PR #14605, opened 2022-11-07
ARROW-18109: [Go] Initial Unary Arithmetic

zeroshade:


go/arrow/compute/internal/kernels/_lib/base_arithmetic.cc
```diff
@@ -125,6 +133,69 @@ struct MultiplyChecked {
...
 };
 
+struct AbsoluteValue {
+    template <typename T, typename Arg>
+    static constexpr T Call(Arg input) {
+        if constexpr(is_same_v<Arg, float>) {
+            *(((int*)&input)+0) &= 0x7fffffff;
+            return input;
+        } else if constexpr(is_same_v<Arg, double>) {
+            *(((int*)&input)+1) &= 0x7fffffff;
```
cyb70289, line 143: Manipulating the sign bit looks hacking. Type punning is not safe and prohibits optimizations (have to move data between floating point and general purpose registers back and forth).
https://godbolt.org/z/srd3jjvMj

Also, I'm surprised s390x CI passed, looks this code should break on big-endian machines.

zeroshade: This C++ code is only compiled and run on amd64 and arm64 machines which are both little-endian. s390x gets the pure Go implementation

zeroshade: Also, I'm using clang, not gcc and it looks like clang recognizes it and optimizes correctly: https://godbolt.org/z/v1Eha8r41

I spot the same `addps` in the output for the x86-64 assembly only it's vectorized so it's `vaddps`

cyb70289 (approved):

PR #14607, opened 2022-11-08
MINOR: [R][CI] Use `docker.io/library/r-base` instead of `docker.io/rocker/r-base`

eitsupi:
Unless there is a special reason, it is better to use `docker.io/library/r-base`, Docker Official Images.
It is built from the same source, but `docker.io/library/r-base` is always up-to-date because it is built automatically.

### DIFF
--- a/dev/tasks/tasks.yml
+++ b/dev/tasks/tasks.yml
@@ -1309,7 +1309,7 @@ tasks:
     ci: azure
     template: r/azure.linux.yml
     params:
-      r_org: rocker
+      r_org: library
       r_image: r-base
       r_tag: latest
       flags: '-e ARROW_DEPENDENCY_SOURCE=BUNDLED'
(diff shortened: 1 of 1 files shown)


raulcd:
@assignUser @nealrichardson you might want to take a look on this one

assignUser:
It is maintained by the rocker community so they are probably updated pretty much at the same time. But the library/r-base image is a multich arch image which is actually a reason to switch. We would  not be taking advantage of it immediatly but it should make it easier to test R stuff on different architectures.

image

assignUser:
@github-actions crossbow submit  test-r-library-* test-r-*minimal*

thisisnic:
@assignUser Will this PR fix the problem alluded to [here](https://github.com/apache/arrow/pull/14654#issuecomment-1317010006) or do we still need to do something else?

kou:
This is not related to https://github.com/apache/arrow/pull/14654#issuecomment-1317010006 because this doesn't change `dev/tasks/r/github.packages.yml`.

eitsupi:
@assignUser @kou Can you merge this? This is a really minor change.

assignUser (approved):

kou (approved):
+1

PR #14609, opened 2022-11-08
ARROW-18287: [C++][CMake] Add support for Brotli/utf8proc provided by vcpkg

kou:
vcpkg provides CMake packages for Brotli and utf8proc with "unofficial-" prefix. So we need to care about them separately.

See also: https://github.com/microsoft/vcpkg/pull/27687


kou:
+1

PR #14610, opened 2022-11-09
ARROW-18289: [Release][vcpkg] Add a script to update vcpkg's arrow port

kou:


kou (commented):
+1

assignUser:
Oh I totally missed this! Very nice! We should add this to the release guide! I will open an issue.

PR #14613, opened 2022-11-09
ARROW-18152: [Python] DataFrame Interchange Protocol for pyarrow Table

AlenkaF:
### Produce a `__dataframe__` object
- [x] Implement the `DataFrame`, `Column` and `Buffers` class
- [x] Test `pa.Table` -> `pd.DataFrame`

What should be looked into after the initial test:
- [x] Data without missing values (produce a validity buffer in case of no missing values)
_Update:_ Columns without missing values are defined as non-nullable for now.
- [x] Boolean values do not transfer correctly (only the first element is produced)
_Update:_ casting boolean column/array to `uint8` solves this issue (boolean arrays are bit packed which is not supported by the protocol)
- [x] Variable-length strings 
_Update:_  Bit-width for the offset buffer dtype must be set to 32 instead of 64.
- [x] DictionaryArray
_Update:_  Pandas implementation seems to expect the column of categories to be an instance of PandasColumn instead of general `__dataframe__` column object.
```python
File "/Users/alenkafrim/repos/pyarrow-dev-9/lib/python3.9/site-packages/pandas/core/interchange/from_dataframe.py", line 185, in categorical_column_to_series
    assert isinstance(cat_column, PandasColumn), "categories must be a PandasColumn"
AssertionError: categories must be a PandasColumn
```
- [x] Bitmasks
_Update:_ Pandas implementation doesn't yet support bitmasks:
```python
File "/Users/alenkafrim/repos/pyarrow-dev-9/lib/python3.9/site-packages/pandas/core/interchange/from_dataframe.py", line 395, in buffer_to_ndarray
    raise NotImplementedError(f"Conversion for {dtype} is not yet supported.")
NotImplementedError: Conversion for (<DtypeKind.BOOL: 20>, 1, 'b', '=') is not yet supported.
```

---

This code in the PR tested with pandas implementation as a consumer currently works with **integers**, **floats**, **booleans**, **strings** and **timestamps** without missing values:
```python
import pyarrow as pa
import pandas as pd
from datetime import datetime as dt

table = pa.table(
    {
        "a": [1, 2, 3, 4],  # dtype kind INT = 0
        "b": [3, 4, 5, 6],  # dtype kind INT = 0
        "c": [1.5, 2.5, 3.5, 4.5],  # dtype kind FLOAT = 2
        "d": [9, 10, 11, 12],  # dtype kind INT = 0
        "e": [True, True, False, False],  # dtype kind BOOLEAN = 20
        "f": ["a", "", "c", "d"],  # dtype kind STRING = 21
        "g": [dt(2007, 7, 13), dt(2007, 7, 14),
              dt(2007, 7, 15), dt(2007, 7, 16)] # dtype kind DATETIME = 22
    }
)

exchange_df = table.__dataframe__()
exchange_df._df

# pyarrow.Table
# a: int64
# b: int64
# c: double
# d: int64
# e: bool
# f: string
# g: timestamp[us]
# ----
# a: [[1,2,3,4]]
# b: [[3,4,5,6]]
# c: [[1.5,2.5,3.5,4.5]]
# d: [[9,10,11,12]]
# e: [[true,true,false,false]]
# f: [["a","","c","d"]]
# g: [[2007-07-13 00:00:00.000000,2007-07-14 00:00:00.000000,2007-07-15 00:00:00.000000,2007-07-16 00:00:00.000000]]

from pandas.core.interchange.from_dataframe import from_dataframe
from_dataframe(exchange_df)
   a  b    c   d  e  f          g
0  1  3  1.5   9  1  a 2007-07-13
1  2  4  2.5  10  1    2007-07-14
2  3  5  3.5  11  0  c 2007-07-15
3  4  6  4.5  12  0  d 2007-07-16
```

---

### Consume a `__dataframe__` object
- [ ] Implement  `from_dataframe` method


python/pyarrow/interchange/column.py
```diff
@@ -0,0 +1,358 @@
...
+        # Store the column as a private attribute
+        self._col = column
+        self._allow_copy = allow_copy
+
+    def size(self) -> int:
+        """
+        Size of the column, in elements.
+        """
+        if isinstance(self._col, pa.Array):
+            len = self._col.to_numpy(zero_copy_only=False).size
```
AlenkaF, line 92: There should be a better way to do this.

python/pyarrow/interchange/column.py
```diff
@@ -0,0 +1,362 @@
...
+        """
+        dtype = self._col.type
+        try:
+            bit_width = dtype.bit_width
+        except ValueError:  # in case of a variable-length strings
+            bit_width = 8
+        # In case of bool data type, bit_width is 1 and has to be multiplied
+        # by 8 (why is that not the case for other dtypes?)
+        if pa.types.is_boolean(dtype):
+            bit_width *= 8
```
AlenkaF, line 140: This is my initial idea of the `bit_width` definition but still needs to be tested and corrected.

python/pyarrow/interchange/column.py
```diff
@@ -0,0 +1,362 @@
...
+            i = 0
+            for start in range(0, chunk_size * n_chunks, chunk_size):
+                yield PyArrowColumn(
+                    array.slice(start, chunk_size), self._allow_copy
+                )
+                i += 1
+            # In case when the size of the chunk is such that the resulting
+            # list is one less chunk then n_chunks -> append an empty chunk
+            if i == n_chunks - 1:
+                yield PyArrowColumn(pa.array([]), self._allow_copy)
```
AlenkaF, line 258: Example: selecting 5 chunks for an array of length 12. If `chunk_size=2`, we get 6 chunks, if `chunk_size=3`, we get 4 chunks =) So we end up producing 4 chunks with `chunk_size=3` plus an empty chunk.

python/pyarrow/interchange/dataframe.py
```diff
@@ -0,0 +1,132 @@
...
+    @property
+    def metadata(self) -> dict[str, Any]:
+        # The metadata for the data frame, as a dictionary with string keys.
+        # Add schema metadata here (pandas metadata, ot custom metadata)
+        if self._df.schema.metadata:
+            schema_metadata = {k.decode('utf8'): v.decode('utf8')
+                           for k, v in self._df.schema.metadata.items()}
+            return schema_metadata
+        else:
+            return {}
```
AlenkaF, line 71: Added this part after testing `pa.Table -> pd.DataFrame` to not get an error for reading the metadata from a NoneType.

python/pyarrow/interchange/column.py
jorisvandenbossche: I think `len(self._col)` should work (in contrast to Table, the Array doesn't have a explicit attribute for it)

honno:
FWIW we have a compliance suite for interchange protocol adopters over at [data-apis/dataframe-interchange-tests](https://github.com/data-apis/dataframe-interchange-tests). It's a bit awkward to use as you'd have to write a compatibility layer in `wrappers.py`, but might be interesting. (Cool to see you're working on this!)

AlenkaF:
Thanks for the info @honno, the compliance suite is definitely something we have to use!
(Cool to see you continue working on the data API standards! 😉 )

AlenkaF:
Will close this PR as I moved the work into another branch: https://github.com/apache/arrow/pull/14804

PR #14614, opened 2022-11-09
ARROW-18251: [CI][Python] Fix AMD64 macOS 12 Python 3 job

raulcd:


jorisvandenbossche:
Hmm, that doesn't seem to have helped. Looking back at the two logs I linked in the JIRA, also the setuptools version changed between those two (65.4.1 -> 65.5.0).

assignUser:
It looks like there is [an issue](https://github.com/apache/arrow/actions/runs/3428644826/jobs/5713204152#step:5:1436) with the naming schema of the `libarrow_dataset.<version>.dylib` : ` [411/465] Linking CXX shared library debug/libarrow_dataset.1100.0.0.dylib` vs `Library not loaded: '@rpath/libarrow_dataset.1100.dylib'`

raulcd:
@jorisvandenbossche @AlenkaF this is now failing on the cython test as it was before. Any idea around that? https://github.com/apache/arrow/actions/runs/3429883915/jobs/5716110145
It also fails the cython test if we use Python 3.10 as seen on this run with the same issue: https://github.com/apache/arrow/actions/runs/3428990620/jobs/5714012255
Should I set `PYARROW_TEST_CYTHON=OFF` temporarily and tackle this on a different PR?

jorisvandenbossche:
Yes, that's the expected failure again. Let's deal with that for a separate JIRA/PR.

>  Should I set PYARROW_TEST_CYTHON=OFF temporarily and tackle this on a different PR?

I am not sure if that env variable is actually still doing something. Maybe the easiest is to pass a flag to pytest to explicitly skip that one test (something like `-k "not test_cython_api"`)

jorisvandenbossche:
Although it _is_ curious why `PYARROW_BUNDLE_ARROW_CPP=ON` is needed only for Python 3.11 ...

raulcd:
> Although it _is_ curious why `PYARROW_BUNDLE_ARROW_CPP=ON` is needed only for Python 3.11 ...

Agree, not sure why is necessary only for Python 3.11

raulcd:
@jorisvandenbossche the job is successful now (skipping the cython test). The other job failure feels unrelated and seems to be a temporary connection issue as it worked on the previous commit.

kou (approved):
+1

For `PYARROW_BUNDLE_ARROW_CPP`, we may be able to remove it by the following patch:

```diff
diff --git a/ci/scripts/python_test.sh b/ci/scripts/python_test.sh
index 824865b923..53505875e8 100755
--- a/ci/scripts/python_test.sh
+++ b/ci/scripts/python_test.sh
@@ -26,6 +26,7 @@ export ARROW_SOURCE_DIR=${arrow_dir}
 export ARROW_TEST_DATA=${arrow_dir}/testing/data
 export PARQUET_TEST_DATA=${arrow_dir}/cpp/submodules/parquet-testing/data
 export LD_LIBRARY_PATH=${ARROW_HOME}/lib:${LD_LIBRARY_PATH}
+export DYLD_LIBRARY_PATH=${ARROW_HOME}/lib${DYLD_LIBRARY_PATH:+:${DYLD_LIBRARY_PATH}}
 export ARROW_GDB_SCRIPT=${arrow_dir}/cpp/gdb_arrow.py
 
 # Enable some checks inside Python itself
```

This may also fix the Cython test failure.

We can try it in a follow-up task.

PR #14615, opened 2022-11-09
ARROW-18285: [R] Fix for failing test after lubridate 1.9 release

paleolimbot:
This fixes test failures resulting from an updated lubridate package, whose update removed some functionality that we supported (apparently `lubridate::yq(2021.1)` was a thing). This works in our source because we cast to `string()` before doing any further processing.


nealrichardson:
On the current test error, it looks like some of the floats do get handled by lubridate, only 2 of the values don't parse--so is removal the right fix?

rok (commented):
`yq`/`Yq` on string still seems to be a thing? https://github.com/tidyverse/lubridate/blob/main/tests/testthat/test-parsers.R#L746-L753
I think the float option is odd but don't have strong feelings on keeping/cutting it.

rok:
Sorry I misread the float/string part.

nealrichardson:
I agree it's odd but it does seem like lubridate does still support it. An alternative to removing the feature would be to changing the test not to have trailing zeros in the test data, since those are supported the same. I might suggest that as the quick fix for the test failures. 

In terms of ongoing maintenance, maybe you could file an issue on lubridate (or check to see if one exists) and see if this is something they intend to support. We know they plan a major refactor soon, so maybe the maintainers can tell us if this is going to stay or go, and we can drop it or not based on that.

PR #14602, opened 2022-11-07
ARROW-18270: [Python] Remove gcc 4.9 compatibility code

pitrou:
We can use std::regex now that we needn't support gcc 4.9 anymore.

PR #14626, opened 2022-11-13
MINOR: [C++] Mark header-defined function inline

pitrou:
Without `inline`, this would produce multiple definition error when used from several compilation units.

Followup to https://github.com/apache/arrow/pull/14602


pitrou:
+1, will merge if CI green

PR #14632, opened 2022-11-14
MINOR: [Docs][C++] Improve docs for ConcatenateTables

rok:
This is to improve [ConcatenateTables docs](https://github.com/apache/arrow/blob/9d86ede69af14eacc7968684b7e9535cfc8e3907/cpp/src/arrow/table.h#L291).

### DIFF
--- a/cpp/src/arrow/table.h
+++ b/cpp/src/arrow/table.h
@@ -286,7 +286,18 @@ struct ARROW_EXPORT ConcatenateTablesOptions {
   static ConcatenateTablesOptions Defaults() { return {}; }
 };
 
-/// \brief Construct table from multiple input tables.
+/// \brief Construct a new table from multiple input tables. Underlying column arrays
+/// are concatenated without copying except in case of schema missmatch in which case
+/// columns will be cast to promoted types which will cause a copy. Tables will be
+/// concatenated in order they are provided in and order of rows within tables will be
+/// preserved.
+///
+/// \param[in] tables a std::vector of Tables to be concatenated
+/// \param[in] options specify how to unify schema of input tables
+/// \param[in] memory_pool MemoryPool to be used if null-filled arrays need to
+/// be created.
+/// \return Status
+
 ARROW_EXPORT
 Result<std::shared_ptr<Table>> ConcatenateTables(
     const std::vector<std::shared_ptr<Table>>& tables,
--- a/cpp/src/arrow/table.h
+++ b/cpp/src/arrow/table.h
@@ -286,16 +286,20 @@ struct ARROW_EXPORT ConcatenateTablesOptions {
   static ConcatenateTablesOptions Defaults() { return {}; }
 };
 
-/// \brief Construct a new table from multiple input tables. Underlying column arrays
-/// are concatenated without copying except in case of schema missmatch in which case
-/// columns will be cast to promoted types which will cause a copy. Tables will be
-/// concatenated in order they are provided in and order of rows within tables will be
-/// preserved.
+/// \brief Construct a new table from multiple input tables.
+///
+/// The new table is assembled from existing column chunks without copying,
+/// if schemas are identical. If schemas do not match exactly, an attempt is
+/// made to unify them, and then column chunks are converted to their
+/// respective unified datatype, which will probably incur a copy.
+///
+/// Tables are concatenated in order they are provided in and the order of
+/// rows within tables will be preserved.
 ///
 /// \param[in] tables a std::vector of Tables to be concatenated
 /// \param[in] options specify how to unify schema of input tables
 /// \param[in] memory_pool MemoryPool to be used if null-filled arrays need to
-/// be created.
+/// be created or if existing column chunks need to endure type conversion
 /// \return Status
 
 ARROW_EXPORT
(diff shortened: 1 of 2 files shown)


cpp/src/arrow/table.h
```diff
@@ -286,7 +286,18 @@ struct ARROW_EXPORT ConcatenateTablesOptions {
   static ConcatenateTablesOptions Defaults() { return {}; }
 };
 
-/// \brief Construct table from multiple input tables.
+/// \brief Construct a new table from multiple input tables. Underlying column arrays
+/// are concatenated without copying except in case of schema missmatch in which case
+/// columns will be cast to promoted types which will cause a copy. Tables will be
+/// concatenated in order they are provided in and order of rows within tables will be
+/// preserved.
```
pitrou, line 293: We should keep the brief line brief. I'd also like to suggest some wording changes.

```suggestion
/// \brief Construct a new table from multiple input tables.
///
/// The new table is assembled from existing column chunks without copying,
/// if schemas are identical. If schemas do not match exactly, an attempt is
/// made to unify them, and then column chunks are converted to their
/// respective unified datatype, which will probably incur a copy.
///
/// Tables are concatenated in order they are provided in and the order of
/// rows within tables will be preserved.
```

cpp/src/arrow/table.h
```diff
@@ -286,7 +286,18 @@ struct ARROW_EXPORT ConcatenateTablesOptions {
...
+///
+/// \param[in] tables a std::vector of Tables to be concatenated
+/// \param[in] options specify how to unify schema of input tables
+/// \param[in] memory_pool MemoryPool to be used if null-filled arrays need to
+/// be created.
```
pitrou, line 298:
```suggestion
/// \param[in] memory_pool MemoryPool to be used if null-filled arrays need to
/// be created or if existing column chunks need to be endure type conversion
```

cpp/src/arrow/table.h
```diff
@@ -286,7 +286,18 @@ struct ARROW_EXPORT ConcatenateTablesOptions {
...
+/// \return Status
```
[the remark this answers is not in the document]
rok:
```suggestion
/// \return new Table
```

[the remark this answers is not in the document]
rok: Ugh sorry.

cpp/src/arrow/table.h
```diff
@@ -286,7 +286,22 @@ struct ARROW_EXPORT ConcatenateTablesOptions {
   static ConcatenateTablesOptions Defaults() { return {}; }
 };
 
-/// \brief Construct table from multiple input tables.
+/// \brief Construct a new table from multiple input tables.
+///
+/// The new table is assembled from existing column chunks without copying,
+/// if schemas are identical. If schemas do not match exactly, an attempt is
+/// made to unify them, and then column chunks are converted to their
```
rok, line 293: @pitrou could docs of other functions be referenced from docstrings? E.g. it would make sense to link to docs of [PromoteTableToSchema](https://github.com/apache/arrow/blob/9d86ede69af14eacc7968684b7e9535cfc8e3907/cpp/src/arrow/table.h#L296-L310). Also `PromoteTableToSchema` docs doesn't seem to be public. Is that ideal?

pitrou: Yes, you can reference `PromoteTableToSchema` here. And, yes, we can publish the doc to the latter in the Sphinx API docs.

rok: Added.

rok:

rok:
Docs changes should be [here](http://crossbow.voltrondata.com/pr_docs/14632/cpp/api/table.html#_CPPv4N5arrow17ConcatenateTablesERKNSt6vectorINSt10shared_ptrI5TableEEEE24ConcatenateTablesOptionsP10MemoryPool) once built.

cpp/src/arrow/table.h
```diff
@@ -286,7 +286,23 @@ struct ARROW_EXPORT ConcatenateTablesOptions {
...
+/// respective unified datatype, which will probably incur a copy.
+/// :func:`arrow::PromoteTableToSchema` is used to unify schemas.
```
wjones127, line 295: Schemas will only be unified if requested.

```suggestion
/// if schemas are identical. If schemas do not match exactly and 
/// unifySchemas is enabled in options (off by default), an attempt is
/// made to unify them, and then column chunks are converted to their
/// respective unified datatype, which will probably incur a copy.
/// :func:`arrow::PromoteTableToSchema` is used to unify schemas.
```

cpp/src/arrow/table.h
```diff
@@ -286,7 +286,24 @@ struct ARROW_EXPORT ConcatenateTablesOptions {
   static ConcatenateTablesOptions Defaults() { return {}; }
 };
 
-/// \brief Construct table from multiple input tables.
+/// \brief Construct a new table from multiple input tables.
+///
+/// The new table is assembled from existing column chunks without copying,
+/// if schemas are identical. If schemas do not match exactly and
```
rok, line 292:
```suggestion
/// if schemas are identical. If schemas do not match exactly and
```

wjones127 (approved):
Thanks, Rok! LGTM

rok:
Thanks all! :)

PR #14633, opened 2022-11-14
ARROW-18257: [Python] pass back time types with correct type class

0x26res:
https://issues.apache.org/jira/browse/ARROW-18257

### DIFF
--- a/python/pyarrow/public-api.pxi
+++ b/python/pyarrow/public-api.pxi
@@ -97,6 +97,10 @@ cdef api object pyarrow_wrap_data_type(
         out = SparseUnionType.__new__(SparseUnionType)
     elif type.get().id() == _Type_DENSE_UNION:
         out = DenseUnionType.__new__(DenseUnionType)
+    elif type.get().id() == _Type_TIME32:
+        out = Time32Type.__new__(Time32Type)
+    elif type.get().id() == _Type_TIME64:
+        out = Time64Type.__new__(Time64Type)
     elif type.get().id() == _Type_TIMESTAMP:
         out = TimestampType.__new__(TimestampType)
     elif type.get().id() == _Type_DURATION:
--- a/python/pyarrow/tests/test_types.py
+++ b/python/pyarrow/tests/test_types.py
@@ -1138,17 +1138,3 @@ def test_types_come_back_with_specific_type():
         schema = pa.schema([pa.field("field_name", arrow_type)])
         type_back = schema.field("field_name").type
         assert type(type_back) is type(arrow_type)
-
-
-@pytest.mark.parametrize("unit", ["us", "ns"])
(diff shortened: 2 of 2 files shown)


python/pyarrow/public-api.pxi
```diff
@@ -97,6 +97,10 @@ cdef api object pyarrow_wrap_data_type(
         out = SparseUnionType.__new__(SparseUnionType)
     elif type.get().id() == _Type_DENSE_UNION:
         out = DenseUnionType.__new__(DenseUnionType)
+    elif type.get().id() == _Type_TIME32:
+        out = Time32Type.__new__(Time32Type)
+    elif type.get().id() == _Type_TIME64:
+        out = Time64Type.__new__(Time64Type)
```
0x26res, line 103: @jorisvandenbossche I've noticed that date32 and date64 are missing here as well. But they would require more effort because Date32Type and Date64Type don't exist. I think I would need to introduced them in `types.pxi`. I'm not sure if this is necessary though, given these types don't have a custom field (unlike Time32Type which has a `unit`).

Also should the if statement in that function be in any specific order?

jorisvandenbossche: > But they would require more effort because Date32Type and Date64Type don't exist. I'm not sure if this is necessary though, given these types don't have a custom field (unlike Time32Type which has a

Yes, that's indeed because they don't have any custom logic or attributes (just like primitive numeric data types also don't have a custom subclass)

> Also should the if statement in that function be in any specific order?

I don't think so, current order looks fine

python/pyarrow/tests/test_types.py
```diff
@@ -1131,3 +1131,24 @@ def test_hashing(items):
...
+    array = pa.array([1, 2, 3], pa.time64(unit))
+    assert array.type.unit == unit
+    assert isinstance(array.type, pa.Time64Type)
+
+
+@pytest.mark.parametrize("unit", ["s", "ms"])
+def test_can_read_unit_from_time_32_array(unit: str):
+    array = pa.array([1, 2, 3], pa.time32(unit))
+    assert array.type.unit == unit
+    assert isinstance(array.type, pa.Time32Type)
```
jorisvandenbossche, line 1154: I am not fully sure those two tests are needed. We already have `test_time32_units` and `test_time64_units` which cover the fact that those subclasses have a `unit` property. And the test you added above will ensure that this subclass is preserved when getting the type from an array of schema.

jorisvandenbossche (commented):
Thanks for the PR!

python/pyarrow/tests/test_types.py
0x26res: I've removed these 2.

jorisvandenbossche (approved):
Thanks!

PR #14645, opened 2022-11-15
ARROW-18290: [C++] Escape all special chars in URI-encoding

pitrou:
By default, the AWS SDK doesn't URI-escape some special chars such as '='. This works fine with AWS endpoints, but not with some non-AWS S3-compatible endpoints.


milesgranger (approved):
Works for me. :+1:  (testing with 1.10.7 SDK)

Edit: Also, thank you! :)

pitrou:
Great, thank you! Will merge then.

PR #14646, opened 2022-11-15
ARROW-18269: [C++] Handle slash character in Hive-style partition values

vibhatha:
This PR includes a modification to the partition key generation. Expecting input data column with URI data for partition, the URI encoding is added to consider that situation.

### DIFF
--- a/cpp/src/arrow/dataset/partition.cc
+++ b/cpp/src/arrow/dataset/partition.cc
@@ -299,7 +299,7 @@ Result<PartitionPathFormat> KeyValuePartitioning::Format(
 
     ARROW_ASSIGN_OR_RAISE(auto match, ref_value.first.FindOneOrNone(*schema_));
     if (match.empty()) continue;
-    // VIBHATHA: upto this point the / is preserved A/Z like things are still there
+
     auto value = ref_value.second.scalar();
 
     const auto& field = schema_->field(match[0]);
@@ -452,7 +452,6 @@ HivePartitioningOptions HivePartitioningFactoryOptions::AsHivePartitioningOption
   HivePartitioningOptions options;
   options.segment_encoding = segment_encoding;
   options.null_fallback = null_fallback;
-  options.split_path = !ignore_separator;
   return options;
 }
 
@@ -752,18 +751,12 @@ Result<std::optional<KeyValuePartitioning::Key>> HivePartitioning::ParseKey(
   return Key{std::move(name), std::move(value)};
 }
 
-//VIBHATHA:CODE: here the segments extraction part shouldn't remove the slashes 
 Result<std::vector<KeyValuePartitioning::Key>> HivePartitioning::ParseKeys(
     const std::string& path) const {
   std::vector<Key> keys;
-  auto parent_path = fs::internal::GetAbstractPathParent(path).first;
-  std::vector<std::string> segments;
-  if (hive_options_.split_path) {
-    segments = fs::internal::SplitAbstractPath(parent_path);
-  } else {
-    segments.push_back(std::move(parent_path));
-  }
-  for (const auto& segment : segments) {
+
+  for (const auto& segment :
+       fs::internal::SplitAbstractPath(fs::internal::GetAbstractPathParent(path).first)) {
     ARROW_ASSIGN_OR_RAISE(auto maybe_key, ParseKey(segment, hive_options_));
     if (auto key = maybe_key) {
       keys.push_back(std::move(*key));
@@ -779,6 +772,7 @@ Result<PartitionPathFormat> HivePartitioning::FormatValues(
 
   for (int i = 0; i < schema_->num_fields(); ++i) {
     const std::string& name = schema_->field(i)->name();
+
     if (values[i] == nullptr) {
       segments[i] = "";
     } else if (!values[i]->is_valid) {
@@ -786,18 +780,9 @@ Result<PartitionPathFormat> HivePartitioning::FormatValues(
       // field_index <-> path nesting relation
       segments[i] = name + "=" + hive_options_.null_fallback;
     } else {
-      segments[i] = name + "=" + values[i]->ToString();
+      segments[i] = name + "=" + arrow::internal::UriEscape(values[i]->ToString());
     }
   }
-
-  // auto flatten_segments = [&]() -> std::string {
-  //   std::string flattened_segment = "";
-  //   std::hash<std::string> str_hash;
-  //   for(const std::string& segment : segments) {
-  //     flattened_segment += str_hash(segment);
-  //   }
-  //   return flattened_segment;
-  // };
 
   return PartitionPathFormat{fs::internal::JoinAbstractPath(std::move(segments)), ""};
 }
@@ -826,8 +811,6 @@ class HivePartitioningFactory : public KeyValuePartitioningFactory {
       const std::vector<std::string>& paths) override {
     auto options = options_.AsHivePartitioningOptions();
     for (auto path : paths) {
-      /// VIBHATHA: I think the issue here is that generally the path string
-      /// contains the schema_field=
       for (auto&& segment : fs::internal::SplitAbstractPath(path)) {
         ARROW_ASSIGN_OR_RAISE(auto maybe_key,
                               HivePartitioning::ParseKey(segment, options));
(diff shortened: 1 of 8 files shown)


python/pyarrow/tests/test_dataset.py
```diff
@@ -4912,3 +4912,33 @@ def test_read_table_nested_columns(tempdir, format):
...
+def test_dataset_partition_with_slash(tmpdir):
+    from pyarrow import dataset as ds
+
+    path = tmpdir / "slash-writer-x"
+
+    dt_table = pa.Table.from_arrays([
+        pa.array([1, 2, 3, 4, 5], pa.int32()),
+        pa.array(["experiment/A/f.csv", "experiment/B/f.csv",
+                  "experiment/A/d.csv", "experiment/C/k.csv",
+                  "experiment/M/i.csv"], pa.utf8())], ["exp_id", "exp_meta"])
```
westonpace, line 4926: Can you make it so at least two rows end up in the same partition?  E.g. you could just change `experiment/A/d.csv` to `experiment/A/f.csv`.

westonpace (commented):
This fixes hive partitioning.  Do we need to also fix directory partitioning or filename partitioning?

python/pyarrow/tests/test_dataset.py
vibhatha: Sure, I can do that

vibhatha:
> This fixes hive partitioning. Do we need to also fix directory partitioning or filename partitioning?

Good question. Should we do in a separate PR?

westonpace:
> Good question. Should we do in a separate PR?

I think that's fine.  I also don't know if we do the URI decoding with directory or filename partitioning.  If I recall, URI decoding was originally added specifically for compatibility with the way that Spark wrote datasets.

vibhatha:
> > Good question. Should we do in a separate PR?
> 
> I think that's fine. I also don't know if we do the URI decoding with directory or filename partitioning. If I recall, URI decoding was originally added specifically for compatibility with the way that Spark wrote datasets.

Ah I see, that's good to know. I created a [JIRA](https://issues.apache.org/jira/browse/ARROW-18338) here and I will work on this issue.

python/pyarrow/tests/test_dataset.py
```diff
@@ -4912,3 +4912,33 @@ def test_read_table_nested_columns(tempdir, format):
...
+    )
+
+    read_table = ds.dataset(
+        source=path,
+        format='parquet',
+        partitioning='hive',
+        schema=pa.schema([pa.field("exp_id", pa.int32()),
+                         pa.field("exp_meta", pa.utf8())])
+    ).to_table().combine_chunks()
+
```
wjones127, line 4943: Could we also assert what the escaped partition directories are named? Given we are trying to be compatible with other systems, it seems like it would be wise to enforce what the Uri-encoded form is, rather than just asserting we can roundtrip.

wjones127 (changes requested):

python/pyarrow/tests/test_dataset.py
vibhatha: Shouldn’t it be handled by the encoder test cases?

wjones127: That tests that the encoding function works, but not that it is used to write the directory names.

I was thinking of a test like this:

```python
   file_paths = list(path.iterdir())
   assert Path("exp_meta=experiment%2FA%2Ff.csv/part-0.parquet") in file_paths
```

vibhatha: I see your point. We could write a test, then it should reflect platform dependent test cases too?

vibhatha: I added a check.

wjones127 (approved):
Thanks for the new test. This looks good!

cpp/src/arrow/dataset/partition.cc
```diff
@@ -780,7 +780,7 @@ Result<PartitionPathFormat> HivePartitioning::FormatValues(
       // field_index <-> path nesting relation
       segments[i] = name + "=" + hive_options_.null_fallback;
     } else {
-      segments[i] = name + "=" + values[i]->ToString();
+      segments[i] = name + "=" + arrow::internal::UriEscape(values[i]->ToString());
```
pitrou, line 783: Can you add tests on the C++ side to make sure it is covered?

python/pyarrow/tests/test_dataset.py
```diff
@@ -4912,3 +4914,40 @@ def test_read_table_nested_columns(tempdir, format):
...
+    assert dt_table == read_table.sort_by("exp_id")
+
+    exp_meta = dt_table.column(1).to_pylist()
+    exp_meta = sorted(list(dict.fromkeys(exp_meta)))  # take unique
```
pitrou, line 4949:
```suggestion
    exp_meta = sorted(set(exp_meta))  # take unique
```

python/pyarrow/tests/test_dataset.py
```diff
@@ -4912,3 +4914,40 @@ def test_read_table_nested_columns(tempdir, format):
...
+    encoded_paths = ["exp_meta="+quote(path, safe='') for path in exp_meta]
```
pitrou, line 4950: Style:
```suggestion
    encoded_paths = ["exp_meta=" + quote(path, safe='') for path in exp_meta]
```

python/pyarrow/tests/test_dataset.py
```diff
@@ -4912,3 +4914,40 @@ def test_read_table_nested_columns(tempdir, format):
...
+        partitioning=['exp_meta'],
+        partitioning_flavor='hive',
+    )
+
+    read_table = ds.dataset(
+        source=path,
+        format='parquet',
+        partitioning='hive',
+        schema=pa.schema([pa.field("exp_id", pa.int32()),
+                         pa.field("exp_meta", pa.utf8())])
```
pitrou, line 4943: Style
```suggestion
        schema=pa.schema([pa.field("exp_id", pa.int32()),
                          pa.field("exp_meta", pa.utf8())])
```

python/pyarrow/tests/test_dataset.py
```diff
@@ -4912,3 +4914,40 @@ def test_read_table_nested_columns(tempdir, format):
...
+    dt_table = pa.Table.from_arrays([
+        pa.array([1, 2, 3, 4, 5], pa.int32()),
+        pa.array(["experiment/A/f.csv", "experiment/B/f.csv",
+                  "experiment/A/f.csv", "experiment/C/k.csv",
+                  "experiment/M/i.csv"], pa.utf8())], ["exp_id", "exp_meta"])
+
+    ds.write_dataset(
+        data=dt_table,
+        base_dir=path,
+        format='parquet',
```
pitrou, line 4933: Can you use "ipc" so that this can run even if Parquet is disabled?

pitrou (changes requested):

cpp/src/arrow/dataset/partition.cc
vibhatha: Sure I will add

python/pyarrow/tests/test_dataset.py
vibhatha: I used archery lint, though didn't format it like this, but I changed manually.

cpp/src/arrow/dataset/partition_test.cc
```diff
@@ -1048,5 +1051,60 @@ TEST(TestStripPrefixAndFilename, Basic) {
...
+  arrow::NumericBuilder<arrow::Int64Type> builder;
+  ASSERT_OK(builder.AppendValues({0, 1, 2, 3, 4}));
+  ASSERT_OK(builder.Finish(&arrays[0]));
+  arrow::StringBuilder string_builder;
+  std::vector<std::string> partitions = {"experiment/A/f.csv", "experiment/B/f.csv",
+                                         "experiment/A/f.csv", "experiment/C/k.csv",
+                                         "experiment/M/i.csv"};
+  ASSERT_OK(string_builder.AppendValues(partitions));
+  ASSERT_OK(string_builder.Finish(&arrays[1]));
+  auto table = arrow::Table::Make(schema, arrays);
```
pitrou, line 1073: Can we use `ArrayFromJSON` or even `TableFromJSON` instead of building these step-by-step?

pitrou (changes requested):
Looks mostly good, thank you!

cpp/src/arrow/dataset/partition_test.cc
vibhatha: Good catch. We should be able to.

pitrou (approved):

PR #14550, opened 2022-10-29
ARROW-17301: [C++] Implement compute function "binary_slice"

kshitij12345:
Implements `binary_slice_bytes` similar to `utf8_slice_codeunits`.

Mostly based on `utf8_slice_codeunits`.

TODO:
* [x] C++ Tests 
* [x] Python Tests

PR #14647, opened 2022-11-15
ARROW-18321: [R] Add tests for binary_slice kernel

paleolimbot:
This PR adds a test for the kernel implemented in ARROW-17301 (#14550).

Are there any R functions that map to this? (i.e., is there any `register_binding()` we should do for an R function?)

### DIFF
--- a/r/R/dplyr-funcs-string.R
+++ b/r/R/dplyr-funcs-string.R
@@ -495,8 +495,19 @@ register_bindings_string_other <- function() {
         stop <- 0
       }
 
+      # if the input is a string we use "utf8_slice_codeunits"; if the
+      # input is binary we use "binary_slice". This does not consider
+      # a binary Scalar.
+      x_is_binary <- inherits(x, "ArrowObject") &&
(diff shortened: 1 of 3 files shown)


nealrichardson:
> Are there any R functions that map to this? (i.e., is there any `register_binding()` we should do for an R function?)

`substr()`? Assuming this also works on `string()` type?

nealrichardson:
> We already do `substr()` (which maps to the `utf8_slice_codeunits()` function...this doesn't work on a string type).
> 

Ah ok. Would it make sense to call `binary_slice` inside of `substr()` if the type of the input is `binary()`? Or maybe it's not worth bothering with.

> Fun discovery in the process, though:
> 
> ```r
> substr(list(as.raw(1:5)), 1, 100)
> #> [1] "as.raw(c(0x01, 0x02, 0x03, 0x04, 0x05))"
> ```

Yeah that is truly odd. Without `list()` it is more sensible.

```
> as.character(list(as.raw(1:5)))
[1] "as.raw(c(0x01, 0x02, 0x03, 0x04, 0x05))"
> as.character(as.raw(1:5))
[1] "01" "02" "03" "04" "05"
```

thisisnic (approved):
LGTM!

PR #14655, opened 2022-11-16
[MINOR] Add Flight SQL Example Link

thatstatsguy:
Add Go example to documentation for Flight SQL

### DIFF
--- a/docs/source/format/FlightSql.rst
+++ b/docs/source/format/FlightSql.rst
@@ -35,6 +35,13 @@ new RPC methods described here.
 .. warning:: Flight SQL is **experimental** and changes to the
              protocol may still be made.
 
+Examples
+===========
+
+Examples for implementing Flight SQL are available in the following languages:
+
+- `Go <https://github.com/apache/arrow/tree/master/go/arrow/flight/flightsql/example/>`_
+
 RPC Methods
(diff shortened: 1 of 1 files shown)


lidavidm:
Thanks!

This is meant as a specification document. Perhaps the right place to put this is in the Go docs themselves?

thatstatsguy:
Thanks for the feedback David! 

I guess the underlying issue here is I really struggled to find an example for implementing Apache Arrow SQL. 

I think others may have struggled too and would appreciate a central view of examples available (even if it's still experimental)

No objection to moving it elsewhere given it's a spec doc - just wanted to flag the issue I had :)

lidavidm:
The [Cookbook](https://arrow.apache.org/cookbook/) is what we use for examples. That said Flight SQL examples would probably be too long to fit that well.

I also don't think the examples in the source tree are really meant as examples - at least the C++/Java ones are really harnesses for unit testing.

That said, the example is already in the [Go docs](https://pkg.go.dev/github.com/apache/arrow/go/v10@v10.0.0/arrow/flight/flightsql/example). I would be ok with adding C++/Go/Java links to 'external resources' with a disclaimer.

amol-:

PR #7335, opened 2020-06-03
ARROW-9018: [C++] Remove APIs that were marked as deprecated in 0.17.0 and prior

wesm:
These are mostly Status APIs that now use Result.

PR #6863, opened 2020-04-07
ARROW-8361: [C++] Add Result<T> APIs to Buffer methods and functions

pitrou:

PR #14657, opened 2022-11-16
ARROW-18343: [C++] Remove AllocateBitmap() with out parameter

js8544:
This variant of AllocateBitmap had been declared but not defined in several years. It's therefore safe to assume that noone is relying on it.

### DIFF
--- a/cpp/src/arrow/buffer.cc
+++ b/cpp/src/arrow/buffer.cc
@@ -184,10 +184,6 @@ Result<std::shared_ptr<Buffer>> AllocateBitmap(int64_t length, MemoryPool* pool)
   return std::move(buf);
 }
 
-Status AllocateBitmap(MemoryPool* pool, int64_t length, std::shared_ptr<Buffer>* out) {
-  return AllocateBitmap(length, pool).Value(out);
-}
-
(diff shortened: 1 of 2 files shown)


pitrou:
Can you find out since when that has been the case? If there have been at least two releases without, perhaps we can safely remove it :-)

js8544:
> Can you find out since when that has been the case? If there have been at least two releases without, perhaps we can safely remove it :-)

The definition is removed in [this PR](https://github.com/apache/arrow/pull/7335/files#diff-590dc662378ffd51c79ee710b3058838e95a9eb16877435b131018f0becb9faaL307) but the api is kept. It appeared to be a mistake to remove the definition back then because this api is not marked deprecated in [this PR](https://github.com/apache/arrow/pull/6863/files#diff-2edce6fb9f08ce495de42b07497d9d43d57b12051ef5aa997ad6056eb207b70dR514) (or perhaps it should be marked deprecated but was overlooked). Either way since it has been like this for over 2 years, perhaps it's best to remove this api?

pitrou:
> Either way since it has been like this for over 2 years, perhaps it's best to remove this api now?

Yes, definitely.

js8544:
This api is removed.

pitrou (approved):
Thanks @js8544 !

PR #14658, opened 2022-11-16
ARROW-18342: [C++] AsofJoinNode support for Boolean data field

rtpsw:
See https://issues.apache.org/jira/browse/ARROW-18342

### DIFF
--- a/cpp/src/arrow/compute/exec/asof_join_node_test.cc
+++ b/cpp/src/arrow/compute/exec/asof_join_node_test.cc
@@ -31,6 +31,7 @@
 #include "arrow/compute/exec/util.h"
 #include "arrow/compute/kernels/row_encoder.h"
 #include "arrow/compute/kernels/test_util.h"
+#include "arrow/testing/generator.h"
 #include "arrow/testing/gtest_util.h"
 #include "arrow/testing/matchers.h"
 #include "arrow/testing/random.h"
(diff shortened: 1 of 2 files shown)


cpp/src/arrow/compute/exec/asof_join_node_test.cc
```diff
@@ -83,6 +85,7 @@ Result<BatchesWithSchema> MakeBatchesFromNumString(
   batches.schema = schema;
   int n_fields = schema->num_fields();
   for (auto num_batch : num_batches.batches) {
+    Datum two(ConstantArrayGenerator::Int32(num_batch.length, 2));
```
wjones127, line 88: Why not a scalar here? (and remove `#include "arrow/testing/generator.h"` above)
```suggestion
    Datum two(Int32Scalar(2));
```

wjones127 (commented):
I have two suggestions, but looks good so far.

cpp/src/arrow/compute/exec/asof_join_node_test.cc
```diff
@@ -91,6 +93,18 @@ Result<BatchesWithSchema> MakeBatchesFromNumString(
...
+        // the next 4 lines compute `as_bool` as `(bool)(x - 2*(x/2))`, i.e., the low bit
+        // of `x`. Here, `x` stands for `num_batch.values[i]`, which is an `int64` value.
+        // Taking the low bit is a somewhat arbitrary way of obtaining both `true` and
+        // `false` values from the `int64` values in the test data, in order to get good
+        // testing coverage. A simple cast to a Boolean value would not get good coverage
+        // because all positive values would be cast to `true`.
+        ARROW_ASSIGN_OR_RAISE(Datum div_two, Divide(num_batch.values[i], two));
+        ARROW_ASSIGN_OR_RAISE(Datum rounded, Multiply(div_two, two));
+        ARROW_ASSIGN_OR_RAISE(Datum low_bit, Subtract(num_batch.values[i], rounded));
+        ARROW_ASSIGN_OR_RAISE(Datum as_bool, Cast(low_bit, type));
```
westonpace, line 106: Hmm, we have a bit_wise_and kernel but I don't think there is a C++ helper function (e.g. `BitWiseAnd`) wrapping it.  This is probably fine.

westonpace (approved):

wjones127 (approved):

PR #14659, opened 2022-11-16
ARROW-18334: [C++] Handle potential non-commutativity by rebinding

bkietz:
This ensures that kernels in every bound expression definitely correspond to the argument types after the arguments have been reordered. Note: if the operation is in fact not commutative (for example if we added a string concatenating kernel to add_checked) then Canonicalize will break expressions using that kernel.


cpp/src/arrow/compute/exec/expression.cc
```diff
@@ -634,6 +634,20 @@ std::optional<Out> FoldLeft(It begin, It end, const BinOp& bin_op) {
   return folded;
 }
 
+template <typename BinOp, typename It,
+          typename Value = typename std::iterator_traits<It>::value_type,
+          typename Out = decltype(std::declval<BinOp>()(std::declval<Value>(),
+                                                        std::declval<Value>()))>
+std::optional<Out> MaybeFoldLeft(It begin, It end, const BinOp& bin_op) {
```
westonpace, line 641: Nit: Is this the same thing as `std::accumulate`?  Could we maybe use `MaybeAccumulate` or, in a brief comment, reference `std::accumulate` for new readers who might not be familiar with the term "folding" (I often think of this as "reduce" but I just looked it up and it seems there is a `std::reduce` but it doesn't guarantee **left**-folding and thus can run in parallel)

cpp/src/arrow/compute/exec/expression.cc
```diff
@@ -856,6 +870,27 @@ bool IsBinaryAssociativeCommutative(const Expression::Call& call) {
...
+  // but the mismatch in declared argument types is potentially problematic
+  // if we ever start using the Expression::Call::kernel field more than
+  // we do currently. Check and rebind if necessary.
+  //
+  // The more correct fix for this problem is to ensure that all kernels of
+  // functions which are commutative be commutative as well, which would
+  // obviate rebinding like this. In the context of ARROW-18334, this
+  // would require rewriting KernelSignature so that a single kernel can
+  // handle both `timestamp + duration` and `duration + timestamp`.
+  if (call.kernel->signature->MatchesInputs(GetTypes(call.arguments))) {
```
westonpace, line 888: Potentially naive question: are we guaranteed to be bound at this point (e.g. is `call.kernel` guaranteed to exist)?

cpp/src/arrow/compute/exec/expression_test.cc
```diff
@@ -69,6 +70,39 @@ Expression true_unless_null(Expression argument) {
...
+  } else if constexpr (std::is_same_v<Period, std::micro>) {
+    unit = TimeUnit::MICRO;
+  } else if constexpr (std::is_same_v<Period, std::milli>) {
+    unit = TimeUnit::MILLI;
+  } else {
+    unit = TimeUnit::SECOND;
+    int_value = std::chrono::duration_cast<std::chrono::seconds>(d).count();
+  }
+  return literal(*MakeScalar(int_value)->CastTo(duration(unit)));
+}
```
westonpace, line 104: nit: This seems generally useful (note, I thought that on the earlier version before the literals route too) to be something hidden in a test case.  Maybe a follow-up PR to add tests and expose the functionality at some point (though it's probably just one in a sea of QoL features we could offer C++ users).

cpp/src/arrow/compute/exec/expression_test.cc
```diff
@@ -258,10 +292,11 @@ TEST(Expression, ToString) {
   EXPECT_EQ(literal(std::make_shared<BinaryScalar>(Buffer::FromString("az"))).ToString(),
             "\"617A\"");
 
-  auto ts = *MakeScalar("1990-10-23 10:23:33")->CastTo(timestamp(TimeUnit::NANO));
-  EXPECT_EQ(literal(ts).ToString(), "1990-10-23 10:23:33.000000000");
+  using namespace arrow_literals;
```
westonpace, line 295: This is an unexpected place to find the `using` statement.  Why not pull it outside the test case?

cpp/src/arrow/compute/exec/expression_test.cc
```diff
@@ -1098,6 +1122,12 @@ TEST(Expression, CanonicalizeAnd) {
 
   // catches and_kleene even when it's a subexpression
   ExpectCanonicalizesTo(is_valid(and_(b, true_)), is_valid(and_(true_, b)));
+
+  auto ts = field_ref("ts_ns");
+  using namespace arrow_literals;
+  ExpectCanonicalizesTo(add(ts, literal(5min)), add(literal(5min), ts));
+  ExpectCanonicalizesTo(add(add(ts, literal(5min)), add(literal(5min), literal(5min))),
+                        add(add(add(literal(5min), literal(5min)), literal(5min)), ts));
```
westonpace, line 1130: Wouldn't `add(literal(5min), literal(5min))` simplify to `add(literal(10min)` since all the args are literals?

cpp/src/arrow/compute/exec/expression_test.cc
```diff
@@ -69,6 +70,39 @@ Expression true_unless_null(Expression argument) {
...
+  TimeUnit::type unit;
+  if constexpr (std::is_same_v<Period, std::nano>) {
+    unit = TimeUnit::NANO;
+  } else if constexpr (std::is_same_v<Period, std::micro>) {
+    unit = TimeUnit::MICRO;
+  } else if constexpr (std::is_same_v<Period, std::milli>) {
+    unit = TimeUnit::MILLI;
+  } else {
+    unit = TimeUnit::SECOND;
+    int_value = std::chrono::duration_cast<std::chrono::seconds>(d).count();
```
westonpace, line 101: So we grab the largest possible resolution and then rely on casting later?  Seems reasonable but just asking for clarification.

westonpace (approved):
Some nits though I expect most are simply misunderstanding

westonpace:
Ah, looks like some new lint fun if we want to use a literals namespace.

cpp/src/arrow/compute/exec/expression_test.cc
bkietz: This is only checking the canonicalization pass, which (among other things) flattens and refolds chains with literals first. This sets up an easy situation for constant folding to do what you describe and eagerly evaluate constant calls. The distinct passes are an implementation detail of the main entry point `SimplifyWithGuarantee`, which might need to run cheap passes like constant folding multiple times if more expensive passes like guarantee substitution happen to yield a constant call.

cpp/src/arrow/compute/exec/expression_test.cc
bkietz: I was trying to keep the scope of the literals contained. As it is, not even this was contained enough for lint https://github.com/apache/arrow/actions/runs/3481185552/jobs/5821942926#step:5:903

cpp/src/arrow/compute/exec/expression_test.cc
bkietz: The logical public place to include it would be as a constructor for `Datum`, but that'd incur `#include <chrono>` for everything which uses Datum. Maybe that's an acceptable cost, WDYT @pitrou ?

cpp/src/arrow/compute/exec/expression.cc
bkietz: Well, that was my intent and the doc for each of these passes states that the argument should be bound... but I guess we're not specifically checking. I'll add a check to Canonicalize and FoldConstants

cpp/src/arrow/compute/exec/expression_test.cc
bkietz: I'll add a comment. If the argument is `std::chrono::nanoseconds`, then use TimeUnit::NANO. If the argument is `std::chrono::microseconds`, ... If the argument is none of `std::chrono::{nanoseconds, microseconds, milliseconds}` then use seconds

cpp/src/arrow/compute/exec/expression.cc
bkietz: It is the same as std::accumulate. I'll just remove these helpers; they don't add much

bkietz:
Well, that's hilarious fun. The linter will not allow `using namespace`, while the compiler misidentifies [using individual operators as a declaration](https://github.com/apache/arrow/actions/runs/3482249863/jobs/5824360020#step:11:1285)

cpp/src/arrow/compute/exec/expression_test.cc
pitrou: How large is `<chrono>` typically? Perhaps we can add a lower-level helper to `arrow/util/time.h`:
```c++
template <typename Rep, typename Period>
int64_t DurationToUnit(std::chrono::duration<Rep, Period> d, TimeUnit::type unit) {
  ...
}
```

cpp/src/arrow/datum.h
```diff
@@ -116,6 +117,27 @@ struct ARROW_EXPORT Datum {
   explicit Datum(std::string value);
   explicit Datum(const char* value);
 
+  // Convenience constructors for a DurationScalar from std::chrono::nanoseconds
```
pitrou, line 120: Why put this in Datum instead of Scalar or even DurationScalar?

cpp/src/arrow/util/literal_operators.h
```diff
@@ -0,0 +1,56 @@
...
+namespace arrow {
+namespace util {
+namespace arrow_literals {
+
+inline TimestampScalar TimestampScalarFromString(const char* s, size_t length,
+                                                 TimeUnit::type unit) {
+  TimestampScalar out{timestamp(unit)};
+  out.is_valid = ::arrow::internal::ParseValue(
+      ::arrow::internal::checked_cast<const TimestampType&>(*out.type), s, length,
+      &out.value);
```
pitrou, line 34: Unfortunately, it sounds like a bad idea to silently instantiate a null scalar for invalid timestamp literals.
If C++ had a macro language we could raise a compile-time error instead...

bkietz: It's possible to emit the compile time error... in C++20 :laughing:

cpp/src/arrow/datum.h
bkietz: Datum already has many convenience constructors, so it seemed like a logical place. Moreover it enabled my immediate use case of `literal(5min)`. I can also add it to DurationScalar if you like

cpp/src/arrow/util/literal_operators.h
bkietz: The operator could return `Result<TimestampScalar>` or `optional<TimestampScalar>` instead

westonpace: I know I encouraged it but maybe let's drop the timestamp literals in favor of simple helper functions:

```
ARROW_ASSIGN_OR_RAISE(Datum christmas, ParseTimestampISO8601("2022-12-25T00:00:00+00:00"));
```

is probably compact enough.  Then we can keep the duration literals since it looks like you're using the standard `chrono` ones and they should be pretty familiar to people (and we don't have to worry about invalid cases).

bkietz: okay, will revert them shortly

cpp/src/arrow/datum.h
```diff
@@ -86,21 +86,26 @@ struct ARROW_EXPORT Datum {
...
-            bool IsArray = std::is_base_of<Array, T>::value,
-            bool IsScalar = std::is_base_of<Scalar, T>::value,
+  template <typename T, typename TV = typename std::remove_reference_t<T>,
+            bool IsArray = std::is_base_of_v<Array, T>,
+            bool IsScalar = std::is_base_of_v<Scalar, T>,
             typename = enable_if_t<IsArray || IsScalar>>
   Datum(T&& value)  // NOLINT implicit conversion
       : Datum(std::make_shared<TV>(std::forward<T>(value))) {}
 
+  // Many Scalars are copyable, let that happen
```
westonpace, line 104: Aren't all datums just wrappers around shared pointers (and thus, copyable)?

westonpace (approved):

westonpace:
Approved caveat minor lint issue.  Thanks for doing this.

cpp/src/arrow/datum.h
pitrou: I think the comment means that the underlying C++ value type (e.g. ChunkedArray) is copyable.

pitrou (approved):
The Datum/Scalar API changes look fine to me.

kou:
@github-actions crossbow submit test-fedora-r-clang-sanitizer test-ubuntu-18.04-r-sanitizer

kou (approved):
+1

PR #14662, opened 2022-11-17
GH-31905: [DevTools] Add linting to Cython files

### LINKED ISSUE #31905: [Python] Add linting to our cython files
The new versions for flake8 do not support linting cython files. Based on [this github issue ](https://github.com/PyCQA/flake8/issues/1482)it was never supposed to be supported.

We updated pre-commit and removed configuration for flake8 linting on cython as part of this ticket: https://issues.apache.org/jira/browse/ARROW-16531

We should find a way of linting our cython files.

**Reporter**: [Raúl Cumplido](https://issues.apache.org/jira/browse/ARROW-16544) / @raulcd
**Assignee**: [Marco Gorelli](https://issues.apache.org/jira/browse/ARROW-16544) / @MarcoGorelli
#### PRs and other links:
- [GitHub Pull Request #14662](https://github.com/apache/arrow/pull/14662)

<sub>**Note**: *This issue was originally created as [ARROW-16544](https://issues.apache.org/jira/browse/ARROW-16544). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

MarcoGorelli:
* Closes: GH-31905


MarcoGorelli:
> ⚠️ Ticket **has not been started in JIRA**, please click 'Start Progress'.

done

kou:
@github-actions crossbow submit -g python

jorisvandenbossche:

MarcoGorelli:
Hi - is there still interest in addressing https://issues.apache.org/jira/browse/ARROW-16544 ?

If not, I'll close, no worries

raulcd:
@github-actions crossbow submit -g python

raulcd:
The workflow that executes linting on PRs is not using pre-commit but `archery lint`:
https://github.com/apache/arrow/blob/master/.github/workflows/dev.yml#L56
https://github.com/apache/arrow/blob/master/docker-compose.yml#L1670
We probably want to add (or follow up) a PR adding a new archery linter for cython to ensure this gets executed on the GitHub workflow on PRs: 
https://github.com/apache/arrow/blob/master/dev/archery/archery/cli.py#L252-L265
I can see `cython-lint` being used on pandas. I think this would be useful. @jorisvandenbossche what are your thoughts on adding this?

jorisvandenbossche:

jorisvandenbossche:
@MarcoGorelli sorry for the slow follow-up here. I think this would be valuable to add yes (and otherwise we should certainly merge the clean-ups you found with this). 

Agreed with @raulcd that ideally this would be exposed through archery. The way this is done for flake8 is we wrap the command line:

https://github.com/apache/arrow/blob/13335454729bcc3e2a3062f5ba7a8da7fc39751f/dev/archery/archery/lang/python.py#L34-L36

and then run that here:

https://github.com/apache/arrow/blob/13335454729bcc3e2a3062f5ba7a8da7fc39751f/dev/archery/archery/utils/lint.py#L219-L235

Given that cython-lint also has a command line, I suppose that should be straightforward to add (but that can also be done as a follow-up).

For the actual diff here, I only have some concern that the changes in the docstring (because of line length) might be messing up how it shows, so checking that with the doc build.

MarcoGorelli:
thanks for your review - sure, I'll take a look at archery

MarcoGorelli:
Sure, have added to Archery - from https://github.com/apache/arrow/actions/runs/4211619690/jobs/7310085926 it looks like it ran

.pre-commit-config.yaml
```diff
@@ -47,3 +47,8 @@ repos:
           - python
         exclude: vendored
         args: [--config, python/setup.cfg]
+  - repo: https://github.com/MarcoGorelli/cython-lint
+    rev: v0.12.4
+    hooks:
+      - id: cython-lint
+        args: [--max-line-length=120]
```
assignUser, line 54: nit: flake8 uses 88 as line length, should this match or are the reasons to increase it for cython files?

assignUser (approved):
Thanks for the PR I remember a discussion on Zulip that there was no adequate cython linter. Good to see that there is one now :D  

+1

.pre-commit-config.yaml
MarcoGorelli: Thanks for your review

It was just to keep the diff down until there was some agreement on whether you wanted this

If you're on board with it, I can go ahead and reduce it to 88

MarcoGorelli: Hi @assignUser - there are some doctests where the result is longer than 88 lines, e.g.

https://github.com/apache/arrow/blob/1ab1f6f159ebf4df1216f28288207a357bbb4017/python/pyarrow/table.pxi#L3843-L3849

` # doctest: +NORMALIZE_WHITESPACE` won't work here (as far as I can tell) because it's a long string without whitespace,
so inserting a newline will make doctest fail

Is it OK to `doctest: +SKIP` this one?

assignUser: Sounds good to me but I am not a python maintainer so I will defer to @jorisvandenbossche :)

AlenkaF: I would suggest making the docstring example data smaller, like so:

```python
>>> import pyarrow as pa
>>> n_legs = pa.array([2, 4, 100])
>>> names = ["n_legs", "animals"]
>>> batch = pa.record_batch([n_legs, animals], names=names)
>>> batch.to_pandas()
Construct a Table   n_legs    animals
0       2   Flamingo
1       4      Horse
2     100  Centipede

Construct a Table from a RecordBatch:

>>> pa.Table.from_batches([batch])
pyarrow.Table
n_legs: int64
animals: string
----
n_legs: [[2,4,100]]
animals: [["Flamingo","Horse","Centipede"]]

Construct a Table from a sequence of RecordBatches:

>>> pa.Table.from_batches([batch, batch])
pyarrow.Table
n_legs: int64
animals: string
----
n_legs: [[2,4,100],[2,4,100]]
animals: [["Flamingo","Horse","Centipede"],["Flamingo","Horse","Centipede"]]
```
I think that should get under the limit of 88.

MarcoGorelli: thanks for your reviews

for now I've put the `--no-pycodestyle` flag, which only does the linting checks (e.g. unused variables/imports/dangerous patterns, rather than style nitpicks) - is that OK?

There's hundreds of cases which would need fixing to get under the 88 limit, which at the moment I'm a bit too busy for I'm afraid

Else I'll update with the 88 limit when I get a chance, no worries

assignUser:
I'd be happy to merge this but I'd like an approval from someone that actually works on the cython files :) @jorisvandenbossche ?

.pre-commit-config.yaml
AlenkaF: I think that is fine 👍 
I would suggest to open an issue to update the 88 limit as a follow up so that it can be tracked and we do not forget to do that at some point.

AlenkaF (approved):
I think this is great, thank you @MarcoGorelli for all your work and perseverance!

PR #14679, opened 2022-11-18
ARROW-15470: [R] Set null value in CSV writer

wjones127:


r/R/csv.R
```diff
@@ -418,6 +418,8 @@ CsvTableReader$create <- function(file,
 #' The `CsvWriteOptions$create()` factory method takes the following arguments:
 #' - `include_header` Whether to write an initial header line with column names
 #' - `batch_size` Maximum number of rows processed at a time. Default is 1024.
+#' - `null_string` The string to be written for null values. Must not contain
+#'   quotation marks. Default is `NA`.
```
nealrichardson, line 422: Is this the current default, or is this a behavior change?

r/R/csv.R
```diff
@@ -456,24 +458,31 @@ CsvReadOptions$create <- function(use_threads = option_use_threads(),
...
 
 readr_to_csv_write_options <- function(include_header,
-                                       batch_size = 1024L) {
+                                       batch_size = 1024L,
+                                       na = "NA") {
   assert_that(is_integerish(batch_size, n = 1, finite = TRUE), batch_size > 0)
   assert_that(is.logical(include_header))
+  assert_that(is.character(na))
+  assert_that(length(na) == 1)
+  assert_that(!grepl('"', na), msg = "na argument must not contain quote characters.")
```
nealrichardson, line 467: Only `"` is invalid? What about `'`?

r/tests/testthat/test-csv.R
```diff
@@ -422,6 +422,25 @@ test_that("Write a CSV file with invalid batch size", {
   )
 })
 
+test_that("Write a CSV with custom NA value", {
+  tbl_out1 <- write_csv_arrow(tbl_no_dates, csv_file, na = "NULL_VALUE")
```
nealrichardson, line 426: Is it worth adding a test that `na = ""` works? I'd imagine that would be a common choice,.

r/R/csv.R
wjones127: Just `"` 😇 

https://github.com/apache/arrow/blob/b9dd41607cb7dd7afd50e3ceb99c68e79e7733a0/cpp/src/arrow/csv/writer.cc#L104

r/R/csv.R
wjones127: For both `write_csv_arrow` and `write_dataset` it was empty string by default. I was originally thinking of aligning with readr default of "NA", at least in `write_csv_arrow`. But now I'm not so sure. @thisisnic what do you think?

thisisnic: Hmm, I'm not sure. It'd be nice to have at least `write_csv_arrow()` behaviour aligned with `readr`, but this would be a change in behaviour that users might not want if they're used to the current behaviour.  What do you think @nealrichardson?

wjones127: Perhaps our goal should be to support the `readr` options, but not necessarily to match the `readr` defaults. Does that seem like a reasonable policy? Right now the function takes `readr` arguments as a convenience, and also has Arrow-specific parameter that may overlap. It seems like if we wanted an API that matched the defaults, we should have a separate API that exclusively took `readr` options.

thisisnic: I think that's a reasonable compromise, I don't think we should be changing default values in this PR just to match readr, but would be good to support the parameter, so that works.

thisisnic:
I tried running this locally, but I don't think both `na` and `null_strings` have been hooked up to the same thing, as when I supply the `na` argument instead of `null_strings` I get:
```
Error in `check_additional_args()` at r/R/dataset-format.R:442:6:
! `na` is not a valid argument for your chosen `format`.
ℹ Supported arguments: `include_header`, `batch_size`, and `null_string`.
```

wjones127:
> I tried running this locally, but I don't think both `na` and `null_strings` have been hooked up to the same thing, as when I supply the `na` argument instead of `null_strings` I get:
> 
> ```
> Error in `check_additional_args()` at r/R/dataset-format.R:442:6:
> ! `na` is not a valid argument for your chosen `format`.
> ℹ Supported arguments: `include_header`, `batch_size`, and `null_string`.
> ```

Ah yes. As written, this enables `na` in `write_csv_arrow()` and `null_string` in `write_dataset()`, but not the other way around. I can update so each function supports both.

r/R/csv.R
```diff
@@ -722,9 +731,10 @@ write_csv_arrow <- function(x,
 
   if (is.null(write_options)) {
     write_options <- readr_to_csv_write_options(
-      include_header = include_header,
-      batch_size = batch_size
+      col_names = include_header,
+      na = na
     )
+    # TODO: what to do with batch_size??
```
wjones127, line 737: I think I mostly have it, but the model I'm going for requires the `readr_to_csv_*_options` parameter set to be disjoint from the `Csv*Options$create` parameter set. The fact that we have `batch_size` here (when it's already exposed via `write_options`) throws a wrench in that plan 😞 

If there were a `readr` equivalent I would map it do that, but I don't believe there is.

thisisnic: I guess part of the issue is that for the `Csv*Options` objects, we have `create()` methods which set up all the options initially, but no methods for updating the different options later.  

What do you think about maybe adding methods for updating these objects?

wjones127: I could do that. This PR might be a little big though :)

r/tests/testthat/test-dataset-csv.R
```diff
@@ -247,7 +247,7 @@ test_that("readr parse options", {
       format = "text",
       quo = "\"",
     ),
-    "Ambiguous"
```
wjones127, line 250: We previously checked for ambiguous options, but why? Is it not sufficient to just say these short-hand ones are unrecognized? cc @nealrichardson @ianmcook

r/R/dataset-format.R
```diff
@@ -482,3 +465,49 @@ FileWriteOptions$create <- function(format, ...) {
...
+      ": ",
+      oxford_paste(unrec_opts)
+    ),
+    call. = FALSE
+    )
+  }
+
+  if (length(arrow_opts) > 0 && length(readr_opts) > 0) {
+    abort(c(
+      "Additional writing options must be Arrow-style or readr-style, but not both.",
```
wjones127, line 492:
```suggestion
      "Additional options must be Arrow-style or readr-style, but not both.",
```

r/tests/testthat/test-dataset-csv.R
thisisnic: Though it's bad practice, it's normal in R to be able to refer to function parameters using a shortened version, so this is a special case where we don't allow this if it's unclear whether the shortened version is referring to an Arrow or a readr option.

r/tests/testthat/test-type.R
```diff
@@ -94,7 +94,7 @@ test_that("infer_type() can infer blob type", {
 
   expect_equal(infer_type(blob::blob()), binary())
 
-  big_ish_raw <- raw(2 ^ 20)
+  big_ish_raw <- raw(2^20)
```
wjones127, line 97: Apologies, I ran `make style-all` and so we have some other style fixes around. Should I revert these and add them in a separate PR?

r/tests/testthat/test-dataset-csv.R
wjones127: 😱  I'll add some generic handling for that in the dispatch function then.

r/tests/testthat/test-type.R
thisisnic: It doesn't look like there are that many, all good.

r/R/csv.R
```diff
@@ -455,25 +457,32 @@ CsvReadOptions$create <- function(use_threads = option_use_threads(),
...
-readr_to_csv_write_options <- function(include_header,
-                                       batch_size = 1024L) {
+readr_to_csv_write_options <- function(include_header = TRUE,
+                                       batch_size = 1024L,
+                                       na = "") {
   assert_that(is_integerish(batch_size, n = 1, finite = TRUE), batch_size > 0)
   assert_that(is.logical(include_header))
+  assert_that(is.character(na))
+  assert_that(length(na) == 1)
+  assert_that(!grepl('"', na), msg = "na argument must not contain quote characters.")
```
nealrichardson, line 467: Should this assertion move down into `CsvWriteOptions$create`?

nealrichardson (commented):
One more edge case to check for but otherwise LGTM

r/R/csv.R
```diff
@@ -455,25 +457,32 @@ CsvReadOptions$create <- function(use_threads = option_use_threads(),
...
+    batch_size = as.integer(batch_size),
+    null_string = na
   )
 }
 
 #' @rdname CsvReadOptions
 #' @export
 CsvWriteOptions <- R6Class("CsvWriteOptions", inherit = ArrowObject)
-CsvWriteOptions$create <- function(include_header = TRUE, batch_size = 1024L) {
+CsvWriteOptions$create <- function(include_header = TRUE, batch_size = 1024L, null_string = "") {
```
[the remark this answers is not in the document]
wjones127: It should now fail the assertion.

r/R/csv.R
wjones127: Good point. I don't even know why we have the assertions here if it just calls into `CsvWriteOptions$create`

thisisnic (approved):

PR #14681, opened 2022-11-20
ARROW-18367: [C++] Enable the creation of named table relations

ljishen:
- changes also include fixes from running IWYU over the directory cpp/src/arrow/engine/substrait

### DIFF
--- a/cpp/src/arrow/compute/exec/options.h
+++ b/cpp/src/arrow/compute/exec/options.h
@@ -85,6 +85,21 @@ class ARROW_EXPORT TableSourceNodeOptions : public ExecNodeOptions {
   int64_t max_batch_size;
 };
 
+/// \brief Define a lazy resolved Arrow table.
+///
+/// The table uniquely identified by the names can typically be resolved at the time when
+/// the plan is to be consumed.
+///
+/// This node is for serialization purposes only and can never be executed.
+class ARROW_EXPORT NamedTableNodeOptions : public ExecNodeOptions {
+ public:
+  NamedTableNodeOptions(std::vector<std::string> names, std::shared_ptr<Schema> schema)
+      : names(std::move(names)), schema(schema) {}
+
+  std::vector<std::string> names;
+  std::shared_ptr<Schema> schema;
+};
+
 /// \brief An extended Source node which accepts a schema
 ///
 /// ItMaker is a maker of an iterator of tabular data.
--- a/cpp/src/arrow/engine/substrait/expression_internal.cc
+++ b/cpp/src/arrow/engine/substrait/expression_internal.cc
@@ -19,16 +19,43 @@
 
 #include "arrow/engine/substrait/expression_internal.h"
 
+#include <algorithm>
+#include <array>
+#include <cstdint>
+#include <cstring>
+#include <functional>
 #include <memory>
+#include <optional>
+#include <string>
+#include <string_view>
+#include <type_traits>
 #include <utility>
+#include <vector>
 
+#include <google/protobuf/descriptor.h>
+
+#include "arrow/array/array_base.h"
+#include "arrow/array/array_nested.h"
+#include "arrow/array/array_primitive.h"
+#include "arrow/array/util.h"
+#include "arrow/buffer.h"
 #include "arrow/builder.h"
+#include "arrow/compute/api_scalar.h"
 #include "arrow/compute/exec/expression.h"
 #include "arrow/compute/exec/expression_internal.h"
+#include "arrow/engine/substrait/extension_set.h"
 #include "arrow/engine/substrait/extension_types.h"
+#include "arrow/engine/substrait/options.h"
 #include "arrow/engine/substrait/type_internal.h"
 #include "arrow/result.h"
+#include "arrow/scalar.h"
 #include "arrow/status.h"
+#include "arrow/type.h"
+#include "arrow/util/checked_cast.h"
+#include "arrow/util/decimal.h"
+#include "arrow/util/endian.h"
+#include "arrow/util/logging.h"
+#include "arrow/util/small_vector.h"
 #include "arrow/util/string.h"
 #include "arrow/visit_scalar_inline.h"
 
--- a/cpp/src/arrow/engine/substrait/relation_internal.cc
+++ b/cpp/src/arrow/engine/substrait/relation_internal.cc
@@ -17,23 +17,45 @@
 
 #include "arrow/engine/substrait/relation_internal.h"
 
-#include "arrow/compute/api_scalar.h"
+#include <cstdint>
+#include <functional>
+#include <iterator>
+#include <memory>
+#include <string>
+#include <string_view>
+#include <type_traits>
+#include <utility>
+#include <variant>
+#include <vector>
+
+#include "arrow/compute/api_aggregate.h"
+#include "arrow/compute/exec/exec_plan.h"
+#include "arrow/compute/exec/expression.h"
 #include "arrow/compute/exec/options.h"
+#include "arrow/compute/kernel.h"
+#include "arrow/dataset/dataset.h"
+#include "arrow/dataset/discovery.h"
 #include "arrow/dataset/file_base.h"
 #include "arrow/dataset/file_ipc.h"
 #include "arrow/dataset/file_parquet.h"
 #include "arrow/dataset/plan.h"
 #include "arrow/dataset/scanner.h"
+#include "arrow/datum.h"
 #include "arrow/engine/substrait/expression_internal.h"
+#include "arrow/engine/substrait/extension_set.h"
+#include "arrow/engine/substrait/options.h"
+#include "arrow/engine/substrait/relation.h"
 #include "arrow/engine/substrait/type_internal.h"
+#include "arrow/filesystem/filesystem.h"
 #include "arrow/filesystem/localfs.h"
-#include "arrow/filesystem/path_util.h"
+#include "arrow/filesystem/type_fwd.h"
 #include "arrow/filesystem/util_internal.h"
+#include "arrow/io/type_fwd.h"
+#include "arrow/status.h"
+#include "arrow/type.h"
 #include "arrow/util/checked_cast.h"
 #include "arrow/util/string.h"
 #include "arrow/util/uri.h"
-
-#include <memory>
 
 namespace arrow {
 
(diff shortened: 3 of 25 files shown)


cpp/src/arrow/engine/substrait/relation_internal.cc
```diff
@@ -689,7 +695,68 @@ Result<std::unique_ptr<substrait::ReadRel>> ScanRelationConverter(
...
+  if (file_system_dataset != nullptr) {
+    ARROW_RETURN_NOT_OK(SetReadRelationLocalFiles(file_system_dataset, read_rel));
+    return std::move(read_rel);
+  }
+
+  auto* in_memory_dataset = dynamic_cast<dataset::InMemoryDataset*>(dataset);
+  if (in_memory_dataset != nullptr) {
+    ARROW_ASSIGN_OR_RAISE(auto scanner_builder, in_memory_dataset->NewScan());
+    ARROW_ASSIGN_OR_RAISE(auto scanner, scanner_builder->Finish());
+    ARROW_ASSIGN_OR_RAISE(auto num_rows, scanner->CountRows());
```
westonpace, line 748: We could maybe just add a `NumRows` method to `InMemoryDataset` to avoid having to create and run a scan which might save a bit of complexity.

cpp/src/arrow/engine/substrait/serde.h
```diff
@@ -25,19 +25,34 @@
...
+}  // namespace compute
+
+namespace dataset {
+class WriteNodeOptions;
+}
+
 namespace engine {
 
+class ExtensionIdRegistry;
+class ExtensionSet;
```
westonpace, line 54: I'm not sure this is better than using the `type_fwd.h` files.

cpp/src/arrow/engine/substrait/type_internal.cc
```diff
@@ -471,8 +482,14 @@ void ToProtoGetDepthFirstNames(const FieldVector& fields,
...
-  if (schema.metadata()) {
-    return Status::Invalid("::substrait::NamedStruct does not support schema metadata");
+  const auto& metadata = schema.metadata();
+  if (metadata) {
+    std::unordered_map<std::string, std::string> metadata_map{};
+    metadata->ToUnorderedMap(&metadata_map);
+    metadata_map.erase(Table::kTableNameKey);
+    if (!metadata_map.empty()) {
+      return Status::Invalid("::substrait::NamedStruct does not support schema metadata");
+    }
```
westonpace, line 492: Maybe only check this if the conversion strictness is `EXACT_ROUNDTRIP`?  Otherwise it seems ok to drop/ignore the metadata.

cpp/src/arrow/table.h
```diff
@@ -41,6 +41,8 @@ class MemoryPool;
 /// \brief Logical table as sequence of chunked arrays
 class ARROW_EXPORT Table {
  public:
+  static constexpr char const kTableNameKey[] = "tableName";
+
```
westonpace, line 45: Given that this key is specific to Substrait use I'm not sure this is the right place for this.  In fact the "table" here is not an Arrow table but more of a database table (which is more similar to an arrow dataset).  Can you move this constant into the substrait package?

cpp/src/arrow/engine/substrait/relation_internal.cc
```diff
@@ -689,7 +695,68 @@ Result<std::unique_ptr<substrait::ReadRel>> ScanRelationConverter(
     read_rel_lfs->mutable_items()->AddAllocated(read_rel_lfs_ffs.release());
   }
   read_rel->set_allocated_local_files(read_rel_lfs.release());
-  return std::move(read_rel);
+
+  return Status::OK();
+}
+
+Status SetReadRelationNamedTable(const dataset::InMemoryDataset* dataset,
```
westonpace, line 702:
```suggestion
Status SetReadRelationNamedTable(const dataset::InMemoryDataset& dataset,
```
The style guide prefers const reference over const pointer:

> Non-optional input parameters should usually be values or const references

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -2221,6 +2252,67 @@ TEST(Substrait, BasicPlanRoundTrippingEndToEnd) {
...
+  auto filter = compute::equal(compute::field_ref("A"), compute::field_ref("B"));
+
+  auto declarations = compute::Declaration::Sequence(
+      {compute::Declaration(
+           {"scan", dataset::ScanNodeOptions{dataset, scan_options}, "s"}),
+       compute::Declaration({"filter", compute::FilterNodeOptions{filter}, "f"})});
+
+  std::shared_ptr<ExtensionIdRegistry> sp_ext_id_reg = MakeExtensionIdRegistry();
+  ExtensionIdRegistry* ext_id_reg = sp_ext_id_reg.get();
+  ExtensionSet ext_set(ext_id_reg);
```
westonpace, line 2279: You probably don't need to create an `ExtensionIdRegistry` (it will just use the default if you do not supply one).

westonpace (changes requested):
I have a few requests.  This seems like a fairly clever way to produce named table relations.

The Substrait serializer works on declarations.  There is no need for a declaration to have a node factory (or even have a corresponding exec node).  Would it be simpler to do something like:

```
/// Declare a named table, there is currently no way to consume this node
class NamedTableOptions {
  public:
    std::vector<std::string> names;
    std::shared_ptr<Schema> table_schema;
};

...

Declaration named_tbl{"named_table", NamedTableOptions{{"table", "1"}, schema}};

...

Result<compute::ExecNode*> MakeNamedTableNode(compute::ExecPlan* plan,
                                               std::vector<compute::ExecNode*> inputs,
                                               const compute::ExecNodeOptions& options) {
  return Status::Invalid("The named table node is for serialization purposes only and can never be converted into an exec plan or executed");
}
...

DCHECK_OK(registry->AddFactory("named_table", MakeNamedTableNode));
```

ljishen:
@westonpace Thanks for the great suggestions! I'll fix them and update the PR.

cpp/src/arrow/engine/substrait/expression_internal.cc
```diff
@@ -19,16 +19,44 @@
...
+#include <cstring>
+#include <functional>
 #include <memory>
+#include <optional>
+#include <string>
+#include <string_view>
+#include <type_traits>
 #include <utility>
-
-#include "arrow/builder.h"
```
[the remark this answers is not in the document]
ljishen: I used the `build-support/iwyu/iwyu.sh` to generate these includes. I've updated the code to use `type_fwd.h` files as much as possible.

ljishen:
> I have a few requests. This seems like a fairly clever way to produce named table relations.
> 
> The Substrait serializer works on declarations. There is no need for a declaration to have a node factory (or even have a corresponding exec node). Would it be simpler to do something like:
> 
> ```
> /// Declare a named table, there is currently no way to consume this node
> class NamedTableOptions {
>   public:
>     std::vector<std::string> names;
>     std::shared_ptr<Schema> table_schema;
> };
> 
> ...
> 
> Declaration named_tbl{"named_table", NamedTableOptions{{"table", "1"}, schema}};
> 
> ...
> 
> Result<compute::ExecNode*> MakeNamedTableNode(compute::ExecPlan* plan,
>                                                std::vector<compute::ExecNode*> inputs,
>                                                const compute::ExecNodeOptions& options) {
>   return Status::Invalid("The named table node is for serialization purposes only and can never be converted into an exec plan or executed");
> }
> ...
> 
> DCHECK_OK(registry->AddFactory("named_table", MakeNamedTableNode));
> ```

Yes, I believe this is a better way to solve this.

lidavidm:
I'm not caught up on the Substrait work. I believe Weston has just been busy lately and should be able to catch up. @westonpace, let me know if you want me to step in here instead.

ljishen:
> I'm not caught up on the Substrait work. I believe Weston has just been busy lately and should be able to catch up. @westonpace, let me know if you want me to step in here instead.

Thanks for your response.

cpp/src/arrow/compute/exec/options.h
```diff
@@ -85,6 +85,19 @@ class ARROW_EXPORT TableSourceNodeOptions : public ExecNodeOptions {
   int64_t max_batch_size;
 };
 
+/// \brief Define a lazy resolved Arrow table.
+///
+/// The table uniquely identified by the names can typically be resolved at the time when
+/// the plan is to be consumed.
```
westonpace, line 91:
```suggestion
/// the plan is to be consumed.
///
/// This node is for serialization purposes only and can never be executed
```

cpp/src/arrow/engine/substrait/relation_internal.cc
```diff
@@ -638,6 +659,10 @@ Result<std::shared_ptr<Schema>> ExtractSchemaToBind(const compute::Declaration&
   } else if (declr.factory_name == "filter") {
     auto input_declr = std::get<compute::Declaration>(declr.inputs[0]);
     ARROW_ASSIGN_OR_RAISE(bind_schema, ExtractSchemaToBind(input_declr));
+  } else if (declr.factory_name == "named_table") {
+    const auto& opts =
+        checked_cast<const compute::NamedTableNodeOptions&>(*(declr.options));
```
westonpace, line 664:
```suggestion
        checked_cast<const compute::NamedTableNodeOptions&>(*declr.options);
```

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -2260,6 +2293,57 @@ TEST(SubstraitRoundTrip, BasicPlanEndToEnd) {
...
+
+  NamedTableProvider table_provider =
+      [&input_table, &table_names](
+          const std::vector<std::string>& names) -> Result<compute::Declaration> {
+    if (table_names != names) {
+      return Status::Invalid("Table name mismatch");
+    }
+    std::shared_ptr<compute::ExecNodeOptions> options =
+        std::make_shared<compute::TableSourceNodeOptions>(input_table);
+    return compute::Declaration("table_source", {}, options, "mock_source");
```
westonpace, line 2332:
```suggestion
    return compute::Declaration("table_source", {}, std::move(options), "mock_source");
```

westonpace (changes requested):
David is right, I was on vacation last week.  I apologize for the delay.  This looks good, a few minor nits and then I think we can go.  Always feel free to `@westonpace` as that is the most likely way to get something to show up in my notifications in a way I see it.

westonpace (approved):
@ljishen I ran the CI and your changes exposed an IWYU problem in an unrelated file (acero_test.cc).  I've added a small commit (https://github.com/apache/arrow/pull/14681/commits/4fbe953caadd6af4186ee421e936028750629335) to fix that and will merge when this passes.

ljishen:
@westonpace thanks for the fix!

PR #14527, opened 2022-10-27
ARROW-15641: [C++][Python] UDF Aggregate Function Implementation

vibhatha:
Initial Draft of Aggregate UDFs for Python

- [x] Improve the existing logic with more tests
- [ ] Find out a missing aggregate function and try to apply it with the current API (check if the applied theory sustains)
- [x] Using a custom object for state holding rather than Arrow data structures
- [x] Improvement function docs and comments
- [ ] Adding binary and other arity tests
- [x] Self-review 1
- [ ] Self-review 2

PR #14320, opened 2022-10-05
ARROW-16212: [C++][Python] Register Multiple Kernels for a UDF

vibhatha:
This PR includes the support for multi-kernel registration with Scalar UDFs for Python.

PR #14043, opened 2022-09-05
ARROW-17613: [C++] Add function execution API for a preconfigured kernel

rtpsw:
See https://issues.apache.org/jira/browse/ARROW-17613

PR #14682, opened 2022-11-20
GH-32916: [C++] [Python] User-defined tabular functions

### LINKED ISSUE #32916: [C++] [Python] User-defined tabular functions
Currently, only a stateless user-defined function is supported in PyArrow. This issue will add support for a user-defined tabular function, which is a user-function implemented in Python that returns a stateful stream of tabular data.

**Reporter**: [Yaron Gvili](https://issues.apache.org/jira/browse/ARROW-17676) / @rtpsw
**Assignee**: [Yaron Gvili](https://issues.apache.org/jira/browse/ARROW-17676) / @rtpsw
#### PRs and other links:
- [GitHub Pull Request #14682](https://github.com/apache/arrow/pull/14682)

<sub>**Note**: *This issue was originally created as [ARROW-17676](https://issues.apache.org/jira/browse/ARROW-17676). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

rtpsw:
See https://issues.apache.org/jira/browse/ARROW-17676
* Closes: #32916


rtpsw:
The CI job failures seem unrelated to this PR's changes.

python/pyarrow/_compute.pyx
```diff
@@ -2629,14 +2718,98 @@ def register_scalar_function(func, function_name, function_doc, in_types,
       21
     ]
     """
+    return register_scalar_like_function(GetRegisterScalarFunction(),
+                                         func, function_name, function_doc, in_types,
+                                         out_type, func_registry)
+
+
+def register_tabular_function(func, function_name, function_doc, in_types, out_type,
```
wjones127, line 2726: I'm not sure if "tabular" is the right word. I would have expected that to be a function that takes a table/record batch and returns a new table / record batch.

What about "batch generator function"?

```suggestion
def register_batch_generator_function(func, function_name, function_doc, in_types, out_type,
```

python/pyarrow/tests/test_udf.py
```diff
@@ -504,3 +504,49 @@ def test_input_lifetime(unary_func_fixture):
...
+            if self.n == 0:
+                batch = _record_batch_from_iters(schema, [], [])
+            else:
+                self.n -= 1
+                batch = _record_batch_for_range(schema, self.n)
+            return pc.udf_result_from_record_batch(batch)
+    return Generator()
+
+
+def test_udt():
```
wjones127, line 540: Could we add a test validating the behavior of when an exception is raised in the function?

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -105,21 +167,109 @@ Status RegisterScalarFunction(PyObject* user_function, ScalarUdfWrapperCallback
...
+      options, registry);
+}
+
+Status RegisterTabularFunction(PyObject* user_function, ScalarUdfWrapperCallback wrapper,
+                               const ScalarUdfOptions& options,
+                               compute::FunctionRegistry* registry) {
+  if (options.arity.num_args != 0 || options.arity.is_varargs) {
+    return Status::Invalid("tabular function must have no arguments");
+  }
+  return RegisterScalarLikeFunction(
```
wjones127, line 204: Shouldn't we validate the the `out_type` is a StructType?

python/pyarrow/_compute.pyx
```diff
@@ -2629,14 +2718,98 @@ def register_scalar_function(func, function_name, function_doc, in_types,
...
+        ScalarUdfContext. It must return a callable that
+        returns on each invocation a StructArray matching
+        the out_type, where an empty array indicates end.
+    function_name : str
+        Name of the function. This name must be globally unique.
+    function_doc : dict
+        A dictionary object with keys "summary" (str),
+        and "description" (str).
+    in_types : Dict[str, DataType]
+        Must be an empty dictionary.
```
wjones127, line 2751: Why have this parameter at all?

python/pyarrow/_compute.pyx
```diff
@@ -2629,14 +2718,98 @@ def register_scalar_function(func, function_name, function_doc, in_types,
...
+        Output type of the function.
+    func_registry : FunctionRegistry
+        Optional function registry to use instead of the default global one.
+    """
+    return register_scalar_like_function(GetRegisterTabularFunction(),
+                                         func, function_name, function_doc, in_types,
+                                         out_type, func_registry)
+
+
+def register_scalar_like_function(register_func, func, function_name, function_doc, in_types,
```
wjones127, line 2762: the table functions don't take any `in_types`, so I don't see why it needs to share that logic with scalar like. Perhaps we should simply add a utility function to handle the common doc validation for UDF registration?

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -105,21 +167,109 @@ Status RegisterScalarFunction(PyObject* user_function, ScalarUdfWrapperCallback
...
+    registry = compute::GetFunctionRegistry();
+  }
+  ARROW_ASSIGN_OR_RAISE(auto func, registry->GetFunction(func_name));
+  if (func->kind() != Function::SCALAR) {
+    return Status::Invalid("tabular function of non-scalar kind");
+  }
+  auto arity = func->arity();
+  if (arity.num_args != 0 || arity.is_varargs) {
+    return Status::Invalid("tabular function of non-null arity");
+  }
```
wjones127, line 235: Why doesn't this validation happen during registration?

wjones127 (changes requested):
Thanks, Yaron. This all looks very useful.

I have left several suggestions.

westonpace (commented):
I think tabular UDFs would be a great feature.  However, I think I would rather see them implemented as an exec node (they would be "batch UDFs"), so that they can run on greater than memory data.

For example, a good use case for this I think is the "drop_null" compute kernel (which we actually have today).  Given a record batch, it will drop any rows that have a null value in any column (or one could imagine supplying a list of columns).

To implement this we would need options like:

```
class TableUdfOptions {
  Expression table_udf;
};
```

Fitting a tabular UDF to an `Expression` would be a little tricky though.  I think one way you could do it is to treat the "input schema" for the expression as a single column schema, where that one column is a struct array consisting of each of the columns of the table.

So, to drop nulls on the entire table it would just be `drop_null(!0)` and to drop nulls from a selection of columns you could do something like `drop_null(select(!0), [0, 1, 3])`

It might be worth checking how Ibis represents tabular functions too.

wjones127:
@westonpace I think you may misunderstand what this is adding; tabular function is a misnomer. These are more like UDF record batch generators or UDF source node. They are take no inputs, and emit record batches / struct arrays.

westonpace:
> @westonpace I think you may misunderstand what this is adding; tabular function is a misnomer. These are more like UDF record batch generators or UDF source node. They are take no inputs, and emit record batches / struct arrays.

You are correct, I did misunderstand. This does also seem to be the definition of table functions from the literature.  Thanks for the clarification.

In that case, could this be a source node?

rtpsw:
Sorry, I should have given more context for this PR. @wjones127's [comment](https://github.com/apache/arrow/pull/14682#issuecomment-1322559310) is correct. The concept of user-defined tabular (UDT) functions is [clarified here](https://github.com/apache/arrow/pull/14043#issuecomment-1257178331). In quick-reply to a couple of questions about the naming:
1. Why "UDT"? Logically, a UDT function returns a (user-defined) table. The subdivision of the resulting table into batches does not affect this.
2. Why "tabular"? Physically, the function produces (a sequence of) structures equivalent to a table. The code in this PR only supports flat struct-arrays, and interprets them as a table.

I'm open to a different name than "tabular", in particular because it caused confusion. However, I'm not in favor of names with "batch" in them due to point 1 above. Could you propose a name considering the above 2 points?

Still related to context for this PR: @wjones127, you asked why have an in-type parameter at all. While the current implementation does not handle UDT inputs, I'm designing the UDT API to support inputs - see [this explanation](https://github.com/apache/arrow/pull/14043#issuecomment-1257178331).

python/pyarrow/_compute.pyx
```diff
@@ -2629,14 +2718,98 @@ def register_scalar_function(func, function_name, function_doc, in_types,
...
+        the out_type, where an empty array indicates end.
+    function_name : str
+        Name of the function. This name must be globally unique.
+    function_doc : dict
+        A dictionary object with keys "summary" (str),
+        and "description" (str).
+    in_types : Dict[str, DataType]
+        Must be an empty dictionary.
+    out_type : DataType
+        Output type of the function.
```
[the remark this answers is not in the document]
rtpsw: It should be a Python object that wraps a `arrow::DataType`.  This gets passed to `pyarrow_unwrap_data_type`.

python/pyarrow/_compute.pyx
```diff
@@ -2629,14 +2718,98 @@ def register_scalar_function(func, function_name, function_doc, in_types,
...
+    Register a user-defined scalar-like function.
+
+    A scalar-like function is a callable accepting a first
+    context argument of type ScalarUdfContext as well as
+    possibly additional Arrow arguments, and returning a
+    an Arrow result appropriate for the kind of function.
+    A scalar function and a tabular function are examples
+    for scalar-like functions.
+    This function is normally not called directly but via
+    register_scalar_function or register_tabular_function.
```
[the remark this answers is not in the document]
rtpsw: It is intended for developers or power-users. By passing a `register_func` argument holding a custom `CRegisterScalarLikeFunction`, one may extend with a new kind of UDX. Having said that, I won't object to making it internal if people think it is too early to decide what the extension API should be.

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -17,35 +17,92 @@
...
+#include "arrow/compute/kernel.h"
 #include "arrow/python/common.h"
+#include "arrow/util/checked_cast.h"
 
 namespace arrow {
 
 using compute::ExecResult;
 using compute::ExecSpan;
+using compute::Function;
+using compute::OutputType;
```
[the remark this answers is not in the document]
rtpsw: I'll remove the `using` statements and fix the references.

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -17,35 +17,92 @@
...
+  Result<std::unique_ptr<compute::KernelState>> operator()(
+      compute::KernelContext*, const compute::KernelInitArgs&) {
+    return std::make_unique<PythonScalarUdfKernelState>(function);
+  }
+
+  std::shared_ptr<OwnedRefNoGIL> function;
+};
+
+struct PythonTableUdfKernelInit {
+  PythonTableUdfKernelInit(std::shared_ptr<OwnedRefNoGIL> function_maker,
```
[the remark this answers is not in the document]
rtpsw: See [this post](https://github.com/apache/arrow/pull/14682#issuecomment-1322610029).

python/pyarrow/tests/test_udf.py
```diff
@@ -504,3 +504,49 @@ def test_input_lifetime(unary_func_fixture):
...
+            self.n = 3
+
+        def __call__(self, ctx):
+            if self.n == 0:
+                batch = _record_batch_from_iters(schema, [], [])
+            else:
+                self.n -= 1
+                batch = _record_batch_for_range(schema, self.n)
+            return pc.udf_result_from_record_batch(batch)
+    return Generator()
```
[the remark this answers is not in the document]
rtpsw: A solution with `yield` has crossed my mind, but I haven't tried it. I agree this is something to look into, aiming for a nicer syntax for end-users.

python/pyarrow/_compute.pyx
rtpsw: See [this post](https://github.com/apache/arrow/pull/14682#issuecomment-1322610029).

python/pyarrow/_compute.pyx
rtpsw: The design is that they will take inputs, but only no-inputs is currently implemented. See [this post](https://github.com/apache/arrow/pull/14682#issuecomment-1322610029).

python/pyarrow/src/arrow/python/udf.cc
rtpsw: I'm not sure what you mean; I don't see `out_type` here.

python/pyarrow/tests/test_udf.py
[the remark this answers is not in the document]
rtpsw: I was able to make this work in a quick-and-dirty way. Please see if the following code seems reasonable from both a library and a user point of view. If so, I'll organize it in a cleaner way.

I added this as a library function:
```
def make_udt_func(schema, batch_gen):
    def udf_func(ctx):
        class UDT:
            def __init__(self):
                self.caller = None
            def __call__(self, ctx):
                try:
                    if self.caller is None:
                        self.caller, ctx = batch_gen(ctx).send, None
                    batch = self.caller(ctx)
                except StopIteration:
                    batch = _record_batch_from_iters(schema, *[[] for i in schema])
                return batch.to_struct_array()
        return UDT()
    return udf_func
```
and replaced `func - datasource1` with this user-code:
```
    def batch_gen(ctx):
        for i in range(3, 0, -1):
            ctx = yield _record_batch_for_range(schema, n - 1)
    func = make_udt_func(schema, batch_gen)
```

python/pyarrow/compute.py
```diff
@@ -80,7 +80,10 @@
     list_functions,
     _group_by,
     # Udf
+    get_record_batches_from_tabular_function,
```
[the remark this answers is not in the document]
rtpsw: I like the idea of encapsulating this in a class. However, I suspect these classes are supposed to correspond to `arrow::compute::Function::Kind` values, and adding a new value here would require many changes in Arrow where this enum is used. If so, I'd prefer to defer, or find a more localized alternative.

python/pyarrow/tests/test_udf.py
rtpsw: Done.

python/pyarrow/src/arrow/python/udf.cc
rtpsw: OK, I see what you mean now. I'll add a check.

python/pyarrow/_compute.pyx
rtpsw: I'll fix to `NotImplemented` in this case.

rtpsw:
The CI jobs failures appear to be unrelated. In particular, no Python test, including `test_udf.py`, is counted as "failed".

python/pyarrow/_compute.pyx
```diff
@@ -141,6 +153,38 @@ cdef wrap_hash_aggregate_kernel(const CHashAggregateKernel* c_kernel):
     return kernel
 
 
+cdef class RecordBatchIterator(_Weakrefable):
```
westonpace, line 156: Is there any way we can use a RecordBatchReader instead of introducing a new type which seems pretty similar?

python/pyarrow/_compute.pyx
```diff
@@ -2558,8 +2602,31 @@ def _get_scalar_udf_context(memory_pool, batch_length):
...
+                                                CFunctionRegistry* registry)
+
+cdef class RegisterScalarLikeFunction(_Weakrefable):
+    cdef CRegisterScalarLikeFunction register_func
+
+    cdef void init(self, const CRegisterScalarLikeFunction register_func):
+        self.register_func = register_func
+
+
+cdef GetRegisterScalarFunction():
```
westonpace, line 2616: Nit: Should probably be `get_register_scalar_function`

python/pyarrow/_compute.pyx
```diff
@@ -2558,8 +2602,31 @@ def _get_scalar_udf_context(memory_pool, batch_length):
...
+    cdef RegisterScalarLikeFunction reg = RegisterScalarLikeFunction.__new__(RegisterScalarLikeFunction)
+    reg.register_func = RegisterScalarFunction
+    return reg
+
+
+cdef GetRegisterTabularFunction():
```
westonpace, line 2622: Nit: Should probably be `get_register_tabular_function`

python/pyarrow/_compute.pyx
westonpace: Maybe `Must be an empty dictionary (planned for future use)`?

python/pyarrow/_compute.pyx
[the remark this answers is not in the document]
westonpace: Can it be a schema?  That would be simpler for users I think.  We could do the conversion to a struct type ourselves.

python/pyarrow/_compute.pyx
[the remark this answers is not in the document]
westonpace: They could create a new kind of UDX but is there any point if there is no node that consumes those types of functions?  At the moment I would agree to making this internal (we can keep the API docs, I think it's just a matter of adding a `_` in front of the name)

python/pyarrow/_compute.pyx
```diff
@@ -2678,5 +2831,37 @@ def register_scalar_function(func, function_name, function_doc, in_types,
...
+        c_func_registry = (<FunctionRegistry>func_registry).registry
+
+    c_register_func = (<RegisterScalarLikeFunction>register_func).register_func
+
+    check_status(c_register_func(c_function,
+                                 <function[CallbackUdf]> &_scalar_udf_callback,
+                                 c_options, c_func_registry))
+
+
+def get_record_batches_from_tabular_function(function_name, func_registry=None):
```
westonpace, line 2846: Is this the table UDF equivalent of `call_function`?  In other words, a way to use the function outside of an exec plan?

westonpace (commented):
A few questions inline.

What's the ultimate goal here?  Do you hope to support an exec plan node that acts as a source node and is given a function name?  Or are you planning on purely using this via `get_record_batches_from_tabular_function`?

rtpsw:
> What's the ultimate goal here? Do you hope to support an exec plan node that acts as a source node and is given a function name?

The short answer is yes. A longer answer is that such source nodes have a specific structure - they accept parameters in a standard way and employ a user-defined function for the implementation.

> Or are you planning on purely using this via `get_record_batches_from_tabular_function`?

`get_record_batches_from_tabular_function` is more direct way to access the data produced by the user-defined function.

python/pyarrow/_compute.pyx
rtpsw: I'll look into this.

python/pyarrow/_compute.pyx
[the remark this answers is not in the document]
rtpsw: I'll look into this.

python/pyarrow/_compute.pyx
rtpsw: > Is this the table UDF equivalent of `call_function`?

I didn't think about it this way, but I guess so. Perhaps this should be integrated into `call_function` at some point. In any case, your comment leads me to a couple of changes I'd make here:
- Rename to `call_tabular_function`.
- Add to the signature an `args` parameter defaulting to `None`, and document that for now it is reserved and should be empty. In the future, it will be used for passing arguments to the function call.

> In other words, a way to use the function outside of an exec plan?

Yes, that's how I was thinking about it.

python/pyarrow/_compute.pyx
rtpsw: I included this as a separate commit, so you could review whether this is what you had in mind.

python/pyarrow/_compute.pyx
rtpsw: I included the changes just described.

python/pyarrow/_compute.pyx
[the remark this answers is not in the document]
rtpsw: I made it internal.

> They could create a new kind of UDX but is there any point if there is no node that consumes those types of functions?

Note that a new kind of UDX could introduce a different way of configuration (e.g., using a callable returning a SQL script) while still outputting in a standard way (e.g., tabular data), so existing nodes could consume its output.

python/pyarrow/_compute.pyx
rtpsw: Fixed.

rtpsw:
The CI job failures seem unrelated to this PR's changes.

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -83,16 +135,17 @@ struct PythonUdf : public compute::KernelState {
...
+                     compute::ExecResult* out) {
   auto udf = static_cast<PythonUdf*>(ctx->kernel()->data.get());
   return SafeCallIntoPython([&]() -> Status { return udf->Exec(ctx, batch, out); });
 }
 
-}  // namespace
-
-Status RegisterScalarFunction(PyObject* user_function, ScalarUdfWrapperCallback wrapper,
-                              const ScalarUdfOptions& options) {
+Status RegisterScalarLikeFunction(PyObject* user_function,
```
westonpace, line 144: I think this PR muddles the meaning of the word "scalar" which is already a confusing enough word as it is.  Do we want to maybe pick a new word entirely instead of "scalar like"?  We should also rename ScalarUdfWrapperCallback, ScalarUdfOptions as well (at the very least to ScalarLike).  What would you imagine is not a "scalar like" function?  The pattern (something with a python function object and options that describe the behavior of that function object) seems pretty generic.

If we can't think of anything that wouldn't fit this pattern then maybe "PythonFunction", "PythonUdfWrapperCallback", and "PythonUdfOptions"?

Or maybe, compared to aggregate udfs, the unique thing is that there is one function.  So maybe "SingleFunction", "SingleFunctionUdfWrapperCallback", and "SingleFunctionUdfOptions"?

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -105,21 +158,117 @@ Status RegisterScalarFunction(PyObject* user_function, ScalarUdfWrapperCallback
...
+  }
+  return RegisterScalarLikeFunction(
+      user_function,
+      PythonTableUdfKernelInit{std::make_shared<OwnedRefNoGIL>(user_function), wrapper},
+      wrapper, options, registry);
+}
 
+namespace  {
+
+Result<std::shared_ptr<RecordBatch>> RecordBatchFromArray(
```
westonpace, line 206: Can you use `RecordBatch::FromStructArray` defined in `record_batch.h`?

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -105,21 +158,117 @@ Status RegisterScalarFunction(PyObject* user_function, ScalarUdfWrapperCallback
...
+    const std::string& func_name, const std::vector<Datum>& args,
+    compute::FunctionRegistry* registry) {
+  if (args.size() != 0) {
+    return Status::NotImplemented("non-empty arguments to tabular function");
+  }
+  if (registry == NULLPTR) {
+    registry = compute::GetFunctionRegistry();
+  }
+  ARROW_ASSIGN_OR_RAISE(auto func, registry->GetFunction(func_name));
+  if (func->kind() != compute::Function::SCALAR) {
```
westonpace, line 227: I think we can tackle this in a future PR but I think we will eventually want to introduce a new function kind for tabular functions.  Or, if we have to pick one of the two options we have today, I think vector would be more accurate.  Since we have no inputs I think we might get away with this for now but if we had inputs I would expect the length of the array returned from a scalar function to match the length of the incoming inputs and I don't think this will be true here.

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -105,21 +158,117 @@ Status RegisterScalarFunction(PyObject* user_function, ScalarUdfWrapperCallback
...
+    // in exec.cc and to never invoking the source function, so 1 is passed instead
+    ARROW_ASSIGN_OR_RAISE(auto datum, func_exec->Execute(args, /*passed_length=*/1));
+    if (!datum.is_array()) {
+      return Status::Invalid("UDF result of non-array kind");
+    }
+    std::shared_ptr<Array> array = datum.make_array();
+    if (array->length() == 0) {
+      return IterationTraits<std::shared_ptr<RecordBatch>>::End();
+    }
+    return RecordBatchFromArray(std::move(schema), std::move(array));
```
westonpace, line 267: Right now it seems a tabular function must return a struct array.  I think that is "ok" but for consistency with the rest of the kernels might it be better to return an ExecSpan?

Of course, we have no concept of exec span in python so the python UDF would still return a struct array.  The C++ wrapper could then convert that returned struct array into an exec span.

Although, continuing down this line, would it make more sense to have the python UDFs return record batches?  pyarrow doesn't really have much concept / functionality for struct array.

westonpace (commented):
Sorry it took me so long to do an in-depth review of this.  Overall I really like these changes, thank you for this.

I have some questions around terminology and structure but I do think I understand the concept now and I like the new functionality.

Thinking ahead I'm imagining both a "function source" node to represent a query like `SELECT * FROM my_tabular_udf()` and then also a "apply" node to represent something like `SELECT * FROM my_table t CROSS APPLY my_tabular_udf(t.id)`.  I think an "apply" node was recently proposed to substrait as well.  Does this match your understanding?

westonpace (changes requested):
Sorry, meant to click "request changes" in that last review (makes it easier for me to keep track).

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -17,35 +17,83 @@
...
+      : function_maker(function_maker), cb(cb) {
+    Py_INCREF(function_maker->obj());
+  }
+
+  Result<std::unique_ptr<compute::KernelState>> operator()(
+      compute::KernelContext* ctx, const compute::KernelInitArgs&) {
+    ScalarUdfContext udf_context{ctx->memory_pool(), /*batch_length=*/0};
+    std::unique_ptr<OwnedRefNoGIL> function;
+    RETURN_NOT_OK(SafeCallIntoPython([this, &udf_context, &function] {
+      OwnedRef empty_tuple(PyTuple_New(0));
```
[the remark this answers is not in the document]
rtpsw: The UDT kernel structure is explained in [this post](https://github.com/apache/arrow/pull/14043#issuecomment-1255332996) - here is the relevant part:

> The general idea is that the end-user is driving from PyArrow and registers a UDT. The UDT is a Python-implemented function that may be invoked multiple times, each at a different source node in the execution plan. Each such invocation returns a stream object implemented in Python that is managed in a kernel state. Invoking the kernel returns tabular data that is part of the dynamically generated stream.

The empty-tuple in the invocation here is the arguments to the UDT; in the current implementation the UDT accepts no arguments (in the future I expect it would). This UDT invocation returns a stateful stream in the form of a kernel state. It does so by calling a Python (no-arguments) function that returns a stateful Python callable, which is wrapped by the kernel state. An invocation of the stateful callable, via the kernel state, return the next tabular batch of data. So, in comparison to what you wrote, the PR has:
- KernelInit -> return a stateful stream
- KernelCall -> return the next batch in the stream

I think this is a reasonable kernel structure because the KernelInit creates state and the KernelCall uses the state to produce one output batch, as kernels normally do. IIUC, in your proposed structure, the KernelCall would produce the entire stream of batches, which is technically possible but not typical for kernels.

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -17,35 +17,83 @@
...
 
 namespace {
 
-struct PythonUdf : public compute::KernelState {
-  ScalarUdfWrapperCallback cb;
+struct PythonScalarUdfKernelState : public compute::KernelState {
+  explicit PythonScalarUdfKernelState(std::shared_ptr<OwnedRefNoGIL> function)
+      : function(function) {}
+
   std::shared_ptr<OwnedRefNoGIL> function;
```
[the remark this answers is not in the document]
rtpsw: I'll look into this.

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -17,35 +17,83 @@
...
+      return Status::OK();
+    }));
+    if (!PyCallable_Check(function->obj())) {
+      return Status::TypeError("Expected a callable Python object.");
+    }
+    return std::make_unique<PythonScalarUdfKernelState>(
+        std::move(function));
+  }
+
+  std::shared_ptr<OwnedRefNoGIL> function_maker;
```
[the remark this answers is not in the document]
rtpsw: I'll look into this.

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -68,8 +116,12 @@ struct PythonUdf : public compute::KernelState {
     // unwrapping the output for expected output type
     if (is_array(result.obj())) {
       ARROW_ASSIGN_OR_RAISE(std::shared_ptr<Array> val, unwrap_array(result.obj()));
-      if (!output_type->Equals(*val->type())) {
-        return Status::TypeError("Expected output datatype ", output_type->ToString(),
+      ARROW_ASSIGN_OR_RAISE(TypeHolder type, output_type.Resolve(ctx, batch.GetTypes()));
```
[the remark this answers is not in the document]
rtpsw: I'll look into this.

python/pyarrow/src/arrow/python/udf.cc
rtpsw: Will do.

python/pyarrow/src/arrow/python/udf.cc
rtpsw: Sounds reasonable to me.

python/pyarrow/src/arrow/python/udf.cc
rtpsw: I'll look into this. I vaguely remember that I had some reason for using `StructArray`, but maybe this can be worked out.

python/pyarrow/src/arrow/python/udf.cc
westonpace: That sounds good to me.  We can worry about distinguishing when we have a case that doesn't fit the pattern.

python/pyarrow/src/arrow/python/udf.cc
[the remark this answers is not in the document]
westonpace: I think your solution will work.  It essentially boils down to whether we want one "init" per input batch/row or one "init" per plan.  I don't know enough to know the pros / cons and I think we can always adjust as needed.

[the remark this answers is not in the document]
rtpsw: OK, keeping the existing kernel structure.

Just to clarify, KernelInit gets invoked once in both structures - mine returns a stream and yours would return a stream maker. Then, in my structure KernelCall gets invoked multiple times, returning an output batch each time, whereas in your structure it would get called just once but would return an iterator that would get invoked multiple times, returning an output batch each time. So, the difference between the two structures is syntactic. As noted, the reason I chose my structure is because it is typical to kernels.

python/pyarrow/src/arrow/python/udf.cc
rtpsw: Reviewing this, I believe I used `StructArray` because the [original result-handling code](https://github.com/apache/arrow/blob/5c1044fce55ed0e373a622cb8ee3b97a1a34799a/python/pyarrow/src/arrow/python/udf.cc#L68-L83) deals with arrays only, so introducing a new type of result should probably be done with care and consistency. You should know this original code - WDYT?

python/pyarrow/src/arrow/python/udf.cc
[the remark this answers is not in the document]
rtpsw: No, we can't do this change. The kernel init's `function` is different than the kernel state's. The former `function` refers to a UDT-maker ([example](https://github.com/apache/arrow/pull/14682/files#diff-0f8e15b029a7828bd24b7063ac03bc1e33537546b7eb7475c17d062d4d7bf1bfR522), which returns a UDT ([example](https://github.com/apache/arrow/pull/14682/files#diff-0f8e15b029a7828bd24b7063ac03bc1e33537546b7eb7475c17d062d4d7bf1bfR523)), which is referred to by the latter. Thus, each UDT is a distinct Python object that is held by a distinct kernel state.

westonpace:
@vibhatha did you want to take a look?

vibhatha:
> @vibhatha did you want to take a look?

@westonpace definitely. I will take a look.

python/pyarrow/src/arrow/python/udf.cc
rtpsw: Thinking about this some more, since "tabular"characterizes the output of a function whereas "scalar"/"vector" characterizes its unit of operation, a tabular function may be tabular-scalar or tabular-vector. So, I think the right way to capture a function being tabular is something to discuss (separately).

python/pyarrow/_compute.pxd
```diff
@@ -21,11 +21,11 @@ from pyarrow.lib cimport *
 from pyarrow.includes.common cimport *
 from pyarrow.includes.libarrow cimport *
 
-cdef class ScalarUdfContext(_Weakrefable):
+cdef class UdfContext(_Weakrefable):
```
vibhatha, line 24: For the moment, let's leave this as `ScalarUdfContext`. The reason is not that you are wrong, for the moment the version we have only supports Scalr UDFs and probably Tabular UDFs with your PR. But the aggregate UDFs are work in progress (there is a PR for that too https://github.com/apache/arrow/pull/14527). Once we finalize this work, then let's see if we want to extend from `UDFContext` and write the others depending on the way the experimental API would evolve.

python/pyarrow/_compute.pyx
```diff
@@ -2507,7 +2519,7 @@ cdef CExpression _bind(Expression filter, Schema schema) except *:
         deref(pyarrow_unwrap_schema(schema).get())))
 
 
-cdef class ScalarUdfContext:
+cdef class UdfContext:
```
vibhatha, line 2522: same here

python/pyarrow/_compute.pyx
```diff
@@ -2519,7 +2531,7 @@ cdef class ScalarUdfContext:
         raise TypeError("Do not call {}'s constructor directly"
                         .format(self.__class__.__name__))
 
-    cdef void init(self, const CScalarUdfContext &c_context):
+    cdef void init(self, const CUdfContext &c_context):
```
vibhatha, line 2534: same here

python/pyarrow/_compute.pxd
vibhatha: There is another issue, that I recall that I didn't name it like this. `CUdf` can get mixed up with `CudfContext` or related things. Need to come up with better naming too.

vibhatha: And probably if we are to do a name change for a specific reason, shall we do it in a separate PR and just keep the feature introduced to this PR?

python/pyarrow/_compute.pyx
```diff
@@ -2568,31 +2580,53 @@ cdef inline CFunctionDoc _make_function_doc(dict func_doc) except *:
...
+    cdef RegisterUdf reg = RegisterUdf.__new__(RegisterUdf)
+    reg.register_func = RegisterScalarFunction
+    return reg
+
+
+cdef get_register_tabular_function():
+    cdef RegisterUdf reg = RegisterUdf.__new__(RegisterUdf)
+    reg.register_func = RegisterTabularFunction
+    return reg
+
```
vibhatha, line 2626: appreciate your effort to make this more generic, but with the other UDFs, we have a complex strucutre for the callbacks. In that case this generalization won't be that useful as far as I feel. For the moment, shall we keep them separate? There would be a sort of `DRY` violated here, but once we generalized this API after our experimental version, we should be able to streamline these. 

cc @westonpace

python/pyarrow/tests/test_udf.py
```diff
@@ -283,7 +283,7 @@ def test_scalar_udf_array_binary(binary_func_fixture):
                           )
 
 
-def test_scalar_udf_array_ternary(ternary_func_fixture):
```
vibhatha, line 286: I see where you might have thought instead of `ScalarUdf` to go with `Udf` prefix. I am just curious, how is tabular function different from a regular scalar function in terms of usage and anatomy? Would appreciate the idea in your workds. I have sort of an idea compiled from your code, but would like to know more.

vibhatha:
@rtpsw if possible in this PR or in a follow up PR shall we add some documentation [here](https://arrow.apache.org/docs/python/compute.html#user-defined-functions) for tabular UDFs.

python/pyarrow/_compute.pxd
rtpsw: I don't mind undoing this name change. IIRC, @westonpace requested it, so he should have an opinion.

As a side note, I wouldn't generally make an effort just to avoid a conflict with another pending PR. If a PR is a step in the right direction, then this is good enough, and any conflict will be resolved in remaining PRs.

rtpsw: > There is another issue, that I recall that I didn't name it like this. `CUdf` can get mixed up with `CudfContext` or related things. Need to come up with better naming too.

I don't follow. At least in the recent commit, [`UdfContext` is a wrapper of `CUdfContext`](https://github.com/apache/arrow/blob/4236e37aaeef7fb92de4b241f3202f4e8a40463d/python/pyarrow/_compute.pyx#L2522-L2535). Where's the mix up?

python/pyarrow/_compute.pyx
rtpsw: You mean revert to the original `register_scalar_udf` code, then duplicate it and adapt to `register_tabular_udf`? The disadvantage of code repetition is clear, but what is the advantage? The remaining PR would still need to resolve the conflict, since you say the callback structure changed in that PR, only this time the conflict would not be caught by source control tools, because the `register_scalar_udf` code would not have changed.

python/pyarrow/tests/test_udf.py
rtpsw: See [this post](https://github.com/apache/arrow/pull/14682/#discussion_r1050597587) about similarity and difference.

python/pyarrow/_compute.pxd
vibhatha: I think I didn't put it accurately, I am not worried about the conflicts or rebasing the changes, but constant renaming in each PR until we figure things out.

rtpsw: I [reverted](https://github.com/apache/arrow/pull/14682/commits/8a4d820131ba919147a0704ff91acf100c745384) the naming changes.

python/pyarrow/_compute.pyx
rtpsw: Done.

python/pyarrow/_compute.pyx
rtpsw: Done.

python/pyarrow/_compute.pyx
rtpsw: The proposed common design for scalar and tabular function registration requires a parameter to distinguish between the two. I considered 3 options to do so:
1. Using a Boolean parameter, such as `is_tabular`. This seemed the least elegant to me, and is definitely the least extensible.
2. Using and enum parameter, which would currently have two values for scalar and tabular. This is better but seemed undesirable because it is not clear what this enum is. Is it the function's kind? the function output's kind? something else? See also [this post](https://github.com/apache/arrow/pull/14682/#discussion_r1050597587).
3. Using an interface parameter. This seemed elegant and extensible to me. The interface just describes how the function is registered, and is easy enough to implement. At least for the time being, the intention is for this interface to be internal.

python/pyarrow/src/arrow/python/udf.cc
westonpace: Right, I think we are balancing between:

 * Array-in / Array-out is a simple model so tabular functions can return arrays
 * Struct arrays aren't used much in pyarrow but record batches are

I think there are merits to both.  However, I don't feel too strongly about either choice.  So we can always leave it how it is and adjust if needed.  This feature will still be fairly experimental for a while I think.

python/pyarrow/src/arrow/python/udf.cc
westonpace: What would an example of a tabular-scalar function be?  Would that be a function that returns one row for each input row but might return many columns?

python/pyarrow/_compute.pyx
westonpace: My preference would be to keep this and, if we have something better when we add aggregate udfs, we can switch to that then.  That being said, I'm not 100% certain I follow the arguments here.

python/pyarrow/src/arrow/python/udf.cc
rtpsw: Exactly.

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -105,21 +192,109 @@ Status RegisterScalarFunction(PyObject* user_function, ScalarUdfWrapperCallback
...
+  std::vector<TypeHolder> in_types;
+  ARROW_ASSIGN_OR_RAISE(auto func_exec,
+                        GetFunctionExecutor(func_name, in_types, NULLPTR, registry));
+  auto next_func =
+      [schema,
+       func_exec = std::move(func_exec)]() -> Result<std::shared_ptr<RecordBatch>> {
+    std::vector<Datum> args;
+    // passed_length of -1 or 0 with args.size() of 0 leads to an empty ExecSpanIterator
+    // in exec.cc and to never invoking the source function, so 1 is passed instead
+    ARROW_ASSIGN_OR_RAISE(auto datum, func_exec->Execute(args, /*passed_length=*/1));
```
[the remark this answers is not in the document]
rtpsw: Good point. This statement is true for the recent commit in this PR. but it is not inherently true for UDTFs, since one can imagine a UDTF that respects an upper limit on the size of batches it outputs. So, I think we should document this statement and create an issue to fix this later.

python/pyarrow/tests/test_udf.py
```diff
@@ -504,3 +504,112 @@ def test_input_lifetime(unary_func_fixture):
...
+            def __call__(self, ctx):
+                try:
+                    if self.caller is None:
+                        self.caller, ctx = batch_gen(ctx).send, None
+                    batch = self.caller(ctx)
+                except StopIteration:
+                    arrays = [pa.array([], type=field.type)
+                              for field in schema]
+                    batch = pa.RecordBatch.from_arrays(
+                        arrays=arrays, schema=schema)
```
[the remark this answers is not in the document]
rtpsw: The reason is that, with the current design here, a UDTF must return an array (see [this post](https://github.com/apache/arrow/pull/14682/#discussion_r1045132971)). I don't know whether it's a good enough reason... If we'd like to allow a `StopIteration`, then I'd prefer this be done in a separate issue.

python/pyarrow/src/arrow/python/udf.cc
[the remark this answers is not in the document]
rtpsw: Added to-do comment in [recent commit](https://github.com/apache/arrow/pull/14682/commits/dc61c55437a8a7510b36b78d52bde9bab3648cbd).

rtpsw:
@westonpace, is this ready to go pending CI?

python/pyarrow/src/arrow/python/udf.cc
westonpace: This particular line will still need to change at some point though right?  Right now you're asserting this must be scalar but the tabular functions you are interested in (generators) are not scalar.

westonpace (commented):
I'm not quite sure yet what we're converging too.  In addition to my comments there is an ask in https://github.com/apache/arrow/pull/14320 to simplify the way we register UDFs and I think that ask applies here as well.

I'm also a little concerned that, without some way to execute these UDFs in the exec plan, all we've really done is come up with a very complicated way of executing python generators.  Are you planning on addressing the use of tabular UDFs in the exec plan soon?  Can we create an issue for that?

python/pyarrow/tests/test_udf.py
```diff
@@ -504,3 +504,112 @@ def test_input_lifetime(unary_func_fixture):
...
+def make_udt_func(schema, batch_gen):
+    def udf_func(ctx):
+        class UDT:
+            def __init__(self):
+                self.caller = None
+
+            def __call__(self, ctx):
+                try:
+                    if self.caller is None:
+                        self.caller, ctx = batch_gen(ctx).send, None
```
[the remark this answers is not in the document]
rtpsw: This is due to the Python generator protocol. When the `UDT` instance is initialized, it has `self.caller` set to `None`. The first time its `__call__` method is invoke, it sets up `self.caller` as the `send` method of a generator, preparing to invoke it in the next line. The first invocation of this `send` method [requires a `None` argument](https://docs.python.org/3/reference/expressions.html#generator.send), which explains the assignment you asked about. In future invocations of `__call__`, the `send` method is passed `ctx`, which is returned by the `yield` expression in the generator.

python/pyarrow/tests/test_udf.py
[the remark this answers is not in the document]
rtpsw: I agree with much of what @westonpace writes here.

> What you've done here is use regular UDFs and allow the "init" function to be "given this input create a generator" and the "execute function" to be "return the next batch of data from the generator".

There are two main views for what's going on in this PR. One view is that a tabular UDF is not a scalar UDF, and they should be modeled separately, and hence this PR is abusing the scalar UDF construction. A second view is that both a scalar UDF and a tabular UDF are forms of a common concept, which is not explicit in the code - a kind of array provider (currently, stateless with arguments, or stateful without arguments) that is internally modeled by an "init" and an "execute" functions. I think both views are valid, and I take the latter view which is the easier one.

> It smells a little bit to me of "using a hammer because you have one lying around and not because its the best tool for the job"

I believe this smell is an artifact of the first view and goes away with the second view. In particular, I think that if we decided to define a new kind of function category with the second view in mind, we'd probably be content with an "init" and "execute" functions anyway.

>  I suspect you may run into trouble once you want to start passing in arguments (you'd want those to go to the init function right? I'm not sure how you can do that).

A related discussion is [here](https://github.com/apache/arrow/pull/14682/#discussion_r1029168570).

UDT arguments wouldn't be passed to the "init" nor to the "execute" function but to the registered UDT function (like `udf_func` in the tester) returned by the Python UDT maker (like `make_udt_func` in the tester). These arguments, or values derived from them, can be saved on the returned UDT instance.

rtpsw:
> I'm not quite sure yet what we're converging too. In addition to my comments there is an ask in #14320 to simplify the way we register UDFs and I think that ask applies here as well.

Do you mean as part of this PR or in a future one?

> I'm also a little concerned that, without some way to execute these UDFs in the exec plan, all we've really done is come up with a very complicated way of executing python generators. Are you planning on addressing the use of tabular UDFs in the exec plan soon? Can we create an issue for that?

I believe the PR code supports executing a plan with a registered UDT, and it's just a matter of adding a test case or two to cover this. I could try to do that as part of this PR.

Side note: I agree we have here a rather complicated way of Arrow operating Python user-code, be it for a generator or whatever function, but if so then that applies already to UDFs, not only to UDTs.

python/pyarrow/tests/test_udf.py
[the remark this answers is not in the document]
rtpsw: Regarding `StopIteration`, since it is handled by the `UDT` class, perhaps what we'd want is to make (something like) this class part of the API so that the user would only need to provide the generator.

westonpace:
> Do you mean as part of this PR or in a future one?

Future PR is fine.  I just want to make sure that:
 * There are concrete plans to extend this (with the caveat that nothing is certain of course)
 * You know in advance there could be some upcoming changes to these APIs (still very experimental I think)

westonpace:
> Side note: I agree we have here a rather complicated way of Arrow operating Python user-code, be it for a generator or whatever function, but if so then that applies already to UDFs, not only to UDTs.

True, but for UDFs we are executing them as part of a streaming execution.  We haven't gotten there yet with UDTs (though I believe we can)

python/pyarrow/src/arrow/python/udf.cc
rtpsw: Yes, this will need to change late when we distinguish between a tabular-attribute and a scalar-attribute of a function.

python/pyarrow/tests/test_udf.py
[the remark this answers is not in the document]
westonpace: > Regarding StopIteration, since it is handled by the UDT class, perhaps what we'd want is to make (something like) this class part of the API so that the user would only need to provide the generator.

Agreed.  That would be a good solution.  It can wait for a future PR if needed.

> UDT arguments wouldn't be passed to the "init" nor to the "execute" function but to the registered UDT function (like udf_func in the tester) returned by the Python UDT maker (like make_udt_func in the tester). These arguments, or values derived from them, can be saved on the returned UDT instance.

First, to clarify, by "UDT arguments" I am referring to the input rows (i.e. a UDT with non-null arity).  I think that is what you are referring to as well.  I agree that these args will need to be passed to the equivalent of `udf_func`.  However, `udf_func` is currently called by the kernel's init function.

In other words, if I add these prints:

```
// udf.cc CallTabularFunction
  std::cout << "About to get executor" << std::endl;
  ARROW_ASSIGN_OR_RAISE(auto func_exec,
                        GetFunctionExecutor(func_name, in_types, NULLPTR, registry));
  std::cout << "Retrieved executor" << std::endl;

// test_udf.py
    def udf_func(ctx):
        class UDT:
            ...
        print("Creating UDT")
        return UDT()
```

Then I see the output:

```
About to get executor
Creating UDT
Retrieved executor
```

The problem is that a kernel's init function does not accept arguments today (e.g. you can't have a unary or binary init function).  So you will need to solve this problem when you're ready to create UDTs that have non-null arity.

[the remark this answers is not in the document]
rtpsw: > First, to clarify, by "UDT arguments" I am referring to the input rows (i.e. a UDT with non-null arity). I think that is what you are referring to as well.

I think we're on the same page here, though I actually had in mind simpler arguments, like numbers or strings, rather than (a table with) input rows, which I guess is also possible.

> The problem is that a kernel's init function does not accept arguments today (e.g. you can't have a unary or binary init function). So you will need to solve this problem when you're ready to create UDTs that have non-null arity.

Though I haven't validated this, the way I think it could work is via `KernelInitArgs`. That is, the arguments would be placed in an appropriate `FunctionOptions` and passed to `FunctionExecutor::Init`, which would deliver them to `Kernel::Init`. This would arrive at `PythonTableUdfKernelInit::operator()(compute::KernelContext* ctx, const compute::KernelInitArgs& init_args)`, whose logic would be changed to extract arguments from `init_args` and place them in a tuple, replacing the current `empty_tuple` in the PR code. These arguments would then be passed via the `cb` callback as extra argument (in addition to `ctx`) to `udf_func`, which would bake them into the UDT instance it returns.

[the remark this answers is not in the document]
rtpsw: > Agreed. That would be a good solution. It can wait for a future PR if needed.

https://github.com/apache/arrow/issues/33879

rtpsw:
> Future PR is fine. I just want to make sure that:
> 
>     * There are concrete plans to extend this (with the caveat that nothing is certain of course)
> 
>     * You know in advance there could be some upcoming changes to these APIs (still very experimental I think)

OK. Is there an existing issue for this ask, or should one be created?

westonpace:
> OK. Is there an existing issue for this ask, or should one be created?

I just created https://github.com/apache/arrow/issues/33883

westonpace:
> I think we're on the same page here, though I actually had in mind simpler arguments, like numbers or strings, rather than (a table with) input rows, which I guess is also possible.

SQL server does this with `CROSS APPLY` and I think postgres with `LATERAL JOIN`.  That being said, we can wait until someone has a compelling use case before we worry too much.  I think the case you care about is using it as a source of data? So we can refine this case before we worry too much about the advanced case.

rtpsw:
> SQL server does this with `CROSS APPLY` and I think postgres with `LATERAL JOIN`. That being said, we can wait until someone has a compelling use case before we worry too much. I think the case you care about is using it as a source of data? So we can refine this case before we worry too much about the advanced case.

Yes, my interest is in something closer to a source of data than to a cross-apply. See also [this post](https://github.com/apache/arrow/pull/14682/#issuecomment-1322610029) and the explanation linked there.

rtpsw:
> I believe the PR code supports executing a plan with a registered UDT, and it's just a matter of adding a test case or two to cover this. I could try to do that as part of this PR.

https://github.com/apache/arrow/issues/33884

PR #14683, opened 2022-11-21
Added support for S3 reader for perquet format in go.

writoblocknaut:


go/parquet/s3/s3_reader.go
```diff
@@ -0,0 +1,305 @@
...
+	if f.footerOffset <= int64(footerSize) {
+		return fmt.Errorf("parquet: file too small (size=%d)", f.footerOffset)
+	}
+
+	buf := make([]byte, footerSize)
+	// backup 8 bytes to read the footer size (first four bytes) and the magic bytes (last 4 bytes)
+	n, err := f.r.ReadAt(buf, f.footerOffset-int64(footerSize))
+	if err != nil {
+		return fmt.Errorf("parquet: could not read footer: %w", err)
+	}
```
zeroshade, line 176: All of this logic already exists in the `file` package of the Parquet module and is unnecessary to reproduce. All you need to do is have something that meets the `parquet.ReaderAtSeeker` interface and everything else will be handled automatically.

go/parquet/s3/s3_reader.go
```diff
@@ -0,0 +1,305 @@
...
+)
+
+type S3file struct {
+	source.ParquetFile
+}
+
+func (rdr S3file) ReadAt(p []byte, off int64) (n int, err error) {
+	rdr.Seek(off, io.SeekCurrent)
+	return rdr.Read(p)
+}
```
zeroshade, line 54: You could like just pass a pointer to this to `file.OpenParquetFile` and it would work. The rest of this file is then unnecessary and just copying code.

zeroshade (commented):
Ultimately, I don't think there's a need to add a specific support for S3 as anything that can support the `io.ReaderAt` + `io.Seeker` interfaces can be used with the existing library and there are multiple different S3 libraries which can provide that interface.

zeroshade:
Going to close this for now, feel free to reopen and tag me if needed

PR #14697, opened 2022-11-22
ARROW-18265: [C++][Python] Support FieldRef to work with ListElement

milesgranger:
Will fix [ARROW-18265](https://issues.apache.org/jira/browse/ARROW-18265)

- [x] Support list types other than `Type::LIST`
- [x] Add C++ tests

Example
```python
In [1]: arr = pa.array([{'a': [{'b': 1}, {'c': 'hi'}]}])

In [2]: pc.struct_field(arr, '.a[1].c')
Out[2]:
<pyarrow.lib.StringArray object at 0x7efbc741f040>
[
  "hi"
]

In [3]: pc.struct_field(arr, '.a[0].b')
Out[3]:
<pyarrow.lib.Int64Array object at 0x7efc77398fa0>
[
  1
]

In [4]:
```

### DIFF
--- a/cpp/src/arrow/compute/kernels/scalar_nested.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_nested.cc
@@ -447,11 +447,12 @@ struct StructFieldFunctor {
                                 union_array.GetFlattenedField(index, ctx->memory_pool()));
           break;
         }
-        case Type::LIST: {
-          const auto& list_array = checked_cast<const ListArray&>(*current);
+        case Type::LIST:
+        case Type::LARGE_LIST:
+        case Type::FIXED_SIZE_LIST: {
           Datum idx(index);
           ARROW_ASSIGN_OR_RAISE(Datum result,
-                                CallFunction("list_element", {list_array, idx}));
+                                CallFunction("list_element", {*current, idx}));
           current = result.make_array();
           break;
         }
@@ -468,7 +469,7 @@ struct StructFieldFunctor {
   static Status CheckIndex(int index, const DataType& type) {
     if (!ValidParentType(type)) {
       return Status::TypeError("struct_field: cannot subscript field of type ", type);
-    } else if (type.id() != Type::LIST && (index < 0 || index >= type.num_fields())) {
+    } else if (!IsBaseListType(type) && (index < 0 || index >= type.num_fields())) {
       return Status::Invalid("struct_field: out-of-bounds field reference to field ",
                              index, " in type ", type, " with ", type.num_fields(),
                              " fields");
@@ -476,9 +477,13 @@ struct StructFieldFunctor {
     return Status::OK();
   }
 
+  static bool IsBaseListType(const DataType& type) {
+    return dynamic_cast<const BaseListType*>(&type) != nullptr;
+  }
+
   static bool ValidParentType(const DataType& type) {
     return type.id() == Type::STRUCT || type.id() == Type::DENSE_UNION ||
-           type.id() == Type::SPARSE_UNION || type.id() == Type::LIST;
+           type.id() == Type::SPARSE_UNION || IsBaseListType(type);
   }
 };
 
@@ -495,8 +500,8 @@ Result<TypeHolder> ResolveStructFieldType(KernelContext* ctx,
   }
 
   for (const auto& index : field_path.indices()) {
-    if (type->id() == Type::LIST) {
-      auto list_type = checked_cast<const ListType*>(type);
+    if (StructFieldFunctor::IsBaseListType(*type->GetSharedPtr())) {
+      auto list_type = checked_cast<const BaseListType*>(type);
       type = list_type->value_type().get();
     } else {
       RETURN_NOT_OK(StructFieldFunctor::CheckIndex(index, *type));
--- a/cpp/src/arrow/compute/kernels/scalar_nested.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_nested.cc
@@ -447,6 +447,14 @@ struct StructFieldFunctor {
                                 union_array.GetFlattenedField(index, ctx->memory_pool()));
           break;
         }
+        case Type::LIST: {
+          const auto& list_array = checked_cast<const ListArray&>(*current);
+          Datum idx(index);
+          ARROW_ASSIGN_OR_RAISE(Datum result,
+                                CallFunction("list_element", {list_array, idx}));
+          current = result.make_array();
+          break;
+        }
         default:
           // Should have been checked in ResolveStructFieldType
           return Status::TypeError("struct_field: cannot reference child field of type ",
@@ -460,7 +468,7 @@ struct StructFieldFunctor {
   static Status CheckIndex(int index, const DataType& type) {
     if (!ValidParentType(type)) {
       return Status::TypeError("struct_field: cannot subscript field of type ", type);
-    } else if (index < 0 || index >= type.num_fields()) {
+    } else if (type.id() != Type::LIST && (index < 0 || index >= type.num_fields())) {
       return Status::Invalid("struct_field: out-of-bounds field reference to field ",
                              index, " in type ", type, " with ", type.num_fields(),
                              " fields");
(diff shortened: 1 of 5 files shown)


cpp/src/arrow/type.cc
```diff
@@ -1066,12 +1066,32 @@ struct FieldPathGetImpl {
...
+          children = get_children(*out);
+          index = 0;
+        } else if (out == nullptr && children->size() == 1 && index >= 0) {
+          // WIP: Perhaps a dangerous assumption?
+          // For list types when previous was not a FieldPath, but
+          // a string name referencing a list type. Then the first iteration
+          // with index (referencing the list item), might look like it's
+          // out of bounds when actually we don't care about this index for
+          // type checking, only the kernel does; we need the list value type
+          // which is now the only child in the children vector.
```
jorisvandenbossche, line 1090: Can you show a concrete example of the case you are thinking of?

(also adding some tests for FieldPath::Get new capabilities (in C++) will probably be helpful to undestand the behaviour)

milesgranger: Sure, this path happens with the `".a[1].c"` case, where the 'a' goes through string name, and the second '[1]' is through this FieldPath, but is also out of bounds from children (index 1) whereas `".a[0].c"` works b/c 0 is the index we want anyway from the list type. 

I'll work on adding the FieldPath::Get tests in C++. Thanks!

jorisvandenbossche: But at this point we always have FieldPaths, no? (regardless whether they originally came from string names or not?)

milesgranger: The first 'a', in our example, will be matched [here in the second overloaded member function](https://github.com/apache/arrow/blob/7ae4705c629655eb37b4200229d30f3a2af9e154/cpp/src/arrow/type.cc#L1407) then the second (`[1]`) goes to the first overloaded function taking `FieldPath`.

cpp/src/arrow/type_test.cc
```diff
@@ -434,6 +434,51 @@ TEST(TestFieldRef, DotPathRoundTrip) {
...
+        {field("c", list_type(struct_({field("d", int32()), field("e", int8())})))});
+
+    // Note numeric values here go unused, outside of indicating a further step, during
+    // FindAll, only used by kernels for which list element to take. When FindAll
+    // encounters a list, it assumes the list value type. Not b/c they are numeric, but
+    // they are the position referencing a specific list element.
+    EXPECT_THAT(FieldRef("c", 0, "d").FindAll(*type), ElementsAre(FieldPath{0, 0, 0}));
+    EXPECT_THAT(FieldRef("c", 0, "e").FindAll(*type), ElementsAre(FieldPath{0, 0, 1}));
+    EXPECT_THAT(FieldRef("c", 1, "d").FindAll(*type), ElementsAre(FieldPath{0, 1, 0}));
+    EXPECT_THAT(FieldRef("c", 1, "e").FindAll(*type), ElementsAre(FieldPath{0, 1, 1}));
```
jorisvandenbossche, line 455: Maybe also cover "failure" cases like `FieldRef("c", "non-integer", "d").FindAll(*type)`  where there is not an integer index at the location of the list level.

jorisvandenbossche (commented):
(just a quick look at the tests)

cpp/src/arrow/type_test.cc
milesgranger: Good call adding the variant for leading list type, it did fail when index > 0 as the parent wasn't being passed in that scenario. Works in https://github.com/apache/arrow/pull/14697/commits/c2fd9eee2fb0d7e536441394d2fd7912844ee61c but will come back later to dress it up a bit more. :)

cpp/src/arrow/compute/kernels/scalar_nested.cc
```diff
@@ -460,17 +469,21 @@ struct StructFieldFunctor {
   static Status CheckIndex(int index, const DataType& type) {
     if (!ValidParentType(type)) {
       return Status::TypeError("struct_field: cannot subscript field of type ", type);
-    } else if (index < 0 || index >= type.num_fields()) {
+    } else if (!IsBaseListType(type) && (index < 0 || index >= type.num_fields())) {
```
jorisvandenbossche, line 472: In theory, we could check the index for fixed sized list arrays as well (not for variable size list arrays, though, so not sure that is worth it)

cpp/src/arrow/type.cc
```diff
@@ -1060,19 +1060,40 @@ struct FieldPathGetImpl {
...
+        auto IsBaseListType = [](const DataType& type) {
+          return dynamic_cast<const BaseListType*>(&type) != nullptr;
+        };
+
+        // For lists, we don't care about the index, jump right into the list value type.
+        // The index here is used in the kernel to grab the specific list element.
+        // Children then are fields from list type, and out will be the children from
+        // that field, thus index 0.
+        if (out != nullptr && IsBaseListType(*(*out)->type())) {
+          children = get_children(*out);
```
jorisvandenbossche, line 1087: I don't fully understand why we need the `get_children` here (because then we call `get_children` twice in a single iteration? since there is another call a bit below) 
Can you add a clarifying comment about that?

jorisvandenbossche: (or to put it differently: why isn't this initial get_children needed when parent is a struct type? What's the difference between both?)

cpp/src/arrow/type_test.cc
```diff
@@ -434,6 +434,64 @@ TEST(TestFieldRef, DotPathRoundTrip) {
...
+  for (auto list_type : {gen_type(list), gen_type(large_list)}) {
+    // Single list type
+    auto type = struct_(
+        {field("c", list_type(struct_({field("d", int32()), field("e", int8())})))});
+
+    // Note numeric values here go unused, outside of indicating a further step, during
+    // FindAll, only used by kernels for which list element to take. When FindAll
+    // encounters a list, it assumes the list value type. Not b/c they are numeric, but
+    // they are the position referencing a specific list element.
+    EXPECT_THAT(FieldRef("c", 0, "d").FindAll(*type), ElementsAre(FieldPath{0, 0, 0}));
```
jorisvandenbossche, line 452: Can you also add variants of those tests that pass a schema instead of type to `FindAll`?

pitrou:
I find it a bit uneasy how this PR is conflating a purely metadata-based field lookup (the current semantics of `FieldRef`) and a dynamic index lookup into a list. Is this actually important to support?

jorisvandenbossche:
> a purely metadata-based field lookup (the current semantics of `FieldRef`)

I am not sure that this is still entirely correct nowadays, since we already do use `FieldRef` in other places in the compute engine to refer to fields (eg to specify the sort key, in projections (although there wrapped in an expression), ...). 
And it is certainly true that right now we only support struct/union field lookups, but also those are not purely metadata-based, as the actual lookup of values (not just the type/field from a schema) also needs to potentially calculate validity bitmaps. While for list types, the lookup of the type is also metadata-only. It's only when looking up actual values that you need a nontrivial computation.

> Is this actually important to support?

If you want to work with list types, this is a pretty basic operation. Now, the actual operation is already supported (`list_element` kernel), here it is only about being able to express this in a field reference. So this is somewhat in the bucket of "user convenience", as you can already achieve something similar with the kernel. 
Now, for example also Substrait has this concept (https://github.com/substrait-io/substrait/blob/7f272f13f22cd5f5842baea42bcf7961e6251881/proto/substrait/algebra.proto#L932-L938)

Another example of database that supports this in queries: https://cloud.google.com/bigquery/docs/reference/standard-sql/arrays#accessing_array_elements

jorisvandenbossche:
And to be clear, that's also basically the question that I asked on the JIRA: "do we want to use FieldRef/FieldPath for this?" (because it does deviate from how it is currently implemented) It could also be a different (new to add) data structure (that would be used throughout the compute code instead of the current FieldRef/FieldPath)

pitrou:
Hmm... The convenience argument is convincing, I admit.

Still, I wonder if we shouldn't make it optional in FieldRef, perhaps by adding a `bool allow_list_lookup = false` argument to FieldRef lookup APIs?

amol-:

PR #14698, opened 2022-11-22
ARROW-18360: [Python] Don't crash when schema=None in FlightClient.do_put

milesgranger:
Will close [ARROW-18360](https://issues.apache.org/jira/browse/ARROW-18360)


python/pyarrow/_flight.pyx
```diff
@@ -1548,6 +1548,9 @@ cdef class FlightClient(_Weakrefable):
         writer : FlightStreamWriter
         reader : FlightMetadataReader
         """
+        if schema is None:
```
lidavidm, line 1551: Cython lets you declare the parameter as `Schema schema not None` which would be better https://cython.readthedocs.io/en/latest/src/userguide/extension_types.html#extension-types-and-none

lidavidm: Also, in general I question if `pyarrow_unwrap_schema` et al should accept None by default, it seems such errors would be caught if they raised and translating None to nullptr had to be opt in

lidavidm: Finally, it seems the C++ side should guard against nullptr better too

milesgranger: Yes, the `not None` is nice, I initially used it in a previous PR but it was advised to raise a more informative error; I can make the switch back here.

> Also, in general I question if pyarrow_unwrap_schema et al should accept None by default, it seems such errors would be caught if they raised and translating None to nullptr had to be opt in

I checked the impl of `pyarrow_unwrap_schema` and it seems quite decisive in it's willingness to return an empty pointer so left it alone; it anyway maps nice enough from Python's `None` to `nullptr` in it's current form. :sweat_smile: 

> Finally, it seems the C++ side should guard against nullptr better too

Agreed, I'll try to make that better in this PR, thanks. :)

milesgranger: Updated in https://github.com/apache/arrow/pull/14698/commits/d38124fc1276757c2eb04b8c77afc67247284766, when testing with removing the `not None` check, the error is somewhat uninformative so left the `not None` in Cython.

```python
In [1]: import pyarrow.flight as flight

In [2]: client = flight.FlightClient('grpc+tls://localhost:9643', disable_server_verification=True)

In [3]: client.do_put(flight.FlightDescriptor.for_command('foo'), schema=None)
FlightUnavailableError                    Traceback (most recent call last)
Cell In [3], line 1
----> 1 client.do_put(flight.FlightDescriptor.for_command('foo'), schema=None)

File ~/Projects/arrow/python/pyarrow/_flight.pyx:1559, in pyarrow._flight.FlightClient.do_put()
   1557
   1558         with nogil:
-> 1559             check_flight_status(self.client.get().DoPut(
   1560                 deref(c_options),
   1561                 c_descriptor,

File ~/Projects/arrow/python/pyarrow/_flight.pyx:71, in pyarrow._flight.check_flight_status()
     69             raise FlightUnauthorizedError(message, detail_msg)
     70         elif detail.get().code() == CFlightStatusUnavailable:
---> 71             raise FlightUnavailableError(message, detail_msg)
     72
     73 size_detail = FlightWriteSizeStatusDetail.UnwrapStatus(status)

FlightUnavailableError: Flight returned unavailable error, with message: failed to connect to all  
addresses; last error: UNKNOWN: Failed to connect to remote host: Connection refused. gRPC 
client debug context: UNKNOWN:Failed to pick subchannel {created_time:"2022-11-
22T20:23:29.049965339+01:00", children:[UNKNOWN:failed to connect to all addresses; last error: 
UNKNOWN: Failed to connect to remote host: Connection refused {grpc_status:14, 
created_time:"2022-11-22T20:23:29.049961187+01:00"}]}. Additionally, could not finish writing 
record batches before closing. Client context: Invalid: nullptr for Schema not allowed.
```

cpp/src/arrow/ipc/writer.cc
```diff
@@ -1377,6 +1377,10 @@ namespace internal {
 Result<std::unique_ptr<RecordBatchWriter>> OpenRecordBatchWriter(
     std::unique_ptr<IpcPayloadWriter> sink, const std::shared_ptr<Schema>& schema,
     const IpcWriteOptions& options) {
+  // constructor for IpcFormatWriter here dereferences ptr to schema.
+  if (schema == nullptr) {
+    return Status::Invalid("nullptr for Schema not allowed.");
```
lidavidm, line 1382: nit: no need for a period here

lidavidm (approved):

BryanCutler (approved):
LGTM, thanks for the fix @milesgranger

lidavidm:
Thanks. Looks like the MinGW failures are unrelated.

PR #14700, opened 2022-11-22
MINOR: [C++] Fix CMake deps for minimal Flight benchmark build

lidavidm:
Otherwise it errors if you try to just build the Flight benchmark


kou (approved):
+1

PR #14702, opened 2022-11-22
ARROW-18382: [C++] Set ADDRESS_SANITIZER in fuzzing builds

pitrou:
This will silence a spurious reported leak in OSS-Fuzz builds.


pitrou:

kou (approved):
+1

PR #14704, opened 2022-11-22
ARROW-18383: [C++] Avoid global variables for thread pools and at-fork handlers

pitrou:
Initialization order of module globals is undefined.
    
In a particular case, the IO thread pool would first be instantiated at library load, registering an at-fork handler.
Then, only after, the at-fork handlers would be initialized, losing the handler registered just before.

### DIFF
--- a/cpp/src/arrow/io/interfaces.cc
+++ b/cpp/src/arrow/io/interfaces.cc
@@ -50,12 +50,14 @@ using internal::ThreadPool;
 
 namespace io {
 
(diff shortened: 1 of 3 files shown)


pitrou:

pitrou:
@paleolimbot This seems to fix the R issues with `mclapply` you discovered.

pitrou:

lidavidm (approved):

PR #14705, opened 2022-11-22
ARROW-15812: [R] Accept col_names in open_dataset for CSV

wjones127:


r/R/csv.R
```diff
@@ -478,7 +478,7 @@ CsvWriteOptions$create <- function(include_header = TRUE, batch_size = 1024L) {
   )
 }
 
-readr_to_csv_read_options <- function(skip, col_names, col_types) {
+readr_to_csv_read_options <- function(skip = 0, col_names = TRUE, col_types = NULL) {
```
thisisnic, line 481: Given it's not used here, should we just remove the `col_types` argument entirely?

r/R/dataset-format.R
```diff
@@ -312,7 +326,30 @@ csv_file_format_read_opts <- function(schema = NULL, ...) {
...
+
+  if (any(is_readr_opt)) {
+    # Catch cases when the user specifies a mix of Arrow C++ options and
+    # readr-style options
+    if (!all(is_readr_opt)) {
+      abort(c(
+        "Use either Arrow read options or readr read options, not both.",
+        i = sprintf("Passed Arrow options: %s.", oxford_paste(opt_names[is_arrow_opt])),
+        i = sprintf("Passed readr options: %s.", oxford_paste(opt_names[is_readr_opt]))
+      ))
```
thisisnic, line 347:
```suggestion
      abort(c(
        "Additional CSV reading options must be Arrow-style or readr-style, but not both.",
        i = sprintf("Arrow options used: %s.", oxford_paste(opt_names[is_arrow_opt])),
        i = sprintf("readr options used: %s.", oxford_paste(opt_names[is_readr_opt]))
      ))
```
How about this slight rephrase which shifts the focus from the person using the function to the usage of the function itself, and adds in "must" as per [tidyverse style](https://style.tidyverse.org/error-messages.html)?

r/tests/testthat/test-dataset-csv.R
```diff
@@ -322,6 +320,49 @@ test_that("Column names can be inferred from schema", {
...
+  ds <- open_dataset(
+    header_csv_dir,
+    format = "csv",
+    col_names = FALSE,
+    skip = 1
+  )
+  expect_equal(names(ds), c("f0", "f1"))
+  expect_equal(ds %>% collect(), set_names(tbl, c("f0", "f1")))
+
+  expect_error(
```
thisisnic, line 360: The text of this error comes from C++ so we shouldn't test it in R; we could either catch this error and add in a more R-style error message, or just test that the code raises an error without testing the specific text of the error itself.

r/tests/testthat/test-dataset-csv.R
```diff
@@ -376,7 +417,7 @@ test_that("skip argument in open_dataset", {
     header_csv_dir,
     format = "csv",
     schema = schema(int = int32(), dbl = float64()),
-    skip = 1
+    skip_rows = 1
```
thisisnic, line 420: Why this change?

thisisnic (commented):
Thanks for making this change! Just a few comments.

r/tests/testthat/test-dataset-csv.R
wjones127: Ah looked at this again and I shouldn't have changed it. It was complaining that `skip` was a readr style arg, but we were internally appending an Arrow-style arg `column_names` when we passed an explicit schema. It's a little messy, but I basically changed it so internally we choose to add `column_names` (Arrow-style) or `col_names` (readr-style) based on what other arguments were passed.

r/R/csv.R
thisisnic: OK, I see we use it later when we're listing which ones we do and do not support.  Maybe this bit of the code needs refactoring in other ways, as it feels a bit brittle to depend on this here?

thisisnic (approved):

PR #14708, opened 2022-11-22
ARROW-18292: [Release][Python] Upload .wheel/.tar.gz for release not RC

kou:


raulcd (approved):
LGTM, I've tested the change locally and downloaded correctly the 25 artifacts

PR #14715, opened 2022-11-23
ARROW-18390: [CI][Python] Update spark test modules to match spark master

raulcd:


raulcd:
@github-actions crossbow submit test-conda-python-*-spark-*

raulcd:
Match the spark module refactoring done on this commit: https://github.com/apache/spark/commit/4f2630e308caa107d1addedbf94a919a18ab3ed4

raulcd:
@github-actions crossbow submit test-conda-python-*-spark-*

raulcd:
@jorisvandenbossche what do you think about this to fix the nightly failures? I am not sure in which version of spark this will be released, I suppose if we decide to add newer versions of spark to the CI matrix that contain this change we can update the if to match the newer versions but this currently fixes the issue.

kou:
@kiszk You may want to review this as an Apache Spark committer.

kiszk:
@raulcd Thank you. Almost looks good to me.

Can we enhance the decision logic of `SPARK_VERSION`?  If version is (<= 2 or) <= 3.3, we use the current package names. Otherwise, we use the new package names. As a result, we could avoid further changes for a while.

jorisvandenbossche:
I would personally also be fine with just merging this as is and handling updating it once it start failing again when those test changes are released, but if there is an easy way to check the actual spark version, that sounds good as well.

AlenkaF:
@github-actions crossbow submit test-conda-python-*-spark-*

AlenkaF:
@github-actions crossbow submit test-conda-python-*-spark-*

kiszk:
nit: `if [ ${old_test_modules} == "true" ]; then` -> `if [ "${old_test_modules}" == "true" ]; then`

kiszk:
@github-actions crossbow submit test-conda-python--spark-

raulcd:
@github-actions crossbow submit test-conda-python-*-spark-*

raulcd:
@kiszk GitHub markdown is removing the required `*`, should be: `@github-actions crossbow submit test-conda-python-*-spark-*`. I might have permissions again :) trying

kiszk:
The problem seems to be a format of `SPARK_VERSION`.

https://github.com/ursacomputing/crossbow/actions/runs/3539915127/jobs/5942357685#step:5:7657 implies `SPARK_VERSION` is `v3.1.2`, not `3.1.2`. On the other hand, `integration_spark.sh` assumes that the format is `3.1.2`.

IMHO, we should take a major version (e.g. `3.`) by using `${SPARK_VERSION:1:2}`.

ci/scripts/integration_spark.sh
```diff
@@ -73,14 +73,38 @@ pushd ${spark_dir}
...
+  # Spark <= 2.x or <= 3.3.x
+  if [ "${SPARK_VERSION:0:2}" == "2." ]; then
+    old_test_modules=true
+  elif [ "${SPARK_VERSION:0:2}" == "3." ]; then
+    if [ "${SPARK_VERSION:2:2}" == "1." ] || \
+       [ "${SPARK_VERSION:2:2}" == "2." ] || \
+       [ "${SPARK_VERSION:2:2}" == "3." ]; then
+        old_test_modules=true
+    fi
+  fi
```
kou, line 88: FYI: We may be able to use `case` for this:

```suggestion
  case "${SPARK_VERSION}" in
    1.*|2.*|3.0.*|3.1.*|3.2.*|3.3.*)
      old_test_modules=true
      ;;
    *)
      old_test_modules=false
      ;;
  esac
```

kiszk: Beautiful. Maybe `v1.*|v2.*|v3.0.*|v3.1.*|v3.2.*|v3.3.*` could work in this case.

raulcd:
@github-actions crossbow submit test-conda-python-*-spark-*

ci/scripts/integration_spark.sh
raulcd: Thanks @kou ! Yes, this is much better than what I was doing!

raulcd:
> @raulcd The problem seems to be a format of `SPARK_VERSION`.

@kiszk you are correct. I got confused by the other check on the script which seems to be wrong then:
```
if [ "${SPARK_VERSION:0:2}" == "2." ]; then
  # https://github.com/apache/spark/blob/master/docs/sql-pyspark-pandas-with-arrow.md#compatibility-setting-for-pyarrow--0150-and-spark-23x-24x
  export ARROW_PRE_0_15_IPC_FORMAT=1
fi
```
I've updated to:
```

if [ "${SPARK_VERSION:1:2}" == "2." ]; then
```
even though it was not used as we don't have any nightly test for versions lower than `3.1.2`.

raulcd:
@kiszk @kou this should be ready now, see the last builds from crossbow. Thanks for your help.

kiszk (approved):

kou (approved):
+1

PR #14714, opened 2022-11-23
ARROW-18389: [CI][Python] Update nightly test-conda-python-3.7-pandas-0.24 to pandas >= 1.0

raulcd:

PR #14716, opened 2022-11-23
ARROW-18392: [Python] Fix test_s3fs_wrong_region; set anonymous=True

milesgranger:
Will fix [ARROW-18392](https://issues.apache.org/jira/browse/ARROW-18392)

### DIFF
--- a/python/pyarrow/tests/test_fs.py
+++ b/python/pyarrow/tests/test_fs.py
@@ -1321,6 +1321,7 @@ def test_s3fs_wrong_region():
     from pyarrow.fs import S3FileSystem
 
     # wrong region for bucket
+    # anonymous=True incase CI/etc has invalid credentials
     fs = S3FileSystem(region='eu-north-1', anonymous=True)
 
     msg = ("When getting information for bucket 'voltrondata-labs-datasets': "
--- a/python/pyarrow/tests/test_fs.py
+++ b/python/pyarrow/tests/test_fs.py
@@ -1321,7 +1321,7 @@ def test_s3fs_wrong_region():
     from pyarrow.fs import S3FileSystem
 
     # wrong region for bucket
-    fs = S3FileSystem(region='eu-north-1')
+    fs = S3FileSystem(region='eu-north-1', anonymous=True)
 
     msg = ("When getting information for bucket 'voltrondata-labs-datasets': "
            r"AWS Error UNKNOWN \(HTTP status 301\) during HeadBucket "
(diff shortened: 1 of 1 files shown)


raulcd:
@github-actions crossbow submit test-conda-python-*-pandas-*

python/pyarrow/tests/test_fs.py
```diff
@@ -1321,22 +1321,23 @@ def test_s3fs_wrong_region():
     from pyarrow.fs import S3FileSystem
 
     # wrong region for bucket
-    fs = S3FileSystem(region='eu-north-1')
+    fs = S3FileSystem(region='eu-north-1', anonymous=True)
```
assignUser, line 1324: I'd add a comment explaining why this is important. Do we have other tests s3 tests that could also need this?

raulcd:
@github-actions crossbow submit test-conda-python-*-pandas-*

python/pyarrow/tests/test_fs.py
milesgranger: Done, and I found only one other test that uses a real AWS call, aptly named `test_s3_real_aws` but that already sets `anonymous=True`. All others appear to use the endpoint override / minio.

raulcd:
The pandas-0.24 job failure is expected to fail and should be fixed on: https://github.com/apache/arrow/pull/14714
The pandas nightly and upstream_devel failures are unrelated and I think should be fixed on its own ticket. I have opened this ticket to track them: https://issues.apache.org/jira/browse/ARROW-18394

python/pyarrow/tests/test_fs.py
```diff
@@ -1321,22 +1321,24 @@ def test_s3fs_wrong_region():
...
-    with pytest.raises(OSError, match=msg) as exc:
+           "'eu-north-1' while the bucket is located in 'us-east-2'.")
+    with pytest.raises(OSError) as exc:
         fs.get_file_info("voltrondata-labs-datasets")
 
-    # Sometimes fails on unrelated network error, so next call would also fail.
-    if 'NETWORK_CONNECTION' in str(exc.value):
+    # Sometimes fails on unrelated network or permission error, so next
+    # call would also fail.
+    if 'UNKNOWN' not in str(exc.value):
```
pitrou, line 1336: Why "not in"? Don't we want this condition to be a bit more specific?

milesgranger: Ya, good point. Since the main issue was `anonymous=True` needing to be set, I'll reset everything else.

pitrou (approved):
+1, thank you @milesgranger

PR #14718, opened 2022-11-23
ARROW-18397: [C++] Clear S3 region resolver client at S3 shutdown

pitrou:
This should hopefully suppress a failed assertion on recent AWS SDK versions.

See https://github.com/aws/aws-sdk-cpp/issues/2204 for upstream issue report.


pitrou:
@kou This fixes the test-s3fs crashes on MinGW builds.

kou (approved):
+1

Thanks!

PR #14722, opened 2022-11-23
GH-14755: [Python] Expose QuotingStyle to Python

frederickjansen:
This exposes the QuotingStyle option to Python. I tested this with the v10 tag since I wasn't able to build the master branch for some reason.
I'm happy to use enums instead of strings. Both seem to be used in Python and I wasn't sure which one was preferred.

(no Jira ticket since I don't have an account and can't create one)


python/pyarrow/tests/test_csv.py
```diff
@@ -1908,6 +1908,20 @@ def test_write_read_round_trip():
...
+    t = pa.Table.from_arrays([[1, 2, 3], ["a", "b", "c"]], ["c1", "c2"])
+    buf = io.BytesIO()
+    for write_options, res in [
+        (WriteOptions(quoting_style='none'), b'"c1","c2"\n1,a\n2,b\n3,c\n'),
+        (WriteOptions(), b'"c1","c2"\n1,"a"\n2,"b"\n3,"c"\n'),
+        (WriteOptions(quoting_style='all_valid'), b'"c1","c2"\n"1","a"\n"2","b"\n"3","c"\n'),
+    ]:
+        with CSVWriter(buf, t.schema, write_options=write_options) as writer:
+            writer.write_table(t)
+        buf.seek(0)
```
pitrou, line 1921: No need for `seek(0)` before `getvalue()` AFAIR.

python/pyarrow/tests/test_csv.py
```diff
@@ -1908,6 +1908,20 @@ def test_write_read_round_trip():
                              parse_options=parse_options)
 
 
+def test_write_quoting_style():
+    t = pa.Table.from_arrays([[1, 2, 3], ["a", "b", "c"]], ["c1", "c2"])
```
pitrou, line 1912: Let's add some null values there?

pitrou: For example:
```suggestion
    t = pa.Table.from_arrays([[1, 2, None], ["a", None, "c"]], ["c1", "c2"])
```

python/pyarrow/tests/test_csv.py
```diff
@@ -1908,6 +1908,20 @@ def test_write_read_round_trip():
                              parse_options=parse_options)
 
 
+def test_write_quoting_style():
```
pitrou, line 1911: Can you also expand `test_write_options` on line 324 above?

python/pyarrow/tests/test_csv.py
```diff
@@ -1908,6 +1908,20 @@ def test_write_read_round_trip():
...
+    for write_options, res in [
+        (WriteOptions(quoting_style='none'), b'"c1","c2"\n1,a\n2,b\n3,c\n'),
+        (WriteOptions(), b'"c1","c2"\n1,"a"\n2,"b"\n3,"c"\n'),
+        (WriteOptions(quoting_style='all_valid'), b'"c1","c2"\n"1","a"\n"2","b"\n"3","c"\n'),
+    ]:
+        with CSVWriter(buf, t.schema, write_options=write_options) as writer:
+            writer.write_table(t)
+        buf.seek(0)
+        assert buf.getvalue() == res
+
```
pitrou, line 1923: Perhaps also test with special characters? This would probably raise with `quoting_style='none'`

python/pyarrow/_csv.pyx
```diff
@@ -1337,6 +1355,17 @@ cdef class WriteOptions(_Weakrefable):
     def delimiter(self, value):
         deref(self.options).delimiter = _single_char(value)
 
+    @property
+    def quoting_style(self):
+        """
+        Quoting style for CSV writing.
```
pitrou, line 1361: We would rather reuse the same description as in the `WriteOptions` docstring above.

python/pyarrow/_csv.pyx
```diff
@@ -1288,20 +1301,25 @@ cdef class WriteOptions(_Weakrefable):
         CSV data
     delimiter : 1-character string, optional (default ",")
         The character delimiting individual cells in the CSV data.
+    quoting_style : str, optional (default "needed")
+        Whether to quote values, and if so, which quoting style to use.
+        Accepted values are "needed", "all_valid", "none"
```
pitrou, line 1306: We would probably like to expand the explanation a bit, for example:
```suggestion
        The following values are accepted:
        - "needed" (default): only enclose values in quotes when needed.
        - "all_valid": enclose all valid values in quotes; nulls are not quoted.
        - "none": do not enclose any values in quotes; values containing
          special characters (such as quotes, cell delimiters or line endings)
          will raise an error.
```

python/pyarrow/_csv.pyx
```diff
@@ -1337,6 +1355,17 @@ cdef class WriteOptions(_Weakrefable):
     def delimiter(self, value):
         deref(self.options).delimiter = _single_char(value)
 
+    @property
+    def quoting_style(self):
+        """
+        Quoting style for CSV writing.
+        """
+        return deref(self.options).quoting_style
```
pitrou, line 1363: Does this actually return the same values as expected by the constructor and property setter?

pitrou (changes requested):
Thanks @frederickjansen! You'll find some comments below.

Also, a new issue should be created to associate with this PR. Do you have a [JIRA](https://issues.apache.org/jira/projects/ARROW/) account? If not, you can create a GitHub issue instead.

frederickjansen:
Thanks for the feedback @pitrou. I'll make the changes. I also created ticket https://github.com/apache/arrow/issues/14755.

Should I change the interface to an Enum instead of strings? I see both in the project and I'm not sure which one is preferable.

python/pyarrow/tests/test_csv.py
frederickjansen: I believe it's still needed to start writing from the start of the buffer again in the next loop. Otherwise the buffer contains the contents of the previous test. I could put the call after `getvalue()` as well if that makes it more clear.

python/pyarrow/tests/test_csv.py
frederickjansen: Done

python/pyarrow/_csv.pyx
frederickjansen: Done

python/pyarrow/_csv.pyx
frederickjansen: Done, thanks for writing this out.

python/pyarrow/_csv.pyx
frederickjansen: It converted to `int` since that's what the C++ enums default to I think. Fixed by doing the reverse of `unwrap_quoting_style`.

python/pyarrow/tests/test_csv.py
pitrou: It's ok, thanks.

pitrou:
> Should I change the interface to an Enum instead of strings? I see both in the project and I'm not sure which one is preferable.

I think a string is ok in this case.

pitrou (approved):

pitrou:
@jorisvandenbossche Do you also want to take a look at this?

python/pyarrow/_csv.pyx
```diff
@@ -1288,20 +1312,30 @@ cdef class WriteOptions(_Weakrefable):
         CSV data
     delimiter : 1-character string, optional (default ",")
         The character delimiting individual cells in the CSV data.
+    quoting_style : str, optional (default "needed")
+        Whether to quote values, and if so, which quoting style to use.
+        The following values are accepted:
+        - "needed" (default): only enclose values in quotes when needed.
```
jorisvandenbossche, line 1318:
```suggestion
        The following values are accepted:

        - "needed" (default): only enclose values in quotes when needed.
```

(small formatting nit, but I _think_ this is needed for it to render correctly in the html version, restructured text typically needs a blank line before a list)

python/pyarrow/_csv.pyx
```diff
@@ -1337,6 +1371,23 @@ cdef class WriteOptions(_Weakrefable):
     def delimiter(self, value):
         deref(self.options).delimiter = _single_char(value)
 
+    @property
+    def quoting_style(self):
+        """
+        Whether to quote values, and if so, which quoting style to use.
+        The following values are accepted:
+        - "needed" (default): only enclose values in quotes when needed.
```
jorisvandenbossche, line 1379:
```suggestion
        The following values are accepted:

        - "needed" (default): only enclose values in quotes when needed.
```

same here

jorisvandenbossche (approved):
Looks good! Just a small formatting comment

frederickjansen:
Thanks @jorisvandenbossche, good catch. The list now looks good when building the docs.

PR #14719, opened 2022-11-23
MINOR: [Developer] Add triage users

assignUser:
This PR adds @raulcd @toddfarmer and me as "triage users" ([INFRA docs](https://cwiki.apache.org/confluence/display/INFRA/Git+-+.asf.yaml+features#Git.asf.yamlfeatures-AssigningexternalcollaboratorswiththetriageroleonGitHub)). This gives us permissions to manage issues.  

@toddfarmer has been spearheading the JIRA -> Github migration and @raulcd and me will need to work on the release and merge scripts to adapt them to the migration.

PR #14727, opened 2022-11-24
MINOR: [Archery] Add collaborators to list of roles with access to trigger bot tasks

raulcd:
I've been added as triager but now if I try to trigger crossbow tasks they fail with permissions issues:
https://github.com/apache/arrow/pull/14715#issuecomment-1326243342
```
Only contributors can submit requests to this bot. Please ask someone from the community for help with getting the first commit in.
The Archery job run can be found at: https://github.com/apache/arrow/actions/runs/3539688563
```
This should fix the issue

### DIFF
--- a/dev/archery/archery/bot.py
+++ b/dev/archery/archery/bot.py
@@ -143,7 +143,7 @@ class CommentBot:
             # https://developer.github.com/v4/enum/commentauthorassociation/
             # Checking  privileges here enables the bot to respond
             # without relying on the handler.
-            allowed_roles = {'OWNER', 'MEMBER', 'CONTRIBUTOR'}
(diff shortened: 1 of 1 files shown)


pitrou:
I didn't know about the "collaborator" role. This looks good to me. @kou What do you think?

raulcd:
@pitrou for reference: https://github.com/apache/arrow/pull/14719

jorisvandenbossche (approved):
Since we already allow `contributor`, which is "lower in rank" as `collaborator`, this seems logical to do

assignUser (approved):
Thanks!

kou:
+1

PR #14729, opened 2022-11-24
ARROW-18399: [Python] Reduce warnings during tests

milesgranger:
Will fix [ARROW-18399](https://issues.apache.org/jira/browse/ARROW-18399)

Also closes [ARROW-18125](https://issues.apache.org/jira/browse/ARROW-18125)

### DIFF
--- a/python/pyarrow/parquet/core.py
+++ b/python/pyarrow/parquet/core.py
@@ -1165,10 +1165,8 @@ class ParquetDatasetPiece:
         -------
         metadata : FileMetaData
         """
-        parquet = self.open()
-        meta = parquet.metadata
-        parquet.close(True)
-        return meta
+        with self.open() as parquet:
+            return parquet.metadata
 
     def open(self):
         """
@@ -1177,6 +1175,10 @@ class ParquetDatasetPiece:
         reader = self.open_file_func(self.path)
         if not isinstance(reader, ParquetFile):
             reader = ParquetFile(reader, **self.file_options)
+
+        # ensure reader knows it's responsible for closing source
+        # since we opened the source here internally.
+        reader._close_source = True
         return reader
 
     def read(self, columns=None, use_threads=True, partitions=None,
--- a/python/pyarrow/parquet/core.py
+++ b/python/pyarrow/parquet/core.py
@@ -1165,8 +1165,10 @@ class ParquetDatasetPiece:
         -------
         metadata : FileMetaData
         """
-        with self.open() as parquet:
-            return parquet.metadata
+        parquet = self.open()
+        meta = parquet.metadata
+        parquet.close(True)
+        return meta
 
     def open(self):
         """
--- a/python/pyarrow/tests/parquet/test_datetime.py
+++ b/python/pyarrow/tests/parquet/test_datetime.py
@@ -17,6 +17,7 @@
 
 import datetime
 import io
+import warnings
 
 import numpy as np
 import pytest
@@ -321,7 +322,11 @@ def test_coerce_int96_timestamp_overflow(pq_reader_method, tempdir):
     # with the default resolution of ns, we get wrong values for INT96
     # that are out of bounds for nanosecond range
     tab_error = get_table(pq_reader_method, filename)
-    assert tab_error["a"].to_pylist() != oob_dts
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore",
+                                "Discarding nonzero nanoseconds in conversion",
+                                UserWarning)
+        assert tab_error["a"].to_pylist() != oob_dts
 
     # avoid this overflow by specifying the resolution to use for INT96 values
     tab_correct = get_table(
--- a/python/pyarrow/tests/parquet/test_datetime.py
+++ b/python/pyarrow/tests/parquet/test_datetime.py
@@ -297,8 +297,6 @@ def test_coerce_int96_timestamp_unit(unit):
 
 @pytest.mark.pandas
 @pytest.mark.parametrize('pq_reader_method', ['ParquetFile', 'read_table'])
-@pytest.mark.filterwarnings(
-    "ignore:Discarding nonzero nanoseconds in conversion:UserWarning")
 def test_coerce_int96_timestamp_overflow(pq_reader_method, tempdir):
 
     def get_table(pq_reader_method, filename, **kwargs):
--- a/python/pyarrow/tests/parquet/test_datetime.py
+++ b/python/pyarrow/tests/parquet/test_datetime.py
@@ -297,6 +297,8 @@ def test_coerce_int96_timestamp_unit(unit):
 
 @pytest.mark.pandas
 @pytest.mark.parametrize('pq_reader_method', ['ParquetFile', 'read_table'])
+@pytest.mark.filterwarnings(
+    "ignore:Discarding nonzero nanoseconds in conversion:UserWarning")
 def test_coerce_int96_timestamp_overflow(pq_reader_method, tempdir):
 
     def get_table(pq_reader_method, filename, **kwargs):
--- a/python/pyarrow/tests/test_pandas.py
+++ b/python/pyarrow/tests/test_pandas.py
@@ -96,9 +96,13 @@ def _check_pandas_roundtrip(df, expected=None, use_threads=False,
     if expected is None:
         expected = df
 
-    tm.assert_frame_equal(result, expected, check_dtype=check_dtype,
-                          check_index_type=('equiv' if preserve_index
-                                            else False))
+    # pandas.testing generates a
+    # DeprecationWarning: elementwise comparison failed
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", DeprecationWarning)
+        tm.assert_frame_equal(result, expected, check_dtype=check_dtype,
+                              check_index_type=('equiv' if preserve_index
+                                                else False))
 
 
 def _check_series_roundtrip(s, type_=None, expected_pa_type=None):
@@ -2112,7 +2116,12 @@ class TestConvertListTypes:
         ])
 
         series = pd.Series(data.to_pandas())
-        tm.assert_series_equal(series, expected)
+
+        # pandas.testing generates a
+        # DeprecationWarning: elementwise comparison failed
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", DeprecationWarning)
+            tm.assert_series_equal(series, expected)
 
     @pytest.mark.parametrize('t,data,expected', [
         (
@@ -2167,10 +2176,12 @@ class TestConvertListTypes:
              .to_pandas())
 
         # pandas.testing generates a
+        # DeprecationWarning: elementwise comparison failed
         # numpy.VisibleDeprecationWarning: Creating an ndarray
         #     from ragged nested sequences ...
         with warnings.catch_warnings():
             warnings.simplefilter("ignore", np.VisibleDeprecationWarning)
+            warnings.simplefilter("ignore", DeprecationWarning)
             tm.assert_series_equal(
                 s, pd.Series([[[1, 2, 3], [4]], None], dtype=object),
                 check_names=False)
--- a/python/pyarrow/tests/test_pandas.py
+++ b/python/pyarrow/tests/test_pandas.py
@@ -2165,9 +2165,15 @@ class TestConvertListTypes:
         s = (pa.array([[[1, 2, 3], [4]], None],
                       type=pa.large_list(pa.large_list(pa.int64())))
              .to_pandas())
-        tm.assert_series_equal(
-            s, pd.Series([[[1, 2, 3], [4]], None], dtype=object),
-            check_names=False)
+
+        # pandas.testing generates a
+        # numpy.VisibleDeprecationWarning: Creating an ndarray
+        #     from ragged nested sequences ...
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", np.VisibleDeprecationWarning)
+            tm.assert_series_equal(
+                s, pd.Series([[[1, 2, 3], [4]], None], dtype=object),
+                check_names=False)
 
     def test_large_binary_list(self):
         for list_type_factory in (pa.list_, pa.large_list):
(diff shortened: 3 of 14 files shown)


milesgranger:
This appears to remove all errors in the CI [without](https://github.com/apache/arrow/actions/runs/3547620857/jobs/5957939770#step:6:5553) and [with pandas](https://github.com/apache/arrow/actions/runs/3547620857/jobs/5957939831#step:6:5696)

I assume we don't want to turn warnings into errors? Maybe that's too frustrating since unrelated warnings can always pop up in subsequent PRs?

pitrou:
> I assume we don't want to turn warnings into errors? Maybe that's too frustrating since unrelated warnings can always pop up in subsequent PRs?

Yes, it would definitely be frustrating.

AlenkaF (approved):
Went through the changes and the CI, looks good to me 👍

python/pyarrow/conftest.py
```diff
@@ -265,3 +267,19 @@ def add_fs(doctest_namespace, request, tmp_path):
...
+    """
+    There are occasions when a PytestUnraisableExceptionWarning is
+    raised from pyx files or other __del__ impls that doesn't properly
+    call the initial caller's .close when creating sockets. Here we'll
+    ensure sockets are closed.
+    """
+    class socket_(socket.socket):
+        def __del__(self):
+            if not self._closed:
+                self.close()
```
pitrou, line 283: We definitely don't want to monkeypatch a core type just to avoid some warnings.

If some of our routines fail to close a socket we should try to improve them. Otherwise we should just live with the warnings.

python/pyarrow/parquet/core.py
```diff
@@ -1165,8 +1165,10 @@ def get_metadata(self):
         -------
         metadata : FileMetaData
         """
-        with self.open() as parquet:
-            return parquet.metadata
+        parquet = self.open()
+        meta = parquet.metadata
+        parquet.close(True)
```
pitrou, line 1170: Hmm, isn't it a bug if a `with` statement fails to close the file it just opened?

pitrou:
(sorry, some of those comments were actually written days ago, but I had forgotten to submit the review...)

python/pyarrow/tests/test_pandas.py
```diff
@@ -2164,9 +2172,17 @@ def test_nested_large_list(self):
...
                       type=pa.large_list(pa.large_list(pa.int64())))
              .to_pandas())
-        tm.assert_series_equal(
-            s, pd.Series([[[1, 2, 3], [4]], None], dtype=object),
-            check_names=False)
+
+        # pandas.testing generates a
+        # DeprecationWarning: elementwise comparison failed
+        # numpy.VisibleDeprecationWarning: Creating an ndarray
+        #     from ragged nested sequences ...
```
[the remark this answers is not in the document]
milesgranger: These warnings are happening inside of the `pandas.testing` module though.  Unless I'm misunderstanding.. should we open an issue with pandas then?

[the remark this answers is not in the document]
pitrou: Ah! Yes, perhaps open an issue with them, in case there's a workaround?

python/pyarrow/conftest.py
milesgranger: Yeah, this felt dirty. Commit has been dropped, I'll poke around at it a bit more.

python/pyarrow/parquet/core.py
milesgranger: It's slightly awkward since ParquetFile distinguishes between opening and closing based on if it was the one who opened it. Indeed, `open()` opens the source for `ParquetFile` and thus should probably mark it as being responsible for closing. :+1: https://github.com/apache/arrow/pull/14729/commits/e845fc85c99ed3e484a320b48234c0dcf861a19e

python/pyarrow/tests/test_pandas.py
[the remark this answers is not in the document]
milesgranger: Hmm, actually, when making an example for the issue to report, I _think_ it maybe how our implementation of converting a variable list array to pandas:

```python
import pyarrow as pa
import pandas as pd
import pandas.testing as tm

s = (pa.array([[[1, 2, 3], [4]], None],
                      type=pa.large_list(pa.large_list(pa.int64())))
             .to_pandas())

# This will raise the warning
tm.assert_series_equal(
     s, pd.Series([[[1, 2, 3], [4]], None], dtype=object),
     check_names=False)

# but so will this:
s.equals(pd.Series([[[1, 2, 3], [4]], None], dtype=object))

# this will not
tm.assert_series_equal(
     pd.Series([[[1, 2, 3], [4]], None], dtype=object),
     pd.Series([[[1, 2, 3], [4]], None], dtype=object),
     check_names=False)
```

Therefore I think a new issue to be made to correct this on our end after all. :thinking:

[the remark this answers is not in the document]
milesgranger: For now, I've made issues 
- https://github.com/apache/arrow/issues/14759 
- https://github.com/apache/arrow/issues/14760. 

I can create separate PRs for them I think if that's agreeable.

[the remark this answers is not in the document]
pitrou: Yes, separate PRs sound fine.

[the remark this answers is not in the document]
jorisvandenbossche: See my answer in https://github.com/apache/arrow/issues/14759, for the warning coming from pandas' testing, I think we can ignore that warning on our side (and maybe report it to pandas)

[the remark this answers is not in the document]
milesgranger: Thanks @jorisvandenbossche! Okay, I'll add back the ignoring of this error then here, and also removed saying this closed [ARROW-17651](https://issues.apache.org/jira/browse/ARROW-17651)

python/pyarrow/tests/parquet/test_datetime.py
```diff
@@ -297,6 +297,8 @@ def test_coerce_int96_timestamp_unit(unit):
 
 @pytest.mark.pandas
 @pytest.mark.parametrize('pq_reader_method', ['ParquetFile', 'read_table'])
+@pytest.mark.filterwarnings(
+    "ignore:Discarding nonzero nanoseconds in conversion:UserWarning")
```
jorisvandenbossche, line 301: Where is this warning coming from?

python/pyarrow/tests/test_pandas.py
```diff
@@ -95,9 +96,13 @@ def _check_pandas_roundtrip(df, expected=None, use_threads=False,
...
-    tm.assert_frame_equal(result, expected, check_dtype=check_dtype,
-                          check_index_type=('equiv' if preserve_index
-                                            else False))
+    # pandas.testing generates a
+    # DeprecationWarning: elementwise comparison failed
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", DeprecationWarning)
+        tm.assert_frame_equal(result, expected, check_dtype=check_dtype,
+                              check_index_type=('equiv' if preserve_index
+                                                else False))
```
jorisvandenbossche, line 105: Could we be more specific in which warning we are catching? (so are not catching any future deprecation warning, but only specifically the current numpy one)

(for example with `filterwarnings()` instead of `simplefilter` where we can also specify the module / message, similarly to the `pytest.mark.filterwarnings`. Or just use that mark)

python/pyarrow/tests/test_pandas.py
```diff
@@ -182,6 +187,8 @@ def test_column_index_names_are_preserved(self):
         df.columns.names = ['a']
         _check_pandas_roundtrip(df, preserve_index=True)
 
+    @pytest.mark.filterwarnings(
+        "ignore:parsing timezone aware datetimes:DeprecationWarning")
```
jorisvandenbossche, line 191: Do you remember for which pandas version this happened?

milesgranger: I don't I was trying locally between a couple Python version environments and I too cannot get it to recreate.. so trying in CI now https://github.com/apache/arrow/pull/14729/commits/a599d65fd09579c5d1a4195a7f41f356b958d544 (same for comment about 'Discarding nonzero nanoseconds')

python/pyarrow/tests/test_substrait.py
```diff
@@ -223,7 +224,11 @@ def table_provider(names):
     assert res_tb == test_table_1
 
 
+@pytest.mark.filterwarnings("ignore::pytest.PytestUnraisableExceptionWarning")
 def test_named_table_invalid_table_name():
+    # Warning generated from pyx which is unable to be propagated to caller
+    # ref: https://docs.pytest.org/en/6.2.x/usage.html#warning-about
+    #      -unraisable-exceptions-and-unhandled-thread-exceptions
```
jorisvandenbossche, line 231: Is this something we should actually try to fix?

python/pyarrow/tests/test_pandas.py
jorisvandenbossche: Yeah, it might be that bumping the minimum pandas version to 1.0 in the meantime has made that disappear already

python/pyarrow/tests/parquet/test_datetime.py
milesgranger: Haivng a hard time getting it to show up agaiin locally, but it's iin CI: https://github.com/milesgranger/arrow/actions/runs/3573348441/jobs/6007303891#step:6:5679

I _think_ it's coming from `tab_correct.to_pandas` call. But not sure, will try more to get it to happen locally.

python/pyarrow/tests/test_substrait.py
jorisvandenbossche: The ignored exception is:

```
  File "/home/joris/scipy/repos/arrow/python/pyarrow/tests/test_substrait.py", line 238, in table_provider
    raise Exception("Unrecognized table name")
Exception: Unrecognized table name
```

which is IMO something we should fix instead (the error should be raised to the user)

python/pyarrow/tests/test_pandas.py
milesgranger: Okay, well it's not in CI that I can see.. not sure how I managed to get that. will remove the filter.

python/pyarrow/tests/test_substrait.py
milesgranger: That seems prudent. I'll make an issue for it separately, and have removed this filter in https://github.com/apache/arrow/pull/14729/commits/20b1a30622ecf722783c1f0485d5f37c110a8deb

python/pyarrow/tests/test_pandas.py
```diff
@@ -2164,9 +2172,17 @@ def test_nested_large_list(self):
...
-        tm.assert_series_equal(
-            s, pd.Series([[[1, 2, 3], [4]], None], dtype=object),
-            check_names=False)
+
+        # pandas.testing generates a
+        # DeprecationWarning: elementwise comparison failed
+        # numpy.VisibleDeprecationWarning: Creating an ndarray
+        #     from ragged nested sequences ...
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", np.VisibleDeprecationWarning)
```
milesgranger, line 2181: Given [our discussion about numpy 1.24](https://github.com/apache/arrow/issues/14759#issuecomment-1330436673) should I remove this warning since it'll be going away?

jorisvandenbossche: I would still catch/ignore it for now, since we will have test builds for some time that have older numpy

python/pyarrow/tests/parquet/test_datetime.py
jorisvandenbossche: Searching for this phrase, pandas raises this warning when you convert a Timestamp object to a python datetime.datetime object:

```
In [22]: pd.Timestamp("2022-01-01 01:01:01.123456789").to_pydatetime()
<ipython-input-22-7cef94b12f00>:1: UserWarning: Discarding nonzero nanoseconds in conversion.
  pd.Timestamp("2022-01-01 01:01:01.123456789").to_pydatetime()
Out[22]: datetime.datetime(2022, 1, 1, 1, 1, 1, 123456)
```

jorisvandenbossche: But I don't directly see how this test is causing such a warning, since it uses dates that doesn't include nanoseconds

jorisvandenbossche: Ah, or maybe from this line:

```
    # with the default resolution of ns, we get wrong values for INT96
    # that are out of bounds for nanosecond range
    tab_error = get_table(pq_reader_method, filename)
    assert tab_error["a"].to_pylist() != oob_dts
```

since you get "wrong" values, those have overflown, and the resulting wrong datetime will probably have nanoseconds, and thus can raise this warning. That should be OK to ignore then.

python/pyarrow/tests/test_substrait.py
milesgranger: https://github.com/apache/arrow/issues/14766

python/pyarrow/tests/test_pandas.py
milesgranger: Good point. Done in https://github.com/apache/arrow/pull/14729/commits/ee97c6bdc475b6e8e78c827cfea278272189f5c8

python/pyarrow/tests/parquet/test_datetime.py
milesgranger: Should I move to a `warnings.catch` over this line specifically then to make it more clear where the warning is being raised?

jorisvandenbossche: > Should I move to a `warnings.catch` over this line specifically then to make it more clear where the warning is being raised?

That's probably most specific indeed (and also documents the root cause as well)

milesgranger: https://github.com/apache/arrow/pull/14729/commits/ab4a31a4a1dbca63f8e35eb3553f8b43f9d0f71a

jorisvandenbossche (approved):

jorisvandenbossche:
Can you merge master to get rid of the failure?

jorisvandenbossche:
Actually, it's only the label workflow (the linter passed), so don't worry

PR #14733, opened 2022-11-25
Update message.cpp

saurabhacellere:


nealrichardson:
This doesn't seem useful as is. Is this PR incomplete?

PR #14735, opened 2022-11-25
ARROW-18406: [C++] Can't build Arrow with Substrait on Ubuntu 20.04

westonpace:


westonpace:
This is something I think we will eventually want a good way to regress.  Perhaps a nightly job for Ubuntu 20.04 that uses the system libraries instead of conda?  I say this because I think this is the second or third time a regression of this style has been introduced.

westonpace:
Also, to be clear, the change here was not actually a code change.  We always had `has_enum_()`.  However, in older versions of protoc it is only valid to call `has_xyz()` if `xyz` is a message.  If `xyz` is an enum or other simple type (in this case a string) then you have to instead check against the default value for that type (because protobuf, in these cases, does not distinguish between the two)

PR #14739, opened 2022-11-25
ARROW-18409: [GLib][Plasma] Suppress deprecated warning in building plasma-glib

kou:
If we always get "Plasma is deprecated since Arrow 10.0.0. ..." warning from plasma/common.h, we can't use -Dwerror=true Meson option with plama-glib.


kou:
+1

PR #14741, opened 2022-11-26
ARROW-18106: [C++] JSON reader ignores explicit schema with default unexpected_field_behavior="infer"

benibus:
See: [ARROW-18106](https://issues.apache.org/jira/browse/ARROW-18106)

The current `ChunkedArrayBuilder` always uses the promotion graph for type conversions if it's provided (even for fields in the explicit schema). The bug in question occurs because conversion errors aren't propagated in `InferringChunkedArrayBuilder`. Instead, it will attempt to promote the type on failure, which is undesirable in cases like this.

This PR doesn't change any of that. It only restricts type inference to unexpected fields while fields present in the schema remain "strongly-typed" - so conversion errors for expected fields will still be reported if `UnexpectedFieldBehavior::InferType` is specified.

### DIFF
--- a/cpp/src/arrow/json/chunked_builder.cc
+++ b/cpp/src/arrow/json/chunked_builder.cc
@@ -37,6 +37,7 @@ using internal::checked_cast;
 using internal::TaskGroup;
 
 namespace json {
+namespace {
 
 Status MakeChunkedArrayBuilder(const std::shared_ptr<TaskGroup>& task_group,
                                MemoryPool* pool, const PromotionGraph* promotion_graph,
@@ -484,6 +485,11 @@ Status MakeChunkedArrayBuilder(const std::shared_ptr<TaskGroup>& task_group,
   return Status::OK();
 }
 
+}  // namespace
+
+// This overload is exposed to the user and will only be called once on instantiation to
+// canonicalize any explicitly-defined fields. Such fields won't be subject to
+// type inference/promotion
 Status MakeChunkedArrayBuilder(const std::shared_ptr<TaskGroup>& task_group,
                                MemoryPool* pool, const PromotionGraph* promotion_graph,
                                const std::shared_ptr<DataType>& type,
--- a/cpp/src/arrow/json/chunked_builder.cc
+++ b/cpp/src/arrow/json/chunked_builder.cc
@@ -38,6 +38,12 @@ using internal::TaskGroup;
 
 namespace json {
 
+Status MakeChunkedArrayBuilder(const std::shared_ptr<TaskGroup>& task_group,
+                               MemoryPool* pool, const PromotionGraph* promotion_graph,
+                               const std::shared_ptr<DataType>& type,
+                               bool allow_promotion,
+                               std::shared_ptr<ChunkedArrayBuilder>* out);
+
 class NonNestedChunkedArrayBuilder : public ChunkedArrayBuilder {
  public:
   NonNestedChunkedArrayBuilder(const std::shared_ptr<TaskGroup>& task_group,
@@ -404,7 +410,7 @@ class ChunkedStructArrayBuilder : public ChunkedArrayBuilder {
 
         std::shared_ptr<ChunkedArrayBuilder> child_builder;
         RETURN_NOT_OK(MakeChunkedArrayBuilder(task_group_, pool_, promotion_graph_, type,
-                                              &child_builder));
+                                              /*allow_promotion=*/true, &child_builder));
         child_builders_.emplace_back(std::move(child_builder));
       }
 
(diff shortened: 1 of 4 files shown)


cpp/src/arrow/json/reader_test.cc
```diff
@@ -305,5 +305,117 @@ TEST(ReaderTest, ListArrayWithFewValues) {
...
+    {"a": {"b": null, "c": {"d": [{"e": true}]}}}
+  ])");
+  ASSERT_OK_AND_ASSIGN(auto expected_table, Table::FromRecordBatches({expected_batch}));
+
+  std::shared_ptr<io::InputStream> input;
+  std::shared_ptr<TableReader> reader;
+  ASSERT_OK(MakeStream(json, &input));
+  ASSERT_OK_AND_ASSIGN(reader, TableReader::Make(default_memory_pool(), input,
+                                                 read_options, parse_options));
+  ASSERT_OK_AND_ASSIGN(auto actual_table, reader->Read());
```
pitrou, line 370: This snippet occurs quite a few times, factor it out in a test fixture perhaps?

cpp/src/arrow/json/reader_test.cc
```diff
@@ -305,5 +305,117 @@ TEST(ReaderTest, ListArrayWithFewValues) {
...
+  AssertTablesEqual(*actual_table, *expected_table);
+
+  json += std::string(R"({"a": {"b": "2022-09-05T08:08:46.000"}})") + "\n";
+  ASSERT_OK(MakeStream(json, &input));
+  ASSERT_OK_AND_ASSIGN(reader, TableReader::Make(default_memory_pool(), input,
+                                                 read_options, parse_options));
+  ASSERT_RAISES(Invalid, reader->Read());
```
pitrou, line 377: Can we also test something about the error message (to make sure we aren't triggering some other condition than expected)?

cpp/src/arrow/json/reader_test.cc
```diff
@@ -305,5 +305,117 @@ TEST(ReaderTest, ListArrayWithFewValues) {
...
+  ASSERT_OK_AND_ASSIGN(reader, TableReader::Make(default_memory_pool(), input,
+                                                 read_options, parse_options));
+  ASSERT_OK_AND_ASSIGN(auto actual_table, reader->Read());
+  AssertTablesEqual(*actual_table, *expected_table);
+
+  json += std::string(R"({"a": {"b": "2022-09-05T08:08:03.000", "c": {}}})") + "\n";
+  ASSERT_OK(MakeStream(json, &input));
+  ASSERT_OK_AND_ASSIGN(reader, TableReader::Make(default_memory_pool(), input,
+                                                 read_options, parse_options));
+  ASSERT_RAISES(Invalid, reader->Read());
```
pitrou, line 417: Same here.

cpp/src/arrow/json/chunked_builder.cc
```diff
@@ -466,5 +484,13 @@ Status MakeChunkedArrayBuilder(const std::shared_ptr<TaskGroup>& task_group,
   return Status::OK();
 }
 
+Status MakeChunkedArrayBuilder(const std::shared_ptr<TaskGroup>& task_group,
```
pitrou, line 487: IIUC, this overload is only called for top-level fields of an explicit schema? Add a comment explaining this?

pitrou (commented):
Thanks @benibus . This seems ok on the principle, some suggestions to improve readability.

pitrou:
@benibus This needs rebasing/merging and then fixing conflicts, and also ideally addressing review comments.

cpp/src/arrow/json/reader_test.cc
```diff
@@ -325,6 +335,108 @@ TEST(ReaderTest, FailOnInvalidEOF) {
...
+  parse_options.explicit_schema = schema({field("t", timestamp(TimeUnit::SECOND))});
+
+  std::shared_ptr<io::InputStream> input;
+  std::shared_ptr<TableReader> reader;
+  for (auto behavior : {UnexpectedFieldBehavior::Error, UnexpectedFieldBehavior::Ignore,
+                        UnexpectedFieldBehavior::InferType}) {
+    parse_options.unexpected_field_behavior = behavior;
+    EXPECT_RAISES_WITH_MESSAGE_THAT(
+        Invalid,
+        ::testing::StartsWith("Invalid: Failed of conversion of JSON to timestamp[s]"),
```
pitrou, line 354: I hadn't noticed this before (it predates your PR), but is "Failed of conversion" grammatically correct?

benibus: No, it isn't. I can go ahead and fix it - I think a couple other tests use the message too.

pitrou (approved):

PR #14742, opened 2022-11-27
ARROW-18413: [C++][Parquet] Expose page index info from ColumnChunkMetaData

wgtmac:
This is the first step to support page index of parquet.

### DIFF
--- a/cpp/src/parquet/metadata.cc
+++ b/cpp/src/parquet/metadata.cc
@@ -314,16 +314,14 @@ class ColumnChunkMetaData::ColumnChunkMetaDataImpl {
 
   inline std::optional<IndexLocation> GetColumIndexLocation() const {
     if (column_->__isset.column_index_offset && column_->__isset.column_index_length) {
-      return IndexLocation{.index_file_offset_bytes = column_->column_index_offset,
-                           .offset_index_length = column_->column_index_length};
+      return IndexLocation{column_->column_index_offset, column_->column_index_length};
     }
     return std::nullopt;
   }
 
   inline std::optional<IndexLocation> GetOffsetIndexLocation() const {
     if (column_->__isset.offset_index_offset && column_->__isset.offset_index_length) {
-      return IndexLocation{.index_file_offset_bytes = column_->offset_index_offset,
-                           .offset_index_length = column_->offset_index_length};
+      return IndexLocation{column_->offset_index_offset, column_->offset_index_length};
     }
     return std::nullopt;
   }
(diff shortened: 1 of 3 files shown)


wgtmac:
@pitrou @emkornfield Can you please take a look? As I have mentioned in other thread, I will work on the page index and break it down into small commits to make it review-friendly.

cpp/src/parquet/metadata.h
```diff
@@ -171,6 +171,13 @@ class PARQUET_EXPORT ColumnChunkMetaData {
   int64_t total_uncompressed_size() const;
   std::unique_ptr<ColumnCryptoMetaData> crypto_metadata() const;
 
+  bool has_column_index() const;
+  int64_t column_index_offset() const;
```
emkornfield, line 175: I think it might be a clearer contract to have something like:

```
struct IndexLocation {
   int64_t index_file_offset_bytes;
   int32_t offset_index_length
}

optional<IndexLocation> GetColumIndexLocation();
optional<IndexLocation> GetOffsetIndexLocation();
```

Thoughts?

wgtmac: SGTM, I will change it shortly.

wgtmac:
I have addressed your comment, and the unsuccessful CI checks are unrelated to my change. Can you please take a look again? @emkornfield

cpp/src/parquet/metadata.cc
```diff
@@ -312,6 +312,20 @@ class ColumnChunkMetaData::ColumnChunkMetaDataImpl {
     }
   }
 
+  inline std::optional<IndexLocation> GetColumIndexLocation() const {
```
pitrou, line 315: `inline` is implied if the method is defined inside the `class` body, so it's superfluous here.

cpp/src/parquet/metadata.h
```diff
@@ -118,6 +119,12 @@ struct PageEncodingStats {
   int32_t count;
 };
 
+/// \brief Public struct for location to page index in ColumnChunkMetaData.
+struct IndexLocation {
+  int64_t index_file_offset_bytes;
+  int32_t offset_index_length;
```
pitrou, line 125: I'm not sure why the complex naming, we could just have
```suggestion
  /// File offset of the given index, in bytes
  int64_t offset;
  /// Length of the given index, in bytes
  int32_t length;
```

pitrou (changes requested):
Thank you. Is it possible to add basic tests for this?

cpp/src/parquet/metadata.cc
wgtmac: Fixed

cpp/src/parquet/metadata.h
wgtmac: Fixed

pitrou (approved):
Thanks for the update @wgtmac !

pitrou:
Test failures are unrelated, will merge.

PR #14743, opened 2022-11-27
replace sample_frac with slice_sample

davechilders:
`dplyr::sample_frac()` has been superseded by `dplyr::slice_sample()`

### DIFF
--- a/r/vignettes/dataset.Rmd
+++ b/r/vignettes/dataset.Rmd
@@ -313,7 +313,7 @@ percentage of rows from each batch:
 sampled_data <- ds %>%
   filter(year == 2015) %>%
   select(tip_amount, total_amount, passenger_count) %>%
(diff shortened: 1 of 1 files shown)


nealrichardson:
Thanks for catching this. Interestingly, since this vignette was originally written, `slice_sample()` is now supported (somewhat) on Arrow queries, so you wouldn't have to drop down to use a UDF to perform it. So we might want a better example altogether of this feature. 

Thoughts @djnavarro @wjones127 @thisisnic?

wjones127:
If we can use `slice_sample()` directly on Arrow queries, the example is still useful since we use it later to call `predict()` on each batch with the model we fit from the sample. I'd be +1 on changing to use `slice_sample()` directly on the Arrow data and keeping the rest of the example as is.

amol-:

PR #14749, opened 2022-11-28
ARROW-18280: [C++][Python] Support slicing to end in list_slice kernel

milesgranger:
Will close [ARROW-18280](https://issues.apache.org/jira/browse/ARROW-18280)


cpp/src/arrow/compute/kernels/scalar_nested_test.cc
```diff
@@ -168,6 +173,20 @@ TEST(TestScalarNested, ListSliceFixedOutput) {
...
+                               "[[2, 3], [5, null], [null, null], null]");
+      if (input->type()->id() == Type::FIXED_SIZE_LIST) {
+        CheckScalarUnary("list_slice", input, expected, &args);
+      } else {
+        EXPECT_RAISES_WITH_MESSAGE_THAT(
+            NotImplemented,
+            ::testing::HasSubstr("Unable to produce FixedSizeListArray from "
+                                 "non-FixedSizeListArray without `stop` being set."),
+            CallFunction("list_slice", {input}, &args));
+      }
```
wjones127, line 189: Is this a case that is possible?

```suggestion
      }
      
      args.start = 3;
      args.stop = std::nullopt;
      expected = ArrayFromJSON(fixed_size_list(value_type, 0),
                               "[[], [], [], null]");
      if (input->type()->id() == Type::FIXED_SIZE_LIST) {
        CheckScalarUnary("list_slice", input, expected, &args);
      }
```

wjones127 (commented):
Looks pretty good. I have a question about an edge case.

cpp/src/arrow/compute/kernels/scalar_nested.cc
```diff
@@ -131,13 +124,26 @@ struct ListSlice {
...
-          &builder));
+      const auto stop = [&]() {
+        if (opts.stop.has_value()) {
+          return static_cast<int32_t>(opts.stop.value());
+        } else {
+          DCHECK_EQ(list_type->id(), arrow::Type::FIXED_SIZE_LIST);
+          return checked_cast<const FixedSizeListType*>(list_type)->list_size();
+        }
+      }();
+      const auto size = std::max(stop - static_cast<int32_t>(opts.start), 0);
```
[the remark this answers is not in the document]
milesgranger: This covers the case of your suggested case where `stop` is null and the fixed size list stop is then set to the list_size, which could be less than the start.

cpp/src/arrow/compute/kernels/scalar_nested_test.cc
milesgranger: That can indeed be possible, and a nice extra test case. Thank you!

cpp/src/arrow/compute/kernels/scalar_nested.cc
```diff
@@ -131,13 +124,26 @@ struct ListSlice {
...
-                          static_cast<int32_t>(opts.stop.value() - opts.start)),
-          &builder));
+      const auto stop = [&]() {
+        if (opts.stop.has_value()) {
+          return static_cast<int32_t>(opts.stop.value());
+        } else {
+          DCHECK_EQ(list_type->id(), arrow::Type::FIXED_SIZE_LIST);
+          return reinterpret_cast<const FixedSizeListType*>(list_type)->list_size();
+        }
+      }();
```
jorisvandenbossche, line 143: Code question: is there a reason this is done as a lambda, instead of just a if/else block?

milesgranger: ...because one can declare `stop` as a `const` for the cost of one extra line? haha. 

It actually spawned from my trying to satisfy MinGW thnking I was illegally casting a different ListType than `FixedSizeList` in the second arm.. However, in retrospect I think that was because I was using the ternary operator `condition ? expr : expr` which MinGW apparently would type check both expressions.

Anyway, I can remove the lambda if that's preferred.

jorisvandenbossche (approved):
Looks good to me!

PR #14756, opened 2022-11-28
MINOR: [Doc][Python] pull the test data in one command

haim0n:


jorisvandenbossche:
@haim0n Thanks for the contribution. I am wondering though, is there a specific reason for this? We have other code blocks on this page that also consist of multiple commands, and where we also don't do this.

haim0n:
@jorisvandenbossche , you're right. I walked the path of the newbie setting up the dev env for the first time. So the one liner was a speedup for me (replaces 3 copy/paste actions) and allows to skip this mechanical step. 

Some of the other commands contain comments between them and provide some learning curve to new comers- so my guess they're nice to have and probably should be left intact.

jorisvandenbossche:
> I walked the path of the newbie setting up the dev env for the first time. So the one liner was a speedup for me (replaces 3 copy/paste actions) and allows to skip this mechanical step.

That's of course something a newcomer also doesn't necessarily know, but you can actually copy paste those three lines at once (at least _most_ terminals should support executing those one by one if pasting multiple lines)

haim0n:
mmm no - Bash treats `$` as variable and those lines can't be just pasted.

pitrou:
I don't really like this change. It makes the snippet easier to paste, but at the expense of readability.

raulcd:
For completeness, if you are navigating the documentation: https://arrow.apache.org/docs/developers/python.html#environment-setup-and-build
and try to copy & paste or use the sphinx copy to clipboard button it already skips the `$`. This is what my clipboard copies:
```
pushd arrow
git submodule update --init
export PARQUET_TEST_DATA="${PWD}/cpp/submodules/parquet-testing/data"
export ARROW_TEST_DATA="${PWD}/testing/data"
popd
```
If you try to copy directly from the source rst file, yes, you have to copy&paste for each command because the `$` gets copied to the clipboard but that should be ok.

amol-:

PR #14758, opened 2022-11-29
GH-33317: [C++] Utility method to ensure an array object meetings an alignment requirement

### LINKED ISSUE #33317: [C++] Utility method to ensure an array object meetings an alignment requirement
This would look something like:

EnsureAligned(Buffer|Array|ChunkedArray|RecordBatch|Table, int minimum_alignment, MemoryPool\* memory_pool);

It would fail if MemoryPool's alignment < minimum_alignment
It would iterate through each buffer of the object, if the object is not aligned properly, it would reallocate and copy the buffer (using memory_pool)

It would return a new object where every buffer is guaranteed to meet the alignment requirements.

**Reporter**: [Weston Pace](https://issues.apache.org/jira/browse/ARROW-18119) / @westonpace
**Assignee**: [Sanjiban Sengupta](https://issues.apache.org/jira/browse/ARROW-18119) / @sanjibansg
#### PRs and other links:
- [GitHub Pull Request #14758](https://github.com/apache/arrow/pull/14758)

<sub>**Note**: *This issue was originally created as [ARROW-18119](https://issues.apache.org/jira/browse/ARROW-18119). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

sanjibansg:
This PR adds a utility function which is responsible for ensuring that all the buffers of an arrow object are properly aligned. It checks all the buffers in an arrow object for alignment, and if not aligned properly, then allocates a buffer by specifying the required alignment and copies data from the previous buffer.
* Closes: #33317

### DIFF
--- a/cpp/src/arrow/util/align_util.cc
+++ b/cpp/src/arrow/util/align_util.cc
@@ -26,56 +26,63 @@ namespace arrow {
 
 namespace util {
 
-Result<std::shared_ptr<Array>> EnsureAlignment(const Array& object, int64_t alignment,
+Result<std::shared_ptr<Buffer>> EnsureAlignment(const std::shared_ptr<Buffer>& object, int64_t alignment,
                                                MemoryPool* memory_pool) {
-  std::vector<std::shared_ptr<Buffer>> buffers_ = object.data()->buffers;
+    auto buffer_address = object->address();
+    if ((buffer_address % alignment) != 0) {
+        ARROW_ASSIGN_OR_RAISE(
+                auto new_buffer, AllocateBuffer(object->size(), alignment, memory_pool));
+        std::memcpy(new_buffer->mutable_data(), object->data(), object->size());
+        return new_buffer;
+    } else {
+        return object;
+    }
+}
+
+Result<std::shared_ptr<Array>> EnsureAlignment(const std::shared_ptr<Array>& object, int64_t alignment,
+                                               MemoryPool* memory_pool) {
+  std::vector<std::shared_ptr<Buffer>> buffers_ = object->data()->buffers;
   for (size_t i = 0; i < buffers_.size(); ++i) {
     if (buffers_[i]) {
-      auto buffer_address = buffers_[i]->address();
-      if ((buffer_address % alignment) != 0) {
-        ARROW_ASSIGN_OR_RAISE(
-            auto new_buffer, AllocateBuffer(buffers_[i]->size(), alignment, memory_pool));
-        std::memcpy(new_buffer->mutable_data(), buffers_[i]->data(), buffers_[i]->size());
-        buffers_[i] = std::move(new_buffer);
+        ARROW_ASSIGN_OR_RAISE(buffers_[i], EnsureAlignment(buffers_[i], alignment, memory_pool));
       }
-    }
   }
   auto new_array_data =
-      ArrayData::Make(object.data()->type, object.data()->length, std::move(buffers_),
-                      object.data()->GetNullCount(), object.data()->offset);
+      ArrayData::Make(object->data()->type, object->data()->length, std::move(buffers_),
+                      object->data()->GetNullCount(), object->data()->offset);
   return MakeArray(new_array_data);
 }
 
-Result<std::shared_ptr<ChunkedArray>> EnsureAlignment(const ChunkedArray& object,
+Result<std::shared_ptr<ChunkedArray>> EnsureAlignment(const std::shared_ptr<ChunkedArray>& object,
                                                       int64_t alignment,
                                                       MemoryPool* memory_pool) {
-  ArrayVector chunks_ = object.chunks();
-  for (int i = 0; i < object.num_chunks(); ++i) {
+  ArrayVector chunks_ = object->chunks();
+  for (int i = 0; i < object->num_chunks(); ++i) {
     ARROW_ASSIGN_OR_RAISE(chunks_[i],
-                          EnsureAlignment(*object.chunk(i), alignment, memory_pool));
+                          EnsureAlignment(object->chunk(i), alignment, memory_pool));
   }
-  return ChunkedArray::Make(std::move(chunks_), object.type());
+  return ChunkedArray::Make(std::move(chunks_), object->type());
 }
 
-Result<std::shared_ptr<RecordBatch>> EnsureAlignment(const RecordBatch& object,
+Result<std::shared_ptr<RecordBatch>> EnsureAlignment(const std::shared_ptr<RecordBatch>& object,
                                                      int64_t alignment,
                                                      MemoryPool* memory_pool) {
-  ArrayVector columns_ = object.columns();
-  for (int i = 0; i < object.num_columns(); ++i) {
+  ArrayVector columns_ = object->columns();
+  for (int i = 0; i < object->num_columns(); ++i) {
     ARROW_ASSIGN_OR_RAISE(columns_[i],
-                          EnsureAlignment(*object.column(i), alignment, memory_pool));
+                          EnsureAlignment(object->column(i), alignment, memory_pool));
   }
-  return RecordBatch::Make(object.schema(), object.num_rows(), std::move(columns_));
+  return RecordBatch::Make(object->schema(), object->num_rows(), std::move(columns_));
 }
 
-Result<std::shared_ptr<Table>> EnsureAlignment(const Table& object, int64_t alignment,
+Result<std::shared_ptr<Table>> EnsureAlignment(const std::shared_ptr<Table>& object, int64_t alignment,
                                                MemoryPool* memory_pool) {
-  std::vector<std::shared_ptr<ChunkedArray>> columns_ = object.columns();
-  for (int i = 0; i < object.num_columns(); ++i) {
+  std::vector<std::shared_ptr<ChunkedArray>> columns_ = object->columns();
+  for (int i = 0; i < object->num_columns(); ++i) {
     ARROW_ASSIGN_OR_RAISE(columns_[i],
-                          EnsureAlignment(*object.column(i), alignment, memory_pool));
+                          EnsureAlignment(object->column(i), alignment, memory_pool));
   }
-  return Table::Make(object.schema(), std::move(columns_), object.num_rows());
+  return Table::Make(object->schema(), std::move(columns_), object->num_rows());
 }
 
 }  // namespace util
--- a/cpp/src/arrow/util/align_util.cc
+++ b/cpp/src/arrow/util/align_util.cc
@@ -26,63 +26,189 @@ namespace arrow {
 
 namespace util {
 
-Result<std::shared_ptr<Buffer>> EnsureAlignment(const std::shared_ptr<Buffer>& object, int64_t alignment,
-                                               MemoryPool* memory_pool) {
-    auto buffer_address = object->address();
-    if ((buffer_address % alignment) != 0) {
-        ARROW_ASSIGN_OR_RAISE(
-                auto new_buffer, AllocateBuffer(object->size(), alignment, memory_pool));
-        std::memcpy(new_buffer->mutable_data(), object->data(), object->size());
-        return new_buffer;
-    } else {
-        return object;
+bool CheckAlignment(const Buffer& buffer, const int64_t& alignment) {
+  return buffer.address() % alignment == 0;
+}
+
+bool CheckAlignment(const ArrayData& array, const int64_t& alignment) {
+  for (const auto& buffer : array.buffers) {
+    if (buffer) {
+      if (!CheckAlignment(*buffer, alignment)) return false;
     }
+  }
+  return true;
 }
 
-Result<std::shared_ptr<Array>> EnsureAlignment(const std::shared_ptr<Array>& object, int64_t alignment,
-                                               MemoryPool* memory_pool) {
-  std::vector<std::shared_ptr<Buffer>> buffers_ = object->data()->buffers;
-  for (size_t i = 0; i < buffers_.size(); ++i) {
-    if (buffers_[i]) {
-        ARROW_ASSIGN_OR_RAISE(buffers_[i], EnsureAlignment(buffers_[i], alignment, memory_pool));
+bool CheckAlignment(const Array& array, const int64_t& alignment) {
+  bool align_dictionary = true, align_children = true;
+
+  if (array.type()->id() == Type::DICTIONARY) {
+    align_dictionary = CheckAlignment(*array.data()->dictionary, alignment);
+  }
+
+  for (const auto& child : array.data()->child_data) {
+    if (child) {
+      if (!CheckAlignment(*child, alignment)) align_children = false;
+    }
+  }
+  return CheckAlignment(*array.data(), alignment) && align_dictionary && align_children;
+}
+
+bool CheckAlignment(const ChunkedArray& array, const int64_t& alignment,
+                    std::vector<bool>& needs_alignment, const int& offset) {
+  needs_alignment.resize(needs_alignment.size() + array.num_chunks(), false);
+  for (auto i = 0; i < array.num_chunks(); ++i) {
+    if (array.chunk(i) && !CheckAlignment(*array.chunk(i), alignment))
+      needs_alignment[i + offset] = true;
+  }
+  return std::find(needs_alignment.begin() + offset,
+                   needs_alignment.begin() + offset + array.num_chunks(),
+                   true) == needs_alignment.begin() + offset + array.num_chunks();
+}
+
+bool CheckAlignment(const RecordBatch& batch, const int64_t& alignment,
+                    std::vector<bool>& needs_alignment) {
+  needs_alignment.resize(batch.num_columns(), false);
+  for (auto i = 0; i < batch.num_columns(); ++i) {
+    if (batch.column(i) && !CheckAlignment(*batch.column(i), alignment))
+      needs_alignment[i] = true;
+  }
+  return std::find(needs_alignment.begin(), needs_alignment.end(), true) ==
+         needs_alignment.end();
+}
+
+bool CheckAlignment(const Table& table, const int64_t& alignment,
+                    std::vector<bool>& needs_alignment) {
+  needs_alignment.resize(table.num_columns(), false);
+  for (auto i = 1; i <= table.num_columns(); ++i) {
+    if (table.column(i - 1) &&
+        !CheckAlignment(*table.column(i - 1), alignment, needs_alignment,
+                        (i - 1) * (1 + table.column(i - 1)->num_chunks())))
+      needs_alignment[i * table.column(i - 1)->num_chunks() + i - 1] = true;
+  }
+  return std::find(needs_alignment.begin(), needs_alignment.end(), true) ==
+         needs_alignment.end();
+}
+
+Result<std::shared_ptr<Buffer>> EnsureAlignment(std::shared_ptr<Buffer> buffer,
+                                                const int64_t& alignment,
+                                                MemoryPool* memory_pool) {
+  if (!CheckAlignment(*buffer, alignment)) {
+    ARROW_ASSIGN_OR_RAISE(auto new_buffer,
+                          AllocateBuffer(buffer->size(), alignment, memory_pool));
+    std::memcpy(new_buffer->mutable_data(), buffer->data(), buffer->size());
+    return new_buffer;
+  } else {
+    return std::move(buffer);
+  }
+}
+
+Result<std::shared_ptr<ArrayData>> EnsureAlignment(std::shared_ptr<ArrayData> array_data,
+                                                   const int64_t& alignment,
+                                                   MemoryPool* memory_pool) {
+  if (!CheckAlignment(*array_data, alignment)) {
+    std::vector<std::shared_ptr<Buffer>> buffers_ = array_data->buffers;
+    for (size_t i = 0; i < buffers_.size(); ++i) {
+      if (buffers_[i]) {
+        ARROW_ASSIGN_OR_RAISE(
+            buffers_[i], EnsureAlignment(std::move(buffers_[i]), alignment, memory_pool));
       }
+    }
+
+    for (auto& it : array_data->child_data) {
+      ARROW_ASSIGN_OR_RAISE(it, EnsureAlignment(std::move(it), alignment, memory_pool));
+    }
+
+    if (array_data->type->id() == Type::DICTIONARY) {
+      ARROW_ASSIGN_OR_RAISE(
+          array_data->dictionary,
+          EnsureAlignment(std::move(array_data->dictionary), alignment, memory_pool));
+    }
+
+    auto new_array_data = ArrayData::Make(
+        array_data->type, array_data->length, std::move(buffers_), array_data->child_data,
+        array_data->dictionary, array_data->GetNullCount(), array_data->offset);
+    return std::move(new_array_data);
+
+  } else {
+    return std::move(array_data);
   }
-  auto new_array_data =
-      ArrayData::Make(object->data()->type, object->data()->length, std::move(buffers_),
-                      object->data()->GetNullCount(), object->data()->offset);
-  return MakeArray(new_array_data);
 }
 
-Result<std::shared_ptr<ChunkedArray>> EnsureAlignment(const std::shared_ptr<ChunkedArray>& object,
-                                                      int64_t alignment,
-                                                      MemoryPool* memory_pool) {
-  ArrayVector chunks_ = object->chunks();
-  for (int i = 0; i < object->num_chunks(); ++i) {
-    ARROW_ASSIGN_OR_RAISE(chunks_[i],
-                          EnsureAlignment(object->chunk(i), alignment, memory_pool));
-  }
-  return ChunkedArray::Make(std::move(chunks_), object->type());
-}
-
-Result<std::shared_ptr<RecordBatch>> EnsureAlignment(const std::shared_ptr<RecordBatch>& object,
-                                                     int64_t alignment,
-                                                     MemoryPool* memory_pool) {
-  ArrayVector columns_ = object->columns();
-  for (int i = 0; i < object->num_columns(); ++i) {
-    ARROW_ASSIGN_OR_RAISE(columns_[i],
-                          EnsureAlignment(object->column(i), alignment, memory_pool));
-  }
-  return RecordBatch::Make(object->schema(), object->num_rows(), std::move(columns_));
-}
-
-Result<std::shared_ptr<Table>> EnsureAlignment(const std::shared_ptr<Table>& object, int64_t alignment,
+Result<std::shared_ptr<Array>> EnsureAlignment(std::shared_ptr<Array> array,
+                                               const int64_t& alignment,
                                                MemoryPool* memory_pool) {
-  std::vector<std::shared_ptr<ChunkedArray>> columns_ = object->columns();
-  for (int i = 0; i < object->num_columns(); ++i) {
-    ARROW_ASSIGN_OR_RAISE(columns_[i],
-                          EnsureAlignment(object->column(i), alignment, memory_pool));
+  ARROW_ASSIGN_OR_RAISE(auto new_array_data,
+                        EnsureAlignment(array->data(), alignment, memory_pool));
+
+  if (new_array_data.get() == array->data().get()) {
+    return std::move(array);
+  } else {
+    return MakeArray(std::move(new_array_data));
   }
-  return Table::Make(object->schema(), std::move(columns_), object->num_rows());
+}
+
+Result<std::shared_ptr<ChunkedArray>> EnsureAlignment(std::shared_ptr<ChunkedArray> array,
+                                                      const int64_t& alignment,
+                                                      MemoryPool* memory_pool) {
+  std::vector<bool> needs_alignment;
+  if (!CheckAlignment(*array, alignment, needs_alignment)) {
+    ArrayVector chunks_ = array->chunks();
+    for (int i = 0; i < array->num_chunks(); ++i) {
+      if (needs_alignment[i] && chunks_[i]) {
+        ARROW_ASSIGN_OR_RAISE(
+            chunks_[i], EnsureAlignment(std::move(chunks_[i]), alignment, memory_pool));
+      }
+    }
+    return ChunkedArray::Make(std::move(chunks_), array->type());
+  } else {
+    return std::move(array);
+  }
+}
+
+Result<std::shared_ptr<RecordBatch>> EnsureAlignment(std::shared_ptr<RecordBatch> batch,
+                                                     const int64_t& alignment,
+                                                     MemoryPool* memory_pool) {
+  std::vector<bool> needs_alignment;
+  if (!CheckAlignment(*batch, alignment, needs_alignment)) {
+    ArrayVector columns_ = batch->columns();
+    for (int i = 0; i < batch->num_columns(); ++i) {
+      if (needs_alignment[i] && columns_[i]) {
+        ARROW_ASSIGN_OR_RAISE(
+            columns_[i], EnsureAlignment(std::move(columns_[i]), alignment, memory_pool));
+      }
+    }
+    return RecordBatch::Make(batch->schema(), batch->num_rows(), std::move(columns_));
+  } else {
+    return std::move(batch);
+  }
+}
+
+Result<std::shared_ptr<Table>> EnsureAlignment(std::shared_ptr<Table> table,
+                                               const int64_t& alignment,
+                                               MemoryPool* memory_pool) {
+  std::vector<bool> needs_alignment;
+  if (!CheckAlignment(*table, alignment, needs_alignment)) {
+    std::vector<std::shared_ptr<ChunkedArray>> columns_ = table->columns();
+    for (int i = 1; i <= table->num_columns(); ++i) {
+      if (columns_[i - 1] && needs_alignment[i * columns_[i - 1]->num_chunks() + i - 1]) {
+        ArrayVector chunks_ = columns_[i - 1]->chunks();
+        for (size_t j = 0; j < chunks_.size(); ++j) {
+          if (chunks_[j] &&
+              needs_alignment[j + (i - 1) * (1 + columns_[i - 1]->num_chunks())]) {
+            ARROW_ASSIGN_OR_RAISE(chunks_[j], EnsureAlignment(std::move(chunks_[j]),
+                                                              alignment, memory_pool));
+          }
+        }
+        ARROW_ASSIGN_OR_RAISE(
+            columns_[i - 1],
+            ChunkedArray::Make(std::move(chunks_), columns_[i - 1]->type()));
+      }
+    }
+    return Table::Make(table->schema(), std::move(columns_), table->num_rows());
+  } else {
+    return std::move(table);
+  }
 }
 
 }  // namespace util
--- a/cpp/src/arrow/util/align_util.cc
+++ b/cpp/src/arrow/util/align_util.cc
@@ -36,58 +36,62 @@ bool CheckAlignment(const ArrayData& array, int64_t alignment) {
       if (!CheckAlignment(*buffer, alignment)) return false;
     }
   }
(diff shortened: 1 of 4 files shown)


cpp/src/arrow/util/align_util.h
```diff
@@ -64,5 +66,23 @@ inline BitmapWordAlignParams BitmapWordAlign(const uint8_t* data, int64_t bit_of
   return p;
 }
 
+template <typename T>
```
westonpace, line 69: Why is this templated?  Right now this relies on `T::data()` existing.  What other types of `T` do you expect to accept?

cpp/src/arrow/util/align_util.h
```diff
@@ -64,5 +66,23 @@ inline BitmapWordAlignParams BitmapWordAlign(const uint8_t* data, int64_t bit_of
...
+                       MemoryPool* memory_pool) {
+  std::vector<std::shared_ptr<Buffer>> buffers_ = object->data()->buffers;
+  for (size_t i = 0; i < buffers_.size(); ++i) {
+    if (buffers_[i]) {
+      auto buffer_address = buffers_[i]->address();
+      if ((buffer_address % alignment) != 0) {
+        ARROW_ASSIGN_OR_RAISE(
+            auto new_buffer, AllocateBuffer(buffers_[i]->size(), alignment, memory_pool));
+        std::memcpy(new_buffer->mutable_data(), buffers_[i]->data(), buffers_[i]->size());
+        object->data()->buffers[i] = std::move(new_buffer);
```
westonpace, line 80: It'd be nice (and more consistent with the rest of the code base) if we could do this by returning a new `std::shared_ptr<Array>` instead of modifying the existing array in-place.  You should be able to create a new `ArrayData` with the new buffers and then create a new `Array` from that.

cpp/src/arrow/util/align_util.h
```diff
@@ -64,5 +66,23 @@ inline BitmapWordAlignParams BitmapWordAlign(const uint8_t* data, int64_t bit_of
   return p;
 }
 
+template <typename T>
+Status EnsureAlignment(std::shared_ptr<T> object, int64_t alignment,
```
westonpace, line 70: This method doesn't recurse into child arrays.  For example, a `StructArray` is going to have `object->children()` that will need to be visited as well.

westonpace (changes requested):
This is a good start but we need to be more thorough.

cpp/src/arrow/util/align_util.h
sanjibansg: Made the change, thanks!

cpp/src/arrow/util/align_util.cc
```diff
@@ -26,56 +26,63 @@ namespace arrow {
 
 namespace util {
 
-Result<std::shared_ptr<Array>> EnsureAlignment(const Array& object, int64_t alignment,
+Result<std::shared_ptr<Buffer>> EnsureAlignment(const std::shared_ptr<Buffer>& object, int64_t alignment,
                                                MemoryPool* memory_pool) {
```
westonpace, line 30:
```suggestion
Result<std::shared_ptr<Buffer>> EnsureAlignment(std::shared_ptr<Buffer> object, int64_t alignment,
                                               MemoryPool* memory_pool) {
```

Also, the method should use move appropriate so I can do...

```
std::shared_ptr<Buffer> foo = ...
foo = EnsureAlignment(std::move(foo), 64, default_memory_pool())
// Should only make a copy of `foo` or the `shared_ptr` if the buffer was not aligned.
```

cpp/src/arrow/util/align_util.cc
```diff
@@ -26,56 +26,63 @@ namespace arrow {
...
-  std::vector<std::shared_ptr<Buffer>> buffers_ = object.data()->buffers;
+    auto buffer_address = object->address();
+    if ((buffer_address % alignment) != 0) {
+        ARROW_ASSIGN_OR_RAISE(
+                auto new_buffer, AllocateBuffer(object->size(), alignment, memory_pool));
+        std::memcpy(new_buffer->mutable_data(), object->data(), object->size());
+        return new_buffer;
+    } else {
+        return object;
```
westonpace, line 38:
```suggestion
        return std::move(object);
```

Needed if you follow the previous advice and change to `std::shared_ptr<Buffer> object` instead of `const std::shared_ptr<Buffer>&`

cpp/src/arrow/util/align_util.cc
```diff
@@ -26,56 +26,63 @@ namespace arrow {
 
 namespace util {
 
-Result<std::shared_ptr<Array>> EnsureAlignment(const Array& object, int64_t alignment,
+Result<std::shared_ptr<Buffer>> EnsureAlignment(const std::shared_ptr<Buffer>& object, int64_t alignment,
```
westonpace, line 29:
```suggestion
Result<std::shared_ptr<Buffer>> EnsureAlignment(const std::shared_ptr<Buffer>& buffer, int64_t alignment,
```

cpp/src/arrow/util/align_util.cc
```diff
@@ -26,56 +26,63 @@ namespace arrow {
...
+        ARROW_ASSIGN_OR_RAISE(
+                auto new_buffer, AllocateBuffer(object->size(), alignment, memory_pool));
+        std::memcpy(new_buffer->mutable_data(), object->data(), object->size());
+        return new_buffer;
+    } else {
+        return object;
+    }
+}
+
+Result<std::shared_ptr<Array>> EnsureAlignment(const std::shared_ptr<Array>& object, int64_t alignment,
```
westonpace, line 42:
```suggestion
Result<std::shared_ptr<Array>> EnsureAlignment(const std::shared_ptr<Array>& array, int64_t alignment,
```

cpp/src/arrow/util/align_util.cc
```diff
@@ -26,56 +26,63 @@ namespace arrow {
...
-        std::memcpy(new_buffer->mutable_data(), buffers_[i]->data(), buffers_[i]->size());
-        buffers_[i] = std::move(new_buffer);
+        ARROW_ASSIGN_OR_RAISE(buffers_[i], EnsureAlignment(buffers_[i], alignment, memory_pool));
       }
-    }
   }
   auto new_array_data =
-      ArrayData::Make(object.data()->type, object.data()->length, std::move(buffers_),
-                      object.data()->GetNullCount(), object.data()->offset);
+      ArrayData::Make(object->data()->type, object->data()->length, std::move(buffers_),
```
westonpace, line 51: You still need to handle the dictionary and the children.

cpp/src/arrow/util/align_util.cc
```diff
@@ -26,56 +26,63 @@ namespace arrow {
...
+                      object->data()->GetNullCount(), object->data()->offset);
   return MakeArray(new_array_data);
 }
 
-Result<std::shared_ptr<ChunkedArray>> EnsureAlignment(const ChunkedArray& object,
+Result<std::shared_ptr<ChunkedArray>> EnsureAlignment(const std::shared_ptr<ChunkedArray>& object,
```
westonpace, line 56:
```suggestion
Result<std::shared_ptr<ChunkedArray>> EnsureAlignment(const std::shared_ptr<ChunkedArray>& array,
```

cpp/src/arrow/util/align_util.cc
```diff
@@ -26,56 +26,63 @@ namespace arrow {
...
     ARROW_ASSIGN_OR_RAISE(chunks_[i],
-                          EnsureAlignment(*object.chunk(i), alignment, memory_pool));
+                          EnsureAlignment(object->chunk(i), alignment, memory_pool));
   }
-  return ChunkedArray::Make(std::move(chunks_), object.type());
+  return ChunkedArray::Make(std::move(chunks_), object->type());
 }
 
-Result<std::shared_ptr<RecordBatch>> EnsureAlignment(const RecordBatch& object,
+Result<std::shared_ptr<RecordBatch>> EnsureAlignment(const std::shared_ptr<RecordBatch>& object,
```
westonpace, line 67:
```suggestion
Result<std::shared_ptr<RecordBatch>> EnsureAlignment(const std::shared_ptr<RecordBatch>& batch,
```

cpp/src/arrow/util/align_util.cc
```diff
@@ -26,56 +26,63 @@ namespace arrow {
...
     ARROW_ASSIGN_OR_RAISE(columns_[i],
-                          EnsureAlignment(*object.column(i), alignment, memory_pool));
+                          EnsureAlignment(object->column(i), alignment, memory_pool));
   }
-  return RecordBatch::Make(object.schema(), object.num_rows(), std::move(columns_));
+  return RecordBatch::Make(object->schema(), object->num_rows(), std::move(columns_));
 }
 
-Result<std::shared_ptr<Table>> EnsureAlignment(const Table& object, int64_t alignment,
+Result<std::shared_ptr<Table>> EnsureAlignment(const std::shared_ptr<Table>& object, int64_t alignment,
```
westonpace, line 78:
```suggestion
Result<std::shared_ptr<Table>> EnsureAlignment(const std::shared_ptr<Table>& table, int64_t alignment,
```

cpp/src/arrow/util/align_util.h
```diff
@@ -68,19 +68,25 @@ inline BitmapWordAlignParams BitmapWordAlign(const uint8_t* data, int64_t bit_of
 
 namespace util {
 
-Result<std::shared_ptr<Array>> EnsureAlignment(const Array& object, int64_t alignment,
-                                               MemoryPool* memory_pool);
+Result<std::shared_ptr<Buffer>> EnsureAlignment(const std::shared_ptr<Buffer>& object,
```
westonpace, line 71:
```suggestion
Result<std::shared_ptr<Buffer>> EnsureAlignment(const std::shared_ptr<Buffer>& buffer,
```

cpp/src/arrow/util/align_util.h
```diff
@@ -68,19 +68,25 @@ inline BitmapWordAlignParams BitmapWordAlign(const uint8_t* data, int64_t bit_of
...
+                                                int64_t alignment,
+                                                MemoryPool* memory_pool);
 
-Result<std::shared_ptr<ChunkedArray>> EnsureAlignment(const ChunkedArray& object,
-                                                      int64_t alignment,
-                                                      MemoryPool* memory_pool);
+Result<std::shared_ptr<Array>> EnsureAlignment(const std::shared_ptr<Array>& object,
```
westonpace, line 75:
```suggestion
Result<std::shared_ptr<Array>> EnsureAlignment(const std::shared_ptr<Array>& array,
```

cpp/src/arrow/util/align_util.h
```diff
@@ -68,19 +68,25 @@ inline BitmapWordAlignParams BitmapWordAlign(const uint8_t* data, int64_t bit_of
...
+                                               int64_t alignment,
+                                                MemoryPool* memory_pool);
 
-Result<std::shared_ptr<RecordBatch>> EnsureAlignment(const RecordBatch& object,
-                                                     int64_t alignment,
-                                                     MemoryPool* memory_pool);
+Result<std::shared_ptr<ChunkedArray>> EnsureAlignment(
+    const std::shared_ptr<ChunkedArray>& object, int64_t alignment,
```
westonpace, line 80:
```suggestion
    const std::shared_ptr<ChunkedArray>& array, int64_t alignment,
```

cpp/src/arrow/util/align_util.h
```diff
@@ -68,19 +68,25 @@ inline BitmapWordAlignParams BitmapWordAlign(const uint8_t* data, int64_t bit_of
...
+    MemoryPool* memory_pool);
 
-Result<std::shared_ptr<Table>> EnsureAlignment(const Table& object, int64_t alignment,
-                                               MemoryPool* memory_pool);
+Result<std::shared_ptr<RecordBatch>> EnsureAlignment(
+    const std::shared_ptr<RecordBatch>& object, int64_t alignment,
```
westonpace, line 84:
```suggestion
    const std::shared_ptr<RecordBatch>& batch, int64_t alignment,
```

cpp/src/arrow/util/align_util.h
```diff
@@ -68,19 +68,25 @@ inline BitmapWordAlignParams BitmapWordAlign(const uint8_t* data, int64_t bit_of
...
+    MemoryPool* memory_pool);
+
+Result<std::shared_ptr<Table>> EnsureAlignment(const std::shared_ptr<Table>& object,
```
westonpace, line 87:
```suggestion
Result<std::shared_ptr<Table>> EnsureAlignment(const std::shared_ptr<Table>& table,
```

westonpace (changes requested):
In most cases the alignment will already be correct.  We want this method to be really fast in that scenario.  So this means that methods should look like...

```
std::shared_ptr<Array> Foo(std::shared_ptr<Array> thing) {
  if (needs_alignment(...)) {
    // a copy needs to be made
  } else {
    return std::move(thing);
  }
}
```

This way we can call it like so...

```
std::shared_ptr<Array> aligned = EnsureAlignment(std::move(unaligned), ...);
```

Then, if `unaligned` is already aligned it will be a straight move into `aligned` without ever making any copies (not even copies of `shared_ptr`)

cpp/src/arrow/util/align_util.h
sanjibansg: Now checking the children as well

cpp/src/arrow/util/align_util.cc
sanjibansg: Made the change, thanks!

cpp/src/arrow/util/align_util.cc
sanjibansg: Made the change.

cpp/src/arrow/util/align_util.cc
sanjibansg: Handling children and dictionary.

cpp/src/arrow/util/align_util.cc
sanjibansg: Made the change, thanks!

cpp/src/arrow/util/align_util.cc
sanjibansg: Made the change.

cpp/src/arrow/util/align_util.h
sanjibansg: Made the change.

cpp/src/arrow/util/align_util.h
sanjibansg: Made the change.

cpp/src/arrow/util/align_util.h
sanjibansg: Made the change.

cpp/src/arrow/util/align_util.h
sanjibansg: Made the change.

cpp/src/arrow/util/align_util.cc
sanjibansg: Made the change.

cpp/src/arrow/util/align_util.cc
sanjibansg: Made the change.

cpp/src/arrow/util/align_util.cc
sanjibansg: Made the change.

sanjibansg:
> In most cases the alignment will already be correct. We want this method to be really fast in that scenario. So this means that methods should look like...
> 
> ```
> std::shared_ptr<Array> Foo(std::shared_ptr<Array> thing) {
>   if (needs_alignment(...)) {
>     // a copy needs to be made
>   } else {
>     return std::move(thing);
>   }
> }
> ```
> 
> This way we can call it like so...
> 
> ```
> std::shared_ptr<Array> aligned = EnsureAlignment(std::move(unaligned), ...);
> ```
> 
> Then, if `unaligned` is already aligned it will be a straight move into `aligned` without ever making any copies (not even copies of `shared_ptr`)

The current implementation might be a bit complicated, so comments and suggestions for optimized approaches will be very helpful. Thanks in advance! 

Currently, we have `Check` functions for various Arrow objects, which check their constituents for alignments. For `ChunkedArray`, `RecordBatch`, and `Table`, we use a boolean vector to track the constituents which require alignment thus avoiding unnecessary function calls.

cpp/src/arrow/util/align_util.cc
```diff
@@ -0,0 +1,215 @@
...
+    if (child) {
+      if (!CheckAlignment(*child, alignment)) align_children = false;
+    }
+  }
+  return CheckAlignment(*array.data(), alignment) && align_dictionary && align_children;
+}
+
+bool CheckAlignment(const ChunkedArray& array, const int64_t& alignment,
+                    std::vector<bool>& needs_alignment, const int& offset) {
+  needs_alignment.resize(needs_alignment.size() + array.num_chunks(), false);
```
westonpace, line 59: Why `needs_alignment.size() + array.num_chunks()` and not `offset + array.num_chunks()`?

cpp/src/arrow/util/align_util.cc
```diff
@@ -0,0 +1,215 @@
...
+  for (auto i = 0; i < array.num_chunks(); ++i) {
+    if (array.chunk(i) && !CheckAlignment(*array.chunk(i), alignment))
+      needs_alignment[i + offset] = true;
+  }
+  return std::find(needs_alignment.begin() + offset,
```
westonpace, line 64: Instead of doing `std::find` at the end (which will loop through `needs_alignment`) could you do something like...

```
bool at_least_one_needs_alignment = false;
for (...) {
  if (...) {
    needs_alignment[i + offset] = true;
    at_least_one_needs_alignment = true;
  }
}
return at_least_one_needs_alignment;
```

cpp/src/arrow/util/align_util_test.cc
```diff
@@ -145,6 +150,84 @@ TEST(BitmapWordAlign, UnalignedDataStart) {
...
 }
-
 }  // namespace internal
+
+TEST(EnsureAlignment, Array) {
+  MemoryPool* pool = default_memory_pool();
+  auto rand = ::arrow::random::RandomArrayGenerator(1923);
+  auto random_array = rand.UInt8(/*size*/ 50, /*min*/ 0, /*max*/ 100,
+                                 /*null_probability*/ 0, /*alignment*/ 512, pool);
+  ASSERT_OK_AND_ASSIGN(auto aligned_array,
```
westonpace, line 160: Maybe assert that `CheckAlignment(..., 1024)` returns `false` before you call `EnsureAlignment`?

westonpace (changes requested):
The approach looks ok.  I have a few suggestions.

cpp/src/arrow/util/align_util.h
```diff
@@ -63,6 +64,43 @@ inline BitmapWordAlignParams BitmapWordAlign(const uint8_t* data, int64_t bit_of
   p.aligned_start = data + (bit_offset + p.leading_bits) / 8;
   return p;
 }
-
 }  // namespace internal
+
+namespace util {
+
+ARROW_EXPORT bool CheckAlignment(const Buffer& buffer, const int64_t& alignment);
```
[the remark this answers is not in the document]
sanjibansg: Oh yes, sorry, I did that by mistake. Corrected that. Thanks for pointing it out.

cpp/src/arrow/util/align_util.h
```diff
@@ -63,6 +64,43 @@ inline BitmapWordAlignParams BitmapWordAlign(const uint8_t* data, int64_t bit_of
...
+ARROW_EXPORT bool CheckAlignment(const ArrayData& array, const int64_t& alignment);
+ARROW_EXPORT bool CheckAlignment(const Array& array, const int64_t& alignment);
+ARROW_EXPORT bool CheckAlignment(const ChunkedArray& array, const int64_t& alignment,
+                                 std::vector<bool>& needs_alignment,
+                                 const int& offset = 0);
```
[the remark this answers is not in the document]
sanjibansg: Made the change, thanks for the suggestion.

cpp/src/arrow/util/align_util.cc
```diff
@@ -0,0 +1,215 @@
...
+  if (array.type()->id() == Type::DICTIONARY) {
+    align_dictionary = CheckAlignment(*array.data()->dictionary, alignment);
+  }
+
+  for (const auto& child : array.data()->child_data) {
+    if (child) {
+      if (!CheckAlignment(*child, alignment)) align_children = false;
+    }
+  }
+  return CheckAlignment(*array.data(), alignment) && align_dictionary && align_children;
```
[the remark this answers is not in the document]
sanjibansg: Moved the checking to the `CheckAlignment` function for `ArrayData`.

cpp/src/arrow/util/align_util.cc
```diff
@@ -0,0 +1,215 @@
...
+bool CheckAlignment(const Array& array, const int64_t& alignment) {
+  bool align_dictionary = true, align_children = true;
+
+  if (array.type()->id() == Type::DICTIONARY) {
+    align_dictionary = CheckAlignment(*array.data()->dictionary, alignment);
+  }
+
+  for (const auto& child : array.data()->child_data) {
+    if (child) {
+      if (!CheckAlignment(*child, alignment)) align_children = false;
```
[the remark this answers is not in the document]
sanjibansg: Made the change, thanks.

cpp/src/arrow/util/align_util.cc
sanjibansg: The resize is basically to add space for the extra elements that should be added for the check. 
For example, if we have a table made up of 2 chunked arrays, which then are made of 2 arrays each. Then, initially `needs_alignment` will be of size 2 (because of 2 chunked arrays). When the `CheckAlignment()` is called for the first chunked array, `needs_alignment` will be resized to 4 `(/*needs_alignment.size()/* 2 + /*array.num_chunks()*/ 2), thus now in needs_alignment, first two bits indicate status of the 2 arrays of the first chunk array, the third one for the ChunkArray as a whole, and the last one for the other chunk array which yet to be checked. The similar operation gets repeated for the other chunked array.

Instead if we use offset, then for the first iteration here, offset will be 0. So, the resize will enforce its size to 2, which will not be enough for storing the alignment  check bits of the first two arrays of the first chunk array, and the second chunk array.

cpp/src/arrow/util/align_util.h
```diff
@@ -63,6 +64,43 @@ inline BitmapWordAlignParams BitmapWordAlign(const uint8_t* data, int64_t bit_of
...
+ARROW_EXPORT bool CheckAlignment(const RecordBatch& batch, const int64_t& alignment,
+                                 std::vector<bool>& needs_alignment);
+ARROW_EXPORT bool CheckAlignment(const Table& table, const int64_t& alignment,
+                                 std::vector<bool>& needs_alignment);
```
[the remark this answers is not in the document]
sanjibansg: Added documentation for CheckAlignment functions. Thanks for the suggestion.

cpp/src/arrow/util/align_util_test.cc
sanjibansg: Made the change, thanks.

westonpace (changes requested):
I believe the unit tests are failling because `CopySlice` is not guaranteed to give you an unaligned array.

cpp/src/arrow/util/align_util_test.cc
```diff
@@ -59,6 +63,18 @@ void CheckBitmapWordAlign(const uint8_t* data, int64_t bit_offset, int64_t lengt
   }
 }
 
+arrow::Result<std::shared_ptr<Array>> CreateUnalignedArray(std::shared_ptr<Array> array) {
+  BufferVector sliced_buffers(array->data()->buffers.size(), nullptr);
+  for (std::size_t i = 0; i < array->data()->buffers.size(); ++i) {
+    if (array->data()->buffers[i]) {
+      sliced_buffers[i] = array->data()->buffers[i]->CopySlice(0, 2).MoveValueUnsafe();
```
westonpace, line 70: Sorry, this comment was meant to accompany the review but it seems Github swallowed it.  `CopySafe` allocates a new buffer and copies the slice into this new buffer.  The newly allocated buffer will have at least 64 byte alignment but, depending on the system allocator (or just bad luck), this alignment could be greater.

Instead you should use [`SliceBuffer`](https://github.com/apache/arrow/blob/main/cpp/src/arrow/buffer.h#L322).  However, `SliceBuffer` returns a view only (this is inevitable here) so you need to make sure to keep the original array alive.  To do that change this method to accept `const Array&` instead of `std::shared_ptr<Array>`.  Then you should be ok.

cpp/src/arrow/util/align_util_test.cc
```diff
@@ -59,6 +63,18 @@ void CheckBitmapWordAlign(const uint8_t* data, int64_t bit_offset, int64_t lengt
...
+    }
+  }
+  auto sliced_array_data =
+      ArrayData::Make(array->type(), array->length(), std::move(sliced_buffers));
```
westonpace, line 74: The length should be smaller here correct?  Since your arrays are all `uint8` you should decrease the length to match the slice.

cpp/src/arrow/util/align_util_test.cc
sanjibansg: made the change, thanks for the suggestion.

cpp/src/arrow/util/align_util_test.cc
```diff
@@ -59,6 +63,18 @@ void CheckBitmapWordAlign(const uint8_t* data, int64_t bit_offset, int64_t lengt
   }
 }
 
+arrow::Result<std::shared_ptr<Array>> CreateUnalignedArray(const Array& array) {
+  BufferVector sliced_buffers(array.data()->buffers.size(), nullptr);
```
westonpace, line 67:
```suggestion
  // Slicing by 1 would create an invalid array if the type was wider than
  // 1 byte so double-check that the array is a 1-byte type
  EXPECT_EQ(array.type_id(), Type::UINT8);
  BufferVector sliced_buffers(array.data()->buffers.size(), nullptr);
```

cpp/src/arrow/util/align_util_test.cc
```diff
@@ -59,6 +63,18 @@ void CheckBitmapWordAlign(const uint8_t* data, int64_t bit_offset, int64_t lengt
...
+  for (std::size_t i = 0; i < array.data()->buffers.size(); ++i) {
+    if (array.data()->buffers[i]) {
+      sliced_buffers[i] = SliceBuffer(array.data()->buffers[i], 1, 2);
+    }
+  }
+  auto sliced_array_data =
+      ArrayData::Make(array.type(), array.length(), std::move(sliced_buffers));
```
westonpace, line 74:
```suggestion
  auto sliced_array_data =
      ArrayData::Make(array.type(), /*length=*/2, std::move(sliced_buffers));
```

This works today, but only because you never call `Validate` or any kind of compute function on the returned array.  This could be a time bomb for a future maintainer.  Better to make sure the array is properly formed which means the length of the array must match the length of its buffers.

westonpace (changes requested):
Thanks, can you make two small changes in the test?  They aren't strictly needed but they will help prevent future surprises for maintainers.

cpp/src/arrow/util/align_util_test.cc
sanjibansg: made the change, thanks for the suggestion.

westonpace (approved):
Thanks for adding this utility!

kou:
@sanjibansg It seems that this breaks our nightly CI. Could you check this?

https://github.com/ursacomputing/crossbow/actions/runs/4518357593/jobs/7958180632#step:6:1378

```text
[ 34%] Building CXX object src/arrow/CMakeFiles/arrow_objlib.dir/util/async_util.cc.o
cd /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/build/src/arrow && /usr/lib64/ccache/c++  -DARROW_EXPORTING -DARROW_HAVE_RUNTIME_AVX2 -DARROW_HAVE_RUNTIME_AVX512 -DARROW_HAVE_RUNTIME_BMI2 -DARROW_HAVE_RUNTIME_SSE4_2 -DARROW_HAVE_SSE4_2 -DARROW_HDFS -DARROW_WITH_BACKTRACE -DARROW_WITH_BROTLI -DARROW_WITH_BZ2 -DARROW_WITH_LZ4 -DARROW_WITH_RE2 -DARROW_WITH_SNAPPY -DARROW_WITH_TIMING_TESTS -DARROW_WITH_UTF8PROC -DARROW_WITH_ZLIB -DARROW_WITH_ZSTD -DURI_STATIC_BUILD -DUTF8PROC_STATIC -I/root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/build/src -I/root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/src -I/root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/src/generated -isystem /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/thirdparty/flatbuffers/include -isystem /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/thirdparty/hadoop/include -isystem /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/build/boost_ep-prefix/src/boost_ep -isystem /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/build/orc_ep-install/include -isystem /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/build/zstd_ep-install/include -isystem /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/build/protobuf_ep-install/include -isystem /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/build/utf8proc_ep-install/include -isystem /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/build/re2_ep-install/include -isystem /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/build/xsimd_ep/src/xsimd_ep-install/include -isystem /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/build/jemalloc_ep-prefix/src  -Wno-noexcept-type -O2 -g -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches    -m64 -mtune=generic -fdiagnostics-color=always  -Wall -fno-semantic-interposition -msse4.2  -DNDEBUG -ftree-vectorize -fPIC   -pthread -std=c++1z -o CMakeFiles/arrow_objlib.dir/util/async_util.cc.o -c /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/src/arrow/util/async_util.cc
/root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/src/arrow/util/align_util.cc: In function 'arrow::Result<std::shared_ptr<arrow::Buffer> > arrow::util::EnsureAlignment(std::shared_ptr<arrow::Buffer>, int64_t, arrow::MemoryPool*)':
/root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/src/arrow/util/align_util.cc:104:12: error: could not convert 'new_buffer' from 'std::unique_ptr<arrow::Buffer>' to 'arrow::Result<std::shared_ptr<arrow::Buffer> >'
     return new_buffer;
            ^~~~~~~~~~
make[2]: *** [src/arrow/CMakeFiles/arrow_objlib.dir/util/align_util.cc.o] Error 1
```

sanjibansg:
> @sanjibansg It seems that this broke our nightly CI. Could you check this?
> 
> https://github.com/ursacomputing/crossbow/actions/runs/4518357593/jobs/7958180632#step:6:1375
> 
> ```
> [ 34%] Building CXX object src/arrow/CMakeFiles/arrow_objlib.dir/util/align_util.cc.o
> cd /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/build/src/arrow && /usr/lib64/ccache/c++  -DARROW_EXPORTING -DARROW_HAVE_RUNTIME_AVX2 -DARROW_HAVE_RUNTIME_AVX512 -DARROW_HAVE_RUNTIME_BMI2 -DARROW_HAVE_RUNTIME_SSE4_2 -DARROW_HAVE_SSE4_2 -DARROW_HDFS -DARROW_WITH_BACKTRACE -DARROW_WITH_BROTLI -DARROW_WITH_BZ2 -DARROW_WITH_LZ4 -DARROW_WITH_RE2 -DARROW_WITH_SNAPPY -DARROW_WITH_TIMING_TESTS -DARROW_WITH_UTF8PROC -DARROW_WITH_ZLIB -DARROW_WITH_ZSTD -DURI_STATIC_BUILD -DUTF8PROC_STATIC -I/root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/build/src -I/root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/src -I/root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/src/generated -isystem /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/thirdparty/flatbuffers/include -isystem /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/thirdparty/hadoop/include -isystem /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/build/boost_ep-prefix/src/boost_ep -isystem /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/build/orc_ep-install/include -isystem /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/build/zstd_ep-install/include -isystem /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/build/protobuf_ep-install/include -isystem /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/build/utf8proc_ep-install/include -isystem /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/build/re2_ep-install/include -isystem /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/build/xsimd_ep/src/xsimd_ep-install/include -isystem /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/build/jemalloc_ep-prefix/src  -Wno-noexcept-type -O2 -g -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches    -m64 -mtune=generic -fdiagnostics-color=always  -Wall -fno-semantic-interposition -msse4.2  -DNDEBUG -ftree-vectorize -fPIC   -pthread -std=c++1z -o CMakeFiles/arrow_objlib.dir/util/align_util.cc.o -c /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/src/arrow/util/align_util.cc
> ...
> /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/src/arrow/util/align_util.cc: In function 'arrow::Result<std::shared_ptr<arrow::Buffer> > arrow::util::EnsureAlignment(std::shared_ptr<arrow::Buffer>, int64_t, arrow::MemoryPool*)':
> /root/rpmbuild/BUILD/apache-arrow-12.0.0.dev290/cpp/src/arrow/util/align_util.cc:104:12: error: could not convert 'new_buffer' from 'std::unique_ptr<arrow::Buffer>' to 'arrow::Result<std::shared_ptr<arrow::Buffer> >'
>      return new_buffer;
>             ^~~~~~~~~~
> make[2]: *** [src/arrow/CMakeFiles/arrow_objlib.dir/util/align_util.cc.o] Error 1
> ```

Thanks for letting me know. I will make a fix.

kou:
FYI: We can test this case by commenting `@github-actions crossbow submit amazon-linux-2-amd64`.

sanjibansg:
@kou It is passing now! https://github.com/apache/arrow/pull/34754

cc: @westonpace

Issue #14761, opened 2022-11-29
[Dev] Update labeler workflow to use new component language labels


raulcd:
### Describe the enhancement requested

As discussed: https://github.com/apache/arrow/pull/14731#discussion_r1034503973
Update labeler to use new labels format for components like: `Component: python`

### Component(s)

Developer Tools

raulcd:
>  Is that something we could configure to use different labels? (to deduplicate this) I don't directly find any dependabot configuration.

we don't have specific configuration for dependabot, we seem to be using the defaults. We can change the labels if we add our own config based on this:
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#labels

raulcd:
I've created a specific issue to track the improvement to remove duplicated labels and add dependabot custom config. Closing this issue as the associated PR has been merged.

PR #14762, opened 2022-11-29
GH-14761: [Dev] Update labels on PR labeler to use new Component ones

raulcd:


raulcd:
* issue: #14761

.github/workflows/dev_pr/labeler.yml
```diff
@@ -15,53 +15,53 @@
...
-lang-go:
+"Component: go":
   - go/**/*
 
-lang-java:
+"Component: java":
   - java/**/*
 
-lang-js:
+"Component: js":
```
jorisvandenbossche, line 33:
```suggestion
"Component: JavaScript":
```

On the short term, we should probably match the labels exactly as the components on JIRA (we can rename them after the issues migration)? So for example "js" is "JavaScript"

raulcd: Thanks! I have updated all the labels to match the existing ones on JIRA.

jorisvandenbossche (approved):

jorisvandenbossche:
And I renamed all relevant labels

PR #14764, opened 2022-11-29
ARROW-18123: [Python] Fix writing files with multi-byte characters in file name

milesgranger:
Will close [ARROW-18123](https://issues.apache.org/jira/browse/ARROW-18123)

### DIFF
--- a/cpp/src/arrow/util/uri.cc
+++ b/cpp/src/arrow/util/uri.cc
@@ -18,12 +18,8 @@
 #include "arrow/util/uri.h"
 
 #include <algorithm>
-#include <codecvt>
 #include <cstring>
-#include <iostream>
-#include <locale>
 #include <sstream>
-#include <string>
 #include <string_view>
 #include <vector>
 
@@ -35,24 +31,23 @@ namespace internal {
 
 namespace {
 
-std::wstring_view TextRangeToView(const UriTextRangeStructW& range) {
+std::string_view TextRangeToView(const UriTextRangeStructA& range) {
   if (range.first == nullptr) {
-    return L"";
+    return "";
   } else {
     return {range.first, static_cast<size_t>(range.afterLast - range.first)};
   }
 }
 
-std::string TextRangeToString(const UriTextRangeStructW& range) {
-  std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
-  return converter.to_bytes(std::wstring(TextRangeToView(range)));
+std::string TextRangeToString(const UriTextRangeStructA& range) {
+  return std::string(TextRangeToView(range));
 }
 
 // There can be a difference between an absent field and an empty field.
 // For example, in "unix:/tmp/foo", the host is absent, while in
 // "unix:///tmp/foo", the host is empty but present.
 // This function helps distinguish.
-bool IsTextRangeSet(const UriTextRangeStructW& range) { return range.first != nullptr; }
+bool IsTextRangeSet(const UriTextRangeStructA& range) { return range.first != nullptr; }
 
 #ifdef _WIN32
 bool IsDriveSpec(const std::string_view s) {
@@ -117,10 +112,10 @@ bool IsValidUriScheme(const std::string_view s) {
 struct Uri::Impl {
   Impl() : string_rep_(""), port_(-1) { memset(&uri_, 0, sizeof(uri_)); }
 
-  ~Impl() { uriFreeUriMembersW(&uri_); }
+  ~Impl() { uriFreeUriMembersA(&uri_); }
 
   void Reset() {
-    uriFreeUriMembersW(&uri_);
+    uriFreeUriMembersA(&uri_);
     memset(&uri_, 0, sizeof(uri_));
     data_.clear();
     string_rep_.clear();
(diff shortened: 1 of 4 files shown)


cpp/src/arrow/util/uri.cc
```diff
@@ -245,19 +256,25 @@ const std::string& Uri::ToString() const { return impl_->string_rep_; }
...
+  std::wcout << "Wide URI: " << wide_uri << std::endl;
+
+  const auto& s = impl_->KeepString(wide_uri);
+  impl_->string_rep_ = converter.to_bytes(s);
+
+  const wchar_t* error_pos;
+  if (uriParseSingleUriExW(&impl_->uri_, s.data(), s.data() + s.size(), &error_pos) !=
       URI_SUCCESS) {
-    return Status::Invalid("Cannot parse URI: '", uri_string, "'");
+    return Status::Invalid("Cannot parse URI: '", uri_string, "' at: '",
```
milesgranger, line 269: Hi @pitrou, I know most of the PR is a bit of a mess, but I was thinking this specific area needed to have `wstring` support to parse the example in the issue (`例.parquet`) but while it now compiles, it still fails to parse it. And I'm out of ideas now. :sweat_smile:

pitrou:
As the comments on the JIRA diagnose, this is not a problem with URI parsing, but with the Python side not supporting relative paths.

Encoding the example filename to be URI-compatible gives the following:
```py
>>> from urllib.parse import quote
>>> quote("例.parquet")
'%E4%BE%8B.parquet'
```

milesgranger:
Apologies, maybe I misunderstood. I was reading relative paths in those comments in the context of a mutli-byte character, because a relative path "foo.parquet" works fine:

```python
In [1]: table = pa.table({'col1': pa.array([5, 6])})

In [2]: pq.write_table(table, 'foo.parquet')

In [3]: pq.write_table(table, '例.parquet')
ArrowInvalid                              Traceback (most recent call last)
Cell In [3], line 1
----> 1 pq.write_table(table, '例.parquet')
...
ArrowInvalid: Cannot parse URI: '例.parquet'
```

The and thus the error leading me to this attempt.

pitrou:
I suppose that's because "foo.parquet" in that case is a valid (local) URL.

jorisvandenbossche:
> I suppose that's because "foo.parquet" in that case is a valid (local) URL

So is "foo.parquet" considered a valid local URL, but "例.parquet" not? (because URLs cannot have multi-byte characters?)

But in any case, our LocalFileSystem works with local paths and does support such file names for example when inspecting with `get_file_info` (I created two empty files with those names):

```
In [1]: from pyarrow.fs import LocalFileSystem

In [2]: fs = LocalFileSystem()

In [3]: fs.get_file_info("foo.parquet")
Out[3]: <FileInfo for 'foo.parquet': type=FileType.File, size=0>

In [4]: fs.get_file_info("例.parquet")
Out[4]: <FileInfo for '例.parquet': type=FileType.File, size=0>
```

So since that the filesystem object itself can handle it, it seems there is going something wrong with how this is called in `write_table`.

jorisvandenbossche:
OK, so the problem is that we use `_resolve_filesystem_and_path` to preprocess the path. And for handling relative paths, this helper function assumes that the file should exist (it was written for when _reading_ data ...):

https://github.com/apache/arrow/blob/54e17920eee65e4227eba889aadbdfeb66c114cd/python/pyarrow/fs.py#L169-L185

And so if the file doesn't exist as a local (potentially relative) path, we fall back to use `from_uri`. And for this case, that then fails because indeed for a URI, the multi-byte character is invalid .. 
It actually also fails for "foo.parquet", because we don't support plain paths as URIs, but this gives:

```
In [12]: LocalFileSystem.from_uri("foo.parquet")
...
ArrowInvalid: URI has empty scheme: 'foo.parquet'
```

And that error about "empty scheme" is specifically checked for, to ignore that in this case:

https://github.com/apache/arrow/blob/54e17920eee65e4227eba889aadbdfeb66c114cd/python/pyarrow/fs.py#L184-L191

So I think the gist is that `_resolve_filesystem_and_path` wasn't written initially with _writing_ in mind (when files don't (need to) exist yet), and so we should update that logic.

jorisvandenbossche:
(as a result of the above, you also have the very confusing situation were the writing actually works if the file already exists, and will happily overwrite it in that case ..)

python/pyarrow/fs.py
```diff
@@ -183,11 +184,11 @@ def _resolve_filesystem_and_path(
     if not exists_locally:
         try:
             filesystem, path = FileSystem.from_uri(path)
-        except ValueError as e:
+        except pa.ArrowInvalid as e:
             # neither an URI nor a locally existing path, so assume that
             # local path was given and propagate a nicer file not found error
             # instead of a more confusing scheme parsing error
-            if "empty scheme" not in str(e):
+            if "URI" not in str(e):
```
jorisvandenbossche, line 191: This might be too broad. I think the "Unrecognized filesystem type in URI: .." error is certainly one we want to keep raising. 
Otherwise we will assume a local file path, which will then error because of getting something that looks like a URI (and then the user might think URIs are not supported in general, while they maybe just made a typo in the URI type, or are using an unsupported file system)

jorisvandenbossche (approved):
Thanks, looks good!

PR #14765, opened 2022-11-29
ARROW-18402: [C++] Expose `DeclarationInfo`

rtpsw:
See https://issues.apache.org/jira/browse/ARROW-18402

### DIFF
--- a/cpp/src/arrow/engine/substrait/relation.h
+++ b/cpp/src/arrow/engine/substrait/relation.h
@@ -1,0 +1,35 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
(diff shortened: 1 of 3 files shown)


rtpsw:
cc @westonpace

cpp/src/arrow/engine/substrait/relation.h
```diff
@@ -0,0 +1,35 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#pragma once
+
+#include "arrow/compute/exec/exec_plan.h"
```
westonpace, line 20:
```suggestion
#include <memory>

#include "arrow/compute/exec/exec_plan.h"
```

Minor IWYU nit

westonpace (changes requested):
Looks good, thanks!  This will be needed for extension relations.

Marking "request changes" because we should also add `relation.h` to `src/arrow/engine/api.h`

PR #14768, opened 2022-11-29
MINOR: Quick fix to the labeler for CPP files.

amol-:


assignUser (approved):
The workflow will still fail as the changes have to be on master to take effect.

raulcd (approved):
Thanks

PR #14771, opened 2022-11-29
ARROW-18416: [R] Update NEWS for 10.0.1

thisisnic:


r/NEWS.md
```diff
@@ -17,7 +17,16 @@
...
+# arrow 10.0.1.9000
+
+# arrow 10.0.1
+
+Minor fixes and updates:
+
+* Fixes for failing test after lubridate 1.9 release ([ARROW-18285](https://issues.apache.org/jira/browse/ARROW-18285))
+* Update to ensure compatibility with changes in dev purrr ([ARROW-18305](https://issues.apache.org/jira/browse/ARROW-18305))
+* Fix to correctly handle `.data` pronoun in `dplyr::group_by()` ([ARROW-18131](https://issues.apache.org/jira/browse/ARROW-18131))
+* Add deprecation cycle for `pull()` change ([ARROW-18132](https://issues.apache.org/jira/browse/ARROW-18132))
```
paleolimbot, line 29: I think that commit went in for 10.0.0, correct? (I seem to remember it had to or else the submission wouldn't go through). I would probably just remove this (unless it is, in fact, new since the last release)

paleolimbot (approved):

r/NEWS.md
thisisnic: I think you're right.  Other than you remembering this, is there any other way I could have known this? The ticket still says 10.0.1 and 11.0.0, should it be updated to say 10.0.0, or is this a change which was cherry-picked to the CRAN release after the RC had already been cut/approved?

thisisnic, line 29:
```suggestion
```

paleolimbot: I think that's probably why we should keep the CRAN-specific release branch in the Arrow main repo (rather than on Neal's fork). You probably would have had to look at that branch history to know that.

r/NEWS.md
```diff
@@ -17,7 +17,15 @@
   under the License.
 -->
 
-# arrow 10.0.0.9000
+# arrow 10.0.1.9000
+
+# arrow 10.0.1
```
thisisnic, line 22:
```suggestion
# arrow 10.0.0.9000
```

thisisnic: We don't do this manually

r/NEWS.md
```diff
@@ -19,6 +19,12 @@
 
 # arrow 10.0.0.9000
 
+Minor fixes and updates:
```
thisisnic, line 22:
```suggestion
Minor improvements and fixes:
```
This is the tidyverse way of doing it

PR #14530, opened 2022-10-27
ARROW-18174: [R] Fix compile of altrep.cpp on some builds

paleolimbot:
After ARROW-17187 (#14271), we get at least two nightly build failures because of how cpp11 objects are constructed in some new test functions. This PR makes the constructors less ambiguous so that no compilers give us trouble!

PR #14774, opened 2022-11-29
ARROW-18412: [C++][R] Windows build fails because of missing ChunkResolver symbols

paleolimbot:
Fix as suggested by @kou in https://github.com/apache/arrow/pull/14530#issuecomment-1328341447 !


kou (approved):
+1

Thanks!

Issue #14775, opened 2022-11-29
[Go] array.UnionBuilder.Len implementations always return 0


nwt:
### Describe the bug, including details regarding any error messages, version, and platform.

array.(*DenseUnionBuilder).Len and array.(*SparseUnionBuilder).Len always return 0. They should instead return the number of values appended.

This bug exists because these methods are promoted (via unionBuilder) from (*builder).Len, which reads builder.length, but DenseUnionBuilder and SparseUnionBuilder don't update that field.

It is a problem because some nested type builders require working child builder Len methods.  For example, at present, a ListBuilder for a union child type always builds empty lists and a MapBuilder for a union key type always builds empty maps.

One way to fix this is to add a (*unionBuilder).Len implementation that calls unionBuilder.typesBuilder.Len.
```go
// Len returns the current number of elements in the builder.
func (b *unionBuilder) Len() int { return b.typesBuilder.Len() }
```

### Component(s)

Go

jorisvandenbossche:
Issue resolved by pull request 14776
https://github.com/apache/arrow/pull/14776

PR #14750, opened 2022-11-28
GH-14720: [Dev] Update merge_arrow_pr script to accept GitHub issues

amol-:
* Closes: #14720

PR #14776, opened 2022-11-29
GH-14775: [Go] Fix UnionBuilder.Len implementations

nwt:
(*DenseUnionBuilder).Len and (*SparseUnionBuilder).Len always return 0. They should instead return the number of values appended.

This bug exists because these methods are promoted (via unionBuilder) from (*builder).Len, which reads builder.length, but DenseUnionBuilder and SparseUnionBuilder don't update that field.

Fix by adding a (*unionBuilder).Len implementation that calls unionBuilder.typesBuilder.Len.

### DIFF
--- a/go/arrow/array/union.go
+++ b/go/arrow/array/union.go
@@ -745,6 +745,9 @@ func (b *unionBuilder) Child(idx int) Builder {
 	return b.children[idx]
 }
 
+// Len returns the current number of elements in the builder.
+func (b *unionBuilder) Len() int { return b.typesBuilder.Len() }
+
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Thanks for this! Once all the tests in the CI pass, I'll merge this.

nwt:
@zeroshade: You're welcome! And thanks to you for all the work you've done in this repository!

zeroshade:
Actually, I need to wait until the merge script is updated from #14750 before I can merge this.

jorisvandenbossche:
@zeroshade I took the liberty to merge this as a final test of the merge script changes from https://github.com/apache/arrow/pull/14750

PR #14369, opened 2022-10-11
ARROW-14656: [Python] Add sorting for StructArray

Jedi18:
This is the first PR for https://github.com/apache/arrow/pull/11659 (splitting up this PR into multiple PRs). This PR adds a sort helper function for the array, chunked_array and array of structs (StructArray) and chunked array of structs sorting.

A `NestedValuesComparator` is added which is planned to be a common comparator for anything that has nested fields and columns. Follow up PRs will update the Table, RecordBatch etc sorting implementations to use this `NestedValuesComparator`.

Related issue: https://github.com/apache/arrow/issues/33206

PR #14781, opened 2022-11-30
GH-14778: [Python] Add (Chunked)Array sort() and RecordBatch.sort_by methods

### LINKED ISSUE #14778: [Python] Add sort_by helper to Table, RecordBatch, Array

Sorting is currently inconvenient in PyArrow as it requires chaining multiple compute functions to achieve the expected result. It would be good to provide helpers that make possible to simply perform `X.sort_by(...)`

### Component(s)

Python

amol-:
* Closes: #14778


amol-:
@jorisvandenbossche I should have addressed your feedbacks from https://github.com/apache/arrow/pull/14369#pullrequestreview-1156213617

python/pyarrow/array.pxi
```diff
@@ -2743,6 +2766,36 @@ cdef class StructArray(Array):
...
+    def sort(self, order="ascending", fieldname=None, **options):
+        """
+        Sort the StructArray
+
+        Parameters
+        ----------
+        order : str, default "ascending"
+            Which order to sort values in.
+            Accepted values are "ascending", "descending".
+        fieldname : str or None, default None
```
jorisvandenbossche, line 2778: >> We currently don't use "fieldname" anywhere. Maybe just `field` as the keyword argument?
>
> `field` implied a `Field` instance to me, that's why I went for `fieldname`.

Good point. Another suggestion, maybe something like `by=` ? 

(I suppose that in the future we might want to allow sorting by multiple fields?)

python/pyarrow/array.pxi
```diff
@@ -1399,6 +1399,29 @@ cdef class Array(_PandasConvertible):
...
+        Sort the Array
+
+        Parameters
+        ----------
+        order : str, default "ascending"
+            Which order to sort values in.
+            Accepted values are "ascending", "descending".
+        **options : dict, optional
+            Additional sorting options
+            As allowed by :class:`SortOptions`
```
jorisvandenbossche, line 1413: Good idea to just pass through options, that automatically also allows to use `null_placement`.

I know this are indeed options, but for consistency I would suggest to use `**kwargs`. For example Array.sum also uses kwargs to pass through options.

jorisvandenbossche: Another nit: I see we have this in other places as well, but the "dict" in the docstring as type for `**options` is quite misleading .. (you can't pass a dict in practice as argument)

(but since we use this everywhere, we can tackle that in a separate PR)

python/pyarrow/tests/test_array.py
```diff
@@ -3287,3 +3287,34 @@ def test_to_pandas_timezone():
...
+    assert sorted_arr.to_pylist() == [None, 35, 7, 5]
+
+
+def test_struct_array_sort():
+    arr = pa.StructArray.from_arrays([
+        pa.array([5, 7, 7, 35], type=pa.int64()),
+        pa.array(["foo", "car", "bar", "foobar"])
+    ], names=["a", "b"])
+
+    sorted_arr = arr.sort("descending", fieldname="a")
```
jorisvandenbossche, line 3314: Also add `arr.sort()` case without fieldname specified?

python/pyarrow/tests/test_table.py
```diff
@@ -2204,3 +2204,100 @@ def test_table_cast_invalid():
 
     table = pa.table({'a': [None, 1], 'b': [False, True]})
     assert table.cast(new_schema).schema == new_schema
+
+
+def test_table_sort():
```
jorisvandenbossche, line 2209: There is already a `test_table_sort_by` a little bit above. Can you move those new tests over there?

(this Table test might also be redundant with what there already is)

python/pyarrow/tests/test_table.py
```diff
@@ -2204,3 +2204,100 @@ def test_table_cast_invalid():
...
+    assert sorted_rb_dict["b"] == [1, 4, 3, 2]
+    assert sorted_rb_dict["c"] == ["car", "foo", "bar", "foobar"]
+
+    sorted_rb = rb.sort_by([("a", "ascending"), ("b", "ascending")])
+    sorted_rb_dict = sorted_rb.to_pydict()
+    assert sorted_rb_dict["a"] == [5, 7, 7, 35]
+    assert sorted_rb_dict["b"] == [2, 3, 4, 1]
+    assert sorted_rb_dict["c"] == ["foobar", "bar", "foo", "car"]
+
+    # test multi-key record batch sorter (> 8 sort keys)
```
jorisvandenbossche, line 2245: Just wondering: is there something specific about more than 8 keys?

python/pyarrow/array.pxi
amol-: Will rename to `kwargs`, regarding the `dict` type, that's what the numpydoc documentation suggests -> https://numpydoc.readthedocs.io/en/latest/format.html#parameters

> When documenting variable length positional, or keyword arguments, leave the leading star(s) in front of the name:
> ```
> *args : tuple
>     Additional arguments should be passed as keyword arguments
> **kwargs : dict, optional
>     Extra arguments to `metric`: refer to each metric documentation for a
>     list of all possible arguments.
> ```

python/pyarrow/tests/test_array.py
amol-: Until https://github.com/apache/arrow/pull/14369 is merged, that isn't supported.

amol-:
@jorisvandenbossche please re-review

python/pyarrow/tests/test_table.py
jorisvandenbossche: Do you know about this one? (or it was just copied from the other PR?) It seems overly complex for the functionality that this PR is adding

jorisvandenbossche:
I just realized an issue with the simple workaround for sorting a StructArray by selecting one of its fields, and that is that this ignores top-level nulls ..

Consider this example:

```
In [25]: arr = pa.StructArray.from_arrays([pa.array([5, 3, 4, 2, 1]), pa.array([1, 2, 3, 4, 5])], names=['a', 'b'], mask=pa.array([False, True, False, False, False]))

In [27]: arr.to_pylist()
Out[27]: [{'a': 5, 'b': 1}, None, {'a': 4, 'b': 3}, {'a': 2, 'b': 4}, {'a': 1, 'b': 5}]

In [30]: arr_sorted = arr.take(pc.sort_indices(arr.field('a')))

In [31]: arr_sorted.to_pylist()
Out[31]: [{'a': 1, 'b': 5}, {'a': 2, 'b': 4}, None, {'a': 4, 'b': 3}, {'a': 5, 'b': 1}]
```

This is due to what the `field()` method returns, of course. But we also have the `StructArray.flatten()` method, which gives the correct array to sort by (it only does this for all fields, while we only need one; but we could also directly use the `StructArray::GetFlattenedField` which is used by `flatten()` under the hood):

```
In [36]: arr_sorted2 = arr.take(pc.sort_indices(arr.flatten()[0]))

In [37]: arr_sorted2.to_pylist()
Out[37]: [{'a': 1, 'b': 5}, {'a': 2, 'b': 4}, {'a': 4, 'b': 3}, {'a': 5, 'b': 1}, None]
```

amol-:
> I just realized an issue with the simple workaround for sorting a StructArray by selecting one of its fields, and that is that this ignores top-level nulls ..
> 

Oh! Thanks for catching this. I took for granted that `field` would account for the array validity bitmap. At least, that seemed a reasonable expectation. I'll make sure to update the StructArray helper to account for the base array validity bitmap.

python/pyarrow/tests/test_table.py
Jedi18: There are two record batch sorters, RadixRecordBatchSorter and MultipleKeyRecordBatchSorter. I wanted to test both implementations so that's why this was added.

https://github.com/apache/arrow/blob/9b0f3d71a99f0d14f6d8e51524dc5a74ca5cea8e/cpp/src/arrow/compute/kernels/vector_sort.cc#L1260

jorisvandenbossche: @Jedi18 thanks for the clarification! Since this stripped down version of the PR actually doesn't touch the RecordBatch sort implementation, it might not be needed to add those tests (I assume those two variants of sorters are also already tested in C++)?

amol-:
@jorisvandenbossche the `StructArray.sort` does now account for the parent array nulls. Also I simplified the test for recordbatch. While I recognise we haven't modified that in this PR, it's still good to keep the test around.

jorisvandenbossche (approved):

PR #14783, opened 2022-11-30
MINOR: Setup linking between PR and Issue in GitHub

amol-:


assignUser (approved):
LGTM  :+1: We can add/refine functionality later on if needed.

jorisvandenbossche:
If we update the issue body, can we leave out the comment? Because then that is just duplicated (and in general the bot comments, if they are just informational without requiring action, are a bit notification noise)

.github/workflows/dev_pr/link.js
```diff
@@ -82,11 +82,22 @@ async function commentJIRAURL(github, context, pullRequestNumber, jiraID) {
...
+    const pullInfo = await github.pulls.get({
+        pull_number: pullRequestNumber,
+        owner: context.repo.owner,
+        repo: context.repo.repo,
+    }).data;
+    await github.pulls.update({
+      owner: context.repo.owner,
+      repo: context.repo.repo,
+      pull_number: pullRequestNumber,
+      body: pullInfo.body + "\n" + message;
```
raulcd, line 99: we can retrieve the body from the context:
```suggestion
      body: context.payload.pull_request.body + "\n" + message
```
and avoid doing the request for the PR. I've tested this on my fork and the only annoying thing is that we might want to deal with the body being null, see here: https://github.com/raulcd/arrow/pull/56#issue-1467688003

amol-:
> If we update the issue body, can we leave out the comment? Because then that is just duplicated (and in general the bot comments, if they are just informational without requiring action, are a bit notification noise)

Agree, made https://github.com/apache/arrow/issues/14790 to track the improvement suggestion.

raulcd (approved):
+1

Issue #14784, opened 2022-11-30
[Dev] Allow users to assign issues with a comment on the issue


raulcd:
### Describe the enhancement requested

With the migration to GitHub users don't have permissions to assign issues to themselves. This was possible with JIRA and we should allow and encourage users to work on issues. The current dev_pr workflow will auto-assign if a PR is opened and the issue did not have any assignee but it is good to do this preemptively to tell the rest of the community that someone is working on the issue.

Other projects like pandas or Ibis allow this to be done via a comment like `take` or `/take` on the issue itself.

We can update the comment bot to do that.

### Component(s)

Developer Tools

kou:
Issue resolved by pull request 14785
https://github.com/apache/arrow/pull/14785

PR #14785, opened 2022-11-30
GH-14784: [Dev] Add possibility to autoassign on GitHub issue comment

raulcd:
This PR will allow a user to type `take` in an issue comment and the issue will automatically be assigned to them.


raulcd:
I have tested this workflow on my fork: https://github.com/raulcd/arrow/issues/60#issuecomment-1332289389

jorisvandenbossche (approved):
Looks good!

.github/workflows/comment_bot.yml
```diff
@@ -174,3 +174,18 @@ jobs:
...
           args: "--force"
+
+  issue_assign:
+    name: "Assign issue"
+    permissions:
+      issues: write
+    if: github.event.comment.body == 'take'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
assignUser, line 185: We can remove this checkout by making the .js file into an action, that would be more idiomatic for gha.

assignUser: but I can work on that as a follow up too :)

raulcd: Thanks for the comment, it made me realise I could simplify it a lot as when I started I thought I might need more code for it. The action is not needed though.

assignUser (approved):
Nice!

kou (approved):
+1

PR #14800, opened 2022-12-01
ARROW-18391: [R] Fix the version selector dropdown in the dev docs

thisisnic:
For context, this overrides the default navbar HTML with a custom version which makes a single change - adding a span called "version" which the JS will override.  An identically-named span existed in the template used with Bootstrap 3, which is why no JavaScript changes were needed.  This approach is almost identical to that in this PR to the pkgdown package which implements the same thing: https://github.com/r-lib/pkgdown/pull/2072 

I built this locally, and it appears to have successfully added the dropdown back in.

image

### DIFF
--- a/r/pkgdown/templates/navbar.html
+++ b/r/pkgdown/templates/navbar.html
@@ -1,0 +1,37 @@
+{{#navbar}}
+<nav class="navbar fixed-top navbar-{{{type}}} navbar-expand-lg bg-{{{bg}}}">
+  <div class="container">
+    {{#includes}}{{{before_title}}}{{/includes}}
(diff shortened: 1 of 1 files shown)


kou:
Can I confirm what problem is fixed by this?

I confirmed the current https://arrow.apache.org/docs/r/ . The version dropdown shows "10.0.0.9000 (dev)" not  "10.0.0 (release)". Is it a problem that is fixed by this PR?

I also confirmed that "9.0.0" in https://arrow.apache.org/docs/r/ can move to https://arrow.apache.org/docs/9.0/r/ . I think that it's expected.
I also confirmed that "10.0.0 (release)" in https://arrow.apache.org/docs/9.0/r/ can move to https://arrow.apache.org/docs/r/ . I think that it's expected too.

kou (approved):
Thanks.
I understand.

I couldn't confirm the change details but +1 because you showed that it works on your local.

PR #14803, opened 2022-12-01
ARROW-18420: [C++][Parquet] Introduce ColumnIndex & OffsetIndex

wgtmac:
Basically this patch has defined the interface of `ColumnIndex` and `OffsetIndex`. Implementation classes are also provided to deserialize byte stream, wrap thrift message and provide access to their attributes.

BTW, the naming style throughout the code base looks confusing to me. I have tried to follow what I have understood from the parquet sub-directory. Please correct me if anything is incorrect.

### DIFF
--- a/cpp/src/parquet/page_index.cc
+++ b/cpp/src/parquet/page_index.cc
@@ -34,22 +34,22 @@ namespace {
 
 template <typename DType>
 void Decode(std::unique_ptr<typename EncodingTraits<DType>::Decoder>& decoder,
-            const std::string& input, std::vector<typename DType::c_type>& output,
-            size_t index) {
-  if (ARROW_PREDICT_FALSE(index >= output.size())) {
+            const std::string& input, std::vector<typename DType::c_type>* output,
+            size_t output_index) {
+  if (ARROW_PREDICT_FALSE(output_index >= output->size())) {
     throw ParquetException("Index out of bound");
   }
 
   decoder->SetData(/*num_values=*/1, reinterpret_cast<const uint8_t*>(input.c_str()),
                    static_cast<int>(input.size()));
-  decoder->Decode(&output[index], /*max_values=*/1);
+  decoder->Decode(&output->at(output_index), /*max_values=*/1);
 }
 
 template <>
 void Decode<BooleanType>(std::unique_ptr<BooleanDecoder>& decoder,
-                         const std::string& input, std::vector<bool>& output,
-                         size_t index) {
-  if (ARROW_PREDICT_FALSE(index >= output.size())) {
+                         const std::string& input, std::vector<bool>* output,
+                         size_t output_index) {
+  if (ARROW_PREDICT_FALSE(output_index >= output->size())) {
     throw ParquetException("Index out of bound");
   }
 
@@ -57,13 +57,13 @@ void Decode<BooleanType>(std::unique_ptr<BooleanDecoder>& decoder,
   decoder->SetData(/*num_values=*/1, reinterpret_cast<const uint8_t*>(input.c_str()),
                    static_cast<int>(input.size()));
   decoder->Decode(&value, /*max_values=*/1);
-  output[index] = value;
+  output->at(output_index) = value;
 }
 
 template <>
 void Decode<ByteArrayType>(std::unique_ptr<ByteArrayDecoder>&, const std::string& input,
-                           std::vector<ByteArray>& output, size_t index) {
-  if (ARROW_PREDICT_FALSE(index >= output.size())) {
+                           std::vector<ByteArray>* output, size_t output_index) {
+  if (ARROW_PREDICT_FALSE(output_index >= output->size())) {
     throw ParquetException("Index out of bound");
   }
 
@@ -72,7 +72,7 @@ void Decode<ByteArrayType>(std::unique_ptr<ByteArrayDecoder>&, const std::string
     throw ParquetException("Invalid encoded byte array length");
   }
 
-  auto& decoded = output.at(index);
+  auto& decoded = output->at(output_index);
   decoded.len = static_cast<uint32_t>(input.size());
   decoded.ptr = reinterpret_cast<const uint8_t*>(input.data());
 }
@@ -113,8 +113,8 @@ class TypedColumnIndexImpl : public TypedColumnIndex<DType> {
       if (!column_index_.null_pages[i]) {
         // The check on `num_pages` has guaranteed the cast below is safe.
         non_null_page_indices_.emplace_back(static_cast<int32_t>(i));
-        Decode<DType>(plain_decoder, column_index_.min_values[i], min_values_, i);
-        Decode<DType>(plain_decoder, column_index_.max_values[i], max_values_, i);
+        Decode<DType>(plain_decoder, column_index_.min_values[i], &min_values_, i);
+        Decode<DType>(plain_decoder, column_index_.max_values[i], &max_values_, i);
       }
     }
     DCHECK_EQ(num_non_null_pages, non_null_page_indices_.size());
--- a/cpp/src/parquet/page_index.cc
+++ b/cpp/src/parquet/page_index.cc
@@ -170,6 +170,7 @@ std::unique_ptr<ColumnIndex> ColumnIndex::Make(const ColumnDescriptor& descr,
       return std::make_unique<TypedColumnIndexImpl<FLBAType>>(descr, column_index);
     case Type::UNDEFINED:
       ::arrow::Unreachable("Cannot make ColumnIndex of an unknown type");
+      return nullptr;
   }
 }
 
--- a/cpp/src/parquet/page_index.cc
+++ b/cpp/src/parquet/page_index.cc
@@ -15,9 +15,9 @@
 // specific language governing permissions and limitations
 // under the License.
 
-#include "parquet/page_index.h"
 #include "parquet/encoding.h"
-#include "parquet/metadata.h"
+#include "parquet/page_index.h"
+#include "parquet/statistics.h"
 #include "parquet/thrift_internal.h"
 
 #include <map>
@@ -46,154 +46,70 @@ class TypedColumnIndexImpl : public TypedColumnIndex<DType> {
  public:
   using T = typename DType::c_type;
 
-  explicit TypedColumnIndexImpl(const ColumnDescriptor& descr,
-                                const std::vector<bool>& null_pages,
-                                const std::vector<std::string>& min_values,
-                                const std::vector<std::string>& max_values,
-                                const BoundaryOrder& boundary_order,
-                                const bool has_null_count = false,
-                                const std::vector<int64_t>& null_counts = {})
-      : null_pages_(null_pages),
-        encoded_min_values_(min_values),
-        encoded_max_values_(max_values),
-        boundary_order_(boundary_order),
-        has_null_count_(has_null_count),
-        null_counts_(null_counts) {
+  TypedColumnIndexImpl(const ColumnDescriptor& descr,
+                       const format::ColumnIndex& column_index)
+      : column_index_(column_index) {
+    min_values_.reserve(column_index_.null_pages.size());
+    max_values_.reserve(column_index_.null_pages.size());
     // Decode min and max values into a compact form (i.e. w/o null page)
     auto plain_decoder = MakeTypedDecoder<DType>(Encoding::PLAIN, &descr);
     T value;
-    for (size_t i = 0; i < null_pages_.size(); ++i) {
-      if (!null_pages_[i]) {
-        // page index -> min/max slot index
-        page_indexes_.emplace(i, min_values_.size());
-        Decode<DType>(plain_decoder, encoded_min_values_[i], &value);
+    for (size_t i = 0; i < column_index_.null_pages.size(); ++i) {
+      if (column_index_.null_pages[i]) {
+        min_values_.push_back(std::nullopt);
+        max_values_.push_back(std::nullopt);
+      } else {
+        Decode<DType>(plain_decoder, column_index_.min_values[i], &value);
         min_values_.push_back(value);
-        Decode<DType>(plain_decoder, encoded_max_values_[i], &value);
+        Decode<DType>(plain_decoder, column_index_.max_values[i], &value);
         max_values_.push_back(value);
       }
     }
   }
 
-  explicit TypedColumnIndexImpl(const ColumnDescriptor& descr,
-                                const format::ColumnIndex& column_index)
-      : TypedColumnIndexImpl(
-            descr, column_index.null_pages, column_index.min_values,
-            column_index.max_values,
-            static_cast<BoundaryOrder>(static_cast<int>(column_index.boundary_order)),
-            column_index.__isset.null_counts, column_index.null_counts) {}
-
-  int64_t num_pages() const override { return static_cast<int64_t>(null_pages_.size()); }
-
-  bool is_null_page(int64_t page_id) const override {
-    if (page_id >= num_pages()) {
-      throw ParquetException("Page index is out of bound");
-    }
-    return null_pages_[page_id];
+  const std::vector<bool>& null_pages() const override {
+    return column_index_.null_pages;
   }
 
-  BoundaryOrder boundary_order() const override { return boundary_order_; }
-
-  bool has_null_counts() const override { return has_null_count_; }
-
-  int64_t null_count(int64_t page_id) const override {
-    if (page_id >= num_pages()) {
-      throw ParquetException("Page index is out of bound");
-    }
-    return null_counts_[page_id];
+  const std::vector<std::string>& encoded_min_values() const override {
+    return column_index_.min_values;
   }
 
-  T min_value(int64_t page_id) const override {
-    return min_values_[GetMinMaxSlot(page_id)];
+  const std::vector<std::string>& encoded_max_values() const override {
+    return column_index_.max_values;
   }
 
-  T max_value(int64_t page_id) const override {
-    return max_values_[GetMinMaxSlot(page_id)];
+  BoundaryOrder boundary_order() const override {
+    return static_cast<BoundaryOrder>(static_cast<int>(column_index_.boundary_order));
   }
 
-  const std::string& encoded_min(int64_t page_id) const override {
-    if (page_id >= num_pages()) {
-      throw ParquetException("Page index is out of bound");
-    }
-    return encoded_min_values_[page_id];
+  bool has_null_counts() const override { return column_index_.__isset.null_counts; }
+
+  const std::vector<int64_t>& null_counts() const override {
+    return column_index_.null_counts;
   }
 
-  const std::string& encoded_max(int64_t page_id) const override {
-    if (page_id >= num_pages()) {
-      throw ParquetException("Page index is out of bound");
-    }
-    return encoded_max_values_[page_id];
-  }
+  const std::vector<std::optional<T>>& min_values() const override { return min_values_; }
 
-  const std::vector<bool>& null_pages() const override { return null_pages_; }
-
-  const std::vector<int64_t>& null_counts() const override { return null_counts_; }
-
-  const std::vector<T>& min_values() const override { return min_values_; }
-
-  const std::vector<T>& max_values() const override { return max_values_; }
-
-  std::vector<int64_t> GetValidPageIndices() const override {
-    std::vector<int64_t> valid_page_indices;
-    std::for_each(page_indexes_.cbegin(), page_indexes_.cend(),
-                  [&](const std::pair<size_t, size_t>& v) {
-                    valid_page_indices.push_back(v.first);
-                  });
-    return valid_page_indices;
-  }
+  const std::vector<std::optional<T>>& max_values() const override { return max_values_; }
 
  private:
-  size_t GetMinMaxSlot(int64_t page_id) const {
-    if (page_id >= static_cast<int64_t>(null_pages_.size())) {
-      throw ParquetException("Page index is out of bound");
-    }
-    if (null_pages_[page_id]) {
-      throw ParquetException("Cannot get min/max value of null page");
-    }
-    auto iter = page_indexes_.find(page_id);
-    if (iter == page_indexes_.cend()) {
-      throw ParquetException("min/max value is unavailable");
-    }
-    return iter->second;
-  }
-
-  /// Values that are copied directly from the thrift message.
-  std::vector<bool> null_pages_;
-  std::vector<std::string> encoded_min_values_;
-  std::vector<std::string> encoded_max_values_;
-  BoundaryOrder boundary_order_;
-  bool has_null_count_;
-  std::vector<int64_t> null_counts_;
-
-  /// page_id -> slot_id in the buffer of min_values_ & max_values_
-  std::map<size_t, size_t> page_indexes_;
-
-  /// Decoded typed min/max values.
-  std::vector<T> min_values_;
-  std::vector<T> max_values_;
+  /// Wrapped thrift column index.
+  const format::ColumnIndex column_index_;
+  /// Decoded typed min/max values. Null pages are set to std::nullopt.
+  std::vector<std::optional<T>> min_values_;
+  std::vector<std::optional<T>> max_values_;
 };
 
 class OffsetIndexImpl : public OffsetIndex {
  public:
-  explicit OffsetIndexImpl(std::vector<PageLocation> page_locations)
-      : page_locations_(std::move(page_locations)) {}
-
   explicit OffsetIndexImpl(const format::OffsetIndex& offset_index) {
+    page_locations_.reserve(offset_index.page_locations.size());
     for (const auto& page_location : offset_index.page_locations) {
-      page_locations_.emplace_back();
-      auto& location = page_locations_.back();
-      location.offset = page_location.offset;
-      location.compressed_page_size = page_location.compressed_page_size;
-      location.first_row_index = page_location.first_row_index;
+      page_locations_.emplace_back(PageLocation{page_location.offset,
+                                                page_location.compressed_page_size,
+                                                page_location.first_row_index});
     }
-  }
-
-  int64_t num_pages() const override { return page_locations_.size(); }
-
-  const PageLocation& GetPageLocation(int64_t page_id) const override {
-    if (page_id >= num_pages()) {
-      throw ParquetException("Page index is out of bound");
-    }
-    return page_locations_[page_id];
   }
 
   const std::vector<PageLocation>& GetPageLocations() const override {
--- a/cpp/src/parquet/page_index.cc
+++ b/cpp/src/parquet/page_index.cc
@@ -43,7 +43,7 @@ void Decode(std::unique_ptr<typename EncodingTraits<DType>::Decoder>& decoder,
 template <>
 void Decode<ByteArrayType>(std::unique_ptr<ByteArrayDecoder>&, const std::string& src,
                            ByteArray* dst) {
-  DCHECK_LE(src.size(), std::numeric_limits<uint32_t>::max());
+  DCHECK_LE(src.size(), static_cast<size_t>(std::numeric_limits<uint32_t>::max()));
   dst->len = static_cast<uint32_t>(src.size());
   dst->ptr = reinterpret_cast<const uint8_t*>(src.data());
 }
--- a/cpp/src/parquet/page_index.cc
+++ b/cpp/src/parquet/page_index.cc
@@ -25,7 +25,6 @@
 #include "arrow/util/unreachable.h"
 
 #include <limits>
-#include <map>
 #include <numeric>
 
 namespace parquet {
@@ -42,7 +41,10 @@ void Decode(std::unique_ptr<typename EncodingTraits<DType>::Decoder>& decoder,
 
   decoder->SetData(/*num_values=*/1, reinterpret_cast<const uint8_t*>(input.c_str()),
                    static_cast<int>(input.size()));
-  decoder->Decode(&output->at(output_index), /*max_values=*/1);
+  const auto num_values = decoder->Decode(&output->at(output_index), /*max_values=*/1);
+  if (ARROW_PREDICT_FALSE(num_values != 1)) {
+    throw ParquetException("Could not decode statistics value");
+  }
 }
 
 template <>
@@ -56,7 +58,10 @@ void Decode<BooleanType>(std::unique_ptr<BooleanDecoder>& decoder,
   bool value;
   decoder->SetData(/*num_values=*/1, reinterpret_cast<const uint8_t*>(input.c_str()),
                    static_cast<int>(input.size()));
-  decoder->Decode(&value, /*max_values=*/1);
+  const auto num_values = decoder->Decode(&value, /*max_values=*/1);
+  if (ARROW_PREDICT_FALSE(num_values != 1)) {
+    throw ParquetException("Could not decode statistics value");
+  }
   output->at(output_index) = value;
 }
 
@@ -72,9 +77,8 @@ void Decode<ByteArrayType>(std::unique_ptr<ByteArrayDecoder>&, const std::string
     throw ParquetException("Invalid encoded byte array length");
   }
 
-  auto& decoded = output->at(output_index);
-  decoded.len = static_cast<uint32_t>(input.size());
-  decoded.ptr = reinterpret_cast<const uint8_t*>(input.data());
+  output->at(output_index) = {/*len=*/static_cast<uint32_t>(input.size()),
+                              /*ptr=*/reinterpret_cast<const uint8_t*>(input.data())};
 }
 
 template <typename DType>
@@ -86,17 +90,16 @@ class TypedColumnIndexImpl : public TypedColumnIndex<DType> {
                        const format::ColumnIndex& column_index)
       : column_index_(column_index) {
     // Make sure the number of pages is valid and it does not overflow to int32_t.
-    if (ARROW_PREDICT_FALSE(column_index_.null_pages.size() >=
-                            static_cast<size_t>(std::numeric_limits<int32_t>::max())) ||
-        column_index_.null_pages.size() != column_index_.min_values.size() ||
-        column_index_.min_values.size() != column_index_.max_values.size() ||
+    const size_t num_pages = column_index_.null_pages.size();
+    if (num_pages >= static_cast<size_t>(std::numeric_limits<int32_t>::max()) ||
+        column_index_.min_values.size() != num_pages ||
+        column_index_.max_values.size() != num_pages ||
         (column_index_.__isset.null_counts &&
-         column_index_.null_counts.size() != column_index_.null_pages.size())) {
+         column_index_.null_counts.size() != num_pages)) {
       throw ParquetException("Invalid column index");
     }
 
-    size_t num_pages = column_index_.null_pages.size();
-    size_t num_non_null_pages = static_cast<size_t>(std::accumulate(
+    const size_t num_non_null_pages = static_cast<size_t>(std::accumulate(
         column_index_.null_pages.cbegin(), column_index_.null_pages.cend(), 0,
         [](int32_t num_non_null_pages, bool null_page) {
           return num_non_null_pages + (null_page ? 0 : 1);
@@ -104,8 +107,8 @@ class TypedColumnIndexImpl : public TypedColumnIndex<DType> {
     DCHECK_LE(num_non_null_pages, num_pages);
 
     // Allocate slots for decoded values.
-    min_values_.resize(num_pages);
-    max_values_.resize(num_pages);
+    min_values_.resize(num_non_null_pages);
+    max_values_.resize(num_non_null_pages);
     non_null_page_indices_.reserve(num_non_null_pages);
 
     // Decode min and max values according to the physical type.
--- a/cpp/src/parquet/page_index.cc
+++ b/cpp/src/parquet/page_index.cc
@@ -56,12 +56,12 @@ class TypedColumnIndexImpl : public TypedColumnIndex<DType> {
   TypedColumnIndexImpl(const ColumnDescriptor& descr,
                        const format::ColumnIndex& column_index)
       : column_index_(column_index) {
-    size_t num_non_null_pages = std::accumulate(
-        column_index_.null_pages.cbegin(), column_index_.null_pages.cend(), 0U,
-        [](size_t num_non_null_pages, bool null_page) {
-          return num_non_null_pages + (null_page ? 0U : 1U);
+    int32_t num_non_null_pages = std::accumulate(
+        column_index_.null_pages.cbegin(), column_index_.null_pages.cend(), 0,
+        [](int32_t num_non_null_pages, bool null_page) {
+          return num_non_null_pages + (null_page ? 0 : 1);
         });
-    if (num_non_null_pages != 0U) {
+    if (num_non_null_pages > 0) {
       min_values_.reserve(num_non_null_pages);
       max_values_.reserve(num_non_null_pages);
       // Decode min and max values into a compact form (i.e. w/o null page)
--- a/cpp/src/parquet/page_index.cc
+++ b/cpp/src/parquet/page_index.cc
@@ -55,14 +55,12 @@ class TypedColumnIndexImpl : public TypedColumnIndex<DType> {
     auto plain_decoder = MakeTypedDecoder<DType>(Encoding::PLAIN, &descr);
     T value;
     for (size_t i = 0; i < column_index_.null_pages.size(); ++i) {
-      if (column_index_.null_pages[i]) {
-        min_values_.push_back(std::nullopt);
-        max_values_.push_back(std::nullopt);
-      } else {
+      if (!column_index_.null_pages[i]) {
+        non_null_page_indices_.emplace_back(static_cast<int32_t>(i));
         Decode<DType>(plain_decoder, column_index_.min_values[i], &value);
-        min_values_.push_back(value);
+        min_values_.emplace_back(value);
         Decode<DType>(plain_decoder, column_index_.max_values[i], &value);
-        max_values_.push_back(value);
+        max_values_.emplace_back(value);
       }
     }
   }
@@ -89,16 +87,22 @@ class TypedColumnIndexImpl : public TypedColumnIndex<DType> {
     return column_index_.null_counts;
   }
 
-  const std::vector<std::optional<T>>& min_values() const override { return min_values_; }
+  const std::vector<T>& min_values() const override { return min_values_; }
 
-  const std::vector<std::optional<T>>& max_values() const override { return max_values_; }
+  const std::vector<T>& max_values() const override { return max_values_; }
+
+  const std::vector<int32_t> GetNonNullPageIndices() const override {
+    return non_null_page_indices_;
+  }
 
  private:
   /// Wrapped thrift column index.
   const format::ColumnIndex column_index_;
   /// Decoded typed min/max values. Null pages are set to std::nullopt.
-  std::vector<std::optional<T>> min_values_;
-  std::vector<std::optional<T>> max_values_;
+  std::vector<T> min_values_;
+  std::vector<T> max_values_;
+  /// A list of page indices for not-null pages.
+  std::vector<int32_t> non_null_page_indices_;
 };
 
 class OffsetIndexImpl : public OffsetIndex {
--- a/cpp/src/parquet/page_index.cc
+++ b/cpp/src/parquet/page_index.cc
@@ -107,8 +107,8 @@ class TypedColumnIndexImpl : public TypedColumnIndex<DType> {
     DCHECK_LE(num_non_null_pages, num_pages);
 
     // Allocate slots for decoded values.
-    min_values_.resize(num_non_null_pages);
-    max_values_.resize(num_non_null_pages);
+    min_values_.resize(num_pages);
+    max_values_.resize(num_pages);
     non_null_page_indices_.reserve(num_non_null_pages);
 
     // Decode min and max values according to the physical type.
--- a/cpp/src/parquet/page_index.cc
+++ b/cpp/src/parquet/page_index.cc
@@ -27,17 +27,16 @@ namespace parquet {
 namespace {
 
 template <typename DType>
-void PlainDecode(const ColumnDescriptor* descr, const std::string& src,
-                 typename DType::c_type* dst) {
-  auto decoder = MakeTypedDecoder<DType>(Encoding::PLAIN, descr);
-  decoder->SetData(1, reinterpret_cast<const uint8_t*>(src.c_str()),
+void Decode(std::unique_ptr<typename EncodingTraits<DType>::Decoder>& decoder,
+            const std::string& src, typename DType::c_type* dst) {
+  decoder->SetData(/*num_values=*/1, reinterpret_cast<const uint8_t*>(src.c_str()),
                    static_cast<int>(src.size()));
-  decoder->Decode(dst, 1);
+  decoder->Decode(dst, /*max_values=*/1);
 }
 
 template <>
-void PlainDecode<ByteArrayType>(const ColumnDescriptor* descr, const std::string& src,
-                                ByteArray* dst) {
+void Decode<ByteArrayType>(std::unique_ptr<ByteArrayDecoder>&, const std::string& src,
+                           ByteArray* dst) {
   dst->len = static_cast<uint32_t>(src.size());
   dst->ptr = reinterpret_cast<const uint8_t*>(src.c_str());
 }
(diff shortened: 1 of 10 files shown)


wgtmac:
@pitrou @emkornfield Can you please take a look at the proposed interface and implementation? Once they are finalized, I will add unit tests. Thank you very much!

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,143 @@
...
+/// \brief BoundaryOrder is a proxy around format::BoundaryOrder.
+enum class PARQUET_EXPORT BoundaryOrder { UNORDERED = 0, ASCENDING = 1, DESCENDING = 2 };
+
+/// \brief ColumnIndex is a proxy around format::ColumnIndex.
+class PARQUET_EXPORT ColumnIndex {
+ public:
+  /// \brief Create a ColumnIndex from a serialized thrift message.
+  static std::unique_ptr<ColumnIndex> Make(const ColumnDescriptor* descr,
+                                           const void* serialized_index,
+                                           uint32_t* inout_index_len,
```
pitrou, line 42: Why "inout"?

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,143 @@
...
+namespace parquet {
+
+/// \brief BoundaryOrder is a proxy around format::BoundaryOrder.
+enum class PARQUET_EXPORT BoundaryOrder { UNORDERED = 0, ASCENDING = 1, DESCENDING = 2 };
+
+/// \brief ColumnIndex is a proxy around format::ColumnIndex.
+class PARQUET_EXPORT ColumnIndex {
+ public:
+  /// \brief Create a ColumnIndex from a serialized thrift message.
+  static std::unique_ptr<ColumnIndex> Make(const ColumnDescriptor* descr,
```
pitrou, line 40: Why not a const ref? Is it allowed to be null?

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,143 @@
...
+                                           uint32_t* inout_index_len,
+                                           const ReaderProperties& properties);
+
+  virtual ~ColumnIndex() = default;
+
+  /// \brief Returns number of pages in this column index.
+  virtual int64_t num_pages() const = 0;
+
+  /// \brief Returns if all values are null in a single page.
+  virtual bool null_page(int64_t page_id) const = 0;
```
pitrou, line 51:
```suggestion
  virtual bool is_null_page(int64_t page_id) const = 0;
```

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,143 @@
...
+#include <vector>
+
+#include "parquet/platform.h"
+#include "parquet/statistics.h"
+#include "parquet/types.h"
+
+namespace parquet {
+
+/// \brief BoundaryOrder is a proxy around format::BoundaryOrder.
+enum class PARQUET_EXPORT BoundaryOrder { UNORDERED = 0, ASCENDING = 1, DESCENDING = 2 };
```
pitrou, line 34: Uppercase can clash with third-party macros unfortunately, we should avoid it as much as possible.
```suggestion
enum class PARQUET_EXPORT BoundaryOrder { Unordered = 0, Ascending = 1, Descending = 2 };
```

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,143 @@
...
+
+  /// \brief Returns if all values are null in a single page.
+  virtual bool null_page(int64_t page_id) const = 0;
+
+  /// \brief Returns whether both min_values and max_values are
+  /// orderd and if so, in which direction.
+  virtual BoundaryOrder boundary_order() const = 0;
+
+  /// \brief Returns if null count is available.
+  virtual bool HasNullCount() const = 0;
```
pitrou, line 58: We don't have a very well-defined policy on casing, but generally simple accessors should be lowercase.
```suggestion
  virtual bool has_null_counts() const = 0;
```

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,143 @@
...
+
+  /// \brief Returns null count for a single page.
+  virtual int64_t null_count(int64_t page_id) const = 0;
+
+  /// \brief Returns all null indicator for each page in batch.
+  virtual const std::vector<bool>& GetNullPages() const = 0;
```
pitrou, line 64:
```suggestion
  virtual const std::vector<bool>& null_pages() const = 0;
```

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,143 @@
...
+
+  /// \brief Returns null count for each page in batch.
+  virtual const std::vector<int64_t>& GetNullCounts() const = 0;
```
pitrou, line 67:
```suggestion
  virtual const std::vector<int64_t>& null_counts() const = 0;
```

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,143 @@
...
+
+  /// \brief The minimum value of a single page. Throws if it is null page.
+  virtual std::string GetEncodedMin(int64_t page_id) const = 0;
```
pitrou, line 70: If the string is persisted in the index, we don't need to copy it on return. So either return a `const std::string&`, or a `std::string_view`.

pitrou: Also, call this `encoded_min`?

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,143 @@
...
+  using T = typename DType::c_type;
+
+  /// \brief The minimum value of a single page. Throws if it is null page.
+  virtual T min_value(int64_t page_id) const = 0;
+
+  /// \brief The maximum value of a single page. Throws if it is null page.
+  virtual T max_value(int64_t page_id) const = 0;
+
+  /// \brief The minimum value of every valid page.
+  virtual const std::vector<T>& GetMinValues() const = 0;
```
pitrou, line 89: `min_values`?

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,143 @@
...
+
+  /// \brief The maximum value of every valid page.
+  virtual const std::vector<T>& GetMaxValues() const = 0;
+
+  /// \brief Returns list of page index of all valid pages.
```
pitrou, line 94: Is it useful to have this in addition from `null_pages`?

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,143 @@
...
+using FLBAColumnIndex = TypedColumnIndex<FLBAType>;
+
+/// \brief PageLocation is a proxy around format::PageLocation.
+struct PARQUET_EXPORT PageLocation {
+  /// File offset of the data page.
+  int64_t offset_;
+  /// Total compressed size of the data page and header.
+  int32_t compressed_page_size_;
+  // row id of the first row in the page within the row group.
+  int64_t first_row_index_;
```
pitrou, line 114: If it's plain value struct, there's no need to suffix the members with an underscore IMHO (there are no methodes where a confusion could arise between local and member).

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,143 @@
...
+  virtual int64_t num_pages() const = 0;
+
+  /// \brief Returns offset for a single page.
+  virtual int64_t offset(int64_t page_id) const = 0;
+
+  /// \brief Returns total compressed size for a single page.
+  virtual int32_t compressed_page_size(int64_t page_id) const = 0;
+
+  /// \brief Returns row id of the first row for a single page.
+  virtual int64_t first_row_index(int64_t page_id) const = 0;
```
pitrou, line 137: Is it useful to have these or can the user just rely on `GetPageLocations`?

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,143 @@
...
+
+  /// \brief Returns all page locations in the offset index.
+  virtual const std::vector<PageLocation>& GetPageLocations() const = 0;
+};
```
pitrou, line 141: Perhaps expose a method that returns the `PageLocation` for a given row id?

pitrou (commented):
Some comments on the API, I didn't look at the implementation.

cpp/src/parquet/page_index.h
wgtmac: I was following the pattern of FileMetaData's ctor: https://github.com/apache/arrow/blob/master/cpp/src/parquet/metadata.h#L277 where FileMetaData needs to know the serialized size of the thrift message. I will change it to input parameter only since here we don't care about the serialized size.

cpp/src/parquet/page_index.h
wgtmac: This makes it easy to know which page_id is valid when the return values from line 92 are used.

cpp/src/parquet/page_index.h
wgtmac: Changed it to a const ref

cpp/src/parquet/page_index.h
wgtmac: Fixed

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,137 @@
...
+  virtual bool is_null_page(int64_t page_id) const = 0;
+
+  /// \brief Returns whether both min_values and max_values are
+  /// orderd and if so, in which direction.
+  virtual BoundaryOrder boundary_order() const = 0;
+
+  /// \brief Returns if null count is available.
+  virtual bool has_null_counts() const = 0;
+
+  /// \brief Returns null count for a single page.
```
emkornfield, line 60: would this throw has_null_count is false?

what do you think about returning optional<int64_t>?

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,137 @@
...
+  virtual int64_t null_count(int64_t page_id) const = 0;
+
+  /// \brief The minimum value of a single page. Throws if it is null page.
+  virtual const std::string& encoded_min(int64_t page_id) const = 0;
```
emkornfield, line 64: what do you think about returning optional<string_view> here?

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,137 @@
...
+
+  /// \brief The maximum value of a single page. Throws if it is null page.
+  virtual const std::string& encoded_max(int64_t page_id) const = 0;
+
+  /// \brief Returns all null indicator for each page in batch.
+  virtual const std::vector<bool>& null_pages() const = 0;
```
emkornfield, line 70: is there a reason these two vectors are exposed here in addition to the accessors above? If these are exposed should the accessors above be defined inline as calls to this function + an accessor?

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,137 @@
...
+  /// \brief Returns null count for a single page.
+  virtual int64_t null_count(int64_t page_id) const = 0;
+
+  /// \brief The minimum value of a single page. Throws if it is null page.
+  virtual const std::string& encoded_min(int64_t page_id) const = 0;
+
+  /// \brief The maximum value of a single page. Throws if it is null page.
+  virtual const std::string& encoded_max(int64_t page_id) const = 0;
+
+  /// \brief Returns all null indicator for each page in batch.
```
emkornfield, line 69:
```suggestion
  /// \brief Returns all null indicator for each page in index.
```
is batch == index?

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,137 @@
...
+  virtual const std::vector<bool>& null_pages() const = 0;
+
+  /// \brief Returns null count for each page in batch.
```
emkornfield, line 72: same question about batch here?

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,137 @@
...
+
+  /// \brief The minimum value of every valid page.
+  virtual const std::vector<T>& min_values() const = 0;
+
+  /// \brief The maximum value of every valid page.
+  virtual const std::vector<T>& max_values() const = 0;
+
+  /// \brief Returns list of page index of all valid pages.
+  /// It can be used to understand values returned from min_values/max_values.
+  virtual std::vector<int64_t> GetValidPageIndices() const = 0;
```
emkornfield, line 96: why not reference here?

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,137 @@
...
+};
+
+/// \brief Typed implementation of ColumnIndex.
+template <typename DType>
+class PARQUET_EXPORT TypedColumnIndex : public ColumnIndex {
+ public:
+  using T = typename DType::c_type;
+
+  /// \brief The minimum value of a single page. Throws if it is null page.
+  virtual T min_value(int64_t page_id) const = 0;
```
emkornfield, line 83: generally same questions above on API:
1.  Should we return optional for values that might not be available?
2. should we be exposing both Vectors and individual element accessors?

cpp/src/parquet/page_index.cc
```diff
@@ -0,0 +1,254 @@
...
+  /// Values that are copied directly from the thrift message.
+  std::vector<bool> null_pages_;
+  std::vector<std::string> encoded_min_values_;
+  std::vector<std::string> encoded_max_values_;
+  BoundaryOrder boundary_order_;
+  bool has_null_count_;
+  std::vector<int64_t> null_counts_;
+
+  /// page_id -> slot_id in the buffer of min_values_ & max_values_
+  std::map<size_t, size_t> page_indexes_;
```
emkornfield, line 168: I guess there is a trade-off but is there  a reason this data-structure is used instead of a std::vector<size_t>, that stores -1 for slots that are all null and the corresponding slot value otherwise?

Also any reason for map vs unordered_map?

cpp/src/parquet/page_index.h
wgtmac: I mean vectorized here.

cpp/src/parquet/page_index.h
wgtmac: I just provide two kind of accessors for caller to choose where appropriate. The suggestion sounds good to make the per-page accessor inline.

cpp/src/parquet/page_index.h
wgtmac: The result is computed on-demand.

cpp/src/parquet/page_index.h
wgtmac: 1. I think returning optional makes sense and it is better than throwing exception.
2. This is a public-facing API, so I'd keep two kinds of accessors for better usability.

cpp/src/parquet/page_index.cc
wgtmac: The trade-off is that we have the chance to save some memory if there is any null page. Using `map` instead of `unordered_map` because GetValidPageIndices() is able to generate ordered page indices simply by iterating the map.

wgtmac:
In the last commit, I have removed all the per-page accessors. They are now actually thin wrappers to the corresponding thrift class. How about this? @emkornfield @pitrou

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,119 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#pragma once
+
+#include "parquet/exception.h"
+#include "parquet/platform.h"
+#include "parquet/schema.h"
```
pitrou, line 22: Suggestion: use forward-declarations for `ColumnDescriptor` and `ReaderProperties` so that you can minimize transitive inclusions in this `.h`.

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,119 @@
...
+  /// \brief Create a ColumnIndex from a serialized thrift message.
+  static std::unique_ptr<ColumnIndex> Make(const ColumnDescriptor& descr,
+                                           const void* serialized_index,
+                                           uint32_t index_len,
+                                           const ReaderProperties& properties);
+
+  virtual ~ColumnIndex() = default;
+
+  /// \brief Returns a list of boolean values to determine the validity of the
+  /// corresponding min and max values.
```
pitrou, line 43:
```suggestion
  /// \brief A bitmap with a bit set for each data page that has only null values.
  ///
  /// The length of this vector is equal to the number of data pages in the column.
```

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,119 @@
...
+  virtual const std::vector<bool>& null_pages() const = 0;
+
+  /// \brief Returns a list of encoded lower bound for the values of each page. For null
+  /// pages the default value is an empty string. Readers must make sure that list entries
+  /// are populated before using them by inspecting null_pages.
```
pitrou, line 48:
```suggestion
  /// \brief A vector of encoded lower bounds for each data page in this column.
  ///
  /// `null_pages` should be inspected first, as only pages with non-null values
  /// may have their lower bounds populated.
```

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,119 @@
...
+  virtual const std::vector<std::string>& encoded_min_values() const = 0;
+
+  /// \brief Returns a list of encoded upper bound for the values of each page. For null
+  /// pages the default value is an empty string. Readers must make sure that list entries
+  /// are populated before using them by inspecting null_pages.
```
pitrou, line 53:
```suggestion
  /// \brief A vector of encoded upper bounds for each data page in this column.
  ///
  /// `null_pages` should be inspected first, as only pages with non-null values
  /// may have their upper bounds populated.
```

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,119 @@
...
+  virtual const std::vector<std::string>& encoded_max_values() const = 0;
+
+  /// \brief Returns whether both min_values and max_values are orderd and if so, in which
+  /// direction.
```
pitrou, line 57:
```suggestion
  /// \brief The ordering of lower and upper bounds.
  ///
  /// The boundary order applies accross all lower bounds, and all upper bounds,
  /// respectively. However, the order between lower bounds and upper bounds
  /// cannot be derived from this.
```

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,119 @@
...
+  virtual BoundaryOrder boundary_order() const = 0;
+
+  /// \brief Returns if null count is available.
```
pitrou, line 60:
```suggestion
  /// \brief Whether per-page null count information is available.
```

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,119 @@
...
+  virtual bool has_null_counts() const = 0;
+
+  /// \brief Returns A list containing the number of null values for each page.
```
pitrou, line 63:
```suggestion
  /// \brief An optional vector with the number of null values in each data page.
  ///
  /// `has_null_counts` should be called first to determine if this information is available.
```

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,119 @@
...
+};
+
+/// \brief Typed implementation of ColumnIndex.
+template <typename DType>
+class PARQUET_EXPORT TypedColumnIndex : public ColumnIndex {
+ public:
+  using T = typename DType::c_type;
+
+  /// \brief Returns a list of lower bound for the values of every non-null page.
+  /// Excluding non-null pages helps binary search if the values are ordered.
```
pitrou, line 74:
```suggestion
  /// \brief A vector of lower bounds for each data page in this column.
  ///
  /// This is like `encoded_min_values`, but with the values decoded according to
  /// the column's physical type.
  /// `min_values` and `max_values` can be used together with `boundary_order`
  /// in order to prune some data pages when searching for specific values.
```

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,119 @@
...
+  virtual const std::vector<T>& min_values() const = 0;
+
+  /// \brief Returns a list of upper bound for the values of every non-null page.
+  /// Excluding non-null pages helps binary search if the values are ordered.
```
pitrou, line 78:
```suggestion
  /// \brief A vector of upper bounds for each data page in this column.
  ///
  /// Just like `min_values`, but for upper bounds instead of lower bounds.
```

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,119 @@
...
+  virtual const std::vector<T>& max_values() const = 0;
+
+  /// \brief Returns a list of page indices for not-null pages. It is helpful to
+  /// understand the original page id in the values returned from min_values()
+  /// and max_values() above.
```
pitrou, line 83: Why expose this `TypedColumnIndex` rather than `ColumnIndex`?

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,119 @@
...
+  virtual const std::vector<int32_t> GetNonNullPageIndices() const = 0;
```
pitrou, line 84:
```suggestion
  virtual const std::vector<int32_t>& non_null_page_indices() const = 0;
```

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,119 @@
...
+using ByteArrayColumnIndex = TypedColumnIndex<ByteArrayType>;
+using FLBAColumnIndex = TypedColumnIndex<FLBAType>;
+
+/// \brief PageLocation is a proxy around format::PageLocation.
+struct PARQUET_EXPORT PageLocation {
+  /// File offset of the data page.
+  int64_t offset;
+  /// Total compressed size of the data page and header.
+  int32_t compressed_page_size;
+  // row id of the first row in the page within the row group.
```
pitrou, line 101:
```suggestion
  /// Row id of the first row in the page within the row group.
```

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,119 @@
...
+ public:
+  /// \brief Create a OffsetIndex from a serialized thrift message.
+  static std::unique_ptr<OffsetIndex> Make(const void* serialized_index,
+                                           uint32_t index_len,
+                                           const ReaderProperties& properties);
+
+  virtual ~OffsetIndex() = default;
+
+  /// \brief Returns all page locations in the offset index.
+  virtual const std::vector<PageLocation>& GetPageLocations() const = 0;
```
pitrou, line 116:
```suggestion
  /// \brief A vector of locations for each data page in this column.
  virtual const std::vector<PageLocation>& page_locations() const = 0;
```

cpp/src/parquet/page_index.cc
```diff
@@ -0,0 +1,174 @@
...
+      : column_index_(column_index) {
+    min_values_.reserve(column_index_.null_pages.size());
+    max_values_.reserve(column_index_.null_pages.size());
+    // Decode min and max values into a compact form (i.e. w/o null page)
+    auto plain_decoder = MakeTypedDecoder<DType>(Encoding::PLAIN, &descr);
+    T value;
+    for (size_t i = 0; i < column_index_.null_pages.size(); ++i) {
+      if (!column_index_.null_pages[i]) {
+        non_null_page_indices_.emplace_back(static_cast<int32_t>(i));
+        Decode<DType>(plain_decoder, column_index_.min_values[i], &value);
```
pitrou, line 60: Is it efficient to decode each value separately, or is it possible to batch-decode all min values at once?

cpp/src/parquet/page_index.cc
```diff
@@ -0,0 +1,174 @@
...
+class TypedColumnIndexImpl : public TypedColumnIndex<DType> {
+ public:
+  using T = typename DType::c_type;
+
+  TypedColumnIndexImpl(const ColumnDescriptor& descr,
+                       const format::ColumnIndex& column_index)
+      : column_index_(column_index) {
+    min_values_.reserve(column_index_.null_pages.size());
+    max_values_.reserve(column_index_.null_pages.size());
+    // Decode min and max values into a compact form (i.e. w/o null page)
```
pitrou, line 54: Hmm, why not, but then `reserve` should be called with the number of non-null data pages (meaning you probably want two separate loops).

cpp/src/parquet/page_index.cc
```diff
@@ -0,0 +1,174 @@
...
+  const std::vector<std::string>& encoded_min_values() const override {
+    return column_index_.min_values;
+  }
+
+  const std::vector<std::string>& encoded_max_values() const override {
+    return column_index_.max_values;
+  }
+
+  BoundaryOrder boundary_order() const override {
+    return static_cast<BoundaryOrder>(static_cast<int>(column_index_.boundary_order));
```
pitrou, line 81: Should use `LoadEnumSafe` here to avoid undefined behavior (UB) on corrupt/malicious data files.

cpp/src/parquet/page_index.cc
```diff
@@ -0,0 +1,174 @@
...
+    case Type::DOUBLE:
+      return std::make_unique<TypedColumnIndexImpl<DoubleType>>(descr, column_index);
+    case Type::BYTE_ARRAY:
+      return std::make_unique<TypedColumnIndexImpl<ByteArrayType>>(descr, column_index);
+    case Type::FIXED_LEN_BYTE_ARRAY:
+      return std::make_unique<TypedColumnIndexImpl<FLBAType>>(descr, column_index);
+    default:
+      break;
+  }
+  DCHECK(false) << "Should not be able to reach this code";
```
pitrou, line 160: Can use `arrow::Unreachable()` from `arrow/util/unreachable.h`.

pitrou (commented):
Thanks for this @wgtmac . Here are some more comments.

cpp/src/parquet/page_index.h
wgtmac: Users cannot simply get decoded min/max values from a generic ColumnIndex. `class Statistics` also have the same issue and thus `class TypedStatistics` is provided to save users from the boilerplate code of decoding min/max values.

cpp/src/parquet/page_index.cc
wgtmac: Unfortunately the encoded min/max values are stored in the thrift object. Since they use PLAIN encoding, I think the actual decoding should be trivial.

cpp/src/parquet/page_index.cc
```diff
@@ -0,0 +1,174 @@
...
+  decoder->SetData(/*num_values=*/1, reinterpret_cast<const uint8_t*>(src.c_str()),
+                   static_cast<int>(src.size()));
+  decoder->Decode(dst, /*max_values=*/1);
+}
+
+template <>
+void Decode<ByteArrayType>(std::unique_ptr<ByteArrayDecoder>&, const std::string& src,
+                           ByteArray* dst) {
+  dst->len = static_cast<uint32_t>(src.size());
+  dst->ptr = reinterpret_cast<const uint8_t*>(src.c_str());
```
emkornfield, line 41: src->data

cpp/src/parquet/page_index.cc
emkornfield: Instead can we remove the default from the switch? or at least throw an exception, crashing seems like the wrong choice if a new enum is added.

cpp/src/parquet/page_index.cc
```diff
@@ -0,0 +1,174 @@
...
+  std::vector<T> max_values_;
+  /// A list of page indices for not-null pages.
+  std::vector<int32_t> non_null_page_indices_;
+};
+
+class OffsetIndexImpl : public OffsetIndex {
+ public:
+  explicit OffsetIndexImpl(const format::OffsetIndex& offset_index) {
+    page_locations_.reserve(offset_index.page_locations.size());
+    for (const auto& page_location : offset_index.page_locations) {
```
emkornfield, line 112: I'm a little bit concerned about the copying around vectors?  How big due we typically expected indexes to be?

cpp/src/parquet/page_index.cc
wgtmac: > Instead can we remove the default from the switch? or at least throw an exception, crashing seems like the wrong choice if a new enum is added.

Agreed. In this way, the developer gets noticed there is something to fix at compile time instead of run time.

cpp/src/parquet/page_index.cc
wgtmac: The size depends on the number of pages in a column chunk. If we don't copy them and instead wrap a thrift OffsetIndex inside the implementation class, then it pays to get a vector of PageLocations (as we cannot expose thrift internals). 

I can refactor it to remove the vector accessor by providing only the per-page accessors. Then pimpl can be utilized to remove the virtual functions to reduce the cost for every access.

cpp/src/parquet/page_index.h
wgtmac: Now only "parquet/types.h" is included in the header file.

cpp/src/parquet/page_index.h
wgtmac: Fixed

cpp/src/parquet/page_index.h
wgtmac: Fixed

cpp/src/parquet/page_index.h
wgtmac: Fixed

cpp/src/parquet/page_index.h
wgtmac: Fixed

cpp/src/parquet/page_index.h
wgtmac: Fixed

cpp/src/parquet/page_index.h
wgtmac: Fixed

cpp/src/parquet/page_index.h
wgtmac: Fixed

cpp/src/parquet/page_index.h
wgtmac: Fixed

cpp/src/parquet/page_index.h
wgtmac: Fixed

cpp/src/parquet/page_index.cc
wgtmac: Fixed

cpp/src/parquet/page_index.cc
wgtmac: Fixed

cpp/src/parquet/page_index.cc
```diff
@@ -0,0 +1,174 @@
...
+            const std::string& src, typename DType::c_type* dst) {
+  decoder->SetData(/*num_values=*/1, reinterpret_cast<const uint8_t*>(src.c_str()),
+                   static_cast<int>(src.size()));
+  decoder->Decode(dst, /*max_values=*/1);
+}
+
+template <>
+void Decode<ByteArrayType>(std::unique_ptr<ByteArrayDecoder>&, const std::string& src,
+                           ByteArray* dst) {
+  dst->len = static_cast<uint32_t>(src.size());
```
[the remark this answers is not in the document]
wgtmac: Added a DCHECK here.

cpp/src/parquet/page_index.cc
```diff
@@ -0,0 +1,188 @@
...
+            const std::string& src, typename DType::c_type* dst) {
+  decoder->SetData(/*num_values=*/1, reinterpret_cast<const uint8_t*>(src.c_str()),
+                   static_cast<int>(src.size()));
+  decoder->Decode(dst, /*max_values=*/1);
+}
+
+template <>
+void Decode<ByteArrayType>(std::unique_ptr<ByteArrayDecoder>&, const std::string& src,
+                           ByteArray* dst) {
+  DCHECK_LE(src.size(), static_cast<size_t>(std::numeric_limits<uint32_t>::max()));
```
pitrou, line 46: Since this depends on the file's correctness, this should raise an exception instead of asserting out.

cpp/src/parquet/page_index.cc
```diff
@@ -0,0 +1,188 @@
...
+    if (num_non_null_pages != 0U) {
+      min_values_.reserve(num_non_null_pages);
+      max_values_.reserve(num_non_null_pages);
+      // Decode min and max values into a compact form (i.e. w/o null page)
+      auto plain_decoder = MakeTypedDecoder<DType>(Encoding::PLAIN, &descr);
+      T value;
+      for (size_t i = 0; i < column_index_.null_pages.size(); ++i) {
+        if (!column_index_.null_pages[i]) {
+          non_null_page_indices_.emplace_back(static_cast<int32_t>(i));
+          Decode<DType>(plain_decoder, column_index_.min_values[i], &value);
```
pitrou, line 73: This seems to assume that `min_values` and `max_values` have the same length as `null_pages`. What if it's not the case? We should detect this situation and raise an exception.

cpp/src/parquet/page_index.cc
```diff
@@ -0,0 +1,188 @@
...
+                       const format::ColumnIndex& column_index)
+      : column_index_(column_index) {
+    size_t num_non_null_pages = std::accumulate(
+        column_index_.null_pages.cbegin(), column_index_.null_pages.cend(), 0U,
+        [](size_t num_non_null_pages, bool null_page) {
+          return num_non_null_pages + (null_page ? 0U : 1U);
+        });
+    if (num_non_null_pages != 0U) {
+      min_values_.reserve(num_non_null_pages);
+      max_values_.reserve(num_non_null_pages);
```
pitrou, line 66: Can also reserve on `non_null_page_indices_`.

cpp/src/parquet/page_index.cc
```diff
@@ -0,0 +1,188 @@
...
+      // Decode min and max values into a compact form (i.e. w/o null page)
+      auto plain_decoder = MakeTypedDecoder<DType>(Encoding::PLAIN, &descr);
+      T value;
+      for (size_t i = 0; i < column_index_.null_pages.size(); ++i) {
+        if (!column_index_.null_pages[i]) {
+          non_null_page_indices_.emplace_back(static_cast<int32_t>(i));
```
pitrou, line 72: Before downcasting here we should probably check above that the size doesn't go above the limit (unlikely, but who knows).

pitrou:
I just realized: in this API, `encoded_min_values` and `min_values` use different conventions? (one is dense, the other is sparse).

cpp/src/parquet/page_index.h
```diff
@@ -0,0 +1,132 @@
...
+  /// \brief A vector of upper bounds for each data page in this column.
+  ///
+  /// Just like `min_values`, but for upper bounds instead of lower bounds.
+  virtual const std::vector<T>& max_values() const = 0;
+
+  /// \brief A vector of page indices for not-null pages.
+  ///
+  /// It is helpful to understand the original page id in the values returned from
+  /// min_values() and max_values() above.
+  virtual const std::vector<int32_t>& non_null_page_indices() const = 0;
```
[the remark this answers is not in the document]
wgtmac: Done

wgtmac:
> I just realized: in this API, `encoded_min_values` and `min_values` use different conventions? (one is dense, the other is sparse).

After checking the potential use cases of `ColumnIndex`, I think it makes sense to follow same convention for `encoded_min_values()` and `min_values()` to have same number of pages. For predicate push down, the binary search for sorted column index can be simply done by iterating page_ids returned from `non_null_page_indices()`. The latest commit reflects this change. Please take a look, thanks! @pitrou

cpp/src/parquet/page_index.cc
emkornfield: i guess it was a general concern for both I just mentioned it here, @pitrou if you feel comfortable with it it seems OK to me.

cpp/src/parquet/page_index.cc
```diff
@@ -0,0 +1,229 @@
...
+      throw ParquetException("Invalid column index");
+    }
+
+    size_t num_pages = column_index_.null_pages.size();
+    size_t num_non_null_pages = static_cast<size_t>(std::accumulate(
+        column_index_.null_pages.cbegin(), column_index_.null_pages.cend(), 0,
+        [](int32_t num_non_null_pages, bool null_page) {
+          return num_non_null_pages + (null_page ? 0 : 1);
+        }));
+    DCHECK_LE(num_non_null_pages, num_pages);
```
emkornfield, line 102: this should probably be an exception? it seems the data structures are independent?

wgtmac: `num_non_null_pages` is computed from `column_index_.null_pages` vector and `num_pages` is the size of that vector. `num_non_null_pages` is guaranteed to be less than or equal to `num_pages` and IMHO it is not worth a runtime check. I simply put a debug check here just in case.

wgtmac:
> Thanks @wgtmac . Is it possible to add some tests for this? One option is to rely on data files from https://github.com/apache/parquet-testing/tree/master/data (you might also upload your own there if desired)

I have added `page_index_test.cc` to test deserialization of page index objects. Please take a look, thanks!

cpp/src/parquet/page_index.cc
```diff
@@ -0,0 +1,229 @@
...
+#include <numeric>
+
+namespace parquet {
+
+namespace {
+
+template <typename DType>
+void Decode(std::unique_ptr<typename EncodingTraits<DType>::Decoder>& decoder,
+            const std::string& input, std::vector<typename DType::c_type>& output,
+            size_t index) {
```
pitrou, line 38: I'd rather keep the former signature
```suggestion
void Decode(std::unique_ptr<typename EncodingTraits<DType>::Decoder>& decoder,
            const std::string& input, typename DType::c_type* out) {
```

cpp/src/parquet/page_index.cc
```diff
@@ -0,0 +1,229 @@
...
+                       const format::ColumnIndex& column_index)
+      : column_index_(column_index) {
+    // Make sure the number of pages is valid and it does not overflow to int32_t.
+    if (column_index_.null_pages.size() != column_index_.min_values.size() ||
+        column_index_.min_values.size() != column_index_.max_values.size() ||
+        ARROW_PREDICT_FALSE(column_index_.null_pages.size() >=
+                            static_cast<size_t>(std::numeric_limits<int32_t>::max()))) {
+      throw ParquetException("Invalid column index");
+    }
+
```
pitrou, line 95: Can we also check the size of `null_counts` if present?

cpp/src/parquet/page_index_test.cc
```diff
@@ -0,0 +1,208 @@
...
+  PARQUET_ASSIGN_OR_THROW(source, ::arrow::io::ReadableFile::Open(path));
+  PARQUET_ASSIGN_OR_THROW(auto buffer,
+                          source->ReadAt(index_location->offset, index_location->length));
+  PARQUET_THROW_NOT_OK(source->Close());
+
+  // Deserialize column index.
+  auto properties = default_reader_properties();
+  auto descr = file_metadata->schema()->Column(column_id);
+  std::unique_ptr<ColumnIndex> column_index = ColumnIndex::Make(
+      *descr, buffer->data(), static_cast<uint32_t>(buffer->size()), properties);
```
pitrou, line 108: Can we perhaps add an API to `ColumnReader` or `RowGroupReader` that would fetch the `ColumnIndex` of a column?

(if so, then perhaps do the same for `OffsetIndex`)

cpp/src/parquet/page_index_test.cc
```diff
@@ -0,0 +1,208 @@
...
+
+  // Verify only partial data as there are too many pages.
+  ASSERT_EQ(num_pages, column_index->null_pages().size());
+  ASSERT_EQ(has_null_counts, column_index->has_null_counts());
+  ASSERT_EQ(boundary_order, column_index->boundary_order());
+  for (size_t i = 0; i < page_indices.size(); ++i) {
+    size_t page_id = page_indices.at(i);
+    ASSERT_EQ(null_pages.at(i), column_index->null_pages().at(page_id));
+    if (has_null_counts) {
+      ASSERT_EQ(null_counts.at(i), column_index->null_counts().at(page_id));
```
pitrou, line 120: This is a duplicate line.

cpp/src/parquet/page_index_test.cc
```diff
@@ -0,0 +1,208 @@
...
+    }
+  }
+}
+
+TEST(PageIndex, ReadInt64ColumnIndex) {
+  const int column_id = 5;
+  const size_t num_pages = 528;
+  const BoundaryOrder::type boundary_order = BoundaryOrder::Unordered;
+  const std::vector<size_t> page_indices = {0, 99, 426, 520};
+  const std::vector<bool> null_pages = {false, false, false, false};
```
pitrou, line 143: Is there any case in the example file where one of these values is true?

cpp/src/parquet/page_index_test.cc
```diff
@@ -0,0 +1,208 @@
...
+  const bool has_null_counts = true;
+  const std::vector<int64_t> null_counts = {0, 0, 0, 0};
```
pitrou, line 145: Similarly, is there any case where a page has some null entries?

cpp/src/parquet/page_index_test.cc
```diff
@@ -0,0 +1,208 @@
...
+  const bool has_null_counts = true;
+  const std::vector<int64_t> null_counts = {0, 0, 0};
+  const std::vector<bool> min_values = {false, false, false};
+  const std::vector<bool> max_values = {true, true, true};
+
+  TestReadTypedColumnIndex<BooleanType>(column_id, num_pages, boundary_order,
+                                        page_indices, null_pages, min_values, max_values,
+                                        has_null_counts, null_counts);
+}
+
```
pitrou, line 207: Is there also a fixed-length byte array column that we can test?

pitrou (commented):
Nice tests, thank you! Some additional comments below.

cpp/src/parquet/page_index.cc
wgtmac: Same here. Unfortunately the `std::vector<bool>` is a special case which cannot alter its values in place. That's why I have changed the signature.

cpp/src/parquet/page_index_test.cc
wgtmac: That is exactly in the plan of next patch. :)

cpp/src/parquet/page_index.cc
pitrou: Hmm, I see. At least we should avoid non-const references, as per the style guide, so perhaps:
```suggestion
void Decode(std::unique_ptr<typename EncodingTraits<DType>::Decoder>& decoder,
            const std::string& input, std::vector<typename DType::c_type>* output,
            size_t output_index) {
```

cpp/src/parquet/page_index_test.cc
pitrou: You can generate such a file with another writer and add it to `parquet-testing` perhaps? That would ensure we're compatible with other implementations (mere roundtripping doesn't guarantee that).

cpp/src/parquet/page_index_test.cc
wgtmac: Sorry I don't understand why it is duplicate. This line checks the `null_counts` vector when has_null_counts is true. This is different from two lines above which checks `null_pages ` vector.

cpp/src/parquet/page_index_test.cc
wgtmac: Yes, I found one file with null_page = true. Will add it to a new case.

cpp/src/parquet/page_index_test.cc
pitrou: Woops, I hadn't read carefully enough, sorry...

cpp/src/parquet/page_index_test.cc
wgtmac: I did some research and found that FIXED_LENGTH_BYTE_ARRAY type is only used for `interval` and `decimal` logical types. For `interval` type, statistics is disabled by the spec. For `decimal` type, `INT32`, `INT64` and `BYTE_ARRAY` physical types are used by default in the `parquet-mr`, I cannot simply use `parquet-cli` command line to generate the file. Will look into the test cases to find some examples.

mapleFU:
@wgtmac Seems that arrow uses git submodules to manage the `parquet-testing`, and maybe you can download submodules, checkout to lastest version and commit

pitrou:
@wgtmac Feel free to ping when this is ready for review again.

cpp/src/parquet/page_index.cc
wgtmac: Fixed

cpp/src/parquet/page_index_test.cc
wgtmac: Add test case `FixedLengthByteArrayColumnIndex`

cpp/src/parquet/page_index_test.cc
wgtmac: Added test case `FixedLengthByteArrayColumnIndex`

wgtmac (commented):
I think the unit test is complete now. Please take a look when you get the chance. Thanks! @pitrou

cpp/src/parquet/page_index_test.cc
```diff
@@ -0,0 +1,260 @@
...
+  for (size_t i = 0; i < page_indices.size(); ++i) {
+    size_t page_id = page_indices.at(i);
+    ASSERT_EQ(null_pages.at(i), column_index->null_pages().at(page_id));
+    if (has_null_counts) {
+      ASSERT_EQ(null_counts.at(i), column_index->null_counts().at(page_id));
+    }
+    // min/max values are only meaningful for non-null pages.
+    if (!null_pages.at(i)) {
+      if constexpr (std::is_same_v<T, double>) {
+        ASSERT_DOUBLE_EQ(min_values.at(i), typed_column_index->min_values().at(page_id));
```
pitrou, line 128: This doesn't look right. In the latest version of the PR, `min_values` and `max_values` are dense, i.e. they only contain entries for non-null pages. So you shouldn't be looking at `min_values().at(page_id)` but `min_values().at(non_null_page_index)`.

This also means that none of your tests has both null and non-null pages.

cpp/src/parquet/page_index_test.cc
```diff
@@ -0,0 +1,260 @@
...
+  const int column_id = 0;
+  const size_t num_pages = 2;
+  const BoundaryOrder::type boundary_order = BoundaryOrder::Ascending;
+  const std::vector<size_t> page_indices = {0, 1};
+  const std::vector<bool> null_pages = {true, true};
+  // It seems that the null_counts are malformed.
+  const bool has_null_counts = true;
+  const std::vector<int64_t> null_counts = {-1, -1};
+  const std::vector<int32_t> min_values = {};
+  const std::vector<int32_t> max_values = {};
```
pitrou, line 253: Seems like a bad example to be honest. Can you find or generate a test case which has both null and non-null pages in a single column?

cpp/src/parquet/page_index_test.cc
```diff
@@ -0,0 +1,260 @@
...
+
+  TestReadTypedColumnIndex<Int32Type>(
+      "datapage_v1-uncompressed-checksum.parquet", column_id, num_pages, boundary_order,
```
pitrou, line 256: If you read this file, you'll see that all data is actually non-null. So the column index in this file may be bogus?
(or you're reading it wrong?)

pitrou (changes requested):
I pushed a couple of minor changes, but it seems the tests are still lacking a bit. See comments.

cpp/src/parquet/page_index.cc
```diff
@@ -0,0 +1,234 @@
...
+    const size_t num_non_null_pages = static_cast<size_t>(std::accumulate(
+        column_index_.null_pages.cbegin(), column_index_.null_pages.cend(), 0,
+        [](int32_t num_non_null_pages, bool null_page) {
+          return num_non_null_pages + (null_page ? 0 : 1);
+        }));
+    DCHECK_LE(num_non_null_pages, num_pages);
+
+    // Allocate slots for decoded values.
+    min_values_.resize(num_non_null_pages);
```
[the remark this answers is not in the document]
pitrou: Sorry, I was reading it wrong again. My bad!

cpp/src/parquet/page_index_test.cc
wgtmac: There isn't. Most of the files are even created without page index. Sigh :(

pitrou: But why not generate a test file yourself? There's probably an implementation that would allow to do it?

cpp/src/parquet/page_index_test.cc
wgtmac: I think the page index is corrupted. The parquet-cli has provided same reading:

```
➜ parquet-cli column-index datapage_v1-uncompressed-checksum.parquet
row-group 0:
column index for column a:
Boundary order: ASCENDING
                      null count  min                                       max
page-0                        -1  <none>                                    <none>
page-1                        -1  <none>                                    <none>

offset index for column a:
                          offset   compressed size       first row index
page-0                         4             10268                     0
page-1                     10272             10268                  2560

column index for column b:
Boundary order: ASCENDING
                      null count  min                                       max
page-0                        -1  <none>                                    <none>
page-1                        -1  <none>                                    <none>

offset index for column b:
                          offset   compressed size       first row index
page-0                     20540             10268                     0
page-1                     30808             10268                  2560
```

pitrou (approved):

pitrou:
Thanks a lot! I think this is good to go now, will just wait for CI.

PR #14806, opened 2022-12-01
ARROW-18421: [C++][ORC] Add accessor for stripe information in reader

LouisClt:
See https://issues.apache.org/jira/browse/ARROW-18421


cpp/src/arrow/adapters/orc/adapter.cc
```diff
@@ -217,6 +217,8 @@ class ORCFileReader::Impl {
 
   int64_t NumberOfRows() { return static_cast<int64_t>(reader_->getNumberOfRows()); }
 
+  int64_t NumberOfRows(int64_t stripe) { return stripes_[stripe].num_rows; }
```
wgtmac, line 220: Should we rename it to `NumberOfRowsInStripe` to avoid unintentional overloading?

wgtmac: In addition, it is better to validate `stripe` and throw if out of range.

LouisClt: Well maybe it will be clearer to rename it NumberOfRowsInStripe indeed.
What kind of exception would you find appropriate, `liborc::InvalidArgument` ?

wgtmac: It seems that the orc adapter already throws exceptions from liborc. Maybe it is good enough to throw `liborc::InvalidArgument` for now.

wgtmac: @kou Can you advise if we need to be more consistent to the project style which returns status instead of throwing exception?

pitrou: We shouldn't throw exceptions in Arrow. Also, I don't think there's a need for checking the input value: the number of stripes is readily available.

pitrou:
Why not expose the full stripe information? We don't want to create tons of little accessors for each different piece of information.

pitrou:
We can change it slightly and expose it as:
```c++
struct StripeInformation {
  int64_t offset;
  int64_t length;
  int64_t num_rows;
  int64_t first_row_id;
};
```
(note the use of signed integers to match our API conventions better)

LouisClt:
Ok, that seems fine to me. I will make the changes then.
A prototype like this would be OK ?
```cpp
StripeInformation GetStripeInformation(int64_t stripe);
```

pitrou:
Yes, that sounds fine.

LouisClt:
So this seems to be good now. I added the "StripeInformation" class in the header, and modified the implementation accordingly.
I added one more small test on that.

pitrou (approved):
+1, thank you @LouisClt

PR #14808, opened 2022-12-01
GH-14805: [Format] C Data Interface: clarify nullability of buffer pointers

### LINKED ISSUE #14805: [Format] C Data Interface: clarify nullability of buffer pointers

In https://github.com/apache/arrow-nanoarrow/issues/76 it was reported that some implementations may export zero-length buffers as a null pointer in the ArrowArray structure.

Given that this seems to be an accepted behavior in several implementations, we should probably make it explicitly legal in the C Data Interface spec.

### Component(s)

Format

pitrou:
* Closes: #14805


pitrou:

lidavidm (approved):

PR #14814, opened 2022-12-01
GH-14875: [C++] C Data Interface: check imported buffer for non-null

### LINKED ISSUE #14875: [Python][C++] C Data Interface incorrect validate failures

Spinning off from #14814: 

When testing round trips of empty arrays between Python and Go using the C Data Interface, I found an issue with the binary and string data type arrays.

The data types: `pa.binary()`, `pa.large_binary()`, `pa.string()`, `pa.large_string()` all throw an error when calling `validate(full=True)` after the `_import_from_c` that contained a null value data buffer:

```
  File "/home/zeroshade/Projects/GitHub/arrow/go/arrow/cdata/test/test_export_to_cgo.py", line 218, in test
    b.validate(full=True)
  File "pyarrow/array.pxi", line 1501, in pyarrow.lib.Array.validate
pyarrow.lib.ArrowInvalid: Value data buffer is null
```

Following up from #14805 clarifying that buffers can be null in a 0-length array. My guess here is rather than the offsets buffer, the issue is the second data buffer which would contain the actual binary/utf-8 data if the array had a length >0. But that's just a theory, I haven't confirmed it.

### Component(s)

C++, Python

pitrou:
The C data interface may expose null data pointers for zero-sized buffers.
Make sure that all buffer pointers remain non-null internally.

Followup to GH-14805

* Closes: #14875

### DIFF
--- a/cpp/src/arrow/array/validate.cc
+++ b/cpp/src/arrow/array/validate.cc
@@ -459,14 +459,17 @@ struct ValidateArrayImpl {
       if (buffer == nullptr) {
         continue;
       }
-      int64_t min_buffer_size = -1;
+      int64_t min_buffer_size = 0;
       switch (spec.kind) {
         case DataTypeLayout::BITMAP:
-          min_buffer_size = bit_util::BytesForBits(length_plus_offset);
+          // If length == 0, buffer size can be 0 regardless of offset
+          if (data.length > 0) {
+            min_buffer_size = bit_util::BytesForBits(length_plus_offset);
+          }
           break;
         case DataTypeLayout::FIXED_WIDTH:
-          if (MultiplyWithOverflow(length_plus_offset, spec.byte_width,
-                                   &min_buffer_size)) {
+          if (data.length > 0 && MultiplyWithOverflow(length_plus_offset, spec.byte_width,
+                                                      &min_buffer_size)) {
             return Status::Invalid("Array of type ", type.ToString(),
                                    " has impossibly large length and offset");
           }
--- a/cpp/src/arrow/c/bridge.cc
+++ b/cpp/src/arrow/c/bridge.cc
@@ -1544,13 +1544,18 @@ struct ArrayImporter {
 
   Status ImportBitsBuffer(int32_t buffer_id, bool is_null_bitmap = false) {
     // Compute visible size of buffer
-    int64_t buffer_size = bit_util::BytesForBits(c_struct_->length + c_struct_->offset);
+    int64_t buffer_size =
+        (c_struct_->length > 0)
+            ? bit_util::BytesForBits(c_struct_->length + c_struct_->offset)
+            : 0;
     return ImportBuffer(buffer_id, buffer_size, is_null_bitmap);
   }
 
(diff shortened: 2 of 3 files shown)


pitrou:
@zeroshade  @alippai @alamb @viirya @tustvold Would it be easy to try this against the Go and Rust Arrow implementations?

alamb:
Sorry @pitrou  I am not sure how to test this against Rust. Did you have anything specific have in mind beyond the existing integration CI check https://github.com/apache/arrow/actions/runs/3595507889/jobs/6055082178 ?

pitrou:
@alamb AFAIR there are Rust-PyArrow integration tests through the C Data Interface, no?

wjones127:
@pitrou I think we just need to add a test case to [this file](https://github.com/apache/arrow-rs/blob/master/arrow-pyarrow-integration-testing/src/lib.rs) roundtripping an empty array, and then test that against PyArrow built off of this branch, correct?

If so, I can do that.

pitrou:
@wjones127 That would probably work, yes.

wjones127:
I added tests here: https://github.com/apache/arrow-rs/pull/3276

And locally tested them against PyArrow built from this branch. All tests pass 👍

zeroshade:
@pitrou I'll build a pyarrow off this branch and test a roundtrip of an empty array between pyarrow and Go via C data. I'll comment again after I set it up.

zeroshade:
@pitrou when I do round trips of the empty arrays, after fixing the crash in Go, I run into a few issues with pyarrow upon importing the empty array from C:

The data types: `pa.binary()`, `pa.large_binary()`, `pa.string()`, `pa.large_string()` all throw an error when calling `validate(full=True)` after the `_import_from_c` that contained a null value data buffer:

```
  File "/home/zeroshade/Projects/GitHub/arrow/go/arrow/cdata/test/test_export_to_cgo.py", line 218, in test
    b.validate(full=True)
  File "pyarrow/array.pxi", line 1501, in pyarrow.lib.Array.validate
pyarrow.lib.ArrowInvalid: Value data buffer is null
```

I also ran into other cases where it through an error in `_import_from_c` because of null values but I think those are my fault, I'm going to double check and try again. But I'm pretty sure those validate's shouldn't be failing in this case if we're saying it's acceptable for a 0-length array to contain a null value data buffer.

pitrou:
@zeroshade Nice catch, can you open an issue?

zeroshade:
I've opened #14877 to fix the Go side of these, though some of the python tests have to be commented out until #14875 is resolved.

pitrou:
@zeroshade This draft PR now passes the uncommented Go tests (assuming you install the corresponding PyArrow version).

zeroshade:
@pitrou if this passes the tests now, would it make sense to uncomment them in this PR? Or should we do that in a separate PR?

pitrou:
@zeroshade We can't uncomment them until a PyArrow wheel is released that incorporates the fix, AFAIU.

pitrou:
@paleolimbot FYI

lidavidm (approved):

PR #14818, opened 2022-12-02
ARROW-17374: [C++] Snappy package may be built without CMAKE_BUILD_TYPE

kou:
For example, conda's Snappy package is built without CMAKE_BUILD_TYPE.


kou (commented):
+1

PR #14827, opened 2022-12-02
ARROW-18422: [C++] Delete EnumTraits, use enum reflection utilities

bkietz:
Trying again with ARROW-13296 now that we have c++17


cpp/src/arrow/util/reflection_internal.h
```diff
@@ -91,21 +92,260 @@ PropertyTuple<Properties...> MakeProperties(Properties... props) {
...
+struct DefaultEnumMembers<Enum, Limits, Max, Max, Members...> {
+  using members = sequence<Members...>;
+};
+
+template <typename Enum, typename Limits, int I, int Max, Enum... Members>
+struct DefaultEnumMembers
+    : std::conditional_t<
+          IsValidEnumMember<Enum, I>(),
+          DefaultEnumMembers<Enum, Limits, I + 1, Max, Members..., static_cast<Enum>(I)>,
+          DefaultEnumMembers<Enum, Limits, I + 1, Max, Members...>> {};
```
pitrou, line 266: Hmm... this seems to lead to O(N) compile-time complexity where N is `std::numeric_limits<std::underlying_type_t<Enum>>::max() - std::numeric_limits<std::underlying_type_t<Enum>>::min()`. That sounds concerning, especially if compile-time evaluation is slow (which it may very well be, depending on the compiler).

bkietz: That's why this is limited to enumerations which are only one byte wide, so we'll do a maximum of 256 iterations per enum. I wouldn't expect that to have a heavy compile time cost

pitrou:
Congrats for making this work, but... this seems uncomfortably heroic. I would really like to use automated introspection facilities instead of specializing EnumTraits by hand, but not at the price of relying on such primitives.

@lidavidm Thoughts?

cpp/src/arrow/util/reflection_internal.h
```diff
@@ -91,21 +92,260 @@ PropertyTuple<Properties...> MakeProperties(Properties... props) {
...
+
+  constexpr explicit array(const T (&arr)[N])  // NOLINT
+      : array{static_cast<const T*>(arr)} {}
+
+  [[nodiscard]] static constexpr size_t size() { return N; }
+  [[nodiscard]] constexpr const T* data() const { return values_; }
+  [[nodiscard]] constexpr T* data() { return values_; }
+  [[nodiscard]] constexpr const T* begin() const { return values_; }
+  [[nodiscard]] constexpr const T* end() const { return begin() + N; }
+  [[nodiscard]] constexpr const T& operator[](size_t i) const { return begin()[i]; }
```
edponce, line 179: why two different idioms for indexing?

lidavidm:
This is really cool to see, but I would also prefer boilerplate over the magic here. I'm also worried about whether this might affect compatibility with future compiler versions, if the representation of those vendor extensions changes.

PR #14831, opened 2022-12-02
ARROW-18423: [Python] Expose reading a schema from an IPC message

ankoh:
Jira Issue: https://issues.apache.org/jira/browse/ARROW-18423

### DIFF
--- a/python/pyarrow/tests/test_ipc.py
+++ b/python/pyarrow/tests/test_ipc.py
@@ -795,6 +795,12 @@ def test_message_read_from_compressed(example_messages):
         assert result.equals(message)
 
 
+def test_message_read_schema(example_messages):
+    batches, messages = example_messages
+    schema = pa.ipc.read_schema(messages[0])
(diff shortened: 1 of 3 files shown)


python/pyarrow/tests/test_ipc.py
```diff
@@ -795,6 +795,12 @@ def test_message_read_from_compressed(example_messages):
         assert result.equals(message)
 
 
+def test_message_read_schema(example_messages):
+    batches, messages = example_messages
+    schema = pa.ipc.read_schema(messages[0])
+    schema.equals(batches[1].schema)
```
pitrou, line 801: Should probably be
```suggestion
    assert schema.equals(batches[1].schema)
```

pitrou (changes requested):
Thanks @ankoh . This is a reasonable addition, I just have one suggestion.

pitrou (approved):
+1, LGTM. The CI failures do not look related.

PR #14102, opened 2022-09-12
ARROW-17635: [Python][CI] Sync conda recipe with the arrow-cpp feedstock

h-vetinari:
Corresponds to status of feedstock as of https://github.com/conda-forge/arrow-cpp-feedstock/pull/848, minus obvious & intentional divergences in the setup here (with the exception of unpinning xsimd, which was [pinned](https://github.com/apache/arrow/blob/apache-arrow-9.0.0/cpp/cmake_modules/ThirdpartyToolchain.cmake#L2245) as of 9.0.0, but isn't anymore).

PR #14832, opened 2022-12-04
GH-14828: [CI][Conda] Sync with conda-forge, fix nightly jobs

### LINKED ISSUE #14828: [CI][Conda] Nightly CI jobs aren't maintained

Nightly CI jobs for Conda were fixed by @h-vetinari and @xhochy in GH-14102 but most nightly CI jobs are failing again this past month.

Can we maintain them? If we can't maintain them, can we remove them? Are they useful to maintain https://github.com/conda-forge/arrow-cpp-feedstock ?

http://crossbow.voltrondata.com/

Task Name | Since Last Successful Build | Last Successful Commit | Last Successful Build | First Failure | 9 Days Ago | 8 Days Ago | 7 Days Ago | 6 Days Ago | 5 Days Ago | 4 Days Ago | 3 Days Ago | 2 Days Ago | Most Recent Failure
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
conda-win-vs2019-py37-r40 | 78 days | 5e6da78 | pending | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-win-vs2019-py38 | 65 days | 60c9383 | pending | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-linux-gcc-py310-arm64 | 29 days | 8e3a1e1 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-linux-gcc-py310-cpu | 29 days | 8e3a1e1 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-linux-gcc-py37-arm64 | 29 days | 8e3a1e1 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-linux-gcc-py37-cpu-r40 | 29 days | 8e3a1e1 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-linux-gcc-py37-cpu-r41 | 29 days | 8e3a1e1 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-linux-gcc-py37-cuda | 29 days | 8e3a1e1 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-linux-gcc-py37-ppc64le | 29 days | 8e3a1e1 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-linux-gcc-py38-arm64 | 29 days | 8e3a1e1 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-linux-gcc-py38-cpu | 29 days | 8e3a1e1 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-linux-gcc-py38-cuda | 29 days | 8e3a1e1 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-linux-gcc-py38-ppc64le | 29 days | 8e3a1e1 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-linux-gcc-py39-arm64 | 29 days | 8e3a1e1 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-linux-gcc-py39-cpu | 29 days | 8e3a1e1 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-linux-gcc-py39-cuda | 29 days | 8e3a1e1 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-linux-gcc-py39-ppc64le | 29 days | 8e3a1e1 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-osx-arm64-clang-py38 | 29 days | 8e3a1e1 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-osx-arm64-clang-py39 | 29 days | 8e3a1e1 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-osx-clang-py37-r40 | 29 days | 8e3a1e1 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-osx-clang-py37-r41 | 29 days | 8e3a1e1 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-osx-clang-py38 | 29 days | 8e3a1e1 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-osx-clang-py39 | 29 days | 8e3a1e1 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-linux-gcc-py310-cuda | 15 days | 501b799 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-linux-gcc-py310-ppc64le | 15 days | 501b799 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-osx-arm64-clang-py310 | 15 days | 501b799 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail
conda-osx-clang-py310 | 15 days | 501b799 | pass | fail | fail | fail | fail | fail | fail | fail | fail | fail | fail

### Component(s)

Continuous Integration, Packaging

h-vetinari:
Synching after https://github.com/conda-forge/arrow-cpp-feedstock/pull/875, which does quite a lot of things, see this [summary](https://github.com/conda-forge/arrow-cpp-feedstock/pull/875#pullrequestreview-1197658827). I'm not keeping the commit history here, but it might be instructive to check the commits there to see why certain changes came about.

It also fixes the CI that was broken by https://github.com/apache/arrow/commit/a3ef64b9583c3a92fac73e74e9e03e5d22a6d8f6 (undoing the changes of #14102 in `tasks.yml`).

Finally, it adapts to conda making a long-planned [switch](https://github.com/conda-forge/conda-forge.github.io/issues/1586) w.r.t. to the format / extension of the artefacts it produces.

I'm very likely going to need some help (or at least pointers) for the R-stuff. CC @xhochy
(for context, I never got a response to https://github.com/conda-forge/r-arrow-feedstock/pull/55, but I'll open a PR to build against libarrow 10).

Once this is done, I can open issues to tackle the tests that shouldn't be failing, resp. the segfaults on PPC resp. in conjunction with `sparse`.
* Closes: #14828


h-vetinari:
I have a tentative PR for r-arrow, but it's going to need some more work: https://github.com/conda-forge/r-arrow-feedstock/pull/58

kou:
@github-actions crossbow submit -g conda

h-vetinari:
OK, I think I cracked the r-issues in https://github.com/conda-forge/r-arrow-feedstock/pull/58, and updated things correspondingly here.

Also:
* avoid patch (line-ending issues when applying on windows), patch `gcs_server` fixture directly see https://github.com/apache/arrow/pull/14832/commits/759e7050d452539f58eabdb522028bc58421e6cb
* add r-jobs for linux-aarch64 & osx-arm64 (these were in fact broken by https://github.com/apache/arrow/commit/8139cdd3af4531fdabe6b3ebeff0a9cd30b7d16e and need `${R_ARGS}` to pass in cross-compilation)
* homogenize naming of conda jobs, see https://github.com/apache/arrow/pull/14832/commits/d08195784ffc1747458e36a779379800879ac3b3

h-vetinari:
@github-actions crossbow submit -g conda

h-vetinari:
@github-actions crossbow submit -g conda

python/pyarrow/tests/conftest.py
```diff
@@ -184,9 +184,16 @@ def gcs_server():
     args = [sys.executable, '-m', 'testbench', '--port', str(port)]
     proc = None
     try:
+        # check first if testbench module is available
+        import testbench  # noqa:F401
```
h-vetinari, line 188: Not doing this adds 30+ min to the CI runs, because (on conda-forge CI, with all tests), because each test using `GcsFileSystem` runs into a lengthy timeout game during `fs.create_dir(bucket)`

python/pyarrow/tests/conftest.py
```diff
@@ -184,9 +184,16 @@ def gcs_server():
...
+        # start server
         proc = subprocess.Popen(args, env=env)
-    except OSError as e:
+        # Make sure the server is alive.
+        assert proc.poll() is None
```
h-vetinari, line 192: Moving the assert is not strictly speaking necessary anymore, but still cleaner, because with the current setup this can just fail during the setup of the `gcsfs` fixture, which unnecessarily errors the respective tests, rather than just skipping them.

dev/tasks/conda-recipes/r-arrow/build.sh
```diff
@@ -1,9 +1,14 @@
...
-$R CMD INSTALL --build r/.
+# set C++17 due to abseil
+export ARROW_R_CXXFLAGS="${ARROW_R_CXXFLAGS} -std=c++17"
+
+if [[ "${target_platform}" == osx-* ]]; then
+    # See https://conda-forge.org/docs/maintainer/knowledge_base.html#newer-c-features-with-old-sdk
+    export ARROW_R_CXXFLAGS="${ARROW_R_CXXFLAGS} -D_LIBCPP_DISABLE_AVAILABILITY"
+fi
+
+${R} CMD INSTALL --build r/. ${R_ARGS}
```
h-vetinari, line 14: `${R_ARGS}` is apparently necessary to be able to cross-compile

dev/tasks/conda-recipes/r-arrow/build.sh
```diff
@@ -1,9 +1,14 @@
...
+#!/bin/bash
 set -ex
 
 export DISABLE_AUTOBREW=1
-# See https://conda-forge.org/docs/maintainer/knowledge_base.html#newer-c-features-with-old-sdk
-export ARROW_R_CXXFLAGS="${ARROW_R_CXXFLAGS} -D_LIBCPP_DISABLE_AVAILABILITY"
 
-$R CMD INSTALL --build r/.
+# set C++17 due to abseil
+export ARROW_R_CXXFLAGS="${ARROW_R_CXXFLAGS} -std=c++17"
```
h-vetinari, line 7: I realized later that this comment is not 100% accurate; it's less about abseil, and more about arrow having move the C++17, full stop. Unless people want a specific fix for that, I'll fix that in the feedstock and it'll be part of the next sync.

dev/tasks/conda-recipes/r-arrow/bld.bat
```diff
@@ -1,9 +1,14 @@
...
-IF %ERRORLEVEL% NEQ 0 exit 1
-cp %RECIPE_DIR%/install.libs.R r/src
-IF %ERRORLEVEL% NEQ 0 exit 1
-set "MAKEFLAGS=-j%CPU_COUNT%"
-"%R%" CMD INSTALL --build r
-IF %ERRORLEVEL% NEQ 0 exit 1
+@echo on
+
+bash %RECIPE_DIR%/build_win.sh
+IF %ERRORLEVEL% NEQ 0 exit 1
```
h-vetinari, line 4: This file had mangled line endings, see https://github.com/apache/arrow/pull/14832/commits/115fe7573902adb30f60a1b39c2eaa3759fdb349

dev/tasks/tasks.yml
```diff
@@ -240,220 +242,208 @@ tasks:
...
   #   an independent feedstock as it doesn't have the Python but the
-  #   R dimension. To limit the number of CI jobs, we are building `r-arrow`
-  #   with the Python 3.7 jobs.
+  #   R dimension.
   # * The files in `dev/tasks/conda-recipes/.ci_support/` are automatically
   #   generated and to be synced regularly from the feedstock. We have no way
   #   yet to generate them inside the arrow repository automatically.
 
-  conda-linux-gcc-py37-cpu-r40:
+  conda-linux-x64-cpu-r41:
```
h-vetinari, line 250: The overall diff in this file is pretty busted. Happy to explain if questions arise (also see individual commits)

h-vetinari (commented):
I think this is getting close to finished. For now it's a pretty faithful reproduction of the state of the feedstocks (https://github.com/conda-forge/arrow-cpp-feedstock/pull/911 & https://github.com/conda-forge/r-arrow-feedstock/pull/58). However, I made two more adaptations for the CI here that IMO make sense:

* Remove the openssl1.1.1 jobs - these exist on the feedstock because some packages in conda-forge haven't made the jump to openssl 3 yet, but that gap is rapidly closing and we're not getting much here for the (almost) duplication of CI jobs.
* Run tests for pyarrow only for one python version - there's very little variance between versions, and I don't expect this to miss much if anything. On the other hands, running the tests for 4 python versions takes quite a bunch longer, especially in emulation (linux-aarch64).
  * This is still a strict improvement over the status quo of not running tests; as discussed above, I'll raise issues for the test skips / segfaults once this PR is merged

Thoughts @kou @pitrou @raulcd @jorisvandenbossche @xhochy?

PS. The commits are pretty incremental, but somewhat redundant. I didn't bother squashing because that happens upon merge anyway. But happy to do so if it helps for reviewing.

python/pyarrow/tests/conftest.py
```diff
@@ -184,9 +184,16 @@ def gcs_server():
...
+    except (ModuleNotFoundError, OSError) as e:
         pytest.skip(f"Command {args} failed to execute: {e}")
+    except AssertionError as e:
```
pitrou, line 195: I would rather have an explicit `if` rather than an assert:
```suggestion
    if proc.poll() is not None:
```

pitrou:
@h-vetinari I see the PyArrow test suite is run only for Python 3.10, is that expected?

h-vetinari:
> However, I made two more adaptations for the CI here that IMO make sense:
> [...]
> 
>  * Run tests for pyarrow only for one python version - there's very little variance between versions, and I don't expect this to miss much if anything. On the other hands, running the tests for 4 python versions takes quite a bunch longer, especially in emulation (linux-aarch64).
>       
>    * This is still a strict improvement over the status quo of not running tests; as discussed above, I'll raise issues for the test skips / segfaults once this PR is merged

☝️

h-vetinari:
Context: Running the test suite 4 times makes the aarch builds take up to 4h.

pitrou:
Oops, I see. What about the x64 builds?

h-vetinari:
> Oops, I see. What about the x64 builds?

It's feasible on linux-x64 (~1.5h total); on win and osx it takes 2.5 - 3h for running it for all 4 versions.

I don't mind running it for all versions, I just didn't want to regress the runtime too much (because previously the test suite wasn't run at all). If you feel you have enough resources, then we can run the suite for all python versions, but again, I really doubt we'd catch anything additionally (and for such corner cases, we still run the suite for everything on the feedstock).

pitrou:
> I don't mind running it for all versions, I just didn't want to regress the runtime too much (because previously the test suite wasn't run at all).

Hmm, I see. How about doing it only on Linux x64?

h-vetinari:
Linux-x64 is the least failure-prone platform, but if you want, sure. :)

pitrou:
I meant run tests for all Python versions on Linux-x64, and for a single version on other platforms as you're doing here.

h-vetinari:
That's indeed how I understood it, but just wanted to add that the extra python version coverage is happening on the least failure-prone platform (so I'm still unsure this would catch anything).

I'll add a commit to do this when I get back to my computer

h-vetinari:
@github-actions crossbow submit -g conda

kou:
Can we remove `conda-osx-arm64-cpu-r41` and `conda-osx-arm64-cpu-r42` jobs because they failed to run `R -e "library('arrow')"`?

https://dev.azure.com/ursacomputing/crossbow/_build/results?buildId=40908&view=logs&j=cf796865-97b7-5cd1-be8e-6e00ce4fd8cf&t=88ee2fb8-46fd-5c68-fde3-1c8d31ba2a5f&l=28318

```text
+ /usr/..._placehold_/bin/R -e 'library('\''arrow'\'')'
/usr/..._placehold_/bin/R: line 271: /usr/..._placehold_/lib/R/bin/exec/R: Bad CPU type in executable
/usr/..._placehold_/bin/R: line 271: /usr/..._placehold_/lib/R/bin/exec/R: Undefined error: 0
```
Azure Pipelines doesn't provide arm64 macOS yet. So we can't run cross-compiled packages.

Or we may be able to ignore `$R -e "library('arrow')"` in `dev/tasks/conda-recipes/r-arrow/meta.yaml` https://github.com/apache/arrow/blob/master/dev/tasks/conda-recipes/r-arrow/meta.yaml#LL61 on cross-compiling.

h-vetinari:
> Can we remove `conda-osx-arm64-cpu-r41` and `conda-osx-arm64-cpu-r42` jobs because they failed to run `R -e "library('arrow')"`?

I'm pretty sure this is solved by https://github.com/apache/arrow/pull/14832/commits/b4147098fcdd229964bf3c97da129e2a5e4265e9.

> Azure Pipelines doesn't provide arm64 macOS yet. So we can't run cross-compiled packages.

Same for conda-forge, so we're cross-compiling there as well. I think the issue was that we need some kind of synch of `build_steps.sh` and `run_docker_build.sh`; while there's a comment here saying they have been modified for crossbow, I don't know what changes these are (aside from obvious ones like `R_CONFIG`).

In any case, I added a minimal [change](https://github.com/conda-forge/arrow-cpp-feedstock/blob/main/.scripts/build_steps.sh#L48-L50) from the conda-forge `build.sh` that should fix this.

h-vetinari:
> Can we remove `conda-osx-arm64-cpu-r41` and `conda-osx-arm64-cpu-r42` jobs because they failed to run `R -e "library('arrow')"`?

Other than that, I'd like to not remove them - they still serve as catching (build) regressions, not least due to the need to cross-compile.

h-vetinari:
@github-actions crossbow submit -g conda

h-vetinari:
As an aside, if you have the rights (e.g. @kou), you could cancel the currently running conda-builds (including the conda nightlies) in the crossbow queue? They're either obsoleted by the last few commits or (for the nightlies) going to fail anyway.

kou:
> if you have the rights (e.g. @kou), you could cancel the currently running conda-builds (including the conda nightlies) in the crossbow queue? 

I can do it but it seems that there are jobs only for `actions-d906b1c7f1` now.
Sorry for not canceling other jobs as soon as possible.

h-vetinari:
@github-actions crossbow submit -g conda

h-vetinari:
@kou @pitrou 
This should be ready, PTAL

h-vetinari:
> This should be ready, PTAL

Ugh, one more unexpected thing - I got caught off guard because `conda-osx-x64-cpu-r41` passed, but now `conda-osx-x64-cpu-r42` failed... 😑

h-vetinari:
@github-actions crossbow submit -g conda

h-vetinari:
The two failing jobs had spurious errors:
```
E   OSError: When reading information for key 'README.txt' in bucket 'mf-nwp-models': AWS Error NETWORK_CONNECTION during HeadObject operation: curlCode: 28, Timeout was reached
```

```
E                   PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\VSSADM~1\\AppData\\Local\\Temp\\tmpc_geax_g\\z\\part-0.arrow'
```

They were both passing before (and will again), so for now I'm not touching them.

xhochy:
Changes look good. Any further help you would need from my side?

dev/tasks/conda-recipes/r-arrow/build.sh
kou: Could you add this note as comment in this file?

kou (approved):
+1

I can't review these changes deeply. Sorry.

It seems that these 2 errors aren't caused by conda. I think that we should improve PyArrow's tests.

h-vetinari:
> Changes look good. Any further help you would need from my side?

Happy to hear. You could review/merge https://github.com/conda-forge/r-arrow-feedstock/pull/58, but otherwise I think we're good.

jorisvandenbossche (approved):
Assuming the changes to the conda feedstock files are just adapted from upstream (as we would like them to be, without unnecessary changes on our side), I didn't review that part, but the changes to pyarrow itself look fine.

h-vetinari:
Alternatively, I guess an argument could be made that _one_ job per arch (counting cpu/cuda) could be enough, which builds all python versions, and then latest R on top.

raulcd:
> Looking at the two remaining failures

I have manually re-run `conda-linux-aarch64-cpu-r41` and it has succeeded

raulcd (approved):
I've tried reviewing the different commits individually and looks good to me

kou:
Can we merge this as-is? This fixes our nightly CI.
We can work on some improvements as follow-up tasks.

kou:
No objection. I'll merge this.

PR #14843, opened 2022-12-05
GH-14842: [C++] Propagate some errors in JSON chunker

### LINKED ISSUE #14842: [C++] Propagate some errors in JSON chunker

The current JSON `Chunker` defers all error reporting to a later parsing stage when `ParseOptions::newlines_in_values = true`. However, this poses issues when sequentially chunking buffers. Consider this:
```
std::shared_ptr<Buffer> whole, rest;
// Trailing right-bracket
chunker->Process(Buffer::FromString("{\"a\":0}}"), &whole, &rest);
```
Here, `whole` will be `{"a":0}` but `rest` will be `}`, which doesn't start a valid JSON block. As such, for the next buffer, you can't then call `ProcessWithPartial` with `rest` as its `partial` argument without crashing (via DCHECK, if enabled). This effectively prevents us from handling the error at all.

### Component(s)

C++

benibus:
* Closes: #14842

### DIFF
--- a/cpp/src/arrow/json/chunker.cc
+++ b/cpp/src/arrow/json/chunker.cc
@@ -125,13 +125,12 @@ namespace {
 class ParsingBoundaryFinder : public BoundaryFinder {
  public:
   Status FindFirst(string_view partial, string_view block, int64_t* out_pos) override {
-    // NOTE: We could bubble up JSON parse errors here, but the actual parsing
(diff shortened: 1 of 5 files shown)


benibus:
Thanks, @pitrou. This should be all set for a second look.

pitrou (approved):
Thanks for the update @benibus !

PR #14845, opened 2022-12-05
ARROW-18424: [C++] Fix Doxygen error on ARROW_ENGINE_EXPORT

rtpsw:
See https://issues.apache.org/jira/browse/ARROW-18424


pitrou (approved):
Thanks @rtpsw !

PR #14847, opened 2022-12-05
ARROW-14999: [C++] Optional field name equality checks for map and list type

wjones127:
BREAKING CHANGE

Two changes for "internal fields" (fields within ListTypes and MapTypes):

 * ListType and MapType internal field names only matter in comparisons when `check_metadata` is `true`.
 * Nullability of MapType internal fields now matters in comparison.

### DIFF
--- a/cpp/src/arrow/compare.cc
+++ b/cpp/src/arrow/compare.cc
@@ -43,6 +43,7 @@
 #include "arrow/util/bitmap_ops.h"
 #include "arrow/util/bitmap_reader.h"
 #include "arrow/util/checked_cast.h"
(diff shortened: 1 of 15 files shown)


pitrou:
Seems like the Java test changes won't work as the latest stable PyArrow is used?

wjones127:
CI failures are unrelated.

paleolimbot (approved):
The R side looks good to me! Thanks!

pitrou (commented):
Thanks a lot @wjones127 ! This looks good on the princople, a couple minor comments.

pitrou (approved):
+1, nice work @wjones127 !

PR #14352, opened 2022-10-08
ARROW-17642: [C++] Add ordered aggregation

rtpsw:
See https://issues.apache.org/jira/browse/ARROW-17642

PR #14867, opened 2022-12-07
GH-14866: [C++] Remove internal GroupBy implementation

### LINKED ISSUE #14866: [C++] Remove internal GroupBy implementation

Currently there are two ways to compute a group by.  The supported way is to use an aggregate node in an exec plan.  The second (internal) way is to use the internal function `arrow::internal::GroupBy`.  This internal function simulates, but does not actually use, an aggregate node.

The internal implementation has caused issues in the past where we did not notice an error in the aggregate node's invocation of aggregate kernels since we use the internal function for testing aggregates and it behaves slightly differently.  The internal implementation also requires maintenance and significantly complicated #14352 .

I would like to remove the internal implementation.  Unfortunately, the internal implementation is used by tests, benchmarks, and pyarrow.  However, we should be able to update those bindings to a friendly wrapper around exec plans.

### Component(s)

C++

### LINKED ISSUE #34238: [C++][Python] Segfault when calling groupby on table with misaligned chunks

- I have a table with ~100k records and 4 chunks of various sizes `[32768, 32768, 29599, 16692]`
- key1 has got high cardinality (30k)
- key2 has got low cardinality (10)

| key1       | key2    |     value |
|:-----------|:--------|----------:|
| KEY1_12226 | KEY1_08 | 0.348599  |
| KEY1_10214 | KEY1_08 | 0.954173  |
| KEY1_26821 | KEY1_09 | 0.416615  |
| KEY1_24557 | KEY1_06 | 0.883226  |
| KEY1_27823 | KEY1_08 | 0.0127225 |

I'm trying to do a groupby on key1, key2 to get the sum of value. It works fine in general. But when I preprocess the data I misalign the chunks in my table and it fails.

```
import numpy as np
import pyarrow as pa

KEYS_1 = [f"KEY1_{i:05d}" for i in range(30_000)]
KEYS_2 = [f"KEY1_{i:02d}" for i in range(10)]
SIDE = ["LEFT", "RIGHT"]


def generate_table(sizes):
    batches = [
        pa.record_batch(
            [
                np.random.choice(KEYS_1, size),
                np.random.choice(KEYS_2, size),
                np.random.rand(size),
            ],
            ["key1", "key2", "value"],
        )
        for size in sizes
    ]
    return pa.Table.from_batches(batches)


table = generate_table([32768, 32768, 29599, 16692])

# This works well:
pa.TableGroupBy(table, ["key1", "key2"]).aggregate(
    [
        ["value", "sum"],
    ]
)

# This misaligns the chunks
table = table.set_column(
    table.schema.get_field_index("value"),
    "value",
    table["value"].combine_chunks(),
)

print("HERE")
pa.TableGroupBy(table, ["key1", "key2"]).aggregate(
    [
        ["value", "sum"],
    ]
)  # segfault :-(
print("NEVER THERE")

```

It took me a while to go to the bottom of the problem. The size of the chunks and the cardinality of the keys seem to play an important factor in weather it fails or not.

The short term solution is for me to call combine_chunks before the groupby.

This is on pyarrow 11.0.0, python 3.9

### Component(s)

Python

westonpace:
* Closes: #14866
* Closes: #34238


python/pyarrow/_compute.pyx
```diff
@@ -2211,11 +2213,18 @@ def _group_by(args, keys, aggregations):
...
             GroupBy(c_args, c_keys, c_aggregations)
         )
 
-    return wrap_datum(result)
+    result_batches = []
+    for c_batch in c_agg_batches:
+        result_batch = []
+        for c_column in c_batch.values:
+            result_batch.append(wrap_datum(c_column))
+        result_batches.append(result_batch)
```
jorisvandenbossche, line 2225: Could this use the `ExecBatch::ToRecordBatch` to return a list of batches instead? (that seems simpler to work with, and also simplifies the code here)

cpp/src/arrow/compute/kernels/hash_aggregate_test.cc
```diff
@@ -185,7 +184,8 @@ Result<Datum> GroupByUsingExecPlan(const BatchesWithSchema& input,
 Result<Datum> GroupByUsingExecPlan(const std::vector<Datum>& arguments,
```
rtpsw, line 184: Perhaps this could just be called `GroupBy` now.

cpp/src/arrow/compute/exec/source_node.cc
```diff
@@ -308,9 +308,6 @@ struct SchemaSourceNode : public SourceNode {
     auto& schema = cast_options.schema;
     auto io_executor = cast_options.io_executor;
 
-    if (io_executor == NULLPTR) {
-      io_executor = plan->exec_context()->executor();
-    }
```
rtpsw, line 313: Is this removal intended?

westonpace: It's not terribly relevant.  I was initially wanting to use `exec_batch_source` here but ran into a problem because I was not transferring off the background generator.  Incidentally, I think it may be close to time to fix the background generator to remove this limitation, but I have enough on my plate for the moment.

This change was simply because, a few lines down (on line 316) we have a very similar if condition:

```
    if (io_executor == NULLPTR) {
      io_executor = io::internal::GetIOThreadPool();
    }
```

I don't think we need both of these statements and defaulting to the I/O thread pool seemed like the better default.  If that is not the correct default, or there is some subtlety I am missing, let me know and I can revert this.

python/pyarrow/_compute.pyx
westonpace: Yes, I was a bit torn on this one.  The simplest thing might be for `arrow::compute::GroupBy` to return `std::shared_ptr<RecordBatch>`.  However, I don't have column names, so I would be making those up.  Also, the inputs are datums, and so it seemed like a mismatch to receive datums (not arrays) and return a record batch (and not an exec batch).  So then I ended up with a list of lists of arrays which is unpleasant too.

I could return a list of record batches but then I would have to copy the `CreateSimpleSchema` method which invents names for these columns.  Since the caller of this function has those names, and this function is private, I figured it best to leave that work for the caller.

That being said, after sleeping on this a bit, maybe a better change would be to change `arrow::compute::GroupBy` to receive arrays (not datums) and then returning a record batch wouldn't be inconsistent.

cpp/src/arrow/compute/kernels/hash_aggregate_test.cc
westonpace: I'm a little hesitant to use `GroupBy` since that would mean the `GroupBy` function calls the `GroupBy` function which is a little confusing.  I changed it to `RunGroupBy`.

python/pyarrow/_compute.pyx
westonpace: Ok, I ended up promoting `arrow::compute::GroupBy` to a "proper" convenience function.  It now accepts arrays, returns a table, is a bit friendlier with field names, checks for invalid input, is added to the api.h file, and has unit tests.

rtpsw:
@jorisvandenbossche, it looks like this PR is pending your review. Could you take a look? I have [another PR](https://github.com/apache/arrow/pull/14352/), which is important to me, that is waiting for this one.

jorisvandenbossche:
> Ok, I ended up promoting `arrow::compute::GroupBy` to a "proper" convenience function. It now accepts arrays, returns a table, is a bit friendlier with field names, checks for invalid input, is added to the api.h file, and has unit tests.


Sorry for the slow reply. I don't fully understand how the above can work correctly (working on arrays, instead of on chunked arrays), because you now calculate the groupby batch by batch? But then those results should be "merged" somehow, and not just concatenated?

jorisvandenbossche:
To illustrate what I mean, using a small example (and using this branch):

```
In [7]: table = pa.table({'key': ['a', 'b', 'a', 'b', 'a', 'b'], 'col': range(6)})

In [8]: table = pa.Table.from_batches(table.to_batches(max_chunksize=3))

In [9]: table.to_pandas()
Out[9]: 
  key  col
0   a    0
1   b    1
2   a    2
3   b    3
4   a    4
5   b    5

In [10]: table.group_by('key').aggregate([('col', 'sum')]).to_pandas()
Out[10]: 
   col_sum key
0        2   a
1        1   b
2        8   b
3        4   a
```

I created a table consisting of multiple chunks, and then the result is incorrect as it is the concatentation of individual results of each chunk.

rtpsw:
@westonpace, though I didn't review the code carefully, it looks like this PR removes code that is refactored and used in [my ordered/segmented aggregation PR](https://github.com/apache/arrow/pull/14352/). GIven this, and the correctness problem @jorisvandenbossche is pointing out in this PR, might it make sense to wait with your removal until my PR is merged? or to include your removal in my PR?

westonpace:
@jorisvandenbossche it should be using an exec plan internally with an aggregate node.  The aggregate node knows how to maintain state from batch to batch.  However, I agree your example is pointing out a bug in my code.  I'll take a look.

@rtpsw I will try merging this with your branch (and then create a third PR) just to make sure it works.  I don't know if I can get to it before tomorrow morning.  Either way, if there is concern about this approach, we can merge yours and clean up with mine.

The basic idea is that we have kernel functions for arrays / single batches and exec plans for multiple batches (which should include chunked arrays).

I don't see any value in maintaining a third path for chunked arrays when they should just be a special case of multiple batches / exec plans.

jorisvandenbossche:
> it should be using an exec plan internally with an aggregate node.

But I assumed that it would be the purpose of the `GroupBy(..)` helper to do this? (its doc comment says "The result will be calculated using an exec plan with an aggregate node")

westonpace:
Yes.  I see the problem now.  The group by helper needs an overload that accepts chunked arrays, converts them to a table, and then uses that as input.

westonpace:
@rtpsw https://github.com/apache/arrow/commit/0f2b458ec45c851707bf8f6f93e0e703af359f26 is an example of layering this PR on top of your ordered groupby changes.  I couldn't get the aggregate node working because I don't think it works at the moment and didn't want to dive too deep into that problem just yet.  However, I don't see any reason your ordered aggregation won't work with this PR. Also, once things are working, we should be able to go further and:

 * remove chunked array from datum
 * remove all grouper.cc/grouper.h changes which use spans and instead only use batches

westonpace:
I'm going to rebase this and address the problem Joris raised.

westonpace:
@jorisvandenbossche Thanks for pointing out that problem.  I think I've addressed your concerns (and I've added your example as a test case).

westonpace:
@jorisvandenbossche friendly ping now that the holidays are over.

python/pyarrow/table.pxi
```diff
@@ -5383,11 +5383,9 @@ list[tuple(str, str, FunctionOptions)]
             for col_name, (aggr_name, _) in zip(columns, group_by_aggrs)
         ] + self.keys
 
-        result = _pc()._group_by(
+        agg_tables = []
```
jorisvandenbossche, line 5386: Left-over variable of the previous iteration?

jorisvandenbossche (approved):
Looks good! (only looked at the cython code again, and the expected behaviour)

Just some small clean-ups needed

westonpace:
I've rebased this and will merge if CI is still passing.

python/pyarrow/table.pxi
westonpace: I've removed this.

python/pyarrow/tests/test_table.py
```diff
@@ -2024,6 +2024,18 @@ def sorted_by_keys(d):
         "values_count": [1]
     }
 
+    table = pa.table({
+        'keys': ['a', 'b', 'a', 'b', 'a', 'b'],
+        'values': range(6)})
+    table_with_chunks = pa.Table.from_batches(
+        table.to_batches(max_chunksize=3))
+    r = table_with_chunks.group_by('keys').aggregate([('values', 'sum')])
+    print(r)
```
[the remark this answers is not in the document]
westonpace: Removed.

westonpace:
closes #34238

jorisvandenbossche:
@westonpace BTW, if you want that the issue gets automatically closed, you need to add the "closes #34238" to the top comment, and not only in a comment like the one above (our tooling still won't automatically assign and milestone the issue though, it's only github that will automatically close it then)

westonpace:
@jorisvandenbossche Oh!  Thanks for catching that.  I hadn't realized that (I think this is the first PR I've done that closed multiple issues).

PR #14868, opened 2022-12-07
MINOR: [Doc][Python] fix timestamps in dataframe output

fdzuJ:
Current console output differ from input


jorisvandenbossche (approved):
Thanks!

PR #14874, opened 2022-12-07
GH-14873: [Java] DictionaryEncoder can decode without building a DictionaryHashTable

### LINKED ISSUE #14873: [Java] DictionaryEncoder can decode without building a DictionaryHashTable

Today DictionaryEncoder always forces the building of a DictionaryHashTable in the constructor. It can be avoided in scenarios where only decoding is required.

### Component(s)

Java

gf2121:
Today DictionaryEncoder always forces the building of a DictionaryHashTable in the constructor. It can be avoided in scenarios where only decoding is required.

see: https://github.com/apache/arrow/issues/14873
* Closes: #14873


lidavidm (approved):
Thanks!

PR #14877, opened 2022-12-07
GH-14876: [Go] Handling Crashes in C Data interface

### LINKED ISSUE #14876: [Go] Address Crashes for empty C Data arrays with nil buffers

Following up from #14805: Go's `cdata` package needs to address handling nil data buffers for 0 length empty arrays.

### Component(s)

Go

zeroshade:
* Closes: #14876

### DIFF
--- a/go/arrow/cdata/cdata.go
+++ b/go/arrow/cdata/cdata.go
@@ -37,6 +37,7 @@ package cdata
 import "C"
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"reflect"
@@ -476,17 +477,26 @@ func (imp *cimporter) doImport(src *CArrowArray) error {
 		}
 
 		bufs := []*memory.Buffer{nil, nil, nil}
+		var err error
 		if imp.arr.n_buffers == 3 {
 			// legacy format exported by older arrow c++ versions
-			bufs[1] = imp.importFixedSizeBuffer(1, 1)
-			bufs[2] = imp.importFixedSizeBuffer(2, int64(arrow.Int32SizeBytes))
+			if bufs[1], err = imp.importFixedSizeBuffer(1, 1); err != nil {
+				return err
+			}
+			if bufs[2], err = imp.importFixedSizeBuffer(2, int64(arrow.Int32SizeBytes)); err != nil {
+				return err
+			}
 		} else {
 			if err := imp.checkNumBuffers(2); err != nil {
 				return err
 			}
 
-			bufs[1] = imp.importFixedSizeBuffer(0, 1)
-			bufs[2] = imp.importFixedSizeBuffer(1, int64(arrow.Int32SizeBytes))
+			if bufs[1], err = imp.importFixedSizeBuffer(0, 1); err != nil {
+				return err
+			}
+			if bufs[2], err = imp.importFixedSizeBuffer(1, int64(arrow.Int32SizeBytes)); err != nil {
+				return err
+			}
 		}
 
 		children := make([]arrow.ArrayData, len(imp.children))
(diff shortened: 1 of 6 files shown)


go/arrow/cdata/cdata.go
```diff
@@ -583,7 +597,19 @@ func (imp *cimporter) importFixedSizePrimitive() error {
 		}
 		values = imp.importBitsBuffer(1)
 	}
-	imp.data = array.NewData(imp.dt, int(imp.arr.length), []*memory.Buffer{nulls, values}, nil, int(imp.arr.null_count), int(imp.arr.offset))
+
+	var dict *array.Data
+	if dt, ok := imp.dt.(*arrow.DictionaryType); ok {
```
pitrou, line 602: Sorry, I don't really read Go, but what happens if `ok` is false here?

go/arrow/cdata/cdata.go
```diff
@@ -613,6 +639,12 @@ func (imp *cimporter) checkNumBuffers(n int64) error {
 func (imp *cimporter) importBuffer(bufferID int, sz int64) *memory.Buffer {
 	// this is not a copy, we're just having a slice which points at the data
 	// it's still owned by the C.ArrowArray object and its backing C++ object.
+	if imp.cbuffers[bufferID] == nil {
+		if sz != 0 {
+			panic("invalid buffer")
```
pitrou, line 644: "panic" bubbles a regular catchable error, or it's a controlled crash?

go/arrow/cdata/cdata_exports.go
```diff
@@ -353,7 +366,7 @@ func exportArray(arr arrow.Array, out *CArrowArray, outSchema *CArrowSchema) {
 		buffers := allocateBufferPtrArr(len(arr.Data().Buffers()))
 		for i := range arr.Data().Buffers() {
 			buf := arr.Data().Buffers()[i]
-			if buf == nil {
+			if buf == nil || buf.Len() == 0 {
```
pitrou, line 369: This doesn't seem mandatory either.

go/arrow/cdata/test/test_export_to_cgo.py
```diff
@@ -161,6 +162,73 @@ def test_batch_roundtrip(self):
...
+            b.validate(full=True)
+            assert a.to_pylist() == b.to_pylist()
+            assert a.type == b.type
+            del a
+            del b
+    return test
+
 
 if __name__ == '__main__':
+    for typ in _test_pyarrow_types:
```
pitrou, line 230: You don't have to do this manually, you can use a sub-test instead:
https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests

go/arrow/cdata/cdata.go
zeroshade: if `ok` is false here, `dict` remains `nil` and we just construct the `ArrayData` object without a dictionary as normal.

go/arrow/cdata/cdata.go
zeroshade: `panic` *can* be caught and handled via `defer` and `recover`, otherwise it will be a controlled crash with a stack trace (unless debug info was stripped when building).

go/arrow/cdata/cdata_exports.go
zeroshade: The reason for this is line 374 below. In the case where `buf.Len() == 0` there is no allocated data to grab a pointer to via `&buf.Bytes()[0]` and you end up panic'ing with an error of attempting to get index 0 of a 0 length slice. Thus if we create an empty array, it's better to just use a nil pointer.

The alternative here would be to use `&buf.Buf()[0]` instead, which points at the reserved bytes (since creating a new buffer will, by default, automatically reserve 64 bytes if it wasn't expanded) but I thought it better to not force us to keep that memory around for a 0 length array.

go/arrow/cdata/test/test_export_to_cgo.py
zeroshade: Ah nice! I'll update this to use that. thanks!

go/arrow/cdata/cdata.go
pitrou: Can't you return a regular error then?

zeroshade: sure, had to propagate the error through to the callers since this method previously didn't return an error. latest commit updates this with the error instead of panic.

go/arrow/cdata/cdata_exports.go
pitrou: Sounds ok to me.

PR #14878, opened 2022-12-07
GH-14865: [Go][Parquet] Address several memory leaks of buffers in pqarrow

### LINKED ISSUE #14865: [Go] pqarrow.WriteArrowToColumn leaks memory from its memory.Allocator

I am testing a benchmark with different parquet libraries to write an Arrow buffer to file. My benchmark checks the allocator is empty at the end of its run. I found that pqarrow retains a memory.Buffer if maybeParentNulls is true, and never releases it. I looked through the API to find a Release() function or some functionality that would release this buffer but did not find anything.

I noticed the unit tests do not check the allocator is zeroed out.

This is the if-statement I am concerned about:
```
// WriteArrowToColumn writes apache arrow columnar data directly to a ColumnWriter.
// Returns non-nil error if the array data type is not compatible with the concrete
// writer type.
// 
// leafArr is always a primitive (possibly dictionary encoded type).
// Leaf_field_nullable indicates whether the leaf array is considered nullable
// according to its schema in a Table or its parent array.
func WriteArrowToColumn(ctx context.Context, cw file.ColumnChunkWriter, leafArr arrow.Array, defLevels, repLevels []int16, leafFieldNullable bool) error {
   // Leaf nulls are canonical when there is only a single null element after a list
   // and it is at the leaf.
   colLevelInfo := cw.LevelInfo()
   singleNullable := (colLevelInfo.DefLevel == colLevelInfo.RepeatedAncestorDefLevel+1) && leafFieldNullable
   maybeParentNulls := colLevelInfo.HasNullableValues() && !singleNullable

   if maybeParentNulls {
      buf := memory.NewResizableBuffer(cw.Properties().Allocator())
---NON-RELEASED ALLOC HERE--->      buf.Resize(int(bitutil.BytesForBits(cw.Properties().WriteBatchSize())))
      cw.SetBitsBuffer(buf)
   }
...
```

This is the suspicious allocation (I added a PrintStack call in my own custom debug allocator to print this):
```
goroutine 19 [running]:                                                                                                                                                                                                                                                                  
runtime/debug.Stack()                                                                                                                                                                                                                                                                    
        /usr/local/go/src/runtime/debug/stack.go:24 +0x65                                                                                                                                                                                                                                
runtime/debug.PrintStack()                                                                                                                                                                                                                                                               
        /usr/local/go/src/runtime/debug/stack.go:16 +0x19                                                                                                                                                                                                                                
gitlab.eng.vmware.com/taurus/data-mesh.git/compact-lake/rows.(*VerboseAllocator).Allocate(0xc0002d5da0, 0x80)                                                                                                                                                                            
        /home/rick/data-mesh/compact-lake/rows/buffer_test.go:145 +0x6a                                                                                                                                                                                                                  
github.com/apache/arrow/go/v11/arrow/memory.(*Buffer).Reserve(0xc00011ee10, 0xc0001596b0?)                                                                                                                                                                                               
        /home/rick/go/pkg/mod/github.com/apache/arrow/go/v11@v11.0.0-20221206133351-50a164ec7f64/arrow/memory/buffer.go:110 +0x5b                                                                                                                                                        
github.com/apache/arrow/go/v11/arrow/memory.(*Buffer).resize(0xc00011ee10, 0x80, 0xf0?)                                                                                                                                                                                                  
        /home/rick/go/pkg/mod/github.com/apache/arrow/go/v11@v11.0.0-20221206133351-50a164ec7f64/arrow/memory/buffer.go:130 +0xf4                                                                                                                                                        
github.com/apache/arrow/go/v11/arrow/memory.(*Buffer).Resize(...)                                                                                                                                                                                                                        
        /home/rick/go/pkg/mod/github.com/apache/arrow/go/v11@v11.0.0-20221206133351-50a164ec7f64/arrow/memory/buffer.go:119                                                                                                                                                              
github.com/apache/arrow/go/v11/parquet/pqarrow.WriteArrowToColumn({0xddcd60, 0xc000130008}, {0xde0c40, 0xc00029e180}, {0xde2040?, 0xc0001736c0?}, {0xc0002d8500, 0x1, 0x80}, {0x0, ...}, ...)                                                                                            
        /home/rick/go/pkg/mod/github.com/apache/arrow/go/v11@v11.0.0-20221206133351-50a164ec7f64/parquet/pqarrow/encode_arrow.go:219 +0x159                                                                                                                                              
github.com/apache/arrow/go/v11/parquet/pqarrow.(*ArrowColumnWriter).Write(0xc000159b60, {0xddcd60, 0xc000130008})                                                                                                                                                                        
        /home/rick/go/pkg/mod/github.com/apache/arrow/go/v11@v11.0.0-20221206133351-50a164ec7f64/parquet/pqarrow/encode_arrow.go:195 +0x533                                                                                                                                              
gitlab.eng.vmware.com/taurus/data-mesh.git/compact-lake/rows.(*Buffer).writeColumns(0xcbd4d5?, {0xddcd60, 0xc000130008}, 0xc000066c90?, {0xc0003382d0, 0x1, 0x0?}, 0x8?)                                                                                                                 
        /home/rick/data-mesh/compact-lake/rows/buffer.go:352 +0x3fa                                                                                                                                                                                                                      
gitlab.eng.vmware.com/taurus/data-mesh.git/compact-lake/rows.(*Buffer).WriteParquet(0xc000159eb8, {0xddcd60, 0xc000130008}, {0xddbaf0?, 0xc0002d5da0?}, {0xdd80a0, 0xc00011a2a0})                                                                                                        
        /home/rick/data-mesh/compact-lake/rows/buffer.go:323 +0x8dd                                    
gitlab.eng.vmware.com/taurus/data-mesh.git/compact-lake/rows.TestWriteOneRowParquet(0xc000105d40)      
        /home/rick/data-mesh/compact-lake/rows/buffer_test.go:78 +0x671                                                                     
testing.tRunner(0xc000105d40, 0xcfc730)                               
        /usr/local/go/src/testing/testing.go:1439 +0x102                                                                                    
created by testing.(*T).Run                                           
        /usr/local/go/src/testing/testing.go:1486 +0x35f  
```

### Component(s)

Go

zeroshade:
* Closes: #14865

### DIFF
--- a/go/parquet/file/column_writer.go
+++ b/go/parquet/file/column_writer.go
@@ -55,6 +55,7 @@ type ColumnChunkWriter interface {
 
 	LevelInfo() LevelInfo
 	SetBitsBuffer(*memory.Buffer)
+	HasBitsBuffer() bool
 }
 
 func computeLevelInfo(descr *schema.Column) (info LevelInfo) {
@@ -154,6 +155,7 @@ func newColumnWriterBase(metaData *metadata.ColumnChunkMetaDataBuilder, pager Pa
 	return ret
 }
 
+func (w *columnWriter) HasBitsBuffer() bool              { return w.bitsBuffer != nil }
 func (w *columnWriter) SetBitsBuffer(buf *memory.Buffer) { w.bitsBuffer = buf }
 
 func (w *columnWriter) LevelInfo() LevelInfo { return w.levelInfo }
@@ -534,6 +536,13 @@ func (w *columnWriter) Close() (err error) {
 
 		w.defLevelSink.Reset(0)
 		w.repLevelSink.Reset(0)
+		if w.bitsBuffer != nil {
+			w.bitsBuffer.Release()
+			w.bitsBuffer = nil
+		}
+
+		w.currentEncoder.Release()
+		w.currentEncoder = nil
 	}
 	return err
 }
--- a/go/parquet/file/column_writer.go
+++ b/go/parquet/file/column_writer.go
@@ -520,6 +520,20 @@ func (w *columnWriter) Close() (err error) {
 
 		w.FlushBufferedDataPages()
 
+		// ensure we release and reset everything even if we
+		// error out from the chunk statistics handling
+		defer func() {
+			w.defLevelSink.Reset(0)
+			w.repLevelSink.Reset(0)
+			if w.bitsBuffer != nil {
+				w.bitsBuffer.Release()
+				w.bitsBuffer = nil
+			}
+
+			w.currentEncoder.Release()
+			w.currentEncoder = nil
+		}()
+
 		var chunkStats metadata.EncodedStatistics
 		chunkStats, err = w.getChunkStatistics()
 		if err != nil {
@@ -533,16 +547,6 @@ func (w *columnWriter) Close() (err error) {
 			w.metaData.SetStats(chunkStats)
 		}
 		err = w.pager.Close(w.hasDict, w.fallbackToNonDict)
-
-		w.defLevelSink.Reset(0)
-		w.repLevelSink.Reset(0)
-		if w.bitsBuffer != nil {
-			w.bitsBuffer.Release()
-			w.bitsBuffer = nil
-		}
-
-		w.currentEncoder.Release()
-		w.currentEncoder = nil
 	}
 	return err
 }
--- a/go/parquet/internal/encoding/encoder.go
+++ b/go/parquet/internal/encoding/encoder.go
@@ -79,6 +79,14 @@ func newEncoderBase(e format.Encoding, descr *schema.Column, mem memory.Allocato
 	}
 }
 
+func (e *encoder) Release() {
+	poolbuf := e.sink.buf
+	memory.Set(poolbuf.Buf(), 0)
+	poolbuf.ResizeNoShrink(0)
+	bufferPool.Put(poolbuf)
+	e.sink = nil
+}
+
 // ReserveForWrite allocates n bytes so that the next n bytes written do not require new allocations.
 func (e *encoder) ReserveForWrite(n int)           { e.sink.Reserve(n) }
 func (e *encoder) EstimatedDataEncodedSize() int64 { return int64(e.sink.Len()) }
@@ -124,6 +132,16 @@ func (d *dictEncoder) Reset() {
 	d.idxValues = d.idxValues[:0]
 	d.idxBuffer.ResizeNoShrink(0)
 	d.memo.Reset()
+}
+
+func (d *dictEncoder) Release() {
+	d.encoder.Release()
+	d.idxBuffer.Release()
+	if m, ok := d.memo.(BinaryMemoTable); ok {
+		m.Release()
+	} else {
+		d.memo.Reset()
+	}
 }
 
 // append the passed index to the indexbuffer
--- a/go/parquet/internal/encoding/types.go
+++ b/go/parquet/internal/encoding/types.go
@@ -71,6 +71,7 @@ type TypedEncoder interface {
 	Allocator() memory.Allocator
 	// Type returns the underlying physical type this encodes.
 	Type() parquet.Type
+	Release()
 }
 
 // DictEncoder is a special kind of TypedEncoder which implements Dictionary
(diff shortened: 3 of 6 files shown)


pitrou:
I'm curious, doesn't Go have a GC?

zeroshade:
@pitrou Yes. But with the Arrow library for Go, we provide an `Allocator` interface to allow consumers to use a custom allocator rather than the default Go allocator so that they could, for instance, call out to `jemalloc` and have that manage their Arrow memory, hiding it from the Go GC so the Go GC no longer has to scan that. Or any other way they want to manage their memory.

As such, the objects have `Release` and `Retain` methods to track reference counting and eagerly release memory. The DefaultAllocator is just using Go's GC and setting slices to nil, etc. when they get released. So we have a test allocator `CheckedAllocator` which can track the allocations and frees to see if there's anywhere that a buffer is kept around without having called Free on it (which would cause a memory leak if they weren't using the Default Allocator, or can cause excessive creation of objects the GC will have to clean up). That's what we're referring to as memory leaks here.

go/parquet/file/column_writer.go
```diff
@@ -534,6 +536,13 @@ func (w *columnWriter) Close() (err error) {
 
 		w.defLevelSink.Reset(0)
 		w.repLevelSink.Reset(0)
+		if w.bitsBuffer != nil {
+			w.bitsBuffer.Release()
```
pitrou, line 540: Can there be a leak in `return err` above?

pitrou:
To make things clear: when `w.bitsBuffer` goes away, it wouldn't automatically free the underlying memory? Can't you use something like `SetFinalizer` to make this more robust?

zeroshade:
To be clear: if the user is using `memory.DefaultAllocator` then it *should* free the underlying memory at some point when the GC cleans everything up. But if they are using a custom allocator of some kind, then it will not actually call Free unless `Release` is called on `w.bitsBuffer`.

One thing I've been thinking of every so often is to try to leverage SetFinalizer more with the usage of the Allocators, but there are benefits to allowing users to have that granular control over *when* memory is released (well, at least when it's set to nil for the GC to pick it up) since SetFinalizer isn't guaranteed to run at any specific time.

go/parquet/file/column_writer.go
zeroshade: Good catch, i should shift this above that `return err` and use a `defer` to ensure it is always called.

pitrou:
Right, what I meant though is that _if_ `Release` wasn't called explicitly then `SetFinalizer` could _still_ avoid an eternal leak (and/or print a helpful warning message about the lack of an explicit `Release` call).

zeroshade:
Currently, you can compile with `-tags assert` and it will turn on a bunch of debug assertions which will panic or throw errors under certain conditions (such as calling `Release` too many times) and it's encouraged for users to utilize the `CheckedAllocator` for their testing to ensure they call release appropriately.

I could potentially look at adding finalizers to the buffers created by `NewResizableBuffer` though with a debug assertion to print a message if it wasn't released properly. I'll add an issue to look into doing that. Does that all make sense? I mostly don't want to blanket add the `SetFinalizer` since that could increase GC time and most users are probably just using the default allocator, so I'd prefer to keep it behind a build tag.

pitrou:
To be clear, that was just a suggestion. Testing is not necessarily sufficient because users generally won't bother exercising edge cases/error cases (as the comment I posted shows in your case). But that's a valid compromise in the end.

Documentation would probably also be useful to inform users about this.

zeroshade:
Sounds good. I'll add two issues: one to improve the Documentation on this and another to investigate potentially leveraging SetFinalizer to handle these issues.

necro351:
Hi, I'd like to test this change out, but TBH its my first time testing a change from a forked repo in GitHub. To try and test the change I added this line to my go.mod file:

```
replace github.com/apache/arrow/go/v11 => github.com/zeroshade/arrow/go/arrow memory-leak-pqarrow
```

...but when I try to do a go build I get...

```
go: errors parsing go.mod:
/home/rick/data-mesh/go.mod:64: replace github.com/zeroshade/arrow/go/arrow: version "v0.0.0-20221212203100-c369cafa0990" invalid: missing github.com/zeroshade/arrow/go/arrow/go.mod at revision c369cafa0990
```

Google is providing solutions to a bunch of related problems but not this one exactly, and I already tried a few things so I figured I'd ask for guidance here. Let me know and I will definitely test out your fork. Thanks in advance for the fix.

zeroshade:
Try changing it to be:

```
replace github.com/apache/arrow/go/v11 => github.com/zeroshade/arrow/go memory-leak-pqarrow
```

and then run `go mod tidy` afterwards. Let me know if that works.

zeroshade:
@necro351 any luck trying this?

necro351:
Deadline came up at work, but the change looks like it is fixing the issue. I'll repro locally and if I still see the incorrect behavior I'll comment on this thread.

PR #14879, opened 2022-12-07
PARQUET-2201: [parquet-cpp] Add stress test for RecordReader ReadRecords and SkipRecords.

fatemehp:
This pull request adds a stress test for testing ReadRecords and SkipRecords for required, optional, and repeated fields. It will create random pages and run a random set of skip and read calls to consume all the pages. It will then compare the expected output with the actual output from the record reader.


fatemehp:
@emkornfield could you take a look?

pitrou:
@fatemehp It seems there are CI failures, can we take a look at them?

emkornfield:
Looks like ASAN/UBSAN are still potentially real: 	 `61 - parquet-reader-test (Failed)`

fatemehp:
I was able to reproduce the bug by running the test 1000 times.

I found a bug when calling ReadRecords(/*num_records=*/0), and handled that case. I am not sure if a separate pull request is necessary for that change.

Also, I changed the main loop for the stress test to reset the reader after each batch of read/skip. This is consistent with how the reader will be used.

I ran the stress test 1000 times after the fixes and it passes all of them.

fatemehp:
@emkornfield can you take another look?

emkornfield (approved):
CI failures look unrelated at this point.

mapleFU:
https://github.com/apache/arrow/issues/34496
Report a segfault on ut here.

PR #14900, opened 2022-12-09
GH-14869: [C++] Add Cflags.private defining <lib>_STATIC to <lib>.pc.in.

### LINKED ISSUE #14869: [C++] arrow.pc should have -DARROW_STATIC for Windows static builds

Without, the generated pc file is insufficient (at least without "manually" defining ARROW_STATIC, which is unpleasant).

Quick hack fix: https://github.com/lukester1975/arrow/commit/2a8efd9c0bf69fe1b466e157bd69e83a757c926e

* Cflags.private is quite new (https://gitlab.freedesktop.org/pkg-config/pkg-config/-/merge_requests/13), so this approach might be unpalatable.
* pkgconfiglite is too old to include that (no commits since 2016??). It does quietly ignore the field, though.
* vcpkg does its own merging of Cflags.private into Cflags, so not an issue if pkg-config doesn't understand Cflags.private there (my case).
* Obviously this is applying to all platforms, not just Windows, but should do no harm...?

Seems like there should be some sort of fix here rather than asking vcpkg to patch it!

Regards

### Component(s)

C++

lukester1975:
* Closes: #14869

### DIFF
--- a/cpp/src/arrow/arrow-testing.pc.in
+++ b/cpp/src/arrow/arrow-testing.pc.in
@@ -27,3 +27,4 @@ Version: @ARROW_VERSION@
 Requires: arrow
 Libs: -L${libdir} -larrow_testing
 Cflags: -I${gtest_includedir}
+Cflags.private: -DARROW_TESTING_STATIC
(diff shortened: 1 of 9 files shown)


kou (approved):
+1

PR #14911, opened 2022-12-11
MINOR: [Java] ArrowBuf#setOne should have int64 params

gf2121:
`ArrowBuf#setOne` should have int64 params


java/memory/memory-core/src/main/java/org/apache/arrow/memory/ArrowBuf.java
```diff
@@ -1184,7 +1184,7 @@ public ArrowBuf setZero(long index, long length) {
    * @param length length of bytes to set.
    * @return this ArrowBuf
    */
-  public ArrowBuf setOne(int index, int length) {
+  public ArrowBuf setOne(long index, long length) {
```
lidavidm, line 1187: Could we leave behind an `@Deprecated` version of this with the original signature, then remove it in v12?

lidavidm (approved):

PR #6131, opened 2020-01-07
ARROW-7505: [Java] Remove Netty dependency for ArrowBuf

liyafan82:
This is part of the first step of issue ARROW-4526.
In this step, we remove netty dependency for ArrowBuf, BufferAllocator and ReferenceManager.

In this issue, we remove the dependency for ArrowBuf.
The task for BufferAllocator and ReferenceManager will not start until ARROW-7329 is finished.

PR #14913, opened 2022-12-11
GH-14912: [Java] Remove usage of PlatformDependent in arrow-vector, arrow-jdbc and arrow-algorithm

### LINKED ISSUE #14912: [Java] Remove usage of PlatformDependent in arrow-vector

For some reason our system need to set netty param `io.netty.noUnsafe=true`. We have replaced `arrow-memory-netty` with `arrow-memory-unsafe`. But there are still some utils in `arrow-vector` like `DictionaryEecoder`, `VectorSchemaRootAppender` using `PlatformDependent` which will cause `NullPointerException` when `io.netty.noUnsafe=true` .

IMO netty params should not affect the work of `arrow-vector`, can we replace the usage of `PlatformDependent` by `ArrowBuf#setxxx` ?

### Component(s)

Java

gf2121:
* Closes: #14912

### DIFF
--- a/java/adapter/jdbc/pom.xml
+++ b/java/adapter/jdbc/pom.xml
@@ -86,11 +86,6 @@
             <scope>test</scope>
         </dependency>
 
-        <dependency>
-            <groupId>io.netty</groupId>
-            <artifactId>netty-common</artifactId>
-        </dependency>
-
     </dependencies>
 
 </project>
--- a/java/adapter/jdbc/src/main/java/org/apache/arrow/adapter/jdbc/consumer/ClobConsumer.java
+++ b/java/adapter/jdbc/src/main/java/org/apache/arrow/adapter/jdbc/consumer/ClobConsumer.java
@@ -23,10 +23,9 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 
 import org.apache.arrow.memory.ArrowBuf;
+import org.apache.arrow.memory.util.MemoryUtil;
 import org.apache.arrow.vector.BitVectorHelper;
 import org.apache.arrow.vector.VarCharVector;
-
-import io.netty.util.internal.PlatformDependent;
 
 /**
  * Consumer which consume clob type values from {@link ResultSet}.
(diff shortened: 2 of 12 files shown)


lidavidm (commented):
Since we also maintain a copy of `sun.misc.Unsafe` in [MemoryUtil](https://github.com/apache/arrow/blob/master/java/memory/memory-core/src/main/java/org/apache/arrow/memory/util/MemoryUtil.java), can we add our own `getInt`, `getLong`, etc. there and replace Netty Unsafe with our Unsafe? That way we can at least keep the same behavior as before.

gf2121:
Thanks @lidavidm for feedback!

I think you are meaning to delegate methods on `Unsafe` in `MemoryUtil` (like what `PlatformDependent` do, just discard the params around netty). But i'm also seeing that `ArrowBuf` directly replaced `PlatformDependent.xxx` with `MemoryUtil.UNSAFE.xxx` in https://github.com/apache/arrow/pull/6131. So I wonder if the logic in `PlatformDependent`(like [this](https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/internal/PlatformDependent0.java#L727)) is still wanted:

* If yes, we can introduce delegating methods in this PR and replaces the direct operation on `MemoryUtil#UNSAFE` in a follow up.

* If no, we can just operate on `MemoryUtil#UNSAFE` without introducing extra methods in `MemoryUtil`.

gf2121:
`PlatformDependent` was used in `arrow-algorithm`, `arrow-jdbc`, `arrow-memory-netty`, `arrow-vector`. I've removed the usages in `arrow-algorithm`, `arrow-jdbc` and `arrow-vector`, and removed the dependency of `netty-common` in `arrow-algorithm` and `arrow-jdbc`. It would take a bit more job to remove dependency of `netty-common`  in `arrow-vector` as it is using some Netty collection utils llike `IntObjectMap`. Maybe do that in follow up.

lidavidm (approved):
Thanks!

CC @lwhite1 you may want to glance through this too

lidavidm:
I filed #14936 for dropping Netty entirely.

maosuhan:
How about merging this MR first? We are also blocked by this netty unsafe issue. @lidavidm

lidavidm:
I rebased and will merge after that

PR #14921, opened 2022-12-12
MINOR: [Go][CI] Increase workflow timeout due to benchmarking

zeroshade:
With the added benchmarking running, the Go workflows occasionally take longer than the 30 minutes, so we need to increase the workflow timeout time.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -45,7 +45,7 @@ jobs:
     name: AMD64 Debian 11 Go ${{ matrix.go }}
     runs-on: ubuntu-latest
     if: ${{ !contains(github.event.pull_request.title, 'WIP') }}
-    timeout-minutes: 30
+    timeout-minutes: 60
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


pitrou:
Uh, we run benchmarks on CI builds?

zeroshade:
@pitrou via conbench, @jonkeane and @alistaire47 can explain more as it was their request to set it up this way.

pitrou:
But are trusting GH runners to provide stable runtimes?

alistaire47:
Jon's on PTO at the moment, but AFAIK this setup is a fairly temporary state of affairs. I'm in the middle of a larger project to rearrange how we run benchmarks, and hopefully before too long it won't be too hard to run this through [arrow-benchmarks-ci](https://github.com/voltrondata-labs/arrow-benchmarks-ci) like the rest of the benchmarks (though we'll have to figure out what machines make sense given capacity on that side). Not sure exactly when that will be ready, but likely Jan-Feb.

And this is all based on my assumptions; if Jon says otherwise, I defer to them.

assignUser (approved):
Should be ok with the current CI capacity :+1:

kou:
It seems that this doesn't solve the problem: https://github.com/apache/arrow/actions/runs/3689554605

For example: https://github.com/apache/arrow/actions/runs/3689554605/jobs/6245623568#step:7:1134

```text
*** Test killed with quit: ran too long (11m0s).
```

PR #14922, opened 2022-12-12
ARROW-16266: [R] Add StructArray$create()

thisisnic:


r/R/array.R
```diff
@@ -450,6 +440,44 @@ StructArray <- R6Class("StructArray",
   )
 )
 
+StructArray$create <- function(...) {
+  dots <- list2(...)
```
paleolimbot, line 444: Would starting this with:

```r
batch <- record_batch(...)
```

...allow you to avoid everything below? I think in C++ you can create a struct array from:

```r
std::shared_ptr<arrow::RecordBatch> batch;
arrow::StructArray::Make(batch->columns(), batch->schema()->field_names());
``

Comment (thisisnic) [reply r/R/array.R:444]: I'm not sure how to make that work; I tried start off with:

```
dots <- list2(...)

stopifnot(length(dots) > 0)

if (is.null(names(dots))) {
  names(dots) <- rep_len("", length(dots))
}

batch <- record_batch(dots)
```

but I just end up with

`Error: Unknown: only data frames are allowed as unnamed arguments to be auto spliced`

Comment (wjones127) [reply r/R/array.R:444]: Perhaps the solution is just to have some non-empty default names? perhaps `x1, x2, x3, ...`?

Comment (paleolimbot) [reply r/R/array.R:444]: I'm sorry for forgetting about this thread! I think Will nailed it...the easiest way around `record_batch()`'s auto-splicing of data.frames from unnamed arguments thing is just to give all the arguments names. Maybe using `rlang::dots_list(..., .named = TRUE)`?

``` r

struct_array <- function(...) {
  args <- rlang::dots_list(..., .named = TRUE)
  batch <- record_batch(!!! args)
  array_ptr <- arrow:::allocate_arrow_array()
  schema_ptr <- arrow:::allocate_arrow_schema()
  batch$export_to_c(array_ptr, schema_ptr)
  Array$import_from_c(array_ptr, schema_ptr)
}

struct_array(Array$create(1), Array$create(2))
#> StructArray
#> <struct<Array$create(1): double, Array$create(2): double>>
#> -- is_valid: all not null
#> -- child 0 type: double
#>   [
#>     1
#>   ]
#> -- child 1 type: double
#>   [
#>     2
#>   ]
```

<sup>Created on 2022-12-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

thisisnic: That doesn't work as it adds names to things which shouldn't be named like data.frames.

paleolimbot: Can you reprex your initial problem? (i.e., what usage caused the `Error: Unknown: only data frames are allowed as unnamed arguments to be auto spliced`). I'm wondering if the initial error might be fixed by `record_batch(!!!dots)` instead of `record_batch(dots)`?

thisisnic: Ah thanks, but solved it a little differently. I've ended up sticking with the previous non-record batch approach as I wanted to avoid creating additional R6 objects en route, and this approach seems to work now.

paleolimbot (commented):
This is looking great!

My concern about using `record_batch()` isn't about any love for `record_batch()` per se, but I do want to make sure that using both of them is the same user experience because the object they are generating is functionally identical. The ability to re-use existing `record_batch()` code would also mean that users of `StructArray$create()` would benefit from our existing tests for `record_batch()` which covers things like recycling of length-1 R vectors (and maybe some other things neither of us has considered yet). It's fine to not use it; however, I would prefer that the common pieces of code be reused by both functions somehow (or that if different behaviour is desired that it be documented in a comment somewhere so that somebody like me in the future doesn't come along and obliterate the work you've put in here trying to refactor them into a common constructor).

thisisnic:
Finally got this working better with a combination of suggestions, mind taking another look @paleolimbot ?  I'm intentionally deferring the need for it to work on every possible input seeing as it's not clear that we'd need them.

paleolimbot (approved):
Just a suggestion to fix the failing test! The error that the test is failing with is not good and could be improved for both `StructArray$create()` and `RecordBatch$create()` (no need for a followup ticket unless you're passionate about this one!).

PR #14926, opened 2022-12-13
MINOR: [Docs][Python] Add example of not nullable type in Schema

martin-kokos:
It took me a while to figure out from the docs how to use a not nullable type in a schema. The inclusion of strings "nullable", "not null" right on the Schema page makes it more easy to find.
Example like this also makes it more clear the tuple Schema takes is shorthand version of Field.

### DIFF
--- a/python/pyarrow/types.pxi
+++ b/python/pyarrow/types.pxi
@@ -3270,10 +3270,12 @@ def schema(fields, metadata=None):
     >>> import pyarrow as pa
     >>> pa.schema([
     ...     ('some_int', pa.int32()),
-    ...     ('some_string', pa.string())
+    ...     ('some_string', pa.string()),
+    ...     pa.field('some_required_string', pa.string(), nullable=False)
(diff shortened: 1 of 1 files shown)


python/pyarrow/types.pxi
```diff
@@ -3271,9 +3271,11 @@ def schema(fields, metadata=None):
     >>> pa.schema([
     ...     ('some_int', pa.int32()),
     ...     ('some_string', pa.string())
```
raulcd, line 3273: I am not entirely sure if this will fix the CI error as I have not tested but the trailing comma is missing and might be the causing issue:
https://github.com/apache/arrow/actions/runs/3683826993/jobs/6240509816#step:6:6016
```suggestion
    ...     ('some_string', pa.string()),
```

raulcd (commented):
Thanks for the PR @martin-kokos

kou (approved):
+1

PR #14928, opened 2022-12-13
GH-14927 : [Dev] Crossbow submit does not work with fine grained PATs 

### LINKED ISSUE #14927: [Dev] Crossbow submit does not work with fine grained PATs

While updating the  expired `CROSSBOW_GITHUB_TOKEN` I noticed that the pygit2 login does not work with the new fine grained PATs, as those represent a substantial improvement in security we should fix this.

### Component(s)

Developer Tools

assignUser:
The old variant with token passed as name does only work for classic PATS, passing the token as password works for both classic and fine grained PATs.
* Closes: #14927

### DIFF
--- a/dev/archery/archery/crossbow/core.py
+++ b/dev/archery/archery/crossbow/core.py
@@ -200,7 +200,7 @@ class GitRemoteCallbacks(PygitRemoteCallbacks):
 
         if (allowed_types &
                 pygit2.credentials.GIT_CREDENTIAL_USERPASS_PLAINTEXT):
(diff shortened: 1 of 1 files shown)


raulcd (approved):
I have tested myself. With my previous non-fine grained token I was able to submit a crossbow task with both using token as user and password.
With a new fine-grained token it only worked using the token as a password. Otherwise I got the following error:
```
archery.crossbow.core.CrossbowError: Wrong oauth personal access token
```

PR #14930, opened 2022-12-13
ARROW-17332: [R] error parsing folder path with accent ('c:/Público') in read_csv_arrow

paleolimbot:
This PR fixes a bug that prevented some filenames with non-ASCII characters from being openable. The probable culprit is `normalizePath()`, which does some handling of special characters but does not mark the encoding of its output in a way that cpp11's conversion to `std::string` understands.

Because most of our test environments have UTF-8 as the session encoding, this usually works by accident, and it may work by accident in a latin-1 locale too (judging mostly by the fact that our issue thread is not overflowing with complaints of unopenable files, which may or may not be a good metric).

I've added a test for converting the out the output of `normalizePath()` and making sure it's marked as UTF-8. I'll try to replicate this using Docker, too and see if there's any additional test we could add.

The change here brings arrow in line with what the vroom package does for this operation: https://github.com/tidyverse/vroom/blob/main/R/path.R#L322-L324

### DIFF
--- a/r/tests/testthat/test-filesystem.R
+++ b/r/tests/testthat/test-filesystem.R
@@ -129,21 +129,6 @@ test_that("LocalFileSystem + Selector", {
   expect_equal(sum(types == FileType$Directory), 1L)
 })
 
-test_that("clean_path_abs() marks its output as utf-8", {
-  f <- tempfile(fileext = "Público")
-  on.exit(unlink(f))
-
-  file.create(f)
-  f_utf8 <- enc2utf8(normalizePath(f, winslash = "/"))
-  f_latin1 <- iconv(f_utf8, "utf-8", "latin1")
-
-  expect_identical(clean_path_abs(f_utf8), f_utf8)
-  expect_identical(Encoding(clean_path_abs(f_utf8)), Encoding(f_utf8))
-
-  expect_identical(clean_path_abs(f_latin1), f_latin1)
-  expect_identical(Encoding(clean_path_abs(f_latin1)), Encoding(f_utf8))
-})
-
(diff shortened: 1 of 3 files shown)


r/tests/testthat/test-filesystem.R
```diff
@@ -130,6 +129,21 @@ test_that("LocalFileSystem + Selector", {
...
 
+test_that("clean_path_abs() marks its output as utf-8", {
+  f <- tempfile(fileext = "Público")
+  on.exit(unlink(f))
+
+  file.create(f)
+  f_utf8 <- enc2utf8(normalizePath(f, winslash = "/"))
+  f_latin1 <- iconv(f_utf8, "utf-8", "latin1")
+
+  expect_identical(clean_path_abs(f_utf8), f_utf8)
```
thisisnic, line 140: Just to check, is it intended that this test passes with or without the change made in this PR, as does the similar one below?

paleolimbot: `expect_identical(clean_path_abs(f_utf8), f_utf8)` and `expect_identical(clean_path_abs(f_latin1), f_latin1)` will both pass because `identical()` knows about character encodings; `expect_identical(Encoding(clean_path_abs(f_utf8)), Encoding(f_utf8))` and `expect_identical(Encoding(clean_path_abs(f_latin1)), Encoding(f_utf8))` both fail for me on current master (could be just MacOS).

thisisnic: Nah, the second ones in those groups fail for me too on Ubuntu, it was just that particular expectation I was asking about.

thisisnic (approved):
Great!

PR #14934, opened 2022-12-13
ARROW-18427: [C++] Support negative tolerance in `AsofJoinNode`

rtpsw:
See https://issues.apache.org/jira/browse/ARROW-18427


rtpsw:
cc @icexelloss @westonpace 

My challenge in is this PR was to add support for a negative tolerance without modifying the original code and algorithm beyond recognition. This is not so simple because a negative tolerance requires a future join, which requires per-key entry queues, and the original code was not designed with this in mind. Hopefully, the PR makes sense. The additional and well-covering test cases give me enough confidence in it.

rtpsw:
I should note that there are opportunities to improve performance, in particular around the management of shared pointers to record batches. I could handle this in a separate issue, unless you say as-of-join performance must be ensured in every push to the master branch.

rtpsw:
@westonpace, could you take a look?

westonpace:
> I should note that there are opportunities to improve performance, in particular around the management of shared pointers to record batches. I could handle this in a separate issue, unless you say as-of-join performance must be ensured in every push to the master branch.

Generally, yes.  There should not be regression on master.  Specifically, I think it's probably ok to have a regression here.  We haven't really released the asof join node (I don't see it in the 9.0.0 or 10.0.0 release notes) and there is no mention of it in the documentation.

cpp/src/arrow/compute/exec/asof_join_node.cc
```diff
@@ -140,6 +172,17 @@ struct MemoStore {
   // Stores last known values for all the keys
 
   struct Entry {
+    Entry() = default;
+
+    Entry(OnType time, std::shared_ptr<arrow::RecordBatch> batch, row_index_t row)
+        : time(time), batch(batch), row(row) {}
+
+    void swap(Entry& other) {
```
westonpace, line 180: We generally discourage mutable references but given this is supposed to be an analogue of `std::swap` it seems ok.

westonpace (changes requested):
Sorry for the delay looking at this (Christmas vacation).  This looks like a solid addition, though it will need a rebase so I am marking "request changes".

Is negative tolerance something that is often used?  Wouldn't it be more intuitive to specify an absolute tolerance that works in both directions? (e.g. grab the closest entry within 20ms of the LHS in either direction)?

rtpsw:
> Generally, yes. There should not be regression on master. Specifically, I think it's probably ok to have a regression here. We haven't really released the asof join node (I don't see it in the 9.0.0 or 10.0.0 release notes) and there is no mention of it in the documentation.

I'll see what I can conveniently do to improve.

rtpsw:
> Sorry for the delay looking at this (Christmas vacation). This looks like a solid addition, though it will need a rebase so I am marking "request changes".

Sure.

> Is negative tolerance something that is often used? Wouldn't it be more intuitive to specify an absolute tolerance that works in both directions? (e.g. grab the closest entry within 20ms of the LHS in either direction)?

@icexelloss brought up the negative tolerance, so he should be able to say more about the use cases. Regarding the idea of both-directions, I still can't speak about the use cases, but my guess is the implementation shouldn't be too hard to include in this PR (likely by changing `bool negative` to an enum).

rtpsw:
@icexelloss confirms the use case of negative tolerance is a recurring one, and says that the use case of a double-sided tolerance is not, at least for the time being. @westonpace, could you clarify the work remaining for this PR?

westonpace:
I don't know why the checks didn't run.  Once this is passing I think we can merge it.  Can you close and reopen the PR real quick so the tests run?

rtpsw:
@westonpace, looks like this should be ready to go.

PR #14938, opened 2022-12-13
GH-14937: [C++] Add rank kernel benchmarks

### LINKED ISSUE #14937: [C++] Add rank kernel benchmarks

We should add benchmarks for the `rank` compute function similar to the existing ones for the `sort_indices` compute function.

### Component(s)

C++

pitrou:
* Closes: #14937

### DIFF
--- a/cpp/src/arrow/compute/kernels/vector_sort_benchmark.cc
+++ b/cpp/src/arrow/compute/kernels/vector_sort_benchmark.cc
@@ -29,35 +29,59 @@ namespace arrow {
 namespace compute {
 constexpr auto kSeed = 0x0ff1ce;
 
-static void ArraySortIndicesBenchmark(benchmark::State& state,
-                                      const std::shared_ptr<Array>& values) {
+//
+// Array sort/rank benchmark helpers
+//
+
+struct SortRunner {
+  explicit SortRunner(benchmark::State&) {}
+
+  Status operator()(const std::shared_ptr<Array>& values) const {
+    return SortIndices(*values).status();
+  }
+  Status operator()(const std::shared_ptr<ChunkedArray>& values) const {
+    return SortIndices(*values).status();
+  }
+};
+
+struct RankRunner {
+  explicit RankRunner(benchmark::State& state) {
+    options = RankOptions::Defaults();
+    options.tiebreaker = static_cast<RankOptions::Tiebreaker>(state.range(2));
+  }
+
+  RankOptions options;
+
+  Status operator()(const std::shared_ptr<Array>& values) const {
+    return CallFunction("rank", {values}, &options).status();
+  }
+};
+
+template <typename Runner, typename ArrayLike>
+static void ArraySortFuncBenchmark(benchmark::State& state, const Runner& runner,
+                                   const std::shared_ptr<ArrayLike>& values) {
   for (auto _ : state) {
-    ABORT_NOT_OK(SortIndices(*values).status());
+    ABORT_NOT_OK(runner(values));
   }
   state.SetItemsProcessed(state.iterations() * values->length());
 }
 
-static void ChunkedArraySortIndicesBenchmark(
-    benchmark::State& state, const std::shared_ptr<ChunkedArray>& values) {
-  for (auto _ : state) {
-    ABORT_NOT_OK(SortIndices(*values).status());
-  }
-  state.SetItemsProcessed(state.iterations() * values->length());
-}
-
-static void ArraySortIndicesInt64Benchmark(benchmark::State& state, int64_t min,
-                                           int64_t max) {
+template <typename Runner>
+static void ArraySortFuncInt64Benchmark(benchmark::State& state, const Runner& runner,
+                                        int64_t min, int64_t max) {
   RegressionArgs args(state);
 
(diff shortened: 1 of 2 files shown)


pitrou:
cc @WillAyd

WillAyd (commented):
thanks for the ping! this makes sense to me

cpp/src/arrow/compute/kernels/vector_sort_benchmark.cc
```diff
@@ -301,5 +333,38 @@ BENCHMARK(TableSortIndicesInt64Wide)
...
+
+void ArrayRankSetArgs(benchmark::internal::Benchmark* bench) {
+  // 3 benchmark arguments: size, inverse null proportion, rank tiebreaker
+  bench->Unit(benchmark::kNanosecond);
+  bench->ArgNames({"", "", "tiebreaker"});
+
+  // Use only a subset of kInverseNullProportions as the cartesian product of
+  // arguments is large already.
+  const std::vector<ArgsType> inverse_null_proportions{10, 1, 0};
+  // Don't bother with Max as it should have the same perf as Min
```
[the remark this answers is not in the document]
pitrou: Well, we don't have a `Last` option :-)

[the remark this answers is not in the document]
WillAyd: This is a great point

cpp/src/arrow/compute/kernels/vector_sort_benchmark.cc
```diff
@@ -301,5 +333,38 @@ BENCHMARK(TableSortIndicesInt64Wide)
     })
     ->Unit(benchmark::TimeUnit::kNanosecond);
 
+//
+// Rank benchmark declarations
+//
+
+void ArrayRankSetArgs(benchmark::internal::Benchmark* bench) {
+  // 3 benchmark arguments: size, inverse null proportion, rank tiebreaker
+  bench->Unit(benchmark::kNanosecond);
```
westonpace, line 342: Mostly just a question but is there any particular reason to pin this?  I believe `kNanosecond` is the default and setting it this way prevents the user from adjusting it on the command line.

westonpace (approved):
This looks good to me.  Thanks for adding.

westonpace:
For both rank & sort do we have any kind of UTF8 benchmarks?

cpp/src/arrow/compute/kernels/vector_sort_benchmark.cc
pitrou: Hmm... I blindly reproduced the existing options for SortIndices benchmarks...

But now I see that it was probably because `RegressionSetArgs`, for whatever reason, forces another unit.
Perhaps we should simply remove all these directives. As you point out, they are actually user-hostile while not bringing any obvious value on their own.

pitrou:
> For both rank & sort do we have any kind of UTF8 benchmarks?

We don't.

cpp/src/arrow/compute/kernels/vector_sort_benchmark.cc
westonpace: Ah, I didn't see the setting in `RegrssionSetArgs`.  I don't think it's too big of a deal but agree it might be nice to let these drift and people can pin it all if they care.

That being said, if we back this out, it would probably disturb continuous benchmarking efforts so maybe not worth it.

pitrou: I would hope that continuous benchmarking efforts are keyed on meaningful metrics such as bytes/sec or items/sec, at least where they exist (and IIRC they do).

WillAyd:
I would be happy to try and add that in a follow up

cpp/src/arrow/util/benchmark_util.h
```diff
@@ -46,12 +39,22 @@ struct BenchmarkArgsType<benchmark::internal::Benchmark* (
...
+
+static const CpuInfo* cpu_info = CpuInfo::GetInstance();
+
+static const int64_t kL1Size = cpu_info->CacheSize(CpuInfo::CacheLevel::L1);
+static const int64_t kL2Size = cpu_info->CacheSize(CpuInfo::CacheLevel::L2);
+static const int64_t kL3Size = cpu_info->CacheSize(CpuInfo::CacheLevel::L3);
+static const int64_t kCantFitInL3Size = kL3Size * 4;
+static const std::vector<int64_t> kMemorySizes = {kL1Size, kL2Size, kL3Size,
+                                                  kCantFitInL3Size};
+// 0 is treated as "no nulls"
```
cyb70289, line 55: The reciprocal representation is a bit confusing to me. Maybe we can used "per ten thousand" unit. And move 0 to the first element, so null proportion is increasing.
[0, 1, 100, 1000, 5000, 10000] / 10000.0

cyb70289 (approved):

cpp/src/arrow/util/benchmark_util.h
pitrou: I find it a bit confusing too, but it might be too late to change as well (it would perhaps disrupt benchmark history).

Issue #14931, opened 2022-12-13
pyarrow 10.0.1 is missing libarrow_python.so


david-engelmann:
### Describe the bug, including details regarding any error messages, version, and platform.

Builds with pyarrow as a dependency are not able to build with the most recent release of pyarrow.
The error states the following:
`.../libarrow_python.so needed by ..., missing and no know rule to make it`

If anyone knows how I can resolve this error so I can support pyarrow10, please let me know!

### Component(s)

Python

jorisvandenbossche:
Can you provide a bit more detail? Do you have a problem running pyarrow itself, or with another package depending on pyarrow? (assuming the second since I see your linked turbodbc PR) Is it when packaging another packge? How did you install pyarrow? etc

david-engelmann:
I see that pyarrow is installed in the environment when I debug. pip list shows 

```
...
pyarrow            10.0.1
...
```
Unfortunately, this is for the turbodbc and the issue occurs during the Earthly CI flow which utilizes mamba to install pyarrow. The error occurs when trying to build turbodbc with the following command

` cmake -DBOOST_ROOT=\${CONDA_PREFIX} -DBUILD_COVERAGE=ON -DCMAKE_INSTALL_PREFIX=./dist  -DPYTHON_EXECUTABLE=\$(which python) -GNinja .. && ninja && cmake --build . --target install && cd dist && python setup.py sdist`

I'll include the relevant section of the build failed output below

```
 -- Found PkgConfig: /opt/conda/envs/turbodbc-dev/bin/pkg-config (found version "0.29.2")
              +build *failed* | -- Found /opt/conda/envs/turbodbc-dev/lib/libarrow.so in /opt/conda/envs/turbodbc-dev/lib
              +build *failed* | -- Found the Arrow core library: /opt/conda/envs/turbodbc-dev/lib/libarrow.so
              +build *failed* | -- Found the Arrow Python library: ARROW_PYTHON_LIB_PATH-NOTFOUND
              +build *failed* | -- Detecting unixODBC library
              +build *failed* | --   Found header files at: /opt/conda/envs/turbodbc-dev/include
              +build *failed* | --   Found library at: /opt/conda/envs/turbodbc-dev/lib/libodbc.so
              +build *failed* | CMake Warning (dev) at /opt/conda/envs/turbodbc-dev/share/cmake-3.25/Modules/FindPackageHandleStandardArgs.cmake:438 (message):
[...]

PR #14944, opened 2022-12-14
GH-14943: [Python] Fix pyarrow.get_libraries() order

### LINKED ISSUE #14943: [Python] pyarrow.get_libraries() returns libraries with wrong order

`pyarrow.get_libraries()` returns `['arrow', 'arrow_python']` but it should be `['arrow_python', 'arrow']` because `libarrow_python.so` depends on `libarrow.so`.

### Component(s)

Python

kou:
pyarrow.get_libraries() returns ['arrow', 'arrow_python'] but it should be ['arrow_python', 'arrow'] because libarrow_python.so depends on libarrow.so.
* Closes: #14943

### DIFF
--- a/python/pyarrow/__init__.py
+++ b/python/pyarrow/__init__.py
@@ -415,7 +415,7 @@ def get_libraries():
     Return list of library names to include in the `libraries` argument for C
     or Cython extensions using pyarrow
     """
-    return ['arrow', 'arrow_python']
+    return ['arrow_python', 'arrow']
 
 
 def create_library_symlinks():


jorisvandenbossche:
Just out of curiosity: is this a problem in practice? We already have this for years, so it would seem strange that his has never come up before?

kou:
I don't have a real problem case. I just noticed this in #14931 and https://github.com/blue-yonder/turbodbc/pull/369 .

> We already have this for years, so it would seem strange that this has never come up before?

I suspect nobody uses `pyarrow.get_libraries()` before because we receive multiple `find_package(ArrowPython)` related reports after we release 10.0.0...

david-engelmann:
@kou I reversed the order in my turbo setup.py and it didn't change anything. I think the issue in https://github.com/blue-yonder/turbodbc/pull/369 is the missing libarrow.a and libarrow_python.a files

kou:
Sorry for confusing you. I didn't mean that https://github.com/blue-yonder/turbodbc/pull/369 requires this.
I just wanted to say that I noticed this when I see your code.

david-engelmann:
@kou True, btw, we were able to resolve all the issues occurring in that PR, thank you for all your help!

kou:
If nobody objects this, I'll merge this in next week.

david-engelmann:
> If nobody objects this, I'll merge this in next week.

switching the order in setup.py had no effect on the turbodbc build

Issue #14949, opened 2022-12-14
[CI][Release] Source Release and Merge Script jobs are currently failing on GHA


raulcd:
### Describe the bug, including details regarding any error messages, version, and platform.

This has been raised on Zulip.
The jobs for `Source Release and Merge Script` are failing on several PRs and master. Example of failures:
- https://github.com/apache/arrow/actions/runs/3687535449
- https://github.com/apache/arrow/pull/14926
- https://github.com/apache/arrow/actions/runs/3693047779/jobs/6252584126

From what I understand the error seems to be around the following not being found (but I might be mistaken as the output is quite difficult to follow):
```
 -   :path=>"java/adapter/avro/pom.xml"},
-  {:hunks=>
-    [["-        <version>11.0.0-SNAPSHOT</version>",
-      "+        <version>12.0.0-SNAPSHOT</version>"]],
-   :path=>"java/adapter/jdbc/pom.xml"},
-  {:hunks=>
-    [["-        <version>11.0.0-SNAPSHOT</version>",
-      "+        <version>12.0.0-SNAPSHOT</version>"]],
-   :path=>"java/adapter/orc/pom.xml"},
-  {:hunks=>
-    [["-    <version>11.0.0-SNAPSHOT</version>",
-      "+    <version>12.0.0-SNAPSHOT</version>"]],
```
I can't find any change lately that might be the cause. The first commit on master that seems to fail due to this issue is unrelated: https://github.com/apache/arrow/commit/4647739e46fc1fffb5e89b28f3faee4500a8ef9e

### Component(s)

Continuous Integration

raulcd:
I can't seem to reproduce locally:
```
(pyarrow-dev) ~/code/arrow  (master)  $ git pull upstream master
From github.com:apache/arrow
 * branch            master     -> FETCH_HEAD
Already up to date.
(pyarrow-dev) ~/code/arrow  (master)  $ ci/scripts/release_test.sh $(pwd)
+ arrow_dir=/home/raulcd/code/arrow
+ pushd /home/raulcd/code/arrow
~/code/arrow ~/code/arrow
+ dev/release/run-test.rb
Loaded suite release
Started
.........
Finished in 26.797402892 seconds.
---------------------------------------------------------------------------------------------------------------------------------------------------------------
9 tests, 9 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
---------------------------------------------------------------------------------------------------------------------------------------------------------------
[...]

PR #14954, opened 2022-12-14
GH-14940: [Go][Parquet] Fix Encryption Column writing

### LINKED ISSUE #14940: [Go][Parquet] writer.WriteBatch silently fail when using encryption

Hello, I am trying to leverage the go arrow library to encrypt/decrypt parquet files.
Unfortunately, it seems that the encrypted values are not written in the parquet file when calling the WriteBatch function.
I have written a test function that writes and then reads a parquet file. The function is working correctly without encryption, but when I add encryption, the values for the encrypted columns do not exist. This is confirmed by two reasons:

- The size of the encrypted file is smaller than the unencrypted one
- When I read the encrypted file, the reader.HasNext always returns false, and the number of rows in the row group is 0.

After noticing this behavior and testing different configurations (serial/buffer, encoding, compression), I've looked at the test suite and noticed what I assume is a bug.

I’ve run the test suite and all the tests pass. The decryptions work perfectly for the files coming from the repo https://github.com/apache/parquet-testing, but for the file created in the previous test (the encryption one), the test is actually silently failing.
The number of rows is zero, even if it should be 50. The test does check if the number of rows read equals the number of rows in the stats, but it does not check if there are rows at all.

I am assuming that something is wrong with the flush when using encryption.

I have forked the repo and updated the test to include a check for the number of rows. You can find the forked repo here => https://github.com/ahmb84/arrow/tree/parquet-go-add-test-for-number-of-rows-in-decryption

Let me know if my assumptions are correct or if I am missing anything. Thank you!

### Component(s)

Go, Parquet

zeroshade:
Missed a `default` causing a fall through to always return an error that wasn't being returned through and checked. 

Test updated to validate number of values/rows written. Error is now propagated through if encountered.
* Closes: #14940

### DIFF
--- a/go/parquet/encryption_read_config_test.go
+++ b/go/parquet/encryption_read_config_test.go
@@ -99,6 +99,7 @@ type TestDecryptionSuite struct {
 	colEncryptionKey1   string
 	colEncryptionKey2   string
 	fileName            string
+	rowsPerRG           int
 }
 
 func (d *TestDecryptionSuite) TearDownSuite() {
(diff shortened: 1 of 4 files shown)


zeroshade:
The failed CI are due to #14949

PR #14956, opened 2022-12-15
ARROW-18425: [Substrait] Add Substrait→Acero mapping for round operation

EpsilonPrime:
Add Subtrait→Acero mapping for round operation to match https://github.com/substrait-io/substrait/pull/322.

### DIFF
--- a/cpp/src/arrow/engine/substrait/extension_set.cc
+++ b/cpp/src/arrow/engine/substrait/extension_set.cc
@@ -24,6 +24,7 @@
 #include <sstream>
 #include <unordered_set>
 
+#include "arrow/compute/api.h"
 #include "arrow/engine/substrait/options.h"
 #include "arrow/type.h"
 #include "arrow/type_fwd.h"
@@ -670,10 +671,10 @@ struct ExtensionIdRegistryImpl : ExtensionIdRegistry {
 template <typename Enum>
 class EnumParser {
  public:
-  explicit EnumParser(const std::vector<std::string>& options) {
+  explicit EnumParser(const std::vector<std::string>& options, int offset) {
     for (std::size_t i = 0; i < options.size(); i++) {
-      parse_map_[options[i]] = static_cast<Enum>(i + 1);
-      reverse_map_[static_cast<Enum>(i + 1)] = options[i];
+      parse_map_[options[i]] = static_cast<Enum>(i + offset);
+      reverse_map_[static_cast<Enum>(i + offset)] = options[i];
     }
   }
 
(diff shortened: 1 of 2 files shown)


cpp/src/arrow/engine/substrait/extension_set.cc
```diff
@@ -790,6 +797,34 @@ ExtensionIdRegistry::SubstraitCallToArrow DecodeOptionlessUncheckedArithmetic(
...
+             compute::RoundMode::HALF_TO_ODD},
+            compute::RoundMode::HALF_TOWARDS_INFINITY));
+    ARROW_ASSIGN_OR_RAISE(std::vector<compute::Expression> value_args,
+                          GetValueArgs(call, 0));
+    std::shared_ptr<compute::RoundOptions> options =
+        std::make_shared<compute::RoundOptions>();
+    std::optional<std::vector<std::string> const*> s_arg = call.GetOption("s");
+    if (s_arg.has_value()) {
+      // Substrait will enforce at least one choice is present so this is safe.
+      options->ndigits = std::stol(s_arg.value()->at(0), NULLPTR, 10);
```
EpsilonPrime, line 821: The std::stol call can raise an exception if the option is non-numeric or if the value will overflow ndigits (an int64 value).  Do we have a standard way of checking these values that won't risk an exception?

cpp/src/arrow/engine/substrait/extension_set.cc
```diff
@@ -24,6 +24,7 @@
 #include <sstream>
 #include <unordered_set>
 
+#include "arrow/compute/api.h"
```
westonpace, line 27: We try to avoid including `/api.h` files internally as they tend to be quite large and increase compilation times.  Instead we prefer many smaller includes.  The `api.h` files are for external use and also to help up distinguish public API from private API.

cpp/src/arrow/engine/substrait/extension_set.cc
```diff
@@ -670,10 +671,10 @@ struct ExtensionIdRegistryImpl : ExtensionIdRegistry {
 template <typename Enum>
 class EnumParser {
  public:
-  explicit EnumParser(const std::vector<std::string>& options) {
+  explicit EnumParser(const std::vector<std::string>& options, int offset) {
```
westonpace, line 674: This is a good cleanup.  However, the need for `kUnspecified` came from a time when Substrait handled enums differently.  Another potential change, that might be simpler, would be to remove `kUnspecified` from the other enums.

cpp/src/arrow/engine/substrait/extension_set.cc
westonpace: Probably a moot point since it seems `s` will remain a value argument (and not an option) but the general philosophy is that, if a call might raise an exception like that, we need to check externally to ensure it does not.  Arrow code should not raise exceptions.

We have a suite of string to value parsing functions though.  Should you encounter this in the future the functions in `src/arrow/util/value_parsing.h` should come in handy.  In particular I think this could be a call to `arrow::internal::ParseValue<Int64Type>`

westonpace (changes requested):
This is a bit out of date now (I think we will be sticking with `s` as a value argument) but overall the code looks very solid.  Thanks!

EpsilonPrime:
Obsoleted by https://github.com/apache/arrow/pull/33775.

PR #14959, opened 2022-12-15
ARROW-18437: [C++][Parquet] Fix encoder for DELTA_BINARY_PACKED when flushing more than once

mapleFU:
When flush more than one block, `DELTA_BINARY_PACKED` will be corrupt, because it didn't reset the context after flush.

I'll add some tests this weekend.


cpp/src/parquet/encoding.cc
```diff
@@ -2138,8 +2138,6 @@ class DeltaBitPackEncoder : public EncoderImpl, virtual public TypedEncoder<DTyp
           "but it's " +
           std::to_string(values_per_block % mini_blocks_per_block));
     }
-    // Reserve enough space at the beginning of the buffer for largest possible header.
-    PARQUET_THROW_NOT_OK(sink_.Advance(kMaxPageHeaderWriterSize));
```
wgtmac, line 2142: Please make sure it still works correctly if there are zero values (i.e. an empty block).

mapleFU:
```c++
template <typename DType>
void DeltaBitPackEncoder<DType>::FlushBlock() {
  if (values_current_block_ == 0) {
    return;
  }
```

Don't worry, I will add some tests this weekend...

mapleFU: Oh, you're right, I make this wrong. Let me just make it simple

wgtmac: I was concerned that there is no sufficient space reserved for the header if the block contains empty values. :)

mapleFU:
@wjones127 @pitrou @rok Mind take a took? I don't want to modify testing for a lot, so a just add a `ROUND_TRIP_TIME` to make it build multiple times. /cc @wgtmac

rok (changes requested):
Thanks for looking into this @mapleFU, I didn't realize encoders get reused like this.
I'm adding a couple of naming suggestions.

mapleFU:
Resolved now, thanks @rok

mapleFU:
@pitrou Mind take a look?

pitrou (approved):

mapleFU:
Retrigger CI, seems the macos error is not caused by me

mapleFU:
@pitrou can we merge this patch?

pitrou:
@rok Can you give this a final look and merge if ok?

rok:
Merged. Thanks for noticing and fixing this @mapleFU !

PR #14962, opened 2022-12-15
GH-14961: [Ruby] Use newer extpp for C++17

### LINKED ISSUE #14961: [Ruby] Use newer extpp for C++17

Recent extpp support C++17. We can remove C++17 workaround with recent extpp.

### Component(s)

Ruby

kou:
Recent extpp support C++17. We can remove C++17 workaround with recent extpp.
* Closes: #14961


kou:
+1

PR #14964, opened 2022-12-15
GH-33596: [C++][Parquet] Parquet page index read support

### LINKED ISSUE #33596: [C++][Parquet] Parquet page index read support
Implement read support for parquet page index and expose it from the reader API.

**Reporter**: [Gang Wu](https://issues.apache.org/jira/browse/ARROW-18434) / @wgtmac
**Assignee**: [Gang Wu](https://issues.apache.org/jira/browse/ARROW-18434) / @wgtmac
#### PRs and other links:
- [GitHub Pull Request #14964](https://github.com/apache/arrow/pull/14964)

<sub>**Note**: *This issue was originally created as [ARROW-18434](https://issues.apache.org/jira/browse/ARROW-18434). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

wgtmac:
Basically, the patch provides following implementation:
- Define `class RowGroupPageIndexReader` to read page index from a parquet row group. It internally leverages implementation from Apache Impala [link](https://github.com/apache/impala/blob/efa426453a8af3728bc272b9158f5564ce37e0ea/be/src/exec/parquet/parquet-page-index.cc#L41) to merge I/O chunks of page index in the same row group.
- Define `class PageIndexReader` to create `RowGroupPageIndexReader` for each row group.
- `ParquetFileReader` internally creates and caches a single `PageIndexReader` object and exposes it to the end user.

Limitation:
- Reading page index from encrypted parquet file is not yet supported. It takes some effort to understand the specs and will be done in a separate patch after the completion of write logic of page index.
* Closes: #33596

### DIFF
--- a/cpp/src/parquet/file_reader.cc
+++ b/cpp/src/parquet/file_reader.cc
@@ -305,7 +305,12 @@ class SerializedFile : public ParquetFileReader::Contents {
 
   std::shared_ptr<PageIndexReader> GetPageIndexReader() override {
     if (!file_metadata_) {
-      throw ParquetException("Cannot get PageIndexReader as file metadata is not ready");
+      // Usually this won't happen if user calls one of the static Open() functions
+      // to create a ParquetFileReader instance. But if user calls the constructor
+      // directly and calls GetPageIndexReader() before Open() then this could happen.
+      throw ParquetException(
+          "Cannot call GetPageIndexReader() due to missing file metadata. Did you "
+          "forget to call ParquetFileReader::Open() first?");
     }
     if (!page_index_reader_) {
       page_index_reader_ = PageIndexReader::Make(source_.get(), file_metadata_,
--- a/cpp/src/parquet/file_reader.cc
+++ b/cpp/src/parquet/file_reader.cc
@@ -303,7 +303,7 @@ class SerializedFile : public ParquetFileReader::Contents {
 
   std::shared_ptr<FileMetaData> metadata() const override { return file_metadata_; }
 
-  std::shared_ptr<PageIndexReader> page_index_reader() override {
+  std::shared_ptr<PageIndexReader> GetPageIndexReader() override {
     if (!file_metadata_) {
       throw ParquetException("Cannot get PageIndexReader as file metadata is not ready");
     }
@@ -797,7 +797,7 @@ std::shared_ptr<FileMetaData> ParquetFileReader::metadata() const {
 }
 
 std::shared_ptr<PageIndexReader> ParquetFileReader::page_index_reader() {
-  return contents_->page_index_reader();
+  return contents_->GetPageIndexReader();
 }
 
 std::shared_ptr<RowGroupReader> ParquetFileReader::RowGroup(int i) {
--- a/cpp/src/parquet/file_reader.cc
+++ b/cpp/src/parquet/file_reader.cc
@@ -40,6 +40,7 @@
 #include "parquet/exception.h"
 #include "parquet/file_writer.h"
 #include "parquet/metadata.h"
+#include "parquet/page_index.h"
 #include "parquet/platform.h"
 #include "parquet/properties.h"
 #include "parquet/schema.h"
@@ -302,6 +303,17 @@ class SerializedFile : public ParquetFileReader::Contents {
 
   std::shared_ptr<FileMetaData> metadata() const override { return file_metadata_; }
 
+  std::shared_ptr<PageIndexReader> page_index_reader() override {
+    if (!file_metadata_) {
+      throw ParquetException("Cannot get PageIndexReader as file metadata is not ready");
+    }
+    if (!page_index_reader_) {
+      page_index_reader_ = PageIndexReader::Make(source_.get(), file_metadata_,
+                                                 properties_, file_decryptor_);
+    }
+    return page_index_reader_;
+  }
+
   void set_metadata(std::shared_ptr<FileMetaData> metadata) {
     file_metadata_ = std::move(metadata);
   }
@@ -522,7 +534,7 @@ class SerializedFile : public ParquetFileReader::Contents {
   int64_t source_size_;
   std::shared_ptr<FileMetaData> file_metadata_;
   ReaderProperties properties_;
-
+  std::shared_ptr<PageIndexReader> page_index_reader_;
   std::shared_ptr<InternalFileDecryptor> file_decryptor_;
 
   // \return The true length of the metadata in bytes
@@ -782,6 +794,10 @@ void ParquetFileReader::Close() {
 
 std::shared_ptr<FileMetaData> ParquetFileReader::metadata() const {
   return contents_->metadata();
+}
+
+std::shared_ptr<PageIndexReader> ParquetFileReader::page_index_reader() {
+  return contents_->page_index_reader();
 }
 
 std::shared_ptr<RowGroupReader> ParquetFileReader::RowGroup(int i) {
--- a/cpp/src/parquet/file_reader.cc
+++ b/cpp/src/parquet/file_reader.cc
@@ -796,7 +796,7 @@ std::shared_ptr<FileMetaData> ParquetFileReader::metadata() const {
   return contents_->metadata();
 }
 
-std::shared_ptr<PageIndexReader> ParquetFileReader::page_index_reader() {
+std::shared_ptr<PageIndexReader> ParquetFileReader::GetPageIndexReader() {
   return contents_->GetPageIndexReader();
 }
 
--- a/cpp/src/parquet/file_reader.h
+++ b/cpp/src/parquet/file_reader.h
@@ -32,6 +32,7 @@ namespace parquet {
 
 class ColumnReader;
 class FileMetaData;
+class PageIndexReader;
 class PageReader;
 class RowGroupMetaData;
 
@@ -98,6 +99,7 @@ class PARQUET_EXPORT ParquetFileReader {
     virtual void Close() = 0;
     virtual std::shared_ptr<RowGroupReader> GetRowGroup(int i) = 0;
     virtual std::shared_ptr<FileMetaData> metadata() const = 0;
+    virtual std::shared_ptr<PageIndexReader> page_index_reader() = 0;
   };
 
   ParquetFileReader();
@@ -132,6 +134,9 @@ class PARQUET_EXPORT ParquetFileReader {
 
   // Returns the file metadata. Only one instance is ever created
   std::shared_ptr<FileMetaData> metadata() const;
+
+  // Returns the page index reader. Only one instance is ever created
+  std::shared_ptr<PageIndexReader> page_index_reader();
 
   /// Pre-buffer the specified column indices in all row groups.
   ///
--- a/cpp/src/parquet/file_reader.h
+++ b/cpp/src/parquet/file_reader.h
@@ -99,12 +99,6 @@ class PARQUET_EXPORT ParquetFileReader {
     virtual void Close() = 0;
     virtual std::shared_ptr<RowGroupReader> GetRowGroup(int i) = 0;
     virtual std::shared_ptr<FileMetaData> metadata() const = 0;
-    // Returns a PageIndexReader instance for the parquet file.
-    // If the file does not have page index, nullptr may be returned. Since it pays to
-    // check existence of page index in the file, it is possible to return a non-nullptr
-    // even if page index does not exist. It is the caller's responsibility to check
-    // the return value of follow-up calls to PageIndexReader.
-    // WARNING: The returned PageIndexReader must not outlive the ParquetFileReader.
     virtual std::shared_ptr<PageIndexReader> GetPageIndexReader() = 0;
   };
 
@@ -141,8 +135,16 @@ class PARQUET_EXPORT ParquetFileReader {
   // Returns the file metadata. Only one instance is ever created
   std::shared_ptr<FileMetaData> metadata() const;
 
-  // Returns the page index reader. Only one instance is ever created
-  std::shared_ptr<PageIndexReader> page_index_reader();
+  /// Returns the PageIndexReader. Only one instance is ever created.
+  ///
+  /// If the file does not have the page index, nullptr may be returned.
+  /// Because it pays to check existence of page index in the file, it
+  /// is possible to return a non null value even if page index does
+  /// not exist. It is the caller's responsibility to check the return
+  /// value and follow-up calls to PageIndexReader.
+  ///
+  /// WARNING: The returned PageIndexReader must not outlive the ParquetFileReader.
+  std::shared_ptr<PageIndexReader> GetPageIndexReader();
 
   /// Pre-buffer the specified column indices in all row groups.
   ///
--- a/cpp/src/parquet/file_reader.h
+++ b/cpp/src/parquet/file_reader.h
@@ -99,7 +99,13 @@ class PARQUET_EXPORT ParquetFileReader {
     virtual void Close() = 0;
     virtual std::shared_ptr<RowGroupReader> GetRowGroup(int i) = 0;
     virtual std::shared_ptr<FileMetaData> metadata() const = 0;
-    virtual std::shared_ptr<PageIndexReader> page_index_reader() = 0;
+    // Returns a PageIndexReader instance for the parquet file.
+    // If the file does not have page index, nullptr may be returned. Since it pays to
+    // check existence of page index in the file, it is possible to return a non-nullptr
+    // even if page index does not exist. It is the caller's responsibility to check
+    // the return value of follow-up calls to PageIndexReader.
+    // WARNING: The returned PageIndexReader must not outlive the ParquetFileReader.
+    virtual std::shared_ptr<PageIndexReader> GetPageIndexReader() = 0;
   };
 
   ParquetFileReader();
--- a/cpp/src/parquet/page_index.cc
+++ b/cpp/src/parquet/page_index.cc
@@ -338,12 +338,10 @@ class PageIndexReaderImpl : public PageIndexReader {
       auto read_range = PageIndexReader::DeterminePageIndexRangesInRowGroup(
           *file_metadata_->RowGroup(row_group_ordinal));
       if (index_selection.column_index && read_range.column_index.has_value()) {
-        read_ranges.emplace_back(::arrow::io::ReadRange{read_range.column_index->offset,
-                                                        read_range.column_index->length});
+        read_ranges.push_back(*read_range.column_index);
       }
       if (index_selection.offset_index && read_range.offset_index.has_value()) {
-        read_ranges.emplace_back(::arrow::io::ReadRange{read_range.offset_index->offset,
-                                                        read_range.offset_index->length});
+        read_ranges.push_back(*read_range.offset_index);
       }
     }
     PARQUET_IGNORE_NOT_OK(input_->WillNeed(read_ranges));
@@ -392,12 +390,10 @@ RowGroupIndexReadRange PageIndexReader::DeterminePageIndexRangesInRowGroup(
 
   RowGroupIndexReadRange read_range;
   if (ci_end != -1) {
-    read_range.column_index =
-        RowGroupIndexReadRange::ReadRange{ci_start, ci_end - ci_start};
+    read_range.column_index = {ci_start, ci_end - ci_start};
   }
   if (oi_end != -1) {
-    read_range.offset_index =
-        RowGroupIndexReadRange::ReadRange{oi_start, oi_end - oi_start};
+    read_range.offset_index = {oi_start, oi_end - oi_start};
   }
   return read_range;
 }
--- a/cpp/src/parquet/page_index.cc
+++ b/cpp/src/parquet/page_index.cc
@@ -18,6 +18,7 @@
 #include "parquet/page_index.h"
 #include "parquet/encoding.h"
 #include "parquet/exception.h"
+#include "parquet/metadata.h"
 #include "parquet/schema.h"
 #include "parquet/statistics.h"
 #include "parquet/thrift_internal.h"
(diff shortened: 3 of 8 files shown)


wgtmac:
@emkornfield @pitrou Can you please take a look when you get the chance? Thanks!

cpp/src/parquet/page_index.cc
```diff
@@ -184,8 +185,241 @@ class OffsetIndexImpl : public OffsetIndex {
...
 
+class RowGroupPageIndexReaderImpl : public RowGroupPageIndexReader {
+ public:
+  RowGroupPageIndexReaderImpl(::arrow::io::RandomAccessFile* input,
+                              std::shared_ptr<RowGroupMetaData> row_group_metadata,
+                              const ReaderProperties& properties,
+                              int32_t row_group_ordinal,
+                              std::shared_ptr<InternalFileDecryptor> file_decryptor)
+      : input_(input),
+        row_group_metadata_(row_group_metadata),
```
emkornfield, line 196: nit: std::move

cpp/src/parquet/page_index.h
```diff
@@ -126,4 +130,75 @@ class PARQUET_EXPORT OffsetIndex {
...
+/// \brief Interface for reading the page index from a Parquet row group.
+class PARQUET_EXPORT RowGroupPageIndexReader {
+ public:
+  virtual ~RowGroupPageIndexReader() = default;
+
+  /// \brief Read column index of a column chunk.
+  ///
+  /// \param[in] i column id of the column chunk.
+  /// \returns error Result if either column id is invalid or column index does not exist.
+  virtual ::arrow::Result<std::shared_ptr<ColumnIndex>> GetColumnIndex(int32_t i) = 0;
```
emkornfield, line 142: Sorry if I missed this in prior reviews, but core parquet code uses exceptions generally instead of result.  It probably pays to be consistent.

cpp/src/parquet/page_index.h
```diff
@@ -126,4 +130,75 @@ class PARQUET_EXPORT OffsetIndex {
...
+  /// \param[out] column_index_start Base start of column index of all column chunks.
+  /// \param[out] column_index_size Total size of column index of all column chunks.
+  /// \param[out] offset_index_start Base start of offset index of all column chunks.
+  /// \param[out] offset_index_size Total size of offset index of all column chunks.
+  /// \param[out] has_column_index Returns true when at least a partial column index are
+  /// found. Returns false when there is absolutely no column index for the row group.
+  /// \param[out] has_offset_index Returns true when at least a partial offset index are
+  /// found. Returns false when there is absolutely no offsets index for the row group.
+  static void DeterminePageIndexRangesInRowGroup(
+      const RowGroupMetaData& row_group_metadata, int64_t* column_index_start,
```
emkornfield, line 199: could we consolidate output parameters into a struct?  It would probably be better to simply return that struct instead of having multiple output values.  I seem to recall defining a struct already for "start + size" for the indexes, could we re-use that?

cpp/src/parquet/file_reader.h
```diff
@@ -98,6 +99,7 @@ class PARQUET_EXPORT ParquetFileReader {
     virtual void Close() = 0;
     virtual std::shared_ptr<RowGroupReader> GetRowGroup(int i) = 0;
     virtual std::shared_ptr<FileMetaData> metadata() const = 0;
+    virtual std::shared_ptr<PageIndexReader> page_index_reader() = 0;
```
emkornfield, line 102: nit: should probably be const OR be GetPageIndexReader to be the same as GetRowGroup.   We should document behavior if Page indices are not written.

emkornfield: Same comment applies below.

cpp/src/parquet/page_index.h
```diff
@@ -126,4 +130,75 @@ class PARQUET_EXPORT OffsetIndex {
   virtual const std::vector<PageLocation>& page_locations() const = 0;
 };
 
+/// \brief Interface for reading the page index from a Parquet row group.
```
emkornfield, line 133:
```suggestion
/// \brief Interface for reading the page index for a Parquet row group.
```
the previous phrasing makes it sound like page index is stored within rowgroups (at least to me)

cpp/src/parquet/page_index.h
```diff
@@ -126,4 +130,75 @@ class PARQUET_EXPORT OffsetIndex {
...
+};
+
+/// \brief Interface for reading the page index from a Parquet file.
+class PARQUET_EXPORT PageIndexReader {
+ public:
+  virtual ~PageIndexReader() = default;
+
+  /// \brief Create a PageIndexReader instance.
+  static std::shared_ptr<PageIndexReader> Make(
+      ::arrow::io::RandomAccessFile* input, std::shared_ptr<FileMetaData> file_metadata,
```
emkornfield, line 158: ownership/lifetime semantics of passing a raw RandomAccessFile needs to be clarified.

cpp/src/parquet/page_index.h
```diff
@@ -126,4 +130,75 @@ class PARQUET_EXPORT OffsetIndex {
...
+  ///
+  /// The PageIndexReader implementation can optionally prefetch and cache page index
+  /// that may be read later. Follow-up read should not fail even if WillNeed() is not
+  /// called, or the requested page index is out of range from WillNeed() call.
+  ///
+  /// \param[in] row_group_indices list of row group ordinal to read page index later.
+  /// \param[in] need_column_index tell if column index is required later.
+  /// \param[in] need_offset_index tell if offset index is required later.
+  virtual void WillNeed(const std::vector<int32_t>& row_group_indices,
+                        bool need_column_index, bool need_offset_index) = 0;
```
emkornfield, line 177: consider making this a struct:
struct IndexSelection {
  bool need_column_index = false;
  bool need_offset_index = false;
}

This allows for nicer syntax `WillNeed(row_group_indices, {.need_column_index=true, .need_offset_index = true}`, once C++20 is supported in the code base.

cpp/src/parquet/page_index.cc
```diff
@@ -184,8 +185,241 @@ class OffsetIndexImpl : public OffsetIndex {
...
 
+void PageIndexReader::DeterminePageIndexRangesInRowGroup(
+    const RowGroupMetaData& row_group_metadata, int64_t* column_index_start,
+    int64_t* column_index_size, int64_t* offset_index_start, int64_t* offset_index_size,
+    bool* has_column_index, bool* has_offset_index) {
+  int64_t ci_start = std::numeric_limits<int64_t>::max();
+  int64_t oi_start = std::numeric_limits<int64_t>::max();
+  int64_t ci_end = -1;
+  int64_t oi_end = -1;
+  for (int i = 0; i < row_group_metadata.num_columns(); ++i) {
```
emkornfield, line 388: should there be a version of this method that also takes in the columns selected?

cpp/src/parquet/page_index.cc
```diff
@@ -184,8 +185,241 @@ class OffsetIndexImpl : public OffsetIndex {
...
+  std::shared_ptr<InternalFileDecryptor> file_decryptor_;
+};
+
 }  // namespace
 
+void PageIndexReader::DeterminePageIndexRangesInRowGroup(
+    const RowGroupMetaData& row_group_metadata, int64_t* column_index_start,
+    int64_t* column_index_size, int64_t* offset_index_start, int64_t* offset_index_size,
+    bool* has_column_index, bool* has_offset_index) {
+  int64_t ci_start = std::numeric_limits<int64_t>::max();
```
emkornfield, line 384: per comment above consider making a byterange struct here (@pitrou do you recall if something like this exists already in parquet or Arrow)?

cpp/src/parquet/page_index.cc
```diff
@@ -184,8 +185,241 @@ class OffsetIndexImpl : public OffsetIndex {
...
+  int64_t oi_start = std::numeric_limits<int64_t>::max();
+  int64_t ci_end = -1;
+  int64_t oi_end = -1;
+  for (int i = 0; i < row_group_metadata.num_columns(); ++i) {
+    auto col_chunk = row_group_metadata.ColumnChunk(i);
+
+    auto column_index_location = col_chunk->GetColumnIndexLocation();
+    if (column_index_location.has_value()) {
```
emkornfield, line 392: consider refactoring this logic to be consolidated for page/offset index.

cpp/src/parquet/page_index.h
emkornfield: Also, maybe nullptr or optional should be returned if the column index doesn't exist, since this doesn't seem like an error case unless there is a different way for checking which columns have indeces present before this code is run.

emkornfield:
@wgtmac took a pass through mostly looking at interfaces, let me know what you think of the suggestions.  Sorry for the delay, catching up after the holidays.

cpp/src/parquet/file_reader.h
wgtmac: Renamed it to GetPageIndexReader and added comment in detail.

cpp/src/parquet/page_index.h
wgtmac: Use exception instead of status for the return type. Now it returns nullptr if page index does not exist and throw if column ordinal or block ordinal is out of bound.

cpp/src/parquet/page_index.h
wgtmac: What about `::arrow::io::ReadRange`?

cpp/src/parquet/page_index.h
wgtmac: The return value must not outlive any of the input parameters. Added comment to explain it.

cpp/src/parquet/page_index.h
wgtmac: Fixed

cpp/src/parquet/page_index.cc
wgtmac: That is feasible. However, it would complicate the implementation if user has provided selected columns and then requests to read page index of a non-selected column.

cpp/src/parquet/page_index.cc
wgtmac: What about `::arrow::io::ReadRange`?

wgtmac: I have consolidated the return value to a single struct `RowGroupIndexReadRange` which internally leverages `::arrow::io::ReadRange`. It looks better now.

cpp/src/parquet/page_index.h
```diff
@@ -126,4 +132,94 @@ class PARQUET_EXPORT OffsetIndex {
...
+};
+
+struct IndexSelection {
+  /// Specifies whether to read the column index.
+  bool column_index = false;
+  /// Specifies whether to read the offset index.
+  bool offset_index = false;
+};
+
+struct RowGroupIndexReadRange {
```
wgtmac, line 162: What about this? @emkornfield

wgtmac:
I have added a reader test to cover the new interface. Now it is complete and ready to review. Any feedback is appreciated. @pitrou @emkornfield @wjones127

emkornfield:
Sorry for delay will be looking at this by EOD friday

cpp/src/parquet/page_index.cc
emkornfield: it feels like this optimization could be worth it, especially in the tail.  I've observed parquet files with O(1000) of columns, it isn't clear to me if page indexes are generally enabled for all columns or not though.  I think if we provided this method, then we simply document behavior as throwing an exception if the request wasn't requested, but it seems fine todo as a follow-up if we want it.

cpp/src/parquet/file_reader.cc
```diff
@@ -302,6 +303,17 @@ class SerializedFile : public ParquetFileReader::Contents {
 
   std::shared_ptr<FileMetaData> metadata() const override { return file_metadata_; }
 
+  std::shared_ptr<PageIndexReader> GetPageIndexReader() override {
+    if (!file_metadata_) {
+      throw ParquetException("Cannot get PageIndexReader as file metadata is not ready");
```
emkornfield, line 308: nit: can we instruct the api caller what went wrong here (i.e. should they call a certain method?)

cpp/src/parquet/page_index.cc
```diff
@@ -184,8 +185,219 @@ class OffsetIndexImpl : public OffsetIndex {
...
+  std::shared_ptr<ColumnIndex> GetColumnIndex(int32_t i) override {
+    if (i < 0 || i >= row_group_metadata_->num_columns()) {
+      throw ParquetException("Invalid column {} to get column index", i);
+    }
+
+    auto col_chunk = row_group_metadata_->ColumnChunk(i);
+
+    std::unique_ptr<ColumnCryptoMetaData> crypto_metadata = col_chunk->crypto_metadata();
+    if (crypto_metadata != nullptr && file_decryptor_ == nullptr) {
+      ParquetException::NYI("Cannot read encrypted column index yet");
```
emkornfield, line 212: given this error should the second part of the check above be removed?

cpp/src/parquet/page_index.cc
```diff
@@ -184,8 +185,219 @@ class OffsetIndexImpl : public OffsetIndex {
...
+      PARQUET_ASSIGN_OR_THROW(column_index_buffer_,
+                              input_->ReadAt(index_read_range_.column_index->offset,
+                                             index_read_range_.column_index->length));
+    }
+
+    auto buffer = column_index_buffer_.get();
+    int64_t buffer_offset =
+        column_index_location->offset - index_read_range_.column_index->offset;
+    uint32_t length = static_cast<uint32_t>(column_index_location->length);
+    DCHECK_GE(buffer_offset, 0);
```
emkornfield, line 234: if location->length is derived from data in the parquet file, we should be checking eagerly that the length is within the allowed range.

Could we also add a comment on why the down-cast

cpp/src/parquet/page_index.cc
```diff
@@ -184,8 +185,219 @@ class OffsetIndexImpl : public OffsetIndex {
...
+    DCHECK_LE(buffer_offset + length, index_read_range_.column_index->length);
+
+    auto descr = row_group_metadata_->schema()->Column(i);
+    std::shared_ptr<ColumnIndex> column_index;
+    try {
+      column_index =
+          ColumnIndex::Make(*descr, buffer->data() + buffer_offset, length, properties_);
+    } catch (...) {
+      throw ParquetException("Cannot deserialize column index for column {}", i);
```
emkornfield, line 243: seems like we should propogate information about the caught error here?

cpp/src/parquet/page_index.cc
```diff
@@ -184,8 +185,219 @@ class OffsetIndexImpl : public OffsetIndex {
...
+  /// Read offset index of a column chunk.
+  std::shared_ptr<OffsetIndex> GetOffsetIndex(int32_t i) override {
+    if (i < 0 || i >= row_group_metadata_->num_columns()) {
+      throw ParquetException("Invalid column {} to get offset index", i);
+    }
+
+    auto col_chunk = row_group_metadata_->ColumnChunk(i);
+
+    std::unique_ptr<ColumnCryptoMetaData> crypto_metadata = col_chunk->crypto_metadata();
+    if (crypto_metadata != nullptr && file_decryptor_ == nullptr) {
```
emkornfield, line 257: same comment about removing file_decriptor_ check?

cpp/src/parquet/page_index.cc
```diff
@@ -184,8 +185,219 @@ class OffsetIndexImpl : public OffsetIndex {
...
+    if (offset_index_buffer_ == nullptr) {
+      PARQUET_ASSIGN_OR_THROW(offset_index_buffer_,
+                              input_->ReadAt(index_read_range_.offset_index->offset,
+                                             index_read_range_.offset_index->length));
+    }
+
+    auto buffer = offset_index_buffer_.get();
+    int64_t buffer_offset =
+        offset_index_location->offset - index_read_range_.offset_index->offset;
+    uint32_t length = static_cast<uint32_t>(offset_index_location->length);
```
emkornfield, line 279: same comment as above on casts.

cpp/src/parquet/page_index.cc
```diff
@@ -184,8 +185,219 @@ class OffsetIndexImpl : public OffsetIndex {
...
+      throw ParquetException("Missing column index read range");
+    }
+
+    if (offset_index_buffer_ == nullptr) {
+      PARQUET_ASSIGN_OR_THROW(offset_index_buffer_,
+                              input_->ReadAt(index_read_range_.offset_index->offset,
+                                             index_read_range_.offset_index->length));
+    }
+
+    auto buffer = offset_index_buffer_.get();
```
emkornfield, line 276: is this a pointer? could we at least to auto* i not spell out the whle type.

cpp/src/parquet/page_index.cc
```diff
@@ -184,8 +185,219 @@ class OffsetIndexImpl : public OffsetIndex {
...
+    uint32_t length = static_cast<uint32_t>(offset_index_location->length);
+    DCHECK_GE(buffer_offset, 0);
+    DCHECK_LE(buffer_offset + length, index_read_range_.offset_index->length);
+
+    std::shared_ptr<OffsetIndex> offset_index;
+    try {
+      offset_index =
+          OffsetIndex::Make(buffer->data() + buffer_offset, length, properties_);
+    } catch (...) {
+      throw ParquetException("Cannot deserialize offset index for column {}", i);
```
emkornfield, line 288: propogate exception information from the caugh tone?

cpp/src/parquet/page_index.cc
```diff
@@ -184,8 +185,219 @@ class OffsetIndexImpl : public OffsetIndex {
...
+      auto read_range = PageIndexReader::DeterminePageIndexRangesInRowGroup(
+          *file_metadata_->RowGroup(row_group_ordinal));
+      if (index_selection.column_index && read_range.column_index.has_value()) {
+        read_ranges.push_back(*read_range.column_index);
+      }
+      if (index_selection.offset_index && read_range.offset_index.has_value()) {
+        read_ranges.push_back(*read_range.offset_index);
+      }
+    }
+    PARQUET_IGNORE_NOT_OK(input_->WillNeed(read_ranges));
```
emkornfield, line 347: could you add a comment on why this is being ignored.

cpp/src/parquet/page_index.h
emkornfield: Yes, thank you!

cpp/src/parquet/page_index.h
```diff
@@ -126,4 +132,94 @@ class PARQUET_EXPORT OffsetIndex {
...
+  /// \brief Advise the reader which part of page index will be read later.
+  ///
+  /// The PageIndexReader implementation can optionally prefetch and cache page index
+  /// that may be read later. Follow-up read should not fail even if WillNeed() is not
+  /// called, or the requested page index is out of range from WillNeed() call.
+  ///
+  /// \param[in] row_group_indices list of row group ordinal to read page index later.
+  /// \param[in] index_selection tell if any of the page index is required later.
+  virtual void WillNeed(const std::vector<int32_t>& row_group_indices,
+                        IndexSelection index_selection) = 0;
```
emkornfield, line 205: nit, even though this might be more effient, probably pays to use const IndexSelection& as the formal parameter.

emkornfield (changes requested):
Need to review tests in a little bit more detail and left some comments but this is looking good thank you for the changes.  @pitrou did you want  review this at all?

cpp/src/parquet/file_reader.cc
wgtmac: Usually this won't happen if user calls one of the static Open() functions to create a ParquetFileReader instance. But if user calls the constructor directly and calls GetPageIndexReader() before Open() then this could happen. Add a comment to explain it and makes the exception message clear.

cpp/src/parquet/page_index.cc
wgtmac: `ThriftDeserializer::DeserializeMessage` requires a `uint32_t*` to be the in/out parameter for the length of thrift message. And `ColumnIndex::Make()` simply reuse uint32_t for simplicity. BTW this is not a down-cast, column_index_location->length is in the type of `int32_t`.

cpp/src/parquet/page_index.cc
wgtmac: I was thinking that the WillNeed call to the input stream is not required (served as a hint) and then we probably don't care about its failure. But this may hide and postpone the IO issue. Changed it to use `PARQUET_THROW_NOT_OK`.

cpp/src/parquet/page_index.cc
wgtmac: I agree. Remove the try/catch block and simply let it throw.

cpp/src/parquet/page_index.cc
wgtmac: I agree that if a parquet file contains many columns, it would be painful to coalesce them into a single I/O chunk. In our use case to optimize reading file tail, we simply cache the coalesced chunk of all columns for repeated read because it is difficult to predict which columns will be accessed next. I can add a function to optionally specify requested columns and by default all columns are requested.

cpp/src/parquet/page_index.cc
wgtmac: Removed unnecessary check

cpp/src/parquet/page_index.cc
wgtmac: Removed unnecessary check

cpp/src/parquet/page_index.cc
wgtmac: Added a check to validate the range.

cpp/src/parquet/page_index.h
wgtmac: I'm sorry but I didn't get your point here. I haven't used the const reference as the input.

cpp/src/parquet/page_index.cc
wgtmac: @pitrou Do you have the chance to take a look into this? I'd like to hear your advice on the proposed `WillNeed` and `WillNotNeed` interfaces we have discussed here. Especially the granularity of row group or column.

wgtmac:
I have added both `row_group_ordinals` and `column_ordinals` to the `WillNeed` function. The comment has been modified to explain the new contract. Please let me know if anything is unclear. @emkornfield

cpp/src/parquet/page_index.cc
lidavidm: I only took a brief look but I think the current granularity of row group + column index is sufficient. That was the scenario we benchmarked (selecting columns out of a file) and we didn't need anything finer-grained. (IIRC, we ended up with only a single row group per file, too.)

wgtmac:
**C++ / AMD64 Ubuntu 22.04 C++ ASAN UBSAN (pull_request) Cancelled after 75m**

@kou Some checks hit the timeout limit of 75 min and it happens a lot. Does it make sense to increase the limit in the workflow file?

kou:
The job finished in 1h 5m 4s on the current master: https://github.com/apache/arrow/actions/runs/4038548251/jobs/6942621768
Why execution time of the job was increased? Are changes in this pull request related?

wgtmac:
> The job finished in 1h 5m 4s on the current master: https://github.com/apache/arrow/actions/runs/4038548251/jobs/6942621768 Why execution time of the job was increased? Are changes in this pull request related?

There is nothing special in this PR. I have seen the same timeout for several times.

I have checked other PRs and found the same issue here: https://github.com/apache/arrow/actions/runs/4035720668/jobs/6937808549

Perhaps the code base is growing and it is time to rethink the timeout limit. @kou

kou:
OK. Could you open a new issue for this? We can ignore the timeout in this pull request.

wgtmac:
> OK. Could you open a new issue for this? We can ignore the timeout in this pull request.

Opened https://github.com/apache/arrow/issues/33920

wgtmac:
@emkornfield Gentle ping :)

wgtmac:
@pitrou May I request your review on this? Thanks!

cpp/src/parquet/page_index.h
emkornfield:
```suggestion
                        const IndexSelection& index_selection) = 0;
```
is what I was suggesting, even though it might be technically less efficient, it is likely less suprising.

emkornfield (commented):
Other then function signature this looks reasonable to me.  Will aim to merge once those are addressed (no earlier then Thursday to leave time for other comments).

wgtmac:
> Other then function signature this looks reasonable to me. Will aim to merge once those are addressed (no earlier then Thursday to leave time for other comments).

All comments have been addressed. Please take a look again. Thanks @emkornfield !

cpp/src/parquet/page_index.h
```diff
@@ -126,4 +132,123 @@ class PARQUET_EXPORT OffsetIndex {
...
+
+  /// \brief Read offset index of a column chunk.
+  ///
+  /// \param[in] i column ordinal of the column chunk.
+  /// \returns offset index of the column or nullptr if it does not exist.
+  /// \throws ParquetException if the index is out of bound.
+  virtual std::shared_ptr<OffsetIndex> GetOffsetIndex(int32_t i) = 0;
+};
+
+struct IndexSelection {
```
pitrou, line 155: Call this `PageIndexSelection`, otherwise this is a bit vague as it sits in the `parquet` namespace?

cpp/src/parquet/page_index.h
```diff
@@ -126,4 +132,123 @@ class PARQUET_EXPORT OffsetIndex {
...
+  /// 2) If WillNeed() has been called for a specific row group, follow-up calls to get
+  ///    page index are limited to columns and index type requested by WillNeed().
+  ///    So it MAY FAIL if columns that are not requested by WillNeed() are requested.
+  /// 3) Later calls to WillNeed() MAY OVERRIDE previous calls of same row groups.
+  /// For example,
+  /// 1) If WillNeed() is not called for row group 0, then follow-up calls to read
+  ///    column index and/or offset index of all columns of row group 0 should not
+  ///    fail if its page index exists.
+  /// 2) If WillNeed() is called for columns 0 and 1 for row group 0, then follow-up
+  ///    call to read page index of column 2 for row group 0 WILL FAIL even if its
```
pitrou, line 214: "MAY FAIL" rather?

cpp/src/parquet/page_index.h
```diff
@@ -126,4 +132,123 @@ class PARQUET_EXPORT OffsetIndex {
...
+  ///    page index exists.
+  /// 3) If WillNeed() is called for row group 0 with offset index only, then
+  ///    follow-up call to read column index of row group 0 WILL FAIL even if
```
pitrou, line 217: "MAY FAIL"?

cpp/src/parquet/page_index.h
```diff
@@ -126,4 +132,123 @@ class PARQUET_EXPORT OffsetIndex {
...
+  /// \brief Advise the reader page index of these row groups will not be read any more.
+  ///
+  /// The PageIndexReader implementation has the opportunity to cancel any prefetch or
+  /// release resource that are related to these row groups.
+  ///
+  /// \param[in] row_group_indices list of row group ordinal that whose page index will
+  /// not be accessed any more.
+  virtual void WillNotNeed(const std::vector<int32_t>& row_group_indices) = 0;
+
+  /// \brief Determines the column index and offset index ranges for the given row group.
```
pitrou, line 241:
```suggestion
  /// \brief Determine the column index and offset index ranges for the given row group.
```

cpp/src/parquet/page_index.h
```diff
@@ -126,4 +132,123 @@ class PARQUET_EXPORT OffsetIndex {
...
+  ///
+  /// \param[in] row_group_metadata row group metadata to get column chunk metadata.
+  /// \param[in] columns list of column ordinals to get page index. If the list is empty,
+  ///            it means all columns in the row group.
+  /// \returns RowGroupIndexReadRange of the specified row group. Throws ParquetException
+  ///          if the selected column ordinal is out of bound or metadata of page index
+  ///          is corrupted.
+  /// found. Returns false when there is absolutely no offsets index for the row group.
```
pitrou, line 249: This is weirdly formatted/indented.

cpp/src/parquet/page_index.cc
```diff
@@ -184,8 +185,294 @@ class OffsetIndexImpl : public OffsetIndex {
...
+
+  auto merge_range = [](const std::optional<IndexLocation>& index_location,
+                        int64_t* start, int64_t* end) {
+    if (index_location.has_value()) {
+      if (index_location->offset < 0 || index_location->length <= 0) {
+        throw ParquetException("Invalid index location: offset ", index_location->offset,
+                               " length ", index_location->length);
+      }
+      *start = std::min(*start, index_location->offset);
+      *end = std::max(*end, index_location->offset + index_location->length);
```
pitrou, line 445: Do we want to guard against addition overflow? We try to make the Parquet reader robust against invalid/corrupt data (some fuzzing is done, but it doesn't involve optional features as bloom filters, page index, encryption...).

cpp/src/parquet/reader_test.cc
```diff
@@ -1059,4 +1060,165 @@ TEST(TestFileReader, TestOverflowInt16PageOrdinal) {
   }
 }
 
+struct PageIndexReaderParam {
+  PageIndexReaderParam(const std::vector<int32_t>& row_group_indices,
+                       const std::vector<int32_t>& column_indices, bool need_column_index,
+                       bool need_offset_index)
+      : row_group_indices(row_group_indices),
+        column_indices(column_indices),
+        index_selection{need_column_index, need_offset_index} {}
```
pitrou, line 1069: I don't think you need to write this constructor explicitly, aggregate initialization will work automatically.

cpp/src/parquet/page_index.cc
pitrou: You can use `AddWithOverflow` for that.

cpp/src/parquet/reader_test.cc
```diff
@@ -1059,4 +1060,165 @@ TEST(TestFileReader, TestOverflowInt16PageOrdinal) {
...
+                                           PageIndexReaderParam({0}, {0}, false, false),
+                                           PageIndexReaderParam({0}, {5}, true, true),
+                                           PageIndexReaderParam({0}, {5}, true, false),
+                                           PageIndexReaderParam({0}, {5}, false, true),
+                                           PageIndexReaderParam({0}, {5}, false, false),
+                                           PageIndexReaderParam({0}, {0, 5}, true, true),
+                                           PageIndexReaderParam({0}, {0, 5}, true, false),
+                                           PageIndexReaderParam({0}, {0, 5}, false, true),
+                                           PageIndexReaderParam({0}, {0, 5}, false,
+                                                                false)));
```
pitrou, line 1209: Thanks a lot for the thorough testing @wgtmac !

wgtmac:
All comments are addressed. Thank you for the review @pitrou

cpp/src/parquet/page_index.cc
```diff
@@ -441,8 +442,11 @@ RowGroupIndexReadRange PageIndexReader::DeterminePageIndexRangesInRowGroup(
         throw ParquetException("Invalid index location: offset ", index_location->offset,
                                " length ", index_location->length);
       }
+      int64_t index_end = 0;
+      ::arrow::internal::AddWithOverflow(index_location->offset, index_location->length,
```
pitrou, line 446: Hmm, ok, but you should check the return value, which is the whole point.

pitrou: Filed https://github.com/apache/arrow/issues/34003 to make this mandatory.

pitrou (approved):
Nice work @wgtmac , thanks a lot!

cpp/src/parquet/page_index.cc
wgtmac: > Hmm, ok, but you should check the return value, which is the whole point.

My bad! Sorry about that

pitrou:
Ok, I think the fix worked. I also rebased on latest master.

PR #14974, opened 2022-12-15
ARROW-18436: [C++] Ensure correct (un)escaping of special characters in URI paths

pitrou:


pitrou:
@westonpace @vibhatha This may affect Substrait, though it should be for the better.

pitrou:
@jorisvandenbossche @AlenkaF This sanitizes the `filesystem_from_uri` behavior (see unit tests).

lidavidm (approved):

vibhatha:
> @westonpace @vibhatha This may affect Substrait, though it should be for the better.

looks good to me, @westonpace do you see any issues?

cpp/src/arrow/util/uri.cc
```diff
@@ -58,10 +59,10 @@ bool IsDriveSpec(const std::string_view s) {
 
 }  // namespace
 
-std::string UriEscape(const std::string& s) {
+std::string UriEscape(const std::string_view& s) {
```
westonpace, line 62: Minor nit: I think we've been preferring to pass `string_view` by value in Substrait per https://quuxplusone.github.io/blog/2021/11/09/pass-string-view-by-value/ and @bkietz 's advice

cpp/src/arrow/util/uri.cc
```diff
@@ -306,14 +311,29 @@ Status Uri::Parse(const std::string& uri_string) {
...
+  }
+  std::string out;
 #ifdef _WIN32
-  // Path is supposed to start with "X:/..."
-  return "file:///" + path;
+  // Turn "/" separators into "\", as Windows recognizes both but uriparser
+  // only the latter.
+  std::string fixed_path(path);
+  std::replace(fixed_path.begin(), fixed_path.end(), '/', '\\');
+  out.resize(8 + 3 * fixed_path.length() + 1);
```
westonpace, line 325: Minor nit: Maybe a comment explaining this math?  Probably not needed if this is a normal Windows thing.

westonpace (commented):
I agree it's for the better, thanks.  A few small thoughts.

cpp/src/arrow/util/uri.cc
pitrou: Ah, good point, thanks.

cpp/src/arrow/util/uri.cc
pitrou: This is simply as per the `uriWindowsFilenameToUriStringA` doc:
https://uriparser.github.io/doc/api/latest/Uri_8h.html#a422dc4a2b979ad380a4dfe007e3de845

cpp/src/arrow/util/uri.cc
pitrou: I opened https://github.com/apache/arrow/issues/15171 for other instances of this.

PR #14976, opened 2022-12-15
GH-14975: [Python] Dataset.sort_by

### LINKED ISSUE #14975: [Python] Add Dataset.sort_by

We currently support forting `Table` and `RecordBatch` using a `X.sort_by` method.
It would be helpful to have a similar helper method for `Dataset`

### Component(s)

Python

amol-:
* Closes: #14975

- [x] Proof of concept using an ExecPlan
- [x] Add test to filter and then sort to confirm lazy filtering works with sorting.

### DIFF
--- a/python/pyarrow/table.pxi
+++ b/python/pyarrow/table.pxi
@@ -4758,11 +4758,11 @@ cdef class Table(_PandasConvertible):
         if isinstance(sorting, str):
             sorting = [(sorting, "ascending")]
 
-        indices = _pc().sort_indices(
-            self,
-            options=_pc().SortOptions(sort_keys=sorting, **kwargs)
-        )
-        return self.take(indices)
+        res = _pc()._exec_plan._sort_source(self, output_type=Table,
+                                            sort_options=_pc().SortOptions(
+                                                sort_keys=sorting
+                                            ))
+        return res
 
     def join(self, right_table, keys, right_keys=None, join_type="left outer",
              left_suffix=None, right_suffix=None, coalesce_keys=True,
--- a/python/pyarrow/table.pxi
+++ b/python/pyarrow/table.pxi
@@ -4760,7 +4760,7 @@ cdef class Table(_PandasConvertible):
 
         res = _pc()._exec_plan._sort_source(self, output_type=Table,
                                             sort_options=_pc().SortOptions(
-                                                sort_keys=sorting
+                                                sort_keys=sorting, **kwargs
                                             ))
         return res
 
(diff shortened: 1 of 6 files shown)


python/pyarrow/table.pxi
```diff
@@ -4758,11 +4758,11 @@ cdef class Table(_PandasConvertible):
...
 
-        indices = _pc().sort_indices(
-            self,
-            options=_pc().SortOptions(sort_keys=sorting, **kwargs)
-        )
-        return self.take(indices)
+        res = _pc()._exec_plan._sort_source(self, output_type=Table,
+                                            sort_options=_pc().SortOptions(
+                                                sort_keys=sorting, **kwargs
+                                            ))
```
jorisvandenbossche, line 4764: Is there any advantage is using the exec plan version here, instead of the current version? (but I suppose also no disadvantage)

amol-: I didn't benchmark it, I can if you want to have numbers, but I guessed that it's faster to do the whole operation in C++ once than having to sort indices and create a temporary arrays of indices just to then take the actual values from the table.

jorisvandenbossche (approved):

kou:
@github-actions crossbow submit example-python-minimal-build-*

kou:
@amol- It seems that this breaks `example-python-minimal-build-*` nightly CI jobs:

https://github.com/ursacomputing/crossbow/actions/runs/3772771994/jobs/6413886549#step:3:7216

```text
______________________________ test_table_sort_by ______________________________

    def test_table_sort_by():
        table = pa.table([
            pa.array([3, 1, 4, 2, 5]),
            pa.array(["b", "a", "b", "a", "c"]),
        ], names=["values", "keys"])
    
>       assert table.sort_by("values").to_pydict() == {
            "keys": ["a", "a", "b", "b", "c"],
            "values": [1, 2, 3, 4, 5]
        }

pyarrow/tests/test_table.py:2198: 

>   res = _pc()._exec_plan._sort_source(self, output_type=Table,
E   AttributeError: module 'pyarrow.compute' has no attribute '_exec_plan'

pyarrow/table.pxi:4761: AttributeError
```

Could you fix this?

python/pyarrow/table.pxi
jorisvandenbossche: For future reference, at the time I replaced this with a manual sort_indices call in a follow-up PR (https://github.com/apache/arrow/pull/15268, for enabling this in a minimal build). But so it also appears that the Acero-based implementation is currently actually slower, see https://github.com/apache/arrow/pull/34249#issuecomment-1468028906 for some analysis (so as long as that is not solved, eg by allowing configuring the batch size, we should keep this manual implementation for Table/RecordBatch.sort_by).

PR #14952, opened 2022-12-14
MINOR: [Go][CI][Benchmarking] Fix GitHub info in Go benchmarks

austin3dickey:
This PR changes `go_bench_adapt.py` to be more accurate about reporting GitHub information to Conbench from a local computer.

In most cases, this script will be run on the master branch during [this workflow](https://github.com/apache/arrow/blob/794aefd6409a3da339ecc1a3d1ea6466b3d9ddb6/.github/workflows/go.yml#L94). This PR shouldn't change that use case. 

However, rarely, someone might [run benchmarks locally](https://conbench.ursa.dev/?search=branch) on a branch on their computer. If that's the case, we want the GitHub information to be scraped from the local repo, instead of assuming it's on the master branch. This is accomplished by passing `None` to `github`.

PR #14982, opened 2022-12-15
MINOR: [Go][CI][Benchmarking] Add timeout to go benchmarks

zeroshade:
As mentioned by @kou here: https://github.com/apache/arrow/pull/14952#issuecomment-1352401925 it looks like we crossed a threshold on the Go benchmarks and have added enough benchmarks for the Compute module that they are collectively taking more than 10m to run (which is the default timeout). So this updates the timeout used for running the benchmarks to 20 minutes.

CC @austin3dickey


kou:
@zeroshade It's still failing... https://github.com/apache/arrow/actions/runs/3707051286/jobs/6282999469#step:7:432

```text
*** Test killed with quit: ran too long (11m0s).
```

Could you test this on your fork before you merge?

PR #14984, opened 2022-12-15
GH-14973: [JS] Fixes bug where serialised data types are not JSON valid when using tableFromIPC().toString()

### LINKED ISSUE #14973: [JS] Using tableFromIPC().toString() results in invalid stringified JSON when containing dates
For reference, let's use the following code sample where we have a single row table:
```javascript
const tableAsArray = tableFromIPC(someSource).toArray();

console.log(tableAsArray);
// [{ id: 'foo', name: 'bar', dob: Fri Dec 15 2000 12:00:00 GMT+0100 (Central European Standard Time) }]
```
Considering the `dob` property above is a date object, this should be the JSON string:
```javascript
const tableAsString = tableFromIPC(someSource).toString();

console.log(tableAsString);
// '[{"id":"foo","name":"bar","dob":"2000-12-15T11:00:00.000Z"}]'
```
Unfortunately, that's not what happens. Instead this is how the string looks like:
```javascript
const tableAsString = tableFromIPC(someSource).toString();

console.log(tableAsString);
// '[{"id":"foo","name":"bar","dob":Fri Dec 15 2000 12:00:00 GMT+0100 (Central European Standard Time)}]'
```
Note that not only the date has not been converted to JSON (`dateObject.toJSON()`) but also, it is an unquoted string. This is of course a problem as one cannot parse back this string to JSON as the parser will throw an exception.

### Component(s)

JavaScript

arteconceito:
This PR fixes the bug reported by #14973 leveraging the native JavaScript `JSON.stringify()` function to serialize an Arrow Table into a valid JSON string. The use of this native function guarantees that all data types will be serialised into valid ones. This allows the reverse operation of deserialising the same string back into a JavaScript object through the use of `JSON.parse()`.
* Closes: #14973

### DIFF
--- a/js/src/table.ts
+++ b/js/src/table.ts
@@ -245,7 +245,7 @@ export class Table<T extends TypeMap = any> {
      * @returns A string representation of the Table rows.
      */
     public toString() {
-        return `[\n  ${this.toArray().join(',\n  ')}\n]`;
+        return JSON.stringify(this.toArray());
     }
 
     /**


arteconceito:
Hi @AlenkaF
This failing test above, is it worth re-running it? Or, do you know anyone who may know what it is about?
Thanks in advance.

trxcllnt (changes requested):
I believe the Vector and Row have a `toJSON()` method. Could we implement this as a new `toJSON()` method on Table (and ideally RecordBatch)?

arteconceito:
Yeah, sure @trxcllnt. Whenever I have some time I will look into that 👍

arteconceito:
@trxcllnt I had a look at the `.toJSON()` methods and I do not see any issue there as those methods return JavaScript objects, not strings. The issue being solved here is only when JavaScript objects are serialised to strings over the `.toString()` methods. I will go over those `.toString()` methods and fix where necessary.

Let me know in case I am missing something. Thx!

domoritz:
I want to note that the goal of `toString` is not to produce a de-serializable representation. `toJSON` should produce valid JSON.

arteconceito:
What you describe @domoritz  is in fact consistent to what happens for example with `new Date().toJSON()`. In such case you get a JSON serialised date string. But, I want to point out that the current implementation of the `.toJSON()` methods does not follow that principle, rather behaves as I described [here](https://github.com/apache/arrow/pull/14984#issuecomment-1363062607).

domoritz:
Do I follow correctly that the proposed fix in this pull request is not the final patch you want us to review? If so, can you mark it as a draft and ping us when it's ready?

arteconceito:
As discussed [here](https://github.com/apache/arrow/pull/14984#issuecomment-1363062607), I am closing this PR as it is no longer relevant.

PR #14989, opened 2022-12-16
GH-33600: [Go][Parquet] Panic in bitmap writer

### LINKED ISSUE #33600: [Go] firstTimeBitmapWriter.Finish() panics with list of 8 structs
Even after [ARROW-17169](https://issues.apache.org/jira/browse/ARROW-17169) I still get a panic at the same location.

Below is a test case that panics:
```go

func (ps *ParquetIOTestSuite) TestStructWithListOf8Structs() {
	bldr := array.NewStructBuilder(memory.DefaultAllocator, arrow.StructOf(
		arrow.Field{
			Name: "l",
			Type: arrow.ListOf(arrow.StructOf(
				arrow.Field{Name: "a", Type: arrow.BinaryTypes.String},
			)),
		},
	))
	defer bldr.Release()

	lBldr := bldr.FieldBuilder(0).(*array.ListBuilder)
	stBldr := lBldr.ValueBuilder().(*array.StructBuilder)
	aBldr := stBldr.FieldBuilder(0).(*array.StringBuilder)

	bldr.AppendNull()
	bldr.Append(true)
	lBldr.Append(true)
	for i := 0; i < 8; i++ {
		stBldr.Append(true)
		aBldr.Append(strconv.Itoa(i))
	}

	arr := bldr.NewArray()
	defer arr.Release()

	field := arrow.Field{Name: "x", Type: arr.DataType(), Nullable: true}
	expected := array.NewTable(arrow.NewSchema([]arrow.Field{field}, nil),
		[]arrow.Column{*arrow.NewColumn(field, arrow.NewChunked(field.Type, []arrow.Array{arr}))}, -1)
	defer expected.Release()

	ps.roundTripTable(expected, false)
}
```

I've tried to trim down the input data and this is as minimal as I could get it. And yes:
- wrapping struct with initial null is required
- the inner list needs to contain 8 structs (or any multiple of 8)

**Reporter**: [Min-Young Wu](https://issues.apache.org/jira/browse/ARROW-18438) / @minyoung
**Assignee**: [Min-Young Wu](https://issues.apache.org/jira/browse/ARROW-18438) / @minyoung
#### PRs and other links:
- [GitHub Pull Request #14989](https://github.com/apache/arrow/pull/14989)

<sub>**Note**: *This issue was originally created as [ARROW-18438](https://issues.apache.org/jira/browse/ARROW-18438). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

minyoung:
There is still a panic in go's bitmap writer, but specifically for when the list contains a multiple of 8 structs (the outer struct starting with null is also required).

Not sure where the off by one actually is, I somewhat blindly based this PR on https://github.com/apache/arrow/pull/14183
* Closes: #33600

### DIFF
--- a/go/parquet/file/column_writer_types.gen.go.tmpl
+++ b/go/parquet/file/column_writer_types.gen.go.tmpl
@@ -141,9 +141,9 @@ func (w *{{.Name}}ColumnChunkWriter) WriteBatchSpaced(values []{{.name}}, defLev
     }
 
     if w.bitsBuffer != nil {
-      w.writeValuesSpaced(vals, info.batchNum, w.bitsBuffer.Bytes(), 0)
+      w.writeValuesSpaced(vals, batch, w.bitsBuffer.Bytes(), 0)
     } else {
-      w.writeValuesSpaced(vals, info.batchNum, validBits, validBitsOffset+valueOffset)
+      w.writeValuesSpaced(vals, batch, validBits, validBitsOffset+valueOffset)
     }
     w.commitWriteAndCheckPageLimit(batch, info.numSpaced())
     valueOffset += info.numSpaced()
--- a/go/parquet/file/column_writer_types.gen.go.tmpl
+++ b/go/parquet/file/column_writer_types.gen.go.tmpl
@@ -141,9 +141,9 @@ func (w *{{.Name}}ColumnChunkWriter) WriteBatchSpaced(values []{{.name}}, defLev
     }
 
     if w.bitsBuffer != nil {
-      w.writeValuesSpaced(vals, batch, w.bitsBuffer.Bytes(), 0)
+      w.writeValuesSpaced(vals, info.batchNum, batch, w.bitsBuffer.Bytes(), 0)
     } else {
-      w.writeValuesSpaced(vals, batch, validBits, validBitsOffset+valueOffset)
+      w.writeValuesSpaced(vals, info.batchNum, batch, validBits, validBitsOffset+valueOffset)
     }
     w.commitWriteAndCheckPageLimit(batch, info.numSpaced())
     valueOffset += info.numSpaced()
@@ -160,8 +160,8 @@ func (w *{{.Name}}ColumnChunkWriter) writeValues(values []{{.name}}, numNulls in
   }
 }
 
-func (w *{{.Name}}ColumnChunkWriter) writeValuesSpaced(spacedValues []{{.name}}, numValues int64, validBits []byte, validBitsOffset int64) {
-  if len(spacedValues) != int(numValues) {
+func (w *{{.Name}}ColumnChunkWriter) writeValuesSpaced(spacedValues []{{.name}}, numRead, numValues int64, validBits []byte, validBitsOffset int64) {
+  if len(spacedValues) != int(numRead) {
     w.currentEncoder.(encoding.{{.Name}}Encoder).PutSpaced(spacedValues, validBits, validBitsOffset)
   } else {
     w.currentEncoder.(encoding.{{.Name}}Encoder).Put(spacedValues)
--- a/go/parquet/file/column_writer_types.gen.go
+++ b/go/parquet/file/column_writer_types.gen.go
@@ -136,9 +136,9 @@ func (w *Int32ColumnChunkWriter) WriteBatchSpaced(values []int32, defLevels, rep
 		}
 
 		if w.bitsBuffer != nil {
-			w.writeValuesSpaced(vals, batch, w.bitsBuffer.Bytes(), 0)
+			w.writeValuesSpaced(vals, info.batchNum, batch, w.bitsBuffer.Bytes(), 0)
 		} else {
-			w.writeValuesSpaced(vals, batch, validBits, validBitsOffset+valueOffset)
+			w.writeValuesSpaced(vals, info.batchNum, batch, validBits, validBitsOffset+valueOffset)
 		}
 		w.commitWriteAndCheckPageLimit(batch, info.numSpaced())
 		valueOffset += info.numSpaced()
@@ -154,8 +154,8 @@ func (w *Int32ColumnChunkWriter) writeValues(values []int32, numNulls int64) {
 	}
 }
 
-func (w *Int32ColumnChunkWriter) writeValuesSpaced(spacedValues []int32, numValues int64, validBits []byte, validBitsOffset int64) {
-	if len(spacedValues) != int(numValues) {
+func (w *Int32ColumnChunkWriter) writeValuesSpaced(spacedValues []int32, numRead, numValues int64, validBits []byte, validBitsOffset int64) {
+	if len(spacedValues) != int(numRead) {
 		w.currentEncoder.(encoding.Int32Encoder).PutSpaced(spacedValues, validBits, validBitsOffset)
 	} else {
 		w.currentEncoder.(encoding.Int32Encoder).Put(spacedValues)
@@ -294,9 +294,9 @@ func (w *Int64ColumnChunkWriter) WriteBatchSpaced(values []int64, defLevels, rep
 		}
 
 		if w.bitsBuffer != nil {
-			w.writeValuesSpaced(vals, batch, w.bitsBuffer.Bytes(), 0)
+			w.writeValuesSpaced(vals, info.batchNum, batch, w.bitsBuffer.Bytes(), 0)
 		} else {
-			w.writeValuesSpaced(vals, batch, validBits, validBitsOffset+valueOffset)
+			w.writeValuesSpaced(vals, info.batchNum, batch, validBits, validBitsOffset+valueOffset)
 		}
 		w.commitWriteAndCheckPageLimit(batch, info.numSpaced())
 		valueOffset += info.numSpaced()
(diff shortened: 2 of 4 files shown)


zeroshade:
@minyoung I'm so sorry, this fell off my radar with the holidays and new year. 

Personally, I'd rather understand where the off by one error actually is and *why* it needs that +1 there instead of just blindly adding it. When I get a chance to dig into this, I'll try but if you get some time beforehand and are able to dig into this some more and figure out why that `+1` is needed, please add a comment into the code and let me know and i'll be happy to merge this.

minyoung:
@zeroshade the `+1` is almost certainly a hack. I tried digging into this, but am not sure what the right solution is.

Disclaimer, my investigation may or may not correct since I'm not super familiar with the codebase and am making leaps of assumptions at various points.

What I've found so far is:

At the very lowest level, the panic is caused by `firstTimeBitmapWriter` being given less data than expected
```golang
buf := []byte{0} // only enough buffer for the data provided to AppendWord
bw := utils.NewFirstTimeBitmapWriter(buf, 0, 9) // we say that there will be 9 values
bw.AppendWord(0b11111111, 8) // we give 8 values
bw.Finish() // panic writing the 9'th bit out
```

And that discrepancy is coming from [defLevelsToBitmapInternal](https://github.com/apache/arrow/blob/go/v10.0.1/go/parquet/file/level_conversion.go#L172-L173):
* e.g. `defLevelsToBitmapInternal(defLevels=[0, 1, 2, 3, 4, 5, 6, 7, 8], out=&{ValidBits: [0]}, hasRepeatedParent=true)`
* there is only one byte available in `out.ValidBits`
* there are >8 definition levels given, with some values being 0 (meaning that the root value is null)
  * the length of this slice is used to initialize `NewFirstTimeBitmapWriter`
  * since `hasRepeatedParent = true`, only a [subset](https://github.com/apache/arrow/blob/go/v10.0.1/go/parquet/file/level_conversion.go#L151-L163) of the definition levels are given to the bitmap writer
  * this is where `AppendWord` is called with fewer bits than indicated in `NewFirstTimeBitmapWriter`

So, a thought I had would be to make `firstTimeBitmapWriter` tolerate getting less data which means having `firstTimeBitmapWriter.Finish` not write out `bw.curByte` (this is what's ultimately causing the panic) if there's no value to write out. But I'm unclear how to do that since testing for `bw.curByte == 0` is almost certainly wrong (0 is a valid value to want to write out).

The other way then is to increase the available buffer (`out.ValidBits`), this is where the `+1` comes into play. Well, why are we giving 8 bits when there are 9 definition levels? This has something to do with the nested structure of the data. To simplify things a little, we're only dealing with a single deeply nested leaf column, so there are multiple columns readers all wrapping the same definition levels used at the leaf.

At the top of the nesting, enough bytes are given to `file.DefRepLevelsToListInfo`/`file.DefLevelsToBitmap`, where it does not error and returns the [number of values read](https://github.com/apache/arrow/blob/go/v10.0.1/go/parquet/file/level_conversion.go#L176). This [feeds to the next layer](https://github.com/apache/arrow/blob/go/v10.0.1/go/parquet/pqarrow/column_readers.go#L289), where we [allocate enough bytes for the indicated number of values](https://github.com/apache/arrow/blob/go/v10.0.1/go/parquet/pqarrow/column_readers.go#L272). So, the parent reads 8 values (remember `hasRepeatedParent=true`, so only a subset is used), the child allocates space for that many values (1 byte), and we then ultimately end up at `defLevelsToBitmapInternal(defLevels=[0, 1, 2, 3, 4, 5, 6, 7, 8], out=&{ValidBits: [0]}, hasRepeatedParent=true)`.

Maybe the repetition/definition levels should be pruned at each level so that each child only sees a subset of those values instead of the full set of values?
Maybe the parent should tell the child to allocate as much space as it was told to allocate?

The `+1` hack works because `firstTimeBitmapWriter` will at most want to write one last byte in `Finish` (and this is why we need 8 values to trigger the panic).

zeroshade:
@minyoung So I spent a while digging into this and I think the issue is that your data type in the test says the field is not nullable, but you are adding a null..

Currently your test defines the builder like so:

```go
bldr := array.NewStructBuilder(memory.DefaultAllocator, arrow.StructOf(
		arrow.Field{
			Name: "l",
			Type: arrow.ListOf(arrow.StructOf(
				arrow.Field{Name: "a", Type: arrow.BinaryTypes.String},
			)),
		},
	))
```

If I add `Nullable: true` so it looks like this:

```go
bldr := array.NewStructBuilder(memory.DefaultAllocator, arrow.StructOf(
		arrow.Field{
			Name: "l",
			Type: arrow.ListOf(arrow.StructOf(
				arrow.Field{Name: "a", Type: arrow.BinaryTypes.String},
			)),
                       Nullable: true,
		},
	))
```

Then the test no longer fails *even without the +1 hack*. The issue is that the code makes some assumptions based on the data types since it generates the corresponding Parquet schema which, in the case of your original version, says it's a required field (ie: no nulls allowed). But you're adding a null to write. So things get wonky because of the mismatch of putting a null into a field marked required and non-null. Can you please try testing and confirm that? I think this is the source of your problem, and not an actual bug in the code.

minyoung:
@zeroshade you are correct about the first test case (my bad about that). What about the second test case where everything is marked as nullable though?

zeroshade:
@minyoung I found the issue and the solution that's not a hack:

in parquet/file/column_writer_types.gen.go.tmpl lines 143 - 147 change this:

```go
        if w.bitsBuffer != nil {
            w.writeValuesSpaced(vals, info.batchNum, w.bitsBuffer.Bytes(), 0)
        } else {
            w.writeValuesSpaced(vals, info.batchNum, validBits, validBitsOffset+valueOffset)
       }
```
To this instead:
```go
        if w.bitsBuffer != nil {
            w.writeValuesSpaced(vals, batch, w.bitsBuffer.Bytes(), 0)
        } else {
            w.writeValuesSpaced(vals, batch, validBits, validBitsOffset+valueOffset)
       }
```

Note the change from `info.batchNum` to `batch`. It turns out that in this scenario we should be passing the full batch size to write and not just the number of raw values it found. This way it properly calculates the number of nulls in the parent (non-leaf) columns. After you make this change, re-run `go generate` so that it re-generates all of the writers.

Finally, on line 289 of `pqarrow/column_readers.go`, change the argument to BuildArray from `validityIO.Read` to `lenBound`. That way it creates the correctly sized null bitmap buffer.

in my testing that was enough to solve the problem, leave your new tests in for good measure so this bug doesn't creep back :)

minyoung:
@zeroshade thanks for the suggestions! I tried applying your suggestions, but other things starts failing (see the build results). Am I interpreting your suggestion incorrectly?

go/parquet/file/column_writer_types.gen.go.tmpl
```diff
@@ -141,9 +141,9 @@ func (w *{{.Name}}ColumnChunkWriter) WriteBatchSpaced(values []{{.name}}, defLev
     }
 
     if w.bitsBuffer != nil {
-      w.writeValuesSpaced(vals, info.batchNum, w.bitsBuffer.Bytes(), 0)
+      w.writeValuesSpaced(vals, batch, w.bitsBuffer.Bytes(), 0)
     } else {
-      w.writeValuesSpaced(vals, info.batchNum, validBits, validBitsOffset+valueOffset)
+      w.writeValuesSpaced(vals, batch, validBits, validBitsOffset+valueOffset)
```
zeroshade, line 146: Looks like there's a reason I did it this way in the first place, haha!

So, there's probably a more elegant solution, but for a quick fix i was able to solve the failed tests by passing *both* `batch` and `info.batchNum` and then modifying `writeValuesSpaced` to take in both, with `batch` being the current `numValues` and `info.batchNum` (probably should rename this, but not necessary for this PR) being passed as `numRead`.  The new `writeValuesSpaced` looked like so:

```go
func (w *{{.Name}}ColumnChunkWriter) writeValuesSpaced(spacedValues []{{.name}}, numRead, numValues int64, validBits []byte, validBitsOffset int64) {
if len(spacedValues) != int(numRead) {
    w.currentEncoder.(encoding.{{.Name}}Encoder).PutSpaced(spacedValues, validBits, validBitsOffset)
  } else {
    w.currentEncoder.(encoding.{{.Name}}Encoder).Put(spacedValues)
  }
  if w.pageStatistics != nil {
    nulls := int64(len(spacedValues)) - numValues
    w.pageStatistics.(*metadata.{{.Name}}Statistics).UpdateSpaced(spacedValues, validBits, validBitsOffset, nulls)
  }
}
```

The issue was that since `len(spacedValues) == int(numRead)` it wasn't calling `PutSpaced` when it should be (which will remove the null values according to the bitmap and only write the non-null values as per the parquet spec).

*sigh* There's gotta be a more elegant way to do this, but at least this works.

zeroshade:
@minyoung sorry for the confusion, I thought I had run all the tests after making that change but apparently I only ran the tests you added (it's been a long week *sigh*).

Anyway, see my recent review comment for the solution to the failures you're seeing.

minyoung:
@zeroshade Thanks for having a look! I applied your most recent comment and all the tests seems to be passing now 🎉  I would not have known to also look at the column writers.

Since we're updating `column_writer_types.gen.go`, does that mean that previously generated parquet files are incorrect? If so, is there an easy way to detect/fix this?

zeroshade:
It would only be an issue for cases that hit the original bug we were trying to fix in the first place (ie: the nested columns with nulls that caused the original panic). I can't think of an easy way to detect and fix this though unfortunately.

minyoung:
> It would only be an issue for cases that hit the original bug we were trying to fix in the first place (ie: the nested columns with nulls that caused the original panic). I can't think of an easy way to detect and fix this though unfortunately.

Thanks for confirming, I assumed as much since pyarrow is able the open the parquet file without any errors.

PR #15010, opened 2022-12-17
GH-15009: [R] stringr 1.5.0 with the str_like function is already released

### LINKED ISSUE #15009: [R] stringr 1.5.0 with the str_like function is already released

stringr 1.5.0 was released on 2022-12-02.
So the following special handling of `str_like` is no longer necessary.

https://github.com/apache/arrow/blob/7e5730f4f64c9e6229840b439832e90cc84489ee/r/R/dplyr-funcs-doc.R#L318

### Component(s)

R

eitsupi:
* Closes: #15009

### DIFF
--- a/r/data-raw/docgen.R
+++ b/r/data-raw/docgen.R
@@ -85,9 +85,7 @@ library(dplyr)
 library(purrr)
 
 # Functions that for whatever reason cause xref problems, so don't hyperlink
-do_not_link <- c(
-  "stringr::str_like" # Still only in the unreleased version
-)
+do_not_link <- c()
 
 package_notes <- list(
   stringr = "Pattern modifiers `coll()` and `boundary()` are not supported in any functions."
--- a/r/R/dplyr-funcs-doc.R
+++ b/r/R/dplyr-funcs-doc.R
@@ -315,7 +315,7 @@
 #' * [`str_dup()`][stringr::str_dup()]
 #' * [`str_ends()`][stringr::str_ends()]
 #' * [`str_length()`][stringr::str_length()]
-#' * `str_like()`: not yet in a released version of `stringr`, but it is supported in `arrow`
+#' * [`str_like()`][stringr::str_like()]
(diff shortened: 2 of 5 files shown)


r/data-raw/docgen.R
```diff
@@ -85,9 +85,7 @@ library(dplyr)
 library(purrr)
 
 # Functions that for whatever reason cause xref problems, so don't hyperlink
-do_not_link <- c(
-  "stringr::str_like" # Still only in the unreleased version
-)
+do_not_link <- c()
```
thisisnic, line 88: Given this us now an empty vector, how about we remove this and update the code that uses it?

thisisnic (commented):
One small change, otherwise looks good; thanks for making this PR!

r/data-raw/docgen.R
eitsupi: That would be fine, but considering the possibility of similar cases occurring in the future, I thought it was worth leaving it for then.

thisisnic: That's logical. The reason I suggested removing it is that these cases occur very infrequently if at all; this is the only example out of >200 functions and happened because we implemented a function from the dev version of stringr, which is highly unusual and there are no plans to do this again.

That said it's so minor, I'm not going to block merging because of it; let me know if you'd rather leave it in and I'll approve this PR.

eitsupi: Thank you for the explanation. I don't know how `str_like` was implemented here, so your observation that the same is unlikely to happen in the future makes sense to me.

However, updating the script that generates the documentation seems generally more compelling to track in a separate PR, given that PR commits are squashed, so I would still prefer to keep it in this PR. (I think it would be far easier to revert from the future if this process were removed in another PR)

thisisnic (approved):
Thanks for making this update!

r/data-raw/docgen.R
thisisnic: That's reasonable, makes sense.

PR #13661, opened 2022-07-20
ARROW-17436: [C++] Use -O2 instead of -O3 for RELEASE builds

wesm:
Motivated by investigation in #13654. To be discussed

PR #15022, opened 2022-12-19
ARROW-18231: [C++][CMake] Add support for overriding optimization level

kou:
We can overriding optimization level by
`-DCMAKE_CXX_FLAGS_${CONFIG}=-OX` such as
`-DCMAKE_CXX_FLAGS_RELEASE=-O0` with this change.

Note that we can't use CXXFLAGS environment variable because CMake uses
`${CMAKE_CXX_FLAGS} ${CMAKE_CXX_FLAGS_${CONFIG}}` order and optimization flags are specified
in `CMAKE_CXX_FLAGS_${CONFIG}` not `CMAKE_CXX_FLAGS`. (CMAKE_CXX_FLAGS CMake variable is
initialized by CXXFLAGS environment variable.)

See also: https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS.html

> The flags in this variable will be passed to the compiler before
> those in the per-configuration `CMAKE_<LANG>_FLAGS_<CONFIG>` variant,
> and before flags added by the `add_compile_options()` or
> `target_compile_options()` commands.

This change also cleans up the followings:

* C/CXX flags related codes
  * Use `CMAKE_C_FLAGS_${BUILD_TYPE}`/`CMAKE_CXX_FLAGS_${BUILD_TYPE}` instead of putting all flags to `CMAKE_C_FLAGS`/`CMAKE_CXX_FLAGS`
* `externalproject_add()` related codes
  * Extract common options as `EP_XXX`
* Use `string(APPEND)`/`list(APPEND)` as much as possible instead of `set(XXX "${XXX} ...")`/`set(XXX ${XXX} ...)`


kou:
@github-actions crossbow submit -g nightly-tests -g nightly-packaging -g nightly-release

kou:
@github-actions crossbow submit -g nightly-tests -g nightly-packaging -g nightly-release

kou:
@github-actions crossbow submit verify-rc-source-python-linux-ubuntu-18.04-amd64

kou:
@github-actions crossbow submit -g nightly-tests -g nightly-packaging -g nightly-release

kou:
@github-actions crossbow submit test-r-install-local-minsizerel

kou:
@github-actions crossbow submit test-r-depsource-system test-r-dev-duckdb

kou:
@github-actions crossbow submit -g nightly-tests -g nightly-packaging -g nightly-release

kou:
This is ready to review/merge.

cpp/cmake_modules/SetupCxxFlags.cmake
```diff
@@ -604,76 +604,42 @@ endif()
...
 
 if(NOT MSVC)
-  string(REPLACE "-O3" "" CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE}")
-  string(REPLACE "-O3" "" CMAKE_CXX_FLAGS_RELWITHDEBINFO
-                 "${CMAKE_CXX_FLAGS_RELWITHDEBINFO}")
-
-  set(RELEASE_FLAGS "-O2 -DNDEBUG")
+  set(C_RELEASE_FLAGS "")
+  if(CMAKE_C_FLAGS_RELEASE MATCHES "-O3")
+    string(APPEND C_RELEASE_FLAGS " -O2")
```
mapleFU, line 616: Hi, kou, does this patch means, when we use release, the default build is "-O3 -DNDEBUG -O2", would it use O2 to final build it rather than O3? @kou

mapleFU: Personally I think it should be `if not`?

kou: Yes.
We don't have any benefit with `-O3`.
See also: #32703 and #13661

> -O3 can produce longer build times and larger binaries for next to no performance benefits.

mapleFU: I think you're right, but here I think that, why no 

```
if (CMAKE_CXX_FLAGS_RELEASE NOT MATCHES "-O3")
  string(APPEND CXX_RELEASE_FLAGS " -O2")
endif()
```

If I want to test O3, maybe we should have ability to use it?

kou: If this uses the condition, `-O3 -DNDEBUG` isn't change. We want to use `-O3 -DNDEBUG -O2` here.

We need to comment out the code when we try `-O3`. Sorry. We may want to add an option to force using `-O3`.

mapleFU: Thanks, the code and comment here is clear. I just want to make clear why we we O2, because when compiling I found the argument is "-O3 -NDEBUG -O2", which make me confused.

I'll try some benchmark under x86, if I got benefits when enabling O3, I'd like to add a flag here.

PR #15029, opened 2022-12-19
GH-15028: [R][Docs] `NOT_CRAN` should be `"true"` instead of `TRUE` in R

### LINKED ISSUE #15028: [R][Docs] `NOT_CRAN` should be `"true"` instead of `TRUE` in R

Currently on the installation manual it says the following,

https://github.com/apache/arrow/blob/dac3a00f87f0a63519ef4d96a3968e56fcfa3248/r/vignettes/install.Rmd#L120

but I think it needs to be written as `Sys.setenv("NOT_CRAN" = "true")` because the `NOT_CRAN` environment variable set by the `devtools` package is `true`, not `TURE`.

https://github.com/r-lib/devtools/blob/aa3f88bc693d049e1b524ca1a86b300e2c954765/R/R.R#L27
https://github.com/r-lib/devtools/commit/4a444805dadf02ccbba7d0776753ebc661fb9bcc

### Component(s)

R

eitsupi:
* Closes: #15028


assignUser (approved):
+1 we check/use the lower case version everywhere too.

PR #15031, opened 2022-12-19
ARROW-18351: [C++][FlightRPC] Fix crash in DoExchange with UCX

lidavidm:


lidavidm:
I looked at adding a unit test - however it would mean refactoring the internals quite a bit to expose the right bits to the unit test. I think that's worth considering separately:

- A core class that just implements the state machine for handling calls
- A wrapper class that connects the state machine to UCX

Then we can test the core state machine more easily.

cyb70289 (approved):

PR #15037, opened 2022-12-19
ARROW-17538: [C++] Import schema when importing array stream

lidavidm:
This lets us check for and report errors instead of crashing.

### DIFF
--- a/cpp/src/arrow/c/bridge.cc
+++ b/cpp/src/arrow/c/bridge.cc
@@ -1786,12 +1786,12 @@ class ArrayStreamBatchReader : public RecordBatchReader {
     }
     std::shared_ptr<Schema> schema;
     struct ArrowSchema c_schema = {};
-    Status status = StatusFromCError(stream, stream->get_schema(stream, &c_schema));
+    auto status = StatusFromCError(stream, stream->get_schema(stream, &c_schema));
     if (status.ok()) {
       status = ImportSchema(&c_schema).Value(&schema);
     }
     if (!status.ok()) {
-      stream->release(stream);
+      ArrowArrayStreamRelease(stream);
       return status;
     }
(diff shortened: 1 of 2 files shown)


cpp/src/arrow/c/bridge.cc
```diff
@@ -1777,17 +1779,30 @@ class ArrayStreamBatchReader : public RecordBatchReader {
...
-      ARROW_CHECK_OK(StatusFromCError(stream_.get_schema(&stream_, &c_schema)));
-      schema_ = ImportSchema(&c_schema).ValueOrDie();
+  static Result<std::shared_ptr<RecordBatchReader>> Make(
+      struct ArrowArrayStream* stream) {
+    if (ArrowArrayStreamIsReleased(stream)) {
+      return Status::Invalid("Cannot import released ArrowArrayStream");
+    }
+    std::shared_ptr<Schema> schema;
+    struct ArrowSchema c_schema = {};
+    Status status = StatusFromCError(stream, stream->get_schema(stream, &c_schema));
```
kou, line 1789:
```suggestion
    auto status = StatusFromCError(stream, stream->get_schema(stream, &c_schema));
```

cpp/src/arrow/c/bridge.cc
```diff
@@ -1777,17 +1779,30 @@ class ArrayStreamBatchReader : public RecordBatchReader {
...
+    if (status.ok()) {
+      status = ImportSchema(&c_schema).Value(&schema);
     }
-    return schema_;
+    if (!status.ok()) {
+      stream->release(stream);
```
kou, line 1794: Should we use `ArrowArrayStreamRelease()` here?

```suggestion
      ArrowArrayStreamRelease(stream);
```

kou (approved):
+1

wjones127 (approved):
This is a nice improvement :)

PR #15048, opened 2022-12-20
ARROW-18394: [Python][CI] Fix nightly job using pandas dev (temporarily skip tests)

jorisvandenbossche:


jorisvandenbossche:
@github-actions crossbow submit test-conda-python-3.8-pandas-nightly test-conda-python-3.9-pandas-upstream_devel

jorisvandenbossche:
@github-actions crossbow submit test-conda-python-3.8-pandas-nightly test-conda-python-3.9-pandas-upstream_devel

jorisvandenbossche:
@github-actions crossbow submit test-conda-python-3.8-pandas-nightly test-conda-python-3.9-pandas-upstream_devel

python/pyarrow/tests/test_compute.py
```diff
@@ -1994,12 +2000,12 @@ def test_assume_timezone():
                                  "2009-12-31T04:20:20.004132",
                                  "2010-01-01T05:25:25.005321",
                                  "2010-01-03T06:30:30.006163",
-                                 "2010-01-04T07:35:35",
-                                 "2006-01-01T08:40:40",
-                                 "2005-12-31T09:45:45",
-                                 "2008-12-28",
-                                 "2008-12-29",
-                                 "2012-01-01 01:02:03"])
+                                 "2010-01-04T07:35:35.0",
```
raulcd, line 2003: just curious why are those being updated? are formats like `"2008-12-28"` not supported anymore?

ci/scripts/install_pandas.sh
```diff
@@ -36,7 +36,7 @@ else
 fi
 
 if [ "${pandas}" = "upstream_devel" ]; then
-  pip install git+https://github.com/pandas-dev/pandas.git --no-build-isolation
+  pip install git+https://github.com/pandas-dev/pandas.git
```
raulcd, line 39: I wonder why we did use `--no-build-isolation` in the first place. LGTM

raulcd (commented):
I've reviewed the skips and the opened issues on pandas and all that LGTM. I added a couple comments, one of them (the timestamps one) I am not sure I understand why is required.

python/pyarrow/tests/test_compute.py
jorisvandenbossche: That string itself, yes, but not the mixture of different formats in a single list (but I am planning to raise an issue about this, because it seems too pedantic, as all those strings are ISO strings)

ci/scripts/install_pandas.sh
jorisvandenbossche: I think just to be more efficient (all dependencies like numpy and cython were already installed in the env (to build pyarrow), so no need to install those in an isolation build environment). But now pandas started to depend on `versioneer` as build-time dependency (instead of vendoring it), so this is the cleanest solution (otherwise we would have to install versioneer first)

PR #15049, opened 2022-12-20
GH-14968: [Python] Fix segfault for dataset ORC write

### LINKED ISSUE #14968: [Python] `write_dataset(table, format='orc')`: segfault
I'm on CPython 3.10.8 on Linux on x86_64.

```
$ python
Python 3.10.8 (main, Nov 15 2022, 15:04:54) [GCC 12.2.1 20220819 (Red Hat 12.2.1-2)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pyarrow
>>> pyarrow.__version__
'10.0.1'
>>> import pyarrow.dataset
>>> t = pyarrow.table({"a": range(10)})
>>> pyarrow.dataset.write_dataset(t, format='orc', base_dir='/tmp')
Segmentation fault (core dumped)
```

If what I try to do shouldn't even work (e.g., if `write_dataset()` isn't supposed to be able to write ORC) then we probably want a nicer error message.

I tried searching existing GitHub issues for that, but didn't find anything applicable. Sorry if that's already tracked elsewhere.

### Component(s)

Python

jgehrcke:
This is my attempt to address https://github.com/apache/arrow/issues/14968 pragmatically.

This is my first PR for Arrow, so I'd appreciate a careful look and all the pointers :).
* Closes: #14968

### DIFF
--- a/cpp/src/arrow/dataset/file_base.h
+++ b/cpp/src/arrow/dataset/file_base.h
@@ -200,6 +200,9 @@ class ARROW_DS_EXPORT FileFormat : public std::enable_shared_from_this<FileForma
       fs::FileLocator destination_locator) const = 0;
 
   /// \brief Get default write options for this format.
+  ///
+  /// May return nullptr if this file format does not yet support writing
+  /// datasets.
   virtual std::shared_ptr<FileWriteOptions> DefaultWriteOptions() = 0;
 
  protected:
--- a/cpp/src/arrow/dataset/file_base.h
+++ b/cpp/src/arrow/dataset/file_base.h
@@ -201,8 +201,8 @@ class ARROW_DS_EXPORT FileFormat : public std::enable_shared_from_this<FileForma
 
   /// \brief Get default write options for this format.
   ///
-  /// May return nullptr if this file format does not yet support writing
-  /// datasets.
+  /// May return null shared_ptr if this file format does not yet support
+  /// writing datasets.
   virtual std::shared_ptr<FileWriteOptions> DefaultWriteOptions() = 0;
 
  protected:
--- a/python/pyarrow/_dataset.pyx
+++ b/python/pyarrow/_dataset.pyx
@@ -889,13 +889,12 @@ cdef class FileFormat(_Weakrefable):
     def make_write_options(self):
         sp_write_options = self.format.DefaultWriteOptions()
         if sp_write_options.get() == nullptr:
-           # DefaultWriteOptions() may return `nullptr` which means that
-           # the format does not yet support writing datasets.
-           raise NotImplementedError(
-               "Writing datasets not yet implemented for this file format."
+            # DefaultWriteOptions() may return `nullptr` which means that
+            # the format does not yet support writing datasets.
+            raise NotImplementedError(
+                "Writing datasets not yet implemented for this file format."
             )
         return FileWriteOptions.wrap(sp_write_options)
-
 
     @property
     def default_extname(self):
--- a/python/pyarrow/_dataset.pyx
+++ b/python/pyarrow/_dataset.pyx
@@ -775,6 +775,14 @@ cdef class FileWriteOptions(_Weakrefable):
 
     @staticmethod
     cdef wrap(const shared_ptr[CFileWriteOptions]& sp):
+
+        if sp.get() == nullptr:
+            # DefaultWriteOptions() may return `nullptr` which means that
+            # the format does not yet support writing datasets.
+            raise NotImplementedError(
+                "Writing datasets not yet implemented for this file format."
+            )
+
         type_name = frombytes(sp.get().type_name())
 
         classes = {
--- a/python/pyarrow/_dataset.pyx
+++ b/python/pyarrow/_dataset.pyx
@@ -775,14 +775,6 @@ cdef class FileWriteOptions(_Weakrefable):
 
     @staticmethod
     cdef wrap(const shared_ptr[CFileWriteOptions]& sp):
-
-        if sp.get() == nullptr:
-            # DefaultWriteOptions() may return `nullptr` which means that
-            # the format does not yet support writing datasets.
-            raise NotImplementedError(
-                "Writing datasets not yet implemented for this file format."
-            )
-
         type_name = frombytes(sp.get().type_name())
 
         classes = {
@@ -895,7 +887,15 @@ cdef class FileFormat(_Weakrefable):
         return Fragment.wrap(move(c_fragment))
 
     def make_write_options(self):
-        return FileWriteOptions.wrap(self.format.DefaultWriteOptions())
+        sp_write_options = self.format.DefaultWriteOptions()
+        if sp_write_options.get() == nullptr:
+           # DefaultWriteOptions() may return `nullptr` which means that
+           # the format does not yet support writing datasets.
+           raise NotImplementedError(
+               "Writing datasets not yet implemented for this file format."
+            )
+        return FileWriteOptions.wrap(sp_write_options)
+
 
     @property
     def default_extname(self):
--- a/python/pyarrow/tests/test_dataset.py
+++ b/python/pyarrow/tests/test_dataset.py
@@ -3081,19 +3081,19 @@ def test_orc_writer_not_implemented_for_dataset():
         pytest.skip('this case is covered by test_orc_format_not_supported')
 
     with pytest.raises(
-            NotImplementedError,
-            match="Writing datasets not yet implemented for this file format"
-        ):
-            ds.write_dataset(
-                pa.table({"a": range(10)}), format='orc', base_dir='/tmp'
-            )
+        NotImplementedError,
+        match="Writing datasets not yet implemented for this file format"
+    ):
+        ds.write_dataset(
+            pa.table({"a": range(10)}), format='orc', base_dir='/tmp'
+        )
 
     of = ds.OrcFileFormat()
     with pytest.raises(
-            NotImplementedError,
-            match="Writing datasets not yet implemented for this file format"
-        ):
-            of.make_write_options()
+        NotImplementedError,
+        match="Writing datasets not yet implemented for this file format"
+    ):
+        of.make_write_options()
 
 
 @pytest.mark.pandas
--- a/python/pyarrow/tests/test_dataset.py
+++ b/python/pyarrow/tests/test_dataset.py
@@ -3074,6 +3074,23 @@ def test_orc_format_not_supported():
             ds.dataset(".", format="orc")
 
 
+def test_orc_writer_not_implemented_for_dataset():
+    with pytest.raises(
+            NotImplementedError,
+            match="Writing datasets not yet implemented for this file format"
+        ):
+            ds.write_dataset(
+                pa.table({"a": range(10)}), format='orc', base_dir='/tmp'
+            )
+
+    of = ds.OrcFileFormat()
+    with pytest.raises(
+            NotImplementedError,
+            match="Writing datasets not yet implemented for this file format"
+        ):
+            of.make_write_options()
+
+
 @pytest.mark.pandas
 def test_csv_format(tempdir, dataset_reader):
     table = pa.table({'a': pa.array([1, 2, 3], type="int64"),
--- a/python/pyarrow/tests/test_dataset.py
+++ b/python/pyarrow/tests/test_dataset.py
@@ -3074,12 +3074,8 @@ def test_orc_format_not_supported():
             ds.dataset(".", format="orc")
 
 
+@pytest.mark.orc
 def test_orc_writer_not_implemented_for_dataset():
-    try:
-        from pyarrow.dataset import OrcFileFormat  # noqa
-    except ImportError:
-        pytest.skip('this case is covered by test_orc_format_not_supported')
-
     with pytest.raises(
         NotImplementedError,
         match="Writing datasets not yet implemented for this file format"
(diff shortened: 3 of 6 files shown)


jgehrcke:
The Windows build failed with
```
================================== FAILURES ===================================
_________________ test_orc_writer_not_implemented_for_dataset _________________
    def test_orc_writer_not_implemented_for_dataset():
        with pytest.raises(
                NotImplementedError,
                match="Writing datasets not yet implemented for this file format"
            ):
>               ds.write_dataset(
                    pa.table({"a": range(10)}), format='orc', base_dir='/tmp'
                )
pyarrow\tests\test_dataset.py:3082: 
pyarrow\dataset.py:946: in write_dataset
    format = _ensure_format(format)
obj = 'orc'
    def _ensure_format(obj):
        if isinstance(obj, FileFormat):
            return obj
        elif obj == "parquet":
            if not _parquet_available:
                raise ValueError(_parquet_msg)
            return ParquetFileFormat()
        elif obj in {"ipc", "arrow"}:
            return IpcFileFormat()
        elif obj == "feather":
            return FeatherFileFormat()
        elif obj == "csv":
            return CsvFileFormat()
        elif obj == "orc":
            if not _orc_available:
>               raise ValueError(_orc_msg)
E               ValueError: The pyarrow installation is not built with support for the ORC file format.
pyarrow\dataset.py:296: ValueError
```

Adding a commit.

jgehrcke:
Ran `archery lint --python --fix` locally and pushed the corresponding change.

jgehrcke:
CI is green :heavy_check_mark:  Would appreciate review @AlenkaF @westonpace @jorisvandenbossche -- appreciated!

python/pyarrow/tests/test_dataset.py
```diff
@@ -3074,6 +3074,28 @@ def test_orc_format_not_supported():
             ds.dataset(".", format="orc")
 
 
+def test_orc_writer_not_implemented_for_dataset():
+    try:
+        from pyarrow.dataset import OrcFileFormat  # noqa
+    except ImportError:
+        pytest.skip('this case is covered by test_orc_format_not_supported')
```
AlenkaF, line 3081: You could also skip the test with a pytest mark, like here: https://github.com/apache/arrow/blob/f67009aac21e8e615743bbc2933fdbf8376953f0/python/pyarrow/tests/test_dataset.py#L3004

AlenkaF (commented):
Thank you for debugging and contributing @jgehrcke !

This looks great. I have one small comment in the test skip plus a question: did you plan to update/correct the documentation also (as Weston suggested https://github.com/apache/arrow/issues/14968#issuecomment-1355542182)?

jgehrcke:
Thank you Alenka! :)

> did you plan to update/correct the documentation also (as Weston suggested https://github.com/apache/arrow/issues/14968#issuecomment-1355542182)?

I looked a bit for "the DefaultWriteOptions comment docs" and did not find those quickly, so I opted for adding the documenting comment to where `DefaultWriteOptions()` is called. 

Is this `\brief ...` what Weston was referring to?
https://github.com/apache/arrow/blob/d18412e07f5aad968a73e1f44bcfa0c63461211f/cpp/src/arrow/dataset/file_base.h#L203

> You could also skip the test with a pytest mark, like here:

Oh, thanks for the pointer! Does our CI invoke pytest with `-m "not orc"` to deselect those tests when testing a build without ORC support? Or would the user need to manually add `-m "not orc"`?

In case of the win32 CI run (which AFAIU did not use a pyarrow build with ORC support) the pytest cmd is `py.test -r sxX --durations=15 --pyargs pyarrow.tests`. And yet, all tests in `test_orc.py` are skipped (`pyarrow\tests\test_orc.py sssssssssssss   `. I had a quick look but didn't find the magic responsible for that. In other words: what's the magic that _automatically_ deselects orc-marked tests whenever necessary?

jgehrcke:
Btw, when looking at https://github.com/apache/arrow/blob/d18412e07f5aad968a73e1f44bcfa0c63461211f/ci/appveyor-cpp-build.bat I see that the c++ lib has `-DARROW_ORC=ON` but the pyarrow build does not have the corresponding `PYARROW_WITH_ORC=ON` -- is that intentional?

AlenkaF:
> I looked a bit for "the DefaultWriteOptions comment docs" and did not find those quickly, so I opted for adding the documenting comment to where `DefaultWriteOptions()` is called.
> 
> Is this `\brief ...` what Weston was referring to?

I would add a note in the:
https://arrow.apache.org/docs/dev/cpp/api/dataset.html#_CPPv4N5arrow7dataset13OrcFileFormat19DefaultWriteOptionsEv

and maybe make the orc support in the dataset more explicit in the C++ and Python documentation (as you have mentioned in the issue):
https://arrow.apache.org/docs/dev/cpp/dataset.html
https://arrow.apache.org/docs/python/dataset.html

> Oh, thanks for the pointer! Does our CI invoke pytest with `-m "not orc"` to deselect those tests when testing a build without ORC support? Or would the user need to manually add `-m "not orc"`?
> 
> In case of the win32 CI run (which AFAIU did not use a pyarrow build with ORC support) the pytest cmd [is](https://github.com/apache/arrow/blob/d18412e07f5aad968a73e1f44bcfa0c63461211f/ci/appveyor-cpp-build.bat#L130) `py.test -r sxX --durations=15 --pyargs pyarrow.tests`. And yet, all tests in `test_orc.py` are skipped (`pyarrow\tests\test_orc.py sssssssssssss `. I had a quick look but didn't find the magic responsible for that. In other words: what's the magic that _automatically_ deselects orc-marked tests whenever necessary?

You can get the info about it in these two conftest files 🪄 😉 
https://github.com/apache/arrow/blob/5e25af8860e3859bb4d17f22e4dbaa2309e5076d/python/pyarrow/tests/conftest.py#L48-L70
https://github.com/apache/arrow/blob/5e25af8860e3859bb4d17f22e4dbaa2309e5076d/python/pyarrow/conftest.py#L22-L82

> Btw, when looking at https://github.com/apache/arrow/blob/d18412e07f5aad968a73e1f44bcfa0c63461211f/ci/appveyor-cpp-build.bat I see that the c++ lib has -DARROW_ORC=ON but the pyarrow build does not have the corresponding PYARROW_WITH_ORC=ON -- is that intentional?

I am not sure about that. I would suspect it just wasn't added to the Win CI build yet as I haven't found any issue with CI failures due to `PYARROW_WITH_ORC=ON` in Jira 🤷‍♀️

jgehrcke:
Thanks for all those pointers Alenka! I have added commits for tweaking docs, and for using `pytest.mark.orc` instead of `pytest.skip`.

AlenkaF (approved):
This is great, thank you!

edponce (approved):

python/pyarrow/_dataset.pyx
```diff
@@ -775,6 +775,14 @@ cdef class FileWriteOptions(_Weakrefable):
 
     @staticmethod
     cdef wrap(const shared_ptr[CFileWriteOptions]& sp):
+
+        if sp.get() == nullptr:
+            # DefaultWriteOptions() may return `nullptr` which means that
+            # the format does not yet support writing datasets.
+            raise NotImplementedError(
+                "Writing datasets not yet implemented for this file format."
+            )
```
kou, line 784: How about doing this in `FileFormat.make_write_options()` not here?

```diff
diff --git a/python/pyarrow/_dataset.pyx b/python/pyarrow/_dataset.pyx
index 26c9f503bd..b2512806c3 100644
--- a/python/pyarrow/_dataset.pyx
+++ b/python/pyarrow/_dataset.pyx
@@ -887,7 +887,14 @@ cdef class FileFormat(_Weakrefable):
         return Fragment.wrap(move(c_fragment))
 
     def make_write_options(self):
-        return FileWriteOptions.wrap(self.format.DefaultWriteOptions())
+        sp_write_options = self.format.DefaultWriteOptions()
+        if sp_write_options.get() == nullptr:
+            # DefaultWriteOptions() may return `nullptr` which means that
+            # the format does not yet support writing datasets.
+            raise NotImplementedError(
+                "Writing datasets not yet implemented for this file format."
+            )
+        return FileWriteOptions.wrap(sp_write_options)
 
     @property
     def default_extname(self):
```

If we should check a return value of `WriteDefaultFormat()`, we should do it where `WriteDefaultFormats()` is used.

cpp/src/arrow/dataset/file_base.h
```diff
@@ -200,6 +200,9 @@ class ARROW_DS_EXPORT FileFormat : public std::enable_shared_from_this<FileForma
       fs::FileLocator destination_locator) const = 0;
 
   /// \brief Get default write options for this format.
+  ///
+  /// May return nullptr if this file format does not yet support writing
```
kou, line 204: How about using `empty shared_ptr` instead of `nullptr`?

This method always returns `std::shared_ptr<FileWriteOptions>` but this may return `std::shared_ptr<FileWriteOptions>` that stores `nullptr`.

It seems that a `std::shared_ptr` that stores `nullptr` is called `empty`:

https://en.cppreference.com/w/cpp/memory/shared_ptr

> A shared_ptr may also own no objects, in which case it is called empty

python/pyarrow/_dataset.pyx
jgehrcke: @kou I agree, now also seeing this very clearly :).

cpp/src/arrow/dataset/file_base.h
jgehrcke: Well. This is interesting! Thanks for raising this.

My C++ is veeery rusty. So, I needed to read a bit. I looked at 
https://www.nextptr.com/question/qa1372136808/shared_ptr-initialized-with-nullptr-is-null-or-empty
and
https://stackoverflow.com/questions/37234969/is-it-correct-to-return-null-shared-ptr

and decided to
```cpp
return std::shared_ptr<FileWriteOptions>(nullptr);
```
I think that's what you meant, right? I like that this is explicit. Pushing a commit.

Someone [pointed out that](https://stackoverflow.com/questions/37234969/is-it-correct-to-return-null-shared-ptr#comment103737709_37235251)

> I would not use return `shared_ptr<Object>(nullptr);` as `return shared_ptr<Object>();` is enough. It is a style preference: I don't write unnecessary stuff, as it might make think others that they are necessary.

I humbly disagree, and like the longer version more. But I usually don't write C++ so :shrug: 

The terminology discussion around "empty" is a bit of a comedy show to me :-) :clown_face:  Not sure if the term "empty" is useful in practice. From said cppreference page:

> an empty shared_ptr may have a non-null stored pointer

Someone on Stack Overflow _tries to_ clarify:

> The page you linked calls a shared_ptr that owns nothing empty, and a shared_ptr that points to nothing (i.e., whose get() == nullptr) null. (Empty is used in this sense by the standard; null isn't.)

Then someone says

> A null-but-nonempty shared_ptrs can be useful!

Someone else replies 

> What can a null-but-nonempty shared_ptr do that a non-null-and-nonempty shared_ptr can't?

Not sure what I am saying and if I want to know about this at all, but it is possible for a shared_ptr to be non-null but still be empty, as of the _aliasing constructor_ which  

> originates from Bloomberg and was proposed for the standard before it was implemented in Boost (see [N1851](http://www.open-std.org/jtc1/SC22/wg21/docs/papers/2005/n1851.pdf)). I prefer the standard's term "shares ownership with r" to the phrasing "owns whatever r owns"

SO quotes all from [here](https://stackoverflow.com/questions/25920681/what-is-the-difference-between-an-empty-and-a-null-stdshared-ptr-in-c).

Trying to fill this rabbit hole and put a lid on it :rabbit: 

Thanks for the detour :palm_tree:

kou: > and decided to
> 
> ```c++
> return std::shared_ptr<FileWriteOptions>(nullptr);
> ```
> 
> I think that's what you meant, right? I like that this is explicit. Pushing a commit.

Ah, sorry. I didn't meant that we should change the existing code. I just meant that we can improve the added comment: 

```suggestion
  /// May return empty shared_ptr if this file format does not yet support writing
```

Could you revert the `return std::shared_ptr<FileWriteOptions>(nullptr);` change in this pull request?

jgehrcke: Ha. OK. 

I will revert the code change.

About the comment, however: I am not convinced that "empty shared_ptr" is correct. "empty shared_ptr" means primarily that the pointer owns nothing. Whether or not this is a shared_ptr that points to nothing (i.e., whose get() == nullptr) is not defined by the word "empty".

It's seemingly not as simple as "It seems that a std::shared_ptr that stores nullptr is called empty".

This following paragraph makes it rather clear to me (emphasis mine):

> A shared_ptr that does not own any pointer is called an empty shared_ptr. **A shared_ptr that points to no object is called a null shared_ptr** and shall not be [dereferenced](https://cplusplus.com/shared_ptr::operator*). Notice though that **an empty shared_ptr is not necessarily a null shared_ptr, and a null shared_ptr is not necessarily an empty shared_ptr**.

(https://cplusplus.com/reference/memory/shared_ptr/)

Given these statements the term "null shared_ptr" seems to be more appropriate than "empty shared_ptr".

kou: Wow! I didn't know the "null shared_ptr" term.
You're right. We should use "null shared_ptr" here.

kou (approved):
+1

PR #15057, opened 2022-12-21
MINOR: [Doc][Python] In from_uri docs, show that gcsfs is also supported

Dhul-Husni:
Screenshot 2022-12-21 at 13 36 59

### DIFF
--- a/python/pyarrow/_fs.pyx
+++ b/python/pyarrow/_fs.pyx
@@ -425,7 +425,7 @@ cdef class FileSystem(_Weakrefable):
         """
         Create a new FileSystem from URI or Path.
 
-        Recognized URI schemes are "file", "mock", "s3fs", "hdfs" and "viewfs".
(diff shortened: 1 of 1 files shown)


python/pyarrow/_fs.pyx
```diff
@@ -425,7 +425,7 @@ cdef class FileSystem(_Weakrefable):
         """
         Create a new FileSystem from URI or Path.
 
-        Recognized URI schemes are "file", "mock", "s3fs", "hdfs" and "viewfs".
+        Recognized URI schemes are "file", "mock", "s3fs", "gcsfs", "hdfs" and "viewfs".
```
kou, line 428:
```suggestion
        Recognized URI schemes are "file", "mock", "s3fs", "gs", "gcs", "hdfs" and "viewfs".
```

kou (approved):
+1

Good catch!

PR #15061, opened 2022-12-21
GH-15043: [Python] [Docs] Update docstring for pyarrow.decompress

### LINKED ISSUE #15043: [Python] [Docs] Change docstring for `decompressed_size` arg in pyarrow.decompress to reflect implementation

The docstring for `decompressed_size` in `pyarrow.decompress` and the underlying `Codec.decompress` states:

> decompressed_size : int, default None
>     If not specified, will be computed if the codec is able to determine the uncompressed buffer size. 

However, you can see from the implementation that the behavior isn't possibly and that the argument is effectively required because we always raise when it's `None`:

https://github.com/apache/arrow/blob/4e9158d373df105f01ba9d6052cfb9ab7ecdcbeb/python/pyarrow/io.pxi#L2091-L2124

For example,

```python
import pyarrow as pa

pa.decompress(pa.compress(bytes('mydata','utf-8')))
```

Raises:

> ValueError: Must pass decompressed_size for <pyarrow.Codec name=lz4 compression_level=1> codec

I think the [original intent](https://github.com/apache/arrow/pull/1345) was to support detecting or estimating the size of the output buffer for codecs that support it (like Snappy) but, the last time [this was brought up](https://issues.apache.org/jira/browse/ARROW-3877), it was marked as low priority and remains unimplemented.

Should the argument be made required and the bit about "If not specified" be removed?

### Component(s)

Python

amoeba:
I opted to keep `decompressed_size` optional to reduce churn (optional->required->optional) if we ever add in detection/estimation for certain codecs in the future. Let me know what you think.
* Closes: #15043

### DIFF
--- a/python/pyarrow/io.pxi
+++ b/python/pyarrow/io.pxi
@@ -2088,7 +2088,7 @@ cdef class Codec(_Weakrefable):
             out_buf.resize(output_length)
             return out_buf
 
-    def decompress(self, object buf, decompressed_size=None, asbytes=False,
+    def decompress(self, object buf, decompressed_size, asbytes=False,
                    memory_pool=None):
         """
         Decompress data from buffer-like object.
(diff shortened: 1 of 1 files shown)


jorisvandenbossche (approved):
Thanks!

PR #15063, opened 2022-12-21
GH-15062: [C++] Simplify EnumParser behavior

### LINKED ISSUE #15062: Remove unused kUnspecified enum value and simplify EnumParser.

In engine/substrait/extension_set.cc EnumParser has special logic to ignore unspecified values which are typically not used.  Removing the unspecified enum when it's not used will allow the parser logic to be simplified.

### Component(s)

C++

EpsilonPrime:
Removes the unspecified enum from engine/substrait/extension_set.cc when it is not being used allowing the parser logic to be simplified.
* Closes: #15062

### DIFF
--- a/cpp/src/arrow/engine/substrait/extension_set.cc
+++ b/cpp/src/arrow/engine/substrait/extension_set.cc
@@ -706,8 +706,8 @@ class EnumParser {
 };
 
 enum class TemporalComponent { kUnspecified = 0, kYear, kMonth, kDay, kSecond };
-static std::vector<std::string> kTemporalComponentOptions = {"UNSPECIFIED", "YEAR", "MONTH", "DAY",
-                                                             "SECOND"};
+static std::vector<std::string> kTemporalComponentOptions = {"UNSPECIFIED", "YEAR",
+                                                             "MONTH", "DAY", "SECOND"};
 static EnumParser<TemporalComponent> kTemporalComponentParser(kTemporalComponentOptions);
 
 enum class OverflowBehavior { kSilent = 0, kSaturate, kError };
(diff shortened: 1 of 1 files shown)


cpp/src/arrow/engine/substrait/extension_set.cc
```diff
@@ -706,11 +706,11 @@ class EnumParser {
 };
 
 enum class TemporalComponent { kUnspecified = 0, kYear, kMonth, kDay, kSecond };
-static std::vector<std::string> kTemporalComponentOptions = {"YEAR", "MONTH", "DAY",
-                                                             "SECOND"};
+static std::vector<std::string> kTemporalComponentOptions = {"UNSPECIFIED", "YEAR",
+                                                             "MONTH", "DAY", "SECOND"};
```
westonpace, line 710: Why is there still a need for `UNSPECIFIED` here?

EpsilonPrime: Temporal extraction (in function tDecodeTemporalExtractionMapping) uses the kUnspecified state to determine if the enum argument was not passed to the function (to provide a more readable error message).  One alternative I considered was defaulting the value to kYear but that doesn't preserve the existing behavior.  I could remove kUnspecified if the less specific error is acceptable (I will have to investigate to see what the wording will be).

EpsilonPrime: Removing that parse extra parse check yields a rather useful "The value BOGUS is not an expected enum value" message when the value can't be parsed so I've removed it.

EpsilonPrime:
It looks like the one test failure is a spurious one.  This pull request should be fine to incorporate as is.

westonpace (approved):
Thanks for the cleanup!

PR #14769, opened 2022-11-29
MINOR: [Python] Bump max_line_length = 88

milesgranger:

PR #15067, opened 2022-12-22
MINOR: [Python] Fix pre-commit and archery flake8 to honor pyarrow config

jorisvandenbossche:
https://github.com/apache/arrow/pull/14769 changed the flake8 config for pyarrow to raise max line length to 88, but our linting checks didn't actually follow that.


dev/archery/archery/utils/lint.py
```diff
@@ -230,6 +230,7 @@ def python_linter(src, fix=False):
 
     yield LintResult.from_cmd(
         flake8("--extend-exclude=" + ','.join(flake8_exclude),
+               "--config=python/setup.cfg",
                setup_py, src.pyarrow, os.path.join(src.python, "examples"),
```
assignUser, line 234: For this to work in the docker lint job we need to specify the absolute path as archery is called from outside of the arrow. (see [here](https://github.com/apache/arrow/actions/runs/3757323229/jobs/6384396332#step:5:837))

This should do it:
```suggestion
        flake8("--extend-exclude=" + ','.join(flake8_exclude),
               "--config=" + os.path.join(src.python, "setup.cfg"),
               setup_py, src.pyarrow, os.path.join(src.python, "examples"),
```

assignUser, line 234: For this to work in the docker lint job we need to specify the absolute path as archery is called from outside of the arrow. (see [here](https://github.com/apache/arrow/actions/runs/3757323229/jobs/6384396332#step:5:837))

This should™ do it:
```suggestion
        flake8("--extend-exclude=" + ','.join(flake8_exclude),
               "--config=" + os.path.join(src.python, "setup.cfg"),
               setup_py, src.pyarrow, os.path.join(src.python, "examples"),
```

.pre-commit-config.yaml
```diff
@@ -46,3 +46,4 @@ repos:
           - file
           - python
         exclude: vendored
+        args: [--config, python/setup.cfg]
```
assignUser, line 49: I don't think we need the absolute path here as precommit should run within the arrow dir by default?

assignUser (changes requested):
A small fix is needed for the lint job.

.pre-commit-config.yaml
jorisvandenbossche: Yes, I assume so

raulcd (approved):

PR #15075, opened 2022-12-23
GH-15040: [C++] Improve pkg-config support for ARROW_BUILD_SHARED=OFF

### LINKED ISSUE #15040: [R][CI] Arrow nightly binaries require aws sdk system dependency

It seems that https://github.com/apache/arrow/pull/14235 introduced aws-sdk as a system dependency for the ubuntu libarrow binaries created by the R nightlies. @raulcd detected the issue while working on the cookbook: https://github.com/apache/arrow/pull/14235#issuecomment-1357539110 . 

We confirmed that it is loading the nighly binary and not building from source https://github.com/raulcd/arrow-cookbook/actions/runs/3733167177/jobs/6333699829#step:6:2572.  Changing the scripts and workflows to use bundled dependencies in any case should fix the issue.

### Component(s)

Continuous Integration, R

kou:
If `ARROW_BUILD_SHARED=OFF` is specified, `pkg-config --cflags --libs arrow` doesn't work. Because `pkg-config --cflags --libs arrow` misses build options for static linking. Users need to specify `pkg-config --cflags --libs --static arrow` for static linking even when `ARROW_BUILD_SHARED=OFF` is specified.

This change adds support for `pkg-config --cflags --libs arrow` with `ARROW_BUILD_SHARED=OFF`. The command outputs build options for linking against `libarrow.a`. But these build options don't use static linking for dependencies. On the other hand, `pkg-config --cflags --libs --static arrow` still use static linking for all libraries including Apache Arrow C++ and its dependencies.

With this change, Apache Arrow R doesn't need to use `pkg-config --static`. Because `pkg-config --cflags --libs arrow` works for both of `libarrow.so` and `libarrow.a`.
* Closes: #15040

### DIFF
--- a/ci/scripts/r_install_system_dependencies.sh
+++ b/ci/scripts/r_install_system_dependencies.sh
@@ -35,81 +35,9 @@ if [ "$ARROW_S3" == "ON" ] || [ "$ARROW_GCS" == "ON" ] || [ "$ARROW_R_DEV" == "T
   fi
 
   # Install curl and OpenSSL for S3/GCS support
-  #
-  # We need to install all dependencies explicitly to use "pkg-config
-  # --static --libs libcurl" result. Because libcurl-dev/libcurl-devel
-  # don't depend on packages that are only needed for "pkg-config
-  # --static".
   case "$PACKAGE_MANAGER" in
     apt-get)
-      # "pkg-config --static --libs libcurl" has
-      #   * -lnghttp2
-      #   * -lidn2
-      #   * -lrtmp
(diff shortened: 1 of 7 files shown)


kou:

kou:

kou:

kou:

kou:

kou:
This is ready to review/merge.

kou:
I merge this. If there are problems, I'll fix them in follow-up pull requests.

PR #15076, opened 2022-12-23
MINOR: [R] Should build ignore __pycache__ directory

eitsupi:
I noticed that the `inst/__pycache__` directory remains in the released R arrow package version 10.0.1.

```r
system.file("__pycache__/demo_flight_server.cpython-38.pyc", package = "arrow")
#> [1] "/usr/local/lib/R/site-library/arrow/__pycache__/demo_flight_server.cpython-38.pyc"
```

### DIFF
--- a/r/.Rbuildignore
+++ b/r/.Rbuildignore
@@ -31,4 +31,4 @@ STYLE.md
 ^revdep$
 ^vignettes$
 ^PACKAGING\.md$
-^__pycache__$
+^inst/__pycache__$
--- a/r/.Rbuildignore
+++ b/r/.Rbuildignore
@@ -31,3 +31,4 @@ STYLE.md
 ^revdep$
 ^vignettes$
 ^PACKAGING\.md$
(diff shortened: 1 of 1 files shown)


r/.Rbuildignore
```diff
@@ -31,3 +31,4 @@ STYLE.md
 ^revdep$
 ^vignettes$
 ^PACKAGING\.md$
+^__pycache__$
```
assignUser, line 34: The path has needs to be relative to the root dir.
```suggestion
^inst/__pycache__$
```

assignUser: (or alternatively we need to remove the line start/end markers `^ $` to ignore  pycache dirs regardless of location in the package. I prefer the more explicit version above though.

assignUser (changes requested):
Thanks for catching this. A minor change is needed.

assignUser (approved):
+1 I`ll merge after CI

PR #15077, opened 2022-12-23
GH-14907: [R] right_join() function does not produce the expected outcome

### LINKED ISSUE #14907: [R] right_join() function does not produce the expected outcome

Hi,

I noticed something strange today when I was using arrow datasets. I cannot give a reproducible example but you can get the idea from the code below. I have `ccaei` as an arrow dataset. When I try to use `right_join()` with an R tibble before using `collect()`, it gives me wrong numbers (the number of distinct `ENROLID` is less than that present in `outpatients`). I get the correct number when I use `right_join()` after using `collect()`, although this is computationally inefficient. Could you help me with this?

This gives a really weird number
```
ccaei |>  
  filter(ADMDATE >= as_date("2016-10-01")) |> 
  filter(!is.na(ENROLID)) |> 
  select(ENROLID, ADMDATE) |> 
  right_join(outpatients) |> 
  collect() |> count(ENROLID)
```

This makes sense
```
ccaei |>  
  filter(ADMDATE >= as_date("2016-10-01")) |> 
  filter(!is.na(ENROLID)) |> 
  select(ENROLID, ADMDATE) |> 
  collect() |> 
    right_join(outpatients) |> 
  count(ENROLID)
```

Not sure where the mistake came from.

### Component(s)

R

thisisnic:
* Closes: #14907


r/tests/testthat/test-dplyr-join.R
```diff
@@ -371,3 +371,24 @@ test_that("full joins handle keep", {
     )
   }
 })
+
+test_that("right_join correctly coalesces keys", {
```
wjones127, line 375: Instead of another test, could we modify the existing `left` and `to_join` to have keys that don't fully overlap?

https://github.com/apache/arrow/blob/785ab5f3c3aa4e5f2e30655d4765ca729bee7aa2/r/tests/testthat/test-dplyr-join.R#L20-L27

Then we get coverage from:

https://github.com/apache/arrow/blob/785ab5f3c3aa4e5f2e30655d4765ca729bee7aa2/r/tests/testthat/test-dplyr-join.R#L143-L152

r/R/dplyr-join.R
```diff
@@ -72,7 +72,18 @@ right_join.arrow_dplyr_query <- function(x,
...
+
+  # Initially keep join keys so we can coalesce them after when keep=FALSE
+  query <- do_join(x, y, by, copy, suffix, ..., keep = TRUE, join_type = "RIGHT_OUTER")
+
+  # If we are doing a right outer join and not keeping the join keys of
+  # both sides, we need to coalesce. Otherwise, rows that exist in the
+  # RHS will have NAs for the join keys.
+  if (!keep) {
+    query$selected_columns <- post_join_projection(names(x), names(y), handle_join_by(by, x, y), suffix)
+  }
```
wjones127, line 84: This works, but it might be straightforward to fix here:

https://github.com/apache/arrow/blob/785ab5f3c3aa4e5f2e30655d4765ca729bee7aa2/r/R/dplyr-join.R#L37-L43

in the case of right join, we need to do the `setdiff` to produce the `left_output` and leave `right_output` as `names(y)`.

wjones127 (changes requested):
Thanks for creating this. This works, but I think the underlying is cause is elsewhere, so might be better to fix there.

r/tests/testthat/test-dplyr-join.R
thisisnic: Good call, done now.

r/R/dplyr-join.R
thisisnic: I tried the way you suggested, but I think I still need the `post_join_projection()` call; if I revert what I've done and just make the changes suggested above, I get an error in my test as the schema expects `some_grouping.x` and `some_grouping.y` columns to have been created, but they have not.

wjones127: Try this (works for me locally):

```R
  left_output <- if (!keep && join_type == "RIGHT_OUTER") {
    setdiff(names(x), by)
  } else {
    names(x)
  }
  right_output <- if (keep || join_type %in% c("FULL_OUTER", "RIGHT_OUTER")) {
    names(y)
  } else {
    setdiff(names(y), by)
  }
```

thisisnic: Thanks, I was missing the `!keep!` from the `left_output`!

wjones127 (approved):
Looks good. Thanks!

PR #15083, opened 2022-12-24
GH-33566: [C++] Add support for nullary and n-ary aggregate functions

### LINKED ISSUE #33566: [C++] Error consuming Substrait plan which uses count function: "only unary aggregate functions are currently supported"
ARROW-17523 added support for the Substrait extension function "count", but when I write code which produces a Substrait plan which calls it, and then try to run it in Acero, I get an error.

The plan:

```r

message of type 'substrait.Plan' with 3 fields set
extension_uris {
  extension_uri_anchor: 1
  uri: "https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml"
}
extension_uris {
  extension_uri_anchor: 2
  uri: "https://github.com/substrait-io/substrait/blob/main/extensions/functions_comparison.yaml"
}
extension_uris {
  extension_uri_anchor: 3
  uri: "https://github.com/substrait-io/substrait/blob/main/extensions/functions_aggregate_generic.yaml"
}
extensions {
  extension_function {
    extension_uri_reference: 3
    function_anchor: 2
    name: "count"
  }
}
relations {
  rel {
    aggregate {
      input {
        project {
          common {
            emit {
              output_mapping: 9
              output_mapping: 10
              output_mapping: 11
              output_mapping: 12
              output_mapping: 13
              output_mapping: 14
              output_mapping: 15
              output_mapping: 16
              output_mapping: 17
            }
          }
          input {
            read {
              base_schema {
                names: "int"
                names: "dbl"
                names: "dbl2"
                names: "lgl"
                names: "false"
                names: "chr"
                names: "verses"
                names: "padded_strings"
                names: "some_negative"
                struct_ {
                  types {
                    i32 {
                      nullability: NULLABILITY_NULLABLE
                    }
                  }
                  types {
                    fp64 {
                      nullability: NULLABILITY_NULLABLE
                    }
                  }
                  types {
                    fp64 {
                      nullability: NULLABILITY_NULLABLE
                    }
                  }
                  types {
                    bool_ {
                      nullability: NULLABILITY_NULLABLE
                    }
                  }
                  types {
                    bool_ {
                      nullability: NULLABILITY_NULLABLE
                    }
                  }
                  types {
                    string {
                      nullability: NULLABILITY_NULLABLE
                    }
                  }
                  types {
                    string {
                      nullability: NULLABILITY_NULLABLE
                    }
                  }
                  types {
                    string {
                      nullability: NULLABILITY_NULLABLE
                    }
                  }
                  types {
                    fp64 {
                      nullability: NULLABILITY_NULLABLE
                    }
                  }
                }
              }
              local_files {
                items {
                  uri_file: "file:///tmp/RtmpsBsoZJ/file1915f604cff4a"
                  parquet {
                  }
                }
              }
            }
          }
          expressions {
            selection {
              direct_reference {
                struct_field {
                }
              }
              root_reference {
              }
            }
          }
          expressions {
            selection {
              direct_reference {
                struct_field {
                  field: 1
                }
              }
              root_reference {
              }
            }
          }
          expressions {
            selection {
              direct_reference {
                struct_field {
                  field: 2
                }
              }
              root_reference {
              }
            }
          }
          expressions {
            selection {
              direct_reference {
                struct_field {
                  field: 3
                }
              }
              root_reference {
              }
            }
          }
          expressions {
            selection {
              direct_reference {
                struct_field {
                  field: 4
                }
              }
              root_reference {
              }
            }
          }
          expressions {
            selection {
              direct_reference {
                struct_field {
                  field: 5
                }
              }
              root_reference {
              }
            }
          }
          expressions {
            selection {
              direct_reference {
                struct_field {
                  field: 6
                }
              }
              root_reference {
              }
            }
          }
          expressions {
            selection {
              direct_reference {
                struct_field {
                  field: 7
                }
              }
              root_reference {
              }
            }
          }
          expressions {
            selection {
              direct_reference {
                struct_field {
                  field: 8
                }
              }
              root_reference {
              }
            }
          }
        }
      }
      groupings {
        grouping_expressions {
          selection {
            direct_reference {
              struct_field {
                field: 3
              }
            }
            root_reference {
            }
          }
        }
      }
      measures {
        measure {
          function_reference: 2
          phase: AGGREGATION_PHASE_INITIAL_TO_RESULT
          output_type {
            i64 {
              nullability: NULLABILITY_NULLABLE
            }
          }
          invocation: AGGREGATION_INVOCATION_ALL
        }
      }
    }
  }
}
```

The error:

```java

Error: NotImplemented: Only unary aggregate functions are currently supported
/home/nic2/arrow/cpp/src/arrow/engine/substrait/relation_internal.cc:587  converter(aggregate_call)
/home/nic2/arrow/cpp/src/arrow/engine/substrait/serde.cc:153  FromProto(plan_rel.has_root() ? plan_rel.root().input() : plan_rel.rel(), ext_set, conversion_options)
```

I have no idea what the "phase" and "invocation" fields above do, but previous attempts to get Acero to consume this plan led to errors due to me using default values instead of the ones specified there (e.g. "Not Implemented: Unsupported aggregation phase 'AGGREGATION_PHASE_UNSPECIFIED'"), so I just changed them to see if it helped.

**Reporter**: [Nicola Crane](https://issues.apache.org/jira/browse/ARROW-18403) / @thisisnic
**Assignee**: [Felipe Oliveira](https://issues.apache.org/jira/browse/ARROW-18403)
#### PRs and other links:
- [GitHub Pull Request #15083](https://github.com/apache/arrow/pull/15083)

<sub>**Note**: *This issue was originally created as [ARROW-18403](https://issues.apache.org/jira/browse/ARROW-18403). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

felipecrv:
- [x] Add ability to pass 0 or more than 1 target fields via the Aggregate API
 - [x] Add support for nullary `count` -- `count(*)`
 - [x] Add a n-ary aggregate function to test changes `*`
 
 `*` I implemented a `"covariant(y, x)"` aggregation function and used it to test the Aggregate API changes, but it's not present in this PR now that I intend to focus on passing the CI tests and get a final review
* Closes: #33566


vibhatha:
@westonpace @wjones127 could please help with running the CIs and reviews.

westonpace (changes requested):
This seems reasonable to me, though at the time this is completely untested.  Can we introduce a simple nilary aggregate function (`COUNT(*)` is the only one I can think of) or multi-argument aggregate function (maybe something like simple linear regression) even if just for test purposes?

cpp/src/arrow/compute/exec/test_util.cc
```diff
@@ -416,7 +416,19 @@ static inline void PrintToImpl(const std::string& factory_name,
       *os << "function=" << agg.function << "<";
       if (agg.options) PrintTo(*agg.options, os);
       *os << ">,";
-      *os << "target=" << agg.target.ToString() << ",";
+      *os << "target=";
+      if (agg.target.size() == 0) {
+        *os << "*";
+      } else if (agg.target.size() == 1) {
+        *os << agg.target[0].ToString();
+      } else {
```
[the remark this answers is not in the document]
felipecrv: The else case contain a `*os << "("` before `target[0]`. Should I remove the ()-wrapping?

cpp/src/arrow/engine/substrait/extension_set.cc
```diff
@@ -883,21 +883,27 @@ ExtensionIdRegistry::SubstraitCallToArrow DecodeConcatMapping() {
...
+        ARROW_ASSIGN_OR_RAISE(compute::Expression arg, call.GetValueArg(0));
+        const FieldRef* arg_ref = arg.field_ref();
+        if (!arg_ref) {
+          return Status::Invalid("Expected an aggregate call ", call.id().uri, "#",
+                                 call.id().name, " to have a direct reference");
+        }
+        return compute::Aggregate{std::move(fixed_arrow_func), *arg_ref};
+      }
+      default:
+        break;
```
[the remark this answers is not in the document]
felipecrv: I started the PR from here and had a second commit containing `case 0` before I started changing the `Aggregate` class signature. Final version of this PR should contain the `case 0` again.

felipecrv:
> Can we introduce a simple nilary aggregate function (COUNT(*) is the only one I can think of) or multi-argument aggregate function (maybe something like simple linear regression) even if just for test purposes?

This is my plan, but I wanted to send this draft PR to get early feedback. I will add the n-ary function to the TODO list in the description of the PR.

cpp/src/arrow/compute/api_aggregate.h
```diff
@@ -186,16 +188,38 @@ class ARROW_EXPORT IndexOptions : public FunctionOptions {
...
+            std::vector<FieldRef> target, std::string name = "")
+      : function(std::move(function)),
+        options(std::move(options)),
+        target(std::move(target)),
+        name(std::move(name)) {}
+
+  Aggregate(std::string function, std::shared_ptr<FunctionOptions> options,
+            FieldRef target, std::string name = "")
+      : Aggregate(std::move(function), std::move(options),
+                  /*target=*/std::vector<FieldRef>{std::move(target)}, std::move(name)) {}
```
vibhatha, line 203: do we need `/*target=*/` here?

cpp/src/arrow/compute/exec/aggregate.cc
```diff
@@ -38,33 +38,38 @@ namespace internal {
...
-    const std::vector<TypeHolder>& in_types) {
+    const std::vector<std::vector<TypeHolder>>& in_types) {
   if (aggregates.size() != in_types.size()) {
     return Status::Invalid(aggregates.size(), " aggregate functions were specified but ",
                            in_types.size(), " arguments were provided.");
   }
 
   std::vector<const HashAggregateKernel*> kernels(in_types.size());
 
+  std::vector<TypeHolder> aggregate_in_types;
```
vibhatha, line 49: maybe reserve size to the vector?

cpp/src/arrow/compute/exec/aggregate.cc
```diff
@@ -121,21 +129,36 @@ Result<Datum> GroupBy(const std::vector<Datum>& arguments, const std::vector<Dat
...
+  if (!aggregates.empty()) {
     ARROW_ASSIGN_OR_RAISE(args_batch, ExecBatch::Make(arguments));
 
     // Construct and initialize HashAggregateKernels
-    auto argument_types = args_batch.GetTypes();
+    std::vector<std::vector<TypeHolder>> aggs_argument_types(aggregates.size());
+    {
+      // Contains the flattened list of aggregate arguments. We use the size of
+      // each Aggregate::target to re-group the aggregate argument types.
+      auto argument_types = args_batch.GetTypes();
```
vibhatha, line 140: should we do `const auto&` here?

cpp/src/arrow/compute/api_aggregate.h
felipecrv: Redundant indeed. I will remove.

cpp/src/arrow/compute/api_aggregate.h
```diff
@@ -186,16 +188,38 @@ class ARROW_EXPORT IndexOptions : public FunctionOptions {
...
   /// the name of the aggregation function
   std::string function;
 
   /// options for the aggregation function
   std::shared_ptr<FunctionOptions> options;
 
-  // fields to which aggregations will be applied
-  FieldRef target;
+  /// zero or more fields to which aggregations will be applied
+  std::vector<FieldRef> target;
```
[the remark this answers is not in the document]
felipecrv: I thought about pluralizing this but it became useful to refer to this as the individual target when processing multiple aggregate calls -- each one with its own target. A names that is pluralizable here is nice because of that. So keeping `target` or renaming it to something like `target_bundle`, `target_set` would keep the code that process this more understandable.

cpp/src/arrow/compute/exec/aggregate.cc
felipecrv: Given that this is the loop body:

```cpp
      // {in_types[i]..., uint32()}
      aggregate_in_types = in_types[i];
      aggregate_in_types.emplace_back(uint32());
```

do you prefer a `.reserve(2)` for the common case of `in_types[i].size() == 1` or should I reserve within the loop body like this?

```diff
       // {in_types[i]..., uint32()}
+      aggregate_in_types.reserve(in_types[i].size() + 1);
       aggregate_in_types = in_types[i];
       aggregate_in_types.emplace_back(uint32());
```

cpp/src/arrow/compute/api_aggregate.h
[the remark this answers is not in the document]
vibhatha: Let’s keep target.

cpp/src/arrow/compute/exec/aggregate.cc
vibhatha: Sorry if I misread this, if it’s hard to guess the exact value, let’s keep it as it is. 

cc @westonpace thoughts?

cpp/src/arrow/compute/exec/aggregate.cc
felipecrv: It can't be a reference because `GetTypes` doesn't return a reference. (Technically it can, but won't really prevent the construction of the vector within the stack).

I think moving the elements from the vector like this will give us more: `TypeHolder` contains a `shared_ptr` -- cheaper to move than to copy.

```diff
       for (size_t j = 0; j < aggregates.size(); j++) {
         const size_t num_agg_args = aggregates[j].target.size();
         for (size_t k = 0; k < num_agg_args && i < argument_types.size(); k++, i++) {
-          aggs_argument_types[j].push_back(argument_types[i]);
+          aggs_argument_types[j].push_back(std::move(argument_types[i]));
         }
       }
 ```

File: cpp/src/arrow/compute/exec/aggregate.cc [hunk 7]
```diff
@@ -38,33 +38,38 @@ namespace internal {
...
 }
 
 Result<std::vector<std::unique_ptr<KernelState>>> InitKernels(
     const std::vector<const HashAggregateKernel*>& kernels, ExecContext* ctx,
-    const std::vector<Aggregate>& aggregates, const std::vector<TypeHolder>& in_types) {
+    const std::vector<Aggregate>& aggregates,
+    const std::vector<std::vector<TypeHolder>>& in_types) {
   std::vector<std::unique_ptr<KernelState>> states(kernels.size());
 
+  std::vector<TypeHolder> agg_in_types;
```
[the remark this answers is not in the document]
felipecrv: Same question as above.

cpp/src/arrow/compute/exec/aggregate.cc
felipecrv: I'm adding a commit fixing both cases using the second approach.

cpp/src/arrow/compute/exec/aggregate.cc
felipecrv: I fixed up the commit with the `std::move` as described above.

cpp/src/arrow/compute/exec/aggregate.cc
```diff
@@ -38,33 +38,39 @@ namespace internal {
...
                            in_types.size(), " arguments were provided.");
   }
 
   std::vector<const HashAggregateKernel*> kernels(in_types.size());
 
+  std::vector<TypeHolder> aggregate_in_types;
   for (size_t i = 0; i < aggregates.size(); ++i) {
     ARROW_ASSIGN_OR_RAISE(auto function,
                           ctx->func_registry()->GetFunction(aggregates[i].function));
+    // {in_types[i]..., uint32()}
```
westonpace, line 53: This comment is a bit cryptic

cpp/src/arrow/compute/exec/aggregate.cc
```diff
@@ -38,33 +38,39 @@ namespace internal {
...
+    aggregate_in_types.reserve(in_types[i].size() + 1);
```
westonpace, line 54:
```suggestion
```
I'm not sure what purpose this `reserve` has

cpp/src/arrow/compute/exec/aggregate.cc
```diff
@@ -75,13 +81,13 @@ Result<std::vector<std::unique_ptr<KernelState>>> InitKernels(
...
     KernelContext kernel_ctx{ctx};
-    ARROW_ASSIGN_OR_RAISE(states[i],
-                          kernels[i]->init(&kernel_ctx, KernelInitArgs{kernels[i],
-                                                                       {
-                                                                           in_types[i],
-                                                                           uint32(),
-                                                                       },
-                                                                       options}));
+    // {in_types[i]..., uint32()}
+    agg_in_types.reserve(in_types[i].size() + 1);
```
westonpace, line 85:
```suggestion
```
Same as above.  Cryptic comment and dubious reserve.

cpp/src/arrow/compute/exec/plan_test.cc
```diff
@@ -1298,17 +1305,53 @@ TEST(ExecPlanExecution, ScalarSourceScalarAggSink) {
...
+      {"source", SourceNodeOptions{scalar_data.schema, scalar_data.gen(/*parallel=*/false,
+                                                                       /*slow=*/false)}},
+      {"aggregate", AggregateNodeOptions{/*aggregates=*/{
+                        {"count_all", "count(*)"},
+                    }}},
+      {"sink", SinkNodeOptions{&sink_gen}},
+  });
+
+  // index can't be tested as it's order-dependent
+  // mode/quantile can't be tested as they're technically vector kernels
```
westonpace, line 1346: I'm not sure this comment is needed?

cpp/src/arrow/compute/exec/plan_test.cc
```diff
@@ -1298,17 +1305,53 @@ TEST(ExecPlanExecution, ScalarSourceScalarAggSink) {
...
+  ASSERT_OK(sequence.AddToPlan(plan.get()));
+
+  auto exec_batch = ExecBatchFromJSON({int64()}, {ArgShape::SCALAR}, R"([[6]])");
+
+  ASSERT_THAT(StartAndCollect(plan.get(), sink_gen),
+              Finishes(ResultWith(UnorderedElementsAreArray({
+                  std::move(exec_batch),
+              }))));
 }
```
westonpace, line 1355:
```suggestion
TEST(ExecPlanExecution, ScalarSourceStandaloneNullaryScalarAggSink) {
  BatchesWithSchema scalar_data;
  scalar_data.batches = {
      ExecBatchFromJSON({int32(), boolean()}, {ArgShape::SCALAR, ArgShape::SCALAR},
                        "[[5, null], [5, false], [5, false]]"),
      ExecBatchFromJSON({int32(), boolean()}, "[[5, true], [null, false], [7, true]]")};
  scalar_data.schema = schema({
      field("a", int32()),
      field("b", boolean()),
  });

  Declaration plan = Declaration::Sequence(
      {{"source",
        SourceNodeOptions{scalar_data.schema, scalar_data.gen(/*parallel=*/false,
                                                              /*slow=*/false)}},
       {"aggregate", AggregateNodeOptions{/*aggregates=*/{
                         {"count_all", "count(*)"},
                     }}}});
  ASSERT_OK_AND_ASSIGN(BatchesWithCommonSchema actual_batches,
                       DeclarationToExecBatches(std::move(plan)));

  auto expected = ExecBatchFromJSON({int64()}, {ArgShape::SCALAR}, R"([[6]])");
  AssertExecBatchesEqualIgnoringOrder(actual_batches.schema, actual_batches.batches,
                                      {expected});
}
```

We are slowly trying to move away from `StartAndCollect` to `DeclarationToXyz` methods in tests.  The latter is a bit more compact and reflects the public API (`StartAndCollect` is an internal method)  You don't need to fix old tests but new tests should be written with the new style.

cpp/src/arrow/compute/exec/test_util.cc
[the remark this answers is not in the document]
westonpace: Ah, I see.  What you have is better then.

python/pyarrow/table.pxi
```diff
@@ -5364,30 +5364,31 @@ list[tuple(str, str, FunctionOptions)]
         ----
         keys: [["a","b","c"]]
         """
-        columns = [a[0] for a in aggregations]
+        target_cols = [a[0] if isinstance(a[0], (list, tuple)) else [a[0]] for a in aggregations]
```
westonpace, line 5367: I don't know if we want to be pythonic and accept any iterable here or not.  However, I suppose we're pretty strict elsewhere.  Can you update the example in the docstring with a call to `count(*)`?

CC @amol- or @jorisvandenbossche for canonical advice

python/pyarrow/includes/libarrow.pxd
```diff
@@ -2489,6 +2489,8 @@ cdef extern from "arrow/compute/exec/aggregate.h" namespace \
     cdef cppclass CAggregate "arrow::compute::Aggregate":
         c_string function
         shared_ptr[CFunctionOptions] options
+        vector[CFieldRef] target
```
westonpace, line 2492: I'm confused how we got away with this field never existing.  Perhaps we never use `Aggregate` directly?

westonpace (changes requested):
Sorry for the delay in review.  This is pretty good.  A few suggestions but I don't see anything critical.

python/pyarrow/includes/libarrow.pxd
felipecrv: I was very confused about that as well! The `GroupBy` code called by Python ignored the `FieldRef`s coming from the `Aggregate::target` and mapped the `arguments` to the group-by by position. Now that the code relies on the `.size()` of `Aggregate::target`, we started having issues.

cpp/src/arrow/compute/exec.cc
```diff
@@ -163,6 +163,34 @@ Result<ExecBatch> ExecBatch::Make(std::vector<Datum> values) {
       continue;
     }
 
+    if (length != value.length()) {
+      // all the arrays should have the same length
+      return -1;
+    }
```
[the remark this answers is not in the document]
felipecrv: It's a recoverable error in the `ExecBatch::Make` loop when there are only scalar `Datum`s. I can change it to return an `optional<int64_t>` because I want to express that inference of the length failing is a non-fatal. `GroupBy` can use this, but ultimately it's `ExecBatch::Make` that performs the last validation and returns a more appropriate error `Status`.

cpp/src/arrow/compute/exec.cc
```diff
@@ -163,6 +163,34 @@ Result<ExecBatch> ExecBatch::Make(std::vector<Datum> values) {
...
+  }
+
+  return length == -1 ? 1 : length;
+}
+
+Result<ExecBatch> ExecBatch::Make(std::vector<Datum> values, int64_t length) {
+  if (length < 0) {
```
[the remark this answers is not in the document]
felipecrv: That was my initial idea (avoid the repetition) as well, but for the reasons described above, added more complications. I can append a commit to this PR converging this to a single function and we can compare.

cpp/src/arrow/compute/exec.cc
```diff
@@ -147,9 +147,9 @@ ExecBatch ExecBatch::Slice(int64_t offset, int64_t length) const {
   return out;
 }
 
-Result<ExecBatch> ExecBatch::Make(std::vector<Datum> values) {
+int64_t ExecBatch::InferLength(const std::vector<Datum>& values) {
```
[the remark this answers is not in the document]
felipecrv: answered in the other threads below

cpp/src/arrow/compute/exec/aggregate.cc
felipecrv: Should I just remove it? It's an allusion to the parameter unpacking syntax from C++ templates.

cpp/src/arrow/compute/exec/aggregate_node.cc
```diff
@@ -46,14 +46,26 @@ namespace {
...
+    } else {
+      *ss << input_schema.field(target[0])->name();
+      for (size_t k = 1; k < target.size(); k++) {
+        *ss << ", " << input_schema.field(target[k])->name();
+      }
+    }
     if (aggs[i].options) {
-      *ss << ", " << aggs[i].options->ToString();
+      auto* options_type = aggs[i].options->options_type();
+      if (options_type->num_properties() > 0) {
```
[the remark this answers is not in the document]
felipecrv: This function is sometimes (maybe at all times) called after NULL `Aggregate::options` get their `options` replaced by a pointer to a instance of the default options for that aggregate function.

That happens here https://github.com/apache/arrow/blob/master/cpp/src/arrow/compute/exec/aggregate_node.cc#L113

So I ended up adding the first function that needs no options at all, but providing default options is required. So I created a options type with 0 properties. That works as a placeholder for the future, when this function might require properties.

I could drop all this and make this render the string `count_all(*, {})`.

cpp/src/arrow/compute/exec/plan_test.cc
felipecrv: not added by me. I can remove.

felipecrv: oh.  Bad copy and paste.

cpp/src/arrow/compute/kernels/hash_aggregate.cc
```diff
@@ -108,19 +108,31 @@ Result<TypeHolder> ResolveGroupOutputType(KernelContext* ctx,
...
+                            OutputType(ResolveGroupOutputType)),
+      std::move(init));
+}
+
+HashAggregateKernel MakeUnaryKernel(KernelInit init) {
+  return MakeKernel(KernelSignature::Make({InputType(Type::UINT32)},
+                                          OutputType(ResolveGroupOutputType)),
+                    std::move(init));
+}
+
```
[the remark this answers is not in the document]
felipecrv: Very confusing. I decided to mention the confusing arity because the `Arity` is mentioned in the context where `Make*Kernel` is called:

```cpp
  {
    auto func = std::make_shared<HashAggregateFunction>(
        "hash_count", Arity::Binary(), hash_count_doc, &default_count_options);

    DCHECK_OK(func->AddKernel(
        MakeKernel(InputType::Any(), HashAggregateInit<GroupedCountImpl>)));
    DCHECK_OK(registry->AddFunction(std::move(func)));
  }

  {
    auto func = std::make_shared<HashAggregateFunction>(
        "hash_count_all", Arity::Unary(), hash_count_all_doc, &default_count_all_options);

    DCHECK_OK(func->AddKernel(MakeUnaryKernel(HashAggregateInit<GroupedCountAllImpl>)));
    auto status = registry->AddFunction(std::move(func));
    DCHECK_OK(status);
  }
```

As I understood it, there is the arity of the kernels and the arity of the aggregation functions. The former needs to have arity+1 to support aggregations combined with group-by.

cpp/src/arrow/engine/substrait/extension_set.cc
```diff
@@ -883,21 +883,37 @@ ExtensionIdRegistry::SubstraitCallToArrow DecodeConcatMapping() {
...
-    return compute::Aggregate{std::move(fixed_arrow_func), nullptr, *arg_ref, ""};
+
+    switch (call.size()) {
+      case 0: {
+        if (call.id().name == "count") {
+          fixed_arrow_func += "count_all";
+          return compute::Aggregate{std::move(fixed_arrow_func), ""};
+        }
+        return Status::Invalid("Expected aggregate call ", call.id().uri, "#",
+                               call.id().name, " to have at least one argument");
```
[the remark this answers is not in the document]
felipecrv: Exactly.

cpp/src/arrow/type.h
```diff
@@ -1692,9 +1692,7 @@ class ARROW_EXPORT FieldRef : public util::EqualityComparable<FieldRef> {
   FieldRef(int index) : impl_(FieldPath({index})) {}  // NOLINT runtime/explicit
 
   /// Construct a nested FieldRef.
-  FieldRef(std::vector<FieldRef> refs) {  // NOLINT runtime/explicit
-    Flatten(std::move(refs));
-  }
+  explicit FieldRef(std::vector<FieldRef> refs) { Flatten(std::move(refs)); }
```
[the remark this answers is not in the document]
felipecrv: I was mindful of that, but this wasted a lot of my time and is bound to waste other people's time if they start dealing with `vector<FieldRef>` anywhere.

python/pyarrow/table.pxi
felipecrv: I will update the example and Python help is very welcome! I tried to copy patterns I saw elsewhere in this codebase because I can't claim I'm able to write pythonic Python :)

cpp/src/arrow/compute/exec/plan_test.cc
felipecrv: Fixed.

cpp/src/arrow/compute/exec/aggregate.cc
felipecrv: Fixed in a manual commit.

cpp/src/arrow/compute/exec.cc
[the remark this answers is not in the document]
felipecrv: I added a commit that adds a reusable function for this.

cpp/src/arrow/compute/exec/aggregate.cc
westonpace: vector copy assignment should do a reserve internally and if it doesn't, they probably have a good reason not to :)

cpp/src/arrow/compute/exec.cc
```diff
@@ -164,13 +171,43 @@ Result<ExecBatch> ExecBatch::Make(std::vector<Datum> values) {
...
+            "Cannot infer ExecBatch length without at least one value");
+      }
+      break;
+
+    case kInvalidValues:
+      return Status::Invalid(
+          "Arrays used to construct an ExecBatch must have equal length");
+
+    default:
+      DCHECK(length < 0 || length == inferred_length);
```
westonpace, line 208: I wonder if we should make `length != inferred_length` an invalid status instead of a failed DCHECK.  I could go either way on this.

cpp/src/arrow/compute/exec/aggregate_node.cc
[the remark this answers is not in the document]
westonpace: I kind of like `count_all(*, {})` to help distinguish options from arguments.

cpp/src/arrow/compute/kernels/hash_aggregate.cc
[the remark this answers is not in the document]
westonpace: Ok, I agree.  Unary makes sense given the context.

westonpace (approved):
This looks good.  I think I lean slightly towards getting rid of `num_properties` but I'll leave it up to you.

cpp/src/arrow/compute/exec/aggregate.cc
westonpace: nvm, I see.  You are copying and adding one.  Still, I think it's simpler without the reserve.

cpp/src/arrow/compute/exec/aggregate_node.cc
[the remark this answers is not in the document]
felipecrv: I will revert the commit that introduced `num_properties` and push again.

cpp/src/arrow/compute/exec.cc
felipecrv: Makes sense as the caller could proceed using the mistaken `length` for other things. I will add a commit changing this.

felipecrv: Done.

cpp/src/arrow/compute/exec/aggregate_node.cc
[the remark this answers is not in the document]
felipecrv: Done.

cpp/src/arrow/compute/exec.h
```diff
@@ -174,7 +175,12 @@ struct ARROW_EXPORT ExecBatch {
 
   explicit ExecBatch(const RecordBatch& batch);
 
-  static Result<ExecBatch> Make(std::vector<Datum> values);
+  /// \brief Infers the ExecBatch length from values.
```
pitrou, line 178:
```suggestion
  /// \brief Infer the ExecBatch length from values.
```

cpp/src/arrow/compute/exec/aggregate.cc
```diff
@@ -121,27 +127,48 @@ Result<Datum> GroupBy(const std::vector<Datum>& arguments, const std::vector<Dat
   ExecSpanIterator argument_iterator;
 
   ExecBatch args_batch;
-  if (!arguments.empty()) {
-    ARROW_ASSIGN_OR_RAISE(args_batch, ExecBatch::Make(arguments));
+  std::optional<int64_t> inferred_length = ExecBatch::InferLength(arguments);
+  if (!inferred_length.has_value()) {
+    inferred_length = ExecBatch::InferLength(keys);
+  }
+  DCHECK(inferred_length.has_value());
```
pitrou, line 134: Are we sure this always succeeds? Otherwise we should return an error to the caller instead of crashing out.

cpp/src/arrow/compute/exec/aggregate.cc
```diff
@@ -121,27 +127,48 @@ Result<Datum> GroupBy(const std::vector<Datum>& arguments, const std::vector<Dat
...
-    auto argument_types = args_batch.GetTypes();
+    std::vector<std::vector<TypeHolder>> aggs_argument_types(aggregates.size());
+    {
+      // Contains the flattened list of aggregate arguments. We use the size of
+      // each Aggregate::target to re-group the aggregate argument types.
+      auto argument_types = args_batch.GetTypes();
+      size_t i = 0;
+      for (size_t j = 0; j < aggregates.size(); j++) {
+        const size_t num_agg_args = aggregates[j].target.size();
+        for (size_t k = 0; k < num_agg_args && i < argument_types.size(); k++, i++) {
```
pitrou, line 148: `i < argument_types.size()` isn't necessary, is it?

cpp/src/arrow/compute/exec/aggregate.cc
```diff
@@ -121,27 +127,48 @@ Result<Datum> GroupBy(const std::vector<Datum>& arguments, const std::vector<Dat
...
+          aggs_argument_types[j].push_back(std::move(argument_types[i]));
+        }
+      }
+      DCHECK_EQ(i, argument_types.size())
+          << "argument_types should contain input types for all the aggregates.";
+    }
 
-    ARROW_ASSIGN_OR_RAISE(kernels, GetKernels(ctx, aggregates, argument_types));
+    ARROW_ASSIGN_OR_RAISE(kernels, GetKernels(ctx, aggregates, aggs_argument_types));
```
pitrou, line 156: Isn't it slightly silly to bother merging all aggregates together, and then call `GetKernels` which just iterates over them? Or am I missing something?

cpp/src/arrow/compute/api_aggregate.h
```diff
@@ -56,6 +58,13 @@ class ARROW_EXPORT ScalarAggregateOptions : public FunctionOptions {
   uint32_t min_count;
 };
 
+class ARROW_EXPORT CountAllOptions : public FunctionOptions {
```
pitrou, line 61: Hmm, what is this for? If a given function doesn't support any specific options, then a `nullptr` FunctionOptions should simply be passed.

cpp/src/arrow/compute/kernels/hash_aggregate.cc
```diff
@@ -223,6 +235,53 @@ void VisitGroupedValuesNonNull(const ExecSpan& batch, ConsumeValue&& valid_func)
...
+    num_groups_ = new_num_groups;
+    return counts_.Append(added_groups * sizeof(int64_t), 0);
+  }
+
+  Status Merge(GroupedAggregator&& raw_other,
+               const ArrayData& group_id_mapping) override {
+    auto other = checked_cast<GroupedCountAllImpl*>(&raw_other);
+
+    auto counts = reinterpret_cast<int64_t*>(counts_.mutable_data());
+    auto other_counts = reinterpret_cast<const int64_t*>(other->counts_.mutable_data());
```
pitrou, line 256:
```suggestion
    auto other_counts = reinterpret_cast<const int64_t*>(other->counts_.data());
```

cpp/src/arrow/compute/exec.cc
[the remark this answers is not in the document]
felipecrv: I've written a single helper function (private) and used it to implement the public `InferLength` and `Make` functions.

cpp/src/arrow/compute/exec/aggregate.cc
felipecrv: Yes. The length should always be inferrable from the `keys`.

cpp/src/arrow/compute/api_aggregate.h
felipecrv: Unfortunately, I can't pass `nullptr` as the "default options" in the kernel configuration. I tried, but then it causes crashes because when you pass no options in the kernel calls, `options` is set to the default options as after some point, `options` is expected to be non-null.

cpp/src/arrow/compute/exec.cc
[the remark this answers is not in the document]
felipecrv: I can change the public `InferLength` member function to return a `Result` instead of an `std::optional`.

cpp/src/arrow/compute/api_aggregate.h
pitrou: Hmm, hopefully this can be fixed, then. Some other compute functions don't have an associated options class.

cpp/src/arrow/compute/exec/aggregate.cc
felipecrv: Not needed, but perhaps good to have to make it obvious that access never goes over the array size?

cpp/src/arrow/compute/exec/aggregate.cc
felipecrv: Hmmm yes. I didn't consider changing the type signatures of these functions, but they are internal and don't have many callsites, so I think I can collapse these transformation steps.

cpp/src/arrow/compute/exec/aggregate.cc
felipecrv: I'm pushing a commit that turns this into a `ARROW_ASSIGN_OR_RAISE`.

cpp/src/arrow/compute/exec/aggregate.cc
felipecrv: Pushing a commit that removes it. The DCHECK after the loop makes it clear.

cpp/src/arrow/compute/exec.cc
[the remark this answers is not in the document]
felipecrv: Done.

cpp/src/arrow/compute/api_aggregate.h
felipecrv: I will open a separate PR fixing this then.

cc @westonpace @vibhatha

python/pyarrow/table.pxi
```diff
@@ -5358,36 +5358,45 @@ list[tuple(str, str, FunctionOptions)]
...
+        keys: [["a","b","c"]]
         >>> t.group_by("keys").aggregate([])
         pyarrow.Table
         keys: string
         ----
         keys: [["a","b","c"]]
         """
-        columns = [a[0] for a in aggregations]
+        target_cols = [a[0] if isinstance(a[0], (list, tuple)) else [
+            a[0]] for a in aggregations]
```
jorisvandenbossche, line 5375: Why is this change needed? Because now the target cols are always expected to be a list?

python/pyarrow/table.pxi
```diff
@@ -5358,36 +5358,45 @@ list[tuple(str, str, FunctionOptions)]
...
             group_by_aggrs.append(aggr)
 
         # Build unique names for aggregation result columns
         # so that it's obvious what they refer to.
-        column_names = [
-            aggr_name.replace("hash", col_name)
-            for col_name, (aggr_name, _) in zip(columns, group_by_aggrs)
+        out_column_names = [
+            aggr_name.replace("hash", "_".join(target))
+            for target, aggr_name, _ in group_by_aggrs
```
jorisvandenbossche, line 5391: We might want to special case "count_all" here (or nullary kernels in general, if we can infer that), because right now this adds a leading underscore to the resulting name: `"_count_all"`

python/pyarrow/table.pxi
jorisvandenbossche: Looking further at the below changes, maybe we can combine all this extraction/validation in the single `for aggr in aggrfuncs:` loop below?  
That might give more readable code (compared to several of those list comprehensions), something like:

```
group_by_aggrs = []
for aggr in aggregations:
    if len(aggr) == 2:
        target, func = aggr
        opt = None
    elif:
        target, func, opt = aggr
    # check func name ...
    # convert target to list if needed ...
    group_by_aggrs.append((target, func, opt))
```

python/pyarrow/table.pxi
```diff
@@ -5334,7 +5334,7 @@ class TableGroupBy:
         Parameters
         ----------
         aggregations : list[tuple(str, str)] or \
-list[tuple(str, str, FunctionOptions)]
+list[tuple(str|list[str]|tuple(str*), str, FunctionOptions)]
```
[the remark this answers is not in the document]
jorisvandenbossche: Basically, IIUC the "only" thing that this PR is adding is that the target column name (first argument of the tuple) can be a list of names instead of a single name (string). API wise I think that is reasonable? (I am not sure alternatives would be better, like not passing any column name and inferring that from the length of the tuple ..).

(one alternative would be to make it more structured, and provide a some namedtuple or small class to define an "aggregation", and then this argument would be a list of such objects. But that won't necessarily make it easier to use, and we probably want to keep supporting the plain tuple as well)

But personally I do find the above line explaining the type completely unreadable, though. I would simplify to just "list of tuples" and explain what each tuple can contain in the explanation below.

python/pyarrow/_compute.pyx
```diff
@@ -2202,12 +2202,18 @@ def _group_by(args, keys, aggregations):
...
+    field_ref = 0
+    for aggr_arg_names, aggr_func_name, aggr_opts in aggregations:
         c_aggr.function = tobytes(aggr_func_name)
         if aggr_opts is not None:
             c_aggr.options = (<FunctionOptions?>aggr_opts).wrapped
         else:
             c_aggr.options = <shared_ptr[CFunctionOptions]>nullptr
+        for _ in aggr_arg_names:
+            c_aggr.target.push_back(CFieldRef(<int> field_ref))
+            field_ref += 1
```
jorisvandenbossche, line 2215: Not for this PR, but would it make sense for this code to start referring by field name instead of integer field? (but then we would need to be able to pass the `args` as an actual table/recordbatch instead of a list of arrays?)

jorisvandenbossche (commented):
Took a look at the python side of the PR, added a few comments but looks good!

Can you add the new count_all kernel to the docs? (the tables in https://arrow.apache.org/docs/dev/cpp/compute.html#grouped-aggregations-group-by and https://arrow.apache.org/docs/dev/python/compute.html#grouped-aggregations)

This PR also (automatically) exposes a "count_all" kernel in pyarrow.compute, but I am not sure how useful this is:

```
In [1]: import pyarrow.compute as pc

In [2]: pc.count_all()
Out[2]: <pyarrow.Int64Scalar: 0>
```

It just returns zero .. I am not sure if there are places currently in the Python API where this could actually be used right now.

python/pyarrow/table.pxi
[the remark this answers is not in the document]
felipecrv: @jorisvandenbossche thank you for taking a look!

I will try to rewrite the doc-comment and push again.

python/pyarrow/table.pxi
felipecrv: I kinda did this on purpose to keep the rule more general, but I understand that the result might seem weird. I will special case nullary aggregations.

python/pyarrow/_compute.pyx
felipecrv: I'm not super familiar with the integration here but I think you're right and the Python side doesn't seem to provide named columns.

westonpace: I had simplified this in #14867 to just get rid of the field refs but I think I'll need to revisit that logic after this merges.  I think it could be made a lot simpler.

Today we are going Table -> vector<Array> -> C++ -> Table -> ExecPlan -> Table -> python

Because python is giving C++ a vector of arrays there are no "names" and named refs wouldn't make sense.  It also means that C++ has to invent names for the output columns.  Then python just renames and clobbers those invented names.

I think this would be a lot simpler as just:

Table -> C++ -> ExecPlan -> Table -> python

I'll put #14867 on hold for now and take a look again when this merges.

cpp/src/arrow/compute/exec/aggregate.cc
felipecrv: Clustering the types as a `vector<vector<TypeHolder>>` made the logic of both `GetKernels`, `InitKernels`, and `ResolverKernels` much simpler. Otherwise I need to inline this loop logic into these 3 functions. That avoids allocating extra vectors, but is it really worth it?

cpp/src/arrow/compute/api_aggregate.h
felipecrv: It's a small fix, so I'm pushing a commit here.

python/pyarrow/table.pxi
felipecrv: `_` removed

felipecrv:
> Some of these failing CI checks are a segmentation fault in the group_by method so I think they are relevant: https://github.com/apache/arrow/actions/runs/3944389169/jobs/6750236015

@westonpace these are due to the removal of the checks against `i` that were suggested above. I reverted the commit, but that uncovered the fact that some extra input validation is needed. I'm investigating and will push a fix soon.

> This PR also (automatically) exposes a "count_all" kernel in pyarrow.compute, but I am not sure how useful this is:

@jorisvandenbossche I changed the python code to never expose nullary scalar aggregate functions.

felipecrv:
@pitrou @westonpace @jorisvandenbossche this is ready for another round of reviews.

python/pyarrow/table.pxi
```diff
@@ -5368,6 +5368,9 @@ list[tuple(str, str, FunctionOptions)]
             List of tuples made of aggregation column names followed
             by function names and optionally aggregation function options.
             Pass empty list to get a single row for each group.
+            The column names can be a string, an empty list or a list of
+            column names, for unary, nullary and n-ary aggregation functions
+            respectively.
```
jorisvandenbossche, line 5373:
```suggestion
            List of tuples, where each tuple is one aggregation specification
            and consists of: aggregation column name followed
            by function name and optionally aggregation function option.
            Pass empty list to get a single row for each group.
            The column name can be a string, an empty list or a list of
            column names, for unary, nullary and n-ary aggregation functions
            respectively.
```

jorisvandenbossche (approved):
The latest python part looks good!

westonpace (approved):
I tested by rebasing #14867 on top of this and everything looked good.

cpp/src/arrow/compute/exec.h
```diff
@@ -174,7 +175,10 @@ struct ARROW_EXPORT ExecBatch {
 
   explicit ExecBatch(const RecordBatch& batch);
 
-  static Result<ExecBatch> Make(std::vector<Datum> values);
+  /// \brief Infer the ExecBatch length from values.
+  static Result<int64_t> InferLength(const std::vector<Datum>& values);
```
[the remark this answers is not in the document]
felipecrv: This would be the docs for the internal (.cc-only) `DoInferLength`, this one returns `Status` (wrapped in `Result`) for errors. I will rebase to fix conflicts and add these docs to `DoInferLength` instead.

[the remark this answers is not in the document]
westonpace: Yes, you're correct, this suggestion is in the wrong place.  Given it's internal you could probably use `//` and be more concise as well.  At a glance it wasn't obvious to me that the operation could "fail" and a small comment to explain those scenarios would be good.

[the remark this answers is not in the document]
westonpace: Just looked at what you added and it looks good

PR #15088, opened 2022-12-25
GH-15084: [Ruby] Use common keys when keys.nil? in Table#join

### LINKED ISSUE #15084: [Ruby] Use common keys when keys.nil? in Table#join
#### Target method

Arrow::Table#join

#### Target argument

keys

#### Proposed feature

Implicit common key(s) will used when `keys` is nil.

#### Impact of this request

`keys` will become an optional argument and nil may be the default value (keys = nil).

I am testing this feature in RedAmber on the line below.

https://github.com/heronshoes/red_amber/blob/main/lib/red_amber/data_frame_combinable.rb#L382-L383

### Component(s)

Ruby

heronshoes:
### What I did

* Add support for `keys = nil` in `Table#join`
  * When `keys.nil?` common keys in self and right will automatically set to keys.
* Add assertions without argument keys.
* Add missing comments for other overload cases.

### What I checked

* It works the same way in RedAmber.

### Related Issue

* Closes: #15084

### DIFF
--- a/ruby/red-arrow/lib/arrow/table.rb
+++ b/ruby/red-arrow/lib/arrow/table.rb
@@ -478,6 +478,8 @@ module Arrow
     #   @macro join_common_before
     #   @macro join_common_after
     #
+    # @since 11.0.0
+    #
     # @overload join(right, key, type: :inner, left_outputs: nil, right_outputs: nil)
     #   Join right by a key.
     #
--- a/ruby/red-arrow/lib/arrow/table.rb
+++ b/ruby/red-arrow/lib/arrow/table.rb
@@ -507,7 +507,7 @@ module Arrow
     #
     # @since 7.0.0
     def join(right, keys=nil, type: :inner, left_outputs: nil, right_outputs: nil)
-      keys ||= columns.map(&:name).intersection(right.columns.map(&:name))
+      keys ||= column_names.intersection(right.column_names)
       plan = ExecutePlan.new
       left_node = plan.build_source_node(self)
       right_node = plan.build_source_node(right)
--- a/ruby/red-arrow/lib/arrow/table.rb
+++ b/ruby/red-arrow/lib/arrow/table.rb
@@ -448,40 +448,52 @@ module Arrow
       self.class.new(schema, packed_arrays)
     end
 
+    # Join another Table by matching with keys.
+    #
+    # @!macro join_common_before
+    #   @param right [Arrow::Table] The right table.
+    #
+    #   Join columns with `right` on join key columns.
+    #
+    # @!macro join_common_after
+    #   @param type [Arrow::JoinType] How to join.
+    #   @param left_outputs [::Array<String, Symbol>] Output columns in
+    #     `self`.
+    #
+    #     If both of `left_outputs` and `right_outputs` aren't
+    #     specified, all columns in `self` and `right` are
+    #     outputted.
+    #   @param right_outputs [::Array<String, Symbol>] Output columns in
+    #     `right`.
+    #
+    #     If both of `left_outputs` and `right_outputs` aren't
+    #     specified, all columns in `self` and `right` are
+    #     outputted.
+    #   @return [Arrow::Table]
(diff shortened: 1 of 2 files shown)


ruby/red-arrow/lib/arrow/table.rb
```diff
@@ -492,7 +504,8 @@ def pack
     #   @macro join_common_after
     #
     # @since 7.0.0
-    def join(right, keys, type: :inner, left_outputs: nil, right_outputs: nil)
+    def join(right, keys = nil, type: :inner, left_outputs: nil, right_outputs: nil)
```
kou, line 507: Could you remove spaces around `=` for default value? We use the style in this project.

```suggestion
    def join(right, keys=nil, type: :inner, left_outputs: nil, right_outputs: nil)
```

ruby/red-arrow/test/test-table.rb
```diff
@@ -1049,27 +1049,29 @@ def setup
...
-                                    ]),
-                   table1.join(table2, "key"))
+      expected = Arrow::Table.new([
+                                    ["key", [1, 3]],
+                                    ["number", [10, 30]],
+                                    ["key", [1, 3]],
+                                    ["string", ["one", "three"]]
+                                  ])
+      assert_equal(expected, table1.join(table2, "key"))
+      assert_equal(expected, table1.join(table2))
```
kou, line 1059: Could you create a new test without changing existing `keys: String` test?
`keys: String` test is for specifying a `String` as `keys`. But new assertion is for specifying nothing. `test("no keys")` is a new test name candidate.

ruby/red-arrow/test/test-table.rb
```diff
@@ -1049,27 +1049,29 @@ def setup
...
-                                    ]),
-                   table1.join(table2, :key))
+      expected = Arrow::Table.new([
+                                    ["key", [1, 3]],
+                                    ["number", [10, 30]],
+                                    ["key", [1, 3]],
+                                    ["string", ["one", "three"]]
+                                  ])
+      assert_equal(expected, table1.join(table2, :key))
+      assert_equal(expected, table1.join(table2))
```
kou, line 1074: This is redundant. Because it's the same assertion as the above newly added assertion.

ruby/red-arrow/test/test-table.rb
```diff
@@ -1079,15 +1081,16 @@ def setup
...
+      expected = Arrow::Table.new([
+                                    ["key1", [1, 2]],
+                                    ["key2", [100, 20]],
+                                    ["number", [1100, 2020]],
+                                    ["key1", [1, 2]],
+                                    ["key2", [100, 20]],
+                                    ["string", ["1-100", "2-20"]],
+                                  ])
+      assert_equal(expected, table1.join(table2, ["key1", :key2]))
+      assert_equal(expected, table1.join(table2))
```
kou, line 1093: ditto.

ruby/red-arrow/lib/arrow/table.rb
heronshoes: OK. Thank you.

ruby/red-arrow/test/test-table.rb
heronshoes: Thank you for your kind comment.

ruby/red-arrow/lib/arrow/table.rb
```diff
@@ -448,40 +448,52 @@ def pack
...
+    #   If key(s) are not supplied, common keys in self and right are used.
     #
-    #       If both of `left_outputs` and `right_outputs` aren't
-    #       specified, all columns in `self` and `right` are
-    #       outputted.
-    #     @return [Arrow::Table]
-    #       The joined `Arrow::Table`.
+    #   @macro join_common_before
+    #   @macro join_common_after
+    #
```
kou, line 480:
```suggestion
    #
    # @since 11.0.0
    #
```

kou (approved):
+1

heronshoes:
Thanks!!

PR #15090, opened 2022-12-25
GH-15087: [Release] Slow down downloading RC binaries from GitHub

### LINKED ISSUE #15087: [Release] Add delay/retry to GitHub artifact downloads in download_rc_binaries.py

From https://github.com/apache/arrow-adbc/issues/271#issuecomment-1364587840

GitHub Release artifact downloads may fail with 403 if we're too aggressive. We should add a bit of delay and/or retry to avoid this.

### Component(s)

Release

lidavidm:
GitHub appears to have a rate limit on downloading artifacts from a release, which we easily run into. Adding a bit of a delay appears to avoid this. Also, force cURL itself retry failed downloads.
* Closes: #15087

### DIFF
--- a/dev/release/download_rc_binaries.py
+++ b/dev/release/download_rc_binaries.py
@@ -22,8 +22,10 @@ import concurrent.futures as cf
 import functools
 import json
 import os
+import random
 import re
 import subprocess
+import time
 import urllib.request
 
 
@@ -190,12 +192,18 @@ class GitHub(Downloader):
             print("Already downloaded", dest_path)
             return
 
-        return self._download_url(
+        delay = random.randint(0, 3)
+        print(f"Waiting {delay} seconds to avoid rate limit")
+        time.sleep(delay)
+
+        self._download_url(
             url,
             dest_path,
             extra_args=[
                 "--header",
                 "Accept: application/octet-stream",
+                # Also retry 403s
+                "--retry-all-errors",
             ],
         )
 


dev/release/download_rc_binaries.py
```diff
@@ -190,12 +192,18 @@ def _download_file(self, dest, asset):
...
+        time.sleep(delay)
+
+        self._download_url(
             url,
             dest_path,
             extra_args=[
                 "--header",
                 "Accept: application/octet-stream",
+                # Also retry 403s
+                "--retry-all-errors",
```
kou, line 206: It seems that we need to specify `--retry N` too:

https://curl.se/docs/manpage.html#--retry-all-errors

> Retry on any error. This option is used together with [--retry](https://curl.se/docs/manpage.html#--retry).

kou (approved):
+1

We may want to reduce the default concurrency in apache/arrow-adbc too:

```diff
diff --git a/dev/release/verify-release-candidate.sh b/dev/release/verify-release-candidate.sh
index ab13e99..afd033a 100755
--- a/dev/release/verify-release-candidate.sh
+++ b/dev/release/verify-release-candidate.sh
@@ -609,6 +609,7 @@ test_binary_distribution() {
     ${PYTHON:-python3} "$ARROW_SOURCE_DIR/dev/release/download_rc_binaries.py" \
                        $VERSION $RC_NUMBER \
                        --dest="${BINARY_DIR}" \
+                       --num_parallel 1 \
                        --package_type=github \
                        --repository="${SOURCE_REPOSITORY}" \
                        --tag="apache-arrow-adbc-${VERSION}-rc${RC_NUMBER}"
```

We may also want to use GitHub token provided by GitHub Actions on GitHub Actions to avoid rate limit.

dev/release/download_rc_binaries.py
lidavidm: Good catch!

lidavidm: Oh wait, the `_download_url` implementation always passes `--retry 5`

https://github.com/apache/arrow/blob/8e9165e83eac7acd0137ef40f397dc05cfacd872/dev/release/download_rc_binaries.py#L115-L116

lidavidm:
https://github.com/apache/arrow/issues/15091 to use the token, though I'll table this for now since the URL is different to go through the API

PR #15097, opened 2022-12-27
GH-15096: [C++] Substrait ProjectRel Emit Optimization

### LINKED ISSUE #15096: [C++]: Substrait ProjectRel Emit Optimization

At the moment, we are using nested `ProjectNode`s to facilitate this. The first one is the projection in the Substrait plan and the other projection is the one to select the emit fields. But this could be further optimized by just using a single `ProjectNode` by only using the expressions which need to be in the output. 

### Component(s)

C++

vibhatha:
This PR introduces an optimization to remove the additional `ProjectNode` used in handling emit feature for `ProjectRel`.

* Closes: #15096

### DIFF
--- a/cpp/src/arrow/engine/substrait/serde_test.cc
+++ b/cpp/src/arrow/engine/substrait/serde_test.cc
@@ -216,6 +216,20 @@ void CheckRoundTripResult(const std::shared_ptr<Table> expected_table,
   compute::AssertTablesEqualIgnoringOrder(merged_expected, output_table);
 }
 
+void countProjectNodeOptionsInDeclarations(const compute::Declaration& input,
+                                           int& counter) {
+  const auto& options = input.options;
+  const auto& inputs = input.inputs;
+  const auto* maybe_project_option =
+      dynamic_cast<const compute::ProjectNodeOptions*>(options.get());
+  if (maybe_project_option != NULL) {
+    counter++;
+  }
+  for (const auto& in : inputs) {
+    countProjectNodeOptionsInDeclarations(std::get<compute::Declaration>(in), counter);
+  }
+}
+
 TEST(Substrait, SupportedTypes) {
   auto ExpectEq = [](std::string_view json, std::shared_ptr<DataType> expected_type) {
     ARROW_SCOPED_TRACE(json);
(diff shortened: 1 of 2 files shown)


vibhatha:
cc @westonpace what do you think about this change?

westonpace (changes requested):
This makes sense to me but needs a unit test.  Preferably one where the emit references both newly added fields (e.g. fields added by the project) and incoming fields (e.g. fields in the input to the project).

vibhatha:
@westonpace I added a test case and seems like test cases are good. But there are two CIs failing but I am not sure if they are related.

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -2551,6 +2551,122 @@ TEST(SubstraitRoundTrip, ProjectRelOnFunctionWithEmit) {
...
+                "struct": {
+                "types": [{
+                  "i32": {}
+                }, {
+                  "i32": {}
+                }]
+              }
+            },
+            "namedTable": {
+              "names": ["A"]
```
westonpace, line 2625:
```suggestion
              "names": ["TABLE"]
```

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -2551,6 +2551,122 @@ TEST(SubstraitRoundTrip, ProjectRelOnFunctionWithEmit) {
...
+      [5, 5, true],
+      [2, 2, true]
+  ])"});
+  NamedTableProvider table_provider = AlwaysProvideSameTable(std::move(input_table));
+
+  ConversionOptions conversion_options;
+  conversion_options.named_table_provider = std::move(table_provider);
+
+  CheckRoundTripResult(std::move(expected_table), buf,
+                       /*include_columns=*/{}, conversion_options);
```
westonpace, line 2667: Does this actually verify that there is only one project node in the created plan?  Wouldn't it be possible to create two project nodes and still pass?

westonpace (changes requested):
I'm not sure the test is quite verifying this optimization.

cpp/src/arrow/engine/substrait/serde_test.cc
vibhatha: This one only validates the final output, let me work on a comprehensive test case.

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -216,6 +216,20 @@ void CheckRoundTripResult(const std::shared_ptr<Table> expected_table,
   compute::AssertTablesEqualIgnoringOrder(merged_expected, output_table);
 }
 
+void countProjectNodeOptionsInDeclarations(const compute::Declaration& input,
```
westonpace, line 219:
```suggestion
void CountProjectNodeOptionsInDeclarations(const compute::Declaration& input,
```

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -216,6 +216,20 @@ void CheckRoundTripResult(const std::shared_ptr<Table> expected_table,
...
+                                           int& counter) {
```
westonpace, line 220: Can you change this to just return an `int` instead of taking an out parameter?

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -216,6 +216,20 @@ void CheckRoundTripResult(const std::shared_ptr<Table> expected_table,
...
+  const auto& options = input.options;
+  const auto& inputs = input.inputs;
+  const auto* maybe_project_option =
+      dynamic_cast<const compute::ProjectNodeOptions*>(options.get());
+  if (maybe_project_option != NULL) {
+    counter++;
+  }
```
westonpace, line 227:
```suggestion
  if (input.factory_name == "project") {
    counter++;
  }
  const auto& inputs = input.inputs;
```

Isn't it easier to just look at the name?

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -216,6 +216,20 @@ void CheckRoundTripResult(const std::shared_ptr<Table> expected_table,
...
+  for (const auto& in : inputs) {
+    countProjectNodeOptionsInDeclarations(std::get<compute::Declaration>(in), counter);
```
westonpace, line 229:
```suggestion
    CountProjectNodeOptionsInDeclarations(std::get<compute::Declaration>(in), counter);
```

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -2648,21 +2744,32 @@ TEST(SubstraitRoundTrip, ProjectRelOnFunctionWithAllEmit) {
...
+  std::shared_ptr<ExtensionIdRegistry> sp_ext_id_reg = MakeExtensionIdRegistry();
+  ExtensionIdRegistry* ext_id_reg = sp_ext_id_reg.get();
+  ExtensionSet ext_set(ext_id_reg);
+  ASSERT_OK_AND_ASSIGN(auto sink_decls, DeserializePlans(
+                                            *buf, [] { return kNullConsumer; },
+                                            ext_id_reg, &ext_set, conversion_options));
+  auto& other_declrs = std::get<compute::Declaration>(sink_decls[0].inputs[0]);
+  int num_projections = 0;
+  countProjectNodeOptionsInDeclarations(other_declrs, num_projections);
+  ASSERT_EQ(num_projections, 2);
```
westonpace, line 2772:
```suggestion
  int num_projections = CountProjectNodeOptionsInDeclarations(other_declrs);
  ASSERT_EQ(num_projections, 2);
```

westonpace (changes requested):

cpp/src/arrow/engine/substrait/serde_test.cc
vibhatha: Right, I missed this. Thanks for noting this one.

cpp/src/arrow/engine/substrait/serde_test.cc
vibhatha: I added a static counter, but I am not confident with this change. Could you please check?

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -216,6 +216,18 @@ void CheckRoundTripResult(const std::shared_ptr<Table> expected_table,
   compute::AssertTablesEqualIgnoringOrder(merged_expected, output_table);
 }
 
+int CountProjectNodeOptionsInDeclarations(const compute::Declaration& input) {
+  static int counter = 0;
```
westonpace, line 220: Why static?

vibhatha: Ah, I see your point, I updated the logic and also add it as a validation where emit is used.

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -216,6 +216,33 @@ void CheckRoundTripResult(const std::shared_ptr<Table> expected_table,
...
+  if (input.factory_name == "project") {
+    counter++;
+  }
+  const auto& inputs = input.inputs;
+  for (const auto& in : inputs) {
+    return counter +
+           CountProjectNodeOptionsInDeclarations(std::get<compute::Declaration>(in));
+  }
+  return counter;
+}
```
westonpace, line 230:
```suggestion
int CountProjectNodeOptionsInDeclarations(const compute::Declaration& input) {
  int counter = 0;
  if (input.factory_name == "project") {
    counter++;
  }
  const auto& inputs = input.inputs;
  for (const auto& in : inputs) {
    counter += CountProjectNodeOptionsInDeclarations(std::get<compute::Declaration>(in));
  }
  return counter;
}
```

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -216,6 +216,33 @@ void CheckRoundTripResult(const std::shared_ptr<Table> expected_table,
...
+/// Validate the number of expected ProjectNodes when an emit is associated with
+/// relations.
+void ValidateEmit(int expected_projections, const std::shared_ptr<Buffer>& buf,
```
westonpace, line 233:
```suggestion
/// Validate the number of expected ProjectNodes
///
/// Project nodes are sometimes added by emit elements and we may want to
/// verify that we are not adding too many
void ValidateNumProjectNodes(int expected_projections, const std::shared_ptr<Buffer>& buf,
```

I think we should be more direct with the naming.  This method counts project nodes whether they come from an emit or not.

cpp/src/arrow/engine/substrait/serde_test.cc
vibhatha: Did the previous code only work because the max number of projections was 2?

vibhatha: Btw, I updated the PR to include this suggestion.

westonpace: For it to fail you would need a node with multiple inputs.  So something like a join where there was a project on the right hand side.

westonpace (approved):
CI failures appear unrelated.  Thanks for the feature!

PR #15099, opened 2022-12-27
MINOR: [Release] Support older versions of CURL in GitHub downloader

wjones127:
The `--retry-all-errors` parameter is available starting in 7.71.0 - June 24 2020, while the latest curl I could install in Ubuntu 20.04 from apt-get is 7.68.0.

### DIFF
--- a/dev/release/download_rc_binaries.py
+++ b/dev/release/download_rc_binaries.py
@@ -133,6 +133,12 @@ class Downloader:
             raise Exception(f"Downloading {url} failed\n"
                             f"stdout: {stdout}\nstderr: {stderr}")
 
+    def _curl_version(self):
+        cmd = ["curl", "--version"]
+        out = subprocess.run(cmd, capture_output=True, check=True).stdout
+        match = re.search(r"curl (\d+)\.(\d+)\.(\d+) ", out.decode())
+        return (int(match.group(1)), int(match.group(2)), int(match.group(3)))
+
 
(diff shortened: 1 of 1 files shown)


dev/release/download_rc_binaries.py
```diff
@@ -203,7 +203,9 @@ def _download_file(self, dest, asset):
                 "--header",
                 "Accept: application/octet-stream",
                 # Also retry 403s
-                "--retry-all-errors",
+                "--retry",
+                "5",
```
kou, line 207: They are redundant because `download_url()` already have them.

dev/release/download_rc_binaries.py
```diff
@@ -203,7 +203,9 @@ def _download_file(self, dest, asset):
...
+                "--fail"
```
kou, line 208: How about retrying without `--retry-all-errors` when `_download_url()` with `--retry-all-errors` is failed?

```python
try:
  self._download_url(..., "--retry-all-errors", ...)
except ...:
  self._download_url(...)
```

Or we can check `curl` version to confirm whether we can use `--retry-all-errors` or not.

wjones127: Sure. I changed to check the version, though not 100% confident my method is very portable. I welcome suggestions on that.

kou (approved):
+1

Let's try this approach!

PR #15101, opened 2022-12-27
GH-15100: [C++][Parquet] Add benchmark for reading strings from Parquet

### LINKED ISSUE #15100: Add benchmarks for reading and writing strings

To evaluate #14353 we should benchmark reading and writing string data.

### Component(s)

Benchmarking, C++, Parquet

wjones127:
* Closes: #15100

### DIFF
--- a/cpp/src/parquet/arrow/reader_writer_benchmark.cc
+++ b/cpp/src/parquet/arrow/reader_writer_benchmark.cc
@@ -202,6 +202,54 @@ BENCHMARK_TEMPLATE2(BM_WriteColumn, true, DoubleType);
 BENCHMARK_TEMPLATE2(BM_WriteColumn, false, BooleanType);
 BENCHMARK_TEMPLATE2(BM_WriteColumn, true, BooleanType);
 
+int32_t kInfiniteUniqueValues = -1;
+
+std::shared_ptr<::arrow::Table> RandomStringTable(int64_t length, int64_t unique_values,
+                                                  int64_t null_percentage) {
+  std::shared_ptr<::arrow::DataType> type = ::arrow::utf8();
+  std::shared_ptr<::arrow::Array> arr;
+  ::arrow::random::RandomArrayGenerator generator(500);
+  double null_probability = static_cast<double>(null_percentage) / 100.0;
+  if (unique_values == kInfiniteUniqueValues) {
+    arr = generator.String(length, /*min_length=*/3, /*max_length=*/32,
+                           /*null_probability=*/null_probability);
+  } else {
+    arr = generator.StringWithRepeats(length, /*unique=*/unique_values,
+                                      /*min_length=*/3, /*max_length=*/32,
+                                      /*null_probability=*/null_probability);
+  }
+  return ::arrow::Table::Make(
+      ::arrow::schema({::arrow::field("column", type, null_percentage > 0)}), {arr});
+}
+
+template <bool nullable>
+static void BM_WriteBinaryColumn(::benchmark::State& state) {
+  std::shared_ptr<::arrow::Table> table =
+      RandomStringTable(BENCHMARK_SIZE, state.range(1), state.range(0));
+
+  while (state.KeepRunning()) {
+    auto output = CreateOutputStream();
+    EXIT_NOT_OK(
+        WriteTable(*table, ::arrow::default_memory_pool(), output, BENCHMARK_SIZE));
+  }
+
+  int64_t total_bytes = table->column(0)->chunk(0)->data()->buffers[1]->size();
+  state.SetItemsProcessed(BENCHMARK_SIZE * state.iterations());
+  state.SetBytesProcessed(total_bytes * state.iterations());
+}
+
+BENCHMARK_TEMPLATE1(BM_WriteBinaryColumn, false)
+    ->Args({/*null_probability*/ 0, /*unique_values*/ 2})
+    ->Args({/*null_probability*/ 0, /*unique_values*/ 32})
+    ->Args({/*null_probability*/ 0, /*unique_values*/ kInfiniteUniqueValues});
+BENCHMARK_TEMPLATE1(BM_WriteBinaryColumn, true)
+    ->Args({/*null_probability*/ 1, /*unique_values*/ 32})
+    ->Args({/*null_probability*/ 50, /*unique_values*/ 32})
+    ->Args({/*null_probability*/ 99, /*unique_values*/ 32})
+    ->Args({/*null_probability*/ 1, /*unique_values*/ kInfiniteUniqueValues})
+    ->Args({/*null_probability*/ 50, /*unique_values*/ kInfiniteUniqueValues})
+    ->Args({/*null_probability*/ 99, /*unique_values*/ kInfiniteUniqueValues});
+
 template <typename T>
 struct Examples {
   static constexpr std::array<T, 2> values() { return {127, 128}; }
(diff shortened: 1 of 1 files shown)


wjones127:
@ursabot please benchmark command=cpp-micro --suite-filter=parquet-arrow-reader-writer-benchmark

wjones127:
@ursabot please benchmark command=cpp-micro --suite-filter=parquet-arrow-reader-writer-benchmark

wjones127:
@ursabot please benchmark

cpp/src/parquet/arrow/reader_writer_benchmark.cc
```diff
@@ -197,6 +197,52 @@ BENCHMARK_TEMPLATE2(BM_WriteColumn, true, DoubleType);
...
+BENCHMARK(BM_WriteBinaryColumn)
+    ->Args({/*null_probability*/ 0, /*unique_values*/ 2})
+    ->Args({/*null_probability*/ 0, /*unique_values*/ 32})
+    ->Args({/*null_probability*/ 0, /*unique_values*/ kInfiniteUniqueValues})
+    ->Args({/*null_probability*/ 1, /*unique_values*/ 32})
+    ->Args({/*null_probability*/ 50, /*unique_values*/ 32})
+    ->Args({/*null_probability*/ 99, /*unique_values*/ 32})
+    ->Args({/*null_probability*/ 1, /*unique_values*/ kInfiniteUniqueValues})
+    ->Args({/*null_probability*/ 50, /*unique_values*/ kInfiniteUniqueValues})
+    ->Args({/*null_probability*/ 99, /*unique_values*/ kInfiniteUniqueValues});
```
pitrou, line 244: Can you use `ArgNames` to make the arguments more descriptive in the output?

cpp/src/parquet/arrow/reader_writer_benchmark.cc
```diff
@@ -197,6 +197,52 @@ BENCHMARK_TEMPLATE2(BM_WriteColumn, true, DoubleType);
...
+  std::shared_ptr<::arrow::Array> arr;
+  ::arrow::random::RandomArrayGenerator generator(500);
+  double null_probability = static_cast<double>(null_percentage) / 100.0;
+  if (unique_values == kInfiniteUniqueValues) {
+    arr = generator.String(length, /*min_length=*/3, /*max_length=*/32,
+                           /*null_probability=*/null_probability);
+  } else {
+    arr = generator.StringWithRepeats(length, /*unique=*/unique_values,
+                                      /*min_length=*/3, /*max_length=*/32,
+                                      /*null_probability=*/null_probability);
```
pitrou, line 214: Do we expect duplicate values to show different performance characteristics? Otherwise, the `unique_values` argument may not be very useful.

cpp/src/parquet/arrow/reader_writer_benchmark.cc
```diff
@@ -197,6 +197,52 @@ BENCHMARK_TEMPLATE2(BM_WriteColumn, true, DoubleType);
 BENCHMARK_TEMPLATE2(BM_WriteColumn, false, BooleanType);
 BENCHMARK_TEMPLATE2(BM_WriteColumn, true, BooleanType);
 
+int32_t kInfiniteUniqueValues = -1;
+
+std::shared_ptr<::arrow::Table> RandomStringTable(int64_t length, int64_t unique_values,
+                                                  int64_t null_percentage) {
+  std::shared_ptr<::arrow::DataType> type = ::arrow::utf8();
+  std::shared_ptr<::arrow::Array> arr;
+  ::arrow::random::RandomArrayGenerator generator(500);
```
pitrou, line 206: Is this the seed?

cpp/src/parquet/arrow/reader_writer_benchmark.cc
```diff
@@ -197,6 +197,52 @@ BENCHMARK_TEMPLATE2(BM_WriteColumn, true, DoubleType);
...
+  } else {
+    arr = generator.StringWithRepeats(length, /*unique=*/unique_values,
+                                      /*min_length=*/3, /*max_length=*/32,
+                                      /*null_probability=*/null_probability);
+  }
+  return ::arrow::Table::Make(
+      ::arrow::schema({::arrow::field("column", type, null_percentage > 0)}), {arr});
+}
+
+static void BM_WriteBinaryColumn(::benchmark::State& state) {
```
pitrou, line 220: Does it use the PLAIN encoding? Add a comment?

cpp/src/parquet/arrow/reader_writer_benchmark.cc
```diff
@@ -319,6 +365,29 @@ BENCHMARK_TEMPLATE2(BM_ReadColumn, true, BooleanType)
...
+BENCHMARK(BM_ReadBinaryColumn)
+    ->Args({/*null_probability*/ 0, /*unique_values*/ 2})
+    ->Args({/*null_probability*/ 0, /*unique_values*/ 32})
+    ->Args({/*null_probability*/ 0, /*unique_values*/ kInfiniteUniqueValues})
+    ->Args({/*null_probability*/ 1, /*unique_values*/ 32})
+    ->Args({/*null_probability*/ 50, /*unique_values*/ 32})
+    ->Args({/*null_probability*/ 99, /*unique_values*/ 32})
+    ->Args({/*null_probability*/ 1, /*unique_values*/ kInfiniteUniqueValues})
+    ->Args({/*null_probability*/ 50, /*unique_values*/ kInfiniteUniqueValues})
+    ->Args({/*null_probability*/ 99, /*unique_values*/ kInfiniteUniqueValues});
```
pitrou, line 389: Similar questions and comments here.

cpp/src/parquet/arrow/reader_writer_benchmark.cc
```diff
@@ -197,6 +197,52 @@ BENCHMARK_TEMPLATE2(BM_WriteColumn, true, DoubleType);
...
+  std::shared_ptr<::arrow::Table> table =
+      RandomStringTable(BENCHMARK_SIZE, state.range(1), state.range(0));
+
+  while (state.KeepRunning()) {
+    auto output = CreateOutputStream();
+    EXIT_NOT_OK(
+        WriteTable(*table, ::arrow::default_memory_pool(), output, BENCHMARK_SIZE));
+  }
+
+  int64_t total_bytes = table->column(0)->chunk(0)->data()->buffers[1]->size();
```
pitrou, line 230: This is the offsets buffer but doesn't take into account the data buffer.

cpp/src/parquet/arrow/reader_writer_benchmark.cc
wjones127: I was thinking it would depending on encoding (RLE vs PLAIN), and IIUC we automatically switch from RLE to plain when we hit a certain threshold of unique values. Though perhaps it's sufficient to just benchmark 32 unique values and infinite, rather than several different cardinalities. I will also make sure to enforce the encoding.

cpp/src/parquet/arrow/reader_writer_benchmark.cc
wjones127: Yeah, the same seed is hardcoded elsewhere in the file. Is there somewhere we should be taking the seed from?

pitrou: No, it's ok. I would just suggest naming the parameter, e.g.:
```suggestion
  ::arrow::random::RandomArrayGenerator generator(/*seed=*/ 500);
```

cpp/src/parquet/arrow/reader_writer_benchmark.cc
wjones127: Good idea. If I do the same to the other benchmarks in the file, would that break the continuity of their results in conbench?

cpp/src/parquet/arrow/reader_writer_benchmark.cc
wjones127: Thanks. I thought I was just doing the data buffer, not the offsets. But probably a good idea to include both.

cpp/src/parquet/arrow/reader_writer_benchmark.cc
wjones127: I added a comment near the parameters of each benchmark, explaining we are using the `unique_values` to trigger the code paths for dictionary and plain encodings. I tried to add a test within the benchmark to validate we are getting the expected encodings. But I found that it was too complicated, as the encodings can change from page to page and also apply to the definition and repetition levels (IIUC).

cpp/src/parquet/arrow/reader_writer_benchmark.cc
pitrou: I'm not sure, though I think we can live with that in any case.

cpp/src/parquet/arrow/reader_writer_benchmark.cc
pitrou: I see. Can you just confirm that the expected encodings are used (and add a comment)?

pitrou: Just saw the comment below, sorry. Please disregard. :-)

pitrou (approved):

PR #15104, opened 2022-12-28
ARROW-15732: [C++] Do not use any CPU threads in execution plan when use_threads is false

westonpace:
This PR gets rid of the "executor==nullptr" style of execution from Acero.  Now a CPU executor must always be defined.  This also updates the DeclarationToXyz methods to support 5 "standard" modes of operation:

* sync / threaded - All CPU work is done on the CPU thread pool.  The calling thread sleeps until the plan is done.
* sync / serial - All CPU work is done on the calling thread.  The calling thread becomes a serial executor until plan is done.
* async / threaded - All CPU work is done on the CPU thread pool.  The calling thread returns immediately.
* async / serial - All CPU work is done on a plan-specific 1-thread pool.  The calling thread returns immediately.
* async / custom exec context - All CPU work is done on the provided executor which cannot be nullptr and the caller is responsible for keeping it alive.

For backwards compatibility the "old style" of serial execution (which is still used by R/python) will still be accepted.  Under the hood a 1-thread pool will be created and its lifetime will be bound to the exec plan (this is similar to async / serial when using DeclarationToXyz).

Since DeclarationToXyz doesn't support "ordered sinks" we cannot yet migrate the bindings and so they will continue to rely on the deprecated API.

All unit tests are updated to use the new API.  Some unit tests were not actually running in parallel (even though they said they were) and now that they are running in parallel a few changes had to be made to accept / handle non-deterministic result order.

As part of this process some unit tests were migrated away from StartAndCollect in favor of DeclarationToXyz.  Hopefully we can eventually migrate almost all of the unit tests to DeclarationToXyz but that is a longer term effort.

### DIFF
--- a/cpp/src/arrow/compute/exec/exec_plan.cc
+++ b/cpp/src/arrow/compute/exec/exec_plan.cc
@@ -17,6 +17,7 @@
 
 #include "arrow/compute/exec/exec_plan.h"
 
+#include <atomic>
 #include <optional>
 #include <sstream>
 #include <unordered_map>
@@ -43,6 +44,7 @@
 namespace arrow {
 
 using internal::checked_cast;
+using internal::ThreadPool;
 using internal::ToChars;
 
 namespace compute {
@@ -50,9 +52,12 @@ namespace compute {
 namespace {
 
 struct ExecPlanImpl : public ExecPlan {
-  explicit ExecPlanImpl(QueryOptions options, ExecContext* exec_context,
-                        std::shared_ptr<const KeyValueMetadata> metadata = NULLPTR)
-      : metadata_(std::move(metadata)), query_context_(options, *exec_context) {}
+  explicit ExecPlanImpl(QueryOptions options, ExecContext exec_context,
+                        std::shared_ptr<const KeyValueMetadata> metadata = NULLPTR,
+                        std::shared_ptr<ThreadPool> owned_thread_pool = nullptr)
+      : metadata_(std::move(metadata)),
+        query_context_(options, exec_context),
+        owned_thread_pool_(std::move(owned_thread_pool)) {}
 
   ~ExecPlanImpl() override {
     if (started_ && !finished_.is_finished()) {
@@ -90,6 +95,15 @@ struct ExecPlanImpl : public ExecPlan {
     if (started_) {
       return Status::Invalid("restarted ExecPlan");
     }
+    if (query_context_.exec_context()->executor() == nullptr) {
+      return Status::Invalid(
+          "An exec plan must have an executor for CPU tasks.  To run without threads use "
+          "a SerialExeuctor (the arrow::compute::DeclarationTo... methods should take "
+          "care of this for you and are an easier way to execute an ExecPlan.");
+    }
+    if (query_context_.io_context()->executor() == nullptr) {
+      return Status::Invalid("An exec plan must have an I/O executor for I/O tasks.");
+    }
 
     started_ = true;
 
@@ -99,8 +113,8 @@ struct ExecPlanImpl : public ExecPlan {
     // If no source node schedules any tasks (e.g. they do all their word synchronously as
     // part of StartProducing) then the plan may be finished before we return from this
     // call.
-    Future<> scheduler_finished =
-        util::AsyncTaskScheduler::Make([this](util::AsyncTaskScheduler* async_scheduler) {
+    Future<> scheduler_finished = util::AsyncTaskScheduler::Make(
+        [this](util::AsyncTaskScheduler* async_scheduler) {
           QueryContext* ctx = query_context();
           RETURN_NOT_OK(ctx->Init(ctx->max_concurrency(), async_scheduler));
 
@@ -120,6 +134,8 @@ struct ExecPlanImpl : public ExecPlan {
           // away soon (or at least be replaced by a sub-scheduler to facilitate OT)
           for (auto& n : nodes_) {
             RETURN_NOT_OK(n->Init());
+          }
+          for (auto& n : nodes_) {
             async_scheduler->AddSimpleTask([&] { return n->finished(); });
           }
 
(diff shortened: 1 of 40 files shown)


cpp/src/arrow/compute/exec/exec_plan.h
```diff
@@ -419,37 +429,116 @@ struct ARROW_EXPORT Declaration {
...
 
 /// \brief Asynchronous version of \see DeclarationToTable
+///
+/// The behavior of use_threads is slightly different than the synchronous version since
+/// we cannot run synchronously on the calling thread.  Instead, if use_threads=false then
+/// a new thread pool will be created with a single thread and this will be used for all
+/// compute work.
+///
+/// If a custom exec context is provided then the value of `use_threads` will be ignored
+/// and the executor in the custom context will be used.  However, the executor must
```
lidavidm, line 447: It doesn't seem to be possible to pass both an exec context and use_threads anymore.

cpp/src/arrow/compute/exec/exec_plan.cc
```diff
@@ -43,16 +44,20 @@
...
 
 namespace {
 
 struct ExecPlanImpl : public ExecPlan {
-  explicit ExecPlanImpl(QueryOptions options, ExecContext* exec_context,
-                        std::shared_ptr<const KeyValueMetadata> metadata = NULLPTR)
-      : metadata_(std::move(metadata)), query_context_(options, *exec_context) {}
+  explicit ExecPlanImpl(QueryOptions options, ExecContext exec_context,
+                        std::shared_ptr<const KeyValueMetadata> metadata = NULLPTR,
+                        std::shared_ptr<ThreadPool> owned_thread_pool = nullptr)
```
lidavidm, line 57: nit: be consistent with NULLPTR/nullptr (note that NULLPTR is really only necessary in headers, not in source files)

cpp/src/arrow/compute/exec/exec_plan.cc
```diff
@@ -90,6 +95,15 @@ struct ExecPlanImpl : public ExecPlan {
     if (started_) {
       return Status::Invalid("restarted ExecPlan");
     }
+    if (query_context_.exec_context()->executor() == nullptr) {
+      return Status::Invalid(
+          "An exec plan must have an executor for CPU tasks.  To run without threads use "
+          "a SerialExeuctor (the arrow::compute::DeclarationTo... methods should take "
+          "care of this for you and are an easier way to execute an ExecPlan.");
```
lidavidm, line 102: nit: unclosed parenthesis :)

cpp/src/arrow/compute/exec/sink_node.cc
```diff
@@ -381,19 +388,11 @@ class ConsumingSinkNode : public ExecNode, public BackpressureControl {
...
-          [this, finish_st](const Status& st) {
-            // Prefer the plan error over the consumer error
-            Status final_status = finish_st & st;
-            finished_.MarkFinished(final_status);
-            return final_status;
-          });
-    });
+    if (finish_st.ok()) {
+      plan_->query_context()->async_scheduler()->AddSimpleTask(
+          [this] { return consumer_->Finish(); });
```
lidavidm, line 393: Shouldn't we still mark finished_ as finished?

lidavidm (commented):
LGTM overall. Thanks for cleaning up the tests.

We do have a lot of different modes of execution, though I suppose it's just because of sync/async × threaded/serial.

lidavidm (approved):

cpp/src/arrow/util/async_util.cc
```diff
@@ -188,8 +192,20 @@ class AsyncTaskSchedulerImpl : public AsyncTaskScheduler {
...
+      // will prevent any other task finishing and marking the scheduler finished
+      // while we are running the abort callback.
+      running_tasks_++;
+      aborted = true;
+    }
+    if (aborted) {
+      lk.unlock();
+      std::move(abort_callback_)(st);
+      lk.lock();
+      running_tasks_--;
```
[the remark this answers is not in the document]
westonpace: Running callbacks while holding locks is problematic.  As an example, the abort callback was added here so we can call StopProducing in the exec plan when a task fails.  If we do this while holding the lock then we have to be careful, in every node, that we don't use the async task scheduler in any way during the call to StopProducing.  Admittedly, that should probably be obvious (if we are aborting then scheduling new tasks is meaningless).  However, this is just more overhead for node authors to worry about and I've been trying as best I can to keep the scheduler fool-proof.

So I run the callback outside of the lock.  That was a bit of a pain because I need to make sure the scheduler doesn't end itself, and get destroyed, while the abort callback is running.  For that reason I treat the abort callback as a "pseudo-task" by incrementing then decrementing running_tasks_.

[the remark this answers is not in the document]
lidavidm: I read it backwards :grimacing: thought it was lock then unlock, not the other way around.

cpp/src/arrow/compute/exec/exec_plan.h
westonpace: Ah, good catch.  I went through several iterations of the API and this comment only made sense somewhere in the middle.

cpp/src/arrow/compute/exec/exec_plan.cc
westonpace: Thanks, fixed!

cpp/src/arrow/compute/exec/exec_plan.cc
westonpace: Fixed

westonpace:
Thanks for the quick review.

> We do have a lot of different modes of execution, though I suppose it's just because of sync/async × threaded/serial.

Yep.  The async-serial mode is newly introduced here and it would be nice to avoid it.  However, I can't easily switch R/python over to sync-serial because I need to wait until we fix ordering (or add ordering to the DeclarationToXyz methods which I don't want to do).

Fixing ordering in this PR would make the PR too large so I think I'm stuck with this.

PR #15108, opened 2022-12-29
ARROW-17823: [C++] [Parquet] Prefer std::make_shared/std::make_unique over constructor with new in Parquet

mapleFU:
This patch reuse the arrow issue https://issues.apache.org/jira/browse/ARROW-17823 .

Advantage: readabilty, exception safety and efficiency(only for shared_ptr).

Cases that don't apply: When calling a private/protected constructor within class member function, make_shared/unique can't work.

### DIFF
--- a/cpp/src/parquet/arrow/reader.cc
+++ b/cpp/src/parquet/arrow/reader.cc
@@ -431,8 +431,7 @@ class RowGroupReaderImpl : public RowGroupReader {
       : impl_(impl), row_group_index_(row_group_index) {}
 
   std::shared_ptr<ColumnChunkReader> Column(int column_index) override {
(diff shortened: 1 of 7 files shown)


wjones127 (commented):
Thanks for doing this.

Isn't this one a candidate as well?

https://github.com/apache/arrow/blob/80b1e0e755a748112e1daba3c9234276c4193d42/cpp/src/parquet/arrow/reader.cc#L434-L435

mapleFU:
Nice catch! Too much code here and I miss `parquet/arrow` part. PTAL again @wjones127

wjones127 (approved):

wjones127:
Thanks @mapleFU!

PR #14235, opened 2022-09-26
ARROW-17692: [R] Add support for building with system AWS SDK C++

thisisnic:
This PR uses "pkg-config --static ... arrow" to collect build flags. "pkg-config --static ... arrow" reports suitable build flags that depend on build options and used libraries for Apache Arrow C++. This works with the system AWS SDK C++.

PR #15111, opened 2022-12-29
GH-15110: [R] [CI] Windows build fails in packaging job

### LINKED ISSUE #15110: [R] [CI] Windows build fails in packaging job

One example: https://github.com/ursacomputing/crossbow/actions/runs/3799348531/jobs/6462070137

```
C:\rtools42\x86_64-w64-mingw32.static.posix\bin/ld.exe: ../windows/arrow-10.0.1.100000336/lib/x64-ucrt/libcurl.a(libcurl_la-version.o):(.text+0x1bc): undefined reference to `nghttp2_version'
collect2.exe: error: ld returned 1 exit status
no DLL was created
ERROR: compilation failed for package 'arrow'
* removing 'D:/a/_temp/Library/arrow'
Warning message:
In install.packages("arrow", type = "source", repos = sub("file://",  :
  installation of package 'arrow' had non-zero exit status
```

I suspect this has to do with #14235 (and I'll try to fix it in a PR in a sec)

### Component(s)

Continuous Integration, R

jonkeane:
Add new dependencies to windows builds following #14235
* Closes: #15110

### DIFF
--- a/ci/scripts/r_windows_build.sh
+++ b/ci/scripts/r_windows_build.sh
@@ -71,7 +71,7 @@ if [ -d mingw64/lib/ ]; then
   # Move the 64-bit versions of libarrow into the expected location
   mv mingw64/lib/*.a $DST_DIR/lib/x64
   # These are from https://dl.bintray.com/rtools/mingw{32,64}/
-  cp $MSYS_LIB_DIR/mingw64/lib/lib{thrift,snappy,zstd,lz4,brotli*,bz2,crypto,curl,ss*,utf8proc,re2,aws*}.a $DST_DIR/lib/x64
+  cp $MSYS_LIB_DIR/mingw64/lib/lib{thrift,snappy,zstd,lz4,brotli*,bz2,crypto,curl,ss*,utf8proc,re2,aws*,nghttp2}.a $DST_DIR/lib/x64
 fi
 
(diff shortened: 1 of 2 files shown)


jonkeane:

jonkeane:

wjones127 (approved):

assignUser:
This seems to be the correct fix for the issue but I don't think #14235 or the follow up #15075 are the cause (even if they would be my first guess too)  as the R win  4.2 checks were green as recent as a0d16306229fc08f9dc64361f9459806e02b5932 and only started failing with aac30873fd9b112d236a7359fcc8a60f04ebfead (on merge, within the PR the check was green...) the rtools version used is unchanged  and so is the runner image, so I am unsure what could have caused the change?

I'll test this for HEAD and this PR locally.

assignUser (approved):
ah wow, just saw your investigation in #15110 @jonkeane I agree it likely is the updated  curl version, I had a look at the changelog and there were changes related to nghttp2 but nothing I could immediatly identify as cause of the issue. It still makes sense as the cause!

jonkeane:
Ah yes, sorry I should have updated the description here with that info too. Agreed that that PR is likely not the root cause, but something upstream in mingw dependency land

Issue #15115, opened 2022-12-29
[R][CI] pyarrow tests fail on macos 10.13 due to missing pyarrow wheel


assignUser:
### Describe the bug, including details regarding any error messages, version, and platform.

see https://github.com/apache/arrow/issues/14829#issuecomment-1367506270

### Component(s)

Continuous Integration, R

PR #14582, opened 2022-11-03
ARROW-18240: [R] head() is crashing on some nightly builds

paleolimbot:
Trying to address these two failures:

- https://github.com/ursacomputing/crossbow/actions/runs/3368026889/jobs/5586109693#step:10:3813
- https://github.com/ursacomputing/crossbow/actions/runs/3368024633/jobs/5586105172#step:9:3813

PR #15116, opened 2022-12-29
GH-15114: [R][C++][CI] Homebrew can't install Python 3.11 on GHA runners

### LINKED ISSUE #15114: [R][C++][CI] Homebrew can't install 3.11 on GHA runners

The GHA runners have a non-brew python 3.11 installed that blocks install of the brew 3.11:
```
The formula built, but is not symlinked into /usr/local
Could not symlink bin/2to3-3.11
Target /usr/local/bin/2to3-3.11
already exists. You may want to remove it:
  rm '/usr/local/bin/2to3-3.11'
```
[log](https://github.com/ursacomputing/crossbow/actions/runs/3799325237/jobs/6461740362#step:4:459)

### Component(s)

R, Continuous Integration

assignUser:
* Closes: #15114

### DIFF
--- a/dev/tasks/homebrew-formulae/github.macos.yml
+++ b/dev/tasks/homebrew-formulae/github.macos.yml
@@ -29,6 +29,10 @@ jobs:
       {{ macros.configure_homebrew_arrow(formula)|indent }}
       - name: Test formula
         run: |
+          # The github runners have a non-brew 3.11 installed and the existing symlinks cause
+          # the installation of brew 3.11 to fail
+          find /usr/local/bin -lname '*/Library/Frameworks/Python.framework/*' -delete
+          
           brew install -v --HEAD apache-arrow
           brew test apache-arrow
           brew audit --strict apache-arrow
--- a/dev/tasks/homebrew-formulae/github.macos.yml
+++ b/dev/tasks/homebrew-formulae/github.macos.yml
@@ -29,15 +29,7 @@ jobs:
       {{ macros.configure_homebrew_arrow(formula)|indent }}
       - name: Test formula
         run: |
-          # The github runners have a non-brew Python 3.11 installed and the existing symlinks cause
-          # the installation of brew python@3.11 to fail
-          rm '/usr/local/bin/2to3-3.11' || echo 'Nothing to delete'
-          rm '/usr/local/bin/2to3' || echo 'Nothing to delete'
-          rm '/usr/local/bin/idle3.11' || echo 'Nothing to delete'
-          rm '/usr/local/bin/pydoc3.11' || echo 'Nothing to delete'
-          rm '/usr/local/bin/python3.11' || echo 'Nothing to delete'
-          rm '/usr/local/bin/python3.11-config' || echo 'Nothing to delete'
-
+          {{ macros.github_cleanup_macos_python()|indent(10) }}
           brew install -v --HEAD apache-arrow
           brew test apache-arrow
           brew audit --strict apache-arrow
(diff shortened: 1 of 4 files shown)


assignUser:
@github-actions crossbow submit homebrew-cpp homebrew-r-*

assignUser:
It does not look like it but the fix worked xD
the macos 11 jobs are still failing because of the `head()` issue (#14582 ) and the 10.13 jobs are now failing because of the missing pyarrow wheels (#15115)

dev/tasks/homebrew-formulae/github.macos.yml
```diff
@@ -29,6 +29,10 @@ jobs:
       {{ macros.configure_homebrew_arrow(formula)|indent }}
       - name: Test formula
         run: |
+          # The github runners have a non-brew 3.11 installed and the existing symlinks cause
+          # the installation of brew 3.11 to fail
+          find /usr/local/bin -lname '*/Library/Frameworks/Python.framework/*' -delete
```
jonkeane, line 34: IMHO, this is a pretty expansive glob that could impact a number of different files — any way that we could make it be a bit more specific to the issue we're seeing? Either (or both of): looking for 3.11 specifically, only running this on (specific) github runners? Or maybe even better: `brew link --overwrite python@3.11` which is suggested in the message?

jonkeane (commented):
One substantive comment. 

On the crossbow builds: 

It looks like the homebrew-r-autobrew jobs are all still failing: the 10.13 ones should be resolved by #15117 yeah? (and then the 11 ones by the head issue).

But what's up with the homebrew-r-brew failure? Is that expected? It looks like it's the same failure we were seeing for homebrew-cpp, no?

dev/tasks/homebrew-formulae/github.macos.yml
assignUser: Yeah the idea was to not have to add multiple things manually after retrying a bunch but being eplicit is probably for the best (e.g. `rm '/usr/local/bin/2to3-3.11'`). 

This is only done on the affected runners (non-10.13), using `overwrite` would require us to change it in the brew internals which would affect everyone not just these (wrongly configured) CI runners.

I'll switch this to the explicit removal, that will probably also fix the brew job for which the glob apparently failed...

jonkeane (approved):
I'm still -0.1 on manually removing these files versus trying to use the brew-specific recommendation here (mostly because it looks like this might become a non-trivial maintenance burden for us in the future), but happy to merge this now so we can stem the failures as much as possible.

assignUser:
After some investigating it looks like this might be a brew issue related to brew update, as this only happens after that. ([reprex](https://github.com/assignUser/test-repo-a/actions/runs/3823464060/jobs/6504662910#step:2:18)) I'll open an issue over in homebrew. 

We can try installing python manually with overwrite instead of deleting the things. I will try this in a moment

assignUser:
@github-actions crossbwo submit hombrew-cpp homebrew-r-*

assignUser:
@github-actions crossbow submit homebrew-cpp homebrew-r-*

re-trigger after rebase

assignUser:
> I just merged the `head()` PR fix (#14582) if you want to try rebasing on that!

oh yeah :D

assignUser:
@github-actions crossbow submit homebrew-cpp homebrew-r-brew

dev/tasks/macros.jinja
```diff
@@ -235,6 +235,7 @@ on:
       brew unlink python@2 || true
       brew config
       brew doctor || true
+      brew install --overwrite python@3.11 python@3.10
```
jonkeane, line 238: You mentioned in the comments submitting an issue to homebrew — could we link to that here in a comment? Just so we know _why_ we need the overwriting situation?

jonkeane (approved):
😱 I thought this approach would be simpler — but I didn't think it would be _this much_ simpler. 

One small comment asking for a comment, but happy to merge this otherwise.

PR #15117, opened 2022-12-29
GH-15115: [R][CI] pyarrow tests fail on macos 10.13 due to missing pyarrow wheel

### LINKED ISSUE #15115: [R][CI] pyarrow tests fail on macos 10.13 due to missing pyarrow wheel

assignUser:
* Closes: #15115

### DIFF
--- a/r/tests/testthat/test-python.R
+++ b/r/tests/testthat/test-python.R
@@ -24,6 +24,8 @@ test_that("install_pyarrow", {
   skip_if_not_installed("reticulate")
   # PyArrow doesn't support Python 3.6 or earlier
   skip_on_python_older_than("3.7")
+  # no pyarrow wheels for macos 10.13
+  skip_if(Sys.info()["sysname"] == "Darwin" && Sys.info()["release"] < 14)
 
   venv <- try(reticulate::virtualenv_create("arrow-test"))
   # Bail out if virtualenv isn't available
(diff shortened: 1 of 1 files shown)


assignUser:
@github-actions crossbow submit homebrew-r-autobrew

r/tests/testthat/test-python.R
```diff
@@ -24,6 +24,8 @@ test_that("install_pyarrow", {
   skip_if_not_installed("reticulate")
   # PyArrow doesn't support Python 3.6 or earlier
   skip_on_python_older_than("3.7")
+  # no pyarrow wheels for macos 10.13
+  skip_if(Sys.info()["sysname"] == "Darwin" && Sys.info()["release"] < 14)
```
jonkeane, line 28: IIUC the release here should actually be 18 (cf https://en.wikipedia.org/wiki/Darwin_(operating_system))

jonkeane (changes requested):
I suspect this will still fail, but that might be the head issue we were seeing in the other PR.

r/tests/testthat/test-python.R
assignUser: oh wow that versioning scheme is confusing :D

assignUser:
@github-actions crossbow submit homebrew-r-autobrew

and yes the head() issue will probably still pop up but the pyarrow tests should all be skipped like this (the first explicitly and the others due to missing pyarrow)

jonkeane (approved):

PR #15121, opened 2022-12-29
MINOR: [Python] add generated lib.h file to pyarrow gitignore

westonpace:


westonpace:
Not sure what changed but I've been seeing this file pop up unexpectedly for a while

kou (approved):
+1

Good catch! This was generated since #14925.

PR #10627, opened 2021-06-30
PARQUET-490: [C++][Parquet] Basic support for reading DELTA_BINARY_PACKED data

shanhuuang:
Only works for int32/int64 when delta_bit_width_<=32.

1. Update the implement of DeltaBitPackDecoder
2. Revise the code of putting/getting ZigZag int and the unit test
3. Add the implement of putting/getting ZigZag int64 and unit test

PR #15124, opened 2022-12-30
GH-15052: [C++][Parquet] Fix DELTA_BINARY_PACKED decoder when reading only one value

### LINKED ISSUE #15052: [C++][Parquet] Getting only 0 when reading DELTA_BINARY_PACKED

I have encountered a possible bug where when reading values one at a time with the low level API of parquet reader
(as done similarly in the example: [/cpp/examples/parquet/low_level_api/reader_writer.cc](https://github.com/apache/arrow/blob/d18412e07f5aad968a73e1f44bcfa0c63461211f/cpp/examples/parquet/low_level_api/reader_writer.cc))
in a row with DELTA_BINARY_PACKED encoding the results are all 0 regardless of file content.

aka this `rows_read = int64_reader->ReadBatch(1, &definition_level, &repetition_level,
                                                    &value, &values_read);` gives a zero value
but this `rows_read = int64_reader->ReadBatch(2, nullptr, nullptr, values, &values_read);` gives the correct value.

So this seems not to occur when reading bigger batch-sizes (>1).

The problem might be somewhere within `DeltaBitPackDecoder<DType>::GetInternal`

### Component(s)

Parquet

mapleFU:
This patch trying to fix https://github.com/apache/arrow/issues/15052 . The problem is mentioned here: https://github.com/apache/arrow/issues/15052#issuecomment-1367486164

When read 1 value, DeltaBitPackDecoder will not call `InitBlock`, causing it always read `last_value_`.

Seems the problem is introduced in https://github.com/apache/arrow/pull/10627 and https://github.com/amol-/arrow/commit/d982bedcf5e03d44c01949b192da54a8c1e525d8

I will add some test tonight
* Closes: #15052


cpp/src/parquet/encoding.cc
```diff
@@ -2479,7 +2481,12 @@ class DeltaBitPackDecoder : public DecoderImpl, virtual public TypedDecoder<DTyp
       if (ARROW_PREDICT_FALSE(values_current_mini_block_ == 0)) {
         if (ARROW_PREDICT_FALSE(!block_initialized_)) {
           buffer[i++] = last_value_;
-          if (ARROW_PREDICT_FALSE(i == max_values)) break;
+          if (ARROW_PREDICT_FALSE(i == max_values)) {
+            if (i != static_cast<int>(total_value_count_)) {
+              InitBlock();
+            }
+            break;
+          }
```
rok, line 2489: I find this difficult to read, could you add a comment about what this logic is doing?

cpp/src/parquet/encoding.cc
```diff
@@ -2459,7 +2459,9 @@ class DeltaBitPackDecoder : public DecoderImpl, virtual public TypedDecoder<DTyp
         ParquetException::EofException();
       }
       if (bit_width_data[i] > kMaxDeltaBitWidth) {
-        throw ParquetException("delta bit width larger than integer bit width");
+        throw ParquetException("delta bit width " + std::to_string(bit_width_data[i]) +
+                               " larger than integer bit width " +
+                               std::to_string(kMaxDeltaBitWidth));
```
rok, line 2464: Nice!

rok (approved):
Thanks for doing this @mapleFU! This is a valuable fix.

cpp/src/parquet/encoding_test.cc
```diff
@@ -1413,5 +1436,16 @@ TYPED_TEST(TestDeltaBitPackEncoding, BasicRoundTrip) {
   }
 }
 
+TYPED_TEST(TestDeltaBitPackEncoding, ZeroRoundTrip) {
```
rok, line 1439: Would this be a better name?

```suggestion
TYPED_TEST(TestDeltaBitPackEncoding, SmallBatchRoundTrip) {
```

cpp/src/parquet/encoding.cc
mapleFU: To be honest, the logic is a bit trickey here, I will try to explain it. I'm bad in English, so maybe you can edit it directly if you can explain it better.

mapleFU:
@tachyonwill I think this patch might fix `DELTA_BINARY_PACKED` issue, mind take a look?

cpp/src/parquet/encoding.cc
```diff
@@ -2479,7 +2481,18 @@ class DeltaBitPackDecoder : public DecoderImpl, virtual public TypedDecoder<DTyp
...
         if (ARROW_PREDICT_FALSE(!block_initialized_)) {
           buffer[i++] = last_value_;
-          if (ARROW_PREDICT_FALSE(i == max_values)) break;
+          if (ARROW_PREDICT_FALSE(i == max_values)) {
+            // When block is uninitialized and `i` reaches max_values, there are different
+            // cases:
+            // 1. `i == total_value_count_`. It means that the page may have only one
+            // value, and we should not initialize any block.
+            // 2. `i != total_value_count_`. It means that user just read the first value
+            // in the page, so, we should initialize the incoming block.
```
rok, line 2490:
```suggestion
            // When block is uninitialized and i reaches max_values we have two
            // different possibilities:
            // 1. i == total_value_count_, which means that the page may have only one
            // value and we should not initialize any block.
            // 2. i != total_value_count_ which means that user just read the first value
            // in the page, so we should initialize the incoming block.
```

mapleFU:
Retrigger CI because I think the fail is not caused by me, but still it looks like a bug: https://github.com/apache/arrow/actions/runs/3808810467/jobs/6479699758

By the way, is there any method that I can retrigger the CI?

mapleFU:
@pitrou Mind take a look here?

pitrou:
> @pitrou Mind take a look here?

In a few days certainly :-)

rok:
> By the way, is there any method that I can retrigger the CI?

Once your first PR is merged CI will start automatically.

cpp/src/parquet/encoding.cc
```diff
@@ -2479,7 +2481,18 @@ class DeltaBitPackDecoder : public DecoderImpl, virtual public TypedDecoder<DTyp
...
+          if (ARROW_PREDICT_FALSE(i == max_values)) {
+            // When block is uninitialized and i reaches max_values we have two
+            // different possibilities:
+            // 1. i == total_value_count_, which means that the page may have only one
+            // value and we should not initialize any block.
+            // 2. i != total_value_count_ which means that user just read the first value
+            // in the page, so we should initialize the incoming block.
+            if (i != static_cast<int>(total_value_count_)) {
+              InitBlock();
+            }
```
pitrou, line 2493: Wouldn't it be clearer to write this as:
```suggestion
            if (total_value_count_ != 1) {
              InitBlock();
            }
```

cpp/src/parquet/encoding.cc
```diff
@@ -2479,7 +2481,18 @@ class DeltaBitPackDecoder : public DecoderImpl, virtual public TypedDecoder<DTyp
       if (ARROW_PREDICT_FALSE(values_current_mini_block_ == 0)) {
         if (ARROW_PREDICT_FALSE(!block_initialized_)) {
           buffer[i++] = last_value_;
-          if (ARROW_PREDICT_FALSE(i == max_values)) break;
+          if (ARROW_PREDICT_FALSE(i == max_values)) {
```
pitrou, line 2484: Perhaps add this check
```suggestion
          DCHECK_EQ(i, 1);  // we're at the beginning of the page
          if (ARROW_PREDICT_FALSE(i == max_values)) {
```

cpp/src/parquet/encoding_test.cc
```diff
@@ -1324,6 +1324,29 @@ class TestDeltaBitPackEncoding : public TestEncodingBase<Type> {
     CheckRoundtripSpaced(valid_bits, valid_bits_offset);
   }
 
+  void ExecuteSteps(int nvalues, int repeats, int read_batch) {
```
pitrou, line 1327: Why not test different batch sizes as part of `CheckRoundtrip` instead?

mapleFU: It's ok, but most test current not use it. So I'd like to keep it simple here. If later we want test `batch size` for other encoding, we can move it to `CheckRoundtrip`

cpp/src/parquet/encoding.cc
mapleFU: By the way, should we add `ARROW_PREDICT_FALSE` here? @pitrou

cpp/src/parquet/encoding.cc
mapleFU: Nice catch, would add it

mapleFU:
oops, I got a 
```
______________________ [doctest] pyarrow._fs.PyFileSystem ______________________
EXAMPLE LOCATION UNKNOWN, not showing all tests of that example
??? >>> gfs = github.GithubFileSystem('apache', 'arrow', sha='ec51aec4d15035f4d9d6a1c4346d0a2b9a37fb75')
UNEXPECTED EXCEPTION: HTTPError('403 Client Error: rate limit exceeded for url: https://api.github.com/repos/apache/arrow/git/trees/ec51aec4d15035f4d9d6a1c4346d0a2b9a37fb75')
  File "/opt/conda/envs/arrow/lib/python3.9/doctest.py", line 1334, in __run
    exec(compile(example.source, filename, "single",
  File "<doctest pyarrow._fs.PyFileSystem[1]>", line 1, in <module>
  File "/opt/conda/envs/arrow/lib/python3.9/site-packages/fsspec/spec.py", line 76, in __call__
    obj = super().__call__(*args, **kwargs)
  File "/opt/conda/envs/arrow/lib/python3.9/site-packages/fsspec/implementations/github.py", line 56, in __init__
    self.ls("")
  File "/opt/conda/envs/arrow/lib/python3.9/site-packages/fsspec/implementations/github.py", line 158, in ls
    r.raise_for_status()
  File "/opt/conda/envs/arrow/lib/python3.9/site-packages/requests/models.py", line 1021, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 403 Client Error: rate limit exceeded for url: https://api.github.com/repos/apache/arrow/git/trees/ec51aec4d15035f4d9d6a1c4346d0a2b9a37fb75
/opt/conda/envs/arrow/lib/python3.9/site-packages/pyarrow/_fs.cpython-39-x86_64-linux-gnu.so:None: UnexpectedException
=============================== warnings summary ===============================
opt/conda/envs/arrow/lib/python3.9/site-packages/pytest_cython/plugin.py:57
  /opt/conda/envs/arrow/lib/python3.9/site-packages/pytest_cython/plugin.py:57: PytestRemovedIn8Warning: The (fspath: py.path.local) argument to DoctestModule is deprecated. Please use the (path: pathlib.Path) argument instead.
  See https://docs.pytest.org/en/latest/deprecations.html#fspath-argument-for-node-constructors-replaced-with-pathlib-path
    return DoctestModule.from_parent(parent, fspath=path)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=========================== short test summary info ============================
SKIPPED [2] opt/conda/envs/arrow/lib/python3.9/site-packages/_pytest/doctest.py:455: all tests skipped by +SKIP option
============= 1 failed, 223 passed, 2 skipped, 1 warning in 8.70s ==============
```

Seems that maybe we should rerun it?

cpp/src/parquet/encoding.cc
pitrou: It's not useful since this case is exceptional already.

cpp/src/parquet/encoding_test.cc
pitrou: Well, it would actually be simpler without adding a separate test method... and it would exercise the code more thoroughly as well.

pitrou:
> oops, I got a
> 
> [...]
> 
> Seems that maybe we should rerun it?

No worries, this is unrelated to his PR.

mapleFU:
> Well, it would actually be simpler without adding a separate test method... and it would exercise the code more thoroughly as well.

So how can I organize that code? I didn't make it clear. Is it like:

## case 1

```c++
template <typename Type>
class TestEncodingBase : public ::testing::Test {
 public:
...
 private:
   std::vector<int> small_read_batch_size_;
};
```

and 

```c++
TYPED_TEST(TestDeltaBitPackEncoding, SmallBatchRoundTrip) {
  // .. set draws_ 1, 3
  ASSERT_NO_FATAL_FAILURE(this->Execute(1, 1));
  ASSERT_NO_FATAL_FAILURE(this->Execute(1, 2));
  ASSERT_NO_FATAL_FAILURE(this->Execute(2, 2));
  ASSERT_NO_FATAL_FAILURE(this->ExecuteSteps(10, 10));
  ASSERT_NO_FATAL_FAILURE(this->ExecuteSteps(10, 10));
  ASSERT_NO_FATAL_FAILURE(this->ExecuteSteps(1, 10));
}
```

Or 

## case 2

```c++
template <typename Type>
class TestDeltaBitPackEncoding : public TestEncodingBase<Type> {
 public:
  using c_type = typename Type::c_type;
  static constexpr int TYPE = Type::type_num;
  static constexpr size_t ROUND_TRIP_TIMES = 3;

  void InitReadBatchSizedData(...);
};
```

Can you help me make it clear? @pitrou

pitrou:
You could have something like:
```c++
  const std::vector<int> kBatchSizes = {...};

  void CheckRoundtrip() override {
    auto encoder =
        MakeTypedEncoder<Type>(Encoding::DELTA_BINARY_PACKED, false, descr_.get());
    auto decoder = MakeTypedDecoder<Type>(Encoding::DELTA_BINARY_PACKED, descr_.get());
    for (size_t i = 0; i < ROUND_TRIP_TIMES; ++i) {
      for (const auto batch_size : kBatchSizes} {
        ...
```

mapleFU:
Seems that would causing our testing run a long time, should I make `kBatchSizes` configurable, and just using another `Execute`? Like:

```c++
template <typename Type>
class TestDeltaBitPackEncoding : public TestEncodingBase<Type> {
 public:
  using c_type = typename Type::c_type;
  static constexpr int TYPE = Type::type_num;
  static constexpr size_t ROUND_TRIP_TIMES = 3;

  void InitReadBatchSizedData(...);
};

  void CheckRoundtrip() override {
    ...
    auto read_batch_sizes = extra_read_batch_sizes;
    read_batch_sizes.push_back(num_values_);
    for (size_t i = 0; i < ROUND_TRIP_TIMES; ++i) {
      for (int read_batch_size: read_batch_sizes) {
        encoder->Put(draws_, num_values_);
        encode_buffer_ = encoder->FlushValues();

        decoder->SetData(num_values_, encode_buffer_->data(),
                         static_cast<int>(encode_buffer_->size()));

        int values_decoded_sum = 0;
        while (values_decoded_sum < num_values_) {
          int values_decoded =
              decoder->Decode(decode_buf_ + values_decoded_sum, read_batch_size);
          values_decoded_sum += values_decoded;
        }
        ASSERT_EQ(num_values_, values_decoded_sum);
        ASSERT_NO_FATAL_FAILURE(VerifyResults<c_type>(decode_buf_, draws_, num_values_));
      }
    }
  }
```

and

```
void ExecuteSteps(int nvalues, int repeats, std::vector<int> extra_read_batch_size) {
  this->InitReadBatchSizedData(nvalues, repeats, extra_read_batch_size);
  this->CheckRoundtrip();
}
```

I'm arguing this, because when I set `const std::vector<int> EXTRA_READ_BATCH_SIZES = {1, 10};` and make `CheckRoundTrip` run it every time, the testing will spend 5 minutes in debug mode, which makes me mad :( `ExecuteSteps` will only finished in one second

mapleFU:
@pitrou comment fixed and merged your patch which could make test faster. Mind take a look?

pitrou (approved):
+1

PR #15125, opened 2022-12-30
GH-28074: [C++][Dataset] Handle NaNs correctly in Parquet predicate push-down

### LINKED ISSUE #28074: [C++][Dataset] Handle NaNs correctly in Parquet predicate push-down
The Parquet spec (in parquet.thrift) says the following about handling of floating-point statistics:
```Java

   * (*) Because the sorting order is not specified properly for floating
   *     point values (relations vs. total ordering) the following
   *     compatibility rules should be applied when reading statistics:
   *     - If the min is a NaN, it should be ignored.
   *     - If the max is a NaN, it should be ignored.
   *     - If the min is +0, the row group may contain -0 values as well.
   *     - If the max is -0, the row group may contain +0 values as well.
   *     - When looking for NaN values, min and max should be ignored.
```

It appears that the dataset code uses the following filter expression when doing Parquet predicate push-down (in `file_parquet.cc`):
```c++

    return and_(greater_equal(field_expr, literal(min)),
                less_equal(field_expr, literal(max)));
```

A NaN value will fail that filter and yet may be found in the given Parquet column chunk.

We may instead need a "greater_equal_or_nan" comparison that returns true if either value is NaN.

**Reporter**: [Antoine Pitrou](https://issues.apache.org/jira/browse/ARROW-12264) / @pitrou
**Assignee**: [Sanjiban Sengupta](https://issues.apache.org/jira/browse/ARROW-12264) / @sanjibansg
#### Related issues:
- [Specify a well-defined sorting order for float and double types](https://issues.apache.org/jira/browse/PARQUET-1222) (relates to)
#### PRs and other links:
- [GitHub Pull Request #15125](https://github.com/apache/arrow/pull/15125)

<sub>**Note**: *This issue was originally created as [ARROW-12264](https://issues.apache.org/jira/browse/ARROW-12264). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

sanjibansg:
This PR fixes the issue of handling NaNs in the Parquet predicate push-down. 
While computing the valid bounds for a column, if the max or min of the column is null, the range should ignore that. 
* Closes: #28074

### DIFF
--- a/cpp/src/arrow/dataset/file_parquet.cc
+++ b/cpp/src/arrow/dataset/file_parquet.cc
@@ -101,10 +101,10 @@ Result<std::shared_ptr<SchemaManifest>> GetSchemaManifest(
 bool isNan(const Scalar& value) {
   if (value.type->Equals(*float32())) {
     const FloatScalar& float_scalar = checked_cast<const FloatScalar&>(value);
-    return isnan(float_scalar.value);
+    return std::isnan(float_scalar.value);
   } else if (value.type->Equals(*float64())) {
     const DoubleScalar& double_scalar = checked_cast<const DoubleScalar&>(value);
-    return isnan(double_scalar.value);
+    return std::isnan(double_scalar.value);
   }
   return false;
 }
--- a/cpp/src/arrow/dataset/file_parquet.cc
+++ b/cpp/src/arrow/dataset/file_parquet.cc
@@ -126,7 +126,12 @@ std::optional<compute::Expression> ColumnChunkStatisticsAsExpression(
   auto column_metadata = metadata.ColumnChunk(schema_field.column_index);
   auto statistics = column_metadata->statistics();
   const auto& field = schema_field.field;
-  return ParquetFileFragment::EvaluateStatisticsAsExpression(field, statistics);
+
+  if (statistics == nullptr) {
+    return std::nullopt;
+  }
+
+  return ParquetFileFragment::EvaluateStatisticsAsExpression(*field, *statistics);
 }
 
 void AddColumnIndices(const SchemaField& schema_field,
@@ -278,15 +283,11 @@ Result<bool> IsSupportedParquetFile(const ParquetFileFormat& format,
 }  // namespace
 
 std::optional<compute::Expression> ParquetFileFragment::EvaluateStatisticsAsExpression(
-    std::shared_ptr<Field> field, std::shared_ptr<parquet::Statistics> statistics) {
-  if (statistics == nullptr) {
-    return std::nullopt;
-  }
-
-  auto field_expr = compute::field_ref(field->name());
+    const Field& field, const parquet::Statistics& statistics) {
+  auto field_expr = compute::field_ref(field.name());
 
   // Optimize for corner case where all values are nulls
-  if (statistics->num_values() == 0 && statistics->null_count() > 0) {
+  if (statistics.num_values() == 0 && statistics.null_count() > 0) {
     return is_null(std::move(field_expr));
   }
 
@@ -295,8 +296,8 @@ std::optional<compute::Expression> ParquetFileFragment::EvaluateStatisticsAsExpr
     return std::nullopt;
   }
 
-  auto maybe_min = min->CastTo(field->type());
-  auto maybe_max = max->CastTo(field->type());
+  auto maybe_min = min->CastTo(field.type());
+  auto maybe_max = max->CastTo(field.type());
 
   if (maybe_min.ok() && maybe_max.ok()) {
     min = maybe_min.MoveValueUnsafe();
@@ -312,7 +313,7 @@ std::optional<compute::Expression> ParquetFileFragment::EvaluateStatisticsAsExpr
     if (min->Equals(max)) {
       auto single_value = compute::equal(field_expr, compute::literal(std::move(min)));
 
-      if (statistics->null_count() == 0) {
+      if (statistics.null_count() == 0) {
         return single_value;
       }
       return compute::or_(std::move(single_value), is_null(std::move(field_expr)));
@@ -332,7 +333,7 @@ std::optional<compute::Expression> ParquetFileFragment::EvaluateStatisticsAsExpr
       in_range = compute::and_(std::move(lower_bound), std::move(upper_bound));
     }
 
-    if (statistics->null_count() != 0) {
+    if (statistics.null_count() != 0) {
       return compute::or_(std::move(in_range), compute::is_null(field_expr));
     }
     return in_range;
--- a/cpp/src/arrow/dataset/file_parquet.cc
+++ b/cpp/src/arrow/dataset/file_parquet.cc
@@ -98,13 +98,15 @@ Result<std::shared_ptr<SchemaManifest>> GetSchemaManifest(
   return manifest;
 }
 
-bool isNan(const Scalar& value) {
-  if (value.type->Equals(*float32())) {
-    const FloatScalar& float_scalar = checked_cast<const FloatScalar&>(value);
-    return std::isnan(float_scalar.value);
-  } else if (value.type->Equals(*float64())) {
-    const DoubleScalar& double_scalar = checked_cast<const DoubleScalar&>(value);
-    return std::isnan(double_scalar.value);
+bool IsNan(const Scalar& value) {
+  if (value.is_valid) {
+    if (value.type->id() == Type::FLOAT) {
+      const FloatScalar& float_scalar = checked_cast<const FloatScalar&>(value);
+      return std::isnan(float_scalar.value);
+    } else if (value.type->id() == Type::DOUBLE) {
+      const DoubleScalar& double_scalar = checked_cast<const DoubleScalar&>(value);
+      return std::isnan(double_scalar.value);
+    }
   }
   return false;
 }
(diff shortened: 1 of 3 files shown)


cpp/src/arrow/dataset/file_parquet.cc
```diff
@@ -143,18 +147,25 @@ std::optional<compute::Expression> ColumnChunkStatisticsAsExpression(
...
       return compute::or_(std::move(single_value), is_null(std::move(field_expr)));
     }
-
+    
     auto lower_bound =
-        compute::greater_equal(field_expr, compute::literal(std::move(min)));
-    auto upper_bound = compute::less_equal(field_expr, compute::literal(std::move(max)));
+        compute::greater_equal(field_expr, compute::literal(min));
+    auto upper_bound = compute::less_equal(field_expr, compute::literal(max));
+    auto in_range = compute::Expression();
```
westonpace, line 154:
```suggestion
    compute::Expression in_range;
```

westonpace (changes requested):
This looks like the right logic but I think we should be checking for NaN and not using `is_valid` (which checks for null).

So something like...

bool isNan(const Scalar& scalar) {
  if (IsFloat(scalar)) {
    const FloatScalar& float_scalar = checked_cast<const FloatScalar&>(scalar);
    return isnan(float_scalar);
  } else if (IsDouble(scalar)) {
    // ...
  } else {
    return false;
  }
}

cpp/src/arrow/dataset/file_parquet.cc
sanjibansg: made the changes, thanks!

westonpace (commented):
This looks correct to me. @pitrou did you want to do a quick sanity check?

How hard would it be to mock up some kind of test to check this?  Unfortunately, `ColumnChunkStatisticsAsExpression` isn't really a public method and so it might be tricky coming up with a test case without reproducing an offending parquet file which may not be easy to do.

cpp/src/arrow/dataset/file_parquet.cc
```diff
@@ -98,6 +98,17 @@ Result<std::shared_ptr<SchemaManifest>> GetSchemaManifest(
   return manifest;
 }
 
+bool isNan(const Scalar& value) {
```
pitrou, line 101: Please, let's follow the coding conventions
```suggestion
bool IsNan(const Scalar& value) {
```

cpp/src/arrow/dataset/file_parquet.cc
```diff
@@ -98,6 +98,17 @@ Result<std::shared_ptr<SchemaManifest>> GetSchemaManifest(
...
+  if (value.type->Equals(*float32())) {
```
pitrou, line 102: Doing a full type equality check is unnecessarily slow here
```suggestion
  if (value.type->id() == Type::FLOAT) {
```

cpp/src/arrow/dataset/file_parquet.cc
```diff
@@ -98,6 +98,17 @@ Result<std::shared_ptr<SchemaManifest>> GetSchemaManifest(
...
+    const FloatScalar& float_scalar = checked_cast<const FloatScalar&>(value);
+    return std::isnan(float_scalar.value);
```
pitrou, line 104: Do we know for sure that `float_scalar.valid` is true?

cpp/src/arrow/dataset/file_parquet.cc
```diff
@@ -144,17 +159,23 @@ std::optional<compute::Expression> ColumnChunkStatisticsAsExpression(
       return compute::or_(std::move(single_value), is_null(std::move(field_expr)));
     }
 
-    auto lower_bound =
-        compute::greater_equal(field_expr, compute::literal(std::move(min)));
-    auto upper_bound = compute::less_equal(field_expr, compute::literal(std::move(max)));
+    auto lower_bound = compute::greater_equal(field_expr, compute::literal(min));
```
pitrou, line 162: Perhaps the code could be better organized, to that NaN handling happens in a single place?
(note that `min->Equals(max)` can only be false if there's a NaN...)

cpp/src/arrow/dataset/file_parquet.cc
```diff
@@ -135,6 +146,10 @@ std::optional<compute::Expression> ColumnChunkStatisticsAsExpression(
     min = maybe_min.MoveValueUnsafe();
     max = maybe_max.MoveValueUnsafe();
 
+    if (isNan(*min) && isNan(*max)) {
```
pitrou, line 149: Please add a comment explaining the reasons why NaN needs to be handled explicitly.

pitrou (changes requested):

pitrou:
> Unfortunately, `ColumnChunkStatisticsAsExpression` isn't really a public method and so it might be tricky coming up with a test case without reproducing an offending parquet file which may not be easy to do.

I agree we should ideally unit test `ColumnChunkStatisticsAsExpression`. Perhaps it can be exposed as an internal API?

cpp/src/arrow/dataset/file_parquet.cc
sanjibansg: Thanks for the review, apologies for missing that out. Made the change!

cpp/src/arrow/dataset/file_parquet.cc
sanjibansg: Made the change, thanks!

cpp/src/arrow/dataset/file_parquet.cc
sanjibansg: Checking for validity now.

cpp/src/arrow/dataset/file_parquet.cc
sanjibansg: Added a comment mentioning the NaN handling.

cpp/src/arrow/dataset/file_parquet.cc
sanjibansg: Apologies, I did not fully understand that. Can you elaborate on considering the inequality for min and max, and determining NaNs? Thanks!

wjones127:
I believe I have generated a file that could be used for a test case here: https://github.com/apache/parquet-testing/pull/35

Does that seem sufficient?

westonpace:
Looks like the parquet tests are failing because they can't find the parquet file.  Do you maybe need to update the submodule version?  I haven't done this recently and am not sure of the right commands.

sanjibansg:
> Looks like the parquet tests are failing because they can't find the parquet file. Do you maybe need to update the submodule version? I haven't done this recently and am not sure of the right commands.
@westonpace 
I think by mistake, I updated the `arrow-testing` submodule as well while updating `parquet-testing`. Will it be okay?

cpp/src/arrow/dataset/file_parquet.cc
```diff
@@ -306,6 +277,69 @@ Result<bool> IsSupportedParquetFile(const ParquetFileFormat& format,
 
 }  // namespace
 
+std::optional<compute::Expression> ParquetFileFragment::EvaluateStatisticsAsExpression(
+    std::shared_ptr<Field> field, std::shared_ptr<parquet::Statistics> statistics) {
```
westonpace, line 281:
```suggestion
std::optional<compute::Expression> ParquetFileFragment::EvaluateStatisticsAsExpression(
    const Field& field, const parquet::Statistics& statistics) {
```

Minor nit: prefer accepting `const X&` over `std::shared_ptr<X>` or even `const std::shared_ptr<X>&` if you can.

westonpace (commented):
This looks good.  I don't see a problem with updating the arrow testing repo but it might be nice to keep changes isolated if you can.

cpp/src/arrow/dataset/file_parquet.cc
sanjibansg: Made the change, thanks!

sanjibansg:
> This looks good. I don't see a problem with updating the arrow testing repo but it might be nice to keep changes isolated if you can.

Yes, of course, and sorry, I will be more careful next time.

cpp/src/arrow/dataset/file_parquet.cc
```diff
@@ -306,6 +282,65 @@ Result<bool> IsSupportedParquetFile(const ParquetFileFormat& format,
...
+      return compute::or_(std::move(single_value), is_null(std::move(field_expr)));
+    }
+
+    auto lower_bound = compute::greater_equal(field_expr, compute::literal(min));
+    auto upper_bound = compute::less_equal(field_expr, compute::literal(max));
+    compute::Expression in_range;
+
+    // If either minimum or maximum is NaN, it should be ignored for the
+    // range computation
+    if (IsNan(*min)) {
```
westonpace, line 328:
```suggestion
    if (min->Equals(max)) {
      auto single_value = compute::equal(field_expr, compute::literal(std::move(min)));

      if (statistics.null_count() == 0) {
        return single_value;
      }
      return compute::or_(std::move(single_value), is_null(std::move(field_expr)));
    }

    auto lower_bound = compute::greater_equal(field_expr, compute::literal(min));
    auto upper_bound = compute::less_equal(field_expr, compute::literal(max));
    compute::Expression in_range;

    // Since the minimum & maximum values are NaN, useful statistics
    // cannot be extracted for checking the presence of a value within
    // range
    if (IsNan(*min) && IsNan(*max)) {
      return std::nullopt;
    }

    // If either minimum or maximum is NaN, it should be ignored for the
    // range computation
    if (IsNan(*min)) {
```
Per @pitrou 's earlier request, let's try and keep all the NaN handling grouped together.

westonpace (changes requested):
One minor nit

cpp/src/arrow/dataset/file_parquet.cc
sanjibansg: made the change, thanks!

westonpace (approved):

westonpace:
Thanks!

PR #15129, opened 2022-12-30
GH-15059: [C++][Acero] populate guarantee columns from expression intstead of fragment

### LINKED ISSUE #15059: [C++] The new scan node should use values from fragment guarantees instead of loading them from disk

The main reason we need to do this is because the columns are not always going to be on the disk (right now the new scan node fails in this case).   It's also a performance enhancement to skip loading of these columns as well.  The solution will, I suspect, also lay the groundwork for adding support for the augmented columns as well (filename, batch index, file index)

### Component(s)

C++

westonpace:
If a fragment has a guarantee like `x == 5` then we don't need to load the column `x` from disk and can instead just use the scalar `5`.  This is not just a performance improvement.  In many cases, users will create partitioned datasets without actually storing the partition value as a separate column (e.g. the file `my_dataset/x=5/foo.parquet` will not have a column named `x`)
* Closes: #15059

### DIFF
--- a/cpp/src/arrow/util/utf8.h
+++ b/cpp/src/arrow/util/utf8.h
@@ -31,7 +31,7 @@ namespace util {
 
 // Convert a UTF8 string to a wstring (either UTF16 or UTF32, depending
 // on the wchar_t width).
-ARROW_EXPORT Result<std::wstring> UTF8ToWideString(const std::string_view& source);
+ARROW_EXPORT Result<std::wstring> UTF8ToWideString(std::string_view source);
 
 // Similarly, convert a wstring to a UTF8 string.
 ARROW_EXPORT Result<std::string> WideStringToUTF8(const std::wstring& source);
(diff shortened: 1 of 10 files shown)


westonpace:
It turns out the scan_node is pretty broken with respect to marking "finished" correctly. It could be fixed but would add quite a bit of complexity and I would rather wait and remove the concept of finished altogether.

westonpace:
Coming back to this now that the various dependent PRs have all been merged.  I think this is in good shape now.

cpp/src/arrow/util/utf8.h
```diff
@@ -31,7 +31,7 @@ namespace util {
 
 // Convert a UTF8 string to a wstring (either UTF16 or UTF32, depending
 // on the wchar_t width).
-ARROW_EXPORT Result<std::wstring> UTF8ToWideString(const std::string& source);
+ARROW_EXPORT Result<std::wstring> UTF8ToWideString(const std::string_view& source);
```
lidavidm, line 34: nit: pass string_view by value?

lidavidm (approved):

PR #15131, opened 2022-12-30
ARROW-18195: [C++] Fix case_when produces bad data when condition has nulls

wjones127:
When the condition has nulls, we were going into a branch that did not advance the `offset`. On subsequent blocks (each 64 values), the `mask` (which tells us which positions have already been set by previous cases) was always already set, so the condition was always evaluated as False regardless of its true values. So the first 64 values would be correct, but every value afterwards would skip the condition with nulls and use the next condition (or the else).

### DIFF
--- a/cpp/src/arrow/compute/kernels/scalar_if_else.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_if_else.cc
@@ -1568,6 +1568,7 @@ Status ExecArrayCaseWhen(KernelContext* ctx, const ExecSpan& batch, ExecResult* 
             }
           }
         }
+        offset += block_length;
       });
     }
   }
--- a/cpp/src/arrow/compute/kernels/scalar_if_else_test.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_if_else_test.cc
@@ -1218,6 +1218,25 @@ void TestCaseWhenFixedSize() {
          ArrayFromJSON(type, "[30, 31, 32, 33, 34, null, 36, 37, null]")},
         ArrayFromJSON(type, "[10, 11, 12, 23, 34, null, 26, 37, null]"));
 
+    // ARROW-18195
+    auto len = 65;
+
+    ASSERT_OK_AND_ASSIGN(auto temp_cond0, MakeArrayOfNull(boolean(), 1));
+    ASSERT_OK_AND_ASSIGN(auto temp_cond1, MakeArrayFromScalar(BooleanScalar(true), len - 1));
+    ASSERT_OK_AND_ASSIGN(auto cond, Concatenate({temp_cond0, temp_cond1}));
+
+    ASSERT_OK_AND_ASSIGN(auto ones, MakeArrayFromScalar(Int64Scalar(1), len));
+
+    ASSERT_OK_AND_ASSIGN(auto temp_res0, MakeArrayOfNull(int64(), 1));
+    ASSERT_OK_AND_ASSIGN(auto temp_res1, MakeArrayFromScalar(Int64Scalar(1), len - 1));
+    ASSERT_OK_AND_ASSIGN(auto res, Concatenate({temp_res0, temp_res1}));
+
+    CheckScalar(
+      "case_when",
+      {MakeStruct({cond}), ones},
+      res
+    );
+
     // Error cases
     EXPECT_RAISES_WITH_MESSAGE_THAT(
         Invalid,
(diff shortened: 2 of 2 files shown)


wjones127:
Double checked, and @thisisnic's repros are all fixed by this:

``` r

# Specific conditions where this happens: a table with one NA and 64 or more non-NA values
test_df = tibble::tibble(x = c(NA, rep("foo", 64)))
test_arrow = arrow_table(test_df)

# the non-arrow version; all the final values are 1
test_df %>%
  mutate(y = case_when(x == 'foo' ~ 1, is.na(x) ~ NA_real_)) %>%
  tail()
#> # A tibble: 6 × 2
#>   x         y
#>   <chr> <dbl>
#> 1 foo       1
#> 2 foo       1
#> 3 foo       1
#> 4 foo       1
#> 5 foo       1
#> 6 foo       1

# the arrow version; the final values is NA
test_arrow %>%
  mutate(y = case_when(x == 'foo' ~ 1, is.na(x) ~ NA_real_)) %>%
  collect() %>%
  tail()
#> # A tibble: 6 × 2
#>   x         y
#>   <chr> <dbl>
#> 1 foo       1
#> 2 foo       1
#> 3 foo       1
#> 4 foo       1
#> 5 foo       1
#> 6 foo       1

# it's fine if there are less than 65 values in the table (i.e. but still contains an NA)
test_arrow[1:64,] %>%
  mutate(y = case_when(x == 'foo' ~ 1, is.na(x) ~ NA_real_)) %>%
  collect() %>%
  tail()
#> # A tibble: 6 × 2
#>   x         y
#>   <chr> <dbl>
#> 1 foo       1
#> 2 foo       1
#> 3 foo       1
#> 4 foo       1
#> 5 foo       1
#> 6 foo       1

# everything is fine when the comparison is being done on doubles and return value is char
test_df2 = tibble::tibble(x = c(NA, rep(1, 64)))
test_arrow2 = arrow_table(test_df2)
test_arrow2 %>%
  mutate(y = case_when(x == 1 ~ "winning", is.na(x) ~ NA_character_)) %>%
  collect() %>%
  tail()
#> # A tibble: 6 × 2
#>       x y      
#>   <dbl> <chr>  
#> 1     1 winning
#> 2     1 winning
#> 3     1 winning
#> 4     1 winning
#> 5     1 winning
#> 6     1 winning

# also breaks when source value is boolean and target value is double
test_df3 = tibble::tibble(x = c(NA, rep(TRUE, 64)))
test_arrow3 = arrow_table(test_df3)
test_arrow3 %>%
  mutate(y = case_when(x == TRUE ~ 1, is.na(x) ~ NA_real_)) %>%
  collect() %>%
  tail()
#> # A tibble: 6 × 2
#>   x         y
#>   <lgl> <dbl>
#> 1 TRUE      1
#> 2 TRUE      1
#> 3 TRUE      1
#> 4 TRUE      1
#> 5 TRUE      1
#> 6 TRUE      1

# also broken for when target is integer
test_df4 = tibble::tibble(x = c(NA, rep(TRUE, 64)))
test_arrow4 = arrow_table(test_df4)
test_arrow4 %>%
  mutate(y = case_when(x == TRUE ~ 1L, is.na(x) ~ 2L)) %>%
  collect() %>%
  tail()
#> # A tibble: 6 × 2
#>   x         y
#>   <lgl> <int>
#> 1 TRUE      1
#> 2 TRUE      1
#> 3 TRUE      1
#> 4 TRUE      1
#> 5 TRUE      1
#> 6 TRUE      1

# broken for logical to logical
test_df5 = tibble::tibble(x = c(NA, rep(TRUE, 64)))
test_arrow5 = arrow_table(test_df5)
test_arrow5 %>%
  mutate(y = case_when(x == TRUE ~ TRUE, is.na(x) ~ FALSE)) %>%
  collect() %>%
  tail()
#> # A tibble: 6 × 2
#>   x     y    
#>   <lgl> <lgl>
#> 1 TRUE  TRUE 
#> 2 TRUE  TRUE 
#> 3 TRUE  TRUE 
#> 4 TRUE  TRUE 
#> 5 TRUE  TRUE 
#> 6 TRUE  TRUE
```

<sup>Created on 2022-12-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

pitrou:
Hmm, it is worrying that we let this slip through. Do you think there would be an easy way to test this more thoroughly?

wjones127:
I think the ifelse kernel has a more thorough tests that we can adapt for case when:

https://github.com/apache/arrow/blob/63b91cc1f7131356537ab9cbb84ed108d6f9102e/cpp/src/arrow/compute/kernels/scalar_if_else_test.cc#L508-L518

wjones127:
@pitrou I've written a new test function that exercises all code paths within `ExecArrayCaseWhen`. I validated it hits all branches by making sure it hit relevant breakpoints in lldb. This includes whether there is an else value and whether there are nulls in the condition.

pitrou (approved):

PR #15142, opened 2022-12-31
GH-15141: [C++] fix for unstable test due to unstable sort

### LINKED ISSUE #15141: [CI] arrow-compute: ExecPlanExecution.StressSourceOrderBy may failed

1. https://github.com/apache/arrow/actions/runs/3812471160/jobs/6485697443
2. https://github.com/apache/arrow/actions/runs/3808810467/jobs/6479699758

Seems that the case may failed. And I didn't modify the case which could cause it failed.

### Component(s)

C++

westonpace:
The sorting done by orderby is not stable.  This means, given the input:

a | b
--- | ---
1 | false
1 | true

the test could have generated both `[false, true]` and `[true, false]` for the `b` column.  We likely did not encounter this before https://github.com/apache/arrow/commit/498b645e1d09306bf5399a9a019a5caa99513815 because the entire thing was run serially (even though there was a `parallel` option it was not setup correctly).

Now that things are properly running parallel the results are non-deterministic.  We could remove the `b` column but I feel it is a better stress test to have at least one payload column.  So I changed the test to only compare the key array and not the payload array.
* Closes: #15141

### DIFF
--- a/cpp/src/arrow/compute/exec/plan_test.cc
+++ b/cpp/src/arrow/compute/exec/plan_test.cc
@@ -808,7 +808,9 @@ TEST(ExecPlanExecution, StressSourceOrderBy) {
                            TableFromExecBatches(input_schema, random_data.batches));
       ASSERT_OK_AND_ASSIGN(auto sort_indices, SortIndices(original, options));
       ASSERT_OK_AND_ASSIGN(auto expected, Take(original, sort_indices));
-      AssertTablesEqual(*actual, *expected.table());
+      AssertSchemaEqual(actual->schema(), expected.table()->schema());
+      AssertArraysEqual(*actual->column(0)->chunk(0),
+                        *expected.table()->column(0)->chunk(0));
     }
   }
 }


westonpace:
@vibhatha would you be interested in providing a review / sanity check?

vibhatha:
> @vibhatha would you be interested in providing a review / sanity check?

Sure @westonpace, I will.

vibhatha (approved):
@westonpace make sense to me. It’s not possible to guarantee the outcome of this. 

Just a question, is there a sort option which gives precedence to the index of the row and decide which comes first, when we have a tie like this?

westonpace:
> Just a question, is there a sort option which gives precedence to the index of the row and decide which comes first, when we have a tie like this?

That's called a "stable sort". The underlying sort kernel (SortIndices) **is** stable.  However, if the plan is run in parallel, then there is no guarantee the batches will accumulate in the same order.  So even if the sort kernel is stable the sort node is not.

Once we add proper ordering we can add a stable option to the sort node which resequences the data before sorting so that the sort node can remain stable.

However, now that I write this, I realize it might be best to only apply my change when testing the parallel case, and to use the old comparison in the non-parallel case.

vibhatha:
> However, now that I write this, I realize it might be best to only apply my change when testing the parallel case, and to use the old comparison in the non-parallel case.

Yes, I also think it’s better that way.

westonpace:
> Yes, I also think it’s better that way.

Hmm, I tried this but it turns out not to be so simple.  I'm going to proceed with this how it is for now.  We can worry about a full comparison later when we add a stable sort.  I'll add a new issue requesting that.  I'll merge this so it doesn't bother CI

PR #15149, opened 2023-01-02
ARROW-18318: [Python] Expose Scalar.validate()

buaazhwb:


pitrou (approved):
Thanks @buaazhwb ! I rebased and will merge if CI passes.

buaazhwb:
@pitrou  Not yet. The public registeration has been disabled. Are there any doc for creating a jira account？

pitrou:
@buaazhwb I should be able to create one for you. Do I just use the same ID and e-mail address as for Github?

Note that in the future, we will only be using Github issues, so please don't create any new JIRA issues.

buaazhwb:
@pitrou OK. Thanks for your help!

PR #15152, opened 2023-01-02
GH-15150: [C++][FlightRPC] Wait for side effects in DoAction

### LINKED ISSUE #15150: [C++][FlightRPC] DoAction cannot be fully streaming
As noted by Kou, existing applications/tests may assume that DoAction is not actually a streaming call.

> This might break `wheel-macos-big-sur-*`. For example: https://github.com/ursacomputing/crossbow/actions/runs/3815472986/jobs/6490471956#step:11:435

```text
_________________________ test_none_action_side_effect _________________________

    def test_none_action_side_effect():
        """Ensure that actions are executed even when we don't consume iterator.
    
        See https://issues.apache.org/jira/browse/ARROW-14255
        """
    
        with ActionNoneFlightServer() as server, \
                FlightClient(('localhost', server.port)) as client:
            client.do_action(flight.Action("append", b""))
            r = client.do_action(flight.Action("get_value", b""))
>           assert json.loads(next(r).body.to_pybytes()) == [True]
E           assert [] == [True]
E             Right contains one more item: True
E             Full diff:
E             - [True]
E             + []

test-arm64-env/lib/python3.10/site-packages/pyarrow/tests/test_flight.py:2132: AssertionError
----------------------------- Captured stderr call -----------------------------
/Users/voltrondata/github-actions-runner/_work/crossbow/crossbow/arrow/cpp/src/arrow/status.cc:137: DoAction result was not fully consumed: Cancelled: Flight cancelled call, with message: CANCELLED. Detail: Cancelled
```

_Originally posted by @kou in https://github.com/apache/arrow/issues/15118#issuecomment-1368377678_

lidavidm:
* Closes: #15150

### DIFF
--- a/cpp/src/arrow/flight/flight_test.cc
+++ b/cpp/src/arrow/flight/flight_test.cc
@@ -1545,7 +1545,7 @@ TEST_F(TestCancel, DoActionSideEffect) {
   FlightCallOptions options;
   options.stop_token = stop_source.token();
   // Will block for a bit, but not forever
-  ASSERT_OK_AND_ASSIGN(auto stream, client_->DoAction(options, {"inc"}));
+  ASSERT_OK_AND_ASSIGN(auto stream, client_->DoAction(options, {"inc", nullptr}));
   // Side effect should have happened
   ASSERT_EQ(1, Server()->CheckCounter());
   stop_source.RequestStop(Status::Cancelled("StopSource"));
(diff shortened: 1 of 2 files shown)


lidavidm:

lidavidm:
I can't seem to reproduce those segfaults locally, but the formerly failing test doesn't fail anymore.

wjones127 (approved):

kou (approved):
+1

Thanks!

PR #15159, opened 2023-01-02
ARROW-15735: [C++] Hash aggregate functions to return first and last value from a group.

sanjibansg:
This PR adds a hash aggregate function which returns the first and last value in a batch based on an order column.


JacekPliszka:
I am surprised by the folding of kernel implementation.  Are compilers so good to optimize it?

Naively it looks like double the computing effort.

nealrichardson:
Looks like #34912 achieved this.

PR #15160, opened 2023-01-02
MINOR: [Docs][Java] Fix a link to IPC RecordBatch message

neshkeev:
The link used to lead to the page that says 

> The contents of this document have relocated to the main [Columnar Specification](https://arrow.apache.org/docs/format/Columnar.html#format-columnar) page.

now the link leads to the exact place in the documentation


kou (approved):
+1

Good catch!

PR #15166, opened 2023-01-03
ARROW-17302: [R] Configure curl timeout policy for S3

thisisnic:


wjones127 (approved):
Looks good!

To test, I tried to trigger a timeout using:

```R
library(arrow)
fs <- s3_bucket("voltrondata-labs-datasets", request_timeout=0.001)
fs$ls("diamonds")
```

But wasn't able to trigger it. I verified with LLDB that the parameters were being correctly passed down, so I think this a limitation of the AWS SDK. 🤷

PR #15169, opened 2023-01-03
GH-15168: [GLib] Add support for half float

### LINKED ISSUE #15168: [GLib] Add support for half float

Half float/16-bit float isn't implemented yet.

### Component(s)

GLib

kou:
* Closes: #15168


kou:
+1

PR #15177, opened 2023-01-03
GH-15174: [Go][FlightRPC] Expose Flight Server Desc and RegisterFlightService

### LINKED ISSUE #15174: [Go] Expose FlightService_ServiceDesc and RegisterFlightServiceServer

In order to incorporate a flight service into an existing grpc server, it is required to expose the Service Descriptor for the flight service. We should also expose the convenience function of `RegisterFlightServiceServer` as a convenient use to register a Flight service onto an existing `grpc.ServiceRegistrar`.

### Component(s)

FlightRPC, Go

zeroshade:
CC @mariusvniekerk 
* Closes: #15174

### DIFF
--- a/go/arrow/flight/client.go
+++ b/go/arrow/flight/client.go
@@ -234,7 +234,7 @@ type ClientMiddleware struct {
 }
 
 type client struct {
-	conn        *grpc.ClientConn
+	conn        grpc.ClientConnInterface
 	authHandler ClientAuthHandler
 
 	FlightServiceClient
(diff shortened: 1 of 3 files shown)


mariusvniekerk:
Looks like it would work nicely for using flight with an already existing grpc server.

HammadB:
Is there any plan to expose these to python so that you can use arrowflight alongside an existing python grpc server?

zeroshade:
@jorisvandenbossche Thoughts on @HammadB's suggestion?

PR #15179, opened 2023-01-03
GH-15042: [C++][Parquet] Update stats on subsequent batches of dictionaries

### LINKED ISSUE #15042: [Python][Parquet] Column statistics incorrect for Dictionary Column in Parquet

When writing a column that is a 

`pa.dictionary(pa.int32(), pa.string())` to a Parquet file incorrect statistics are produced.  I believe the statistics are calculated from the contents of the first chunk rather than all chunks.  Since the CSV parser produces chunks that may not contain all dictionary rows, this results in incorrect statistics to be produced.

A test case is below that demonstrates the problem without using the CSV parser:

```python
import pyarrow as pa
import pyarrow.parquet as pq

schema = pa.schema({"field_1": pa.dictionary(pa.int32(), pa.string())})

# The ordering of the values don't matter, but they must
# be stored in seperate chunks (as they would be if they are parsed from CSV)
arr_1 = pa.array(["rusty", "sean", "aa"]).dictionary_encode()
arr_2 = pa.array(["zzz", "frank"]).dictionary_encode()

t = pa.Table.from_batches(
    [
        pa.record_batch([arr_1], names=["field_1"]),
        pa.record_batch([arr_2], names=["field_1"]),
    ]
)

# If this is commented the bug does not occur.
t = t.unify_dictionaries()

with pq.ParquetWriter("example.parquet", schema) as writer:
    writer.write_table(t)

# The Parquet stats will be:
# ┌─────────────────┬─────────────────┐
# │ stats_min_value │ stats_max_value │
# ├─────────────────┼─────────────────┤
# │ aa              │ sean            │
# └─────────────────┴─────────────────┘
#
# The stats should be:
#
# ┌─────────────────┬─────────────────┐
# │ stats_min_value │ stats_max_value │
# ├─────────────────┼─────────────────┤
# │ aa              │ zzz             │
# └─────────────────┴─────────────────┘
#
```

Workaround: if you call

`t = t.combine_chunks()`

Before calling `write_table()` the proper column statistics are written.

The difficulty with having improper column statistics is that query engines (Athena, Trino) use column statistics to create predicate pushdowns as part of their query execution.  If these query plans are incorrect resulting in data that exists in the Parquet file not being returned as part of the query result.

### Component(s)

C++, Parquet, Python

wjones127:
* Closes: #15042

### DIFF
--- a/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
+++ b/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
@@ -4154,7 +4154,7 @@ TEST_P(TestArrowWriteDictionary, StatisticsUnifiedDictionary) {
         ArrayFromJSON(::arrow::utf8(), R"(["b", "c", "d", "a"])");
     std::vector<std::shared_ptr<::arrow::Array>> test_indices = {
         ArrayFromJSON(::arrow::int32(),
-                      R"([0, null, 3, 0, null, 3])"),  // ["b", null "a", "b", null, "a"]
+                      R"([3, null, 3, 3, null, 3])"),  // ["a", null "a", "a", null, "a"]
         ArrayFromJSON(
             ::arrow::int32(),
             R"([0, 3, null, 0, null, 1])")};  // ["b", "a", null, "b", null, "c"]
--- a/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
+++ b/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
@@ -4157,7 +4157,7 @@ TEST_P(TestArrowWriteDictionary, StatisticsUnifiedDictionary) {
                       R"([0, null, 3, 0, null, 3])"),  // ["b", null "a", "b", null, "a"]
         ArrayFromJSON(
             ::arrow::int32(),
-            R"([0, 3, null, 0, null, 1])")};  // ["b", "c", null, "b", "c", null]
+            R"([0, 3, null, 0, null, 1])")};  // ["b", "a", null, "b", null, "c"]
 
     ::arrow::ArrayVector chunks = {
         std::make_shared<DictionaryArray>(dict_type, test_indices[0], test_dictionary),
--- a/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
+++ b/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
@@ -80,6 +80,7 @@ using arrow::DataType;
 using arrow::Datum;
 using arrow::DecimalType;
 using arrow::default_memory_pool;
+using arrow::DictionaryArray;
 using arrow::ListArray;
 using arrow::PrimitiveArray;
 using arrow::ResizableBuffer;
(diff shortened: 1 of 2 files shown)


cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -4138,6 +4139,74 @@ TEST_P(TestArrowWriteDictionary, Statistics) {
...
+    // It's important there are no duplicate values in the dictionary, otherwise
+    // we trigger the WriteDense() code path which side-steps dictionary encoding.
+    std::shared_ptr<::arrow::Array> test_dictionary =
+        ArrayFromJSON(::arrow::utf8(), R"(["b", "c", "d", "a"])");
+    std::vector<std::shared_ptr<::arrow::Array>> test_indices = {
+        ArrayFromJSON(::arrow::int32(),
+                      R"([0, null, 3, 0, null, 3])"),  // ["b", null "a", "b", null, "a"]
+        ArrayFromJSON(
+            ::arrow::int32(),
+            R"([0, 3, null, 0, null, 1])")};  // ["b", "c", null, "b", "c", null]
```
westonpace, line 4160:
```suggestion
            R"([0, 1, null, 0, 1, null])")};  // ["b", "c", null, "b", "c", null]
```

I like what you have in the comment because then the min/max of row group 0 / chunk 0  is different from row group 0 / chunk 1.  Right now it looks like your indices don't match your comment and we have:

// ["b", null, "a", "b", null, "c"]

This leads to a/b being the min/max in stats0 but a/b is the min/max in both chunks of stats0.  To reproduce I think we want what you have in the comment which would mean chunk 0 is a/b and chunk 1 is b/c and so stats0 should be a/c.

westonpace (changes requested):
I'm not fully convinced the unit test has the correct test data (though I could be misreading it)

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
wjones127: You are correct; I updated the data but forgot the comment. I did a weird thing where the chunks are 6/6 but the row groups are 9/3, anticipating that hit more interesting conditions in the writer (but maybe this is unnecessary).

wjones127: So I like the data as is. The first row group contains the first chunk plus the first three rows of the next chunk. Then the second group contains the last three elements of the second chunk. I've updated the comment so it is accurate again.

westonpace: Write...so I think (but could be wrong) this would lead to three calls to `WriteArrowDictionary`:

Call #1: (no previous dictionary) min=a, max=b, nulls=2
Call #2: (previous dictionary is equal) min=a, max=b, nulls=1
Call #3: (no previous dictionary) min=b, max=c, nulls=1

So if the bug was still in place, and it was using the first chunk to determine row-group statistics, it would still get the correct answer in this case.

Admittedly, the null count would still be wrong (it would report 2 nulls for stat0), so the test case itself wouldn't pass with the old code.  But I think it would get further than it should.

wjones127: Ah yes I see now. You are correct (just verified in lldb). I will change it so call 1 and 2 will have a different max.

(Also funny to realize how much PR 1, 2, and 3 of this repo have been mentioned 🤣 )

westonpace (approved):
Thanks for looking into this!

PR #15180, opened 2023-01-04
GH-18481: [C++] prefer casting literal over casting field ref

### LINKED ISSUE #18481: [C++][Dataset] Allow more aggresive implicit casts for literals
After ARROW-8919, a literal in an Expression may cause unnecessary implicit casting of a column. For example `equal(field_ref("i8"), literal(1))` will cause column i8 to be promoted to the type of the literal (int32) for comparison. Since we have access to the literal value at bind time we could examine `1` and determine that it can safely be \*de\*moted to int8, which produces a semantically equivalent and more performant filter.

**Reporter**: [Ben Kietzman](https://issues.apache.org/jira/browse/ARROW-11402) / @bkietz
#### PRs and other links:
- [GitHub Pull Request #15180](https://github.com/apache/arrow/pull/15180)

<sub>**Note**: *This issue was originally created as [ARROW-11402](https://issues.apache.org/jira/browse/ARROW-11402). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

westonpace:
I ran into this problem while trying to work out partition pruning in the new scan node.  I feel like this is a somewhat naive approach but it seems to work.

I think it would fail if a `DispatchBest` existed where a n-ary kernel existed with non-equal types.  For example, if there was a function foo(int8, int32) and it had a dispatch best of some kind.
* Closes: #18481

### DIFF
--- a/cpp/src/arrow/compute/exec/expression.cc
+++ b/cpp/src/arrow/compute/exec/expression.cc
@@ -18,6 +18,7 @@
 #include "arrow/compute/exec/expression.h"
 
 #include <algorithm>
+#include <cmath>
 #include <optional>
 #include <unordered_map>
 #include <unordered_set>
@@ -221,7 +222,11 @@ bool Expression::Equals(const Expression& other) const {
   }
 
   if (auto lit = literal()) {
-    return lit->Equals(*other.literal());
+    // The scalar NaN is not equal to the scalar NaN but the literal NaN
+    // is equal to the literal NaN (e.g. the expressions are equal even if
+    // the values are not)
+    EqualOptions equal_options = EqualOptions::Defaults().nans_equal(true);
+    return lit->scalar()->Equals(other.literal()->scalar(), equal_options);
   }
 
   if (auto ref = field_ref()) {
@@ -442,7 +447,12 @@ TypeHolder SmallestTypeFor(const arrow::Datum& value) {
     }
     case Type::DOUBLE: {
       double doub = value.scalar_as<DoubleScalar>().value;
-      if (double(float(doub)) == doub) {
+      if (!std::isfinite(doub)) {
+        // Special values can be float
+        return float32();
+      }
+      // Test if float representation is the same
+      if (static_cast<double>(static_cast<float>(doub)) == doub) {
         return float32();
       }
       return float64();
--- a/cpp/src/arrow/compute/exec/expression.cc
+++ b/cpp/src/arrow/compute/exec/expression.cc
@@ -18,7 +18,6 @@
 #include "arrow/compute/exec/expression.h"
 
 #include <algorithm>
-#include <cmath>
 #include <optional>
 #include <unordered_map>
 #include <unordered_set>
--- a/cpp/src/arrow/compute/exec/expression.cc
+++ b/cpp/src/arrow/compute/exec/expression.cc
@@ -368,6 +368,153 @@ bool Expression::IsSatisfiable() const {
 
 namespace {
 
+TypeHolder SmallestTypeFor(const arrow::Datum& value) {
+  switch (value.type()->id()) {
+    case Type::INT8:
+      return int8();
+    case Type::UINT8:
+      return uint8();
+    case Type::INT16: {
+      int16_t i16 = value.scalar_as<Int16Scalar>().value;
+      if (i16 <= std::numeric_limits<int8_t>::max() &&
+          i16 >= std::numeric_limits<int8_t>::min()) {
+        return int8();
+      }
+      return int16();
+    }
+    case Type::UINT16: {
+      uint16_t ui16 = value.scalar_as<UInt16Scalar>().value;
+      if (ui16 <= std::numeric_limits<uint8_t>::max()) {
+        return uint8();
+      }
+      return uint16();
+    }
+    case Type::INT32: {
+      int32_t i32 = value.scalar_as<Int32Scalar>().value;
+      if (i32 <= std::numeric_limits<int8_t>::max() &&
+          i32 >= std::numeric_limits<int8_t>::min()) {
+        return int8();
+      }
+      if (i32 <= std::numeric_limits<int16_t>::max() &&
+          i32 >= std::numeric_limits<int16_t>::min()) {
+        return int16();
+      }
+      return int32();
+    }
+    case Type::UINT32: {
+      uint32_t ui32 = value.scalar_as<UInt32Scalar>().value;
+      if (ui32 <= std::numeric_limits<uint8_t>::max()) {
+        return uint8();
+      }
+      if (ui32 <= std::numeric_limits<uint16_t>::max()) {
+        return uint16();
+      }
+      return uint32();
+    }
+    case Type::INT64: {
+      int64_t i64 = value.scalar_as<Int64Scalar>().value;
+      if (i64 <= std::numeric_limits<int8_t>::max() &&
+          i64 >= std::numeric_limits<int8_t>::min()) {
+        return int8();
+      }
+      if (i64 <= std::numeric_limits<int16_t>::max() &&
+          i64 >= std::numeric_limits<int16_t>::min()) {
+        return int16();
+      }
+      if (i64 <= std::numeric_limits<int32_t>::max() &&
+          i64 >= std::numeric_limits<int32_t>::min()) {
+        return int32();
+      }
+      return int64();
+    }
+    case Type::UINT64: {
+      uint64_t ui64 = value.scalar_as<UInt64Scalar>().value;
+      if (ui64 <= std::numeric_limits<uint8_t>::max()) {
+        return uint8();
+      }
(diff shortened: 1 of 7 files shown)


cpp/src/arrow/compute/exec/expression.cc
```diff
@@ -368,6 +368,153 @@ bool Expression::IsSatisfiable() const {
 
 namespace {
 
+TypeHolder SmallestTypeFor(const arrow::Datum& value) {
```
westonpace, line 371: I feel like this is a rather naive approach but it seems to work.

cpp/src/arrow/compute/exec/expression.cc
```diff
@@ -368,6 +368,153 @@ bool Expression::IsSatisfiable() const {
...
+        return uint16();
+      }
+      if (ui64 <= std::numeric_limits<uint32_t>::max()) {
+        return uint32();
+      }
+      return uint64();
+    }
+    case Type::DOUBLE: {
+      double doub = value.scalar_as<DoubleScalar>().value;
+      if (double(float(doub)) == doub) {
```
westonpace, line 445: Is this the best way to determine if a double can be exactly represented by a float?

bkietz: This doesn't catch `nan`s. I'd recommend including an explicit clause for `nan`s and `inf`s. For the value itself, though, I think this roundtrip is reasonable.

bkietz: What do you think of checking `fmod(doub) == 0` and demoting to integer? This would simplify expressions like `i32 > 0.0` to an integer comparison. Perhaps not worth it?

cpp/src/arrow/compute/exec/expression_test.cc
```diff
@@ -593,8 +594,23 @@ TEST(Expression, BindWithImplicitCasts) {
     ExpectBindsTo(cmp(field_ref("dict_str"), field_ref("str")),
                   cmp(cast(field_ref("dict_str"), utf8()), field_ref("str")));
 
+    // Should prefer the literal
     ExpectBindsTo(cmp(field_ref("dict_i32"), literal(int64_t(4))),
-                  cmp(cast(field_ref("dict_i32"), int64()), literal(int64_t(4))));
+                  cmp(field_ref("dict_i32"), literal(int32_t(4))));
+    ExpectBindsTo(cmp(field_ref("dict_i32"), literal(int64_t(4))),
```
bkietz, line 600: Please add a case which demonstrates the behavior in the presence of unsigned integers. `SmallestTypeFor` preserves signedness information, which can produce some odd edge cases which we should be explicit about. For example:

```suggestion
    ExpectBindsTo(cmp(field_ref("i8"), literal(uint8_t(4))),
                  cmp(cast(field_ref("i8"), int16()), literal(int16_t(4))));
    ExpectBindsTo(cmp(field_ref("dict_i32"), literal(int64_t(4))),
```

bkietz (changes requested):
I think the general approach is acceptable, modulo a few nits.

For future reference, my own plan for this issue was to write this as a simplification pass. This approach is much simpler and only introduces a few minor edge cases. Nicely done

cpp/src/arrow/compute/exec/expression.cc
westonpace: At the moment I'm thinking no for floating->integral.  In general it should be easy enough to convert from one literal to another in just about any serialization of expressions (e.g. change 1.0 to 1).

westonpace: I've added a check for `nan`/`inf`

cpp/src/arrow/compute/exec/expression_test.cc
westonpace: Done.

westonpace:
I've addressed review.  In doing so I encountered a problem #33990 and addressed that here as well.

Closes #33990

westonpace:
The test failures revealed another issue which I addressed here.  In `DispatchBest` for simple arithmetic we were declaring `float32` to be the common type between `decimal` and `float32`.  However, I believe `float64` is a more accurate choice.  This also matches the [rule we use for trigonometric functions](https://github.com/apache/arrow/blob/apache-arrow-11.0.0/cpp/src/arrow/compute/kernels/scalar_arithmetic.cc#L749-L753).  CC @lidavidm for second opinion.

@nealrichardson there was another R test failure which I deemed inevitable and I have updated the test.  It is a slight behavior change.  Does this seem acceptable: https://github.com/apache/arrow/pull/15180/files#diff-45391fbe156c77f99e14090369faed8e11ec49583634fa1a51b634ef8b9eb5bf

nealrichardson:
Let me take a look. If this PR does what I think it's doing, there's probably a bunch of R code I can delete now that attempted the same.

lidavidm:
float64 sounds fine to me (I'm surprised it was float32 actually)

r/tests/testthat/test-type.R
```diff
@@ -280,7 +280,9 @@ test_that("infer_type() gets the right type for Expression", {
   expect_equal(y$type(), infer_type(y))
   expect_equal(infer_type(y), float64())
   expect_equal(add_xy$type(), infer_type(add_xy))
-  expect_equal(infer_type(add_xy), float64())
+  # even though 10 is a float64, arrow will clamp it to the narrowest
```
nealrichardson, line 283: This is only because both are scalars? If the float64 were corresponding to a field in the data, I would think that downcasting to float32 would be undesirable.

nealrichardson:
It would be good to run the macrobenchmarks on this. I would expect some speedup, mainly in the python queries (assuming we have any that would trigger this) since much of this is already handled in the R code, because we should be avoiding some big casts. Perhaps more importantly though, I ran into some issues with decimal types in #14553 and had to [exclude decimals from the type conversion](https://github.com/apache/arrow/pull/14553/files#diff-63c0f9982496837d6863f0aeaa0b2dff573175f2219ab0dcacd45da3c076c516R178-R188). I noticed this because a TPC-H query errored as a result of trying to keep operations on decimals.

r/tests/testthat/test-type.R
westonpace: Correct.

westonpace:
> It would be good to run the macrobenchmarks on this. I would expect some speedup

I will but I'm not certain we will get any speedup.

> mainly in the python queries (assuming we have any that would trigger this)

We don't.  The python queries are extremely basic.

> I ran into some issues with decimal types 

Yes.  In fact, for some of the simpler TPC-H queries, that conversion from decimal to double is a majority of the execution time.  Unfortunately, this PR does not fix that particular case because we don't support decimal arithmetic so a cast is inevitable.

The implementation is a bit subtle.  All of our non-decimal implicit casts go from "narrow type" to "wider type" (e.g. it is a safe cast).  For example, if we have `add<int8, int16>` it will become `add<int16, int16>`.  So by shrinking the literal as much as possible we ensure it is the one that will be cast (or avoid the cast).

With decimal implicit casts we actually go in the opposite direction.  We go from decimal (a wide type) to float64 (a more narrow type).  Since we don't shrink a float64 literal into a decimal literal we still end up casting the array and not the literal.

westonpace:
@ursabot please benchmark lang=R

nealrichardson:
I pulled this, removed the R code that tries to make scalar inputs match the type of the corresponding fields where safe and appropriate, and ran the tests. I was surprised to see that `field<int32> * 1<float64>` resulted in `float32`. The R code I would be removing checks that `1<float64>` can be represented as `int32` (the field's type) without loss, and if so, it casts it to match the field. It's fine with me if the acero code doesn't want to use that rule, but I found the result of `float32`, i.e. casting both inputs to something else, surprising.

westonpace:
> The R code I would be removing checks that 1<float64> can be represented as int32 (the field's type) without loss, and if so, it casts it to match the field.

@bkietz had suggested a similar rule.  I am a little bit unsure about switching over types like that but we can add it if we want.

> I found the result of float32, i.e. casting both inputs to something else, surprising.

I found this surprising as well (I added a test case for it to draw attention to it).  I think it is the correct thing to do though (although I may be biased as I think avoiding this situation would end up being tricky).  Plus, if a user wants, they can always explicitly cast to get their desired behavior.  Also, even if we adopted the above rule, this scenario could still occur with something like `field<int32> * 1.5<float64>` (1.5 can safely be represented as `float32` but not an integer).

Yes, you have the negative consequence that the result of that 1.5x might not be easily represented in `float32` for all values of `x` but that would be true if the right hand side were an array too.  In other words, if `field<int32> * 1<float64>` should be `float64` for precision purposes then one could argue that `field<int32> * field<float32>` should be `float64` with similar reasoning.  That is, if we are going to implicitly cast we should implicitly cast to the target that has the best chance of representing all results.

PR #15181, opened 2023-01-04
MINOR: Use http in preview-docs job summary

assignUser:
S3 does not support https so the link does not work atm.


assignUser:

assignUser:
+1

PR #15182, opened 2023-01-04
GH-15074: [Parquet][C++] change 16-bit page_ordinal to 32-bit

### LINKED ISSUE #15074: [Parquet][C++] Using int32_t to store page_ordinal

When a Page can be well compressed in `PLAIN` format, if the estimate size is much more larger than compressed size, the Page can be very small. And a 512MB row group may contains more than int16_t::max pages, causing it overflow in reader / writer.

Since parquet-mr uses `int` to store page_ordinal, why don't we use same int32_t? Or we can check-overflow when using the `page_ordinal`?

@pitrou 

### Component(s)

C++, Parquet

mapleFU:
As we mentioned in https://github.com/apache/arrow/issues/15074 . `int16_t page_ordinal` may causing overflow. So, we need to change it to 32-bit.

* [x] Implement the logic
* [x] Testing
  * [x] Upload a file with more than `int16_t` pages in parquet-testing.
* Closes: #15074

### DIFF
--- a/cpp/src/parquet/column_reader.cc
+++ b/cpp/src/parquet/column_reader.cc
@@ -261,7 +261,7 @@ class SerializedPageReader : public PageReader {
 
  private:
   void UpdateDecryption(const std::shared_ptr<Decryptor>& decryptor, int8_t module_type,
-                        const std::string& page_aad);
+                        std::string* page_aad);
 
   void InitDecryption();
 
@@ -329,8 +329,7 @@ void SerializedPageReader::InitDecryption() {
 }
 
 void SerializedPageReader::UpdateDecryption(const std::shared_ptr<Decryptor>& decryptor,
-                                            int8_t module_type,
-                                            const std::string& page_aad) {
+                                            int8_t module_type, std::string* page_aad) {
   DCHECK(decryptor != nullptr);
   if (crypto_ctx_.start_decrypt_with_dictionary_page) {
     std::string aad = encryption::CreateModuleAad(
@@ -338,8 +337,8 @@ void SerializedPageReader::UpdateDecryption(const std::shared_ptr<Decryptor>& de
         crypto_ctx_.column_ordinal, kNonPageOrdinal);
     decryptor->UpdateAad(aad);
   } else {
-    encryption::QuickUpdatePageAad(page_aad, page_ordinal_);
-    decryptor->UpdateAad(page_aad);
+    encryption::QuickUpdatePageAad(page_ordinal_, page_aad);
+    decryptor->UpdateAad(*page_aad);
   }
 }
 
@@ -366,7 +365,7 @@ std::shared_ptr<Page> SerializedPageReader::NextPage() {
       try {
         if (crypto_ctx_.meta_decryptor != nullptr) {
           UpdateDecryption(crypto_ctx_.meta_decryptor, encryption::kDictionaryPageHeader,
-                           data_page_header_aad_);
+                           &data_page_header_aad_);
         }
         deserializer.DeserializeMessage(reinterpret_cast<const uint8_t*>(view.data()),
                                         &header_size, &current_page_header_,
@@ -394,7 +393,7 @@ std::shared_ptr<Page> SerializedPageReader::NextPage() {
 
     if (crypto_ctx_.data_decryptor != nullptr) {
       UpdateDecryption(crypto_ctx_.data_decryptor, encryption::kDictionaryPage,
-                       data_page_aad_);
+                       &data_page_aad_);
     }
 
     // Read the compressed data page.
--- a/cpp/src/parquet/column_reader.cc
+++ b/cpp/src/parquet/column_reader.cc
@@ -291,7 +291,7 @@ class SerializedPageReader : public PageReader {
 
   // The ordinal fields in the context below are used for AAD suffix calculation.
   CryptoContext crypto_ctx_;
-  int16_t page_ordinal_;  // page ordinal does not count the dictionary page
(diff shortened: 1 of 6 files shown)


cpp/src/parquet/encryption/encryption_internal.cc
```diff
@@ -599,7 +599,13 @@ static std::string ShortToBytesLe(int16_t input) {
 
 std::string CreateModuleAad(const std::string& file_aad, int8_t module_type,
                             int16_t row_group_ordinal, int16_t column_ordinal,
-                            int16_t page_ordinal) {
+                            int32_t page_ordinal) {
+  if (ARROW_PREDICT_FALSE(page_ordinal > std::numeric_limits<int16_t>::max())) {
```
wgtmac, line 603: I'd suggest explicitly explain that it is the restriction by AAD suffix of page ordinal: https://github.com/apache/parquet-format/blob/master/Encryption.md#442-aad-suffix

mapleFU: Okay. The standard seems to be ambiguous. The format doesn't says that `page cannot larger than i16`. But AAD requires it. I follow the implemention in parquet-mr: If the page not use AAD, larger than i16 is ok. Otherwise, we cannot use it.

By the way, seems compiler flag not use `no-narrowing`, which may cause some possible narrow-cast issue.

mapleFU:
By the way, it's so hard for current C++ code to generate a file with more than INT16_MAX pages. But it's possible in parquet-mr, because it can split page by rows.

cpp/src/parquet/encryption/encryption_internal.cc
mapleFU: @pitrou Whats your opinion? I just follow the Java implementions here.

mapleFU:
Testing is done, patch can be reviewed now.

cpp/src/parquet/reader_test.cc
```diff
@@ -1015,4 +1017,40 @@ INSTANTIATE_TEST_SUITE_P(Lz4CodecTests, TestCodec, ::testing::ValuesIn(test_code
                          testing::PrintToStringParamName());
 #endif  // ARROW_WITH_LZ4
 
+TEST(TestFileReader, TestOverflowI16) {
+  ReaderProperties reader_props;
```
pitrou, line 1021: Can you add a comment pointing to the GH issue number?

cpp/src/parquet/reader_test.cc
```diff
@@ -1015,4 +1017,40 @@ INSTANTIATE_TEST_SUITE_P(Lz4CodecTests, TestCodec, ::testing::ValuesIn(test_code
...
+  EXPECT_EQ(1, metadata_ptr->num_columns());
+  auto row_group = file_reader->RowGroup(0);
+
+  {
+    auto column_reader =
+        std::dynamic_pointer_cast<TypedColumnReader<BooleanType>>(row_group->Column(0));
+    EXPECT_NE(nullptr, column_reader);
+    std::array<bool, 1024> boolean_values{};
+    int64_t value_sum = 0;
+    while (value_sum < 40000) {
```
pitrou, line 1035: Instead of hardcoding 40000 here, can you just read the column until the end?

cpp/src/parquet/reader_test.cc
```diff
@@ -1015,4 +1017,40 @@ INSTANTIATE_TEST_SUITE_P(Lz4CodecTests, TestCodec, ::testing::ValuesIn(test_code
...
+      int64_t value_num = 0;
+      column_reader->ReadBatch(/* batch_size= */ 1024, nullptr, nullptr,
```
pitrou, line 1037:
```suggestion
      column_reader->ReadBatch(/*batch_size=*/ kBatchLength, nullptr, nullptr,
```

mapleFU:
These tests are so flackey...Never get a all-test-pass CI these days...

pitrou (approved):
Thanks @mapleFU . I did a couple changes and will merge this if CI is green.

mapleFU:
Still one failed, lol

cpp/src/parquet/column_reader.cc
```diff
@@ -329,17 +329,16 @@ void SerializedPageReader::InitDecryption() {
...
   if (crypto_ctx_.start_decrypt_with_dictionary_page) {
     std::string aad = encryption::CreateModuleAad(
         decryptor->file_aad(), module_type, crypto_ctx_.row_group_ordinal,
         crypto_ctx_.column_ordinal, kNonPageOrdinal);
     decryptor->UpdateAad(aad);
   } else {
-    encryption::QuickUpdatePageAad(page_aad, page_ordinal_);
-    decryptor->UpdateAad(page_aad);
+    encryption::QuickUpdatePageAad(page_ordinal_, page_aad);
+    decryptor->UpdateAad(*page_aad);
```
mapleFU, line 341: Was all `const std::string&` changed to `std::string*` because of aad? @pitrou

mapleFU: Since I didn't find it neccessary to change to mutable pointer, why do you did this change? Would `page_aad` be nullptr?

mapleFU: @pitrou

pitrou: See https://github.com/apache/arrow/pull/15182/commits/8089b4e641d96b3dbceae512d15535efd4f68a31 : the `const std::string&` was actually mutated by the AAD functions!

PR #15183, opened 2023-01-04
GH-15151: [C++] ]Adding RecordBatchReaderSource to solve an issue in R API

### LINKED ISSUE #15151: Should `GetIOThreadPool()` be accessible from installed headers?

In #14582 it was found that using the CPU thread pool in `arrow::compute::MakeReaderGenerator()` caused problems when the number of CPU threads was limited (as it often is on CI machines with few available cores). The solution was to use the IO thread pool for this; however, `arrow::io::internal::GetIOThreadPool()` is not available in any installed headers. I don't know what the best way to make this available would be (or whether creating a source node from a record batch reader should be baked into the internals somewhere); however, my hack of:

```r
namespace arrow {
namespace io {
namespace internal {
arrow::internal::ThreadPool* GetIOThreadPool();
}
}  // namespace io
}  // namespace arrow
```

...in the R package should almost certainly not exist.

### Component(s)

C++

vibhatha:
This PR includes the factory `record_batch_reader_source` for the Acero. This is a source node which takes in a `RecordBatchReader` as the data source along an executor which gives the freedom to choose the threadpool required for the execution. Also an example shows how this can be used in Acero.

- [x] Self-review
* Closes: #15151

### DIFF
--- a/cpp/examples/arrow/execution_plan_documentation_examples.cc
+++ b/cpp/examples/arrow/execution_plan_documentation_examples.cc
@@ -784,10 +784,9 @@ arrow::Status RecordBatchReaderSourceSinkExample() {
   std::shared_ptr<arrow::RecordBatchReader> reader =
       std::make_shared<arrow::TableBatchReader>(table);
 
-  ARROW_ASSIGN_OR_RAISE(
-      cp::ExecNode * source,
-      cp::MakeExecNode("record_batch_reader_source", plan.get(), {},
-                       cp::RecordBatchReaderSourceNodeOptions{table->schema(), reader}));
+  ARROW_ASSIGN_OR_RAISE(cp::ExecNode * source,
+                        cp::MakeExecNode("record_batch_reader_source", plan.get(), {},
+                                         cp::RecordBatchReaderSourceNodeOptions{reader}));
   ARROW_RETURN_NOT_OK(cp::MakeExecNode(
       "order_by_sink", plan.get(), {source},
       cp::OrderBySinkNodeOptions{
--- a/cpp/examples/arrow/execution_plan_documentation_examples.cc
+++ b/cpp/examples/arrow/execution_plan_documentation_examples.cc
@@ -761,6 +761,41 @@ arrow::Status TableSinkExample() {
   std::cout << "Results : " << output_table->ToString() << std::endl;
   return arrow::Status::OK();
 }
+
+// (Doc section: Table Sink Example)
+
+// (Doc section: RecordBatchReaderSource Example)
+
+/// \brief An example showing the usage of a RecordBatchReader as the data source.
+///
+/// RecordBatchReaderSourceSink Example
+/// This example shows how a record_batch_reader_source can be used
+/// in an execution plan. This includes the source node
+/// receiving data from a TableRecordBatchReader.
+
+arrow::Status RecordBatchReaderSourceSinkExample() {
+  ARROW_ASSIGN_OR_RAISE(std::shared_ptr<cp::ExecPlan> plan,
+                        cp::ExecPlan::Make(*cp::threaded_exec_context()));
+
+  std::cout << "basic data created" << std::endl;
+
+  arrow::AsyncGenerator<std::optional<cp::ExecBatch>> sink_gen;
+  ARROW_ASSIGN_OR_RAISE(auto table, GetTable());
+  std::shared_ptr<arrow::RecordBatchReader> reader =
+      std::make_shared<arrow::TableBatchReader>(table);
+
+  ARROW_ASSIGN_OR_RAISE(
+      cp::ExecNode * source,
+      cp::MakeExecNode("record_batch_reader_source", plan.get(), {},
+                       cp::RecordBatchReaderSourceNodeOptions{table->schema(), reader}));
+  ARROW_RETURN_NOT_OK(cp::MakeExecNode(
+      "order_by_sink", plan.get(), {source},
+      cp::OrderBySinkNodeOptions{
+          cp::SortOptions{{cp::SortKey{"a", cp::SortOrder::Descending}}}, &sink_gen}));
+
+  return ExecutePlanAndCollectAsTableWithCustomSink(plan, table->schema(), sink_gen);
+}
+
 // (Doc section: Table Sink Example)
 
 enum ExampleMode {
@@ -777,7 +812,8 @@ enum ExampleMode {
   KSELECT = 10,
   WRITE = 11,
   UNION = 12,
-  TABLE_SOURCE_TABLE_SINK = 13
+  TABLE_SOURCE_TABLE_SINK = 13,
+  RECORD_BATCH_READER_SOURCE = 14
 };
 
 int main(int argc, char** argv) {
@@ -847,6 +883,10 @@ int main(int argc, char** argv) {
     case TABLE_SOURCE_TABLE_SINK:
       PrintBlock("TableSink Example");
       status = TableSinkExample();
+      break;
+    case RECORD_BATCH_READER_SOURCE:
+      PrintBlock("RecordBatchReaderSource Example");
+      status = RecordBatchReaderSourceSinkExample();
       break;
     default:
       break;
--- a/cpp/examples/arrow/execution_plan_documentation_examples.cc
+++ b/cpp/examples/arrow/execution_plan_documentation_examples.cc
@@ -772,9 +772,6 @@ arrow::Status TableSinkExample() {
 /// receiving data from a TableRecordBatchReader.
 
 arrow::Status RecordBatchReaderSourceSinkExample() {
-  ARROW_ASSIGN_OR_RAISE(std::shared_ptr<cp::ExecPlan> plan,
-                        cp::ExecPlan::Make(*cp::threaded_exec_context()));
-
   ARROW_ASSIGN_OR_RAISE(auto table, GetTable());
   std::shared_ptr<arrow::RecordBatchReader> reader =
       std::make_shared<arrow::TableBatchReader>(table);
@@ -783,7 +780,7 @@ arrow::Status RecordBatchReaderSourceSinkExample() {
   return ExecutePlanAndCollectAsTable(std::move(reader_source));
 }
 
-// (Doc section: Table Sink Example)
+// (Doc section: RecordBatchReaderSource Example)
 
 enum ExampleMode {
   SOURCE_SINK = 0,
--- a/cpp/examples/arrow/execution_plan_documentation_examples.cc
+++ b/cpp/examples/arrow/execution_plan_documentation_examples.cc
@@ -561,8 +561,6 @@ arrow::Status SourceOrderBySinkExample() {
 
   ARROW_ASSIGN_OR_RAISE(auto basic_data, MakeSortTestBasicBatches());
 
-  std::cout << "basic data created" << std::endl;
-
   arrow::AsyncGenerator<std::optional<cp::ExecBatch>> sink_gen;
 
   auto source_node_options = cp::SourceNodeOptions{basic_data.schema, basic_data.gen()};
@@ -777,22 +775,12 @@ arrow::Status RecordBatchReaderSourceSinkExample() {
   ARROW_ASSIGN_OR_RAISE(std::shared_ptr<cp::ExecPlan> plan,
                         cp::ExecPlan::Make(*cp::threaded_exec_context()));
 
-  std::cout << "basic data created" << std::endl;
-
-  arrow::AsyncGenerator<std::optional<cp::ExecBatch>> sink_gen;
   ARROW_ASSIGN_OR_RAISE(auto table, GetTable());
   std::shared_ptr<arrow::RecordBatchReader> reader =
       std::make_shared<arrow::TableBatchReader>(table);
-
-  ARROW_ASSIGN_OR_RAISE(cp::ExecNode * source,
-                        cp::MakeExecNode("record_batch_reader_source", plan.get(), {},
-                                         cp::RecordBatchReaderSourceNodeOptions{reader}));
-  ARROW_RETURN_NOT_OK(cp::MakeExecNode(
-      "order_by_sink", plan.get(), {source},
-      cp::OrderBySinkNodeOptions{
-          cp::SortOptions{{cp::SortKey{"a", cp::SortOrder::Descending}}}, &sink_gen}));
-
-  return ExecutePlanAndCollectAsTableWithCustomSink(plan, table->schema(), sink_gen);
+  cp::Declaration reader_source{"record_batch_reader_source",
+                                cp::RecordBatchReaderSourceNodeOptions{reader}};
+  return ExecutePlanAndCollectAsTable(std::move(reader_source));
 }
 
 // (Doc section: Table Sink Example)
--- a/cpp/src/arrow/compute/exec/options.h
+++ b/cpp/src/arrow/compute/exec/options.h
@@ -128,13 +128,9 @@ class ARROW_EXPORT SchemaSourceNodeOptions : public ExecNodeOptions {
 
 class ARROW_EXPORT RecordBatchReaderSourceNodeOptions : public ExecNodeOptions {
  public:
-  RecordBatchReaderSourceNodeOptions(std::shared_ptr<Schema> schema,
-                                     std::shared_ptr<RecordBatchReader> reader,
+  RecordBatchReaderSourceNodeOptions(std::shared_ptr<RecordBatchReader> reader,
                                      arrow::internal::Executor* io_executor = NULLPTR)
-      : schema(schema), reader(std::move(reader)), io_executor(io_executor) {}
-
-  /// \brief The schema of the record batches from the iterator
-  std::shared_ptr<Schema> schema;
+      : reader(std::move(reader)), io_executor(io_executor) {}
 
   /// \brief The RecordBatchReader which acts as the data source
   std::shared_ptr<RecordBatchReader> reader;
--- a/cpp/src/arrow/compute/exec/options.h
+++ b/cpp/src/arrow/compute/exec/options.h
@@ -136,10 +136,10 @@ class ARROW_EXPORT RecordBatchReaderSourceNodeOptions : public ExecNodeOptions {
   /// \brief The schema of the record batches from the iterator
   std::shared_ptr<Schema> schema;
 
-  /// \brief A maker of an iterator which acts as the data source
+  /// \brief The RecordBatchReader which acts as the data source
   std::shared_ptr<RecordBatchReader> reader;
 
-  /// \brief The executor to use for scanning the iterator
+  /// \brief The executor to use for the reader
   ///
   /// Defaults to the default I/O executor.
   arrow::internal::Executor* io_executor;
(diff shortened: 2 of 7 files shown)


vibhatha:
cc @westonpace @paleolimbot could you please take a look.

paleolimbot (commented):
Awesome! I *think* this will fix my problem, but could you also update the R package code? That will tell you in a hurry whether or not this works. I tried:

```cpp
// [[arrow::export]]
std::shared_ptr<compute::ExecNode> ExecNode_SourceNode(
    const std::shared_ptr<compute::ExecPlan>& plan,
    const std::shared_ptr<arrow::RecordBatchReader>& reader) {
  arrow::compute::RecordBatchReaderSourceNodeOptions options{reader->schema(), reader};
  return MakeExecNodeOrStop("record_batch_reader_source", plan.get(), {}, options);
}
```

Replacing

https://github.com/apache/arrow/blob/63b91cc1f7131356537ab9cbb84ed108d6f9102e/r/src/compute-exec.cpp#L459-L468

...and that seemed to work. Maybe also remove

https://github.com/apache/arrow/blob/63b91cc1f7131356537ab9cbb84ed108d6f9102e/r/src/compute-exec.cpp#L33-L40

while you're there?

cpp/src/arrow/compute/exec/options.h
```diff
@@ -126,6 +126,25 @@ class ARROW_EXPORT SchemaSourceNodeOptions : public ExecNodeOptions {
...
 
+class ARROW_EXPORT RecordBatchReaderSourceNodeOptions : public ExecNodeOptions {
+ public:
+  RecordBatchReaderSourceNodeOptions(std::shared_ptr<Schema> schema,
+                                     std::shared_ptr<RecordBatchReader> reader,
+                                     arrow::internal::Executor* io_executor = NULLPTR)
+      : schema(schema), reader(std::move(reader)), io_executor(io_executor) {}
+
+  /// \brief The schema of the record batches from the iterator
+  std::shared_ptr<Schema> schema;
```
westonpace, line 137: `RecordBatchReader` has a `schema()` method. We should use that instead of requiring it to be specified separately.

cpp/src/arrow/compute/exec/plan_test.cc
```diff
@@ -344,6 +344,47 @@ void TestSourceSink(
...
+                {
+                    {"record_batch_reader_source",
+                     RecordBatchReaderSourceNodeOptions{exp_batches.schema, reader}},
+                    {"sink", SinkNodeOptions{&sink_gen}},
+                })
+                .AddToPlan(plan.get()));
+
+  ASSERT_THAT(StartAndCollect(plan.get(), sink_gen),
+              Finishes(ResultWith(UnorderedElementsAreArray(exp_batches.batches))));
+}
```
westonpace, line 368: You are specifying an executor but you aren't verifying that it is being used.  Also, there is no test where we do not specify an executor.

Testing that the executor gets used would be a little tricky (maybe someday we can add a `TotalTasksRun` method to the executor).  You would probably need to wrap the batch generator with something that verifies it is on the I/O executor (e.g. using OwnsThisThread or something similar).

I am probably being a bit harsh here as I suspect we test some of the other sources in a similar fashion.  If we're only going to have one simple test I would prefer a test using the default instead of one specifying a custom executor as that is the more common case.

cpp/src/arrow/compute/exec/test_util.h
```diff
@@ -127,6 +127,10 @@ ARROW_TESTING_EXPORT
 Result<std::vector<std::shared_ptr<RecordBatch>>> ToRecordBatches(
     const BatchesWithSchema& batches);
 
+ARROW_TESTING_EXPORT
+Result<std::shared_ptr<RecordBatchReader>> ToRecordBatcheReader(
+    const BatchesWithSchema& batches_with_schema);
+
```
westonpace, line 133:
```suggestion
ARROW_TESTING_EXPORT
Result<std::shared_ptr<RecordBatchReader>> ToRecordBatchReader(
    const BatchesWithSchema& batches_with_schema);

```

cpp/src/arrow/compute/exec/test_util.cc
```diff
@@ -290,6 +290,18 @@ Result<std::vector<std::shared_ptr<RecordBatch>>> ToRecordBatches(
   return record_batches;
 }
 
+Result<std::shared_ptr<RecordBatchReader>> ToRecordBatcheReader(
```
westonpace, line 293:
```suggestion
Result<std::shared_ptr<RecordBatchReader>> ToRecordBatchReader(
```

cpp/src/arrow/compute/exec/source_node.cc
```diff
@@ -327,6 +327,55 @@ struct SchemaSourceNode : public SourceNode {
...
+    const auto& cast_options =
+        checked_cast<const RecordBatchReaderSourceNodeOptions&>(options);
+    auto& reader = cast_options.reader;
+    auto& schema = cast_options.schema;
+    auto io_executor = cast_options.io_executor;
+
+    if (io_executor == NULLPTR) {
+      io_executor = plan->query_context()->exec_context()->executor();
+    }
+
```
westonpace, line 347:
```suggestion
```
This will initialize the io_executor to the CPU executor, which we do not want.  You already have an `if (io_executor == NULLPTR)` block below

cpp/src/arrow/compute/exec/plan_test.cc
```diff
@@ -344,6 +344,47 @@ void TestSourceSink(
...
+  ASSERT_OK(Declaration::Sequence(
+                {
+                    {"record_batch_reader_source",
+                     RecordBatchReaderSourceNodeOptions{exp_batches.schema, reader}},
+                    {"sink", SinkNodeOptions{&sink_gen}},
+                })
+                .AddToPlan(plan.get()));
+
+  ASSERT_THAT(StartAndCollect(plan.get(), sink_gen),
+              Finishes(ResultWith(UnorderedElementsAreArray(exp_batches.batches))));
```
westonpace, line 367: Let's use `DeclarationToExecBatches` for new tests (feel free to fix existing tests too).

westonpace (changes requested):
Thanks for doing this.  I have a few suggestions.

cpp/src/arrow/compute/exec/plan_test.cc
vibhatha: @westonpace I am working on a wrapper to address this. Let me evaluate options and update the PR. Also, I wouldn't say this is `harsh` but `thorough` as I see other sources are not properly tested like this. I think it would be better to test them too. I will create a ticket to cover this. Thanks for thinking in this angel.

vibhatha: By the way `TotalTasksRun` sounds fun, could you elaborate a bit what you expect here, I am happy to work on that task.

vibhatha: Probably external to this PR. Appreciate your thoughts.

cpp/src/arrow/compute/exec/plan_test.cc
vibhatha: Would you mind, if I do this in another PR for the rest of the items?

westonpace: I don't mind if you fix the other tests in a separate PR.  However, any newly added tests should follow the new format.  `TEST(ExecPlanExecution, SourceGroupedSum)` is a good example from this file that you can follow.  This test, for example, would become something like:

```
  auto exp_batches = MakeBasicBatches();
  ASSERT_OK_AND_ASSIGN(std::shared_ptr<RecordBatchReader> reader, to_reader(exp_batches));
  Declaration plan("record_batch_reader_source", RecordBatchReaderSourceNodeOptions{exp_batches.schema, reader});
  ASSERT_OK_AND_ASSIGN(auto out_batches, DeclarationToExecBatches(std::move(plan)));
  AssertExecBatchesEqualIgnoringOrder(exp_batches.schema, exp_batches.batches, out_batches);
```

cpp/src/arrow/compute/exec/plan_test.cc
westonpace: > Are you referring to IOThreadPool being used as the default instead of NULLPTR?

What you had before might actually have been correct (I thought you were using `executor` in the options but on closer examination it was only being used in the exec_ctx).  I was saying the most common case is `RecordBatchReaderSourceNodeOptions options(reader)` where the executor is not specified and is left to be the default.  We should have our test verify that this case works correctly.  At the very least we should make sure we can run such a plan without crashing.  Long term we should check, after we ran the plan, to ensure that there were tasks submitted to the I/O thread pool.

This kind of test should not need to create a thread pool at all.  We should only need to create a thread pool when testing a custom executor in the node.

vibhatha: I updated the code, is it okay?

paleolimbot (approved):
From the R side of things...looks great! Thank you!

cpp/src/arrow/compute/exec/plan_test.cc
```diff
@@ -344,6 +344,40 @@ void TestSourceSink(
...
+    std::function<Result<std::shared_ptr<RecordBatchReader>>(const BatchesWithSchema&)>
+        to_reader) {
+  for (bool parallel : {false, true}) {
+    SCOPED_TRACE(parallel ? "parallel/merged" : "serial");
+    auto exp_batches = MakeBasicBatches();
+    ASSERT_OK_AND_ASSIGN(std::shared_ptr<RecordBatchReader> reader,
+                         to_reader(exp_batches));
+    RecordBatchReaderSourceNodeOptions options{reader};
+    Declaration plan("record_batch_reader_source", std::move(options));
+    ASSERT_OK_AND_ASSIGN(auto result, DeclarationToExecBatches(plan, false));
```
westonpace, line 357:
```suggestion
    ASSERT_OK_AND_ASSIGN(auto result, DeclarationToExecBatches(plan, parallel));
```

cpp/src/arrow/compute/exec/source_node.cc
```diff
@@ -327,6 +327,53 @@ struct SchemaSourceNode : public SourceNode {
...
+  static Result<ExecNode*> Make(ExecPlan* plan, std::vector<ExecNode*> inputs,
+                                const ExecNodeOptions& options) {
+    RETURN_NOT_OK(ValidateExecNodeInputs(plan, inputs, 0, kKindName));
+    const auto& cast_options =
+        checked_cast<const RecordBatchReaderSourceNodeOptions&>(options);
+    auto& reader = cast_options.reader;
+    auto io_executor = cast_options.io_executor;
+
+    if (reader == NULLPTR) {
```
westonpace, line 343:
```suggestion
    if (reader == nullptr) {
```

cpp/src/arrow/compute/exec/source_node.cc
```diff
@@ -327,6 +327,53 @@ struct SchemaSourceNode : public SourceNode {
...
+      return Status::Invalid(kKindName, " requires a reader which is not null");
+    }
+
+    if (io_executor == NULLPTR) {
```
westonpace, line 347:
```suggestion
    if (io_executor == nullptr) {
```

cpp/src/arrow/compute/exec/source_node.cc
```diff
@@ -327,6 +327,53 @@ struct SchemaSourceNode : public SourceNode {
...
+                                                          generator);
+  }
+
+  static Result<arrow::AsyncGenerator<std::optional<ExecBatch>>> MakeGenerator(
+      const std::shared_ptr<RecordBatchReader>& reader,
+      arrow::internal::Executor* io_executor) {
+    const auto& schema = reader->schema();
+    auto to_exec_batch =
+        [schema](const std::shared_ptr<RecordBatch>& batch) -> std::optional<ExecBatch> {
+      if (batch == NULLPTR || *batch->schema() != *schema) {
```
westonpace, line 362:
```suggestion
      if (batch == nullptr) {
```
I don't believe we should need to make sure the batches produced by the record batch reader have the correct schema.

cpp/src/arrow/compute/exec/source_node.cc
```diff
@@ -327,6 +327,53 @@ struct SchemaSourceNode : public SourceNode {
...
+
+    ARROW_ASSIGN_OR_RAISE(auto generator, MakeGenerator(reader, io_executor));
+    return plan->EmplaceNode<RecordBatchReaderSourceNode>(plan, reader->schema(),
+                                                          generator);
+  }
+
+  static Result<arrow::AsyncGenerator<std::optional<ExecBatch>>> MakeGenerator(
+      const std::shared_ptr<RecordBatchReader>& reader,
+      arrow::internal::Executor* io_executor) {
+    const auto& schema = reader->schema();
```
westonpace, line 359:
```suggestion
```

westonpace (changes requested):
A few more suggestions

cpp/examples/arrow/execution_plan_documentation_examples.cc
```diff
@@ -761,6 +761,40 @@ arrow::Status TableSinkExample() {
...
+/// RecordBatchReaderSourceSink Example
+/// This example shows how a record_batch_reader_source can be used
+/// in an execution plan. This includes the source node
+/// receiving data from a TableRecordBatchReader.
+
+arrow::Status RecordBatchReaderSourceSinkExample() {
+  ARROW_ASSIGN_OR_RAISE(std::shared_ptr<cp::ExecPlan> plan,
+                        cp::ExecPlan::Make(*cp::threaded_exec_context()));
+
+  std::cout << "basic data created" << std::endl;
```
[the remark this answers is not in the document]
vibhatha: You're correct.

cpp/examples/arrow/execution_plan_documentation_examples.cc
```diff
@@ -761,6 +759,30 @@ arrow::Status TableSinkExample() {
   std::cout << "Results : " << output_table->ToString() << std::endl;
   return arrow::Status::OK();
 }
+
+// (Doc section: Table Sink Example)
+
```
westonpace, line 764:
```suggestion
```

cpp/examples/arrow/execution_plan_documentation_examples.cc
```diff
@@ -761,6 +759,30 @@ arrow::Status TableSinkExample() {
...
+///
+/// RecordBatchReaderSourceSink Example
+/// This example shows how a record_batch_reader_source can be used
+/// in an execution plan. This includes the source node
+/// receiving data from a TableRecordBatchReader.
+
+arrow::Status RecordBatchReaderSourceSinkExample() {
+  ARROW_ASSIGN_OR_RAISE(std::shared_ptr<cp::ExecPlan> plan,
+                        cp::ExecPlan::Make(*cp::threaded_exec_context()));
+
```
westonpace, line 777:
```suggestion
```

cpp/src/arrow/compute/exec/plan_test.cc
```diff
@@ -344,6 +344,40 @@ void TestSourceSink(
...
+    AssertExecBatchesEqualIgnoringOrder(result.schema, result.batches,
+                                        exp_batches.batches);
+  }
+}
+
+void TestRecordBatchReaderSourceSinkError(
+    std::function<Result<std::shared_ptr<RecordBatchReader>>(const BatchesWithSchema&)>
+        to_reader) {
+  ASSERT_OK_AND_ASSIGN(auto plan, ExecPlan::Make());
+  std::shared_ptr<Schema> no_schema;
```
westonpace, line 367:
```suggestion
```

cpp/src/arrow/compute/exec/test_util.cc
```diff
@@ -290,6 +290,18 @@ Result<std::vector<std::shared_ptr<RecordBatch>>> ToRecordBatches(
   return record_batches;
 }
 
+Result<std::shared_ptr<RecordBatchReader>> ToRecordBatchReader(
+    const BatchesWithSchema& batches_with_schema) {
+  std::vector<std::shared_ptr<RecordBatch>> record_batches;
+  for (auto batch : batches_with_schema.batches) {
+    ARROW_ASSIGN_OR_RAISE(auto record_batch,
+                          batch.ToRecordBatch(batches_with_schema.schema));
+    record_batches.push_back(record_batch);
```
westonpace, line 299:
```suggestion
    record_batches.push_back(std::move(record_batch));
```

westonpace (changes requested):
A final few cleanup nits and then we can merge.

cpp/examples/arrow/execution_plan_documentation_examples.cc
vibhatha: Ah, shouldn't we put this in either side of the code-segment that we want to refer to?

For instance

```c++
// (Doc section: Table Sink Example)

/// Code refer to

// (Doc section: Table Sink Example)
```
Even if this is the correct way to do, there is a typo in my code in the `RecordBatchReaderSource` section.

`// (Doc section: RecordBatchReaderSource Example)`, I haven't closed that properly. I did close it now.

westonpace: Yes, sorry.  I hadn't realized this was missing before.  You are correct about the Table Sink Example part.

vibhatha:
@westonpace btw, I just searched for places where I could have missed `move` and found out the following in other sections of the code base. I haven't included it in here yet, but here is the diff (I tested locally and works fine). 

```
diff --git a/cpp/src/arrow/compute/exec/test_util.cc b/cpp/src/arrow/compute/exec/test_util.cc
index 038171c84..912428acf 100644
--- a/cpp/src/arrow/compute/exec/test_util.cc
+++ b/cpp/src/arrow/compute/exec/test_util.cc
@@ -273,8 +273,8 @@ Result<std::vector<std::shared_ptr<ExecBatch>>> ToExecBatches(
     const BatchesWithSchema& batches_with_schema) {
   std::vector<std::shared_ptr<ExecBatch>> exec_batches;
   for (auto batch : batches_with_schema.batches) {
-    auto exec_batch = std::make_shared<ExecBatch>(batch);
-    exec_batches.push_back(exec_batch);
+    //auto exec_batch = std::make_shared<ExecBatch>(batch);
+    exec_batches.push_back(std::make_shared<ExecBatch>(batch));
   }
   return exec_batches;
 }
@@ -285,7 +285,7 @@ Result<std::vector<std::shared_ptr<RecordBatch>>> ToRecordBatches(
   for (auto batch : batches_with_schema.batches) {
     ARROW_ASSIGN_OR_RAISE(auto record_batch,
                           batch.ToRecordBatch(batches_with_schema.schema));
-    record_batches.push_back(record_batch);
+    record_batches.push_back(std::move(record_batch));
   }
   return record_batches;
 }
```

Is this acceptable?

cpp/src/arrow/compute/exec/plan_test.cc
vibhatha: The follow up PR is WIP here: https://github.com/apache/arrow/pull/15288

westonpace:
> Is this acceptable?

Yes, both of those changes look correct.  Do you want to apply them real quick?

vibhatha:
> > Is this acceptable?
> 
> Yes, both of those changes look correct. Do you want to apply them real quick?

Yes I will. One minute.

vibhatha:
@westonpace Done. Ready to review.

westonpace (approved):
Thanks for looking into this.  I'll merge when CI passes.

vibhatha:
@westonpace CIs passed.

PR #15184, opened 2023-01-04
GH-15185: [C++][Parquet] Improve documentation for Parquet Reader column_indices

### LINKED ISSUE #15185: Improvement for the documentation of Parquet reader

Improve the comments so that it is clearer to which level the column_indexes refer to in the different reader methods.

### Component(s)

C++, Documentation, Parquet

LouisClt:
This aims to fix the documentation and improve it,, by better specifying to what level the column_indices given in argument refer to.
* Closes: #15185

### DIFF
--- a/cpp/src/parquet/arrow/reader.h
+++ b/cpp/src/parquet/arrow/reader.h
@@ -133,6 +133,7 @@ class PARQUET_EXPORT FileReader {
   // fully-materialized arrow::Array instances
   //
   // Returns error status if the column of interest is not flat.
+  // The indicated column index is relative to the schema
   virtual ::arrow::Status GetColumn(int i, std::unique_ptr<ColumnReader>* out) = 0;
 
   /// \brief Return arrow schema for all the columns.
@@ -225,7 +226,19 @@ class PARQUET_EXPORT FileReader {
 
   /// \brief Read the given columns into a Table
   ///
-  /// The indicated column indices are relative to the schema
+  /// The indicated column indices are relative to the internal representation
+  /// of the parquet table. For instance :
+  /// 0 foo.bar
+  ///       foo.bar.baz           0
+  ///       foo.bar.baz2          1
+  ///   foo.qux                   2
+  /// 1 foo2                      3
+  /// 2 foo3                      4
+  ///
+  /// i=0 will read foo.bar.baz, i=1 will read only foo.bar.baz2 and so on.
+  /// To retrive all the indices corresponding to the upper level arrow schema,
+  /// one can use manifest().schema_fields and get all the indices of the leaves
+  /// for each particular upper level field index.
   virtual ::arrow::Status ReadTable(const std::vector<int>& column_indices,
                                     std::shared_ptr<::arrow::Table>* out) = 0;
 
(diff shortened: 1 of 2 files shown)


cpp/src/parquet/arrow/reader.h
```diff
@@ -225,7 +226,19 @@ class PARQUET_EXPORT FileReader {
...
+  ///       foo.bar.baz           0
+  ///       foo.bar.baz2          1
+  ///   foo.qux                   2
+  /// 1 foo2                      3
+  /// 2 foo3                      4
+  ///
+  /// i=0 will read foo.bar.baz, i=1 will read only foo.bar.baz2 and so on.
+  /// To retrive all the indices corresponding to the upper level arrow schema,
+  /// one can use manifest().schema_fields and get all the indices of the leaves
+  /// for each particular upper level field index.
```
wjones127, line 241: What do you think of this?

```suggestion
  /// Only leaf fields have indices; foo itself doesn't have an index.
  /// To get the index for a particular leaf field, one can use
  /// manifest().schema_fields to get the top level fields, and then walk the
  /// tree to identify the relevant leaf fields and access its column_index.
  /// To get the total number of leaf fields, use FileMetadata.num_columns().
```

wjones127 (commented):
Related: I think this doc comment is misleading as well:

https://github.com/apache/arrow/blob/2410d36773d04ebe1b13a54748a947eed6fb304e/cpp/src/parquet/metadata.h#L285

It returns the number of leaf fields in the schema, not the number of top-level ones (at least as most people would understand). Would you mind updating that as well?

cpp/src/parquet/arrow/reader.h
LouisClt: Yes, your version seems to be more readable than mine.
I also updated the comments for the "num_columns()" method.

cpp/src/parquet/metadata.h
```diff
@@ -282,11 +282,20 @@ class PARQUET_EXPORT FileMetaData {
 
   bool Equals(const FileMetaData& other) const;
 
-  /// \brief The number of top-level columns in the schema.
+  /// \brief The number of parquet "leaf" fields.
```
wgtmac, line 285: In the parquet specs, the official wording is `column`. 

I'd suggest follow the same syntax.

https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L680

LouisClt: Yes, it can't hurt to be coherent. I made the change.

cpp/src/parquet/metadata.h
```diff
@@ -282,11 +282,20 @@ class PARQUET_EXPORT FileMetaData {
...
+  /// flattened. This method returns the number of columns in the flattened
   /// version.
+  /// For instance, if the schema looks like this :
+  /// 0 foo.bar
+  ///       foo.bar.baz           0
+  ///       foo.bar.baz2          1
+  ///   foo.qux                   2
+  /// 1 foo2                      3
+  /// 2 foo3                      4
+  /// This method will return 5, because there is 5 "leaf" fields (so 5
```
wjones127, line 297:
```suggestion
  /// This method will return 5, because there are 5 "leaf" fields (so 5
```

wjones127 (approved):
This looks good. Thanks @LouisClt!

PR #15194, opened 2023-01-04
GH-15102: [C++] Could not decompress arrow stream sent from Java arrow SDK

### LINKED ISSUE #15102: [C++] Could not decompress arrow stream sent from Java arrow SDK

Hi guys,
I am using arrow java sdk to send  lz4/zstd stream to a server which is running c++ arrow sdk, but the c++ arrow report a failure: "Negative buffer resize: -1"

After some investigation, I found that Java compress process is not compatible with C++.
In Java, the class AbstractCompressionCodec has one optimization, if the buffer size after compression is larger than the uncompressed buffer, AbstractCompressionCodec::compress() will attach the uncompressed buffer directly and mark the buffer size with "-1".
But in C++, reader.cc::DecompressBuffer() does not check buffer size with "-1". It allocates the buffer directly and so the error "Negative buffer resize: -1" is thrown.

### Component(s)

C++, Java

benibus:
* Closes: #15102

### DIFF
--- a/cpp/src/arrow/ipc/options.h
+++ b/cpp/src/arrow/ipc/options.h
@@ -73,8 +73,8 @@ struct ARROW_EXPORT IpcWriteOptions {
   ///
   /// For example, if min_space_savings = 0.1, a 100-byte body buffer won't undergo
   /// compression if its expected compressed size exceeds 90 bytes. If this option is
-  /// set to 0.0, compression will be used indiscriminately. If no codec was supplied, this
-  /// option is ignored.
+  /// set to 0.0, compression will be used indiscriminately. If no codec was supplied,
+  /// this option is ignored.
   double min_space_savings = 0.0;
 
   /// \brief Use global CPU thread pool to parallelize any computational tasks
--- a/cpp/src/arrow/ipc/options.h
+++ b/cpp/src/arrow/ipc/options.h
@@ -18,6 +18,7 @@
 #pragma once
 
 #include <cstdint>
+#include <optional>
 #include <vector>
 
 #include "arrow/io/caching.h"
@@ -73,9 +74,14 @@ struct ARROW_EXPORT IpcWriteOptions {
   ///
   /// For example, if min_space_savings = 0.1, a 100-byte body buffer won't undergo
   /// compression if its expected compressed size exceeds 90 bytes. If this option is
-  /// set to 0.0, compression will be used indiscriminately. If no codec was supplied,
-  /// this option is ignored.
-  double min_space_savings = 0.0;
+  /// unset, compression will be used indiscriminately. If no codec was supplied, this
+  /// option is ignored.
+  ///
+  /// Values outside of the range [0,1] are handled as errors.
+  ///
+  /// Note that enabling this option may result in unreadable data for Arrow C++ versions
+  /// prior to 12.0.0.
+  std::optional<double> min_space_savings;
 
   /// \brief Use global CPU thread pool to parallelize any computational tasks
   /// like compression
--- a/cpp/src/arrow/ipc/options.h
+++ b/cpp/src/arrow/ipc/options.h
@@ -72,8 +72,8 @@ struct ARROW_EXPORT IpcWriteOptions {
   /// Space savings is calculated as (1.0 - compressed_size / uncompressed_size).
   ///
   /// For example, if min_space_savings = 0.1, a 100-byte body buffer won't undergo
-  /// compression if its expected compressed size exceeds 90 bytes. If this option isn't
-  /// set, compression will be used indiscriminately. If no codec was supplied, this
+  /// compression if its expected compressed size exceeds 90 bytes. If this option is
+  /// set to 0.0, compression will be used indiscriminately. If no codec was supplied, this
   /// option is ignored.
   double min_space_savings = 0.0;
 
--- a/cpp/src/arrow/ipc/options.h
+++ b/cpp/src/arrow/ipc/options.h
@@ -18,6 +18,7 @@
 #pragma once
 
 #include <cstdint>
+#include <optional>
 #include <vector>
 
 #include "arrow/io/caching.h"
@@ -67,14 +68,17 @@ struct ARROW_EXPORT IpcWriteOptions {
   /// May only be UNCOMPRESSED, LZ4_FRAME and ZSTD.
   std::shared_ptr<util::Codec> codec;
 
-  /// @brief Always use compression, regardless of the resulting output size
+  /// @brief Minimum space savings percentage required for compression to be applied
   ///
-  /// If false, compression will be skipped for body buffers where the expected compressed
-  /// size matches/exceeds the uncompressed size. Otherwise, compression will be applied
-  /// indiscriminately. This option is ignored if no codec was supplied.
+  /// Space savings is calculated as (1.0 - compressed_size / uncompressed_size).
   ///
-  /// This is true by default due to backwards-compatibility concerns.
-  bool compress_always = true;
+  /// For example, if min_space_savings = 0.1, a 100-byte body buffer won't undergo
+  /// compression if its expected compressed size exceeds 90 bytes. If this option isn't
+  /// set, compression will be used indiscriminately. If no codec was supplied, this
+  /// option is ignored.
+  ///
+  /// Unset by default due to backwards-compatibility concerns
+  std::optional<double> min_space_savings;
 
   /// \brief Use global CPU thread pool to parallelize any computational tasks
   /// like compression
--- a/cpp/src/arrow/ipc/options.h
+++ b/cpp/src/arrow/ipc/options.h
@@ -18,7 +18,6 @@
 #pragma once
 
 #include <cstdint>
-#include <optional>
 #include <vector>
 
 #include "arrow/io/caching.h"
@@ -76,9 +75,7 @@ struct ARROW_EXPORT IpcWriteOptions {
   /// compression if its expected compressed size exceeds 90 bytes. If this option isn't
   /// set, compression will be used indiscriminately. If no codec was supplied, this
   /// option is ignored.
-  ///
-  /// Unset by default due to backwards-compatibility concerns
-  std::optional<double> min_space_savings;
+  double min_space_savings = 0.0;
 
   /// \brief Use global CPU thread pool to parallelize any computational tasks
   /// like compression
--- a/cpp/src/arrow/ipc/options.h
+++ b/cpp/src/arrow/ipc/options.h
@@ -68,7 +68,7 @@ struct ARROW_EXPORT IpcWriteOptions {
   /// May only be UNCOMPRESSED, LZ4_FRAME and ZSTD.
   std::shared_ptr<util::Codec> codec;
 
-  /// @brief Minimum space savings percentage required for compression to be applied
+  /// \brief Minimum space savings percentage required for compression to be applied
   ///
   /// Space savings is calculated as (1.0 - compressed_size / uncompressed_size).
   ///
--- a/cpp/src/arrow/ipc/options.h
+++ b/cpp/src/arrow/ipc/options.h
@@ -67,6 +67,15 @@ struct ARROW_EXPORT IpcWriteOptions {
   /// May only be UNCOMPRESSED, LZ4_FRAME and ZSTD.
   std::shared_ptr<util::Codec> codec;
 
+  /// @brief Always use compression, regardless of the resulting output size
+  ///
+  /// If false, compression will be skipped for body buffers where the expected compressed
+  /// size matches/exceeds the uncompressed size. Otherwise, compression will be applied
+  /// indiscriminately. This option is ignored if no codec was supplied.
+  ///
+  /// This is true by default due to backwards-compatibility concerns.
+  bool compress_always = true;
+
   /// \brief Use global CPU thread pool to parallelize any computational tasks
   /// like compression
   bool use_threads = true;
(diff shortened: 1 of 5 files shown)


lidavidm:
We should probably make sure this is covered in a unit test, and check in a file to the test data repo that covers this, too, so that any language using the integration testing infrastructure is also tested. (Also, we _may_ want to implement this optimization in C++, though I fear that has a high likelihood of generating backwards-incompatible files...)

pitrou:
> Also, we _may_ want to implement this optimization in C++, though I fear that has a high likelihood of generating backwards-incompatible files...

Agreed that it unfortunately wouldn't be very friendly at this point. We can revisit in two or three years probably...

lidavidm:
I'll do that. Oddly enough, I don't see any practical way in the Java implementation to actually write a compressed file or stream, without subclassing and poking the writer implementation...so I'll be making some other changes.

lidavidm:
https://github.com/apache/arrow-testing/pull/85

benibus:
So, I went out on a limb and implemented the optimization on the C++ side. Given the backwards-compatibility concerns, I added `IpcWriteOptions::compress_always`, which enables it (but not by default). I don't know if this is an ideal solution in the long-term, but I could imagine it being useful regardless of any current file incompatibilities. That being said, I partially did it to make testing the reader easier,  since it wouldn't be straightforward otherwise. In any case, let me know what you think.

Also, the original changes in `reader.cc` weren't correct, so that's been fixed.

cpp/src/arrow/ipc/options.h
```diff
@@ -67,6 +67,15 @@ struct ARROW_EXPORT IpcWriteOptions {
...
   std::shared_ptr<util::Codec> codec;
 
+  /// @brief Always use compression, regardless of the resulting output size
+  ///
+  /// If false, compression will be skipped for body buffers where the expected compressed
+  /// size matches/exceeds the uncompressed size. Otherwise, compression will be applied
+  /// indiscriminately. This option is ignored if no codec was supplied.
+  ///
+  /// This is true by default due to backwards-compatibility concerns.
+  bool compress_always = true;
```
pitrou, line 77: We could perhaps get smart and make this an optional floating-point threshold. For example only compress if it would save at least 10% of the buffer size.

On the flip side, compression is typically much slower than decompression, and compression is attempted anyway, so the end-to-end cost might not vary that much. @lidavidm @westonpace Thoughts?

pitrou: A more advanced strategy would be to compress only parts of the buffer and compute a compression ratio on those sampled parts; that's what Dask [does](https://github.com/dask/distributed/blob/0063de53fed5e4e2e409940213c6265867e6635d/distributed/protocol/compression.py#L153).

pitrou: Of course, we don't have to implement an advanced strategy here. However, we must make an API choice and I think it's better to be forward-looking (and implement the heuristic in the most simplistic way, for now).

lidavidm: I think the threshold sounds good.

I suppose it may not save encoder time, but it could save some decoder time, right? Else I suppose I don't see the point of the optimization at all.

pitrou: It would save decoder time. But also, if using a sophisticated encoding strategy as mentioned above, it would also save encoder time.

benibus:
This is ready for a second look.

I switched over to using a `min_space_savings` percentage instead of just a boolean (the name is a bit weird, but it [supposedly has some precedence](https://en.wikipedia.org/wiki/Data_compression_ratio) so that's what I went with - no strong opinions though).

I'm also not quite sure whether the parameter should be a literal `std::optional` or just `double = 0.0` by default. The effect should be the same either way. Some kind of range-validation is probably in order though, at least in debug mode.

pitrou:
I think `double = 0.0` is enough.

lidavidm:
Do we want to incorporate https://github.com/apache/arrow-testing/pull/85 ?

benibus:
I think it's a good idea. The unit test coverage here is fairly superficial (and relies on the reader's implementation details).

If we ever do a Dask-like sampling optimization then I suspect we'd need to reassess the generated files, but any uncompressible input should be sufficient for now.

lidavidm:
We can always add more generated files when needed.

If you can confirm the generated files fail without the patch/pass with the patch, then I can merge the new files and then you can bump the submodule commit as part of this PR.

cpp/src/arrow/ipc/options.h
```diff
@@ -67,6 +67,16 @@ struct ARROW_EXPORT IpcWriteOptions {
   /// May only be UNCOMPRESSED, LZ4_FRAME and ZSTD.
   std::shared_ptr<util::Codec> codec;
 
+  /// @brief Minimum space savings percentage required for compression to be applied
+  ///
+  /// Space savings is calculated as (1.0 - compressed_size / uncompressed_size).
+  ///
+  /// For example, if min_space_savings = 0.1, a 100-byte body buffer won't undergo
+  /// compression if its expected compressed size exceeds 90 bytes. If this option isn't
+  /// set, compression will be used indiscriminately. If no codec was supplied, this
```
pitrou, line 76:
```suggestion
  /// compression if its expected compressed size exceeds 90 bytes. If this option is
  /// set to 0.0, compression will be used indiscriminately. If no codec was supplied, this
```

cpp/src/arrow/ipc/writer.cc
```diff
@@ -176,19 +176,43 @@ class RecordBatchSerializer {
...
+    ARROW_ASSIGN_OR_RAISE(auto result,
+                          AllocateResizableBuffer(maximum_length + sizeof(int64_t)));
+    ARROW_ASSIGN_OR_RAISE(auto actual_length,
                           codec->Compress(buffer.size(), buffer.data(), maximum_length,
                                           result->mutable_data() + sizeof(int64_t)));
+    // FIXME: Not the most sophisticated way to handle this. Ideally, you'd want to avoid
+    // pre-compressing the entire buffer via some kind of sampling method. As the feature
+    // gains adoption, this may become a worthwhile optimization.
+    if (!ShouldCompress(buffer.size(), actual_length)) {
+      if (buffer.size() < actual_length || buffer.size() > maximum_length) {
```
pitrou, line 201: I'm not sure why `buffer.size() < actual_length` if you're passing `/*shrink_to_fit=*/false` below?

cpp/src/arrow/ipc/writer.cc
```diff
@@ -176,19 +176,43 @@ class RecordBatchSerializer {
...
-    int64_t actual_length;
-    ARROW_ASSIGN_OR_RAISE(actual_length,
+    ARROW_ASSIGN_OR_RAISE(auto result,
+                          AllocateResizableBuffer(maximum_length + sizeof(int64_t)));
+    ARROW_ASSIGN_OR_RAISE(auto actual_length,
                           codec->Compress(buffer.size(), buffer.data(), maximum_length,
                                           result->mutable_data() + sizeof(int64_t)));
+    // FIXME: Not the most sophisticated way to handle this. Ideally, you'd want to avoid
+    // pre-compressing the entire buffer via some kind of sampling method. As the feature
+    // gains adoption, this may become a worthwhile optimization.
```
pitrou, line 199: I've opened GH-33885 for this, which you can reference here.

cpp/src/arrow/ipc/writer.cc
pitrou: (though it's probably harmless)

cpp/src/arrow/ipc/options.h
```diff
@@ -67,6 +67,16 @@ struct ARROW_EXPORT IpcWriteOptions {
...
+  /// option is ignored.
```
pitrou, line 77: Also, can you add that enabling non-zero savings can make the data unreadable for pre-12.0.0 Arrow C++ versions?

cpp/src/arrow/ipc/read_write_test.cc
```diff
@@ -720,6 +771,11 @@ TEST_F(TestWriteRecordBatch, WriteWithCompression) {
     write_options.use_threads = false;
     read_options.use_threads = false;
     CheckRoundtrip(*batch, write_options, read_options);
+
+    ASSERT_OK_AND_ASSIGN(write_options.codec, MockCodec::Create(codec));
+    // 200% savings is impossible, so compression/decompression should be skipped
+    write_options.min_space_savings = 2.0;
```
pitrou, line 777: A more interesting check would be a `min_space_savings` between 0 and 1, but without random (uncompressible) data, to check that compression isn't enabled.

pitrou: (something like `min_space_savings = 2.0` should perhaps raise an error?)

pitrou (commented):
Some comments, but I also agree with @lidavidm that more testing would be worthwhile.

cpp/src/arrow/ipc/writer.cc
benibus: It was strictly to zero the excess padding without manually using `memset` in a separate path. TBH, I'm not _entirely_ sure if zeroing is necessary in this context, but I erred on the side of caution since the original allocation would've been pre-initialized.

cpp/src/arrow/ipc/writer.cc
```diff
@@ -176,19 +176,43 @@ class RecordBatchSerializer {
                                    field_nodes_, buffer_meta_, options_, &out_->metadata);
   }
 
+  bool ShouldCompress(int64_t uncompressed_size, int64_t compressed_size) const {
+    auto max_compressed_size = static_cast<int64_t>(
+        std::floor((1.0 - options_.min_space_savings) * uncompressed_size));
+    return compressed_size <= max_compressed_size;
```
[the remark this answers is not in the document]
benibus: You're right, that isn't taken into account here.

My only issue is that you'd lose the ability to easily say "only compress if doing so wouldn't be strictly counterproductive", which seems like a fairly common/sensible constraint. Instead, one would need to specify an arbitrary non-zero value to get that behavior.

I suspect that using `std::optional` was the better choice after all. I hadn't fully consider this edge-case before suggesting its removal.

benibus:
I'll try to complete this in the next few days - didn't mean to let it sit.

benibus:
Update: Integration files still pass.

cpp/src/arrow/ipc/options.h
```diff
@@ -67,6 +68,21 @@ struct ARROW_EXPORT IpcWriteOptions {
   /// May only be UNCOMPRESSED, LZ4_FRAME and ZSTD.
   std::shared_ptr<util::Codec> codec;
 
+  /// @brief Minimum space savings percentage required for compression to be applied
```
lidavidm, line 71: nit: Arrow mostly uses `\brief` not Javadoc-style `@brief`

```suggestion
  /// \brief Minimum space savings percentage required for compression to be applied
```

cpp/src/arrow/ipc/read_write_test.cc
```diff
@@ -735,6 +737,69 @@ TEST_F(TestWriteRecordBatch, WriteWithCompression) {
...
+  auto batch = RecordBatchFromJSON(schema({field("n", int64())}), R"([
+    {"n":0},{"n":1},{"n":2},{"n":3},{"n":4},
+    {"n":5},{"n":6},{"n":7},{"n":8},{"n":9}])");
+
+  auto prefixed_size = [](const Buffer& buffer) -> int64_t {
+    if (buffer.size() < int64_t(sizeof(int64_t))) return 0;
+    return bit_util::FromLittleEndian(util::SafeLoadAs<int64_t>(buffer.data()));
+  };
+  auto content_size = [](const Buffer& buffer) -> int64_t {
+    return buffer.size() - sizeof(int64_t);
```
lidavidm, line 751: nit: should this have a check like prefixed_size?

lidavidm (approved):
LGTM, thank you.

I'll merge the testing PR. Then the submodule commit can be bumped here.

cpp/src/arrow/ipc/read_write_test.cc
benibus: It should be validated by the check in the loop (`ASSERT_GT(compressed_size, 0)`).

Main reason for the check in prefixed_size was because the memory-read could theoretically yield a positive value even if `buffer.size() < 8` (due to padding).

lidavidm:
Ok, you should be able to bump to https://github.com/apache/arrow-testing/pull/85

benibus:
Alright, should be good to go.

lidavidm:
@zeroshade it looks like Go has the same bug as C++ here

@benibus we should have Go skip the new files for now

benibus:
@lidavidm Ok, they should be skipped now (I assume archery was the right place to do this?)

lidavidm:
Yup, that looks good. Thanks.

lidavidm:
Filed #34385 for Go

benibus:
@zeroshade Thanks! Should be all set now

lidavidm:
Seems Go still panics in CI

zeroshade:
Weird, I'll investigate and get back to you @benibus, in theory it should be working properly unless i mucked something up in my unit tests....

zeroshade (approved):

zeroshade:
Fixed the Go issue, I made a silly mistake. Integration tests are all green now! :smile:

wgtmac:
Does the current PR solve this issue: https://github.com/apache/arrow/issues/34432 ? @benibus @lidavidm

lidavidm:
I don't believe so.

lidavidm (approved):

PR #15201, opened 2023-01-05
GH-15200: [C++] Created benchmarks for round kernels.

### LINKED ISSUE #15200: [C++][Arrow] Add benchmarks for the round functions.

### Component(s)

C++

EpsilonPrime:
The four existing kernel functions Ceil, Floor, Round, and Trunc gain benchmarks with this change.
* Closes: #15200


EpsilonPrime:
One thought about the number of benchmarks -- there is a benchmark for every type and round mode (10 each).  Some alternatives could be to try to fold the round mode tests into the type (for less visibility) or come up with a "less frequently used" benchmark concept if having less benchmarks is ideal.

cyb70289:
> One thought about the number of benchmarks -- there is a benchmark for every type and round mode (10 each). Some alternatives could be to try to fold the round mode tests into the type (for less visibility) or come up with a "less frequently used" benchmark concept if having less benchmarks is ideal.

It does looks too many benchmarks are generated. And I guess most of them are similar.
How long will all these cases run? If the round modes don't impact performance, I agree that we can only test one of them. And probably few data types, e.g., int64/uint64/double.

cyb70289:
Do you plan to add benchmark for decimal types? It can be in a followup PR.

cpp/src/arrow/compute/kernels/scalar_round_benchmark.cc
```diff
@@ -0,0 +1,111 @@
...
+  options.round_mode = static_cast<RoundMode>(Mode);
+
+  for (auto _ : state) {
+    ABORT_NOT_OK(CallFunction(func_name, {val}, &options));
+  }
+  state.SetItemsProcessed(state.iterations() * array_size);
+}
+
+void SetRoundArgs(benchmark::internal::Benchmark* bench) {
+  for (const auto inverse_null_proportion : std::vector<ArgsType>({100, 0})) {
```
cyb70289, line 56: Add more null proportions like other benchmarks?
Of course, we should first try to reduce the total test cases here.

EpsilonPrime: I copied the null proportion behavior from the arithmetic benchmarks which appeared to be the closest to the round behavior.  Do you have a specific set of null proportions I should try?

EpsilonPrime:
> Do you plan to add benchmark for decimal types? It can be in a followup PR.

Done.

EpsilonPrime:
All of the tests together are about 3 minutes.  I've selected 3 of the round modes to run normally and disabled the rest using a command line flag.

cpp/src/arrow/compute/kernels/scalar_round_benchmark.cc
```diff
@@ -0,0 +1,123 @@
...
```
wjones127, line 56: Would be nice to label these args:

```suggestion
  bench->ArgNames({"size", "inverse_null_proportion"})
  
  for (const auto inverse_null_proportion : std::vector<ArgsType>({100, 0})) {
```

EpsilonPrime:
I've updated the arithmetic benchmarks to also have labels.  Thanks for the suggestion!

cpp/src/arrow/compute/kernels/scalar_round_benchmark.cc
cyb70289: Ah, that's okay.

cpp/src/arrow/compute/kernels/scalar_round_benchmark.cc
```diff
@@ -0,0 +1,125 @@
...
+}
+
+template <typename ArrowType, RoundMode Mode>
+static void Ceil(benchmark::State& state) {
+  RoundArrayBenchmark<ArrowType, Mode>(state, "ceil");
+}
+
+template <typename ArrowType, RoundMode Mode>
+static void Floor(benchmark::State& state) {
+  RoundArrayBenchmark<FloatType, Mode>(state, "floor");
```
cyb70289, line 70: `FloatType` looks wrong. Should it be `ArrowType`? Same for Round and Trunc.

cyb70289 (changes requested):

cpp/src/arrow/compute/kernels/scalar_round_benchmark.cc
EpsilonPrime: Good catch, fixed (although I had to remove the Decimal128 benchmarks for now).

cyb70289: We can leave decimals as followup task. Can you open an issue to track it? Thanks.

cyb70289:
Travis CI error is not related.

cyb70289 (approved):

cyb70289:
Not related to this PR, but the benchmark result is not obvious to me.
Why is `double` faster than `int64`? I suppose integer rounding does nothing.

```
Ceil<Int64Type, RoundMode::DOWN>/size:1048576/inverse_null_proportion:0                      323253 ns       323251 ns        
2161 bytes_per_second=3.02107G/s items_per_second=405.481M/s null_percent=0 size=1048.58k

Ceil<DoubleType, RoundMode::DOWN>/size:1048576/inverse_null_proportion:0                      89834 ns        89832 ns         
7795 bytes_per_second=10.871G/s items_per_second=1.45908G/s null_percent=0 size=1048.58k
```

PR #15210, opened 2023-01-05
GH-20512: [Python] Numpy conversion doesn't account for ListArray offset

### LINKED ISSUE #20512: [Python] Quadratic memory usage of Table.to_pandas with nested data
Reading nested Parquet data and then converting it to a Pandas DataFrame shows quadratic memory usage and will eventually run out of memory for reasonably small files. I had initially thought this was a regression since 7.0.0, but it looks like 7.0.0 has similar quadratic memory usage that kicks in at higher row counts.

Example code to generate nested Parquet data:
```python

import numpy as np
import random
import string
import pandas as pd

_characters = string.ascii_uppercase + string.digits + string.punctuation

def make_random_string(N=10):
    return ''.join(random.choice(_characters) for _ in range(N))

nrows = 1_024_000
filename = 'nested.parquet'

arr_len = 10
nested_col = []
for i in range(nrows):
    nested_col.append(np.array(
            [{
                'a': None if i % 1000 == 0 else np.random.choice(10000, size=3).astype(np.int64),
                'b': None if i % 100 == 0 else random.choice(range(100)),
                'c': None if i % 10 == 0 else make_random_string(5)
            } for i in range(arr_len)]
        ))
df = pd.DataFrame({'c1': nested_col})
df.to_parquet(filename)
```
And then read into a DataFrame with:
```python

import pyarrow.parquet as pq
table = pq.read_table(filename)
df = table.to_pandas()
```
Only reading to an Arrow table isn't a problem, it's the to_pandas method that exhibits the large memory usage. I haven't tested generating nested Arrow data in memory without writing Parquet from Pandas but I assume the problem probably isn't Parquet specific.

Memory usage I see when reading different sized files on a machine with 64 GB RAM:

|Num rows|Memory used with 10.0.1 (MB)|Memory used with 7.0.0 (MB)|
|
|-|-|-|-|
|32,000|362|361|
|
|64,000|531|531|
|
|128,000|1,152|1,101|
|
|256,000|2,888|1,402|
|
|512,000|10,301|3,508|
|
|1,024,000|38,697|5,313|
|
|2,048,000|OOM|20,061|
|
|4,096,000| |OOM|
<br>
<br>With Arrow 10.0.1, memory usage approximately quadruples when row count doubles above 256k rows. With Arrow 7.0.0 memory usage is more linear but then quadruples from 1024k to 2048k rows.
<br>
<br>PyArrow 8.0.0 shows similar memory usage to 10.0.1 so it looks like something changed between 7.0.0 and 8.0.0.|

**Environment**: Python 3.10.8 on Fedora Linux 36. AMD Ryzen 9 5900 X with 64 GB RAM
**Reporter**: [Adam Reeve](https://issues.apache.org/jira/browse/ARROW-18400) / @adamreeve
**Assignee**: [Will Jones](https://issues.apache.org/jira/browse/ARROW-18400) / @wjones127
#### Original Issue Attachments:
- [test_memory.py](https://issues.apache.org/jira/secure/attachment/13054045/test_memory.py)
#### PRs and other links:
- [GitHub Pull Request #15210](https://github.com/apache/arrow/pull/15210)

<sub>**Note**: *This issue was originally created as [ARROW-18400](https://issues.apache.org/jira/browse/ARROW-18400). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

wjones127:
* Closes: #20512

### DIFF
--- a/cpp/src/arrow/array/array_list_test.cc
+++ b/cpp/src/arrow/array/array_list_test.cc
@@ -509,18 +509,6 @@ class TestListArray : public ::testing::Test {
     ASSERT_RAISES(Invalid, ValidateOffsets(2, {0, 7, 4}, values));
   }
 
-  void TestSliced() {
-    auto arr = ArrayFromJSON(list(int16()), "[[1, 2], [3, 4, 5], [6], [7, 8]]");
-
-    auto arr_sliced = arr->Slice(0, 2);
-    auto expected_sliced = ArrayFromJSON(list(int16()), "[[1, 2], [3, 4, 5]]");
-    AssertArraysEqual(*expected_sliced, *arr_sliced);
-
-    auto values = checked_cast<ListArray*>(arr_sliced.get())->values();
-    auto expected_values = ArrayFromJSON(int16(), "[1, 2, 3, 4, 5]");
-    AssertArraysEqual(*expected_values, *values);
-  }
-
   void TestCornerCases() {
     // ARROW-7985
     ASSERT_OK(builder_->AppendNull());
@@ -612,8 +600,6 @@ TYPED_TEST(TestListArray, TestFlattenNonEmptyBackingNulls) {
 }
 
 TYPED_TEST(TestListArray, ValidateOffsets) { this->TestValidateOffsets(); }
-
-TYPED_TEST(TestListArray, TestSliced) { this->TestSliced(); }
 
 TYPED_TEST(TestListArray, CornerCases) { this->TestCornerCases(); }
 
--- a/cpp/src/arrow/array/array_list_test.cc
+++ b/cpp/src/arrow/array/array_list_test.cc
@@ -509,6 +509,18 @@ class TestListArray : public ::testing::Test {
     ASSERT_RAISES(Invalid, ValidateOffsets(2, {0, 7, 4}, values));
   }
 
+  void TestSliced() {
+    auto arr = ArrayFromJSON(list(int16()), "[[1, 2], [3, 4, 5], [6], [7, 8]]");
+
+    auto arr_sliced = arr->Slice(0, 2);
+    auto expected_sliced = ArrayFromJSON(list(int16()), "[[1, 2], [3, 4, 5]]");
+    AssertArraysEqual(*expected_sliced, *arr_sliced);
+
+    auto values = checked_cast<ListArray*>(arr_sliced.get())->values();
+    auto expected_values = ArrayFromJSON(int16(), "[1, 2, 3, 4, 5]");
+    AssertArraysEqual(*expected_values, *values);
+  }
+
   void TestCornerCases() {
     // ARROW-7985
     ASSERT_OK(builder_->AppendNull());
@@ -600,6 +612,8 @@ TYPED_TEST(TestListArray, TestFlattenNonEmptyBackingNulls) {
 }
 
 TYPED_TEST(TestListArray, ValidateOffsets) { this->TestValidateOffsets(); }
+
+TYPED_TEST(TestListArray, TestSliced) { this->TestSliced(); }
 
 TYPED_TEST(TestListArray, CornerCases) { this->TestCornerCases(); }
 
--- a/python/pyarrow/tests/test_pandas.py
+++ b/python/pyarrow/tests/test_pandas.py
@@ -4522,18 +4522,18 @@ def test_list_no_duplicate_base():
 
     np_arr = chunked_arr.to_numpy()
 
-    expected = np.array([np.array(part) for part in [[1, 2], [3, 4, 5], [6]]])
+    expected = np.array([[1, 2], [3, 4, 5], [6]], dtype="object")
     for left, right in zip(np_arr, expected):
         npt.assert_array_equal(left, right)
 
-    expected_base = np.array([[1, 2, 3, 4, 5, 6]])
+    expected_base = np.array([[1, 2, 3, 4, 5, 6]], dtype="object")
     npt.assert_array_equal(np_arr[0].base, expected_base)
 
     np_arr_sliced = chunked_arr.slice(1, 3).to_numpy()
 
-    expected = np.array([[3, 4, 5], [6]])
+    expected = np.array([[3, 4, 5], [6]], dtype="object")
     for left, right in zip(np_arr_sliced, expected):
         npt.assert_array_equal(left, right)
 
-    expected_base = np.array([[3, 4, 5, 6]])
+    expected_base = np.array([[3, 4, 5, 6]], dtype="object")
     npt.assert_array_equal(np_arr_sliced[0].base, expected_base)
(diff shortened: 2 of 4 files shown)


cpp/src/arrow/array/array_list_test.cc
```diff
@@ -509,6 +509,18 @@ class TestListArray : public ::testing::Test {
...
+  void TestSliced() {
+    auto arr = ArrayFromJSON(list(int16()), "[[1, 2], [3, 4, 5], [6], [7, 8]]");
+
+    auto arr_sliced = arr->Slice(0, 2);
+    auto expected_sliced = ArrayFromJSON(list(int16()), "[[1, 2], [3, 4, 5]]");
+    AssertArraysEqual(*expected_sliced, *arr_sliced);
+
+    auto values = checked_cast<ListArray*>(arr_sliced.get())->values();
+    auto expected_values = ArrayFromJSON(int16(), "[1, 2, 3, 4, 5]");
+    AssertArraysEqual(*expected_values, *values);
```
wjones127, line 521: Should we make this expected behavior or not?

jorisvandenbossche: I just commented on the JIRA as well, but so in the current design, we shouldn't change that. It is `Flatten()` that already implements this, and the `values` being unsliced ensures that they still match the `offsets` of a sliced ListArray.

It is certainly confusing, though. I wonder if we should give it a more scary name like "raw_values"

wjones127:
I re-ran the original reproduction and it seems memory usage is no longer quadratic:

| Num rows | Memory usage (10.0.1) | Memory usage (after) |
|     ---: |       --:             | ---: |
| 256k | 2,153,767,662 | 1,102,736,461 |
| 512k | 8,496,047,798 | 2,185,596,364 |


<details>
<summary>Code for test</summary>

Write test file:
```python
import numpy as np
import random
import string
import tracemalloc
import pyarrow as pa
import pyarrow.parquet as pq

_characters = string.ascii_uppercase + string.digits + string.punctuation

def make_random_string(N=10):
    return ''.join(random.choice(_characters) for _ in range(N))

nrows = 256_000
filename = 'nested_pandas.parquet'

arr_len = 10
nested_col = []
for i in range(nrows):
    nested_col.append(np.array(
            [{
                'a': None if i % 1000 == 0 else np.random.choice(10000, size=3).astype(np.int64),
                'b': None if i % 100 == 0 else random.choice(range(100)),
                'c': None if i % 10 == 0 else make_random_string(5)
            } for i in range(arr_len)]
        ))

table = pa.table({'c1': nested_col})

# table = pa.table({
#     'c1': pa.array([list(range(random.randint(1, 20))) for _ in range(nrows)])
# })

# Writing to .parquet and loading it into arrow again
pq.write_table(table, filename)
```

Then measure:
```python
import tracemalloc
import pyarrow.parquet as pq

filename = '/Users/willjones/Documents/arrows/arrow/python/nested_pandas.parquet'
tracemalloc.start()
table_from_parquet = pq.read_table(filename)

out = table_from_parquet.to_pandas()

print(tracemalloc.get_traced_memory())
```
</details>

cpp/src/arrow/array/array_list_test.cc
jorisvandenbossche: > The problem with Flatten though is that it removes the null values from the values array, but in this case we want them.

Because we use the offsets to slice into the flat values, and those offsets take into account potential values behind a null?

This corner case is covered by existing tests?

python/pyarrow/tests/test_pandas.py
```diff
@@ -4513,3 +4513,27 @@ def test_does_not_mutate_timedelta_nested():
     df = table.to_pandas()
 
     assert df["timedelta_2"][0].to_pytimedelta() == timedelta_2[0]
+
+
+def test_list_no_duplicate_base():
```
jorisvandenbossche, line 4518: There is a `TestConvertListTypes` class that groups some list type related tests, maybe can move this somewhere there.

cpp/src/arrow/array/array_list_test.cc
wjones127: I should add a test.

And you might be right about flatten, it does include the nulls in the values (not the list). For some reason earlier I had gotten the impression it doesn't, while I was debugging.

```python
>>> import pyarrow as pa
>>> arr = pa.array([[1, 2], [3, 4, 5], [6, None], [7, 8]])
>>> arr.flatten()
<pyarrow.lib.Int64Array object at 0x129b2c820>
[
  1,
  2,
  3,
  4,
  5,
  6,
  null,
  7,
  8
]
```

python/pyarrow/tests/test_pandas.py
wjones127: I can move it there.

cpp/src/arrow/array/array_list_test.cc
jorisvandenbossche: Yes, but your example uses an actual null inside a list, not a "null list". For example, here the first null is not in the flattened output:

```
In [3]: arr = pa.array([[1, 2], None, [3, None]])

In [4]: arr.flatten()
Out[4]: 
<pyarrow.lib.Int64Array object at 0x7f8b1fbcaa40>
[
  1,
  2,
  3,
  null
]
```

And I suppose behind that null can be any data (the default when constructed above is that there is no data behind, so the offset doesn't increment for that list element):

```
In [5]: arr.offsets
Out[5]: 
<pyarrow.lib.Int32Array object at 0x7f8b4f0f5a80>
[
  0,
  2,
  2,
  4
]
```

It's a bit tricky to construct manually, but something like:

```
In [10]: arr = pa.ListArray.from_arrays(pa.array([0, 2, 4, 6]), pa.array([1, 2, 99, 99, 3, None]), mask=pa.array([False, True, False]))

In [11]: arr
Out[11]: 
<pyarrow.lib.ListArray object at 0x7f8b1fbcba60>
[
  [
    1,
    2
  ],
  null,
  [
    3,
    null
  ]
]

In [12]: arr.flatten()
Out[12]: 
<pyarrow.lib.Int64Array object at 0x7f8b4f065960>
[
  1,
  2,
  3,
  null
]

In [13]: arr.values
Out[13]: 
<pyarrow.lib.Int64Array object at 0x7f8b4f065780>
[
  1,
  2,
  99,
  99,
  3,
  null
]

In [14]: arr.offsets
Out[14]: 
<pyarrow.lib.Int32Array object at 0x7f8b1fbc9300>
[
  0,
  2,
  4,
  6
]
```

But, so I am not sure if for this case you actually need to flattened (with nulls removed) or not for this case of converting to numpy.

wjones127: > Because we use the offsets to slice into the flat values, and those offsets take into account potential values behind a null?

Oh now I remember, and I think I understand what you are saying here better now. For the fixed size lists, the values behind a null entry in the list are removed when we call `Flatten()`. When we go back to try to reconstruct the lists based on offsets, the offsets produced by `value_offset` are all invalid since they don't account for the values we dropped in `Flatten()`.

jorisvandenbossche:
Yes, agreed it would be nice to include this one in the release, given the quadratic memory issues. I did a review, and all looks good to me. @wjones127 I am just going to add one more test with the "hidden" null values (the reason we can't use Flatten), which I think is currently not covered by the tests (unless it was already covered by existing tests).

jorisvandenbossche (approved):

AlenkaF (approved):
+1, thank you for working on this @wjones127!

PR #15211, opened 2023-01-05
GH-27494: [R] Implement RPrimitiveConverter for Decimal type

### LINKED ISSUE #27494: [R] Implement RPrimitiveConverter for Decimal type
This succeeds:
```java

Array$create(1)$cast(decimal(10, 2))
```
but this fails:
```java

Array$create(1, type = decimal(10, 2))
```
with error:
```java

NotImplemented: Extend
```
because the `Extend` method of the `RPromitiveConverter` class for the Decimal type is not yet implemented.

The error is thrown here: <https://github.com/apache/arrow/blob/7184c3f46981dd52c3c521b2676796e82f17da77/r/src/r_to_arrow.cpp#L601>

**Reporter**: [Ian Cook](https://issues.apache.org/jira/browse/ARROW-11631) / @ianmcook
**Assignee**: [Dewey Dunnington](https://issues.apache.org/jira/browse/ARROW-11631) / @paleolimbot
#### Related issues:
- [[C++] Error when creating Table and using schema with decimal type](https://github.com/apache/arrow/issues/29501) (is duplicated by)
- [[R] Creating Decimal type using the example fails](https://github.com/apache/arrow/issues/30526) (relates to)
- [[R] Unit tests for decimals / ARROW-4565](https://github.com/apache/arrow/issues/15983) (relates to)
- [[R] Use Converter API to convert SEXP to Array/ChunkedArray](https://github.com/apache/arrow/issues/26534) (relates to)
#### PRs and other links:
- [GitHub Pull Request #15211](https://github.com/apache/arrow/pull/15211)

<sub>**Note**: *This issue was originally created as [ARROW-11631](https://issues.apache.org/jira/browse/ARROW-11631). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

paleolimbot:
Implements conversion from R to Decimal128/256 in such a way that it works with `Array$create()`.

Before this PR:

``` r
Array$create(1)$cast(decimal(10, 2))
#> Array
#> <decimal128(10, 2)>
#> [
#>   1.00
#> ]

Array$create(1, type = decimal128(10, 2))
#> Error in `value[[3L]]()`:
#> ! NotImplemented: Extend
#> ℹ You might want to try casting manually with `Array$create(...)$cast(...)`.

#> Backtrace:
#>     ▆
#>  1. └─Array$create(1, type = decimal128(10, 2))
#>  2.   └─base::tryCatch(...) at r/R/array.R:198:2
#>  3.     └─base (local) tryCatchList(expr, classes, parentenv, handlers)
#>  4.       └─base (local) tryCatchOne(expr, names, parentenv, handlers[[1L]])
#>  5.         └─value[[3L]](cond)
#>  6.           └─rlang::abort(...) at r/R/array.R:202:6
Array$create(1, type = decimal256(10, 2))
#> Error in `value[[3L]]()`:
#> ! NotImplemented: Extend
#> ℹ You might want to try casting manually with `Array$create(...)$cast(...)`.

#> Backtrace:
#>     ▆
#>  1. └─Array$create(1, type = decimal256(10, 2))
#>  2.   └─base::tryCatch(...) at r/R/array.R:198:2
#>  3.     └─base (local) tryCatchList(expr, classes, parentenv, handlers)
#>  4.       └─base (local) tryCatchOne(expr, names, parentenv, handlers[[1L]])
#>  5.         └─value[[3L]](cond)
#>  6.           └─rlang::abort(...) at r/R/array.R:202:6
```

<sup>Created on 2023-01-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

After this PR:

``` r
Array$create(1)$cast(decimal(10, 2))
#> Array
#> <decimal128(10, 2)>
#> [
#>   1.00
#> ]

Array$create(1, type = decimal128(10, 2))
#> Array
#> <decimal128(10, 2)>
#> [
#>   1.00
#> ]
Array$create(1, type = decimal256(10, 2))
#> Array
#> <decimal256(10, 2)>
#> [
#>   1.00
#> ]
```

<sup>Created on 2023-01-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

TODO: test!
* Closes: #27494

### DIFF
--- a/r/src/r_to_arrow.cpp
+++ b/r/src/r_to_arrow.cpp
@@ -727,9 +727,33 @@ class RPrimitiveConverter<T, enable_if_t<is_timestamp_type<T>::value>>
 template <typename T>
 class RPrimitiveConverter<T, enable_if_t<is_decimal_type<T>::value>>
     : public PrimitiveConverter<T, RConverter> {
+  using ValueType = typename arrow::TypeTraits<T>::BuilderType::ValueType;
+
  public:
   Status Extend(SEXP x, int64_t size, int64_t offset = 0) override {
-    return Status::NotImplemented("Extend");
+    RETURN_NOT_OK(this->Reserve(size - offset));
+    int32_t precision = this->primitive_type_->precision();
+    int32_t scale = this->primitive_type_->scale();
+
+    auto append_value = [this, precision, scale](double value) {
+      auto converted = ValueType::FromReal(value, precision, scale);
+      RETURN_NOT_OK(converted);
+      this->primitive_builder_->UnsafeAppend(converted.ValueUnsafe());
+      return Status::OK();
+    };
+
+    auto append_null = [this]() {
+      this->primitive_builder_->UnsafeAppendNull();
+      return Status::OK();
+    };
+
+    if (ALTREP(x)) {
+      return VisitVector(RVectorIterator_ALTREP<double>(x, offset), size, append_null,
+                         append_value);
+    } else {
+      return VisitVector(RVectorIterator<double>(x, offset), size, append_null,
+                         append_value);
+    }
   }
 };
 
--- a/r/src/r_to_arrow.cpp
+++ b/r/src/r_to_arrow.cpp
@@ -1309,7 +1309,7 @@ std::shared_ptr<arrow::ChunkedArray> vec_to_arrow_ChunkedArray(
   if (can_convert_native(x) && type->id() != Type::EXTENSION) {
     // short circuit if `x` is an altrep vector that shells a chunked Array
     auto maybe = altrep::vec_to_arrow_altrep_bypass(x);
-    if (maybe.get()) {
+    if (maybe.get() && maybe->type() == type) {
       return maybe;
     }
 
--- a/r/src/r_to_arrow.cpp
+++ b/r/src/r_to_arrow.cpp
@@ -727,7 +727,7 @@ class RPrimitiveConverter<T, enable_if_t<is_timestamp_type<T>::value>>
 template <typename T>
 class RPrimitiveConverter<T, enable_if_t<is_decimal_type<T>::value>>
     : public PrimitiveConverter<T, RConverter> {
-  using ValueType = typename arrow::TypeTraits<T>::BuilderType::ValueType;
+  using ValueType = typename arrow::TypeTraits<T>::CType;
 
  public:
   Status Extend(SEXP x, int64_t size, int64_t offset = 0) override {
--- a/r/src/r_to_arrow.cpp
+++ b/r/src/r_to_arrow.cpp
@@ -736,9 +736,8 @@ class RPrimitiveConverter<T, enable_if_t<is_decimal_type<T>::value>>
     int32_t scale = this->primitive_type_->scale();
 
     auto append_value = [this, precision, scale](double value) {
-      auto converted = ValueType::FromReal(value, precision, scale);
-      RETURN_NOT_OK(converted);
-      this->primitive_builder_->UnsafeAppend(converted.ValueUnsafe());
+      ARROW_ASSIGN_OR_RAISE(ValueType converted, ValueType::FromReal(value, precision, scale));
+      this->primitive_builder_->UnsafeAppend(converted);
       return Status::OK();
     };
 
--- a/r/src/r_to_arrow.cpp
+++ b/r/src/r_to_arrow.cpp
@@ -1309,7 +1309,7 @@ std::shared_ptr<arrow::ChunkedArray> vec_to_arrow_ChunkedArray(
   if (can_convert_native(x) && type->id() != Type::EXTENSION) {
     // short circuit if `x` is an altrep vector that shells a chunked Array
     auto maybe = altrep::vec_to_arrow_altrep_bypass(x);
-    if (maybe.get() && maybe->type() == type) {
+    if (maybe.get() && maybe->type()->Equals(type)) {
       return maybe;
     }
 
--- a/r/src/r_to_arrow.cpp
+++ b/r/src/r_to_arrow.cpp
@@ -736,7 +736,8 @@ class RPrimitiveConverter<T, enable_if_t<is_decimal_type<T>::value>>
     int32_t scale = this->primitive_type_->scale();
 
     auto append_value = [this, precision, scale](double value) {
-      ARROW_ASSIGN_OR_RAISE(ValueType converted, ValueType::FromReal(value, precision, scale));
+      ARROW_ASSIGN_OR_RAISE(ValueType converted,
+                            ValueType::FromReal(value, precision, scale));
       this->primitive_builder_->UnsafeAppend(converted);
       return Status::OK();
     };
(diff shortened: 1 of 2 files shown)


ianmcook:
Awesomesauce! Thanks @paleolimbot!

r/tests/testthat/test-Array.R
```diff
@@ -1321,3 +1321,19 @@ test_that("Array to C-interface", {
   delete_arrow_schema(schema_ptr)
   delete_arrow_array(array_ptr)
 })
+
+test_that("direct creation of Decimal Arrays (ARROW-11631)", {
+
+  decimal_array <- Array$create(1, type = decimal128(10, 2))
+  decimal_array2 <- Array$create(1, type = decimal256(10, 2))
```
paleolimbot, line 1328: Could the test Array include an `NA` and something whose precision might get truncated (i.e, maybe `Array$create(c(1, 1 / 3, NA), type = decimal128(10, 2))`)? (The `NA` is to get full test coverage since there's a branch for nulls in the C++; the truncation is to check that `type` gets passed through properly).

The purpose of `decimal_array2` isn't clear to me here...am I missing something?

r/tests/testthat/test-Array.R
```diff
@@ -1321,3 +1321,19 @@ test_that("Array to C-interface", {
...
+
+  expect_equal(
```
paleolimbot, line 1330: Because there's a branch for ALTREP, maybe `decimal_array2` could be `Array$create(1:10, type = decimal128(10, 2))`?

paleolimbot (commented):
Awesome!

Just a few comments to test all the branches of the C++. I'm going to try to solicit a review of the C++ too since this is my first time using the Converter API in any meaningful way.

r/tests/testthat/test-Array.R
paleolimbot: (In trying my own example here I'm realizing that I didn't consider integer arrays...hang tight and I'll add it to the C++!)

r/src/r_to_arrow.cpp
```diff
@@ -727,9 +727,33 @@ class RPrimitiveConverter<T, enable_if_t<is_timestamp_type<T>::value>>
...
   Status Extend(SEXP x, int64_t size, int64_t offset = 0) override {
-    return Status::NotImplemented("Extend");
+    RETURN_NOT_OK(this->Reserve(size - offset));
+    int32_t precision = this->primitive_type_->precision();
+    int32_t scale = this->primitive_type_->scale();
+
+    auto append_value = [this, precision, scale](double value) {
+      auto converted = ValueType::FromReal(value, precision, scale);
+      RETURN_NOT_OK(converted);
+      this->primitive_builder_->UnsafeAppend(converted.ValueUnsafe());
```
westonpace, line 741:
```suggestion
      ARROW_ASSIGN_OR_RAISE(ValueType converted, ValueType::FromReal(value, precision, scale));
      this->primitive_builder_->UnsafeAppend(converted);
```
You can use `ARROW_ASSIGN_OR_RAISE` here.  Also, I'm personally not a huge fan of auto but YMMV.

r/src/r_to_arrow.cpp
```diff
@@ -727,9 +727,33 @@ class RPrimitiveConverter<T, enable_if_t<is_timestamp_type<T>::value>>
 template <typename T>
 class RPrimitiveConverter<T, enable_if_t<is_decimal_type<T>::value>>
     : public PrimitiveConverter<T, RConverter> {
+  using ValueType = typename arrow::TypeTraits<T>::BuilderType::ValueType;
```
westonpace, line 730:
```suggestion
  using ValueType = typename arrow::TypeTraits<T>::CType;
```
I think this will work also and is a bit more direct.

r/tests/testthat/test-Array.R
westonpace: > The purpose of decimal_array2 isn't clear to me here...am I missing something?

Didn't you write this?  My interpretation is that `decimal_array2` is meant to verify the `decimal256` type works (in addition to the 128 bit variant).

> Could the test Array include an NA and something whose precision might get truncated (i.e, maybe Array$create(c(1, 1 / 3, NA), type = decimal128(10, 2)))?

I think truncation is less interesting to me than NA.  Testing truncation would just be verifying that `FromReal` works correctly and we can assume it does (or, if we have concerns, should test it elsewhere).  Testing `NA` will test that R's `NA` is getting properly recognized (e.g. adding coverage for `append_null`).

r/tests/testthat/test-Array.R
```diff
@@ -1321,3 +1321,19 @@ test_that("Array to C-interface", {
   delete_arrow_schema(schema_ptr)
   delete_arrow_array(array_ptr)
 })
+
+test_that("direct creation of Decimal Arrays (ARROW-11631)", {
```
westonpace, line 1325: I'm a little confused.  I expected to see an R array of doubles get converted to an Arrow array.  I didn't expect to see decimals get created directly.

westonpace (commented):
A few thoughts

r/tests/testthat/test-Array.R
paleolimbot: I'm not sure I know what the distinction is between those two things!

r/tests/testthat/test-Array.R
paleolimbot: I missed the 128/256 difference!

r/tests/testthat/test-Array.R
westonpace: Well, in python it would be something like:

```
# create an array of python decimals
x = [decimal.Decimal('12.34'), decimal.Decimal('23.45')]
# convert to an array of arrow decimals
arr = pa.array(x)
# arr
# <pyarrow.lib.Decimal128Array object at 0x7fd6a40dd120>
# [
#   12.34,
#   23.45
# ]
```

It looks like there are tests like this in `test-Array.R`:

```
test_that("Integer Array", {
  ints <- c(1:10, 1:10, 1:5)
  x <- expect_array_roundtrip(ints, int32())
})
```

Also in`test-Array.R` I see something like...

```
  expect_error(
    Array$create(as.double(1:10), type = decimal(4, 2)),
    "You might want to try casting manually"
  )
```

Shouldn't this work now?

paleolimbot: > Shouldn't this work now?

It does...that test fails and should be updated!

> `x = [decimal.Decimal('12.34'), decimal.Decimal('23.45')]`

Maybe the difference is that R doesn't have decimals (only reals)? Perhaps a less ambiguous title for the test would be "can convert R integer/double to decimal"?

westonpace: True, you can't do proper round trip if there is no native decimals.  I see now my confusion.  I read `Array$create(1, type = decimal128(10, 2))` too quickly and thought it was creating an Arrow scalar from a single R value.  I didn't realize it was treating `1` as an array of size 1.

r/tests/testthat/test-Array.R
thisisnic: Hmm, interesting...I now get this result:

```
> Array$create(c(1:10, NA))$cast(decimal128(10, 2))
Error: Invalid: Precision is not great enough for the result. It should be at least 12
/home/nic2/arrow/cpp/src/arrow/compute/exec.cc:828  kernel_->exec(kernel_ctx_, input, out)
/home/nic2/arrow/cpp/src/arrow/compute/exec.cc:796  ExecuteSingleSpan(input, &output)
/home/nic2/arrow/cpp/src/arrow/compute/function.cc:276  executor->Execute(input, &listener)
```

Looks like a bug given 10 significant digits is more than enough to represent the integers 1 through 10

thisisnic: Will update the tests to use 12 and open a ticket

thisisnic:
Looks like something is broken here:
```
> Array$create(c(1:10, NA), type = decimal128(12, 2))
Array
<decimal128(12, 2)>
[
  0.00,
  0.00,
  0.00,
  0.00,
  0.00,
  0.00,
  0.00,
  0.00,
  0.00,
  0.00,
  0.00
]
```

thisisnic:
> @thisisnic I think you were seeing that as a result of the force-push that overwrote the C++ I added to handle integers. I know you're off today so I merged both branches and added some tests to get this in before feature freeze slightly_smiling_face

Thanks for making that update, and sorry about the overwrite, brain fart!

PR #15213, opened 2023-01-05
GH-15212: [C++] fix sliced list array writing in ORC

### LINKED ISSUE #15212: ORC writer doesn't work on sliced list arrays

Follow up on #15210, where there was another misuse of `BaseListArray.values()`.

This can be reproduced with the unit test:

```cpp
TEST_F(TestORCWriterTrivialNoConversion, writeSlicedBatch) {
  std::shared_ptr<Table> table = GenerateRandomTable(table_schema, /*size=*/100, /*min_num_chunks=*/1, /*max_num_chunks*/1, /*probability=*/0);
  table = table->Slice(20, 60);
  
  AssertTableWriteReadEqual(table, table, kDefaultSmallMemStreamSize / 16);
}
```

### Component(s)

C++

wjones127:
Writing sliced list and map arrays will produce incorrect data. This happens because the arrays are recreated using `BaseListArray.values()`, which doesn't account for the offsets, and the offsets was not passed. To fix, I changed to pass along the offsets.
* Closes: #15212

### DIFF
--- a/cpp/src/arrow/adapters/orc/adapter_test.cc
+++ b/cpp/src/arrow/adapters/orc/adapter_test.cc
@@ -536,6 +536,15 @@ TEST_F(TestORCWriterTrivialNoConversion, writeFilledChunkAndSelectField) {
                             &selected_indices);
 }
 
+TEST_F(TestORCWriterTrivialNoConversion, writeSlicedBatch) {
+  std::shared_ptr<Table> table =
+      GenerateRandomTable(table_schema, /*size=*/100, /*min_num_chunks=*/1,
+                          /*max_num_chunks*/ 1, /*probability=*/0);
+  table = table->Slice(20, 60);
+
+  AssertTableWriteReadEqual(table, table, kDefaultSmallMemStreamSize / 16);
+}
+
 class TestORCWriterTrivialWithConversion : public ::testing::Test {
  public:
   TestORCWriterTrivialWithConversion() {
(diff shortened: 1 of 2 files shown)


cpp/src/arrow/adapters/orc/adapter_test.cc
```diff
@@ -536,6 +536,15 @@ TEST_F(TestORCWriterTrivialNoConversion, writeFilledChunkAndSelectField) {
                             &selected_indices);
 }
 
+TEST_F(TestORCWriterTrivialNoConversion, writeSlicedBatch) {
+  std::shared_ptr<Table> table =
+      GenerateRandomTable(table_schema, /*size=*/100, /*min_num_chunks=*/1,
```
wgtmac, line 541: Does it cover type like `list<list<X>>` to make sure that offset of the inner list is correct?

wjones127: The test doesn't cover that case. But it should be fine, since `NormalizeArray` is recursive. So  `NormalizeArray(list<list<X>>)` calls `NormalizeArray(list<X>)`, which calls `NormalizeArray(X)`. The offsets at each level should be handled.

pitrou:
Can we rephrase the PR title to describe what the PR does? "XXX doesn't work" is ambiguous (does it describe what the PR is supposed to fix?).

pitrou (approved):

PR #13983, opened 2022-08-27
ARROW-15006: [Python][CI][Doc] Enable numpydoc check PR03

amoeba:
Adds an additional numypdoc check to CI (PR03) and fixes all corresponding violations. 

This includes one minor breaking change to parameter order in `pyarrow.parquet.read_table` which makes order more uniform across `ParquetDataset.__new__`, `read_table`, and `read_pandas` which have related signatures.

Note this does not fully resolve [ARROW-15006](https://issues.apache.org/jira/browse/ARROW-15006).

PR #15214, opened 2023-01-06
ARROW-15006: [Python][Doc] Add five more numpydoc checks to CI

amoeba:
This adds the numpydoc checks GL10, PR04, PR05, RT03, and YD01 to CI and fixes the associated issues in docstrings.

These checks are:

- GL10: reST directives {directives} must be followed by two colons
- PR04: Parameter "{param_name}" has no type
- PR05: Parameter "{param_name}" type should not finish with "."
- RT03: Return value has no description
- YD01: No Yields section found

https://numpydoc.readthedocs.io/en/latest/validation.html

### DIFF
--- a/python/pyarrow/_dataset.pyx
+++ b/python/pyarrow/_dataset.pyx
@@ -2634,6 +2634,9 @@ cdef class Scanner(_Weakrefable):
             The number of batches to read ahead in a file. This might not work
             for all file formats. Increasing this number will increase
             RAM usage but could also improve IO utilization.
+        fragment_scan_options : FragmentScanOptions, default None
+            Options specific to a particular scan and fragment type, which
+            can change between different scans of the same dataset.
         use_threads : bool, default True
             If enabled, then maximum parallelism will be used determined by
             the number of available CPU cores.
@@ -2644,9 +2647,6 @@ cdef class Scanner(_Weakrefable):
         memory_pool : MemoryPool, default None
             For memory allocations, if required. If not specified, uses the
             default pool.
-        fragment_scan_options : FragmentScanOptions, default None
-            Options specific to a particular scan and fragment type, which
-            can change between different scans of the same dataset.
         use_threads : bool, default True
             If enabled, then maximum parallelism will be used determined by
             the number of available CPU cores.
--- a/python/pyarrow/_dataset.pyx
+++ b/python/pyarrow/_dataset.pyx
@@ -2513,16 +2513,6 @@ cdef class Scanner(_Weakrefable):
         ----------
         dataset : Dataset
             Dataset to scan.
-        use_threads : bool, default True
-            If enabled, then maximum parallelism will be used determined by
-            the number of available CPU cores.
-        use_async : bool, default True
-            This flag is deprecated and is being kept for this release for
-            backwards compatibility.  It will be removed in the next
-            release.
-        memory_pool : MemoryPool, default None
-            For memory allocations, if required. If not specified, uses the
-            default pool.
         columns : list of str, default None
             The columns to project. This can be a list of column names to
             include (order and duplicates will be preserved), or a dictionary
@@ -2611,16 +2601,6 @@ cdef class Scanner(_Weakrefable):
             fragment to scan.
         schema : Schema, optional
             The schema of the fragment.
-        use_threads : bool, default True
-            If enabled, then maximum parallelism will be used determined by
-            the number of available CPU cores.
-        use_async : bool, default True
-            This flag is deprecated and is being kept for this release for
-            backwards compatibility.  It will be removed in the next
-            release.
-        memory_pool : MemoryPool, default None
-            For memory allocations, if required. If not specified, uses the
-            default pool.
         columns : list of str, default None
             The columns to project. This can be a list of column names to
             include (order and duplicates will be preserved), or a dictionary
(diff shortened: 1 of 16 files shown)


amoeba:
Hey @jorisvandenbossche, since you reviewed [the last numpydoc PR](https://github.com/apache/arrow/pull/13983), would you want to take a look here? This adds five of the ten checks we discussed ([Jira](https://issues.apache.org/jira/browse/ARROW-15006), [Google Sheet](https://docs.google.com/spreadsheets/d/1NVzQe3MA4Mh-AqsWUmwIZu9r5gB-Lmugz7mge3stYvU)) adding and the diff is relatively straightforward compared to the last one.

python/pyarrow/_dataset.pyx
```diff
@@ -532,12 +532,12 @@ cdef class InMemoryDataset(Dataset):
 
     Parameters
     ----------
-    source : The data for this dataset.
+    source : The data for this dataset
```
wjones127, line 535: Why are the periods deleted? Other parameter descriptions, like the ones below, keep theirs.

wjones127: Actually this one makes sense; the description here is in the spot for the datatype, which isn't supposed to have a period. I think this comment needs to be fixed to have a type, then this description should go on the next line.

python/pyarrow/_dataset_parquet.pyx
```diff
@@ -67,11 +67,11 @@ cdef class ParquetFileFormat(FileFormat):
     Parameters
     ----------
     read_options : ParquetReadOptions
-        Read options for the file.
+        Read options for the file
```
wjones127, line 70: This one is puzzling though. Why remove this period?

python/pyarrow/_dataset_parquet.pyx
```diff
@@ -236,9 +236,9 @@ class RowGroupInfo:
 
     Parameters
     ----------
-    id : the group id.
-    metadata : the rowgroup metadata.
-    schema : schema of the rows.
+    id : the group id
```
wjones127, line 239: Similar to before, the place after the colon is for the type, so the description should be moved to the next line I think.

python/pyarrow/_dataset.pyx
amoeba: Good eye. Numpydoc catches the first issue (the period) but doesn't catch the ultimate issue which is a poorly-formed docstring. I was focusing too narrowly while fixing these. I don't think we have an automated way of catching this type of issue (a non-type sitting in the position a type would normally go).

python/pyarrow/_dataset_parquet.pyx
amoeba: Not sure what I was doing there. I'll add the period back in.

python/pyarrow/_dataset_parquet.pyx
amoeba: Same response as above too, though I think I can fix this in this PR.

amoeba:
Thanks for the review @wjones127. This is ready for another look.

python/pyarrow/ipc.pxi
```diff
@@ -57,11 +57,11 @@ class WriteStats(_WriteStats):
...
-    num_messages : number of messages.
-    num_record_batches : number of record batches.
-    num_dictionary_batches : number of dictionary batches.
-    num_dictionary_deltas : delta of dictionaries.
-    num_replaced_dictionaries : number of replaced dictionaries.
+    num_messages : number of messages
+    num_record_batches : number of record batches
+    num_dictionary_batches : number of dictionary batches
+    num_dictionary_deltas : delta of dictionaries
+    num_replaced_dictionaries : number of replaced dictionaries
```
wjones127, line 64:
```suggestion
    num_messages : int
        Number of messages.
    num_record_batches : int
        Number of record batches.
    num_dictionary_batches : int
        Number of dictionary batches.
    num_dictionary_deltas : int
        Delta of dictionaries.
    num_replaced_dictionaries : int
        Number of replaced dictionaries.
```

python/pyarrow/ipc.pxi
```diff
@@ -84,11 +84,11 @@ class ReadStats(_ReadStats):
...
```
wjones127, line 91:
```suggestion
    num_messages : int
        Number of messages.
    num_record_batches : int
        Number of record batches.
    num_dictionary_batches : int
        Number of dictionary batches.
    num_dictionary_deltas : int
        Delta of dictionaries.
    num_replaced_dictionaries : int
        Number of replaced dictionaries.
```

python/pyarrow/ipc.pxi
```diff
@@ -108,14 +108,14 @@ cdef class IpcReadOptions(_Weakrefable):
     ----------
     ensure_native_endian : bool
```
wjones127, line 109:
```suggestion
    ensure_native_endian : bool, default True
```

python/pyarrow/ipc.pxi
```diff
@@ -108,14 +108,14 @@ cdef class IpcReadOptions(_Weakrefable):
...
         Whether to convert incoming data to platform-native endianness.
-        Default is true.
+        Default is true
```
wjones127, line 111:
```suggestion
```

python/pyarrow/parquet/core.py
```diff
@@ -1309,7 +1313,7 @@ def get_index(self, key):
 
         Parameters
         ----------
-        key : The value for which we want to known the index.
+        key : The value for which we want to known the index
```
wjones127, line 1316:
```suggestion
        key : str or int
            The value for which we want to known the index.
```

python/pyarrow/parquet/core.py
```diff
@@ -1222,6 +1225,7 @@ def read(self, columns=None, use_threads=True, partitions=None,
         Returns
         -------
         table : pyarrow.Table
+            The pierce as a pyarrow.Table
```
wjones127, line 1228:
```suggestion
            The piece as a pyarrow.Table.
```

wjones127 (changes requested):
Saw a couple more instances of description in the type spot.

Also not going to correct the periods missing from end of description, since we aren't planning on enforcing PR09 and RT05, but did include that correct in my examples. (I know Joris has asked this of my PRs in the past.)

amoeba:
Thanks @wjones127, those look good. The help is appreciated :)

wjones127 (approved):

PR #15217, opened 2023-01-06
ARROW-18161: [Ruby] Refer source input in sub objects

kou:
`Arrow::Table` refers the source input to prevent garbage collecting the source input. If the source input is garbage collected, `Arrow::Table` touches invalid memory. It causes a crash.

`Arrow::Table` may return a sub object (table, record batch, chunked array and array) by its method. For example, `table[0].data` returns a chunked array of the first column. If the chunked array doesn't refer the source input and table is garbage collected, the chunked array touches invalid memory. It causes a crash.

We can avoid the crash by referring the source input from sub objects too.


kou:
+1

PR #15223, opened 2023-01-06
GH-15203: [Java] Implement writing compressed files

### LINKED ISSUE #15203: [Java] ArrowFileWriter/ArrowStreamWriter lack compression support

Their reader counterparts support compression, but the writers don't give you any way to enable it. Furthermore, the design of the compression API means that enabling this will require modifying several classes: the compression API closes the input buffer after compression, but this runs contrary to the expectation of VectorUnloader, which isn't supposed to mutate the source VectorSchemaRoot. (Arguably, this is a flaw in the VectorUnloader API that prevents efficient memory usage, but fixing _that_ would mean invalidating essentially all current Arrow Java code.)

Also there are various other flaws (e.g. VectorUnloader says it is OK to pass a null compression codec but will crash if you do so because it assumes the codec is never null.)

See also: #15102

It is possible to manually generate compressed data yourself in Java, but this requires taping together several low-level APIs yourself.

### Component(s)

Java

lidavidm:
* Closes: #15203

### DIFF
--- a/java/vector/src/main/java/org/apache/arrow/vector/ipc/message/ArrowRecordBatch.java
+++ b/java/vector/src/main/java/org/apache/arrow/vector/ipc/message/ArrowRecordBatch.java
@@ -76,10 +76,28 @@ public class ArrowRecordBatch implements ArrowMessage {
    * @param nodes   field level info
    * @param buffers will be retained until this recordBatch is closed
    * @param bodyCompression compression info.
+   * @param alignBuffers Whether to align buffers to an 8 byte boundary.
    */
   public ArrowRecordBatch(
       int length, List<ArrowFieldNode> nodes, List<ArrowBuf> buffers,
       ArrowBodyCompression bodyCompression, boolean alignBuffers) {
+    this(length, nodes, buffers, bodyCompression, alignBuffers, /*retainBuffers*/ true);
+  }
+
+  /**
+   * Construct a record batch from nodes.
+   *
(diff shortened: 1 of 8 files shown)


java/vector/src/main/java/org/apache/arrow/vector/ipc/message/ArrowRecordBatch.java
```diff
@@ -89,7 +107,9 @@ public ArrowRecordBatch(
     List<ArrowBuffer> arrowBuffers = new ArrayList<>(buffers.size());
     long offset = 0;
     for (ArrowBuf arrowBuf : buffers) {
-      arrowBuf.getReferenceManager().retain();
+      if (retainBuffers) {
```
davisusanibar, line 110: Could you try to explain why working with compression options impact or is needed to add this flag  "retainBuffers"? 

What will be the error if this flag retainBuffers is not added?

lidavidm: https://github.com/apache/arrow/blob/8199fa79a4440ca3976c5712ccf3bc3eaff0eae2/java/vector/src/main/java/org/apache/arrow/vector/VectorUnloader.java#L102-L105

lidavidm:
@lwhite1 would you also like to take a look?

davisusanibar (approved):

PR #15229, opened 2023-01-06
GH-15228: [R][CI] Add GitHub PAT to jobs that are reaching limit [WIP]

### LINKED ISSUE #15228: [R] [CI] Add Github PAT to jobs that are reaching limit

This passes the already extant github pat down to where it can be used during installation of R dependencies (which are hitting rate limits)

```
#12 899.5 > remotes::install_deps(dependencies = TRUE, INSTALL_opts = '')
#12 913.6 Error: HTTP error 403.
#12 913.6   API rate limit exceeded for 40.122.243.81. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
#12 913.6 
#12 913.6   Rate limit remaining: 0/60
#12 913.6   Rate limit reset at: 2023-01-06 00:59:48 UTC
#12 913.6 
#12 913.6   To increase your GitHub API rate limit
#12 913.6   - Use `usethis::create_github_token()` to create a Personal Access Token.
#12 913.6   - Use `usethis::edit_r_environ()` and add the token as `GITHUB_PAT`.
#12 913.6 Execution halted
#12 ERROR: executor failed running [/bin/bash -o pipefail -c /arrow/ci/scripts/r_deps.sh /arrow]: exit code: 1
------
 > [10/17] RUN /arrow/ci/scripts/r_deps.sh /arrow:
```
https://github.com/ursacomputing/crossbow/actions/runs/3851498487/jobs/6562738845#step:5:3350

### Component(s)

Continuous Integration, R

jonkeane:
This allows for authenticated github calls during the builds which should prevent us from being ratelimited
* Closes: #15228

### DIFF
--- a/ci/docker/linux-apt-r.dockerfile
+++ b/ci/docker/linux-apt-r.dockerfile
@@ -27,53 +27,50 @@ ENV R_PRUNE_DEPS=${r_prune_deps}
 ARG r_duckdb_dev=FALSE
 ENV R_DUCKDB_DEV=${r_duckdb_dev}
 
-ARG github_pat=""
-ENV GITHUB_PAT=${github_pat}
-
 # Build R
 # [1] https://www.digitalocean.com/community/tutorials/how-to-install-r-on-ubuntu-18-04
 # [2] https://linuxize.com/post/how-to-install-r-on-ubuntu-18-04/#installing-r-packages-from-cran
 ARG r=3.6
 RUN apt-get update -y && \
     apt-get install -y \
-    dirmngr \
-    apt-transport-https \
-    software-properties-common && \
+        dirmngr \
+        apt-transport-https \
+        software-properties-common && \
     wget -qO- https://cloud.r-project.org/bin/linux/ubuntu/marutter_pubkey.asc | \
-    tee -a /etc/apt/trusted.gpg.d/cran_ubuntu_key.asc && \
+        tee -a /etc/apt/trusted.gpg.d/cran_ubuntu_key.asc && \
     # NOTE: Only R >= 4.0 is available in this repo
     add-apt-repository 'deb https://cloud.r-project.org/bin/linux/ubuntu '$(lsb_release -cs)'-cran40/' && \
     apt-get install -y \
-    r-base=${r}* \
-    r-recommended=${r}* \
-    # system libs needed by core R packages
-    libxml2-dev \
-    libgit2-dev \
-    libssl-dev \
-    # install clang to mirror what was done on Travis
-    clang \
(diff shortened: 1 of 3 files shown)


jonkeane:
@github-actions crossbow submit test-r-depsource-system test-r-dev-duckdb test-r-ubuntu-22.04 test-r-gcc-11 test-r-gcc-12

jonkeane:
@github-actions crossbow submit test-r-depsource-system test-r-dev-duckdb test-r-ubuntu-22.04 test-r-gcc-11 test-r-gcc-12

jonkeane:
@github-actions crossbow submit test-r-depsource-system test-r-dev-duckdb test-r-ubuntu-22.04 test-r-gcc-11 test-r-gcc-12

ci/docker/linux-apt-r.dockerfile
```diff
@@ -27,50 +27,53 @@ ENV R_PRUNE_DEPS=${r_prune_deps}
 ARG r_duckdb_dev=FALSE
 ENV R_DUCKDB_DEV=${r_duckdb_dev}
 
+ARG github_pat=""
```
assignUser, line 30: :x: this permanently embeds the pat into the image. This is not super dangerous for the crossbow builds as those PATs expire after the job is run but people do run these jobs locally and could have this envvar exported for use in their local R things. (Also it is just bad practice, maybe we need to use an non expiring pat for some reason at some point ...)

Here the first link explaining the issue + solution I found https://medium.com/marionete/pass-secure-information-for-building-docker-images-8adeafe08355

assignUser (changes requested):
We need to use build secrets for this.

ci/docker/linux-apt-r.dockerfile
jonkeane: Cool, would you mind taking this over and doing the right thing?

kou (commented):
Could you revert needless changes?
(It seems that they are done by editor's auto formatter.)

jonkeane:
> Could you revert needless changes?
> (It seems that they are done by editor's auto formatter.)

Yes, of course. I ended up reverting that whole commit anyway since it's not the best way to do it. @assignUser has an idea for how to do this correctly + is going to push a better way to this branch with a better approach of making sure those github calls can be authenticated safely + get around the limits we're hitting. So for now I'll mark this PR as a WIP (or will close it if @assignUser prefers that)

assignUser:
I have looked at this again and the issue is not trivial due to our use of archery, I will explain in the issue and close this PR for now.

PR #15240, opened 2023-01-07
GH-15216: [C++][Parquet] Parquet writer accepts RecordBatch

### LINKED ISSUE #15216: [C++][Parquet] Parquet arrow::FileWriter accepts arrow::RecordBatch in the buffered row group mode

Current state of the parquet [arrow::FileWriter](https://github.com/apache/arrow/blob/master/cpp/src/parquet/arrow/writer.h#L53)
- It does not support buffered row group mode. User has to write the whole column chunk one by one.
- It only accepts `arrow::Table` in the tabular form, or `arrow::Array`/`arrow::ChunkedArray` in the columnar form. If one only has `arrow::RecordBatch` on hand, additional transformation is required before calling the writer.

The goal of this issue aims to enhance the arrow::FileWriter as below:

- [ ] Accept writing `arrow::RecordBatch`.
- [ ] Support buffered row group mode.

### Component(s)

C++, Parquet

wgtmac:
- Parquet arrow::FileWriter supports buffered row group mode.
 - Parquet arrow::FileWriter accepts arrow::RecordBatch.
* Closes: #15216

### DIFF
--- a/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
+++ b/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
@@ -4982,9 +4982,10 @@ TEST(TestReadWriteArrow, WriteAndReadRecordBatch) {
 
   // Prepare schema
   auto schema = ::arrow::schema(
-      {::arrow::field("a", ::arrow::int64()), ::arrow::field("b", ::arrow::utf8()),
-       ::arrow::field("c", ::arrow::struct_({::arrow::field("c1", ::arrow::int64()),
-                                             ::arrow::field("c2", ::arrow::utf8())}))});
+      {::arrow::field("a", ::arrow::int64()),
+       ::arrow::field("b", ::arrow::struct_({::arrow::field("b1", ::arrow::int64()),
+                                             ::arrow::field("b2", ::arrow::utf8())})),
+       ::arrow::field("c", ::arrow::utf8())});
   std::shared_ptr<SchemaDescriptor> parquet_schema;
   ASSERT_OK_NO_THROW(ToParquetSchema(schema.get(), *writer_properties,
                                      *arrow_writer_properties, &parquet_schema));
@@ -4992,18 +4993,18 @@ TEST(TestReadWriteArrow, WriteAndReadRecordBatch) {
 
   // Prepare data
   auto record_batch = ::arrow::RecordBatchFromJSON(schema, R"([
-      [1,     "alfa",  {"c1": -3,   "c2": "1"}],
-      [null,  "alfa",  {"c1": null, "c2": "22"}],
-      [3,     "beta",  {"c1": -2,   "c2": "333"}],
-      [null,  "gama",  {"c1": null, "c2": null}],
-      [5,     null,    {"c1": -1,   "c2": "-333"}],
-      [6,     "alfa",  {"c1": null, "c2": "-22"}],
-      [7,     "beta",  {"c1": 0,    "c2": "-1"}],
-      [8,     "beta",  {"c1": null, "c2": null}],
-      [9,     null  ,  {"c1": 1,    "c2": "0"}],
-      [null,  "gama",  {"c1": null, "c2": ""}],
-      [11,    "foo",   {"c1": 2,    "c2": "1234"}],
-      [12,    "bar",   {"c1": null, "c2": "4321"}]
+      [1,    {"b1": -3,   "b2": "1"   }, "alfa"],
+      [null, {"b1": null, "b2": "22"  }, "alfa"],
+      [3,    {"b1": -2,   "b2": "333" }, "beta"],
+      [null, {"b1": null, "b2": null  }, "gama"],
+      [5,    {"b1": -1,   "b2": "-333"}, null  ],
+      [6,    {"b1": null, "b2": "-22" }, "alfa"],
+      [7,    {"b1": 0,    "b2": "-1"  }, "beta"],
+      [8,    {"b1": null, "b2": null  }, "beta"],
+      [9,    {"b1": 1,    "b2": "0"   }, null  ],
+      [null, {"b1": null, "b2": ""    }, "gama"],
+      [11,   {"b1": 2,    "b2": "1234"}, "foo" ],
+      [12,   {"b1": null, "b2": "4321"}, "bar" ]
     ])");
 
   // Create writer to write data via RecordBatch.
--- a/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
+++ b/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
@@ -4895,7 +4895,8 @@ TYPED_TEST(TestIntegerAnnotateDecimalTypeParquetIO, SingleNullableDecimalColumn)
 template <typename TestType>
 class TestBufferedParquetIO : public TestParquetIO<TestType> {
  public:
-  void WriteBufferedFile(const std::shared_ptr<Array>& values, int64_t batch_size) {
+  void WriteBufferedFile(const std::shared_ptr<Array>& values, int64_t batch_size,
+                         int* num_row_groups) {
     std::shared_ptr<GroupNode> schema =
         MakeSimpleSchema(*values->type(), Repetition::OPTIONAL);
     SchemaDescriptor descriptor;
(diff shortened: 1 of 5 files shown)


wgtmac:
@emkornfield @pitrou @wjones127 Could you please take a look? Thanks!

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -4727,5 +4727,111 @@ std::vector<NestedFilterTestCase> GenerateMapFilteredTestCases() {
...
+};
+
+TYPED_TEST_SUITE(TestBufferedParquetIO, TestTypes);
+
+TYPED_TEST(TestBufferedParquetIO, SingleColumnOptionalBufferedWriteSmall) {
+  constexpr int64_t batch_size = SMALL_SIZE / 4;
+  std::shared_ptr<Array> values;
+  ASSERT_OK(NullableArray<TypeParam>(SMALL_SIZE, 10, kDefaultSeed, &values));
+  this->WriteBufferedFile(values, batch_size);
+  ASSERT_NO_FATAL_FAILURE(this->ReadAndCheckSingleColumnFile(*values));
```
wjones127, line 4766: IIUC, the purpose of using `NewBufferedRowGroup()` between calls to `WriteRecordBatch()` is to write multiple batches to the same row group, right? Could you add an assertion here that the final number of row groups output is correct?

wjones127: Also, do we test what happens when we write a record batch that is larger than `WriterProperties.max_row_group_length`?

cpp/src/parquet/arrow/writer.cc
```diff
@@ -355,6 +381,40 @@ class FileWriterImpl : public FileWriter {
...
+    auto WriteBatch = [&](int64_t offset, int64_t size) {
+      int column_index_start = 0;
+      for (int i = 0; i < batch.num_columns(); i++) {
+        ChunkedArray chunkedArray(batch.column(i));
+        ARROW_ASSIGN_OR_RAISE(
+            std::unique_ptr<ArrowColumnWriterV2> writer,
+            ArrowColumnWriterV2::Make(chunkedArray, offset, size, schema_manifest_,
+                                      row_group_writer_, column_index_start));
+        RETURN_NOT_OK(writer->Write(&column_write_context_));
+        column_index_start += writer->leaf_count();
```
wjones127, line 410: I think this seems correct, but it would be good to test this path with data containing a struct array.

cpp/src/parquet/arrow/writer.h
```diff
@@ -114,6 +115,14 @@ class PARQUET_EXPORT FileWriter {
   virtual ::arrow::Status WriteColumnChunk(
       const std::shared_ptr<::arrow::ChunkedArray>& data) = 0;
 
+  /// \brief Start a new buffered row group.
+  ///
+  /// Returns an error if not all columns have been written.
+  virtual ::arrow::Status NewBufferedRowGroup() = 0;
+
+  /// \brief Write a RecordBatch into the buffered row group.
```
wjones127, line 123: Should clarify:

 * That multiple batches can be written to the same row group.
 * What happens when you write more data than `WriterProperties.max_row_group_length`?
 * When do the batches get flushed to the output stream?

```suggestion
  /// \brief Write a RecordBatch into the buffered row group.
  ///
  /// Multiple RecordBatches can be written into the same row
  /// group through this method. 
```

wjones127 (changes requested):
This looks very useful! I have noted a few places where we could use a bit more tests and documentation to clarify some behavior.

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
wgtmac: Changed the writer to respect `WriterProperties.max_row_group_length` in the `WriterRecordBatch`. Also, I have changed the test case to check number of row groups created.

cpp/src/parquet/arrow/writer.cc
wgtmac: Fixed

cpp/src/parquet/arrow/writer.h
wgtmac: Added more comments to clarify all behaviors mentioned above.

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -4727,5 +4727,153 @@ std::vector<NestedFilterTestCase> GenerateMapFilteredTestCases() {
...
+      [3,     "beta",  {"c1": -2,   "c2": "333"}],
+      [null,  "gama",  {"c1": null, "c2": null}],
+      [5,     null,    {"c1": -1,   "c2": "-333"}],
+      [6,     "alfa",  {"c1": null, "c2": "-22"}],
+      [7,     "beta",  {"c1": 0,    "c2": "-1"}],
+      [8,     "beta",  {"c1": null, "c2": null}],
+      [9,     null  ,  {"c1": 1,    "c2": "0"}],
+      [null,  "gama",  {"c1": null, "c2": ""}],
+      [11,    "foo",   {"c1": 2,    "c2": "1234"}],
+      [12,    "bar",   {"c1": null, "c2": "4321"}]
```
wjones127, line 4841: In order to exercise that `leaf_count` code on the struct array, we need another array after (since it's added at the end of the loop below). So I suggest move the string to the end.

```suggestion
      [1,    {"b1": -3,   "b2": "1"}],    "alfa"]
      [null, {"b1": null, "b2": "22"}],   "alfa"]
      [3,    {"b1": -2,   "b2": "333"}],  "beta"]
      [null, {"b1": null, "b2": null}],   "gama"]
      [5,    {"b1": -1,   "b2": "-333"}], null, ]
      [6,    {"b1": null, "b2": "-22"}],  "alfa"]
      [7,    {"b1": 0,    "b2": "-1"}],   "beta"]
      [8,    {"b1": null, "b2": null}],   "beta"]
      [9,    {"b1": 1,    "b2": "0"}],    null  ]
      [null, {"b1": null, "b2": ""}],     "gama"]
      [11,   {"b1": 2,    "b2": "1234"}], "foo",]
      [12,   {"b1": null, "b2": "4321"}]  "bar",]
```

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -4727,5 +4727,153 @@ std::vector<NestedFilterTestCase> GenerateMapFilteredTestCases() {
...
+  auto sink = CreateOutputStream();
+  // Limit the max number of rows in a row group to 10
+  auto writer_properties = WriterProperties::Builder().max_row_group_length(10)->build();
+  auto arrow_writer_properties = default_arrow_writer_properties();
+
+  // Prepare schema
+  auto schema = ::arrow::schema(
+      {::arrow::field("a", ::arrow::int64()), ::arrow::field("b", ::arrow::utf8()),
+       ::arrow::field("c", ::arrow::struct_({::arrow::field("c1", ::arrow::int64()),
+                                             ::arrow::field("c2", ::arrow::utf8())}))});
```
wjones127, line 4822:
```suggestion
  auto schema = ::arrow::schema(
      {::arrow::field("a", ::arrow::int64()),
       ::arrow::field("b", ::arrow::struct_({::arrow::field("b1", ::arrow::int64()),
                                             ::arrow::field("b2", ::arrow::utf8())})),
       ::arrow::field("c", ::arrow::utf8())});
```

wjones127 (changes requested):
Looking very good. I suggest one small change to a test so we use that `leaf_count()` call within the loop properly.

wgtmac:
> Looking very good. I suggest one small change to a test so we use that `leaf_count()` call within the loop properly.

All comments have been addressed. Please take a look again. Thanks!

wjones127:
Looks good. @wgtmac could you rebase? Have conflicts since I merged the other PR. Once that's complete, this should be good to merge.

wgtmac:
> Looks good. @wgtmac could you rebase? Have conflicts since I merged the other PR. Once that's complete, this should be good to merge.

@wjones127 Thanks for your review!

I have rebased and updated the PR. Please check again.

wjones127 (approved):
Excellent! Thank you @wgtmac

PR #15241, opened 2023-01-07
GH-14923: [C++][Parquet] Fix DELTA_BINARY_PACKED problem on reading the last block with malford bit-width

### LINKED ISSUE #14923: [C++][Parquet] DeltaBitPackDecoder expects all miniblock bitwidths to be present for the last block

This is a follow-up to [this discussion](https://github.com/apache/arrow/pull/14191#discussion_r1042303944).
Arrow's reader is not compliant with the spec, and requires all miniblock bitwidth values of the last block even if not necessary. This forces arrow writer to write all miniblock bitwidths to be able to roundtrip which is not a big issue. However it does mean Arrow reader would not be able to read bitpacked buffer written by another parquet writer that does not write all miniblock bitwidths.

### Component(s)

C++, Parquet

mapleFU:
Problem is mentioned here: https://github.com/apache/arrow/issues/14923

This patch fixes that issue. And the code is a bit complex.
- Rename variables, since original name is confusing for me. 
  - `block_initialized_` -> `first_block_initialized_`, because it's a mask that indicates that if the first block in page is initialized.
  - `total_value_count_` -> `total_values_remaining_`. Because it's not `total values within a page`, it means `remaing values to be decoded within a page`
  - `values_count_current_mini_block_` -> `values_remaining_current_mini_block_`, ditto
- Add variables
  - `total_value_count_`: the total value numbers within a page.
- Change Syntax
  -  Change `InitBlock()` to `InitBlock()` and `InitMiniBlock`
- Implemention, most logic is in `InitBlock()` and `InitMiniBlock`
- Testing. Thanks @rok. I use a page within 65 values with bitwidth `32 32 165 165`.

And personally, I use the code here for testing:

```c++
   for (uint32_t i = num_miniblocks; i < mini_blocks_per_block_; i++) {
-    bit_width_data[i] = 0;
+    //    bit_width_data[i] = 0;
+    bit_width_data[i] = static_cast<uint8_t>(random());
   }
```

The code works well in both debug and release mode.

* Closes: #14923

### DIFF
--- a/cpp/src/parquet/encoding.cc
+++ b/cpp/src/parquet/encoding.cc
@@ -2453,16 +2453,25 @@ class DeltaBitPackDecoder : public DecoderImpl, virtual public TypedDecoder<DTyp
     if (!decoder_->GetZigZagVlqInt(&min_delta_)) ParquetException::EofException();
 
     // read the bitwidth of each miniblock
+    uint32_t miniblock_values_sum = 0;
     uint8_t* bit_width_data = delta_bit_widths_->mutable_data();
-    for (uint32_t i = 0; i < mini_blocks_per_block_; ++i) {
-      if (!decoder_->GetAligned<uint8_t>(1, bit_width_data + i)) {
+    for (uint32_t current_mini_block_idx = 0; current_mini_block_idx < mini_blocks_per_block_; ++current_mini_block_idx) {
+      if (!decoder_->GetAligned<uint8_t>(1, bit_width_data + current_mini_block_idx)) {
         ParquetException::EofException();
       }
-      if (bit_width_data[i] > kMaxDeltaBitWidth) {
-        throw ParquetException("delta bit width " + std::to_string(bit_width_data[i]) +
-                               " larger than integer bit width " +
-                               std::to_string(kMaxDeltaBitWidth));
+
+      if (bit_width_data[current_mini_block_idx] > kMaxDeltaBitWidth) {
+        if (miniblock_values_sum < total_value_count_) {
+          throw ParquetException(
+              "delta bit width " +
+              std::to_string(
+                  bit_width_data[current_mini_block_idx * values_current_mini_block_]) +
+              " larger than integer bit width " + std::to_string(kMaxDeltaBitWidth));
+        } else {
+          break;
+        }
       }
+      miniblock_values_sum += values_current_mini_block_;
     }
     mini_block_idx_ = 0;
     delta_bit_width_ = bit_width_data[0];
--- a/cpp/src/parquet/encoding.cc
+++ b/cpp/src/parquet/encoding.cc
@@ -2446,7 +2446,7 @@ class DeltaBitPackDecoder : public DecoderImpl, virtual public TypedDecoder<DTyp
 
     delta_bit_widths_ = AllocateBuffer(pool_, mini_blocks_per_block_);
     first_block_initialized_ = false;
-    value_sum_up_to_current_block_ = 0;
+    values_num_up_to_current_block_ = 0;
     values_current_mini_block_ = 0;
   }
 
@@ -2455,7 +2455,7 @@ class DeltaBitPackDecoder : public DecoderImpl, virtual public TypedDecoder<DTyp
 
     // read the bitwidth of each miniblock
     uint8_t* bit_width_data = delta_bit_widths_->mutable_data();
-    uint32_t miniblock_values_sum = value_sum_up_to_current_block_;
+    uint32_t miniblock_values_sum = values_num_up_to_current_block_;
     for (uint32_t current_mini_block_idx = 0;
          current_mini_block_idx < mini_blocks_per_block_; ++current_mini_block_idx) {
       if (!decoder_->GetAligned<uint8_t>(1, bit_width_data + current_mini_block_idx)) {
@@ -2476,7 +2476,7 @@ class DeltaBitPackDecoder : public DecoderImpl, virtual public TypedDecoder<DTyp
       }
       miniblock_values_sum += values_per_mini_block_;
     }
-    value_sum_up_to_current_block_ = std::min(miniblock_values_sum, total_value_count_);
+    values_num_up_to_current_block_ = std::min(miniblock_values_sum, total_value_count_);
     mini_block_idx_ = 0;
     delta_bit_width_ = bit_width_data[0];
     values_current_mini_block_ = values_per_mini_block_;
@@ -2495,8 +2495,8 @@ class DeltaBitPackDecoder : public DecoderImpl, virtual public TypedDecoder<DTyp
         if (ARROW_PREDICT_FALSE(!first_block_initialized_)) {
           buffer[i++] = last_value_;
           DCHECK_EQ(i, 1);  // we're at the beginning of the page
-          DCHECK_EQ(value_sum_up_to_current_block_, 0);
-          value_sum_up_to_current_block_ = 1;
+          DCHECK_EQ(values_num_up_to_current_block_, 0);
+          values_num_up_to_current_block_ = 1;
           if (ARROW_PREDICT_FALSE(i == max_values)) {
             // When block is uninitialized and i reaches max_values we have two
             // different possibilities:
@@ -2561,9 +2561,13 @@ class DeltaBitPackDecoder : public DecoderImpl, virtual public TypedDecoder<DTyp
   uint32_t total_value_count_;
 
   // If the page doesn't contain any block, `first_block_initialized_` will
-  // always be false.
+  // always be false. Otherwise, it will be true when first block initialized.
   bool first_block_initialized_;
-  uint32_t value_sum_up_to_current_block_;
+  // The sum of values up to current block.
+  // If the page doesn't contain any value, it will always be 0.
+  // If the page only contains one value, it would be 1 if value loaded.
+  // Otherwise, it would be the sum value up to "current" block.
+  uint32_t values_num_up_to_current_block_;
   T min_delta_;
   uint32_t mini_block_idx_;
   std::shared_ptr<ResizableBuffer> delta_bit_widths_;
--- a/cpp/src/parquet/encoding.cc
+++ b/cpp/src/parquet/encoding.cc
@@ -2447,26 +2447,26 @@ class DeltaBitPackDecoder : public DecoderImpl, virtual public TypedDecoder<DTyp
     total_values_remaining_ = total_value_count_;
     delta_bit_widths_ = AllocateBuffer(pool_, mini_blocks_per_block_);
     first_block_initialized_ = false;
-    values_num_up_to_current_mini_block_ = 0;
     values_remaining_current_mini_block_ = 0;
   }
 
-  void InitBlock() {
+  void InitBlock(int prefix_values) {
     if (!decoder_->GetZigZagVlqInt(&min_delta_))
       ParquetException::EofException("InitBlock EOF");
 
     // read the bitwidth of each miniblock
     uint8_t* bit_width_data = delta_bit_widths_->mutable_data();
-    uint32_t miniblock_values_sum = values_num_up_to_current_mini_block_;
+    // If the current block contains the last mini block,
+    // current_num_values may greater than total_values_remaining_.
+    uint32_t current_num_values = prefix_values;
     for (uint32_t current_mini_block_idx = 0;
          current_mini_block_idx < mini_blocks_per_block_; ++current_mini_block_idx) {
       if (!decoder_->GetAligned<uint8_t>(1, bit_width_data + current_mini_block_idx)) {
         ParquetException::EofException("Decode bit-width EOF");
       }
-
       if (ARROW_PREDICT_FALSE(bit_width_data[current_mini_block_idx] >
                               kMaxDeltaBitWidth)) {
-        if (miniblock_values_sum < total_values_remaining_) {
+        if (current_num_values < total_values_remaining_) {
           throw ParquetException("delta bit width " +
                                  std::to_string(bit_width_data[current_mini_block_idx]) +
                                  " larger than integer bit width " +
@@ -2475,7 +2475,7 @@ class DeltaBitPackDecoder : public DecoderImpl, virtual public TypedDecoder<DTyp
         // according to the parquet standard, we should ignore the bit_width_data here.
         // cannot break because still need to read remaining bit-width from decoder.
       }
-      miniblock_values_sum += values_per_mini_block_;
+      current_num_values += values_per_mini_block_;
     }
     mini_block_idx_ = 0;
     delta_bit_width_ = bit_width_data[0];
@@ -2495,8 +2495,6 @@ class DeltaBitPackDecoder : public DecoderImpl, virtual public TypedDecoder<DTyp
         if (ARROW_PREDICT_FALSE(!first_block_initialized_)) {
           buffer[i++] = last_value_;
           DCHECK_EQ(i, 1);  // we're at the beginning of the page
-          DCHECK_EQ(values_num_up_to_current_mini_block_, 0);
-          values_num_up_to_current_mini_block_ = 1;
           if (ARROW_PREDICT_FALSE(i == max_values)) {
             // When block is uninitialized and i reaches max_values we have two
             // different possibilities:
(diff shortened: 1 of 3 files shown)


mapleFU:
@pitrou @rok PTAL (^_^)

mapleFU:
Wait, it's still some bug, sorry. Let me fix it

mapleFU:
I change the code to 

```c++
    for (uint32_t j = values_current_mini_block; j < values_per_mini_block_; j++) {
      bit_writer_.PutValue(random(), bit_width);
    }
```

And run multiple times with ASAN, test still passed. Though I don't know how to mock in release, seens we cannot use gmock when `DeltaBitPackEncoder` is always in `.cc` file.

mapleFU:
@pitrou @rok PTAL (^_^) I think I fix it, and leaving some comments.

mapleFU:
Well, it's a bit weird.

```c++
TEST_F(DeltaBitPackEncoding, MalfordMiniblockBitWidth) {
  std::shared_ptr<ColumnDescriptor> descr_ = ExampleDescr<Int32Type>();
  auto decoder = MakeTypedDecoder<Int32Type>(Encoding::DELTA_BINARY_PACKED, descr_.get());
  using c_type = parquet::Int32Type::c_type;

  unsigned char good_data[] = "\200\001\004A\237\224\316\362\r\242\220\203-  ";
  int encode_buffer_size = 273;
  int num_values = 65;
  std::vector<uint8_t> output_bytes = std::vector<uint8_t>(num_values * sizeof(c_type));
  auto decode_buf = reinterpret_cast<c_type*>(output_bytes.data());
  decoder->SetData(num_values, &good_data[0], encode_buffer_size);
```

@rok How do you build the `good_data`? Seems `encode_buffer_size = 273` is too large for `good_data`, and in `Release` mode, memcpy will reach the stack guard, and raise an asan exception.

mapleFU:
Well, seems that both buffer is 273B. And I need to padding the `good_data`. @rok would you mind add the generating method of these two datas? Seems leaving a bit-string here is confusing.

mapleFU:
This time it failed because unpack32_avx512 meets an ASAN issue. Maybe I need to know how input data is generated and do some adjustions. Go to sleep now.

mapleFU:
(By the way, my PC using AMD's CPU, which I can't reproduce unpack32_avx512 problem. So it's important for me to know how test data generated...)

rok:
Hey @mapleFU, sorry for not replying earlier, I am preoccupied with the Jira->Github migration :).
I generated the data by applying [this diff](https://github.com/apache/arrow/issues/14923#issuecomment-1366657071) and stepping through GDB in CLion probably up [to here](https://github.com/apache/arrow/blob/6c35c4ae4124a5012234ac485ea1ee22f2a5259a/cpp/src/parquet/encoding_test.cc#L1340) and copied the data. I am on AMD Ryzen 7 5700G, Ubuntu 22.02.
I can try getting the data string in some other way as GDB could be interfering somehow?

mapleFU:
Sorry for being crazy and at you for multiple times. I finally make clear the meaning of them, and put the meaning in `encoding_test.cc`:

```
  // Both good_data and bad_data is a DELTA_BINARY_PACKED INT32 Page with 65 values.
  // For needed miniblocks, there bit-widths are all 32.
  // There bit-widths for unneeded miniblocks are different:
  // * good_data's unneeded bit-width is 0.
  // * bad_data's unneeded bit-width is 165.
```

mapleFU:
Test regenerated, and representing using Hex.

mapleFU:
@pitrou @wjones127 Mind take a look? Since I made lots of naming changes( described https://github.com/apache/arrow/pull/15241#issue-1523751523 ), and update a Hex testing format. Don't know whether it's ok.

cpp/src/parquet/encoding.cc
```diff
@@ -2542,10 +2558,21 @@ class DeltaBitPackDecoder : public DecoderImpl, virtual public TypedDecoder<DTyp
   uint32_t values_per_block_;
   uint32_t mini_blocks_per_block_;
   uint32_t values_per_mini_block_;
-  uint32_t values_current_mini_block_;
   uint32_t total_value_count_;
 
-  bool block_initialized_;
+  uint32_t total_value_remaining_;
```
pitrou, line 2563:
```suggestion
  uint32_t total_values_remaining_;
```

cpp/src/parquet/encoding.cc
```diff
@@ -2542,10 +2558,21 @@ class DeltaBitPackDecoder : public DecoderImpl, virtual public TypedDecoder<DTyp
...
+  // Remaining values in current mini block. If the current block is the last mini block,
+  // values_remaining_current_mini_block_ may greater than total_value_remaining_.
+  uint32_t values_remaining_current_mini_block_;
+
+  // If the page doesn't contain any block, `first_block_initialized_` will
+  // always be false. Otherwise, it will be true when first block initialized.
+  bool first_block_initialized_;
+  // The sum of values up to current block.
```
pitrou, line 2571:
```suggestion
  // The number of values up to current mini-block (non including).
```

cpp/src/parquet/encoding_test.cc
```diff
@@ -1427,5 +1429,66 @@ TYPED_TEST(TestDeltaBitPackEncoding, BasicRoundTrip) {
...
+  auto decoder = MakeTypedDecoder<Int32Type>(Encoding::DELTA_BINARY_PACKED, descr.get());
+  using c_type = parquet::Int32Type::c_type;
+
+  int encode_buffer_size = 274;
+  int num_values = 65;
+  c_type* decode_buf = nullptr;
+  std::vector<uint8_t> output_bytes;
+  int values_decoded = 0;
+
+  // Both good_data and bad_data is a DELTA_BINARY_PACKED INT32 Page with 65 values.
```
pitrou, line 1443: Both of these data are "good", neither is "bad".

cpp/src/parquet/encoding_test.cc
```diff
@@ -1427,5 +1429,66 @@ TYPED_TEST(TestDeltaBitPackEncoding, BasicRoundTrip) {
...
+  // For needed miniblocks, there bit-widths are all 32.
+  // There bit-widths for unneeded miniblocks are different:
+  // * good_data's unneeded bit-width is 0.
+  // * bad_data's unneeded bit-width is 165.
+  // We use Hex to represent good_data and bad_data.
+  {
+    char good_data_hex[] =
```
pitrou, line 1450: Should use `std::string` or `std::string_view` to have length information.

cpp/src/parquet/encoding_test.cc
```diff
@@ -1427,5 +1429,66 @@ TYPED_TEST(TestDeltaBitPackEncoding, BasicRoundTrip) {
...
+        "80010441BDA08DB708A488C0F1012020000089F79E53946D93CA242D48B21478591FDE6E4855CC03"
+        "7024A4E8E56E1D20687C8BFB4DE880E73CE2E49E873BEB0B1DAD0E0169C7951C411CA76AD949D8ED"
+        "FD9E07A485DF4410B4486253A6280335BB87494B7D61E40A8D487731BBC83A37E1E8EE2F5431150C"
+        "F46D7224B3C7C271AC3220F2C6663289427E1CE3B00F8C34D5DDA845664AF36F56ACF6FED2339911"
+        "7AB54F6667A80EA78D08DC61B4262CE28960B46B2D93785ED9933175A910415B50B30E9889C70ECD"
+        "84C8856531E78B4E05B2AB27B1233E4BAB722869D785F5B8B1049955EFE200000000C6C2201E1836"
+        "C0579B202A6A7395C761044AD151D5CEE992CBBA5AE937E54F29D28F86CC0627B9D1";
+    unsigned char good_data[274];
```
pitrou, line 1458: Please don't hard code values like that. Instead you can use a `std::string` or a `std::vector<uint8_t>`.

cpp/src/parquet/encoding_test.cc
```diff
@@ -1427,5 +1429,66 @@ TYPED_TEST(TestDeltaBitPackEncoding, BasicRoundTrip) {
...
+    for (int i = 0; i < 274; ++i) {
+      auto s = ::arrow::ParseHexValue(&good_data_hex[i * 2], &good_data[i]);
+      ASSERT_TRUE(s.ok());
```
pitrou, line 1461: ASSERT_OK

cpp/src/parquet/encoding_test.cc
```diff
@@ -1427,5 +1429,66 @@ TYPED_TEST(TestDeltaBitPackEncoding, BasicRoundTrip) {
...
+
+  {
+    char bad_data_hex[] =
+        "80010441BDA08DB708A488C0F1012020A5A589F79E53946D93CA242D48B21478591FDE6E4855CC03"
+        "7024A4E8E56E1D20687C8BFB4DE880E73CE2E49E873BEB0B1DAD0E0169C7951C411CA76AD949D8ED"
+        "FD9E07A485DF4410B4486253A6280335BB87494B7D61E40A8D487731BBC83A37E1E8EE2F5431150C"
+        "F46D7224B3C7C271AC3220F2C6663289427E1CE3B00F8C34D5DDA845664AF36F56ACF6FED2339911"
+        "7AB54F6667A80EA78D08DC61B4262CE28960B46B2D93785ED9933175A910415B50B30E9889C70ECD"
+        "84C8856531E78B4E05B2AB27B1233E4BAB722869D785F5B8B1049955EFE200000000C6C2201E1836"
+        "C0579B202A6A7395C761044AD151D5CEE992CBBA5AE937E54F29D28F86CC0627B9D1";
```
pitrou, line 1479: Is this different from the other data above? IIUC, the trailing bytes should be different, but they seem identical.

cpp/src/parquet/encoding_test.cc
```diff
@@ -1427,5 +1429,66 @@ TYPED_TEST(TestDeltaBitPackEncoding, BasicRoundTrip) {
...
+    unsigned char bad_data[274];
+    for (int i = 0; i < 274; ++i) {
```
pitrou, line 1481: Please don't copy/paste code while it could trivially be factored out.

pitrou (commented):
It's not obvious to me why you're making so many changes to the decoder. Was the suggested fix in https://github.com/apache/arrow/issues/14923#issuecomment-1369902019 not sufficient?

cpp/src/parquet/encoding_test.cc
mapleFU: Other data is equal. (In fact, they're all generated by `this->Execute(65, 1)` )

mapleFU:
In most case, https://github.com/apache/arrow/issues/14923#issuecomment-1369902019 is right. But we may have the case above:

```c++
if (total_value_count_ != 1) {
  InitBlock();
}
```

Here handling it would becoming trickey. And update so much code because some name really makes me feel confusing. But maybe I can make it simplier

pitrou:
I don't understand, what would be tricky?

mapleFU:
Let's go back to the original code:

```c++
      if (ARROW_PREDICT_FALSE(values_remaining_current_mini_block_ == 0)) {
        if (ARROW_PREDICT_FALSE(!first_block_initialized_)) {
          buffer[i++] = last_value_;
          DCHECK_EQ(i, 1);  // we're at the beginning of the page
          DCHECK_EQ(values_num_up_to_current_mini_block_, 0);
          values_num_up_to_current_mini_block_ = 1;
          if (ARROW_PREDICT_FALSE(i == max_values)) {
            // When block is uninitialized and i reaches max_values we have two
            // different possibilities:
            // 1. total_value_count_ == 1, which means that the page may have only
            // one value (encoded in the header), and we should not initialize
            // any block.
            // 2. total_value_count_ != 1, which means we should initialize the
            // incoming block for subsequent reads.
            if (total_value_count_ != 1) {
              InitBlock();
            }
            break;
          }
          InitBlock();
        }
```

And assume the page has 33 values:
1. `total_value_count_ == 1`, call `InitBlock`
2. `current_num_values = 0`, read block one, change to 32.
3. `current_num_values = 32`, read block one, change to 64.

But here, it should be `1` firstly.

pitrou:
Why not, but can you add a test that would reproduce the issue?

mapleFU:
> Why not, but can you add a test that would reproduce the issue?

Yes, but adding test case is trickey... I cannot use `gmock` to mock writing non-zero-values...

pitrou:
> Sure, it's easy to write that in my personal develop enviroment. But I cannot submit these "mock" code to the codebase, which make writing unit test a bit more trickey.

You can write the code that mutates the encoder output directly in the unit test. Schematically:
```c++
std::vector<uint8_t> encoded = ...;
for (auto i = encoded.length() - 33; i < encoded.length() - 2; ++i) {
  encoded[i] = 0xFF;
}
```

> By the way, do you think using Hex to represent binary is ok?

That works, yes.

mapleFU:
Well, I found that rok's testing data already covers that case:

And assume the page has 65 values and 32 values per-mini-block:

```
total_value_count_ == 1, call InitBlock
current_num_values = 0, read block one, change to 32.
current_num_values = 32, read block one, change to 64.
current_num_values = 64, read block one, change to 96.
```

When changing to `InitBlock(is_init_page)`, I meet another trickey problem:

```c++
while (i < max_values) {
  ... decode values from miniblocks
}
total_values_remaining_ -= max_values;
```

So, `InitBlock()` should have a counter, or it should decrease `total_values_remaining_` per loop.

I think there are some fixing:
1. current my implemention is ok here, because it maintains a `current miniblock counter`
2. The change like below:

```c++
while (i < max_values) {
  ... decode values from miniblocks, when i increase, decrease remaining
  ... or passing i into InitBlock
}
CHECK_EQ(i, max_values);
```

@pitrou what's your optionion here? Maintaining a `current value counter` or `passing i into InitBlock`, like `InitBlock(int prefix_counter)`, or decrease `remaining counter` every time we inc?

(Personally I'd like to call `InitBlock(int prefix_counter)`, but I'd like to know what's your opinion)

mapleFU:
I change my implemention to use `i` counter. If you have some ideas you can edit my patch directly https://github.com/apache/arrow/pull/15241#issue-1523751523

mapleFU:
(By the way, no idea why Travis CI failed...)

pitrou:
Ok, I was not satisfied with the test and implementation, so I ended up rewriting them.

mapleFU:
I update the description here: https://github.com/apache/arrow/pull/15241#issue-1523751523

If you like, you can edit it for your new changes

rok (approved):
This looks great. I didn't realize ULEB128 values were byte aligned. Good to know!
Thanks for doing this @mapleFU & @pitrou and sorry I was not able to contribute more.

pitrou:
Hmm, interestingly, there are test failures...

mapleFU:
> Hmm, interestingly, there are test failures...

@pitrou I think I got why it failed, when read batches data, previously, we will point miniblock to a "bad" miniblock, and never read from it. But now, our code may `InitMiniBlock` for last "bad" miniblock, causing the failure.

mapleFU:
```c++
        } else {
          ++mini_block_idx_;
          if (mini_block_idx_ < mini_blocks_per_block_) {
            InitMiniBlock(delta_bit_widths_->data()[mini_block_idx_]);
          } else {
            InitBlock();
          }
        }
```

Previously, `bit_width_` will point to a "bad" width, but it will never been read. Now, `InitMiniBlock` will read it

mapleFU:
I guess change the code like this would work:

```c++
          if (mini_block_idx_ < mini_blocks_per_block_) {
            if (i >= total_values_remaining_) {
                break;
            }
            InitMiniBlock(delta_bit_widths_->data()[mini_block_idx_]);
          } else {
            InitBlock();
          }
```

@pitrou

mapleFU:
On my MacOs:

```
[DELTA_BINARY_PACKED] values_per_block: 128, mini_blocks_per_block: 4, total_value_count: 2, last_value: -49
Header consume 5 bytes
Block mini delta is: 83, consume bytes: 2
miniblock bitwidth: 255
miniblock bitwidth: 255
miniblock bitwidth: 255
miniblock bitwidth: 0
miniblock bitwidth consume bytes: 4
```

pitrou:
Thanks @mapleFU !

mapleFU:
@wjones127 Mind take a look :) ?

mapleFU:
@pitrou would you mind merge this patch? Or I should wait @wjones127  to take a look?

pitrou:
I'll wait a bit for @wjones127 to chime in. Otherwise I'll probably merge tomorrow.

PR #15245, opened 2023-01-07
ARROW-18377: MIGRATION: Automate component labels from issue form content

assignUser:
tested on fork: https://github.com/assignUser/arrow/issues/9

It only triggers on the initial opening of the issue so we can make sane assumptions of the layout (that could be broken by editing it).

### DIFF
--- a/.github/workflows/issue_bot.yml
+++ b/.github/workflows/issue_bot.yml
@@ -1,0 +1,63 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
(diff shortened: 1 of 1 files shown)


.github/workflows/issue_bot.yml
```diff
@@ -0,0 +1,63 @@
...
+jobs:
+  label_components:
+    name: Label Components
+    if: github.event.issue.pull_request == null
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v6
+        with:
+          script: |
+            let split_body = context.payload.issue.body.split('### Component(s)');
```
rok, line 38: How about capturing a regex e.g. `### Component\(s\)\n(.+)\n` instead?

rok:
Thanks for doing this @assignUser !

.github/workflows/issue_bot.yml
assignUser: We could but it doesn't really bring any advantages here imo :D

rok: Only if the template text changes. :)
I'm ok with this too.

rok (approved):

.github/workflows/issue_bot.yml
rok: Shall we merge?

assignUser: :rocket:

rok:
Jira seems unresponsive when I run the merge script. I might be throttled due to all my migration API abuse. Will retry later.

rok:
I went through the current issues and labelled them as originally stated (where it made sense).

PR #15252, opened 2023-01-08
GH-15246: [R] Now saving row names

### LINKED ISSUE #15246: [R] arrow::write_feather doesn't save row names

R code snippet:
```r
> df <- data.frame(a=1:3,b=4:6)
> rownames(df) <- c("row1","row2","row3")
> df
     a b
row1 1 4
row2 2 5
row3 3 6
> arrow::write_feather(x = df, sink="~/tmp/df")
> df2 <- arrow::read_feather(file = "~/tmp/df")
> df2
  a b
1 1 4
2 2 5
3 3 6
```

### Component(s)

R

OfekShilon:
Not sure why `row.names` was ever removed.
[This is the commit](https://github.com/sirensolutions/arrow/commit/a6d531a3f409706d7dcde4562913c8589be3402a#diff-659e9fa6b66e5a72b4e3f9ac79ffddf08f92d9ea3d7aa45bd8c73b9a022fa2e5) that added the entire metadata.R, it seems a part of an unrelated effort. 
I believe the other attributes listed in `removed_attributes` were removed because they're serialized/deserialized by other means. `row.names` doesn't belong there.

* Closes: #15246

### DIFF
--- a/r/R/metadata.R
+++ b/r/R/metadata.R
@@ -125,9 +125,9 @@ apply_arrow_r_metadata <- function(x, r_metadata) {
 remove_attributes <- function(x) {
   removed_attributes <- character()
   if (identical(class(x), c("tbl_df", "tbl", "data.frame"))) {
-    removed_attributes <- c("class", "row.names", "names")
+    removed_attributes <- c("class", "names")
   } else if (inherits(x, "data.frame")) {
-    removed_attributes <- c("row.names", "names")
+    removed_attributes <- "names"
   } else if (inherits(x, "factor")) {
     removed_attributes <- c("class", "levels")
   } else if (inherits(x, c("integer64", "Date", "arrow_binary", "arrow_large_binary"))) {


nealrichardson:
> Not sure why `row.names` was ever removed. [This is the commit](https://github.com/sirensolutions/arrow/commit/a6d531a3f409706d7dcde4562913c8589be3402a#diff-659e9fa6b66e5a72b4e3f9ac79ffddf08f92d9ea3d7aa45bd8c73b9a022fa2e5) that added the entire metadata.R, it seems a part of an unrelated effort. I believe the other attributes listed in `removed_attributes` were removed because they're serialized/deserialized by other means. `row.names` doesn't belong there.
> 
> * Closes: [[R] arrow::write_feather doesn't save row names #15246](https://github.com/apache/arrow/issues/15246)

row names are removed from metadata for a few reasons:

* They are length `n`, so they can easily get too large to store in schema metadata
* If rows are filtered in arrow prior to converting to an R data.frame, the row names won't be filtered (they're not in the table), and their length won't match the resulting data.

If you want to keep row names, they need to be added to the table. We could try to develop a convention for how to do that automatically in the R package, though the best solution is probably for you to be explicit about adding the column to the table yourself.

OfekShilon:
> row names are removed from metadata for a few reasons:
> 
> * They are length `n`, so they can easily get too large to store in schema metadata
> * If rows are filtered in arrow prior to converting to an R data.frame, the row names won't be filtered (they're not in the table), and their length won't match the resulting data.
> 
> If you want to keep row names, they need to be added to the table. We could try to develop a convention for how to do that automatically in the R package, though the best solution is probably for you to be explicit about adding the column to the table yourself.

Thanks for your prompt reply. 

As users we obviously expect the save/load roundtrip to be a no-op.  Not sure silently dropping row.names (or other attributes)  just because they were less convenient implementation-wise is the best design choice.

I can suggest quite a few ways to avoid this unpleasant surprise:
(1) Serializing row.names as regular column data (with some designated name) and deserializing them directly into an attribute, (best for the user)
(2) Adding a `save_row_names` parameter to write_feather, even defaulting to false - _but have it documented_,
(3) A warning like "You're trying to save a large object with row names. Consider putting them at a column for better performance and stability",   (probably best for the implementer)
etc. etc.
Hopefully something like this can be done.

OfekShilon:
After measuring the impact on large datasets - we see the performance impact, and opted to save row.names via temporary columns as you suggested.
I believe the same treatment should generally apply to all attributes of length nrows - notably data.table indices.

OfekShilon:
We eventually opted to go with your suggestion and are saving/loading `row.names` using a dedicated column. You might want to consider doing it as part of the package and perhaps more generally - for all attributes of length nrows (notably data.table indices).

thisisnic:
I'm going to close this PR for now as we're not looking to merge this as-is, but I have updated issue #15246 to reflect the fact that we could consider implementing support for row names in future.

PR #15266, opened 2023-01-09
ARROW-17715: [CI][C++] reduce the parallelism for cmake

kiszk:
Current setting causes timeout during C++ compilation.


kiszk:
@kou Sorry, I made a mistake...

PR #15268, opened 2023-01-09
GH-14976: [Python] Avoid dependency on exec plan in Table.sort_by to fix minimal tests

jorisvandenbossche:
* Closes: #14976


jorisvandenbossche:
@github-actions crossbow submit example-python-minimal-build-fedora-conda example-python-minimal-build-ubuntu-venv

jorisvandenbossche:
The alternative is to simply mark those tests with `@pytest.mark.dataset` as a proxy to indicate that exec plan is not available. 

But since this method worked previously with a minimal python install, we can easily keep it working.

kou (approved):
+1

Thanks!

PR #15270, opened 2023-01-09
ARROW-16480: [R] Update read_csv_arrow and open_dataset parse_options, read_options, and convert_options to take lists

thisisnic:


r/tests/testthat/test-csv.R
```diff
@@ -664,3 +664,27 @@ test_that("Shows an error message when trying to read a timestamp with time zone
     "CSV conversion error to timestamp\\[ns\\]: expected no zone offset in"
   )
 })
+
+test_that("CSV reading/parsing/convert options can be passed in as lists", {
+  tf <- tempfile()
+  on.exit(unlink(tf))
+
+  writeLines('"x"\nNA\nNA\n"NULL"\n\n"foo"\n', tf, )
+  readLines(tf)
```
wjones127, line 673: Small nit: we make a lot of temp files in these tests, when we could just do this in-memory. Don't worry about the rest of the file, but maybe simplify to:

```suggestion
  tf <- buffer(charToRaw('"x"\nNA\nNA\n"NULL"\n\n"foo"\n'))
```

wjones127: Although, is this the right output? I'm not sure the null values are being parsed correctly or `ignore_empty_lines` is working, but `skip_rows` does seem to be working.

```r
> library(arrow)

Attaching package: ‘arrow’

The following object is masked from ‘package:utils’:

    timestamp

> buf <- buffer(charToRaw('"x"\nNA\nNA\n"NULL"\n\n"foo"\n'))
> tab1 <- read_csv_arrow(
+     buf,
+     convert_options = list(null_values = c("NA", "NULL")),
+     parse_options = list(ignore_empty_lines = FALSE),
+     read_options = list(skip_rows = 1L)
+   )
> tab1
    NA
1   NA
2 NULL
3     
4  foo
```

thisisnic: What's the impact of creating temporary files?  We could looks to make it more efficient (having a single tempfile we overwrite each time perhaps?) if it's slowing things down, though I would prefer to have them keep using files given that this is the topic of these tests.

wjones127: Just slowing things down. The C++ tests generally write to buffers, not files, for single file tests. But if people would prefer to keep as files that's fine too. 

> though I would prefer to have them keep using files given that this is the topic of these tests.

If there is a test that is specifically about the interaction of CSV data and the filesystem, I agree that makes sense. For example, writing a dataset and verifying the directory structure on disk is expected. But most of these don't really care about the filesystem; writing CSV data to a buffer is just as valid as to a temp file.

thisisnic: OK, that makes sense, thanks for explaining.  If the difference is significant, we should definitely look into it. Any idea how much impact it has?

r/tests/testthat/test-dataset-csv.R
```diff
@@ -444,11 +444,35 @@ test_that("skip argument in open_dataset", {
...
     regexp = "`schema` must be an object of class 'Schema' not 'function'.",
     fixed = TRUE
   )
 })
+
+test_that("CSV reading/parsing/convert options can be passed in as lists", {
+  tf <- tempfile()
+  on.exit(unlink(tf))
+
+  writeLines('"x"\n"y"\nNA\nNA\n"NULL"\n\n"foo"\n', tf, )
```
wjones127, line 459: nit: does this trailing comma thing do anything?
```suggestion
  writeLines('"x"\n"y"\nNA\nNA\n"NULL"\n\n"foo"\n', tf)
```

PR #15273, opened 2023-01-09
MINOR: [C++] Mark TypeHolder operator bool() as const

felipecrv:
To be able to use it when the `TypeHolder` is const.


pitrou (approved):
+1, I will just wait for CI to make sure there are no relevant failures

PR #15278, opened 2023-01-09
GH-14474: Opportunistically delete R references to shared pointers where possible

### LINKED ISSUE #14474: How do I remove a fileSystemDataset object without calling garbage collection from R?
I would like to create a fileSystemDataset object in a temp folder, process it in batches, and then remove it. This works fine on Linux and Mac but on Windows a file lock prevents removal of the temp folder. I think the lock is created by arrow and is not released until manually call the garbage collector. I don't think I should (or am allowed to) call the garbage collector from inside a function that is part of a CRAN hosted R package. So how do I remove the file lock in Windows so I can delete the fileSystemDataset?

Reproducible example below.

``` r
library(arrow)

# create a FileSystemDataset object

filename <- here::here("tmp")
write_dataset(cars, filename, format = "feather")
ds <- open_dataset(filename, format = "feather")
ds
#> FileSystemDataset with 1 Feather file
#> speed: double
#> dist: double
#> 
#> See $metadata for additional Schema metadata

# process the file in batches
scanner <- ScannerBuilder$create(ds)$BatchSize(batch_size = 4)$Finish()
reader <- scanner$ToRecordBatchReader()

batch_num <- 1
while(!is.null(batch <- reader$read_next_batch())) {
  print(paste("Reading batch", batch_num, "with", nrow(batch), "rows"))
  batch_num <- batch_num + 1
}
#> [1] "Reading batch 1 with 4 rows"
#> [1] "Reading batch 2 with 4 rows"
#> [1] "Reading batch 3 with 4 rows"
#> [1] "Reading batch 4 with 4 rows"
#> [1] "Reading batch 5 with 4 rows"
#> [1] "Reading batch 6 with 4 rows"
#> [1] "Reading batch 7 with 4 rows"
#> [1] "Reading batch 8 with 4 rows"
#> [1] "Reading batch 9 with 4 rows"
#> [1] "Reading batch 10 with 4 rows"
#> [1] "Reading batch 11 with 4 rows"
#> [1] "Reading batch 12 with 4 rows"
#> [1] "Reading batch 13 with 2 rows"

rm(reader)
rm(scanner)
rm(ds)

# remove the file
rc <- unlink(filename, recursive = TRUE)
if(rc == 1) print("removal of file failed")
#> [1] "removal of file failed"

file.exists(filename)
#> [1] TRUE

# call gc()
gc()
#>           used (Mb) gc trigger  (Mb) max used (Mb)
#> Ncells 1115105 59.6    2401181 128.3  1234217 66.0
#> Vcells 1937727 14.8    8388608  64.0  3294370 25.2

# remove the file
rc <- unlink(filename, recursive = TRUE)
if(rc == 1) print("removal of file failed")

file.exists(filename)
#> [1] FALSE

```

<sup>Created on 2022-10-19 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> - Session info ---------------------------------------------------------------
#>  setting  value                       
#>  version  R version 4.0.5 (2021-03-31)
#>  os       Windows 10 x64              
#>  system   x86_64, mingw32             
#>  ui       RTerm                       
#>  language (EN)                        
#>  collate  English_United States.1252  
#>  ctype    English_United States.1252  
#>  tz       America/New_York            
#>  date     2022-10-19                  
#> 
#> - Packages -------------------------------------------------------------------
#>  package     * version date       lib source        
#>  arrow       * 9.0.0.2 2022-10-02 [1] CRAN (R 4.0.5)
#>  assertthat    0.2.1   2019-03-21 [1] CRAN (R 4.0.5)
#>  backports     1.4.0   2021-11-23 [1] CRAN (R 4.0.5)
#>  bit           4.0.4   2020-08-04 [1] CRAN (R 4.0.5)
#>  bit64         4.0.5   2020-08-30 [1] CRAN (R 4.0.5)
#>  cli           3.0.1   2021-07-17 [1] CRAN (R 4.0.5)
#>  crayon        1.5.1   2022-03-26 [1] CRAN (R 4.0.5)
#>  DBI           1.1.2   2021-12-20 [1] CRAN (R 4.0.5)
#>  digest        0.6.27  2020-10-24 [1] CRAN (R 4.0.5)
#>  dplyr         1.0.8   2022-02-08 [1] CRAN (R 4.0.5)
#>  ellipsis      0.3.2   2021-04-29 [1] CRAN (R 4.0.5)
#>  evaluate      0.14    2019-05-28 [1] CRAN (R 4.0.5)
#>  fansi         0.5.0   2021-05-25 [1] CRAN (R 4.0.5)
#>  fastmap       1.1.0   2021-01-25 [1] CRAN (R 4.0.5)
#>  fs            1.5.0   2020-07-31 [1] CRAN (R 4.0.5)
#>  generics      0.1.2   2022-01-31 [1] CRAN (R 4.0.5)
#>  glue          1.6.2   2022-02-24 [1] CRAN (R 4.0.5)
#>  here          1.0.1   2020-12-13 [1] CRAN (R 4.0.5)
#>  highr         0.9     2021-04-16 [1] CRAN (R 4.0.5)
#>  htmltools     0.5.2   2021-08-25 [1] CRAN (R 4.0.5)
#>  knitr         1.36    2021-09-29 [1] CRAN (R 4.0.5)
#>  lifecycle     1.0.1   2021-09-24 [1] CRAN (R 4.0.5)
#>  magrittr      2.0.1   2020-11-17 [1] CRAN (R 4.0.5)
#>  pillar        1.7.0   2022-02-01 [1] CRAN (R 4.0.5)
#>  pkgconfig     2.0.3   2019-09-22 [1] CRAN (R 4.0.5)
#>  purrr         0.3.4   2020-04-17 [1] CRAN (R 4.0.5)
#>  R6            2.5.1   2021-08-19 [1] CRAN (R 4.0.5)
#>  reprex        2.0.1   2021-08-05 [1] CRAN (R 4.0.5)
#>  rlang         1.0.2   2022-03-04 [1] CRAN (R 4.0.5)
#>  rmarkdown     2.10    2021-08-06 [1] CRAN (R 4.0.5)
#>  rprojroot     2.0.2   2020-11-15 [1] CRAN (R 4.0.5)
#>  rstudioapi    0.13    2020-11-12 [1] CRAN (R 4.0.5)
#>  sessioninfo   1.1.1   2018-11-05 [1] CRAN (R 4.0.5)
#>  stringi       1.7.5   2021-10-04 [1] CRAN (R 4.0.5)
#>  stringr       1.4.0   2019-02-10 [1] CRAN (R 4.0.5)
#>  styler        1.5.1   2021-07-13 [1] CRAN (R 4.0.5)
#>  tibble        3.1.2   2021-05-16 [1] CRAN (R 4.0.5)
#>  tidyselect    1.1.2   2022-02-21 [1] CRAN (R 4.0.5)
#>  tzdb          0.2.0   2021-10-27 [1] CRAN (R 4.0.5)
#>  utf8          1.2.1   2021-03-12 [1] CRAN (R 4.0.5)
#>  vctrs         0.3.8   2021-04-29 [1] CRAN (R 4.0.5)
#>  withr         2.5.0   2022-03-03 [1] CRAN (R 4.0.5)
#>  xfun          0.25    2021-08-06 [1] CRAN (R 4.0.5)
#>  yaml          2.2.1   2020-02-01 [1] CRAN (R 4.0.5)
#> 
#> [1] C:/Users/adam.DESKTOP-D3KQQA1/Documents/R/win-library/4.0
#> [2] C:/Program Files/R/R-4.0.5/library
```

</details>

paleolimbot:
(Still trying to set up a Windows environment so I can check to see if this helps at all)
* Closes: #14474

### DIFF
--- a/r/R/record-batch-reader.R
+++ b/r/R/record-batch-reader.R
@@ -100,7 +100,11 @@ RecordBatchReader <- R6Class("RecordBatchReader",
     read_table = function() Table__from_RecordBatchReader(self),
     Close = function() RecordBatchReader__Close(self),
     export_to_c = function(stream_ptr) ExportRecordBatchReader(self, stream_ptr),
-    ToString = function() self$schema$ToString()
+    ToString = function() self$schema$ToString(),
+    .unsafe_delete = function() {
+      RecordBatchReader__UnsafeDelete(self)
+      super$.unsafe_delete()
+    }
   ),
   active = list(
     schema = function() RecordBatchReader__schema(self)
@@ -255,6 +259,8 @@ as_record_batch_reader.arrow_dplyr_query <- function(x, ...) {
   # See query-engine.R for ExecPlan/Nodes
   plan <- ExecPlan$create()
   final_node <- plan$Build(x)
+  on.exit(plan$.unsafe_delete())
+
   plan$Run(final_node)
 }
 
--- a/r/R/table.R
+++ b/r/R/table.R
@@ -339,7 +339,10 @@ as_arrow_table.Dataset <- function(x, ...) {
 #' @rdname as_arrow_table
 #' @export
 as_arrow_table.arrow_dplyr_query <- function(x, ...) {
-  out <- as_arrow_table(as_record_batch_reader(x))
+  reader <- as_record_batch_reader(x)
+  on.exit(reader$.unsafe_delete())
+
+  out <- as_arrow_table(reader)
   # arrow_dplyr_query holds group_by information. Set it on the table metadata.
   set_group_attributes(
     out,
--- a/r/R/arrow-object.R
+++ b/r/R/arrow-object.R
@@ -52,7 +52,12 @@ ArrowObject <- R6Class("ArrowObject",
       # call the underlying shared pointer's reset() method for the
       # shared_ptr<SubclassType> in C++.
       self$`.:xp:.` <- NULL
-      invisible(self)
+
+      # Return NULL, because keeping this R6 object in scope is not a good idea.
+      # This syntax would allow the rare use that has to actually do this to
+      # do `object <- object$.unsafe_delete()` and reduce the chance that an
+      # IDE like RStudio will try try to call other methods which will error
+      invisible(NULL)
     }
   )
 )
--- a/r/R/arrow-object.R
+++ b/r/R/arrow-object.R
@@ -45,6 +45,14 @@ ArrowObject <- R6Class("ArrowObject",
         cat(self$ToString(), "\n", sep = "")
       }
       invisible(self)
+    },
+    .unsafe_delete = function() {
+      # The best we can do in a generic way is to set the underlying
+      # pointer to NULL. Subclasses specialize this so that we can actually
+      # call the underlying shared pointer's reset() method for the
+      # shared_ptr<SubclassType> in C++.
+      self$`.:xp:.` <- NULL
+      invisible(self)
     }
   )
 )
--- a/r/R/arrowExports.R
+++ b/r/R/arrowExports.R
@@ -448,6 +448,10 @@ ExecPlan_ToString <- function(plan) {
   .Call(`_arrow_ExecPlan_ToString`, plan)
 }
 
+ExecPlan_UnsafeDelete <- function(plan) {
+  invisible(.Call(`_arrow_ExecPlan_UnsafeDelete`, plan))
+}
+
 ExecNode_output_schema <- function(node) {
   .Call(`_arrow_ExecNode_output_schema`, node)
 }
@@ -1758,6 +1762,10 @@ RecordBatchReader__schema <- function(reader) {
 
 RecordBatchReader__Close <- function(reader) {
   invisible(.Call(`_arrow_RecordBatchReader__Close`, reader))
+}
+
+RecordBatchReader__UnsafeDelete <- function(reader) {
+  invisible(.Call(`_arrow_RecordBatchReader__UnsafeDelete`, reader))
 }
 
 RecordBatchReader__ReadNext <- function(reader) {
--- a/r/R/dataset-write.R
+++ b/r/R/dataset-write.R
@@ -151,6 +151,8 @@ write_dataset <- function(dataset,
   }
 
   plan <- ExecPlan$create()
+  on.exit(plan$.unsafe_delete())
+
   final_node <- plan$Build(dataset)
   if (!is.null(final_node$extras$sort %||% final_node$extras$head %||% final_node$extras$tail)) {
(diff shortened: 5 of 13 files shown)


r/R/record-batch-reader.R
```diff
@@ -255,6 +259,8 @@ as_record_batch_reader.arrow_dplyr_query <- function(x, ...) {
   # See query-engine.R for ExecPlan/Nodes
   plan <- ExecPlan$create()
   final_node <- plan$Build(x)
+  on.exit(plan$.unsafe_delete())
```
wjones127, line 262: Should we do the same everywhere else a plan is created? For example:

https://github.com/apache/arrow/blob/37a79659f82a7dbe0a55858e59fa5eaa380a266f/r/R/dataset-write.R#L153

https://github.com/apache/arrow/blob/37a79659f82a7dbe0a55858e59fa5eaa380a266f/r/R/query-engine.R#L381

r/R/table.R
```diff
@@ -339,7 +339,10 @@ as_arrow_table.Dataset <- function(x, ...) {
 #' @rdname as_arrow_table
 #' @export
 as_arrow_table.arrow_dplyr_query <- function(x, ...) {
-  out <- as_arrow_table(as_record_batch_reader(x))
+  reader <- as_record_batch_reader(x)
+  on.exit(reader$.unsafe_delete())
```
wjones127, line 343: Similarly, should we do this everywhere a record batch reader is created an consumed within a function? For example:

https://github.com/apache/arrow/blob/37a79659f82a7dbe0a55858e59fa5eaa380a266f/r/R/dplyr.R#L297

https://github.com/apache/arrow/blob/37a79659f82a7dbe0a55858e59fa5eaa380a266f/r/R/python.R#L111

wjones127 (commented):
Code seems sensible. There are a few places where I wonder if we want to add the delete logic to.

r/R/record-batch-reader.R
paleolimbot: Good catch! I added an `on.exit()` for both of these.

r/R/table.R
paleolimbot: Good catch! I added an `on.exit()` for the first one...for the second one the input is very specifically a `Table` which we have no mechanism for cleaning up (and probably don't need to?). If you think it's easier to wrap one's head around if the rule is "temporary record batch readers are always deleted with `on.exit()`" I'm game (otherwise I'm inclined to constrain this type of deleting to ExecPlan-related things).

wjones127 (approved):
This looks good, assuming CI jobs are successful.

PR #15288, opened 2023-01-10
GH-15284: [C++] Use DeclarationToExecBatches in Acero plan tests

### LINKED ISSUE #15284: [C++] Use DeclarationToExecBatches in Acero plan tests

The `DeclarationToExecBatches` provides a much neater way to enhance the existing test cases in `plan_test.cc`. 

### Component(s)

C++

vibhatha:
This PR includes a change to the existing test cases in `plan_test.cc` for Acero-based plan evaluation. 

At the moment, using the `StartAndCollect` method and manually passing the `FutureMatcher` and expected results are done and the validation is done on the results obtained from the `StartAndCollect` method. Most of the updated test cases in this PR uses the a common format. 

Replacing that format and using `DeclarationToExecBatches` makes the code more readable and easy to debug. 
### TODO
- [x] Convert other tests with sort options and custom sinks
- [x] Self review 1
- [x] Self review 2
- [x] Check CI Status
- [x] Mark Ready to Review

* Closes: #15284

### DIFF
--- a/cpp/src/arrow/compute/exec/plan_test.cc
+++ b/cpp/src/arrow/compute/exec/plan_test.cc
@@ -317,7 +317,8 @@ void TestSourceSink(
     };
     Declaration plan(source_factory_name,
                      OptionsType{exp_batches.schema, element_it_maker});
-    ASSERT_OK_AND_ASSIGN(auto result, DeclarationToExecBatches(plan, parallel));
+    ASSERT_OK_AND_ASSIGN(auto result,
+                         DeclarationToExecBatches(std::move(plan), parallel));
     AssertExecBatchesEqualIgnoringOrder(result.schema, result.batches,
                                         exp_batches.batches);
   }
@@ -780,7 +781,8 @@ TEST(ExecPlanExecution, StressSourceSink) {
           schema({field("a", int32()), field("b", boolean())}), num_batches);
       Declaration plan("source", SourceNodeOptions{random_data.schema,
                                                    random_data.gen(parallel, slow)});
-      ASSERT_OK_AND_ASSIGN(auto result, DeclarationToExecBatches(plan, parallel));
+      ASSERT_OK_AND_ASSIGN(auto result,
+                           DeclarationToExecBatches(std::move(plan), parallel));
       AssertExecBatchesEqualIgnoringOrder(result.schema, result.batches,
                                           random_data.batches);
     }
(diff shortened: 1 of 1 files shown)


cpp/src/arrow/compute/exec/plan_test.cc
```diff
@@ -322,26 +306,16 @@ void TestSourceSink(
...
-                                      {"sink", SinkNodeOptions{&sink_gen}},
-                                  })
-                .AddToPlan(plan.get()));
-
-  ASSERT_THAT(StartAndCollect(plan.get(), sink_gen),
-              Finishes(ResultWith(UnorderedElementsAreArray(exp_batches.batches))));
+  Declaration plan(source_factory_name,
+                   OptionsType{exp_batches.schema, element_it_maker});
+  ASSERT_OK_AND_ASSIGN(auto result,
+                       DeclarationToExecBatches(std::move(plan), /*use_threads=*/false));
```
westonpace, line 317: Any particular reason for `use_threads=false` here?

cpp/src/arrow/compute/exec/plan_test.cc
```diff
@@ -928,52 +894,34 @@ TEST(ExecPlanExecution, StressSourceSinkStopped) {
...
-  ASSERT_THAT(StartAndCollect(plan.get(), sink_gen),
-              Finishes(ResultWith(UnorderedElementsAreArray(
-                  {ExecBatchFromJSON({int32(), boolean()}, "[]"),
-                   ExecBatchFromJSON({int32(), boolean()}, "[[6, false]]")}))));
+  Declaration plan = Declaration::Sequence(
+      {{"source", SourceNodeOptions{basic_data.schema, basic_data.gen(/*parallel=*/false,
+                                                                      /*slow=*/false)}},
+       {"filter", FilterNodeOptions{equal(field_ref("i32"), literal(6))}}});
+  ASSERT_OK_AND_ASSIGN(auto result,
+                       DeclarationToExecBatches(std::move(plan), /*user_threads=*/false));
```
westonpace, line 902:
```suggestion
                       DeclarationToExecBatches(std::move(plan), /*use_threads=*/false));
```

cpp/src/arrow/compute/exec/plan_test.cc
```diff
@@ -1046,26 +994,21 @@ TEST(ExecPlanExecution, SourceMinMaxScalar) {
...
     auto input = MakeGroupableBatches(/*multiplicity=*/parallel ? 100 : 1);
     auto minmax_opts = std::make_shared<ScalarAggregateOptions>();
-    auto expected_result = ExecBatch::Make(
-        {ScalarFromJSON(struct_({field("min", int32()), field("max", int32())}),
-                        R"({"min": -8, "max": 12})")});
-
-    ASSERT_OK_AND_ASSIGN(auto plan, ExecPlan::Make());
-    AsyncGenerator<std::optional<ExecBatch>> sink_gen;
+    auto ty = struct_({field("min", int32()), field("max", int32())});
+    auto expected_table = TableFromJSON(schema({field("struct", ty)}), {R"([
```
westonpace, line 998:
```suggestion
    auto min_max_type = struct_({field("min", int32()), field("max", int32())});
    auto expected_table = TableFromJSON(schema({field("struct", min_max_type)}), {R"([
```

cpp/src/arrow/compute/exec/plan_test.cc
```diff
@@ -1046,26 +994,21 @@ TEST(ExecPlanExecution, SourceMinMaxScalar) {
...
-                Finishes(ResultWith(UnorderedElementsAreArray({*expected_result}))));
+    Declaration plan = Declaration::Sequence(
+        {{"source", SourceNodeOptions{input.schema, input.gen(parallel, /*slow=*/false)}},
+         {"aggregate",
+          AggregateNodeOptions{
+              /*aggregates=*/{{"min_max", std::move(minmax_opts), "i32", "min_max"}},
+              /*keys=*/{}}}});
+    ASSERT_OK_AND_ASSIGN(auto result_table,
+                         DeclarationToTable(std::move(plan), parallel));
+    AssertTablesEqual(*result_table, *expected_table);
```
westonpace, line 1011:
```suggestion
    // No need to ignore order since there is only 1 row
    AssertTablesEqual(*result_table, *expected_table);
```

cpp/src/arrow/compute/exec/plan_test.cc
```diff
@@ -1232,30 +1167,22 @@ TEST(ExecPlanExecution, SourceFilterProjectGroupedSumTopK) {
...
+  Declaration plan = Declaration::Sequence(
+      {{"source", SourceNodeOptions{basic_data.schema,
+                                    basic_data.gen(/*parallel=*/false, /*slow=*/false)}},
+       {"aggregate", AggregateNodeOptions{
+                         /*aggregates=*/{{"sum", nullptr, "i32", "sum(i32)"},
+                                         {"any", nullptr, "bool", "any(bool)"}},
+                     }}});
+  auto exp_batches = {
+      ExecBatchFromJSON({int64(), boolean()}, {ArgShape::SCALAR, ArgShape::SCALAR},
+                        "[[22, true]]"),
```
westonpace, line 1181:
```suggestion
                        "[[22, true]]")
```

westonpace (changes requested):
A few minor suggestions.  Thanks for the cleanup.

Looks like you preferred to set `use_threads=false` which I suppose more accurately reflected the old behavior.  Can we instead just not specify it (leave it parallel, the default) unless needed?

cpp/src/arrow/compute/exec/plan_test.cc
vibhatha: Not really, just left the test case as it hasn’t used parallel mode. Should we prefer testing both modes at all times? 

PS: I sort of did that, but again removed that thinking I am introducing which is not in the current test cases.

vibhatha: I am setting this to default just now... I will confirm one's it is tested.

vibhatha: Just for the curiosity I repeated this on the `use_threads=false` setting (which is the same state this was run before making this change, I guess).

The following was my observation

```bash
/home/asus/github/fork/arrow/cpp/src/arrow/compute/exec/plan_test.cc:887: Failure
Value of: sink_gen()
Expected: value value value is equal to ExecBatch
    # Rows: 4
    0: Array[-2147483380,808065407,-681488731,829937304]
    1: Array[true,false,false,false]

  Actual: 16-byte object <F0-5B 46-2E D1-55 00-00 90-75 47-2E D1-55 00-00>, whose value (ExecBatch
    # Rows: 4
    0: Array[-2147483418,113924762,1252144620,-193012790]
    1: Array[false,true,true,true]
) doesn't match, whose value ExecBatch
    # Rows: 4
    0: Array[-2147483418,113924762,1252144620,-193012790]
    1: Array[false,true,true,true]
 doesn't match
Google Test trace:
/home/asus/github/fork/arrow/cpp/src/arrow/compute/exec/plan_test.cc:866: single threaded
/home/asus/github/fork/arrow/cpp/src/arrow/compute/exec/plan_test.cc:863: unslowed


-----------------------------------------------------------------------------------------------------

/home/asus/github/fork/arrow/cpp/src/arrow/compute/exec/plan_test.cc:887: Failure
Value of: sink_gen()
Expected: value value value is equal to ExecBatch
    # Rows: 4
    0: Array[-2147483380,808065407,-681488731,829937304]
    1: Array[true,false,false,false]

  Actual: 16-byte object <00-66 F7-B1 AD-55 00-00 40-DE F7-B1 AD-55 00-00>, whose value (ExecBatch
    # Rows: 4
    0: Array[-2147483522,-417100609,768701078,-1848250513]
    1: Array[true,false,true,true]
) doesn't match, whose value ExecBatch
    # Rows: 4
    0: Array[-2147483522,-417100609,768701078,-1848250513]
    1: Array[true,false,true,true]
 doesn't match
Google Test trace:
/home/asus/github/fork/arrow/cpp/src/arrow/compute/exec/plan_test.cc:866: parallel
/home/asus/github/fork/arrow/cpp/src/arrow/compute/exec/plan_test.cc:863: unslowed


-----------------------------------------------------------------------------------------------------

/home/asus/github/fork/arrow/cpp/src/arrow/compute/exec/plan_test.cc:887: Failure
Value of: sink_gen()
Expected: value value value is equal to ExecBatch
    # Rows: 4
    0: Array[-2147483380,808065407,-681488731,829937304]
    1: Array[true,false,false,false]

  Actual: 16-byte object <B0-75 1A-B3 48-56 00-00 30-06 1D-B3 48-56 00-00>, whose value (ExecBatch
    # Rows: 4
    0: Array[-2147483522,-417100609,768701078,-1848250513]
    1: Array[true,false,true,true]
) doesn't match, whose value ExecBatch
    # Rows: 4
    0: Array[-2147483522,-417100609,768701078,-1848250513]
    1: Array[true,false,true,true]
 doesn't match
Google Test trace:
/home/asus/github/fork/arrow/cpp/src/arrow/compute/exec/plan_test.cc:866: parallel
/home/asus/github/fork/arrow/cpp/src/arrow/compute/exec/plan_test.cc:863: unslowed
```

For 160 repetitions (bash on 8 runs), I observed 3 failures. Could say 3/8 fails. Do you have a clue why this could be happening? The failing test is `ExecPlanExecution.StressSourceSinkStopped`. A single run includes all test cases. 

But when run just this case over 100 repetitions, it failed only 2 times. I suspect a resource allocation issue. Appreciate your thoughts.

westonpace: Ah, I have a fix for that test here: https://github.com/apache/arrow/pull/15253/files#diff-b4e59a9cedddaaa33c03f220c03ac812f9ebdfb06692c9f91c5f8f5988363b57R853 

However, I don't know if that PR will get finished in time for the release.  I'll open an independent PR with the fix.  The test does this:

 * Generate a bunch of random batches
 * Start the plan
 * Wait for one batch to arrive
 * Cancel/abort the plan
 * Verify the one batch that arrived was the first of the random batches

However, there is no guarantee, when run in parallel, that the first batch to complete will be the first batch generated.

Since the test doesn't fully consume the plan we can't use DeclarationToXyz here.  This means when run serially it is using "single-thread mode" which is still slightly parallel enough to cause random ordering.

In my fix I changed it so parallel=false would run synchronously (and have dependent order) and parallel=true would not require it to be the first batch generated.

vibhatha:
@westonpace does this PR needs further changes?

cpp/src/arrow/compute/exec/plan_test.cc
vibhatha: Shall we hold this until this PR is merged?

westonpace: No, sorry.  I thought you were modifying the test you were referring to but I see now you are not.  I think this PR is good to go.

PR #15291, opened 2023-01-10
GH-15290: [C++][Compute] Optimize IfElse kernel AAS/ASA case when the scalar is null

### LINKED ISSUE #15290: [C++][Compute] Optimize IfElse kernel AAS/ASA case when the scalar is null

In AAS and ASA case, when the scalar is null, we only need to construct a new validity bitmap and copy the input array data to output. We can skip looping over the input array.

### Component(s)

C++

js8544:

Closes #15290

# Rationale for this change

In AAS and ASA case, when the scalar is null, we only need to construct a new validity bitmap and copy the input array data to output. We can skip looping over the input array.

The above optimization for numeric and binary arrays. 

# Are these changes tested?

Tested with the original IfElse tests.

No.
* Closes: #15290

### DIFF
--- a/cpp/src/arrow/compute/kernels/scalar_if_else.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_if_else.cc
@@ -738,14 +738,14 @@ struct IfElseFunctor<Type, enable_if_base_binary<Type>> {
     const uint8_t* right_data = right.buffers[2].data;
 
     if (!left.is_valid) {  // left is null scalar, only need to copy right data to output
-      auto& out_data = *out->array_data();
+      auto* out_data = out->array_data().get();
       auto offset_length = (cond.length + 1) * sizeof(OffsetType);
-      ARROW_ASSIGN_OR_RAISE(out_data.buffers[1], ctx->Allocate(offset_length));
-      memcpy(out_data.buffers[1]->mutable_data(), right_offsets, offset_length);
+      ARROW_ASSIGN_OR_RAISE(out_data->buffers[1], ctx->Allocate(offset_length));
+      std::memcpy(out_data->buffers[1]->mutable_data(), right_offsets, offset_length);
 
       auto right_data_length = right_offsets[right.length] - right_offsets[0];
-      ARROW_ASSIGN_OR_RAISE(out_data.buffers[2], ctx->Allocate(right_data_length));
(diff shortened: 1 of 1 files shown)


cpp/src/arrow/compute/kernels/scalar_if_else.cc
```diff
@@ -723,12 +734,24 @@ struct IfElseFunctor<Type, enable_if_base_binary<Type>> {
...
   static Status Call(KernelContext* ctx, const ArraySpan& cond, const Scalar& left,
                      const ArraySpan& right, ExecResult* out) {
-    std::string_view left_data = internal::UnboxScalar<Type>::Unbox(left);
-    auto left_size = static_cast<OffsetType>(left_data.size());
-
     const auto* right_offsets = right.GetValues<OffsetType>(1);
     const uint8_t* right_data = right.buffers[2].data;
 
+    if (!left.is_valid) {  // left is null scalar, only need to copy right data to output
+      auto& out_data = *out->array_data();
```
lidavidm, line 741: nit: we mostly use pointers instead of non-const references

cpp/src/arrow/compute/kernels/scalar_if_else.cc
```diff
@@ -723,12 +734,24 @@ struct IfElseFunctor<Type, enable_if_base_binary<Type>> {
...
+      auto offset_length = (cond.length + 1) * sizeof(OffsetType);
+      ARROW_ASSIGN_OR_RAISE(out_data.buffers[1], ctx->Allocate(offset_length));
+      memcpy(out_data.buffers[1]->mutable_data(), right_offsets, offset_length);
```
lidavidm, line 744: nit: stick with std::memcpy for consistency?

lidavidm (approved):
LGTM, just a couple nits.

It looks like CheckWithDifferentShapes in scalar_if_else_test.cc should hit all these new branches. Have you confirmed?

cpp/src/arrow/compute/kernels/scalar_if_else.cc
js8544: done

cpp/src/arrow/compute/kernels/scalar_if_else.cc
js8544: done

PR #15293, opened 2023-01-10
GH-15292: [C++] Typeclass alias is missing in ExtensionArray

### LINKED ISSUE #15292: [C++] Typeclass alias is missing in ExtensionArray

All array classes have the `using TypeClass = xxx` type alias except for ExtensionArray.

With the alias, it will be easier to write templated inline visitors such as the following code block:

```Cpp
class ExampleVisitor {
 public:
  template <typename T>
  Status Visit(const T& arr) {
    if constexpr (arrow::is_number_type<typename T::TypeClass>::value) {
      ///
    } else {
      ///
    }
    return Status::OK();
  }
};
```

Current it fails to compile with message: `No type named 'TypeClass' in 'arrow::ExtensionArray'`.

### Component(s)

C++

js8544:

Closes #15292 

# Rationale for this change

All array classes have the `using TypeClass = xxx` type alias except for ExtensionArray.

With the alias, it will be easier to write templated inline visitors such as the following code block:

```Cpp
class ExampleVisitor {
 public:
  template <typename T>
  Status Visit(const T& arr) {
    if constexpr (arrow::is_number_type<typename T::TypeClass>::value) {
      ///
    } else {
      ///
    }
    return Status::OK();
  }
};
```

Current it fails to compile with message: `No type named 'TypeClass' in 'arrow::ExtensionArray'`.
* Closes: #15292

### DIFF
--- a/cpp/src/arrow/extension_type.h
+++ b/cpp/src/arrow/extension_type.h
@@ -103,6 +103,7 @@ class ARROW_EXPORT ExtensionType : public DataType {
 /// \brief Base array class for user-defined extension types
 class ARROW_EXPORT ExtensionArray : public Array {
  public:
+  using TypeClass = ExtensionType;
   /// \brief Construct an ExtensionArray from an ArrayData.
   ///
   /// The ArrayData must have the right ExtensionType.


pitrou (approved):
LGTM, thank you @js8544

PR #15294, opened 2023-01-10
MINOR: [C++] Optimize Gandiva log function

js8544:
Gandiva's `log` function computes log_of_base twice.


pitrou (approved):
Looks like an obvious improvement.

PR #17877, opened 2023-01-10
PARQUET-2225: [C++][Parquet] Allow reading dense with RecordReader

fatemehp:
Currently ReadRecords reads spaced by default. Some readers may need to read the values dense, and reading spaced is less efficient than reading dense. We need an option for reading dense.


cpp/src/parquet/column_reader.h
```diff
@@ -278,9 +278,13 @@ class PARQUET_EXPORT RecordReader {
...
   /// Note that for repeated fields, a record may have more than one value
-  /// and all of them are read.
+  /// and all of them are read. If read_dense_for_nullable is true, it will
+  /// not leave any space for null values. Otherwise, it will read spaced.
+  /// Readers must call Reset() before switching between reading dense and
+  /// spaced since reading dense will not update the valid_bits_.
   /// \return number of records read
-  virtual int64_t ReadRecords(int64_t num_records) = 0;
+  virtual int64_t ReadRecords(int64_t num_records,
+                              bool read_dense_for_nullable = false) = 0;
```
emkornfield, line 287: Thinking about this API, is there a reason you can think of for someone between to switch between Dense/Non-Dense?  Would it make sense pass this as an option to Make instead?

cpp/src/parquet/column_reader.cc
```diff
@@ -1762,7 +1763,7 @@ class TypedRecordReader : public TypedColumnReaderImpl<DType>,
   }
 
   // Return number of logical records read
-  int64_t ReadRecordData(int64_t num_records) {
+  int64_t ReadRecordData(int64_t num_records, bool read_dense_for_nullable) {
```
wgtmac, line 1766: If `read_dense_for_nullable` is set to true but the column has null values, it seems that we haven't checked it and the reader may fail with random error or read wrong values.

cpp/src/parquet/column_reader.cc
```diff
@@ -1787,21 +1788,34 @@ class TypedRecordReader : public TypedColumnReaderImpl<DType>,
...
+        validity_io.valid_bits = valid_bits_->mutable_data();
+        validity_io.valid_bits_offset = values_written_;
+
+        DefLevelsToBitmap(def_levels() + start_levels_position,
+                          levels_position_ - start_levels_position, leaf_info_,
+                          &validity_io);
+        values_to_read = validity_io.values_read - validity_io.null_count;
+        null_count = validity_io.null_count;
+        ARROW_DCHECK_GE(values_to_read, 0);
+        ReadValuesSpaced(validity_io.values_read, null_count);
```
wgtmac, line 1811: What about checking if `null_count` is zero and calling `ReadValuesDense` if that is true? Actually `ReadValuesSpaced()` internally calls `decoder->DecodeSpaced()` which goes to the fast path `decoder->Decode()` if `null_count` is ZERO.

cpp/src/parquet/column_reader.cc
```diff
@@ -1787,21 +1788,34 @@ class TypedRecordReader : public TypedColumnReaderImpl<DType>,
...
+        values_written_ += values_to_read;
+        // Any values with def_level < max_def_level is considered null.
+        null_count = levels_position_ - start_levels_position - values_to_read;
+      } else {
+        ValidityBitmapInputOutput validity_io;
+        validity_io.values_read_upper_bound = levels_position_ - start_levels_position;
+        validity_io.valid_bits = valid_bits_->mutable_data();
+        validity_io.valid_bits_offset = values_written_;
+
+        DefLevelsToBitmap(def_levels() + start_levels_position,
```
wgtmac, line 1805: IIUC, your major saving comes from skipping decoding def/ref levels if one is pretty sure the nullable column does not have any null.

cpp/src/parquet/column_reader.h
wgtmac: Can we derive the null_count info from ColumnChunkMetaData or PageHeader and automatically enable it? To be conservative, we cannot enable it if null_count stats if missing. This would be much safer and user-friendly.

fatemehp:
Adding more clarification here.

The change proposed here is about the vector of values that is returned. Currently, we first come up with the location of the null values and then make a vector that has an empty space for the null values (reading spaced). When reading dense, we do not make space for the null values.

For example:
def_levels: [0, 1]
values: [10]

Reading spaced: [null, 10]
Reading dense: [10]

The change here is meaningful for nullable columns. The savings come when we have null values. The issue is that 1) it is inefficient to come up with the exact space of the nulls and move the values around to make space for null values and 2) Some readers may want to indeed read dense and so they have to move the null values out again.

fatemehp:
> In that case, I assume the caller does not care about the null values and will not be able to sync values of different columns (because they may have different null slots), am I right?

Yes, that is correct.

fatemehp:
> Is it better to add a new function to support this case?

I am open to suggestions regarding the API. I am inclined to pass this as a parameter when making the reader as suggested by @emkornfield.

fatemehp:
Copying benchmark results here:

```
-------------------------------------------------------------
Benchmark                                Time         
-------------------------------------------------------------
Optional BatchSize:1000/ReadDense:1    1006171 ns      
Optional BatchSize:1000/ReadDense:0   13019233 ns   
```

fatemehp:
Copying the full benchmark result here for reference:
```
--------------------------------------------------------------------------------------------------------------------------
Benchmark                                                                Time             CPU   Iterations UserCounters...
--------------------------------------------------------------------------------------------------------------------------
ColumnReaderSkipInt32/Repetition:0/BatchSize:1000                   237521 ns       237769 ns         2951 bytes_per_second=20.0546G/s
ColumnReaderSkipInt32/Repetition:1/BatchSize:1000                  1617840 ns      1601169 ns          453 bytes_per_second=1.58552G/s
ColumnReaderSkipInt32/Repetition:2/BatchSize:1000                  2513241 ns      2499972 ns          279 bytes_per_second=1102.55M/s
ColumnReaderReadBatchInt32/Repetition:0/BatchSize:1000              249705 ns       249864 ns         2705 bytes_per_second=19.0839G/s
ColumnReaderReadBatchInt32/Repetition:1/BatchSize:1000             1604586 ns      1601909 ns          449 bytes_per_second=1.58478G/s
ColumnReaderReadBatchInt32/Repetition:2/BatchSize:1000             2597184 ns      2596807 ns          276 bytes_per_second=1061.44M/s
RecordReaderSkipRecords/Repetition:0/BatchSize:1000                 269624 ns       270945 ns         2547 bytes_per_second=17.5991G/s
RecordReaderSkipRecords/Repetition:1/BatchSize:1000                1627094 ns      1625718 ns          431 bytes_per_second=1.56157G/s
RecordReaderSkipRecords/Repetition:2/BatchSize:1000               12275821 ns     12246340 ns           57 bytes_per_second=225.075M/s
RecordReaderReadRecords/Repetition:0/BatchSize:1000/ReadDense:1     314982 ns       314958 ns         2163 bytes_per_second=15.1397G/s
RecordReaderReadRecords/Repetition:0/BatchSize:1000/ReadDense:0     314255 ns       313987 ns         2141 bytes_per_second=15.1865G/s
RecordReaderReadRecords/Repetition:1/BatchSize:1000/ReadDense:1    1122195 ns      1116475 ns          653 bytes_per_second=2.27383G/s
RecordReaderReadRecords/Repetition:1/BatchSize:1000/ReadDense:0    8818219 ns      8792272 ns           84 bytes_per_second=295.67M/s
RecordReaderReadRecords/Repetition:2/BatchSize:1000/ReadDense:1   11038168 ns     11003551 ns           66 bytes_per_second=250.496M/s
RecordReaderReadRecords/Repetition:2/BatchSize:1000/ReadDense:0   19544598 ns     19487114 ns           36 bytes_per_second=141.444M/s
```

pitrou:
> I am open to suggestions regarding the API. I am inclined to pass this as a parameter when making the reader as suggested by @emkornfield.

Given the RecordReader API where the output buffer is allocated internally, I think this makes sense indeed.

cpp/src/parquet/column_reader.h
```diff
@@ -311,13 +311,14 @@ class PARQUET_EXPORT RecordReader {
   static std::shared_ptr<RecordReader> Make(
```
emkornfield, line 311: please add parameter docs.

cpp/src/parquet/column_reader.h
```diff
@@ -367,6 +368,7 @@ class PARQUET_EXPORT RecordReader {
   uint8_t* values() const { return values_->mutable_data(); }
 
   /// \brief Number of values written including nulls (if any)
+  /// If read_dense_for_nullable_ is true this will not include nulls.
```
emkornfield, line 371: nit: please try to phrase things relative to publicly visible APIs.  E.g. if this Reader was constructed with read_dense_for_nullable set to true.  I think rephrasing the `\brief` a little bit could make sense since we essentially contradict (including nulls) if any on the very next line.

cpp/src/parquet/column_reader.cc
```diff
@@ -1829,21 +1831,34 @@ class TypedRecordReader : public TypedColumnReaderImpl<DType>,
...
+                          levels_position_ - start_levels_position, leaf_info_,
+                          &validity_io);
+        values_to_read = validity_io.values_read - validity_io.null_count;
+        null_count = validity_io.null_count;
+        ARROW_DCHECK_GE(values_to_read, 0);
```
emkornfield, line 1853: is ARROW_DCHECK_GE defined (looks like it was just DHECK_EQ beore.

cpp/src/parquet/column_reader_test.cc
```diff
@@ -757,6 +760,63 @@ TEST_F(RecordReaderTest, BasicReadRepeatedField) {
...
 
+// Tests reading the nullable values without saving a space for nulls.
+TEST_F(RecordReaderTest, ReadDenseForNullable) {
+  Init(/*max_def_level=*/1, /*max_rep_level=*/1, Repetition::REPEATED,
+       /*read_dense_for_nullable=*/true);
+
+  // Records look like: {[10], null, [20, 20], null, [30, 30, 30], null}
+  std::vector<std::shared_ptr<Page>> pages;
+  std::vector<int32_t> values = {10, 20, 20, 30, 30, 30};
+  std::vector<int16_t> def_levels = {1, 0, 1, 1, 0, 1, 1, 1, 0};
```
emkornfield, line 771: we should probably add a test for when 3 level encoding lists so we can check behavior for a list containing a null element and an empty list?

emkornfield (changes requested):
Mostly looks OK, my main concern is sufficient test coverage for more complicated repated fields (it would also be good to verify with no lists).

cpp/src/parquet/column_reader.h
fatemehp: Done

cpp/src/parquet/column_reader.h
fatemehp: Done.

cpp/src/parquet/column_reader.cc
fatemehp: Done.

cpp/src/parquet/column_reader.cc
```diff
@@ -1829,21 +1831,34 @@ class TypedRecordReader : public TypedColumnReaderImpl<DType>,
...
-      null_count = validity_io.null_count;
-      DCHECK_GE(values_to_read, 0);
-      ReadValuesSpaced(validity_io.values_read, null_count);
+      if (read_dense_for_nullable_) {
+        DCHECK_GE(values_to_read, 0);
+        ReadValuesDense(values_to_read);
+        // Total values, if reading dense, we don't have the nulls in the values
+        // vector.
+        values_written_ += values_to_read;
+        // Any values with def_level < max_def_level is considered null.
```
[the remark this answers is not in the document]
fatemehp: values_to_read here are really values, not records.

cpp/src/parquet/column_reader_test.cc
fatemehp: Done. Thanks for the suggestion, helped catch a bug reading Optional fields. I refactored the code a bit to make it less error-prone.

fatemehp:
@emkornfield could you take another look?

fatemehp:
Updating the benchmark numbers. The 10X improvement above is not accurate, it was due to a bug in the code for reading optional fields. For the benchmark data, the improvement for optional fields is more like 10%, and for repeated fields 40%.

Nevertheless, we need this API since some readers may want to read dense.

```
--------------------------------------------------------------------------------------------------------------------------
Benchmark                                                                Time             CPU   Iterations UserCounters...
--------------------------------------------------------------------------------------------------------------------------
ColumnReaderSkipInt32/Repetition:0/BatchSize:1000                   234597 ns       233231 ns         2529 bytes_per_second=20.4448G/s
ColumnReaderSkipInt32/Repetition:1/BatchSize:1000                  1613926 ns      1608305 ns          442 bytes_per_second=1.57848G/s
ColumnReaderSkipInt32/Repetition:2/BatchSize:1000                  2469379 ns      2462899 ns          280 bytes_per_second=1119.15M/s
ColumnReaderReadBatchInt32/Repetition:0/BatchSize:1000              257132 ns       256459 ns         2457 bytes_per_second=18.5931G/s
ColumnReaderReadBatchInt32/Repetition:1/BatchSize:1000             1629986 ns      1625472 ns          442 bytes_per_second=1.56181G/s
ColumnReaderReadBatchInt32/Repetition:2/BatchSize:1000             2530858 ns      2514407 ns          282 bytes_per_second=1096.22M/s
RecordReaderSkipRecords/Repetition:0/BatchSize:1000                 267614 ns       267870 ns         2459 bytes_per_second=17.801G/s
RecordReaderSkipRecords/Repetition:1/BatchSize:1000                1615133 ns      1613831 ns          443 bytes_per_second=1.57308G/s
RecordReaderSkipRecords/Repetition:2/BatchSize:1000               11080021 ns     11021551 ns           66 bytes_per_second=250.087M/s
RecordReaderReadRecords/Repetition:0/BatchSize:1000/ReadDense:1     312089 ns       312874 ns         2143 bytes_per_second=15.2406G/s
RecordReaderReadRecords/Repetition:0/BatchSize:1000/ReadDense:0     320200 ns       320023 ns         2128 bytes_per_second=14.9001G/s
RecordReaderReadRecords/Repetition:1/BatchSize:1000/ReadDense:1    7913189 ns      7885560 ns           89 bytes_per_second=329.667M/s
RecordReaderReadRecords/Repetition:1/BatchSize:1000/ReadDense:0    8907880 ns      8879646 ns           77 bytes_per_second=292.76M/s
RecordReaderReadRecords/Repetition:2/BatchSize:1000/ReadDense:1   11947824 ns     11942741 ns           54 bytes_per_second=230.797M/s
RecordReaderReadRecords/Repetition:2/BatchSize:1000/ReadDense:0   20702187 ns     20626693 ns           32 bytes_per_second=133.63M/s
```

emkornfield:
@fatemahp looks like build failures might be due to doc bugs:

```
/arrow/cpp/src/parquet/column_reader.h:306:6: error: '\param' command used in a comment that is not attached to a function declaration [-Werror,-Wdocumentation]
/// \param[in] column descriptor
```

cpp/src/parquet/column_reader.cc
```diff
@@ -1232,7 +1232,7 @@ int64_t TypedColumnReaderImpl<DType>::Skip(int64_t num_values_to_skip) {
       // Jump to the right offset in the Page
       int64_t values_read = 0;
       InitScratchForSkip();
-      ARROW_DCHECK_NE(this->scratch_for_skip_, nullptr);
```
emkornfield, line 1235: sorry for the churn, i guess if these macros already existed, then we should use the ARROW_ version to avoid conflicts.  I didn't realize they were both defined in the project.

cpp/src/parquet/column_reader.cc
```diff
@@ -1803,7 +1805,100 @@ class TypedRecordReader : public TypedColumnReaderImpl<DType>,
...
+        ++(*null_count);
+      }
+    }
+    // Any values with def_level < max_def_level is considered null.
+    DCHECK_EQ(*null_count, levels_position_ - start_levels_position - *values_to_read);
+    ReadValuesDense(*values_to_read);
+  }
+
+  // Reads spaced for optional or repeated fields.
+  void ReadSpacedForNullable(int64_t start_levels_position, int64_t* values_to_read,
```
emkornfield, line 1882: nit: even though it is verbose, I think being more descriptive here (e.g. ReadSpacedForNonRequired, or ReadSpacedForOptionalOrRepeated) might be clearer

cpp/src/parquet/column_reader.cc
```diff
@@ -1803,7 +1805,100 @@ class TypedRecordReader : public TypedColumnReaderImpl<DType>,
...
+    // read.
+    DCHECK_GE(levels_position_, start_levels_position);
+
+    // When reading dense we need to figure out number of values to read.
+    const int16_t* def_levels = this->def_levels();
+    for (int i = start_levels_position; i < levels_position_; ++i) {
+      if (def_levels[i] == this->max_def_level_) {
+        ++(*values_to_read);
+      } else {
+        ++(*null_count);
```
emkornfield, line 1873: could we document behavior here with respect to empty lists (i.e. why they are considered null).

emkornfield (changes requested):
Mostly nits and documentation requests, generally looks OK to me otherwise.

cpp/src/parquet/column_reader.cc
fatemehp: Done.

cpp/src/parquet/column_reader.cc
fatemehp: Done.

cpp/src/parquet/column_reader.cc
fatemehp: I added tests for reading spaced and empty lists, and it counts them as null. So the behavior is consistent between the two cases. I added a comment.

The write path pays specific attention to the repeated_ancestor_def_level (see MaybeCalculateValidityBits in column_writer). However, in the read path we seem to ignore it?

emkornfield: it could be we have a bug for non-3 level encoded fields there, we can probably handle that separately (maybe we should open up an issue to add specific test cases here).

emkornfield:
@fatemehp seems like CI failures might be relevant here?

cpp/src/parquet/column_reader.cc
fatemehp: I was handling this incorrectly in the code, and I fixed it. For repeated fields, we need to check to see if there is a nullable ancestor in the path or not. If there is then the repeated field is nullable otherwise it is required. I have updated the code accordingly.

Apart from determining if a repeated field is nullable or not, I don't think the handling of repeated fields is different from optional fields.

When reading dense, anything with a def_level < max_def_level will be counted in the null count (including empty lists). We do not leave space for nulls.

For reading spaced, we rely on internal::DefLevelsToBitmap and ReadValuesSpaced to take care of counting the nulls and leaving space for nulls. I have not changed this code path so it should work the same as before.

fatemehp: I changed the contract so that we do not populate null_count when reading dense. It turns out that null_count is the number of null values where we left space for it in the values vector. I clarified the null_count() contract. I found this through arrow internal tests. See [reconstruct_internal_test.cc testcase: ThreeLevelListOptionalOptional](https://github.com/apache/arrow/blob/45918a90a6ca1cf3fd67c256a7d6a240249e555a/cpp/src/parquet/arrow/reconstruct_internal_test.cc#L805). Here, the empty list was not counted towards the null count in the record reader.

fatemehp:
@emkornfield, @wjones127, I have fixed the issues, and all the tests appear to be passing. Could you take another look?

Note that I changed the contract so that when reading dense, the null_count is always 0. I have added more description here: https://github.com/apache/arrow/pull/17877#discussion_r1113542602

cpp/src/parquet/column_reader.h
```diff
@@ -364,10 +373,15 @@ class PARQUET_EXPORT RecordReader {
...
 
   /// \brief Decoded values, including nulls, if any
+  /// FLBA and ByteArray types do not use this array and read into their own
+  /// builders.
   uint8_t* values() const { return values_->mutable_data(); }
 
-  /// \brief Number of values written including nulls (if any)
-  /// There is no read-ahead/buffering for values.
+  /// \brief Number of values written, including space left for nulls if any.
+  /// If this Reader was constructed with read_dense_for_nullable(), there is no space for
```
wgtmac, line 381: I didn't see any function named `read_dense_for_nullable()`. Should we use `read_dense_for_nullable_` instead?

cpp/src/parquet/column_reader.cc
```diff
@@ -1803,7 +1805,104 @@ class TypedRecordReader : public TypedColumnReaderImpl<DType>,
...
+    const int64_t start_levels_position = levels_position_;
+    // Note that repeated records may be required or nullable. If they have
+    // an optional parent in the path, they will be nullable, otherwise,
+    // they are required. We use leaf_info_->HasNullableValues() that looks
+    // at repeated_ancestor_def_level to determine if it is required or
+    // nullable. Even if they are required, we may have to read ahead and
+    // delimit the records to get the right number of values and they will
+    // have associated levels.
+    int64_t records_read = DelimitRecords(num_records, values_to_read);
+    if (!nullable_values()) {
```
wgtmac, line 1821: When will this branch be hit?

cpp/src/parquet/column_reader.h
fatemehp: Done.

cpp/src/parquet/column_reader.cc
fatemehp: This will be hit for required repeated fields. Please see the comment 2 lines above this.

emkornfield:
Generally looks OK to me, I think merging your other PR caused a merge conflict, could you rebase.  Also given the sensitivity of the code I'd like to do another pass tomorrow with fresh eyes.

fatemehp:
I tried to merge, the conflicts from the other pull request were minor.

cpp/src/parquet/column_reader_benchmark.cc
```diff
@@ -56,19 +58,36 @@ class BenchmarkHelper {
...
+  RecordReader* ResetRecordReader(bool read_dense_for_nullable) {
+    std::unique_ptr<PageReader> pager;
+    pager.reset(new test::MockPageReader(pages_));
+    internal::LevelInfo level_info;
+    level_info.def_level = descr_->max_definition_level();
+    level_info.rep_level = descr_->max_repetition_level();
+    record_reader_ = internal::RecordReader::Make(
+        descr_.get(), level_info, ::arrow::default_memory_pool(),
+        /*read_dictionary=*/false,
+        /*read_dense_for_nullable=*/read_dense_for_nullable);
```
wgtmac, line 77: nit: the comment is redundant as the argument is self-descriptive

cpp/src/parquet/column_reader_benchmark.cc
```diff
@@ -123,35 +142,82 @@ static void ColumnReaderReadBatchInt32(::benchmark::State& state) {
...
+static void RecordReaderSkipRecords(::benchmark::State& state) {
+  const auto repetition = static_cast<Repetition::type>(state.range(0));
+  const auto batch_size = static_cast<int64_t>(state.range(1));
+
+  BenchmarkHelper helper(repetition, /*num_pages=*/16, /*levels_per_page=*/80000);
+
+  // Vectors to read the values into.
+  for (auto _ : state) {
+    state.PauseTiming();
+    RecordReader* reader = helper.ResetRecordReader(/*read_dense_for_nullable=*/true);
```
wgtmac, line 183: Does the value of `read_dense_for_nullable` really matters here? Or should we set it to false?

cpp/src/parquet/column_reader_benchmark.cc
```diff
@@ -123,35 +142,82 @@ static void ColumnReaderReadBatchInt32(::benchmark::State& state) {
...
     ->Args({1, 1000})
-    ->Args({1, 10000})
-    ->Args({1, 100000})
-    ->Args({2, 100})
-    ->Args({2, 1000})
-    ->Args({2, 10000})
-    ->Args({2, 100000});
+    ->Args({2, 1000});
+
+BENCHMARK(RecordReaderSkipRecords)
```
wgtmac, line 207: Why not keep different batch sizes as before?

cpp/src/parquet/column_reader_test.cc
```diff
@@ -644,20 +644,27 @@ TEST_F(TestPrimitiveReader, TestNonDictionaryEncodedPagesWithExposeEncoding) {
   pages_.clear();
 }
 
-class RecordReaderTest : public ::testing::Test {
+class RecordReaderPrimitiveTypeTest : public ::testing::TestWithParam<bool> {
```
wgtmac, line 647:
```suggestion
using ReadDenseForNullable = bool;
class RecordReaderPrimitiveTypeTest : public ::testing::TestWithParam<ReadDenseForNullable> {
```

wgtmac: This may provide better readability. Otherwise one has to go all the way to the end to understand what does it mean.

cpp/src/parquet/column_reader_test.cc
```diff
@@ -703,9 +710,187 @@ class RecordReaderTest : public ::testing::Test {
...
+  CheckReadValues(/*expected_values=*/{20, 20, 30, 30, 30},
+                  /*expected_defs=*/{1, 1, 1, 1, 1},
+                  /*expected_reps=*/{0, 1, 0, 1, 1});
+  record_reader_->Reset();
+  CheckState(/*values_written=*/0, /*null_count=*/0, /*levels_written=*/0,
+             /*levels_position=*/0);
+}
+
+// Tests reading a nullable repeated field.
+TEST_P(RecordReaderPrimitiveTypeTest, ReadNullableRepeated) {
```
wgtmac, line 889: Could you please add a test case for `empty list` and `list of null`?
For example:
- {[10], [] , [20, 20], [], [30, 30, 30]}
- {[10], [null], [20, 20], [30, null, 30], [null, null]}

So we can make sure cases for `max_def_level > 1` are well-covered.

cpp/src/parquet/column_reader_test.cc
fatemehp: Done

cpp/src/parquet/column_reader_test.cc
fatemehp: I modified ReadNullableRepeated to cover these cases.

cpp/src/parquet/column_reader_benchmark.cc
fatemehp: Just to keep it less cluttered. The previous benchmark showed a batch size of ~1000 has reasonable performance, so using that moving forward. One could always benchmark with different batch sizes.

cpp/src/parquet/column_reader_benchmark.cc
fatemehp: Set to false.

cpp/src/parquet/column_reader_benchmark.cc
fatemehp: Done.

fatemehp:
@wgtmac I have tried to address your comments. Please take a look. Thanks!

wgtmac (commented):
Looking good from my side. Thanks @fatemehp !

Let's wait for @emkornfield to take another pass.

fatemehp:
@emkornfield could you take another look and merge this?

PR #19706, opened 2023-01-10
GH-18818: [R] Create a field ref to a field in a struct

### LINKED ISSUE #18818: [R] Create a field ref to a field in a struct
See also ARROW-11259. This probably needs to be a `$` and `[[` method on `Expression`

**Reporter**: [Neal Richardson](https://issues.apache.org/jira/browse/ARROW-13858) / @nealrichardson
#### Related issues:
- [[C++] Support nested FieldRefs](https://github.com/apache/arrow/issues/29587) (is blocked by)

<sub>**Note**: *This issue was originally created as [ARROW-13858](https://issues.apache.org/jira/browse/ARROW-13858). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

nealrichardson:
This PR implements `$.Expression` and `[[.Expression` methods, such that if the Expression is a FieldRef, it returns a nested FieldRef. This required revising some assumptions in a few places, particularly that if an Expression is a FieldRef, it has a `name`, and that all FieldRefs correspond to a Field in a Schema. In the case where the Expression is not a FieldRef, it will create an Expression call to `struct_field` to extract the field, iff the Expression has a knowable `type`, the type is `StructType`, and the field name exists in the struct. 

Things not done because they weren't needed to get this working:

  * `Expression$field_ref()` take a vector to construct a nested ref
  * Method to return vector of nested components of a field ref in R

Next steps for future PRs:

* Wrap this in [tidyr::unpack()](https://tidyr.tidyverse.org/reference/pack.html) method (but unfortunately, unpack() is not a generic)
* https://github.com/apache/arrow/issues/33756
* https://github.com/apache/arrow/issues/33757
* https://github.com/apache/arrow/issues/33760

* Closes: #18818


paleolimbot (commented):
Happy to review more as this progresses...just a few observations:

- I think `ls(x)` is not a robust way to check for collisions. You may have to use `get()` or `mget()` with `inherits = TRUE` since the symbol might exist in a parent environment?
- Can `Exression$field_ref()` be updated to to accept a vector? Then you could avoid the C++ `compute___expr__nested_field_ref()` and do it all in R.?
- There should probably be a way to get all the components of a field ref (maybe `compute___expr__get_field_ref_name()` should return `cpp11::strings()`?

nealrichardson:
> Happy to review more as this progresses...just a few observations:
> 
> * I think `ls(x)` is not a robust way to check for collisions. You may have to use `get()` or `mget()` with `inherits = TRUE` since the symbol might exist in a parent environment?

For R6 this works because I'm just looking for a method name, and inheritance isn't handled via nested environments it seems (i.e. `class_name` is defined in the base `ArrowObject` class but it shows up in `ls()` for `Expression`). 

We've been using this trick for a while for Tables and RecordBatches: https://github.com/apache/arrow/blob/master/r/R/arrow-tabular.R#L153-L160

> * Can `Exression$field_ref()` be updated to to accept a vector? Then you could avoid the C++ `compute___expr__nested_field_ref()` and do it all in R.?

It could, and that was my first instinct too, but that's actually not helpful for the `$` method, where you have a FieldRef and the name of the thing you want to nest further. You'd still have to call out to C++ to get the vector of field names in the path of the existing FieldRef. 

> * There should probably be a way to get all the components of a field ref (maybe `compute___expr__get_field_ref_name()` should return `cpp11::strings()`?

Probably. I'm going to focus on getting the essential dplyr stuff working first and we'll see what's required for that. 

One further complication that is completely punted here is that FieldRefs can be from strings or integers, so a nested path could contain a mix of integers and strings: https://github.com/apache/arrow/blob/master/cpp/src/arrow/type.h#L1651-L1660 So just switching to `std::vector<std::string>` everywhere isn't a simple solution.

r/tests/testthat/test-dplyr-query.R
```diff
@@ -714,3 +714,48 @@ test_that("Scalars in expressions match the type of the field, if possible", {
...
+        nested = df_col$a,
+        times2 = df_col$a * 2
+      ) %>%
+      filter(nested > 7) %>%
+      summarize(sum(times2)) %>%
+      collect(),
+    nested_data
+  )
+
+  # Now with Dataset
```
paleolimbot, line 744: Is there a difference in behaviour that would be expected with Dataset? (If so, maybe add a note to that comment and if not, maybe remove it?)

r/src/expression.cpp
```diff
@@ -19,7 +19,7 @@
 
 #include <arrow/compute/api_scalar.h>
 #include <arrow/compute/exec/expression.h>
-
+#include <iostream>
```
paleolimbot: Is this addition intentional?

r/tests/testthat/test-expression.R
```diff
@@ -76,6 +76,16 @@ test_that("Field reference expression schemas and types", {
   expect_equal(x$type(), int32())
 })
 
+test_that("Nested field refs", {
+  x <- Expression$field_ref("x")
+  nested <- x$y
+  expect_r6_class(nested, "Expression")
+  expect_r6_class(x[["y"]], "Expression")
+  expect_r6_class(nested$z, "Expression")
+  # Should this instead be NULL?
```
paleolimbot, line 85: An error seems safer for now?

r/src/expression.cpp
```diff
@@ -79,6 +110,7 @@ std::shared_ptr<compute::Expression> compute___expr__scalar(
 
 // [[arrow::export]]
 std::string compute___expr__ToString(const std::shared_ptr<compute::Expression>& x) {
+  // TODO: something different if is field ref and IsNested?
   return x->ToString();
```
paleolimbot, line 114: Is the string representation awful for a nested reference? (Or do we use it to implicitly do stuff?) From the lack of failing tests here, it seems like we don't do a lot of testing of expression objects for whether or not they are field references.

paleolimbot (approved):
Just a few bits!

It looks like integers nestings aren't supported yet...that seems fine for the purposes of arrow--dplyr wrapping.

Was `devtools::document()` run after the S3 methods were added? (I didn't see a NAMESPACE file update but maybe GitHub is hiding it from me)

r/src/expression.cpp
nealrichardson: Nope, removed.

r/tests/testthat/test-dplyr-query.R
nealrichardson: There is a different code path used to create a ScanNode for a Dataset, and it identifies all fields used in the projection so they can be pushed down into the scan: https://github.com/apache/arrow/blob/master/r/R/query-engine.R#L37-L41

This was segfaulting (nested FieldRef doesn't have `.name()`) until I modified that function. I'll add more to this comment.

r/src/expression.cpp
nealrichardson: It's not awful, it's just not pretty. `FieldRef.Nested(FieldRef.Name(df_col) FieldRef.Name(a))`, whereas for a non-nested FieldRef, you just get `col`, not `FieldRef.Name(col)`. For R, I might want to print that as `df_col$a`, or `df_col[["a"]]` if we want to avoid messing with whether `a` needs backticks. 

Trouble is, even if we improve this for where we call ToString from R, it doesn't change how it works when it is called recursively inside an Expression's ToString method. So you still would get e.g. `multiply_checked(FieldRef.Nested(FieldRef.Name(df_col) FieldRef.Name(a)), 2)`, and there's nothing we could do about that. 

I'll just remove this comment.

r/tests/testthat/test-dplyr-query.R
```diff
@@ -759,3 +759,27 @@ test_that("Can use nested field refs", {
...
+  )
+})
+
+test_that("nested field ref error handling", {
+  expect_error(
+    example_data %>%
+      arrow_table() %>%
+      mutate(x = int$nested) %>%
+      compute()
+  )
```
paleolimbot, line 784: What is the expected error? (Or did you mean `expect_snapshot_error()`?)

paleolimbot (approved):
Just one `expect_error()` without an expected error to fix and this is good to go! Thank you!

r/src/expression.cpp
```diff
@@ -46,13 +46,26 @@ std::shared_ptr<compute::Expression> compute___expr__call(std::string func_name,
...
 std::vector<std::string> field_names_in_expression(
     const std::shared_ptr<compute::Expression>& x) {
   std::vector<std::string> out;
+  std::vector<arrow::FieldRef> nested;
+
   auto field_refs = FieldsInExpression(*x);
   for (auto f : field_refs) {
-    out.push_back(*f.name());
+    if (f.IsNested()) {
+      // We keep the top-level field name.
```
jorisvandenbossche, line 63: This might not be used in practice (in a `mutate` call where you select the field, you also directly specify the resulting column name), but otherwise it might also make sense to keep the innermost field name?

r/tests/testthat/test-dplyr-query.R
nealrichardson: It's the usual "No match for `field_ref->ToString()` in `schema->ToString()`". I wanted to check (which I did visually) that it was sensible, which it is, and then the test basically asserts that it doesn't try to compute something wrong and doesn't segfault. Since the error comes from C++ I thought it best not to assert something specific about it.

paleolimbot: It can be a minimal regex; however, I think it's important since `expect_error()` will swallow errors from completely invalid code (e.g., `expect_error(this_is_something(that_will)$never_run))` gives silence).

r/R/expression.R
```diff
@@ -89,6 +92,56 @@ Expression$create <- function(function_name,
...
+
+get_nested_field <- function(expr, name) {
+  if (expr$is_field_ref()) {
+    # Make a nested field ref
+    out <- compute___expr__nested_field_ref(expr, name)
+  } else {
+    # Use the struct_field kernel, but that only works if:
+    # * expr has a knowable type (has a schema set)
+    # * that type is struct
+    # * `name` exists in the struct (bc we have to map to an integer position)
```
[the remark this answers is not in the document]
jorisvandenbossche: Opened https://github.com/apache/arrow/issues/33745

r/src/expression.cpp
jorisvandenbossche: You can also specify field refs (well, generic expressions), but then you also need to pass the resulting name for the schema. See the second Project signature at 

https://github.com/apache/arrow/blob/4e439f6a597180c5fc8ff1552c860cecd33736c5/cpp/src/arrow/dataset/scanner.h#L463-L484

which gets translated to ScanOptions.projection. It seems that is also what the R bindings actually do inside `ExecNode_Scan` (it will convert the materialized_field_names back to FieldRefs). Now, the scanner itself will also just use the top-level name of a nested field ref to do pruning of what it needs to read, so right now preserving the nested field ref is not useful. But ideally in the future we would optimize that for formats that can do that (like parquet, cfr https://github.com/apache/arrow/issues/33167)

nealrichardson: Thanks for the pointers. I've deferred cleaning this up to #33760 since I see a few places where it could be more involved than just deleting code.

PR #24372, opened 2023-01-10
GH-14863: [C++] Add appender functions to array builders that can take optionals

### LINKED ISSUE #14863: [C++] Add `Append(std::optional<T>...)` convenience methods to builders

When you have a `std::optional` of the right value type, it would be convenient to append it directly to a concrete `ArrayBuilder` subclass instead of having to query whether it has a value:
```c++
template <typename V>
Status Append(const std::optional<V>& value) {
  return (value) ? Append(*value) : AppendNull();
}

template <typename V>
void UnsafeAppend(const std::optional<V>& value) {
  if (value) {
    UnsafeAppend(*value);
  } else {
    UnsafeAppendNull();
  }
}
```   

### Component(s)

C++

felipecrv:
This adds more member functions (different names) to some concrete implementations of `ArrayBuilder`.

* Closes: #14863

### DIFF
--- a/cpp/src/arrow/array/builder_base.h
+++ b/cpp/src/arrow/array/builder_base.h
@@ -41,11 +41,15 @@ namespace internal {
 template <class Builder, class V>
 class ArrayBuilderExtraOps {
  public:
+  /// \brief Append a value from an optional or null if it has no value.
   Status AppendOrNull(const std::optional<V>& value) {
     auto* self = static_cast<Builder*>(this);
     return value.has_value() ? self->Append(*value) : self->AppendNull();
   }
 
+  /// \brief Append a value from an optional or null if it has no value.
+  ///
+  /// Unsafe methods don't check existing size.
   void UnsafeAppendOrNull(const std::optional<V>& value) {
     auto* self = static_cast<Builder*>(this);
     return value.has_value() ? self->UnsafeAppend(*value) : self->UnsafeAppendNull();
--- a/cpp/src/arrow/array/builder_base.h
+++ b/cpp/src/arrow/array/builder_base.h
@@ -36,6 +36,24 @@
 
 namespace arrow {
 
+namespace internal {
+
+template <class Builder, class V>
+class ArrayBuilderExtraOps {
+ public:
+  Status AppendOrNull(const std::optional<V>& value) {
+    auto* self = static_cast<Builder*>(this);
+    return value.has_value() ? self->Append(*value) : self->AppendNull();
+  }
+
+  void UnsafeAppendOrNull(const std::optional<V>& value) {
+    auto* self = static_cast<Builder*>(this);
+    return value.has_value() ? self->UnsafeAppend(*value) : self->UnsafeAppendNull();
+  }
+};
+
+}  // namespace internal
+
 /// \defgroup numeric-builders Concrete builder subclasses for numeric types
 /// @{
 /// @}
--- a/cpp/src/arrow/testing/matchers.h
+++ b/cpp/src/arrow/testing/matchers.h
@@ -431,9 +431,9 @@ DataEqMatcher DataEqArray(T type, const std::vector<std::optional<ValueType>>& v
 
   for (auto value : values) {
     if (need_safe_append) {
-      builder.UnsafeAppendOrNull(value);
-    } else {
       DCHECK_OK(builder.AppendOrNull(value));
+    } else {
+      builder.UnsafeAppendOrNull(value);
     }
   }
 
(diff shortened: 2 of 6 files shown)


felipecrv:
@pitrou

cpp/src/arrow/array/builder_base.h
```diff
@@ -36,6 +36,48 @@
...
+  }
+
+ public:
+  Status AppendOrNull(const std::optional<V>& value) {
+    return AppendOptional(value, false);
+  }
+
+  Status AppendOrEmptyValue(const std::optional<V>& value) {
+    return AppendOptional(value, true);
+  }
```
pitrou, line 68: I don't think appending an empty value for an empty optional is an interesting use case, what do you think @bkietz ?

cpp/src/arrow/testing/matchers.h
```diff
@@ -430,14 +430,10 @@ DataEqMatcher DataEqArray(T type, const std::vector<std::optional<ValueType>>& v
...
 
   for (auto value : values) {
-    if (value) {
-      if (need_safe_append) {
-        builder.UnsafeAppend(*value);
-      } else {
-        DCHECK_OK(builder.Append(*value));
-      }
+    if (need_safe_append) {
+      builder.UnsafeAppendOrNull(value);
```
pitrou, line 434: It's interesting that the `need_safe_append` use seems flipped. It probably should be:
```c++
    if (need_safe_append) {
      DCHECK_OK(builder.AppendOrNull(value));
    } else {
      builder.UnsafeAppendOrNull(value);
    }
```


@bkietz What do you think?

felipecrv: I noticed that too. And decided to preserve the confusion. It would be great to fix it though by removing the `!` and then flipping the cases.

felipecrv: (but that is assuming the current code is not buggy)

pitrou: I think it is.

cpp/src/arrow/array/builder_base.h
felipecrv: Removed this and pushed again.

cpp/src/arrow/testing/matchers.h
felipecrv: Swapped.

felipecrv:
@pitrou @lidavidm I've found an use-case for this, so I fixed comments and would like to merge it.

cpp/src/arrow/array/builder_base.h
```diff
@@ -145,6 +164,11 @@ class ARROW_EXPORT ArrayBuilder {
   virtual Status AppendScalar(const Scalar& scalar, int64_t n_repeats);
   virtual Status AppendScalars(const ScalarVector& scalars);
 
+  /// \brief Append a scalar value from a optional or null if it has no value.
```
pitrou, line 167: "from a optional" seems incorrect here?

cpp/src/arrow/array/builder_base.h
```diff
@@ -36,6 +37,24 @@
 
 namespace arrow {
 
+namespace internal {
+
+template <class Builder, class V>
+class ArrayBuilderExtraOps {
+ public:
+  Status AppendOrNull(const std::optional<V>& value) {
```
pitrou, line 45: Can you add docstrings for these two methods?

pitrou (commented):
LGTM, just two comments

cpp/src/arrow/array/builder_base.h
felipecrv: Done.

cpp/src/arrow/array/builder_base.h
```diff
@@ -145,6 +168,11 @@ class ARROW_EXPORT ArrayBuilder {
   virtual Status AppendScalar(const Scalar& scalar, int64_t n_repeats);
   virtual Status AppendScalars(const ScalarVector& scalars);
 
+  /// \brief Append a scalar value from an optional or null if it has no value.
```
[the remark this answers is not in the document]
felipecrv: Oh right. The comment on this one shouldn't be like the others.

> Also, TBH, I'm not sure I understand the point of this API. Are there situations where this would be useful?

I'm planning to use it in the REE array builder. Representing the current open run value with a `shared_ptr<Scalar>` and length.

felipecrv, line 171:
```suggestion
  /// \brief Append a scalar value from the shared pointer or a null scalar if the pointer is null.
```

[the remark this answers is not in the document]
pitrou: Going through a `shared_ptr<Scalar>` would be too costly IMHO.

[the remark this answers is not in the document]
felipecrv: Ok, I will keep this in the builder code then.

pitrou (approved):
+1, thanks for this @felipecrv !

PR #29185, opened 2023-01-10
GH-15264: [C++] Add scanner tests for disabling readahead and fix relevant bugs

### LINKED ISSUE #15264: [C++] An error occurred when I set batch_readahead = 0 for parquet ScanOptions

In arrow cpp version 8.0.0, the error message is "Could not read from readahead_queue". And the details are as follows:

Firstly, I set use_threads to false, batch_readahead to 0, and set batch_size, dataset_schema, project, and filter in setScanOptions();.
```
    // project & filter
    setScanOptions();
```
Secondly, I use ParquetFileFormat::ScanBatchesAsync function to scan a ParquetFileFragment.

Then I follow the code below:
```
    // parquetFileFormat is an object of ParquetFileFormat class
    // parquetFileFragment is an object of FileFragment class
    auto recordBatchIterator = arrow::MakeGeneratorIterator(
                                std::move(parquetFileFormat->ScanBatchesAsync(parquetScanOptions, parquetFileFragment).ValueOrDie()));
    std::shared_ptr<arrow::RecordBatch> recordBatch;
    auto recordBatch_res = recordBatchIterator.Next();
    std::shared_ptr<arrow::RecordBatch> recordBatch;
    if (recordBatch_res.ok())
        recordBatch= recordBatch_res.ValueOrDie();
```

Finally, during debugging, the error message in recordBatch_res is "Could not read from readahead_queue".
I found this error message in the SerialReadaheadGenerator class, probably because of the data structure util::SpscQueue<std::shared_ptr<Future<T>>> readahead_queue_.
```
template <typename T>
using SpscQueue = arrow_vendored::folly::ProducerConsumerQueue<T>;
```
**The size of the corresponding ProducerConsumerQueue class must be greater than or equal to 2, which is described in the code comments. Therefore, if I set batch_readahead to 0, the size of ProducerConsumerQueue is 1, and an error occurs when I read the parquet file.**

> src\arrow\util\async_generator.h  583
> src\arrow\vendored\ProducerConsumerQueue.h  75

**My requirement is not to use multithreading, but when I set use_threads to false, if batch_readahead * batch_size > the number of rows in a row_group which to be read, it will read multiple row_groups at the same time in the parquetFileFragment. This means that when I turn use_threads off, multithreaded reads still occur in my code. My existing code doesn't support multithreading, so it reads errors.**

**Then I want to set batch_readahead to 0, and the above error occurs, which may be a bug. So, could you please fix this bug or help me to avoid reading multiple row_groups at the same time?**

### Component(s)

C++

westonpace:
* Closes: #15264

### DIFF
--- a/cpp/src/arrow/dataset/scanner.cc
+++ b/cpp/src/arrow/dataset/scanner.cc
@@ -988,16 +988,25 @@ Result<compute::ExecNode*> MakeScanNode(compute::ExecPlan* plan,
 
   AsyncGenerator<EnumeratedRecordBatch> merged_batch_gen;
   if (require_sequenced_output) {
-    ARROW_ASSIGN_OR_RAISE(merged_batch_gen,
-                          MakeSequencedMergedGenerator(std::move(batch_gen_gen),
-                                                       scan_options->fragment_readahead));
+    if (scan_options->fragment_readahead > 1) {
+      ARROW_ASSIGN_OR_RAISE(merged_batch_gen, MakeSequencedMergedGenerator(
+                                                  std::move(batch_gen_gen),
+                                                  scan_options->fragment_readahead));
+    } else {
+      merged_batch_gen = MakeConcatenatedGenerator(std::move(batch_gen_gen));
+    }
   } else {
     merged_batch_gen =
         MakeMergedGenerator(std::move(batch_gen_gen), scan_options->fragment_readahead);
   }
 
-  auto batch_gen = MakeReadaheadGenerator(std::move(merged_batch_gen),
-                                          scan_options->fragment_readahead);
+  AsyncGenerator<EnumeratedRecordBatch> batch_gen;
+  if (scan_options->fragment_readahead > 1) {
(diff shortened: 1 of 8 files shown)


westonpace:
Failures appear to be unrelated

cpp/src/arrow/dataset/scanner.cc
```diff
@@ -988,16 +988,25 @@ Result<compute::ExecNode*> MakeScanNode(compute::ExecPlan* plan,
...
   }
 
-  auto batch_gen = MakeReadaheadGenerator(std::move(merged_batch_gen),
-                                          scan_options->fragment_readahead);
+  AsyncGenerator<EnumeratedRecordBatch> batch_gen;
+  if (scan_options->fragment_readahead > 1) {
+    batch_gen = MakeReadaheadGenerator(std::move(merged_batch_gen),
+                                       scan_options->fragment_readahead);
+  } else {
+    batch_gen = merged_batch_gen;
```
lidavidm, line 1008: nit: move here?

lidavidm (approved):

PR #30309, opened 2023-01-11
GH-15200: [C++] Add specific ndigits value for round benchmarks.

### LINKED ISSUE #15200: [C++][Arrow] Add benchmarks for the round functions.

EpsilonPrime:
Sets a specific non-negative ndigits value to ensure that the integer-based round benchmarks are exercising a useful codepath.

This makes sense as a call to Round with an integer type and an ndigits=0 option or parameter should be optimized away by any reasonable plan optimizer.
* Closes: #15200


EpsilonPrime:
`Round<Int64Type, RoundMode::DOWN>/size:4194304/inverse_null_proportion:100                        2776893 n`

EpsilonPrime:
I took a deeper look at these behavior and found that aren't any integer kernels for the round functionality.  Ceil/Trunc/Floor only have floating point types implemented.  Round has floating point types as well as decimal types (which we don't yet have benchmarks for).  The benchmarks for these types are likely dominated by conversion to a floating point type.

cpp/src/arrow/compute/kernels/scalar_round_benchmark.cc
```diff
@@ -45,6 +45,7 @@ static void RoundArrayBenchmark(benchmark::State& state, const std::string& func
       rand.Numeric<ArrowType>(array_size, min, max, args.null_proportion));
   RoundOptions options;
   options.round_mode = static_cast<RoundMode>(Mode);
+  options.ndigits = 1;  // Round to the appropriate tenth.
```
cyb70289, line 48: Looks the default `ndigits=0` is the common case.

EpsilonPrime: For floating point types, yes.   For integer types, `ndigits=0` should be optimized away.  I suppose I could set it up to run 0/1 for floating point and just 1 for integer types.  Not that it makes a difference here but the two arg version of round (the version that will be exposed to Substrait) is going to be using random values from -6 to +6.

EpsilonPrime:
Obsolete after incoming round benchmark changes.

PR #33603, opened 2023-01-11
GH-32240: [C#] Support decompression of IPC format buffers

### LINKED ISSUE #32240: [C#] Add decompression support for Record Batches
C# Implementation does not support reading batches written in other implementations of Arrow when the compression is specified in IPC Write options.

e.g. Reading this batch from pyarrow in C# will fail:

pyarrow.ipc.RecordStreamBatchWriter(sink, schema, options=pyarrow,ipcWriteOptions(compression="lz4"))

 

This is to support decompression (lz4 & zstd) in the C# implementation.

**Reporter**: [Rishabh Rana](https://issues.apache.org/jira/browse/ARROW-16921) / @rishabh-rana
#### Related issues:
- [[C#] Support compression in IPC format](https://github.com/apache/arrow/issues/32370) (is duplicated by)

<sub>**Note**: *This issue was originally created as [ARROW-16921](https://issues.apache.org/jira/browse/ARROW-16921). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

adamreeve:

Closes https://github.com/apache/arrow/issues/32240 / https://issues.apache.org/jira/browse/ARROW-16921

This PR implements decompression support for Arrow IPC format files and streams in the dotnet/C# library.

This PR originally provided a reflection based implementation to avoid pulling in new dependencies but now only adds a new `ICompressionCodecFactory` interface that users must implement for decompression support, although I intend to add a new package that implements this in #33893. The original description is below:

> The main concern raised in the above Jira issue was that we don't want to add new NuGet package dependencies to support decompression formats that won't be needed by most users, so a default `CompressionProvider` implementation has been added that uses reflection to use the `ZstdNet` package for ZSTD decompression and `K4os.Compression.LZ4.Streams` and `CommunityToolkit.HighPerformance` for LZ4 Frame support if they are available. The `netstandard1.3` target has decompression support disabled due to some reflection functionality being missing, and neither `ZstdNet` or `K4os.Compression.LZ4.Streams` support `netstandard1.3`.
> 
> The `ArrowFileReader` and `ArrowStreamReader` constructors accept an `ICompressionProvider` parameter to allow users to provide their own compression provider if they want to use different dependencies.
> 
> ### Alternatives to consider
> 
> An alternative approach that could be considered instead of reflection is to use these extra dependencies as build time dependencies but not make them dependencies of the NuGet package. I tested this out in https://github.com/adamreeve/arrow/commit/4544afde6fef12337c7b188cc497da0bc1bf829d and it seems to work reasonably well too but required bumping the version of `System.Runtime.CompilerServices.Unsafe` under the `netstandard2.0` and `netcoreapp3.1` targets. This reduces all the reflection boilerplate but seems pretty hacky as now Apache.Arrow.dll depends on these extra dlls and we rely on the dotnet runtime behaviour of not trying to load them until they're used. So I think the reflection approach is better.
> 
> Another alternative would be to move decompression support into a separate NuGet package (eg.  `Apache.Arrow.Compression`) that depends on `Apache.Arrow` and has an implementation of `ICompressionProvider` that users can pass in to the `ArrowFileReader` constructor, or maybe has a way to register itself with the `Apache.Arrow` package so it only needs to be configured once. That would seem cleaner to me but I'm not sure how much work it would be to set up a whole new package.

# Are these changes tested?

Yes, new unit tests have been added. Test files have been created with a Python script that is included in the PR due to only decompression support being added and not compression support.

Yes, this implements a new feature but in a backwards compatible way.
* Closes: #32240


lidavidm:
FWIW, the Java implementation does take the approach of separating out the compression dependencies into a new package, with only the interface (and a no-op implementation) in the core package. You do have to know to pass the compression codec factory explicitly in that case, unfortunately.

adamreeve:
Thanks @lidavidm, yes I think having a separate package would make sense for dotnet too, and maybe that can be added in a separate PR so for now I've stripped this back to remove the reflection based implementation and just have an implementation in the test project.

I've also renamed things to be a bit more consistent with the Java API, although I've kept the `ICompressionCodec` API more narrowly focused so it only deals with decompression of dotnet memory buffers and Arrow buffer creation is handled by the internal `IBufferCreator` implementation, whereas in the Java API this is part of the `CompressionCodec` interface.

One question I have is whether it makes sense for the public compression API (`ICompressionCodec` and `ICompressionCodecFactory`) to live in the `Apache.Arrow.Ipc` namespace, or whether this should be moved down to `Apache.Arrow` if there's a possibility it might be used outside of the IPC format?

lidavidm:
Java also puts the compression interface in its IPC package. C++ treats it as a more general utility (since it's also used for Parquet, etc.). Strictly in terms of Arrow I think it's fine to be in Ipc.

csharp/src/Apache.Arrow/Ipc/DecompressingBufferCreator.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        public ArrowBuffer CreateBuffer(ReadOnlyMemory<byte> source)
+        {
+            if (source.Length < 8)
+            {
+                throw new Exception($"Invalid compressed data buffer size ({source.Length}), expected at least 8 bytes");
+            }
+
+            // First 8 bytes give the uncompressed data length
+            var uncompressedLength = BinaryPrimitives.ReadInt64LittleEndian(source.Span.Slice(0, 8));
```
wjones127, line 44: This seems sensible, but I couldn't find where in the spec this come from?

wjones127 (approved):
This all looks good! I just had a minor question.

csharp/src/Apache.Arrow/Ipc/DecompressingBufferCreator.cs
adamreeve: The buffer compression strategy is described at https://github.com/apache/arrow/blob/apache-arrow-11.0.0/format/Message.fbs#L59, I'll add a reference to this in a comment

wjones127: Ah it was right there. Thanks!

wjones127 (approved):

westonpace:
I see a script to generate some .arrow files and then also those .arrow files checked in.  Did you mean to commit the generated arrow files?

adamreeve:
> I see a script to generate some .arrow files and then also those .arrow files checked in. Did you mean to commit the generated arrow files?

Yes, I didn't think we'd want to depend on being able to run Python with PyArrow for the dotnet tests to run, but wanted to keep the script there partly as documentation on what should be in the files and so that people could regenerate them if needed. Do you think there's a better way to handle this?

westonpace:
> Yes, I didn't think we'd want to depend on being able to run Python with PyArrow for the dotnet tests to run

I agree it is nice not to need this.

> Do you think there's a better way to handle this?

For a few small files I don't think it's something we need to worry about too much.  There is a separate repository however, https://github.com/apache/arrow-testing which stores various test binaries and is already included as a submodule in apache/arrow.  If you find you need to generate more files in the future you could consider storing them there.

csharp/src/Apache.Arrow/Ipc/ArrowStreamReader.cs
```diff
@@ -31,26 +31,41 @@ public class ArrowStreamReader : IArrowReader, IDisposable
...
+        public ArrowStreamReader(Stream stream, MemoryAllocator allocator, ICompressionCodecFactory compressionCodecFactory, bool leaveOpen)
         {
             if (stream == null)
                 throw new ArgumentNullException(nameof(stream));
 
-            _implementation = new ArrowStreamReaderImplementation(stream, allocator, leaveOpen);
+            _implementation = new ArrowStreamReaderImplementation(stream, allocator, compressionCodecFactory, leaveOpen);
         }
 
         public ArrowStreamReader(ReadOnlyMemory<byte> buffer)
```
Tommo56700, line 71: Why not add decoder support for ArrowMemoryReaderImplementation?

adamreeve: That wasn't intentional, I'd missed that there was a separate implementation for reading from a memory buffer. I'll make a PR to fix this.

adamreeve: PR opened at #34108

PR #33608, opened 2023-01-11
GH-33607: [C++] Support optional additional arguments for inline visit functions

### LINKED ISSUE #33607: [C++] Support optional additional arguments for inline visit functions

Sometimes we need extra arguments for the `Visit` function, the most common of which is output parameters to save the result. For now, we need to have a member variable and a member function to get the result:
```cpp
class ExampleVisitor {
 public:
  template <typename T>
  Status Visit(const T& arr) {
    /// Do stuff and save result to output_;
    return Status::OK();
  }

  std::shared_ptr<Array> GetOutput() { return output_; }

 private:
  std::shared_ptr<Array> output_;
};

ExampleVisitor visitor;
RETURN_NOT_OK(VisitArrayInline(*arr, &visitor));
*output = visitor.GetOutput();
```

It will be more convenient to write a Visitor if the VisitArrayInline function supports additional args for the `Visit` method:
```cpp
class ExampleVisitorWithArg {
 public:
  template <typename T>
  Status Visit(const T& arr, std::shared_ptr<Array>* output) {
    /// Do stuff and save result to output directly;
    return Status::OK();
  }
};

ExampleVisitorWithArg visitor;
RETURN_NOT_OK(VisitArrayInline(*arr, &visitor, &output));
```

### Component(s)

C++

js8544:

Closes #33607

# Rationale for this change

Sometimes we need extra arguments for the `Visit` function, the most common of which is output parameters to save the result. For now, we need to have a member variable and a member function to get the result:
```cpp
class ExampleVisitor {
 public:
  template <typename T>
  Status Visit(const T& arr) {
    /// Do stuff and save result to output_;
    return Status::OK();
  }

  std::shared_ptr<Array> GetOutput() { return output_; }

 private:
  std::shared_ptr<Array> output_;
};

ExampleVisitor visitor;
RETURN_NOT_OK(VisitArrayInline(*arr, &visitor));
*output = visitor.GetOutput();
```

It will be more convenient to write a Visitor if the VisitArrayInline function supports additional args for the `Visit` method:
```cpp
class ExampleVisitorWithArg {
 public:
  template <typename T>
  Status Visit(const T& arr, std::shared_ptr<Array>* output) {
    /// Do stuff and save result to output directly;
    return Status::OK();
  }
};

ExampleVisitorWithArg visitor;
RETURN_NOT_OK(VisitArrayInline(*arr, &visitor, &output));
```

# Are these changes tested?

Not sure if we need to explicitly test this feature, since no runtime behavior is changed. The existing tests all pass.

There is a small chance, like in [diff.cc:402](https://github.com/apache/arrow/pull/33608/files#diff-ef9d394f251ee719f9896b864246bd97a81165c08195d4846d5d99e589e5c1c0L402), `VisitTypeInline` is declared as a friend function. The declaration needs to be changed. But I am not sure if any outside user would do this.

### DIFF
--- a/cpp/src/arrow/visit_array_inline.h
+++ b/cpp/src/arrow/visit_array_inline.h
@@ -28,7 +28,7 @@ namespace arrow {
     return visitor->Visit(                                                               \
         internal::checked_cast<const typename TypeTraits<TYPE_CLASS##Type>::ArrayType&>( \
             array),                                                                      \
-        args...);
+        std::forward<ARGS>(args)...);
 
 /// \brief Apply the visitors Visit() method specialized to the array type
 ///
--- a/cpp/src/arrow/visit_array_inline.h
+++ b/cpp/src/arrow/visit_array_inline.h
@@ -27,11 +27,14 @@ namespace arrow {
   case TYPE_CLASS##Type::type_id:                                                        \
     return visitor->Visit(                                                               \
         internal::checked_cast<const typename TypeTraits<TYPE_CLASS##Type>::ArrayType&>( \
-            array));
+            array),                                                                      \
+        args...);
 
 /// \brief Apply the visitors Visit() method specialized to the array type
 ///
 /// \tparam VISITOR Visitor type that implements Visit() for all array types.
+/// \tparam ARGS Additional arguments, if any, will be passed to the Visit function after
+/// the `arr` argument
 /// \return Status
 ///
 /// A visitor is a type that implements specialized logic for each Arrow type.
@@ -46,8 +49,8 @@ namespace arrow {
 /// ExampleVisitor visitor;
 /// VisitArrayInline(some_array, &visitor);
 /// ```
-template <typename VISITOR>
-inline Status VisitArrayInline(const Array& array, VISITOR* visitor) {
+template <typename VISITOR, typename... ARGS>
+inline Status VisitArrayInline(const Array& array, VISITOR* visitor, ARGS&&... args) {
   switch (array.type_id()) {
     ARROW_GENERATE_FOR_ALL_TYPES(ARRAY_VISIT_INLINE);
     default:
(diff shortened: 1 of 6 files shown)


cpp/src/arrow/visit_array_inline.h
```diff
@@ -27,11 +27,14 @@ namespace arrow {
   case TYPE_CLASS##Type::type_id:                                                        \
     return visitor->Visit(                                                               \
         internal::checked_cast<const typename TypeTraits<TYPE_CLASS##Type>::ArrayType&>( \
-            array));
+            array),                                                                      \
+        args...);
```
pitrou, line 31: Wouldn't it be better to use something like `std::forward<decltype(args)>`?

js8544: Definitely! Changed to std::forward

pitrou:
Ok, is it possible to exploit this new possibility in some select places? It will validate that it works and also show the benefits.

(you can of course grep through the codebase to find candidates)

cpp/src/arrow/visit_type_inline.h
```diff
@@ -71,31 +77,33 @@ inline Status VisitTypeInline(const DataType& type, VISITOR* visitor) {
...
+template <typename VISITOR, typename... ARGS>
+inline auto VisitType(const DataType& type, VISITOR&& visitor, ARGS&&... args)
+    -> decltype(std::forward<VISITOR>(visitor)(type, args...)) {
   switch (type.id()) {
     ARROW_GENERATE_FOR_ALL_TYPES(TYPE_VISIT_INLINE);
     default:
       break;
   }
-  return std::forward<VISITOR>(visitor)(type);
+  return std::forward<VISITOR>(visitor)(type, args...);
```
pitrou, line 88: Should forward here and below as well.

js8544: done

js8544:
> Ok, is it possible to exploit this new possibility in some select places? It will validate that it works and also show the benefits.
> 
> (you can of course grep through the codebase to find candidates)

Good point. I've applied the new `VisitArrayInline`in column_writer.cc, which saves a copy construction of `std::shared_ptr<Buffer>`. And `VisitTypeInline` in builder.h, which eliminates the need for a wrapper for the visitor.

pitrou (approved):
LGTM, thank you @js8544

PR #33613, opened 2023-01-11
GH-15189: [R] Skip S3 tests on MacOS 10.13

### LINKED ISSUE #15189: [CI][R] `mutex lock failed: Invalid argument` on MacOS 10.13 on process shutdown

In the MacOS 10.13 runners, we see:

```
libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
```

On the latest nightly build ( https://github.com/ursacomputing/crossbow/actions/runs/3827980013/jobs/6513063333#step:7:22433 ).

Jacob did some research and it seems this might be something along the lines of a mutex lock attempt after the mutex has been destructed ( https://stackoverflow.com/questions/66773247/libcabi-dylib-terminating-with-uncaught-exception-of-type-std-1system-er ).

Unfortunately I don't have an Apple Clang-12 environment locally and so I'm not able to reproduce this at the moment. Perhaps even more unfortunately, MacOS 10.13 is the environment CRAN uses to build the R package binary that most Intel mac users will use.

### Component(s)

C++, R

paleolimbot:

Closes #15189

# Rationale for this change

The curl/ssl runtime on 10.13 results in a segfault when the process exits (even though all tests pass), so we get a spurious failure on our 10.13 runner test.

Updates the `skip_if_not_available()` function to special case the "s3" feature. The "re2" feature is handled similarly to prevent spurious valgrind errors from being reported.

# Are these changes tested?

These changes can only be tested via crossbow + the 10.13 runner (and locally on my own 10.13 machine after the PR is live and can be pulled there).

Nope!
* Closes: #15189

### DIFF
--- a/r/R/arrow-package.R
+++ b/r/R/arrow-package.R
@@ -205,6 +205,11 @@ on_linux_dev <- function() {
     grepl("devel", R.version.string)
 }
 
+on_macos_10_13_or_lower <- function() {
+  identical(unname(Sys.info()["sysname"]), "Darwin") &&
+    package_version(unname(Sys.info()["release"])) < "18.0.0"
+}
+
 option_use_threads <- function() {
   !is_false(getOption("arrow.use_threads"))
 }
--- a/r/tests/testthat/helper-skip.R
+++ b/r/tests/testthat/helper-skip.R
@@ -43,7 +43,7 @@ skip_if_not_available <- function(feature) {
   if (feature == "s3") {
     sys_info <- Sys.info()
     is_macos_1013 <- identical(unname(Sys.info()["sysname"]), "Darwin") &&
-      identical(unname(Sys.info()["release"]), "17.7.0")
+      package_version(unname(Sys.info()["release"])) < "18.0.0"
     if (is_macos_1013) {
       skip("curl/ssl runtime on MacOS 10.13 is too old")
     }
--- a/r/tests/testthat/helper-skip.R
+++ b/r/tests/testthat/helper-skip.R
@@ -41,10 +41,7 @@ skip_if_not_available <- function(feature) {
   # curl/ssl on MacOS is too old to support S3 filesystems without
   # crashing when the process exits.
   if (feature == "s3") {
-    sys_info <- Sys.info()
-    is_macos_1013 <- identical(unname(Sys.info()["sysname"]), "Darwin") &&
-      package_version(unname(Sys.info()["release"])) < "18.0.0"
-    if (is_macos_1013) {
+    if (on_macos_10_13_or_lower()) {
       skip("curl/ssl runtime on MacOS 10.13 is too old")
     }
   }
--- a/r/tests/testthat/helper-skip.R
+++ b/r/tests/testthat/helper-skip.R
@@ -38,6 +38,17 @@ skip_if_not_available <- function(feature) {
     skip_on_linux_devel()
   }
 
+  # curl/ssl on MacOS is too old to support S3 filesystems without
+  # crashing when the process exits.
+  if (feature == "s3") {
+    sys_info <- Sys.info()
+    is_macos_1013 <- identical(unname(Sys.info()["sysname"]), "Darwin") &&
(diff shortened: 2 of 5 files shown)


kou:
> The curl/ssl runtime on 10.13 results in a segfault when the process exits (even though all tests pass), so we get a spurious failure on our 10.13 runner test.

Oh... How about disabling S3 support instead of skipping S3 tests?

```diff
diff --git a/dev/tasks/homebrew-formulae/autobrew/apache-arrow-static.rb b/dev/tasks/homebrew-formulae/autobrew/apache-arrow-static.rb
index 84206978f3..b5f5158cde 100644
--- a/dev/tasks/homebrew-formulae/autobrew/apache-arrow-static.rb
+++ b/dev/tasks/homebrew-formulae/autobrew/apache-arrow-static.rb
@@ -35,7 +35,7 @@ class ApacheArrowStatic < Formula
 
   depends_on "boost" => :build
   depends_on "cmake" => :build
-  depends_on "aws-sdk-cpp-static"
+  depends_on "aws-sdk-cpp-static" if MacOS.version > :hige_sierra
   depends_on "brotli"
   depends_on "lz4"
   depends_on "snappy"
@@ -62,7 +62,7 @@ class ApacheArrowStatic < Formula
       -DARROW_JSON=ON
       -DARROW_MIMALLOC=ON
       -DARROW_PARQUET=ON
-      -DARROW_S3=ON
+      -DARROW_S3=#{MacOS.version > :hige_sierra ? "ON" : "OFF"}
       -DARROW_VERBOSE_THIRDPARTY_BUILD=ON
       -DARROW_WITH_BROTLI=ON
       -DARROW_WITH_BZ2=ON
diff --git a/dev/tasks/homebrew-formulae/autobrew/apache-arrow.rb b/dev/tasks/homebrew-formulae/autobrew/apache-arrow.rb
index cb55c3a7d9..cb09d8f614 100644
--- a/dev/tasks/homebrew-formulae/autobrew/apache-arrow.rb
+++ b/dev/tasks/homebrew-formulae/autobrew/apache-arrow.rb
@@ -32,7 +32,7 @@ class ApacheArrow < Formula
   # NOTE: if you add something here, be sure to add to PKG_LIBS in r/tools/autobrew
   depends_on "boost" => :build
   depends_on "cmake" => :build
-  depends_on "aws-sdk-cpp"
+  depends_on "aws-sdk-cpp" if MacOS.version > :hige_sierra
   depends_on "brotli"
   depends_on "lz4"
   depends_on "snappy"
@@ -54,7 +54,7 @@ class ApacheArrow < Formula
       -DARROW_JSON=ON
       -DARROW_MIMALLOC=ON
       -DARROW_PARQUET=ON
-      -DARROW_S3=ON
+      -DARROW_S3=#{MacOS.version > :high_sierra ? "ON" : "OFF"}
       -DARROW_VERBOSE_THIRDPARTY_BUILD=ON
       -DARROW_WITH_BROTLI=ON
       -DARROW_WITH_BZ2=ON
```

assignUser:
maybe both? :D

assignUser:
Oh I forgot that the bianry would than have s3 disabled :scream: good catch!

>I'm not sure we have an explicit test of a static build of the R package on 10.13 that is then tested on a higher MacOS version

Not that I am aware of? Not sure that it's necessary as the test are run on e.g. the M1 builds and a bunch of intel mac builds too.

Also I recently skipped some pyarrow test on 10.13, could you unify this with the skip here? 
https://github.com/apache/arrow/commit/1aa8f3554781de81630f3a334e146a73177a2fa3#diff-6175f71a19c38052968303a04b873b8a73cac854aceb91c879f3f49a8f4c9357R28

assignUser:
Awesome thanks! I canceled the previous run to free up the runner :)

assignUser (approved):
This is one of the longest lasting fail streaks in the nightly report finally broken! :tada: Thanks a lot @paleolimbot !

PR #33615, opened 2023-01-11
GH-14997: [Release] Ensure archery release tasks works with both new style GitHub issues and old style JIRA issues

### LINKED ISSUE #14997: [Release][Archery] Update archery release curate to support GitHub issues

During a release before creating a release candidate we use the `archery release curate <version>` command to generate a report with tickets that have a wrong version assigned on JIRA. A summary of the report looks like:

```
Total number of JIRA tickets assigned to version 11.0.0: 367

Total number of applied patches since version 10.0.0: 337

Patches with assigned issue in version 11.0.0:
- https://github.com/apache/arrow/commit/180b1c2ac8aebe9055622e7d9b947916f4cc84ae ARROW-14832: [R] Implement bindings for stringr::str_remove and stringr::str_remove_all (#14644)
...
JIRA issues in version 11.0.0 without a linked patch:
- https://issues.apache.org/jira/browse/ARROW-18437
```
This helps us remove issues that have a wrong version, etcetera.
We should update this report to support GitHub issues showing issues that are incorrectly milestoned on the version we are going to Release.
It would also be nice if we update the number of GitHub issues and the Patches with assigned issues to show also the ones from GitHub. 

### Component(s)

Archery

### LINKED ISSUE #14999: [Release][Archery] Update archery release changelog to support GitHub issues

The `archery release changelog` command is currently used on a couple of steps during the release.
- https://github.com/apache/arrow/blob/master/dev/release/01-prepare.sh#L73-L79
- https://github.com/apache/arrow/blob/master/dev/release/post-03-website.sh#L159-L160

The first one appends issues to the current Changelog on the maintenance branch. Example: https://github.com/apache/arrow/commit/6776681ce08a2811419fc7d4628abe510783cc69 
The second one updates the website announcement with the Changelog.

Currently this only picks issues from JIRA so our changelog will be incomplete for the next Release.

I've added an example of the current output on this public gist: https://gist.github.com/raulcd/6063ef71d5301fcbce88109748624d51

### Component(s)

Archery

### LINKED ISSUE #15002: [Release][Archery] Update archery release cherry-pick to support GitHub issues

We use the `archery release cherry-pick` command during the Release in order to update the maintenance branch once it has been created in order to pick the individual commits that are tagged on JIRA with the corresponding version:
https://github.com/apache/arrow/blob/master/docs/source/developers/release.rst#create-or-update-the-corresponding-maintenance-branch

We should update the `cherry-pick` command to allow picking individual commits from GitHub issues that are Milestoned with the new version too otherwise we have to manually cherry-pick those issues from GitHub.

An example of execution (the local `maint-11.0.0` branch is fake from `maint-10.0.0`):
```
$ archery release --jira-cache /tmp/jiracache cherry-pick 11.0.0 --continue
git checkout -b maint-11.0.0 master
git cherry-pick 9b23bc45778d9191b229b61c036bd348b1e11602
git cherry-pick 66874f5a075525dd440f4d60e0e54042dcbe71ad
git cherry-pick e0ca46598fbaba01afe2feef9eeb7385ddf95028
```
Those commits are tagged with version `11.0.0` on JIRA but are not on the branch, we should replicate that workflow with GitHub issues that are part of the milestone.

This is the main logic for which commits to pick: https://github.com/apache/arrow/blob/master/dev/archery/archery/release/core.py#L465-L488

### Component(s)

Archery

raulcd:
I've decided to do all the archery release tasks on a single PR:
* Closes: #14997
* Closes: #14999
* Closes: #15002

### DIFF
--- a/dev/archery/archery/release/core.py
+++ b/dev/archery/archery/release/core.py
@@ -25,6 +25,7 @@ import shelve
 import warnings
 
 from git import Repo
+from github import Github
 from jira import JIRA
 from semver import VersionInfo as SemVer
 
@@ -58,13 +59,28 @@ class Version(SemVer):
             release_date=getattr(jira_version, 'releaseDate', None)
         )
 
+    @classmethod
+    def from_milestone(cls, milestone):
+        return cls.parse(
+            milestone.title,
+            released=milestone.state == "closed",
+            release_date=milestone.due_on
+        )
+
+
+ORIGINAL_ARROW_REGEX = re.compile(
+    r"\*This issue was originally created as " +
+    r"\[(?P<issue>ARROW\-(?P<issue_id>(\d+)))\]"
+)
+
 
 class Issue:
 
-    def __init__(self, key, type, summary):
+    def __init__(self, key, type, summary, github_issue=None):
         self.key = key
         self.type = type
         self.summary = summary
+        self._github_issue = github_issue
 
     @classmethod
     def from_jira(cls, jira_issue):
@@ -74,6 +90,23 @@ class Issue:
             summary=jira_issue.fields.summary
         )
 
+    @classmethod
+    def from_github(cls, github_issue):
+        original_jira = cls.original_jira_id(github_issue)
+        key = original_jira or github_issue.number
+        return cls(
+            key=key,
+            type=next(
+                iter(
+                    [
+                        label for label in github_issue.labels
+                        if label.name.startswith("Type:")
+                    ]
+                ), None),
+            summary=github_issue.title,
+            github_issue=github_issue
+        )
+
     @property
     def project(self):
         return self.key.split('-')[0]
@@ -81,6 +114,25 @@ class Issue:
     @property
     def number(self):
         return int(self.key.split('-')[1])
+
+    @cached_property
+    def pr(self):
+        if self.is_pr:
+            return self._github_issue.as_pull_request()
+
+    @cached_property
+    def is_pr(self):
+        return bool(self._github_issue and self._github_issue.pull_request)
+
+    @classmethod
+    def original_jira_id(cls, github_issue):
+        # All migrated issues contain body
+        if not github_issue.body:
+            return None
+        matches = ORIGINAL_ARROW_REGEX.search(github_issue.body)
+        if matches:
+            values = matches.groupdict()
+            return values['issue']
 
 
 class Jira(JIRA):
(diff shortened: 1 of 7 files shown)


raulcd:
This is an example of how the new changelog for 11.0.0 will look with a mix of issues from JIRA and GITHUB: https://gist.github.com/raulcd/1eebbfc28e890d9abbb248db48ba5775

dev/archery/archery/release/core.py
```diff
@@ -285,8 +354,8 @@ def __repr__(self):
         return f"<{self.__class__.__name__} {self.version!r} {status}>"
 
     @staticmethod
-    def from_jira(version, jira=None, repo=None):
-        return Release(version, jira, repo)
+    def from_issue_tracker(version, issue_tracker=None, repo=None):
```
kou, line 357: Can we remove this? It seems that this is used only for test.

dev/archery/archery/release/core.py
```diff
@@ -269,13 +336,15 @@ def __init__(self, version, jira, repo):
...
+            version = issue_tracker.project_version(version)
+
         elif not isinstance(version, Version):
             raise TypeError(version)
 
         self.version = version
-        self.jira = jira
         self.repo = repo
+        self.issue_tracker = issue_tracker
+        self._github_token = github_token
```
kou, line 347: It seems that this is needless.

dev/archery/archery/release/core.py
```diff
@@ -250,15 +323,9 @@ def __new__(self, version, jira=None, repo=None):
...
-            jira = Jira()
-        elif isinstance(jira, str):
-            jira = Jira(jira)
-        elif not isinstance(jira, (Jira, CachedJira)):
-            raise TypeError("`jira` argument must be a server url or a valid "
-                            "Jira instance")
-
+    def __init__(self, version, repo, github_token=None, issue_tracker=None):
+        if not issue_tracker:
+            issue_tracker = IssueTracker(github_token=github_token)
```
kou, line 328: How about requiring `issue_tracker` instead of accepting optional `github_token`?

dev/archery/archery/release/core.py
```diff
@@ -74,68 +90,122 @@ def from_jira(cls, jira_issue):
...
+        if self._github_issue:
+            return [label for label in self._github_issue.labels
+                    if label.name.startswith("Component:")]
+
+    def is_component(self, component_name):
+        if components := self.components:
+            for component in components:
+                if component.name == f"Component: {component_name}":
+                    return True
+        return False
```
kou, line 143: It seems that they aren't used.

dev/archery/archery/release/core.py
```diff
@@ -74,68 +90,122 @@ def from_jira(cls, jira_issue):
...
-        return int(self.key.split('-')[1])
+        if isinstance(self.key, str):
+            return int(self.key.split('-')[1])
+        else:
+            return self.key
+
+    @cached_property
+    def pr(self):
+        if self.is_pr:
+            return self._github_issue.as_pull_request()
```
kou, line 126: It seems that this is not used.

dev/archery/archery/release/core.py
```diff
@@ -74,68 +90,122 @@ def from_jira(cls, jira_issue):
...
+
+    @cached_property
+    def is_pr(self):
+        return bool(self._github_issue and self._github_issue.pull_request)
```
kou, line 130: It seems that `self._github_issue` is needed only for this. How about setting `self.is_pr` instead of keeping `self._github_issue`?

dev/archery/archery/release/core.py
raulcd: I did use both the PR and components at one point when I was testing things and thought that there was no harm on keeping them in case we want to improve / add more functionality to the scripts to have them available but I plan for a bigger refactor on archery in general an I'll tackle creating a small utility library around those in the future so I've removed it as it is currently not used.

dev/archery/archery/release/core.py
raulcd: removed

rok (approved):

dev/archery/archery/release/core.py
kou: I see. I don't object it.

FYI: Reusing this class for Jira issue may not be a good idea. If I implement this, I'll define `GitHubIssue(Issue)`  and `JiraIssue(Issue)` and `JiraIssue.is_pr` always returns `False`.

kou (approved):
+1 after we rebase on master.

pitrou:
Style issue: when reading "Archery release tasks takes into account both GitHub and old Apache Arrow JIRA issues", we don't know whether that's describing the situation pre-PR or post-PR.

To fix this, it is preferrable to use an infinitive to describe what the PR *does*. For example: "Ensure that Archery release tasks work with both ...".

@wjones127 for suggestions

PR #33620, opened 2023-01-11
GH-33619:  [Documentation] Update PR template

### LINKED ISSUE #33619: [Documentation] Update PR template
There were some additonal reviews and comments in the initial PR after merge e.g.:
> If we start to use this, that will get duplicated with the `* Closes ..` that the bot adds. Do we then want to remove the bot action? (or some way to unify those? but since this is free form, it might be difficult to have the bot check it)

_Originally posted by @jorisvandenbossche in https://github.com/apache/arrow/pull/15250#discussion_r1066727192_

assignUser:
### Rationale for this change

I will keep this as a draft for now please feel free to suggest additional changes!

I have switched to smaller headers for now and removed the first head + closes keyword to avoid redundancy.

* Closes: #33619

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,4 +1,3 @@
-# Which issue does this PR close?
 
 <!--
 Thanks for opening a pull request!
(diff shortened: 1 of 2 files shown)


kevingurney:
Looks good to me! Thanks for making this change, @assignUser!

jorisvandenbossche (approved):

jorisvandenbossche:
> Following up on the discussion about stripping comments from the pull request description when merging, I believe adding the following around [line 583 of merge_arrow_pr.py](https://github.com/apache/arrow/blob/c287cfe3cf384bd8b9faf1a52d439b902a457f8d/dev/merge_arrow_pr.py#L583) should work:

Short term that would be good to do to avoid those long noisy commit messages. Longer term, IMO we should discuss using only the title of the PR.

.github/pull_request_template.md
```diff
@@ -50,7 +48,7 @@ We typically require tests for all PRs in order to:
 If tests are not included in your PR, please explain why (for example, are they covered by existing tests)?
 -->
 
-# Are there any user-facing changes?
+### Are there any user-facing changes?
```
wjones127, line 51: Another piece of feedback:

1. The `breaking-change` label didn't exist. I decided to add it as `Breaking Change` just now, since there isn't a good reason to kebab-case AFAIK.
2. I don't think non-committers can apply labels, right? So perhaps the instructions simply need to have a checkbox `[ ] This PR contains breaking changes` or something like that?

wjones127: Actually, I'll handle this in #33660

jorisvandenbossche (approved):
This is already a big improvement IMO, so let's merge this

jorisvandenbossche:
Whoops, sorry, the linter was failing on the code line in archery .. This will be fixed by https://github.com/apache/arrow/pull/15067

PR #33634, opened 2023-01-12
GH-20385: [C++][Parquet] Reject partial load of an extension type

westonpace:
When the parquet reader is performing a partial read it will try and maintain field structure.  So, for example, given a schema of `points: struct<x: int32, y:int32>` and a load of `points.x` it will return `points: struct<x: int32>`.  However, if there is an extension type `points: Point` where `Point` has a storage type `struct<x: int32, y: int32>` then suddenly the reference `points.x` no longer makes sense.
* Closes: #20385

### DIFF
--- a/cpp/src/parquet/arrow/reader.cc
+++ b/cpp/src/parquet/arrow/reader.cc
@@ -844,8 +844,7 @@ Status GetReader(const SchemaField& field, const std::shared_ptr<Field>& arrow_f
     RETURN_NOT_OK(GetReader(field, storage_field, ctx, out));
     if (*out) {
       auto storage_type = (*out)->field()->type();
-      if (!storage_type->Equals(
-              checked_cast<const ExtensionType&>(*arrow_field->type()).storage_type())) {
+      if (!storage_type->Equals(storage_field->type())) {
         return Status::Invalid(
             "Due to column pruning only part of an extension's storage type was loaded.  "
             "An extension type cannot be created without all of its fields");
--- a/cpp/src/parquet/arrow/reader.cc
+++ b/cpp/src/parquet/arrow/reader.cc
@@ -842,7 +842,16 @@ Status GetReader(const SchemaField& field, const std::shared_ptr<Field>& arrow_f
     auto storage_field = arrow_field->WithType(
         checked_cast<const ExtensionType&>(*arrow_field->type()).storage_type());
     RETURN_NOT_OK(GetReader(field, storage_field, ctx, out));
-    *out = std::make_unique<ExtensionReader>(arrow_field, std::move(*out));
+    if (*out) {
+      auto storage_type = (*out)->field()->type();
+      if (!storage_type->Equals(
+              checked_cast<const ExtensionType&>(*arrow_field->type()).storage_type())) {
+        return Status::Invalid(
+            "Due to column pruning only part of an extension's storage type was loaded.  "
+            "An extension type cannot be created without all of its fields");
+      }
+      *out = std::make_unique<ExtensionReader>(arrow_field, std::move(*out));
+    }
     return Status::OK();
   }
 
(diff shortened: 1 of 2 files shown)


westonpace:
This should fix the issue I believe but it might not be the best fix.  I wonder if there is some place higher up (e.g. where we are converting the path strings into some kind of parquet column references) that we might reject the request.

cpp/src/parquet/arrow/reader.cc
```diff
@@ -842,7 +842,16 @@ Status GetReader(const SchemaField& field, const std::shared_ptr<Field>& arrow_f
     auto storage_field = arrow_field->WithType(
         checked_cast<const ExtensionType&>(*arrow_field->type()).storage_type());
     RETURN_NOT_OK(GetReader(field, storage_field, ctx, out));
-    *out = std::make_unique<ExtensionReader>(arrow_field, std::move(*out));
+    if (*out) {
+      auto storage_type = (*out)->field()->type();
+      if (!storage_type->Equals(
+              checked_cast<const ExtensionType&>(*arrow_field->type()).storage_type())) {
```
pitrou, line 848: Isn't this simply
```suggestion
      if (!storage_type->Equals(storage_field->type()) {
```

jorisvandenbossche:
I think ideally we would still allow reading a field of an extension type, but in that case just return the plain field (not wrapped in an extension type, since that is no longer possible)

westonpace:
> I think ideally we would still allow reading a field of an extension type, but in that case just return the plain field (not wrapped in an extension type, since that is no longer possible)

It could be possible but would require another abstract method in `ExtensionType`.  Something like `std::shared_ptr<ExtensionType> ExtensionType::WithType(std::shared_ptr<Type> storage_type)`.  However, in cases like `Point` that would be meaningless.  However, in some cases like `MyCustomStruct` it might make sense.

jorisvandenbossche:
Sorry, I don't understand your last comment. Why does it need a new method? (you already can get the storage type / array from an extension type / array) And why would that be meaningless for the Point example? (I think it can makes sense to want to read just one of x or y)

westonpace:
Sorry, I will clarify.  If the column `my_points` has the type `Point` and you ask for `my_points.x` then your only option is to return `x: int32`.  However, if the column `my_points` has the type `AnnotatedStruct<x: int32, y: int32>` then it may very well make sense to return `AnnotatedStruct<x: int32>`.

`PointType::WithType(struct<x: int32>)` would return `nullptr`.  `AnnotatedStructType::WithType(struct<x: int32>)` would return something valid (a new data type with the given storage type but reusing the metadata / annotations on the original type).

westonpace:
Actually, never mind, after thinking this through, a nested reference always discards the parents.  The above behavior would only with Substrait's mask references (though it is still a valid concern, that is for later).

It appears that the C++ parquet reader has the ability to support this kind of "masked projection" but the Arrow parquet reader does not expose an API for it.  I'll have to read through in more detail.

amol-:

amol-:
@AlenkaF @wjones127 can you take a look at this one as soon as you have a chance? It has been pending review for a while and it would be great to lead it to completion as it's an helpful improvement.

AlenkaF (approved):
LGTM.

cpp/src/parquet/arrow/reader.cc
wjones127: I think Antoine is correct. This can be simplified.

zeroshade (approved):
this LGTM though I think antoine's suggestion should be applied first.

westonpace:
> this LGTM though I think antoine's suggestion should be applied first.

Agreed.  I've let this slide for a while.  I'll try and refresh this next week.

pitrou:
@westonpace Do you want to update this PR?

PR #33636, opened 2023-01-12
GH-15000: [Dev][Release] Update changelog.md with latest release changelog

### LINKED ISSUE #15000: [Release] Changelog.md on master branch has not been updated since 6.0.1

While investigating https://github.com/apache/arrow/issues/14999 I've realised we never merge from the maintenance branch the Changelog to our master branch and the current Changelog.md on master has not been updated since the 6.0.1 release: https://github.com/apache/arrow/blob/master/CHANGELOG.md
We are updating the changelog on the maintenance branch though: https://github.com/apache/arrow/commit/6776681ce08a2811419fc7d4628abe510783cc69
Maybe we should remove the Changelog from the repository and just generate a version changelog for the website announcement.

Related issues (closed as duplicate):

- https://github.com/apache/arrow/issues/30928
- https://github.com/apache/arrow/issues/18757

### Component(s)

Developer Tools

raulcd:
The release changelog hadn't been updated since the release 6.0.1.

This PR regenerates the changelog adding information up until 10.0.1 using:
```
archery release changelog regenerate
```

Changes have been made to archery to generate it on this PR: https://github.com/apache/arrow/pull/33615 in order to be able to retrieve the issues from GitHub instead of JIRA.

* Closes: #15000

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,30 +1,2392 @@
 
-# Apache Arrow 6.0.1 (2021-11-18)
+# Apache Arrow 10.0.1 (2022-11-22 08:00:00)
+
+## New Features and Improvements
+
+* [ARROW-17487](https://issues.apache.org/jira/browse/ARROW-17487) - [Python][Packaging][CI] Add support for Python 3.11 (#14499)
+* [ARROW-17635](https://issues.apache.org/jira/browse/ARROW-17635) - [Python][CI] Sync conda recipe with the arrow-cpp feedstock (#14102)
+* [ARROW-18054](https://issues.apache.org/jira/browse/ARROW-18054) - [Python][CI] Enable Cython tests on windows wheels (#13552)
+* [ARROW-18080](https://issues.apache.org/jira/browse/ARROW-18080) - [C++] Remove gcc <= 4.9 workarounds (#14441)
+* [ARROW-18092](https://issues.apache.org/jira/browse/ARROW-18092) - [CI][Conan] Update versions of gRPC related dependencies (#14453)
+* [ARROW-18093](https://issues.apache.org/jira/browse/ARROW-18093) - [CI][Conda][Windows] Disable ORC (#14454)
+* [ARROW-18162](https://issues.apache.org/jira/browse/ARROW-18162) - [C++] Add Arm SVE compiler options (#14515)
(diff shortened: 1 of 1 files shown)


raulcd:
I've regenerated the changelog fully as a way of testing my changes to archery release scripts using GitHub instead of JIRA and in order to add the missing releases from the Changelog. There are some minor changes to some old changelogs mainly due to issues being assigned to a single milestone on GitHub but to more than one fix version on JIRA. We used to add tickets to both the minor release and the next major release but on GitHub we have only added it to the minor milestone. Mainly because an issue can only have a single milestone.

As a practical example: https://issues.apache.org/jira/browse/ARROW-3353
This issue appeared on both the 0.11.1 and 0.12.0 changelogs. It now appears only on the 0.11.1 changelog as the GitHub assigned issue is set to the 0.11.1 milestone: https://github.com/apache/arrow/issues/19685

I don't think that is a problem as we are already delivering all tickets that pertain to a minor release on the next major release, like:
- 11.0.0 will include all tickets that were part of 10.0.1 even if they are not included in the 11.0.0 changelog.

@kou any thoughts?

kou:
How about cherry-picking old changelogs from release branch instead of re-generating changelogs?

For example, `git cherry-pick f786b8d5e034217dda420bda10d9749c0b66e971` for 10.0.1: https://github.com/apache/arrow/commit/f786b8d5e034217dda420bda10d9749c0b66e971
(Unfortunately, the command reports a merge conflict... But resolving the conflict isn't hard.)

How about adding a post release task that cherry-picks a changelog to master like ADBC?
FYI: https://github.com/apache/arrow-adbc/pull/335 ADBC replaces the changelog file but I think that we can use cherry-pick approach.

raulcd:
That makes sense to me, we can add a post release task for it and I can cherry pick all the old changelog commits individually to master.

There is still a slight change on the generation of new changelogs due to the GitHub issues being on a single milestone, instead of more than one fix version but I think that's ok for newly generated changelogs.

kou:
> There is still a slight change on the generation of new changelogs due to the GitHub issues being on a single milestone, instead of more than one fix version but I think that's ok for newly generated changelogs.

I think so too.

raulcd:
Closing this as we will take a different approach as discussed on the comments.

PR #33648, opened 2023-01-13
GH-33640: [C++] Add backpressure to asof join node

### LINKED ISSUE #33640: [C++] as-of-join backpressure for large sources

Add backpressure support to as-of-join in order to handle large sources. This is a merge into master of (at least most of) https://github.com/westonpace/arrow/pull/21/files

### Component(s)

C++

westonpace:
If any input starts to accumulate too many batches then we ask that input to slow down.
* Closes: #33640


westonpace:
CC @rtpsw can you please review

cpp/src/arrow/compute/exec/asof_join_node.cc
```diff
@@ -380,6 +501,23 @@ class InputState {
     }
   }
 
+  static Result<std::unique_ptr<InputState>> Make(
+      size_t index, TolType tolerance, bool must_hash, bool may_rehash,
+      KeyHasher* key_hasher, ExecNode* node, ExecNode* output,
+      std::atomic<int32_t>& backpressure_counter,
+      const std::shared_ptr<arrow::Schema>& schema, const col_index_t time_col_index,
+      const std::vector<col_index_t>& key_col_index) {
+    constexpr size_t low_threshold = 4, high_threshold = 8;
```
rtpsw, line 510: Do we want to keep these hard-coded in this PR?

cpp/src/arrow/compute/exec/test_nodes.h
```diff
@@ -0,0 +1,52 @@
...
+
+#include <string>
+
+#include "arrow/compute/exec/options.h"
+#include "arrow/compute/exec/test_util.h"
+
+namespace arrow {
+namespace compute {
+
+struct ConcatNodeOptions : public ExecNodeOptions {
```
rtpsw, line 26: Perhaps this should be named `BackpressureOptions` and moved to `arrow/compute/exec/options.h`, while `ConcatNodeOptions` should derive from it or include it as a member. Not critical for this PR, but seems useful.

rtpsw (approved):

rtpsw:
@westonpace, looked reasonable to me. I approved but posted a couple of comments for consideration.

cpp/src/arrow/compute/exec/asof_join_node.cc
icexelloss: Ok for now. Let's try not make additional changes on top of the original one here.

cpp/src/arrow/compute/exec/test_nodes.h
icexelloss: We can do this as follow up if needed - Let's try to keep the PR as is (unless there is something that we really don't like and want to change in the original PR).

rtpsw (changes requested):

icexelloss (approved):
LGTM.

icexelloss:
Oh wait, slightly confused, which original PR on demo-branch was this? (Or is this combination of multiple)?

rtpsw:
@westonpace, I ended up [requesting a change here](https://github.com/apache/arrow/pull/33648#pullrequestreview-1247776597).

Side note: This [CI job](https://github.com/apache/arrow/actions/runs/3908770342/jobs/6679192718) complains about a missing parameter doc, which should be easy to fix.

cpp/src/arrow/compute/exec/test_nodes.h
westonpace: Hmm, the concat node is only meant as an example for exec node developers right now.  There's no way to actually use it in a deployed environment.  So let's wait until there is a real node that needs backpressure options and then we can move it.

westonpace: And, actually, there is already `arrow::compute::BackpressureOptions` in `exec/options.h` for the sink node.  I think I'll just rip out the concat node for now since it isn't used in tests anymore.

rtpsw:
The [CI job failure](https://github.com/apache/arrow/actions/runs/3917238974/jobs/6696908870) looks unrelated.

rtpsw:
LGTM.

PR #33653, opened 2023-01-13
GH-15289: [Ruby] Return self when saving Table to csv

### LINKED ISSUE #15289: [Ruby] Return other than @table.raw_records when saving Table to csv
#### Target method

Arrow::Table#save

#### Proposed feature

Change return value of Table#save when saving to csv files.

Current code returns `@table.raw_records` (Array of row Arrays). When saving a very big Table, it takes long time to display a large Array with interactive execution such as irb or Jupyter.

When we save as 'arrow' or 'arrows', it simply returns 'true'.

Candidates of return value:

- true
- self
- size of Table

#### Impact of this request

It will improve the user experience of interactive execution. The return value of #save is rarely used.

https://github.com/apache/arrow/blob/master/ruby/red-arrow/lib/arrow/table-saver.rb#L174-L176

### Component(s)

Ruby

heronshoes:
# Rationale for this change

Change return value of Table#save when saving Table to csv files.

Change return value to self (`@table`).

# Are these changes tested?

Add test for saving to csv.

Return value of Table#save is undetermined.

* Closes: #15289

### DIFF
--- a/ruby/red-arrow/lib/arrow/table-saver.rb
+++ b/ruby/red-arrow/lib/arrow/table-saver.rb
@@ -51,6 +51,7 @@ module Arrow
         raise ArgumentError, message
       end
       __send__(custom_save_method)
+      @table
     end
 
     private
--- a/ruby/red-arrow/test/test-table.rb
+++ b/ruby/red-arrow/test/test-table.rb
@@ -720,6 +720,11 @@ class TableTest < Test::Unit::TestCase
                                            schema: @table.schema))
           end
 
+          test("csv, return value") do
+            output = create_output(".csv")
+            assert_equal(@table, @table.save(output))
+          end
+
           test("csv.gz") do
             output = create_output(".csv.gz")
             @table.save(output)


kou:
How about doing this in `TableSaver#save`?

```diff
diff --git a/ruby/red-arrow/lib/arrow/table-saver.rb b/ruby/red-arrow/lib/arrow/table-saver.rb
index f19261cced..c33e641438 100644
--- a/ruby/red-arrow/lib/arrow/table-saver.rb
+++ b/ruby/red-arrow/lib/arrow/table-saver.rb
@@ -51,6 +51,7 @@ module Arrow
         raise ArgumentError, message
       end
       __send__(custom_save_method)
+      @table
     end
 
     private

```

If we did this in `TableSaver#save`, custom save method implementers don't need to care about this.

heronshoes:
Thanks. I will do it there.

kou (approved):
+1

heronshoes:
Thanks!

PR #33656, opened 2023-01-13
GH-33655: [C++][Parquet] Write parquet columns in parallel

### LINKED ISSUE #33655: [C++][Parquet] Write columns in parallel for parquet writer
### Background
he parquet arrow reader supports reading columns in parallel in the executor, but the writer does not support it yet.

### Goal
It is pretty straight-forward to support it in the buffered row group mode when RecordBatch is used. For non-buffered row group mode, it is non-trivial to do the same thing because column chunks are written one by one and flushed to the output. Therefore, this issue aims to support writing columns in parallel in buffered row group only.

### Component(s)

C++, Parquet

wgtmac:

Closes #33655 

 - Add use_threads and executor options to `ArrowWriterProperties`.
 - `parquet::arrow::FileWriter` writes columns in parallel when buffered row group is enabled.
 - Only `WriteRecordBatch()` is supported.

# Are these changes tested?

Added `TEST(TestArrowReadWrite, MultithreadedWrite)` in the `arrow_reader_writer_test.cc`

* Closes: #33655

### DIFF
--- a/cpp/src/parquet/arrow/writer.cc
+++ b/cpp/src/parquet/arrow/writer.cc
@@ -287,13 +287,12 @@ class FileWriterImpl : public FileWriter {
       : schema_(std::move(schema)),
         writer_(std::move(writer)),
         row_group_writer_(nullptr),
+        column_write_context_(pool, arrow_properties.get()),
         arrow_properties_(std::move(arrow_properties)),
         closed_(false) {
     if (arrow_properties_->use_threads()) {
-      column_write_context_.resize(schema_->num_fields(),
-                                   {pool, arrow_properties_.get()});
-    } else {
-      column_write_context_.emplace_back(pool, arrow_properties_.get());
+      parallel_column_write_contexts_.resize(schema_->num_fields(),
+                                             {pool, arrow_properties_.get()});
     }
   }
 
@@ -340,7 +339,7 @@ class FileWriterImpl : public FileWriter {
           std::unique_ptr<ArrowColumnWriterV2> writer,
           ArrowColumnWriterV2::Make(*data, offset, size, schema_manifest_,
                                     row_group_writer_));
-      return writer->Write(&column_write_context_.back());
+      return writer->Write(&column_write_context_);
     }
 
     return Status::NotImplemented("Unknown engine version.");
@@ -411,8 +410,8 @@ class FileWriterImpl : public FileWriter {
 
     auto WriteBatch = [&](int64_t offset, int64_t size) {
       std::vector<std::unique_ptr<ArrowColumnWriterV2>> writers;
-
       int column_index_start = 0;
+
       for (int i = 0; i < batch.num_columns(); i++) {
         ChunkedArray chunkedArray{batch.column(i)};
         ARROW_ASSIGN_OR_RAISE(
@@ -420,21 +419,22 @@ class FileWriterImpl : public FileWriter {
             ArrowColumnWriterV2::Make(chunkedArray, offset, size, schema_manifest_,
                                       row_group_writer_, column_index_start));
         column_index_start += writer->leaf_count();
-        writers.emplace_back(std::move(writer));
+        if (arrow_properties_->use_threads()) {
+          writers.emplace_back(std::move(writer));
+        } else {
+          RETURN_NOT_OK(writer->Write(&column_write_context_));
+        }
       }
 
-      auto executor = arrow_properties_->executor() == nullptr
-                          ? ::arrow::internal::GetCpuThreadPool()
-                          : arrow_properties_->executor();
+      if (arrow_properties_->use_threads()) {
+        DCHECK_EQ(parallel_column_write_contexts_.size(), writers.size());
+        RETURN_NOT_OK(::arrow::internal::OptionalParallelFor(
+            /*use_threads=*/true, static_cast<int>(writers.size()),
+            [&](int i) { return writers[i]->Write(&parallel_column_write_contexts_[i]); },
+            arrow_properties_->executor()));
+      }
 
-      return ::arrow::internal::OptionalParallelFor(
-          arrow_properties_->use_threads(), static_cast<int>(writers.size()),
-          [&](int i) {
-            auto ctx = arrow_properties_->use_threads() ? &column_write_context_.at(i)
-                                                        : &column_write_context_.back();
-            return writers[i]->Write(ctx);
-          },
-          executor);
+      return Status::OK();
     };
 
     int64_t offset = 0;
@@ -457,8 +457,7 @@ class FileWriterImpl : public FileWriter {
   const WriterProperties& properties() const { return *writer_->properties(); }
 
   ::arrow::MemoryPool* memory_pool() const override {
-    DCHECK(!column_write_context_.empty());
-    return column_write_context_.back().memory_pool;
+    return column_write_context_.memory_pool;
   }
 
   const std::shared_ptr<FileMetaData> metadata() const override {
@@ -474,12 +473,14 @@ class FileWriterImpl : public FileWriter {
 
   std::unique_ptr<ParquetFileWriter> writer_;
   RowGroupWriter* row_group_writer_;
-  // If arrow_properties_.use_threads() is true, the vector size is equal to
-  // schema_->num_fields() to make it thread-safe. Otherwise, the vector size
-  // is always 1 which is shared by all column chunks.
-  std::vector<ArrowWriteContext> column_write_context_;
+  ArrowWriteContext column_write_context_;
   std::shared_ptr<ArrowWriterProperties> arrow_properties_;
   bool closed_;
+
+  /// If arrow_properties_.use_threads() is true, the vector size is equal to
+  /// schema_->num_fields() to make it thread-safe. Otherwise, the vector is
+  /// empty and column_write_context_ above is shared by all columns.
+  std::vector<ArrowWriteContext> parallel_column_write_contexts_;
 };
 
 FileWriter::~FileWriter() {}
--- a/cpp/src/parquet/arrow/writer.cc
+++ b/cpp/src/parquet/arrow/writer.cc
@@ -413,10 +413,10 @@ class FileWriterImpl : public FileWriter {
       int column_index_start = 0;
 
       for (int i = 0; i < batch.num_columns(); i++) {
-        ChunkedArray chunkedArray{batch.column(i)};
+        ChunkedArray chunked_array{batch.column(i)};
         ARROW_ASSIGN_OR_RAISE(
             std::unique_ptr<ArrowColumnWriterV2> writer,
-            ArrowColumnWriterV2::Make(chunkedArray, offset, size, schema_manifest_,
+            ArrowColumnWriterV2::Make(chunked_array, offset, size, schema_manifest_,
                                       row_group_writer_, column_index_start));
         column_index_start += writer->leaf_count();
         if (arrow_properties_->use_threads()) {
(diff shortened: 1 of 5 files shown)


wgtmac:
@wjones127 @emkornfield @pitrou Please take a look when you have time. Thanks!

wgtmac:
The new test failed on Windows:
```
[ RUN      ] TestArrowReadWrite.MultithreadedWrite
unknown file: error: SEH exception with code 0xc0000005 thrown in the test body.
[  FAILED  ] TestArrowReadWrite.MultithreadedWrite (12 ms)
```

cpp/src/parquet/arrow/writer.cc
```diff
@@ -333,7 +340,7 @@ class FileWriterImpl : public FileWriter {
           std::unique_ptr<ArrowColumnWriterV2> writer,
           ArrowColumnWriterV2::Make(*data, offset, size, schema_manifest_,
                                     row_group_writer_));
-      return writer->Write(&column_write_context_);
+      return writer->Write(&column_write_context_.back());
```
mapleFU, line 343: Would a writer doesn't have any columns?

cpp/src/parquet/arrow/writer.cc
```diff
@@ -403,17 +410,31 @@ class FileWriterImpl : public FileWriter {
     }
 
     auto WriteBatch = [&](int64_t offset, int64_t size) {
+      std::vector<std::unique_ptr<ArrowColumnWriterV2>> writers;
+
       int column_index_start = 0;
       for (int i = 0; i < batch.num_columns(); i++) {
-        ChunkedArray chunkedArray(batch.column(i));
+        ChunkedArray chunkedArray{batch.column(i)};
```
mapleFU, line 417: Here would `chunkedArray` dtor in stack because it outlives writer's lifetime?

cpp/src/parquet/properties.h
```diff
@@ -843,6 +863,9 @@ class PARQUET_EXPORT ArrowWriterProperties {
     bool store_schema_;
     bool compliant_nested_types_;
     EngineVersion engine_version_;
+
+    bool use_threads_;
+    ::arrow::internal::Executor* executor_;
```
cyb70289, line 868: Is `executor_` guaranteed to be initialized?

cpp/src/parquet/arrow/writer.cc
```diff
@@ -403,17 +410,31 @@ class FileWriterImpl : public FileWriter {
...
         ARROW_ASSIGN_OR_RAISE(
             std::unique_ptr<ArrowColumnWriterV2> writer,
             ArrowColumnWriterV2::Make(chunkedArray, offset, size, schema_manifest_,
                                       row_group_writer_, column_index_start));
-        RETURN_NOT_OK(writer->Write(&column_write_context_));
         column_index_start += writer->leaf_count();
+        writers.emplace_back(std::move(writer));
```
cyb70289, line 423: I understand it's to unify use-thread and non-thread code, but it's looks not ideal to always create such a vector for non-thread case.

cpp/src/parquet/arrow/writer.cc
```diff
@@ -403,17 +410,31 @@ class FileWriterImpl : public FileWriter {
...
+                          : arrow_properties_->executor();
+
+      return ::arrow::internal::OptionalParallelFor(
+          arrow_properties_->use_threads(), static_cast<int>(writers.size()),
+          [&](int i) {
+            auto ctx = arrow_properties_->use_threads() ? &column_write_context_.at(i)
+                                                        : &column_write_context_.back();
+            return writers[i]->Write(ctx);
+          },
+          executor);
```
cyb70289, line 437: This code is a bit tedious IMO. Probably the code can be much simpler if it is only called for use-thead case.

cpp/src/parquet/arrow/writer.cc
wgtmac: Then it is invalid to call `WriteColumnChunk`. Maybe I need to add a check to protect it.

cpp/src/parquet/arrow/writer.cc
wgtmac: If the optimization is aggressive, it may be true. I am also suspecting this is the root cause of test failure.

cpp/src/parquet/arrow/writer.cc
wgtmac: Agreed. Let me refactor it to share common logic and avoid unnecessary vector creation.

cpp/src/parquet/properties.h
wgtmac: No, the default value is nullptr and executor created by `::arrow::internal::GetCpuThreadPool()` will be used by default. If this is not a good idea, I can set the default value to `::arrow::internal::GetCpuThreadPool()` to make follow-up use much simpler.

wgtmac (commented):
Thanks for the review @cyb70289 @mapleFU

cpp/src/parquet/properties.h
cyb70289: When is it default nullified? I didn't see related code.

wgtmac: I have changed to return `::arrow::internal::GetCpuThreadPool()` if user does not provide a valid executor. Please check the related change in the `properties.cc`

wgtmac:
```
::arrow::internal::Executor* ArrowWriterProperties::executor() const {
  return executor_ != nullptr ? executor_ : ::arrow::internal::GetCpuThreadPool();
}
```

wgtmac:
**C++ / AMD64 Ubuntu 20.04 C++ ASAN UBSAN** [(link)](https://github.com/apache/arrow/actions/runs/3918495063/jobs/6698929302
) is failing with following error:

```
[90/890] Building CXX object CMakeFiles/substrait.dir/substrait_ep-generated/substrait/extensions/extensions.pb.cc.o
FAILED: CMakeFiles/substrait.dir/substrait_ep-generated/substrait/extensions/extensions.pb.cc.o 
/usr/bin/ccache /usr/lib/ccache/clang++-14  -DADDRESS_SANITIZER -DARROW_HAVE_RUNTIME_AVX2 -DARROW_HAVE_RUNTIME_AVX512 -DARROW_HAVE_RUNTIME_BMI2 -DARROW_HAVE_RUNTIME_SSE4_2 -DARROW_HAVE_SSE4_2 -DARROW_NO_DEPRECATED_API -DARROW_UBSAN -DARROW_WITH_RE2 -DARROW_WITH_UTF8PROC -Isubstrait_ep-generated -Iprotobuf_ep-install/include -Isrc -I/arrow/cpp/src -I/arrow/cpp/src/generated -Qunused-arguments -fcolor-diagnostics  -Wall -Wextra -Wdocumentation -Wshorten-64-to-32 -Wno-missing-braces -Wno-unused-parameter -Wno-constant-logical-operand -Wno-return-stack-address -Wno-unknown-warning-option -Wno-pass-failed -msse4.2  -fsanitize=address -DADDRESS_SANITIZER -fsanitize=undefined -fno-sanitize=alignment,vptr,function,float-divide-by-zero -fno-sanitize-recover=all -fsanitize-coverage=pc-table,inline-8bit-counters,edge,no-prune,trace-cmp,trace-div,trace-gep -fsanitize-blacklist=/arrow/cpp/build-support/sanitizer-disallowed-entries.txt -g -Werror -O0 -ggdb -fPIC   -fsanitize-coverage=pc-table,inline-8bit-counters,edge,no-prune,trace-cmp,trace-div,trace-gep -std=c++17 -Wno-error=shorten-64-to-32 -MD -MT CMakeFiles/substrait.dir/substrait_ep-generated/substrait/extensions/extensions.pb.cc.o -MF CMakeFiles/substrait.dir/substrait_ep-generated/substrait/extensions/extensions.pb.cc.o.d -o CMakeFiles/substrait.dir/substrait_ep-generated/substrait/extensions/extensions.pb.cc.o -c substrait_ep-generated/substrait/extensions/extensions.pb.cc
In file included from substrait_ep-generated/substrait/extensions/extensions.pb.cc:4:
In file included from substrait_ep-generated/substrait/extensions/extensions.pb.h:24:
In file included from protobuf_ep-install/include/google/protobuf/arena.h:52:
protobuf_ep-install/include/google/protobuf/arena_impl.h:45:10: fatal error: 'sanitizer/asan_interface.h' file not found
#include <sanitizer/asan_interface.h>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

@kou Do you have any idea?

kou:
It's not related to this pull request. I've opened a new issue for it: #33667

westonpace (commented):
This could, potentially, lead to a nested parallelism deadlock if multiple parquet files are being written at the same time.

However, the only place that could happen today is the dataset writer and since `use_threads` defaults to false I think we are ok for now.

The reason this happens is:

The thread that calls OptionalParallelFor will block waiting for the column writers.  If this thread is a user thread that is fine.  If this thread is a thread pool thread then that essentially becomes a "wasted thread" that can't be used in the pool.  If the number of files being written is equal to or greater than the number of threads in the thread pool then there is a potential that all threads become wasted threads and no work can be done.

The fix would be to add a `WriteRecordBatchAsync` method that calls `ParallelForAsync` and returns the future.  This can then be safely called in parallel, even by thread pool threads (assuming they don't block on that future but wrap it up into a higher level `AllComplete` call later).  The `WriteRecordBatch` method could then just return `WriteRecordBatchAsync(...).status()`.

For now, could you potentially add a comment that this must be false if the user is writing multiple files in parallel?

The comment could either go on the `use_threads` property or the `WriteRecordBatch` method.  My preference would be `WriteRecordBatch`.

lidavidm (approved):
Would you mind filing a ticket for the async version Weston suggested?

cpp/src/parquet/arrow/writer.cc
```diff
@@ -403,16 +409,31 @@ class FileWriterImpl : public FileWriter {
...
+          writers.emplace_back(std::move(writer));
+        } else {
+          RETURN_NOT_OK(writer->Write(&column_write_context_));
+        }
       }
+
+      if (arrow_properties_->use_threads()) {
+        DCHECK_EQ(parallel_column_write_contexts_.size(), writers.size());
+        RETURN_NOT_OK(::arrow::internal::OptionalParallelFor(
+            /*use_threads=*/true, static_cast<int>(writers.size()),
```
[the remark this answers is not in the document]
westonpace: Thanks, those comments look good!

wgtmac:
> Would you mind filing a ticket for the async version Weston suggested?

Thanks @lidavidm and @westonpace for the review! I have opened the [issue](https://github.com/apache/arrow/issues/33710) to track the progress.

cpp/src/parquet/arrow/writer.cc
```diff
@@ -403,16 +409,31 @@ class FileWriterImpl : public FileWriter {
     }
 
     auto WriteBatch = [&](int64_t offset, int64_t size) {
+      std::vector<std::unique_ptr<ArrowColumnWriterV2>> writers;
       int column_index_start = 0;
+
       for (int i = 0; i < batch.num_columns(); i++) {
-        ChunkedArray chunkedArray(batch.column(i));
+        ChunkedArray chunkedArray{batch.column(i)};
```
cyb70289, line 416: Nit: I know it's from existing code, but can you rename `chunkedArray` to `chunked_array`? Thanks.

wgtmac: Good catch! Fixed the name. Please check @cyb70289

cyb70289 (approved):

cyb70289:
Thanks @wgtmac !

PR #33658, opened 2023-01-13
GH-33638: [C++] Removing ExecPlan::Make deprecation warning

### LINKED ISSUE #33638: R package has pragmas that are supressing diagnostics

In compute-exec.cpp, we have:

https://github.com/apache/arrow/blob/3146588a0c2b016a757d0f3f86d152c814b734e9/r/src/compute-exec.cpp#L44-L51

...which is not allowed on CRAN. Ideally we need to find a way to not use any of these pragmas in the R code by fixing it to not use deprecated behaviour. A workaround may be to remove the deprecation by patching a C++ file when we do the vendor step (but hopefully we don't have to do this).

### Component(s)

R

westonpace:
See issue for rationale
* Closes: #33638


paleolimbot (approved):
Thank you!

Issue #15280, opened 2023-01-09
[C++] Move Acero out of libarrow


wjones127:
### Describe the enhancement requested

`libarrow_acero`? `libacero`?

The core libarrow is largely stable while Acero remains experimental. It would be nice to eventually let these libraries evolve somewhat independently, like arrow-rs and datafusion.

- [x] #34630

### Component(s)

C++

wjones127:
`libarrow_dataset` depends on Acero, and is tightly integrated with it. Should it be moved into `libacero`? or renamed `libacero_dataset`?

kou:
I prefer to `libacero` because Gandiva uses `libgandiva` not `libarrow_gandiva`.
ADBC also uses `libadbc_*` not `libarrow_adbc_*`.

jorisvandenbossche:
What part would we want to move to a separate library? 
Right now Acero lives in `arrow/compute/exec/`, right? Would we only move that, or the full of `arrow/compute` ? (xref ARROW-8891 - *"Split non-cast compute kernels into a separate shared library"*)

rok:
How usable are compute kernels now without Acero? How usable is Acero without kernels?

rok:
So then another option is to split further to `libkernels`, `libacero`. I don't have an opinion on the split though :).

paleolimbot:
I would love to see this split, although the ability to distribute it separately (e.g., so that a separate R or Python 'acero' package could exist) is something I hope this would be a step in the direction of.

wjones127:
> What part would we want to move to a separate library?
Right now Acero lives in arrow/compute/exec/, right? Would we only move that, or the full of arrow/compute ? (xref [ARROW-8891](https://issues.apache.org/jira/browse/ARROW-8891) - "Split non-cast compute kernels into a separate shared library")

Yes, just `arrow/compute/exec/`. Splitting compute functions is its own issue and much messier.

westonpace:
`libacero` will be easier than `libkernels` (there are still parts of the core lib that depend on compute kernels, I think) though both seem valid.

The only gray area I can see is group-by.  There is currently a standalone group-by implementation that runs outside of Acero, that is marked internal, that I am [trying to get rid of](https://github.com/apache/arrow/pull/14867).

My vote would be that "group-by" is a part of Acero.  Note, the `hash_` aggregate kernels themselves could remain as part of compute/kernels.  Aggregations would still be usable outside of Acero.

[...]

PR #33660, opened 2023-01-13
GH-33659: [Developer Tools] Add definition of Breaking Change and Critical Fix

### LINKED ISSUE #33659: [Developer Tools] Define Breaking Change and Critical Fix

Add them to PR template and suggest labels.

Follow up on: https://lists.apache.org/thread/s49bkj8ckgcjvjskhhz26xy1xscrswdp

### Component(s)

Developer Tools

wjones127:
* Closes: #33659

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -59,9 +59,8 @@ If there are user-facing changes then we may require documentation to be updated
 <!--
 If there are any breaking changes to public APIs, please uncomment the following line:
 -->
-<!-- **This PR includes breaking changes to public APIs.** -->
-
+* [ ] This PR includes breaking changes to public APIs.
 <!--
 If the changes fix either a security vulnerability or a bug that caused incorrect or invalid data to be produced. This is intended to mark changes that may affect users without their knowledge. For this reason, fixing bugs that cause errors or crashes don't count, since those are usually obvious.
 -->
-<!-- **This PR contains a "Critical Fix".** -->
+* [ ] This PR contains a "Critical Fix".
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -57,5 +57,11 @@ If there are user-facing changes then we may require documentation to be updated
 -->
 
 <!--
-If there are any breaking changes to public APIs, please add the `breaking-change` label.
+If there are any breaking changes to public APIs, please uncomment the following line:
 -->
+<!-- **This PR includes breaking changes to public APIs.** -->
+
+<!--
+If the changes fix either a security vulnerability or a bug that caused incorrect or invalid data to be produced. This is intended to mark changes that may affect users without their knowledge. For this reason, fixing bugs that cause errors or crashes don't count, since those are usually obvious.
+-->
+<!-- **This PR contains a "Critical Fix".** -->
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -57,10 +57,10 @@ If there are user-facing changes then we may require documentation to be updated
 -->
 
 <!--
-If there are any breaking changes to public APIs, please uncomment the following line:
+If there are any breaking changes to public APIs, please check the box below:
 -->
 * [ ] This PR includes breaking changes to public APIs.
 <!--
-If the changes fix either a security vulnerability or a bug that caused incorrect or invalid data to be produced. This is intended to mark changes that may affect users without their knowledge. For this reason, fixing bugs that cause errors or crashes don't count, since those are usually obvious.
+Check the box below if the changes fix either a security vulnerability or a bug that caused incorrect or invalid data to be produced. We use this to highlight fixes to issues that may affect users without their knowledge. For this reason, fixing bugs that cause errors or crashes don't count, since those are usually obvious.
 -->
 * [ ] This PR contains a "Critical Fix".
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -57,10 +57,11 @@ If there are user-facing changes then we may require documentation to be updated
 -->
 
 <!--
-If there are any breaking changes to public APIs, please check the box below:
+If there are any breaking changes to public APIs, please uncomment the line below and explain which changes are breaking.
 -->
-* [ ] This PR includes breaking changes to public APIs.
+<!-- **This PR includes breaking changes to public APIs.** -->
+
 <!--
-Check the box below if the changes fix either (a) a security vulnerability, (b) a bug that caused incorrect or invalid data to be produced, or (c) a bug that causes a crash (even when the API contract is upheld). We use this to highlight fixes to issues that may affect users without their knowledge. For this reason, fixing bugs that cause errors don't count, since those are usually obvious.
+Pleas uncomment the line below (and provide explanation) if the changes fix either (a) a security vulnerability, (b) a bug that caused incorrect or invalid data to be produced, or (c) a bug that causes a crash (even when the API contract is upheld). We use this to highlight fixes to issues that may affect users without their knowledge. For this reason, fixing bugs that cause errors don't count, since those are usually obvious.
 -->
-* [ ] This PR contains a "Critical Fix".
+<!-- **This PR contains a "Critical Fix".** -->
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -62,6 +62,6 @@ If there are any breaking changes to public APIs, please uncomment the line belo
 <!-- **This PR includes breaking changes to public APIs.** -->
 
 <!--
-Pleas uncomment the line below (and provide explanation) if the changes fix either (a) a security vulnerability, (b) a bug that caused incorrect or invalid data to be produced, or (c) a bug that causes a crash (even when the API contract is upheld). We use this to highlight fixes to issues that may affect users without their knowledge. For this reason, fixing bugs that cause errors don't count, since those are usually obvious.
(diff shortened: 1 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -57,5 +57,11 @@ If there are user-facing changes then we may require documentation to be updated
 -->
 
 <!--
-If there are any breaking changes to public APIs, please add the `breaking-change` label.
--->
\ No newline at end of file
+If there are any breaking changes to public APIs, please uncomment the following line:
```
assignUser, line 60: I have seen projects use markdown check boxes in pr templates for things like these and other |checklist| item like tests etc... might be a good use case here too?

assignUser:
Thanks I really like this! A needed step into the direction of properly versioning each component independently!

.github/pull_request_template.md
wjones127: I will look into that.

.github/pull_request_template.md
```diff
@@ -57,5 +57,11 @@ If there are user-facing changes then we may require documentation to be updated
...
+-->
+<!-- **This PR includes breaking changes to public APIs.** -->
+
+<!--
+If the changes fix either a security vulnerability or a bug that caused incorrect or invalid data to be produced. This is intended to mark changes that may affect users without their knowledge. For this reason, fixing bugs that cause errors or crashes don't count, since those are usually obvious.
+-->
```
westonpace, line 66: This is a little confusing to me.

If we were previously giving the wrong answer, and we fixed something to give a new answer, would I mark it as both breaking change and critical fix?

Also, hopefully most security fixes would go unnoticed by a user. I wouldn't expect them (by default) to change the output or the API.

Also, do we want to carve out an exception for experimental APIs?

wjones127: Hmm I guess I should clarify since Rok had the same question. The difference is "Breaking Change" is when we change our promise about how an API behaves, while "Critical Fix" changes the implementation to align with the existing promise. For example, if we have a function `add(x, y)` that promised to add two numbers, but we found that `add(2, 2)` returned `5`, fixing that to return `4` is a Critical Fix, not a breaking change. In the Arrow project, this typically shows up in file readers and writers, where we sometimes parse or serialize data incorrectly. Fixing those bugs don't really constitute breaking changes IMO.

Another way to put it is "Breaking Changes" are the reasons you might not want to upgrade until you have time to adapt your code, while "Critical Changes" highlight the risk of *not* upgrading.

wjones127: > Also, do we want to carve out an exception for experimental APIs?

And that's a good question on experimental APIs... I'm inclined to say it's still good to mark them as breaking changes. When we mark them as experimental, we are telling users to expect the API to change, but not that we won't warn them if they do 😄. That being said, in the future if we decide to do semantic versioning, we would probably treat breaking changes in experimental APIs differently from those in stable APIs.

wjones127: If we wanted, we could consider a separate tag for `Experimental API Change`.

westonpace: > Another way to put it is "Breaking Changes" are the reasons you might not want to upgrade until you have time to adapt your code, while "Critical Changes" highlight the risk of not upgrading.

This description is very helpful, let's include some of it.

> If we wanted, we could consider a separate tag for Experimental API Change.

No, I think your rationale is ok.  Even in our experimental APIs we usually avoid breaking changes if we can so it isn't like they occur all that often.

.github/pull_request_template.md
```diff
@@ -57,5 +57,10 @@ If there are user-facing changes then we may require documentation to be updated
...
-If there are any breaking changes to public APIs, please add the `breaking-change` label.
--->
\ No newline at end of file
+If there are any breaking changes to public APIs, please check the box below:
+-->
+* [ ] This PR includes breaking changes to public APIs.
+<!--
+Check the box below if the changes fix either a security vulnerability or a bug that caused incorrect or invalid data to be produced. We use this to highlight fixes to issues that may affect users without their knowledge. For this reason, fixing bugs that cause errors or crashes don't count, since those are usually obvious.
+-->
+* [ ] This PR contains a "Critical Fix".
```
jorisvandenbossche, line 66: Minor argument against using checkboxes for this: this adds a "todo" list to each PR, and then the GitHub UI will show this for example as "1 of 2 tasks" done in the pull request listing

(ideally github would also allow you use a "form" and not just a template, so this could be an actual checkbox that would add the label)

jorisvandenbossche: Also, if there are breaking changes to the public API, we should maybe ask to actually list / describe the exact change(s), instead of just checking the box? (and potentially give some rationale for making the change, although that could also refer to the issue)

docs/source/developers/reviewing.rst
```diff
@@ -255,3 +255,41 @@ Social aspects
...
+Breaking changes and critical fixes are separate: breaking changes alter the
+API contract, while critical fixes make the implementation align with the
+existing API contract. For example, fixing a bug that caused a Parquet reader
+to skip rows containing the number 42 is a critical fix but not a breaking change,
+since the row skipping wasn't behavior a reasonable user would rely on.
+
+These labels are used in the release to highlight changes that users ought to be
+aware of when they consider upgrading library versions. Breaking changes help
+identify reasons when users may wish to wait to upgrade until they have time to
+adapt their code, while critical fixes highlight the risk in *not* upgrading.
```
jorisvandenbossche, line 286: This is a nice description, very clear!

.github/pull_request_template.md
wjones127: > (ideally github would also allow you use a "form" and not just a template, so this could be an actual checkbox that would add the label)

Yeah I think to get close to that we could have our bot add the labels if these boxes are checked (or using some other indicator).

wjones127: Based on your feedback, I've reverted the instructions to "uncomment the line below" and also requested the creator provide explanation on which specific changes are breaking or critical fix.

docs/source/developers/reviewing.rst
```diff
@@ -255,3 +255,43 @@ Social aspects
...
+
+In addition, we use the following labels to indicate priority:
+
+* **Priority: Blocker**: Indicates the PR **must** be merged before the next
```
raulcd, line 292: At the moment for our release tasks we use the labels only on issues, not on PRs. Here we refer to labels on PRs and issues without distinction: `Indicates issues that are high priority.`. We also enforce the update of the milestone only on issues when merging a PR, not on PRs.
Right now the commit title on the merge script allows us to link to the PR because we use the following: `commit_title = f'{self.title} (#{self.number})'` which links both the issue `GH-XXXX` and the PR `(#YYYYY)` even though at the moment we don't use the PR on our release tasks, we could.
I suppose my concern is around the workflow changes and whether we want to force those labels only on issues, only on PRs, on both, what happens if a PR and an issue contain different types of labels, etcetera.
This might affect the discussion about removing the merge script or at least how we link issues with PRs.

wjones127: Thanks for pointing this out. I should change to say these labels should be applied to the issue, not the PR.

docs/source/developers/reviewing.rst
```diff
@@ -255,3 +255,43 @@ Social aspects
...
+* **Critical Fix**: The change fixes either: (a) a security vulnerability;
+  (b) a bug that causes incorrect or invalid data to be produced;
+  or (c) a bug that causes a crash (while the API contract is upheld).
+  This is intended to mark fixes to issues that may affect users without their
+  knowledge. For this reason, fixing bugs that cause errors don't count, since 
+  those bugs are usually obvious. Bugs that cause crashes are considered critical
+  because they are a possible vector of Denial-of-Service attacks.
+* **Breaking Change**: The change breaks backwards compatibility in a public API.
+  For changes in C++, this does not include changes that simply break ABI
+  compatibility, except for the few places where we do guarantee ABI
```
kou, line 275: Why do we not include ABI compatibility?
I think that we need to bump major version when we break API or ABI compatibility.
I thought that we use "Breaking Change" to determine whether we should bump major/minor version.

BTW, we will be able to check ABI compatibility automatically by CI. (Some projects do it.) So we may not check ABI compatibility manually.

kou (approved):
+1

docs/source/developers/reviewing.rst
wjones127: What I meant is that we don't guarantee that your code compiled against the headers of Arrow 10.0.0 can be linked against the Arrow 11.0.0 shared libraries. But we do try to keep it stable such that you don't need to change your code when you upgrade the dependency.

For example, IIUC changing a function signature from

```cpp
void MyFunction(int arg1);
```

To:

```cpp
void MyFunction(int arg1, int arg2 = 0);
```

Breaks ABI compatibility, since it alters the symbol for `MyFunction` in the shared library. But downstream projects won't have to change their code when they upgrade. So this paragraph means we wouldn't label this change as a "Breaking Change".

kou: Thanks. I see.

My understanding: We use "Breaking Change" to provide upgrade difficulty for users. We don't use "Breaking Change" to determine which version component (major/minor/patch) should be bumped.

docs/source/developers/reviewing.rst
jorisvandenbossche: Personally, as a reviewer, I would find it useful to have those labels on the PR as well ..

docs/source/developers/reviewing.rst
wjones127: For now, yes.

wjones127: > We don't use "Breaking Change" to determine which version component (major/minor/patch) should be bumped.

I suspect this won't be viable until we do https://github.com/apache/arrow/issues/15280, but worth discussing after.

PR #33664, opened 2023-01-14
GH-14981: [R] Forward compatibility with dplyr::join_by() 

### LINKED ISSUE #14981: [R] Forward compatibility for dplyr::join_by() output

`join_by()`, a function coming in dplyr 1.1.0, returns a `dplyr_join_by` object: https://github.com/tidyverse/dplyr/blob/main/R/join-by.R#L274-L284

We should make sure we can map that output to the existing equality join arguments in case someone starts using it.

### Component(s)

R

ianmcook:

Closes #14981

# Rationale for this change

dplyr 1.1.0 introduces a new function `join_by()` for specifying join conditions. This PR adds support for `join_by()` in dplyr joins on Arrow objects. The support is limited only to equality conditions. Code added in this PR throws an error if the user specifies inequality conditions or uses helper functions in `join_by()`.

https://www.tidyverse.org/blog/2022/11/dplyr-1-1-0-is-coming-soon/#join-improvements

- Code to handle `join_by()` in dplyr joins on Arrow objects with equality conditions
- Tests of handling of `join_by()`, which are skipped when the version of dplyr is less than `1.0.99.9000` which is the current version number of the development version of dplyr on GitHub which that become version `1.1.0` on CRAN.

# Are these changes tested?

Yes

Yes, the new dplyr syntax for specifying join conditions is supported, but use of this new syntax is optional. The old dplyr join syntax will continue to work. There are no breaking changes in this PR.
* Closes: #14981

### DIFF
--- a/r/R/dplyr-join.R
+++ b/r/R/dplyr-join.R
@@ -136,6 +136,17 @@ handle_join_by <- function(by, x, y) {
   if (is.null(by)) {
     return(set_names(intersect(names(x), names(y))))
   }
+  if (inherits(by, "dplyr_join_by")) {
+    if (!all(by$condition == "==" & by$filter == "none")) {
+      abort(
+        paste0(
+          "Inequality conditions and helper functions ",
+          "are not supported in `join_by()` expressions."
+        )
+      )
+    }
+    by <- set_names(by$y, by$x)
+  }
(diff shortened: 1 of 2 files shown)


paleolimbot (approved):
Thank you!

PR #33665, opened 2023-01-15
GH-33657: [C++] arrow-dataset.pc doesn't depend on parquet.pc without ARROW_PARQUET=ON

### LINKED ISSUE #33657: [c++] Does the dataset feature have a hard dependency on parquet?

C++: There is the option `ARROW_DATASET`. Does setting this to ON also require setting `ARROW_PARQUET` to ON?
Rationale: The file `arrow-dataset.pc` contains a hard dependency to `parquet`. 

PS: I am only a vcpkg maintainer and not an arrow user. 

### Component(s)

C++

kou:

Closes #33657

# Rationale for this change

If `arrow-dataset.pc` has `parquet` in `Requires` without `ARROW_PARQUET=ON`, the `arrow-dataset.pc` is invalid.

This PR includes `ARROW_PARQUET=ON` check to determine suitable `Requires` value.

# Are these changes tested?

Yes.

Yes. `arrow-dataset.pc` without `ARROW_PARQUET=ON` users can use `arrow-dataset.pc` with this.
* Closes: #33657


kou:
+1

PR #33674, opened 2023-01-15
GH-15256: [C++][Dataset] Add support for writing with Partitioning::Default()

### LINKED ISSUE #15256: [C++][Dataset] arrow::dataset::Partitioning::Default() can't be used for writing dataset

Because `arrow::dataset::DefaultPartitioning::Format()` isn't implemented:

https://github.com/apache/arrow/blob/2acc51a7d5304c3fc6a432f1c09946547ca91d74/cpp/src/arrow/dataset/partition.cc#L92-L95

It's required in `WriteBatch()`:

https://github.com/apache/arrow/blob/2acc51a7d5304c3fc6a432f1c09946547ca91d74/cpp/src/arrow/dataset/file_base.cc#L380-L381

Is it expected that we can't use `arrow::dataset::Partitioning::Default()` for writing dataset?
If it's expected, how about removing `arrow::dataset::Partitioning::Default()` because it's useless?
If it's not expected, how about implementing `arrow::dataset::DefaultPartitioning::Format()` like the following?

```diff
diff --git a/cpp/src/arrow/dataset/partition.cc b/cpp/src/arrow/dataset/partition.cc
index 46cdf9023c..13add35fb8 100644
--- a/cpp/src/arrow/dataset/partition.cc
+++ b/cpp/src/arrow/dataset/partition.cc
@@ -90,8 +90,7 @@ std::shared_ptr<Partitioning> Partitioning::Default() {
     }
 
     Result<PartitionPathFormat> Format(const compute::Expression& expr) const override {
-      return Status::NotImplemented("formatting paths from ", type_name(),
-                                    " Partitioning");
+      return PartitionPathFormat{"", ""};
     }
 
     Result<PartitionedBatches> Partition(
```

### Component(s)

C++

kou:

It writes all data into one directory.

### Are these changes tested?

Yes.

Yes.

* Closes: #15256

### DIFF
--- a/cpp/src/arrow/dataset/partition_test.cc
+++ b/cpp/src/arrow/dataset/partition_test.cc
@@ -189,42 +189,10 @@ TEST_F(TestPartitioning, Partition) {
                   expected_expressions);
 }
 
-TEST_F(TestPartitioning, DefaultPartitioningPartition) {
-  auto dataset_schema =
-      schema({field("a", int32()), field("b", utf8()), field("c", uint32())});
-
+TEST_F(TestPartitioning, DefaultPartitioningIsDirectoryPartitioning) {
   auto partitioning = Partitioning::Default();
-  std::string json = R"([{"a": 3,    "b": "x",  "c": 0},
-                         {"a": 3,    "b": "x",  "c": 1},
-                         {"a": 1,    "b": null, "c": 2},
-                         {"a": null, "b": null, "c": 3},
-                         {"a": null, "b": "z",  "c": 4},
-                         {"a": null, "b": null, "c": 5}
-                       ])";
-  std::vector<std::string> expected_batches = {json};
-  std::vector<compute::Expression> expected_expressions = {literal(true)};
-  AssertPartition(partitioning, dataset_schema, json, dataset_schema, expected_batches,
-                  expected_expressions);
-}
-
-TEST_F(TestPartitioning, DefaultPartitioningParse) {
-  partitioning_ = Partitioning::Default();
-
-  AssertParse("/hello", literal(true));
-  AssertParse("", literal(true));  // no segments to parse
-}
-
-TEST_F(TestPartitioning, DefaultPartitioningFormat) {
-  partitioning_ = Partitioning::Default();
-
-  written_schema_ = partitioning_->schema();
-
-  AssertFormat(and_(equal(field_ref("alpha"), literal(0)),
-                    equal(field_ref("beta"), literal("hello"))),
-               "");
-  AssertFormat(and_(equal(field_ref("alpha"), literal(0)), is_null(field_ref("beta"))),
-               "");
-  AssertFormat(literal(true), "");
+  ASSERT_EQ(partitioning->type_name(), "directory");
+  AssertSchemaEqual(partitioning->schema(), schema({}));
 }
 
 TEST_F(TestPartitioning, DirectoryPartitioning) {
--- a/cpp/src/arrow/dataset/partition_test.cc
+++ b/cpp/src/arrow/dataset/partition_test.cc
@@ -189,6 +189,44 @@ TEST_F(TestPartitioning, Partition) {
                   expected_expressions);
 }
 
+TEST_F(TestPartitioning, DefaultPartitioningPartition) {
+  auto dataset_schema =
+      schema({field("a", int32()), field("b", utf8()), field("c", uint32())});
+
+  auto partitioning = Partitioning::Default();
+  std::string json = R"([{"a": 3,    "b": "x",  "c": 0},
+                         {"a": 3,    "b": "x",  "c": 1},
+                         {"a": 1,    "b": null, "c": 2},
+                         {"a": null, "b": null, "c": 3},
+                         {"a": null, "b": "z",  "c": 4},
+                         {"a": null, "b": null, "c": 5}
+                       ])";
+  std::vector<std::string> expected_batches = {json};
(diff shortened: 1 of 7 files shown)


pitrou:
Shouldn't this be given a more descriptive name than "default"?

kou:
"flat"? "nothing"?

jorisvandenbossche:
"Flat" sounds good to me.

cc @westonpace

westonpace:
I'm fine with default. I think I'd prefer "none" over "flat".  "flat" implies to me that something is still happening.  E.g. there is still some kind of partitioning.

We currently have `HivePartitioning` and `DirectoryPartitioning` so how about `NoPartitioning`?

kou:
I'm OK with `NoPartitioning`.

jorisvandenbossche:
> I think I'd prefer "none" over "flat". "flat" implies to me that something is still happening. E.g. there is still some kind of partitioning.

There also _is_ still some kind of partitioning, I think? I.e. a single flat directory? I would interpret "No" partitioning as a single file.

westonpace:
@jorisvandenbossche you might be thinking of FilenamePartitioning (which I forgot to mention) which gives you:

```
x=7_chunk0.parquet
x=7_chunk1.parquet
x=10_chunk0.parquet
```

This partitioning is only going to split up files when there are too many rows. So, if you set `max_rows_per_file` to unlimited then you would get a single file for each write.  The output will be:

```
chunk0.parquet
chunk1.parquet
chunk2.parquet
```

...and there will be no meaningful information in the filenames.

jorisvandenbossche:
No, I was thinking about the latter. 
How I interpret it, with the datasets API, we _always_ create a partitioned dataset, since we always create a (possibly nested) directory of files (in contrast to the Parquet ``write_table`` API which writes single files). Even if the different files / parts have no meaningful additional information embedded in the name or path, I personally still consider that as partitioned (but again, that's my interpretation, I don't know if that's the common interpretation of "partitioned")

westonpace:
@pitrou can be tiebreaker then :).  I don't like `FlatPartitioning` (I would think this was equivalent to filename partitioning) and Joris is not a fan of `NoPartitioning` (since he would expect no files).  Maybe `ChunkPartitioning` (since files will still potentially be broken into chunks if needed)?

kou:
Can we use `FilenamePartitioning` with an empty schema as the default partitioning?

westonpace:
I suppose all partitioning schemes, given an empty schema, should behave exactly the same.  That might be a better solution.

For example, someone working with Spark will always want to use the hive partitioning scheme.  Sometimes there might not be any partitioning columns.  They still would think they are working with "the hive scheme with no columns".

I'm not sure how much this scenario is tested.

jorisvandenbossche:
> I'm not sure how much this scenario is tested.

From Python that is certainly tested, since if you don't pass any partitioning columns in `pyarrow.dataset.write_dataset`, we create a default partitioning object with an empty schema (and the default is DirectoryPartitioning). 

---

> Maybe `ChunkPartitioning` (since files will still potentially be broken into chunks if needed)?

The downside of that is that also for other schemes like HivePartitioning files _also_ get broken into chunks in addition to the hive-like directories, so that is not a distinguishing feature. 

Maybe the original "Default" partitioning is a decent name in the end, since "default" is ambiguous enough to avoid such conflicting interpretations of "flat" or "no" .. ;)

cpcloud:
@kou Thanks for the PR! This has been open for some months now without activity, so I'm going to close it out!

westonpace:
I didn't mean to reopen.  @kou can reopen if desired.  However, I do think it would be good to resolve this issue.

kou:
@westonpace OK! We need to find a consensus approach to resolve this.
How about returning a `DictionaryPartitioning` with an empty schema (like PyArrow) by `Partitioning::Default()` instead of a `DefaultPartitioning` defined internally?

kou:
OK. I'll do it.

kou:
@westonpace Could you review this?

CI failures are unrelated:
* R failures are caused by #34711
* MinGW failures will be fixed by #34768

cpp/src/arrow/dataset/partition_test.cc
```diff
@@ -189,6 +189,12 @@ TEST_F(TestPartitioning, Partition) {
                   expected_expressions);
 }
 
+TEST_F(TestPartitioning, DefaultPartitioningIsDirectoryPartitioning) {
+  auto partitioning = Partitioning::Default();
+  ASSERT_EQ(partitioning->type_name(), "directory");
+  AssertSchemaEqual(partitioning->schema(), schema({}));
+}
+
```
westonpace, line 197: Can you also add a test to verify directory partitioning works with an empty schema?

```

TEST_F(TestPartitioning, DirectoryPartitioningEmpty) {
  partitioning_ = std::make_shared<DirectoryPartitioning>(schema({}));
  written_schema_ = partitioning_->schema();

  // No partitioning info
  AssertParse("", literal(true));
  // Files can be in subdirectories
  AssertParse("/foo/", literal(true));
  // Partitioning info is discarded on write
  AssertFormat(equal(field_ref("alpha"), literal(7)), "");
}
```

westonpace (approved):
Thanks!  We might need https://github.com/apache/arrow/pull/34872 to make sure the tests run.

cpp/src/arrow/dataset/partition_test.cc
kou: Sure!
Done!

westonpace:
CI failures are unrelated.

PR #33684, opened 2023-01-16
GH-15171: [C++] Pass std::string_view by value

### LINKED ISSUE #15171: [C++] Pass std::string_view by value, not by const-ref

See https://quuxplusone.github.io/blog/2021/11/09/pass-string-view-by-value/

### Component(s)

C++

ucasfl:

Closes #15171. 

# Rationale for this change

Pass std::string_view by value, not by const-ref
# Are these changes tested?

No.

* Closes: #15171

### DIFF
--- a/cpp/src/arrow/array/builder_binary.h
+++ b/cpp/src/arrow/array/builder_binary.h
@@ -480,7 +480,7 @@ class ARROW_EXPORT FixedSizeBinaryBuilder : public ArrayBuilder {
     return Append(reinterpret_cast<const uint8_t*>(value));
   }
 
-  Status Append(const std::string_view& view) {
+  Status Append(std::string_view view) {
     ARROW_RETURN_NOT_OK(Reserve(1));
     UnsafeAppend(view);
     return Status::OK();
@@ -662,7 +662,7 @@ class ARROW_EXPORT ChunkedBinaryBuilder {
     return builder_->Append(value, length);
   }
 
-  Status Append(const std::string_view& value) {
+  Status Append(std::string_view value) {
     return Append(reinterpret_cast<const uint8_t*>(value.data()),
                   static_cast<int32_t>(value.size()));
   }
--- a/cpp/src/arrow/compute/exec/tpch_node_test.cc
+++ b/cpp/src/arrow/compute/exec/tpch_node_test.cc
@@ -98,9 +98,9 @@ void VerifyUniqueKey(std::unordered_set<int32_t>* seen, const Datum& d, int32_t 
   }
 }
 
-void VerifyStringAndNumber_Single(const std::string_view& row,
-                                  const std::string_view& prefix, const int64_t i,
-                                  const int32_t* nums, bool verify_padding) {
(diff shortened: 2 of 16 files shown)


cpp/CMakeLists.txt
```diff
@@ -49,6 +49,8 @@ endif()
 
 set(ARROW_VERSION "11.0.0-SNAPSHOT")
 
+set(CMAKE_EXPORT_COMPILE_COMMANDS 1) # Write compile_commands.json
```
pitrou, line 52: I don't think we want to enable this by default. You should pass this yourself to CMake using `-DCMAKE_EXPORT_COMPILE_COMMANDS=1`.

cpp/src/arrow/vendored/datetime/tz.cpp
```diff
@@ -3597,19 +3597,17 @@ tzdb::locate_zone(const std::string& tz_name) const
 {
```
pitrou, line 3597: This is a vendored file (as the file path suggests), so we shouldn't do such changes to it. Can you revert them?

pitrou (changes requested):

cpp/CMakeLists.txt
ucasfl: OK

cpp/src/arrow/vendored/datetime/tz.cpp
ucasfl: done.

pitrou (approved):

pitrou:
@ursabot please benchmark lang=C++

cyb70289:
Rebased to fix lint error.

cyb70289:
Looks Glib & Ruby CI error is not related? cc @kou  
https://github.com/apache/arrow/actions/runs/3935751955/jobs/6731706736#step:10:204

kou:
Not related. It's sometimes happen. I couldn't reproduce it on local environment...

cyb70289 (approved):

westonpace:
@ucasfl thanks for cleaning this up!

jgehrcke:
Some curiosity-driven feedback CC @ucasfl @westonpace @pitrou. 

I read https://quuxplusone.github.io/blog/2021/11/09/pass-string-view-by-value/ with curiosity and was wondering if this patch here resulted in a measurable performance difference somewhere on the user-facing surface of Arrow.

Before Christmas I build a benchmark that I called `dataset-serialize`. Among others, it measures the time for dataset CSV serialization, writing to memory (/dev/shm). It also runs on bare metal. Both in combination yields quite some stability.

When this commit landed, said benchmark got faster by (roughly) three percent:

image

The plot/data can be further explored here: https://conbench.ursa.dev/benchmarks/bc6d37ffe2614078814efb01402f64dd/

Of course this is just correlation (once this was merged, the benchmark got a tad faster) and not necessarily causation (for corroborating this I understand too little of the impact of this change). But yeah, it's probably fair to say that it's not unlikely that this change here was responsible for the perf change.

PR #33693, opened 2023-01-16
GH-33666: [R] Remove extraneous argument to semi_join

### LINKED ISSUE #33666: [R] Two tests are failing with to-be-released dplyr dependency

```
Error (test-dplyr-join.R:241): semi_join
<rlib_error_dots_nonempty/rlib_error_dots/rlang_error/error/condition>
Error in `semi_join(., to_join, by = "some_grouping", keep = TRUE)`: `...` must be empty.
x Problematic argument:
* keep = TRUE
Backtrace:
 1. arrow:::compare_dplyr_binding(...)
      at test-dplyr-join.R:241:2
 6. dplyr:::semi_join.data.frame(., to_join, by = "some_grouping", keep = TRUE)
```

(Looks like we just need to remove `keep = TRUE`, which was never an argument to `semi_join()` anyway?)

```
Error (test-dplyr-funcs-datetime.R:3641): with_tz() and force_tz() works
<dplyr:::mutate_error/rlang_error/error/condition>
Error in `mutate(., timestamps_with_tz_1 = force_tz(timestamps, "Europe/Brussels", 
    roll_dst = c("boundary", "post")))`: i In argument: `timestamps_with_tz_1 = force_tz(...)`.
Caused by error in `force_tz()`:
! unused argument (roll_dst = c("boundary", "post"))
Backtrace:
  1. arrow:::compare_dplyr_binding(...)
       at test-dplyr-funcs-datetime.R:3641:2
  6. dplyr:::mutate.data.frame(...)
  7. dplyr:::mutate_cols(.data, dplyr_quosures(...), by)
  9. dplyr:::mutate_col(dots[[i]], data, mask, new_columns)
 10. mask$eval_all_mutate(quo)
 11. dplyr (local) eval()
```

### Component(s)

R

thisisnic:
This PR removes the `keep` argument from the test for `semi_join()`, which are causing the unit tests to fail.  It also removes the argument `suffix` argument (which is not part of the dplyr function signature) from the function signature here.

Closes: #33666

### DIFF
--- a/r/R/dplyr-join.R
+++ b/r/R/dplyr-join.R
@@ -116,9 +116,8 @@ semi_join.arrow_dplyr_query <- function(x,
                                         y,
                                         by = NULL,
                                         copy = FALSE,
-                                        suffix = c(".x", ".y"),
                                         ...) {
-  do_join(x, y, by, copy, suffix, ..., join_type = "LEFT_SEMI")
+  do_join(x, y, by, copy, ..., join_type = "LEFT_SEMI")
 }
 semi_join.Dataset <- semi_join.ArrowTabular <- semi_join.RecordBatchReader <- semi_join.arrow_dplyr_query
 
--- a/r/R/dplyr-join.R
+++ b/r/R/dplyr-join.R
@@ -116,8 +116,9 @@ semi_join.arrow_dplyr_query <- function(x,
                                         y,
                                         by = NULL,
                                         copy = FALSE,
+                                        suffix = c(".x", ".y"),
                                         ...) {
-  do_join(x, y, by, copy, ..., join_type = "LEFT_SEMI")
+  do_join(x, y, by, copy, suffix, ..., join_type = "LEFT_SEMI")
 }
 semi_join.Dataset <- semi_join.ArrowTabular <- semi_join.RecordBatchReader <- semi_join.arrow_dplyr_query
 
--- a/r/tests/testthat/test-dplyr-join.R
+++ b/r/tests/testthat/test-dplyr-join.R
@@ -82,7 +82,7 @@ test_that("left_join with join_by", {
       left_join(
         to_join %>%
           rename(the_grouping = some_grouping),
-          join_by(some_grouping == the_grouping)
+        join_by(some_grouping == the_grouping)
       ) %>%
       collect(),
     left
@@ -244,7 +244,6 @@ test_that("semi_join", {
       collect(),
     left
   )
-
 })
 
 test_that("anti_join", {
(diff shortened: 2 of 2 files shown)


thisisnic:
I'm struggling to work out why the linter is failing this; it works fine for me locally(!)

raulcd:
> I'm struggling to work out why the linter is failing this; it works fine for me locally(!)

There was a PR merged (https://github.com/apache/arrow/pull/33620#issuecomment-1383702537) that broke it on master. I have merged the fix on master if you rebase should be fixed.

r/tests/testthat/test-dplyr-join.R
```diff
@@ -82,7 +82,7 @@ test_that("left_join with join_by", {
       left_join(
         to_join %>%
           rename(the_grouping = some_grouping),
-          join_by(some_grouping == the_grouping)
+        join_by(some_grouping == the_grouping)
```
paleolimbot, line 85: Is this change intentional?

r/R/dplyr-join.R
```diff
@@ -116,9 +116,8 @@ semi_join.arrow_dplyr_query <- function(x,
                                         y,
                                         by = NULL,
                                         copy = FALSE,
-                                        suffix = c(".x", ".y"),
                                         ...) {
-  do_join(x, y, by, copy, suffix, ..., join_type = "LEFT_SEMI")
+  do_join(x, y, by, copy, ..., join_type = "LEFT_SEMI")
```
paleolimbot, line 120: Are changes in this file necessary to make the tests pass? (Or do they have another purpose, like perhaps making this function more consistent with other arrow join implementations?)

paleolimbot (approved):
Tiny things! Thanks!

r/R/dplyr-join.R
thisisnic: It makes the API consistent with dplyr

r/tests/testthat/test-dplyr-join.R
thisisnic: It's the consequence of tunning the styler.

r/R/dplyr-join.R
paleolimbot: I'm ok with that change; however, I think it should be consistent for all of our join wrappers (anti, left, right, etc.). Currently they all define the `suffix` argument explicitly (and I think the default value of `c(".x", ".y")` is necessary and may result in an error if you try to `semi_join` overlapping columns?).

thisisnic: I've removed this change as the `anti_join()` signature is also different to dplyr's.  I don't quite understand what you mean; please can we discuss this in #33709?

PR #33694, opened 2023-01-16
MINOR: [C++][Parquet] Rephrase decimal annotation

wgtmac:

- Rename the function in the writer properties to `enable_short_decimal_stored_as_integer`.
- Rephrase the comment to explain the behavior in detail.

### DIFF
--- a/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
+++ b/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
@@ -3443,8 +3443,8 @@ TEST(ArrowReadWrite, Decimal256AsInt) {
   auto table = ::arrow::Table::Make(::arrow::schema({field("root", type)}), {array});
 
   parquet::WriterProperties::Builder builder;
-  // Enforce integer type to store short decimal type
-  auto writer_properties = builder.enable_short_decimal_stored_as_integer()->build();
+  // Allow small decimals to be stored as int32 or int64.
+  auto writer_properties = builder.enable_store_decimal_as_integer()->build();
   auto props_store_schema = ArrowWriterProperties::Builder().store_schema()->build();
 
   CheckConfiguredRoundtrip(table, table, writer_properties, props_store_schema);
@@ -4821,8 +4821,8 @@ class TestIntegerAnnotateDecimalTypeParquetIO : public TestParquetIO<TestType> {
     auto arrow_schema = ::arrow::schema({::arrow::field("a", values->type())});
 
     parquet::WriterProperties::Builder builder;
-    // Enforce integer type to store short decimal type
-    auto writer_properties = builder.enable_short_decimal_stored_as_integer()->build();
+    // Allow small decimals to be stored as int32 or int64.
+    auto writer_properties = builder.enable_store_decimal_as_integer()->build();
     std::shared_ptr<SchemaDescriptor> parquet_schema;
     ASSERT_OK_NO_THROW(ToParquetSchema(arrow_schema.get(), *writer_properties,
                                        *default_arrow_writer_properties(),
--- a/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
+++ b/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
@@ -3443,8 +3443,8 @@ TEST(ArrowReadWrite, Decimal256AsInt) {
   auto table = ::arrow::Table::Make(::arrow::schema({field("root", type)}), {array});
 
   parquet::WriterProperties::Builder builder;
-  // Enforce integer type to annotate decimal type
-  auto writer_properties = builder.enable_integer_annotate_decimal()->build();
+  // Enforce integer type to store short decimal type
+  auto writer_properties = builder.enable_short_decimal_stored_as_integer()->build();
   auto props_store_schema = ArrowWriterProperties::Builder().store_schema()->build();
 
   CheckConfiguredRoundtrip(table, table, writer_properties, props_store_schema);
@@ -4821,8 +4821,8 @@ class TestIntegerAnnotateDecimalTypeParquetIO : public TestParquetIO<TestType> {
     auto arrow_schema = ::arrow::schema({::arrow::field("a", values->type())});
 
     parquet::WriterProperties::Builder builder;
-    // Enforce integer type to annotate decimal type
-    auto writer_properties = builder.enable_integer_annotate_decimal()->build();
+    // Enforce integer type to store short decimal type
+    auto writer_properties = builder.enable_short_decimal_stored_as_integer()->build();
     std::shared_ptr<SchemaDescriptor> parquet_schema;
     ASSERT_OK_NO_THROW(ToParquetSchema(arrow_schema.get(), *writer_properties,
                                        *default_arrow_writer_properties(),
--- a/cpp/src/parquet/properties.h
+++ b/cpp/src/parquet/properties.h
@@ -456,7 +456,7 @@ class PARQUET_EXPORT WriterProperties {
     ///
     /// In Parquet, DECIMAL can be stored in any of the following physical types:
     /// - int32: for 1 <= precision <= 9.
-    /// - int64: for 1 <= precision <= 18; precision < 10 will produce a warning.
+    /// - int64: for 10 <= precision <= 18.
     /// - fixed_len_byte_array: precision is limited by the array size.
     ///   Length n can store <= floor(log_10(2^(8*n - 1) - 1)) base-10 digits.
     /// - binary: precision is unlimited. The minimum number of bytes to store
--- a/cpp/src/parquet/properties.h
+++ b/cpp/src/parquet/properties.h
@@ -459,8 +459,8 @@ class PARQUET_EXPORT WriterProperties {
     /// - int64: for 1 <= precision <= 18; precision < 10 will produce a warning.
     /// - fixed_len_byte_array: precision is limited by the array size.
     ///   Length n can store <= floor(log_10(2^(8*n - 1) - 1)) base-10 digits.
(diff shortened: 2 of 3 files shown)


wgtmac:
@pitrou What about this?

wjones127:
Breadcrumb: These changes are in response to comments:

 * https://github.com/apache/arrow/pull/15244#discussion_r1071045668
 * https://github.com/apache/arrow/pull/15244#discussion_r1071045128

cpp/src/parquet/properties.h
```diff
@@ -452,19 +452,39 @@ class PARQUET_EXPORT WriterProperties {
...
     }
 
-    /// Enable integer type to annotate decimal type as below:
-    ///   int32: 1 <= precision <= 9
-    ///   int64: 10 <= precision <= 18
-    /// Default disabled.
-    Builder* enable_integer_annotate_decimal() {
-      integer_annotate_decimal_ = true;
+    /// Enable decimal logical type with 1 <= precision <= 18 to be stored as
+    /// integer physical type.
```
wjones127, line 456:
```suggestion
    /// Allow decimals with precision <= 18 to be stored as integers.
```

cpp/src/parquet/properties.h
```diff
@@ -452,19 +452,39 @@ class PARQUET_EXPORT WriterProperties {
...
+    ///
+    /// According to the specs, DECIMAL can be used to annotate the following types:
```
wjones127, line 458:
```suggestion
    /// In Parquet, DECIMAL can be stored in any of the following physical types:
```

cpp/src/parquet/properties.h
```diff
@@ -452,19 +452,39 @@ class PARQUET_EXPORT WriterProperties {
...
+    /// - int32: for 1 <= precision <= 9.
+    /// - int64: for 1 <= precision <= 18; precision < 10 will produce a warning.
+    /// - fixed_len_byte_array: precision is limited by the array size.
+    ///   Length n can store <= floor(log_10(2^(8*n - 1) - 1)) base-10 digits.
+    /// - binary: precision is not limited, but is required. precision is not limited,
+    ///   but is required. The minimum number of bytes to store the unscaled value
+    ///   should be used.
```
wjones127, line 465:
```suggestion
    /// - binary: precision is unlimited. The minimum number of bytes to store the unscaled value
    ///   is used.
```

cpp/src/parquet/properties.h
```diff
@@ -452,19 +452,39 @@ class PARQUET_EXPORT WriterProperties {
...
-    ///   int64: 10 <= precision <= 18
-    /// Default disabled.
-    Builder* enable_integer_annotate_decimal() {
-      integer_annotate_decimal_ = true;
+    /// Enable decimal logical type with 1 <= precision <= 18 to be stored as
+    /// integer physical type.
+    ///
+    /// According to the specs, DECIMAL can be used to annotate the following types:
+    /// - int32: for 1 <= precision <= 9.
+    /// - int64: for 1 <= precision <= 18; precision < 10 will produce a warning.
```
wjones127, line 460: "precision < 10 will produce a warning." is this true of our implementation? Maybe we should skip saying that.

cpp/src/parquet/properties.h
```diff
@@ -452,19 +452,39 @@ class PARQUET_EXPORT WriterProperties {
...
+    /// By default, this is DISABLED and all decimal types annotate fixed_len_byte_array.
+    ///
+    /// When enabled, the C++ writer will use following physical types to store decimals:
+    /// - int32: for 1 <= precision <= 9.
+    /// - int64: for 10 <= precision <= 18.
+    /// - fixed_len_byte_array: for precision > 18.
+    ///
+    /// As a consequence, decimal columns stored in integer types are more compact
+    /// but in a risk that the parquet file may not be readable by previous Arrow C++
+    /// versions or other implementations.
```
wjones127, line 476:
```suggestion
    /// As a consequence, decimal columns stored in integer types are more compact.
```

I'm not sure it's worth saying that older versions don't support them. In Arrow C++, read support for these was added at the same time for the other physical types for decimals ([original pr](https://github.com/apache/parquet-cpp/pull/403)). And that was a long time ago, before version 0.10.0. I'm sure it was similarly well supported in parquet-mr.

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -3443,8 +3443,8 @@ TEST(ArrowReadWrite, Decimal256AsInt) {
   auto table = ::arrow::Table::Make(::arrow::schema({field("root", type)}), {array});
 
   parquet::WriterProperties::Builder builder;
-  // Enforce integer type to annotate decimal type
-  auto writer_properties = builder.enable_integer_annotate_decimal()->build();
+  // Enforce integer type to store short decimal type
```
wjones127, line 3446:
```suggestion
  // Allow small decimals to be stored as int32 or int64.
```

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -3443,8 +3443,8 @@ TEST(ArrowReadWrite, Decimal256AsInt) {
...
+  auto writer_properties = builder.enable_short_decimal_stored_as_integer()->build();
```
wjones127, line 3447:
```suggestion
  auto writer_properties = builder.enable_store_decimal_as_integer()->build();
```

cpp/src/parquet/properties.h
wjones127: We can omit `1 <= precision` since that's obvious.

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -4821,8 +4821,8 @@ class TestIntegerAnnotateDecimalTypeParquetIO : public TestParquetIO<TestType> {
     auto arrow_schema = ::arrow::schema({::arrow::field("a", values->type())});
 
     parquet::WriterProperties::Builder builder;
-    // Enforce integer type to annotate decimal type
-    auto writer_properties = builder.enable_integer_annotate_decimal()->build();
+    // Enforce integer type to store short decimal type
```
wjones127, line 4824:
```suggestion
    // Allow small decimals to be stored as int32 or int64.
```

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
wjones127: I think this slightly shorter name is sufficient.

wjones127 (changes requested):
Thanks for doing this. I've added several suggestions :)

cpp/src/parquet/properties.h
wjones127: FYI @pitrou

cpp/src/parquet/properties.h
wgtmac: This part of text is copied from the [specs](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#decimal). Should I leave them to be consistent?

cpp/src/parquet/properties.h
wgtmac: This is copied from the specs. In our impl no warning is produced because we have fixed physical types for decimals and guarantee that precision < 10 will not use int64 type.

cpp/src/parquet/properties.h
wgtmac: I'd prefer to keep 1 <= precision <= 18 to be consistent with the specs. Otherwise we may infer that precision <= 0 (including negative precisions) also apply this.

wjones127: Do negative precision decimals exist?

cpp/src/parquet/properties.h
wjones127: I'd rather write about what our function does than what is in the spec. The spec is documentation for us (the developers), but this documentation is for our users.

cpp/src/parquet/properties.h
wgtmac: I don't think negative precision decimals exist, but keeping `1 <= precision` doesn't hurt.

cpp/src/parquet/properties.h
wgtmac: That makes sense. I have changed the comment accordingly.

cpp/src/parquet/properties.h
wgtmac: I have checked that both apache/parquet-mr and apache/impala support reading this long ago. It seems that the native parquet reader in the velox does not handle this yet. I'm not familiar with other implementations but there may be some do not support it.

IMHO, we cannot take care of all other implementations. There is no risk to the parquet-cpp itself because the reader support is already in and it is well documented by the parquet specs. So I have removed the comment about the risk. @wjones127 @pitrou

wjones127: Thanks for confirming that.

cpp/src/parquet/properties.h
wjones127: Then perhaps we should change to this? So that we document our libraries behavior, and not just what the spec says?

```suggestion
    /// - int64: for 10 <= precision <= 18.
```

wjones127: Once this is done, I think this is ready to be merged :)

wgtmac: Fixed. Please check again.

wjones127 (approved):

wjones127:
> I found that this feature is already included in the maint-11.0.0 branch. Should I cherry-pick this once merged?

I'm afraid it might be too late for that, since the release vote has already passed. So this will have to be a minor breaking change in the 12.0.0 release.

PR #33703, opened 2023-01-16
GH-14748:[C++] Modify some comments about shared_ptr's ownership in parquet-cpp.

### LINKED ISSUE #14748: [C++][Parquet] column_reader.HasNext() throws an exception "Access violation executing location"
I'm trying to use C++ API to read a parquet file generated by a C# application based on PraquetSharp. As the title said, I encountered a problem of loading the UUID column. The schema, column description and sample code are shown below respectively.
image
image
image
Is the method I called incorrect?

P.S I can successfully load the FixedLenByteArray Column from the file generated by C++ API with "logical_type=NONE".

### Component(s)

C++

zzzzwj:
### Rationale for this change

The documentations of some functions that return a shared_ptr are fuzzy. So I add some comments ahead these functions to make them clear.

Add some comments on some functions which return a shared_ptr.

### Are these changes tested?

No, only comments/documentations are affected.

Yes, the update on documentation will affect user's coding.

* Closes: #14748

### DIFF
--- a/cpp/src/parquet/file_reader.h
+++ b/cpp/src/parquet/file_reader.h
@@ -53,7 +53,8 @@ class PARQUET_EXPORT RowGroupReader {
   const RowGroupMetaData* metadata() const;
 
   // Construct a ColumnReader for the indicated row group-relative
-  // column. Ownership is shared with the RowGroupReader.
+  // column. Ownership is shared with the RowGroupReader. This object
+  // must not outlive the RowGroupReader.
   std::shared_ptr<ColumnReader> Column(int i);
 
   // Construct a ColumnReader, trying to enable exposed encoding.


cpp/src/parquet/encryption/internal_file_decryptor.h
```diff
@@ -76,12 +76,17 @@ class InternalFileDecryptor {
...
   std::shared_ptr<Decryptor> GetFooterDecryptor();
+  /// This object must not outlive the InternalFileDecryptor.
   std::shared_ptr<Decryptor> GetFooterDecryptorForColumnMeta(const std::string& aad = "");
+  /// This object must not outlive the InternalFileDecryptor.
   std::shared_ptr<Decryptor> GetFooterDecryptorForColumnData(const std::string& aad = "");
+  /// This object must not outlive the InternalFileDecryptor.
   std::shared_ptr<Decryptor> GetColumnMetaDecryptor(
       const std::string& column_path, const std::string& column_key_metadata,
       const std::string& aad = "");
+      /// This object must not outlive the InternalFileDecryptor.
```
mapleFU, line 89: fmt?

cpp/src/parquet/column_reader.h
```diff
@@ -151,7 +151,7 @@ class PARQUET_EXPORT PageReader {
   }
 
   // @returns: shared_ptr<Page>(nullptr) on EOS, std::shared_ptr<Page>
-  // containing new Page otherwise
+  // containing new Page otherwise. This object must not outlive the PageReader.
```
mapleFU, line 154: Why, seems that `shared_ptr<Page>` can outlives PageReader, isn't it just a shared buffer?

wgtmac: I would say it is fine if the buffer (as well as the page object) outlives the PageReader. But its lifecycle is bound to the memory pool which allocates the buffer.

cpp/src/parquet/arrow/test_util.h
```diff
@@ -463,12 +463,16 @@ Status MakeEmptyListsArray(int64_t size, std::shared_ptr<Array>* out_array) {
   return Status::OK();
 }
 
+// Please make sure that the return value should have an object assigned explicitly
+// if you want to use the pointer contained.
```
pitrou, line 467: I don't understand what this comment means, nor why you are adding it in multiple places. Would you like to explain?

zzzzwj: Hi pitrou. I think I misunderstood the meaning of [emkornfield](https://github.com/emkornfield). According to https://github.com/apache/arrow/issues/14748#issuecomment-1333336429, I thought that when a function returns a shared_ptr containing a new object (the pointer contained is allocated in the function), then it need to add the comment "sth. must not outlive some object". This function is static, which belongs to no object. So, I put my word to replace that comment. And you can refer to #14748, I open that issue because I directly used the bare pointer without explicitly assignment to an object, and it causes the access violation.

zzzzwj: Does that make sense? And is it necessary to make this PR? I think it was my problem that triggers the access violation, but not the docs.

cpp/src/parquet/column_reader.h
zzzzwj: Sorry for the late response. This comment may be incorrect. According to https://github.com/apache/arrow/issues/14748#issuecomment-1333336429, I think [emkornfield](https://github.com/emkornfield)'s meaning is that, we need to make sure that the object "Column(i)" must outlive the "Column(i).get()", right? If so, this comment should be "shared_ptr\<Page\>" must outlive its pointer contained (Page *).
I'm not that familiar with C++. I don't know if I understand correctly or not, if not, please let me know. Thank you.

cpp/src/parquet/encryption/internal_file_decryptor.h
zzzzwj: Sorry, I didn't notice that comment. Will fix in next commit if has.

cpp/src/parquet/arrow/test_util.h
emkornfield: sorry for chiming, in there were at least a few places that seemed to be incorrect on specifying the the independence of the shared_ptr, I think my intent was really only places in column_reader.h.

zzzzwj: Oh, really sorry, it looks like I truly misunderstood your meaning. I thought that you wanted me to verify all the ownership of functions that return a shared_ptr. I'll have a careful check and then make a change on this PR.

emkornfield (approved):

mapleFU:
@emkornfield @wjones127 @zzzzwj What's the status of this patch? Should we rebase master, retrigger CI and merge it?

amol-:
> @emkornfield @wjones127 @zzzzwj What's the status of this patch? Should we rebase master, retrigger CI and merge it?

Follow up on this, according to https://github.com/apache/arrow/pull/33703#discussion_r1072218248 it seems my understanding is that the comment in the PR is in fact incorrect and we probably want to close this PR. Is my understanding correct?

PR #33712, opened 2023-01-17
GH-15139: [C++] Improve bzip2 static library path detection for arrow.pc

### LINKED ISSUE #15139: [C++] arrow.pc is missing dependencies with Windows static builds

I have been having to manually edit `arrow.pc` from a vcpkg installation of a static build of arrow (using triplet `x64-windows-static-md`) for a few reasons. #14869 was a start, but now for the tricky (to me) one - missing dependencies. It doesn't seem to be a vcpkg issue, as non-vcpkg builds also generate "unconsumable" pkg-config files, so I'm posting here. Or maybe it goes all the way up to cmake, I don't know (I'm a cmake avoider...).

The following is using arrow master as of a0d16306229fc08f9dc64361f9459806e02b5932.

For example, using VS2022 (17.4.3) (and cmake 3.24.202208181-MSVC_2 from there) configured as:

```
cmake -G Ninja -DVCPKG_TARGET_TRIPLET=x64-windows-static-md -DARROW_WITH_BROTLI=ON -DARROW_WITH_BZ2=ON 
  -DARROW_WITH_LZ4=ON -DARROW_WITH_SNAPPY=ON -DARROW_WITH_ZLIB=ON -DARROW_WITH_ZSTD=ON 
  -DARROW_BUILD_TESTS=OFF -DCMAKE_INSTALL_PREFIX=/path/to/somewhere -DARROW_DEPENDENCY_SOURCE=VCPKG 
  -DARROW_BUILD_STATIC=ON -DARROW_BUILD_SHARED=OFF -DARROW_BUILD_TESTS=OFF 
  -DARROW_DEPENDENCY_USE_SHARED=OFF -DARROW_WITH_BACKTRACE=OFF 
  -DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreadedDebugDLL -DCMAKE_BUILD_TYPE=Debug ../..
```

when built and installed generates `arrow.pc`:

```
...
Requires:
Requires.private:
Libs: -L${libdir} -larrow /path/to/arrow.git/cpp/vcpkg_installed/x64-windows-static-md/lib/snappy.lib optimized;/path/to/arrow.git/cpp/vcpkg_installed/x64-windows-static-md/lib/bz2.lib;debug;/path/to/arrow.git/cpp/vcpkg_installed/x64-windows-static-md/debug/lib/bz2d.lib
Libs.private:
Cflags: -I${includedir} -DARROW_STATIC
Cflags.private:
```

This is essentially the same as an install via vcpkg. There are a few problems here:

- Most of the compression deps are missing (libbrotlidec libbrotlienc zlib liblz4 libzstd).The cmake invocation generates the following:

  ```
  -- pkg-config package for libbrotlidec for static link isn't found
  -- pkg-config package for libbrotlienc for static link isn't found
  -- pkg-config package for zlib for static link isn't found
  ...
  ```

  so not surprising, I suppose! pc files are there in the `vcpkg_installed/x64-windows-static-md/debug/lib/pkgconfig` dir, however (and the libs themselves a level up).

- `Libs` has `bz2` and `snappy` without `-l` and with a file extension and path prefix. I don't think this is ideal, as without the `-l` the file may be treated as a plain object file rather than import library causing issues with certain toolchains. Ultimately this could be worked around in meson, maybe, (or libpkgconf I think it uses) but...
- `bzip2` should ideally be a `Requires` too - since there are separate release and debug pkgconfig dirs they will cope with the file naming difference.
- And `snappy` too, but its .pc file is also broken (missing completely here; `Libs: -L${libdir} -l`, i.e. not `-lsnappy`, with a vcpkg install of arrow).
  I haven't investigated that at all yet, so keep in `Libs`, but without a path.

I.e. a usable `arrow.pc` would look something like:

```
...
Requires: bzip2 libzstd liblz4 zlib libbrotlidec libbrotlienc libbrotlicommon
Libs: -L${libdir} -larrow -lsnappy
Cflags: -I"${includedir}" -DARROW_STATIC
```

Now I'm sure there are a gazillion of other cases to cope with where doing this would be wrong, but at least in the "get everything via vcpkg" case this `arrow.pc` works for me (and similarly editied `arrow.pc` from a vcpkg install of arrow).

As an aside, the same set up of options (except no `CMAKE_MSVC_RUNTIME_LIBRARY` and `x64-linux` triplet) on Fedora 37 (cmake 3.25.1) generates:

```
...
Requires: snappy libbrotlidec libbrotlienc zlib liblz4 libzstd
Requires.private:
Libs: -L${libdir} -larrow optimized;/path/to/arrow.git/cpp/vcpkg_installed/x64-linux/lib/libbz2.a;debug;/path/to/arrow.git/cpp/vcpkg_installed/x64-linux/debug/lib/libbz2d.a -larrow_bundled_dependencies
Libs.private:
Cflags: -I${includedir} -DARROW_STATIC
Cflags.private:
```

This is more reasonable, though I personally think `bzip2` is handled wrong (for the same reasons as above), even if it "works". The `debug`/`optimized` stuff is still problematic but vcpkg are adding support for that there (https://github.com/microsoft/vcpkg/pull/23898), so will start to work at some point - but obviously potentially not if consuming the pc file via another route.

I have spent some time hacking `ThirdpartyToolchain.cmake` to see if I could make the pc files more sane, but unsuccessfully. The best I could do was getting bzip2 "better" with:

```
if(ARROW_WITH_BZ2)
  resolve_dependency(BZip2
    PC_PACKAGE_NAMES
    bzip2)

  # resolve_dependency(BZip2)

  # if(${BZip2_SOURCE} STREQUAL "SYSTEM")
  #   string(APPEND ARROW_PC_LIBS_PRIVATE " ${BZIP2_LIBRARIES}")
  # endif()

  # if(NOT TARGET BZip2::BZip2)
  #   add_library(BZip2::BZip2 UNKNOWN IMPORTED)
  #   set_target_properties(BZip2::BZip2
  #                         PROPERTIES IMPORTED_LOCATION "${BZIP2_LIBRARIES}"
  #                                    INTERFACE_INCLUDE_DIRECTORIES "${BZIP2_INCLUDE_DIR}")
  # endif()
endif()
```

but what that breaks I've no idea for other cases. As I say, not a cmake person and would like to keep it that way 😀 Happy to try any suggestions, though.

Thanks!

### Component(s)

C++

kou:
### Rationale for this change

If bzip2 is installed for debug build and release build on Windows, find_package(BZip2)
returns both static library paths of them. 

We should use one of detected library paths in arrow.pc.

The bzip2's official build system doesn't provide bzip2.pc but vcpkg provides bzip2.pc. We can use it instead of detecting bzip2's static library path by ourselves.

### Are these changes tested?

Yes. We test these changes in #15139 manually.

No.
* Closes: #15139


kou:
+1

PR #33713, opened 2023-01-17
MINOR: [Documentation] Remove trailing spaces in CODEOWNERS file

cyb70289:


cyb70289:
A bit surprised github doesn't highlight trailing spaces for a new file. Even gerrit has such a feature.

assignUser (approved):
Ah good catch! (though this is a cosmetic rather than a functional change as `Trailing spaces are ignored unless they are quoted with backslash ` ^=^)

cyb70289:
> though this is a cosmetic rather than a functional change

Indeed. Just to remove the annoying red blocks in `git show`.

PR #33716, opened 2023-01-17
WIP: DO NOT MERGE: Apache Arrow Flight SQL adapter for PostgreSQL plan

kou:
This is a document to discuss this project.

See
https://github.com/kou/arrow/blob/flight-sql-postgresql/flight-sql-postgresql.md for rendered version.

### DIFF
--- a/flight-sql-postgresql.md
+++ b/flight-sql-postgresql.md
@@ -1,0 +1,102 @@
+# Apache Arrow Flight SQL adapter for PostgreSQL
+
+MySQL has a plugin API to open a new socket, that is used for
+implementing MySQL's X protocol, but PostgreSQL doesn't provide such
+API. So we need to implement it with background worker process API.
+
+https://www.postgresql.org/docs/current/bgworker.html
+
+We can access one database by one user per background worker process
+with SPI (Server Programming Interface).
+
+See also: https://www.postgresql.org/docs/current/spi.html
+
+We can also use libpg to connect local PostgreSQL but it has
+performance penalty because it uses the PostgreSQL protocol to connect
+to local PostgreSQL.  (Serializations/deserializations are happen.)
+
+If we use SPI, we need to create one background worker process per
+Apache Arrow Flight SQL connection. We will be able to use background
+worker process' "shared memory" feature to communicate between a
+Apache Arrow Flight SQL server process and a process that uses SPI.
+
+```mermaid
+sequenceDiagram
+    participant C as Client
+    participant M as PostgreSQL(master)
+    participant D as PostgreSQL(bgworker/Flight SQL dispatcher)
+    participant W as PostgreSQL(bgworker/Flight SQL worker)
+    M->>D: Spawn
+    note right of D: Listen gRPC socket
+    C->>D: Connect with Flight SQL protocol
+    D->>W: Spawn
+    C->>D: Send a query
+    D->>W: Pass the given query via shared memory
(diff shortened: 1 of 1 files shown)


alamb:
Thank you for starting this discussion @kou  -- we are interested in this feature at InfluxData and we may have some more detail to share soon.

jacobmarble:
I'm also curious about use cases. Deployed [PostgreSQL FE/BE](https://www.postgresql.org/docs/15/protocol.html) clients number in the zillions, so there would seem to be more interest in a proxy enabling those applications to use FlightSQL services.

kou:
@alamb @jacobmarble Thanks for your comments!

> It might help to start this document with an expected usecase.

It makes sense. I'll add it.

> I originally thought it was to allow clients that spoke the postgres (front-end / back-end protocol) protocol to connect to a FlightSQL server
>
> But now it is primarily to allow clients that speak FlightSQL to speak directly to a posgres server

Ah, sorry. Your understanding is correct. I didn't think the use case. I think that we cover this use case by ADBC. (cc: @lidavidm)

Is PostgreSQL protocol important in your use case? ADBC isn't suitable?

BTW, I think that the approach you mentioned has serialize/deserialize overhead that Apache Arrow project wants to eliminate.

lidavidm:
Yes, the ADBC driver wraps libpq and should let you work with other databases that use the PostgreSQL wire protocol, with the caveat that it has to convert the data (of course). 

Both tools serve complementary purposes: as you mention, there are many projects besides PostgreSQL that use its wire protocol. The ADBC driver provides access to them. At the same time, PostgreSQL and its derivatives are quite popular. The hope is that by pushing the conversion onto the server and using columnar data/Arrow for transport and at the edge, we can achieve better performance overall for clients that want Arrow data. 

I suppose you have a third goal: using the PostgreSQL protocol/drivers/clients as the lingua franca, placing it sort of in between the roles of Flight SQL and JDBC/ODBC? I don't think anyone's actively working towards that.

bkmgit:
What is needed for authentication? Could pluggable authentication modules be used?  How often could authentication be required - in particular are there performance requirements?

kou:
> What is needed for authentication?

Integration with user information in PostgreSQL. We should not require users to configure additional authentication related things.

> Could pluggable authentication modules be used?

I think that we shouldn't do it. We should only use information in PostgreSQL. 

> How often could authentication be required - in particular are there performance requirements?

Always. All connections must be associated with an user.

alamb:
> Ok. And those clients use the wire protocol directly(ish), so they can't take advantage of the JDBC or ODBC Flight SQL drivers, presumably?

That is correct. In the ideal future world, it will be FlightSQL (via ADBC perhaps or native support) all the way down. Until we get there it is important to work with the existing ecosystem

> ADBC may not help there, unless they're using DBAPI/SQLAlchemy in Python, where the ADBC bindings implement DBAPI and work(ish) with the rest of the stack.

👍

kou:
I've added use cases:

* https://github.com/kou/arrow/blob/flight-sql-postgresql/flight-sql-postgresql.md#use-cases
* https://github.com/apache/arrow/pull/33716/commits/3ed670e9cc5e8d88af969963fe0ae7dc28212e15

> * ETL
> * Interactive/exploratory data analysis

kou:
I've created https://github.com/apache/arrow-flight-sql-postgresql .

PR #33718, opened 2023-01-17
GH-33717: [Go] Flight SQL Server handle StreamChunk errors

### LINKED ISSUE #33717: [Go] FlightSQL server elides errors in StreamChunks

When sending a stream of results as a response to a  `DoGetStatement` (or indeed any other `DoGet` request). Any error returned over the `StreamChunk` channel will be silently dropped. The expected behaviour is for the error to propogate to the gRPC client.

This is occurring because when the `DoGet` handler detects that the `StreamChunk` contains an error it returns the contents of the `err` value, which will always be `nil` if that code path is being followed (https://github.com/apache/arrow/blob/master/go/arrow/flight/flightsql/server.go#L635).

### Component(s)

Go

mhilton:
### Rationale for this change

Errors returned in the `StreamChunk` channel of the Flight SQL server's `DoGet` method are currently being dropped rather than returned to gRPC client. This is a fix for that problem.

Fix the bug detailed above and add tesing for the fixed code path.

### Are these changes tested?

Yes, the majority of the PR add appropriate test cases.

Only the fixed bug.

* Closes: #33717

### DIFF
--- a/go/arrow/flight/flightsql/server.go
+++ b/go/arrow/flight/flightsql/server.go
@@ -632,7 +632,7 @@ func (f *flightSqlServer) DoGet(request *flight.Ticket, stream flight.FlightServ
 
 	for chunk := range cc {
 		if chunk.Err != nil {
-			return err
+			return chunk.Err
 		}
 
 		wr.SetFlightDescriptor(chunk.Desc)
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM, I'll merge it after all the CI pass.

Thanks much for the fix @mhilton!

PR #33728, opened 2023-01-17
GH-33726: [CI][Go] Set host name in Go benchmarks

### LINKED ISSUE #33726: Set consistent host name in Go benchmarks in CI

Currently Go benchmarks are running in CI, but because the particular runner nodes vary and we do not set a consistent hostname, we can't see the history of the benchmarks because the hostnames keep changing, e.g.

- fv-az570-835 https://conbench.ursa.dev/benchmarks/6948d6fcaa08438d9e3a31a05fb7a62a/
- fv-az361-674 https://conbench.ursa.dev/benchmarks/aeb05a213dac45e59f62dc2f6d9e888d/
- Mac-1673971649551.local https://conbench.ursa.dev/benchmarks/68559e0a74a34327a130f9dc719e7778/
- Mac-1672779927260.local https://conbench.ursa.dev/benchmarks/e05538be04f749f9b3d0f5299b1aeca9/

The solution for this is to set an environment variable called `CONBENCH_MACHINE_INFO_NAME` to a consistent value so [this code](https://github.com/conbench/conbench/blob/main/benchadapt/python/benchadapt/_machine_info.py#L161) will pick it up and use it instead. We're running on two types of runners, so we'll need to insert the env var in https://github.com/apache/arrow/blob/master/.github/workflows/go.yml in both the envs on [L94-98](https://github.com/apache/arrow/blob/master/.github/workflows/go.yml#L94-L98) and [L264-268](https://github.com/apache/arrow/blob/master/.github/workflows/go.yml#L264-L268). We can hardcode values in those two locations with names for the types of runners they are, e.g. something like `amd64-debian-11` and `amd64-macos-11`, respectively. Some of our other host names, for reference: 

- ec2-m5-4xlarge-us-east-2
- arm64-t4g-linux-compute
- ursa-i9-9960x

Long-term, we plan to move these benchmarks out of Arrow's CI and together with the rest in [voltrondata-labs/arrow-benchmarks-ci](https://github.com/voltrondata-labs/arrow-benchmarks-ci), but there's work to do before we're ready for that, and in the mean time, cleaning up our naming will let us see the history we're generating for Go benchmarks.

### Component(s)

Continuous Integration, Go

alistaire47:
Closes #33726; see issue for comprehensive explanation. cc @zeroshade 

### Rationale for this change

Go benchmarks are sending host names from runners that vary slightly, preventing us from seeing the full history of the benchmarks on Conbench.

Sets host name in Go benchmarks via environment variable so benchmark history will be available on Conbench.

### Are these changes tested?

Yes, for the benchadapt code that will pick it up: https://github.com/conbench/conbench/blob/main/benchadapt/python/tests/test_result.py#L107-L114

No
* Closes: #33726

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -96,6 +96,7 @@ jobs:
           CONBENCH_EMAIL: ${{ secrets.CONBENCH_EMAIL }}
           CONBENCH_PASSWORD: ${{ secrets.CONBENCH_PASS }}
           CONBENCH_REF: ${{ github.ref_name }}
+          CONBENCH_MACHINE_INFO_NAME: amd64-debian-11
         run: |
           pip install benchadapt@git+https://github.com/conbench/conbench.git@main#subdirectory=benchadapt/python
           python ci/scripts/go_bench_adapt.py
@@ -266,6 +267,7 @@ jobs:
           CONBENCH_EMAIL: ${{ secrets.CONBENCH_EMAIL }}
           CONBENCH_PASSWORD: ${{ secrets.CONBENCH_PASS }}
           CONBENCH_REF: ${{ github.ref_name }}
+          CONBENCH_MACHINE_INFO_NAME: amd64-macos-11
         run: |
           pip install benchadapt@git+https://github.com/conbench/conbench.git@main#subdirectory=benchadapt/python
           python ci/scripts/go_bench_adapt.py


zeroshade (approved):
LGTM

assignUser (approved):
+1

PR #33731, opened 2023-01-17
GH-15231: [C++][Benchmarking] Add new memory pool metrics and track in benchmarks

### LINKED ISSUE #15231: [Benchmarking][C++] Track memory usage in C++ microbenchmarks

Google Bench supports providing a `MemoryManager` class to its `RegisterMemoryManager()` method. Then each benchmark will report the following in the JSON output:

 * `allocs_per_iter`
 * `max_bytes_used`
 * `total_allocated_bytes` (if available)
 * `net_heap_growth` (if available)

Perhaps we can setup a `arrow::MemoryPool` that is a `MemoryManager`?

See:
https://github.com/google/benchmark/blob/main/docs/user_guide.md#memory-usage
https://github.com/google/benchmark/blob/62edc4fb00e1aeab86cc69c70eafffb17219d047/src/json_reporter.cc#L291-L305

Not sure, but I think once it's in the output in the JSON then Conbench will have it saved. ([It seems to save the whole JSON output](https://github.com/voltrondata-labs/benchmarks/blob/e8a8d415af589744a5a455a16701f244a8a7cd63/benchmarks/cpp_micro_benchmarks.py#L134).) And then it's a matter of processing and rendering it in Conbench. @jonkeane do I understand correctly?

### Component(s)

Benchmarking, C++

wjones127:
### Rationale for this change

We would like to track memory usage in many of our benchmarks. In order to isolate memory metrics per benchmark, we pass in a new `ProxyMemoryPool` to each benchmark, and collect the metrics from that at the conclusion of each benchmark.

This only works for benchmarks where there are no allocations that bypass the passed pool and instead use the `default_memory_pool()`. Unfortunately most do right now, so this is only enabled for two benchmarks: builders and CSV converter.

 * `MemoryPool` now tracks both `num_allocations` and `total_bytes_allocated`.
 * There is now a class `BenchmarkMemoryTracker` which can be instantiated in benchmark files to get access to the `ProxyMemoryPool` that takes measurements and saves them to Google benchmark.
 * Wired up memory tracking to `arrow-csv-converter-benchmark` and `arrow-builder-benchmark`

### Are these changes tested?

* [x] Add tests

* [x] Validate performance impact of new metrics in memory pools.

No breaking changes.
* Closes: #15231


wjones127:
@ursabot please benchmark lang=C++

rok:
Would there be need to expose [jemalloc stats](https://github.com/apache/arrow/blob/1d9366f19e4b9846b33cc0c7bd7941cb5f482d74/cpp/src/arrow/memory_pool_test.cc#L190-L199) as well?

wjones127:
Local benchmarks confirm this doesn't seem to have a meaningful affect on performance:

```
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Non-regressions: (20)
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
                                                   benchmark      baseline     contender  change %                                                                                                                                                                                                    counters
     AllocateDeallocate<SystemAlloc>/size:16777216/real_time    264.865114    270.309814     2.056   {'family_index': 1, 'per_family_instance_index': 3, 'run_name': 'AllocateDeallocate<SystemAlloc>/size:16777216/real_time', 'repetitions': 10, 'repetition_index': 5, 'threads': 1, 'iterations': 2646218}
                              TouchArea/size:65536/real_time   2073.088409   2102.201688     1.404                             {'family_index': 0, 'per_family_instance_index': 1, 'run_name': 'TouchArea/size:65536/real_time', 'repetitions': 10, 'repetition_index': 8, 'threads': 1, 'iterations': 338503}
    AllocateTouchDeallocate<Mimalloc>/size:1048576/real_time  58004.604073  58547.859656     0.937    {'family_index': 4, 'per_family_instance_index': 2, 'run_name': 'AllocateTouchDeallocate<Mimalloc>/size:1048576/real_time', 'repetitions': 10, 'repetition_index': 9, 'threads': 1, 'iterations': 12050}
                            TouchArea/size:1048576/real_time  33798.339093  34108.716406     0.918                            {'family_index': 0, 'per_family_instance_index': 2, 'run_name': 'TouchArea/size:1048576/real_time', 'repetitions': 10, 'repetition_index': 6, 'threads': 1, 'iterations': 20679}
                           TouchArea/size:16777216/real_time 555616.013949 559728.605189     0.740                            {'family_index': 0, 'per_family_instance_index': 3, 'run_name': 'TouchArea/size:16777216/real_time', 'repetitions': 10, 'repetition_index': 8, 'threads': 1, 'iterations': 1238}
    AllocateTouchDeallocate<SystemAlloc>/size:4096/real_time    205.279964    206.720969     0.702  {'family_index': 2, 'per_family_instance_index': 0, 'run_name': 'AllocateTouchDeallocate<SystemAlloc>/size:4096/real_time', 'repetitions': 10, 'repetition_index': 2, 'threads': 1, 'iterations': 3397743}
      AllocateDeallocate<SystemAlloc>/size:1048576/real_time     86.631662     87.188996     0.643    {'family_index': 1, 'per_family_instance_index': 2, 'run_name': 'AllocateDeallocate<SystemAlloc>/size:1048576/real_time', 'repetitions': 10, 'repetition_index': 2, 'threads': 1, 'iterations': 8012878}
        AllocateDeallocate<SystemAlloc>/size:65536/real_time     86.715660     87.233966     0.598      {'family_index': 1, 'per_family_instance_index': 1, 'run_name': 'AllocateDeallocate<SystemAlloc>/size:65536/real_time', 'repetitions': 10, 'repetition_index': 4, 'threads': 1, 'iterations': 8064265}
 AllocateTouchDeallocate<SystemAlloc>/size:1048576/real_time  33744.497321  33941.250271     0.583 {'family_index': 2, 'per_family_instance_index': 2, 'run_name': 'AllocateTouchDeallocate<SystemAlloc>/size:1048576/real_time', 'repetitions': 10, 'repetition_index': 5, 'threads': 1, 'iterations': 20702}
                               TouchArea/size:4096/real_time    105.445068    106.045668     0.570                             {'family_index': 0, 'per_family_instance_index': 0, 'run_name': 'TouchArea/size:4096/real_time', 'repetitions': 10, 'repetition_index': 8, 'threads': 1, 'iterations': 5587137}
            AllocateDeallocate<Mimalloc>/size:4096/real_time    615.564121    618.911726     0.544          {'family_index': 3, 'per_family_instance_index': 0, 'run_name': 'AllocateDeallocate<Mimalloc>/size:4096/real_time', 'repetitions': 10, 'repetition_index': 2, 'threads': 1, 'iterations': 1133612}
   AllocateTouchDeallocate<SystemAlloc>/size:65536/real_time   2183.956326   2194.576623     0.486  {'family_index': 2, 'per_family_instance_index': 1, 'run_name': 'AllocateTouchDeallocate<SystemAlloc>/size:65536/real_time', 'repetitions': 10, 'repetition_index': 7, 'threads': 1, 'iterations': 321459}
       AllocateTouchDeallocate<Mimalloc>/size:4096/real_time    738.431923    741.599917     0.429      {'family_index': 4, 'per_family_instance_index': 0, 'run_name': 'AllocateTouchDeallocate<Mimalloc>/size:4096/real_time', 'repetitions': 10, 'repetition_index': 9, 'threads': 1, 'iterations': 945778}
        AllocateDeallocate<Mimalloc>/size:16777216/real_time 541207.905798 542746.952466     0.284         {'family_index': 3, 'per_family_instance_index': 3, 'run_name': 'AllocateDeallocate<Mimalloc>/size:16777216/real_time', 'repetitions': 10, 'repetition_index': 7, 'threads': 1, 'iterations': 1220}
         AllocateDeallocate<SystemAlloc>/size:4096/real_time     81.856769     82.027025     0.208       {'family_index': 1, 'per_family_instance_index': 0, 'run_name': 'AllocateDeallocate<SystemAlloc>/size:4096/real_time', 'repetitions': 10, 'repetition_index': 5, 'threads': 1, 'iterations': 8543875}
           AllocateDeallocate<Mimalloc>/size:65536/real_time   2680.987560   2681.028080     0.002          {'family_index': 3, 'per_family_instance_index': 1, 'run_name': 'AllocateDeallocate<Mimalloc>/size:65536/real_time', 'repetitions': 10, 'repetition_index': 0, 'threads': 1, 'iterations': 261797}
         AllocateDeallocate<Mimalloc>/size:1048576/real_time  35094.429406  35048.055175    -0.132         {'family_index': 3, 'per_family_instance_index': 2, 'run_name': 'AllocateDeallocate<Mimalloc>/size:1048576/real_time', 'repetitions': 10, 'repetition_index': 5, 'threads': 1, 'iterations': 19992}
   AllocateTouchDeallocate<Mimalloc>/size:16777216/real_time 950131.925658 948438.976006    -0.178     {'family_index': 4, 'per_family_instance_index': 3, 'run_name': 'AllocateTouchDeallocate<Mimalloc>/size:16777216/real_time', 'repetitions': 10, 'repetition_index': 4, 'threads': 1, 'iterations': 725}
      AllocateTouchDeallocate<Mimalloc>/size:65536/real_time   4142.543854   4133.527899    -0.218     {'family_index': 4, 'per_family_instance_index': 1, 'run_name': 'AllocateTouchDeallocate<Mimalloc>/size:65536/real_time', 'repetitions': 10, 'repetition_index': 2, 'threads': 1, 'iterations': 170360}
AllocateTouchDeallocate<SystemAlloc>/size:16777216/real_time 562788.074673 556769.522773    -1.069 {'family_index': 2, 'per_family_instance_index': 3, 'run_name': 'AllocateTouchDeallocate<SystemAlloc>/size:16777216/real_time', 'repetitions': 10, 'repetition_index': 0, 'threads': 1, 'iterations': 1247}
```

wjones127:
> Would there be need to expose [jemalloc stats](https://github.com/apache/arrow/blob/1d9366f19e4b9846b33cc0c7bd7941cb5f482d74/cpp/src/arrow/memory_pool_test.cc#L190-L199) as well?

@rok you mean expose in the benchmark output? We have a limited number of fields there.

rok:
> @rok you mean expose in the benchmark output? We have a limited number of fields there.

If we'd want it we'd probably want a separate benchmark altogether. It would probably give different results as jemalloc does a bit of guessing about how much memory it should release etc.

wjones127:
A few points I would like feedback on:

 * I added `total_bytes_allocated` and `num_allocations`, since they are metrics available and seem like they would be useful. For example, it might help remind of us places where we aren't calling Reserve on array builders when we should be. Does that seem sensible?
 * In order to be able to use these metrics for individual benchmarks, we needed to expose a `ResetStatistics()` method. However, this changes the semantics of the statistics, since they are no longer guaranteed to represent the global state since the beginning of the program. That's fine as long as no one was relying on them for an accurate picture of memory use at the time, but I do worry it could be a foot-gun.
    - An alternative would be to keep a separate optional struct for "local stats", and allow that to be reset while the global one is left alone. And in most cases it could be left blank. Does that seem preferable?
 * I added memory tracking only on benchmarks that I felt would benefit from it. For example, I excluded flight because AFAIK many of the allocations happen in gRPC outside of the memory pool, so this method of measurement wouldn't be particularly useful there.

lidavidm:
We already have ProxyMemoryPool, right? Ideally, we'd make use of that (and fix any places that are hardcoding a memory pool).

wjones127:
> We already have ProxyMemoryPool, right? Ideally, we'd make use of that (and fix any places that are hardcoding a memory pool).

That's a good idea. I'll need to look at how hard it is to identify hardcoded memory pools. (I wonder if we can add a lint for that?).

westonpace:
> I added total_bytes_allocated and num_allocations, since they are metrics available and seem like they would be useful. For example, it might help remind of us places where we aren't calling Reserve on array builders when we should be. Does that seem sensible?

Yes, it seems like it could be useful in some cases.  Atomic increments aren't free but we don't allocate in critical sections so I think it's fine.

> In order to be able to use these metrics for individual benchmarks, we needed to expose a ResetStatistics() method. However, this changes the semantics of the statistics, since they are no longer guaranteed to represent the global state since the beginning of the program. That's fine as long as no one was relying on them for an accurate picture of memory use at the time, but I do worry it could be a foot-gun. 

As long as no one calls this it should be fine.  It looks like we're only calling this in the benchmarks, I think it's ok then.  Maybe don't expose it in python/r or maybe you could make it private and make the benchmark utilities a friend.

> I added memory tracking only on benchmarks that I felt would benefit from it. For example, I excluded flight because AFAIK many of the allocations happen in gRPC outside of the memory pool, so this method of measurement wouldn't be particularly useful there. Do the ones I chose seem reasonable?

No strong opinions.  Seems like a generally harmless addition.  Pity we can't get malloc statistics (maybe we could with jemalloc? but that can be for a future PR).

westonpace (approved):

rok:
> No strong opinions. Seems like a generally harmless addition. Pity we can't get malloc statistics (maybe we could with jemalloc? but that can be for a future PR).

@westonpace If you need jemalloc stats outside of benchmarking they are exposed [here](https://github.com/apache/arrow/blob/1d9366f19e4b9846b33cc0c7bd7941cb5f482d74/cpp/src/arrow/memory_pool_test.cc#L190-L199).

wjones127:
> Pity we can't get malloc statistics (maybe we could with jemalloc? but that can be for a future PR).

On one hand that would capture allocations not in the memory pool, which is nice. On the other hand, it means we are looking at the memory pool's requests for allocations, which might make the metrics less useful, since they are the allocations we have less direct control over and may even be non-deterministic (especially other benchmarks can become noisy neighbors).

wjones127:
I actually want to see if the ProxyMemoryPool approach is viable and cleaner, so I am going to return this to a draft state for now.

rok:
> On one hand that would capture allocations not in the memory pool, which is nice. On the other hand, it means we are looking at the memory pool's requests for allocations, which might make the metrics less useful, since they are the allocations we have less direct control over and may even be non-deterministic (especially other benchmarks can become noisy neighbors).

Indeed, that would make this proposal more of a "unit benchmark" and jemalloc alternative an "integration benchmark". I do think the current proposal is more useful for developers.

wjones127:
> Indeed, that would make this proposal more of a "unit benchmark" and jemalloc alternative an "integration benchmark". I do think the current proposal is more useful for developers.

Yeah perhaps the jemalloc approach would be appropriate for the macro benchmarks?

wjones127:
Experimenting here for now: https://github.com/wjones127/arrow/pull/1/files

So far, I like using `ProxyPool` as it avoids the `ResetStatistics` conundrum. But need to see how feasible it is to pass down the memory pool to everything. I did find a clever way to detect allocations we are missing though :)

wjones127:
I have good news and bad news.

Bad news is that most high-level functionality (CSV, Parquet, Acero) has at least some memory pool allocations that just go directly to the default memory pool. So I'm not able to add memory benchmarks to them using this method yet. The only two benchmarks that we can measure today are the builders and the CSV converter. 😢 

Good news is the by adding the `num_allocations` metric on memory pools, we can now write tests that assert all allocations happened on the passed memory pool. Basically just assert that `ProxyMemoryPool.num_allocations()` is equal to `default_memory_pool()->num_allocations()` minus the existing allocation count when the proxy pool was initialized. 🥳 

I propose that we merge this, and in follow-up PRs we iteratively fix the memory pool handling in features and enable the memory benchmarks as we go. Of course, we should also add tests as described above to make sure we don't get regressions in the memory pool handling.

westonpace:
Seems like a good plan.  If nothing else this exercise is worth it just to help detect places where we are improperly using the default memory pool.

wjones127:
@ursabot please benchmark lang=C++

cpp/src/arrow/util/benchmark_util.h
```diff
@@ -21,7 +21,10 @@
 
 #include "benchmark/benchmark.h"
 
+#include "arrow/memory_pool.h"
+#include "arrow/type_fwd.h"
 #include "arrow/util/cpu_info.h"
+#include "arrow/util/logging.h"  // IWYU pragma: keep
```
westonpace, line 27: We normally try to keep `logging.h` out of public headers.  Though I don't really know if `benchmark_util.h` is a public header or not :shrug: 

Why is everything defined inline here instead of in an cc file?

cpp/src/arrow/util/benchmark_util.h
```diff
@@ -136,4 +139,62 @@ struct RegressionArgs {
...
+
+/// \brief Track memory pool allocations in benchmarks.
+///
+/// Instantiate as a global variable to register the hooks into Google Benchmark
+/// to collect memory metrics. Before each benchmark, a new ProxyMemoryPool is
+/// created. It can then be accessed with memory_pool(). Once the benchmark is
+/// complete, the hook will record the maximum memory used, the total bytes
+/// allocated, and the total number of allocations. If no allocations were seen,
+/// (for example, if you forgot to pass down the memory pool), then these metrics
+/// will not be saved.
```
westonpace, line 190: How does this work with multi-threaded benchmarks?  Is the memory manager invoked per thread in some way?  Or one single memory manager shared across all threads?

I suspect the latter but it might be good to clarify here for future readers.

westonpace (approved):

cpp/src/arrow/util/benchmark_util.h
wjones127: I will put the implementation in a CC file.

cpp/src/arrow/util/benchmark_util.h
westonpace: To be clear, I am not interested in running multiple different benchmarks at the same time.  Rather, I am curious about running a single benchmark across multiple threads (https://github.com/google/benchmark/blob/main/docs/user_guide.md#multithreaded-benchmarks)  For example, we have some of these in `expression_benchmark.cc`.

cpp/src/arrow/util/benchmark_util.h
wjones127: On second thought, this seems like more trouble than it's worth.

cpp/src/arrow/util/benchmark_util.h
wjones127: Got it. I made a comment about both of those.

wjones127:
Follow-up issue created here: https://github.com/apache/arrow/issues/34058

wjones127:
@ursabot please benchmark lang=C++

PR #33732, opened 2023-01-17
GH-33209: [C++] Support for reading JSON Datasets

### LINKED ISSUE #33209: [C++] Bind the JSON RecordBatchReader to Dataset
**Reporter**: [Alessandro Molina](https://issues.apache.org/jira/browse/ARROW-18005) / @amol-
**Assignee**: [Ben Harkins](https://issues.apache.org/jira/browse/ARROW-18005) / @benibus

<sub>**Note**: *This issue was originally created as [ARROW-18005](https://issues.apache.org/jira/browse/ARROW-18005). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

benibus:
This adds initial support the JSON file format to the Dataset library. Since there's currently no public API for writing JSON files, this only deals with the reader-side facilities.
* Closes: #33209

### DIFF
--- a/cpp/src/arrow/dataset/file_json.cc
+++ b/cpp/src/arrow/dataset/file_json.cc
@@ -17,15 +17,29 @@
 
 #include "arrow/dataset/file_json.h"
 
+#include <algorithm>
 #include <unordered_set>
+#include <vector>
 
+#include "arrow/compute/exec.h"
+#include "arrow/compute/exec/expression.h"
 #include "arrow/dataset/dataset_internal.h"
+#include "arrow/dataset/scanner.h"
+#include "arrow/dataset/type_fwd.h"
 #include "arrow/io/buffered.h"
+#include "arrow/io/interfaces.h"
+#include "arrow/io/type_fwd.h"
 #include "arrow/json/chunker.h"
 #include "arrow/json/parser.h"
 #include "arrow/json/reader.h"
+#include "arrow/record_batch.h"
+#include "arrow/type.h"
+#include "arrow/util/async_generator.h"
+#include "arrow/util/bit_util.h"
 #include "arrow/util/checked_cast.h"
+#include "arrow/util/delimiting.h"
 #include "arrow/util/logging.h"
+#include "arrow/util/thread_pool.h"
 
 namespace arrow {
 
@@ -149,7 +163,7 @@ FieldRef ToUniversalRef(const FieldPath& path, const Schema& schema) {
   for (auto it = path.begin(); it != path.end(); ++it) {
     DCHECK_LT(*it, static_cast<int>(fields->size()));
     const auto& child_field = *(*fields)[*it];
-    refs.push_back(FieldRef(child_field.name()));
+    refs.emplace_back(child_field.name());
     if (it + 1 != path.end()) {
       auto&& child_type = checked_cast<const StructType&>(*child_field.type());
       fields = &child_type.fields();
--- a/cpp/src/arrow/dataset/file_json.cc
+++ b/cpp/src/arrow/dataset/file_json.cc
@@ -127,6 +127,14 @@ class JsonFragmentScanner : public FragmentScanner {
   int num_scanned_ = 0;
 };
 
+// Return the same parse options, but disable any options that could interfere with
+// fragment inspection
+json::ParseOptions GetInitialParseOptions(json::ParseOptions options) {
+  options.explicit_schema = nullptr;
+  options.unexpected_field_behavior = json::UnexpectedFieldBehavior::InferType;
+  return options;
+}
+
 Result<std::shared_ptr<StructType>> ParseToStructType(
     std::string_view data, const json::ParseOptions& parse_options, MemoryPool* pool) {
   auto full_buffer = std::make_shared<Buffer>(data);
@@ -258,14 +266,9 @@ Result<Future<ReaderPtr>> DoOpenReader(
   struct State {
     State(const JsonFragmentScanOptions& json_options,
           const std::shared_ptr<ScanOptions>& scan_options)
-        : parse_options(json_options.parse_options),
+        : parse_options(GetInitialParseOptions(json_options.parse_options)),
           read_options(json_options.read_options),
-          scan_options(scan_options) {
-      // We selectively ignore some user options, primarily those that will influence the
-      // output schema.
-      parse_options.explicit_schema = nullptr;
-      parse_options.unexpected_field_behavior = json::UnexpectedFieldBehavior::InferType;
-    }
+          scan_options(scan_options) {}
     json::ParseOptions parse_options;
     json::ReadOptions read_options;
     std::shared_ptr<const ScanOptions> scan_options;
@@ -339,9 +342,10 @@ Result<std::shared_ptr<InspectedFragment>> DoInspectFragment(
 
   ARROW_ASSIGN_OR_RAISE(auto first_block,
                         stream->Peek(format_options.read_options.block_size));
-  ARROW_ASSIGN_OR_RAISE(
-      auto struct_type,
-      ParseToStructType(first_block, format_options.parse_options, pool));
+
+  auto parse_options = GetInitialParseOptions(format_options.parse_options);
+  ARROW_ASSIGN_OR_RAISE(auto struct_type,
+                        ParseToStructType(first_block, parse_options, pool));
 
   std::vector<std::string> column_names;
   column_names.reserve(struct_type->num_fields());
--- a/cpp/src/arrow/dataset/file_json.cc
+++ b/cpp/src/arrow/dataset/file_json.cc
@@ -1,0 +1,291 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#include "arrow/dataset/file_json.h"
+#include "arrow/dataset/dataset_internal.h"
+#include "arrow/io/buffered.h"
+#include "arrow/json/chunker.h"
+#include "arrow/json/parser.h"
+#include "arrow/json/reader.h"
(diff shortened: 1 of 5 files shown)


benibus:
I attempted to implement the newer `ScanNode` methods. However, I ran into some issues that have prevented feature parity with the legacy API. Primarily, the methods that use `ScanOptions` are able to prevent loading top-level fields if none of its children are expected to be materialized. However, I haven't figured out a good way to do this using `FragmentScanRequest` since it only provides `FieldPath`s (relative to the inspected fragment).

I suppose there's a possibility that the tests in `FileFormatScanNodeMixin` are expecting too much of the format implementation, though. From what I can tell, CSV is the only other format that implements the new API and even then it doesn't support nested fields - so the relevant tests in that fixture haven't been used anywhere before.

benibus:
For reference, [here](https://github.com/benibus/arrow/blob/cb56130e6d311bd0d178a398ae114aec58fdcf9b/cpp/src/arrow/dataset/test_util_internal.h#L1345) is where that test would fail (in `FileFormatScanNodeMixin::TestScanProjectedNested`), whereas the equivalent test in `FileFormatScanMixin` currently passes.

cpp/src/arrow/dataset/file_json.cc
```diff
@@ -0,0 +1,414 @@
...
+  RETURN_NOT_OK(parser->Finish(&parsed));
+
+  return checked_pointer_cast<StructType>(parsed->type());
+}
+
+Result<std::shared_ptr<Schema>> ParseToSchema(std::string_view data,
+                                              const json::ParseOptions& parse_options,
+                                              MemoryPool* pool) {
+  ARROW_ASSIGN_OR_RAISE(auto type, ParseToStructType(data, parse_options, pool));
+  return ::arrow::schema(type->fields());
```
westonpace, line 139: Any particular need for `::arrow::`?

cpp/src/arrow/dataset/file_json.cc
```diff
@@ -0,0 +1,414 @@
...
+ private:
+  ReaderPtr reader_;
+  int64_t block_size_;
+  int num_batches_;
+  int num_scanned_ = 0;
+};
+
+Result<std::shared_ptr<StructType>> ParseToStructType(
+    std::string_view data, const json::ParseOptions& parse_options, MemoryPool* pool) {
+  if (!pool) pool = default_memory_pool();
```
westonpace, line 119: This is just an observation but it seems like this is something file formats probably shouldn't have to worry about.  It would be better to normalize the memory pool in the scanner I think.

That being said, I'm not surprised if we don't.  But maybe a potential for a follow-up PR.

cpp/src/arrow/dataset/file_json.cc
```diff
@@ -0,0 +1,414 @@
...
+
+// Make a new schema consisting only of the top-level fields in the dataset schema that:
+//  (a) Have children that require materialization
+//  (b) Have children present in `physical_schema`
+//
+// The resulting schema can be used in reader instantiation to ignore unused fields. Note
+// that `physical_schema` is only of structural importance and its data types are ignored
+// when constructing the final schema.
+Result<std::shared_ptr<Schema>> GetPartialSchema(const ScanOptions& scan_options,
+                                                 const Schema& physical_schema) {
```
westonpace, line 187: This "shouldn't" be needed in the new scanner.  I don't remember if it's needed in the old path or not.  If it is then we need to keep it unfortunately.  We can re-examine later when adding support for the new path.

cpp/src/arrow/dataset/file_json.cc
```diff
@@ -0,0 +1,414 @@
...
+      // We selectively ignore some user options, primarily those that will influence the
+      // output schema.
+      parse_options.explicit_schema = nullptr;
+      parse_options.unexpected_field_behavior = json::UnexpectedFieldBehavior::InferType;
+      // We don't wish to contend with the scanner's CPU threads if multiple fragments are
+      // being scanned in parallel, so fragment-level parallelism gets conditionally
+      // disabled.
+      if (this->scan_options && this->scan_options->use_threads) {
+        read_options.use_threads = false;
+      }
```
westonpace, line 262: We've done this in the past to avoid nested deadlock.  That shouldn't be a problem if we use the async paths (though I don't know what the async path for the JSON reader looks like).

If there are only 1 or 2 files and there is a potential to speed up JSON decoding with multiple threads then that seems like something we'd want to do.

westonpace (changes requested):
This is exciting.  Sorry you took this on at a time when we have two mirror paths in the file format.  Hopefully this will all be getting simplified soon.

cpp/src/arrow/dataset/file_json.cc
benibus: Yeah that was a "better safe than sorry" decision. After looking into it, both `ExecContext` and `ScanOptions` provide valid defaults for their memory pool so it probably isn't necessary.

cpp/src/arrow/dataset/file_json.cc
benibus: There was a similar comment in `file_csv.cc` that I based that on (since the CSV reader has similar async properties). Admittedly, I wasn't completely on board with the reasoning myself.

Otherwise, I don't think there's any real risk of nested parallelism.

cpp/src/arrow/dataset/file_json.cc
```diff
@@ -0,0 +1,414 @@
...
+  }
+
+  static Future<std::shared_ptr<FragmentScanner>> Make(
+      const FragmentScanRequest& scan_request,
+      const JsonFragmentScanOptions& format_options,
+      const JsonInspectedFragment& inspected, Executor* cpu_executor) {
+    auto parse_options = format_options.parse_options;
+    ARROW_ASSIGN_OR_RAISE(parse_options.explicit_schema,
+                          GetSchema(scan_request, inspected));
+    parse_options.unexpected_field_behavior = json::UnexpectedFieldBehavior::Ignore;
```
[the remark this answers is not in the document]
benibus: Since the schema derived by `GetSchema` may not contain all of the fields in the fragment schema, forcefully ignoring the remaining fields is the optimal way to ensure non-requested fields don't get parsed/converted. Setting `unexpected_field_behavior` to `Error` or `InferType` wouldn't make much sense here.

The alternative would be post-processing each returned `RecordBatch` to remove unneeded fields - unless I'm thinking about this wrong.

cpp/src/arrow/dataset/file_json.cc
benibus: Yeah, from what I can tell, the new scanner does a lot of the legwork here (as far as top-level fields, at least). For the old path, this was what was required to get the common format tests to pass.

cpp/src/arrow/dataset/file_json.cc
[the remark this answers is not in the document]
westonpace: Ah, that makes sense.  Yes, I agree that `unexpected_field_behavior` is meaningless in a datasets context where the fields requested can change from query to query.

cpp/src/arrow/dataset/file_json.cc
```diff
@@ -0,0 +1,405 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#include "arrow/dataset/file_json.h"
+
+#include <unordered_set>
```
westonpace, line 20:
```suggestion
#include <algorithm>
#include <unordered_set>
#include <vector>
```
algorithm for std::sort

cpp/src/arrow/dataset/file_json.cc
```diff
@@ -0,0 +1,405 @@
...
+
+#include "arrow/dataset/dataset_internal.h"
+#include "arrow/io/buffered.h"
+#include "arrow/json/chunker.h"
+#include "arrow/json/parser.h"
+#include "arrow/json/reader.h"
+#include "arrow/util/checked_cast.h"
+#include "arrow/util/logging.h"
```
westonpace, line 28:
```suggestion
#include "arrow/compute/exec.h"
#include "arrow/compute/exec/expression.h"
#include "arrow/dataset/dataset_internal.h"
#include "arrow/dataset/scanner.h"
#include "arrow/dataset/type_fwd.h"
#include "arrow/io/buffered.h"
#include "arrow/io/interfaces.h"
#include "arrow/io/type_fwd.h"
#include "arrow/json/chunker.h"
#include "arrow/json/parser.h"
#include "arrow/json/reader.h"
#include "arrow/record_batch.h"
#include "arrow/type.h"
#include "arrow/util/async_generator.h"
#include "arrow/util/bit_util.h"
#include "arrow/util/checked_cast.h"
#include "arrow/util/delimiting.h"
#include "arrow/util/logging.h"
#include "arrow/util/thread_pool.h"
```

I'm not sure how practical IWYU is when we don't have the tool wired up :laughing: so feel free to treat this as a minor nit.

cpp/src/arrow/dataset/file_json_test.cc
```diff
@@ -0,0 +1,216 @@
...
+                         ::testing::ValuesIn(TestFormatParams::Values()),
+                         TestFormatParams::ToTestNameString);
+
+TEST_P(TestJsonFormatScanNode, Scan) { TestScan(); }
+TEST_P(TestJsonFormatScanNode, ScanMissingFilterField) { TestScanMissingFilterField(); }
+TEST_P(TestJsonFormatScanNode, ScanProjected) { TestScanProjected(); }
+TEST_P(TestJsonFormatScanNode, ScanProjectedMissingColumns) {
+  TestScanProjectedMissingCols();
+}
+
```
westonpace, line 210: Just scanning through the CSV tests I wonder if there are a few JSON-specific tests that might be nice to have:

 * Confirm that alternate read/parse options (e.g. alternate block size or newlines allowed) get passed in correctly
 * Does the JSON reader handle the the presence of a unicode BOM well?  Maybe not as big an issue here where we have to explicitly handle it for the CSV reader.

westonpace (changes requested):
A few minor nits from a tidy/iwyu pass but mostly clean, thank you.

cpp/src/arrow/dataset/file_json_test.cc
benibus: The reader doesn't explicitly handle BOM, but I suspect that rapidjson does, as it's at least [referenced in their docs](https://rapidjson.org/classrapidjson_1_1_encoding.html). Either way, I added a test for it (which currently passes).

benibus:
@felipecrv Feel free to take a look if you find the time.

westonpace (approved):
Thanks for the addition.  A very cool feature.

felipecrv:
@benibus sorry, I only noticed this today. Feel free to mention my handle as soon as PR is out of draft status.

benibus:
@felipecrv No worries! I'll let you in earlier on the next one.

Issue #33734, opened 2023-01-18
[Go] arrow library is not compatible with grpc < 1.45 due to use of reflection experimental interface


chrisirhc:
### Describe the enhancement requested

When attempting to use arrow library in projects with grpc < 1.45, the reflection was added in v1.45.0 via https://github.com/grpc/grpc-go/commit/18564ff61d5505d955c7bd1adc28e4f1ed96300c . 

This is due to a single line that references an experimental interface in grpc.reflection package:
https://github.com/apache/arrow/blob/c8d6110a26c41966e539e9fa2f5cb8c31dc2f0fe/go/arrow/flight/server.go#L97-L99

The interface is defined as:
https://github.com/grpc/grpc-go/blob/4c776ec01572d55249df309251900554b46adb41/reflection/serverreflection.go#L69-L83

I propose to inline this interface so that the go arrow library can be used in projects with earlier versions of grpc which don't contain this experimental interface. This should maintain the reflection capabilities introduced in https://github.com/apache/arrow/commit/07e7009154dc64967543ccd6462841443a8586b7 but make go arrow library compatible with grpc < 1.45.

### Component(s)

Go

assignUser:
cc @zeroshade

zeroshade:
Issue resolved by pull request 33735
https://github.com/apache/arrow/pull/33735

PR #33735, opened 2023-01-18
GH-33734: [Go] make compatible with grpc < 1.45

chrisirhc:
### Rationale for this change

Define interface from grpc/reflection rather than importing. The interface is marked as experiment, so defining it here is also a step toward improving stability of the go arrow library.

* As above, inlining the interface

### Are these changes tested?

I believe the the change is already tested via:
https://github.com/apache/arrow/blob/c8d6110a26c41966e539e9fa2f5cb8c31dc2f0fe/go/arrow/flight/flight_test.go#L325-L326
And also via confirming via reading https://github.com/grpc/grpc-go/blob/4e4d8288ff4fb99150a526165c767e6fac3bd5ec/Documentation/server-reflection-tutorial.md . However, I have not done any manual testing beyond this.

No.

### Feedback requested
I'd pasted the code verbatim from: https://github.com/grpc/grpc-go/blob/4c776ec01572d55249df309251900554b46adb41/reflection/serverreflection.go#L69-L83

Should I have pasted only what's necessary and link to that code block instead?

### DIFF
--- a/go/arrow/flight/server.go
+++ b/go/arrow/flight/server.go
@@ -24,7 +24,6 @@ import (
 
 	"github.com/apache/arrow/go/v11/arrow/flight/internal/flight"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/reflection"
 )
 
 type (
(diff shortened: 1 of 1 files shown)


go/arrow/flight/server.go
```diff
@@ -64,6 +63,24 @@ func RegisterFlightServiceServer(s grpc.ServiceRegistrar, srv FlightServer) {
...
+// for a custom implementation to return zero values for the
+// grpc.ServiceInfo values in the map.
+//
+// Experimental
+//
+// Notice: This type is EXPERIMENTAL and may be changed or removed in a
+// later release.
+type serviceInfoProvider interface {
+	GetServiceInfo() map[string]grpc.ServiceInfo
+}
```
zeroshade, line 82: Can you add a small comment about the fact that we're inlining this in order to provide compatibility with grpc < 1.45?

zeroshade (changes requested):
LGTM just one nitpick

go/arrow/flight/server.go
chrisirhc: Yep sure thing. Any thoughts about the comment copied verbatim from https://github.com/grpc/grpc-go/blob/4c776ec01572d55249df309251900554b46adb41/reflection/serverreflection.go#L69-L83 ?

chrisirhc:
Addressed comment.

go/arrow/flight/server.go
zeroshade: I think it's fine copying the comment verbatim for now.

zeroshade (approved):
Thanks much @chrisirhc ! once the CI passes I'll merge this

PR #33736, opened 2023-01-18
PARQUET-2232: [C++]  Add an api to ColumnChunkMetaData to indicate if the column chunk uses a bloom filter

jp0317:
Although bloom filter is not fully supported in parquet-cpp for now, it can be useful to provide an api that tells if a column chunk is using bloom filters. This would lead to better understanding of file characteristics.

Changes are tested with unit tests.

Thanks.

### DIFF
--- a/cpp/src/parquet/metadata.cc
+++ b/cpp/src/parquet/metadata.cc
@@ -279,7 +279,7 @@ class ColumnChunkMetaData::ColumnChunkMetaDataImpl {
   const std::vector<PageEncodingStats>& encoding_stats() const { return encoding_stats_; }
 
   inline std::optional<int64_t> bloom_filter_offset() const {
-    if(column_metadata_->__isset.bloom_filter_offset) {
+    if (column_metadata_->__isset.bloom_filter_offset) {
       return column_metadata_->bloom_filter_offset;
     }
     return std::nullopt;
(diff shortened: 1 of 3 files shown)


cpp/src/parquet/metadata.h
```diff
@@ -171,6 +171,7 @@ class PARQUET_EXPORT ColumnChunkMetaData {
 
   const std::vector<Encoding::type>& encodings() const;
   const std::vector<PageEncodingStats>& encoding_stats() const;
+  bool has_bloom_filter() const;
```
wgtmac, line 174:
```suggestion
  std::optional<int64_t> bloom_filter_offset() const;
```

wgtmac: It would be more user-friendly to provide this function, like the `GetColumnIndexLocation` below.

jp0317: Thanks, changed as suggested. PTAL.

mapleFU (approved):
Though I'm not a committer, but currently I'm working on parquet BF. The interface LGTM

cpp/src/parquet/metadata.cc
```diff
@@ -278,6 +278,13 @@ class ColumnChunkMetaData::ColumnChunkMetaDataImpl {
 
   const std::vector<PageEncodingStats>& encoding_stats() const { return encoding_stats_; }
 
+  inline std::optional<int64_t> bloom_filter_offset() const {
+    if(column_metadata_->__isset.bloom_filter_offset) {
```
wjones127, line 282:
```suggestion
    if (column_metadata_->__isset.bloom_filter_offset) {
```

wjones127 (commented):
Looks good, apart from the style lint.

wjones127:
CI failures are unrelated.

PR #33738, opened 2023-01-18
GH-33737: [C++] simplify exec plan tracing

### LINKED ISSUE #33737: [C++] Simplify tracing in exec plan

The old tracing model starts a span when a node starts and ends the span when the node marks itself finished.  Some nodes start an additional InputReceived span with the above mentioned span as parent.  This makes it rather difficult to tell where time is actually being spent because large blocks of the span represent idle time.  It does not accurately reflect time spent.

I've would like to change the model to use async scheduler tasks as spans.  In practice, this would mean that there is now a span per fragment.  It may have child spans for each of the nodes that runs on the fragment (simple nodes may just mark their execution as an event).  This also will allow us to get rid of the ExecNode::finsihed_ future as they are no longer really necessary (they can show up as "waiting for finish" spans that don't really provide any useful information).

### Component(s)

C++

westonpace:
This PR does two things.  First, it requires that all "tasks" (for the AsyncTaskScheduler, not the executor) have a name.  Second, it simplifies and cleans up the way that exec nodes report their tracing using a `TracedNode` helper class.

* Closes: #33737


westonpace:
Example Plan Trace:

image

Note: I can already see something I find rather interesting in the above trace.  The final pipeline calls `ScalarAggregateNode::Finish` which finalizes the aggregates and calls `SinkNode::InputReceived`. Admittedly, finalizing aggregates is not much work, however, all `InputReceived` should be doing is pushing the batch onto a queue.  I'm surprised how much longer it takes to run `InputReceived`.

westonpace:
CC @mbrobbel please review if you have a chance

westonpace:
Looks like I still have a CI issue.  Converting to draft while I work that out.

cpp/src/arrow/compute/exec/sink_node.cc
```diff
@@ -335,19 +327,13 @@ class ConsumingSinkNode : public ExecNode, public BackpressureControl {
...
   }
 
   void InputReceived(ExecNode* input, ExecBatch batch) override {
-    EVENT(span_, "InputReceived", {{"batch.length", batch.length}});
-    util::tracing::Span span;
-    START_COMPUTE_SPAN_WITH_PARENT(
-        span, span_, "InputReceived",
-        {{"node.label", label()}, {"batch.length", batch.length}});
-
+    auto scope = TraceInputReceived(batch);
```
joosthooz, line 336: If the sink applies backpressure (e.g. the datasetwriter), it does not result in an event being created on the span as is the case for the normal SinkNode here https://github.com/apache/arrow/pull/33738/files#diff-967cff6ef1964402635ac0769dece741ca0ed58bcefb8d669ecfe3ed8371998eR175
Is there a way to do this? For example, we could compare the value of `backpressure_counter_` before and after calling `Consume()`, but then we wouldn't be able to know if backpressure was applied or released.

cpp/src/arrow/compute/exec/filter_node.cc
```diff
@@ -103,17 +103,9 @@ class FilterNode : public MapNode {
   }
 
   void InputReceived(ExecNode* input, ExecBatch batch) override {
-    EVENT(span_, "InputReceived", {{"batch.length", batch.length}});
```
joosthooz, line 106: Does this need a `TraceInputReceived` or `NoteInputReceived`?

cpp/src/arrow/compute/exec/project_node.cc
```diff
@@ -96,17 +96,9 @@ class ProjectNode : public MapNode {
   }
 
   void InputReceived(ExecNode* input, ExecBatch batch) override {
-    EVENT(span_, "InputReceived", {{"batch.length", batch.length}});
```
joosthooz, line 99: Does this need a `TraceInputReceived`?

cpp/src/arrow/dataset/file_base.cc
```diff
@@ -556,19 +558,8 @@ class TeeNode : public compute::MapNode {
   }
 
   void InputReceived(compute::ExecNode* input, compute::ExecBatch batch) override {
-    EVENT(span_, "InputReceived", {{"batch.length", batch.length}});
```
joosthooz, line 559: Does this need a `TraceInputReceived`?

cpp/src/arrow/dataset/file_base.cc
```diff
@@ -533,6 +534,7 @@ class TeeNode : public compute::MapNode {
       MapNode::Finish(std::move(finish_st));
       return;
     }
+    auto scope = TraceFinish();
```
joosthooz, line 537: Why do we do a `TraceFinish` here and not inside `DatasetWriter::Finish`? That's the only thing inside this trace and it seems weird to only trace it if it is inside a TeeNode (e.g. it is missing here https://github.com/apache/arrow/pull/33738/files#diff-2caf4e9bd3f139e05e55dca80725d8a9c436f5ccf65c76a37cebfa6ee9b36a6aL418)

joosthooz (commented):
Thanks a lot for this, I still have to try this branch for myself but I left some questions!

joosthooz:
In the figure, why does `WaitForFinish(SinkNode:)` end earlier than the `ScalarAggregate`? Can we add a name (and maybe even an id number in case there are multiple) to the names so that we know which node a span refers to? Lastly, there is a SinkNode but it doesn't seem to perform any work

cpp/src/arrow/compute/exec/sink_node.cc
westonpace: The backpressure is applied in the dataset writer (it has it's own queue).  However, we can add tracing events in the dataset writer.  I'm kind of interested now to see what a dataset writer trace looks like.  I will add something.

cpp/src/arrow/compute/exec/filter_node.cc
westonpace: It gets it from `MapNode`.  This will be more obvious in https://github.com/apache/arrow/pull/15253 because subclasses will no longer implement `InputReceived` (they will use `MapNode::InputReceived` and instead just implement `ProcessBatch`)

westonpace:
> In the figure, why does WaitForFinish(SinkNode:) end earlier than the ScalarAggregate?

The code looks roughly like:

```
void SinkNode::ReceiveLastBatch(batch) {
  output_queue.Enqueue(batch); // 4
  finished.MarkFinished(); // 5
}

void AggregateNode::ReceiveLastBatch(batch) {
    Enqueue(batch); // 2
    aggregates = ComputeAggregates(); // 3
    output->ReceiveLastBatch(batch);
    finished_.MarkFinished(); // 6
}

void SourceNode::ReceiveLastBatch(batch) {
  output->ReceiveLastBatch(batch); // 1
  finished_.MarkFinished(); // 7
}
```
> Can we add a name (and maybe even an id number in case there are multiple) to the names so that we know which node a span refers to?

All of the node-specific spans and events should have the node label as an attribute.  I don't think it's displayed here.  The node label defaults (I think) to NodeType:NodeCounter but I'll check

> Lastly, there is a SinkNode but it doesn't seem to perform any work

There are two kinds of sinks.  The SinkNode has an external queue.  All it does is push batches into the queue.  So no, it should not be doing any work.  The ConsumingSinkNode assumes the batch is consumed as part of the plan (e.g. dataset write) and has no output but it does do work.

westonpace:
Here's a trace from a dataset write.  There are still things that could be cleaned up here.  Pretty much all of the `DatasetWriter::` traces are a mix of active CPU time and idle I/O time and it isn't clear what is what.

Screenshot 2023-01-18 at 17-51-42 Jaeger UI

cpp/src/arrow/compute/exec/sink_node.cc
joosthooz: Yes, I think that makes a lot of sense especially because the dataset writer submits tasks to the IO executor that can run in parallel. All that would otherwise be lost behind a single ConsumingSinkNode span

westonpace:
@joosthooz I slightly changed things so the current task will be used as parent and not the scheduler.  This makes it more clear that WriteAndCheckBackpressure is actually creating some of those following spans.

However, at this point, I think we are veering from my original goal which was "remove the dependence on the exec node finished future so I can get away with it but don't break OT worse than it already was".

I think I'd like to merge this in as it is.  Would you be interested in investigating better ways of handling spans in a future PR?

> Shouldn't there also be a span created in the delta that gets submitted to the IO executor in WriteNext? That's where the actual writing (and parquet encoding & compression) is being performed

I think I/O of any kind is generally interesting enough to always justify a span.

joosthooz:
Hi, I gave this branch a spin, and it seems that the nesting has become inconsistent: 
image
There's 2 ReadBatch spans under InitialTask. 1 of these has all the FragmentsToBatches as its child spans (these were nested under the SourceNode before). The other keeps recursively nesting more ReadBatch spans. Each has a ProcessMorsel, that has the filter, project and sink spans nested under each other. Then the dataset writer also keeps nesting WriteAndCheckBackpressure.
image
Is there a way to go back to making most of these spans siblings again?
Do we want to change the organization of the spans in this PR from having 1 span for each node in the graph, each having a span for every chunk of data it processes (how it was before), to having a ProcessMorsel for each chunk of data, each having a span for each node it traverses through?
I think I can help in a follow-up PR, especially for the dataset writer.

westonpace:
>  Is there a way to go back to making most of these spans siblings again?

Yes.  The recursion is probably somewhat accurate but I agree it makes it harder to read.  Having them as siblings makes sense too.  I will revert back to this understanding.

>  Do we want to change the organization of the spans in this PR from having 1 span for each node in the graph, each having a span for every chunk of data it processes (how it was before), to having a ProcessMorsel for each chunk of data, each having a span for each node it traverses through?

Yes, I believe so.  A more generic term than `ProcessMorsel` would be "pipeline" or "plan fragment".  Acero is (implicitly) a "plan fragment" engine in that we have one task per batch per fragment.  Thinking about it this way it would be nice if we had something like the conceptual model:

 * ProcessMorsel
   * Filter
   * Project
   * Sink

But today it ends up being (because the last part of each node is to call the downstream node):

 * ProcessMorsel
   * Filter
     * Project
       * Sink

Yet another case where the conceptual/logical understanding is different than the physical understanding.  Although, perhaps there is some merit for the physical understanding as it mirrors reality more closely.  Perhaps it depends on the goal of the reader.  If someone is trying to improve the threading and execution of the plan itself they might want the physical model.  If someone is trying to focus on the performance of a single node they might want the logical model.  I'll leave that for a follow-up PR.

>  I think I can help in a follow-up PR, especially for the dataset writer.

Great.  Mentally, when I think of the dataset writer, I think there are two parts.  The first part should be the trailing part of the fragment/pipeline that feeds the writer.  In this first part we partition the batch, select the appropriate file queues, and deposit the batches into the queues.  There is then a separate dedicated thread task to write each batch to the writer.

westonpace:
I'm going to merge this as-is to unblock the error handling cleanup.  We can fine tune in future PRs.

lidavidm:
I was about to start reviewing it, sorry for the delay :sweat_smile: In any case, Joost already took a look here fortunately.

westonpace:
> I was about to start reviewing it, sorry for the delay sweat_smile In any case, Joost already took a look here fortunately.

No problem.  I might be moving a little fast but I think the tracing stuff is still pretty experimental at the moment.  One thing I forgot to note is that I have upgraded OT from 1.4 to 1.8.  I found that 1.4 could not connect directly to a connector for some reason.  1.8 seems to work out of the box.  Also, I noticed that 1.8 now has a jaeger exporter which could potentially avoid the need to have a collector at all.  I tried to enable it but quickly ran into trouble with the bundled build which seems to be pretty custom.

Issue #33743, opened 2023-01-18
[Java] Release outstanding buffers when BaseAllocator is being closed


zhztheplayer:
### Describe the enhancement requested

This is mainly aim to enhance [BaseAllocator#close()](https://github.com/apache/arrow/blob/4e439f6a597180c5fc8ff1552c860cecd33736c5/java/memory/memory-core/src/main/java/org/apache/arrow/memory/BaseAllocator.java#L370-L454) to implement the original design of its super method `BufferAllocator#close()`: 

https://github.com/apache/arrow/blob/4e439f6a597180c5fc8ff1552c860cecd33736c5/java/memory/memory-core/src/main/java/org/apache/arrow/memory/BufferAllocator.java#L88-L95

The implementation should be fast enough to not impact current allocation process much. Also we should put detailed information of this clean-up action into allocator-close logs.

### Component(s)

Java

PR #33744, opened 2023-01-18
GH-33743: [Java] Release outstanding buffers when BaseAllocator is being closed

### LINKED ISSUE #223: [Java] Release outstanding buffers when BaseAllocator is being closed

This is mainly aim to enhance [BaseAllocator#close()](https://github.com/apache/arrow/blob/4e439f6a597180c5fc8ff1552c860cecd33736c5/java/memory/memory-core/src/main/java/org/apache/arrow/memory/BaseAllocator.java#L370-L454) to implement the original design of its super method `BufferAllocator#close()`: 

https://github.com/apache/arrow/blob/4e439f6a597180c5fc8ff1552c860cecd33736c5/java/memory/memory-core/src/main/java/org/apache/arrow/memory/BufferAllocator.java#L88-L95

The implementation should be fast enough to not impact current allocation process much. Also we should put detailed information of this clean-up action into allocator-close logs.

### Component(s)

Java

zhztheplayer:
See https://github.com/apache/arrow-java/issues/223

1. Close outstanding buffer(ledger)s during allocator-close;
2. Report warning messages into log after the outstanding buffers were closed;
3. To help implement the feature, construct a linked-list of buffer ledgers and hold a reference to tail element from buffer alloctor.

### Are these changes tested?

1. Current UTs and added UTs.
2. Benchmarking by AllocatorBenchmarks

No
* Closes: apache/arrow-java#223
* GitHub Issue: #223

### DIFF
--- a/java/memory/memory-core/src/main/java/org/apache/arrow/memory/BaseAllocator.java
+++ b/java/memory/memory-core/src/main/java/org/apache/arrow/memory/BaseAllocator.java
@@ -68,6 +68,12 @@ abstract class BaseAllocator extends Accountant implements BufferAllocator {
   private final BaseAllocator parentAllocator;
   private final Map<BaseAllocator, Object> childAllocators;
   private final ArrowBuf empty;
+
+  // linked list to hold references to alive ledgers
+  private BufferLedger tail = null;
+
+  private final Object REF_LOCK = new Object();
+
   // members used purely for debugging
   private final IdentityHashMap<BufferLedger, Object> childLedgers;
   private final IdentityHashMap<Reservation, Object> reservations;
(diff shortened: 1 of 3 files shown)


lidavidm (changes requested):
What is the purpose? I don't think we want this.

This silently suppresses an application logic error, and forces all allocations to go through a lock. The exception you get if there are unclosed buffers is very intentional.

The close() method's docstring is admittedly misleading, but I do not think the intent was to just close all buffers for you. (I think it's speaking more towards releasing any pooled or cached memory, which is definitely true for the Netty implementation.)

lidavidm:
This patch forces everyone to take the opposite tradeoff, however. 

Some sort of nursery that auto-closes buffers may be useful, but I think it should be separate from the base allocator. (The allocator also has a 'listener' to be notified of allocation/deallocation events; possibly you could build this mechanism on top of that.) And nondeterministic collection of buffers is liable to bite you; the GC does not see into native allocations and you can run out of memory despite having 'free' memory because the GC has not collected the small Java-side objects that hold the large native buffers.

zhztheplayer:
Thanks and would you suggest to add a new implementation of interface `BufferAllocator` (e.g. `AutoCleanBufferAllocator`)? If yes I can try start from there.

> the GC does not see into native allocations and you can run out of memory despite having 'free' memory because the GC has not collected the small Java-side objects that hold the large native buffers.

Yes I can image something like that too. What I can try is to reuse the JVM's direct memory counter which can trigger GC when reaching its limit, or rebuild a similar counter like that. Not sure if there are some better solutions but this one is probably OK since it was at least from JVM itself.

zhztheplayer:
> I don't think it needs a new interface.

I meant a new implementation of the interface, e.g. 

`class AutoCleanBufferAllocator implements BufferAllocator`

lidavidm:
Ah, ok. That sounds fine. I would explore the allocation listener further, though...

zhztheplayer:
I see. I'll find out to what extent can allocation listener to be leveraged in this feature. Thanks for the suggestion.

amol-:

PR #33764, opened 2023-01-18
GH-15109: [Python] Allow creation of non empty struct array with zero field

### LINKED ISSUE #15109: Can't create a non empty StructArray with no field using `StructArray.from_array`

I want to create a StructArray with several rows but no fields/columns:

```
    array = pa.StructArray.from_arrays(arrays=[], names=[], mask=pa.array([True, True]))
    assert array.type == pa.struct([])
    assert len(array) == 0 # This is wrong
```

As commented in the code `len(array)` is wrong and should be 2 (according to the provided mask).

I found a work around but it's not ideal:

```
    array = pa.StructArray.from_arrays(
        arrays=[pa.nulls(2, pa.null())], names=["DELETE"], mask=pa.array([True, True])
    )
    assert len(array) == 2
    empty_array = array.cast(pa.struct([]))
    assert len(empty_array) == 2
    assert empty_array.type == pa.struct([])
```

### Component(s)

Python

0x26res:
@jorisvandenbossche 
- I made minor changes to the C++ code, I couldn't help myself
- I've followed your suggestion, I've used `{}` to represent an empty struct python value, instead of `[]`. Because that's what is used by `as_py`
- I've also passed the mask to `pa.array` because we want it to be applied to the resulting struct array.

* Closes: #15109


jorisvandenbossche (approved):
Looks good to me, thanks!

PR #33768, opened 2023-01-18
GH-33767: [Go] Clear out parameter in ArrowArrayStream.get_next

### LINKED ISSUE #33767: [Go] Exported ArrowArrayStream.get_next doesn't handle uninitialized ArrowArrays well

`get_next` should set `ArrowArray.release` to `NULL` when there are no more records. However, the current implementation instead tries to _release_ the out-parameter. This is harmless when the out-parameter is 0-initialized (the implementation will skip the call) but otherwise it'll crash (after jumping to a random garbage address).

### Component(s)

Go

lidavidm:
* Closes: #33767


zeroshade (approved):
LGTM

lidavidm:
I noticed the unit test doens't actually assert that err is nil (seems Go doesn't complain about unused return values?) so I'll fix that

PR #33774, opened 2023-01-19
GH-15200: [C++] Add specific ndigits value for round

### LINKED ISSUE #15200: [C++][Arrow] Add benchmarks for the round functions.

EpsilonPrime:
Adds binary round kernel for float types.
Provides fallback for binary round from int types to double. Does not provide fallback for binary round for decimal types.

Minor:
Fixes fallback behavior to only affect the first argument. Corrects fallback error message to display the original type instead of the fallback type when the underlying type cannot be found. Round benchmark changed to be driven like arithmetic benchamrk instead of the string benchmark.

Trivial:
Renames Arithmetic to Round to match the containing module. Removes some unreachable code.

* Closes: #15200


EpsilonPrime:
Replaces PR 30309.

EpsilonPrime:
Wrong bug, I'll try again.

PR #33775, opened 2023-01-19
GH-33588: [Substrait] Add Substrait→Acero mapping for round operationMajor:

EpsilonPrime:
Adds binary round kernel for float types.
Provides fallback for binary round from int types to double. Does not provide fallback for binary round for decimal types.

Minor:
Fixes fallback behavior to only affect the first argument. Corrects fallback error message to display the original type instead of the fallback type when the underlying type cannot be found. Round benchmark changed to be driven like arithmetic benchamrk instead of the string benchmark.

Trivial:
Renames Arithmetic to Round to match the containing module. Removes some unreachable code.

* Closes: #33588

### DIFF
--- a/cpp/src/arrow/compute/api_scalar.h
+++ b/cpp/src/arrow/compute/api_scalar.h
@@ -91,6 +91,15 @@ class ARROW_EXPORT RoundOptions : public FunctionOptions {
   RoundMode round_mode;
 };
 
+class ARROW_EXPORT RoundBinaryOptions : public FunctionOptions {
+ public:
+  explicit RoundBinaryOptions(RoundMode round_mode = RoundMode::HALF_TO_EVEN);
+  static constexpr char const kTypeName[] = "RoundBinaryOptions";
+  static RoundBinaryOptions Defaults() { return RoundBinaryOptions(); }
+  /// Rounding and tie-breaking mode
+  RoundMode round_mode;
+};
+
 enum class CalendarUnit : int8_t {
   NANOSECOND,
   MICROSECOND,
@@ -881,6 +890,20 @@ Result<Datum> Sign(const Datum& arg, ExecContext* ctx = NULLPTR);
 ARROW_EXPORT
 Result<Datum> Round(const Datum& arg, RoundOptions options = RoundOptions::Defaults(),
                     ExecContext* ctx = NULLPTR);
+
+/// \brief Round a value to a given precision.
+///
+/// If argument is null the result will be null.
+///
+/// \param[in] arg1 the value rounded
+/// \param[in] arg2 the number of significant digits to round to
+/// \param[in] options rounding options (rounding mode and number of digits), optional
+/// \param[in] ctx the function execution context, optional
+/// \return the element-wise rounded value
+ARROW_EXPORT
+Result<Datum> RoundBinary(const Datum& arg1, const Datum& arg2,
+                          RoundBinaryOptions options = RoundBinaryOptions::Defaults(),
+                          ExecContext* ctx = NULLPTR);
 
 /// \brief Round a value to a given multiple.
 ///
--- a/cpp/src/arrow/compute/api_scalar.h
+++ b/cpp/src/arrow/compute/api_scalar.h
@@ -881,9 +881,9 @@ Result<Datum> Sign(const Datum& arg, ExecContext* ctx = NULLPTR);
 
 /// \brief Round a value to a given precision.
 ///
-/// If argument is null the result will be null.
+/// If arg is null the result will be null.
 ///
-/// \param[in] arg the value rounded
+/// \param[in] arg the value to be rounded
 /// \param[in] options rounding options (rounding mode and number of digits), optional
 /// \param[in] ctx the function execution context, optional
 /// \return the element-wise rounded value
@@ -893,11 +893,15 @@ Result<Datum> Round(const Datum& arg, RoundOptions options = RoundOptions::Defau
 
 /// \brief Round a value to a given precision.
 ///
-/// If argument is null the result will be null.
+/// If arg1 is null the result will be null.
+/// If arg2 is null then the default number of significant digits (zero) will be used instead.
+/// If arg2 is negative, then the rounding place will be shifted to the left (thus -1 would
+/// correspond to rounding to the nearest ten).  If positive, the rounding place will shift
+/// to the right (and +1 would correspond to rounding to the nearest tenth).
 ///
-/// \param[in] arg1 the value rounded
+/// \param[in] arg1 the value to be rounded
 /// \param[in] arg2 the number of significant digits to round to
-/// \param[in] options rounding options (rounding mode and number of digits), optional
+/// \param[in] options rounding options, optional
 /// \param[in] ctx the function execution context, optional
 /// \return the element-wise rounded value
 ARROW_EXPORT
--- a/cpp/src/arrow/compute/api_scalar.h
+++ b/cpp/src/arrow/compute/api_scalar.h
@@ -894,10 +894,11 @@ Result<Datum> Round(const Datum& arg, RoundOptions options = RoundOptions::Defau
 /// \brief Round a value to a given precision.
 ///
 /// If arg1 is null the result will be null.
-/// If arg2 is null then the default number of significant digits (zero) will be used instead.
-/// If arg2 is negative, then the rounding place will be shifted to the left (thus -1 would
-/// correspond to rounding to the nearest ten).  If positive, the rounding place will shift
-/// to the right (and +1 would correspond to rounding to the nearest tenth).
+/// If arg2 is null then the default number of significant digits (zero) will be used
+/// instead. If arg2 is negative, then the rounding place will be shifted to the left
+/// (thus -1 would correspond to rounding to the nearest ten).  If positive, the rounding
+/// place will shift to the right (and +1 would correspond to rounding to the nearest
+/// tenth).
 ///
 /// \param[in] arg1 the value to be rounded
 /// \param[in] arg2 the number of significant digits to round to
--- a/cpp/src/arrow/compute/api_scalar.h
+++ b/cpp/src/arrow/compute/api_scalar.h
@@ -894,11 +894,10 @@ Result<Datum> Round(const Datum& arg, RoundOptions options = RoundOptions::Defau
 /// \brief Round a value to a given precision.
 ///
 /// If arg1 is null the result will be null.
-/// If arg2 is null then the default number of significant digits (zero) will be used
-/// instead. If arg2 is negative, then the rounding place will be shifted to the left
-/// (thus -1 would correspond to rounding to the nearest ten).  If positive, the rounding
-/// place will shift to the right (and +1 would correspond to rounding to the nearest
-/// tenth).
+/// If arg2 is null then the result will be null. If arg2 is negative, then the rounding
+/// place will be shifted to the left (thus -1 would correspond to rounding to the nearest
+/// ten).  If positive, the rounding place will shift to the right (and +1 would
+/// correspond to rounding to the nearest tenth).
 ///
 /// \param[in] arg1 the value to be rounded
 /// \param[in] arg2 the number of significant digits to round to
--- a/cpp/src/arrow/compute/kernels/scalar_round.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_round.cc
@@ -25,9 +25,7 @@
 #include "arrow/compare.h"
 #include "arrow/compute/api_scalar.h"
 #include "arrow/compute/cast.h"
-#include "arrow/compute/kernels/base_arithmetic_internal.h"
 #include "arrow/compute/kernels/common.h"
-#include "arrow/compute/kernels/util_internal.h"
 #include "arrow/type.h"
 #include "arrow/type_traits.h"
 #include "arrow/util/decimal.h"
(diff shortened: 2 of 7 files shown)


cpp/src/arrow/compute/api_scalar.h
```diff
@@ -882,6 +891,20 @@ ARROW_EXPORT
 Result<Datum> Round(const Datum& arg, RoundOptions options = RoundOptions::Defaults(),
                     ExecContext* ctx = NULLPTR);
 
+/// \brief Round a value to a given precision.
+///
+/// If argument is null the result will be null.
+///
+/// \param[in] arg1 the value rounded
+/// \param[in] arg2 the number of significant digits to round to
+/// \param[in] options rounding options (rounding mode and number of digits), optional
```
westonpace, line 900:
```suggestion
/// \param[in] options rounding options (rounding mode), optional
```
Or just get rid of the parentheses section entirely.

cpp/src/arrow/compute/kernels/scalar_round_arithmetic_test.cc
```diff
@@ -18,7 +18,6 @@
 #include <algorithm>
 #include <cmath>
 #include <memory>
-#include <string>
```
westonpace, line 21: Our guideline for includes is [`iwyu`](https://include-what-you-use.org/).  We don't always follow it perfectly (the conformance tool doesn't like type_fwd files) but it is what we aim for.  Please don't remove includes if they are used in the file (I still see many instances of `std::string`) even if the file compiles otherwise (transitive includes are potentially unstable).

cpp/src/arrow/compute/kernels/scalar_round_arithmetic_test.cc
```diff
@@ -113,7 +112,7 @@ class TestBaseUnaryRoundArithmetic : public ::testing::Test {
   // (Array, Array)
   void AssertUnaryOp(UnaryFunction func, const std::shared_ptr<Array>& arg,
                      const std::shared_ptr<Array>& expected) {
-    ASSERT_OK_AND_ASSIGN(auto actual, func(arg, options_, nullptr));
+    ASSERT_OK_AND_ASSIGN(auto actual, func(arg, options_, nullptr))
```
westonpace, line 115: While this does compile and work we try and add a `;` to the end of these kinds of macros anyways for readability
```suggestion
    ASSERT_OK_AND_ASSIGN(auto actual, func(arg, options_, nullptr));
```

cpp/src/arrow/compute/kernels/scalar_round_arithmetic_test.cc
```diff
@@ -965,6 +1086,97 @@ TYPED_TEST(TestUnaryRoundFloating, Round) {
...
+      {-1, "[320, 0.0, 0.0, 0.0, -0.0, -40, -0.0]"},
+      {0, "[320, 4, 3, 5, -3, -35, -3]"},
+      {1, "[320, 3.5, 3.1, 4.5, -3.2, -35.1, -3]"},
+      {2, "[320, 3.5, 3.08, 4.5, -3.21, -35.12, -3.05]"},
+  }};
+  this->SetRoundMode(RoundMode::HALF_TOWARDS_INFINITY);
+  for (const auto& pair : ndigits_and_expected) {
+    this->AssertBinaryOp(RoundBinary, values, pair.first, pair.second);
+  }
+}
```
westonpace, line 1178: Can you add some unit tests that consider nulls in both the values and the num_digits arguments?  Also maybe a few tests with scalars (esp. using a scalar for num_digits and an array for values which should be equivalent to unary rounding)

cpp/src/arrow/engine/substrait/extension_set.cc
```diff
@@ -790,6 +796,30 @@ ExtensionIdRegistry::SubstraitCallToArrow DecodeOptionlessUncheckedArithmetic(
...
+        compute::RoundMode round_mode,
+        ParseOptionOrElse(
+            call, "rounding", kRoundModeParser,
+            {compute::RoundMode::DOWN, compute::RoundMode::UP,
+             compute::RoundMode::TOWARDS_ZERO, compute::RoundMode::TOWARDS_INFINITY,
+             compute::RoundMode::HALF_DOWN, compute::RoundMode::HALF_UP,
+             compute::RoundMode::HALF_TOWARDS_ZERO,
+             compute::RoundMode::HALF_TOWARDS_INFINITY, compute::RoundMode::HALF_TO_EVEN,
+             compute::RoundMode::HALF_TO_ODD},
+            compute::RoundMode::HALF_TOWARDS_INFINITY));
```
westonpace, line 812: It appears you are defaulting to `HALF_TOWARDS_INFINITY` but shouldn't the default be `HALF_TO_EVEN`?

cpp/src/arrow/engine/substrait/extension_set.cc
```diff
@@ -790,6 +796,30 @@ ExtensionIdRegistry::SubstraitCallToArrow DecodeOptionlessUncheckedArithmetic(
...
+    ARROW_ASSIGN_OR_RAISE(std::vector<compute::Expression> value_args,
+                          GetValueArgs(call, 0));
+    std::shared_ptr<compute::RoundBinaryOptions> options =
```
westonpace, line 815: Do we want to optimize and call the unary round if the second value is a scalar?  If not in this PR can we create a follow-up github issue so we don't lose track of it?  Or maybe round_binary itself can fallback to unary rounding if the second argument is scalar.

cpp/src/arrow/compute/kernels/scalar_round.cc
```diff
@@ -452,6 +468,127 @@ struct Round<ArrowType, kRoundMode, enable_if_decimal<ArrowType>> {
...
+  using State = RoundOptionsWrapper<RoundBinaryOptions>;
+
+  explicit RoundBinary(const State& state, const DataType& out_ty) {}
+
+  template <typename T = ArrowType, typename CType0 = typename TypeTraits<T>::CType0,
+            typename CType1 = typename TypeTraits<T>::CType1>
+  enable_if_floating_value<CType> Call(KernelContext* ctx, CType0 arg0, CType1 arg1,
+                                       Status* st) const {
+    // Do not process Inf or NaN because they will trigger the overflow error at end of
+    // function.
```
westonpace, line 483: Do you have any tests with infinite or NaN?

cpp/src/arrow/compute/kernels/scalar_round.cc
```diff
@@ -751,60 +877,25 @@ ArrayKernelExec GenerateArithmeticWithFixedIntOutType(detail::GetTypeId get_id)
...
-      }
-    }
-    return Status::OK();
-  }
 };
 
-/// An ArithmeticFunction that promotes only decimal arguments to double.
-struct ArithmeticDecimalToFloatingPointFunction : public ArithmeticFunction {
-  using ArithmeticFunction::ArithmeticFunction;
+/// An RoundFunction that promotes only decimal arguments to double.
```
westonpace, line 896:
```suggestion
/// A RoundFunction that promotes only decimal arguments to double.
```

westonpace (changes requested):
Thanks for the contribution, I have a few questions / suggestions.

cpp/src/arrow/engine/substrait/extension_set.cc
EpsilonPrime: HALF_TOWARDS_INFINITY mirrors Ceiling, AWAY_FROM_ZERO mirrors the way humans round.  What rationale is there for HALF_TO_EVEN?  Is it because it's the first option listed in the proposed YAML definition?

cpp/src/arrow/engine/substrait/extension_set.cc
EpsilonPrime: I did't add it in this PR because there was already a lot of moving parts.  I've created GH-33833 to handle investigation into optimization of this feature.

cpp/src/arrow/compute/api_scalar.h
```diff
@@ -882,6 +891,20 @@ ARROW_EXPORT
 Result<Datum> Round(const Datum& arg, RoundOptions options = RoundOptions::Defaults(),
                     ExecContext* ctx = NULLPTR);
 
+/// \brief Round a value to a given precision.
+///
+/// If argument is null the result will be null.
```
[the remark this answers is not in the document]
EpsilonPrime: Added.

cpp/src/arrow/compute/api_scalar.h
```diff
@@ -882,6 +891,20 @@ ARROW_EXPORT
...
+///
+/// \param[in] arg1 the value rounded
```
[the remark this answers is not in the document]
EpsilonPrime: Added.

cpp/src/arrow/compute/api_scalar.h
```diff
@@ -882,6 +891,20 @@ ARROW_EXPORT
...
+/// \param[in] arg2 the number of significant digits to round to
```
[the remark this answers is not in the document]
EpsilonPrime: Defined.

cpp/src/arrow/compute/api_scalar.h
EpsilonPrime: Done.

cpp/src/arrow/compute/kernels/scalar_round_arithmetic_test.cc
EpsilonPrime: I've been spoiled by having iwyu being enforced by the compiler.  I wonder what it would take to get the project to a point where we can enjoy that.  For now I'll tell clion to stop warning me about unnecessary includes.

cpp/src/arrow/compute/kernels/scalar_round_arithmetic_test.cc
EpsilonPrime: Another clion special.  I'll turn off the option warning about empty statements.

cpp/src/arrow/compute/kernels/scalar_round.cc
EpsilonPrime: updated.

cpp/src/arrow/compute/kernels/scalar_round.cc
EpsilonPrime: Yes, they are in TEST_F(TestBinaryRoundFloating, Round) around line 1161.

cpp/src/arrow/engine/substrait/extension_set.cc
westonpace: Yes, for some reason I thought the spec declared that the first option specified should be the default.  However, upon re-reading, it appears I was incorrect.

> Options can be left unspecified and the consumer is free to choose which implementation to use.

That being said, HALF_TO_EVEN is the default in Arrow if you are not using Substrait and it is the default IEEE rounding mode.  I don't know exactly why but, after some googling, it seems that the most likely reason is that it tends to be the most unbiased: https://stackoverflow.com/questions/45129046/why-floats-rounding-use-round-to-even

I would agree though, that AWAY_FROM_ZERO would be my intuitive expectation.

cpp/src/arrow/compute/kernels/scalar_round_arithmetic_test.cc
EpsilonPrime: Done.

cpp/src/arrow/engine/substrait/extension_set.cc
EpsilonPrime: Done.

cpp/src/arrow/compute/kernels/scalar_round.cc
```diff
@@ -906,9 +980,8 @@ void AddNullExec(ScalarFunction* func) {
   DCHECK_OK(func->AddKernel(std::move(input_types), OutputType(null()), NullToNullExec));
 }
 
-template <typename Op, typename FunctionImpl = ArithmeticFunction>
-std::shared_ptr<ScalarFunction> MakeArithmeticFunction(std::string name,
-                                                       FunctionDoc doc) {
+template <typename Op, typename FunctionImpl = RoundFunction>
+std::shared_ptr<ScalarFunction> MakeRoundFunction(std::string name, FunctionDoc doc) {
```
westonpace, line 984: You didn't add this but...is this used?  I think this file must have been generated from scalar_arithmetic.cc and not quite cleaned up all the way.

cpp/src/arrow/compute/kernels/scalar_round.cc
```diff
@@ -918,11 +991,11 @@ std::shared_ptr<ScalarFunction> MakeArithmeticFunction(std::string name,
...
 }
 
-// Like MakeArithmeticFunction, but for arithmetic ops that need to run
+// Like MakeRoundFunction, but for arithmetic ops that need to run
 // only on non-null output.
-template <typename Op, typename FunctionImpl = ArithmeticFunction>
-std::shared_ptr<ScalarFunction> MakeArithmeticFunctionNotNull(std::string name,
-                                                              FunctionDoc doc) {
+template <typename Op, typename FunctionImpl = RoundFunction>
+std::shared_ptr<ScalarFunction> MakeRoundFunctionNotNull(std::string name,
```
westonpace, line 997: Same here.  Is this used?

westonpace (changes requested):
I think one small opportunity for a bit of cleanup but overall looks good to me.  Thanks for doing this!

cpp/src/arrow/compute/kernels/scalar_round.cc
EpsilonPrime: Dropped.

cpp/src/arrow/compute/kernels/scalar_round.cc
EpsilonPrime: Dropped.

westonpace (approved):
Thanks for the addition!  s3fs test failures are unrelated.

PR #33785, opened 2023-01-19
GH-33741: [Python] Address docstrings in Data Types Factory Functions

### LINKED ISSUE #33741: [Python] Address docstrings in Data Types Factory Functions

Ensure docstrings for [Data Types Factory Functions](https://arrow.apache.org/docs/python/api/datatypes.html#factory-functions) have an Examples section.

### Component(s)

Python

AlenkaF:
### Rationale for this change
Ensure docstrings for [Data Types Factory Functions](https://arrow.apache.org/docs/python/api/datatypes.html#factory-functions) have an Examples section.

Docstrings are added to listed Data Types Factory Functions.

### Are these changes tested?
Yes, locally with `pytest --doctest-cython --disable-warnings` pyarrow and on the CI with `Python / AMD64 Conda Python 3.9 Sphinx & Numpydoc` build.

No.
* Closes: #33741

### DIFF
--- a/python/pyarrow/types.pxi
+++ b/python/pyarrow/types.pxi
@@ -2887,6 +2887,16 @@ def date64():
 def float16():
     """
     Create half-precision floating point type.
+
+    Examples
+    --------
+    Create an instance of float16 type:
+
+    >>> import pyarrow as pa
+    >>> pa.float16()
+    DataType(halffloat)
+    >>> print(pa.float16())
+    halffloat
     """
     return primitive_type(_Type_HALF_FLOAT)
 
@@ -2894,6 +2904,26 @@ def float16():
 def float32():
     """
     Create single-precision floating point type.
+
+    Examples
+    --------
+    Create an instance of float32 type:
+
+    >>> import pyarrow as pa
+    >>> pa.float32()
+    DataType(float)
+    >>> print(pa.float32())
+    float
+
+    Create an array with float32 type:
+
+    >>> pa.array([0.0, 1.0, 2.0], type=pa.float32())
+    <pyarrow.lib.FloatArray object at ...>
+    [
+      0,
+      1,
+      2
+    ]
     """
     return primitive_type(_Type_FLOAT)
 
@@ -2901,6 +2931,26 @@ def float32():
 def float64():
     """
     Create double-precision floating point type.
+
+    Examples
+    --------
+    Create an instance of float64 type:
+
+    >>> import pyarrow as pa
+    >>> pa.float64()
+    DataType(double)
+    >>> print(pa.float64())
+    double
+
+    Create an array with float64 type:
+
+    >>> pa.array([0.0, 1.0, 2.0], type=pa.float64())
+    <pyarrow.lib.DoubleArray object at ...>
+    [
+      0,
+      1,
+      2
+    ]
     """
     return primitive_type(_Type_DOUBLE)
 
--- a/python/pyarrow/types.pxi
+++ b/python/pyarrow/types.pxi
@@ -2320,6 +2320,28 @@ def field(name, type, bint nullable=True, metadata=None):
     Returns
     -------
     field : pyarrow.Field
+
+    Examples
+    --------
+    Create an instance of pyarrow.Field:
+
+    >>> import pyarrow as pa
+    >>> pa.field('key', pa.int32())
+    pyarrow.Field<key: int32>
+    >>> pa.field('key', pa.int32(), nullable=False)
+    pyarrow.Field<key: int32 not null>
+
+    >>> field = pa.field('key', pa.int32(),
+    ...                  metadata={"key": "Something important"})
+    >>> field
+    pyarrow.Field<key: int32>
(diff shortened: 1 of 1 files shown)


python/pyarrow/types.pxi
```diff
@@ -2347,69 +2369,260 @@ cdef set PRIMITIVE_TYPES = set([
...
+    >>> pa.bool_()
+    DataType(bool)
+    >>> print(pa.bool_())
+    bool
+
+    Create a ``Field`` type with a boolean type
+    and a name:
+
+    >>> pa.field('null_field', pa.bool_())
+    pyarrow.Field<null_field: bool>
```
danepitkin, line 2409: should we rename `null_field` -> `bool_field`?

danepitkin (commented):
A few small suggestions, but overall LGTM!

python/pyarrow/types.pxi
```diff
@@ -2509,6 +2724,16 @@ def timestamp(unit, tz=None):
     >>> pa.timestamp('s', tz='+07:30')
     TimestampType(timestamp[s, tz=+07:30])
 
+    Use timestamp type when creating a scalar object:
+
+    >>> from datetime import datetime
+    >>> pa.scalar(datetime(2012, 1, 1),
+    ...           type=pa.timestamp('s', tz='+07:30'))
```
jorisvandenbossche, line 2731:
```suggestion
    >>> pa.scalar(datetime(2012, 1, 1), type=pa.timestamp('s', tz='+07:30'))
```

(this fits on a single line for PEP8?)

python/pyarrow/types.pxi
```diff
@@ -2509,6 +2724,16 @@ def timestamp(unit, tz=None):
...
+    <pyarrow.TimestampScalar: datetime.datetime(2012, 1, 1, 7, 30, tzinfo=pytz.FixedOffset(450))>
+    >>> pa.scalar(datetime(2012, 1, 1),
+    ...           type=pa.timestamp('us'))
```
jorisvandenbossche, line 2734:
```suggestion
    >>> pa.scalar(datetime(2012, 1, 1), type=pa.timestamp('us'))
```

python/pyarrow/types.pxi
jorisvandenbossche: And I would maybe show a timezone name instead of a fixed offset (in general there are not many reasons to use a fixed offset, I think). Maybe "UTC", to show that that is also an option.

python/pyarrow/types.pxi
```diff
@@ -2664,41 +2901,145 @@ def month_day_nano_interval():
...
+    Create an instance of 32-bit date type:
+
+    >>> import pyarrow as pa
+    >>> pa.date32()
+    DataType(date32[day])
+
+    Create a scalar with 32-bit date type:
+
+    >>> from datetime import datetime
+    >>> pa.scalar(datetime(2012, 1, 1), type=pa.date32())
```
jorisvandenbossche, line 2936: Do we want to show creating it from `datetime.date` instead? (or in addition)

python/pyarrow/types.pxi
```diff
@@ -2810,6 +3237,26 @@ def large_binary():
...
+    --------
+    Create an instance of large variable-length binary type:
+
+    >>> import pyarrow as pa
+    >>> pa.large_binary()
+    DataType(large_binary)
+
+    and use the type to create an array:
+
+    >>> pa.array(['foo', 'bar'] * 50, type=pa.large_binary())
```
jorisvandenbossche, line 3251: I don't know if the `* 50` is needed here. I would maybe use the same `['foo', 'bar', 'baz']` as you used above for the normal binary type. The `* 50` might indicate that it is "larger", but at the same time it is still tiny data, so I am not sure that helps.

python/pyarrow/types.pxi
```diff
@@ -2914,6 +3454,34 @@ cpdef MapType map_(key_type, item_type, keys_sorted=False):
...
+
+    >>> import pyarrow as pa
+    >>> pa.map_(pa.string(), pa.int32())
+    MapType(map<string, int32>)
+    >>> pa.map_(pa.string(), pa.int32(), keys_sorted=True)
+    MapType(map<string, int32, keys_sorted>)
+
+    Use MapType to create an array:
+
+    >>> data = [[{'key': 'a', 'value': 1}, {'key': 'b', 'value': 2}]]
```
jorisvandenbossche, line 3470:
```suggestion
    >>> data = [[{'key': 'a', 'value': 1}, {'key': 'b', 'value': 2}], [{'key': 'c', 'value': 3}]]
```

By adding a second element of the array, it might be clearer that this is actually like a nested list (where the first element of the map has two key/values mappings, and the second element has one). Currently someone not familiar with the repr might interpret it as an array of length 2

jorisvandenbossche (commented):
Some tiny comments, but generally looks good!

jorisvandenbossche:
Thanks!

PR #33792, opened 2023-01-19
GH-33789: [Go] Add Err() to RecordReader 

### LINKED ISSUE #33789: [Go] RecordReader has no way to propagate errors

RecordReader's methods don't return `err`, so there's no way to propagate errors. For this reason, exported streams in the C Data Interface have no way of returning errors, either. 

Changing the interface would of course be a breaking change. The alternative is to declare this:

```
type ClosableRecordReader interface {
        RecordReader
        Closable
}
```

which gives us one place to report errors (at the end).

### Component(s)

Go

lidavidm:
### Rationale for this change

Add Err() to the RecordReader interface so we can report errors.

### Are these changes tested?

This is tested in the C Data Interface.

**This is a breaking change.**
* Closes: #33789

### DIFF
--- a/go/arrow/array/record.go
+++ b/go/arrow/array/record.go
@@ -37,6 +37,7 @@ type RecordReader interface {
 
 	Next() bool
 	Record() arrow.Record
+	Err() error
 }
 
 // simpleRecords is a simple iterator over a collection of records.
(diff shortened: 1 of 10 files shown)


zeroshade (approved):
LGTM

PR #33793, opened 2023-01-19
GH-15054: [Python] Finalize S3 on pytest exit to prevent shutdown crashes

### LINKED ISSUE #15054: [CI][Python] wheel-manylinux2014-* sometimes crashed on pytest exit

wheel-manylinux2014-cp38-amd64 2022-12-20 nighty:

https://github.com/ursacomputing/crossbow/actions/runs/3738813751/jobs/6345300403#step:9:775

```text
= 4294 passed, 382 skipped, 18 xfailed, 2 xpassed, 4 warnings in 522.05s (0:08:42) =
/arrow/ci/scripts/python_wheel_unix_test.sh: line 95:    79 Segmentation fault      (core dumped) python -m pytest -r s --pyargs pyarrow
```

wheel-manylinux2014-cp39-arm64 2022-12-20 nighty:

https://app.travis-ci.com/github/ursacomputing/crossbow/builds/259026563#L5708

```text
= 4324 passed, 352 skipped, 18 xfailed, 2 xpassed, 1 warning in 141.20s (0:02:21) =
/arrow/ci/scripts/python_wheel_unix_test.sh: line 95:    15 Segmentation fault      (core dumped) python -m pytest -r s --pyargs pyarrow
```

### Component(s)

Continuous Integration, Packaging, Python

westonpace:
For some reason destroying an S3Client at shutdown can cause a crash in our tests.  It appears to be because S3's logging is already shut down.  I wouldn't think this would happen because I would think the linker would tear down S3 shared libraries AFTER it tears down Arrow shared libraries.

However, we can also finalize S3 in conftest.py to prevent this crash.

It would still be good to understand the root cause.  Otherwise we need to advise our users to call finalize_s3.
* Closes: #15054

### DIFF
--- a/python/pyarrow/conftest.py
+++ b/python/pyarrow/conftest.py
@@ -265,3 +265,14 @@ def add_fs(doctest_namespace, request, tmp_path):
         doctest_namespace["local_path"] = str(tmp_path)
         doctest_namespace["path"] = str(path)
     yield
+
+# hookimpl(trylast=True) is required to ensure this runs after session scoped fixtures
+# which may try and use S3.  See https://github.com/pytest-dev/pytest/issues/7484
+@pytest.hookimpl(trylast=True)
+def pytest_sessionfinish(session):
+    print('Session finished, finalizing s3')
+    try:
+        from pyarrow.fs import finalize_s3
+        finalize_s3()
+    except ImportError:
+        pass
--- a/python/pyarrow/conftest.py
+++ b/python/pyarrow/conftest.py
@@ -270,7 +270,6 @@ def add_fs(doctest_namespace, request, tmp_path):
 # which may try and use S3.  See https://github.com/pytest-dev/pytest/issues/7484
 @pytest.hookimpl(trylast=True)
 def pytest_sessionfinish(session):
-    print('Session finished, finalizing s3')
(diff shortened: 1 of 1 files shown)


lidavidm:
> because I would think the linker would tear down S3 shared libraries AFTER it tears down Arrow shared libraries.

In a wheel though, we statically link everything, and then destruction order is undefined as usual?

python/pyarrow/conftest.py
```diff
@@ -265,3 +265,13 @@ def add_fs(doctest_namespace, request, tmp_path):
         doctest_namespace["local_path"] = str(tmp_path)
         doctest_namespace["path"] = str(path)
     yield
+
```
raulcd, line 268: I think the linter error is only because it expects two empty lines here.
```suggestion


```

westonpace:
> In a wheel though, we statically link everything, and then destruction order is undefined as usual?

Yes, I think that could explain it.  Still, this fix would require pyarrow users to know about and remember calling the finalize S3 method.  I'm going to pause this idea and make an attempt at an arrow::Shutdown.

westonpace:
Closing in favor of #33858

PR #33799, opened 2023-01-20
GH-33783: [C#] Update release verification to use .NET 7.0

### LINKED ISSUE #33783: [Release][C#] Release verification tasks fail with new version of C# 7.0.x

dotnet 7.0.0 was released on November and we never added new tasks for it. https://dotnet.microsoft.com/en-us/download/dotnet/7.0
We currently only have verification jobs with 6.0.202. If I run verification locally with Ubuntu 22.04 with .NET 6.0.202 C# jobs are successful:
```
===========================
Build and test C# libraries
===========================
└ Ensuring that C# is installed...
└ Installed C# at  (.NET 6.0.202)
You can invoke the tool using the following command: sourcelink
Tool 'sourcelink' (version '3.1.1') was successfully installed.
/tmp/arrow-11.0.0.ReaWC/apache-arrow-11.0.0/csharp /tmp/arrow-11.0.0.ReaWC/apache-arrow-11.0.0 ~/code/arrow
  Determining projects to restore...
```

but it fails if I upgrade dotnet to `7.0.102`:
```
===========================
Build and test C# libraries
===========================
└ Ensuring that C# is installed...
└ Found C# at  (.NET 7.0.102)

Welcome to .NET 7.0!
---------------------
SDK Version: 7.0.102
...
dev/release/verify-release-candidate.sh: line 341: 129149 Segmentation fault      (core dumped) dotnet tool install --tool-path ${csharp_bin} sourcelink
Failed to verify release candidate. See /tmp/arrow-11.0.0.lNQyX for details.
```

### Component(s)

C#, Release

westonpace:
Update verification script to download and install 7.0 but run tests against 6.0 if the 6.0 SDK is present.  Update examples to use 7.0 only.
* Closes: #33783

### DIFF
--- a/dev/release/verify-release-candidate.sh
+++ b/dev/release/verify-release-candidate.sh
@@ -823,16 +823,7 @@ test_csharp() {
 
   pushd csharp
 
-  # Have to use --list-sdks and not --list-runtimes because the
-  # flight tests use asp net core and it will only run if the SDK
-  # is present
-  if dotnet --list-sdks | grep 6\.0; then
-    dotnet test --framework net6.0
-    TESTS_RUN=1
-  fi
-
-  dotnet test --framework net7.0
-  TESTS_RUN=1
+  dotnet test
 
   if [ "${SOURCE_KIND}" = "local" -o "${SOURCE_KIND}" = "git" ]; then
     dotnet pack -c Release
--- a/dev/release/verify-release-candidate.sh
+++ b/dev/release/verify-release-candidate.sh
@@ -347,18 +347,15 @@ install_csharp() {
 
   show_info "Ensuring that C# is installed..."
 
-  if which dotnet > /dev/null 2>&1; then
+  if dotnet --version | grep 7\.0 > /dev/null 2>&1; then
     local csharp_bin=$(dirname $(which dotnet))
-    if ! which sourcelink > /dev/null 2>&1; then
-      local dotnet_tools_dir=$HOME/.dotnet/tools
-      if [ -d "${dotnet_tools_dir}" ]; then
-        PATH="${dotnet_tools_dir}:$PATH"
-      fi
-    fi
     show_info "Found C# at $(which csharp) (.NET $(dotnet --version))"
   else
+    if which dotnet > /dev/null 2>&1; then
+      show_info "dotnet found but it is the wrong version and will be ignored."
+    fi
     local csharp_bin=${ARROW_TMPDIR}/csharp/bin
-    local dotnet_version=6.0.202
+    local dotnet_version=7.0.102
     local dotnet_platform=
     case "$(uname)" in
       Linux)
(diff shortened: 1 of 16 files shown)


westonpace:
@github-actions crossbow submit --group verify-rc-source --param release=11.0.0 --param rc=0

westonpace:
@github-actions crossbow submit --group verify-rc-binaries --param release=11.0.0 --param rc=0

westonpace:
@github-actions crossbow submit --group verify-rc-binaries

westonpace:
@github-actions crossbow submit --group verify-rc-source

dev/release/verify-release-candidate.sh
```diff
@@ -347,18 +347,15 @@ install_csharp() {
 
   show_info "Ensuring that C# is installed..."
 
-  if which dotnet > /dev/null 2>&1; then
+  if dotnet --version | grep 7\.0 > /dev/null 2>&1; then
```
eerhardt, line 350: what happens if `dotnet` is not installed? Does this script error out?

eerhardt (commented):
I almost wonder if we need both 6.0 and 7.0.

Would it be better to just require 7.0 when doing development on the library?

westonpace:
> I almost wonder if we need both 6.0 and 7.0.

Are you asking if it is worth testing with 6.0?  I agree that may be excessive.  I was erring on the side of caution but if you want me to refine this so its a strict upgrade to 7.0 I'd be fine with that.

> Would it be better to just require 7.0 when doing development on the library?

You should be able to get away with doing development with just 7.0.  The only downside will be that you will need to make sure to run your tests with `dotnet test --framework net7.0` or else it will try and run them for each target and fail on 6.0 (I believe Visual Studio IDE makes this a drop down and only tests a single framework at a time).

dev/release/verify-release-candidate.sh
westonpace: The `if` fails and it downloads and installs a temporary local copy of .NET 7.

eerhardt:
By default, VS will run both:

image

westonpace:
@eerhardt circling back on this.  Would you rather we just upgrade everything to .net7 (and remove references to .net6) to keep the developer experience simple?

eerhardt:
I'm fine with having both .NET 6 and .NET 7 support, I guess, if we need it.

Maybe let's just do .NET 7 for now. And if someone finds a reason to keep supporting .NET 6, we can add it then?

westonpace:
@github-actions crossbow submit --group verify-rc-source-csharp

westonpace:
@github-actions crossbow submit --group verify-rc-source

westonpace:
> Maybe let's just do .NET 7 for now. And if someone finds a reason to keep supporting .NET 6, we can add it then?

Done.  CI failures are unrelated and the wheel verification builds seem to be passing.  I'm going to go ahead and merge this.

PR #33802, opened 2023-01-20
GH-33801: [Python] Expose C++ ExtensionTypes/ExtensionArrays in pyarrow

### LINKED ISSUE #33801: C++ Extension Types aren't correctly exposed in pyarrow

Version: master branch (11.0.0)
Platform: Ubuntu 20.04

Neither `__arrow_ext_class__`  nor `__arrow_ext_scalar_class__` are exposed on `BaseExtensionType`.

This results in the following sort of errors when trying to access a C++ ExtensionArray/ExtensionType from pyarrow:

```
AttributeError: 'pyarrow.lib.BaseExtensionType' object has no attribute '__arrow_ext_class__'
```

See the following, for example:

- https://github.com/apache/arrow/issues/32291
- https://github.com/apache/arrow/pull/10565#issuecomment-890893166

### Component(s)

Python

sjperkins:
* Closes #33801 

### Rationale for this change

C++ Extension Types are not correctly exposed in pyarrow

`__arrow_ext_class__` and `__arrow_ext_scalar_class__` have been moved from `ExtensionType` to `BaseExtensionType` in types.pxi.

### Are these changes tested?

Yes, a test has been added to `test_cython.py`. There may be better locations for this, but the existing cython testing machinery here is useful for generating a C++ extension type on the fly.

I don't believe there are any user-facing changes as `__arrow_ext_class__` and `__arrow_ext_scalar_class__` are moved into a base class.

### DIFF
--- a/python/pyarrow/tests/test_cython.py
+++ b/python/pyarrow/tests/test_cython.py
@@ -164,6 +164,40 @@ def test_cython_api(tmpdir):
 
 
 @pytest.mark.cython
+def test_extension_type(tmpdir):
+    with tmpdir.as_cwd():
+        # Set up temporary workspace
+        pyx_file = 'extensions.pyx'
+        shutil.copyfile(os.path.join(here, pyx_file),
+                        os.path.join(str(tmpdir), pyx_file))
+        # Create setup.py file
+        setup_code = setup_template.format(pyx_file=pyx_file,
+                                           compiler_opts=compiler_opts,
+                                           test_ld_path=test_ld_path)
+        with open('setup.py', 'w') as f:
+            f.write(setup_code)
+
+        subprocess_env = test_util.get_modified_env_with_pythonpath()
+
+        # Compile extension module
+        subprocess.check_call([sys.executable, 'setup.py',
+                               'build_ext', '--inplace'],
+                              env=subprocess_env)
+
+    sys.path.insert(0, str(tmpdir))
+    mod = __import__('extensions')
+
+    uuid_type = mod._make_uuid_type()
+    assert uuid_type.extension_name == "uuid"
+    assert uuid_type.storage_type == pa.binary(16)
+
+    array = mod._make_uuid_array()
+    assert array.to_pylist() == [b'abcdefghijklmno0', b'0onmlkjihgfedcba']
+    assert array[0].as_py() == b'abcdefghijklmno0'
+    assert array[1].as_py() == b'0onmlkjihgfedcba'
+
+
+@pytest.mark.cython
 def test_visit_strings(tmpdir):
     with tmpdir.as_cwd():
         # Set up temporary workspace
--- a/python/pyarrow/tests/test_cython.py
+++ b/python/pyarrow/tests/test_cython.py
@@ -163,8 +163,6 @@ def test_cython_api(tmpdir):
                               env=subprocess_env)
 
 
-
-
 @pytest.mark.cython
 def test_visit_strings(tmpdir):
     with tmpdir.as_cwd():
--- a/python/pyarrow/tests/test_cython.py
+++ b/python/pyarrow/tests/test_cython.py
@@ -163,38 +163,6 @@ def test_cython_api(tmpdir):
                               env=subprocess_env)
 
 
-@pytest.mark.cython
-def test_extension_type(tmpdir):
-    with tmpdir.as_cwd():
-        # Set up temporary workspace
-        pyx_file = 'extensions.pyx'
-        shutil.copyfile(os.path.join(here, pyx_file),
-                        os.path.join(str(tmpdir), pyx_file))
-        # Create setup.py file
-        setup_code = setup_template.format(pyx_file=pyx_file,
-                                           compiler_opts=compiler_opts,
-                                           test_ld_path=test_ld_path)
-        with open('setup.py', 'w') as f:
-            f.write(setup_code)
-
-        subprocess_env = test_util.get_modified_env_with_pythonpath()
-
-        # Compile extension module
-        subprocess.check_call([sys.executable, 'setup.py',
-                               'build_ext', '--inplace'],
-                              env=subprocess_env)
-
-    sys.path.insert(0, str(tmpdir))
-    mod = __import__('extensions')
-
-    uuid_type = mod._make_uuid_type()
-    assert uuid_type.extension_name == "uuid"
-    assert uuid_type.storage_type == pa.binary(16)
-
-    array = mod._make_uuid_array()
-    assert array.to_pylist() == [b'abcdefghijklmno0', b'0onmlkjihgfedcba']
-    assert array[0].as_py() == b'abcdefghijklmno0'
-    assert array[1].as_py() == b'0onmlkjihgfedcba'
 
 
 @pytest.mark.cython
--- a/python/pyarrow/types.pxi
+++ b/python/pyarrow/types.pxi
@@ -836,6 +836,27 @@ cdef class BaseExtensionType(DataType):
         DataType.init(self, type)
         self.ext_type = <const CExtensionType*> type.get()
 
+    def __arrow_ext_class__(self):
+        """Return an extension array class to be used for building or
+        deserializing arrays with this extension type.
+
+        This method should return a subclass of the ExtensionArray class. By
+        default, if not specialized in the extension implementation, an
+        extension type array will be a built-in ExtensionArray instance.
+        """
+        return ExtensionArray
+
+    def __arrow_ext_scalar_class__(self):
+        """Return an extension scalar class for building scalars with this
+        extension type.
+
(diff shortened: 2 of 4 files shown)


python/pyarrow/tests/test_cython.py
```diff
@@ -163,6 +163,40 @@ def test_cython_api(tmpdir):
...
+    mod = __import__('extensions')
+
+    uuid_type = mod._make_uuid_type()
+    assert uuid_type.extension_name == "uuid"
+    assert uuid_type.storage_type == pa.binary(16)
+
+    array = mod._make_uuid_array()
+    assert array.to_pylist() == [b'abcdefghijklmno0', b'0onmlkjihgfedcba']
+    assert array[0].as_py() == b'abcdefghijklmno0'
+    assert array[1].as_py() == b'0onmlkjihgfedcba'
```
sjperkins, line 197: Check that we can access extension array as list and it's scalar values.

python/pyarrow/types.pxi
```diff
@@ -836,6 +836,26 @@ cdef class BaseExtensionType(DataType):
...
 
+    def __arrow_ext_class__(self):
+        """Return an extension array class to be used for building or
+        deserializing arrays with this extension type.
+
+        This method should return a subclass of the ExtensionArray class. By
+        default, if not specialized in the extension implementation, an
+        extension type array will be a built-in ExtensionArray instance.
+        """
+        return ExtensionArray
```
sjperkins, line 847: It occurred to me that perhaps we should be generating dummy ExtensionArray classes, perhaps from the `extension_name`? However, this does seem to just work.

python/pyarrow/tests/test_cython.py
```diff
@@ -163,6 +163,40 @@ def test_cython_api(tmpdir):
                               env=subprocess_env)
 
 
+@pytest.mark.cython
+def test_extension_type(tmpdir):
```
sjperkins, line 167: This isn't strictly a cython test, but the cython machinery is useful for generating a C++ Extension Type on the fly. Perhaps this test should be moved elsewhere?

python/pyarrow/tests/test_cython.py
```diff
@@ -163,6 +163,40 @@ def test_cython_api(tmpdir):
...
+                              env=subprocess_env)
+
+    sys.path.insert(0, str(tmpdir))
+    mod = __import__('extensions')
+
+    uuid_type = mod._make_uuid_type()
+    assert uuid_type.extension_name == "uuid"
+    assert uuid_type.storage_type == pa.binary(16)
+
+    array = mod._make_uuid_array()
```
jorisvandenbossche, line 194: Can you add an assert of that the resulting array is an instance of BaseExtensionType? (and maybe that the `array.type == uuid_type`)

python/pyarrow/types.pxi
jorisvandenbossche: > However, this does seem to just work.

Yes, this works as long as you are fine with the base implementation of that class (eg it's implementation of to_numpy, etc). 
Once you want to customize that, you'll need a custom class, but I am not sure if that's straightforward to do at the moment ("register" such a custom class for extension type that is actually implemented in C++). In any case, for this PR using ExtensionArray is fine (since a dummy generated class would not give anything of additional functionality, just a different name of the class)

python/pyarrow/types.pxi
```diff
@@ -968,27 +988,6 @@ cdef class ExtensionType(BaseExtensionType):
         """
         return NotImplementedError
 
-    def __arrow_ext_class__(self):
```
jorisvandenbossche, line 971: I am wondering if we would like to keep this here as well. It would give some duplication, but for someone looking at `ExtensionType` for implementing a subclass yourself, it's nice that the 4 special dunder methods are all together, and not spread between this class and the base class. 

(to avoid duplication, we can maybe leave out the docstring on the base class?)

python/pyarrow/tests/test_cython.py
jorisvandenbossche: I think it is fine either way. It could also be moved to `test_extension_type.py`, and then you can import some of the needed utilities from test_cython.py.

python/pyarrow/tests/test_cython.py
```diff
@@ -163,6 +163,40 @@ def test_cython_api(tmpdir):
...
+    assert array.to_pylist() == [b'abcdefghijklmno0', b'0onmlkjihgfedcba']
+    assert array[0].as_py() == b'abcdefghijklmno0'
+    assert array[1].as_py() == b'0onmlkjihgfedcba'
+
+
```
jorisvandenbossche, line 199: I am not sure it's necessarily needed, but another test you could add is put this extension array in a RecordBatch, send it through IPC to ensure it's still the correct extension type afterwards. There are some helpers to make this easier in `test_extension_type.py` (see the `ipc_write_batch`/`ipc_read_batch` usage)

This would require exposing an additional function in cython wrapped C++ snippet to register the type.

jorisvandenbossche (commented):
Thanks a lot for this PR! (and sorry for the slow feedback)

Looking good, added some comments.

jorisvandenbossche:
@sjperkins FYI I also opened an issue about having custom subclasses for extension types implemented in C++ (instead of using the base class BaseExtensionType/ExtensionArray): https://github.com/apache/arrow/issues/33997

sjperkins:
> @sjperkins FYI I also opened an issue about having custom subclasses for extension types implemented in C++ (instead of using the base class BaseExtensionType/ExtensionArray): #33997

Thanks @jorisvandenbossche for mentioning this, as well as the review. I'm intending to address your review comments next week.

python/pyarrow/tests/test_cython.py
sjperkins: Moved to `test_extension_type.py`

python/pyarrow/tests/test_cython.py
sjperkins: Done

python/pyarrow/types.pxi
sjperkins: Moved `__arrow_ext_class__` and `__arrow_ext_scalar__` dunder methods to the `ExtensionType` class and left minimally documented implementations on `BaseExtensionType`.

python/pyarrow/tests/test_cython.py
sjperkins: > I am not sure it's necessarily needed, but another test you could add is put this extension array in a RecordBatch, send it through IPC to ensure it's still the correct extension type afterwards. There are some helpers to make this easier in `test_extension_type.py` (see the `ipc_write_batch`/`ipc_read_batch` usage)

Added an IPC serialisation test here, which does fail. I'll discuss further down.

> This would require exposing an additional function in cython wrapped C++ snippet to register the type.

Which additional function is required here? I wonder if this is possible at all?

The reason for this is that default implementations  for the `__arrow_ext_class__` and `__arrow_ext_scalar__` methods are provided on `BaseExtensionType`,  but not the `__arrow_ext_serialize__` and `__arrow_ext_deserialize__` methods. These would ideally call into the C++ `Serialize/Deserialize` methods but its not immediately clear how this can be accomplished at the moment without implementing https://github.com/apache/arrow/issues/33997 in some form.

sjperkins:
Apologies for triggering the assigned reviewer and label bots with the force pushes.

python/pyarrow/tests/test_cython.py
sjperkins: Perhaps the way forward is to expose these methods in `CExtensionType` in `libarrow.pxd`?

sjperkins: Using `pa.register_extension_type(uuid_type)` produces the desired output.

```python
    uuid_type = mod._make_uuid_type()
    ...

    pa.register_extension_type(uuid_type)

    buf = ipc_write_batch(pa.RecordBatch.from_arrays([array], ["uuid"]))
    del array

    batch = ipc_read_batch(buf)
    array = batch.column(0)
    assert array.type == uuid_type
```

sjperkins: Ah, had to register it at the C++ layer (obviously!). With https://github.com/apache/arrow/pull/33802/commits/a17e926162c9ed6f3abb5a0b5b42e35f1fa79bc0, `test_cpp_extension_in_python` runs through fine.

python/pyarrow/tests/test_cython.py
sjperkins: Resolving as I'm assuming this is OK

sjperkins:
Is there anything else needed from me on this PR?

jorisvandenbossche (approved):
No, nothing else needed, this looks good! (and thanks for the ping :))

sjperkins:
Thanks for the review and merging @jorisvandenbossche

PR #33806, opened 2023-01-20
GH-33723: [C++] re2::RE2::RE2() result must be checked

### LINKED ISSUE #33723: [C++] re2::RE2::RE2() result must be checked

`re2::RE2::RE2()` may be failed. We should check the `re2::RE2::RE2()` result with `re2::RE2::ok()`.

For example:

```diff
diff --git a/cpp/src/arrow/compute/kernels/scalar_string_ascii.cc b/cpp/src/arrow/compute/kernels/scalar_string_ascii.cc
index d3d0ac3201..b2b9d47c02 100644
--- a/cpp/src/arrow/compute/kernels/scalar_string_ascii.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_string_ascii.cc
@@ -1681,6 +1681,10 @@ struct FindSubstringRegex {
 
   template <typename OutValue, typename... Ignored>
   OutValue Call(KernelContext*, std::string_view val, Status*) const {
+    if (!regex_match_->ok()) {
+      // TODO: Report error
+      return -1;
+    }
     re2::StringPiece piece(val.data(), val.length());
     re2::StringPiece match;
     if (RE2::PartialMatch(piece, *regex_match_, &match)) {
```

Gandiva also doesn't check `re2::RE2::RE2()` result.

If `re2::RE2::RE2()` is failed, a program is crashed like #25633 .

### Component(s)

C++, C++ - Gandiva

js8544:
### Rationale for this change

RE2 construction needs to be checked.

Check all RE2 object status.

### Are these changes tested?

Tested with existing tests.

No.
* Closes: #33723

### DIFF
--- a/cpp/src/arrow/compute/kernels/scalar_string_ascii.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_string_ascii.cc
@@ -20,8 +20,6 @@
 #include <iterator>
 #include <memory>
 #include <string>
-#include "arrow/result.h"
-#include "arrow/util/macros.h"
 
 #ifdef ARROW_WITH_RE2
 #include <re2/re2.h>
@@ -29,6 +27,8 @@
 
 #include "arrow/array/builder_nested.h"
 #include "arrow/compute/kernels/scalar_string_internal.h"
+#include "arrow/result.h"
+#include "arrow/util/macros.h"
 #include "arrow/util/string.h"
 #include "arrow/util/value_parsing.h"
 
--- a/cpp/src/arrow/compute/kernels/scalar_string_ascii.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_string_ascii.cc
@@ -21,6 +21,7 @@
 #include <memory>
 #include <string>
 #include "arrow/result.h"
+#include "arrow/util/macros.h"
 
 #ifdef ARROW_WITH_RE2
 #include <re2/re2.h>
@@ -1503,13 +1504,17 @@ struct MatchLike {
     static const RE2::Options kRE2Options = MakeRE2Options<StringType>();
     // A LIKE pattern matching this regex can be translated into a substring search.
     static const RE2 kLikePatternIsSubstringMatch(R"(%+([^%_]*[^\\%_])?%+)", kRE2Options);
-    RETURN_NOT_OK(RegexStatus(kLikePatternIsSubstringMatch));
     // A LIKE pattern matching this regex can be translated into a prefix search.
     static const RE2 kLikePatternIsStartsWith(R"(([^%_]*[^\\%_])?%+)", kRE2Options);
-    RETURN_NOT_OK(RegexStatus(kLikePatternIsStartsWith));
     // A LIKE pattern matching this regex can be translated into a suffix search.
     static const RE2 kLikePatternIsEndsWith(R"(%+([^%_]*))", kRE2Options);
-    RETURN_NOT_OK(RegexStatus(kLikePatternIsEndsWith));
+    static bool global_checked = false;
+    if (ARROW_PREDICT_FALSE(!global_checked)) {
+      RETURN_NOT_OK(RegexStatus(kLikePatternIsSubstringMatch));
+      RETURN_NOT_OK(RegexStatus(kLikePatternIsStartsWith));
+      RETURN_NOT_OK(RegexStatus(kLikePatternIsEndsWith));
+      global_checked = true;
+    }
 
     auto original_options = MatchSubstringState::Get(ctx);
     auto original_state = ctx->state();
--- a/cpp/src/arrow/compute/kernels/scalar_string_ascii.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_string_ascii.cc
@@ -18,7 +18,9 @@
 #include <algorithm>
 #include <cctype>
 #include <iterator>
+#include <memory>
 #include <string>
+#include "arrow/result.h"
 
 #ifdef ARROW_WITH_RE2
 #include <re2/re2.h>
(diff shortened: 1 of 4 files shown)


cpp/src/arrow/compute/kernels/scalar_string_ascii.cc
```diff
@@ -1505,6 +1508,13 @@ struct MatchLike {
     static const RE2 kLikePatternIsStartsWith(R"(([^%_]*[^\\%_])?%+)", kRE2Options);
     // A LIKE pattern matching this regex can be translated into a suffix search.
     static const RE2 kLikePatternIsEndsWith(R"(%+([^%_]*))", kRE2Options);
+    static bool global_checked = false;
+    if (ARROW_PREDICT_FALSE(!global_checked)) {
+      RETURN_NOT_OK(RegexStatus(kLikePatternIsSubstringMatch));
+      RETURN_NOT_OK(RegexStatus(kLikePatternIsStartsWith));
+      RETURN_NOT_OK(RegexStatus(kLikePatternIsEndsWith));
+      global_checked = true;
+    }
```
js8544, line 1517: Are there better ways to check global RE2 objects?

cpp/src/gandiva/regex_functions_holder.cc
```diff
@@ -34,27 +31,44 @@ std::string& RemovePatternEscapeChars(const FunctionNode& node, std::string& pat
...
+  static const RE2 is_substr_regex(R"(\.\*([^\.\*])*\.\*)");
+
+  static bool global_checked = false;
+  if (ARROW_PREDICT_FALSE(!global_checked)) {
+    if (ARROW_PREDICT_FALSE(
+            !(starts_with_regex.ok() && ends_with_regex.ok() && is_substr_regex.ok()))) {
+      return node;
+    }
+    global_checked = true;
+  }
```
js8544, line 49: Are there better ways to check global RE2 objects?

cpp/src/arrow/compute/kernels/scalar_string_ascii.cc
```diff
@@ -1505,6 +1508,13 @@ struct MatchLike {
     static const RE2 kLikePatternIsStartsWith(R"(([^%_]*[^\\%_])?%+)", kRE2Options);
     // A LIKE pattern matching this regex can be translated into a suffix search.
     static const RE2 kLikePatternIsEndsWith(R"(%+([^%_]*))", kRE2Options);
+    static bool global_checked = false;
```
kou, line 1511: Can we use `std::once_flag`?

cpp/src/arrow/compute/kernels/scalar_string_ascii.cc
```diff
@@ -1505,6 +1508,13 @@ struct MatchLike {
...
+    if (ARROW_PREDICT_FALSE(!global_checked)) {
+      RETURN_NOT_OK(RegexStatus(kLikePatternIsSubstringMatch));
+      RETURN_NOT_OK(RegexStatus(kLikePatternIsStartsWith));
+      RETURN_NOT_OK(RegexStatus(kLikePatternIsEndsWith));
```
kou, line 1515: If we have an error in these patterns, does this code report an error on the second call? (It seems that this code reports an error only on the first call.)

kou (commented):
Could you also add tests for error pattern cases?

cpp/src/arrow/compute/kernels/scalar_string_ascii.cc
js8544: I thought about using `std::call_once` but the only solution I can come up with uses exceptions:
```cpp
    static std::once_flag global_checked;
    try {
      std::call_once(global_checked, [] {
        auto check_global = [] {
          RETURN_NOT_OK(RegexStatus(kLikePatternIsSubstringMatch));
          RETURN_NOT_OK(RegexStatus(kLikePatternIsStartsWith));
          RETURN_NOT_OK(RegexStatus(kLikePatternIsEndsWith));
        };
        auto status = check_global();
        if (!status.ok()) { // throw an exception so that once_flag is not set
          throw std::invalid_argument(status.ToString());
        }
      });
    } catch (const std::invalid_argument& e) {
      return Status::Invalid(e.what());
    }
```
I couldn't find a more elegant solution. WDYT?

js8544: I also feel like a static boolean is enough for our case. As it avoids false positives, i.e., erroneous regexes are always checked.

js8544:
> Could you also add tests for error pattern cases?

Added a TestBaseBinaryKernels: FindSubstringRegexWrongPattern test

cpp/src/arrow/compute/kernels/scalar_string_ascii.cc
kou: Ah, I see. `std::call_once` isn't suitable here.

cpp/src/arrow/compute/kernels/scalar_string_ascii.cc
kou: I see. Thanks.

kou (approved):
+1

cpp/src/arrow/compute/kernels/scalar_string_test.cc
```diff
@@ -19,6 +19,7 @@
 #include <string>
 #include <utility>
 #include <vector>
+#include "arrow/compute/exec.h"
```
westonpace, line 22: Can you keep the includes grouped together?

cpp/src/gandiva/regex_functions_holder.cc
```diff
@@ -17,15 +17,12 @@
 
 #include "gandiva/regex_functions_holder.h"
 #include <regex>
+#include "arrow/util/macros.h"
```
westonpace, line 20: I am not very familiar with the gandiva code base.  Is it ok/normal to use Arrow macros here?  I assume so.

westonpace (approved):
One minor include nit would be nice to fix.

cpp/src/arrow/compute/kernels/scalar_string_test.cc
js8544: done

cpp/src/gandiva/regex_functions_holder.cc
js8544: Yep. Gandiva uses various headers in the arrow/util directory.

PR #13294, opened 2022-06-02
ARROW-16730: [C++] Bump vendored jemalloc version

pitrou:
Bump to 5.3.0, which promises performance improvements.

PR #33808, opened 2023-01-20
GH-20272: [C++] Bump version of bundled AWS SDK

### LINKED ISSUE #20272: [C++] Bump version of bundled AWS SDK
The latest version on the 1.8 line is 1.8.186.
We could also try to switch to the 1.9 line, but there were blocking issues last time we tried.

Note we should bump the dependent AWS library versions at the same time.

**Reporter**: [Antoine Pitrou](https://issues.apache.org/jira/browse/ARROW-16727) / @pitrou

<sub>**Note**: *This issue was originally created as [ARROW-16727](https://issues.apache.org/jira/browse/ARROW-16727). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

js8544:
### Rationale for this change

Bump AWS SDK version to 1.10.55.

Bump AWS SDK version to 1.10.55.

* Closes: #20272


cpp/cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -4728,8 +4729,9 @@ macro(build_awssdk)
                                               "10")
     # Workaround for https://github.com/aws/aws-sdk-cpp/issues/1750
     set(AWSSDK_PATCH_COMMAND "sed" "-i.bak" "-e" "s/\"-Werror\"//g"
-                             "<SOURCE_DIR>/cmake/compiler_settings.cmake")
+                             "<SOURCE_DIR>/cmake/compiler_settings.cmake" "&&")
```
kou, line 4732: Can we remove this patch?
It seems that 1.10.55 includes the fix for https://github.com/aws/aws-sdk-cpp/issues/1750 .

cpp/cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -4728,8 +4729,9 @@ macro(build_awssdk)
...
   endif()
+  list(APPEND AWSSDK_PATCH_COMMAND "${AWSSDK_SOURCE}/prefetch_crt_dependency.sh")
```
kou, line 4734: Can we use `<SOURCE_DIR>`?

```suggestion
  list(APPEND AWSSDK_PATCH_COMMAND "<SOURCE_DIR>/prefetch_crt_dependency.sh")
```

See https://cmake.org/cmake/help/latest/module/ExternalProject.html for `<SOURCE_DIR>`.

kou: BTW, I think that we should not use this. If we use this, this code doesn't work on Windows. (Windows doesn't provide `sh`.)

cpp/cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -4643,7 +4644,7 @@ macro(build_awssdk)
   set(AWSSDK_CMAKE_ARGS
       ${AWSSDK_COMMON_CMAKE_ARGS}
       -DOPENSSL_ROOT_DIR=${OPENSSL_ROOT_HINT}
-      -DBUILD_DEPS=OFF
+      -DBUILD_DEPS=ON # We need to build aws-crt-cpp from source
```
kou, line 4647: Can we build `aws-crt-cpp` by ourselves instead of using this and `prefetch_crt_dependency.sh` like other dependencies such as `aws-c-common` for portability?

js8544: I see. I'll try building it manually.

cpp/cmake_modules/ThirdpartyToolchain.cmake
js8544: done

js8544:
I've changed to building aws-crt-cpp manually. It now works on my ubuntu machine. Let's see if the CI checks pass.

cpp/cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -4673,7 +4761,19 @@ macro(build_awssdk)
...
+      aws-c-cal
+      aws-c-compression
+      aws-c-http
+      aws-c-io
+      aws-c-mqtt
+      aws-c-s3
+      aws-c-sdkutils
+      s2n
+      crypto
+      ssl
```
kou, line 4775: Are they artifacts by `s2n-tls`?
Then we should not create `AWS::crypto` and `AWS::ssl`. We should create `AWS::s2n-tls` and associate `libcrypto.a` and `libssl.a` with `AWS::s2n-tls`.

cpp/cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -4714,35 +4814,145 @@ macro(build_awssdk)
...
                       URL_HASH "SHA256=${ARROW_AWSSDK_BUILD_SHA256_CHECKSUM}"
                       CMAKE_ARGS ${AWSSDK_CMAKE_ARGS}
-                      PATCH_COMMAND ${AWSSDK_PATCH_COMMAND}
                       BUILD_BYPRODUCTS ${AWS_CPP_SDK_COGNITO_IDENTITY_STATIC_LIBRARY}
                                        ${AWS_CPP_SDK_CORE_STATIC_LIBRARY}
                                        ${AWS_CPP_SDK_IDENTITY_MANAGEMENT_STATIC_LIBRARY}
                                        ${AWS_CPP_SDK_S3_STATIC_LIBRARY}
                                        ${AWS_CPP_SDK_STS_STATIC_LIBRARY}
-                      DEPENDS aws_c_event_stream_ep)
+                      DEPENDS aws_c_event_stream_ep aws_crt_cpp_ep)
```
kou, line 4955: Can we remove `aws_c_event_stream_ep` here because `aws_crt_cpp_ep` depends on `aws_c_event_stream_ep`?

```suggestion
                      DEPENDS aws_crt_cpp_ep)
```

cpp/thirdparty/versions.txt
```diff
@@ -107,6 +129,17 @@ DEPENDENCIES=(
...
+  "ARROW_AWS_C_HTTP_URL aws-c-http-${ARROW_AWS_C_HTTP_BUILD_VERSION}.tar.gz https://github.com/awslabs/aws-c-http/archive/${ARROW_AWS_C_HTTP_BUILD_VERSION}.tar.gz"
+  "ARROW_AWS_C_IO_URL aws-c-io-${ARROW_AWS_C_IO_BUILD_VERSION}.tar.gz https://github.com/awslabs/aws-c-io/archive/${ARROW_AWS_C_IO_BUILD_VERSION}.tar.gz"
+  "ARROW_AWS_C_MQTT_URL aws-c-mqtt-${ARROW_AWS_C_MQTT_BUILD_VERSION}.tar.gz https://github.com/awslabs/aws-c-mqtt/archive/${ARROW_AWS_C_MQTT_BUILD_VERSION}.tar.gz"
+  "ARROW_AWS_C_S3_URL aws-c-s3-${ARROW_AWS_C_S3_BUILD_VERSION}.tar.gz https://github.com/awslabs/aws-c-s3/archive/${ARROW_AWS_C_S3_BUILD_VERSION}.tar.gz"
+  "ARROW_AWS_C_SDKUTILS_URL aws-c-sdkutils-${ARROW_AWS_C_SDKUTILS_BUILD_VERSION}.tar.gz https://github.com/awslabs/aws-c-sdkutils/archive/${ARROW_AWS_C_SDKUTILS_BUILD_VERSION}.tar.gz"
+  "ARROW_AWS_LC_URL aws-lc-${ARROW_AWS_LC_BUILD_VERSION}.tar.gz https://github.com/awslabs/aws-lc/archive/${ARROW_AWS_LC_BUILD_VERSION}.tar.gz"
+  "ARROW_S2N_URL s2n-${ARROW_S2N_BUILD_VERSION}.tar.gz https://github.com/awslabs/s2n/archive/${ARROW_S2N_BUILD_VERSION}.tar.gz"
```
kou, line 141: Could you keep alphabetical order?

cpp/cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -4612,6 +4699,7 @@ endif()
 macro(build_awssdk)
   message(STATUS "Building AWS C++ SDK from source")
   set(AWSSDK_PREFIX "${CMAKE_CURRENT_BINARY_DIR}/awssdk_ep-install")
+  set(AWSSDK_SOURCE "${CMAKE_CURRENT_BINARY_DIR}/awssdk_ep-prefix/src/awssdk_ep")
```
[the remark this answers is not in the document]
js8544: done

cpp/cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -455,6 +455,93 @@ else()
...
+           "https://github.com/awslabs/aws-lc/archive/${ARROW_AWS_LC_BUILD_VERSION}.tar.gz"
+  )
+endif()
+
+if(DEFINED ENV{ARROW_S2N_URL})
+  set(S2N_SOURCE_URL "$ENV{ARROW_S2N_URL}")
+else()
+  set_urls(S2N_SOURCE_URL
+           "https://github.com/awslabs/s2n/archive/${ARROW_S2N_BUILD_VERSION}.tar.gz")
+endif()
```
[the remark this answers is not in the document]
js8544: Interestingly the archive files are stored in s2n not s2n-tls. https://github.com/awslabs/s2n-tls/archive/v1.3.27.tar.gz is not found but https://github.com/awslabs/s2n/archive/v1.3.27.tar.gz is.

cpp/cmake_modules/ThirdpartyToolchain.cmake
js8544: It turns out they are not needed. I've removed these two.

cpp/cmake_modules/ThirdpartyToolchain.cmake
js8544: done

cpp/thirdparty/versions.txt
js8544: done

js8544:
I am stuck on [this problem](https://app.travis-ci.com/github/apache/arrow/jobs/594451329) for some time. Due to [this](https://github.com/apache/arrow/blob/master/cpp/src/arrow/filesystem/CMakeLists.txt#L63), we need to initialize AWS SDK twice in arrow-s3fs-test because AWS SDK is statically linked. With the new 1.10 version, its initialization involves adding an OpenSSL engine [here](https://github.com/aws/s2n-tls/blob/main/utils/s2n_random.c#L388). The problem is, OpenSSL is dynamically linked, so the second initialization will fail due to a name conflict. The problem is the same as [this issue](https://github.com/aws/s2n-tls/issues/3445). 

In short, AWS SDK is statically linked but OpenSSL is dynamically linked, so we can't init AWS twice as we have been done in arrow-s3fs-test.

What is the best way to resolve this? Any help is appreciated. cc @kou @pitrou

pitrou:
I think we should simply switch to dynamic linking for the AWS SDK. Static linking is fine for leaf dependencies (such as lz4), but it should be unsupported for complex dependencies such as AWS, gRPC, etc.

pitrou:
> If we use dynamic linking for bundled libraries, we need to install them but it may overwrite existing libraries.

Well, yes, but so what? People who are concerned about that should use their distribution's packages.

js8544:
I found out that S2N provides an option `S2N_INTERN_LIBCRYPTO` to statically link against and "internalize" libcrypto. With [this option on](https://github.com/apache/arrow/pull/33808/commits/1b9bcdc54fe9a319b03fa1bdb51549c980915716) all tests are passing. We can stay with static linkage for now :). cc @kou @pitrou

kou:
> > If we use dynamic linking for bundled libraries, we need to install them but it may overwrite existing libraries.
> 
> Well, yes, but so what? People who are concerned about that should use their distribution's packages.

I think that this is a bad manner. But I just realized that I have never discussed this before. My concern may be overthinking.

cpp/cmake_modules/ThirdpartyToolchain.cmake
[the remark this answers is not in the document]
kou: `awslabs` -> `aws`: https://github.com/aws/s2n-tls/archive/v1.3.27.tar.gz

Could you use `S2N_TLS` instead of `S2N` like `ARROW_S2N_TLS_URL` for all s2n-tls related variables?

BTW, the latest release is 1.3.35: https://github.com/aws/s2n-tls/releases
Can we use the latest version?

cpp/cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -4714,35 +4798,129 @@ macro(build_awssdk)
...
+  add_dependencies(AWS::aws-c-io aws_c_io_ep)
+
   externalproject_add(aws_c_event_stream_ep
                       ${EP_COMMON_OPTIONS}
                       URL ${AWS_C_EVENT_STREAM_SOURCE_URL}
                       URL_HASH "SHA256=${ARROW_AWS_C_EVENT_STREAM_BUILD_SHA256_CHECKSUM}"
                       CMAKE_ARGS ${AWSSDK_COMMON_CMAKE_ARGS}
                       BUILD_BYPRODUCTS ${AWS_C_EVENT_STREAM_STATIC_LIBRARY}
-                      DEPENDS aws_checksums_ep)
+                      DEPENDS aws_checksums_ep aws_c_common_ep aws_c_io_ep)
```
kou, line 4839: Can we remove `aws_c_common_ep` here because it seems that `aws_c_io_ep` depends on `awx_c_common_ep`:

```suggestion
                      DEPENDS aws_checksums_ep aws_c_io_ep)
```

cpp/cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -4714,35 +4798,129 @@ macro(build_awssdk)
...
+                      DEPENDS aws_c_io_ep aws_c_compression_ep)
+  add_dependencies(AWS::aws-c-http aws_c_http_ep)
+
+  externalproject_add(aws_c_mqtt_ep
+                      ${EP_COMMON_OPTIONS}
+                      URL ${AWS_C_MQTT_SOURCE_URL}
+                      URL_HASH "SHA256=${ARROW_AWS_C_MQTT_BUILD_SHA256_CHECKSUM}"
+                      CMAKE_ARGS ${AWSSDK_COMMON_CMAKE_ARGS}
+                      BUILD_BYPRODUCTS ${AWS_C_MQTT_STATIC_LIBRARY}
+                      DEPENDS aws_c_io_ep aws_c_http_ep)
```
kou, line 4875:
```suggestion
                      DEPENDS aws_c_http_ep)
```

kou (commented):
Thanks!
I think that we can merge this only with some more small cleanups.

cpp/cmake_modules/ThirdpartyToolchain.cmake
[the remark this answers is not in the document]
kou: @js854 Could you also confirm the above comment?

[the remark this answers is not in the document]
js8544: @kou Oh sorry I missed it. I am looking at it now.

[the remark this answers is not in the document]
js8544: @kou Changed in #abfa9774c

kou:
@github-actions crossbow submit -g cpp -g r

kou:
@github-actions crossbow submit -g cpp -g r

kou:
@github-actions crossbow submit -g cpp -g r

kou:
@github-actions crossbow submit -g cpp -g r

kou (approved):
+1

I've disabled arrow-s3fs-test with bundled aws-sdk-cpp.

kou:

assignUser:
@pitrou @kou When updating bundled dependencies please make sure to also update our fallback version in the artifactory to avoid something like https://github.com/apache/arrow/issues/34120#issuecomment-1426080976 this was missed in the recent round of version bumps (e.g. https://github.com/apache/arrow/pull/13294).

kou:
Oh, sorry.
We need a CI job for fallback cases. We can add an environment variable or a CMake variable to force using fallback cases for the job. Could you open an issue for this?

And we should have a script to mirror upstream archives to Artifactory. Could you open an issue for this too?

assignUser:
I have updated the missing files for now and opened a few issues: https://github.com/apache/arrow/issues/34131 https://github.com/apache/arrow/issues/34132 https://github.com/apache/arrow/issues/34129

PR #33810, opened 2023-01-20
GH-33377: [Python] Table.drop should support passing a single column

### LINKED ISSUE #33377: [Python] Table.drop should support passing a single column
```Java

----> 2 data = dataset.drop("Churn") 

/usr/local/lib/python3.7/dist-packages/pyarrow/table.pxi in pyarrow.lib.Table.drop()
    KeyError: "Column 'C' not found"
```

Also, for consistency, it would probably be good to have a `Table.drop_column` alias, as all the other methods are named `Table.add_column`, `Table.append_column` and `Table.set_column`

**Reporter**: [Alessandro Molina](https://issues.apache.org/jira/browse/ARROW-18189) / @amol-

<sub>**Note**: *This issue was originally created as [ARROW-18189](https://issues.apache.org/jira/browse/ARROW-18189). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

danepitkin:
### Rationale for this change

Provide a better user experience in pyarrow when working with `Table`.

Allow `Table.drop()` to accept a single column name as a `str` argument.
Provide a wrapper `Table.drop_column(str)`, which calls `Table.drop()`, to match similar APIs such as `add_column()`, `append_column()`.

### Are these changes tested?

Updated the pytest for `Table.drop()` and added a pytest for `Table.drop_column()`. Verified both test cases ran successfully locally.

Yes, a new pyarrow API is added. The existing pyarrow API is backwards compatible, but does support additional types of input now (str). The doc strings are updated so I assume the python API reference will be auto-updated somehow? Let me know if this is not the case.

* Closes: #33377

### DIFF
--- a/python/pyarrow/table.pxi
+++ b/python/pyarrow/table.pxi
@@ -4664,7 +4664,7 @@ cdef class Table(_PandasConvertible):
 
         return pyarrow_wrap_table(c_table)
 
-    def drop(self, columns):
+    def drop_columns(self, columns):
         """
         Drop one or more columns and return a new table.
 
@@ -4693,7 +4693,7 @@ cdef class Table(_PandasConvertible):
 
         Drop one column:
 
-        >>> table.drop("animals")
+        >>> table.drop_columns("animals")
         pyarrow.Table
         n_legs: int64
         ----
@@ -4701,7 +4701,7 @@ cdef class Table(_PandasConvertible):
 
         Drop one or more columns:
 
-        >>> table.drop(["n_legs", "animals"])
+        >>> table.drop_columns(["n_legs", "animals"])
         pyarrow.Table
         ...
         ----
@@ -4725,34 +4725,20 @@ cdef class Table(_PandasConvertible):
 
         return table
 
-    def drop_column(self, name):
+    def drop(self, columns):
         """
-        Create new Table with the named column removed.
+        Alias of Table.drop_columns for backwards compatibility.
 
-        Parameters
-        ----------
-        name : str
-            Name of column to remove.
-
-        Returns
-        -------
-        Table
-            New table without the column.
-
-        Examples
+        Warnings
         --------
-        >>> import pyarrow as pa
-        >>> import pandas as pd
-        >>> df = pd.DataFrame({'n_legs': [2, 4, 5, 100],
-        ...                    'animals': ["Flamingo", "Horse", "Brittle stars", "Centipede"]})
-        >>> table = pa.Table.from_pandas(df)
-        >>> table.drop_column('animals')
-        pyarrow.Table
-        n_legs: int64
-        ----
-        n_legs: [[2,4,5,100]]
+        This API is depreacted in favor of Table.drop_columns.
         """
-        return self.drop(name)
+        import warnings
+        warnings.warn(
+            "Table.drop is deprecated, use Table.drop_columns.",
+            DeprecationWarning)
+
+        return self.drop_column(columns)
 
     def group_by(self, keys):
         """Declare a grouping over the columns of the table.
--- a/python/pyarrow/table.pxi
+++ b/python/pyarrow/table.pxi
@@ -4670,18 +4670,18 @@ cdef class Table(_PandasConvertible):
 
         Parameters
         ----------
-        columns : list of str
-            List of field names referencing existing columns.
+        columns : str or list[str]
+            Field name(s) referencing existing column(s).
 
         Raises
         ------
         KeyError
-            If any of the passed columns name are not existing.
+            If any of the passed column names do not exist.
 
         Returns
         -------
         Table
-            New table without the columns.
+            New table without the column(s).
 
         Examples
         --------
@@ -4693,19 +4693,22 @@ cdef class Table(_PandasConvertible):
 
         Drop one column:
 
-        >>> table.drop(["animals"])
+        >>> table.drop("animals")
         pyarrow.Table
         n_legs: int64
         ----
         n_legs: [[2,4,5,100]]
 
-        Drop more columns:
+        Drop one or more columns:
 
         >>> table.drop(["n_legs", "animals"])
         pyarrow.Table
         ...
         ----
         """
+        if isinstance(columns, str):
+            columns = [columns]
+
         indices = []
         for col in columns:
             idx = self.schema.get_field_index(col)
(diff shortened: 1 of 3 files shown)


danepitkin:
This is my first Arrow PR. I am very open to any feedback or suggestions you might have!

rok:
> This is my first Arrow PR. I am very open to any feedback or suggestions you might have!

Welcome @danepitkin! Thanks for opening the PR!

> I followed the instructions as stated in the original GH issue, but there is now some redundancy between `Table.drop(<str or list(str)>)`, `Table.drop_column(<str>)`, and `Table.remove_column(<index>)`. In my opinion, it would have been nicer to consolidate these from the start. However, since backwards compatibility is important, would you say this PR is the right approach?

I would slightly prefer `Table.drop_column(<str or list(str)>)` over `Table.drop_column(<str>)` and `Table.drop(<list(str)>)`. I don't know if `.drop` -> `.drop_column` makes sense at this point. [Pandas uses `.drop` for everything](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.drop.html) but it does index dropping as well. Perhaps you could just alias the two and introduce a deprecation warning? (e.g. [here](https://github.com/apache/arrow/blob/a16c54567ada6729311fd26bdbee4b5e61901410/python/pyarrow/plasma.py#L118)).

danepitkin:
Thanks, Rok!

> I would slightly prefer `Table.drop_column(<str or list(str)>)` over `Table.drop_column(<str>)` and `Table.drop(<list(str)>)`. I don't know if `.drop` -> `.drop_column` makes sense at this point. [Pandas uses `.drop` for everything](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.drop.html) but it does index dropping as well. Perhaps you could just alias the two and introduce a deprecation warning? (e.g. [here](https://github.com/apache/arrow/blob/a16c54567ada6729311fd26bdbee4b5e61901410/python/pyarrow/plasma.py#L118)).

I like this plan. Also, I think this is what the original issue was requesting and I just misread it! I'll update the ticket with `Table.drop_column(<str or list(str)>)` and mark `Table.drop(<list(str)>)` with a deprecation warning. While Arrow does append the word `_column` to other `Table` modification APIs (e.g. `add_column` in arrow, `add` in pandas), I'd be hesitant to remove `Table.drop()` anytime soon, if ever. Breaking backwards compatibility for minor API UX improvements goes against providing a good UX in general. Just wanted to document my thinking here. Overall, I am aligned with you!

rok:
> I'd be hesitant to remove `Table.drop()` anytime soon, if ever. Breaking backwards compatibility for minor API UX improvements goes against providing a good UX in general.

Indeed. I think it also depends a bit on the capability evolution - if we ever introduce `.drop_rows` it would make much more sense to deprecate `.drop` than it does now.

danepitkin:
I decided to name it `Table.drop_columns` instead of `Table.drop_column` to better communicate the possible actions.

Also, instead of a wrapper with a warning, what if we alias it directly with `drop = drop_columns`? I'm not sure if we can still add the warning anymore, but maybe we don't need to officially deprecate the API if there is no timeline for removal.

jorisvandenbossche:
> Also, instead of a wrapper with a warning, what if we alias it directly with `drop = drop_columns`? I'm not sure if we can still add the warning anymore, but maybe we don't need to officially deprecate the API if there is no timeline for removal.

I don't have a strong opinion, but if we don't actually plan to remove it in a few releases, we can indeed also just alias it as you said without warning.

jorisvandenbossche (commented):
Looks good!

python/pyarrow/tests/test_table.py
```diff
@@ -1050,21 +1050,36 @@ def test_table_set_column():
...
+        table.drop_columns(['d'])
+
+
+def test_table_drop():
+    """ test the alias of drop_columns"""
+    a = pa.array(range(5))
+    b = pa.array([-10, -5, 0, 5, 10])
+
+    table = pa.Table.from_arrays([a, b], names=('a', 'b'))
+    t2 = table.drop('b')
```
jorisvandenbossche, line 1079: If we keep the warning, we can assert that this raises a warning (`pytest.raises`), which at the same time also suppresses the warning (we don't want that running our test suite prints this expected warning)

python/pyarrow/table.pxi
```diff
@@ -4708,6 +4711,61 @@ cdef class Table(_PandasConvertible):
 
         return table
 
+    def drop(self, columns):
+        """
+        Drop one or more columns and return a new table.
+
+        Alias of Table.drop_columns for backwards compatibility.
+
+        Parameters
```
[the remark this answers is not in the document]
danepitkin: Thanks! I tried this at first, but thought I was getting a document build error. I probably was just building it wrong 😬 haha. I'll do this because I very much dislike duplication.

python/pyarrow/tests/test_table.py
danepitkin: Thank you! I'm going to remove the warning since there is no plan for deprecation. Good to know for the future though!

python/pyarrow/table.pxi
```diff
@@ -4708,6 +4711,11 @@ cdef class Table(_PandasConvertible):
 
         return table
 
+    drop = drop_columns
+    """
+    Alias of Table.drop_columns, kept for backwards compatibility.
+    """
```
jorisvandenbossche, line 4717: Is this docstring for sphinx? (interactively checking the docstring in a terminal/notebook, that still gives the `drop_columns` one)

danepitkin: Great catch!! Will fix this.

danepitkin: I'm going to remove the alias and use a wrapper. I'm not sure if it's possible to update the docstring of an alias without affecting the original method's docstring.

I tried this:
```
    drop = drop_columns
    drop.__doc__ = """Alias of Table.drop_columns, kept for backwards compatibility."""
```
..but got an error:
```
  ...
  File "pyarrow/table.pxi", line 4729, in init pyarrow.lib
AttributeError: attribute '__doc__' of 'method_descriptor' objects is not writable
```

jorisvandenbossche: I suppose that's because it's implemented in cython, and then that attribute is not writeable as it would be in python (just a guess, didn't actually verify)

But anyway, the wrapper looks good!

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

rok:
Congrats @danepitkin !

PR #33812, opened 2023-01-21
GH-33796: [C++] Fix wrong arrow-testing.pc config with system GoogleTest

### LINKED ISSUE #33796: [C++] arrow-testing.pc.in's Cflags are incorrectly set if gtest isn't built as part of the arrow build
If we don't build gtest as part of the arrow build process (we could get it pre-built somewhere else), then the following variable is unset: 
https://github.com/apache/arrow/blob/bf8780d0ff794c50312d799a9e877430e99dcf8b/cpp/src/arrow/arrow-testing.pc.in#L22

Which is currently only set in the `macro(build_gtest)` cmake function found here: https://github.com/apache/arrow/blob/359f28ba9d62a5e8456d92dfbe5b16b790019edd/cpp/cmake_modules/ThirdpartyToolchain.cmake#L2003

As a result the Cflags generated in: 
https://github.com/apache/arrow/blob/bf8780d0ff794c50312d799a9e877430e99dcf8b/cpp/src/arrow/arrow-testing.pc.in#L29
End up being just `-I`, which then causes an `-I` to appear in the compile command for users building against the arrow project, which in our case (and I assume all cases?) is invalid. Specifically, the behavior we saw was nothing inside of the `std::` namespace could be referenced anymore
As an example, taking out a sub portion of our compile command which was generated with this issue:
`-pthread -I -std=gnu++17`

A solution here would be to not generate any Cflags in the case that `GTEST_INCLUDE_DIR` isn't set.  The `-I` in `Cflags: -I${gtest_includedir}` needs to be created conditionally.  I'll try to add a PR in the next few days to address the issue.

### Component(s)

C++

kou:
### Rationale for this change

Empty `-I` in `Cflags:` generates an invalid build command line.

Add `Requires: gtest` if `gtest.pc` exists.

Add `Cflags: -I...` and `Libs: /.../libgtest.a` if `gtest.pc` doesn't exist.

Bundled GoogleTest isn't supported yet.

### Are these changes tested?

Yes.

Yes.
* Closes: #33796


kou:
+1

PR #13806, opened 2022-08-05
ARROW-17276: [Go][Integration] Implement IPC handling for union type

zeroshade:
With this, the Go implementation finally fully supports IPC handling for All the Arrow DataTypes!

PR #33822, opened 2023-01-21
GH-32613: [C++] Simplify IPC writer for dense unions

### LINKED ISSUE #32613: [C++] Simplify IPC writer for dense unions
In ARROW-10580 we fixed the Arrow C++ implementation so that dense union offsets are always (non-strictly) monotonic for a given child, as mandated by the spec.

The IPC writer implementation, however, still assumes that dense union offsets may be in any order:
https://github.com/apache/arrow/blob/5719576c611929dd790f7f8a1ae3169a8f96f7f1/cpp/src/arrow/ipc/writer.cc#L476-L485

This can probably be simplified, making it slightly less costly to emit a sliced union array.

**Reporter**: [Antoine Pitrou](https://issues.apache.org/jira/browse/ARROW-17339) / @pitrou
#### Related issues:
- [[C++] When Validating, ensure DenseUnionArray offsets are increasing](https://github.com/apache/arrow/issues/26543) (relates to)

<sub>**Note**: *This issue was originally created as [ARROW-17339](https://issues.apache.org/jira/browse/ARROW-17339). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

rtadepalli:
JIRA: https://issues.apache.org/jira/browse/ARROW-17339 
Closes: https://github.com/apache/arrow/issues/32613

### Rationale for this change
Dense union offsets are always non-strictly monotonic for any given child as mandated by the spec, The C++ implementation still assumes that the offsets may be in any order. This can be improved.

Just a change to eliminate looping over the size of a `DenseUnionArray` twice.

### Are these changes tested?

I am not functionally changing anything. All changes respect the spec, and behavior should be 1:1 with the existing implementation. I believe existing tests should suffice.

There are no user facing changes for this.

* Closes: #32613


rtadepalli:
Hello @kou, I have fixed the broken clang-format test. Thanks.

ramasai1:
Hello, can `TestS3FS.OpenOutputStreamBackgroundWrites` be restarted? I am not sure if it is related to my changes. Thanks.

kou:

kou:
Do existing tests cover this case?

rtadepalli:
I think they should. I was looking at the implementation in Go as a reference, and it doesn't seem like the PR that introduced the change (https://github.com/apache/arrow/pull/13806) added any additional tests for finding the offset. I am of the idea right now that existing tests should break if this change is wrong. Please let me know if this is not true, and a unit test is warranted.

kou:
If you add a bug in the target logic and there are tests for this case, these existing tests will be failed.
Could you try it? Or do you want me to do it?

rtadepalli:
But won't those tests run as part of the CI pipeline?

I can definitely try to run them, but would need some help 😅. Would running all tests in the ipc/ package be sufficient?

kou:
You can just push a commit that includes a bug to this branch.
Then CI runs all existing tests with the buggy code. If there are tests that cover this case, they will be failed.

rtadepalli:
Hello @kou , I have introduced a bug in the code. I removed the +1 while calculating the child offset.

rtadepalli:
Looks like `arrow-ipc-read-write-test` is failing with my bug 👍 

I will revert the bug if the test results seem satisfactory to you.

kou:
Thanks.
It seems that we have a test for this case.

Could you revert the bug?

rtadepalli:
I have revert the three commits in which I pushed bugs.

kou (approved):
+1

PR #33831, opened 2023-01-22
GH-33830: Clarify handling of Null values in REE encoding

### LINKED ISSUE #33830: Clarify handling of Null values in REE encoding
In the Columnar format, the handling of null values in the REE type is not described properly.

- The null count of the parent array should be 0, since the parent has no validity bitmap (and there is a general assumption in arrow code that a non-zero null count means the standard validity bitmap is present)
- The document only indirectly describes how Null values are encoded

Thanks @tustvold for catching this!

[1] [https://github.com/apache/arrow/pull/13333#discussion_r1083470362](https://github.com/apache/arrow/pull/13333#discussion_r1083470362)

### Component(s)

Documentation, Format

zagto:
* Closes: #33830

### DIFF
--- a/docs/source/format/Columnar.rst
+++ b/docs/source/format/Columnar.rst
@@ -824,7 +824,7 @@ In Run-end-encoded form, this could appear as:
 
 ::
 
-    * Length: 7, Null count: 2
+    * Length: 7, Null count: 0
     * Child Arrays:
 
       * run_ends (Int32):
--- a/docs/source/format/Columnar.rst
+++ b/docs/source/format/Columnar.rst
@@ -816,7 +816,7 @@ run ends array all are positive and in strictly ascending order. A run end canno
 null.
 
 The REE parent has no validity bitmap, and it's null count field should always be 0.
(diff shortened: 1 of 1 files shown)


docs/source/format/Columnar.rst
```diff
@@ -815,6 +815,9 @@ A run must have have a length of at least 1. This means the values in the
 run ends array all are positive and in strictly ascending order. A run end cannot be
 null.
 
+The REE parent has no validity bitmap, and it's null count field should always be 0.
+validity bitmap. Null values are encoded as runs with the value null.
```
westonpace, line 819:
```suggestion
Null values are encoded as runs with the value null.
```

zagto: Thanks! fixed

tustvold (approved):

westonpace (approved):
This matches my understanding (parent array has no buffers and thus, no null count)

zeroshade (approved):

alamb (commented):
Thank you @zagto

PR #33837, opened 2023-01-23
MINOR: [C++][clang-tidy] Disable modernize-use-nodiscard check 

felipecrv:
Not all functions that return a value should have a `[[nodiscard]]`.

`[[nodiscard]]` adds noise and is better used for functions that:
 1. Return a `Status` or some kind of error code
 2. Have a confusing name like the STL container's `empty()`. The STL uses  `[[nodiscard]]` for `empty()` so people are less likely to call it believing it has the same effect of `clear()`.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -22,6 +22,7 @@ Checks: |
   google-*,
   modernize-*,
   -modernize-use-trailing-return-type,
+  -modernize-use-nodiscard,
 # produce HeaderFilterRegex from cpp/build-support/lint_exclusions.txt with:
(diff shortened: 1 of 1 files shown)


felipecrv:
@pitrou

pitrou (approved):
Good idea. I know @lidavidm was bitten by this at some point :-)

felipecrv:
> Good idea. I know @lidavidm was bitten by this at some point :-)

I've seen that code review. :)

lidavidm:
Thank you!

And yes, the clangd LSP server is very aggressive about suggesting this, which will affect anyone using Emacs, VSCode, etc.

PR #33838, opened 2023-01-23
MINOR: [C++][R] Declare FieldRef(std::vector<FieldRef>) constructor explicit

felipecrv:
This implicit constructor can make passing `std::vector<FieldRef>`s around very confusing. You might think you are passing a vector, but it's implicitly being converted into a single `FieldRef`.

### DIFF
--- a/cpp/src/arrow/type.h
+++ b/cpp/src/arrow/type.h
@@ -1692,9 +1692,7 @@ class ARROW_EXPORT FieldRef : public util::EqualityComparable<FieldRef> {
   FieldRef(int index) : impl_(FieldPath({index})) {}  // NOLINT runtime/explicit
 
   /// Construct a nested FieldRef.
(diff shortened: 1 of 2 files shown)


felipecrv:
@pitrou @rok

rok:
Looks good to me, but I'd ask @paleolimbot for his opinion.

rok:
Looks good to me, but I'd ask @paleolimbot for his opinion.

westonpace (approved):
Having run into this mistake at least once before when working on scanner stuff I can agree it was a bit unexpected.  I suspect it was not intentional (the other constructors here are implicit so it is an easy pattern to follow) but CC @AlenkaF who introduced this method to verify.

felipecrv:
@westonpace I also think it was not intentional given the many other implicit ctors in the file.

pitrou (approved):
Good idea @felipecrv, thanks for contributing this.

PR #33841, opened 2023-01-23
GH-33840: [Go] Improve SQLite Flight SQL Example and provide mainprog

### LINKED ISSUE #33840: [Go] Improve SQLite Flight SQL example and provide simple mainprog

To assist with testing and facilitate easier ADBC testing, the Go SQLite Flight SQL example can be expanded to provide a simple mainprog to launch it (which can then be installed with `go install` or `go get`). This requires a couple improvements to the example to better support features of Flight SQL and not be as limited by the types it accepts (such as unions).

### Component(s)

Developer Tools, Go

zeroshade:
### Rationale for this change

Better facilitate Flight SQL testing by providing a simple mainprog which can be easily installed via `go get` or `go install` for testing and for running CI against.

Improvements to the FlightSQL SQLite Go example and the creation of a mainprog at github.com/apache/arrow/go/v11/arrow/flight/flightsql/cmd/sqlite_flightsql_server

### Are these changes tested?

Covered by existing tests in sqlite_server_test.go

* Closes: #33840

### DIFF
--- a/go/arrow/flight/record_batch_reader.go
+++ b/go/arrow/flight/record_batch_reader.go
@@ -234,3 +234,30 @@ func StreamChunksFromReader(rdr array.RecordReader, ch chan<- StreamChunk) {
 		}
 	}
 }
+
+func ConcatenateReaders(rdrs []array.RecordReader, ch chan<- StreamChunk) {
+	defer close(ch)
+	defer func() {
+		for _, r := range rdrs {
+			r.Release()
+		}
+
+		if err := recover(); err != nil {
+			ch <- StreamChunk{Err: fmt.Errorf("panic while reading: %s", err)}
+		}
+	}()
+
+	for _, r := range rdrs {
+		for r.Next() {
+			rec := r.Record()
+			rec.Retain()
+			ch <- StreamChunk{Data: rec}
+		}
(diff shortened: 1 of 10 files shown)


go/arrow/flight/record_batch_reader.go
```diff
@@ -234,3 +234,30 @@ func StreamChunksFromReader(rdr array.RecordReader, ch chan<- StreamChunk) {
...
+
+func ConcatenateReaders(rdrs []array.RecordReader, ch chan<- StreamChunk) {
+	defer close(ch)
+	defer func() {
+		for _, r := range rdrs {
+			r.Release()
+		}
+
+		if err := recover(); err != nil {
+			ch <- StreamChunk{Err: fmt.Errorf("panic while reading: %s", err)}
```
lidavidm, line 246: yikes, is a panic typical here?

lidavidm (approved):

go/arrow/flight/record_batch_reader.go
zeroshade: it shouldn't be, but it is possible so i'm protecting against other server writers throwing panics in their implementation by catching it and returning an error instead of bringing down the server.

PR #14505, opened 2022-10-25
ARROW-16707 [C++] Implement Rank kernel on chunked arrays

WillAyd:
Initial implementation - could use some guidance on a few points

PR #33846, opened 2023-01-24
GH-32050: [C++] Implement Rank kernel on chunked arrays

### LINKED ISSUE #32050: [C++] Implement Rank kernel on chunked arrays
In ARROW-16234 the rank kernel was implemented for plain arrays, it should also be implemented for chunked arrays.

**Reporter**: [Antoine Pitrou](https://issues.apache.org/jira/browse/ARROW-16707) / @pitrou
**Assignee**: [Will Ayd](https://issues.apache.org/jira/browse/ARROW-16707) / @WillAyd
#### PRs and other links:
- [GitHub Pull Request #14505](https://github.com/apache/arrow/pull/14505)

<sub>**Note**: *This issue was originally created as [ARROW-16707](https://issues.apache.org/jira/browse/ARROW-16707). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

benibus:
This is a duplicate/continuation of https://github.com/apache/arrow/pull/14505.

William Ayd did the bulk of the work here. There was a somewhat nontrivial rebase that needed to occur (due to one of my recent PRs) so Will's commits have been squashed with the credits intact.

Besides that, I did some refactoring to avoid duplicating `ChunkedArraySorter`'s logic in `ChunkedArrayRanker`. 

* Closes: #32050

### DIFF
--- a/cpp/src/arrow/compute/kernels/vector_rank.cc
+++ b/cpp/src/arrow/compute/kernels/vector_rank.cc
@@ -246,111 +246,24 @@ class ChunkedArrayRanker : public TypeVisitor {
     using T = typename GetViewType<InType>::T;
     using ArrayType = typename TypeTraits<InType>::ArrayType;
 
-    const auto num_chunks = chunked_array_.num_chunks();
-    if (num_chunks == 0) {
+    if (chunked_array_.num_chunks() == 0) {
       return Status::OK();
     }
+
+    ARROW_ASSIGN_OR_RAISE(
+        NullPartitionResult sorted,
+        SortChunkedArray(ctx_, indices_begin_, indices_end_, physical_type_,
+                         physical_chunks_, order_, null_placement_));
+
     const auto arrays = GetArrayPointers(physical_chunks_);
-
-    ArraySortOptions array_options(order_, null_placement_);
-
-    ARROW_ASSIGN_OR_RAISE(auto array_sorter, GetArraySorter(*physical_type_));
-
-    // See related ChunkedArraySort method for comments
-    std::vector<NullPartitionResult> sorted(num_chunks);
-    int64_t begin_offset = 0;
-    int64_t end_offset = 0;
-    int64_t null_count = 0;
-    for (int i = 0; i < num_chunks; ++i) {
-      const auto array = checked_cast<const ArrayType*>(arrays[i]);
-      end_offset += array->length();
-      null_count += array->null_count();
-      sorted[i] = array_sorter(indices_begin_ + begin_offset, indices_begin_ + end_offset,
-                               *array, begin_offset, array_options);
-      begin_offset = end_offset;
-    }
-    DCHECK_EQ(end_offset, indices_end_ - indices_begin_);
-
-    auto resolver = ChunkedArrayResolver(arrays);
-    if (sorted.size() > 1) {
-      auto merge_nulls = [&](uint64_t* nulls_begin, uint64_t* nulls_middle,
-                             uint64_t* nulls_end, uint64_t* temp_indices,
-                             int64_t null_count) {
-        if (has_null_like_values<typename ArrayType::TypeClass>::value) {
-          PartitionNullsOnly<StablePartitioner>(nulls_begin, nulls_end, resolver,
-                                                null_count, null_placement_);
-        }
-      };
-      auto merge_non_nulls = [&](uint64_t* range_begin, uint64_t* range_middle,
-                                 uint64_t* range_end, uint64_t* temp_indices) {
-        MergeNonNulls<ArrayType>(range_begin, range_middle, range_end, arrays,
-                                 temp_indices);
-      };
-
-      MergeImpl merge_impl{null_placement_, std::move(merge_nulls),
-                           std::move(merge_non_nulls)};
-      // std::merge is only called on non-null values, so size temp indices accordingly
-      RETURN_NOT_OK(merge_impl.Init(ctx_, indices_end_ - indices_begin_ - null_count));
-
-      while (sorted.size() > 1) {
-        auto out_it = sorted.begin();
-        auto it = sorted.begin();
-        while (it < sorted.end() - 1) {
-          const auto& left = *it++;
-          const auto& right = *it++;
-          DCHECK_EQ(left.overall_end(), right.overall_begin());
-          const auto merged = merge_impl.Merge(left, right, null_count);
-          *out_it++ = merged;
-        }
-        if (it < sorted.end()) {
-          *out_it++ = *it++;
-        }
-        sorted.erase(out_it, sorted.end());
-      }
-    }
-
-    DCHECK_EQ(sorted.size(), 1);
-    DCHECK_EQ(sorted[0].overall_begin(), indices_begin_);
-    DCHECK_EQ(sorted[0].overall_end(), indices_end_);
-    // Note that "nulls" can also include NaNs, hence the >= check
-    DCHECK_GE(sorted[0].null_count(), null_count);
-
-    auto value_selector = [resolver](int64_t index) {
+    auto value_selector = [resolver = ChunkedArrayResolver(arrays)](int64_t index) {
       return resolver.Resolve<ArrayType>(index).Value();
     };
     auto rankings =
-        CreateRankings<T>(ctx_, sorted[0], null_placement_, tiebreaker_, value_selector);
+        CreateRankings<T>(ctx_, sorted, null_placement_, tiebreaker_, value_selector);
     ARROW_ASSIGN_OR_RAISE(auto output, rankings);
     *output_ = output;
     return Status::OK();
-  }
-
-  template <typename ArrayType>
-  void MergeNonNulls(uint64_t* range_begin, uint64_t* range_middle, uint64_t* range_end,
-                     const std::vector<const Array*>& arrays, uint64_t* temp_indices) {
-    const ChunkedArrayResolver left_resolver(arrays);
-    const ChunkedArrayResolver right_resolver(arrays);
-
-    if (order_ == SortOrder::Ascending) {
-      std::merge(range_begin, range_middle, range_middle, range_end, temp_indices,
-                 [&](uint64_t left, uint64_t right) {
-                   const auto chunk_left = left_resolver.Resolve<ArrayType>(left);
-                   const auto chunk_right = right_resolver.Resolve<ArrayType>(right);
-                   return chunk_left.Value() < chunk_right.Value();
-                 });
-    } else {
-      std::merge(range_begin, range_middle, range_middle, range_end, temp_indices,
-                 [&](uint64_t left, uint64_t right) {
-                   const auto chunk_left = left_resolver.Resolve<ArrayType>(left);
-                   const auto chunk_right = right_resolver.Resolve<ArrayType>(right);
-                   // We don't use 'left > right' here to reduce required
-                   // operator. If we use 'right < left' here, '<' is only
-                   // required.
-                   return chunk_right.Value() < chunk_left.Value();
-                 });
-    }
-    // Copy back temp area into main buffer
-    std::copy(temp_indices, temp_indices + (range_end - range_begin), range_begin);
   }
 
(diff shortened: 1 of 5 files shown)


benibus:
cc @pitrou

WillAyd:
Looks nice - thanks for the help on this

westonpace:
Why are we interested in compute on chunked arrays in the first place?  For example, for rank, couldn't we use an exec plan with an aggregate node?

My understanding is that we have:

Single batch -> function registry
Stream of batches -> Execution plans

I worry that maintaining a chunked array path for all the aggregate kernels may be tedious without much value.

pitrou:
I assume there are many cases where using the compute facilities is much easier than creating an exec plan (which also doesn't really have a good solution in Python).

Ideally the vector kernel API for chunked arrays would be based on iterative chunk processing. With some manpower this would be a good move IMHO, and exec plans could easily build on top of that.

westonpace:
> I assume there are many cases where using the compute facilities is much easier than creating an exec plan

The exec plan does not support window functions today which is what we would need to easily do this rank function.  That being said, if it did support window functions, I would argue that it is easier than using the compute facilities directly.  For example, here is `Sum` support for chunked arrays:

```
std::shared_ptr<Table> ChunkedArrayToTable(std::shared_ptr<ChunkedArray> chunked_array) {
  auto table_schema = schema({field("values", chunked_array->type())});
  int64_t length = chunked_array->length();
  return Table::Make(std::move(table_schema), {std::move(chunked_array)}, length);
}

Result<std::shared_ptr<ChunkedArray>> Sum(std::shared_ptr<ChunkedArray> chunked_array,
                                          const ScalarAggregateOptions& options) {
  auto table = ChunkedArrayToTable(std::move(chunked_array));
  Declaration plan = Declaration::Sequence(
      {{"table_source", TableSourceNodeOptions(std::move(table))},
       {"aggregate",
        AggregateNodeOptions({{"sum",
                               std::make_shared<ScalarAggregateOptions>(options),
                               {0},
                               "values_sum"}},
                             {})}});
  ARROW_ASSIGN_OR_RAISE(auto sum_table, DeclarationToTable(std::move(plan)));
  return sum_table->column(0);
}

TEST(RankTest, Test) {
  auto test_array =
      ChunkedArrayFromJSON(int32(), {R"([4, 5, null])", R"([9])", R"([7, 6, 10, 8])",
                                     R"([11, 11])", R"([14])", R"([14])", R"([16])"});
  ScalarAggregateOptions options;
  ASSERT_OK_AND_ASSIGN(auto result, Sum(std::move(test_array), options));
  auto expected = ChunkedArrayFromJSON(int64(), {R"([115])"});
  AssertChunkedEqual(*expected, *result);
  std::cout << result->ToString() << std::endl;
}
```

> Ideally the vector kernel API for chunked arrays would be based on iterative chunk processing.

This is how the aggregate and project nodes work today so I think we'd be reinventing that for non-window function cases.

> With some manpower this would be a good move IMHO, and exec plans could easily build on top of that.

I don't think we could build exec plans on top of this because exec plans assume the input doesn't fit in memory which is not possible with chunked arrays.

Window functions (e.g. rank) do generally require sorted input, and so "doesn't fit in memory" is a problem case.  However, we are getting pretty close at adding spilling support for hash-join and it would be straightforward to extend that to order-by.  At that point a larger-than-memory rank should be straightforward.

However, I don't have the manpower to dedicate to this, so as I said before, I won't stand in the way of a more incremental approach with chunked arrays.  It just seems unfortunate.

pitrou:
What do you call "window function" exactly? The rank function doesn't fit in my understanding of a window function.

westonpace:
A scalar function produces 1 output row per input row and the and is stateless (the answer is completely determined by the 1 input row)

An aggregate function produces only 1 output row and is stateful (the answer may depend on more than just a single input row)

A window function produces 1 output row per input (similar to a scalar function) and is stateful (similar to an aggregate function)

The "window" you use to determine the output can be based on a number of things but a very common window is "all rows up to this point ordered by X"

E.g. "rank", in postgres, is expressed as:

[`SELECT RANK () OVER ( ORDER BY c ) FROM my_table;`](http://sqlfiddle.com/#!17/261c7/1)

You can also compute the rank within the context of another column (e.g. month):

`SELECT val, RANK () OVER ( PARTITION BY month ORDER BY val ) FROM my_table;`

Also note that all aggregate functions can also be used as window functions.  In that case you simply broadcast the result of the aggregate to all rows in window.

pitrou:
> I don't think we could build exec plans on top of this because exec plans assume the input doesn't fit in memory which is not possible with chunked arrays.

I mean the VectorKernel interface for chunked arrays could be a streaming interface such as:
```c++
  virtual Result<std::shared_ptr<ArrayVector>> ConsumeChunks(const ArrayVector&) = 0;
  virtual Result<std::shared_ptr<ArrayVector>> Finish() = 0;
```

cpp/src/arrow/compute/kernels/vector_sort_test.cc
```diff
@@ -2182,5 +2185,146 @@ TEST(TestRankForFixedSizeBinary, RankFixedSizeBinary) {
       ArrayFromJSON(binary_type, R"(["aaa", "   ", "eee", null, "eee", null, "   "])"));
 }
 
+TEST(TestRankForReal, RankRealChunked) {
```
pitrou, line 2188: Instead of essentially duplicating existing tests, perhaps we can generate `ChunkedArray` checks from the `Array` tests, or vice-versa?

cpp/src/arrow/compute/kernels/vector_sort_test.cc
```diff
@@ -2182,5 +2185,146 @@ TEST(TestRankForFixedSizeBinary, RankFixedSizeBinary) {
...
+  for (auto real_type : ::arrow::FloatingPointTypes()) {
+    for (auto null_placement : AllNullPlacements()) {
+      for (auto tiebreaker : AllTiebreakers()) {
+        for (auto order : AllOrders()) {
+          AssertRankEmpty(real_type, order, null_placement, tiebreaker);
+        }
+
+        AssertRankSimple(
+            ChunkedArrayFromJSON(real_type, {"[2.1, 3.2]", "[1.0, 0.0, 5.5]"}),
+            null_placement, tiebreaker);
```
pitrou, line 2198: Regardless of what strategy is chosen, it would be nice to exercise empty chunks as well.

cpp/src/arrow/compute/kernels/vector_rank.cc
```diff
@@ -25,6 +25,120 @@ namespace {
 // ----------------------------------------------------------------------
 // Rank implementation
 
+template <typename T>
+Result<Datum> CreateRankings(ExecContext* ctx, const NullPartitionResult& sorted,
+                             const NullPlacement null_placement,
+                             const RankOptions::Tiebreaker tiebreaker,
+                             std::function<T(int64_t)> value_selector) {
```
pitrou, line 32: Ideally, the `std::function` overhead would be removed by the compiler after inlining, but perhaps it's better to avoid it upfront (including in debug builds) by passing this as a template parameter instead?
```c++
template <typename ValueSelector, typename T = std::invoke_result_t<ValueSelector, int64_t>>
Result<Datum> CreateRankings(ExecContext* ctx, const NullPartitionResult& sorted,
                             const NullPlacement null_placement,
                             const RankOptions::Tiebreaker tiebreaker,
                             ValueSelector&& value_selector) {
```

cpp/src/arrow/compute/kernels/vector_rank.cc
```diff
@@ -77,121 +191,90 @@ class ArrayRanker : public TypeVisitor {
...
 
-      case RankOptions::Min: {
-        T curr_value, prev_value{};
-        rank = 0;
+    auto value_selector = [&arr](int64_t index) {
+      return GetView::LogicalValue(arr.GetView(index));
+    };
+    auto rankings =
+        CreateRankings<T>(ctx_, sorted, null_placement_, tiebreaker_, value_selector);
+    ARROW_ASSIGN_OR_RAISE(auto output, rankings);
```
pitrou, line 200: I'm curious, why didn't you collapse these two lines in one? The `rankings` temporary doesn't seem useful...

pitrou: (you might need to parenthesize the `CreateRankings` call because of syntax limitations, though)

pitrou: Also, this can perhaps even be `ARROW_ASSIGN_OR_RAISE(*output_, ....`

cpp/src/arrow/compute/kernels/vector_rank.cc
```diff
@@ -77,121 +191,90 @@ class ArrayRanker : public TypeVisitor {
...
-            rank = (it - sorted.overall_begin()) + 1;
-          }
-          out_begin[*it] = rank;
-          prev_value = curr_value;
-        }
+class ChunkedArrayRanker : public TypeVisitor {
+ public:
+  // TODO: here we accept order / null_placement / tiebreaker as separate arguments
+  // whereas the ArrayRanker accepts them as the RankOptions struct; this is consistent
+  // with ArraySorter / ChunkedArraySorter, so likely should refactor ArrayRanker
```
pitrou, line 218: If the refactor is small enough, why not do it in this PR?

cpp/src/arrow/compute/kernels/vector_rank.cc
```diff
@@ -237,6 +323,32 @@ class RankMetaFunction : public MetaFunction {
...
+    auto length = chunked_array.length();
+    auto buffer_size = bit_util::BytesForBits(
+        length * std::static_pointer_cast<UInt64Type>(sort_type)->bit_width());
+    std::vector<std::shared_ptr<Buffer>> buffers(2);
+    ARROW_ASSIGN_OR_RAISE(buffers[1],
+                          AllocateResizableBuffer(buffer_size, ctx->memory_pool()));
+    auto out = std::make_shared<ArrayData>(sort_type, length, buffers, 0);
+    auto sort_begin = out->GetMutableValues<uint64_t>(1);
+    auto sort_end = sort_begin + length;
+    std::iota(sort_begin, sort_end, 0);
```
pitrou, line 344: Most of this should probably be shared with the `Array` code path?

pitrou (commented):
Thanks a lot @benibus ! This looks good on the principle, just some minor questions and comments.

pitrou:
@benibus Feel free to ping when this PR is ready again.

cpp/src/arrow/compute/kernels/vector_sort_test.cc
```diff
@@ -1962,15 +1962,50 @@ INSTANTIATE_TEST_SUITE_P(AllNull, TestTableSortIndicesRandom,
 // ----------------------------------------------------------------------
 // Tests for Rank
 
-void AssertRank(const std::shared_ptr<Array>& input, SortOrder order,
-                NullPlacement null_placement, RankOptions::Tiebreaker tiebreaker,
+struct RankTestCase {
```
pitrou, line 1965: Hmm... sorry to bother you with this, but it would be more idiomatic to use a GTest fixture (i.e. a class derived from `::testing::Test`).

You could have something like:
```c++
class TestRank : public ::testing::Test {
 public:
  void SetInput(const std::shared_ptr<Array>& array) {
    ...
```

benibus: Yeah I agree. That would unify the base test names too (as of right now, it's kind of a pain to filter out the rank tests as a set).

pitrou (approved):
Thanks for this! I just pushed a trivial cleanup and will merge.

benibus:
Thanks! ...I somehow missed that those were all the exact same line.

WillAyd (approved):
Looks great - thanks again for taking this over. Learned a few things looking at your implementation 👍

benibus:
@WillAyd No problem! Thanks for the work done on your end. I learned a few things as well.

PR #33847, opened 2023-01-24
GH-33701: [C++] Add support for LTO (link time optimization) build

### LINKED ISSUE #33701: [R] Link-time optimization reports violations of one-definition rule in the R package

The warning:

```
* checking whether package ‘arrow’ can be installed ... [18m/17m] WARNING
Found the following significant warnings:
  /arrow/cpp/src/arrow/type.h:323: warning: virtual table of type ‘struct NestedType’ violates one definition rule [-Wodr]
  /arrow/cpp/src/arrow/type.h:288: warning: virtual table of type ‘struct FixedWidthType’ violates one definition rule [-Wodr]
See ‘/arrow/r/check/arrow.Rcheck/00install.out’ for details.
```

‘/arrow/r/check/arrow.Rcheck/00install.out’ (for details): https://dev.azure.com/ursacomputing/crossbow/_build/results?buildId=42712&view=logs&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=6c939d89-0d1a-51f2-8b30-091a7a82e98c&l=1591

```
/arrow/cpp/src/arrow/type.h:323: warning: virtual table of type ‘struct NestedType’ violates one definition rule [-Wodr]
  323 | class ARROW_EXPORT NestedType : public DataType, public ParametricType {
      | 
/arrow/r/check/arrow.Rcheck/00_pkg_src/arrow/libarrow/arrow-10.0.1.9000/include/arrow/type.h:323:20: note: the conflicting type defined in another translation unit
  323 | class ARROW_EXPORT NestedType : public DataType, public ParametricType {
      |                    ^
/arrow/r/check/arrow.Rcheck/00_pkg_src/arrow/libarrow/arrow-10.0.1.9000/include/arrow/type.h:177:22: note: virtual method ‘storage_id’
  177 |   virtual Type::type storage_id() const { return id_; }
      |                      ^
<built-in>: note: ought to match virtual method ‘__cxa_pure_virtual’ but does not
/arrow/cpp/src/arrow/type.h:288: warning: virtual table of type ‘struct FixedWidthType’ violates one definition rule [-Wodr]
  288 | class ARROW_EXPORT FixedWidthType : public DataType {
      | 
/arrow/r/check/arrow.Rcheck/00_pkg_src/arrow/libarrow/arrow-10.0.1.9000/include/arrow/type.h:288:20: note: the conflicting type defined in another translation unit
  288 | class ARROW_EXPORT FixedWidthType : public DataType {
      |                    ^
/arrow/r/check/arrow.Rcheck/00_pkg_src/arrow/libarrow/arrow-10.0.1.9000/include/arrow/type.h:177:22: note: virtual method ‘storage_id’
  177 |   virtual Type::type storage_id() const { return id_; }
      |                      ^
<built-in>: note: ought to match virtual method ‘__cxa_pure_virtual’ but does not
```

I think this is an R package problem...somehow we are mixing up the includes or the linking in such a way that we're picking up includes outside of the package directory (which we really don't want in case there's a user install of Arrow that's a different version!).

### Component(s)

R

kou:
### Rationale for this change

Some base type classes don't have hidden method implementations. It may define these classes in each translation unit. It may cause one-definition-rule violation with `-flto`.

Define at least one hidden method to prevent defining these base type classes in each translation unit.

### Are these changes tested?

How to reproduce:

```bash
CFLAGS="-flto" CXXFLAGS="-flto" cmake ...
cmake --build ...
```

This reports link errors without this change.

No.
* Closes: #33701

### DIFF
--- a/cpp/src/arrow/type.h
+++ b/cpp/src/arrow/type.h
@@ -288,24 +288,36 @@ std::shared_ptr<DataType> GetPhysicalType(const std::shared_ptr<DataType>& type)
 class ARROW_EXPORT FixedWidthType : public DataType {
  public:
   using DataType::DataType;
+  // This is only for preventing defining this class in each
+  // translation unit to avoid one-definition-rule violation.
+  ~FixedWidthType() override;
 };
 
 /// \brief Base class for all data types representing primitive values
 class ARROW_EXPORT PrimitiveCType : public FixedWidthType {
  public:
   using FixedWidthType::FixedWidthType;
+  // This is only for preventing defining this class in each
+  // translation unit to avoid one-definition-rule violation.
+  ~PrimitiveCType() override;
 };
 
 /// \brief Base class for all numeric data types
 class ARROW_EXPORT NumberType : public PrimitiveCType {
  public:
   using PrimitiveCType::PrimitiveCType;
+  // This is only for preventing defining this class in each
+  // translation unit to avoid one-definition-rule violation.
+  ~NumberType() override;
 };
 
 /// \brief Base class for all integral data types
 class ARROW_EXPORT IntegerType : public NumberType {
  public:
   using NumberType::NumberType;
+  // This is only for preventing defining this class in each
+  // translation unit to avoid one-definition-rule violation.
+  ~IntegerType() override;
   virtual bool is_signed() const = 0;
 };
 
@@ -313,6 +325,9 @@ class ARROW_EXPORT IntegerType : public NumberType {
 class ARROW_EXPORT FloatingPointType : public NumberType {
  public:
   using NumberType::NumberType;
+  // This is only for preventing defining this class in each
+  // translation unit to avoid one-definition-rule violation.
+  ~FloatingPointType() override;
   enum Precision { HALF, SINGLE, DOUBLE };
   virtual Precision precision() const = 0;
 };
@@ -323,6 +338,9 @@ class ParametricType {};
 class ARROW_EXPORT NestedType : public DataType, public ParametricType {
  public:
   using DataType::DataType;
+  // This is only for preventing defining this class in each
+  // translation unit to avoid one-definition-rule violation.
+  ~NestedType() override;
 };
 
 /// \brief The combination of a field name and data type, with optional metadata
@@ -650,6 +668,9 @@ class ARROW_EXPORT DoubleType
 class ARROW_EXPORT BaseBinaryType : public DataType {
  public:
   using DataType::DataType;
+  // This is only for preventing defining this class in each
+  // translation unit to avoid one-definition-rule violation.
+  ~BaseBinaryType() override;
 };
 
 constexpr int64_t kBinaryMemoryLimit = std::numeric_limits<int32_t>::max() - 1;
@@ -893,6 +914,9 @@ class ARROW_EXPORT Decimal256Type : public DecimalType {
 class ARROW_EXPORT BaseListType : public NestedType {
  public:
   using NestedType::NestedType;
+  // This is only for preventing defining this class in each
+  // translation unit to avoid one-definition-rule violation.
+  ~BaseListType() override;
   const std::shared_ptr<Field>& value_field() const { return children_[0]; }
 
   std::shared_ptr<DataType> value_type() const { return children_[0]->type(); }
@@ -1209,6 +1233,9 @@ class ARROW_EXPORT DenseUnionType : public UnionType {
 class ARROW_EXPORT TemporalType : public FixedWidthType {
  public:
   using FixedWidthType::FixedWidthType;
+  // This is only for preventing defining this class in each
+  // translation unit to avoid one-definition-rule violation.
+  ~TemporalType() override;
 
   DataTypeLayout layout() const override {
     return DataTypeLayout(
--- a/cpp/src/arrow/type.cc
+++ b/cpp/src/arrow/type.cc
@@ -454,6 +454,16 @@ std::string TypeHolder::ToString(const std::vector<TypeHolder>& types) {
 
 // ----------------------------------------------------------------------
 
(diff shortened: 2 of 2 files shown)


kou:
@github-actions crossbow submit test-r-rhub-debian-gcc-devel-lto-latest

kou:
@paleolimbot This is another approach.

cpp/src/arrow/type.h
```diff
@@ -288,31 +288,46 @@ std::shared_ptr<DataType> GetPhysicalType(const std::shared_ptr<DataType>& type)
 class ARROW_EXPORT FixedWidthType : public DataType {
  public:
   using DataType::DataType;
+  // This is only for preventing defining this class in each
+  // translation unit to avoid one-definition-rule violation.
+  ~FixedWidthType() override;
```
cyb70289, line 293: Curious why it causes ODR violation.

kou: To be honest, I don't know much about this but...

I tried to create a small example to reproduce this case but couldn't create it... So I explain what I observed:

If we don't have this, `.o` of a C++ code that has `std::make_shared<arrow::XXXType>` such as `std::make_shared<arrow::ListType>` has the following `nm` result:

```console
$ nm --demangle src/arrow/CMakeFiles/arrow_objlib.dir/compute/kernels/scalar_cast_numeric.cc.o | grep 'vtable for arrow::FloatingPointType'
00000000 W vtable for arrow::FloatingPointType
```

`W` is a global weak symbol.

https://linux.die.net/man/1/nm

>  If lowercase, the symbol is local; if uppercase, the symbol is global (external). 

> "W"
> "w"
>
> The symbol is a weak symbol that has not been specifically tagged as a weak object symbol.

If we have this, the symbol becomes a undefined symbol:

```console
$ nm --demangle src/arrow/CMakeFiles/arrow_objlib.dir/compute/kernels/scalar_cast_numeric.cc.o | grep 'vtable for arrow::FloatingPointType'
         U vtable for arrow::FloatingPointType
```

https://linux.die.net/man/1/nm

> "U"
>
> The symbol is undefined. 

If there are multiple `W vtable for arrow::FloatingPointType` in our `.o`. linking `libarrow.so` with `-flto` is failed:

```text
[650/1111] Linking CXX shared library debug/libarrow.so.1100.0.0
FAILED: debug/libarrow.so.1100.0.0 
: && /bin/c++ -fPIC -Wno-noexcept-type -flto -fdiagnostics-color=always  -Wall -Wno-conversion -Wno-sign-conversion -Wunused-result -fno-semantic-interposition -msse4.2  -g -Werror -O0 -ggdb -shared -Wl,-soname,libarrow.so.1100 -o debug/libarrow.so.1100.0.0 ...(**/*.o)... && :
cpp/src/arrow/type.h:313: error: virtual table of type ‘struct FloatingPointType’ violates one definition rule [-Werror=odr]
  313 | class ARROW_EXPORT FloatingPointType : public NumberType {
      | 
cpp/src/arrow/type.h:313:20: note: the conflicting type defined in another translation unit
  313 | class ARROW_EXPORT FloatingPointType : public NumberType {
      |                    ^
<built-in>: note: virtual method ‘__cxa_pure_virtual’
...
```

It seems that `W` and `-flto` is a bad combination. If we change `W` to `U` by defining a not-inlined deconstructor, we don't get the ODR error.

Do you notice anything from this observation?

cyb70289: I don't know why it happens. Clang doesn't complain ODR issue.
cc @pitrou, @westonpace, @lidavidm for comments.

pitrou, line 293: You probably don't need to define the destructor separately?
```suggestion
  ~FixedWidthType() override = default;
```

westonpace: I understand why adding that destructor changes W to U.  You are defining a non-inline key function.  Therefore the vtable is defined in only one place.

However, I don't understand why it is an ODR error.  For example, in the [clang docs](https://lld.llvm.org/missingkeyfunction.html):

> When a class has no non-pure, non-inline, virtual functions, there is no key function, and the compiler is forced to emit the vtable in every translation unit that references the class. In this case, it is emitted in a COMDAT section, which allows the linker to eliminate all duplicate copies. This is still wasteful in terms of object file size and link time, so it’s always advisable to ensure there is at least one eligible function that can serve as the key function.

kou: I tried `= default`. If we add `= default`, generated symbol is still `W` not `U`. It seems that we need to define the destructor separately for `U`.

kou: It seems that this is only happen with GCC (GNU ld?).
I couldn't reproduce this with `CC=clang CXX=clang++ CFLAGS="-flto" CXXFLAGS="-flto" cmake ...`. (`W` symbol is still generated but there is no link error.)

Is it acceptable that we add this change only for GCC and LTO?

cyb70289: I think this change is okay. And I wonder how you find this solution, it's not obvious from the error log.

cyb70289 (approved):
+1

cpp/src/arrow/type.h
kou: It's just a try and error...

There are some abstract classes for `arrow::DataType` family. Most of them are in error messages but `arrow::DataType` isn't shown. I noticed that `arrow::DataType` is `U` and others are `W` in `nm --demangle` result. I found that `arrow::DataType::~DataType()` is defined in `type.cc` but other abstract classes don't define their destructors. So I tried to define them.

westonpace: Yes, that is what it means.  I think "wasteful" is too strong of a word though.  I think the consequence should generally be minor enough that it isn't really a concern.

PR #33853, opened 2023-01-24
GH-33852: [Go] Return a catalog/schema from Flight SQL example server

### LINKED ISSUE #33852: [Go] Flight SQL example server should return at least one catalog and schema

This is necessary for integration testing with the Flight SQL driver. Although SQLite does not really have catalogs/schemas, we should at least expose some structure there. (And, attached databases do mean that SQLite basically has the notion of catalog.)

### Component(s)

Go

lidavidm:
### Rationale for this change

The example server returns 0 catalogs and 0 schemas because it is based on SQLite. But SQLite does sort of have a concept of catalogs via attached databases. And we should still pretend there is a single (unnamed) schema so that clients get a consistent view of database metadata.

Update how GetTables/GetCatalogs/GetDbSchemas are implemented.

### Are these changes tested?

Tests were updated.

No, the example is for internal development.
* Closes: #33852

### DIFF
--- a/go/arrow/flight/flightsql/example/sqlite_server.go
+++ b/go/arrow/flight/flightsql/example/sqlite_server.go
@@ -70,7 +70,7 @@ func genRandomString() []byte {
 
 func prepareQueryForGetTables(cmd flightsql.GetTables) string {
 	var b strings.Builder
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM

PR #33855, opened 2023-01-24
MINOR: [Go] Don't panic on unknown SQLite type in example server

lidavidm:
### Rationale for this change

GetTables just panics right now because SQLite internal tables don't always have a defined type for all columns. 

Just mimic the C++ implementation and assign NullType for such columns.

### Are these changes tested?

Yes

No, this is an example for internal development.


zeroshade (approved):
LGTM

PR #33858, opened 2023-01-25
GH-15054: [C++] Change s3 finalization to happen after arrow threads finished, add pyarrow exit hook

### LINKED ISSUE #15054: [CI][Python] wheel-manylinux2014-* sometimes crashed on pytest exit

westonpace:
CRITICAL FIX: When statically linking error with AWS it was possible to have a crash on shutdown/exit.  Now that should no longer be possible.

BREAKING CHANGE: S3 can only be initialized and finalized once.

BREAKING CHANGE: S3 (the AWS SDK) will not be finalized until after all CPU & I/O threads are finished.
* Closes: #15054

### DIFF
--- a/cpp/src/arrow/filesystem/s3fs.cc
+++ b/cpp/src/arrow/filesystem/s3fs.cc
@@ -2573,7 +2573,7 @@ namespace {
 
 struct AwsInstance : public ::arrow::internal::Executor::Resource {
   AwsInstance() : is_initialized_(false), is_finalized_(false) {}
-  ~AwsInstance() { Finalize(); }
+  ~AwsInstance() { Finalize(/*from_destructor=*/true); }
 
   // Returns true iff the instance was newly initialized with `options`
   Result<bool> EnsureInitialized(const S3GlobalOptions& options) {
@@ -2590,12 +2590,17 @@ struct AwsInstance : public ::arrow::internal::Executor::Resource {
 
   bool IsInitialized() { return !is_finalized_ && is_initialized_; }
 
-  void Finalize() {
+  void Finalize(bool from_destructor = false) {
     bool expected = true;
     is_finalized_.store(true);
     if (is_initialized_.compare_exchange_strong(expected, false)) {
-      RegionResolver::ResetDefaultInstance();
-      Aws::ShutdownAPI(aws_options_);
+      if (from_destructor) {
+        ARROW_LOG(WARNING)
+            << " arrow::fs::FinalizeS3 was not called even though S3 was initialized.  "
+               "This could lead to a segmentation fault at exit";
+        RegionResolver::ResetDefaultInstance();
+        Aws::ShutdownAPI(aws_options_);
+      }
     }
   }
 
--- a/cpp/src/arrow/filesystem/s3fs.cc
+++ b/cpp/src/arrow/filesystem/s3fs.cc
@@ -2571,11 +2571,38 @@ Result<std::shared_ptr<io::OutputStream>> S3FileSystem::OpenAppendStream(
 
 namespace {
 
-std::mutex aws_init_lock;
 Aws::SDKOptions aws_options;
 
 struct AwsInstance : public ::arrow::internal::Executor::Resource {
-  explicit AwsInstance(const S3GlobalOptions& options) {
+  AwsInstance() : is_initialized_(false), is_finalized_(false) {}
+  ~AwsInstance() { Finalize(); }
+
+  // Returns true iff the instance was newly initialized with `options`
+  Result<bool> EnsureInitialized(const S3GlobalOptions& options) {
+    bool expected = false;
+    if (is_finalized_.load()) {
+      return Status::Invalid("Attempt to initialize S3 after it has been finalized");
+    }
+    if (is_initialized_.compare_exchange_strong(expected, true)) {
+      DoInitialize(options);
+      return true;
+    }
+    return false;
+  }
+
+  bool IsInitialized() { return !is_finalized_ && is_initialized_; }
+
+  void Finalize() {
+    bool expected = true;
+    is_finalized_.store(true);
+    if (is_initialized_.compare_exchange_strong(expected, false)) {
+      RegionResolver::ResetDefaultInstance();
+      Aws::ShutdownAPI(aws_options);
+    }
+  }
+
+ private:
+  void DoInitialize(const S3GlobalOptions& options) {
     Aws::Utils::Logging::LogLevel aws_log_level;
 
 #define LOG_LEVEL_CASE(level_name)                             \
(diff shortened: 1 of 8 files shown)


cpp/src/arrow/filesystem/s3fs.cc
```diff
@@ -2568,84 +2568,92 @@ namespace {
...
+    switch (options.log_level) {
+      LOG_LEVEL_CASE(Fatal)
+      LOG_LEVEL_CASE(Error)
+      LOG_LEVEL_CASE(Warn)
+      LOG_LEVEL_CASE(Info)
+      LOG_LEVEL_CASE(Debug)
+      LOG_LEVEL_CASE(Trace)
+      default:
+        aws_log_level = Aws::Utils::Logging::LogLevel::Off;
+    }
```
vibhatha, line 2591: is the new tab an expected change?

westonpace: Yes.  The initialize and finalize methods were file-scoped methods previously.  Now they have become the constructor/destructor of the `AwsInstance` class.  Since they are in a class they get some indentation.

zeroshade (approved):
I don't know if I'm the best to review here, but at least at the moment I don't see anything glaringly problematic with this and it seems like a good idea in general. 

It feels like it should be better tested, but I'm not particularly sure how would be a good way to do so. But that might just be because it's the end of the day lol.

cpp/src/arrow/filesystem/s3fs.cc
```diff
@@ -2573,98 +2573,105 @@ namespace {
...
-  RegionResolver::ResetDefaultInstance();
-  Aws::ShutdownAPI(aws_options);
-  aws_initialized.store(false);
-  return Status::OK();
+std::shared_ptr<AwsInstance>* GetAwsInstance(const S3GlobalOptions& options) {
+  static auto instance = CreateAwsInstance(options);
+  return &instance;
 }
 
+bool IsAwsInitialized() { return !!GetAwsInstance({}); }
```
felipecrv, line 2647: If this gets called before `InitializeS3`, the `options` passed there will be ignored and an instance based on `{}` options will stick around. I think bringing back the `bool` flag can avoid these risks.

westonpace:
Thanks @felipecrv , that is a good point regarding IsInitialized.  I added the flag back but put it inside the instance object.  I put everything inside the instance object and I think the code is simpler now.  What do you think of this version?

felipecrv (approved):
This LGTM. Thank you!

westonpace:
These R crashes are a little concerning since they seem to be crashes at shutdown.  Maybe it is related.

westonpace:
Yes, the R crashes are relevant.  With this fix we ensure that `AWS::ShutdownAPI` is the last call we make to AWS.  However, AWS has static global objects of its own.  So it is entirely possible that these have been destroyed by the time `AWS::ShutdownAPI` is called.  In fact, by prolonging the call to `ShutdownAPI` here it seems we are making the R situation worse because it is more likely these other static globals have been destroyed.  There is an open AWS issue for this at https://github.com/aws/aws-sdk-cpp/issues/1550  .  However, it does not sound like it will be resolved quickly.  In the meantime an R atexit hook of some kind will be required (technically it has always been required but we've gotten lucky).

r/R/arrow-package.R
```diff
@@ -147,6 +153,12 @@ supported_dplyr_methods <- list(
   # Register extension types that we use internally
   reregister_extension_type(vctrs_extension_type(vctrs::unspecified()))
 
+  # Registers a callback to run at session exit
+  .onLoad <- function(libname, pkgname) {
+    print(parent.env(environment()))
+    reg.finalizer(parent.env(environment()), finalize_s3, onexit=TRUE)
+  }
+
```
paleolimbot, line 161: I know you asked about this earlier and found `.onUnload()`...just putting a note here in case this gets forgotten about.

If `.onUnload()` isn't working you can create your own environment an register a finalizer on that:

```r
s3_finalizer <- new.env(parent = emptyenv())

.onLoad <- function(...) {
  # ...
  reg.finalizer(s3_finalizer, finalize_s3, onexit=TRUE)
}
```

You probably shouldn't register a finalizer on the package namespace because somebody else might have done that (maybe even R).

westonpace:
The broken C++ tests are unrelated (will be fixed by #35145).  It seems the R tests are passing now.  Would appreciate a review from @paleolimbot or @thisisnic on the R changes.

@kou do you know if cglib / ruby is using S3?  If so, is it calling `FinalizeS3` in some kind of shutdown hook?  Or do we need a change there too.

r/R/arrow-package.R
paleolimbot: There is also `.onLoad()` defined twice here as written (you should be able to use the existing `.onLoad()` to register this hook if you can't in fact use `.onUnload()`.

westonpace: Ah, ok.  I will change to this pattern instead of `parent.env(environment())`

westonpace: Also, I just realized I left the print in >_<

paleolimbot: And if `.onUnload()` really isn't working make sure to put a comment in explaining that (so I don't forget about this and try to consolidate our namespace unload incorrectly)

westonpace: Ok, I think I've resolved all your concerns now.  Thanks for looking at this.

westonpace: > And if .onUnload() really isn't working make sure to put a comment in explaining that (so I don't forget about this and try to consolidate our namespace unload incorrectly)

Added a comment

kou:
> do you know if cglib / ruby is using S3? If so, is it calling `FinalizeS3` in some kind of shutdown hook? Or do we need a change there too.

Yes, I know. GLib/Ruby is using S3.
It doesn't call `FinalizeS3()` automatically but it provides APIs that can call `InitializeS3()` and `FinalizeS3()` explicitly. So we don't need changes for GLib/Ruby here for now.
If a report from community is raised, we can consider this.

westonpace:
Thanks, that sounds reasonable.  I'll merge when CI passes then.

assignUser:
I have added the appropriate labels for visibility

PR #33862, opened 2023-01-25
GH-33825: [Python] Expose pyarrow.dataset.get_partition_keys publicly (get key/value from partition expression)

### LINKED ISSUE #33825: [Python] Introspect partition keys and values in fragments.

It's not possible to programmatically determine the values of partition keys in a fragment. Fragments have a `partition_expression` attribute, but the `Expression` type doesn't allow any further introspection. I don't want to have to parse the string representation of the expression.
```python
In []: dataset.partitioning.schema
Out[]: 
year: int32
month: int32

In []: fragment = next(dataset.get_fragments())

In []: fragment.partition_expression
Out[]: <pyarrow.compute.Expression ((year == 2013) and (month == 1))>
```

My broader use case is more performant (speed and memory) aggregation of partitioned data. Using `pc._group_by` requires loaded arrays, so it ignores that the data is already partitioned. And iterating `get_fragments` is crippled if one can't identify the fragment.

### Component(s)

Python

jorisvandenbossche:
#### Rationale for this change

We have an existing "semi-private" `pyarrow.dataset._get_partition_keys` function (to get the partitioning field's key/value from the partition expression of a certain fragment). This is used by external projects (eg dask), and generally useful for advanced users, so let's just make it public. 

* Closes: #33825


danepitkin:
LGTM!

amol-:

PR #33865, opened 2023-01-25
GH-33864: [Go] Don't directly coerce cgo.Handle to unsafe.Pointer

### LINKED ISSUE #33864: [Go] Exported ArrowArray may panic with "misuse of an invalid handle"

```
panic: runtime/cgo: misuse of an invalid Handle

goroutine 17 [running, locked to thread]:
runtime/cgo.Handle.Value(...)
	/usr/local/go/src/runtime/cgo/handle.go:124
github.com/apache/arrow/go/v11/arrow/cdata.releaseExportedArray(0x55889925b4d0)
	/root/go/pkg/mod/github.com/apache/arrow/go/v11@v11.0.0-20230123220137-8449c553710a/arrow/cdata/exports.go:104 +0x1b8
```

We saw this in ADBC as well: https://github.com/apache/arrow-adbc/commit/5029f7b8ed1e60dd7d41f175af0abee6ccadb02e Experimentally, heap-allocating the Go Handle would avoid this at the cost of an extra indirection. That makes sense; the documentation for Handle states:

```go
// Some C functions accept a void* argument that points to an arbitrary
// data value supplied by the caller. It is not safe to coerce a cgo.Handle
// (an integer) to a Go unsafe.Pointer, but instead we can pass the address
// of the cgo.Handle to the void* parameter, as in this variant of the
// previous example:
```

(https://github.com/golang/go/blob/master/src/runtime/cgo/handle.go)

### Component(s)

Go

lidavidm:
### Rationale for this change

Converting cgo.Handle to unsafe.Pointer is not allowed as per the docs. Doing so appears to cause occasional panics at runtime, possibly because something corrupts or collects the handle unexpectedly.

Heap-allocate the Handle to preserve it.

### Are these changes tested?

We were unable to reproduce this outside of an integration test.

No user-facing changes.
* Closes: #33864

### DIFF
--- a/go/arrow/cdata/cdata_exports.go
+++ b/go/arrow/cdata/cdata_exports.go
@@ -396,7 +396,7 @@ func exportArray(arr arrow.Array, out *CArrowArray, outSchema *CArrowSchema) {
 
 	arr.Data().Retain()
 	h := cgo.NewHandle(arr.Data())
-	out.private_data = unsafe.Pointer(&h)
+	out.private_data = createHandle(h)
 	out.release = (*[0]byte)(C.goReleaseArray)
 	switch arr := arr.(type) {
 	case array.ListLike:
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
LGTM

PR #33866, opened 2023-01-25
GH-33872: [C++] Remove hacky shared_ptr construction in AppendScalar

### LINKED ISSUE #33872: [C++] Replace shared_ptr creation hack before AppendScalarImpl instantiation

`AppendScalarImpl` was created for appending scalars based on a contiguous array of `std::shared_ptr<Scalar>`, so when a single `Scalar` reference is passed, a dummy `shared_ptr` is created before the instantiation of `AppendScalarImpl`.

```cpp
  std::shared_ptr<Scalar> shared{const_cast<Scalar*>(&scalar), [](Scalar*) {}};
  return AppendScalarImpl{&shared, &shared + 1, n_repeats, this}.Convert();
```

This seems non-idiomatic [1] and likely inefficient as a control block needs to be heap-allocated for the `shared_ptr`. An alternative way (that simplifies `AppendScalarImpl` as well) is a custom iterator class the wraps the source iterator with an extra dereference operation.

[1] Pointers to `shared_ptr` as class members set a bad-style precedent

### Component(s)

C++

felipecrv:
`AppendScalarImpl` was created for appending scalars based on a contiguous array of `std::shared_ptr<Scalar>`, so when a single `Scalar` reference is passed, a dummy `shared_ptr` is created before the instantiation of `AppendScalarImpl`.

```cpp
  std::shared_ptr<Scalar> shared{const_cast<Scalar*>(&scalar), [](Scalar*) {}};
  return AppendScalarImpl{&shared, &shared + 1, n_repeats, this}.Convert();
```

This seems non-idiomatic [1] and likely inefficient as a control block needs to be heap-allocated for the `shared_ptr`. An alternative way (that simplifies `AppendScalarImpl` as well) is a custom iterator class the wraps the source iterator with an extra dereference operation.

[1] Pointers to `shared_ptr` as class members set a bad-style precedent

### Component(s)

C++

* Closes: #33872

### DIFF
--- a/cpp/src/arrow/array/builder_base.cc
+++ b/cpp/src/arrow/array/builder_base.cc
@@ -101,6 +101,7 @@ Status ArrayBuilder::Advance(int64_t elements) {
 
 namespace {
 
+template <typename ConstIterator>
 struct AppendScalarImpl {
   template <typename T>
   enable_if_t<has_c_type<T>::value || is_decimal_type<T>::value ||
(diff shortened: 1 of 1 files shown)


felipecrv:
@pitrou @lidavidm

lidavidm (approved):
Do you mind filing an issue for this? This goes beyond a minor PR IMO.

Generally I would be hesitant to add extra templated code here (readability, code bloat, etc.) However I think it makes more sense to cater for the use case of appending a single scalar as this PR does.

assignUser (changes requested):
Yes this is not a minor change, I will request changes to avoid a merge before we have a linked issue :)

felipecrv:
@lidavidm @assignUser done.

assignUser (commented):
Thanks!

assignUser (approved):

PR #33868, opened 2023-01-25
GH-33867: [Go][FlightSQL] Allow passing grpc call options to PreparedStatement methods

### LINKED ISSUE #33867: [Go][FlightSQL] Allow passing grpc call options to prepared statement methods

Currently you cannot pass `grpc.CallOption`s to Execute/ExecuteUpdate/Close on `flightsql.PreparedStatement` objects. We should allow passing call options so that they can get propagated, rather than just copying the call options which were used for the `Prepare` and just propagating those.

### Component(s)

Go

zeroshade:
### Rationale for this change
To allow propagating timeouts among other call options when using PreparedStatements with FlightSQL we need to update the function signatures to allow passing the call options to them.

The signatures for the methods on PreparedStatement objects have been updated to accept `opts... grpc.CallOption` as their last parameter. This shouldn't break any existing usage of the methods (since they are optional) but would cause a problem if consumers set up interfaces based on them.

* Closes: #33867

### DIFF
--- a/go/arrow/flight/flightsql/client.go
+++ b/go/arrow/flight/flightsql/client.go
@@ -355,7 +355,6 @@ func (c *Client) Prepare(ctx context.Context, mem memory.Allocator, query string
 
 	prep = &PreparedStatement{
 		client:        c,
-		opts:          opts,
 		handle:        result.PreparedStatementHandle,
 		datasetSchema: dsSchema,
 		paramSchema:   paramSchema,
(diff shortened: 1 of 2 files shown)


lidavidm (approved):
1. Do we want to label this breaking-change? 
2. Is there a way we can add a quick unit test?

zeroshade:
1. I was considering it but wasn't sure. It won't break any *usage* of the methods (since call sites won't be modified because it's an optional amount of params) but if consumers built interfaces that depended on the prepared statement methods, then those might break. thoughts?
2. I thought the tests were covering it but it looks like i never actually *put* any call options into the test suite call options. I'm updating the tests now so these get covered and tested.

lidavidm:
I think while we're doing major-version releases, let's be conservative and label it.

lidavidm (approved):

PR #33877, opened 2023-01-26
Java pushdown filter

noahfrn:
Conditional on #14287 - This adds an interface through JNI for strings to be passed, and then parsed into expressions to perform pushdown filtering via the datasets API.

### DIFF
--- a/java/dataset/src/main/cpp/jni_wrapper.cc
+++ b/java/dataset/src/main/cpp/jni_wrapper.cc
@@ -399,7 +399,7 @@ JNIEXPORT void JNICALL Java_org_apache_arrow_dataset_jni_JniWrapper_closeDataset
 /*
  * Class:     org_apache_arrow_dataset_jni_JniWrapper
  * Method:    createScanner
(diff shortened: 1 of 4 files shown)


zinking:
@NoahFournier  could this be pushed forward ?

NoahFournier:
Hey @zinking, I think the future direction for filtering in Arrow will likely be provided via substrait. To work around this for now, you can take a look at: [G-Research/datafusion-java](https://github.com/G-Research/datafusion-java) to see if that unblocks you right away.

zinking:
I think it's fine for now, there are plenty of serialization schemes for expressions.

NoahFournier:
What do you mean by serialization schemes for expressions ? Would be interested to know how you worked around this.

zinking:
> What do you mean by serialization schemes for expressions ? Would be interested to know how you worked around this.

you can see my workaround [here] (https://github.com/apache/arrow/commit/b1426b1460cfde0cec562f92380b84af1021de50) using json serializations. 

subtrait might be the correct direction. but it's not progressing as expected.

PR #33886, opened 2023-01-26
GH-33699: [C++] Increase timeout of c++ tests when running under valgrind and shorten long tests

### LINKED ISSUE #33699: [CI][C++] Nightly tests for valgrind have been failing for the last

Based on http://crossbow.voltrondata.com/
The [test-conda-cpp-valgrind](https://github.com/ursacomputing/crossbow/runs/10663265308) job has been failing on our nightlies for the last 26 days.
The latest error includes several test errors:
```
89% tests passed, 8 tests failed out of 70

Label Time Summary:
arrow-tests        = 1201.63 sec*proc (35 tests)
arrow_compute      = 1991.41 sec*proc (12 tests)
arrow_dataset      = 1085.71 sec*proc (10 tests)
arrow_substrait    =  45.52 sec*proc (1 test)
filesystem         =  70.61 sec*proc (3 tests)
parquet-tests      = 799.20 sec*proc (9 tests)
plasma-tests       = 294.49 sec*proc (3 tests)
unittest           = 5417.95 sec*proc (70 tests)

Total Test time (real) = 2787.52 sec

The following tests FAILED:
	 31 - arrow-compute-scalar-test (Timeout)
	 32 - arrow-compute-vector-test (Timeout)
	 37 - arrow-compute-hash-join-node-test (Timeout)
	 39 - arrow-compute-tpch-node-test (Timeout)
	 40 - arrow-compute-union-node-test (Timeout)
	 49 - arrow-dataset-scanner-test (Timeout)
	 50 - arrow-dataset-file-csv-test (Timeout)
	 65 - parquet-arrow-test (Timeout)
Errors while running CTest
```

### Component(s)

C++, Continuous Integration

westonpace:
This partially addresses #33699.  Once these tests are passing we can analyze the runtime and try and scope our valgrind coverage to something appropriate.
* Closes: #33699

### DIFF
--- a/cpp/src/arrow/compute/kernels/vector_sort_test.cc
+++ b/cpp/src/arrow/compute/kernels/vector_sort_test.cc
@@ -45,6 +45,15 @@ using internal::checked_pointer_cast;
 
 namespace compute {
 
+#ifdef ARROW_VALGRIND
+using RealArrowTypes = ::testing::Types<FloatType>;
+
+using IntegralArrowTypes = ::testing::Types<UInt32Type>;
+using TemporalArrowTypes = ::testing::Types<Date32Type, TimestampType, Time32Type>;
+
+using DecimalArrowTypes = ::testing::Types<Decimal128Type>;
+#endif
+
 std::vector<SortOrder> AllOrders() {
   return {SortOrder::Ascending, SortOrder::Descending};
 }
@@ -320,10 +329,15 @@ class TestNthToIndicesRandom : public TestNthToIndicesBase<ArrowType> {
   }
 };
 
+#ifdef ARROW_VALGRIND
+using NthToIndicesableTypes =
+    ::testing::Types<Int16Type, FloatType, Decimal128Type, StringType>;
+#else
 using NthToIndicesableTypes =
     ::testing::Types<UInt8Type, UInt16Type, UInt32Type, UInt64Type, Int8Type, Int16Type,
                      Int32Type, Int64Type, FloatType, DoubleType, Decimal128Type,
                      StringType>;
+#endif
 
 TYPED_TEST_SUITE(TestNthToIndicesRandom, NthToIndicesableTypes);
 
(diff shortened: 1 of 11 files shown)


westonpace:
@github-actions crossbow submit test-conda-cpp-valgrind

assignUser:
Looks like they are still timing out https://dev.azure.com/ursacomputing/crossbow/_build/results?buildId=43127&view=logs&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=d9b15392-e4ce-5e4c-0c8c-b69645229181&l=6947

westonpace:
I think splitting arrow-compute-scalar-test is inevitable at this point.  There are 1600 tests.  Right now, under valgrind, each takes an average of 700ms, which is not extremely long.  I tried to identify some of the worst culprits and while there are some things I can do I don't see any real solution.

The rest of the failures I can improve the timing on.

westonpace:
Looks like I'll have to tackle both issues after all :)

I broke the scalar test into multiple tests.  Some kernel tests were still pretty long so as most of those tests run on a large combination of types so I reduced the combinations when testing with valgrind.

The scanner test was using too much data.  I turns out that sorting an array with 2Mi rows is pretty slow when running in valgrind.

The CSV tests were slow because they were testing many combinations of compression and writing CSV files (even 50B CSV files) is very slow under valgrind (the conversion to string in particular).  I figure the various compressions are independently tested already and so removed those when running under valgrind.

The hash tests were simply running too many iterations.

westonpace:
@github-actions crossbow submit test-conda-cpp-valgrind

assignUser:
Awesome looks like the timeout is fixed :D

westonpace:
Yes, though the appveyor test is concerning (and confusing :)

js8544:
> Yes, though the appveyor test is concerning (and confusing :)

From the error message I guess it's a MSVC bug where it somehow treats the anonymous namespace in scalar_arithmetis_test.cc and the anonymous namespace in test_util.cc (in which the four typed ValidateOutput are defined) as the same namespace. The unqualified name lookup stops once it finds the name ValidateOutput in the same namespace.

westonpace:
> The fact that MSVC looks up names in source files seems even wilder to me....

Ah, that's probably it.  I will try fully qualifying.  I suspect that it isn't looking in different source files because it is a unity build so I think, in this case, scalar_arithmetic_test and test_util have been smashed together into the same file.

js8544:
> > The fact that MSVC looks up names in source files seems even wilder to me....
> 
> Ah, that's probably it. I will try fully qualifying. I suspect that it isn't looking in different source files because it is a unity build so I think, in this case, scalar_arithmetic_test and test_util have been smashed together into the same file.

Oh I see. Makes sense.

westonpace:
@js8544 's suggestion worked.  Looks like the remaining test failures are spurious.  Any last thoughts @assignUser ?

assignUser (approved):
I trust you on the implementation details and on the ci side it looks good :+1:

cpp/src/arrow/compute/kernels/vector_sort_test.cc
```diff
@@ -45,6 +45,15 @@ using internal::checked_pointer_cast;
 
 namespace compute {
 
+#ifdef ARROW_VALGRIND
+using RealArrowTypes = ::testing::Types<FloatType>;
```
pitrou, line 49: Ah, I didn't know you could redefine an alias with `using`.
As a style note, I would personally prefer if this created a distinct alias, to avoid confusion - and potential leaking accross unity builds?
(e.g. replace `RealArrowTypes` with `RealSortTypes` or something)

pitrou:
Thanks a lot @westonpace !

PR #12190, opened 2022-01-19
ARROW-15372: [C++][Gandiva] Gandiva now depends on boost/crc.hpp which is missing from the trimmed boost archive

kszucs:
See build error https://github.com/ursacomputing/crossbow/runs/4871392838?check_suite_focus=true#step:5:11762

PR #33890, opened 2023-01-26
GH-33851: [C++] Update bundled boost version

### LINKED ISSUE #33851: [C++] Bump version of bundled boost to 1.81.0

In addition to bumping the version in versions.txt, we need to update our trimmed version that we host (90% smaller than the full boost). See `cpp/build-support/trim-boost.sh`. Edit `BOOST_VERSION` and/or provide on the command line. Then upload the bundle to artifactory (see https://issues.apache.org/jira/browse/ARROW-14336). 

While you're in there, you should edit the comment about uploading to https://github.com/ursa-labs/thirdparty/releases/edit/latest, which we've switched off of. (It also appears in versions.txt but isn't used anywhere else anymore.) 

May also need to edit `Boost_ADDITIONAL_VERSIONS` in ThirdpartyToolchain.cmake. Check for other places in that file too.

### Component(s)

C++

paleolimbot:
This PR updates the bundled version of Boost, as it was suggested by a maintainer of CRAN (R packaging) that the older version of boost might be responsible for an Arrow build failure on Fedora/clang (#33819).

Closes #33851.

The resulting tarball also has to be uploaded to https://apache.jfrog.io/ui/native/arrow/thirdparty/7.0.0/ (right?) to kick in and almost certainly needs to be tested through a round of CI. I'm not sure how to do either of those things but perhaps @assignUser does?
* Closes: #33851


nealrichardson:
@kou mentioned on the issue: "All Apache Arrow committers/PMC members can upload a file to https://apache.jfrog.io/artifactory/arrow/thirdparty/ from https://apache.jfrog.io/ui/repos/tree/General/arrow/thirdparty with their ASF account."

nealrichardson:
There's also a curl command suggested here: https://github.com/apache/arrow/issues/18890#issuecomment-1377663132

assignUser:
The ui for artifactory is a bit slow but to upload one file it is fine :D (would need to login to get the api key anyway ^^) 
We also need to update the url and sha in `cpp/thridparty/versions.txt`.

Once that is uploaded and changed you can test it locally with `archery docker run -e boost_SOURCE=BUNDLED ubuntu-cpp`. For ci we probably want to run the full nightlies to make sure there are no hiccups: `@github-actions crossbow submit nightly-*`

assignUser:
@github actions crossbow submit r-binary-packages

That job is missing from nightly for some reason ^^

nealrichardson:
You can test this in your regular build env too without docker. That might make debug easier. 

You might want to `rm -rf cpp/build/*` before building, just so any previous boost_ep would be cleared out.

assignUser:
It's the internet and not you ;) https://github.com/apache/arrow/issues/33939

pitrou:
Why "7.0.0"? Does this version number have a special meaning?

nealrichardson:
I would guess that 7.0.0 was when we moved these bundles to artifactory and someone at the time thought it would be useful to associate them with an arrow version. But AFAICT it shouldn't matter, the dependency's version is in its filename, so boost_1.75.0 can live alongside boost_1.81.0.

nealrichardson:
> Is Boost the only dependency that gets downloaded in this way? (If it is, I'm happy to change the URL in this PR if somebody tells me where it is).

No, but it's the only one we alter in this way to reduce the size. The rest are just mirrors of the official sources for redundancy, so that if one site/CDN goes down, we're less likely to get a failed build. 

I wouldn't just change the URL since we've shipped several versions of the library with it. You could just put the boost bundle in the thirdparty directory (i.e. up one level) and update the URL for it. Then whenever we bump the dependency versions for everything else, we can put them in that directory. (You could make an issue for that.)

kou:
> Why "7.0.0"? Does this version number have a special meaning?

It doesn't have a special meaning. It’s just for historical reason.
We needed to create a release page to use GitHub as file upload location: https://github.com/ursa-labs/thirdparty/releases
"apache-arrow-7.0.0" is a tag for it.

We changed our thirdparty mirror to Artifactory in #12190. We can remove the tag name from URL in the PR but we keep only "7.0.0" in the PR.

We can use different URL such as https://apache.jfrog.io/artifactory/arrow/thirdparty/boost/boost_X_Y_Z.tar.gz for new files.

assignUser:
Well we can only change the URL for future versions. We have to keep the 7.0.0 folder around so previously released versions also work (or we decide to drop them at some point?) Also agree that that is a separate issue 

The hash join fail feels familiar but I have nothing specific.

kou:
test-alpine-linux-cpp doesn't use bundled Boost. So it's not related to this.

kou (approved):
+1

PR #33893, opened 2023-01-26
GH-32240: [C#] Add new Apache.Arrow.Compression package to implement IPC decompression

### LINKED ISSUE #32240: [C#] Add decompression support for Record Batches

adamreeve:
### Rationale for this change

This further addresses #32240 and is a follow up to PR #33603 to provide an implementation of the `ICompressionCodecFactory` interface in a new `Apache.Arrow.Compression` package. Making this a separate package means users who don't need IPC decompression support don't need to pull in extra dependencies.

Adds a new `Apache.Arrow.Compression` package and moves the existing compression codec implementations used for testing into this package.

### Are these changes tested?

There are unit tests verifying the decompression support, but this also affects the release scripts and I'm not sure how to fully test these.

Yes, this adds a new package users can install for IPC decompression support, so documentation has been updated.
* Closes: #32240


csharp/src/Apache.Arrow.Compression/Apache.Arrow.Compression.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="CommunityToolkit.HighPerformance" Version="8.0.0" />
```
adamreeve, line 8: This dependency is only needed for the  `AsStream` extension methods on `Memory<byte>` and `ReadOnlyMemory<byte>` which could be re-implemented if we wanted to minimize extra dependencies.

assignUser:
@github-actions crossbow submit nuget verify-rc-source-csharp*

assignUser:
@adamreeve I have kicked of some additional CI jobs that should be sufficent to test the changes to the ci side of things and building the nuget packages. The release script is tested with `ci/scripts/release_test.sh` but the test themselves are implemented in Ruby. 

@kou can you give details on if additional tests are needed for this PR and if so how to implement them?

assignUser:
looks like the binary is properly uploaded: https://github.com/ursacomputing/crossbow/releases/tag/actions-c460b32b4b-github-nuget

assignUser (commented):
CI side looks good, cant comment on the C# +1

kou:
@github-actions crossbow submit nuget

kou:
@eerhardt Could you review this?

kou:
We don't have a test for `post-06-csharp.sh` and don't need it. Because it just pushes built packages to NuGet.

csharp/src/Apache.Arrow.Compression/Apache.Arrow.Compression.csproj
westonpace: I think this is fine.

csharp/test/Apache.Arrow.Compression.Tests/ArrowFileReaderTests.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            for (var i = 0; i < numRows; ++i)
+            {
+                Assert.Equal(i, intArray.GetValue(i));
+                Assert.True(Math.Abs(floatArray.GetValue(i).Value - 0.1f * i) < 1.0e-6);
+            }
+        }
+
+        [Fact]
+        public void ErrorReadingCompressedFileWithoutCodecFactory()
```
westonpace, line 53: Do you want an equivalent test on the stream reader?

westonpace (approved):

csharp/test/Apache.Arrow.Compression.Tests/ArrowFileReaderTests.cs
adamreeve: Good point, I hadn't bothered as it would mostly test the same code path, but the tests shouldn't make assumptions like that so I've added one now.

csharp/src/Apache.Arrow.Compression/Apache.Arrow.Compression.csproj
eerhardt: I found that the LZ4 library offers APIs that takes Spans. We should just use those APIs instead, and then we can get rid of this dependency.

https://github.com/MiloszKrajewski/K4os.Compression.LZ4/blob/fa8b8e038b500d565efe12769db097852a28ddf7/src/K4os.Compression.LZ4/LZ4Codec.cs#L139-L144

Note that there are 2 LZ4 libraries - one for streams (which this pr is using now) and one that doesn't use streams - https://www.nuget.org/packages/K4os.Compression.LZ4

csharp/src/Apache.Arrow.Compression/Lz4CompressionCodec.cs
```diff
@@ -18,7 +18,7 @@
 using CommunityToolkit.HighPerformance;
 using K4os.Compression.LZ4.Streams;
 
-namespace Apache.Arrow.Tests.Compression
+namespace Apache.Arrow.Compression
 {
     internal sealed class Lz4CompressionCodec : ICompressionCodec
```
eerhardt, line 23: Since the `Lz4CompressionCodec` doesn't need to be disposed, would it make sense to just have a singleton object and return that from `CreateCodec`? It would save on a tiny allocation every time this is called.

(I have the same comment for the `NoOpBufferCreator` in the previous PR, but I missed my window to add the comment there. Maybe we could update that as well?)

csharp/src/Apache.Arrow.Compression/Apache.Arrow.Compression.csproj
adamreeve: The two libraries actually implement different formats. The `K4os.Compression.LZ4.Streams` library implements the LZ4 frame format, and the `K4os.Compression.LZ4` library implements the LZ4 block format. Arrow IPC uses the frame format (https://github.com/apache/arrow/blob/apache-arrow-11.0.0/format/Message.fbs#L46-L48), so we need to use the streams library.

csharp/src/Apache.Arrow.Compression/Lz4CompressionCodec.cs
adamreeve: Yes that would make sense, I will update both.

csharp/src/Apache.Arrow.Compression/Apache.Arrow.Compression.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
```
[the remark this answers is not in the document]
adamreeve: Yes there's no reason to target both, I'm not sure why I did that.

csharp/src/Apache.Arrow.Compression/Apache.Arrow.Compression.csproj
eerhardt: I've opened https://github.com/MiloszKrajewski/K4os.Compression.LZ4/issues/79 to request this is added to the `K4os.Compression.LZ4` library.

adamreeve: Thanks Eric, your issue prompted me to dig more into the `K4os` library though and I realised that since I initially started working on this, a new API has been added that allows using the frame format with more types than just `Stream`, so I've switched to this and removed the `CommunityToolkit.HighPerformance` dependency.

csharp/test/Apache.Arrow.Compression.Tests/Apache.Arrow.Compression.Tests.csproj
```diff
@@ -0,0 +1,32 @@
...
+    <TargetFramework>net7.0</TargetFramework>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.5.0" />
+    <PackageReference Include="xunit" Version="2.4.0" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
+    <PackageReference Include="coverlet.collector" Version="1.2.0" />
```
eerhardt, line 13:
```suggestion
```

I know the Arrow.Flight.Tests have this, but we don't actually use it. We should remove them until they are used to reduce restore times.

csharp/test/Apache.Arrow.Compression.Tests/Apache.Arrow.Compression.Tests.csproj
```diff
@@ -0,0 +1,32 @@
...
+
+  <ItemGroup>
+    <None Remove="Resources\ipc_lz4_compression.arrow" />
+    <EmbeddedResource Include="Resources\ipc_lz4_compression.arrow" />
+    <None Remove="Resources\ipc_lz4_compression.arrow_stream" />
+    <EmbeddedResource Include="Resources\ipc_lz4_compression.arrow_stream" />
+    <None Remove="Resources\ipc_zstd_compression.arrow" />
+    <EmbeddedResource Include="Resources\ipc_zstd_compression.arrow" />
+    <None Remove="Resources\ipc_zstd_compression.arrow_stream" />
+    <EmbeddedResource Include="Resources\ipc_zstd_compression.arrow_stream" />
```
eerhardt, line 29:
```suggestion
    <EmbeddedResource Include="Resources\ipc_lz4_compression.arrow" />
    <EmbeddedResource Include="Resources\ipc_lz4_compression.arrow_stream" />
    <EmbeddedResource Include="Resources\ipc_zstd_compression.arrow" />
    <EmbeddedResource Include="Resources\ipc_zstd_compression.arrow_stream" />
```

Are the `None Remove`s necessary?

eerhardt (commented):
Just a couple last nits. I think this can be merged after they are addressed.

Thanks for the great work here, @adamreeve!

csharp/test/Apache.Arrow.Compression.Tests/Apache.Arrow.Compression.Tests.csproj
adamreeve: No, I don't think there's any need for these, thanks

eerhardt (approved):
LGTM. Thanks again!

I'll merge this soon, assuming no more feedback is submitted.

PR #33897, opened 2023-01-27
GH-33652: [C++][Parquet] Add interface total_compressed_bytes_written

### LINKED ISSUE #33652: [C++][Parquet] Interface total_bytes_written is Confusing

Hi, all, when using parquet's `total_bytes_written`, I found it really confusing.

In `RowGroupWriter`:

```c++
class PARQUET_EXPORT RowGroupWriter {
 public:
  // Forward declare a virtual class 'Contents' to aid dependency injection and more
  // easily create test fixtures
  // An implementation of the Contents class is defined in the .cc file
  struct Contents {
    virtual ~Contents() = default;
    virtual int num_columns() const = 0;
    virtual int64_t num_rows() const = 0;

    // to be used only with ParquetFileWriter::AppendRowGroup
    virtual ColumnWriter* NextColumn() = 0;
    // to be used only with ParquetFileWriter::AppendBufferedRowGroup
    virtual ColumnWriter* column(int i) = 0;

    virtual int current_column() const = 0;
    virtual void Close() = 0;

    // total bytes written by the page writer
    virtual int64_t total_bytes_written() const = 0;
    // total bytes still compressed but not written
    virtual int64_t total_compressed_bytes() const = 0;

    virtual bool buffered() const = 0;
  };
```

1. Syntax of `total_compressed_bytes()` is clear
2. total_bytes_written means "total bytes written by page writer".

And in `ColumnWriter`, the interface says:

```c++
  /// \brief The total size of the compressed pages + page headers. Some values
  /// might be still buffered and not written to a page yet
  virtual int64_t total_compressed_bytes() const = 0;

  /// \brief The total number of bytes written as serialized data and
  /// dictionary pages to the ColumnChunk so far
  virtual int64_t total_bytes_written() const = 0;
```

In `ColumnWriterImpl`, it's:

```c++
  // Records the total number of uncompressed bytes written by the serializer
  int64_t total_bytes_written_;

  // Records the current number of compressed bytes in a column
  int64_t total_compressed_bytes_;
```

So, in fact, the `total_bytes_written()` in `RowGroup` just means the uncompressed bytes it output? Is there any interface for compressed bytes? In non-buffer mode, we can just call `sink_.Tell()`, but in buffer mode, how can we get compressed-output length?

### Component(s)

C++, Parquet

mapleFU:

Talked in https://github.com/apache/arrow/issues/33652 . Main issue is that `total_bytes_written` is confusing, because it only tells the size of "uncompressed" bytes size written. For buffered page writer, the actually written size cannot be known until all column chunk is written and call `sink_.Tell()`.

I'd like to:
* [x] Add interface for PageWriter
* [x] Add interface for ColumnWriter
* [x] Add interface for RowGroupWriter
* [x] Testing them

### Are these changes tested?

I'd like to test it later.

User can get excatly size written by: `total_compressed_bytes_written() + total_compressed_bytes()` for ColumnWriter before finish writing.

* Closes: #33652


mapleFU:
@pitrou @wjones127 I think the main code is already finished. If you think the interface and description is ok, I will add some simple tests here.

wjones127:
This looks good so far!

mapleFU:
Well, by the way, I found a trickey issue. In `RowGroupSerializer`, it have both:

```c++
int64_t total_bytes_written() const override;
```

and member:

```c++
int64_t total_bytes_written_;
```

When `RowGroupSerializer` is closed, all bytes will accumulated to `total_bytes_written_`, and use it to finish metadata. But `total_bytes_written()` will always got 0. Should I add comment for this behavior? Or making it return `total_bytes_written_` once closed?

@pitrou @wjones127 @wgtmac

wjones127:
>  Or making it return total_bytes_written_ once closed?

I think this is better. I would rather the behavior is not surprising than have to document special cases like that.

cpp/src/parquet/column_writer.h
```diff
@@ -103,6 +103,8 @@ class PARQUET_EXPORT PageWriter {
   // Return the number of uncompressed bytes written (including header size)
   virtual int64_t WriteDictionaryPage(const DictionaryPage& page) = 0;
 
+  virtual int64_t total_compressed_bytes_written() const = 0;
```
wgtmac, line 106: Add a comment?

cpp/src/parquet/column_writer.h
```diff
@@ -136,8 +138,15 @@ class PARQUET_EXPORT ColumnWriter {
 
   /// \brief The total number of bytes written as serialized data and
```
wgtmac, line 139:
```suggestion
  /// \brief The total number of uncompressed bytes written as serialized data and
```

wgtmac: Is it enough?

cpp/src/parquet/column_writer.h
```diff
@@ -136,8 +138,15 @@ class PARQUET_EXPORT ColumnWriter {
...
   /// dictionary pages to the ColumnChunk so far
+  /// These bytes are uncompressed bytes.
   virtual int64_t total_bytes_written() const = 0;
 
+  /// \brief The total number of bytes written as serialized data and
+  /// dictionary pages to the ColumnChunk so far.
+  /// If the column is uncompressed, the value would be equal to
```
wgtmac, line 146: Is this true? Will total_compressed_bytes_written() be greater than total_bytes_written() because of page headers and column chunk metadata?

cpp/src/parquet/file_writer.h
```diff
@@ -90,8 +92,12 @@ class PARQUET_EXPORT RowGroupWriter {
    */
   int64_t num_rows() const;
 
+  /// \brief total uncompressed bytes written by the page writer
   int64_t total_bytes_written() const;
+  /// \brief total bytes still compressed but not written
```
wgtmac, line 97: Should we explain it can only exist in the buffered row group?

cpp/src/parquet/column_writer.h
mapleFU: Column Chunk Metadata will not be added to count. And both the compressed size and uncompressed would be written to column chunk.

page header will not be compressed.

cpp/src/parquet/file_writer.h
mapleFU: Sure, I can write code to verify this.

cpp/src/parquet/column_writer.h
mapleFU: Let me add these comment, wait a minute

mapleFU:
@wgtmac Some comment fixed.

By the way, after writing some comments, I thinks previous comment is really leaky... Maybe users have already misused these interface...

wjones127:
Yes, I think adding the comment should help understand. I wish these parts of the codebase were better documented.

Are you still planning on adding tests?

mapleFU:
> Yes, I think adding the comment should help understand. I wish these parts of the codebase were better documented.
> 
> Are you still planning on adding tests?

Yes, I will, but I'm a little busy this week, and adding tests for bytes_written is boring and troublesome, because we only know that I "would no less than xx bytes" or "should less than xx bytes", but actually don't know it's really length when writing...

I'll add it this weekend...

cpp/src/parquet/stream_writer.cc
```diff
@@ -294,7 +294,6 @@ void StreamWriter::EndRow() {
     }
     // Initialize for each row with size already written
     // (compressed + uncompressed).
-    //
     row_group_size_ = row_group_writer_->total_bytes_written() +
                       row_group_writer_->total_compressed_bytes();
```
mapleFU, line 298: I guess here is better to use `total_compressed_bytes_written`

mapleFU:
@wjones127 Test added, maybe I'm bad at organize testing, you can comment or edit it directly..

cpp/src/parquet/column_writer.cc
```diff
@@ -526,7 +530,13 @@ class SerializedPageWriter : public PageWriter {
   int64_t num_values_;
   int64_t dictionary_page_offset_;
   int64_t data_page_offset_;
+  // The uncompressed page size the page writer already
+  //  written.
```
westonpace, line 534:
```suggestion
  // The uncompressed page size the page writer has already
  //  written.
```

cpp/src/parquet/column_writer.cc
```diff
@@ -526,7 +530,13 @@ class SerializedPageWriter : public PageWriter {
...
   int64_t total_uncompressed_size_;
+  // The compressed page size the page writer already
+  //  written.
```
westonpace, line 537:
```suggestion
  // The compressed page size the page writer has already
  //  written.
```

cpp/src/parquet/column_writer.h
```diff
@@ -131,14 +134,23 @@ class PARQUET_EXPORT ColumnWriter {
   /// \brief The number of rows written so far
   virtual int64_t rows_written() const = 0;
 
-  /// \brief The total size of the compressed pages + page headers. Some values
-  /// might be still buffered and not written to a page yet
+  /// \brief The total size of the compressed pages + page headers. Values
+  /// are still buffered and not written to a pager yet
```
westonpace, line 138: Is this the number of bytes that are buffered but not yet written?

cpp/src/parquet/column_writer.h
westonpace: Do both of them record bytes from dictionary pages written?

cpp/src/parquet/column_writer_test.cc
```diff
@@ -1143,5 +1143,143 @@ TEST(TestColumnWriter, WriteDataPageV2Header) {
...
+    schema_descriptor_ = std::make_unique<SchemaDescriptor>();
+    schema_descriptor_->Init(node_);
+  }
+
+  std::shared_ptr<parquet::Int32Writer> BuildWriter(Compression::type compression,
+                                                    bool buffered,
+                                                    bool enable_dictionary = false) {
+    auto builder = WriterProperties::Builder();
+    builder.disable_dictionary()
+        ->disable_dictionary()
```
westonpace, line 1165:
```suggestion
```
You don't need this since you enable/disable below.

cpp/src/parquet/column_writer_test.cc
```diff
@@ -1143,5 +1143,143 @@ TEST(TestColumnWriter, WriteDataPageV2Header) {
...
+TEST_F(TestSizeEstimated, NonBufferedDictionary) {
+  auto required_writer = this->BuildWriter(Compression::UNCOMPRESSED, false, true);
+
+  // Write half page
+  // for dict, keep value equal
+  int32_t dict_value = 1;
+  for (int32_t i = 0; i < 50; i++) {
+    required_writer->WriteBatch(1, nullptr, nullptr, &dict_value);
+  }
+  // Page flushed, check size
```
westonpace, line 1244:
```suggestion
```
I don't think the page is flushed yet based on the comment below.

cpp/src/parquet/column_writer_test.cc
```diff
@@ -1143,5 +1143,143 @@ TEST(TestColumnWriter, WriteDataPageV2Header) {
...
+  EXPECT_EQ(0, required_writer->total_bytes_written());
+  EXPECT_EQ(0, required_writer->total_compressed_bytes());
+  EXPECT_EQ(0, required_writer->total_compressed_bytes_written());
+  // write a huge batch to trigger page flush
+  for (int32_t i = 0; i < 50000; i++) {
```
westonpace, line 1249: Just a question.  Why do we need such a large batch to flush a page when writing a dictionary?

cpp/src/parquet/column_writer_test.cc
```diff
@@ -1143,5 +1143,143 @@ TEST(TestColumnWriter, WriteDataPageV2Header) {
...
+#ifndef ARROW_WITH_SNAPPY
+  GTEST_SKIP() << "Test requires snappy compression";
+#endif
+  auto required_writer = this->BuildWriter(Compression::SNAPPY, true);
+
+  // Write half page
+  for (int32_t i = 0; i < 50; i++) {
+    required_writer->WriteBatch(1, nullptr, nullptr, &i);
+  }
+  // Page flushed, check size
```
westonpace, line 1270:
```suggestion
```
If it's flushed then why is bytes written 0 here?

westonpace (commented):
I have a few questions but I think this would be a valuable addition.  I don't know much about the parquet writer internals but the feature looks pretty straightforward.  If I can remember I will try and hook this up to the dataset writer in a month or two so we can use support sizing row groups by bytes instead of rows.

cpp/src/parquet/column_writer_test.cc
mapleFU: In parquet-format v2, Dictionary has 2 pages:
1. (Dict index page) PLAIN
2. (Data Page) RLE_DICTIONARY

`RLE_DICTIONARY` uses `RLE/Bit-packing` as encoding. It can well compress the page when all values are equal. Arrow C++ Parquet default use `size` as Page Split condition. So, I use a large value here to trigger page flush.

mapleFU:
@westonpace Let me explain it detail here:

Parquet, which may contains multiple rowgroups, rowgroups may contain multiple columns, and columns may contain multiple pages.

For a `ParquetFileWriter`, use can use `AppendBufferedRowGroup` and `AppendRowGroup` to acquire a buffered and unbuffered rowgroup writer.

For buffered `RowGroupWriter`, it can:

```
for column in columns:
  columns[i] = file_writer->AppendBufferedRowGroup()

for value in value-batches:
  column_writer = columns[i]->column()
  typed_column_writer->Write()
```

For unbuffered `RowGroupWriter`, it can:

```
for column in columns:
  columns[i] = file_writer->AppendRowGroup()
  writeAllValues(columns[i])
```

Below are user-known api which you may familiar with. Now let's explain the interface of this patch, and why previous patch is trickey.

An active Column uses `(Typed)ColumnWriter` to write values, and ColumnWriter holds a `PageWriter` to write page to sink.  If no dict is enabled, it would be like:

## Type1: non-buffered-non-dict:

```
Only Have One Active Column Writer
Column2: |  DataPage1 |   DataPage2 | Buffered-Values that didn't becoming a page |
```

For it, the output bytes are:
1. `RowGroupWriter.total_compressed_bytes`: 0
2.`RowGroupWriter.total_bytes_written`: It **doesn't** mean the `Column1-Data-Bytes + Column2-DataPage1 + Column2-DataPage2`. In fact, it means the **uncompressed** `Column1-IO-Bytes + Column2-DataPage1 + Column2-DataPage2` if compression like zstd enabled
3. (New in this patch) `RowGroupWriter.total_bytes_written`: It means the `Column1-Data-Bytes + Column2-DataPage1 + Column2-DataPage2`. It will be a little less than really io-bytes, because page and column may write some metadata. But they will be close
4. `ColumnWriter.total_compressed_bytes`: 0
5. `ColumnWriter.total_bytes_written`: **uncompressed** `Column2-DataPage1 + Column2-DataPage2`
6. (New in this patch) `ColumnWriter.total_compressed_bytes_written`: It means the `Column2-DataPage1 + Column2-DataPage2`.

## Type-2 non-buffered-dict:

Now, things would be a little different. `ColumnWriter` has a `data_pages_` vector for pages.

```
Only Have One Active Column Writer
Column2:  | Buffered-Values that didn't becoming a page |
* Column2: buffers DataPage1  and  DataPage2
* Column2: Building Dictionary Page
```

Now, let's revisit the ColumnWriter's size:
1. `ColumnWriter.total_compressed_bytes`: **compressed** `Column2-DataPage1 + Column2-DataPage2`
2. `ColumnWriter.total_bytes_written`: 0
3. (New in this patch) `ColumnWriter.total_compressed_bytes_written`: 0

## Type3: buffered-non-dict:

```
Have Two Active Column Writer
Column2: |  DataPage1 |  Buffered-Values that didn't becoming a page |
Column2: |  DataPage1 |   DataPage2 | Buffered-Values that didn't becoming a page |
```

1. `ColumnWriter.total_compressed_bytes`: Compressed `c1.DataPage1 + c2.DataPage1 + c3.DataPage2`
2. `ColumnWriter.total_bytes_written`: 0
3. (New in this patch) `ColumnWriter.total_compressed_bytes_written`: 0

cpp/src/parquet/column_writer.h
mapleFU: No, it's a bit hard to understand, maybe you can take a look at https://github.com/apache/arrow/pull/33897#issuecomment-1440131022

cpp/src/parquet/column_writer_test.cc
mapleFU: Yes, nice catch

mapleFU:
And you may curious that what's the size didn't writen to a page, or holded by Dictionary. That's `EstimatedDataEncodedSize()`.

westonpace (approved):
Thanks for the clarifications.  I'll let @wjones127 take another look but the overall descriptions here make sense to me.

cpp/src/parquet/column_writer.h
```diff
@@ -104,12 +104,15 @@ class PARQUET_EXPORT PageWriter {
   // Return the number of uncompressed bytes written (including header size)
   virtual int64_t WriteDictionaryPage(const DictionaryPage& page) = 0;
 
+  /// \brief The total number of bytes written as serialized data and
```
wgtmac, line 107: What does `serialized data` mean here? Flushed data pages?

mapleFU: It means written to sink. In fact, the comment is pasted from below:

```c++
  /// \brief The total number of bytes written as serialized data and
  /// dictionary pages to the ColumnChunk so far
  virtual int64_t total_bytes_written() const = 0;
```

in `ColumnWriter`

mapleFU: (By the way, the way `Serialize` may means write to sink, which is like `SerializedPageWriter`. If I'm wrong, please correct me)

wgtmac: Simply put
- written: compressed and flushed to sink
- compressed: compressed only and not flushed to sink

Am I correct?

wgtmac:
> @westonpace Let me explain it detail here:
> 
> Parquet, which may contains multiple rowgroups, rowgroups may contain multiple columns, and columns may contain multiple pages.
> 
> For a `ParquetFileWriter`, use can use `AppendBufferedRowGroup` and `AppendRowGroup` to acquire a buffered and unbuffered rowgroup writer.
> 
> For buffered `RowGroupWriter`, it can:
> 
> ```
> for column in columns:
>   columns[i] = file_writer->AppendBufferedRowGroup()
> 
> for value in value-batches:
>   column_writer = columns[i]->column()
>   typed_column_writer->Write()
> ```
> 
> For unbuffered `RowGroupWriter`, it can:
> 
> ```
> for column in columns:
>   columns[i] = file_writer->AppendRowGroup()
>   writeAllValues(columns[i])
> ```
> 
> Above are user-known api which you may familiar with. Now let's explain the interface of this patch, and why previous patch is trickey.
> 
> An active Column uses `(Typed)ColumnWriter` to write values, and ColumnWriter holds a `PageWriter` to write page to sink. If no dict is enabled, it would be like:
> 
> ## Type1: non-buffered-non-dict:
> ```
> Only Have One Active Column Writer
> Column2: |  DataPage1 |   DataPage2 | Buffered-Values that didn't becoming a page |
> ```
> 
> For it, the output bytes are:
> 
> 1. `RowGroupWriter.total_compressed_bytes`: 0
>    2.`RowGroupWriter.total_bytes_written`: It **doesn't** mean the `Column1-Data-Bytes + Column2-DataPage1 + Column2-DataPage2`. In fact, it means the **uncompressed** `Column1-IO-Bytes + Column2-DataPage1 + Column2-DataPage2` if compression like zstd enabled
> 2. (New in this patch) `RowGroupWriter.total_bytes_written`: It means the `Column1-Data-Bytes + Column2-DataPage1 + Column2-DataPage2`. It will be a little less than really io-bytes, because page and column may write some metadata. But they will be close
> 3. `ColumnWriter.total_compressed_bytes`: 0
> 4. `ColumnWriter.total_bytes_written`: **uncompressed** `Column2-DataPage1 + Column2-DataPage2`
> 5. (New in this patch) `ColumnWriter.total_compressed_bytes_written`: It means the `Column2-DataPage1 + Column2-DataPage2`.
> 
> ## Type-2 non-buffered-dict:
> Now, things would be a little different. `ColumnWriter` has a `data_pages_` vector for pages.
> 
> ```
> Only Have One Active Column Writer
> Column2:  | Buffered-Values that didn't becoming a page |
> * Column2: buffers DataPage1  and  DataPage2
> * Column2: Building Dictionary Page
> ```
> 
> Now, let's revisit the ColumnWriter's size:
> 
> 1. `ColumnWriter.total_compressed_bytes`: **compressed** `Column2-DataPage1 + Column2-DataPage2`
> 2. `ColumnWriter.total_bytes_written`: 0
> 3. (New in this patch) `ColumnWriter.total_compressed_bytes_written`: 0
> 
> ## Type3: buffered-non-dict:
> ```
> Have Two Active Column Writer
> 
> (All-Values are Buffered)
> 
> Column1: |  DataPage1 |  Buffered-Values that didn't becoming a page | 
> Column2: |  DataPage1 |   DataPage2 | Buffered-Values that didn't becoming a page |
> ```
> 
> 1. `ColumnWriter.total_compressed_bytes`: Compressed `c1.DataPage1 + c2.DataPage1 + c3.DataPage2`
> 2. `ColumnWriter.total_bytes_written`: 0
> 3. (New in this patch) `ColumnWriter.total_compressed_bytes_written`: 0

It seems that every reviewer or reader may ask the similar questions. It would be worthy to add this explanation somewhere in the code.

mapleFU:
@wgtmac https://github.com/apache/arrow/pull/33897#discussion_r1118315497

Yes. `Compressed` is a trickey syntax and misleading-prune. It just means that "Page is built, but not flush to sink yet"...

cpp/src/parquet/column_writer.h
mapleFU: Yes

mapleFU:
@wjones127 Mind take a look?

wgtmac (approved):
Looking good on my side. Thanks for doing the dirty job!

wjones127 (approved):
This looks good. All the metrics can be a bit confusing, but this is a helpful improvement. Thanks @mapleFU

Issue #33850, opened 2023-01-24
[C++] Allow Substrait's default extension provider to be configured


westonpace:
### Describe the enhancement requested

A custom extension provider can be specified when running a query.  However, in some cases (e.g. when Acero is embedded in a larger program that provides its own extension handlers) it would be simpler to change the default instead of having to specify the custom provider with every query run.

### Component(s)

C++

rtpsw:
take

rtpsw:
@westonpace, @icexelloss, this issue may be a no-op, because [`ExtensionProvider::kDefaultExtensionProvider` is public](https://github.com/apache/arrow/blob/177a0d4400e61f44dbd045ce05150bba28dd1d74/cpp/src/arrow/engine/substrait/options.h#L77) and can already be directly set. We could have some kind of guarding around its setting, if desired.

icexelloss:
IMO `kDefaultExtensionProvider` is not a public API so we should not change it directly. The API can be as simple as changing that variable under the hood but we still need a nice public API for it.

westonpace:
I agree with @icexelloss .  We should move `kDefaultExtensionProvider` into the `.cc` file and/or make it const.

rtpsw:
We're on the same page on that. My question is whether we need anything more, like some kind of guard, in the implementation of the API for setting.

westonpace:
You mean like a mutex / lock_guard?  Yes, that would be a good idea.  Or am I misunderstanding?

rtpsw:
Yes, and wanted to make sure there's nothing else needed that I might be missing.

westonpace:
I think that would be sufficient.

PR #33909, opened 2023-01-28
GH-33899: [C++] Add NamedTapRel relation as a Substrait extension

### LINKED ISSUE #33899: [C++] Add `NamedTapRel` relation as a Substrait extension

A `NamedTapRel` is a relation that passes the input through but also process it for some side effect. A useful example for a side-effect is writing to external storage.

### Component(s)

C++

rtpsw:
See #33899. This PR adds `NamedTapRel` and a simple test case with a no-op tap (i.e., just passing-through).
* Closes: #33899

### DIFF
--- a/cpp/proto/substrait/extension_rels.proto
+++ b/cpp/proto/substrait/extension_rels.proto
@@ -44,3 +44,9 @@ message AsOfJoinRel {
     repeated .substrait.Expression by = 2;
   }
 }
+
+message NamedTapRel {
+  string kind = 1;
+  string name = 2;
+  repeated string columns = 3;
+}
--- a/cpp/proto/substrait/extension_rels.proto
+++ b/cpp/proto/substrait/extension_rels.proto
@@ -45,8 +45,15 @@ message AsOfJoinRel {
   }
 }
 
+// Named tap relation
+//
+// A tap is a relation having a single input relation that it passes through, while also
+// causing some side-effect, e.g., writing to external storage.
 message NamedTapRel {
+  // The kind of tap
   string kind = 1;
+  // A name used to configure the tap, e.g., a URI defining the destination of writing
   string name = 2;
+  // Column names for the tap's output. Their number must be 0 or equal to the input's.
   repeated string columns = 3;
 }
--- a/cpp/proto/substrait/extension_rels.proto
+++ b/cpp/proto/substrait/extension_rels.proto
@@ -54,6 +54,7 @@ message NamedTapRel {
   string kind = 1;
   // A name used to configure the tap, e.g., a URI defining the destination of writing
   string name = 2;
-  // Column names for the tap's output. Their number must be 0 or equal to the input's.
+  // Column names for the tap's output. If specified there must be one name per field.
+  // If empty, field names will be automatically generated.
   repeated string columns = 3;
 }
--- a/cpp/src/arrow/engine/substrait/options.cc
+++ b/cpp/src/arrow/engine/substrait/options.cc
@@ -28,6 +28,19 @@
 namespace arrow {
 namespace engine {
 
+namespace {
+
+std::vector<compute::Declaration::Input> MakeInputDeclarations(
+    const std::vector<DeclarationInfo> inputs) {
+  std::vector<compute::Declaration::Input> input_decls(inputs.size());
+  for (size_t i = 0; i < inputs.size(); i++) {
+    input_decls[i] = inputs[i].declaration;
+  }
+  return input_decls;
+}
+
+}  // namespace
+
 class BaseExtensionProvider : public ExtensionProvider {
  public:
   Result<RelationInfo> MakeRel(const std::vector<DeclarationInfo>& inputs,
@@ -51,6 +64,11 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
       substrait_ext::AsOfJoinRel as_of_join_rel;
       rel.UnpackTo(&as_of_join_rel);
       return MakeAsOfJoinRel(inputs, as_of_join_rel, ext_set);
+    }
+    if (rel.Is<substrait_ext::NamedTapRel>()) {
+      substrait_ext::NamedTapRel named_tap_rel;
+      rel.UnpackTo(&named_tap_rel);
+      return MakeNamedTapRel(inputs, named_tap_rel, ext_set);
     }
     return Status::NotImplemented("Unrecognized extension in Susbstrait plan: ",
                                   rel.DebugString());
@@ -111,14 +129,42 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
     compute::AsofJoinNodeOptions asofjoin_node_opts{std::move(input_keys), tolerance};
 
     // declaration
-    std::vector<compute::Declaration::Input> input_decls(inputs.size());
-    for (size_t i = 0; i < inputs.size(); i++) {
-      input_decls[i] = inputs[i].declaration;
-    }
+    auto input_decls = MakeInputDeclarations(inputs);
     return RelationInfo{
         {compute::Declaration("asofjoin", input_decls, std::move(asofjoin_node_opts)),
          std::move(schema)},
         std::move(field_output_indices)};
+  }
+
+  Result<RelationInfo> MakeNamedTapRel(const std::vector<DeclarationInfo>& inputs,
+                                       const substrait_ext::NamedTapRel& named_tap_rel,
+                                       const ExtensionSet& ext_set) {
+    if (inputs.size() != 1) {
+      return Status::Invalid(
+          "substrait_ext::NamedTapNode requires a single table but got: ", inputs.size());
+    }
+
+    auto schema = inputs[0].output_schema;
+    int num_fields = schema->num_fields();
+    if (named_tap_rel.columns_size() != num_fields) {
+      return Status::Invalid("Got ", named_tap_rel.columns_size(),
+                             " NamedTapRel columns but expected ", num_fields);
+    }
+    int i = 0;
+    FieldVector fields(static_cast<size_t>(num_fields));
+    for (const auto& column : named_tap_rel.columns()) {
+      fields[i] = field(column, schema->field(i)->type());
+      i++;
+    }
+    auto renamed_schema = arrow::schema(std::move(fields));
+    std::shared_ptr<compute::ExecNodeOptions> named_tap_opts =
+        std::make_shared<NamedTapNodeOptions>(named_tap_rel.name(),
+                                              std::move(renamed_schema));
+    auto input_decls = MakeInputDeclarations(inputs);
+    return RelationInfo{{compute::Declaration(named_tap_rel.kind(), input_decls,
+                                              std::move(named_tap_opts)),
+                         std::move(renamed_schema)},
+                        std::nullopt};
   }
 };
 
--- a/cpp/src/arrow/engine/substrait/options.cc
+++ b/cpp/src/arrow/engine/substrait/options.cc
@@ -30,7 +30,7 @@ namespace engine {
 
 namespace {
 
-std::vector<compute::Declaration::Input> MakeDeclarationInputss(
+std::vector<compute::Declaration::Input> MakeDeclarationInputs(
     const std::vector<DeclarationInfo>& inputs) {
   std::vector<compute::Declaration::Input> input_decls(inputs.size());
   for (size_t i = 0; i < inputs.size(); i++) {
@@ -133,7 +133,7 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
     compute::AsofJoinNodeOptions asofjoin_node_opts{std::move(input_keys), tolerance};
 
     // declaration
-    auto input_decls = MakeDeclarationInputss(inputs);
+    auto input_decls = MakeDeclarationInputs(inputs);
     return RelationInfo{
         {compute::Declaration("asofjoin", input_decls, std::move(asofjoin_node_opts)),
          std::move(schema)},
@@ -146,7 +146,7 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
                                        const ExtensionSet& ext_set) {
     if (inputs.size() != 1) {
       return Status::Invalid(
-          "substrait_ext::NamedTapNode requires a single table but got: ", inputs.size());
+          "substrait_ext::NamedTapRel requires a single input but got: ", inputs.size());
     }
 
     ARROW_ASSIGN_OR_RAISE(auto tap_func_name,
@@ -164,7 +164,7 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
     std::shared_ptr<compute::ExecNodeOptions> named_tap_opts =
         std::make_shared<NamedTapNodeOptions>(named_tap_rel.name(),
                                               std::move(renamed_schema));
-    auto input_decls = MakeDeclarationInputss(inputs);
+    auto input_decls = MakeDeclarationInputs(inputs);
     return RelationInfo{
         {compute::Declaration(tap_func_name, input_decls, std::move(named_tap_opts)),
          std::move(renamed_schema)},
--- a/cpp/src/arrow/engine/substrait/options.cc
+++ b/cpp/src/arrow/engine/substrait/options.cc
@@ -19,6 +19,7 @@
 
 #include <google/protobuf/util/json_util.h>
 #include <mutex>
+
 #include "arrow/compute/exec/asof_join_node.h"
 #include "arrow/compute/exec/options.h"
 #include "arrow/engine/substrait/expression_internal.h"
@@ -174,27 +175,27 @@ std::shared_ptr<ExtensionProvider> default_extension_provider() {
   return ExtensionProvider::kDefaultExtensionProvider;
 }
 
-NamedTapProvider kDefaultNamedTapProvider =
+namespace {
+
+NamedTapProvider g_default_named_tap_provider =
     [](const std::string& tap_kind, std::vector<compute::Declaration::Input> inputs,
        const std::string& tap_name,
        std::shared_ptr<Schema> tap_schema) -> Result<compute::Declaration> {
   return Status::NotImplemented("Named tap provider must be given");
 };
 
-namespace {
-
 std::mutex g_default_named_tap_provider_mutex;
 
-}
+}  // namespace
 
 NamedTapProvider default_named_tap_provider() {
-  std::unique_lock lock(g_default_named_tap_provider_mutex);
-  return kDefaultNamedTapProvider;
+  std::unique_lock<std::mutex> lock(g_default_named_tap_provider_mutex);
+  return g_default_named_tap_provider;
 }
 
 void set_default_named_tap_provider(NamedTapProvider provider) {
-  std::unique_lock lock(g_default_named_tap_provider_mutex);
-  kDefaultNamedTapProvider = provider;
+  std::unique_lock<std::mutex> lock(g_default_named_tap_provider_mutex);
+  g_default_named_tap_provider = provider;
 }
 
 }  // namespace engine
--- a/cpp/src/arrow/engine/substrait/options.cc
+++ b/cpp/src/arrow/engine/substrait/options.cc
@@ -181,7 +181,7 @@ NamedTapProvider g_default_named_tap_provider =
     [](const std::string& tap_kind, std::vector<compute::Declaration::Input> inputs,
        const std::string& tap_name,
        std::shared_ptr<Schema> tap_schema) -> Result<compute::Declaration> {
-  return Status::NotImplemented("Named tap provider must be given");
+  return Status::NotImplemented("Plan contained a NamedTapRel but no provider configured");
 };
 
 std::mutex g_default_named_tap_provider_mutex;
--- a/cpp/src/arrow/engine/substrait/options.cc
+++ b/cpp/src/arrow/engine/substrait/options.cc
@@ -66,7 +66,7 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
     if (rel.Is<substrait_ext::AsOfJoinRel>()) {
       substrait_ext::AsOfJoinRel as_of_join_rel;
       rel.UnpackTo(&as_of_join_rel);
-      return MakeAsOfJoinRel(conv_opts, inputs, as_of_join_rel, ext_set);
+      return MakeAsOfJoinRel(inputs, as_of_join_rel, ext_set);
     }
     if (rel.Is<substrait_ext::NamedTapRel>()) {
       substrait_ext::NamedTapRel named_tap_rel;
@@ -78,8 +78,7 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
   }
 
  private:
-  Result<RelationInfo> MakeAsOfJoinRel(const ConversionOptions& conv_opts,
-                                       const std::vector<DeclarationInfo>& inputs,
+  Result<RelationInfo> MakeAsOfJoinRel(const std::vector<DeclarationInfo>& inputs,
                                        const substrait_ext::AsOfJoinRel& as_of_join_rel,
                                        const ExtensionSet& ext_set) {
     if (inputs.size() < 2) {
--- a/cpp/src/arrow/engine/substrait/options.cc
+++ b/cpp/src/arrow/engine/substrait/options.cc
@@ -181,7 +181,8 @@ NamedTapProvider g_default_named_tap_provider =
     [](const std::string& tap_kind, std::vector<compute::Declaration::Input> inputs,
        const std::string& tap_name,
        std::shared_ptr<Schema> tap_schema) -> Result<compute::Declaration> {
-  return Status::NotImplemented("Plan contained a NamedTapRel but no provider configured");
+  return Status::NotImplemented(
+      "Plan contained a NamedTapRel but no provider configured");
 };
 
 std::mutex g_default_named_tap_provider_mutex;
--- a/cpp/src/arrow/engine/substrait/options.cc
+++ b/cpp/src/arrow/engine/substrait/options.cc
@@ -18,6 +18,7 @@
 #include "arrow/engine/substrait/options.h"
 
 #include <google/protobuf/util/json_util.h>
+#include <mutex>
 #include "arrow/compute/exec/asof_join_node.h"
 #include "arrow/compute/exec/options.h"
 #include "arrow/engine/substrait/expression_internal.h"
@@ -157,13 +158,11 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
     std::vector<std::string> columns(named_tap_rel.columns().begin(),
                                      named_tap_rel.columns().end());
     ARROW_ASSIGN_OR_RAISE(auto renamed_schema, schema->WithNames(columns));
-    std::shared_ptr<compute::ExecNodeOptions> named_tap_opts =
-        std::make_shared<NamedTapNodeOptions>(named_tap_rel.name(),
-                                              std::move(renamed_schema));
     auto input_decls = MakeDeclarationInputs(inputs);
-    ARROW_ASSIGN_OR_RAISE(auto decl,
-                          conv_opts.named_tap_mapper(named_tap_rel.kind(), input_decls,
-                                                     std::move(named_tap_opts)));
+    ARROW_ASSIGN_OR_RAISE(
+        auto decl,
+        conv_opts.named_tap_provider(named_tap_rel.kind(), input_decls,
+                                     named_tap_rel.name(), std::move(renamed_schema)));
     return RelationInfo{{std::move(decl), std::move(renamed_schema)}, std::nullopt};
   }
 };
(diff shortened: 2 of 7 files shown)


cpp/src/arrow/engine/substrait/options.h
```diff
@@ -83,6 +84,14 @@ class ARROW_ENGINE_EXPORT ExtensionProvider {
 
 ARROW_ENGINE_EXPORT std::shared_ptr<ExtensionProvider> default_extension_provider();
 
+struct ARROW_ENGINE_EXPORT NamedTapNodeOptions : public compute::ExecNodeOptions {
+  NamedTapNodeOptions(const std::string& name, std::shared_ptr<Schema> schema)
+      : name(name), schema(std::move(schema)) {}
+
+  std::string name;
+  std::shared_ptr<Schema> schema;
+};
```
westonpace, line 93: I'm not sure what the plan is here.  I thought originally the goal was to mimic the named table.  In that case I'd expect to see something like `NamedTableProvider` (e.g. `NamedTapProvider`) that translates names into actual nodes (e.g. a `TeeNode`).

In that case I wouldn't expect there to be any "named tap" equivalent in Acero.  I'm not sure what `NamedTapNode` and `AddPassFactory` are providing in this PR?

westonpace (changes requested):
Took a quick look.  The Substrait side of this makes sense but I'm a little unclear on what the plan is for the C++ side of things.

rtpsw:
@westonpace, in the [recent commit](https://github.com/apache/arrow/pull/33909/commits/65542c3cbbe28228d8105ab49430a9b3844a77d5), when coding up `kDefaultNamedTapKindMapper`, I reused the code structure for `kDefaultNamedTableProvider`. However, it looks like it is not properly initialized - I observed the error message `C++ exception with description "bad_function_call" thrown in the test body` when calling its `operator()`. Is this intended? Or shall I fix the initialization of both?

rtpsw:
Note that the factory registry does not support scoping, which could be useful for cleaning up after and isolating testing (here, due to `AddPassFactory`) and in other use cases, like with other scoped registries. @westonpace, do you agree? open an issue?

westonpace:
> @westonpace, in the [recent commit](https://github.com/apache/arrow/pull/33909/commits/65542c3cbbe28228d8105ab49430a9b3844a77d5), when coding up kDefaultNamedTapKindMapper, I reused the code structure for kDefaultNamedTableProvider. However, it looks like it is not properly initialized - I observed the error message C++ exception with description "bad_function_call" thrown in the test body when calling its operator(). Is this intended? Or shall I fix the initialization of both?

What would you initialize it to?  There is no default named table provider.  I'm pretty sure we are using `kDefaultNamedTapKindMapper` as sentinel value only.  Since `std::function` evaluates to false if default initialized:

```
        if (!conversion_options.named_table_provider) {
          return Status::Invalid(
              "plan contained a named table but a NamedTableProvider has not been "
              "configured");
        }
```

westonpace:
> Note that the factory registry does not support scoping, which could be useful for cleaning up after and isolating testing (here, due to AddPassFactory) and in other use cases, like with other scoped registries. @westonpace, do you agree? open an issue?

I'm not opposed to this but I don't think it is really necessary.  The tests all rely on the fact that the initialization methods are setup as "call-once" so calling them repeatedly won't be a problem.

When dealing with the function registry it seemed important to support nested registries because you might have query-specific functions (e.g. if a UDF is embedded this could be one way to handle it).  However, the odds of having query-specific nodes seems pretty slim to me.

cpp/src/arrow/engine/substrait/options.h
westonpace: If you return exec node options (similar to named tap provider) then you could bypass the need to encode properties into the name (you wouldn't even really need `kind`):

```
# in python
def tap_provider(name):
  if name == 'one':
    return TeeNodeOptions('/tmp/dataset_one')
  elif name == 'two':
    return TeeNodeOptions('/tmp/dataset_two')
  else:
    raise Error(...)
```

or you could move the name encoding / decoding into python

```
# in python
def tap_provider(name)
  path = get_path_from_name(name)
  return TeeNodeOptions(path)
```

or you could still use the kind mapping

```
# in python
def tap_provider(name)
  kind = get_kind_from_name(name)
  if kind == 'tee':
   return TeeNodeOptions(path)...
```

However, this feature is still pretty experimental so I don't mind sticking with kind<->factory name mapping for now if that is what you would prefer.

cpp/proto/substrait/extension_rels.proto
```diff
@@ -44,3 +44,16 @@ message AsOfJoinRel {
...
+// Named tap relation
+//
+// A tap is a relation having a single input relation that it passes through, while also
+// causing some side-effect, e.g., writing to external storage.
+message NamedTapRel {
+  // The kind of tap
+  string kind = 1;
+  // A name used to configure the tap, e.g., a URI defining the destination of writing
+  string name = 2;
+  // Column names for the tap's output. Their number must be 0 or equal to the input's.
```
westonpace, line 57:
```suggestion
  // Column names for the tap's output. If specified there must be one name per field.
  // This can be empty and field names will be generated
```

cpp/src/arrow/engine/substrait/options.cc
```diff
@@ -28,36 +28,58 @@
 namespace arrow {
 namespace engine {
 
+namespace {
+
+std::vector<compute::Declaration::Input> MakeDeclarationInputss(
```
westonpace, line 33:
```suggestion
std::vector<compute::Declaration::Input> MakeDeclarationInputs(
```

cpp/src/arrow/engine/substrait/options.cc
```diff
@@ -28,36 +28,58 @@
...
+      rel.UnpackTo(&named_tap_rel);
+      return MakeNamedTapRel(conv_opts, inputs, named_tap_rel, ext_set);
     }
     return Status::NotImplemented("Unrecognized extension in Susbstrait plan: ",
                                   rel.DebugString());
   }
 
  private:
-  Result<RelationInfo> MakeAsOfJoinRel(const std::vector<DeclarationInfo>& inputs,
+  Result<RelationInfo> MakeAsOfJoinRel(const ConversionOptions& conv_opts,
```
westonpace, line 81: Any particular reason to add `conv_opts` here?

cpp/src/arrow/engine/substrait/options.cc
```diff
@@ -111,15 +133,43 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
     compute::AsofJoinNodeOptions asofjoin_node_opts{std::move(input_keys), tolerance};
 
     // declaration
-    std::vector<compute::Declaration::Input> input_decls(inputs.size());
-    for (size_t i = 0; i < inputs.size(); i++) {
-      input_decls[i] = inputs[i].declaration;
-    }
+    auto input_decls = MakeDeclarationInputss(inputs);
```
westonpace, line 136:
```suggestion
    auto input_decls = MakeDeclarationInputs(inputs);
```

cpp/src/arrow/engine/substrait/options.cc
```diff
@@ -111,15 +133,43 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
...
         std::move(field_output_indices)};
   }
+
+  Result<RelationInfo> MakeNamedTapRel(const ConversionOptions& conv_opts,
+                                       const std::vector<DeclarationInfo>& inputs,
+                                       const substrait_ext::NamedTapRel& named_tap_rel,
+                                       const ExtensionSet& ext_set) {
+    if (inputs.size() != 1) {
+      return Status::Invalid(
+          "substrait_ext::NamedTapNode requires a single table but got: ", inputs.size());
```
westonpace, line 149:
```suggestion
          "substrait_ext::NamedTapRel requires a single input but got: ", inputs.size());
```

cpp/src/arrow/engine/substrait/options.h
```diff
@@ -93,6 +106,11 @@ struct ARROW_ENGINE_EXPORT ConversionOptions {
   /// The default behavior will return an invalid status if the plan has any
   /// named table relations.
   NamedTableProvider named_table_provider = kDefaultNamedTableProvider;
+  /// \brief A custom strategy to be used for mapping a tap kind to a function name
+  ///
+  /// The default behavior will return an invalid status if the plan has any
+  /// named tap relations.
```
westonpace, line 112: Couldn't a default implementation pass the `kind` through directly?  E.g. `pass_for_named_tap -> pass_for_named_tap`?

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -5355,5 +5380,97 @@ TEST(Substrait, AsOfJoinDefaultEmit) {
...
+      [&input_schema](
+          const std::vector<std::string>& names) -> Result<std::shared_ptr<Table>> {
+        if (names.size() != 1) {
+          return Status::Invalid("Multiple test table names");
+        }
+        if (names[0] == "T") {
+          return TableFromJSON(input_schema, {"[[2, 1, 1.1], [4, 1, 2.1], [6, 2, 3.1]]"});
+        }
+        return Status::Invalid("Unknown test table name ", names[0]);
+      });
```
westonpace, line 5460: Use `AlwaysProvideSameTable` for brevity.

westonpace (changes requested):
Looking at this further I do worry that `kind` will be too restrictive.  For example, if you are writing to a temporary output directory and the `name` specifies the path then how do you configure the directory to write to?

Perhaps that might be what you intended with the nesting of node factories?  E.g. you would register a node factory that had the output directory hard-coded as a part of the node factory?

rtpsw:
> What would you initialize it to?

I would have the default implementation return the invalid status, like the one you pointed to.

rtpsw:
> I'm not opposed to this but I don't think it is really necessary.

The purpose I had in mind for this is test isolation. I agree it's not strictly necessary.

cpp/src/arrow/engine/substrait/options.cc
rtpsw: It gets passed all the way to `MakeNamedTapRel`, which uses `conv_opts.named_tap_mapper`.

cpp/proto/substrait/extension_rels.proto
```diff
@@ -44,3 +44,17 @@ message AsOfJoinRel {
...
+//
+// A tap is a relation having a single input relation that it passes through, while also
+// causing some side-effect, e.g., writing to external storage.
+message NamedTapRel {
+  // The kind of tap
+  string kind = 1;
+  // A name used to configure the tap, e.g., a URI defining the destination of writing
+  string name = 2;
+  // Column names for the tap's output. If specified there must be one name per field.
+  // If empty, field names will be automatically generated.
```
icexelloss, line 58: Why don't we make the columns required - i.e., if it's missing or doesn't match the number of fields in the data then it's an illegal message?

cpp/src/arrow/engine/substrait/options.cc
```diff
@@ -28,36 +28,58 @@
...
+  }
+  return input_decls;
+}
+
+}  // namespace
+
 class BaseExtensionProvider : public ExtensionProvider {
  public:
-  Result<RelationInfo> MakeRel(const std::vector<DeclarationInfo>& inputs,
+  Result<RelationInfo> MakeRel(const ConversionOptions& conv_opts,
```
icexelloss, line 46: Why do we need to add the ConversionOptions here?

cpp/src/arrow/engine/substrait/options.h
icexelloss: In my mind we should do the following:

(1) Add "NamedTapRel" as a purely abstract relation in Acero. The Acero code itself doesn't care or know how to turn a NamedTapRel message into a Declaration.

(2) Add a C++ extension API that allows downstream users / implementer of the NamedTapRel to tell Acero "hey, if you see a NamedTapRel message, call into the custom substrait extension function that I provide here and I will return a declaration to u" (basically this: https://github.com/apache/arrow/issues/33850)

This way I believe we don't need most the stuff here (including changes to pass conv_opts, named_tap_mapper, etc)

icexelloss: @westonpace @rtpsw WDYT?

cpp/src/arrow/type.h
```diff
@@ -1968,6 +1968,8 @@ class ARROW_EXPORT Schema : public detail::Fingerprintable,
   Result<std::shared_ptr<Schema>> SetField(int i,
                                            const std::shared_ptr<Field>& field) const;
 
+  Result<std::shared_ptr<Schema>> WithNames(const std::vector<std::string>& names) const;
```
icexelloss, line 1971: What does this function do? (Missing docstring)

cpp/src/arrow/engine/substrait/options.h
icexelloss: Maybe the API for (2) can also be "Hey, if you see an certain type of message (is there a way to detect a message is an extension message)?) pass it to the custom extension function that I provide and I will return a declaration to u"

cpp/proto/substrait/extension_rels.proto
icexelloss: I think we should use 
https://github.com/substrait-io/substrait/blob/main/proto/substrait/algebra.proto#L268

To define the named tal rel similar to how we implement Asof join extension with MultiRelExtension

cpp/src/arrow/engine/substrait/options.h
icexelloss: I think with this approach, we can basically support any ExtensionSingleRel, ExtensionMultiRel and ExtensionLeafRel in the future with a single extension point (the custom substrait provider)

rtpsw: Taking it a step further, we could just add `NamedTapRel` without any callback code, because one would be able (with #33850 and without this PR) register a custom extension provider that would handle it in `MakeRel` (and forward what it doesn't handle to the original/default extension provider). One step more, we could add nothing and just let the extension interpret an `Any` message. This feels like stepping in the wrong direction, which aggressively prioritizes generality. I feel we should have some structure in Acero for this, at least so that structures would get reused instead of reinvented.

cpp/src/arrow/engine/substrait/options.cc
rtpsw: See [this discussion](https://github.com/apache/arrow/pull/33909#discussion_r1093869541).

cpp/src/arrow/type.h
rtpsw: It gives a schema with the names changed, similar to other `With*` methods. I could add a docstring.

cpp/src/arrow/engine/substrait/options.h
icexelloss: IMO I don't think this is over generalized - it matches the level of generalization in substrait extension protobuf - Leaf, single rel and multiple rel. As something experimental at this stage I think we do benefit from generalization since we make very minimal changes to Acero code itself and more free to experiment / iterator faster on our side.

cpp/src/arrow/engine/substrait/options.h
```diff
@@ -67,6 +68,10 @@ using NamedTableProvider =
     std::function<Result<compute::Declaration>(const std::vector<std::string>&)>;
 static NamedTableProvider kDefaultNamedTableProvider;
 
+using NamedTapKindMapper = std::function<Result<std::string>(const std::string&)>;
+static NamedTapKindMapper kDefaultNamedTapKindMapper =
```
icexelloss, line 72: If you wanted to add this, I think we should probably match the signature of kDefaultNamedTableProvider and returns a Declaration instead?

rtpsw: A string-typed argument is not sufficient for creating a declaration for an internal (non-source) node - the input declarations are required, and probably options too. Another reason for the differing signatures is that here the string-typed parameter is for the tap kind, which is semantically different than a table name.

cpp/src/arrow/engine/substrait/options.h
rtpsw: There's also the option of having both - a user would use the specific structures when they fit and the more general one when they don't. If we go with this approach, this PR will be for a specific structure. I'd also like to hear @westonpace's opinion.

icexelloss: IFWIW,  agree a specific structure can be useful however in this case I don't feel it is much more useful than a general one in this case. For example, the specific structure for the NamedTapNode make function returns a NamedTapNode option, and doesn't not allow the user to return a declaration with a custom node option (e.g, in our case, a WriteSmoothNodeOption), so we need to handle that in our code after calling this NamedTapNode make function here (or totally by pass NamedTapNode function here). This an be done but I feel that is unnecessary indirection.

cpp/src/arrow/engine/substrait/options.cc
westonpace: But `MakeAsOfJoinRel` does not call `MakeNamedTapRel`.

rtpsw: Ah, sorry. Right, it's not needed here in `MakeAsOfJoinRel`; it's only needed in `MakeNamedTapRel`.

cpp/src/arrow/engine/substrait/options.h
westonpace: I agree with @rtpsw that going too generic (protobuf.Any in and Declaration out) is not a good idea.  This is the exact API for the `ExtensionProvider` and so we already have this.  You could use this for `WriteSmoothNodeOption` and it would end up looking just like the extension that is in place for AsofJoinNode.  If that works for you then there is nothing we need to do :)

So if we want to go less generic then we should look at what Acero provides at each step to simplify things:

### Generic Node

The next step down in generality would be to remove protobuf from the picture using Substrait/Arrow literals:

```
message GenericOptions {
  message Property {
    string key = 0;
    Literal value = 1;
  }
  repeated Property properties = 1;
}

message GenericNode {
  string kind = 0;
  GenericOptions options = 1;
}
```

The API in Arrow would be:

```
void RegisterGenericExtensionHandler(string kind, function<Declaration(unordered_map<string, Scalar>)> handler)
```

In other words, given properties as a map of arrow scalars, create a declaration

**Pros:**
  * No need for extension author to write proto file
  * Can be used for any extension type
**Cons:**
  * Need to create an entire ExecNode for your custom behavior

### Generic Mapper

An even more focused approach could be to remove the need to create an exec node at all.  This would be focused on the 1 input in / 1 input out / no accumulation case (e.g. what MapNode handles).  From what I know of your write need this would fit this case.

The Substrait would be pretty much identical to the above.

```
message MapNode {
  string kind = 0;
  GenericOptions options = 1;
}
```

However, the Arrow API would be:

```
using Mapper = function<ExecBatch(ExecBatch)>;
void RegisterGenericMapHandler(string kind, function<Mapper(unordered_map<string, Scalar>)> handler)
```

**Pros:**
 * No need to write proto file
 * No need to create an exec node
**Cons:**
 * Only handles 1-in relations that don't accumulate

westonpace: Also, sorry, I was starting a bit with a blank slate above.  I think, in this PR, the current idea is that instead of using `GenericOptions` we use `string name` and encode the options into the name.  This is probably less flexible but in many cases users might be more familiar with encoding options into strings (ala URL encoding) instead of dealing with Arrow scalars.  I'm fine with either approach.

westonpace: So sticking with `name` I think the two approaches are:

### Generic Node
```
void RegisterGenericExtensionHandler(string kind, function<Declaration(string)> handler);
```
### Generic Mapper
```
using Mapper = function<ExecBatch(ExecBatch)>;
void RegisterGenericMapHandler(string kind, function<Mapper(string)> handler);
```

cpp/proto/substrait/extension_rels.proto
westonpace: I have no strong opinion on this either.  I do agree there will be cases where the tap doesn't care about the names but forcing the producer to invent field names in this case isn't much burden.

icexelloss:
@rtpsw With https://github.com/apache/arrow/pull/34042 merged I think we can simply this PR? I.e. we don't really need to use the `MakeNamedTapRel` in this PR since it will be implemented in the custom substrait consumer anyway?

rtpsw:
My understanding is that even with the [most generic solution Weston described](https://github.com/westonpace), we would need some message defined since the current design has the extension provider check for the message using `google::protobuf::Any::Is<...>()`. Is there a different place, external to Arrow rather than in its `extension_rels.proto`, where we could define the message?

rtpsw:
@westonpace, are you leaning toward accepting the current version of the PR, or something close to it? @icexelloss would be OK with reducing the PR to just adding the message, in case you'd prefer this.

icexelloss:
IMO we should reduce this PR to be just the extension protobuf. The main reason is that most of the other change in this PR will not be used by us anyway because we will use another custom extension provider to handle the NamedTapMessage so it will not be hitting any of the code changes in this PR. Therefore introducing a bunch of code that will not be used doesn't seem necessarily.

westonpace:
If we changed `NamedTapKindMapper` to return a Declaration I would be happy to proceed with something like this.

icexelloss:
@westonpace would u be ok with having just the proto change in 
https://github.com/apache/arrow/pull/33909/files#diff-878a08387dc84ac471ccb8f129a6e032154fc82d457f8cc59bfcbe012920339b 

I think since the we can register custom substrait extension consumer, we won't really need to call into the implementation for "NamedTapRel" in the default substrait consumer. 

If there is a use case can justifies support NamedTapRel in the default consumer I think we can add that later (but currently, our use case doesn't need it). WDTY?

icexelloss:
One thing I am thinking is for testing maybe we can do the following:

* construct a plan that contains a NamedTapRel
* register a custom substrait consumer that can handle NamedTapRel (returns some declaration, but not necessarily need to execute it) 
* Verify the Declaration 

This way the test serves and both an example for using NamedTapRel and also mimicing how we will be using this internally.

icexelloss:
> A mapper that returns a declaration seems simple enough (and this PR is 80% of the way there) and this makes it clear what the intent of the proto message is even if you don't end up using it.

Fair enough. Let do that @rtpsw

cpp/src/arrow/engine/substrait/options.h
icexelloss: Somehow missed this message - I am leaning towards "Generic Node" since it allows more flexibility to define node that is not just "map-like" and also we don't need to write protobuf file. 

I have a question - if we go the "Generic Node" route it looks like we can achieve register custom handlers via either `RegisterGenericExtensionHandler` or `set_default_extension_provider` [https://github.com/apache/arrow/commit/7423f0332cb11eb780f421c07bac71f87bf44a03#diff-8c5c127e8db8e52b40519262ee2b7e0179d4a39e5bcfe8b2165173b2b352d6a9R139](url) so technically we don't need `set_default_extension_provider` if we have `RegisterGenericExtensionHandler`?

icexelloss:
@rtpsw @westonpace 

I realize there might be some confusion so I want to clarify here.

There seems to be two (or three) proposed solutions:
(1) The one that is implemented in the PR now
(2) What Weston wrote 5 days ago here (https://github.com/apache/arrow/pull/33909#discussion_r1096045166) (The Generic Node / Generic Map) (sorry I somehow missed this message and just read that Today)

It's not clear to me if we have agreement to go with one vs the other? Both seem workable to me. (2) does seem a bit more generic / future proof and (1) seems pretty much done. 

Li

westonpace:
If 2 is acceptable then I would ask that we please move to that.  I think the only change is to have the mapper return a declaration instead of a string.  If we feel there is still some ambiguity I'd be happy to make a PR / diff real quick to show what I am thinking.

rtpsw:
The [recent commit](https://github.com/apache/arrow/pull/33909/commits/0b1147bcd64dfebd266a098d9e6c72588e112afd) has `NamedTapKindMapper` returning a `Declaration`. @westonpace, let me know if anything is missing.

icexelloss:
2 is acceptable to me. @westonpace are u asking for "Generic Node" or "Generic Mapper"?

cpp/src/arrow/engine/substrait/options.h
```diff
@@ -67,6 +69,16 @@ using NamedTableProvider =
     std::function<Result<compute::Declaration>(const std::vector<std::string>&)>;
 static NamedTableProvider kDefaultNamedTableProvider;
 
+using NamedTapKindMapper = std::function<Result<compute::Declaration>(
+    const std::string&, std::vector<compute::Declaration::Input>,
+    std::shared_ptr<compute::ExecNodeOptions>)>;
```
westonpace, line 74: This looks great.  I think we can do one more small tweak.
```suggestion
using NamedTapKindMapper = std::function<Result<compute::Declaration>(
    const std::string&, std::vector<compute::Declaration::Input>,
    std::shared_ptr<Schema>, std::string name)>;
```

cpp/src/arrow/engine/substrait/options.h
```diff
@@ -76,13 +88,22 @@ class ARROW_ENGINE_EXPORT ExtensionProvider {
...
-  virtual Result<RelationInfo> MakeRel(const std::vector<DeclarationInfo>& inputs,
+  virtual Result<RelationInfo> MakeRel(const ConversionOptions& conv_opts,
+                                       const std::vector<DeclarationInfo>& inputs,
                                        const ExtensionDetails& ext_details,
                                        const ExtensionSet& ext_set) = 0;
 };
 
 ARROW_ENGINE_EXPORT std::shared_ptr<ExtensionProvider> default_extension_provider();
 
+struct ARROW_ENGINE_EXPORT NamedTapNodeOptions : public compute::ExecNodeOptions {
```
westonpace, line 99: We can pass the schema and name directly instead of using this class.  If we still want to wrap the two of them in some kind of object (to reduce the # of args passed into the mapper?) then we should include `kind`, do not extend `ExecNodeOptions`, and call it something like NamedTap or NamedTapOptions.

In other words, we shouldn't try to form the node options from the protobuf.  That will be the mapper's job.  We just need to pass whatever was in the protobuf along to the mapper and let the mapper figure out what node options are appropriate.

cpp/src/arrow/engine/substrait/options.h
```diff
@@ -93,6 +114,10 @@ struct ARROW_ENGINE_EXPORT ConversionOptions {
   /// The default behavior will return an invalid status if the plan has any
   /// named table relations.
   NamedTableProvider named_table_provider = kDefaultNamedTableProvider;
+  /// \brief A custom strategy to be used for mapping a tap kind to a function name
+  ///
+  /// The default mapper returns a declaration whose factory name is equal to the tap kind
+  NamedTapKindMapper named_tap_mapper = kDefaultNamedTapKindMapper;
```
westonpace, line 120: How do you feel about renaming this to `NamedTapProvider` for consistency?

cpp/src/arrow/engine/substrait/options.h
```diff
@@ -67,6 +69,16 @@ using NamedTableProvider =
     std::function<Result<compute::Declaration>(const std::vector<std::string>&)>;
 static NamedTableProvider kDefaultNamedTableProvider;
 
+using NamedTapKindMapper = std::function<Result<compute::Declaration>(
+    const std::string&, std::vector<compute::Declaration::Input>,
+    std::shared_ptr<compute::ExecNodeOptions>)>;
+static NamedTapKindMapper kDefaultNamedTapKindMapper =
```
westonpace, line 75: Do you think we will want to make the default named tap mapper (or the default named table provider for that matter) configurable?

cpp/src/arrow/engine/substrait/options.h
```diff
@@ -67,6 +69,16 @@ using NamedTableProvider =
...
+    [](const std::string& kind, std::vector<compute::Declaration::Input> inputs,
+       std::shared_ptr<compute::ExecNodeOptions> options)
+    -> Result<compute::Declaration> {
+  return compute::Declaration(kind, inputs, options);
```
westonpace, line 79: If we remove `NamedTapNodeOptions` like some of my other comments suggest then I suppose there is no more meaningful default mapper.  That being said, I don't know of any use case where this default would be correct.  Perhaps we just return an error here like we do with named tables?

westonpace (changes requested):
Sorry, I hadn't noticed that commit.  I agree this looks close.  I have a few more questions / minor suggestions

cpp/src/arrow/engine/substrait/options.h
rtpsw: Done.

cpp/src/arrow/engine/substrait/options.h
rtpsw: Done.

cpp/src/arrow/engine/substrait/options.h
rtpsw: Done.

cpp/src/arrow/engine/substrait/options.h
rtpsw: I made it configurable.

cpp/src/arrow/engine/substrait/options.h
rtpsw: Done.

icexelloss (approved):
LGTM

cpp/src/arrow/engine/substrait/options.cc
```diff
@@ -129,5 +175,28 @@ std::shared_ptr<ExtensionProvider> default_extension_provider() {
   return ExtensionProvider::kDefaultExtensionProvider;
 }
 
+namespace {
+
+NamedTapProvider g_default_named_tap_provider =
+    [](const std::string& tap_kind, std::vector<compute::Declaration::Input> inputs,
+       const std::string& tap_name,
+       std::shared_ptr<Schema> tap_schema) -> Result<compute::Declaration> {
+  return Status::NotImplemented("Named tap provider must be given");
```
westonpace, line 184:
```suggestion
  return Status::NotImplemented("Plan contained a NamedTapRel but no provider configured");
```

westonpace (approved):
I think this works.  Thank you for your persistence.

PR #33910, opened 2023-01-28
MINOR: [C++][Parquet] Add const to DictEncoder member functions

wgtmac:
### Rationale for this change

Some member functions of parquet dictionary encoder should be const.

Two member functions of parquet dictionary encoder are changed to be const.

### Are these changes tested?

Make sure all existing tests pass.

### DIFF
--- a/cpp/src/parquet/encoding.cc
+++ b/cpp/src/parquet/encoding.cc
@@ -486,7 +486,7 @@ class DictEncoderImpl : public EncoderImpl, virtual public DictEncoder<DType> {
 
   ~DictEncoderImpl() override { DCHECK(buffered_indices_.empty()); }
 
-  int dict_encoded_size() override { return dict_encoded_size_; }
+  int dict_encoded_size() const override { return dict_encoded_size_; }
 
   int WriteIndices(uint8_t* buffer, int buffer_len) override {
     // Write bit width in first byte
@@ -601,7 +601,7 @@ class DictEncoderImpl : public EncoderImpl, virtual public DictEncoder<DType> {
 
   /// Writes out the encoded dictionary to buffer. buffer must be preallocated to
   /// dict_encoded_size() bytes.
-  void WriteDict(uint8_t* buffer) override;
+  void WriteDict(uint8_t* buffer) const override;
 
   /// The number of entries in the dictionary.
   int num_entries() const override { return memo_table_.size(); }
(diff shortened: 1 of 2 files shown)


rok (approved):

wgtmac:
Thanks @rok!

Any chance to take a look? @wjones127

cpp/src/parquet/encoding.cc
```diff
@@ -649,15 +649,15 @@ class DictEncoderImpl : public EncoderImpl, virtual public DictEncoder<DType> {
 };
 
 template <typename DType>
-void DictEncoderImpl<DType>::WriteDict(uint8_t* buffer) {
+void DictEncoderImpl<DType>::WriteDict(uint8_t* buffer) const {
```
wjones127, line 652: I'm not 100% sure on the C++ conventions here, but I'm not sure I agree this should be labelled `const`. We are modifying `memo_table_`, right? And probably doing so in a way that wouldn't be safe to call from multiple threads concurrently?

@lidavidm Am I wrong on this?

lidavidm: Weirdly enough, MemoTable::CopyValues is also marked const. But I agree this shouldn't be const semantically, even if it technically can be.

wgtmac: > I'm not 100% sure on the C++ conventions here, but I'm not sure I agree this should be labelled `const`. We are modifying `memo_table_`, right? And probably doing so in a way that wouldn't be safe to call from multiple threads concurrently?
> 
> @lidavidm Am I wrong on this?

We are not modifying memo_table_ here. Instead this function copies all values from memo_table_ to the output buffer. It is read-only and doesn’t change anything.

I’m not sure when it will be called concurrently. 

In my context, the dict encoder is called to collect and aggregate dictionary entries of all row groups in a parquet writer. Finally it outputs the aggregated dictionary in a const function. So it makes sense to me to label it as const.

@lidavidm @wjones127

wgtmac: @pitrou WDYT?

pitrou: Indeed, `MemoTable::CopyValues` being `const` is fine AFAICT.

pitrou (approved):
This looks reasonable to me. Thanks @wgtmac !

cpp/src/parquet/encoding.cc
lidavidm: Sorry about that! It's a little unclear that a method called CopyValues modifies its argument and not itself, but I should've looked more closely.

PR #33912, opened 2023-01-29
GH-33911: [C++] Add missing std::forward to Result::ValueOrElse

### LINKED ISSUE #33911: [C++] Broken univeral forwarding in Result::ValueOrElse

https://github.com/apache/arrow/blob/4f1d255f3dc57457e5c78d98c4b76fc523cf9961/cpp/src/arrow/result.h#L369-L375

The parameter `generate_alternative` is passed by univeral forwarding, but `std::forward` is missing to use it.

This may result in incorrect overloading selection or compilation errors, such as a class type with the method `operator()() &&`.

### Component(s)

C++

PragmaTwice:
It fixes #33911.

------

From #33911:

> https://github.com/apache/arrow/blob/4f1d255f3dc57457e5c78d98c4b76fc523cf9961/cpp/src/arrow/result.h#L369-L375
>
> The parameter `generate_alternative` is passed by univeral forwarding, but `std::forward` is missing to use it.
>
> This may result in incorrect overloading selection or compilation errors, such as a class type with the method `operator()() &&`.
* Closes: #33911

### DIFF
--- a/cpp/src/arrow/result.h
+++ b/cpp/src/arrow/result.h
@@ -371,7 +371,7 @@ class [[nodiscard]] Result : public util::EqualityComparable<Result<T>> {
     if (ok()) {
       return MoveValueUnsafe();
     }
-    return generate_alternative();
+    return std::forward<G>(generate_alternative)();
   }
 
   /// Apply a function to the internally stored value to produce a new result or propagate
--- a/cpp/src/arrow/result_test.cc
+++ b/cpp/src/arrow/result_test.cc
@@ -28,6 +28,7 @@
 #include "arrow/testing/gtest_compat.h"
 #include "arrow/testing/gtest_util.h"
 #include "arrow/testing/matchers.h"
+#include "arrow/util/functional.h"
 
 namespace arrow {
 
@@ -793,6 +794,15 @@ TEST(ResultTest, MatcherExplanations) {
     EXPECT_THAT(listener.str(),
                 testing::StrEq("whose error \"Type error: XXX\" doesn't match"));
   }
+}
+
+TEST(ResultTest, ValueOrGeneratedMoveOnlyGenerator) {
+  Result<MoveOnlyDataType> result = Status::Invalid("");
+  internal::FnOnce<MoveOnlyDataType()> alternative_generator = [] {
+    return MoveOnlyDataType{kIntElement};
+  };
+  auto out = std::move(result).ValueOrElse(std::move(alternative_generator));
+  EXPECT_EQ(*out.data, kIntElement);
 }
 
 }  // namespace


pitrou:
@bkietz @felipecrv @benibus Does this look ok?

westonpace:
Let's add a unit test (below requires including `arrow/util/functional.h`):

```
TEST(ResultTest, ValueOrGeneratedMoveOnlyGenerator) {
  Result<MoveOnlyDataType> result = Status::Invalid("");
  internal::FnOnce<MoveOnlyDataType()> alternative_generator = [] {
    return MoveOnlyDataType{kIntElement};
  };
  auto out = std::move(result).ValueOrElse(std::move(alternative_generator));
  EXPECT_EQ(*out.data, kIntElement);
}
```

I'll point out that `ValueOrElse` is not used at all within the code base (other than in unit tests) but this concern seems legitimate.

cpp/src/arrow/result.h
```diff
@@ -371,7 +371,7 @@ class [[nodiscard]] Result : public util::EqualityComparable<Result<T>> {
     if (ok()) {
       return MoveValueUnsafe();
     }
-    return generate_alternative();
+    return std::forward<G>(generate_alternative)();
```
felipecrv, line 374: Note that from the comment, `generate_alternative` is supposed to be a factory not an actual `G` or parameters that can be forwarded to `G`'s constructor. This is a way to avoid evaluating the parameter when `Result` already contains something. 

Functional Programming languages make this kind of lazy parameter (call-by-name) very easy to declare. In scalar this would be `alternative: => T` instead of `alternative: T`.

https://www.scala-lang.org/api/current/scala/Option.html#getOrElse[B%3E:A](default:=%3EB):B

felipecrv: @PragmaTwice a way to fix the linked issue would be to ensure somehow that `G` is a lambda.

felipecrv: An `enable_if` using `is_invocable_r` would restrict the calls only to generators `G` that return `T`.

[1] https://en.cppreference.com/w/cpp/types/is_invocable

cpp/src/arrow/result.h
```diff
@@ -371,7 +371,7 @@ class [[nodiscard]] Result : public util::EqualityComparable<Result<T>> {
     if (ok()) {
```
felipecrv, line 371: Additionally, I don't think this should be `G&&` as the intention here is to let the compiler not even materialize `G` (e.g. by allocating an object on the stack and calling some constructor for `G`), but instead just splice the code from the lambda into the inlined code for this function.

This is how predicates are passed to functions like `find_if` in the standard library.

```cpp
template< class InputIt, class UnaryPredicate >
constexpr InputIt find_if( InputIt first, InputIt last, UnaryPredicate p );
```

pitrou: > Additionally, I don't think this should be `G&&` as the intention here is to let the compiler not even materialize `G` (e.g. by allocating an object on the stack and calling some constructor for `G`), but instead just splice the code from the lambda into the inlined code for this function.

I'm not sure what you mean here. `G&&` is a [forwarding reference](https://en.cppreference.com/w/cpp/language/reference#Forwarding_references) according to the C++ reference.

> This is how predicates are passed to functions like `find_if` in the standard library.

Is this the actual declaration in a standard library implementation, or the abstract signature as shown in the reference?
(also, since `find_if` typically calls the predicate several times, it can probably not `std::forward` it)

felipecrv: > > Additionally, I don't think this should be `G&&` as the intention here is to let the compiler not even materialize `G` (e.g. by allocating an object on the stack and calling some constructor for `G`), but instead just splice the code from the lambda into the inlined code for this function.
> 
> I'm not sure what you mean here. `G&&` is a [forwarding reference](https://en.cppreference.com/w/cpp/language/reference#Forwarding_references) according to the C++ reference.

Yes, but why do we want to forward `G` when we want the compiler to completely inline the code it represents?

> 
> > This is how predicates are passed to functions like `find_if` in the standard library.
> 
> Is this the actual declaration in a standard library implementation, or the abstract signature as shown in the reference? 

Yes. This is a common way of passing lambdas that you want to be completely inlined if they are inline at the callsite.

See https://github.com/llvm/llvm-project/blob/main/libcxx/include/__algorithm/find_if.h#L23

> (also, since `find_if` typically calls the predicate several times, it can probably not `std::forward` it)

Again, I don't understand why we should forward it. We only call it, don't store/pass it anywhere.

Another example from the STL is how `std::sort` inlines `Compare __c` everywhere. If you pass a lambda it will inline the lambda, if you pass a function pointer or an `std::function` it will invoke it by following the pointer and so on.

pitrou: This is becoming pedantic. Do we want to solve #33911 or not?

pitrou: (and allow for https://github.com/apache/arrow/pull/33912#issuecomment-1409091955)

felipecrv: Sorry. In my other comment I gave a suggestion on how to solve the issue.

Proof that it works as intended: https://godbolt.org/z/5c17zjG89

felipecrv: My full proposal is (1) Remove the `&&` from `G` and (2) add the `enable_if` requirement.

westonpace: > Proof that it works as intended: https://godbolt.org/z/5c17zjG89

I'm a little confused.  In this example I get:

```
function &:
copy
copy

function && (with std::move):
move
move
```

If I change to `G&&` I get:

```

function &:

function && (with std::move):
```

It seems the version without the moves (and especially without the copies) would be preferred.

Also, if your generator is stateful (e.g. a generator object that returns the next id in a sequence) then it would make a big difference if you use & or &&.

I agree, in this case, it seems rather unlikely that someone is going to be passing in a stateful (or complex to copy, can't be inlined) generator.   Still, I don't think there is any penalty to the &&.  And the standard [does sometimes pass functions by forwarding reference](https://en.cppreference.com/w/cpp/utility/functional/invoke).

westonpace: [An even more applicable example](https://en.cppreference.com/w/cpp/utility/optional/or_else)

felipecrv: > It seems the version without the moves (and especially without the copies) would be preferred.

You're right @westonpace. I stand corrected.

The implementation from `optional` you linked does exactly as this PR plus a `static_assert` with a nice error message that we could adopt.

felipecrv (approved):

cpp/src/arrow/result.h
PragmaTwice: Thank for your reviews and suggestions!

But I want to point out that there are some difference between `Result::ValueOrElse` and `std::optional::or_else`: `Result<T>::ValueOrElse` returns a value typed `T`, while `std::optional<T>::or_else` typed `std::optional<T>`, so that the generator in `std::optional::or_else` can returns something like `std::nullopt` and then the `or_else` operations can be "chained" (e.g. `v.and_then().or_else().transform()`) and more expressive to achieve "monadic".

PragmaTwice:
> Let's add a unit test (below requires including `arrow/util/functional.h`):
> 
> ```
> TEST(ResultTest, ValueOrGeneratedMoveOnlyGenerator) {
>   Result<MoveOnlyDataType> result = Status::Invalid("");
>   internal::FnOnce<MoveOnlyDataType()> alternative_generator = [] {
>     return MoveOnlyDataType{kIntElement};
>   };
>   auto out = std::move(result).ValueOrElse(std::move(alternative_generator));
>   EXPECT_EQ(*out.data, kIntElement);
> }
> ```
> 
> I'll point out that `ValueOrElse` is not used at all within the code base (other than in unit tests) but this concern seems legitimate.

So thanks! Added.

PragmaTwice:
I notice that a CI job [(R / AMD64 Windows C++ RTools 40 ucrt64)](https://github.com/apache/arrow/actions/runs/4050545764/jobs/6968031026) failed, but I cannot find the specific compilation error after reading the CI log. :rofl:

Could anyone help me? So thanks!

westonpace:
The log is [here](https://github.com/apache/arrow/actions/runs/4050545764/jobs/6968031026) though the error appears to be a failure to build abseil so I'm pretty sure it's unrelated to your PR.

PragmaTwice:
> The log is [here](https://github.com/apache/arrow/actions/runs/4050545764/jobs/6968031026) though the error appears to be a failure to build abseil so I'm pretty sure it's unrelated to your PR.

Got it. Thanks.

pitrou:
Thanks for the contribution @PragmaTwice !

PR #33917, opened 2023-01-29
GH-33892: [R] Map `dplyr::n()` to `count_all` kernel

### LINKED ISSUE #33892: [R] Map `dplyr::n()` to `count_all` kernel

Currently the dplyr translation code in the arrow R package translates `dplyr::n()` to `sum(1)`. https://github.com/apache/arrow/pull/15083 adds a `count_all` C++ kernel. We should map `dplyr::n()` to that.

### Component(s)

R

### LINKED ISSUE #33960: [R] Output schema for aggregation is sometimes innacurate

Discovered in https://github.com/duckdb/duckdb/issues/5895 and ended up in 32-bit integers (our calculated output schema) interpreted as 64-bit integers (the actual resulting schema).

In 11.0.0 you can now get a RecordBatchReader from an Arrow dplyr query without starting to evaluate the query, so `as_record_batch_reader(query)$schema` is safe. We might be able to replace our current schema inference with that except we'll need to be careful about performance because that process involves building up the entire query from scratch including any nested components. I don't know how often we request that schema but maybe we can request it less as part of this PR.

An example where our calculated output is incorrect:

``` r

small <- tibble(a = c(letters, letters)) |> as_arrow_table()

query <- small |> count(a)
query  
#> Table (query)
#> a: string
#> n: int32
#> 
#> See $.data for the source Arrow object

collect(query, as_data_frame = FALSE)
#> Table
#> 26 rows x 2 columns
#> $a <string>
#> $n <int64>
as_record_batch_reader(query)$schema
#> Schema
#> a: string
#> n: int64
```

<sup>Created on 2023-01-31 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### Component(s)

R

ianmcook:
### Rationale for this change

This PR is a follow-up to #15083. It allows the R package to register bindings to nullary aggregation functions, and it remaps `dplyr::n()` to the nullary aggregation function `count_all`.

This PR also:
- Prepares the R bindings to support aggregation functions with 2+ arguments, although none yet exist in the C++ library
- Removes the heuristics that were used to infer the data types of aggregates, replacing that with actual type determination

### Are these changes tested?

Yes, through existing tests.

No.
* Closes: #33892
* Closes: #33960

### DIFF
--- a/r/R/dplyr-collect.R
+++ b/r/R/dplyr-collect.R
@@ -188,7 +188,7 @@ implicit_schema <- function(.data) {
     group_fields <- new_fields[.data$group_by_vars]
     hash <- length(.data$group_by_vars) > 0
     agg_fields <- new_fields[setdiff(names(new_fields), .data$group_by_vars)]
-    agg_fields <- fix_aggregated_types(.data$aggregations, agg_fields, hash)
+    agg_fields <- fix_aggregated_types(agg_fields, .data$aggregations, hash)
     new_fields <- c(group_fields, agg_fields)
   }
   schema(!!!new_fields)
--- a/r/R/dplyr-collect.R
+++ b/r/R/dplyr-collect.R
@@ -179,16 +179,18 @@ implicit_schema <- function(.data) {
       new_fields <- c(left_fields, right_fields)
     }
   } else {
+    hash <- length(.data$group_by_vars) > 0
     # The output schema is based on the aggregations and any group_by vars
-    new_fields <- map(summarize_fields(.data), ~ .$type(old_schm))
+    new_fields <- c(
+      aggregate_types(.data, hash, old_schm),
+      group_types(.data, old_schm)
+    )
     # * Put group_by_vars first (this can't be done by summarize,
     #   they have to be last per the aggregate node signature,
     #   and they get projected to this order after aggregation)
     # * Determine the output types of the aggregations
     group_fields <- new_fields[.data$group_by_vars]
-    hash <- length(.data$group_by_vars) > 0
     agg_fields <- new_fields[setdiff(names(new_fields), .data$group_by_vars)]
-    agg_fields <- fix_aggregated_types(agg_fields, .data$aggregations, hash)
     new_fields <- c(group_fields, agg_fields)
   }
   schema(!!!new_fields)
--- a/r/R/dplyr-collect.R
+++ b/r/R/dplyr-collect.R
@@ -180,17 +180,13 @@ implicit_schema <- function(.data) {
     }
   } else {
     # The output schema is based on the aggregations and any group_by vars
-    new_fields <- map(summarize_projection(.data), ~ .$type(old_schm))
+    new_fields <- map(summarize_fields(.data), ~ .$type(old_schm))
     # * Put group_by_vars first (this can't be done by summarize,
     #   they have to be last per the aggregate node signature,
     #   and they get projected to this order after aggregation)
-    # * Infer the output types from the aggregations
+    # * Determine the output types of the aggregations
     group_fields <- new_fields[.data$group_by_vars]
-    hash <- length(.data$group_by_vars) > 0
-    agg_fields <- imap(
-      new_fields[setdiff(names(new_fields), .data$group_by_vars)],
-      ~ agg_fun_output_type(.data$aggregations[[.y]][["fun"]], .x, hash)
-    )
+    agg_fields <- new_fields[setdiff(names(new_fields), .data$group_by_vars)]
     new_fields <- c(group_fields, agg_fields)
   }
   schema(!!!new_fields)
--- a/r/R/dplyr-collect.R
+++ b/r/R/dplyr-collect.R
@@ -184,7 +184,10 @@ implicit_schema <- function(.data) {
     # The group_by vars come first (this can't be done by summarize; they have
     # to be last per the aggregate node signature, and they get projected to
     # this order after aggregation)
-    new_fields <- c(group_types(.data), aggregate_types(.data, hash))
+    new_fields <- c(
+      group_types(.data, old_schm),
+      aggregate_types(.data, hash, old_schm)
+    )
   }
   schema(!!!new_fields)
 }
--- a/r/R/dplyr-collect.R
+++ b/r/R/dplyr-collect.R
@@ -180,18 +180,14 @@ implicit_schema <- function(.data) {
     }
   } else {
     hash <- length(.data$group_by_vars) > 0
-    # The output schema is based on the aggregations and any group_by vars
+    # The output schema is based on the aggregations and any group_by vars.
+    # The group_by vars come first (this can't be done by summarize; they have
+    # to be last per the aggregate node signature, and they get projected to
+    # this order after aggregation)
     new_fields <- c(
-      aggregate_types(.data, hash, old_schm),
-      group_types(.data, old_schm)
+      group_types(.data, old_schm),
+      aggregate_types(.data, hash, old_schm)
     )
-    # * Put group_by_vars first (this can't be done by summarize,
-    #   they have to be last per the aggregate node signature,
-    #   and they get projected to this order after aggregation)
-    # * Determine the output types of the aggregations
-    group_fields <- new_fields[.data$group_by_vars]
-    agg_fields <- new_fields[setdiff(names(new_fields), .data$group_by_vars)]
-    new_fields <- c(group_fields, agg_fields)
   }
   schema(!!!new_fields)
 }
--- a/r/R/dplyr-collect.R
+++ b/r/R/dplyr-collect.R
@@ -186,7 +186,9 @@ implicit_schema <- function(.data) {
     #   and they get projected to this order after aggregation)
     # * Determine the output types of the aggregations
     group_fields <- new_fields[.data$group_by_vars]
+    hash <- length(.data$group_by_vars) > 0
     agg_fields <- new_fields[setdiff(names(new_fields), .data$group_by_vars)]
+    agg_fields <- fix_aggregated_types(.data$aggregations, agg_fields, hash)
     new_fields <- c(group_fields, agg_fields)
   }
   schema(!!!new_fields)
--- a/r/R/dplyr-collect.R
+++ b/r/R/dplyr-collect.R
@@ -184,10 +184,7 @@ implicit_schema <- function(.data) {
     # The group_by vars come first (this can't be done by summarize; they have
     # to be last per the aggregate node signature, and they get projected to
     # this order after aggregation)
-    new_fields <- c(
-      group_types(.data, old_schm),
-      aggregate_types(.data, hash, old_schm)
-    )
+    new_fields <- c(group_types(.data), aggregate_types(.data, hash))
   }
   schema(!!!new_fields)
 }
--- a/r/R/dplyr-summarize.R
+++ b/r/R/dplyr-summarize.R
@@ -310,10 +310,19 @@ summarize_projection <- function(.data) {
   c(
     unlist(unname(imap(
       .data$aggregations,
-      ~setNames(
-        .x$data,
-        paste(.y[length(.x$data) > 0], seq_along(.x$data), sep = "..")
-      )
+      ~if (length(.x$data) > 1) {
+          setNames(
+            .x$data,
+            paste(.y, seq_along(.x$data), sep = "..")
+          )
+        } else if (length(.x$data) > 0) {
+          setNames(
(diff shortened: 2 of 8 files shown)


paleolimbot (commented):
I think this needs review from @nealrichardson...I'm skeptical of futher complicating the aggregation process (which is already rather complicated). I'll run the ursabot benchmarks, but a local example where this makes a meaningful difference would be helpful for me with respect to justifying additional complexity.

ianmcook:
I'm seeing an ~8% speed improvement when I run this locally on my Intel Mac:
```r
library(arrow)
library(dplyr)
library(tictoc)

z <- arrow_table(z = seq_len(4e9))

tic()
z %>% summarize(n()) %>% collect()
toc()
```
I am having trouble getting the package to build with Dataset so that's about the largest data I can test on at the moment.

nealrichardson:
I'll try to take a look this week.

r/R/dplyr-summarize.R
```diff
@@ -322,15 +301,76 @@ arrow_eval_or_stop <- function(expr, mask) {
...
+# library, the `tdigest` and `hash_tdigest` kernels have different output types.
+# The `tdigest` kernel returns `Float64`, but the `hash_tdigest` kernel returns
+# `FixedSizeList[Float64]`. The system that the R bindings use to infer the
+# output types of expressions does not account for this. It infers the output
+# type of `hash_tdigest` as `Float64`. This function is used to correct this.
+fix_aggregated_types <- function(fields, aggs, hash) {
+  imap(
+    fields,
+    ~if (hash && aggs[[.y]]$fun == "tdigest") {
+        fixed_size_list_of(float64(), 1L)
```
nealrichardson, line 356: Is it always float64()? (looks like it is today, but probably not safe to assume that things like timestamp wouldn't return timestamp)

Is it always length 1? (looks like not, the length depends on the number of quantiles you give it in the options)

I think this would generalize better:

```suggestion
        fixed_size_list_of(.x, length(aggs[[.y]][["options"]][["q"]])
```

r/R/dplyr-summarize.R
```diff
@@ -322,15 +301,76 @@ arrow_eval_or_stop <- function(expr, mask) {
...
+# that will be returned by an aggregation
+aggregated_fields <- function(aggs) {
+  map(
+    aggs,
+    ~Expression$create(.$fun, args = .$data, options = .$options)
+  )
+}
+
+# Unlike with other pairs of non-hash/hash aggregate kernels in the Arrow C++
+# library, the `tdigest` and `hash_tdigest` kernels have different output types.
```
nealrichardson, line 347: Perhaps a safer way to do the type determination is to build expressions with the hash versions. The prose documentation says the hash kernels are unary, but according to the error messages you get, as well as the inline function docs in hash_aggregate.cc, they're binary. After some trial and error, I got this to work:

```
> x <- Expression$create("tdigest", Expression$field_ref("a"), options=list(q=c(.1, .2)))
> x$type(schema(a = int32()))
Float64
double

> x <- Expression$create("hash_tdigest", Expression$field_ref("a"), Expression$scalar(1L)$cast(uint32()), options=list(q=c(.1, .2)))
> x$type(schema(a = int32()))
FixedSizeListType
fixed_size_list<item: double>[2]
```

So the second argument is the group hash, and that is `uint32()`.

r/R/dplyr-summarize.R
```diff
@@ -322,15 +301,76 @@ arrow_eval_or_stop <- function(expr, mask) {
...
+    character(0)
+  }
+}
+
+# This function returns a list of expressions representing the aggregated fields
+# that will be returned by an aggregation
+aggregated_fields <- function(aggs) {
+  map(
+    aggs,
+    ~Expression$create(.$fun, args = .$data, options = .$options)
```
nealrichardson, line 342: See the comment below. Basically, `if (hash)` then it's `paste0("hash_", .$fun)` and `c(.$data, Expression$scalar(Scalar$create(1L, type = uint32())))`

r/R/dplyr-summarize.R
```diff
@@ -322,15 +301,76 @@ arrow_eval_or_stop <- function(expr, mask) {
...
+  } else if (length(data) > 0) {
+    name
+  } else {
+    character(0)
+  }
+}
+
+# This function returns a list of expressions representing the aggregated fields
+# that will be returned by an aggregation
+aggregated_fields <- function(aggs) {
```
nealrichardson, line 339: Technically these aren't "Fields". Perhaps `aggregation_exprs` instead?

r/R/dplyr-summarize.R
nealrichardson: Nevermind, see above, I think we can construct `hash_` Expressions and get the actual output type without any munging.

r/R/query-engine.R
```diff
@@ -101,7 +101,11 @@ ExecPlan <- R6Class("ExecPlan",
         # plus group_by_vars (last)
         # TODO: validate that none of names(aggregations) are the same as names(group_by_vars)
         # dplyr does not error on this but the result it gives isn't great
-        node <- node$Project(summarize_projection(.data))
+        projection <- summarize_projection(.data)
+        # skip projection if no grouping and all aggregate functions are nullary
```
nealrichardson, line 105: Why?

r/R/dplyr-summarize.R
```diff
@@ -322,15 +301,76 @@ arrow_eval_or_stop <- function(expr, mask) {
...
+        aggregate_target_names(.x$data, .y)
+      )
+    ))),
+    .data$selected_columns[.data$group_by_vars]
+  )
+}
+
+# This function determines what names to give to the fields used in aggregations
+# (the "targets"). When an aggregate function takes 2 or more fields as targets,
+# this function gives the fields unique names by appending `..1`, `..2`, etc.
```
nealrichardson, line 324: Why? I'm not sure about this. Since there's no code that actually does this in this PR, maybe save it for #33959, where you can prove out what works/is necessary?

r/R/dplyr-summarize.R
```diff
@@ -322,15 +301,76 @@ arrow_eval_or_stop <- function(expr, mask) {
   out
 }
 
+# This function returns a list of expressions that can be used to project the
```
nealrichardson, line 304:
```suggestion
# This function returns a list of expressions that is used to project the
```

nealrichardson (commented):
Thanks for taking this on. Here are some suggestions. I think in the end this PR will result in simplifying the code a bit, it should be more understandable what's going on and will remove some hacks/imprecision.

r/R/query-engine.R
ianmcook: In general we want to avoid adding unnecessary projections to the ExecPlan. This is a case where we can do that.

r/R/dplyr-summarize.R
ianmcook: I plan to do a PR for #33959 soon. I would prefer to leave this code in and change it as needed in that PR. In the meantime, it is unreachable code, which I know isn't great. But taking it out leaves things in an awkward state where obvious cases are unhandled in the code.

ianmcook:
@nealrichardson thanks for your review! I think I addressed or responded to all the issues you raised. I also changed some of the helper functions so that they return types instead of Expressions. They are only ever used to assemble the schema, so it is clearer this way.

r/R/dplyr-summarize.R
ianmcook: Very cool! 

I changed it to work like this, but I used `Scalar$create(1L, uint32())` instead of `Expression$scalar(1L)$cast(uint32())` as the groups argument.

r/R/query-engine.R
nealrichardson: Would something be harmed if it were projected to having 0 columns? Would that be even more efficient?

ianmcook: My understanding is that adding a project node before an aggregate node will never improve performance, and that the only real reason to do it is to ensure that the aggregate node has all the columns it needs.

I just checked with Weston about this. Here's our conversation:

**Ian:**
Which of these two ExecPlans would execute the most quickly and efficiently?

1. Read source data containing columns A,B,C,D --> project to only columns A and B --> aggregate sum of A over groups in B --> output to table

2. Read source data containing columns A,B,C,D --> aggregate sum of A over groups in B --> output to table

In other words, does adding an unnecessary projection help, hurt, or not make any difference?

**Weston**:
I assume by "project to" you mean an actual project node and not a push down projection? In reality you'd want to avoid reading C and D from disk so you'd want the projection in the read.

**Ian**:
correct, actual project node

**Weston**:
An actual project node, with just field references, is effectively free.

Now, if the follow-up node was a filter node, then you might be paying more for plan 2.  We don't use selection vectors so the filter node would have to materialize the data and it would have to materialize your two junk columns.

The aggregate node, on the other hand, doesn't output any input columns.  So I'd expect it will throw them away in it's InputReceived and the performance of 1 & 2 will be more or less the same.

r/R/dplyr-summarize.R
```diff
@@ -322,15 +301,71 @@ arrow_eval_or_stop <- function(expr, mask) {
...
+  } else if (length(data) > 0) {
+    name
+  } else {
+    character(0)
+  }
+}
+
+# This function returns a named list of the data types of the aggregate columns
+# returned by an aggregation
+aggregate_types <- function(.data, hash, schema = NULL) {
```
nealrichardson, line 339: Is schema needed as an argument here? The Expressions in $data will all bring schema with them.

r/R/dplyr-summarize.R
```diff
@@ -322,15 +301,71 @@ arrow_eval_or_stop <- function(expr, mask) {
...
+  map(
+    .data$aggregations,
+    ~if (hash) {
+      Expression$create(
+        paste0("hash_", .$fun),
+        # hash aggregate kernels must be passed another argument representing
+        # the groups, so we pass in a dummy scalar, since the groups will not
+        # affect the type that an aggregation returns
+        args = c(.$data, Scalar$create(1L, uint32())),
```
nealrichardson, line 348: Minor optimization: we could define `.hash_arg <- Scalar$create(1L, uint32())` outside the function so we don't have to keep doing it. There's small (<1ms) overhead in creating R6 objects, and this creates 2 every time you call it. When I was doing some benchmarking of the query building a while back, these all added up such that our conbench TPC-H query benchmarks were being driven more by them than evaluating the query itself (since they're not using really big data).

nealrichardson (commented):
A couple of small suggestions but otherwise LGTM, thanks for taking this on!

r/R/dplyr-summarize.R
ianmcook: The code in `dplyr-collect.R` passes `old_schm` to this function. That's how it worked before and I didn't want to change in for fear of breaking it.

nealrichardson: That could predate the schema tracking in Expressions. I'd try removing it, and if the tests pass, I guess it's not needed? After all, that's what the tests are there to tell us.

r/R/dplyr-summarize.R
ianmcook: Defining it outside of a function doesn't work because of build order. So I did the next best thing and pulled the definition outside of the `map()` so it at least it only incurs the overhead once per function call.

r/R/dplyr-collect.R
```diff
@@ -179,18 +179,18 @@ implicit_schema <- function(.data) {
       new_fields <- c(left_fields, right_fields)
     }
   } else {
+    hash <- length(.data$group_by_vars) > 0
     # The output schema is based on the aggregations and any group_by vars
-    new_fields <- map(summarize_projection(.data), ~ .$type(old_schm))
+    new_fields <- c(
+      aggregate_types(.data, hash, old_schm),
+      group_types(.data, old_schm)
+    )
```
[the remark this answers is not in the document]
ianmcook: Indeed. Thanks for noticing this.

r/R/dplyr-summarize.R
ianmcook: I removed it and it looks like the tests pass.

ianmcook: Oops, I take that back. It looks like there is an edge case with datasets where the old schema does need to be passed, and this did cause a test to fail: https://github.com/apache/arrow/actions/runs/4147404413/jobs/7174247433#step:5:24083

ianmcook: I reverted the removal and restored the schema argument here.

nealrichardson (approved):

PR #33918, opened 2023-01-29
GH-33904: [R] improve behavior of s3_bucket

### LINKED ISSUE #33904: [R] Using s3_bucket with non-AWS S3-compatible storage is confusing

If I want to use arrow with non-AWS, S3-compatible storage, I might find the `s3_bucket` convenience function from our documentation [[1](https://arrow.apache.org/docs/dev/r/articles/arrow.html#connecting-to-cloud-storage), [2](https://arrow.apache.org/docs/dev/r/articles/fs.html)].  The documentation doesn't always mention non-AWS storage but I might see in the help page for `s3_bucket` that I can pass options to `S3FileSystem$Create` using dots and those options include `endpoint_override`. I might then try the following:

```r
s3_bucket(
    "my-bucket",
    endpoint_override = "example.org", 
    anonymous = TRUE
)
```

However, this results in the following error:

```
# Error: IOError: Bucket 'my-bucket' not found
# /Users/bryce/src/apache/arrow/cpp/src/arrow/filesystem/s3fs.cc:829  ResolveRegionUncached(bucket)
# /Users/bryce/src/apache/arrow/cpp/src/arrow/filesystem/s3fs.cc:357  ResolveS3BucketRegion(bucket)
# /Users/bryce/src/apache/arrow/cpp/src/arrow/filesystem/filesystem.cc:719  S3Options::FromUri(uri, out_path)
```

You can see that we're trying to resolve the bucket's AWS region which fails as expected. Interestingly, the user that reported this issue to me found they could prevent this error by creating a bucket on AWS with the same name at which point the region resolution would succeed (pointlessly), the `S3FileSystem` createsd would respect their `endpoint_override`, and future calls like $ls() would target the correct endpoint. This could actually cause some significant frustration for non-AWS users.

If I go back to the docs and read further down on https://arrow.apache.org/docs/dev/r/articles/fs.html#file-systems-that-emulate-s3, I find that the recommended way to do what I want is directly through `S3FileSystem$create`:

```r
S3FileSystem$create(
    endpoint_override = "example.org",
    anonymous = TRUE
)
```

I was made aware through other channels while researching this issue that this syntax would also have worked (though it's not documented sufficiently):

```r
s3_bucket("s3://anonymous@my-bucket?endpoint_override=example.org")
```

With the increasing number of non-AWS providers of S3-compatible storage (e.g., [Ceph](https://docs.ceph.com/en/latest/radosgw/s3/), [DigitalOcean](https://www.digitalocean.com/products/spaces)], I only think we'll have more users of non-AWS, S3-compatible storage and those users will run into confusion over (1) how things work and (2) how things are documented.

I'd like to:

1. Do a pass over the documentation (online, and R help pages)
  a.  To make the messaging clear that we support AWS, GCS, and S3-compatible storage providers
  b. Add more examples to `s3_bucket` to make it clearer how a user can go about doing that
3. Change the behavior of `s3_bucket` to special-case passing `endpoint_override` to skip region resolution and just jump directly to creating an `S3FileSystem`.

### Component(s)

R

cboettig:
`s3_bucket` has needless conditional logic that defaulted to filesystem construction using `FileSystem$from_uri`, and only falling back on `S3FileSystem$create()` when additional optional arguments were supplied.  

This approach is more verbose opaque, and not properly documented, leading to errors in numerous edge cases described below.  All of these problems are easily avoided by using the `S3FileSystem$create()` constructor directly, (which also corresponds to the documentation for the S3 Filesytem in pyarrow).

As you know, the S3 SDK is capable of reading user keys, secret keys, region, endpoint, and scheme from environmental variables and from `.aws` config files, and users who have encountered the S3 system in other open source software are very familiar with these uses.  

The current implementation of `s3_bucket` needlessly seeks to construct the `s3://` URI notation instead,  `s3://{access_key}:{secret_key}@?scheme={scheme}&endpoint_override={uri_escaped_enpoint}` and create the filesystem object using `FileSystem$from_uri`:

https://github.com/apache/arrow/blob/4dd5cedb21d7b58d837bdb3c0d35a5cd80fd9f4b/r/R/filesystem.R#L463

This is not the correct URI construction, as it ignores additional options such as key, secret key, scheme, and endpoint override (and possibly others) that the user has passed to the s3_bucket() function.  This creates errors and unpredictable behavior to the user -- for instance observe that:

```r
s3 <- s3_bucket("neon4cast-scores", endpoint_override="data.ecoforecast.org", anonymous=TRUE)
```

Works as anticipated -- but only because the exact same bucket-name requested from the MINIO-based https://data.ecoforecast.org/ , happens to also exist on Amazon's default endpoint, s3.amazonaws.com (allowing the `from_uri()` method to erroneously succeed, connecting to the wrong endpoint, and then be replaced by a connection to the correct endpoint triggered by the presence of `...` .  (It took me some time to track down this behavior, as I and many colleagues have used `s3_bucket()` with independent endpoints for over a year, not realizing why some bucket names would mysteriously fail while others worked).  

I do not think this problem is isolated to the use of `endpoint_override`, I don't think that `from_uri` construction is preferable to `S3FileSystem$create()` even when a user is working against the default s3.awsamazon.com endpoints, and at any event, requires more care that the URI is constructed with respect to user-facing options (including respecting `AWS_REGION` as documented in the SDK. I believe this would address https://github.com/apache/arrow/issues/33904, but emphasize that _no special conditional handling_ is needed here to support endpoint_override or other options -- the AWS SDK is already well designed to handle alternative endpoints and all of these other options just fine, we just let the standard S3FileSystem$create method do it's job.

I think the change I propose here would make the code base robust to these edge cases, improve performance by decreasing redundant API calls, and requires no change to any of the documentation or function API.

* Closes: #33904

### DIFF
--- a/r/R/filesystem.R
+++ b/r/R/filesystem.R
@@ -456,22 +456,8 @@ default_s3_options <- list(
 #' @export
 s3_bucket <- function(bucket, ...) {
   assert_that(is.string(bucket))
-  args <- list2(...)
-
-  # Use FileSystemFromUri to detect the bucket's region
-  if (!is_url(bucket)) {
-    bucket <- paste0("s3://", bucket)
-  }
-  fs_and_path <- FileSystem$from_uri(bucket)
-  fs <- fs_and_path$fs
-  # If there are no additional S3Options, we can use that filesystem
-  # Otherwise, take the region that was detected and make a new fs with the args
-  if (length(args)) {
-    args$region <- fs$region
-    fs <- exec(S3FileSystem$create, !!!args)
-  }
-  # Return a subtree pointing at that bucket path
-  SubTreeFileSystem$create(fs_and_path$path, fs)
+  fs <- S3FileSystem$create(...)
+  SubTreeFileSystem$create(bucket, fs)
 }
 
 #' Connect to a Google Cloud Storage (GCS) bucket


cboettig:
Okay, I see in the test suite the test now fails line: https://github.com/apache/arrow/blob/master/r/tests/testthat/test-filesystem.R#L184

This occurs because 

```r
fs <- FileSystem$from_uri("s3://ursa-labs-r-test")
```

extracts infers the region from the bucket* while `S3FileSystem$create()` obeys the region assignment configured (i.e. past as an argument, specified in the environmental variable, or configured in a `.aws` configuration file, then falling back on the default `us-east-1`)

If desired, I could add two lines to auto-detect this region, which is returned by the header, i.e.

```
 curl -I https://s3.amazonaws.com/ursa-labs-r-test/
 ```
 
Note that while AWS S3 API provides get-bucket-location, this requires permission is not required to get bucket location from the header. AWS's own docs say to use the HEAD method to detect region on v4 API anyway: https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadBucket.html

 
It's not totally clear that such guessing of the region, rather than obeying the same conventions as S3FileSystem, is really the right choice though.  It may be worth a closer comparison to other interfaces to see how they handle this.  I believe there are valid use cases, e.g. where a bucket is mirrored across many regions, and a user wants to detect the region *in which the code is executing*, rather than guess the region from the bucket info (which iiuc, would result in a machine in US-west region, say, using data from the default us-east-1 region, rather than realizing that the bucket was mirrored in the local region).  

I'm not an expert on AWS and handling of regions, but if the region was always intended to be referred only from the bucket name then AWS would have no reason to expose the configuration option in the first place.

**Comment by amoeba** [comment]
Hey @cboettig, I'm not sure what the best course of action is here but thank you for the PR and discussion. I think we have to keep breaking changes in mind, which makes this more challenging.

As far as how other S3 libraries handle this, there seems to be two broad categories: (1) SDKs and (2) higher-level applications. The SDKs tend not to offer region guessing and instead will error out if region isn't provided one way or another whereas higher level applications like s3fs will look up region for you. I consider the arrow R package to be a higher level application rather than an SDK so I think it makes sense that we have the machinery we do (both in the R package and in Arrow C++) for region detection.

Related, I'm not sure we'd want to add any region-detection code into the R package because we already offer a more advanced facility for this in Arrow C++.

I've tried a few alternative ways to fix the very valid issues you bring up while also not breaking the API and I'm not there yet. I might ask @thisisnic and @paleolimbot to have a look too if they can spare the time and I will return to this when I have time.

**Comment by cboettig** [comment]
Thanks for the discussion!  @amoeba I definitely agree with your point about higher-level function and breaking changes.  How is region guessing handled in arrow's other client interfaces (e.g. python?)  Is region resolution in the C++ level available with something other than FileSystem$from_uri?

It seems to me that natural way forward here would be to access the region from a HEAD request, which seems to be the recommended mechanism at least on the REST API.  Doing this through the SDK is probably the preferable to a REST call, though I'm fuzzy on those details, maybe [this](https://sdk.amazonaws.com/cpp/api/LATEST/aws-cpp-sdk-s3/html/class_aws_1_1_s3_1_1_s3_request.html))



```r
scheme = "https"
endpoint_override = "s3.amazonaws.com"
bucket = "ursa-labs-r-test"

resp <- httr::HEAD(glue::glue("{scheme}://{endpoint_override}/{bucket}"))
region <- resp$headers$`x-amz-bucket-region`
```

It looks like the SDK is already built to handle those endpoint parameters appropriately (i.e. passed in, or inferred from a config file on env vars), which is obviously the right way to do it, not what I've sketched above.  Can you tell if there is a function in the AWS S3 C++ SDK that does this, just inspecting the bucket header to determine the region?

thisisnic:
> How is region guessing handled in arrow's other client interfaces (e.g. python?) Is region resolution in the C++ level available with something other than FileSystem$from_uri?

pyarrow has `fs.resolve_s3_region()`: https://arrow.apache.org/docs/python/generated/pyarrow.fs.resolve_s3_region.html#pyarrow.fs.resolve_s3_region

> Can you tell if there is a function in the AWS S3 C++ SDK that does this, just inspecting the bucket header to determine the region?

Yes; if we follow it allll the way down, we get to this function here: https://github.com/apache/arrow/blob/a0d0ecee71b99b45f2c1b269c70033d39b982d00/cpp/src/arrow/filesystem/s3fs.cc#L580-L584

I had a look at the code in the AWS S3 C++ SDK and it looks like it's doing exactly what you suggest there.

cboettig:
thanks @thisisnic , nice! 

It's beyond my depth to see precisely how to fix this PR to implement this, but conceptually I think this is spot-on to what I'd love to see in the R bindings too (i.e. method to get region, and then a simple call to get_region before calling `S3FileSystem$create()` and `SubTreeFileSystem$create()` )

thisisnic:
> I like the idea of `arrow::resolve_s3_region()` (and possibly deprecating `s3_bucket()` in favour of `resolve_s3_region()` + `S3FileSystem$create()`?).
> 
> Can we fix `endpoint_override` in the meantime? Even if it's just a `warning()` that the argument is about to be ignored?

I'd like there to be some sort of utility function at some point, so users don't have to interact directly with objects, even if we do consider deprecating `s3_bucket()`, but that's a point we can come back to in another issue/PR perhaps.  I think a warning in the short-term is a great solution here.

cboettig:
I do like `s3_bucket()` as part of the R API, and think we should keep it.  It's a concise wrapper around a common pattern of:
1. guess region (optional)
2. create s3 filesystem
3. create subtree filesystem

When introducing new users to the S3 Filesystem, learning one new function with a nice obvious name that starts with `s3_` is far less intimidating than the three considerably more verbose methods using the less familiar `$create()` constructor (I say this after having tried  showing `s3_bucket()` to basically anyone who has ever used R and will listen to me over the past year :joy: ).  Most other APIs offer similar patterns, as does arrow with the `gs_bucket()` analogue.  

I totally get the desire for a quick fix.  I think the natural thing to do right now is: https://github.com/apache/arrow/pull/34009/files, i.e. since `from_uri` is ignoring all optional args already and thus only gives the intended behavior when optional args are not provided (as in the unit test), then we should only hit it when no optional arguments have been included.  

I *think* that should pass the current tests and preserve the existing behavior in all cases where the existing behavior was desirable (e.g. guessing bucket region when no region is given when using the default endpoint), while only altering things where the existing behavior was flawed (e.g. currently a user of even the default endpoint who manually requests a region different from what the bucket header says will find their requested region ignored).

eitsupi:
I am wondering if this can be closed in favor of already merged #34009.

thisisnic:
@eitsupi Well spotted, closing this now.

PR #33922, opened 2023-01-30
MINOR: [Ruby][Docs] Add a short example of Arrow::Table#join to README

qnm:
### Rationale for this change

The code for `Apache::Table` `join` has recently been merged (#12108, #30086) into the Ruby gem, but the current documentation for `join` suggests referring to Jira, which then links into GitHub, where the commit can be found.

This change instead provides a short example of a join.

I have added an example to the README and removed the link to the now-completed bug report.

### Are these changes tested?

No, but is documentation only.

Yes - this PR is documentation.

### DIFF
--- a/ruby/README.md
+++ b/ruby/README.md
@@ -152,4 +152,19 @@ table.group('name').sum('amount')
 ```
 
 ### Joining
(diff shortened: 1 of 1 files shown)


kou (approved):
+1

Issue #24868, opened 2020-05-06
[C++] Add a Tensor logical value type with varying dimensions, implemented using ExtensionType


asfimport:
Support for tensor in Table, RecordBatch, etc. where each row is a tensor of a different shape (e.g images of different sizes), but of the same underlying type (e.g. int32). Implemented as an ExtensionType, so no need to change the format. 

I don't see needing each row being a tensor with a different number of dimensions, so if the implementation for that falls out easily of the use case with each row in the table having a tensor with the same number of dimensions, great. If it adds a lot of complexity, that case would be postponed.

**Reporter**: [Christian Hudon](https://issues.apache.org/jira/browse/ARROW-8714) / @chrish42
**Watchers**: [Rok Mihevc](https://issues.apache.org/jira/browse/ARROW-8714) / @rok
#### Related issues:
- [[C++] Add a Tensor logical value type with constant shape, implemented using ExtensionType](https://github.com/apache/arrow/issues/15483) (relates to)

<sub>**Note**: *This issue was originally created as [ARROW-8714](https://issues.apache.org/jira/browse/ARROW-8714). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

raulcd:
I was going to cherry-pick this due to it being milestoned for 14.0.0 before, should I pick it @jorisvandenbossche  or should I leave it for 15.0.0?

rok:
I'm +0 for 14.0.0 for the same reasons as Joris. Since it's just a doc change it'd not expect it to cause issues with the release.

Issue #15483, opened 2017-09-26
[C++] Add a Tensor logical value type with constant shape, implemented using ExtensionType


asfimport:
In an Arrow table, we would like to add support for a column that has values cells each containing a tensor value, with all tensors having the same shape/dimensions. These would be stored as a binary value, plus some metadata to store type and shape/strides.

**Reporter**: [Wes McKinney](https://issues.apache.org/jira/browse/ARROW-1614) / @wesm
**Assignee**: [Rok Mihevc](https://issues.apache.org/jira/browse/ARROW-1614) / @rok
**Watchers**: [Rok Mihevc](https://issues.apache.org/jira/browse/ARROW-1614) / @rok
#### Related issues:
- [[Python] Store sequences of arbitrary ndarrays (with same type) in Tensor value type](https://github.com/apache/arrow/issues/22239) (is related to)
- [[C++] Add a Tensor logical value type with varying dimensions, implemented using ExtensionType](https://github.com/apache/arrow/issues/24868) (is related to)
#### PRs and other links:
- [GitHub Pull Request #8510](https://github.com/apache/arrow/pull/8510)

<sub>**Note**: *This issue was originally created as [ARROW-1614](https://issues.apache.org/jira/browse/ARROW-1614). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

PR #10565, opened 2021-06-21
ARROW-638: [C++] Complex Number Support via ExtensionTypes

sjperkins:
xref https://issues.apache.org/jira/browse/ARROW-638

https://github.com/apache/arrow/pull/10452 contains the PR implemented Complex Numbers as First Class Types

PR #33925, opened 2023-01-30
GH-33923: [Docs] Tensor canonical extension type specification

### LINKED ISSUE #33923: [Docs] Tensor canonical extension type specification

We propose to add a canonical extension type for tensors and this issue is covering the specification for the extension type.

### Component(s)

Documentation

AlenkaF:
### Rationale for this change

There have been quite a lot of discussions connected to the tensor support in Arrow Tables/RecorBatches. This PR is a specification proposal to add tensors as a canonical type extensions and is meant to be sent to the Mailing list for discussion and vote.

Specification for canonical extension type for fixed sized tensors.

**Open question**

Should metadata include the `"dim_names"` key to (optionally) specify dimension names when creating the Arrow FixedShapeTensorArray? 

* Closes: #33923

### DIFF
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -94,4 +94,4 @@ Fixed size tensor
   * boolean indicating the order of elements in memory with key
     "is_row_major".
 
-  For example: `{ "shape": [2, 5], "is_row_major": True }`
+  For example: `{ "shape": [2, 5], "is_row_major": true }`
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -90,6 +90,6 @@ Fixed size tensor
   The metadata must be a valid JSON object including:
 
   * shape of the contained tensors as an array with key “shape”,
-  * string defining the order of elements in memory with key “order”.
+  * boolean indicating the order of elements in memory with key “order”.
 
   For example: `{ "shape": [2, 5], "is_row_major": True }`
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -84,7 +84,7 @@ Fixed shape tensor
 
 * Extension type parameters:
 
-  * **value_type** = Arrow DataType or Field of the tensor elements.
+  * **value_type** = the Arrow data type of individual tensor elements.
   * **shape** = the physical shape of the contained tensors
     as an array.
 
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -83,13 +83,15 @@ Fixed size tensor
 
   * **value_type** = Arrow DataType of the tensor elements
   * **shape** = shape of the contained tensors as a tuple
-  * **is_row_major** = boolean indicating the order of elements in memory
+  * **is_row_major** = boolean indicating the order of elements
+    in memory
 
 * Description of the serialization:
 
   The metadata must be a valid JSON object including:
 
-  * shape of the contained tensors as an array with key “shape”,
-  * boolean indicating the order of elements in memory with key “order”.
+  * shape of the contained tensors as an array with key "shape",
+  * boolean indicating the order of elements in memory with key
+    "is_row_major".
 
   For example: `{ "shape": [2, 5], "is_row_major": True }`
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -77,11 +77,11 @@ Fixed size tensor
 
 * Extension name: `arrow.fixed_size_tensor`.
 
-* The storage type of the extension: ``List``.
+* The storage type of the extension: ``FixedSizeList``.
 
 * Extension type parameters:
 
-  * **value_type** = pyarrow DataType of the tensor elements
+  * **value_type** = Arrow DataType of the tensor elements
   * **shape** = shape of the contained tensors as a tuple
   * **order** = string indicating the order of elements in memory;
     ‘C’ for row major order and ‘F’ for column major order
@@ -90,10 +90,10 @@ Fixed size tensor
 
   The metadata MUST be a valid JSON object including:
 
-  * shape of the contained tensors as a tuple with key “shape”,
+  * shape of the contained tensors as an array with key “shape”,
   * string defining the order of elements in memory with key “order”.
 
-  For example: `{ “shape”: (2, 5), “order”: ‘C’}`
+  For example: `{ "shape": [2, 5], "order": "C" }`
 
   Implementations MAY include implementation-specific metadata by using a
   namespaced key. For example `{"package.name": {"my": "metadata"}}`
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -79,8 +79,7 @@ Fixed shape tensor
 
 * The storage type of the extension: ``FixedSizeList`` where:
 
-  * **value_type** is the data type of individual tensor elements
-    and is an instance of Arrow ``DataType`` or ``Field``.
+  * **value_type** is the data type of individual tensor elements.
   * **list_size** is the product of all the elements in tensor shape.
 
 * Extension type parameters:
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -95,22 +95,22 @@ Fixed shape tensor
     length and equal to the number of dimensions.
 
     ``dim_names`` are used if the dimensions have well-known
-    names and they map to the physical order (row-major).
+    names and they map to the physical layout (row-major).
 
   * **permutation**  = indices of the desired ordering of the
     original dimensions, defined as an array.
 
     The indices contain a permutation of the values [0, 1, .., N-1] where
     N is the number of dimensions. The permutation indicates which
-    dimension of the logical order corresponds to which dimension of the
+    dimension of the logical layout corresponds to which dimension of the
     physical tensor (the i-th dimension of the logical view corresponds
     to the dimension with number permutations[i] of the physical tensor).
 
     **Permutation is only needed in case the logical order of
     the tensor is a permutation of the physical order (row-major).**
 
-    When logical and physical order are equal, the permutation will
-    always equal to the original order of dimensions ([0, 1, .., N-1])
+    When logical and physical layout are equal, the permutation will
+    always equal to the original layout of dimensions ([0, 1, .., N-1])
     and therefore doesn't need to be defined.
 
 * Description of the serialization:
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -109,9 +109,10 @@ Fixed shape tensor
     **Permutation is only needed in case the logical order of
     the tensor is a permutation of the physical order (row-major).**
 
-    When logical and physical layout are equal, the permutation will
-    always equal to the original layout of dimensions ([0, 1, .., N-1])
-    and therefore doesn't need to be defined.
+    When logical and physical layout are equal, the permutation will always
+    be ([0, 1, .., N-1]) and is therefore absent. Same holds the other way
+    round: if permutation parameter is absent, it is assumed that logical
+    layout matches the physical one.
 
 * Description of the serialization:
 
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -110,9 +110,7 @@ Fixed shape tensor
     the tensor is a permutation of the physical order (row-major).
 
     When logical and physical layout are equal, the permutation will always
-    be ([0, 1, .., N-1]) and can therefore be left out. Same holds the other way
-    round: if permutation parameter is absent, it is assumed that logical
-    layout matches the physical one.
+    be ([0, 1, .., N-1]) and can therefore be left out.
 
 * Description of the serialization:
 
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -79,8 +79,8 @@ Fixed shape tensor
 
 * The storage type of the extension: ``FixedSizeList`` where:
 
-  * **value_type** is the data type of individual tensors and
-    is an instance of ``pyarrow.DataType`` or ``pyarrow.Field``.
+  * **value_type** is the data type of individual tensor elements
+    and is an instance of Arrow ``DataType`` or ``Field``.
   * **list_size** is the product of all the elements in tensor shape.
 
 * Extension type parameters:
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -88,12 +88,9 @@ Fixed size tensor
 
 * Description of the serialization:
 
-  The metadata MUST be a valid JSON object including:
+  The metadata must be a valid JSON object including:
 
   * shape of the contained tensors as an array with key “shape”,
   * string defining the order of elements in memory with key “order”.
 
   For example: `{ "shape": [2, 5], "order": "C" }`
-
-  Implementations MAY include implementation-specific metadata by using a
-  namespaced key. For example `{"package.name": {"my": "metadata"}}`
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -94,7 +94,7 @@ Fixed shape tensor
     as an array. The length of it should be equal to the shape
     length and equal to the number of dimensions.
 
-    ``dim_names`` are used if the dimensions have well-known
+    ``dim_names`` can be used if the dimensions have well-known
     names and they map to the physical layout (row-major).
 
   * **permutation**  = indices of the desired ordering of the
@@ -104,13 +104,13 @@ Fixed shape tensor
     N is the number of dimensions. The permutation indicates which
     dimension of the logical layout corresponds to which dimension of the
     physical tensor (the i-th dimension of the logical view corresponds
-    to the dimension with number permutations[i] of the physical tensor).
+    to the dimension with number ``permutations[i]`` of the physical tensor).
 
-    **Permutation is only needed in case the logical order of
-    the tensor is a permutation of the physical order (row-major).**
+    Permutation can be useful in case the logical order of
+    the tensor is a permutation of the physical order (row-major).
 
     When logical and physical layout are equal, the permutation will always
-    be ([0, 1, .., N-1]) and is therefore absent. Same holds the other way
+    be ([0, 1, .., N-1]) and can therefore be left out. Same holds the other way
     round: if permutation parameter is absent, it is assumed that logical
     layout matches the physical one.
 
@@ -132,5 +132,5 @@ Fixed shape tensor
 
 .. note::
 
-  Elements in an fixed shape tensor extension array are stored
+  Elements in a fixed shape tensor extension array are stored
   in row-major/C-contiguous order.
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -79,9 +79,9 @@ Fixed shape tensor
 
 * The storage type of the extension: ``FixedSizeList`` where:
 
-  * **value_type** is the data type of the child array (individual tensor) and
+  * **value_type** is the data type of individual tensors and
     is an instance of ``pyarrow.DataType`` or ``pyarrow.Field``.
-  * **list_size** is the size of the child array (individual tensor).
+  * **list_size** is the product of all the elements in tensor shape.
 
 * Extension type parameters:
 
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -83,8 +83,7 @@ Fixed size tensor
 
   * **value_type** = Arrow DataType of the tensor elements
   * **shape** = shape of the contained tensors as a tuple
-  * **order** = string indicating the order of elements in memory;
-    ‘C’ for row major order and ‘F’ for column major order
+  * **is_row_major** = boolean indicating the order of elements in memory
 
 * Description of the serialization:
 
@@ -93,4 +92,4 @@ Fixed size tensor
   * shape of the contained tensors as an array with key “shape”,
   * string defining the order of elements in memory with key “order”.
 
-  For example: `{ "shape": [2, 5], "order": "C" }`
+  For example: `{ "shape": [2, 5], "is_row_major": True }`
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -72,4 +72,28 @@ same rules as laid out above, and provide backwards compatibility guarantees.
 Official List
 =============
 
-No canonical extension types have been standardized yet.
+Fixed size tensor
+=================
+
+* Extension name: `arrow.fixed_size_tensor`.
+
+* The storage type of the extension: ``List``.
+
+* Extension type parameters:
+
+  * **value_type** = pyarrow DataType of the tensor elements
+  * **shape** = shape of the contained tensors as a tuple
+  * **order** = string indicating the order of elements in memory;
+    ‘C’ for row major order and ‘F’ for column major order
+
+* Description of the serialization:
+
+  The metadata MUST be a valid JSON object including:
+
+  * shape of the contained tensors as a tuple with key “shape”,
+  * string defining the order of elements in memory with key “order”.
+
+  For example: `{ “shape”: (2, 5), “order”: ‘C’}`
+
+  Implementations MAY include implementation-specific metadata by using a
+  namespaced key. For example `{"package.name": {"my": "metadata"}}`
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -87,15 +87,15 @@ Fixed shape tensor
 
   * **value_type** = Arrow DataType of the tensor elements
   * **shape** = shape of the contained tensors as a tuple
-  * **is_row_major** = boolean indicating the order of elements
-    in memory
 
 * Description of the serialization:
 
-  The metadata must be a valid JSON object including:
+  The metadata must be a valid JSON object including shape of
+  the contained tensors as an array with key "shape".
 
-  * shape of the contained tensors as an array with key "shape",
-  * boolean indicating the order of elements in memory with key
-    "is_row_major".
+  For example: `{ "shape": [2, 5]}`
 
-  For example: `{ "shape": [2, 5], "is_row_major": true }`
+.. note::
+
+  Elements in an fixed shape tensor extension array are stored
+  in row-major/C-contiguous order.
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -105,12 +105,22 @@ Fixed shape tensor
 
     ``{ "shape": [100, 200, 500], "dim_names": ["C", "H", "W"]}``
 
+    The ``dim_names`` metadata can be added if the dimensions have
+    well-known names that can map to the physical order (row-major).
+
   - **"permutation"** holds indexes of the desired ordering of the
     original dimensions. Also defined as an array.
 
     Example of permuted 3-dimensional tensor:
 
     ``{ "shape": [100, 200, 500], "permutation": [2, 0, 1]}``
+
+    **Permutation key in the metadata is needed only in case the logical
+    order of the tensor doesn't equal the physical order (row-major).**
+
+    When logical and physical order are equal, the permutation metadata
+    is not added as it will always equal the original order of dimensions
+    (example ``[0, 1, 2]``).
 
 .. note::
 
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -135,3 +135,13 @@ Fixed shape tensor
 
   Elements in a fixed shape tensor extension array are stored
   in row-major/C-contiguous order.
+
+.. note::
+
+  Other Data Structures in Arrow include a
+  `Tensor (Multi-dimensional Array) <https://arrow.apache.org/docs/format/Other.html>`_
+  to be used as a message in the interprocess communication machinery (IPC).
+
+  This structure has no relationship with the Fixed shape tensor extension type defined
+  by this specification. With defining an extension type one can use fixed shape tensors
+  as elements in a field of a RecordBatch or a Table.
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -91,9 +91,26 @@ Fixed shape tensor
 * Description of the serialization:
 
   The metadata must be a valid JSON object including shape of
-  the contained tensors as an array with key "shape".
+  the contained tensors as an array with key **"shape"**
 
-  For example: `{ "shape": [2, 5]}`
+  - example: ``{ "shape": [2, 5]}``
+
+  and optional:
+
+  - **"dim_names"** holds explicit names to tensor dimensions
+    as an array. The length of it should be equal to the shape
+    length and equal to the number of dimensions.
+
+    Example of dim_names metadata for NCHW ordered data:
+
+    ``{ "shape": [100, 200, 500], "dim_names": ["C", "H", "W"]}``
+
+  - **"permutation"** holds indexes of the desired ordering of the
+    original dimensions. Also defined as an array.
+
+    Example of permuted 3-dimensional tensor:
+
+    ``{ "shape": [100, 200, 500], "permutation": [2, 0, 1]}``
 
 .. note::
 
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -95,17 +95,23 @@ Fixed shape tensor
     length and equal to the number of dimensions.
 
     ``dim_names`` are used if the dimensions have well-known
-    names and map to the physical order (row-major).
+    names and they map to the physical order (row-major).
 
-  * **permutation**  = indexes of the desired ordering of the
-    original dimensions. Also defined as an array.
+  * **permutation**  = indices of the desired ordering of the
+    original dimensions, defined as an array.
 
-    **Permutation key in the metadata is needed only in case the logical
-    order of the tensor doesn't equal the physical order (row-major).**
+    The indices contain a permutation of the values [0, 1, .., N-1] where
+    N is the number of dimensions. The permutation indicates which
+    dimension of the logical order corresponds to which dimension of the
+    physical tensor (the i-th dimension of the logical view corresponds
+    to the dimension with number permutations[i] of the physical tensor).
 
-    When logical and physical order are equal, the permutation metadata
-    is not added as it will always equal the original order of dimensions
-    (example ``[0, 1, 2]``).
+    **Permutation is only needed in case the logical order of
+    the tensor is a permutation of the physical order (row-major).**
+
+    When logical and physical order are equal, the permutation will
+    always equal to the original order of dimensions ([0, 1, .., N-1])
+    and therefore doesn't need to be defined.
 
 * Description of the serialization:
 
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -106,7 +106,7 @@ Fixed shape tensor
     ``{ "shape": [100, 200, 500], "dim_names": ["C", "H", "W"]}``
 
     The ``dim_names`` metadata can be added if the dimensions have
-    well-known names that can map to the physical order (row-major).
+    well-known names and map to the physical order (row-major).
 
   - **"permutation"** holds indexes of the desired ordering of the
     original dimensions. Also defined as an array.
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -86,34 +86,19 @@ Fixed shape tensor
 * Extension type parameters:
 
   * **value_type** = Arrow DataType of the tensor elements
-  * **shape** = shape of the contained tensors as a tuple
+  * **shape** = shape of the contained tensors as an array
 
-* Description of the serialization:
+  Optional parameters:
 
-  The metadata must be a valid JSON object including shape of
-  the contained tensors as an array with key **"shape"**
-
-  - example: ``{ "shape": [2, 5]}``
-
-  and optional:
-
-  - **"dim_names"** holds explicit names to tensor dimensions
+  * **dim_names** = explicit names to tensor dimensions
     as an array. The length of it should be equal to the shape
     length and equal to the number of dimensions.
 
-    Example of dim_names metadata for NCHW ordered data:
+    ``dim_names`` are used if the dimensions have well-known
+    names and map to the physical order (row-major).
 
-    ``{ "shape": [100, 200, 500], "dim_names": ["C", "H", "W"]}``
-
-    The ``dim_names`` metadata can be added if the dimensions have
-    well-known names and map to the physical order (row-major).
-
-  - **"permutation"** holds indexes of the desired ordering of the
+  * **permutation**  = indexes of the desired ordering of the
     original dimensions. Also defined as an array.
-
-    Example of permuted 3-dimensional tensor:
-
-    ``{ "shape": [100, 200, 500], "permutation": [2, 0, 1]}``
 
     **Permutation key in the metadata is needed only in case the logical
     order of the tensor doesn't equal the physical order (row-major).**
@@ -121,6 +106,22 @@ Fixed shape tensor
     When logical and physical order are equal, the permutation metadata
     is not added as it will always equal the original order of dimensions
     (example ``[0, 1, 2]``).
+
+* Description of the serialization:
+
+  The metadata must be a valid JSON object including shape of
+  the contained tensors as an array with key **"shape"** plus optional
+  dimension names with keys **"dim_names"** and ordering of the
+  dimensions with key **"permutation"**.
+
+  - Example: ``{ "shape": [2, 5]}``
+  - Example with ``dim_names`` metadata for NCHW ordered data:
+
+    ``{ "shape": [100, 200, 500], "dim_names": ["C", "H", "W"]}``
+
+  - Example of permuted 3-dimensional tensor:
+
+    ``{ "shape": [100, 200, 500], "permutation": [2, 0, 1]}``
 
 .. note::
 
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -85,10 +85,11 @@ Fixed shape tensor
 
 * Extension type parameters:
 
-  * **value_type** = Arrow DataType of the tensor elements
-  * **shape** = shape of the contained tensors as an array
+  * **value_type** = Arrow DataType or Field of the tensor elements.
+  * **shape** = the physical shape of the contained tensors
+    as an array.
 
-  Optional parameters:
+  Optional parameters describing the logical layout:
 
   * **dim_names** = explicit names to tensor dimensions
     as an array. The length of it should be equal to the shape
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -72,10 +72,10 @@ same rules as laid out above, and provide backwards compatibility guarantees.
 Official List
 =============
 
-Fixed size tensor
-=================
+Fixed shape tensor
+==================
 
-* Extension name: `arrow.fixed_size_tensor`.
+* Extension name: `arrow.fixed_shape_tensor`.
 
 * The storage type of the extension: ``FixedSizeList``.
 
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -129,6 +129,9 @@ Fixed shape tensor
 
     ``{ "shape": [100, 200, 500], "permutation": [2, 0, 1]}``
 
+    This is the physical layout shape and the the shape of the logical
+    layout would in this case be ``[500, 100, 200]``.
+
 .. note::
 
   Elements in a fixed shape tensor extension array are stored
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -77,7 +77,11 @@ Fixed shape tensor
 
 * Extension name: `arrow.fixed_shape_tensor`.
 
-* The storage type of the extension: ``FixedSizeList``.
+* The storage type of the extension: ``FixedSizeList`` where:
+
+  * **value_type** is the data type of the child array (individual tensor) and
+    is an instance of ``pyarrow.DataType`` or ``pyarrow.Field``.
+  * **list_size** is the size of the child array (individual tensor).
 
 * Extension type parameters:
 
--- a/docs/source/format/CanonicalExtensions.rst
+++ b/docs/source/format/CanonicalExtensions.rst
@@ -143,5 +143,5 @@ Fixed shape tensor
   to be used as a message in the interprocess communication machinery (IPC).
 
   This structure has no relationship with the Fixed shape tensor extension type defined
-  by this specification. With defining an extension type one can use fixed shape tensors
+  by this specification. Instead, this extension type lets one use fixed shape tensors
   as elements in a field of a RecordBatch or a Table.


docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,28 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+    ‘C’ for row major order and ‘F’ for column major order
+
+* Description of the serialization:
+
+  The metadata MUST be a valid JSON object including:
+
+  * shape of the contained tensors as a tuple with key “shape”,
+  * string defining the order of elements in memory with key “order”.
+
+  For example: `{ “shape”: (2, 5), “order”: ‘C’}`
```
jorisvandenbossche, line 96:
```suggestion
  For example: `{ "shape":  [2, 5], "order": "C" }`
```

To make it formatted as valid JSON. Also, JSON doesn't know "tuple" but has an "array" (https://www.json.org/json-en.html), so you can update that in the text above.

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,28 @@ same rules as laid out above, and provide backwards compatibility guarantees.
 Official List
 =============
 
-No canonical extension types have been standardized yet.
+Fixed size tensor
+=================
+
+* Extension name: `arrow.fixed_size_tensor`.
+
+* The storage type of the extension: ``List``.
```
rok, line 80: I think we'd want [FixedSizeList](https://github.com/apache/arrow/blob/master/format/Schema.fbs#L99) for fixed size tensor.
```suggestion
* The storage type of the extension: ``FixedSizeList``.
```

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,28 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+
+  The metadata MUST be a valid JSON object including:
+
+  * shape of the contained tensors as a tuple with key “shape”,
+  * string defining the order of elements in memory with key “order”.
+
+  For example: `{ “shape”: (2, 5), “order”: ‘C’}`
+
+  Implementations MAY include implementation-specific metadata by using a
+  namespaced key. For example `{"package.name": {"my": "metadata"}}`
```
jorisvandenbossche, line 99: I forgot to ask earlier, but what is the benefit of explicitly allowing this _inside_ the serialized metadata? Because the extension metadata is part of the general custom key-value metadata of the Field (where the extension metadata is stored under the `"ARROW:extension:metadata"` key, https://arrow.apache.org/docs/dev/format/Columnar.html#extension-types). So users can still store custom implementation-specific metadata in the general Field metadata, if needed (of course, this isn't seen by the "deserialize" step to instantiate an ExtensionType from the serialized schema, so it would work differently)

Are there concrete use cases for allowing this?

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,28 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+Fixed size tensor
+=================
+
+* Extension name: `arrow.fixed_size_tensor`.
+
+* The storage type of the extension: ``List``.
+
+* Extension type parameters:
+
+  * **value_type** = pyarrow DataType of the tensor elements
```
rok, line 84:
```suggestion
  * **value_type** = Arrow DataType of the tensor elements
```

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,28 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+  * **shape** = shape of the contained tensors as a tuple
+  * **order** = string indicating the order of elements in memory;
+    ‘C’ for row major order and ‘F’ for column major order
```
rok, line 87: Perhaps we should store this as `is_row_major` boolean instead?

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,28 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+
+* Description of the serialization:
+
+  The metadata MUST be a valid JSON object including:
+
+  * shape of the contained tensors as a tuple with key “shape”,
```
rok, line 93: I think it's called array in JSON.
```suggestion
  * shape of the contained tensors as an array with key “shape”,
```

docs/source/format/CanonicalExtensions.rst
rok, line 96:
```suggestion
  For example: `{ “shape”: [2, 5], “is_row_major”: true, "metadata": {...}}`
```

rok (changes requested):
Thanks for doing this @AlenkaF !
I suggested some changes.
Also: a question I hit on the C++ implementation last week:
what happens when one wants to work with two `fixed_size_tensor` extensions (let's say they have different shapes) at once? I think this will be a common occurrence and because of the namespace collision two extensions can't be registered at once.

docs/source/format/CanonicalExtensions.rst
AlenkaF: No, there is no concrete use case and I agree with what you have stated. Will remove the MAY part of the metadata spec.

docs/source/format/CanonicalExtensions.rst
AlenkaF: I have also first thought about storing this as `is_row_major` boolean but then decided to use numpy definition. Am OK with both.

AlenkaF:
> Also: a question I hit on the C++ implementation last week:
> what happens when one wants to work with two `fixed_size_tensor` extensions (let's say they have different shapes) at once? I think this will be a common occurrence and because of the namespace collision two extensions can't be registered at once.

When working on pyarrow implementation locally, this wasn't an issue.
From the Python [documentation](https://arrow.apache.org/docs/python/generated/pyarrow.register_extension_type.html):

> Registration needs an extension type instance, but then works for any instance of the same subclass regardless of parametrization of the type.

And you can see this case in the tests I uploaded to gist:
https://gist.github.com/AlenkaF/95fb41f461fb792396bb20dd502b4112#file-02_tensor_extension_tests-py-L77-L96

I guess it is the same in C++?

rok:
Only one extension type with a given name can be *registered* at a time. Multiple can coexist, but register can only keep one extension per name. I suppose users won't really want to register these types (or is that a requirement for using compute?), but will want an easy way to instantiate them.

docs/source/format/CanonicalExtensions.rst
rok: We use currently use [row_major](https://github.com/apache/arrow/blob/4acf7c17689c926841a5a471828970eda2f496c5/cpp/src/arrow/tensor.cc#L263) in Arrow, but don't use `order`. I just slightly `is_prefer row_major`.

jorisvandenbossche:
> Only one extension type with a given name can be _registered_ at a time. Multiple can coexist, but register can only keep one extension per name.

It's only the name that is being registered, along with the methods to serialize / deserialize. The actual metadata (i.e. the only thing that differs between two different extension type instances of the same type) isn't part of the type class that is being registered, and so for a single registered (parametrized) type, you can have many _instance_ alive at the same time with a different parametrization (i.e. different metadata).  
That's the goal of having parametrized types (and so you _don't_ want to have those parameters (metadata) in the name of the extension type. Otherwise you would have to register every possible parametrization).

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,24 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+  * **value_type** = Arrow DataType of the tensor elements
+  * **shape** = shape of the contained tensors as a tuple
+  * **is_row_major** = boolean indicating the order of elements in memory
+
+* Description of the serialization:
+
+  The metadata must be a valid JSON object including:
+
+  * shape of the contained tensors as an array with key “shape”,
+  * string defining the order of elements in memory with key “order”.
```
rok, line 93:
```suggestion
  * boolean indicating the order of elements in memory with key “order”.
```

rok (approved):

AlenkaF:
Thank you for reviewing Rok!

If Joris agrees this can be sent to the ML when the C++ implementation is ready. If I remember correctly, I would start a new discussion thread for the tensor canonical extension specification adding some explanation, link to the C++ implementation and pyarrow example. Please correct me if I am wrong.

I am working on PyArrow extension example to have it ready for illustration.

docs/source/format/CanonicalExtensions.rst
rok: @jorisvandenbossche any thoughts on `is_row_major` vs `order`?

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,26 @@ same rules as laid out above, and provide backwards compatibility guarantees.
 Official List
 =============
 
-No canonical extension types have been standardized yet.
+Fixed size tensor
+=================
+
+* Extension name: `arrow.fixed_size_tensor`.
```
jorisvandenbossche, line 78: One more thought about the naming: in the discussions (eg https://github.com/apache/arrow/issues/15483), we typically spoke about a fixed or constant _shape_ (and a fixed size can still mean varying shapes). 
So I wonder if a more correct name would be "fixed shape tensor" ?

pitrou:
Hey @lhoestq, you (and/or other people at HuggingFace) might be interested in reviewing this proposed addition. It will also be discussed on the Arrow dev ML.

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,26 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+
+* The storage type of the extension: ``FixedSizeList``.
```
pitrou, line 80: Can you describe the value type and the size of the fixed-size list?

docs/source/format/CanonicalExtensions.rst
AlenkaF: That would be better, yes.

rok: As we don't use this anywhere yet and "shape" is more precise I'd prefer it too.
(Varying shape issue is here: https://github.com/apache/arrow/issues/24868)

docs/source/format/CanonicalExtensions.rst
jorisvandenbossche: @AlenkaF I think we should probably describe the size in context of the tensor (i.e. the product of the shape), instead of generic explanation of FixedSizeList type. Also "size of the child array (individual tensor)." isn't fully correct, since the "child array" is generally considered to be the full child array (which would have a size of "(length of TensorArray) x (size of each individual tensor)")

AlenkaF: Oh, thanks for explaining! Will correct 👍

mariosasko:
@jorisvandenbossche 

> I noticed that in the HuggingFace Datasets implementation, there is a [comment](https://github.com/huggingface/datasets/blob/5b793dd8c43bf6e85f165238becb3c64f6cd3ed0/src/datasets/features/features.py#L659-L660) about using variable instead of fixed size list array. That is resolved now, and it's fine to use fixed-size lists?

Yes, the fixed-sized version seems to work now (requires `pyarrow>=10.0.0` to pass the tests).

> I think none of the existing custom tensor extension type implementations have this (eg neither of Ray and HuggingFace Datasets have this. @mariosasko do you know if that has come up in HuggingFace?)

No, that has yet to be requested on our side (so not super important for us right now).

docs/source/format/CanonicalExtensions.rst
AlenkaF: Changed the description in https://github.com/apache/arrow/pull/33925/commits/37e83db3787b4a94a359c577b04e4583f7f60818

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,30 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+
+  * **value_type** is the data type of individual tensors and
+    is an instance of ``pyarrow.DataType`` or ``pyarrow.Field``.
+  * **list_size** is the product of all the elements in tensor shape.
+
+* Extension type parameters:
+
+  * **value_type** = Arrow DataType of the tensor elements
+  * **shape** = shape of the contained tensors as a tuple
+  * **is_row_major** = boolean indicating the order of elements
```
thomasw21, line 90: Can we not store `stride` instead? https://pytorch.org/docs/stable/generated/torch.Tensor.stride.html#torch.Tensor.stride

rok: In the [Zulip discussion](https://ursalabs.zulipchat.com/#narrow/stream/180245-dev/topic/Canonical.20extension.20type.20for.20tensors/near/324882917) we are leaning towards canonical type always storing row-major and letting applications store strides in metadata. Any arguments for or against from you or your users would be most welcome at this point!

thomasw21: So you mean removing `is_row_major` as a parameter? I guess the issue boils down to how fast you want the loading/saving? In torch terminology, making a tensor contiguous (I'm guessing is the same as row major) makes a copy because the underlying memory representation changes. The same would apply to loading.

rok: > So you mean removing is_row_major as a parameter?

Yes. We would just use the physical layout of the source and not change memory layout when going in and out of the extension. We would provide an option to store the layout as metadata.

jorisvandenbossche: > In torch terminology, making a tensor contiguous (I'm guessing is the same as row major) makes a copy because the underlying memory representation changes.

My understanding is that contiguous tensors in torch are indeed always row-major, but so that also means that if you have such a contiguous tensor, you don't need any copy to put this in the proposed extension TensorArray (or you can get it out without a copy).

thomasw21: > Yes. We would just use the physical layout of the source and not change memory layout when going in and out of the extension. We would provide an option to store the layout as metadata.

This makes sense. I do think that in terms of UX it helps to be agnostic of those kind of consideration (and so having the format handle this in an underlying fashion it helps). But it makes sense to have a simple design.

> My understanding is that contiguous tensors in torch are indeed always row-major, but so that also means that if you have such a contiguous tensor, you don't need any copy to put this in the proposed extension TensorArray (or you can get it out without a copy).

Yes for contiguous it's perfect. The proposition of having `stride` as well, would make a zero copy system as well.

thomasw21: Sometimes users will manipulate non_contiguous formats without realising. Typically `torch` uses a `to(memory_format=torch.channels_last)` which essentially is a equivalent to `.permute(0,2,3,1).contiguous()`. This API is often used in computer vision as some operation run faster when the tensors are not actually stored in "row major" fashion. So loading images in that specific format directly would help, which means storing non row major tensors.

```python
import torch

device = torch.device("cpu")
x = torch.randn(2,3,5,7, device=device)
y = x.to(memory_format=torch.channels_last)

# Assert that they don't share memory
do_share_memory = x.data_ptr() == y.data_ptr()
print(f"Channels Last shares memory: {do_share_memory}")

# Assert that channels last are not contiguous
print(f"Channels Last is contiguous: {y.is_contiguous()}")

# Data storage is equivalent to contiguous NHWC
z = x.permute(0,2,3,1).contiguous()
is_storage_equal = all(z_data == y_data for z_data, y_data in zip(z.storage(), y.storage()))
print(f"Channels Last storage is the same as the NHWC format: {is_storage_equal}")
```

More on this: https://pytorch.org/blog/accelerating-pytorch-vision-models-with-channels-last-on-cpu/

pitrou: Note that "contiguous" can be ambiguous. There are two possible meanings:
* physical contiguity: all values span a single contiguous area in memory (there are no "holes" without values in them), regardless of how they are ordered
* logical contiguity: values can be walked in either "C" or "F" order

Numpy uses the latter meaning. Note that logical contiguity implies physical contiguity, but the converse is not true, for example if you swap some dimensions in a more than 3+D array.

For now, it seems this spec and discussion is talking about _physical_ contiguity, as it doesn't care how the values are logically accessed (the interpretation of data is not part of the spec).

pitrou: Thus, by adding some metadata describing the dimension names in physical (row-major) order (for example `["N", "H", "W", "C"]`), we would leave the matter of logical contiguity to consumers of the data.

pitrou: (that said, there are situations where there are no dimension names, only dimension indices, such as when tensors are coming from Numpy arrays, so perhaps a slightly different convention is needed?)

jorisvandenbossche: I think for this spec, the "physical" contiguity is a given, because that's the only possibility that it would match with our memory layout for FixedSizeList (or any other type, for that matter). 

But I do think we _also_ can say that the data should be interpreted as row-major ("logical" contiguity). At least, that's what we would use by default when converting to/from arrays (it's of course open for discussion if we should do that, but to be more explicit: that's what the current (example) implementations do), and applications could still ignore that and interpret the data however they want.

jorisvandenbossche: > Thus, by adding some metadata describing the dimension names in physical (row-major) order (for example `["N", "H", "W", "C"]`), we would leave the matter of logical contiguity to consumers of the data.

But those dimension names only make sense if you _assume_ the data is in logical row-major order, no? Because otherwise if the data could be interpreted logically in different ways, you still don't know which logical dimension name matches with which actual data values?

pitrou: I see two problems with a "is_row_major" flag:
* when it's false, you don't know what the actual layout is (is it column-major, or is it in a non-trivial order like in the transpose() example)?
* saying that it "should be interpreted as row-major" doesn't say what the dimensions actually are. If you are shipping arrays Numpy then it's not a problem, because dimensions are anonymous (you are just doing abstract math on tensors). But if you are shipping ML tensors, the information might be incomplete (IIUC, you can have row-major "NHWC" or row-major "NCHW").

pitrou: > But those dimension names only make sense if you _assume_ the data is in logical row-major order, no?

The logical row-major order is given by the dimension names, so this is a bit circular?

rok: > (that said, there are situations where there are no dimension names, only dimension indices, such as when tensors are coming from Numpy arrays, so perhaps a slightly different convention is needed?)

So `dim_indexes` instead of `dim_names`?

rok: I think order of dimension names given by the application at creation time should define the logical contiguity and logical strides.

rok: Consequently a consumer can derive their own strides for a different logical view by (I think, didn't really check this):
```python
def strides(dim_names):
  dim_indexes = np.argsort(dim_names, physical_dim_names)
  view_shape = permute(shape, dim_indexes)
  return compute_row_major_strides(view_shape)
```

jorisvandenbossche: > I see two problems with a "is_row_major" flag:

To be clear, I am not advocating for this flag. I am advocating for removing it, and state in the spec it is always assumed to be row major.

> saying that it "should be interpreted as row-major" doesn't say what the dimensions actually are

Yes, and for that reason we can have the optional dimension names

> The logical row-major order is given by the dimension names, so this is a bit circular?

I am not fully sure what's the consequence of it being circular, but does that then mean those two indeed go together? I.e. the order of the dimension names match the data if you assume it to be row-major (if that's the case, we still need to specify this, so you know how to match the list of dimension names to the data)

> So `dim_indexes` instead of `dim_names`?

I don't understand how that would be useful. That would just be [0, 1, 2] for 3D tensors? It's the optional name that could give meaning to an application (is it the width of the image? or the color channel?)

rok: > I don't understand how that would be useful. That would just be [0, 1, 2] for 3D tensors? It's the optional name that could give meaning to an application (is it the width of the image? or the color channel?)

I was referring to the numpy case where there are no `dim_names`. I do agree `dim_names` is the more useful option.

lhoestq (commented):
Thanks everyone for diving into this :)

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,30 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+
+  The metadata must be a valid JSON object including shape of
+  the contained tensors as an array with key "shape".
+
+  For example: `{ "shape": [2, 5]}`
+
+.. note::
+
+  Elements in an fixed shape tensor extension array are stored
+  in row-major/C-contiguous order.
```
paleolimbot, line 101: Just a note that this will prevent storing R matrices as zero-copy. I seem to remember an earlier version of this had an option for storing column-major tensors zero-copy?

paleolimbot: I think this is also the case for Julia: https://docs.julialang.org/en/v1/manual/performance-tips/#man-performance-column-major

jorisvandenbossche: Do you have the same problem right now with FixedSizeList arrays? Or what is this converted to in R now? (can you convert that to a single matrix?)

Also, even if we allow a different order or custom strides for each _individual tensor_, the full array backing the FixedSizeListArray (the flat values child array) still needs the first dimension (with size == length of the logical array) with the biggest strides. So if R doesn't support that, I don't think zero-copy conversion is ever possible?

paleolimbot: We'd just have to make a chunked array where every chunk had a single element. We don't currently support converting a fixed-size list as a matrix (if we did it would be faster -- even if it wouldn't be zero copy).

jorisvandenbossche: On further thinking, I am also not sure this is actually a problem. Languages like Julia and R can still create an array/matrix from the raw memory of the ListArray zero-copy. This array will only look transposed to their users if you need to prevent the copy. It's then up to the application/user to decide if you want to keep that or not (have a strided view in expected dimension order, or create a copy)?

paleolimbot: > This array will only look transposed to their users if you need to prevent the copy. 

Users have always been able to transport tensors if they are willing to keep track of that kind of metadata themselves.

One potential use of this is communicating a single view of one very large tensor, for which case copies are expensive and in most cases unnecessary.

pitrou: Well, if you have only one tensor, then both ways would work anyway, right?

AlenkaF:
Hello @changhiskhan (plus anybody from Lance), @clarkzinzow, @scottjlee (plus anybody from Ray), you might also be interested in reviewing this proposed spec for fixed shape tensor type. A discussion e-mail has also been sent to the Arrow dev ML.

jorisvandenbossche:
@paleolimbot thanks for the R example (it's nice to see how easy it is!) 
One small comment on the conversion to/from an R matrix in your example: typically, if you start from an array of a certain dimension, the dimension of the actual shape in the metadata of the TensorType is the original dimension - 1 (since the first dimension becomes the length of the logical TensorArray). In your example, you are converting a single array to a 1-len TensorArray (which is probably related to what you said earlier about having a "chunked array where every chunk had a single element", but just to note that I think this won't be the typical way to convert to/from TensorArray)

jorisvandenbossche:
From inline discussion at https://github.com/apache/arrow/pull/33925#discussion_r1094350467:

> [@thomasw21] Sometimes users will manipulate non_contiguous formats without realising. Typically `torch` uses a `to(memory_format=torch.channels_last)` which essentially is a equivalent to `.permute(0,2,3,1).contiguous()`. This API is often used in computer vision as some operation run faster when the tensors are not actually stored in "row major" fashion. So loading images in that specific format directly would help, which means storing non row major tensors.
> 
> ...
> 
> More on this: https://pytorch.org/blog/accelerating-pytorch-vision-models-with-channels-last-on-cpu/

@thomasw21 the page you link mentions the difference between "physical order" and "logical order". So the actual, physical layout in memory that pytorch uses can either be channels first (NCHW) or channels last (NHWC). But in both cases, this physical order is row-major. When interacting with those tensors as a user, pytorch however always _shows_ the data as if it's NCHW, independent of the physical layout, i.e. what they call the "logical order". And it's when viewing the data as NCHW while the physical layout is NHWC that you get custom strides / non-C-contiguous layout _for that view_. But when considering the physical layout NHWC, the data is still row-major. 

And for the purpose of this specification and being able to transfer FixedShapeTensor data using Arrow, it's the physical order that matters, and optionally the dimension order that matches the physical row-major layout (i.e. NHWC or NCHW for pytorch).  

> So loading images in that specific format directly would help, which means storing non row major tensors.

So given the above, my understanding is that it is still perfectly possible to store either formats that pytorch uses as row-major data in the proposed FixedShapeTensor array, and be able to load it directly (without copy). 
You could then use the proposed `"dim_names"` to keep track of the actual layout (for example if the physical layout is "channels last" (NHWC), you could specify `"dim_names": ["H", "W", "C"]`, and when reading those data into a pytorch tensor, you know the physical order, but are still free to _view_ as a different logical order).

docs/source/format/CanonicalExtensions.rst
thomasw21: > [@jorisvandenbossche ] @thomasw21 the page you link mentions the difference between "physical order" and "logical order". So the actual, physical layout in memory that pytorch uses can either be channels first (NCHW) or channels last (NHWC). But in both cases, this physical order is row-major. When interacting with those tensors as a user, pytorch however always shows the data as if it's NCHW, independent of the physical layout, i.e. what they call the "logical order". And it's when viewing the data as NCHW while the physical layout is NHWC that you get custom strides / non-C-contiguous layout for that view. But when considering the physical layout NHWC, the data is still row-major.

I think I have a hard time understanding this point. To me, torch has the ability to change the underlying "physical" ordering depending on how you interpret data. See Figure 1 of that blogpost
image

thomasw21: I don't know if that helps the discussion:

```python
import torch

def get_1d_memory_buffer(tensor):
    return "".join(hex(elt) for elt in x.storage().untyped().byte())

x = torch.randn(2,3)
y = torch.empty(3,2).transpose(0,1)

# Fill y with x data
y[:] = x

assert x.shape == y.shape
assert get_1d_memory_buffer(x) != get_1d_memory_buffer(y)
# You can try printing `x` and `y` and you'll see that the tensors are the same
```

jorisvandenbossche: > To me, torch has the ability to change the underlying "physical" ordering depending on how you interpret data. 

Yes, but both of those use row-major / C-contiguous memory layout. What you change is the order of the dimensions (C-H-W or H-W-C), and to do that while keeping row-major layout, changing between channels first/last layout requires actually shuffling the data in memory (and thus requires a copy). But either "physical order" is row-major and thus can be stored in the proposed FixedShapeTensor array without copy.

jorisvandenbossche: > I don't know if that helps the discussion:

For your example, `x` and `y` indeed have a different layout of the values in memory, while seemingly representing the same tensor (in you print them). But that's because `y` is transposed after it was created (and this only changed the logical order, not the physical), and thus has custom strides. 

But you could still store both x and y without copy in a FixedShapeTensor array. The difference is that `x` would be stored as shape (2, 3), and y would be stored as (3, 2). And assume that in this dummy example the dimensions are called A and B, and if you want to keep track of the correct logical order, you would store the dimension names for `x` as `["A", "B"]` and for `y` as `["B", "A"]`. Then afterwards, after reading such data and the application knows it always wants the data in (A, B) order, it can transpose the data read from the stored `y` (just as you did when creating `y`)

thomasw21: Btw I fixed a bug in my script ...

> The difference is that x would be stored as shape (2, 3), and y would be stored as (3, 2)

I understand that there exist a permutation of dimensions that allows me to get a "row major" format. I think it doesn't change how you store the permutation information, ie via dimension names or stride. It felt like a natural concept to me to store `stride`, as this would allow just provide a better generalisation IMO. But I do understand if the current extension would focus on pure row_major.

jorisvandenbossche: OK, I understand (we might have been talking past each other a bit, as I was assuming you want to have `strides` to allow zero-copy for all cases, while I tried to convince you that it's not needed). 

It's certainly true that we _could_ store strides, but I am not sure it would be a better generalization of (or a full replacement for) dimension names. 
Consider for example that you have channels-last physical data (NHWC), but you view it as channels-first logically (NCHW). To store the data with the logical dimension order, this would requires a `strides` parameter. But assume you only store the strides in the `FixedShapeTensor` type and not the dimension names, then when consuming that data, you know the strides associated with it, but you still don't know for sure what the dimensions mean (because both NHWC viewed as NCHW, or NCHW viewed as NHWC would give you custom strides). Of course, if you know where the data is coming from and you know that it's from a pytorch context, then you can assume that the logical order is NCHW (that's how pytorch [always shows it](https://pytorch.org/blog/accelerating-pytorch-vision-models-with-channels-last-on-cpu/): "No matter what the physical order is, tensor shape and stride will always be depicted in the order of NCHW"), and that information combined with the strides ensures you know if the physical layout is channels-first or last. 
But that requires application-specific context to know that. While if you store the dimension names that match the order assuming a row-major layout, then you can infer the same information (and how to transpose it to get your desired logical order), but without requiring this application-specific knowledge (assuming different applications would use consistent dimension names, so you can recognize those).

So my current understanding is that dimension names are the more generalizable information. 

In addition, pushing the strides logic (how to translate the given dimension order and your desired dimension order to strides) to the application to deal with, keeps the implementation of the FixedShapeTensorType itself simpler, not requiring every implementation to deal with custom strides.

rok:
@ezyang and @fchollet or their colleagues might have some input here as well.
TLDR: We're discussing design of tensor extension arrays for Arrow. Biggest discussion point so far is whether we should store strides or is dimension order enough.

lhoestq:
Regarding what `dim_names` would imply for `pyarrow` and `numpy`:

If I have a numpy array, convert it to arrow using `pa.array`, and then call `to_numpy()`, ideally I would expect to get the exact same numpy array.

If we use `dim_names` and the array has a physical order that is different that the logical order (but still row-major), does that mean I would have to give names to the dimensions to do [numpy -> arrow] and to do [arrow -> numpy] to end up with the same array back ?

Arrow could also give arbitrary names and save the original logical order for the dimensions.

ezyang:
Some recommendations from me (not officially representing PyTorch team, but just based on my experience.)

* Don't encode dimension names. We tried this with named tensor, but it's too restrictive and the propagation rules are too difficult to right. The correct way to do it is first class dims (https://github.com/facebookresearch/torchdim) but this is a UX problem that shouldn't be in arrow
* You can choose to represent strides or dimension ordering. Dimension ordering is good enough to represent memory ordering, but not good enough to represent a broadcasted input. We are internally experimenting with only recording memory ordering for our mobile runtime, since strides are too expressive and they do not support aliasing in their runtime anyway. It is a lossy conversion (distinct strides can map to the same dimension ordering), but it is not too difficult to define a "best" stride for any given dimension ordering.

rok:
@ezyang am I right to read this as: `dim_names` are good enough for exact in-memory representation but for wouldn't suffice for pytorch's internal representation due to propagation rules etc?

ezyang:
They work perfectly fine if you don't do anything with the names, but yeah, in PyTorch, when you start doing operations on the tensor, it's totally nonobvious how to propagate them.

rok:
@ezyang So to be clear: the proposal as is (with `dim_names`) is good for pytorch?
As I understand it you could zero-copy use it but would have to add your own way of handling strides which would not hurt performance/memory?

ezyang:
Actually, what exactly is the dim_names proposal. I assumed that the ordering of the names tell you how to interpret the physical data layout (outside in). If so, if you loaded one of these tensors into PyTorch, we would just ignore the dim names entirely and just give you a tensor in the physical layout. To reinterpret the meaning of the data is application specific; after all, who decides if you call it N or BATCH or something else? Which wouldn't really be in the purview of a framework to decide

ezyang:
If you wanted PyTorch to *automatically* restride; eg give the same logical view for both row-major and col-major matrix, you need a different rep. A layout permutation is sufficiently general that it would work for us. Strides would work too. A fixed enum of names and their logical positions would work as well (but people typically dislike that.)

rok:
> Actually, what exactly is the dim_names proposal.

The proposed logic for calculating `strides` from `dim_names` [is here](https://github.com/apache/arrow/pull/8510/files#diff-eccd12baabdd9d2f59798d3791dfa0fef6272622c18b614bd54a395d08f05956R46-R67) and tests it should pass would be:
```
auto ext_type_4 = extension::tensor_array(int64(), {3, 4, 7}, {"x", "y", "z"});
ASSERT_EQ(ext_type_4->strides(), (std::vector<int64_t>{224, 56, 8}));
ASSERT_EQ(ext_type_4->strides({"x", "y", "z"}), (std::vector<int64_t>{224, 56, 8}));
ASSERT_EQ(ext_type_4->strides({"y", "z", "x"}), (std::vector<int64_t>{56, 8, 224}));
ASSERT_EQ(ext_type_4->strides({"z", "x", "y"}), (std::vector<int64_t>{8, 224, 56}));
```

> I assumed that the ordering of the names tell you how to interpret the physical data layout (outside in). If so, if you loaded one of these tensors into PyTorch, we would just ignore the dim names entirely and just give you a tensor in the physical layout. To reinterpret the meaning of the data is application specific; after all, who decides if you call it N or BATCH or something else? Which wouldn't really be in the purview of a framework to decide

Right! Our current thinking is to provide `dim_names` in the extension and the `->strides(dim_names))` helper to allow applications derive their own strides if they chose another dimension order. That implies the application has some knowledge of `dim_names`, but I think that's a given?

> If you wanted PyTorch to automatically restride; eg give the same logical view for both row-major and col-major matrix, you need a different rep. A layout permutation is sufficiently general that it would work for us. Strides would work too. A fixed enum of names and their logical positions would work as well (but people typically dislike that.)

Storing a layout permutation is an interesting idea too, but that requires knowledge of the "natural order".

ezyang:
> Storing a layout permutation is an interesting idea too, but that requires knowledge of the "natural order".

The natural order is whatever the physical layout is, with contiguous dimensions on the right.

rok:
> > Storing a layout permutation is an interesting idea too, but that requires knowledge of the "natural order".
> 
> The natural order is whatever the physical layout is, with contiguous dimensions on the right.

Sure, but if application is not images and we just receive a shaped tensor without `dim_names` we can't know what they are?

ezyang:
I mean, one way or another the out of bound communication has to happen. In the cases of names, the out of band communication is "sender and receiver agree on common meaning of particular string names" and works pretty OK for specific domains (e.g., images) where these conventions agree. With the case of a layout permutation, the communication is "sender and receiver agree on the semantics of the positions of logical dimensions (e.g., after you apply the permutation)." This is more general because you don't have to have agreed on names for a particular dimension. The user code can then do whatever it needs (e.g., pass `channels_last=True/False` to an operator on the image) to line up the logical interpretation.

rok:
That makes sense yeah. Though I think passing `dim_names` is equally general to agreeing on natural order + passing `permutation` in terms of passing information. Except that `dim_names` gives both in one variable (if we assume dimension names are universal), while natural order + `permutation` requires two.
If I understand correctly it's more about usability/readability tradeoffs. Am I correct to assume that at pytorch entry time tensor would be interpreted and passed on as a pytorch object and wouldn't cause cause friction?

ezyang:
> if we assume dimension names are universal

This is a bonkers thing to assume though, like just get out of batch dimension and image and text data and like, the tensors dimensions can mean literally anything. You're not going to be able to standardize for every case.

pitrou:
Stepping back a little, ISTM that there are two major categories of use cases that should be satisfied:
* applications where the dimensions have well-known names that can map to physical order
* applications where the dimensions don't have well-known names (they are anonymous) but logical order is a non-trivial permutation of physical order (for example a Numpy array)

Is there a third category of use cases?

rok:
> > if we assume dimension names are universal
> 
> This is a bonkers thing to assume though, like just get out of batch dimension and image and text data and like, the tensors dimensions can mean literally anything. You're not going to be able to standardize for every case.

Indeed :D. And that's just on the application entry/exit. If you want to serialize a tensor somewhere mid computation graph you might as well use UUID for dimension naming.

> Is there a third category of use cases?

Nothing significant comes to mind.

rok:
@pitrou given the two cases - first could be solved with `dim_names` and second with either `strides` or `permutation`. Or did you want to make another point?

pitrou:
@rok I'm just saying that these should be satisfied by the final proposal that you (or someone else) will submit on the ML.

rok:
I was hoping you had a solution :D.
I'm starting to think optional `strides` might be most robust proposal here.

rok:
> `strides` will be annoying to consume by implementations, while not bringing any additional flexibility

How so? I see mostly as metadata to be passed between application. Only case where I'd use strides is when creating TensorArray with n-dim tensor elements from a single n+1-dim tensor.

pitrou:
> How so? I see mostly as metadata to be passed between application.

Ah, fair enough, perhaps I should have said "applications" instead of "implementations". But the point still holds.

rok:
> Ah, fair enough, perhaps I should have said "applications" instead of "implementations". But the point still holds.

Strides are the most common API (that is not `C` or `F`). For applications that don't use strides we could provide helpers for conversion to `C` or `F`.

pitrou:
I will circle back to https://github.com/apache/arrow/pull/33925#issuecomment-1422986313 and let you take that into account.

jorisvandenbossche:
Personally I am not sure what `strides` would provide that `dim_names` and/or `permutation` couldn't do, while being more difficult to work with and not being a full solution (it doesn't allow for names).

So I would go for either just an optional `dim_names` (with the convention that if you don't have actual names for dimensions but would like to indicate a typical permutation, you can use integer names to do that), or optional `dim_names` + optional `permutation` (if we think that it is important to be able to specify both separately, or if it's cleaner to not mix integers in the dim_names).

Both of those should cover the two use cases that Antoine mentioned above at https://github.com/apache/arrow/pull/33925#issuecomment-1422986313

(in the end, it's maybe mostly a choice between `strides` and `permutation`, as both hold the same information but in a different form, and we need `dim_names` anyway if we want to give the option to preserve dimension name information for applications that do natively have those)

rok:
I think optional `dim_names` + optional `permutation` would be a good design.
One potential issue would be: when a row-major application would pass data to a column-major application we'd want to force passing of permutation so data is not misinterpreted. I propose `permutation` is mandatory (it can be derived from `C`/`F` by our implementation easily).

AlenkaF:
I have updated the spec according to the discussion in this PR and the ML plus the talk with @rok and @jorisvandenbossche. You can find the change in _Description of the serialization_ section where `dim_names` and `permutations` are now included as an _optional_ metadata.

Please have a look at the text and give comments/suggest changes. Once the PR is ready I will send the new version to the ML for a vote.

Thank you everybody for the participation and feedback!

Screenshot 2023-02-15 at 08 12 16

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,57 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+* The storage type of the extension: ``FixedSizeList`` where:
+
+  * **value_type** is the data type of individual tensors and
+    is an instance of ``pyarrow.DataType`` or ``pyarrow.Field``.
+  * **list_size** is the product of all the elements in tensor shape.
+
+* Extension type parameters:
+
+  * **value_type** = Arrow DataType of the tensor elements
+  * **shape** = shape of the contained tensors as a tuple
```
jorisvandenbossche, line 89: `dim_names` and `permutation` are also parameters (optional). So maybe we could move the actual description of the parameters here, and below in the "Description of the serialization" bullet point only focus on the formatting of the parameters in the serialized metadata?

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,57 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+  - **"dim_names"** holds explicit names to tensor dimensions
+    as an array. The length of it should be equal to the shape
+    length and equal to the number of dimensions.
+
+    Example of dim_names metadata for NCHW ordered data:
+
+    ``{ "shape": [100, 200, 500], "dim_names": ["C", "H", "W"]}``
+
+    The ``dim_names`` metadata can be added if the dimensions have
+    well-known names that can map to the physical order (row-major).
```
jorisvandenbossche, line 109:
```suggestion
    well-known names and map to the physical order (row-major).
```

I am not sure the "can" is needed here. I would say that if you have dimension names, they always _can_ be mapped to the physical order, but what is relevant here is that indeed you _should_ list them in that order.

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,57 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+
+  - **"permutation"** holds indexes of the desired ordering of the
+    original dimensions. Also defined as an array.
```
jorisvandenbossche, line 112: Suggestion to add a more explicit explanation of what those values should be, something like: "The indices contain a permutation of the values [0, 1, .., N-1] where N is the number of dimensions. The permutation indicates which dimension of the logical order corresponds to which dimension of the physical tensor (the i-th dimension of the logical view corresponds to the dimension with number ``permutations[i]`` of the physical tensor).

docs/source/format/CanonicalExtensions.rst
AlenkaF: OK, will do - I misunderstood and thought we were only talking about defining the metadata for this cases, not adding them to parameters.

jorisvandenbossche: There is not a strict difference between both. The parameters are the information a user would need to provide to fully define an instance of this extension type (in in a C++ or Python API), and so the optional dim_names/permutation is part of that (even though they are not essential for the interpretation, hence they are optional). 
_Most_ of the parameters are then represented in the serizalization metadata (but `value_type` is not, because that parameter is already part of the physical storage type).

docs/source/format/CanonicalExtensions.rst
rok: Perhaps something like:
```
  - **"permutation"** holds indexes of the desired logical ordering of the
    existing physical dimensions and can be used to derive strides.
    Also defined as an array.
```
Also adding an example of permutation for representing NHWC data as NCHW could be something to do? We should also state the convention somewhere if we're going to use it for examples (N = element, H = image height, W = image width, C = color channel).

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,57 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+
+    Example of permuted 3-dimensional tensor:
+
+    ``{ "shape": [100, 200, 500], "permutation": [2, 0, 1]}``
+
+    **Permutation key in the metadata is needed only in case the logical
+    order of the tensor doesn't equal the physical order (row-major).**
```
rok: I'd slightly prefer layout to order:
```suggestion
    layout of the tensor doesn't equal the physical layout (row-major).**
```

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,64 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+    dimension of the logical order corresponds to which dimension of the
+    physical tensor (the i-th dimension of the logical view corresponds
+    to the dimension with number permutations[i] of the physical tensor).
+
+    **Permutation is only needed in case the logical order of
+    the tensor is a permutation of the physical order (row-major).**
+
+    When logical and physical order are equal, the permutation will
+    always equal to the original order of dimensions ([0, 1, .., N-1])
+    and therefore doesn't need to be defined.
```
rok, line 114:
```suggestion
    and therefore doesn't need to be defined. If permutation is absent row-major layout is assumed.
```

AlenkaF: Row-major layout is always assumed? Just the dims are not permutated in case the param is absent.

rok: Sorry, yes. In case permutation is absent, it is assumed logical layout matches the physical one.

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,65 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+  * **value_type** = Arrow DataType of the tensor elements
+  * **shape** = shape of the contained tensors as an array
+
+  Optional parameters:
+
+  * **dim_names** = explicit names to tensor dimensions
+    as an array. The length of it should be equal to the shape
+    length and equal to the number of dimensions.
+
+    ``dim_names`` are used if the dimensions have well-known
```
jorisvandenbossche, line 97:
```suggestion
    ``dim_names`` can be used if the dimensions have well-known
```

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,65 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+
+    ``{ "shape": [100, 200, 500], "dim_names": ["C", "H", "W"]}``
+
+  - Example of permuted 3-dimensional tensor:
+
+    ``{ "shape": [100, 200, 500], "permutation": [2, 0, 1]}``
+
+.. note::
+
+  Elements in an fixed shape tensor extension array are stored
```
jorisvandenbossche, line 135:
```suggestion
  Elements in a fixed shape tensor extension array are stored
```

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,65 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+    names and they map to the physical layout (row-major).
+
+  * **permutation**  = indices of the desired ordering of the
+    original dimensions, defined as an array.
+
+    The indices contain a permutation of the values [0, 1, .., N-1] where
+    N is the number of dimensions. The permutation indicates which
+    dimension of the logical layout corresponds to which dimension of the
+    physical tensor (the i-th dimension of the logical view corresponds
+    to the dimension with number permutations[i] of the physical tensor).
```
jorisvandenbossche, line 107:
```suggestion
    to the dimension with number ``permutations[i]`` of the physical tensor).
```

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,65 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+
+    **Permutation is only needed in case the logical order of
+    the tensor is a permutation of the physical order (row-major).**
+
+    When logical and physical layout are equal, the permutation will always
+    be ([0, 1, .., N-1]) and is therefore absent. Same holds the other way
```
jorisvandenbossche, line 113:
```suggestion
    be ([0, 1, .., N-1]) and can therefore be left out. Same holds the other way
```

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,65 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+    round: if permutation parameter is absent, it is assumed that logical
+    layout matches the physical one.
```
jorisvandenbossche, line 115: I think this last sentence is a bit redundant with the previous one?

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,65 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+    original dimensions, defined as an array.
+
+    The indices contain a permutation of the values [0, 1, .., N-1] where
+    N is the number of dimensions. The permutation indicates which
+    dimension of the logical layout corresponds to which dimension of the
+    physical tensor (the i-th dimension of the logical view corresponds
+    to the dimension with number permutations[i] of the physical tensor).
+
+    **Permutation is only needed in case the logical order of
+    the tensor is a permutation of the physical order (row-major).**
```
jorisvandenbossche, line 110:
```suggestion
    Permutation can be useful in case the logical order of
    the tensor is a permutation of the physical order (row-major).
```

I would maybe use the softer "can be useful" instead of "is needed", because even in the latter case, it's not strictly "needed", you are perfectly allowed to just store as the physical layout, or eg use dim_names to convey this information.

Nitpick: for the rest only keys are put in bold, so I also wouldn't put this in bold (it's not thé most important aspect of the whole spec)

jorisvandenbossche (commented):
Thanks for the update, looks good! Just a bunch of tiny wording nitpicks.

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,65 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+    original dimensions, defined as an array.
+
+    The indices contain a permutation of the values [0, 1, .., N-1] where
+    N is the number of dimensions. The permutation indicates which
+    dimension of the logical layout corresponds to which dimension of the
+    physical tensor (the i-th dimension of the logical view corresponds
+    to the dimension with number ``permutations[i]`` of the physical tensor).
+
+    Permutation can be useful in case the logical order of
+    the tensor is a permutation of the physical order (row-major).
```
paleolimbot, line 110: Does this apply to the entire child array or to each slice of the child array?

If it applies to the entire child array this is problematic for the ability to concatenate, subset, and/or slice extension arrays without an extension type. I would argue that if that is the case, this shouldn't be an extension type (it should be a core type that can be special-cased or not supported accordingly).

rok: Shape, permutation and shape all apply to a single element of an array.

rok: If you were to have an un-chunked tensor array, then the underlying buffer would have to have the first stride be the largest.

paleolimbot: Got it! If everything applies element-wise that seems like it wouldn't cause any deviation in the rules of the underlying storage type?

rok: I don't think so. The underlying `fixed_size_list` type is just a buffer we by default interpret as row-major. The producer has the option to store data with their own layout and provide the permutation needed to interpret their layout with at read time.

jorisvandenbossche: Indeed, we still follow the underlying storage type, so operations like slicing, filtering, take etc all work correctly (because indeed otherwise it could not be done as an extension type, but would have to be a new core type). And that's also the reason that the tensor has to be row-major instead of column-major (because otherwise it would violate the integrity of individual elements of the array).

GregoryKimball:
Thank you everyone for this discussion. The proposal looks good to me, and I would like your help checking my understanding.

In [RAPIDS libcudf](https://github.com/rapidsai/cudf), we would use an nested List type to represent the proposed Tensor type. In the case where `permutation` is not provided, I believe we could use the elements zero-copy and then create offsets based on the `shape` parameter. 

So this Tensor example with 2 rows of (2,3) tensors:
```
value_type: int32
shape: (2,3)
elements: [1,2,3,4,5,6,7,8,9,10,11,12]
```
would correspond to these List column children:
```
data_type: int32
offsets: [0,2,4]
  offsets: [0,3,6,9,12]
  elements: [1,2,3,4,5,6,7,8,9,10,11,12]
```

Would you please let me know if my conversion looks correct?

jorisvandenbossche:
> In [RAPIDS libcudf](https://github.com/rapidsai/cudf), we would use an nested List type to represent the proposed Tensor type. In the case where `permutation` is not provided, I believe we could use the elements zero-copy and then create offsets based on the `shape` parameter.

@GregoryKimball it seems that cudf only has a variable size List type (https://docs.rapids.ai/api/cudf/stable/user_guide/cudf.listdtype)? In the proposal here we use a fixed size list type, which doesn't have offsets. Of course you can use a variable size List type as well, but because each tensor element has the same shape (for this extension type, there is discussion to add a second type with variable shape), the offsets will increase regularly and are somewhat superfluous.
But the main buffer with all values can still be used zero-copy (which is most important here).

> would correspond to these List column children:

It's up to cudf to decide which data type to use exactly, but I am not fully sure if it is needed to use a _nested_ list type. For your example where you are using `LIST<LIST<INT32>>`, you could also use `LIST<INT32>` where each element of that list array is one flat tensor element (the `elements` array is still the same, you just have less offsets). But in the end it will depend on what you exactly want to do with this data which representation will be most useful.

rok:
Great to see libcudf underlying buffer layout is compatible with this proposal!
I do think variable shape case will additionally need to store tensor shapes as they generally probably cannot be derived.
The variable shape tensor array issue by @chrish42 is #24868.

GregoryKimball:
Thank you @jorisvandenbossche for your reply. Yes, in libcudf we only have a variable-length List type. It's expensive for us to introduce a new type, so our first step to support the proposed fixed-length Tensor type would be to convert to variable-length List type. Thank you for confirming that you expect the elements could be used zero-copy. 

Also, thank you for your idea to use flat lists (e.g. `LIST<INT32>`) instead of nested lists to represent tensors. Unless I'm mistaken, libcudf algorithms for joins, aggregations, stream compaction, etc would all correctly operate on flattened tensors. 

Thank you @rok for highlighting the variable-shape Tensor type proposal. I'll take a closer look at #24858 🙏

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,30 @@ same rules as laid out above, and provide backwards compatibility guarantees.
 Official List
 =============
 
-No canonical extension types have been standardized yet.
+Fixed shape tensor
+==================
+
+* Extension name: `arrow.fixed_shape_tensor`.
+
+* The storage type of the extension: ``FixedSizeList`` where:
```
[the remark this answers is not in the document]
lhoestq: Note that instead of `strides` we could store the dimensions permutation (as it seems to be the current direction) - and it could be a different one per extension array.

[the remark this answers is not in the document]
rok: I believe this was addressed and we can probably mark it as resolved?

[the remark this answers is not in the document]
jorisvandenbossche: Yes, I think in general this was discussed, just not here but in one of the many other threads.

docs/source/format/CanonicalExtensions.rst
AlenkaF: Sorry, I forgot to answer your comment here. I think the example of the permutation representing NHWC data as NCHW belongs to the implementations and the documentation of it. I plan to add examples of the usage and this would be a great thing to add.

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,67 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+
+  * **value_type** is the data type of individual tensor elements
+    and is an instance of Arrow ``DataType`` or ``Field``.
```
jorisvandenbossche, line 83:
```suggestion
  * **value_type** is the data type of individual tensor elements.
```

I think saying just this is sufficient (even "DataType" is strictly speaking a Arrow C++ specific term, and not necessarily each implementation uses that term). It should also be clear that FixedSizeList is an existing arrow type, and thus it follows from FixedSizeList definition how its child array is defined by a field with a type.

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,66 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+* Extension name: `arrow.fixed_shape_tensor`.
+
+* The storage type of the extension: ``FixedSizeList`` where:
+
+  * **value_type** is the data type of individual tensor elements.
+  * **list_size** is the product of all the elements in tensor shape.
+
+* Extension type parameters:
+
+  * **value_type** = Arrow DataType or Field of the tensor elements.
```
jorisvandenbossche, line 87:
```suggestion
  * **value_type** = the Arrow data type of individual tensor elements.
```

Removing the mention to Field here as well? (like it was removed from the equivalent item above describing the FixedSizeList)

jorisvandenbossche (approved):

extabgrad:
MATLAB’s Deep Learning Toolbox uses n-dimensional arrays which fit quite well with the proposal. It also has a special datatype called “dlarray” which is responsible for automatic differentiation. This datatype allows labelled dimensions, however, the labels are quite compatible with a permutation. We group dimension labels into categories – spatial/continuous, channel, batch, temporal, unspecified, and potential extensions – where the spatial and unspecified labels (and perhaps others in future) are allowed to be used for multiple dimensions. We find that ‘H’ and ‘W’ are very image-centric and do not extend arbitrarily, while ‘U’ might be used for additional dimensions needed for a variety of purposes in different contexts, such as filter groups, point cloud index and so forth. Consequently we need to keep track of permutations within dimensions with the same name.

Therefore our main input would be to request that permutation and dim_names are not mutually exclusive.

We would also like to ensure that the format will support complex data.

Questions:
1.	Could you clarify whether shape, dim_names and permutation are listed in left-to-right or right-to-left order. As in, is the contiguous dimension the right-most (C, python) or left-most (MATLAB, Fortran) dimension?
2.	Are 1-dimensional and 0-dimensional arrays allowed?
3.	“fixed_shape_array” or variations on multi-dimensional or nd array might be a more appropriate name given the mathematical implications of the term ‘tensor’.

Joss Knight
Development Manager
GPU and Deep Learning Team
MathWorks UK

rok:
Thanks you for the input and description of MATLAB's Deep Learning Toolbox @extabgrad !

> Therefore our main input would be to request that permutation and dim_names are not mutually exclusive.

As per [discussion on the mailing list](https://lists.apache.org/thread/zhdkotvjczcomzooxr5jtj2g0fmsyk5m) and this proposal I believe `permutation` and `dim_names` will not be mutually exclusive.

> We would also like to ensure that the format will support complex data.

Just to be clear: by complex you mean diverse not complex as in complex numbers?

> Questions:
> 
> 1. Could you clarify whether shape, dim_names and permutation are listed in left-to-right or right-to-left order. As in, is the contiguous dimension the right-most (C, python) or left-most (MATLAB, Fortran) dimension?

`shape`, `dim_names` and `permutation` would map to a row-major (C, python) physical layout of data. Data in the underlying buffer would have row-major layout.

> 2. Are 1-dimensional and 0-dimensional arrays allowed?

This wasn't discussed, but the current language doesn't forbid them, so I suppose they are allowed. Do you think we should explicitly allow them?

> 3. “fixed_shape_array” or variations on multi-dimensional or nd array might be a more appropriate name given the mathematical implications of the term ‘tensor’.

That's an interesting point. Do you know of a source where this is discussed/argued? I'd be in favor of array-like naming too. The consideration here would be that deep learning frameworks are pushing tensor name and seem to have a lot of momentum.

jorisvandenbossche:
> > We would also like to ensure that the format will support complex data.
> 
> Just to be clear: by complex you mean diverse not complex as in complex numbers?

And if you mean complex numbers: the `value_type` in this proposal can be _any_ arrow data type. At the moment, there is no "complex" data type defined in the Arrow format spec, though. But there is a proposal to add a canonical extension type for complex data (https://github.com/apache/arrow/pull/10565), and assuming that this would be included in Arrow, then you can specify a tensor array that uses complex numbers as the individual tensor elements.

> > 5\. “fixed_shape_array” or variations on multi-dimensional or nd array might be a more appropriate name given the mathematical implications of the term ‘tensor’.
> 
> That's an interesting point. Do you know of a source where this is discussed/argued? I'd be in favor of array-like naming too. The consideration here would be that deep learning frameworks are pushing tensor name and seem to have a lot of momentum.

I don't have an opinion on the whole "array vs tensor" debate, but one aspect in favor of "Tensor" is that it avoids the confusion about whether "FixedShapeArray" is actually an array or type class, and the duplication in the name of "FixedShapeArrayArray".

extabgrad:
> Just to be clear: by complex you mean diverse not complex as in complex numbers?

No, I mean complex numbers, which are increasingly used in AI workflows.

> `shape`, `dim_names` and `permutation` would map to a row-major (C, python) physical layout of data. Data in the underlying buffer would have row-major layout.

I find 'row-major' to be ambiguous. You can have a row major layout and still list dimensions left-to-right [row column page batch].

> > 2. Are 1-dimensional and 0-dimensional arrays allowed?
> 
> This wasn't discussed, but the current language doesn't forbid them, so I suppose they are allowed. Do you think we should explicitly allow them?

I don't know, I just know they are important in numpy.

> That's an interesting point. Do you know of a source where this is discussed/argued? I'd be in favor of array-like naming too. The consideration here would be that deep learning frameworks are pushing tensor name and seem to have a lot of momentum.

image

Are you confident it is clear this datatype is only to be used in machine learning contexts? Perhaps it should be called 'machine-learning-tensor' then?

rok:
> No, I mean complex numbers, which are increasingly used in AI workflows.

Interesting! As Joris states there is an independent effort to enable that.

> > `shape`, `dim_names` and `permutation` would map to a row-major (C, python) physical layout of data. Data in the underlying buffer would have row-major layout.
> 
> I find 'row-major' to be ambiguous. You can have a row major layout and still list dimensions left-to-right [row column page batch].

This proposal currently states: `Elements in a fixed shape tensor extension array are stored in row-major/C-contiguous order.`. We can amend that to be more general. Could you state what left-to-right means? I assume it's equal to [TensorFlow's minor-to-major](https://www.tensorflow.org/xla/shapes).

> Are you confident it is clear this datatype is only to be used in machine learning contexts? Perhaps it should be called 'machine-learning-tensor' then?

I hope it'll be used in a wider context. `FixedShapeNDArray`?

extabgrad:
> This proposal currently states: `Elements in a fixed shape tensor extension array are stored in row-major/C-contiguous order.`. We can amend that to be more general. Could you state what left-to-right means? I assume it's equal to [TensorFlow's minor-to-major](https://www.tensorflow.org/xla/shapes).


I can see it's confusing, since we are distinguishing between the order dimensions are indexed and the order they are stored. Left-to-right means new higher dimensions are added on the right, and right-to-left means they are added on the left.

The minor_to_major field definition states that the left-most entry refers to the contiguous dimension in memory. The order of indexing is then defined by the numbers in this field, so [0 1 2] means that X(i,j,k) is indexing position i+(M*j)+(M*N*k). In typical parlance where i represents the row index, this is therefore column-major layout. A minor_to_major of [1 0 2] would therefore be a strict row-major layout so that j can represent the column index. But of course by row-major people typically mean [2 1 0], which means actually k is the column index, since the left-most dimension will be the highest (pages).

If MATLAB data were stored in row-major format but still indexed left-to-right then i would become the column index and no permutation would be needed to translate between MATLAB data and other row-major formats. However, since everyone uses the row,column indexing convention and MATLAB indexes left-to-right, the data is therefore stored in column-major. If 'permutation' is an inherently left-to-right spec then it is [1 0 2] for MATLAB (swap the 'first' two dimensions in memory). But if it's a right-to-left spec then it would be [0 2 1] (swap the 'last' two dimensions in memory).

Hence the ambiguity - so I thought it best to check!

I note that permutation is effectively the inverse of minor_to_major. The former is the memory order relative to the indexing order and the latter is the other way around.

docs/source/format/CanonicalExtensions.rst
```diff
@@ -72,4 +72,76 @@ same rules as laid out above, and provide backwards compatibility guarantees.
...
+
+.. note::
+
+  Other Data Structures in Arrow include a
+  `Tensor (Multi-dimensional Array) <https://arrow.apache.org/docs/format/Other.html>`_
+  to be used as a message in the interprocess communication machinery (IPC).
+
+  This structure has no relationship with the Fixed shape tensor extension type defined
+  by this specification. With defining an extension type one can use fixed shape tensors
+  as elements in a field of a RecordBatch or a Table.
```
AlenkaF, line 147: @lidavidm just letting u know I have added a note on the IPC Tensor structure.

lidavidm: Thanks! The very last sentence is a little confusing to me - is it saying that this extension type lets us use tensors (not the IPC object, but conceptual tensors) as an Arrow data type?

AlenkaF: If I understand what you are saying, then I guess so: the canonical extension type for fixed shape tensors is used to define an Extension Array (which can be used in a Table/RecordBatch). And the storage type is Arrow data type.

lidavidm: Ok, how about this? (Just a small tweak)

```suggestion
.. note::

  Other Data Structures in Arrow include a
  `Tensor (Multi-dimensional Array) <https://arrow.apache.org/docs/format/Other.html>`_
  to be used as a message in the interprocess communication machinery (IPC).

  This structure has no relationship with the Fixed shape tensor extension type defined
  by this specification. Instead, this extension type lets one use fixed shape tensors
  as elements in a field of a RecordBatch or a Table.
```

AlenkaF: Better! :+1:

docs/source/format/CanonicalExtensions.rst
AlenkaF: @rok I have added an example of permutation for representing NCHW data as NHWC in the Python bindings documentation PR: https://github.com/apache/arrow/pull/34957/files

PR #33938, opened 2023-01-30
GH-33937: [C++] Implement the "covariance" function/kernels

### LINKED ISSUE #33937: Create the first binary aggregate function kernel to serve as an example for other implementations

#15083 introduced support for binary aggregate kernels, but there isn't a single such implementation right now. Having one would help with issues like #15103.

### Component(s)

C++

felipecrv:

This adds the `"covariance"` aggregate kernel and some refactoring that was needed to be able to support binary aggregate kernels.

### Are these changes tested?

Only one small unit test. More thorough tests and careful code review is needed before merging.

- [ ] Document the `"covariance"` kernel
* Closes: #33937

### DIFF
--- a/cpp/src/arrow/CMakeLists.txt
+++ b/cpp/src/arrow/CMakeLists.txt
@@ -459,6 +459,7 @@ if(ARROW_COMPUTE)
   list(APPEND
        ARROW_SRCS
        compute/kernels/aggregate_basic.cc
+       compute/kernels/aggregate_covariance.cc
        compute/kernels/aggregate_mode.cc
        compute/kernels/aggregate_quantile.cc
        compute/kernels/aggregate_tdigest.cc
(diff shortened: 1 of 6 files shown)


cpp/src/arrow/visitor_generate.h
```diff
@@ -65,4 +65,50 @@ namespace arrow {
...
+  ACTION(Int64, ARG1);                                      \
+  ACTION(UInt64, ARG1)
+
+#define ARROW_GENERATE_FOR_ALL_NUMERIC_TYPES2(ACTION, ARG1) \
+  ARROW_GENERATE_FOR_ALL_INTEGER_TYPES2(ACTION, ARG1);      \
+  ACTION(HalfFloat, ARG1);                                  \
+  ACTION(Float, ARG1);                                      \
+  ACTION(Double, ARG1)
+
+#define ARROW_GENERATE_FOR_ALL_TYPES2(ACTION, ARG1)    \
```
felipecrv, line 84: There were some changes in this file since I did this and there's a better approach I can take that doesn't need these new lines here.

PR #14241, opened 2022-09-26
ARROW-17846: [C++] Use `if constexpr` in CSV subsystem

pitrou:

PR #33941, opened 2023-01-30
MINOR: [C++] Fix a stale comment and declare a function constexpr

felipecrv:


cpp/src/arrow/type.h
```diff
@@ -2126,7 +2126,7 @@ Result<std::shared_ptr<Schema>> UnifySchemas(
 
 namespace internal {
 
-static inline bool HasValidityBitmap(Type::type id) {
+inline constexpr bool HasValidityBitmap(Type::type id) {
```
felipecrv, line 2129: This was mentioned in another PR of mine, but I'm pushing it here instead to reduce the noise there.

js8544: constexpr functions are implicitly inline so we don't need the inline keywork here, right?

felipecrv: Pushed a change.

js8544: I think the arrow codebase tends to omit inline keyword for constexpr functions, for example in type_trait.h and [this PR](https://github.com/apache/arrow/pull/14241). 
P.S. I don't think the inline specifier has much to do with inline function expansion for modern compilers. It's more for making sure ODR is not violated.

pitrou: Yes, according to the C++ reference, "A constexpr specifier used in a function declaration implies inline".
https://en.cppreference.com/w/cpp/language/constexpr

pitrou (approved):
LGTM, thank you @felipecrv

PR #33942, opened 2023-01-31
GH-15193: [C++][Parquet] Parquet FuzzReader add some fixed batch size

### LINKED ISSUE #15193: [C++][Parquet] Fuzz using several batch sizes

Exercising several batch sizes in `parquet::arrow::internal::FuzzReader` may help find more reader/decoder bugs.

### Component(s)

C++

mapleFU:
### Rationale for this change

Currently, parquet FuzzReader will only read fixed batch size. This patch change it using some fixed batch size.

Change parquet FuzzReader use some fixed batch size.

### Are these changes tested?

These changes are already testing in parquet :)

Because `FuzzReader` is an EXPORT api, a Fuzzer user may facing "error" when reading. But I think it's ok

* Closes: #15193

### DIFF
--- a/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
+++ b/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
@@ -5072,5 +5072,25 @@ TEST(TestArrowReadWrite, MultithreadedWrite) {
   ASSERT_NO_FATAL_FAILURE(::arrow::AssertTablesEqual(*table, *result));
 }
 
+TEST(TestArrowReadWrite, FuzzReader) {
+  constexpr size_t kMaxFileSize = 1024 * 1024 * 1;
+  {
+    auto path = test::get_data_file("PARQUET-1481.parquet", /*is_good=*/false);
+    PARQUET_ASSIGN_OR_THROW(auto source, ::arrow::io::MemoryMappedFile::Open(
+                                             path, ::arrow::io::FileMode::READ));
+    PARQUET_ASSIGN_OR_THROW(auto buffer, source->Read(kMaxFileSize));
+    auto s = internal::FuzzReader(buffer->data(), buffer->size());
+    ASSERT_NOT_OK(s);
+  }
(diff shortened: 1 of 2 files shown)


mapleFU:
@pitrou @wjones127  PTAL (^_^)

mapleFU:
By the way, because `generate_fuzz_corpus.cc` uses `Compression::BROTLI`, but `ARROW_FUZZING` compile flag not introduce `ARROW_WITH_BROTLI`, should we include brotli when fuzzing enable in cmake?

pitrou:
@github-actions crossbow submit -g fuzz

mapleFU:
Seems it passed? @pitrou

pitrou:
I think we want to ensure that it doesn't fail early on valid Parquet data, so we should probably add a trivial test for that (check that it returns `Status::OK` for some arbitrarily chosen Parquet data).

mapleFU:
> I think we want to ensure that it doesn't fail early on valid Parquet data, so we should probably add a trivial test for that (check that it returns `Status::OK` for some arbitrarily chosen Parquet data).

And seems that we should assure it failed on some bad input? By the way, what do you think of https://github.com/apache/arrow/pull/33942#issuecomment-1409804794 ?

pitrou:
> And seems that we should assure it failed on some bad input?

Yes, indeed!

> By the way, what do you think of https://github.com/apache/arrow/pull/33942#issuecomment-1409804794 ?

That sounds like a good idea to me.

mapleFU:
@pitrou Updated, PTAL
I'm not so familiar with CMake, so let me change brotli dependency in another patch later...

mapleFU:
@pitrou Hi, I update a testing and pass the ut, PTAL again?

pitrou:
@github-actions crossbow submit -g fuzz

pitrou (approved):
Thanks for the update @mapleFU . I will merge if the CI job is green.

PR #33945, opened 2023-01-31
GH-33944: [C++] Allow FnOnce to be used in ThreadPool::Submit

### LINKED ISSUE #33944: [C++] Support FnOnce in ThreadPool::Submit

I'd like to use FnOnce for the async task scheduler stuff in the query context.  However, there is currently a call to std::bind in ThreadPool::Submit that forces a copy.

### Component(s)

C++

westonpace:
* Closes: #33944

### DIFF
--- a/cpp/src/arrow/util/thread_pool.h
+++ b/cpp/src/arrow/util/thread_pool.h
@@ -186,6 +186,46 @@ class ARROW_EXPORT Executor {
                   std::forward<Args>(args)...);
   }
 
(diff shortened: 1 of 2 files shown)


westonpace:
Leaving as a draft at the moment.  I do not like the fact that this isn't an overload but if I try and replace the existing implementation then I get errors about some of the invocations.

westonpace:
Abandoned

Issue #33924, opened 2023-01-30
[Format] Fixed shape Tensor as a canonical extension type


AlenkaF:
### Describe the enhancement requested

Umbrella ticket for adding support for a column in an Arrow table that has values cells each containing a tensor value, with all tensors having the same shape/dimensions.

Initial discussion mostly happened in the C++ issue: #15483 / [ARROW-1614](https://issues.apache.org/jira/browse/ARROW-1614)

We propose to add a canonical extension type for tensors and this umbrella issue is including the draft for the specification plus the implementations (C++ , PyArrow - as an example or as an implementation in the codebase).

- [x] https://github.com/apache/arrow/issues/33923
- [x] https://github.com/apache/arrow/issues/15483
- [x] https://github.com/apache/arrow/issues/33947

PR #33948, opened 2023-01-31
GH-33947: [Python][Docs] Tensor canonical type extension example

### LINKED ISSUE #33947: [Python][Docs] Tensor canonical type extension example

We propose to add a canonical extension type for tensors and this issue is adding an example of the implementation and the usage of it in PyArrow. Mainly to be added to the ML together with the proposed canonical type specification.

### Component(s)

Python

AlenkaF:
### Rationale for this change
There have been quite a lot of discussions connected to the tensor support in Arrow Tables/RecorBatches (https://github.com/apache/arrow/issues/33924). This PR is a Python implementation example for the specification proposal to add tensors as a canonical type extensions which will be sent to the Mailing list for discussion and vote.

- Implementation of the fixed size tensor canonical extension type in PyArrow plus tests (as an example in test_extension_type.py)
- Example of the implementation and example of the usage in `pyarrow.Table` (added to extending_types.rst)

Both are **only illustrative**, do not merge!

### Are these changes tested?
- Tests for the implementation are added to test_extension_type.py

No.

* Closes: #33947

cc @rok @jorisvandenbossche


python/pyarrow/tests/test_extension_type.py
```diff
@@ -1079,3 +1082,279 @@ def test_array_constructor_from_pandas():
         pd.Series([1, 2, 3], dtype="category"), type=IntegerType()
     )
     assert result.equals(expected)
+
+
+class TensorType(pa.ExtensionType):
```
jorisvandenbossche, line 1087: Use `FixedShapeTensorType`, to follow the name change?

And the same for the TensorArray

python/pyarrow/tests/test_extension_type.py
```diff
@@ -1079,3 +1082,279 @@ def test_array_constructor_from_pandas():
...
+    def __init__(self, value_type, shape, is_row_major):
+        self._value_type = value_type
+        self._shape = shape
+        self._is_row_major = is_row_major
+        size = math.prod(shape)
+        pa.ExtensionType.__init__(self, pa.list_(self._value_type, size),
+                                  'arrow.fixed_size_tensor')
+
+    @property
+    def dtype(self):
```
jorisvandenbossche, line 1119:
```suggestion
    def value_type(self):
```

(to match the parameter)

python/pyarrow/tests/test_extension_type.py
```diff
@@ -1079,3 +1082,279 @@ def test_array_constructor_from_pandas():
...
+      [
+        100,
+        200,
+        300,
+        400
+      ]
+    ]
+    """
+
+    def to_numpy_tensor_list(self):
```
jorisvandenbossche, line 1199: It might be more useful to showcase conversion to/from a single numpy array (with ndim+1)?

python/pyarrow/tests/test_extension_type.py
```diff
@@ -1079,3 +1082,279 @@ def test_array_constructor_from_pandas():
...
+    pa.register_extension_type(tensor_type)
+    yield tensor_type, tensor_class
+    # teardown
+    try:
+        pa.unregister_extension_type('arrow.fixed_size_tensor')
+    except KeyError:
+        pass
+
+
+def test_generic_ext_type_tensor():
```
jorisvandenbossche, line 1246: I think you can remove the "generic_ext_type" part of the test names everywhere (that's from tests above that contrasted tests for py extension types with generic (non-python specific) extension types)

python/pyarrow/tests/test_extension_type.py
AlenkaF: Do you mean to change the method `to_numpy_tensor_list` to `to_numpy_tensor` where an array of numpy ndarrays is resulted or do you mean it would be useful to add an example as a docstring? Sorry, not sure about what you mean.

jorisvandenbossche: I meant to change what the method does. But also not returning an "array of arrays", but a single ndarray with 1 more dimension as the shape of the individual tensors.

(so for TensorArray with tensor shape of (2, 3) of length 10, return single array of shape (10, 2, 3))

AlenkaF: Got it, thanks!!

python/pyarrow/tests/test_extension_type.py
```diff
@@ -1079,3 +1082,279 @@ def test_array_constructor_from_pandas():
...
+    def to_numpy_tensor(self):
+        """
+        Convert tensor extension array to a numpy array (with dim+1).
+
+        Examples
+        --------
+
+        """
+        np_flat = np.array([])
+        for tensor in self.storage:
```
jorisvandenbossche, line 1208: It should be possible to do this without iterating over each element. If you do `to_numpy()` on the storage array's child values, you get the full column as one flat numpy array, that we can then reshape to the correct shape:

Given a storage array of fixed size list:

```In [29]: arr = pa.array([[1, 2], [3, 4]], pa.list_(pa.int64(), 2))

In [30]: arr
Out[30]: 
<pyarrow.lib.FixedSizeListArray object at 0x7ffb663b1000>
[
  [
    1,
    2
  ],
  [
    3,
    4
  ]
]

In [32]: arr.values
Out[32]: 
<pyarrow.lib.Int64Array object at 0x7ffb8b0848e0>
[
  1,
  2,
  3,
  4
]

In [33]: arr.values.to_numpy()
Out[33]: array([1, 2, 3, 4])

In [34]: arr.values.to_numpy().reshape(2, 2)
Out[34]: 
array([[1, 2],
       [3, 4]])
```

python/pyarrow/tests/test_extension_type.py
```diff
@@ -1079,3 +1082,282 @@ def test_array_constructor_from_pandas():
...
+      [
+        100,
+        200,
+        300,
+        400
+      ]
+    ]
+    """
+
+    def to_numpy_tensor(self):
```
AlenkaF, line 1212: @jorisvandenbossche I haven't used permutations parameter in `to_numpy_tensor()` or in `from_numpy_tensor()` as the shape of the ndarray changes if the dimensions are permutated with [transpose](https://numpy.org/doc/stable/reference/generated/numpy.transpose.html). What do you think?

jorisvandenbossche: I think that is fine to leave that out. It's up the user / application to know if they want to use the permutations anyway. You could maybe add an example of doing this using the available API: `np_arr = arr.to_numpy_tensor(); np_arr.transpose(arr.type.permutation)`

python/pyarrow/tests/test_extension_type.py
```diff
@@ -1079,3 +1082,282 @@ def test_array_constructor_from_pandas():
...
+        """
+        Convert tensor extension array to a numpy array (with dim+1).
+        """
+        np_flat = np.array(self.storage.values)
```
jorisvandenbossche, line 1216:
```suggestion
        np_flat = np.asarray(self.storage.values)
```

That will avoid a copy when not needed

python/pyarrow/tests/test_extension_type.py
rok: I'd perhaps name these methods `to_numpy` and `from_numpy`. Tensor part is already implied from context and `Tensor` class [uses the same names](https://github.com/apache/arrow/blob/master/python/pyarrow/tests/test_tensor.py#L78).

jorisvandenbossche: I wouldn't use `to_numpy()`, because that is an existing method on pyarrow Arrays, and from which you expect to get a 1D numpy array of the same length as the pyarrow array. 
While a nD array mostly will behave consistently as a 1D array (eg doing `result[0]` in both cases would give the same), I am still a bit hesitant to break the existing contract of `to_numpy()` in this case.

rok: That's fair. How about `to_numpy_array`? I'm skeptical of the word "tensor" as it doesn't appear in numpy.

AlenkaF: `to_ndarray`?

jorisvandenbossche: I would personally keep "numpy" in the name, given that it's explicitly a numpy ndarray, and not from some other package

AlenkaF: `to_numpy_ndarray`?

jorisvandenbossche: That's quite close to the original `to_numpy_tensor` ;) But either way, both are fine for me

rok: A nit: I think there are two schools of thought regarding naming tensors. One calls them tensors, the other multidimensional arrays. I think numpy is in the former camp and hence I'd avoid calling it `*_tensor`. :)

AlenkaF: @rok, can we leave as is or would you prefer ndarray to tensor?

AlenkaF: Sorry, GitHub failed to show me your last message while I had the site open. I will change to `to_numpy_ndarray` then.

AlenkaF:
Closing in favour of https://github.com/apache/arrow/pull/34883 and https://github.com/apache/arrow/pull/34957

PR #33951, opened 2023-01-31
GH-33936: [Go] C Data Interface: export dummy buffer for nil buffers

### LINKED ISSUE #33936: [Go] C Data Interface: export a dummy non-null buffer for null/0-size buffers

Related to #14875, #14805. 

Go exports a `NULL` buffer if the source buffer is either `nil` or 0-sized. While technically this is valid by #14805, libarrow < 12 doesn't handle it, and it would be easiest just to export a dummy buffer (ideally a static global) to make interoperability easier.

### Component(s)

Go

lidavidm:
### Rationale for this change

While #14805 clarified that NULL buffers are valid in the C Data Interface, not all implementations handle this correctly (or were fixed, but earlier versions still exist).

Export a non-NULL dummy buffer for nil/empty buffers to ensure compatibility.

### Are these changes tested?

A regression test is included.
* Closes: #33936

### DIFF
--- a/go/arrow/cdata/cdata_exports.go
+++ b/go/arrow/cdata/cdata_exports.go
@@ -381,7 +381,7 @@ func exportArray(arr arrow.Array, out *CArrowArray, outSchema *CArrowSchema) {
 		// unions don't have validity bitmaps, but we keep them shifted
 		// to make processing easier in other contexts. This means that
 		// we have to adjust for union arrays
-		if !internal.HasValidityBitmap(arr.DataType().ID(), flatbuf.MetadataVersionV5) {
+		if !internal.DefaultHasValidityBitmap(arr.DataType().ID()) {
 			out.n_buffers--
 			nbuffers--
 			bufs = bufs[1:]
@@ -390,7 +390,7 @@ func exportArray(arr arrow.Array, out *CArrowArray, outSchema *CArrowSchema) {
 		for i := range bufs {
 			buf := bufs[i]
 			if buf == nil || buf.Len() == 0 {
-				if i > 0 || !internal.HasValidityBitmap(arr.DataType().ID(), flatbuf.MetadataVersionV5) {
+				if i > 0 || !internal.DefaultHasValidityBitmap(arr.DataType().ID()) {
 					// apache/arrow#33936: export a dummy buffer to be friendly to
 					// implementations that don't import NULL properly
 					buffers[i] = (*C.void)(unsafe.Pointer(&C.kGoCdataZeroRegion))
--- a/go/arrow/cdata/cdata_exports.go
+++ b/go/arrow/cdata/cdata_exports.go
@@ -48,6 +48,8 @@ import (
 	"github.com/apache/arrow/go/v12/arrow"
 	"github.com/apache/arrow/go/v12/arrow/array"
 	"github.com/apache/arrow/go/v12/arrow/endian"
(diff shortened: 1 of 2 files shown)


go/arrow/cdata/cdata_exports.go
```diff
@@ -385,7 +388,13 @@ func exportArray(arr arrow.Array, out *CArrowArray, outSchema *CArrowSchema) {
 		for i := range bufs {
 			buf := bufs[i]
 			if buf == nil || buf.Len() == 0 {
-				buffers[i] = nil
+				if i > 0 {
```
pitrou, line 391: This implies that buffer 0 is the null bitmap, which might not be true for all types (such as unions).

By the way, the hardcoded check for unions above could instead use of more generic API (such as `HasValidityBitmap` in Arrow C++).

zeroshade: @lidavidm There's a `HasValidityBitmap` in `arrow/internal/utils.go` you can import `github.com/apache/arrow/go/v12/arrow/internal` and use `internal.DefaultHasValidityBitmap`

lidavidm: Updated.

go/arrow/cdata/cdata_exports.go
```diff
@@ -376,7 +381,7 @@ func exportArray(arr arrow.Array, out *CArrowArray, outSchema *CArrowSchema) {
 		// unions don't have validity bitmaps, but we keep them shifted
 		// to make processing easier in other contexts. This means that
 		// we have to adjust for union arrays
-		if arr.DataType().ID() == arrow.DENSE_UNION || arr.DataType().ID() == arrow.SPARSE_UNION {
+		if !internal.HasValidityBitmap(arr.DataType().ID(), flatbuf.MetadataVersionV5) {
```
zeroshade, line 384: you can use `DefaultHasValidityBitmap` to avoid having to specify the metadataversion (it'll always default to the most recent)

go/arrow/cdata/cdata_exports.go
```diff
@@ -385,7 +390,13 @@ func exportArray(arr arrow.Array, out *CArrowArray, outSchema *CArrowSchema) {
...
 			buf := bufs[i]
 			if buf == nil || buf.Len() == 0 {
-				buffers[i] = nil
+				if i > 0 || !internal.HasValidityBitmap(arr.DataType().ID(), flatbuf.MetadataVersionV5) {
+					// apache/arrow#33936: export a dummy buffer to be friendly to
+					// implementations that don't import NULL properly
+					buffers[i] = (*C.void)(unsafe.Pointer(&C.kGoCdataZeroRegion))
+				} else {
+					buffers[i] = nil
+				}
```
zeroshade, line 399: should this be behind some sort of flag so that we can easily switch back to using `nil` when the other implementations catch up? Or is that not worth it?

lidavidm: If/when we decide to remove it later, I think we should just remove it from the code entirely, instead of trying to flag it.

zeroshade (approved):
LGTM just the one question but it's not blocking

go/arrow/cdata/cdata_exports.go
lidavidm: The C++ side fix isn't even shipping until Arrow 12

PR #10289, opened 2021-05-10
PARQUET-1798: [C++] Review logic around automatic assignment of field_id's

westonpace:
Questions:

- This is my first PR in the parquet namespace, I'm not sure of all the special rules.
- The field ID generation doesn't happen on the `parquet::schema` -> `arrow::schema` phase but on the `parquet::format::schema` -> `parquet::schema` phase.  So in order to test I had to add `#include "generated/parquet_types.h"` to `arrow_schema_test.cc` and I wasn't sure if I was allowed to reference the `generated/*` files like that.
- This PR simply allows user specified field id's to be persisted.  Is that sufficient for PARQUET-1798 (the title is rather general) or should I open up a dedicated JIRA?

PR #33955, opened 2023-01-31
GH-33954: [C++][Parquet] Preserve field-id for nested type

### LINKED ISSUE #33954: [C++][Parquet] FieldToNode does not preserve field-id for nested type

The function below converts arrow schema to parquet schema.
```cpp
Status FieldToNode(const std::string& name, const std::shared_ptr<Field>& field,
                   const WriterProperties& properties,
                   const ArrowWriterProperties& arrow_properties, NodePtr* out);
```

It should preserve field-id of each parquet column but the implementation fails to do it for nested type.

### Component(s)

C++, Parquet

wgtmac:
### Rationale for this change

Function below converts arrow schema to parquet schema and preserves field-id for primitive types but loses that information for nested types.
```cpp
Status FieldToNode(const std::string& name, const std::shared_ptr<Field>& field,
                   const WriterProperties& properties,
                   const ArrowWriterProperties& arrow_properties, NodePtr* out);
```

In the Apache Iceberg, field-ids are assigned to all column types. We need to keep field-id of all columns.

Fix `FieldToNode` to preserve field-id of STRUCT/MAP/LIST types.

### Are these changes tested?

Added a test case in the `cpp/src/parquet/arrow/arrow_schema_test.cc`.

No.
* Closes: #33954

### DIFF
--- a/cpp/src/parquet/arrow/arrow_schema_test.cc
+++ b/cpp/src/parquet/arrow/arrow_schema_test.cc
@@ -1327,6 +1327,54 @@ TEST_F(TestConvertRoundTrip, FieldIdPreserveExisting) {
   ASSERT_EQ(thrift_field_ids, expected_field_ids);
 }
 
+TEST_F(TestConvertRoundTrip, FieldIdPreserveAllColumnTypes) {
+  std::vector<std::shared_ptr<Field>> arrow_fields;
+  arrow_fields.push_back(::arrow::field("c1", INT32, true, FieldIdMetadata(2)));
+  arrow_fields.push_back(::arrow::field("c2", DOUBLE, true, FieldIdMetadata(4)));
+  arrow_fields.push_back(::arrow::field("c3", DECIMAL_8_4, true, FieldIdMetadata(6)));
+  arrow_fields.push_back(::arrow::field("c4", UTF8, true, FieldIdMetadata(8)));
+
+  auto inner_struct =
+      ::arrow::struct_({::arrow::field("c5_1_1", UTF8, true, FieldIdMetadata(14))});
+  arrow_fields.push_back(::arrow::field(
+      "c5",
+      ::arrow::struct_({::arrow::field("c5_1", inner_struct, true, FieldIdMetadata(12)),
+                        ::arrow::field("c5_2", INT64, true, FieldIdMetadata(16))}),
+      true, FieldIdMetadata(10)));
+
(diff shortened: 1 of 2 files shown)


wgtmac:
@pitrou @wjones127 Could you please take a look? Thanks!

pitrou (approved):
This looks good to me. @jorisvandenbossche Do you remember if there were any problems with Parquet field ids?

jorisvandenbossche:
> This looks good to me. @jorisvandenbossche Do you remember if there were any problems with Parquet field ids?

I remember that we had previous PRs where we wondered what to do with it. Your comment at this PR summarizes how we currently handle it (https://github.com/apache/arrow/pull/10289#issuecomment-839634828):

> Based on my understanding, it seems that we should:
> * when reading from Parquet, reflect Parquet field_ids (if any) under the `PARQUET:field_id` metadata key
> * when writing to Parquet, generate Parquet field_ids from the `PARQUET:field_id` metadata key (if present)
> * not attempt to auto-generate any field_ids if they are not present in metadata

So correctly preserving the field-ids for nested types seems to follow that. 

@wgtmac the current test does a full roundtrip of the arrow<->parquet schemas, so I assume this indirectly also covers that we correctly write them to Parquet if the information is present in the field metadata of an Arrow schema?

wgtmac:
> > This looks good to me. @jorisvandenbossche Do you remember if there were any problems with Parquet field ids?
> 
> I remember that we had previous PRs where we wondered what to do with it. Your comment at this PR summarizes how we currently handle it ([#10289 (comment)](https://github.com/apache/arrow/pull/10289#issuecomment-839634828)):
> 
> > Based on my understanding, it seems that we should:
> > 
> > * when reading from Parquet, reflect Parquet field_ids (if any) under the `PARQUET:field_id` metadata key
> > * when writing to Parquet, generate Parquet field_ids from the `PARQUET:field_id` metadata key (if present)
> > * not attempt to auto-generate any field_ids if they are not present in metadata
> 
> So correctly preserving the field-ids for nested types seems to follow that.
> 
> @wgtmac the current test does a full roundtrip of the arrow<->parquet schemas, so I assume this indirectly also covers that we correctly write them to Parquet if the information is present in the field metadata of an Arrow schema?

Yes, the current fix preserves all field-ids that exist in the `PARQUET:field_id` metadata key for all data types. This is required if we intend to write a parquet file which contains nested types to follow what the Apache Iceberg expects. @jorisvandenbossche

PR #33965, opened 2023-02-01
GH-33875: [Go] Handle writing LargeString and LargeBinary types

### LINKED ISSUE #33875: [Go] How to handle large lists?

I'm encountering a panic when trying to generate a parquet with a very large list of (nested) strings in it:
```
panic: runtime error: slice bounds out of range [:-2147483636]

goroutine 22 [running]:
github.com/apache/arrow/go/v10/parquet/pqarrow.writeDenseArrow(0x1400059b4a0, {0x105684320, 0x14008189c80}, {0x105686490?, 0x14009402410?}, {0x14657c7c000, 0x4000b00, 0x8000000}, {0x14647c7a000, 0x4000b00, ...}, ...)
        /Users/minyoung/go/pkg/mod/github.com/apache/arrow/go/v10@v10.0.1/parquet/pqarrow/encode_arrow.go:442 +0x22b0
github.com/apache/arrow/go/v10/parquet/pqarrow.WriteArrowToColumn({0x10567d4a0, 0x140008667e0}, {0x105684320, 0x14008189c80}, {0x105686490?, 0x14009402410?}, {0x14657c7c000, 0x4000b00, 0x8000000}, {0x14647c7a000, ...}, ...)
        /Users/minyoung/go/pkg/mod/github.com/apache/arrow/go/v10@v10.0.1/parquet/pqarrow/encode_arrow.go:227 +0x1e8
github.com/apache/arrow/go/v10/parquet/pqarrow.(*ArrowColumnWriter).Write(0x14008879a08, {0x10567d4a0, 0x140008667e0})
        /Users/minyoung/go/pkg/mod/github.com/apache/arrow/go/v10@v10.0.1/parquet/pqarrow/encode_arrow.go:195 +0x420
github.com/apache/arrow/go/v10/parquet/pqarrow.(*FileWriter).WriteColumnChunked(0x14000976000, 0x1400166c840?, 0x14008879ae0?, 0x1?)
        /Users/minyoung/go/pkg/mod/github.com/apache/arrow/go/v10@v10.0.1/parquet/pqarrow/file_writer.go:282 +0xb8
github.com/apache/arrow/go/v10/parquet/pqarrow.(*FileWriter).WriteColumnData(0x14000976000?, {0x1056861f0, 0x1400059b4f0})
        /Users/minyoung/go/pkg/mod/github.com/apache/arrow/go/v10@v10.0.1/parquet/pqarrow/file_writer.go:291 +0xb0
github.com/apache/arrow/go/v10/parquet/pqarrow.(*FileWriter).Write(0x14000976000, {0x105685228, 0x14000964330})
        /Users/minyoung/go/pkg/mod/github.com/apache/arrow/go/v10@v10.0.1/parquet/pqarrow/file_writer.go:206 +0x1f4
main.generateParquet.func1()
        /Users/minyoung/repo/core/services/gorilla/cmd/adhoc/main.go:578 +0x3f8
golang.org/x/sync/errgroup.(*Group).Go.func1()
        /Users/minyoung/go/pkg/mod/golang.org/x/sync@v0.0.0-20220819030929-7fc1605a5dde/errgroup/errgroup.go:75 +0x5c
created by golang.org/x/sync/errgroup.(*Group).Go
        /Users/minyoung/go/pkg/mod/golang.org/x/sync@v0.0.0-20220819030929-7fc1605a5dde/errgroup/errgroup.go:72 +0xa4
exit status 2
```

I can recreate the error with:

```golang
func (ps *ParquetIOTestSuite) TestLargeListOfStrings() {
	bldr := array.NewStructBuilder(memory.DefaultAllocator, arrow.StructOf(
		arrow.Field{
			Nullable: true,
			Name:     "l",
			Type: arrow.ListOf(arrow.BinaryTypes.String),
		},
	))
	defer bldr.Release()

	lBldr := bldr.FieldBuilder(0).(*array.ListBuilder)
	sBldr := lBldr.ValueBuilder().(*array.StringBuilder)

	for i := 0; i < 10; i++ {
		bldr.Append(true)
		for j := 0; j < 5; j++ {
			lBldr.Append(true)
			sBldr.Append(strings.Repeat(fmt.Sprintf("%d:%d|", i, j), 250000000))
		}
	}

	arr := bldr.NewArray()
	defer arr.Release()

	field := arrow.Field{Name: "x", Type: arr.DataType(), Nullable: true}
	expected := array.NewTable(arrow.NewSchema([]arrow.Field{field}, nil),
		[]arrow.Column{*arrow.NewColumn(field, arrow.NewChunked(field.Type, []arrow.Array{arr}))}, -1)
	defer expected.Release()

	ps.roundTripTable(expected, false)
}
```

From skimming through the code and some breakpointing, it seems like the [int32 offsets](https://github.com/apache/arrow/blob/master/go/parquet/pqarrow/encode_arrow.go#L442) used has overflowed, hence the error with a number that's around -2^31. It seems like maybe I should use `LargeListOf` or maybe `LargeString` in the schema, but when attempting to do so I get a [not implemented yet](https://github.com/apache/arrow/blob/master/go/parquet/pqarrow/schema.go#L376-L377) error.

How much work would it be to implement `LargeListOf`/`LargeString`? And if that were done, would it resolve the panic?

Writing out smaller row groups does work (presumably because the buffers and offsets are reset), but we would like to keep the number of row groups small (we only access a very small number of columns at a time, and our datasets can get very very wide).

### Component(s)

Go

minyoung:
### Rationale for this change

Handle writing `array.LargeString` and `array.LargeBinary` data types. This allows parquet files to contain more than 2G worth of binary data in a single column chunk.

### Are these changes tested?

Unit tests included

* Closes: #33875

### DIFF
--- a/go/parquet/file/record_reader.go
+++ b/go/parquet/file/record_reader.go
@@ -756,10 +756,16 @@ func newByteArrayRecordReader(descr *schema.Column, info LevelInfo, dtype arrow.
 		mem = memory.DefaultAllocator
 	}
 
+	dt, ok := dtype.(arrow.BinaryDataType)
+	// arrow.DecimalType will also come through here, which we want to treat as binary
+	if !ok {
+		dt = arrow.BinaryTypes.Binary
+	}
+
 	return &binaryRecordReader{&recordReader{
 		recordReaderImpl: &byteArrayRecordReader{
 			createPrimitiveRecordReader(descr, mem, bufferPool),
-			array.NewBinaryBuilder(mem, dtype.(arrow.BinaryDataType)),
+			array.NewBinaryBuilder(mem, dt),
 			nil,
 		},
 		leafInfo:  info,
--- a/go/parquet/file/record_reader.go
+++ b/go/parquet/file/record_reader.go
@@ -751,20 +751,15 @@ type byteArrayRecordReader struct {
 	valueBuf []parquet.ByteArray
 }
 
-func newByteArrayRecordReader(descr *schema.Column, info LevelInfo, mem memory.Allocator, bufferPool *sync.Pool) RecordReader {
+func newByteArrayRecordReader(descr *schema.Column, info LevelInfo, dtype arrow.DataType, mem memory.Allocator, bufferPool *sync.Pool) RecordReader {
 	if mem == nil {
 		mem = memory.DefaultAllocator
-	}
-
-	dt := arrow.BinaryTypes.Binary
-	if descr.LogicalType().Equals(schema.StringLogicalType{}) {
-		dt = arrow.BinaryTypes.String
 	}
 
 	return &binaryRecordReader{&recordReader{
 		recordReaderImpl: &byteArrayRecordReader{
 			createPrimitiveRecordReader(descr, mem, bufferPool),
-			array.NewBinaryBuilder(mem, dt),
+			array.NewBinaryBuilder(mem, dtype.(arrow.BinaryDataType)),
 			nil,
 		},
 		leafInfo:  info,
@@ -841,10 +836,10 @@ func (br *byteArrayRecordReader) GetBuilderChunks() []arrow.Array {
 
 // TODO(mtopol): create optimized readers for dictionary types after ARROW-7286 is done
 
-func NewRecordReader(descr *schema.Column, info LevelInfo, readDict bool, mem memory.Allocator, bufferPool *sync.Pool) RecordReader {
+func NewRecordReader(descr *schema.Column, info LevelInfo, dtype arrow.DataType, mem memory.Allocator, bufferPool *sync.Pool) RecordReader {
 	switch descr.PhysicalType() {
 	case parquet.Types.ByteArray:
-		return newByteArrayRecordReader(descr, info, mem, bufferPool)
+		return newByteArrayRecordReader(descr, info, dtype, mem, bufferPool)
 	case parquet.Types.FixedLenByteArray:
 		return newFLBARecordReader(descr, info, mem, bufferPool)
 	default:
--- a/go/parquet/pqarrow/encode_arrow.go
+++ b/go/parquet/pqarrow/encode_arrow.go
@@ -231,6 +231,10 @@ type binaryarr interface {
 	ValueOffsets() []int32
 }
 
+type binary64arr interface {
+	ValueOffsets() []int64
+}
+
 func writeDenseArrow(ctx *arrowWriteContext, cw file.ColumnChunkWriter, leafArr arrow.Array, defLevels, repLevels []int16, maybeParentNulls bool) (err error) {
 	noNulls := cw.Descr().SchemaNode().RepetitionType() == parquet.Repetitions.Required || leafArr.NullN() == 0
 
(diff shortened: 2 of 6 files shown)


go/parquet/pqarrow/encode_arrow.go
```diff
@@ -421,12 +425,21 @@ func writeDenseArrow(ctx *arrowWriteContext, cw file.ColumnChunkWriter, leafArr
...
+		case arrow.STRING, arrow.BINARY:
+			off32 := leafArr.(binaryarr).ValueOffsets()
+			offsets = make([]int64, len(off32))
+			for i, o := range off32 {
+				offsets[i] = int64(o)
+			}
+		case arrow.LARGE_STRING, arrow.LARGE_BINARY:
+			offsets = leafArr.(binary64arr).ValueOffsets()
+		default:
+			return xerrors.New(fmt.Sprintf("invalid column type to write to ByteArray: %s", leafArr.DataType().Name()))
```
zeroshade, line 439: Rather than pay the cost to copy the offsets, i'd rather just duplicate the loop at line 453 based on whether you're using int32 or int64 offsets. Duplicating the 4 lines of code is better than the cost of copying a potentially large offset array just to adjust for the int32/int64. 

When go1.20 releases I'd consider it viable to upgrade us to using go1.18 as a default requirement, and we can make a generic version of the loop via a function and just have that work. But for now, let's just duplicate the loop and avoid copying the offset slice.

minyoung: That makes sense. Will shuffle the code around to use `ValueOffsets` as-is

go/parquet/file/record_reader.go
```diff
@@ -751,14 +751,15 @@ type byteArrayRecordReader struct {
...
 		mem = memory.DefaultAllocator
 	}
 
-	dt := arrow.BinaryTypes.Binary
-	if descr.LogicalType().Equals(schema.StringLogicalType{}) {
-		dt = arrow.BinaryTypes.String
+	dt, ok := dtype.(arrow.BinaryDataType)
+	// arrow.DecimalType will also come through here, which we want to treat as binary
+	if !ok {
+		dt = arrow.BinaryTypes.Binary
```
zeroshade, line 762: I don't like requiring the `arrow.DataType` here as part of the point of this interface was that you could create a record reader without needing to explicitly code the underlying datatype, it determines it from the parquet schema. Could we instead use the metadata to determine whether or not we need to use a LargeString/LargeBinary builder by checking how large the total data column is? Or some other way of determining it? I'm not sure if that would work, but it would be preferable if it is possible.

minyoung: > Could we instead use the metadata to determine whether or not we need to use a LargeString/LargeBinary builder by checking how large the total data column is?

This is a little tricky to do. We could potentially do https://github.com/apache/arrow/commit/8b526e5afe5dc1b9f2a8b264e2693f406161cbf2, but that's only summing the column chunk sizes and does not take into account the column chunk dictionary. Imagine 2M rows each referencing the same 1k string. The column chunk would have 1 string value in the dictionary, and all the rows would use that dictionary value. At [decoding time](https://github.com/apache/arrow/blob/go/v10.0.1/go/parquet/internal/utils/typed_rle_dict.gen.go#L1137), the values would get filled out to >2G.

In the context of `pqarrow.WithStoreSchema()`, it might also be a little surprising if the arrow schema says that a column is `arrow.LargeString`, but you get an `arrow.String` back (since total data column size is low). Since this specific change is only for `pqarrow.WithStoreSchema()`, so we could revert this change and say that reads will always return `arrow.String`

zeroshade: hmm, okay that's fair. I think it's fine as is for now. I agree that if the stored schema says `arrow.LargeString` you should get that back.

zeroshade (approved):

PR #33967, opened 2023-02-01
GH-33953: [Java] Pass custom headers on every request

### LINKED ISSUE #33839: [Java][FlightRPC] FlightCallHeaders added to FlightClient::getStream aren't attached to call

I'm using the Java flight-core version 10.0.1 and grpc-core version 1.50.2.  I'm having trouble attaching headers to a GET call to a server.
```java
Ticket ticket = new Ticket(/* ... */);
CallHeaders headers = new FlightCallHeaders();
headers.insert("foo", "bar");
CallOption opt = new HeaderCallOption(headers);
try (FlightStream stream = client.getStream(ticket, opt)) {
    // ...
}
```
I have logs that show the headers added to `opt` are not making it to the server.  My best guess at why this happens is due to https://github.com/apache/arrow/blob/master/java/flight/flight-core/src/main/java/org/apache/arrow/flight/FlightClient.java#L309.  `FlightClient` is ignoring the new stub returned by gRPC which has the header-injection interceptor added, see https://github.com/grpc/grpc-java/blob/master/stub/src/main/java/io/grpc/stub/MetadataUtils.java#L55.

Am I doing something wrong here?

### Component(s)

FlightRPC, Java

### LINKED ISSUE #33953: [Java] JDBC driver does not send custom headers on DoGet

Follow-up of #33874, specifically https://github.com/apache/arrow/issues/33874#issuecomment-1405501350

### Component(s)

Java

aiguofer:
### Rationale for this change

Some flight requests don't send custom headers. This PR should fix that.

Ensure custom headers are sent across on every request.

### Are these changes tested?

No

Custom headers should now be attached to every call.

* Closes: apache/arrow#33953
* Closes: apache/arrow#33839
* GitHub Issue: #33953


aiguofer:
Unsure what that failure is... maybe the `ManagedChannel` shutdown is not being propagated to other channels? Or maybe it's not waiting long enough for the shutdown to complete (there's a hardcoded 5 sec wait there I believe)

lidavidm:
It'll be a while before I can review here but that test is known to be flaky

aiguofer:
Great, good to know! We're currently using a JDBC driver jar I built with these changes for development of our service so we're not currently blocked. Hoping these changes can make it into the next release so we can have users use a published jar once we get to beta. I really appreciate your help and time!

java/flight/flight-core/src/main/java/org/apache/arrow/flight/FlightClient.java
```diff
@@ -353,10 +355,11 @@ public void onCompleted() {
    */
   public ExchangeReaderWriter doExchange(FlightDescriptor descriptor, CallOption... options) {
     Preconditions.checkNotNull(descriptor, "descriptor must not be null");
-    final io.grpc.CallOptions callOptions = CallOptions.wrapStub(asyncStub, options).getCallOptions();
+    FlightServiceStub stub = CallOptions.wrapStub(asyncStub, options);
+    final io.grpc.CallOptions callOptions = stub.getCallOptions();
 
     try {
-      final ClientCall<ArrowMessage, ArrowMessage> call = interceptedChannel.newCall(doExchangeDescriptor, callOptions);
+      final ClientCall<ArrowMessage, ArrowMessage> call = stub.getChannel().newCall(doExchangeDescriptor, callOptions);
```
lidavidm, line 362: Is it possible to factor out a helper that can take care of the wrapStub - getChannel - newCall dance?

lidavidm (commented):
Can we add unit tests here?

java/flight/flight-core/src/main/java/org/apache/arrow/flight/FlightClient.java
aiguofer: it'd be nice to add something like:

```
    public <RequestT, ResponseT> ClientCall<RequestT, ResponseT> newCall(MethodDescriptor<RequestT, ResponseT> descriptor) {
      return getChannel().newCall(descriptor, getCallOptions());
    }
```

to the stub so we could just do `stub.newCall(descriptor)`, but it looks like that class is auto-generated so I'm not sure where to add it.

aiguofer:
Sure I'll take a stab at writing some unit tests. I was thinking maybe something like `ConnectionTest` with a middleware that checks for the specific header values.

java/flight/flight-sql-jdbc-driver/src/test/java/org/apache/arrow/driver/jdbc/CustomHeaderTest.java
```diff
@@ -0,0 +1,118 @@
...
+    properties.put("answer", "42");
+    properties.put("has_towel", "true");
+
+    try (Connection connection = DriverManager.getConnection(
+        "jdbc:arrow-flight-sql://" + FLIGHT_SERVER_TEST_RULE.getHost() + ":" +
+            FLIGHT_SERVER_TEST_RULE.getPort(), properties)) {
+
+      try {
+        // We don't care if the call fails, we just want to make sure the headers got passed
+        connection.getMetaData().getTableTypes();
```
aiguofer, line 102: It'd be great to have a sequence of calls that tests most grpc methods, but at this point I'm not familiar enough with Arrow SQL to write that.

Depending on what those calls might look like, it could be better to have multiple tests, but figured this simple test is at least a good starting point.

java/flight/flight-core/src/main/java/org/apache/arrow/flight/FlightClient.java
lidavidm: It can just be a private method, no?

aiguofer: Oh I meant this would be great as a public method in `FlightServiceStub`. There's a lot of code in there that could use it, but I believe that is an auto-generated source file so I don't know how to change it.

I went ahead and created a private method in `FlightClient` to do the same.

java/flight/flight-sql-jdbc-driver/src/test/java/org/apache/arrow/driver/jdbc/CustomHeaderTest.java
aiguofer: Yeah makes sense, unfortunately I don't know the code well enough to do that. I know I encountered the issue by having properties in the JDBC connection that didn't make it to the server, and that's how I've been testing that the changes work.

Unfortunately I don't think I have the time right now to dig into it to write better unit tests... some help there would be great.

java/flight/flight-core/src/main/java/org/apache/arrow/flight/FlightClient.java
lidavidm: It's generated by gRPC, so it's not under our control

aiguofer:
@lidavidm Now that I'm a bit more familiar with the code base I was able to write some better unit tests. I moved the tests to `flight-core`, removed the use of rules, and I'm now testing each specific flight method instead of using the JDBC connection proxy.

aiguofer:
for the future... is there some command I can run to auto-format code to checkstyle? I haven't thought about checkstyles compliance in years, this should be automated.

lidavidm:
> for the future... is there some command I can run to auto-format code to checkstyle? I haven't thought about checkstyles compliance in years, this should be automated.

Unfortunately no. I would much prefer if we could switch to a formatter like google-java-format rather than checkstyle but that would be a somewhat disruptive refactor.

That said - feel free to raise an issue and get opinions from other Java contributors.

lidavidm (approved):
This is great, thank you!

lidavidm:
Actually, I filed https://github.com/apache/arrow/issues/34364 for checkstyle - feel free to chime in there

Issue #33668, opened 2023-01-15
Reading flat dataset with `partitioning="hive"` results in partition schema equal to dataset schema


jhostetler:
### Describe the bug, including details regarding any error messages, version, and platform.

Example code:
```
% ls flat/
4dc78608-2739-4b67-942a-6eb39d39fe62.999.parquet
% python
>>> hive_ds = pyarrow.dataset.dataset("flat/", partitioning="hive", format="parquet")
>>> hive_ds.partitioning
<pyarrow._dataset.HivePartitioning object at 0x1012a7990>
>>> hive_ds.partitioning.schema
_index_: int64
label: list<item: string>
  child 0, item: string
score: list<item: float>
  child 0, item: float
>>> hive_ds.schema
_index_: int64
label: list<item: string>
  child 0, item: string
score: list<item: float>
  child 0, item: float
>>>
>>> flat_ds = pyarrow.dataset.dataset("flat/", format="parquet")
>>> flat_ds.schema
_index_: int64
label: list<item: string>
  child 0, item: string
score: list<item: float>
  child 0, item: float
>>> flat_ds.partitioning
>>> flat_ds.partitioning is None
True
```

Notice how when using `partitioning="hive"`, the partitioning schema is the same as the dataset schema. Since the data is not, in fact, partitioned, I would expect `hive_ds.partitioning` to be `None`.

PyArrow 10.0.1, Linux, same result with data on local filesystem and when retrieving from GCS.

### Component(s)

Python

westonpace:
I can confirm.  I reproduced this with the latest and agree it is a bug.

westonpace:
Labeling as critical as this causes invalid results given valid input (not sure if this is the same label being discussed in https://github.com/apache/arrow/pull/33660/files or not).

vibhatha:
take

vibhatha:
I am removing my assignment, since I am mostly maintaining Arrow Java.

PR #33969, opened 2023-02-01
GH-33668: [Python] Reading flat dataset with partitioning="hive" results in partition schema equal to dataset schema

vibhatha:
This PR includes a minor modification to `HivePartitioningFactory` to fix an error caused when reading non-partitioned data in hive format. And test cases for Python and C++ has been added to evaluate the update.

### DIFF
--- a/cpp/src/arrow/dataset/partition.cc
+++ b/cpp/src/arrow/dataset/partition.cc
@@ -827,7 +827,7 @@ class HivePartitioningFactory : public KeyValuePartitioningFactory {
   Result<std::shared_ptr<Partitioning>> Finish(
       const std::shared_ptr<Schema>& schema) const override {
     if (dictionaries_.empty()) {
-      return std::make_shared<HivePartitioning>(schema, dictionaries_);
+      return std::make_shared<HivePartitioning>(arrow::schema({}), dictionaries_);
     } else {
       for (FieldRef ref : field_names_) {
         // ensure all of field_names_ are present in schema
--- a/cpp/src/arrow/dataset/partition.cc
+++ b/cpp/src/arrow/dataset/partition.cc
@@ -825,17 +825,17 @@ class HivePartitioningFactory : public KeyValuePartitioningFactory {
   }
 
   Result<std::shared_ptr<Partitioning>> Finish(
-      const std::shared_ptr<Schema>& schema) const override {
+      const std::shared_ptr<Schema>& dataset_schema) const override {
     if (dictionaries_.empty()) {
-      return std::make_shared<HivePartitioning>(arrow::schema({}), dictionaries_);
+      return std::make_shared<HivePartitioning>(schema({}), dictionaries_);
     } else {
       for (FieldRef ref : field_names_) {
         // ensure all of field_names_ are present in schema
-        RETURN_NOT_OK(ref.FindOne(*schema));
+        RETURN_NOT_OK(ref.FindOne(*dataset_schema));
       }
 
       // drop fields which aren't in field_names_
-      auto out_schema = SchemaFromColumnNames(schema, field_names_);
+      auto out_schema = SchemaFromColumnNames(dataset_schema, field_names_);
 
       return std::make_shared<HivePartitioning>(std::move(out_schema), dictionaries_,
                                                 options_.AsHivePartitioningOptions());
--- a/cpp/src/arrow/dataset/partition.cc
+++ b/cpp/src/arrow/dataset/partition.cc
@@ -827,7 +827,7 @@ class HivePartitioningFactory : public KeyValuePartitioningFactory {
   Result<std::shared_ptr<Partitioning>> Finish(
       const std::shared_ptr<Schema>& dataset_schema) const override {
     if (dictionaries_.empty()) {
-      return std::make_shared<HivePartitioning>(schema({}), dictionaries_);
+      return std::make_shared<HivePartitioning>(dataset_schema, dictionaries_);
     } else {
       for (FieldRef ref : field_names_) {
         // ensure all of field_names_ are present in schema
--- a/cpp/src/arrow/dataset/partition_test.cc
+++ b/cpp/src/arrow/dataset/partition_test.cc
@@ -84,8 +84,16 @@ class TestPartitioning : public ::testing::Test {
   void AssertInspect(const std::vector<std::string>& paths,
                      const std::vector<std::shared_ptr<Field>>& expected) {
     ASSERT_OK_AND_ASSIGN(auto actual, factory_->Inspect(paths));
-    ASSERT_EQ(*actual, Schema(expected));
-    ASSERT_OK_AND_ASSIGN(partitioning_, factory_->Finish(actual));
+    // ASSERT_EQ(*actual, Schema(expected));
+    // ASSERT_OK_AND_ASSIGN(partitioning_, factory_->Finish(actual));
+    Schema expected_schema(expected);
+    ASSERT_EQ(*actual, expected_schema);
+    // The dataset will usually have more than just partition fields.  The partitioning
+    // schema should ignore these.
+    std::vector<std::shared_ptr<Field>> all_fields(expected);
+    all_fields.push_back(field("field_unrelated_to_partitioning", int8()));
+    ASSERT_OK_AND_ASSIGN(partitioning_, factory_->Finish(schema(std::move(all_fields))));
+    AssertSchemaEqual(expected_schema, *partitioning_->schema());
   }
 
   void AssertPartition(const std::shared_ptr<Partitioning> partitioning,
@@ -572,6 +580,11 @@ TEST_F(TestPartitioning, DiscoverHiveSchema) {
   // missing path segments will not cause an error
   AssertInspect({"/alpha=0/beta=1", "/beta=2/alpha=3", "/gamma=what"},
                 {Int("alpha"), Int("beta"), Str("gamma")});
+
+  // No partitioning information should yield an empty partitioning even if
+  // there are fields in the dataset schema
+  AssertInspect({"/foo/chunk.parquet", "/bar/chunk.parquet", "/baz"}, {});
+  AssertInspect({"chunk.parquet"}, {});
 }
 
 TEST_F(TestPartitioning, HiveDictionaryInference) {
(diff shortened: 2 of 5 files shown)


vibhatha:
cc @westonpace there are a few CIs failing, not sure if they are related. Could you please review this PR?

cpp/src/arrow/dataset/partition.cc
```diff
@@ -827,7 +827,7 @@ class HivePartitioningFactory : public KeyValuePartitioningFactory {
   Result<std::shared_ptr<Partitioning>> Finish(
       const std::shared_ptr<Schema>& schema) const override {
     if (dictionaries_.empty()) {
-      return std::make_shared<HivePartitioning>(schema, dictionaries_);
+      return std::make_shared<HivePartitioning>(arrow::schema({}), dictionaries_);
```
westonpace, line 830: Nit: Is the `arrow::` prefix necessary here?

cpp/src/arrow/dataset/partition_test.cc
```diff
@@ -1107,5 +1108,88 @@ TEST(TestStripPrefixAndFilename, Basic) {
                                           "year=2019/month=12/day=01"));
 }
 
+TEST_F(TestPartitioning, HivePartitionReadEvaluation) {
```
westonpace, line 1111: This test is too complex.  Could you create a narrower unit test that focuses on the needed behavior?  For example, there should be no need to create a filesystem and actually write files if your goal is to test the partitioning factory.  The partitioning factory takes strings & schemas as input and outputs a partitioning object.

For example, you can reproduce the issue with this:

```diff
diff --git a/cpp/src/arrow/dataset/partition_test.cc b/cpp/src/arrow/dataset/partition_test.cc
index 3e681a9cb..0e22152e0 100644
--- a/cpp/src/arrow/dataset/partition_test.cc
+++ b/cpp/src/arrow/dataset/partition_test.cc
@@ -83,8 +83,14 @@ class TestPartitioning : public ::testing::Test {
   void AssertInspect(const std::vector<std::string>& paths,
                      const std::vector<std::shared_ptr<Field>>& expected) {
     ASSERT_OK_AND_ASSIGN(auto actual, factory_->Inspect(paths));
-    ASSERT_EQ(*actual, Schema(expected));
-    ASSERT_OK_AND_ASSIGN(partitioning_, factory_->Finish(actual));
+    Schema expected_schema(expected);
+    ASSERT_EQ(*actual, expected_schema);
+    // The dataset will usually have more than just partition fields.  The partitioning
+    // schema should ignore these.
+    std::vector<std::shared_ptr<Field>> all_fields(expected);
+    all_fields.push_back(field("field_unrelated_to_partitioning", int8()));
+    ASSERT_OK_AND_ASSIGN(partitioning_, factory_->Finish(schema(std::move(all_fields))));
+    AssertSchemaEqual(expected_schema, *partitioning_->schema());
   }
 
   void AssertPartition(const std::shared_ptr<Partitioning> partitioning,
@@ -571,6 +577,11 @@ TEST_F(TestPartitioning, DiscoverHiveSchema) {
   // missing path segments will not cause an error
   AssertInspect({"/alpha=0/beta=1", "/beta=2/alpha=3", "/gamma=what"},
                 {Int("alpha"), Int("beta"), Str("gamma")});
+
+  // No partitioning information should yield an empty partitioning even if
+  // there are fields in the dataset schema
+  AssertInspect({"/foo/chunk.parquet", "/bar/chunk.parquet", "/baz"}, {});
+  AssertInspect({"chunk.parquet"}, {});
 }
 
 TEST_F(TestPartitioning, HiveDictionaryInference) {
```

python/pyarrow/tests/parquet/test_dataset.py
```diff
@@ -1925,3 +1925,20 @@ def test_write_to_dataset_kwargs_passed(tempdir, write_dataset_kwarg):
...
+
+
+def test_partition_schema_validity(tempdir):
+    data = {
+        "a": [1, 2, 1, 2, 3, 4],
+        "b": [10, 20, 30, 40, 50, 60]
+    }
+
+    table = pa.Table.from_pydict(data)
+    pq.write_table(table, tempdir / "table.parquet")
```
westonpace, line 1937: A full round-trip test here is ok as it is much simpler to write such a test and we don't fully expose `PartitioningFactory` so it wouldn't be possible to do the simpler test.

python/pyarrow/tests/test_dataset.py
```diff
@@ -5084,10 +5084,8 @@ def test_dataset_partition_with_slash(tmpdir):
     read_table = ds.dataset(
         source=path,
         format='ipc',
-        partitioning='hive',
-        schema=pa.schema([pa.field("exp_id", pa.int32()),
-                          pa.field("exp_meta", pa.utf8())])
-    ).to_table().combine_chunks()
+        schema=dt_table.schema,
```
westonpace, line 5087: Does the schema even need to be specified here?

westonpace (changes requested):
I think the failures are unrelated and the fix looks correct (thanks) but we should simplify the C++ unit test.

python/pyarrow/tests/test_dataset.py
vibhatha: Not at all, but there was a bug when the schema was specified.

vibhatha: To evaluate that, I left it as it is but simplified it.

cpp/src/arrow/dataset/partition.cc
vibhatha: There is an arugment `schema` to the function, so it would need either refactoring argument or using the namespace. Any thoughts?

cpp/src/arrow/dataset/partition_test.cc
vibhatha: I removed the added test case and updated with your suggestion. I understand your concern.

vibhatha: should we include another test case? Or this change would be satisfactory?

python/pyarrow/dataset.py
```diff
@@ -437,7 +437,7 @@ def _filesystem_dataset(source, schema=None, filesystem=None,
     FileSystemDataset
     """
     format = _ensure_format(format or 'parquet')
-    partitioning = _ensure_partitioning(partitioning)
+    partitioning = _ensure_partitioning(partitioning, schema)
```
jorisvandenbossche, line 440: This doesn't seem right to me, since the user-specified `schema` is meant to be the full dataset schema, while the `partitioning` functions takes the schema of the only the partitioning (although it is then a bit surprising that no tests fails). 

Is this change needed to fix the issue (apart from the fix in C++)? 
(on first sight, it doesn't seem needed to have the python test you added passing)

python/pyarrow/tests/parquet/test_dataset.py
```diff
@@ -1925,3 +1925,20 @@ def test_write_to_dataset_kwargs_passed(tempdir, write_dataset_kwarg):
         pq.write_to_dataset(table, path, **{key: arg})
         _name, _args, kwargs = mock_write_dataset.mock_calls[0]
         assert kwargs[key] == arg
+
+
+def test_partition_schema_validity(tempdir):
```
jorisvandenbossche, line 1930: Can you move this tests to `tests/test_dataset.py`? (this file is specifically for the ParquetDataset implemention part of `pyarrow.parquet`)

python/pyarrow/tests/test_dataset.py
jorisvandenbossche: I am a trying to gauge the diff here, but did something actually change? It seems that the schema that was specified manually before should be identical to `dt_table.schema`, and so this change is just a simplification? 
But what is then the "bug when the schema was specified" you are mentioning?

vibhatha: In a previous PR I added this test case. It works without providing the schema (with the current change) and we don't need the change I did in the `_ensure_scheme`. But if a user provide a schema in this method, I get nulls for the `exp_meta` column after making this change. That was the bug encountered. I wasn't quite confident changing the test case to make things work. Do you agree with the C++ change to fix this bug? If so, for Python we could just leave the `schema`, but in case someone provides it there, we could run into trouble.

python/pyarrow/dataset.py
vibhatha: Yes this wasn't needed for the Python test case I added in this PR. But it was needed for the previous one when the schema is specified. If it is not specified even with this change it works fine. But when specified I get `null` values in the `exp_meta` column.

python/pyarrow/tests/test_dataset.py
jorisvandenbossche: > It works without providing the schema

But a schema_already_ was specified before this PR? I don't understand what this PR is changing about this test. 

It might be clearer if you add a new test that fails on current master, and is fixed by this PR.

vibhatha: I have added a new test to evaluate the failing idea in the `test_dataset.py`, but an existing test case fails with the current modification.

vibhatha: The issue is with the existing test case. When the schema is provided that case failed, but I realized the fix was to handle it where the `_ensure_scheme` is defined.

vibhatha: I could be wrong here. Any thoughts?

jorisvandenbossche: > The issue is with the existing test case. When the schema is provided that case failed

Can you show the error that you got for this test before you made the changes to `_ensure_partitioning`?

(and to be clear, the test also fails without the modifications to the test code? As those changes are only stylistic?)

vibhatha: Ah it is simple. Basically the data in the `exp_meta` is all `null` values, it shouldn't be the case. 
What I observed was, when we create the scheme, we should pass the schema in that case. Maybe that is a wrong fix.

cpp/src/arrow/dataset/partition.cc
westonpace: Let's disambiguate then.  Can you change the argument name to `dataset_schema`?

cpp/src/arrow/dataset/partition_test.cc
westonpace: This is fine.  It covers your change.  Are there other branches you feel we are not testing?

cpp/src/arrow/dataset/partition.cc
vibhatha: Sure. I will.

cpp/src/arrow/dataset/partition_test.cc
vibhatha: Not quite. But I felt the need for a roundtrip test. Other than that, I don’t have strong concerns.

python/pyarrow/tests/test_dataset.py
westonpace: I am also confused.  Here is my understanding of this test:

 * A user wants to partition on column exp_meta
 * The values in exp_meta contain slashes
 * We should be able to partition on that column and still write and then read the dataset

The files that will get created will be of the shape:

```
<tmpdir>/slash-writer-x/exp_meta=experiment%2FA%2Ff.csv/chunk-0.parquet
```

There should be no need to specify a schema manually to read this dataset.

It should also be harmless to specify a schema.  However, doing so makes the test less clear.

vibhatha: I agree, we don't have to provide the schema. In the PR we added this, I have included the schema. I think this could be my misunderstanding. 

But my concern is, if one does specify that schema, with this new change, we are getting null values in the `exp_meta` column.

vibhatha: In theory we don't need to pass the schema here, but if one does, we get invalid results. I am not quite sure which one to change :thinking:

cpp/src/arrow/dataset/partition_test.cc
westonpace: We have quite a few roundtrip tests for dataset reading and writing already (python's test_dataset has quite a few) so I'd be opposed to another one unless there was a specific reason.

python/pyarrow/tests/test_dataset.py
westonpace: > In theory we don't need to pass the schema here, but if one does, we get invalid results

In the current code?  Or in this PR?  Or both?

Both passing and not passing the schema should work.  If passing the schema gives invalid results then I would ask we follow @jorisvandenbossche 's advice and create a test that isolates this behavior as much as possible.  This may require further debugging to figure out exactly what happened.

vibhatha: Yes, I will further debug and see how to replicate this with an isolated test.

westonpace: I am going to remove my review request until we get this sorted out, just for my own book-keeping.  Feel free to re-request as soon as this is ready for another look.

westonpace (changes requested):
Marking "request changes" while we investigate whether there is a bug with specifying the schema or not.

vibhatha:
> Marking "request changes" while we investigate whether there is a bug with specifying the schema or not.

Yeah, I will investigate this further.

amol-:

Issue #33972, opened 2023-02-01
[Python] Remove redundant S3 call


Fokko:
### Describe the enhancement requested

Hey all,

First of all thanks everyone for working on PyArrow! Really loving it so far. I'm currently working on PyIceberg that will load an Iceberg table in PyArrow. For those unfamiliar with Apache Iceberg. This is a table format that focusses on having huge tables (petabyte size). PyIceberg makes you life easier by taking care of statistics to boost performance, and all the schema maintenance. For example, if you change the partitioning of an Iceberg table, you don't have to directly rewrite all the files, you can do this in an incremental way.

Now I'm running into some performance issues, and I noticed that PyArrow is doing more queries than required to S3. I went down the rabbit hole, and was able to narrow it down to:

```python
import pyarrow.dataset as ds
from pyarrow.fs import S3FileSystem
ONE_MEGABYTE = 1024 * 1024

client_kwargs = {
    "endpoint_override": "http://localhost:9000",
    "access_key": "admin",
    "secret_key": "password",
}
parquet_format = ds.ParquetFileFormat(
    use_buffered_stream=True,
    pre_buffer=True,
    buffer_size=8 * ONE_MEGABYTE
)
fs = S3FileSystem(**client_kwargs)
with fs.open_input_file("warehouse/wh/nyc/taxis/data/tpep_pickup_datetime_day=2022-04-30/00003-4-89e0ad58-fb77-4512-8679-6f26d8d6ef28-00033.parquet") as fout:
    # First get the fragment
    fragment = parquet_format.make_fragment(fout, None)
    print(f"Schema: {fragment.physical_schema}")
    arrow_table = ds.Scanner.from_fragment(
        fragment=fragment
    ).to_table()
```

I need the schema first, because it can be that a column got renamed, but the the file hasn't been rewritten against the latest schema. The same goes for filtering, if you change a column name, and the file still has the old name in there, then you would like to leverage the predicate pushdown of PyArrow to not load the data in memory at all.

When looking into the minio logs I can see that it does four requests.

1. A head to check if the file exists
2. The last 64kb from the Parquet file to get the schema
3. Another last 64kb from the parquet file to get the schema
4. A nice beefy 1978578kb request to fetch the data

Looking at the tests, we shouldn't fetch the footer twice:

```python
# with default discovery, no metadata loaded
with assert_opens([fragment.path]):
    fragment.ensure_complete_metadata()
assert fragment.row_groups == [0, 1]

[...]

PR #33974, opened 2023-02-01
GH-33973: [Python][Docs] Update documentation for Parquet filter keyword

### LINKED ISSUE #33973: [Python] Parquet `to_table` pythondoc is outdated

The filter states that it needs it in DNF format, but since the `use_legacy_dataset=False` is now the default, we can also push in an expression. If you supply a DNF format, it will be converted into an expression later on: 

https://github.com/apache/arrow/blob/master/python/pyarrow/parquet/core.py#L2414-L2416

### Component(s)

Python

Fokko:
### Rationale for this change

I wrote a converter from an arbitrary expression to DNF, but this was not needed after learning that it just accepts an expression now.

Closes #33973


westonpace:
I don't have a problem with this change but wonder if the root cause might be a little more fundamental.  I wonder if `ParquetDataset` itself should be deprecated.  The docs you probably want are [pyarrow.dataset.dataset](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.dataset.html#pyarrow.dataset.dataset) and [pyarrow.dataset.Dataset](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.Dataset.html#pyarrow.dataset.Dataset) (though filter would be provided by [pyarrow.dataset.Dataset.to_table](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.Dataset.html#pyarrow.dataset.Dataset.to_table) which unfortunately redirects its documentation to [pyarrow.dataset.Scanner.from_dataset](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.Scanner.html#pyarrow.dataset.Scanner.from_dataset) :cold_sweat: )

Fokko:
@westonpace 

> The docs you probably want are [pyarrow.dataset.dataset](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.dataset.html#pyarrow.dataset.dataset) and [pyarrow.dataset.Dataset](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.Dataset.html#pyarrow.dataset.Dataset)

I would agree with you, but I went for a lower API because I want to re-use the connection. More background is provided [here](https://github.com/apache/arrow/issues/33972), and the dataset only accepts paths. If you think it is worthwhile to accept `NativeFile` there as well, let me know, and happy to raise a PR.

> which unfortunately redirects its documentation to [pyarrow.dataset.Scanner.from_dataset](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.Scanner.html#pyarrow.dataset.Scanner.from_dataset) 😰 )

Do you want me to create a PR to copy those docs? Because of the redirect, the arguments are also not showing up in PyCharm. Let me know and I'll create a PR.

westonpace:
> Do you want me to create a PR to copy those docs? Because of the redirect, the arguments are also not showing up in PyCharm. Let me know and I'll create a PR.

Yes, I think that is a good idea but I might CC @jorisvandenbossche or @amol- to weigh in on whether they know some better way to avoid the duplication or have a preference here (I normally focus on the C++ end of things.)

jorisvandenbossche:
@Fokko thanks for the catch! You are currently updating the docstring of ParquetDataset, but we should do the same update for read_table:

https://github.com/apache/arrow/blob/78a8da42d1fc364e84286afc88a25fa5b819d092/python/pyarrow/parquet/core.py#L2775

(I think it's also `read_table` you are using in PyIceberg, and not `ParquetDataset`?)




> > Do you want me to create a PR to copy those docs? Because of the redirect, the arguments are also not showing up in PyCharm. Let me know and I'll create a PR.
> 
> Yes, I think that is a good idea but I might CC @jorisvandenbossche or @amol- to weigh in on whether they know some better way to avoid the duplication or have a preference here (I normally focus on the C++ end of things.)

Yeah, I think we should prefer some duplication if that gives better docstrings. I agree the indirection for the user right now isn't very user friendly. 
We might be able to share some part of the docstring and inject that in multiple places to avoid duplicating the actual content, if that doesn't make things too complicated.

Fokko:
@jorisvandenbossche Updating table one is a good suggestion indeed. Updated that one as well 👍🏻 

> (I think it's also read_table you are using in PyIceberg, and not ParquetDataset?)

Currently, [we use the `read_table` indeed](https://github.com/apache/iceberg/blob/master/python/pyiceberg/io/pyarrow.py#L510-L517). I've also played around with the ParquetDataset, and it looked very similar. However, we don't need the lazy nature of the dataset, so directly loading a table makes more sense in our situation.

> Yeah, I think we should prefer some duplication if that gives better docstrings. I agree the indirection for the user right now isn't very user friendly.
> We might be able to share some part of the docstring and inject that in multiple places to avoid duplicating the actual content, if that doesn't make things too complicated.

Makes a lot of sense, I think sharing would be best. Let me create a separate PR for that

Fokko:
@jorisvandenbossche first a bit of cleanup in https://github.com/apache/arrow/pull/34034

jorisvandenbossche:
One more thing about the docstring: for both read_table and ParquetDataset, the docstring also injects the content of `_DNF_filter_doc` (which holds the actual explanation of how the filter is expressed in this list of tuples form). We should probably update that to start with saying that the filter can either be a pyarrow Expression, or this DNF list of tuples form.

Fokko:
@jorisvandenbossche Good call. I've updated the docstring and added an example.

Fokko:
@jorisvandenbossche WDYT?

jorisvandenbossche (approved):
Thanks!

PR #33978, opened 2023-02-01
GH-33970: [C#] Make schema field names case sensitive

### LINKED ISSUE #33970: [C#] Column names should be case sensitive

In the Schema constructor in [Schema.cs](https://github.com/apache/arrow/blob/master/csharp/src/Apache.Arrow/Schema.cs), the casing of field names is ignored when constructing the `_fieldsDictionary` (by using `StringComparer.OrdinalIgnoreCase` when constructing the dictionary).

This disallows multiple fields whose name only differs in casing, even though it should be allowed.

### Component(s)

C#

DanTm99:
Remove the use of a comparer that ignores casing when building the fields dictionary for a schema. 

This allows for fields with names that differ only in casing.

Currently, fields with names that only differ in casing will be seen as having the same name and will cause an Exception to be thrown due to being seen as having duplicate keys.
* Closes: #33970


westonpace (approved):
It seems this has been there from the very beginning so I could be convinced it was a mistake.  It's odd to me that a dictionary is used at all.  For example, it seems this will still prevent correct loading of tables that have duplicate column names (which is legal in at least some other implementations).

@eerhardt any opinions?

eerhardt:
Thanks for the contribution, @DanTm99! The code change looks good.

Any chance you could write a quick unit test or two for this?

eerhardt:
> It's odd to me that a dictionary is used at all. For example, it seems this will still prevent correct loading of tables that have duplicate column names 

How do duplicate column names work? If you look it up by name, you get the last one? But each gets an index?

So if I had 2 columns named "Foo", the Count would be `2`, and index 0 would be the first column. Index 1 would be the second column. And if I said `schema.GetFieldByName("Foo")` I'd get the 2nd column?

westonpace:
> How do duplicate column names work? If you look it up by name, you get the last one? But each gets an index?

Yes, each column gets an index and index-lookups always work.  In pyarrow a name lookup raises an error if there are duplicate columns.  In arrow-rs it appears it always picks the first one.

DanTm99:
@eerhardt I've added a few unit tests

eerhardt (approved):
LGTM. Thanks!

CI failure is unrelated.

DanTm99:
@westonpace @eerhardt I've created issue #34076 regarding the exception being thrown when multiple fields have the same name

PR #33983, opened 2023-02-01
MINOR: [Go] Use shared cache in SQLite example.

brettbuddin:
### Rationale for this change

When calling this server with a few concurrent requests, you can easily force the Go stdlib to spawn a new connection. Since only the `:memory:` DSN is specified, the stdlib doesn't know any better and effectively creates a new database. The result is a lot of `no such table` errors.

This change adds the `?cache=shared` as specified at:

https://www.sqlite.org/sharedcache.html#enabling_shared_cache_mode

This ensures that these new connections participate in same in-memory database and can see all the tables created at server start.

### Are these changes tested?

I haven't written any additional tests here, because this behavior is difficult to test. Suggestions are welcome on an approach here.

No.

* Closes: #33983

### DIFF
--- a/go/arrow/flight/flightsql/example/cmd/sqlite_flightsql_server/main.go
+++ b/go/arrow/flight/flightsql/example/cmd/sqlite_flightsql_server/main.go
@@ -40,7 +40,13 @@ func main() {
 
 	flag.Parse()
 
-	srv, err := example.NewSQLiteFlightSQLServer()
+	db, err := example.CreateDB()
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer db.Close()
+
+	srv, err := example.NewSQLiteFlightSQLServer(db)
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/go/arrow/flight/flightsql/example/sqlite_server.go
+++ b/go/arrow/flight/flightsql/example/sqlite_server.go
@@ -140,19 +140,7 @@ func prepareQueryForGetKeys(filter string) string {
 		` ORDER BY pk_catalog_name, pk_schema_name, pk_table_name, pk_key_name, key_sequence`
 }
 
-type Statement struct {
-	stmt   *sql.Stmt
-	params [][]interface{}
-}
-
-type SQLiteFlightSQLServer struct {
-	flightsql.BaseServer
-	db *sql.DB
-
-	prepared sync.Map
-}
-
-func NewSQLiteFlightSQLServer() (*SQLiteFlightSQLServer, error) {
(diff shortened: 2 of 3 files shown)


zeroshade:
@lidavidm can you think of a good way to test this properly?

lidavidm:
Hmm, you can't easily force Go to open two connections?

Possibly have one client start a transaction (with just a BEGIN TRANSACTION) and hopefully that ties up a connection in the pool, forcing the next request to use a new connection?

brettbuddin:
It's possible this is causing some failures in CI. I'll have to go through it and see the impact.

zeroshade:
@brettbuddin looks like you're getting "table already exists" errors in the CI. I'm guessing because I put the table creation into the startup of the server so when it goes to run the next test the tables already exist because of the cached option.

You could potentially put the table creation into the test itself via `SetupSuite` and then have the tests run in parallel?

brettbuddin:
@zeroshade I took a crack at it. I realized that the server shutdown is never closing the SQLite database so even though we're creating a new server structure and overwriting `srv` in the tests, it will continue to think its in the same memory space. We'd have the same problem if we simply pushed it down into the tests without closure.

I'm not exactly happy with the shape of what I have though. It seems like the owning Flight server should be able to ensure this is called without me manually calling it the teardown. function. However, there isn't a suitable hook point without being too invasive in `server.go` and friends. The `Shutdown()` I've added feels tacked on, because it is.

Either `db.Close()` will need to be called or a new "memory space" will need to be created explicitly for each server start-up via something like `file:<random-id-for-server>?mode=memory&cache=shared` in dialing. My vote is for proper closure, because it's less articulation. However, if concurrent execution of test cases (like you mentioned) is a priority, the new memory space idea should be considered.

zeroshade:
@brettbuddin I agree with you, there should be a better hook point. Do you want to take a crack at being invasive and putting together a better hook point? If not, I think this solution is fine for now until we put together a better one.

Thoughts?

brettbuddin:
@zeroshade After working the problem a bit I wasn't able to come up with a solution where augmenting the Flight server gave me what I needed. To me, it's more clear to separate the creation of the database from the gRPC service—making a dependency relationship. In `main.go` this results in the ability to just `defer` closure for when the server is completely out of the way, and in tests we just articulate closure in the suite's test teardown.

The consequence is an added step in creation of the server. Given my experience with this example: the point is to provide a binary to use in integration testing. It's not so important to me how many steps it takes to create that gRPC service layer. Thoughts?

zeroshade (approved):
I like it, LGTM

PR #33991, opened 2023-02-01
updated JSON to be JSONL and improved performance

dhowden:
This is just for reference later.


dhowden:
Apologies, was supposed to be against my own fork.

PR #33992, opened 2023-02-02
GH-33993: [Java] Let OS assign port in tests while creating Flight server

### LINKED ISSUE #32954: [C++][Java][FlightRPC] Get rid of FlightTestUtil.getStartedServer etc.
Anything expecting to bind to a random port in CI is an antipattern and makes tests flaky. All tests should bind to port 0 and let the OS assign a port.

**Reporter**: [David Li](https://issues.apache.org/jira/browse/ARROW-17718) / @lidavidm

<sub>**Note**: *This issue was originally created as [ARROW-17718](https://issues.apache.org/jira/browse/ARROW-17718). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

### LINKED ISSUE #33993: [Java] Let OS assign port in FlightTestUtil#getStartedServer

Today, the flight server is getting started on a random port in the ranges 49152-54152. This is a cause for flaky tests, and a quick mitigation would be to use `java.net.ServerSocket` by passing in 0 as the port value, and letting the OS assign a port. A more comprehensive fix can be implemented as part of #32954.

### Component(s)

Java

rtadepalli:
### Rationale for this change

Sometimes tests are flaky because the current port selection algorithm selects a random port between the ranges 49152-54152. This change makes it such that the OS will assign the next free port without us searching for a port in this range.

This PR makes use of  `java.net.ServerSocket`'s ability to assign the next free port if 0 is passed as the port number.

### Are these changes tested?

If existing tests pass, there's confirmation that the Flight server is able to come up successfully, thereby no new tests should be needed.

No, there are no user facing changes.

* Closes: #32954
* Closes: #33993

### DIFF
--- a/java/flight/flight-core/src/test/java/org/apache/arrow/flight/FlightTestUtil.java
+++ b/java/flight/flight-core/src/test/java/org/apache/arrow/flight/FlightTestUtil.java
@@ -52,9 +52,12 @@ public class FlightTestUtil {
     IOException lastThrown = null;
     T server = null;
     for (int x = 0; x < 3; x++) {
-      final ServerSocket dynamicPortSocket = new ServerSocket(0);
-      final int port = dynamicPortSocket.getLocalPort();
-      dynamicPortSocket.close();
+      int port;
+      try (final ServerSocket dynamicPortSocket = new ServerSocket(0)) {
+        port = dynamicPortSocket.getLocalPort();
+      } catch (SecurityException | IOException e) {
+        throw new RuntimeException("Unable to create a ServerSocket instance. ", e);
+      }
 
       final Location location = Location.forGrpcInsecure(LOCALHOST, port);
       lastThrown = null;
--- a/java/flight/flight-core/src/test/java/org/apache/arrow/flight/FlightTestUtil.java
+++ b/java/flight/flight-core/src/test/java/org/apache/arrow/flight/FlightTestUtil.java
@@ -20,6 +20,7 @@ package org.apache.arrow.flight;
 import java.io.File;
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
+import java.net.ServerSocket;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


java/flight/flight-core/src/test/java/org/apache/arrow/flight/FlightTestUtil.java
```diff
@@ -51,7 +52,10 @@ public static <T> T getStartedServer(Function<Location, T> newServerFromLocation
     IOException lastThrown = null;
     T server = null;
     for (int x = 0; x < 3; x++) {
```
rtadepalli, line 54: Kept the retry in case the port gets before the lambda is applied and the server is created.

lidavidm:
Thank you! That said, the issue is actually about getting rid of this antipattern entirely - we can and should bind to port 0 directly, and create the client by reading the assigned port off of the server. (That's why the issue is about getting rid of the antipattern-using test utilities).

This is an improvement though, so we can just file a new issue for this?

rtadepalli:
Yeah I was thinking that this is somewhat of a dumb fix to what you mentioned in the ticket. Sure, I can file a new ticket for this.

rtadepalli:
I have created: https://github.com/apache/arrow/issues/33993

Will amend my commit to include the new issue number.

rtadepalli:
@lidavidm I'd like to work on the original ticket as well. Hopefully I'll have something up soon. Sorry about the confusion here.

lidavidm:
No worries, as mentioned this is still better than it was before.

java/flight/flight-core/src/test/java/org/apache/arrow/flight/FlightTestUtil.java
```diff
@@ -51,7 +52,10 @@ public static <T> T getStartedServer(Function<Location, T> newServerFromLocation
...
-      final int port = 49152 + RANDOM.nextInt(5000);
+      final ServerSocket dynamicPortSocket = new ServerSocket(0);
```
lidavidm, line 55: nit: maybe try-with-resources this just to be safe?

lidavidm (approved):
Thanks again!

I'll merge this in the morning.

PR #33994, opened 2023-02-02
GH-33971: [C++] Fix AdaptiveIntBuilder to always populate data buffer

### LINKED ISSUE #33971: [Python] segfault when calling nbytes on empty table with dictionary field

Using `pyarrow.Table.nbytes` on an empty table with a dictionary field causes a segmentation fault.

```python
import pyarrow as pa

schema = pa.schema(
    [
        pa.field("foo", pa.dictionary(pa.int32(), pa.string())),
    ]
)
table = pa.table({"foo": []}, schema=schema)

print(table.nbytes)  # segmentation fault  python
```

A few notes:
- `get_total_buffer_size` works (which is a good workaround for now)
- It works if the table has got one or more row (including nulls)
- It works if the table is empty but there are not dictionary fields
- I'm using `pyarrow==11.0.0` and Python 3.9.16 (but it is happening in `10.0.1` as well, so not a new bug)

### Component(s)

Python

westonpace:
If the AdaptiveIntBuilder was empty it would yield a null for the values buffer.  The byte_size.h utilities were not expecting this which led to the error reported in the issue.

Example:

```
>>> pa.array([], type=pa.dictionary(pa.int32(), pa.string())).buffers()
[None, None]
```
* Closes: #33971

### DIFF
--- a/cpp/src/arrow/array/array_dict_test.cc
+++ b/cpp/src/arrow/array/array_dict_test.cc
@@ -157,6 +157,18 @@ TYPED_TEST(TestDictionaryBuilder, MakeBuilder) {
   AssertArraysEqual(expected, *result);
 }
 
+TYPED_TEST(TestDictionaryBuilder, Empty) {
+  DictionaryBuilder<TypeParam> dictionary_builder;
+  ASSERT_OK_AND_ASSIGN(std::shared_ptr<Array> empty_arr, dictionary_builder.Finish());
+  std::shared_ptr<DictionaryArray> empty_dict_arr =
+      std::static_pointer_cast<DictionaryArray>(empty_arr);
+  // Ensure that the indices value buffer is initialized
+  ASSERT_NE(nullptr, empty_dict_arr->data()->buffers[1]);
+  // Ensure that the dictionary's value buffer is initialized
+  ASSERT_NE(nullptr, empty_dict_arr->dictionary());
+  ASSERT_NE(nullptr, empty_dict_arr->dictionary()->data()->buffers[1]);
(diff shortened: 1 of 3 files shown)


cpp/src/arrow/array/array_dict_test.cc
```diff
@@ -157,6 +157,18 @@ TYPED_TEST(TestDictionaryBuilder, MakeBuilder) {
   AssertArraysEqual(expected, *result);
 }
 
+TYPED_TEST(TestDictionaryBuilder, Empty) {
+  DictionaryBuilder<TypeParam> dictionary_builder;
+  ASSERT_OK_AND_ASSIGN(std::shared_ptr<Array> empty_arr, dictionary_builder.Finish());
+  std::shared_ptr<DictionaryArray> empty_dict_arr =
+      std::static_pointer_cast<DictionaryArray>(empty_arr);
```
pitrou, line 164: `checked_pointer_cast` here?

pitrou (changes requested):
LGTM on the principle, just a nit.

cpp/src/arrow/array/array_dict_test.cc
westonpace: Fixed in https://github.com/apache/arrow/pull/33994/commits/c2d0a3631d157698fb9cc78ae1bb8db852b92085#diff-b6db8f2d68f397dbb43cd18c35eaf00d28fea796dd870d7019ceafcf04eda7e7R164

lidavidm (approved):

PR #34000, opened 2023-02-02
MINOR: [C++] Fix compilation error in hash_join_benchmark

js8544:
### Rationale for this change

[`HashJoinImpl::Init`](https://github.com/apache/arrow/blob/master/cpp/src/arrow/compute/exec/hash_join.h#L51) expects `output_batch_callback` and `finished_callback` to return `Status`, while the currently passed arguments return void. It caused compilation errors on my machine. 

```
error: cannot convert ‘arrow::compute::JoinBenchmark::JoinBenchmark(arrow::compute::BenchmarkSettings&)::<lambda(int64_t, arrow::compute::ExecBatch)>’ to ‘arrow::compute::HashJoinImpl::OutputBatchCallback’ {aka ‘std::function<arrow::Status(long int, arrow::compute::ExecBatch)>’}
```

Fix the return types.

### Are these changes tested?

Yes. It fixes the compilation error on my machine.

No.

### DIFF
--- a/cpp/src/arrow/compute/exec/hash_join_benchmark.cc
+++ b/cpp/src/arrow/compute/exec/hash_join_benchmark.cc
@@ -160,7 +160,8 @@ class JoinBenchmark {
         &ctx_, settings.join_type, settings.num_threads, &(schema_mgr_->proj_maps[0]),
         &(schema_mgr_->proj_maps[1]), std::move(key_cmp), std::move(filter),
         std::move(register_task_group_callback), std::move(start_task_group_callback),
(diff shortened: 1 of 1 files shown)


js8544:
The expected return type was changed in [this PR](cpp/src/arrow/compute/exec/hash_join.h). Not sure why our CI process doesn't catch this compile error.

kou (approved):
+1

We need additional `-DARROW_BUILD_OPENMP_BENCHMARKS=ON` option to build this benchmark. Our CI jobs don't cover it.

PR #34002, opened 2023-02-02
GH-34001: [C++] Add more rvalue overload for Status::operator&

### LINKED ISSUE #34001: [C++] rvalue lhs cannot be moved into returned value in Status::operator&

https://github.com/apache/arrow/blob/a0d0ecee71b99b45f2c1b269c70033d39b982d00/cpp/src/arrow/status.h#L160-L161

Currently there are two overloads for `Status::operator&` to achieve "reference forwarding", but only for the rhs (i.e. `const Status& s` or `Status&& s`), the reference type of lhs value (i.e. `*this`) cannot be forwarded. e.g.
```c++
Status v;
Status f(/* some parameters */);
Status g(/* some parameters */);

auto v1 = v & f(); // OK! the rhs rvalue can be moved into result
auto v2 = f() & v; // Ooops, the lhs rvalue cannot be moved into result
auto v3 = f() & g(); // Ooops, same as above
```

We can add more overloads to forward reference of lhs for better performance.

### Component(s)

C++

PragmaTwice:
It fixes #34001

-----

From #34001:

> https://github.com/apache/arrow/blob/a0d0ecee71b99b45f2c1b269c70033d39b982d00/cpp/src/arrow/status.h#L160-L161
>
> Currently there are two overloads for `Status::operator&` to achieve "reference forwarding", but only for the rhs (i.e. `const > Status& s` or `Status&& s`), the reference type of lhs value (i.e. `*this`) cannot be forwarded. e.g.
> ```c++
> Status v = ...;
> Status f(/* some parameters */);
> Status g(/* some parameters */);
>
> auto v1 = v & f(); // OK! the rhs rvalue can be moved into result
> auto v2 = f() & v; // Ooops, the lhs rvalue cannot be moved into result
> auto v3 = f() & g(); // Ooops, same as above
> ```
>
> We can add more overloads to forward reference of lhs for better performance.
* Closes: #34001

### DIFF
--- a/cpp/src/arrow/status.h
+++ b/cpp/src/arrow/status.h
@@ -157,8 +157,10 @@ class ARROW_EXPORT [[nodiscard]] Status : public util::EqualityComparable<Status
   inline bool Equals(const Status& s) const;
 
   // AND the statuses.
-  inline Status operator&(const Status& s) const noexcept;
-  inline Status operator&(Status&& s) const noexcept;
+  inline Status operator&(const Status& s) const& noexcept;
+  inline Status operator&(const Status& s) && noexcept;
+  inline Status operator&(Status&& s) const& noexcept;
+  inline Status operator&(Status&& s) && noexcept;
   inline Status& operator&=(const Status& s) noexcept;
   inline Status& operator&=(Status&& s) noexcept;
 
@@ -422,7 +424,7 @@ bool Status::Equals(const Status& s) const {
 /// \cond FALSE
 // (note: emits warnings on Doxygen < 1.8.15,
 //  see https://github.com/doxygen/doxygen/issues/6295)
-Status Status::operator&(const Status& s) const noexcept {
+Status Status::operator&(const Status& s) const& noexcept {
   if (ok()) {
     return s;
   } else {
@@ -430,11 +432,27 @@ Status Status::operator&(const Status& s) const noexcept {
   }
 }
 
-Status Status::operator&(Status&& s) const noexcept {
+Status Status::operator&(const Status& s) && noexcept {
+  if (ok()) {
+    return s;
+  } else {
+    return std::move(*this);
+  }
+}
+
+Status Status::operator&(Status&& s) const& noexcept {
   if (ok()) {
     return std::move(s);
   } else {
     return *this;
+  }
+}
+
+Status Status::operator&(Status&& s) && noexcept {
+  if (ok()) {
+    return std::move(s);
+  } else {
+    return std::move(*this);
   }
 }
 
--- a/cpp/src/arrow/status_test.cc
+++ b/cpp/src/arrow/status_test.cc
@@ -104,6 +104,21 @@ TEST(StatusTest, AndStatus) {
   res = Status::Invalid("foo") & Status::IOError("bar");
   ASSERT_TRUE(res.IsInvalid());
 
+  ASSERT_TRUE((a & Status::OK()).ok());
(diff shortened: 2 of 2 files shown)


westonpace:
The change seems fine but after the discussion in https://github.com/apache/arrow/pull/13304 I'm not sure I'd advise anyone to use `f() & g()` as there is no way to guarantee short-circuiting and many readers will assume that is what is happening.

mapleFU:
I think it's ok, but I wonder if it would introduce any public api break? If not, glad to use it

PragmaTwice:
@pitrou Hi, could you take a look at this PR? So thanks!

cpp/src/arrow/status.h
```diff
@@ -157,8 +157,10 @@ class ARROW_EXPORT [[nodiscard]] Status : public util::EqualityComparable<Status
   inline bool Equals(const Status& s) const;
 
   // AND the statuses.
-  inline Status operator&(const Status& s) const noexcept;
-  inline Status operator&(Status&& s) const noexcept;
+  inline Status operator&(const Status& s) const& noexcept;
+  inline Status operator&(const Status& s) && noexcept;
+  inline Status operator&(Status&& s) const& noexcept;
+  inline Status operator&(Status&& s) && noexcept;
```
cyb70289, line 163: Looks to me "pass by value" is simpler. To reduce half of the functions. There might be overhead of an additional move, should be trivial.
```
Status operator&(Status s) const&;
Status operator&(Status s) &&;
```

cpcloud:
Is this PR still viable? If not, please close out.

amol-:
@westonpace is this a PR that we can merge? The addition seems helpful with minor side effects

westonpace:
I don't see any response to https://github.com/apache/arrow/pull/34002#discussion_r1099944171  I'm fine with this being merged.  It's not really high on my priority list.  At the very least it needs a rebase though.

PragmaTwice:
It appears that there are only a few individuals in the Arrow community who have good knowledge of C++.

To begin with, I am confident that my code is accurate. Comparable code can be found in the standard library, such as `std::optional<T>::value_or`. In future versions of C++ (23 or later), we may even use *deducing this* to apply universal forwarding on `*this` and prevent redundant code.

Furthermore, I am uncertain why method functions within class scope necessitate an `inline` specifier. This is implicit in the C++ standard and does not require explicit declaration.

cpp/src/arrow/status.h
PragmaTwice: I do not understand what you mean by the "simpler" that you mentioned. 

It should be like this in c++23 or later:

```c++
template <typename Self, typename T, std::enable_if_t<std::is_same_v<std::remove_cvref_t<T>, Status>, int> = 0>
Status operator&(this Self&& self, T&& other) {
     if (ok()) return std::forward<T>(other);
    return std::forward<Self>(self);
}
```

(hand-written, you can check it by compiler and lang-server)

mapleFU:
@PragmaTwice It's `value_or` like code here: https://github.com/llvm-mirror/libcxx/blob/master/include/optional#L132-L133 ? And you use explicit ctor using `(const Status&)` and `Status&&` to represent universal reference? I think both your change and https://github.com/apache/arrow/pull/34002#discussion_r1099944171 is ok.

westonpace:
> I am currently not interested in advancing this PR.

Fair enough.  I will close this.  Should someone else need it or should you change your mind you are always welcome to open a new PR with this change.

PR #34008, opened 2023-02-02
GH-34007: [C++] Add an array_span_mutable interface to ExecResult

### LINKED ISSUE #34007: [C++] Add an array_span_mutable interface to ExecResult

This is a child issue of #33628.

Like for Buffer, we separate two interface for getting the underlying `array_span` for `ExecResult`, one const and one mutable. This avoids the awkward `const_cast` we currently use.

### Component(s)

C++

js8544:
### Rationale for this change

Like for Buffer, we separate two interface for getting the underlying `array_span` for `ExecResult`, one const and one mutable. This avoids the awkward `const_cast` we currently use.

Add the new mutable interface and update current calls to this method.

### Are these changes tested?

Tested with existing tests.

No.
* Closes: #34007

### DIFF
--- a/cpp/examples/arrow/udf_example.cc
+++ b/cpp/examples/arrow/udf_example.cc
@@ -56,7 +56,7 @@ arrow::Status SampleFunction(cp::KernelContext* ctx, const cp::ExecSpan& batch,
   const int64_t* x = batch[0].array.GetValues<int64_t>(1);
   const int64_t* y = batch[1].array.GetValues<int64_t>(1);
   const int64_t* z = batch[2].array.GetValues<int64_t>(1);
-  auto* out_values = out->array_span_mutable()->GetValues<int64_t>(1);
+  int64_t* out_values = out->array_span_mutable()->GetValues<int64_t>(1);
   for (int64_t i = 0; i < batch.length; ++i) {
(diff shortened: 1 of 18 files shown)


westonpace:
+1 on the idea of `array_span_mutable`.

I'm -0.5 on rewriting everything to auto.  Arrow follows the [google C++ style guide](https://google.github.io/styleguide/cppguide.html#Type_deduction) which states:

> The fundamental rule is: use type deduction only to make the code clearer or safer, and do not use it merely to avoid the inconvenience of writing an explicit type. When judging whether the code is clearer, keep in mind that your readers are not necessarily on your team, or familiar with your project, so types that you and your reviewer experience as unnecessary clutter will very often provide useful information to others. For example, you can assume that the return type of make_unique<Foo>() is obvious, but the return type of MyWidgetFactory() probably isn't.

Though that leaves plenty of room for interpretation.  We also have not applied the rule consistently and large chunks of the code base use auto extensively.

However, I would generally discourage rewriting something to auto just because you can.

js8544:
> I would generally discourage rewriting something to auto just because you can.
Makes sense to me. I've reverted the changes to `auto`.

westonpace (approved):
Thanks for this addition!

PR #34009, opened 2023-02-02
GH-33904: [R] improve behavior of s3_bucket - work-around

### LINKED ISSUE #33904: [R] Using s3_bucket with non-AWS S3-compatible storage is confusing

cboettig:
I think this is another option to address #33918.

This retains the existing behavior when no additional arguments (region, endpoint, etc) are supplied.  

If any optional arguments are supplied, this approach will by-pass the `$from_uri` call.  I think this is reasonable, given that the `from_uri()` call is implicitly ignoring optional arguments already, which is giving rise to the current bugs.  

* Closes: #33904

### DIFF
--- a/r/R/filesystem.R
+++ b/r/R/filesystem.R
@@ -462,8 +462,8 @@ s3_bucket <- function(bucket, ...) {
   if (!is_url(bucket)) {
     bucket <- paste0("s3://", bucket)
   }
-  
-  if (!length(args)) {  
+
+  if (!length(args)) {
     fs_and_path <- FileSystem$from_uri(bucket)
     fs <- fs_and_path$fs
   } else {
--- a/r/R/filesystem.R
+++ b/r/R/filesystem.R
@@ -458,19 +458,20 @@ s3_bucket <- function(bucket, ...) {
   assert_that(is.string(bucket))
   args <- list2(...)
 
-  # Use FileSystemFromUri to detect the bucket's region
-  if (!is_url(bucket)) {
-    bucket <- paste0("s3://", bucket)
-  }
+  # If user specifies args, they must specify region as arg, env var, or config
+  if (length(args) == 0) {
+    # Use FileSystemFromUri to detect the bucket's region
+    if (!is_url(bucket)) {
+      bucket <- paste0("s3://", bucket)
+    }
 
-  if (!length(args)) {
     fs_and_path <- FileSystem$from_uri(bucket)
     fs <- fs_and_path$fs
   } else {
-  # If there are no additional S3Options, we can use that filesystem
-  # If user specifies args, they must specify region as arg, env var, or config
+    # If there are no additional S3Options, we can use that filesystem
     fs <- exec(S3FileSystem$create, !!!args)
   }
+
   # Return a subtree pointing at that bucket path
   SubTreeFileSystem$create(bucket, fs)
 }
--- a/r/R/filesystem.R
+++ b/r/R/filesystem.R
@@ -462,12 +462,13 @@ s3_bucket <- function(bucket, ...) {
   if (!is_url(bucket)) {
     bucket <- paste0("s3://", bucket)
   }
-  fs_and_path <- FileSystem$from_uri(bucket)
-  fs <- fs_and_path$fs
+  
+  if (!length(args)) {  
+    fs_and_path <- FileSystem$from_uri(bucket)
+    fs <- fs_and_path$fs
+  } else {
   # If there are no additional S3Options, we can use that filesystem
-  # Otherwise, take the region that was detected and make a new fs with the args
-  if (length(args)) {
-    args$region <- fs$region
+  # If user specifies args, they must specify region as arg, env var, or config
     fs <- exec(S3FileSystem$create, !!!args)
   }
   # Return a subtree pointing at that bucket path
--- a/r/R/filesystem.R
+++ b/r/R/filesystem.R
@@ -472,7 +472,7 @@ s3_bucket <- function(bucket, ...) {
     fs <- exec(S3FileSystem$create, !!!args)
   }
   # Return a subtree pointing at that bucket path
-  SubTreeFileSystem$create(fs_and_path$path, fs)
+  SubTreeFileSystem$create(bucket, fs)
 }
 
 #' Connect to a Google Cloud Storage (GCS) bucket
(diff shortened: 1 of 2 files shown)


cboettig:
I think this work-around is much preferable to a warning for now.  If you introduce a warning that says `endpoint_override` is not supported by s3_bucket, I have a good number of users across a few different packages who will be confused by the warning, because it works just fine as long as the bucket name is not unique.  (and yes, users on custom endpoints will often pick common bucket names already in use somewhere on AWS, since they have no reason to need unique names).

paleolimbot (commented):
This change looks reasonable...thank you for making the change!

Could you add a test that makes sure that at least one parameter is not ignored? I'm not sure what the best call to `s3_bucket()` is that will both succeed and not ignore a parameter...perhaps that's best done on the `test-minio.R` (make sure we can connect to the minio server using `s3_bucket(..., endpoint_override = "xxx")`?

r/R/filesystem.R
```diff
@@ -462,12 +462,13 @@ s3_bucket <- function(bucket, ...) {
...
+  if (!length(args)) {
+    fs_and_path <- FileSystem$from_uri(bucket)
+    fs <- fs_and_path$fs
+  } else {
   # If there are no additional S3Options, we can use that filesystem
-  # Otherwise, take the region that was detected and make a new fs with the args
-  if (length(args)) {
-    args$region <- fs$region
+  # If user specifies args, they must specify region as arg, env var, or config
     fs <- exec(S3FileSystem$create, !!!args)
```
paleolimbot, line 472: Looks like you need to define `fs_and_path` here somehow?

cboettig:
@paleolimbot ah simple, `fs_and_path$path` is just the `bucket`, should be fixed now.  

Also checked locally this time (apologies for not doing so before, had to fiddle a bit to get `devtools::load_all()` to compile with S3 support).  Seems to be working on my end.

One of the CI tasks seems to have timed out, not sure precisely but doesn't look like a related issue?

thisisnic:
CI timeout will be unrelated to this PR as we've seen it do that on another one; have manually re-triggered it for you.

r/tests/testthat/test-s3-minio.R
```diff
@@ -61,6 +61,13 @@ fs$CreateDir(now)
 # Clean up when we're all done
 withr::defer(fs$DeleteDir(now))
 
+test_that("Confirm s3_bucket works with endpoint_override", {
+  
+  
+  bucket <- s3_bucket("test", endpoint_override = "https://play.min.io")
```
paleolimbot, line 67: Was something wrong with the localhost version?

cboettig: @paleolimbot good question, that's something I'm trying to figure out too.  It failed to find the bucket in the previous test, https://github.com/apache/arrow/actions/runs/4093029707/jobs/7073398005, but I don't think that's because of any change I made -- in my own local testing I could never reproduce that error.  I see the desire to use localhost, but it's not quite clear how to ensure the localhost deploy works before running the test.

It's not so clear to me that any of the other tests rely on the localhost bucket existing?  

https://play.min.io is the official test service for the MINIO platform, not sure if that's ideal choice here for the tests, but then the test suite here isn't really built out and my objective is just submit a patch for this confirmed bug in the codebase.

cboettig:
It looks like things are passing now, except for again a stalled checks that don't seem related to the pull request(?)

r/tests/testthat/test-s3-minio.R
paleolimbot: Hmm...the localhost version should definitely work. I will do a local pull tomorrow morning and see what I can come up with 🙂

cboettig:
ah, good catch!  

Good to know that the URI format requires all optional arguments to be included in the URI and aren't read from config or manual options even when using `$create()`, that's a bit surprising to me but guess it makes sense.  (My play.minio example must have passed because again it used a bucket name available on both AWS and play.minio endpoints! whoops!)

Yup, all looks good now to me as well and works as expected.  Thanks much!

cboettig:
:tada:

PR #34013, opened 2023-02-02
GH-33960: [C++] Add DeclarationToSchema and DeclarationToString helper methods.

### LINKED ISSUE #33960: [R] Output schema for aggregation is sometimes innacurate

westonpace:
Also cleans up the Acero server example to use the DeclarationToXyz methods
* Closes: #33960

### DIFF
--- a/cpp/src/arrow/compute/exec/exec_plan.cc
+++ b/cpp/src/arrow/compute/exec/exec_plan.cc
@@ -863,12 +863,14 @@ Result<std::unique_ptr<RecordBatchReader>> DeclarationToReader(
     std::shared_ptr<Schema> schema() const override { return schema_; }
 
     Status ReadNext(std::shared_ptr<RecordBatch>* record_batch) override {
-      DCHECK(!!iterator_) << "call to ReadNext on already closed reader";
+      if (!iterator_) {
+        return Status::Invalid("call to ReadNext on already closed reader");
+      }
       return iterator_->Next().Value(record_batch);
     }
 
     Status Close() override {
-      if (!!iterator_) {
+      if (!iterator_) {
         // Already closed
         return Status::OK();
       }
--- a/cpp/src/arrow/compute/exec/exec_plan.cc
+++ b/cpp/src/arrow/compute/exec/exec_plan.cc
@@ -575,6 +575,52 @@ bool Declaration::IsValid(ExecFactoryRegistry* registry) const {
   return !this->factory_name.empty() && this->options != nullptr;
 }
 
+namespace {
+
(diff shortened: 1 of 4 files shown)


westonpace:
I'm not sure if this fully closes #33960 but it was inspired by it.

cpp/src/arrow/compute/exec/exec_plan.cc
```diff
@@ -822,6 +868,10 @@ Result<std::unique_ptr<RecordBatchReader>> DeclarationToReader(
     }
 
     Status Close() override {
+      if (!!iterator_) {
```
lidavidm, line 871: Should this just be `!iterator_`?

westonpace: :cold_sweat: yes.  Fixed.

lidavidm (approved):

ianmcook:
#33917 provides an immediate fix for #33960. I unlinked this from #33917 since it does not.

westonpace:
> https://github.com/apache/arrow/pull/33917 provides an immediate fix for https://github.com/apache/arrow/issues/33960. I unlinked this from https://github.com/apache/arrow/issues/33960 since it does not. Nope, the bots wouldn't let me do that without changing the PR title, I think.

I thought the merge script would allow me to merge this without updating the issue but it seems to have closed it anyway (I suppose GH does that automatically).  I reopened #33960 manually.

lidavidm:
In this case, the PR title and description still reference that issue and the sidebar notes that the issue is linked, so yeah, it seems GitHub did that

PR #34014, opened 2023-02-02
GH-33920: [C++][CI] Disable Flight SQL in sanitizer job

### LINKED ISSUE #33920: [C++][CI] C++ / AMD64 Ubuntu 22.04 C++ ASAN UBSAN (pull_request) Cancelled after 75m

**C++ / AMD64 Ubuntu 22.04 C++ ASAN UBSAN** frequently hits the timeout limit of 75 min. Does it make sense to increase the limit in the workflow file?

Examples:
https://github.com/apache/arrow/actions/runs/4035720668/jobs/6937808549
https://github.com/apache/arrow/actions/runs/4034613084/jobs/6935958187

Successful check also takes 1h 5m 4s on the current master:
https://github.com/apache/arrow/actions/runs/4038548251/jobs/6942621768

### Component(s)

Continuous Integration

kou:
### Rationale for this change

We want to reduce CI time.

This disables Flight SQL to use system gRPC and ProtoBuf.
We can reduce build time by avoiding bundled gRPC and ProtoBuf.

### Are these changes tested?

Yes.

No.
* Closes: #33920

### DIFF
--- a/cpp/cmake_modules/ThirdpartyToolchain.cmake
+++ b/cpp/cmake_modules/ThirdpartyToolchain.cmake
@@ -3997,6 +3997,13 @@ if(ARROW_WITH_GRPC)
     else()
       message(FATAL_ERROR "Cannot find grpc++ headers in ${GRPC_INCLUDE_DIR}")
     endif()
+    if(ARROW_USE_ASAN)
+      # Disable ASAN in system gRPC.
+      add_library(gRPC::grpc_asan_suppressed INTERFACE IMPORTED)
+      target_compile_definitions(gRPC::grpc_asan_suppressed
+                                 INTERFACE "GRPC_ASAN_SUPPRESSED")
+      target_link_libraries(gRPC::grpc++ INTERFACE gRPC::grpc_asan_suppressed)
+    endif()
   endif()
 endif()
 
--- a/cpp/cmake_modules/ThirdpartyToolchain.cmake
+++ b/cpp/cmake_modules/ThirdpartyToolchain.cmake
@@ -1649,8 +1649,8 @@ macro(build_protobuf)
 endmacro()
 
 if(ARROW_WITH_PROTOBUF)
-  if(ARROW_WITH_GRPC)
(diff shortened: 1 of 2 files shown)


westonpace (approved):

westonpace:
I ran this locally and does prevent building protobuf.  We still might also want to turn off GCS (although that can be a separate PR)

westonpace (changes requested):
Actually, I didn't notice, but one of the flight tests is failing.  Maybe flight (and not just flight SQL) require newer protobuf?

westonpace:
I spent quite some time today trying to debug this with no luck.  It is a very consistent failure in this configuration.  I can reproduce with docker.

From what I can tell it is including and linking the correct versions of protobuf and grpc.

gdb isn't terribly useful (grpc's symbols are stripped and so I can't inspect the rpc.context object).

js8544:
I can also reproduce this locally. Here's what I can find out with gdb:

1. The UB happens because `rpc.context.initial_metadata_corked_` is not initialized.
2. The constructor `ClientContext::ClientContext()` is called, but only a few member variables are initialized, `initial_metadata_corked_` is not one of them.
3.  This only happens when UBSAN and ASAN are on. In a regular build `rpc.context` is properly initialized.

js8544:
I found the root cause. GRPC's [`gpr_mu`](https://github.com/grpc/grpc/blob/master/include/grpc/support/sync_posix.h#L28) mutex has different structures depending on whether UBSAN is on. So we have to compile GRPC by ourselves if we want to enable UBSAN.
cc @kou @westonpace

kou:
Good catch!
I've added `-DGRPC_ASAN_SUPPRESSED`.

kou:
+1

Passed in 34min: https://github.com/apache/arrow/actions/runs/4092011740/jobs/7056458825

westonpace:
@js8544 great catch indeed!  Very impressive investigating.  Thanks for looking at it.

PR #34015, opened 2023-02-03
GH-33972: [C++] Pass in metadata to ParquetReader

### LINKED ISSUE #33972: [Python] Remove redundant S3 call

Fokko:
Closes #33972

* Closes: #33972

### DIFF
--- a/cpp/src/arrow/dataset/file_parquet.cc
+++ b/cpp/src/arrow/dataset/file_parquet.cc
@@ -347,11 +347,23 @@ Result<std::shared_ptr<Schema>> ParquetFileFormat::Inspect(
 
 Result<std::shared_ptr<parquet::arrow::FileReader>> ParquetFileFormat::GetReader(
     const FileSource& source, const std::shared_ptr<ScanOptions>& options) const {
-  return GetReaderAsync(source, options).result();
+  return GetReaderAsync(source, options, nullptr).result();
+}
+
+Result<std::shared_ptr<parquet::arrow::FileReader>> ParquetFileFormat::GetReader(
+    const FileSource& source, const std::shared_ptr<ScanOptions>& options,
+    const std::shared_ptr<parquet::FileMetaData>& metadata) const {
+  return GetReaderAsync(source, options, metadata).result();
 }
 
 Future<std::shared_ptr<parquet::arrow::FileReader>> ParquetFileFormat::GetReaderAsync(
     const FileSource& source, const std::shared_ptr<ScanOptions>& options) const {
+  return GetReaderAsync(source, options, nullptr);
+}
+
+Future<std::shared_ptr<parquet::arrow::FileReader>> ParquetFileFormat::GetReaderAsync(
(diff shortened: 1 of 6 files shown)


westonpace:
This metadata behavior is kind of subtle and I'm hoping we have a lot of changes coming to file_parquet.cc (page indices, bloom filters, etc.) so let's add a unit test.  I sent you a PR.

Fokko:
Thanks @westonpace for adding tests, much appreciated

westonpace (approved):

Fokko:
Thanks @westonpace for merging and the support in this issue 👍🏻

Issue #33091, opened 2022-09-28
[R][CI] Remove ubuntu-18.04 from nixlibs & prebuilt binaries


asfimport:
The new dts compiled centos-7 binaries ([ARROW-17594]) should be able to replace the ubuntu-18.04 binaries.

**Reporter**: [Jacob Wujciak](https://issues.apache.org/jira/browse/ARROW-17876) / @assignUser

<sub>**Note**: *This issue was originally created as [ARROW-17876](https://issues.apache.org/jira/browse/ARROW-17876). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

kou:
We can't use the `centos-7` binaries on Ubuntu 20.04 because the `centos-7` binaries uses OpenSSL 1.0 and Ubuntu 20.04 uses OpenSSL 1.1. OpenSSL 1.0 and 1.1 are incompatible.
See also: #34048

Issue #32292, opened 2022-07-05
[R] Build linux binaries on older image (like manylinux2014)


asfimport:
ARROW-16752 observed that even with newer compilers installed on centos 7, you can't use binaries built on ubuntu 18.04 because ubuntu 18.04 has glibc 2.27 but centos 7 only has 2.17. But if we built the binaries on centos 7 with devtoolset-7 or 8 or something, all features could compile and we'd work with older glibc. 

Things built against older glibc are guaranteed to work with newer versions, and you can't just upgrade glibc because it would break the system. So for maximum compatibility, build with the oldest glibc. This strategy is like how they python manylinux standards work (IIUC). 

**Reporter**: [Neal Richardson](https://issues.apache.org/jira/browse/ARROW-16976) / @nealrichardson
#### Related issues:
- [[R][Packaging] Build binaries with devtoolset 8 on CentOS 7](https://github.com/apache/arrow/issues/32842) (is fixed by)

<sub>**Note**: *This issue was originally created as [ARROW-16976](https://issues.apache.org/jira/browse/ARROW-16976). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

kou:
I tried to use CentOS 7 binaries with Ubuntu 18.04 in https://github.com/apache/arrow/pull/34020#issuecomment-1416603703 but we can't do it:

https://github.com/ursacomputing/crossbow/actions/runs/4089420508/jobs/7052165676#step:12:265

```text
 Error: package or namespace load failed for ‘arrow’ in dyn.load(file, DLLpath = DLLpath, ...):
 unable to load shared object '/usr/local/lib/R/site-library/00LOCK-arrow/00new/arrow/libs/arrow.so':
  /usr/local/lib/R/site-library/00LOCK-arrow/00new/arrow/libs/arrow.so: undefined symbol: _ZNK5arrow8DataType18ComputeFingerprintB5cxx11Ev
```

Because CentOS 7 + devtoolset doesn't use cxx11 ABI but Ubuntu 18.04 requires cxx11 ABI.

Can we close this?

PR #13470, opened 2022-06-30
ARROW-16764: [Packaging][deb] Drop support for Debian GNU/Linux buster

kou:
It will reach EOL at 2022-08.

See also: https://wiki.debian.org/DebianReleases

PR #34020, opened 2023-02-03
GH-33800: [Packaging] Drop support for Ubuntu 18.04

### LINKED ISSUE #33800: [Packaging] Drop support for Ubuntu 18.04

Ubuntu 18.04 will reach End of Standard Support on 2023-04: https://wiki.ubuntu.com/Releases

> Version | Code name | Docs | Release | End of Standard Support | End of Life
> -- | -- | -- | -- | -- | --
> Ubuntu 18.04.6 LTS | Bionic Beaver | Changes | September 17.2021 | April 2023 | April 2028

We'll release 12.0.0 on 2023-04 so 12.0.0 doesn't need Ubuntu 18.04 support. We can drop support for Ubuntu 18.04 support now because the maintenance branch for 11.0.0 is already created.

FYI: We can require CMake 3.16 or later after we drop support for Ubuntu 18.04 because Ubuntu 20.04 ships CMake 3.16 and EPEL for CentOS 7 ships CMake 3.17.

### Component(s)

Packaging

kou:
### Rationale for this change

Ubuntu 18.04 will reach End of Standard Support on 2023-04: https://wiki.ubuntu.com/Releases

> Version | Code name | Docs | Release | End of Standard Support | End of Life
> -- | -- | -- | -- | -- | --
> Ubuntu 18.04.6 LTS | Bionic Beaver | Changes | September 17.2021 | April 2023 | April 2028

Remove Ubuntu 18.04 related things.

### Are these changes tested?

Yes.

Yes.

* Closes: #33800


kou:
@github-actions crossbow submit -g nightly-tests -g nightly-packaging -g nightly-release

kou:

kou:

kou:

assignUser:
Please don't merge the changes to r-binary-packages yet as I am fairly sure we need that version due to compiler versions etc. and can't replace it with 20.04. But can't check `nixlibs.R` right now. Right @nealrichardson ?

kou:
Do we need to support R 3.6?
Do we need to support Debian buster?

If we can drop support for them, we can replace `ubuntu-18.04` binaries with binaries built on Ubuntu 20.04.

nealrichardson:
> Do we need to support R 3.6? 

Yes, but that shouldn't matter for the libarrow binaries we build.
 
> Do we need to support Debian buster?

The libarrow binaries we build are optional. As long as arrow compiles everywhere it should, it's ok if we don't provide binaries that work there, they're just an installation speedup.

> If we can drop support for them, we can replace `ubuntu-18.04` binaries with binaries built on Ubuntu 20.04.

#32292 and #33091 should let us switch to using binaries built on CentOS 7, IIUC, so this would be moot. We'll basically have two versions of libarrow binaries: one built against openssl 1 and the other against openssl 3.

I would prefer to leave the r-binary-packages jobs unmodified here and remove the ubuntu-18.04 from it by addressing those other issues.

kou:
OK. Then I work on #32292 and #33091 before this.

kou:
#34048 is for #32292 and #33091.

kou:
@github-actions crossbow submit -g nightly-release -g nightly-tests -g nightly-packaging

ci/docker/conda.dockerfile
```diff
@@ -16,7 +16,7 @@
 # under the License.
 
 ARG arch=amd64
-FROM ${arch}/ubuntu:18.04
+FROM ${arch}/ubuntu:22.04
```
assignUser, line 19: I don't mind using 22 but is there a reason we skip 20?

dev/tasks/linux-packages/apache-arrow-apt-source/apt/debian-buster/Dockerfile
assignUser, line 1: Buster has also reached EOL for a while right?

assignUser (approved):
LGTM :+1:

ci/docker/conda.dockerfile
kou: No.
It's just no-reason to use Ubuntu 20.04. :-)

dev/tasks/linux-packages/apache-arrow-apt-source/apt/debian-buster/Dockerfile
kou: Yes.
I should have removed this in #13470.

kou:
@github-actions crossbow submit test-conda-python-3.10

PR #34032, opened 2023-02-03
GH-29746: [C++][Python] Approximate count distinct kernel

jbapple:
This patch provides "hll", an abbreviation for HyperLogLog, which can approximately count the number of distinct elements in a stream (or set of streams) with less memory than exact count distinct.

To do this, we pull in Apache DataSketches - a vendored version, as the most recent released version has some issues that casue our CI to fail.

Changes to the cpp and python directories, as well as a full copy of https://github.com/apache/datasketches-cpp

### Are these changes tested?

Yes

Yes - the addition of the hll function. This PR contains an update to the docs.

* Closes: #29746


assignUser:
Thanks for the PR, sounds useful

. I personally am not a fan of vendoring entire repos if there is no need for it. In this case you seem to know a specific rev that works, so I would prefer to get the repo via FetchContent (as we are bumping cmake to 3.16 for 12.0.0)/EP and use the provided cmake code & config to integrate it into the build process. 

But I am no cpp maintainer so I will defer to other opinions.

westonpace:
I agree with @assignUser .  Plus, new repos bring in the question of code size, security, etc.  Probably we will also want a new `ARROW_WITH_DATASKETCHES` cmake option that controls inclusion (and if not included we just don't get those kernels and that's ok). That being said, the datasketches project looks pretty neat.  Do you have much experience with it?  Do you know if there would be other kernels beyond hll that are available?

westonpace:
Alternatively, maybe it's time to start pioneering the ability to write an extension that adds new compute functions?  E.g. DuckDb is heading in this direction (https://duckdb.org/docs/extensions/overview.html) and it seems like it would be a way to help keep the compute kernels more maintainable (both kernel maintainers not wanting to be affected by engine changes and vice versa).

jbapple:
> That being said, the datasketches project looks pretty neat. Do you have much experience with it? Do you know if there would be other kernels beyond hll that are available?

There are other useful kernels, including frequent items / heavy hitters, reservoir sampling, quantiles for non-numeric entities, and exterme-value quantiles (like p99.9).

https://datasketches.apache.org/docs/Architecture/MajorSketchFamilies.html

PR #34034, opened 2023-02-03
GH-30774: [Python] Remove deprecated `use_async`

Fokko:
Has been deprecated since Arrow 7.0.0, time to make things nice and clean:

https://issues.apache.org/jira/browse/ARROW-13554

* Closes: #30774


jorisvandenbossche (approved):
Thanks for the PR, looks good!

(I just linked it to another older issue, since we already had one open about it (that is linked to from the original PR))

PR #34036, opened 2023-02-03
GH-34035: [C++] Internal header file included from public one breaks build of external projects

### LINKED ISSUE #34035: [C++] Internal header file included from public one breaks build of external projects

There are a number of public header files that include internal header files:
```
$ find cpp/src/arrow/ -name "*.h" | xargs grep -l '_internal.h' | grep -v '_internal.h$'
cpp/src/arrow/compute/exec/swiss_join.h
cpp/src/arrow/compute/exec/util.h
cpp/src/arrow/compute/kernels/common.h
cpp/src/arrow/compute/kernels/row_encoder.h
cpp/src/arrow/dataset/test_util.h
```
This causes the build of external projects to break when including the public header because the internal one cannot be found since it is not part of the install.

Besides fixing the above public headers, it would be useful to add a (push-time?) check that ensures the above list is empty.

### Component(s)

C++

rtpsw:
See #34035
* Closes: #34035


cpp/src/arrow/compute/exec/util.h
```diff
@@ -57,6 +56,15 @@
...
 namespace arrow {
+
+namespace internal {
+namespace tracing {
+
+struct Scope;
+
+}  // namespace tracing
+}  // namespace internal
+
```
westonpace, line 67: Can we instead add `Scope` to `arrow/util/type_fwd.h` (and include that here)

cpp/src/arrow/compute/exec/util.h
```diff
@@ -234,6 +242,8 @@ class bit_util {
 }  // namespace util
 namespace compute {
 
+class ExecNode;
```
westonpace, line 245: Instead can we include `arrow/compute/type_fwd.h`?

cpp/src/arrow/compute/exec/util.h
```diff
@@ -36,7 +36,6 @@
 #include "arrow/util/logging.h"
 #include "arrow/util/mutex.h"
 #include "arrow/util/thread_pool.h"
-#include "arrow/util/tracing_internal.h"
```
westonpace, line 39: Note, this header is still an internal header as it includes `arrow/util/logging.h` which is perhaps our most egregious improperly named header.

westonpace (changes requested):
Thanks for taking this on.

cpp/src/arrow/compute/exec/util.h
rtpsw: Though this won't break an external project build, right?

cpp/src/arrow/compute/exec/util.h
rtpsw: Handling this in the next commit.

cpp/src/arrow/compute/exec/util.h
rtpsw: Looks like the header is already included and the forward declaration here can be removed.

cpp/src/arrow/compute/exec/util.h
westonpace: Ah, I don't know.  But we have a check that ensures we don't expose `logging.h` from `api.h`.  So we need to keep `map_node.h` out of `api.h` until this is resolved.

rtpsw: Done.

PR #34042, opened 2023-02-04
GH-33850: [C++] Allow Substrait's default extension provider to be configured

### LINKED ISSUE #33850: [C++] Allow Substrait's default extension provider to be configured

rtpsw:
See #33850
* Closes: #33850

### DIFF
--- a/cpp/src/arrow/engine/substrait/options.cc
+++ b/cpp/src/arrow/engine/substrait/options.cc
@@ -122,11 +122,23 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
   }
 };
 
-std::shared_ptr<ExtensionProvider> ExtensionProvider::kDefaultExtensionProvider =
+namespace {
+
+std::shared_ptr<ExtensionProvider> kDefaultExtensionProvider =
     std::make_shared<DefaultExtensionProvider>();
 
+std::mutex kDefaultExtensionProviderMutex;
+
+}  // namespace
+
 std::shared_ptr<ExtensionProvider> default_extension_provider() {
-  return ExtensionProvider::kDefaultExtensionProvider;
+  std::unique_lock<std::mutex> lock(kDefaultExtensionProviderMutex);
+  return kDefaultExtensionProvider;
+}
+
+void set_default_extension_provider(const std::shared_ptr<ExtensionProvider>& provider) {
+  std::unique_lock<std::mutex> lock(kDefaultExtensionProviderMutex);
+  kDefaultExtensionProvider = provider;
 }
 
 }  // namespace engine
--- a/cpp/src/arrow/engine/substrait/options.cc
+++ b/cpp/src/arrow/engine/substrait/options.cc
@@ -124,21 +124,21 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
 
 namespace {
 
-std::shared_ptr<ExtensionProvider> kDefaultExtensionProvider =
+std::shared_ptr<ExtensionProvider> g_default_extension_provider =
     std::make_shared<DefaultExtensionProvider>();
 
-std::mutex kDefaultExtensionProviderMutex;
+std::mutex g_default_extension_provider_mutex;
 
 }  // namespace
 
 std::shared_ptr<ExtensionProvider> default_extension_provider() {
-  std::unique_lock<std::mutex> lock(kDefaultExtensionProviderMutex);
-  return kDefaultExtensionProvider;
+  std::unique_lock<std::mutex> lock(g_default_extension_provider_mutex);
+  return g_default_extension_provider;
 }
 
 void set_default_extension_provider(const std::shared_ptr<ExtensionProvider>& provider) {
-  std::unique_lock<std::mutex> lock(kDefaultExtensionProviderMutex);
-  kDefaultExtensionProvider = provider;
+  std::unique_lock<std::mutex> lock(g_default_extension_provider_mutex);
+  g_default_extension_provider = provider;
 }
 
 }  // namespace engine
--- a/cpp/src/arrow/engine/substrait/options.h
+++ b/cpp/src/arrow/engine/substrait/options.h
@@ -74,14 +74,19 @@ class ARROW_ENGINE_EXPORT ExtensionDetails {
 
 class ARROW_ENGINE_EXPORT ExtensionProvider {
  public:
-  static std::shared_ptr<ExtensionProvider> kDefaultExtensionProvider;
   virtual ~ExtensionProvider() = default;
   virtual Result<RelationInfo> MakeRel(const std::vector<DeclarationInfo>& inputs,
                                        const ExtensionDetails& ext_details,
                                        const ExtensionSet& ext_set) = 0;
 };
 
+/// \brief Get the default extension provider
 ARROW_ENGINE_EXPORT std::shared_ptr<ExtensionProvider> default_extension_provider();
+/// \brief Set the default extension provider
+///
+/// \input[in] provider the new provider to be set as default
+ARROW_ENGINE_EXPORT void set_default_extension_provider(
+    const std::shared_ptr<ExtensionProvider>& provider);
 
 /// Options that control the conversion between Substrait and Acero representations of a
 /// plan.
--- a/cpp/src/arrow/engine/substrait/options.h
+++ b/cpp/src/arrow/engine/substrait/options.h
@@ -84,7 +84,7 @@ class ARROW_ENGINE_EXPORT ExtensionProvider {
 ARROW_ENGINE_EXPORT std::shared_ptr<ExtensionProvider> default_extension_provider();
 /// \brief Set the default extension provider
 ///
-/// \input[in] provider the new provider to be set as default
+/// \param[in] provider the new provider to be set as default
 ARROW_ENGINE_EXPORT void set_default_extension_provider(
     const std::shared_ptr<ExtensionProvider>& provider);
 


rtpsw:
cc @westonpace @icexelloss

cpp/src/arrow/engine/substrait/options.cc
```diff
@@ -122,11 +122,23 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
...
+std::shared_ptr<ExtensionProvider> kDefaultExtensionProvider =
     std::make_shared<DefaultExtensionProvider>();
 
+std::mutex kDefaultExtensionProviderMutex;
+
+}  // namespace
+
 std::shared_ptr<ExtensionProvider> default_extension_provider() {
-  return ExtensionProvider::kDefaultExtensionProvider;
+  std::unique_lock<std::mutex> lock(kDefaultExtensionProviderMutex);
```
icexelloss, line 135: Don't think we need the lock here - IMO this method doesn't need to be thread safe (I cannot imagine who would want to do this in multi threaded way)

cpp/src/arrow/engine/substrait/options.cc
```diff
@@ -122,11 +122,23 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
...
+  return kDefaultExtensionProvider;
+}
+
+void set_default_extension_provider(const std::shared_ptr<ExtensionProvider>& provider) {
```
icexelloss, line 139: Can we document these method?

icexelloss: Also - do we need to expose these in public headers?

icexelloss:
One thing wasn't clear to be is how do we dispatch some of the message to our extension provider vs use the default extension provider. For example, for the asof join message, I would imagine that being passed to the existing Acero extension provider - how would we handle that?

rtpsw:
> One thing wasn't clear to be is how do we dispatch some of the message to our extension provider vs use the default extension provider. For example, for the asof join message, I would imagine that being passed to the existing Acero extension provider - how would we handle that?

The usual pattern would be chaining of an extension provider - something along the lines of
```
class MyExtensionProvider {
public:
  MyExtensionProvider(const std::shared_ptr<ExtensionProvider>& parent_provider) :
      parent_provider_(parent_provider) {}

  Result<RelationInfo> MakeRel(const std::vector<DeclarationInfo>& inputs,
                               const google::protobuf::Any& rel,
                               const ExtensionSet& ext_set) override {
    if (rel.Is<substrait_ext::MyRel>()) {
      substrait_ext::MyRel my_rel;
      rel.UnpackTo(&my_rel);
      return MakeMyRel(inputs, my_rel, ext_set);
    }
    return parent_provider_->MakeRel(inputs, rel, ext_set);
  }

  // MakeMyRel implementation here

private:
  std::shared_ptr<ExtensionProvider> parent_provider_;
};
```
and for initialization
```
set_default_extension_provider(std::make_shared<MyExtensionProvider>(default_extension_provider());
```

icexelloss:
IMO we can handle the "fallback" behavior in Acero instead of custom code so not everyone that wants to provide custom extension need to implement this.

cpp/src/arrow/engine/substrait/options.cc
icexelloss: Not a big issue though - IMO this is a bit of overkill but wouldn't hurt.

rtpsw:
> IMO we can handle the "fallback" behavior in Acero instead of custom code so not everyone that wants to provide custom extension need to implement this.

What do you have in mind for the API? My suggestion would be to encapsulate the chaining pattern via an API like:
```
chain_extension_provider(const std::shared_ptr<ExtensionProvider>& provider);
```
Here `provider` would not need to call a parent because this would be done automatically.

cpp/src/arrow/engine/substrait/options.cc
westonpace: Better safe than sorry I think.

westonpace:
We could make the chaining of extension providers more explicit (e.g. Acero could have a stack of extension providers).  Or we could create a base class that we offer as a helper / example.  However, I'm not sure that this is a good idea.  I'd rather stick to something like what we already have as it is very simple.  Yes, most providers will want to do something similar but I don't think it is a large burden for them.

icexelloss:
>  I'd rather stick to something like what we already have as it is very simple. Yes, most providers will want to do something similar but I don't think it is a large burden for them.

This sounds fine to me. We can implement the fallback logic internally.

cpp/src/arrow/engine/substrait/options.cc
icexelloss: SGTM

cpp/src/arrow/engine/substrait/options.cc
```diff
@@ -122,11 +122,23 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
   }
 };
 
-std::shared_ptr<ExtensionProvider> ExtensionProvider::kDefaultExtensionProvider =
+namespace {
+
+std::shared_ptr<ExtensionProvider> kDefaultExtensionProvider =
     std::make_shared<DefaultExtensionProvider>();
 
+std::mutex kDefaultExtensionProviderMutex;
```
westonpace, line 130:
```suggestion
std::shared_ptr<ExtensionProvider> default_extension_provider =
    std::make_shared<DefaultExtensionProvider>();

std::mutex default_extension_provider_mutex;
```

These aren't constants so we can't use `kName`.  Will this work?

cpp/src/arrow/engine/substrait/options.cc
rtpsw: It is declared in the public header `options.h` and documented there.

cpp/src/arrow/engine/substrait/options.cc
rtpsw: You're right we can't use `kName`. I believe the right way to use is `g_name`, which I just used in the recent commit.

westonpace (approved):
Thanks for the feature.

kou:
@github-actions crossbuw submit test-ubuntu-18.04-cpp

kou:
@rtpsw @westonpace This broke `test-ubuntu-18.04-cpp` job:

https://github.com/ursacomputing/crossbow/actions/runs/4121522608/jobs/7117290841#step:5:1695

```text
FAILED: src/arrow/engine/CMakeFiles/arrow_substrait_shared.dir/substrait/options.cc.o 
/usr/local/bin/sccache /usr/bin/c++  -DARROW_ENGINE_EXPORTING -DARROW_HAVE_RUNTIME_AVX2 -DARROW_HAVE_RUNTIME_AVX512 -DARROW_HAVE_RUNTIME_BMI2 -DARROW_HAVE_RUNTIME_SSE4_2 -DARROW_HAVE_SSE4_2 -DARROW_HDFS -DARROW_MIMALLOC -DARROW_NO_DEPRECATED_API -DARROW_WITH_BROTLI -DARROW_WITH_BZ2 -DARROW_WITH_LZ4 -DARROW_WITH_RE2 -DARROW_WITH_SNAPPY -DARROW_WITH_TIMING_TESTS -DARROW_WITH_UTF8PROC -DARROW_WITH_ZLIB -DARROW_WITH_ZSTD -DURI_STATIC_BUILD -DUTF8PROC_STATIC -Darrow_substrait_shared_EXPORTS -Isrc -I/arrow/cpp/src -I/arrow/cpp/src/generated -Isubstrait_ep-generated -isystem /arrow/cpp/thirdparty/flatbuffers/include -isystem /arrow/cpp/thirdparty/hadoop/include -isystem orc_ep-install/include -isystem zstd_ep-install/include -isystem utf8proc_ep-install/include -isystem xsimd_ep/src/xsimd_ep-install/include -isystem jemalloc_ep-prefix/src -isystem mimalloc_ep/src/mimalloc_ep/include/mimalloc-2.0 -Wno-noexcept-type  -fdiagnostics-color=always  -Wall -Wno-conversion -Wno-sign-conversion -Wunused-result -fno-semantic-interposition -msse4.2  -g -Werror -O0 -ggdb -fPIC   -pthread -std=c++1z -MD -MT src/arrow/engine/CMakeFiles/arrow_substrait_shared.dir/substrait/options.cc.o -MF src/arrow/engine/CMakeFiles/arrow_substrait_shared.dir/substrait/options.cc.o.d -o src/arrow/engine/CMakeFiles/arrow_substrait_shared.dir/substrait/options.cc.o -c /arrow/cpp/src/arrow/engine/substrait/options.cc
/arrow/cpp/src/arrow/engine/substrait/options.cc:130:6: error: 'mutex' in namespace 'std' does not name a type
 std::mutex g_default_extension_provider_mutex;
      ^~~~~
/arrow/cpp/src/arrow/engine/substrait/options.cc: In function 'std::shared_ptr<arrow::engine::ExtensionProvider> arrow::engine::default_extension_provider()':
/arrow/cpp/src/arrow/engine/substrait/options.cc:135:8: error: 'unique_lock' is not a member of 'std'
   std::unique_lock<std::mutex> lock(g_default_extension_provider_mutex);
        ^~~~~~~~~~~
/arrow/cpp/src/arrow/engine/substrait/options.cc:135:8: note: suggested alternative: 'unique_copy'
   std::unique_lock<std::mutex> lock(g_default_extension_provider_mutex);
        ^~~~~~~~~~~
        unique_copy
/arrow/cpp/src/arrow/engine/substrait/options.cc:135:25: error: 'mutex' is not a member of 'std'
   std::unique_lock<std::mutex> lock(g_default_extension_provider_mutex);
                         ^~~~~
/arrow/cpp/src/arrow/engine/substrait/options.cc:135:37: error: 'g_default_extension_provider_mutex' was not declared in this scope
   std::unique_lock<std::mutex> lock(g_default_extension_provider_mutex);
                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/arrow/cpp/src/arrow/engine/substrait/options.cc:135:37: note: suggested alternative: 'default_extension_provider'
   std::unique_lock<std::mutex> lock(g_default_extension_provider_mutex);
                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                     default_extension_provider
/arrow/cpp/src/arrow/engine/substrait/options.cc:135:32: error: 'lock' was not declared in this scope
   std::unique_lock<std::mutex> lock(g_default_extension_provider_mutex);
                                ^~~~
/arrow/cpp/src/arrow/engine/substrait/options.cc:135:32: note: suggested alternative: 'clock'
   std::unique_lock<std::mutex> lock(g_default_extension_provider_mutex);
                                ^~~~
                                clock
/arrow/cpp/src/arrow/engine/substrait/options.cc: In function 'void arrow::engine::set_default_extension_provider(const std::shared_ptr<arrow::engine::ExtensionProvider>&)':
/arrow/cpp/src/arrow/engine/substrait/options.cc:140:8: error: 'unique_lock' is not a member of 'std'
   std::unique_lock<std::mutex> lock(g_default_extension_provider_mutex);
        ^~~~~~~~~~~
/arrow/cpp/src/arrow/engine/substrait/options.cc:140:8: note: suggested alternative: 'unique_copy'
   std::unique_lock<std::mutex> lock(g_default_extension_provider_mutex);
        ^~~~~~~~~~~
        unique_copy
/arrow/cpp/src/arrow/engine/substrait/options.cc:140:25: error: 'mutex' is not a member of 'std'
   std::unique_lock<std::mutex> lock(g_default_extension_provider_mutex);
                         ^~~~~
/arrow/cpp/src/arrow/engine/substrait/options.cc:140:37: error: 'g_default_extension_provider_mutex' was not declared in this scope
   std::unique_lock<std::mutex> lock(g_default_extension_provider_mutex);
                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/arrow/cpp/src/arrow/engine/substrait/options.cc:140:37: note: suggested alternative: 'default_extension_provider'
   std::unique_lock<std::mutex> lock(g_default_extension_provider_mutex);
                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                     default_extension_provider
/arrow/cpp/src/arrow/engine/substrait/options.cc:140:32: error: 'lock' was not declared in this scope
   std::unique_lock<std::mutex> lock(g_default_extension_provider_mutex);
                                ^~~~
/arrow/cpp/src/arrow/engine/substrait/options.cc:140:32: note: suggested alternative: 'clock'
   std::unique_lock<std::mutex> lock(g_default_extension_provider_mutex);
                                ^~~~
                                clock
```

Could you check this?

rtpsw:
@kou, I'll handle this. I don't have the same platform readily available, but I have a reasonable guess of the cause.

rtpsw:
@westonpace, could you please reopen this PR?  I already pushed a commit to the same branch.

icexelloss:
Yaron, Probably easier to create a new branch./PR to fix the issue. I am
not sure how reopening a merged PR would work.

rtpsw:
Create a [fix-PR](https://github.com/apache/arrow/pull/34075).

PR #34048, opened 2023-02-06
GH-32292: [R][Packaging] Use binaries built on CentOS 7 for Ubuntu < 22.04

### LINKED ISSUE #32292: [R] Build linux binaries on older image (like manylinux2014)

### LINKED ISSUE #33091: [R][CI] Remove ubuntu-18.04 from nixlibs & prebuilt binaries

kou:
### Rationale for this change

Binaries built on CentOS 7 uses `_GLIBCXX_USE_CXX11_ABI=0`. So we can't use them on Ubuntu 20.04 because
Ubuntu 20.04 uses `_GLIBCXX_USE_CXX11_ABI=1`. If we use them on Ubuntu 20.04, our R package must use
`_GLIBCXX_USE_CXX11_ABI=0` too.

There is another problem for Ubuntu 20.04. CentOS 7 uses OpenSSL 1.0 but Ubuntu 20.04 uses OpenSSL 1.1.
OpenSSL 1.0 and 1.1 are incompatible. So we can't use binaries built on CentOS 7. We need binaries for OpenSSL
1.0 and OpenSSL 1.1. We can't use the same binaries on CentOS 7 (OpenSSL 1.0) and Ubuntu 20.04 (OpenSSL 1.1).

This changes add `-D_GLIBCXX_USE_CXX11_ABI=0` to `Cflags` in `arrow.pc`. Our R package uses `Cflags` in
`arrow.pc`. So users don't need to specify `-D_GLIBCXX_USE_CXX11_ABI=0` explicitly.

This changes replace binaries built on Ubuntu 18.04 with binaries built on CentOS 7 with OpenSSL 1.1.
(The "ubuntu-18.04" binaries are replaced with the "linux-openssl-1.1" binaries.)

The "centos-7" binaries are renamed to the "linux-openssl-1.0" binaries.

The "ubuntu-22.04" binaries are renamed to the "linux-openssl-3.0" binaries.

### Are these changes tested?

Yes.

No.

* Closes: GH-33091
* Closes: #32292


kou:

cpp/src/arrow/CMakeLists.txt
```diff
@@ -576,6 +576,19 @@ if(ARROW_BUILD_BUNDLED_DEPENDENCIES)
...
+  set(GLIBCXX_USE_CXX11_ABI_SOURCE
+      ${CMAKE_CURRENT_BINARY_DIR}/try_compile_glibcxx_use_cxx_abi.cc)
+  file(WRITE ${GLIBCXX_USE_CXX11_ABI_SOURCE} "#if !_GLIBCXX_USE_CXX11_ABI\n"
+                                             "#error Not using CXX11 ABI\n" "#endif\n")
+  try_compile(IS_GLIBCXX_USE_CXX11_ABI ${CMAKE_CURRENT_BINARY_DIR}/try_compile
+              SOURCES ${GLIBCXX_USE_CXX11_ABI_SOURCE})
+  if(NOT IS_GLIBCXX_USE_CXX11_ABI)
+    string(APPEND ARROW_PC_CFLAGS " -D_GLIBCXX_USE_CXX11_ABI=0")
+  endif()
+endif()
```
assignUser, line 590: Is this really needed? We have been building and testing the R package with the dts centos 7 binary for a while now and thez worked without this? Or is this needed to use this binary on systems without dts installed?

kou: Yes, this is needed.
devtoolset-8 uses `_GLIBCXX_USE_CXX11_ABI=0` by default. So we don't need to specify it explicitly for our R package on CentOS 7.
But g++ on Ubuntu 18.04/20.04 uses `_GLIBCXX_USE_CXX11_ABI=1` by default. If we mix `_GLIBCXX_USE_CXX11_ABI=0` Apache Arrow C++ binaries (built on CentOS 7) and `_GLIBCXX_USE_CXX11_ABI=1` our R package, our R package can't find needed symbols. It causes a link error.

kou:

kou:

ci/scripts/r_install_system_dependencies.sh
```diff
@@ -37,10 +37,10 @@ if [ "$ARROW_S3" == "ON" ] || [ "$ARROW_GCS" == "ON" ] || [ "$ARROW_R_DEV" == "T
   # Install curl and OpenSSL for S3/GCS support
   case "$PACKAGE_MANAGER" in
     apt-get)
-      apt-get install -y libcurl4-openssl-dev libssl-dev
+      apt-get install -y libcurl4-openssl-dev libssl-dev pkg-config
```
nealrichardson, line 40: How common is it to be on ubuntu and not have pkg-config installed? Does this mean we would require pkg-config in order for the binary to be used?

cpp/src/arrow/CMakeLists.txt
assignUser: Ah I see, thanks for the explanation!

ci/scripts/r_install_system_dependencies.sh
kou: > How common is it to be on ubuntu and not have pkg-config installed?

Hmm. I'm not sure preference of R users on Ubuntu but I think that users who have GCC also have pkg-config.

But `rocker/r-ver:4.0.0` and `rocker/r-ver:3.6.3` doesn't have pkg-config. So we need this.

> Does this mean we would require pkg-config in order for the binary to be used?

Yes.

We can fallback to source build when a user doesn't have pkg-config:

```diff
diff --git a/r/configure b/r/configure
index ff6a9dacc4..08fa58d327 100755
--- a/r/configure
+++ b/r/configure
@@ -176,6 +176,7 @@ else
       if [ "${ARROW_DEPENDENCY_SOURCE}" = "AUTO" ] && \
            [ "${PKG_CONFIG_AVAILABLE}" = "false" ]; then
         export ARROW_DEPENDENCY_SOURCE=BUNDLED
+        export LIBARROW_BINARY=false
         echo "**** pkg-config not installed, setting ARROW_DEPENDENCY_SOURCE=BUNDLED"
       fi
 
```

Do you want me to add this change to this PR?

kou:
We can't use the `centos-7` binary on Ubuntu 20.04 because `centos-7` binary built with OpenSSL 1.0 and Ubuntu 20.04 uses OpenSSL 1.1.

https://github.com/ursacomputing/crossbow/actions/runs/4103886242/jobs/7078891973#step:12:266

> ```text
>   /opt/R/4.1.3/lib/R/library/00LOCK-arrow/00new/arrow/libs/arrow.so: undefined symbol: HMAC_CTX_cleanup
> ```

FYI: `HMAC_CTX_cleanup()` is referred by aws-sdk-cpp.

Here are solution candidates:

1. Add `centos-7-openssl-1.1` binary (We can use OpenSSL 1.1 on CentOS 7 because EPEL provides `openssl11-devel)
2. Add `ubuntu-20.04` binary 
3. Updating aws-sdk-cpp (#33808) may resolve this because OpenSSL related code is changed in the latest aws-sdk-cpp (hint: s2n-tls)

I'll try 3. first but is there any preference?

nealrichardson:
If upgrading aws-sdk-cpp solves it, great, but doesn't google-cloud also require openssl?

Could we bundle openssl rather than require it from the system? I think the answer from last time I asked this was "no because security", but also as I understand it, the python wheels include openssl in them.

ci/scripts/r_install_system_dependencies.sh
nealrichardson: I'd rather ship a binary that just works. Would it be terrible to grep for the CXX11_ABI in arrow.pc and add it to `PKG_CFLAGS`?

If we were to set `LIBARROW_BINARY=false` if no pkg-config, it should be in its own `"${PKG_CONFIG_AVAILABLE}" = "false"` check, not connected to whatever the value of `ARROW_DEPENDENCY_SOURCE` is since that's about building libarrow from source.

kou:

ci/scripts/r_install_system_dependencies.sh
kou: > I'd rather ship a binary that just works. Would it be terrible to grep for the CXX11_ABI in arrow.pc and add it to `PKG_CFLAGS`?

It will work. I've added the logic but I'm afraid of increasing ad-hoc logics again.

> If we were to set `LIBARROW_BINARY=false` if no pkg-config, it should be in its own `"${PKG_CONFIG_AVAILABLE}" = "false"` check, not connected to whatever the value of `ARROW_DEPENDENCY_SOURCE` is since that's about building libarrow from source.

It makes sense.

kou:
Upgrading aws-sdk-cpp doesn't solve the problem.

> Could we bundle openssl rather than require it from the system? I think the answer from last time I asked this was "no because security", but also as I understand it, the python wheels include openssl in them.

Hmm. I want to stop bundling OpenSSL in wheel because we can't release a fixed version in a timely manner for now.
For example, we released 10.0.1 for https://nvd.nist.gov/vuln/detail/CVE-2022-3786 .
CVE-2022-3786 was published on 2022-11-01,
the 10.0.1 vote https://lists.apache.org/thread/rlkrj9lnfmwgn7kq8hvmzf06l5z6w30k was started on 2022-11-17 and
the vote was carried on 2022-11-23 https://lists.apache.org/thread/ozo2k7l7jhc0wj3wsrpqgsvy4fosprv8 .

Anyway, I try bundling OpenSSL.

kou:

kou:

kou:

kou:
We can't use static linking for OpenSSL:

https://github.com/ursacomputing/crossbow/actions/runs/4140935566/jobs/7160057196#step:6:1857

```text
/arrow/r/libarrow/dist/s2n_tls_ep-prefix/src/s2n_tls_ep/crypto/s2n_fips.c:21:6: error: #error "Interning with OpenSSL fips-validated libcrypto is not currently supported. See https://github.com/aws/s2n-tls/issues/2741"
     #error "Interning with OpenSSL fips-validated libcrypto is not currently supported. See https://github.com/aws/s2n-tls/issues/2741"
      ^~~~~
```

nealrichardson:
Ok, so we're back to needing three different binaries then, essentially for each version of openssl. Would the broadest compatibility be to build them all with CentOS 7 + devtoolset 8, just with openssl 1 (default on the machine), openssl 1.1, and openssl 3. Using CentOS 7 would have the oldest glibc version, so it would work everywhere?

kou:

kou:

kou:

kou:

kou:

kou:

kou:

kou:

kou:
@nealrichardson This is ready to merge.

This replaces the `ubuntu-18.04` binaries with the `centos-7-openssl-1.1` binaries. The `centos-7-openssl-1.1` binaries uses OpenSSL 1.1 but built on CentOS 7 instead of Ubuntu 18.04. The `centos-7-openssl-1.1` binaries work on Debian GNU/Linux buster, Ubuntu 18.04 and Ubuntu 20.04. 

This doesn't change `ubuntu-22.04` binaries that is for OpenSSL 3. Because there aren't OpenSSL 3 packages for CentOS 7 and CentOS 7 will reach EOL on 2024-06-30 https://wiki.centos.org/About/Product . I think that we don't want to maintain CentOS 7 + OpenSSL 3.

Do you have any concern?

nealrichardson (approved):
Thanks @kou! Looks great to me. 

@assignUser you want to take one more look?

r/tools/nixlibs.R
```diff
@@ -99,8 +99,8 @@ download_binary <- function(lib) {
 # * Some other string: a "distro-version" that corresponds to a binary that is
 #   available, to override what this function may discover by default.
 #   Possible values are:
-#    * "centos-7" (gcc 8 (devtoolset), openssl 1, glib 2.17)
-#    * "ubuntu-18.04" (gcc 8, openssl 1, glib 2.27)
+#    * "centos-7" (gcc 8 (devtoolset), _GLIBCXX_USE_CXX11_ABI=0, openssl 1.0)
```
assignUser, line 102: I think renaming the other centos7 binary would communicate the difference between the two better (without having to look in nixlibs.R)?
```suggestion
#    * "centos-7-openssl-1.0" (gcc 8 (devtoolset), _GLIBCXX_USE_CXX11_ABI=0, openssl 1.0)
```

assignUser (commented):
I am going to start another CI run with the cpp jobs to make sure the changes to cmake don't affect any other builds.

assignUser:
@github-actions crossbow submit -g cpp r-binary-packages

r/tools/nixlibs.R
kou: I kept the `centos-7` name for backward compatibility. (I thought that some users may specify `centos-7` explicitly.)

I'll rename them as the followings:

* `centos-7` -> `openssl-1.0`
* `centos-7-openssl-1.1` -> `openssl-1.1`
* `ubuntu-22.04` -> `openssl-3.0`

kou:
@github-actions crossbow submit -g cpp r-binary-packages

kou:
@github-actions crossbow submit -g cpp r-binary-packages

kou:

kou:
I'll merge this tomorrow if nobody objects it.

assignUser (approved):
+1

assignUser:
(I merged now so it the change affects the next packaging-nightly run)

PR #34050, opened 2023-02-06
GH-31910: [C++] Add support for Substrait cast expression

### LINKED ISSUE #31910: [C++] Add support for Substrait cast expression
It would seem that Substrait treats cast as a dedicated "expression kind" and not a scalar function call.  We should add support for this expression kind.

**Reporter**: [Weston Pace](https://issues.apache.org/jira/browse/ARROW-16550) / @westonpace

<sub>**Note**: *This issue was originally created as [ARROW-16550](https://issues.apache.org/jira/browse/ARROW-16550). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

thisisnic:
* Closes: #31910

### DIFF
--- a/cpp/src/arrow/engine/substrait/expression_internal.cc
+++ b/cpp/src/arrow/engine/substrait/expression_internal.cc
@@ -1018,6 +1018,7 @@ Result<std::unique_ptr<substrait::Expression>> ToProto(
   }
 
   auto out = std::make_unique<substrait::Expression>();
+
   if (auto datum = expr.literal()) {
     ARROW_ASSIGN_OR_RAISE(auto literal, ToProto(*datum, ext_set, conversion_options));
     out->set_allocated_literal(literal.release());
--- a/cpp/src/arrow/engine/substrait/expression_internal.cc
+++ b/cpp/src/arrow/engine/substrait/expression_internal.cc
@@ -333,14 +333,15 @@ Result<compute::Expression> FromProto(const substrait::Expression& expr,
       ARROW_ASSIGN_OR_RAISE(auto type,
                             FromProto(cast_exp.type(), ext_set, conversion_options));
 
+      if(cast_exp.failure_behavior() == substrait::Expression_Cast_FailureBehavior::Expression_Cast_FailureBehavior_FAILURE_BEHAVIOR_THROW_EXCEPTION){
+        return compute::call("cast", std::move(input), compute::CastOptions::Safe(type));
+      } else if(cast_exp.failure_behavior() == substrait::Expression_Cast_FailureBehavior::Expression_Cast_FailureBehavior_FAILURE_BEHAVIOR_RETURN_NULL){
 
-      // do we want safe or unsafe?  
-      // e.g. compute::CastOptions::Safe(int64())
-      // e.g. compute::CastOptions::Unsafe(int32())
-      // e.g. compute::CastOptions::Safe(timestamp(TimeUnit::NANO))
-
-      compute::CastOptions cast_options;
-      return compute::call("cast", std::move(input), compute::CastOptions::Safe(type));
+      // e.g. if unspecified  
+      } else {
+        return compute::call("cast", std::move(input), compute::CastOptions::Unsafe(type));
+      }
+      
     }
 
     default:
--- a/cpp/src/arrow/engine/substrait/expression_internal.cc
+++ b/cpp/src/arrow/engine/substrait/expression_internal.cc
@@ -325,14 +325,19 @@ Result<compute::Expression> FromProto(const substrait::Expression& expr,
       ARROW_ASSIGN_OR_RAISE(auto input,
                             FromProto(cast_exp.input(), ext_set, conversion_options));
 
-      ARROW_ASSIGN_OR_RAISE(auto type,
+      ARROW_ASSIGN_OR_RAISE(auto type_nullable,
                             FromProto(cast_exp.type(), ext_set, conversion_options));
+
+      if (!type_nullable.second &&
+          conversion_options.strictness == ConversionStrictness::EXACT_ROUNDTRIP) {
+        return Status::Invalid("Substrait cast type must be of nullable type");
+      }
 
       if (cast_exp.failure_behavior() ==
           substrait::Expression_Cast_FailureBehavior::
               Expression_Cast_FailureBehavior_FAILURE_BEHAVIOR_THROW_EXCEPTION) {
         return compute::call("cast", {std::move(input)},
-                             compute::CastOptions::Safe(type.first));
+                             compute::CastOptions::Safe(type_nullable.first));
       } else if (cast_exp.failure_behavior() ==
                  substrait::Expression_Cast_FailureBehavior::
                      Expression_Cast_FailureBehavior_FAILURE_BEHAVIOR_RETURN_NULL) {
--- a/cpp/src/arrow/engine/substrait/expression_internal.cc
+++ b/cpp/src/arrow/engine/substrait/expression_internal.cc
@@ -337,9 +337,10 @@ Result<compute::Expression> FromProto(const substrait::Expression& expr,
         return compute::call("cast", std::move(input), compute::CastOptions::Safe(type));
       } else if(cast_exp.failure_behavior() == substrait::Expression_Cast_FailureBehavior::Expression_Cast_FailureBehavior_FAILURE_BEHAVIOR_RETURN_NULL){
         return Status::NotImplemented("Unsupported cast failure behavior: Expression_Cast_FailureBehavior_FAILURE_BEHAVIOR_RETURN_NULL");
-      // e.g. if unspecified  
+      // i.e. if unspecified  
       } else {
-        return compute::call("cast", std::move(input), compute::CastOptions::Unsafe(type));
+        return Status::Invalid(
+            "substrait::Expression::Cast::FailureBehavior unspecified; must be FAILURE_BEHAVIOR_RETURN_NULL or FAILURE_BEHAVIOR_THROW_EXCEPTION)";
       }
       
     }
@@ -1015,7 +1016,6 @@ Result<std::unique_ptr<substrait::Expression>> ToProto(
   }
 
   auto out = std::make_unique<substrait::Expression>();
-
   if (auto datum = expr.literal()) {
     ARROW_ASSIGN_OR_RAISE(auto literal, ToProto(*datum, ext_set, conversion_options));
     out->set_allocated_literal(literal.release());
--- a/cpp/src/arrow/engine/substrait/expression_internal.cc
+++ b/cpp/src/arrow/engine/substrait/expression_internal.cc
@@ -1012,6 +1012,7 @@ Result<std::unique_ptr<substrait::Expression>> ToProto(
     return Status::Invalid("ToProto requires a bound Expression");
   }
 
+
   auto out = std::make_unique<substrait::Expression>();
   if (auto datum = expr.literal()) {
     ARROW_ASSIGN_OR_RAISE(auto literal, ToProto(*datum, ext_set, conversion_options));
--- a/cpp/src/arrow/engine/substrait/expression_internal.cc
+++ b/cpp/src/arrow/engine/substrait/expression_internal.cc
@@ -339,7 +339,7 @@ Result<compute::Expression> FromProto(const substrait::Expression& expr,
         return compute::call("cast", {std::move(input)},
                              compute::CastOptions::Safe(std::move(type_nullable.first)));
       } else if (cast_exp.failure_behavior() ==
-                 substrait::Expression_Cast_FailureBehavior::
+                 substrait::Expression::Cast::FailureBehavior::
                      Expression_Cast_FailureBehavior_FAILURE_BEHAVIOR_RETURN_NULL) {
         return Status::NotImplemented(
             "Unsupported cast failure behavior: "
--- a/cpp/src/arrow/engine/substrait/expression_internal.cc
+++ b/cpp/src/arrow/engine/substrait/expression_internal.cc
@@ -320,6 +320,13 @@ Result<compute::Expression> FromProto(const substrait::Expression& expr,
       return function_converter(substrait_call);
     }
 
+    case substrait::Expression::kCast: {
+      const auto& cast_exp = expr.cast()
+
+
+      return compute::call("cast");
+    }
+
     default:
       break;
   }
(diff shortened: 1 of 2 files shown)


cpp/src/arrow/engine/substrait/expression_internal.cc
```diff
@@ -320,6 +320,30 @@ Result<compute::Expression> FromProto(const substrait::Expression& expr,
...
+      ARROW_ASSIGN_OR_RAISE(auto type,
+                            FromProto(cast_exp.type(), ext_set, conversion_options));
+
+      if(cast_exp.failure_behavior() == substrait::Expression_Cast_FailureBehavior::Expression_Cast_FailureBehavior_FAILURE_BEHAVIOR_THROW_EXCEPTION){
+        return compute::call("cast", std::move(input), compute::CastOptions::Safe(type));
+      } else if(cast_exp.failure_behavior() == substrait::Expression_Cast_FailureBehavior::Expression_Cast_FailureBehavior_FAILURE_BEHAVIOR_RETURN_NULL){
+
+      // e.g. if unspecified  
+      } else {
+        return compute::call("cast", std::move(input), compute::CastOptions::Unsafe(type));
```
westonpace, line 342: I notice you switch between safe and unsafe here.  I'm not sure safe/unsafe (i.e. is this value invalid?) is the same as throw/null (i.e. this value is invalid, what should we do?).  For example, an unsafe cast from int16 to int8 would overflow 500 to -12 instead of assigning null.  It seems that Substrait has no concept of safe/unsafe (might be a good addition to the spec).

cpp/src/arrow/engine/substrait/expression_internal.cc
```diff
@@ -320,6 +320,30 @@ Result<compute::Expression> FromProto(const substrait::Expression& expr,
...
+      // the type should come from cast_expr.type()
+      
+      // this will be a Result<std::pair<std::shared_ptr<DataType>
+      ARROW_ASSIGN_OR_RAISE(auto type,
+                            FromProto(cast_exp.type(), ext_set, conversion_options));
+
+      if(cast_exp.failure_behavior() == substrait::Expression_Cast_FailureBehavior::Expression_Cast_FailureBehavior_FAILURE_BEHAVIOR_THROW_EXCEPTION){
+        return compute::call("cast", std::move(input), compute::CastOptions::Safe(type));
+      } else if(cast_exp.failure_behavior() == substrait::Expression_Cast_FailureBehavior::Expression_Cast_FailureBehavior_FAILURE_BEHAVIOR_RETURN_NULL){
+
```
westonpace, line 339: I don't think Acero has any capability to do this today (might be a nice addition) so the best we can do is return a NotImplemented error.

cpp/src/arrow/engine/substrait/expression_internal.cc
```diff
@@ -320,6 +320,30 @@ Result<compute::Expression> FromProto(const substrait::Expression& expr,
       return function_converter(substrait_call);
     }
 
+    case substrait::Expression::kCast: {
+      const auto& cast_exp = expr.cast();
+      ARROW_ASSIGN_OR_RAISE(auto input,
+                            FromProto(cast_exp.input(), ext_set, conversion_options));
+
+      // need to convert from the Substrait type spec to the relevant Arrow thing
+      // do we need a switch/case or can we just get code similar to the below FromProto?
```
westonpace, line 329: Using `FromProto` as you are doing is the way to go.  The second half of the FromProto return is nullability.  In Arrow nullability is a field concept and not a type concept.  In Substrait it is a part of the type.  So we have to return two pieces when we convert to Arrow.

Acero has no concept of nullability so for the purposes of consumption we generally ignore it (this is a [documented caveat](https://arrow.apache.org/docs/dev/cpp/streaming_execution.html#types)).

This does raise a bit of a pickle when the user requests EXACT_ROUNDTRIP as we may lose information like this.  For example, one thing we will have to consider when implementing the reverse operation is whether the target type should be marked nullable or not.

A limited but safe approach could be to just pick one (probably nullable) and, if we are in EXACT_ROUNDTRIP, then reject any incoming plan that tries to cast to a non-nullable field.

A more extensible approach might be to only reject a plan if it changes the nullability of the input type.  Then, when converting to proto, we would have to make sure the output nullability of the cast matches the input nullability.  I don't yet have a good enough understanding of what the ToProto path will look like to know if this is feasible or not.

westonpace (commented):
A few thoughts, apologies if it was presumptuous to make notes on a draft.

cpp/src/arrow/engine/substrait/expression_internal.cc
thisisnic: Thanks for highlighting this, I can see that these are different things, so I'll update this.

cpp/src/arrow/engine/substrait/expression_internal.cc
thisisnic: Thanks, done!

thisisnic:
I figured a sensible approach to this PR is to write a test for the desired behaviour, and then implement something to make the test pass. I noticed other tests in a similar region of code seem to use this "Arrow -> serialize to Substrait -> deserialize to Arrow -> check things match" pattern.  However, on further inspection, it doesn't seem like a lot of 'Arrow -> Substrait' serialization is implemented.  Am I better off just manually specifying the plan using `SubstraitFromJSON()`, and testing I get the desired Arrow behaviour?

cpp/src/arrow/engine/substrait/expression_internal.cc
vibhatha: @westonpace this is a great analysis on the issue. The exact roundtrip was sort of a recurring discussion when we started wrtiting the ToProtos for relations. Can we guarantee the exactness in roundtripping? It is very tricky when we have to make that choice. Should we put a set of rules where we ignore the plans with this kind of inputs or we do not support exact roundtrip for such cases. I am not sure whether this can be generalised. But my understanding is roundtripping could be limited to such complex cases.

cpp/src/arrow/engine/substrait/expression_internal.cc
```diff
@@ -989,7 +1011,6 @@ Result<std::unique_ptr<substrait::Expression>> ToProto(
   }
 
   auto out = std::make_unique<substrait::Expression>();
-
```
vibhatha, line 992: maybe we should keep the newline here?

cpp/src/arrow/engine/substrait/expression_internal.cc
thisisnic: Thanks! I went for the simpler solution you suggestion for the moment - rejecting the plan if we try to cast to a non-nullable field when strictness is set to `EXACT_ROUNDTRIP`.

cpp/src/arrow/engine/substrait/expression_internal.cc
thisisnic: My mistake, added back in!

thisisnic:
OK, I think this is finally at the point where I feel it's no longer a draft PR and ready for more review.  Thanks for all the really detailed suggestions+help so far; this has been really useful learning for me!

cpp/src/arrow/engine/substrait/expression_internal.cc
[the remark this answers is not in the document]
thisisnic: Makes sense, have marked such plans as invalid now.

cpp/src/arrow/engine/substrait/expression_internal.cc
```diff
@@ -320,6 +320,38 @@ Result<compute::Expression> FromProto(const substrait::Expression& expr,
...
+          substrait::Expression_Cast_FailureBehavior::
+              Expression_Cast_FailureBehavior_FAILURE_BEHAVIOR_THROW_EXCEPTION) {
+        return compute::call("cast", {std::move(input)},
+                             compute::CastOptions::Safe(type_nullable.first));
+      } else if (cast_exp.failure_behavior() ==
+                 substrait::Expression_Cast_FailureBehavior::
+                     Expression_Cast_FailureBehavior_FAILURE_BEHAVIOR_RETURN_NULL) {
+        return Status::NotImplemented(
+            "Unsupported cast failure behavior: "
+            "Expression_Cast_FailureBehavior_FAILURE_BEHAVIOR_RETURN_NULL");
```
westonpace, line 346:
```suggestion
            "FAILURE_BEHAVIOR_RETURN_NULL");
```

cpp/src/arrow/engine/substrait/expression_internal.cc
```diff
@@ -320,6 +320,38 @@ Result<compute::Expression> FromProto(const substrait::Expression& expr,
...
+      if (!type_nullable.second &&
+          conversion_options.strictness == ConversionStrictness::EXACT_ROUNDTRIP) {
+        return Status::Invalid("Substrait cast type must be of nullable type");
+      }
+
+      if (cast_exp.failure_behavior() ==
+          substrait::Expression_Cast_FailureBehavior::
+              Expression_Cast_FailureBehavior_FAILURE_BEHAVIOR_THROW_EXCEPTION) {
+        return compute::call("cast", {std::move(input)},
+                             compute::CastOptions::Safe(type_nullable.first));
```
westonpace, line 340:
```suggestion
                             compute::CastOptions::Safe(std::move(type_nullable.first)));
```
minor nit

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -774,6 +774,127 @@ TEST(Substrait, CallExtensionFunction) {
...
 
+TEST(Substrait, CallCast) {
+  ExtensionSet ext_set;
+  ConversionOptions conversion_options;
+
+  ASSERT_OK_AND_ASSIGN(auto buf,
+                       internal::SubstraitFromJSON("Expression", R"({
+  "selection": {
+      "directReference": {
+        "structField": {
```
westonpace, line 785: You use camelCase here but snake_case below (i.e. `direct_reference` and `struct_field`).  I was a bit surprised it works but it seems protobuf supports both.

However, let's pick one and be consistent.

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -774,6 +774,127 @@ TEST(Substrait, CallExtensionFunction) {
   }
 }
 
+TEST(Substrait, CallCast) {
```
westonpace, line 777:
```suggestion
TEST(Substrait, Cast) {
```

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -774,6 +774,127 @@ TEST(Substrait, CallExtensionFunction) {
...
+                                                   /*ignore_unknown_fields=*/false))
+
+  ASSERT_OK_AND_ASSIGN(auto expr, DeserializeExpression(*buf, ext_set));
+
+  ASSERT_TRUE(expr.call());
+
+  ASSERT_THAT(expr.call()->arguments[0].call()->function_name, "cast");
+}
+
+TEST(Substrait, CallCastRequiresFailureBehavior) {
```
westonpace, line 819:
```suggestion
TEST(Substrait, CastRequiresFailureBehavior) {
```
Minor nit

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -774,6 +774,127 @@ TEST(Substrait, CallExtensionFunction) {
...
+        },
+        "failure_behavior": "FAILURE_BEHAVIOR_UNSPECIFIED"
+      }
+    }
+  }
+})",
+                                                   /*ignore_unknown_fields=*/false))
+
+  EXPECT_THAT(DeserializeExpression(*buf, ext_set, conversion_options),
+              Raises(StatusCode::Invalid));
```
westonpace, line 855:
```suggestion
  EXPECT_THAT(DeserializeExpression(*buf, ext_set, conversion_options),
              Raises(StatusCode::Invalid, HasSubstr("FailureBehavior unspecified")));
```

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -774,6 +774,127 @@ TEST(Substrait, CallExtensionFunction) {
...
+        },
+        "failure_behavior": "FAILURE_BEHAVIOR_THROW_EXCEPTION"
+      }
+    }
+  }
+})",
+                                                   /*ignore_unknown_fields=*/false))
+
+  EXPECT_THAT(DeserializeExpression(*buf, ext_set, conversion_options),
+              Raises(StatusCode::Invalid));
```
westonpace, line 895:
```suggestion
  EXPECT_THAT(DeserializeExpression(*buf, ext_set, conversion_options),
              Raises(StatusCode::Invalid, HasSubstr("must be of nullable type")));
```

westonpace (changes requested):
This looks good.  Just a few minor nits.  Thanks for doing this!

cpp/src/arrow/engine/substrait/expression_internal.cc
```diff
@@ -320,6 +320,38 @@ Result<compute::Expression> FromProto(const substrait::Expression& expr,
...
+                            FromProto(cast_exp.type(), ext_set, conversion_options));
+
+      if (!type_nullable.second &&
+          conversion_options.strictness == ConversionStrictness::EXACT_ROUNDTRIP) {
+        return Status::Invalid("Substrait cast type must be of nullable type");
+      }
+
+      if (cast_exp.failure_behavior() ==
+          substrait::Expression_Cast_FailureBehavior::
+              Expression_Cast_FailureBehavior_FAILURE_BEHAVIOR_THROW_EXCEPTION) {
```
vibhatha, line 338: nit
```suggestion
              Expression::Cast::FailureBehavior:: Expression_Cast_FailureBehavior_FAILURE_BEHAVIOR_THROW_EXCEPTION) {
```

vibhatha, line 338: nit

```suggestion
          substrait::Expression::Cast::FailureBehavior::
              Expression_Cast_FailureBehavior_FAILURE_BEHAVIOR_THROW_EXCEPTION) {
```

cpp/src/arrow/engine/substrait/expression_internal.cc
```diff
@@ -320,6 +320,38 @@ Result<compute::Expression> FromProto(const substrait::Expression& expr,
...
+        return compute::call("cast", {std::move(input)},
+                             compute::CastOptions::Safe(type_nullable.first));
+      } else if (cast_exp.failure_behavior() ==
+                 substrait::Expression_Cast_FailureBehavior::
+                     Expression_Cast_FailureBehavior_FAILURE_BEHAVIOR_RETURN_NULL) {
```
vibhatha, line 343: nit
```suggestion
                 substrait::Expression::Cast::FailureBehavior::
                     Expression_Cast_FailureBehavior_FAILURE_BEHAVIOR_RETURN_NULL) {
```

vibhatha:
@thisisnic thanks for working on this one. PR looks good to me and just added a minor suggestions.

cpp/src/arrow/engine/substrait/serde_test.cc
thisisnic: I am also surprised, didn't even know it could be supplied both ways - must have copied and pasted multiple bits from elsewhere.  Fixed now!

westonpace (approved):
Thanks for doing this!

PR #34038, opened 2023-02-03
GH-34037: [Python][Docs] Fix Table.drop docstring

Fokko:
It throws errors in the CI, while I was working on other docstrings:

```
pyarrow.lib.Table.drop
-> pyarrow.lib.Table.drop(self, columns)
PR01: Parameters {'columns'} not documented

pyarrow.lib.Table.drop
-> pyarrow.lib.Table.drop(self, columns)
PR01: Parameters {'columns'} not documented

pyarrow.lib.Table.drop
-> pyarrow.lib.Table.drop(self, columns)
PR01: Parameters {'columns'} not documented

Total number of docstring violations: 3
```

### Are there any user-facing changes?

* Closes: #34037

PR #34052, opened 2023-02-06
GH-34051: [C++] GcsFileSystem lazily starts sequential reads

### LINKED ISSUE #34051: [C++] Avoid unnecessary downloads in `GcsFileSystem::OpenInputFile()`

`GcsFileSystem::OpenInputFile()` eagerly starts a download to implement the `io::InputStream` aspect of the API.  However, some applications (notably those reading Parquet files) do not use this aspect of the API.  Starting a download is relatively slow in GCS, I use 25ms as a rule of thumb to predict the download time. 

This download should start lazily, so applications that won't use it do not have to pay for this cost.

### Component(s)

C++

coryan:
`OpenInputFile()` returns a `io::RandomAccessFile` which supports sequential reads as well as random access reads. The previous implementation eagerly started a sequential read, but many applications do not use that aspect of the API. Because GCS has fairly high latency, this can slow down applications that are only going to read data using `ReadAt()`. This includes applications using Parquet files via PyArrow.

Fixes #34051 

Change the GcsFileSystem class to lazily start the download used to implement the `io::InputFile` APIs.

### Are these changes tested?

I think so: the existing tests cover the affected functions.

No.

* Closes: #34051

### DIFF
--- a/cpp/src/arrow/filesystem/gcsfs.cc
+++ b/cpp/src/arrow/filesystem/gcsfs.cc
@@ -282,9 +282,6 @@ class GcsRandomAccessFile : public arrow::io::RandomAccessFile {
     ARROW_RETURN_NOT_OK(InitializeStream());
     return stream_->Read(nbytes);
   }
-  const arrow::io::IOContext& io_context() const override {
-    return stream_->io_context();
-  }
   //@}
 
   //@{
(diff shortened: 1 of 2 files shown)


coryan:
The failure in [Python / AMD64 Conda Python 3.9 Sphinx & Numpydoc](https://github.com/apache/arrow/actions/runs/4106241652/jobs/7084210511) seems unrelated, or at least I cannot figure out how it relates to the changes in this PR.  If the failure was indeed caused by this PR I would appreciate a hint in the right direction.

kou (approved):
+1

The doc CI job failure will be fixed by #34038.

PR #34054, opened 2023-02-06
GH-34053: [C++][Parquet] Write parquet page index

### LINKED ISSUE #34053: [C++][Parquet] Write parquet page index

Parquet C++ reader has supported reading page index from file. Now it is time to implement the write logic.

### Component(s)

C++, Parquet

wgtmac:
### Rationale for this change

Parquet C++ reader supports reading page index from file, but the writer does not yet support writing it.

Parquet file writer collects page index from all data pages and serializes page index into the file.

### Are these changes tested?

Not yet, will be added later.

`WriterProperties::enable_write_page_index()` and `WriterProperties::disable_write_page_index()` have been added to toggle it on and off.
* Closes: #34053

### DIFF
--- a/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
+++ b/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
@@ -21,6 +21,7 @@
 #pragma warning(disable : 4800)
 #endif
 
+#include "gmock/gmock.h"
 #include "gtest/gtest.h"
 
 #include <cstdint>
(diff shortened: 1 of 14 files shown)


wgtmac:
I have encountered several issues while working on this patch. They must be resolved before proceeding this. So I have switched it to draft and will work on the blocking issues first.

etseidl:
@wgtmac another issue to consider as you implement the page index is rows that span multiple pages. With nested columns, it is possible to have single rows that are so large that they exceed the requested page size.  arrow-cpp currently will honor the page size by splitting these rows across multiple pages.  The current parquet spec, however, seems to require that pages begin at row boundaries (i.e. the repetition level R is 0 for the first value in each page, see [here](https://github.com/apache/parquet-format/blob/5205dc7b7c0b910ea6af33cadbd2963c0c47c726/src/main/thrift/parquet.thrift#L564) and [here](https://github.com/apache/parquet-format/blob/5205dc7b7c0b910ea6af33cadbd2963c0c47c726/src/main/thrift/parquet.thrift#L918)).  Do you concur and think this should be another blocking issue or part of this PR?

wgtmac:
> @wgtmac another issue to consider as you implement the page index is rows that span multiple pages. With nested columns, it is possible to have single rows that are so large that they exceed the requested page size. arrow-cpp currently will honor the page size by splitting these rows across multiple pages. The current parquet spec, however, seems to require that pages begin at row boundaries (i.e. the repetition level R is 0 for the first value in each page, see [here](https://github.com/apache/parquet-format/blob/5205dc7b7c0b910ea6af33cadbd2963c0c47c726/src/main/thrift/parquet.thrift#L564) and [here](https://github.com/apache/parquet-format/blob/5205dc7b7c0b910ea6af33cadbd2963c0c47c726/src/main/thrift/parquet.thrift#L918)). Do you concur and think this should be another blocking issue or part of this PR?

Thanks for the information. @etseidl 

Yes I have already noticed that a record may span across different pages. But in the parquet-cpp, the page size check always happens at the end of each batch. Therefore it guarantees that a page will not split any record. Please check this function as well as where it is called for reference: https://github.com/apache/arrow/blob/master/cpp/src/parquet/column_writer.cc#L1376

etseidl:
> Yes I have already noticed that a record may span across different pages. But in the parquet-cpp, the page size check always happens at the end of each batch. Therefore it guarantees that a page will not split any record. Please check this function as well as where it is called for reference: https://github.com/apache/arrow/blob/master/cpp/src/parquet/column_writer.cc#L1376

Perhaps I'm misunderstanding, but it appears that the function you referenced is called after a batch of values is written...I don't see where it is guaranteed that the end of a batch is also the end of a row.  But thanks for working on the page indexes, I think it's an important feature that arrow-cpp currently lacks.

wgtmac:
> > Yes I have already noticed that a record may span across different pages. But in the parquet-cpp, the page size check always happens at the end of each batch. Therefore it guarantees that a page will not split any record. Please check this function as well as where it is called for reference: https://github.com/apache/arrow/blob/master/cpp/src/parquet/column_writer.cc#L1376
> 
> Perhaps I'm misunderstanding, but it appears that the function you referenced is called after a batch of values is written...I don't see where it is guaranteed that the end of a batch is also the end of a row. But thanks for working on the page indexes, I think it's an important feature that arrow-cpp currently lacks.

Please correct me if I am wrong. At least the arrow parquet writer guarantees this by calling `ColumnWriter::WriteArrow` like this: https://github.com/apache/arrow/blob/master/cpp/src/parquet/arrow/writer.cc#L154. Yes, the ParquetFileWriter itself does not prevent this.

emkornfield:
> Please correct me if I am wrong. At least the arrow parquet writer guarantees this by calling ColumnWriter::WriteArrow like this: https://github.com/apache/arrow/blob/master/cpp/src/parquet/arrow/writer.cc#L154. Yes, the ParquetFileWriter itself does not prevent this.

I think @etseidl is correct here, WriteArrow is working on leaf arrays and  IIRC rep/def levels in the code references are the only way to recover record boundaries.  Sorry its been a busy week will aim to catchup on reviews next week.  It would also be nice to not special case this for Arrow even it does somehow work there.

wgtmac:
> > Please correct me if I am wrong. At least the arrow parquet writer guarantees this by calling ColumnWriter::WriteArrow like this: https://github.com/apache/arrow/blob/master/cpp/src/parquet/arrow/writer.cc#L154. Yes, the ParquetFileWriter itself does not prevent this.
> 
> I think @etseidl is correct here, WriteArrow is working on leaf arrays and IIRC rep/def levels in the code references are the only way to recover record boundaries. Sorry its been a busy week will aim to catchup on reviews next week. It would also be nice to not special case this for Arrow even it does somehow work there.

@emkornfield Thanks for the explanation! No problem and this is not ready to review due to a series of blocking issues ahead.

I strongly agree that writing via arrow should not be a special case. It sounds like splitting page at record boundary is a new blocking issue now.

wgtmac:
Opened an issue for the CI failure which is unrelated to this PR: https://github.com/apache/arrow/issues/34328

cpp/src/parquet/statistics.cc
```diff
@@ -494,6 +494,8 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
                       int64_t null_count, int64_t distinct_count, bool has_min_max,
                       bool has_null_count, bool has_distinct_count, MemoryPool* pool)
       : TypedStatisticsImpl(descr, pool) {
+    has_null_count_ = has_null_count;
+    has_distinct_count_ = has_distinct_count;
```
kou, line 498: This is related to the GLib failure.
If this is a correct change, we can change the GLib test:

```diff
diff --git a/c_glib/test/parquet/test-statistics.rb b/c_glib/test/parquet/test-statistics.rb
index 4f21ebf00c..0367084c88 100644
--- a/c_glib/test/parquet/test-statistics.rb
+++ b/c_glib/test/parquet/test-statistics.rb
@@ -51,7 +51,7 @@ class TestParquetStatistics < Test::Unit::TestCase
 
   test("#has_n_distinct_values?") do
     assert do
-      @statistics.has_n_distinct_values?
+      not @statistics.has_n_distinct_values?
     end
   end
 
```

kou: It seems that this is a correct change.
How about adding tests related to this to `cpp/src/parquet/statistics_test.cc` or spliting this to another pull request?

kou: #34351/34355 is related.

mapleFU: I meet the same problem here, I think the syntax of "has_xxx" is like that, for a writer:
* Writer can assure that if has right null-count ( if it not has any bugs )
* Currently I found that ndv is never collected. If a user collect ndv in page1, but not collect ndv in page 2, it should be abandon.

For reader:
* When deserialize, reader should assume that ndv and null_count can be unset ( but currently, it doesn't work like this)
* Deserialized statistics can call merge, but if either `null_count` or `ndv` is unset, all null_count should be discarded.

wgtmac: @kou Thanks for the information! Let's solve it in a separate PR because removing the change in statistics does not affect this one.

@mapleFU As you are fixing statistics and hitting the same issue, do you think solving them all together is a good idea? I have left some comments in your PR and let's discuss there. I'll rebase my PR once yours is merged. Thanks!

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -5146,5 +5147,96 @@ TEST(TestArrowReadWrite, FuzzReader) {
   }
 }
 
+TEST(TestArrowReadWrite, WriteReadPageIndexRoundTrip) {
+  // Enable page index to the writer.
+  auto writer_properties = WriterProperties::Builder()
+                               .enable_write_page_index()
```
emkornfield, line 5153: it would be nice to add an assertion to at least one other test to confirm the page index is in fact not generated.

cpp/src/parquet/column_writer.cc
```diff
@@ -1482,15 +1538,14 @@ class TypedColumnWriterImpl : public ColumnWriterImpl, public TypedColumnWriter<
 
   void WriteValuesSpaced(const T* values, int64_t num_values, int64_t num_spaced_values,
                          const uint8_t* valid_bits, int64_t valid_bits_offset,
-                         int64_t num_levels) {
+                         int64_t num_levels, int64_t num_nulls) {
```
emkornfield, line 1541: we should clarify what num_nulls means here.  We've had problem in the past on interpretation of empty-lists as nulls (I believe this is what the parquet spec expects), and num_null I think it usually derived from Arrow's concept but i could be mistaken.

emkornfield:
Apologies if I missed it but do you plan to add truncation functionality for large stat values?  Also, not sure what parquet-mr does here but should we stop saving stats on individual pages if the functionality is turned on.

wgtmac:
> Apologies if I missed it but do you plan to add truncation functionality for large stat values? Also, not sure what parquet-mr does here but should we stop saving stats on individual pages if the functionality is turned on.

No, stats truncation is not part of the plan here. We can open a new issue to disable writing stats in the page header (which is deprecated by specs IIRC) once writing page index has been supported.

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
wgtmac: Added a check here: https://github.com/apache/arrow/pull/34054/files#diff-4bc4a077197b0bc71088567adb45a3a20e850290d76cb7af52e2f6b8c19ff6ebR5091

cpp/src/parquet/column_writer.cc
wgtmac: You're right. I have added a doc string to explain them.

etseidl:
> > Apologies if I missed it but do you plan to add truncation functionality for large stat values?
> 
> Large stats truncation is already implemented by parquet-cpp: https://github.com/apache/arrow/blob/main/cpp/src/parquet/statistics.h#L145. So we are good here.

Does the referenced code only apply to statistics in the  chunk metadata (and page headers)? Statistics in the page indexes can and should be truncated since there are no guarantees that those statistics are actually contained in the page. See  [second bullet](https://github.com/apache/parquet-format/blob/master/PageIndex.md#technical-approach) under "Some observations:".

wgtmac:
> > > Apologies if I missed it but do you plan to add truncation functionality for large stat values?
> > 
> > 
> > Large stats truncation is already implemented by parquet-cpp: https://github.com/apache/arrow/blob/main/cpp/src/parquet/statistics.h#L145. So we are good here.
> 
> Does the referenced code only apply to statistics in the chunk metadata (and page headers)? Statistics in the page indexes can and should be truncated since there are no guarantees that those statistics are actually contained in the page. See [second bullet](https://github.com/apache/parquet-format/blob/master/PageIndex.md#technical-approach) under "Some observations:".

It applies to both page stats (in the header) and column chunk stats. In this PR, page index directly collects page stats after truncation. So it should not be a problem here. @etseidl

emkornfield:
> It applies to both page stats (in the header) and column chunk stats. In this PR, page index directly collects page stats after truncation. So it should not be a problem here. @etseidl

I didn't follow through the code exactly but the referenced code seems to simply drop statistics not-truncate them per the other comment

wgtmac:
> I didn't follow through the code exactly but the referenced code seems to simply drop statistics not-truncate them per the other comment

Yes, simply dropping stats other than truncating stats is less error-prone. This in turn will drop column index of that column.

etseidl:
> Yes, simply dropping stats other than truncating stats is less error-prone. This in turn will drop column index of that column.

I agree, truncating UTF-8 encoded strings is a pain (I know, I've done it). But the default cutoff for statistics is 4KB. That can lead to some pretty big footers when you have hundreds or thousands of pages. For comparison, parquet-mr uses a default of 64 bytes as a max for the column index statistics. I'm not saying truncation has to be part of this PR, just that it should be considered in the future.

wgtmac:
All blocking issues have been resolved. 

Please take a look, thanks! @emkornfield @westonpace @wjones127

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -5146,5 +5158,96 @@ TEST(TestArrowReadWrite, FuzzReader) {
...
+  auto schema_node = std::static_pointer_cast<GroupNode>(parquet_schema->schema_root());
+
+  // Prepare data and each row group contains 4 rows.
+  auto record_batch = ::arrow::RecordBatchFromJSON(schema, R"([
+      [1,     "a"],
+      [2,     "b"],
+      [3,     "c"],
+      [null,  "d"],
+      [5,     null],
+      [6,     "f"]
```
emkornfield, line 5184: it would be nice to verify a repeated column also.

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -5146,5 +5158,96 @@ TEST(TestArrowReadWrite, FuzzReader) {
...
+      auto offset_index = row_group_index_reader->GetOffsetIndex(c);
+      ASSERT_NE(offset_index, nullptr);
+      ASSERT_EQ(num_pages, offset_index->page_locations().size());
+      ASSERT_EQ(0, offset_index->page_locations()[0].first_row_index);
+    }
+
+    // Verify column index of c0.
+    auto c0_column_index = row_group_index_reader->GetColumnIndex(0);
+    ASSERT_NE(c0_column_index, nullptr);
+    ASSERT_EQ(num_pages, c0_column_index->null_pages().size());
```
emkornfield, line 5233: could you remind me why this is null_pages?

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -5146,5 +5158,96 @@ TEST(TestArrowReadWrite, FuzzReader) {
...
+  ASSERT_NE(page_index_reader, nullptr);
+
+  auto encode_int64 = [=](int64_t value) {
+    return std::string(reinterpret_cast<const char*>(&value), sizeof(int64_t));
+  };
+
+  const std::vector<std::string> c0_min_values = {encode_int64(1), encode_int64(5)};
+  const std::vector<std::string> c0_max_values = {encode_int64(3), encode_int64(6)};
+  const std::vector<std::string> c1_min_values = {"a", "f"};
+  const std::vector<std::string> c1_max_values = {"d", "f"};
```
emkornfield, line 5213: should we have a test to verify dropped statistics with large value?

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -5146,5 +5158,96 @@ TEST(TestArrowReadWrite, FuzzReader) {
...
+
+    // Verify column index of c1.
+    auto c1_column_index = row_group_index_reader->GetColumnIndex(1);
+    ASSERT_NE(c1_column_index, nullptr);
+    ASSERT_EQ(num_pages, c1_column_index->null_pages().size());
+    ASSERT_EQ(BoundaryOrder::Ascending, c1_column_index->boundary_order());
+    ASSERT_EQ(c1_min_values[rg], c1_column_index->encoded_min_values()[0]);
+    ASSERT_EQ(c1_max_values[rg], c1_column_index->encoded_max_values()[0]);
+    ASSERT_TRUE(c1_column_index->has_null_counts());
+    ASSERT_EQ(c1_null_counts[rg], c1_column_index->null_counts()[0]);
```
emkornfield, line 5248: it seems like we should be verifying we the offset indexes as well?

emkornfield: Also first row index?

cpp/src/parquet/page_index.cc
```diff
@@ -426,6 +426,354 @@ class PageIndexReaderImpl : public PageIndexReader {
...
+      comparator = MakeComparator<DType>(descr_);
+    } catch (const ParquetException&) {
+      /// Simply return unordered for unsupported comparator.
+      return BoundaryOrder::Unordered;
+    }
+
+    /// Check if both min_values and max_values are in ascending order.
+    bool is_ascending = true;
+    for (size_t i = 1; i < min_values.size(); ++i) {
+      if (comparator->Compare(min_values[i], min_values[i - 1]) ||
```
emkornfield, line 559: this doesn't look write should this have an inequality?

emkornfield: or is this <= relationship for compare?

cpp/src/parquet/page_index_test.cc
```diff
@@ -416,4 +416,420 @@ TEST(PageIndex, DeterminePageIndexRangesInRowGroupWithMissingPageIndex) {
...
+    return std::string(reinterpret_cast<const char*>(&value), sizeof(float));
+  };
+
+  // Float values with no specific order.
+  std::vector<EncodedStatistics> page_stats(3);
+  page_stats.at(0).set_null_count(0).set_min(encode(2.2F)).set_max(encode(4.4F));
+  page_stats.at(1).set_null_count(0).set_min(encode(1.1F)).set_max(encode(5.5F));
+  page_stats.at(2).set_null_count(0).set_min(encode(3.3F)).set_max(encode(6.6F));
+
+  TestWriteTypedColumnIndex(schema::Float("c1"), page_stats, BoundaryOrder::Unordered,
```
emkornfield, line 534: could we add a comment? on why unored?

cpp/src/parquet/page_index_test.cc
```diff
@@ -416,4 +416,420 @@ TEST(PageIndex, DeterminePageIndexRangesInRowGroupWithMissingPageIndex) {
...
+}
+
+TEST(PageIndex, WriteFloatColumnIndex) {
+  auto encode = [=](float value) {
+    return std::string(reinterpret_cast<const char*>(&value), sizeof(float));
+  };
+
+  // Float values with no specific order.
+  std::vector<EncodedStatistics> page_stats(3);
+  page_stats.at(0).set_null_count(0).set_min(encode(2.2F)).set_max(encode(4.4F));
```
emkornfield, line 530: I might be misrembering but I though indexes were more stringent around NaN comparison (because there was previously undefined behavior, we should probably fix NaN comparisons if this hasn't been done already).

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -5146,5 +5158,96 @@ TEST(TestArrowReadWrite, FuzzReader) {
...
+  };
+
+  const std::vector<std::string> c0_min_values = {encode_int64(1), encode_int64(5)};
+  const std::vector<std::string> c0_max_values = {encode_int64(3), encode_int64(6)};
+  const std::vector<std::string> c1_min_values = {"a", "f"};
+  const std::vector<std::string> c1_max_values = {"d", "f"};
+  const std::vector<int64_t> c0_null_counts = {1, 0};
+  const std::vector<int64_t> c1_null_counts = {0, 1};
+
+  const size_t num_pages = 1;
```
emkornfield, line 5217: it would be good to verify multiple pages in a row group also, end to end.

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
wgtmac: `null_pages` is a required list to indicate whether each page is a null page or not. So its size can be used to count number of pages here.

```thrift
struct ColumnIndex {
  /**
   * A list of Boolean values to determine the validity of the corresponding
   * min and max values. If true, a page contains only null values, and writers
   * have to set the corresponding entries in min_values and max_values to
   * byte[0], so that all lists have the same length. If false, the
   * corresponding entries in min_values and max_values must be valid.
   */
  1: required list<bool> null_pages

  /**
   * Two lists containing lower and upper bounds for the values of each page
   * determined by the ColumnOrder of the column. These may be the actual
   * minimum and maximum values found on a page, but can also be (more compact)
   * values that do not exist on a page. For example, instead of storing ""Blart
   * Versenwald III", a writer may set min_values[i]="B", max_values[i]="C".
   * Such more compact values must still be valid values within the column's
   * logical type. Readers must make sure that list entries are populated before
   * using them by inspecting null_pages.
   */
  2: required list<binary> min_values
  3: required list<binary> max_values

  /**
   * Stores whether both min_values and max_values are ordered and if so, in
   * which direction. This allows readers to perform binary searches in both
   * lists. Readers cannot assume that max_values[i] <= min_values[i+1], even
   * if the lists are ordered.
   */
  4: required BoundaryOrder boundary_order

  /** A list containing the number of null values for each page **/
  5: optional list<i64> null_counts
}
```

cpp/src/parquet/page_index_test.cc
wgtmac: There is already some logic handling NaN values: https://github.com/apache/arrow/blob/main/cpp/src/parquet/statistics.cc#L300. I will look into it later to see if more actions are required.

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
wgtmac: It has been verified in line 5227 already. Let me move it after column index to be more intuitive.

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
wgtmac: Added `TEST(TestArrowReadWrite, WriteReadPageIndexRoundTripWithMultiplePages)` to cover this.

cpp/src/parquet/page_index_test.cc
wgtmac: Added `TEST(TestArrowReadWrite, WriteReadPageIndexRoundTripWithNaNs)` to cover `NaN`, `+0.0` and `-0.0`.

wgtmac:
All feedbacks have been addressed. Could you please take a look again? @emkornfield

wgtmac:
Gentle ping @emkornfield

wgtmac:
@pitrou Do you have time to take a look? It would be nice to be included in the v12.0.0 release.

emkornfield:
Sorry will try to do a review pass over this weekend.

wgtmac:
> Sorry will try to do a review pass over this weekend.

Thank you very much! @emkornfield

emkornfield:
sorry still on my radar to look at.

wgtmac:
> sorry still on my radar to look at.

No problem. It is never easy to review a large PR like this :)

wgtmac:
Gentle ping @emkornfield :)

emkornfield:
Starting to look again now, sorry for the delay.

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -5146,5 +5158,361 @@ TEST(TestArrowReadWrite, FuzzReader) {
...
+  auto read_properties = default_arrow_reader_properties();
+  auto reader = ParquetFileReader::Open(std::make_shared<BufferReader>(buffer));
+  auto metadata = reader->metadata();
+  ASSERT_EQ(2, metadata->num_row_groups());
+
+  // Make sure page index reader is not null.
+  auto page_index_reader = reader->GetPageIndexReader();
+  ASSERT_NE(page_index_reader, nullptr);
+
+  auto encode_int64 = [=](int64_t value) {
```
emkornfield, line 5205:
```suggestion
  auto encode_int64 = [](int64_t value) {
```
Or is the capture doing something there?

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -5146,5 +5158,361 @@ TEST(TestArrowReadWrite, FuzzReader) {
...
+  auto reader = ParquetFileReader::Open(std::make_shared<BufferReader>(buffer));
+  auto metadata = reader->metadata();
+  ASSERT_EQ(2, metadata->num_row_groups());
+
+  // Make sure page index reader is not null.
+  auto page_index_reader = reader->GetPageIndexReader();
+  ASSERT_NE(page_index_reader, nullptr);
+
+  const size_t num_pages = 1;
+  for (int rg = 0; rg < metadata->num_row_groups(); ++rg) {
```
emkornfield, line 5309: Are all the loops necessary in this test?  it seems like there should only be two row-groups and one column?

Given the repetiveness of these tests, is there someway to make the logic more readable (some sort of custom gmock matcher?)

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -5146,5 +5158,361 @@ TEST(TestArrowReadWrite, FuzzReader) {
...
+      // Large stats have been dropped, so does the column index.
+      ASSERT_EQ(column_index, nullptr);
+    }
+
+    // Verify offset index.
+    for (int c = 0; c < metadata->num_columns(); ++c) {
+      auto offset_index = row_group_index_reader->GetOffsetIndex(c);
+      ASSERT_NE(offset_index, nullptr);
+      ASSERT_EQ(num_pages, offset_index->page_locations().size());
+      ASSERT_EQ(0, offset_index->page_locations()[0].first_row_index);
```
emkornfield, line 5334: small nit: at least in [later versions of gtest](https://google.github.io/googletest/primer.html) expected value goes last (I forget what version we ware running here and i think it might have flip-flopped at some pont).

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
emkornfield: or maybe helper functions to consolidate elements into a vector and use EXPECT_THAT assertions over array elements?

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -5146,5 +5158,361 @@ TEST(TestArrowReadWrite, FuzzReader) {
...
+
+    // Verify offset index.
+    auto offset_index = row_group_index_reader->GetOffsetIndex(col_idx);
+    ASSERT_NE(offset_index, nullptr);
+    ASSERT_EQ(num_pages, offset_index->page_locations().size());
+    int64_t prev_offset = 0;
+    for (size_t page_idx = 0; page_idx < num_pages; ++page_idx) {
+      const auto& page_location = offset_index->page_locations()[page_idx];
+      ASSERT_EQ(static_cast<int64_t>(page_idx * 2), page_location.first_row_index);
+      ASSERT_GT(page_location.offset, prev_offset);
```
emkornfield, line 5423: my main question here is how do we know the actual offset here is correct (same with other tests)?

cpp/src/parquet/column_page.h
```diff
@@ -64,33 +65,39 @@ class DataPage : public Page {
   Encoding::type encoding() const { return encoding_; }
   int64_t uncompressed_size() const { return uncompressed_size_; }
   const EncodedStatistics& statistics() const { return statistics_; }
+  std::optional<int64_t> first_row_index() const { return first_row_index_; }
```
emkornfield, line 68: could we add some docs on when this is expected to be present, what it is (and it is relative to RowGroup or complete file?

cpp/src/parquet/column_writer.cc
```diff
@@ -335,6 +340,10 @@ class SerializedPageWriter : public PageWriter {
     if (meta_encryptor_ != nullptr) {
       UpdateEncryption(encryption::kColumnMetaData);
     }
+
+    // Serialized page writer does not need to adjust page offsets.
+    FinishPageIndex(/*final_position=*/0);
```
emkornfield, line 345: Maybe, FinishPageIndexes (it took me looking up the actual method implementation when reading through this that this finished both, column and offsets).

cpp/src/parquet/column_writer.h
```diff
@@ -91,7 +93,9 @@ class PARQUET_EXPORT PageWriter {
       bool buffered_row_group = false,
       std::shared_ptr<Encryptor> header_encryptor = NULLPTR,
       std::shared_ptr<Encryptor> data_encryptor = NULLPTR,
-      bool page_write_checksum_enabled = false);
+      bool page_write_checksum_enabled = false,
+      ColumnIndexBuilder* column_index_builder = NULLPTR,
```
emkornfield, line 97: for completeness we should clarify that the builders must outlive the the writer?

cpp/src/parquet/metadata.h
```diff
@@ -506,6 +506,21 @@ class PARQUET_EXPORT RowGroupMetaDataBuilder {
   std::unique_ptr<RowGroupMetaDataBuilderImpl> impl_;
 };
 
+/// \brief Public struct for location to all page indexes in a parquet file.
+struct PageIndexLocation {
+  /// Alias type of page index location of a row group. The index location
+  /// is located by column ordinal. If the column does not have the page index,
+  /// its value is set to std::nullopt.
+  using RowGroupIndexLocation = std::vector<std::optional<IndexLocation>>;
+  /// Alis type of page index location of a parquet file. The index location
```
emkornfield, line 515:
```suggestion
  /// Alias type of page index location of a parquet file. The index location
```

cpp/src/parquet/page_index.cc
```diff
@@ -426,6 +426,354 @@ class PageIndexReaderImpl : public PageIndexReader {
...
+      /// This is a non-null page but it lacks of meaningful min/max values.
+      /// Discard the column index.
+      state_ = BuilderState::kDiscarded;
+      return;
+    }
+
+    if (column_index_.__isset.null_counts && stats.has_null_count) {
+      column_index_.null_counts.emplace_back(stats.null_count);
+    } else {
+      column_index_.__isset.null_counts = false;
```
emkornfield, line 483: does it pay to clear null_counts here?

cpp/src/parquet/page_index.cc
```diff
@@ -426,6 +426,354 @@ class PageIndexReaderImpl : public PageIndexReader {
...
+  BoundaryOrder::type DetermineBoundaryOrder(const std::vector<T>& min_values,
+                                             const std::vector<T>& max_values) const {
+    DCHECK_EQ(min_values.size(), max_values.size());
+    if (min_values.empty()) {
+      return BoundaryOrder::Unordered;
+    }
+
+    std::shared_ptr<TypedComparator<DType>> comparator;
+    try {
+      comparator = MakeComparator<DType>(descr_);
```
emkornfield, line 550: doesn't logical type need to be accounted for here for things like Decimal type?

cpp/src/parquet/page_index_test.cc
```diff
@@ -416,4 +416,420 @@ TEST(PageIndex, DeterminePageIndexRangesInRowGroupWithMissingPageIndex) {
...
+      ASSERT_EQ(page_stats[i].max(), column_index->encoded_max_values()[i]);
+      if (has_null_counts) {
+        ASSERT_EQ(page_stats[i].null_count, column_index->null_counts()[i]);
+      }
+    }
+  }
+}
+
+TEST(PageIndex, WriteInt32ColumnIndex) {
+  auto encode = [=](int32_t value) {
```
emkornfield, line 494:
```suggestion
  auto encode = [](int32_t value) {
```

cpp/src/parquet/page_index_test.cc
```diff
@@ -416,4 +416,420 @@ TEST(PageIndex, DeterminePageIndexRangesInRowGroupWithMissingPageIndex) {
...
+  auto sink = CreateOutputStream();
+  PageIndexLocation location;
+  builder->WriteTo(sink.get(), &location);
+  PARQUET_ASSIGN_OR_THROW(auto buffer, sink->Finish());
+  ASSERT_EQ(num_row_groups, location.column_index_location.size());
+  ASSERT_EQ(num_row_groups, location.offset_index_location.size());
+
+  // Verify data of deserialized page index.
+  auto properties = default_reader_properties();
+  for (size_t rg = 0; rg < num_row_groups; ++rg) {
```
emkornfield, line 789: would unwinding the loops + appropriate helper methods make this signifantly more verbose?

Having logic in tests, at least for me, makes them much more difficult to read (sometimes it is warranted but sometimes it is cleaner to try make things less complex).

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
wgtmac: Unfortunately we don't. We may need to change the `PageReader` to return a list of offsets to data pages, otherwise we have no idea. The only thing we are sure is that the page offsets are in the ascending order so here is why I use `ASSERT_GT` to check offsets.

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
wgtmac: I found a related discussion here: https://groups.google.com/a/chromium.org/g/chromium-dev/c/3aQU5iM5Ov0

cpp/src/parquet/page_index.cc
wgtmac: No, I have cleared it at line 508 already.

cpp/src/parquet/page_index.cc
wgtmac: It has already been taken care of because `descr_` knows the logical type and sort order.

cpp/src/parquet/page_index_test.cc
wgtmac: I have rewritten the tests. Please take a look, thanks! @emkornfield

wjones127 (approved):
This looks good to me. I also tested it locally by turning on page index by default and verify the tests pass for C++ and Python (after changing assertions that it is turned off by default).

cpp/src/parquet/page_index.cc
wjones127: `descr_` does, it doesn't seem to use the logical type here:

https://github.com/apache/arrow/blob/82d48a93d0dc5ea7826207f154d9f6a1250ec57b/cpp/src/parquet/statistics.cc#L811

But it does seem to be handled by the sort order argument:

https://github.com/apache/arrow/blob/82d48a93d0dc5ea7826207f154d9f6a1250ec57b/cpp/src/parquet/statistics_test.cc#L77-L82

wgtmac: Yes, actually the sort order is deduced from either logical_type or converted_type as below:

```cpp
  SortOrder::type sort_order() const {
    auto la = logical_type();
    auto pt = physical_type();
    return la ? GetSortOrder(la, pt) : GetSortOrder(converted_type(), pt);
  }
```

Detail can be found here: https://github.com/apache/arrow/blob/main/cpp/src/parquet/schema.h. So I think we are good. @wjones127 @emkornfield

wgtmac:
> This looks good to me. I also tested it locally by turning on page index by default and verify the tests pass for C++ and Python (after changing assertions that it is turned off by default).

Thanks for your review!

wgtmac:
@wjones127 Could we make it to the v12.0.0 release?

wjones127:
I will merge this later today if @emkornfield has no further comments.

PR #34057, opened 2023-02-06
GH-34056: [C++] Add Utility function to simplify converting any row-based structure into an `arrow::RecordBatchReader` or  an `arrow::Table`

### LINKED ISSUE #34056: [C++] Add Utility function to simplify converting any row-based structure into an `arrow::RecordBatchReader` or an `arrow::Table`

*Describe the enhancement requested*

Arrow is column based, but often clients need to import external data sources that are stored in a row based fashion. To help simplify the process, I propose we create a `RowsToBatches` utility function that can take any valid C++ range (`std::begin`/`std::end` is defined for `T`) and returns an `arrow::RecordBatchReader` (convertible to an `arrow::Table`). This is particularly useful when useful when the data types for each column are not known at compile time - like in the case of an `std::variant`

The interface could look like the following (simplified for clarity)

```
Result<std::shared_ptr<RecordBatchReader>>> RowsToBatches(const std::shared_ptr<Schema>& schema, std::reference_wrapper<Range> rows, DataPointConvertor&& data_point_convertor);
```

See linked pull request for full details. The client would only need to provide their `Schema` and a callable type that converts their structure’s types into the associated arrow types.

If the client type is not a C++ range, they can either add iterators or write a wrapper/adaptor that provides the iterators for the type.

*Example Usage:*

```
auto IntConvertor = [](ArrayBuilder& array_builder, int value) {
	return static_cast<Int64Builder&>(array_builder).Append(value);
};
std::vector<std::vector<int>> data = {{1, 2, 4}, {5, 6, 7}};
auto batches = RowsToBatches(kTestSchema, std::ref(data), IntConvertor);
```
*Example Supported Types:*
 - `std::vector<std::vector<std::variant<int, bsl::string>>>`
 - `std::vector<MyRowStruct>`

### Component(s)

C++

gringasalpastor:
*Are these changes tested?*

The following tests are provided:
- basic usage
- const ranges
- custom struct accessor
- usage with `std::variant`

* Closes: #34056


gringasalpastor:
@wjones127 If you could take a look, thanks.

cpp/src/arrow/util/rows_to_batches.h
```diff
@@ -0,0 +1,135 @@
...
+           data_point_convertor)]() mutable -> Result<std::shared_ptr<RecordBatch>> {
+    if (rows_ittr == rows_ittr_end) return NULLPTR;
+
+    ARROW_ASSIGN_OR_RAISE(
+        auto record_batch_builder,
+        RecordBatchBuilder::Make(schema, default_memory_pool(), batch_size));
+
+    for (size_t i = 0; i < batch_size and (rows_ittr != rows_ittr_end);
+         i++, std::advance(rows_ittr, 1)) {
+      size_t col_index = 0;
```
wjones127, line 116: This is int in `GetField()`, so should match that.
```suggestion
      int col_index = 0;
```

cpp/src/arrow/util/rows_to_batches.h
```diff
@@ -0,0 +1,135 @@
...
+
+/// \param[in] data_point_convertor - client provided callable type that will convert
+/// the structure’s data points into the corresponding arrow types. The convertor must
+/// return an error `Status` if an error happens during conversion.
+
+/// \param[in] row_accessor - In the common case where the value of each row iterated
+/// over is it's self also directly iterable, the client can just use the default.
+/// the provided callable must take the values of the otter `rows` range and return a
+/// `std::reference_wrapper<Range>` to the data points in a given row.
+/// see: /ref `MakeDefaultRowAccessor`
```
wjones127, line 90: This is obvious in most cases, but worth saying for some others: it needs to return the data points in the proper order.

Also: "otter `rows`"?

```suggestion
/// \param[in] row_accessor - In the common case where the value of each row iterated
/// over is it's self also directly iterable, the client can just use the default.
/// The provided callable must take the values of the otter `rows` range and return a
/// `std::reference_wrapper<Range>` to the data points in a given row. The data points
/// must be in order of their corresponding fields in the schema.
/// see: /ref `MakeDefaultRowAccessor`
```

cpp/src/arrow/util/rows_to_batches.h
```diff
@@ -0,0 +1,135 @@
...
+       schema = schema, row_accessor = std::forward<RowAccessor>(row_accessor),
+       data_point_convertor = std::forward<DataPointConvertor>(
+           data_point_convertor)]() mutable -> Result<std::shared_ptr<RecordBatch>> {
+    if (rows_ittr == rows_ittr_end) return NULLPTR;
+
+    ARROW_ASSIGN_OR_RAISE(
+        auto record_batch_builder,
+        RecordBatchBuilder::Make(schema, default_memory_pool(), batch_size));
+
+    for (size_t i = 0; i < batch_size and (rows_ittr != rows_ittr_end);
```
wjones127, line 114: In this project we always prefer `&&`:
```suggestion
    for (size_t i = 0; i < batch_size && (rows_ittr != rows_ittr_end);
```

cpp/src/arrow/util/rows_to_batches.h
```diff
@@ -0,0 +1,135 @@
...
+/// see: /ref `MakeDefaultRowAccessor`
+
+/// \return `Result<std::shared_ptr<RecordBatchReader>>>` result will be a
+/// `std::shared_ptr<RecordBatchReader>>` if not errors occurred, else an error status.
+template <class Range, class DataPointConvertor,
+          class RowAccessor = decltype(detail::MakeDefaultRowAccessor())>
+[[nodiscard]] typename std::enable_if_t<detail::is_range<Range>::value,
+                                        Result<std::shared_ptr<RecordBatchReader>>>
+/* Result<std::shared_ptr<RecordBatchReader>>> */ RowsToBatches(
+    const std::shared_ptr<Schema>& schema, std::reference_wrapper<Range> rows,
```
wjones127, line 99: `std::reference_wrapper` is new to me. Is there a reason you use that rather than taking `const Range&`? It seems like this requires that users wrap their types with `std::ref` or `std::cref`.

wjones127 (changes requested):
Thanks for submitting this! It would be nice to have at least one example with nested row data.

I'd suggest either adding another test, or to be more complete it would be interesting to see how [the rapidjson example](https://github.com/apache/arrow/blob/master/cpp/examples/arrow/rapidjson_row_converter.cc) would be changed to use this.

cpp/src/arrow/util/rows_to_batches.h
```diff
@@ -0,0 +1,135 @@
...
+    DataPointConvertor&& data_point_convertor,
+    RowAccessor&& row_accessor = detail::MakeDefaultRowAccessor()) {
+  const std::size_t batch_size = 1024;
```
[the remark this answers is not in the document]
gringasalpastor: sure, will update to make it an parameter

cpp/src/arrow/util/rows_to_batches.h
gringasalpastor: Good question. The intention was to block the client from passing a temporary (`const T&&`). `reference_wrapper ` does this by using `= delete` on that constructor. A temporary doesn't make sense because the lambda will be called after the end of the function call.

After more investigation, I found a stackoverflow showing that you can use `= delete`  on free functions. I am going to try to do that so can can just pass a `const T&`

gringasalpastor: Updated PR to use `const &&` and delete overload for `const T&&`

cpp/src/arrow/util/rows_to_batches_test.cc
```diff
@@ -0,0 +1,120 @@
...
+
+auto IntConvertor = [](ArrayBuilder& array_builder, int value) {
+  return static_cast<Int64Builder&>(array_builder).Append(value);
+};
+
+bool CompareJson(const arrow::Table& arrow_table, const std::string& json,
+                 const std::string& field_name) {
+  const auto col = arrow_table.GetColumnByName(field_name);
+  return arrow::ChunkedArrayFromJSON(col->type(), {json})->Equals(col);
+};
```
wjones127, line 42:
```suggestion
}
```

cpp/src/arrow/util/rows_to_batches.h
```diff
@@ -0,0 +1,159 @@
...
+
+namespace detail {
+
+// Default identity function row accessor. Used to for the common case where the value
+// of each row iterated over is it's self also directly iterable.
+[[nodiscard]] constexpr inline auto MakeDefaultRowAccessor() {
+  return [](auto& x) -> Result<decltype(std::ref(x))> { return std::ref(x); };
+}
+
+// Meta-funciton to check if a type `T` is a range (iterable using `std::begin()` /
```
wjones127, line 38:
```suggestion
// Meta-function to check if a type `T` is a range (iterable using `std::begin()` /
```

cpp/src/arrow/util/rows_to_batches_test.cc
```diff
@@ -0,0 +1,120 @@
...
+
+TEST(RowsToBatches, Variant) {
+  auto VariantConvertor = [](ArrayBuilder& array_builder,
+                             const std::variant<int, std::string>& value) {
+    if (std::holds_alternative<int>(value))
+      return dynamic_cast<Int64Builder&>(array_builder).Append(std::get<int>(value));
+    else
+      return dynamic_cast<arrow::StringBuilder&>(array_builder)
+          .Append(std::get<std::string>(value).c_str(),
+                  std::get<std::string>(value).length());
```
wjones127, line 105:
```suggestion
          .Append(std::get<std::string>(value));
```

wjones127 (changes requested):
Just two minor changes, then I think this is ready to go.

gringasalpastor:
> Just two minor changes, then I think this is ready to go.

Committed the last 2 suggested changes, let me know if there is anything else you would like to see updated @wjones127

wjones127:
@gringasalpastor There seems to be one CI failure that is legitimate. Could you address that?

```
/arrow/cpp/src/arrow/util/rows_to_batches.h:107: error: The following parameters of arrow::util::RowsToBatches(const std::shared_ptr< Schema > &schema, const Range &rows, DataPointConvertor &&data_point_convertor, RowAccessor &&row_accessor=detail::MakeDefaultRowAccessor(), MemoryPool *pool=default_memory_pool(), const std::size_t batch_size=1024) are not documented:
  parameter 'pool'
  parameter 'batch_size' (warning treated as error, aborting now)
1
```

gringasalpastor:
> @gringasalpastor There seems to be one CI failure that is legitimate. Could you address that?
> 
> ```
> /arrow/cpp/src/arrow/util/rows_to_batches.h:107: error: The following parameters of arrow::util::RowsToBatches(const std::shared_ptr< Schema > &schema, const Range &rows, DataPointConvertor &&data_point_convertor, RowAccessor &&row_accessor=detail::MakeDefaultRowAccessor(), MemoryPool *pool=default_memory_pool(), const std::size_t batch_size=1024) are not documented:
>   parameter 'pool'
>   parameter 'batch_size' (warning treated as error, aborting now)
> 1
> ```

@wjones127 Thanks, I just pushed a new commit to add documentation for these.

wjones127 (approved):
Thanks @gringasalpastor!

PR #34060, opened 2023-02-07
GH-34059: [C++] Add a fetch node based on a batch index

### LINKED ISSUE #34059: [C++] Create a fetch node based on a batch index property

This node will eventually be usable mid-plan allowing us to get rid of the concept of a TopKSink and also to add Substrait consumer support for fetch.

This feature introduces the idea of an "ExecBatch::index" which is described more fully in https://docs.google.com/document/d/1MfVE9td9D4n5y-PTn66kk4-9xG7feXs1zSFf-qxQgPs/edit

This feature is part of a larger set of features and this node will not be very useful on its own until some of those features are added (currently, all nodes except table_source will either fail to set the index or will reset it back to -1 and the fetch node will fail).  See #32991 for the larger roadmap.

### Component(s)

C++

westonpace:
This PR introduces the concept of `ExecBatch:index` but does not yet do much with it.  As a proof of concept this PR adds a fetch node which can be inserted anywhere in the plan (not just at the sink) to satisfy `LIMIT x OFFSET y` (Substrait calls this fetch and so I have also).

This PR also introduces two sequencing accumulation queues which will be useful, I hope, for anyone implementing nodes that rely on ordered execution.

This PR unfortunately introduces a new query option which is whether or not the sink node should pay the small performance hit required to sequence output.  While considering how best to add this option I realized we will probably have more query options in the near future regarding "how much RAM to use" (e.g. spillover) and potentially more beyond that.

So I have taken all the options and put them into `arrow::compute::QueryOptions` (this already existed but it was not user facing and I added more things to it).  I added a new DeclarationToXyz overload that accepts QueryOptions.  This has, unfortunately, led to a bit of overload explosion but I think this should be the last new addition to the overload set (and we can deprecate the older overloads at some point).

This PR also includes a new `gen::Gen / gen::TestGen` facility for generating test tables for input.  I'd like to eventually use this to simplify some of the existing exec plan tests as well.  I'm willing to split this into a separate PR if that makes sense.
* Closes: #34059


cpp/src/arrow/compute/exec.h
```diff
@@ -209,6 +212,11 @@ struct ARROW_EXPORT ExecBatch {
   /// whether any values are Scalar.
   int64_t length = 0;
 
+  /// \brief index of this batch in a sorted stream of batches
+  ///
+  /// This index must be strictly monotoic starting at 0 without gaps
```
lidavidm, line 217: nit
```suggestion
  /// This index must be strictly monotonic starting at 0 without gaps, or kUnsequencedIndex
```

cpp/src/arrow/compute/exec/exec_plan.h
```diff
@@ -408,6 +408,54 @@ struct ARROW_EXPORT Declaration {
...
+  /// \brief Should the plan use a legacy batching strategy
+  ///
+  /// This is currently in place only to support the Scanner::ToTable
+  /// method.  This method relies on batch indices from the scanner
+  /// remaining consistent.  This is impractical in the ExecPlan which
+  /// might slice batches as needed (e.g. for a join)
+  ///
+  /// However, it still works for simple plans and this is the only way
+  /// we have at the moment for maintaining implicit order.
+  bool use_legacy_batching;
```
lidavidm, line 423: nit: `= false` like all the other fields?

cpp/src/arrow/compute/exec/accumulation_queue.h
```diff
@@ -53,5 +56,98 @@ class AccumulationQueue {
...
+  virtual ~SequencingQueue() = default;
+
+  /// Insert a batch into the queue
+  ///
+  /// This will insert the batch into the queue.  If this batch was the next batch
+  /// to deliver then this will trigger 1+ calls to the process callback to generate
+  /// 1+ tasks.
+  ///
+  /// The task generated by this call will be executed immediately.  The remaining
+  /// tasks will be scheduled using the schedule callback.
```
lidavidm, line 102: nit: this is a little vague to me (without having read the implementation yet). "The task" here is "The first task" right?

cpp/src/arrow/compute/exec/accumulation_queue.cc
```diff
@@ -54,5 +59,112 @@ void AccumulationQueue::Clear() {
...
+
+ private:
+  Status DoProcess(std::unique_lock<std::mutex>&& lk) {
+    while (!queue_.empty() && queue_.top().index == next_index_) {
+      ExecBatch next(queue_.top());
+      queue_.pop();
+      next_index_++;
+      lk.unlock();
+      // If we bail here we don't hold the lock so that is ok.  is_processing_ will
+      // never switch to true so no other threads can process but that should be ok
```
lidavidm, line 142: never switch to false?

cpp/src/arrow/testing/generator.h
```diff
@@ -234,4 +236,84 @@ class ARROW_TESTING_EXPORT ConstantArrayGenerator {
...
+    std::vector<GeneratorField> column_gens);
+// For generating batches with 0 columns (though they can still have length)
+ARROW_TESTING_EXPORT std::unique_ptr<DataGenerator> EmptyGen();
+
+ARROW_TESTING_EXPORT std::unique_ptr<GTestDataGenerator> TestGen(
+    std::vector<std::shared_ptr<ArrayGenerator>> column_gens);
+ARROW_TESTING_EXPORT std::unique_ptr<GTestDataGenerator> TestGen(
+    std::vector<GeneratorField> column_gens);
+// For generating batches with 0 columns (though they can still have length)
+ARROW_TESTING_EXPORT std::unique_ptr<GTestDataGenerator> EmptyTestGen();
```
lidavidm, line 298: possibly instead of duplicate methods, a method (or instance method of DataGenerator) could convert one to the other?

cpp/src/arrow/compute/exec/fetch_node.cc
```diff
@@ -0,0 +1,207 @@
...
+
+  Page NextPage(const ExecBatch& batch) {
+    int64_t rows_in_batch_to_skip = 0;
+    if (rows_to_skip_ > 0) {
+      rows_in_batch_to_skip = std::min(rows_to_skip_, batch.length);
+      rows_to_skip_ -= rows_in_batch_to_skip;
+    }
+
+    int64_t rows_in_batch_to_send = 0;
+    if (rows_to_send_ > 0) {
```
lidavidm, line 62: hmm, don't we only want to send rows if remaining rows_to_skip_ == 0?

cpp/src/arrow/compute/exec/accumulation_queue.h
westonpace: Yes.  Basically, don't create a new thread task for the batch of data that just arrived, since it's already in the processor's cache.  So if the batches arrive:

1 2 4 3 5

then only a single thread task is scheduled (for 3).

cpp/src/arrow/compute/exec/fetch_node.cc
westonpace: That should (I think) be addressed on line 63.  If `rows_to_skip_ > 0` at this point then we know `rows_in_batch_to_skip == batch.length` and therefore `rows_in_batch_to_send == 0`.

I'm pretty sure the test cases cover this scenario as well

lidavidm: Ah, indeed. Ok, sounds good.

cpp/src/arrow/testing/generator.h
westonpace: Not sure which duplication you are referring to.  Do you mean the `Gen`/`TestGen` duplication?  That should be doable...

`auto test_gen = Gen(...).FailOnError();`

lidavidm (approved):

cpp/src/arrow/testing/generator.h
lidavidm: Yes, Gen/TestGen. But not a big deal.

westonpace: It would be nice to avoid the need for `EmptyGen` too.  I'll revisit this.

westonpace: Ok, I've managed to collapse `Gen/TestGen` and get rid of `EmptyGen`.  Now, if you want a test gen you can do:

```
gen::Gen({...})->FailOnError()->Table(...);
```

cpp/src/arrow/compute/exec/accumulation_queue.h
westonpace: I cleaned up the wording a little and added an extra paragraph.

cpp/src/arrow/compute/exec/exec_plan.h
westonpace: Done (also got rid of the explicit constructor)

cpp/src/arrow/compute/exec.h
westonpace: Added with slightly different wording.

PR #34065, opened 2023-02-07
GH-34063: [C++] Avoid waste in `GcsFileSystem::ReadAt()`

### LINKED ISSUE #34063: [C++] Stop downloads ASAP in `GcsFileSystem`

`GcsFileSystem` uses `ReadFromOffset()` to implement `Seek()` and reads from the beginning of the file.  I think I was too clever by half and reused that to implement `ReadAt()`.  The code works, because the download is interrupted when enough bytes are received.  But the code is wasteful, because more data than needed may be returned by the service and consume memory in the local socket buffers, incur higher egress fees (if applicable) and potentially dirty the socket pool as these downloads do not terminate gracefully.

`GcsFileSystem` should use `gcs::ReadRange(start, end)` in the implementation of `ReadAt()`.  Fortunately both `gcs::ReadRange()` and `gcs::ReadFromOffset()` can be used at the same time, and when default initialized they have no effect.  That means we can make the change with minimal changes.

### Component(s)

C++

coryan:
With this change `GcsFileSystem::ReadAt()` will only request *exactly* the bytes it needs.

Fixes #34063.

### Are these changes tested?

There are existing tests for `ReadAt()` this is just a performance change.

No.

* Closes: #34063

### DIFF
--- a/cpp/src/arrow/filesystem/gcsfs.cc
+++ b/cpp/src/arrow/filesystem/gcsfs.cc
@@ -243,7 +243,7 @@ class GcsOutputStream : public arrow::io::OutputStream {
 };
 
 using InputStreamFactory = std::function<Result<std::shared_ptr<GcsInputStream>>(
-    gcs::Generation, gcs::ReadFromOffset)>;
+    gcs::Generation, gcs::ReadRange, gcs::ReadFromOffset)>;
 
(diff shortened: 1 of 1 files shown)


coryan:
I am attributing the  `Python / AMD64 Conda Python 3.9 Sphinx & Numpydoc` failure to #34038.

lidavidm (approved):
Thanks!

Issue #33982, opened 2023-02-01
[Python] DataFrame Interchange Protocol


AlenkaF:
### Describe the enhancement requested

This is an umbrella issue for all tasks connected to the Dataframe Interchange Protocol.
https://data-apis.org/dataframe-protocol/latest/index.html

- [x] https://github.com/apache/arrow/issues/33346
- [x] https://github.com/apache/arrow/issues/33926
- [x] https://github.com/apache/arrow/issues/33980
- [x] https://github.com/apache/arrow/issues/33984
- [x] Implement DataFrame Interchange Protocol Compliance Suite https://github.com/data-apis/dataframe-interchange-tests/pull/21

### Component(s)

Python

stinodego:
I noticed that there is currently no support yet in the `from_dataframe` function for boolean columns with bitsize 1. Is there an issue for implementing this?

```python
import polars as pl
import pyarrow as pa
import pyarrow.interchange

df = pl.Series("a", [True, False], dtype=pl.Boolean).to_frame()
dfi = df.__dataframe__()
pa.interchange.from_dataframe(dfi)
# NotImplementedError: Conversion for (<DtypeKind.BOOL: 20>, 1, 'b', '=') is not yet supported.
```

AlenkaF:
No, there is no issue - I think we missed it as we didn't have the use case then.
Will open a new issue to implement it, thanks!

AlenkaF:
Closing this issue as all tasks are done.

PR #34069, opened 2023-02-07
GH-34068: [C++][Gandiva] Add extra functions

### LINKED ISSUE #34068: Provide common transcendental / bitwise operations in Gandiva

We (the ALICE experiment at the LHC) use Arrow and Gandiva as basis of our data analysis framework. However, we are currently maintaining a patch on top of the upstream arrow which introduces some basic bitwise and transcendental operations. In particular we need to support:

* 16 / 8 bit boolean operations.
* not operations on unsigned quantities.
* float sqrtf, logf, log10f, powerf, absf
* float trigonometric functions

We actually have the code already and we use it in production, we will open a PR and we would be happy to donate it upstream to simplify our adoption of vanilla arrow releases.

### Component(s)

C++ - Gandiva

ktf:
### Rationale for this change

A number of trivial functions, or trivial functions acting on different data types (e.g. float rather than double) is currently missing in Gandiva. This PR  introduces a number of them, as requested in [GH-34068](https://github.com/apache/arrow/issues/34068). While the patch is quite straightforward, the goal is to avoid having to maintain them in our own private branch and we are happy to donate it to the community.

* Additional native functions for float32 in Gandiva
* Extend bitwise operations for more int types
* Adapt to upstream trigonometric function definitions
* Prevent clash with c++ variant

### Are these changes tested?

We updated the unit tests for added functions and we (ALICE experiment at the LHC) are currently using them to analyse  terabytes (if not yet petabytes) of data.

There are not any user facing changes unless you consider the availability of the new Gandiva functions as such. As far as we understand, this PR is not breaking any public APIs.
* Closes: #34068

### DIFF
--- a/cpp/src/gandiva/function_registry_math_ops.cc
+++ b/cpp/src/gandiva/function_registry_math_ops.cc
@@ -79,6 +79,10 @@ std::vector<NativeFunction> GetMathOpsFunctionRegistry() {
       MATH_UNARY_OPS_FLOAT(expf, {}), MATH_UNARY_OPS_FLOAT(logf, {}),
       MATH_UNARY_OPS_FLOAT(log10f, {}),
 
+      MATH_UNARY_OPS_FLOAT(sqrtf, {}), MATH_UNARY_OPS_FLOAT(cbrtf, {}),
+      MATH_UNARY_OPS_FLOAT(expf, {}), MATH_UNARY_OPS_FLOAT(logf, {}),
+      MATH_UNARY_OPS_FLOAT(log10f, {}),
+
       MATH_BINARY_UNSAFE(log, {}),
 
       BINARY_SYMMETRIC_SAFE_NULL_IF_NULL(power, {"pow"}, float64),
--- a/cpp/src/gandiva/function_registry_math_ops.cc
+++ b/cpp/src/gandiva/function_registry_math_ops.cc
@@ -16,7 +16,6 @@
 // under the License.
 
 #include "gandiva/function_registry_math_ops.h"
-
 #include "gandiva/function_registry_common.h"
 
 namespace gandiva {
--- a/cpp/src/gandiva/function_registry_math_ops.cc
+++ b/cpp/src/gandiva/function_registry_math_ops.cc
@@ -16,6 +16,7 @@
 // under the License.
 
 #include "gandiva/function_registry_math_ops.h"
+
 #include "gandiva/function_registry_common.h"
 
 namespace gandiva {
@@ -27,6 +28,11 @@ namespace gandiva {
       UNARY_SAFE_NULL_IF_NULL(name, ALIASES, uint64, float64),  \
       UNARY_SAFE_NULL_IF_NULL(name, ALIASES, float32, float64), \
       UNARY_SAFE_NULL_IF_NULL(name, ALIASES, float64, float64)
+
+#define MATH_UNARY_OPS_FLOAT(name, ALIASES)                    \
+  UNARY_SAFE_NULL_IF_NULL(name, ALIASES, int32, float32),      \
+      UNARY_SAFE_NULL_IF_NULL(name, ALIASES, uint32, float32), \
+      UNARY_SAFE_NULL_IF_NULL(name, ALIASES, float32, float32)
 
 #define MATH_BINARY_UNSAFE(name, ALIASES)                          \
   BINARY_UNSAFE_NULL_IF_NULL(name, ALIASES, int32, float64),       \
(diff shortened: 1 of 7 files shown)


ktf:
@aalkin FYI.

cpp/src/gandiva/function_registry_math_ops.cc
```diff
@@ -62,11 +73,16 @@ namespace gandiva {
 std::vector<NativeFunction> GetMathOpsFunctionRegistry() {
   static std::vector<NativeFunction> math_fn_registry_ = {
       MATH_UNARY_OPS(cbrt, {}), MATH_UNARY_OPS(exp, {}), MATH_UNARY_OPS(log, {}),
-      MATH_UNARY_OPS(log10, {}),
+      MATH_UNARY_OPS(log10, {}), MATH_UNARY_OPS(sqrt, {}),
```
js8544, line 76: There are existing sqrt functions in function_registry_arithmetis.cc.

cpp/src/gandiva/function_registry_math_ops.cc
```diff
@@ -62,11 +73,16 @@ namespace gandiva {
...
+
+      MATH_UNARY_OPS_FLOAT(sqrtf, {}), MATH_UNARY_OPS_FLOAT(cbrtf, {}),
+      MATH_UNARY_OPS_FLOAT(expf, {}), MATH_UNARY_OPS_FLOAT(logf, {}),
+      MATH_UNARY_OPS_FLOAT(log10f, {}),
```
js8544, line 80: Why not overload the existing functions?

cpp/src/gandiva/function_registry_math_ops.cc
```diff
@@ -85,9 +101,16 @@ std::vector<NativeFunction> GetMathOpsFunctionRegistry() {
       MATH_UNARY_OPS(sinh, {}), MATH_UNARY_OPS(cosh, {}), MATH_UNARY_OPS(tanh, {}),
       MATH_UNARY_OPS(cot, {}), MATH_UNARY_OPS(radians, {}),
       MATH_UNARY_OPS(degrees, {"udfdegrees"}), MATH_BINARY_SAFE(atan2, {}),
+      MATH_UNARY_OPS_FLOAT(sinf, {}), MATH_UNARY_OPS_FLOAT(cosf, {}),
+      MATH_UNARY_OPS_FLOAT(asinf, {}), MATH_UNARY_OPS_FLOAT(acosf, {}),
+      MATH_UNARY_OPS_FLOAT(tanf, {}), MATH_UNARY_OPS_FLOAT(atanf, {}),
+      MATH_UNARY_OPS_FLOAT(sinhf, {}), MATH_UNARY_OPS_FLOAT(coshf, {}),
+      MATH_UNARY_OPS_FLOAT(tanhf, {}), MATH_UNARY_OPS_FLOAT(cotf, {}),
+      MATH_BINARY_SAFE_FLOAT(atan2f, {}),
```
js8544, line 109: same as above.

cpp/src/gandiva/function_registry_math_ops.cc
```diff
@@ -85,9 +101,16 @@ std::vector<NativeFunction> GetMathOpsFunctionRegistry() {
...
 
       // decimal functions
       UNARY_SAFE_NULL_IF_NULL(abs, {}, decimal128, decimal128),
+      UNARY_SAFE_NULL_IF_NULL(absf, {}, float32, float32),
```
js8544, line 113: same as above

js8544 (commented):
Thanks for adding these functions! I have a few comments.

cpp/src/gandiva/function_registry_math_ops.cc
```diff
@@ -16,6 +16,7 @@
 // under the License.
 
 #include "gandiva/function_registry_math_ops.h"
+
```
ktf, line 19:
```suggestion
```

cpp/src/gandiva/function_registry_math_ops.cc
ktf: Sorry about that, I think I made a mess with the rebase. I will push an updated commit soon.

cpp/src/gandiva/function_registry_math_ops.cc
ktf: You mean like:

```suggestion
      MATH_UNARY_OPS_FLOAT(log10, {}),
```

we can do that, we just liked the fact there was no implicit conversion. Do you have strong feelings about that?

aalkin: The reasoning here is that we deal with limited precision data and we want to avoid automatic promotion to double from float, which has memory and performance implications. Having float-only functions allows us to spot the issues where extra precision may be used incorrectly at compilation time.

js8544:
@kou Could you have a look at this PR?

cpp/src/gandiva/function_registry_math_ops.cc
js8544: On second thought I think appending an "f" to function names makes sense. Since the new functions differ from existing ones only by their return types but not input types, I agree that they should not have the same names.

kou:
This pull request includes multiple logical changes:

> * Additional native functions for float32 in Gandiva
> * Extend bitwise operations for more int types
> * Adapt to upstream trigonometric function definitions
> * Prevent clash with c++ variant

Could you split this pull request to multiple small pull requests for each logical change?
Could you also split #34068 to corresponding subissues like #33982?

ktf:
Ok, I have started the splitting, starting with https://github.com/apache/arrow/issues/34117. Could we proceed with that one so that I get a grasp of the whole procedure, before I do the rest?

kou:
It makes sense.

cpcloud:
Is this still a viable PR?

kou:
OK.

I close this for now. We can restart this with multiple smaller PRs.

PR #34079, opened 2023-02-08
GH-34077: [Go] Implement RunEndEncoded Scalar

### LINKED ISSUE #34077: [Go] Implement RunEndEncoded scalar

RunEndEncoded arrays are implemented, now we need to implement scalar run-end-encoded values

### Component(s)

Go

zeroshade:
Implementing RunEndEncoded scalar values

### Are these changes tested?
Unit tests are added

* Closes: #34077

### DIFF
--- a/go/arrow/array/util.go
+++ b/go/arrow/array/util.go
@@ -440,6 +440,23 @@ func (n *nullArrayFactory) create() *Data {
 			childData[i] = n.createChild(dt, i, n.len)
 			defer childData[i].Release()
 		}
+	case *arrow.RunEndEncodedType:
+		bldr := NewBuilder(n.mem, dt.RunEnds())
+		defer bldr.Release()
+
+		switch b := bldr.(type) {
+		case *Int16Builder:
+			b.Append(int16(n.len))
+		case *Int32Builder:
+			b.Append(int32(n.len))
+		case *Int64Builder:
+			b.Append(int64(n.len))
+		}
+
+		childData[0] = bldr.newData()
+		defer childData[0].Release()
+		childData[1] = n.createChild(dt.Encoded(), 1, 1)
+		defer childData[1].Release()
 	case arrow.UnionType:
 		bufs[0].Release()
 		bufs[0] = nil
--- a/go/arrow/scalar/nested.go
+++ b/go/arrow/scalar/nested.go
@@ -20,7 +20,6 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"strings"
 
 	"github.com/apache/arrow/go/v12/arrow"
 	"github.com/apache/arrow/go/v12/arrow/array"
@@ -749,12 +748,11 @@ func NewDenseUnionScalar(v Scalar, code arrow.UnionTypeCode, dt *arrow.DenseUnio
 type RunEndEncoded struct {
 	scalar
 
-	RunLength int64
-	Value     Scalar
+	Value Scalar
 }
 
-func NewRunEndEncodedScalar(v Scalar, runLength int64, dt *arrow.RunEndEncodedType) *RunEndEncoded {
-	return &RunEndEncoded{scalar: scalar{dt, true}, RunLength: runLength, Value: v}
+func NewRunEndEncodedScalar(v Scalar, dt *arrow.RunEndEncodedType) *RunEndEncoded {
+	return &RunEndEncoded{scalar: scalar{dt, v.IsValid()}, Value: v}
 }
 
 func (s *RunEndEncoded) Release() {
@@ -779,8 +777,13 @@ func (s *RunEndEncoded) Validate() (err error) {
 	}
 
 	if s.Type.ID() != arrow.RUN_END_ENCODED {
-		err = fmt.Errorf("%w: run-end-encoded scalar should not have type %s",
+		return fmt.Errorf("%w: run-end-encoded scalar should not have type %s",
 			arrow.ErrInvalid, s.Type)
+	}
+
+	if !arrow.TypeEqual(s.Value.DataType(), s.Type.(*arrow.RunEndEncodedType).Encoded()) {
+		return fmt.Errorf("%w: run-end-encoded scalar value type %s does not match type %s",
+			arrow.ErrInvalid, s.Value.DataType(), s.Type)
 	}
 	return
 }
@@ -789,12 +792,11 @@ func (s *RunEndEncoded) ValidateFull() error { return s.Validate() }
 
 func (s *RunEndEncoded) equals(rhs Scalar) bool {
 	other := rhs.(*RunEndEncoded)
-	return s.RunLength == other.RunLength && Equals(s.Value, other.Value)
+	return Equals(s.Value, other.Value)
 }
 
 func (s *RunEndEncoded) String() string {
-	return fmt.Sprintf("{run_length=%d, value=%v}",
-		s.RunLength, s.Value)
+	return s.Value.String()
 }
 
 func (s *RunEndEncoded) CastTo(to arrow.DataType) (Scalar, error) {
@@ -806,11 +808,14 @@ func (s *RunEndEncoded) CastTo(to arrow.DataType) (Scalar, error) {
 		return s, nil
 	}
 
-	if to.ID() == arrow.STRING {
-		return NewStringScalar("[" +
-			strings.Repeat(s.Value.String(), int(s.RunLength)) + "]"), nil
+	if otherREE, ok := to.(*arrow.RunEndEncodedType); ok {
+		sc, err := s.Value.CastTo(otherREE.Encoded())
+		if err != nil {
+			return nil, err
+		}
+
+		return NewRunEndEncodedScalar(sc, otherREE), nil
 	}
 
-	return nil, fmt.Errorf("%w: cannot convert non-nil run-end-encoded scalar to type %s",
-		arrow.ErrInvalid, to)
+	return s.Value.CastTo(to)
 }
--- a/go/arrow/scalar/nested.go
+++ b/go/arrow/scalar/nested.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/apache/arrow/go/v12/arrow"
 	"github.com/apache/arrow/go/v12/arrow/array"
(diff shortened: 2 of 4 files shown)


zeroshade:
CC @felipecrv

go/arrow/scalar/nested.go
```diff
@@ -744,3 +745,72 @@ func (s *DenseUnion) ChildValue() Scalar { return s.Value }
...
+	return s.RunLength == other.RunLength && Equals(s.Value, other.Value)
+}
+
+func (s *RunEndEncoded) String() string {
+	return fmt.Sprintf("{run_length=%d, value=%v}",
+		s.RunLength, s.Value)
+}
+
+func (s *RunEndEncoded) CastTo(to arrow.DataType) (Scalar, error) {
+	if !s.Valid {
```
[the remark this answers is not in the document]
lidavidm: that said it's at least consistent

go/arrow/scalar/scalar.go
```diff
@@ -664,6 +671,28 @@ func GetScalar(arr arrow.Array, idx int) (Scalar, error) {
...
+
+		runEnds := arr.RunEndsArr()
+		switch ends := runEnds.(type) {
+		case *array.Int16:
+			runLength = int64(ends.Value(physicalIndex))
+		case *array.Int32:
+			runLength = int64(ends.Value(physicalIndex))
+		case *array.Int64:
+			runLength = int64(ends.Value(physicalIndex))
+		}
```
lidavidm, line 691: Apologies if this discussion was already hashed out, but does it make sense to have a run-end-encoded scalar? Wouldn't 'scalar at logical index N' just be the scalar of the underlying type?

lidavidm: I suppose at least here, we do adjust the runLength for the logical offset

go/arrow/scalar/scalar_test.go
```diff
@@ -1416,3 +1427,56 @@ func TestUnionScalars(t *testing.T) {
...
+	defer mem.AssertSize(t, 0)
+
+	runEnds, _, _ := array.FromJSON(mem, arrow.PrimitiveTypes.Int32, strings.NewReader(`[100, 200, 300, 400, 500]`))
+	defer runEnds.Release()
+
+	values, _, _ := array.FromJSON(mem, arrow.BinaryTypes.String, strings.NewReader(`["Hello", "beautiful", "world", "of", "RLE"]`))
+	defer values.Release()
+
+	reeArray := array.NewRunEndEncodedArray(runEnds, values, 500, 0)
+	defer reeArray.Release()
```
lidavidm, line 1442: Is there a plan to implement FromJSON for REE? (Possibly using an analogous encoding to unions like `[run_length, value]`?)

go/arrow/scalar/scalar.go
zeroshade: @felipecrv did you discuss the run-end-encoded scalar with @pitrou or anyone else? @lidavidm makes a good point that there's the possibility that we don't actually need a specific RunEndEncoded scalar type.

go/arrow/scalar/scalar_test.go
zeroshade: `FromJSON` was already implemented for REE, and like dictionary arrays, REE arrays are expected to be expanded in the JSON. Processing the JSON will encode it into run-end-encoding

go/arrow/scalar/scalar.go
lidavidm: To me, this scalar makes sense as representing the physical value for a physical index, but not the logical value for a logical index (if that makes...logical sense?)

lidavidm: Of course, Arrow hasn't really differentiated like that in the past since effectively everything was a 'physical' value (even if there was sort of not really a distinction between physical and logical types). But dictionary and run-end encoding kind of changes that up (the latter more than the former), and what happens with StringView when we get there?

felipecrv: @zeroshade Antoine has suggested that I add a RunEndEncoded scalar type instead of adding hacks to make things work without it.

> To me, this scalar makes sense as representing the physical value for a physical index, but not the logical value for a logical index (if that makes...logical sense?)

If you look at the REE array as an array of runs, the scalar that is typed by the REE type is a single run.

lidavidm:
Then it seems weird to get a scalar by logical index as we do here, but changing the meaning of getting a scalar would also be inconsistent.

go/arrow/array/util.go
```diff
@@ -440,6 +440,23 @@ func (n *nullArrayFactory) create() *Data {
...
+		case *Int32Builder:
+			b.Append(int32(n.len))
+		case *Int64Builder:
+			b.Append(int64(n.len))
+		}
+
+		childData[0] = bldr.newData()
+		defer childData[0].Release()
+		childData[1] = n.createChild(dt.Encoded(), 1, 1)
+		defer childData[1].Release()
```
felipecrv, line 459: This is not related to the scalar type, right?

go/arrow/scalar/nested.go
```diff
@@ -744,3 +745,72 @@ func (s *DenseUnion) ChildValue() Scalar { return s.Value }
 func NewDenseUnionScalar(v Scalar, code arrow.UnionTypeCode, dt *arrow.DenseUnionType) *DenseUnion {
 	return &DenseUnion{scalar: scalar{dt, v.IsValid()}, TypeCode: code, Value: v}
 }
+
+type RunEndEncoded struct {
+	scalar
+
+	RunLength int64
```
felipecrv, line 752: Nitpick: I called it `run_end` in C++ and it fit more naturally with the rest of the code.

For example:
 - it's typed by the "run end" type
 - the value 0 is invalid (no run ends array should contain 0) representing the null scalar

go/arrow/scalar/nested.go
```diff
@@ -744,3 +745,72 @@ func (s *DenseUnion) ChildValue() Scalar { return s.Value }
...
+	Value     Scalar
+}
+
+func NewRunEndEncodedScalar(v Scalar, runLength int64, dt *arrow.RunEndEncodedType) *RunEndEncoded {
+	return &RunEndEncoded{scalar: scalar{dt, true}, RunLength: runLength, Value: v}
```
felipecrv, line 757: instead of passing `true` here, you can pass `runLength > 0` as `runLength == 0` is the null/invalid REE scalar

go/arrow/scalar/nested.go
```diff
@@ -744,3 +745,72 @@ func (s *DenseUnion) ChildValue() Scalar { return s.Value }
...
+			arrow.ErrInvalid, s.Type)
+	}
+	return
+}
+
+func (s *RunEndEncoded) ValidateFull() error { return s.Validate() }
+
+func (s *RunEndEncoded) equals(rhs Scalar) bool {
+	other := rhs.(*RunEndEncoded)
+	return s.RunLength == other.RunLength && Equals(s.Value, other.Value)
```
felipecrv, line 792: if one of the `RunLength`s is `0`, we shouldn't take the values into account in the comparison

go/arrow/scalar/nested.go
```diff
@@ -744,3 +745,72 @@ func (s *DenseUnion) ChildValue() Scalar { return s.Value }
...
+}
+
+func (s *RunEndEncoded) String() string {
+	return fmt.Sprintf("{run_length=%d, value=%v}",
```
felipecrv, line 796: `run_end=%d`?

go/arrow/scalar/nested.go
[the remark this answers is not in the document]
felipecrv: In some type systems (like Scala's) `Null` is a sub-type of every reference type meaning that a null reference to any  type T can be used where a T is expected (`T :> Null`). So this is logically sound.

image

go/arrow/scalar/nested.go
```diff
@@ -744,3 +745,72 @@ func (s *DenseUnion) ChildValue() Scalar { return s.Value }
...
+		s.RunLength, s.Value)
+}
+
+func (s *RunEndEncoded) CastTo(to arrow.DataType) (Scalar, error) {
+	if !s.Valid {
+		return MakeNullScalar(to), nil
+	}
+
+	if arrow.TypeEqual(s.Type, to) {
```
felipecrv, line 805: Shouldn't we try to cast the run lengths here? s.run_end_type could be int16 and to.run_end_type int32/64... which permits a cast as long as the value types also work.

go/arrow/scalar/scalar.go
```diff
@@ -908,6 +937,51 @@ func MakeArrayFromScalar(sc Scalar, length int, mem memory.Allocator) (arrow.Arr
...
+			if totalLen > math.MaxInt32 {
+				return nil, fmt.Errorf("%w: final length overflows int16 run ends", arrow.ErrInvalid)
+			}
+
+			v := int32(totalLen)
+			endBytes = (*[4]byte)(unsafe.Pointer(&v))[:]
+		case arrow.INT64:
+			if totalLen > math.MaxInt64 {
+				return nil, fmt.Errorf("%w: final length overflows int16 run ends", arrow.ErrInvalid)
+			}
```
felipecrv, line 963: This check is not needed for int64, right?

go/arrow/scalar/scalar.go
```diff
@@ -908,6 +937,51 @@ func MakeArrayFromScalar(sc Scalar, length int, mem memory.Allocator) (arrow.Arr
...
+		case arrow.INT16:
+			if totalLen > math.MaxInt16 {
+				return nil, fmt.Errorf("%w: final length overflows int16 run ends", arrow.ErrInvalid)
+			}
+
+			v := int16(totalLen)
+			endBytes = (*[2]byte)(unsafe.Pointer(&v))[:]
+		case arrow.INT32:
+			if totalLen > math.MaxInt32 {
+				return nil, fmt.Errorf("%w: final length overflows int16 run ends", arrow.ErrInvalid)
```
felipecrv, line 955: typo: s/int16/int32

go/arrow/array/util.go
zeroshade: yea, i just wanted to also update `MakeArrayOfNull` since it's "connected" but not directly related.

go/arrow/scalar/nested.go
zeroshade: that's a good point, i like that and i'll make that change.

go/arrow/scalar/nested.go
zeroshade: `scalar.Equals` already handles that portion of the comparison as a common implementation check `scalar/compare.go`. We only get to the `equals` method on a type if both sides are `Valid`. Which if i update the above to make `Valid` be `runLength > 0` means that case will be handled by the common implementation

go/arrow/scalar/scalar.go
zeroshade: Sure it is, the total len could be larger than an `int64` which is why I'm using `uint64` to hold the value

felipecrv (approved):
LGTM

go/arrow/scalar/scalar_test.go
lidavidm: I guess what I mean is, why are we still constructing the array manually here? But it doesn't really matter

lidavidm (approved):

zeroshade:
I'll merge this EOD barring any comments objecting

PR #34084, opened 2023-02-08
GH-34080: [Python] Add support for round_binary to python

### LINKED ISSUE #34080: [Python] Remove warning "Python binding for RoundBinaryOptions not exposed"

This warning has been showing up since the addition of the round_binary kernel.  I assume we simply need to add bindings.

### Component(s)

Python

westonpace:
* Closes: #34080

### DIFF
--- a/python/pyarrow/tests/test_array.py
+++ b/python/pyarrow/tests/test_array.py
@@ -714,8 +714,7 @@ def test_struct_from_arrays():
         pa.StructArray.from_arrays(
             arrays, fields, mask=pa.chunked_array([mask]))
 
-    # Non-empty array with no fields
-    # https://github.com/apache/arrow/issues/15109
+    # Non-empty array with no fields https://github.com/apache/arrow/issues/15109
     arr = pa.StructArray.from_arrays([], [], mask=mask)
     assert arr.is_null() == mask
     assert arr.to_pylist() == [None, {}, {}]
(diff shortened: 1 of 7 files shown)


python/pyarrow/tests/test_array.py
```diff
@@ -714,7 +714,8 @@ def test_struct_from_arrays():
         pa.StructArray.from_arrays(
             arrays, fields, mask=pa.chunked_array([mask]))
 
-    # Non-empty array with no fields https://github.com/apache/arrow/issues/15109
```
Fokko, line 717: Nit: Looks like an unnecessary change

Fokko (approved):

python/pyarrow/tests/test_array.py
westonpace: pylint was complaining for me.  But, oddly, it was complaining about an 80 character line limit and I thought we allowed 90.  I'll revert and see if CI passes.  Maybe I have a stray .pylintrc in my environment somewhere.

westonpace: Ah, my install of archery was out of date it seems.  I updated it and the warning went away.  I suspect we changed where we store the flake config.

westonpace: Thanks for the catch

Fokko: I believe PyArrow allows 87 chars at most. Thanks for fixing 👍🏻

wjones127 (approved):

PR #34089, opened 2023-02-09
GH-34088: [Python]: Fix typo in get_writer

### LINKED ISSUE #34088: [Python] Typo in get_writer

When trying to open an IPC **writer** on something that is not a file object the error is:

```
Unable to read from object of type: {0}
```

It should be `write` instead of `read`.

### Component(s)

Python

westonpace:
* Closes: #34088


Fokko (approved):
Great catch!

PR #34096, opened 2023-02-09
GH-34086: [C++][Parquet] Fix writing num_rows to data page v2

### LINKED ISSUE #34086: [C++][Parquet] Parquet V2 page headers have incorrect number of rows

When writing Parquet files with version 2 page headers, the `num_rows` field is incorrect.  This appears to be because in `column_writer.cc ColumnWriterImpl::BuildDataPageV2()` `num_values` is passed twice to the constructor for `DataPageV2`.  The 4th argument should be `num_rows`.

To reproduce:
```python
import pyarrow.parquet as pq
import pyarrow as pa
table = pa.table({'col0': [[1,2,3]]})
pq.write_table(table, 'bug.parquet', data_page_version="2.0")
```
Examining with parquet-cli:
```sh
% parquet-cli pages bug.parquet
Column: col0.list.item
--------------------------------------------------------------------------------
  page   type  enc  count   avg size   size       rows     nulls   min / max
  0-D    dict  S _  3       8.00 B     24 B      
  0-1    data  _ R  3       2.67 B     8 B        3        0       "1" / "3"
```
"rows" should be 1.
Rewriting the file with parquet-mr gives:
```sh
% parquet-cli pages bug-mr.parquet
Column: col0.list.element
--------------------------------------------------------------------------------
  page   type  enc  count   avg size   size       rows     nulls   min / max
  0-0    data  _ D  3       5.00 B     15 B       1        0       
```

### Component(s)

C++, Parquet

wgtmac:
### Rationale for this change

The C++ parquet writer does not correctly fill num_rows field to DataPageV2 header.

ColumnWriter keeps track of number of rows buffered in the current data page and then fills it into header of data page v2.

### Are these changes tested?

A test case has been added to make sure the data page header has been set correctly for required, optional and repeated columns.

No.
* Closes: #34086

### DIFF
--- a/cpp/src/parquet/column_writer.cc
+++ b/cpp/src/parquet/column_writer.cc
@@ -642,6 +642,7 @@ class ColumnWriterImpl {
         allocator_(properties->memory_pool()),
         num_buffered_values_(0),
         num_buffered_encoded_values_(0),
+        num_buffered_rows_(0),
         rows_written_(0),
         total_bytes_written_(0),
         total_compressed_bytes_(0),
@@ -742,9 +743,12 @@ class ColumnWriterImpl {
   // case.
   int64_t num_buffered_values_;
 
-  // The total number of stored values. For repeated or optional values, this
-  // number may be lower than num_buffered_values_.
+  // The total number of stored values in the data page. For repeated or optional
+  // values, this number may be lower than num_buffered_values_.
   int64_t num_buffered_encoded_values_;
+
+  // Total number of rows buffered in the data page.
+  int64_t num_buffered_rows_;
 
   // Total number of rows written with this ColumnWriter
   int64_t rows_written_;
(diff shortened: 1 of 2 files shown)


wgtmac:
@pitrou @wjones127 Please take a look when you have time. Thanks!

cpp/src/parquet/column_writer.cc
```diff
@@ -854,8 +858,21 @@ void ColumnWriterImpl::AddDataPage() {
...
 
+namespace {
+
+int32_t safe_cast_64_to_32(int64_t value) {
+  if (value > std::numeric_limits<int32_t>::max() ||
+      value < std::numeric_limits<int32_t>::min()) {
+    throw ParquetException("Cannot cast ", value, " to int32_t");
+  }
+  return static_cast<int32_t>(value);
+}
```
wjones127, line 872: Is this necessary? Is it even possible to write more than `std::numeric_limits<int32_t>::max()` values to a row group in parquet?

wjones127 (commented):
This looks good, although I am unsure whether it makes sense to do the checked cast.

cpp/src/parquet/column_writer.cc
wgtmac: I have removed the check. Please take a look again. Thanks @wjones127

wjones127 (approved):
Looks good. Thanks for fixing this!

etseidl:
@mapleFU the issue is with nested columns (a list column for instance).  The number of values in a given page for that column will be different from the number of rows.  The previous writer was using `num_values` for both the `num_values` and `num_rows` fields in the page header.  This PR fixes this to use `num_rows` for the `num_rows` field.

mapleFU:
@etseidl I got it, seems once a page has def-level or def-level and rep-level, it values and num_rows would be not equal. And Page Header just using num_value, would could be more or less than real storing rows.

Because the `ColumnWriter` only has `rows_written_`, so it needs a `num_buffered_rows_` to get rows it actually written.

Am I right?

etseidl:
> @etseidl I got it, seems once a page has def-level or def-level and rep-level, it values and num_rows would be not equal. And Page Header just using num_value, would could be more or less than real storing rows.
> 
> 
> 
> Because the `ColumnWriter` only has `rows_written_`, so it needs a `num_buffered_rows_` to get rows it actually written.
> 
> 
> 
> Am I right?

Yes!

PR #34099, opened 2023-02-09
GH-34098: [Python][Docs] Expand dataset method docstrings

### LINKED ISSUE #34098: [Python][Docs] Fix Dataset docstrings

Many `Dataset.from_{table,fragment,batches}` don't have a docstring, or they refer to `Scanner.from_table`. It would be better to plug in the actual docstring in there.

### Component(s)

Python

Fokko:
Many `Dataset.from_{table,fragment,batches}` don't have a docstring, or they refer to `Scanner.from_table`. It would be better to plug in the actual docstring in there.

* Closes: #34098


python/pyarrow/_dataset.pyx
```diff
@@ -2647,9 +2551,11 @@ cdef class Scanner(_Weakrefable):
     @staticmethod
     def from_batches(source, *, Schema schema=None, object columns=None,
                      Expression filter=None, int batch_size=_DEFAULT_BATCH_SIZE,
+                     int batch_readahead=_DEFAULT_BATCH_READAHEAD,
```
Fokko, line 2554: Added the missing arguments

python/pyarrow/_dataset.pyx
```diff
@@ -2393,53 +2434,6 @@ cdef class Scanner(_Weakrefable):
 
     A scanner is the class that glues the scan tasks, data fragments and data
     sources together.
-
-    Parameters
```
Fokko, line 2397: I removed these since the default constructor only accepts a pointer to another dataset, and doesn't accept these arguments.

jorisvandenbossche:
@Fokko Thanks for the PR! Generally it looks great, but in practice it seems the `.format(..)` might not work for docstrings? If I test out this branch, I get empty docstrings for eg `pyarrow.dataset.Dataset.to_batches`

In pandas, they use a `@doc` decorator that does this, I assume to overcome this limitation (https://github.com/pandas-dev/pandas/blob/7d545f0849b8502974d119684bef744382cb55be/pandas/util/_decorators.py#L340-L398, it can do a lot more than just filling in some parts, so it's more complicated than what we would need). But I don't know if this would work in cython code. This would be used like

```
class Scanner:

    @doc(_scanner_arguments_doc)
    def to_batches(..):
        """
        Read the dataset as materialized record batches.

        Parameters
        ----------
        {0}
        """
```

Instead of calling the `.format` inline (and the `@doc` decorator basically does that under the hood)


In our own parquet module we use the simpler approach of afterwards assigning `__doc__`:

https://github.com/apache/arrow/blob/ddfa8eed9b188fcc7b38767d1858c2588c588f05/python/pyarrow/parquet/core.py#L3019-L3029

Similarly like that, you could also leave the docstrings as you updated them in this PR, but do the string interpolation as a next step:

```
class Scanner:

    @doc(_scanner_arguments_doc)
    def to_batches(..):
        """
        Read the dataset as materialized record batches.

        Parameters
        ----------
        {0}
        """
    ...

Scanner.to_batches.__doc__ = Scanner.to_batches.__doc__.format(_scanner_arguments_doc)
```

Fokko:
@jorisvandenbossche Thanks for the pointers. I'm quite comfortable with Python, and the examples point to `.py` files. For some reason, this doesn't seem to work with the `.pyx` files. Let me take another stab at it.

Fokko:
```
==================================== ERRORS ====================================
________________________ ERROR collecting test session _________________________
opt/conda/envs/arrow/lib/python3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1014: in _gcd_import
    ???
<frozen importlib._bootstrap>:991: in _find_and_load
    ???
<frozen importlib._bootstrap>:975: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:671: in _load_unlocked
    ???
opt/conda/envs/arrow/lib/python3.8/site-packages/_pytest/assertion/rewrite.py:168: in exec_module
    exec(co, module.__dict__)
opt/conda/envs/arrow/lib/python3.8/site-packages/pyarrow/conftest.py:103: in <module>
    import pyarrow.dataset  # noqa
opt/conda/envs/arrow/lib/python3.8/site-packages/pyarrow/dataset.py:23: in <module>
    from pyarrow._dataset import (  # noqa
pyarrow/_dataset.pyx:561: in init pyarrow._dataset
    ???
E   AttributeError: attribute '__doc__' of 'method_descriptor' objects is not writable
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
=============================== 1 error in 0.95s ===============================
```

Line 561:
```python
Dataset.scanner.__doc__ = Dataset.scanner.__doc__.format(_scanner_arguments_doc)
```

Fokko:
I don't think that the decorator will work, it is also overwriting the `__doc__` attribute: https://github.com/pandas-dev/pandas/blob/7d545f0849b8502974d119684bef744382cb55be/pandas/util/_decorators.py#L383-L390

jorisvandenbossche:
Yes, so the issues with not being able to write to `__doc__` is a general CPython issue. Antoine actaully opened an issue about it (so I suppose we tried this before ..): https://github.com/python/cpython/issues/91309

I suppose just copying the docstring around as you did now is the best we can do ..

Fokko:
Ah, great catch on the cpython issue. I'm also not super happy about copying it, but I think updating shouldn't be the biggest issue. At least there are checks that make sure that the arguments are documented. I think it is worth copying the docstrings to make it more explicit, and therefore friendlier, to the end user.

python/pyarrow/_dataset.pyx
```diff
@@ -348,63 +479,297 @@ cdef class Dataset(_Weakrefable):
...
+            batch_size=batch_size,
+            batch_readahead=batch_readahead,
+            fragment_readahead=fragment_readahead,
+            fragment_scan_options=fragment_scan_options,
+            use_threads=use_threads,
+            memory_pool=memory_pool
+        ).head(num_rows)
+
+    def count_rows(self,
+                   object columns=None,
```
jorisvandenbossche, line 700: For count_rows, doing a column selection doesn't really make sense? (it shouldn't influence the result) 
Maybe we can keep it in `**kwargs` to be sure, but at least we can maybe limit its section in the docstring here

(starting to make special cases of course makes it harder to maintain the repeated docstrings ...)

jorisvandenbossche:
Yes, certainly agree that some duplication is worth the improved usability of those docstrings.

I do wonder if we might want to go for some middle ground, and add it to the most used methods (like to_table, to_batches, and some others). With methods like `count_rows` and `head` you probably typically won't use those keywords, so in those methods referring to the docstring of another method might be good enough.

python/pyarrow/_dataset.pyx
Fokko: I'd rather strip away the columns that don't make sense for a special operation. The problem is that the `*kwargs` also need to have a docstring, which would have a reference to another function, but then in PyCharm it cannot infer the arguments from the docstring (and you have to go back to the docs).

Columns make sense in a SQL case, where `count(col)` will only count the valid columns, instead of a `count(1)` like it is now :)

Fokko: @jorisvandenbossche Sorry for the long wait, I just removed the `count` for `count_rows`. I think we're okay with maintaining the docstrings 👍🏻

Fokko:
@jorisvandenbossche any further thoughts on this?

jorisvandenbossche (approved):
Thanks for the ping, let's go with this!

Fokko:
Awesome, thanks @jorisvandenbossche 🚀

PR #34100, opened 2023-02-09
GH-34094: [C++] Increase Boost minimum version for clang >= 16

### LINKED ISSUE #34094: [C++] Boost minimum version needs increasing for clang16 builds
Although the bundled boost version was increased in #33890, the clang16 build is still failing on CRAN because the machine it's being tested on has an older version already installed, and so the bundled version is not being used, and the build is failing.

We need to set the minimum version of Boost to a certain value if Arrow C++ is being compiled with clang16 in `arrow/cpp/cmake_modules/ThirdpartyToolchain.cmake`.

Currently the minimum value is determined by this block: https://github.com/apache/arrow/blob/4a5fadbb7200ae71a151841bac3cbf35dac10258/cpp/cmake_modules/ThirdpartyToolchain.cmake#L1060-L1064

assignUser:
### Rationale for this change

Build fails with older Boost on Clang >= 16

No our dependency management will download the required Boost version if it is not installed on the system.
* Closes: #34094

### DIFF
--- a/cpp/cmake_modules/ThirdpartyToolchain.cmake
+++ b/cpp/cmake_modules/ThirdpartyToolchain.cmake
@@ -1057,8 +1057,8 @@ macro(build_boost)
   set(BOOST_VENDORED TRUE)
 endmacro()
 
-if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND CMAKE_CXX_COMPILER_VERSION
-                                              VERSION_GREATER 15)
+if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER
+                                              15)
   # GH-34094 Older versions of Boost use the deprecated std::unary_function in
   # boost/container_hash/hash.hpp and support for that was removed in clang 16
   set(ARROW_BOOST_REQUIRED_VERSION "1.81")
--- a/cpp/cmake_modules/ThirdpartyToolchain.cmake
+++ b/cpp/cmake_modules/ThirdpartyToolchain.cmake
@@ -1057,7 +1057,11 @@ macro(build_boost)
   set(BOOST_VENDORED TRUE)
 endmacro()
 
-if(ARROW_BUILD_TESTS)
+if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND CMAKE_CXX_COMPILER_VERSION
+                                              VERSION_GREATER_EQUAL 16)
+  # GH-34094 we require newer Boost versions for Clang 16 or newer 
+  set(ARROW_BOOST_REQUIRED_VERSION "1.81.0")
+elseif(ARROW_BUILD_TESTS)
   set(ARROW_BOOST_REQUIRED_VERSION "1.64")
 else()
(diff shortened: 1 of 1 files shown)


thisisnic:
Thanks for updating this!  Looks like there's a linter error which needs fixing, but aside from that, I tried this locally in a container, and it appears to be working - the Arrow C++ build now fails when being built using clang16 with `Boost_SOURCE=SYSTEM` and an older version of Boost, but uses the bundled version with `Boost_SOURCE=AUTO` or `Boost_SOURCE=BUNDLED`.

assignUser:
>  now fails when being built using clang16 with Boost_SOURCE=SYSTEM and an older version of Boost, but uses the bundled version with Boost_SOURCE=AUTO or Boost_SOURCE=BUNDLED.

Nice, that's the expected result :tada:

cpp/cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -1057,7 +1057,11 @@ macro(build_boost)
   set(BOOST_VENDORED TRUE)
 endmacro()
 
-if(ARROW_BUILD_TESTS)
+if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND CMAKE_CXX_COMPILER_VERSION
+                                              VERSION_GREATER_EQUAL 16)
+  # GH-34094 we require newer Boost versions for Clang 16 or newer
```
nealrichardson, line 1062: I'd add another comment line about how older boost versions use the long-deprecated std::unary_function in boost/container_hash/hash.hpp and that support has been removed from clang 16. That way we know why this is here and under what conditions we can remove it.

assignUser: AH yes that's a good clarification, thanks

thisisnic (approved):

westonpace (approved):
I haven't really been following this discussion but this change seems safe to me.  If someone is using clang 16 then either they are on a very new OS (which probably has a more recent version of boost) or they are advanced enough they can get a newer boost and comfortable with using newer libs.

cpp/cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -1057,7 +1057,12 @@ macro(build_boost)
   set(BOOST_VENDORED TRUE)
 endmacro()
 
-if(ARROW_BUILD_TESTS)
+if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND CMAKE_CXX_COMPILER_VERSION
+                                              VERSION_GREATER_EQUAL 16)
```
kou, line 1061: We can't use `VERSION_GREATER_EQUAL` for now because it's available since CMake 3.7:
https://cmake.org/cmake/help/latest/command/if.html#version-greater-equal

nealrichardson: How about this then?

```suggestion
                                              VERSION_GREATER 15)
```

kou: We can use it.

kou (approved):
+1

Issue #33976, opened 2023-02-01
[Python] Add low-level pyarrow bindings for Acero


jorisvandenbossche:
The Acero query engine is one of the few parts of the Arrow C++ libraries that is not directly exposed in the pyarrow bindings. While you can already run queries using Acero through a substrait plan (`pyarrow.substrait.run_query`; but that requires first building that plan, and limits to what Acero supports of Substrait), and Acero is already used under the hood in some Dataset methods (`pyarrow.dataset.Dataset.join/sort_by`, but that doesn't currently allow building up a full query), there are some benefits of having direct, low-level bindings:

* It would be useful for direct users (or downstream packages) of pyarrow that don't want / need to go through substrait (yet another thing to learn / implement)
* It would make it easier for developers to test new Acero functionality from python
* You can use features of Acero that are not (yet) available in substrait
* It could (potentially) give an interface for people developing custom exec nodes
* It would be useful for people that want to do things "right now" (e.g. 12.0.0) instead of waiting for Substrait (tooling) to mature

The idea would be to add a new `pyarrow.acero` module, with low level bindings for the Acero execution engine. The goal would _not_ be to build a complete user-friendly query interface (like ibis or dplyr), but provide rather direct access to build up a query with the Declaration interface. Minimum abilities would be to construct a plan, inspect plans created through other means (eg from substrait), and execute a plan (to_table, to_reader, write). An initial goal would be to be able to use this interface ourselves in the places that currently have custom cython usage of the exec plan such as Dataset.join (i.e. refactoring `_exec_plan.pyx`).

Non-exhaustive list of subtasks:

- [ ] Basic API to create a query plan / declaration (including the options classes)
- [ ] Replace the internal usage of execplan in Dataset join/sort_by methods
- [ ] Improve usage of compute functions with (field) expressions
- [ ] Ability to use registered UDFs in the query (specify the function registry)
- [ ] Basic bindings for ExecPlan object? (not create it directly, but be able to inspect it when created from declaration or through substrait)
- [ ] (enable using custom exec nodes?)

thisisnic:
This issue has been marked as stale because it has had no activity in the past 2 years. Please remove the stale label or comment below, or this issue will be closed in 14 days.

PR #34102, opened 2023-02-09
GH-33976: [Python] Initial bindings for acero Declaration and ExecNodeOptions classes

jorisvandenbossche:
First step for GH-33976, adding basic bindings for the different ExecNodeOptions classes and the Declaration class to combine those in a query.

Some notes on what is and what is not included in this PR:

* For source nodes, didn't expose the generic `SourceNodeOptions` et al, only the concrete `TableSourceNodeOptions` (should probably also add `RecordBatchReaderSourceNodeOptions`)
* Didn't yet expose any sink nodes. The table sink is implicitly used by `Declaration.to_table()`, and given that there is currently no explicit API to manually convert to ExecPlan and execute it, explicit table sink node bindings didn't seem necessary. 
* Also didn't yet expose the order_by sink node, because this requires a custom sink when collecting as a Table, and it's not directly clear how this is possible with the Declaration interface. This requires https://github.com/apache/arrow/issues/34248 to be fixed first.
* Leaving dataset-based scan and write nodes for a follow-up PR
* Basic class for `Declaration` with a `to_table` method to execute the plan and consume it into a Table, and a `to_reader()` to get a RecordBatchReader (could also further add a `to_batches()` method)

--

* Issue: #33976


python/pyarrow/_acero.pyx
```diff
@@ -0,0 +1,459 @@
...
+            resources).
+
+        Returns
+        -------
+        pyarrow.Table
+        """
+        cdef:
+            shared_ptr[CTable] c_table
+
+        c_table = GetResultValue(DeclarationToTable(self.decl, use_threads))
```
jorisvandenbossche, line 458: Should I use `DeclarationToTable` or `DeclarationToTableAsync` here?

python/pyarrow/_acero.pyx
```diff
@@ -0,0 +1,459 @@
...
+            vector[CFieldRef] c_left_keys
+            vector[CFieldRef] c_right_keys
+            vector[CFieldRef] c_left_output
+            vector[CFieldRef] c_right_output
+
+        # join type
+        # try:
+        #     c_join_type = self._join_type_map[join_type]
+        # except KeyError:
+        #     raise ValueError("Unsupported join type")
```
jorisvandenbossche, line 250: I am getting compile errors with this more compact option, so wrote it out with if/elses below (but will still look into the errors)

danepitkin:
Awesome stuff! I'm not too familiar with Acero, but looks like a good starting foundation for pyarrow.

python/pyarrow/_acero.pyx
westonpace: Use `DeclarationToTable`.  We have not historically exposed async methods to python and I don't see much benefit in doing so.

python/pyarrow/_acero.pyx
```diff
@@ -0,0 +1,459 @@
...
+    producing new batches.
+
+    The "project" operation rearranges, deletes, transforms, and
+    creates columns. Each output column is computed by evaluating
+    an expression against the source record batch.
+
+    Parameters
+    ----------
+    expressions : list of pyarrow.compute.Expression
+        List of expressions to evaluate against the source batch.
```
westonpace, line 150: These must be scalar expressions.  That is, every function called in the expression must meet the following criteria:

 * The expression is stateless (the value of one row is independent of the value of all other rows)
 * The expression returns a value for each input row
 * The expression must accept arrays/scalars as inputs and return an array as output

This is probably mostly important when users are creating UDFs (also, it might be nice to mention that UDFs can be used here provided they meet the criteria)

python/pyarrow/_acero.pyx
```diff
@@ -0,0 +1,459 @@
...
+    names : list of str
+        List of names for each of the ouptut columns (same length as
+        `expressions`). If `names` is not provided, the string
+        representations of exprs will be used.
+    """
```
westonpace, line 155: The output of this node is not clear, though it may be intuitive enough.  This node will output batches with the same length as the input batches that have one column for each expression.

python/pyarrow/_acero.pyx
```diff
@@ -0,0 +1,459 @@
...
+    ----------
+    aggregates : list of tuples
+        Aggregations which will be applied to the targetted fields.
+        Specified as a list of tuples, where each tuple is one aggregation
+        specification and consists of: aggregation column name followed
+        by function name, aggregation function options object and the
+        output field name.
+        The column name can be a string, an empty list or a list of
+        column names, for unary, nullary and n-ary aggregation functions
+        respectively.
```
westonpace, line 213: I don't know if it helps to have some examples here.  Similar to group_by I suspect this will be a little confusing.

python/pyarrow/_acero.pyx
```diff
@@ -0,0 +1,459 @@
...
+    keys : list, optional
```
westonpace, line 214: A list of what?

python/pyarrow/_acero.pyx
```diff
@@ -0,0 +1,459 @@
...
+
+class HashJoinNodeOptions(_HashJoinNodeOptions):
+    """
+    Make a node which implements join operation using hash join strategy.
+
+    Parameters
+    ----------
+    join_type : str
+        Type of join. One of "left semi", "right semi", "left anti",
+        "right anti", "inner", "left outer", "right outer", "full outer".
```
westonpace, line 315: Is there a default?  I wonder if it makes sense to default to `inner`?

python/pyarrow/_acero.pyx
```diff
@@ -0,0 +1,459 @@
...
+                    _true,
+                    <c_string>tobytes(output_suffix_for_left),
+                    <c_string>tobytes(output_suffix_for_right)
+                )
+            )
+
+
+class HashJoinNodeOptions(_HashJoinNodeOptions):
+    """
+    Make a node which implements join operation using hash join strategy.
```
westonpace, line 309: This seems a bit sparse.  Maybe...

Make a node which joins two inputs together based on common keys.  Each row from the left input will be matched with rows from the right input where the keys are equal.  Null keys do not match other null keys.  If any key is null then that row will not match.  Which columns are output, and the behavior of rows that do not match, will depend on the join type.

inner: for each row in the left input there will be one output row per matching row in the right input.  If there are no matching rows in the right input then there will be no output row for that input row.  The output columns will be the left_keys, followed by the right_keys (these will be identical to the left_keys), followed by the left_output, followed by the right_output.

left outer: behaves the same as an inner join.  However, if there are no matching rows in the right input then there will be one output row that contains null for each column in right_keys and right_output.  A row in the right input, that has no matching row in the left input, will not be output.  The output columns will be identical to an inner join.

right outer: for each row in the right input there will be one output row per matching row in the left input.  If there are no matching rows in the left input then there will be one output row containing null in left_keys and left_output.  A row in the left input, that has no matching row in the right input, will not be output.  The output columns will be identical to an inner join.

full outer: behaves the same as a left outer join.  However, if there is a row in the right input, that has no matching row in the left input, then there will be one output row containing null for each column in left_keys and left_output.  The output columns will be identical to an inner join.

left semi: for each row in the left input there will be exactly one output row, if and only if, there is a matching row in the right input.  The output columns will be the left_keys followed by the left_output.

right semi: for each row in the right input there will be exactly one output row, if and only if, there is a matching row in the left input.  The output columns will be the right_keys followed by the right_output.

left anti: for each row in the left input there will be exactly one output row, if and only if, there is not a matching row in the right input.  The output will be the same as left semi.

right anti: for each row in the right input there will be exactly one output row, if and only if, there is not a matching row in the left input.  The output will be the same as right semi.

python/pyarrow/_acero.pyx
```diff
@@ -0,0 +1,459 @@
...
+
+    A Declaration represents an unconstructed ExecNode, and potentially
+    more since its inputs may also be Declarations or when constructed
+    with ``from_sequence``.
+
+    Parameters
+    ----------
+    factory_name : str
+        The ExecNode factory name, such as "table_source", "filter",
+        "project" etc.
```
westonpace, line 360: I think users may be confused by this since it might not be clear to them why a "factory name" is a necessary concept at all.

python/pyarrow/_acero.pyx
```diff
@@ -0,0 +1,459 @@
...
+        return frombytes(GetResultValue(DeclarationToString(self.decl)))
+
+    def __repr__(self):
+        return "<pyarrow.acero.Declaration>\n{0}".format(str(self))
+
+    def to_table(self, use_threads=True):
+        """
+        Run the declaration and collect the results into a table.
+
+        This method will implicitly add a sink node to the declaration
```
westonpace, line 438: Do users care about the existence of a sink node?  I suppose they might if they are creating custom nodes and aware of such things.

westonpace (commented):
Initial thoughts

python/pyarrow/_acero.pxd
```diff
@@ -0,0 +1,44 @@
...
+from pyarrow.includes.libarrow cimport *
+
+
+cdef class ExecNodeOptions(_Weakrefable):
+    cdef:
+        shared_ptr[CExecNodeOptions] wrapped
+
+    cdef const CExecNodeOptions* get_options(self) except NULL
+    cdef void init(self, const shared_ptr[CExecNodeOptions]& sp)
+    cdef inline shared_ptr[CExecNodeOptions] unwrap(self)
```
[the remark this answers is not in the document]
jorisvandenbossche: Ah, that's a good reminder that I still need to check where to add `nogil` in general. I suppose for the options it is less critical, but certainly for executing with eg DeclarationToTable I should certainly put that in a nogil block.

python/pyarrow/_acero.pyx
```diff
@@ -0,0 +1,459 @@
...
+from pyarrow.compute import field
+
+# Initialize()  # Initialise support for Datasets in ExecPlan
+
+
+cdef class ExecNodeOptions(_Weakrefable):
+    __slots__ = ()  # avoid mistakingly creating attributes
+
+    cdef const CExecNodeOptions* get_options(self) except NULL:
+        return self.wrapped.get()
```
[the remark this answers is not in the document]
jorisvandenbossche: This is the pattern we had for the compute options classes, so started with that. But in the end it's not actually being used here, so can remove it then.

python/pyarrow/_acero.pyx
```diff
@@ -0,0 +1,459 @@
...
+    cdef void init(self, const shared_ptr[CExecNodeOptions]& sp):
+        self.wrapped = sp
+
+    cdef inline shared_ptr[CExecNodeOptions] unwrap(self):
+        return self.wrapped
+
+    # def __repr__(self):
+    #     type_name = self.__class__.__name__
+    #     # Remove {} so we can use our own braces
+    #     string_repr = frombytes(self.get_options().ToString())[1:-1]
```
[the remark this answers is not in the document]
jorisvandenbossche: Sorry, I should have added some comments to the WIP things. This was a reminder to see if we want a repr. But I agree that for the option classes here, that's not very important. Typically you only create them just to directly pass to a Declaration constructor (inspecting options created by someone else seems not a typical use case here)

jorisvandenbossche:
@westonpace thanks for the review! Question for the (useful) comments with ideas to clarify the docstrings: I mostly copied things from existing doc comments in the C++ code or explanations from the C++ user guide. While updating here based on your feedback, would it be useful to directly also update it in the C++ souce? (but maybe you prefer to keep the doc comments in options.h/exec_plan.h shorer?)

westonpace:
> While updating here based on your feedback, would it be useful to directly also update it in the C++ source?

Yes, if you don't mind.  The lengthy join explanation was mostly just because I was recently doing some Substrait work and so I had the detailed explanation still loaded up in short term memory.  I think it would be quite helpful to add to the C++ documentation as well.

> (but maybe you prefer to keep the doc comments in options.h/exec_plan.h shorter?)

I don't mind [long](https://github.com/apache/arrow/blob/main/cpp/src/arrow/util/async_util.h#L41-L65) [doc](https://github.com/apache/arrow/blob/main/cpp/src/arrow/util/async_generator.h#L40-L69) [comments](https://github.com/apache/arrow/blob/main/cpp/src/arrow/dataset/dataset_writer.h#L60-L83) when the situation warrants. :laughing:

python/pyarrow/_acero.pyx
```diff
@@ -0,0 +1,459 @@
...
+    #     return f"{type_name}({string_repr})"
+
+    # def __eq__(self, FunctionOptions other):
+    #     return self.get_options().Equals(deref(other.get_options()))
```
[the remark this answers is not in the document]
jorisvandenbossche: @westonpace the ExecNodeOptions class in C++ does not define a `operator==` overload, so I don't think it allows such equality check? (or where you thinking about pure pointer equality?) 

Anyway, that's also not very important here (if we would like to add support for pickling those objects (and declarations / execplans in general, that might become more important, but that is definitely out of scope for this PR).

python/pyarrow/_acero.pyx
jorisvandenbossche: Given that we will have multiple possible sinks at some point, it might be useful to be aware of the general pattern (and also for people that might read the C++ user guide alongside this, I think it is OK to be explicit here)

python/pyarrow/_acero.pyx
jorisvandenbossche: And the factory name is needed for the registry? (so that you can register your own custom nodes) 

Do you suggest using a different name here? (eg just "name" instead of "factory name") Or to explain it better what this "factory name" is?

python/pyarrow/_acero.pyx
jorisvandenbossche: I don't have a strong opinion about a default. I think it's fine for a low level interface to require to specify it. On the C++ side there is one simplified constructor signature that defaults to inner, the others are explicit.

python/pyarrow/_acero.pyx
```diff
@@ -0,0 +1,459 @@
...
+        else:
+            self.wrapped.reset(
+                new CProjectNodeOptions(c_expressions)
+            )
+
+
+class ProjectNodeOptions(_ProjectNodeOptions):
+    """
+    Make a node which executes expressions on input batches,
+    producing new batches.
```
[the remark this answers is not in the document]
jorisvandenbossche: This is from 

https://github.com/apache/arrow/blob/e8e50a3ca96ec8eb2f2beb383cf9005ce1c93486/cpp/src/arrow/compute/exec/options.h#L186

and it also _does_ produce new batches (a batch of the same length for each input batch). But maybe most nodes produce batches, so that might not be the most helpful distinguishing descriptor? 

How does "producing batches of the same length with new columns" sound?

python/pyarrow/_acero.pyx
```diff
@@ -0,0 +1,459 @@
...
+
+    The "project" operation rearranges, deletes, transforms, and
+    creates columns. Each output column is computed by evaluating
```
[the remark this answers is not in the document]
jorisvandenbossche: I think the "transform" term more clearly highlights that you can use the existing columns to transform/create new columns. So personally I think it is fine to leave in both terms.

python/pyarrow/_acero.pyx
jorisvandenbossche: > These must be scalar expressions

Are those criteria already documented or described somewhere? We have the section about "scalar" compute functions (https://arrow.apache.org/docs/dev/cpp/compute.html#element-wise-scalar-functions), but that is also not that explicit.

jorisvandenbossche: I see we have a `IsScalarExpression` that checks for this in the code base, and essentially checks for the expression being a scalar literal, a field reference, or calls to scalar functions (or a combination of those)

python/pyarrow/_acero.pyx
jorisvandenbossche: A list of "field references", where each field reference can either be a string name, integer reference, or fieldref expression. 

Updated the documentation and added some more test coverage for this. 

Using integers (i.e. a FieldPath) does work out of the box, but given that this position is "relative" to the input of the node (i.e. output of the previous node) and not for the full plan, that's probably not something we want to encourage.

python/pyarrow/_acero.pyx
jorisvandenbossche: Given that this suggestion is a larger addition, going to leave that for a separate PR.

jorisvandenbossche:
@westonpace I did some small updates to the doc comments / user guide based on your feedback, would appreciate if you could take a look at those changes.

docs/source/cpp/streaming_execution.rst
```diff
@@ -652,8 +652,9 @@ Example of using ``table_source``
 ----------
 
 ``filter`` operation, as the name suggests, provides an option to define data filtering 
-criteria. It selects rows matching a given expression. Filters can be written using 
-:class:`arrow::compute::Expression`. For example, if we wish to keep rows where the value 
+criteria. It selects rows where the given expression evaluates to true. Filters can bewritten using
```
westonpace, line 655:
```suggestion
criteria. It selects rows where the given expression evaluates to true. Filters can be written using
```

python/pyarrow/_acero.pyx
```diff
@@ -0,0 +1,476 @@
...
+from pyarrow.includes.common cimport *
+from pyarrow.includes.libarrow cimport *
+from pyarrow.includes.libarrow_dataset cimport *
+from pyarrow.lib cimport (Table, check_status, pyarrow_unwrap_table, pyarrow_wrap_table,
+                          RecordBatchReader)
+from pyarrow.lib import frombytes, tobytes
+from pyarrow._compute cimport Expression, FunctionOptions, _ensure_field_ref, _true
+from pyarrow.compute import field
+
+# Initialize()  # Initialise support for Datasets in ExecPlan
```
westonpace, line 34: Uncomment? (I don't remember if I asked about this last time, sorry)

python/pyarrow/_acero.pyx
```diff
@@ -0,0 +1,459 @@
...
+
+        if names is not None:
+            if len(names) != len(expressions):
+                raise ValueError("dd")
+
+            for name in names:
+                c_names.push_back(<c_string>tobytes(name))
+
+            self.wrapped.reset(
+                new CProjectNodeOptions(c_expressions, c_names)
```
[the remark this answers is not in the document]
westonpace: In retrospect, I think immutable is better. In python you can do `MyOptions(x=7, y=3)`.  So you don't really need the ability to do:

```
MyOptions options;
options.x = 7;
options.y = 3;
```

python/pyarrow/_acero.pyx
```diff
@@ -0,0 +1,476 @@
...
+    left_keys : str, Expression or list
+        Key fields from left input. Each key can be a string column name
+        or a field expression, or a list of such field references.
+    right_keys : str, Expression or list
+        Key fields from right input. See `left_keys` for details.
+    left_output : list, optional
+        Output fields passed from left input. If left and right output
+        fields are not specified, all valid fields from both left and
+        right input will be output
+    right_output : list, optional
```
westonpace, line 323: Do these need to be lists of field references?  Or can they be lists of strings?

python/pyarrow/_acero.pyx
westonpace: Yes, the factory name is currently needed for the registry.  Either you need to explain what it is and explain, in each of the options, what factory name they should use or you need to make the name a part of the options class and remove it from here.

Otherwise, a user might not (for example) know to use "aggregate" when creating `AggregateNodeOptions`

westonpace (approved):
This looks quite nice.  A few more small thoughts.

python/pyarrow/_acero.pyx
jorisvandenbossche: We could indeed make the factory name unnecessary in the API, if we attach a factory name to the Python option classes. But I think I might opt to stay closer to the C++ API here to require a factory name. But indeed will mention in each option class docstring the name of its factory.

python/pyarrow/_acero.pyx
jorisvandenbossche: Will just remove for now, will do the Datasets scan/write node option classes in a follow-up PR.

python/pyarrow/_acero.pyx
jorisvandenbossche: Yes, they can be any form of field "reference", so either a field_ref or a string (or int). Will clarify in the docstring.

jorisvandenbossche:
The integration test failure seems unrelated and seems to happen on other PRs as well, so going to merge this. 

@westonpace thanks for the review!

Issue #14870, opened 2022-12-07
[C++][Parquet] Support min_value and max_value Statistics


pitrou:
### Describe the enhancement requested

The `Statistics` [structure](https://github.com/apache/parquet-format/blob/28e6e16f6d46cf8f0eff8f95ba2ca1a3371f258c/src/main/thrift/parquet.thrift#L199-L225) in Parquet files provides two ways of specifying lower and upper bounds for a data page:
* `min` and `max` are legacy fields for compatibility with older writers, with ill-defined comparison semantics in most cases except for signed integers
* `min_value` and `max_value` are "new" fields (introduced in 2017! - see https://github.com/apache/parquet-format/commit/041708da1af52e7cb9288c331b542aa25b68a2b6 and https://github.com/apache/parquet-format/commit/bef5438990116725af041cdd8ced2bca0ed2608a) with well-defined comparison semantics depending on the logical type

Currently Parquet C++ supports only the legacy fields `min` and `max`. We should add support for reading and writing the newer ones, with the appropriate semantics on the write path.

### Component(s)

Parquet

pitrou:
cc @fatemehp @wgtmac

wgtmac:
AFAIK, the statistics in the data page header has been deprecated after column index (part of the page index) has been introduced.

CMIW, parquet cpp is in good state with the parquet spec:
- On the read path, the statistics in the ColumnChunkMetaData already consumes `min_value` & `max_value` from the thrift object if available: https://github.com/apache/arrow/blob/master/cpp/src/parquet/metadata.cc#L97
- On the write path, we have written both `min_value` and `min` for backward-compatibility: https://github.com/apache/arrow/blob/master/cpp/src/parquet/thrift_internal.h#L292

wgtmac:
> AFAIK, the statistics in the data page header has been deprecated after column index (part of the page index) has been introduced.
> 
> CMIW, parquet cpp is in good state with the parquet spec:
> 
> * On the read path, the statistics in the ColumnChunkMetaData already consumes `min_value` & `max_value` from the thrift object if available: https://github.com/apache/arrow/blob/master/cpp/src/parquet/metadata.cc#L97
> * On the write path, we have written both `min_value` and `min` for backward-compatibility: https://github.com/apache/arrow/blob/master/cpp/src/parquet/thrift_internal.h#L292

The confusing thing is that both `Statistics` and `EncodedStatistics` use `min` and `max` for both old and new meanings.

[...]

PR #34107, opened 2023-02-09
GH-34106: [C++][Parquet] Fix updating page stats for WriteArrowDictionary

### LINKED ISSUE #34106: [C++][Parquet] Fix updating page statistics for WriteArrowDictionary

Commit for this issue https://github.com/apache/arrow/issues/15042 has fixed the missing statistics, but it still has some issues. For example, if a single write has been split into more than one batch which results in several data pages, the statistics of each page is not correctly updated.

### Component(s)

C++, Parquet

wgtmac:
### Rationale for this change

`ColumnWriter::WriteArrowDictionary` has tried to update stats but has problem if a single write has been split into batches and more than one page is written.

Make sure every write of batch has updated the stats.

### Are these changes tested?

Add test case which fails without the fix.

No.
* Closes: #34106

### DIFF
--- a/cpp/src/parquet/column_writer.cc
+++ b/cpp/src/parquet/column_writer.cc
@@ -1484,6 +1484,39 @@ Status TypedColumnWriterImpl<DType>::WriteArrowDictionary(
   std::shared_ptr<::arrow::Array> dictionary = data.dictionary();
   std::shared_ptr<::arrow::Array> indices = data.indices();
 
+  auto update_stats = [&](int64_t num_chunk_levels,
+                          const std::shared_ptr<Array>& chunk_indices) {
+    // TODO(PARQUET-2068) This approach may make two copies.  First, a copy of the
+    // indices array to a (hopefully smaller) referenced indices array.  Second, a copy
+    // of the values array to a (probably not smaller) referenced values array.
+    //
+    // Once the MinMax kernel supports all data types we should use that kernel instead
+    // as it does not make any copies.
+    ::arrow::compute::ExecContext exec_ctx(ctx->memory_pool);
+    exec_ctx.set_use_threads(false);
+
+    std::shared_ptr<::arrow::Array> referenced_dictionary;
+    PARQUET_ASSIGN_OR_THROW(::arrow::Datum referenced_indices,
+                            ::arrow::compute::Unique(*chunk_indices, &exec_ctx));
+
+    // On first run, we might be able to re-use the existing dictionary
+    if (referenced_indices.length() == dictionary->length()) {
+      referenced_dictionary = dictionary;
+    } else {
+      PARQUET_ASSIGN_OR_THROW(
+          ::arrow::Datum referenced_dictionary_datum,
+          ::arrow::compute::Take(dictionary, referenced_indices,
+                                 ::arrow::compute::TakeOptions(/*boundscheck=*/false),
+                                 &exec_ctx));
+      referenced_dictionary = referenced_dictionary_datum.make_array();
+    }
+
+    int64_t non_null_count = chunk_indices->length() - chunk_indices->null_count();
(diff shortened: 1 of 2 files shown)


wgtmac:
Converted to draft because I hit another issue: https://github.com/apache/arrow/issues/14870. The C++ parquet reader does not parse column statistics correctly here: https://github.com/apache/arrow/blob/master/cpp/src/parquet/column_reader.cc#L214
```cpp
// Extracts encoded statistics from V1 and V2 data page headers
template <typename H>
EncodedStatistics ExtractStatsFromHeader(const H& header) {
  EncodedStatistics page_statistics;
  if (!header.__isset.statistics) {
    return page_statistics;
  }
  const format::Statistics& stats = header.statistics;
  if (stats.__isset.max) {
    page_statistics.set_max(stats.max);
  }
  if (stats.__isset.min) {
    page_statistics.set_min(stats.min);
  }
  if (stats.__isset.null_count) {
    page_statistics.set_null_count(stats.null_count);
  }
  if (stats.__isset.distinct_count) {
    page_statistics.set_distinct_count(stats.distinct_count);
  }
  return page_statistics;
}

```

cpp/src/parquet/column_writer.cc
```diff
@@ -1471,6 +1504,7 @@ Status TypedColumnWriterImpl<DType>::WriteArrowDictionary(
                       AddIfNotNull(rep_levels, offset));
     std::shared_ptr<Array> writeable_indices =
         indices->Slice(value_offset, batch_num_spaced_values);
+    update_stats(/*num_chunk_levels=*/batch_size, writeable_indices);
```
mapleFU, line 1507: Would `page_statistics_` be a nullptr here?

wgtmac: Yes, please check: https://github.com/apache/arrow/blob/master/cpp/src/parquet/column_writer.cc#L1069

mapleFU: Well, I think here we should:
`if (page_statistics) update_stats(/*num_chunk_levels=*/batch_size, writeable_indices);`

wgtmac: Good catch!

wgtmac:
@westonpace @wjones127 Please take a look. Thanks!

cpp/src/parquet/column_writer.cc
```diff
@@ -1484,6 +1484,39 @@ Status TypedColumnWriterImpl<DType>::WriteArrowDictionary(
...
+      PARQUET_ASSIGN_OR_THROW(
+          ::arrow::Datum referenced_dictionary_datum,
+          ::arrow::compute::Take(dictionary, referenced_indices,
+                                 ::arrow::compute::TakeOptions(/*boundscheck=*/false),
+                                 &exec_ctx));
+      referenced_dictionary = referenced_dictionary_datum.make_array();
+    }
+
+    int64_t non_null_count = chunk_indices->length() - chunk_indices->null_count();
+    page_statistics_->IncrementNullCount(num_chunk_levels - non_null_count);
```
wjones127, line 1515: Question: is the reason this isn't just set to `chunk_indices->null_count()` that the null count stats isn't meant to include cases where higher levels (parent fields) are null? Or something else?

wjones127 (approved):
Thanks for fixing this!

cpp/src/parquet/column_writer.cc
wgtmac: Yes. `chunk_indices->null_count()` is the null count for the current leaf only. `num_chunk_levels - non_null_count` also counts null values from ancestor (e.g. empty list is also considered as null but not included in the `chunk_indices->null_count()`).

wgtmac:
Gentle ping @wjones127

PR #34112, opened 2023-02-10
GH-34138: [C++][Parquet] Fix parsing stats from min_value/max_value

### LINKED ISSUE #34138: [C++][Parquet] Fix parsing stats from min_value/max_value

The code below does not check and read from stats.min_value/max_value. If reading from a parquet file where the stats apply min_value/max_value, we are unable to read any column statistics at all.
```cpp
// Extracts encoded statistics from V1 and V2 data page headers
template <typename H>
EncodedStatistics ExtractStatsFromHeader(const H& header) {
  EncodedStatistics page_statistics;
  if (!header.__isset.statistics) {
    return page_statistics;
  }
  const format::Statistics& stats = header.statistics;
  if (stats.__isset.max) {
    page_statistics.set_max(stats.max);
  }
  if (stats.__isset.min) {
    page_statistics.set_min(stats.min);
  }
  if (stats.__isset.null_count) {
    page_statistics.set_null_count(stats.null_count);
  }
  if (stats.__isset.distinct_count) {
    page_statistics.set_distinct_count(stats.distinct_count);
  }
  return page_statistics;
}
```

### Component(s)

C++, Parquet

wgtmac:
### Rationale for this change

The code below does not read from stats.min_value/max_value at all.
```cpp
// Extracts encoded statistics from V1 and V2 data page headers
template <typename H>
EncodedStatistics ExtractStatsFromHeader(const H& header) {
  EncodedStatistics page_statistics;
  if (!header.__isset.statistics) {
    return page_statistics;
  }
  const format::Statistics& stats = header.statistics;
  if (stats.__isset.max) {
    page_statistics.set_max(stats.max);
  }
  if (stats.__isset.min) {
    page_statistics.set_min(stats.min);
  }
  if (stats.__isset.null_count) {
    page_statistics.set_null_count(stats.null_count);
  }
  if (stats.__isset.distinct_count) {
    page_statistics.set_distinct_count(stats.distinct_count);
  }
  return page_statistics;
}
```

Do similar thing from parquet-mr to check and read min_value/max_value from thrift stats.

### Are these changes tested?

Some test cases fail after the fix. Fixed them to make sure it is covered.

No.

* Closes: #34138

### DIFF
--- a/cpp/src/parquet/column_reader.cc
+++ b/cpp/src/parquet/column_reader.cc
@@ -211,10 +211,15 @@ EncodedStatistics ExtractStatsFromHeader(const H& header) {
     return page_statistics;
   }
   const format::Statistics& stats = header.statistics;
-  if (stats.__isset.max) {
+  // Use the new V2 min-max statistics over the former one if it is filled
+  if (stats.__isset.max_value && stats.__isset.min_value) {
+    // TODO: check if the column_order is TYPE_DEFINED_ORDER.
+    page_statistics.set_max(stats.max_value);
+    page_statistics.set_min(stats.min_value);
+  } else if (stats.__isset.max && stats.__isset.min) {
+    // TODO: check created_by to see if it is corrupted for some types.
+    // TODO: check if the sort_order is SIGNED.
     page_statistics.set_max(stats.max);
-  }
-  if (stats.__isset.min) {
     page_statistics.set_min(stats.min);
   }
   if (stats.__isset.null_count) {
--- a/cpp/src/parquet/column_reader.cc
+++ b/cpp/src/parquet/column_reader.cc
@@ -212,15 +212,23 @@ EncodedStatistics ExtractStatsFromHeader(const H& header) {
   }
   const format::Statistics& stats = header.statistics;
   // Use the new V2 min-max statistics over the former one if it is filled
-  if (stats.__isset.max_value && stats.__isset.min_value) {
+  if (stats.__isset.max_value || stats.__isset.min_value) {
     // TODO: check if the column_order is TYPE_DEFINED_ORDER.
-    page_statistics.set_max(stats.max_value);
-    page_statistics.set_min(stats.min_value);
-  } else if (stats.__isset.max && stats.__isset.min) {
+    if (stats.__isset.max_value) {
+      page_statistics.set_max(stats.max_value);
+    }
+    if (stats.__isset.min_value) {
+      page_statistics.set_min(stats.min_value);
+    }
+  } else if (stats.__isset.max || stats.__isset.min) {
     // TODO: check created_by to see if it is corrupted for some types.
     // TODO: check if the sort_order is SIGNED.
-    page_statistics.set_max(stats.max);
-    page_statistics.set_min(stats.min);
+    if (stats.__isset.max) {
+      page_statistics.set_max(stats.max);
+    }
+    if (stats.__isset.min) {
+      page_statistics.set_min(stats.min);
+    }
   }
   if (stats.__isset.null_count) {
     page_statistics.set_null_count(stats.null_count);
--- a/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
+++ b/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
@@ -4083,6 +4083,16 @@ TEST_P(TestArrowWriteDictionary, Statistics) {
   std::vector<std::vector<std::string>> expected_min_max_ = {
       {"a", "b"}, {"b", "c"}, {"a", "d"}, {"", ""}};
 
+  const std::vector<std::vector<std::vector<std::string>>> expected_min_by_page = {
+      {{"b", "a"}, {"b", "a"}}, {{"b", "b"}, {"b", "b"}}, {{"c", "a"}, {"c", "a"}}};
+  const std::vector<std::vector<std::vector<std::string>>> expected_max_by_page = {
+      {{"b", "a"}, {"b", "a"}}, {{"c", "c"}, {"c", "c"}}, {{"d", "a"}, {"d", "a"}}};
+  const std::vector<std::vector<std::vector<bool>>> expected_has_min_max_by_page = {
+      {{true, true}, {true, true}},
+      {{true, true}, {true, true}},
+      {{true, true}, {true, true}},
+      {{false}, {false}}};
+
   for (std::size_t case_index = 0; case_index < test_dictionaries.size(); case_index++) {
     SCOPED_TRACE(test_dictionaries[case_index]->type()->ToString());
     ASSERT_OK_AND_ASSIGN(std::shared_ptr<::arrow::Array> dict_encoded,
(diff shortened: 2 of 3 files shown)


wgtmac:
@pitrou @wjones127 @westonpace Could you please take a look?

cpp/src/parquet/column_reader.cc
```diff
@@ -211,10 +211,15 @@ EncodedStatistics ExtractStatsFromHeader(const H& header) {
     return page_statistics;
   }
   const format::Statistics& stats = header.statistics;
-  if (stats.__isset.max) {
+  // Use the new V2 min-max statistics over the former one if it is filled
```
mapleFU, line 214: Previously, `page_statistics` will handle min-max separately. This patch changes it to once have all min-max, otherwise, cannot use min-max

mapleFU (commented):
Rest LGTM

cpp/src/parquet/column_reader.cc
wgtmac: Yes, page stats without either min or max is corrupted and cannot be used. Check parquet-mr for detail: https://github.com/apache/parquet-mr/blob/5290bd5e0ee5dc30db0576e2bfc6eea335c465cf/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java#L797

mapleFU: Although in parquet.thrift, min-max can exist only one. But I think handling it like this is ok

wjones127 (approved):
This looks good. @wgtmac before we merge, could you create a new issue for this bug fix? Also, if you could create issues for the TODOs as well, that would be appreciated.

westonpace:
Does this mean the following (testing my understanding)?

If min is set but max is not set then we should ignore both min and max

wjones127:
> Does this mean the following (testing my understanding)?
> 
> If min is set but max is not set then we should ignore both min and max

Yes, it does mean we will. Do you foresee that as an issue? It sounds like Java implementation takes the same approach.

wgtmac:
> Does this mean the following (testing my understanding)?
> 
> If min is set but max is not set then we should ignore both min and max

When only one of min and max exists, it usually happens when a binary value has an extreme length or a floating value has NaN. In this case, the stats provide little value and make it tricker to use.

wgtmac:
> This looks good. @wgtmac before we merge, could you create a new issue for this bug fix? Also, if you could create issues for the TODOs as well, that would be appreciated.

I have created a new issue for this PR and updated the title.

A separate issue has been created for the TODO items: https://github.com/apache/arrow/issues/34139

Thanks for the review! @wjones127

mapleFU (approved):
Rest LGTM

wgtmac:
Gentle ping @wjones127 @pitrou 

Once this gets merged, I will rebase https://github.com/apache/arrow/pull/34107 which is blocked by it.

emkornfield:
CC @fatemehp

westonpace:
> Yes, it does mean we will. Do you foresee that as an issue? It sounds like Java implementation takes the same approach.

In datasets, for row group statistics, we [recently added a check](https://github.com/apache/arrow/pull/15125) that was roughly...

```
if (is_nan(min) && is_nan(max)) {
  // Ignore statistics
} else if (is_nan(min)) {
  // Assume x <= max
} else if(is_nan(max)) {
  // Assume x >= min
} else {
  // Assume min <= x <= max
}
```

In other words, if one of min or max is NaN then we still use the other side of the equality.  I think my primary concern is to validate that is a safe assumption.  In other words, I want to make sure we aren't using garbage data in our handling of row groups.

wjones127:
@westonpace that makes sense.

> When only one of min and max exists, it usually happens when a binary value has an extreme length or a floating value has NaN. In this case, the stats provide little value and make it tricker to use.

It seems like we do have handling for these two cases. See Weston's message for NaN handling and `max_statistics_size` on [WriterProperties](https://arrow.apache.org/docs/cpp/api/formats.html#_CPPv4N7parquet16WriterPropertiesE). Based on that, I'd actually prefer we keep the ability to parse just the min or max if only one is available.

cpp/src/parquet/column_reader.cc
```diff
@@ -211,10 +211,15 @@ EncodedStatistics ExtractStatsFromHeader(const H& header) {
...
+  if (stats.__isset.max_value && stats.__isset.min_value) {
+    // TODO: check if the column_order is TYPE_DEFINED_ORDER.
+    page_statistics.set_max(stats.max_value);
+    page_statistics.set_min(stats.min_value);
```
wjones127, line 218:
```suggestion
  if (stats.__isset.max_value || stats.__isset.min_value) {
    // TODO: check if the column_order is TYPE_DEFINED_ORDER.
    if (stats.__isset.max_value) {
      page_statistics.set_max(stats.max_value);
    }
    if (stats.__isset.min_value) {
      page_statistics.set_min(stats.min_value);
    }
```

wjones127 (changes requested):

westonpace:
> It seems like we do have handling for these two cases. 

Just for clarity, the PR I linked (and my thoughts) were about how we currently handle row group statistics.  I'm not sure if the rules are identical for page statistics.  I mainly wanted to make sure my understanding of the row group statistics wasn't invalid.

wgtmac:
> > It seems like we do have handling for these two cases.
> 
> Just for clarity, the PR I linked (and my thoughts) were about how we currently handle row group statistics. I'm not sure if the rules are identical for page statistics. I mainly wanted to make sure my understanding of the row group statistics wasn't invalid.

IIUC, row group statistics are aggregated from page statistics so they should share the same rules. The parquet thrift message definition does allow only one side of min or max exist:
```thrift
/**
 * Statistics per row group and per page
 * All fields are optional.
 */
struct Statistics {
   /**
    * DEPRECATED: min and max value of the column. Use min_value and max_value.
    *
    * Values are encoded using PLAIN encoding, except that variable-length byte
    * arrays do not include a length prefix.
    *
    * These fields encode min and max values determined by signed comparison
    * only. New files should use the correct order for a column's logical type
    * and store the values in the min_value and max_value fields.
    *
    * To support older readers, these may be set when the column order is
    * signed.
    */
   1: optional binary max;
   2: optional binary min;
   /** count of null value in the column */
   3: optional i64 null_count;
   /** count of distinct values occurring */
   4: optional i64 distinct_count;
   /**
    * Min and max values for the column, determined by its ColumnOrder.
    *
    * Values are encoded using PLAIN encoding, except that variable-length byte
    * arrays do not include a length prefix.
    */
   5: optional binary max_value;
   6: optional binary min_value;
}
```

On the other side, the story of page index is different. The column index definition does require existence of both min and max values if it is not a null page:
```thrift
/**
 * Description for ColumnIndex.
 * Each <array-field>[i] refers to the page at OffsetIndex.page_locations[i]
 */
struct ColumnIndex {
  /**
   * A list of Boolean values to determine the validity of the corresponding
   * min and max values. If true, a page contains only null values, and writers
   * have to set the corresponding entries in min_values and max_values to
   * byte[0], so that all lists have the same length. If false, the
   * corresponding entries in min_values and max_values must be valid.
   */
  1: required list<bool> null_pages

  /**
   * Two lists containing lower and upper bounds for the values of each page
   * determined by the ColumnOrder of the column. These may be the actual
   * minimum and maximum values found on a page, but can also be (more compact)
   * values that do not exist on a page. For example, instead of storing ""Blart
   * Versenwald III", a writer may set min_values[i]="B", max_values[i]="C".
   * Such more compact values must still be valid values within the column's
   * logical type. Readers must make sure that list entries are populated before
   * using them by inspecting null_pages.
   */
  2: required list<binary> min_values
  3: required list<binary> max_values

  /**
   * Stores whether both min_values and max_values are ordered and if so, in
   * which direction. This allows readers to perform binary searches in both
   * lists. Readers cannot assume that max_values[i] <= min_values[i+1], even
   * if the lists are ordered.
   */
  4: required BoundaryOrder boundary_order

  /** A list containing the number of null values for each page **/
  5: optional list<i64> null_counts
}
```

So I am fine with parsing only one side min or max values from page/row group statistics. @westonpace @wjones127

cpp/src/parquet/column_reader.cc
wgtmac: Fixed. Please take a look again. Thanks @wjones127 !

cpp/src/parquet/column_reader.cc
mapleFU: So we revert back to previous mode that only has min or max is ok?

wgtmac:
The CI build is failed due to a recent branch rename and will be fixed by https://github.com/apache/arrow/pull/34233

wgtmac:
@westonpace Could you please take another pass?

wjones127 (approved):

PR #34116, opened 2023-02-10
GH-34117: [C++][Gandiva] Introduce ability to cast from int8 to float32 and int32

### LINKED ISSUE #34117: Extend cast operators for int8

In our analysis, we have the need to produce int8 data which can then optionally be processed as int32 or float32.

### Component(s)

C++ - Gandiva

ktf:
### Rationale for this change

In our analysis, we need to be able to extend data from int8 to float32 and int32.

Ability to cast int8 to float32 and int32 in Gandiva expressions.

### Are these changes tested?

Yes. We run our analysis using them and they produce the expected result.

Nothing apart the newly introduced capabilities.
* Closes: #34117


kou (commented):
Could you add tests for them?

ktf:
Where are the tests for the current casting operators?

kou:
https://github.com/apache/arrow/blob/master/cpp/src/gandiva/precompiled/arithmetic_ops_test.cc#L555-L593

PR #34125, opened 2023-02-10
GH-34076: [C#] Allow schema fields with duplicate names

### LINKED ISSUE #34076: [C#] Exception thrown when creating schema with multiple columns with the same name

When creating a Schema [Schema.cs](https://github.com/apache/arrow/blob/master/csharp/src/Apache.Arrow/Schema.cs), if the list of Fields contains fields that share the same name, an exception is thrown when constructing the fields dictionary.

The use of a dictionary disallows multiple fields that have the same name, even though it should be allowed.

### Component(s)

C#

DanTm99:
Skip fields with duplicate names when building the `_fieldsDictionary` in the Schema constructor.
This allows schemas to have multiple fields with the same name, as the field name is used as the key for this dictionary.

Deprecate the existing `IReadOnlyDictionary<string, Field> Fields` and expose a new `IReadOnlyList<Field> FieldsList` and a new `ILookup<string, Field> FieldsLookup`. Replace usages of `Fields` with `FieldsList` and `FieldsLookup`.
This allows for access to fields that were omitted from the `_fieldsDictionary`.
* Closes: #34076


csharp/src/Apache.Arrow/Schema.cs
```diff
@@ -22,32 +22,31 @@ namespace Apache.Arrow
 {
     public partial class Schema
     {
-        public IReadOnlyDictionary<string, Field> Fields
-        {
-            get => _fieldsDictionary;
-        }
+        [Obsolete("Use `FieldList` instead")]
+        public IReadOnlyDictionary<string, Field> Fields => _fieldsDictionary;
```
eerhardt, line 26: We could still support this API, couldn't we? We could use a custom `IReadOnlyDictionary` that stored multiple `Field` instances per string. And when someone tried accessing a field by name, if there was more than 1 field, we could define behavior of either:

* throw an exception
* return the last field with that name

DanTm99: When enumerating the key value pairs, either:
- multiple key value pairs would have the same key
- some fields would be missing

In the remarks for the `IReadOnlyDictionary` docs, it is stated that:
> Each pair must have a unique key.

This means the former would violate this and could lead to behaviour that the user would not expect.

If both 
- a single field is returned (instead of throwing an exception) when someone tries accessing a field by name for which there are multiple fields
- duplicate fields are skipped when enumerating

then this is equivalent to just skipping duplicate keys when building the dictionary.

This API can still be supported by skipping duplicate keys when building the dictionary, but that would lead to the dictionary being an inaccurate representation of the state of the schema when there are multiple fields with the same name. Because of this, deprecating it could make it clear that using it is to be avoided.

westonpace: Do we want to add an `ILookup<string, Field>` which might be handy on wide schemas?  It's still an API break but might be useful.

DanTm99: @westonpace Exposing an `ILookup<string, Field>` alongside an `IReadOnlyList<Field>` would provide more flexibility to users. I think doing so could be a good idea

DanTm99: I've added the `ILookup<string, Field>`

csharp/src/Apache.Arrow/Schema.cs
```diff
@@ -22,98 +22,91 @@ namespace Apache.Arrow
 {
     public partial class Schema
     {
-        public IReadOnlyDictionary<string, Field> Fields
-        {
-            get => _fieldsDictionary;
-        }
-
+        [Obsolete("Use `FieldsList` instead")]
```
westonpace, line 25:
```suggestion
        [Obsolete("Use `FieldsLookup` instead")]
```

westonpace (approved):
This seems pretty good to me.  Thanks for taking a crack at this.  I notice some of these changes seem to be linter inspired (removing whitespace, preferring `Assert.Equal`, etc.)  Do we want to setup CI to hook some kind of linter in?  Not for this PR but maybe something to think about for later.

csharp/src/Apache.Arrow/Schema.cs
DanTm99: Since `FieldsList` wasn't previously exposed, `Fields` was the only way of getting a list of all the fields (via `Fields.Values`). 

So for some use cases `FieldsList` would be preferred over `Fields`, but for others `FieldsLookup` would be preferred over `Fields`. I've updated the message to mention both, so the user can decide which would be best to use for their use case.

csharp/src/Apache.Arrow/Ipc/ArrowStreamWriter.cs
```diff
@@ -522,13 +522,12 @@ private protected async Task WriteDictionaryAsync(Field field, CancellationToken
             long id = DictionaryMemo.GetId(field);
             IArrowArray dictionary = DictionaryMemo.GetDictionary(id);
 
-            var fieldsDictionary = new Dictionary<string, Field> {
-                { dictionaryField.Name, dictionaryField } };
+            var fieldsList = new List<Field> { dictionaryField };
```
eerhardt, line 525:
```suggestion
            var fields = new Field[] { dictionaryField };
```

Array implements `IReadOnlyList`, so we can just create an array here instead of a List and save on an allocation.

csharp/src/Apache.Arrow/Schema.cs
```diff
@@ -44,80 +43,74 @@ public partial class Schema
...
 
-        public Field GetFieldByIndex(int i)
-        {
-            return _fields[i];
-        }
+        public Field GetFieldByIndex(int i) => _fieldsList[i];
 
-        public Field GetFieldByName(string name) =>
-            Fields.TryGetValue(name, out Field field) ? field : null;
+        public Field GetFieldByName(string name) => FieldsLookup[name].First();
```
eerhardt, line 72:
```suggestion
        public Field GetFieldByName(string name) => FieldsLookup[name].FirstOrDefault();
```

eerhardt: Can you add a test here to ensure `null` is returned when a field doesn't exist?

eerhardt (commented):
Thanks for the contribution, @DanTm99!

I just have a couple comments. After these are addressed, I believe this can be merged.

csharp/src/Apache.Arrow/Ipc/ArrowStreamWriter.cs
DanTm99: Done

csharp/src/Apache.Arrow/Schema.cs
DanTm99: DOne

PR #34127, opened 2023-02-10
GH-34101: [Go][Parquet] NewSchemaManifest creates wrong schema field

### LINKED ISSUE #34101: [Go] pqarrow.NewSchemaManifest creates wrong schema field for array object fields

I'm using the arrow V11 package for reading parquet files.

Calling `pqarrow.NewFileReader` creates the schema manifest using `pqarrow.NewSchemaManifest`.
When dealing with objects array fields, the record readers create an array of primitives of the first field in the object. for example:
[[“a”:1, “b”:2},[“a”:3, “b”:4}] -> [1,3]

diving into the code I found this in `pqarrow/schema.go -> listToSchemaField`:
```Go
listGroup := listNode.(*schema.GroupNode)
if listGroup.NumFields() == 1 && (listGroup.Name() == "array" || strings.HasSuffix(listGroup.Name(), "_tuple")) {
	// list of primitive type
	if err := groupToStructField(listGroup, currentLevels, ctx, out, &out.Children[0]); err != nil {
		return err
	}
} else {
	if err := nodeToSchemaField(listGroup.Field(0), currentLevels, ctx, out, &out.Children[0]); err != nil {
		return err
	}
}
```

At this stage, the listGroup is the inner dictionary. Assuming the above example listGroup.NumFields() returns 2 and the schema node is converted to a primitive field taking the first field in the group (else clause).
It also seems like the `groupToStructField` function is more suitable for this case.

changing the condition solves the problem and the field is being read as expected.
```Go
listGroup := listNode.(*schema.GroupNode)
if listGroup.NumFields() != 1 && (listGroup.Name() == "array" || strings.HasSuffix(listGroup.Name(), "_tuple")) {
	if err := groupToStructField(listGroup, currentLevels, ctx, out, &out.Children[0]); err != nil {
		return err
	}
} else {
	// list of primitive type
	if err := nodeToSchemaField(listGroup.Field(0), currentLevels, ctx, out, &out.Children[0]); err != nil {
		return err
	}
}
```

Can you take a look and fix it? the current state is misleading as the user might think this is the real value

### Component(s)

Go, Parquet

zeroshade:
### Rationale for this change
Bug was found in the handling of the special cases when converting Parquet Schemas to Arrow Schemas.

Parquet Schemas containing a repeated group (list) field with multiple child fields, even when named "array" or "<list_name>_tuple" will properly convert to a list of struct now.

### Are these changes tested?
Yes, unit test was added for this case.

Fixes a bug that caused users to get a list of a struct with a single field rather than a list of struct with multiple fields.

* Closes: #34101

### DIFF
--- a/go/parquet/pqarrow/schema_test.go
+++ b/go/parquet/pqarrow/schema_test.go
@@ -18,7 +18,6 @@ package pqarrow_test
 
 import (
 	"encoding/base64"
-	"fmt"
 	"testing"
 
 	"github.com/apache/arrow/go/v12/arrow"
@@ -366,7 +365,5 @@ func TestListStructBackwardCompatible(t *testing.T) {
 
 	arrsc, err := pqarrow.FromParquet(pqSchema, nil, metadata.KeyValueMetadata{})
 	assert.NoError(t, err)
-	fmt.Println(arrowSchema)
-	fmt.Println(arrsc)
 	assert.True(t, arrowSchema.Equal(arrsc))
 }
--- a/go/parquet/pqarrow/schema_test.go
+++ b/go/parquet/pqarrow/schema_test.go
@@ -342,9 +342,12 @@ func TestListStructBackwardCompatible(t *testing.T) {
 	pqSchema := schema.NewSchema(schema.MustGroup(schema.NewGroupNode("root", parquet.Repetitions.Required, schema.FieldList{
 		schema.Must(schema.NewGroupNodeLogical("answers", parquet.Repetitions.Optional, schema.FieldList{
 			schema.Must(schema.NewGroupNode("array", parquet.Repetitions.Repeated, schema.FieldList{
-				schema.MustPrimitive(schema.NewPrimitiveNodeLogical("type", parquet.Repetitions.Optional, schema.StringLogicalType{}, parquet.Types.ByteArray, -1, -1)),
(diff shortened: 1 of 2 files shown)


go/parquet/pqarrow/schema_test.go
```diff
@@ -309,3 +310,63 @@ func TestConvertArrowStruct(t *testing.T) {
...
+					arrow.Field{Name: "type", Type: arrow.BinaryTypes.String, Nullable: true, Metadata: meta},
+					arrow.Field{Name: "rdata", Type: arrow.BinaryTypes.String, Nullable: true, Metadata: meta},
+					arrow.Field{Name: "class", Type: arrow.BinaryTypes.String, Nullable: true, Metadata: meta},
+				), Nullable: true}), Nullable: true, Metadata: meta},
+		}, nil)
+
+	arrsc, err := pqarrow.FromParquet(pqSchema, nil, metadata.KeyValueMetadata{})
+	assert.NoError(t, err)
+	fmt.Println(arrowSchema)
+	fmt.Println(arrsc)
```
lidavidm, line 370: Did you mean to leave these in?

lidavidm (approved):

go/parquet/pqarrow/schema_test.go
zeroshade: GAH i'll fix those lol. it took a little bit of printf debugging to figure out why the schemas didn't match haha.

PR #3257, opened 2018-12-23
ARROW-3928: [Python] Deduplicate Python objects when converting binary, string, date, time types to object arrays

wesm:
This adds a `deduplicate_objects` option to all of the `to_pandas` methods. It works with string types, date types (when `date_as_object=True`), and time types.

I also made it so that `ScalarMemoTable` can be used with `string_view`, for more efficient memoization in this case.

I made the default for `deduplicate_objects` is True. When the ratio of unique strings to the length of the array is low, not only does this use drastically less memory, it is also faster. I will write some benchmarks to show where the "crossover point" is when the overhead of hashing makes things slower.

Let's consider a simple case where we have 10,000,000 strings of length 10, but only 1000 unique values:

```
In [50]: import pandas.util.testing as tm                                                                 

In [51]: unique_values = [tm.rands(10) for i in range(1000)]                                                                 

In [52]: values = unique_values * 10000                                                                                      

In [53]: arr = pa.array(values)                                                                                              

In [54]: timeit arr.to_pandas()                                                                                              
236 ms ± 1.69 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [55]: timeit arr.to_pandas(deduplicate_objects=False)                                                                     
730 ms ± 12.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Almost 3 times faster in this case. The different in memory use is even more drastic

```
In [44]: unique_values = [tm.rands(10) for i in range(1000)]                                                                 

In [45]: values = unique_values * 10000                                                                                      

In [46]: arr = pa.array(values)                                                                                              

In [49]: %memit result11 = arr.to_pandas()                                                                                   
peak memory: 1505.89 MiB, increment: 76.27 MiB

In [50]: %memit result12 = arr.to_pandas(deduplicate_objects=False)                                                          
peak memory: 2202.29 MiB, increment: 696.11 MiB
```

As you can see, this is a huge problem. If our bug reports about Parquet memory use problems are any indication, users have been suffering from this issue for a long time. 

When the strings are mostly unique, then things are slower as expected, the peak memory use is higher because of the hash table

```
In [17]: unique_values = [tm.rands(10) for i in range(500000)]                                                               

In [18]: values = unique_values * 2                                                                                          

In [19]: arr = pa.array(values)                                                                                              

In [20]: timeit result = arr.to_pandas()                                                                                     
177 ms ± 574 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

[...]

PR #34128, opened 2023-02-10
GH-34104: [Python] update deduplicate_objects default in docs to match implementation

### LINKED ISSUE #34104: [Python][Docs] deduplicate doesn't match docs

in `pyarrow/array.pxi` deduplicate_objects has a default value set to true but the docs say its false. 

https://github.com/apache/arrow/blob/fc1f9ebbc4c3ae77d5cfc2f9322f4373d3d19b8a/python/pyarrow/array.pxi#L724
https://github.com/apache/arrow/blob/fc1f9ebbc4c3ae77d5cfc2f9322f4373d3d19b8a/python/pyarrow/array.pxi#L690

### Component(s)

Python

westonpace:
* Closes: #34104

### DIFF
--- a/python/pyarrow/array.pxi
+++ b/python/pyarrow/array.pxi
@@ -721,7 +721,7 @@ cdef class _PandasConvertible(_Weakrefable):
             If False, all timestamps are converted to datetime64[ns] dtype.
         use_threads : bool, default True
             Whether to parallelize the conversion using multiple threads.
-        deduplicate_objects : bool, default False
+        deduplicate_objects : bool, default True
             Do not create multiple copies Python objects when created, to save
             on memory use. Conversion will be slower.
         ignore_metadata : bool, default False


westonpace:
I confirmed that the default behavior does indeed deduplicate:

```
>>> x = pa.array(['one', 'one'])
>>> y = x.to_pandas()
>>> id(y[0]) == id(y[1])
True
>>> z = x.to_pandas(deduplicate_objects=False)
>>> id(z[0]) == id(z[1])
False
```

AlenkaF (approved):

jorisvandenbossche:
I was a bit surprised by this default, but indeed it has been True as default from the start: https://github.com/apache/arrow/pull/3257 (the PR mentions some rationale for the default; the default is only False on the C++ side)

But be careful with a test like that, because Python itself also does some small string interning. If you create such a list in Python:

```python
>>> l = ["one", "one"]
>>>  id(l[0]) == id(l[1])
True
```

But it seems that with the Python C API we use when creating those strings in the to_pandas conversion, this string interning is not used.

jorisvandenbossche (approved):

PR #34133, opened 2023-02-10
GH-34737: [C#] C Data interface for schemas and types

### LINKED ISSUE #33856: [C#] Implement C Data Interface for C#

C# is one of the few independent implementations of Arrow that doesn't support the C Data Interface.

Supporting this would allow the [ParquetSharp](https://github.com/G-Research/ParquetSharp) library, which wraps the Parquet C++ implementation, to produce and consume Arrow arrays and record batches.

### Subtasks

* [x] #34737
* [ ] Support moving field and schema metadata over C data interface
* [ ] Define C Data Interface for arrays

### Component(s)

C#

### LINKED ISSUE #34737: [C#] Define C Data Interface for schemas/fields/types

wjones127:
### Rationale for this change

This starts the C Data Interface implementation for C# with integration for `ArrowSchema`. `ArrowArray` will come in a follow-up PR.

* Adds classes `CArrowSchema` and `ImportedArrowSchema` which allow interacting with the `CArrowSchema`.
* Adds integration tests with PyArrow, inspired by the similar integration tests in [arrow-rs](https://github.com/apache/arrow-rs/blob/master/arrow/src/pyarrow.rs)

### Are these changes tested?

Yes, the PyArrow integration tests validate the functionality.

This only adds new APIs, and doesn't change any existing ones.

* Closes: #33856
* Closes: #34737


csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,519 @@
...
+        /// <param name="schema">An uninitialized CArrowSchema.</param>
+        public static void ExportDataType(IArrowType datatype, out CArrowSchema schema)
+        {
+            schema.format = GetFormat(datatype);
+            schema.name = null;
+            schema.metadata = null;
+            schema.flags = GetFlags(datatype);
+
+            schema.children = ConstructChildren(datatype);
+            schema.n_children = datatype is NestedType ? ((NestedType)datatype).Fields.Count : 0;
```
kou, line 139: How about setting them at once in `ConstructChildren()` because they are related?

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,519 @@
...
+                    CArrowSchema.ExportField(fields[i], out c_schema);
+                    IntPtr exported_schema = c_schema.AllocateAsPtr();
+                    pointer_list[i] = exported_schema;
+                }
+
+                return pointer_list;
+            }
+            else
+            {
+                return (IntPtr*)IntPtr.Zero;
```
kou, line 107: I'm not familiar with C# but is it safe?
Is there `nullptr` in C#?

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,519 @@
...
+                schema.release(ptr);
+            }
+            Marshal.FreeHGlobal(ptr);
+        }
+
+        /// <summary>
+        /// Export to an existing pointer
+        /// </summary>
+        /// <param name="ptr"></param>
+        /// <returns></returns>
```
kou, line 226: Should we fill them?

csharp/test/Apache.Arrow.Tests/CDataInterfaceSchemaTests.cs
```diff
@@ -0,0 +1,302 @@
...
+                IntPtr imported_ptr = c_schema.AllocateAsPtr();
+
+                using (Py.GIL())
+                {
+                    dynamic py_datatype = py_field.type;
+                    py_datatype._export_to_c(imported_ptr.ToInt64());
+                }
+
+                var dataTypeComparer = new ArrayTypeComparer(field.DataType);
+                var imported_type = new ImportedArrowSchema(imported_ptr);
```
kou, line 147: How about using the same naming convention? (`camelCase` v.s. `snake_case`)

wjones127: That's a good point. I need to fix the code style.

csharp/src/Apache.Arrow/C/CArrowSchema.cs
wjones127: Yes, filled this one.

csharp/src/Apache.Arrow/C/CArrowSchema.cs
wjones127: I was originally trying to see if I could do something like

```csharp
(schema.children, schema.n_children) = ConstructChildren(datatype);
```

but couldn't figure out how to import support for tuples. I also tried a signature like:

```csharp
void SetChildren(datatype, out CArrowSchema schema);
```

But then C# wants me to set every field of `schema` within `SetChildren` (it doesn't know it's only being called from a function that does).

Open to suggestions :)

ci/scripts/csharp_test.sh
```diff
@@ -21,6 +21,17 @@ set -ex
 
 source_dir=${1}/csharp
 
+# Python and PyArrow are required for C Data Interface tests.
+if [ -z "${PYTHON}" ]; then
+  if [ ! which python3 > /dev/null 2>&1 ]; then
+    export PYTHON=python
+  else
+    export PYTHON=python3
+  fi
```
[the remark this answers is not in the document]
wjones127: Thanks!

wjones127:
@adamreeve would you be interested in reviewing?

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,339 @@
...
+    /// </remarks>
+    [StructLayout(LayoutKind.Sequential)]
+    unsafe public struct CArrowSchema
+    {
+        [MarshalAs(UnmanagedType.LPStr)]
+        public string format;
+        [MarshalAs(UnmanagedType.LPStr)]
+        public string name;
+        [MarshalAs(UnmanagedType.LPStr)]
+        public string metadata;
```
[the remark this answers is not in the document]
wjones127: Good point. This isn't nul-terminated so I think there's probably no default marshaller for this.

csharp/test/Apache.Arrow.Tests/CDataInterfaceSchemaTests.cs
```diff
@@ -0,0 +1,300 @@
...
+using System.Linq;
+using System.Runtime.InteropServices;
+using Apache.Arrow.C;
+using Apache.Arrow.Types;
+using Python.Runtime;
+using Xunit;
+
+namespace Apache.Arrow.Tests
+{
+    public class CDataSchemaTest
```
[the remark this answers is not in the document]
wjones127: That's a good point. I will add that, but also want to make sure that we don't accidentally skip this in CI without realizing it.

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,339 @@
...
+    /// 
+    /// <remarks>
+    /// This is used to export <see cref="ArrowType"/>, <see cref="Field"/>, or
+    /// <see cref="Schema"/> to other languages. It matches the layout of the
+    /// ArrowSchema struct described in https://github.com/apache/arrow/blob/main/cpp/src/arrow/c/abi.h.
+    /// </remarks>
+    [StructLayout(LayoutKind.Sequential)]
+    unsafe public struct CArrowSchema
+    {
+        [MarshalAs(UnmanagedType.LPStr)]
```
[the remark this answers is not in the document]
wjones127: This doesn't seem to be recognized by `netstandard1.3` and `netstandard2.0`. 😞

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,339 @@
...
+        public string format;
+        [MarshalAs(UnmanagedType.LPStr)]
+        public string name;
+        [MarshalAs(UnmanagedType.LPStr)]
+        public string metadata;
+        public long flags;
+        public long n_children;
+        public IntPtr* children;
```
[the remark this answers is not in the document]
wjones127: Yeah looking at this, I don't know how I would do this without unsafe:

```csharp
Marshal.AllocHGlobal(numFields * sizeof(IntPtr))
```

And that's called in the constructor. (Compiler says that you can't call `sizeof(IntPtr)` outside of an unsafe context; not sure why.)

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,339 @@
...
+
+            // TODO: when we implement MapType, make sure to set the KEYS_SORTED flag.
+            return flags;
+        }
+
+        private static IntPtr* ConstructChildren(IArrowType datatype)
+        {
+            if (datatype is NestedType)
+            {
+                IReadOnlyList<Field> fields = ((NestedType)datatype).Fields;
```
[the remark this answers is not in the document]
wjones127: Thanks, that is nifty.

csharp/src/Apache.Arrow/C/CArrowSchema.cs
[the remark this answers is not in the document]
adamreeve: Ah yeah you may need to handle these manually then, eg. see https://github.com/G-Research/ParquetSharp/blob/467d99298fb5a5b9d5935b9c8dbde95e63954dd3/csharp/StringUtil.cs#L22

wjones127:
Thanks for the feedback @adamreeve. I've realized I could simplify the API by changing the `Export*` methods into constructors for `CArrowSchema`. And I've moved the `ImportedArrowSchema` into an internal class within `CArrowSchema` so that it can handle the common import and disposal logic. Let me know what you think.

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,559 @@
...
+// under the License.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices;
+using Apache.Arrow.Types;
+
+[UnmanagedFunctionPointer(CallingConvention.StdCall)]
```
westonpace, line 25: Why `StdCall`?

csharp/src/Apache.Arrow/C/CArrowSchema.cs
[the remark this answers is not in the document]
westonpace: What about [`IntPtr.Size`](https://learn.microsoft.com/en-us/dotnet/api/system.intptr.size?view=net-7.0#system-intptr-size)?

```
Marshal.AllocHGlobal(numFields * IntPtr.Size);
```

westonpace (commented):
Sorry, just barely started looking at this but will have to come back to it tomorrow.

csharp/src/Apache.Arrow/C/CArrowSchema.cs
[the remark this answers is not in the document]
adamreeve: Yeah I think `IntPtr.Size` works here, for other struct types you'd generally want to use `Marshal.SizeOf` in this scenario

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,559 @@
...
+        /// <summary>
+        /// Export to an existing pointer
+        /// </summary>
+        /// <param name="ptr">An allocated but uninitialized pointer.</param>
+        public void Export(IntPtr ptr)
+        {
+            Marshal.StructureToPtr<CArrowSchema>(this, ptr, false);
+        }
+
+        public static IntPtr AllocateUninitialized()
```
adamreeve, line 224: This method needs documentation, and should mention needing to call `FreePtr`. The examples in the `Import` methods below should probably also include calls to `FreePtr` for completeness.

adamreeve:
> Thanks for the feedback @adamreeve. I've realized I could simplify the API by changing the `Export*` methods into constructors for `CArrowSchema`. And I've moved the `ImportedArrowSchema` into an internal class within `CArrowSchema` so that it can handle the common import and disposal logic. Let me know what you think.

:+1: nice, the new API looks a lot more straightforward to work with

csharp/src/Apache.Arrow/C/CArrowSchema.cs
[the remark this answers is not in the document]
wjones127: Okay I've narrowed it so `unsafe` only appears when constructing and accessing children. Does that seem like the improvement we are looking for?

csharp/src/Apache.Arrow/C/CArrowSchema.cs
wjones127: To be honest, I'm not sure on this. It was the default and seemed to work with the integration tests fine.

csharp/src/Apache.Arrow/C/CArrowSchema.cs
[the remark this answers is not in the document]
adamreeve: 👍 yeah I think that is preferable

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,584 @@
...
+        }
+
+        /// <summary>
+        /// Allocate an unmanaged pointer and copy this instances data to it.
+        /// </summary>
+        /// <remarks>
+        /// To avoid a memory leak, you must call <see cref="FreePtr"/> on this
+        /// pointer when done using it.
+        /// </remarks>
+        public IntPtr AllocateAsPtr()
```
adamreeve, line 211: Maybe a nitpick but I think `ExportAsPtr` or maybe just having an override of `Export` would more clearly describe the main intent of this method. I guess `AllocateAndExport` would be more descriptive but is getting a little verbose

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,584 @@
...
+            }
+
+            public void Dispose()
+            {
+                // We only call release on a root-level schema, not child ones.
+                if (_isRoot)
+                {
+                    _data.release(_handle);
+                }
+            }
```
westonpace, line 455: `Dispose` should be idempotent.  Also, there is [a pattern](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose) that is typically used with `Dispose()` and `Dispose(bool)` though that is probably not strictly needed here since there is no managed state.

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,584 @@
...
+                        "I" => new UInt32Type(),
+                        "l" => new Int64Type(),
+                        "L" => new UInt64Type(),
+                        _ => throw new InvalidDataException($"Indices must be an integer, but got format string {format}"),
+                    };
+
+                    var dictionarySchema = new ImportedArrowSchema(_data.dictionary, /*is_root*/ false);
+                    ArrowType dictionaryType = dictionarySchema.GetAsType();
+
+                    bool ordered = (_data.flags & CArrowSchema.ArrowFlagNullable) == CArrowSchema.ArrowFlagNullable;
```
westonpace, line 478: ArrowFlagDictionaryOrdered?

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,584 @@
...
+                    if (_data.n_children != 1)
+                    {
+                        throw new Exception("Expected list type to have exactly one child.");
+                    }
+                    ImportedArrowSchema childSchema;
+                    if (_data.GetChild(0) == IntPtr.Zero)
+                    {
+                        throw new Exception("Expected list type child to be non-null.");
+                    }
+                    childSchema = new ImportedArrowSchema(_data.GetChild(0));
```
westonpace, line 495: `is_root=false`?

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,584 @@
...
+                        "s" => new Int16Type(),
+                        "S" => new UInt16Type(),
+                        "i" => new Int32Type(),
+                        "I" => new UInt32Type(),
+                        "l" => new Int64Type(),
+                        "L" => new UInt64Type(),
+                        _ => throw new InvalidDataException($"Indices must be an integer, but got format string {format}"),
+                    };
+
+                    var dictionarySchema = new ImportedArrowSchema(_data.dictionary, /*is_root*/ false);
```
westonpace, line 475:
```suggestion
                    var dictionarySchema = new ImportedArrowSchema(_data.dictionary, isRoot: false);
```

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,584 @@
...
+                {
+                    var child_schemas = new ImportedArrowSchema[_data.n_children];
+
+                    for (int i = 0; i < _data.n_children; i++)
+                    {
+                        if (_data.GetChild(i) == IntPtr.Zero)
+                        {
+                            throw new Exception("Expected struct type child to be non-null.");
+                        }
+                        child_schemas[i] = new ImportedArrowSchema(_data.GetChild(i));
```
westonpace, line 511: `is_root=false`?

csharp/src/Apache.Arrow/C/StringUtil.cs
```diff
@@ -0,0 +1,67 @@
...
+                for (length = 0; s[length] != '\0'; ++length)
+                {
+                }
+
+                return System.Text.Encoding.UTF8.GetString(s, length);
+            }
+#endif
+        }
+    }
+}
```
westonpace, line 67: newline

westonpace (changes requested):
Took a bit more thorough scan through.  Thanks for working on this.

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,584 @@
...
+    /// <remarks>
+    /// This is used to export <see cref="ArrowType"/>, <see cref="Field"/>, or
+    /// <see cref="Schema"/> to other languages. It matches the layout of the
+    /// ArrowSchema struct described in https://github.com/apache/arrow/blob/main/cpp/src/arrow/c/abi.h.
+    /// </remarks>
+    [StructLayout(LayoutKind.Sequential)]
+    public struct CArrowSchema
+    {
+        public IntPtr format;
+        public IntPtr name;
```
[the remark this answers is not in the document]
wjones127: Also see: https://github.com/apache/arrow/pull/34133#discussion_r1119262146

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,584 @@
...
+        public IntPtr dictionary;
+        [MarshalAs(UnmanagedType.FunctionPtr)]
+        public ReleaseCArrowSchema release;
+        public IntPtr private_data;
+
+        private static string GetFormat(IArrowType datatype)
+        {
+            TypeFormatter formatter = new TypeFormatter();
+            datatype.Accept(formatter);
+            return formatter.FormatString;
```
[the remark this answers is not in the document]
wjones127: Thanks. That is a lot cleaner :)

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,584 @@
...
+                FormatString = $"w:{datatype.ByteWidth}";
+            }
+            // Decimal
+            public void Visit(Decimal128Type datatype)
+            {
+                FormatString = $"d:{datatype.Precision},{datatype.Scale}";
+            }
+            public void Visit(Decimal256Type datatype)
+            {
+                FormatString = $"w:{datatype.Precision},{datatype.Scale},256";
```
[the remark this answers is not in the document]
wjones127: Thanks for catching this. Reminds me I meant to test decimals and timestamps.

csharp/src/Apache.Arrow/C/CArrowSchema.cs
wjones127: Thanks. I've updated the tests to catch this.

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,599 @@
...
+    /// 
+    /// <remarks>
+    /// This is used to export <see cref="ArrowType"/>, <see cref="Field"/>, or
+    /// <see cref="Schema"/> to other languages. It matches the layout of the
+    /// ArrowSchema struct described in https://github.com/apache/arrow/blob/main/cpp/src/arrow/c/abi.h.
+    /// </remarks>
+    [StructLayout(LayoutKind.Sequential)]
+    public struct CArrowSchema
+    {
+        public IntPtr format;
```
eerhardt, line 42: We shouldn't be using `IntPtr` for data that is actually pointers. Instead, use the correct pointer type.

```suggestion
        public byte* format;
```

See https://learn.microsoft.com/en-us/dotnet/api/system.intptr?view=net-7.0#remarks. An `IntPtr` is

> Represents a signed integer where the bit-width is the same as a pointer.

We should use pointers when we really mean pointers, and `IntPtr` (or `nint` in newer C# versions) to mean "a number that is the same bit-width as a pointer", for example `size_t`.

wjones127: That makes sense for the string fields. What type would you suggest for `children` and `dictionary`?

https://github.com/apache/arrow/blob/54f8fe377099ab9904e9858aa26c63caa199e3df/cpp/src/arrow/c/abi.h#L40-L41

Originally I had

```
public IntPtr* children;
public IntPtr dictionary;
```

eerhardt: > What type would you suggest for children and dictionary?

```C#
public CArrowSchema** children;
public CArrowSchema* dictionary;
```

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,599 @@
...
+        public IntPtr name;
+        public IntPtr metadata;
+        public long flags;
+        public long n_children;
+        public IntPtr children;
+        public IntPtr dictionary;
+        [MarshalAs(UnmanagedType.FunctionPtr)]
+        public ReleaseCArrowSchema release;
```
eerhardt, line 50: We should use function pointers here, not managed delegates.

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/unsafe-code#function-pointers

(Coding in GitHub, I'm not 100% sure this code will compile/run, but hopefully you get the idea):

```suggestion
        [MarshalAs(UnmanagedType.FunctionPtr)]
        public delegate* unmanaged<CArrowSchema*, void> release;
```

csharp/src/Apache.Arrow/C/CArrowSchema.cs
wjones127: Such a type issues a warning:
```
csharp(CS8500): This takes the address of, gets the size of, or declares a pointer to a managed type ('CArrowSchema')
```

Also is this how to allocate such a type?

```csharp
var pointerList = (CArrowSchema**)Marshal.AllocHGlobal(numFields * IntPtr.Size);
```

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,599 @@
...
+            {
+                return IntPtr.Zero;
+            }
+        }
+
+        /// <summary>
+        /// Initialize the exported C schema as an Arrow type.
+        /// </summary>
+        /// <param name="datatype">The Arrow type to export.</param>
+        public CArrowSchema(IArrowType datatype)
```
eerhardt, line 222: I'm a bit confused by the intention/use case here. I assumed the intention was so C# code could consume "C Data Interface" memory already in memory created by some other system. 

Is the use case here the other way? Maybe instead of putting these as methods on the `CArrowSchema` struct, it would be better to have a separate class/API/factory that created these. And then `CArrowSchema` struct would stay about defining the data structure, and not contain the logic of turning C# objects into the data structure.

csharp/src/Apache.Arrow/C/CArrowSchema.cs
eerhardt: From https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0208 (a similar warning):

> A managed type is:
> any reference type
> any struct that contains a reference type as a field or property

We need to make sure `CArrowSchema` doesn't contain any reference types as a field or property. i.e. the managed delegate that I pointed out in the other comment.

csharp/src/Apache.Arrow/C/CArrowSchema.cs
wjones127: That makes sense. The syntax here is for C# 9.0 and above, so I'll need to figure out the equivalent for C# 8.0.

eerhardt: https://github.com/apache/arrow/blob/48df467eb5c684aafbd168ad056c64bb2f6706e8/csharp/Directory.Build.props#L37

just update this to be `latest` instead. Not sure why we are setting `8.0` explicitly.

eerhardt: Also, if it makes it easier, we can make the C Data Interface only available on NETCOREAPP. I don't think we need to support it on `netstandard` at all. At least not until someone actually needs it.

eerhardt: Can you elaborate on your reasons for targeting `netstandard2.1` instead of `net6.0`? I see that ParquetSharp also targets .NET Framework (`net471`). Would you need the C Data interface on .NET Framework?

adamreeve: It's an open source project so we try to support a broad range of platforms, I don't know of specific users that would need .netstandard 2.1 specifically that couldn't use net6.0 though. And no we wouldn't need this supported in net471. So after considering this some more it's probably fine if the C data interface support is netcoreapp only as we could provide arrow integration as a separate library or add a new target and limit support to that

wjones127:
I have refactored to make `CArrowSchema` simply use real pointers for its members, and separated out the import and export functions to their own classes. So now instead of passing around `IntPtr`, users will be passing around `CArrowSchema*`. I like that this makes the API more strongly typed, as IntPtr drops the information about what type it is pointing to.

csharp/src/Apache.Arrow/C/CArrowSchemaExporter.cs
```diff
@@ -0,0 +1,231 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices;
+using Apache.Arrow.Types;
+
+namespace Apache.Arrow.C
+{
+    public sealed class CArrowSchemaExporter
```
adamreeve, line 26: This class should be static if it's only going to have static methods

csharp/src/Apache.Arrow/C/CArrowSchemaImporter.cs
```diff
@@ -0,0 +1,285 @@
...
+        /// <examples>
+        /// Typically, you will allocate a uninitialized CArrowSchema pointer,
+        /// pass that to external function, and then use this method to import
+        /// the result.
+        /// 
+        /// <code>
+        /// IntPtr importedPtr = CArrowSchema.AllocateUninitialized();
+        /// foreign_export_function(importedPtr);
+        /// ArrowType importedType = CArrowSchema.ImportType(importedPtr);
+        /// CArrowSchema.FreePtr(importedPtr);
```
adamreeve, line 39: Should be `Free` rather than `FreePtr`, and similarly for docs in other methods

csharp/src/Apache.Arrow/C/CArrowSchemaImporter.cs
```diff
@@ -0,0 +1,285 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Apache.Arrow.Types;
+
+namespace Apache.Arrow.C
+{
+    public sealed class CArrowSchemaImporter
```
adamreeve, line 25: This class should probably be static too

adamreeve (commented):
Looks good to me, I've just commented on some minor things to tidy up

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,122 @@
...
+                throw new ArgumentOutOfRangeException("Child index out of bounds.");
+            }
+            if (children == null)
+            {
+                throw new Exception("Children array is null but n_children is not zero.");
+            }
+            unsafe
+            {
+                return children[i];
+            }
```
eerhardt, line 119:
```suggestion
            return children[i];
```

Whole struct is unsafe.

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// <remarks>
+        /// Known valid flags are <see cref="ArrowFlagDictionaryOrdered" />,
+        /// <see cref="ArrowFlagNullable" />, and <see cref="ArrowFlagMapKeysSorted" />.
+        /// </remarks>
+        public bool GetFlag(long flag)
+        {
+            return (flags & flag) == flag;
+        }
+
+        internal CArrowSchema* GetChild(int i)
```
eerhardt, line 106:
```suggestion
        internal readonly CArrowSchema* GetChild(int i)
```

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,122 @@
...
+        public const long ArrowFlagMapKeysSorted = 4;
+
+        /// <summary>
+        /// Get the value of a particular flag.
+        /// </summary>
+        /// <remarks>
+        /// Known valid flags are <see cref="ArrowFlagDictionaryOrdered" />,
+        /// <see cref="ArrowFlagNullable" />, and <see cref="ArrowFlagMapKeysSorted" />.
+        /// </remarks>
+        public bool GetFlag(long flag)
```
eerhardt, line 101:
```suggestion
        public readonly bool GetFlag(long flag)
```

csharp/src/Apache.Arrow/C/CArrowSchema.cs
eerhardt, line 106: Any reason this is `internal`?

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+            return (flags & flag) == flag;
+        }
+
+        internal CArrowSchema* GetChild(int i)
+        {
+            if (i >= n_children)
+            {
+                throw new ArgumentOutOfRangeException("Child index out of bounds.");
```
eerhardt, line 110:
```suggestion
                throw new ArgumentOutOfRangeException($"Child index '{i}' out of bounds.");
```

csharp/src/Apache.Arrow/C/StringUtil.cs
```diff
@@ -0,0 +1,66 @@
...
+            var utf8 = System.Text.Encoding.UTF8;
+            int byteCount = utf8.GetByteCount(str);
+            byte* byteArray = (byte*)Marshal.AllocHGlobal(byteCount + 1);
+
+            fixed (char* chars = str)
+            {
+                utf8.GetBytes(chars, str.Length, byteArray, byteCount);
+            }
+
+            // Need to make sure it is nul-terminated.
```
eerhardt, line 38:
```suggestion
            // Need to make sure it is null-terminated.
```

csharp/src/Apache.Arrow/C/StringUtil.cs
```diff
@@ -0,0 +1,66 @@
...
+#if NETSTANDARD2_1_OR_GREATER
+            return Marshal.PtrToStringUTF8(ptr);
+#else
+            if (ptr == null)
+            {
+                return null;
+            }
+
+            unsafe
+            {
```
eerhardt, line 55:
```suggestion
```

The method is already unsafe.

csharp/src/Apache.Arrow/C/CArrowSchemaExporter.cs
```diff
@@ -0,0 +1,268 @@
...
+            else
+            {
+                return null;
+            }
+        }
+
+        private static unsafe void ReleaseCArrowSchema(CArrowSchema* schema)
+        {
+            if (schema == null) return;
+            if (schema->release == null) return;
```
eerhardt, line 238: Why is this check necessary?

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,122 @@
...
+            ptr->children = null;
+            ptr->dictionary = null;
+            ptr->release = null;
+            ptr->private_data = null;
+
+            return ptr;
+        }
+
+        /// <summary>
+        /// Free a pointer that was allocated in <see cref="AllocateAsPtr"/>.
```
eerhardt, line 65: What is `AllocateAsPtr`?

csharp/src/Apache.Arrow/C/CArrowSchemaImporter.cs
```diff
@@ -0,0 +1,285 @@
...
+                    "C" => new UInt8Type(),
+                    "s" => new Int16Type(),
+                    "S" => new UInt16Type(),
+                    "i" => new Int32Type(),
+                    "I" => new UInt32Type(),
+                    "l" => new Int64Type(),
+                    "L" => new UInt64Type(),
+                    "e" => new HalfFloatType(),
+                    "f" => new FloatType(),
+                    "g" => new DoubleType(),
```
eerhardt, line 240: These should all use the `.Default` property, where applicable. No reason to allocate extra objects here.

csharp/src/Apache.Arrow/C/CArrowSchemaImporter.cs
```diff
@@ -0,0 +1,285 @@
...
+        }
+
+        private sealed unsafe class ImportedArrowSchema : IDisposable
+        {
+            private readonly CArrowSchema* _data;
+            private readonly bool _isRoot;
+
+            public ImportedArrowSchema(CArrowSchema* handle)
+            {
+                _data = (CArrowSchema*)handle;
```
eerhardt, line 97:
```suggestion
                _data = handle;
```

eerhardt, line 97: (nit) The ctor parameter should be the same name as the backing field. It makes it easier to read and know that these things are the same thing.

csharp/src/Apache.Arrow/C/CArrowSchemaImporter.cs
```diff
@@ -0,0 +1,285 @@
...
+            using var importedSchema = new ImportedArrowSchema(ptr);
+            return importedSchema.GetAsSchema();
+        }
+
+        private sealed unsafe class ImportedArrowSchema : IDisposable
+        {
+            private readonly CArrowSchema* _data;
+            private readonly bool _isRoot;
+
+            public ImportedArrowSchema(CArrowSchema* handle)
```
eerhardt, line 95: Are we checking the `handle` pointer for `null`? This is passed directly here from a public API.

csharp/src/Apache.Arrow/C/CArrowSchemaImporter.cs
```diff
@@ -0,0 +1,285 @@
...
+            {
+                _isRoot = isRoot;
+            }
+
+            public void Dispose()
+            {
+                // We only call release on a root-level schema, not child ones.
+                if (_isRoot && _data->release != null)
+                {
+                    _data->release(_data);
```
eerhardt, line 115: I don't think we should be releasing anything that we don't own. `_data` was passed by the public API, right? We don't own it, we shouldn't dispose of it.

eerhardt (commented):
This is looking pretty good. I like the general structure you have now.

csharp/src/Apache.Arrow/C/CArrowSchemaImporter.cs
wjones127: Is there a way in C# to indicate that something is meant to take ownership of a pointer? TBH I think I would prefer that over relying on the user to remember to call the release callback in the appropriate situation.

For reference, that's how the C++ import functions work:

https://github.com/apache/arrow/blob/5b2fbade23eda9bc95b1e3854b19efff177cd0bd/cpp/src/arrow/c/bridge.h#L86-L94

csharp/src/Apache.Arrow/C/CArrowSchemaExporter.cs
wjones127: `release` is set to null in a release function to indicate it has already been released. This check makes it idempotent.

See: https://arrow.apache.org/docs/format/CDataInterface.html#release-callback-semantics-for-producers

csharp/src/Apache.Arrow/C/CArrowSchema.cs
wjones127: I thought about exposing it, but got worried it could be misused easily. If someone where to try to import a child schema instead of the root `CArrowSchema`, they would call the wrong release callback. Consumers "MUST not call any of its children’s release callbacks (including the optional dictionary)". See https://arrow.apache.org/docs/format/CDataInterface.html#release-callback-semantics-for-consumers

csharp/src/Apache.Arrow/C/CArrowSchema.cs
wjones127: That should be referencing `Create`. Fixed.

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// <see cref="ArrowFlagNullable" />, and <see cref="ArrowFlagMapKeysSorted" />.
+        /// </remarks>
+        public bool GetFlag(long flag)
+        {
+            return (flags & flag) == flag;
+        }
+
+        internal CArrowSchema* GetChild(int i)
+        {
+            if (i >= n_children)
```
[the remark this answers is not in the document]
wjones127: Cool!

csharp/src/Apache.Arrow/C/CArrowSchemaImporter.cs
wjones127: I've added a remark to these methods to indicate the release callback will be called. And I've added a unit test to make sure that it is called, even if the import fails.

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,122 @@
...
+        public long n_children;
+        public CArrowSchema** children;
+        public CArrowSchema* dictionary;
+        public delegate* unmanaged[Stdcall]<CArrowSchema*, void> release;
+        public void* private_data;
+
+
+        public static unsafe CArrowSchema* New()
+        {
+            var ptr = (CArrowSchema*)Marshal.AllocHGlobal(Marshal.SizeOf<CArrowSchema>());
```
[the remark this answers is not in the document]
adamreeve: Hi @eerhardt, would you mind explaining this a bit more? I would have thought that `Marshal.SizeOf` is the correct approach here as we want to know the size of the marshaled representation. In this case we might be lucky that this matches the size of the struct in managed memory as the struct only contains pointers and longs, and we don't need to worry about any ordering differences as all the members have the same size. I'd be concerned about future changes altering the struct representation and breaking this assumption, although that does seem unlikely. And I don't think we can assume applications will use NativeAOT or Trimming?

We could store the size in a static member if we want to avoid the overhead of the `Marshal.SizeOf` call each time, or just hard-code the size as a constant and add a check that this matches the marshaled size somewhere. I guess even just adding a test that `sizeof(CArrowSchema) == Marshal.SizeOf<CArrowSchema>()` would alleviate my concerns.

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,124 @@
...
+
+
+        /// <summary>
+        /// Whether this field is semantically nullable (regardless of whether it actually has null values)
+        /// </summary>
+        public const long ArrowFlagDictionaryOrdered = 1;
+        /// <summary>
+        /// For dictionary-encoded types, whether the ordering of dictionary indices is semantically meaningful.
+        /// </summary>
+        public const long ArrowFlagNullable = 2;
```
westonpace, line 92:
```suggestion
        /// <summary>
        /// For dictionary-encoded types, whether the ordering of dictionary indices is semantically meaningful.
        /// </summary>
        public const long ArrowFlagDictionaryOrdered = 1;
        /// <summary>
        /// Whether this field is semantically nullable (regardless of whether it actually has null values)
        /// </summary>
        public const long ArrowFlagNullable = 2;
```
Looks like these comments got swapped

csharp/src/Apache.Arrow/C/CArrowSchemaExporter.cs
```diff
@@ -0,0 +1,265 @@
...
+        /// Export a type to a <see cref="CArrowSchema"/>.
+        /// </summary>
+        /// <param name="datatype">The datatype to export</param>
+        /// <param name="schema">An allocated but uninitialized CArrowSchema pointer.</param>
+        /// <example>
+        /// <code>
+        /// CArrowSchema* exportPtr = CArrowSchema.Create();
+        /// CArrowSchemaExporter.ExportType(dataType, exportPtr);
+        /// foreign_import_function(exportPtr);
+        /// CArrowSchema.Free(exportPtr);
```
westonpace, line 38: Won't this trigger a call to `exportPtr->release`?  Is that safe given that you've already passed ownership during the `foreign_import_function`.

csharp/src/Apache.Arrow/C/CArrowSchemaExporter.cs
```diff
@@ -0,0 +1,265 @@
...
+            return flags;
+        }
+
+        private static unsafe CArrowSchema** ConstructChildren(IArrowType datatype, out long numChildren)
+        {
+            if (datatype is NestedType nestedType)
+            {
+                IReadOnlyList<Field> fields = nestedType.Fields;
+                int numFields = fields.Count;
+                numChildren = numFields;
```
westonpace, line 197: Is there a chance that `numChildren` could be 0?  Does `AllocHGlobal` return `IntPtr.Zero` in that case?

Maybe more of a concern on import but I seem to recall a zero-field union being a useless, but technically possible, data type.

csharp/src/Apache.Arrow/C/CArrowSchemaImporter.cs
```diff
@@ -0,0 +1,301 @@
...
+
+        /// <summary>
+        /// Import C pointer as an <see cref="Schema"/>.
+        /// </summary>
+        /// <remarks>
+        /// This will call the release callback on the passed struct, even if
+        /// this function fails.
+        /// </remarks>
+        /// <examples>
+        /// Typically, you will allocate a uninitialized CArrowSchema pointer,
```
westonpace, line 85:
```suggestion
        /// Typically, you will allocate an uninitialized CArrowSchema pointer,
```

csharp/src/Apache.Arrow/C/CArrowSchemaImporter.cs
```diff
@@ -0,0 +1,301 @@
...
+        /// </code>
+        /// </examples>
+        public static unsafe ArrowType ImportType(CArrowSchema* ptr)
+        {
+            using var importedType = new ImportedArrowSchema(ptr);
+            return importedType.GetAsType();
+        }
+
+        /// <summary>
+        /// Import C pointer as an <see cref="Field"/>.
```
westonpace, line 53:
```suggestion
        /// Import C pointer as a <see cref="Field"/>.
```

westonpace (commented):
This looks good.  I have a few remaining questions / nits.

csharp/src/Apache.Arrow/C/CArrowSchemaExporter.cs
wjones127: Okay read up on it. In the C++ bridge, the consumer moves the data a struct it owns by memcpy'ing it to the struct and then calling release on the source. So all import functions are expected to call `release` on the original struct (but might copy data off of it before then).

wjones127: https://github.com/apache/arrow/blob/427f1b649e36ef3c64dd61ed86808463a440eb94/cpp/src/arrow/c/helpers.h#L70-L75

westonpace: So let's split these into two points.

1. I agree that `CArrowSchema*` needs to be freed.  I'm not sure it has to be heap-allocated.  I think it could be stack-allocated but I don't remember if `struct` was guaranteed to be stack allocated or not.  It's probably large enough (with all the strings) to justify heap allocation so I think this is good.

2. I'm not sure that it is safe to assume that `release` has been set to null.  I do agree it is safe when calling Arrow-C++'s import function.  I just don't think this is part of the spec.  And, the spec is very clear that it is the consumer's responsibility to call the top-level `release` function.  So, while Arrow does copy everything into its own memory during it's import function I don't think this is spec-mandated.  Maybe it would be a good idea to add this mandate to the spec?

westonpace: Ah, regarding point 1, the strings are just pointers in the struct so...if you could mandate it is stack-allocated, I think that would be fine too.  But still, this is unlikely to really matter.  These structs/functions should be more or less invisible to the user most of the time.

wjones127: I was thinking that stack-allocated values in C# would be managed memory, and that wouldn't be accessible from C/other languages. But I might be mistaken on that.

wjones127: > I'm not sure that it is safe to assume that release has been set to null. I do agree it is safe when calling Arrow-C++'s import function. I just don't think this is part of the spec

You're right this isn't guaranteed in the spec. It's more something that needs to be negotiated in each API. So perhaps I just need to say something like that in the examples?

westonpace: Well, the problem is that you really DO need to release the `CArrowSchema*` regardless of what the foreign import function does.  So, is there a way to do that without calling `release` if it exists?

Maybe the pattern could be:

```
/// CArrowSchema* exportPtr = CArrowSchema.Create();
/// CArrowSchemaExporter.ExportType(dataType, exportPtr);
/// foreign_import_function(exportPtr);
/// exportPtr->release = IntPtr.Zero;
/// CArrowSchema.Free(exportPtr);
```

However, this seems potentially error prone.

wjones127: I think there are two hard rules:

 * `CArrowSchema` should only be free'd if `CArrowSchema.release` is null.
 * `CArrowSchema.release` should only be set to null when either the release callback is called or the data is moved. (So I think that line you added is not ideal)

If the `foreign_import_function` doesn't set it null, then a different pattern should be used. For example, the foreign importer should be the one to allocate the CArrowSchema and be responsible for freeing it. I think this is the reason that it's intended that the consumers allocate and own `CArrowSchema`, rather than the producer. But when interacting with R and Python, it's sometimes convenient to have the C++/Rust/... producer make these structs, so the move pattern is nice.

westonpace: I was all set to disagree with you. However, reading in detail, I think you're correct.

I don't like this but it might be the safest general pattern:

```
/// CArrowSchema* exportPtr = CArrowSchema.Create();
/// CArrowSchemaExporter.ExportType(dataType, exportPtr);
/// foreign_import_function(exportPtr);
/// if (exportPtr->release != IntPtr.Zero)
/// {
///     // Generally a foreign import function should copy the release callback
///     // and set this to IntPtr.Zero.  If they do not then it will not be possible for
///     // them to call the release callback.  Perhaps the imported expected to allocate
///     // the CArrowSchema instance itself?
///     CArrowSchema.Free(exportPtr);
///     throw new InvalidOperationException("Foreign import function failed to fully consume schema");
/// }
/// CArrowSchema.Free(exportPtr);
```

csharp/src/Apache.Arrow/C/CArrowSchemaImporter.cs
eerhardt: Can you show me in the code where `release` is called? I'm not seeing the code for it. (Yes, I do see the comment, but I don't see the backing code, nor the place in https://arrow.apache.org/docs/format/CDataInterface.html where it says that when you "import" a C Data schema object, the incoming pointer is released.

I feel this is a bad pattern - releasing resources that you don't own/didn't create. It would be really surprising to me that I couldn't get a `CArrowSchema*` from somewhere, and I couldn't call `CArrowSchemaImporter.ImportSchema` multiple times on it.

cc @pitrou who has the "git blame" on the above bridge.h file.

wjones127: > I feel this is a bad pattern - releasing resources that you don't own/didn't create.

To be clear: releasing doesn't necessarily mean freeing memory. It could also mean decrementing a reference count on a shared pointer. The release callback is just a way for the consumer to inform the producer that it is no longer using those resources. If the producer also isn't using those resources, it will free them. That's commonly the case with `ArrowSchema`, since it's all just small temporary values we created for exchange. But in other cases it might keep them around. That will be the case with `ArrowArray`, where the buffers will be behind reference counted pointers that the producer might still have references to.

> It would be really surprising to me that I couldn't get a CArrowSchema* from somewhere, and I couldn't call CArrowSchemaImporter.ImportSchema multiple times on it.

I think it's basically just a trade off. Either you can only import once, and the release callback is handled for you. Or you are allowed to import multiple times, but you are responsible for also calling the release callback (or else risking memory leaks). It seems the current design expects that most situations people only care to import something once, and that it's more important to prevent accidental memory leaks than it is support multiple imports.

> Can you show me in the code where release is called?

Indeed it's a little complicated:

 1. When you import, you create a `SchemaImporter`: https://github.com/apache/arrow/blob/08fb86107b001e2e1903ae0e416ff51386d4239e/cpp/src/arrow/c/bridge.cc#L1212-L1213
 2. When the schema importer is passed the `ArrowSchema`, it passes that to an `ExportGuard` https://github.com/apache/arrow/blob/08fb86107b001e2e1903ae0e416ff51386d4239e/cpp/src/arrow/c/bridge.cc#L831
 3. When the `SchemaImporter` destructor is called, the `ExportGuard` destructor is called, and that calls the release callback https://github.com/apache/arrow/blob/08fb86107b001e2e1903ae0e416ff51386d4239e/cpp/src/arrow/c/util_internal.h#L61

csharp/src/Apache.Arrow/C/CArrowSchema.cs
```diff
@@ -0,0 +1,124 @@
...
+    public unsafe struct CArrowSchema
+    {
+        public byte* format;
+        public byte* name;
+        public byte* metadata;
+        public long flags;
+        public long n_children;
+        public CArrowSchema** children;
+        public CArrowSchema* dictionary;
+        public delegate* unmanaged[Stdcall]<CArrowSchema*, void> release;
```
eerhardt, line 42: In the future (when we start targeting net6.0+), I think we will want to drop the `[Stdcall]` here and just use the default:

```
        public delegate* unmanaged<CArrowSchema*, void> release;
```

But that doesn't work in the older TFMs, so what we have now is fine.

csharp/src/Apache.Arrow/C/CArrowSchemaExporter.cs
```diff
@@ -0,0 +1,278 @@
...
+        /// CArrowSchemaExporter.ExportType(field, exportPtr);
+        /// foreign_import_function(exportPtr);
+        /// CArrowSchema.Free(exportPtr);
+        /// </code>
+        /// </example>
+        public static unsafe void ExportField(Field field, CArrowSchema* schema)
+        {
+            ExportType(field.DataType, schema);
+            schema->name = StringUtil.ToCStringUtf8(field.Name);
+            // TODO: field metadata
```
eerhardt, line 89: Can you log issues for these TODOs?

eerhardt (approved):
I just had a few more comments that can be addressed post merge.

Thanks for the contribution here, @wjones127!

I'll merge this early this week, unless I see more feedback.

csharp/src/Apache.Arrow/C/CArrowSchema.cs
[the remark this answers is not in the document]
eerhardt: Sorry I missed this question earlier.

See https://github.com/dotnet/runtime/issues/33802 for a lot of good discussion on this topic. In this case, `sizeof(CArrowSchema)` and `Marshal.SizeOf<CArrowSchema>()` will always return the same size. It is just that `sizeof(CArrowSchema)` is more efficient and is guaranteed to work in trimed/NativeAOT apps.

csharp/src/Apache.Arrow/C/CArrowSchemaImporter.cs
pitrou: > I feel this is a bad pattern - releasing resources that you don't own/didn't create

This is the entire point. Someone (the producer) decided to share those resources with you, and they need to know when you are finished using those resources.

As @wjones127 said, the release callback will generally not be a straight deallocation; instead, it may decrement a dedicated reference count (or unpin the backing object if the producer is implemented in a GC-based language).

eerhardt: I have 2 questions about this:

> they need to know when you are finished using those resources.

1. Isn't the fact that the function call returned to the caller enough to know that I'm finished using those resources?

>  it may decrement a dedicated reference count 

2. How do I "increment a dedicated reference count"? For example, if I received a `CArrowSchema*` from somewhere, and I need to pass it into a C# Importer and a C++ importer method. How could I tell the `CArrowSchema*` that it needs to allow `release` to be called twice?

wjones127: > Isn't the fact that the function call returned to the caller enough to know that I'm finished using those resources?

I think this will make more sense when you look at array import rather than schema import. The main point of the C data interface is to allow one implementation (the producer) to share array buffers with another implementation (the consumer). If the lifetime of those buffers was limited to the import function call, then that means the consumer would either be limited to using the arrays during the import or else have to copy the buffers (at which point you might as well use IPC).

As an example, imagine we used the C data interface to create bindings to DataFusion to run queries. 

```csharp
RecordBatch batch = ... 
DataFusionContext context;

// Uses C data interface to import batch (as a struct array)
context.ImportBatch("my_table", batch);

// This query is applied to the imported array
var result = context.Sql("SELECT sum(x * y) FROM my_table");
```

Here, if we assumed it was safe to cleanup the batch after calling `ImportBatch`, then we couldn't call the subsequent query to do any computation on it. We need those buffers to live as long as the consumer still needs them.

wjones127: > How do I "increment a dedicated reference count"? For example, if I received a CArrowSchema* from somewhere, and I need to pass it into a C# Importer and a C++ importer method. How could I tell the CArrowSchema* that it needs to allow release to be called twice?

I don't think you can import a `CArrowSchema*` more than once for that reason. So either the original producer would have to produce a `CArrowSchema` for each consumer, or else you could chain them, so that the C# importer imports the original one, and then the C# exporter creates a `CArrowSchema` that could be imported by C++.

csharp/src/Apache.Arrow/C/CArrowSchemaExporter.cs
pitrou: I think this discussion is a bit pedantic as it all depends how the consumer API (`foreign_import_function`) is defined, and that is not specified by the C Data Interface.

The C Data Interface only says what should happen with the _contents_ of the C struct. How the struct itself is allocated is left for the producer or consumer to decide (and may indeed depend on which language those are implemented in).

I would generally suggest that the producer handles the lifetime of the struct it populates, and the consumer moves the contents of the producer-populated struct to its own managed area.

(also, in real-world examples you would hope that `foreign_import_function` is able to return an error)

csharp/src/Apache.Arrow/C/CArrowSchemaExporter.cs
pitrou: Well, if the release callback is already null, the consumer cannot have called this function...

csharp/src/Apache.Arrow/C/CArrowSchemaExporter.cs
wjones127: I've created an umbrella issue to track the remaining C data interface tasks (https://github.com/apache/arrow/issues/33856) and another one to track remaining data types to implement (https://github.com/apache/arrow/issues/34736).

csharp/test/Apache.Arrow.Tests/CDataInterfaceSchemaTests.cs
```diff
@@ -0,0 +1,427 @@
...
+            Assert.Throws<NullReferenceException>(() =>
+            {
+                CArrowSchemaImporter.ImportType(cSchema);
+            });
+            Assert.True(wasCalled);
+            CArrowSchema.Free(cSchema);
+        }
+    }
+
+    public class CDataSchemaPythonTest
```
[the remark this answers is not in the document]
wjones127: Done.

eerhardt:
@westonpace - I see you still have changes requested. Is that stale? or are you still requesting changes on the current proposal?

The macOS 12 Python 3 leg is still failing. Does anyone know if that is related to these changes? @wjones127, can you merge with `main` to see if the errors go away?

wjones127:
@eerhardt I think we can ignore that error for now. The test is flakey and we are still looking into the root cause.

https://github.com/apache/arrow/issues/34743

westonpace (approved):
Sorry, my review was stale.  I'm happy with where this is now.

csharp/src/Apache.Arrow/C/CArrowSchema.cs
pitrou: @westonpace Arrow C++ does build on 32-bit Linux (we have nightly crossbow tests for that).

PR #34134, opened 2023-02-10
GH-34118: [C++][Python] Make # of S3 event loop threads configurable

### LINKED ISSUE #34118: Allow configuration of size of AWS event loop thread pool

When calling `DoInitializeS3`, arrow creates initialises the AWS API, which by default creates a thread pool for the background AWS event loop that uses one thread per physical core on the system.

This is particularly noticeable when using pyarrow where merely importing `from pyarrow.fs import PyFileSystem` will cause eager initialisation of the AWS API, so you're paying (something) for these threads even when you have no intention of using them.

This is rather unfriendly when running a multi-process or some otherhow parallelised process on a multicore box since it leads to oversubscription. Moreover, it may well not be the best option depending on the number of concurrent connections this arrow process is going to be making to the AWS api: quoth [the documentation](https://awslabs.github.io/aws-crt-cpp/class_aws_1_1_crt_1_1_io_1_1_event_loop_group.html)

> The number of threads used depends on your use-case. IF you have a maximum of less than a few hundred connections 1 thread is the ideal threadCount.

It would be nice if there were a way to control the size of this thread pool in the same way that one can control the number of IO threads arrow uses using `ARROW_IO_THREADS`. [Aside: AFAICT there's no programmatic way of control _arrow's_ thread pool size, it must be done via environment variables, which is also rather unfriendly].

I think the following diff is kind of a sketch in this direction, although it just unilaterally sets the size of the thread pool available to a single thread.

```patch
diff --git a/cpp/src/arrow/filesystem/s3fs.cc b/cpp/src/arrow/filesystem/s3fs.cc
index 16ffe2526..a71ec93f7 100644
--- a/cpp/src/arrow/filesystem/s3fs.cc
+++ b/cpp/src/arrow/filesystem/s3fs.cc
@@ -2604,6 +2604,13 @@ Status DoInitializeS3(const S3GlobalOptions& options) {
   // This configuration options is only available with AWS SDK 1.9.272 and later.
   aws_options.httpOptions.compliantRfc3986Encoding = true;
 #endif
+  aws_options.ioOptions.clientBootstrap_create_fn = []() {
+    Aws::Crt::Io::EventLoopGroup eventLoopGroup(1);
+    Aws::Crt::Io::DefaultHostResolver defaultHostResolver(eventLoopGroup, 8, 30);
+    auto clientBootstrap = Aws::MakeShared<Aws::Crt::Io::ClientBootstrap>(ALLOCATION_TAG, eventLoopGroup, defaultHostResolver);
+    clientBootstrap->EnableBlockingShutdown();
+    return clientBootstrap;
+  };
   Aws::InitAPI(aws_options);
   aws_initialized.store(true);
   return Status::OK();
```

I'm not really familiar with the arrow layout so I don't know how to plumb this in to any configuration options that might abound: is there such a thing or would one just introduce a (new?) env var?

### Component(s)

C++

westonpace:
This also changes the default # of threads to 1 per advice in the linked PR.
* Closes: #34118


westonpace:
Under the category of "nothing is ever easy" it appears that some older AWS SDK versions do not support I/O options as they were not using the CRT library under the hood.  This was causing the ` R / AMD64 Windows C++ RTools 40 ucrt64 (pull_request) ` build to fail.  I've attempted to add a cmake check for this and conditionally set the option.

westonpace:
CC @benibus @felipecrv do one of you want to take a look?

cpp/src/arrow/CMakeLists.txt
```diff
@@ -497,6 +497,17 @@ if(ARROW_FILESYSTEM)
     list(APPEND ARROW_SRCS filesystem/hdfs.cc)
   endif()
   if(ARROW_S3)
+
```
felipecrv, line 500: nitpick: blank line can go :scissors:

cpp/src/arrow/CMakeLists.txt
```diff
@@ -497,6 +497,17 @@ if(ARROW_FILESYSTEM)
...
+    try_compile(S3_HAS_CRT ${CMAKE_CURRENT_BINARY_DIR}/try_compile
+                SOURCES "${CMAKE_CURRENT_SOURCE_DIR}/filesystem/try_compile/check_s3fs_crt.cc"
+                CMAKE_FLAGS "-DINCLUDE_DIRECTORIES=${CURRENT_INCLUDE_DIRECTORIES}"
+                LINK_LIBRARIES ${AWSSDK_LINK_LIBRARIES} CXX_STANDARD 17)
+
+    if(S3_HAS_CRT)
+      message(STATUS "AWS SDK is new enought to have CRT support")
```
felipecrv, line 507: *enough

felipecrv (approved):

benibus (commented):
No major objections on first glance, just a minor suggestion.

cpp/src/arrow/CMakeLists.txt
westonpace: Thanks, removed.

cpp/src/arrow/filesystem/s3fs.cc
```diff
@@ -2591,6 +2596,20 @@ Status DoInitializeS3(const S3GlobalOptions& options) {
...
+  aws_options.ioOptions.clientBootstrap_create_fn =
+      [ev_threads = options.num_event_loop_threads]() {
+        // https://github.com/aws/aws-sdk-cpp/blob/1.11.15/src/aws-cpp-sdk-core/source/Aws.cpp#L65
+        Aws::Crt::Io::EventLoopGroup eventLoopGroup(ev_threads);
+        Aws::Crt::Io::DefaultHostResolver defaultHostResolver(
+            eventLoopGroup, /*maxHosts=*/8, /*maxTTL=*/30);
+        auto clientBootstrap = Aws::MakeShared<Aws::Crt::Io::ClientBootstrap>(
+            "Aws_Init_Cleanup", eventLoopGroup, defaultHostResolver);
+        clientBootstrap->EnableBlockingShutdown();
+        return clientBootstrap;
```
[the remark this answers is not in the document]
westonpace: Ah, good catch.  Not intentional, thanks.  I've made this change.

felipecrv (approved):

PR #34137, opened 2023-02-11
GH-34136: [C++] Add a concept of ordering to ExecPlan

### LINKED ISSUE #34136: [C++] Add the concept of "ordering" to an exec node, reject non-sensible plans

Every node has an "ordering" which describes what the batch index of the batches produced by that node corresponds to.

Source nodes will generally have an ordering based on the data they are generating.  For in-memory data this will be implicit "row number" ordering (done in this PR).  For the scanner this will generally be implicit "file number / row number" ordering (future PR).  In both cases the user should eventually be able to assert an explicit ordering if they know for a fact that their data is ordered (future PR).  If the data is partitioned the scanner could generate an explicit ordering based on the partition columns (future PR).

Mapping nodes like filter/project/fetch will simply propagate the ordering of their input.

Some nodes, such as order_by (future PR) and asof join (this PR) will establish a new ordering.

Other nodes, such as the aggregate node, the hash-join node, and the union node will destroy the ordering (this PR).  Their output will have no meaningful order.  Although, in all these cases one could devise an alternative order-preserving approach (future PR):

 * An aggregate node could pretty easily establish an ordering based on the keys or an implicit ordering if there are no keys (if you only have 1 row then it is ordered :)
 * A hash-join node could similarly choose to order by one or more of the join keys (unclear at the moment if there would be much cost associated with this)
 * To preserve ordering, a union node would need to read from all inputs at the same time and output in round-robin fashion.  Alternatively, a union node could choose to fully drain one input before starting on the next input.  Both approaches would have performance implications

The fetch node will now fail to validate if it is placed in a plan that does not make sense (e.g. immediately following a hash-join node).

The sink node, if sequenced output is requested, will similarly fail to validate if the plan does not have an established ordering.

This all sounds far more complex than it is :)

### Component(s)

C++

westonpace:
In addition, it is now possible to bypass the I/O executor on the `record_batch_source`, `exec_batch_source`, and `array_vector_source`.

It is now possible to create a source node from a `gen::Gen` generator.

BREAKING CHANGE: The default executor for `record_batch_source`, `exec_batch_source`, and `array_vector_source` was (erroneously) the plan's CPU executor.  It now defaults properly to the I/O executor.
* Closes: #34136

### DIFF
--- a/cpp/src/arrow/compute/exec/exec_plan.h
+++ b/cpp/src/arrow/compute/exec/exec_plan.h
@@ -28,6 +28,7 @@
 
 #include "arrow/compute/api_vector.h"
 #include "arrow/compute/exec.h"
+#include "arrow/compute/ordering.h"
 #include "arrow/compute/type_fwd.h"
 #include "arrow/type_fwd.h"
 #include "arrow/util/future.h"
@@ -110,9 +111,6 @@ class ARROW_EXPORT ExecPlan : public std::enable_shared_from_this<ExecPlan> {
   std::string ToString() const;
 };
 
-static const std::optional<std::vector<SortKey>> kImplicitOrdering = {{}};
-static const std::optional<std::vector<SortKey>> kUnordered = std::nullopt;
-
 class ARROW_EXPORT ExecNode {
  public:
   using NodeVector = std::vector<ExecNode*>;
@@ -160,16 +158,14 @@ class ARROW_EXPORT ExecNode {
   /// know that all values of x in a batch with index N will be less than
   /// or equal to all values of x in a batch with index N+k (assuming k > 0)
   ///
-  /// Note that there is an important distinction between std::nullopt and
-  /// an empty vector.  If the value is std::nullopt then the node is asserting
-  /// that there is no predictable output order to its batches.  For example,
-  /// a hash-join has no predictable output order.
+  /// Note that an ordering can be both Ordering::Unordered and Ordering::Implicit.
+  /// A node's output should be marked Ordering::Unordered if the order is
+  /// non-deterministic.  For example, a hash-join has no predictable output order.
   ///
-  /// If the value is an empty vector then the data is "implicitly ordered".
-  /// This means there is a meaningful order but that odering is not represented
-  /// by any column in the data.  The most common case for this is when reading
-  /// data from an in-memory table.  The data has an implicit "row order" which
-  /// is not neccesarily represented in the data set.
+  /// If the ordering is Ordering::Implicit then there is a meaningful order but that
+  /// odering is not represented by any column in the data.  The most common case for this
+  /// is when reading data from an in-memory table.  The data has an implicit "row order"
+  /// which is not neccesarily represented in the data set.
   ///
   /// A typical map node will not modify the ordering.  Nothing needs to be done
   /// other than ensure the index assigned to output batches is the same as the
@@ -190,7 +186,7 @@ class ARROW_EXPORT ExecNode {
   /// Nodes that maintain ordering should be careful to avoid introducing gaps
   /// in the batch index.  This may require emitting empty batches in order to
   /// maintain continuity.
-  virtual const std::optional<std::vector<SortKey>>& ordering() const;
+  virtual const Ordering& ordering() const;
 
   /// Upstream API:
   /// These functions are called by input nodes that want to inform this node
--- a/cpp/src/arrow/compute/exec/exec_plan.h
+++ b/cpp/src/arrow/compute/exec/exec_plan.h
@@ -156,7 +156,10 @@ class ARROW_EXPORT ExecNode {
   ///
   /// In other words, given the ordering {{"x", SortOrder::Ascending}} we
   /// know that all values of x in a batch with index N will be less than
-  /// or equal to all values of x in a batch with index N+k (assuming k > 0)
+  /// or equal to all values of x in a batch with index N+k (assuming k > 0).
+  /// Furthermore, we also know that values will be sorted within a batch.
+  /// Any row N will have a value of x that is less than the value for
+  /// any row N+k.
   ///
   /// Note that an ordering can be both Ordering::Unordered and Ordering::Implicit.
   /// A node's output should be marked Ordering::Unordered if the order is
@@ -167,7 +170,7 @@ class ARROW_EXPORT ExecNode {
   /// is when reading data from an in-memory table.  The data has an implicit "row order"
   /// which is not neccesarily represented in the data set.
   ///
-  /// A typical map node will not modify the ordering.  Nothing needs to be done
+  /// A filter or project node will not modify the ordering.  Nothing needs to be done
   /// other than ensure the index assigned to output batches is the same as the
   /// input batch that was mapped.
   ///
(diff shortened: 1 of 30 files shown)


westonpace:
This PR should be complete but it builds on #34059 which has not yet merged.

lidavidm (approved):

westonpace:
@lidavidm Sorry, I requested your review a bit too early.  I started working on the order by node and realized that the ordering I had in place didn't support the ability to specify null placement.  So I changed it to a proper class similar to SortNodeOptions.  The change is pretty minor but I'll leave up unless you want to take another look.

cpp/src/arrow/compute/ordering.h
```diff
@@ -0,0 +1,120 @@
...
+  bool Equals(const Ordering& other) const;
+  std::string ToString() const;
+
+  bool is_implicit() const { return is_implicit_; }
+  bool is_unordered() const { return !is_implicit_ && sort_keys_.empty(); }
+
+  std::vector<SortKey> sort_keys() { return sort_keys_; }
+  NullPlacement null_placement() { return null_placement_; }
+
+  static const Ordering& Imiplicit() {
```
lidavidm, line 96: typo: Implicit

lidavidm (approved):

cpp/src/arrow/compute/ordering.h
westonpace: :facepalm:

lidavidm (approved):

cpp/src/arrow/compute/exec/exec_plan.h
```diff
@@ -146,6 +148,46 @@ class ARROW_EXPORT ExecNode {
...
 
+  /// \brief the ordering of the output batches
+  ///
+  /// This does not guarantee the batches will be emitted by this node
+  /// in order.  Instead it guarantees that the batches will have their
+  /// ExecBatch::index property set in a way that respects this ordering.
+  ///
+  /// In other words, given the ordering {{"x", SortOrder::Ascending}} we
+  /// know that all values of x in a batch with index N will be less than
+  /// or equal to all values of x in a batch with index N+k (assuming k > 0)
```
jorisvandenbossche, line 159: If there is an order based on a column "x", does that also guarantee something about the order _within_ each batch? (or only between batches as this paragraph explains)

cpp/src/arrow/compute/exec/exec_plan.h
```diff
@@ -146,6 +148,46 @@ class ARROW_EXPORT ExecNode {
...
+  /// Note that an ordering can be both Ordering::Unordered and Ordering::Implicit.
+  /// A node's output should be marked Ordering::Unordered if the order is
+  /// non-deterministic.  For example, a hash-join has no predictable output order.
+  ///
+  /// If the ordering is Ordering::Implicit then there is a meaningful order but that
+  /// odering is not represented by any column in the data.  The most common case for this
+  /// is when reading data from an in-memory table.  The data has an implicit "row order"
+  /// which is not neccesarily represented in the data set.
+  ///
+  /// A typical map node will not modify the ordering.  Nothing needs to be done
```
jorisvandenbossche, line 170: What is exactly meant with "map node"? (I don't find this term used anywhere else in the compute / Acero code or docs) Do we mean a node that uses typical element-wise scalar kernels? Also something like a filter node will always preserve ordering.

cpp/src/arrow/compute/exec/exec_plan.h
```diff
@@ -146,6 +148,46 @@ class ARROW_EXPORT ExecNode {
...
+  /// new ordering based on the hash keys).
+  ///
+  /// Some nodes will require an ordering.  For example, a fetch node or an
+  /// asof join node will only function if the input data is ordered (for fetch
+  /// it is enough to be implicitly ordered.  For an asof join the ordering must
+  /// be explicit and compatible with the on key.)
+  ///
+  /// Nodes that maintain ordering should be careful to avoid introducing gaps
+  /// in the batch index.  This may require emitting empty batches in order to
+  /// maintain continuity.
```
jorisvandenbossche, line 188: Just wondering, but assume you would do a filter operation that filters a large part of the data, and so you might end up with several empty batches, does that affect for example sink nodes like writing files? (do we skip empty batches there, or do we then potentially write empty files for them?)

cpp/src/arrow/compute/exec/exec_plan.h
jorisvandenbossche: OK, I searched for "map node" and not for "MapNode" ;) 
And I see that indeed both Project and Filter node inherit from MapNode (it's still a term that is not used in our documentation though)

jorisvandenbossche (commented):
Didn't look in detail at the C++ code, mostly took a look to see if I could follow the new logic, and added a few questions. 

One more question: assume you would use something like `DeclarationToTable`, does that automatically use the ordering / batch indices if there is one, or do you still need to indicate this manually you want to use it? (like the generic `SinkNodeOptions` has a `sequence_delivery` parameter with a default of false, but I don't see that exposed through the DeclarationTo.. versions)

And looking forward to have this!

jorisvandenbossche:
> One more question: assume you would use something like `DeclarationToTable`, does that automatically use the ordering / batch indices if there is one, or do you still need to indicate this manually you want to use it? (like the generic `SinkNodeOptions` has a `sequence_delivery` parameter with a default of false, but I don't see that exposed through the DeclarationTo.. versions)

Ah, I see there is a new variant of `DeclarationToTable` that uses QueryOptions as parameter, and that options struct has a `sequence_output` keyword. So I assume that answers my question.

Follow-up question on this: the default for `sequence_output` is false, so does that mean that even if you have a query plan with an order_by node at the end before consuming as eg a table, you won't get the data as ordered by default, because for the ordering to be respected, you need to explicitly enable it with `sequence_output`?

cpp/src/arrow/compute/exec/exec_plan.h
westonpace: Yes, it should guarantee that the ordering exists within the batch as well.

cpp/src/arrow/compute/exec/exec_plan.h
westonpace: Ah, yes, this is probably an internal term.  I'll update this.

cpp/src/arrow/compute/exec/exec_plan.h
westonpace: Good question.  The dataset writer will discard empty batches without writing anything.  However, the sink node still respects empty batches.  For example, if one were doing `dataset.to_batches(...)` then they might see an empty batch.

I'm fairly certain this is consistent with the current implementation and not a change in behavior.

westonpace:
> even if you have a query plan with an order_by node at the end before consuming as eg a table, you won't get the data as ordered by default, because for the ordering to be respected, you need to explicitly enable it with sequence_output?

Correct.  Maybe the default should be true when the input to the sink node is ordered.  This would mean we only default to false if there is an aggregate or join.  Given the cost of this sequencing should generally be pretty reasonable I think it would be an ok default (and users could still disable it if they wanted).

cpp/src/arrow/compute/exec/exec_plan.h
westonpace: I've updated the wording here to mention ordering within a batch.

cpp/src/arrow/compute/exec/exec_plan.h
westonpace: I changed this to "A filter or project node..."

westonpace:
> Correct. Maybe I should change to an optional and the default (nullopt) would sequence when the input to the sink node is ordered. This would mean we only default to false if there is an aggregate or join. Given the cost of this sequencing should generally be pretty reasonable I think it would be an ok default (and users could still disable it if they wanted).

I've done this.  The default (nullopt) means "sequence if there is any ordering".  It can be set to true to get "fail validation if there is no meaningful ordering" or false to get "never sequence and maximize performance even if there is a meaningful ordering".

cpp/src/arrow/compute/exec/exec_plan.h
```diff
@@ -156,7 +156,10 @@ class ARROW_EXPORT ExecNode {
   ///
   /// In other words, given the ordering {{"x", SortOrder::Ascending}} we
   /// know that all values of x in a batch with index N will be less than
-  /// or equal to all values of x in a batch with index N+k (assuming k > 0)
+  /// or equal to all values of x in a batch with index N+k (assuming k > 0).
+  /// Furthermore, we also know that values will be sorted within a batch.
+  /// Any row N will have a value of x that is less than the value for
```
jorisvandenbossche, line 161:
```suggestion
  /// Any row N will have a value of x that is less than or equal to the value for
```

jorisvandenbossche:
This is ready to be merged?

westonpace:
> This is ready to be merged?

Yes, thanks.

PR #34140, opened 2023-02-11
GH-15173: [C++][Parquet] Fixing ByteStreamSplit Standard broken

### LINKED ISSUE #15173: [Parquet][C++] ByteStreamSplitDecoder broken in presence of nulls

ByteStreamSplitDecoder is initialized https://github.com/apache/arrow/blob/ceec7950e8c6e9a63f48d27c284c56938df3598d/cpp/src/parquet/encoding.cc#L2908 with a `num_values` parameter. The code assumes that this is the number of non-null values but in fact this is the value count including nulls. This breaks decoding in two ways. The first is the check `num_values * static_cast<int64_t>(sizeof(T)) > len` which won't be true if there are nulls. The second is that `num_values` is used as the stride parameter for decoding the stream instead of the non-null value count, so even if the aforementioned check is removed, the decoding will be wrong.  The java code in parquet-mr seems to have similar problems...

I think there are two different approaches to fixing this. The first is to ascertain the non-null value count before decoding; with V2 page headers this can be determined from the header but for V1 headers this might require decoding all levels first. The second approach is to assume that the length of the values data is what it is and divide this length by the element size to get the stride. This seems to be compatible with the way both the C++ and Java encoders write these pages but it probably should be specified in the specification that there is no padding / extraneous byte at the end of the page.

### Component(s)

C++, Parquet

mapleFU:
### Rationale for this change

This patch fixes bugs in BYTE_STREAM_SPLIT. It enforce the parquet ByteStreamSplit not to use padding, and force using `sizeof(data) / sizeof(T)` as num-of-values.

1. Update num-of-values in ByteStreamSplit
2. Add testing

### Are these changes tested?

Yes

No

* Closes: #15173


mapleFU:
@pitrou @wgtmac @wjones127 PTAL

cpp/src/parquet/encoding.cc
```diff
@@ -2937,11 +2937,16 @@ ByteStreamSplitDecoder<DType>::ByteStreamSplitDecoder(const ColumnDescriptor* de
 template <typename DType>
 void ByteStreamSplitDecoder<DType>::SetData(int num_values, const uint8_t* data,
                                             int len) {
-  DecoderImpl::SetData(num_values, data, len);
-  if (num_values * static_cast<int64_t>(sizeof(T)) > len) {
-    throw ParquetException("Data size too small for number of values (corrupted file?)");
+  if (num_values * static_cast<int64_t>(sizeof(T)) < len) {
+    throw ParquetException("Data size too large for number of values (padding file?)");
```
wjones127, line 2941: It might be nice to be more specific.
```suggestion
    throw ParquetException("Data size too large for number of values (padding in byte stream split data page?)");
```

wjones127 (commented):
Looks good. Just one suggestion on the exception description.

mapleFU:
Comment fixed

PR #34156, opened 2023-02-13
GH-34150: [C++] [Python] Fix improper initialization of ConversionOptions

### LINKED ISSUE #34150: [C++] [Python] Fix improper initialization of `ConversionOptions`

In a debugging session, I observed that `ConversionOptions` being passed from `_substrait.pyx` was improperly initialized, leading to a SIGSEGV. In particular, the extension provider member had a different value than `default_extension_provider()` returns. I suspect this is because `ConversionOptions` is currently a POD, whose initialization of the extension provider member occurs before that of `g_default_extension_provider` being returned from `default_extension_provider`. The fix should ensure correct order of initialization. In addition, the definition of `CConversionOptions` in `libarrow_substrait.pxd` should be updated to match `ConversionOptions`.

### Component(s)

C++, Python

rtpsw:
This PR fixes a bug that caused a crash. See #34150 for more details.
* Closes: #34150


rtpsw:
@westonpace, could you suggest what tests to add?

python/pyarrow/includes/libarrow_substrait.pxd
```diff
@@ -34,10 +50,26 @@ cdef extern from "arrow/engine/substrait/options.h" namespace "arrow::engine" no
         BEST_EFFORT \
             "arrow::engine::ConversionStrictness::BEST_EFFORT"
 
+    cdef cppclass CExtensionDetails \
+            "arrow::ending::ExtensionDetails"
```
icexelloss, line 54: This should be "engine"?

Does it not error out if this is wrong?

python/pyarrow/includes/libarrow_substrait.pxd
```diff
@@ -34,10 +50,26 @@ cdef extern from "arrow/engine/substrait/options.h" namespace "arrow::engine" no
...
+
+    cdef cppclass CExtensionProvider \
+            "arrow::ending::ExtensionProvider":
```
icexelloss, line 57: engine

icexelloss:
@rtpsw IIUC, there are two changes in this PR:
(1) Fix ConversionOptions initialization 
(2) Add python bindings for ExtensionDetails, set_default_extension_provider, ExtensionProvider, ExtensionSet, etc

If that is correct, I think we don't necessarily need to mix the two - the first is a bug fix and second is additional functionality.

I am comfortable with approving (1) and but I am not sure about (2) so probably need Weston's input (I believe VD folks are in a conference so there might be delay)

python/pyarrow/includes/libarrow_substrait.pxd
rtpsw: It doesn't error in CI either. Surprising.

rtpsw:
> @rtpsw IIUC, there are two changes in this PR: (1) Fix ConversionOptions initialization (2) Add python bindings for ExtensionDetails, set_default_extension_provider, ExtensionProvider, ExtensionSet, etc
> 
> If that is correct, I think we don't necessarily need to mix the two - the first is a bug fix and second is additional functionality.
> 
> I am comfortable with approving (1) and but I am not sure about (2) so probably need Weston's input (I believe VD folks are in a conference so there might be delay)

The reason for (2) is because API signatures of `ConversionOptions` bring in other symbols. There might be some symbols we could reduce, but not all of them.

icexelloss:
I see - so (2) is required to fix (1) - ok

icexelloss:
@rtpsw This is tested internally and works? (I am surprised that nothing breaks because of the engine typo)

rtpsw:
> @rtpsw This is tested internally and works? 

There are no new tests, but the change in `ConversionOptions` is covered by existing tests. The difficulty with a distinguishing test for this change, i.e., a test that fails pre-PR code but passes post-PR, is that it would need to somehow tinker with the order of initialization (see #34150) - there is no precedent of doing this in Arrow that I know of.

> (I am surprised that nothing breaks because of the engine typo)

Yes, that surprised me too. I'll try to figure out why.

rtpsw:
> Yes, that surprised me too. I'll try to figure out why.

My investigation suggests that Cython only parses `pxd` files, not compile them, but it does compile any symbol in a `pxd` file that is used in a `pyx` file that is compiles. In the [recent commit](https://github.com/apache/arrow/pull/34156/commits/eb8231834c75697b5b32e12874b506cdd7498fa3), I added a dummy function that "touches" the symbols added to the `pxd` here. I observed that typos were caught by Cython, and fixing them allowed the compilation to succeed.

icexelloss:
Thanks - This makes sense. The PR looks good to me. But probably need a final approval from @westonpace

python/pyarrow/includes/libarrow_substrait.pxd
```diff
@@ -34,10 +51,27 @@ cdef extern from "arrow/engine/substrait/options.h" namespace "arrow::engine" no
...
+
+    cdef shared_ptr[CExtensionProvider] default_extension_provider()
+    cdef void set_default_extension_provider(
+        const shared_ptr[CExtensionProvider]& provider)
+
     cdef cppclass CConversionOptions \
             "arrow::engine::ConversionOptions":
+        CConversionOptions()
         ConversionStrictness conversion_strictness
         function[CNamedTableProvider] named_table_provider
```
icexelloss, line 73: Hold on - we are missing named_tap_provider here - does that not trigger any error?

rtpsw: This doesn't trigger an error, and this is expected. Cython doesn't care if a symbol exists in C++ but is not exposed in a `pxd`. This allows symbols in a `pxd` to be included on a need-basis. If you'd like this symbol to be exposed, I could add (and touch) it.

icexelloss: Hmm..why do we expose extension_provider and not named_tap_provider then?

icexelloss: I see. I am indifferent about exposing named_tap_provider - but we feel we either (1) expose all or (2) expose minimal. (Although I am not sure which one Pyarrow generally follows)

Does fixing the bug require exposing extension_provider in here?

rtpsw: > (Although I am not sure which one Pyarrow generally follows)

I don't know myself, but my guess is it is mixed. You'd probably find pretty much all public Arrow symbols in PyArrow, but not all in sub-components like Arrow Substrait.

> Does fixing the bug require exposing extension_provider in here?

The bug is due to `ConversionOptions` default constructor, so it has to be exposed. I'll check if the bug is fixed with a minimal change including it that I can get to compile.

westonpace:
Does #34209 work?  I think static locals should be easier because:

 * Works without modifying python (which means we don't have to repeat this in R)
 * Minor benefit to load time because the singleton is only initialized if used
 * Consistent with some of our newer handling of singletons (sorry for not suggesting this earlier)

rtpsw:
It will take a bit of time for me to check, but I suspect static locals would only fix initialization order in C++ but not in Cython. Specifically, I suspect that Cython sees different header definitions for `ConversionOptions` (those generated from the `pxd` file) than C++ does, and doesn't see the inline initialization of its members in the C++ header, leading to an improperly initialized object. I don't mind using static locals to fix the C++ order of initialization, but I'd still suggest adding a constructor to `ConversionOptions`, where its members are initialized, and expose it to Cython. I added a reduced changeset here for just this exposure.

westonpace:
Static locals should fix cython.  Initialization is no longer a global thing and has nothing to do with header order.  The fields will be initialized the first time `default_extension_provider` is called.  So there is no way for cython to mess it up.

rtpsw:
> Static locals should fix cython. Initialization is no longer a global thing and has nothing to do with header order. The fields will be initialized the first time `default_extension_provider` is called. So there is no way for cython to mess it up.

I used your code but, unfortunately, I'm still seeing the same SIGSEGV due to an improperly initialized `extension_provider` member of `ConversionOptions`. How about combining both your and my changes?

westonpace:
> Update: I confirmed in the same env that switching from inline member initialization to a constructor in ConversionOptions fixes the SIGSEGV.

That is odd.  As long as it doesn't involve python changes I don't mind moving to using constructor initialization over member initialization (the Arrow codebase is not at all consistent here)

python/pyarrow/_substrait.pyx
```diff
@@ -26,6 +27,13 @@ from pyarrow.includes.libarrow cimport *
 from pyarrow.includes.libarrow_substrait cimport *
 
 
+def _check_conversion_options():
+    cdef:
+        CConversionOptions conv_opts
+
+    return conv_opts.strictness == BEST_EFFORT
+
+
```
westonpace, line 36:
```suggestion
```

Let's get rid of this.  It'll just confuse future maintainers more than it helps I think.

python/pyarrow/tests/test_substrait.py
```diff
@@ -34,6 +34,10 @@
 pytestmark = [pytest.mark.dataset, pytest.mark.substrait]
 
 
+def test_conversion_options():
+    assert substrait._check_conversion_options()
+
+
```
westonpace, line 40:
```suggestion
```

westonpace (changes requested):
The move to constructor initialization is fine.  Let's get rid of the test utility.

westonpace (approved):
Thanks!

rtpsw:
Thanks, @westonpace. I was able to use the latest Arrow nightly version to fix the bug.

Do you think #34209 should also be merged?

Issue #34111, opened 2023-02-10
[CI][Packaging][Java] java-jars was broken by updating bundled aws-sdk-cpp


kou:
### Describe the bug, including details regarding any error messages, version, and platform.

#33808 broke java-jars job:

https://github.com/apache/arrow/pull/33808#issuecomment-1424873183

https://github.com/ursacomputing/crossbow/actions/runs/4138641959/jobs/7155286836#step:6:4265

```text
Undefined symbols for architecture x86_64:
  "_SSLClose", referenced from:
      _s_handle_shutdown in libaws-c-io.a(secure_transport_tls_channel_handler.c.o)
  "_SSLCopyPeerTrust", referenced from:
      _s_drive_negotiation in libaws-c-io.a(secure_transport_tls_channel_handler.c.o)
  "_SSLCreateContext", referenced from:
      _s_tls_handler_new in libaws-c-io.a(secure_transport_tls_channel_handler.c.o)
  "_SSLHandshake", referenced from:
      _s_drive_negotiation in libaws-c-io.a(secure_transport_tls_channel_handler.c.o)
  "_SSLRead", referenced from:
      _s_process_read_message in libaws-c-io.a(secure_transport_tls_channel_handler.c.o)
  "_SSLSetCertificate", referenced from:
      _s_tls_handler_new in libaws-c-io.a(secure_transport_tls_channel_handler.c.o)
  "_SSLSetConnection", referenced from:
      _s_tls_handler_new in libaws-c-io.a(secure_transport_tls_channel_handler.c.o)
  "_SSLSetIOFuncs", referenced from:
      _s_tls_handler_new in libaws-c-io.a(secure_transport_tls_channel_handler.c.o)
  "_SSLSetPeerDomainName", referenced from:
      _s_tls_handler_new in libaws-c-io.a(secure_transport_tls_channel_handler.c.o)
  "_SSLSetProtocolVersionMin", referenced from:
      _s_tls_handler_new in libaws-c-io.a(secure_transport_tls_channel_handler.c.o)
  "_SSLSetSessionOption", referenced from:
      _s_tls_handler_new in libaws-c-io.a(secure_transport_tls_channel_handler.c.o)
  "_SSLWrite", referenced from:
      _s_process_write_message in libaws-c-io.a(secure_transport_tls_channel_handler.c.o)
  "_SecCertificateCreateWithData", referenced from:
[...]

PR #34158, opened 2023-02-13
GH-34157: [C++] Configure bundled AWS SDK to use aws-lc instead of OpenSSL

### LINKED ISSUE #34157: [C++] Configure bundled AWS SDK to use aws-lc instead of OpenSSL

Using OpenSSL causes various issues like https://github.com/apache/arrow/pull/33808#issuecomment-1408247269 and https://github.com/apache/arrow/issues/34111. We should try to use aws-lc for libcrypto and libssl. We need to hide them inside s2n-tls to avoid name conflicts with OpenSSL used by other libraries.

### Component(s)

C++

js8544:
Using OpenSSL causes various issues like https://github.com/apache/arrow/pull/33808#issuecomment-1408247269 and https://github.com/apache/arrow/issues/34111. We should try to use aws-lc for libcrypto and libssl. We need to hide them inside s2n-tls to avoid name conflicts with OpenSSL used by other libraries.

* Closes: #34157


js8544:
Closed in favor of https://github.com/apache/arrow/pull/34159

Issue #34044, opened 2023-02-04
[Go] Build fails with noasm build tag


nwt:
### Describe the bug, including details regarding any error messages, version, and platform.

A Go build with the noasm build tag fails.
```console
$ git describe --dirty --tags
apache-arrow-12.0.0.dev-41-gecb1f2e31

$ go build -tags noasm ./...
# github.com/apache/arrow/go/v12/arrow/compute/internal/kernels
arrow/compute/internal/kernels/basic_arithmetic_noasm.go:27:9: undefined: getGoArithmeticOpFloatingSameType
arrow/compute/internal/kernels/basic_arithmetic_noasm.go:35:9: undefined: getGoArithmeticFixedIntOut
```

### Component(s)

Go

zeroshade:
Issue resolved by pull request 34045
https://github.com/apache/arrow/pull/34045

PR #34167, opened 2023-02-13
GH-34055: [Go][CI] Add test run in CI that uses noasm tag

### LINKED ISSUE #34055: [Go][CI] Add test run in CI that uses `noasm` tag

#34044 found that the build failed with the `noasm` tag, to prevent this from happening again we should add a CI run to build and test using the `noasm` tag.

### Component(s)

Continuous Integration, Go

zeroshade:
### Rationale for this change
A bug was found in the `noasm` build for the Go lib in #34044 so adding a CI run with the `noasm` tag should prevent that from happening again in the future.

Updating the `go_test.sh` script to allow for an env var to tell it to add the `noasm` tag to the `go test` command

Updating go.yml to add a `noasm` CI run

No

* Closes: #34055


.github/workflows/go.yml
```diff
@@ -369,3 +369,40 @@ jobs:
...
+        with:
+          fetch-depth: 0
+          submodules: recursive
+      - name: Install go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ matrix.go }}
+          cache: true
+          cache-dependency-path: go/go.sum
+      - name: Install staticcheck
```
lidavidm, line 402: side note, why isn't this one of the things archery runs? does it need to build the code to work?

lidavidm (approved):

.github/workflows/go.yml
zeroshade: It's part of the building of the Go docker images, which is why this step doesn't exist in jobs that use the archery images. I just didn't feel like making an entirely separate docker image for building/running with the `noasm` tag.

If you look at the primary jobs which use archery, they don't have a separate `install staticcheck` step, it's in the dockerfile.

lidavidm: Sure, I just mean that - is it necessary to run staticcheck separately in all pipelines, vs just once in the Archery pipeline?

assignUser: Can we not run `go_test.sh` with the noasm flag in one of the exisisting jobs instead of creating another 2 jobs? The go matrix is pretty big already :D

zeroshade: @lidavidm we call staticcheck in the `go_test.sh` file so that's why I did it here, though technically we don't necessarily need it in all the pipelines but we should have it run in the both `asm` and `noasm` pipelines since it changes which files actually get included.

that said, I can definitely shoehorn this into one of the existing jobs instead as @assignUser requested :smile: I'll update this tomorrow to avoid making another job

zeroshade:
@assignUser updated to just have `go_test.sh` run the tests twice, once without and once with the `noasm` tag. So all the existing jobs just get the additional run of tests using `noasm` now.

assignUser (approved):
Thanks!

PR #34168, opened 2023-02-13
GH-33880: [C++] Improve I/O tracing

### LINKED ISSUE #33880: Add/improve tracing in the dataset writer

The telemetry code in the dataset writer currently does not trace the asynchronous tasks that are submitted to the I/O thread pool that perform the actual encoding, compression, and writing.

Follow-up to https://github.com/apache/arrow/pull/33738, where @westonpace noted: "Mentally, when I think of the dataset writer, I think there are two parts. The first part should be the trailing part of the fragment/pipeline that feeds the writer. In this first part we partition the batch, select the appropriate file queues, and deposit the batches into the queues. There is then a separate dedicated thread task to write each batch to the writer."

### Component(s)

C++

joosthooz:
### Rationale for this change
The structure of the opentelemetry traces have changed significantly in this recent PR https://github.com/apache/arrow/pull/33738, this PR aims to improve some things that were left out of scope at that time. Specifically this PR aims to improve the tracing of the I/O side of things. We'll add spans for low-level I/O calls (e.g. `write`) and portions of the Parquet writer and Datasetwriter. The goal is to be able to distinguish work performed on the I/O threadpool versus the normal threadpool and see how much time was actually spent on I/O or on (de/en)coding and (de)compression.

### Are these changes tested?
This is a very early WIP

Arrow releases do not have telemetry enabled so these changes should not affect normal users. But users that are interested in the performance of Acero will likely want to use the telemetry. For these users, the structure and details of the spans will likely be changed by this PR. The current documentation of the telemetry does not include any details about this yet.

* Closes: #33880


mapleFU:
What's the status of this patch? Can the patch be review now?

joosthooz:
> What's the status of this patch? Can the patch be review now?

@mapleFU Sorry for the delayed response, and thanks for checking in. I didn't have time to work on this but I'll try to get this out of draft by the end of the week!

joosthooz:
@mapleFU I think this is ready for some feedback. This PR requires some additional context. I've focused mostly on the Dataset writers and readers, and I think that some of the "normal" readers/writers are still not yet instrumented with spans. I think we should leave that out of scope. As noted in the description above, this PR is a follow-up to a previous PR that changed the structure of traces. Before, there were top-level spans for each graph node, and each item of data would have a child span underneath each of those.
Here, the goal is to use spans' parent-child relation to show how morsels flow through the graph:
image
These morsels are produced by the readers:
image
But because of how the code is written, I have not been able to connect these together. In the ideal world, there would be a connection between ReadNextAsync -> MakeChunkedBatchGenerator -> ProcessMorsel. But that would require more invasive code changes (passing along a span context with an ExecBatch, for example).
image
What I'd still like to do is somehow add an attribute to certain spans whose duration can be aggregated to get some sensible total (like CPU time spent on the read side, each kernel, and the write side of things). And then I've been working on a notebook to visualize a trace into a per-thread activity overview, it would be nice if we can add something like that but I'm not sure how.
Looking forward to any feedback

cpp/src/arrow/csv/reader.cc
```diff
@@ -456,6 +463,23 @@ class BlockDecodingOperator {
...
+            -> Result<DecodedBlock> {
+          ARROW_ASSIGN_OR_RAISE(auto decoded_arrays,
+                                arrow::internal::UnwrapOrRaise(maybe_decoded_arrays));
+
+          ARROW_ASSIGN_OR_RAISE(auto batch,
+                                state->DecodedArraysToBatch(std::move(decoded_arrays)));
+          raw_span->SetAttribute("arrow.csv.output_batch_size_bytes",
+                                 util::TotalBufferSize(*batch));
+          return DecodedBlock{std::move(batch), bytes_parsed_or_skipped};
+        });
```
westonpace, line 481: I think there is too much duplicated code in these alternate paths. Is there any way we can just pass `span` to the callback's capture list and use `ATTRIBUTE_ON_CURRENT_SPAN`?  Also, what happened to `END_SPAN_ON_FUTURE_COMPLETION`?

cpp/src/arrow/csv/reader.cc
```diff
@@ -911,9 +964,25 @@ class StreamingReaderImpl : public ReaderMixin,
...
 
     auto self = shared_from_this();
-    return rb_gen().Then([self, rb_gen, max_readahead](const DecodedBlock& first_block) {
-      return self->InitFromBlock(first_block, std::move(rb_gen), max_readahead, 0);
+    auto init_finished = rb_gen().Then([self, rb_gen, max_readahead
+#ifdef ARROW_WITH_OPENTELEMETRY
+                                        ,
+                                        init_span = std::move(init_span),
+                                        read_span = std::move(read_span)
+#endif
```
westonpace, line 972: Why is this `ifdef` block needed?  If tracing is disabled then `init_span` and `read_span` are an empty unique_ptr.  Copying it around should be trivial.

cpp/src/arrow/csv/reader.cc
```diff
@@ -911,9 +964,25 @@ class StreamingReaderImpl : public ReaderMixin,
...
+    ](const DecodedBlock& first_block) {
+      auto fut = self->InitFromBlock(first_block, std::move(rb_gen), max_readahead, 0);
+#ifdef ARROW_WITH_OPENTELEMETRY
+      opentelemetry::nostd::shared_ptr<opentelemetry::trace::Span> raw_span =
+          ::arrow::internal::tracing::UnwrapSpan(read_span.details.get());
+      raw_span->SetAttribute("batch.size_bytes",
+                             util::TotalBufferSize(*first_block.record_batch));
+#endif
```
westonpace, line 980: Why not `ATTRIBUTE_ON_CURRENT_SPAN`?

cpp/src/arrow/dataset/dataset_writer.cc
```diff
@@ -202,7 +207,18 @@ class DatasetWriterFileQueue {
...
     rows_currently_staged_ += delta_staged;
-    staged_batches_.push_back(std::move(batch));
+    {
+      util::tracing::Span span;
+      START_SPAN(span, "DatasetWriter::Push",
+                 {{"batch.size_rows", batch->num_rows()},
+                  {"rows_currently_staged", rows_currently_staged_},
+                  // staged_rows_count is updated at the end, after this push and possibly
+                  // multiple pops
+                  //        {"staged_rows_count", writer_state_->staged_rows_count},
```
westonpace, line 217:
```suggestion
```

cpp/src/arrow/dataset/dataset_writer.cc
```diff
@@ -202,7 +207,18 @@ class DatasetWriterFileQueue {
...
     uint64_t delta_staged = batch->num_rows();
     rows_currently_staged_ += delta_staged;
-    staged_batches_.push_back(std::move(batch));
+    {
+      util::tracing::Span span;
+      START_SPAN(span, "DatasetWriter::Push",
+                 {{"batch.size_rows", batch->num_rows()},
+                  {"rows_currently_staged", rows_currently_staged_},
+                  // staged_rows_count is updated at the end, after this push and possibly
+                  // multiple pops
```
westonpace, line 216: I'm not sure I understand this comment

cpp/src/arrow/dataset/dataset_writer.cc
```diff
@@ -233,6 +249,18 @@ class DatasetWriterFileQueue {
...
+          uint64_t size_bytes = util::TotalBufferSize(*batch);
+          uint64_t num_buffers = 0;
+          for (auto column : batch->columns()) {
+            num_buffers += column->data()->buffers.size();
+          }
+          util::tracing::Span span;
+          START_SPAN(span, "DatasetWriter::WriteNext",
+                     {{"threadpool", "IO"},
+                      {"batch.size_bytes", size_bytes},
+                      {"batch.num_buffers", num_buffers}});
```
westonpace, line 262: Why do I need to know `num_buffers`?

cpp/src/arrow/dataset/dataset_writer.cc
```diff
@@ -611,12 +639,14 @@ class DatasetWriter::DatasetWriterImpl {
       backpressure =
           writer_state_.rows_in_flight_throttle.Acquire(next_chunk->num_rows());
       if (!backpressure.is_finished()) {
+        EVENT(scheduler_->span(), "DatasetWriter::Backpressure::TooManyRowsQueued");
```
westonpace, line 642: Why both `EVENT` and `EVENT_ON_CURRENT_SPAN`?

cpp/src/arrow/util/async_util.cc
```diff
@@ -294,6 +282,25 @@ class ThrottledAsyncTaskSchedulerImpl
...
+        return (*target)();
+      }
+      int cost() const override { return target->cost(); }
+      std::string_view name() const override { return target->name(); }
+      std::unique_ptr<Task> target;
+      opentelemetry::nostd::shared_ptr<opentelemetry::trace::Span> parent_span;
+    };
+    task = std::make_unique<WrapperTask>(
+        std::move(task), arrow::internal::tracing::GetTracer()->GetCurrentSpan());
+#endif
```
westonpace, line 303: This is exactly the same `WrapperTask` implementation as above.  Can we remove the duplication?

cpp/src/arrow/util/async_util.cc
```diff
@@ -168,6 +146,25 @@ class AsyncTaskSchedulerImpl : public AsyncTaskScheduler {
...
```
westonpace, line 167: I'm not sure this works the same.  `task` here is perhaps a bit of a misnomer.  It returns a future which does the actual work.  So you've wrapped the "submission" part of a task which is normally not the very interesting part (usually it just schedules the task on an executor and returns the future).

Before, I was starting the task here and then marking it finished when the task's future actually completes.  So it was wrapping both the submission and the task itself.

cpp/src/arrow/util/async_util.cc
```diff
@@ -306,7 +313,10 @@ class ThrottledAsyncTaskSchedulerImpl
     std::optional<Future<>> maybe_backoff = throttle_->TryAcquire(latched_cost);
     if (maybe_backoff) {
 #ifdef ARROW_WITH_OPENTELEMETRY
-      TraceTaskQueued(task.get(), span());
+      EVENT(span(), "Task submission throttled",
+            {{"task.name", ::opentelemetry::nostd::string_view(task->name().data(),
+                                                               task->name().size())},
+             {"task.cost", task->cost()}});
```
westonpace, line 319: `span()` is the scheduler's span.  This puts an event on the scheduler's span that submission was throttled.  I don't think that's terribly interesting.

What I was aiming for with `TraceTaskQueued` was to actually start the task's span.  Then, later, in `TraceTaskSubmitted`, it would register a "task submitted" event.

So a throttled task (that is queued for 1 time unit and takes 2 time units to run) would look something like this...

```
T+0 Task Span Begins
T+1 Task Submitted
T+3 Task Ends
```

A non-throttled task would look something like:

```
T+0 Task Begins
T+2 Task Ends
```

cpp/src/arrow/util/async_util.cc
```diff
@@ -168,6 +146,25 @@ class AsyncTaskSchedulerImpl : public AsyncTaskScheduler {
...
+        auto scope = arrow::internal::tracing::GetTracer()->WithActiveSpan(parent_span);
+        return (*target)();
+      }
+      int cost() const override { return target->cost(); }
+      std::string_view name() const override { return target->name(); }
+      std::unique_ptr<Task> target;
+      opentelemetry::nostd::shared_ptr<opentelemetry::trace::Span> parent_span;
+    };
+    task = std::make_unique<WrapperTask>(
+        std::move(task), arrow::internal::tracing::GetTracer()->GetCurrentSpan());
```
westonpace, line 166: Note that I was intentionally not using `GetCurrentSpan` as the parent.  We are submitting a new task here.  I would like that new task to appear as its own span and not as a child of the thing that submitted it.  For example, in Acero, in most plans, the source node is the one that schedules all the tasks.  If we use `GetCurrentSpan` here then it will look like all the work being done is a part of the source node.

cpp/src/arrow/util/io_util.cc
```diff
@@ -1628,6 +1629,8 @@ Result<int64_t> FileRead(int fd, uint8_t* buffer, int64_t nbytes) {
   HANDLE handle = reinterpret_cast<HANDLE>(_get_osfhandle(fd));
 #endif
   int64_t total_bytes_read = 0;
+  util::tracing::Span span;
+  START_SPAN(span, "FileRead", {{"fd", fd}});
```
westonpace, line 1633: This is very deep in the I/O & filesystem hierarchy. Maybe it would be better to put these spans in `LocalFileSystem` instead?

cpp/src/parquet/arrow/reader.cc
```diff
@@ -286,13 +287,21 @@ class FileReaderImpl : public FileReader {
...
                 {"parquet.arrow.physicaltype", phys_type},
-                {"parquet.arrow.records_to_read", records_to_read}});
-#endif
+                { "parquet.arrow.records_to_read",
+                  records_to_read }});
+
+    auto status = reader->NextBatch(records_to_read, out);
+
+    uint64_t size_bytes = ::arrow::util::TotalBufferSize(*out->get());
+    ATTRIBUTE_ON_CURRENT_SPAN("parquet.arrow.output_batch_size_bytes", size_bytes);
```
westonpace, line 300: Why do we need to guard `ATTRIBUTE_ON_CURRENT_SPAN` with an ifdef?  Couldn't we have left the ifdef where it was and moved `::arrow::util::TotalBufferSize` inside the macro?

docs/source/cpp/opentelemetry.rst
```diff
@@ -85,3 +85,88 @@ at http://localhost:16686. Note that unlike with other methods of exporting
...
+Each morsel has a toplevel span called ProcessMorsel.
+Currently, the DatasetScanner cannot connect its output to the ProcessMorsel 
+spans due to the asynchronous structure of the code.
+The dataset writer will gather batches of data in its staging area, and will 
+issue a write operation once it has enough rows.
+This is represented by the DatasetWriter::Push and DatasetWriter::Pop spans.
+These also carry the current fill level of the staging area.
+This means that some Morsels will not trigger a write.
+Only if a morsel causes the staging area to overflow its threshold,
+a DatasetWriter::Pop is triggered that will perform a write operation.
```
westonpace, line 155: Acero has a task scheduler.  I've had the most luck when I structured spans in this way:

 * There is one span for the plan itself that starts when `StartProducing` is called and ends when the plan finishes.
 * Within that span there is a span for the scheduler.  It should have more or less the same lifetime (these top two spans can be combined if desired)
 * As a child of that each thread task has its own span.
 * Various operations (filtering / projection / etc.) are children of their thread task and may be nested as needed (e.g. a filtering span may have child spans for the expression evaluation)

This is sort of what I had in `async_util.cc` but I think these changes have altered that.

westonpace (changes requested):
I think we can tackle this PR more easily as multiple PRs but I will leave it up to you.  We could have on PR for Acero related things and one for the I/O & readers.

The spans in the CSV reader seem overly obtrusive, I'd like to see if we can find a way to avoid paths where we have an `#ifdef` and then repeat logic in both the if and else blocks.

I've made a few comments about how tracing was working in Acero with the task scheduler.  I don't think the changes here are necessarily improving the way that was handled.  Maybe we can talk this through in more detail.

PR #34170, opened 2023-02-13
GH-33618: [C++] add option needs_extended_file_info and implement localFS

### LINKED ISSUE #33618: [C++] Improve arrow::fs::FileSelect performance for `IsFile()` and `IsDirectory()`

Here is an example `std::filesystem`:

```
for (const auto &path : std::filesystem::directory_iterator(dataDir)) {
    if (path.is_regular_file()) {
     // logic here
    }
  }
 ```
 
Here is a similar example using `arrow::fs::FileSelect()`:

```
  auto selector = arrow::fs::FileSelector();
  selector.base_dir = dir;
  selector.allow_not_found = false;
  selector.recursive = false;
  auto fileInfo = fs_->GetFileInfo(selector);
  for (const auto &pathInfo: fileInfo.ValueOrDie()) {
    pathInfo.IsFile();
    // OR
    pathInfo.IsDirectory();
  }
  ```
  
The `arrow::fs::FileSelector()` example takes 10x as long. Under the hood there is a stat() call which fills in extra metadata information, and this increases time spent making the call. `std::filesystem` instead uses `readdir`, which checks whether an entry is a file or directory, but does not provide extra metadata information. This results in a much faster system call.

`std::filesystem::directory_iterator` gives the user a `directory_entry` which does have a `file_size` attribute. Calling `file_size` the first time triggers a stat call to the file but a stat call is not made if you don't call `file_size`. 
 
 I propose a similar adjustment to the `arrow::fs` API, which allows a user to perform an entry identity check with a `readdir` system call, such that a `stat` call is only used if the user requests the extra metadata (e.g. file size and mtime).

### Component(s)

C++

anjakefala:
### Rationale for this change

When set to false on localFS, use `std::filesystem` to set FileType instead of `stat()`. Does nothing for the other filesystems.

Improves performance 10x on checks whether entities are directories or files on localFS. file size and mtime are not set.

Added variable `needs_extended_file_info` (default `true`) to struct `FileSelector`.

When `false`, `StatSelector` (called by `GetFileInfo (FileSelector)`) will call `IdentifyFile` instead of `StatFile`. `IdentifyFile` will then use `std::filesystem` to set the `.type`, instead of using `stat()`.

### Are these changes tested?

A basic test was added.

Yes, an optional variable was added to the `FileSelector` struct. Would a cookbook example or docstring update be appropriate?
* Closes: #33618

### DIFF
--- a/cpp/src/arrow/filesystem/filesystem.h
+++ b/cpp/src/arrow/filesystem/filesystem.h
@@ -132,7 +132,10 @@ struct ARROW_EXPORT FileSelector {
   bool recursive;
   /// The maximum number of subdirectories to recurse into.
   int32_t max_recursion;
-  /// If true, use stat() to obtain extended file metadata.
+  /// If false, may skip retrieving size and modification time metadata.
+  ///
+  /// On local filesystem, setting to false avoids an additional system call.
+  /// On other filesystems, this may have no effect.
   bool needs_extended_file_info;
 
   FileSelector()
--- a/cpp/src/arrow/filesystem/filesystem.h
+++ b/cpp/src/arrow/filesystem/filesystem.h
@@ -132,8 +132,14 @@ struct ARROW_EXPORT FileSelector {
   bool recursive;
   /// The maximum number of subdirectories to recurse into.
   int32_t max_recursion;
+  /// If true, use stat() to obtain extended file metadata.
+  bool needs_extended_file_info;
 
-  FileSelector() : allow_not_found(false), recursive(false), max_recursion(INT32_MAX) {}
+  FileSelector()
+      : allow_not_found(false),
+        recursive(false),
+        max_recursion(INT32_MAX),
+        needs_extended_file_info(true) {}
 };
 
 /// \brief FileSystem, path pair
--- a/cpp/src/arrow/filesystem/localfs.cc
+++ b/cpp/src/arrow/filesystem/localfs.cc
@@ -228,7 +228,7 @@ Result<FileInfo> IdentifyFile(const std::filesystem::path path) {
   info.set_mtime(kNoTime);
   info.set_size(kNoSize);
   // TODO does it need to be wstring in windows?
-  info.set_path(path.generic_string());
+  info.set_path(path.string());
 
   return info;
 }
@@ -263,7 +263,14 @@ Status StatSelector(const PlatformFilename& dir_fn, const FileSelector& select,
 
 Status IdentifyFileSelector(const PlatformFilename& dir_fn, const FileSelector& select,
                             int32_t nesting_depth, std::vector<FileInfo>* out) {
-  // TODO handle case where it does not exist
+  ARROW_ASSIGN_OR_RAISE(bool exists, FileExists(dir_fn));
+  if (!exists) {
+    if (select.allow_not_found) {
+      return Status::OK();
+    }
+    return Status::Invalid("No such directory: ", dir_fn.ToString());
+  }
+
   auto result = std::filesystem::directory_iterator(dir_fn.ToString());
 
   for (const auto& path : result) {
@@ -274,9 +281,9 @@ Status IdentifyFileSelector(const PlatformFilename& dir_fn, const FileSelector& 
     }
     if (nesting_depth < select.max_recursion && select.recursive &&
         info.type() == FileType::Directory) {
-      // TODO again, does the Windows wstring case need to be handledS
+      // TODO again, does the Windows wstring case need to be handled
       ARROW_ASSIGN_OR_RAISE(auto path_pf,
-                            PlatformFilename::FromString(path.path().generic_string()));
+                            PlatformFilename::FromString(path.path().string()));
       RETURN_NOT_OK(IdentifyFileSelector(path_pf, select, nesting_depth + 1, out));
     }
   }
--- a/cpp/src/arrow/filesystem/localfs.cc
+++ b/cpp/src/arrow/filesystem/localfs.cc
@@ -17,6 +17,7 @@
 
 #include <chrono>
 #include <cstring>
+#include <filesystem>
 #include <memory>
 #include <sstream>
 #include <utility>
@@ -210,6 +211,22 @@ Result<FileInfo> StatFile(const std::string& path) {
   return info;
 }
 
+Result<FileInfo> IdentifyFile(const std::string& path) {
+  FileInfo info;
+
+  if (std::filesystem::is_directory(path)) {
+    info.set_type(FileType::Directory);
+  } else if (std::filesystem::is_regular_file(path)) {
+    info.set_type(FileType::File);
+  }
+
+  info.set_mtime(kNoTime);
+  info.set_size(kNoSize);
+  info.set_path(path);
+
+  return info;
+}
+
 #endif
 
 Status StatSelector(const PlatformFilename& dir_fn, const FileSelector& select,
@@ -228,7 +245,12 @@ Status StatSelector(const PlatformFilename& dir_fn, const FileSelector& select,
 
   for (const auto& child_fn : *result) {
     PlatformFilename full_fn = dir_fn.Join(child_fn);
-    ARROW_ASSIGN_OR_RAISE(FileInfo info, StatFile(full_fn.ToNative()));
+    FileInfo info;
+    if (select.needs_extended_file_info == true) {
+      ARROW_ASSIGN_OR_RAISE(info, StatFile(full_fn.ToNative()));
+    } else {
+      ARROW_ASSIGN_OR_RAISE(info, IdentifyFile(full_fn.ToNative()));
+    }
     if (info.type() != FileType::NotFound) {
       out->push_back(std::move(info));
     }
--- a/cpp/src/arrow/filesystem/localfs.cc
+++ b/cpp/src/arrow/filesystem/localfs.cc
@@ -211,6 +211,8 @@ Result<FileInfo> StatFile(const std::string& path) {
   return info;
 }
 
+#endif
+
 Result<FileInfo> IdentifyFile(const std::string& path) {
   FileInfo info;
 
@@ -226,8 +228,6 @@ Result<FileInfo> IdentifyFile(const std::string& path) {
 
   return info;
 }
-
-#endif
 
 Status StatSelector(const PlatformFilename& dir_fn, const FileSelector& select,
                     int32_t nesting_depth, std::vector<FileInfo>* out) {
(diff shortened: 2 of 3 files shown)


cpp/src/arrow/filesystem/filesystem.h
```diff
@@ -132,8 +132,14 @@ struct ARROW_EXPORT FileSelector {
   bool recursive;
   /// The maximum number of subdirectories to recurse into.
   int32_t max_recursion;
-
-  FileSelector() : allow_not_found(false), recursive(false), max_recursion(INT32_MAX) {}
+  /// If true, use stat() to obtain extended file metadata.
```
wjones127, line 135: Since the default is `true`, maybe better to frame as what happens if false. Also maybe preferable to frame in implementation-agnostic way (stat isn't invoked on Windows, and isn't applicable to remote filesystems):
```suggestion
  /// If false, may skip retrieving size and modification time metadata.
  ///
  /// On local filesystem, setting to false avoids an additional system call.
  /// On other filesystems, this may have no effect.
```

cpp/src/arrow/filesystem/localfs.cc
```diff
@@ -210,6 +211,22 @@ Result<FileInfo> StatFile(const std::string& path) {
   return info;
 }
 
+Result<FileInfo> IdentifyFile(const std::string& path) {
```
wjones127, line 214: This is defined within the POSIX branch (else) of the `#ifdef _WIN32` condition above. Since it's platform agnostic, I think you can pull this out to the top-level.

cpp/src/arrow/filesystem/localfs_test.cc
```diff
@@ -472,6 +472,24 @@ TYPED_TEST(TestLocalFS, StressGetFileInfoGenerator) {
   }
 }
 
+TYPED_TEST(TestLocalFS, NeedsExtendedFileInfo) {
+  // Test setting needs_extended_file_info to true and false
```
wjones127, line 476: Comment isn't accurate. I think the true case is well covered by other tests though.
```suggestion
  // Test setting needs_extended_file_info to false
```

wjones127:
As a bonus, it would be interesting to add a benchmark to [localfs_benchmark.cc](https://github.com/apache/arrow/blob/master/cpp/src/arrow/filesystem/localfs_benchmark.cc). This would make sure we detect if there's ever a regression in the performance of this function.

wjones127 (changes requested):

cpp/src/arrow/filesystem/localfs.cc
```diff
@@ -210,6 +211,22 @@ Result<FileInfo> StatFile(const std::string& path) {
...
+  FileInfo info;
+
+  if (std::filesystem::is_directory(path)) {
+    info.set_type(FileType::Directory);
+  } else if (std::filesystem::is_regular_file(path)) {
+    info.set_type(FileType::File);
+  }
```
anjakefala, line 221: `std::filesystem` supports a bunch of file types: https://en.cppreference.com/w/cpp/filesystem. Should we add any of the others to `else if`? Add a case for an empty path?

wjones127: We should probably ignore entries that aren't files or directories. Except maybe symlinks.

cpp/src/arrow/filesystem/localfs.cc
```diff
@@ -228,7 +247,12 @@ Status StatSelector(const PlatformFilename& dir_fn, const FileSelector& select,
 
   for (const auto& child_fn : *result) {
     PlatformFilename full_fn = dir_fn.Join(child_fn);
-    ARROW_ASSIGN_OR_RAISE(FileInfo info, StatFile(full_fn.ToNative()));
+    FileInfo info;
+    if (select.needs_extended_file_info == true) {
+      ARROW_ASSIGN_OR_RAISE(info, StatFile(full_fn.ToNative()));
+    } else {
+      ARROW_ASSIGN_OR_RAISE(info, IdentifyFile(full_fn.ToNative()));
```
wjones127, line 254: From [docs on std::filesystem::status](https://en.cppreference.com/w/cpp/filesystem/status) (which `std::filesystem::is_directory` calls):

> The information provided by this function is usually also provided as a byproduct of directory iteration, and may be obtained by the member functions of [filesystem::directory_entry](https://en.cppreference.com/w/cpp/filesystem/directory_entry). During directory iteration, calling status again is unnecessary.

The original example that seemed more performant used `std::filesystem::directory_iterator`, but this approach seems to instead make separate calls again for each path instead of using the information provided by the directory iterator. Is there a reason you moved away from the original approach?

anjakefala: Good point! I am making changes to address this. I was trying to avoid duplicating bunches of `StatSelector` code basically. But there are better ways of handling that.

anjakefala:
A note that all `TODO`s will be removed before I mark the PR as ready! I just wanted to see the CI results with the current changes.

cpp/src/arrow/filesystem/localfs_test.cc
```diff
@@ -472,6 +472,55 @@ TYPED_TEST(TestLocalFS, StressGetFileInfoGenerator) {
...
+    if (info.path() == "AB") {
+      AssertFileInfo(info, "AB", FileType::Directory);
+    } else if (info.path() == "AB/CD") {
+      AssertFileInfo(info, "AB/CD", FileType::Directory);
+    } else if (info.path() == "AB/ab") {
+      AssertFileInfo(info, "AB/ab", FileType::File);
+    } else {
+      // TODO how would I raise an error in a test?
+      ASSERT_TRUE(false);
+    }
```
wjones127, line 513: Instead of this if loop, perhaps you could use `SortInfos` like the other tests and assert equality of each element. See example here:

https://github.com/apache/arrow/blob/9651563d6e8d34ddd43d5882de8134393590d4ce/cpp/src/arrow/filesystem/s3fs_test.cc#L730-L735

Same applies to the earlier for loop.

anjakefala: Ah, I could not find an example of how inconsistent ordering was handled with the `FileSelector` case! Thank you!

cpp/src/arrow/filesystem/localfs.cc
```diff
@@ -310,7 +361,11 @@ Result<std::vector<FileInfo>> LocalFileSystem::GetFileInfo(const FileSelector& s
   RETURN_NOT_OK(ValidatePath(select.base_dir));
   ARROW_ASSIGN_OR_RAISE(auto fn, PlatformFilename::FromString(select.base_dir));
   std::vector<FileInfo> results;
-  RETURN_NOT_OK(StatSelector(fn, select, 0, &results));
+  if (select.needs_extended_file_info == true) {
```
wgtmac, line 364:
```suggestion
  if (select.needs_extended_file_info) {
```

cpp/src/arrow/filesystem/localfs.cc
```diff
@@ -240,6 +261,36 @@ Status StatSelector(const PlatformFilename& dir_fn, const FileSelector& select,
   return Status::OK();
 }
 
+Status IdentifyFileSelector(const PlatformFilename& dir_fn, const FileSelector& select,
```
wgtmac, line 264: It seems to share a lot of common logic from `StatSelector`. Is is possible to modify it directly?

cpp/src/arrow/filesystem/localfs.cc
```diff
@@ -212,6 +213,26 @@ Result<FileInfo> StatFile(const std::string& path) {
...
+  } else if (std::filesystem::is_regular_file(path) ||
+             std::filesystem::is_symlink(path)) {
+    info.set_type(FileType::File);
+  } else {
+    info.set_type(FileType::Unknown);
+  }
+
+  info.set_mtime(kNoTime);
+  info.set_size(kNoSize);
+  // TODO does it need to be wstring in windows?
```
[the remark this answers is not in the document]
anjakefala: This is a draft! Part of the goal was to ensure all TODOs were addressed before merging. But, I wanted to communicate the work that I knew remained.

cpp/src/arrow/filesystem/localfs.cc
anjakefala: `IdentifyFile` requires an `std::filesystem::path` while `StatFile` requires `std::wstring` or `std::string`. Because it can use `std::filesystem::path` `IdentifyFile` takes advantage of the performance boost of `directory_iterator` for its loop. Basically, There is enough unique logic that combining them is looking uglier. The main thing they have in common is the boilerplate at the start that checks for the existence of the file. I think the better thing to do is to factor out that code, instead of trying to combine the rest.

anjakefala:
Test failure on AMD64 Windows C++:

```
[ RUN      ] TestLocalFS/1.NeedsExtendedFileInfo
D:/a/arrow/arrow/cpp/src/arrow/filesystem/test_util.cc:137: Failure
  info.type()
    Which is: FileType::Unknown
  type
    Which is: FileType::Directory
For path 'AB'
D:/a/arrow/arrow/cpp/src/arrow/filesystem/localfs_test.cc:494: Failure
  infos.size()
    Which is: 1
  3
[  FAILED  ] TestLocalFS/1.NeedsExtendedFileInfo, where TypeParam = arrow::fs::internal::ExtendedLengthPathFormatter (1 ms)
```

amoeba:
I'm taking a look at the test failure above and after that just finishing up this PR. So far I was able to reproduce the failure locally. The `FileInfo` that fails the assertion has an absolute path of `"//?/C:/msys64/tmp/test-localfs-u2pia6rh/AB"` which looks problematic so I'm going to look at what's going on there next.

amoeba:
I'm seeing now that this only fails for the ExtendedLengthPathFormatter variant and, on WIN32, we prepend `//?/`, see:

https://github.com/apache/arrow/blob/e83c23bb86156ef32d073c8eb507d601847350e1/cpp/src/arrow/filesystem/localfs_test.cc#L62-L69

Not sure why yet but if I remove that prefix, the entire test passes now. More later.

amoeba:
I'm making good progress finishing off the remainder of this over on [my fork](https://github.com/amoeba/arrow/tree/ARROW-33618/FileSelector) and should have this ready for review within one or two sessions of work on it.

anjakefala:
@amoeba Are you planning to open a fresh PR with your fork? If you are, can you link me to it, and I'll close this one!

amoeba:
I just created a draft PR at https://github.com/apache/arrow/pull/44176.

PR #34177, opened 2023-02-14
GH-15145: [C++][Parquet] Specify Dict Encoding Type

### LINKED ISSUE #15145: [C++][Parquet] The DictEncoder is always PLAIN_DICTIONARY even in parquet_v2 format

In `DictEncoderImpl`, the encoding of it is fixed, which is `PLAIN_DICTIONARY`. In our standard, it should be `RLE_DICTIONARY` or `PLAIN_DICTIONARY`, and should be decide by parquet version.

Though the final format maybe right, the temporary encoding might be trickey here.

I'd like to fix this if you'd like it

### Component(s)

C++, Parquet

mapleFU:
### Rationale for this change

In parquet, for v1 writer, it will use `PLAIN_DICTIONARY` for dict index page. In parquet v2, `RLE_DICTIONARY` and `PLAIN` is used. However, the encoder will always be `PLAIN_DICTIONARY`, which is really confusing.

This patch is a draft. In `parquet::WriterProperties`, we're able to get `dictionary_index_encoding()`, and `ColumnWriter` would use it. However, testing usally uses `MakeEncoder(Encoding::PLAIN, /* use_dict */ true)`. So I wonder if it's ok.

Change Dict Encoding using specified encoding.

### Are these changes tested?

Currently not.

Users using parquet internal `MakeEncoder` would failed.

* Closes: #15145

### DIFF
--- a/cpp/src/parquet/column_writer.cc
+++ b/cpp/src/parquet/column_writer.cc
@@ -1043,7 +1043,7 @@ Status ConvertDictionaryToDense(const ::arrow::Array& array, MemoryPool* pool,
 }
 
 static inline bool IsDictionaryEncoding(Encoding::type encoding) {
-  return encoding == Encoding::PLAIN_DICTIONARY;
+  return encoding == Encoding::PLAIN_DICTIONARY || encoding == Encoding::RLE_DICTIONARY;
 }
 
 template <typename DType>
--- a/cpp/src/parquet/encoding.cc
+++ b/cpp/src/parquet/encoding.cc
@@ -480,11 +480,18 @@ class DictEncoderImpl : public EncoderImpl, virtual public DictEncoder<DType> {
  public:
   typedef typename DType::c_type T;
 
-  explicit DictEncoderImpl(const ColumnDescriptor* desc, MemoryPool* pool)
-      : EncoderImpl(desc, Encoding::PLAIN_DICTIONARY, pool),
+  explicit DictEncoderImpl(const ColumnDescriptor* desc, Encoding::type dict_encoding,
+                           MemoryPool* pool)
+      : EncoderImpl(desc, dict_encoding, pool),
         buffered_indices_(::arrow::stl::allocator<int32_t>(pool)),
         dict_encoded_size_(0),
-        memo_table_(pool, kInitialHashTableSize) {}
+        memo_table_(pool, kInitialHashTableSize) {
+    if (dict_encoding != Encoding::PLAIN_DICTIONARY &&
+        dict_encoding != Encoding::RLE_DICTIONARY) {
+      throw ParquetException("Pass Non-Dictionary encoding " +
+                             EncodingToString(dict_encoding) + " to DictEncoder");
+    }
+  }
 
   ~DictEncoderImpl() override { DCHECK(buffered_indices_.empty()); }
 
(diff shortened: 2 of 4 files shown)


mapleFU:
@pitrou @wjones127 @wgtmac Mind take a look?

cpp/src/parquet/column_writer.cc
```diff
@@ -1043,7 +1043,7 @@ Status ConvertDictionaryToDense(const ::arrow::Array& array, MemoryPool* pool,
 }
 
 static inline bool IsDictionaryEncoding(Encoding::type encoding) {
-  return encoding == Encoding::PLAIN_DICTIONARY;
+  return encoding == Encoding::PLAIN_DICTIONARY || encoding == Encoding::RLE_DICTIONARY;
```
wgtmac, line 1046: Can we remove this function and instead adding a function to return if the ColumnWriter uses dictionary encoding?

cpp/src/parquet/encoding.cc
```diff
@@ -480,11 +480,18 @@ class DictEncoderImpl : public EncoderImpl, virtual public DictEncoder<DType> {
  public:
   typedef typename DType::c_type T;
 
-  explicit DictEncoderImpl(const ColumnDescriptor* desc, MemoryPool* pool)
-      : EncoderImpl(desc, Encoding::PLAIN_DICTIONARY, pool),
+  explicit DictEncoderImpl(const ColumnDescriptor* desc, Encoding::type dict_encoding,
```
wgtmac, line 483: To provide some detail, the encoding type of dictionary encoded pages is depending on the format version
- Format v1: `PLAIN_DICTIONARY` is used for both dict page and data page. 
- Format v2: dict page uses `PLAIN` and data page uses `RLE_DICTIONARY`

So it is kind of weird to manually pass `dict_encoding` because they are the same thing but relevant to different parquet formats.

mapleFU: Right, currently they will be decided by `dictionary_index_page()` in `WriterProperties`

And currently, there encoding logic are similiar

wgtmac:
I'd propose to fix it by using dict encoding type of format v2 and only deal with the deprecated `PLAIN_DICTIONARY` when talking to thrift messages.

For example, `WriterProperties` has dealt with the format already: [link](https://github.com/apache/arrow/blob/master/cpp/src/parquet/properties.h#L577) 
```cpp
  inline Encoding::type dictionary_index_encoding() const {
    if (parquet_version_ == ParquetVersion::PARQUET_1_0) {
      return Encoding::PLAIN_DICTIONARY;
    } else {
      return Encoding::RLE_DICTIONARY;
    }
  }

  inline Encoding::type dictionary_page_encoding() const {
    if (parquet_version_ == ParquetVersion::PARQUET_1_0) {
      return Encoding::PLAIN_DICTIONARY;
    } else {
      return Encoding::PLAIN;
    }
  }
```

The DictDecoder always uses `RLE_DICTIONARY` to create decoder and check the encoding type of data pages: [link](https://github.com/apache/arrow/blob/master/cpp/src/parquet/encoding.cc#L1470)
```cpp
  explicit DictDecoderImpl(const ColumnDescriptor* descr,
                           MemoryPool* pool = ::arrow::default_memory_pool())
      : DecoderImpl(descr, Encoding::RLE_DICTIONARY),
        dictionary_(AllocateBuffer(pool, 0)),
        dictionary_length_(0),
        byte_array_data_(AllocateBuffer(pool, 0)),
        byte_array_offsets_(AllocateBuffer(pool, 0)),
        indices_scratch_space_(AllocateBuffer(pool, 0)) {}
```

And the ColumnReader always resets the dict index encoding to `RLE_DICTIONARY`: [link](https://github.com/apache/arrow/blob/master/cpp/src/parquet/column_reader.cc#L853)
```cpp
    Encoding::type encoding = page.encoding();

    if (IsDictionaryIndexEncoding(encoding)) {
      encoding = Encoding::RLE_DICTIONARY;
    }
```

So I'd prefer doing the same thing on the writer side. @mapleFU

mapleFU:
@wgtmac personally I wonder if there are user using format v1. If they're using format v1, would it be confusing if v1 user meets RLE_DICTIONARY?

And I wonder what's other contributors' opinion...

wgtmac:
> @wgtmac personally I wonder if there are user using format v1. If they're using format v1, would it be confusing if v1 user meets RLE_DICTIONARY?
> 
> And I wonder what's other contributors' opinion...

What I have proposed is to use `RLE_DICTIONARY` internally in the implementations of `ColumnWriter` and `DictEncoder`. This makes less confusion for developers. The users are still free to put `PLAIN_DICTIONARY` or `RLE_DICTIONARY` to the `WriterProperties`.

PR #34178, opened 2023-02-14
GH-34023: [Docs] Version warning about viewing old docs doesn't work for versions >= 10

### LINKED ISSUE #34023: [Docs] Version warning about viewing old docs doesn't work for versions >= 10
Currently, if viewing older docs (eg https://arrow.apache.org/docs/9.0/python/index.html), you get a warning about that you are not viewing the latest docs. However, now that the newest "old" docs (those that are stored in a version/ subdirectory) is 10.0, that doesn't seem to work for this version (eg https://arrow.apache.org/docs/10.0/python/index.html doesn't show the warning)

My first guess is that it's caused by a bump in number of characters in the major version number (7, 8, 9 etc -> 10), and that this if check no longer works as expected because of that:

https://github.com/apache/arrow/blob/177a0d4400e61f44dbd045ce05150bba28dd1d74/docs/source/_static/versionwarning.js#L104-L105

AlenkaF:
### Rationale for this change
Due to a bump in number of characters in the major version number (7, 8, 9 etc -> 10) the check for versions in `versionwarning.js` no longer works as expected.

Update of the checks in `versionwarning.js` file.

### Are these changes tested?
Yes, on the apache/arrow-site PR:

- https://github.com/apache/arrow-site/pull/312
- https://arrow.apache.org/docs/10.0/index.html

No.
* Closes: #34023

### DIFF
--- a/docs/source/_static/versionwarning.js
+++ b/docs/source/_static/versionwarning.js
@@ -30,7 +30,7 @@
             // Path of the page
             var location_array = location.pathname.split('/');
             var versionPath = location_array[2];
-            var versionNumber = Number(versionPath.match(/^\d+/))
+            var majorVersionNumber = Number(versionPath.match(/^\d+/))
             var subPath = location_array[3];
             var filePath = location_array.slice(3).join('/');
             // Links to stable or dev versions
@@ -55,7 +55,7 @@
                         $('.container-fluid').prepend(`${showWarning}`)
                     }
                 });
-            } else if (versionNumber < 4) {
+            } else if (majorVersionNumber < 4) {
                 // old versions 1.0, 2.0 or 3.0
                 $.ajax({
                     type: 'HEAD',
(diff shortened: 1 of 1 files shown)


docs/source/_static/versionwarning.js
```diff
@@ -30,6 +30,7 @@
             // Path of the page
             var location_array = location.pathname.split('/');
             var versionPath = location_array[2];
+            var versionNumber = Number(versionPath.match(/^\d+/))
```
raulcd, line 33: This is `majorVersionNumber` from the `major.minor.patch` format, right? A really minor nit I would probably rename it so it does not cause confusion.

raulcd (approved):
Thanks @AlenkaF ! LGTM just a really minor nit

docs/source/_static/versionwarning.js
AlenkaF: Yes, correct. It is the major part of the version format. Will rename 👍

PR #34184, opened 2023-02-14
GH-34154: [Python] Add `is_nan` method to Array and Expression

### LINKED ISSUE #34154: [Python] Add `is_nan` for Python

We currently use `field.is_null(nan_is_null=True) & field.is_valid()` for filtering on NaN values, but would be great to fall back to `is_nan`. The abovementioned expression doesn't seem to catch NaN values.

### Component(s)

Python

Fokko:
* Closes: GH-34154


Fokko:
Is this expected? 🔍 

```
>       assert result.equals(expected)
E       assert False
E        +  where False = <built-in method equals of pyarrow.lib.BooleanArray object at 0x7f41dfe37c90>(<pyarrow.lib.BooleanArray object at 0x7f41dfe37a60>\n[\n  false,\n  false,\n  false,\n  false,\n  true\n])
E        +    where <built-in method equals of pyarrow.lib.BooleanArray object at 0x7f41dfe37c90> = <pyarrow.lib.BooleanArray object at 0x7f41dfe37c90>\n[\n  false,\n  false,\n  false,\n  null,\n  true\n].equals
```

AlenkaF:
Looking at the tests in C++, I think it is expected:

https://github.com/apache/arrow/blob/53752adc6b81166cd4ee7db5a819494042f29197/cpp/src/arrow/compute/kernels/scalar_validity_test.cc#L121-L132
https://github.com/apache/arrow/blob/53752adc6b81166cd4ee7db5a819494042f29197/cpp/src/arrow/compute/kernels/scalar_validity_test.cc#L229-L244

python/pyarrow/tests/test_compute.py
```diff
@@ -1634,6 +1634,10 @@ def test_is_null():
     expected = pa.array([False, False, False, True, True])
     assert result.equals(expected)
 
+    result = arr.is_nan()
+    expected = pa.array([False, False, False, False, True])
```
jorisvandenbossche, line 1638:
```suggestion
    expected = pa.array([False, False, False, None, True])
```

jorisvandenbossche:
> Is this expected?

It's not super clear from the assert message, but so it is the "null" that is different (see my inline comment). As Alenka mentioned above, that is indeed expected (and tested on the C++ side) behaviour. 

In general, predicate functions (checking something, returning a boolean) propagate nulls (in the idea of "you don't know the value, so you also don't know if it is NaN or not, so whether it should be True or False). 
(one exception on that rule is of course `is_null` to check for nulls itself)

Fokko:
I learned something today. Thanks @AlenkaF for pointing that out.

Updated the PR with also adding the option of returning `None`'s in the docstring 👍🏻

python/pyarrow/tests/test_compute.py
Fokko: Thanks for the suggestion, fixed this in a commit together with updating the docstring 👍🏻

AlenkaF:
Great, thanks for adding this!
One thing I see missing is taking into account that `is_nan` expression only handles numeric types. What happens if we call `is_nan` on an array of strings or any non-numeric type?

python/pyarrow/array.pxi
```diff
@@ -1253,6 +1253,16 @@ cdef class Array(_PandasConvertible):
         options = _pc().NullOptions(nan_is_null=nan_is_null)
         return _pc().call_function('is_null', [self], options)
 
+    def is_nan(self):
+        """
+        Return BooleanArray indicating the nan values.
+
+        Returns
+        -------
+        array : boolean Array, where null values are None
```
jorisvandenbossche, line 1262: What does this "where null values are None" mean exactly?

python/pyarrow/_compute.pyx
```diff
@@ -2436,6 +2436,19 @@ cdef class Expression(_Weakrefable):
         options = NullOptions(nan_is_null=nan_is_null)
         return Expression._call("is_null", [self], options)
 
+    def is_nan(self):
+        """
+        Check whether the expression is nan.
```
jorisvandenbossche, line 2441:
```suggestion
        Check whether the expression is NaN.
```

python/pyarrow/array.pxi
```diff
@@ -1253,6 +1253,16 @@ cdef class Array(_PandasConvertible):
         options = _pc().NullOptions(nan_is_null=nan_is_null)
         return _pc().call_function('is_null', [self], options)
 
+    def is_nan(self):
+        """
+        Return BooleanArray indicating the nan values.
```
jorisvandenbossche, line 1258:
```suggestion
        Return BooleanArray indicating the NaN values.
```

python/pyarrow/table.pxi
```diff
@@ -327,6 +327,32 @@ cdef class ChunkedArray(_PandasConvertible):
         options = _pc().NullOptions(nan_is_null=nan_is_null)
         return _pc().call_function('is_null', [self], options)
 
+    def is_nan(self):
+        """
+        Return boolean array indicating the nan values.
```
jorisvandenbossche, line 332:
```suggestion
        Return boolean array indicating the NaN values.
```

python/pyarrow/tests/test_compute.py
```diff
@@ -1635,6 +1636,23 @@ def test_is_null():
...
 
+def test_is_nan():
+    arr = pa.array([1, 2, 3, None, np.nan])
+    result = arr.is_nan()
+    expected = pa.array([False, False, False, None, True])
+    assert result.equals(expected)
+
+    with pytest.raises(
+            ArrowNotImplementedError, match="has no kernel matching input types"):
+        arr = pa.array(["1", "2", None], type=pa.string())
```
jorisvandenbossche, line 1647:
```suggestion
    arr = pa.array(["1", "2", None], type=pa.string())
    with pytest.raises(
            ArrowNotImplementedError, match="has no kernel matching input types"):
```

And the same for the one below (although unlikely here, but that avoids that it might be accidentally the array creation that is erroring, instead of calling is_nan)

jorisvandenbossche (commented):
Code looks good, added some comments on the docstrings.

You also still have a related doctest failure:

```
__________________ [doctest] pyarrow.lib.ChunkedArray.is_nan ___________________
331 ChunkedArray.is_nan(self)
332 
333         Return boolean array indicating the nan values.
334 
335         Examples
336         --------
337         >>> import pyarrow as pa
338         >>> import numpy as np
339         >>> arr = pa.chunked_array([[2, np.nan, 4], [4, None, 100]])
340         >>> arr.is_nan()
Differences (unified diff with -expected +actual):
    @@ -1,12 +1,10 @@
    -<pyarrow.lib.ChunkedArray object at ...>
    +<pyarrow.lib.ChunkedArray object at 0x7f91d0ff3360>
     [
       [
         false,
         true,
    -    false
    -  ],
    -  [
         false,
    -    None,
    +    false,
    +    null,
         false
       ]
```

FYI you can check those locally as well: https://arrow.apache.org/docs/dev/developers/python.html#doctest

python/pyarrow/array.pxi
Fokko: That a null value, will result in a Python None:
https://github.com/apache/arrow/blob/df625f2abf7f460b9935503998c075ad9f3c5e62/python/pyarrow/tests/test_compute.py#L1639-L1643

jorisvandenbossche: Ah, OK, I didn't get that at first. Now, I don't think that is something we need to mention here in this docstring (that's true for _all_ functions that return arrow data). And it's also only when converting to native Python objects that this happens, the actual result here is a pyarrow.array, and that doesn't use None (so the explanation would be relevant for a method like `Array.to_pylist()`.

The reason you see it in the tests, is because we allow you to use None to specify a null value when constructing an array (but the None gets converted to a proper "null").

Fokko: @jorisvandenbossche Okay, that's fine with me as well. It was not directly obvious for me, but it makes sense now I'm aware of it.

jorisvandenbossche (approved):
Thanks!

PR #34192, opened 2023-02-15
GH-34191: [C++] Ensure using the same ProtoBuf in bundled ORC

### LINKED ISSUE #34191: [C++] Bundled ORC may not use the same ProtoBuf

We may specify wrong `-DPROTO*` values when we use bundled ORC. Because `FindProtobuf.cmake` provided by CMake and `protobuf-config.cmake` provided by ProtoBuf is incompatible. If we specify wrong `-DPROTO*` values, ORC uses bundled ProtoBuf. It may conflict ProtoBuf used by Arrow.

https://github.com/Homebrew/homebrew-core/pull/121566 hits this problem.

### Component(s)

C++

kou:
### Rationale for this change

We can't detect correct ProtoBuf related values to pass bundled ORC with the current approach.

This change uses generator expression to get `CONFIG` depended value correctly.

### Are these changes tested?

Yes.

No.
* Closes: #34191


kou:
@github-actions crossbow submit -g cpp -g r -g wheel

kou:

kou:

kou:
+1

PR #13330, opened 2022-06-07
ARROW-16772: [C++] Implement encode and decode functions for Run-Length encoding

zagto:

PR #34195, opened 2023-02-15
GH-32105: [C++] Encode and decode Run-End Encoded vectors

### LINKED ISSUE #20351: [C++] Kernel input type matcher for RLE
Builds on top of ARROW-17261

**Reporter**: [Tobias Zagorni](https://issues.apache.org/jira/browse/ARROW-17262) / @zagto
#### PRs and other links:
- [GitHub Pull Request #13971](https://github.com/apache/arrow/pull/13971)

<sub>**Note**: *This issue was originally created as [ARROW-17262](https://issues.apache.org/jira/browse/ARROW-17262). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

### LINKED ISSUE #32105: [C++] Implement encode and decode functions for Run-Length encoding
**Reporter**: [Tobias Zagorni](https://issues.apache.org/jira/browse/ARROW-16772) / @zagto
#### PRs and other links:
- [GitHub Pull Request #13330](https://github.com/apache/arrow/pull/13330)

<sub>**Note**: *This issue was originally created as [ARROW-16772](https://issues.apache.org/jira/browse/ARROW-16772). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

felipecrv:
* Closes #13330

* Closes: #32105

### DIFF
--- a/cpp/src/arrow/compute/api_vector.h
+++ b/cpp/src/arrow/compute/api_vector.h
@@ -80,6 +80,15 @@ class ARROW_EXPORT DictionaryEncodeOptions : public FunctionOptions {
   NullEncodingBehavior null_encoding_behavior = MASK;
 };
 
+/// \brief Options for the run-end encode function
+class ARROW_EXPORT RunEndEncodeOptions : public FunctionOptions {
+ public:
+  explicit RunEndEncodeOptions(std::shared_ptr<DataType> run_end_type = int32());
+  static constexpr char const kTypeName[] = "RunEndEncodeOptions";
+
+  std::shared_ptr<DataType> run_end_type;
+};
+
 class ARROW_EXPORT ArraySortOptions : public FunctionOptions {
  public:
   explicit ArraySortOptions(SortOrder order = SortOrder::Ascending,
@@ -546,6 +555,29 @@ Result<Datum> DictionaryEncode(
     const Datum& data,
     const DictionaryEncodeOptions& options = DictionaryEncodeOptions::Defaults(),
     ExecContext* ctx = NULLPTR);
+
+/// \brief Run-end-encode values in an array-like object
+///
+/// \param[in] value array-like input
+/// \param[in] ctx the function execution context, optional
+/// \return result with same shape and type as input
+///
+/// \since 12.0.0
+/// \note API not yet finalized
+ARROW_EXPORT
+Result<Datum> RunEndEncode(const Datum& value, const RunEndEncodeOptions& options,
+                           ExecContext* ctx = NULLPTR);
+
+/// \brief Decode a Run-End Encoded array to a plain array
+///
+/// \param[in] value run-end-encoded input
+/// \param[in] ctx the function execution context, optional
+/// \return result with same shape and type as input
+///
+/// \since 12.0.0
+/// \note API not yet finalized
+ARROW_EXPORT
+Result<Datum> RunEndDecode(const Datum& value, ExecContext* ctx = NULLPTR);
 
 ARROW_EXPORT
 Result<Datum> CumulativeSum(
--- a/cpp/src/arrow/compute/api_vector.h
+++ b/cpp/src/arrow/compute/api_vector.h
@@ -558,10 +558,13 @@ Result<Datum> DictionaryEncode(
 
 /// \brief Run-end-encode values in an array-like object
 ///
+/// The returned run-end encoded type uses the same value type of the input and
+/// run-end type defined in the options.
+///
 /// \param[in] value array-like input
 /// \param[in] options configures encoding behavior
 /// \param[in] ctx the function execution context, optional
-/// \return result with same shape and type as input
+/// \return result with same shape but run-end encoded
 ///
 /// \since 12.0.0
 /// \note API not yet finalized
@@ -571,9 +574,12 @@ Result<Datum> RunEndEncode(const Datum& value, const RunEndEncodeOptions& option
 
 /// \brief Decode a Run-End Encoded array to a plain array
 ///
+/// The output data type is the same as the values array type of run-end encoded
+/// input.
+///
 /// \param[in] value run-end-encoded input
 /// \param[in] ctx the function execution context, optional
-/// \return result with same shape and type as input
+/// \return plain array resulting from decoding the run-end encoded input
 ///
 /// \since 12.0.0
 /// \note API not yet finalized
--- a/cpp/src/arrow/compute/api_vector.h
+++ b/cpp/src/arrow/compute/api_vector.h
@@ -559,6 +559,7 @@ Result<Datum> DictionaryEncode(
 /// \brief Run-end-encode values in an array-like object
 ///
 /// \param[in] value array-like input
+/// \param[in] options configures encoding behavior
 /// \param[in] ctx the function execution context, optional
 /// \return result with same shape and type as input
 ///
--- a/cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
+++ b/cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
@@ -620,6 +620,7 @@ void RegisterVectorRunEndEncode(FunctionRegistry* registry) {
   }
   add_kernel(boolean());
   add_kernel(null());
+  // TODO(GH-34195): Add support for more types
 
   DCHECK_OK(registry->AddFunction(std::move(function)));
 }
@@ -643,6 +644,7 @@ void RegisterVectorRunEndDecode(FunctionRegistry* registry) {
   }
   add_kernel(boolean());
   add_kernel(null());
+  // TODO(GH-34195): Add support for more types
 
   DCHECK_OK(registry->AddFunction(std::move(function)));
 }
--- a/cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
+++ b/cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
@@ -227,6 +227,38 @@ Status ValidateRunEndType(int64_t input_length) {
   return Status::OK();
 }
 
+/// \brief Preallocate the ArrayData for the run-end encoded version
+/// of the input array
+template <typename RunEndType, bool has_validity_buffer>
+Result<std::shared_ptr<ArrayData>> PreallocateREEData(const ArraySpan& input_array,
+                                                      int64_t physical_length,
+                                                      int64_t physical_null_count,
+                                                      MemoryPool* pool) {
+  ARROW_ASSIGN_OR_RAISE(
+      auto run_ends_buffer,
+      AllocateBuffer(physical_length * RunEndType().byte_width(), pool));
+  std::shared_ptr<Buffer> validity_buffer = NULLPTR;
+  if constexpr (has_validity_buffer) {
+    ARROW_ASSIGN_OR_RAISE(validity_buffer, AllocateBitmap(physical_length, pool));
+  }
+  ARROW_ASSIGN_OR_RAISE(
+      auto values_buffer,
+      AllocatePrimitiveBuffer(physical_length, *input_array.type, pool));
+
+  auto ree_type = std::make_shared<RunEndEncodedType>(std::make_shared<RunEndType>(),
+                                                      input_array.type->GetSharedPtr());
+  auto run_ends_data =
+      ArrayData::Make(ree_type->run_end_type(), physical_length,
+                      {NULLPTR, std::move(run_ends_buffer)}, /*null_count=*/0);
+  auto values_data = ArrayData::Make(
+      ree_type->value_type(), physical_length,
+      {std::move(validity_buffer), std::move(values_buffer)}, physical_null_count);
+
+  return ArrayData::Make(std::move(ree_type), input_array.length, {NULLPTR},
+                         {std::move(run_ends_data), std::move(values_data)},
+                         /*null_count=*/0);
+}
+
 template <typename RunEndType, typename ValueType, bool has_validity_buffer>
 class RunEndEncodeImpl {
  private:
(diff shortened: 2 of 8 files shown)


felipecrv:
@westonpace @zeroshade @lidavidm this is now ready for review.

felipecrv:
I just noticed the input type matcher isn't really necessary for these kernels, so if you want I can remove this commit from the PR and push again.

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
```diff
@@ -0,0 +1,652 @@
...
+}
+
+template <>
+bool ReadValueImpl<BooleanType, false>::ReadValue(const uint8_t* input_validity,
+                                                  const void* input_values, CType* out,
+                                                  int64_t read_offset) const {
+  *out = bit_util::GetBit(reinterpret_cast<const uint8_t*>(input_values), read_offset);
+  return true;
+}
+
```
zeroshade, line 63: Shouldn't we need another explicit implementation here for handling variable and fixed length binary arrays? I don't see anything elsewhere for handling encoding/decoding of that type.

felipecrv: The function definition currently declares kernels for `numeric`, `bool`, and `null` types. The `Type::NA` implementation was a TODO in the initial code by Tobias, and I added a specialized implementation for them. I think I will need something similar that doesn't use this for fixed-length binary arrays. Can I create a separate issue for those and work on getting this PR merged with these initial types. I'm trying to get to a steady flow of small PRs.

zeroshade: that's perfectly fine, was just curious as I didn't see a TODO or other comment when I was skimming through the PR and wanted to make sure that I understood the code correctly that there wasn't an implementation for binary types

felipecrv: Oh, right. Added a TODO mentioned the newly created issue.

cpp/src/arrow/compute/kernels/vector_run_end_encode_test.cc
```diff
@@ -0,0 +1,211 @@
...
+    result.expected_run_ends_json = "[1, 2, 3]";
+    result.string = "Type min, max, & null values";
+    return result;
+  }
+
+  std::shared_ptr<Array> input;
+  std::shared_ptr<Array> expected_values;
+  std::string expected_run_ends_json;
+  // only used for gtest output
+  std::string string;
```
lidavidm, line 77: nit: `string` isn't very descriptive, maybe `description` if an instance of this is meant to be a test case?

cpp/src/arrow/compute/kernels/vector_run_end_encode_test.cc
```diff
@@ -0,0 +1,211 @@
...
+
+}  // namespace
+
+class TestRunEndEncodeDecode : public ::testing::TestWithParam<
+                                   std::tuple<REETestData, std::shared_ptr<DataType>>> {
+ public:
+  void AddArtificialOffsetInChildArray(ArrayData* array, int64_t offset) {
+    auto& child = array->child_data[1];
+    auto builder = MakeBuilder(child->type).ValueOrDie();
+    ARROW_CHECK_OK(builder->AppendNulls(offset));
```
lidavidm, line 88: You can use ASSERT_OK and ASSERT_OK_AND_ASSIGN

Then you can call this as `ASSERT_NO_FATAL_FAILURE(AddArtificialOffsetInChildArray(...))`

cpp/src/arrow/compute/kernels/vector_run_end_encode_test.cc
```diff
@@ -0,0 +1,211 @@
...
+
+  auto encoded = encoded_datum.array();
+  this->AddArtificialOffsetInChildArray(encoded.get(), 100);
+  ASSERT_OK_AND_ASSIGN(Datum datum_without_first, RunEndDecode(encoded));
+  auto array_without_first = datum_without_first.make_array();
+  ASSERT_OK(array_without_first->ValidateFull());
+  ASSERT_ARRAYS_EQUAL(*array_without_first, *data.input);
+}
+
+INSTANTIATE_TEST_SUITE_P(
```
lidavidm, line 167: Is it possible to test scalars and chunked arrays here, too?

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
```diff
@@ -0,0 +1,654 @@
...
+template <typename ArrowType, bool has_validity_buffer>
+struct WriteValueImpl {
```
lidavidm, line 65: Is there any possibility to consolidate with https://github.com/apache/arrow/blob/f74d5285803c23e6f553d655ea5f184d06d26607/cpp/src/arrow/compute/kernels/copy_data_internal.h

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
```diff
@@ -0,0 +1,654 @@
...
+#include "arrow/compute/kernels/common_internal.h"
+#include "arrow/util/checked_cast.h"
+#include "arrow/util/ree_util.h"
+
+namespace arrow {
+namespace compute {
+namespace internal {
+
+template <typename ArrowType, bool has_validity_buffer>
+struct ReadValueImpl {
```
lidavidm, line 30: Is it possible to use one of the existing typed visitors instead of a handrolled loop?

lidavidm (commented):
I need to take a closer look at the kernel implementations themselves, but here are some initial comments

cpp/src/arrow/compute/api_vector.h
```diff
@@ -577,6 +586,30 @@ Result<Datum> DictionaryEncode(
     const DictionaryEncodeOptions& options = DictionaryEncodeOptions::Defaults(),
     ExecContext* ctx = NULLPTR);
 
+/// \brief Run-end-encode values in an array-like object
+///
+/// \param[in] value array-like input
+/// \param[in] options configures encoding behavior
+/// \param[in] ctx the function execution context, optional
+/// \return result with same shape and type as input
```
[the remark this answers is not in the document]
felipecrv: Fixed. Pushing soon.

cpp/src/arrow/compute/kernels/vector_run_end_encode_test.cc
felipecrv: Renamed to `description`.

cpp/src/arrow/compute/kernels/vector_run_end_encode_test.cc
felipecrv: Thank you for the tip. Done.

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
felipecrv: Do you mean adding function to `copy_data_internal.h` classes that can copy + repeat values into array buffers?

cpp/src/arrow/compute/kernels/vector_run_end_encode_test.cc
felipecrv: The implementation doesn't handle `RunEndEncodedScalar`, so I will have to add that first ;-) Thanks for reminding me of that.

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
felipecrv: I think I can use `visit_data_inline.h` for some steps but not everything and the readability of the loops might become worse. I will try to do it.

cpp/src/arrow/compute/kernels/vector_run_end_encode_test.cc
lidavidm: Ah, whoops!

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
lidavidm: Possibly - if nothing there seems useful then don't bother

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
lidavidm: Just a suggestion - if it wouldn't improve it (and it doesn't seem like it could be a variant of one of the visitors) then don't worry about it

cpp/src/arrow/compute/kernels/vector_run_end_encode_test.cc
lidavidm: It seems things didn't change (looking at GitHub)?

felipecrv: Pushed now.

cpp/src/arrow/compute/kernels/vector_run_end_encode_test.cc
felipecrv: This is also pushed now.

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
felipecrv: Can you take a look at the last commit? The implementation is longer and relies on a lot of inlining to be efficient.

felipecrv: I decided against this solution and added a commit that reverts it.

cpp/src/arrow/compute/kernels/vector_run_end_encode_test.cc
felipecrv: Well, vector functions don't really deal with scalars, so scalars are not needed. cc @westonpace

I'm looking into chunked arrays now.

felipecrv: Done. Lots of chunked array tests!

felipecrv:
@westonpace @zeroshade this is now ready for final review. I've addressed all of @lidavidm's comments.

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
```diff
@@ -0,0 +1,672 @@
...
+Status RunEndEncodeNullArray(KernelContext* ctx, const ArraySpan& input_array,
+                             ExecResult* output) {
+  using RunEndCType = typename RunEndType::c_type;
+
+  const int64_t input_length = input_array.length;
+  auto input_array_type = input_array.type->GetSharedPtr();
+  DCHECK(input_array_type->id() == Type::NA);
+
+  int64_t num_output_runs = 0;
+  if (input_length > 0) {
```
zeroshade, line 365: same as above

zeroshade (changes requested):
Just had a few nitpicks really, overall looks really good to me.

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
```diff
@@ -0,0 +1,672 @@
...
+namespace compute {
+namespace internal {
+
+template <typename ArrowType, bool has_validity_buffer, typename Enable = void>
+struct ReadValueImpl {};
+
+// Numeric and primitive C-compatible types
+template <typename ArrowType, bool has_validity_buffer>
+struct ReadValueImpl<ArrowType, has_validity_buffer, enable_if_has_c_type<ArrowType>> {
+  using CType = typename ArrowType::c_type;
```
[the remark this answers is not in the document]
felipecrv: Removing the `BooleanType` specializations will be nice indeed, but the enable_if was done in preparation to support non-primitive types. I'm following the same pattern of `visit_data_inline.h` [1].

[1] https://github.com/apache/arrow/blob/main/cpp/src/arrow/visit_data_inline.h#L37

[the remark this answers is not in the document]
lidavidm: I'll note that that header predates the C++17 migration, so if you think `if constexpr` can clean up the code over template metaprogramming, I'd lean towards that

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
```diff
@@ -0,0 +1,672 @@
...
+
+  [[nodiscard]] bool ReadValue(const uint8_t* input_validity, const void* input_values,
+                               CType* out, int64_t read_offset) const {
+    bool valid = true;
+    if constexpr (has_validity_buffer) {
+      valid = bit_util::GetBit(input_validity, read_offset);
+    }
+    if (valid) {
+      *out = (reinterpret_cast<const CType*>(input_values))[read_offset];
+    }
```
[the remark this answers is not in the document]
felipecrv: Indeed, and having this `cmov` waiting for the `valid` to be loaded is bad for speculative execution. I will fix.

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
```diff
@@ -0,0 +1,672 @@
...
+template <typename RunEndType>
+Status ValidateRunEndType(int64_t input_length) {
+  using RunEndCType = typename RunEndType::c_type;
+  constexpr int64_t kRunEndMax = std::numeric_limits<RunEndCType>::max();
+  if (input_length > kRunEndMax) {
+    return Status::Invalid(
+        "Cannot run-end encode Arrays with more elements than the "
+        "run end type can hold: ",
+        kRunEndMax);
+  }
```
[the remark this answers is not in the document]
felipecrv: I will add the extra boolean check.

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
```diff
@@ -0,0 +1,672 @@
...
+  Status Exec() {
+    const int64_t input_length = input_array_.length;
+    const int64_t input_offset = input_array_.offset;
+    const auto* input_validity = input_array_.buffers[0].data;
+    const auto* input_values = input_array_.buffers[1].data;
+
+    // First pass: count the number of runs
+    int64_t num_valid_runs = 0;
+    int64_t num_output_runs = 0;
+    if (input_length > 0) {
```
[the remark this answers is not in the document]
felipecrv: I tried that at first, the code got very repetitive and more bug-prone. I would have to repeat the next block with a hard-coded 0.

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
```diff
@@ -0,0 +1,672 @@
...
+          input_array_.length, input_array_.offset, input_validity, input_values);
+      std::tie(num_valid_runs, num_output_runs) = counting_loop.CountNumberOfRuns();
+    }
+
+    // Allocate the output array data
+    std::shared_ptr<ArrayData> output_array_data;
+    int64_t validity_buffer_size = 0;  // in bytes
+    {
+      ARROW_ASSIGN_OR_RAISE(auto run_ends_buffer,
+                            AllocateBuffer(num_output_runs * RunEndType().bit_width(),
```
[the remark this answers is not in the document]
felipecrv: Oh! yes :grimacing:

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
```diff
@@ -0,0 +1,672 @@
...
+                                         {NULLPTR, std::move(run_ends_buffer)},
+                                         /*null_count=*/0);
+    auto values_data = ArrayData::Make(input_array_type, num_output_runs, {NULLPTR},
+                                       /*null_count=*/num_output_runs);
+
+    output_array_data =
+        ArrayData::Make(std::move(ree_type), input_length, {NULLPTR},
+                        {std::move(run_ends_data), std::move(values_data)},
+                        /*null_count=*/0);
+  }
```
felipecrv, line 390: @zeroshade early-returning for the 0 case means repeating this block. Don't you think it's more elegant that this code can handle any length and as such doesn't have to be duplicated?

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
```diff
@@ -0,0 +1,672 @@
...
+
+  auto add_kernel = [&function](const std::shared_ptr<DataType>& ty) {
+    for (const auto& run_end_type : {int16(), int32(), int64()}) {
+      auto exec = GenerateTypeAgnosticPrimitive<RunEndDecodeExec>(ty);
+      auto input_type = std::make_shared<RunEndEncodedType>(run_end_type, ty);
+      auto sig = KernelSignature::Make({InputType(input_type)}, OutputType({ty}));
+      VectorKernel kernel(sig, exec);
+      DCHECK_OK(function->AddKernel(std::move(kernel)));
+    }
+  };
```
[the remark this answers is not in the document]
felipecrv: Sure, I will add the comment. The decision was conscious. I've done something similar in the builder (appending a slice can leave two runs with the same value). The complexity that handling these cases introduce can kill performance in some cases and give many surprises to people using the library IMHO. Not worth it for the rather small memory savings.

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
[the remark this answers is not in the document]
felipecrv: Done.

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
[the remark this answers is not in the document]
felipecrv: I rewrote it to use the `if constexpr (is_same_type<...>)` and I will wait to know what you think about following the pattern in `visit_data_inline.h`.

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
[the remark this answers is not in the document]
felipecrv: Done.

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
```diff
@@ -0,0 +1,672 @@
...
+                                           ctx_->memory_pool()));
+      std::shared_ptr<Buffer> validity_buffer = NULLPTR;
+      if constexpr (has_validity_buffer) {
+        validity_buffer_size = bit_util::BytesForBits(num_output_runs);
+        ARROW_ASSIGN_OR_RAISE(validity_buffer,
+                              AllocateBuffer(validity_buffer_size, ctx_->memory_pool()));
+      }
+      ARROW_ASSIGN_OR_RAISE(auto values_buffer,
+                            AllocateBuffer(bit_util::BytesForBits(
+                                               num_output_runs * ValueType().bit_width()),
```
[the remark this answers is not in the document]
felipecrv: That's true in the case above, but `ValueType` can be `BooleanType` which fits in a single bit.

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
```diff
@@ -623,6 +579,13 @@ void RegisterVectorRunEndEncode(FunctionRegistry* registry) {
   auto function = std::make_shared<VectorFunction>("run_end_encode", Arity::Unary(),
                                                    run_end_encode_doc);
 
+  // NOTE: When the input to run_end_encode() is a ChunkedArray, the output is also a
+  // ChunkedArray with the same number of chunks as the input. Each chunk in the output
+  // has the same logical length as the corresponding chunk in the input. This simplicity
+  // has a small downside: if a run of identical values crosses a chunk boundary, this run
+  // cannot be encoded as a single run in the output. This is a conscious trade-off as
+  // trying to solve this corner-case would complicate the implementation,
+  // require reallocations, and could create surprising behavior for users of this API.
```
felipecrv, line 588: @zeroshade what you think? Should this be part of the function documentation?

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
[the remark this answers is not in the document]
felipecrv: (the builder takes a slice of another ree array, so just concatenating blindly is much simpler)

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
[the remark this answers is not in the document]
zeroshade: for a `BooleanType` you should use `AllocateBitmap` instead, right?

`BytesForBits(num_output_runs * bit_width())` should be identical to `byte_width()` even in the boolean case, shouldn't it?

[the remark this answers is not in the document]
felipecrv: `Boolean::bit_width()` returns 1:


```cpp
// BooleanType within arrow use a single bit instead of the C 8-bits layout.
int bit_width() const final { return 1; }
```

Then the `BooleanType::byte_width()` which is the default from `DataType`

```cpp
    /// \brief Returns the type's fixed byte width, if any. Returns -1
    /// for non-fixed-width types, and should only be used for
    /// subclasses of FixedWidthType
    virtual int32_t byte_width() const final {
      int32_t num_bits = this->bit_width();
      return num_bits > 0 ? num_bits / 8 : -1;
    }
  ```

evaluates to 0.

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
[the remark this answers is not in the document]
lidavidm: We can leave it as is if other reviewers are happy. Possibly file an issue to continue consolidating kernel utilities like this together where possible

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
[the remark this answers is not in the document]
felipecrv: I opened a PR adding `BytesForBits` in the filtering code after this research: https://github.com/apache/arrow/pull/34521

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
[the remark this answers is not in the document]
felipecrv: I just opened #34522 which removes the BooleanType specialization from `visit_data_inline.h`, but leaves that `enable_if` for selecting implementations across different categories of types.

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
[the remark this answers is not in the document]
zeroshade: oy, well that's inconvenient....

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
zeroshade: okay that's fair in this case.

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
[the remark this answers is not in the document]
zeroshade: would it make sense to put the output allocating into a function which you can just call with the 0 and do the early return for input_length == 0? 

I agree that the repeated `if (input_length > 0)` is preferable to duplicating the whole block. I'd be fine with either leaving it as is or shifting the allocation of the output to a function which is called during the early return. Up to you.

zeroshade (approved):

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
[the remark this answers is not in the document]
felipecrv: Good news. I just added a commit that removes the need for `BytesForBits` and does the right clearing of trailing bits when allocating bitmaps.

cpp/src/arrow/compute/kernels/vector_run_end_encode.cc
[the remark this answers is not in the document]
felipecrv: I extracted the allocation function and rewrote this to early-return on 0 length.

zeroshade (approved):
LGTM the code definitely looks a bit simpler with the extracted prealloc method + early return.... if i do say so myself :stuck_out_tongue:

PR #34208, opened 2023-02-15
GH-33143: [C++] Kernel to convert timestamp with timezone to wall time

### LINKED ISSUE #33143: [C++] Kernel to convert timestamp with timezone to wall time
We have `assume_timezone` to go from wall time timestamp to timestamp with a timezone. We might want a reverse operation to go from timestamp with a timezone to wall time.

This is not needed for computation within Arrow, but would be needed if an application outsides of Arrow consumes wall time. E.g.: https://stackoverflow.com/questions/73275465/how-to-keep-original-datatime-in-pyarrow-table/73276431

**Reporter**: [Rok Mihevc](https://issues.apache.org/jira/browse/ARROW-17935) / @rok
**Watchers**: [Rok Mihevc](https://issues.apache.org/jira/browse/ARROW-17935) / @rok
#### Related issues:
- [[C++] Kernel to localize naive timestamps to a timezone (preserving clock-time)](https://github.com/apache/arrow/issues/28744) (is related to)

<sub>**Note**: *This issue was originally created as [ARROW-17935](https://issues.apache.org/jira/browse/ARROW-17935). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

rok:
### Rationale for this change

We have `assume_timezone` to go from wall time timestamp to timestamp with a timezone. We might want a reverse operation to go from timestamp with a timezone to wall time.

This is not needed for computation within Arrow, but would be needed if an application outsides of Arrow consumes wall time. E.g.: https://stackoverflow.com/questions/73275465/how-to-keep-original-datatime-in-pyarrow-table/73276431

This adds a `local_time` C++ kernel.

### Are these changes tested?

Yes.

Yes, `local_time` C++ kernel is added.
* Closes: #33143


docs/source/cpp/compute.rst
```diff
@@ -1445,6 +1445,8 @@ For timestamps inputs with non-empty timezone, localized timestamp components wi
 +--------------------+------------+-------------------+---------------+----------------------------+-------+
 | is_leap_year       | Unary      | Timestamp, Date   | Boolean       |                            |       |
 +--------------------+------------+-------------------+---------------+----------------------------+-------+
+| local_time         | Unary      | Timestamp         | Timestamp     |                            |       |
```
westonpace, line 1448: Can we add a note?  I don't think it will be obvious what this function does.

cpp/src/arrow/compute/api_scalar.h
```diff
@@ -1518,6 +1518,16 @@ ARROW_EXPORT Result<Datum> AssumeTimezone(const Datum& values,
...
                                              ExecContext* ctx = NULLPTR);
 
+/// \brief LocalTime converts timestamp to timezone naive local timestamp
+///
+/// \param[in] values input to convert to local time
+/// \param[in] ctx the function execution context, optional
+/// \return the resulting datum
+///
+/// \since 12.0.0
+/// \note API not yet finalized
```
westonpace, line 1528: I was going to suggest removing this but now I see this is everywhere.  Hmm.... :(  I'm not sure how true this is these days.

westonpace (approved):
Thanks, as always, for your diligence here.

docs/source/cpp/compute.rst
rok: Good point. Added some text and moved under the `Timezone handling` chapter. I'll trigger a docs preview to make sure I've got the markup right.

cpp/src/arrow/compute/api_scalar.h
rok: Well, you could see it as an opportunity to sync the API with substrait :D

rok:

rok:
Docs preview: http://crossbow.voltrondata.com/pr_docs/34208/cpp/compute.html#timezone-handling

docs/source/cpp/compute.rst
```diff
@@ -1556,11 +1556,20 @@ Input timestamps are assumed to be relative to the timezone given in
 UTC-relative timestamps with the timezone metadata set to the above value.
 An error is returned if the timestamps already have the timezone metadata set.
 
+`local_time` function converts UTC-relative timestamps to local "timezone-naive"
+timestamp. The timezone is taken from the timezone metadata of the input
```
westonpace, line 1560:
```suggestion
timestamps. The timezone is taken from the timezone metadata of the input
```

docs/source/cpp/compute.rst
```diff
@@ -1556,11 +1556,20 @@ Input timestamps are assumed to be relative to the timezone given in
...
+timestamps. This function is the inverse of `assume_timezone`. Please note:
+**all temporal functions already operate on timestamps as if they were in local
+time of metadata provided timezone**. Using `local_time` is only meant to be
```
westonpace, line 1563:
```suggestion
time of the metadata provided timezone**. Using `local_time` is only meant to be
```

westonpace (approved):
Thanks, the description is clear to me.  Two small grammar nits.

rok:

rok:
Thanks @westonpace, merged the grammar changes.

westonpace:
The test failures seem unrelated (I've opened a new issue for one which appears to be an intermittent segfault).

jorisvandenbossche:
Nice to see this functionality! 

One comment on the name of the function: given that "local_time" could also be interpreted as returning just the time (a `time32/64()` type) and not both date and time (and theoretically, this could also be a kernel we would want to add at some point), we could maybe use `"local_timestamp"` instead?  
(I know that some will day that "local timestamp" is a contradictio in terminis, and that it should be "local datetime" instead, but since we don't use "datetime" for any other kernel name and "timestamp" is the name of the type anyway, that seems the best option).

For reference, in Joda they call this `toLocalDateTime` (and they also have a `toLocalDate` and `toLocalTime`), in pandas the naming is confusing (`tz_localize(None)`), and in R's lubridate there is a `local_time` that does _somewhat_ related, except that it returns the difference in seconds instead of new timestamp object.

cpp/src/arrow/compute/kernels/scalar_temporal_unary.cc
```diff
@@ -1801,8 +1831,9 @@ const FunctionDoc ceil_temporal_doc{
 const FunctionDoc round_temporal_doc{
     "Round temporal values to the nearest multiple of specified time unit",
     ("Null values emit null.\n"
-     "An error is returned if the values have a defined timezone but it\n"
-     "cannot be found in the timezone database."),
+     "If timezone is not given then timezone naive timestamp in UTC are\n"
+     "returned. An error is returned if the values have a defined timezone\n"
+     "but it cannot be found in the timezone database."),
```
jorisvandenbossche, line 1836: Was this change intentional? Or was it meant to be in `local_time_doc`? 
(and if intentional, why only for round and not ceil/floor?)

cpp/src/arrow/compute/kernels/scalar_temporal_unary.cc
```diff
@@ -1784,6 +1807,13 @@ const FunctionDoc is_dst_doc{
      "An error is returned if the values do not have a defined timezone."),
     {"values"}};
 
+const FunctionDoc local_time_doc{
+    "Convert timestamp to a timezone-naive local time timestamp",
+    ("LocalTime converts a timestamp to a local time of timestamps timezone\n"
```
jorisvandenbossche, line 1812: Something is missing in this sentence?

cpp/src/arrow/compute/kernels/scalar_temporal_unary.cc
```diff
@@ -1784,6 +1807,13 @@ const FunctionDoc is_dst_doc{
...
+     "and removes timezone metadata. If input is in UTC or doesn't have\n"
+     "timezone metadata, it is returned as is.\n"
```
jorisvandenbossche, line 1814: Returned "as is" is not entirely correct for UTC, as I assume the "UTC" timezone is still removed?

jorisvandenbossche (commented):
And a few doc comments/questions

rok:
Thanks for the review @jorisvandenbossche. I've opened a PR to address it: #34263.

PR #34209, opened 2023-02-15
GH-34150: [C++] Fix error due to improper initialization of conversion option defaults

### LINKED ISSUE #34150: [C++] [Python] Fix improper initialization of `ConversionOptions`

westonpace:
* Closes: #34150

### DIFF
--- a/cpp/src/arrow/engine/substrait/options.cc
+++ b/cpp/src/arrow/engine/substrait/options.cc
@@ -170,45 +170,60 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
 
 namespace {
 
-std::shared_ptr<ExtensionProvider> g_default_extension_provider =
-    std::make_shared<DefaultExtensionProvider>();
+template <typename T>
+class ConfigurableSingleton {
+ public:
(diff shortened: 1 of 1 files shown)


westonpace:
@rtpsw do we think this is still a good idea?

rtpsw:
> @rtpsw do we think this is still a good idea?

I do (see also [my post here](https://github.com/apache/arrow/pull/34156#issuecomment-1435941288)). It could help avoid improper initialization from C++ users.

PR #34229, opened 2023-02-16
GH-34228: [R] Add LIB_DIR when Arrow is found via pkg-config

### LINKED ISSUE #34228: [R] Extra features not found when Arrow is detected with pkg

I am well beyond my depth here but I was having trouble getting R to detect that my Arrow build had the datasets feature enabled (I don't normally build R but built it recently to debug an issue and ran into this).

I get the message `echo "*** Arrow C++ libraries found via pkg-config at $FOUND_LIB_DIR"` so R is finding Arrow that way.  However, later, it seems to want to find the extra features in this way:

```
  # Check for features
  ARROW_OPTS_CMAKE="$LIB_DIR/cmake/Arrow/ArrowOptions.cmake"

  arrow_built_with() {
    # Function to check cmake options for features
    grep -i 'set('"$1"' "ON")' $ARROW_OPTS_CMAKE >/dev/null 2>&1
  }
```

Since `LIB_DIR` is unset it does not find the `ArrowOptions.cmake` file (even though I have one and it is installed).  I have a fix that sets `LIB_DIR` in the pkg-config path and will create a PR but I have no idea if that is the best fix.

### Component(s)

R

westonpace:
* Closes: #34228

### DIFF
--- a/r/configure
+++ b/r/configure
@@ -110,6 +110,7 @@ else
     PKG_CFLAGS="$PKGCONFIG_CFLAGS $PKG_CFLAGS"
     PKG_LIBS="${PKGCONFIG_LIBS}"
     PKG_DIRS="${PKGCONFIG_DIRS}"
+    LIB_DIR=${FOUND_LIB_DIR}
 
     # Check for version mismatch
     PC_LIB_VERSION=`pkg-config --modversion arrow`


westonpace:
As mentioned in the issue, I am not entirely sure what is going on or if this is the best approach but it works in my limited use.

kou (approved):
+1

PR #34231, opened 2023-02-17
GH-34230: [Java] Call allocation listener on BaseAllocator#wrapForeignAllocation

### LINKED ISSUE #34230: [Java]  Call allocation listener on BaseAllocator#wrapForeignAllocation

Buffer created by BaseAllocator#wrapForeignAllocation calls `AllocationListener#onRelease` but the Allocator didn't call `AllocationListener#onAllocate()` when the buffer was created.

### Component(s)

Java

zhztheplayer:
#34230 
* Closes: #34230

### DIFF
--- a/java/memory/memory-core/src/main/java/org/apache/arrow/memory/BaseAllocator.java
+++ b/java/memory/memory-core/src/main/java/org/apache/arrow/memory/BaseAllocator.java
@@ -234,6 +234,41 @@ abstract class BaseAllocator extends Accountant implements BufferAllocator {
   }
 
   @Override
+  public ArrowBuf wrapForeignAllocation(ForeignAllocation allocation) {
+    assertOpen();
+    final long size = allocation.getSize();
+    listener.onPreAllocation(size);
+    AllocationOutcome outcome = this.allocateBytes(size);
+    if (!outcome.isOk()) {
+      if (listener.onFailedAllocation(size, outcome)) {
+        // Second try, in case the listener can do something about it
+        outcome = this.allocateBytes(size);
+      }
+      if (!outcome.isOk()) {
+        throw new OutOfMemoryException(createErrorMsg(this, size,
+            size), outcome.getDetails());
+      }
+    }
+    try {
+      final AllocationManager manager = new ForeignAllocationManager(this, allocation);
+      final BufferLedger ledger = manager.associate(this);
+      final ArrowBuf buf =
+          new ArrowBuf(ledger, /*bufferManager=*/null, size, allocation.memoryAddress());
+      buf.writerIndex(size);
+      listener.onAllocation(size);
+      return buf;
+    } catch (Throwable t) {
+      try {
+        releaseBytes(size);
+        allocation.release0();
+      } catch (Throwable e) {
+        t.addSuppressed(e);
+      }
+      throw t;
+    }
+  }
+
+  @Override
   public ArrowBuf buffer(final long initialRequestSize) {
     assertOpen();
 
--- a/java/memory/memory-core/src/main/java/org/apache/arrow/memory/BaseAllocator.java
+++ b/java/memory/memory-core/src/main/java/org/apache/arrow/memory/BaseAllocator.java
@@ -260,6 +260,10 @@ abstract class BaseAllocator extends Accountant implements BufferAllocator {
     } catch (Throwable t) {
       try {
         releaseBytes(size);
+      } catch (Throwable e) {
+        t.addSuppressed(e);
+      }
+      try {
         allocation.release0();
       } catch (Throwable e) {
         t.addSuppressed(e);
--- a/java/memory/memory-netty/src/test/java/org/apache/arrow/memory/CountingAllocationListener.java
+++ b/java/memory/memory-netty/src/test/java/org/apache/arrow/memory/CountingAllocationListener.java
@@ -17,9 +17,11 @@
 
 package org.apache.arrow.memory;
 
-// Allocation listener
-// It counts the number of times it has been invoked, and how much memory allocation it has seen
-// When set to 'expand on fail', it attempts to expand the associated allocator's limit
+/**
+ * Allocation listener.
+ * It counts the number of times it has been invoked, and how much memory allocation it has seen
+ * When set to 'expand on fail', it attempts to expand the associated allocator's limit.
+ */
 final class CountingAllocationListener implements AllocationListener {
   private int numPreCalls;
   private int numCalls;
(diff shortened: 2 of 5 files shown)


java/memory/memory-netty/src/test/java/org/apache/arrow/memory/CountingAllocationListener.java
```diff
@@ -0,0 +1,111 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.arrow.memory;
+
+// Allocation listener
```
lidavidm, line 20: nit, but please stick to JavaDoc syntax for this?

java/memory/memory-core/src/main/java/org/apache/arrow/memory/BaseAllocator.java
```diff
@@ -233,6 +233,41 @@ private void childClosed(final BaseAllocator childAllocator) {
...
+      final BufferLedger ledger = manager.associate(this);
+      final ArrowBuf buf =
+          new ArrowBuf(ledger, /*bufferManager=*/null, size, allocation.memoryAddress());
+      buf.writerIndex(size);
+      listener.onAllocation(size);
+      return buf;
+    } catch (Throwable t) {
+      try {
+        releaseBytes(size);
+        allocation.release0();
```
lidavidm, line 263: can this be in a separate try-catch?

java/memory/memory-core/src/main/java/org/apache/arrow/memory/BaseAllocator.java
```diff
@@ -233,6 +233,41 @@ private void childClosed(final BaseAllocator childAllocator) {
     listener.onChildRemoved(this, childAllocator);
   }
 
+  @Override
+  public ArrowBuf wrapForeignAllocation(ForeignAllocation allocation) {
+    assertOpen();
+    final long size = allocation.getSize();
+    listener.onPreAllocation(size);
+    AllocationOutcome outcome = this.allocateBytes(size);
```
lidavidm, line 241: Hmm, if we don't force allocation the behavior is different - the memory is already allocated, we're just trying to account for it. This will 'eagerly' release resources (via the listener) to make room while the existing behavior is to do it 'lazily' (the next allocation will trigger things if needed). Is that what you want?

java/memory/memory-netty/src/test/java/org/apache/arrow/memory/TestForeignAllocation.java
```diff
@@ -54,6 +54,55 @@ public void wrapForeignAllocation() {
...
+      assertEquals(0, listenedAllocator.getAllocatedMemory());
+    }
+    assertEquals(1, listener.getNumPreCalls());
+    assertEquals(1, listener.getNumCalls());
+    assertEquals(1, listener.getNumReleaseCalls());
+    assertEquals(16, listener.getTotalMem());
+  }
+
+  @Test(expected = OutOfMemoryException.class)
+  public void wrapForeignAllocationFailedWithAllocationListener() {
```
lidavidm, line 85: Can we also test the case where the listener frees some memory and the allocation succeeds?

lidavidm (changes requested):
Thanks, I left some comments.

java/memory/memory-core/src/main/java/org/apache/arrow/memory/BaseAllocator.java
zhztheplayer: Thanks for noticing this. I think it's not a very big issue as long as we treat "accountation of foreign allocation" as another type of "allocation" too. Since sometimes it can be tricky to track on actual foreign allocation at real-time from Java side.

Also another solution would be just excluding foreign allocation from `listener.release()` too. But in that way one will lose tracking if he didn't set up a independent path to report the foreign allocations.

java/memory/memory-netty/src/test/java/org/apache/arrow/memory/TestForeignAllocation.java
zhztheplayer: Sure, I'll add a case

zhztheplayer:
@lidavidm Would you like to take another look? Thanks.

java/memory/memory-core/src/main/java/org/apache/arrow/memory/BaseAllocator.java
zhztheplayer: Thanks but would you give some hints on this? Or would you suggest adding a local variable `Throwable xx` to save the error for further handling?

lidavidm: What I mean is that `release0` should be called even if `releaseBytes` throws.

java/memory/memory-core/src/main/java/org/apache/arrow/memory/BaseAllocator.java
lidavidm: As long as it's consistent. I think the behavior here is OK, and consistent with the "allocator" actually being a "ledger".

java/memory/memory-core/src/main/java/org/apache/arrow/memory/BaseAllocator.java
zhztheplayer: Thanks. Updated.

java/memory/memory-netty/src/test/java/org/apache/arrow/memory/TestForeignAllocation.java
```diff
@@ -54,6 +57,80 @@ public void wrapForeignAllocation() {
...
+      }
+    }
+  }
+
+  @Test
+  public void wrapForeignAllocationWithAllocationListenerReclaimingSpace() {
+    final long bufferSize = 16;
+    final long limit = 2 * bufferSize - 1;
+
+    final List<ArrowBuf> bufferedToBeFreed = new ArrayList<>();
```
lidavidm, line 113: nit: typo 

```suggestion
    final List<ArrowBuf> buffersToBeFreed = new ArrayList<>();
```

lidavidm (approved):
Thanks, just a typo.

lidavidm (approved):

Issue #14917, opened 2022-12-12
[C++] Issues with GTest when building Arrow C++ on MacOS


AlenkaF:
### Describe the bug, including details regarding any error messages, version, and platform.

I have seen errors with GTest when building Arrow C++ on MacOS come up regularly, probably due to lack of some symbols or incompatibility of the supplied library . An error log is similar to:

```
FAILED: debug/arrow-array-test 
[...]

PR #3440, opened 2019-01-20
ARROW-4306: [Release] Update website, write blog post for 0.12.0 release

wesm:

PR #34234, opened 2023-02-17
GH-34235: [Python] Add `join_asof` binding

### LINKED ISSUE #34235: [Python] Add `join_asof` binding

The `AsofJoinNode` currently exists in the C++ implementation but there is no binding to it in PyArrow. It would be great to add this!

### Component(s)

Python

judahrand:
* Closes: #34235


judahrand:
I'm struggling to run the tests locally as I can't get Arrow to build on an M1 Mac.

```
  "testing::Matcher<std::__1::basic_string_view<char, std::__1::char_traits<char> > const&>::Matcher(char const*)", referenced from:
      testing::Matcher<std::__1::basic_string_view<char, std::__1::char_traits<char> > const&> testing::internal::MatcherCastImpl<std::__1::basic_string_view<char, std::__1::char_traits<char> > const&, char const*>::CastImpl<true>(char const* const&, std::__1::integral_constant<bool, true>, std::__1::integral_constant<bool, true>) in array_test.cc.o
      testing::Matcher<std::__1::basic_string_view<char, std::__1::char_traits<char> > const&> testing::internal::MatcherCastImpl<std::__1::basic_string_view<char, std::__1::char_traits<char> > const&, char const*>::CastImpl<true>(char const* const&, std::__1::integral_constant<bool, true>, std::__1::integral_constant<bool, true>) in array_binary_test.cc.o
make[2]: *** [src/arrow/CMakeFiles/arrow-array-test.dir/build.make:208: build/debug/arrow-array-test] Error 1
make[1]: *** [CMakeFiles/Makefile2:1714: src/arrow/CMakeFiles/arrow-array-test.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[ 54%] Built target arrow-public-api-test
```

AlenkaF:
> I'm struggling to run the tests locally as I can't get Arrow to build on an M1 Mac.

Can you try adding `-DGTest_SOURCE="BUNDLED"` CMake flag? See https://github.com/apache/arrow/issues/14917

judahrand:
@AlenkaF I believe that this is ready for a round of review when you have time.

AlenkaF:
Hi @judahrand , thank you for your contribution! ⭐ 

The code LGTM 👍  I would maybe add tests to `test_table.py` also, as in https://github.com/apache/arrow/pull/12452/files#diff-72bd29ba764c85f18fbf9e74898b72d47ed8a1b04be879c1a5b2a59382e2eaef

judahrand:
> Hi @judahrand , thank you for your contribution! ⭐
> 
> The code LGTM 👍 I would maybe add tests to `test_table.py` also, as in https://github.com/apache/arrow/pull/12452/files#diff-72bd29ba764c85f18fbf9e74898b72d47ed8a1b04be879c1a5b2a59382e2eaef

Done

judahrand:
The Appveyor failure looks like a timeout?

AlenkaF:
Yeah, I see the Appveyor build is failing on other PRs also.

judahrand:
@AlenkaF Do you need me to do anything else on this PR?

AlenkaF (approved):
This looks good to me, thanks!
Will wait for Joris to have time to look at it also.

AlenkaF:
Just one more thing: could you rebase to latest master to get AppVeyor working? (related issue was closed: https://github.com/apache/arrow/issues/34296)

jorisvandenbossche:
@judahrand thanks for working on this!

Before diving into the details, I have two general comments:

- We are in the middle of refactoring how we expose the Acero / ExecPlan features, and I have a PR that exposes the Declaration object and ExecNodeOptions subclasses in Python (https://github.com/apache/arrow/pull/34102). Once that is merged, it should be the goal that also the asof join could be exposed by adding an `AsofJoinNodeOptions` class in pyarrow. 
  Ideally, I would prefer that we can do this directly, but I know that the mentioned PR isn't merged yet (I hope it can be merged in one of the coming days, though)

- I think we should try to do a better job of explaining what the "asof" exactly is and does in the docstring (I also noted that AsOfJoin is missing in the C++ user guide (will open an issue about that), although it has some reference docs), since I think this is generally not a very well known join type: what is the difference with a normal join? What is the difference between the "on" and "by" join keys?

python/pyarrow/_dataset.pyx
```diff
@@ -514,6 +514,53 @@ cdef class Dataset(_Weakrefable):
...
+        right_dataset : dataset
+            The dataset to join to the current one, acting as the right dataset
+            in the join operation.
+        on : str
+            The column from current dataset that should be used as the on key
+            of the join operation left side.
+        by : str or list[str]
+            The columns from current dataset that should be used as the by keys
+            of the join operation left side.
+        tolerance : int
```
jorisvandenbossche, line 535: If the "on" key is a timestamp column, what value can be used here? (not an int?)

python/pyarrow/_dataset.pyx
```diff
@@ -514,6 +514,53 @@ cdef class Dataset(_Weakrefable):
...
+            The tolerance for inexact "on" key matching. A right row is considered
+            a match with the left row `right.on - left.on <= tolerance`. The
+            `tolerance` may be:
+                - negative, in which case a past-as-of-join occurs;
+                - or positive, in which case a future-as-of-join occurs;
+                - or zero, in which case an exact-as-of-join occurs.
```
jorisvandenbossche, line 541:
```suggestion

            - negative, in which case a past-as-of-join occurs;
            - or positive, in which case a future-as-of-join occurs;
            - or zero, in which case an exact-as-of-join occurs.
```

Formatting nitpick: restructuredtext is quite strict about this, and so a list shouldn't use indentation but requires a blank line before and after

python/pyarrow/_exec_plan.pyx
```diff
@@ -390,6 +390,120 @@ def _perform_join(join_type, left_operand not None, left_keys,
...
+
+    c_input_keys.push_back(c_right_keys)
+
+    # By default expose all columns on both left and right table
+    if isinstance(left_operand, Table):
+        left_columns = left_operand.column_names
+    elif isinstance(left_operand, Dataset):
+        left_columns = left_operand.schema.names
+    else:
+        raise TypeError("Unsupported left join member type")
```
jorisvandenbossche, line 462: Those `left_columns`/`right_columns` is not really used, except for checking the column collisions? 
What happens if we don't check for this here in cython and there actually is a column collision? Does the C++ implementation give an error for that as well?

python/pyarrow/table.pxi
```diff
@@ -4993,6 +4993,78 @@ cdef class Table(_PandasConvertible):
...
+        >>> df1 = pd.DataFrame({'id': [1, 2, 3],
+        ...                     'year': [2020, 2022, 2019]})
+        >>> df2 = pd.DataFrame({'id': [3, 4],
+        ...                     'year': [2020, 2021],
+        ...                     'n_legs': [5, 100],
+        ...                     'animal': ["Brittle stars", "Centipede"]})
+        >>> t1 = pa.Table.from_pandas(df1).sort_by('year')
+        >>> t2 = pa.Table.from_pandas(df2).sort_by('year')
+
+        >>> t1.join_asof(t2, 'year', 'id', 1).combine_chunks().sort_by('year')
```
jorisvandenbossche, line 5049: Can you use explicit keyword names (except for the first table argument)? I think that will make it easier to understand what those arguments are.

python/pyarrow/tests/test_dataset.py
```diff
@@ -4839,6 +4839,57 @@ def test_dataset_join_collisions(tempdir):
...
+
+    t2 = pa.Table.from_pydict({
+        "colB": [2, 9, 15],
+        "col3": ["a", "b", "g"],
+        "colC": [1., 3., 5.]
+    })
+    ds.write_dataset(t2, tempdir / "t2", format="ipc")
+    ds2 = ds.dataset(tempdir / "t2", format="ipc")
+
+    result = ds1.join_asof(ds2, "colA", "col2", 1, "colB", "col3")
```
jorisvandenbossche, line 4859: Also here, can you use keyword names instead of positional arguments (except for the first argument) to make it a bit more readable?

python/pyarrow/tests/test_table.py
```diff
@@ -2297,6 +2297,49 @@ def test_table_join_many_columns():
...
+        "col2": ["a", "b", "a", "b", "f"]
+    })
+
+    t2 = pa.Table.from_pydict({
+        "colB": [2, 9, 15],
+        "col3": ["a", "b", "g"],
+        "colC": [1., 3., 5.]
+    })
+
+    r = t1.join_asof(t2, "colA", "col2", 1, "colB", "col3")
```
jorisvandenbossche, line 2313: Some additional test case ideas to ensure good coverage:

- A test where the left/right column names are the same, so you can rely on not having to specify right_on/by
- A test where the `by` keys is a list of columns instead of a single one

jorisvandenbossche:
@judahrand I merged https://github.com/apache/arrow/pull/34102, so you should be able to add a `AsofJoinNodeOptions` subclass, and then update `_perform_join_asof` to use that options class together with Declaration (similar as I am doing in https://github.com/apache/arrow/pull/3440)

python/pyarrow/_exec_plan.pyx
judahrand: I believe the reason I adding this in is that it was cause the C++ implementation to segfault.

judahrand: The `left_columns/right_columns` variables are also used to filter out the 'special' Dataset columns which we get back if the operands are datasets.

python/pyarrow/_dataset.pyx
judahrand: This is a good question actually... I'm not 100% sure how this is intended to work. The C++ implementation exclusively accepts an `int64_t` for the tolerance.

python/pyarrow/tests/test_table.py
judahrand: > A test where the left/right column names are the same, so you can rely on not having to specify right_on/by

This is now tested.

> A test where the by keys is a list of columns instead of a single one

This is now tested.

> and what happens if passing an empty list?

It seems like it just doesn't perform the join over any partitions - this is also now tested.

python/pyarrow/_exec_plan.pyx
jorisvandenbossche: OK, I see. I added ScanNodeOptions, so you should be able to update this now.

python/pyarrow/_acero.pyx
```diff
@@ -340,6 +340,78 @@ class HashJoinNodeOptions(_HashJoinNodeOptions):
...
+    ----------
+    left_on : str, Expression
+        The left key on which the join operation should be performed.
+        Can be a string column name or a field expression.
+    left_by: str, Expression or list[str]
+        The left keys on which the join operation should be performed.
+        Each key can be a string column name or a field expression,
+        or a list of such field references.
+    right_operand : Table or Dataset
+        The right operand for the join operation.
```
jorisvandenbossche, line 399: This can be left out here in the options docstring (that's not a parameter)

python/pyarrow/_dataset.pyx
```diff
@@ -857,6 +857,54 @@ cdef class Dataset(_Weakrefable):
...
+        ----------
+        right_dataset : dataset
+            The dataset to join to the current one, acting as the right dataset
+            in the join operation.
+        on : str
+            The column from current dataset that should be used as the on key
+            of the join operation left side.
+        by : str or list[str]
+            The columns from current dataset that should be used as the by keys
+            of the join operation left side.
```
jorisvandenbossche, line 877: We will need to explain here what's the difference between "on" and "by" keys.

python/pyarrow/tests/test_exec_plan.py
```diff
@@ -321,3 +321,77 @@ def test_join_extension_array_column():
...
+        "colB": [99, 20, 10],
+        "colVals": ["Z", "B", "A"],
+        "colUniq": [100, 200, 300],
+        "colA": [99, 2, 1],
+        "on": [2, 3, 4],
+    })
+
+    msg = "colVals present in both tables. AsofJoin does not support column collisions."
+    with pytest.raises(ValueError, match=msg):
+        ep._perform_join_asof(t1, "on", ["colA", "colB"], t2, "on", ["colA", "colB"], 1)
```
jorisvandenbossche, line 397: Another failure case to test: what happens if you pass by keys of different length? (or specifying columns where the type doesn't match)

jorisvandenbossche (commented):
Thanks for the updates!

jorisvandenbossche:
@judahrand to be clear, there is still interest in this!

judahrand:
I'll try and find the time to push this through 🙏

python/pyarrow/_acero.pyx
judahrand: https://github.com/apache/arrow/pull/34234/commits/6f8db25fec78546e19c91e405a3f1c14a0ddcd64

python/pyarrow/tests/test_exec_plan.py
judahrand: https://github.com/apache/arrow/pull/34234/commits/102f965f83ba702997b74dd72626b78970cfb91a

python/pyarrow/_dataset.pyx
judahrand: https://github.com/apache/arrow/pull/34234/commits/d66a2e8089d422cfdf9bd605c628cda45e24142f

judahrand:
@jorisvandenbossche Sorry this took me so long to revisit! I think I've dealt with all of your comments?

judahrand:
> @jorisvandenbossche Sorry this took me so long to revisit! I think I've dealt with all of your comments?

Merged main in which should fix the appveyor test failure.

python/pyarrow/acero.py
```diff
@@ -253,6 +254,93 @@ def _perform_join(join_type, left_operand, left_keys,
...
+
+    join_opts = AsofJoinNodeOptions(
+        left_on, left_by, right_on, right_by, tolerance
+    )
+    decl = Declaration(
+        "asofjoin", options=join_opts, inputs=[left_source, right_source]
+    )
+
+    # Last time I checked the asof join node does not support using threads.
+    result_table = decl.to_table(use_threads=False)
```
AlenkaF, line 334: Yeah, _I think_ it uses threads by default.

https://github.com/apache/arrow/blob/57e9386d4a3a18dfa3c5269c49efaee93aabea2b/cpp/src/arrow/acero/CMakeLists.txt#L177-L182

But it is a bit unclear to me how the `use_threads` in the `DeclarationToTable` corresponds to the `use_threads` in the specific join. 

https://github.com/apache/arrow/blob/57e9386d4a3a18dfa3c5269c49efaee93aabea2b/python/pyarrow/_acero.pyx#L495-L501

https://github.com/apache/arrow/blob/57e9386d4a3a18dfa3c5269c49efaee93aabea2b/cpp/src/arrow/acero/exec_plan.cc#L779-L790

I would think the threading here is separate from the threading in the specific join - but I am not very familiar with the acero codebase. Have you tried setting it to `True`/default? What happens in that case?

AlenkaF (commented):
Thank you for continuing the work on this PR! Much appreciated.
I only have small questions/comments, nothing mayor.

It will have to be approved by somebody more experienced before emerging though ;)

python/pyarrow/_dataset.pyx
```diff
@@ -874,6 +874,63 @@ cdef class Dataset(_Weakrefable):
...
+            An inexact match is used on the “on” key.i.e., a row is considered a
+            match iff left_on - tolerance <= right_on <= left_on.
+
+            The input table must be sorted by the “on” key. Must be a single
+            field of a common type.
+
+            Currently, the “on” key must be an integer, date, or timestamp type.
+        by : str or list[str]
+            The columns from current dataset that should be used as the keys
+            of the join operation left side.
```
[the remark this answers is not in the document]
judahrand: https://github.com/apache/arrow/pull/34234/commits/acb082e8da4647c13a60484f654d7ca1425f7f6e

python/pyarrow/acero.py
judahrand: `use_threads=True` does seem to work! It didn't when I originally opened this PR way back when but now all seems well!

AlenkaF:
Thanks!

AppVeyor failure is not connected - seems to be failing elsewhere also. Opened https://github.com/apache/arrow/issues/38431

I have re-run `C GLib & Ruby` job, don't think the failure is connected though.

Otherwise this looks good to me 👍  @jorisvandenbossche would you like to have a look before we merge?

judahrand:
@jorisvandenbossche Do you think you might be able to take another look at this? I know I've been quite slow at pushing this PR forward but I think it potentially ready to be merged now.

judahrand:
@jorisvandenbossche ping?

judahrand:
@jorisvandenbossche @AlenkaF ping?

python/pyarrow/table.pxi
```diff
@@ -4859,6 +4859,91 @@ cdef class Table(_Tabular):
             output_type=Table
         )
 
+    def join_asof(self, right_table, on, by, tolerance, right_on=None, right_by=None):
+        """
+        Perform an asof join between this table and another one.
+
```
jorisvandenbossche, line 4865: I would still like to see a bit more expanded explanation (apart from the individual keyword explanations) about _what_ and asof join exactly is.

Something indicating it does 1) an inexact join, 2) on a sorted dataset, 3) potentially first joining on other attributes, and 4) typically useful for time series data that are not perfectly aligned. Are that the most relevant characteristics?

python/pyarrow/table.pxi
```diff
@@ -4859,6 +4859,91 @@ cdef class Table(_Tabular):
...
+        ...                     'year': [2020, 2022, 2019]})
+        >>> df2 = pd.DataFrame({'id': [3, 4],
+        ...                     'year': [2020, 2021],
+        ...                     'n_legs': [5, 100],
+        ...                     'animal': ["Brittle stars", "Centipede"]})
+        >>> t1 = pa.Table.from_pandas(df1).sort_by('year')
+        >>> t2 = pa.Table.from_pandas(df2).sort_by('year')
+
+        >>> t1.join_asof(
+        ...     t2, on='year', by='id', tolerance=1
```
jorisvandenbossche, line 4926: The fact that there is no repetition of the vaues in the by "id" key in the example data makes that it is difficult to see what exactly happens with the "on" key?

python/pyarrow/tests/test_exec_plan.py
```diff
@@ -323,6 +323,83 @@ def test_join_extension_array_column():
...
+        -5,
+        {
+            "colA": [1, 1, 5, 6, 7],
+            "col2": ["a", "b", "a", "b", "f"],
+            "colC": [None, None, 1., None, None],
+        },
+    ),
+])
+@pytest.mark.parametrize("use_datasets", [False, True])
+def test_join_asof(tolerance, expected, use_datasets):
```
jorisvandenbossche, line 353: This file of `test_exec_plan.py` exist more for historical reasons, from before we had the Acero bindings. Do the test you added here add anything in addition to what is either tested in `test_acero.py` or `test_table.py`/`test_dataset.py`? If so, I think you can leave out those tests, or if it does test something else, add that to one of the other files.

jorisvandenbossche (commented):
Again apologies for the slow review! (and thanks for keep pinging us)

I added some comments on the docstring in acero.pyx (the options class), but most of them should also apply on the docstrings in table/dataset methods.

python/pyarrow/tests/test_exec_plan.py
judahrand: https://github.com/apache/arrow/pull/34234/commits/5fcce5f689d4ff42af13baec2816f8e49345e075

judahrand: https://github.com/apache/arrow/pull/34234/commits/5fcce5f689d4ff42af13baec2816f8e49345e075

python/pyarrow/table.pxi
judahrand: https://github.com/apache/arrow/pull/34234/commits/e6da4df5cae7464439ae51c92ca1f43012a331a1

python/pyarrow/table.pxi
judahrand: https://github.com/apache/arrow/pull/34234/commits/905bb54307da77541085e694c04d10ce92eee56b

judahrand:
@jorisvandenbossche I believe I've dealt will all the feedback 😄

jorisvandenbossche (approved):

PR #34259, opened 2023-02-19
GH-34258: [C++]  Add suppport for building with system flatbuffers

### LINKED ISSUE #34258: [C++] Allow building against an installed flatbuffers library

Almost all dependencies can be provided by the system. Flatbuffers is one of those exceptions.

See also: https://github.com/apache/arrow/pull/13775#pullrequestreview-1062900676.

### Component(s)

C++

tobim:

* Make it possible to use `-Dflatbuffers_SOURCE=SYSTEM`. 
* Change `update-flatbuffers.sh` so that the output directory can optionally be passed in.
* Generate flatbuffers headers with `update-flatbuffers.sh` when flatbuffers is provided by the system.
* Remove plasma file generation from `update-flatbuffers.sh`.

### Are these changes tested?

This change is opt-in, so I'm not sure if it makes sense to extend the build matrix for it.
* Closes: #34258

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -541,7 +541,20 @@ include_directories(${CMAKE_CURRENT_BINARY_DIR}/src)
 include_directories(src)
 
 # Compiled flatbuffers files
-include_directories(src/generated)
+if("${flatbuffers_SOURCE}" STREQUAL "SYSTEM")
+  file(GLOB _flatbuffers_files "${CMAKE_CURRENT_SOURCE_DIR}/../format/*.fbs")
+  add_custom_command(OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/src/generated
+                     COMMAND "${BUILD_SUPPORT_DIR}/update-flatbuffers.sh"
+                             "${CMAKE_CURRENT_BINARY_DIR}/src/generated"
+                     DEPENDS ${_flatbuffers_files}
+                     VERBATIM)
+  unset(_flatbuffers_files)
+  add_custom_target(generate-flatbuffers
+                    DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/src/generated)
+  include_directories(${CMAKE_CURRENT_BINARY_DIR}/src/generated)
+else()
+  include_directories(src/generated)
+endif()
 
 #
 # Visibility
@@ -652,9 +665,9 @@ endif()
 # Note that arrow::hadoop is a header only target that refers
 # cpp/thirdparty/hadoop/include/. See
 # cpp/cmake_modules/ThirdpartyToolchain.cmake for details.
-set(ARROW_SHARED_LINK_LIBS arrow::flatbuffers arrow::hadoop)
+set(ARROW_SHARED_LINK_LIBS flatbuffers::flatbuffers arrow::hadoop)
 set(ARROW_SHARED_INSTALL_INTERFACE_LIBS)
-set(ARROW_STATIC_LINK_LIBS arrow::flatbuffers arrow::hadoop)
+set(ARROW_STATIC_LINK_LIBS flatbuffers::flatbuffers arrow::hadoop)
 set(ARROW_STATIC_INSTALL_INTERFACE_LIBS)
 
 if(ARROW_USE_BOOST)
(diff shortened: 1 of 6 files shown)


cpp/cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -313,15 +320,23 @@ endmacro()
...
+    target_include_directories(flatbuffers::flatbuffers
+                               INTERFACE "${THIRDPARTY_DIR}/flatbuffers/include")
+  endif()
+endmacro()
+
+resolve_dependency(flatbuffers
+                   IS_RUNTIME_DEPENDENCY
+                   FALSE
+                   USE_CONFIG
+                   TRUE)
```
kou, line 339: Can we remove this?
In general, `Findflatbuffers.cmake` doesn't exist. So we don't need this.

cpp/build-support/update-flatbuffers.sh
```diff
@@ -18,30 +18,22 @@
...
+# it defaults to /cpp/src/generated.
 
 set -euo pipefail
 
 CWD="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" && pwd)"
 SOURCE_DIR="$CWD/../src"
-PYTHON_SOURCE_DIR="$CWD/../../python"
 FORMAT_DIR="$CWD/../../format"
-TOP="$FORMAT_DIR/.."
 FLATC="flatc"
```
kou, line 30: How about running `flatc` from CMake directly instead of using this shell script?

Because:
* We need to use `flatbuffers::flatc` instead of `flatc`.
* If we use shell script, we can't use system FlatBuffers on Windows.

cpp/CMakeLists.txt
```diff
@@ -541,7 +541,20 @@ include_directories(${CMAKE_CURRENT_BINARY_DIR}/src)
...
+if("${flatbuffers_SOURCE}" STREQUAL "SYSTEM")
+  file(GLOB _flatbuffers_files "${CMAKE_CURRENT_SOURCE_DIR}/../format/*.fbs")
+  add_custom_command(OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/src/generated
+                     COMMAND "${BUILD_SUPPORT_DIR}/update-flatbuffers.sh"
+                             "${CMAKE_CURRENT_BINARY_DIR}/src/generated"
+                     DEPENDS ${_flatbuffers_files}
+                     VERBATIM)
+  unset(_flatbuffers_files)
+  add_custom_target(generate-flatbuffers
+                    DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/src/generated)
```
kou, line 553: How about one more custom target that updates bundled generated files in `src/generated/` like `update-generated-flatbuffers`?
If we provide it, we can remove `cpp/build-support/update-flatbuffers.sh`.

cpp/CMakeLists.txt
```diff
@@ -541,7 +541,20 @@ include_directories(${CMAKE_CURRENT_BINARY_DIR}/src)
...
+  include_directories(${CMAKE_CURRENT_BINARY_DIR}/src/generated)
+else()
+  include_directories(src/generated)
```
kou, line 556: Could you attach them to `arrow::flatbuffers` instead of using `include_directories()`?

cpp/cmake_modules/ThirdpartyToolchain.cmake
kou, line 339: Could you add a new `ARROW_FLATBUFFERS_USE_SHARED` CMake option?
If it's true, we use `flatbuffers::flatbuffers_shared` instead of `flatbuffers::flatbuffers`.

How about keeping the current `arrow::flatbuffers` target? It refers the bundled FlatBuffers with `flatbuffers_SOURCE == TRUE`. It refers `flatbuffers::flatbuffers_shared` with `ARROW_FLATBUFFERS_USE_SHARED == TRUE`. It refers `flatbuffers::flatbuffers` otherwise.

cpp/cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -313,15 +320,23 @@ endmacro()
...
+                                     "${THIRDPARTY_DIR}/flatbuffers/include")
+  else()
+    target_include_directories(flatbuffers::flatbuffers
+                               INTERFACE "${THIRDPARTY_DIR}/flatbuffers/include")
+  endif()
+endmacro()
+
+resolve_dependency(flatbuffers
+                   IS_RUNTIME_DEPENDENCY
+                   FALSE
```
kou, line 337: Is this correct?
It seems that `flatbuffers::flatbuffers` depends on `libflatbuffers.a` and `flatbuffers::flatbuffers_shared` depends on `libflatbuffers.so`.

amol-:

tobim:
Sorry for not following up. I don't have capacity to work on this at the moment. I still plan to integrate the change request eventually.

PR #34263, opened 2023-02-20
GH-33143: [C++] Naming and doc/test changes for local_time compute kernel

### LINKED ISSUE #33143: [C++] Kernel to convert timestamp with timezone to wall time

rok:
### Rationale for this change

A better naming for local_time kernel was proposed in post merge review of #34208.

Change `local_time` to `local_timestamp` and related docs/test changes.

### Are these changes tested?

Yes.

Changing `local_time` to `local_timestamp` is a user facing change. But since it was not yet released we can probably treat it as non-breaking.
* Closes: #33143


kou:
Could you open a new issue for this?
This is not a MINOR change.
See also: https://github.com/apache/arrow/blob/main/CONTRIBUTING.md#minor-fixes

rok:
Sorry @kou this is indeed not minor. @jorisvandenbossche thanks for linking to the original issue, I think this mapping is satisfactory?
I think this is also ready for review.

cpp/src/arrow/compute/kernels/scalar_temporal_unary.cc
```diff
@@ -1807,11 +1807,12 @@ const FunctionDoc is_dst_doc{
...
     {"values"}};
 
-const FunctionDoc local_time_doc{
+const FunctionDoc local_timestamp_doc{
     "Convert timestamp to a timezone-naive local time timestamp",
-    ("LocalTime converts a timestamp to a local time of timestamps timezone\n"
-     "and removes timezone metadata. If input is in UTC or doesn't have\n"
-     "timezone metadata, it is returned as is.\n"
+    ("LocalTimestamp converts timezone-aware timestamp to local timestamp\n"
+     "of the given timestamps timezone and removes timezone metadata.\n"
```
jorisvandenbossche, line 1813: "timestamps timezone" -> should that be "timestamp's timezone" ? (otherwise I don't fully understand the sentence)

Maybe we can also include "wall clock time" somewhere in the explanation?

rok: Changed and added.

westonpace:
This looks fine to me.

jorisvandenbossche (approved):

PR #34275, opened 2023-02-21
GH-34411: [Python] Change array constructor to accept pyarrow array

### LINKED ISSUE #34411: [Python] accept pyarrow Array object in the array() constructor
Subset of https://github.com/apache/arrow/issues/21761 specifically for accepting Array objects in the `pa.array(..)` function.

AlenkaF:
### Rationale for this change

Currently, `pyarrow.array` doesn't accept pyarrow Arrays and this PR adds a check to allow that.

* Closes: #34411

### DIFF
--- a/python/pyarrow/array.pxi
+++ b/python/pyarrow/array.pxi
@@ -232,6 +232,11 @@ def array(object obj, type=None, mask=None, size=None, from_pandas=None,
     else:
         c_from_pandas = from_pandas
 
+    if isinstance(obj, Array):
+        if type is not None and not obj.type.equals(type):
+            obj = obj.cast(type)
+        return obj
+
     if hasattr(obj, '__arrow_array__'):
         return _handle_arrow_array_protocol(obj, type, mask, size)
     elif _is_array_like(obj):
--- a/python/pyarrow/array.pxi
+++ b/python/pyarrow/array.pxi
@@ -234,7 +234,7 @@ def array(object obj, type=None, mask=None, size=None, from_pandas=None,
 
     if isinstance(obj, Array):
         if type is not None and not obj.type.equals(type):
-            obj = obj.cast(type, safe=safe)
+            obj = obj.cast(type, safe=safe, memory_pool=memory_pool)
         return obj
 
     if hasattr(obj, '__arrow_array__'):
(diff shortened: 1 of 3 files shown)


python/pyarrow/array.pxi
```diff
@@ -232,6 +232,11 @@ def array(object obj, type=None, mask=None, size=None, from_pandas=None,
     else:
         c_from_pandas = from_pandas
 
+    if isinstance(obj, Array):
+        if type is not None and not obj.type.equals(type):
+            obj = obj.cast(type)
```
jorisvandenbossche, line 237: We probably should pass through the `safe` and `memory_pool` keywords to `cast` (`safe` can also easily be tested)

python/pyarrow/tests/test_array.py
```diff
@@ -3381,3 +3381,15 @@ def test_struct_array_sort():
...
+
+
+def test_array_accepts_pyarrow_array():
+    arr = pa.array([1, 2, 3])
+    result = pa.array(arr)
+
+    assert arr == result
+
+    arr_uint = pa.array([1, 2, 3], type=pa.uint8())
+    result = pa.array(arr, type=pa.uint8())
```
jorisvandenbossche, line 3393: Can you add an extra test where the passed `type` is different? (eg int64 in this case)

AlenkaF: You mean adding a test where `arr` type and type passed is different? I do not think I follow - it should be the case here (`arr` is of type int64 and the test passes uint8 type to check the casting?)

I will rename `arr_uint` to `expected` just to make it clearer.

python/pyarrow/array.pxi
AlenkaF: Hm, [pc.cast](https://arrow.apache.org/docs/python/generated/pyarrow.compute.cast.html) doesn't accept the `memory_pool` keyword argument?

python/pyarrow/tests/test_array.py
jorisvandenbossche: Ah, yes, I was misreading the `arr_uint` as the array being passed to `pa.array(..)`

python/pyarrow/array.pxi
AlenkaF: To update the thread: as we discussed I will add the change to the `cast()` method to accept `memory_pool` argument as a part of this PR.

jorisvandenbossche (approved):
Thanks!

Issue #34278, opened 2023-02-21
[C++] Expose schema in named table provider


icexelloss:
### Describe the enhancement requested

Currently, the named table provider takes "names" as input:
https://github.com/apache/arrow/blob/11ae62a1426a1594ed750be1f25334fe7198356d/cpp/src/arrow/engine/substrait/options.h#L68

In some cases, it is useful for the implementer of the provider to have access to the schema as well. 
One case we have currently is that we are using named table to read from a custom remote storage and the file schema in the storage doesn't match 100% with the node schema (e.g., int32 vs int64 difference). If we have the node schema, we can add cast operation in the named table provider to avoid downstream schema/data mismatching issue.
 

### Component(s)

C++

icexelloss:
take

westonpace:
Issue resolved by pull request 34279
https://github.com/apache/arrow/pull/34279

PR #34279, opened 2023-02-21
GH-34278: [C++] Expose schema in named table provider

icexelloss:
### Rationale for this change

https://github.com/apache/arrow/issues/34278
This PR adds an parameter "schema" in the `NamedTableProvider` interface

### Are these changes tested?
Added test to check schema in named table provider

This is a developer API change.

### DIFF
--- a/cpp/src/arrow/engine/substrait/options.h
+++ b/cpp/src/arrow/engine/substrait/options.h
@@ -66,7 +66,7 @@ enum class ARROW_ENGINE_EXPORT ConversionStrictness {
 };
 
 using NamedTableProvider = std::function<Result<compute::Declaration>(
-    const std::vector<std::string>&, const std::shared_ptr<Schema>)>;
+    const std::vector<std::string>&, const Schema&)>;
 static NamedTableProvider kDefaultNamedTableProvider;
 
 using NamedTapProvider = std::function<Result<compute::Declaration>(
--- a/cpp/src/arrow/engine/substrait/options.h
+++ b/cpp/src/arrow/engine/substrait/options.h
@@ -65,8 +65,8 @@ enum class ARROW_ENGINE_EXPORT ConversionStrictness {
   BEST_EFFORT,
 };
 
-using NamedTableProvider =
-    std::function<Result<compute::Declaration>(const std::vector<std::string>&)>;
+using NamedTableProvider = std::function<Result<compute::Declaration>(
+    const std::vector<std::string>&, const std::shared_ptr<Schema>)>;
 static NamedTableProvider kDefaultNamedTableProvider;
 
 using NamedTapProvider = std::function<Result<compute::Declaration>(
--- a/cpp/src/arrow/engine/substrait/serde_test.cc
+++ b/cpp/src/arrow/engine/substrait/serde_test.cc
@@ -832,7 +832,7 @@ TEST(Substrait, ReadRel) {
 /// \brief Create a NamedTableProvider that provides `table` regardless of the name
 NamedTableProvider AlwaysProvideSameTable(std::shared_ptr<Table> table) {
   return [table = std::move(table)](const std::vector<std::string>&,
-                                    const std::shared_ptr<Schema>) {
+                                    const Schema&) {
     std::shared_ptr<compute::ExecNodeOptions> options =
         std::make_shared<compute::TableSourceNodeOptions>(table);
     return compute::Declaration("table_source", {}, options, "mock_source");
@@ -2312,11 +2312,11 @@ TEST(SubstraitRoundTrip, FilterNamedTable) {
   NamedTableProvider table_provider =
     [&input_table, &table_names, &dummy_schema](
           const std::vector<std::string>& names,
-          const std::shared_ptr<Schema> schema) -> Result<compute::Declaration> {
+          const Schema& schema) -> Result<compute::Declaration> {
     if (table_names != names) {
       return Status::Invalid("Table name mismatch");
     }
-    if (!dummy_schema->Equals(*schema)) {
+    if (!dummy_schema->Equals(schema)) {
       return Status::Invalid("Table schema mismatch");
     }
 
@@ -3057,7 +3057,7 @@ TEST(SubstraitRoundTrip, JoinRel) {
 
   NamedTableProvider table_provider = [left_table, right_table](
                                           const std::vector<std::string>& names,
-                                          const std::shared_ptr<Schema>) {
+                                          const Schema&) {
     std::shared_ptr<Table> output_table;
     for (const auto& name : names) {
       if (name == "left") {
@@ -3209,7 +3209,7 @@ TEST(SubstraitRoundTrip, JoinRelWithEmit) {
 
   NamedTableProvider table_provider = [left_table, right_table](
                                           const std::vector<std::string>& names,
-                                          const std::shared_ptr<Schema>) {
+                                          const Schema&) {
     std::shared_ptr<Table> output_table;
     for (const auto& name : names) {
       if (name == "left") {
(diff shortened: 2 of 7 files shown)


icexelloss:
cc @westonpace This should be ready for review

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -3200,21 +3207,22 @@ TEST(SubstraitRoundTrip, JoinRelWithEmit) {
...
-          }
-          if (name == "right") {
-            output_table = right_table;
-          }
-        }
-        std::shared_ptr<compute::ExecNodeOptions> options =
-            std::make_shared<compute::TableSourceNodeOptions>(std::move(output_table));
-        return compute::Declaration("table_source", {}, options, "mock_source");
-      };
+  NamedTableProvider table_provider = [left_table, right_table](
```
icexelloss, line 3210: Only interface and format change here

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -4114,21 +4123,21 @@ TEST(Substrait, SetRelationBasic) {
...
-      if (name == "T2") {
-        output_table = table2;
-      }
-    }
-    std::shared_ptr<compute::ExecNodeOptions> options =
-        std::make_shared<compute::TableSourceNodeOptions>(std::move(output_table));
-    return compute::Declaration("table_source", {}, options, "mock_source");
-  };
+  NamedTableProvider table_provider =
+      [table1, table2](const std::vector<std::string>& names, std::shared_ptr<Schema>) {
```
icexelloss, line 4127: Only interface and format change here

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -3049,21 +3055,22 @@ TEST(SubstraitRoundTrip, JoinRel) {
...
-          }
-          if (name == "right") {
-            output_table = right_table;
-          }
-        }
-        std::shared_ptr<compute::ExecNodeOptions> options =
-            std::make_shared<compute::TableSourceNodeOptions>(std::move(output_table));
-        return compute::Declaration("table_source", {}, options, "mock_source");
-      };
+  NamedTableProvider table_provider = [left_table, right_table](
```
icexelloss, line 3058: Only interface and format change here

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -2309,11 +2310,16 @@ TEST(SubstraitRoundTrip, FilterNamedTable) {
...
   NamedTableProvider table_provider =
-      [&input_table, &table_names](
-          const std::vector<std::string>& names) -> Result<compute::Declaration> {
+    [&input_table, &table_names, &dummy_schema](
+          const std::vector<std::string>& names,
+          const std::shared_ptr<Schema> schema) -> Result<compute::Declaration> {
     if (table_names != names) {
       return Status::Invalid("Table name mismatch");
     }
+    if (!dummy_schema->Equals(*schema)) {
```
icexelloss, line 2319: Added checks here to ensure schema is passed in correctly.

cpp/src/arrow/engine/substrait/options.h
```diff
@@ -65,8 +65,8 @@ enum class ARROW_ENGINE_EXPORT ConversionStrictness {
   BEST_EFFORT,
 };
 
-using NamedTableProvider =
-    std::function<Result<compute::Declaration>(const std::vector<std::string>&)>;
+using NamedTableProvider = std::function<Result<compute::Declaration>(
+    const std::vector<std::string>&, const std::shared_ptr<Schema>)>;
```
westonpace, line 69: Please use `const Schema&`.  `shared_ptr` should only be needed if the caller is likely to extend ownership.  In the rare case that is needed the caller can just make their own copy.

If you do think it is likely that caller will need to extend ownership you can use `const std::shared_ptr<Schema>&` (`const &` and not just `const`).

westonpace (changes requested):
Looks good. Just one small change.

cpp/src/arrow/engine/substrait/options.h
icexelloss: Thanks sense. Will push another revision.

icexelloss: Changed to `const Schema&`

python/pyarrow/_substrait.pyx
```diff
@@ -27,7 +27,7 @@ from pyarrow.includes.libarrow cimport *
 from pyarrow.includes.libarrow_substrait cimport *
 
 
-cdef CDeclaration _create_named_table_provider(dict named_args, const std_vector[c_string]& names):
+cdef CDeclaration _create_named_table_provider(dict named_args, const std_vector[c_string]& names, const CSchema& schema):
```
westonpace, line 30: I suppose we would eventually want to pass this schema on to the python function correct?

Something like:

```
    c_schema_sp = shared_ptr[CSchema](schema)
    py_schema = wrap_schema(c_schema_sp)

    py_table = named_args["provider"](py_names, py_schema)
```

Do you not need this for your use case?

icexelloss: Yeah the python side support for this is left out but we only use named table provider via the C++ API.

I didn't do the python side because (1) we don't use it and (2) I am not very familiar with the python C extension. (Took me a while to figure the current C extension changes just to get it working).

icexelloss: I can spend a bit time to see if I can get this working on the Python side. Looks like your example code is simple enough.

icexelloss: Hmm, any quick ideas on the error?

```
    py_names = []
    for i in range(names.size()):
        c_name = names[i]
        py_names.append(frombytes(c_name))
    c_schema_sp = shared_ptr[CSchema](schema)
                                    ^
------------------------------------------------------------

pyarrow/_substrait.pyx:42:37: Cannot convert 'shared_ptr[CSchema]' to Python object
gmake[2]: *** [CMakeFiles/_substrait_pyx.dir/build.make:71: CMakeFiles/_substrait_pyx] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:785: CMakeFiles/_substrait_pyx.dir/all] Error 2
```

icexelloss: Hmm I might be importing `wrap_schema` from the wrong place:

```
from pyarrow.lib import frombytes, tobytes, wrap_schema
```

I think it's complaining about 
```
py_schema = wrap_schema(c_schema_sp)
```
(error trace is weird on the wrong line)

icexelloss: Ok I managed to get extesion to compile with

```
c_in_schema = make_shared[CSchema](move(schema))
```

Let me fix  the python test next

westonpace: Hmm, the first error was probably that I didn't point out that you would need to add `c_schema_sp` to your `cdef` section (so cython was recognizing `c_schema_sp` as a python object and not a C++ object).

icexelloss:
@westonpace pushed another revision
* Fixed lint
* Addressed https://github.com/apache/arrow/pull/34279#discussion_r1114657183 and exposed schema via python named table provider (also added test and updated the doc)

python/pyarrow/_substrait.pyx
icexelloss: Actually - hold off - I think I didn't handle ownership correctly - should be probably doing a copy instead of move here. Let me see if I can fix.

icexelloss: Ok I pushed another revision - I think this is correct now

westonpace (approved):
Thanks for covering the bindings as well.

PR #34281, opened 2023-02-21
GH-34280: [C++][Python] Clarify meaning of row_group_size and change default to 1Mi

### LINKED ISSUE #34280: [C++][Python] Clarify meaning of "row_group_size" and change default to something more reasonable

The `row_group_size` property in `pyarrow.parquet.write_table` is described as:

> Maximum size of each written row group. If None, the row group size will be the minimum of the Table size and 64 * 1024 * 1024.

This limit is in # of rows but that is not obvious from the description.  Furthermore, 64Mi is an extremely high limit for row group size.  I believe it is perhaps based on the Java implementation, however the Java implementation treats this number as "bytes" and not "rows" (64MiB row groups is very reasonable).

Perhaps the best solution would be to add support for a limit in terms of bytes.  In the meantime, I think we should lower the default limit to 1Mi rows.

### Component(s)

C++

westonpace:
BREAKING CHANGE: Changes the default row group size when writing parquet files.
* Closes: #34280

### DIFF
--- a/cpp/src/parquet/properties.h
+++ b/cpp/src/parquet/properties.h
@@ -265,7 +265,7 @@ class PARQUET_EXPORT WriterProperties {
     }
 
     /// Specify the max number of rows to put in a single row group.
-    /// Default 64Mi rows.
+    /// Default 1Mi rows.
     Builder* max_row_group_length(int64_t max_row_group_length) {
       max_row_group_length_ = max_row_group_length;
       return this;
--- a/cpp/src/parquet/properties.h
+++ b/cpp/src/parquet/properties.h
@@ -132,7 +132,7 @@ static constexpr int64_t kDefaultDataPageSize = 1024 * 1024;
 static constexpr bool DEFAULT_IS_DICTIONARY_ENABLED = true;
 static constexpr int64_t DEFAULT_DICTIONARY_PAGE_SIZE_LIMIT = kDefaultDataPageSize;
 static constexpr int64_t DEFAULT_WRITE_BATCH_SIZE = 1024;
-static constexpr int64_t DEFAULT_MAX_ROW_GROUP_LENGTH = 64 * 1024 * 1024;
+static constexpr int64_t DEFAULT_MAX_ROW_GROUP_LENGTH = 1024 * 1024;
 static constexpr bool DEFAULT_ARE_STATISTICS_ENABLED = true;
 static constexpr int64_t DEFAULT_MAX_STATISTICS_SIZE = 4096;
 static constexpr Encoding::type DEFAULT_ENCODING = Encoding::PLAIN;
@@ -264,8 +264,8 @@ class PARQUET_EXPORT WriterProperties {
       return this;
     }
 
-    /// Specify the max row group length.
-    /// Default 64M.
+    /// Specify the max number of rows to put in a single row group.
+    /// Default 64Mi rows.
     Builder* max_row_group_length(int64_t max_row_group_length) {
       max_row_group_length_ = max_row_group_length;
       return this;
--- a/python/pyarrow/parquet/core.py
+++ b/python/pyarrow/parquet/core.py
@@ -1038,7 +1038,7 @@ Examples
         ----------
         table_or_batch : {RecordBatch, Table}
         row_group_size : int, default None
-            Maximum # of rows in each written row group. If None,
+            Maximum number of rows in each written row group. If None,
             the row group size will be the minimum of the input
             table or batch length and 1024 * 1024.
         """
@@ -1057,7 +1057,7 @@ Examples
         ----------
         batch : RecordBatch
         row_group_size : int, default None
-            Maximum # of rows in written row group. If None, the
+            Maximum number of rows in written row group. If None, the
             row group size will be the minimum of the RecordBatch
             size and 1024 * 1024.
         """
@@ -1072,8 +1072,8 @@ Examples
         ----------
         table : Table
         row_group_size : int, default None
-            Maximum # of rows in each written row group. If None, the
-            row group size will be the minimum of the Table size
+            Maximum number of rows in each written row group. If None,
+            the row group size will be the minimum of the Table size
             and 1024 * 1024.
 
         """
@@ -3153,9 +3153,9 @@ Parameters
 table : pyarrow.Table
 where : string or pyarrow.NativeFile
 row_group_size : int
-    Maximum # of rows in each written row group. If None, the
-    row group size will be the minimum of the Table size
-    and 1024 * 1024.
+    Maximum number of rows in each written row group. If None, the
+    row group size will be the minimum of the Table size and
+    1024 * 1024.
 {}
 **kwargs : optional
     Additional options for ParquetWriter
--- a/python/pyarrow/parquet/core.py
+++ b/python/pyarrow/parquet/core.py
@@ -1038,9 +1038,9 @@ Examples
         ----------
         table_or_batch : {RecordBatch, Table}
         row_group_size : int, default None
-            Maximum size of each written row group. If None, the
-            row group size will be the minimum of the input
-            table or batch length and 64 * 1024 * 1024.
+            Maximum # of rows in each written row group. If None,
+            the row group size will be the minimum of the input
+            table or batch length and 1024 * 1024.
         """
         if isinstance(table_or_batch, pa.RecordBatch):
             self.write_batch(table_or_batch, row_group_size)
@@ -1057,9 +1057,9 @@ Examples
         ----------
         batch : RecordBatch
         row_group_size : int, default None
-            Maximum size of each written row group. If None, the
+            Maximum # of rows in written row group. If None, the
             row group size will be the minimum of the RecordBatch
-            size and 64 * 1024 * 1024.
+            size and 1024 * 1024.
         """
         table = pa.Table.from_batches([batch], batch.schema)
         self.write_table(table, row_group_size)
@@ -1072,9 +1072,9 @@ Examples
         ----------
         table : Table
         row_group_size : int, default None
-            Maximum size of each written row group. If None, the
+            Maximum # of rows in each written row group. If None, the
             row group size will be the minimum of the Table size
-            and 64 * 1024 * 1024.
+            and 1024 * 1024.
 
         """
         if self.schema_changed:
@@ -3153,9 +3153,9 @@ Parameters
 table : pyarrow.Table
 where : string or pyarrow.NativeFile
 row_group_size : int
-    Maximum size of each written row group. If None, the
+    Maximum # of rows in each written row group. If None, the
     row group size will be the minimum of the Table size
-    and 64 * 1024 * 1024.
+    and 1024 * 1024.
 {}
 **kwargs : optional
     Additional options for ParquetWriter


cpp/src/parquet/properties.h
```diff
@@ -264,8 +264,8 @@ class PARQUET_EXPORT WriterProperties {
       return this;
     }
 
-    /// Specify the max row group length.
-    /// Default 64M.
+    /// Specify the max number of rows to put in a single row group.
+    /// Default 64Mi rows.
```
wjones127, line 268:
```suggestion
    /// Default 1Mi rows.
```

wjones127 (changes requested):
Cool! I know DuckDB has been complaining about this for a bit.

wjones127:
Also FYI I think DuckDB defaults to 100,000 (unless I read their source wrong earlier). In delta-rs we mimicked that with the PyArrow writer in https://github.com/delta-io/delta-rs/pull/818

python/pyarrow/parquet/core.py
```diff
@@ -1038,9 +1038,9 @@ def write(self, table_or_batch, row_group_size=None):
         ----------
         table_or_batch : {RecordBatch, Table}
         row_group_size : int, default None
-            Maximum size of each written row group. If None, the
-            row group size will be the minimum of the input
-            table or batch length and 64 * 1024 * 1024.
+            Maximum # of rows in each written row group. If None,
```
jorisvandenbossche, line 1041:
```suggestion
            Maximum number of rows in each written row group. If None,
```

(I suppose "#" is quite universal, but just using the word "number" seems even more explicit and not much longer

jorisvandenbossche:
cc @rjzamora do you have an idea if this would affect dask? I didn't find that dask exposes this keyword in the to_parquet function explicitly, but I suppose it can be passed through. But I can imagine that if most users rely on the default, and if you are then reading with `split_row_groups=True`, that might affect dask users?

westonpace:
> Also FYI I think [DuckDB defaults to 100,000](https://github.com/Mytherin/duckdb/commit/3b8ad037bff978b263fd06ec9d0635fcb049e92a#diff-a95d5e017c81184e18f0f04c5df3b72061fd80555d581a4ce163af5deca3dac0R394) (unless I read their source wrong earlier).

100k is probably workable but I don't think ideal.  100k in an int32 column would mean, at most (e.g. no encodings / compression), 400KiB per column.  If you are reading scattershot from a file (e.g. 12 out of 100 columns) then this starts to degrade performance on HDD (and probably SSD as well) and S3 (but would be fine on NVME or hot-in-memory).

That being said, no single default is going to work for all cases (100k is better for single row reads for example).  I personally think it would be more useful to make scanners that are more robust against large row groups (e.g. pyarrow's could be improved) by supporting reading at the page level instead of the row group level (still not 100% convinced this is possible).  So at the moment I'm still leaning towards 1Mi but I could be convinced otherwise.

wjones127:
> I personally think it would be more useful to make scanners that are more robust against large row groups (e.g. pyarrow's could be improved) by supporting reading at the page level instead of the row group level (still not 100% convinced this is possible). So at the moment I'm still leaning towards 1Mi but I could be convinced otherwise.

I agree with this assessment. For highly selective / single row reads, we should be optimizing the PageIndex and Page pruning paths. That work seems to be in progress:

* Read: https://github.com/apache/arrow/issues/33596
* Write: https://github.com/apache/arrow/issues/34053

cpp/src/parquet/properties.h
westonpace: Ah, good catch.  Thanks.  I've made this change.

python/pyarrow/parquet/core.py
westonpace: I agree.  I've switched to `number`.

rjzamora:
>do you have an idea if this would affect dask? I didn't find that dask exposes this keyword in the to_parquet function explicitly, but I suppose it can be passed through. But I can imagine that if most users rely on the default, and if you are then reading with split_row_groups=True, that might affect dask users?

Thanks for the ping @jorisvandenbossche !  If I understand correctly, this change is more likely to benefit Dask than it is to cause problems. When using `to_parquet(..., engine="pyarrow")`, users sometimes end up with oversized row-groups that cause memory errors during ETL. These users can already pass `row_group_size` through to mitigate this issue, but it would be nice if the default made OOM errors a bit less common.  With that said, the size of a single row completely depends on the properties of the dataset. Therefore, I don't have any strong feelings about the row-count default. If it were up to me, pyarrow would fall back to a byte-size default, rather than a row-count default (like cudf).

XinyuZeng:
There are some old discussions before: https://lists.apache.org/thread/g868kbv4jzszwrs65yv45ym1wq5ztyvq

Also, arrow-rs set the default to 1 Mi, just for reference:https://github.com/apache/arrow-rs/blob/master/parquet/src/file/properties.rs#L83

westonpace:
> There are some old discussions before

Thank you.  I thought I remembered some past discussion on this topic.  It appears the conclusion reached in the email thread was the same (the default should be 64MiB if possible, but 1Mi rows if not).

There seems to be consensus for this change.  I will merge this tomorrow.

wjones127 (approved):

westonpace:
This change had some significant performance regression for writing parquet files.  This may be unavoidable but I've opened https://github.com/apache/arrow/issues/34374 to at least investigate.

Mytherin:
Thanks for this change! This is great.

> 100k is probably workable but I don't think ideal. 100k in an int32 column would mean, at most (e.g. no encodings / compression), 400KiB per column. If you are reading scattershot from a file (e.g. 12 out of 100 columns) then this starts to degrade performance on HDD (and probably SSD as well) and S3 (but would be fine on NVME or hot-in-memory).

Agreed, we might change DuckDB's default to be larger (e.g. 500K or 1 million) as well. The problem with making it too large is that many data sets are inherently not very large, but still benefit immensely from the parallelism that multi-row groups offer. Parallelism starts to matter significantly already when dealing with a few million rows. 

> That being said, no single default is going to work for all cases (100k is better for single row reads for example). I personally think it would be more useful to make scanners that are more robust against large row groups (e.g. pyarrow's could be improved) by supporting reading at the page level instead of the row group level (still not 100% convinced this is possible). So at the moment I'm still leaning towards 1Mi but I could be convinced otherwise.

The challenge there is that you need to align the multiple columns somehow, and there is no guarantee that pages are aligned between columns. For example, you might have a column that has rows partitioned in 50K pages (`[0..50K][50K..100K]...`) and a column that has rows partitioned in 60K pages (`[0..60K][60K..120K]...`). Splitting pages over multiple threads doesn't work nicely for that reason.

The writer could guarantee that pages are aligned so that each page has the same number of rows (e.g. `120K` rows each). But from the readers' perspective you are dependent on how the pages are laid out in the file.

Another alternative is to parallelize over the number of columns instead, as each column can be read independently. That does not fit very cleanly into most execution workflows, however, and the available parallelism is entirely query-dependent (e.g. reading 2 columns means you are limited to 2 cores).

westonpace:
> The challenge there is that you need to align the multiple columns somehow, and there is no guarantee that pages are aligned between columns. For example, you might have a column that has rows partitioned in 50K pages ([0..50K][50K..100K]...) and a column that has rows partitioned in 60K pages ([0..60K][60K..120K]...). Splitting pages over multiple threads doesn't work nicely for that reason.

In my head, the way this would be addressed is in stages:

 * I/O Stage: Each column reader independently reads ahead 8-16MB worth of data pages.  This could be configurable and tweaked and could even change dynamically based on how quickly a column is being read.
 * Decoding Stage: Decoding is parallel across columns but has batch semantics.  In other words, the "batch decoder" is asked to decode a batch of X rows (e.g. 1 million).  Space could be allocated at this point for the output batch.  This then delegates to the various column decoders so each column decoder is asked to decode X items.  So maybe a column reads 5 pages to fill this batch or maybe it reads 50.

I have no idea how close Arrow's parquet-cpp reader is to this implementation or whether it is even feasible.

wgtmac:
> The writer could guarantee that pages are aligned so that each page has the same number of rows (e.g. 120K rows each). But from the readers' perspective you are dependent on how the pages are laid out in the file.

I agree with this. Aligning page boundaries benefits reading by avoiding unnecessary I/O and decoding. No special care is required on the reader side.

> I have no idea how close Arrow's parquet-cpp reader is to this implementation or whether it is even feasible.

Arrow's parquet-cpp reader currently reads pages of a single column chunk in a sequential fashion. The column reader does not know offset/length of all pages in advance. Without the knowledge of page index, it is difficult to schedule I/O and decoding in an efficient way. This is on my plan to contribute but I cannot promise the time frame yet.


According to my experience, 1 Mi rows seems small to me if the schema contains only a few columns. We'd better decide the row group by `num_of_rows` and `estimated_compressed_size` together.

python/pyarrow/parquet/core.py
```diff
@@ -3153,9 +3153,9 @@ def write_table(table, where, row_group_size=None, version='2.4',
 table : pyarrow.Table
 where : string or pyarrow.NativeFile
 row_group_size : int
-    Maximum size of each written row group. If None, the
-    row group size will be the minimum of the Table size
-    and 64 * 1024 * 1024.
+    Maximum number of rows in each written row group. If None, the
+    row group size will be the minimum of the Table size and
+    1024 * 1024.
```
mapleFU, line 3158: it's "bytes" or other?

mapleFU (commented):
The change LGTM.

I guess different storage backend can have different RowGroup size. It depends on the cost of fetching data from storage, computation parallism and the ability of skipping.

Noverdays, NVMe SSD provides high ability of IO parallism and fast random-read, rather than seek and read in HDD, SSD can have smaller default RowGroupSize

However, for these user who are using S3 or HDFS backend, they may have larger maximum-bindwidth, but they can benifits from sequential read and large RowGroup Size.

Maybe it's better to define default RowGroup length for different FS or storage backend.

mapleFU:
By the way, it's funny that, in arrow's io-merging in S3, the default range limit is 32M :)

```c++
class ARROW_EXPORT ReadRangeCache {
 public:
  static constexpr int64_t kDefaultHoleSizeLimit = 8192;
  static constexpr int64_t kDefaultRangeSizeLimit = 32 * 1024 * 1024;
```

/cc @pitrou

python/pyarrow/parquet/core.py
westonpace: It is rows.

westonpace:
> Maybe it's better to define default RowGroup length for different FS or storage backend.

Agreed.

> By the way, it's funny that, in arrow's io-merging in S3, the default range limit is 32M :)

AWS [suggests](https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance-guidelines.html) 8MB or 16MB.  32MB is probably fine.  It's still possible to get 32MiB reads.  For example, when reading multiple columns at a time.

My own testing with HDD shows that ~4MB is good enough to get sequential reads.

1Mi rows, without any encodings, will yield ~4-8MiB for a column (assuming 4 or 8 bytes per value) so in many cases I think the reads will still be large enough.

> Maybe it's better to define default RowGroup length for different FS or storage backend.

That is probably a good idea.

mapleFU:
@westonpace HDD 4MB Sequential read means, at least 4MB per row-group or column-chunk?
I think currently, we don't support split row-group by rows. How can we use 1Mi Rows to split?

PR #34288, opened 2023-02-22
GH-34122: [C++] Allow calling function registry functions without requiring a Substrait mapping

### LINKED ISSUE #34122: [C++] Use special URI to allow calling UDFs via Substrait

We should key on a special URI (e.g. https://apache.org/arrow/udf) to recognize that a Substrait call is actually looking for an Acero registered UDF.  Then we can bypass the normal mapping and just search for the function by name and pass all arguments as-is.

### Component(s)

C++

westonpace:
This introduces the special URI `urn:arrow:substrait_simple_extension_function` which can be interpreted as "just call a function from the Arrow function registry by name".  It should only be used in plans where the consumer is Acero as these plans will not likely make any sense elsewhere.
* Closes: #34122


cpp/src/arrow/engine/substrait/function_test.cc
```diff
@@ -188,6 +188,12 @@ TEST(FunctionMapping, ValidCases) {
        {int8(), int8()},
        "-119",
        int8()},
+      {{kArrowSimpleExtensionFunctionsUri, "add_checked"},
```
icexelloss, line 191: I wonder if this also works for UDFs? e.g., if I register a Python UDF by name "my_func", I can also use the same `::arrow::compute::call` to invoke it via `kSimpleSubstraitToArrow` or is there more to it?

icexelloss (approved):
Looks good to me. Have a question though.

cpp/src/arrow/engine/substrait/function_test.cc
westonpace: It should work for any scalar UDF used in a project node or filter node.  It would also work for aggregate UDFs (though these cannot be specified in python).

`::arrow::compute::call` is evaluated by looking up the function, by name, in the function registry given to the plan.  So the UDF will either need to be added to the default function registry or the custom function registry will need to be specified when the plan is run (I believe pyarrow.substrait.run_query exposes this configuration property but, if not, we can add it easily).

icexelloss: thx for the explaination

icexelloss:
@westonpace I am working on the substrait producer side of this - IIUC, the producer should generate sth like this?
```
extension_uris {
  extension_uri_anchor: 1
}
extension_uris {
  extension_uri_anchor: 2
  uri: "urn:arrow:substrait_simple_extension_function"
}
extensions {
  extension_function {
    extension_uri_reference: 1
    function_anchor: 1
    name: "multiply"
  }
}
extensions {
  extension_function {
    extension_uri_reference: 2
    function_anchor: 2
    name: "my_udf"
  }
}
```

westonpace:
> @westonpace I am working on the substrait producer side of this - IIUC, the producer should generate sth like this?

Looks correct to me. That would call a function named `my_udf` in the function registry.

icexelloss:
Thanks @westonpace I will take a follow up to add a test to use  run_query with UDFs

PR #34289, opened 2023-02-22
GH-33727: [Python] array() errors if pandas categorical column has dictionary as string not object

### LINKED ISSUE #33727: [Python] array() errors if pandas categorical column has dictionary as string not object

pandas 1.5.2
pyarrow 10.0.1

If you convert a pandas Series with dtype `string[pyarrow]` to `category`, the categories will be `string[pyarrow]`. So far, so good.
However, when you try writing the resulting object to parquet, PyArrow fails as it does not recognize its own datatype.

### Reproducer
```python
>>> import pandas as pd
>>> df = pd.DataFrame({"x": ["foo", "bar", "foo"], dtype="string[pyarrow]")
>>> df.dtypes.x
string[pyarrow]
>>> df = df.astype("category")
>>> df.dtypes.x
CategoricalDtype(categories=['bar', 'foo'], ordered=False)
>>> df.dtypes.x.categories.dtype
string[pyarrow]
>>> df.to_parquet("foo.parquet")
pyarrow.lib.ArrowInvalid: ("Could not convert <pyarrow.StringScalar: 'bar'> with type pyarrow.lib.StringScalar: did not recognize Python value type when inferring an Arrow data type", 'Conversion failed for column x with type category')
```
### Workaround
```python
df = df.astype(
    {
        k: pd.CategoricalDtype(v.categories.astype(object))
        for k, v in df.dtypes.items()
        if isinstance(v, pd.CategoricalDtype)
        and v.categories.dtype == "string[pyarrow]"
    }
)
```

### Component(s)

Python

AlenkaF:
### Rationale for this change
Currently writing a pandas dataframe with categorical column of dtype `string[pyarrow]` fails. The reason for this is that when category with `string[pyarrow]` dtype is converted to an array in pyarrow it results in a `ChunkedArray,` not `Array`, and then `DictionaryArray.from_arrays`  fails.

`_handle_arrow_array_protocol` method in _array.pxi_ is updated so that in case of a `ChunkedArray` with one chunk, the result is a `pyarrow.Array` and not `pa.ChunkedArray.`

### Are these changes tested?
Yes. Tests are added to:

- python/pyarrow/tests/parquet/test_pandas.py
- python/pyarrow/tests/test_pandas.py
- python/pyarrow/tests/test_array.py

No.
* Closes: #33727

### DIFF
--- a/python/pyarrow/array.pxi
+++ b/python/pyarrow/array.pxi
@@ -111,6 +111,8 @@ def _handle_arrow_array_protocol(obj, type, mask, size):
     if not isinstance(res, (Array, ChunkedArray)):
         raise TypeError("The object's __arrow_array__ method does not "
                         "return a pyarrow Array or ChunkedArray.")
+    if isinstance(res, ChunkedArray) and res.num_chunks==1:
+        res = res.chunk(0)
     return res
 
 
--- a/python/pyarrow/tests/parquet/test_pandas.py
+++ b/python/pyarrow/tests/parquet/test_pandas.py
@@ -26,6 +26,7 @@ from pyarrow.fs import LocalFileSystem, SubTreeFileSystem
 from pyarrow.tests.parquet.common import (
     parametrize_legacy_dataset, parametrize_legacy_dataset_not_supported)
 from pyarrow.util import guid
+from pyarrow.vendored.version import Version
 
 try:
     import pyarrow.parquet as pq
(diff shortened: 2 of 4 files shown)


jorisvandenbossche (approved):
This seems a good short term fix for this specific case. When pandas starts to use pyarrow chunked arrays more and more, we might need to think about what `pa.array(pandas_series)` should do if the pandas object has multiple chunks, but let's leave that for another issue.

Just one small remark on the tests

jorisvandenbossche:
Thanks! Opened https://github.com/apache/arrow/issues/34755 for the broader question about how to handle chunked arrays in `pa.array()`

Issue #25025, opened 2020-05-22
[C++] Split non-cast compute kernels into a separate shared library


asfimport:
Since we are going to implement a lot more precompiled kernels, I am not sure it makes sense to require all of them to be compiled unconditionally just to get access to `compute::Cast`, which is needed in many different contexts.

After ARROW-8792 is merged, I would suggest creating a plugin hook for adding a bundle of kernels from a shared library outside of libarrow.so, and then moving all the object code outside of Cast to something like libarrow_compute.so. Then we can change the CMake flags to compile Cast kernels always (?) and then opt in to building the additional kernels package separately

**Reporter**: [Wes McKinney](https://issues.apache.org/jira/browse/ARROW-8891) / @wesm
**Watchers**: [Rok Mihevc](https://issues.apache.org/jira/browse/ARROW-8891) / @rok
#### Related issues:
- [[C++] C++ array kernels framework and execution buildout (umbrella issue)](https://github.com/apache/arrow/issues/25028) (relates to)
- [[Python] Allow disabling more components](https://github.com/apache/arrow/issues/33126) (relates to)

<sub>**Note**: *This issue was originally created as [ARROW-8891](https://issues.apache.org/jira/browse/ARROW-8891). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

benibus:
@lidavidm @westonpace @zeroshade @felipecrv
I've been working on the PR for this - just wanted to give an update and get some design opinions.

Currently, I've set things up in the build system so that everything outside of `compute/kernels` is built into libarrow unconditionally - along with:
- `scalar_cast_*.cc`
- `vector_selection.cc` for `take` (used by dictionary casts, parquet)
- `vector_hash.cc` for `unique` (used by parquet)

If `ARROW_COMPUTE` is enabled, then the remaining kernel sources are built into libarrow_compute, which links against libarrow. Alternatively, we could introduce a new flag for this, but as it stands, `-DARROW_COMPUTE=ON` still gives you all the kernels (and we could unconditionally compile code that uses casts - i.e. the CSV writer, STL tests, etc).

Assuming that sounds reasonable (if not, let me know), I just need to determine how registration will work for the extra kernels. Currently, this is done on the first invocation of `compute::GetFunctionRegistry` in libarrow.so. However, it'll no longer be possible to register the extra kernels in this way given that their registration functions are in a different object that libarrow doesn't link against. As I see it, there are a couple possibilities:
[...]

PR #34295, opened 2023-02-22
GH-34388: [C++] Build core compute kernels unconditionally

### LINKED ISSUE #34388: [C++] Build core compute kernels unconditionally

From https://github.com/apache/arrow/issues/25025
> Since we are going to implement a lot more precompiled kernels, I am not sure it makes sense to require all of them to be compiled unconditionally just to get access to compute::Cast, which is needed in many different contexts.

This would entail building a small subset of kernels into libarrow by default. The remaining kernels could be conditionally included via cmake options.

### Component(s)

C++

benibus:
This includes the core compute machinery in libarrow by default - in addition to all cast kernels and several other kernels that are either dependencies of `cast` (`take`) or utilized in libarrow/libparquet (`unique`, `filter`). The remaining kernels won't be built/registered unless `ARROW_COMPUTE=ON` (note that this would slightly change the option's meaning, as currently, nothing in arrow/compute is built unless it's set).

Initially this was more substantial as the original goal was to build the extra kernels as a shared library (suggested in the orginal issue). After some discussion in the issue thread, I opted not to do that - primarily because I can't personally see the utility of a separate lib here, even ignoring the complexity it introduces. However, there may be a good reason that simply hasn't occured to me.
* Closes: #34388

### DIFF
--- a/cpp/src/arrow/compute/CMakeLists.txt
+++ b/cpp/src/arrow/compute/CMakeLists.txt
@@ -26,8 +26,22 @@ arrow_add_pkg_config("arrow-compute")
 # Unit tests
 #
 
+# As of 2023-02-28, the following kernels are always present:
+#  - array_filter
+#  - array_take
+#  - cast
+#  - dictionary_encode
+#  - drop_null
+#  - filter
+#  - indices_nonzero
+#  - take
+#  - unique
+#  - value_counts
+#
+# Tests that use additional kernels should specify REQUIRE_ALL_KERNELS to avoid
+# being included in minimal builds. See: GH-34388
 function(ADD_ARROW_COMPUTE_TEST REL_TEST_NAME)
-  set(options)
+  set(options REQUIRE_ALL_KERNELS)
   set(one_value_args PREFIX)
   set(multi_value_args LABELS)
   cmake_parse_arguments(ARG
@@ -35,6 +49,10 @@ function(ADD_ARROW_COMPUTE_TEST REL_TEST_NAME)
                         "${one_value_args}"
                         "${multi_value_args}"
                         ${ARGN})
+
+  if(ARG_REQUIRE_ALL_KERNELS AND (NOT ARROW_COMPUTE))
+    return()
+  endif()
 
   if(ARG_PREFIX)
     set(PREFIX ${ARG_PREFIX})
--- a/cpp/src/arrow/compute/CMakeLists.txt
+++ b/cpp/src/arrow/compute/CMakeLists.txt
@@ -26,7 +26,7 @@ arrow_add_pkg_config("arrow-compute")
 # Unit tests
 #
 
-# As of 2023-02-28, the following kernels are always present:
+# The following kernels are always present:
 #  - array_filter
 #  - array_take
 #  - cast
--- a/cpp/src/arrow/compute/kernels/CMakeLists.txt
+++ b/cpp/src/arrow/compute/kernels/CMakeLists.txt
@@ -18,10 +18,11 @@
 # ----------------------------------------------------------------------
 # Scalar kernels
 
+add_arrow_compute_test(scalar_cast_test SOURCES scalar_cast_test.cc test_util.cc)
+
 add_arrow_compute_test(scalar_type_test
                        SOURCES
                        scalar_boolean_test.cc
-                       scalar_cast_test.cc
                        scalar_nested_test.cc
                        scalar_string_test.cc
                        test_util.cc)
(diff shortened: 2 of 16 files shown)


benibus:
One thing I haven't decided is how to deal with the compute unit tests since most of them make heavy use of the extra kernels, so a good chunk of them will fail without them. Easiest option would be force `ARROW_COMPUTE=ON` if `ARROW_BUILD_TESTS=ON` (not the worst idea in the world, i guess - as this is a packaging-focused feature). Alternatively, we could just not build the tests in question - although that would include most of the tests in `compute/exec`.

(Also, I'll look into the unity build failures - not quite sure what's going wrong there...)

benibus:
cc @felipecrv

felipecrv (approved):

felipecrv:
This isn't enough to close the issue, right? Do you want to associate this one with a sub-issue of #25025 so you can merge it before working on the shared library setup?

benibus:
Well to be fair, that might still be the right move as it'd be easy to make that a follow-up PR if we decide to go down that road (assuming we get the library boundaries right in this one).

westonpace:
The goal is to eventually reduce the size of the "core" library right?  Do we have any idea how slim this set of baseline kernels is compared to the full set?

cpp/src/arrow/compute/kernels/CMakeLists.txt
```diff
@@ -18,10 +18,11 @@
 # ----------------------------------------------------------------------
 # Scalar kernels
 
+add_arrow_compute_test(scalar_cast_test SOURCES scalar_cast_test.cc test_util.cc)
```
westonpace, line 21: Why did you peel this off into its own test?

westonpace (commented):
Seems like a reasonable set of minimal kernels.  Do you want to add a CI job (or does one already exist?) that builds without ARROW_COMPUTE to ensure basic functionality (e.g. parquet reading/writing and csv reading/writing) still works?

benibus:
> Do we have any idea how slim this set of baseline kernels is compared to the full set?

Here's what would be present in the default build:
```
array_filter
array_take
cast
dictionary_encode
drop_null
filter
indices_nonzero
take
unique
value_counts
```
There are currently 240 kernels in the full set, so it's a pretty deep cut.

> Do you want to add a CI job (or does one already exist?) that builds without ARROW_COMPUTE to ensure basic functionality (e.g. parquet reading/writing and csv reading/writing) still works?

That would be a good idea, yes. AFAIK none of the existing jobs build without ARROW_COMPUTE. Even if they did, the CSV writer/STL tests wouldn't be included and libparquet wouldn't be built at all.

cpp/src/arrow/compute/kernels/CMakeLists.txt
benibus: It was so that the casts could be tested in isolation with/without the complete registry, given that none of the other scalar-type kernels are present in the base set. Also, the cast kernels seemed logically distinct enough to warrant a separate test IMO.

The testing situation isn't exactly ideal at the moment... The following will fail if ARROW_COMPUTE isn't enabled:
```
arrow-compute-scalar-type-test
arrow-compute-scalar-if-else-test
arrow-compute-scalar-temporal-test
arrow-compute-scalar-math-test
arrow-compute-scalar-utility-test
arrow-compute-vector-test
arrow-compute-aggregate-test
arrow-compute-expression-test
arrow-compute-plan-test
arrow-compute-hash-join-node-test
arrow-compute-asof-join-node-test
arrow-compute-groupby-test
```
I went into a bit more detail in an [earlier comment](https://github.com/apache/arrow/pull/34295#issuecomment-1441138674).

westonpace: Ah, ok.  That makes sense.

westonpace:
This failure looks related: https://github.com/apache/arrow/actions/runs/4247620924/jobs/7385865305

Perhaps just a changing of include orders has angered the unity build gremlins in some way.  Given the two implementations are identical maybe we could put them in util_internal.h?

benibus:
I still need to add the CI job, but in preparation, I set things up so that certain tests won't be built without the complete kernel registry - so we wouldn't need any special ctest flags to avoid  expected failures.

The unity build redefinition errors should be fixed now. Most of the problematic code in scalar_round.cc was actually completely unused, so I just removed it.

cpp/src/arrow/compute/CMakeLists.txt
```diff
@@ -26,8 +26,22 @@ arrow_add_pkg_config("arrow-compute")
 # Unit tests
 #
 
+# As of 2023-02-28, the following kernels are always present:
```
westonpace, line 29:
```suggestion
# The following kernels are always present:
```
I'm not sure the date is useful (and it will be hard to keep correct).  Users can always use git blame if they need it.

westonpace (approved):

westonpace:
@felipecrv / @lidavidm I haven't been following the discussion on #25025 very closely.  However, this change seems good.  I assume we want to proceed with it?

lidavidm:
Yes, either way, I think this is a necessary first step before we can unbundle the kernels + I hope it is easier to review this way

PR #34312, opened 2023-02-23
GH-34293: [Java] Error loading native libraries on Windows

### LINKED ISSUE #34293: [Java] Error loading native libraries on Windows

`org.apache.arrow.dataset.jni.JniLoader` cannot load "x86_64\arrow_dataset_jni.dll" from JAR file("arrow-dataset-11.0.0.jar" from Maven central repository) and throws an exception. However, this example works fine on Arrow 10.0.1.

Stack trace

```
java.lang.IllegalStateException: error loading native libraries: java.io.FileNotFoundException: x86_64\arrow_dataset_jni.dll
	at org.apache.arrow.dataset.jni.JniLoader.load(JniLoader.java:94)
	at org.apache.arrow.dataset.jni.JniLoader.loadRemaining(JniLoader.java:74)
	at org.apache.arrow.dataset.jni.JniLoader.ensureLoaded(JniLoader.java:61)
	at org.apache.arrow.dataset.jni.NativeMemoryPool.getDefault(NativeMemoryPool.java:34)
	at QuickStart.main(QuickStart.java:23)
```

QuickStart.java:

```java
// Reading a Parquet file using FileSystemDatasetFactory
import org.apache.arrow.dataset.file.FileFormat;
import org.apache.arrow.dataset.file.FileSystemDatasetFactory;
import org.apache.arrow.dataset.jni.NativeMemoryPool;
import org.apache.arrow.dataset.scanner.ScanOptions;
import org.apache.arrow.dataset.scanner.Scanner;
import org.apache.arrow.dataset.source.Dataset;
import org.apache.arrow.dataset.source.DatasetFactory;
import org.apache.arrow.memory.BufferAllocator;
import org.apache.arrow.memory.RootAllocator;
import org.apache.arrow.vector.VectorSchemaRoot;
import org.apache.arrow.vector.ipc.ArrowReader;

import java.nio.file.Path;

public class QuickStart {
    public static void main(String[] args) {
        String uri = Path.of("data1.parquet").toUri().toString();
        System.out.println(uri);
        ScanOptions options = new ScanOptions(32768);
        try (
                BufferAllocator allocator = new RootAllocator();
                DatasetFactory datasetFactory = new FileSystemDatasetFactory(allocator, NativeMemoryPool.getDefault(), FileFormat.PARQUET, uri);
                Dataset dataset = datasetFactory.finish();
                Scanner scanner = dataset.newScan(options);
                ArrowReader reader = scanner.scanBatches()
        ) {
            while (reader.loadNextBatch()) {
                try (VectorSchemaRoot root = reader.getVectorSchemaRoot()) {
                    System.out.print(root.contentToTSVString());
                }
            }
        } catch (Exception e) {
            e.printStackTrace();
        }
    }
}

```

Version: arrow 11.0.0

Dependencies:
- org.apache.arrow:arrow-vector:11.0.0
- org.apache.arrow:arrow-memory:11.0.0
- org.apache.arrow:arrow-memory-netty:11.0.0
- org.apache.arrow:arrow-dataset:11.0.0

Platform:

- OS: Windows 11 x64
- JDK version: 17

### Component(s)

Java

davisusanibar:
To fix error on Java native libraries related to `File.separator` on Windows OS: Use `/` to separate path instead of `\`.
* Closes: apache/arrow#34293
* GitHub Issue: #34293


davisusanibar:

lidavidm (commented):
Are there tests we can add to cover this?

rtadepalli:
I don't think we should remove `File.separator`. Instead, could we try using the [`file(TO_CMAKE_PATH)`](https://cmake.org/cmake/help/latest/command/file.html) class in `CMakeLists.txt` to generate the DLL file name appropriately using `\` as a separator if the platform is Windows?

java/dataset/src/test/java/org/apache/arrow/dataset/file/TestFileSystemDataset.java
```diff
@@ -386,6 +388,27 @@ public void testBaseCsvRead() throws Exception {
...
+      Assertions.assertThrows(FileNotFoundException.class, () -> {
+        try (final InputStream stream = TestFileSystemDataset.class.getResourceAsStream(pathWithFileSeparator)) {
+          if (stream == null) {
+            throw new FileNotFoundException(pathWithFileSeparator);
+          }
+        }
+      }, "FileNotFound \\avroschema\\user.avsc");
+    }
+  }
+
```
lidavidm, line 411: I don't see why we're testing the Java standard library here. My previous review was asking why this issue was not caught by any test at all on Windows: it implies that we don't test any of this JNI code on Windows, and that is the real issue I want to fix.

lidavidm (changes requested):

java/dataset/src/test/java/org/apache/arrow/dataset/file/TestFileSystemDataset.java
davisusanibar: The current JUnit tests are able to read resources without any issues.

Basically, this error was not detected because of the lack of checking of `getResourceAsStream` in the main code to load native libraries.

davisusanibar: `File.separator` cause the problem on Windows OS by [ClassLoader](https://docs.oracle.com/javase/8/docs/api/java/lang/ClassLoader.html#getResource-java.lang.String-:~:text=The%20name%20of%20a%20resource%20is%20a%20%27/%27%2Dseparated%20path%20name%20that%20identifies%20the%20resource.) java docs.

lidavidm: I don't see how this test changes that. And I don't see what you mean: it seems this should have failed and been caught if any of the tests ran on Windows at all. So the main problem is to make sure we test JNI code on Windows somewhere.

lidavidm: Sure, though without that, it's hard to validate that we've fixed things here

davisusanibar:

davisusanibar:
With version: 11.0.0
```
java.lang.IllegalStateException: error loading native libraries: java.io.FileNotFoundException: x86_64\arrow_dataset_jni.dll
	at org.apache.arrow.dataset.jni.JniLoader.load(JniLoader.java:94)
	at org.apache.arrow.dataset.jni.JniLoader.loadRemaining(JniLoader.java:74)
	at org.apache.arrow.dataset.jni.JniLoader.ensureLoaded(JniLoader.java:61)
	at org.apache.arrow.dataset.jni.NativeMemoryPool.getDefault(NativeMemoryPool.java:34)
	at me.Recipe.main(Recipe.java:25)
```

With new version arrow-dataset-12.0.0-SNAPSHOT: https://github.com/ursacomputing/crossbow/releases/tag/actions-b5ef0c7d3c-github-java-jars
```
- wget https://github.com/ursacomputing/crossbow/releases/download/actions-b5ef0c7d3c-github-java-jars/arrow-dataset-12.0.0-SNAPSHOT.jar
- mvn install:install-file "-Dfile=C:\Users\dsusanibar\jni\arrow-dataset-12.0.0-SNAPSHOT.jar" "-DgroupId=org.apache.arrow" "-DartifactId=arrow-dataset" "-Dversion=12.0.0-SNAPSHOT" "-Dpackaging=jar"
```

Cookbook: https://arrow.apache.org/cookbook/java/dataset.html#query-data-content-for-file
```
import org.apache.arrow.dataset.file.FileFormat;
import org.apache.arrow.dataset.file.FileSystemDatasetFactory;
import org.apache.arrow.dataset.jni.NativeMemoryPool;
import org.apache.arrow.dataset.scanner.ScanOptions;
import org.apache.arrow.dataset.scanner.Scanner;
import org.apache.arrow.dataset.source.Dataset;
import org.apache.arrow.dataset.source.DatasetFactory;
import org.apache.arrow.memory.BufferAllocator;
import org.apache.arrow.memory.RootAllocator;
import org.apache.arrow.vector.VectorSchemaRoot;
import org.apache.arrow.vector.ipc.ArrowReader;

public class Recipe {
    public static void main(String[] args) {
        String uri = "file:///C:\\Users\\dsusanibar\\IdeaProjects\\win-cookbooks\\src\\main\\resources\\files\\data1.parquet";
        ScanOptions options = new ScanOptions(/*batchSize*/ 32768);
        try (
                BufferAllocator allocator = new RootAllocator();
                DatasetFactory datasetFactory = new FileSystemDatasetFactory(allocator, NativeMemoryPool.getDefault(), FileFormat.PARQUET, uri);
                Dataset dataset = datasetFactory.finish();
                Scanner scanner = dataset.newScan(options);
                ArrowReader reader = scanner.scanBatches()
        ) {
            while (reader.loadNextBatch()) {
                try (VectorSchemaRoot root = reader.getVectorSchemaRoot()) {
                    System.out.print(root.contentToTSVString());
                }
            }
        } catch (Exception e) {
            e.printStackTrace();
        }
    }
}
```

davisusanibar:
Just filled issue for CI Java JNI modules https://github.com/apache/arrow/issues/34763

java/dataset/src/test/java/org/apache/arrow/dataset/file/TestFileSystemDataset.java
lidavidm: This test is still not useful

davisusanibar:
Hi @kou could you help me on this PR?

`File.separator` cause the problem on Windows OS by [ClassLoader](https://docs.oracle.com/javase/8/docs/api/java/lang/ClassLoader.html#getResource-java.lang.String-:~:text=The%20name%20of%20a%20resource%20is%20a%20%27/%27%2Dseparated%20path%20name%20that%20identifies%20the%20resource.) java docs, it was introduced on Arrow 11 release and this is not able to resolve for Windows OS.

rtadepalli:
Is there an issue with the CMake approach I pointed out above? The .dll generation pipeline is in the repo, and we could make CMake use a different separator for generating DLLs. No code would need to be changed, unless this is the preferred way.

davisusanibar:
> Is there an issue with the CMake approach I pointed out above? The .dll generation pipeline is in the repo, and we could make CMake use a different separator for generating DLLs. No code would need to be changed, unless this is the preferred way.

Yes, this could solve some scenarios. In this case this is needed at runtime to inspect native libraries inside the jar, there are no another way that maintain this code inside java classes. Do you agree?

rtadepalli:
My apologies if I am misunderstanding something here -- I am thinking that the issue is that there is a DLL file named `x86_64\arrow_dataset_jni.dll` and loading this fails with a `FileNotFoundException` because the actual DLL is named `x86_64/arrow_dataset_jni.dll`. 

If this is right, would changing the name of the generated DLL file and using `File.separator` not work? I am a bit unclear on what you mean by "there are no another way that maintain this code inside java classes" -- do you please mind explaining? Do you mean to say that even if we make the CMake change, we would still need to make changes in the application because loading the JAR would fail?

davisusanibar:
> My apologies if I am misunderstanding something here -- I am thinking that the issue is that there is a DLL file named `x86_64\arrow_dataset_jni.dll` and loading this fails with a `FileNotFoundException` because the actual DLL is named `x86_64/arrow_dataset_jni.dll`.
> 
> If this is right, would changing the name of the generated DLL file and using `File.separator` not work? I am a bit unclear on what you mean by "there are no another way that maintain this code inside java classes" -- do you please mind explaining? Do you mean to say that even if we make the CMake change, we would still need to make changes in the application because loading the JAR would fail?

1. CMake did the work without problems: Create folder `x86_64` and put the native library inside that `arrow_dataset_jni.dll`
2. Next step is Java time, Java at runtime need to localize and load that native library specified by the filename argument.
3. `File.separator` (Windows) is translated/converted to `\` (that generate problems), and File.separator (Unix's) is translated/converted to `/` (working without problems)
4. BUT Java independently of the OS mention [getResource](https://docs.oracle.com/javase/8/docs/api/java/lang/ClassLoader.html#getResource-java.lang.String-:~:text=The%20name%20of%20a%20resource%20is%20a%20%27/%27%2Dseparated%20path%20name%20that%20identifies%20the%20resource.):  `The name of a resource is a '/'-separated path name that identifies the resource.` 

For that reason in case is needed to identified a resource (as the DLL) is needed to use `/` instead of `File.separator` independently of the OS.

rtadepalli:
Oh ok, interesting. Thank you for the explanation. I was unaware that loading a resource requires `/`, and the native filesystem path separator does not apply. It all makes sense now, and changing CMakeLists.txt to produce `x86_64\arrow_dataset_jni.dll` would not do much since we're not operating with the file system abstraction, but rather with the resource abstraction.

kou:
@davisusanibar Sure! What failure should I fix?

kou:
I see. Then can we work on #34763 before this?

I have a local Windows environment but it doesn't have Java. So it's helpful for me that we can try the problem on GitHub Actions.

Do you want to work on #34763?
Or do you want me to work on #34763?

davisusanibar:
> I see. Then can we work on #34763 before this?

Current Dataset 11.0.0 is not working on Windows OS, this PR is to be able to work this on next release 12.0.0. We could work on #34763 as an improvement on the next releases.

> 
> I have a local Windows environment but it doesn't have Java. So it's helpful for me that we can try the problem on GitHub Actions.
> 
> Do you want to work on #34763? Or do you want me to work on #34763?

I'll work on that, but not for release 12.

kou:
I see but I can't help you without #34763...

ssvendsen:
Any chance this this could be included in version 12 independently of #34763? Windows support is broken in version 11 anyway.

lidavidm:
I think feature freeze has already passed, unfortunately - this could possibly have been included but I still want to at least remove the useless test (if we're going to defer proper testing for a different issue)

davisusanibar:
Just close this and consider to be part of https://github.com/apache/arrow/issues/34763

lidavidm:
I don't see why we have to close this?

FWIW, I never merged this because there is unaddressed feedback.

davisusanibar:
> I don't see why we have to close this?
> 
> FWIW, I never merged this because there is unaddressed feedback.

I apologize for the long time ago.

Would it be okay to remove this test `TestFileSystemDataset.testResourceToLoad` to address the last feedback?

lidavidm:
Yes.

lidavidm (approved):

PR #34320, opened 2023-02-24
MINOR:[CI] Fix automatic issue labeling.

assignUser:
### Rationale for this change

The issue labeling was failing because it only got the first 30 label (we have 56) everything on page 2 was regarded as "non-existent' throwing an error.
  
Raise the number of items per page to 100 ensuring that all labels are fetched.


assignUser:
works: https://github.com/apache/arrow/issues/34321

PR #34331, opened 2023-02-24
GH-34332: [Go][FlightRPC] Add driver for `database/sql` framework

### LINKED ISSUE #34332: [Go][FlightRPC] Add driver for `database/sql` framework

Due to it's wide adoption and ease-of-use, a driver for Golang's `database/sql` framework will help new users and applications using different data-sources.

Add a driver implementation for FlighSQL to allow the use.

### Component(s)

FlightRPC, Go

srebhan:
### Rationale for this change

Using Golang's `database/sql` framework is well known, offers goodies like connection pooling and is easy to use. Therefore using FlightSQL trough this framework is a good starting point for users performing simple queries, inserts etc.

This PR adds an `database/sql/driver` implementation currently supporting `sqlite` and `InfluxData IOx` (query only). Unit-tests are added using the SQLite server example implementation and the driver and driver settings are documented.

### Are these changes tested?

Yes, a test-suite is added for the driver. Futhermore, the IOx backend is additionally tested against a real local instance using [this code](https://github.com/srebhan/go-flightsql-example).

This PR does not contain breaking changes. All modifications to the FlightSQL client code are transparent to the user.
* Closes: #34332


lidavidm:
How does this differ from adbc/sqldriver? https://github.com/apache/arrow-adbc/blob/main/go/adbc/sqldriver/driver.go

zeroshade:
@srebhan Would it make more sense to add IOx backend testing to the CI for adbc/sqldriver (as pointed to by @lidavidm) along with any necessary changes there?

At the moment we're trying to keep the arrow module/package compatible with go1.17+ so at a minimum you'd need to add `go:build` constraints to not break the CI here. But it would probably be better to consolidate efforts in the ADBC sqldriver (which would provide a `database/sql` wrapper for *anything* that implements ADBC including the FlightSQL driver) rather than duplicating a large part of that work here.

What do you think?

srebhan:
@zeroshade thanks for your comment. I can surely work on making this compatible with Go 1.17 and avoid breaking tests by removing the IOx backend tests (which only supports a subset of the operations anyway). I thought you were running CI tests in "short" mode to strip out integration tests...

Regarding the relation to the ADBC driver: I'm aware of this driver, but I do want the leanest solution possible adhering to the look-and-feel of `database/sql` as close as possible. For _me personally_ there is no value in a wrapper around a wrapper. I even thought of getting rid of the different backend variants by putting in the specific options as parameters to make this a thin wrapper around the FlightSQL client. Does that make sense?

zeroshade:
@srebhan from the arrow side it's more of a maintenance question. it's hard to justify maintaining two separate implementations. The ADBC FlightSQL driver is basically a thin wrapper around the FlightSQL client already and the `adbc/sqldriver` package shouldn't be adding much in the way of overhead (honestly I'd be surprised if any overhead it adds dwarfs the actual network IO costs....). At a minimum we should document the use case, either through your work here or by documenting how to leverage ADBC's flight sql driver. 

Just as a thought, would you be able to create a benchmark that uses `database/sql` against IOx that we can run with both of your solution here and the ADBC driver so we can compare and see if there's a significant difference (thus justifying the separate implementation)?

go/arrow/flight/flightsql/driver.go
```diff
@@ -0,0 +1,772 @@
...
+			return nil, fmt.Errorf("reading record failed: %w", err)
+		}
+
+		if rows.schema == nil {
+			rows.schema = record.Schema()
+		}
+		if !rows.schema.Equal(record.Schema()) {
+			return nil, fmt.Errorf("mixed schemas %w", ErrNotSupported)
+		}
+		rows.records = append(rows.records, record)
```
zeroshade, line 256: This is only grabbing the first record, you need to either do:

```go
for reader.Next() {
    record := reader.Record()
    record.Retain()
    rows.records = append(rows.records, record)
}
if reader.Err() {
    // do something with error
}
```

or you need to call `reader.Read()` repeatedly until you get a non-nil `err`. `io.EOF` indicates no more records, and anything else would be an error it reports.

go/arrow/flight/flightsql/driver.go
```diff
@@ -0,0 +1,772 @@
...
+			return fmt.Errorf("schema: %w", err)
+		}
+		fields = append(fields, arrow.Field{
+			Name:     arg.Name,
+			Type:     dt,
+			Nullable: true,
+		})
+	}
+
+	schema := s.stmt.ParameterSchema()
```
zeroshade, line 285: if your `toArrowDataType` auto-conversion checks don't match the types in the ParameterSchema, you could end up with a panic or other failure. If `s.stmt.ParameterSchema()` is non-nil, you should use that for converting the arguments rather than inferring the data types.

go/arrow/flight/flightsql/driver.go
```diff
@@ -0,0 +1,772 @@
...
+
+func (d *Driver) BeginTx(ctx context.Context, opts sql.TxOptions) (driver.Tx, error) {
+	tx, err := d.client.BeginTransaction(ctx)
+	if err != nil {
+		return nil, err
+	}
+	d.txn = tx
+
+	return &Tx{tx: tx, timeout: d.timeout}, nil
+}
```
zeroshade, line 488: storing the transaction at the driver level means that this will break if there are multiple connections with their own transactions. You need a separate connection object and to store the transaction at the connection level instead, not the driver level.

go/arrow/flight/flightsql/driver.go
```diff
@@ -0,0 +1,772 @@
...
+
+type DriverConfig struct {
+	Backend  string
+	Address  string
+	Username string
+	Password string
+	Database string
+	Token    string
+	Timeout  time.Duration
+}
```
zeroshade, line 709: Also need to add configuration options for TLS and such.

zeroshade (changes requested):
Separate from the discussion as to whether or not we should just leverage the ADBC `adbc/sqldriver` package instead, I did a quick review on the code and left a bunch of comments.

If we end up deciding to go ahead with this then I'll give a more thorough review.

jacobmarble:
> @srebhan from the arrow side it's more of a maintenance question. it's hard to justify maintaining two separate implementations. The ADBC FlightSQL driver is basically a thin wrapper around the FlightSQL client already and the `adbc/sqldriver` package shouldn't be adding much in the way of overhead (honestly I'd be surprised if any overhead it adds dwarfs the actual network IO costs....). At a minimum we should document the use case, either through your work here or by documenting how to leverage ADBC's flight sql driver.
> 
> Just as a thought, would you be able to create a benchmark that uses `database/sql` against IOx that we can run with both of your solution here and the ADBC driver so we can compare and see if there's a significant difference (thus justifying the separate implementation)?

I'm not sure performance is the concern (@srebhan please correct or confirm). IMO, this driver is valuable because:
1. [one conventional golang import statement](https://go.dev/doc/database/open-handle#database_driver)

```golang
import _ "github.com/apache/arrow/go/v12/arrow/flight/flightsql"
...
sql.Open("flightsql", dsn)
```
vs
```golang
import "github.com/apache/arrow-adbc/go/adbc/driver/flightsql"
import "github.com/apache/arrow-adbc/go/adbc/sqldriver"
sql.Register("flightsql", sqldriver.Driver{flightsql.Driver{}})
...
sql.Open("flightsql", dsn)
```

2. easy to discover by googling `FlightSQL golang database/sql`
3. fewer dependencies

All of the above are satisfied by this PR, and not by the ADBC wrapper.

---

On the other hand, this PR means the Go/Arrow community would maintain two FlightSQL `database/sql` drivers, [which has been discussed before](https://github.com/apache/arrow-adbc/issues/315#issuecomment-1372695586).

Here's an alternative that satisfies 1 and 2, but not 3 (which hardly matters these days), and retains a the implementation:

New module (including go.mod) containing only:
```golang
import (
	"github.com/apache/arrow-adbc/go/adbc/driver/flightsql"
	"github.com/apache/arrow-adbc/go/adbc/sqldriver"
)

func init() {
	sql.Register("flightsql", sqldriver.Driver{
		Driver: flightsql.Driver{},
	})
}
```

The module could live as `github.com/apache/arrow/go/v12/arrow/flight/flightsql/sqldriver`, or it could live in the ADBC repo and get a vanity name like `go.arrow.apache.org/flightsql/sqldriver`.

zeroshade:
@jacobmarble I like the idea of having a separate module solely consisting of registering the flightsql driver + a README with the options / how-to-use documentation etc.

I recommend putting it in the ADBC repo solely because in the past I've had difficulties with a submodule of the Arrow Go module due to the `v12` major version in the import path. Could you put up a PR for this on the ADBC repo and myself and @lidavidm will take a look at it as soon as we can to move this forward.

Thanks much!

srebhan:
@jacobmarble it's not only about wrapping the registration, it's about the fundamental addition of (IMO unnecessary) complexity. The ADBC `database/sql` driver is a wrapper around `ADBC` as such it adds another indirection from the database, i.e. `Application -> database/sql -> ADBC -> FlightSQL -> Backend` vs. `Application -> database/sql -> FlightSQL -> Backend`. It's like writing a wrapper around ODBC... What is the benefit, given that `database/sql` is the defacto standard in go? The clear drawback for the user is to deal with ADBC. From the docs it's not yet clear to me which "ADBC drivers" I need. Are they bundled with the driver? If so, how to deal with growing dependencies and code when adding new drivers? What if CGO is not allowed in the application?

I'm certainly willing to work on this PR, but it sounds like there is some resistance to the idea. Please let me know if there is a realistic chance of getting this merged!

lidavidm:
You depend on the Flight SQL driver to use Flight SQL. There is no cgo required. The cgo bits are only if you want to 1) use a non-Go ADBC driver or 2) expose a Go driver to a non-Go application. The proposed module would mean that the user just imports a single module that re-exports the interface, without explicitly worrying about ADBC.

We can add documentation as necessary for searchability (and IMO, having two implementations would make searchability worse, not better). I don't think the answer to "there is no documentation" is "write a second implementation".

Again, the primary concern is maintenance. Are you/InfluxData willing to continue maintaining the implementation here? My other concern is just confusing users by having two implementations that don't functionally differ.

I can see the argument for a separate implementation if there is a specific benefit for users, but I don't think that's been addressed beyond having fewer layers in principle. 

See what was done for Python: instead of maintaining specific Flight SQL<->DBAPI wrappers, we did the same thing where there is just an ADBC<->DBAPI wrapper. It also means we can iterate on those parts separately from/more quickly than the core library.

jacobmarble:
Here's my alternative proposal:
https://github.com/apache/arrow-adbc/pull/480

As documented in the PR example, use is:
```golang
import (
	"database/sql"
	_ "github.com/apache/arrow-adbc/go/adbc/sqldriver/flightsql"
)

func Example() {
	db, err := sql.Open("flightsql", "uri=grpc://localhost:12345")
	if err != nil {
		panic(err)
	}

	_ = db.Ping()
}
```

I still find `github.com/apache/arrow-adbc/go/adbc/sqldriver/flightsql` to be very long-winded, but if documentation consistently points to it, then I think we'll help folks out.

srebhan:
> The proposed module would mean that the user just imports a single module that re-exports the interface, without explicitly worrying about ADBC.

@lidavidm that's exactly my point. I my view, `database/sql` (think of it as "Golang Database connectivity") and `ADBC` are two independent standards. So **if I do not worry about ADBC** why am I forces to import it? People probably want to use the one or the other. So I do not suggest to have two implementations, I suggest to remove the ADBC wrapper and make it a wrapper around FlightSQL as this is what is done for every other database in Golang!

zeroshade:
There is possibly some value in having the separate Flight SQL <--> `database/sql` here, as far as contributing goes, though I'd prefer more reasoning than just an "in principle" argument against the extra layers. That said, it's likely not *too* much maintenance overhead to maintain this implementation assuming we get it right.

@srebhan If we can get this to a better spot with good testing, I'd be okay with getting this merged. But I do want to make it clear that the documentation / examples will likely still be pointing at using the ADBC implementation.

I've already left a bunch of comments above that need to be addressed (above and beyond the failing CI). In addition, I'd prefer if this were put into it's own subdirectory of the `flightsql` directory rather than directly in the `flightsql` package itself.

jacobmarble:
Whether the technical merits of the "direct driver" are truly meaningful, I support it for other reasons.

---

> it's likely not too much maintenance overhead to maintain this implementation assuming we get it right.

> If we can get this to a better spot with good testing, I'd be okay with getting this merged. But I do want to make it clear that the documentation / examples will likely still be pointing at using the ADBC implementation.

IMHO, the existence of a "direct driver" is part of "getting it right". Eventually, someone will create such a driver, and it's better that it be a canonical driver live in this repository, rather than `github.com/StartCo/goflightsql`. (I'm thinking about the various Kafka golang clients. Confluent's driver depends on cgo, so it isn't as popular as the "pure Go" drivers published by Segment.io and Shopify.)

Both implementations should behave the same way, as much as possible. Unit tests can help.

> I my view, database/sql (think of it as "Golang Database connectivity") and ADBC are two independent standards. So if I do not worry about ADBC why am I forced to depend on it? People probably want to use the one or the other. So I do not suggest to have two implementations, I suggest to remove the ADBC wrapper and make it a wrapper around FlightSQL as this is what is done for every other database in Golang!

This illustrates a concern I have about the Apache Arrow ecosystem. The relationship between these APIs (Arrow, Flight, FlightSQL, Arrow, database/sql) are not obvious from the outside.

The most common FlightSQL client persona wants a basic SQL driver to access a database that someone else chose. Things like Arrow and ADBC don't matter to this persona, but things like "pure Go" and few dependencies do matter.

go/arrow/flight/flightsql/driver.go
```diff
@@ -0,0 +1,772 @@
...
+	}
+	return r.lastinsert, nil
+}
+
+// RowsAffected returns the number of rows affected by the query.
+func (r *Result) RowsAffected() (int64, error) {
+	if r.affected < 0 {
+		return -1, ErrNotSupported
+	}
+	return r.affected, ErrNotSupported
```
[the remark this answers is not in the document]
srebhan: You are right. Fixed.

go/arrow/flight/flightsql/driver.go
```diff
@@ -0,0 +1,772 @@
...
+// Exec executes a query that doesn't return rows, such
+// as an INSERT or UPDATE.
+func (s *Stmt) Exec(args []driver.Value) (driver.Result, error) {
+	var params []driver.NamedValue
+	for i, arg := range args {
+		params = append(params, driver.NamedValue{
+			Name:    fmt.Sprintf("arg_%d", i),
+			Ordinal: i,
+			Value:   arg,
+		})
```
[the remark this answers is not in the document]
srebhan: You are right. I was under the impression that the `Name` is required (and has to be unique) which is not the case. Remove artificial naming altogether.

go/arrow/flight/flightsql/driver.go
```diff
@@ -0,0 +1,772 @@
...
+	}
+
+	n, err := s.stmt.ExecuteUpdate(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// FIXME: For now we ignore the number of affected records as it seems like
+	// the returned value is always one.
+	_ = n
```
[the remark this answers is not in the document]
srebhan: Fixed.

go/arrow/flight/flightsql/driver.go
```diff
@@ -0,0 +1,772 @@
...
+
+// Query executes a query that may return rows, such as a SELECT.
+func (s *Stmt) Query(args []driver.Value) (driver.Rows, error) {
+	var params []driver.NamedValue
+	for i, arg := range args {
+		params = append(params, driver.NamedValue{
+			Name:    fmt.Sprintf("arg_%d", i),
+			Ordinal: i,
+			Value:   arg,
+		})
```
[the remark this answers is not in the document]
srebhan: There is a subtle difference in the functions, `Query` and `Exec` take a `[]driver.Value` argument while the context variants take a `[]driver.NamedValue`. The difference is that the former is a list of values **that do not have a name**, while the latter are actually named. Therefore, I tried to invent names under the impression that they are required. This doesn't seem to be the case so I removed the artificial naming.

go/arrow/flight/flightsql/driver.go
srebhan: Thanks for the advice and the example. Fixed.

go/arrow/flight/flightsql/driver.go
srebhan: Currently I have nothing to test this against as the example server will not return a parameter schema. Do you have a way to check my (to be written) code will handle the schema correctly?

go/arrow/flight/flightsql/driver.go
srebhan: From what I read about the Golang code there is only one connection per `Driver` instance. Golang will then create a pool of drivers which are assigned to an transaction. There is no way to access the "connection" at the `BeginTx` call... What do you have in mind?

go/arrow/flight/flightsql/driver.go
srebhan: My plan was to extend the driver in a second PR and add TLS customization. Do you prefer to do it in this PR?

srebhan:
@zeroshade thanks again for your thorough review! I've fixed most of your comments (at least IMO ;-)). As I'm not used to your usual habits I left the comments unresolved. Please let me know if I should resolve them by myself or let you go through those.

There are a few point that are open
1. Unification of backends: I think I should not deal with backend-specific settings at that level. So my idea is to unify those and let the user do the specifics on the `DSN` level. I hope to finish this next Friday...
2. TLS implementation/customization: My plan is to implement a mechanism similar to the [MySQL driver](https://github.com/go-sql-driver/mysql) to have a driver-global TLS-config registry. I thought about doing this in a separate PR once this one is merged. What do you think?
3. Moving the driver to another location: You suggested to move this to a separate directory. I'm fine with this, but I'm concerned about testing. The driver, by nature, is closely bound to the API and behavior of the FlightSQL client, therefore it's tests should run whenever something in the client changes. In a separate directory chances are high that the tests are forgotten... Anyway, after unification (see item 1) only `driver.go` and `driver_test.go` will be left. Do you want a separate dir nonetheless?

Thanks again for your support and comments! Much appreciated.

go/arrow/flight/flightsql/driver.go
zeroshade: > From what I read about the Golang code there is only one connection per Driver instance. Golang will then create a pool of drivers which are assigned to an transaction.

That's not quite the case. The `database/sql` package creates one driver instance and maintains a pool of connections. You're guaranteed that a given connection will only be used by one goroutine **at a time**, but you're not guaranteed it won't be used by multiple goroutines in general.

> There is no way to access the "connection" at the BeginTx call... What do you have in mind?

The `Open` function on the Driver should return a connection to the database, not the driver itself. That's where your issue is. `BeginTx` is a method on the `driver.Conn` interface, not `driver.Driver`. You need to fix the implementation to have an actual "connection" type. I highly recommend looking over https://github.com/apache/arrow-adbc/blob/main/go/adbc/sqldriver/driver.go for inspiration.

go/arrow/flight/flightsql/driver.go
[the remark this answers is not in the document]
zeroshade: Looks like the artificial naming is still here and should get removed.

go/arrow/flight/flightsql/driver.go
zeroshade: It should be pretty easy to create a sample testing server to utilize for the tests by following the examples and have that return an artificial schema to test with.

go/arrow/flight/flightsql/driver.go
```diff
@@ -0,0 +1,779 @@
...
+			}
+			record.Retain()
+
+			// Check the schemata
+			if rows.schema == nil {
+				rows.schema = record.Schema()
+			}
+			if !rows.schema.Equal(record.Schema()) {
+				return nil, fmt.Errorf("mixed schemas %w", ErrNotSupported)
+			}
```
zeroshade, line 256: this isn't necessary, all the records in a single stream are currently guaranteed to have the same schema. You should move this above the for loop and just use `reader.Schema()`

go/arrow/flight/flightsql/driver.go
```diff
@@ -0,0 +1,779 @@
...
+	rows := Rows{}
+	for _, endpoint := range info.Endpoint {
+		reader, err := s.client.DoGet(ctx, endpoint.GetTicket())
+		if err != nil {
+			return nil, fmt.Errorf("getting ticket failed: %w", err)
+		}
+
+		for {
+			record, err := reader.Read()
+			if err != nil {
```
zeroshade, line 242: You're incorrectly mixing `Read` and `Next`.

You should either do 

```go
for reader.Next() {
    record := reader.Record()
    record.Retain()
    // do something
}

if reader.Err() != nil {
    // do something with err
}
```

Or remove the `Next` call at the end of the loop. Calling `reader.Read` advances to the next record on its own.

go/arrow/flight/flightsql/driver.go
```diff
@@ -0,0 +1,779 @@
...
+		})
+	}
+
+	schema := s.stmt.ParameterSchema()
+	if schema == nil {
+		schema = arrow.NewSchema(fields, nil)
+	}
+
+	recBuilder := array.NewRecordBuilder(memory.DefaultAllocator, schema)
+	defer recBuilder.Release()
```
zeroshade, line 298: Instead of using `memory.DefaultAllocator` we should allow specifying the allocator by embedding it into a context or by some other way.

go/arrow/flight/flightsql/driver.go
```diff
@@ -0,0 +1,779 @@
...
+}
+
+func setFieldValue(builder array.Builder, arg interface{}) error {
+	switch b := builder.(type) {
+	case *array.BooleanBuilder:
+		switch v := arg.(type) {
+		case bool:
+			b.Append(v)
+		case []bool:
+			b.AppendValues(v, nil)
```
zeroshade, line 582: shouldn't this case be handled by a list builder instead?

go/arrow/flight/flightsql/driver.go
```diff
@@ -0,0 +1,779 @@
...
+	case time.Time:
+		return &arrow.Time64Type{Unit: arrow.Nanosecond}, nil
+	}
+	return nil, fmt.Errorf("type %T: %w", value, ErrNotSupported)
+}
+
+func setFieldValue(builder array.Builder, arg interface{}) error {
+	switch b := builder.(type) {
+	case *array.BooleanBuilder:
+		switch v := arg.(type) {
```
zeroshade, line 578: you also need to handle a union builder

zeroshade (changes requested):
some more comments

zeroshade:
@srebhan 

> Unification of backends: I think I should not deal with backend-specific settings at the driver level. So my idea is to unify those and let the user do the specifics on the DSN level. I hope to finish this next Friday...

Agreed, unifying the backends would be good and is kind of the point of using flightsql in general i think.

> TLS implementation/customization: My plan is to implement a mechanism similar to the [MySQL driver](https://github.com/go-sql-driver/mysql) to have a driver-global TLS-config registry. I thought about doing this in a separate PR once this one is merged. What do you think?

I haven't looked at that too hard but I think it's fine to do it as a separate PR once this is merged.

go/arrow/flight/flightsql/driver.go
srebhan: Looking at that code again, I'm using the schema if given and otherwise create one from the given values. I changed the code a bit to make this more clear, but I don't see what you mean in your initial comment. Can you elaborate a bit @zeroshade?

Regarding creating of a test-server, this is way beyond my current capabilities as it involves parsing the SQL statement itself for adding it to the SQLite mock _or_ writing a complete mockup FlightSQL server...

go/arrow/flight/flightsql/driver.go
srebhan: Can we please also postpone to the point where this is requested by a user? I guess this is not a feature a user of `database/sql` is expecting from the start... :-)

go/arrow/flight/flightsql/driver.go
srebhan: Well the **schema** specified the `BooleanBuilder` and the user passed in a `[]bool`. Isn't this a valid use?

go/arrow/flight/flightsql/driver.go
srebhan: Is this a requirement from the start? Do I need to support _all_ builders? Why doesn't arrow provide a helper for doing this? ;-)

srebhan:
@zeroshade I pushed an update containing fixes to most of your comments above, the unification of the drivers and placement of the driver in a `driver` subdir.

I do have one question to the schema usage (see response to your comment). I'll probably be the wrong person to implement a schema mock, but I'm willing to test if someone implements this.

Regarding the other open points I suggest to first get this PR ready and merged. I'm then willing to work on the following items as follow-up PRs:
1. TLS configuration
2. parameterization of the memory allocator (if we really want to do this without need)
3. mock schema generation for SQLite example (not sure if I can find the time for this)

What do you think?

go/arrow/flight/flightsql/driver.go
zeroshade: For my initial comment, I was referring to the fact that when you attempt to create a new record, if the schema you pass doesn't match the types of the slice of `arrow.Array` you use, then `NewRecord` will panic. So if the conversions you were doing in `toArrowDataType` don't match up with the schema returned from `ParameterSchema` you'd end up getting a panic when you try to create the record due to the mismatch. that's all.

> Regarding creating of a test-server, this is way beyond my current capabilities as it involves parsing the SQL statement itself for adding it to the SQLite mock or writing a complete mockup FlightSQL server...

You don't need a *complete* mockup flightsql server, you only need to implement the necessary methods that would return the ParameterSchema and you can hardcode the schema it returns. For example take a look at the mocked out server used for the ADBC timeout tests: https://github.com/apache/arrow-adbc/blob/main/go/adbc/driver/flightsql/flightsql_adbc_test.go#L709 

it's pretty straightforward I think.

go/arrow/flight/flightsql/driver.go
zeroshade: okay, it can be postponed.

go/arrow/flight/flightsql/driver.go
zeroshade: I don't believe so. Consider the case where the query is something like: `SELECT * FROM <table> WHERE col = ?` and the server returns a parameter schema with a `Boolean` column. If you prepared the above statement, you'd expect a *single* value to be passed, not a slice of values. 

The situation where you have a slice of bools as a single value would be if the expected parameter is supposed to be a *list* of boolean values.

go/arrow/flight/flightsql/driver.go
zeroshade: if you're not going to support *all* value types, then you should at least add a Godoc comment stating that only the types implemented are supported for parameter values.

> Why doesn't arrow provide a helper for doing this? ;-)

It's honestly not a common use case to have an `interface{}` and needing to append it to a `Builder` that isn't already typed. :smile: In the cases where it happens you typically already *know* that the types match so if you type switch on the builder, you can assume the value type or vice versa. The closest case to this would be the way that the `CSV` reader works by having parsing function pointers that it can call which will properly type assert.

As long as what is supported and what is not is explicitly documented for consumers, I'm okay with that.

zeroshade:
@srebhan I think your plan is fine for going ahead. Once you get the CI tests passing again I'll give this another look over.

srebhan:
@zeroshade I found the time to fix the CI issue. The remaining failing test can hardly be due to my changes as I didn't touch any of the mentioned parts/languages...

go/arrow/flight/flightsql/driver.go
srebhan: Understood. Will fix.

go/arrow/flight/flightsql/driver.go
srebhan: Ok, will add a comment and only support scalar values as this is most probably the only thing needed for parameters. If it turns out we require more, we can still add it later...

srebhan:
Thanks @zeroshade, I missed that. Will add the headers...

amol-:

jorisvandenbossche:
@srebhan sorry for closing this a bit too fast

zeroshade:
@srebhan this LGTM for the moment, can you just create issues (referencing this one) for each of the tasks we identified as something that will be done in a followup PR? so that we don't lose track of that work?

I'm good to merge this once those issues are created, potentially creating a top-level umbrella issue with a task list of the separate issues?

srebhan:
@zeroshade I added the two discussed follow ups. The third one (mock schema generation for SQLite example) is already implemented as a test case and thus I omitted an issue.

Did I forget anything? If not I will start implementing the TLS part (#35042) as soon as this PR is merged.

zeroshade (approved):

PR #34323, opened 2023-02-24
GH-34322: [C++][Parquet] Encoding Microbench for ByteArray

mapleFU:
### Rationale for this change

After https://github.com/apache/arrow/pull/14293 . We have `DELTA_BYTE_LENTH` for encoding ByteArray. So, I'd like to have encoding benchmark for them.

### What changes are included in this PR?

Benchmark add some cases.

### Are these changes tested?

No

### Are there any user-facing changes?

No

* Closes: #34322

PR #34336, opened 2023-02-24
GH-34335: [C++][Parquet] Optimize Decoding DELTA_LENGTH_BYTE_ARRAY

### LINKED ISSUE #34335: [C++][Parquet] Improve the performance for Decoding DELTA_LENGTH_BYTE_ARRAY

To be honest, the previous logic is too slow. It's 20 time slower than PLAIN for ByteArray. I'll do some optimizations on it.

### Component(s)

C++, Parquet

mapleFU:
### Rationale for this change

According to https://github.com/apache/arrow/pull/34323 . DELTA_LENGTH_BYTE_ARRAY is much more slower. So do some optimizations.

Some tiny changes

### Are these changes tested?

No more tests included, these code are already tested

No

* Closes: #34335


mapleFU:
This patch is currently for poc and fast verification. If it works and https://github.com/apache/arrow/pull/34323 is merged, I'll making code for it better :)

mapleFU:
```c++
    for (int i = 0; i < max_values; ++i) {
      buffer[i].ptr = data_ptr;
      data_ptr += buffer[i].len;
    }
```

By the way, I don't know would this code cause cache problem or prevent from optimization, causing pipeline hazard or compiler do not optimize here. I'd do some survey.

PR #34340, opened 2023-02-24
GH-34339: [R] Add `skip_rows_after_names` option to `read_csv_arrow`'s options

### LINKED ISSUE #34339: [R] Add `skip_rows_after_names` option to `read_csv_arrow`'s options

Add an option to skip rows after the column names.
This was implemented in C++ and Python (#28410, #10255), but not in R.

### Component(s)

R

eitsupi:
### Rationale for this change

The `skip_rows_after_names` option implemented in C++ was not exposed to R.

The new `skip_rows_after_names` option is now available for the `read_csv_arrow` function.

``` r
> csv <- "a,b\n1,2\n3,4"
> arrow::read_csv_arrow(I(csv), read_options = list(skip_rows_after_names = 1))
# A tibble: 1 × 2
      a     b
  <int> <int>
1     3     4
```

### Are these changes tested?

Tests are added for the new option.

The new `skip_rows_after_names` option can be used.
* Closes: #34339


r/R/csv.R
```diff
@@ -467,6 +473,7 @@ CsvReadOptions <- R6Class("CsvReadOptions",
 CsvReadOptions$create <- function(use_threads = option_use_threads(),
                                   block_size = 1048576L,
                                   skip_rows = 0L,
+                                  skip_rows_after_names = 0L,
```
thisisnic, line 476: Might it be better to add new parameters at the end of the list, so that this won't break code of users who are already using `CsvReadOptions$create()` and passing in parameters without names?

r/R/csv.R
```diff
@@ -385,7 +385,13 @@ CsvTableReader$create <- function(file,
...
 #' `CsvReadOptions$create()` further accepts these additional arguments:
 #'
-#' - `skip_rows` Number of lines to skip before reading data (default 0)
+#' - `skip_rows` Number of lines to skip before reading data (default 0).
+#' - `skip_rows_after_names` Number of lines to skip after the column names (default 0).
+#' This number can be larger than the number of rows in one block, and empty rows are counted.
+#' The order of application is as follows:
+#'   - `skip_rows` is applied (if non-zero);
+#'   - column names aread (unless `column_names` is set);
+#'   - `skip_rows_after_names` is applied (if non-zero).
```
thisisnic, line 394:
```suggestion
#' - `skip_rows_after_names` Number of lines to skip after the column names (default 0).
#'    This number can be larger than the number of rows in one block, and empty rows are counted.
#'    The order of application is as follows:
#'      - `skip_rows` is applied (if non-zero);
#'      - column names are read (unless `column_names` is set);
#'      - `skip_rows_after_names` is applied (if non-zero).
```

thisisnic (changes requested):
Thanks for implementing this, I know there are quite a few users who will appreciate this feature!  Just a couple of minor changes needed.

r/R/csv.R
eitsupi: Thank you for finding this. This is obviously copied from the following, so the following line also needs to be corrected too.
(I will send you a PR later)

https://github.com/apache/arrow/blob/6a4bcb36c091fea07c03c57b2e31dd29f9846ac2/python/pyarrow/_csv.pyx#L113

thisisnic (approved):

eitsupi:
> I absolutely agree that it would make a lot more sense if they were side-by-side in documentation such as vignettes. I don't think we currently mention these options in the vignettes at all, but that could make for a really helpful follow-up issue/PR.

Thanks for the suggestion. I have created a follow-up issue #34383 and would be happy to discuss it here.

PR #34343, opened 2023-02-24
GH-34334: [Go][CSV] Support list fields

### LINKED ISSUE #34334: [Go] CSV not handling all extensions

Seems like CSV missing some types at the moment (including BinaryType and BinaryFixedLengthType) - https://github.com/apache/arrow/blob/main/go/arrow/csv/common.go#L217

Also, missing handling of ArrowExtensionTypes (similar to https://github.com/apache/arrow/issues/34330). 

Will be happy to contribute support for some of the missing types unless there are some restrictions on supporting those types in a csv?

### Component(s)

Go

yevgenypats:
This PR only handles
- list (of all supported type in recursive manner) for both reading and writing CSVs.
- extensions

* Closes: #34334


go/arrow/csv/reader.go
```diff
@@ -721,6 +734,39 @@ func (r *Reader) parseDecimal256(field array.Builder, str string, prec, scale in
...
+		return
+	}
+	if !(strings.HasPrefix(str, "{") || strings.HasSuffix(str, "}")) {
+		r.err = errors.New("invalid list format. should start with '{' and end with '}'")
+		field.AppendNull()
+		return
+	}
+	str = strings.TrimPrefix(str, "{")
+	str = strings.TrimSuffix(str, "}")
+	strs := strings.Split(str, ",")
```
zeroshade, line 749: You probably need to use an actual `csv` reader here instead of just splitting on `","`. The reason is so that quotes are handled properly. Imagine a list of strings where one of the strings has a comma in them. Normal CSV handling would use quotes like `foo,bar,"something,baz"` to handle it. But if you just split on `","` it won't handle the quotes correctly.

go/arrow/csv/reader.go
```diff
@@ -721,6 +734,39 @@ func (r *Reader) parseDecimal256(field array.Builder, str string, prec, scale in
...
+	if len(strs) == 0 {
+		field.AppendNull()
+		return
+	}
```
zeroshade, line 753: an empty list is not equivalent to null. You should call `AppendEmpty` or you can call `Append(true)` and just not add anything to the value builder.

go/arrow/csv/reader.go
```diff
@@ -721,6 +734,39 @@ func (r *Reader) parseDecimal256(field array.Builder, str string, prec, scale in
...
+}
+
+func (r *Reader) parseFixedSizeBinary(field array.Builder, str string, byteWidth int) {
+	if r.isNull(str) {
+		field.AppendNull()
+		return
+	}
+	field.(*array.FixedSizeBinaryBuilder).Append([]byte(str))
+}
```
zeroshade, line 768: We should base64 decode the string so that we can get actual binary data. We should also verify the length of the decoded binary against `byteWidth`

go/arrow/csv/reader_test.go
```diff
@@ -350,10 +358,12 @@ rec[2]["f32"]: [(null)]
...
 rec[2]["ts"]: [(null)]
+rec[2]["list(i64)"]: [(null)]
+rec[2]["uuid"]: (extension_type<storage=fixed_size_binary[16]>)[(null)]
 `
-
-	if got, want := out.String(), want; got != want {
-		t.Fatalf("invalid output:\ngot= %s\nwant=%s\n", got, want)
+	got, want := out.String(), want
+	if diff := cmp.Diff(want, got); diff != "" {
+		t.Fatalf("invalid output: (-want, +got) = %s", diff)
```
zeroshade, line 366: instead of adding a new dependency, we could use `github.com/stretchr/testify/assert` which will also create the diff for you as you need by using `assert.Equal`

go/arrow/csv/reader.go
zeroshade: I'd argue we probably want to make sure that any solution to allow the customization of the behavior *should* default to just calling the underlying storage type if there is no customized version available.

It might make sense to allow providing a custom marshaller / unmarshaller when constructing the Array Type or the ExtensionType itself? And then just defaulting to the underlying storage if that custom one is nil? Thoughts?

yevgenypats:
@zeroshade I've updated this PR to only include list support as figured out it will be easier to do extensions in a follow-up PR potentially after https://github.com/apache/arrow/pull/34454.

go/arrow/csv/reader.go
```diff
@@ -721,6 +726,32 @@ func (r *Reader) parseDecimal256(field array.Builder, str string, prec, scale in
...
+func (r *Reader) parseList(field array.Builder, str string) {
+	if r.isNull(str) {
+		field.AppendNull()
+		return
+	}
+	if !(strings.HasPrefix(str, "{") || strings.HasSuffix(str, "}")) {
+		r.err = errors.New("invalid list format. should start with '{' and end with '}'")
+		field.AppendNull()
+		return
+	}
```
zeroshade, line 738: The logic here isn't quite correct. It should be 

`if !strings.HasPrefix(str, "{") || !strings.HasSuffix(str, "}") {`

The current logic would only error if *both* are missing. Is there a unit test which tests the case where there is only a prefix "{" or only a suffix "}"?

go/arrow/csv/reader.go
```diff
@@ -721,6 +726,32 @@ func (r *Reader) parseDecimal256(field array.Builder, str string, prec, scale in
...
+	valueBldr := listBldr.ValueBuilder()
+	reader := csv.NewReader(strings.NewReader(str))
+	items, err := reader.Read()
+	if err != nil {
+		r.err = err
+		return
+	}
+	for _, str := range items {
+		r.initFieldConverter(valueBldr)(str)
+	}
```
zeroshade, line 752: a future enhancement might be to cache this somehow... but it's not needed for this PR

zeroshade:
@yevgenypats This looks good to me other than my last two comments. Please add a couple tests for checking error cases and an empty list case. After that this LGTM :smile: thanks again for this!

yevgenypats:
Thanks @zeroshade ! BTW - any idea why the `GO / AMD64 macOS 11 Go 1.17/1.18 - CGO` fails? seems unrelated to this PR.

zeroshade:
@yevgenypats it is unrelated to this PR, I'm addressing it in #34488

go/arrow/csv/reader.go
```diff
@@ -731,15 +731,19 @@ func (r *Reader) parseList(field array.Builder, str string) {
...
 		return
 	}
-	if !(strings.HasPrefix(str, "{") || strings.HasSuffix(str, "}")) {
+	if !(strings.HasPrefix(str, "{") && strings.HasSuffix(str, "}")) {
 		r.err = errors.New("invalid list format. should start with '{' and end with '}'")
-		field.AppendNull()
 		return
 	}
 	str = strings.TrimPrefix(str, "{")
 	str = strings.TrimSuffix(str, "}")
```
zeroshade, line 739: I think this can be condensed into a single call to `strings.Trim(str, "{}")`

zeroshade:
If you rebase in from master, that'll pick up the fix for the failing macos tests. Once you add a test for the error case, I'm happy to merge this :)

yevgenypats:
> If you rebase in from master, that'll pick up the fix for the failing macos tests. Once you add a test for the error case, I'm happy to merge this :)

Awesome. Done!

go/arrow/csv/reader.go
yevgenypats: Good catch. Fixed and added tests.

go/arrow/csv/reader.go
yevgenypats: I can change this but I feel the explicit way is clearer and potentially more efficient. But let me know if you want it the other way.

zeroshade: Looking at the implementations from the `strings` package, I'm pretty sure that calling `strings.Trim` is more efficient than calling `TrimPrefix` and `TrimSuffix` separately. At minimum it reduces the number of code branches that get used. Personally I'd prefer using `strings.Trim` over the separate calls to prefix/suffix.

zeroshade:
@yevgenypats Turns out there was a typo in the go workflow yaml in the fix for the cgo macos tests, please rebase main again to get the Go workflows running again (we fixed the typo) thanks!

go/arrow/csv/reader.go
yevgenypats: Updated

zeroshade (approved):
LGTM

Thanks much!

PR #34348, opened 2023-02-25
GH-34347: [C++] Add an initial fuzz test for the scanner

### LINKED ISSUE #34347: [C++] Add an end-to-end fuzz test for the new scan node

The scanner has quite a few unit tests.  However, I think it would benefit from some more robust end-to-end testing as well.

As further justification, I have found now three different issues (#15059 #18481 and one I am fixing in the PR for this issue) just getting this fuzz test started.

### Component(s)

C++

westonpace:
This is just an starting point.  There are a number of different things that will hopefully be eventually added (support for more formats, different scanning scenarios, running multiple scans in parallel).

It's currently exposed as a unit test and can maybe be a bit slow (takes ~30 seconds with one thread on my machine).  Eventually it might make more sense to change this to a nightly test that runs for a longer duration.  We should also probably have a version that runs on release mode.

I also fixed an issue encountered with partition guarantee normalization.  I could separate that into a separate PR if desired.
* Closes: #34347


lidavidm (approved):

westonpace:
These test failures are legimate.  I hacked at this a little today and it quickly started to get out of control :laughing: 

The basic issue is that I am transferring to a model where all expressions (projections, filters, guarantees) get normalized as we start to work with them.  This allows us to cleanly handle duplicate column names and similar issues.  Unfortunately, it's exposing a number of places in the current datasets code where we are explicitly relying on named refs and I am needing to work around that.

amol-:

PR #34355, opened 2023-02-26
GH-34351: [C++][Parquet] Statistic: tiny optimization

### LINKED ISSUE #34351: [C++][Parquet] Statistics Merge ignore setting `has` flag

In `src/parquet/statistics.cc`:

```c++
   void Merge(const TypedStatistics<DType>& other) override {
     this->num_values_ += other.num_values();
     if (other.HasNullCount()) {
-      this->statistics_.null_count += other.null_count();
+      this->IncrementNullCount(other.null_count());
     }
     if (other.HasDistinctCount()) {
-      this->statistics_.distinct_count += other.distinct_count();
+      this->IncrementDistinctCount(other.distinct_count());
     }
     if (other.HasMinMax()) {

```

The original code ignore setting `has` flag.

### Component(s)

C++, Parquet

mapleFU:
### Rationale for this change

This patch does some tiny optimizations on Parquet C++ Statistics. It does:

1. For min-max, using `std::string`. Because assume the case like that:

```c++
EncodedStatistics c1;
// do some operations
EncodedStatistics c2 = c1;
c2.set_max("dasdasdassd");
```

After c2 set, c1 would be set too. So I use std::string here.

2. Force setting ndv count during merging, and add some comments

As we talked above.

### Are these changes tested?

Test by unittest.

No 

* Closes: #34351


mapleFU:
```
===============================================================================
Omission: parquet::RowGroupMetaData::Equals() isn't stable. [test: #==(TestParquetRowGroupMetadata)]
/Users/runner/work/arrow/arrow/c_glib/test/parquet/test-row-group-metadata.rb:48:in `block in <class:TestParquetRowGroupMetadata>'
===============================================================================
.........F
===============================================================================
Failure: test: #has_n_distinct_values?(TestParquetStatistics):
        @statistics.has_n_distinct_values?
        |           |
        |           false
        #<Parquet::Int32Statistics:0x7f86979871f8 ptr=0x7f8698c1c8d0>
/Users/runner/work/arrow/arrow/c_glib/test/parquet/test-statistics.rb:53:in `block in <class:TestParquetStatistics>'
     50:   end
     51: 
     52:   test("#has_n_distinct_values?") do
  => 53:     assert do
     54:       @statistics.has_n_distinct_values?
     55:     end
     56:   end
===============================================================================
..............O
```

emmmm

mapleFU:
After go through the piece of code, I found that current impl is ok, because we mostly only use statistics on writer. But when `ExtractStatisticsFromPageHeader` or other reader part is in, things will get a bit more complex.

As for now, we can assume that:

1. Writer can assure that if has right null-count ( if it not has any bugs )
2. Currently I found that ndv is never collected. If a user collect ndv in page1, but not collect ndv in page 2, it should be abandon.

For reader:

1. When deserialize, reader should assume that ndv and null_count can be unset ( but currently, it doesn't work like this)
2. Deserialized statistics can **not** call merge or other mutation methods

Currently, a writer will not has bug on merging. But if a reader, checks `has_null_count` or `ndv`, it will get the wrong result

cpp/src/parquet/statistics.cc
```diff
@@ -495,9 +495,12 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
                       bool has_null_count, bool has_distinct_count, MemoryPool* pool)
       : TypedStatisticsImpl(descr, pool) {
     TypedStatisticsImpl::IncrementNumValues(num_values);
+    // Currently, `has_null_count` argument is not used.
+    // Internal has_null_count_ would always be true.
```
wgtmac, line 499: IMO, if `has_null_count` is not used we should remove it to reduce confusion and future misuse. Otherwise, we should respect it from the input argument.

cpp/src/parquet/statistics.cc
```diff
@@ -620,7 +626,11 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
  private:
   const ColumnDescriptor* descr_;
   bool has_min_max_ = false;
+  // Currently, has_null_count_ is always true, and would
```
wgtmac, line 629: This is confusing because the default value below is `false`

cpp/src/parquet/statistics.cc
```diff
@@ -553,10 +556,13 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
...
-      this->statistics_.null_count += other.null_count();
+      this->IncrementNullCount(other.null_count());
     }
-    if (other.HasDistinctCount()) {
-      this->statistics_.distinct_count += other.distinct_count();
+    if (HasDistinctCount() && other.HasDistinctCount()) {
+      this->IncrementDistinctCount(other.distinct_count());
+    } else {
+      this->has_distinct_count_ = false;
+      this->has_distinct_count_ = 0;
```
wgtmac, line 565:
```suggestion
      statistics_.distinct_count = 0;
```

cpp/src/parquet/statistics.cc
```diff
@@ -620,7 +626,11 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
  private:
   const ColumnDescriptor* descr_;
   bool has_min_max_ = false;
+  // Currently, has_null_count_ is always true, and would
+  // be collected and encoded to `EncodedStatistics`.
   bool has_null_count_ = false;
+  // Currently, has_distinct_count_ would not be encoded
```
wgtmac, line 632: It seems that `statistics_` already has a set of `hasXXX` variables. Can we reuse those directly?

```cpp
class PARQUET_EXPORT EncodedStatistics {
  std::shared_ptr<std::string> max_, min_;
  bool is_signed_ = false;

 public:
  EncodedStatistics()
      : max_(std::make_shared<std::string>()), min_(std::make_shared<std::string>()) {}


  int64_t null_count = 0;
  int64_t distinct_count = 0;

  bool has_min = false;
  bool has_max = false;
  bool has_null_count = false;
  bool has_distinct_count = false;
```

cpp/src/parquet/statistics_test.cc
```diff
@@ -377,6 +377,35 @@ class TestStatistics : public PrimitiveTypedTest<TestType> {
     ASSERT_EQ(total->max(), std::max(statistics1->max(), statistics2->max()));
   }
 
+  void TestMergeEmpty() {
```
wgtmac, line 380: It would be better to add a case where either side does not have a valid stats (min/max/null_count/distinct_count) meaning that the merged stats is dropped.

cpp/src/parquet/statistics.cc
mapleFU: Yes, default value is false, but ctor will ensure that it is true. I can update it to `true`.

cpp/src/parquet/statistics_test.cc
mapleFU: Sure, I will add them

cpp/src/parquet/statistics.cc
mapleFU: I agree with you, but I need other's opinion

mapleFU:
@kou I guess ruby test should not set has ndv, because current parquet writer will never put valid ndv. Let me set it to false later.

As @wgtmac says, maybe it's better to make it "false"

kou:
OK!

mapleFU:
@pitrou @emkornfield I need some idea about the code here. The problem is that:
1. When writing statistics, the code is ok.
2. `Statistics::Make` would create statistics using `EncodedStats` for read, then the code here would be a disaster

cpp/src/parquet/statistics.cc
wjones127: IIUC, this class is only used in the writer and we always write the null count in statistics? If so, then I agree we can remove this.

mapleFU: After a weird patch, this class is used in reader, but ndv is not used.

wjones127: what is ndv?

mapleFU: Sorry for make misunderstanding. NDV means num-different values. Load statistics is introduced in `ColumnChunkMetaData`, and it's weird that:

1. Writer will never set ndv.
2. Reader should read ndv.

The behavior here is trickey

mapleFU:
This patch is neccessary but not visited for a long time. I'll close and recreate this this weekend.

PR #34358, opened 2023-02-27
GH-34284: [Java][FlightRPC] Fixed issue with prepared statement getting sent twice

### LINKED ISSUE #34284: [Java] JDBC Flight SQL driver calls prepare() twice

[From user@](https://lists.apache.org/thread/9smgkz1x3wl27yl7nj55d8qdmlmhpgm2):

> I noticed that when I use ArrowFlightJdbcDriver to run a prepare statement
> query. It sends prepare query twice to the backend flight server.
> 
> The first query is triggered in ArrowFlightMetaImpl#prepare method: link:
> https://github.com/apache/arrow/blob/45918a90a6ca1cf3fd67c256a7d6a240249e555a/java/flight/flight-sql-jdbc-driver/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightMetaImpl.java#L134
> 
> 
> The second query is triggered
> in ArrowFlightJdbcFactory#newPreparedStatement: link
> https://github.com/apache/arrow/blob/45918a90a6ca1cf3fd67c256a7d6a240249e555a/java/flight/flight-sql-jdbc-driver/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightJdbcFactory.java#L84
> 
> Could you shed some light on this? Why it's called twice? Does backend
> flight server need skip the second duplicate query in such case?

### Component(s)

FlightRPC, Java

rtadepalli:
### Rationale for this change

The `AvaticaConnection` class within the Calcite library [executes](https://github.com/apache/calcite-avatica/blob/b57eb7cd31a90d3f46b65c13832b398be5b0dad9/core/src/main/java/org/apache/calcite/avatica/AvaticaConnection.java#L357) the `prepareStatement(...)` function on a DB connection in the following order:
1. Calls `prepare()` on the `MetaImpl` class, which in our case is `ArrowFlightMetaImpl`. This function executes a command and returns a prepared statement.
2. Calls `newPreparedStatement` on the JDBC factory class, which in our case is `ArrowFlightJdbcFactory`. This function today creates an entirely new `PreparedStatement` without checking the statement cache, thereby causing 2 executions.

This PR fixes the issue by checking whether or not the `StatementHandle` has been recorded before, and if so, uses the results of the previously executed command. If not, it creates a brand new `PreparedStatement`.

My understanding is that `ConcurrentHashMap` (which is what is being used to keep a mapping of statement handles to prepared statements) is eventually consistent, so there _may_ be cases where the prepared statement execution runs twice.

Check to see if the `PreparedStatement` has run before, and if so, use that instead of executing the command to contact the server again.

### Are these changes tested?

Existing tests should cover this case. Namely, `ArrowFlightPreparedStatementTest` I _think_ should cover all cases.

There are no user facing changes.
* Closes: #34284

### DIFF
--- a/java/flight/flight-sql-jdbc-driver/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightPreparedStatement.java
+++ b/java/flight/flight-sql-jdbc-driver/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightPreparedStatement.java
@@ -81,12 +81,12 @@ public class ArrowFlightPreparedStatement extends AvaticaPreparedStatement
   }
 
   static ArrowFlightPreparedStatement newPreparedStatement(final ArrowFlightConnection connection,
-                                                           final ArrowFlightSqlClientHandler.PreparedStatement preparedStmt,
-                                                           final StatementHandle statementHandle,
-                                                           final Signature signature,
-                                                           final int resultSetType,
-                                                           final int resultSetConcurrency,
-                                                           final int resultSetHoldability) throws SQLException {
+      final ArrowFlightSqlClientHandler.PreparedStatement preparedStmt,
+      final StatementHandle statementHandle,
+      final Signature signature,
+      final int resultSetType,
+      final int resultSetConcurrency,
+      final int resultSetHoldability) throws SQLException {
     return new ArrowFlightPreparedStatement(
         connection, preparedStmt, statementHandle,
         signature, resultSetType, resultSetConcurrency, resultSetHoldability);
--- a/java/flight/flight-sql-jdbc-driver/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightPreparedStatement.java
+++ b/java/flight/flight-sql-jdbc-driver/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightPreparedStatement.java
@@ -39,12 +39,12 @@ public class ArrowFlightPreparedStatement extends AvaticaPreparedStatement
 
   private final ArrowFlightSqlClientHandler.PreparedStatement preparedStatement;
 
-  private ArrowFlightPreparedStatement(final ArrowFlightConnection connection,
-                                       final ArrowFlightSqlClientHandler.PreparedStatement preparedStatement,
-                                       final StatementHandle handle,
-                                       final Signature signature, final int resultSetType,
-                                       final int resultSetConcurrency,
-                                       final int resultSetHoldability)
+  public ArrowFlightPreparedStatement(final ArrowFlightConnection connection,
+                                      final ArrowFlightSqlClientHandler.PreparedStatement preparedStatement,
+                                      final StatementHandle handle,
+                                      final Signature signature, final int resultSetType,
+                                      final int resultSetConcurrency,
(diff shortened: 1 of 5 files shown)


rtadepalli:
I have left `ArrowFlightPreparedStatement#createNewPreparedStatement` as is, in case. There are 0 usages of that method right now, though.

java/flight/flight-sql-jdbc-driver/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightPreparedStatement.java
```diff
@@ -39,12 +39,12 @@ public class ArrowFlightPreparedStatement extends AvaticaPreparedStatement
 
   private final ArrowFlightSqlClientHandler.PreparedStatement preparedStatement;
 
-  private ArrowFlightPreparedStatement(final ArrowFlightConnection connection,
-                                       final ArrowFlightSqlClientHandler.PreparedStatement preparedStatement,
-                                       final StatementHandle handle,
-                                       final Signature signature, final int resultSetType,
-                                       final int resultSetConcurrency,
-                                       final int resultSetHoldability)
+  public ArrowFlightPreparedStatement(final ArrowFlightConnection connection,
```
lidavidm, line 42: Instead of a public constructor, would it be possible to add a new static (package-private) factory method?

lidavidm (commented):
Thanks for the PR!

Would it be possible to add a test case demonstrating this? Or would that require too much mocking/mucking with the internals?

java/flight/flight-sql-jdbc-driver/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightPreparedStatement.java
rtadepalli: Changed.

rtadepalli:
I can write a test, but I think that would involve messing with the internals a fair bit. I did a POC of sorts, and I think it would require mocking out `FlightSqlClient`, `FlightClient`, in addition to making a few variables non-final to inject mocks. The only way I can think of testing this is to make these changes and count the number of invocations for `FlightClient#doAction` after `prepare(...)` is  called -- please let me know if you have any other ideas.

lidavidm:
Hmm, perhaps an easier way is to use the FlightSqlProducer implementation that's already in the tests. It could be modified to track how many prepared statements have been created, then a test can assert on that value.

rtadepalli:
Oh wow I didn't really think about that. Made testing a lot easier, thanks!

Didn't convert `FlightSqlProducer` into an abstract class but rather took the easy way out of just creating a static map counter. Hope that is fine.

lidavidm:
Oops! I didn't mean the concrete class, there's a subclass that overrides methods and is used for tests

rtadepalli:
I saw `MockFlightSqlProducer`, but this just seemed easier...

rtadepalli:
Alright I have moved the logic down into `MockFlightSqlProducer`.

lidavidm (approved):
Thank you!

PR #34369, opened 2023-02-27
GH-34366: [R] Don't getFromNamespace() the dplyr:::check_name() helper

### LINKED ISSUE #34366: Don't `getFromNamespace()` the `dplyr:::check_name()` helper

Right here arrow grabs the `dplyr:::check_name()` helper:
https://github.com/apache/arrow/blob/969dd290ab93c8629b643134f150e71c1c05c2ff/r/R/dplyr-count.R#L42-L43

We've renamed this in dplyr 1.1.1 to `check_n_name()` as part of https://github.com/tidyverse/dplyr/pull/6722

That PR is part of a larger move towards using some standard internal `check_*()` functions throughout our tidyverse packages, and one of those is `check_name()`, which conflicted with the previous dplyr function name. The `import-standalone-*.R` files that get brought in there are read only, so I'd prefer to not have to tweak them just to make arrow happy.

It seems like `dplyr:::check_n_name()` is fairly simple, is it possible to just inline it in arrow directly? That seems to be the only place arrow does `getFromNamespace()` for dplyr, so it would be great to remove that entirely anyways.

We will likely be releasing dplyr 1.1.1 in 2-3 weeks, so it would be great to have this updated before then!

### Component(s)

R

thisisnic:
We use an internal function from dplyr in the tally function; this PR copies the code directly instead of pulling it from their namespace. This solves an issue whereby an upcoming release of dplyr was failing its revdepchecks.

Fixes #34366


paleolimbot (commented):
This change looks appropriate to me; however, it looks like we don't test the behaviour of this internally (I looked in test-dplyr-count.R). Now that it's in our code base, I think we need to add tests...notably one that will produce the message and one that checks the error when `name` is not a string.

paleolimbot (approved):

PR #34371, opened 2023-02-27
MINOR: [C++][Docs] Remove note about Compute IR in the Acero user guide

jorisvandenbossche:
Compute IR was abandoned in favor of Substrait, and this is already handled above the current section in that doc page.


kou (approved):
+1

PR #34400, opened 2023-03-01
GH-33976: [Python] Rename pyarrow._exec_plan to pyarrow._acero

jorisvandenbossche:
https://github.com/apache/arrow/pull/34102 is adding a new `pyarrow._acero` cython module (we probably also want a public `pyarrow.acero` at some point, but that's for later), and in a next PR I will refactor the existing `_exec_plan.pyx` to use those new bindings. Given that we don't need to keep that as two separate cython files, this PR is already renaming `_exec_plan.pyx` to `_acero.pyx` as a precursor to reduce the diff in the subsequent PRs.

* Issue: #33976


jorisvandenbossche:
Actually, on second thought, the goal should be that the current `_exec_plan.pyx` can be fully done in _python_, so this when doing that refactor, we can rename `_exec_plan.pyx` to `acero.py`, instead of merging it into the new `_acero.pyx`. Hence, a rename in advance doesn't help.

PR #34401, opened 2023-03-01
GH-33976: [Python] Refactor the internal usage of ExecPlan to use new Acero Declaration/ExecNodeOptions bindings

jorisvandenbossche:
This PR refactors our current custom cython implementation of the Table/Dataset.filter/join/group_by/sort_by methods to use the new bindings for Declaration/ExecNodeOptions (https://github.com/apache/arrow/pull/34102).

* Issue: #33976


python/pyarrow/_exec_plan.pyx
```diff
@@ -235,163 +237,121 @@ def _perform_join(join_type, left_operand not None, left_keys,
...
-                c_join_type, c_left_keys, c_right_keys,
-                _true,
-                <c_string>tobytes(left_suffix or ""),
-                <c_string>tobytes(right_suffix or "")
-            ))
+    # TEMP
+    if isinstance(left_operand, Dataset):
+        left_operand = left_operand.to_table()
+    if isinstance(right_operand, Dataset):
+        right_operand = right_operand.to_table()
```
jorisvandenbossche, line 287: This is just a temporary hack because for now there is only a TableSourceNodeOptions, I need to add a proper ScanNodeOptions in another pre-cursor PR

jorisvandenbossche:
With the last commit, I renamed `_exec_plan.pyx` to `_exec_plan.py`, because it no longer contains any cython. That makes github see it as a deleted and new file, though. If you want to see the diff within the file itself, select all but the last commit: https://github.com/apache/arrow/pull/34401/files/11204cb7ff4071b4799bb4f064e8cc9f116351c8

python/pyarrow/tests/test_exec_plan.py
```diff
@@ -209,44 +209,27 @@ def test_filter_table_errors():
     })
 
     with pytest.raises(pa.ArrowTypeError):
-        ep._filter_table(
-            t, pc.divide(pc.field("a"), pc.scalar(2)),
-            output_type=pa.Table
```
jorisvandenbossche, line 214: Since we only ever used `output_type=pa.Table` option for this keyword in `_filter_table`, I removed the keyword and simplified the tests

AlenkaF (approved):
I see nothing obvious missing. The code is hugely optimised and much clearer now so the maintenance of it will benefit big time 👍  Thanks!

PR #34406, opened 2023-03-01
GH-34405: [C++] Add support for custom names in QueryOptions.  Wire this up to Substrait

### LINKED ISSUE #33434: [C++] Add Acero test to ensure names in the root relation of Substrain plans are retained
Currently, Acero retains the names in the root relation when it executes a Substrait plan but there isn't a unit test in place to prevent a future regression.

**Reporter**: [Bryce Mecum](https://issues.apache.org/jira/browse/ARROW-18252) / @amoeba

<sub>**Note**: *This issue was originally created as [ARROW-18252](https://issues.apache.org/jira/browse/ARROW-18252). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

### LINKED ISSUE #34405: [C++] Make it possible to specify custom field names in QueryOptions

This is needed, for example, to run Substrait queries because those queries specify the output column names explicitly.  In general, it seems that it would be a pretty common ask to support column aliases instead of the auto-generated names (which can be quite unwieldy when casts are involved).

### Component(s)

C++

westonpace:
### Rationale for this change

Users want to be able to specify custom column names / aliases instead of using the ones generated by Acero

It is now possible to specify custom column names in QueryOptions.  In addition, the python Substrait bindings now use this feature so that the Substrait plan's names will be respsected.

### Are these changes tested?

Yes.  These are tested directly.  In addition, I added a python test for the Substrait bindings as this is actually a regression there and this should close https://github.com/apache/arrow/issues/33434.

There is new API surface but nothing breaking.
* Closes: #34405
* Closes: gh-33434

### DIFF
--- a/cpp/src/arrow/compute/exec/exec_plan.cc
+++ b/cpp/src/arrow/compute/exec/exec_plan.cc
@@ -951,8 +951,11 @@ struct BatchConverter {
     if (!names.empty()) {
       if (static_cast<int>(names.size()) != schema->num_fields()) {
         return Status::Invalid(
-            "A plan was created with custom field names but the number of names did not "
-            "match the number of output columns");
+            "A plan was created with custom field names but the number of names (",
+            names.size(),
+            ") did not "
+            "match the number of output columns (",
+            schema->num_fields(), ")");
       }
       ARROW_ASSIGN_OR_RAISE(schema, schema->WithNames(names));
     }
--- a/cpp/src/arrow/compute/exec/exec_plan.cc
+++ b/cpp/src/arrow/compute/exec/exec_plan.cc
@@ -624,6 +624,7 @@ Future<std::shared_ptr<Table>> DeclarationToTableImpl(
   ARROW_ASSIGN_OR_RAISE(std::shared_ptr<ExecPlan> exec_plan, ExecPlan::Make(exec_ctx));
   TableSinkNodeOptions sink_options(output_table.get());
   sink_options.sequence_output = query_options.sequence_output;
+  sink_options.names = std::move(query_options.field_names);
   Declaration with_sink =
       Declaration::Sequence({declaration, {"table_sink", sink_options}});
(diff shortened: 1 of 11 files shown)


cpp/src/arrow/compute/exec/options.h
```diff
@@ -614,6 +616,13 @@ class ARROW_EXPORT TableSinkNodeOptions : public ExecNodeOptions {
   ///
   /// \see QueryOptions for more details
   std::optional<bool> sequence_output;
+  /// \brief Custom names to use for the columns.
+  ///
+  /// If specified then names must be provided for all fields. Currently, only a flat
+  /// schema is supported (see ARROW-15901).
```
lidavidm, line 622: nit: maybe we should refer to the GitHub issue instead? (I assume this was copied from above)

cpp/src/arrow/compute/exec/exec_plan.cc
```diff
@@ -925,14 +943,32 @@ struct BatchConverter {
...
+  Result<std::shared_ptr<Schema>> InitializeSchema(
+      const std::vector<std::string>& names) {
+    // By this point this->schema will have been set by the SinkNode.  We potentially
+    // rename it with the names provided by the user and then return this in case the user
+    // wants to know the output schema.
+    if (!names.empty()) {
+      if (static_cast<int>(names.size()) != schema->num_fields()) {
+        return Status::Invalid(
+            "A plan was created with custom field names but the number of names did not "
+            "match the number of output columns");
```
lidavidm, line 955: nit: include expected/actual count in the message for convenience?

lidavidm (approved):

cpp/src/arrow/compute/exec/exec_plan.cc
westonpace: Done.

lidavidm (approved):

PR #34416, opened 2023-03-02
GH-34262: [C++][ORC] Support union type

### LINKED ISSUE #34262: [C++] Support for union type in ORC writer

I've built an union manually with `pa.UnionArray.from_dense()`, not using inference,

In python I'm trying to write that union to an ORC file and I'm getting the following error:

```
pyarrow.lib.ArrowNotImplementedError: Unknown or unsupported Arrow type: dense_union<int64: int64=0, float64: double=1, string: string=2>
```

It seems that this in not supported even in C++ code that, IMO, python uses:

https://github.com/apache/arrow/blob/f2d632e727d4eb7e8a693354c9116ea2518e50a6/cpp/src/arrow/adapters/orc/util.cc#L920

Any plans on adding this feature to arrow?

### Component(s)

C++, Python

wgtmac:
### Rationale for this change

The ORC adapter does not support union type yet.

Support union type to both ORC reader and writer.

### Are these changes tested?

To be added.

No.
* Closes: #34262

### DIFF
--- a/cpp/src/arrow/adapters/orc/adapter_test.cc
+++ b/cpp/src/arrow/adapters/orc/adapter_test.cc
@@ -840,6 +840,23 @@ TEST_F(TestORCWriterSingleArray, WriteListOfMap) {
   AssertArrayWriteReadEqual(array, array, kDefaultSmallMemStreamSize * 10);
 }
 
+TEST_F(TestORCWriterSingleArray, WriteSparseUnion) {
+  const int64_t num_rows = 1024;
+  auto type =
+      sparse_union({field("_union_0", utf8()), field("_union_1", int32())}, {0, 1});
+  auto array = checked_pointer_cast<SparseUnionArray>(rand.ArrayOf(type, num_rows, 0.4));
+  ArrayVector children;
+  for (int i = 0; i < array->num_fields(); ++i) {
+    ASSERT_OK_AND_ASSIGN(auto flattened_child, array->GetFlattenedField(i));
+    children.emplace_back(std::move(flattened_child));
+  }
+  auto flattened_array = std::make_shared<SparseUnionArray>(
+      array->type(), array->length(), std::move(children), array->type_codes(),
+      array->offset());
+  AssertArrayWriteReadEqual(flattened_array, flattened_array,
+                            kDefaultSmallMemStreamSize * 10);
+}
+
 class TestORCWriterMultipleWrite : public ::testing::Test {
  public:
   TestORCWriterMultipleWrite() : rand(kRandomSeed) {}
--- a/cpp/src/arrow/adapters/orc/adapter_test.cc
+++ b/cpp/src/arrow/adapters/orc/adapter_test.cc
@@ -22,6 +22,7 @@
 #include <string>
 
 #include "arrow/adapters/orc/adapter.h"
+#include "arrow/adapters/orc/util.h"
 #include "arrow/array.h"
 #include "arrow/buffer.h"
 #include "arrow/buffer_builder.h"
@@ -34,7 +35,6 @@
 #include "arrow/testing/gtest_util.h"
 #include "arrow/testing/random.h"
 #include "arrow/type.h"
-#include "arrow/util/decimal.h"
 #include "arrow/util/key_value_metadata.h"
 
 namespace liborc = orc;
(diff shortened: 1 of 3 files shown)


wgtmac:
@westonpace @wjones127 Could you take a look please?

cpp/src/arrow/adapters/orc/adapter_test.cc
```diff
@@ -840,6 +840,23 @@ TEST_F(TestORCWriterSingleArray, WriteListOfMap) {
...
+      sparse_union({field("_union_0", utf8()), field("_union_1", int32())}, {0, 1});
+  auto array = checked_pointer_cast<SparseUnionArray>(rand.ArrayOf(type, num_rows, 0.4));
+  ArrayVector children;
+  for (int i = 0; i < array->num_fields(); ++i) {
+    ASSERT_OK_AND_ASSIGN(auto flattened_child, array->GetFlattenedField(i));
+    children.emplace_back(std::move(flattened_child));
+  }
+  auto flattened_array = std::make_shared<SparseUnionArray>(
+      array->type(), array->length(), std::move(children), array->type_codes(),
+      array->offset());
```
westonpace, line 855: Why is this flattening needed?

cpp/src/arrow/adapters/orc/util.cc
```diff
@@ -994,7 +1125,6 @@ Result<std::shared_ptr<DataType>> GetArrowType(const liborc::Type* type) {
       } else {
         return decimal128(precision, scale);
       }
-      break;
```
westonpace, line 997: Why remove this break?

westonpace (commented):
This looks good.  Thank you for adding this.  I have some small questions (probably just my misunderstanding)

westonpace:
Does your test cover the read path?  Also, do you want a test with a dense union?

cpp/src/arrow/adapters/orc/adapter_test.cc
wgtmac: The random array generator fill random unselected values in the child arrays of `SparseUnionArray`. However, orc file only contains dense union type meaning that these unselected values will not be written to the file (so we can never read them back and compare equality in the unit test). In this PR, I fill unselected values to nulls when reading from the file. So flattening the `SparseUnionArray` before writing makes it easy for the roundtrip equality check of `SparseUnionArray`.

cpp/src/arrow/adapters/orc/util.cc
wgtmac: Because it is unreachable.

wgtmac:
> Does your test cover the read path? Also, do you want a test with a dense union?

Yes, I have added a roundtrip test for sparse union type. The read path of dense union will never happen because we always convert orc union to arrow sparse union type. Let me add a test case for batch conversion roundtrip of dense union to make it coverred.

cpp/src/arrow/adapters/orc/util.h
```diff
@@ -35,8 +35,9 @@ Result<std::shared_ptr<DataType>> GetArrowType(const liborc::Type* type);
 
 Result<ORC_UNIQUE_PTR<liborc::Type>> GetOrcType(const Schema& schema);
 
-Status AppendBatch(const liborc::Type* type, liborc::ColumnVectorBatch* batch,
-                   int64_t offset, int64_t length, arrow::ArrayBuilder* builder);
+ARROW_EXPORT Status AppendBatch(const liborc::Type* type,
```
wgtmac, line 38: I have to mark them as `ARROW_EXPORT`. Otherwise the windows build will complain unresolved references for the new tests.

cpp/src/arrow/adapters/orc/util.cc
westonpace: Ah, I see now :).  Thanks.

cpp/src/arrow/adapters/orc/adapter_test.cc
westonpace: Can you maybe add a small comment explaining this for future readers that might not be as familiar with orc?

westonpace: nevermind, I see you already did.  Thanks!

westonpace (approved):
Thanks for adding this support.

PR #34418, opened 2023-03-02
MINOR: [Go][FlightSQL] reference the ADBC drivers in README

jacobmarble:
### Rationale for this change

It is not easy to find the Golang database/sql FlightSQL driver. This change anticipates that people will look for it here first.

Only README content.

### Are these changes tested?

No.

No.

### DIFF
--- a/go/README.md
+++ b/go/README.md
@@ -40,11 +40,17 @@ import (
 )
 
 func main() {
-    db, err := sql.Open("flightsql", "uri=grpc://localhost:12345")
+	dsn := "uri=grpc://localhost:12345;username=mickeymouse;password=p@55w0RD"
+    db, err := sql.Open("flightsql", dsn)
     ...
 }
 ```
 
+DSN option keys are expressed as `k=v`, delimited with `;`. 
+Some options keys are defined in ADBC, others are defined in the FlightSQL ADBC driver.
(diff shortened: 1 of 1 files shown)


jacobmarble:
> * Closes: [[Go][FlightRPC] Add driver for `database/sql` framework #34332](https://github.com/apache/arrow/issues/34332)

This change *does not* close #34332

go/README.md
```diff
@@ -28,6 +28,23 @@ for flat and hierarchical data, organized for efficient analytic operations on
 modern hardware. It also provides computational libraries and zero-copy
 streaming messaging and inter-process communication.
 
+### A note about FlightSQL drivers
+
+Go FlightSQL drivers live in the
+[ADBC repository](https://github.com/apache/arrow-adbc/tree/main/go/adbc).
+In particular, to use the Golang `database/sql` interface:
```
zeroshade, line 35: we should probably document that you can pass multiple options by separating them with `;` and then point to the ADBC documentation (https://arrow.apache.org/adbc/main/driver/go/flight_sql.html) which contains the available options which can be passed in addition to the uri. such as the user and password, etc.

jacobmarble: Great idea.

PR #34420, opened 2023-03-03
GH-34386: [C++] Add a PathFromUriOrPath method

### LINKED ISSUE #34386: [C++] Handle URI paths for filesystems?

Given a URI like `s3://bucket/foo/bar.parquet`, right now, you can turn that URI into a FileSystem instance + a non-URI path. That is useful, but then the FileSystem APIs don't work with URIs. For instance, it's explicitly rejected here: 

https://github.com/apache/arrow/blob/0844924942042619d3e04652c7a2c150850167af/cpp/src/arrow/filesystem/s3fs.cc#L416-L419

So if you have a set of URIs and want to work with them with filesystems, you have to parse the URIs yourself. Given we already do the URI munging internally, it might be nice for FileSystem instances to expose a method to parse URIs into a path as well (raising an error if the URI is incompatible).

### Component(s)

C++

westonpace:
### Rationale for this change

We have some URI parsing indirectly exposed through FilesystemFromUri.  However, this isn't very useful if the user has multiple URIs or if they already have a filesystem.  This method allows that same URI handling to be used even if the user already has a filesystem.

Adds a new arrow::fs::PathFromUriOrPath method

### Are these changes tested?

Yes, via unit tests

There is a new API method but no changes to any existing APIs
* Closes: #34386

### DIFF
--- a/cpp/src/arrow/filesystem/filesystem.cc
+++ b/cpp/src/arrow/filesystem/filesystem.cc
@@ -759,7 +759,7 @@ Result<std::shared_ptr<FileSystem>> FileSystemFromUriOrPath(
   if (internal::DetectAbsolutePath(uri_string)) {
     // Normalize path separators
     if (out_path != nullptr) {
-      *out_path = ToSlashes(uri_string);
+      *out_path = std::string(RemoveTrailingSlash(ToSlashes(uri_string)));
     }
     return std::make_shared<LocalFileSystem>();
   }
--- a/cpp/src/arrow/filesystem/filesystem.cc
+++ b/cpp/src/arrow/filesystem/filesystem.cc
@@ -759,7 +759,8 @@ Result<std::shared_ptr<FileSystem>> FileSystemFromUriOrPath(
   if (internal::DetectAbsolutePath(uri_string)) {
     // Normalize path separators
     if (out_path != nullptr) {
-      *out_path = std::string(RemoveTrailingSlash(ToSlashes(uri_string)));
+      *out_path =
+          std::string(RemoveTrailingSlash(ToSlashes(uri_string), /*preserve_root=*/true));
     }
     return std::make_shared<LocalFileSystem>();
   }
--- a/cpp/src/arrow/filesystem/filesystem.cc
+++ b/cpp/src/arrow/filesystem/filesystem.cc
@@ -738,7 +738,56 @@ Result<std::shared_ptr<FileSystem>> FileSystemFromUriReal(const Uri& uri,
   return Status::Invalid("Unrecognized filesystem type in URI: ", uri_string);
 }
 
+Status CheckFileSystem(const FileSystem* filesystem,
+                       const std::string& expected_type_name, const std::string& uri) {
+  if (filesystem && filesystem->type_name() != expected_type_name) {
+    return Status::Invalid("URI ", uri, " is not compatible with filesystem of type ",
+                           filesystem->type_name());
+  }
+  return Status::OK();
+}
+
 }  // namespace
+
+Result<std::string> PathFromUriOrPath(const FileSystem* filesystem,
+                                      const std::string& uri_string) {
+  if (internal::DetectAbsolutePath(uri_string)) {
+    RETURN_NOT_OK(CheckFileSystem(filesystem, "local", uri_string));
+    // Normalize path separators
+    return ToSlashes(uri_string);
+  }
+
+  ARROW_ASSIGN_OR_RAISE(auto uri, ParseFileSystemUri(uri_string));
+  const auto scheme = uri.scheme();
+  std::string path;
+  if (scheme == "file") {
+    RETURN_NOT_OK(CheckFileSystem(filesystem, "local", uri_string));
+    RETURN_NOT_OK(LocalFileSystemOptions::FromUri(uri, &path));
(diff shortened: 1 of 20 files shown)


lidavidm:
Oh neat, thank you. I'll try to take a look at this soon

cpp/src/arrow/filesystem/filesystem.h
```diff
@@ -476,6 +476,25 @@ Result<std::shared_ptr<FileSystem>> FileSystemFromUri(const std::string& uri,
...
+/// URI.  If that behavior is not desired then NULLPTR can be specified for the
+/// filesystem and that check will be skipped.
+///
+/// uri can be an absolute path instead of a URI.  In that case it will ensure the
+/// filesystem (if supplied) is the local filesystem and will normalize the path's
+/// file separators.
+///
+/// \return The path inside the filesystem that is indicated by the URI.
+ARROW_EXPORT
+Result<std::string> PathFromUriOrPath(const FileSystem* filesystem,
```
lidavidm, line 495: nit: `const FileSystem&`?

cpp/src/arrow/filesystem/filesystem.cc
```diff
@@ -738,8 +738,57 @@ Result<std::shared_ptr<FileSystem>> FileSystemFromUriReal(const Uri& uri,
...
+  if (filesystem && filesystem->type_name() != expected_type_name) {
+    return Status::Invalid("URI ", uri, " is not compatible with filesystem of type ",
+                           filesystem->type_name());
+  }
+  return Status::OK();
+}
+
 }  // namespace
 
+Result<std::string> PathFromUriOrPath(const FileSystem* filesystem,
```
lidavidm, line 752: I suppose this feels like it should be a virtual method on FileSystem to me, still. (Probably with a default implementation that returns NotImplemented.) It would avoid the ifdef tower, and would keep FileSystem as an actual extension point, avoiding a hardcoded list that we have to maintain inside of Arrow. It effectively already is implemented as a (static) method, too.

Then again, other methods above have this same pattern, so I suppose it's consistent.

cpp/src/arrow/filesystem/filesystem.cc
```diff
@@ -738,8 +738,57 @@ Result<std::shared_ptr<FileSystem>> FileSystemFromUriReal(const Uri& uri,
...
+#else
+    return Status::NotImplemented("Got GCS URI but Arrow compiled without GCS support");
+#endif
+  } else if (scheme == "hdfs" || scheme == "viewfs") {
+    RETURN_NOT_OK(CheckFileSystem(filesystem, "hdfs", uri_string));
+    path = uri.path();
+  } else if (scheme == "s3") {
+    RETURN_NOT_OK(CheckFileSystem(filesystem, "s3", uri_string));
+#ifdef ARROW_S3
+    RETURN_NOT_OK(S3Options::FromUri(uri, &path));
```
lidavidm, line 779: The other potential problem is, what if the URI specifies a filesystem that isn't compatible beyond just the type? For instance, we can stuff credentials, endpoint overrides, etc. into the URI - this won't detect that (that may be fine so long as it's clear what "compatible" means).

In that case, though, maybe it's best to have this function do nothing other than turn the URI into a path, and let the application deal with any other logic.

cpp/src/arrow/filesystem/filesystem.cc
westonpace: Ah, can't move the existing methods because you don't have a filesystem at that point.  I'll update this one to be a virtual method.

lidavidm:
Not sure why this was unceremoniously closed. Weston, do you think you would still have time to look at this? It would be useful in https://github.com/apache/arrow/pull/35034

westonpace:
Yes, I will try and look at this.  I think it still needs some changes.

westonpace:
I unified the various paths using a helper method in util_internal.cc.  This highlighted one small difference between the various implementations.  The s3 and gcs filesystems remove any trailing slash in the path while the hdfs and local filesystems do not.  I have opened #35399 for any discussion on that difference but, for the moment, I have changed all the filesystems to always remove the trailing slash for consistency.

lidavidm (approved):

lidavidm:
Hmm, I suppose the trailing slash could change listing behavior in S3.

westonpace:
> Hmm, I suppose the trailing slash could change listing behavior in S3.

S3 was already removing the trailing slash so there is no change to the S3 behavior.  This only affected the local filesystem.

lidavidm:
Ah, I got that mixed up. LGTM then

cpp/src/arrow/filesystem/path_util.cc
```diff
@@ -129,7 +129,17 @@ std::string EnsureLeadingSlash(std::string_view v) {
...
+  if (preserve_root && key.size() == 1) {
+    // If the user gives us "/" then don't return ""
+    return key;
+  }
+#ifdef _WIN32
+  if (preserve_root && key.size() == 3 && key[1] == ':' && key[0] != '/') {
+    // If the user gives us C:/ then don't return C:
+    return key;
+  }
+#endif
```
lidavidm, line 142: Oh fun.

lidavidm (approved):

Issue #32609, opened 2022-08-08
[Python] Type checking support


asfimport:
# mypy and static type checking

As of Python3.6, it has been possible to introduce typing information in the code. This became immensely popular in a short period of time. Shortly after, the tool `mypy` arrived and this has become the industry standard for static type checking inside Python. It is able to check very quickly for invalid types which makes it possible to serve as a pre-commit. It has raised many bugs that I did not see myself and has been a very valuable tool.
## Now what does this mean for PyArrow?

When we run mypy on code that uses PyArrow, you will get error message as follows:

```
some_util_using_pyarrow/hdfs_utils.py:5: error: Skipping analyzing "pyarrow": module is installed, but missing library stubs or py.typed marker
some_util_using_pyarrow/hdfs_utils.py:9: error: Skipping analyzing "pyarrow": module is installed, but missing library stubs or py.typed marker
some_util_using_pyarrow/hdfs_utils.py:11: error: Skipping analyzing "pyarrow.fs": module is installed, but missing library stubs or py.typed marker
```

More information is available here: <https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-library-stubs-or-py-typed-marker>
## You can solve this in three ways:
1. Ignore the message. This, however, will put all types from PyArrow to `Any`, making it unable to find user errors with the PyArrow library
1. Create a Python stub file. This is what previously used to be the standard, however, it no longer a popular option. This is because stubs are extra, next to the source code, while you can also inline the code with type hints, which brings me to our third option.
1. Create a `py.typed` file and use inline type hints. This is the most popular option today because it requires no extra files (except for the py.typed file), allows all the type hints to be with the code (like now in the documentation) and not only provides your users but also the developers of the library themselves with type hints (and hinting of issues inside your IDE).
   
    
   
   My personal opinion already shines through the options, it is 3 as this has shortly become the industry standard since the introduction.
   ## What should we do?
   
   I'd very much like to work on this, however, I don't feel like wasting time. Therefore, I am raising this ticket to see if this had been considered before or if we just didn't get to this yet.
   
   I'd like to open the discussion here:
1. Do you agree with number #3 as type hints.
1. Should we remove the documentation annotations for the type hints given they will be inside the functions? Or should we keep it and specify it in the code? Which would make it double.
   
    

**Reporter**: [Jorrick Sleijster](https://issues.apache.org/jira/browse/ARROW-17335)

[...]

PR #34429, opened 2023-03-03
GH-34428: [Python][Docs] Add docsstring for `make_fragment`

### LINKED ISSUE #34428: [Python] Add missing docs on `make_fragement`

I had to go to the code to see what we could pass into the `make_fragments`

### Component(s)

Python

Fokko:
* Closes: #34428

### DIFF
--- a/python/pyarrow/_dataset.pyx
+++ b/python/pyarrow/_dataset.pyx
@@ -892,7 +892,8 @@ cdef class FileFormat(_Weakrefable):
             If `filesystem` is given, `file` must be a string and specifies
             the path of the file to read from the filesystem.
         partition_expression : Expression, optional
-            The filter expression.
+            An expression that is guaranteed true for all rows in the fragment.  Allows
+            fragment to be potentially skipped while scanning with a filter.
 
         Returns
         -------
--- a/python/pyarrow/_dataset.pyx
+++ b/python/pyarrow/_dataset.pyx
@@ -891,8 +891,13 @@ cdef class FileFormat(_Weakrefable):
         filesystem : Filesystem, optional
             If `filesystem` is given, `file` must be a string and specifies
             the path of the file to read from the filesystem.
-        partition_expression : Expression
+        partition_expression : Expression, optional
             The filter expression.
+
+        Returns
+        -------
+        fragment : Fragment
+            The file fragment
         """
         if partition_expression is None:
             partition_expression = _true
--- a/python/pyarrow/_dataset_parquet.pyx
+++ b/python/pyarrow/_dataset_parquet.pyx
@@ -223,7 +223,7 @@ cdef class ParquetFileFormat(FileFormat):
             An expression that is guaranteed true for all rows in the fragment.  Allows
             fragment to be potentially skipped while scanning with a filter.
         row_groups : Iterable, optional
-            The row groups to include
+            The indices of the row groups to include
 
         Returns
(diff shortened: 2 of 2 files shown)


python/pyarrow/_dataset_parquet.pyx
```diff
@@ -209,6 +209,26 @@ cdef class ParquetFileFormat(FileFormat):
...
+
+        Parameters
+        ----------
+        file : file-like object, path-like or str
+            The file or file path to make a fragment from.
+        filesystem : Filesystem, optional
+            If `filesystem` is given, `file` must be a string and specifies
+            the path of the file to read from the filesystem.
+        partition_expression : Expression, optional
+            The filter expression.
```
westonpace, line 223:
```suggestion
            An expression that is guaranteed true for all rows in the fragment.  Allows
            fragment to be potentially skipped while scanning with a filter.
```

python/pyarrow/_dataset_parquet.pyx
```diff
@@ -209,6 +209,26 @@ cdef class ParquetFileFormat(FileFormat):
...
+        row_groups : Iterable, optional
```
westonpace, line 224: Iterable of what?  I assume numbers (I don't know if specify the iterable type elsewhere).  Also, maybe "The indices of the row groups to include"

Can this be empty?

westonpace (commented):
Thanks for the continuing cleanup.

python/pyarrow/_dataset_parquet.pyx
Fokko: Thanks!

python/pyarrow/_dataset_parquet.pyx
```diff
@@ -209,6 +209,26 @@ cdef class ParquetFileFormat(FileFormat):
...
+            The row groups to include
```
Fokko, line 226:
```suggestion
            The indices of the row groups to include
```

python/pyarrow/_dataset_parquet.pyx
Fokko: Thanks @westonpace. I'd love to add types https://github.com/apache/arrow/issues/32609 but right now it is just guessing indeed.

I just checked, and an empty list `[]` is allowed, and will not ready any row groups, and return an empty table. Setting it to `None` will return all the row groups.

python/pyarrow/_dataset.pyx
```diff
@@ -891,8 +891,13 @@ cdef class FileFormat(_Weakrefable):
         filesystem : Filesystem, optional
             If `filesystem` is given, `file` must be a string and specifies
             the path of the file to read from the filesystem.
-        partition_expression : Expression
+        partition_expression : Expression, optional
             The filter expression.
```
westonpace, line 895:
```suggestion
            An expression that is guaranteed true for all rows in the fragment.  Allows
            fragment to be potentially skipped while scanning with a filter.
```

westonpace (approved):
Sorry, didn't realize there were two versions.  Same clarification on `partition_expression`.  Once that is in I think this is good to go.

python/pyarrow/_dataset.pyx
Fokko: Ah, I missed that one as well. Thanks!

Fokko:
@westonpace Good to go!

PR #34435, opened 2023-03-03
GH-34410: [Python] Allow chunk sizes larger than the default to be used

### LINKED ISSUE #34410: [Python][C++] No longer possible to specify higher chunksize than the default for Parquet writing

See https://github.com/apache/arrow/issues/34374#issuecomment-1449926603 for context

https://github.com/apache/arrow/pull/34281 changed the default row group size (`chunksize` in the C++ WriteTable API). However, that PR changed `DEFAULT_MAX_ROW_GROUP_LENGTH`, which doesn't set the _default_ `chunksize`, but actually caps the max chunk (row group) size regardless of the user-specified `chunksize`.

It seems this constant is used both for setting this max upper cap, and as the default values for `chunk_size` in `WriteTable`. I assume we will have to distinguish those two meanings.

### Component(s)

C++, Parquet, Python

westonpace:
### Rationale for this change

We changed the default chunk size from 64Mi rows to 1Mi rows.  However, it turns out that this property was being treated not just as the default but also as the absolute max.  So it was no longer possible to specify chunk sizes larger than 1Mi rows.  This change separates those two things and restores the max to 64Mi rows.

Pyarrow will now set the `ParquetWriterProperties::max_row_group_length` to 64Mi when constructing a parquet writer.

### Are these changes tested?

Yes.  Unit tests are added.

No.  The previous change #34281 changed two defaults (absolute max and default).  This PR restores the absolute max back to what it was before.  So it is removing a user-facing change.
* Closes: #34410

### DIFF
--- a/python/pyarrow/_parquet.pyx
+++ b/python/pyarrow/_parquet.pyx
@@ -1597,6 +1597,15 @@ cdef shared_ptr[WriterProperties] _create_writer_properties(
         props.encryption(
             (<FileEncryptionProperties>encryption_properties).unwrap())
 
+    # For backwards compatibility reasons we cap the maximum row group size
+    # at 64Mi rows.  This could be changed in the future, though it would be
+    # a breaking change.
+    #
+    # The user can always specify a smaller row group size (and the default
+    # is smaller) when calling write_table.  If the call to write_table uses
+    # a size larger than this then it will be latched to this value.
+    props.max_row_group_length(64*1024*1024)
+
     properties = props.build()
 
     return properties
--- a/python/pyarrow/tests/parquet/test_parquet_writer.py
+++ b/python/pyarrow/tests/parquet/test_parquet_writer.py
@@ -24,7 +24,8 @@ from pyarrow.tests.parquet.common import parametrize_legacy_dataset
 
 try:
     import pyarrow.parquet as pq
-    from pyarrow.tests.parquet.common import _read_table, _test_dataframe
+    from pyarrow.tests.parquet.common import (_read_table, _test_dataframe,
+                                              _range_integers)
 except ImportError:
     pq = None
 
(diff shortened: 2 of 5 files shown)


python/pyarrow/tests/parquet/test_parquet_writer.py
```diff
@@ -202,6 +203,39 @@ def test_parquet_writer_write_wrappers(tempdir, filesystem):
     tm.assert_frame_equal(result, df)
 
 
+@pytest.mark.pandas
```
wjones127, line 206: Perhaps also:

```suggestion
@pytest.mark.large_memory
@pytest.mark.pandas
```

wjones127 (changes requested):
Overall looks good. One suggestion.

python/pyarrow/tests/parquet/test_parquet_writer.py
westonpace: Good idea

wjones127 (approved):

python/pyarrow/_parquet.pyx
```diff
@@ -1597,6 +1597,15 @@ cdef shared_ptr[WriterProperties] _create_writer_properties(
...
             (<FileEncryptionProperties>encryption_properties).unwrap())
 
+    # For backwards compatibility reasons we cap the maximum row group size
+    # at 64Mi rows.  This could be changed in the future, though it would be
+    # a breaking change.
+    #
+    # The user can always specify a smaller row group size (and the default
+    # is smaller) when calling write_table.  If the call to write_table uses
+    # a size larger than this then it will be latched to this value.
+    props.max_row_group_length(64*1024*1024)
```
mapleFU, line 1607: does this causing the bug in https://github.com/apache/arrow/issues/35859#issuecomment-1583614922 ?

PR #34445, opened 2023-03-03
GH-34283 [Python] Add types_mapper support to index for to_pandas

### LINKED ISSUE #34283: [Python] `Table.to_pandas` fails to convert index `dtype` with a custom type mapper

When providing a custom type mapper to `pyarrow.Table.to_pandas`, column dtypes are converted, but not index dtypes.

Example:

```python
import pyarrow as pa
import pandas as pd

df = pd.DataFrame({"s1": ["a", "c", "b"], "i1": [1, 2, 3], "s2": ["d", "e", "f"]}).set_index('s2')
tbl = pa.Table.from_pandas(df)
mapper = {pa.string(): pd.StringDtype("pyarrow")}.get
df2 = tbl.to_pandas(types_mapper=mapper)
print(f"{df2.dtypes}")
print(f"{df2.index.dtype = }")
```

This prints:

```
s1    string[pyarrow]
i1              int64
dtype: object
df2.index.dtype = dtype('O')
```

Column `s1` was mapped to `string[pyarrow]` dtype, however, the index dtype remained `object`. I'd expect the type mapper being used to convert the index `dtype` as well.

Platform: macOS
Pyarrow version: 11.0.0
Build: py310h89f3c6b_2_cpu
Channel: conda-forge

### Component(s)

Python

phofl:

Only respects types_mapper for indexes as well

### Are these changes tested?

Yes

Technically this breaks the API in a way that we would now respect the types_mapper for the index.

- [x] closes #34283

cc @jorisvandenbossche

### DIFF
--- a/python/pyarrow/tests/test_pandas.py
+++ b/python/pyarrow/tests/test_pandas.py
@@ -647,23 +647,6 @@ class TestConvertMetadata:
             result = new_table.to_pandas()
             tm.assert_frame_equal(result, expected)
 
-    @pytest.mark.parametrize("index", ["a", ["a", "b"]])
-    def test_to_pandas_types_mapper_index(self, index):
-        if Version(pd.__version__) < Version("1.5.0"):
-            pytest.skip("ArrowDtype missing")
-        df = pd.DataFrame(
-            {
-                "a": [1, 2],
-                "b": [3, 4],
-                "c": [5, 6],
-            },
-            dtype=pd.ArrowDtype(pa.int64()),
-        ).set_index(index)
-        expected = df.copy()
-        table = pa.table(df)
-        result = table.to_pandas(types_mapper=pd.ArrowDtype)
-        tm.assert_frame_equal(result, expected)
-
 
 class TestConvertPrimitiveTypes:
     """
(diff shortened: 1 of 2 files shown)


AlenkaF (commented):
Thank you for another fix @phofl! 🤩
The solution is much nicer than I thought it would be. And I only have one small nit question in the tests.

python/pyarrow/tests/test_pandas.py
```diff
@@ -647,6 +647,23 @@ def test_mismatch_metadata_schema(self):
...
+            pytest.skip("ArrowDtype missing")
+        df = pd.DataFrame(
+            {
+                "a": [1, 2],
+                "b": [3, 4],
+                "c": [5, 6],
+            },
+            dtype=pd.ArrowDtype(pa.int64()),
+        ).set_index(index)
+        expected = df.copy()
```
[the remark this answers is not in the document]
phofl: Thx! 

Generally in pandas we add copies when expected equals the input. Theoretically you could accidentally modify df in your operation and the test should fail in this case. Copy ensures this

python/pyarrow/tests/test_pandas.py
```diff
@@ -647,6 +647,23 @@ def test_mismatch_metadata_schema(self):
             result = new_table.to_pandas()
             tm.assert_frame_equal(result, expected)
 
+    @pytest.mark.parametrize("index", ["a", ["a", "b"]])
+    def test_to_pandas_types_mapper_index(self, index):
```
jorisvandenbossche, line 651: The other types_mapper related tests are lower in the file (not in one of the classes), so I would move it there (eg just below `test_roundtrip_empty_table_with_extension_dtype_index`)

(Sidenote, seeing that `test_roundtrip_empty_table_with_extension_dtype_index` test, that seems to be something that should be fixed as well, preserving interval index)

jorisvandenbossche (approved):
Looks good!

AlenkaF (approved):

python/pyarrow/tests/test_pandas.py
phofl: Moved

jorisvandenbossche:
Thanks @phofl !

Issue #33999, opened 2023-02-02
[Go] array.String.Value is unsafe


lidavidm:
### Describe the bug, including details regarding any error messages, version, and platform.

String.Value slices a backing string:

```go
func (a *String) Value(i int) string {
	i = i + a.array.data.offset
	return a.values[a.offsets[i]:a.offsets[i+1]]
}
```

Which points directly into the buffer data:

```go
	if vdata := data.buffers[2]; vdata != nil {
		b := vdata.Bytes()
		a.values = *(*string)(unsafe.Pointer(&b))
	}
```

This seems OK when the buffer is Go-allocated since the GC will take care of it (regardless of Release/Retain). But when the buffer is C-allocated (e.g. C Data Interface, mallocator), if the string outlives the array, we have a dangling pointer!

I believe we should copy the string (and frankly, get rid of the unsafe here) to be safe (which also has the side effect that getting a single value out of an array will no longer keep the entire backing array alive). If the extra garbage is an issue, we should consider some sort of string view type.

### Component(s)

Go

rtadepalli:
take

PR #34450, opened 2023-03-04
GH-33999: [Go] Removed  cast from byte[] to string and copied entire string when `Value()` is called

### LINKED ISSUE #68: [Go] array.String.Value is unsafe

String.Value slices a backing string:

```go
func (a *String) Value(i int) string {
	i = i + a.array.data.offset
	return a.values[a.offsets[i]:a.offsets[i+1]]
}
```

Which points directly into the buffer data:

```go
	if vdata := data.buffers[2]; vdata != nil {
		b := vdata.Bytes()
		a.values = *(*string)(unsafe.Pointer(&b))
	}
```

This seems OK when the buffer is Go-allocated since the GC will take care of it (regardless of Release/Retain). But when the buffer is C-allocated (e.g. C Data Interface, mallocator), if the string outlives the array, we have a dangling pointer!

I believe we should copy the string (and frankly, get rid of the unsafe here) to be safe (which also has the side effect that getting a single value out of an array will no longer keep the entire backing array alive). If the extra garbage is an issue, we should consider some sort of string view type.

### Component(s)

Go

rtadepalli:
### Rationale for this change
Remove some unsafe casts and copy a string instead of returning a reference to an underlying array.

Remove `unsafe.Pointer` and add a string deep copy.

### Are these changes tested?
I _think_ existing tests should suffice.

There are no user facing/breaking changes.
* Closes: apache/arrow-go#68
* GitHub Issue: #68

### DIFF
--- a/go/arrow/array/string.go
+++ b/go/arrow/array/string.go
@@ -50,8 +50,11 @@ func (a *String) Reset(data arrow.ArrayData) {
 
 // Value returns the slice at index i. This value should not be mutated.
 func (a *String) Value(i int) string {
+	sb := strings.Builder{}
+
 	i = i + a.array.data.offset
-	return a.values[a.offsets[i]:a.offsets[i+1]]
+	sb.WriteString(a.values[a.offsets[i]:a.offsets[i+1]])
+	return sb.String()
 }
 
 // ValueOffset returns the offset of the value at index i.
@@ -246,7 +249,7 @@ func (a *LargeString) setData(data *Data) {
 
 	if vdata := data.buffers[2]; vdata != nil {
 		b := vdata.Bytes()
-		a.values = *(*string)(unsafe.Pointer(&b))
+		a.values = string(b[:])
 	}
 
 	if offsets := data.buffers[1]; offsets != nil {


go/arrow/array/string.go
```diff
@@ -50,8 +50,11 @@ func (a *String) Reset(data arrow.ArrayData) {
 
 // Value returns the slice at index i. This value should not be mutated.
 func (a *String) Value(i int) string {
+	sb := strings.Builder{}
+
 	i = i + a.array.data.offset
-	return a.values[a.offsets[i]:a.offsets[i+1]]
+	sb.WriteString(a.values[a.offsets[i]:a.offsets[i+1]])
+	return sb.String()
```
zeroshade, line 57: There's no need to use the string builder here, you can just do `string(a.values[a.offsets[i]:a.offsets[i+1]])`

go/arrow/array/string.go
```diff
@@ -246,7 +249,7 @@ func (a *LargeString) setData(data *Data) {
 
 	if vdata := data.buffers[2]; vdata != nil {
 		b := vdata.Bytes()
-		a.values = *(*string)(unsafe.Pointer(&b))
+		a.values = string(b[:])
```
zeroshade, line 252: we shouldn't copy the entire buffer here. Keep the `unsafe.Pointer` here so we only perform copies in the `Value` method and only on the values that are requested.

amol-:

PR #34454, opened 2023-03-04
GH-34453: [Go] Support Builders for user defined extensions

### LINKED ISSUE #34453: [Go] Extension Builder Interface

(This was also send in the mailing list and discussed shortly with @zeroshade ). Copy of what was sent in the mailing list and a PR will quickly follow:

Hopefully this is the right place to ask. As some background I'm [Yevgeny Pats](https://www.linkedin.com/in/yevgeny-pats-5973328b/), Founder @ [CloudQuery](https://github.com/cloudquery/cloudquery) . We are very interested in migrating our protocol and Go type system to Apache Arrow. Extensions are a critical part for us and thus I've the following questions on whether it's a usage problem on my end or something that is not yet available. I'll give here an example for Go but I believe the same issue exists in all libraries/languages.

Here is a [public github gist](https://gist.github.com/yevgenypats/6969e8e598161fc2021612c780bba3eb).

What are the problems:

- The problems are around the abstraction for the extension types. While I understand that the underlying storage needs to be supported in the library we don't have a way for extensions to provide its own builder which means the user needs to know how the extension type stores the type inside the binary. This creates a leaky abstraction and the need for various helper functions like `UUIDToBinary`
- The other way is fine as you can have methods like ToUUID on top of the extension array. But this creates asymmetry in the abstraction. 
- Because we don't control the builder for extensions this cripples into other places like [json](https://github.com/apache/arrow/issues/34292#issuecomment-1446653210) and csv where we can't control marshalling (in the same way we control all other built-in types). So basically for extensions that use binary type as underlying storage in case of json and csv those will always be encoded as base64 which is not very useful (think about uuid, ip address, mac address).

The main point is that I think the right abstraction for extensions should provide all the apis (type, array, builder) just like built-in types, otherwise the abstraction is incomplete or "leaky". Of course we can still have limitations like the custom builder must use an underlying known storage (for it to work over ipc) but it can still control various other types like marshaling, unmarshaling, building, and so on.

Hopefully this gives enough context but would love to elaborate.

Thanks,
Yevgeny

### Component(s)

Go

yevgenypats:
This should serve as discussion as it's a medium change but this should Close https://github.com/apache/arrow/issues/34453 and give users the ability to define custom Builder for their extensions just like they define ExtensionTypes and ExtensionArrays
* Closes: #34453

### DIFF
--- a/go/arrow/array/extension_builder.go
+++ b/go/arrow/array/extension_builder.go
@@ -16,6 +16,8 @@
 
 package array
 
+// ExtensionBuilderWrapper is an interface that you need to implement in your custom extension type if you want to provide a customer builder as well.
+// See example in ./arrow/internal/testing/types/extension_types.go
 type ExtensionBuilderWrapper interface {
 	NewBuilder(bldr *ExtensionBuilder) Builder
 }
--- a/go/arrow/array/extension_builder.go
+++ b/go/arrow/array/extension_builder.go
@@ -1,3 +1,19 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package array
 
 type ExtensionBuilderWrapper interface {
--- a/go/arrow/array/extension_builder.go
+++ b/go/arrow/array/extension_builder.go
@@ -1,0 +1,5 @@
+package array
+
+type ExtensionBuilderWrapper interface {
+	NewBuilder(bldr *ExtensionBuilder) Builder
+}
--- a/go/arrow/datatype_extension.go
+++ b/go/arrow/datatype_extension.go
@@ -20,8 +20,6 @@ import (
 	"fmt"
 	"reflect"
 	"sync"
-
-	"github.com/apache/arrow/go/v12/arrow/memory"
 )
 
 var (
@@ -124,20 +122,16 @@ type ExtensionType interface {
 	// If the storage type is incorrect or something else is invalid with the data this should
 	// return nil and an appropriate error.
 	Deserialize(storageType DataType, data string) (ExtensionType, error)
-	// this should return array.Builder interface but we cannot import due to cycle import, so we use 
-	// interface{} instead. At least for 
-	NewBuilder(mem memory.Allocator, dt ExtensionType) interface{}
 	mustEmbedExtensionBase()
 }
 
 // ExtensionBase is the base struct for user-defined Extension Types which must be
 // embedded in any user-defined types like so:
 //
-//     type UserDefinedType struct {
-//         arrow.ExtensionBase
-//         // any other data
-//     }
-//
+//	type UserDefinedType struct {
+//	    arrow.ExtensionBase
+//	    // any other data
+//	}
 type ExtensionBase struct {
 	// Storage is the underlying storage type
 	Storage DataType
@@ -166,13 +160,10 @@ func (e *ExtensionBase) Fields() []Field {
 	return nil
 }
 
-func (e *ExtensionBase) NewBuilder(mem memory.Allocator, dt ExtensionType) interface{} {
-	return nil
-}
-
 func (e *ExtensionBase) Layout() DataTypeLayout { return e.Storage.Layout() }
 
 // this no-op exists to ensure that this type must be embedded in any user-defined extension type.
+//
 //lint:ignore U1000 this function is intentionally unused as it only exists to ensure embedding happens
 func (ExtensionBase) mustEmbedExtensionBase() {}
 
--- a/go/arrow/datatype_extension.go
+++ b/go/arrow/datatype_extension.go
@@ -122,7 +122,7 @@ type ExtensionType interface {
 	// If the storage type is incorrect or something else is invalid with the data this should
 	// return nil and an appropriate error.
 	Deserialize(storageType DataType, data string) (ExtensionType, error)
-	
+
 	mustEmbedExtensionBase()
 }
 
--- a/go/arrow/datatype_extension.go
+++ b/go/arrow/datatype_extension.go
@@ -122,6 +122,7 @@ type ExtensionType interface {
 	// If the storage type is incorrect or something else is invalid with the data this should
 	// return nil and an appropriate error.
 	Deserialize(storageType DataType, data string) (ExtensionType, error)
+	
 	mustEmbedExtensionBase()
 }
 
--- a/go/arrow/datatype_extension.go
+++ b/go/arrow/datatype_extension.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"reflect"
 	"sync"
+
+	"github.com/apache/arrow/go/v12/arrow/memory"
 )
 
 var (
@@ -122,7 +124,9 @@ type ExtensionType interface {
 	// If the storage type is incorrect or something else is invalid with the data this should
 	// return nil and an appropriate error.
 	Deserialize(storageType DataType, data string) (ExtensionType, error)
-
+	// this should return array.Builder interface but we cannot import due to cycle import, so we use 
+	// interface{} instead. At least for 
+	NewBuilder(mem memory.Allocator, dt ExtensionType) interface{}
 	mustEmbedExtensionBase()
 }
 
@@ -159,6 +163,10 @@ func (e *ExtensionBase) Fields() []Field {
 	if nested, ok := e.Storage.(NestedType); ok {
 		return nested.Fields()
 	}
+	return nil
+}
+
+func (e *ExtensionBase) NewBuilder(mem memory.Allocator, dt ExtensionType) interface{} {
 	return nil
 }
 
--- a/go/arrow/internal/testing/types/extension_test.go
+++ b/go/arrow/internal/testing/types/extension_test.go
@@ -33,16 +33,21 @@ import (
 var testUUID = uuid.New()
 
 func TestExtensionBuilder(t *testing.T) {
-	extBuilder := array.NewExtensionBuilder(memory.DefaultAllocator, types.NewUUIDType())
+	mem := memory.NewCheckedAllocator(memory.DefaultAllocator)
+	defer mem.AssertSize(t, 0)
+	extBuilder := array.NewExtensionBuilder(mem, types.NewUUIDType())
+	defer extBuilder.Release()
 	builder := types.NewUUIDBuilder(extBuilder)
 	builder.Append(testUUID)
 	arr := builder.NewArray()
+	defer arr.Release()
 	arrStr := arr.String()
 	assert.Equal(t, "[\""+testUUID.String()+"\"]", arrStr)
 	jsonStr, err := json.Marshal(arr)
 	assert.NoError(t, err)
 
-	arr1, _, err := array.FromJSON(memory.DefaultAllocator, types.NewUUIDType(), bytes.NewReader(jsonStr))
+	arr1, _, err := array.FromJSON(mem, types.NewUUIDType(), bytes.NewReader(jsonStr))
+	defer arr1.Release()
 	assert.NoError(t, err)
 	assert.Equal(t, arr, arr1)
 }
--- a/go/arrow/internal/testing/types/extension_test.go
+++ b/go/arrow/internal/testing/types/extension_test.go
@@ -33,7 +33,8 @@ import (
 var testUUID = uuid.New()
 
 func TestExtensionBuilder(t *testing.T) {
-	builder := types.NewUUIDBuilder(memory.DefaultAllocator, types.NewUUIDType())
+	extBuilder := array.NewExtensionBuilder(memory.DefaultAllocator, types.NewUUIDType())
+	builder := types.NewUUIDBuilder(extBuilder)
 	builder.Append(testUUID)
 	arr := builder.NewArray()
 	arrStr := arr.String()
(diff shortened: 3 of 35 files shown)


go/arrow/datatype_extension.go
```diff
@@ -122,7 +124,9 @@ type ExtensionType interface {
 	// If the storage type is incorrect or something else is invalid with the data this should
 	// return nil and an appropriate error.
 	Deserialize(storageType DataType, data string) (ExtensionType, error)
-
+	// this should return array.Builder interface but we cannot import due to cycle import, so we use 
+	// interface{} instead. At least for 
+	NewBuilder(mem memory.Allocator, dt ExtensionType) interface{}
```
zeroshade, line 129: should we follow the same pattern as `ArrayType` and have a `BuilderType` method which returns a `reflect.Type` that we use to wrap the `ExtensionBuilder` with? This also avoids the import cycle.

Another thing to consider is that this is going to break any and all existing Extension types in other consumers' codebases. We should probably make a second interface type which contains the `BuilderType` method so that we can just use a type assertion test in `NewBuilder` rather than break existing consumers?

go/arrow/internal/testing/types/extension_types.go
```diff
@@ -18,26 +18,181 @@
...
+func (b *UUIDBuilder) AppendValues(v []uuid.UUID, valid []bool) {
+	data := make([][]byte, len(v))
+	for i, v := range v {
+		data[i] = v[:]
+	}
+	b.ExtensionBuilder.Builder.(*array.FixedSizeBinaryBuilder).AppendValues(data, valid)
+}
+
+func (b *UUIDBuilder) UnmarshalOne(dec *json.Decoder) error {
+	t, err := dec.Token()
```
zeroshade, line 66: Should we have default implementations of the Unmarshal methods for the base `ExtensionBuilder` which just forwards to the storage type?

zeroshade (commented):
Took a look through, thoughts?

go/arrow/datatype_extension.go
yevgenypats: On the first part, we can keep the same pattern, sure. But actually I think this should 1) provide better performance, no reflection. 2) much better type safety, and better developer experience as it is clear what should be the type of this function (I also wanted to return Builder but had to fallback to interface due to some cyclic import that didn't want to fix part of this PR).

Re 2nd thing, not sure I followed, can you share an example of that it's going to break? The only place that we use this function now is [here](https://github.com/apache/arrow/pull/34454/files#diff-9a870facf93f4b8b367b13881398e3918d0627e348079d33017231d2b389d8bcR322)  and I fallback to the previous builder.

go/arrow/internal/testing/types/extension_types.go
yevgenypats: If you don't implement `UnmarshalOne` this should call the underlying ExtensionBuilder `UnmarshalOne`. This part was critical for us because otherwise we can't use any of `Marshal`, `Unmarshal` functions as they will always result `base64` encoded values instead of what is a fit for the extension. For UUID for example we want to unmarshal a uuid or in case of IP Address (a string of ip address while storing this as a binary).

Here is an example of a few types we implemented on top of this PR: https://github.com/cloudquery/filetypes/tree/main/internal/cqarrow (This is just a PoC and will eventually go into our [SDK](https://github.com/cloudquery/plugin-sdk) which is used by all our [plugins](https://github.com/cloudquery/cloudquery) )

yevgenypats:
> Took a look through, thoughts?

@zeroshade  Thanks for the initial review! Commented and also linked to a few example on how we use it.

go/arrow/datatype_extension.go
zeroshade: For the second point: Note that you had to add the `NewBuilder` method to the existing extension types. By adding a new method to the interface, anyone who has their own extension types defined that doesn't already have this method defined will have a compile error when they upgrade to this version of Arrow which adds the method because their structs will no longer meet the interface for `ExtensionType`.

>  But actually I think this should 1) provide better performance, no reflection. 2) much better type safety, and better developer experience as it is clear what should be the type of this function (I also wanted to return Builder but had to fallback to interface due to some cyclic import that didn't want to fix part of this PR).

In general, I wouldn't expect creating a builder to be a performance bottleneck as consumers shouldn't create builders repeatedly. That said, a couple ideas I've had so far:

* Like was previously done for Arrays, we could shift the definition of the `Builder` interface to the `arrow` package directly and then add an alias in the `array` package to ensure we don't break any consumers (with a deprecated message telling people to point at `arrow.Builder` instead.
* If we're going in this direction, rather than passing the allocator and the extension type, we should pass an `ExtensionBuilder` to the method and have this just wrap it and return the wrapped builder. The consumer can also retrieve the extension type and the allocator from the builder directly if they need to. So perhaps something like `WrapBuilder(ExtensionBuilder) Builder`.

go/arrow/internal/testing/types/extension_types.go
zeroshade: Ah, I realized we don't need to manually create default implementations of `UnmarshalOne` etc. because we embed the underlying Storage Builder anyways, so you're right. It will call the underlying one automatically if you don't implement it. All good here then!

go/arrow/datatype_extension.go
yevgenypats: Re breaking, I don't think this will breaking anything as there is a [default](https://github.com/apache/arrow/pull/34454/files#diff-7fe93f07a80f96f7f99153486503543e8bb9443496a71ad078d6c6e9a8a5c39eR169) in the `ExtensionType` that just returns `nil` to keep this backward compatible so if this is not implement I call the old `NewExtensionBuilder` in the `NewBuilder` function

zeroshade (changes requested):
Just request a re-review once this is ready for a new review, thanks.

go/arrow/datatype_extension.go
yevgenypats: @zeroshade can you please take a look at this comment ^ ? (This is why I didn't re-requested a re-review yet as need your guidance here)

yevgenypats: re using `Builder` in the signature I think it might be more complex than it seems (I tried it initially). For example `newData *Data` function which means we need also to take `Data` to `arrow` pacakge. In that case it might be even easier to move `datatype_extension.go` to `array` because we can import from `array` the `arrow` package but not the otherway around (Also, there are even more complications because the interface has private functions which would cause compilation error if we move it to a different package).  Can't think of a much easier way without a big refactor which I think better to avoid atm and just do the runtime check. WDYT ?

zeroshade: Darn, that's really annoying..... I agree it's better to avoid a big refactor atm. 

I still want to change the signature though instead of having consumers have to create both the underlying storage builder & wrap it.

Instead of adding this method to the `ExtensionType` interface, we could just go with an interface defined in the `array` package which would let us use `Builder` in the method definition. something like:

```go
type ExtensionTypeCustomBuilder interface {
    NewBuilder(ExtensionBuilder) Builder
}
```

Then in `builder.go` you can do:

```go
case arrow.EXTENSION:
        typ := dtype.(arrow.ExtensionType)
        bldr := NewExtensionBuilder(mem, typ)
        if custom, ok := typ.(ExtensionTypeCustomBuilder); ok {
               return custom.NewBuilder(bldr)
        }
        return bldr
```

What do you think?

yevgenypats: I like that! Updated.

yevgenypats:
Ready now for another review, though looks like some workflows are failing. I think this is not connected though to this PR

go/arrow/internal/testing/types/extension_types.go
```diff
@@ -18,20 +18,172 @@
...
 
 	"github.com/apache/arrow/go/v12/arrow"
 	"github.com/apache/arrow/go/v12/arrow/array"
+	"github.com/google/uuid"
 	"golang.org/x/xerrors"
 )
 
+type UUIDBuilder struct {
+	*array.ExtensionBuilder
+	dtype *UUIDType
```
zeroshade, line 37: the failures are because this field is unused here

yevgenypats: Oh ok. Fixed. I wasn't able to find it from the action. Where should I look for it next time? (Can I run those linters locally as well?)

zeroshade: So that was actually found by the Go build failing, and I saw it by looking at the log file for the failing Go workflows.

You can also run the linters (or most of the workflows) by following the instructions here: https://arrow.apache.org/docs/developers/continuous_integration/archery.html to install the `archery` utility in the repo and then running `archery docker run <job>` as described [here](https://arrow.apache.org/docs/developers/continuous_integration/docker.html).

go/arrow/array/extension_builder.go
```diff
@@ -0,0 +1,21 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package array
+
+type ExtensionBuilderWrapper interface {
```
zeroshade, line 19: can you add a Godoc comment here to explain what this is for and how to use it? It also might be worthwhile to move the entirety of the `ExtensionBuilder` implementation to this file and update the comment on it to explain the functionality you've added as far as for customizing the behavior.

go/arrow/datatype_extension.go
```diff
@@ -122,18 +122,16 @@ type ExtensionType interface {
 	// If the storage type is incorrect or something else is invalid with the data this should
 	// return nil and an appropriate error.
 	Deserialize(storageType DataType, data string) (ExtensionType, error)
-
```
zeroshade, line 125: please put the empty line back here, I'd prefer to keep a single empty line between the public and private method declarations here.

go/arrow/internal/testing/types/extension_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+var testUUID = uuid.New()
+
+func TestExtensionBuilder(t *testing.T) {
+	extBuilder := array.NewExtensionBuilder(memory.DefaultAllocator, types.NewUUIDType())
+	builder := types.NewUUIDBuilder(extBuilder)
```
zeroshade, line 37: use

```go
mem := memory.NewCheckedAllocator(memory.DefaultAllocator)
defer mem.AssertSize(t, 0)
```

to ensure that the memory is handled properly (for example, you're missing the `Release` calls here for the builders and generated arrays

go/arrow/internal/testing/types/extension_types.go
yevgenypats: Sounds good. Thanks, I'll try those for next PR! Tests passing now.

go/arrow/internal/testing/types/extension_test.go
```diff
@@ -0,0 +1,63 @@
...
+	assert.Equal(t, "[\""+testUUID.String()+"\"]", arrStr)
+	jsonStr, err := json.Marshal(arr)
+	assert.NoError(t, err)
+
+	arr1, _, err := array.FromJSON(memory.DefaultAllocator, types.NewUUIDType(), bytes.NewReader(jsonStr))
+	assert.NoError(t, err)
+	assert.Equal(t, arr, arr1)
+}
+
+func TestExtensionRecordBuilder(t *testing.T) {
```
zeroshade, line 50: it would also be probably worthwhile to add a test for the `String()` method of `UUIDArray` , since you did the work to have it print the string representation of the UUID instead of the raw bytes when calling `String()`

go/arrow/internal/testing/types/extension_types.go
```diff
@@ -18,20 +18,171 @@
...
+	return b
+}
+
+func (b *UUIDBuilder) Append(v uuid.UUID) {
+	b.ExtensionBuilder.Builder.(*array.FixedSizeBinaryBuilder).Append(v[:])
+}
+
+func (b *UUIDBuilder) UnsafeAppend(v uuid.UUID) {
+	b.ExtensionBuilder.Builder.(*array.FixedSizeBinaryBuilder).UnsafeAppend(v[:])
+}
```
zeroshade, line 52: is `UnsafeAppend` actually needed here at all?

go/arrow/internal/testing/types/extension_types.go
```diff
@@ -18,20 +18,171 @@
...
+
+func (b *UUIDBuilder) AppendValues(v []uuid.UUID, valid []bool) {
+	data := make([][]byte, len(v))
+	for i, v := range v {
+		data[i] = v[:]
+	}
+	b.ExtensionBuilder.Builder.(*array.FixedSizeBinaryBuilder).AppendValues(data, valid)
```
zeroshade, line 59: this could be done more efficiently if desired via resize + `UnsafeAppend` or other methods. Not necessary for this PR but just something to think about.

go/arrow/internal/testing/types/extension_types.go
```diff
@@ -18,20 +18,171 @@
...
+		if err != nil {
+			return err
+		}
+		val = data
+	case []byte:
+		data, err := uuid.ParseBytes(v)
+		if err != nil {
+			return err
+		}
+		val = data
```
zeroshade, line 81: Is this case actually getting tested by the tests you wrote? Check the test coverage please to confirm. I think it would only go through the String case. I'm not positive though.

zeroshade (changes requested):
A few more nitpicks. We're almost there! Thanks again for all this!

go/arrow/internal/testing/types/extension_test.go
yevgenypats: I think this is tested via the `MarshalJSON` in a way.

go/arrow/internal/testing/types/extension_types.go
yevgenypats: I just copied that from other places in the code but seems if I remove it then all is good.

zeroshade (changes requested):
just a single whitespace nitpick haha. Fix that and i'll merge this :smile:

go/arrow/internal/testing/types/extension_test.go
zeroshade: that works. sure

go/arrow/datatype_extension.go
```diff
@@ -122,18 +122,17 @@ type ExtensionType interface {
...
+
```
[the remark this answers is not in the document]
yevgenypats: Done.

zeroshade (approved):

PR #34458, opened 2023-03-05
GH-34425: [GLib] Add GArrowRankOptions

### LINKED ISSUE #34425: [GLib] Add support for RankOptions
#### Target option

RankOptions

#### Proposed feature

Add support for RankOptions.

#### Impact of this request

It will be used in rank();

```
rank(input, {sort_keys=[], null_placement=AtEnd, tiebreaker=First}):

Compute numerical ranks of an array (1-based)
----
This function computes a rank of the input array.
By default, null values are considered greater than any other value and
are therefore sorted at the end of the input. For floating-point types,
NaNs are considered greater than any other non-null value, but smaller
than null values. The default tiebreaker is to assign ranks in order of
when ties appear in the input.

The handling of nulls, NaNs and tiebreakers can be changed in RankOptions.
```

Currently, only the default option is available from Ruby.

https://arrow.apache.org/docs/cpp/api/compute.html#_CPPv4N5arrow7compute11RankOptionsE

### Component(s)

GLib, Ruby

kou:
### Rationale for this change

It's just missed.

Just add it.

### Are these changes tested?

Yes.

Yes.
* Closes: #34425

### DIFF
--- a/c_glib/arrow-glib/compute.cpp
+++ b/c_glib/arrow-glib/compute.cpp
@@ -104,12 +104,54 @@ namespace {
     return true;
   }
 
-  bool
-  garrow_sort_key_equal_raw(const arrow::compute::SortKey &sort_key,
-                            const arrow::compute::SortKey &other_sort_key) {
-    return
-      (sort_key.target == other_sort_key.target) &&
-      (sort_key.order == other_sort_key.order);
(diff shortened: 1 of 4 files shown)


kou:
+1

PR #34463, opened 2023-03-06
GH-32619: [Python][Docs] Include options for PyArrow build explicitly

### LINKED ISSUE #32619: [Python][Docs] Include options for PyArrow build explicitly
Currently, <https://arrow.apache.org/docs/developers/python.html#build-and-test> suggests setting PYARROW_WITH_$COMPONENT to 1 for every component used in C++. However, this does not make clear whether every single component has a corresponding flag, and also, due to mentioning the PYARROW_PARALLEL flag directly after, makes it unclear whether there are more environment variables that can affect build.

This could be fixed by including an explicit list of relevant variables and arguments for the PyArrow build.

**Reporter**: [Kae Suarez](https://issues.apache.org/jira/browse/ARROW-17345) / @ksuarez1423

<sub>**Note**: *This issue was originally created as [ARROW-17345](https://issues.apache.org/jira/browse/ARROW-17345). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

AlenkaF:

List of relevant build options for PyArrow added to the python dev docs.

This PR also removes these flags from PyArrow:
- `PYARROW_BUNDLE_BOOST`
- `PYARROW_BUNDLE_ARROW_CPP_HEADERS`
- `PYARROW_WITH_STATIC_PARQUET`
- `PYARROW_PARQUET_USE_SHARED`

* Closes: #32619


docs/source/developers/python.rst
```diff
@@ -586,6 +586,86 @@ Caveats
...
+     - Extra CMake and Arrow options (ex. ``"-DARROW_SIMD_LEVEL"``,
+       ``"-DCMAKE_OSX_ARCHITECTURES"``)
+   * - PYARROW_BOOST_NAMESPACE
+     - Name of install folder for boost (if different from 'boost')
+   * - PYARROW_CXXFLAGS
+     - Extra cxx flags
+   * - PYARROW_WITH_STATIC_PARQUET
+     - Rely on parquet shared libraries where relevant, default OFF
+   * - PYARROW_WITH_STATIC_BOOST
+     - Rely on Boost shared libraries on linking static parquet, default OFF
```
AlenkaF, line 649: There is some inconsistency if I am reading the code correctly. 

`PYARROW_WITH_STATIC_PARQUET` and `PYARROW_WITH_STATIC_BOOST` are set to `false`/`OFF` (`0`) in setup.py:
https://github.com/apache/arrow/blob/e69a53e9ee8a1c14b9556c1009c12903d89a594b/python/setup.py#L174-L179

but the corresponding flags are set to `ON` in CMakeLists.txt:
https://github.com/apache/arrow/blob/e69a53e9ee8a1c14b9556c1009c12903d89a594b/python/setup.py#L309-L312
https://github.com/apache/arrow/blob/e69a53e9ee8a1c14b9556c1009c12903d89a594b/python/CMakeLists.txt#L119-L123

Will create an issue for it if I am not mistaken.

docs/source/developers/python.rst
```diff
@@ -586,6 +586,86 @@ Caveats
...
+   * - PYARROW_GENERATE_COVERAGE
+     - Setting ``Xlinetrace`` flag to true for the Cython compiler
+   * - PYARROW_BUNDLE_ARROW_CPP
+     - Bundle the Arrow C++ libraries, default OFF
+   * - PYARROW_BUNDLE_CYTHON_CPP
+     - Bundle the C++ files generated by Cython, default OFF
+   * - PYARROW_BUNDLE_BOOST
+     - Bundle the Boost libraries when we bundle Arrow C++, default OFF
+   * - PYARROW_BUNDLE_ARROW_CPP_HEADERS
+     - Bundle the Arrow C++ headers
```
AlenkaF, line 659: This flag doesn't seem to be used in setup.py or passed to CMake anymore:
https://github.com/apache/arrow/blob/a3cd962b8bc5c49f33824202dff382af13cfa873/python/setup.py#L196-L197

I will also create an issue, if I am not missing something.

docs/source/developers/python.rst
```diff
@@ -586,6 +586,86 @@ Caveats
...
+   * - PYARROW_WITH_STATIC_BOOST
+     - Rely on Boost shared libraries on linking static parquet, default OFF
+   * - PYARROW_GENERATE_COVERAGE
+     - Setting ``Xlinetrace`` flag to true for the Cython compiler
+   * - PYARROW_BUNDLE_ARROW_CPP
+     - Bundle the Arrow C++ libraries, default OFF
+   * - PYARROW_BUNDLE_CYTHON_CPP
+     - Bundle the C++ files generated by Cython, default OFF
+   * - PYARROW_BUNDLE_BOOST
+     - Bundle the Boost libraries when we bundle Arrow C++, default OFF
```
kou, line 657: I should have removed this in #14925.

This is not used now because Apache Arrow C++'s library part doesn't depend on Boost libraries. (Tests may depend on Boost libraries. Apache Arrow C++'s library part may depend on Boost but it just uses header-only part of Boost.)

docs/source/developers/python.rst
kou: Because `setup.py` uses `STATIC` (not `SHARED`) and `CMakeLists.txt` uses `SHARED` (not `STATIC`).
(We can change `CMakeLists.txt` to use `STATIC` to align names in `setup.py` and `CMakeLists.txt` for easy to understand.)

We can remove `PYARROW_WITH_STATIC_BOOST` because we don't use Boost now.

docs/source/developers/python.rst
```diff
@@ -586,6 +586,86 @@ Caveats
...
+     - Description
+   * - PYARROW_CMAKE_GENERATOR
+     - Example: 'Visual Studio 15 2017 Win64'
+   * - PYARROW_CMAKE_OPTIONS
+     - Extra CMake and Arrow options (ex. ``"-DARROW_SIMD_LEVEL"``,
+       ``"-DCMAKE_OSX_ARCHITECTURES"``)
+   * - PYARROW_BOOST_NAMESPACE
+     - Name of install folder for boost (if different from 'boost')
+   * - PYARROW_CXXFLAGS
+     - Extra cxx flags
```
kou, line 645:
```suggestion
     - Extra C++ compiler flags
```

docs/source/developers/python.rst
```diff
@@ -586,6 +586,86 @@ Caveats
...
+
+List of relevant environment variables that can also be used to build
+PyArrow are:
+
+.. list-table::
+   :widths: 30 30
+   :header-rows: 1
+
+   * - PyArrow environment variable
+     - Description
```
kou, line 636: How about adding one more column for the default value?
Most rows include the default value information in the "Description" column.

docs/source/developers/python.rst
```diff
@@ -586,6 +586,86 @@ Caveats
...
+   * - PYARROW_CMAKE_GENERATOR
+     - Example: 'Visual Studio 15 2017 Win64'
+   * - PYARROW_CMAKE_OPTIONS
+     - Extra CMake and Arrow options (ex. ``"-DARROW_SIMD_LEVEL"``,
+       ``"-DCMAKE_OSX_ARCHITECTURES"``)
```
kou, line 641: * How about specify option value explicitly in example because users need to specify option value?
* How about specify two options in example to show users can specify multiple options?

```suggestion
     - Extra CMake and Arrow options (ex. ``"-DARROW_SIMD_LEVEL=NONE -DCMAKE_OSX_ARCHITECTURES=x86_64;arm64"``)
```

docs/source/developers/python.rst
AlenkaF: Thanks for the explanation!
I can remove the variable from setup.py in this PR if you agree it fits.

docs/source/developers/python.rst
```diff
@@ -586,6 +586,86 @@ Caveats
...
 
+Relevant components and environment variables
+=============================================
+
+List of relevant Arrow CMake flags and corresponding environment variables
+to be used when building PyArrow are:
+
+.. list-table::
+   :widths: 30 30
+   :header-rows: 1
```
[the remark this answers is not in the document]
AlenkaF: Oh yes! Removed it to test and forgot to put it back. Thanks!

docs/source/developers/python.rst
AlenkaF: Sorry if the questions are silly:
But we can link dynamically to Arrow C++ libs in python I think? And it is also a bit confusing for me that setup.py and CMake that is run from setup.py define different ways to link to the libraries.

Yes, it would be much more clear if this could be aligned or at least I can try to understand and add a sentence or two about this to the docs.

kou: > But we can link dynamically to Arrow C++ libs in python I think?

Correct.

> setup.py and CMake that is run from setup.py define different ways to link to the libraries.

Does it mean that `setup.py` says "static or not" and `CMake` says "shared or not"?


Ah, wait. We may be able to remove `PYARROW_WITH_STATIC_PARQUET` too like `PYARROW_WITH_STATIC_BOOST` with the following change:

```diff
diff --git a/python/CMakeLists.txt b/python/CMakeLists.txt
index 8f846348f3..2e62dfb0e5 100644
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -586,7 +586,7 @@ if(PYARROW_BUILD_PARQUET)
             DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
   endif()
 
-  if(PYARROW_PARQUET_USE_SHARED)
+  if(ARROW_BUILD_SHARED)
     if(PYARROW_BUNDLE_ARROW_CPP)
       bundle_arrow_lib(${PARQUET_SHARED_LIB} SO_VERSION ${PARQUET_SO_VERSION})
       if(MSVC)

```

AlenkaF: Great! Will include the change and test it locally.

AlenkaF:
@github-actions crossbow submit wheel-windows*

AlenkaF:

kou (approved):
+1

docs/source/developers/python.rst
```diff
@@ -586,6 +586,89 @@ Caveats
...
+     - Default value
+   * - PYARROW_CMAKE_GENERATOR
+     - Example: 'Visual Studio 15 2017 Win64'
+     - ``''``
+   * - PYARROW_CMAKE_OPTIONS
+     - Extra CMake and Arrow options (ex. ``"-DARROW_SIMD_LEVEL=NONE -DCMAKE_OSX_ARCHITECTURES=x86_64;arm64"``)
+     - ``''``
+   * - PYARROW_BOOST_NAMESPACE
+     - Name of install folder for Boost (if different from 'boost')
+     - ``''``
```
jorisvandenbossche, line 646: @kou can this also be removed then?

docs/source/developers/python.rst
```diff
@@ -586,6 +586,89 @@ Caveats
...
+   * - PYARROW_CXXFLAGS
+     - Extra C++ compiler flags
+     - ``''``
+   * - PYARROW_WITH_STATIC_BOOST
+     - Rely on Boost shared libraries on linking static parquet
+     - ``0`` (``OFF``)
```
jorisvandenbossche, line 652: Here is another Boost reference

docs/source/developers/python.rst
kou: Ah, yes. Sorry, I missed it.

AlenkaF: Missed it too! Thanks, will remove 👍

docs/source/developers/python.rst
```diff
@@ -586,6 +586,83 @@ Caveats
...
+to be used when building PyArrow are:
+
+.. list-table::
+   :widths: 30 30
+   :header-rows: 1
+
+   * - Arrow flags/options
+     - Corresponding environment variables for PyArrow
+   * - ``CMAKE_BUILD_TYPE``
+     - ``PYARROW_BUILD_TYPE`` (release, debug or relwithdebinfo)
```
assignUser, line 602: Usually the build types are capitalized, is this not the case in pyarrow cmake?

assignUser: nvm found my answer in setup.py :D

assignUser (approved):
+1

jorisvandenbossche (approved):

PR #34466, opened 2023-03-06
WIP: Test ARM workflow

raulcd:
PR for testing ARM runners


raulcd:
@kou @assignUser this is working now: https://github.com/apache/arrow/actions/runs/4344862403/jobs/7588770241
We have added self-hosted ARM runners as organization level runners to a newly created `Arrow` runner-group. The group currently contains two repositories, `arrow` and `arrow-adbc` but we can ask infra to add more.

kou:
Wow! Thanks!!!

PR #34468, opened 2023-03-06
GH-34467: [R] Disable DuckDB tests on R versions < 4.0.0  

### LINKED ISSUE #34467: [R] Disable DuckDB tests on R versions < 4.0.0

Tests in the Arrow R package involving DuckDB are failing for R < 4.0.0 due to a newly released version of DuckDB which uses a function from the `tools` package which was only introduced in `tools` 4.0.0; see https://github.com/duckdb/duckdb/issues/6589.  

We should skip DuckDB tests on these builds for now; we can re-enable them at a later point if supporting Arrow with DuckDB on old R versions is thought to be a priority and we implement a workaround, or DuckDB release their own fix.

### Component(s)

R

thisisnic:
This PR skips DuckDB tests on older versions of R.
* Closes: #34467

### DIFF
--- a/r/R/duckdb.R
+++ b/r/R/duckdb.R
@@ -94,7 +94,8 @@ run_duckdb_examples <- function() {
     requireNamespace("duckdb", quietly = TRUE) &&
     packageVersion("duckdb") > "0.2.7" &&
     requireNamespace("dplyr", quietly = TRUE) &&
-    requireNamespace("dbplyr", quietly = TRUE)
+    requireNamespace("dbplyr", quietly = TRUE) &&
+    R.Version()$major >= 4
 }
 
 # Adapted from dbplyr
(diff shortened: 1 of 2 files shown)


eitsupi:
Maybe the following example test should also not be run < R 4.0?
https://github.com/apache/arrow/blob/3ca55cb258af6676f64d4dd11a84cf01bc9f7289/r/R/duckdb.R#L127-L138

thisisnic:
@github-actions crossbow submit test-r-versions

r/R/duckdb.R
```diff
@@ -94,7 +94,8 @@ run_duckdb_examples <- function() {
     requireNamespace("duckdb", quietly = TRUE) &&
     packageVersion("duckdb") > "0.2.7" &&
     requireNamespace("dplyr", quietly = TRUE) &&
-    requireNamespace("dbplyr", quietly = TRUE)
+    requireNamespace("dbplyr", quietly = TRUE) &&
+    R.Version()$major >= 4
```
paleolimbot, line 98: In `skip_r_older_than()` we use `getRversion()`, so maybe `getRversion() > "4.0.0"` for consistency?

paleolimbot (approved):
A suggestion about consistency in checking the R version, but other than that looks great! Thanks!

r/R/duckdb.R
thisisnic: Good catch!

PR #34469, opened 2023-03-06
GH-33801: [Python] Further expose C++ Extension Types in Python

### LINKED ISSUE #33997: [Python] Custom Python type/array subclasses for ExtensionTypes implemented in C++
When wrapping a type (or array) in a pyarrow object, we need to define which Python class to use. Currently, for extension types, this logic lives here in `pyarrow_wrap_data_type`:

https://github.com/apache/arrow/blob/b413ac4f2b6911af5e8241803277caccc43aa3c4/python/pyarrow/public-api.pxi#L114-L120

So there are currently two options:

- The ExtensionType is implemented in Python, by subclassing `pyarrow.(Py)ExtensionType`, and which links to the C++ `arrow::py::PyExtensionType` (a subclass of `arrow::ExtensionType`). In this case, we store the python type instance on the C++ instance, and return this as python object in `pyarrow_wrap_data_type`.
- The ExtensionType is implemented in C++, and then we currently always fall back to wrap this in the `pyarrow.BaseExtenstionType` base class (there is currently a bug in this, but that is getting fixed in [GH-33802](https://github.com/apache/arrow/pull/33802)).

However, that means that for such extension types implemented in C++, there is currently no way to have a "richer" python Type object (or Array object, since that is determined by the Type, and for a BaseExtensionType, that will always use the base ExtensionArray). While for an extension type, you might want to add type-specific attributes or methods. 

For canonical extension types that are implemented in Arrow C++ itself (for example, the currently discussed Tensor extension type in https://github.com/apache/arrow/pull/8510, or a previous effort to add complex type as extension type in https://github.com/apache/arrow/pull/10565), I think it will work today to create a custom subclass of `pyarrow.BaseExtensionType` for the specific canonical type, and then we could add a special case to `pyarrow_wrap_data_type` checking the name of the extension type, and if it is a canonical one we implement ourselves, use the python subclass we implemented ourselves. 

But for extension types that are implemented in C++ externally (or for extension types that are implemented in Arrow C++, but for which we don't provide a custom python subclass), that doesn't work. 
I am wondering to what extent we want to allow "registering" a python class that should be used when wrapping a specific C++ extension type (and to what extent this would be useful for

sjperkins:
### Rationale for this change

Explained in greater detail in https://github.com/apache/arrow/issues/33997, but my summary follows:

* C++ Extension Types currently are not well-exposed in the Python Layer
* In particular, generic `ExtensionArray` and `ExtensionScalar` types are associated with C++ Extensions exposed through `BaseExtensionType`, rather than hand-coded Python types (which would be laborious to generate).

* When `BaseExtensionType.__arrow_ext_class__()` and `BaseExtensionType.__arrow_ext_scalar_class__()` are called, associated Python classes are created on the fly and cached for future calls.
* `BaseExtensionType.__arrow_ext_serialize__()` and `BaseExtensionType.__arrow_ext_deserialize__()` defer to `CExtensionType.Serialize()` and `CExtensionType.Deserialize()` respectively.

### Are these changes tested?

* `__arrow_ext_class__()` and `__arrow_ext_scalar_class__()` are tested
* `__arrow_ext_serialize__()` and `__arrow_ext_deserialize__()` are not tested yet. I'd appreciate feedback on whether the general approach is useful going forward before adding tests.

* I don't believe so.

* Closes: #33997

### DIFF
--- a/python/pyarrow/tests/test_extension_type.py
+++ b/python/pyarrow/tests/test_extension_type.py
@@ -1115,9 +1115,16 @@ def test_cpp_extension_in_python(tmpdir):
     assert uuid_type.extension_name == "uuid"
     assert uuid_type.storage_type == pa.binary(16)
 
+    array_cls = uuid_type.__arrow_ext_class__()
+    scalar_cls = uuid_type.__arrow_ext_scalar_class__()
+    assert array_cls.__name__ == "ExtensionArray(uuid)"
+    assert scalar_cls.__name__ == "ExtensionScalar(uuid)"
+
     array = mod._make_uuid_array()
     assert array.type == uuid_type
+    assert isinstance(array, array_cls)
     assert array.to_pylist() == [b'abcdefghijklmno0', b'0onmlkjihgfedcba']
+    assert isinstance(array[0], scalar_cls)
     assert array[0].as_py() == b'abcdefghijklmno0'
     assert array[1].as_py() == b'0onmlkjihgfedcba'
 
@@ -1127,3 +1134,12 @@ def test_cpp_extension_in_python(tmpdir):
     reconstructed_array = batch.column(0)
     assert reconstructed_array.type == uuid_type
     assert reconstructed_array == array
+
+    storage = pa.array([b'0onmlkjihgfedcba']*4, pa.binary(16))
+    ext_array = pa.ExtensionArray.from_storage(uuid_type, storage)
+    assert isinstance(ext_array, array_cls)
+    assert len(ext_array) == 4
+
+    for i in range(4):
+        assert isinstance(ext_array[i], scalar_cls)
+        assert ext_array[i].as_py() == b'0onmlkjihgfedcba'
--- a/python/pyarrow/includes/libarrow.pxd
+++ b/python/pyarrow/includes/libarrow.pxd
@@ -2693,6 +2693,12 @@ cdef extern from "arrow/extension_type.h" namespace "arrow":
         c_string extension_name()
         shared_ptr[CDataType] storage_type()
 
+        c_string Serialize()
+        CResult[shared_ptr[CDataType]] Deserialize(shared_ptr[CDataType] storage_type,
+                                                   const c_string & serialized_data)
+
+        bint ExtensionEquals(CExtensionType other)
+
         @staticmethod
         shared_ptr[CArray] WrapArray(shared_ptr[CDataType] ext_type,
                                      shared_ptr[CArray] storage)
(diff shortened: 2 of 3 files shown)


python/pyarrow/tests/test_extension_type.py
```diff
@@ -1115,9 +1115,16 @@ def test_cpp_extension_in_python(tmpdir):
...
     assert uuid_type.storage_type == pa.binary(16)
 
+    array_cls = uuid_type.__arrow_ext_class__()
+    scalar_cls = uuid_type.__arrow_ext_scalar_class__()
+    assert array_cls.__name__ == "ExtensionArray(uuid)"
+    assert scalar_cls.__name__ == "ExtensionScalar(uuid)"
+
     array = mod._make_uuid_array()
     assert array.type == uuid_type
+    assert isinstance(array, array_cls)
```
sjperkins, line 1125: prior to this change, only `isinstance(array, pyarrow.ExtensionArray)` would have held.

python/pyarrow/tests/test_extension_type.py
```diff
@@ -1115,9 +1115,16 @@ def test_cpp_extension_in_python(tmpdir):
     assert uuid_type.extension_name == "uuid"
     assert uuid_type.storage_type == pa.binary(16)
 
+    array_cls = uuid_type.__arrow_ext_class__()
+    scalar_cls = uuid_type.__arrow_ext_scalar_class__()
+    assert array_cls.__name__ == "ExtensionArray(uuid)"
+    assert scalar_cls.__name__ == "ExtensionScalar(uuid)"
```
sjperkins, line 1121: The auto-generated class naming schema is similar to that of BaseExtensionType. i.e. `BaseExtensionType(extension<uuid>)`.

python/pyarrow/tests/test_extension_type.py
```diff
@@ -1127,3 +1134,12 @@ def test_cpp_extension_in_python(tmpdir):
     reconstructed_array = batch.column(0)
     assert reconstructed_array.type == uuid_type
     assert reconstructed_array == array
+
+    storage = pa.array([b'0onmlkjihgfedcba']*4, pa.binary(16))
+    ext_array = pa.ExtensionArray.from_storage(uuid_type, storage)
+    assert isinstance(ext_array, array_cls)
```
sjperkins, line 1140: From memory, before to this change, `pa.ExtensionArray.from_storage` would have returned a `pyarrow.ExtensionArray`.

sjperkins:
Would it be possible to re-open this issue? I'm awaiting review from Arrow maintainers here.

amol-:
Yes of course, the cleaning was done exactly to surface the prs that still needed attention

sjperkins:
Closing for reasons discussed in https://github.com/apache/arrow/pull/34483#issuecomment-1491702391

PR #34473, opened 2023-03-06
GH-34472: [Go][FlightRPC] Drain result of DoAction in Flight SQL client

### LINKED ISSUE #34472: [Go][FlightRPC] Flight SQL client does not drain DoAction streams

This is similar to a bug that was in the C++ Flight SQL client. This means that certain operations won't actually complete when the function returns, causing nondeterministic behavior in callers.

### Component(s)

Go

lidavidm:
### Rationale for this change

Unless we drain the result of DoAction, we can't be sure that the RPC actually completed.

Always drain the result of DoAction.

### Are these changes tested?

A new test was added and existing tests were modified.

This fixes potential nondeterministic behavior that users may have previously observed.
* Closes: #34472

### DIFF
--- a/go/arrow/flight/flightsql/client.go
+++ b/go/arrow/flight/flightsql/client.go
@@ -54,6 +54,18 @@ type Client struct {
 	Alloc memory.Allocator
 }
 
+// Ensure the result of a DoAction is fully consumed
+func readUntilEOF(stream pb.FlightService_DoActionClient) error {
+	for {
+		_, err := stream.Recv()
+		if err == io.EOF {
+			return nil
+		} else if err != nil {
+			return err
+		}
+	}
+}
+
 func descForCommand(cmd proto.Message) (*flight.FlightDescriptor, error) {
 	var any anypb.Any
 	if err := any.MarshalFrom(cmd); err != nil {
@@ -395,8 +407,6 @@ func (c *Client) Prepare(ctx context.Context, query string, opts ...grpc.CallOpt
 	if stream, err = c.Client.DoAction(ctx, &action, opts...); err != nil {
 		return
 	}
-	defer stream.CloseSend()
-
 	return parsePreparedStatementResponse(c, c.Alloc, stream)
 }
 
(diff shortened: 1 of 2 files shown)


go/arrow/flight/flightsql/client.go
```diff
@@ -396,6 +408,11 @@ func (c *Client) Prepare(ctx context.Context, query string, opts ...grpc.CallOpt
 		return
 	}
 	defer stream.CloseSend()
+	defer func() {
+		// XXX: assuming server will not return a result and then an error
+		// (or else we need to also try to clean up stmt)
+		err = readUntilEOF(stream)
+	}()
```
zeroshade, line 415: would it make more sense to put this into `parsePreparedStatementResponse` and get it for free in all the spots that just delegate to it?

go/arrow/flight/flightsql/client.go
```diff
@@ -500,6 +522,11 @@ func (c *Client) CancelQuery(ctx context.Context, info *flight.FlightInfo, opts
...
```
zeroshade, line 529: rather than defer this, why not just do it manually after the next if statement?

go/arrow/flight/flightsql/client.go
```diff
@@ -775,9 +816,7 @@ func (tx *Txn) Commit(ctx context.Context, opts ...grpc.CallOption) error {
 	}
 
 	tx.txn = nil
-	if _, err = stream.Recv(); err == io.EOF {
-		err = nil
-	}
+	err = readUntilEOF(stream)
 	return err
```
zeroshade, line 820: `return readUntilEOF(stream)` ?

go/arrow/flight/flightsql/client.go
```diff
@@ -806,10 +845,7 @@ func (tx *Txn) Rollback(ctx context.Context, opts ...grpc.CallOption) error {
 	}
 
 	tx.txn = nil
-	if _, err = stream.Recv(); err == io.EOF {
-		err = nil
-	}
-
+	err = readUntilEOF(stream)
 	return err
```
zeroshade, line 849: same as above

zeroshade (commented):
Looks good, just a few nitpicks

lidavidm:
Cleaned things up:
- CloseSend() needs to be called before Recv() always
- Mocks were made more reliable via Once()
- Addressed comments

zeroshade (approved):
LGTM though i think you missed a spot which could just `return readUntilEOF`

lidavidm:
Whoops, thanks.

lidavidm:
I'll merge this now, CI failures are from #34488

PR #34477, opened 2023-03-07
GH-34422: [R] Expose GcsFileSystem$options

### LINKED ISSUE #34422: [R] Expose GcsFileSystem options in R

Related to https://github.com/apache/arrow/issues/34421 and commented as a TODO:

https://github.com/apache/arrow/blob/50fb5b0df84a3ccaa665965e36a6e757d1af98f1/r/tests/testthat/test-gcs.R#L26-L27

It's not currently possible to test the behavior of GcsFileSystem$create's handling of options because they aren't exposed. 

### Component(s)

R

amoeba:
Closes https://github.com/apache/arrow/issues/34422

### Rationale for this change

Exposing options on GcsFilesystem objects is useful for testing but may also be useful to R arrow users who are using GCS (gs_bucket). The implementation here is one-way which might be a good point for in discussion in code review.

### Are these changes tested?

I've added two tests which pass.

Yes, though I wasn't sure if it made sense to document it. With this change, a user can run the following to get the options:

```r
> bucket <- gs_bucket("voltrondata-labs-datasets")
> bucket$base_fs$options
$anonymous
[1] FALSE

$scheme
[1] "https"

$retry_limit_seconds
[1] 15
```
* Closes: #34422

### DIFF
--- a/r/R/filesystem.R
+++ b/r/R/filesystem.R
@@ -503,7 +503,10 @@ gs_bucket <- function(bucket, ...) {
 #' @rdname FileSystem
 #' @export
 GcsFileSystem <- R6Class("GcsFileSystem",
-  inherit = FileSystem
+  inherit = FileSystem,
+  active = list(
+    options = function() fs___GcsFileSystem__options(self)
+  )
 )
 GcsFileSystem$create <- function(anonymous = FALSE, retry_limit_seconds = 15, ...) {
   # The default retry limit in C++ is 15 minutes, but that is experienced as
--- a/r/R/filesystem.R
+++ b/r/R/filesystem.R
@@ -165,8 +165,8 @@ FileSelector$create <- function(base_dir, allow_not_found = FALSE, recursive = F
 #'    credentials using standard GCS configuration methods.
 #' - `access_token`: optional string for authentication. Should be provided along
 #'   with `expiration`
-#' - `expiration`: optional date representing point at which `access_token` will
-#'   expire.
+#' - `expiration`: `POSIXct`. optional datetime representing point at which
+#'   `access_token` will expire.
 #' - `json_credentials`: optional string for authentication. Point to a JSON
 #'   credentials file downloaded from GCS.
 #' - `endpoint_override`: if non-empty, will connect to provided host name / port,
@@ -505,7 +505,18 @@ gs_bucket <- function(bucket, ...) {
 GcsFileSystem <- R6Class("GcsFileSystem",
   inherit = FileSystem,
   active = list(
-    options = function() fs___GcsFileSystem__options(self)
+    options = function() {
+      out <- fs___GcsFileSystem__options(self)
+
+      # Convert from nanoseconds to POSIXct w/ UTC tz
+      if ("expiration" %in% names(out)) {
+        out$expiration <- as.POSIXct(
+          out$expiration / 1000000000, origin = "1970-01-01", tz = "UTC"
+        )
+      }
+
+      out
+    }
   )
 )
 GcsFileSystem$create <- function(anonymous = FALSE, retry_limit_seconds = 15, ...) {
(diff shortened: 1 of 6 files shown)


r/src/filesystem.cpp
```diff
@@ -422,4 +424,63 @@ std::shared_ptr<fs::GcsFileSystem> fs___GcsFileSystem__Make(bool anonymous,
...
+  fs::GcsOptions opts = fs->options();
+
+  // GcsCredentials
+  out.push_back({"anonymous"_nm = opts.credentials.anonymous() });
+
+  if (opts.credentials.access_token() != "") {
+    out.push_back({"access_token"_nm = opts.credentials.access_token() });
+  }
+
+  // TODO: Cast to posixct?
```
amoeba, line 442: This needs more work. What I'd like is to get this back to a POSIXct but I'm not sure how to do that.

amoeba:
Hey @thisisnic @paleolimbot: I'd love a review on this. This is my first time working with cpp11 so extra scrutiny would be good :). You'll see my implementation is very verbose -- did I miss some automatic conversion helpers that would make this shorter?

r/src/filesystem.cpp
paleolimbot: I would do that back in R land...you could do something like `sexp expiration = ...; expiration.attr("class") = "POSIXct"; expiration.attr("tzone") = "UTC";` but maybe safer to just do `options$expiration <- as.POSIXct(options$expiration, origin = "1970-01-01", tz = "UTC")`.

paleolimbot (commented):
I think what you have here is as non-verbose as you're going to get! For some other options objects we've added an R6 class with accessors (e.g., I think CsvReadOptions are done this way) but that tends to be verbose in its own way.

I addded a note about how to go about POSIXct and I see that CI is failing. Once CI is green I think this is good to go unless Nic says otherwise!

amoeba:
Thanks @paleolimbot. I went with your suggestion on converting to POSIXct on the R side in https://github.com/apache/arrow/pull/34477/commits/b9153a949766f4d3532c3117e44d08e965e318df and also included  a test to cover it. I pushed up https://github.com/apache/arrow/pull/34477/commits/122b219018fad07d6d5f394d95696dc226e1e75e to fix the CI failure but I'm not totally sure about it. CI seems green but I want to look at it tomorrow when I get back to my laptop.

r/R/filesystem.R
```diff
@@ -547,6 +561,11 @@ GcsFileSystem$create <- function(anonymous = FALSE, retry_limit_seconds = 15, ..
     )
   }
 
+  # Stop if expiration isn't a POSIXct
+  if ("expiration" %in% names(options) && !inherits(options$expiration, "POSIXct")) {
+    stop("Option 'expiration' must be of class POSIXct", call. = FALSE)
```
thisisnic, line 566: You could extend this to make it even more helpful by printing out the class of the object, e.g. something like

`paste("Option 'expiration' must be of class POSIXct, not", class(options$expiration)[[1]])`

Only a minor point - feel free to leave it as-is.

r/src/filesystem.cpp
```diff
@@ -422,4 +423,65 @@ std::shared_ptr<fs::GcsFileSystem> fs___GcsFileSystem__Make(bool anonymous,
...
+  fs::GcsOptions opts = fs->options();
+
+  // GcsCredentials
+  out.push_back({"anonymous"_nm = opts.credentials.anonymous()});
+
+  if (opts.credentials.access_token() != "") {
+    out.push_back({"access_token"_nm = opts.credentials.access_token()});
+  }
+
+  // TODO: Cast to posixct?
```
thisisnic, line 441:
```suggestion

```

thisisnic (approved):
A couple of minor suggestions, but otherwise looks good!

r/R/filesystem.R
amoeba: Thanks, this is an improvement.

amoeba:
Thanks @thisisnic for the 👀 .

You'll see I pushed up a couple of commits after your review related to type casting when iterating over GcsOptions' default_metadata. The issue had to with Arrow's use of signed ints for container sizes where cpp11 (and the STL) uses signed ints. I ended up with a static_cast which isn't totally safe w/o a check but I think this is safe enough in practice.

Aside from any comments on the above, this is ready for a merge.

PR #33805, opened 2023-01-20
GH-33804: [Python] Add support for manylinux_2_28 wheel

sjperkins:
Closes #33804 

### Rationale for this change

At some point, it would be useful to support the new C++ ABI `_GLIBCXX_USE_CXX11_ABI=1` in pyarrow wheels, especially when moving to C++17:

- https://github.com/apache/arrow/issues/32415

I wanted to create a pyarrow wheel that supported the above ABI and adapted the existing CENTOS 7 manylinux2014 Dockerfile/wheel to produce a AlmaLinux 8 manylinux_2_28 Dockerfile/wheel.

Publishing wheels with a new ABI needs [careful consideration](https://pypackaging-native.github.io/key-issues/native-dependencies/cpp_deps/) so I think this is low priority, but I thought I'd provide this manylinux_2_28 implementation in case it was useful for current/future adoption. 

### What changes are included in this PR?

A manylinux_2_28 Dockerfile, adopted from the existing manylinux2014 Dockerfile

### Are these changes tested?

Manually tested at present

### Are there any user-facing changes?

Yes, there's a major ABI change, as pyarrow will be compiled with `_GLIBCXX_USE_CXX11_ABI=1`
* Closes: #33804

PR #34483, opened 2023-03-07
GH-33801: [Python] Generate on the fly Python Extension Types wrapping C++ Extension Types

### LINKED ISSUE #33997: [Python] Custom Python type/array subclasses for ExtensionTypes implemented in C++

sjperkins:
Alternative implementation to https://github.com/apache/arrow/pull/34469

### Rationale for this change

Explained in greater detail in https://github.com/apache/arrow/issues/33997, but my summary follows:

* C++ Extension Types currently are not well-exposed in the Python Layer
* In particular, generic `ExtensionArray` and `ExtensionScalar` types are associated with C++ Extensions exposed through `BaseExtensionType`, rather than hand-coded Python types (which would be laborious to generate).

* When `get_cpp_extension_type` is called a Python wrapper CppExtensionType is created and cached, along with Python ExtensionArrays and ExtensionScalars
* `CppExtensionType.__arrow_ext_serialize__()` and `CppExtensionType.__arrow_ext_deserialize__()` defer to `CExtensionType.Serialize()` and `CExtensionType.Deserialize()` respectively.

### Are these changes tested?

* `__arrow_ext_class__()` and `__arrow_ext_scalar_class__()` are tested
* `__arrow_ext_serialize__()` and `__arrow_ext_deserialize__()` are not tested yet. I'd appreciate feedback on whether the general approach is useful going forward before adding tests.

* I don't believe so.

* Closes: #33997

### DIFF
--- a/python/pyarrow/includes/libarrow.pxd
+++ b/python/pyarrow/includes/libarrow.pxd
@@ -2693,6 +2693,12 @@ cdef extern from "arrow/extension_type.h" namespace "arrow":
         c_string extension_name()
         shared_ptr[CDataType] storage_type()
 
+        c_string Serialize()
+        CResult[shared_ptr[CDataType]] Deserialize(shared_ptr[CDataType] storage_type,
+                                                   const c_string & serialized_data)
+
+        bint ExtensionEquals(CExtensionType other)
+
         @staticmethod
         shared_ptr[CArray] WrapArray(shared_ptr[CDataType] ext_type,
                                      shared_ptr[CArray] storage)
--- a/python/pyarrow/public-api.pxi
+++ b/python/pyarrow/public-api.pxi
@@ -117,7 +117,8 @@ cdef api object pyarrow_wrap_data_type(
         if cpy_ext_type != nullptr:
             return cpy_ext_type.GetInstance()
         else:
-            out = BaseExtensionType.__new__(BaseExtensionType)
+            cls = get_cpp_extension_type(ext_type)
+            out = cls.__new__(cls)
     else:
         out = DataType.__new__(DataType)
 
--- a/python/pyarrow/tests/test_extension_type.py
+++ b/python/pyarrow/tests/test_extension_type.py
@@ -1115,9 +1115,16 @@ def test_cpp_extension_in_python(tmpdir):
     assert uuid_type.extension_name == "uuid"
     assert uuid_type.storage_type == pa.binary(16)
 
+    array_cls = uuid_type.__arrow_ext_class__()
+    scalar_cls = uuid_type.__arrow_ext_scalar_class__()
+    assert array_cls.__name__ == "ExtensionArray(uuid)"
+    assert scalar_cls.__name__ == "ExtensionScalar(uuid)"
+
     array = mod._make_uuid_array()
     assert array.type == uuid_type
+    assert isinstance(array, array_cls)
     assert array.to_pylist() == [b'abcdefghijklmno0', b'0onmlkjihgfedcba']
+    assert isinstance(array[0], scalar_cls)
     assert array[0].as_py() == b'abcdefghijklmno0'
     assert array[1].as_py() == b'0onmlkjihgfedcba'
 
@@ -1127,3 +1134,12 @@ def test_cpp_extension_in_python(tmpdir):
     reconstructed_array = batch.column(0)
     assert reconstructed_array.type == uuid_type
     assert reconstructed_array == array
+
+    storage = pa.array([b'0onmlkjihgfedcba']*4, pa.binary(16))
+    ext_array = pa.ExtensionArray.from_storage(uuid_type, storage)
+    assert isinstance(ext_array, array_cls)
+    assert len(ext_array) == 4
+
+    for i in range(4):
+        assert isinstance(ext_array[i], scalar_cls)
+        assert ext_array[i].as_py() == b'0onmlkjihgfedcba'
--- a/python/pyarrow/types.pxi
+++ b/python/pyarrow/types.pxi
@@ -20,6 +20,7 @@ from cpython.pycapsule cimport PyCapsule_CheckExact, PyCapsule_GetPointer
 import atexit
 from collections.abc import Mapping
 import re
+from threading import Lock
 import sys
 import warnings
 
@@ -831,7 +832,6 @@ cdef class BaseExtensionType(DataType):
     """
     Concrete base class for extension types.
     """
-
     cdef void init(self, const shared_ptr[CDataType]& type) except *:
         DataType.init(self, type)
         self.ext_type = <const CExtensionType*> type.get()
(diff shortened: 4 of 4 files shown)


amol-:

amol-:
Yes of course, the cleaning was done exactly to surface the prs that still needed attention

jorisvandenbossche:
@sjperkins sorry about closing all your PRs, and about the slow review here. I will try to get back to this one and https://github.com/apache/arrow/pull/34469 shortly.

Other question: you couldn't reopen this PR yourself? I mean, did the GitHub UI allow that? Because if only committers can reopen PRs, that would be useful to know because the messaging should be different then if we would do some auto-closing in the future.

sjperkins:
> @sjperkins sorry about closing all your PRs, and about the slow review here. I will try to get back to this one and #34469 shortly.

No problem! I assumed that the upcoming Arrow 12.0.0 release was important.

> Other question: you couldn't reopen this PR yourself? I mean, did the GitHub UI allow that? Because if only committers can reopen PRs, that would be useful to know because the messaging should be different then if we would do some auto-closing in the future.

Yes, the UI didn't seem to give me the option to reopen the PR.

sjperkins:
> Yes, the UI didn't seem to give me the option to reopen the PR.

But I do have the ability to close the PR..

sjperkins:
> Yes, the UI didn't seem to give me the option to reopen the PR.

OK, this does seem possible if I close the PR. Apologies I responded late yesterday evening, so I may have conflated the force push in #33805 or missed the button in this issue.

sjperkins:
@jorisvandenbossche Regarding this issue and #33997, they're somewhat exploratory because they dynamically generate Python types.

I'm interested in whether these strategies would be a useful way of exposing pure C++ Extension Types in Python.

For instance, `{from,to}_numpy()` and `{from,to}_pandas` could then be added to these types.

However, perhaps this should be done through the PyExtensionType mechanisms (both Python and C++)?

sjperkins:
> For instance, {from,to}_numpy() and {from,to}_pandas could then be added to these types.

Some further context to motivate for this: It would be useful to efficiently convert nested FixedSizeListArray's into numpy arrays. The default produces deeply nested numpy arrays with `object` dtype. See https://github.com/ratt-ru/casa-arrow/issues/2 for example.

This may also be relevant to #8510 which, from imperfect memory, uses FixedSizeListArray's to represent Tensors.

jorisvandenbossche:
Can you reopen it now? (it might depend on whether you closed it yourself)

sjperkins:
> Can you reopen it now? (it might depend on whether you closed it yourself)

Ah I cannot. I think it does depend on the closer.

jorisvandenbossche:
OK, that's good to know, thanks for checking!

jorisvandenbossche:
> I'm interested in whether these strategies would be a useful way of exposing pure C++ Extension Types in Python.

I am not a huge fan of the idea of creating classes on the fly .. Also, does this give something more useful than wrapping it just in a base class as is done now? (because right now this generated class also doesn't have any extension-type specific logic?)

> Some further context to motivate for this: It would be useful to efficiently convert nested FixedSizeListArray's into numpy arrays. 

Yeah, so if we have a way to register a python class to use as the type class for an extension type implemented in C++ (https://github.com/apache/arrow/issues/33997), you can override the `to_numpy()` method. However, than only works if you call this method directly on the array object. But if you convert a table, or a chunked array, it will still go through the C++ layer which currently falls back on converting the storage array. So we might need to think about a more general mechanism here to tap into this conversion logic.


> This may also be relevant to #8510 which, from imperfect memory, uses FixedSizeListArray's to represent Tensors.

Yes, that one is close to being merged, and then we can expose this in python, which might be a useful exercise to see how this goes / what we can learn from this (but of course it's an internal one, so we can hard code support for it)

sjperkins:
> I am not a huge fan of the idea of creating classes on the fly .. Also, does this give something more useful than wrapping it just in a base class as is done now? (because right now this generated class also doesn't have any extension-type specific logic?)

Yes, I understand the suggestion is a bit exotic. I think what it gives is the ability to associate specific Python classes with C++ Extension Types/Arrays  and attach (mainly) `{from,to}_numpy` functionality to the Python class. There probably are nicer ways to do it, but nothing comes immediately to mind. Perhaps there's a way to do this with the existing C++ PyExtensionType machinery.

> Yeah, so if we have a way to register a python class to use as the type class for an extension type implemented in C++ (#33997), you can override the `to_numpy()` method. However, than only works if you call this method directly on the array object. But if you convert a table, or a chunked array, it will still go through the C++ layer which currently falls back on converting the storage array. So we might need to think about a more general mechanism here to tap into this conversion logic.

Thanks for pointing this out, this gives me a broader perspective on what would be useful.

> 
> > This may also be relevant to #8510 which, from imperfect memory, uses FixedSizeListArray's to represent Tensors.
> 
> Yes, that one is close to being merged, and then we can expose this in python, which might be a useful exercise to see how this goes / what we can learn from this (but of course it's an internal one, so we can hard code support for it)

I'll think about this a bit more. let me know if there're approaches you think would be useful to experiment with.

sjperkins:
Closing because the use of dynamic types is a bit too exotic.

PR #34488, opened 2023-03-07
MINOR: [Go][CI] Add openssl to macos CGO runs

zeroshade:
### Rationale for this change
Fix the failing CI runs

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -313,9 +313,7 @@ jobs:
           ci/scripts/go_build.sh $(pwd)
       - name: Test
         shell: bash
-        run: |
-          export PKG_CONFIG_PATH=/usr/local/Cellar/openssl@3/3.0.8/lib/pkgconfig:$PKG_CONFIG_PATH
-          ci/scripts/go_test.sh $(pwd)
+        run: ci/scripts/go_test.sh $(pwd)
 
   windows-mingw:
     name: AMD64 Windows MinGW ${{ matrix.mingw-n-bits }} CGO
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -309,7 +309,7 @@ jobs:
       - name: Build
         shell: bash
         run: |
-          export PKG_CONFIG_PATH=/usr/local/Cellar/openssl@3/3.0.8/lib/pkgconfig:$PKG_CONFIG_PATH
+           echo "PKG_CONFIG_PATH=/usr/local/Cellar/openssl@3/3.0.8/lib/pkgconfig:$PKG_CONFIG_PATH" >> $GITHUB_ENV
           ci/scripts/go_build.sh $(pwd)
       - name: Test
         shell: bash
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -303,7 +303,7 @@ jobs:
           cache-dependency-path: go/go.sum
       - name: Brew Install Arrow and pkg-config
         shell: bash
-        run: brew install apache-arrow pkg-config
+        run: brew install apache-arrow pkg-config openssl
       - name: Install staticcheck
         run: go install honnef.co/go/tools/cmd/staticcheck@${{ matrix.staticcheck }}
       - name: Build
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -300,7 +300,7 @@ jobs:
         with:
           go-version: ${{ matrix.go }}
           cache: true
-          cache-dependency-path: go/go.sum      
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -300,18 +300,22 @@ jobs:
         with:
           go-version: ${{ matrix.go }}
           cache: true
-          cache-dependency-path: go/go.sum
+          cache-dependency-path: go/go.sum
```
assignUser, line 303: nit:
```suggestion
          cache-dependency-path: go/go.sum
```

.github/workflows/go.yml
```diff
@@ -300,18 +300,22 @@ jobs:
...
       - name: Brew Install Arrow and pkg-config
         shell: bash
         run: brew install apache-arrow pkg-config
       - name: Install staticcheck
         run: go install honnef.co/go/tools/cmd/staticcheck@${{ matrix.staticcheck }}
       - name: Build
         shell: bash
-        run: ci/scripts/go_build.sh $(pwd)
+        run: |
+          export PKG_CONFIG_PATH=/usr/local/Cellar/openssl@3/3.0.8/lib/pkgconfig:$PKG_CONFIG_PATH
```
assignUser, line 312: This should update it for all following steps by setting it once :)
```suggestion
           echo "PKG_CONFIG_PATH=/usr/local/Cellar/openssl@3/3.0.8/lib/pkgconfig:$PKG_CONFIG_PATH" >> $GITHUB_ENV
```

lidavidm, line 312: Hmm, hardcoding the path like this seems brittle. Are we missing a `brew link` or something similar?

.github/workflows/go.yml
lidavidm, line 303: nit: trailing whitespace

.github/workflows/go.yml
```diff
@@ -300,18 +300,22 @@ jobs:
...
+          ci/scripts/go_build.sh $(pwd)
       - name: Test
         shell: bash
-        run: ci/scripts/go_test.sh $(pwd)
+        run: |
+          export PKG_CONFIG_PATH=/usr/local/Cellar/openssl@3/3.0.8/lib/pkgconfig:$PKG_CONFIG_PATH
+          ci/scripts/go_test.sh $(pwd)
```
assignUser, line 318:
```suggestion
        run: ci/scripts/go_test.sh $(pwd)
```

.github/workflows/go.yml
lidavidm: Ah, OpenSSL is special in that regard for Homebrew...

assignUser (commented):
A minor change allows us to not have to export PKG_CONFIG_PATH in each step.

.github/workflows/go.yml
zeroshade: From what I found, it seems like it's explicitly an issue with the openssl brew package not placing `openssl.pc` in the same spot as other pkg-config paths. :( If there's a `brew link` or something that we can do, I haven't been able to find it.

assignUser: IIRC we have something similar hardcoded in a bunch of places because there is no better solution?

lidavidm: Yeah, that makes sense. Oh well...

lidavidm (approved):

.github/workflows/go.yml
```diff
@@ -308,7 +308,9 @@ jobs:
         run: go install honnef.co/go/tools/cmd/staticcheck@${{ matrix.staticcheck }}
       - name: Build
         shell: bash
-        run: ci/scripts/go_build.sh $(pwd)
+        run: |
+           echo "PKG_CONFIG_PATH=/usr/local/Cellar/openssl@3/3.0.8/lib/pkgconfig:$PKG_CONFIG_PATH" >> $GITHUB_ENV
```
kou, line 312: This doesn't affect in this step.

We need a separated step:

```yaml
- name: Prepare PKG_CONFIG_PATH
  shell: bash
  run: |
    echo "PKG_CONFIG_PATH=$(brew --prefix openssl)/lib/pkgconfig:$PKG_CONFIG_PATH" >> $GITHUB_ENV
- name: Build
  shell: bash
  run: ci/scripts/go_build.sh $(pwd)
...
```

And this change causes YAML syntax error: https://github.com/apache/arrow/actions/runs/4358242702/workflow#L308

I think that the indentation of this line is a problem.

assignUser: Also weird how the workflow didn't run on the last commit. Sorry for the indentation error that happens due to my comment suggestion 🙈

PR #34490, opened 2023-03-07
MINOR: [Go][CI] Fix indentation and syntax for yml file

zeroshade:
Fixing issue caused by #34488

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -306,11 +306,13 @@ jobs:
         run: brew install apache-arrow pkg-config
       - name: Install staticcheck
         run: go install honnef.co/go/tools/cmd/staticcheck@${{ matrix.staticcheck }}
-      - name: Build
+      - name: Add To pkg config path
         shell: bash
         run: |
-           echo "PKG_CONFIG_PATH=/usr/local/Cellar/openssl@3/3.0.8/lib/pkgconfig:$PKG_CONFIG_PATH" >> $GITHUB_ENV
-          ci/scripts/go_build.sh $(pwd)
+          echo "PKG_CONFIG_PATH=/usr/local/Cellar/openssl@3/3.0.8/lib/pkgconfig:$PKG_CONFIG_PATH" >> $GITHUB_ENV
+      - name: Build
+        shell: bash
+        run: ci/scripts/go_build.sh $(pwd)
       - name: Test
         shell: bash
         run: ci/scripts/go_test.sh $(pwd)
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -306,11 +306,13 @@ jobs:
         run: brew install apache-arrow pkg-config
       - name: Install staticcheck
         run: go install honnef.co/go/tools/cmd/staticcheck@${{ matrix.staticcheck }}
-      - name: Build
+      - name: Add To pkg config path
         shell: bash
         run: |
-           echo "PKG_CONFIG_PATH=/usr/local/Cellar/openssl@3/3.0.8/lib/pkgconfig:$PKG_CONFIG_PATH" >> $GITHUB_ENV
-          ci/scripts/go_build.sh $(pwd)
+          echo "PKG_CONFIG_PATH=/usr/local/Cellar/openssl@3/3.0.8/lib/pkgconfig:$PKG_CONFIG_PATH" >> $GITHUB_ENV
```
kou, line 312: How about using `brew --preifx` to get `openssl@3` path?

```suggestion
          echo "PKG_CONFIG_PATH=$(brew --prefix openssl@3)/lib/pkgconfig:$PKG_CONFIG_PATH" >> $GITHUB_ENV
```

zeroshade: i didn't know that was a thing! Thanks much @kou!

.github/workflows/go.yml
```diff
@@ -374,7 +376,7 @@ jobs:
 
   linux-arm:
     name: ARM64 Debian 11 GO ${{ matrix.go }}
-    runs-on: ["arm", "linux"]
+    runs-on: ["self-hosted", "linux"]
```
kou, line 379: @raulcd Should we prefer `self-hosted` to `arm` here?

kou (approved):
+1

.github/workflows/go.yml
zeroshade: I was getting a syntax error in my IDE for having "arm" there based on GHA json schema. That's why I changed it.

kou: I see.
I'll wait for @raulcd.

raulcd: I don't mind using `self-hosted` but I think we should also use the architecture in case we want to add other self-hosted architectures in the future.

zeroshade: In the interests of getting the Go tests running again, I'm going to merge this but if @raulcd or @assignUser thinks I should change it back to "arm", just comment here and I'll do so.

assignUser: Yeah what @raulcd said as it also matches the official recommendation.

zeroshade: so, it should be `["self-hosted", "linux", "arm"]` ?

PR #34493, opened 2023-03-07
GH-34492: [Go] Fix missing boolean plain encoder state update

### LINKED ISSUE #34492: PlainBooleanEncoder occasionally emits incorrect data over page boundaries

The `PlainBooleanEncoder` is used in a way indicating that `FlushValues` "resets" its state. However, the intermediate `BitMapWriter`'s position state is not reset. In rare cases this can cause something akin to "stale" values being flushed to the column. The most immediate manifestation of this is in the Boolean statistics: e.g., on the rowgroup stats, if there is a single "true" value in the column, the way it emits maxvalue followed by minvalue reveals that the `true` maxvalue is emitted for both.

I have a fix with tests in an incoming PR.

### Component(s)

Go

aaron-mongodb:
### Rationale for this change

This is a bugfix in the go `PlainBooleanEncoder`. Statistics of boolean columns, as well as data in rare cases, can be malformed. This is a data-correctness bugfix.

`FlushValue` does not reset the intermediary BitMapWriter position. In this way previously-encoded values might "leak" into subsequent `FlushValue` calls. Page emission and statistics emission (for examples) manifest this issue.

### Are these changes tested?

Yes. The most direct test is the new `TestBooleanPlainDecoderAfterFlushing` in internal\encoding\encoding_test.go, but I also added `TestBooleanStatisticsEncoding` as a "consumer" of that encoder.

Question: should I add more testing? I experimentally harnessed a fuzz test targeted towards the Boolean encoder, and that manifested the issue, but that's a fuzz test and also targeted towards Bools alone. I also provisionally implemented a (hacky) extension to the roundtrip tests to have an intermediate flush, but that didn't seem enough to manifest the (somewhat value-count-sensitive) issue.

This is a correctness fix, so other than that, no.

* Closes: #34492

### DIFF
--- a/go/parquet/file/column_writer_test.go
+++ b/go/parquet/file/column_writer_test.go
@@ -256,10 +256,8 @@ func (p *PrimitiveWriterTestSuite) buildReader(nrows int64, compression compress
 	return file.NewColumnReader(p.descr, pagereader, mem, &p.bufferPool)
 }
 
-func (p *PrimitiveWriterTestSuite) buildWriter(_ int64, columnProps parquet.ColumnProperties, version parquet.Version) file.ColumnChunkWriter {
+func (p *PrimitiveWriterTestSuite) buildWriter(_ int64, columnProps parquet.ColumnProperties, opts ...parquet.WriterProperty) file.ColumnChunkWriter {
 	p.sink = encoding.NewBufferWriter(0, mem)
-	opts := make([]parquet.WriterProperty, 0)
-	opts = append(opts, parquet.WithVersion(version))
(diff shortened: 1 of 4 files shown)


zeroshade:
This is certainly a weird little bug, very interesting. Thanks for filing this @aaron-mongodb! Can you rebase `main` and update this? There was a typo in the go workflow yaml and I just merged a fix for that. I want to make sure the Go workflows get run with this test.

zeroshade:
@aaron-mongodb this all looks great to me! Thanks much for the fix. I'm looking forward to more contributions from you guys!

PR #34498, opened 2023-03-08
GH-34404: [Python] Failing tests because pandas.Index can now store all numeric dtypes (not only 64bit versions)

### LINKED ISSUE #34404: [Python] Failing tests because pandas.Index can now store all numeric dtypes (not only 64bit versions)
We have several failing tests in the nightly build (https://github.com/ursacomputing/crossbow/actions/runs/4277727973/jobs/7446784501) because of a change in pandas 2.0: the Index can now store all numeric dtypes, and not just int64/uint64/float64, see https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html#index-can-now-hold-numpy-numeric-dtypes.

Failing tests because of this:

```
FAILED python/pyarrow/tests/test_pandas.py::test_table_from_pandas_schema_index_columns - AssertionError: DataFrame.index are different
FAILED python/pyarrow/tests/parquet/test_dataset.py::test_read_partitioned_directory[False] - AssertionError: Attributes of DataFrame.iloc[:, 2] (column name="foo") are different
FAILED python/pyarrow/tests/parquet/test_dataset.py::test_read_partitioned_directory_s3fs[False] - AssertionError: Attributes of DataFrame.iloc[:, 2] (column name="foo") are different
```

I think all those cases are where now an int32 dtype is preserved, while before it would have been cast to int64 by pandas. But the expected result still uses int64, causing the test failures.

AlenkaF:
### Rationale for this change
Several failing tests in the nightly build (https://github.com/ursacomputing/crossbow/actions/runs/4277727973/jobs/7446784501) 

Due to change in supported dtypes for Index in pandas, the tests expecting `int64`and not `int32` are failing with dev version of pandas. The failing tests are updated to match the new pandas behaviour.
* Closes: #34404

### DIFF
--- a/python/pyarrow/tests/parquet/test_dataset.py
+++ b/python/pyarrow/tests/parquet/test_dataset.py
@@ -735,8 +735,14 @@ def _partition_test_for_filesystem(fs, base_path, use_legacy_dataset=True):
                    .reset_index(drop=True)
                    .reindex(columns=result_df.columns))
 
-    expected_df['foo'] = pd.Categorical(df['foo'], categories=foo_keys)
-    expected_df['bar'] = pd.Categorical(df['bar'], categories=bar_keys)
+    if use_legacy_dataset==True or Version(pd.__version__) < Version("2.0.0"):
+        expected_df['foo'] = pd.Categorical(df['foo'], categories=foo_keys)
+        expected_df['bar'] = pd.Categorical(df['bar'], categories=bar_keys)
+    else:
+        # With pandas 2.0.0 Index can store all numeric dtypes (not just int64/uint64/float64)
+        # Using astype() to create a categorical column preserves original dtype (int32)
+        expected_df['foo'] = expected_df['foo'].astype("category")
+        expected_df['bar'] = expected_df['bar'].astype("category")
 
     assert (result_df.columns == ['index', 'values', 'foo', 'bar']).all()
 
--- a/python/pyarrow/tests/parquet/test_dataset.py
+++ b/python/pyarrow/tests/parquet/test_dataset.py
@@ -735,12 +735,13 @@ def _partition_test_for_filesystem(fs, base_path, use_legacy_dataset=True):
                    .reset_index(drop=True)
                    .reindex(columns=result_df.columns))
 
-    if use_legacy_dataset==True or Version(pd.__version__) < Version("2.0.0"):
+    if use_legacy_dataset or Version(pd.__version__) < Version("2.0.0"):
         expected_df['foo'] = pd.Categorical(df['foo'], categories=foo_keys)
         expected_df['bar'] = pd.Categorical(df['bar'], categories=bar_keys)
     else:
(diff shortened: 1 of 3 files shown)


kou:
@github-actions crossbow submit *pandas*

python/pyarrow/tests/parquet/test_dataset.py
```diff
@@ -735,8 +735,15 @@ def _partition_test_for_filesystem(fs, base_path, use_legacy_dataset=True):
...
-    expected_df['bar'] = pd.Categorical(df['bar'], categories=bar_keys)
+    if use_legacy_dataset or Version(pd.__version__) < Version("2.0.0"):
+        expected_df['foo'] = pd.Categorical(df['foo'], categories=foo_keys)
+        expected_df['bar'] = pd.Categorical(df['bar'], categories=bar_keys)
+    else:
+        # With pandas 2.0.0 Index can store all numeric dtypes (not just
+        # int64/uint64/float64). Using astype() to create a categorical
+        # column preserves original dtype (int32)
+        expected_df['foo'] = expected_df['foo'].astype("category")
+        expected_df['bar'] = expected_df['bar'].astype("category")
```
jorisvandenbossche, line 746: This new way might work for all pandas versions? (then the if/else is not needed)

python/pyarrow/tests/test_compute.py
```diff
@@ -1934,22 +1934,48 @@ def _check_datetime_components(timestamps, timezone=None):
...
+    dayofyear = ts.dt.dayofyear
+    quarter = ts.dt.quarter
+    hour = ts.dt.hour
+    minute = ts.dt.minute
+    second = ts.dt.second.values
+    microsecond = ts.dt.microsecond
+    nanosecond = ts.dt.nanosecond
+    if Version(pd.__version__) >= Version("2.0.0"):
+        # Casting is required because pandas with 2.0.0 various numeric
+        # date/time attributes have dtype int32 (previously int64)
```
jorisvandenbossche, line 1950: Also here, it might not do any harm to just always cast to int64 above, without this extra `if: ` block (if it's already int64, that should be a no-op)

jorisvandenbossche (commented):
Generally looks good (and the CI is good, the remaining failures in the pandas nightly are unrelated), just some potential simplification comment

python/pyarrow/tests/parquet/test_dataset.py
AlenkaF: Unfortunately it doesn't: on older versions of pandas (and in the legacy dataset, donno why, didn't think it makes sense to investigate) the `foo` value type in `result_df ` is `int64` but `.astype("category")` would define the type of `foo` in `expected_df` as `int32`.

Which is just the opposite in newer version of pandas: the `foo` value type in `result_df` is `int32` but `pd.Categorical` defines the type of `foo` in `expected_df` as `int64`.

jorisvandenbossche (approved):
Thanks!

PR #34507, opened 2023-03-08
MINOR: [docs] Include `writer.close()` at end of `with ... as writer` calls.

fpoms:
### Rationale for this change

The guide currently doesn't specify that the `writer.close()` method must be called to complete a file write. This change adds that.

Adds `writer.close()` to the python IPC documentation code examples.

### Are these changes tested?
No tests because this is a change in the documentation.

No, this is a documentation change.

### DIFF
--- a/docs/source/python/ipc.rst
+++ b/docs/source/python/ipc.rst
@@ -80,6 +80,7 @@ this one can be created with :func:`~pyarrow.ipc.new_stream`:
    with pa.ipc.new_stream(sink, batch.schema) as writer:
       for i in range(5):
          writer.write_batch(batch)
+      writer.close()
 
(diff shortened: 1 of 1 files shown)


docs/source/python/ipc.rst
```diff
@@ -80,6 +80,7 @@ this one can be created with :func:`~pyarrow.ipc.new_stream`:
    with pa.ipc.new_stream(sink, batch.schema) as writer:
       for i in range(5):
          writer.write_batch(batch)
+      writer.close()
```
jorisvandenbossche, line 83: I think that exiting from the context should call this automatically? (that could also be mentioned in the text to make this explicit)

PR #34511, opened 2023-03-09
GH-29105: [C++][Parquet] Relax schema checking when writing using StreamWriter

omrimallis:
### Rationale for this change

The converted type expected by StreamWriter when writing int and string values did not match the converted type typically set in the Parquet schema. When writing INT_32 and INT_64, a converted type NONE is typically set in the Parquet schema, but this caused a ParquetException to be raised. This error was encountered when using StreamWriter after converting an Arrow schema to Parquet schema via ToParquetSchema().

Closes #29105.

Relax the schema checking in StreamWriter. The fix was already present for StreamReader since 4d82549, and is now ported to StreamWriter as well, with the shared logic moved to PrimitiveNode.

### Are these changes tested?

Yes. A test was added to verify that StreamWriter is able to write `INT_32`, `INT_64` and `UTF8` types when the `ConvertedType` in the Parquet schema is set to `NONE`.

No.

* Closes: #29105

### DIFF
--- a/cpp/src/parquet/schema.cc
+++ b/cpp/src/parquet/schema.cc
@@ -20,6 +20,7 @@
 #include <algorithm>
 #include <cstring>
 #include <memory>
+#include <set>
 #include <string>
 #include <type_traits>
 #include <utility>
@@ -302,6 +303,34 @@ bool PrimitiveNode::Equals(const Node* other) const {
     return false;
   }
   return EqualsInternal(static_cast<const PrimitiveNode*>(other));
+}
+
+bool PrimitiveNode::CompatibleType(ConvertedType::type converted_type) const {
+  // The converted type in the parquet schema may not match exactly with the
+  // converted type of a value to be read or written. The following is a list
+  // of converted types which are allowed instead.
+  // Each pair given is:
+  //   {<type to be read or written>, <Parquet file converted type>}
+  // So for example {ConvertedType::INT_32, ConvertedType::NONE} means
+  // that a converted type INT_32 can be read or written to Parquet field with
+  // converted type NONE.
+  static const std::set<std::pair<ConvertedType::type, ConvertedType::type> >
+      converted_type_exceptions = {{ConvertedType::INT_32, ConvertedType::NONE},
+                                   {ConvertedType::INT_64, ConvertedType::NONE},
+                                   {ConvertedType::INT_32, ConvertedType::DECIMAL},
+                                   {ConvertedType::INT_64, ConvertedType::DECIMAL},
+                                   {ConvertedType::UTF8, ConvertedType::NONE}};
+
+  if (converted_type != converted_type_) {
+    // The converted type does not always match with the value
+    // provided so check the set of exceptions.
+    if (converted_type_exceptions.find({converted_type, converted_type_}) ==
+        converted_type_exceptions.end()) {
+      return false;
+    }
+  }
+
+  return true;
 }
 
 void PrimitiveNode::Visit(Node::Visitor* visitor) { visitor->Visit(this); }
--- a/cpp/src/parquet/schema.h
+++ b/cpp/src/parquet/schema.h
@@ -222,6 +222,9 @@ class PARQUET_EXPORT PrimitiveNode : public Node {
 
   bool Equals(const Node* other) const override;
 
+  // Check if a value of converted_type can be read or written to this Parquet node.
+  bool CompatibleType(ConvertedType::type converted_type) const;
+
   Type::type physical_type() const { return physical_type_; }
 
   ColumnOrder column_order() const { return column_order_; }
--- a/cpp/src/parquet/stream_reader.cc
+++ b/cpp/src/parquet/stream_reader.cc
@@ -17,30 +17,11 @@
 
 #include "parquet/stream_reader.h"
 
-#include <set>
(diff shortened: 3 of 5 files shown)


mapleFU (commented):
This code is ok for me. However, I think converted type is deprecated. Maybe we should use LogicalType to check within writer, and encapsulate the convert logic in reader? (I'm not so familiar with the standard, if I'm wrong, please correct me)

mapleFU:
Seems `LogicalType::is_compatible` maybe ok here?

omrimallis:
> Maybe we should use LogicalType to check within writer, and encapsulate the convert logic in reader?

Makes sense, converted type is only for backward compatibility according to the spec.

> Seems LogicalType::is_compatible maybe ok here?

I don't think that it can be used here because the `PrimitiveNode` that is being checked might have its `LogicalType` set to null (e.g. in case of a native type like int32 or int64). It is probably better to just do a quick comparison of `LogicalType::Type`. I can give it a try and push an update.

mapleFU:
I think that `PrimitiveNode` should have a logical type. Because it's built by writer, so we don't need some compability checkings. If it doesn't has, just fallback to convert type, but I think it's a bit trickey

wgtmac:
> This error was encountered when using StreamWriter after converting an Arrow schema to Parquet schema via ToParquetSchema().

IMHO, it makes sense to check compatibility when reading because the requested type can differ from what is stored in the file. But for writing, we can take fully control over the schema, so is it better to fix the schema conversion mentioned above?

omrimallis:
It seems to be the common practice to set the logical type (and converted type) to `NONE` for fields that represent 32 and 64 bit integers. The logical type is inferred from the physical type in this case. Changing the schema conversion code could solve the issue, however, new Parquet files will be created with their logical types set to `INT`, which seems to be against the common practice today.

mapleFU:
Hi @omrimallis 

There a two kinds of node during **write** mode:
1. `PrimitiveNode`: For primitive type, which would have physical type and correspond logical type. As you can see, in `PrimitiveNode::PrimitiveNode`, it will force set a logical_type
2. `GroupNode`: A node for Struct/Map/Array, it would have a logicalType, and never have a physical type.

So, I think when writing, we always have a physical type. Things get a bit complex during reading. It will parse type from arrow, and need to handle compability. But `Node` in memory would have logical type.

So, this patch relax checking when writing, but I guess logical_type is enough. If I'm wrong, please correct me

wgtmac:
As `ConvertedType` has been deprecated, what about changing `StreamWriter::CheckColumn` to check `LogicalType` instead of `ConvertedType`? Both the writer and schema are created by ourselves, so we are pretty sure which logical type to expect (including null).

BTW, the `StreamWriter` does not support nested types and we do not need to worry about the logical type of them.

amol-:

wgtmac:
@omrimallis Do you plan to update this?

mapleFU:
If you don't mind, I can write a fixing using `LogicalType` @omrimallis

omrimallis:
@mapleFU Sure, feel free to update it.

mapleFU:
By the way, I'm curious that why you should fixing this. When read from stale Parquet file, the parquet schema is parsed from  
 the old parquet file, so, it has stale `ConvertedType`. But why parquet writer has that type? I'm considering that can we just bypass that problem? Or how do you use this?

mapleFU:
And maybe logic could be reference from `parquet-mr/parquet-arrow/src/main/java/org/apache/parquet/arrow/schema/SchemaConverter.java`. If you think we need it, I can try to port it

omrimallis:
@mapleFU To my understanding, Parquet writers still have to encode `ConvertedType` for each field according to the Parquet spec. So when you create a `PrimitiveNode`, its `ConvertedType` is automatically inferred and set from its `LogicalType` (see the constructor of `PrimitiveNode`). The stream writer can therefore always rely on it being present, except in the special circumstances which this PR solves. We can update the code of `StreamWriter` to check against logical types instead, though I see no special benefit for it currently.

mapleFU:
Hi @omrimallis . Personally I'm ok for this patch. If you want, you can encapsulate in in common header for `stream_reader.cc` and `stream_writer.cc`. My view is that, for reader, we need to support some backward-capability rules. However, for writer, I don't think we should keep the rule, because writer schema is created by user, and user can deduce specifict convert types.

If you want, I think just put it to stream_reader and stream_writer is ok

mapleFU:
@pitrou What do you think of this patch? I think writer itself should have the right schema. Did I miss something here?

pitrou:
@mapleFU Sorry for the delay. Could you first rebase this and fix conflicts?

pitrou:
@mapleFU Are you still interested in taking this up?

mapleFU:
I'm so busy these few days, so maybe I don't have time these month.

For me, `StreamWriter` is a bit wierd,  because although it has a schema when opening the writer, but in fact, it checks depend on the `operator<<` and logical type. This is a bit weird for me. Another problem is that it checks the type every record, I don't think it's a good idea.

As for logical type, we cannot make sure that user is writing using LogicalType or deprecated only-converttype mode, so it makes it more complex, maybe we can draft how to redesign it with LogicalType later, so that nano seconds or others can be written, and type checkings could be easier. @pitrou

PR #34518, opened 2023-03-09
GH-15280: [C++] Refactor to reorganize dependencies as a prequel to moving acero out of libarrow

### LINKED ISSUE #15280: [C++] Move Acero out of libarrow

ildipo:
### Rationale for this change

In order to remove acero from libarrow we need to reorganize the code to reduce entanglement between compute/exec and the rest of the product by removing any dependency from the codebase (except dataset) into compute/exec.

This PR removes some of the dependencies:
* some of the functions in `compute/exec/util.[h,cc]` that are widely used are moved to a new `compute/util.[h,cc] `file
* `key_map` and `key_hash` are widely used and thus moved from compute/exec to compute
* some light_array tests do use functionalities inside `compute/exec`, so the tests are split into 2 parts, one inside `compute` and another inside `compute/exec`

### Are these changes tested?

The existing tests cover all these changes.
One additional test program has been added by splitting light_array_test in two parts.

no

* Closes: #15280


cpp/src/arrow/compute/util_internal.h
```diff
@@ -0,0 +1,31 @@
...
+
+#pragma once
+
+#include "arrow/util/logging.h"
+
+namespace arrow {
+namespace util {
+
+template <typename T>
+void CheckAlignment(const void* ptr) {
```
icexelloss, line 26: Curious why moving this from util.h to a new util_internal.h?

ildipo: this uses a macro that should not end up in the external include files.
util.h ends up in external include files (since we'll need it inside the future libacero)

cpp/src/arrow/compute/light_array_test.cc
```diff
@@ -216,27 +215,6 @@ TEST(KeyColumnArray, SliceBool) {
   }
 }
 
-TEST(KeyColumnArray, FromExecBatch) {
```
westonpace, line 219: Hmm, it's a bit odd to have these tests split this way.  Do you want to just move this entire test file to `exec`?  Maybe later we can, in a follow-up, attempt to pull out the utilities that are needed in `exec/test_util.h` so the entire test file can move out of `exec`.

However, for now, I think it's probably ok that this is tested in exec.

cpp/src/arrow/compute/util_internal.h
westonpace: Yes, internal is correct for this, good catch.  `exec/util.h` has always been "implicitly internal" and I've been wanting to rename it forever but hadn't quite gotten around to it.  This is a good start.

westonpace (approved):
This looks pretty reasonable to me.  I have one suggestion regarding the test.  Let me know what you want to do there and we can move forward.  Thanks for helping to organize this.

icexelloss (approved):
LGTM.

icexelloss:
@ildipo Is on vacation I will double check with him and I will merge

cpp/src/arrow/compute/light_array_test.cc
ildipo: In the coming CR I'm moving a few more things around so I'll change this to a single file as well.

PR #34526, opened 2023-03-10
GH-15107: [C++][Parquet] Parquet Encoder: Support RLE for Boolean

### LINKED ISSUE #15107: [C++][Parquet] Support RLE Encoder for Boolean type

Currently, in our spec, boolean type in data page can use RLE to encode itself. And parquet-mr, parquet-go and other versions all supports it.

Thanks https://github.com/apache/arrow/pull/14147 , we already have a Rle Boolean Decoder. So for testing, just testing it local is ok.

I'd like to submit a implemention and testing this weekend.

### Component(s)

C++, Parquet

mapleFU:
Create RLE Encoder for Boolean.

### Rationale for this change

Boolean current can only use Plain Encoder, here it support RLE.

Create encoder

### Are these changes tested?

Yes

Yes, user can use new kind of encoding.

* Closes: #15107


cpp/src/parquet/encoding.cc
```diff
@@ -2838,6 +2839,120 @@ class DeltaLengthByteArrayDecoder : public DecoderImpl,
...
+
+ protected:
+  template <typename ArrowType>
+  void PutImpl(const ::arrow::Array& values) {
+    if (values.type_id() != ::arrow::Type::BOOL) {
+      throw ParquetException(std::string() + "direct put to " + ArrowType::type_name() +
+                             " from " + values.type()->ToString() + " not supported");
+    }
+    const auto& data = *values.data();
+    PutSpaced(data.GetValues<typename ArrowType::c_type>(1),
```
wgtmac, line 2892: Should we check `values.null_count()` and go to a fast path if there is no null?

cpp/src/parquet/encoding.cc
```diff
@@ -2838,6 +2839,120 @@ class DeltaLengthByteArrayDecoder : public DecoderImpl,
...
+  /// 4 bytes in little-endian, which indicates the length.
+  constexpr static int32_t kRleLengthInBytes = 4;
+
+  // std::vector<bool> in C++ is tricky, because it's a bitmap.
+  // Here RleBooleanEncoder will only append values into it, and
+  // dump values into Buffer, so using it here is ok.
+  std::vector<bool> buffered_append_values_;
+};
+
+void RleBooleanEncoder::Put(const bool* src, int num_values) { PutImpl(src, num_values); }
```
wgtmac, line 2918: Should we support a `const uint8_t*` variant? The decoder has that support.

cpp/src/parquet/encoding.cc
```diff
@@ -2838,6 +2839,120 @@ class DeltaLengthByteArrayDecoder : public DecoderImpl,
...
+template <typename SequenceType>
+void RleBooleanEncoder::PutImpl(const SequenceType& src, int num_values) {
+  for (int i = 0; i < num_values; ++i) {
+    // TODO(mwish): using iterator to batch insert?
+    buffered_append_values_.push_back(src[i]);
+  }
+}
+
+int64_t RleBooleanEncoder::EstimatedDataEncodedSize() {
+  return kRleLengthInBytes + MaxRleBufferSize();
```
wgtmac, line 2933: This may have a risk of over estimation. Could we encode it in a streaming way?

cpp/src/parquet/encoding_test.cc
```diff
@@ -1576,6 +1576,60 @@ TYPED_TEST(TestDeltaBitPackEncoding, NonZeroPaddedMiniblockBitWidth) {
...
+    ASSERT_EQ(num_values_, values_decoded);
+    ASSERT_NO_FATAL_FAILURE(VerifyResultsSpaced<c_type>(decode_buf_, draws_, num_values_,
+                                                        valid_bits, valid_bits_offset));
+  }
+};
+
+TEST_F(TestRleBooleanEncoding, BasicRoundTrip) {
+  ASSERT_NO_FATAL_FAILURE(this->Execute(0, 0));
+  ASSERT_NO_FATAL_FAILURE(this->Execute(2000, 200));
+  ASSERT_NO_FATAL_FAILURE(this->ExecuteSpaced(
```
wgtmac, line 1628: Add a case to cover arrow array as input?

cpp/src/parquet/encoding.cc
mapleFU: Why, since it's just rle estimations

wgtmac: IIUC, max size is estimated for the worse case of bit packing and min size is estimated for repeated rle. So we have a risk of over estimation for the real data here.

cpp/src/parquet/encoding.cc
mapleFU: Seems `PlainEncoder` doesn't support that? I think u8 is ok but just leave it here first.

cpp/src/parquet/encoding.cc
mapleFU: Yes, the syntax for it it's really trickey. MaxSize even possible less than MinSize, and the estimation is really a mass.

Though may have a risk of over estimation, DictEncoder estimate value in this way. And if we want just an estimation, we can just use `size-of-values / 8` as an estimation.

Streaming Encoding needs to refactor `::arrow::util::RleEncoder`, I think it's a bit hard here. I can open an issue and do it later

cpp/src/parquet/encoding_test.cc
mapleFU: Arrow case cannot decode...Let me amend a Decode later

mapleFU:
```
C:\projects\arrow\cpp\build\src\parquet\CMakeFiles\parquet_shared.dir\Unity\unity_1_cxx.cxx
C:/projects/arrow/cpp/src/parquet/encoding.cc(2901): error C2220: the following warning is treated as an error
C:/projects/arrow/cpp/src/parquet/encoding.cc(2901): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data
C:/projects/arrow/cpp/src/parquet/encoding.cc(2901): warning C4244: 'return': conversion from 'int64_t' to 'int', possible loss of data
```

...

cpp/src/parquet/encoding.cc
```diff
@@ -2855,7 +2963,7 @@ class RleBooleanDecoder : public DecoderImpl, virtual public BooleanDecoder {
                              " (corrupt data page?)");
     }
     // Load the first 4 bytes in little-endian, which indicates the length
-    num_bytes = ::arrow::bit_util::ToLittleEndian(SafeLoadAs<uint32_t>(data));
```
mapleFU, line 2858: I think previously here could be strange, it should be `FromLittleEndian`

mapleFU:
@pitrou @emkornfield Mind take a look?

cpp/src/arrow/util/rle_encoding.h
```diff
@@ -209,8 +209,7 @@ class RleEncoder {
     // bit-packed repeated value
     int min_repeated_run_size =
         1 + static_cast<int>(::arrow::bit_util::BytesForBits(bit_width));
-    int repeated_max_size = static_cast<int>(::arrow::bit_util::CeilDiv(num_values, 8)) *
-                            min_repeated_run_size;
```
felipecrv, line 213: Removed code does seem wrong. Has this caused bugs?

cpp/src/parquet/encoding.cc
```diff
@@ -466,6 +465,15 @@ struct DictEncoderTraits<FLBAType> {
 // Initially 1024 elements
 static constexpr int32_t kInitialHashTableSize = 1 << 10;
 
+int RlePreserveBufferSize(int num_values, int bit_width) {
+  // Note: because of the way RleEncoder::CheckBufferFull() is called, we have to
+  // reserve
+  // an extra "RleEncoder::MinBufferSize" bytes. These extra bytes won't be used
+  // but not reserving them would cause the encoder to fail.
+  return 1 + ::arrow::util::RleEncoder::MaxBufferSize(bit_width, num_values) +
+         ::arrow::util::RleEncoder::MinBufferSize(bit_width);
```
felipecrv, line 474: You could avoid this extraction by creating a local variable called `num_values`.

Note that the caller returns `int64_t`, so you can also return `int64_t` and avoid a potential `int` overflow when summing `a + int + int`.

cpp/src/parquet/encoding.cc
```diff
@@ -2838,6 +2839,113 @@ class DeltaLengthByteArrayDecoder : public DecoderImpl,
...
+class RleBooleanEncoder final : public EncoderImpl, virtual public BooleanEncoder {
+ public:
+  explicit RleBooleanEncoder(const ColumnDescriptor* descr, ::arrow::MemoryPool* pool)
+      : EncoderImpl(descr, Encoding::RLE, pool) {}
+
+  int64_t EstimatedDataEncodedSize() override {
+    // FIXME(mwish): should we just use buffered_append_values_.size() / 8
+    //  or just use ::arrow::util::RleEncoder::MaxBufferSize?
+    return kRleLengthInBytes + MaxRleBufferSize();
+  }
```
felipecrv, line 2854: Do one pass over the data to count the number of runs.

cpp/src/arrow/util/rle_encoding.h
mapleFU: Nope, in fact, `static_cast<int>(::arrow::bit_util::CeilDiv(num_values, 8))` has occured twice, so for simplify, I rename it

cpp/src/parquet/encoding.cc
mapleFU: int64_t seems ok, but previously, the code using `RlePreserveBufferSize` never do a overflow checking. So I use int here.

What do you mean by creating local variable?

cpp/src/parquet/encoding.cc
```diff
@@ -2838,6 +2839,109 @@ class DeltaLengthByteArrayDecoder : public DecoderImpl,
...
+
+  constexpr static int32_t kBitWidth = 1;
+  /// 4 bytes in little-endian, which indicates the length.
+  constexpr static int32_t kRleLengthInBytes = 4;
+
+  // std::vector<bool> in C++ is tricky, because it's a bitmap.
+  // Here RleBooleanEncoder will only append values into it, and
+  // dump values into Buffer, so using it here is ok.
+  std::vector<bool> buffered_append_values_;
```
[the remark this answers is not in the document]
mapleFU: Nice catch, I think it won't be too large, because it's just a bitmap. But `ArrowPoolVector<bool>` seems much better?

cpp/src/parquet/encoding.cc
```diff
@@ -512,14 +520,7 @@ class DictEncoderImpl : public EncoderImpl, virtual public DictEncoder<DType> {
...
   int64_t EstimatedDataEncodedSize() override {
-    // Note: because of the way RleEncoder::CheckBufferFull() is called, we have to
-    // reserve
-    // an extra "RleEncoder::MinBufferSize" bytes. These extra bytes won't be used
-    // but not reserving them would cause the encoder to fail.
-    return 1 +
-           ::arrow::util::RleEncoder::MaxBufferSize(
-               bit_width(), static_cast<int>(buffered_indices_.size())) +
-           ::arrow::util::RleEncoder::MinBufferSize(bit_width());
+    return RlePreserveBufferSize(static_cast<int>(buffered_indices_.size()), bit_width());
```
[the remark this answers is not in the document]
mapleFU: They should be same, but they're used in different encoders

cpp/src/arrow/util/rle_encoding.h
felipecrv: Ahh. This change preserves equivalence.

cpp/src/parquet/encoding.cc
felipecrv: I supposes you extracted a new function here for clarity, but you call in multiple places now?

felipecrv: If the extract was just for the sake of clarify, creating a variable would be preferable. You can mark this comment thread as resolved. :)

cpp/src/parquet/encoding.cc
```diff
@@ -2838,6 +2839,109 @@ class DeltaLengthByteArrayDecoder : public DecoderImpl,
...
+        buffered_append_values_.push_back(boolean_array.Value(i));
+      }
+    } else {
+      PARQUET_THROW_NOT_OK(::arrow::VisitArraySpanInline<::arrow::BooleanType>(
+          *boolean_array.data(),
+          [&](bool value) {
+            buffered_append_values_.push_back(value);
+            return Status::OK();
+          },
+          []() { return Status::OK(); }));
```
felipecrv, line 2875: Not handling nulls?

cpp/src/parquet/encoding.cc
mapleFU: Currently I thinks keep it is ok. I can open a issue, and support write `PlainBooleanEncoder` and `RleBooleanEncoder` by u8 and bitmap later.

cpp/src/parquet/encoding.cc
mapleFU: Nulls, in parquet, in handled by multi-level-path writer. It write rep-levels and def-levels in upper layer.

Encoder only handles non-null variables. ColumnWriter may collect some stats for null.

mapleFU:
@rok @wjones127 PTAL :)

mapleFU:
@pitrou mind take a look?

wjones127 (approved):
This looks good to me.

wjones127:
@felipecrv Does this look good to you?

felipecrv:
@wjones127 I'm not familiar with this part of the codebase to really have an opinion. I was asking some questions to learn more about and trying to help with the review.

wjones127:
@felipecrv Just wanted to make sure you are satisfied with responses to your earlier comments before merging :)

cpp/src/parquet/encoding.cc
```diff
@@ -466,6 +466,15 @@ struct DictEncoderTraits<FLBAType> {
 // Initially 1024 elements
 static constexpr int32_t kInitialHashTableSize = 1 << 10;
 
+int RlePreserveBufferSize(int num_values, int bit_width) {
+  // Note: because of the way RleEncoder::CheckBufferFull() is called, we have to
+  // reserve
```
wgtmac, line 471: The format of comment looks strange.

cpp/src/parquet/encoding.cc
```diff
@@ -466,6 +466,15 @@ struct DictEncoderTraits<FLBAType> {
...
+  // an extra "RleEncoder::MinBufferSize" bytes. These extra bytes won't be used
+  // but not reserving them would cause the encoder to fail.
+  return 1 + ::arrow::util::RleEncoder::MaxBufferSize(bit_width, num_values) +
```
wgtmac, line 474: What does 1 mean here?

cpp/src/parquet/encoding.cc
wgtmac: Good catch

cpp/src/parquet/encoding.cc
```diff
@@ -2883,7 +2989,26 @@ class RleBooleanDecoder : public DecoderImpl, virtual public BooleanDecoder {
   int DecodeArrow(int num_values, int null_count, const uint8_t* valid_bits,
                   int64_t valid_bits_offset,
                   typename EncodingTraits<BooleanType>::Accumulator* out) override {
-    ParquetException::NYI("DecodeArrow for RleBooleanDecoder");
+    if (null_count != 0) {
+      ParquetException::NYI("RleBoolean DecodeArrow with null slots");
```
wgtmac, line 2993: Is this a TODO or misuse for `null_count != 0`?

wgtmac (commented):
Overall looking good.

cpp/src/parquet/encoding.cc
mapleFU: Yes...Here implement it just for testing. And later `DecodeArrow with null slots` not have todo. I may do it later

mapleFU:
@rok  does python tests have test for boolean encoding / decoding ? I didn't found it in `python/pyarrow/tests/parquet/test_basic.py`

cpp/src/parquet/encoding.cc
mapleFU: Pasted from `DictEncoder`, let me change it

cpp/src/parquet/encoding.cc
mapleFU: (In fact, in little endian machine, they are just nop, and in big endian machine, they just do byteswap, but the syntax here is ambigious)

cpp/src/parquet/encoding.cc
```diff
@@ -489,6 +489,10 @@ class DictEncoderImpl : public EncoderImpl, virtual public DictEncoder<DType> {
  public:
   typedef typename DType::c_type T;
 
+  /// In data page, the bit width used to encode the entry
+  /// ids stored as 1 byte (max bit width = 32).
+  constexpr static int32_t kDataPageBitWidthBytes = 1;
```
mapleFU, line 494: @wgtmac I extract that `1` to a `kDataPageBitWidthBytes` here.

wgtmac (approved):
LGTM. Thanks @mapleFU!

mapleFU:
```
 55/113 Test  #56: arrow-dataset-scanner-test ................   Passed    4.65 sec
        Start  57: arrow-dataset-file-csv-test
 56/113 Test  #46: arrow-threading-utility-test ..............   Passed   31.67 sec
        Start  58: arrow-dataset-file-json-test
 57/113 Test  #57: arrow-dataset-file-csv-test ...............***Failed   27.37 sec
Running main() from D:\bld\gtest_1679003944036\work\googletest\src\gtest_main.cc
[==========] Running 176 tests from 12 test suites.
[----------] Global test environment set-up.
```

Seems that https://ci.appveyor.com/project/ApacheSoftwareFoundation/arrow/builds/46564157 is unrelated to my patch. Can someone help rerun that case?

raulcd:
> Can someone help rerun that case?

I've triggered a re-run

mapleFU:
Thanks @raulcd

rok:
> @rok does python tests have test for boolean encoding / decoding ? I didn't found it in `python/pyarrow/tests/parquet/test_basic.py`

Sorry for the late reply. `test_column_encoding` indeed does't test boolean columns in RLE encoding. It would probably make sense to add it.

Issue #34527, opened 2023-03-10
pyarrow crashes Python on import on AMD CPUs


Nemecsek:
(It could be related to [issue #28041](https://github.com/apache/arrow/issues/28041))

`pyarrow 11.0.0` (and some previous versions) doesn't work on some AMD x86_64 CPUs such as AMD G-T40 and AMD GX-412TC.

`import pyarrow` crashes python with an error `Illegal instruction` (my free translation from `Ungültiger Maschinenbefehl` in German).
It is due to a **SSE feature** that is not available on the CPUs and is requested by `pyarrow` on import.

According to the [documentation](https://arrow.apache.org/docs/cpp/env_vars.html#cpp-env-vars) a possible solution is to set environment var `ARROW_USER_SIMD_LEVEL=NONE`. It looks like the initial check returns SSE is available for these AMD CPUs even if it is not:

>  **ARROW_USER_SIMD_LEVEL**
The SIMD optimization level to select. By default, Arrow C++ detects the capabilities of the current CPU at runtime and chooses the best execution paths based on that information. One can override the detection by setting this environment variable to a well-defined value. Supported values are:
> *  **NONE** disables any runtime-selected SIMD  
> *  **SSE4_2** enables any SSE2-based optimizations until SSE4.2 (
> *  **AVX** enables any AVX-based optimizations and earlier;
> *  **AVX2** enables any AVX2-based optimizations and earlier;
> * **AVX512** enables any AVX512-based optimizations and earlier.
> This environment variable only has an effect on x86 platforms.

[Here](https://arrow.apache.org/docs/python/env_vars.html) it is stated that 
 > Since PyArrow is based on Arrow C++, its behavior is also affected by the [Arrow C++ environment variables](https://arrow.apache.org/docs/cpp/env_vars.html#cpp-env-vars).

After a number of attempts, it doesn't work: `ARROW_USER_SIMD_LEVEL` is neglected, and the only solution looks like to be a lenghty recompilation of `pyarrow`.

Am I missing something?

### Component(s)

Python

cyb70289:
You have to rebuild pyarrow with sse disabled. See https://github.com/apache/arrow/issues/34268 and https://github.com/apache/arrow/issues/34277.

@abcbarryn shared the build steps https://github.com/apache/arrow/issues/34277#issuecomment-1449455571

Nemecsek:
@cyb70289, it is what I did as it was the only practical solution after hours of testing.
[...]

Issue #34268, opened 2023-02-20
[Python] Pyarrow throw illegal instruction immediately on import on some hardware platforms.


abcbarryn:
### Describe the bug, including details regarding any error messages, version, and platform.

Every version of pyarrow I have tried so far from the latest version back to version 3.0.0 exits with an illegal instruction after installation as soon as I execute:
import pyarrow

### Component(s)

Python

abcbarryn:
I was simply running pip install pyarrow, it appears to build and install fine on the affected systems, but that as soon as you try to import it, it throws an illegal instruction.
I went to the trrouble of installing Anaconda and ran: conda install pyarrow
I got the same result on this system:

(base) static1:~ # python
Python 3.9.13 (main, Aug 25 2022, 23:26:10) 
[GCC 11.2.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pyarrow
Illegal instruction
(base) static1:~ #

abcbarryn:
CPU info:
vendor_id : GenuineIntel
cpu family : 6
model : 15
model name : Intel(R) Xeon(R) CPU 5160 @ 3.00GHz
stepping : 6
microcode : 0xd2
cpu MHz : 3000.116
cache size : 4096 KB
4 CPUs with 2 cores each

Linux static1 3.16.7-48-desktop https://github.com/streamlit/streamlit/pull/1 SMP PREEMPT Wed Oct 26 18:09:22 UTC 2016 (8c027f6) x86_64 GNU/Linux

glibc-2.18-4.44.1.x86_64

gcc version 7.1.1 20170607 [gcc-7-branch revision 248970] (SUSE Linux)

abcbarryn:
I have tried the following python versions: 3.9.4, 3.9.13, 3.9.16, and 3.11 with the same result.

assignUser:
The pyarrow wheels are compiled assuming SSE4.2 is available. You CPU is quite old and does not support that instruction set. You will have to build pyarrow/libarrow yourself see the [docs](https://arrow.apache.org/docs/cpp/env_vars.html#envvar-ARROW_USER_SIMD_LEVEL) on how to control the SIMD level.

abcbarryn:
System information
CPU
vendor_id	: GenuineIntel
cpu family	: 6
model		: 15
model name	: Intel(R) Xeon(R) CPU            5160  @ 3.00GHz
stepping	: 6
microcode	: 0xd2
[...]

PR #34528, opened 2023-03-10
MINOR: [Docs][C++] Explain "Illegal Instruction" on legacy CPU

cyb70289:
### Rationale for this change
We saw "Illegal Instruction" bug repots from time to time. It happens on legacy x86_64 CPUs don't support SSE4.2.
User may misunderstand current document and think that setting `ARROW_USER_SIMD_LEVEL=NONE` will disable SSE instruction at runtime and fix the issue. Clearer explaination is necessary.

Update cpp document to explain "Illegal Instruction" issue and how to fix it.

### Are these changes tested?
No.

No.


cyb70289:
Two recent bug reports.
https://github.com/apache/arrow/issues/34268, https://github.com/apache/arrow/issues/34527

cyb70289:
cc @Nemecsek, @abcbarryn

kou (approved):
+1

PR #34534, opened 2023-03-10
GH-20408: [Go] Implement Encode and Decode functions for REE

### LINKED ISSUE #20408: [Go] Implement Encode and Decode functions for REE
**Reporter**: [Matthew Topol](https://issues.apache.org/jira/browse/ARROW-17713) / @zeroshade

<sub>**Note**: *This issue was originally created as [ARROW-17713](https://issues.apache.org/jira/browse/ARROW-17713). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

zeroshade:
Adding new `run_end_encode` and `run_end_decode` functions to the Go Compute library.

### Are these changes tested?
Yes

* Closes: #20408

### DIFF
--- a/go/arrow/compute/internal/kernels/vector_run_end_encode.go
+++ b/go/arrow/compute/internal/kernels/vector_run_end_encode.go
@@ -231,6 +231,8 @@ func (re *runEndEncodeFSB[R]) PreallocOutput(ctx *exec.KernelCtx, numOutput int6
 
 	valueBuffer := ctx.Allocate(re.width * int(numOutput))
 	reeType := arrow.RunEndEncodedOf(exec.GetDataType[R](), re.valueType)
+	out.Release()
+
 	*out = exec.ExecResult{
 		Type:   reeType,
 		Len:    re.inputLen,
--- a/go/arrow/compute/internal/kernels/vector_run_end_encode.go
+++ b/go/arrow/compute/internal/kernels/vector_run_end_encode.go
@@ -47,13 +47,9 @@ type RunEndsType interface {
 }
 
 func readFixedWidthVal[V exec.FixedWidthTypes](inputValidity, inputValues []byte, offset int64, out *V) bool {
-	if bitutil.BitIsNotSet(inputValidity, int(offset)) {
-		return false
-	}
-
 	sz := int64(unsafe.Sizeof(*out))
 	*out = *(*V)(unsafe.Pointer(&inputValues[offset*sz]))
-	return true
+	return bitutil.BitIsSet(inputValidity, int(offset))
 }
 
 func writeFixedWidthVal[V exec.FixedWidthTypes](result *exec.ExecResult, offset int64, valid bool, value V) {
@@ -66,12 +62,8 @@ func writeFixedWidthVal[V exec.FixedWidthTypes](result *exec.ExecResult, offset 
 }
 
 func readBoolVal(inputValidity, inputValues []byte, offset int64, out *bool) bool {
-	if bitutil.BitIsNotSet(inputValidity, int(offset)) {
-		return false
-	}
-
 	*out = bitutil.BitIsSet(inputValues, int(offset))
-	return true
+	return bitutil.BitIsSet(inputValidity, int(offset))
 }
 
 func writeBoolVal(result *exec.ExecResult, offset int64, valid bool, value bool) {
--- a/go/arrow/compute/internal/kernels/vector_run_end_encode.go
+++ b/go/arrow/compute/internal/kernels/vector_run_end_encode.go
@@ -1,0 +1,953 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build go1.18
+
+package kernels
+
+import (
+	"bytes"
+	"fmt"
+	"sort"
+	"unsafe"
+
+	"github.com/apache/arrow/go/v12/arrow"
+	"github.com/apache/arrow/go/v12/arrow/bitutil"
+	"github.com/apache/arrow/go/v12/arrow/compute/internal/exec"
+	"github.com/apache/arrow/go/v12/arrow/decimal128"
+	"github.com/apache/arrow/go/v12/arrow/decimal256"
+	"github.com/apache/arrow/go/v12/arrow/float16"
+	"github.com/apache/arrow/go/v12/arrow/internal/debug"
+	"github.com/apache/arrow/go/v12/arrow/memory"
+)
+
+type RunEndEncodeState struct {
+	RunEndType arrow.DataType
+}
+
+func (RunEndEncodeState) TypeName() string {
+	return "RunEndEncodeOptions"
+}
+
+type RunEndsType interface {
+	int16 | int32 | int64
+}
+
+func readFixedWidthVal[V exec.FixedWidthTypes](inputValidity, inputValues []byte, offset int64, out *V) bool {
+	if bitutil.BitIsNotSet(inputValidity, int(offset)) {
+		return false
+	}
+
+	sz := int64(unsafe.Sizeof(*out))
+	*out = *(*V)(unsafe.Pointer(&inputValues[offset*sz]))
+	return true
+}
+
+func writeFixedWidthVal[V exec.FixedWidthTypes](result *exec.ExecResult, offset int64, valid bool, value V) {
+	if len(result.Buffers[0].Buf) != 0 {
+		bitutil.SetBitTo(result.Buffers[0].Buf, int(offset), valid)
+	}
+
+	arr := exec.GetData[V](result.Buffers[1].Buf)
+	arr[offset] = value
+}
+
+func readBoolVal(inputValidity, inputValues []byte, offset int64, out *bool) bool {
+	if bitutil.BitIsNotSet(inputValidity, int(offset)) {
+		return false
+	}
+
+	*out = bitutil.BitIsSet(inputValues, int(offset))
+	return true
+}
+
+func writeBoolVal(result *exec.ExecResult, offset int64, valid bool, value bool) {
+	if len(result.Buffers[0].Buf) != 0 {
+		bitutil.SetBitTo(result.Buffers[0].Buf, int(offset), valid)
+	}
+	bitutil.SetBitTo(result.Buffers[1].Buf, int(offset), value)
+}
+
+type runEndEncodeLoopFixedWidth[R RunEndsType, V exec.FixedWidthTypes | bool] struct {
+	inputLen, inputOffset int64
+	inputValidity         []byte
+	inputValues           []byte
+	valueType             arrow.DataType
+
+	readValue  func(inputValidity, inputValues []byte, offset int64, out *V) bool
+	writeValue func(*exec.ExecResult, int64, bool, V)
+}
+
+func (re *runEndEncodeLoopFixedWidth[R, V]) WriteEncodedRuns(out *exec.ExecResult) int64 {
+	outputRunEnds := exec.GetData[R](out.Children[0].Buffers[1].Buf)
+
+	readOffset := re.inputOffset
+	var currentRun V
+	curRunValid := re.readValue(re.inputValidity, re.inputValues, readOffset, &currentRun)
+	readOffset++
+
+	var writeOffset int64
+	for readOffset < re.inputOffset+re.inputLen {
+		var value V
+		valid := re.readValue(re.inputValidity, re.inputValues, readOffset, &value)
+		if valid != curRunValid || value != currentRun {
+			// close the current run by writing it out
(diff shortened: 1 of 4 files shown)


zeroshade:
CC @felipecrv so we can make sure that I didn't miss anything and that this is on par with the C++ impl

go/arrow/compute/vector_run_ends.go
```diff
@@ -0,0 +1,90 @@
...
+
+	return out.(*ArrayDatum).MakeArray(), nil
+}
+
+func RunEndDecode(ctx context.Context, arg Datum) (Datum, error) {
+	return CallFunction(ctx, "run_end_decode", nil, arg)
+}
+
+func RunEndDecodeArray(ctx context.Context, values arrow.Array) (arrow.Array, error) {
+	out, err := RunEndDecode(ctx, &ArrayDatum{Value: values.Data()})
```
felipecrv, line 83: It would be less confusing to call `values` just `input` or `input_array` here since "values" clashes with the name of the fields of the REEs.

zeroshade: that's a good point, i'll fix that.

go/arrow/compute/internal/kernels/vector_run_end_encode.go
```diff
@@ -0,0 +1,953 @@
...
+	int16 | int32 | int64
+}
+
+func readFixedWidthVal[V exec.FixedWidthTypes](inputValidity, inputValues []byte, offset int64, out *V) bool {
+	if bitutil.BitIsNotSet(inputValidity, int(offset)) {
+		return false
+	}
+
+	sz := int64(unsafe.Sizeof(*out))
+	*out = *(*V)(unsafe.Pointer(&inputValues[offset*sz]))
```
felipecrv, line 55: This could always be done without waiting for the load in `BitIsNotSet`, right? Less data dependency.

go/arrow/compute/internal/kernels/vector_run_end_encode.go
```diff
@@ -0,0 +1,953 @@
...
+	var valueBuffer *memory.Buffer
+	bufSpec := re.valueType.Layout().Buffers[1]
+	if bufSpec.Kind == arrow.KindBitmap {
+		valueBuffer = ctx.AllocateBitmap(numOutput)
+	} else {
+		valueBuffer = ctx.Allocate(int(numOutput) * bufSpec.ByteWidth)
+	}
+
+	reeType := arrow.RunEndEncodedOf(exec.GetDataType[R](), re.valueType)
+	out.Release()
```
felipecrv, line 167: Go question: re-assigning to `out` doesn't automatically Release it?

go/arrow/compute/internal/kernels/vector_run_end_encode.go
zeroshade: fair point, I'm not sure if the Go compiler will optimize for this particular case, but it's worthwhile to help the CPU pipeline. I'll reorder this

go/arrow/compute/internal/kernels/vector_run_end_encode.go
zeroshade: it doesn't call the `Release()` method. Go doesn't have destructors.

zeroshade: it's a case of wanting to propagate. Honestly, this is *technically* unnecessary since I have the kernels set to not preallocate anything, but I'm doing this to make sure that a future change doesn't accidentally cause a memory leak or something.

go/arrow/compute/internal/kernels/vector_run_end_encode.go
```diff
@@ -0,0 +1,945 @@
...
+
+func (re *runEndEncodeFSB[R]) PreallocOutput(ctx *exec.KernelCtx, numOutput int64, out *exec.ExecResult) {
+	runEndsBuffer := ctx.Allocate(int(numOutput) * int(SizeOf[R]()))
+	var validityBuffer *memory.Buffer
+	if len(re.inputValidity) > 0 {
+		validityBuffer = ctx.AllocateBitmap(numOutput)
+	}
+
+	valueBuffer := ctx.Allocate(re.width * int(numOutput))
+	reeType := arrow.RunEndEncodedOf(exec.GetDataType[R](), re.valueType)
```
felipecrv, line 233: out.Release() on this one also?

zeroshade: good catch!

go/arrow/compute/internal/kernels/vector_run_end_encode.go
```diff
@@ -0,0 +1,947 @@
...
+	out.Children[1].Buffers[1].WrapBuffer(offsetsBuffer)
+	out.Children[1].Buffers[2].WrapBuffer(valueBuffer)
+}
+
+func (re *runEndEncodeLoopBinary[R, O]) WriteEncodedRuns(out *exec.ExecResult) int64 {
+	outputRunEnds := exec.GetData[R](out.Children[0].Buffers[1].Buf)
+	outputOffsets := exec.GetSpanOffsets[O](&out.Children[1], 1)
+	outputValues := out.Children[1].Buffers[2].Buf
+
+	var readOffset int64
```
felipecrv, line 395: This is starting from `0` when it should start from the offset of the input.

go/arrow/compute/internal/kernels/vector_run_end_encode.go
```diff
@@ -0,0 +1,947 @@
...
+
+	outputOffsets[0], outputOffsets = 0, outputOffsets[1:]
+
+	writeValue := func(valid bool, value []byte) {
+		setValidity(valid)
+		valueOffset += int64(copy(outputValues[valueOffset:], value))
+		outputOffsets[writeOffset] = O(valueOffset)
+	}
+
+	for readOffset < re.inputLen {
```
felipecrv, line 416: `readOffset < re.inputOffset + re.inputLen`

Either that or I'm missing something about Binary array types.

go/arrow/compute/internal/kernels/vector_run_end_encode.go
```diff
@@ -0,0 +1,947 @@
...
+	}
+
+	writeValue(curRunValid, currentRun)
+	outputRunEnds[writeOffset] = R(re.inputLen)
+	return writeOffset + 1
+}
+
+func validateRunEndType[R RunEndsType](length int64) error {
+	runEndMax := MaxOf[R]()
+	if length > int64(runEndMax) {
```
felipecrv, line 436: In my PR I also added a check against negative values after you pointed out this could also be an issue.

go/arrow/compute/internal/kernels/vector_run_end_encode.go
```diff
@@ -0,0 +1,947 @@
...
+	}
+}
+
+func createVarBinaryEncoder[R RunEndsType, O int32 | int64](input *exec.ArraySpan) *runEndEncodeLoopBinary[R, O] {
+	return &runEndEncodeLoopBinary[R, O]{
+		inputLen:      input.Len,
+		inputOffset:   input.Offset,
+		inputValidity: input.Buffers[0].Buf,
+		inputValues:   input.Buffers[2].Buf,
+		offsetValues:  exec.GetSpanOffsets[O](input, 1),
```
felipecrv, line 461: Does this apply `input.Offset`? That makes it a bit too different from the other buffers pointers in this struct.

go/arrow/compute/internal/kernels/vector_run_end_encode.go
```diff
@@ -0,0 +1,947 @@
...
+			inputLen:      input.Len,
+			inputOffset:   input.Offset,
+			inputValidity: input.Buffers[0].Buf,
+			inputValues:   input.Buffers[1].Buf,
+			valueType:     input.Type,
+			readValue:     readBoolVal,
+			writeValue:    writeBoolVal,
+		}
+	// for the other fixed size types, we only need to
+	// handle the different physical representations.
```
felipecrv, line 479: I should send a follow-up PR on the C++ implementation applying this trick.

go/arrow/compute/internal/kernels/vector_run_end_encode.go
```diff
@@ -0,0 +1,947 @@
...
+		writeOffset         int64
+		runLength, numValid int64
+		outputValues        = out.Buffers[1].Buf
+		prevRunEnd          = R(de.inputOffset)
+	)
+
+	for i, runEnd := range de.inputRunEnds[de.inputPhysicalOffset:] {
+		runLength, prevRunEnd = int64(runEnd-prevRunEnd), runEnd
+		// if this run is a null, clear the bits and update writeOffset
+		if len(de.inputValidity) != 0 && bitutil.BitIsNotSet(de.inputValidity, int(de.inputValueOffset+de.inputPhysicalOffset)+i) {
```
felipecrv, line 611: `inputValueOffset` could be just `inputOffset` here, right? It refers to the logical offset of the parent array. (?)

go/arrow/compute/internal/kernels/vector_run_end_encode.go
```diff
@@ -0,0 +1,947 @@
...
+			bitutil.SetBitsTo(out.Buffers[0].Buf, writeOffset, runLength, false)
+			writeOffset += runLength
+			continue
+		}
+
+		// if the output has a validity bitmap, update it with 1s
+		if len(out.Buffers[0].Buf) != 0 {
```
felipecrv, line 618: This condition and `len(de.inputValidity)` could become a single boolean set outside the loop.

go/arrow/compute/internal/kernels/vector_run_end_encode.go
zeroshade: Not quite. the slice of the offsets `re.offsetValues` is already offset based on the offset of the input because `exec.GetSpanOffsets` which is called in `createVarBinaryEncoder` will apply the internal offset itself.

go/arrow/compute/internal/kernels/vector_run_end_encode.go
zeroshade: good point! i'll add that

go/arrow/compute/internal/kernels/vector_run_end_encode.go
zeroshade: one complexity of binary arrays is that the offset values are always absolute offsets into the raw data. So in order for everything to work correctly, the slice for the raw values (`inputValues`) can't be shifted by the offset. In addition, you can't easily shift the validity bitmap in a predictable way because it's a bitmap and you aren't guaranteed the offset is easily divisible by 8. So those can't utilize the offset built in.

However, `exec.GetSpanOffsets` already accounts for the offset when creating the typed slice of offsets allowing the simple change of just not needing to apply the offset to the `readOffset` in the loop for var binary arrays.

go/arrow/compute/internal/kernels/vector_run_end_encode.go
zeroshade: * `de.inputOffset` is the logical offset of the parent array
* `de.inputPhysicalOffset` is the physical offset into the values and runs children arrays
* `de.inputValueOffset` is the internal offset of the values child array

For example, let's say you have the array `[1, 2, 3, 4, 5]`. Then let's take a slice of that, `[2, 3, 4]` and use that as our `values` for a run-end-encoded array. Finally let's add a logical offset to the REE array.

To make up some runs, that gives us the final REE array of:

* Parent: logical offset = 20, len = 30
    * Runs: offset = 0, len = 3, `[15, 30, 50]`
    * Values: offset = 1, len = 3, `[1, 2, 3, 4, 5]`

To get the proper values for the runs, we first find the physical offset (`1`) from the logical offset. Then when we index into the `values` child, we have to also apply the child's offset (`1`). Giving us the first value of `3` (index = `2`). This of course also applies to determining the bitmap validity for nulls.

Does that make sense? This is also tested in the unit tests.

go/arrow/compute/internal/kernels/vector_run_end_encode.go
zeroshade: fair point, we only set `out.Buffers[0].Buf` if `len(de.inputValidity) != 0`.

go/arrow/compute/internal/kernels/vector_run_end_encode.go
felipecrv: Oh right. I encapsulated much of this in a forward iterator in the C++ version and had forgotten there are 3 offset variables to address a physical value sometimes :)

go/arrow/compute/internal/kernels/vector_run_end_encode.go
zeroshade: the same is also true for `exec.GetSpanValues`

The difference between `GetSpanValues` and `GetSpanOffsets` is that `GetSpanOffsets` adds the expected +1 to the length that offsets should have. so `len(exec.GetSpanOffsets(....)) == span.Len+1`

Also, `exec.GetSpanOffsets` doesn't do the `nil` check that `GetSpanValues` does.

go/arrow/compute/internal/kernels/vector_run_end_encode.go
zeroshade: yea, Go doesn't really have `iterators` so there's just the `merged runs` iterator thing, but that's designed for iterating across 2 REE arrays simultaneously and getting the runs that line up rather than just iterating a single REE array. Maybe one day Go will provide a way to hook into the `for ... := range` construction for custom types lol

felipecrv (approved):

go/arrow/compute/internal/kernels/vector_run_end_encode.go
felipecrv: > so there's just the merged runs iterator thing, but that's designed for iterating across 2 REE arrays simultaneously

I took that thigk apart in the C++ version and made a similar class that takes a single `ArraySpan` and allows iterating on the runs. It's not completely C++ iterator compatible, so I think a similar thing could be achieved with Go. Perhaps Go doesn't inline as aggressively and it wouldn't be as efficient.

PR #34549, opened 2023-03-13
GH-34546: [C++] Support casting from large string to string scalar

### LINKED ISSUE #34546: [C++] Hive partitioning values contain newlines for large_string

pyarrow: 11.0.0
python: 3.9

I've been trying to create (and then write) a dataset using pyarrow:

```
file_options = ds.ParquetFileFormat().make_write_options(compression="zstd")
partitions = ds.partitioning(
    schema=pyarrow.schema([
        ("partition1", pyarrow.large_string()),
        ("partition2", pyarrow.large_string()),
        ("3", pyarrow.large_string()),
    ])
dataset = ds.dataset(
    f"gs://{bucket}/path1/",
    partitioning=partitions,
)
table = dataset.scanner().to_table()
ds.write_dataset(
    table,
    f"gs://{bucket}/path1/",
    format="parquet",
    file_options=file_options,
    partitioning=partitions,
    schema=table.schema,
    existing_data_behavior="overwrite_or_ignore",
)
```

This code never seems to turn the partitions into the actual expected strings:

```
pyarrow.lib.ArrowInvalid: google::cloud::Status(INVALID_ARGUMENT: Permanent error CreateResumableUpload: Disallowed unicode characters present in object name 'path1/[
  "partition1_value"
]/[
  "partition2_value"
]/[
  "3_value"
]/part-0.par...' error_info={reason=invalid, domain=global, metadata={http_status_code=400}}). Detail: [errno 22] Invalid argument
```

Am I doing something wrong here? I _thought_ I was following the documentation...

### Component(s)

Parquet, Python

wjones127:
### Rationale for this change

We rely on scalar casting to create partition values. Some systems, such as Polars, always use LargeString arrays instead of String, so we need to make sure we can handle those partition values correctly.

Adds casting function for `LargeStringScalar` and `LargeBinaryScalar` to `StringScalar`.
Tests that we correctly format Hive partition values for large string.

### Are these changes tested?

The casting functions are implicitly tested in other places (like the test added). There are no existing tests for `CastTo` for binary/string scalars.

Fixes a bug.

I guess it is a breaking change, given that Ruby was testing that the serialized value of a LargeBinaryScalar was `"[\n{value}\n]"`. But I don't think that was actually something anyone really wanted, was it?

* Closes: #34546

### DIFF
--- a/c_glib/test/test-large-binary-scalar.rb
+++ b/c_glib/test/test-large-binary-scalar.rb
@@ -38,11 +38,7 @@ class TestLargeBinaryScalar < Test::Unit::TestCase
   end
 
   def test_to_s
-    assert_equal(<<-BINARY.strip, @scalar.to_s)
-[
-  030102
-]
-                 BINARY
+    assert_equal("\x03\x01\x02", @scalar.to_s)
   end
 
   def test_value
--- a/c_glib/test/test-large-string-scalar.rb
+++ b/c_glib/test/test-large-string-scalar.rb
@@ -38,11 +38,7 @@ class TestLargeStringScalar < Test::Unit::TestCase
   end
 
   def test_to_s
-    assert_equal(<<-STRING.strip, @scalar.to_s)
-[
-  "Hello"
-]
-                 STRING
+    assert_equal("Hello", @scalar.to_s)
   end
 
   def test_value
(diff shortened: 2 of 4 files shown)


kou (approved):
+1

westonpace (commented):
Nice catch, thanks for the addition!

PR #34555, opened 2023-03-14
GH-34556: [C++] Add prefix 'ACT_' for 'arrow::Compression::type' to avoid naming conflicts

### LINKED ISSUE #34556: [C++] 'arrow::Compression::type' naming conflicts occur easily

Compression name (arrow::Compression::type) is very easy to conflicts when using arrow library in others project.

### Component(s)

C++

deepzliu:
Because compression name is very easy to conflicts when using arrow library in others project.
* Closes: #34556

### DIFF
--- a/cpp/src/arrow/util/type_fwd.h
+++ b/cpp/src/arrow/util/type_fwd.h
@@ -45,17 +45,18 @@ struct Scope;
 
 struct Compression {
   /// \brief Compression algorithm
+  //   prefix 'ACT'(arrow compression type) to avoid nameing conflicts.
   enum type {
-    UNCOMPRESSED,
-    SNAPPY,
-    GZIP,
-    BROTLI,
-    ZSTD,
-    LZ4,
-    LZ4_FRAME,
-    LZO,
-    BZ2,
-    LZ4_HADOOP
+    ACT_UNCOMPRESSED,
+    ACT_SNAPPY,
+    ACT_GZIP,
+    ACT_BROTLI,
+    ACT_ZSTD,
+    ACT_LZ4,
+    ACT_LZ4_FRAME,
+    ACT_LZO,
+    ACT_BZ2,
+    ACT_LZ4_HADOOP
   };
 };
 
--- a/c_glib/arrow-glib/codec.cpp
+++ b/c_glib/arrow-glib/codec.cpp
@@ -200,21 +200,21 @@ GArrowCompressionType
 garrow_compression_type_from_raw(arrow::Compression::type arrow_type)
 {
   switch (arrow_type) {
-  case arrow::Compression::type::UNCOMPRESSED:
+  case arrow::Compression::type::ACT_UNCOMPRESSED:
     return GARROW_COMPRESSION_TYPE_UNCOMPRESSED;
-  case arrow::Compression::type::SNAPPY:
+  case arrow::Compression::type::ACT_SNAPPY:
     return GARROW_COMPRESSION_TYPE_SNAPPY;
-  case arrow::Compression::type::GZIP:
+  case arrow::Compression::type::ACT_GZIP:
     return GARROW_COMPRESSION_TYPE_GZIP;
-  case arrow::Compression::type::BROTLI:
+  case arrow::Compression::type::ACT_BROTLI:
     return GARROW_COMPRESSION_TYPE_BROTLI;
-  case arrow::Compression::type::ZSTD:
+  case arrow::Compression::type::ACT_ZSTD:
     return GARROW_COMPRESSION_TYPE_ZSTD;
-  case arrow::Compression::type::LZ4:
+  case arrow::Compression::type::ACT_LZ4:
     return GARROW_COMPRESSION_TYPE_LZ4;
-  case arrow::Compression::type::LZO:
+  case arrow::Compression::type::ACT_LZO:
     return GARROW_COMPRESSION_TYPE_LZO;
-  case arrow::Compression::type::BZ2:
+  case arrow::Compression::type::ACT_BZ2:
     return GARROW_COMPRESSION_TYPE_BZ2;
   default:
     return GARROW_COMPRESSION_TYPE_UNCOMPRESSED;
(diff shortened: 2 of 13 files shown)


cpp/src/arrow/util/type_fwd.h
```diff
@@ -45,17 +45,18 @@ struct Scope;
 
 struct Compression {
   /// \brief Compression algorithm
+  //   prefix 'ACT'(arrow compression type) to avoid nameing conflicts.
   enum type {
```
kou, line 49: Can we use `enum class` instead of adding a prefix?

```suggestion
  enum class type {
```

deepzliu: OK, let me try.

deepzliu: 'enum class type {' has the same error if macro definition before, like:`#define SNAPPY 1`.
The error is 'expected identifier before numeric constant'.
Total sample code:
```
#include <iostream>

#define SNAPPY 1

enum class Compression {
  /// \brief Compression algorithm
  UNCOMPRESSED,
  SNAPPY,
  GZIP,
  BROTLI,
  ZSTD,
  LZ4,
  LZ4_FRAME,
  LZO,
  BZ2,
  LZ4_HADOOP
};

int main() {

  std::cout << "UNCOMPRESSED: " << static_cast<int>(Compression::UNCOMPRESSED) << std::endl;
  std::cout << "SNAPPY: " << static_cast<int>(Compression::SNAPPY) << std::endl;

  return 0;
}
```

deepzliu: Ref: https://github.com/fmtlib/fmt/commit/8ed264fcd49de5716204e77c6293e1a7aef46906

mapleFU (commented):
Generally I'm -1 for this. I think it would changes too many interface here.

And if possible, I think enum class is much better

wgtmac (changes requested):
Please use enum class as suggested by other reviewers. The current change may break public APIs.

deepzliu:
(My English is not good, with Chinese at the back.)

I describe the problem I encountered in detail. When I used both Arrow and RocksDB in my project, I encountered definition conflicts such as LZ4 and SNAPPY. In RocksDB, it is a macro definition, and in Arrow, it is an enumeration.

I personally think that in practical applications, this kind of scenario may be more common, it，it's worth to solving this problem.

The solutions are as follows:
1) Use some kind of technology to solve it completely with minimal impact, such as the enum class everyone mentioned. According to the my verification results, cannot solve the problem .
2) Modify the macro definitions of other libraries. However, in my opinion, the high maintenance cost and inconvenient usage are not conducive to promotion and application of Arrow library, especially in my case.
3) Add a prefix or suffix to the Arrow library enumeration, such as my PR, the disadvantages are clearly explained;
4) The user ensures that the Arrow header file is referenced at the front of others in the project. The problem with this method is that, a) the quality of developers is very variable and difficult to guarantee; b) if you use the macro definition in the make parameter, there is no solution;

To sum up, I personally think that solutions 3 may be a method that can fundamentally solve this problem.

---

我详细描述下我遇到的问题。在我的项目中同时用到了Arrow和RocksDB时，遇到LZ4、SNAPPY等定义冲突，在RocksDB是宏定义，在Arrow中是枚举。

个人认为，在实际应用中，这种场景可能会比较常见，所以解决这个问题有一定的价值。

能想到的解决方法有如下：
1）使用某种技术彻底解决，且影响最小，如大家提到的enum class。从验证结果来看，无法解决该问题；
2）改掉其他库的宏定义。但个人认为维护成本高、使用不便捷，不利于Arrow库广泛推广与应用，尤其是我遇到的情况；
3）在Arrow库枚举上加前缀或后缀，如本PR，弊端大家说得很清楚了；
4）使用者保证在项目中Arrow头文件引用在最前面。这个方法的问题在于，a）开发者水平参差不齐，很难保证；b）如果使用make参数中的宏定义，则无解；

综上所述，个人认为方法3可能是能根本解决该问题的方法。

mapleFU:
How do you use arrow compression? Are you using C++ SDK or C SDK? If you're using C++ sdk, the name and namespace would be like `arrow::util::Compression` and `parquet::Compression`. How did you meet conflict here?

Seems that if you code like:

```
using namespace rocksdb;
using namespace parquet;
using namespace arrow::util;
```

It's likely to conflict. But that's not the reason to change the name for it.

deepzliu:
>

deepzliu:
> How do you use arrow compression? Are you using C++ SDK or C SDK? If you're using C++ sdk, the name and namespace would be like `arrow::util::Compression` and `parquet::Compression`. How did you meet conflict here?
> 
> Seems that if you code like:
> 
> ```
> using namespace rocksdb;
> using namespace parquet;
> using namespace arrow::util;
> ```
> 
> It's likely to conflict. But that's not the reason to change the name for it.

namespace is not working for the problem.

wgtmac:
I don't know how does your project use these third-party libraries. Generally it is good practice to follow this: https://www.appsloveworld.com/cplus/100/232/conflict-caused-by-define-macro-and-enum-using-same-name

mapleFU:
https://github.com/apache/arrow/pull/34555#issuecomment-1475533675 @deepzliu Sorry for previously misunderstand this patch. Seems that rocksdb uses code like:

```
#ifdef SNAPPY
#include <snappy.h>
#endif

#ifdef ZLIB
#include <zlib.h>
#endif
```

And it conflict with arrow's definition. @kou What's the style for arrow's enum? Can we just use `kZStd` or use the fixing in this patch? And would it effect the C and other languages api?

kou:
In general, we use the Google C++ Style. It says:

https://google.github.io/styleguide/cppguide.html#Enumerator_Names

> Enumerators (for both scoped and unscoped enums) should be named like [constants](https://google.github.io/styleguide/cppguide.html#Constant_Names), not like [macros](https://google.github.io/styleguide/cppguide.html#Macro_Names). That is, use kEnumName not ENUM_NAME.

But our code doesn't follow the style. (It seems that `clang-format` doesn't support it.)

In general, we should use `kEnumName` style but it breaks many APIs. We should discuss this on `dev@arrow.apache.org`. Could you start a discussion on `dev@`?

Anyway, we can avoid the problem by undefing conflicting macros like the followings:

```c++
#include <rocksdb/...>

#undef SNAPPY
#undef ZLIB
#undef ...

#include <arrow/...>
```

deepzliu:
> In general, we use the Google C++ Style. It says:
> 
> https://google.github.io/styleguide/cppguide.html#Enumerator_Names
> 
> > Enumerators (for both scoped and unscoped enums) should be named like [constants](https://google.github.io/styleguide/cppguide.html#Constant_Names), not like [macros](https://google.github.io/styleguide/cppguide.html#Macro_Names). That is, use kEnumName not ENUM_NAME.
> 
> But our code doesn't follow the style. (It seems that `clang-format` doesn't support it.)
> 
> In general, we should use `kEnumName` style but it breaks many APIs. We should discuss this on `dev@arrow.apache.org`. Could you start a discussion on `dev@`?
> 
> Anyway, we can avoid the problem by undefing conflicting macros like the followings:
> 
> ```c++
> #include <rocksdb/...>
> 
> #undef SNAPPY
> #undef ZLIB
> #undef ...
> 
> #include <arrow/...>
> ```

This is a good solution, but it should not solve some complex project situation, not specific at the moment.

mapleFU:
@deepzliu I know that what you mean, however, if enum is changed, every user using arrow will not have backward-capability. So we may Could you start a discussion on dev@?

deepzliu:
> @deepzliu I know that what you mean, however, if enum is changed, every user using arrow will not have backward-capability. So we may Could you start a discussion on dev@?

How to start a discussion on dev@? never do this before.

kou:
OK. I made this pull request draft for now.

PR #34558, opened 2023-03-14
GH-34572: [Go][CSV] Add binary support for CSV

### LINKED ISSUE #34572: [Go][CSV] Support Binary Types

Right now CSV package (reader and writer) don't support binary types and return a panic if it's encountered in the schema. We want the reader,write to support it and use the base64 encoding.

### Component(s)

Go

yevgenypats:
Adding binary support for CSV
* Closes: #34572


zeroshade:
@yevgenypats Can you create a corresponding issue for this and then update the title to be `GH-<issue#>: [Go] Add binary support for CSV` thanks!

go/arrow/csv/reader.go
```diff
@@ -469,6 +469,10 @@ func (r *Reader) initFieldConverter(bldr array.Builder) func(string) {
 		return func(s string) {
 			r.parseList(bldr, s)
 		}
+	case arrow.BinaryDataType:
+		return func(s string) {
+			r.parseBinaryDataType(bldr, s)
+		}
```
zeroshade, line 475: Can you add a case for `*arrow.LargeStringType` above this (such as with the string case) so that it doesn't get dropped into this case too? (I'm pretty sure the LargeString type does meet this interface)

go/arrow/csv/reader.go
```diff
@@ -469,6 +469,10 @@ func (r *Reader) initFieldConverter(bldr array.Builder) func(string) {
 		return func(s string) {
 			r.parseList(bldr, s)
 		}
+	case arrow.BinaryDataType:
+		return func(s string) {
+			r.parseBinaryDataType(bldr, s)
```
zeroshade, line 474: Like the string case, we probably want to check the `r.stringsCanBeNull` value to determine whether we want to check for nulls or just let it be empty values etc.

go/arrow/csv/reader.go
```diff
@@ -755,6 +759,18 @@ func (r *Reader) parseList(field array.Builder, str string) {
 	}
 }
 
+func (r *Reader) parseBinaryDataType(field array.Builder, str string) {
+	if r.isNull(str) {
+		field.AppendNull()
+		return
+	}
+	decodedVal, err := base64.StdEncoding.DecodeString(str)
```
zeroshade, line 767: do we want to check both the padded and non-padded encodings? (ie: both `StdEncoding` and `RawStdEncoding`, checking the other if the first one errors)

If we don't check both encodings, we should make sure the documentation states which we use.

go/arrow/csv/testdata/types.csv
```diff
@@ -15,7 +15,7 @@
...
 #
-## supported types: bool;int8;int16;int32;int64;uint8;uint16;uint32;uint64;float32;float64;string;timestamp
-true;-1;-1;-1;-1;1;1;1;1;1.1;1.1;str-1;2022-05-09T00:01:01;{1,2,3}
-false;-2;-2;-2;-2;2;2;2;2;2.2;2.2;str-2;2022-05-09T23:59:59;{}
-null;NULL;null;N/A;;null;null;null;null;null;null;null;null;null
\ No newline at end of file
+## supported types: bool;int8;int16;int32;int64;uint8;uint16;uint32;uint64;float32;float64;string;timestamp;list(i64);binary
+true;-1;-1;-1;-1;1;1;1;1;1.1;1.1;str-1;2022-05-09T00:01:01;{1,2,3};AAEC
+false;-2;-2;-2;-2;2;2;2;2;2.2;2.2;str-2;2022-05-09T23:59:59;{};AwQF
+null;NULL;null;N/A;;null;null;null;null;null;null;null;null;null;null
```
zeroshade, line 21: Can we make sure there's a test case for testing an empty value in addition to a null value?

go/arrow/csv/transformer.go
```diff
@@ -215,6 +216,15 @@ func (w *Writer) transformColToStringArr(typ arrow.DataType, col arrow.Array) []
 				res[i] = w.nullValue
 			}
 		}
+	case arrow.BinaryDataType:
```
zeroshade, line 219: same as above, make sure that `LargeString` is covered above this case so it doesn't fall into here.

go/arrow/csv/transformer.go
```diff
@@ -215,6 +216,15 @@ func (w *Writer) transformColToStringArr(typ arrow.DataType, col arrow.Array) []
...
+		arr := col.(*array.Binary)
```
zeroshade, line 220: you can't assume this, it could also be `*array.LargeBinary` or otherwise. If you don't want to duplicate the code and create cases for both `LargeBinary` and `Binary` separately, you can create and use an interface which contains `Value(int) []byte` which would cover both `Binary` and `LargeBinary` types.

zeroshade (changes requested):

go/arrow/csv/reader.go
yevgenypats: I handled it a bit differently as `""` shouldn't mean null in base64 but rather just empty and `(null)` is not a valid base64 encoding anyway so we don't really need to check the `r.StringsCanBeNull`

go/arrow/csv/reader.go
yevgenypats: I would stay with one option for starters. Where should I document it?

go/arrow/csv/testdata/types.csv
yevgenypats: Done

go/arrow/csv/transformer.go
yevgenypats: Good catch. Should be fine now as handling only `BinaryType` in this PR (similar to how StringType only handled now)

yevgenypats:
@zeroshade updated :)

go/arrow/csv/common.go
```diff
@@ -223,6 +223,7 @@ func validate(schema *arrow.Schema) {
 		case *arrow.Date32Type, *arrow.Date64Type:
 		case *arrow.Decimal128Type, *arrow.Decimal256Type:
 		case *arrow.ListType:
+		case arrow.BinaryDataType:
```
zeroshade, line 226: if you're not going to support the `LargeString` and `LargeBinary` types, then this should be just `*arrow.BinaryType` instead of `arrow.BinaryDataType` which is an interface that includes the Large variants

go/arrow/csv/reader.go
zeroshade: Hmm. I'd suggest adding it to the `NewReader` godoc comment as that seems most reasonable to me. To mention that a binary field expects to be base64 encoded with padding as per `base64.StdEncoding`

zeroshade (approved):
Awesome, thanks!

PR #34559, opened 2023-03-14
GH-34165: [Python] Extension array data type should default to the storage type if to_pandas_dtype is not implemented

### LINKED ISSUE #34165: [Python] Extension array data type should default to the storage type if to_pandas_dtype is not implemented

When working on the extension type for tensors in PyArrow I came across a behaviour of the conversion to pandas that could be improved. 

Creating an extension array (fixed shape tensor in this case) and converting it to pandas works well

```python
>>> arr = [[1, 2, 3, 4], [10, 20, 30, 40], [100, 200, 300, 400]]
>>> storage = pa.array(arr, pa.list_(pa.int32(), 4))
>>> tensor = pa.ExtensionArray.from_storage(tensor_type, storage)
>>> tensor.to_pandas()
0            [1, 2, 3, 4]
1        [10, 20, 30, 40]
2    [100, 200, 300, 400]
dtype: object
```
But creating a table with an extension array and then converting it to pandas fails:

```python
>>> data = [
...     pa.array([1, 2, 3]),
...     pa.array(['foo', 'bar', None]),
...     pa.array([True, None, True]),
...     tensor
... ]
>>> my_schema = pa.schema([('f0', pa.int8()),
...                        ('f1', pa.string()),
...                        ('f2', pa.bool_()),
...                        ('tensors_int', tensor_type)])
>>> table = pa.Table.from_arrays(data, schema=my_schema)
>>> table.to_pandas()
  File "pyarrow/array.pxi", line 830, in pyarrow.lib._PandasConvertible.to_pandas
    return self._to_pandas(options, categories=categories,
  File "pyarrow/table.pxi", line 4004, in pyarrow.lib.Table._to_pandas
    mgr = table_to_blockmanager(
  File "/Users/alenkafrim/repos/arrow/python/pyarrow/pandas_compat.py", line 820, in table_to_blockmanager
    blocks = _table_to_blocks(options, table, categories, ext_columns_dtypes)
  File "/Users/alenkafrim/repos/arrow/python/pyarrow/pandas_compat.py", line 1171, in _table_to_blocks
    return [_reconstruct_block(item, columns, extension_columns)
  File "/Users/alenkafrim/repos/arrow/python/pyarrow/pandas_compat.py", line 1171, in <listcomp>
    return [_reconstruct_block(item, columns, extension_columns)
  File "/Users/alenkafrim/repos/arrow/python/pyarrow/pandas_compat.py", line 776, in _reconstruct_block
    pandas_dtype = extension_columns[name]
KeyError: 'tensors_int'
```

The issue is due to the extension array in this example not having `to_pandas_dtype` method implemented. In this case `ext_columns` does not get populated in `_get_extension_dtypes` method with the name of the column with an extension type:

https://github.com/apache/arrow/blob/0368e410be4dac30eada13d307b415165aedc6a7/python/pyarrow/pandas_compat.py#L870-L879

It would be good if it would, in case `to_pandas_dtype` method is not implemented, convert the storage array https://github.com/apache/arrow/blob/0368e410be4dac30eada13d307b415165aedc6a7/python/pyarrow/pandas_compat.py#L776 similar to
https://github.com/apache/arrow/blob/925cbd81427ae02ce897c406a264d53c8813b920/python/pyarrow/array.pxi#L2888-L2889

### Component(s)

Python

AlenkaF:
### Rationale for this change
Method `to_pandas` fails with `KeyError` if a table has an extension array as a column with extension dtype not having `to_pandas_dtype` defined. In this cases we should fall back to storage type of the extension array.

Changes in `arrow_to_pandas.cc` at:
- `GetBlockType` for `ConvertTableToPandas`
- `ConvertChunkedArrayToPandas`
* Closes: #34165

### DIFF
--- a/python/pyarrow/array.pxi
+++ b/python/pyarrow/array.pxi
@@ -1459,6 +1459,9 @@ cdef class Array(_PandasConvertible):
         supported for primitive arrays with the same memory layout as NumPy
         (i.e. integers, floating point, ..) and without any nulls.
 
+        For the extension arrays, this method simply delegates to the
+        underlying storage array.
+
         Parameters
         ----------
         zero_copy_only : bool, default True
--- a/python/pyarrow/array.pxi
+++ b/python/pyarrow/array.pxi
@@ -3072,23 +3072,6 @@ cdef class ExtensionArray(Array):
         # otherwise convert the storage array with the base implementation
         return Array._to_pandas(self.storage, options, **kwargs)
 
-    def to_numpy(self, **kwargs):
-        """
-        Convert extension array to a numpy ndarray.
-
-        This method simply delegates to the underlying storage array.
-
-        Parameters
-        ----------
-        **kwargs : dict, optional
-            See `Array.to_numpy` for parameter description.
-
-        See Also
-        --------
-        Array.to_numpy
-        """
-        return self.storage.to_numpy(**kwargs)
-
 
 cdef dict _array_classes = {
     _Type_NA: NullArray,
--- a/python/pyarrow/tests/test_extension_type.py
+++ b/python/pyarrow/tests/test_extension_type.py
@@ -1127,3 +1127,24 @@ def test_cpp_extension_in_python(tmpdir):
     reconstructed_array = batch.column(0)
     assert reconstructed_array.type == uuid_type
     assert reconstructed_array == array
+
+
+def test_extension_to_pandas_storage_type(registered_period_type):
+    period_type, _ = registered_period_type
+    storage = pa.array([1, 2, 3, 4], pa.int64())
+    arr = pa.ExtensionArray.from_storage(period_type, storage)
+    breakpoint()
+    arr.to_pandas()
+
+    data = [
+        pa.array([1, 2, 3, 4]),
+        pa.array(['foo', 'bar', None, None]),
+        pa.array([True, None, True, False]),
+        arr
+    ]
+    my_schema = pa.schema([('f0', pa.int8()),
+                           ('f1', pa.string()),
+                           ('f2', pa.bool_()),
+                           ('ext', period_type)])
+    table = pa.Table.from_arrays(data, schema=my_schema)
+    table.to_pandas()
--- a/python/pyarrow/tests/test_extension_type.py
+++ b/python/pyarrow/tests/test_extension_type.py
@@ -740,7 +740,21 @@ class PeriodTypeWithClass(PeriodType):
         return PeriodTypeWithClass(freq)
 
 
-@pytest.fixture(params=[PeriodType('D'), PeriodTypeWithClass('D')])
+class PeriodTypeWithToPandasDtype(PeriodType):
+    @classmethod
+    def __arrow_ext_deserialize__(cls, storage_type, serialized):
+        freq = PeriodType.__arrow_ext_deserialize__(
+            storage_type, serialized).freq
+        return PeriodTypeWithToPandasDtype(freq)
+
+    def to_pandas_dtype(self):
+        import pandas as pd
+        return pd.PeriodDtype(freq=self.freq)
+
+
+@pytest.fixture(params=[PeriodType('D'),
+                        PeriodTypeWithClass('D'),
+                        PeriodTypeWithToPandasDtype('D')])
 def registered_period_type(request):
     # setup
     period_type = request.param
@@ -1131,25 +1145,41 @@ def test_cpp_extension_in_python(tmpdir):
 
 def test_extension_to_pandas_storage_type(registered_period_type):
     period_type, _ = registered_period_type
+    np_arr = np.array([1, 2, 3, 4])
     storage = pa.array([1, 2, 3, 4], pa.int64())
     arr = pa.ExtensionArray.from_storage(period_type, storage)
-    arr.to_pandas()
+
+    if isinstance(period_type, PeriodTypeWithToPandasDtype):
+        pandas_dtype = period_type.to_pandas_dtype()
+    else:
+        pandas_dtype = np_arr.dtype
+
+    # Test arrays
+    result = arr.to_pandas()
+    assert result.dtype == pandas_dtype
 
     # Test the change in ConvertChunkedArrayToPandas
     chunked_arr = pa.chunked_array([arr])
-    chunked_arr.to_numpy()
+    result = chunked_arr.to_numpy()
+    assert result.dtype == np_arr.dtype
+
+    result = chunked_arr.to_pandas()
+    # TODO: to_pandas should take use of to_pandas_dtype
+    # if defined!
+    # assert result.dtype == pandas_dtype
+    assert result.dtype == np_arr.dtype
 
     # Test the change in ConvertTableToPandas
-    # data = [
-    #     pa.array([1, 2, 3, 4]),
-    #     pa.array(['foo', 'bar', None, None]),
-    #     pa.array([True, None, True, False]),
-    #     arr
-    # ]
-    # my_schema = pa.schema([('f0', pa.int8()),
-    #                        ('f1', pa.string()),
-    #                        ('f2', pa.bool_()),
-    #                        ('ext', period_type)])
-    # table = pa.Table.from_arrays(data, schema=my_schema)
-
-    # table.to_pandas()
+    data = [
+        pa.array([1, 2, 3, 4]),
+        pa.array(['foo', 'bar', None, None]),
+        pa.array([True, None, True, False]),
+        arr
+    ]
+    my_schema = pa.schema([('f0', pa.int8()),
+                           ('f1', pa.string()),
+                           ('f2', pa.bool_()),
+                           ('ext', period_type)])
+    table = pa.Table.from_arrays(data, schema=my_schema)
+    result = table.to_pandas()
+    assert result["ext"].dtype == pandas_dtype
--- a/python/pyarrow/tests/test_extension_type.py
+++ b/python/pyarrow/tests/test_extension_type.py
@@ -25,6 +25,7 @@ import sys
 
 import numpy as np
 import pyarrow as pa
+from pyarrow.vendored.version import Version
 
 import pytest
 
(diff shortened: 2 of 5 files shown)


python/pyarrow/array.pxi
```diff
@@ -2888,23 +2888,6 @@ cdef class ExtensionArray(Array):
         # otherwise convert the storage array with the base implementation
         return Array._to_pandas(self.storage, options, **kwargs)
 
-    def to_numpy(self, **kwargs):
-        """
-        Convert extension array to a numpy ndarray.
-
-        This method simply delegates to the underlying storage array.
```
jorisvandenbossche, line 2895: Can you move this sentence to the docstring of the to_numpy method on the base class ("For this extension arrays, this method ...")

python/pyarrow/tests/test_extension_type.py
```diff
@@ -1127,3 +1141,45 @@ def test_cpp_extension_in_python(tmpdir):
...
+        pa.array([True, None, True, False]),
+        arr
+    ]
+    my_schema = pa.schema([('f0', pa.int8()),
+                           ('f1', pa.string()),
+                           ('f2', pa.bool_()),
+                           ('ext', period_type)])
+    table = pa.Table.from_arrays(data, schema=my_schema)
+    result = table.to_pandas()
+    assert result["ext"].dtype == pandas_dtype
```
jorisvandenbossche, line 1185: Maybe we can test that specifying `types_mapper` still overrides it?

python/pyarrow/tests/test_extension_type.py
```diff
@@ -1127,3 +1141,45 @@ def test_cpp_extension_in_python(tmpdir):
...
+    # Test the change in ConvertChunkedArrayToPandas
+    chunked_arr = pa.chunked_array([arr])
+    result = chunked_arr.to_numpy()
+    assert result.dtype == np_arr.dtype
+
+    result = chunked_arr.to_pandas()
+    # TODO: to_pandas should take use of to_pandas_dtype
+    # if defined!
+    # assert result.dtype == pandas_dtype
+    assert result.dtype == np_arr.dtype
```
jorisvandenbossche, line 1170: Yeah, for `Array.to_pandas`, we have some custom logic for this in the cython code. We might want to do something similar for ChunkedArray?

jorisvandenbossche (commented):
Looking good!

python/pyarrow/tests/test_extension_type.py
AlenkaF: This doesn't work currently (without the change in this PR also). The problem is that the extension type is unhashable so we can not even define a `types_mapper` with an extension type:

```python
>>> period_type = PeriodType('D')
>>> {period_type: pd.Int64Dtype()}
TypeError: unhashable type: 'PeriodType'
```

The `DataType` base class has the `__hash__` method implemented so I guess that should work for `ExtensionType` subclass also? If we would want `types_mapper` keyword to work for extension types, I guess, we would have to add `__hash__` method in the definition of the type:

```python
class PeriodType(pa.ExtensionType):
    ...
    def __hash__(self):
        return hash(str(self)) 
```

But using `PeriodType` with the `__hash__` method and the `types_mapper` (latest main, not this PR) I get an error on pandas side:

```python
>>> types_mapper = {period_type: pd.Int64Dtype()}.get
>>> table.to_pandas(types_mapper=types_mapper)
  File "pyarrow/array.pxi", line 835, in pyarrow.lib._PandasConvertible.to_pandas
    return self._to_pandas(options, categories=categories,
  File "pyarrow/table.pxi", line 4111, in pyarrow.lib.Table._to_pandas
    mgr = table_to_blockmanager(
  File "/Users/alenkafrim/repos/arrow/python/pyarrow/pandas_compat.py", line 820, in table_to_blockmanager
    blocks = _table_to_blocks(options, table, categories, ext_columns_dtypes)
  File "/Users/alenkafrim/repos/arrow/python/pyarrow/pandas_compat.py", line 1171, in _table_to_blocks
    return [_reconstruct_block(item, columns, extension_columns)
  File "/Users/alenkafrim/repos/arrow/python/pyarrow/pandas_compat.py", line 1171, in <listcomp>
    return [_reconstruct_block(item, columns, extension_columns)
  File "/Users/alenkafrim/repos/arrow/python/pyarrow/pandas_compat.py", line 780, in _reconstruct_block
    pd_ext_arr = pandas_dtype.__from_arrow__(arr)
  File "/Users/alenkafrim/repos/pyarrow-dev/lib/python3.10/site-packages/pandas/core/arrays/numeric.py", line 88, in __from_arrow__
    raise TypeError(
TypeError: Expected array of Int64 type, got extension<test.period<PeriodType>> instead
```

This looks it deserves a separate issue.

jorisvandenbossche: We should indeed ensure that extension types are hashable, but yes that's worth a separate issue. 

Now, for `types_mapper`, you don't have to use a dict (it only provides a convenient way to specify it). The keyword takes a generic callable.

> But using `PeriodType` with the `__hash__` method and the `types_mapper` (latest main, not this PR) I get an error on pandas side:

Ah, yes, that's because the Int64Dtype from pandas doesn't expect an extension type. You can probably use `pd.ArrowDtype` instead, that should probably work (I think that's the only pandas extension dtype that should work with any pyarrow extension type).

AlenkaF: True for the callable. Will add that to the test (a callable that is not a dict and `pd.ArrowDtype`).

It feels a bit out of place that the `types_mapper` argument doesn't really work for extension columns, as the only case that it does is if the function is not a dict and it transforms to a pandas `pd.ArrowDtype`. Maybe adding the info to the docs might be good.

python/pyarrow/tests/test_extension_type.py
```diff
@@ -1127,3 +1142,51 @@ def test_cpp_extension_in_python(tmpdir):
...
+    if isinstance(period_type, PeriodTypeWithToPandasDtype):
+        pandas_dtype = period_type.to_pandas_dtype()
+    else:
+        pandas_dtype = np_arr.dtype
+
+    # Test arrays
+    result = arr.to_pandas()
+    assert result.dtype == pandas_dtype
+
+    # Test the change in ConvertChunkedArrayToPandas
```
jorisvandenbossche, line 1163:
```suggestion
    # Test chunked arrays
```

(later we won't know what this "change" is about)

python/pyarrow/tests/test_extension_type.py
```diff
@@ -1127,3 +1142,51 @@ def test_cpp_extension_in_python(tmpdir):
...
+    chunked_arr = pa.chunked_array([arr])
+    result = chunked_arr.to_numpy()
+    assert result.dtype == np_arr.dtype
+
+    result = chunked_arr.to_pandas()
+    assert result.dtype == pandas_dtype
+
+    # Test the change in ConvertTableToPandas
```
jorisvandenbossche, line 1171:
```suggestion
    # Test Table.to_pandas
```

python/pyarrow/tests/test_extension_type.py
```diff
@@ -1127,3 +1142,51 @@ def test_cpp_extension_in_python(tmpdir):
...
+    result = table.to_pandas()
+    assert result["ext"].dtype == pandas_dtype
+
+    import pandas as pd
+    if Version(pd.__version__) < Version("1.5.0"):
+        pytest.skip("ArrowDtype missing")
+
+        # Check the usage of types_mapper
+        result = table.to_pandas(types_mapper=pd.ArrowDtype)
+        assert isinstance(result["ext"].dtype, pd.ArrowDtype)
```
jorisvandenbossche, line 1192:
```suggestion
    # Check the usage of types_mapper
    result = table.to_pandas(types_mapper=pd.ArrowDtype)
    assert isinstance(result["ext"].dtype, pd.ArrowDtype)
```

This should be de-dented, so that this actually runs for latest pandas versions?

(once you do that, you will also need to trigger crossbow to run the test-conda-python-3.8-pandas-nightly build)

python/pyarrow/tests/test_extension_type.py
```diff
@@ -1127,3 +1142,51 @@ def test_cpp_extension_in_python(tmpdir):
...
+                           ('f1', pa.string()),
+                           ('f2', pa.bool_()),
+                           ('ext', period_type)])
+    table = pa.Table.from_arrays(data, schema=my_schema)
+    result = table.to_pandas()
+    assert result["ext"].dtype == pandas_dtype
+
+    import pandas as pd
+    if Version(pd.__version__) < Version("1.5.0"):
+        pytest.skip("ArrowDtype missing")
```
jorisvandenbossche, line 1188:
```suggestion
    if Version(pd.__version__) >= Version("2.0.0.dev0"):
```

Alternative fix, it might be better to not mark the full test as "skipped", but just only run this part for newer pandas

jorisvandenbossche (commented):
Looks good! Just for the extra types_mapper test, we need to ensure the test is actually run, and check it with CI

jorisvandenbossche:
Hmm, the pandas nightly build is failing to build (not even running the tests), some LLVM related error for gandiva. I don't (yet) see this on the latest crossbow nightly build, though

jorisvandenbossche:
Ah, llvm updated from 15 to 16, and our cmake requires 15. We should probably pin that temporarily in the env files.

AlenkaF:
Thanks for looking at this! I wasn't sure what was the issue for the same reason you mentioned (latest crossbow nightly builds were fine).

jorisvandenbossche:
To make progress on this PR in the meantime, you could add a `llvmdev<16` to `conda_env_gandiva.txt`, I _think_ that should solve it.

AlenkaF:
@github-actions crossbow submit test-conda-python-3.8-pandas-nightly

AlenkaF:
@jorisvandenbossche comparing the test failures in the pandas nightly build for this PR and the latest crossbow nightly build (https://github.com/ursacomputing/crossbow/actions/runs/4559282266) it seems there is no extra test failing.

jorisvandenbossche:
Maybe update once more, then the nightly pandas builds should actually pass

AlenkaF:
@github-actions crossbow submit test-conda-python-3.8-pandas-nightly

AlenkaF:
Nightly is now passing, but I have to look at the failures in the CI.

AlenkaF:
@github-actions crossbow submit test-conda-python-3.8-pandas-nightly

jorisvandenbossche (approved):
Thanks!

PR #34570, opened 2023-03-15
GH-34568: [C++][Python] Expose Run-End Encoded arrays in Python Arrow

### LINKED ISSUE #34568: [Python] Expose Run-End Encoded arrays in PyArrow

The python bindings are setup to warn us if there is a compute function that does not have python bindings.  The fix should be pretty straightforward.  Look at https://github.com/apache/arrow/pull/34084/files for an example.

```
/home/pace/dev/arrow/python/pyarrow/compute.py:206: RuntimeWarning: Python binding for RunEndEncodeOptions not exposed
  warnings.warn("Python binding for {} not exposed"
```

### Component(s)

Python

felipecrv:
* Closes: #34568


felipecrv:
@westonpace @zeroshade

python/pyarrow/tests/test_compute.py
```diff
@@ -3108,3 +3109,11 @@ def test_list_slice_bad_parameters():
         pc.list_slice(arr, 0, 1, step=0)
     with pytest.raises(pa.ArrowInvalid, match=msg + "-1"):
         pc.list_slice(arr, 0, 1, step=-1)
+
+
+def test_run_end_encode():
+    arr = pa.array([1, 1, 1, 2, 2, 1, 1, 1, 1, 1, 3, 3, 3, 3, 3, 3, 3, 3, 3])
+    encoded = pc.run_end_encode(arr, options=pc.RunEndEncodeOptions(pa.int64()))
```
jorisvandenbossche, line 3116: So this actually fails at the moment, because we can't yet put a REE C++ array in a pyarrow.Array object (`pyarrow_wrap_array` fails here)

On the short term, we can maybe put it just in the base class Array (I don't know if there is already an issue about adding proper python bindings?)

felipecrv: Ooops. And I've ran the tests locally, but only now saw the lone F at the end of the green dots :D

I will fix this properly.

python/pyarrow/_compute.pyx
```diff
@@ -1335,6 +1335,27 @@ class DictionaryEncodeOptions(_DictionaryEncodeOptions):
         self._set_options(null_encoding)
 
 
+cdef class _RunEndEncodeOptions(FunctionOptions):
+    def _set_options(self, run_end_type):
+        self.wrapped.reset(new CRunEndEncodeOptions(pyarrow_unwrap_data_type(run_end_type)))
```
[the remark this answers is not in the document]
felipecrv: Committed locally. Pushing soon.

westonpace (approved):
The changes seem fine to me once @jorisvandenbossche 's suggestions are addressed.

felipecrv:
@jorisvandenbossche I added everything related to REEs to Python Arrow (hopefully). Please review and give your suggestions.

python/pyarrow/array.pxi
```diff
@@ -2851,6 +2851,103 @@ cdef class StructArray(Array):
...
+        _run_ends = asarray(run_ends)
+        _values = asarray(values)
+        _logical_offset = <int64_t>0
+
+        type = ensure_type(type, allow_none=True)
+        if type is not None:
+            c_type = pyarrow_unwrap_data_type(type)
+            ree_array.reset(new CRunEndEncodedArray(
+                    c_type, _logical_length, _run_ends.sp_array,
+                    _values.sp_array, _logical_offset))
```
jorisvandenbossche, line 2897: I suppose for this code path you need to ensure that run_ends and values have the correct matching type? (potentially the source of the segfault)

python/pyarrow/tests/test_types.py
```diff
@@ -818,6 +818,19 @@ def test_fields_weakrefable():
...
+    ty = pa.run_end_encoded(pa.int64(), pa.utf8())
+    assert isinstance(ty, pa.RunEndEncodedType)
+    assert ty.run_end_type == pa.int64()
+    assert ty.value_type == pa.utf8()
+
+    with pytest.raises(TypeError):
+        pa.run_end_encoded(pa.int64(), None)
+
+    with pytest.raises(TypeError):
+        pa.run_end_encoded(None, pa.utf8())
```
jorisvandenbossche, line 831: Can you also test with an unsupported run_end type (eg uint)

python/pyarrow/tests/test_types.py
jorisvandenbossche, line 1: There is a `get_many_types` at the top of this file that returns many types used in various generic tests, can you add one run end encoded type there as well?

python/pyarrow/tests/test_convert_builtin.py
```diff
@@ -2351,3 +2351,22 @@ def test_array_from_pylist_offset_overflow():
     assert isinstance(arr, pa.ChunkedArray)
     assert len(arr) == 2**31
     assert len(arr.chunks) > 1
+
+
+def check_run_end_encoded_with_type(ree_type=None):
```
jorisvandenbossche, line 2356: Can you move this to `test_array.py` ? (I _think_ that's a better fit for testing the attributes of the array class and the from_arrays constructor)

jorisvandenbossche: And could you also add a test for `RunEndEncodedArray.from_buffers`? (this gets inherited from the base class, but we should ensure this version actually works and not segfaults or so)

jorisvandenbossche (commented):
Looking great! Didn't have time today to finish a thorough review, just posting some quick comments already.

python/pyarrow/array.pxi
```diff
@@ -2851,6 +2851,103 @@ cdef class StructArray(Array):
...
+
+    @staticmethod
+    def from_arrays(logical_length, run_ends, values, type=None):
+        """
+        Construct RunEndEncodedArray from run_ends and values arrays.
+
+        Parameters
+        ----------
+        logical_length : int
+            The logical length of the run-end encoded array.
```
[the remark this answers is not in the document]
felipecrv: The explicit length is used to validate the `run_ends` array. I can't simply take `run_ends[-1]` without also checking the physical length of `values`.

python/pyarrow/tests/test_convert_builtin.py
jorisvandenbossche: There is a children keyword, so I assume it would take an empty list for the buffers, and a list of children

python/pyarrow/array.pxi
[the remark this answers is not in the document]
jorisvandenbossche: > I can't simply take `run_ends[-1]` without also checking the physical length of `values`.

That's a check that happens anyway? 

```
In [49]: pc.run_end_decode(pa.RunEndEncodedArray.from_arrays(8, [2, 5, 7, 8], [1, 2, 3]))
...
ArrowInvalid: Values array has to be at least as long as run ends array

```

jorisvandenbossche:
Something else: you can create an invalid REE array with non-increasing run ends:

```
In [2]: a= pa.RunEndEncodedArray.from_arrays(5, [2, 4, 2, 5], [1, 2, 3, 4])

In [3]: a
Out[3]: 
<pyarrow.lib.RunEndEncodedArray object at 0x7f8138ffa920>

-- run_ends:
  [
    2,
    4,
    2,
    5
  ]
-- values:
  [
    1,
    2,
    3,
    4
  ]

In [4]: pc.run_end_decode(a)
Segmentation fault (core dumped)
```

And as you can see, decoding it then segfaults. 
The full validation actually catches this:

```
In [4]: a.validate(full=True)
...
ArrowInvalid: Every run end must be strictly greater than the previous run end, but run_ends[2] is 2 and run_ends[1] is 4
```

But the constructor only does the cheap validation (without `full=True`). But I suppose it is always a bit the question and trade-off what is considered a necessary / cheap check and what is only part of the full validation (the same is true for offsets in variable size list/binary)

python/pyarrow/array.pxi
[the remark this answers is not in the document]
felipecrv: Cool. I will remove the `logical_length` parameter and call full validation after the REE is constructed.

python/pyarrow/array.pxi
felipecrv: Rewrote the handling of type completely in this function. Pushing soon.

python/pyarrow/array.pxi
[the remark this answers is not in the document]
felipecrv: [x] Removed length parameter
[x] add `full=True` to `validate` call

python/pyarrow/tests/test_types.py
felipecrv: Added. It raises a `ValueError` as the validation comes from the C++ side. Is that OK or should I write a Python validation that raise `TypeError`?

python/pyarrow/tests/test_types.py
felipecrv: Added 3. One for each valid run-end type.

python/pyarrow/tests/test_convert_builtin.py
felipecrv: I wonder if I should remove the `=None` from `children=None` in my override version of `from_buffers`. What do you say?

python/pyarrow/tests/test_types.py
jorisvandenbossche: ValueError is fine here I think (it's a bit ambiguous anyway, as you can argue that the "type" of the argument, i.e. a DataType is correct, but the exact value (int8 vs int16) is not correct. It's just that the values itself are also types ;))

python/pyarrow/array.pxi
```diff
@@ -2851,6 +2851,169 @@ cdef class StructArray(Array):
...
+        An array holding the values of each run.
+
+        The physical offset to the array is applied.
+        """
+        cdef CRunEndEncodedArray* ree_array = <CRunEndEncodedArray*>(self.ap)
+        return pyarrow_wrap_array(ree_array.values())
+
+    def find_physical_offset(self):
+        """
+        Find the physical offset of this REE array.
```
jorisvandenbossche, line 2994: Maybe you can clarify that this is the offset into the run_ends and values child arrays?

python/pyarrow/array.pxi
```diff
@@ -2851,6 +2851,169 @@ cdef class StructArray(Array):
...
+        """
+        Find the physical length of this REE array.
+
+        The physical length of an REE is the number of physical values (and
+        run-ends) necessary to represent the logical range of values from offset
+        to length.
+
+        Avoid calling this function if the physical length can be estabilished in
+        some other way (e.g. when iterating over the runs sequentially until the
+        end). This function uses binary-search, so it has a O(log N) cost.
```
jorisvandenbossche, line 3011: This paragraph is maybe too detailed for the python docs? (for example, in python I don't think you should typically iterate over the runs)

jorisvandenbossche (commented):
Thanks for the updates! Some small comments.

Can you also list the new classes and functions (that you added to the top-level namespace) in the python docs? (eg in docs/source/python/api/arrays.rst)

BTW, something that is still missing from this PR is the equivalent Scalar subclass (for me it's also fine to add this in a follow-up PR if you would prefer finalizing this one):

```
In [15]: run_ends = [3, 5, 10, 19]
    ...: values = [1, 2, 1, 3]
    ...: ree_array = pa.RunEndEncodedArray.from_arrays(run_ends, values)

In [16]: ree_array[0]
...
NotImplementedError: Wrapping scalar of type run_end_encoded<run_ends: int64, values: int64>
```

And that reminds me, we should also open follow-up issues to implement the pyarrow<->python/numpy conversions. For example `ree_array.to_numpy()` or `pa.array([1, 1, 2], ree_type)`.

python/pyarrow/_compute.pyx
```diff
@@ -1335,6 +1335,31 @@ class DictionaryEncodeOptions(_DictionaryEncodeOptions):
         self._set_options(null_encoding)
 
 
+cdef class _RunEndEncodeOptions(FunctionOptions):
+    def _set_options(self, run_end_type=None):
+        run_end_ty = ensure_type(run_end_type, allow_none=True)
+        if run_end_ty is None:
+            self.wrapped.reset(new CRunEndEncodeOptions())
```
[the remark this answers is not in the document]
felipecrv: Removing. Nice cleanup.

python/pyarrow/array.pxi
felipecrv: That's also a confusing explanation of this concept. :-)

Going with this:

```diff
         """
         Find the physical offset of this REE array.

+        This is the offset of the run that contains the value of the first
+        logical element of this array considering its offet.
+
         This function uses binary-search, so it has a O(log N) cost.
         """
```

python/pyarrow/array.pxi
felipecrv: Removing. Leaving only "This function uses binary-search, so it has a O(log N) cost.".

python/pyarrow/array.pxi
jorisvandenbossche: Sounds good! Adding an example might also help to illustrate the explanation (feel free to do that as a follow-up!)

jorisvandenbossche (approved):

jorisvandenbossche:
Let's already get this merged, we can do some more docs and the python<->arrow conversions as follow-ups. 
Thanks!

jorisvandenbossche:
Opened https://github.com/apache/arrow/issues/34701 to keep track of the general topic

PR #34586, opened 2023-03-16
GH-34216: [Python] Support for reading JSON Datasets With Python

### LINKED ISSUE #34216: [Python] Add bindings for JSON format in Dataset

https://github.com/apache/arrow/pull/33732 adds C++ support for reading (but not writing) JSON datasets. Once that's finalized, we should expose the new format via Python.

### Component(s)

Python

R-JunmingChen:
This PR supports for reading JSON Datasets With Python. As mentioned in [#34216](https://github.com/apache/arrow/issues/34216), only the reading ability are supported.

Please compare the difference between my implemenation of _json.pyx, _json.pyd and _csv.pyx _csv.pyd.
Cause _csv.pyd utilize pointer for cpp class and my implementation doesn't. 

**What changes are included in this PR?**

C++: add inclusion for file_json.h
Python: reference C++ codes and support reading JSON Datasets

**Are these changes tested?**
Yes
6 test samples added in tests/test_dataset.py

* Closes: #34216


python/pyarrow/_dataset.pyx
```diff
@@ -1171,6 +1174,7 @@ cdef class FileFormat(_Weakrefable):
         classes = {
             'ipc': IpcFileFormat,
             'csv': CsvFileFormat,
+            'json':JsonFileFormat,
```
westonpace, line 1177:
```suggestion
            'json': JsonFileFormat,
```

python/pyarrow/_dataset.pyx
```diff
@@ -1307,10 +1311,11 @@ cdef class Fragment(_Weakrefable):
         type_name = frombytes(sp.get().type_name())
 
         classes = {
-            # IpcFileFormat, CsvFileFormat and OrcFileFormat do not have
+            # IpcFileFormat, CsvFileFormat, JsonFileFormat and OrcFileFormat do not have
             # corresponding subclasses of FileFragment
             'ipc': FileFragment,
             'csv': FileFragment,
+            'json': FileFormat,
```
westonpace, line 1318:
```suggestion
            'json': FileFragment,
```

python/pyarrow/_dataset.pyx
```diff
@@ -2176,6 +2182,126 @@ cdef class CsvFileWriteOptions(FileWriteOptions):
...
+        return f"<JsonFileFormat>"
+
+
+cdef class JsonFragmentScanOptions(FragmentScanOptions):
+    """
+    Scan-specific options for JSON fragments.
+
+    Parameters
+    ----------
+    read_options : pyarrow.json.ReadOptions
```
westonpace, line 2254: Add parse_options?

python/pyarrow/_json.pxd
```diff
@@ -0,0 +1,37 @@
...
+# cython: language_level = 3
+
+from pyarrow.includes.libarrow cimport *
+from pyarrow.lib cimport _Weakrefable
+
+
+cdef class ParseOptions(_Weakrefable):
+    cdef:
+        CJSONParseOptions options
+
```
westonpace, line 27: Can we structure these more like we have the CSV options?  It's not clear to users what they can set here and doing something like...

```
options = json.ParseOptions()
options.options.xyz = ...
```

...seems a bit redundant.

westonpace (changes requested):
Thanks for working on this!  I have a few questions / suggestions.  Also, we will need to add some unit tests for this feature.

python/pyarrow/_json.pxd
R-JunmingChen: Hi, the CSV options also use this structure. 
https://github.com/apache/arrow/blob/caa78981259fc029182e677f13183371822ce410/python/pyarrow/_csv.pxd#L32-L38
For Arrow developers who use cython, it indeed seems redundant, but for users, the C part is transparent, users only need to functions of Python part to utilize ParseOptions.

westonpace: Ah, sorry, I was referring to something like this: https://github.com/apache/arrow/blob/main/python/pyarrow/_csv.pyx#L89-L205

I don't see an equivalent for JSON (I see there is a `pyx` file with `ParseOptions` but it does not have a docstring or an `__init__` method that allows users to set the various options).

R-JunmingChen: I believe there is a  __init__ method of ParseOptions in _json.pyx 

https://github.com/apache/arrow/blob/main/python/pyarrow/_json.pyx#L115-L123

R-JunmingChen:
> Thanks for working on this! I have a few questions / suggestions. Also, we will need to add some unit tests for this feature.

Hi, I have added some tests, which references the csv tests. Please review them when you are free

python/pyarrow/tests/test_dataset.py
```diff
@@ -3172,6 +3202,58 @@ def test_csv_fragment_options(tempdir, dataset_reader):
...
+    table = pa.table({'a': pa.array([1, 2, 3], type="int64"),
+                      'b': pa.array([.1, .2, .3], type="float64")})
+
+    path = str(tempdir / 'test.json')
+    out = table.to_pandas().to_json(orient='records')[1:-1].replace('},{', '}\n{')
+    with open(path, 'w') as f:
+        f.write(out)
+    
+    dataset = ds.dataset(path, format=ds.JsonFileFormat(
+        read_options=pa.json.ReadOptions(block_size=64)))
```
westonpace, line 3235: Is there any way to know if this option actually got applied?  If the bindings were completely ignoring `block_size` would this test fail?

Is there another option that is easier to verify?  Maybe `newlines_in_values` (and put some newlines in the test data) because then it will fail if this option is not getting set correctly?

westonpace (changes requested):
These tests look good, thank you for adding them.  One small suggestion I think to improve them.

python/pyarrow/tests/test_dataset.py
R-JunmingChen: Thanks for useful suggestions.
The blocksize is indeed `inappropriate ` here. I will change to `newlines_in_values` soon.

R-JunmingChen: > 

Hi, It seems that `newlines_in_values `  doesn't work.
`import pyarrow.json as pj
json=pj.read_json("test.json",parse_options=pj.ParseOptions(newlines_in_values=False))`

R-JunmingChen: > Is there any way to know if this option actually got applied? If the bindings were completely ignoring `block_size` would this test fail?
> 
> Is there another option that is easier to verify? Maybe `newlines_in_values` (and put some newlines in the test data) because then it will fail if this option is not getting set correctly?

Since the `newlines_in_values` doesn't work for pyarrow.json interface too, I update my `blocksize` tests, referencing the code in test_json.py. In this way, we can check whether the `blocksize` is set.
https://github.com/apache/arrow/blob/5e8db3156c733a31e196683011db113e76ce6a32/python/pyarrow/tests/test_json.py

We may submit a issue for `newlines_in_values` if it indeed fails to work. BTW, I check the code in test_json, there is no test for `newlines_in_values` too.

westonpace (approved):
I've pushed some lint changes.  Assuming CI passes I think this is good to go.

python/pyarrow/tests/test_dataset.py
R-JunmingChen: > Ok, the updated test is good. Please do open an issue for `newlines_in_values` but we don't need to hold up this PR for that.

Sure, I open a issue #35096 for that.

PR #34591, opened 2023-03-16
GH-34590: [C++][ORC] Fix timestamp type mapping between orc and arrow

### LINKED ISSUE #34590: [C++][ORC] Fix timestamp type mapping between orc and arrow

Background: There was an effort to fix inconsistent timestamp types across different SQL-on-Hadoop engines: https://docs.google.com/document/d/1gNRww9mZJcHvUDCXklzjFEQGpefsuR_akCDfWsdE35Q 

In the Apache Orc, two timestamp types are provided:
- TIMESTAMP: timestamp type without timezone, timestamp value is stored in the writer timezone .
- TIMESTAMP_INSTANT: timestamp type with local timezone, timestamp value is stored in the UTC timezone.

arrow::TimestampType has an optional timezone field: https://github.com/apache/arrow/blob/main/cpp/src/arrow/type.h#L1385
- If timezone is provided, values are normalized in UTC.
- If timezone is missing, values can be in any timezone.

Therefore, the type mapping should be as below:
- orc::TIMESTAMP <=> arrow::TimestampType w/o timezone
- orc::TIMESTAMP_INSTANT <=> arrow::TimestampType w/ timezone

### Component(s)

C++

wgtmac:
### Rationale for this change

Background: There was an effort to fix inconsistent timestamp types across different SQL-on-Hadoop engines: https://docs.google.com/document/d/1gNRww9mZJcHvUDCXklzjFEQGpefsuR_akCDfWsdE35Q

In the Apache Orc, two timestamp types are provided:

- TIMESTAMP: timestamp type without timezone, timestamp value is stored in the writer timezone .
- TIMESTAMP_INSTANT: timestamp type with local timezone, timestamp value is stored in the UTC timezone.

arrow::TimestampType has an optional timezone field:
- If timezone is provided, values are normalized in UTC.
- If timezone is missing, values can be in any timezone.

The type mapping is fixed as below:
- orc::TIMESTAMP <=> arrow::TimestampType w/o timezone
- orc::TIMESTAMP_INSTANT <=> arrow::TimestampType w/ timezone

### Are these changes tested?

Make sure all tests pass.

No.
* Closes: #34590

### DIFF
--- a/cpp/src/arrow/adapters/orc/util.cc
+++ b/cpp/src/arrow/adapters/orc/util.cc
@@ -381,6 +381,7 @@ Status AppendBatch(const liborc::Type* type, liborc::ColumnVectorBatch* batch,
       return AppendNumericBatchCast<Date32Builder, int32_t, liborc::LongVectorBatch,
                                     int64_t>(batch, offset, length, builder);
     case liborc::TIMESTAMP:
+    case liborc::TIMESTAMP_INSTANT:
       return AppendTimestampBatch(batch, offset, length, builder);
     case liborc::DECIMAL:
       return AppendDecimalBatch(type, batch, offset, length, builder);
@@ -995,8 +996,16 @@ Result<std::unique_ptr<liborc::Type>> GetOrcType(const DataType& type) {
     case Type::type::DATE32:
       return liborc::createPrimitiveType(liborc::TypeKind::DATE);
     case Type::type::DATE64:
-    case Type::type::TIMESTAMP:
       return liborc::createPrimitiveType(liborc::TypeKind::TIMESTAMP);
+    case Type::type::TIMESTAMP: {
+      const auto& timestamp_type = checked_cast<const TimestampType&>(type);
+      if (!timestamp_type.timezone().empty()) {
+        // The timestamp values stored in the arrow array are normalized to UTC.
+        return liborc::createPrimitiveType(liborc::TypeKind::TIMESTAMP_INSTANT);
+      }
+      // The timestamp values stored in the arrow array can be in any timezone.
+      return liborc::createPrimitiveType(liborc::TypeKind::TIMESTAMP);
+    }
     case Type::type::DECIMAL128: {
       const uint64_t precision =
           static_cast<uint64_t>(checked_cast<const Decimal128Type&>(type).precision());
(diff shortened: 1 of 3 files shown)


wgtmac:
@westonpace @wjones127 Mind take a look? Thanks!

cpp/src/arrow/adapters/orc/util.cc
```diff
@@ -1111,7 +1120,11 @@ Result<std::shared_ptr<DataType>> GetArrowType(const liborc::Type* type) {
     case liborc::CHAR:
       return fixed_size_binary(static_cast<int>(type->getMaximumLength()));
     case liborc::TIMESTAMP:
+      // The timestamp values stored in ORC are in the writer timezone.
       return timestamp(TimeUnit::NANO);
```
wjones127, line 1124: Is this true? I would have expected the values in the "UTC" timezone and the expectations is readers should interpret it in the local time zone.

wgtmac: Yes, you're right.

For `orc::TIMESTAMP` type:
- The Orc writer expects input data (i.e. in the orc::TimestampVectorBatch) to be in the "UTC" timezone and serializes it into the writer timezone: https://github.com/apache/orc/blob/main/c%2B%2B/src/ColumnWriter.cc#L1717
- The Orc reader deserializes the data from writer timezone and restores it into reader timezone: https://github.com/apache/orc/blob/main/c%2B%2B/src/ColumnReader.cc#L336

For `orc::TIMESTAMP_INSTANT` type:
- The Orc writer expects input data to be in the "UTC" timezone and serializes it into the "UTC" timezone: https://github.com/apache/orc/blob/main/c%2B%2B/src/ColumnWriter.cc#L1644
- The Orc reader deserializes the data from "UTC" timezone and no more conversion is needed because writerTimezone and readerTimezone are both "UTC": https://github.com/apache/orc/blob/main/c%2B%2B/src/ColumnReader.cc#L282

We have seen many issues around `orc::TIMESTAMP` type because of the writer-reader timezone conversion, especially with different day-light saving rules. So that's why `orc::TIMESTAMP_INSTANT` type is added and is always preferred over `orc::TIMESTAMP` type if user can take care of the timezone.

wjones127: Okay that makes sense. It does make me wonder if we should then be applying the local time zone to these types, instead of using naive timestamps:

```suggestion
      // The timestamp values stored in ORC are in the writer timezone.
      return timestamp(TimeUnit::NANO, GetLocalTimeZone());
```

wgtmac: This would fail the equality check of timestamp types in the roundtrip tests. Additionally, it seems that there is a lack of utility like `GetLocalTimeZone()`?

IMHO, if writer uses naive timestamp types, reader should follow the same pattern.

cpp/src/arrow/adapters/orc/util.cc
```diff
@@ -1111,7 +1121,20 @@ Result<std::shared_ptr<DataType>> GetArrowType(const liborc::Type* type) {
     case liborc::CHAR:
       return fixed_size_binary(static_cast<int>(type->getMaximumLength()));
     case liborc::TIMESTAMP:
+      // Values of TIMESTAMP type are stored in the writer timezone in the Orc file.
```
wgtmac, line 1124: I have added extensive comment to explain the issue and suggested usage. Please check again. Thanks @wjones127

wjones127 (approved):
Thank you for adding those comments!

PR #34593, opened 2023-03-16
GH-34703: [Python] Set copy=False explicitly when creating a pandas Series

phofl:
### Rationale for this change

pandas will change the default for creating a Series from an array (numpy, arrow, ...) to copy=True when Copy-on-Write is enabled. To avoid this when using it internally, we have to specify copy=False explicitly.

Setting copy=False when creating a Series

### Are these changes tested?

This is equivalent to the current default behavior, so no reason to add any additional tests. 

no

cc @jorisvandenbossche 
* Closes: #34703

### DIFF
--- a/python/pyarrow/array.pxi
+++ b/python/pyarrow/array.pxi
@@ -1664,7 +1664,7 @@ cdef _array_like_to_pandas(obj, options, types_mapper):
     else:
         dtype = None
 
-    result = pandas_api.series(arr, dtype=dtype, name=name)
+    result = pandas_api.series(arr, dtype=dtype, name=name, copy=False)
 
     if (isinstance(original_type, TimestampType) and
             original_type.tz is not None and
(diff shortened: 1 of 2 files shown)


AlenkaF:
Could you open an issue for this PR as it is not a minor change even though the diff is small ;) https://github.com/apache/arrow/blob/main/CONTRIBUTING.md#Minor-Fixes

phofl:
This doesn’t change anything right now, since copy=False is still the default in pandas for quite some time. Not setting this explicitly makes testing in pandas tricky though.

can open an issue if you prefer, but there isn’t really any change at all right now and since you are using series only with arrays internally there won’t be a user facing change at all (if I am not mistaken)

AlenkaF (approved):

AlenkaF:
> This doesn’t change anything right now, since copy=False is still the default in pandas for quite some time. Not setting this explicitly makes testing in pandas tricky though.
> 
> can open an issue if you prefer, but there isn’t really any change at all right now and since you are using series only with arrays internally there won’t be a user facing change at all (if I am not mistaken)

Agree and totally understand. The issue creation has more to do with the workflow we currently have (only docs changes can be minor PR).

Issue #34605, opened 2023-03-17
[C++] Segment fault when accessing output_schema of a named tap provider declaration info


icexelloss:
### Describe the bug, including details regarding any error messages, version, and platform.

In internal testing, we found a segment fault issue when executing a named tap relation. After debugging, the segfault is caused by accessing the output_schema field of the named tap declaration info, caused by this code:

https://github.com/apache/arrow/blob/main/cpp/src/arrow/engine/substrait/options.cc#L134

### Component(s)

C++

PR #34606, opened 2023-03-17
GH-34605: [C++] Don't use std::move when passing shared_ptr to named table …

icexelloss:
…provider and declaration info

### Rationale for this change

In internal testing, we found a segment fault issue when executing a named tap relation. After debugging, the segfault is caused by accessing the output_schema field of the named tap declaration info, caused by this code:

https://github.com/apache/arrow/blob/main/cpp/src/arrow/engine/substrait/options.cc#L134

This PR removes "std::move" when passing shared_ptr of the Schema and thus prevent segfault when accessing this field from named tap provider or DeclarationInfo.

### Are these changes tested?
I verified with internal tests that this PR fixes the issue.

No

### DIFF
--- a/cpp/src/arrow/engine/substrait/options.cc
+++ b/cpp/src/arrow/engine/substrait/options.cc
@@ -161,9 +161,8 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
     ARROW_ASSIGN_OR_RAISE(auto renamed_schema, schema->WithNames(columns));
     auto input_decls = MakeDeclarationInputs(inputs);
     ARROW_ASSIGN_OR_RAISE(
-        auto decl,
-        conv_opts.named_tap_provider(named_tap_rel.kind(), input_decls,
-                                     named_tap_rel.name(), renamed_schema));
+        auto decl, conv_opts.named_tap_provider(named_tap_rel.kind(), input_decls,
+                                                named_tap_rel.name(), renamed_schema));
     return RelationInfo{{std::move(decl), renamed_schema}, std::nullopt};
   }
 };
--- a/cpp/src/arrow/engine/substrait/options.cc
+++ b/cpp/src/arrow/engine/substrait/options.cc
@@ -163,8 +163,8 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
     ARROW_ASSIGN_OR_RAISE(
         auto decl,
         conv_opts.named_tap_provider(named_tap_rel.kind(), input_decls,
-                                     named_tap_rel.name(), std::move(renamed_schema)));
-    return RelationInfo{{std::move(decl), std::move(renamed_schema)}, std::nullopt};
+                                     named_tap_rel.name(), renamed_schema));
+    return RelationInfo{{std::move(decl), renamed_schema}, std::nullopt};
   }
 };
 
--- a/cpp/src/arrow/engine/substrait/options.cc
+++ b/cpp/src/arrow/engine/substrait/options.cc
@@ -163,7 +163,7 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
     ARROW_ASSIGN_OR_RAISE(
         auto decl, conv_opts.named_tap_provider(named_tap_rel.kind(), input_decls,
                                                 named_tap_rel.name(), renamed_schema));
-    return RelationInfo{{std::move(decl), renamed_schema}, std::nullopt};
+    return RelationInfo{{std::move(decl), std::move(renamed_schema)}, std::nullopt};
   }
 };
 


cpp/src/arrow/engine/substrait/options.cc
```diff
@@ -163,8 +163,8 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
     ARROW_ASSIGN_OR_RAISE(
         auto decl,
         conv_opts.named_tap_provider(named_tap_rel.kind(), input_decls,
-                                     named_tap_rel.name(), std::move(renamed_schema)));
-    return RelationInfo{{std::move(decl), std::move(renamed_schema)}, std::nullopt};
+                                     named_tap_rel.name(), renamed_schema));
```
icexelloss, line 166: I think passing `renamed_schema` without `std::move` is correct because `std::move` would nullify all other shared pointers and cause segment fault when either named_tap_provider or Declaration info tries to access it.

icexelloss:
cc @westonpace

cpp/src/arrow/engine/substrait/options.cc
```diff
@@ -161,10 +161,9 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
     ARROW_ASSIGN_OR_RAISE(auto renamed_schema, schema->WithNames(columns));
     auto input_decls = MakeDeclarationInputs(inputs);
     ARROW_ASSIGN_OR_RAISE(
-        auto decl,
-        conv_opts.named_tap_provider(named_tap_rel.kind(), input_decls,
-                                     named_tap_rel.name(), std::move(renamed_schema)));
-    return RelationInfo{{std::move(decl), std::move(renamed_schema)}, std::nullopt};
+        auto decl, conv_opts.named_tap_provider(named_tap_rel.kind(), input_decls,
+                                                named_tap_rel.name(), renamed_schema));
+    return RelationInfo{{std::move(decl), renamed_schema}, std::nullopt};
```
mapleFU, line 166: I think the bug occurs because the renamed_schema is moved twice, but can we copy the first one, and keep `std::move` on the second one?

icexelloss: I don't think that would be sufficient. IIUC, std::move would nullify all other shared_ptr, meaning if inside "named_tap_provider" the shared_ptr is passed to some other object (since it receives a shared_ptr, it is a reasonable thing to do) then it would be nullified as well after we `std::move` the second.

icexelloss:
@westonpace Since this is blocking our internal bump to newer Arrow, I would like to merge this to EoD today to get it into the nightly build (so I can pull into next Monday).  If you are too busy to review this, or somehow missed this, I will merge this EoD and address any comments / request for changes as follow up since this is small change.

I am fairly confident that this change is at least "correct" and have validated that this indeed fixes the issue I am seeing internally.

cpp/src/arrow/engine/substrait/options.cc
mapleFU: `NamedTapProvider` takes `std::shared_ptr` as input.

```c++
using NamedTapProvider = std::function<Result<compute::Declaration>(
    const std::string&, std::vector<compute::Declaration::Input>, const std::string&,
    std::shared_ptr<Schema>)>;
```

`std::shared_ptr` is a smart pointer, which is consitute from a control block(with refcount) and data block. Calling copy to a `std::shared_ptr` would copy the data block, copy the and inc the refcount. When calling move, it will "move" control block and data block from moved pointer.

In this case, `conv_opts.named_tap_provider( ..., renamed_schema))` will **copy** the shared_ptr. When it's copied, the original is not changed. And previously, when move is called, the `renamed_schema` would be set to null.

On the second place, calling move will move the `renamed_schema` to `RelationInfo`'s constructor. It's ok to move, because `renamed_schema` will not be used in the remaining part of this function .

The previous copied `renamed_schema` will not be changed, because it's control block and data block would not be changed.

If `named_tap_provider` just keeps `const std::shared_ptr<T>&`, rather than a `std::shared_ptr<T>`, we need to copy. Otherwise, move the last `renamed_schema` is ok

mapleFU: (If you'd like to keep it, it's also ok, bugfix is much more important. But I think we could make it better)

westonpace: @mapleFU is correct.  It would be ok to leave the second `std::move`.  You should not (usually) use an object after `std::move` has been called on it.  However, it is ok to use the object before `std::move` is called on it.

icexelloss: Ah thanks both. I have added the second move back.

mapleFU (approved):

PR #34617, opened 2023-03-17
GH-34615: [CI][C++] Add CI job for basic format support without ARROW_COMPUTE

### LINKED ISSUE #34615: [CI][C++] Add CI job for basic format support without ARROW_COMPUTE

Following the compute kernel split in https://github.com/apache/arrow/pull/34295, both libarrow and libparquet should now be fully functional without `ARROW_COMPUTE`. Previously, some tests in the core library were conditionally excluded - primarily due their usage of cast kernels. Thus, `ARROW_COMPUTE` is more-or-less treated as mandatory by the current CI jobs (in the interest maximum coverage). However, this should no longer be necessary.

It would be useful to include `ARROW_CSV` as well, since the CSV writer uses `Cast` in its implementation.

### Component(s)

C++, Continuous Integration

### LINKED ISSUE #34655: [CI][C++] arrow-compute-internals-test fails with `No function registered with name: equal` on test-cuda-cpp

The [test-cuda-cpp](https://github.com/ursacomputing/crossbow/actions/runs/4474260045/jobs/7862536357) nightly job fails with the following error:

```
[----------] 3 tests from ExpressionUtils
[ RUN      ] ExpressionUtils.Comparison
/arrow/cpp/src/arrow/compute/expression_test.cc:90: Failure
  maybe_actual
    Which is: Key error: No function registered with name: equal
  maybe_expected
    Which is: "equal"
/arrow/cpp/src/arrow/compute/expression_test.cc:90: Failure
  maybe_actual
    Which is: Key error: No function registered with name: equal
  maybe_expected
    Which is: "less"
/arrow/cpp/src/arrow/compute/expression_test.cc:90: Failure
  maybe_actual
    Which is: Key error: No function registered with name: equal
  maybe_expected
    Which is: "greater"
/arrow/cpp/src/arrow/compute/expression_test.cc:90: Failure
  maybe_actual
    Which is: Key error: No function registered with name: equal
  maybe_expected
    Which is: "na"
/arrow/cpp/src/arrow/compute/expression_test.cc:90: Failure
  maybe_actual
    Which is: Key error: No function registered with name: equal
  maybe_expected
    Which is: "na"
/arrow/cpp/src/arrow/compute/expression_test.cc:92: Failure
  _st.CodeAsString()
    Which is: "Key error"
  Status::CodeAsString(expected.status().code())
    Which is: "NotImplemented"
/arrow/cpp/src/arrow/compute/expression_test.cc:92: Failure
Value of: _st.ToString()
Expected: has substring "no kernel matching input types"
  Actual: "Key error: No function registered with name: equal"
/arrow/cpp/src/arrow/compute/expression_test.cc:92: Failure
  _st.CodeAsString()
    Which is: "Key error"
  Status::CodeAsString(expected.status().code())
    Which is: "NotImplemented"
/arrow/cpp/src/arrow/compute/expression_test.cc:92: Failure
Value of: _st.ToString()
Expected: has substring "no kernel matching input types"
  Actual: "Key error: No function registered with name: equal"
/arrow/cpp/src/arrow/compute/expression_test.cc:92: Failure
  _st.CodeAsString()
    Which is: "Key error"
  Status::CodeAsString(expected.status().code())
    Which is: "NotImplemented"
/arrow/cpp/src/arrow/compute/expression_test.cc:92: Failure
Value of: _st.ToString()
Expected: has substring "no kernel matching input types"
  Actual: "Key error: No function registered with name: equal"
/arrow/cpp/src/arrow/compute/expression_test.cc:90: Failure
  maybe_actual
    Which is: Key error: No function registered with name: equal
  maybe_expected
    Which is: "equal"
/arrow/cpp/src/arrow/compute/expression_test.cc:90: Failure
  maybe_actual
    Which is: Key error: No function registered with name: equal
  maybe_expected
    Which is: "equal"
/arrow/cpp/src/arrow/compute/expression_test.cc:90: Failure
  maybe_actual
    Which is: Key error: No function registered with name: equal
  maybe_expected
    Which is: "less"
/arrow/cpp/src/arrow/compute/expression_test.cc:90: Failure
  maybe_actual
    Which is: Key error: No function registered with name: equal
  maybe_expected
    Which is: "less"
[  FAILED  ] ExpressionUtils.Comparison (0 ms)
```
This can be reproduced locally with:
```
$ archery docker run ubuntu-cuda-cpp
```

### Component(s)

C++, Continuous Integration

benibus:
This adds a crossbow job for `ARROW_IPC`, `ARROW_PARQUET`, and `ARROW_CSV` - based on a minimal Ubuntu image.

The job primarily aims to test the core Arrow library + basic format support without the full kernel registry provided by `ARROW_COMPUTE`. Note that `ARROW_JSON` is implicitly enabled as well, since it's a dependency of `ARROW_TESTING`.

* Closes: #34615
* Closes: #34655

### DIFF
--- a/dev/tasks/tasks.yml
+++ b/dev/tasks/tasks.yml
@@ -1255,6 +1255,16 @@ tasks:
         UBUNTU: 20.04
       image: ubuntu-cpp-thread-sanitizer
 
+  test-ubuntu-20.04-cpp-minimal-with-formats:
+    ci: github
+    template: docker-tests/github.linux.yml
+    params:
+      env:
+        ARROW_CSV: ON
+        ARROW_PARQUET: ON
+        UBUNTU: 20.04
+      image: ubuntu-cpp-minimal
+
 {% for python_version in ["3.7", "3.8", "3.9", "3.10", "3.11"] %}
   test-conda-python-{{ python_version }}:
     ci: github
--- a/dev/tasks/tasks.yml
+++ b/dev/tasks/tasks.yml
@@ -1260,9 +1260,8 @@ tasks:
     template: docker-tests/github.linux.yml
     params:
       env:
-        ARROW_CSV: ON
-        ARROW_PARQUET: ON
         UBUNTU: 20.04
+      flags: "-e ARROW_CSV=ON -e ARROW_PARQUET=ON"
       image: ubuntu-cpp-minimal
 
 {% for python_version in ["3.7", "3.8", "3.9", "3.10", "3.11"] %}
--- a/ci/docker/ubuntu-20.04-cpp-minimal.dockerfile
+++ b/ci/docker/ubuntu-20.04-cpp-minimal.dockerfile
@@ -33,6 +33,7 @@ RUN apt-get update -y -q && \
         libssl-dev \
         libcurl4-openssl-dev \
         python3-pip \
+        tzdata \
         wget && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists*
--- a/ci/docker/ubuntu-22.04-cpp-minimal.dockerfile
+++ b/ci/docker/ubuntu-22.04-cpp-minimal.dockerfile
@@ -33,6 +33,7 @@ RUN apt-get update -y -q && \
         libssl-dev \
         libcurl4-openssl-dev \
         python3-pip \
+        tzdata \
         wget && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists*
--- a/cpp/src/arrow/compute/CMakeLists.txt
+++ b/cpp/src/arrow/compute/CMakeLists.txt
@@ -26,7 +26,14 @@ arrow_add_pkg_config("arrow-compute")
 # Unit tests
 #
 
-# The following kernels are always present:
+set(ARROW_COMPUTE_TEST_PREFIX "arrow-compute")
+set(ARROW_COMPUTE_TEST_LABELS "arrow_compute")
+set(ARROW_COMPUTE_TEST_ARGS PREFIX ${ARROW_COMPUTE_TEST_PREFIX} LABELS
+                            ${ARROW_COMPUTE_TEST_LABELS})
+
+# This will only add the test if ARROW_COMPUTE is enabled, meaning the full kernel registry is available.
+#
+# The following kernels are always present in default builds:
 #  - array_filter
 #  - array_take
 #  - cast
@@ -38,10 +45,12 @@ arrow_add_pkg_config("arrow-compute")
 #  - unique
 #  - value_counts
 #
-# Tests that use additional kernels should specify REQUIRE_ALL_KERNELS to avoid
-# being included in minimal builds. See: GH-34388
+# Also see: GH-34388, GH-34615
 function(ADD_ARROW_COMPUTE_TEST REL_TEST_NAME)
-  set(options REQUIRE_ALL_KERNELS)
+  if(NOT ARROW_COMPUTE)
+    return()
+  endif()
+
   set(one_value_args PREFIX)
   set(multi_value_args LABELS)
   cmake_parse_arguments(ARG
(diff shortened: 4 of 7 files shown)


kou:
There are still some enabled components:

https://github.com/apache/arrow/actions/runs/4462272725/jobs/7836694435?pr=34617#step:6:1866

```text
--   ARROW_BUILD_UTILITIES=ON [default=OFF]
--       Build Arrow commandline utilities
--   ARROW_CSV=ON [default=OFF]
--       Build the Arrow CSV Parser Module
--   ARROW_FILESYSTEM=ON [default=OFF]
--       Build the Arrow Filesystem Layer
--   ARROW_IPC=ON [default=ON]
--       Build the Arrow IPC extensions
--   ARROW_JSON=ON [default=OFF]
--       Build Arrow with JSON support (requires RapidJSON)
--   ARROW_PARQUET=ON [default=OFF]
--       Build the Parquet libraries
```

Are they intentional?

benibus:
> Are they intentional?

`ARROW_BUILD_UTILITIES` and `ARROW_FILESYSTEM` are unintentional (I think the latter is being enabled by `PARQUET_REQUIRE_ENCRYPTION=ON`). Regarding the others, see my edit to the [body message](https://github.com/apache/arrow/pull/34617#issue-1629870680) (I attempted to clarify my intent as well).

benibus:
Also, fair warning: I've never actually messed with CI system before so there may be a better way to do this. I just went with one of the preexisting Ubuntu images since it seemed like a reasonable base (to me, at least).

kou:
OK. Then "minimal" may be misleading. For example, we use "minimal" as "all components are disabled" in https://github.com/apache/arrow/blob/main/cpp/CMakePresets.json#L49 . Can we use other label for this job?

If we enable Parquet, CI time isn't short. How about running this as a nightly CI job instead like `test-*-cpp-*` in https://github.com/apache/arrow/blob/main/dev/tasks/tasks.yml#L1171-L1256 ?

benibus:
I agree that it's misleading.

We could also handle both cases pretty easily, I think. The "minimal" image I have here could disable everything as it implies. Then, the nightly task would use that image but enable CSV/Parquet.

benibus:
@github-actions crossbow submit test-ubuntu-20.04-cpp-minimal-with-formats

.github/workflows/cpp.yml
```diff
@@ -69,6 +69,13 @@ jobs:
             runs-on: ubuntu-latest
             title: AMD64 Conda C++
             ubuntu: "22.04"
+          - arch: amd64
+            clang-tools: "14"
+            image: ubuntu-cpp-minimal
+            llvm: "14"
+            runs-on: ubuntu-latest
+            title: AMD64 Ubuntu 22.04 C++ Minimal
+            ubuntu: "22.04"
```
kou, line 78: Could you revert this?

dev/tasks/tasks.yml
```diff
@@ -1255,6 +1255,16 @@ tasks:
         UBUNTU: 20.04
       image: ubuntu-cpp-thread-sanitizer
 
+  test-ubuntu-20.04-cpp-minimal-with-formats:
+    ci: github
+    template: docker-tests/github.linux.yml
+    params:
+      env:
+        ARROW_CSV: ON
+        ARROW_PARQUET: ON
```
kou, line 1264: We need to pass them as `params.flags` not `params.env`:

```yaml
params:
 flags: "-e ARROW_CSV=ON -e ARROW_PARQUET=ON"
```

kou:
Could you also update "minimal" in the title and the description of this PR and GH-34615.

kou:
@github-actions crossbow submit test-ubuntu-20.04-cpp-minimal-with-formats

raulcd:
I've updated the body to include the closes https://github.com/apache/arrow/issues/34655 as suggested.

raulcd:
The test-cuda-cpp seems to fail with a similar error to the one reported on #34655 :
```
 '_error_or_value25.status()' failed with Key error: No function registered with name: sort_indices
/arrow/cpp/src/arrow/compute/exec/order_by_node_test.cc:61: Failure
Failed
```
@benibus let me know if I should open a different issue for that or should be fixed here. Thanks!

benibus:
@kou @raulcd test-ubuntu-20.04-cpp-minimal-with-formats and test-cuda-cpp should pass on this branch now.

benibus:
@github-actions crossbow submit test-ubuntu-20.04-cpp-minimal-with-formats

kou (approved):
+1

PR #3806, opened 2019-03-04
ARROW-4719: [C#] Implement ChunkedArray, Column and Table in C#

pgovind:
Port ChunkedArray, Column and Table to C#

@eerhardt @chutchinson

PR #34618, opened 2023-03-17
GH-25163: [C#] Support half-float arrays.

### LINKED ISSUE #25163: [C#] Support Float16
With <https://github.com/dotnet/runtime/issues/936>, .NET is getting a `System.Half` type, which is a 16-bit floating point number. Once that type lands in .NET we can implement support for the Float16 type in Arrow.

**Reporter**: [Eric Erhardt](https://issues.apache.org/jira/browse/ARROW-9048) / @eerhardt

<sub>**Note**: *This issue was originally created as [ARROW-9048](https://issues.apache.org/jira/browse/ARROW-9048). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

teo-tsirpanis:
### Rationale for this change

.NET 5 introduced the [`System.Half`](https://devblogs.microsoft.com/dotnet/introducing-the-half-type/) type, which represents 16-bit floats. This PR adds support for them in Apache Arrow.

I multi-targeted the `Apache.Arrow` project to .NET 6 (because .NET 5 is unsupported) and added a `HalfFloatArray` type with a very similar implementation as the other floating-point array types.

I also updated the README.

### Are these changes tested?

Yes. I also refactored the array tests to reduce duplication among the various numeric types.

Yes.
* Closes: #25163


csharp/src/Apache.Arrow/Apache.Arrow.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.3;netstandard2.0;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>netstandard1.3;netstandard2.0;netcoreapp3.1;net6.0</TargetFrameworks>
```
teo-tsirpanis, line 4: How about I remove .NET Standard 1.3 and .NET Core 3.1?

csharp/src/Apache.Arrow/Apache.Arrow.csproj
```diff
@@ -1,7 +1,7 @@
...
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <DefineConstants>$(DefineConstants);UNSAFE_BYTEBUFFER;BYTEBUFFER_NO_BOUNDS_CHECK;ENABLE_SPAN_T</DefineConstants>
```
teo-tsirpanis, line 6: How about I remove `ENABLE_SPAN_T`? Is there any value in keeping it? Some stuff has two implementations and one is untested.

westonpace (commented):
This seems like a fairly minimal set of tests for a new types.  Also, from some searching, it appears there are quite a few places we are visiting the various types:

RecordBatch.Builder.cs
ArrowArrayBuilderFactory.cs
ArrayBuilderTests.cs
ArrowReaderVerifier.cs
ArrowStreamWriter.cs

Do we need to add half-float support there?

Do you know if C# runs the integration tests (these are tests that read/write known golden IPC files from other implementations)? Although I'm not sure if we have half-float arrays in the IPC integration tests.

csharp/README.md
```diff
@@ -126,12 +126,10 @@ for currently available features.
     - Dictionary Encoding
 - Types
     - Tensor
-    - Table
```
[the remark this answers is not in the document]
teo-tsirpanis: Because `Table` is supported. It got added in #3806, which is closed instead of merged for some reason.

teo-tsirpanis:
Thanks for the review @westonpace. I addressed your comments.

> Do you know if C# runs the integration tests […]?

I have no idea.

csharp/src/Apache.Arrow/Apache.Arrow.csproj
eerhardt: The current guidance is to avoid targeting netstandard1.x: https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting

But let's not do it in this PR. Instead, I think we should open a new issue for it. We may want to wait on netcoreapp3.1 though. It was just EOL'd a couple months ago. It sometimes takes people a bit longer to move off these EOL frameworks. It doesn't cost much to keep supporting it for a few months longer.

csharp/src/Apache.Arrow/Apache.Arrow.csproj
eerhardt: You don't want to remove that define. It makes the flatbuffer code work better.

eerhardt:
> Although I'm not sure if we have half-float arrays in the IPC integration tests.

I'm not seeing half-float array in the current integration tests. Is that a test hole?

eerhardt (commented):
Thanks for the contribution, @teo-tsirpanis!

I just have a few comments. This looks really good.

csharp/src/Apache.Arrow/Arrays/HalfFloatArray.cs
```diff
@@ -0,0 +1,46 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Apache.Arrow.Types;
+using System;
+
+namespace Apache.Arrow
+{
+    public class HalfFloatArray : PrimitiveArray<Half>
```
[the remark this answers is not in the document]
teo-tsirpanis: None of the other array types are documented. 😅

docs/source/status.rst
```diff
@@ -101,7 +101,8 @@ Data Types
 
 Notes:
 
-* \(1) Nested dictionaries not supported
+* \(1) FLoat16 support in C# is only available when targeting .NET 6+.
```
eerhardt, line 104:
```suggestion
* \(1) Float16 support in C# is only available when targeting .NET 6+.
```

eerhardt (approved):
Thanks again for the contribution @teo-tsirpanis!

I'll merge this unless I see more feedback.

westonpace:
> I'm not seeing half-float array in the current integration tests. Is that a test hole?

I don't think we need to hold this PR up for it but that does seem like something we'd like to fix at some point.  i don't think too many implementations have half-float support at the moment so maybe it isn't surprising.

Issue #32897, opened 2022-09-09
[C++] Change Acero Aggregate output order


asfimport:
At the moment Acero Aggregates output contains aggregate fields followed by key fields. In Substrait this is the other way around. Also, it is not necessary to keep this static and use a project node to fulfill the requirement. It is rather important to modify this order and update the corresponding test cases. 

**Reporter**: [Vibhatha Lakmal Abeykoon](https://issues.apache.org/jira/browse/ARROW-17656) / @vibhatha

<sub>**Note**: *This issue was originally created as [ARROW-17656](https://issues.apache.org/jira/browse/ARROW-17656). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

westonpace:
Closing as a duplicate of #33616

PR #34627, opened 2023-03-17
GH-34626:  [C++] Add ordered/segmented aggregation Substrait extension

### LINKED ISSUE #34626: [C++] Add ordered/segmented aggregation Substrait extension

Add a Substrait extension for ordered/segmented aggregation, which was added in #32884

### Component(s)

C++

rtpsw:
See #34626
* Closes: #34626

### DIFF
--- a/cpp/proto/substrait/extension_rels.proto
+++ b/cpp/proto/substrait/extension_rels.proto
@@ -61,11 +61,11 @@ message NamedTapRel {
 
 message SegmentedAggregateRel {
   // Grouping keys of the aggregation
-  repeated substrait.Expression.ReferenceSegment grouping_keys = 2;
+  repeated substrait.Expression.ReferenceSegment grouping_keys = 1;
 
   // Segment keys of the aggregation
-  repeated substrait.Expression.ReferenceSegment segment_keys = 3;
+  repeated substrait.Expression.ReferenceSegment segment_keys = 2;
 
   // A list of one or more aggregate expressions along with an optional filter.
-  repeated substrait.AggregateRel.Measure measures = 4;
+  repeated substrait.AggregateRel.Measure measures = 3;
 }
--- a/cpp/proto/substrait/extension_rels.proto
+++ b/cpp/proto/substrait/extension_rels.proto
@@ -60,8 +60,6 @@ message NamedTapRel {
 }
 
 message SegmentedAggregateRel {
-  substrait.RelCommon common = 1;
-
   // Grouping keys of the aggregation
   repeated substrait.Expression.ReferenceSegment grouping_keys = 2;
 
--- a/cpp/proto/substrait/extension_rels.proto
+++ b/cpp/proto/substrait/extension_rels.proto
@@ -58,3 +58,16 @@ message NamedTapRel {
   // If empty, field names will be automatically generated.
   repeated string columns = 3;
 }
+
+message SegmentedAggregateRel {
+  substrait.RelCommon common = 1;
+
+  // Grouping keys of the aggregation
+  repeated substrait.Expression.ReferenceSegment grouping_keys = 2;
+
+  // Segment keys of the aggregation
+  repeated substrait.Expression.ReferenceSegment segment_keys = 3;
+
+  // A list of one or more aggregate expressions along with an optional filter.
+  repeated substrait.AggregateRel.Measure measures = 4;
+}
--- a/cpp/proto/substrait/extension_rels.proto
+++ b/cpp/proto/substrait/extension_rels.proto
@@ -61,10 +61,10 @@ message NamedTapRel {
 
 message SegmentedAggregateRel {
   // Grouping keys of the aggregation
-  repeated substrait.Expression.ReferenceSegment grouping_keys = 1;
+  repeated substrait.Expression.FieldReference grouping_keys = 1;
 
   // Segment keys of the aggregation
-  repeated substrait.Expression.ReferenceSegment segment_keys = 2;
+  repeated substrait.Expression.FieldReference segment_keys = 2;
 
   // A list of one or more aggregate expressions along with an optional filter.
   repeated substrait.AggregateRel.Measure measures = 3;
--- a/cpp/src/arrow/compute/exec/source_node.cc
+++ b/cpp/src/arrow/compute/exec/source_node.cc
@@ -28,8 +28,10 @@
 #include "arrow/compute/expression.h"
 #include "arrow/datum.h"
 #include "arrow/io/util_internal.h"
+#include "arrow/ipc/util.h"
 #include "arrow/result.h"
 #include "arrow/table.h"
+#include "arrow/util/align_util.h"
 #include "arrow/util/async_generator.h"
 #include "arrow/util/async_util.h"
 #include "arrow/util/checked_cast.h"
@@ -102,6 +104,19 @@ struct SourceNode : ExecNode, public TracedNode {
               batch_size = morsel_length;
             }
             ExecBatch batch = morsel.Slice(offset, batch_size);
+            for (auto& value : batch.values) {
+              if (value.is_array()) {
+                ARROW_ASSIGN_OR_RAISE(
+                    value, util::EnsureAlignment(value.make_array(), ipc::kArrowAlignment,
+                                                 default_memory_pool()));
+              }
+              if (value.is_chunked_array()) {
+                ARROW_ASSIGN_OR_RAISE(
+                    value,
+                    util::EnsureAlignment(value.chunked_array(), ipc::kArrowAlignment,
+                                          default_memory_pool()));
+              }
+            }
             if (has_ordering) {
               batch.index = batch_index;
             }
--- a/cpp/src/arrow/compute/exec/source_node.cc
+++ b/cpp/src/arrow/compute/exec/source_node.cc
@@ -110,12 +110,6 @@ struct SourceNode : ExecNode, public TracedNode {
                     value, util::EnsureAlignment(value.make_array(), ipc::kArrowAlignment,
                                                  default_memory_pool()));
               }
-              if (value.is_chunked_array()) {
-                ARROW_ASSIGN_OR_RAISE(
-                    value,
-                    util::EnsureAlignment(value.chunked_array(), ipc::kArrowAlignment,
-                                          default_memory_pool()));
-              }
             }
             if (has_ordering) {
               batch.index = batch_index;
--- a/cpp/src/arrow/engine/substrait/expression_internal.cc
+++ b/cpp/src/arrow/engine/substrait/expression_internal.cc
@@ -143,27 +143,27 @@ Result<compute::Expression> FromProto(const substrait::Expression::ReferenceSegm
                                       const ConversionOptions& conversion_options,
                                       std::optional<compute::Expression> in_expr) {
   auto in_ref = ref;
-  auto& out = in_expr;
+  auto& current = in_expr;
   while (ref != nullptr) {
     switch (ref->reference_type_case()) {
       case substrait::Expression::ReferenceSegment::kStructField: {
         auto index = ref->struct_field().field();
-        if (!out) {
+        if (!current) {
           // Root StructField (column selection)
-          out = compute::field_ref(FieldRef(index));
-        } else if (auto out_ref = out->field_ref()) {
+          current = compute::field_ref(FieldRef(index));
+        } else if (auto current_ref = current->field_ref()) {
           // Nested StructFields on the root (selection of struct-typed column
           // combined with selecting struct fields)
-          out = compute::field_ref(FieldRef(*out_ref, index));
-        } else if (out->call() && out->call()->function_name == "struct_field") {
+          current = compute::field_ref(FieldRef(*current_ref, index));
+        } else if (current->call() && current->call()->function_name == "struct_field") {
           // Nested StructFields on top of an arbitrary expression
           auto* field_options =
-              checked_cast<compute::StructFieldOptions*>(out->call()->options.get());
+              checked_cast<compute::StructFieldOptions*>(current->call()->options.get());
           field_options->field_ref = FieldRef(std::move(field_options->field_ref), index);
         } else {
           // First StructField on top of an arbitrary expression
-          out = compute::call("struct_field", {std::move(*out)},
-                              arrow::compute::StructFieldOptions({index}));
+          current = compute::call("struct_field", {std::move(*current)},
+                                  arrow::compute::StructFieldOptions({index}));
         }
 
         // Segment handled, continue with child segment (if any)
@@ -175,16 +175,16 @@ Result<compute::Expression> FromProto(const substrait::Expression::ReferenceSegm
         break;
       }
       case substrait::Expression::ReferenceSegment::kListElement: {
-        if (!out) {
+        if (!current) {
           // Root ListField (illegal)
           return Status::Invalid(
               "substrait::ListElement cannot take a Relation as an argument");
         }
 
         // ListField on top of an arbitrary expression
-        out = compute::call(
+        current = compute::call(
             "list_element",
-            {std::move(*out), compute::literal(ref->list_element().offset())});
+            {std::move(*current), compute::literal(ref->list_element().offset())});
 
         // Segment handled, continue with child segment (if any)
         if (ref->list_element().has_child()) {
@@ -195,18 +195,33 @@ Result<compute::Expression> FromProto(const substrait::Expression::ReferenceSegm
         break;
       }
       default:
-        // Unimplemented construct, break out of loop
-        out.reset();
+        // Unimplemented construct, break current of loop
+        current.reset();
         ref = nullptr;
     }
   }
-  if (out) {
-    return *std::move(out);
+  if (current) {
+    return *std::move(current);
   }
 
   return Status::NotImplemented(
       "conversion to arrow::compute::Expression from Substrait reference segment: ",
       in_ref ? in_ref->DebugString() : "null");
+}
+
+Result<FieldRef> DirectReferenceFromProto(
+    const substrait::Expression::ReferenceSegment* refseg, const ExtensionSet& ext_set,
+    const ConversionOptions& conversion_options) {
+  ARROW_ASSIGN_OR_RAISE(compute::Expression expr,
+                        FromProto(refseg, ext_set, conversion_options, {}));
+  const FieldRef* field_ref = expr.field_ref();
+  if (field_ref) {
+    return *field_ref;
+  } else {
+    return Status::Invalid(
+        "A direct reference was expected but a more complex expression was given "
+        "instead");
+  }
 }
 
 Result<SubstraitCall> FromProto(const substrait::AggregateFunction& func, bool is_hash,
--- a/cpp/src/arrow/engine/substrait/expression_internal.cc
+++ b/cpp/src/arrow/engine/substrait/expression_internal.cc
@@ -138,6 +138,77 @@ std::string EnumToString(int value, const google::protobuf::EnumDescriptor* desc
   return value_desc->name();
 }
 
+Result<compute::Expression> FromProto(const substrait::Expression::ReferenceSegment* ref,
+                                      const ExtensionSet& ext_set,
+                                      const ConversionOptions& conversion_options,
+                                      std::optional<compute::Expression> in_expr) {
+  auto in_ref = ref;
+  auto& out = in_expr;
+  while (ref != nullptr) {
+    switch (ref->reference_type_case()) {
+      case substrait::Expression::ReferenceSegment::kStructField: {
+        auto index = ref->struct_field().field();
+        if (!out) {
+          // Root StructField (column selection)
+          out = compute::field_ref(FieldRef(index));
+        } else if (auto out_ref = out->field_ref()) {
+          // Nested StructFields on the root (selection of struct-typed column
+          // combined with selecting struct fields)
+          out = compute::field_ref(FieldRef(*out_ref, index));
+        } else if (out->call() && out->call()->function_name == "struct_field") {
+          // Nested StructFields on top of an arbitrary expression
+          auto* field_options =
+              checked_cast<compute::StructFieldOptions*>(out->call()->options.get());
+          field_options->field_ref = FieldRef(std::move(field_options->field_ref), index);
+        } else {
+          // First StructField on top of an arbitrary expression
+          out = compute::call("struct_field", {std::move(*out)},
+                              arrow::compute::StructFieldOptions({index}));
+        }
+
+        // Segment handled, continue with child segment (if any)
+        if (ref->struct_field().has_child()) {
+          ref = &ref->struct_field().child();
+        } else {
+          ref = nullptr;
+        }
+        break;
+      }
+      case substrait::Expression::ReferenceSegment::kListElement: {
+        if (!out) {
+          // Root ListField (illegal)
+          return Status::Invalid(
+              "substrait::ListElement cannot take a Relation as an argument");
+        }
+
+        // ListField on top of an arbitrary expression
+        out = compute::call(
+            "list_element",
+            {std::move(*out), compute::literal(ref->list_element().offset())});
+
+        // Segment handled, continue with child segment (if any)
+        if (ref->list_element().has_child()) {
+          ref = &ref->list_element().child();
+        } else {
+          ref = nullptr;
+        }
+        break;
+      }
+      default:
+        // Unimplemented construct, break out of loop
+        out.reset();
+        ref = nullptr;
+    }
+  }
+  if (out) {
+    return *std::move(out);
+  }
+
+  return Status::NotImplemented(
+      "conversion to arrow::compute::Expression from Substrait reference segment: ",
+      in_ref ? in_ref->DebugString() : "null");
+}
+
 Result<SubstraitCall> FromProto(const substrait::AggregateFunction& func, bool is_hash,
                                 const ExtensionSet& ext_set,
                                 const ConversionOptions& conversion_options) {
@@ -193,67 +264,7 @@ Result<compute::Expression> FromProto(const substrait::Expression& expr,
       }
 
       const auto* ref = &expr.selection().direct_reference();
-      while (ref != nullptr) {
-        switch (ref->reference_type_case()) {
-          case substrait::Expression::ReferenceSegment::kStructField: {
-            auto index = ref->struct_field().field();
-            if (!out) {
-              // Root StructField (column selection)
-              out = compute::field_ref(FieldRef(index));
-            } else if (auto out_ref = out->field_ref()) {
-              // Nested StructFields on the root (selection of struct-typed column
-              // combined with selecting struct fields)
-              out = compute::field_ref(FieldRef(*out_ref, index));
-            } else if (out->call() && out->call()->function_name == "struct_field") {
-              // Nested StructFields on top of an arbitrary expression
-              auto* field_options =
-                  checked_cast<compute::StructFieldOptions*>(out->call()->options.get());
-              field_options->field_ref =
-                  FieldRef(std::move(field_options->field_ref), index);
-            } else {
-              // First StructField on top of an arbitrary expression
-              out = compute::call("struct_field", {std::move(*out)},
-                                  arrow::compute::StructFieldOptions({index}));
-            }
-
-            // Segment handled, continue with child segment (if any)
-            if (ref->struct_field().has_child()) {
-              ref = &ref->struct_field().child();
-            } else {
-              ref = nullptr;
-            }
-            break;
-          }
-          case substrait::Expression::ReferenceSegment::kListElement: {
-            if (!out) {
-              // Root ListField (illegal)
-              return Status::Invalid(
-                  "substrait::ListElement cannot take a Relation as an argument");
-            }
-
-            // ListField on top of an arbitrary expression
-            out = compute::call(
-                "list_element",
-                {std::move(*out), compute::literal(ref->list_element().offset())});
-
-            // Segment handled, continue with child segment (if any)
-            if (ref->list_element().has_child()) {
-              ref = &ref->list_element().child();
-            } else {
-              ref = nullptr;
-            }
-            break;
-          }
-          default:
-            // Unimplemented construct, break out of loop
-            out.reset();
-            ref = nullptr;
-        }
-      }
-      if (out) {
-        return *std::move(out);
-      }
-      break;
+      return FromProto(ref, ext_set, conversion_options, std::move(out));
     }
 
     case substrait::Expression::kIfThen: {
--- a/cpp/src/arrow/engine/substrait/expression_internal.cc
+++ b/cpp/src/arrow/engine/substrait/expression_internal.cc
@@ -209,11 +209,24 @@ Result<compute::Expression> FromProto(const substrait::Expression::ReferenceSegm
       in_ref ? in_ref->DebugString() : "null");
 }
 
+Result<compute::Expression> FromProto(const substrait::Expression::FieldReference* fref,
+                                      const ExtensionSet& ext_set,
+                                      const ConversionOptions& conversion_options,
+                                      std::optional<compute::Expression> in_expr) {
+  if (fref->reference_type_case() !=
+          substrait::Expression::FieldReference::kDirectReference ||
+      fref->root_type_case() != substrait::Expression::FieldReference::kRootReference) {
+    return Status::NotImplemented("substrait::FieldReference not direct root reference");
+  }
+  auto& dref = fref->direct_reference();
+  return FromProto(&dref, ext_set, conversion_options, std::move(in_expr));
+}
+
 Result<FieldRef> DirectReferenceFromProto(
-    const substrait::Expression::ReferenceSegment* refseg, const ExtensionSet& ext_set,
+    const substrait::Expression::FieldReference* fref, const ExtensionSet& ext_set,
     const ConversionOptions& conversion_options) {
   ARROW_ASSIGN_OR_RAISE(compute::Expression expr,
-                        FromProto(refseg, ext_set, conversion_options, {}));
+                        FromProto(fref, ext_set, conversion_options, {}));
   const FieldRef* field_ref = expr.field_ref();
   if (field_ref) {
     return *field_ref;
--- a/cpp/src/arrow/engine/substrait/expression_internal.h
+++ b/cpp/src/arrow/engine/substrait/expression_internal.h
@@ -20,6 +20,7 @@
 #pragma once
 
 #include <memory>
+#include <optional>
 
 #include "arrow/compute/type_fwd.h"
 #include "arrow/datum.h"
@@ -33,6 +34,11 @@ namespace arrow {
 namespace engine {
 
 class SubstraitCall;
+
+ARROW_ENGINE_EXPORT
+Result<compute::Expression> FromProto(const substrait::Expression::ReferenceSegment*,
+                                      const ExtensionSet&, const ConversionOptions&,
+                                      std::optional<compute::Expression>);
 
 ARROW_ENGINE_EXPORT
 Result<compute::Expression> FromProto(const substrait::Expression&, const ExtensionSet&,
--- a/cpp/src/arrow/engine/substrait/expression_internal.h
+++ b/cpp/src/arrow/engine/substrait/expression_internal.h
@@ -36,6 +36,10 @@ namespace engine {
 class SubstraitCall;
 
 ARROW_ENGINE_EXPORT
+Result<FieldRef> DirectReferenceFromProto(const substrait::Expression::ReferenceSegment*,
+                                          const ExtensionSet&, const ConversionOptions&);
+
+ARROW_ENGINE_EXPORT
 Result<compute::Expression> FromProto(const substrait::Expression::ReferenceSegment*,
                                       const ExtensionSet&, const ConversionOptions&,
                                       std::optional<compute::Expression>);
--- a/cpp/src/arrow/engine/substrait/expression_internal.h
+++ b/cpp/src/arrow/engine/substrait/expression_internal.h
@@ -36,11 +36,11 @@ namespace engine {
 class SubstraitCall;
 
 ARROW_ENGINE_EXPORT
-Result<FieldRef> DirectReferenceFromProto(const substrait::Expression::ReferenceSegment*,
+Result<FieldRef> DirectReferenceFromProto(const substrait::Expression::FieldReference*,
                                           const ExtensionSet&, const ConversionOptions&);
 
 ARROW_ENGINE_EXPORT
-Result<compute::Expression> FromProto(const substrait::Expression::ReferenceSegment*,
+Result<compute::Expression> FromProto(const substrait::Expression::FieldReference*,
                                       const ExtensionSet&, const ConversionOptions&,
                                       std::optional<compute::Expression>);
 
--- a/cpp/src/arrow/engine/substrait/options.cc
+++ b/cpp/src/arrow/engine/substrait/options.cc
@@ -190,9 +190,9 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
     // store key fields to be used when output schema is created
     std::vector<int> key_field_ids;
     std::vector<FieldRef> keys;
-    for (auto& key_refseg : seg_agg_rel.grouping_keys()) {
+    for (auto& ref : seg_agg_rel.grouping_keys()) {
       ARROW_ASSIGN_OR_RAISE(auto field_ref,
-                            DirectReferenceFromProto(&key_refseg, ext_set, conv_opts));
+                            DirectReferenceFromProto(&ref, ext_set, conv_opts));
       ARROW_ASSIGN_OR_RAISE(auto match, field_ref.FindOne(*input_schema));
       key_field_ids.emplace_back(std::move(match[0]));
       keys.emplace_back(std::move(field_ref));
@@ -201,9 +201,9 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
     // store segment key fields to be used when output schema is created
     std::vector<int> segment_key_field_ids;
     std::vector<FieldRef> segment_keys;
-    for (auto& key_refseg : seg_agg_rel.segment_keys()) {
+    for (auto& ref : seg_agg_rel.segment_keys()) {
       ARROW_ASSIGN_OR_RAISE(auto field_ref,
-                            DirectReferenceFromProto(&key_refseg, ext_set, conv_opts));
+                            DirectReferenceFromProto(&ref, ext_set, conv_opts));
       ARROW_ASSIGN_OR_RAISE(auto match, field_ref.FindOne(*input_schema));
       segment_key_field_ids.emplace_back(std::move(match[0]));
       segment_keys.emplace_back(std::move(field_ref));
--- a/cpp/src/arrow/engine/substrait/options.cc
+++ b/cpp/src/arrow/engine/substrait/options.cc
@@ -187,24 +187,22 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
     std::vector<int> key_field_ids;
     std::vector<FieldRef> keys;
     for (auto& key_refseg : seg_agg_rel.grouping_keys()) {
-      ARROW_ASSIGN_OR_RAISE(auto expr, FromProto(&key_refseg, ext_set, conv_opts, {}));
-      if (auto field_ref = expr.field_ref()) {
-        ARROW_ASSIGN_OR_RAISE(auto match, field_ref->FindOne(*input_schema));
-        key_field_ids.emplace_back(std::move(match[0]));
-        keys.emplace_back(std::move(*field_ref));
-      }
+      ARROW_ASSIGN_OR_RAISE(auto field_ref,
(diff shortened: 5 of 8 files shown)


rtpsw:
cc @icexelloss

rtpsw:
Marking as a draft until testing is added.

cpp/proto/substrait/extension_rels.proto
```diff
@@ -58,3 +58,9 @@ message NamedTapRel {
   // If empty, field names will be automatically generated.
   repeated string columns = 3;
 }
+
+message SegmentedAggregateRel {
+  substrait.AggregateRel aggregate = 1;
+
+  repeated substrait.AggregateRel.Grouping segment_groupings = 2;
```
icexelloss, line 65: I remember from the discussion that we prefer to use field index here?

cpp/src/arrow/engine/substrait/options.cc
```diff
@@ -166,6 +171,57 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
...
+    }
+
+    auto input_schema = inputs[0].output_schema;
+
+    ConversionOptions conversion_options;
+
+    // store segment key fields to be used when output schema is created
+    std::vector<int> segment_key_field_ids;
+    std::vector<FieldRef> segment_keys;
+    if (seg_agg_rel.segment_groupings_size() > 0) {
```
icexelloss, line 192: We can error out when there is no "segment" key. I cannot think of a valid case where segment is empty and we still want to use segmented aggregation so it's probably better to error out.

cpp/src/arrow/engine/substrait/options.cc
```diff
@@ -166,6 +171,57 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
                                      named_tap_rel.name(), std::move(renamed_schema)));
     return RelationInfo{{std::move(decl), std::move(renamed_schema)}, std::nullopt};
   }
+
+  Result<RelationInfo> MakeSegmentedAggregateRel(
```
icexelloss, line 175: I don't think this belongs in options.cc - this is not related to "options" really

cpp/src/arrow/engine/substrait/options.cc
```diff
@@ -166,6 +171,57 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
...
+      const ExtensionSet& ext_set) {
+    if (inputs.size() != 1) {
+      return Status::Invalid(
+          "substrait_ext::SegmentedAggregateRel requires a single input but got: ",
+          inputs.size());
+    }
+
+    auto input_schema = inputs[0].output_schema;
+
+    ConversionOptions conversion_options;
```
icexelloss, line 187: Should we use the conversion_options from the MakeRel function?

cpp/src/arrow/engine/substrait/options.cc
```diff
@@ -166,6 +171,57 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
...
+              return Status::OK();
+            }));
+
+    const auto& output_schema = decl_info.output_schema;
+    size_t out_size = output_schema->num_fields();
+    std::vector<int> field_output_indices(out_size);
+    for (int i = 0; i < static_cast<int>(out_size); i++) {
+      field_output_indices[i] = i;
+    }
+    return RelationInfo{decl_info, std::move(field_output_indices)};
```
icexelloss, line 223: can we use a null/nullopt field_output_indices? This is no-op right?

cpp/src/arrow/engine/substrait/options.cc
```diff
@@ -166,6 +171,57 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
...
+    if (seg_agg_rel.segment_groupings_size() > 0) {
+      ARROW_RETURN_NOT_OK(internal::ParseAggregateGrouping(
+          seg_agg_rel.segment_groupings(0), ext_set, conversion_options, input_schema,
+          &segment_key_field_ids, &segment_keys));
+    }
+
+    const auto& aggregate = seg_agg_rel.aggregate();
+    ARROW_ASSIGN_OR_RAISE(
+        auto decl_info,
+        internal::ParseAggregateDeclaration(
```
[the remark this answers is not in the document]
icexelloss: I am kind of between here, if we copy the AggregationRel and turn that into a new SegmentedAggregationRel, then I don't know if there is a reasonable way to reuse the deser code for AggregationRel  (or do we need to copy the code)?

If we use this approach, most of the deser code can be shared but the rel message looks a bit weird. So I am not sure which way of better.

@westonpace WDYT?

cpp/src/arrow/engine/substrait/options.cc
rtpsw: There are other `Make*Rel` functions already in `options.cc`, so I think such refactoring could be deferred to a separate PR.

cpp/src/arrow/engine/substrait/options.cc
rtpsw: Yes, I'll fix this.

cpp/src/arrow/engine/substrait/options.cc
rtpsw: `nullopt` has a different meaning - that `field_output_indices` is not defined rather than defined like the identity map - and will lead to an error.

cpp/src/arrow/engine/substrait/options.cc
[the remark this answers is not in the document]
rtpsw: IIRC, we discussed a couple of alternatives and leaned towards the copying one; another alternative was reusing `AggregateRel` in the message and leveraging the existing parsing code for, and while working on the PR I found that this was simpler to implement.

cpp/src/arrow/engine/substrait/relation_internal.h
```diff
@@ -46,5 +51,39 @@ Result<DeclarationInfo> FromProto(const substrait::Rel&, const ExtensionSet&,
 ARROW_ENGINE_EXPORT Result<std::unique_ptr<substrait::Rel>> ToProto(
     const compute::Declaration&, ExtensionSet*, const ConversionOptions&);
 
+namespace internal {
```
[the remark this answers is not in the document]
rtpsw: Indeed the header file is not public, yet the new functions are not really meant to be used sparingly, and the `internal` namespace makes this clear at each call site. I could remove the namespace and add a doc but then the clarity at the call sites would be lost.

cpp/src/arrow/engine/substrait/options.cc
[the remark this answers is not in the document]
westonpace: In Substrait itself we have been discouraging this kind of approach when creating physical relations because:

 * It's too expressive - We don't consume all kinds of AggregateRel (e.g. expressions in a grouping have to be direct references) and, since this is a physical relation, we should only expose what we can consume.
 * Unnecessary coupling - It's not possible to change AggregateRel without potentially changing all the extensions and it's not clear they would always need to change.
 * Directly including the Rel itself leads to some awkwardness like the fact that you now have multiple "inputs".

However, yes, most of the parsing code would then have to be duplicated.  So for something internal I don't know that it is unworkable.  So I don't have a strong opinion but I would lean slightly towards something like...

```
message SegmentedAggregateRel {
  repeated Expression.FieldReference grouping_keys = 0;
  repeated Expression.FieldReference segment_keys = 1;
  repeated substrait.AggregateRel.Measure measures = 2;
}
```

cpp/src/arrow/engine/substrait/options.cc
rtpsw: Done.

cpp/src/arrow/engine/substrait/options.cc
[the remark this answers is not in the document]
rtpsw: @icexelloss, do we have time now to try the approach Weston suggests?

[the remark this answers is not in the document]
icexelloss: Yes I think we should do what Weston suggests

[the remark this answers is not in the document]
rtpsw: Fixed.

rtpsw:
@westonpace, I believe the CI failures are unrelated, but please review them.

rtpsw:
> @rtpsw can you also remove the extra reviewers? Thanks

I don't seem to have permission to do this.

cpp/proto/substrait/extension_rels.proto
```diff
@@ -58,3 +58,16 @@ message NamedTapRel {
...
+  substrait.RelCommon common = 1;
+
+  // Grouping keys of the aggregation
+  repeated substrait.Expression.ReferenceSegment grouping_keys = 2;
+
+  // Segment keys of the aggregation
+  repeated substrait.Expression.ReferenceSegment segment_keys = 3;
+
+  // A list of one or more aggregate expressions along with an optional filter.
+  repeated substrait.AggregateRel.Measure measures = 4;
```
icexelloss, line 72: Not sure about reusing AggregationRel.Measure here, it feels to me this introduced an dependency between AggregationRel and SegmentedAggregateRel which we wanted to avoid? Maybe we should have SegmentedAggregateRel.Measure here - @westonpace WDYT?

cpp/src/arrow/engine/substrait/expression_internal.cc
```diff
@@ -138,6 +138,77 @@ std::string EnumToString(int value, const google::protobuf::EnumDescriptor* desc
...
+
+        // Segment handled, continue with child segment (if any)
+        if (ref->struct_field().has_child()) {
+          ref = &ref->struct_field().child();
+        } else {
+          ref = nullptr;
+        }
+        break;
+      }
+      case substrait::Expression::ReferenceSegment::kListElement: {
```
icexelloss, line 177: We can just unsupport this all together since we don't use this

cpp/src/arrow/engine/substrait/expression_internal.cc
```diff
@@ -138,6 +138,77 @@ std::string EnumToString(int value, const google::protobuf::EnumDescriptor* desc
   return value_desc->name();
 }
 
+Result<compute::Expression> FromProto(const substrait::Expression::ReferenceSegment* ref,
+                                      const ExtensionSet& ext_set,
+                                      const ConversionOptions& conversion_options,
+                                      std::optional<compute::Expression> in_expr) {
+  auto in_ref = ref;
+  auto& out = in_expr;
```
icexelloss, line 146: This is strange - why are we talking `in_expr` as out?

cpp/src/arrow/engine/substrait/expression_internal.cc
```diff
@@ -138,6 +138,77 @@ std::string EnumToString(int value, const google::protobuf::EnumDescriptor* desc
...
+      case substrait::Expression::ReferenceSegment::kStructField: {
+        auto index = ref->struct_field().field();
+        if (!out) {
+          // Root StructField (column selection)
+          out = compute::field_ref(FieldRef(index));
+        } else if (auto out_ref = out->field_ref()) {
+          // Nested StructFields on the root (selection of struct-typed column
+          // combined with selecting struct fields)
+          out = compute::field_ref(FieldRef(*out_ref, index));
+        } else if (out->call() && out->call()->function_name == "struct_field") {
```
icexelloss, line 158: don't think we need to support this now - this is using a field in struct field as the by key?

cpp/src/arrow/engine/substrait/relation_internal.cc
```diff
@@ -74,10 +74,9 @@ struct EmitInfo {
   std::shared_ptr<Schema> schema;
 };
 
-template <typename RelMessage>
-Result<EmitInfo> GetEmitInfo(const RelMessage& rel,
+Result<EmitInfo> GetEmitInfo(const substrait::RelCommon& rel_common,
```
icexelloss, line 77: Why this change

cpp/src/arrow/engine/substrait/relation_internal.cc
```diff
@@ -293,6 +303,90 @@ Status DiscoverFilesFromDir(const std::shared_ptr<fs::LocalFileSystem>& local_fs
   return Status::OK();
 }
 
+namespace internal {
+
+ARROW_ENGINE_EXPORT Status ParseAggregateMeasure(
```
icexelloss, line 308: Is this a refactor?

cpp/src/arrow/engine/substrait/relation_internal.cc
```diff
@@ -293,6 +303,90 @@ Status DiscoverFilesFromDir(const std::shared_ptr<fs::LocalFileSystem>& local_fs
   return Status::OK();
 }
 
+namespace internal {
```
icexelloss, line 306: I would be leaning towards removing the internal namespace here. I don't see this being different from other methods here. Plus the file is already named "relation_internal" so methods here either all belongs to an internal namespace, or don't, but I don't see a reason to split them.

cpp/src/arrow/engine/substrait/relation_internal.cc
```diff
@@ -293,6 +303,90 @@ Status DiscoverFilesFromDir(const std::shared_ptr<fs::LocalFileSystem>& local_fs
...
+    }
+
+    aggregates.push_back(std::move(arrow_agg));
+    return Status::OK();
+  } else {
+    return Status::Invalid("substrait::AggregateFunction not provided");
+  }
+}
+
+ARROW_ENGINE_EXPORT Result<DeclarationInfo> MakeAggregateDeclaration(
```
icexelloss, line 350: Why this change/refactor?

cpp/proto/substrait/extension_rels.proto
rtpsw: Note that this [was suggested](https://github.com/apache/arrow/pull/34627#discussion_r1145005055) by @westonpace.

cpp/src/arrow/engine/substrait/expression_internal.cc
```diff
@@ -138,6 +138,77 @@ std::string EnumToString(int value, const google::protobuf::EnumDescriptor* desc
...
+          auto* field_options =
+              checked_cast<compute::StructFieldOptions*>(out->call()->options.get());
+          field_options->field_ref = FieldRef(std::move(field_options->field_ref), index);
+        } else {
+          // First StructField on top of an arbitrary expression
+          out = compute::call("struct_field", {std::move(*out)},
+                              arrow::compute::StructFieldOptions({index}));
+        }
+
+        // Segment handled, continue with child segment (if any)
```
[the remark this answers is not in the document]
rtpsw: This part of the code is due to refactoring (not new code).

cpp/src/arrow/engine/substrait/expression_internal.cc
```diff
@@ -138,6 +138,77 @@ std::string EnumToString(int value, const google::protobuf::EnumDescriptor* desc
   return value_desc->name();
 }
 
+Result<compute::Expression> FromProto(const substrait::Expression::ReferenceSegment* ref,
```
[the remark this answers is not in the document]
rtpsw: Yes, this is a refactoring which extracted part of the original code that handled a reference segment into a reusable function.

cpp/src/arrow/engine/substrait/relation_internal.cc
rtpsw: This is a refactoring that created two overloads of the functions, the original first accepting a message (of some type, using the template) and the new second accepting a `RelCommon`. This is possible because the implementation of the original function is mostly just looking at the `RelCommon` of the message. This allows `MakeAggregateDeclaration` to be a non-template function by [invoking a non-template `ProcessEmit`](https://github.com/apache/arrow/pull/34627/files#diff-2c7a2e65948433ab4aa1fb595d3132338b0e9684ed57f8021c28e448dd8a8ec2R384).

cpp/src/arrow/engine/substrait/relation_internal.cc
rtpsw: Yes, this is another method-extraction refactoring.

cpp/src/arrow/engine/substrait/relation_internal.cc
rtpsw: My [rationale is explained here](https://github.com/apache/arrow/pull/34627#discussion_r1142728086). I'm fine with removing the namespace, if you feel strongly about it.

cpp/src/arrow/engine/substrait/relation_internal.cc
rtpsw: This is another method-extraction refactoring to support two call-sites, the [original one](https://github.com/apache/arrow/pull/34627/files#diff-2c7a2e65948433ab4aa1fb595d3132338b0e9684ed57f8021c28e448dd8a8ec2R836) and a [new one](https://github.com/apache/arrow/pull/34627/files#diff-8c5c127e8db8e52b40519262ee2b7e0179d4a39e5bcfe8b2165173b2b352d6a9R220).

cpp/src/arrow/engine/substrait/expression_internal.cc
```diff
@@ -138,6 +138,77 @@ std::string EnumToString(int value, const google::protobuf::EnumDescriptor* desc
...
+        } else if (out->call() && out->call()->function_name == "struct_field") {
+          // Nested StructFields on top of an arbitrary expression
+          auto* field_options =
+              checked_cast<compute::StructFieldOptions*>(out->call()->options.get());
+          field_options->field_ref = FieldRef(std::move(field_options->field_ref), index);
+        } else {
+          // First StructField on top of an arbitrary expression
+          out = compute::call("struct_field", {std::move(*out)},
+                              arrow::compute::StructFieldOptions({index}));
+        }
```
westonpace, line 167: The code here is perhaps more complex than the caller needs in many cases.  Ideally it would return a simple field reference.  The challenge is that Arrow's expressions don't support field references into intermediate outputs.  For example, if you have a function `min_max` that returns a struct like `{ "minimum": 0, "maximum": 10} }` then there is no way to say `min_max(x).minimum > 0` using a field reference.  So you end up needing to use the `struct_field` function to create `call("gt", { call("struct_field", { field_ref("x"), "minimum" }), 0 })`.

In many cases (e.g. aggregate keys, join keys, sort keys) Acero will only accept "direct references".  These will always end up being expressible as `FieldRef`.

So, there may be some value in creating a second `DirectReferenceFromProto` that accepts `ReferenceSegment` and returns `FieldRef`.  The implementation could call the existing `FromProto(Expression)`:

```
Result<FieldRef> DirectReferenceFromProto(const substrait::Expression& substrait_expr,
                                          const ExtensionSet& ext_set,
                                          const ConversionOptions& conversion_options) {
  ARROW_ASSIGN_OR_RAISE(compute::Expression expr,
                        FromProto(substrait_expr, ext_set, conversion_options));
  const FieldRef* field_ref = expr.field_ref();
  if (field_ref) {
    return *field_ref;
  } else {
    return Status::Invalid(
        "A direct reference was expected but a more complex expression was given "
        "instead");
  }
}
```

You would still need the change here so I agree with breaking it out into its own method.

cpp/proto/substrait/extension_rels.proto
```diff
@@ -58,3 +58,16 @@ message NamedTapRel {
   // If empty, field names will be automatically generated.
   repeated string columns = 3;
 }
+
+message SegmentedAggregateRel {
+  substrait.RelCommon common = 1;
```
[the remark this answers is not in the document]
westonpace: Agreed.

cpp/proto/substrait/extension_rels.proto
westonpace: I think this is ok.  If this were in the Substrait spec we might separate the two so we could allow the two implementations to diverge in the future.  However, for something internal I think it is fine to keep it simpler.

cpp/src/arrow/compute/exec/source_node.cc
```diff
@@ -102,6 +104,19 @@ struct SourceNode : ExecNode, public TracedNode {
               batch_size = morsel_length;
             }
             ExecBatch batch = morsel.Slice(offset, batch_size);
+            for (auto& value : batch.values) {
+              if (value.is_array()) {
+                ARROW_ASSIGN_OR_RAISE(
+                    value, util::EnsureAlignment(value.make_array(), ipc::kArrowAlignment,
+                                                 default_memory_pool()));
+              }
+              if (value.is_chunked_array()) {
```
westonpace, line 113: `value` can't be a chunked array here.

cpp/src/arrow/engine/substrait/options.cc
```diff
@@ -166,6 +171,57 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
                                      named_tap_rel.name(), std::move(renamed_schema)));
     return RelationInfo{{std::move(decl), std::move(renamed_schema)}, std::nullopt};
   }
+
+  Result<RelationInfo> MakeSegmentedAggregateRel(
+      const ConversionOptions& conv_opts, const std::vector<DeclarationInfo>& inputs,
+      const substrait_ext::SegmentedAggregateRel& seg_agg_rel,
+      const ExtensionSet& ext_set) {
+    if (inputs.size() != 1) {
```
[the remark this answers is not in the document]
westonpace: I think this is no longer a concern right?  Since we are only using the `Measure` part of the original message (and not the part that contains its own inputs).  So we should be able to avoid this `inputs` check.

cpp/src/arrow/engine/substrait/relation_internal.cc
```diff
@@ -293,6 +303,90 @@ Status DiscoverFilesFromDir(const std::shared_ptr<fs::LocalFileSystem>& local_fs
...
+  std::shared_ptr<Schema> aggregate_schema = schema(std::move(output_fields));
+
+  DeclarationInfo aggregate_declaration{
+      compute::Declaration::Sequence(
+          {std::move(input_decl),
+           {"aggregate", compute::AggregateNodeOptions{aggregates, keys, segment_keys}}}),
+      aggregate_schema};
+
+  return ProcessEmit(std::move(agg_common_opt), std::move(aggregate_declaration),
+                     std::move(aggregate_schema));
```
westonpace, line 385: `ProcessEmit` should not be part of this method.  You don't need it when working on an extension method.  This should be left back where it was in the `kAggregate` case of the top-level relation switch statement.  Then you don't need to pass in `RelCommon` and you don't need the changes to `ProcessEmit`.

westonpace (changes requested):
Sorry for the delay, I think there are a few things we can clean up.

cpp/src/arrow/engine/substrait/options.cc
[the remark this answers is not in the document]
rtpsw: I think we still need this check, as the inputs are coming from the caller, not from the `SegmentedAggregateRel` message. We have similar code in the existing `MakeAsOfJoinNodeRel`.

cpp/src/arrow/engine/substrait/relation_internal.cc
rtpsw: Fixed.

cpp/src/arrow/engine/substrait/relation_internal.cc
icexelloss: Yeah I prefer that we remove this namespace.

cpp/proto/substrait/extension_rels.proto
```diff
@@ -58,3 +58,14 @@ message NamedTapRel {
   // If empty, field names will be automatically generated.
   repeated string columns = 3;
 }
+
+message SegmentedAggregateRel {
+  // Grouping keys of the aggregation
+  repeated substrait.Expression.ReferenceSegment grouping_keys = 2;
```
icexelloss, line 64: From looking at existing substrait messages. `FieldReference` is what is commonly used to represent "keys" so I think we should use that type here. @westonpace WDTY?

cpp/proto/substrait/extension_rels.proto
icexelloss: kk sounds good to me.

cpp/proto/substrait/extension_rels.proto
```diff
@@ -58,3 +58,14 @@ message NamedTapRel {
...
+
+  // Segment keys of the aggregation
+  repeated substrait.Expression.ReferenceSegment segment_keys = 3;
```
icexelloss, line 67: Update the index to start from 1?

cpp/src/arrow/engine/substrait/relation_internal.cc
icexelloss: What is the "original one"? I cannot see it from the link.

icexelloss: Nvm I found it.

cpp/src/arrow/engine/substrait/relation_internal.cc
icexelloss: Actually I see what you are doing here - these two methods are not "API" methods in the same sense that the other "FromProto" methods are and takes a lot of internal parameters. 

We can leave as is. But can you put a comment about how the methods here are different from the rest of the methods here. I didn't understand why do meant by "Called sparsely" - but I guess you mean that "these two methods are only meant to be used inside FromProto of AggregationRel and SegementedAggregationRel"?

cpp/proto/substrait/extension_rels.proto
rtpsw: IIRC, Weston's said this is fine in an extension.

cpp/src/arrow/engine/substrait/relation_internal.cc
icexelloss: What is confusing here is that a "internal" in the filename and "internal" in the namespace are two difference things. Here the namespace "internal" means "internals of this internal cc file", which is not obvious. Perhaps we should name this to engine::aggregate to make clear this is internal to aggregate logic.

cpp/src/arrow/engine/substrait/options.cc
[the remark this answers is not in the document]
icexelloss: This is probably OK to keep (minor point)

cpp/proto/substrait/extension_rels.proto
icexelloss: @rtpsw What's the issue with using `FeildReference`? here

rtpsw: There no issue; using `FieldReference` is an alternative. Because `FieldReference` supports non-direct references, such as a reference to multiple fields, the Acero code accepting it would only support a `ReferenceSegment` under it and would fail on anything else. Since we're dealing with a Substrait extension, there's no requirement to use a `FieldReference` or anything, and we can use the exact kind of reference needed, making things simpler.

cpp/proto/substrait/extension_rels.proto
rtpsw: The fix is small but note there's overhead in testing a message-change end-to-end. I fixed it.

icexelloss: ty

cpp/src/arrow/engine/substrait/relation_internal.cc
```diff
@@ -730,64 +810,24 @@ Result<DeclarationInfo> FromProto(const substrait::Rel& rel, const ExtensionSet&
       std::vector<compute::Aggregate> aggregates;
       aggregates.reserve(measure_size);
       // store aggregate fields to be used when output schema is created
-      std::vector<std::vector<int>> agg_src_fieldsets(measure_size);
+      std::vector<std::vector<int>> agg_src_fieldsets;
```
icexelloss, line 813: Why this change?

cpp/src/arrow/engine/substrait/relation_internal.cc
```diff
@@ -730,64 +810,24 @@ Result<DeclarationInfo> FromProto(const substrait::Rel& rel, const ExtensionSet&
...
+      ARROW_ASSIGN_OR_RAISE(
+          auto aggregate_declaration,
+          internal::MakeAggregateDeclaration(
+              std::move(input.declaration), std::move(input_schema), measure_size,
+              std::move(aggregates), std::move(agg_src_fieldsets), std::move(keys),
+              std::move(key_field_ids), {}, {}, ext_set, conversion_options));
 
       return ProcessEmit(std::move(aggregate), std::move(aggregate_declaration),
-                         std::move(aggregate_schema));
+                         aggregate_declaration.output_schema);
```
icexelloss, line 830: Is it valid to access aggregate_declaration.output_schema after std::move(aggregate_declaration? I am unsure how the move works here.

cpp/src/arrow/engine/substrait/relation_internal.h
```diff
@@ -30,6 +30,12 @@
 #include "substrait/algebra.pb.h"  // IWYU pragma: export
 
 namespace arrow {
+namespace compute {
```
icexelloss, line 33: Why do we need this here?

cpp/src/arrow/engine/substrait/expression_internal.h
```diff
@@ -34,6 +35,15 @@ namespace engine {
 
 class SubstraitCall;
 
+ARROW_ENGINE_EXPORT
+Result<FieldRef> DirectReferenceFromProto(const substrait::Expression::ReferenceSegment*,
```
icexelloss, line 39: What is the difference between this and the `FromProto` below?

cpp/src/arrow/engine/substrait/options.cc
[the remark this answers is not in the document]
westonpace: I get it now.  I think, in the past, we had different branches for `kExtensionLeaf`, `kExtensionSingle`, and `kExtensionMulti`.  In my brain this was a `kExtensionSingle` only and so we would verify the number of inputs there.

However, we've merged those all into one path in the extension provider, so we still get `inputs` and need to verify the length.

+1 for leaving this in, sorry for the confusion.

cpp/src/arrow/engine/substrait/expression_internal.h
```diff
@@ -34,6 +35,15 @@ namespace engine {
...
+                                          const ExtensionSet&, const ConversionOptions&);
+
+ARROW_ENGINE_EXPORT
+Result<compute::Expression> FromProto(const substrait::Expression::ReferenceSegment*,
+                                      const ExtensionSet&, const ConversionOptions&,
+                                      std::optional<compute::Expression>);
```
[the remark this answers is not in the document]
westonpace: It's the "current expression".  This method is currently called int he middle of a deserializing an expression tree.  So, for example:

```mermaid
flowchart TD
    A[Call] -->|args| FieldRef
    A -->|args| C
    C[Call*] -->|args| D
    C -->|Call| E
    D[Literal]
    E[FieldRef*]
```

So, when de-referencing `FieldRef` this will be `Call` and when dereferencing `FieldRef*` this will be `Call*`.

However, can we remove this prototype from the header file and put it in an anonymous namespace?  I think it should be an internal method and not exposed.

cpp/src/arrow/engine/substrait/relation_internal.cc
```diff
@@ -293,6 +293,86 @@ Status DiscoverFilesFromDir(const std::shared_ptr<fs::LocalFileSystem>& local_fs
...
+    const ConversionOptions& conversion_options) {
+  FieldVector output_fields;
+  output_fields.reserve(key_field_ids.size() + segment_key_field_ids.size() +
+                        measure_size);
+  // extract aggregate fields to output schema
+  for (const auto& agg_src_fieldset : agg_src_fieldsets) {
+    for (int field : agg_src_fieldset) {
+      output_fields.emplace_back(input_schema->field(field));
+    }
+  }
```
westonpace, line 355: I don't think this is correct.

1. Values come after keys (note, this changed recently in https://github.com/apache/arrow/issues/32897)
2. You're inserting fields here for the inputs to the aggregates.  However, the fields should be based on the outputs.

For example, consider the query is `SELECT SUM(x), COVARIANCE(x,y) GROUP BY key` and the input schema is `{key: int32, x: float32, y: float32}`

I believe this method would return `{x: float32, x: float32, y: float32, key: int32}`.  Instead it should return `{ key: int32, sum(x): float64, covariance(x, y): float64}`.

Note, the names aren't really important (and probably difficult to recreate).  So it'd be fine to do `{key0: int32, measure0: float64, measure1: float64}`.  However, we do need to get the types and # of output fields correct or else it will mess up any project relations added after this.

cpp/src/arrow/engine/substrait/relation_internal.h
```diff
@@ -46,5 +52,46 @@ Result<DeclarationInfo> FromProto(const substrait::Rel&, const ExtensionSet&,
...
+/// \brief Parse an aggregate relation's measure
+///
+/// \param[in] agg_measure the measure
+/// \param[in] ext_set an extension mapping to use in parsing
+/// \param[in] conversion_options options to control how the conversion is done
+/// \param[in] input_schema the schema to which field refs apply
+/// \param[in] is_hash whether the measure is a hash one (i.e., aggregation keys exist)
+/// \param[out] aggregates points to vector to push the parsed measure into
+/// \param[out] agg_src_fieldsets points to vector to push the parsed field set into
+ARROW_ENGINE_EXPORT Status ParseAggregateMeasure(
```
westonpace, line 66: I don't love these methods.  It's not clear why they are public and they have a lot of arguments.  However, I can understand why they are here.

ParseAggregateMeasure is probably inevitable.  Although, I wonder if it might be easier to understand if you return:

```
struct ParsedMeasure {
  compute::Aggregate aggregate;
  std::vector<int> fieldset;
  std::shared_ptr<DataType> output_type;
};

Result<ParsedMeasure> ParseAggregateMeasure(...);
```

I think we could do away with `MakeAggregateDeclaration` if we had a better way of computing the output schema given an input schema and a `compute::AggregateNodeOptions`. I started trying to make a `compute::AggregateNodeOptions::CalculateOutputSchema(const Schema& input_schema)` but it turns out it's rather tricky to determine the output types of measures once we have left Substrait.  So, for the moment, I think we can leave this as-is.

westonpace (changes requested):
I took another look and caught something I had missed previously in the output schema.

cpp/proto/substrait/extension_rels.proto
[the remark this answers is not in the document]
rtpsw: Removed.

cpp/src/arrow/engine/substrait/expression_internal.cc
rtpsw: Fixed.

cpp/src/arrow/engine/substrait/relation_internal.cc
rtpsw: For convenience. The next line reserves the same size. The only difference is that the vector is populated using `push_back` instead of `[i]`.

cpp/src/arrow/engine/substrait/relation_internal.cc
rtpsw: I'll fix this.

cpp/src/arrow/engine/substrait/relation_internal.h
rtpsw: We need a forward declaration of `Aggregate` for the signature of `ParseAggregateMeasure`. However, we no longer need `AggregateNodeOptions`, so I'll remove it.

cpp/src/arrow/engine/substrait/expression_internal.h
rtpsw: The function here returns a `FieldRef` whereas the `FromProto` below returns a `Expression`. This is a fix Weston requested.

cpp/src/arrow/engine/substrait/options.cc
```diff
@@ -165,6 +170,69 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
...
+      return Status::Invalid(
+          "substrait_ext::SegmentedAggregateRel requires a single input but got: ",
+          inputs.size());
+    }
+    if (seg_agg_rel.segment_keys_size() == 0) {
+      return Status::Invalid(
+          "substrait_ext::SegmentedAggregateRel requires at least one segment key");
+    }
```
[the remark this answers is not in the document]
rtpsw: Correct. It's fine that the Substrait producer would need to use the right relation.

cpp/src/arrow/engine/substrait/relation_internal.cc
rtpsw: This is basically refactoring of [original code](https://github.com/apache/arrow/blob/2c9340dd1795bedcc6a43975e26580b4f27912ad/cpp/src/arrow/engine/substrait/relation_internal.cc#L768-L781) (followed by adding similar code for `segment_key_field_ids`). Does that mean the original code is incorrect too? That would be surprising.

cpp/src/arrow/engine/substrait/relation_internal.h
rtpsw: Perhaps something like [the example of `AsofJoinNode::MakeOutputSchema`](https://github.com/apache/arrow/blob/2c9340dd1795bedcc6a43975e26580b4f27912ad/cpp/src/arrow/compute/exec/asof_join_node.cc#L1318-L1322) is what you're looking for?

cpp/src/arrow/engine/substrait/expression_internal.h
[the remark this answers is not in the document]
icexelloss: Thanks Weston for the explanation.

+1 on not exposing this in the header file. I think if we move the MakeSegmentedAggregateRel to expression_internal.cc then we don't need to expose this via the header file. I have a comment about moving these "make extension rel" methods out of options.cc anyway.

cpp/src/arrow/engine/substrait/relation_internal.h
icexelloss: If we need to keep these methods as well, I think we can make them internal methods of relation_internal.cc and remove them from the header.

cpp/src/arrow/engine/substrait/expression_internal.h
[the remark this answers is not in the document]
rtpsw: I don't think `expression_internal.cc`, which deals with Substrait expression parsing, is the right place for any of the `Make*Rel` methods, which deal with Substrait extension handling.

IMO, refactoring of `options.cc` could be reasonable but is outside the scope of this PR.

cpp/src/arrow/engine/substrait/relation_internal.h
icexelloss: I see. I believe if we don't expose these two methods from relation_internal.h then this we don't need this fwd declaration either. See https://github.com/apache/arrow/pull/34627#discussion_r1151928873

cpp/src/arrow/engine/substrait/relation_internal.cc
icexelloss: Actually I think if we remove these two methods from the header I think there is no need to make this internal namespace either.

cpp/src/arrow/engine/substrait/relation_internal.h
rtpsw: I don't see now how we can cleanly avoid exposing the two functions in the "internal"-namespace without putting the function calling them in an inappropriate place. If we can do this cleanly, then I agree.

icexelloss: Sth like this - https://github.com/apache/arrow/commit/e958b6d6cfd8889fd8e5a616eadfe0023cac3b77

I moved all things under relation_internal.cc which I think makes the header file cleaner.

The downside is that the extension logic is now inside relation_internal.cc which makes that file larger, but I still think this is justifiable because "aggregate" and "segment aggregate" does share internal logic here. This is cleaner than exposing the `ParseAggregateMeasure` and `MakeAggregateDeclaration` in `relation_internal.h`.

cpp/src/arrow/engine/substrait/expression_internal.h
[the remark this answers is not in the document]
icexelloss: > However, can we remove this prototype from the header file and put it in an anonymous namespace? I think it should be an internal method and not exposed.

@westonpace Would sth like this look better  to you? https://github.com/apache/arrow/commit/e958b6d6cfd8889fd8e5a616eadfe0023cac3b77

cpp/src/arrow/engine/substrait/options.cc
```diff
@@ -165,6 +170,74 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
...
+          auto parsed_measure,
+          internal::ParseAggregateMeasure(agg_measure, ext_set, conv_opts,
+                                          /*is_hash=*/!keys.empty(), input_schema));
+      aggregates.push_back(std::move(parsed_measure.aggregate));
+      agg_src_fieldsets.push_back(std::move(parsed_measure.fieldset));
+    }
+
+    ARROW_ASSIGN_OR_RAISE(auto decl_info,
+                          internal::MakeAggregateDeclaration(
+                              std::move(inputs[0].declaration), std::move(input_schema),
```
icexelloss, line 227: My C++ knowledge is probably lacking here but do these std::move do anything? It looks like MakeAggregateDeclaration doesn't take a `&&` "move" paramter (rvalue reference parameter) which IIUC is how std::move is supposed to work with.

Similarly I noticed the std::move is used with `EmitProcess` which also doesn't take `&&` move parameter. Perhaps @westonpace can shed some C++ light?

cpp/src/arrow/engine/substrait/relation_internal.h
rtpsw: I understand what you're trying to achieve, however, I think this refactoring of `options.cc` should be out of scope for this PR and in any case should probably be into new files rather than into `relation_internal.cc`, which can't be the right place.

cpp/proto/substrait/extension_rels.proto
rtpsw: Fixed.

cpp/src/arrow/engine/substrait/options.cc
rtpsw: Fixed.

cpp/src/arrow/engine/substrait/relation_internal.cc
icexelloss: @westonpace  I agree this looks strange/incorrect but it does seem to be same as existing code:
https://github.com/apache/arrow/blob/main/cpp/src/arrow/engine/substrait/relation_internal.cc#L771

Do you prefer try to get to bottom of this in this PR or leave as follow up? I am fine either way

icexelloss:
@westonpace I believe the two main remaining items are:

(1) https://github.com/apache/arrow/pull/34627/files#r1151191726

Your comment here about not exposing these methods in the header. I agree with this and made an attempt here:
https://github.com/rtpsw/arrow/pull/3/files

I am able to move those methods internal to `relational_internal.cc` but as a result I also need to move the DefaultExtensionProvider to `relation_internal.cc` because it needs to access those method.

I feel this is a bit cleaner than what is currently in the PR. If you are OK with this I can update the PR with the change. If you want some other way to resolve this / or leave it as is in the PR let me know as well.

(2) https://github.com/apache/arrow/pull/34627#discussion_r1151204240

This is potentially a bug / correctness issue. Given this is not introduced in this PR I am leaning towards debug this as follow up (I actually suspect this could cause some internal issues and asked @rtpsw to test but I also don't want to drag this PR for too long).
If you feel otherwise let me know I can try to get to the bottom of this.

In any case, I would like to move forward with this PR so please let me know your preference of the two issues above or any other  issue that I missed.

cpp/src/arrow/engine/substrait/expression_internal.h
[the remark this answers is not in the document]
westonpace: > @westonpace Would sth like this look better to you? https://github.com/apache/arrow/commit/e958b6d6cfd8889fd8e5a616eadfe0023cac3b77

I'm neutral on the change and don't want to spend too much time bike shedding here.  If you're comfortable let's move forward.

> DirectReferenceFromProto is used by MakeSegmentedAggregateRel in options.cc, so I think it has to be in a header.

Yes, I am fine with `DirectReferenceFromProto`.  I was merely asking about the `FromProto` variant that takes a reference segment.

cpp/src/arrow/engine/substrait/relation_internal.cc
westonpace: We can leave for a follow-up.  Yes, I still think it's incorrect.  Following up an aggregate with a projection had been broken for a long time due to the ordering issue.  So it wouldn't surprise me if this has always been broken.

westonpace (approved):
@icexelloss I agree with your summary.

The header / inclusion issue is minor and we don't need to fix it if there is no obvious solution.  I do think the schema thing is a problem but, given it was already broken, fixing it in a future PR seems fine to me.  We can focus on adding some test cases.  I'll create a github issue.

westonpace:
https://github.com/apache/arrow/issues/34786

icexelloss:
Thanks @westonpace. In that case I will go ahead and merge this. There is definitely room for clean up header/inclusion I would create a follow up to clean this up interest of time.

icexelloss (approved):
LGTM

icexelloss:
https://github.com/apache/arrow/issues/34800

PR #34356, opened 2023-02-26
GH-34330: [Go][Parquet]: Add Extension type support

yevgenypats:
Closes https://github.com/apache/arrow/issues/34330.

This is still WIP as missing some logic around `path_builder.go` but I wanted to open this PR for initial review to make sure Im moving in the right direction.
* Closes: #34330

PR #34631, opened 2023-03-18
GH-34330: [Go][Parquet]: Add Extension type support

### LINKED ISSUE #34330: [Go] Parquet file handling extensions

Hi there! I was trying to use the parquet writer and saw it doesn't support extensions. Im curious if this will be supported or I need to implement some kind of work-around in the meanwhile? Also, if it possible to support it, would love to try and contribute depending on the difficulty or the missing pieces? 

### Component(s)

Go

yevgenypats:
Follow-up instead of https://github.com/apache/arrow/pull/34356
* Closes: #34330

This is not yet complete but I would love some direction on where should I add tests and in what other places I should handle the new extension type.


go/parquet/pqarrow/uuid.go
```diff
@@ -0,0 +1,198 @@
...
+	"strings"
+
+	"github.com/goccy/go-json"
+
+	"github.com/apache/arrow/go/v12/arrow"
+	"github.com/apache/arrow/go/v12/arrow/array"
+	"github.com/google/uuid"
+)
+
+type UUIDBuilder struct {
```
zeroshade, line 16: maybe we should shift the existing UUID extension type out of the `arrow/internal/testing/types` pkg and put it somewhere in the shared `internal` package that both `arrow` and `parquet` packages can use. that way we don't have to duplicate the code.

Not saying you have to, just an idea. I'm also fine with having this separately in here, though since it is only used for testing I'd prefer to keep this to a `_test` package or a `_test.go` file if you don't shift it to the shared internal package. Basically I don't want this to be externally accessible please.

go/parquet/pqarrow/schema_test.go
```diff
@@ -36,9 +36,13 @@ func TestGetOriginSchemaBase64(t *testing.T) {
 	origArrSc := arrow.NewSchema([]arrow.Field{
 		{Name: "f1", Type: arrow.BinaryTypes.String, Metadata: md},
 		{Name: "f2", Type: arrow.PrimitiveTypes.Int64, Metadata: md},
+		{Name: "uuid", Type: pqarrow.NewUUIDType(), Metadata: md},
 	}, nil)
 
 	arrSerializedSc := flight.SerializeSchema(origArrSc, memory.DefaultAllocator)
+	if err := arrow.RegisterExtensionType(pqarrow.NewUUIDType()); err != nil {
```
zeroshade, line 43: don't forget to defer the unregister call. `defer arrow.UnregisterExtensionType("uuid")` it uses the extension name to unregister, so you can either hardcode it or you can do something like:

```go
uuidType := pqarrow.NewUUIDType()
if err := arrow.RegisterExtensionType(uuidType); err != nil {
    t.Fatal(err)
}
defer arrow.UnregisterExtensionType(uuidType.ExtensionName())
```

go/parquet/pqarrow/schema.go
```diff
@@ -354,7 +358,15 @@ func fieldToNode(name string, field arrow.Field, props *parquet.WriterProperties
...
 	case arrow.EXTENSION:
-		return nil, xerrors.New("not implemented yet")
+		return fieldToNode(name, arrow.Field{
+			Name: name,
+			Type: field.Type.(arrow.ExtensionType).StorageType(),
+			Nullable: field.Nullable,
+			Metadata: arrow.MetadataFrom(map[string]string{
+				ExtensionTypeKeyName: field.Type.(arrow.ExtensionType).ExtensionName(),
+			}),
+			// Metadata: ,
```
zeroshade, line 368: what's this comment for?

go/parquet/pqarrow/schema.go
```diff
@@ -32,6 +32,10 @@ import (
 	"golang.org/x/xerrors"
 )
 
+// constants for the extension type metadata keys for the type name and
+// any extension metadata to be passed to deserialize.
+const	ExtensionTypeKeyName = "ARROW:extension:name"
+
```
zeroshade, line 38: you should just use `const ExtensionTypeKeyName = ipc.ExtensionTypeKeyName` or just use `ipc.ExtensionTypeKeyName` directly since it's already exported there. No need to hardcode it here.

go/parquet/pqarrow/schema.go
```diff
@@ -354,7 +358,15 @@ func fieldToNode(name string, field arrow.Field, props *parquet.WriterProperties
...
 			props, arrprops)
 	case arrow.EXTENSION:
-		return nil, xerrors.New("not implemented yet")
+		return fieldToNode(name, arrow.Field{
+			Name: name,
+			Type: field.Type.(arrow.ExtensionType).StorageType(),
+			Nullable: field.Nullable,
+			Metadata: arrow.MetadataFrom(map[string]string{
+				ExtensionTypeKeyName: field.Type.(arrow.ExtensionType).ExtensionName(),
+			}),
```
zeroshade, line 367: You also need to add `ExtensionMetadataKeyName: field.Type.(arrow.ExtensionType).Serialize()`

go/parquet/pqarrow/schema.go
```diff
@@ -948,7 +960,15 @@ func getNestedFactory(origin, inferred arrow.DataType) func(fieldList []arrow.Fi
...
 	nchildren := len(inferred.Children)
 	switch origin.Type.ID() {
-	case arrow.EXTENSION, arrow.SPARSE_UNION, arrow.DENSE_UNION:
+	case arrow.EXTENSION:
+		extType := arrow.GetExtensionType(origin.Type.(arrow.ExtensionType).ExtensionName())
+		if extType == nil {
+			err = xerrors.Errorf("arrow: extension type %q not registered", origin.Type.Name())
+			return
+		}
+		inferred.Field.Type = extType
```
zeroshade, line 969: why `GetExtensionType` using the name when you already have the extension type here via `origin.Type` ?

If the extension type wasn't registered, then when the schema was deserialized you'd only have the storage type anyway and never get into this spot. In addition if the extension type is a parameterized type, then you're losing the metadata that was serialized into the `ExtensionMetadataKey`. 

I think the better solution here would probably be to do something like:

```go
extType := origin.Type.(arrow.ExtensionType)
modified, err = applyOriginalStorageMetadata(arrow.Field{
                 Type: extType.StorageType(), 
                 Metadata: arrow.NewMetadata(
                            []string{ipc.ExtensionTypeKeyName, ExtensionMetadataKeyName},
                            []string{extType.ExtensionName(), extType.Serialize()}),
           }, inferred)
if err != nil {
    return
}

inferred.Field.Type = extType
modified = true
```

Or something to that extent. That way it can properly handle extension types that have children / are defined in terms of a struct/list/etc. If the extension type wasn't registered, that's pretty much what would happen anyways since it would deserialize the schema as the storage type + the metadata.

What do you think? Might also be worthwhile to add a testing extension type that is parameterized and whose storage type is one of the nested types just to ensure it is handled correctly. You can find examples of parameteric types in the `arrow/internal/testing/types/extension_types.go` file.

go/parquet/pqarrow/schema_test.go
```diff
@@ -36,9 +36,13 @@ func TestGetOriginSchemaBase64(t *testing.T) {
 	origArrSc := arrow.NewSchema([]arrow.Field{
 		{Name: "f1", Type: arrow.BinaryTypes.String, Metadata: md},
 		{Name: "f2", Type: arrow.PrimitiveTypes.Int64, Metadata: md},
+		{Name: "uuid", Type: pqarrow.NewUUIDType(), Metadata: md},
```
zeroshade, line 39: in addition to the schema `ToParquet`/`FromParquet` functions. Please add some tests for reading and writing a table / record batches which contain extension types. Such as a simple round trip:

```go
expectedTable := // ... table with extension type column
defer expectedTable.Release()

var buf bytes.Buffer
require.NoError(t, pqarrow.WriteTable(expectedTable, &buf, ....))

actualTable, err := pqarrow.ReadTable(context.Background(), bytes.NewReader(buf.Bytes()), .....)
require.NoError(t, err)
defer actualTable.Release()

assert.True(t, array.TableEqual(expectedTable, actualTable), ...)
```

zeroshade (changes requested):

zeroshade:
@yevgenypats you're missing the Apache license heading on the `uuid.go` file, that's why the linter job failed.

go/internal/types/uuid.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/goccy/go-json"
+
+	"github.com/apache/arrow/go/v12/arrow"
+	"github.com/apache/arrow/go/v12/arrow/array"
+	"github.com/google/uuid"
+)
+
+type UUIDBuilder struct {
+	*array.ExtensionBuilder
+}
```
zeroshade, line 18: instead of this being duplicated now in `arrow/internal/testing/types/` and here, could we remove it from `arrow/internal/testing/types/` and anywhere that references `UUIDType` in the arrow pkg also points to this spot? That way both are sharing this implementation

go/parquet/pqarrow/schema.go
```diff
@@ -948,7 +957,21 @@ func getNestedFactory(origin, inferred arrow.DataType) func(fieldList []arrow.Fi
...
+		extType := origin.Type.(arrow.ExtensionType)
+		modified, err = applyOriginalStorageMetadata(arrow.Field{
+										 Type: extType.StorageType(), 
+										 Metadata: arrow.NewMetadata(
+																[]string{ipc.ExtensionTypeKeyName, ipc.ExtensionMetadataKeyName},
+																[]string{extType.ExtensionName(), extType.Serialize()}),
+							 }, inferred)
+		if err != nil {
+				return
+		}
```
zeroshade, line 970: can you run `gofmt` please to fix the spacing here?

go/parquet/pqarrow/schema_test.go
```diff
@@ -32,13 +34,20 @@ import (
 )
 
 func TestGetOriginSchemaBase64(t *testing.T) {
+	uuidType := types.NewUUIDType()
 	md := arrow.NewMetadata([]string{"PARQUET:field_id"}, []string{"-1"})
+  extMd := arrow.NewMetadata([]string{ipc.ExtensionMetadataKeyName, ipc.ExtensionTypeKeyName, "PARQUET:field_id"}, []string{ uuidType.Serialize(), uuidType.ExtensionName(), "-1"})
```
zeroshade, line 39: same thing, please run gofmt and fix the spacing here.

go/parquet/pqarrow/encode_arrow_test.go
```diff
@@ -1526,6 +1528,7 @@ func TestWriteTableMemoryAllocation(t *testing.T) {
 			arrow.Field{Name: "i64", Type: arrow.PrimitiveTypes.Int64, Nullable: true},
 			arrow.Field{Name: "f64", Type: arrow.PrimitiveTypes.Float64, Nullable: true})},
 		{Name: "arr_i64", Type: arrow.ListOf(arrow.PrimitiveTypes.Int64)},
+		{Name: "uuid", Type: types.NewUUIDType(), Nullable: true},
```
zeroshade, line 1531: please add this to one of the round trip tests (this test doesn't perform a read) so that we have full testing for reading *and* writing the extension type along with confirmation that the resulting Array we read matches what we wrote.

go/parquet/pqarrow/schema.go
hermanschaaf: Fixed

go/arrow/table.go
```diff
@@ -42,20 +43,19 @@ type Table interface {
...
-//				arr := c.(*array.Int32)
-//				// do something with arr
-//			}
-//		case arrow.INT64:
-//			for _, c := range column.Data().Chunks() {
-//				arr := c.(*array.Int64)
-//				// do something with arr
-//			}
-//		case ...
+//	switch column.DataType().ID {
```
disq, line 46: This block was formatted 'badly' according to `goimports`?

disq:
@zeroshade Hi, I've done some things to the project, probably incorrectly. Might need some pointers when you have a chance to take a look.

go/parquet/pqarrow/encode_arrow_test.go
disq: Added as of [these commits](https://github.com/apache/arrow/pull/34631/files/517ace50760cad5749a3c1e927372c90e6fd5b85..05295d3a8ce9d6f619163d0217bdeccd37aaab86)

zeroshade:
@disq I'll take a look early next week. I've been away at a conference this week so I haven't been able to check and respond here. But I should be able to look this over on monday or tuesday.

Sorry for the delay.

go/arrow/table.go
```diff
@@ -69,7 +69,7 @@ type Column struct {
 // of the ref counting.
 func NewColumnFromArr(field Field, arr Array) Column {
 	if !TypeEqual(field.Type, arr.DataType()) {
-		panic("arrow/array: inconsistent data type")
+		panic(fmt.Sprintf("arrow/array: inconsistent data type %s vs %s", field.Type.String(), arr.DataType().String()))
```
disq, line 72: I think having more data in these types of errors is helpful because they don't seem to be day to day errors but errors occurring during development, and having the types in the message removes the 'obvious second step' (ie. running in debugger to investigate) and transforms it into a more actionable, clearer error. There's one more below in L101, also one in the parquet test suite. Easy to revert.

go/arrow/compare.go
```diff
@@ -129,3 +129,15 @@ func TypeEqual(left, right DataType, opts ...TypeEqualOption) bool {
...
+// optionally checking metadata equality for STRUCT types by passing to TypeEqual
+func StorageTypeEqual(left, right DataType, opts ...TypeEqualOption) bool {
+	if leftExt, ok := left.(ExtensionType); ok {
+		left = leftExt.StorageType()
+	}
+	if rightExt, ok := right.(ExtensionType); ok {
+		right = rightExt.StorageType()
+	}
+	return TypeEqual(left, right, opts...)
+}
```
zeroshade, line 143: might make sense to simply implement this as the default implementation of `ExtensionEquals` in `ExtensionBase` that other extension types can implement or otherwise. Honestly I don't see much benefit to having this function over letting consumers just call `TypeEqual` themselves and having the extension types implement `ExtensionEquals` appropriately.

go/arrow/table.go
zeroshade: interesting. *shrug* nothing wrong with this modification honestly.

go/arrow/table.go
zeroshade: I can agree that adding more information in the errors is good. It might make sense to use:

`fmt.Errorf("%w: arrow/array: inconsistent data type %s vs %s", arrow.ErrInvalid, field.Type, arr.DataType)` instead of the `fmt.Sprintf`. You also don't need to call `.String()` manually, since you're using `%s` it'll use the `.String()` method automatically when printing it. So the code can look cleaner.

zeroshade: same goes for the other spots.

go/arrow/table.go
```diff
@@ -96,9 +96,9 @@ func NewColumn(field Field, chunks *Chunked) *Column {
 	}
 	col.data.Retain()
 
-	if !TypeEqual(col.data.DataType(), col.field.Type) {
+	if !StorageTypeEqual(col.data.DataType(), col.field.Type) {
```
zeroshade, line 99: this change doesn't make any sense to me. `TypeEqual` should already end up checking that the storage types are equal by virtue of the extension type itself implementing the equals check.

go/internal/types/extension_types.go
```diff
@@ -209,13 +209,13 @@ func (UUIDType) ExtensionName() string { return "uuid" }
...
-// "uuid-serialized" in order to correctly create a UuidType for testing deserialize.
+// "uuid-serialized" in order to correctly create a UUIDType for testing deserialize.
 func (UUIDType) Deserialize(storageType arrow.DataType, data string) (arrow.ExtensionType, error) {
 	if string(data) != "uuid-serialized" {
 		return nil, fmt.Errorf("type identifier did not match: '%s'", string(data))
 	}
-	if !arrow.TypeEqual(storageType, &arrow.FixedSizeBinaryType{ByteWidth: 16}) {
-		return nil, fmt.Errorf("invalid storage type for UuidType: %s", storageType.Name())
+	if !arrow.StorageTypeEqual(storageType, &arrow.FixedSizeBinaryType{ByteWidth: 16}) {
+		return nil, fmt.Errorf("invalid storage type for UUIDType: %s", storageType.Name())
```
zeroshade, line 218: this change doesn't make any sense to me. The `StorageType` for `UUIDType` is `&arrow.FixedSizeBinaryType{ByteWidth: 16}`. You wouldn't want the underlying storage type to be a different extension type. What's the use case where this would be correct over just using `!arrow.TypeEqual`?

go/parquet/pqarrow/column_readers.go
```diff
@@ -459,8 +459,13 @@ func chunksToSingle(chunked *arrow.Chunked) (arrow.ArrayData, error) {
 
 // create a chunked arrow array from the raw record data
 func transferColumnData(rdr file.RecordReader, valueType arrow.DataType, descr *schema.Column, mem memory.Allocator) (*arrow.Chunked, error) {
+	valueID := valueType.ID()
+	if valueID == arrow.EXTENSION {
+		valueID = valueType.(arrow.ExtensionType).StorageType().ID()
+	}
```
zeroshade, line 465: you probably want to do:

```go
dt := valueType
if valueType.ID() == arrow.EXTENSION {
    dt = valueType.(arrow.ExtensionType).StorageType()
}

// ...
switch dt.ID() {
...
}
```

and then use `dt` everywhere `valueType` is used except for when arrays and such are created.  This is because, for example, if you look at `transferDictionary` it assumes that the data type passed is a `DictionaryType`, which would be incorrect if `valueType` was an extension type whose underlying storage type was a dictionary. You'd want to pass `dt` to `transferDictionary` and then update the returned `Chunked` array's data type to be the extension type (or something like that....).

go/parquet/pqarrow/column_readers.go
```diff
@@ -546,6 +551,14 @@ func transferBinary(rdr file.RecordReader, dt arrow.DataType) *arrow.Chunked {
 			defer chunks[idx].Data().Release()
 			defer chunks[idx].Release()
```
zeroshade, line 552: this seems wrong... it should probably be:

```go
prev := chunks[idx]
chunks[idx] = array.MakeFromData(chunks[idx].Data())
prev.Release()
defer chunks[idx].Release()
```

go/parquet/pqarrow/column_readers.go
```diff
@@ -546,6 +551,14 @@ func transferBinary(rdr file.RecordReader, dt arrow.DataType) *arrow.Chunked {
...
 		}
+	} else if dt.ID() == arrow.EXTENSION && len(chunks) > 0 && arrow.StorageTypeEqual(chunks[0].DataType(), dt) && !arrow.TypeEqual(chunks[0].DataType(), dt) {
```
zeroshade, line 554: this is kinda what i was touching on above. Because you're passing the extension data type down here rather than the expected underlying storage, this becomes a bit more complicated.

I think it might be preferable to do something like:

```go
chunks := brdr.GetBuilderChunks()
storage := dt
if dt.ID() == arrow.EXTENSION {
    storage = dt.(arrow.ExtensionType).StorageType()
}

if storage == arrow.BinaryTypes.String || storage == arrow.BinaryTypes.LargeString {
    // if storage != dt -> do the loop with `NewExtensionArrayWithStorage` 
    // otherwise the look with `MakeFromData`
}
return arrow.NewChunked(dt, chunks)
```

go/parquet/pqarrow/column_readers.go
```diff
@@ -546,6 +551,14 @@ func transferBinary(rdr file.RecordReader, dt arrow.DataType) *arrow.Chunked {
...
+		// convert chunks from the underlying storage type to extension type without copying data
+		etype := dt.(arrow.ExtensionType)
+
+		for idx := range chunks {
+			chunks[idx] = array.NewExtensionArrayWithStorage(etype, chunks[idx])
+			defer chunks[idx].Release()
+		}
```
zeroshade, line 561: probably should be:

```go
for idx, chk := range chunks {
    chunks[idx] = array.NewExtensionArrayWithStorage(etype, chk)
    chk.Release() // NewExtensionArrayWithStorage will call retain on chk, so it still needs to be released
    defer chunks[idx].Release()
}
```

go/parquet/pqarrow/encode_arrow.go
```diff
@@ -449,7 +457,7 @@ func writeDenseArrow(ctx *arrowWriteContext, cw file.ColumnChunkWriter, leafArr
 				data[i] = parquet.ByteArray(valueBuf[offsets[i]:offsets[i+1]])
 			}
 		default:
-			return xerrors.New(fmt.Sprintf("invalid column type to write to ByteArray: %s", leafArr.DataType().Name()))
+			return xerrors.Errorf("invalid column type to write to ByteArray: %s", leafArr.DataType().Name())
```
zeroshade, line 460: if we're in here changing this, might as well use `fmt.Errorf` as `xerrors` is no longer needed and should be phased out. do something like:

`return fmt.Errorf("%w: expected column type %s to be compatible with `ByteArray`", arrow.ErrInvalid, leafArr.DataType())`

go/parquet/pqarrow/encode_arrow_test.go
```diff
@@ -1517,6 +1531,24 @@ func (ps *ParquetIOTestSuite) TestArrowMapTypeRoundTrip() {
 	ps.roundTripTable(tbl, true)
 }
 
+func (ps *ParquetIOTestSuite) TestArrowExtensionTypeRoundTrip() {
+	var testUUID = uuid.New()
```
zeroshade, line 1535: use the `memory.NewCheckedAllocator` pattern so you can `defer mem.AssertSize(ps.T(), 0)` and make sure there's no leaks in the code.

go/parquet/pqarrow/path_builder.go
```diff
@@ -440,7 +440,9 @@ func (p *pathBuilder) Visit(arr arrow.Array) error {
 		}
 		return nil
 	case arrow.EXTENSION:
-		return xerrors.New("extension types not implemented yet")
+		// TODO: (hermanschaaf) handle nested extension types
```
zeroshade, line 443: it should be feasible to just do `return p.Visit(arr.(arrow.ExtensionArray).Storage())` and that would handle nested extension types for you just fine.

zeroshade:
@disq I've gone through and left a bunch of comments, feel free to let me know if anything is unclear or if you need help implementing anything.

go/arrow/compare.go
disq: This is different than `ExtensionEquals`, as that one compares the underlying storage types and doesn't require both types to be extensions. In our case one of the two inputs is `FixedSizeBinaryType` and the is `UUIDType` which is an extension type. It's now only used in `UUIDType.deserialize()` so I've unexported and moved it to `internal/types`.

go/internal/types/extension_types.go
```diff
@@ -209,13 +209,13 @@ func (UUIDType) ExtensionName() string { return "uuid" }
...
 // Deserialize expects storageType to be FixedSizeBinaryType{ByteWidth: 16} and the data to be
-// "uuid-serialized" in order to correctly create a UuidType for testing deserialize.
+// "uuid-serialized" in order to correctly create a UUIDType for testing deserialize.
 func (UUIDType) Deserialize(storageType arrow.DataType, data string) (arrow.ExtensionType, error) {
 	if string(data) != "uuid-serialized" {
 		return nil, fmt.Errorf("type identifier did not match: '%s'", string(data))
 	}
-	if !arrow.TypeEqual(storageType, &arrow.FixedSizeBinaryType{ByteWidth: 16}) {
-		return nil, fmt.Errorf("invalid storage type for UuidType: %s", storageType.Name())
+	if !storageTypeEqual(storageType, &arrow.FixedSizeBinaryType{ByteWidth: 16}) {
```
disq, line 217: The `storageType` passed in here is an ExtensionType so `arrow.TypeEqual` doesn't fit here

go/parquet/pqarrow/column_readers.go
disq: The resulting code seems harder to follow as both cases are now somewhat intertwined.

go/arrow/table.go
disq: This wasn't required and is removed, it's back to `TypeEqual` now.

go/internal/types/extension_types.go
disq: As I mentioned in the comment about the StorageEqual change, `storageType` passed in here is apparently an `ExtensionType`, so `TypeEqual()` doesn't work (as `FixedSizeBinaryType` isn't an extension type... which can be remedied by creating an extension on top of that type). We could also unrolling the `storageEqual()` into here as this is the only call site for that func, but I chose to leave it separate, thinking it could be useful for other test types in the future.

go/parquet/pqarrow/encode_arrow_test.go
disq: I've added checkedallocators everywhere in that test file and looks like everything's a bit leaky. Investigating I could probably use some pointers.

go/arrow/array/table_test.go
```diff
@@ -157,7 +157,13 @@ func TestChunkedInvalid(t *testing.T) {
...
 			t.Fatalf("expected a panic")
 		}
-		if got, want := e.(string), "arrow/array: mismatch data type float64 vs int32"; got != want {
+		errStr, ok := e.(string)
+		if !ok {
+			err := e.(error)
+			errStr = err.Error()
+		}
+
+		if got, want := errStr, fmt.Sprintf("%s: arrow/array: mismatch data type float64 vs int32", arrow.ErrInvalid); got != want {
```
zeroshade, line 166: since we're expecting an error, this should probably be:

```go
err, ok := e.(error)
if !ok {
    t.Fatalf("expected an error")
}

if !errors.Is(err, arrow.ErrInvalid) {
    t.Fatalf("should be an ErrInvalid")
}

if got, want := err.Error(), fmt.Sprintf(........); got != want { ...
```

go/internal/types/extension_types.go
zeroshade: what's the use case where an `ExtensionType` gets passed in here? If an `ExtensionType` is being passed in here, that means something is calling `Deserialize` incorrectly.

go/parquet/pqarrow/column_readers.go
```diff
@@ -543,21 +543,26 @@ func transferBinary(rdr file.RecordReader, dt arrow.DataType) *arrow.Chunked {
...
-		etype := dt.(arrow.ExtensionType)
-
-		for idx := range chunks {
-			chunks[idx] = array.NewExtensionArrayWithStorage(etype, chunks[idx])
-			defer chunks[idx].Release()
+	storage := dt
+	if dt.ID() == arrow.EXTENSION {
+		storage = dt.(arrow.ExtensionType).StorageType()
+	}
+	if storage == arrow.BinaryTypes.String || storage == arrow.BinaryTypes.LargeString || dt.ID() == arrow.EXTENSION {
```
zeroshade, line 550: why is the `dt.ID() == arrow.EXTENSION` needed?

go/parquet/pqarrow/encode_arrow_test.go
zeroshade: you can muck with the env vars `ARROW_CHECKED_ALLOC_FRAMES` and `ARROW_CHECKED_REALLOC_FRAMES` (which default to `4` and `3` respectively) when running the tests. They control how many stack frames up are used to print out the location of the allocation that leaked. By increasing those you can pinpoint which allocation of memory is leaking which can help identify the leak.

You can also step through with the debugger and watch the ref counts of the record/array/data objects respectively to ensure the ref counting goes up and down appropriately.

Most likely there's a case where a new array is being created, or a new ChunkedArray is being created and the original sources aren't getting released (since `NewData`, `NewChunked`, and `MakeFromData` all call retain on their arguments to ensure they don't get wiped out from below them).

go/parquet/pqarrow/column_readers.go
disq: `UUIDType`'s storage is `FixedSizeBinaryType` so there's no way it's going inside that block without adding that check. And without it, `TestArrowExtensionTypeRoundTrip`'s `roundTripTable` checks fail with:

> panic: invalid: arrow/array: mismatch data type fixed_size_binary[16] vs extension_type<storage=fixed_size_binary[16]>

go/parquet/pqarrow/encode_arrow_test.go
```diff
@@ -917,17 +963,20 @@ var fullTypeList = []arrow.DataType{
...
+			ps.readAndCheckSingleColumnFile(mem, data, values)
 		})
 	}
 }
 
-func (ps *ParquetIOTestSuite) roundTripTable(expected arrow.Table, storeSchema bool) {
-	mem := memory.NewCheckedAllocator(memory.NewGoAllocator())
-	defer mem.AssertSize(ps.T(), 0)
+func (ps *ParquetIOTestSuite) roundTripTable(mem memory.Allocator, expected arrow.Table, storeSchema bool) {
+	mem = memory.NewGoAllocator() // FIXME currently overriding allocator to isolate leaks between roundTripTable and caller
```
disq, line 979: Maybe having CheckedAllocator everywhere wasn't a good idea

go/parquet/pqarrow/encode_arrow_test.go
```diff
@@ -126,7 +126,7 @@ func makeDateTimeTypesTable(mem memory.Allocator, expected bool, addFieldMeta bo
 }
 
 func TestWriteArrowCols(t *testing.T) {
-	mem := memory.NewCheckedAllocator(memory.DefaultAllocator)
+	mem := memory.NewCheckedAllocator(memory.NewGoAllocator())
```
zeroshade, line 129: why the switch to `NewGoAllocator` over `DefaultAllocator` ?

go/parquet/pqarrow/column_readers.go
zeroshade: Interesting, I guess I completely forgot about the `fixed size binary` handling when i made my suggestion. Hmm. you might be able to pull this out into a switch then to make it a bit cleaner:

```go
switch {
case dt.ID() == arrow.EXTENSION:
        etype := dt.(arrow.ExtensionType)
        for idx, chk := range chunks {
	        chunks[idx] = array.NewExtensionArrayWithStorage(etype, chk)
                chk.Release() // NewExtensionArrayWithStorage will call retain on chk, so it still needs to be released
                defer chunks[idx].Release()
        }
case dt == arrow.BinaryTypes.String || dt == arrow.BinaryTypes.LargeString:
        for idx := range chunks {
                prev := chunks[idx]
                chunks[idx] = array.MakeFromData(chunks[idx].Data())
                prev.Release()
                defer chunks[idx].Release()
        }
}
```

what do you think?

zeroshade:
@disq looks like you've still got a leak somewhere causing the tests to fail

go/parquet/pqarrow/encode_arrow_test.go
disq: I just wanted to _make sure_ while whacking ~moles~ leaks, will test and switch to the `DefaultAllocator ` when it's all over :)

disq:
> @disq looks like you've still got a leak somewhere causing the tests to fail

Well the project (mostly the tests, but not all) has the leaks now, the extension stuff doesn't seem to have any... but I went out of scope and started hunting those down, still here after two days of debugging and trying things. `ParquetIOTestSuite.roundTripTable` causes more leaks but the assertion is currently disabled so they aren't popping up at the moment.

zeroshade:
@disq ah gotcha. Fair enough, if you want I'm fine with filing a new issue to hunt down those memory leaks and re-enable the assertion, and then leaving it disabled in that test for now. That way the tests can still succeed but we have an issue to track fixing the problem.

go/parquet/internal/encoding/types.go
```diff
@@ -339,6 +339,12 @@ func (b *BufferWriter) Finish() *memory.Buffer {
 	return buf
 }
 
+// Release the underlying buffer and not allocate anything else. To re-use this buffer, Reset() or Finish() should be called
+func (b *BufferWriter) Release() {
```
disq, line 343: This had to be added, otherwise there's no way of completely "releasing the memory" without "allocating" a new buffer. (memory management related quotes used ironically)

go/parquet/pqarrow/encode_arrow_test.go
```diff
@@ -136,7 +138,8 @@ func TestWriteArrowCols(t *testing.T) {
 	manifest, err := pqarrow.NewSchemaManifest(psc, nil, nil)
 	require.NoError(t, err)
 
-	sink := encoding.NewBufferWriter(0, memory.DefaultAllocator)
+	sink := encoding.NewBufferWriter(0, mem)
+	defer sink.Release()
```
disq, line 142: Example use of the `BufferWriter.Release()` mentioned above

go/parquet/pqarrow/file_reader.go
```diff
@@ -356,11 +356,9 @@ func (fr *FileReader) ReadRowGroups(ctx context.Context, indices, rowGroups []in
 
 	// pass pairs of reader and column index to the channel for the
 	// goroutines to read the data
-	for idx, r := range readers {
-		defer func(r *ColumnReader) {
-			r.Release()
-		}(r)
-		ch <- readerInfo{r, idx}
+	for idx := range readers {
+		defer readers[idx].Release()
```
disq, line 360: Cleaned up a little here, shorter and easier to understand, without a shadowing-closure for the defer, and no confusion as of to the value of `r` w.r.t. `range`.

go/parquet/pqarrow/path_builder.go
```diff
@@ -440,7 +440,8 @@ func (p *pathBuilder) Visit(arr arrow.Array) error {
 		}
 		return nil
 	case arrow.EXTENSION:
-		return xerrors.New("extension types not implemented yet")
+		// TODO this needs a test
```
disq, line 443: still TODO

go/parquet/pqarrow/path_builder.go
disq: The test is finally added

go/parquet/pqarrow/encode_arrow_test.go
```diff
@@ -917,17 +959,21 @@ var fullTypeList = []arrow.DataType{
...
 		})
 	}
 }
 
-func (ps *ParquetIOTestSuite) roundTripTable(expected arrow.Table, storeSchema bool) {
-	mem := memory.NewCheckedAllocator(memory.NewGoAllocator())
-	defer mem.AssertSize(ps.T(), 0)
+func (ps *ParquetIOTestSuite) roundTripTable(_ memory.Allocator, expected arrow.Table, storeSchema bool) {
+	mem := memory.NewCheckedAllocator(memory.DefaultAllocator) // FIXME: currently overriding allocator to isolate leaks between roundTripTable and caller
+	//defer mem.AssertSize(ps.T(), 0)                            // FIXME: known leak
```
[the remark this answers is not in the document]
zeroshade: Once you're satisfied (or frustrated enough lol) can you file an issue that references this PR for fixing these memory leaks? i'm fine with leaving fixing these memory leaks for another PR, I just want to have an issue filed to track it so we don't forget about it.

Once the issue is filed and the CI is passing, i'm good with merging this.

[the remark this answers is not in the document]
disq: Filed the issue https://github.com/apache/arrow/issues/35015

go/parquet/pqarrow/schema.go
```diff
@@ -1043,6 +1066,26 @@ func applyOriginalStorageMetadata(origin arrow.Field, inferred *SchemaField) (mo
...
+					if ext != nil {
+						dd, err2 := ext.Deserialize(inferred.Field.Type, extMeta)
+						if err2 != nil {
+							err = err2
+							return
+						}
+						inferred.Field.Type = dd
+					} else {
+						// We'll be here if the extension isn't registered.
+						// Optional? maybe skip this error and leave the type as is?
```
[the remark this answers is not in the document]
zeroshade: yea, I agree. We shouldn't error in this case, just let it pass through and maintain the metadata for the field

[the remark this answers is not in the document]
disq: @zeroshade OK we don't error any more, also added a test to verify this case. Tests should hopefully pass now.

disq:
@zeroshade Is it possible to get a final review, I think I've managed to get all outstanding issues fixed, let me know if there's anything else I can do to help go through with the approval/merge. AMD64 Debian11 1.18 CGO test is getting cancelled after 15m, maybe timing out?

go/internal/types/extension_types.go
zeroshade: I'm still a little confused here, what's the situation (or string of events) where Deserialize is being called on something that *is already* an extension type? Shouldn't `Deserialize` only get called if the type isn't yet an extension type and we have metadata saying it should be?

zeroshade (commented):
I only have one outstanding comment/question left.

disq:
> I only have one outstanding comment/question left.

@zeroshade I can't seem to comment on your comment on the line so quoting it here

> I'm still a little confused here, what's the situation (or string of events) where Deserialize is being called on something that is already an extension type? Shouldn't `Deserialize` only get called if the type isn't yet an extension type and we have metadata saying it should be?

The original call sites for `ext.Deserialize` seems to be in `arrjson/arrjson.go:FieldWrapper.UnmarshalJSON()` and `ipc/metadata.go:typeFromFB`, which existed before this PR. This PR adds a third call site, in `pqarrow/schema.go:applyOriginalStorageMetadata()`, which seem to do the same thing. In summary, they all work like:

- We have some data in a known storage type, and an extension name, `extName` in metadata.
- Call `arrow.GetExtensionType(extName)`.
- If the ext is registered this will get an ext object with the correct (extension) type. (if not, all call sites keep the storage type as-is)
- Then call this newly created ext object's `Deserialize` method to get data into the object.

zeroshade:
@disq Ah that explains it!

We're basically decoding the type twice in the current implementation:

Over [here](https://github.com/apache/arrow/pull/34631/files#diff-0ca8ee00b3bab4abf032cc92b9053fc94f528c103fdb05caabffe58a8e849152R960): we're grabbing the origin type that is an extension type and we *add* the extension type metadata to the field.

Then over [here](https://github.com/apache/arrow/pull/34631/files#diff-0ca8ee00b3bab4abf032cc92b9053fc94f528c103fdb05caabffe58a8e849152R1070): we look for that metadata and recreate the extension type by calling deserialize, but we use `inferred.Field.Type`, which we just set equal to the extension type up above.

Thinking about this a bit more, I realize that the only way you end up in the initial case to begin with is if we already successfully deserialized the extension type when we called `flight.DeserializeSchema` in the first place (inside of `getOriginSchema`). So, we actually don't need the check for the metadata and manual call to `Deserialize` in `applyOriginalStorageMetadata`. We can just do a quick check that the storage type of `extType` matches `inferred.Field.Type` and then we're all good.

I've created a commit for the PR which removes the need for storageTypeEquals and adds a test to ensure that if the extension type is *not* currently registered, we get the storage type + extension metadata when we convert back to an Arrow Schema. But I think you need to check the box to allow maintainers to push to your fork. Alternately, I could just share the diff with you and you can add it yourself. let me know what you think

disq:
> @disq Ah that explains it!

I've pinged @yevgenypats for the access but in the meantime you can just send the diff here to get things moving :)

disq:
> @disq [zeroshade@df72318](https://github.com/zeroshade/arrow/commit/df72318346051cc684488d6f7eb075949188002a) that should be it :)

@zeroshade Committed as-is, thanks! (hopefully no last minute fmt issues now)

zeroshade (approved):

PR #34632, opened 2023-03-18
GH-34536: [Parquet][C++] Overwrite default config for DeltaBitPackEncoder

### LINKED ISSUE #34536: [C++][Parquet] Benchmark and maybe Override DeltaBitPackEncoder Defaults

Parquet C++ DELTA_BINARY_PACKED uses:

* Block size - the number of values per block (default: 128)
* Miniblock count - the number of miniblocks per block (default: 4)

As mentioned in https://github.com/apache/arrow-rs/issues/2282 . We may benchmark it and make it suitable for different bitwidth.

### Component(s)

C++, Parquet

mapleFU:
### Rationale for this change

Change default DeltaBitPackEncoder BlockSize from 32 to 64.

Tiny block size change, and an trivial optimization.

### Are these changes tested?

No

No

* Closes: #34536


mapleFU:
Benchmark data is listed in issue: https://github.com/apache/arrow/issues/34536
However, I didn't test all distributions. There must be some cases 128 is faster than 256. But I think we can change the default

mapleFU:
cc @rok

cpp/src/parquet/encoding.cc
```diff
@@ -2100,9 +2100,12 @@ class DictByteArrayDecoderImpl : public DictDecoderImpl<ByteArrayType>,
 
 template <typename DType>
 class DeltaBitPackEncoder : public EncoderImpl, virtual public TypedEncoder<DType> {
+  static_assert(std::is_same_v<int32_t, typename DType::c_type> ||
+                std::is_same_v<int64_t, typename DType::c_type>);
   // Maximum possible header size
   static constexpr uint32_t kMaxPageHeaderWriterSize = 32;
```
wgtmac, line 2106: It would be better to make them configurable. Either at writer level (via `WriterProperties`) or column level (via `ColumnProperties`). Otherwise user does not have a way to tune performance or mitigate regression at the run time.

wgtmac: Yes, something like `column_properties->set_delta_bit_pack_params(a, b)` ?

rok: Would we want to expose this in other language bindings too? (e.g. Python, R)

rok:
Do we need integration tests with other languages (e.g. Java) to make sure to make sure they can decode different block sizes?

cpp/src/parquet/encoding.cc
mapleFU: It's a bit hacking here. Let me have a try...

rok: Perhaps that would be better as a separate issue.

mapleFU:
Seems Java has a config, but use would not change it:

```java
/**
 * Config for delta binary packing
 */
class DeltaBinaryPackingConfig {
  final int blockSizeInValues;
  final int miniBlockNumInABlock;
  final int miniBlockSizeInValues;

  public DeltaBinaryPackingConfig(int blockSizeInValues, int miniBlockNumInABlock) {
    this.blockSizeInValues = blockSizeInValues;
    this.miniBlockNumInABlock = miniBlockNumInABlock;
    double miniSize = (double) blockSizeInValues / miniBlockNumInABlock;
    Preconditions.checkArgument(miniSize % 8 == 0, "miniBlockSize must be multiple of 8, but it's " + miniSize);
    this.miniBlockSizeInValues = (int) miniSize;
  }

  public static DeltaBinaryPackingConfig readConfig(InputStream in) throws IOException {
    return new DeltaBinaryPackingConfig(BytesUtils.readUnsignedVarInt(in),
            BytesUtils.readUnsignedVarInt(in));
  }

  public BytesInput toBytesInput() {
    return BytesInput.concat(
            BytesInput.fromUnsignedVarInt(blockSizeInValues),
            BytesInput.fromUnsignedVarInt(miniBlockNumInABlock));
  }
}
```

arrow-rs use fixed size writer:

```rust
impl<T: DataType> DeltaBitPackEncoder<T> {
    /// Creates new delta bit packed encoder.
    pub fn new() -> Self {
        Self::assert_supported_type();

        // Size miniblocks so that they can be efficiently decoded
        let mini_block_size = match T::T::PHYSICAL_TYPE {
            Type::INT32 => 32,
            Type::INT64 => 64,
            _ => unreachable!(),
        };

        let num_mini_blocks = DEFAULT_NUM_MINI_BLOCKS;
        let block_size = mini_block_size * num_mini_blocks;
        assert_eq!(block_size % 128, 0);

        DeltaBitPackEncoder {
            page_header_writer: BitWriter::new(MAX_PAGE_HEADER_WRITER_SIZE),
            bit_writer: BitWriter::new(MAX_BIT_WRITER_SIZE),
            total_values: 0,
            first_value: 0,
            current_value: 0, // current value to keep adding deltas
            block_size,       // can write fewer values than block size for last block
            mini_block_size,
            num_mini_blocks,
            values_in_block: 0, // will be at most block_size
            deltas: vec![0; block_size],
            _phantom: PhantomData,
        }
    }
```

I think I can first make it in, and than add an issue to control that behavior.

cpp/src/parquet/encoding.cc
```diff
@@ -2107,9 +2107,12 @@ class DictByteArrayDecoderImpl : public DictDecoderImpl<ByteArrayType>,
...
-  static constexpr uint32_t kValuesPerBlock = 128;
+  static constexpr uint32_t kValuesPerBlock =
+      std::is_same_v<int32_t, typename DType::c_type> ? 128 : 256;
   static constexpr uint32_t kMiniBlocksPerBlock = 4;
```
wjones127, line 2116: Just to make sure I understand, is this where the 128 / 256 are coming from here?
```suggestion
  static constexpr uint32_t kMiniBlocksPerBlock = 4;
  static constexpr uint32_t kValuesPerBlock = kMiniBlocksPerBlock * sizeof(typename DType::c_type);
```

wjones127 (approved):
This looks good to me. @mapleFU could you rebase, so a few more of the C++ CI jobs can be rerun?

cpp/src/parquet/encoding.cc
mapleFU: Yes

mapleFU: Although kValuesPerBlock is exactly equal to `kMiniBlocksPerBlock * sizeof(typename DType::c_type);`. This doesn't means they have some calculated relationship. So I hardcord it

wjones127 (approved):

PR #34645, opened 2023-03-20
GH-34644: [C++] Prefer unsafe casting by default in Substrait 

### LINKED ISSUE #34644: [C++] Prefer unsafe casting by default in Substrait

Substrait specifies a failure behavior (return null or error) for casting.  However, it does not specify what exactly constitutes a failure. When we added bindings for the cast expression we assumed that Substrait wanted a safe cast.  However, I have since learned that most existing engines (e.g. postgres, spark, etc.) will perform an unsafe cast by default and only consider a failure when casting is impossible (e.g. "foo" to integer).

I'd like to eventually see Substrait add support for specifying more precisely what kind of casting should occur.  In the meantime I think we should update the default to unsafe casting to match user expectation.

### Component(s)

C++

westonpace:
### Rationale for this change

See issue

Changes the default cast in Substrait to an unsafe cast.

### Are these changes tested?

Yes.

Yes

BREAKING CHANGE: The default Substrait behavior has changed slightly.  However, we believe this to be more in line with user expectation.
* Closes: #34644

### DIFF
--- a/cpp/src/arrow/engine/substrait/expression_internal.cc
+++ b/cpp/src/arrow/engine/substrait/expression_internal.cc
@@ -336,8 +336,9 @@ Result<compute::Expression> FromProto(const substrait::Expression& expr,
       if (cast_exp.failure_behavior() ==
           substrait::Expression::Cast::FailureBehavior::
               Expression_Cast_FailureBehavior_FAILURE_BEHAVIOR_THROW_EXCEPTION) {
-        return compute::call("cast", {std::move(input)},
-                             compute::CastOptions::Safe(std::move(type_nullable.first)));
+        return compute::call(
+            "cast", {std::move(input)},
+            compute::CastOptions::Unsafe(std::move(type_nullable.first)));
       } else if (cast_exp.failure_behavior() ==
                  substrait::Expression::Cast::FailureBehavior::
                      Expression_Cast_FailureBehavior_FAILURE_BEHAVIOR_RETURN_NULL) {
--- a/cpp/src/arrow/engine/substrait/serde_test.cc
+++ b/cpp/src/arrow/engine/substrait/serde_test.cc
@@ -814,6 +814,19 @@ TEST(Substrait, Cast) {
   ASSERT_TRUE(expr.call());
 
   ASSERT_THAT(expr.call()->arguments[0].call()->function_name, "cast");
+
+  std::shared_ptr<compute::FunctionOptions> call_opts =
+      expr.call()->arguments[0].call()->options;
+
+  ASSERT_TRUE(!!call_opts);
+  std::shared_ptr<compute::CastOptions> cast_opts =
+      std::dynamic_pointer_cast<compute::CastOptions>(call_opts);
+  ASSERT_TRUE(!!cast_opts);
+  // It is unclear whether a Substrait cast should be safe or not.  In the meantime we are
+  // assuming it is unsafe based on the behavior of many SQL engines.
+  ASSERT_TRUE(cast_opts->allow_int_overflow);
+  ASSERT_TRUE(cast_opts->allow_float_truncate);
+  ASSERT_TRUE(cast_opts->allow_decimal_truncate);
 }
 
 TEST(Substrait, CastRequiresFailureBehavior) {


jorisvandenbossche:
(I am also not sure that just using `unsafe` instead of `safe` fully matches "user expectation". We might want to use a more finer grained CastOptions specification)

icexelloss (approved):
LGTM. This is more aligned with behaviors of pandas/spark/SQL databases.

thisisnic (approved):
From what's been said in the discussion, it looks like this makes more sense than before, though we should definitely think more about this.

jorisvandenbossche:
> This is more aligned with behaviors of pandas/spark/SQL databases.

Whoops, my comment above was intended to refer to my comment on the issue (https://github.com/apache/arrow/issues/34644#issuecomment-1477405158), but it seems I forgot to add that. But so for overflow, this PR is _not_ more aligned with eg postgres (didn't check other databases, though). 

So I mostly wanted to mention, if we change the default, we don't necessarily have to use either one of "safe" or "unsafe", since we already have more fine grained control over the option class, we can also use something like

```
auto options = compute::CastOptions::Unsafe(std::move(type_nullable.first));
options.allow_int_overflow = false;
```

Now, I am not up to date with the discussions in substrait around this, so I don't know if that matches better the intended default in substrait. But it might match better with user expectations from _some_ SQL databases (I only checked Postgres)

icexelloss:
> > This is more aligned with behaviors of pandas/spark/SQL databases.
> 
> Whoops, my comment above was intended to refer to my comment on the issue ([#34644 (comment)](https://github.com/apache/arrow/issues/34644#issuecomment-1477405158)), but it seems I forgot to add that. But so for overflow, this PR is _not_ more aligned with eg postgres (didn't check other databases, though).
> 
> So I mostly wanted to mention, if we change the default, we don't necessarily have to use either one of "safe" or "unsafe", since we already have more fine grained control over the option class, we can also use something like
> 
> ```
> auto options = compute::CastOptions::Unsafe(std::move(type_nullable.first));
> options.allow_int_overflow = false;
> ```
> 
> Now, I am not up to date with the discussions in substrait around this, so I don't know if that matches better the intended default in substrait. But it might match better with user expectations from _some_ SQL databases (I only checked Postgres)

The behavior for overflow is a valid concern. I tested on pandas and Spark, both which allows overflow by default and therefore, is different from the PostgreSQL behavior. I am slight +0.5 for matching pandas/Spark behavior because my background is from Python data science tooling but I can see the other argument as well.

@jorisvandenbossche if you feel strongly about the overflow behavior then I don't object to disallowing it by default, but I hope to get some quick resolution as the "allow truncate" behavior is what I care about most.

icexelloss:
@westonpace @jorisvandenbossche I would like to push for some consensus a bit here. Just to reiterate, from my side, the minimal requirement is for the default behavior to support floating number to integer truncate. 

For the overflow behavior, I am leaning towards supporting overflow behavior by default as well but ok with disallow it by default if you guys have strong opinions here. But I would like move forward with this somehow so would like to hear what u guys think.

westonpace:
I think the only workable long term solution is expanding Substrait's capability when it comes to expressing cast behavior.

For the short term I don't have a strong opinion.  As far as I know @icexelloss is the only one consuming cast via Substrait in this way currently so I'd be ok with matching spark over postgres.

icexelloss:
@westonpace @jorisvandenbossche Thanks both for the detailed discussion. Since there is no object I went ahead and merged this one to unblock.

icexelloss:
> I personally think the more sensible behaviour for users is to not allow overflow.

I don't object to this (although I am somewhat neutral). 

> But also, it might not matter that much what the default is for substrait?

From my understanding, substrait doesn't define this in its spec (it also have the option to return null or raise when "cast error" happens but doesn't have a clear spec for what are "cast error")

>  Because it can't have the "good" default for various use cases at the same time. So what matters is that there is a way to specify the behaviour, so that producers of substrait can specify the behaviour so they can choose whathever default they want for the users of their system, and ensure this gets reflected correctly in the substrait plan?

The way I see it. There are two options -
(1) Have a way for end user to configure this (in the user facing API) and pass that all the way from user API -> substrait -> backend. This is certainly the most configurable way.
(2) Have a way for users of Acero (i.e., me) to configure the behavior for their users. E.g., having backend-wide default for casting behaviors. E.g., I might decide I don't care much about overflow for my users so I am gonna allow it. But another user of Acero might decide he/she want to raise for overflow.

From my perspective, (2) is probably good enough for my requirement. (1) feels ideal in theory but might not be necessary.

PR #34654, opened 2023-03-21
GH-34248: [Python] Expose the order_by node

### LINKED ISSUE #34248: [C++] Add an order_by node which can reassign an ordering mid-plan

#34136 added ordering to exec plans.  We can now take our order_by_sink node and make an actual order_by node which can be placed anywhere in a plan.  This can use the same naive implementation in order_by_sink.

### Component(s)

C++

jorisvandenbossche:
Adds Python bindings for the OrderByNode added in https://github.com/apache/arrow/pull/34249 
* Closes: #34248

### DIFF
--- a/python/pyarrow/_acero.pyx
+++ b/python/pyarrow/_acero.pyx
@@ -28,7 +28,10 @@ from pyarrow.includes.libarrow_dataset cimport *
 from pyarrow.lib cimport (Table, pyarrow_unwrap_table, pyarrow_wrap_table,
                           RecordBatchReader)
 from pyarrow.lib import frombytes, tobytes
-from pyarrow._compute cimport Expression, FunctionOptions, _ensure_field_ref, _true
+from pyarrow._compute cimport (
+    Expression, FunctionOptions, _ensure_field_ref, _true,
+    unwrap_null_placement, unwrap_sort_order
+)
 
 
 cdef class ExecNodeOptions(_Weakrefable):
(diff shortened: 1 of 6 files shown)


AlenkaF (approved):
This is very cool =)
I have tried it locally, works great. Can't think of anything to add to the tests.

I am a bit unsure about the reason for the change in `CSortKey` class though. This class was only used in `Select*`, `Rank*` and `SortOptions` where, I guess, a field reference could have only been a string column name.

In `OrderByNodeOptions` it can also be an integer, or an expression. Then, I think, this should also be added in the docstrings for other classes that use `sort_keys` and add this cases to the tests? (maybe not as a part of this PR)

jorisvandenbossche:
> I am a bit unsure about the reason for the change in `CSortKey` class though. This class was only used in `Select*`, `Rank*` and `SortOptions` where, I guess, a field reference could have only been a string column name.

Yeah, this was actually changed a while ago in C++, but the cython bindings exposing it as ``CSortKey(c_string name, ..)`` still worked fine because there is an implicit conversion from string to FieldRef possible. 
For this PR however, I needed to create a SortKey from a FieldRef (to be consistent with the rest of the node options), and so I needed to update the constructor  to  ``CSortKey(CFieldRef ..)``, which also meant updating it in the other places. 

> Then, I think, this should also be added in the docstrings for other classes that use sort_keys and add this cases to the tests? (maybe not as a part of this PR)

Good point. Updated the docstrings and tests

PR #34668, opened 2023-03-21
GH-34667: [C++][Parquet] Test DeltaLengthByteArrayDecoder with invalid inputs

### LINKED ISSUE #34667: [C++][Parquet] DeltaLengthByteArrayDecoder fails to reject empty buffers

If a corrupted Parquet file indicates it has `num_values > 0` in a page while the page is empty, `DeltaLengthByteArrayDecoder` will cause buffer overflow.

### Component(s)

C++, Parquet

wjones127:
### Rationale for this change

We should make sure that we are defensive when it comes to invalid inputs. Otherwise malicious actors might send invalid Parquet files as a way to crash a users service.

Makes sure we reject empty pages by deferring to the DeltaBitPackDecoder to error when the lengths header is missing. Also adds a DCHECK that validates `num_valid_values_` was initialized (it wasn't when `len = 0`). Also adds a few more general tests cases for the encoding, covering empty arrays and empty strings.

### Are these changes tested?

Yes, several unit tests have been added.

* Closes: #34667

### DIFF
--- a/cpp/src/parquet/encoding.cc
+++ b/cpp/src/parquet/encoding.cc
@@ -2708,7 +2708,7 @@ class DeltaLengthByteArrayDecoder : public DecoderImpl,
 
   void SetData(int num_values, const uint8_t* data, int len) override {
     num_values_ = num_values;
-    if (len == 0) {
+    if (ARROW_PREDICT_FALSE(len <= 0)) {
       if (num_values > 0) {
         throw ParquetException(
             "Delta length byte array page was empty, but expected {num_values} values.");
--- a/cpp/src/parquet/encoding.cc
+++ b/cpp/src/parquet/encoding.cc
@@ -2708,7 +2708,14 @@ class DeltaLengthByteArrayDecoder : public DecoderImpl,
 
   void SetData(int num_values, const uint8_t* data, int len) override {
     num_values_ = num_values;
-    if (len == 0) return;
+    if (len == 0) {
+      if (num_values > 0) {
+        throw ParquetException(
+            "Delta length byte array page was empty, but expected {num_values} values.");
+      }
+      num_valid_values_ = 0;
+      return;
+    }
     decoder_ = std::make_shared<::arrow::bit_util::BitReader>(data, len);
     DecodeLengths();
   }
@@ -2723,6 +2730,7 @@ class DeltaLengthByteArrayDecoder : public DecoderImpl,
     // Decode up to `max_values` strings into an internal buffer
     // and reference them into `buffer`.
     max_values = std::min(max_values, num_valid_values_);
+    DCHECK_GE(max_values, 0);
     if (max_values == 0) {
       return 0;
     }
--- a/cpp/src/parquet/encoding.cc
+++ b/cpp/src/parquet/encoding.cc
@@ -2708,10 +2708,6 @@ class DeltaLengthByteArrayDecoder : public DecoderImpl,
 
   void SetData(int num_values, const uint8_t* data, int len) override {
     num_values_ = num_values;
-    if (ARROW_PREDICT_FALSE(len <= 0)) {
-      throw ParquetException(
-          "Delta length byte array page was missing header for lengths.");
-    }
     decoder_ = std::make_shared<::arrow::bit_util::BitReader>(data, len);
     DecodeLengths();
   }
--- a/cpp/src/parquet/encoding.cc
+++ b/cpp/src/parquet/encoding.cc
@@ -2709,12 +2709,8 @@ class DeltaLengthByteArrayDecoder : public DecoderImpl,
   void SetData(int num_values, const uint8_t* data, int len) override {
     num_values_ = num_values;
     if (ARROW_PREDICT_FALSE(len <= 0)) {
-      if (num_values > 0) {
-        throw ParquetException(
-            "Delta length byte array page was empty, but expected {num_values} values.");
-      }
-      num_valid_values_ = 0;
-      return;
+      throw ParquetException(
+          "Delta length byte array page was missing header for lengths.");
     }
     decoder_ = std::make_shared<::arrow::bit_util::BitReader>(data, len);
     DecodeLengths();
--- a/cpp/src/parquet/encoding_test.cc
+++ b/cpp/src/parquet/encoding_test.cc
@@ -1615,7 +1615,7 @@ class TestDeltaLengthByteArrayEncoding : public TestEncodingBase<Type> {
 
     encoder->PutSpaced(draws_, num_values_, valid_bits, valid_bits_offset);
     encode_buffer_ = encoder->FlushValues();
-    decoder->SetData(num_values_ - null_count, encode_buffer_->data(),
+    decoder->SetData(num_values_, encode_buffer_->data(),
                      static_cast<int>(encode_buffer_->size()));
     auto values_decoded = decoder->DecodeSpaced(decode_buf_, num_values_, null_count,
                                                 valid_bits, valid_bits_offset);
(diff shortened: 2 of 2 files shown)


cpp/src/parquet/encoding_test.cc
```diff
@@ -1639,6 +1639,79 @@ TYPED_TEST(TestDeltaLengthByteArrayEncoding, BasicRoundTrip) {
...
+    int values_decoded = decoder->DecodeArrow(num_values, 0, nullptr, 0, &acc);
+    ASSERT_EQ(values_decoded, num_values);
+
+    std::shared_ptr<::arrow::Array> roundtripped_array;
+    ASSERT_OK(acc.builder->Finish(&roundtripped_array));
+    ASSERT_ARRAYS_EQUAL(*array, *roundtripped_array);
+  }
+}
+
+TEST(DeltaLengthByteArrayEncoding, RejectBadBuffer) {
```
wjones127, line 1690: I'd welcome any other ideas for bad buffers we should try.

cpp/src/parquet/encoding_test.cc
```diff
@@ -1759,12 +1832,10 @@ TEST(DeltaLengthByteArrayEncodingAdHoc, ArrowDirectPut) {
...
-  const uint8_t data[] = {
-      0x80, 0x01, 0x04, 0x04, 0x0a, 0x00, 0x01, 0x00, 0x00, 0x00, 0x02, 0x00,
-      0x00, 0x00, 0x48, 0x65, 0x6c, 0x6c, 0x6f, 0x57, 0x6f, 0x72, 0x6c, 0x64,
-      0x46, 0x6f, 0x6f, 0x62, 0x61, 0x72, 0x41, 0x44, 0x42, 0x43, 0x45, 0x46,
-  };
-  auto encoded = Buffer::Wrap(data, sizeof(data));
+  auto encoded =
+      ::arrow::ConcatenateBuffers(
+          {DeltaEncode({5, 5, 6, 6}), std::make_shared<Buffer>("HelloWorldFoobarADBCEF")})
+          .ValueOrDie();
```
wjones127, line 1838: Added for readability.

emkornfield (approved):

cpp/src/parquet/encoding.cc
```diff
@@ -2723,6 +2730,7 @@ class DeltaLengthByteArrayDecoder : public DecoderImpl,
     // Decode up to `max_values` strings into an internal buffer
     // and reference them into `buffer`.
     max_values = std::min(max_values, num_valid_values_);
+    DCHECK_GE(max_values, 0);
```
emkornfield, line 2733: should this be an exception?

wjones127: That might be safer, but as-is we should never trip that. I added this in case someone refactors the class and adds another code path where `num_valid_values_` might not be initialized. Do you still think it's worthwhile?

cpp/src/parquet/encoding.cc
```diff
@@ -2708,7 +2708,14 @@ class DeltaLengthByteArrayDecoder : public DecoderImpl,
 
   void SetData(int num_values, const uint8_t* data, int len) override {
     num_values_ = num_values;
-    if (len == 0) return;
+    if (len == 0) {
```
rok, line 2711:
```suggestion
    if (ARROW_PREDICT_FALSE(len == 0)) {
```

cpp/src/parquet/encoding.cc
```diff
@@ -2708,7 +2708,14 @@ class DeltaLengthByteArrayDecoder : public DecoderImpl,
...
+      if (num_values > 0) {
+        throw ParquetException(
+            "Delta length byte array page was empty, but expected {num_values} values.");
```
rok, line 2714: Does this print `num_values`? Perhaps we can check the message with `EXPECT_THROW_THAT`

cpp/src/parquet/encoding_test.cc
rok: Much more readable now :D

rok (approved):
Thanks for catching this one! Just a couple of suggestions.

cpp/src/parquet/encoding_test.cc
rok: Perhaps we should also check for `num_values` or `len` being negative as well?

cpp/src/parquet/encoding.cc
```diff
@@ -2708,7 +2708,14 @@ class DeltaLengthByteArrayDecoder : public DecoderImpl,
 
   void SetData(int num_values, const uint8_t* data, int len) override {
     num_values_ = num_values;
-    if (len == 0) return;
+    if (len == 0) {
+      if (num_values > 0) {
```
wgtmac, line 2712: I did a quick check. `num_values` is the number of values with nulls considered in a data page.

https://github.com/apache/arrow/blob/main/cpp/src/parquet/column_reader.cc#L922
```cpp
    current_decoder_->SetData(static_cast<int>(num_buffered_values_), buffer,
                              static_cast<int>(data_size));
```

https://github.com/apache/arrow/blob/main/cpp/src/parquet/column_reader.cc#L949
```cpp
  // The total number of values stored in the data page. This is the maximum of
  // the number of encoded definition levels or encoded values. For
  // non-repeated, required columns, this is equal to the number of encoded
  // values. For repeated or optional values, there may be fewer data values
  // than levels, and this tells you how many encoded levels there are in that
  // case.
  int64_t num_buffered_values_;
```

Therefore, the input `num_values > 0 && len == 0` looks valid to me. Should we not throw here but decode nothing from the decoder instead?

mapleFU:
The `num_values` is a tricky word, num_values is from DataPageHeader, and means that `num of non-null value + num of null values`. So it includes null value.

In DeltaLengthByteArray, seems real num_values can comes from DELTA encoding.

mapleFU:
The `num_values` is a tricky word, num_values is from DataPageHeader, and means that `num of non-null value + num of null values`. So it includes null value.

In DeltaLengthByteArray, seems real num_values can comes from DELTA encoding.

cpp/src/parquet/encoding.cc
mapleFU: I think num_values could greater than 0, let me write a test for that

mapleFU:
@wjones127 Would you mind providing me the input file or test case? Or where does this segfault? I think the program would be prevent from fatal in somewhere

mapleFU:
And I think you can add a test case:

```
diff --git a/cpp/src/parquet/encoding_test.cc b/cpp/src/parquet/encoding_test.cc
index 7dcee1152..31ebd4188 100644
--- a/cpp/src/parquet/encoding_test.cc
+++ b/cpp/src/parquet/encoding_test.cc
@@ -1713,7 +1713,7 @@ class TestDeltaLengthByteArrayEncoding : public TestEncodingBase<Type> {
 
     encoder->PutSpaced(draws_, num_values_, valid_bits, valid_bits_offset);
     encode_buffer_ = encoder->FlushValues();
-    decoder->SetData(num_values_ - null_count, encode_buffer_->data(),
+    decoder->SetData(num_values_, encode_buffer_->data(),
:
```

wgtmac:
Thanks @mapleFU for investigation. This is what I mean exactly.

wjones127:
@mapleFU 
> Would you mind providing me the input file or test case? Or where does this segfault? I think the program would be prevent from fatal in somewhere

You are right it didn't actually crash, it just caused a buffer overflow, which tripped the sanitizers. This was discovered in our fuzzing CI.

Thanks for providing the test. It doesn't actually trigger the codepath for `len <= 0` since there is the header for the encoded lengths present:

```
"\x80\U00000001\U00000004\0\0"
```

That does make me wonder whether we should reject an empty buffer as corrupt (missing required header?). But also just loading no values makes sense too.

wjones127:
So I found that the DeltaBitPackDecoder will error if the header is missing, so I think it makes sense for us to also error if it's missing at the beginning of DeltaLengthByteArray. To accomplish this, I actually deleted the branch, so it can get to the `DecodeLengths()` which already handles this error:
 https://github.com/wjones127/arrow/blob/f59fe0a2267b12069586e7c6b0351c48bf296371/cpp/src/parquet/encoding.cc#L2424

wgtmac (approved):
LGTM. Thanks!

cpp/src/parquet/encoding.cc
```diff
@@ -2708,7 +2708,6 @@ class DeltaLengthByteArrayDecoder : public DecoderImpl,
 
   void SetData(int num_values, const uint8_t* data, int len) override {
     num_values_ = num_values;
-    if (len == 0) return;
```
mapleFU, line 2711: nice catch!

mapleFU (approved):

PR #34671, opened 2023-03-22
GH-15233: [C++] Correct thread contention in s3fs

### LINKED ISSUE #15233: pyarrow copy_files hangs indefinitely

We are working on a python package that calls pyarrow's copy_files function to copy a local directory to S3. We notice that this seems to hang indefinitely for a directory, even though it works for individual files.

A simple reproducer seems to be:
```py
from ray.air._internal.remote_storage import upload_to_uri
dir = "/some/directory/local"
uri = "s3://some/s3/bucket"
upload_to_uri(dir, uri)
```

Where ray just wraps pyarrow.fs.copy_files: https://github.com/ray-project/ray/blob/d7b2b49a962bf33dae7a50376f159ab15d80800f/python/ray/air/_internal/remote_storage.py#L195

This results in the following flamegraph.

profile-idle

And an `strace` of that process looks like

```
stat("/some/directory/local/result.json", {st_mode=S_IFREG|0664, st_size=2345, ...}) = 0
getpid()                                = 1639901
futex(0x5580a3b563e4, FUTEX_WAKE_PRIVATE, 1) = 1
getpid()                                = 1639901
futex(0x5580a3b563e4, FUTEX_WAKE_PRIVATE, 1) = 1
getpid()                                = 1639901
futex(0x5580a3b563e4, FUTEX_WAKE_PRIVATE, 1) = 1
getpid()                                = 1639901
futex(0x5580a3b563e4, FUTEX_WAKE_PRIVATE, 1) = 1
getpid()                                = 1639901
futex(0x5580a3b563e4, FUTEX_WAKE_PRIVATE, 1) = 1
getpid()                                = 1639901
futex(0x5580a3b563e4, FUTEX_WAKE_PRIVATE, 1) = 1
getpid()                                = 1639901
futex(0x5580a3b563e4, FUTEX_WAKE_PRIVATE, 1) = 1
getpid()                                = 1639901
futex(0x5580a3b563e4, FUTEX_WAKE_PRIVATE, 1) = 1
getpid()                                = 1639901
getpid()                                = 1639901
getpid()                                = 1639901
futex(0x5580a3d374e8, FUTEX_WAIT_PRIVATE, 0, NULL) = ? ERESTARTSYS (To be restarted if SA_RESTART is set)
--- SIGTSTP {si_signo=SIGTSTP, si_code=SI_KERNEL} ---
--- stopped by SIGTSTP ---
+++ killed by SIGKILL +++
````

This is using
```
.venv ❯ pip show pyarrow                                                                                                                                                                                                    [23/01/6|17:08]
Name: pyarrow
Version: 8.0.0
```
On Python 3.10.9, on a machine running Ubuntu 20.04.5 LTS.

### Component(s)

Python

### PR DESCRIPTION
**Pull request by EpsilonPrime**
Avoids thread contention which can occur when using CopyFiles to copy files from a filesystem other than S3 to the S3FS.  This is accomplished by not trying to complete the merging of a potentially multipart file asynchronously while already in an I/O thread (instead that work is completed in the already active thread).

* Closes: #15233

### DIFF
--- a/cpp/src/arrow/filesystem/s3fs.cc
+++ b/cpp/src/arrow/filesystem/s3fs.cc
@@ -1244,12 +1244,12 @@ class ObjectOutputStream final : public io::OutputStream {
 
     if (current_part_) {
       // Upload last part
-      RETURN_NOT_OK(CommitCurrentPart(true));
+      RETURN_NOT_OK(CommitCurrentPart(/*always_foreground_writes=*/ true));
     }
 
     // S3 mandates at least one part, upload an empty one if necessary
     if (part_number_ == 1) {
-      RETURN_NOT_OK(UploadPart("", 0, true));
+      RETURN_NOT_OK(UploadPart("", 0, /*always_foreground_writes=*/ true));
     }
 
     // Wait for in-progress uploads to finish (if async writes are enabled)
@@ -1307,7 +1307,8 @@ class ObjectOutputStream final : public io::OutputStream {
     if (!current_part_ && nbytes >= part_upload_threshold_) {
       // No current part and data large enough, upload it directly
       // (without copying if the buffer is owned)
-      RETURN_NOT_OK(UploadPart(data, nbytes, false, owned_buffer));
+      RETURN_NOT_OK(UploadPart(data, nbytes,
+                               /*always_foreground_writes=*/ false, owned_buffer));
       pos_ += nbytes;
       return Status::OK();
     }
@@ -1324,7 +1325,7 @@ class ObjectOutputStream final : public io::OutputStream {
 
     if (current_part_size_ >= part_upload_threshold_) {
       // Current part large enough, upload it
-      RETURN_NOT_OK(CommitCurrentPart(false));
+      RETURN_NOT_OK(CommitCurrentPart(/*always_foreground_writes=*/ false));
     }
 
     return Status::OK();
--- a/cpp/src/arrow/filesystem/s3fs.cc
+++ b/cpp/src/arrow/filesystem/s3fs.cc
@@ -1244,12 +1244,12 @@ class ObjectOutputStream final : public io::OutputStream {
 
     if (current_part_) {
       // Upload last part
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: cpp/src/arrow/filesystem/s3fs.cc [hunk 1]
```diff
@@ -1244,12 +1244,12 @@ class ObjectOutputStream final : public io::OutputStream {
 
     if (current_part_) {
       // Upload last part
-      RETURN_NOT_OK(CommitCurrentPart());
+      RETURN_NOT_OK(CommitCurrentPart(true));
```
cyb70289, line 1247: Add a comment to make it easier to reason the boolean argument at caller side?
`CommitCurrentPart(/*always_foreground_writes=*/true)`

EpsilonPrime: Sure, I've gotten away from the practice of adding the extra comments as my IDE does this for me, but done.

westonpace:
I think `CloseAsync` is expected to act as an async method.  By pushing the writes to the foreground it is no longer acting as an async method.

This is similar to the issues we ran into on the read path.  The solution was to make everything async by default and the only time we wrap async with sync is at the API layer (e.g. on that first call from the user to a sync method).  This allows us to keep one async path for all the "internals" and only mirror the two paths at the edges.

Instead, could we change `CopyFile` (and `CopyFiles`) to use async methods?  Or create `CopyFileAysnc` and `CopyFilesAsync` (and change `CopyFiles` to return `CopyFilesAsync.result()`?

EpsilonPrime:
> @EpsilonPrime are you still interested in moving this forward? It looks like there are some failing checks and some review feedbacks to address

@amol- Yes, I'm still working on this but my alternate version of the solution (using futures throughout) is not yet working.

zeroshade:
@EpsilonPrime Just checking back in on this, are you still working on this and we should keep it open? Do you need any assistance in getting it working?

EpsilonPrime:
Solved by another PR.

PR #34672, opened 2023-03-22
GH-34670: [Packaging][C++] Add support for customizing GDB plugin install directory

### LINKED ISSUE #34670: [Packaging][C++] gdb integration is not compatible with conda packaging

gdb apparently needs to reflect the path of a given library (for which we want to have gdb-support) in the path to that support artefact. Currently this is being done by using
https://github.com/apache/arrow/blob/1ba4425fab35d572132cb30eee6087a7dca89853/cpp/src/arrow/CMakeLists.txt#L674-L675

However, this is incompatible with the way that conda-forge works, where the packaging step needs to install the package in order to determine which files need to be packaged at all. The use of `CMAKE_INSTALL_FULL_LIBDIR` means that we'll be baking the path of that temporary build environment in the feedstock into the final artefact, which is both non-functional, and additionally leads to [problems](https://github.com/conda-forge/arrow-cpp-feedstock/issues/935) due to path length.

The core issue is that we do not know the path to the final environment in conda until we're actually in it. In order to support the gdb integration nevertheless, the current work-around is to use a [patch](https://github.com/conda-forge/arrow-cpp-feedstock/blob/main/recipe/patches/0001-don-t-bake-non-relocatable-CMAKE_INSTALL_FULL_LIBDIR.patch) and an [activation script](https://github.com/conda-forge/arrow-cpp-feedstock/blob/main/recipe/activate.sh) (which will always execute upon entering/activating an environment).

@kou asked me in https://github.com/apache/arrow/pull/34652 to open an issue for this. It would be great if we could have a better way to support gdb also in conda-forge, but at the moment I don't see how that would work.

### Component(s)

Packaging

kou:
### Rationale for this change

It's for conda package. We can't determine the final GDB plugin path on building. We can adjust the path by replacing a placeholder on installing. 

Conda package can specify a placeholder for GDB plugin path.

### Are these changes tested?

Yes.

Yes.
* Closes: #34670

### DIFF
--- a/cpp/cmake_modules/DefineOptions.cmake
+++ b/cpp/cmake_modules/DefineOptions.cmake
@@ -615,9 +615,9 @@ that have not been built"
                 OFF)
 
   define_option_string(ARROW_GDB_INSTALL_DIR
-                       "Use a custom install directory for GDB plugin. \
-In general, you don't need to specify this because the default \
-(CMAKE_INSTALL_FULL_BINDIR on Windows, CMAKE_INSTALL_FULL_LIBDIR otherwise) \
+                       "Use a custom install directory for GDB plugin.;\
+In general, you don't need to specify this because the default;\
+(CMAKE_INSTALL_FULL_BINDIR on Windows, CMAKE_INSTALL_FULL_LIBDIR otherwise);\
 is reasonable."
                        "")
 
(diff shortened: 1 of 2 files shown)


kou:
@github-actions crossbow submit -g linux

kou:
@github-actions crossbow submit -g linux

kou:
@github-actions crossbow submit test-conda-python-3.10

PR #34679, opened 2023-03-22
GH-15286: [GLib] Add GArrowIndexOptions

### LINKED ISSUE #15286: [GLib] Add support for IndexOptions
#### Target option

IndexOptions

#### Proposed feature

Add support for IndexOptions.

#### Impact of this request

It will be used in index() which is an basic feature to get index from an Array.

https://arrow.apache.org/docs/cpp/compute.html#aggregations

### Component(s)

GLib, Ruby

kou:
### Rationale for this change

It's needed to use `index` compute function.

Add `GArrowIndexOptions` and related API for easy to use.

### Are these changes tested?

Yes.

Yes.
* Closes: #15286


kou:
+1

PR #34685, opened 2023-03-22
GH-34437: [R] Use FetchNode and OrderByNode

### LINKED ISSUE #31980: [C++] Support limit operation
Either an option to a SinkNode (TopK already takes a number of results to keep) or a streaming LimitNode that only lets N rows through.

**Reporter**: [Neal Richardson](https://issues.apache.org/jira/browse/ARROW-16628) / @nealrichardson

<sub>**Note**: *This issue was originally created as [ARROW-16628](https://issues.apache.org/jira/browse/ARROW-16628). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

### LINKED ISSUE #31982: [C++] Support order by derived column
You can't do the equivalent of

```Java

SELECT x, y FROM table ORDER BY x * y ASC;
```

because sorting requires a named column AND because sorting is only done in a SinkNode. You can project to {x, y, x\*y} then sort on x\*y, but you can't then project back to {x, y} on the sorted data because that's a new ExecPlan and order is not preserved. In R we have to handle this outside of an ExecPlan.

**Reporter**: [Neal Richardson](https://issues.apache.org/jira/browse/ARROW-16631) / @nealrichardson

<sub>**Note**: *This issue was originally created as [ARROW-16631](https://issues.apache.org/jira/browse/ARROW-16631). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

### LINKED ISSUE #34437: [R] Use FetchNode and OrderByNode

See #34059. There's at least [one workaround](https://github.com/apache/arrow/blob/main/r/R/query-engine.R#L224-L233) we can remove and push the work into the ExecPlan instead of massaging the result after. 

### Component(s)

R

nealrichardson:
### Rationale for this change

See also #32991. By using the new nodes, we're closer to having all dplyr query business happening inside the ExecPlan. Unfortunately, there are still two cases where we have to apply operations in R after running a query:

* #34941: Taking head/tail on unordered data, which has non-deterministic results but that should be possible, in the case where the user wants to see a slice of the result, any slice
* #34942: Implementing tail in the FetchNode or similar would enable removing more hacks and workarounds.

Once those are resolved, we can simply further and then move to the new Declaration class.

This removes the use of different SinkNodes and many R-specific workarounds to support sorting and head/tail, so *almost* 
everything we do in a query should be represented in an ExecPlan. 

### Are these changes tested?

Yes. This is mostly an internal refactor, but behavior changes are accompanied by test updates.

The `show_query()` method will print slightly different ExecPlans. In many cases, they will be more informative. 

`tail()` now actually returns the tail of the data in cases where the data has an implicit order (currently only in-memory tables). Previously it was non-deterministic (and would return the head or some other slice of the data).

When printing query objects that include `summarize()` when the `arrow.summarize.sort = TRUE` option is set, the sorting is correctly printed.

It's unclear if there should be changes in performance; running benchmarks would be good but it's also not clear that our benchmarks cover all affected scenarios. 

* Closes: #34437
* Closes: #31980
* Closes: #31982

### DIFF
--- a/r/R/arrowExports.R
+++ b/r/R/arrowExports.R
@@ -444,8 +444,8 @@ ExecPlanReader__PlanStatus <- function(reader) {
   .Call(`_arrow_ExecPlanReader__PlanStatus`, reader)
 }
 
-ExecPlan_run <- function(plan, final_node, sort_options, metadata, head) {
-  .Call(`_arrow_ExecPlan_run`, plan, final_node, sort_options, metadata, head)
+ExecPlan_run <- function(plan, final_node, metadata) {
+  .Call(`_arrow_ExecPlan_run`, plan, final_node, metadata)
 }
 
 ExecPlan_ToString <- function(plan) {
--- a/r/R/arrowExports.R
+++ b/r/R/arrowExports.R
@@ -460,6 +460,10 @@ ExecNode_output_schema <- function(node) {
   .Call(`_arrow_ExecNode_output_schema`, node)
 }
 
+ExecNode_has_ordered_batches <- function(node) {
+  .Call(`_arrow_ExecNode_has_ordered_batches`, node)
+}
+
 ExecNode_Scan <- function(plan, dataset, filter, projection) {
   .Call(`_arrow_ExecNode_Scan`, plan, dataset, filter, projection)
 }
--- a/r/R/arrowExports.R
+++ b/r/R/arrowExports.R
@@ -488,6 +488,14 @@ ExecNode_Union <- function(input, right_data) {
   .Call(`_arrow_ExecNode_Union`, input, right_data)
 }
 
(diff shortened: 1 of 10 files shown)


r/R/arrow-info.R
```diff
@@ -39,6 +39,7 @@ arrow_info <- function() {
     version = packageVersion("arrow"),
     options = opts[grep("^arrow\\.", names(opts))],
     capabilities = c(
+      acero = arrow_with_acero(),
```
thisisnic, line 42: I think you might need a rebase? This was done in #34943 right?

thisisnic (approved):
Looks great, just needs a rebase!

PR #34691, opened 2023-03-22
GH-34639: [C++] Support RecordBatch::FromStructArray even if struct array has nulls/offsets

### LINKED ISSUE #34639: [Python] Info about offset is loss when converting struct array to record batch

Hello,
it seems that when a struct is pointing to a part of a larger array via an offset and we try to convert it to a record batch, that information is lost and we get a record batch with columns having the larger array length.

For now, a workaround is to select all the actual values in the array before converting it to a recordbatch (though this solution does not scale well, slicing does not work).

The following code reproduces the error and shows the workaround: 
```import pyarrow as pa

# create a struct and a table having as column the struct, then split it into record batches
struct_1=pa.StructArray.from_arrays([pa.array([1.,2.]),pa.array(['a','b'])],names=['col1','col2'])
out=pa.Table.from_arrays(arrays=[struct_1],names=['struct'])
batches=out.to_batches(max_chunksize=1)

#convert to  struct arrays, here each array has an offset referencing the table
arrays=[pa.StructArray.from_arrays(batch.columns,names=batch.schema.names) for batch in batches]
#select the struct
modified_arrays=[array.field('struct') for array in arrays]

# just take length of each array to show difference
taken_arrays=[array.take(pa.array([0])) for array in modified_arrays]

for standard,taken in zip(modified_arrays,taken_arrays):
    #arrays are equals
    assert standard==taken
    assert len(standard)==len(taken)==1
    #but record batches are different!
    assert len(pa.RecordBatch.from_struct_array(standard).column(0))==2
    assert len(pa.RecordBatch.from_struct_array(taken).column(0))==1 ```

### Component(s)

Python

### PR DESCRIPTION
**Pull request by westonpace**
### Rationale for this change

A struct array can have a validity map and an offset.  A record batch cannot.  When converting from a struct array to a record batch we were throwing an error if a validity map was present and returning the wrong data if an offset was present.

If a validity map or offset are present then StructArray::Flatten is used to push the offset and validity map into the struct array.  Note: this means that RecordBatch::FromStructArray will not be zero-copy if it has to push down a validity map.

### Are these changes tested?

Yes

Yes, RecordBatch::FromStructArray now takes in a memory pool because it might have to make allocations when pushing validity bitmaps down.

* Closes: #34639

### DIFF
--- a/cpp/src/arrow/record_batch.cc
+++ b/cpp/src/arrow/record_batch.cc
@@ -198,14 +198,20 @@ Result<std::shared_ptr<RecordBatch>> RecordBatch::MakeEmpty(
 }
 
 Result<std::shared_ptr<RecordBatch>> RecordBatch::FromStructArray(
-    const std::shared_ptr<Array>& array) {
+    const std::shared_ptr<Array>& array, MemoryPool* memory_pool) {
   if (array->type_id() != Type::STRUCT) {
     return Status::TypeError("Cannot construct record batch from array of type ",
                              *array->type());
   }
-  if (array->null_count() != 0) {
-    return Status::Invalid(
-        "Unable to construct record batch from a StructArray with non-zero nulls.");
+  if (array->null_count() != 0 || array->offset() != 0) {
+    // If the struct array has a validity map or offset we need to push those into
+    // the child arrays via Flatten since the RecordBatch doesn't have validity/offset
+    const std::shared_ptr<StructArray>& struct_array =
+        internal::checked_pointer_cast<StructArray>(array);
+    ARROW_ASSIGN_OR_RAISE(std::vector<std::shared_ptr<Array>> fields,
+                          struct_array->Flatten(memory_pool));
+    return Make(arrow::schema(array->type()->fields()), array->length(),
+                std::move(fields));
   }
   return Make(arrow::schema(array->type()->fields()), array->length(),
               array->data()->child_data);
--- a/cpp/src/arrow/record_batch.h
+++ b/cpp/src/arrow/record_batch.h
@@ -84,6 +84,7 @@ class ARROW_EXPORT RecordBatch {
   /// This constructs a record batch using the child arrays of the given
   /// array, which must be a struct array.
   ///
+  /// \param[in] array the source array, must be a StructArray
   /// \param[in] pool the memory pool to allocate new validity bitmaps
   ///
   /// This operation will usually be zero-copy.  However, if the struct array has an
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by westonpace** [comment]
CC @felipecrv for review

**Review by felipecrv — APPROVED** [review]

File: cpp/src/arrow/record_batch.cc [hunk 1]
```diff
@@ -198,14 +198,20 @@ Result<std::shared_ptr<RecordBatch>> RecordBatch::MakeEmpty(
...
+    const std::shared_ptr<Array>& array, MemoryPool* memory_pool) {
   if (array->type_id() != Type::STRUCT) {
     return Status::TypeError("Cannot construct record batch from array of type ",
                              *array->type());
   }
-  if (array->null_count() != 0) {
-    return Status::Invalid(
-        "Unable to construct record batch from a StructArray with non-zero nulls.");
+  if (array->null_count() != 0 || array->offset() != 0) {
+    // If the struct array has a validity map or offset we need to push those into
```
wgtmac, line 207: If the struct_array only has an offset (w/o validity map), does `struct_array->Slice(offset, length)` work as well which should be zero copy?

cpp/src/arrow/record_batch.h
```diff
@@ -82,10 +82,15 @@ class ARROW_EXPORT RecordBatch {
   /// \brief Construct record batch from struct array
   ///
   /// This constructs a record batch using the child arrays of the given
-  /// array, which must be a struct array.  Note that the struct array's own
-  /// null bitmap is not reflected in the resulting record batch.
+  /// array, which must be a struct array.
+  ///
+  /// \param[in] pool the memory pool to allocate new validity bitmaps
```
jorisvandenbossche, line 87: Because you added the `pool` param documentation, it now seems to complain that `array` is not documented

cpp/src/arrow/record_batch.cc
```diff
@@ -198,14 +198,20 @@ Result<std::shared_ptr<RecordBatch>> RecordBatch::MakeEmpty(
 }
 
 Result<std::shared_ptr<RecordBatch>> RecordBatch::FromStructArray(
-    const std::shared_ptr<Array>& array) {
+    const std::shared_ptr<Array>& array, MemoryPool* memory_pool) {
```
felipecrv, line 201: It  might be a good idea to review the callsites to this function and ensure they pass the context's memory pool instead of just passing the default one automatically.

westonpace: Good catch.  Fortunately it seems we don't use this method too much.  I forwarded the parameter to two call sites in the JSON reader (although I think we will always be zero-copy in those cases it doesn't hurt).  There is one more call site in Datum::MakeExecBatch but it was already missing a memory pool (and implicitly using the default in other calls, e.g. cast) so I just left it as is.  I'm also not sure this code path is used anywhere.

PR #34693, opened 2023-03-22
GH-34638: [C++][Docs] Add documentation for minimal build flags

### LINKED ISSUE #34638: [C++][Docs] Add documentation for minimal build flags

It would be useful to publicly document which cmake flags are necessary for producing a minimal build of the C++ library - i.e. nothing but basic Arrow/IPC (and possibly Parquet) support. There would be some extra dependencies for test-enabled builds so we might want to make note of those as well.

### Component(s)

C++, Documentation

benibus:
### Rationale for this change

There's an ongoing effort to support smaller builds of Arrow C++ for simple format integration so explicitly documenting the necessary flags would be useful. Also, some recent changes to the build system may require disclosure.

Adds documentation to `cpp/building.rst` regarding:
- Minimal cmake flags for basic format support
- Recent changes to `ARROW_COMPUTE` and how compute kernels are included in builds.

### Are these changes tested?

N/A

N/A
* Closes: #34638

### DIFF
--- a/docs/source/developers/cpp/building.rst
+++ b/docs/source/developers/cpp/building.rst
@@ -317,7 +317,7 @@ several optional system components which you can opt into building by passing
 boolean flags to ``cmake``.
 
 * ``-DARROW_BUILD_UTILITIES=ON`` : Build Arrow commandline utilities
-* ``-DARROW_COMPUTE=ON``: Computational kernel functions and other support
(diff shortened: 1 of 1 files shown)


westonpace (approved):
This seems like a good explanation to me.  Thanks!

PR #34694, opened 2023-03-22
MINOR: [Go][CI] Increase Go benchmark timeout

zeroshade:
### Rationale for this change
The addition of more Go benchmarks for the RunEndEncoding/Decoding pushed the time taken for the `compute` module's benchmarks over the 20m timeout that we were using for the script, but only on one of the runs (other runs finished just fine but very close to the timeout). So I'm increasing the timeout passed to the `go test -bench` command to avoid it failing.


assignUser (approved):
+1

PR #34697, opened 2023-03-22
GH-34696: [C++] Check REE arrays have no null buffer in Validate()

### LINKED ISSUE #34696: [C++] ValidateRunEndEncoded does not check there is no null buffer

We should make sure we check there is no null buffer at the top-level, otherwise we will attempt to count the nulls in `GetNullCount`, which can crash if the null bitmap is invalid. This situation can happen if corrupt data is read through an IPC reader.

### Component(s)

C++

wjones127:
### Rationale for this change

See #34696

Adds another check. Also removes DCHECKs from `SetData()`, since they are redundant with the `Validate()` checks and I don't think we do that for other arrays. Having them present made it harder to get to the root cause of this error, because `SetData()` is called on the array when validating a record batch.

### Are these changes tested?

* [x] Add tests

**This PR contains a "Critical Fix".** Without this, sending a REE array with an invalid null buffer through IPC to the C++ implementation will cause it to crash if the array is validated.
* Closes: #34696

### DIFF
--- a/cpp/src/arrow/array/array_run_end.cc
+++ b/cpp/src/arrow/array/array_run_end.cc
@@ -71,6 +71,7 @@ void RunEndEncodedArray::SetData(const std::shared_ptr<ArrayData>& data) {
   ARROW_CHECK_EQ(data->type->id(), Type::RUN_END_ENCODED);
   const auto* ree_type =
       internal::checked_cast<const RunEndEncodedType*>(data->type.get());
+  ARROW_CHECK_EQ(data->child_data.size(), 2);
   ARROW_CHECK_EQ(ree_type->run_end_type()->id(), data->child_data[0]->type->id());
   ARROW_CHECK_EQ(ree_type->value_type()->id(), data->child_data[1]->type->id());
 
--- a/cpp/src/arrow/array/array_run_end.cc
+++ b/cpp/src/arrow/array/array_run_end.cc
@@ -74,19 +74,6 @@ void RunEndEncodedArray::SetData(const std::shared_ptr<ArrayData>& data) {
   ARROW_CHECK_EQ(ree_type->run_end_type()->id(), data->child_data[0]->type->id());
   ARROW_CHECK_EQ(ree_type->value_type()->id(), data->child_data[1]->type->id());
 
-  DCHECK_EQ(data->child_data.size(), 2);
-
-  // A non-zero number of logical values in this array (offset + length) implies
-  // a non-zero number of runs and values.
-  DCHECK(data->offset + data->length == 0 || data->child_data[0]->length > 0);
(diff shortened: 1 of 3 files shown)


wjones127:
Alternatively, should we be raising here?

https://github.com/apache/arrow/blob/5bbaf6bade2232a0d60e912defb97579a11db83f/cpp/src/arrow/array/validate.cc#L492-L494

felipecrv (approved):

felipecrv:
Thank you for fixing this @wjones127. 

> Alternatively, should we be raising here?
> 
> https://github.com/apache/arrow/blob/5bbaf6bade2232a0d60e912defb97579a11db83f/cpp/src/arrow/array/validate.cc#L492-L494

@wjones127 as RUN_END_ENCODED is a new type, we could add this layout validation in this switch without breaking existing code as REEs are not used yet. This is our chance :-)

cpp/src/arrow/array/array_run_end.cc
```diff
@@ -74,19 +74,6 @@ void RunEndEncodedArray::SetData(const std::shared_ptr<ArrayData>& data) {
...
-  // At least as many values as run_ends
-  DCHECK_GE(data->child_data[1]->length, data->child_data[0]->length);
-
-  // The null count for run-end encoded arrays is always 0. Actual number of
-  // nulls needs to be calculated through other means.
-  DCHECK_EQ(data->null_count, 0);
-
   Array::SetData(data);
   run_ends_array_ = MakeArray(this->data()->child_data[0]);
   values_array_ = MakeArray(this->data()->child_data[1]);
```
felipecrv, line 79: This will segfault if `child_data.size() < 2`, so maybe we should wrap this in size checks to allow invalid invalid data to go through until `Validate` checks it.

wjones127:
> as RUN_END_ENCODED is a new type, we could add this layout validation in this switch without breaking existing code as REEs are not used yet. This is our chance :-)

So we have added an equivalent check to this within the REE array path:

https://github.com/apache/arrow/blob/4a0fc2868cbb7c2e78f76855ad476c260b92b495/cpp/src/arrow/array/validate.cc#L648-L650

My questions is whether we should enable this for other arrays that have always null buffers, such as NullArray and UnionArray. It seems we don't enforce this for these other types right now in validate, and I'd rather not change that in this PR.

cpp/src/arrow/array/array_run_end.cc
wjones127: Added a check again, but also moved it up since we index into `child_data` in earlier checks.

felipecrv (approved):

jorisvandenbossche (approved):
Looks good!

PR #34705, opened 2023-03-23
GH-34687: [CI][Python] Create job to remove old nightly wheels from gemfury

### LINKED ISSUE #34687: [CI][Python] Set up automated cleaning for old nightly wheels on gemfury

I have been given access to the arrow-nightlies gemfury wheels. At the moment we have ~21600 versions of different wheels uploaded there.
We upload around 30 per day. This should be ~2 years of daily wheels uploaded.

We should automate some clean up. Maybe keep the last 3 months?

### Component(s)

Continuous Integration, Python

raulcd:
### Rationale for this change

We haven't been cleaning nightly wheels from gemfury ever.

Add a new job that should clean them. This is mainly reused from the one on arrow-adbc: https://github.com/apache/arrow-adbc/blob/main/ci/scripts/gemfury_clean.rb

### Are these changes tested?

I will trigger via crossbow

No
* Closes: #34687

### DIFF
--- a/dev/tasks/python-wheels/github.clean.yml
+++ b/dev/tasks/python-wheels/github.clean.yml
@@ -1,0 +1,35 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+{% import 'macros.jinja' as macros with context %}
+
+{{ macros.github_header() }}
+
+jobs:
+  build:
+    name: "Clean old wheels"
+    runs-on: ubuntu-latest
+    steps:
+      {{ macros.github_checkout_arrow()|indent }}
+
+      - name: Clean old wheels
+        shell: bash
+        run: |
+          gem install --user-install gemfury
+          ruby ./ci/scripts/gemfury_clean.rb
+    env:
+      GEMFURY_API_TOKEN: {{ '${{ secrets.CROSSBOW_GEMFURY_TOKEN }}' }}
--- a/dev/tasks/python-wheels/github.clean.yml
+++ b/dev/tasks/python-wheels/github.clean.yml
@@ -27,9 +27,9 @@ jobs:
       {{ macros.github_checkout_arrow()|indent }}
 
       - name: Clean old wheels
+        env:
+          GEMFURY_API_TOKEN: {{ '${{ secrets.CROSSBOW_GEMFURY_API_TOKEN }}' }}
         shell: bash
         run: |
           gem install --user-install gemfury
           ruby ./ci/scripts/gemfury_clean.rb
-    env:
-      GEMFURY_API_TOKEN: {{ '${{ secrets.CROSSBOW_GEMFURY_API_TOKEN }}' }}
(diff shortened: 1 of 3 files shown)


raulcd:
@kou @lidavidm @jorisvandenbossche  before trying (and removing a bunch of wheels) I wanted to validate we are ok with keeping 90 days of nightly pyarrow wheels.

dev/tasks/python-wheels/github.clean.yml
```diff
@@ -0,0 +1,35 @@
...
+    steps:
+      {{ macros.github_checkout_arrow()|indent }}
+
+      - name: Clean old wheels
+        shell: bash
+        run: |
+          gem install --user-install gemfury
+          ruby ./ci/scripts/gemfury_clean.rb
+    env:
+      GEMFURY_API_TOKEN: {{ '${{ secrets.CROSSBOW_GEMFURY_TOKEN }}' }}
```
lidavidm, line 35: Is the crossbow token an upload token, or an API token? Gemfury has both

lidavidm (commented):
I'm ok with 90 days, at least

lidavidm:
Oh, I meant it as "at least I'm ok with this" :) personally I think fewer days is fine, even.

dev/tasks/python-wheels/github.clean.yml
lidavidm: Yeah, we need the other token added then

kou:
I'm OK with 90 days.

dev/tasks/python-wheels/github.clean.yml
```diff
@@ -0,0 +1,35 @@
...
+    steps:
+      {{ macros.github_checkout_arrow()|indent }}
+
+      - name: Clean old wheels
+        shell: bash
+        run: |
+          gem install --user-install gemfury
+          ruby ./ci/scripts/gemfury_clean.rb
+    env:
+      GEMFURY_API_TOKEN: {{ '${{ secrets.CROSSBOW_GEMFURY_API_TOKEN }}' }}
```
assignUser, line 35: nit: It is good security practice to not define envvars with creds globally for a job but rather as close to the usage as possible, so usually in the step they are used in.

assignUser (approved):
+1 I am fine with 90 days but imo we could reduce that further to disincentivize long-term use of nightly wheels/use outside of development. 

I assume we will trigger this with a cron job in crossbow?

Side note: another bit of release infra that is duplicated from one of the arrow subprojects. IMO we should really find a way to avoid all of this C&P

jorisvandenbossche:
I think 90 days is more than long enough, and indeed can probably even be a bit less. The nightly wheels are meant to test development versions, so you should never rely on the exact wheel long term.

raulcd:
I am running this locally first. Just for reference the first wheel on gemfury was:
`Yanked pyarrow 0.16.1.dev342 (created 2020-03-25T17:56:25.572+00:00)`

raulcd:
> I assume we will trigger this with a cron job in crossbow?

This will already be covered on the `wheel-*` group and run as part of the nightly-packaging.

> Side note: another bit of release infra that is duplicated from one of the arrow subprojects. IMO we should really find a way to avoid all of this C&P

Maybe we could open an issue, discuss some approaches and propose something on the ML. Without giving too much thought we could move some of this tooling / release to its own repo and either clone or use a submodule.

raulcd:
@github-actions crossbow submit wheel-clean

raulcd:
The failure is expected until the PR is merged because crossbow can't find the ruby script when cloning arrow/main for cleaning the wheels.

lidavidm:
Yes, it would be good to discuss sharing some of the infrastructure now that we have three/four subprojects using similar infrastructure (not sure how similar arrow-julia is to the rest).

dev/tasks/python-wheels/github.clean.yml
```diff
@@ -0,0 +1,35 @@
...
+    steps:
+      {{ macros.github_checkout_arrow()|indent }}
+
+      - name: Clean old wheels
+        env:
+          GEMFURY_API_TOKEN: {{ '${{ secrets.CROSSBOW_GEMFURY_API_TOKEN }}' }}
+        shell: bash
+        run: |
+          gem install --user-install gemfury
+          ruby ./ci/scripts/gemfury_clean.rb
```
kou, line 35: @raulcd It seems that we need to use `arrow/ci/scripts/gemfury_clean.rb`: https://github.com/ursacomputing/crossbow/actions/runs/4587784310/jobs/8101540480

```text
ruby: No such file or directory -- ./ci/scripts/gemfury_clean.rb (LoadError)
```

PR #34707, opened 2023-03-23
GH-34595: [C++] Update google-cloud-cpp to v2.8.0

### LINKED ISSUE #34595: GCS FileSystem - support federated identity

Hello,

It seems that federated / external identities are not supported in the Arrow GcsFileSystem implementation? It would be great to support this. I'm using it in CI with GitHub as an IdP as per the [instructions](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-google-cloud-platform) and it works great for the Google CLI tools.

I'm not sure how this is implemented, is there a standard library from Google that can just be linked / updated? The federated auth mechanism creates a regular Google application creds file, with the "type" set as "external_account". Is there an easy (ish) way to bring this in if the Google libs handle the different standard credential types? Or is that wishful thinking on my part?

I'm interested in the Python component but guess this would come to other languages that use the same underlying code if it was added.

Here is the stack trace from Arrow:

 File "pyarrow/_fs.pyx", line 571, in pyarrow._fs.FileSystem.get_file_info
pyarrow.lib.ArrowInvalid: google::cloud::Status(INVALID_ARGUMENT: Permanent error GetObjectMetadata: Could not create a OAuth2 access token to authenticate the request. The request was not sent, as such an access token is required to complete the request successfully. Learn more about Google Cloud authentication at https://cloud.google.com/docs/authentication. The underlying error message was: Unsupported credential type (external_account) when reading Application Default Credentials file from [/path/to/credentials.json].). Detail: [errno 22] Invalid argument

### Component(s)

Python

coryan:
Newer versions of `google-cloud-cpp` include support for workload identity federation (aka BYOID, aka federated identity). There are many other features and improvements too.  FWIW, this library increased its major version (from 1.x to 2.x) simply because it dropped support for C++11. Since Arrow already requires C++17, this is a non-issue.

### Rationale for this change

This fixes #34595.  Newer versions of `google-cloud-cpp` support federated identity (and other features).

### Are these changes tested?

N/A. The existing tests should cover this.

No.  There are no public facing changes, nor breaks to public APIs.

* Closes: #34595


coryan:
The test failure for `R / AMD64 Windows C++ RTools 40 ucrt64` seems unrelated, it triggers while building Abseil.

kou (approved):
+1

Thanks!

PR #34720, opened 2023-03-24
GH-34633: [C++][Parquet] Fix StreamReader to read decimals

### LINKED ISSUE #34633: [C++][Parquet] Can't read converted type DECIMAL in Parquet Stream Reader C++ API

I'm trying to read the following field in a parquet file using the Stream Reader C++ API (I didn't create the parquet file):

Schema output:
```text
optional fixed_len_byte_array(16) field_id=-1 processstarted_seconds (Decimal(precision=38, scale=18));
```

```c++
// code
optional<std::array<char, 16>> processstarted_seconds;
... 
stream_reader >> columns_obj.processstarted_seconds;
// code
```

When I run the above code, I get the following runtime error message:
```text
what():  Column converted type mismatch.  Column 'processstarted_seconds' has converted type 'DECIMAL' not 'NONE'
```

The error is originating here: https://github.com/apache/arrow/blob/main/cpp/src/parquet/stream_reader.cc#L158

Note that I'm able to read the FIXED_LEN_BYTE_ARRAY type with parquet::ConvertedType::NONE. For example, the following works fine.
Schema output:
```text
optional fixed_len_byte_array(4) field_id=-1 country_char[4];
```

I'm using Arrow-Parquet C++ v9.0.0.

Thanks for your help.

### Component(s)

C++

wgtmac:
### Rationale for this change

StreamReader only supports reading decimals from int32 and int64. User cannot read decimals from FIXED_LENGTH_BYTE_ARRAY or BYTE_ARRAY.

Support StreamReader to read decimals via arrow::Decimal128 type.

### Are these changes tested?

Added two test cases to read them.

Yes, now user can directly use arrow::Decimal128 to read from parquet file.
* Closes: #34633

### DIFF
--- a/cpp/src/parquet/stream_reader.cc
+++ b/cpp/src/parquet/stream_reader.cc
@@ -174,60 +174,6 @@ StreamReader& StreamReader::operator>>(std::string& v) {
   return *this;
 }
 
-StreamReader& StreamReader::operator>>(::arrow::Decimal128& v) {
-  const auto& node = nodes_[column_index_];
-  ::arrow::Decimal128 value;
-
-  if (node->physical_type() == Type::FIXED_LEN_BYTE_ARRAY) {
-    const int type_length = node->type_length();
-    CheckColumn(Type::FIXED_LEN_BYTE_ARRAY, ConvertedType::DECIMAL, type_length);
-
-    FixedLenByteArray flba;
-    Read(&flba);
-    PARQUET_ASSIGN_OR_THROW(value,
-                            ::arrow::Decimal128::FromBigEndian(flba.ptr, type_length));
-  } else if (node->physical_type() == Type::BYTE_ARRAY) {
-    CheckColumn(Type::BYTE_ARRAY, ConvertedType::DECIMAL);
-
-    ByteArray ba;
-    Read(&ba);
-    PARQUET_ASSIGN_OR_THROW(value, ::arrow::Decimal128::FromBigEndian(ba.ptr, ba.len));
-  } else {
-    ParquetException::NYI("Decimal128 is not implemented for non-binary types");
-  }
-
-  v = std::move(value);
-  return *this;
-}
-
-StreamReader& StreamReader::operator>>(optional<::arrow::Decimal128>& v) {
-  const auto& node = nodes_[column_index_];
-  std::optional<::arrow::Decimal128> value = std::nullopt;
-
-  if (node->physical_type() == Type::FIXED_LEN_BYTE_ARRAY) {
-    const int type_length = node->type_length();
-    CheckColumn(Type::FIXED_LEN_BYTE_ARRAY, ConvertedType::DECIMAL, type_length);
-
-    FixedLenByteArray flba;
-    if (ReadOptional(&flba)) {
-      PARQUET_ASSIGN_OR_THROW(value,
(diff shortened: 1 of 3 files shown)


wgtmac:
@wjones127 @westonpace Could you help review this? Thanks!

cpp/src/parquet/stream_reader_test.cc
```diff
@@ -912,5 +913,39 @@ TEST_F(TestReadingDataFiles, Int64Decimal) {
...
 
+TEST_F(TestReadingDataFiles, FLBADecimal) {
+  PARQUET_ASSIGN_OR_THROW(auto infile, ::arrow::io::ReadableFile::Open(
+                                           GetDataFile("fixed_length_decimal.parquet")));
+
+  auto file_reader = ParquetFileReader::Open(infile);
+  auto reader = StreamReader{std::move(file_reader)};
+
+  ::arrow::Decimal128 x;
+  int i;
```
mapleFU, line 924: `int i = 0;` to avoid reader already eof, and `i` is undefined?

wgtmac: I just follow the same pattern of previous cases. These files are well-defined so i will always be updated.

cpp/src/parquet/stream_reader.cc
```diff
@@ -173,6 +174,60 @@ StreamReader& StreamReader::operator>>(std::string& v) {
   return *this;
 }
 
+StreamReader& StreamReader::operator>>(::arrow::Decimal128& v) {
```
westonpace, line 177: Is there anyway to avoid repeating the implementation here?  For example, would something like this work?

```
StreamReader& StreamReader::operator>>(::arrow::Decimal128& v) {
  std::optional<::arrow::Decimal128> maybe_v;
  this >> maybe_v;
  if (!maybe_v) {
    // I'm not sure what happens here, maybe throw an exception?
  }
  v = *maybe_v;
  return this;
}
```

cpp/src/parquet/stream_reader_test.cc
```diff
@@ -912,5 +913,39 @@ TEST_F(TestReadingDataFiles, Int64Decimal) {
...
+
+  auto file_reader = ParquetFileReader::Open(infile);
+  auto reader = StreamReader{std::move(file_reader)};
+
+  ::arrow::Decimal128 x;
+  int i = 0;
+
+  for (i = 1; !reader.eof(); ++i) {
+    reader >> x >> EndRow;
+    EXPECT_EQ(x.ToString(2), ::arrow::Decimal128(i * 100).ToString(2));
```
westonpace, line 928: Why `ToString`.  Can't we compare the decimals directly?

westonpace (commented):
Nice addition.  I'm a little unfamiliar with this code so these questions may be silly.

cpp/src/parquet/stream_reader.cc
wgtmac: Tried to reuse code and throw if it is null.

cpp/src/parquet/stream_reader_test.cc
wgtmac: Double checked that decimal128 does not have scale. Changed it into comparing decimals directly.

westonpace (approved):
Thanks for this addition!

PR #34730, opened 2023-03-26
GH-34729: [C++][Python] Enhanced Arrow<->Pandas map/pydict support

### LINKED ISSUE #34729: Better support for Maps in Pandas

Today (Py)Arrow -> Pandas treats:

1. structs as Python dicts (pydicts)
2. maps as Python list of tuples (i.e. [(key1, value1), (key2, value2), ...]

Treating maps as a list of tuples has various pros (preserve ordering, allows duplicates, speed of iteration/creation). However, many times users simply want a ... map! (i.e. pydict). 

Having to convert every element manually via `dict(map_elem)` is cumbersome, inefficient, and downright nasty when working with arbitrarily nested maps in Pandas.

Today, Pyarrow already supports (pydicts -> arrow maps) when a schema is provided. So, it's a known use-case. 

I propose a simple bool switch in PandasOptions for `table.to_pandas(...)` to generate pydicts for maps. This creates a symmetrical behavior for the (pydict -> arrow maps) flow, as well.

----

As alluded to above, the cons are that:
1. Users can lose (non) ordering.
2. Duplicates will be removed, resulting in potential data loss. This should be made clear to the user.
3. Potential ambiguity when examining data that has both maps and structs

I think the upsides of ergonomic flexibility outweigh these cons.

----

Related, I think there's a bug that precludes (pydicts -> arrow maps) when the type is nested (e.g. list of maps). That should be fixed as well to provide a more featureful map experience.

### Component(s)

C++, Python

mikelui:
### Rationale for this change

Explained in issue #34729 

- Add support for list of maps when converting Arrow to Pandas. There doesn't seem to be a strong reason to omit this. Previously it was a hard error as unsupported, due to a bool check.
- Refactor Arrow Map -> Pandas to support two paths: (1) list of tuples, or (2) pydicts
- Add another option in PandasOptions to enable (2), above
- Bugfix in nested pydicts -> Arrow maps. 
- Unit tests

### Are these changes tested?

Unit tests are added in `test_pandas.py`

- An additional option flag in PandasOptions
- Enable list of maps to Pandas, which was previously disabled
* Closes: #34729

### DIFF
--- a/python/pyarrow/src/arrow/python/arrow_to_pandas.cc
+++ b/python/pyarrow/src/arrow/python/arrow_to_pandas.cc
@@ -894,7 +894,8 @@ Status CheckMapAsPydictsError() {
     PyErr_Fetch(&type, &value, &traceback);
     std::string message;
     RETURN_NOT_OK(internal::PyObject_StdStringStr(value, &message));
-    message += ". If keys are not hashable, then you must use `maps_as_pydicts=False`";
+    message += ". If keys are not hashable, then you must use the option "
+        "[maps_as_pydicts=False (default)]";
 
     // resets the error
     PyErr_SetString(PyExc_TypeError, message.c_str());
@@ -984,7 +985,7 @@ Status ConvertMap(PandasOptions options, const ChunkedArray& data,
           // returns -1 if there are internal errors around hashing/resizing
           return setitem_result == 0 ?
             Status::OK() :
-            Status::UnknownError(
+            Status::UnknownError("[maps_as_pydicts=True] "
                 "Unexpected failure inserting Arrow (key, value) pair into Python dict"
             );
         },
@@ -999,9 +1000,10 @@ Status ConvertMap(PandasOptions options, const ChunkedArray& data,
         out_values);
 
     if (status.ok() && (total_dict_len < total_raw_len)) {
-      const char* message = "After conversion of Arrow map to pydict, "
-          "detected possible data loss due to duplicate keys. "
-          "Input length is [%lld], total pydict length is [%lld].";
+      const char* message = "[maps_as_pydicts=True] "
+          "After conversion of Arrow maps to pydicts, "
+          "detected data loss due to duplicate keys. "
+          "Original input length is [%lld], total converted pydict length is [%lld].";
       std::array<char, 256> buf;
       std::snprintf(buf.data(), buf.size(), message, total_raw_len, total_dict_len);
       ARROW_LOG(WARNING) << buf.data();
--- a/python/pyarrow/src/arrow/python/arrow_to_pandas.cc
+++ b/python/pyarrow/src/arrow/python/arrow_to_pandas.cc
@@ -878,11 +878,10 @@ Status ConvertMapHelper(
 }
 
 // A more helpful error message around TypeErrors that may stem from unhashable keys
-Status CheckMapAsPydictsError() {
+Status CheckMapAsPydictsTypeError() {
   if (ARROW_PREDICT_TRUE(!PyErr_Occurred())) {
     return Status::OK();
   }
-
   if (PyErr_ExceptionMatches(PyExc_TypeError)) {
     // Modify the error string directly, so it is re-raised
     // with our additional info.
@@ -895,12 +894,32 @@ Status CheckMapAsPydictsError() {
     std::string message;
     RETURN_NOT_OK(internal::PyObject_StdStringStr(value, &message));
     message += ". If keys are not hashable, then you must use the option "
-        "[maps_as_pydicts=False (default)]";
+        "[maps_as_pydicts=None (default)]";
 
     // resets the error
     PyErr_SetString(PyExc_TypeError, message.c_str());
   }
   return ConvertPyError();
+}
+
+Status CheckForDuplicateKeys(bool error_on_duplicate_keys,
+                             Py_ssize_t total_dict_len, Py_ssize_t total_raw_len) {
+  if (total_dict_len < total_raw_len) {
+    const char* message =
+        "[maps_as_pydicts] "
+        "After conversion of Arrow maps to pydicts, "
+        "detected data loss due to duplicate keys. "
+        "Original input length is [%lld], total converted pydict length is [%lld].";
+    std::array<char, 256> buf;
+    std::snprintf(buf.data(), buf.size(), message, total_raw_len, total_dict_len);
+
+    if (error_on_duplicate_keys) {
+      return Status::UnknownError(buf.data());
+    } else {
+      ARROW_LOG(WARNING) << buf.data();
+    }
+  }
+  return Status::OK();
 }
 
 Status ConvertMap(PandasOptions options, const ChunkedArray& data,
@@ -947,7 +966,7 @@ Status ConvertMap(PandasOptions options, const ChunkedArray& data,
   PyArrayObject* py_keys = reinterpret_cast<PyArrayObject*>(owned_numpy_keys.obj());
   PyArrayObject* py_items = reinterpret_cast<PyArrayObject*>(owned_numpy_items.obj());
 
-  if (!options.maps_as_pydicts) {
+  if (options.maps_as_pydicts == MapConversionType::DEFAULT) {
     // The default behavior to express an Arrow MAP as a list of [(key, value), ...] pairs
     OwnedRef list_item;
     return ConvertMapHelper(
@@ -972,6 +991,18 @@ Status ConvertMap(PandasOptions options, const ChunkedArray& data,
     Py_ssize_t total_dict_len{0};
     Py_ssize_t total_raw_len{0};
 
+    bool error_on_duplicate_keys;
+    if (options.maps_as_pydicts == MapConversionType::LOSSY) {
+      error_on_duplicate_keys = false;
+    } else if (options.maps_as_pydicts == MapConversionType::STRICT) {
+      error_on_duplicate_keys = true;
+    } else {
+      auto val = std::underlying_type_t<MapConversionType>(options.maps_as_pydicts);
+      return Status::UnknownError(
+          "Received unknown option for maps_as_pydicts: " + std::to_string(val)
+      );
+    }
+
     auto status = ConvertMapHelper(
         [&dict_item, &total_raw_len](int64_t num_pairs) {
           total_raw_len += num_pairs;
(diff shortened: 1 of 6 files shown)


python/pyarrow/src/arrow/python/python_to_arrow.cc
```diff
@@ -762,7 +762,7 @@ class PyListConverter : public ListConverter<T, PyConverter, PyConverterTrait> {
       RETURN_NOT_OK(AppendSequence(value));
     } else if (PySet_Check(value) || (Py_TYPE(value) == &PyDictValues_Type)) {
       RETURN_NOT_OK(AppendIterable(value));
-    } else if (PyDict_Check(value) && this->options_.type->id() == Type::MAP) {
+    } else if (PyDict_Check(value) && this->type()->id() == Type::MAP) {
```
mikelui, line 765: It looks like `options_` isn't updated/set as expected in `MakeConverter`

So, when there's a nested map (e.g. list of map):
1. `type_` is updated to be the list's value_type, but
2. `option_` is passed along unchanged and still has its `type_` set as the list

That causes this check to fail unexpected, since the pydict is seen as having `options_.type->id() == Type::LIST` instead of `type()->id() == Type::MAP`

I am unsure what the expected behavior of `options_` is, but this check certainly is meant to check the current type.

python/pyarrow/src/arrow/python/arrow_to_pandas.cc
```diff
@@ -176,6 +176,7 @@ static inline bool ListTypeSupported(const DataType& type) {
     case Type::DATE32:
     case Type::DATE64:
     case Type::STRUCT:
+    case Type::MAP:
```
mikelui, line 179: There didn't seem to be a strong reason to have this disabled. Users can do nested maps via (list of structs of maps)

python/pyarrow/src/arrow/python/arrow_to_pandas.cc
```diff
@@ -808,52 +808,20 @@ Status ConvertListsLike(PandasOptions options, const ChunkedArray& data,
...
-    const auto& map_arr = checked_cast<const MapArray&>(*data.chunk(c));
-    key_arrays.emplace_back(map_arr.keys());
-    item_arrays.emplace_back(map_arr.items());
-  }
-
-  const auto& map_type = checked_cast<const MapType&>(*data.type());
-  auto key_type = map_type.key_type();
-  auto item_type = map_type.item_type();
+template<typename F1, typename F2, typename F3>
+Status ConvertMapHelper(
```
mikelui, line 812: this change makes sense when combined with the next commit

amol-:
@jorisvandenbossche would you mind having a look at this one? 

I'm not sure how sold I am on a lossy conversion, but it's also true that it's an option so it's user choice.

AlenkaF:
Thank you for the contribution @mikelui!

The Python part of the code looks good to me. There are two long lines to be corrected:

```python
/arrow/python/pyarrow/tests/test_pandas.py:4610:89: E501 line too long (95 > 88 characters)
/arrow/python/pyarrow/tests/test_pandas.py:4627:89: E501 line too long (92 > 88 characters)
```

As for the C++ part and the question if this is a good option to have or not, Joris will probably have a better comment and ideas.

python/pyarrow/src/arrow/python/arrow_to_pandas.cc
wjones127: Agreed. Although it's interesting we check the value type for list, but not the fields of structs? I was thinking you should put `Type::MAP` down with `Type::LIST`, since it is a subclass of ListType, though since the immediate child is `Type::STRUCT` right now it won't do anything meaningful.

python/pyarrow/tests/test_pandas.py
```diff
@@ -4542,3 +4569,48 @@ def test_does_not_mutate_timedelta_nested():
     df = table.to_pandas()
 
     assert df["timedelta_2"][0].to_pytimedelta() == timedelta_2[0]
+
+
+def test_roundtrip_nested_map_table_with_pydicts():
```
wjones127, line 4574: This is a bit paranoid, but could you also test the `to_pandas` conversion on a MapArray that has been sliced (or even a chunked one where each chunk has been sliced)? That often brings out some bugs in these implementations.

python/pyarrow/src/arrow/python/arrow_to_pandas.cc
```diff
@@ -911,6 +877,97 @@ Status ConvertMap(PandasOptions options, const ChunkedArray& data,
...
+    OwnedRef dict_item;
+    return ConvertMapHelper(
+        [&dict_item]([[maybe_unused]] int64_t) {
+          dict_item.reset(PyDict_New());
+          return CheckPyError();
+        },
+        [&dict_item]([[maybe_unused]] int64_t idx, OwnedRef& key_value, OwnedRef& item_value) {
+          auto setitem_result =
+              PyDict_SetItem(dict_item.obj(), key_value.obj(), item_value.obj());
+          RETURN_IF_PYERROR();
```
wjones127, line 954: This should handle if `TypeError` is raised in case `key_value.obj()` isn't hashable? Perhaps we should test that? I think having a key type of a map or list would trigger that.

python/pyarrow/src/arrow/python/python_to_arrow.cc
wjones127: Agreed. `MakeConverter` doesn't seem to expect that options is specific to the type in question; it's templated over it. This change seems correct to me. Thanks for catching it. 👍

wjones127 (changes requested):
This looks like a very nice change. There are two additional tests I'd like to see, and then I think this is ready to go.

python/pyarrow/src/arrow/python/arrow_to_pandas.cc
```diff
@@ -911,6 +877,122 @@ Status ConvertMap(PandasOptions options, const ChunkedArray& data,
   return Status::OK();
 }
 
+// A more helpful error message around TypeErrors that may stem from unhashable keys
+Status CheckMapAsPydictsError() {
```
mikelui, line 881: Manually tested this by remove the `with pytest.raises(TypeError):` in my non-hashable key test, and got:

```
E   TypeError: unhashable type: 'numpy.ndarray'. If keys are not hashable, then you must use `maps_as_pydicts=False`
```

python/pyarrow/src/arrow/python/arrow_to_pandas.cc
mikelui: This is already correctly handled and converted here: https://github.com/kszucs/arrow/blob/master/cpp/src/arrow/python/common.cc#L144

But, I added an additional check and error message, anyway, to provide more direct information to users so that they don't have to google and curiously search through docs for the common case.

python/pyarrow/tests/test_pandas.py
mikelui: Added some additional tests in 0939fa65ab99e709f0ec4f3bf2609f4f353b218b

The actual scenario was a bit difficult to generate. Let me know if the tests are sufficient or if they need more refinement.

mikelui:
What's the right way to handle the errors in:  https://github.com/apache/arrow/actions/runs/4708295556/jobs/8350646115?pr=34730

It looks like newer versions of pandas have modified the testing utilities to avoid this error, and those tests do seem to pass.

mikelui:
To address @amol- 's concerns about data loss, I added a check and warning.

With the new tests that have no duplicate keys, there are no warnings output.

With the test that does have duplicate keys:

```
arrow/python/pyarrow/src/arrow/python/arrow_to_pandas.cc:1006: After conversion of Arrow map to pydict, detected possible data loss due to duplicated keys. Input length is [3], total pydict length is [2].
```

mikelui:
Hmmmm I wonder if it would be more ergonomic to have users supply an optional string, instead of a bool for `maps_as_pydicts`.

That way, users can supply:
* None: default behavior
* "lossy": report warnings when duplicate keys detected
* "strict": if duplicate keys are detected, we raise a python error

I will experiment with this as a new commit.

mikelui:
@wjones127 I've clarified and improved some of the related error logic, and added some extra tests for it. Can you take a look and provide any feedback when you have time?

python/pyarrow/tests/test_pandas.py
```diff
@@ -2128,6 +2128,91 @@ def test_to_list_of_structs_pandas(self):
...
+                         pa.list_(pa.string()))
+        maps = pa.MapArray.from_arrays(offsets, keys, items)
+        data = pa.ListArray.from_arrays([0, 1, 3], maps)
+
+        expected = pd.Series([
+            [[('foo', ['a', 'b']), ('bar', ['c', 'd'])]],
+            [[('baz', []), ('qux', None), ('quux', [None, 'e'])], [('quz', ['f', 'g'])]]
+        ])
+
+        series = pd.Series(data.to_pandas())
```
wjones127, line 2149: I think this could be simplified:

```suggestion
        data = [
            [[('foo', ['a', 'b']), ('bar', ['c', 'd'])]],
            [[('baz', []), ('qux', None), ('quux', [None, 'e'])], [('quz', ['f', 'g'])]]
        ]
        arr = pa.array(data, pa.list_(pa.map_(pa.utf8(), pa.list_(pa.utf8()))))
        series = arr.to_pandas()

        expected = pd.Series(data)
```

python/pyarrow/tests/test_pandas.py
```diff
@@ -2128,6 +2128,91 @@ def test_to_list_of_structs_pandas(self):
...
+        )
+
+        map_chunk_1 = pa.MapArray.from_arrays([0, 2], keys_1, items_1)
+        map_chunk_2 = pa.MapArray.from_arrays([0, 3], keys_2, items_2)
+        map_chunk_3 = pa.MapArray.from_arrays([0, 1], keys_3, items_3)
+        data = pa.chunked_array([
+            pa.ListArray.from_arrays([0, 1], map_chunk_1),
+            pa.ListArray.from_arrays([0, 1], map_chunk_2),
+            pa.ListArray.from_arrays([0, 1], map_chunk_3),
+        ])
```
wjones127, line 2193: A couple thoughts:

 1. I think the chunked array is overcomplicated it in the end (sorry for that distraction). What's more important is the slicing.
 2. This currently doesn't slice within any arrays; it slices at array boundaries, which doesn't trigger the code paths I was hoping for.
 3. It would also be good to slice the inner arrays.

Here's my suggested test case:

```suggestion
        keys = pa.array(['ignore', 'foo', 'bar', 'baz', 'qux', 'quux', 'ignore']).slice(1, 5)
        items = pa.array(
            [['ignore'], ['ignore'], ['a', 'b'], ['c', 'd'], [], None, [None, 'e']],
            pa.list_(pa.string()),
        ).slice(2, 5)
        map = pa.MapArray.from_arrays([0, 2, 4], keys, items)
        arr = pa.ListArray.from_arrays([0, 1, 2], map)
```

wjones127 (changes requested):
Thanks for the update. Just a few more comments.

wjones127 (approved):
Thanks for working on this @mikelui . LGTM

mikelui:
Looks like some merge issue. Is this an external issue or is it something weird about the HEAD I was working off of?

https://gist.github.com/mikelui/95d23d440974d49c80ab354764e5a39b

PR #34732, opened 2023-03-26
GH-34731: [Python] Release GIL when creating RecordBatchReader

### LINKED ISSUE #34731: [Python] MetadataRecordBatchReader does not release GIL in to_reader() method

`MetadataRecordBatchReader.to_reader()` does not release GIL as it creates the RecordBatchReader via call to `MakeRecordBatchReader`. 

`MakeRecordBatchReader` waits for first batch of data to arrive before it returns the RBR => interpreter will essentially 'stop' until the data is available. This can have 'interesting' consequences especially if used on server...

Workaround: first call `stream.schema` - this waits for the first batch of data while not holding GIL, then follow with `stream.to_reader()`

### Component(s)

FlightRPC, Python

lupko:
### Rationale for this change

See #34731

Wrap existing call that creates RBR in `with nogil:` context

### Are these changes tested?

No new tests. 

No

* Closes: #34731

### DIFF
--- a/python/pyarrow/_flight.pyx
+++ b/python/pyarrow/_flight.pyx
@@ -1039,7 +1039,9 @@ cdef class _MetadataRecordBatchReader(_Weakrefable, _ReadPandasMixin):
         """
         cdef RecordBatchReader reader
         reader = RecordBatchReader.__new__(RecordBatchReader)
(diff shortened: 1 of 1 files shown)


lidavidm (approved):
Thank you!

lidavidm:
Failure is #34743

PR #34733, opened 2023-03-26
GH-34724: [Go] Add Table SetColumn method

### LINKED ISSUE #34724: [Go][Table] Data Manipulation

Looks like the python implementation has some nice [helper](https://arrow.apache.org/cookbook/py/data.html#adding-a-column-to-an-existing-table) functions such as `concat_table`, `append_column`, `set_column` would it be useful/make sense for the Go implementation?

Subtasks:

- [x] #34832

### Component(s)

Go

yevgenypats:
Small method similar to the c/c++ implementation.  

* Closes: #34724

### DIFF
--- a/go/arrow/array/table.go
+++ b/go/arrow/array/table.go
@@ -217,6 +217,16 @@ func (tbl *simpleTable) validate() {
 	}
 }
 
+func (tbl *simpleTable) SetColumn(i int, col arrow.Column) error {
+	if tbl.cols[i].DataType() != col.DataType() {
+		return fmt.Errorf("arrow/array: column %q expected type %q but got type %q", col.Name(), tbl.cols[i].DataType(), col.DataType())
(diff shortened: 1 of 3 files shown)


zeroshade (commented):
This looks great to me, can you add a test that just verifies that `SetColumn` actually replaced the column (you get the new column back from the `Column` method)

zeroshade:
@yevgenypats As I mentioned on the other PR, can you make issues for these subtasks so that merging this doesn't end up closing the umbrella "Data Table Manipulation" issue? If you update the title of this PR to point at a subtask issue the bot will update the comments. Then I'll merge this, and the automation won't make things confusing.

Sorry for the difficulty and strangeness there.

yevgenypats:
Closing for now as we try to use `Records` directly instead of the table abstraction.

PR #34734, opened 2023-03-26
GH-34657: [Go] Move Scalar interface to arrow

### LINKED ISSUE #34657: [Go][Array] Expose UntypedValue(i int) interface

Similar to https://github.com/apache/arrow/issues/34377

It will be nice to expose `UntypedValue(i int) interface` on array so any subclass array can implement it.

### Component(s)

Go

yevgenypats:
This is a pre-requirement for https://github.com/apache/arrow/issues/34657 as array interface needs access to the Scalar interface and without it, it will cause cyclic loop.

cc @zeroshade 
* Closes: #34657

### DIFF
--- a/go/arrow/array.go
+++ b/go/arrow/array.go
@@ -23,6 +23,31 @@ import (
 	"github.com/apache/arrow/go/v12/arrow/memory"
 )
 
+// Scalar represents a single value of a specific DataType as opposed to
+// an array.
+//
+// Scalars are useful for passing single value inputs to compute functions
+// (not yet implemented) or for representing individual array elements,
+// (with a non-trivial cost though).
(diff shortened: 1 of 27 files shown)


zeroshade:
I don't think this resolves the issue unfortunately, which is why the existing `GetScalar` method is in the Scalar package. Even if you put the `Scalar` interface into the `arrow` package, if you try to put `GetScalar` as a member of the `Array` objects you'll still run into the issue of trying to actually construct your `Scalar` objects. You'll still run into the cyclical import if you try to call any of the `MakeScalar` methods. So I don't think that this is really worthwhile since it doesn't solve the cyclical import problem.

PR #34735, opened 2023-03-27
MINOR: [C++] Fix build issue where wrong variable checked for re2 prefix dir

mikelui:
### Rationale for this change

CMake variables are case sensitive.

When providing a manual prefix dir for re2 (`re2_ROOT`), the wrong variable (`RE2_ROOT`) would be checked, causing build issues.

typo fix

### Are these changes tested?

Built on my end by providing `-Dre2_ROOT=...`

Possibly if a user decided to set `RE2_ROOT` to something different. Unlikely, though.


kou (approved):
+1

Good catch!

PR #34745, opened 2023-03-27
MINOR: [Python] Respect BUILD_SHARED_LIBS cmake options

mikelui:
### Rationale for this change

Prior to 240ebb75b57bb05551c9103ec3dee11c23fd0aca , both dynamic and static libraries would be built.

After, only dynamic shared libraries were being built.

This change at least gives the user the option to build dynamic or static libs with the `BUILD_SHARED_LIBS` cmake variable

See: https://cmake.org/cmake/help/latest/command/add_library.html

CMake build change

### Are these changes tested?

Yes, build with default value, and `-DBUILD_SHARED_LIBS=OFF`

For the former, got dynamic libarrow_python libs, for the latter got static libarrow_python archive

No, default stays the same

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -358,7 +358,7 @@ if(NOT PYARROW_CPP_LINK_LIBS)
   endif()
 endif()
 
-add_library(arrow_python SHARED ${PYARROW_CPP_SRCS})
+add_library(arrow_python ${PYARROW_CPP_SRCS})
 target_include_directories(arrow_python PUBLIC ${PYARROW_CPP_ROOT_DIR}
                                                ${CMAKE_CURRENT_BINARY_DIR}/pyarrow/src)
 if(NOT CMAKE_VERSION VERSION_LESS 3.16)


kou:
This is not a minor change. Could you open an issue for this pull request and explain your use case for this?

See also:

> If this is not a [minor PR](https://github.com/apache/arrow/blob/main/CONTRIBUTING.md#Minor-Fixes). Could you open an issue for this pull request on GitHub? https://github.com/apache/arrow/issues/new/choose
> 
> Opening GitHub issues ahead of time contributes to the [Openness](http://theapacheway.com/open/#:~:text=Openness%20allows%20new%20users%20the,must%20happen%20in%20the%20open.) of the Apache Arrow project.
> 
> Then could you also rename the pull request title in the following format?
> 
> ```
> GH-${GITHUB_ISSUE_ID}: [${COMPONENT}] ${SUMMARY}
> ```
>
> ...
> 
> See also:
> 
>    * [Other pull requests](https://github.com/apache/arrow/pulls/)
> 
>    * [Contribution Guidelines - How to contribute patches](https://arrow.apache.org/docs/developers/contributing.html#how-to-contribute-patches)

mikelui:
deciding this isn't valuable, for now

PR #34748, opened 2023-03-27
GH-34724: [Go] Add Schema AddField

### LINKED ISSUE #34724: [Go][Table] Data Manipulation

yevgenypats:
Similar to cpp [`AddField`](https://arrow.apache.org/docs/cpp/api/datatype.html#_CPPv4NK5arrow10StructType8AddFieldEiRKNSt10shared_ptrI5FieldEE) API and [`AddColumn`](https://arrow.apache.org/docs/cpp/api/table.html#_CPPv4NK5arrow11RecordBatch9AddColumnEiRKNSt10shared_ptrI5FieldEERKNSt10shared_ptrI5ArrayEE)
* Closes: #34724

### DIFF
--- a/go/arrow/array/table.go
+++ b/go/arrow/array/table.go
@@ -92,8 +92,6 @@ type simpleTable struct {
 	schema *arrow.Schema
 }
 
-var _ arrow.Table = (*simpleTable)(nil)
-
 // NewTable returns a new basic, non-lazy in-memory table.
 // If rows is negative, the number of rows will be inferred from the height
 // of the columns.
--- a/go/arrow/array/table.go
+++ b/go/arrow/array/table.go
@@ -92,6 +92,8 @@ type simpleTable struct {
 	schema *arrow.Schema
 }
 
+var _ arrow.Table = (*simpleTable)(nil)
+
 // NewTable returns a new basic, non-lazy in-memory table.
 // If rows is negative, the number of rows will be inferred from the height
 // of the columns.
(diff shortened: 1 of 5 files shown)


go/arrow/array/table.go
```diff
@@ -198,6 +198,26 @@ func NewTableFromRecords(schema *arrow.Schema, recs []arrow.Record) *simpleTable
 }
 
 func (tbl *simpleTable) Schema() *arrow.Schema      { return tbl.schema }
+
+func (tbl *simpleTable) AddColumn(i int, field arrow.Field, column arrow.Column) (*simpleTable, error) {
```
zeroshade, line 202: Would `InsertColumn` be a better name? You also need to add this to the `Table` interface.

When you do, please add a documentation comment to clarify the semantics that this creates a new table that needs to be released instead of modifying the existing table.

go/arrow/schema.go
```diff
@@ -222,6 +222,18 @@ func (sc *Schema) Equal(o *Schema) bool {
 	return true
 }
 
+func (s *Schema) AddField(i int, field Field) (*Schema, error) {
```
zeroshade, line 225: Like with the table, would `InsertField` be a better name? Personally I see `Add` and immediately think of adding to the end rather than inserting.

zeroshade:
@yevgenypats that's a good point. If the functionality is the same as the C++ one, I agree that keeping the same name would make it easier for someone to transition from one to the other and we can keep it.

yevgenypats:
@zeroshade is it good to go then?

zeroshade (changes requested):
Just add docstring comments on the new functions that explain them and clearly explain the semantics that it creates a new table / new schema rather than modifying the one it is called on. 

LGTM otherwise

zeroshade (changes requested):
Just one issue to be fixed then we're good to merge

go/arrow/array/table.go
```diff
@@ -92,6 +92,8 @@ type simpleTable struct {
 	schema *arrow.Schema
 }
 
+var _ arrow.Table = (*simpleTable)(nil)
```
disq, line 95: This is a compile-time check of `simpleTable` against the `arrow.Table` interface.

zeroshade (approved):

PR #34765, opened 2023-03-29
GH-14917: [C++] Error out when GTest is compiled with a C++ standard lower than 17

### LINKED ISSUE #14779: [C++] Compiling failed on Mac M1

I am getting the following error while compiling Arrow C++. 

```bash
8 warnings generated.
[537/632] Linking CXX executable debug/arrow-utility-test
FAILED: debug/arrow-utility-test 
: && /Users/vibhatha/tools/arrow/lint/clang+llvm-14.0.0-x86_64-apple-darwin/bin/clang++ -fno-aligned-new  -Qunused-arguments -fcolor-diagnostics -ggdb -O0  -Wall -Wextra -Wdocumentation -Wshorten-64-to-32 -Wno-missing-braces -Wno-unused-parameter -Wno-constant-logical-operand -Wno-return-stack-address -Werror -Wno-unknown-warning-option -Wno-pass-failed -march=armv8-a  -g -arch arm64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.3.sdk -mmacosx-version-min=12.0 -Wl,-search_paths_first -Wl,-headerpad_max_install_names -L/opt/homebrew/opt/bison/lib src/arrow/util/CMakeFiles/arrow-utility-test.dir/align_util_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/atfork_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/byte_size_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/cache_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/checked_cast_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/compression_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/decimal_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/formatting_util_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/key_value_metadata_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/hashing_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/int_util_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/io_util_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/iterator_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/logging_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/queue_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/range_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/reflection_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/small_vector_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/stl_util_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/string_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/tdigest_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/test_common.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/time_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/tracing_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/trie_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/uri_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/utf8_util_test.cc.o src/arrow/util/CMakeFiles/arrow-utility-test.dir/value_parsing_test.cc.o -o debug/arrow-utility-test  -Wl,-rpath,/Users/vibhatha/github/fork/arrow/cpp/build/debug -Wl,-rpath,/opt/homebrew/lib  debug/libarrow_testing.1100.0.0.dylib  debug/libarrow.1100.0.0.dylib  /opt/homebrew/opt/openssl@1.1/lib/libcrypto.dylib  /opt/homebrew/opt/openssl@1.1/lib/libssl.dylib  /opt/homebrew/lib/libbrotlienc.dylib  /opt/homebrew/lib/libbrotlidec.dylib  /opt/homebrew/lib/libbrotlicommon.dylib  /opt/homebrew/lib/libaws-cpp-sdk-config.dylib  /opt/homebrew/lib/libaws-cpp-sdk-transfer.dylib  /opt/homebrew/lib/libaws-cpp-sdk-identity-management.dylib  /opt/homebrew/lib/libaws-cpp-sdk-cognito-identity.dylib  /opt/homebrew/lib/libaws-cpp-sdk-sts.dylib  /opt/homebrew/lib/libaws-cpp-sdk-s3.dylib  /opt/homebrew/lib/libaws-cpp-sdk-core.dylib  utf8proc_ep-install/lib/libutf8proc.a  /opt/homebrew/lib/libre2.9.0.0.dylib  googletest_ep-prefix/lib/libgtest_maind.dylib  googletest_ep-prefix/lib/libgtestd.dylib  googletest_ep-prefix/lib/libgmockd.dylib  /opt/homebrew/opt/openssl@1.1/lib/libcrypto.dylib  /opt/homebrew/lib/libaws-crt-cpp.dylib  /opt/homebrew/lib/libaws-c-mqtt.1.0.0.dylib  /opt/homebrew/lib/libaws-c-event-stream.1.0.0.dylib  /opt/homebrew/lib/libaws-c-s3.1.0.0.dylib  /opt/homebrew/lib/libaws-c-auth.1.0.0.dylib  /opt/homebrew/lib/libaws-c-http.1.0.0.dylib  /opt/homebrew/lib/libaws-c-io.1.0.0.dylib  /opt/homebrew/lib/libaws-c-compression.1.0.0.dylib  /opt/homebrew/lib/libaws-c-cal.1.0.0.dylib  -framework Security  /opt/homebrew/lib/libaws-c-sdkutils.1.0.0.dylib  /opt/homebrew/lib/libaws-checksums.1.0.0.dylib  /opt/homebrew/lib/libaws-c-common.1.0.0.dylib  -pthread  -lpthread  -framework CoreFoundation && :
  "__ZNSt3__18to_charsEPcS0_d", referenced from:
      __ZN5arrow8internal7ToCharsIdJEEENSt3__112basic_stringIcNS2_11char_traitsIcEENS2_9allocatorIcEEEET_DpOT0_ in string_test.cc.o
  "__ZNSt3__18to_charsEPcS0_f", referenced from:
      __ZN5arrow8internal7ToCharsIfJEEENSt3__112basic_stringIcNS2_11char_traitsIcEENS2_9allocatorIcEEEET_DpOT0_ in string_test.cc.o
clang-14: error: linker command failed with exit code 1 (use -v to see invocation)
[546/632] Building CXX object src/parquet/CMakeFiles/parquet-arrow-test.dir/arrow/arrow_reader_writer_test.cc.o
ninja: build stopped: subcommand failed.
```
I encounter this issue on Mac M1 while using the clang-14. 

### Component(s)

C++

### LINKED ISSUE #14917: [C++] Issues with GTest when building Arrow C++ on MacOS

HaochengLIU:
### Rationale for this change
On MacOS, the system GTest from brew uses C++14 while arrow is default to C++17.
The build will fail at the linking stage unless users explicity uses bundled GTest.

Users are often confused by unclear linking errors.

At CMake configuration time, the CMake code will automatically detect and 
compare the C++ standard used by GTest and arrow. If a mismatch is detected,
CMake will error out and privide meaningful guidance. It will apply both to
linux and MacOS.

### Are these changes tested?
Covered by Arrow github test worker and tested on local host.

Yes, now users are given better guidance when building Arrow against the system
GTest.

* Closes: #14917
* Closes: #14779

### DIFF
--- a/cpp/cmake_modules/ThirdpartyToolchain.cmake
+++ b/cpp/cmake_modules/ThirdpartyToolchain.cmake
@@ -2230,6 +2230,13 @@ if(ARROW_TESTING)
                      1.10.0
                      USE_CONFIG
                      ${GTEST_USE_CONFIG})
+  get_target_property(gtest_cxx_standard GTest::gtest INTERFACE_COMPILE_FEATURES)
+
+  if((${gtest_cxx_standard} STREQUAL "cxx_std_11") OR (${gtest_cxx_standard} STREQUAL
+                                                       "cxx_std_14"))
+    message(FATAL_ERROR "System GTest is built with a C++ standard lower than 17. Use bundled GTest via passing in CMake flag
+-DGTest_SOURCE=\"BUNDLED\"")
+  endif()
 
   if(GTest_SOURCE STREQUAL "SYSTEM")
     find_package(PkgConfig QUIET)


cpp/cmake_modules/SetupCxxFlags.cmake
HaochengLIU, line 1: arm64 is detected [here](https://github.com/apache/arrow/pull/34765/files#diff-f85a0b8910bc12cd3aa3171bed980e51aefaf30d0fc771ba1621322e973a8d9fR29). Thus it's safe for me to set GTest_SOURCE to be "BUNDLED" when MacOS is detected.

HaochengLIU:
Hi @AlenkaF , after some thoughts, IMO the correct thing here is to fix things at the building level rather than asking every Arrow user to add an extra flag manually in the doc. I have richful experience with CMake thus I go ahead and touch the toolchain file.

I've tested the build settings from https://github.com/apache/arrow/issues/14779 as well as few build configurations with test enabled and they all build smoothly. Looking forward to your thoughts.

Thanks.

AlenkaF:
Thank you for the thoughtful contribution @HaochengLIU !

> Hi @AlenkaF , after some thoughts, IMO the correct thing here is to fix things at the building level rather than asking every Arrow user to add an extra flag manually in the doc. I have richful experience with CMake thus I go ahead and touch the toolchain file.

That would for sure be useful for all that have this issue. My only thought here is that this doesn't seem to be an issue for every MacOS user. For this reason I suggested to first add this to the docs.

The change in the PR looks good to me though (will still wait for the CI and local test to approve it).

@kou what do you think about this change?

cpp/cmake_modules/SetupCxxFlags.cmake
```diff
@@ -115,6 +115,9 @@ elseif(ARROW_CPU_FLAG STREQUAL "aarch64")
   if(ARROW_SIMD_LEVEL STREQUAL "DEFAULT")
     set(ARROW_SIMD_LEVEL "NEON")
   endif()
+  if(APPLE)
+    set(GTest_SOURCE "BUNDLED")
+  endif()
```
kou, line 120: We should not change `*_SOURCE` in `SetupCxxFlags.cmake`.
We should process `*_SOURCE` in `ThirdpartyToolchain.cmake`.

This problem is caused by C++ standard difference, right?
For example, this problem is caused when Apache Arrow C++ uses C++17 and system GoogleTest uses C++14.

Can we detect whether system GoogleTest uses different C++ standard than Apache Arrow C++? If we can detect it, we can fallback to bundled GoogleTest automatically when system GoogleTest uses different C++ standard.

HaochengLIU: Hi Sutou, Good point. I will check CMake and get back to you.

HaochengLIU: After checking the upstream [FindGTest.cmake](https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/FindGTest.cmake) as well as the pkgconfig from GoogleTest on my host, here is my finding:

In order to detect what C++ standard is used by GoogleTest, there is no simple way to do so except by importing it first to my best knowledge. Moreover, once a target is imported, you can not rename or delete it(See Ben's - one of the main CMake developers - comment [here](https://discourse.cmake.org/t/how-to-delete-or-rename-an-imported-target-using-cmake/2372/2?u=haocheng_liu1)). Thus there is no way to fall back to bundled mode.

My new proposed fix is at CMake configuration time, the CMake code will automatically detect and
compare the C++ standard used by GTest and arrow. If a mismatch is detected,
CMake will error out and provide meaningful guidance.  With respect to The principle of least astonishment, I prefer to let the users add these flags by themselves.

```
// sample output from my host
CMake Error at cmake_modules/ThirdpartyToolchain.cmake:2238 (message):
  SYSTEM GTest is built with c++14 but project is built

  with a different standard(17).  Use bundled GTest via passing in CMake flag

  -DGTest_SOURCE="BUNDLED"
Call Stack (most recent call first):
  CMakeLists.txt:505 (include)


-- Configuring incomplete, errors occurred!
```

Looking forward to your thoughts. @kou .

cpp/cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -2230,6 +2230,15 @@ if(ARROW_TESTING)
                      1.10.0
                      USE_CONFIG
                      ${GTEST_USE_CONFIG})
+  get_target_property(gtest_cxx_standard GTest::gtest INTERFACE_COMPILE_FEATURES)
+
+  if(APPLE
+     AND (${gtest_cxx_standard} STREQUAL "cxx_std_14")
```
HaochengLIU, line 2236: GoogleTest just bumped its CMake standard [less then a year ago](https://github.com/google/googletest/blame/main/docs/quickstart-cmake.md) thus we should be fine here

HaochengLIU:
> My only thought here is that this doesn't seem to be an issue for every MacOS user

@AlenkaF I totally agree. With the new change, I believe the doc update is no longer needed since users will do it by themselves with a meaningful error message now?

cpp/cmake_modules/SetupCxxFlags.cmake
kou: Can we use `try_compile()` https://cmake.org/cmake/help/latest/command/try_compile.html for this like we did for gRPC https://github.com/apache/arrow/blob/ddd0a337174e57cdc80b1ee30dc7e787acfc09f6/cpp/src/arrow/flight/CMakeLists.txt#L119-L123 ?

I think that checking `INTERFACE_COMPILE_FEATURES` isn't so portable.
For example, `libgtest-dev` on Debian/Ubuntu still uses `cxx_std_11` not `cxx_std_14`.

For CMake target conflict, we can use different CMake target names (such as `arrow::GTest::gtest`) and CMake variables to refer suitable CMake targets (such as `set(ARROW_GTEST_GTEST arrow::GTest::gtest)`) like we did for Zstandard: https://github.com/apache/arrow/blob/ddd0a337174e57cdc80b1ee30dc7e787acfc09f6/cpp/cmake_modules/ThirdpartyToolchain.cmake#L2577-L2589

kou: We can use `FindXXXAlt.cmake` such as `FindGTestAlt.cmake` to check such additional check.
We can use `FindXXXAlt.cmake` by `resolve_dependency(... HAVE_ALT TRUE)`.

HaochengLIU: Also there is no need to do both try_compile and FindXXXAlt.cmake right? The later one sounds an over engineering solution to me….. it will bloat the cmake code in arrow with limited benefits.

HaochengLIU: > - **I think that checking INTERFACE_COMPILE_FEATURES isn't so portable.
For example, libgtest-dev on Debian/Ubuntu still uses cxx_std_11 not cxx_std_14.**

I agree. As [Arow's minimum CXX_STANDARD requirement is 17](https://github.com/apache/arrow/blob/main/cpp/cmake_modules/SetupCxxFlags.cmake#L128-L130), how about We error out whenever `INTERFACE_COMPILE_FEATURES` from GTest is lower than that?  It will be universal to all platforms.
Ty.

HaochengLIU: (PR updated with item 1) )

kou: The latter.
We can just wrap existing `GTestConfig.cmake` like we did in `FindOpenSSLAlt.cmake`: https://github.com/apache/arrow/blob/main/cpp/cmake_modules/FindOpenSSLAlt.cmake

> In this PR, We limit the scope to error out whenever `INTERFACE_COMPILE_FEATURES` from GTest is lower than that? It will be universal to all platforms and truly detect the issue at CMake configure time.
> 
> Then in another PR/issue, we can discuss how to implement try_compile and FindXXXAlt.cmake as it will need non trivial work.

It makes sense. Could you update the title of this pull request and open a new issue for the follow-up action?

kou: FYI: If we use the latter approach, we can implement an automatically fallback solution instead of raising an error for C++ standard miss match. (If system GoogleTest isn't suitable, we can use our bundled GoogleTest instead of stopping `cmake`.)

HaochengLIU: Thanks for the suggestion @kou , I've updated the title of the PR and opened a new follow up issue. Please take a look when you have time

kou: Thanks!

HaochengLIU: [The failing test ](https://github.com/apache/arrow/actions/runs/4570445331/jobs/8068019026?pr=34765)- test_generic_options - looks like a flaky one

kou (approved):
+1

austin3dickey:
FYI, this PR caused the `cpp-micro` benchmarks to start failing. See linked issue ^

jorisvandenbossche:
Is this limitation necessary on non-MacOS systems? I have been using the conda installed gtest on linux, and don't remember to have run into problems with this

HaochengLIU:
Yes. It's platform agnostic, only depending on which C++ standard gtest is compiled with.

jorisvandenbossche:
But the reported issues all seem to involve people using Mac? As mentioned, I have always been using gtest from conda on linux, and never ran into any issues with it (while now this check avoids a successful compilation)

HaochengLIU:
It depends on how conda builds/ships gtest on linux which I'm not familiar with. When you say "while now this check avoids a successful compilation", does it happen after this https://github.com/apache/arrow/pull/34874?

kou:
We use `INTERFACE_COMPILE_FEATURES` for this approach but it's not perfect because GoogleTest always sets `cxx_std_14` (or `cxx_std_11` in older Google Test): https://github.com/google/googletest/blob/7ee260c54921571b18b15049304426fe151c1265/googletest/cmake/internal_utils.cmake#L204

But GoogleTest may be built with C++17 even when `INTERFACE_COMPILE_FEATURES` is `cxx_std_14` or `cxx_std_11`.
It's occurred when C++ compiler uses C++17 by default. (Recent Linux distributions will use `g++` that uses C++17 by default.)

I'll implement #34813. Please wait for a few days.

FYI: We need a GoogleTest that uses https://github.com/google/googletest/blob/7ee260c54921571b18b15049304426fe151c1265/googletest/include/gtest/internal/gtest-port.h#L2470-L2482

jorisvandenbossche:
> When you say "while now this check avoids a successful compilation", does it happen after this #34874?

No, because when you use ARROW_DEPENDENCY_SOURCE=CONDA, it get actually set to SYSTEM for each of the dependencies (but with the path (ARROW_PACKAGE_PREFIX) pointing to the conda prefix)

@kou thanks for taking a look!

PR #34769, opened 2023-03-29
GH-33976: [Python] Remove usage of TableGroupBy helper in favor of pyarrow.acero.Declaration

jorisvandenbossche:
### Rationale for this change

Now we have the pyarrow.acero building blocks (GH-33976), we can easily construct the Declaration, which `arrow::compute::TableGroupBy` creates under the hood, ourselves in pyarrow. 

### Are these changes tested?

Existing tests are passing.

No


jorisvandenbossche:
@westonpace I don't see the C++ `TableGroupBy` being used anywhere else (except in tests). Now we don't need this anymore for pyarrow, do we still want to keep this? (as convenience for C++ users) Or do I remove this on the C++ side as well?

`BatchGroupBy` is used in a benchmark, but we can easily move the 4-line function to construct the declaration to that file.

westonpace:
> @westonpace I don't see the C++ TableGroupBy being used anywhere else (except in tests). Now we don't need this anymore for pyarrow, do we still want to keep this? (as convenience for C++ users) Or do I remove this on the C++ side as well?

I'm kind of +0 but yes, let's go ahead and remove it.  Having a single interface to Acero is probably easier to maintain long-term.

westonpace (approved):
The python code looks good.  I reviewed some of the group by code again.  I thought it was doing a bit more to patch up the output from the exec plan.  However, I believe I am remembering an older state of the code.  So yes, let's go ahead and remove the C++ side too if you don't mind.

jorisvandenbossche:
On the C++ side, those are publicly exposed in Acero I assume (use of `ARROW_ACERO_EXPORT`). Do we first want to deprecate those, or it is OK to directly remove them given the early stage of Acero?

jorisvandenbossche:
Another question: `groupby_test.cc`, is that just testing the helper interface (and the actual hash aggregation tests are living elsewhere), or is that useful to keep? 
I was first planning to just move the helper defining/executing the Declaration to `groupby_test.cc` to keep it running without exposing the helper function. But if that's not needed, I can just remove the file altogether (I see there are quite some tests involving grouped aggregations in `plan_test.cc`)

westonpace:
@jorisvandenbossche 

If you don't mind, keeping the standalone groupby_test would be helpful.  I want to encourage splitting plan_test up into tests-per-node (see order_by_node_test and fetch_node_test which are similar in size to groupby_test).  Maybe you could rename it to aggregate_node_test?

westonpace (approved):
This looks great.  Thanks a lot!

PR #34777, opened 2023-03-29
MINOR: [C++] Remove std::move when calling ProcessEmit and ProcessExtensionEmit

icexelloss:
### Rationale for this change
I noticed a few places in relation_internal.cc that std::move is used incorrectly. 

* Currently, `std::move` is used when passing arguments to `ProcessEmit` and `ProcessExtensionEmit`, however these two methods takes in const reference and therefore the `std::move` is not needed.
* (orthogonal) Removed unneeded `FromProto` from header

### Are these changes tested?
With existing tests

No

### DIFF
--- a/cpp/src/arrow/engine/substrait/expression_internal.h
+++ b/cpp/src/arrow/engine/substrait/expression_internal.h
@@ -40,11 +40,6 @@ Result<FieldRef> DirectReferenceFromProto(const substrait::Expression::FieldRefe
                                           const ExtensionSet&, const ConversionOptions&);
 
 ARROW_ENGINE_EXPORT
-Result<compute::Expression> FromProto(const substrait::Expression::FieldReference*,
-                                      const ExtensionSet&, const ConversionOptions&,
-                                      std::optional<compute::Expression>);
-
-ARROW_ENGINE_EXPORT
 Result<compute::Expression> FromProto(const substrait::Expression&, const ExtensionSet&,
                                       const ConversionOptions&);
 
--- a/cpp/src/arrow/engine/substrait/relation_internal.cc
+++ b/cpp/src/arrow/engine/substrait/relation_internal.cc
@@ -89,7 +89,7 @@ Result<EmitInfo> GetEmitInfo(const RelMessage& rel,
   }
   emit_info.expressions = std::move(proj_field_refs);
   emit_info.schema = schema(std::move(emit_fields));
-  return emit_info;
+  return std::move(emit_info);
 }
 
 Result<DeclarationInfo> ProcessEmitProject(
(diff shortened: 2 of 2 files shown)


cpp/src/arrow/engine/substrait/expression_internal.h
```diff
@@ -39,11 +39,6 @@ ARROW_ENGINE_EXPORT
 Result<FieldRef> DirectReferenceFromProto(const substrait::Expression::FieldReference*,
                                           const ExtensionSet&, const ConversionOptions&);
 
-ARROW_ENGINE_EXPORT
```
icexelloss, line 42: Per this comment:
https://github.com/apache/arrow/pull/34627#discussion_r1152715709

I noticed and removed it here. Although unrelated to the original PR, I figured this is small enough change to piggy back here. If there are objection I can revert this change and put in separate PR.

cpp/src/arrow/engine/substrait/relation_internal.cc
```diff
@@ -89,7 +89,7 @@ Result<EmitInfo> GetEmitInfo(const RelMessage& rel,
   }
   emit_info.expressions = std::move(proj_field_refs);
   emit_info.schema = schema(std::move(emit_fields));
-  return std::move(emit_info);
```
icexelloss, line 92: I don't think std::move of return value is good practice. (Since this prevents NRVO) so I removed them here.

icexelloss:
@westonpace Minor changes to clean up the code a bit - no rush

cpp/src/arrow/engine/substrait/relation_internal.cc
westonpace: I recall that one of the older compilers (I think mingw) does not recognize that `emit_info` is the return value because the return type is `Result<EmitInfo>` and the type of `emit_info` is `EmitInfo`.  See, for example, https://stackoverflow.com/questions/74459462/why-c-does-not-perform-rvo-to-stdoptional .  However, maybe it was one of the older gcc compilers that we no longer support now that we upgraded to C++17.

We should detect it, if it is a problem, if the nightly tests pass.  However, the existing convention in the code base is still to use `std::move` when the return type is `Result<T>` and the object is `T`.

westonpace (commented):
Thanks.  I definitely agree with removing `std::move` when calling `ProcessEmit`.  I'm not sure about the return statement though (see my reply to your comment).

icexelloss:
> Thanks. I definitely agree with removing `std::move` when calling `ProcessEmit`. I'm not sure about the return statement though (see my reply to your comment).

Got it. I do see in other places where `std::move` is used for return. I guess it is a larger scope discussion whether we want to change that or not. Will revert that part.

cpp/src/arrow/engine/substrait/relation_internal.cc
icexelloss: Got it. I reverted this.

westonpace (approved):
Thanks for the cleanup.

Issue #34058, opened 2023-02-06
[C++] Iteratively fix memory pool passdown and enable memory benchmarks


wjones127:
### Describe the enhancement requested

#15231 enabled memory tracking in benchmarks, but only for components that always used the memory pool passed. Unfortunately, many components will make at least some allocations with the default memory pool, so before we can add memory metrics to their benchmarks we need to fix the implementations first.

#15231 also added a new metric on `MemoryPool`: `num_allocations`. This provides a way to assert that an implementation only uses the passed memory pool. First, wrap the default memory pool in a `ProxyMemoryPool` and record the current `num_allocations` on the global pool. Pass the proxy pool to the function. Then once the function returns, assert that the `num_allocations` on the proxy pool equals the number of allocations on the global pool minus the starting value.

Components:

 * [ ] Parquet read and write
 * [ ] CSV read and write
 * [ ] Acero queries
 * [ ] Dataset scans
 * [ ] Compute functions

### Component(s)

Benchmarking, C++

thisisnic:
This issue has been marked as stale because it has had no activity in the past 2 years. Please remove the stale label or comment below, or this issue will be closed in 14 days.

PR #34803, opened 2023-03-30
GH-34802: [C++][Parquet] Allow passing pool to decoder

### LINKED ISSUE #34802: [C++][Parquet] Decoder: Allow passing `MemoryPool` to Decoder

Currently, only `DictDecoder` will have memory-pool. And some decoders doesn't have MemoryPool, some decoders like `DeltaBitPacked` have memory pool, but always use default pool.

### Component(s)

C++, Parquet

mapleFU:

Add `::arrow::MemoryPool` to Decoder builder.

### Are these changes tested?

no

Parquet user can pass memory pool to decoder now

* Closes: #34802

### DIFF
--- a/cpp/src/parquet/encoding.cc
+++ b/cpp/src/parquet/encoding.cc
@@ -3417,7 +3417,8 @@ std::unique_ptr<Encoder> MakeEncoder(Type::type type_num, Encoding::type encodin
 }
 
 std::unique_ptr<Decoder> MakeDecoder(Type::type type_num, Encoding::type encoding,
-                                     const ColumnDescriptor* descr) {
+                                     const ColumnDescriptor* descr,
+                                     ::arrow::MemoryPool* pool) {
   if (encoding == Encoding::PLAIN) {
     switch (type_num) {
       case Type::BOOLEAN:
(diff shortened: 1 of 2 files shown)


cpp/src/parquet/encoding.cc
```diff
@@ -3417,7 +3417,8 @@ std::unique_ptr<Encoder> MakeEncoder(Type::type type_num, Encoding::type encodin
 }
 
 std::unique_ptr<Decoder> MakeDecoder(Type::type type_num, Encoding::type encoding,
-                                     const ColumnDescriptor* descr) {
+                                     const ColumnDescriptor* descr,
+                                     ::arrow::MemoryPool* pool) {
```
wgtmac, line 3421: Plain decoder does not need a pool?

mapleFU: Some decoders, like `DeltaBitPacked` may requires a pool, so I added. Currently, Plain doesn't need a Pool

cpp/src/parquet/encoding.h
```diff
@@ -458,7 +459,8 @@ std::unique_ptr<DictDecoder<DType>> MakeDictDecoder(
 
 template <typename DType>
 std::unique_ptr<typename EncodingTraits<DType>::Decoder> MakeTypedDecoder(
-    Encoding::type encoding, const ColumnDescriptor* descr = NULLPTR) {
+    Encoding::type encoding, const ColumnDescriptor* descr = NULLPTR,
+    ::arrow::MemoryPool* pool = ::arrow::default_memory_pool()) {
   using OutType = typename EncodingTraits<DType>::Decoder;
   std::unique_ptr<Decoder> base = MakeDecoder(DType::type_num, encoding, descr);
```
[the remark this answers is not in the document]
mapleFU: Fixed

wgtmac (approved):
LGTM

wjones127 (approved):
Thanks for doing this. FYI we'd eventually like to make sure we are always passing down the right memory pool and guarantee it with unit tests https://github.com/apache/arrow/issues/34058

wgtmac:
> Thanks for doing this. FYI we'd eventually like to make sure we are always passing down the right memory pool and guarantee it with unit tests #34058

That's a good improvement!

wjones127:
Test failure is unrelated.

PR #34806, opened 2023-03-30
GH-34790: [Go]: Add array.Diff()

### LINKED ISSUE #34790: [Go] Support Array Diff

Similar to the CPP implementation of Array Diff (Python docs: https://arrow.apache.org/docs/python/generated/pyarrow.Array.html#pyarrow.Array.diff) it would be good to have the same in Go. As far as I can tell, there is an `array.Equal` in Go right now, but not `array.Diff`.

Like its CPP counterpart, given two string arrays `["one", "two", "three"]` and `["two", None, "two-and-a-half", "three"]` it would output a string like this:

```
@@ -0, +0 @@
-"one"
@@ -2, +1 @@
+null
+"two-and-a-half"
```

### Component(s)

Go

hermanschaaf:
This adds an `array.Diff` function that returns an edit script that, when applied to `base`, produces `target`. This mirrors the C++ implementation. 

It does not yet include a string diff formatter.  This can be done in a follow-up.

* Closes: #34790

### DIFF
--- a/go/arrow/array/diff.go
+++ b/go/arrow/array/diff.go
@@ -1,3 +1,19 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package array
 
 import (
--- a/go/arrow/array/diff.go
+++ b/go/arrow/array/diff.go
@@ -47,7 +47,7 @@ import (
 // target: an array of identical type to base whose elements differ from base's
 // mem: memory to store the result will be allocated from this memory pool
 func Diff(mem memory.Allocator, base, target arrow.Array) (*Struct, error) {
-	if arrow.TypeEqual(base.DataType(), target.DataType()) {
+	if !arrow.TypeEqual(base.DataType(), target.DataType()) {
 		return nil, fmt.Errorf("%w: only taking the diff of like-typed arrays is supported", arrow.ErrNotImplemented)
 	}
 	switch base.DataType().ID() {
--- a/go/arrow/array/diff.go
+++ b/go/arrow/array/diff.go
@@ -51,9 +51,9 @@ func Diff(base, target arrow.Array) (inserts []bool, runLengths []int64, err err
 	case arrow.EXTENSION:
 		return Diff(base.(ExtensionArray).Storage(), target.(ExtensionArray).Storage())
 	case arrow.DICTIONARY:
-		return nil, nil, fmt.Errorf("%w: diffing arrays of type %s is not implemented", arrow.ErrNotImplemented, base.DataType().String())
+		return nil, nil, fmt.Errorf("%w: diffing arrays of type %s is not implemented", arrow.ErrNotImplemented, base.DataType())
 	case arrow.RUN_END_ENCODED:
-		return nil, nil, fmt.Errorf("%w: diffing arrays of type %s is not implemented", arrow.ErrNotImplemented, base.DataType().String())
+		return nil, nil, fmt.Errorf("%w: diffing arrays of type %s is not implemented", arrow.ErrNotImplemented, base.DataType())
 	}
 	d := newQuadraticSpaceMyersDiff(base, target)
 	return d.Diff()
@@ -63,10 +63,6 @@ func Diff(base, target arrow.Array) (inserts []bool, runLengths []int64, err err
 type editPoint struct {
 	base   int
 	target int
-}
-
-func (e editPoint) Equals(other editPoint) bool {
-	return e.base == other.base && e.target == other.target
 }
 
 type quadraticSpaceMyersDiff struct {
--- a/go/arrow/array/diff.go
+++ b/go/arrow/array/diff.go
@@ -20,17 +20,16 @@ import (
 	"fmt"
 
 	"github.com/apache/arrow/go/v12/arrow"
-	"github.com/apache/arrow/go/v12/arrow/memory"
 )
 
 // Diff compares two arrays, returning an edit script which expresses the difference
 // between them. The edit script can be applied to the base array to produce the target.
 //
-// An edit script is an array of struct(insert bool, run_length int64).
+// An edit script represents the difference between two arrays.
 // Each element of "insert" determines whether an element was inserted into (true)
 // or deleted from (false) base. Each insertion or deletion is followed by a run of
 // elements which are unchanged from base to target; the length of this run is stored
-// in "run_length". (Note that the edit script begins and ends with a run of shared
+// in RunLength. (Note that the edit script begins and ends with a run of shared
 // elements but both fields of the struct must have the same length. To accommodate this
 // the first element of "insert" should be ignored.)
 //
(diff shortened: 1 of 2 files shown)


go/arrow/array/diff.go
```diff
@@ -0,0 +1,276 @@
...
+//	{"insert": false, "run_length": 1}, // leading run of length 1 ("h")
+//	{"insert": true, "run_length": 3}, // insert("e") then a run of length 3 ("llo")
+//	{"insert": false, "run_length": 0} // delete("o") then an empty run
+//
+// ]
+//
+// base: baseline for comparison
+// target: an array of identical type to base whose elements differ from base's
+// mem: memory to store the result will be allocated from this memory pool
+func Diff(base, target arrow.Array, mem memory.Allocator) (*Struct, error) {
```
zeroshade, line 50: I think we've been trying to put the allocator as the first argument in methods (similar to the convention of `context.Context` being the first parameter), so can we move the allocator to the first argument?

go/arrow/array/diff.go
```diff
@@ -0,0 +1,276 @@
...
+	if base.DataType().Fingerprint() != target.DataType().Fingerprint() {
+		return nil, errors.New("only taking the diff of like-typed arrays is supported")
+	}
```
zeroshade, line 53: Probably better to use `arrow.TypeEqual(base.DataType(), target.DataType())`

go/arrow/array/diff.go
```diff
@@ -0,0 +1,276 @@
...
+	switch base.DataType().ID() {
+	case arrow.EXTENSION:
+		return Diff(base.(ExtensionArray).Storage(), target.(ExtensionArray).Storage(), mem)
+	case arrow.DICTIONARY:
+		return nil, fmt.Errorf("diffing arrays of type %s is not implemented", base.DataType().String())
+	case arrow.RUN_END_ENCODED:
+		return nil, fmt.Errorf("diffing arrays of type %s is not implemented", base.DataType().String())
```
zeroshade, line 60: can you use `%w` and `arrow.ErrNotImplemented` when constructing this error?

go/arrow/array/diff.go
```diff
@@ -0,0 +1,276 @@
...
+	return d
+}
+
+func (d *quadraticSpaceMyersDiff) valuesEqual(baseIndex, targetIndex int) bool {
+	baseNull := d.base.IsNull(baseIndex)
+	targetNull := d.target.IsNull(targetIndex)
+	if baseNull || targetNull {
+		return baseNull && targetNull
+	}
+	return SliceEqual(d.base, int64(baseIndex), int64(baseIndex+1), d.target, int64(targetIndex), int64(targetIndex+1))
```
zeroshade, line 123: yea, i can see where an `UntypedValue` method would potentially be a more efficient alternative here, but given the current things available this works.

go/arrow/array/diff.go
zeroshade, line 50: Is there really a particular benefit to returning the edits as an actual arrow Struct column vs just returning a native Go object?

hermanschaaf: We probably also want `mem` to be the first argument here?

hermanschaaf:
@zeroshade Thank you for your review earlier, this is now ready for another look. I addressed most of your comments in https://github.com/apache/arrow/pull/34806/commits/3f45bac67f9b3e10374ba048aa2d3a1cfd6c08c6 (with a further fix in https://github.com/apache/arrow/pull/34806/commits/ebf70fd8844dc4e0d0c3be6ad3eceaa605833f1c), and switched to native Go return types in https://github.com/apache/arrow/pull/34806/commits/1f15a10f97ac24c07859a603d4dc7f4e19305e07. This way we can more easily compare the `*Struct` / Native Go implementations.

Personally I prefer the native Go return types. It means we don't need to pass a memory allocator to the `Diff` function, and it simplifies the code here a bit. I toyed with introducing an `EditScript` struct and returning that, but that would have the downside of introducing a new public `array.EditScript` type, which I feel is unnecessary in this instance. It seems better to me to return `inserts []bool, runLengths []int64` and document what they contain. 

Let me know your thoughts!

go/arrow/array/diff.go
```diff
@@ -0,0 +1,248 @@
...
+// [
+//
+//	{"insert": false, "run_length": 1}, // leading run of length 1 ("h")
+//	{"insert": true, "run_length": 3}, // insert("e") then a run of length 3 ("llo")
+//	{"insert": false, "run_length": 0} // delete("o") then an empty run
+//
+// ]
+// base: baseline for comparison
+// target: an array of identical type to base whose elements differ from base's
+func Diff(base, target arrow.Array) (inserts []bool, runLengths []int64, err error) {
```
zeroshade, line 46: Maybe have a struct:

```go
type Edit struct {
    Insert bool
    RunLength int64
}
```

and return `(edits []Edit, err error)`? Thoughts?

go/arrow/array/diff.go
```diff
@@ -0,0 +1,248 @@
...
+		return nil, nil, fmt.Errorf("%w: only taking the diff of like-typed arrays is supported", arrow.ErrNotImplemented)
+	}
+	switch base.DataType().ID() {
+	case arrow.EXTENSION:
+		return Diff(base.(ExtensionArray).Storage(), target.(ExtensionArray).Storage())
+	case arrow.DICTIONARY:
+		return nil, nil, fmt.Errorf("%w: diffing arrays of type %s is not implemented", arrow.ErrNotImplemented, base.DataType().String())
+	case arrow.RUN_END_ENCODED:
+		return nil, nil, fmt.Errorf("%w: diffing arrays of type %s is not implemented", arrow.ErrNotImplemented, base.DataType().String())
+	}
```
zeroshade, line 57: You don't need the `.String()` in there, because you're using `%s` in the format string, it'll automatically know to call the `String()` method.

go/arrow/array/diff.go
```diff
@@ -0,0 +1,248 @@
...
+
+// editPoint represents an intermediate state in the comparison of two arrays
+type editPoint struct {
+	base   int
+	target int
+}
+
+func (e editPoint) Equals(other editPoint) bool {
+	return e.base == other.base && e.target == other.target
+}
```
zeroshade, line 70: funny enough, you don't actually need to create this for the trivial case (which this is) the default thing that will happen when performing `==` on a struct is that it just checks `==` for each member. 

You don't have to get rid of this method, just saying that it's **technically** not necessary :smile:

zeroshade (commented):
Mostly looks good to me, just a few nitpicks

go/arrow/array/diff.go
hermanschaaf: I'm fine either way. I think the main downside is that a single `Edit` is somewhat meaningless in isolation - it needs to be part of a full edit script to be useful, as only the edits leading up to this one can give you the position in the array it is referring to.

I also considered something like:

```
type EditScript struct {
     Inserts []bool
     RunLengths []int64
}
```

But ended up going with returning `inserts []bool, runLengths []int64` separately so we don't need to introduce a new exported type to the `array` package. 

We can also do 

```
type Edit struct {
    Insert bool
    RunLength int64
}
```

like you suggested, paired with

```
type Edits []Edit
```

which could then be returned by `Diff`, and perhaps we could then add a `String()` method on the `Edits` type that returns the diff in unified format?

So usage would be like:

```
edits, err := array.Diff(left, right)
if err != nil { ... }
diffStr := edits.String()
```

I think I like this the most :) Lots of options here, and it also depends on how much leeway we have to go our own way in the Go implementation. Let me know your preferred solution and I'll update it.

go/arrow/array/diff.go
hermanschaaf: Good point, fixed!

go/arrow/array/diff.go
hermanschaaf: @zeroshade I've now introduced

```
type Edit struct {
    Insert bool
    RunLength int64
}

type Edits []Edit
```

and changed the signature to return `(edits Edits, err error)`. Let me know what you think!

zeroshade: I agree, I like the idea of the `type Edit struct` pair with `type Edits []Edit` and giving it a `String` method. So your idea works great!

We have as much leeway as we want to go our own way, if something would be "better" or more idiomatic in go if we do it differently than one of the other impls then it's better to be more idiomatic in my opinion.

zeroshade (approved):
Just two questions, one of which is below.

The other question is whether you think it would be useful/interesting to add a benchmark for computing the diffs? Go benchmarks do get tracked on https://conbench.ursa.dev/ so we can compare performance of Go vs other implementations and across changes, so i'm curious whether you think adding a benchmark for the diffs is something that would be useful or not.

Not required at all, just looking for feedback.

go/arrow/array/diff.go
```diff
@@ -0,0 +1,251 @@
...
+//
+// For example for base "hlloo" and target "hello", the edit script would be
+// [
+//
+//	{"insert": false, "run_length": 1}, // leading run of length 1 ("h")
+//	{"insert": true, "run_length": 3}, // insert("e") then a run of length 3 ("llo")
+//	{"insert": false, "run_length": 0} // delete("o") then an empty run
+//
+// ]
+type Edits []Edit
```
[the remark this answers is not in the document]
hermanschaaf: Ah, no I will add the `String` method in a follow-up if that's okay. I've written the code for it in another draft PR but still need to refactor to use the new types and add some more test cases.

hermanschaaf:
@zeroshade re: Benchmarks: For our purposes this would not be particularly useful right now; we're mostly going to be using `Diff` in tests to assert that arrays match our expectations (and print out the differences if they don't), at least for the time being. Performance is not a particular concern in this case, and the algorithm is the same as in the C++ implementation, but we could always add some benchmarks if performance turns out to be an issue later?

PR #34808, opened 2023-03-30
GH-34807: [Go] Handle `io.EOF` when reading parquet footer size and magic bytes

### LINKED ISSUE #34807: [Go] file.NewParquetReader doesn't properly handle all `io.ReaderAt` implementations

Specifically, when supplied an implementation of `parquet.ReaderSeeker` that returns `io.EOF` as an error when `io.ReadAt` perfect reads to the end of stream, it interprets it as a proper error:

https://github.com/apache/arrow/blob/go/v11.0.0/go/parquet/file/file_reader.go#L166-L168

The interface of `io.ReaderAt` states (https://pkg.go.dev/io#ReaderAt):
```
If the n = len(p) bytes returned by ReadAt are at the end of the input source, ReadAt may return either err == EOF or err == nil.
```

### Component(s)

Go

agchang:
* Closes: #34807

### DIFF
--- a/go/parquet/file/file_reader_test.go
+++ b/go/parquet/file/file_reader_test.go
@@ -19,6 +19,8 @@ package file_test
 import (
 	"bytes"
 	"encoding/binary"
+	"github.com/apache/arrow/go/v12/parquet"
+	"github.com/apache/arrow/go/v12/parquet/schema"
 	"io"
 	"math/rand"
 	"testing"
@@ -63,6 +65,19 @@ func checkStatistics(t *testing.T, stats format.Statistics, actual metadata.Enco
 	if stats.IsSetDistinctCount() {
 		assert.Equal(t, stats.GetDistinctCount(), actual.DistinctCount)
 	}
+}
+
+type testReader struct {
+	*bytes.Reader
+}
+
+// ReadAt for testReader returns io.EOF when off + len(b) is exactly the length of the underlying input source.
+func (tr testReader) ReadAt(b []byte, off int64) (int, error) {
+	n, err := tr.Reader.ReadAt(b, off)
+	if err == nil && (int64(n)+off == tr.Size()) {
+		return n, io.EOF
+	}
+	return n, err
 }
 
 type PageSerdeSuite struct {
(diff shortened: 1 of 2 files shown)


zeroshade:
Is there more you intend to do here? I'd say that a unit test which uses an implementation that returns the io.EOF and confirms this fixes it would be good. Otherwise this seems fine to me if you think it's ready for review.

agchang:
> Is there more you intend to do here? I'd say that a unit test which uses an implementation that returns the io.EOF and confirms this fixes it would be good. Otherwise this seems fine to me if you think it's ready for review.

I added a test. I wasn't sure of the best way to get a buffer to an underlying valid parquet file for testing purposes, so I pieced together things from other tests -- let me know if there's a better way.

go/parquet/file/file_reader_test.go
```diff
@@ -271,6 +286,31 @@ func (p *PageSerdeSuite) TestCompression() {
...
+	builder := metadata.NewFileMetadataBuilder(schema, props, nil)
+	fileMetaData, _ := builder.Finish()
+
+	size, _ := serializer.Serialize(fileMetaData, sink, nil)
+
+	binary.Write(sink, binary.LittleEndian, uint32(size))
+	magic := []byte("PAR1")
+	sink.Write(magic)
+	buf := sink.Finish()
+	defer buf.Release()
```
zeroshade, line 307: You should be able to easily do something like:

```go
root, _ := schema.NewGroupNode("schema", parquet.Repetitions.Repeated, schema.FieldList{
        schema.NewInt32Node("int_col", parquet.Repetitions.Required, -1)}, -1)
props := parquet.NewWriterProperties(parquet.WithVersion(parquet.V2_LATEST)

var sink bytes.Buffer
wr := file.NewParquetWriter(&sink, root, file.WithWriterProps(props))
require.NoError(t, wr.Close())

r := bytes.NewReader(buf.Bytes())
_, err := file.NewParquetReader(testReader{Reader: r})
assert.NoError(t, err)
```
And just make sure that without your change the test fails.

agchang: Appreciate the help! Much cleaner. Done.

go/parquet/file/file_reader_test.go
```diff
@@ -19,6 +19,9 @@ package file_test
 import (
 	"bytes"
 	"encoding/binary"
+	"github.com/apache/arrow/go/v12/parquet"
+	"github.com/apache/arrow/go/v12/parquet/schema"
+	"github.com/stretchr/testify/require"
```
zeroshade, line 24: can you run `go fmt` on this / `goimports` so it reorders the imports sorting them appropriately? that's likely what's causing the failing dev script CI tests, these are getting ordered differently resulting in the diffs not being what is expected.

zeroshade: I should really add a workflow for the bot to auto-run the formatting like we have for C++ and clang-format...

zeroshade:
@agchang the import order / `go fmt` issue is the only thing left on here, I'm happy to merge once that's addressed.

Thanks again for this!

go/parquet/file/file_reader_test.go
agchang: Done! Thanks for the pointer.

> I should really add a workflow for the bot to auto-run the formatting like we have for C++ and clang-format...

I can take a look if you guys need help on the go front :-)

agchang: Hmm, still have failing checks, I ran:
`go fmt -d file_reader.go file_reader_test.go` and see no diffs 🤔

zeroshade: you needed to remove the empty line and then use goimports to sort the imports. I've done so and pushed the fix.

zeroshade (approved):
LGTM

PR #34815, opened 2023-03-31
GH-34778: [Java] Only apply ServerInterceptorAdapter logic to Flight service requests

### LINKED ISSUE #34778: [Java] Ability to register additional grpc services with FlightServer

We want to add the [gRPC health endpoint](https://github.com/grpc/grpc-java/blob/master/services/src/generated/main/grpc/io/grpc/health/v1/HealthGrpc.java) to our server. It looks like the ability to do this was added to the [go server with this PR.](https://github.com/apache/arrow/pull/13995/files). It would be great to have something similar for the Java server.

### Component(s)

Java

aiguofer:
### Rationale for this change

The interceptor applies to all methods. When receiving a method for a different service, we don't want to throw an error because the method doesn't exist.

- Ensure added services work correctly
- Add test to make sure registered services work correctly
- Add documentation explaining how to add services

### Are these changes tested?

Yes

Yes, documentation now explains how to add services

* Closes: #34778

### DIFF
--- a/java/flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/ServerInterceptorAdapter.java
+++ b/java/flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/ServerInterceptorAdapter.java
@@ -82,7 +82,14 @@ public class ServerInterceptorAdapter implements ServerInterceptor {
   @Override
   public <ReqT, RespT> Listener<ReqT> interceptCall(ServerCall<ReqT, RespT> call, Metadata headers,
       ServerCallHandler<ReqT, RespT> next) {
-    final CallInfo info = new CallInfo(FlightMethod.fromProtocol(call.getMethodDescriptor().getFullMethodName()));
+    FlightMethod method;
+    try {
+      method = FlightMethod.fromProtocol(call.getMethodDescriptor().getFullMethodName());
+    } catch (IllegalArgumentException ignored) {
+      // Don't add middleware to non Flight methods
+      return Contexts.interceptCall(Context.current(), call, headers, next);
+    }
+    final CallInfo info = new CallInfo(method);
     final List<FlightServerMiddleware> middleware = new ArrayList<>();
     // Use LinkedHashMap to preserve insertion order
     final Map<FlightServerMiddleware.Key<?>, FlightServerMiddleware> middlewareMap = new LinkedHashMap<>();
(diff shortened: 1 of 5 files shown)


aiguofer:
Upon further looking at the code, it seems that this already exists:

```
      // Allow hooking into the gRPC builder. This is not guaranteed to be available on all Arrow versions or
      // Flight implementations.
      builderOptions.computeIfPresent("grpc.builderConsumer", (key, builderConsumer) -> {
        final Consumer<NettyServerBuilder> consumer = (Consumer<NettyServerBuilder>) builderConsumer;
        consumer.accept(builder);
        return null;
      });
```

I can update the PR to only make the fixes to `ServerInterceptorAdapter.java` so it won't throw an exception on non-flight requests and add some docs on how to add extra services.

lidavidm:
Wow, I forgot about that. Yes, docs + any fixes necessary to make it work better would be much welcome.

java/flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/ServerInterceptorAdapter.java
```diff
@@ -82,7 +82,14 @@ public ServerInterceptorAdapter(List<KeyFactory<?>> factories) {
   @Override
   public <ReqT, RespT> Listener<ReqT> interceptCall(ServerCall<ReqT, RespT> call, Metadata headers,
       ServerCallHandler<ReqT, RespT> next) {
-    final CallInfo info = new CallInfo(FlightMethod.fromProtocol(call.getMethodDescriptor().getFullMethodName()));
+    FlightMethod method;
+    try {
+      method = FlightMethod.fromProtocol(call.getMethodDescriptor().getFullMethodName());
```
lidavidm, line 87: I wonder if we could do something like check the service name instead, to avoid catching an exception + to avoid silently passing things through if we add new Flight methods in the future?

lidavidm (commented):
Thank you! LGTM, but I have one question

java/flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/ServerInterceptorAdapter.java
aiguofer: Hmmm makes sense! I'll see what I can do

lidavidm (approved):
Thanks!

Issue #34775, opened 2023-03-29
[R] arrow_table: as.data.frame() sometimes returns a tbl and sometimes a data.frame


daattali:
### Describe the bug, including details regarding any error messages, version, and platform.

The documentation for `arrow_table()` says that you can call `as.data.frame()` on an arrow table object to convert it to a dataframe. It seems like the behaviour is inconsistent. Example:

```r
> class(as.data.frame(arrow::arrow_table(name = "1", mtcars)))
[1] "tbl_df"     "tbl"        "data.frame"
> class(as.data.frame(arrow::arrow_table(mtcars, name = "1")))
[1] "data.frame"
```

I created the same arrow_table in both cases but only changed the order of the arguments, and it returned a different object type.

I would expect it to be base R dataframe since `as.data.frame()` is a generic from base R that is generally assumed to convert to actual `data.frame` objects, not to tibbles. Right now if I want to guarantee a base R data.frame, I need to do `as.data.frame(as.data.frame(table))`. I need to ocnvert it using the same function twice, which is extremely awkward and unintuitive. I strongly believe arrow should not conflate `data.frame()` with `tbl_df`, it can cause errors for users.

### Component(s)

R

thisisnic:
Thanks for reporting this, I agree it's not intended behaviour.  Looks like it's something to do with how metadata is handled, will have a closer look at this.

PR #5399, opened 2019-09-17
ARROW-6337: [R] Changed as_tible to as_dataframe in the R package

jameslamb:
See [ARROW-6337](https://issues.apache.org/jira/browse/ARROW-6337) for background on this PR.

PR #34825, opened 2023-03-31
GH-34775: [R] arrow_table: as.data.frame() sometimes returns a tbl and sometimes a data.frame

thisisnic:
This PR currently updates the table creator to only extract metadata if a single data.frame has been passed in in the dots and nothing else - previously the metadata was extracted from the first item passed in if it is a `data.frame`, resulting in inconsistent behaviour depending on argument order.

It also ensures that the object returned from `as.data.frame()` is always a vanilla data.frame (previously it sometimes was a tibble and sometimes was a data.frame).

``` r
library(arrow)

tbl <- data.frame(x = 1)
x <- arrow_table(tbl, name = "1") 
y <- arrow_table(name = "1", tbl) 
z <- arrow_table(tbl)
x$r_metadata
#> $attributes
#> $attributes$class
#> [1] "data.frame"
#> 
#> 
#> $columns
#> $columns$x
#> NULL
#> 
#> $columns$name
#> NULL
y$r_metadata
#> NULL
z$r_metadata
#> $attributes
#> $attributes$class
#> [1] "data.frame"
#> 
#> 
#> $columns
#> $columns$x
#> NULL

class(as.data.frame(x))
#> [1] "data.frame"
class(as.data.frame(y))
#> [1] "tbl_df"     "tbl"        "data.frame"
class(as.data.frame(z))
#> [1] "data.frame"
```

After:

``` r
library(arrow)

tbl <- data.frame(x = 1)
x <- arrow_table(tbl, name = "1") 
y <- arrow_table(name = "1", tbl) 
z <- arrow_table(tbl)
x$r_metadata
#> NULL
y$r_metadata
#> NULL
z$r_metadata
#> $attributes
#> $attributes$class
#> [1] "data.frame"
#> 
#> 
#> $columns
#> $columns$x
#> NULL

class(as.data.frame(x))
#> [1] "data.frame"
class(as.data.frame(y))
#> [1] "data.frame"
class(as.data.frame(z))
#> [1] "data.frame"
```

I did look at implementing `as_tibble.ArrowTabular()` but it's unnecessary as `as_tibble()` by default will call `as.data.frame()` on any object which doesn't have an S3 method implemented for anyway.


thisisnic:
> > We can't just swap it for as_tibble() inside these functions as we don't have the tibble package as a dependency.
> 
> I think we can safely add tibble to Suggests and use `as_tibble()` in tests that already expect a tibble? Unless I'm misunderstanding the failures?
> 

To clarify, I'm talking about the functions themselves, not the tests. We call `as.data.frame()` in the body of `read_feather()`.

thisisnic:
OK, I'm in favour of taking it on as a new dependency, though it's a big enough change that I'm going to add it as a discussion topic in the dev meeting on Thursday before pushing it ahead.

thisisnic (commented):
I'll wait til we're happy with the overall PR before doing this, but we should absolutely update the docs before merging this.

thisisnic:
Thanks folks for the discussion here, it's good to have more eyes on this.

> If I remember correctly, `read_feather` and `read_parquet` restore the attributes of R as they were before they were written, so if we didn't write tibble to arrow file or parquet, it didn't become tibble, I think.

That's right, they restore properties, but there was a little more going on here; sometimes we return a tibble and other times we return a `data.frame`, which isn't great.  That was the original purpose of this PR - to fix that - but in fixing it, it became more complex.

assignUser:
I like the consistency of always returning a tibble and imo that would make sense seeing the overall alignment of the arrow package with dplyr.

ianmcook:
I believe we should avoid adding a hard dependency on tibble if we can avoid it, and provide a way to return a plain old base R data.frame for users who prefer that.

assignUser:
Ah I thought we had it as an indirect dependency already...

thisisnic:
@ianmcook Would you mind expanding more on why not? I don't feel like you'll be the only person with this opinion, and regardless of what we end up doing, I want to better understand the objections. 

There's also the possible workaround you mentioned elsewhere involving conditionally exporting the S3 generic that we should explore more as if it works it allows us to return tibbles but without the added dependency.

ianmcook:
It's true that tibble would _not_ be our first hard dependency on a tidyverse package, but if I'm not mistaken, it _would_ be our first hard tidyverse package dependency **that is exposed to end users of the arrow package**. I believe all of our other hard tidyverse package dependencies are just under-the-hood stuff. I'm a big tidyverse fan myself, but as a general rule I think we should aim for the arrow package to expose no tidyverse stuff to end users unless they explicitly ask for it.

ianmcook:
> ...or if returning a tibble really is a problem, we should commit to returning only data.frames, because returning a tibble only sometimes is the worst of all worlds.

IMO it would be a reasonable behavior to return a tibble unless (a) the tibble package is not installed, or (b) the user sets an option to specify that they want a plain old data.frame

eitsupi:
I am wondering if such a case would eliminate the need to write (some) R attributes to the files?
Is there a benefit to continuing to write these to files?

thisisnic:
Given that nobody asked us to implement `as_tibble()` and this is contentious, I'm tempted to reduce the scope of this PR to guaranteeing that `as.data.frame()` returns a base R data.frame, picking a new (internal) function to use in the body of the `read_*()` functions instead of `as.data.frame()`, and not implementing `as_tibble()` in this PR.  We'd be left with the problem of still variably returning data.frames vs. tibbles from the `read_*` functions, but can defer that to another discussion/ticket.

thisisnic:
(FWIW I tried the conditional export approach I mentioned above but couldn't get it working).

nealrichardson:
Apologies for being late to the discussion here, and forgive me if I'm missing something, but it seems like the odd behavior @thisisnic reported is a more narrow problem of how the auto-splicing works in `arrow_table() / Table$create()`. That was added in #4704, and there was a TODO issue made about improving the C++ code around it (#22186), though I don't think this particular issue was what they had in mind. This seems worth fixing.

But I'm not sure there's a more general problem to solve; or rather, I'm not sure there's a better solution than the one we have that meets our requirements. It seems like the constraints we're solving for, ranked by priority, are:

1. `as.data.frame(ArrowObject)` should return an object where `is.data.frame()` is `TRUE` and can be used by any other R functions. No Arrow methods required to work with it.
2. Fidelity: when a user saves an R data object to Arrow or Parquet, they should get the same R object back when they load it, including metadata.
3. Internal coherence: the same kind of operation in various places should yield the same kind of result.
4. `tbl`s are useful because they print nicer than vanilla `data.frame`s.
5. Because we want Arrow to have broad adoption and appeal, we want to avoid adding hard dependencies.

Our current approach is: when converting a Table to a data.frame, we [set the class](https://github.com/apache/arrow/blob/main/r/src/array_to_vector.cpp#L1386) as [`c("tbl_df", "tbl", "data.frame")`](https://github.com/apache/arrow/blob/main/r/src/symbols.cpp#L60). instead of just "data.frame" But [if there is R metadata](https://github.com/apache/arrow/blob/main/r/R/metadata.R#L99) saved that overwrites the class, we use that so that we can faithfully round-trip R objects. Leaving aside the autosplice issue, I think that's all that's happening. 

We're relying on how S3 dispatch works, such that if you don't have `tibble` loaded and print a `c("tbl_df", "tbl", "data.frame")`, it just prints like a data.frame, so we can both (a) return tbls and (b) not depend on `tibble` just fine. So we can get `#4` and `#5`, but because of `#2`, we sacrifice a small amount of `#3`: we get the weaker `inherits(x, "data.frame")` requirement and not `identical(class(x), "data.frame")`. I'm of the opinion that this is a reasonable tradeoff since tibbles are just subclasses of data.frame--you always get back a data.frame. We could push back the other way, but I think we have to give up some `#4` as a result. 

To be clear, I think you/we could justify any number of tradeoffs along these dimensions, I just wanted to sketch out what I think they are.

A few other specific thoughts:

* I'm not sure you can make `as.data.frame()` always return `class == "data.frame"` without either sacrificing roundtrip fidelity or inventing a new function that means "take this Arrow Table with metadata and make an R object". The latter doesn't sound like something we want to do.
* The `read_xxx()` functions can't really be strict about `class` without losing roundtrip fidelity. For me, that's enough to say `inherits(x, "data.frame")` is sufficient and we don't have to always return identical class. 
* Seems like `dplyr::collect()` should always return tibble, regardless of other changes you make (I believe it already does, given how the ExecPlan evaluation works).

Additional historical context, in case it's useful in understanding how we got here:

* https://github.com/apache/arrow/pull/4454
* https://github.com/apache/arrow/pull/5399
* https://github.com/apache/arrow/issues/22715

daattali:
As the user who reported the bug this PR is solving, I'd like to very succintly add my opinion.

I personally **strongly** believe that `as.data.frame()` should return a `data.frame`. Plain and simple. That's just what would be intuitively expected. If I cast any object with `as.X()` then I expect the result to be exactly `X`, not a clever derivation/subclass of `X`. If I want a tibble, then I'd use for `as.tibble()`.

Having `as.data.frame(x)` and `as.data.frame(as.data.frame(x))` return different types feels very wrong.

thisisnic:
Thanks for summarising the extra context there that we didn't have before, @nealrichardson, that's super helpful; now I can see why things are the way they currently are.

FWIW, I agree with your points Dean, but I don't see a reasonable solution to that problem which doesn't cause other issues given the different priorities we're balancing.  

I'm still on the fence regarding what is the best solution here, but given that the issue regarding the `as.data.frame` return type is 1) fairly complex, 2) contentious, 3) not the entire issue here, and 4) not something that has been reported by anyone else (yet), I'm going to close this PR and submit a new one for the narrower problem of variable results based on argument ordering.

Open to more discussion here regarding what our priorities *should* be, if not the current ones in the order they are above.

For example, I don't fully see the importance of roundtrip fidelity; if the input and output have the same (non-default) metadata and contents, then is there any harm caused by returning a tibble instead of a data.frame, if a user then just has the additional step of calling `as.data.frame()` to get a vanilla `data.frame`? Haven't sketched that out in practice though.

thisisnic:
> having the end result of this PR be "it was too hard so we didn't fix it" is a sustainable path.

That's not what's being said or done here. I'm separating out the different issues into smaller components.  I'm finishing the small and easily solvable one, and pausing working on this one to wait for more discussion, as, as you said, we don't have many contributors, and I feel that we're starting to sacrifice momentum and speed of resolution for what could be modular issues in favour of heading towards yak shaving at the moment.

PR #34827, opened 2023-03-31
GH-34790: [Go]: Add array.Edits.UnifiedDiff

hermanschaaf:
This adds a `UnifiedDiff(base, target arrow.Array)` method to the `array.Edits` type. It returns a string diff in Unified Diff format. This makes use of the `array.Edits` type returned by the `arrays.Diff()` function added in https://github.com/apache/arrow/pull/34806

- Part of https://github.com/apache/arrow/issues/34790


go/arrow/array/diff.go
```diff
@@ -0,0 +1,326 @@
...
+
+func (d *quadraticSpaceMyersDiff) Diff(mem memory.Allocator) (*Struct, error) {
+	for !d.Done() {
+		d.Next()
+	}
+	return d.GetEdits(mem)
+}
+
+// DiffString returns a representation of the diff in Unified Diff format.
+func DiffString(base, target arrow.Array, mem memory.Allocator) (string, error) {
```
zeroshade, line 280: maybe `UnifiedDiff` instead of `DiffString` ?

zeroshade (approved):

PR #34849, opened 2023-04-02
MINOR: [Go] Add tests for decimal arithmetic operations

izveigor:
### Rationale for this change
There are no tests for arithmetic operations of decimal128 and decimal256 in the project.

Tests cover the following operators:
- Add
- Sub
- Mul
- Div
Add tests for these cases.

### Are these changes tested?
Yes

No


zeroshade (approved):
Thanks for doing this! I had written tests that are in the `compute` package, but it's definitely a good idea to have tests for the Add method right here in this package too.

PR #34854, opened 2023-04-03
GH-34853: [Go] Add TotalRecordSize, TotalArraySize

### LINKED ISSUE #34853: [Go][Util] TotalBufferSize

Similar to [cpp](https://arrow.apache.org/docs/cpp/api/compute.html#_CPPv4NK5arrow5Datum15TotalBufferSizeEv) implementation 

### Component(s)

Go

yevgenypats:
Closes https://github.com/apache/arrow/issues/34853
* Closes: #34853

### DIFF
--- a/go/arrow/array/record.go
+++ b/go/arrow/array/record.go
@@ -77,6 +77,7 @@ func NewRecordReader(schema *arrow.Schema, recs []arrow.Record) (*simpleRecords,
 func (rs *simpleRecords) Retain() {
 	atomic.AddInt64(&rs.refCount, 1)
 }
+
 // Release decreases the reference count by 1.
 // When the reference count goes to zero, the memory is freed.
 // Release may be called simultaneously from multiple goroutines.
--- a/go/arrow/array/record.go
+++ b/go/arrow/array/record.go
@@ -179,14 +179,6 @@ func (rec *simpleRecord) SetColumn(i int, arr arrow.Array) (arrow.Record, error)
 	return NewRecord(rec.schema, arrs, rec.rows), nil
 }
 
-func (rec *simpleRecord) TotalBufferSize() uint64 {
-	var sz uint64
-	for _, arr := range rec.arrs {
-		sz += uint64(arr.Data().Len())
-	}
-	return sz
-}
-
 func (rec *simpleRecord) validate() error {
 	if rec.rows == 0 && len(rec.arrs) == 0 {
 		return nil
--- a/go/arrow/array/record.go
+++ b/go/arrow/array/record.go
@@ -77,7 +77,6 @@ func NewRecordReader(schema *arrow.Schema, recs []arrow.Record) (*simpleRecords,
 func (rs *simpleRecords) Retain() {
 	atomic.AddInt64(&rs.refCount, 1)
 }
-
 // Release decreases the reference count by 1.
 // When the reference count goes to zero, the memory is freed.
 // Release may be called simultaneously from multiple goroutines.
@@ -178,6 +177,14 @@ func (rec *simpleRecord) SetColumn(i int, arr arrow.Array) (arrow.Record, error)
 	arrs[i] = arr
 
 	return NewRecord(rec.schema, arrs, rec.rows), nil
+}
+
+func (rec *simpleRecord) TotalBufferSize() uint64 {
+	var sz uint64
+	for _, arr := range rec.arrs {
+		sz += uint64(arr.Data().Len())
+	}
+	return sz
 }
 
 func (rec *simpleRecord) validate() error {
--- a/go/arrow/util/byte_size.go
+++ b/go/arrow/util/byte_size.go
@@ -27,7 +27,7 @@ func isArrayDataNil(arrayData arrow.ArrayData) bool {
 		return true
 	}
 	if v, ok := arrayData.(*array.Data); ok {
-    return v == nil
+		return v == nil
 	}
 	panic("unknown ArrayData type")
 }
@@ -36,13 +36,13 @@ func totalArrayDataSize(arrayData arrow.ArrayData, seenBuffers map[*memory.Buffe
 	var sum int64
 	var void = struct{}{}
 	for _, buf := range arrayData.Buffers() {
-    if buf == nil {
-        continue
-    }
-    if _, ok := seenBuffers[buf]; !ok {
-        sum += int64(buf.Len())
-        seenBuffers[buf] = void
-    }
+		if buf == nil {
+			continue
+		}
+		if _, ok := seenBuffers[buf]; !ok {
+			sum += int64(buf.Len())
+			seenBuffers[buf] = void
+		}
 	}
 	for _, child := range arrayData.Children() {
 		sum += totalArrayDataSize(child, seenBuffers)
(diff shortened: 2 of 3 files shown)


zeroshade:
@yevgenypats For future reference: the style for naming the PRs is `GH-####: [Go] ...` don't put brackets around the `GH-#####` part and the CI will automatically pick it up and add the "closes" comment for you. I've updated the title for this PR but just mentioning it for your future PRs. Thanks! I'll take a look at this in just a bit. I'm just catching up on notifications from while i was at a conference last week :smile:

go/arrow/util/byte_size.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	switch v := arrayData.(type) {
+	case *array.Data:
+		if v == nil {
+			return true
+		}
+	default:
+		panic("unknown array data type")
+	}
+	return false
```
zeroshade, line 38: probably simpler to do:

```go
if arrayData == nil {
    return true
}

if v, ok := arrayData.(*array.Data); ok {
    return v == nil
}
panic("unknown ArrayData type")
```

go/arrow/util/byte_size.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func totalArrayDataSize(arrayData arrow.ArrayData, seenBuffers map[string]struct{}) int64 {
```
zeroshade, line 41: would be more efficient to just use a `map[*memory.Buffer]struct{}` instead rather than using `fmt.Sprintf("%p", buf)` no need to convert to string for the map key

go/arrow/util/byte_size.go
```diff
@@ -0,0 +1,83 @@
...
+	var sum int64
+	for _, buf := range arrayData.Buffers() {
+		if buf != nil {
+			_, ok := seenBuffers[fmt.Sprintf("%p", buf)]
+			if !ok {
+				sum += int64(buf.Len())
+				seenBuffers[fmt.Sprintf("%p", buf)] = struct{}{}
+			}
```
zeroshade, line 49: can simplify this:

```go
var void = struct{}{}
for _, buf := range arrayData.Buffers() {
    if buf == nil {
        continue
    }
    if _, ok := seenBuffers[buf]; !ok {
        sum += int64(buf.Len())
        seenBuffers[buf] = void
    }
}
```

go/arrow/util/byte_size.go
```diff
@@ -0,0 +1,83 @@
...
+
+func TotalArraySize(arr arrow.Array) int64 {
+	seenBuffer := make(map[string]struct{})
+	return totalArraySize(arr, seenBuffer)
+}
+
+func TotalRecordSize(record arrow.Record) int64 {
+	seenBuffer := make(map[string]struct{})
+	return totalRecordSize(record, seenBuffer)
+}
```
zeroshade, line 83: Can you add docstring comments to these plz? thanks

go/arrow/util/byte_size_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"testing"
+
+	"github.com/apache/arrow/go/v12/arrow"
+	"github.com/apache/arrow/go/v12/arrow/array"
+	"github.com/apache/arrow/go/v12/arrow/memory"
+	"github.com/apache/arrow/go/v12/arrow/util"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTotalArraySizeBasic(t *testing.T) {
```
zeroshade, line 30: Please add tests for the nil cases and for reused buffers and dictionaries please

go/arrow/array/record.go
```diff
@@ -180,6 +179,14 @@ func (rec *simpleRecord) SetColumn(i int, arr arrow.Array) (arrow.Record, error)
 	return NewRecord(rec.schema, arrs, rec.rows), nil
 }
 
+func (rec *simpleRecord) TotalBufferSize() uint64 {
```
zeroshade, line 182: Since you added `TotalRecordSize` in utils, is there actually a need for this separate method? If so, can it just call `TotalRecordSize` and add this to the `arrow.Record` interface?

zeroshade (changes requested):

go/arrow/array/record.go
yevgenypats: I removed this method as we can't add it to the interface given it will cause the cyclical loop

yevgenypats:
Thanks @zeroshade  and sorry for the delay here. Addressed the review.

go/arrow/array/record.go
```diff
@@ -77,7 +77,6 @@ func NewRecordReader(schema *arrow.Schema, recs []arrow.Record) (*simpleRecords,
 func (rs *simpleRecords) Retain() {
 	atomic.AddInt64(&rs.refCount, 1)
 }
-
```
zeroshade, line 80: why remove the empty line?

go/arrow/util/byte_size.go
```diff
@@ -0,0 +1,80 @@
...
+
+func isArrayDataNil(arrayData arrow.ArrayData) bool {
+	if arrayData == nil {
+		return true
+	}
+	if v, ok := arrayData.(*array.Data); ok {
+    return v == nil
+	}
+	panic("unknown ArrayData type")
+}
```
zeroshade, line 33: can you run `go fmt` on the files?

go/arrow/util/byte_size.go
```diff
@@ -0,0 +1,80 @@
...
+
+func totalRecordSize(record arrow.Record, seenBuffers map[*memory.Buffer]struct{}) int64 {
+	var sum int64
+	for _, c := range record.Columns() {
+		sum += totalArraySize(c, seenBuffers)
+	}
+	return sum
+} 
+
+// TotalArraySize The sum of bytes in each buffer referenced by the Array.
```
zeroshade, line 70: just write it like a sentence, ie: `TotalArraySize returns the sum of the number of bytes in each buffer referenced by the Array.`

zeroshade (changes requested):
just some very nitpicky things for formatting, but otherwise LGTM

go/arrow/util/byte_size.go
```diff
@@ -0,0 +1,79 @@
...
+	return sum
+}
+
+// TotalArraySize returns the sum of the number of bytes in each buffer referenced by the Array.
+func TotalArraySize(arr arrow.Array) int64 {
+	seenBuffer := make(map[*memory.Buffer]struct{})
+	return totalArraySize(arr, seenBuffer)
+}
+
+// TotalArraySize The sum of bytes in each buffer referenced by the Record.
```
zeroshade, line 75: same comment as previous, make this a regular sentence (general conventions for godoc comments). `TotalRecordSize returns the......` etc.

yevgenypats: Done

go/arrow/util/byte_size.go
```diff
@@ -0,0 +1,79 @@
...
+	return sum
+}
+
+// TotalArraySize returns the sum of the number of bytes in each buffer referenced by the Array.
+func TotalArraySize(arr arrow.Array) int64 {
+	seenBuffer := make(map[*memory.Buffer]struct{})
+	return totalArraySize(arr, seenBuffer)
+}
+
+// TotalArraySize return the sum of bytes in each buffer referenced by the Record.
```
zeroshade, line 75: fix the name here :) Should be `TotalRecordSize` to be consistent with the function your documenting.

zeroshade (changes requested):

zeroshade (approved):

PR #34864, opened 2023-04-03
GH-34863: [Go] Pow method for Decimal DataTypes

### LINKED ISSUE #34863: [Go] Pow method for Decimal DataTypes

Pow method for Decimal is implemented in many languages (Golang: https://pkg.go.dev/github.com/shopspring/decimal#Decimal.Pow, Rust: https://docs.rs/rust_decimal/latest/rust_decimal/struct.Decimal.html#impl-Pow%3CDecimal%3E-for-Decimal, Python: https://github.com/python/cpython/blob/3.11/Lib/_pydecimal.py#L2298 and others). Because of its popularity I think Golang Arrow should also support this operation.

### Component(s)

Go

izveigor:
### Rationale for this change
Closes: https://github.com/apache/arrow/issues/34863

### Are these changes tested?
Yes

Yes
* Closes: #34863

### DIFF
--- a/go/arrow/decimal128/decimal128.go
+++ b/go/arrow/decimal128/decimal128.go
@@ -134,6 +134,11 @@ func (n Num) Div(rhs Num) (res, rem Num) {
 	return FromBigInt(out), FromBigInt(remainder)
 }
 
+func (n Num) Pow(rhs Num) Num {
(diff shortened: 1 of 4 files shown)


zeroshade (approved):
Thanks much for this! LGTM

PR #34873, opened 2023-04-04
GH-34749 : [Java] Make Zstd compression level configurable

### LINKED ISSUE #34749: [Java] ZstdCompressionCodec uses hard-coding compression level which cannot be configurable

ArrowFileWriter/ArrowStreamWriter add compression support for 12.0.0-dev. But [ZstdCompressionCodec](https://github.com/apache/arrow/blob/196fbe5f5e0669b1f8bf71005c8a4c6a4e0fb2da/java/compression/src/main/java/org/apache/arrow/compression/ZstdCompressionCodec.java#L41) uses hard-coding compression level (level 3) which cannot be configurable. As ZSTD level affects compression speed and ratio, ideally we should make such parameter configurable to developers.

### Component(s)

Java

davisusanibar:
### Rationale for this change

Closes: https://github.com/apache/arrow/issues/34749

Make compression level configurable for Zstd

### Are these changes tested?

Yes

No
* Closes: #34749


java/compression/src/main/java/org/apache/arrow/compression/CommonsCompressionFactory.java
```diff
@@ -27,15 +27,22 @@
...
+
+  public CommonsCompressionFactory(int compressionLevel) {
+    this.compressionLevel = compressionLevel;
+  }
+
+  @Deprecated
+  public static final CommonsCompressionFactory INSTANCE = new CommonsCompressionFactory(3);
 
   @Override
   public CompressionCodec createCodec(CompressionUtil.CodecType codecType) {
```
wgtmac, line 40: Is it better to extend `createCodec` to add an `Optional<int> level`? The downside of the current implementation is that we need to create separate instances for different levels.

davisusanibar: Just updated.

```
ArrowFileWriter writer = new ArrowFileWriter(vectorSchemaRoot, null,
  fileOutputStream.getChannel(), null, IpcOption.DEFAULT,
  CommonsCompressionFactory.INSTANCE, zstdCodec.getCodecType(), /*compressionLevel*/ 7)
```

davisusanibar: I'm going to fill a ticket to implement Builder patter for Reader/Writes class.

java/vector/src/main/java/org/apache/arrow/vector/ipc/ArrowWriter.java
```diff
@@ -119,6 +119,53 @@ protected ArrowWriter(VectorSchemaRoot root, DictionaryProvider provider, Writab
...
+   *
+   * @param root               the vectors to write to the output
+   * @param provider           where to find the dictionaries
+   * @param out                the output where to write
+   * @param option             IPC write options
+   * @param compressionFactory Compression codec factory
+   * @param codecType          Compression codec
+   * @param compressionLevel   Compression level
+   */
+  protected ArrowWriter(VectorSchemaRoot root, DictionaryProvider provider, WritableByteChannel out, IpcOption option,
```
wgtmac, line 133: This duplicates a lot with the previous ctor. Probably we can use one as a delegate or refactor one to accept a created codec?

wgtmac (commented):
Rest LGTM.

davisusanibar:
Hi @lidavidm do you have any comments for this PR?

java/compression/src/test/java/org/apache/arrow/compression/TestArrowReaderWriterWithCompression.java
```diff
@@ -85,4 +85,46 @@ public void testArrowFileZstdRoundTrip() throws Exception {
...
+    try (ArrowFileReader reader =
+             new ArrowFileReader(new ByteArrayReadableSeekableByteChannel(out.toByteArray()),
+                 allocator, CommonsCompressionFactory.INSTANCE)) {
+      Assert.assertEquals(1, reader.getRecordBlocks().size());
+      Assert.assertTrue(reader.loadNextBatch());
+      Assert.assertTrue(root.equals(reader.getVectorSchemaRoot()));
+      Assert.assertFalse(reader.loadNextBatch());
+    }
+
+    // Read the in-memory compressed arrow file without CompressionFactory provided
```
lidavidm, line 119: Isn't this tested elsewhere already?

java/vector/src/main/java/org/apache/arrow/vector/ipc/ArrowWriter.java
lidavidm: Agreed, can we consolidate?

lidavidm (changes requested):

java/compression/src/test/java/org/apache/arrow/compression/TestArrowReaderWriterWithCompression.java
davisusanibar: Deleted

java/vector/src/main/java/org/apache/arrow/vector/ipc/ArrowWriter.java
davisusanibar: Updated

java/vector/src/main/java/org/apache/arrow/vector/ipc/ArrowFileWriter.java
```diff
@@ -73,8 +73,8 @@ public ArrowFileWriter(VectorSchemaRoot root, DictionaryProvider provider, Writa
 
   public ArrowFileWriter(VectorSchemaRoot root, DictionaryProvider provider, WritableByteChannel out,
                          Map<String, String> metaData, IpcOption option, CompressionCodec.Factory compressionFactory,
-                         CompressionUtil.CodecType codecType) {
-    super(root, provider, out, option, compressionFactory, codecType);
+                         CompressionUtil.CodecType codecType, int compressionLevel) {
```
lidavidm, line 76: Well, now this is a breaking change, since the old constructor isn't available anymore.

java/vector/src/main/java/org/apache/arrow/vector/ipc/ArrowStreamWriter.java
```diff
@@ -76,12 +76,13 @@ public ArrowStreamWriter(VectorSchemaRoot root, DictionaryProvider provider, Wri
    * @param option IPC write options
    * @param compressionFactory Compression codec factory
    * @param codecType Codec type
+   * @param compressionLevel Compression level
    * @param out WritableByteChannel for writing.
    */
   public ArrowStreamWriter(VectorSchemaRoot root, DictionaryProvider provider, WritableByteChannel out,
                            IpcOption option, CompressionCodec.Factory compressionFactory,
-                           CompressionUtil.CodecType codecType) {
```
lidavidm, line 83: Ditto here.

java/vector/src/main/java/org/apache/arrow/vector/ipc/ArrowWriter.java
```diff
@@ -72,7 +72,7 @@ protected ArrowWriter(VectorSchemaRoot root, DictionaryProvider provider, Writab
   }
 
   protected ArrowWriter(VectorSchemaRoot root, DictionaryProvider provider, WritableByteChannel out, IpcOption option) {
-    this(root, provider, out, option, NoCompressionCodec.Factory.INSTANCE, CompressionUtil.CodecType.NO_COMPRESSION);
+    this(root, provider, out, option, NoCompressionCodec.Factory.INSTANCE, CompressionUtil.CodecType.NO_COMPRESSION, 3);
```
lidavidm, line 75: Use a constant for the integer?

java/vector/src/main/java/org/apache/arrow/vector/ipc/ArrowWriter.java
lidavidm: This now breaks compatibility. I don't think that's what Gang was suggesting with a delegate constructor.

davisusanibar: Oh, I've decided to refactor constructor to add new parameter needed, .... otherwise will be to maintain the constructor as it is now and add a new constructor for new compression_level args needed. Please let me know how to proceed?

lidavidm: As already suggested: why not add a new constructor, and have the old ones delegate to the new one? That way we can avoid code duplication.

java/vector/src/main/java/org/apache/arrow/vector/ipc/ArrowFileWriter.java
davisusanibar: Updated

java/vector/src/main/java/org/apache/arrow/vector/ipc/ArrowWriter.java
davisusanibar: Updated

java/vector/src/main/java/org/apache/arrow/vector/ipc/ArrowWriter.java
```diff
@@ -53,6 +53,8 @@ public abstract class ArrowWriter implements AutoCloseable {
 
   protected static final Logger LOGGER = LoggerFactory.getLogger(ArrowWriter.class);
 
+  protected static final int compressionLevel = 3;
```
lidavidm, line 56: Constants should use UPPER_SNAKE_CASE.

lidavidm: And this should be clearly named DEFAULT_COMPRESSION_LEVEL

lidavidm (changes requested):

java/vector/src/main/java/org/apache/arrow/vector/ipc/ArrowWriter.java
wgtmac: The default level is tricky here:
- LZ4 default level is 1 with range from 1 to 9
- ZSTD default level is 3 with range from 1 to 22 (also negative levels -1 to -5 are supported)

There is a [discussion](https://github.com/apache/arrow/issues/35287) on the C++ side to add support of `Codec::Options`. Would it be good to do similar things on the Java side so we can accommodate different default levels per codec?

lidavidm: Hmm, we could leave that for a different PR?

Or we could use Optional<Integer> here and switch between constructors that way?

wgtmac: +1 for Optional

davisusanibar: > The default level is tricky here:
> 
> * LZ4 default level is 1 with range from 1 to 9
> * ZSTD default level is 3 with range from 1 to 22 (also negative levels -1 to -5 are supported)
> 
> There is a [discussion](https://github.com/apache/arrow/issues/35287) on the C++ side to add support of `Codec::Options`. Would it be good to do similar things on the Java side so we can accommodate different default levels per codec?

Uhmm, our current compression version only support compression level for Zstd not for LZ4. I was reviewing and is close the same as [Spark Compression Level Zstd](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/io/CompressionCodec.scala).

davisusanibar: > +1 for Optional

Could you help me on this about where o how?

wgtmac: We can use an `Optional<int> level` parameter and pass it to specific codecs via `createCodec`. `Lz4CompressionCodec` and `ZstdCompressionCodec` can respect the value when it is set or use specific default level if value is unset.

BTW, I just checked that `FramedLZ4CompressorOutputStream` does not have the concept of level. It requires user to fine-tune the parameters: https://commons.apache.org/proper/commons-compress/apidocs/org/apache/commons/compress/compressors/lz4/FramedLZ4CompressorOutputStream.Parameters.html

davisusanibar: Updated

java/vector/src/main/java/org/apache/arrow/vector/ipc/ArrowFileWriter.java
```diff
@@ -74,7 +75,13 @@ public ArrowFileWriter(VectorSchemaRoot root, DictionaryProvider provider, Writa
   public ArrowFileWriter(VectorSchemaRoot root, DictionaryProvider provider, WritableByteChannel out,
                          Map<String, String> metaData, IpcOption option, CompressionCodec.Factory compressionFactory,
                          CompressionUtil.CodecType codecType) {
-    super(root, provider, out, option, compressionFactory, codecType);
+    this(root, provider, out, metaData, option, compressionFactory, codecType, Optional.ofNullable(null));
```
wgtmac, line 78:
```suggestion
    this(root, provider, out, metaData, option, compressionFactory, codecType, Optional.empty());
```

java/vector/src/main/java/org/apache/arrow/vector/ipc/ArrowStreamWriter.java
```diff
@@ -81,7 +82,25 @@ public ArrowStreamWriter(VectorSchemaRoot root, DictionaryProvider provider, Wri
   public ArrowStreamWriter(VectorSchemaRoot root, DictionaryProvider provider, WritableByteChannel out,
                            IpcOption option, CompressionCodec.Factory compressionFactory,
                            CompressionUtil.CodecType codecType) {
-    super(root, provider, out, option, compressionFactory, codecType);
+    this(root, provider, out, option, compressionFactory, codecType, Optional.ofNullable(null));
```
wgtmac, line 85: ditto

wgtmac (approved):
LGTM. Thanks!

lidavidm:
There seems to be a build error?

davisusanibar:
@lidavidm could you help me to retry job failed? or Is it needed to review that?

wgtmac:
> @lidavidm could you help me to retry job failed? or Is it needed to review that?

I just re-triggered the failed one.

lidavidm (approved):

PR #34874, opened 2023-04-04
GH-34861: [C++] Move GTest std version check into SYSTEM if branch

### LINKED ISSUE #34861: [Benchmarks] All cpp microbenchmarks are failing with new GTest CMake Error

Since the PR https://github.com/apache/arrow/pull/34765 was merged, the `cpp-micro` benchmarks are failing with the following error:

```
CMake Error at cmake_modules/ThirdpartyToolchain.cmake:2236 (message):
  System GTest is built with a C++ standard lower than 17.  Use bundled GTest
  via passing in CMake flag

  -DGTest_SOURCE="BUNDLED"
```
[Here](https://buildkite.com/apache-arrow/arrow-bci-benchmark-on-arm64-m6g-linux-compute/builds/2413) is an example Buildkite build.

This error message was explicitly added in that PR, so we'll need to take action to correct it.

cc @HaochengLIU

### Component(s)

Benchmarking, C++

HaochengLIU:
### Rationale for this change
Developers have reported that the current GTest std version check will break cpp microbenchmarks in BUNDLE mode.

Move the check into SYSTEM if branch so that it gets triggered in a narrowed scope.

### Are these changes tested?
Tested locally and will be covered by CI.

No

* Closes: #34861

### DIFF
--- a/cpp/cmake_modules/ThirdpartyToolchain.cmake
+++ b/cpp/cmake_modules/ThirdpartyToolchain.cmake
@@ -2229,15 +2229,15 @@ if(ARROW_TESTING)
                      1.10.0
                      USE_CONFIG
                      ${GTEST_USE_CONFIG})
-  get_target_property(gtest_cxx_standard GTest::gtest INTERFACE_COMPILE_FEATURES)
-
-  if((${gtest_cxx_standard} STREQUAL "cxx_std_11") OR (${gtest_cxx_standard} STREQUAL
-                                                       "cxx_std_14"))
-    message(FATAL_ERROR "System GTest is built with a C++ standard lower than 17. Use bundled GTest via passing in CMake flag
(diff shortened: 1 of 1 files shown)


HaochengLIU:
I dont believe the failing tests are related to my change. More like corrupted build farm.

kou (approved):
+1

This will not solve #34861 but I merge this because this is a right change.
Our bundled GoogleTest must be built with the same C++ standard as Arrow C++ because we specify `-DCMAKE_CXX_STANDARD=${CMAKE_CXX_STANDARD}` via `EP_COMMON_CMAKE_ARGS`.
But `build_gtest` doesn't set `INTERFACE_COMPILE_FEATURES`. So the check is meaningless for bundled GoogleTest.

PR #34878, opened 2023-04-04
GH-15070: [Python][CI] Compatibility with pandas 2.0

### LINKED ISSUE #15070: [Python][CI] Remove temporary skips for pandas 2.0 development version
https://github.com/apache/arrow/pull/15048 added some skips for regressions in pandas main. Once those are fixed, we should remove the skips on our side.

jorisvandenbossche:

- The issue with numpy 1.25 in the assert equal helper was fixed in pandas 1.5.3 -> removing the skip (in theory can still run into this error when using an older pandas version with the latest numpy, but that's not something you should do)
- Casting tz-aware strings to datetime64[ns] was not fixed in pandas (https://github.com/pandas-dev/pandas/issues/50140) -> updating our implementation to work around it
- Casting to numpy string dtype (https://github.com/pandas-dev/pandas/issues/50127) is not yet fixed -> updating the skip

No
* Closes: #15070

### DIFF
--- a/python/pyarrow/pandas_compat.py
+++ b/python/pyarrow/pandas_compat.py
@@ -1148,8 +1148,7 @@ def _reconstruct_columns_from_metadata(columns, column_indexes):
         if pandas_dtype == "datetimetz":
             tz = pa.lib.string_to_tzinfo(
                 column_indexes[0]['metadata']['timezone'])
-            dt = level.astype(numpy_dtype)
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
@github-actions crossbow submit test-conda-python-3.8-pandas-nightly test-conda-python-3.9-pandas-upstream_devel

AlenkaF (approved):

PR #34885, opened 2023-04-04
GH-34786: [C++] Fix output schema calculated by Substrait consumer for AggregateRel

### LINKED ISSUE #34786: [C++] Output schema calculated by Substrait consumer for aggregate rel seems incorrect.

The code calculating the output schema is here:

```
      FieldVector output_fields;
      output_fields.reserve(key_field_ids.size() + measure_size);
      // extract aggregate fields to output schema
      for (const auto& agg_src_fieldset : agg_src_fieldsets) {
        for (int field : agg_src_fieldset) {
          output_fields.emplace_back(input_schema->field(field));
        }
      }
      // extract key fields to output schema
      for (int key_field_id : key_field_ids) {
        output_fields.emplace_back(input_schema->field(key_field_id));
      }

      std::shared_ptr<Schema> aggregate_schema = schema(std::move(output_fields));
```

This appears to have two issues:

 * It is inserting the key fields after the measure fields
 * It is inserting measure fields based on the function inputs and not the function outputs

I suspect we are getting away with it in many cases because we are not applying projection / emit after the aggregate.  At the very least, we should add some test cases that do this so we can verify the output schema is correct.  If there is indeed an issue then we should also fix that.

### Component(s)

C++

rtpsw:
See https://github.com/apache/arrow/issues/34786
* Closes: #34786

### DIFF
--- a/cpp/src/arrow/acero/aggregate_node.cc
+++ b/cpp/src/arrow/acero/aggregate_node.cc
@@ -21,6 +21,7 @@
 #include <unordered_map>
 #include <unordered_set>
 
+#include "arrow/acero/aggregate_node.h"
 #include "arrow/acero/exec_plan.h"
 #include "arrow/acero/options.h"
 #include "arrow/acero/query_context.h"
@@ -85,8 +86,43 @@ std::vector<TypeHolder> ExtendWithGroupIdType(const std::vector<TypeHolder>& in_
   return aggr_in_types;
 }
 
-Result<const HashAggregateKernel*> GetKernel(ExecContext* ctx, const Aggregate& aggregate,
+void DefaultAggregateOptions(Aggregate* aggregate_ptr,
+                             const std::shared_ptr<Function> function) {
+  Aggregate& aggregate = *aggregate_ptr;
+  if (aggregate.options == nullptr) {
+    DCHECK(!function->doc().options_required);
+    const auto* default_options = function->default_options();
+    if (default_options) {
+      aggregate.options = default_options->Copy();
+    }
+  }
+}
+
+using GetKernel = std::function<Result<const Kernel*>(ExecContext*, Aggregate*,
+                                                      const std::vector<TypeHolder>&)>;
+
+Result<const Kernel*> GetScalarAggregateKernel(ExecContext* ctx, Aggregate* aggregate_ptr,
+                                               const std::vector<TypeHolder>& in_types) {
+  Aggregate& aggregate = *aggregate_ptr;
+  ARROW_ASSIGN_OR_RAISE(auto function,
+                        ctx->func_registry()->GetFunction(aggregate.function));
+  if (function->kind() != Function::SCALAR_AGGREGATE) {
+    if (function->kind() == Function::HASH_AGGREGATE) {
+      return Status::Invalid("The provided function (", aggregate.function,
+                             ") is a hash aggregate function.  Since there are no "
+                             "keys to group by, a scalar aggregate function was "
+                             "expected (normally these do not start with hash_)");
+    }
+    return Status::Invalid("The provided function(", aggregate.function,
+                           ") is not an aggregate function");
+  }
+  DefaultAggregateOptions(&aggregate, function);
+  return function->DispatchExact(in_types);
+}
+
+Result<const Kernel*> GetHashAggregateKernel(ExecContext* ctx, Aggregate* aggregate_ptr,
                                              const std::vector<TypeHolder>& in_types) {
+  Aggregate& aggregate = *aggregate_ptr;
   const auto aggr_in_types = ExtendWithGroupIdType(in_types);
   ARROW_ASSIGN_OR_RAISE(auto function,
                         ctx->func_registry()->GetFunction(aggregate.function));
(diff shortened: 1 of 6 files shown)


rtpsw:
cc @westonpace @icexelloss

cpp/src/arrow/acero/aggregate_node.cc
```diff
@@ -85,8 +86,43 @@ std::vector<TypeHolder> ExtendWithGroupIdType(const std::vector<TypeHolder>& in_
   return aggr_in_types;
 }
 
-Result<const HashAggregateKernel*> GetKernel(ExecContext* ctx, const Aggregate& aggregate,
+void DefaultAggregateOptions(Aggregate* aggregate_ptr,
```
icexelloss, line 89: Why do we need to add this?

cpp/src/arrow/acero/aggregate_node.cc
```diff
@@ -85,8 +86,43 @@ std::vector<TypeHolder> ExtendWithGroupIdType(const std::vector<TypeHolder>& in_
...
+    if (default_options) {
+      aggregate.options = default_options->Copy();
+    }
+  }
+}
+
+using GetKernel = std::function<Result<const Kernel*>(ExecContext*, Aggregate*,
+                                                      const std::vector<TypeHolder>&)>;
+
+Result<const Kernel*> GetScalarAggregateKernel(ExecContext* ctx, Aggregate* aggregate_ptr,
```
icexelloss, line 104: What is this used for?

cpp/src/arrow/acero/aggregate_node.cc
```diff
@@ -122,53 +158,91 @@ Result<std::unique_ptr<KernelState>> InitKernel(const HashAggregateKernel* kerne
...
   if (aggregates.size() != in_types.size()) {
     return Status::Invalid(aggregates.size(), " aggregate functions were specified but ",
                            in_types.size(), " arguments were provided.");
   }
 
-  std::vector<const HashAggregateKernel*> kernels(in_types.size());
+  std::vector<const Kernel*> kernels(in_types.size());
   for (size_t i = 0; i < aggregates.size(); ++i) {
-    ARROW_ASSIGN_OR_RAISE(kernels[i], GetKernel(ctx, aggregates[i], in_types[i]));
+    ARROW_ASSIGN_OR_RAISE(kernels[i], get_kernel(ctx, &aggregates[i], in_types[i]));
```
icexelloss, line 183: Why this change?

cpp/src/arrow/acero/aggregate_node.cc
```diff
@@ -703,7 +635,7 @@ class GroupByNode : public ExecNode, public TracedNode {
                           {"function.kind", std::string(kind_name()) + "::Consume"}});
       auto ctx = plan_->query_context()->exec_context();
       KernelContext kernel_ctx{ctx};
-      kernel_ctx.SetState(state->agg_states[i].get());
+      kernel_ctx.SetState(state->agg_states[i][0].get());
```
icexelloss, line 638: Why this change?

icexelloss:
@rtpsw Not sure I follow what you are doing - seems like a lot of refactor is done. Can you explain your approach?

cpp/src/arrow/acero/aggregate_node.h
```diff
@@ -0,0 +1,76 @@
...
+namespace aggregate {
+
+using compute::Aggregate;
+using compute::default_exec_context;
+using compute::ExecContext;
+using compute::Kernel;
+using compute::KernelState;
+using compute::RowSegmenter;
+
+struct ARROW_ACERO_EXPORT AggregateNodeArgs {
```
icexelloss, line 38: Why do we need this?

rtpsw:
> @rtpsw Not sure I follow what you are doing - seems like a lot of refactor is done. Can you explain your approach?

In this PR, the main goal is a single method `MakeOutputSchema` providing the output schema for an aggregation. The problem is that the original code has two classes , `ScalarAggregateNode` and `GroupByNode`, for aggregation that do not share much code between them for the purpose of constructing the output schema. To prepare the stage, I started with refactoring the original code to make them share code for this purpose.For this, I needed to encapsulate certain differences between them:
- Get kernel: This is [encapsulated by `GetKernel`](https://github.com/apache/arrow/pull/34885/files#diff-806b2b905d7908f2ce860f38eca2090b9504753ebd72d5fef5f97bd2611359b1R101-R102). The two implementations are `GetScalarAggregateKernel` and `GetHashAggregateKernel`. The latter has the function dispatch on the [types extended with the group-id](https://github.com/apache/arrow/pull/34885/files#diff-806b2b905d7908f2ce860f38eca2090b9504753ebd72d5fef5f97bd2611359b1R81-R87).
- Init kernel: This is [encapsulated by `InitKernel`](https://github.com/apache/arrow/pull/34885/files#diff-806b2b905d7908f2ce860f38eca2090b9504753ebd72d5fef5f97bd2611359b1R143-R144). The two implementations are `InitScalarAggregateKernel` and `InitHashAggregateKernel`. The latter has the kernel-args configured using the types extended with the group-id.
- Resolve kernels: This is [encapsulated by `ResolveKernels`](https://github.com/apache/arrow/pull/34885/files#diff-806b2b905d7908f2ce860f38eca2090b9504753ebd72d5fef5f97bd2611359b1R218-R221). The two implementations are `ResolveScalarAggregateKernels` and `ResolveHashAggregateKernels`. The latter resolves each kernel using the types extended with the group-id.

Additional parts of the refactoring are:
- Adding `MakeAggregateNodeArgs` as a common method for setting up the arguments needed for constructing an aggregation node, whether it is a `ScalarAggregateNode` or a `GroupByNode`.
- Cleaning up `ScalarAggregateNode::Make` and `GroupByNode::Make` to use the above consistently.
- Adding `MakeOutputSchema` that uses `MakeAggregateNodeArgs` to return the output schema that the aggregation node is constructed with.

rtpsw:
Replaced by #34904

PR #34889, opened 2023-04-04
GH-34888: [C++][Parquet] Writer supports adding extra kv meta

### LINKED ISSUE #34888: [C++][Parquet] Writer cannot add extra key-value metadata once opened

Parquet specs support storing key-value metadata provided by the user. However, the parquet-cpp writer can only set it via `ParquetFileWriter::Open()`. Sometimes user may want to add extra information to it while writing. So it is good to support adding extra key-value metadata any time before closing the file writer.

### Component(s)

C++, Parquet

wgtmac:
### Rationale for this change

Parquet specs support storing key-value metadata provided by the user. However, the parquet-cpp writer can only set it via ParquetFileWriter::Open(). Sometimes user may want to add extra information to it while writing. So it is good to support adding extra key-value metadata any time before closing the file writer.

Add a new interface `void AddKeyValueMetadata(std::shared_ptr<const KeyValueMetadata> key_value_metadata)` to the `ParquetFileWriter` class. User can now add more key-value metadata to the file if not closed.

### Are these changes tested?

Added a new `Metadata.TestAddKeyValueMetadata` test to verify key-value metadata added before closing the writer are well preserved.

Yes, user can add custom key-value metadata whenever writer is not closed.
* Closes: #34888

### DIFF
--- a/cpp/src/parquet/file_writer.cc
+++ b/cpp/src/parquet/file_writer.cc
@@ -17,13 +17,13 @@
 
 #include "parquet/file_writer.h"
 
-#include <cstddef>
 #include <memory>
 #include <ostream>
 #include <string>
 #include <utility>
 #include <vector>
 
+#include "arrow/util/key_value_metadata.h"
 #include "parquet/column_writer.h"
 #include "parquet/encryption/encryption_internal.h"
 #include "parquet/encryption/internal_file_encryptor.h"
@@ -338,7 +338,7 @@ class FileSerializer : public ParquetFileWriter::Contents {
       auto file_encryption_properties = properties_->file_encryption_properties();
 
       if (file_encryption_properties == nullptr) {  // Non encrypted file.
-        file_metadata_ = metadata_->Finish();
+        file_metadata_ = metadata_->Finish(key_value_metadata_);
         WriteFileMetaData(*file_metadata_, sink_.get());
       } else {  // Encrypted file
         CloseEncryptedFile(file_encryption_properties);
@@ -376,6 +376,15 @@ class FileSerializer : public ParquetFileWriter::Contents {
 
   RowGroupWriter* AppendBufferedRowGroup() override { return AppendRowGroup(true); }
 
+  void AddKeyValueMetadata(
+      std::shared_ptr<const KeyValueMetadata> key_value_metadata) override {
+    if (key_value_metadata_ == nullptr) {
+      key_value_metadata_ = std::move(key_value_metadata);
+    } else if (key_value_metadata != nullptr) {
+      key_value_metadata_ = key_value_metadata_->Merge(*key_value_metadata);
+    }
+  }
+
   ~FileSerializer() override {
     try {
       Close();
@@ -394,7 +403,7 @@ class FileSerializer : public ParquetFileWriter::Contents {
         properties_(std::move(properties)),
         num_row_groups_(0),
         num_rows_(0),
-        metadata_(FileMetaDataBuilder::Make(&schema_, properties_, key_value_metadata_)) {
+        metadata_(FileMetaDataBuilder::Make(&schema_, properties_)) {
     PARQUET_ASSIGN_OR_THROW(int64_t position, sink_->Tell());
     if (position == 0) {
       StartFile();
@@ -407,7 +416,7 @@ class FileSerializer : public ParquetFileWriter::Contents {
     // Encrypted file with encrypted footer
     if (file_encryption_properties->encrypted_footer()) {
       // encrypted footer
-      file_metadata_ = metadata_->Finish();
+      file_metadata_ = metadata_->Finish(key_value_metadata_);
 
       PARQUET_ASSIGN_OR_THROW(int64_t position, sink_->Tell());
       uint64_t metadata_start = static_cast<uint64_t>(position);
@@ -422,7 +431,7 @@ class FileSerializer : public ParquetFileWriter::Contents {
           sink_->Write(reinterpret_cast<uint8_t*>(&footer_and_crypto_len), 4));
       PARQUET_THROW_NOT_OK(sink_->Write(kParquetEMagic, 4));
     } else {  // Encrypted file with plaintext footer
-      file_metadata_ = metadata_->Finish();
+      file_metadata_ = metadata_->Finish(key_value_metadata_);
       auto footer_signing_encryptor = file_encryptor_->GetFooterSigningEncryptor();
       WriteEncryptedFileMetadata(*file_metadata_, sink_.get(), footer_signing_encryptor,
                                  false);
@@ -611,6 +620,13 @@ RowGroupWriter* ParquetFileWriter::AppendBufferedRowGroup() {
 
 RowGroupWriter* ParquetFileWriter::AppendRowGroup(int64_t num_rows) {
   return AppendRowGroup();
+}
+
+void ParquetFileWriter::AddKeyValueMetadata(
+    std::shared_ptr<const KeyValueMetadata> key_value_metadata) {
+  if (contents_) {
+    contents_->AddKeyValueMetadata(std::move(key_value_metadata));
+  }
 }
 
 const std::shared_ptr<WriterProperties>& ParquetFileWriter::properties() const {
--- a/cpp/src/parquet/file_writer.cc
+++ b/cpp/src/parquet/file_writer.cc
@@ -24,6 +24,7 @@
 #include <vector>
 
 #include "arrow/util/key_value_metadata.h"
+#include "arrow/util/logging.h"
 #include "parquet/column_writer.h"
 #include "parquet/encryption/encryption_internal.h"
 #include "parquet/encryption/internal_file_encryptor.h"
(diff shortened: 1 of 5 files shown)


wgtmac:
@wjones127 @westonpace Please take a look when you have time. Thanks!

cpp/src/parquet/metadata.cc
```diff
@@ -1743,13 +1743,11 @@ void RowGroupMetaDataBuilder::Finish(int64_t total_bytes_written,
 // TODO(PARQUET-595) Support key_value_metadata
```
wjones127, line 1743: I think we can remove this?

cpp/src/parquet/metadata.h
```diff
@@ -510,24 +510,23 @@ class PARQUET_EXPORT FileMetaDataBuilder {
  public:
   // API convenience to get a MetaData reader
   static std::unique_ptr<FileMetaDataBuilder> Make(
-      const SchemaDescriptor* schema, std::shared_ptr<WriterProperties> props,
-      std::shared_ptr<const KeyValueMetadata> key_value_metadata = NULLPTR);
+      const SchemaDescriptor* schema, std::shared_ptr<WriterProperties> props);
```
wjones127, line 513: I know it makes the code more complex, but I think for now we should create a new overload that drops the `key_value_metadata` parameter and add `PARQUET_DEPRECATED` marker to the existing constructor.

cpp/src/parquet/file_writer.cc
```diff
@@ -356,6 +355,15 @@ class FileSerializer : public ParquetFileWriter::Contents {
 
   RowGroupWriter* AppendBufferedRowGroup() override { return AppendRowGroup(true); }
 
+  void AddKeyValueMetadata(
+      std::shared_ptr<const KeyValueMetadata> key_value_metadata) override {
+    if (key_value_metadata_ == nullptr) {
+      key_value_metadata_ = std::move(key_value_metadata);
+    } else if (key_value_metadata != nullptr) {
+      key_value_metadata_ = key_value_metadata_->Merge(*key_value_metadata);
```
mapleFU, line 363: should we add a DCHECK for `key_value_metadata != nullptr`?

cpp/src/parquet/file_writer.h
```diff
@@ -209,6 +212,12 @@ class PARQUET_EXPORT ParquetFileWriter {
   /// until the next call to AppendRowGroup or AppendBufferedRowGroup or Close.
   RowGroupWriter* AppendBufferedRowGroup();
 
+  /// \brief Add key-value metadata to the file.
+  /// \param[in] key_value_metadata the metadata to add.
+  /// \note This will overwrite any existing metadata with the same key.
+  /// It will not take effect if Close() has been called.
```
mapleFU, line 218: Should we throw ex if closed?

cpp/src/parquet/file_writer.cc
```diff
@@ -356,6 +355,15 @@ class FileSerializer : public ParquetFileWriter::Contents {
 
   RowGroupWriter* AppendBufferedRowGroup() override { return AppendRowGroup(true); }
 
+  void AddKeyValueMetadata(
+      std::shared_ptr<const KeyValueMetadata> key_value_metadata) override {
```
mapleFU, line 359: Should it better be a `const std::shared_ptr<const KeyValueMetadata>& key_value_metadata`? Because it doesn't force copy everytime

cpp/src/parquet/metadata.h
mapleFU: Agree

mapleFU (commented):
Rest LGTM

cpp/src/parquet/metadata.cc
wgtmac: Fixed

cpp/src/parquet/metadata.h
wgtmac: Sounds reasonable. Fixed.

The only caveat is that the old overload simply ignores the input `key_value_metadata` which may confuses users.

cpp/src/parquet/file_writer.cc
wgtmac: Fixed

cpp/src/parquet/file_writer.h
wgtmac: Good point! Fixed

cpp/src/parquet/file_writer.cc
wgtmac: Fixed

cpp/src/parquet/metadata.h
```diff
@@ -508,26 +508,29 @@ class PARQUET_EXPORT RowGroupMetaDataBuilder {
 
 class PARQUET_EXPORT FileMetaDataBuilder {
  public:
-  // API convenience to get a MetaData reader
+  // API convenience to get a MetaData builder
+  PARQUET_DEPRECATED("Deprecated. Use overload without KeyValueMetadata instead.")
```
wjones127, line 512: It's nice to say when it was deprecated.
```suggestion
  PARQUET_DEPRECATED("Deprecated in 12.0.0. Use overload without KeyValueMetadata instead.")
```

cpp/src/parquet/metadata.h
wjones127: Ignoring the parameter seems like an even worse breaking change than removing it. Could we simply make this backwards compatible? Either `key_value_metadata` in `Finish` in override, or it should merge with the one provided in the constructor.

wjones127 (changes requested):

wgtmac:
Fixed it to be backward compatible. Please take a look again. Thanks! @wjones127

wgtmac:
https://github.com/apache/arrow/actions/runs/4662936307/jobs/8253813543?pr=34889 is failing:
```
+ doxygen
/arrow/cpp/apidoc /
warning: Tag 'OUTPUT_TEXT_DIRECTION' at line 119 of file 'Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'COLS_IN_ALPHA_INDEX' at line 1118 of file 'Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'FORMULA_TRANSPARENT' at line 1557 of file 'Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'LATEX_SOURCE_CODE' at line 1865 of file 'Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'RTF_SOURCE_CODE' at line 1955 of file 'Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'DOCBOOK_PROGRAMLISTING' at line 2060 of file 'Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'CLASS_DIAGRAMS' at line 2257 of file 'Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'DOT_FONTNAME' at line 2299 of file 'Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'DOT_FONTSIZE' at line 2306 of file 'Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'DOT_TRANSPARENT' at line 2530 of file 'Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
/arrow/cpp/src/parquet/metadata.h:531: error: Found ';' while parsing initializer list! (doxygen could be confused by a macro call without semicolon) (warning treated as error, aborting now)
1
Error: `docker-compose --file /home/runner/work/arrow/arrow/docker-compose.yml run --rm conda-python-docs` exited with a non-zero exit code 1, see the process log above.
```

cpp/src/parquet/file_writer.cc
```diff
@@ -376,6 +377,16 @@ class FileSerializer : public ParquetFileWriter::Contents {
 
   RowGroupWriter* AppendBufferedRowGroup() override { return AppendRowGroup(true); }
 
+  void AddKeyValueMetadata(
+      const std::shared_ptr<const KeyValueMetadata>& key_value_metadata) override {
+    if (key_value_metadata_ == nullptr) {
+      key_value_metadata_ = std::move(key_value_metadata);
+    } else if (key_value_metadata != nullptr) {
+      ARROW_DCHECK(key_value_metadata != nullptr);
```
wjones127, line 385: That seems unnecessary

```suggestion
```

wjones127 (approved):
Just one minor comment, otherwise looks good!

cpp/src/parquet/metadata.h
```diff
@@ -523,10 +523,14 @@ struct PageIndexLocation {
 
 class PARQUET_EXPORT FileMetaDataBuilder {
  public:
-  // API convenience to get a MetaData reader
+  ARROW_DEPRECATED("Deprecated in 12.0.0. Use overload without KeyValueMetadata instead.")
```
wgtmac, line 526:
```suggestion
  ARROW_DEPRECATED("Deprecated in 13.0.0. Use overload without KeyValueMetadata instead.")
```
Should I change this since it may not be included in the 12.0.0 release? @wjones127

wjones127 (approved):

wjones127:
I merged this and then realized I have just missed the window for 12.0.0. @raulcd would it be possible to add this to the release branch? Otherwise I'll put up a PR to adjust the text of the deprecation warning added in this PR (which says a method was deprecated in 12.0.0).

raulcd:
> @raulcd would it be possible to add this to the release branch?

I will add it to 12.0.0

PR #34894, opened 2023-04-04
GH-34868: [Python] Share docstrings between classes

### LINKED ISSUE #34868: [Python] Sharing docstrings between classes

PyArrow duplicates a lot of documentation in order to provide explicit docstring examples. Let's reduce the duplication of docstrings by providing a way to share docstrings between classes. See the way `pandas` did this as an example: https://pandas.pydata.org/docs/development/contributing_docstring.html#sharing-docstrings

A good example of duplication in PyArrow are the classes `Table` and `RecordBatch`. They both provide similar, sometimes identical, top-level implementations and docstrings, while typically only differing in low-level C++ implementation.

Here is an example of duplicative docstring descriptions.
`class RecordBatch:`
```
    @property
    def nbytes(self):
        """
        Total number of bytes consumed by the elements of the record batch.

        In other words, the sum of bytes from all buffer ranges referenced.

        Unlike `get_total_buffer_size` this method will account for array
        offsets.

        If buffers are shared between arrays then the shared
        portion will only be counted multiple times.

        The dictionary of dictionary arrays will always be counted in their
        entirety even if the array only references a portion of the dictionary.

        Examples
        --------
        >>> import pyarrow as pa
        >>> n_legs = pa.array([2, 2, 4, 4, 5, 100])
        >>> animals = pa.array(["Flamingo", "Parrot", "Dog", "Horse", "Brittle stars", "Centipede"])
        >>> batch = pa.RecordBatch.from_arrays([n_legs, animals],
        ...                                     names=["n_legs", "animals"])
        >>> batch.nbytes
        116
        """
        ...
```

`class Table:`
```
    @property
    def nbytes(self):
        """
        Total number of bytes consumed by the elements of the table.

        In other words, the sum of bytes from all buffer ranges referenced.

        Unlike `get_total_buffer_size` this method will account for array
        offsets.

        If buffers are shared between arrays then the shared
        portion will only be counted multiple times.

        The dictionary of dictionary arrays will always be counted in their
        entirety even if the array only references a portion of the dictionary.

        Examples
        --------
        >>> import pyarrow as pa
        >>> import pandas as pd
        >>> df = pd.DataFrame({'n_legs': [None, 4, 5, None],
        ...                    'animals': ["Flamingo", "Horse", None, "Centipede"]})
        >>> table = pa.Table.from_pandas(df)
        >>> table.nbytes
        72
        """
        ...
```

### Component(s)

Python

danepitkin:
### Rationale for this change

Python classes sometimes duplicate docstrings, but change one word such as class name. Add a decorator function as a utility to help deduplicate docstring descriptions. Only works in Python. Does not work in Cython due to this CPython issue https://github.com/python/cpython/issues/91309.

Add a decorator `@doc` that can copy, concatenate, and/or format docstrings between classes.

### Are these changes tested?

Tests added. 

```
>>> import pyarrow
>>> from pyarrow.filesystem import FileSystem, LocalFileSystem, DaskFileSystem, S3FSWrapper
>>> from pyarrow.hdfs import HadoopFileSystem
>>> for fs in [FileSystem, LocalFileSystem, DaskFileSystem, S3FSWrapper, HadoopFileSystem]:
...     print(fs.__name__)
...     print(fs.isdir.__doc__)
... 
FileSystem

        Return True if path is a directory.

        Parameters
        ----------
        path : str
            Path to check.
        
LocalFileSystem

Return True if path is a directory.

Parameters
----------
path : str
    Path to check.

DaskFileSystem

Return True if path is a directory.

Parameters
----------
path : str
    Path to check.

S3FSWrapper

Return True if path is a directory.

Parameters
----------
path : str
    Path to check.

HadoopFileSystem

Return True if path is a directory.

Parameters
----------
path : str
    Path to check.

```
Note that `FileSystem.isdir.__doc__` is not dedented because it does not use the `@doc` decorator.

No
* Closes: #34868

### DIFF
--- a/python/pyarrow/types.py
+++ b/python/pyarrow/types.py
@@ -23,6 +23,7 @@ from pyarrow.lib import (is_boolean_value,  # noqa
                          is_float_value)
 
 import pyarrow.lib as lib
+from pyarrow.util import doc
 
 
 _SIGNED_INTEGER_TYPES = {lib.Type_INT8, lib.Type_INT16, lib.Type_INT32,
@@ -43,9 +44,10 @@ _NESTED_TYPES = {lib.Type_LIST, lib.Type_LARGE_LIST, lib.Type_STRUCT,
                  lib.Type_MAP} | _UNION_TYPES
 
 
+@doc(datatype="null")
 def is_null(t):
     """
-    Return True if value is an instance of a null type.
+    Return True if value is an instance of type: {datatype}.
 
     Parameters
     ----------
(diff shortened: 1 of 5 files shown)


AlenkaF (approved):
I think this is a good utility for writing docstrings 👍
Do we know if there are other places in pyarrow that would benefit from it?

danepitkin:
`arrow/python/pyarrow/types.py` could be a good option! This is the first one that stood out to me, but I haven't yet thoroughly reviewed all python code. Most of the benefit would be in Cython code unfortunately..

jorisvandenbossche:
>  Do we know if there are other places in pyarrow that would benefit from it?

I am not sure we have many places that currently are not in cython, and have repetition in docstrings. Looking through our python files, we do already have some manual templating in parquet/core.py, ipc.py or orc.py (eg to share the explanation of certain parameters between multiple docstrings), but I don't know if they would fit this pattern for using `@doc`.

`pyarrow.filesystem` is also deprecated, so that will be removed in a few releases (not that we can't improve the docs with this PR in the meantime). `types.py` indeed looks like the main obvious candidate where this could be useful.

danepitkin:
Updated `types.py`:
```
>>> import pyarrow
>>> from pyarrow.types import *
>>> print(is_null.__doc__)

Return True if value is an instance of type: null.

Parameters
----------
t : DataType

>>> print(is_boolean.__doc__)

Return True if value is an instance of type: boolean.

Parameters
----------
t : DataType

>>> print(is_map.__doc__)

Return True if value is an instance of type: logical map.

Parameters
----------
t : DataType

>>> print(is_unicode.__doc__)

Alias for is_string.

Parameters
----------
t : DataType

>>> print(is_large_unicode.__doc__)

Alias for is_large_string.

Parameters
----------
t : DataType
```

python/pyarrow/types.py
```diff
@@ -54,351 +56,165 @@ def is_null(t):
...
-    Return True if value is an instance of a union type.
-
-    Parameters
-    ----------
-    t : DataType
-    """
     return t.id in _UNION_TYPES
 
 
+@doc(is_null, datatype="nested")
```
danepitkin, line 164: Do we like `nested` or `nested type`?

python/pyarrow/types.py
```diff
@@ -407,155 +223,71 @@ def is_unicode(t):
...
-    Return True if value is an instance of a date64 (milliseconds) type.
-
-    Parameters
-    ----------
-    t : DataType
-    """
     return t.id == lib.Type_DATE64
 
 
+@doc(is_null, datatype="logical map")
```
danepitkin, line 261: Do we want to keep "logical" in `logical map`?

python/pyarrow/types.py
```diff
@@ -407,155 +223,71 @@ def is_unicode(t):
...
-    Return True if value is an instance of a decimal type.
-
-    Parameters
-    ----------
-    t : DataType
-    """
     return t.id in _DECIMAL_TYPES
 
 
+@doc(is_null, datatype="decimal128")
```
danepitkin, line 271: Added "128" to `decimal128` here.

python/pyarrow/types.py
```diff
@@ -407,155 +223,71 @@ def is_unicode(t):
...
-    Return True if value is an instance of a decimal type.
-
-    Parameters
-    ----------
-    t : DataType
-    """
     return t.id == lib.Type_DECIMAL128
 
 
+@doc(is_null, datatype="decimal256")
```
danepitkin, line 276: Added "256" to `decimal256` here.

python/pyarrow/types.py
```diff
@@ -407,155 +223,71 @@ def is_unicode(t):
...
-    Return True if the value is an instance of an interval type.
-
-    Parameters
-    ----------
-    t : DateType
-    """
     return t.id == lib.Type_INTERVAL_MONTH_DAY_NANO
 
 
+@doc(is_null, datatype="primitive")
```
danepitkin, line 291: Do we like `primitive` or `primitive type` better?

danepitkin: I chose `primitive type` since it's describing a family of types.

python/pyarrow/types.py
danepitkin: I chose `nested type` since it's describing a where the type comes from.

python/pyarrow/types.py
danepitkin: The underlying implementation is `@GARROW_TYPE_MAP: A repeated struct logical type.`

danepitkin:
I applied `@doc` to the types.py!

danepitkin:
One interesting limitation is if there are docstrings that use curly braces `{}`. If it's a doctest, we would need to replace `{}` with an equivalent like `dict()` or `set()`. If it's part of the descriptions, it can be escaped e.g. `{{}}`

example:
```
        Examples
        --------
        >>> import pyarrow as pa
        >>> import pandas as pd
        >>> df = pd.DataFrame({'year': [2020, 2022, 2019, 2021],
        ...                    'n_legs': [2, 4, 5, 100],
        ...                    'animals': ["Flamingo", "Horse", "Brittle stars", "Centipede"]})
```
vs.
```
        Examples
        --------
        >>> import pyarrow as pa
        >>> import pandas as pd
        >>> df = pd.DataFrame(dict('year': [2020, 2022, 2019, 2021],
        ...                    'n_legs': [2, 4, 5, 100],
        ...                    'animals': ["Flamingo", "Horse", "Brittle stars", "Centipede"]))
```

AlenkaF:
OK. Well, I do not have much against that (using `dict` or `set` instead of `{}`).

python/pyarrow/types.py
```diff
@@ -54,351 +56,165 @@ def is_null(t):
...
-    Return True if value is an instance of a boolean type.
-
-    Parameters
-    ----------
-    t : DataType
-    """
     return t.id == lib.Type_BOOL
 
 
+@doc(is_null, datatype="integer")
```
jorisvandenbossche, line 64:
```suggestion
@doc(is_null, datatype="any integer")
```

I would keep the "any" to clearly indicate this is for _any_ integer (i.e. all bitwidths and signed/unsigned)

python/pyarrow/types.py
jorisvandenbossche: I think we can remove that. We don't have a clear definition of "logical" types in the arrow format spec, and the fact that a map type is basically a list of structs is an implementation detail (and adding "logical" here doesn't help making that clear)

jorisvandenbossche:
> If it's a doctest, we would need to replace `{}` with an equivalent like `dict()` or `set()`. If it's part of the descriptions, it can be escaped e.g. `{{}}`

I think also in a doctest you can use the `{{ .. }}` escape approach? 

(and to be clear, the current PR doesn't run into that, right? It's only for when we would want to use it more broadly for docstrings with examples?)

danepitkin:
> > If it's a doctest, we would need to replace `{}` with an equivalent like `dict()` or `set()`. If it's part of the descriptions, it can be escaped e.g. `{{}}`
> 
> I think also in a doctest you can use the `{{ .. }}` escape approach?
> 
> (and to be clear, the current PR doesn't run into that, right? It's only for when we would want to use it more broadly for docstrings with examples?)

Correct! There are no uses of `{}` in the docstrings of this PR. You can also use `{{ .. }}` for descriptions, but my guess is that doctests would fail if you tried to use it in the code examples. I noticed Pandas usage always uses `dict()` instead of `{{ .. }}` in code examples explicitly, but does use `{{ .. }}` in the descriptions.

python/pyarrow/types.py
danepitkin: Good call! Will update

danepitkin:
All feedback is applied!

AlenkaF (approved):

PR #34896, opened 2023-04-05
GH-34893: [C++] Fix run-end encoded array iterator issues that manifest on backwards iteration

### LINKED ISSUE #34893: [C++] Add operator--() to RunEndEncodedArraySpan::Iterator and make it stop violating loop invariants

The value of `RunEndEncodedArraySpan::Iterator::physical_pos_` from the instance returned by `RunEndEncodedArraySpan::end()` is smaller (by 1) than the results of successive applications of `operator++` until `logical_pos_` reaches the logical length of the run-end encoded array.

*Why this is important?*

To support backwards iteration, the `Iterator` returned by `RunEndEncodedArraySpan::end()` should be well-formed.

*Why this is not a bug?*

There is no `operator--()` in `RunEndEncodedArraySpan::Iterator` yet, so you can't go backwards from `end()`. A fix for this issue adds that operator and uses it in tests that guarantee iterators are always valid and don't assume iteration only happens in one direction.

### Component(s)

C++

felipecrv:
### Rationale for this change

The value of `RunEndEncodedArraySpan::Iterator::physical_pos_` from the instance returned by `RunEndEncodedArraySpan::end()` is smaller (by 1) than the results of successive applications of `operator++()` until `logical_pos_` reaches the logical length of the run-end encoded array.

To support backwards iteration, the `Iterator` returned by `RunEndEncodedArraySpan::end()` should be well-formed.

There was no `operator--()` in `RunEndEncodedArraySpan::Iterator` before this PR, so one can't go backwards from `end()` and this latent bug doesn't manifest. 

 - Addition of `RunEndEncodedArraySpan::Iterator::operator--()`
 - Addition of `MergedRunsIterator::operator--()`
 - Rename `isEnd()` to `is_end()` (it was a code-style violation)
 - More test cases

### Are these changes tested?

By existing and new unit tests.

A public function -- `isEnd` -- was renamed, but this code hasn't been released yet, so it's OK.
* Closes: #34893

### DIFF
--- a/cpp/src/arrow/util/ree_util.h
+++ b/cpp/src/arrow/util/ree_util.h
@@ -154,13 +154,13 @@ class RunEndEncodedArraySpan {
     /// The values array can be addressed with this index to get the value
     /// that makes up the run.
     ///
-    /// NOTE: if this Iterator was produced by RunEndEncodedArraySpan::end(),
+    /// NOTE: if this Iterator is equal to RunEndEncodedArraySpan::end(),
     /// the value returned is undefined.
     int64_t index_into_array() const { return physical_pos_; }
 
     /// \brief Return the initial logical position of the run
     ///
-    /// If this Iterator was produced by RunEndEncodedArraySpan::end(), this is
+    /// If this Iterator is equal to RunEndEncodedArraySpan::end(), this is
     /// the same as RunEndEncodedArraySpan::length().
     int64_t logical_position() const { return logical_pos_; }
 
@@ -174,6 +174,16 @@ class RunEndEncodedArraySpan {
     /// Pre-condition: *this != RunEndEncodedArraySpan::end()
     int64_t run_length() const { return run_end() - logical_pos_; }
 
+    /// \brief Check if the iterator is at the end of the array.
+    ///
+    /// This can be used to avoid paying the cost of a call to
+    /// RunEndEncodedArraySpan::end().
+    ///
+    /// \return true if the iterator is at the end of the array
+    bool is_end(const RunEndEncodedArraySpan& span) const {
+      return logical_pos_ >= span.length();
+    }
+
     Iterator& operator++() {
       logical_pos_ = span.run_end(physical_pos_);
       physical_pos_ += 1;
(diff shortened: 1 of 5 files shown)


felipecrv:
@zeroshade

felipecrv:
@zeroshade this is ready for review and would be nice to get merged before release as I rename `isEnd` to `is_end` in a public interface.

cpp/src/arrow/util/ree_util.h
```diff
@@ -231,31 +253,52 @@ class RunEndEncodedArraySpan {
...
+  /// }
+  /// \endcode
+  ///
+  /// \par Or this version that does not look like idiomatic C++, but removes
+  /// the need for calling end() completely:
+  /// \code
+  /// for (auto it = array.begin(); !it.is_end(array); ++it) {
+  ///   // ...
+  /// }
+  /// \endcode
```
zeroshade, line 290: I'm honestly not a fan of this, but can't think of a better way at the moment yet that wouldn't be just offloading the cost somewhere else.  

Could we potentially use some sentinel value for `physical_pos_` (like <0?) that will have the iterator compute it if `++`/`--` are called on the iterator? That would allow it to remain cheap to call `end()`. Or would that just make things worse?

hmm

zeroshade (approved):
this LGTM, just the one nitpick i mentioned above but I don't really have a better solution for it than what you have already.

cpp/src/arrow/util/ree_util.h
felipecrv: Things got tricky because I want to be able to land on a valid value after applying `--` to the Iterator returned by `array.end()`. Introducing a sentinel value would add a check for every `--` and give the compiler a hard time optimizing loops. What I had before this PR kinda worked (taking the physical length), but didn't survive the introduction of `operator--`.


`for (auto it = array.begin(), end = array.end(); it != end; ++it)` is what performance-conscious C++ developers already use to write loops and the only extra cost incurred is a single O(log n) search. So I don't think adding more fields to the iterator and complicating the loop conditions is a good idea.

`for (auto it = array.begin(); !it.is_end(array); ++it)` is weird, but makes the perfectionists (me) very happy about eliminating that extra O(log n) search completely.

PR #34725, opened 2023-03-24
GH-15285: [GLib] Add GArrowMatchSubstringOptions

kou:
### Rationale for this change

It's needed for match substring related kernels.

### What changes are included in this PR?

These changes add `GArrowMatchSubstringOptions`.

These changes also includes bug fixes of `GArrowIndexOptions`. Sorry.

### Are these changes tested?

Yes.

### Are there any user-facing changes?

Yes.
* Closes: #15285

PR #6732, opened 2020-03-26
ARROW-8216: [C++][Compute] Filter out nulls by default

bkietz:
In C++ and R, Filter's default is now to drop from output when a selection slot is null. In Python, `Array.filter` now has a keyword argument `drop_nulls=True`. In R, since the expected behavior is the opposite, the argument is `keep_na = TRUE`. Old behavior is preserved in c_glib; I'm not familiar enough with that binding to add an optional parameter in an idiomatic fashion (@mrkn ?)

PR #34902, opened 2023-04-05
GH-34819: [Ruby] Add Slicer::ColumnCondition#match_substring

### LINKED ISSUE #34819: [Ruby] Add easy to use API for match_substring and related functions

It would be useful to have access to the string matching functions using the Ruby gem.

These are the functions:
https://arrow.apache.org/docs/cpp/compute.html#containment-tests

For example:

```ruby
# Find rows that have a status 200 and contain the string example in the message field
table.slice { |slicer| (slicer['status'] == 200) & (slicer['message'].match_substring 'example') }
```

### Component(s)

Ruby

heronshoes:
### Rationale for this change

C GLib supported `MatchSubstringOptions` in #34725. This PR will use this and will supply an easy to use API in Ruby binding.

- Add a class `Arrow::Slicer::MatchSubstringCondition`
- Add a method `#match_substring` in `Arrow::Slicer::ColumnCondition`

### Are these changes tested?

Yes.

Yes.

* Closes: #34819

### DIFF
--- a/ruby/red-arrow/lib/arrow/slicer.rb
+++ b/ruby/red-arrow/lib/arrow/slicer.rb
@@ -162,6 +162,10 @@ module Arrow
       def reject(&block)
         RejectCondition.new(@column, block)
       end
+
+      def match_substring(value, ignore_case: false)
+        MatchSubstringCondition.new(@column, value, ignore_case)
+      end
     end
 
     class NotColumnCondition < Condition
@@ -349,6 +353,19 @@ module Arrow
           end
         end
         BooleanArray.new(raw_array)
+      end
+    end
+
+    class MatchSubstringCondition < Condition
+      def initialize(column, value, ignore_case)
+        @column = column
+        @options = MatchSubstringOptions.new
+        @options.pattern = value
+        @options.ignore_case = true if ignore_case
+      end
+
+      def evaluate
+        Function.find("match_substring").execute([@column.data], @options).value
       end
     end
   end
--- a/ruby/red-arrow/lib/arrow/slicer.rb
+++ b/ruby/red-arrow/lib/arrow/slicer.rb
@@ -378,10 +378,9 @@ module Arrow
       def evaluate
         datum = Function.find(@function).execute([@column.data], @options)
         if @invert
-          Function.find("invert").execute([datum]).value
-        else
-          datum.value
+          datum = Function.find("invert").execute([datum])
         end
+        datum.value
       end
     end
   end
--- a/ruby/red-arrow/lib/arrow/slicer.rb
+++ b/ruby/red-arrow/lib/arrow/slicer.rb
@@ -164,7 +164,7 @@ module Arrow
       end
 
       def match_substring(substring, ignore_case: false)
-        MatchSubstringCondition.new("match_substring",
+        MatchSubstringFamilyCondition.new("match_substring",
                                     @column, substring, ignore_case)
       end
     end
@@ -357,7 +357,7 @@ module Arrow
       end
     end
 
-    class MatchSubstringCondition < Condition
+    class MatchSubstringFamilyCondition < Condition
       def initialize(function, column, pattern, ignore_case, invert: false)
         @function = function
         @column = column
@@ -368,7 +368,7 @@ module Arrow
       end
 
       def !@
-        MatchSubstringCondition.new(@function,
+        MatchSubstringFamilyCondition.new(@function,
                                     @column,
                                     @options.pattern,
                                     @options.ignore_case?,
--- a/ruby/red-arrow/lib/arrow/slicer.rb
+++ b/ruby/red-arrow/lib/arrow/slicer.rb
@@ -361,7 +361,7 @@ module Arrow
         @column = column
         @options = MatchSubstringOptions.new
         @options.pattern = substring
-        @options.ignore_case = true if ignore_case
+        @options.ignore_case = ignore_case
       end
 
       def evaluate
--- a/ruby/red-arrow/lib/arrow/slicer.rb
+++ b/ruby/red-arrow/lib/arrow/slicer.rb
@@ -165,7 +165,7 @@ module Arrow
 
       def match_substring(substring, ignore_case: false)
         MatchSubstringFamilyCondition.new("match_substring",
-                                    @column, substring, ignore_case)
+                                          @column, substring, ignore_case)
       end
     end
 
@@ -369,10 +369,10 @@ module Arrow
 
       def !@
         MatchSubstringFamilyCondition.new(@function,
-                                    @column,
-                                    @options.pattern,
-                                    @options.ignore_case?,
-                                    invert: !@invert)
+                                          @column,
+                                          @options.pattern,
+                                          @options.ignore_case?,
+                                          invert: !@invert)
       end
 
       def evaluate
--- a/ruby/red-arrow/lib/arrow/slicer.rb
+++ b/ruby/red-arrow/lib/arrow/slicer.rb
@@ -164,7 +164,8 @@ module Arrow
       end
 
       def match_substring(substring, ignore_case: false)
-        MatchSubstringCondition.new(@column, substring, ignore_case)
+        MatchSubstringFamilyCondition.new("match_substring",
+                                          @column, substring, ignore_case)
       end
     end
 
@@ -356,8 +357,9 @@ module Arrow
       end
     end
 
-    class MatchSubstringCondition < Condition
-      def initialize(column, substring, ignore_case)
+    class MatchSubstringFamilyCondition < Condition
+      def initialize(function, column, substring, ignore_case)
+        @function = function
         @column = column
         @options = MatchSubstringOptions.new
         @options.pattern = substring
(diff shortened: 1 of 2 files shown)


ruby/red-arrow/lib/arrow/slicer.rb
```diff
@@ -162,6 +162,10 @@ def select(&block)
       def reject(&block)
         RejectCondition.new(@column, block)
       end
+
+      def match_substring(substring, ignore_case: false)
+        MatchSubstringCondition.new(@column, substring, ignore_case)
```
kou, line 167: How about making this reusable for other `match_substring` family functions?

```suggestion
        MatchSubstringFamilyCondition.new("match_substring", @column, substring, ignore_case)
```

ruby/red-arrow/lib/arrow/slicer.rb
```diff
@@ -351,5 +355,18 @@ def evaluate
         BooleanArray.new(raw_array)
       end
     end
+
+    class MatchSubstringCondition < Condition
+      def initialize(column, substring, ignore_case)
+        @column = column
+        @options = MatchSubstringOptions.new
+        @options.pattern = substring
+        @options.ignore_case = true if ignore_case
```
kou, line 364:
```suggestion
        @options.ignore_case = ignore_case
```

ruby/red-arrow/test/test-slicer.rb
```diff
@@ -484,4 +488,29 @@ def setup
 7	   256	true   
     TABLE
   end
+
+  test("match_substring") do
```
kou, line 492: Could you create sub test case and move `@string_table = ...` to here?

```ruby
sub_test_case("#match_substring") do
  def setup
    super # Not needed?
    # index: is needed?
    @table = Arrow::Table.new(index: [*1..5], string: ["array", "Arrow", "carrot", nil, "window"])
  end
  
  test("String") do
  end

  test("String, ignore_case:") do
  end
end
```

heronshoes:
@kou Thanks! I tried to fix them.

Is it necessary to support '!match_substring' ?
(Such as `!slicer.string.match_substring("abc")`).

ruby/red-arrow/test/test-slicer.rb
```diff
@@ -484,4 +484,38 @@ def setup
...
+      end
+      assert_equal(<<~TABLE, sliced_table.to_s)
+	 index	string
+0	     1	array 
+1	     3	carrot
+2	(null)	(null)
+      TABLE
+    end
+
+    test("match_substring with ignore_case option") do
```
kou, line 508: Could you use simple Ruby syntax like test name instead of English?
I like the style because it's easy to read for me :-)

```suggestion
    test("match_substring(ignore_case:)") do
```

ruby/red-arrow/test/test-slicer.rb
```diff
@@ -484,4 +484,38 @@ def setup
 7	   256	true   
     TABLE
   end
+
+  sub_test_case "MatchSubstringOptions family" do
+    def setup
+      @string_table = Arrow::Table.new(
```
kou, line 490: I think that `string_` is redundant. Because we process only string in this context.

ruby/red-arrow/test/test-slicer.rb
```diff
@@ -484,4 +484,38 @@ def setup
...
+        index: [*1..5],
```
kou, line 491: Why do we need this?

ruby/red-arrow/lib/arrow/slicer.rb
```diff
@@ -351,5 +356,22 @@ def evaluate
...
+        @options = MatchSubstringOptions.new
+        @options.pattern = substring
+        @options.ignore_case = ignore_case
+      end
+
+      def evaluate
+        case @function
+        when "match_substring"
+          Function.find("match_substring").execute([@column.data], @options).value
+        end
```
kou, line 373:
```suggestion
        Function.find(@function).execute([@column.data], @options).value
```

kou (commented):
> Is it necessary to support `'!match_substring'`?

Ah, yes, please!

ruby/red-arrow/test/test-slicer.rb
heronshoes: Sure, thanks.

ruby/red-arrow/test/test-slicer.rb
heronshoes: Yes it's redundant. I feel something strange for the behavior of propagating nulls to other columns, so I put it here to show as example for users. But I will remove `:index` column here.

May be off topic of this PR, is it useful to propagate nulls in slicer?

I prefer to same behavior of `table#filter(boolean_array)` or `slicer.select { |value| .. }`.
(Null is not selected.)

ruby/red-arrow/lib/arrow/slicer.rb
heronshoes: Thanks, I will do so. (May some change needed to implement `#!@` or other family methods.)

ruby/red-arrow/test/test-slicer.rb
kou: It's just for keeping backward compatibility.
See also: #6732

But we may can change it for 12.0.0. (2023-04-11 is code freeze date: https://lists.apache.org/thread/wj2x50s2xyghnjm0pm4gt7yq5187zhwy )
Could you open an issue and a pull request for it?

ruby/red-arrow/lib/arrow/slicer.rb
```diff
@@ -162,6 +162,11 @@ def select(&block)
       def reject(&block)
         RejectCondition.new(@column, block)
       end
+
+      def match_substring(substring, ignore_case: false)
+        MatchSubstringCondition.new("match_substring",
```
kou, line 167: How about `MatchSubstringFamilyCondition` because this is not only for `match_substring`?

ruby/red-arrow/lib/arrow/slicer.rb
```diff
@@ -351,5 +356,33 @@ def evaluate
...
+                                    invert: !@invert)
+      end
+
+      def evaluate
+        datum = Function.find(@function).execute([@column.data], @options)
+        if @invert
+          Function.find("invert").execute([datum]).value
+        else
+          datum.value
+        end
```
kou, line 384: It's not a strong opinion but the following may be simpler:

```suggestion
        if @invert
          datum = Function.find("invert").execute([datum])
        end
        datum.value
```

ruby/red-arrow/test/test-slicer.rb
kou: FYI: The following change will work for it:

```diff
diff --git a/ruby/red-arrow/lib/arrow/table.rb b/ruby/red-arrow/lib/arrow/table.rb
index a14bf90d1f..da79e94b18 100644
--- a/ruby/red-arrow/lib/arrow/table.rb
+++ b/ruby/red-arrow/lib/arrow/table.rb
@@ -316,8 +316,6 @@ module Arrow
         end
       end
 
-      filter_options = Arrow::FilterOptions.new
-      filter_options.null_selection_behavior = :emit_null
       sliced_tables = []
       slicers.each do |slicer|
         slicer = slicer.evaluate if slicer.respond_to?(:evaluate)
@@ -339,7 +337,7 @@ module Arrow
           to += n_rows if to < 0
           sliced_tables << slice_by_range(from, to)
         when ::Array, BooleanArray, ChunkedArray
-          sliced_tables << filter(slicer, filter_options)
+          sliced_tables << filter(slicer)
         else
           message = "slicer must be Integer, Range, (from, to), " +
             "Arrow::ChunkedArray of Arrow::BooleanArray, " +
```

heronshoes: I see. I will open an issue and PR for it. Thanks.

ruby/red-arrow/lib/arrow/slicer.rb
heronshoes: Thnaks!

ruby/red-arrow/lib/arrow/slicer.rb
```diff
@@ -162,6 +162,11 @@ def select(&block)
       def reject(&block)
         RejectCondition.new(@column, block)
       end
+
+      def match_substring(substring, ignore_case: false)
+        MatchSubstringFamilyCondition.new("match_substring",
+                                    @column, substring, ignore_case)
```
kou, line 168:
```suggestion
                                          @column, substring, ignore_case)
```

ruby/red-arrow/lib/arrow/slicer.rb
```diff
@@ -351,5 +356,32 @@ def evaluate
...
+        @options.ignore_case = ignore_case
+        @invert = invert
+      end
+
+      def !@
+        MatchSubstringFamilyCondition.new(@function,
+                                    @column,
+                                    @options.pattern,
+                                    @options.ignore_case?,
+                                    invert: !@invert)
```
kou, line 375:
```suggestion
                                          @column,
                                          @options.pattern,
                                          @options.ignore_case?,
                                          invert: !@invert)
```

kou (approved):
+1

heronshoes:
Thanks for your support!
I will open issue/PR of its family.

PR #34903, opened 2023-04-05
MINOR: [Docs][Python] Remove note about pyarrow.dataset being experimental

jorisvandenbossche:
### Rationale for this change

In practice pyarrow.dataset is quite stable and we have not been treating it as experimental. So after several years of existence, high time to remove the (misleading) "experimental" label


danepitkin:
Better late than never! LGTM 👍

PR #34904, opened 2023-04-05
GH-34786: [C++] Fix output schema calculated by Substrait consumer for AggregateRel

### LINKED ISSUE #34786: [C++] Output schema calculated by Substrait consumer for aggregate rel seems incorrect.

rtpsw:
See #34786
* Closes: #34786

Can replace #34885


rtpsw:
@westonpace, this is a simpler PR, as compared to #34885, following a discussion with @icexelloss.

cpp/src/arrow/acero/aggregate_node.cc
```diff
@@ -77,6 +78,19 @@ namespace acero {
 
 namespace {
 
+template <typename KernelType>
+struct ARROW_ACERO_EXPORT AggregateNodeArgs {
```
icexelloss, line 82: I am actually not sure what does `ARROW_ACERO_EXPORT` does but it seems this is a private struct of the cc file and doesn't need to be exported?

cpp/src/arrow/acero/aggregate_node.cc
```diff
@@ -602,7 +621,7 @@ class GroupByNode : public ExecNode, public TracedNode {
     for (const auto& segment_key_field_id : segment_key_field_ids) {
       if (key_field_id_set.find(segment_key_field_id) != key_field_id_set.end()) {
         return Status::Invalid("Group-by aggregation with field '",
-                               input_schema->field(segment_key_field_id)->name(),
```
icexelloss, line 605: Any reason that we need to change `input_schema` from a pointer reference to a schema reference?

cpp/src/arrow/engine/substrait/relation_internal.h
```diff
@@ -50,11 +50,6 @@ ARROW_ENGINE_EXPORT Result<std::unique_ptr<substrait::Rel>> ToProto(
 
 namespace internal {
 
-struct ParsedMeasure {
```
icexelloss, line 53: Nice cleanup.

icexelloss:
@rtpsw Thanks - at high level I think this looks a lot nicer - 

Are there any test we can add for this?

cpp/src/arrow/acero/aggregate_node.cc
rtpsw: Yes, this was moved here from the header but has no effect here - should be removed.

rtpsw:
> Are there any test we can add for this?

I should have an answer after experimenting with this code a bit.

cpp/src/arrow/acero/aggregate_node.cc
rtpsw: I think this can be a `shared_ptr`, if you prefer.

cpp/src/arrow/acero/aggregate_node.h
```diff
@@ -0,0 +1,53 @@
...
+///
+/// \param[in] input_schema the schema of the input to the node
+/// \param[in] keys the grouping keys for the aggregation
+/// \param[in] segment_keys the segmenting keys for the aggregation
+/// \param[in] aggregates the aggregates for the aggregation
+/// \param[in] exec_ctx the execution context for the aggregation
+ARROW_ACERO_EXPORT Result<std::shared_ptr<Schema>> MakeOutputSchema(
+    const Schema& input_schema, const std::vector<FieldRef>& keys,
+    const std::vector<FieldRef>& segment_keys, const std::vector<Aggregate>& aggregates,
+    ExecContext* exec_ctx = default_exec_context());
```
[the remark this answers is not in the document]
icexelloss: Got it - not a big deal. But maybe add in the docstring why this is needed. (basically to get the output schema, we need to create kernels, and to create kernels we need exec context?)

cpp/src/arrow/acero/aggregate_node.cc
icexelloss: Yeah I think `const shared_ptr<Schema>&` is better because then some of the lines using input_schema can be unchanged.

cpp/src/arrow/acero/aggregate_node.cc
rtpsw: Fixed.

cpp/src/arrow/acero/aggregate_node.cc
rtpsw: Fixed.

cpp/src/arrow/acero/aggregate_node.h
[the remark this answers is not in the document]
rtpsw: Done.

rtpsw:
> > Are there any test we can add for this?
> 
> I should have an answer after experimenting with this code a bit.

Added test-cases.

cpp/src/arrow/acero/aggregate_node.cc
```diff
@@ -665,14 +682,47 @@ class GroupByNode : public ExecNode, public TracedNode {
...
 
+    return AggregateNodeArgs<HashAggregateKernel>{schema(std::move(output_fields)),
+                                                  std::move(key_field_ids),
+                                                  std::move(segment_key_field_ids),
+                                                  std::move(segmenter),
+                                                  std::move(agg_src_fieldsets),
+                                                  std::move(aggs),
+                                                  std::move(agg_kernels),
+                                                  std::move(agg_src_types),
+                                                  {}};
```
icexelloss, line 696: ditto

cpp/src/arrow/acero/aggregate_node.cc
```diff
@@ -665,14 +682,47 @@ class GroupByNode : public ExecNode, public TracedNode {
...
+
+  static Result<ExecNode*> Make(ExecPlan* plan, std::vector<ExecNode*> inputs,
+                                const ExecNodeOptions& options) {
+    RETURN_NOT_OK(ValidateExecNodeInputs(plan, inputs, 1, "GroupByNode"));
+
+    auto input = inputs[0];
+    const auto& aggregate_options = checked_cast<const AggregateNodeOptions&>(options);
+    const auto& keys = aggregate_options.keys;
+    const auto& segment_keys = aggregate_options.segment_keys;
+    // Copy (need to modify options pointer below)
```
icexelloss, line 707: Where did we modify the options?

cpp/src/arrow/acero/aggregate_node.cc
```diff
@@ -665,14 +682,47 @@ class GroupByNode : public ExecNode, public TracedNode {
...
+    auto aggs = aggregate_options.aggregates;
+
+    if (plan->query_context()->exec_context()->executor()->GetCapacity() > 1 &&
+        segment_keys.size() > 0) {
+      return Status::NotImplemented("Segmented aggregation in a multi-threaded plan");
```
icexelloss, line 712: nit: "multi-threaded plan" -> "multi-threaded execution context"

icexelloss:
@rtpsw Looks pretty good - left a few question/comments.

icexelloss:
@westonpace This looks pretty good to me. Can you please take a look too?

cpp/src/arrow/acero/aggregate_node.cc
```diff
@@ -665,14 +682,47 @@ class GroupByNode : public ExecNode, public TracedNode {
...
+    }
+
+    auto input_schema = input->output_schema();
```
westonpace, line 715:
```suggestion
    const auto& input_schema = input->output_schema();
```

westonpace (approved):
This is a great simplification.  Thanks!

cpp/src/arrow/acero/aggregate_node.cc
rtpsw: In the refactoring, I forgot to move this comment to line 294, since `aggregates` there is modified (in line 358). I'll do this.

cpp/src/arrow/acero/aggregate_node.cc
```diff
@@ -665,14 +680,47 @@ class GroupByNode : public ExecNode, public TracedNode {
...
+
+    return AggregateNodeArgs<HashAggregateKernel>{schema(std::move(output_fields)),
+                                                  std::move(key_field_ids),
+                                                  std::move(segment_key_field_ids),
+                                                  std::move(segmenter),
+                                                  std::move(agg_src_fieldsets),
+                                                  std::move(aggs),
+                                                  std::move(agg_kernels),
+                                                  std::move(agg_src_types),
+                                                  /*states=*/{}};
```
icexelloss, line 694: Why states is passed in with ScalarAggregate but not HashAggregate?

rtpsw: Technically, this is because `GroupByNode` does not take states in [its constructor](https://github.com/apache/arrow/blob/d9e672f40e831dd8f6a73a21b3a9bebfc35c5a45/cpp/src/arrow/acero/aggregate_node.cc#L543-L549) whereas `ScalarAggregateNode` [does](https://github.com/apache/arrow/blob/d9e672f40e831dd8f6a73a21b3a9bebfc35c5a45/cpp/src/arrow/acero/aggregate_node.cc#L257-L265). I don't think there is a real design justification for this difference; I just didn't change the original design here.

Issue #34911, opened 2023-04-05
[C++] Add first and last aggregation


icexelloss:
### Describe the enhancement requested

Add first and last aggregation kernel and support in Acero

### Component(s)

C++

icexelloss:
take

westonpace:
We spoke about this a bit externally but to record the conversation:

`first` and `last` can be written in two different ways.  The unary version `first(expr)` is a window aggregate.  It is an aggregate function that depends on the order of the data.  There is also a binary variant (often called `arg_min` and `arg_max`) which returns the smallest value in column X given column Y.  The advantage of the binary variant is that it doesn't depend on the order.  However, it is a little less flexible (e.g. no way to specify custom sort function, not that we support that yet anyways :)).

From our conversation, my understanding is that you are interested in the unary window-aggregate version.  We don't have any window aggregates yet but I do think it would be a good idea to start adding some.  We have pretty much all the building blocks we need to support a "window aggregate node" (or an extension to the current aggregate nodes).  Furthermore, even if we don't build in proper multithreaded support for window functions it should be possible to use window aggregates today as long as your plan is single threaded.

westonpace:
Issue resolved by pull request 34912
https://github.com/apache/arrow/pull/34912

PR #34912, opened 2023-04-05
GH-34911: [C++] Add first and last aggregator

icexelloss:
### Rationale for this change
This PR adds "first" and "last" aggregator and support using those with Acero's segmented aggregation.

- [x] Numeric Scalar Aggregator (bool, int types, floating types)
- [x] Numeric Hash Aggregator (bool, int types, floating types)
- [x] Docstring
- [x] Non-Numeric Scalar Aggregator (string, binary, fixed binary, temporal)
- [x] Non-Numeric Hash Aggregator (string, binary, fixed binary, temporal)
- [x] Add `ordered` flag in aggregate kernels
- [x] Implement and test skip null
- [x] Update compute.rst

### Are these changes tested?
- [x] Compute Kernel Test (Scalar Kernels, all supported datatypes)
- [x] Hash Aggregate Test (Hash Kernels, all supported datatypes)
- [x] Segmented Aggregation Test (Both Scalar and Hash Kernels)

Yes. Added First and Last aggregator.

### DIFF
--- a/cpp/src/arrow/acero/aggregate_node.cc
+++ b/cpp/src/arrow/acero/aggregate_node.cc
@@ -616,7 +616,7 @@ class GroupByNode : public ExecNode, public TracedNode {
   static Result<AggregateNodeArgs<HashAggregateKernel>> MakeAggregateNodeArgs(
       const std::shared_ptr<Schema>& input_schema, const std::vector<FieldRef>& keys,
       const std::vector<FieldRef>& segment_keys, const std::vector<Aggregate>& aggs,
-      ExecContext* ctx) {
+      ExecContext* ctx, const int concurrency) {
     // Find input field indices for key fields
     std::vector<int> key_field_ids(keys.size());
     for (size_t i = 0; i < keys.size(); ++i) {
@@ -673,6 +673,20 @@ class GroupByNode : public ExecNode, public TracedNode {
     // Construct aggregates
     ARROW_ASSIGN_OR_RAISE(auto agg_kernels, GetKernels(ctx, aggs, agg_src_types));
 
+    if (concurrency > 1) {
+      if (segment_keys.size() > 0) {
+        return Status::NotImplemented(
+            "Segmented aggregation in a multi-threaded execution context");
+      }
+
+      for (auto kernel : agg_kernels) {
+        if (kernel->ordered) {
+          return Status::NotImplemented(
+              "Using ordered aggregator in multiple threaded execution is not supported");
+        }
+      }
+    }
+
     ARROW_ASSIGN_OR_RAISE(auto agg_states,
                           InitKernels(agg_kernels, ctx, aggs, agg_src_types));
 
@@ -720,27 +734,13 @@ class GroupByNode : public ExecNode, public TracedNode {
     const auto& keys = aggregate_options.keys;
     const auto& segment_keys = aggregate_options.segment_keys;
     auto aggs = aggregate_options.aggregates;
-    auto concurreny = plan->query_context()->exec_context()->executor()->GetCapacity();
-
-    if (concurreny > 1 && segment_keys.size() > 0) {
-      return Status::NotImplemented(
-          "Segmented aggregation in a multi-threaded execution context");
-    }
+    auto concurrency = plan->query_context()->exec_context()->executor()->GetCapacity();
 
     const auto& input_schema = input->output_schema();
     auto exec_ctx = plan->query_context()->exec_context();
-
-    ARROW_ASSIGN_OR_RAISE(auto args, MakeAggregateNodeArgs(input_schema, keys,
-                                                           segment_keys, aggs, exec_ctx));
-
-    if (concurreny > 1) {
-      for (auto kernel : args.kernels) {
-        if (kernel->ordered) {
-          return Status::NotImplemented(
-              "Using ordered aggregator in multiple threaded execution is not supported");
-        }
-      }
-    }
+    ARROW_ASSIGN_OR_RAISE(
+        auto args, MakeAggregateNodeArgs(input_schema, keys, segment_keys, aggs, exec_ctx,
+                                         concurrency));
 
     return input->plan()->EmplaceNode<GroupByNode>(
         input, std::move(args.output_schema), std::move(args.grouping_key_field_ids),
@@ -1068,9 +1068,9 @@ Result<std::shared_ptr<Schema>> MakeOutputSchema(
                                          exec_ctx, /*concurrency=*/1));
     return std::move(args.output_schema);
   } else {
-    ARROW_ASSIGN_OR_RAISE(
-        auto args, GroupByNode::MakeAggregateNodeArgs(input_schema, keys, segment_keys,
-                                                      aggregates, exec_ctx));
+    ARROW_ASSIGN_OR_RAISE(auto args, GroupByNode::MakeAggregateNodeArgs(
+                                         input_schema, keys, segment_keys, aggregates,
+                                         exec_ctx, /*concurrency=*/1));
     return std::move(args.output_schema);
   }
 }
--- a/cpp/src/arrow/acero/aggregate_node.cc
+++ b/cpp/src/arrow/acero/aggregate_node.cc
@@ -294,6 +294,7 @@ class ScalarAggregateNode : public ExecNode, public TracedNode {
       ExecContext* exec_ctx, size_t concurrency) {
     // Copy (need to modify options pointer below)
     std::vector<Aggregate> aggregates(aggs);
+
     std::vector<int> segment_field_ids(segment_keys.size());
     std::vector<TypeHolder> segment_key_types(segment_keys.size());
     for (size_t i = 0; i < segment_keys.size(); i++) {
@@ -350,7 +351,14 @@ class ScalarAggregateNode : public ExecNode, public TracedNode {
       kernel_intypes[i] = in_types;
       ARROW_ASSIGN_OR_RAISE(const Kernel* kernel,
                             function->DispatchExact(kernel_intypes[i]));
-      kernels[i] = static_cast<const ScalarAggregateKernel*>(kernel);
+      const ScalarAggregateKernel* agg_kernel =
+          static_cast<const ScalarAggregateKernel*>(kernel);
+      if (concurrency > 1 && agg_kernel->ordered) {
+        return Status::NotImplemented(
+            "Using ordered aggregator in multiple threaded execution is not supported");
+      }
+
+      kernels[i] = agg_kernel;
 
       if (aggregates[i].options == nullptr) {
         DCHECK(!function->doc().options_required);
@@ -391,12 +399,13 @@ class ScalarAggregateNode : public ExecNode, public TracedNode {
     auto aggregates = aggregate_options.aggregates;
     const auto& keys = aggregate_options.keys;
     const auto& segment_keys = aggregate_options.segment_keys;
+    const auto concurreny =
+        plan->query_context()->exec_context()->executor()->GetCapacity();
 
     if (keys.size() > 0) {
       return Status::Invalid("Scalar aggregation with some key");
     }
-    if (plan->query_context()->exec_context()->executor()->GetCapacity() > 1 &&
-        segment_keys.size() > 0) {
+    if (concurreny > 1 && segment_keys.size() > 0) {
       return Status::NotImplemented("Segmented aggregation in a multi-threaded plan");
     }
 
@@ -406,7 +415,16 @@ class ScalarAggregateNode : public ExecNode, public TracedNode {
     ARROW_ASSIGN_OR_RAISE(
         auto args,
         MakeAggregateNodeArgs(input_schema, keys, segment_keys, aggregates, exec_ctx,
-                              /*concurrency=*/plan->query_context()->max_concurrency()));
+                              /*concurrency=*/concurreny));
+
+    if (concurreny > 1) {
+      for (auto& kernel : args.kernels) {
+        if (kernel->ordered) {
+          return Status::NotImplemented(
+              "Using ordered aggregator in multiple threaded execution is not supported");
+        }
+      }
+    }
 
     return plan->EmplaceNode<ScalarAggregateNode>(
         plan, std::move(inputs), std::move(args.output_schema), std::move(args.segmenter),
@@ -703,9 +721,9 @@ class GroupByNode : public ExecNode, public TracedNode {
     const auto& keys = aggregate_options.keys;
     const auto& segment_keys = aggregate_options.segment_keys;
     auto aggs = aggregate_options.aggregates;
+    auto concurreny = plan->query_context()->exec_context()->executor()->GetCapacity();
 
-    if (plan->query_context()->exec_context()->executor()->GetCapacity() > 1 &&
-        segment_keys.size() > 0) {
+    if (concurreny > 1 && segment_keys.size() > 0) {
       return Status::NotImplemented(
           "Segmented aggregation in a multi-threaded execution context");
     }
@@ -715,6 +733,15 @@ class GroupByNode : public ExecNode, public TracedNode {
 
     ARROW_ASSIGN_OR_RAISE(auto args, MakeAggregateNodeArgs(input_schema, keys,
                                                            segment_keys, aggs, exec_ctx));
+
+    if (concurreny > 1) {
+      for (auto kernel : args.kernels) {
+        if (kernel->ordered) {
+          return Status::NotImplemented(
+              "Using ordered aggregator in multiple threaded execution is not supported");
+        }
+      }
+    }
 
     return input->plan()->EmplaceNode<GroupByNode>(
         input, std::move(args.output_schema), std::move(args.grouping_key_field_ids),
--- a/cpp/src/arrow/acero/aggregate_node.cc
+++ b/cpp/src/arrow/acero/aggregate_node.cc
@@ -294,7 +294,6 @@ class ScalarAggregateNode : public ExecNode, public TracedNode {
       ExecContext* exec_ctx, size_t concurrency) {
     // Copy (need to modify options pointer below)
     std::vector<Aggregate> aggregates(aggs);
-
     std::vector<int> segment_field_ids(segment_keys.size());
     std::vector<TypeHolder> segment_key_types(segment_keys.size());
     for (size_t i = 0; i < segment_keys.size(); i++) {
--- a/cpp/src/arrow/acero/hash_aggregate_test.cc
+++ b/cpp/src/arrow/acero/hash_aggregate_test.cc
@@ -4232,11 +4232,11 @@ TEST_P(GroupBy, FirstLastNumericAndTemporalTypes) {
 
   const std::string default_expected =
       R"([
-    [1,    1,    3],
-    [2,    0,    0],
-    [3,    null,  null],
-    [4,    3,     5],
-    [null, 4,     1]
+    [1,    1,    3,    null,   null],
+    [2,    0,    0,    0,   0],
+    [3,    null,  null,  null,  null],
+    [4,    3,     5,    3,   5],
+    [null, 4,     1,    4,   1]
     ])";
 
   const std::vector<std::string> date64_table = {R"([
@@ -4260,12 +4260,14 @@ TEST_P(GroupBy, FirstLastNumericAndTemporalTypes) {
 
   const std::string date64_expected =
       R"([
-    [1,    86400000,  259200000 ],
-    [2,    0,   0               ],
-    [3,    null, null           ],
-    [4,    259200000, 432000000 ],
-    [null, 345600000, 86400000]
+    [1,    86400000,259200000,null,null],
+    [2,    0,0,0,0],
+    [3,    null,null,null,null],
+    [4,    259200000,432000000,259200000,432000000],
+    [null, 345600000,86400000,345600000,86400000]
     ])";
+
+  auto skip_nulls = std::make_shared<ScalarAggregateOptions>(false, 1);
 
   for (const auto& ty : types) {
     SCOPED_TRACE(ty->ToString());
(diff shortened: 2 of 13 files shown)


icexelloss:
This is work in progress but I want to put this up because I got segmented aggregation to work with numeric types. Will finish up the rest of the box (see check boxes in the description).

icexelloss:
@westonpace I pretty much followed the way that min/max aggregator is implemented so hopefully there is no surprises here. Still, would be appreciate if you can take a look at whether this general approach is correct.

cpp/src/arrow/compute/kernels/aggregate_basic_internal.h
```diff
@@ -272,8 +273,120 @@ struct MeanKernelInit : public SumLikeInit<KernelClass> {
...
+  }
+
+  void MergeOne(T value) {
+    if (!this->first.has_value()) {
+      this->first = value;
+    }
+    this->last = value;
+  }
+
+  std::optional<T> first = std::nullopt;
```
icexelloss, line 300: WIP: Probably want to make this just T and have a bool `has_values` similar to the hash aggregator version.

cpp/src/arrow/compute/kernels/hash_aggregate.cc
```diff
@@ -1251,6 +1251,210 @@ HashAggregateKernel MakeApproximateMedianKernel(HashAggregateFunction* tdigest_f
...
+      }
+    }
+    return Status::OK();
+  }
+
+  Result<Datum> Finalize() override {
+    ARROW_ASSIGN_OR_RAISE(auto null_bitmap, has_values_.Finish());
+
+    if (!options_.skip_nulls) {
+      return Status::NotImplemented("Don't support first/last with skip nulls = False");
```
icexelloss, line 1364: WIP: Support this(?)

cpp/src/arrow/compute/kernels/hash_aggregate.cc
```diff
@@ -1251,6 +1251,210 @@ HashAggregateKernel MakeApproximateMedianKernel(HashAggregateFunction* tdigest_f
...
+  Status Visit(const DoubleType&) {
+    kernel = MakeKernel(std::move(argument_type), FirstLastInit<DoubleType>);
+    return Status::OK();
+  }
+
+  Status Visit(const HalfFloatType& type) {
+    return Status::NotImplemented("Computing first/last of data of type ", type);
+  }
+
+  Status Visit(const DataType& type) {
```
icexelloss, line 1443: WIP: Support non-numeric types

westonpace (commented):
A few general questions.

Also, how does `first` compare with something like `LIMIT 1`?

Do you think there is value in instead implementing something like `nth_value`?  E.g. https://docs.snowflake.com/en/sql-reference/functions/nth_value

icexelloss:
> Also, how does first compare with something like LIMIT 1?

I see them as different things. `limit` to me seems like "get partial results from the full results". `first` feels like a ordered required aggregation function, which can be used with group by, window aggregations. `limit` can only be used (IIRC) as the last statement in the query plan, while `first` can appear in the middle.

> Do you think there is value in instead implementing something like nth_value? E.g. https://docs.snowflake.com/en/sql-reference/functions/nth_value

I think there is value. I can look into if this is easy to do.

cpp/src/arrow/compute/kernels/aggregate_basic_internal.h
icexelloss: Done

cpp/src/arrow/compute/kernels/hash_aggregate.cc
```diff
@@ -1695,6 +1692,389 @@ struct GroupedMinMaxFactory {
...
+    auto firsts = ArrayData::Make(type_, num_groups_, {null_bitmap, nullptr});
+    auto lasts = ArrayData::Make(type_, num_groups_, {std::move(null_bitmap), nullptr});
+    RETURN_NOT_OK(MakeOffsetsValues(firsts.get(), firsts_));
+    RETURN_NOT_OK(MakeOffsetsValues(lasts.get(), lasts_));
+    return ArrayData::Make(out_type(), num_groups_, {nullptr},
+                           {std::move(firsts), std::move(lasts)});
+  }
+
+  template <typename T = Type>
+  enable_if_base_binary<T, Status> MakeOffsetsValues(
```
icexelloss, line 1915: This is copied from the binary version of the grouped min/max kernel. I will try to refactor this to a base class.

cpp/src/arrow/compute/kernels/hash_aggregate.cc
```diff
@@ -1695,6 +1692,389 @@ struct GroupedMinMaxFactory {
...
+  static constexpr float value() { return std::numeric_limits<float>::infinity(); }
+};
+
+template <>
+struct NullSentinel<double> {
+  static constexpr double value() { return std::numeric_limits<double>::infinity(); }
+};
+
+template <typename Type, typename Enable = void>
+struct GroupedFirstLastImpl final : public GroupedAggregator {
```
icexelloss, line 1714: This class follows the GroupedMinMaxImpl with slight modification of the Consume and Merge function

cpp/src/arrow/acero/hash_aggregate_test.cc
```diff
@@ -4206,6 +4206,187 @@ TEST_P(GroupBy, MinMaxWithNewGroupsInChunkedArray) {
                     /*verbose=*/true);
 }
 
+TEST_P(GroupBy, FirstLastNumericAndTemporalTypes) {
```
icexelloss, line 4209: Test for numeric and temporal types

cpp/src/arrow/acero/hash_aggregate_test.cc
```diff
@@ -4206,6 +4206,187 @@ TEST_P(GroupBy, MinMaxWithNewGroupsInChunkedArray) {
...
+      [2,    "bcd",    "babcd"],
+      [3,    null,    null],
+      [null, "2",    "123"]
+    ])"),
+                      aggregated_and_grouped,
+                      /*verbose=*/true);
+  }
+}
+
+TEST_P(GroupBy, FirstLastFixedSizeBinary) {
```
icexelloss, line 4343: Test for fixed binary types

cpp/src/arrow/compute/kernels/aggregate_basic_internal.h
```diff
@@ -272,8 +273,120 @@ struct MeanKernelInit : public SumLikeInit<KernelClass> {
...
+  using ThisType = FirstLastImpl<ArrowType, SimdLevel>;
+  using StateType = FirstLastState<ArrowType, SimdLevel>;
+
+  FirstLastImpl(std::shared_ptr<DataType> out_type, ScalarAggregateOptions options)
+      : out_type(std::move(out_type)), options(std::move(options)), count(0) {
+    this->options.min_count = std::max<uint32_t>(1, this->options.min_count);
+  }
+
+  Status Consume(KernelContext*, const ExecSpan& batch) override {
+    if (batch[0].is_array()) {
```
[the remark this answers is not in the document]
icexelloss: Actually - This Aggregator can take scalar input. Agree it is probably not too useful but doesn't hurt either. (There is some tests that hits this)

cpp/src/arrow/compute/kernels/hash_aggregate.cc
```diff
@@ -105,22 +105,19 @@ Result<TypeHolder> ResolveGroupOutputType(KernelContext* ctx,
...
-                               KernelInit init) {
-  HashAggregateKernel kernel;
-  kernel.init = std::move(init);
-  kernel.signature = std::move(signature);
-  kernel.resize = HashAggregateResize;
-  kernel.consume = HashAggregateConsume;
-  kernel.merge = HashAggregateMerge;
-  kernel.finalize = HashAggregateFinalize;
+                               KernelInit init, const bool ordered = false) {
+  HashAggregateKernel kernel(std::move(signature), std::move(init), HashAggregateResize,
```
icexelloss, line 109: same as original code, but refactored a bit so that we cannot have an partially initialized kernel here (the constructor makes sure of that)

cpp/src/arrow/compute/kernels/hash_aggregate.cc
```diff
@@ -1251,6 +1251,210 @@ HashAggregateKernel MakeApproximateMedianKernel(HashAggregateFunction* tdigest_f
...
+
+template <>
+struct NullSentinel<float> {
+  static constexpr float value() { return std::numeric_limits<float>::infinity(); }
+};
+
+template <>
+struct NullSentinel<double> {
+  static constexpr double value() { return std::numeric_limits<double>::infinity(); }
+};
```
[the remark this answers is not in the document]
icexelloss: I removed this and ended up reusing AntiExtrema for the sentinel values

cpp/src/arrow/compute/kernels/aggregate_basic_internal.h
```diff
@@ -272,8 +273,120 @@ struct MeanKernelInit : public SumLikeInit<KernelClass> {
...
+
+    ArrayType arr(arr_span.ToArrayData());
+    const auto null_count = arr.null_count();
+    local.has_nulls = null_count > 0;
+    this->count += arr.length() - null_count;
+
+    if (!local.has_nulls) {
+      for (int64_t i = 0; i < arr.length(); i++) {
+        local.MergeOne(arr.GetView(i));
+      }
```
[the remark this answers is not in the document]
icexelloss: I updated this to be close to what you have. Can you take a look if that looks fine to you?

cpp/src/arrow/compute/kernels/hash_aggregate.cc
icexelloss: Done

cpp/src/arrow/compute/kernels/hash_aggregate.cc
icexelloss: Done

icexelloss:
@westonpace I am reasonably happy about the current code and would like to request a review from you.

This PR is a little overwhelming because lots of test, boiler plate code and templates. But I will mark the actual important piece (I think) for you. Hopefully makes it a little easier. I also included some check marks in the overview to summarize that I did.  The main points that we discussed last time should be addressed.

cpp/src/arrow/compute/kernels/aggregate_basic_internal.h
```diff
@@ -272,8 +272,230 @@ struct MeanKernelInit : public SumLikeInit<KernelClass> {
...
+    this->count += scalar.is_valid;
+
+    if (!local.has_nulls || options.skip_nulls) {
+      local.MergeOne(internal::UnboxScalar<ArrowType>::Unbox(scalar));
+    }
+    this->state += local;
+    return Status::OK();
+  }
+
+  Status ConsumeArray(const ArraySpan& arr_span) {
```
icexelloss, line 422: Main logic for the Scala kernel

cpp/src/arrow/compute/kernels/hash_aggregate.cc
```diff
@@ -1695,6 +1692,384 @@ struct GroupedMinMaxFactory {
...
+    // matter what the value is. We never output the uninitialized
+    // first/last value.
+    RETURN_NOT_OK(firsts_.Append(added_groups, AntiExtrema<CType>::anti_min()));
+    RETURN_NOT_OK(lasts_.Append(added_groups, AntiExtrema<CType>::anti_max()));
+    RETURN_NOT_OK(has_values_.Append(added_groups, false));
+    RETURN_NOT_OK(has_nulls_.Append(added_groups, false));
+    return Status::OK();
+  }
+
+  Status Consume(const ExecSpan& batch) override {
```
icexelloss, line 1728: Main logic for the hash kernel

icexelloss: (Consume, Merge and Finalize)

cpp/src/arrow/compute/kernels/hash_aggregate.cc
icexelloss: I failed to address this because the templates make it hard to refactor and share the logic between "MinMax" and "FirstLast" for binary types. if that's ok I would like to leave that as follow up to better refactor this and avoid adding complexity to this PR.

cpp/src/arrow/compute/function_test.cc
```diff
@@ -323,7 +323,7 @@ TEST(ScalarAggregateFunction, DispatchExact) {
 
   std::vector<InputType> in_args = {int8()};
   ScalarAggregateKernel kernel(std::move(in_args), int64(), NoopInit, NoopConsume,
-                               NoopMerge, NoopFinalize);
+                               NoopMerge, NoopFinalize, /*ordered*/ false);
```
westonpace, line 326:
```suggestion
                               NoopMerge, NoopFinalize, /*ordered=*/ false);
```
Minor consistency nit

cpp/src/arrow/compute/kernels/aggregate_basic.cc
```diff
@@ -439,6 +440,46 @@ struct ProductInit {
...
+  ARROW_ASSIGN_OR_RAISE(TypeHolder out_type,
+                        args.kernel->signature->out_type().Resolve(ctx, args.inputs));
+
+  FirstLastInitState<SimdLevel::NONE> visitor(
+      ctx, *args.inputs[0], out_type.GetSharedPtr(),
+      static_cast<const ScalarAggregateOptions&>(*args.options));
+  return visitor.Create();
+}
+
+// For "first" and "last" functions: override finanlize and return the actual value
```
westonpace, line 457:
```suggestion
// For "first" and "last" functions: override finalize and return the actual value
```

cpp/src/arrow/compute/kernels/aggregate_basic.cc
```diff
@@ -894,6 +954,28 @@ const FunctionDoc mean_doc{
     {"array"},
     "ScalarAggregateOptions"};
 
+const FunctionDoc first_last_doc{"Compute the first and last values of an array",
+                                 ("Do not use this directly. This is internal and might"
+                                  "be removed in the future."),
```
westonpace, line 959: We can probably remove this warning.  I do believe that "always calculating the first and last in one pass" is not ideal when the user just wants the first or just wants the last (e.g. compared to min/max where you have to iterate the entire array no matter what).

However, I do think there will still be scenarios where a user will want both the first and last in a single pass.  So let's go ahead and assume this function is here to stay.

cpp/src/arrow/compute/kernels/aggregate_basic_internal.h
```diff
@@ -272,8 +272,230 @@ struct MeanKernelInit : public SumLikeInit<KernelClass> {
 };
 
 // ----------------------------------------------------------------------
-// MinMax implementation
+// Last implementation
+template <typename ArrowType, SimdLevel::type SimdLevel, typename Enable = void>
```
westonpace, line 276:
```suggestion
template <typename ArrowType, typename Enable = void>
```
As far as I can tell `SimdLevel` is completely unused.

westonpace (changes requested):
I've got some initial comments without going into detail in the kernel.  I'll make a pass at trying to simplify the kernel later today and that should help me understand it better.

It should be pretty easy (though maybe not trivial) to add multi-threaded support too.

However, I don't see anything major.  I think adding `ordered` to the aggregate kernel helps prevent new kernel types and seems sufficient.

westonpace:
CC @felipecrv and @benibus who might be interested as they've done some work in kernels recently.

cpp/src/arrow/compute/kernels/aggregate_basic_internal.h
icexelloss: You are right - This is an side effect of my copying the existing MinMax template.. Will remove

cpp/src/arrow/acero/hash_aggregate_test.cc
```diff
@@ -4206,6 +4206,235 @@ TEST_P(GroupBy, MinMaxWithNewGroupsInChunkedArray) {
...
+])",
+                                                  R"([
+    [0,    2],
+    [1,    null],
+    [null, 3]
+])"};
+
+  const std::string default_expected =
+      R"([
+    [1,    1,    3,    null,   null],
```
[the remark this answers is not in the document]
westonpace: @icexelloss ping

ianmcook:
I believe this should also be documented on the [Compute Functions docs page](https://arrow.apache.org/docs/cpp/compute.html), which is generated from [`docs/source/cpp/compute.rst`](https://github.com/apache/arrow/blob/main/docs/source/cpp/compute.rst). @icexelloss could you add a checkbox for that to the PR description please?

icexelloss:
> I believe these functions should also be documented on the [Compute Functions docs page](https://arrow.apache.org/docs/cpp/compute.html), which is generated from [`docs/source/cpp/compute.rst`](https://github.com/apache/arrow/blob/main/docs/source/cpp/compute.rst). @icexelloss could you add a checkbox for that to the PR description please?

Yep will do

cpp/src/arrow/compute/kernels/aggregate_basic_internal.h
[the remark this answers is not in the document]
icexelloss: I think I hit this with one of the test but let me double check

cpp/src/arrow/compute/kernels/aggregate_basic.cc
icexelloss: Sounds good. Will update.

cpp/src/arrow/acero/aggregate_node.cc
```diff
@@ -716,6 +733,15 @@ class GroupByNode : public ExecNode, public TracedNode {
     ARROW_ASSIGN_OR_RAISE(auto args, MakeAggregateNodeArgs(input_schema, keys,
                                                            segment_keys, aggs, exec_ctx));
 
+    if (concurreny > 1) {
```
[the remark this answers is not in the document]
icexelloss: Good point. Let me check and remove.

cpp/src/arrow/acero/hash_aggregate_test.cc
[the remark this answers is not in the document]
icexelloss: Good point. Now I think of this, the behavior of `skip_nulls=false` you described makes more sense. I will update this.

cpp/src/arrow/compute/kernels/aggregate_basic_internal.h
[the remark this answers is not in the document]
westonpace: Yes, it seems my intuition was only correct for scalar functions and not aggregate functions.  We can leave it as it is for now but there is an opportunity for simplification here.

cpp/src/arrow/compute/kernels/aggregate_basic_internal.h
```diff
@@ -272,8 +272,230 @@ struct MeanKernelInit : public SumLikeInit<KernelClass> {
...
+
+    this->state += local;
+    return Status::OK();
+  }
+
+  Status MergeFrom(KernelContext*, KernelState&& src) override {
+    const auto& other = checked_cast<const ThisType&>(src);
+    this->state += other.state;
+    this->count += other.count;
+    return Status::OK();
```
westonpace, line 465: This is interesting.  This method is used because, in multithreaded execution, we have a state-per-thread.  However, you're limited to single threaded here.  And, even if we add multi-threaded support to the aggregate node (via a sequencing queue) I think this part will still have to be sequenced with a single state object.

Technically this should only work if `this` comes before `src` in "data order" which seems like a very odd constraint for this part of the code.  It really seems that ordered kernels shouldn't need any concept of "merge" because there's no way to divide the work.

I was tempted to say we can just return `Status::Invalid` here but it turns out that we do call `MergeFrom` on every batch although this seems to be [a mistake](https://github.com/apache/arrow/blob/release-11.0.0-rc0/cpp/src/arrow/compute/exec.cc#L1101-L1102).  Regardless, it is probably ok to leave this as it is.

cpp/src/arrow/compute/kernels/aggregate_basic_internal.h
```diff
@@ -272,8 +272,230 @@ struct MeanKernelInit : public SumLikeInit<KernelClass> {
...
+      auto null_scalar = MakeNullScalar(child_type);
+      values = {null_scalar, null_scalar};
+    } else if (state.has_values) {
+      ARROW_ASSIGN_OR_RAISE(auto first_scalar, MakeScalar(child_type, state.first));
+      ARROW_ASSIGN_OR_RAISE(auto last_scalar, MakeScalar(child_type, state.last));
+      values = {first_scalar, last_scalar};
+    } else {
+      auto null_scalar = MakeNullScalar(child_type);
+      values = {null_scalar, null_scalar};
+    }
```
westonpace, line 486: This branch seems redundant.  If `state.has_values` is false then `this->count` must be 0 correct?

cpp/src/arrow/acero/hash_aggregate_test.cc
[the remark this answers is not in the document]
lidavidm: As I remember, they mimic dplyr's `na.rm`:

> If na.rm is FALSE an NA value in any of the arguments will cause a value of NA to be returned, otherwise NA values are ignored. 

https://rdrr.io/r/base/Extremes.html

dplyr appears to interpret `na_rm` differently for `slice` (~= first, last):

> Should missing values in order_by be removed from the result? If FALSE, NA values are sorted to the end (like in [arrange()](https://dplyr.tidyverse.org/reference/arrange.html)), so they will only be included if there are insufficient non-missing values to reach n/prop.

https://dplyr.tidyverse.org/reference/slice.html

westonpace (commented):
A few more observations.  None of them major I think.  I think the only outstanding item is how we want to interpret `skip_nulls`.

cpp/src/arrow/acero/hash_aggregate_test.cc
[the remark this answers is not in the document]
westonpace: I think this is closer to [`first`/`last`](https://dplyr.tidyverse.org/reference/nth.html) which states:

> na_rm : Should missing values in x be removed before extracting the value?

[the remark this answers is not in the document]
westonpace: Seems like we're all on the same page here.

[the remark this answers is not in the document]
lidavidm: I'm going to unsubscribe for my inbox's sake, ping me again if something comes up

cpp/src/arrow/compute/kernels/aggregate_basic_internal.h
icexelloss: Updated this

cpp/src/arrow/compute/kernels/aggregate_basic_internal.h
```diff
@@ -272,8 +272,230 @@ struct MeanKernelInit : public SumLikeInit<KernelClass> {
...
+
+  Status Consume(KernelContext*, const ExecSpan& batch) override {
+    if (batch[0].is_array()) {
+      return ConsumeArray(batch[0].array);
+    }
+    return ConsumeScalar(*batch[0].scalar);
+  }
+
+  Status ConsumeScalar(const Scalar& scalar) {
+    StateType local;
```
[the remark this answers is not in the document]
icexelloss: Removed

cpp/src/arrow/compute/kernels/aggregate_basic.cc
icexelloss: Updated

cpp/src/arrow/acero/hash_aggregate_test.cc
```diff
@@ -4206,6 +4206,235 @@ TEST_P(GroupBy, MinMaxWithNewGroupsInChunkedArray) {
                     /*verbose=*/true);
 }
 
+TEST_P(GroupBy, FirstLastBasicTypes) {
+  std::vector<std::shared_ptr<DataType>> types;
+  types.insert(types.end(), boolean());
+  types.insert(types.end(), NumericTypes().begin(), NumericTypes().end());
+  types.insert(types.end(), TemporalTypes().begin(), TemporalTypes().end());
+
+  const std::vector<std::string> default_table = {R"([
```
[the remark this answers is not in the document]
icexelloss: Updated

cpp/src/arrow/acero/aggregate_node.cc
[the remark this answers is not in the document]
icexelloss: Moved to MakeAggregateNodeArgs (there are actually two MakeAggregateNodeArgs methods, one for scalar and one for hash)

cpp/src/arrow/compute/kernels/aggregate_basic_internal.h
icexelloss: As a side note I noticed that it will always merge the an empty state with the first state (the one that is called with ConsumeArray), which seems unnecessary (merge of empty state and first state should just be the first state by definition). 

But anyways, I agree we should probably leave as is for the purpose of this PR.

icexelloss:
@westonpace I addressed all the comments I think.

For skip_nulls=false, it's actually non-trivial to support and I ended up adding three more flags (first_is_null, last_is_null, has_any_values) to support this. Please take look.

westonpace (approved):
This matches the null behavior I would expect.  Thanks for fixing that.  This looks good, appreciate the persistence.

PR #34931, opened 2023-04-06
GH-34930:  [C++] Standardize aggregation column order

### LINKED ISSUE #34908: [C++] Switching the column ordering of "segment key" and "hash key"

Recently, this PR changes the ordering output columns of aggregation to
`(hash)keys + segment_keys + aggregates`
(https://github.com/apache/arrow/blob/main/cpp/src/arrow/acero/aggregate_node.cc#L654)

I wondering it makes more sense to change this to
`segment_keys + (hash)keys + aggregates`

My reasoning is as follows:
This only really affects the result in the case of segmented aggregation so I will solely reason about this in the context of segmented aggregation (in non-segmented aggregation, segment_keys is empty so the result wouldn't change regardless).

In segmented aggregation, the hash key is used within each segment (i.e., we don't combine grouped from different segments together) so in a way, the hash key feels like "second key", i.e., data is first split up by segments, then by hash keys. And therefore, it feels more natural to order the output columns: primary key (segment keys) + secondary key (hash keys) + aggregates. 

To give an example of time ordered aggregations (an application of segmented aggregations). where segment key = "time" and hash key = "id", we almost always wanted the output to be "time, id" instead of "id, time" because the full stream is time ordered so it makes sense to make that the first column.

@westonpace I wonder what is your thoughts on switching this up? If you feel strongly about keeping the current order I think we might still be able to tweak this at the substrait level with emit output to get the order we want, but I figured it's probably easier / more consistent if we just switch up the default ordering. 

I am happy to make the change if that sounds ok but want to discuss first.

 

### Component(s)

C++

### LINKED ISSUE #34930: [C++] Standardize aggregation column order

Change aggregation column order to segment-keys then keys then aggregates.

### Component(s)

C++

rtpsw:
See #34930, #34908
* Closes: #34930
* Closes: #34908

### DIFF
--- a/cpp/src/arrow/acero/aggregate_node.cc
+++ b/cpp/src/arrow/acero/aggregate_node.cc
@@ -651,19 +651,19 @@ class GroupByNode : public ExecNode, public TracedNode {
     // Build field vector for output schema
     FieldVector output_fields{keys.size() + segment_keys.size() + aggs.size()};
 
-    // First output is keys, followed by segment_keys, followed by aggregates themselves
+    // First output is segment keys, followed by keys, followed by aggregates themselves
     // This matches the behavior described by Substrait and also tends to be the behavior
     // in SQL engines
-    for (size_t i = 0; i < keys.size(); ++i) {
-      int key_field_id = key_field_ids[i];
-      output_fields[i] = input_schema->field(key_field_id);
-    }
-    size_t base = keys.size();
     for (size_t i = 0; i < segment_keys.size(); ++i) {
       int segment_key_field_id = segment_key_field_ids[i];
-      output_fields[base + i] = input_schema->field(segment_key_field_id);
+      output_fields[i] = input_schema->field(segment_key_field_id);
     }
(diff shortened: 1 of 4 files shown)


rtpsw:
cc @icexelloss

icexelloss (approved):
LGTM

icexelloss:
@westonpace This one looks good to me. Let me know if this looks good to you.

icexelloss:
@westonpace Since this is a pretty straight forward change and this has been agreed on in the issue:

https://github.com/apache/arrow/issues/34908#issuecomment-1500355286

and I haven't heard back from you. I am going to merge this. If you have some requests for changes we can address as follow up.

westonpace:
> and I haven't heard back from you. I am going to merge this. If you have some requests for changes we can address as follow up.

That's fine.  I'm sorry I didn't get to this earlier.  I've been heads down this week on resolving 12.0.0 blockers and fallen behind on some reviews.

PR #34948, opened 2023-04-07
GH-34897: [Python] Don't attempt to aquire the GIL when Python is finalizing

### LINKED ISSUE #34897: [R][Python] Segfault on session exit after materializing ALTREP character vectors imported from Python

As described and reproduced nicely in https://github.com/rpy2/rpy2-arrow/issues/11

This is almost certainly a result of https://github.com/apache/arrow/pull/34489 since it seems to involve fully materialized arrays. Otherwise, I'm not sure exactly what would cause this specific to shipping arrays over the C Data interface. The traceback reported is below (I will try to reproduce as well):

I suspect this might have to do with the fact that the session shutting down might lead to some things happening with the GIL and R is trying to release that memory at a very inconvenient time.

```
Thread 1 "python" received signal SIGSEGV, Segmentation fault.
0x00005555557868b0 in new_threadstate (interp=0x0, init=1) at ../Python/pystate.c:616
616	../Python/pystate.c: No such file or directory.
(gdb) up
#1  0x00005555557a4ee2 in PyThreadState_New (interp=<optimized out>)
    at ../Python/pystate.c:684
684	../Python/pystate.c: No such file or directory.
(gdb) up
#2  PyGILState_Ensure () at ../Python/pystate.c:1504
1504	in ../Python/pystate.c
(gdb) up
#3  0x00007fffe0ee6003 in arrow::py::NumPyBuffer::~NumPyBuffer() ()
   from /opt/software/python/py310_env/lib/python3.10/site-packages/pyarrow/libarrow_python.so
(gdb) up
#4  0x00007fffe0ecb56d in std::_Sp_counted_ptr_inplace<arrow::ArrayData, std::allocator<arrow::ArrayData>, (__gnu_cxx::_Lock_policy)2>::_M_dispose() ()
   from /opt/software/python/py310_env/lib/python3.10/site-packages/pyarrow/libarrow_python.so
(gdb) up
#5  0x00007fffe0ecb4d5 in std::_Sp_counted_ptr_inplace<arrow::ArrayData, std::allocator<arrow::ArrayData>, (__gnu_cxx::_Lock_policy)2>::_M_dispose() ()
   from /opt/software/python/py310_env/lib/python3.10/site-packages/pyarrow/libarrow_python.so
(gdb) up
#6  0x00007fffde32f04a in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release() ()
   from /opt/software/python/py310_env/lib/python3.10/site-packages/pyarrow/libarrow.so.1100
(gdb) up
#7  0x00007fffdee02a4c in arrow::(anonymous namespace)::ReleaseExportedArray(ArrowArray*) ()
   from /opt/software/python/py310_env/lib/python3.10/site-packages/pyarrow/libarrow.so.1100
(gdb) up
#8  0x00007fffd0aab6f7 in std::_Sp_counted_ptr_inplace<arrow::(anonymous namespace)::ImportedArrayData, std::allocator<arrow::(anonymous namespace)::ImportedArrayData>, (__gnu_cxx::_Lock_policy)2>::_M_dispose() ()
   from /usr/local/packages/R/4.2/lib/R/library/arrow/libs/arrow.so
(gdb) up
#9  0x00007fffd04eda4a in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release (this=0x55555c8b2a70)
    at /usr/include/c++/11/bits/shared_ptr_base.h:168
168		    _M_dispose();
(gdb) 
```

### Component(s)

R

paleolimbot:
This is one possible solution to #34897...I admittedly don't know much about what's involved here but I'm hoping to use this PR to generate a wheel that can help us track down the underlying cause.
* Closes: #34897

### DIFF
--- a/python/pyarrow/src/arrow/python/numpy_convert.cc
+++ b/python/pyarrow/src/arrow/python/numpy_convert.cc
@@ -53,8 +53,10 @@ NumPyBuffer::NumPyBuffer(PyObject* ao) : Buffer(nullptr, 0) {
 }
 
 NumPyBuffer::~NumPyBuffer() {
-  PyAcquireGIL lock;
-  Py_XDECREF(arr_);
+  if (!_Py_IsFinalizing()) {
+    PyAcquireGIL lock;
+    Py_XDECREF(arr_);
+  }
 }
 
 #define TO_ARROW_TYPE_CASE(NPY_NAME, FACTORY) \


danepitkin:
Hmm interesting. From the Python C API docs[1] when acquiring the GIL:
```
Note Calling this function from a thread when the runtime is finalizing will terminate the thread, even if the thread was not created by Python. You can use _Py_IsFinalizing() or [sys.is_finalizing()](https://docs.python.org/3/library/sys.html#sys.is_finalizing) to check if the interpreter is in process of being finalized before calling this function to avoid unwanted termination.
```

I wonder if we should be checking for Py_IsFinalizing everywhere we acquire the GIL? 

[1] https://docs.python.org/3/c-api/init.html#c.PyGILState_Ensure

PR #34955, opened 2023-04-07
GH-34335: [C++][Parquet] Optimize Decoding DELTA_LENGTH_BYTE_ARRAY

### LINKED ISSUE #34335: [C++][Parquet] Improve the performance for Decoding DELTA_LENGTH_BYTE_ARRAY

mapleFU:
### Rationale for this change

According to https://github.com/apache/arrow/pull/34323 . DELTA_LENGTH_BYTE_ARRAY is much more slower. So do some optimizations.

Some tiny changes

* Closes: #34335

### DIFF
--- a/cpp/src/parquet/encoding.cc
+++ b/cpp/src/parquet/encoding.cc
@@ -2738,7 +2738,7 @@ class DeltaLengthByteArrayDecoder : public DecoderImpl,
     int32_t data_size = 0;
     const int32_t* length_ptr =
         reinterpret_cast<const int32_t*>(buffered_length_->data()) + length_idx_;
-    int bytes_left = decoder_->bytes_left();
+    int bytes_offset = len_ - decoder_->bytes_left();
     for (int i = 0; i < max_values; ++i) {
       int32_t len = length_ptr[i];
       if (ARROW_PREDICT_FALSE(len < 0)) {
@@ -2753,7 +2753,7 @@ class DeltaLengthByteArrayDecoder : public DecoderImpl,
     if (ARROW_PREDICT_FALSE(!decoder_->Advance(8 * static_cast<int64_t>(data_size)))) {
       ParquetException::EofException();
     }
-    const uint8_t* data_ptr = data_ + (len_ - bytes_left);
+    const uint8_t* data_ptr = data_ + bytes_offset;
     for (int i = 0; i < max_values; ++i) {
       buffer[i].ptr = data_ptr;
       data_ptr += buffer[i].len;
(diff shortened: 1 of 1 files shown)


mapleFU:
@rok @wjones127 @wgtmac

cpp/src/parquet/encoding.cc
```diff
@@ -2745,6 +2738,7 @@ class DeltaLengthByteArrayDecoder : public DecoderImpl,
     int32_t data_size = 0;
     const int32_t* length_ptr =
         reinterpret_cast<const int32_t*>(buffered_length_->data()) + length_idx_;
+    int bytes_left = decoder_->bytes_left();
```
wjones127, line 2741: nit: could we rewrite this as `bytes_offset`? That way it's clearer why we are computing this before calling `decoder_->Advance()`

```suggestion
    int bytes_offset = len_ - decoder_->bytes_left();
```

cpp/src/parquet/encoding.cc
```diff
@@ -2756,13 +2750,10 @@ class DeltaLengthByteArrayDecoder : public DecoderImpl,
...
     length_idx_ += max_values;
-
-    PARQUET_THROW_NOT_OK(buffered_data_->Resize(data_size));
-    if (decoder_->GetBatch(8, buffered_data_->mutable_data(), data_size) != data_size) {
+    if (ARROW_PREDICT_FALSE(!decoder_->Advance(8 * static_cast<int64_t>(data_size)))) {
       ParquetException::EofException();
     }
-    const uint8_t* data_ptr = buffered_data_->data();
-
+    const uint8_t* data_ptr = data_ + (len_ - bytes_left);
```
wjones127, line 2756:
```suggestion
    const uint8_t* data_ptr = data_ + bytes_offset;
```

cpp/src/parquet/encoding.cc
```diff
@@ -3071,8 +3061,14 @@ class DeltaByteArrayDecoder : public DecoderImpl,
     prefix_len_offset_ = 0;
     num_valid_values_ = num_prefix;
 
+    int bytes_left = decoder_->bytes_left();
+    int suffix_begins = len - bytes_left;
+    if (suffix_begins < 0) {
+      throw ParquetException(
+          "Wrong data in DeltaByteArrayDecoder: bytes_left more than length");
+    }
```
wjones127, line 3069: If this happens, doesn't that mean we have already read past the buffer end (buffer overflow)? Seems like that should be handled in the lengths decoder.

wjones127 (changes requested):

mapleFU:
@wjones127 comment resolved, you can take a look when you have spare time :)

wjones127 (approved):

PR #34958, opened 2023-04-07
GH-34944: [Python] Fix crash when converting non-sequence object with getitem in pa.array()

### LINKED ISSUE #34944: [Python] PyArrow pa.array fails intermittently with custom iterable object

- Arrow versions tested: 10.0.1 and 11.0.0, installed via conda
- Platform: Linux

I worked with @shwina recently on a problem we saw in cudf, and we identified a bug in PyArrow. The bug can be reproduced (but only intermittently) with the following snippet:

```python
import pyarrow as pa

class A:
    def __getitem__(self, key):
        return 3

pa.array(A())
```

I can run this snippet under `pdb` by inserting `breakpoint()` before `pa.array(A())` and then continuing from the breakpoint by pressing `c`. With `pdb`, I consistently get errors like:
```
WARNING: Logging before InitGoogleLogging() is written to STDERR
F20230406 15:12:07.595571 153661 inference.cc:348]  Check failed: _s.ok() Operation failed: internal::ImportDecimalType(&decimal_type_)
Bad status: Unknown error: <built-in function __import__> returned a result with an exception set. Detail: Python exception: SystemError
*** Check failure stack trace: ***
```

Without `pdb`, the error is _sometimes_ this one (which I expect) and _sometimes_ the crash shown above.
```python
  File "/home/bdice/issue.py", line 16, in <module>
    pa.array(A())
  File "pyarrow/array.pxi", line 320, in pyarrow.lib.array
  File "pyarrow/array.pxi", line 39, in pyarrow.lib._sequence_to_array
TypeError: object of type 'A' has no len()
```

We want a `TypeError` to be raised rather than getting the crash above under `pdb`. The crash is reproducible without `pdb` if the snippet is run repeatedly. This might suggest some kind of memory corruption is happening behind the scenes.

### Component(s)

C++, Python

jorisvandenbossche:

Some python objects can pass the `PySequence_Check` without being "proper" sequences with a length, resulting in a subsequent `PySequence_Size` to fail, but we didn't check for python errors there, and so failed to properly raise this as a python exception.
* Closes: #34944

### DIFF
--- a/python/pyarrow/src/arrow/python/python_to_arrow.cc
+++ b/python/pyarrow/src/arrow/python/python_to_arrow.cc
@@ -1099,6 +1099,7 @@ Status ConvertToSequenceAndInferSize(PyObject* obj, PyObject** seq, int64_t* siz
   if (PySequence_Check(obj)) {
     // obj is already a sequence
     int64_t real_size = static_cast<int64_t>(PySequence_Size(obj));
+    RETURN_IF_PYERROR();
     if (*size < 0) {
       *size = real_size;
     } else {
--- a/python/pyarrow/tests/test_convert_builtin.py
+++ b/python/pyarrow/tests/test_convert_builtin.py
@@ -133,6 +133,18 @@ def test_failing_iterator():
         pa.array((1 // 0 for x in range(10)), size=10)
 
 
+class ObjectWithOnlyGetitem:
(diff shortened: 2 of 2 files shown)


bdice (approved):
Fantastic. This looks right. Thanks @jorisvandenbossche for the quick fix!

PR #34961, opened 2023-04-07
GH-34960: [C++] test util Fixing arrow Random Generator for lost nullable info

### LINKED ISSUE #34960: [C++] TestUtil: RandomArrayGenerator will lost the "nullable" info when build nested list

```C++
  {
    auto item = std::make_shared<arrow::Field>("item", arrow::int8(), true);
    auto nestListField =
        std::make_shared<arrow::Field>("list", arrow::list(item), false);
    auto listField =
        std::make_shared<arrow::Field>("list", arrow::list(nestListField), true);
    auto array = rng.ArrayOf(*listField, 428);
    ARROW_EXPECT_OK(array->ValidateFull());

    auto batch = rng.BatchOf({listField}, 428);
    ARROW_EXPECT_OK(batch->ValidateFull());
  }
```

The code above will validate failed, because it lost the nullable info in nested list

### Component(s)

C++

mapleFU:
### Rationale for this change

Currently, nonnull in nested schema will be losted by random generator. This patch make it not lost.

Using ctor with DataType to deduce the field type.

* Closes: #34960

### DIFF
--- a/cpp/src/arrow/testing/random.cc
+++ b/cpp/src/arrow/testing/random.cc
@@ -916,11 +916,11 @@ std::shared_ptr<Array> RandomArrayGenerator::ArrayOf(const Field& field, int64_t
 
     case Type::type::STRUCT: {
       ArrayVector child_arrays(field.type()->num_fields());
-      FieldVector child_fields;
+      FieldVector child_fields(field.type()->num_fields());
       for (int i = 0; i < field.type()->num_fields(); i++) {
         const auto& child_field = field.type()->field(i);
         child_arrays[i] = ArrayOf(*child_field, length, alignment, memory_pool);
-        child_fields.push_back(child_field);
+        child_fields[i] = child_field;
       }
       return *StructArray::Make(
           child_arrays, child_fields,
--- a/cpp/src/arrow/testing/random.cc
+++ b/cpp/src/arrow/testing/random.cc
@@ -793,7 +793,7 @@ std::shared_ptr<Array> RandomArrayGenerator::ArrayOf(const Field& field, int64_t
                 values_length, alignment, memory_pool);                              \
     const auto offsets = OffsetsFromLengthsArray(lengths.get(), force_empty_nulls,   \
                                                  alignment, memory_pool);            \
-    return *ARRAY_TYPE::FromArrays(*offsets, *values);                               \
+    return *ARRAY_TYPE::FromArrays(field.type(), *offsets, *values);                 \
   }
 
   const double null_probability =
@@ -916,14 +916,14 @@ std::shared_ptr<Array> RandomArrayGenerator::ArrayOf(const Field& field, int64_t
 
     case Type::type::STRUCT: {
       ArrayVector child_arrays(field.type()->num_fields());
-      std::vector<std::string> field_names;
+      FieldVector child_fields;
       for (int i = 0; i < field.type()->num_fields(); i++) {
         const auto& child_field = field.type()->field(i);
         child_arrays[i] = ArrayOf(*child_field, length, alignment, memory_pool);
-        field_names.push_back(child_field->name());
+        child_fields.push_back(child_field);
       }
       return *StructArray::Make(
-          child_arrays, field_names,
+          child_arrays, child_fields,
           NullBitmap(length, null_probability, alignment, memory_pool));
     }
 
--- a/cpp/src/arrow/testing/random_test.cc
+++ b/cpp/src/arrow/testing/random_test.cc
@@ -499,43 +499,44 @@ TEST(RandomList, Basics) {
   }
 }
 
-TEST(RandomNestedBatchOf, Basics) {
+TEST(RandomChildFieldNullablity, List) {
   random::RandomArrayGenerator rng(42);
-  {
-    auto item = std::make_shared<arrow::Field>("item", arrow::int8(), true);
-    auto nestListField = std::make_shared<arrow::Field>("list", arrow::list(item), false);
-    auto listField =
-        std::make_shared<arrow::Field>("list", arrow::list(nestListField), true);
-    auto array = rng.ArrayOf(*listField, 428);
-    ARROW_EXPECT_OK(array->ValidateFull());
 
-    auto batch = rng.BatchOf({listField}, 428);
-    ARROW_EXPECT_OK(batch->ValidateFull());
-  }
-  {
-    auto item = std::make_shared<arrow::Field>("item", arrow::int8(), true);
-    auto nestStructField =
-        std::make_shared<arrow::Field>("struct", arrow::struct_({item}), false);
-    auto structField =
-        std::make_shared<arrow::Field>("struct", arrow::struct_({nestStructField}), true);
-    auto array = rng.ArrayOf(*structField, 428);
-    ARROW_EXPECT_OK(array->ValidateFull());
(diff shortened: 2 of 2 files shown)


mapleFU:
@wjones127 @lidavidm Mind take a look?

lidavidm:
I'm OOO so I probably can't get to this for a while, sorry

mapleFU:
Don't worry, you can review it any time :)

cpp/src/arrow/testing/random.cc
```diff
@@ -793,7 +793,7 @@ std::shared_ptr<Array> RandomArrayGenerator::ArrayOf(const Field& field, int64_t
                 values_length, alignment, memory_pool);                              \
     const auto offsets = OffsetsFromLengthsArray(lengths.get(), force_empty_nulls,   \
                                                  alignment, memory_pool);            \
-    return *ARRAY_TYPE::FromArrays(*offsets, *values);                               \
+    return *ARRAY_TYPE::FromArrays(field.type(), *offsets, *values);                 \
```
wgtmac, line 796: Does map type have similar issue?

cpp/src/arrow/testing/random.cc
```diff
@@ -916,14 +916,14 @@ std::shared_ptr<Array> RandomArrayGenerator::ArrayOf(const Field& field, int64_t
...
+      FieldVector child_fields;
       for (int i = 0; i < field.type()->num_fields(); i++) {
         const auto& child_field = field.type()->field(i);
         child_arrays[i] = ArrayOf(*child_field, length, alignment, memory_pool);
-        field_names.push_back(child_field->name());
+        child_fields.push_back(child_field);
       }
       return *StructArray::Make(
-          child_arrays, field_names,
+          child_arrays, child_fields,
```
wgtmac, line 926: Shouldn't the original overload get the data type from `child_arrays`? I saw a comment of the current overload saying `This method does not check that field types and child array types are consistent.` It seems that we need to fix the original `Make` function as well.

cpp/src/arrow/testing/random.cc
mapleFU: No, the problem is a bit tricky. Let me explain it:
1. Arrow has `DataType`, `DataType` doesn't including some nullable info itself. `Field` has nullable info.
2. For `DataType` , it child would be array of `Field`, so DataType has children's nullable info

Here if doesn't pass `field.type()` in ctor, it will mark child as nullable. And Map's constructor already marks it.

cpp/src/arrow/testing/random.cc
mapleFU: Yes, In fact I use `Generator` to generate testing data, but it `ValidationFull` fails, lol
Maybe @lidavidm can take a look?

mapleFU:
@westonpace Hi, I use `BatchOf` to generate a random List/Struct, however, the generated data call `ValidateFull` failed. This patch is a fixing. Would you mind take a look?

cpp/src/arrow/testing/random_test.cc
```diff
@@ -499,6 +499,45 @@ TEST(RandomList, Basics) {
   }
 }
 
+TEST(RandomNestedBatchOf, Basics) {
+  random::RandomArrayGenerator rng(42);
+  {
+    auto item = std::make_shared<arrow::Field>("item", arrow::int8(), true);
+    auto nestListField = std::make_shared<arrow::Field>("list", arrow::list(item), false);
```
lidavidm, line 506: nit: use `arrow::field` helper, remember to use `snake_case` variable names

cpp/src/arrow/testing/random.cc
lidavidm: Specifically, if you look at the case for MapType, it already passes the type. So this is just bringing ListType into alignment.

cpp/src/arrow/testing/random.cc
lidavidm: Sorry, what are you saying fails? I agree the proposed code is more correct.

cpp/src/arrow/testing/random_test.cc
```diff
@@ -499,6 +499,45 @@ TEST(RandomList, Basics) {
   }
 }
 
+TEST(RandomNestedBatchOf, Basics) {
```
lidavidm, line 502: Can this just be part of the `RandomList` suite as a new case?

cpp/src/arrow/testing/random_test.cc
```diff
@@ -499,6 +499,45 @@ TEST(RandomList, Basics) {
...
+    auto nestListField = std::make_shared<arrow::Field>("list", arrow::list(item), false);
+    auto listField =
+        std::make_shared<arrow::Field>("list", arrow::list(nestListField), true);
+    auto array = rng.ArrayOf(*listField, 428);
+    ARROW_EXPECT_OK(array->ValidateFull());
+
+    auto batch = rng.BatchOf({listField}, 428);
+    ARROW_EXPECT_OK(batch->ValidateFull());
+  }
+  {
```
lidavidm, line 515: Can this be made into a `RandomStruct` suite?

cpp/src/arrow/testing/random_test.cc
```diff
@@ -499,6 +499,45 @@ TEST(RandomList, Basics) {
...
+        std::make_shared<arrow::Field>("struct", arrow::struct_({item}), false);
+    auto structField =
+        std::make_shared<arrow::Field>("struct", arrow::struct_({nestStructField}), true);
+    auto array = rng.ArrayOf(*structField, 428);
+    ARROW_EXPECT_OK(array->ValidateFull());
+
+    auto batch = rng.BatchOf({structField}, 428);
+    ARROW_EXPECT_OK(batch->ValidateFull());
+  }
+  {
```
lidavidm, line 527: And `RandomMap` here.

cpp/src/arrow/testing/random_test.cc
lidavidm: And for all three tests, can you call out specifically that we want to test child field nullability is preserved?

cpp/src/arrow/testing/random.cc
mapleFU: Yes, this patch brings `List` and `Struct` into alignment. For testing, I also tests map

cpp/src/arrow/testing/random.cc
lidavidm: Ok, but I don't really get what this thread is trying to discuss.

@wgtmac note that types do not carry nullability info. So we need to use fields here, this isn't a problem with Make.

mapleFU:
Comment resolved

lidavidm (approved):

cpp/src/arrow/testing/random.cc
wgtmac: I just realized that `DataType` does not know the nullability but `Field` does. Thanks for reminding me! @lidavidm

wgtmac:
@raulcd Could you please backport this patch to v12.0.0 release? Thanks!

raulcd:
I've taken a look and as it is a minor bug fix that almost made the feature freeze I've added it to the 12.0.0. I'll try to write something to document more clearly what should be backported once the feature freeze has been done. For bug fixes we should probably include:
* Critical fixes
* Regression issues
* Incorrect data issues

PR #34964, opened 2023-04-07
GH-34962: [Go] Make GetOneForMarshal public on Array interface

### LINKED ISSUE #34962: [Go]: Make GetOneForMarshal(i int) public

GetOneForMarshal(i int) interface{} is already public on `arraymarshal` interface but it seems that it can be safely moved to the main `Array` interface.

### Component(s)

Go

yevgenypats:

* Closes: #34962


zeroshade:
@yevgenypats 

Just a thought, interestingly the `GetOneForMarshal` pretty easily fits the `UntypedValue` method when you think about it..... should we just rename it and use that?

zeroshade (approved):

PR #34972, opened 2023-04-07
GH-34971: [Format] Add non-CPU version of C Data Interface

### LINKED ISSUE #34971: [Format] Enhance C-Data API to support non-cpu use cases

Currently the C-Data API assumes that all the buffers are CPU buffers and there is no way to pass device information through the ABI to allow using it for non-cpu memory allocations. We need to enhance the API to allow passing device and memory information without breaking the C-Data ABI, while also ensuring it is future-proofed for future developments with non-cpu devices. 

### Component(s)

Format

zeroshade:
### Rationale for this change
In order to support non-cpu devices and memory usage, we can add new `ArrowDeviceArray` and `ArrowDeviceArrayStream` structs to the C Data Interface in order to allow for handling these types of memory.

Definitions for a new `ArrowDeviceArray`, `ArrowDeviceArrayStream` and `ArrowDeviceType` enums.

* Closes: #34971


cpp/src/arrow/c/abi.h
```diff
@@ -65,6 +65,69 @@ struct ArrowArray {
...
+  kArrowHexagon = 16,
+} ArrowDeviceType;
+
+struct ArrowDeviceArray {
+  // the private_date and release callback of the arrow array
+  // should contain any necessary information and structures
+  // related to freeing the array according to the device it
+  // is allocated on, rather than having a separate release
+  // callback embedded here.
+  struct ArrowArray* array;
```
pitrou, line 119: If you make this a pointer, then who owns the `struct ArrowArray` area?
You should instead embed the struct.
```suggestion
  struct ArrowArray array;
```

cpp/src/arrow/c/abi.h
```diff
@@ -65,6 +65,69 @@ struct ArrowArray {
...
+  int device_id;
```
pitrou, line 120: Is an `int` sufficient, or do we want to be future-proof and make this `int64_t`?

cpp/src/arrow/c/abi.h
```diff
@@ -65,6 +65,69 @@ struct ArrowArray {
...
+  ArrowDeviceType device_type;
+  // reserve 128 bytes for future expansion
+  // of this struct as non-CPU development expands
+  // so that we can update without ABI breaking
+  // changes.
```
pitrou, line 125: That's assuming those reserved bytes are currently forced to zero, right? This should be specified somewhere.

cpp/src/arrow/c/abi.h
```diff
@@ -106,6 +169,77 @@ struct ArrowArrayStream {
...
+  // device_id matches what is provided here, and whose device_type is the
+  // same as the device_type member of this stream.
+  int (*get_next_device_id)(struct ArrowDeviceArrayStream*, int* out_device_id);
+
+  // Callback to get the next array
+  // (if no error and the array is released, the stream has ended)
+  //
+  // the provided stream_ptr should be the appropriate stream, or
+  // equivalent object, for the device that the data is allocated on
+  // to indicate where the consumer wants the data to be accessible.
```
pitrou, line 208: I'm not sure I understand, why would the consumer be able to choose where the data is accessible? If the data is already allocated (which is certainly a common case), how the producer make it accessible to a new stream?

cpp/src/arrow/c/abi.h
```diff
@@ -106,6 +169,77 @@ struct ArrowArrayStream {
...
+  // if stream_ptr is NULL then the default stream (e.g. CUDA stream 0)
+  // should be used to ensure that the memory is accessible from any stream.
```
pitrou, line 210: Does this make sense for all backends, or are we trying to encode device-specific behaviour in a generic API?

cpp/src/arrow/c/abi.h
pitrou: In the CUDA Array Interface, it's the producer that optionally fills in the stream that must be synchronized on to access the data:
https://numba.readthedocs.io/en/stable/cuda/cuda_array_interface.html#synchronization

Besides, there's an interesting statement about stream lifetime here:
https://numba.readthedocs.io/en/stable/cuda/cuda_array_interface.html#streams

paleolimbot: Should this `void*` instead be a member of `ArrowDeviceArray`? It seems like that pointer might be useful to anybody attempting to access data in those buffers?

cpp/src/arrow/c/abi.h
```diff
@@ -65,6 +65,69 @@ struct ArrowArray {
...
+#ifndef ARROW_C_DEVICE_DATA_INTERFACE
+#define ARROW_C_DEVICE_DATA_INTERFACE
+
+// ArrowDeviceType is compatible with dlpack DLDeviceType for portability
+// it uses the same values for each enum as the equivalent kDL<type> from dlpack.h
+#ifdef __cplusplus
+typedef enum : int32_t {
+#else
+typedef enum {
+#endif
```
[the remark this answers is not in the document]
lidavidm: Explicit `int32_t` should be safe AFAIK, but I agree it's a little goofy + is inconsistent with existing use of `#define`s

cpp/src/arrow/c/abi.h
```diff
@@ -65,6 +65,69 @@ struct ArrowArray {
 
 #endif  // ARROW_C_DATA_INTERFACE
 
+#ifndef ARROW_C_DEVICE_DATA_INTERFACE
+#define ARROW_C_DEVICE_DATA_INTERFACE
+
+// ArrowDeviceType is compatible with dlpack DLDeviceType for portability
```
lidavidm, line 71: It would be great to have docstrings for enums, new structs, etc.

cpp/src/arrow/c/abi.h
```diff
@@ -65,6 +65,69 @@ struct ArrowArray {
...
+// it uses the same values for each enum as the equivalent kDL<type> from dlpack.h
+#ifdef __cplusplus
+typedef enum : int32_t {
+#else
+typedef enum {
+#endif
+  // CPU device, same as using ArrowArray directly
+  kArrowCPU = 1,
```
lidavidm, line 79: Should we explicitly have `0` as the 'unset' value then?

cpp/src/arrow/c/abi.h
```diff
@@ -106,6 +169,77 @@ struct ArrowArrayStream {
...
+  //
+  // The next call to `get_next` should provide an ArrowDeviceArray whose
+  // device_id matches what is provided here, and whose device_type is the
+  // same as the device_type member of this stream.
+  int (*get_next_device_id)(struct ArrowDeviceArrayStream*, int* out_device_id);
+
+  // Callback to get the next array
+  // (if no error and the array is released, the stream has ended)
+  //
+  // the provided stream_ptr should be the appropriate stream, or
```
lidavidm, line 206: to be clear, this is "CUDA stream"?

while this is a draft, the language here is a little imprecise/feels like it assumes too much knowledge of a particular programming model. It would be great to define any terminology needed so that things can reasonably stand alone.

zeroshade: This is referring to whichever type is used by the appropriate backend to represent a stream (or should be null if the backend doesn't utilize this concept). 

I'll update the comments to be more explicit about this.

cpp/src/arrow/c/abi.h
zeroshade: As far as I'm aware, this should make sense for all backends. If they utilize the concept of a stream, they all have some "default" stream that can be used if this is null (the example given being stream 0 for CUDA), so I'm fairly certain this isn't device-specific behavior being encoded. @kkraus14 can you fact-check me on this?

cpp/src/arrow/c/abi.h
zeroshade: Stream lifetime is a tough thing to handle for GPUs because everything is async. So after discussing with @kkraus14 we came to the conclusion that the safest API would be for a consumer to specify which stream they wanted to be able to access the data from and the producer would manage any event and stream synchronizing necessary to make it happen. This puts the stream lifetime management into the consumer (where it's easier to manage) rather than the producer.

@kkraus14 you can probably explain this better than I just did.

cpp/src/arrow/c/abi.h
zeroshade: I'll future proof and set it to an `int64_t` just in case.

cpp/src/arrow/c/abi.h
zeroshade: I'll update the comment to specify that they should be zeroed after allocating an instance of this struct

cpp/src/arrow/c/abi.h
pitrou: This seems a bit gratuitously inconsistent with the handling of the lifetime of memory buffers (which goes through the release callback, telling the producer when it can safely release them).

It is also gratuitously inconsistent with Numba's CUDA array interface, and I would trust the Numba developers' experience here.

cpp/src/arrow/c/abi.h
pitrou: Also sidenote, it would be nice if we could find a less ambiguous term, as "stream" already refers to something else in this interface (though I have no concrete suggestion).

cpp/src/arrow/c/abi.h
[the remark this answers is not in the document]
pitrou: @paleolimbot is right, I think, that this threatens to make the ABI incompatible between C and C++. We could either keep this an untyped enum, or switch to macros as suggested.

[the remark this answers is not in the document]
zeroshade: I got the idea for the ifdef to have it be concretely `int32_t` in C++ from the dlpack C API, but I'm fine with switching to an untyped enum or macros if preferred. I personally would prefer an untyped enum, but if everyone thinks the consistency with the ADBC header is more important I'll switch it to macros.

cpp/src/arrow/c/abi.h
zeroshade: I'll fix the comment format to be docstrings. Good point

cpp/src/arrow/c/abi.h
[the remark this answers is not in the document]
lidavidm: I wasn't thinking ADBC - the C Data Interface already uses macros.

cpp/src/arrow/c/abi.h
lidavidm: I agree with Antoine's comment above that we need a different term for this in this context. But my general point is that it would help to have a set of definitions up front. As it stands this seems to assume the reader is already very familiar with CUDA specifically. That may not be great if the idea is to support non-CPU memory more generally.

cpp/src/arrow/c/abi.h
kkraus14: I drove a lot of the development on the `__cuda_array_interface__` with @seibert and @sklam and the stream was added later and I haven't seen it used. The challenge is that most frameworks internally manage their streams and don't have the ability to share a stream and control its lifetime nicely.

The idea of a consumer of the interface passing a stream to the producer of the interface is that the producer can guarantee that the data is safe to consume on the passed in stream. A bunch of array library maintainers including folks from Tensorflow, PyTorch, JAX, CuPy, etc. all agreed on these semantics: https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack__.html

cpp/src/arrow/c/abi.h
kkraus14: CUDA (NVIDIA GPUs) and ROCm (AMD GPUs) both use the term "stream". The SYCL equivalent is an in-order "queue". I'm not familiar enough with other device programming models to know what their terminology is or if a single entity is enough here.

cpp/src/arrow/c/abi.h
kkraus14: Same conversation going on here? https://github.com/apache/arrow/pull/34972/files#r1161080258

cpp/src/arrow/c/abi.h
kkraus14: One of the nice things in this being a pointer is that `ArrowArray` can evolve without also breaking the `ArrowDeviceArray` abi.

If we're going to inline this maybe we should put it at the end of the `ArrowDeviceArray` struct? I'm not sure if there's magic alignment that compilers do that could lead to UB either way if the `ArrowArray` struct changes.

zeroshade: The `ArrowArray` struct itself can't change because that would break the ABI, but we might potentially create an `ArrowArrayV2` that would be good to easily be able to plug in here. 

That said, "who owns the memory" is a good question, and technically if we did evolve this to point to an `ArrowArrayV2` in theory that would still be breaking because while the ABI would still work just fine (all the sizes are the same) attempting to access the members would fail because the struct itself would be different. So my entire idea of using a pointer here is flawed in general. I'll update this to just embed the struct.

cpp/src/arrow/c/abi.h
kkraus14: CUDA (NVIDIA GPUs) and ROCm (AMD GPUs) both use `int` to specify a device ID in their APIs.

SYCL uses a `cl_device_id` which is an opaque type where I'm not sure if an `int32_t` or `int64_t` is sufficient to capture it. Maybe @aregm could have someone help guide here? 

dlpack has historically used an `int` / `int32_t` for 6+ years.

cpp/src/arrow/c/abi.h
zeroshade: what do we all think about using `queue` here then instead of `stream`? Is that enough to avoid the ambiguity?

cpp/src/arrow/c/abi.h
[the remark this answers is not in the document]
zeroshade: Ah, you're referring to the macros for the FLAGs. Fair enough, I'll switch these to macros so that the storage isn't compiler or language dependant.

cpp/src/arrow/c/abi.h
zeroshade: comment updated.

cpp/src/arrow/c/abi.h
zeroshade: comment formats updated to be docstrings in the same format as found in `bridge.h`

cpp/src/arrow/c/abi.h
zeroshade: I don't think so, it would be invalid for the device_type to be any value that is *not* one of the macro values, an `unset` should most likely be the CPU value. I can't think of a valid use case to have an `unset` value. 

If someone can think of a valid use case for an "unset" value, i'm happy to add one. But i'd rather not add one for the sake of having one if it would technically be invalid for it to ever be used.

cpp/src/arrow/c/abi.h
zeroshade: resolving this in favor of the other conversation on the same topic: https://github.com/apache/arrow/pull/34972/files#r1161080258

Please unresolve and comment if this should be continued separately here for some reason.

cpp/src/arrow/c/abi.h
```diff
@@ -106,6 +212,98 @@ struct ArrowArrayStream {
...
+  int (*get_schema)(struct ArrowDeviceArrayStream* self, struct ArrowSchema* out);
+
+  /// \brief Callback to get the device id for the next array.
+  ///
+  /// This is necessary so that the proper/correct stream pointer can be provided
+  /// to get_next.
+  ///
+  /// The next call to `get_next` should provide an ArrowDeviceArray whose
+  /// device_id matches what is provided here, and whose device_type is the
+  /// same as the device_type member of this stream.
```
westonpace, line 251: I'm not certain I follow.  Isn't the `ArrowDeviceArray` passed to `get_next` an "out" parameter?  Are you saying that the `ArrowDeviceArray` struct itself (not the buffers) needs to be allocated on the device?

cpp/src/arrow/c/abi.h
```diff
@@ -106,6 +212,98 @@ struct ArrowArrayStream {
...
+  /// Because different frameworks use different types to represent this, we
+  /// accept a void* which should then be reinterpreted into whatever the
+  /// appropriate type is (e.g. cudaStream_t) for use by the producer.
+  ///
+  /// \param[in] self The ArrowDeviceArrayStream object itself
+  /// \param[in] queue_ptr The appropriate queue, stream, or
+  /// equivalent object for the device that the data is allocated on
+  /// to indicate where the consumer wants the data to be accessible.
+  /// If queue_ptr is NULL then the default stream (e.g. CUDA stream 0)
+  /// should be used to ensure that the memory is accessible from any stream.
```
westonpace, line 273: I'm a little confused here.  It sounds like I need to call `get_next_device_id` to determine which queue to use and then I need to pass that queue on to the call to `get_next`.  But why?  Why isn't the producer managing the queues?

If the producer controls which device id gets used (`get_next_device_id` seems to suggest this) then why does the consumer need to give it the queue?  For example, if I were merging streams from two different devices it seems like I would do something like (apologies for the butchered pseudo-code)...

```
// Dummy class merging two infinite streams in an inefficient round-robin fashion
class MergedStream {

  int get_next(ArrowDeviceArray* out) {
    if (merged_arrays_.empty()) {
      ArrowDeviceArray arr;
      left_.get_next(&arr);
      merged_arrays_.push(arr);
      right_.get_next(&arr);
      merged_arrays_.push(arr);
    }
    *out = merged_arrays_.pop();
  }
};
```

westonpace (commented):
I have a few questions about how `ArrowDeviceArrayStream` works

cpp/src/arrow/c/abi.h
zeroshade: no, I was referring to the `device_id` member and `device_type` member that get populated in the `ArrowDeviceArray` that is returned from `get_next`

pitrou: This is a weird API choice, all because you want the consumer to pass its CUDA stream of choice...

zeroshade: ultimately this is a consequence of the fact that the existing frameworks and APIs don't provide any good way to manage the stream's lifetime easily which makes having the consumer pass the stream be the safest route to take.

I'm absolutely open to suggestions to make this better as long as the consumer is able to pass in the desired stream.

cpp/src/arrow/c/abi.h
zeroshade: @pitrou the issue is that there isn't a mechanism that you could call in the release callback (to my knowledge) to cleanly control the lifetime. (@kkraus14 correct me if I'm wrong and this isn't what you meant)

cpp/src/arrow/c/abi.h
zeroshade: Correct. This is why there isn't a `void*` in the `ArrowDeviceArray` struct itself.

cpp/src/arrow/c/abi.h
pitrou: That doesn't make sense, does it? How is the consumer supposed to manage the stream's lifetime if "there isn't a mechanism that you could call in the release callback to cleanly control the lifetime"?

cpp/src/arrow/c/abi.h
pitrou: > the existing frameworks and APIs don't provide any good way to manage the stream's lifetime easily

What do you mean by that? Would you care to give a more concrete example? For example CUDA allows you to destroy a stream:
https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__STREAM.html#group__CUDART__STREAM_1gfda584f1788ca983cb21c5f4d2033a62

kkraus14: Many frameworks manage their streams internally and don't expose those streams in their APIs, so there isn't a clean way to share or release ownership of a stream in a way that a stream can be guaranteed to not be destroyed until the release callback is called.

See the conversations here:
- https://github.com/dmlc/dlpack/issues/57
- https://github.com/data-apis/consortium-feedback/issues/1
- https://github.com/data-apis/array-api/pull/106

cpp/src/arrow/c/abi.h
kkraus14: The idea of a consumer providing a stream is that it only needs to manage the lifetime of its provided stream. A flow would look something like:

- Producer creates stream 3
- Producer runs some kernel that produces data safe to consume on stream 3
- Consumer creates stream 4
- Consumer calls Producer API asking for C Data Interface and provides it stream 4
- Producer creates an Event (see cudaEventCreate or hipEventCreate) and records it on stream 3
- Producer waits the passed in stream 4 on the event (see cudaStreamWaitEvent or hipStreamWaitEvent)
- Producer returns the C Data Interface struct to the consumer

With this, the producer's stream 3 can be destroyed at any point safely after the C Data Interface struct is returned so you don't need to worry about two different libraries / frameworks needing to be able to control the lifetimes of the others' streams.

cpp/src/arrow/c/abi.h
zeroshade: anyone have objections to using `int64_t` for future-proof reasons? Should I just use `int32_t` instead?

cpp/src/arrow/c/abi.h
pitrou: Ok, what's wrong with the following:
* Producer creates CUDA stream 3
* Producer runs some kernel that produces data safe to consume on CUDA stream 3
* Consumer gets a C Data Interface export of the previously produced data; the `ArrowArray`'s release callback is responsible for handling the lifetime of stream 3, in addition to the lifetime of the allocated data
* Consumer either syncs on stream 3 directly, or chooses to sync stream 3 to its own stream 4 using the event APIs
* When consumer is done, it calls the release callback, letting the producer dealing with CUDA stream 3 lifetime

cpp/src/arrow/c/abi.h
pitrou: +1 for `int64_t`. We don't know what future device APIs may choose here.

cpp/src/arrow/c/abi.h
pitrou: >     * [Specify synchronization semantics dmlc/dlpack#57](https://github.com/dmlc/dlpack/issues/57)

I tried to read through this and did not find a convincing argument that we shouldn't let the producer expose the original stream value. The entire discussion seemed to revolve around API convenience, not correctness. Also, people there seem to agree that the CUDA Array Interface from Numba is sane.

Did I miss something?

kkraus14: There's a lot of discussion in that issue related to internal stream handling under contexts and schedulers and similar terms. It all boils down to the same discussion of many producers being unable to release or share ownership of their streams.

I think this comment does a good job of summarizing the options that were considered: https://github.com/dmlc/dlpack/issues/57#issuecomment-753220425

And then this comment summarizes discussion of those options: https://github.com/dmlc/dlpack/issues/57#issuecomment-774482973


The lifetime management of streams as defined by the Numba documentation for `__cuda_array_interface__` (https://numba.readthedocs.io/en/stable/cuda/cuda_array_interface.html#streams) requires that keeping the object (typically array) that produces the `__cuda_array_interface__` alive also keeps the stream alive. In most cases libraries don't associate a stream with the object since it's valid to use multiple streams with a single object.

Here's the current state of things across a handful of projects:
- Numba
  - Handles stream lifetime properly for `__cuda_array_interface__` since they store a stream object as part of their array: https://github.com/numba/numba/blob/008077553b558bd183668ecd581d4d0bc54bd32c/numba/cuda/cudadrv/devicearray.py#L119-L139
  - Doesn't support `__dlpack__`
- CuPy
  - Doesn't implement stream lifetime properly for `__cuda_array_interface__` as far as I can tell, where they just get a stream ptr as an integer and if it's not the default stream someone could change the current stream ptr and end up having it be destroyed out from underneath it: https://github.com/cupy/cupy/blob/c92d5bc16293300297b843b4ebb364125697c131/cupy/_core/core.pyx#L258-L262 (cc @leofang)
  - Handles `__dlpack__` properly: https://github.com/cupy/cupy/blob/c92d5bc16293300297b843b4ebb364125697c131/cupy/_core/core.pyx#L285-L327
- PyTorch
  - Doesn't support passing the stream in `__cuda_array_interface__` currently: https://github.com/pytorch/pytorch/blob/def50d253401540cfdc6c0fffa444d0ee643cc11/torch/_tensor.py#L1005-L1066
  - Handles `__dlpack__` properly: https://github.com/pytorch/pytorch/blob/def50d253401540cfdc6c0fffa444d0ee643cc11/torch/_tensor.py#L1345
- Tensorflow and JAX
  - Doesn't support passing the stream in `__cuda_array_interface__` currently: https://github.com/tensorflow/tensorflow/blob/ea6a0f282d2b7ce20891dfc24ec8fe107eeaf22d/tensorflow/compiler/xla/python/py_buffer.cc#L254-L300
  - Doesn't support `__dlpack__`, but has explicit `to` and `from` functions for using dlpack pycapsules but does not handle streams: https://github.com/tensorflow/tensorflow/blob/6a050b6c15ed2a545693bc171f5e95dacbe05839/tensorflow/compiler/xla/python/dlpack.cc#L285-L363
- cuDF
  - Doesn't support passing the stream in `__cuda_array_interface__` currently: https://github.com/rapidsai/cudf/blob/50718e673ff53b18706cf66c6e02cda8e30681fe/python/cudf/cudf/core/column/numerical.py#L169-L191
  - Doesn't support `__dlpack__` but has explicit `to` and `from` functions for using dlpack pycapsules but does not handle streams: https://github.com/rapidsai/cudf/blob/50718e673ff53b18706cf66c6e02cda8e30681fe/cpp/src/interop/dlpack.cpp#L218-L294

kkraus14: cc @rgommers @leofang @tqchen (please feel free to include anyone else) do you happen to know if there was any other discussion captured that could be linked here regarding the decision to have a consumer hand a stream to the producer?

pitrou: Thanks the pointers.

> I think this comment does a good job of summarizing the options that were considered: [dmlc/dlpack#57 (comment)](https://github.com/dmlc/dlpack/issues/57#issuecomment-753220425)

Yes, I read this. It looks like solution S1, which is also the one I'm proposing, is considered the most flexible (I don't understand the "harder for compilers" comment, though).

> And then this comment summarizes discussion of those options: [dmlc/dlpack#57 (comment)](https://github.com/dmlc/dlpack/issues/57#issuecomment-774482973)

I read this too, but it doesn't actually mention S1, for reasons I wasn't able to understand.

>  In most cases libraries don't associate a stream with the object since it's valid to use multiple streams with a single object.

But you have to actually synchronize on the right stream before being able to use the object, right? How does the user know which stream to synchronize on, if they didn't produce the data themselves?

kkraus14: > Yes, I read this. It looks like solution S1, which is also the one I'm proposing, is considered the most flexible (I don't understand the "harder for compilers" comment, though).

From: https://github.com/dmlc/dlpack/issues/57#issuecomment-751715168
> It also brings extra burden to the compilers themselves. The compiler will need to generate optional synchronization code based on the streams, which is non-trivial.

> But you have to actually synchronize on the right stream before being able to use the object, right?

Something / someone needs to guarantee that there isn't a data race with regards to using multiple non-blocking streams, yes. That could be done with events, stream synchronization, or device synchronization.

> How does the user know which stream to synchronize on, if they didn't produce the data themselves?

If you're staying within your framework / library then the expectation is for the framework / library to handle things for the user. If crossing framework / library boundaries, then the expectation is to be reliant on things like interchange protocols to handle the synchronization semantics.

kkraus14:
@pitrou I think the other thing that hasn't been discussed in the above threads with regards to producer vs consumer provided stream is the development burden that will have to be incurred by the consumer in the different situations.

In producer provided stream / event:
- The default behavior would be for the producer to return a stream or event in the struct if appropriate based on the producer
- It forces the consumer to understand and handle device streams / events, which could be a non-trivial ask
- There's no way for the consumer to ask for the data to be synchronized and safe to access from any stream and for the stream or event to be nullptr

In consumer provided stream:
- The default behavior would be the consumer not providing a stream which would inform the producer to guarantee the data to be synchronized and safe to access from any stream
- Someone could use consumer code that doesn't understand or utilize device streams or events in any way without issue

pitrou:
Ok, who is supposed to be the consumer of the C Device Data Interface? I would expect it to be an Arrow implementation, or an Arrow-compatible library. Realistically they probably already deal with CUDA streams if they support CUDA?

cpp/src/arrow/c/abi.h
pitrou: Ok, so Tensorflow doesn't expose its CUDA streams to the user... but can the user request that Tensorflow data be tied to a specific CUDA stream? Otherwise, how can we implement the "consumer-provided CUDA stream" semantics?

kkraus14: > can the user request that Tensorflow data be tied to a specific CUDA stream?

No a user can't request that.

> Otherwise, how can we implement the "consumer-provided CUDA stream" semantics?

Tensorflow can handle the consumer provided stream internally as part of it's implementation in producing dlpack or Arrow C Data Interface structs.

kkraus14:
> Ok, who is supposed to be the consumer of the C Device Data Interface? I would expect it to be an Arrow implementation, or an Arrow-compatible library.

I agree it would be primarily other libraries who want to exchange Arrow data. Users would likely initiate the exchange between the two libraries and the C Device Data Interface would be an implementation detail of how that exchange would work.

I do think there would be some users who want to directly interact with the C Device Data Interface structs though as well, as I had seen similar with `__cuda_array_interface__` and dlpack.

> Realistically they probably already deal with CUDA streams if they support CUDA?

Not necessarily. There's a bunch of CUDA code out in the wild that just uses the default stream implicitly. For example, most numba and cupy cuda code I've seen doesn't use streams. Numba and CuPy internally both support using CUDA Streams but I don't think it would make sense for either of them to directly support the Arrow C Device Data Interface since it doesn't match their memory layout. So that would become a library author or user's responsibility to handle it and CUDA Streams / Events may be a new concept to them.

cpp/src/arrow/c/abi.h
pitrou: Why wouldn't they? They can easily refcount the usage of their own CUDA streams.

kkraus14: > Why wouldn't they? They can easily refcount the usage of their own CUDA streams.

I think that is making a lot of assumptions about how folks use and manage CUDA streams 😄. Again, some places use them similarly to thread pools and only control the lifetime of the pool.

I tried to dig through Tensorflow's code to figure exactly how they're managing the lifetime of their streams but I'm not confident, everything I say below may not be correct:

- Something eventually calls down to `AllocateStream` and `DeallocateStream` (https://github.com/tensorflow/tensorflow/blob/b9fc6a9b611ec373c02e5b5ab432b1d7aff9392e/tensorflow/compiler/xla/stream_executor/cuda/cuda_gpu_executor.cc#L759-L774) to create and destroy CUDA streams.
- These operate on raw ptrs and it looks like there's a class that wraps these, `Stream` (https://github.com/tensorflow/tensorflow/blob/b9fc6a9b611ec373c02e5b5ab432b1d7aff9392e/tensorflow/compiler/xla/stream_executor/stream.cc#L262-L286) which has constructor, destructor, and init functions roughly of what you'd expect.
- I believe these `Stream` objects are managed in a `StreamPool` (https://github.com/tensorflow/tensorflow/blob/b9fc6a9b611ec373c02e5b5ab432b1d7aff9392e/tensorflow/compiler/xla/service/stream_pool.h#L27-L59) which then allows "borrowing" the streams using unique ptrs.

I guess in theory that if they ultimately have `Stream` objects being used that it could be moved into the private data being used by the release callback.

cpp/src/arrow/c/abi.h
kurapov-peter: > CUDA (NVIDIA GPUs) and ROCm (AMD GPUs) both use `int` to specify a device ID in their APIs.
> 
> SYCL uses a `cl_device_id` which is an opaque type where I'm not sure if an `int32_t` or `int64_t` is sufficient to capture it. Maybe @aregm could have someone help guide here?
> 
> dlpack has historically used an `int` / `int32_t` for 6+ years.

DPC++ uses level zero (https://github.com/oneapi-src/level-zero/tree/master) as the primary C API. As to the device ID it is stored as `uint32_t` currently (see https://spec.oneapi.io/level-zero/latest/core/api.html#_CPPv422ze_device_properties_t) and it is not expected to exceed `int64_t`, so +1 as well.

aregm: Once remark, though - the assumption is that Arrow will not go beyond a single host, anyone who decided to go low-level knows how to map [UUID](https://spec.oneapi.io/level-zero/latest/core/api.html#ze-uuid-t) to host/device. But with the upcoming CXL capability, this can be a problem. Let's leave it int64 and address the issue when it comes. So +1.

zeroshade:
@kkraus14 @pitrou Did you two reach a consensus on the question concerning the consumer passing of the stream/queue pointer into `get_next` or is this still an open question that needs to be discussed further?

cpp/src/arrow/c/abi.h
kkraus14: Okay, regardless if we take the producer provided path then I think it makes a lot more sense for the producer to share an Event than a Stream.

An Event can be waited on via `cudaStreamWaitEvent` / `hipStreamWaitEvent` which does a device side wait which would have minimal overhead if it's the same stream or `cudaEventSynchronize` / `hipEventSynchronize` if blocking host code is desired.

kkraus14:
> @kkraus14 @pitrou Did you two reach a consensus on the question concerning the consumer passing of the stream/queue pointer into `get_next` or is this still an open question that needs to be discussed further?

I am still in favor of passing the stream/queue, but if @pitrou has a strong preference for returning it from the producer instead I would just ask that we use an Event instead of a Stream/Queue.

pitrou:
Thanks @gmarkall for the detailed insight! (also, hi :-))

> I also think it would also have been more difficult to have the consumer supply the stream for CAI, because it is just a property containing a dict

This is also how the C Data Interface works, though of course it's a C struct rather than a Python dict.
This is also why I would originally prefer the CUDA stream to be provider-supplied, because it can be part of that interface...

So, to sum it up, we have:
* in favour of the producer-supplied CUDA stream approach: allows to expose the CUDA stream in the C Data Interface (not only the C Stream Interface, which is more specialized - for example, it's sync not async)
* in favour of the consumer-supplied CUDA stream approach: allows the consumer to ignore synchronization issues entirely, especially by passing a default stream

Am I missing something in this bullet list?

zeroshade:
so, @gmarkall @pitrou @kkraus14 do we have a consensus on the best way forward here? You three definitely have a lot more insight into this than I do, so I'll move forward with whichever approach you all can agree on :smile:

kkraus14:
> so, @gmarkall @pitrou @kkraus14 do we have a consensus on the best way forward here? You three definitely have a lot more insight into this than I do, so I'll move forward with whichever approach you all can agree on 😄

I'm comfortable with either option assuming we use an `Event` pointer as opposed to a `Stream` pointer in the producer-supplied case, with a slight preference to the consumer-supplied case.

pitrou:
In that case, a bit more research is needed to find out what the equivalent of a CUDA `Event` would be for other backends.

cpp/src/arrow/c/abi.h
leofang: > do you happen to know if there was any other discussion captured that could be linked here regarding the decision to have a consumer hand a stream to the producer

Sorry I wasn't able to respond promptly. Is the question still open?

In the case of CAI, [it is required](https://numba.readthedocs.io/en/stable/cuda/cuda_array_interface.html#streams) that someone handles the exporting stream's lifetime properly:
> Like data, CUDA streams also have a finite lifetime. It is therefore required that a Producer exporting data on the interface with an associated stream ensures that the exported stream’s lifetime is equal to or surpasses the lifetime of the object from which the interface was exported.

and this was considered a burden when discussing the DLPack support. A few libraries like Numba, for example, had to [hold the reference to the underlying stream](https://numba.readthedocs.io/en/stable/cuda/cuda_array_interface.html#producing-arrays). I believe this was the main concern for DLPack to place the requirement on the consumer instead of the producer.

kkraus14:
cc @aregm @oleksandr-pavlyk

Any chance you could provide feedback from the SYCL perspective or ping someone who could?

zeroshade:
@pitrou Do we need an equivalent for `Event` or just use a `void*` and let other backends use whatever they would need appropriately?

@aregm @oleksandr-pavlyk just poking again to see if either of you can respond to @kkraus14's comment? Thanks!

Just to clarify: are we considering that we'd add the `Event` to the `ArrowDeviceArray` struct and remove the stream argument from the `get_next` ?

pitrou:
> @pitrou Do we need an equivalent for `Event` or just use a `void*` and let other backends use whatever they would need appropriately?

Probably a `int64_t`, so that the event can fit in the field regardless of the backend (some backends may use a pointer, but some may use some kind of opaque handle).

However, we should probably define what the field _means_ for each backend. Otherwise we'll have a half-spec that nobody can implement rigorously.

Also, I think this means the device array must have its own release callback, to properly handle the lifetime of said event.

oleksandr-pavlyk:
I assume we are talking about USM-based chunks. Dereferencing a USM pointer, generically, is only safe in kernels executed on the same device and using the same context that were used to perform the allocation. USM pointers of certain type (USM-shared and USM-host) can also be accessed from host. Deallocation of USM allocations requires a context.

SYCL provides a mechanism to recover the device where the allocation was made from the pointer and the context (see `sycl::get_pointer_device`, [sec. 4.8.4](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_unified_shared_memory_pointer_queries)). Therefore, a USM allocation should be thought of as a tuple of a pointer and the associated SYCL context. 

To share USM allocations across libraries (modules) in the same process space, the same context must be used. Two context instances constructed from the same arguments (either SYCL device, or device selector callable) are going to be different.

To facilitate sharing USM allocations Intel has proposed an [extension](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_default_context.asciidoc) and implemented it in oneAPI DPC++ to define a canonical default context for each SYCL platform.

This extension allows to use USM pointer and a SYCL device to describe USM allocations bound to the default platform context. 
The SYCL device in this tuple can be represented by an integer, which is the position of the device in the vector of all devices known to SYCL runtime, obtained using `sycl::device::get_devices()`. This position is guaranteed to be stable (same for different modules in a process, and the same between process runs).

Hence USM allocations (only if bound to the default platform context) can be shared using DLPack and represented in DLPack struct using a pointer and a integral device identifier. (DLPack support for USM allocations is implemented in [IntelPython/dpctl](https://github.com/IntelPython/dpctl/blob/master/dpctl/tensor/_dlpack.pyx)).

I would therefore not recommend deleting information to reconstruct context from `ArrowDeviceArray` unless a Arrow were to restrict itself to heterogeneous systems in which SYCL only sees a single platform (i.e. only CUDA devices, only Level-Zero devices, or only HIP devices).

To support sharing of USM allocations which may still be due to be accessed by scheduled, but as yet unfinished kernels, one can use SYCL events (one may need to support providing more than one events). Consumer would then gate kernels accessing shared allocations to avoid race conditions.

@tchen suggested in https://github.com/dmlc/dlpack/issues/57#issuecomment-753696812 that instead of using events, a consumer would give a process a stream (or `sycl::queue`) that it would use to process shared USM allocation and the produce would insert a barrier gated by the events from submitted tasks that may yet modified the shared allocation. 

Support for such barriers is yet to make it into the SYCL standard (see [this oneAPI extension](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_enqueue_barrier.asciidoc) which is implemented in oneAPI DPC++ compiler). Such mechanism is implemented in IntelPython/dpctl (see [here](https://github.com/IntelPython/dpctl/blob/master/dpctl/tensor/_usmarray.pyx#L934-L935)).

zeroshade:
@oleksandr-pavlyk Thanks for the context and information there! Given the proposals so far here and what you mentioned in your comment. Do you have a preference between the concept of a producer-supplied event vs a consumer provided stream (or `syscl::queue`)?

Also, you said:

> I would therefore not recommend deleting information to reconstruct context from `ArrowDeviceArray`....

What information were you referring to as deleting?

oleksandr-pavlyk:
I was referring to the comment as I understood it:

> Just to clarify: are we considering that we'd add the Event to the ArrowDeviceArray struct and remove the stream argument from the get_next ?

In SYCL, the stream corresponds to in-order `sycl::queue` which provides a way to retrieve the `sycl::context` associated with USM allocation. Now that I reread your comment, I see that I _misunderstood_ the comment. It is not stating that stream is going to be removed from the `ArrowDeviceArray` structure, but rather only from arguments of `get_next`.

Focusing on the synchronization question part only from now on. Event-based synchronization is _cheaper_ than stream-based one (since it requires submitting queue barriers), at least in the SYCL world. Additionally, stream-based synchronization is not even possible with current SYCL 2020 without leveraging `oneapi_queue_barrier` extension. This may be a weaker argument since the entire enterprise of sharing USM allocations without explicitly sharing underlying context requires an extensions too.

zeroshade:
@oleksandr-pavlyk currently there isn't a stream that is a member of the `ArrowDeviceArray` struct. I appreciate the information and context, but can we boil it down to a conclusion/recommendation?

There are two questions that I can think of based on the information you've provided that I'd like to get a concrete opinion/conclusion from  you on if at all possible:

1. Should we include a event/stream/queue/some combination of these in the `ArrowDeviceArray` struct directly? If so, should it be some combination of these objects or should it just be one?
2. Would it be preferable to have the Producer provide an event inside the `ArrowDeviceArray` struct when it produces the array (consumer has to perform synchronization), or should we have the consumer provide a queue/stream to the producer that it would like the memory accessible from (producer performs synchronization)?

If there's something I'm not mentioning please let me know. I just want to try to move this proposal forward in a way that makes the most sense and is future proof-ed, so it would help to clarify precisely what your conclusions are in regards to those two questions. Thanks much!

cpp/src/arrow/c/abi.h
zeroshade: Since it seems we're going to take the Producer providing an event path, there isn't really a need for the `get_next_device_id` callback anymore, correct? Or am I missing something?

kkraus14 (approved):
Latest changes LGTM, thanks @zeroshade!

zeroshade:
@pitrou @lidavidm @paleolimbot @oleksandr-pavlyk @westonpace @leofang @gmarkall

I've updated the header and comments based on the discussions here for the past few weeks. If possible I'd like to reach a consensus here before the end of the week at which point I'll update the arrow dev mailing list for more discussion and hopefully a vote.

cpp/src/arrow/c/abi.h
```diff
@@ -65,6 +85,111 @@ struct ArrowArray {
...
+  /// \brief Reserved bytes for future expansion.
+  ///
+  /// As non-CPU development expands we can update this struct
+  /// without ABI breaking changes. This also rounds out the
+  /// total size of this struct to be 128 bytes (power of 2)
+  /// on 64-bit systems. These bytes should be zero'd out after 
+  /// allocation in order to ensure safe evolution of the ABI in 
+  /// the future.
+  int64_t reserved[3];
+  int32_t reserved_addl;
```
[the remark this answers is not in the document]
zeroshade: Nice catch on the compiler padding I forgot about that, I confirmed locally that if i remove the additional `int32_t reserved_addl;` then the static assert is valid (though it wouldn't be valid on a 32-bit machine so I'm not going to actually add that assert to the code).

cpp/src/arrow/c/abi.h
```diff
@@ -65,6 +85,111 @@ struct ArrowArray {
...
+  /// the semantics of the id will be hardware dependant.
+  int64_t device_id;
+  /// \brief The type of device which can access this memory.
+  ArrowDeviceType device_type;
+  /// \brief An event-like object to synchronize on if needed.
+  ///
+  /// Many devices, like GPUs, are primarily asynchronous with
+  /// respect to CPU processing. As such in order to safely access
+  /// memory, it is often necessary to have an object to synchronize
+  /// processing on. Since different devices will use different types
```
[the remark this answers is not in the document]
lidavidm: I suppose, but I think that decreases the value of this if a producer can't know what type to use based on the device type alone - that means they have to write producer-specific code.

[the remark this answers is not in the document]
kkraus14: Beyond the `cuda` and `hip` paths which you referenced below I'm unclear on what other devices would pass here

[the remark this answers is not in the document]
lidavidm: We can leave it then if it's unclear (maybe just specify it for CUDA?)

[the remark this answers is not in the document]
zeroshade: currently about three lines below where this thread is (on line 172-173) I specify `cudaEvent_t` for CUDA and `hipEvent_t` for HIP. 

@kkraus14 Do OpenCL/Vulkan/Metal/ROCM/OneAPI have equivalent structures that we could reference? (or any other device in the device list of enums)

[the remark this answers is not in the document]
kkraus14: - OpenCL: http://people.cs.bris.ac.uk/~simonm/workshops/BSC_2013/opencl:course:bsc/Slides/OpenCL_events.pdf
- Vulkan: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkEvent.html
- Metal: https://developer.apple.com/documentation/metal/mtlevent
- ROCM is HIP
- OneAPI: There's SYCL events: https://developer.codeplay.com/products/computecpp/ce/2.11.0/api-reference/classcl_1_1sycl_1_1event but I'm not clear if that directly translates to OneAPI or not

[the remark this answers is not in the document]
zeroshade: sounds good, i'll update the comments with this list of types :)

gmarkall:
> @pitrou @lidavidm @paleolimbot @oleksandr-pavlyk @westonpace @leofang @gmarkall
> 
> I've updated the header and comments based on the discussions here for the past few weeks. If possible I'd like to reach a consensus here before the end of the week at which point I'll update the arrow dev mailing list for more discussion and hopefully a vote.

Many thanks for the ping - I've no further comments.

leofang:
cc: @shwina for vis

oleksandr-pavlyk (approved):
Looks good to me. Thank you.

gmarkall (approved):

cpp/src/arrow/c/abi.h
```diff
@@ -65,6 +85,110 @@ struct ArrowArray {
...
+/// \brief Vulkan buffer for next-gen graphics
+#define ARROW_DEVICE_VULKAN = 7
+/// \brief Metal for Apple GPU
+#define ARROW_DEVICE_METAL = 8
+/// \brief Verilog simulator buffer
+#define ARROW_DEVICE_VPI = 9
+/// \brief ROCm GPUs for AMD GPUs
+#define ARROW_DEVICE_ROCM = 10
+/// \brief Pinned ROCm CPU memory allocated by hipMallocHost
+#define ARROW_DEVICE_ROCMHOST = 11
```
JohnZed, line 123:
```suggestion
#define ARROW_DEVICE_ROCM_HOST = 11
```
underscore to be consistent with ARROW_DEVICE_CUDA_HOST

cpp/src/arrow/c/abi.h
```diff
@@ -65,6 +85,110 @@ struct ArrowArray {
...
+  /// memory, it is often necessary to have an object to synchronize
+  /// processing on. Since different devices will use different types
+  /// to specify this we use a void* that can be coerced into
+  /// whatever the device appropriate type is (e.g. cudaEvent_t for
+  /// CUDA and hipEvent_t for HIP).
+  ///
+  /// If synchronization is not needed this can be null. If this is
+  /// non-null, then it should be used to call the appropriate sync
+  /// method for the device (e.g. cudaStreamWaitEvent / hipStreamWaitEvent).
+  void* sync_event;
```
GregoryKimball, line 178: Would you please help me understand the `void* sync_event` pointer a bit better in the context of CUDA/C++? Is this a function pointer that is expected to be called with a `cudaStream_t` parameter provided by the application? Would there be a benefit from storing a `cudaStream_t` in `ArrowDeviceArray` (in the `reserved` bytes or elsewhere)?

kkraus14: In the case of CUDA, this would be a pointer to a `cudaEvent_t`. It would be the producer's responsibility to create the event and call `cudaEventRecord`  to ensure the relevant work in the stream(s) are captured in the event. The consumer can then call `cudaStreamWaitEvent` which is typically a device-side more efficient stream synchronization mechanism than `cudaStreamSynchronize` (if they need to wait for host code they can still use `cudaEventSynchronize` instead).

If both sides are on the same stream, then the `cudaStreamWaitEvent` call should have negligible overhead.

cpp/src/arrow/c/abi.h
```diff
@@ -106,6 +239,74 @@ struct ArrowArrayStream {
 
 #endif  // ARROW_C_STREAM_INTERFACE
 
+#ifndef ARROW_C_DEVICE_STREAM_INTERFACE
+#define ARROW_C_DEVICE_STREAM_INTERFACE
+
+/// \brief Equivalent to ArrowArrayStream, but for ArrowDeviceArrays.
```
pitrou, line 245: Same here: describe the struct and its members in the spec, not in the `.h`.

pitrou (changes requested):
I would require that the device interface is adequately described in the spec, not in the `.h` file.

The spec is currently free-standing (the C Data Interface can be implemented entirely by reading the spec, without looking at any external source code) and it would be nice to keep it like that.

For clarity I would make the C Device Data Interface spec a separate page in the format docs.

cpp/src/arrow/c/abi.h
```diff
@@ -106,6 +239,74 @@ struct ArrowArrayStream {
...
+/// device, if a producer wants data to be produced on multiple devices
+/// then multiple streams should be provided. One per device.
+struct ArrowDeviceArrayStream {
+  /// \brief The device that this stream produces data on.
+  ///
+  /// All ArrowDeviceArrays that are produced by this
+  /// stream should have the same device_type as set
+  /// here. The device_type needs to be provided here
+  /// so that consumers can provide the correct type
+  /// of queue_ptr when calling get_next.
```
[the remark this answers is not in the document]
zeroshade: good point, will update.

zeroshade:
@pitrou I've added a full prose spec writeup for the documentation. Please have a look at your earliest convenience

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,613 @@
...
+CUDA for NVIDIA GPUs. Since copying data from host to device and back is
+expensive, the ideal would be to be able to leave the data on the device
+for as long as possible, even when passing it between runtimes and
+libraries.
+
+The Arrow C Device data interface builds on the existing C data interface
+by adding a very small, stable set of C definitions to it. For non-C/C++
+languages and runtimes, translating the C definitions to corresponding
+C FFI declarations should be just as simple as with the current C data
+interface.
```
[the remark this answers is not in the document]
zeroshade: I do that, just later in the prose rather than right here. But I can add that extra bit here if you think it's necessary.

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,613 @@
...
+
+Rationale
+=========
+
+The current :ref:`C Data Interface <c-data-interface>`, and most
+implementations of it, make the assumption that all data buffers provided
+are CPU buffers. Since Apache Arrow is designed to be a universal in-memory
+format for representing tabular ("columnar") data, there will be the desire
+to leverage this data on Non-CPU hardware such as GPUs. One example of such
+a case is the `RAPIDS cuDF library` which uses the Arrow memory format with
```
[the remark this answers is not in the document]
zeroshade: It does link to rapids, i'll update it to use the `:ref:` syntax but the link is at the bottom of the file.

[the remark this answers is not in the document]
zeroshade: sorry, not `:ref:`, but adding the `_` to indicate it. I'm still more used to md than rst lol

docs/source/format/CDeviceDataInterface.rst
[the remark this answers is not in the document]
zeroshade: added a sentence here to explain. Let me know what you think

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,613 @@
...
+
+The stream source is not assumed to be thread-safe. Consumers wanting to
+call ``get_next`` from several threads should ensure those calls are 
+serialized.
+
+Updating this specification
+===========================
+
+Once this specification is supported in an official Arrow release, the C ABI
+is frozen. This means that the ``ArrowDeviceArray`` structure definition
```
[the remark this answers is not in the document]
zeroshade: Added a note here to clarify the language. Let me know your thoughts, thanks!

lidavidm (approved):

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,628 @@
...
+    * CUDA: ``cudaEvent_t*``
+    * ROCm: ``hipEvent_t*``
+    * OpenCL: ``cl_event*``
+    * Vulkan: ``VkEvent*``
+    * Metal: ``MTLEvent*``
+    * OneAPI: ``sycl::event*``
+
+    If an event is provided, then the producer MUST ensure that the event
+    is triggered/recorded at the end of the processing stream once the data
+    is considered available for use.
```
pitrou, line 291: This is not clear to me. What is the "processing stream"? Can we make this simpler? e.g.:
> If an event is provided, then the producer MUST ensure that the exported data is available on the device before the even becomes ready. The consumer SHOULD wait on the event before trying to access the exported data.

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,628 @@
...
+    different devices will use different types to specify this, we use a
+    void* which can be coerced into a pointer to whatever the device
+    appropriate type is.
+
+    If synchronization is not needed, this can be null. If this is non-null
+    then it MUST be used to call the appropriate sync method for the device
+    (e.g. ``cudaStreamWaitEvent`` or ``hipStreamWaitEvent``) before attempting
+    to access the memory in the buffers.
+
+    Expected types to coerce this ``void*`` to depending on the reported
```
pitrou, line 279: Perhaps detail this in a separate paragraph about event types?
For example:
```restructuredtext
Synchronization event types
---------------------------

The table below lists the event types supported for each the device type.
If no event type is supported ("N/A"), then the ``sync_event`` member
should always be null.

+--------------------------+----------------------+--------+
| Device type              | Actual event type    | Notes  |
+==========================+======================+========+
| ARROW_DEVICE_CPU         | N/A                  |        |
+--------------------------+----------------------+--------+
| ARROW_DEVICE_CUDA        | ``cudaEvent_t*``     |        |
+--------------------------+----------------------+--------+
| ...                      | ...                  |        |
+--------------------------+----------------------+--------+
```

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,628 @@
...
+
+.. c:member:: int64_t ArrowDeviceArray.reserved[3]
+
+    As non-CPU development expands, there may be a need to expand this
+    structure. In order to do so without potentially breaking ABI changes,
+    we reserve 24 bytes at the end of the object. This also has the added
+    benefit of bringing the total size of this structure to exactly 128
+    bytes (a power of 2) on 64-bit systems. These bytes should be zero'd
+    out after allocation in order to ensure safe evolution of the ABI in
+    the future.
```
pitrou, line 302: I'm still skeptical about this, but if we want to keep this, replace "should" with "MUST"?

pitrou: Also, to be clear, those should be zeroed by the producer (which is responsible for initializing the whole struct).

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,628 @@
...
+
+.. note::
+    Rather than store the shape / types of the data alongside the
```
pitrou, line 305: Not sure this note is useful?

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,628 @@
...
+To export the data itself, we transfer ownership to the consumer through
+the release callback. This example will use CUDA, but the equivalent calls
+could be used for any device:
+
+.. code-block:: c
+
+    static void release_int32_device_array(struct ArrowArray* array) {
+        assert(array->n_buffers == 2);
+        // destroy the event
+        cudaEvent_t* ev_ptr = reinterpret_cast<cudaEvent_t*>(array->private_data);
```
pitrou, line 385: This wouldn't work since `array` is a `struct ArrowArray*` here? Need to cast to `struct ArrowDeviceArray*` first.

(note the cast might technically be UB...)

pitrou: Also, this is sample C code, not C++, so should only use C-style casts (not `reinterpret_cast`).

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,628 @@
...
+        // free the buffers and the buffers array
+        status = cudaFree(array->buffers[1]);
+        assert(status == cudaSuccess);
+        free(array->buffers);
+
+        // mark released
+        array->release = NULL;
+    }
+
+    __host__ void export_int32_device_array(void* cudaAllocdPtr,
```
pitrou, line 399: If `__host__` required? If so, it should also be added to the release function above.

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,628 @@
...
+                .children = NULL,
+                .dictionary = NULL,
+                // bookeeping
+                .release = &release_int32_device_array,
+                .private_data = reinterpret_cast<void*>(ev_ptr),
+            },
+            .device_id = static_cast<int64_t>(device),
+            .device_type = ARROW_DEVICE_CUDA,
+            // pass the event pointer to the consumer
+            .sync_event = reinterpret_cast<void*>(ev_ptr),
```
pitrou, line 436: Use C-style cast here.

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,628 @@
...
+        };
+
+        // allocate list of buffers
+        array->array.buffers = (const void**)malloc(sizeof(void*) * array->array.n_buffers);
+        assert(array->array.buffers != NULL);
+        array->array.buffers[0] = NULL;
+        array->array.buffers[1] = cudaAllocdPtr;
+    }
+
+================
```
pitrou, line 446: Formatting nit: should match over/underlining length with title length, otherwise I think reST won't like you :-)

pitrou (commented):
Looks quite good to me! Some minor comments about the docs.

cpp/src/arrow/c/abi.h
```diff
@@ -65,6 +85,119 @@ struct ArrowArray {
...
+  /// the private_data and release callback of the arrow array
+  /// should contain any necessary information and structures
+  /// related to freeing the array according to the device it
+  /// is allocated on, rather than having a separate release
+  /// callback embedded here.
+  struct ArrowArray array;
+  /// \brief The device id to identify a specific device
+  /// if multiple of this type are on the system.
+  ///
+  /// the semantics of the id will be hardware dependant.
```
wgtmac, line 161:
```suggestion
  /// the semantics of the id will be hardware dependent.
```

wgtmac (commented):
Some cosmetic comments

docs/source/format/CDeviceDataInterface.rst
zeroshade: One of the reasons for the creation of things such as the `Cuda Array Interface` that numba uses was in order to pass the shape and type of the data buffer along with the pointers. Since that kind of interface is what we're targeting here and we aren't putting that information into the `ArrowDeviceArray` struct, I felt it was good to explicitly state that producers should use the `ArrowSchema` struct to pass this information. But I'm open to removing this note if people feel it is unnecessary.

docs/source/format/CDeviceDataInterface.rst
zeroshade: In this example, you can see that we store the event pointer in the `private_data` member of the `ArrowArray` member inside of `ArrowDeviceArray` and release should be called using `device_array->array.release(&device_array->array)`. That ensures that there's no UB.

pitrou: Oh, I misread. Thanks for the clarification!

docs/source/format/CDeviceDataInterface.rst
kkraus14: Given this code block is marked as `c` and there's no kernel code here we should definitely remove this as it's not needed (and likely wouldn't work unless using `nvcc` and would still likely warn).

cpp/src/arrow/c/abi.h
```diff
@@ -15,10 +15,30 @@
...
+/// without necessarily using the Arrow libraries or reinventing
+/// the wheel. Developers can choose between tight integration
+/// with the Arrow software project or minimal integration with
+/// the Arrow format only.
+
 #pragma once
 
 #include <stdint.h>
 
+/// \defgroup Arrow C-Data Interface
```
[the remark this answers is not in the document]
zeroshade: Makes sense, I've added links to the specs and reduced the detail of the comments.

docs/source/format/CDeviceDataInterface.rst
zeroshade: I've simplified and updated this, let me know if the version is good or if you think it should be further simplified.

docs/source/format/CDeviceDataInterface.rst
zeroshade: Good idea, i've updated with the table and it's own section.

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,628 @@
...
+    (still very low) possibility it might change slightly. Once it is
+    supported in an official Arrow release and the "experimental" tag is
+    removed from it, this section will apply and the ABI will be frozen.
+    
+    The reason for the "experimental" tag is because we don't know what we
+    don't know. While it was attempted to ensure this is generic enough to
+    work with a multitude of different frameworks, it's also possible that
+    something was missed. Once there is some usage of this and we are
+    confident there isn't any necessary modifications, the "experimental"
+    tag will be removed and the ABI frozen.
```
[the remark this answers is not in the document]
zeroshade: Is the new version better?

docs/source/format/CDeviceDataInterface.rst
pitrou: Ok, but do we still want the bullet list below?

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,688 @@
...
+.. c:member:: ArrowDeviceType ArrowDeviceArray.device_type
+
+    The type of the device which can access the buffers in the array.
+
+.. c:member:: void* ArrowDeviceArray.sync_event
+
+    Optional. An event-like object to synchronize on if needed.
+
+    Many devices, like GPUs, are primarily asynchronous with respect to
+    CPU processing. As such in order to safely access memory, it is often
```
pitrou, line 268:
```suggestion
    CPU processing. As such, in order to safely access device memory, it is often
```

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,688 @@
...
+    necessary to have an object to synchronize processing with. Since
+    different devices will use different types to specify this, we use a
+    void* which can be coerced into a pointer to whatever the device
```
pitrou, line 271:
```suggestion
    ``void*`` which can be coerced into a pointer to whatever the device
```

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,688 @@
...
+---------------
+
+The ``ArrowDeviceType`` typedef is used to indicate what type of device the
+provided memory buffers were allocated on. This, in conjunction with the
+``device_id``, should be sufficient to reference the correct data buffers.
+
+We then use macros to define values for different device types. The provided
+macro values are compatible with the widely used `dlpack`_ ``DLDeviceType``
+definition values, using the same value for each as the equivalent
+``kDL<type>`` enum from dlpack.h. The list will be kept in sync with those
```
pitrou, line 158:
```suggestion
``kDL<type>`` enum from ``dlpack.h``. The list will be kept in sync with those
```

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,688 @@
...
+
+    Indicates usage of a Verilog simulator buffer.
+
+.. c:macro:: ARROW_DEVICE_ROCM
+
+    An AMD device using the `ROCm`_ stack.
+
+.. c:macro:: ARROW_DEVICE_ROCM_HOST
+
+    CPU memory pinned and page-locked allocated ``hipMallocHost``.
```
pitrou, line 205:
```suggestion
    CPU memory that was pinned and page-locked by ROCm by using ``hipMallocHost``.
```

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,688 @@
...
+
+.. c:member:: int64_t ArrowDeviceArray.reserved[3]
+
+    As non-CPU development expands, there may be a need to expand this
+    structure. In order to do so without potentially breaking ABI changes,
+    we reserve 24 bytes at the end of the object. This also has the added
+    benefit of bringing the total size of this structure to exactly 128
+    bytes (a power of 2) on 64-bit systems. These bytes MUST be zero'd
+    out after initialization by the producer in order to ensure safe 
+    evolution of the ABI in the future.
```
pitrou, line 302: I would not mention the dubious "added benefit" of making the size a power of two.

docs/source/format/CDeviceDataInterface.rst
pitrou: Then you should clearly state the context of this note, i.e. a comparison with the Cuda Array Interface. Otherwise it's confusing.

Perhaps you also want to move this at the end into a "Interoperability with other interchange formats" paragraph?

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,688 @@
...
+Notes:
+
+* \(1) Currently unknown if framework has an event type to support.
+* \(2) Extension Device has producer defined semantics and thus if
+       synchronization is needed for an extension device, the producer
+       should document the type.
+
+
+Semantics
+=========
```
pitrou, line 360: Perhaps we should make clear somewhere that, amongst all the information exported by the C Device Data Interface, only the data buffers reside in device memory, all the rest is in CPU memory.

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,688 @@
...
+        status = cudaEventCreate(ev_ptr);
+        assert(status == cudaSuccess);
+
+        // record event on the stream, assuming that the passed in
+        // stream is where the work to produce the data will be processing.
+        status = cudaEventRecord(*ev_ptr, stream);
+        assert(status == cudaSuccess);
+
+        // initialize fields
+        *array = (struct ArrowDeviceArray) {
```
pitrou, line 475: Should `memset` `*array` before, or explicitly zero out the reserved area.

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,688 @@
...
+            .array = (struct ArrowArray) {
+                .length = length,
+                .null_count = 0,
+                .offset = 0,
+                .n_buffers = 2,
+                .n_children = 0,
+                .children = NULL,
+                .dictionary = NULL,
+                // bookeeping
```
pitrou, line 484:
```suggestion
                // bookkeeping
```

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,688 @@
...
+    }
+
+=======================
+Device Stream Interface
+=======================
+
+Like the :ref:`C stream interface <c-stream-interface>`, the C Device data
+interface also specifies a higher-level structure for easing communication
+of streaming data within a single process. Defining an ``ArrowDeviceArrayStream``
+structure.
```
pitrou, line 516: Is there a leftover sentence here?
```suggestion
of streaming data within a single process.
```

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,688 @@
...
+The ArrowDeviceArrayStream structure
+------------------------------------
+
+The ``ArrowDeviceArrayStream`` provides a device type that can access the
+resulting data along with the required callbacks to interact with a
+streaming source of Arrow arrays. It has the following fields:
+
+.. c:member:: ArrowDeviceType device_type
+
+    The device type that this stream produces data on. All
```
pitrou, line 571: Mandatory or optional?

docs/source/format/CDeviceDataInterface.rst
```diff
@@ -0,0 +1,688 @@
...
+    Data allocated on a Qualcomm Hexagon DSP device.
+
+The ArrowDeviceArray structure
+------------------------------
+
+The ``ArrowDeviceArray`` structure embeds the C data ``ArrowArray`` structure
+and adds additional information necessary for consumers to use the data. It
+has the following fields:
+
+.. c:member:: struct ArrowArray ArrowDeviceArray.array
```
pitrou, line 241: For each of these, define if it's mandatory or optional.

docs/source/format/CDeviceDataInterface.rst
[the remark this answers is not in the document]
pitrou: Seems good to me, thanks!

pitrou (commented):
Some minor comments, but LGTM overall.

docs/source/format/CDeviceDataInterface.rst
zeroshade: added at the top of the `Memory Management` section.

docs/source/format/CDeviceDataInterface.rst
zeroshade: added

docs/source/format/CDeviceDataInterface.rst
zeroshade: done

docs/source/format/CDeviceDataInterface.rst
zeroshade: added the new section

pitrou (approved):

PR #34984, opened 2023-04-08
GH-34565: [C++] Teach dataset_writer to accept custom filename functor

### LINKED ISSUE #34565: [C++] Teach dataset_writer to accept custom functor on base file name template

Hi, 
I want to report a feature request as nowadays dataset_writer only supports [basename_template with kIntegerToken](https://github.com/apache/arrow/blob/main/cpp/src/arrow/dataset/dataset_writer.cc#L427-L441). Say I have file 0.parquet, 1,parquet, ..., 10.parquet, 11.parquet and 12.parquet. It does not work with alphabetical sorter and downstream users must implement lexicographic sorter accordingly.  In my case, I need to touch quite a few codebases to support hive style partition parquet in my org.

I propose to add a new option which allows left padding with zeros. so the names will be 001,parquet, ...,010.parquet, 011.parquet and 012.parquet. Let me know thoughts and suggestions as I can help with the contribution.

### Component(s)

C++

HaochengLIU:
### Rationale for this change

Existing basename_template will only use a monotonically increasing int as new filenames. when there is needs for custom filenames(left padding, hash-code), downstream users must rename the files in a post-processing step.

A new functor is added to FileSystemDatasetWriteOptions which allows users to provide a custom name for dataset_writer.

### Are these changes tested?

Yes. Unit tests are added for normal and ill-formed lambdas.

Yes. It allows users to customize output file names.
* Closes: #34565

### DIFF
--- a/cpp/src/arrow/dataset/dataset_writer.cc
+++ b/cpp/src/arrow/dataset/dataset_writer.cc
@@ -21,6 +21,7 @@
 #include <memory>
 #include <mutex>
 #include <unordered_map>
+#include <unordered_set>
 
 #include "arrow/filesystem/path_util.h"
 #include "arrow/record_batch.h"
@@ -308,10 +309,21 @@ class DatasetWriterDirectoryQueue {
   }
 
   Result<std::string> GetNextFilename() {
-    auto basename = ::arrow::internal::Replace(write_options_.basename_template,
-                                               kIntegerToken, ToChars(file_counter_++));
+    std::optional<std::string> basename;
+    if (write_options_.basename_template_functor == nullptr) {
+      basename = ::arrow::internal::Replace(write_options_.basename_template,
+                                            kIntegerToken, ToChars(file_counter_++));
+    } else {
+      basename = ::arrow::internal::Replace(
+          write_options_.basename_template, kIntegerToken,
+          write_options_.basename_template_functor(file_counter_++));
+    }
     if (!basename) {
       return Status::Invalid("string interpolation of basename template failed");
+    }
+    if (!used_filenames_.insert(*basename).second) {
+      return Status::Invalid("filename ", *basename,
+                             " is already used before. Check basename_template_functor");
     }
     return fs::internal::ConcatAbstractPath(directory_, prefix_ + *basename);
   }
@@ -406,6 +418,7 @@ class DatasetWriterDirectoryQueue {
       latest_open_file_tasks_.reset();
       latest_open_file_ = nullptr;
     }
+    used_filenames_.clear();
     return Status::OK();
   }
 
@@ -418,6 +431,7 @@ class DatasetWriterDirectoryQueue {
   DatasetWriterState* writer_state_;
   Future<> init_future_;
   std::string current_filename_;
+  std::unordered_set<std::string> used_filenames_;
   DatasetWriterFileQueue* latest_open_file_ = nullptr;
   std::unique_ptr<util::ThrottledAsyncTaskScheduler> latest_open_file_tasks_;
   uint64_t rows_written_ = 0;
--- a/cpp/src/arrow/dataset/file_base.h
+++ b/cpp/src/arrow/dataset/file_base.h
@@ -404,6 +404,12 @@ struct ARROW_DS_EXPORT FileSystemDatasetWriteOptions {
   /// {i} will be replaced by an auto incremented integer.
   std::string basename_template;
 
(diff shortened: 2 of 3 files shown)


HaochengLIU:
@westonpace Mind taking a look when you have time? Ty

cpp/src/arrow/dataset/dataset_writer.cc
```diff
@@ -308,11 +309,24 @@ class DatasetWriterDirectoryQueue {
...
 
   Result<std::string> GetNextFilename() {
-    auto basename = ::arrow::internal::Replace(write_options_.basename_template,
-                                               kIntegerToken, ToChars(file_counter_++));
+    std::optional<std::string> basename;
+    if (write_options_.basename_template_functor == nullptr) {
+      basename = ::arrow::internal::Replace(write_options_.basename_template,
+                                            kIntegerToken, ToChars(file_counter_++));
+    } else {
+      basename = ::arrow::internal::Replace(
```
westonpace, line 317: If a user is providing a `basename_template_functor` do we need to use `basename_template` at all?  Can this just be...

```
basename = write_options_.basename_template_functor(file_counter_++);
```

I don't think it matters too much but I think it would be simpler to say "either you use the template OR you use a custom function".

cpp/src/arrow/dataset/dataset_writer.cc
```diff
@@ -308,11 +309,24 @@ class DatasetWriterDirectoryQueue {
...
+    }
     if (!basename) {
       return Status::Invalid("string interpolation of basename template failed");
     }
+    if (used_filenames_.find(*basename) != used_filenames_.end()) {
+      return Status::Invalid("filename ", *basename,
+                             " is already used before. Check basename_template_functor");
+    } else {
+      used_filenames_.insert(*basename);
+    }
```
westonpace, line 329: Small nit, not a big deal, but you could combine the `find` and `insert` calls...

```
if (!used_filenames_.insert(*basename).second) {
  return Status::Invalid(...);
}
```

cpp/src/arrow/dataset/file_base.h
```diff
@@ -404,6 +404,10 @@ struct ARROW_DS_EXPORT FileSystemDatasetWriteOptions {
   /// {i} will be replaced by an auto incremented integer.
   std::string basename_template;
 
+  /// A functor which will be applied on basename_template to replace auto incremented
+  /// integer only if provided.
```
westonpace, line 408:
```suggestion
  /// A functor which will be applied on an incremented counter.  The result will be inserted into
  /// the basename_template in place of {i}.
  ///
  /// This can be used, for example, to left-pad the file counter.
```

westonpace (changes requested):
A few small suggestions but this looks good.  Thank you for the contribution!

cpp/src/arrow/dataset/dataset_writer.cc
HaochengLIU: good point, updated

cpp/src/arrow/dataset/file_base.h
HaochengLIU: updated, your version is much easier to digest

raulcd:
> One quick question: Do you happen to know when Arrow 12.0.0 will be released? Would like to get this PR in so I could use at work

We have a feature freeze policy for our Releases, see: https://arrow.apache.org/docs/dev/developers/release.html#creating-a-release-candidate
The code freeze is scheduled to be performed today. I plan to do it in the next couple of hours.

HaochengLIU:
Thanks for the info Raul. @westonpace If possible can I catch the last train :) ?

HaochengLIU:
It seems weston is on west coast. @raulcd Can you guide if we could wait until he is online? Ty

raulcd:
> It seems weston is on west coast. @raulcd Can you guide if we could wait until he is online? Ty

Sure, I can wait slightly longer. I'll try to ping him if there's no answer but in order to follow the current schedule for the release I'll branch the 12.0.0 release today (CEST timezone).

HaochengLIU:
SGTM. Thanks Raul

cpp/src/arrow/dataset/dataset_writer.cc
westonpace: Ok.  I am happy keeping it how it is.  If a user wants to avoid basename_template they can just set it to `"{i}"`.

westonpace (approved):

PR #34986, opened 2023-04-08
GH-34657: [Go] Add ValueString(i int) string to array

### LINKED ISSUE #34657: [Go][Array] Expose UntypedValue(i int) interface

yevgenypats:

* Closes: apache/arrow#34657
* GitHub Issue: #34657

### DIFF
--- a/go/arrow/array/binary.go
+++ b/go/arrow/array/binary.go
@@ -58,7 +58,7 @@ func (a *Binary) Value(i int) []byte {
 }
 
 // ValueString returns the string at index i
-func (a *Binary) ValueString(i int) string {
+func (a *Binary) ValueStr(i int) string {
 	if a.IsNull(i) {
 		return NullValueStr
 	}
@@ -67,7 +67,7 @@ func (a *Binary) ValueString(i int) string {
 
 // ValueStr returns the string at index i without performing additional allocations.
 // The string is only valid for the lifetime of the Binary array.
-func (a *Binary) ValueStr(i int) string {
+func (a *Binary) ValueString(i int) string {
 	b := a.Value(i)
 	return *(*string)(unsafe.Pointer(&b))
 }
@@ -114,7 +114,7 @@ func (a *Binary) String() string {
 		case a.IsNull(i):
 			o.WriteString("(null)")
 		default:
-			fmt.Fprintf(o, "%q", a.ValueStr(i))
+			fmt.Fprintf(o, "%q", a.ValueString(i))
 		}
 	}
 	o.WriteString("]")
@@ -200,13 +200,13 @@ func (a *LargeBinary) Value(i int) []byte {
 	return a.valueBytes[a.valueOffsets[idx]:a.valueOffsets[idx+1]]
 }
 
-func (a *LargeBinary) ValueString(i int) string {
+func (a *LargeBinary) ValueStr(i int) string {
 	if a.IsNull(i) {
 		return NullValueStr
 	}
 	return base64.StdEncoding.EncodeToString(a.Value(i))
 }
-func (a *LargeBinary) ValueStr(i int) string {
+func (a *LargeBinary) ValueString(i int) string {
 	b := a.Value(i)
 	return *(*string)(unsafe.Pointer(&b))
 }
@@ -253,7 +253,7 @@ func (a *LargeBinary) String() string {
 		case a.IsNull(i):
 			o.WriteString("(null)")
 		default:
-			fmt.Fprintf(&o, "%q", a.ValueStr(i))
+			fmt.Fprintf(&o, "%q", a.ValueString(i))
 		}
 	}
 	o.WriteString("]")
--- a/go/arrow/array/binary.go
+++ b/go/arrow/array/binary.go
@@ -18,6 +18,7 @@ package array
 
 import (
 	"bytes"
+	"encoding/base64"
 	"fmt"
 	"strings"
 	"unsafe"
@@ -59,6 +60,9 @@ func (a *Binary) Value(i int) []byte {
 // ValueString returns the string at index i without performing additional allocations.
 // The string is only valid for the lifetime of the Binary array.
 func (a *Binary) ValueString(i int) string {
+	if a.IsNull(i) {
+		return "(null)"
+	}
 	b := a.Value(i)
 	return *(*string)(unsafe.Pointer(&b))
 }
@@ -192,8 +196,10 @@ func (a *LargeBinary) Value(i int) []byte {
 }
 
 func (a *LargeBinary) ValueString(i int) string {
-	b := a.Value(i)
-	return *(*string)(unsafe.Pointer(&b))
+	if a.IsNull(i) {
+		return NullValueStr
+	}
+	return base64.StdEncoding.EncodeToString(a.Value(i))
 }
 
 func (a *LargeBinary) ValueOffset(i int) int64 {
--- a/go/arrow/array/binary.go
+++ b/go/arrow/array/binary.go
@@ -21,6 +21,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"strings"
+	"unsafe"
 
 	"github.com/apache/arrow/go/v12/arrow"
 	"github.com/goccy/go-json"
@@ -62,6 +63,13 @@ func (a *Binary) ValueString(i int) string {
 		return NullValueStr
 	}
 	return base64.RawURLEncoding.EncodeToString(a.Value(i))
+}
+
+// ValueStr returns the string at index i without performing additional allocations.
+// The string is only valid for the lifetime of the Binary array.
+func (a *Binary) ValueStr(i int) string {
+	b := a.Value(i)
+	return *(*string)(unsafe.Pointer(&b))
 }
 
 func (a *Binary) ValueOffset(i int) int {
@@ -197,6 +205,10 @@ func (a *LargeBinary) ValueString(i int) string {
 		return NullValueStr
 	}
 	return base64.StdEncoding.EncodeToString(a.Value(i))
+}
+func (a *LargeBinary) ValueStr(i int) string {
+	b := a.Value(i)
+	return *(*string)(unsafe.Pointer(&b))
 }
 
 func (a *LargeBinary) ValueOffset(i int) int64 {
--- a/go/arrow/array/binary.go
+++ b/go/arrow/array/binary.go
@@ -200,13 +200,13 @@ func (a *LargeBinary) Value(i int) []byte {
 	return a.valueBytes[a.valueOffsets[idx]:a.valueOffsets[idx+1]]
 }
 
-func (a *LargeBinary) ValueString(i int) string {
+func (a *LargeBinary) ValueStr(i int) string {
 	if a.IsNull(i) {
 		return NullValueStr
 	}
 	return base64.StdEncoding.EncodeToString(a.Value(i))
 }
-func (a *LargeBinary) ValueStr(i int) string {
+func (a *LargeBinary) ValueString(i int) string {
 	b := a.Value(i)
 	return *(*string)(unsafe.Pointer(&b))
 }
--- a/go/arrow/array/binary.go
+++ b/go/arrow/array/binary.go
@@ -62,7 +62,7 @@ func (a *Binary) ValueString(i int) string {
 	if a.IsNull(i) {
 		return NullValueStr
 	}
-	return base64.RawURLEncoding.EncodeToString(a.Value(i))
+	return base64.StdEncoding.EncodeToString(a.Value(i))
 }
 
 // ValueStr returns the string at index i without performing additional allocations.
--- a/go/arrow/array/binary.go
+++ b/go/arrow/array/binary.go
@@ -21,7 +21,6 @@ import (
 	"encoding/base64"
 	"fmt"
 	"strings"
-	"unsafe"
 
 	"github.com/apache/arrow/go/v12/arrow"
 	"github.com/goccy/go-json"
@@ -57,14 +56,12 @@ func (a *Binary) Value(i int) []byte {
 	return a.valueBytes[a.valueOffsets[idx]:a.valueOffsets[idx+1]]
 }
 
-// ValueString returns the string at index i without performing additional allocations.
-// The string is only valid for the lifetime of the Binary array.
+// ValueString returns the string at index i
 func (a *Binary) ValueString(i int) string {
 	if a.IsNull(i) {
-		return "(null)"
+		return NullValueStr
 	}
-	b := a.Value(i)
-	return *(*string)(unsafe.Pointer(&b))
+	return base64.RawURLEncoding.EncodeToString(a.Value(i))
 }
 
 func (a *Binary) ValueOffset(i int) int {
--- a/go/arrow/array/binary.go
+++ b/go/arrow/array/binary.go
@@ -200,13 +200,13 @@ func (a *LargeBinary) Value(i int) []byte {
 	return a.valueBytes[a.valueOffsets[idx]:a.valueOffsets[idx+1]]
 }
 
-func (a *LargeBinary) ValueStr(i int) string {
+func (a *LargeBinary) ValueString(i int) string {
 	if a.IsNull(i) {
 		return NullValueStr
 	}
 	return base64.StdEncoding.EncodeToString(a.Value(i))
 }
-func (a *LargeBinary) ValueString(i int) string {
+func (a *LargeBinary) ValueStr(i int) string {
 	b := a.Value(i)
 	return *(*string)(unsafe.Pointer(&b))
 }
--- a/go/arrow/array/binary.go
+++ b/go/arrow/array/binary.go
@@ -114,7 +114,7 @@ func (a *Binary) String() string {
 		case a.IsNull(i):
 			o.WriteString("(null)")
 		default:
-			fmt.Fprintf(o, "%q", a.ValueString(i))
+			fmt.Fprintf(o, "%q", a.ValueStr(i))
 		}
 	}
 	o.WriteString("]")
@@ -253,7 +253,7 @@ func (a *LargeBinary) String() string {
 		case a.IsNull(i):
 			o.WriteString("(null)")
 		default:
-			fmt.Fprintf(&o, "%q", a.ValueString(i))
+			fmt.Fprintf(&o, "%q", a.ValueStr(i))
 		}
 	}
 	o.WriteString("]")
--- a/go/arrow/array/binarybuilder.go
+++ b/go/arrow/array/binarybuilder.go
@@ -125,19 +125,6 @@ func (b *BinaryBuilder) AppendNull() {
 	b.UnsafeAppendBoolToBitmap(false)
 }
 
-func (b *BinaryBuilder) AppendFromString(s string) error {
-	if s == NullValueStr {
-		b.AppendNull()
-		return nil
-	}
-	data, err := base64.StdEncoding.DecodeString(s)
-	if err != nil {
-		return err
-	}
-	b.Append(data)
-	return nil
-} 
-
 func (b *BinaryBuilder) AppendEmptyValue() {
 	b.Reserve(1)
 	b.appendNextOffset()
@@ -307,11 +294,18 @@ func (b *BinaryBuilder) AppendValueFromString(s string) error {
 		b.AppendNull()
 		return nil
 	}
-	decodedVal, err := base64.StdEncoding.DecodeString(s)
-	if err != nil {
-		return fmt.Errorf("could not decode base64 string: %w", err)
+	switch b.dtype.ID() {
+	case arrow.BINARY, arrow.LARGE_BINARY:
+		decodedVal, err := base64.StdEncoding.DecodeString(s)
+		if err != nil {
+			return fmt.Errorf("could not decode base64 string: %w", err)
+		}
+		b.Append(decodedVal)
+	case arrow.STRING, arrow.LARGE_STRING:
+		b.Append([]byte(s))
+	default:
+		return fmt.Errorf("cannot append string to type %s", b.dtype)
 	}
-	b.Append(decodedVal)
 	return nil
 }
 
--- a/go/arrow/array/binarybuilder.go
+++ b/go/arrow/array/binarybuilder.go
@@ -125,6 +125,19 @@ func (b *BinaryBuilder) AppendNull() {
 	b.UnsafeAppendBoolToBitmap(false)
 }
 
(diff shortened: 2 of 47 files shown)


go/arrow/array/binary.go
```diff
@@ -56,9 +57,17 @@ func (a *Binary) Value(i int) []byte {
...
 func (a *Binary) ValueString(i int) string {
+	if a.IsNull(i) {
+		return NullValueStr
+	}
+	return base64.StdEncoding.EncodeToString(a.Value(i))
+}
+
+// ValueStr returns the string at index i without performing additional allocations.
+// The string is only valid for the lifetime of the Binary array.
+func (a *Binary) ValueStr(i int) string {
```
zeroshade, line 70: Hmm i'd prefer to avoid the breaking change. Can we choose a different name for the function so we can avoid having to rename the `*Binary.ValueString` method?

yevgenypats: @zeroshade  I can do the other way around. `ValueStr` for the new function? WDYT?

zeroshade: That's fine for me. Alternately, now that we have `GetOneForMarshal` publicly available, in theory with some trickery it should be pretty easy to leverage that and some strong typing with `String` methods we might not even need this `ValueStr` method but instead could simply just be able to `fmt.Sprint(GetOneForMarshal(...))` to get text. what do you think?

yevgenypats: Interesting. I prefer though to make it a public function `ValueStr` and the reverse `AppendFromValueStr` so it's easy to test it. I did call `GetOneForMarshal` in some of the `ValueStr` implementation where it made sense but I think in any case better to expose a proper function.

yevgenypats: I've updated. Not sure why the tests now fail though in the CI

zeroshade: You missed a few spots for `arraymarshal`:
image

yevgenypats: Ok seems I was missing a rebase. fixed now.

go/arrow/array/binary.go
```diff
@@ -105,7 +114,7 @@ func (a *Binary) String() string {
 		case a.IsNull(i):
 			o.WriteString("(null)")
 		default:
-			fmt.Fprintf(o, "%q", a.ValueString(i))
+			fmt.Fprintf(o, "%q", a.ValueStr(i))
```
zeroshade, line 117: this entire switch can probably be replaced with just calling `a.ValueStr(i)` right? since it already does the `a.IsNull(i)` check

go/arrow/array/binarybuilder_test.go
```diff
@@ -38,6 +38,7 @@ func TestBinaryBuilder(t *testing.T) {
 			ab.AppendNull()
 		} else {
 			ab.Append(v)
+
```
zeroshade, line 41: unnecessary empty line

go/arrow/array/list.go
```diff
@@ -52,6 +52,13 @@ func NewListData(data arrow.ArrayData) *List {
 
 func (a *List) ListValues() arrow.Array { return a.values }
 
+func (a *List) ValueStr(i int) string {
+	if !a.IsValid(i) {
+		return NullValueStr
+	}
+	return fmt.Sprintf("%v", a.newListValue(i))
```
zeroshade, line 59: same comment as for FixedSizeList. needs to defer call release on the return from `newListValue` and please add a test that verifies no leaked memory

go/arrow/array/list.go
```diff
@@ -175,6 +182,12 @@ func NewLargeListData(data arrow.ArrayData) *LargeList {
 
 func (a *LargeList) ListValues() arrow.Array { return a.values }
 
+func (a *LargeList) ValueStr(i int) string {
+	if !a.IsValid(i) {
+		return NullValueStr
+	}
+	return fmt.Sprintf("%v", a.newListValue(i))
```
zeroshade, line 189: same comment as above

go/arrow/array/list.go
```diff
@@ -416,6 +429,9 @@ func (b *baseListBuilder) AppendEmptyValue() {
 	b.Append(true)
 }
 
+// func (b *ListBuilder) AppendFromString(s string) {
+// 	b.AppendFromStringArray([]string{s})
+// }
```
zeroshade, line 434: remove commented code?

go/arrow/array/map.go
```diff
@@ -300,6 +300,10 @@ func (b *MapBuilder) ValueBuilder() Builder {
 	return b.listBuilder.ValueBuilder()
 }
 
+func (b *MapBuilder) AppendValueFromString(s string) error {
+	panic("not implemented")
```
zeroshade, line 304: can we just return `arrow.ErrNotImplemented` instead of panicing?

go/arrow/array/numeric.gen.go.tmpl
```diff
@@ -82,6 +82,32 @@ func (a *{{.Name}}) setData(data *Data) {
 	}
 }
 
+func (a *{{.Name}}) ValueStr(i int) string {
+	if a.IsNull(i) {
+		return NullValueStr
+	}
+{{if or (eq .Name "Date32") (eq .Name "Date64") -}}
+	return a.values[i].ToTime().Format("2006-01-02")
```
zeroshade, line 90: these types have a `FormattedString()` method that you can call instead. just do `a.values[i].FormattedString()`

go/arrow/array/numeric.gen.go.tmpl
```diff
@@ -82,6 +82,32 @@ func (a *{{.Name}}) setData(data *Data) {
...
+{{else if or (eq .Name "Time32") (eq .Name "Time64") -}}
+	return a.values[i].ToTime(a.DataType().(*{{.QualifiedType}}Type).Unit).Format("15:04:05.999999999")
```
zeroshade, line 92: `Time32` and `Time64` also have `FormattedString` methods which take the unit. so this should be:

`return a.values[i].FormattedString(a.DataType().(*{{.QualifiedType}}Type).Unit)`

go/arrow/array/numeric.gen.go.tmpl
```diff
@@ -82,6 +82,32 @@ func (a *{{.Name}}) setData(data *Data) {
...
+{{else if or (eq .Name "Timestamp") -}}
+	return a.values[i].ToTime(a.DataType().(*{{.QualifiedType}}Type).Unit).Format("2006-01-02 15:04:05.999999999")
+{{else if (eq .Name "Duration") -}}	
+	// return value and suffix as a string such as "12345ms"
+	return fmt.Sprintf("%d%s", a.values[i], a.DataType().(*{{.QualifiedType}}Type).Unit.String())
```
zeroshade, line 97: you don't need the `.String()` since you're already using `%s` in the format

go/arrow/array/numeric.gen.go.tmpl
```diff
@@ -82,6 +82,32 @@ func (a *{{.Name}}) setData(data *Data) {
...
+	return a.values[i].ToTime(a.DataType().(*{{.QualifiedType}}Type).Unit).Format("15:04:05.999999999")
+{{else if or (eq .Name "Timestamp") -}}
+	return a.values[i].ToTime(a.DataType().(*{{.QualifiedType}}Type).Unit).Format("2006-01-02 15:04:05.999999999")
+{{else if (eq .Name "Duration") -}}	
+	// return value and suffix as a string such as "12345ms"
+	return fmt.Sprintf("%d%s", a.values[i], a.DataType().(*{{.QualifiedType}}Type).Unit.String())	
+{{else if or (eq .Name "Int8") (eq .Name "Int16") (eq .Name "Int32") (eq .Name "Int64") -}}
+  return strconv.FormatInt(int64(a.Value(i)), 10)
+{{else if or (eq .Name "Uint8") (eq .Name "Uint16") (eq .Name "Uint32") (eq .Name "Uint64") -}}
+  return strconv.FormatUint(uint64(a.Value(i)), 10)
```
zeroshade, line 101: any reason not to just use `fmt.Sprintf("%d", a.values[i])`

go/arrow/array/numericbuilder.gen.go.tmpl
```diff
@@ -184,6 +184,124 @@ func (b *{{.Name}}Builder) newData() (data *Data) {
...
+		b.AppendNull()
+		return nil
+	}
+  {{if or (eq .Name "Date32") -}}
+  	tm, err := time.Parse("2006-01-02", s)
+    if err != nil {
+      b.AppendNull()
+      return err
+    }
+    b.Append(arrow.Date32FromTime(tm))
```
zeroshade, line 198: fix the indentation formatting please

go/arrow/array/numericbuilder.gen.go.tmpl
```diff
@@ -184,6 +184,124 @@ func (b *{{.Name}}Builder) newData() (data *Data) {
...
+    b.Append(val)
+  {{else if or (eq .Name "Timestamp") -}}
+  	v, err := arrow.TimestampFromString(s, b.dtype.Unit)
+    if err != nil {
+      b.AppendNull()
+      return err
+    }
+    b.Append(v)
+  {{else if (eq .Name "Duration") -}}	
+    return fmt.Errorf("AppendValueFromString not implemented for Duration")
```
zeroshade, line 228: please prefix the error with `%w` and use `arrow.ErrNotImplemented` so that we consistently wrap that error for this scenario.

go/arrow/array/numericbuilder.gen.go.tmpl
```diff
@@ -184,6 +184,124 @@ func (b *{{.Name}}Builder) newData() (data *Data) {
...
+  {{else if or (eq .Name "Int8") -}}
+    v, err := strconv.ParseInt(s, 10, 8)
+    if err != nil {
+      b.AppendNull()
+      return err
+    }
+    b.Append(int8(v))
```
zeroshade, line 235: You can condense all of these to:

```
v, err := strconv.ParseInt(s, 10, {{.Size}})
if err != nil {
    b.AppendNull()
    return err
}
b.Append({{.name}}(v))
```

instead of repeating it

go/arrow/array/numericbuilder.gen.go.tmpl
```diff
@@ -184,6 +184,124 @@ func (b *{{.Name}}Builder) newData() (data *Data) {
...
+      return err
+    }
+    b.Append(float32(v))
+  {{else if or (eq .Name "Float64") -}}
+    v, err := strconv.ParseFloat(s, 64)
+    if err != nil {
+      b.AppendNull()
+      return err
+    }
+    b.Append(float64(v))
```
zeroshade, line 298: same thing as with the integral types, you can condense these by using `{{.Size}}` and `{{.name}}`

go/arrow/array/numericbuilder.gen.go.tmpl
```diff
@@ -184,6 +184,124 @@ func (b *{{.Name}}Builder) newData() (data *Data) {
...
+  {{else}}
+    return fmt.Errorf("AppendValueFromString not implemented for {{.Name}}")
```
zeroshade, line 300: same comment as above, use `%w` and wrap the `arrow.ErrNotImplemented`

go/arrow/array/numericbuilder.gen_test.go.tmpl
```diff
@@ -51,9 +51,10 @@ func TestNew{{.Name}}Builder(t *testing.T) {
 	ab.Append(8)
 	ab.Append(9)
 	ab.Append(10)
+  ab.AppendValueFromString(11)
```
zeroshade, line 54: indentation

go/arrow/array/string.go
```diff
@@ -188,6 +195,7 @@ func (a *LargeString) Value(i int) string {
 	i = i + a.array.data.offset
 	return a.values[a.offsets[i]:a.offsets[i+1]]
 }
+func (a *LargeString) ValueStr(i int) string { return fmt.Sprintf("%q", a.Value(i)) }
```
zeroshade, line 198: why not just return `a.Value(i)` ?

go/arrow/array/struct.go
```diff
@@ -81,6 +81,15 @@ func NewStructData(data arrow.ArrayData) *Struct {
 func (a *Struct) NumField() int           { return len(a.fields) }
 func (a *Struct) Field(i int) arrow.Array { return a.fields[i] }
 
+func (a *Struct) ValueStr(i int) string {
```
zeroshade, line 84: add a docstring comment on here to indicate that the value is returned as a JSON string please.

go/arrow/array/struct.go
```diff
@@ -81,6 +81,15 @@ func NewStructData(data arrow.ArrayData) *Struct {
...
+	var buf bytes.Buffer
+	enc := json.NewEncoder(&buf)
+	if err := enc.Encode(a.GetOneForMarshal(i)); err != nil {
+		return ""
```
zeroshade, line 88: we shouldn't swallow the error. either return `err.String()` or `panic(err)` please

go/arrow/array/struct.go
```diff
@@ -351,6 +360,14 @@ func (b *StructBuilder) newData() (data *Data) {
 	return
 }
 
+func (b *StructBuilder) AppendValueFromString(s string) error {
+	if !strings.HasPrefix(s, "{") && !strings.HasSuffix(s, "}") {
+		return fmt.Errorf("invalid string for struct")
```
zeroshade, line 365: add `%w` and wrap `arrow.ErrInvalid` please

go/arrow/array/union.go
```diff
@@ -343,6 +343,21 @@ func (a *SparseUnion) MarshalJSON() ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
+func (a *SparseUnion) Value(i int) string {
+	fieldList := a.unionType.Fields()
+	field := fieldList[a.ChildID(i)]
+	f := a.Field(a.ChildID(i))
+	return fmt.Sprintf("{%s=%v}", field.Name, f.GetOneForMarshal(i))
+}
```
zeroshade, line 351: I think this is unnecesary. having the `ValueStr` is fine

zeroshade (changes requested):

go/arrow/array/binary.go
```diff
@@ -192,6 +201,12 @@ func (a *LargeBinary) Value(i int) []byte {
 }
 
 func (a *LargeBinary) ValueString(i int) string {
+	if a.IsNull(i) {
+		return NullValueStr
+	}
+	return base64.StdEncoding.EncodeToString(a.Value(i))
```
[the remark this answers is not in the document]
yevgenypats: Yeah, seems I was missing one commit from the previous review. fixed it.

go/arrow/array/binary.go
yevgenypats: I just kept it the same. as it was. also returned it to be `ValueString` as it doesn't need any change

go/arrow/array/binary.go
```diff
@@ -56,9 +57,17 @@ func (a *Binary) Value(i int) []byte {
...
 }
 
-// ValueString returns the string at index i without performing additional allocations.
-// The string is only valid for the lifetime of the Binary array.
+// ValueString returns the string at index i
 func (a *Binary) ValueString(i int) string {
+	if a.IsNull(i) {
+		return NullValueStr
+	}
+	return base64.StdEncoding.EncodeToString(a.Value(i))
```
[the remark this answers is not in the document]
yevgenypats: Yeah good catch. fixed.

go/arrow/array/binarybuilder_test.go
```diff
@@ -50,11 +51,16 @@ func TestBinaryBuilder(t *testing.T) {
...
+	// Zm9v is foo in base64
+	assert.NoError(t, ab.AppendValueFromString("Zm9v"))
 
 	ar := ab.NewBinaryArray()
+	assert.Equal(t, "Zm9v", ar.ValueString(5))
+
 	ab.Release()
 	ar.Release()
 
+
```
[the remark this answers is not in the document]
yevgenypats: Sure. BTW - maybe we should open an issue on adding a linter that runs fmt on all files as a pre-set ? this way it wont need review as I'll be able to just see in the CI. WDYT ? :)

go/arrow/array/encoded.go
```diff
@@ -192,6 +193,15 @@ func (r *RunEndEncoded) GetPhysicalLength() int {
 	return encoded.GetPhysicalLength(r.data)
 }
 
+func (r *RunEndEncoded) ValueStr(i int) string {
```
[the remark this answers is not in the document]
yevgenypats: hmm. maybe I should just use then `json.Marshal(r.GetOneForMarshal(i))` for consistency in this case?

go/arrow/array/binarybuilder_test.go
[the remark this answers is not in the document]
zeroshade: I think that it would definitely be a good idea to add that to the existing linting. Right now the linter just runs `staticcheck`, but doesn't check the formatting.

go/arrow/array/encoded.go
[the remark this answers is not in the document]
zeroshade: I think that's actually the difference here between `GetOneForMarshal` and `ValueStr`. `GetOneForMarshal` would return the whole run-mapping etc. While I think `ValueStr` should just return the actual "value" for the given "logical" index, so that a consumer doesn't need to do that themselves if they are actually trying to retrieve the values from the run-end-encoded array. It also maintains consistency with the Dictionary encoded arrays, where `ValueStr` returns the actual dictionary value, not the dictionary index. Does that make sense?

We should also explicitly document this in a doc string on the function.

go/arrow/array/binarybuilder.go
```diff
@@ -289,6 +289,26 @@ func (b *BinaryBuilder) appendNextOffset() {
...
+	if s == NullValueStr {
+		b.AppendNull()
+		return nil
+	}
+	switch b.dtype.ID() {
+	case arrow.BINARY, arrow.LARGE_BINARY:
+		decodedVal, err := base64.StdEncoding.DecodeString(s)
+		if err != nil {
+			return fmt.Errorf("could not decode base64 string: %w", err)
+		}
```
zeroshade, line 302: do we want to try falling back to `RawStdEncoding` if this errors so we can handle both situations of padding or no padding?

go/arrow/array/builder.go
```diff
@@ -82,6 +85,7 @@ type Builder interface {
 	UnmarshalOne(*json.Decoder) error
 	Unmarshal(*json.Decoder) error
 
+
```
zeroshade, line 88: extraneous line

zeroshade:
@yevgenypats I've resolved the outdated issues you've fixed but there's still outstanding changes that need to be made such as https://github.com/apache/arrow/pull/34986#discussion_r1162987868 and the other related comments.

go/arrow/array/binarybuilder.go
yevgenypats: hmm. I think we should try to always use `std` otherwise this fallback will propagate everywhere because we want catch if someone encoded something with `RawStdEncoding` instead of `StdEncoding`.

zeroshade: that's fair. I suppose if we run into an issue for individuals that are using non-padded base64 for this they'll file an issue and it's fairly easy to add the support. so this is fine.

go/arrow/array/numeric.gen.go.tmpl
yevgenypats: not sure. I think it's more efficient and we also do the same in CSV reader/writer but I don't mind changing it.

go/arrow/array/numericbuilder.gen.go.tmpl
yevgenypats: The problem with the indentation here is because in this file we use tabs instead of spaces. should we all turn it into spaces or should I keep using tabs in this file while in all `.go` files we do use spaces I believe. wdyt?

go/arrow/array/numeric.gen.go.tmpl
zeroshade: i just like condensing this where possible :smile:

go/arrow/array/numericbuilder.gen.go.tmpl
zeroshade: I'm fine with converting the entire file to spaces. Go for it

yevgenypats:
Thanks @zeroshade ! I addressed other comments. I only have one outstanding question for the template indentation and yeah if your or someone would be able to set the CI step that that does that automatically would be awesome as Im not super familiar with the arrow CI (so someone knowledgable would be able to do that faster I think - Opened issue - https://github.com/apache/arrow/issues/35079 )

go/arrow/array/numericbuilder.gen.go.tmpl
```diff
@@ -225,79 +225,30 @@ func (b *{{.Name}}Builder) AppendValueFromString(s string) error {
...
-      return err
+        b.AppendNull()
+        return err
     }
-    b.Append(uint64(v))
-  {{else if or (eq .Name "Float32") -}}
-    v, err := strconv.ParseFloat(s, 32)
+    b.Append({{.name}}(v))
+  {{else if or (eq .Name "Uint8") (eq .Name "Uint16") (eq .Name "Uint32") (eq .Name "Uint64") -}}
+    v, err := strconv.ParseInt(s, 10, {{.Size}})
```
zeroshade, line 237: `ParseUint` ?

zeroshade (changes requested):

go/arrow/array/struct_test.go
```diff
@@ -290,15 +291,17 @@ func TestStructArrayStringer(t *testing.T) {
...
 	defer arr.Release()
 
-	want := "{[1.1 (null) 1.3 1.4] [1 2 (null) 4]}"
+	assert.Equal(t, "{\"f1\":1.1,\"f2\":1}\n", arr.ValueStr(4))
+	want := "{[1.1 (null) 1.3 1.4 1.1] [1 2 (null) 4 1]}"
 	got := arr.String()
 	if got != want {
 		t.Fatalf("invalid string representation:\ngot = %q\nwant= %q", got, want)
 	}
+	// assert.Equal(t, "[1.1 (null) 1.3 1.4]", arr.ValueStr(0))
```
zeroshade, line 304: why is this commented out?

zeroshade (approved):

PR #34991, opened 2023-04-09
GH-34987: [PYTHON]  Truthy/Falsy for boolean scalars.

### LINKED ISSUE #34987: `bool(pa.scalar(False))` evaluates to `True`

This can easily cause bugs: `assert pa.scalar(False)` will not raise `AssertionError`!

### Component(s)

Python

zfoobar:
added __bool__ implementation to provide Truthy/Falsey for boolean scalars in pyarrow they were tested locally, waiting for ci
the user will be exposed to  truthy/falsey for True/False inputs respectively

* Closes: #34987

### DIFF
--- a/ruby/red-arrow/lib/arrow/slicer.rb
+++ b/ruby/red-arrow/lib/arrow/slicer.rb
@@ -163,8 +163,37 @@ module Arrow
         RejectCondition.new(@column, block)
       end
 
-      def match_substring(substring, ignore_case: false)
-        MatchSubstringFamilyCondition.new("match_substring",
+      def end_with?(substring, ignore_case: false)
+        MatchSubstringFamilyCondition.new("ends_with",
+                                          @column, substring, ignore_case)
+      end
+
+      def match_like?(pattern, ignore_case: false)
+        MatchSubstringFamilyCondition.new("match_like",
+                                          @column, pattern, ignore_case)
+      end
+
+      def match_substring?(pattern, ignore_case: nil)
+        case pattern
+        when String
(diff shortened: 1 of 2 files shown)


randolf-scholz:
I think it should just be `return self.as_py()`, no?

randolf-scholz:
Following my https://github.com/apache/arrow/issues/34987#issuecomment-1501083185, I think it should be:

```python
def __bool__(self) -> bool:
    if self.is_valid:
        return self.as_py()
    raise TypeError("Boolean value of null is ambiguous.")
```

zfoobar:
Good move. I have to see why it’s failing the macos runner as wellSent from my iPhoneOn Apr 9, 2023, at 2:40 AM, Randolf Scholz ***@***.***> wrote:﻿
Following my #34987 (comment), I think it should be:
def __bool__(self) -> bool:
    if self.is_valid:
        return self.as_py()
    raise TypeError("Boolean value of null is ambiguous.")

randolf-scholz:
Possible Unit-Test:

```python
import pyarrow as pa
import pytest

def test_BooleanScalar__bool__():
    t  = pa.scalar(True, type=pa.bool_())
    f  = pa.scalar(False, type=pa.bool_())
    m  = pa.scalar(None, type=pa.bool_())
    assert bool(t) is True
    assert bool(f) is False
    with pytest.raises(TypeError):
        bool(m)
```

zfoobar:
Got some great insight from Randolf. I wrote up a unit test for this, currently testing in fork.

PR #34993, opened 2023-04-09
GH-34987: [Python] Truthy/Falsy for boolean scalars

### LINKED ISSUE #34987: `bool(pa.scalar(False))` evaluates to `True`

zfoobar:
Added exception handling for null value case (thanks Randolf) Added unit test

* Closes: #34987


zfoobar:
@AlenkaF I noticed the MacOS12 AMD64 runner failed for this upstream - I can't reproduce. Test does not time out on my side. MacOS 13/M1, test runs in a few seconds.

AlenkaF:
> @AlenkaF I noticed the MacOS12 AMD64 runner failed for this upstream - I can't reproduce. Test does not time out on my side. MacOS 13/M1, test runs in a few seconds.

Hm, I see both MacOS and AppVeyor have the same issue and I can see the same time out on other PRs also - it is not connected to this PR.

AlenkaF:
> @AlenkaF it looks like the jury is still out on this PR. Do you want me to close it or expand the discussion to include _bool support for the other data types?

Oh, so sorry for being unresponsive! Will look at the discussion on the issue and see what would be the best thing to do. If there is still no activity from us by the end of this week, please ping me again 🙏

danepitkin:
This implementation LGTM! 

I think supporting `__bool__` for Boolean Scalars makes sense, and I agree we raise exception on NULL since its ambiguous. I would forego any implementation of `__bool__` with other scalars for the time being. There are bigger philosophical questions for other scalars such as how to handle 3-value logic.

pitrou:
Having `__bool__` return something else than a boolean, or raise an exception, is generally rather nasty as it breaks a widespread expectation about boolean testing.

Also the boolean value of `None` is well-defined:
```pycon
>>> bool(None)
False
```

randolf-scholz:
@pitrou It's not quite as easy...

- `bool(float("nan"))` is `True`. Historically, 𝙽𝚊𝙽-floats were often used as missing value indicators.
- `bool(None)` is `False`.
- `bool(NotImplemented)` is `True`, but gives a `DeprecationWarning`.
- `bool(pandas.NA)` raises `TypeError`.
- `pyarrow.NA.as_py()` returns `None`.

The last point would be a reason to return None. However, one should consider how and when `bool` evaluation might pop up.

- assert statements.
- branching logic.

Coercing a missing boolean scalar to `False` instead of raising an error can potentially lead to some very nasty and hard to debug issues. I'd wager that in the vast majority of cases, branching logic based on a missing boolean is just nonsense and should be dismissed.

There is also some inherent inconsistency with conversion to python if one takes this route.
In a perfect world, one would expect the following diagram to commute:

    ┌─────────┐         ┌─────────┐
    │pa Scalar├───op───►│pa Scalar│
    └────┬────┘         └────┬────┘
         │                   │
         │                   │
       as_py               as_py
         │                   │
         ▼                   ▼
    ┌─────────┐         ┌─────────┐
    │py Scalar├───op───►│py Scalar│
    └─────────┘         └─────────┘

However, consider this example:


```python
import pyarrow as pa

pa_x = pa.scalar(None, type=pa.int64())
pa_y = pa.compute.greater(x, None)
result = y.as_py()   # None

py_x = pa_x.as_py()
result = py_x > 0   # TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

So we see that if we translated to the python world immediately, there would have been a `TypeError`.

In order to make the diagram "commute", the only reasonable solution is therefore to raise a `TypeError` when converting the null-bool to python. This way `result` is the same in both branches - a `TypeError`.

By coercing the null-bool to `False`, one hides this `TypeError` which as said before can lead to all sorts of hard to debug bugs.

AlenkaF:
There has already been changes in the meantime, see https://github.com/apache/arrow/pull/45818. The linked PR adds `__bool__` method to the `BooleanScalar` class and so I will close this PR. Sorry this PR didn't come through.

I will keep the issue open though as the discussion there had different ideas then what is currently implemented.

PR #35001, opened 2023-04-10
GH-32107: [C++] Create Filter Kernels for REE values

felipecrv:
### Rationale for this change

Run-end encoded arrays can very compactly represent filters with repeated values and run-end encoded arrays can be filtered efficiently without inflating all the runs for filtering. This PR implements algorithms that make it possible to leverage these possibilities.

### What changes are included in this PR?

- [x] REE x REE  filters kernels (for the same types supported in `run_end_encode`) 
         Closes #13657
- [x] Plain values x REE filter
- [x] REE values x Plain filter (boolean bitmap)

### Are these changes tested?

By unit tests that handle all the REE/Plain combinations.

### Are there any user-facing changes?

Users don't control which filter kernels get picked directly, so there isn't a user-facing API area yet.
* Closes: #32107

PR #35002, opened 2023-04-10
MINOR: [C++] Make the utilities used to implement run_end_encode available to other compute kernels

felipecrv:
### Rationale for this change

Make it easier to implement correct kernels dealing with run-end encoded data.

Extraction of code to a header so kernels like REE filter kernels (see #35001) can use them.

### Are these changes tested?

Indirectly via the `run_end_encode` and `run_end_decode` tests.

No, these are internal compute headers.

### DIFF
--- a/cpp/src/arrow/CMakeLists.txt
+++ b/cpp/src/arrow/CMakeLists.txt
@@ -402,6 +402,7 @@ list(APPEND
      compute/registry.cc
      compute/kernels/codegen_internal.cc
      compute/kernels/row_encoder.cc
+     compute/kernels/ree_util_internal.cc
      compute/kernels/scalar_cast_boolean.cc
      compute/kernels/scalar_cast_dictionary.cc
      compute/kernels/scalar_cast_extension.cc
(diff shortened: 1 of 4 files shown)


felipecrv:
@zeroshade @benibus

zeroshade (approved):
LGTM

cpp/src/arrow/CMakeLists.txt
```diff
@@ -402,6 +402,7 @@ list(APPEND
      compute/registry.cc
      compute/kernels/codegen_internal.cc
      compute/kernels/row_encoder.cc
+     compute/kernels/ree_util_internal.cc
```
benibus, line 405: Does this need to be included in the default build or should it be conditionally compiled when `ARROW_COMPUTE=ON` alongside `vector_run_end_encode.cc` in [this block](https://github.com/apache/arrow/blob/3a3d1ccde4023cb769eb8a2c10531ee6dce666dc/cpp/src/arrow/CMakeLists.txt#L428)? If you expect the utils will be used outside of the REE-kernels then it's probably fine where it is.

benibus (commented):
Looks pretty straightforward to me! Just one comment.

cpp/src/arrow/CMakeLists.txt
felipecrv: `RunEndEncode` is extremely useful when writing unit tests, so I would consider this "core" and `RunEndDecode` will probably play a central role in `compute` as something we can reach out to whenever a function can't handle REE directly.

benibus: Alright, sounds good. Just wanted to make sure that it was intentional.

PR #35005, opened 2023-04-10
GH-35004: [C++] Remove RelationInfo

rtpsw:
See #35004
* Closes: #35004


rtpsw:
cc @westonpace @icexelloss

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -4356,7 +4356,7 @@ TEST(Substrait, PlanWithAsOfJoinExtension) {
           "extension_multi": {
             "common": {
               "emit": {
-                "outputMapping": [0, 1, 2, 5]
+                "outputMapping": [0, 1, 2, 3]
```
icexelloss, line 4359: Why these change

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -5475,7 +5475,7 @@ TEST(Substrait, AsOfJoinDefaultEmit) {
             }
           }
         },
-        "names": ["time", "key", "value1", "time2", "key2", "value2"]
```
icexelloss, line 5478: Why these change?

cpp/src/arrow/engine/substrait/serde_test.cc
rtpsw: Before these indices where input indices but now they are output indices. The output schema has these columns: "time", "key", "value1", "value2".

cpp/src/arrow/engine/substrait/serde_test.cc
rtpsw: Before the names corresponded to the input indices but now they correspond to the output indices.

icexelloss: Actually I see - this is the correct output schema. @rtpsw Did you test this with internal code? I think we also need to update our ibis-substrait producer to produce an emit outputting that matches this.

icexelloss: > No, I haven't tested this code internally yet. Do you want this test done before proceeding here? I agree the Substrait producer would have to change.

Yeah I think so. We want to make sure internal code can work with change we make here (with slight modification)

rtpsw: This appears to work in internal testing.

rtpsw:
@westonpace, is this good to go from your standpoint?

icexelloss (approved):
LGTM

westonpace (commented):
I'm fine with simplifying things.  But what changed to make this possible?  My understanding was that we needed this because Ibis had decided that the output of an asof-join node was:

```
left fields | right fields
```

and not...

```
left fields | right fields (except for those used in on/by)
```

Was there a corresponding change made in ibis-substrait?

rtpsw:
> But what changed to make this possible?

I'm not qualified to answer for Substrait, but here are some point to consider:

1. The [Substrait spec for column selection](https://substrait.io/tutorial/sql_to_substrait/#column-selection-and-emit) speaks about emit output indices (see also [this post](https://github.com/apache/arrow/issues/35004#issue-1660861440)).
2. The [Substrait spec for joining](https://substrait.io/relations/logical_relations/#join-types) does not describe anything about as-of join.
3. The [Substrait algebra proto](https://github.com/substrait-io/substrait/blob/b5d7ed26a17c0a0bd6d0779d312942e5216ea9fa/proto/substrait/algebra.proto) (also [here](https://github.com/substrait-io/substrait/blob/b5d7ed26a17c0a0bd6d0779d312942e5216ea9fa/proto/substrait/algebra.proto)) does not define an as-of-join relation.
4. The [as-of-join relation is defined as an extension](https://github.com/apache/arrow/blob/a2548a79a191c280966d869c7778643fd64fdcd6/cpp/proto/substrait/extension_rels.proto#L29-L46) in Arrow.

This suggests to me that we are allowed to change the output schema for as-of-join within Arrow, and we need to adhere to the emit output indices defined by Substrait.

icexelloss:
@westonpace The ibis substrait side change is this:
https://github.com/ibis-project/ibis-substrait/pull/554

icexelloss:
@westonpace I went ahead and merged this - let me know if you have any questions happy to answer.

westonpace:
> @westonpace I went ahead and merged this - let me know if you have any questions happy to answer.

No, that's fine.  We probably need to document things like the expected output order of extension nodes at some point.  Once that's done we will need to be more hesitant about this kind of change since it's a backwards incompatible change.  Currently, the node isn't documented, and I'm not aware of anyone using it.  Though this might be something we want to mention in the 13.0.0 release nodes.

PR #35010, opened 2023-04-10
GH-35009: [C#] First level ArrayTypes IEnumerable<T>

### LINKED ISSUE #35009: [C#] Add IEnumerable Extensions to Arrays

It would be nice to add IEnumerable Interface to Array like

```c#
public class StringArray: BinaryArray, IEnumerable<string>
```

### Component(s)

C#

Platob:
Add IEnumerable interface on primitive, binary, string, decimal array types

### Are these changes tested?

I made unit tests in Arrow.Tests

No, Only new feature to use IEnumerable + Linq
* Closes: #35009


csharp/src/Apache.Arrow/Arrays/StringArray.cs
```diff
@@ -91,5 +93,16 @@ public string GetString(int index, Encoding encoding = default)
                     return encoding.GetString(data, bytes.Length);
             }
         }
+
+        // IEnumerable methods
+        public new IEnumerator<string> GetEnumerator()
+        {
+            return Enumerable.Range(0, Length).Select(i => GetString(i)).GetEnumerator();
```
eerhardt, line 100: Instead of using `Enumerable.Range`, can we instead build `Enumerator` structs for these? That will be better performing.

See https://github.com/dotnet/runtime/blob/388edb64f004e1c24b350edca43cf9fd2fdb7a4e/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/List.cs#L1145-L1215 for an example.

Platob: Ok i was hesitating to do it

PR #35013, opened 2023-04-10
GH-34983: [C++] Preserve map values nullability on C Data Interface import

### LINKED ISSUE #34983: [C++] C data bridge doesn't preserve map values nullability on import

The nullability of the values field is lost here, since we pass as a type and not a field:

https://github.com/apache/arrow/blob/82d48a93d0dc5ea7826207f154d9f6a1250ec57b/cpp/src/arrow/c/bridge.cc#L1105

### Component(s)

C++

wjones127:
### Rationale for this change

Map types with non-nullable values don't roundtrip right now through C Data interface, when they arguably should.

I also found that field names don't roundtrip, but I think that might be a feature, not a bug. Therefore I added unit tests enforcing it for now.
 
### Are these changes tested?

Yes, added various unit tests for export and roundtrip.

* Closes: #34983

### DIFF
--- a/cpp/src/arrow/c/bridge.cc
+++ b/cpp/src/arrow/c/bridge.cc
@@ -1102,7 +1102,13 @@ struct SchemaImporter {
     }
 
     bool keys_sorted = (c_struct_->flags & ARROW_FLAG_MAP_KEYS_SORTED);
-    type_ = map(value_type->field(0)->type(), value_type->field(1)->type(), keys_sorted);
+    bool values_nullable = value_type->field(1)->nullable();
+    // Some implementations of Arrow (such as Rust) use a non-standard field name
+    // for key ("keys") and value ("values") fields. For simplicity, we override
+    // them on import.
+    auto values_field =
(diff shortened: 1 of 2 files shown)


cpp/src/arrow/c/bridge.cc
```diff
@@ -1102,7 +1102,13 @@ struct SchemaImporter {
     }
 
     bool keys_sorted = (c_struct_->flags & ARROW_FLAG_MAP_KEYS_SORTED);
-    type_ = map(value_type->field(0)->type(), value_type->field(1)->type(), keys_sorted);
+    bool values_nullable = value_type->field(1)->nullable();
+    // Some implementations of Arrow (such as Rust) use a non-standard field name
+    // for key ("keys") and value ("values") fields. For simplicity, we override
+    // them on import.
```
paleolimbot, line 1108: I need to fix this in the nanoarrow implementation (I think it currently errors for the case where the names are incorrect!)

paleolimbot (approved):
Seems reasonable to me!

PR #35017, opened 2023-04-10
GH-35014: [Python] Make sure unit tests can run without acero

### LINKED ISSUE #35014: [Python] PyArrow tests cannot be run without Acero built

Our minimal job is failing:

https://github.com/ursacomputing/crossbow/actions/runs/4653280100/jobs/8234017973

There don't seem to be any pytest markers allowing skipping acero tests.

https://github.com/apache/arrow/blob/main/python/pyarrow/tests/test_acero.py

### Component(s)

Python

wjones127:
### Rationale for this change

Fixes two failing nightlies:

 * [example-python-minimal-build-fedora-conda](https://github.com/ursacomputing/crossbow/actions/runs/4653275808/jobs/8234009627)
 * [example-python-minimal-build-ubuntu-venv](https://github.com/ursacomputing/crossbow/actions/runs/4653280100/jobs/8234017973)

Adds a pytest mark for Acero, since it is now required.

### Are these changes tested?

Yes, this fixes existing tests. Validated it works locally.

No.
* Closes: #35014

### DIFF
--- a/python/pyarrow/tests/parquet/test_dataset.py
+++ b/python/pyarrow/tests/parquet/test_dataset.py
@@ -1448,7 +1448,6 @@ def test_write_to_dataset_no_partitions_s3fs(
     "ignore:'ParquetDataset:FutureWarning",
     "ignore:'partition_filename_cb':FutureWarning")
 @pytest.mark.pandas
-@pytest.mark.dataset
 @parametrize_legacy_dataset_not_supported
 def test_write_to_dataset_with_partitions_and_custom_filenames(
     tempdir, use_legacy_dataset
@@ -1471,7 +1470,7 @@ def test_write_to_dataset_with_partitions_and_custom_filenames(
                         partition_by, partition_filename_callback,
                         use_legacy_dataset=use_legacy_dataset)
 
-    dataset = pq.ParquetDataset(path)
+    dataset = pq.ParquetDataset(path, use_legacy_dataset=use_legacy_dataset)
 
     # ARROW-3538: Ensure partition filenames match the given pattern
     # defined in the local function partition_filename_callback
--- a/python/pyarrow/tests/parquet/test_dataset.py
+++ b/python/pyarrow/tests/parquet/test_dataset.py
@@ -1448,6 +1448,7 @@ def test_write_to_dataset_no_partitions_s3fs(
     "ignore:'ParquetDataset:FutureWarning",
     "ignore:'partition_filename_cb':FutureWarning")
 @pytest.mark.pandas
+@pytest.mark.dataset
 @parametrize_legacy_dataset_not_supported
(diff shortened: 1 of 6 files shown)


wjones127:
@github-actions crossbow submit example-python-minimal-build-ubuntu-venv

python/pyarrow/tests/test_table.py
```diff
@@ -1989,6 +1989,7 @@ def test_table_select():
     assert result.equals(expected)
 
 
+@pytest.mark.acero
 def test_table_group_by():
```
jorisvandenbossche, line 1993: There are some other `join` tests in this file marked with `pytest.mark.dataset` that can then be changed to "acero"

jorisvandenbossche (commented):
Strictly speaking, we should also mark `test_exec_plan.py` with "acero", but because that is already marked as "dataset", this doesn't show up (but only a subset of the tests there actually need datasets). Enabling datasets currently requires acero, right? So then no need to mark things that require datasets also with acero?

I think we have been typically been using the "dataset" marker up to now, because essentially that gave the same. Given that it is possible to have acero but not datasets, this seems like a good change. But it's also something we don't really test I think? (the minimal build has neither)

python/pyarrow/tests/parquet/test_dataset.py
```diff
@@ -1448,6 +1448,7 @@ def test_write_to_dataset_no_partitions_s3fs(
     "ignore:'ParquetDataset:FutureWarning",
     "ignore:'partition_filename_cb':FutureWarning")
 @pytest.mark.pandas
+@pytest.mark.dataset
```
raulcd, line 1451: This should not be required because we are already marking all as requiring dataset: https://github.com/apache/arrow/blob/main/python/pyarrow/tests/test_dataset.py#L59

wjones127: That's a different file; this one just requires parquet.

python/pyarrow/tests/test_table.py
wjones127: In theory we could, but getting the checks like `not isinstance(left_operand, (Table, ds.Dataset))` to work without datasets sounds like it's more trouble than it's worth at the moment.

wjones127: Actually found a way to do that, writing a little stub class for the module. LMK if there's a better way we would prefer to handle that.

wjones127:
> Enabling datasets currently requires acero, right? So then no need to mark things that require datasets also with acero?

Yes. Datasets requires acero. So this is just marking tests that can run when acero is enabled but datasets isn't.

wjones127:
✅ Tested locally with Acero enabled and Datasets disabled.

jorisvandenbossche (approved):
Thanks!

PR #35028, opened 2023-04-11
GH-35027: [Go]: Use base64.StdEncoding in FixedSizeBinaryBuilder Unmarshal

### LINKED ISSUE #35027: [Go] FixedSizeBinary Marshal/Unmarshal returns error

When using `MarshalJSON` on a `FixedSizeBinary`, I would expect that calling `UnmarshalJSON` on a `FixedSizeBinaryBuilder` with the produced JSON would read the original rows back into the builder. However right now these operations are not symmetrical, and an error is returned instead:

```
failed to unmarshal json: illegal base64 data at input byte 6
```

This is because the `FixedSizeBinaryBuilder` is using `base64.RawStdEncoding` while the `json.Marshal` method is using `base64.StdEncoding`. It seems like we should change the `FixedSizeBinaryBuilder` implementation to match the `json.Marshal` call and also use `base64.StdEncoding`.

I will open a pull request to fix this shortly.

### Component(s)

Go

hermanschaaf:
This changes `FixedSizeBinaryBuilder`'s Unmarshal to use `base64.StdEncoding` instead of `base64.RawStdEncoding`. The previous implementation caused an error when unmarshaling from JSON previously produced by a `FixedSizeBinary` array instance.

I also added a test for Marshal/Unmarshal to set the expectation that these operations should mirror one another.
* Closes: #35027

### DIFF
--- a/go/arrow/array/fixedsize_binary_test.go
+++ b/go/arrow/array/fixedsize_binary_test.go
@@ -109,3 +109,43 @@ func TestFixedSizeBinarySlice(t *testing.T) {
 		t.Fatalf("got=%q, want=%q", got, want)
 	}
 }
+
+func TestFixedSizeBinary_MarshalUnmarshalJSON(t *testing.T) {
+	mem := memory.NewCheckedAllocator(memory.NewGoAllocator())
+	defer mem.AssertSize(t, 0)
+
+	dtype := &arrow.FixedSizeBinaryType{ByteWidth: 4}
+	b := array.NewFixedSizeBinaryBuilder(mem, dtype)
+	defer b.Release()
+
+	var data = [][]byte{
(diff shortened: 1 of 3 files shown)


hermanschaaf:
I'm not sure why some tests are failing for certain architectures and not others. Any help/advice here would be appreciated 🙇

zeroshade:
@hermanschaaf It looks like the common thread is an issue with Go 1.18 (as opposed to Go 1.17 which is succeeding)

disq:
> I'm not sure why some tests are failing for certain architectures and not others. Any help/advice here would be appreciated 🙇

Looks like the go 1.18 tests are more comperehensive, should work now.

zeroshade (approved):

PR #35034, opened 2023-04-11
GH-35033: [Java] [Datasets] Add support for multi-file datasets from Java

### LINKED ISSUE #35033: [Java] [Datasets] Add support for multi-file datasets from Java

I'd like to add support for multi-file datasets from Java, without having to keep them in the same directory. This functionality already exists in the C++ libraries, as well as multiple client bindings (Python, R, etc). 

### Component(s)

Java

noahfrn:
I'd like to add support for multi-file datasets (in different directories) in the Java bindings.
This PR adds to the existing Datasets API to allow an array of URIs to be passed in. The logic for whether the URIs passed are valid or not exists in the JNI layer.
* Closes: #35033


NoahFournier:
Unsure why the flight-core test is failing, doesn't seem related to my change.

lidavidm:
@davisusanibar

java/dataset/src/main/cpp/jni_wrapper.cc
```diff
@@ -533,6 +535,59 @@ Java_org_apache_arrow_dataset_file_JniWrapper_makeFileSystemDatasetFactory(
...
+  std::shared_ptr<arrow::dataset::FileFormat> file_format =
+      JniGetOrThrow(GetFileFormat(file_format_id));
+  arrow::dataset::FileSystemFactoryOptions options;
+
+  std::vector<std::string> uri_vec = ToStringVector(env, uris);
+
+  // If not all URIs, throw exception
+  if (!std::all_of(uri_vec.begin(), uri_vec.end(), arrow::fs::internal::IsLikelyUri)) {
+    JniThrow("All sources must be valid URIs.");
+  }
```
davisusanibar, line 559: What about to only delegate the validation to `arrow::fs::internal::IsLikelyUri`, they could response with the invalid URI error message `java.lang.RuntimeException: Unrecognized filesystem type in URI: http://example.com`

```suggestion
std::all_of(uri_vec.begin(), uri_vec.end(), arrow::fs::internal::IsLikelyUri);
```

java/dataset/src/main/cpp/jni_wrapper.cc
```diff
@@ -533,6 +535,59 @@ Java_org_apache_arrow_dataset_file_JniWrapper_makeFileSystemDatasetFactory(
...
+    auto fs = JniGetOrThrow(arrow::fs::FileSystemFromUri(s, &output_path));
+    return {fs, output_path};
+  });
+
+  // If all URIs, ensure that they all share a FileSystem type
+  if (std::unique(filesystems.begin(), filesystems.end(),
+        [] (const auto& p1, const auto& p2) {
+          return p1.first->type_name() == p2.first->type_name();
+        }) - filesystems.begin() != 1) {
+    JniThrow("Different filesystems are not supported in a multi-file dataset.");
```
davisusanibar, line 575: Could be possible to add the message that contains detail about what different filesystem the user are trying to use?

java/dataset/src/main/java/org/apache/arrow/dataset/file/JniWrapper.java
```diff
@@ -45,6 +45,17 @@ private JniWrapper() {
    */
   public native long makeFileSystemDatasetFactory(String uri, int fileFormat);
 
+  /**
+   * Create FileSystemDatasetFactory and return its native pointer. The pointer is pointing to a
+   * intermediate shared_ptr of the factory instance.
+   *
+   * @param uris List of file uris to read, each path pointing to an individual file
```
davisusanibar, line 52: Is it working only for specific individual file or it is also working for a directory path that contains many individual files?

java/dataset/src/test/java/org/apache/arrow/dataset/file/TestFileSystemDataset.java
```diff
@@ -101,6 +101,29 @@ public void testBaseParquetRead() throws Exception {
...
+    List<ArrowRecordBatch> datum = collectResultFromFactory(factory, options);
+
+    assertScanBatchesProduced(factory, options);
+    assertEquals(2, datum.size());
+    datum.forEach(batch -> assertEquals(1, batch.getLength()));
+    checkParquetReadResult(schema, expectedJsonUnordered, datum);
+
+    AutoCloseables.close(datum);
+
+  }
```
davisusanibar, line 125: Could be possible to add testing for exception throws: (1) Invalid URI , (2) ensure that they all share a FileSystem type

java/dataset/src/main/java/org/apache/arrow/dataset/file/JniWrapper.java
NoahFournier: It will only work for specific individual files, and confirmed this by testing locally. I believe this matches the behaviour of the arrow C++ libraries. We would need to use a selector in order to crawl a directory.

lidavidm: Is there any purpose in keeping the `String, int` overload if we have `String[], int` which should be strictly more general?

java/dataset/src/main/cpp/jni_wrapper.cc
```diff
@@ -533,6 +535,60 @@ Java_org_apache_arrow_dataset_file_JniWrapper_makeFileSystemDatasetFactory(
...
+
+  using FsPathPair = std::pair<std::shared_ptr<arrow::fs::FileSystem>, std::string>;
+
+  std::shared_ptr<arrow::dataset::FileFormat> file_format =
+      JniGetOrThrow(GetFileFormat(file_format_id));
+  arrow::dataset::FileSystemFactoryOptions options;
+
+  std::vector<std::string> uri_vec = ToStringVector(env, uris);
+
+  // If not all URIs, throw exception
```
lidavidm, line 556: We should also be checking if no URIs were passed

java/dataset/src/main/cpp/jni_wrapper.cc
lidavidm: We should be using https://github.com/apache/arrow/pull/34420

NoahFournier: Agreed on both counts, I'll wait for #34420  to clean this up a bit. and then create a better exception.

java/dataset/src/main/java/org/apache/arrow/dataset/file/JniWrapper.java
NoahFournier: The `String, int` overload correctly handles the user passing a directory, which the `String[], int` overload does not (Each `String` needs to be a complete uri to a source). I'll look into whether I could handle both directories and complete URIs in the `String[], int` implementation, in which case we wouldn't need `String, int`.

lidavidm: Aha, thank you.

java/dataset/src/main/cpp/jni_wrapper.cc
ianmcook: @NoahFournier FYI https://github.com/apache/arrow/pull/34420 is merged now

NoahFournier:
Sorry for the delay on this one. I've added the test as requested, and cleaned up the implementation using #34420

lidavidm (approved):
Thank you!

PR #35046, opened 2023-04-11
MINOR: [C++] io_util.cc: fix for macOS without MACH_TASK_BASIC_INFO

barracuda156:
Earlier macOS have no `MACH_TASK_BASIC_INFO`, but they have an equivalent `TASK_BASIC_INFO`. Use the latter as a fall back on Apple when the former is undefined. Fixes the build on earlier systems. Tested on macOS 10.6 (fixes the build), tested on the latest macOS (i.e. the patch does not break anything and works as supposed to).

We use a similar patch for `folly` and elsewhere: https://github.com/macports/macports-ports/blob/eaa29a4816054657e65d001562d7f77ba5157e40/devel/folly/files/patch-older-systems.diff#L232-L251

### DIFF
--- a/cpp/src/arrow/util/io_util.cc
+++ b/cpp/src/arrow/util/io_util.cc
@@ -2139,6 +2139,7 @@ int64_t GetCurrentRSS() {
 
 #elif defined(__APPLE__)
   // OSX ------------------------------------------------------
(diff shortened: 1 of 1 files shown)


felipecrv (approved):

barracuda156:
@pitrou Thank you for help!

PR #35048, opened 2023-04-11
GH-34907: [Docs][R] Version selector reports that release version is dev

### LINKED ISSUE #34907: [Docs][R] Version selector reports that release version is dev

image

It shows as dev when you're on the release version (per the URL). When you select release version, it goes to the release version URL but updates to the dev element in the selector. You can't select the dev version from here--the only way to use the selector to get to the dev version is to select an old number version (which does work correctly), and then you can select "dev", which takes you to the dev URL.

cc @thisisnic 

### Component(s)

Documentation, R

thisisnic:
This PR updates the version number, which was missed in the CRAN release.
* Closes: #34907

### DIFF
--- a/r/pkgdown/assets/versions.json
+++ b/r/pkgdown/assets/versions.json
@@ -5,7 +5,7 @@
     },
     {
         "name": "11.0.0.3 (release)",
-        "version": "/"
+        "version": ""
     },
     {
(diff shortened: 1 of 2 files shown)


thisisnic:
I've also opened https://github.com/apache/arrow-site/pull/342

PR #35051, opened 2023-04-11
GH-35042: [Go][FlightSQL driver] Add TLS configuration

### LINKED ISSUE #35042: [Go][FlightSQL driver] Allow custom TLS configurations

Currently, the `database/sql` FlightSQL driver does not allow for TLS connections. However, it should be possible to enable TLS connections and also specify custom TLS parameters such as certificates to check against, minimum version etc.

This was also discussed in the initial driver PR #34331.

### Component(s)

FlightRPC, Go

srebhan:
This PR adds the ability to enable and customize TLS parameters for Golang's `database/sql` driver.
* Closes: apache/arrow#35042
* GitHub Issue: #35042

### DIFF
--- a/go/arrow/flight/flightsql/driver/README.md
+++ b/go/arrow/flight/flightsql/driver/README.md
@@ -35,7 +35,7 @@ connection pooling, transactions combined with ease of use (see (#usage)).
 
 ## Prerequisits
 
-* Go 1.19+
+* Go 1.17+
 * Installation via `go get -u github.com/apache/arrow/go/v12/arrow/flight/flightsql`
 * Backend speaking FlightSQL
 
@@ -111,6 +111,23 @@ to limit the maximum time an operation can take. This prevents calls that wait
 forever, e.g. if the backend is down or a query is taking very long. When
 not set, the driver will use an _infinite_ timeout.
 
+#### `tls`
+
+The `tls` parameter allows to enable and customize Transport-Layer-Security
+settings. There are some special values for the parameters:
+
+* `disabled` or `false` will disable TLS for this server connection. In this
+  case all other settings are ignored.
+* `enabled` or `true` will force TLS for this server connection. In this case
+  the system settings for trusted CAs etc will be used.
+* `skip-verify` will enable TLS for this server connection but will not verify
+  the server certificate. **This is a security risk and should not be used!**
+
+Any other value will be interpreted as the name of a custom configuration. Those
+configurations must be registered either by
+[creating the DSN from configuration](#driver-config-usage) or by calling
+`RegisterTLSConfig()` (see [TLS setup](#tls-setup) for details).
+
 ## Driver config usage
 
 Alternatively to specifying the DSN directly you can fill the `DriverConfig`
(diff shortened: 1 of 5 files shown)


go/arrow/flight/flightsql/driver/README.md
```diff
@@ -35,7 +35,7 @@ connection pooling, transactions combined with ease of use (see (#usage)).
 
 ## Prerequisits
```
disq, line 36:
```suggestion
## Prerequisites
```

go/arrow/flight/flightsql/driver/config.go
```diff
@@ -19,9 +19,53 @@ import (
...
+
+	"github.com/google/uuid"
+)
+
+// TLS configuration registry
+var (
+	tlsConfigRegistry = map[string]*tls.Config{
+		"skip-verify": {InsecureSkipVerify: true},
+	}
+	tlsRegistryMutex sync.Mutex
```
zeroshade, line 33: would it make sense to use an `RWMutex` so the `GetTLSConfig` function can use `RLock` and `RUnlock` ?

go/arrow/flight/flightsql/driver/config.go
```diff
@@ -30,14 +74,15 @@ type DriverConfig struct {
...
 }
 
 func NewDriverConfigFromDSN(dsn string) (*DriverConfig, error) {
 	u, err := url.Parse(dsn)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("invalid URL: %w", err)
 	}
 
 	// Sanity checks on the given connection string
```
zeroshade, line 88: as I'm looking at this, is there any reason that this doesn't follow the general consistency of other implementations by using a scheme of `grpc+tcp` vs `grpc+unix` vs `grpc+tls`?

go/arrow/flight/flightsql/driver/config.go
srebhan: I can certainly do this, but I don't expect this to be in a hot-path. Remember, the mutex is locked if you register a config _and_ when you are calling `Open` (which in turn does the DSN parsing). Do you want me to use a `RWMute`?

zeroshade: I agree that it's not likely in a hot path and unlikely to be a cause for contention. It's probably alright for it to be a regular mutex. this is fine.

go/arrow/flight/flightsql/driver/config.go
srebhan: I guess that's because I wanted to keep the same scheme other SQL drivers use (see https://github.com/xo/dburl#example-urls). I think we could also support the ones above even though I don't like the redundant `grpc`...

srebhan: I think this is unrelated to this PR. Are you OK with opening an issue for this request and proceed with this PR?

zeroshade: sure thing

zeroshade: Created issue #35080

zeroshade (approved):
LGTM, support for the other schemes moved to #35080

Issue #34940, opened 2023-04-06
[R][Doc] Remove discussion of Scanner from vignettes


nealrichardson:
### Describe the enhancement requested

It's at least in data_objects.Rmd, and it's not really accurate: we don't use the Scanner to evaluate dplyr queries anymore. Plus, I think Scanner is deprecated or on its way to deprecation. 

### Component(s)

Documentation, R

thisisnic:
This issue has been marked as stale because it has had no activity in the past 2 years. Please remove the stale label or comment below, or this issue will be closed in 14 days.

PR #35055, opened 2023-04-11
GH-34436: [R] Bindings for JSON Dataset 

### LINKED ISSUE #34436: [R] Bindings for JSON Dataset

See #33732 where it was added to the C++ library

### Component(s)

R

thisisnic:
* Closes: #34436


r/R/dataset-format.R
```diff
@@ -113,6 +115,35 @@ ParquetFileFormat$create <- function(...,
...
 
+#' JSON dataset file format
+#'
+#' @description
+#' A `JsonFileFormat` is a [FileFormat] subclass which holds information about how to
+#' read and parse the files included in a JSON `Dataset`.
+#'
+#' @section Factory:
+#' `JsonFileFormat$create()` can take options in the form of lists passed through as `parse_options`,
+#'  or `read_options` parameters.
```
westonpace, line 126: Is it clear to the user what they should pass in for `parse_options` and `read_options`?

Also, some of the parse options don't apply when used in a datasets context (as opposed to reading a single file).  I believe we ignore `explicit_schema` (in favor of the dataset schema) and `unexpected_field_behavior` (we default to ignore I think since we need to tolerate partial reads).

westonpace (approved):
The C++ / bindings changes look good to me.  I didn't look too much at the R part.

cpp/src/arrow/dataset/api.h
```diff
@@ -36,4 +36,7 @@
 #ifdef ARROW_PARQUET
 #include "arrow/dataset/file_parquet.h"
 #endif
+#ifdef ARROW_JSON
+#include "arrow/dataset/file_json.h"
+#endif
```
pitrou, line 41: These lines are already included above, no need to add them again.

r/tests/testthat/test-dataset-json.R
```diff
@@ -0,0 +1,58 @@
...
+      select(string = chr, integer = int) %>%
+      filter(integer > 6) %>%
+      summarize(mean = mean(integer))
+  )
+  # Collecting virtual partition column works
+  expect_equal(
+    collect(ds) %>% arrange(part) %>% pull(part),
+    c(rep(5, 10), rep(6, 10))
+  )
+})
```
pitrou, line 58: `JsonFragmentScanOptions` doesn't look tested, is it?

cpp/src/arrow/dataset/api.h
thisisnic: Ah, must have crept in when I rebased, good catch!

r/R/dataset-format.R
```diff
@@ -532,6 +565,22 @@ ParquetFragmentScanOptions$create <- function(use_buffered_stream = FALSE,
   dataset___ParquetFragmentScanOptions__Make(use_buffered_stream, buffer_size, pre_buffer)
 }
 
+#' @usage NULL
+#' @format NULL
+#' @rdname FragmentScanOptions
+#' @export
+JsonFragmentScanOptions <- R6Class("JsonFragmentScanOptions", inherit = FragmentScanOptions)
```
thisisnic, line 572: I've only created bindings for this here, as it mirrors the CSV implementation, but given that https://github.com/apache/arrow/issues/34940 implies that the scanner is deprecated, should I remove this? @nealrichardson

r/R/dataset-format.R
thisisnic: Good point, have added a bit more in the docs about these params now.

r/R/dataset-format.R
nealrichardson: No, it's fine, Acero needs these too, not just the old Scanner.

r/R/dataset-format.R
```diff
@@ -532,6 +575,22 @@ ParquetFragmentScanOptions$create <- function(use_buffered_stream = FALSE,
...
+JsonFragmentScanOptions$create <- function(...) {
+  dots <- list2(...)
+  parse_opt_choices <- dots[names(dots) %in% names(formals(JsonParseOptions$create))]
+  read_opt_choices <- dots[names(dots) %in% names(formals(JsonReadOptions$create))]
+
+  parse_options <- do.call(JsonParseOptions$create, parse_opt_choices)
+  read_options <- do.call(JsonReadOptions$create, read_opt_choices)
+
```
paleolimbot, line 590: Should there be an error for elements of `dots` that are unaccounted for?

paleolimbot (commented):
This is so close!

I know that we don't necessarily have a way to extract options that we set on a Dataset; however, I think we should add a test with an `open_dataset()` that passes some though (i.e., `use_threads` and/or `newlines_in_values` and/or `schema`), and if possible, test that they actually did something.

For `use_threads = FALSE`: I believe that `option_use_threads()` returns FALSE on Windows, so this is sort of implicitly tested; however, I think at least one test should explicitly attempt to pass it.

For `newlines_in_values = TRUE`: You could do something like write `{"key": "\nvalue"}\n` to a file and attempt collecting to make sure the newline in the value comes through.

For `schema`: You could do something like write `{"key": "value"}\n` to a file and attempt collecting with `schema(key = int32())` and make sure it errors?

r/R/dataset-format.R
thisisnic: Good call, added this plus a test

r/R/dataset-format.R
```diff
@@ -113,6 +115,45 @@ ParquetFileFormat$create <- function(...,
...
+#' A `JsonFileFormat` is a [FileFormat] subclass which holds information about how to
+#' read and parse the files included in a JSON `Dataset`.
+#'
+#' @section Factory:
+#' `JsonFileFormat$create()` can take options in the form of lists passed through as `parse_options`,
+#'  or `read_options` parameters.
+#'
+#'  Available `read_options` parameters:
+#'  * `use_threads`: Whether to use the global CPU thread pool. Default `TRUE`. If `FALSE`, JSON input must end with an
+#'  empty line.
```
westonpace, line 130: It seems odd these two things are related.  Am I missing something?

r/R/dataset-format.R
```diff
@@ -113,6 +115,45 @@ ParquetFileFormat$create <- function(...,
...
+#'  * `block_size`: Block size we request from the IO layer; also determines size of chunks when `use_threads`
+#'   is `TRUE`.
```
westonpace, line 132: This leads me to wonder what determines the size of chunks when `use_threads` is false?  Although maybe there just aren't chunks?

westonpace (approved):
Just a few minor questions but this seems fine to me otherwise.

r/R/dataset-format.R
thisisnic: I must have copied and pasted this to the wrong place, good catch!

r/R/dataset-format.R
thisisnic: I have no idea; I copied from https://github.com/apache/arrow/blob/95972cca1f65179b63a9fe5262d1092547e3d137/r/R/csv.R#L378-L384

thisisnic:
> This is so close!
> 
> I know that we don't necessarily have a way to extract options that we set on a Dataset; however, I think we should add a test with an `open_dataset()` that passes some though (i.e., `use_threads` and/or `newlines_in_values` and/or `schema`), and if possible, test that they actually did something.

Actually, I'm going to push back on this.  We don't test these options fully when reading CSV datasets, and given they relate to internal mechanisms of how things run not what is returned, it's going to be difficult to test here, and should be tested in the C++ anyway.  Happy to write tests to check that these options are passed through successfully though.

r/R/dataset-format.R
westonpace: Good enough for me :)

thisisnic:
> I don't think that our previous lack of of test coverage should mean that we continue to add features that are not tested. I am also happy to defer to another review on this, though.
> 
> How about:
> 
> ```r
> library(arrow, warn.conflicts = FALSE)
> #> Some features are not enabled in this build of Arrow. Run `arrow_info()` for more information.
> library(testthat, warn.conflicts = FALSE)
> 
> test_that("JsonParseOptions are passed through to the dataset", {
>   tf <- tempfile()
>   on.exit(unlink(tf))
>   write('{"key": "value"}', tf)
>   
>   expect_error(
>     # (but instead use open_dataset())
>     read_json_arrow(tf, schema = schema(key = int32())),
>     "parse error"
>   )
> })
> #> Test passed 😀
> ```
> 
> (Ideally we'd be able to do something like `dataset$parse_options$schema` instead, which I think you've started to implement with the CSV options, but this seems slightly easier)

Feel free to push to the branch @paleolimbot

paleolimbot (approved):
Let's track separately! (See #36138)

PR #35067, opened 2023-04-12
GH-35067: [JavaScript] toString for signed `BigNum`s

### LINKED ISSUE #22932: [JS] decimal toString does not support negative values
The main description is here: <https://github.com/apache/arrow/issues/5397>

Also, I have a simple test case (slightly changed generate-test-data.js and generated-data-validators):
```java

export const decimal = (length = 2, nullCount = length * 0.2 | 0, scale = 0, precision = 38) => vectorGenerator.visit(new Decimal(scale, precision), length, nullCount);

function fillDecimal(length: number) {
    // const BPE = Uint32Array.BYTES_PER_ELEMENT; // 4
    const array = new Uint32Array(length);
    // const max = (2 ** (8 * BPE)) - 1;
    // for (let i = -1; ++i < length; array[i] = rand() * max * (rand() > 0.5 ? -1 : 1));
    array[0] = 0;
    array[1] = 1286889712;
    array[2] = 2218195178;
    array[3] = 4282345521;
    array[4] = 0;
    array[5] = 16004768;
    array[6] = 3587851993;
    array[7] = 126217744;
    return array;
}
```
and the expected value should be
```java

expect(vector.get(0).toString()).toBe('-1000000000000000000000000000000000000');
expect(vector.get(1).toString()).toBe('1000000000000000000000000000000000000');
```
However, the actual first value is 339282366920938463463374607431768211456 which is wrong! The second value is correct by the way.

I believe the bug is in the function called 

function decimalToString<T extends BN<BigNumArray>>(a: T) because it cannot return a negative value at all.

[arrow/js/src/util/bn.ts](https://github.com/apache/arrow/blob/d54425de19b7dbb2764a40355d76d1c785cf64ec/js/src/util/bn.ts#L99)

Line 99 

**Reporter**: [Andong Zhan](https://issues.apache.org/jira/browse/ARROW-6575)
#### Related issues:
- [[JS] decimalToString function in bn.ts does not handle negative decimals](https://github.com/apache/arrow/issues/22980) (is duplicated by)

<sub>**Note**: *This issue was originally created as [ARROW-6575](https://issues.apache.org/jira/browse/ARROW-6575). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

aljazerzen:

Closes apache/arrow#22932

Basically, signed negative numbers were displayed as very large positive numbers.

### Are these changes tested?

Yes, I've also added tests for `bn.ts` (BigNum) that was not tested before.

Negative numbers stored in BigNum will now be displayed as negative numbers instead of very large positive numbers.

Note that this change also affects decimals, because they are stored in BigNum as signed numbers. Decimals still don't convert to string correctly, because inserting the decimal dot is not implemented. Ref apache/arrow-js#100.
* Closes: apache/arrow#35067
* GitHub Issue: #35067

### DIFF
--- a/js/src/util/bn.ts
+++ b/js/src/util/bn.ts
@@ -102,7 +102,7 @@ export const bigNumToString: { <T extends BN<BigNumArray>>(a: T): string } = (<T
         return unsignedBigNumToString(a);
     }
 
-    const array = new Uint16Array(a.buffer, a.byteOffset, a.byteLength / 2);
+    let array = new Uint16Array(a.buffer, a.byteOffset, a.byteLength / 2);
 
     // detect positive numbers
     const highOrderWord = new Int16Array([array[array.length - 1]])[0];
@@ -111,6 +111,7 @@ export const bigNumToString: { <T extends BN<BigNumArray>>(a: T): string } = (<T
     }
 
     // flip the negative value
+    array = array.slice();
     let carry = 1;
     for (let i = 0; i < array.length; i++) {
         const elem = array[i];
--- a/js/src/util/bn.ts
+++ b/js/src/util/bn.ts
@@ -97,7 +97,30 @@ export const bigNumToString: { <T extends BN<BigNumArray>>(a: T): string } = (<T
         return `${bigIntArray[0]}`;
     }
 
-    return unsignedBigNumToString(a);
+    // unsigned numbers
+    if (!a.signed) {
+        return unsignedBigNumToString(a);
+    }
+
+    const array = new Uint16Array(a.buffer, a.byteOffset, a.byteLength / 2);
+
+    // detect positive numbers
+    const highOrderWord = new Int16Array([array[array.length - 1]])[0];
+    if (highOrderWord >= 0) {
+        return unsignedBigNumToString(a);
+    }
+
+    // flip the negative value
+    let carry = 1;
+    for (let i = 0; i < array.length; i++) {
+        const elem = array[i];
+        const updated = ~elem + carry;
+        array.set([updated], i);
+        carry &= elem === 0 ? 1 : 0;
+    }
+
+    const negated = unsignedBigNumToString(<any>array);
+    return `-${negated}`;
 });
 
 /** @ignore */
--- a/js/src/util/bn.ts
+++ b/js/src/util/bn.ts
@@ -36,17 +36,17 @@ function BigNum(this: any, x: any, ...xs: any) {
 }
 
 BigNum.prototype[isArrowBigNumSymbol] = true;
-BigNum.prototype.toJSON = function <T extends BN<BigNumArray>>(this: T) { return `"${bignumToString(this)}"`; };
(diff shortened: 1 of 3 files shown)


trxcllnt (approved):
Looks good to me, thanks @aljazerzen!

js/src/util/bn.ts
```diff
@@ -90,12 +90,52 @@ function bignumToNumber<T extends BN<BigNumArray>>(bn: T) {
...
+        return unsignedBigNumToString(a);
+    }
+
+    // flip the negative value
+    array = array.slice();
+    let carry = 1;
+    for (let i = 0; i < array.length; i++) {
+        const elem = array[i];
+        const updated = ~elem + carry;
+        array.set([updated], i);
```
trxcllnt, line 119:
```suggestion
        array[i] = updated;
```

aljazerzen:
I've spent a little time trying to get tests to pass, but something strange is happening:

the `bn-tests.ts` are failing only on `-t es5 -m umd`. I've debugged it to the finding that in [this line](https://github.com/apache/arrow/pull/35067/files#diff-f0dbeeadc965880cf706bd8b8776f24dacf74f260b1faa5c60e4d62dbb76ecd7R101), `a.signed` is false, even though I've constructed a signed number or a decimal (which is signed).

It may be a problem with how these prototypes are assigning to each other, but that is beyond me.

trxcllnt:
@aljazerzen 99% sure closure is mangling too aggressively, and we need to do `if(!a['signed'])` to defeat it.

js/src/util/bn.ts
```diff
@@ -90,12 +90,52 @@ function bignumToNumber<T extends BN<BigNumArray>>(bn: T) {
...
-export const bignumToString: { <T extends BN<BigNumArray>>(a: T): string } = (<T extends BN<BigNumArray>>(a: T) => a.byteLength === 8 ? `${new a['BigIntArray'](a.buffer, a.byteOffset, 1)[0]}` : decimalToString(a));
+export const bigNumToString: { <T extends BN<BigNumArray>>(a: T): string } = (<T extends BN<BigNumArray>>(a: T) => {
+    // use BigInt native implementation
+    if (a.byteLength === 8) {
+        const bigIntArray = new a['BigIntArray'](a.buffer, a.byteOffset, 1);
+        return `${bigIntArray[0]}`;
+    }
+
+    // unsigned numbers
+    if (!a.signed) {
```
trxcllnt, line 101:
```suggestion
    if (!a['signed']) {
```

aljazerzen:
@domoritz can you review this? @trxcllnt can it be merged with only one approval?

domoritz:
One review is fine. We can merge this.

aljazerzen:
Great. Please go ahead, as I don't have the permissions.

domoritz:
Thank you. I had to get back to my computer to use the merge script.

PR #35075, opened 2023-04-12
GH-34539: [C++] Fix throttled scheduler to avoid stack overflow in dataset writer

### LINKED ISSUE #34539: [C++][Python] Setting rows per group cause segmentation fault writing parquet with write_dataset

Writing parquet file crashes with segmentation fault with [`write_dataset`](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.write_dataset.html) and setting `min_rows_per_file` and `max_rows_per_file`.

It seems the crash depends on particular number for example setting `10_000` for both parameters works while `60_000` crashes.

`pyarrow==11.0`

See an example and dedicated docker environment

## segfault.py

```python
import pyarrow as pa
from pyarrow import dataset as ds
from pyarrow import fs

TOTAL = 2**20

schema = pa.schema(
    [
        pa.field("id", pa.uint64()),
    ]
)

data = (pa.RecordBatch.from_pylist([{"id": i}], schema=schema) for i in range(TOTAL))

file_system, path = fs.FileSystem.from_uri("file:///tmp/data")


ds.write_dataset(
    data=data,
    base_dir=path,
    format="parquet",
    schema=schema,
    filesystem=file_system,
    min_rows_per_group=60_000,
    max_rows_per_group=60_000,
)
```

## Dockerfile

```shell
FROM python:3.10.10-slim-buster
RUN pip install pyarrow==11.0
COPY segfault.py /tmp/segfault.py
```

## Run the example

- `docker build . -t pyarrow-segfault`
- `docker run --rm -it pyarrow-segfault bash`
- `python /tmp/segfault.py`

### Component(s)

Python

westonpace:
### Rationale for this change

Fixes a bug in the throttled scheduler.

The throttled scheduler will no longer recurse in the ContinueTasks loop if the continued task was immediately finished.

### Are these changes tested?

Yes, I added a new stress test that exposed the stack overflow very reliably on a standard Linux system.

No.
* Closes: #34539

### DIFF
--- a/cpp/src/arrow/util/async_util.cc
+++ b/cpp/src/arrow/util/async_util.cc
@@ -322,7 +322,7 @@ class ThrottledAsyncTaskSchedulerImpl
       return true;
     } else {
       lk.unlock();
-      return SubmitTask(std::move(task), latched_cost);
+      return SubmitTask(std::move(task), latched_cost, /*in_continue=*/false);
     }
   }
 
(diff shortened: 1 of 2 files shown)


westonpace:
CC @benibus for review since you faced something very similar recently :smile:

lidavidm (approved):

snizovtsev:
I've tested the branch on my workload and can confirm that stack overflows crashes had gone.

PR #35103, opened 2023-04-13
GH-34907: [Docs][R] Version selector reports that release version is dev

### LINKED ISSUE #14969: [R] pkgdown built-in search doesn't work

Built-in search does not seem working.

image

### Component(s)

R, Website

### LINKED ISSUE #34907: [Docs][R] Version selector reports that release version is dev

thisisnic:
This PR moves the functionality to select a different version of the docs from a dropdown menu (which replaces the version number span on the page) to a link in the footer, which links to a HTML page with clickable links to the different docs versions.

This resolves the problem in #34907 and also fixes #14969 in which the JavaScript used in the dropdown interferes with the JavaScript code used for the search box (present since the upgrade to BS5).

This change also simplifies the maintenance of the version selection functionality, though does remove the ability to navigate directly to the same reference page in a different version - overall I think this minor loss of functionality is a worthwhile trade for simpler maintenance.

Footer before (note that the author information is already in the sidebar, so this is duplication):
image

Footer after:
image

* Closes: #34907


thisisnic:
I am wondering if this is even necessary and whether anyone is actually using older versions of the docs? Is there somewhere it might make sense to add a note that these older versions of the docs are available, with an example URL, and just let people find it themselves?

r/vignettes/prevdocs.Rmd
```diff
@@ -0,0 +1,12 @@
+---
+title: "Current and Older Versions of this Documentation"
+output: rmarkdown::html_vignette
+---
+
+```{r, echo=FALSE}
+versions <- jsonlite::read_json("https://arrow.apache.org/docs/r/versions.json")
```
kou, line 7: How about reading local `r/pkgdown/assets/versions.json` instead of `https://arrow.apache.org/docs/r/versions.json`?
It seems that we don't need to use https://arrow.apache.org/docs/r/versions.json here.

kou (commented):
It seems that we can't move to newer versions from older versions with this approach. For example, we don't have a 13.0.0 link in the 12.0.0 docs.
Is it acceptable for R users?
(I don't have a strong opinion for this because I'm not a R user.)

thisisnic:
>  Is it acceptable for R users?

I don't know as I don't know if anyone uses this functionality at all, and nobody requested it.  However, I'd say the main use case is probably someone coming to the docs but having to work with an older version of the package. They can still navigate to the latest version by re-navigating to the main page.

Alternatively, we could have that footer link point to a page which is published independently (e.g. so to a URL like `https://arrow.apache.org/docs/r/versions.html`).  This might be a better solution than having this as an article actually.

thisisnic:
I've updated this now so that instead of adding an article-which-isn't-an-article, we instead generate a HTML page when we do a release, which contains links to all other docs versions.  This page is updated every release, so it always contains an up-to-date version of available docs versions.

dev/release/utils-update-docs-versions.py
```diff
@@ -26,6 +26,7 @@
 
 main_versions_path = dir_path + "/docs/source/_static/versions.json"
 r_versions_path = dir_path + "/r/pkgdown/assets/versions.json"
+r_html_path = dir_path + "/r/pkgdown/assets/versions.html"
```
kou, line 29: Could you add a generated HTML to this pull request?
We should version control the HTML, right?

dev/release/utils-update-docs-versions.py
```diff
@@ -87,3 +88,15 @@
...
+
+# Load the updated versions JSON file
+with open(r_versions_path) as json_file:
+    data = json.load(json_file)
+
+# Write HTML to file
+with open(r_html_path, 'w') as html_file:
+    html_file.write('<!DOCTYPE html>\n<html>\n<body>')
+    for i in data:
+        html_file.write('<p><a href="../' + i['version'] + 'r/">' + i['name'] + '</a></p>\n')
```
kou, line 100: I think that `i['version']` and `i['name']` will not include any HTML special characters such as `<` but how about escape them for safety?

```python
import html
html.escape(i['version'])
```

How about using f-string?

```suggestion
        html_file.write(f'<p><a href="../{i["version"]}/r/">{i["name"]}</a></p>\n')
```

thisisnic: Thanks, done!

r/pkgdown/assets/versions.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+<body><p><a href="../dev/r/">11.0.0.9000 (dev)</a></p>
+<p><a href="../r/">11.0.0 (release)</a></p>
```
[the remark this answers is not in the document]
thisisnic: With the R docs, we only go to 3 levels (to reflect the Arrow release versions), so 11.0.0 effectively is the same as 11.0.0.3.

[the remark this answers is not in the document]
kou: I see.

kou (approved):
+1

PR #35108, opened 2023-04-13
GH-35107: [FlightSQL]: Use `uint8` to refer to 8 bit unsigned integers rather than `uint1`

### LINKED ISSUE #35107: [Format][FlightSQL] Spec refers to "uint1" type (which seems Java specific) rather than "uint8"

While implement FlightSQL in our project, we found that the FlightSql.proto format reference refers to a type `unit1` (not `uint16` or `uint8`)

https://github.com/apache/arrow/blob/6432a2382e870633e5060b5e6ea481f28ca0c725/format/FlightSql.proto#L1327-L1328

What `uint1` means was slightly unclear as the Arrow spec refers to "Int" types with `bitWidth` of 8, 16, 32 or 64 but the rust implementation has UInt8, UInt16, UInt32 and UInt64. 

https://github.com/apache/arrow/blob/6432a2382e870633e5060b5e6ea481f28ca0c725/format/Schema.fbs#L145-L148

## Go  uses `uint8`:

https://github.com/apache/arrow/blob/6432a2382e870633e5060b5e6ea481f28ca0c725/go/arrow/flight/flightsql/schema_ref/reference_schemas.go#L66

## C++  uses `uint8`:

https://github.com/apache/arrow/blob/6432a2382e870633e5060b5e6ea481f28ca0c725/cpp/src/arrow/flight/sql/server.cc#L1145

## Java uses "UINT1"
https://github.com/apache/arrow/blob/6432a2382e870633e5060b5e6ea481f28ca0c725/java/flight/flight-sql/src/main/java/org/apache/arrow/flight/sql/FlightSqlProducer.java#L851-L852

Reading the Java Arrow implementation, it appears to use the term UINT1 for 8 bit integers:
https://github.com/apache/arrow/blob/6432a2382e870633e5060b5e6ea481f28ca0c725/java/vector/src/main/java/org/apache/arrow/vector/ipc/JsonFileReader.java#L368-L381

However, inconsistently, while other parts of the flightsql spec refer to uint32
https://github.com/apache/arrow/blob/6432a2382e870633e5060b5e6ea481f28ca0c725/format/FlightSql.proto#L34

The Java code uses UINT4 for this  https://github.com/apache/arrow/blob/6432a2382e870633e5060b5e6ea481f28ca0c725/java/flight/flight-sql/src/main/java/org/apache/arrow/flight/sql/FlightSqlProducer.java#L874

# Proposal
I propose that we change the proto spec to be consistent and use `uint<bitwidth>` to refer to integer types. So specifically,  change `uint1` to  `uint8` to match C++/Rust/Go as well as the convention in the rest of Flight.proto

Here is the reference to details of our discussion https://github.com/influxdata/influxdb_iox/pull/7532/files#r1165307397

### Component(s)

Format

alamb:
### Rationale for this change

The spec is inconsistent -- see details on https://github.com/apache/arrow/issues/35107

Use `uint8` to refer to 8 bit unsigned integers rather than `uint1`

### Are these changes tested?
No, they are comment only changed

This clarifies a small corner case in the document
* Closes: #35107

### DIFF
--- a/format/FlightSql.proto
+++ b/format/FlightSql.proto
@@ -1324,8 +1324,8 @@ enum UpdateDeleteRules {
  *  key_sequence: int not null,
  *  fk_key_name: utf8,
  *  pk_key_name: utf8,
- *  update_rule: uint1 not null,
- *  delete_rule: uint1 not null
+ *  update_rule: uint8 not null,
+ *  delete_rule: uint8 not null
  * >
  * The returned data should be ordered by fk_catalog_name, fk_db_schema_name, fk_table_name, fk_key_name, then key_sequence.
  * update_rule and delete_rule returns a byte that is equivalent to actions declared on UpdateDeleteRules enum.
@@ -1370,8 +1370,8 @@ message CommandGetExportedKeys {
  *  key_sequence: int not null,
  *  fk_key_name: utf8,
  *  pk_key_name: utf8,
- *  update_rule: uint1 not null,
- *  delete_rule: uint1 not null
+ *  update_rule: uint8 not null,
+ *  delete_rule: uint8 not null
  * >
  * The returned data should be ordered by pk_catalog_name, pk_db_schema_name, pk_table_name, pk_key_name, then key_sequence.
  * update_rule and delete_rule returns a byte that is equivalent to actions:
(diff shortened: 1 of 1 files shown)


alamb:
cc @lidavidm

lidavidm:
I'm away from a computer for a while but in principle it seems good to me.

kou (approved):
+1

I think that we don't need a vote for this change because this change is just a notation change.

alamb:
> +1

> I think that we don't need a vote for this change because this change is just a notation change.

I agree -- thank you -- I don't view this as a change of the spec, rather a textual clarification.

alamb:
I'll plan to leave this open for a few more days in case anyone else would like time to comment

lidavidm (approved):
+1

Agree that we don't need a vote for a clarification.

PR #35113, opened 2023-04-13
GH-35112: [Python] Expose keys_sorted in python MapType

### LINKED ISSUE #35112: [python] Expose keys_sorted in MapType

That field is present in the C++ representation of MapType but not accessible in the python API. 

I think it's just a matter of adding it to the types.pxi.

### Component(s)

Python

0x26res:
### Rationale for this change

It not possible to read `keys_sorted` in the python API

- expose keys_sorted in `cdef class MapType` / types.pxi
- add tests

### Are these changes tested?

yes

We're exposing keys_sorted but I guess the documentation will update itself from the `"""` pydoc (?)

This is not an API breaking change

* Closes: #35112


danepitkin:
LGTM! I'll let someone with committers rights finalize this review.

I believe you can ignore the appveyor error, since there's been issues on main with pytests timing out.

AlenkaF:
The binding looks good to me also +1, thank you for the contribution!

I do have a general, probably silly question, about the keyword in general. Looking at the C++ and the tests, it is meant as a "metadata" keyword and not a "check" that the data is actually sorted, right? What I mean is, you can have a `MapType` defined with `keys_sorted=True` but using it in Scalars for example, the keys do not actually have to be sorted (ascending?):

```python
>>> ty = pa.map_(pa.string(), pa.int8(), keys_sorted=True)
>>> v = [('b', 2), ('a', 1)]
>>> s = pa.scalar(v, type=ty)
>>> s
<pyarrow.MapScalar: [('b', 2), ('a', 1)]>
```

And Dane is correct, the failing tests are not connected to this PR.

danepitkin:
Good catch, @AlenkaF . Is that a possible bug that the underlying C++ implementation doesn't maintain a sorted order?

0x26res:
Are we then saying that:
- This field is pure meta data and it's left to the user to provide sorted keys? 
- it won't be enforced by arrow?

For context, I'm not really using that field. I just need to be able to access it in order to create slightly modified copies of schemas. For example if I want to change the type of nested fields (int32 -> int64). Then I need to make copy of `pa.map_
` types, preserving the keys_sorted metadata.

| So maybe the question is do we want to add a check for it in PyArrow Array and Scalar?

Maybe we should create a follow up issue to do this. It would involve making some change that may break some stuff at runtime (if someone was previously providing unsorted data with `keys_sorted=True`).

As far as this MR is concerned, I think we should just improve the doc for that field (and probably update the doc in the C++ MapType class).

AlenkaF:
Yeah, the issue of `keys_sorted` being only metadata or not should be a separate from this PR. We can open a new issue to start a discussion on that.

I agree that for this PR the aim is to expose the parameter as a property so we are able to get the information from the Data Type.

My suggestion for the docs would be to only make it explicit in PyArrow for example: "_Should_ the entries be sorted according to keys."

0x26res:
@AlenkaF thanks for the suggestion, I've updated the comment.

AlenkaF (approved):
Thanks!

PR #35114, opened 2023-04-13
GH-35124: [C++] Avoid unnecessary copy when outputting join result

### LINKED ISSUE #35124: [C++] Possibly an unnecessary copy when outputting join result batch

See this line: https://github.com/apache/arrow/blob/913c34506000fd77444ea7b7f24dcb6402003d6b/cpp/src/arrow/acero/swiss_join_internal.h#L571

According to the email discussion with @save-buffer :

> I _think_ the compiler will automatically emit the move because it sees we’re copying from an object that’ll never be used again [1], but adding the std::move would be good just to remove any ambiguity.

### Component(s)

C++

zanmato1984:
### Rationale for this change

Seems this https://github.com/apache/arrow/blob/913c34506000fd77444ea7b7f24dcb6402003d6b/cpp/src/arrow/acero/swiss_join_internal.h#L571 is an unnecessary copy when outputting join result batch. According to Sasha K.:

> I _think_ the compiler will automatically emit the move because it sees we’re copying from an object that’ll never be used again [1], but adding the std::move would be good just to remove any ambiguity.

Adding `std::move()` to disambiguate/eliminate the unnecessary copy.

### Are these changes tested?

No. This change is possibly undetectable, and the existing tests should cover it well.

No.

* Closes: #35124

### DIFF
--- a/cpp/src/arrow/acero/swiss_join_internal.h
+++ b/cpp/src/arrow/acero/swiss_join_internal.h
@@ -568,7 +568,7 @@ class JoinResultMaterialize {
       if (num_rows_appended < num_rows_to_append) {
         ExecBatch batch;
         ARROW_RETURN_NOT_OK(Flush(&batch));
-        ARROW_RETURN_NOT_OK(output_batch_fn(batch));
+        ARROW_RETURN_NOT_OK(output_batch_fn(std::move(batch)));
         num_rows_to_append -= num_rows_appended;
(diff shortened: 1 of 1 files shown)


save-buffer (approved):
@westonpace Please take a look as well when you get the chance

westonpace (approved):
Thanks for the cleanup.

Issue #35116, opened 2023-04-13
[C++] Need a nightly test that compiles AVX2


westonpace:
### Describe the enhancement requested

I use AVX2 locally and ran into a few regressions.  It seems we do not have a CI job that is testing AVX2.

### Component(s)

Continuous Integration

assignUser:
Apparently gha runners use these [azure machine](https://learn.microsoft.com/en-us/azure/virtual-machines/dv2-dsv2-series#dsv2-series). All listed CPUs support at least AVX2 (not 512 though which has caused issues before...) so this should be no issue.

PR #35117, opened 2023-04-13
GH-35115: [C++] Moved util_avx2.cc from acero to compute

### LINKED ISSUE #35115: [C++] util_avx2.cc is in acero but should be in compute

As part of the split to acero the file `arrow/compute/util.cc` and `arrow/util/compute/exec/util.cc` got moved around a bit.  It seems that `util_avx2.cc` ended up in acero but it's functions are actually needed by parquet.  This means, when compiling with AVX2, you will get undefined symbol errors if you try to use parquet.

### Component(s)

C++

westonpace:
### Rationale for this change

The file util_avx2.cc contains implementation for methods defined in `src/arrow/compute/util.cc` but it was placed in `src/arrow/acero`.  This leads to undefined symbol errors when compiling against AVX2.

The file util_avx2.cc is moved from `acero` to `compute`.

### Are these changes tested?

Regrettably, no.  See #35116 

No.
* Closes: #35115


icexelloss:
Sorry missed this - LGTM

PR #35120, opened 2023-04-13
GH-35118: [FlightSQL] Use `int32` to refer to 32-bit integers rather than `int`

### LINKED ISSUE #35118: [Format][FlightSQL] Spec refers to "int" type (which seems Java specific) rather than "int32"

While implement FlightSQL in our project, we found that the FlightSql.proto format reference refers to a type `int`

https://github.com/apache/arrow/blob/c03ca8f7c901eba498dc79b184e5a72a5149f236/format/FlightSql.proto#L1370

What `int` means was slightly unclear as the Arrow spec refers to "Int" types with `bitWidth` of 8, 16, 32 or 64 but the rust implementation has Int8, Int16, Int32 and Int64.

https://github.com/apache/arrow/blob/c03ca8f7c901eba498dc79b184e5a72a5149f236/format/Schema.fbs#L145-L148

## Go uses `Int32`:

https://github.com/apache/arrow/blob/c03ca8f7c901eba498dc79b184e5a72a5149f236/go/arrow/flight/flightsql/schema_ref/reference_schemas.go#L63

## C++ uses `Int32`:

https://github.com/apache/arrow/blob/c03ca8f7c901eba498dc79b184e5a72a5149f236/cpp/src/arrow/flight/sql/server.cc#L1144

## Java uses `INT`:

https://github.com/apache/arrow/blob/c03ca8f7c901eba498dc79b184e5a72a5149f236/java/flight/flight-sql/src/main/java/org/apache/arrow/flight/sql/FlightSqlProducer.java#L848

which is defined at:

https://github.com/apache/arrow/blob/c03ca8f7c901eba498dc79b184e5a72a5149f236/java/adapter/jdbc/src/test/java/org/apache/arrow/adapter/jdbc/h2/JdbcToArrowDataTypesTest.java#L90

According to [The Java Language Specification](http://java.sun.com/docs/books/jls/third_edition/html/j3TOC.html) from [Section 4.2: Primitive Types and Values](https://docs.oracle.com/javase/specs/jls/se20/html/jls-4.html#jls-4.2):

> The integral types are `byte`, `short`, `int`, and `long`, whose values are 8-bit, 16-bit, 32-bit and 64-bit signed two's-complement integers, respectively, and `char`, whose values are 16-bit unsigned integers representing UTF-16 code units 

Therefore, we can conclude that `INT` in arrow Java refers to 32-bit integer.

# Proposal

I propose that we change the proto spec to be consistent and use `int<bitwidth>` to refer to integer types. So specifically, change `int` to `int32` to match C++/Rust/Go as well as the convention in the rest of Flight.proto

Here is the reference to details of our discussion https://github.com/influxdata/influxdb_iox/pull/7546#discussion_r1165830088

### Component(s)

Format

appletreeisyellow:
### Rationale for this change

The spec is inconsistent -- see details on #35118 

Use `int32` to refer to 32-bit integers rather than `int`

### Are these changes tested?

No, only comments are changed

This clarifies a small corner case in the document

* Closes: https://github.com/apache/arrow/issues/35118
* Closes: #35118

### DIFF
--- a/format/FlightSql.proto
+++ b/format/FlightSql.proto
@@ -1271,7 +1271,7 @@ message CommandGetTableTypes {
  *  table_name: utf8 not null,
  *  column_name: utf8 not null,
  *  key_name: utf8,
- *  key_sequence: int not null
+ *  key_sequence: int32 not null
  * >
  * The returned data should be ordered by catalog_name, db_schema_name, table_name, key_name, then key_sequence.
  */
@@ -1321,7 +1321,7 @@ enum UpdateDeleteRules {
  *  fk_db_schema_name: utf8,
  *  fk_table_name: utf8 not null,
  *  fk_column_name: utf8 not null,
- *  key_sequence: int not null,
+ *  key_sequence: int32 not null,
  *  fk_key_name: utf8,
  *  pk_key_name: utf8,
  *  update_rule: uint1 not null,
@@ -1367,7 +1367,7 @@ message CommandGetExportedKeys {
  *  fk_db_schema_name: utf8,
  *  fk_table_name: utf8 not null,
  *  fk_column_name: utf8 not null,
- *  key_sequence: int not null,
+ *  key_sequence: int32 not null,
  *  fk_key_name: utf8,
  *  pk_key_name: utf8,
  *  update_rule: uint1 not null,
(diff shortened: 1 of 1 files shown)


alamb (approved):
Thanks @appletreeisyellow 

fyi @lidavidm and @kou   here is a similar one to https://github.com/apache/arrow/pull/35108

kou (approved):
+1

lidavidm (approved):
+1

Thank you!

alamb:
Related PR: https://github.com/apache/arrow/pull/35108

alamb:
Thanks again @appletreeisyellow

PR #35129, opened 2023-04-14
GH-35059: [C++] Fix "hash_count" for run-end encoded inputs

### LINKED ISSUE #35059: [C++] `hash_count` kernel miscounts when run-end encoded array contains null

Passing run-end encoded array in unit tests leads to these failures:

```cpp
[ RUN      ] GroupBy/GroupBy.CountOnly/0
/home/felipeo/code/arrow/cpp/src/arrow/testing/gtest_util.cc:100: Failure
Failed

@@ -0, +0 @@
-{key_0: 1, hash_count: 2}
+{key_0: 1, hash_count: 3}
@@ -2, +2 @@
-{key_0: 3, hash_count: 0}
+{key_0: 3, hash_count: 2}
Expected:
  -- is_valid: all not null
  -- child 0 type: int64
    [
      1,
      2,
      3,
      null
    ]
  -- child 1 type: int64
    [
      2,
      3,
      0,
      2
    ]
Actual:
  -- is_valid: all not null
  -- child 0 type: int64
    [
      1,
      2,
      3,
      null
    ]
  -- child 1 type: int64
    [
      3,
      3,
      2,
      2
    ]
    ```

### Component(s)

C++

### PR DESCRIPTION
**Pull request by felipecrv**
### Rationale for this change

Fixing a bug.

Changes to the `"hash_count"` kernel implementation to handle REE and union arrays correctly.

- [x] Generic (potentially slow) implementation
- [x] REE-specialized implementation

### Are these changes tested?

Yes, by modifying the existing unit tests.

* Closes: #35059

### DIFF
--- a/cpp/src/arrow/compute/kernels/hash_aggregate.cc
+++ b/cpp/src/arrow/compute/kernels/hash_aggregate.cc
@@ -45,6 +45,7 @@
 #include "arrow/util/cpu_info.h"
 #include "arrow/util/int128_internal.h"
 #include "arrow/util/int_util_overflow.h"
+#include "arrow/util/ree_util.h"
 #include "arrow/util/task_group.h"
 #include "arrow/util/tdigest.h"
 #include "arrow/util/thread_pool.h"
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by felipecrv** [comment]
@westonpace @jorisvandenbossche

File: cpp/src/arrow/testing/gtest_util.cc [hunk 1]
```diff
@@ -427,6 +428,24 @@ std::shared_ptr<Table> TableFromJSON(const std::shared_ptr<Schema>& schema,
...
+Result<std::shared_ptr<Table>> RunEndEncodeTableColumns(
+    const Table& table, const std::vector<int>& column_indices) {
+  const int num_columns = table.num_columns();
+  std::vector<std::shared_ptr<ChunkedArray>> encoded_columns;
+  encoded_columns.reserve(num_columns);
+  for (int i = 0; i < num_columns; i++) {
+    if (std::find(column_indices.begin(), column_indices.end(), i) !=
+        column_indices.end()) {
+      EXPECT_OK_AND_ASSIGN(auto run_end_encoded, compute::RunEndEncode(table.column(i)));
+      EXPECT_EQ(run_end_encoded.kind(), Datum::CHUNKED_ARRAY);
```
westonpace, line 440: Given you're returning a result anyways you could just use ARROW_ASSIGN_OR_RAISE and ARROW_RETURN_NOT_OK here right?

cpp/src/arrow/compute/kernels/hash_aggregate.cc
```diff
@@ -305,6 +306,56 @@ struct GroupedCountImpl : public GroupedAggregator {
...
+          }
+        } else {
+          if (!is_valid) {
+            for (int64_t i = 0; i < it.run_length(); ++i, ++g) {
+              counts[*g] += 1;
+            }
+          } else {
+            g += it.run_length();
+          }
+        }
```
westonpace, line 339: Can you collapse these two branches?

```
if (is_valid == count_valid) {
  ...
}
```

Or does that not somehow generate the same benefit?  Seems like either way there is a single if comparison.  I expect the compiler would generate the same for `if (x == true)` that it does for `if (x)`.

westonpace (changes requested):
One small nit for consistency in test routines to only use `EXPECT` if not returning a statusy thing.

cpp/src/arrow/testing/gtest_util.cc
felipecrv: Isn't this better for things used in tests? These are not really expected to fail, so these checks are more assert-like than regular error checking. I found this to be a pattern in this file.

cpp/src/arrow/compute/kernels/hash_aggregate.cc
felipecrv: Interesting catch. `count_valid` is a template param, so the compiler will generate the same code `if (is_valid)` in one instantiation and `if (!is_valid)` on another. No problem there. The final code is quite a logic exercise though :-) 

```cpp
          if (is_valid == count_valid) {
            for (int64_t i = 0; i < it.run_length(); ++i, ++g) {
              counts[*g] += 1;
            }
          } else {
            g += it.run_length();
          }
 ```

I will change it.

File: cpp/src/arrow/testing/gtest_util.cc (hunk 1 above)
Comment (westonpace) [reply cpp/src/arrow/testing/gtest_util.cc:440]: If that's the case then change the return type to `std::shared_ptr<Table>` and change the end to...

```
EXPECT_OK_AND_ASSIGN(auto table, Table::Make(...));
return table;
```

I don't really mind one pattern or the other.  It's just the mixing of patterns we should avoid.

cpp/src/arrow/compute/kernels/hash_aggregate.cc
felipecrv: Done.

cpp/src/arrow/testing/gtest_util.cc
felipecrv: Done. Pushed two new commits.

westonpace (approved):
Thanks for also extending the test to cover all the options as well.

PR #34919, opened 2023-04-06
GH-34918: [C++] Update vendored double-conversion 3.2.1

liujiacheng777:
Modifications based on doubleConversion version 3.2.1
* Closes: #34918

PR #35135, opened 2023-04-14
GH-35134: [C++] Add `arrow_vendored` namespace around double-conversion library

### LINKED ISSUE #35134: [C++] Place vendored double-conversion library in a private namespace

Double conversion is vendored in Arrow: 
* https://github.com/apache/arrow/blob/main/cpp/src/arrow/vendored/double-conversion/double-conversion.h#L33
* https://github.com/apache/arrow/blob/main/cpp/src/arrow/util/double_conversion.h#L26

This can cause symbol collisions when the Arrow library is linked into other libraries which use a different version of double-conversion.

The vendored version should be placed into its own namespace to avoid these symbol collisions.

This is a sub-issue of #34932

### Component(s)

C++

mikelui:
### Rationale for this change

See #35134

This change avoids potential symbol collisions when linking Arrow within other libraries.

This actually seems like the original intention (a private namespace) per the following merge: 767c953
The message in that merge is confusing, because it implies that the merge made the library private, when it actually did the opposite (specifically in 3b89b19). A quote is below:

>  Also, make its use private, because of Windows DLL exports. 

* an `update.sh` script to 
  * automatically pull a new version, 
  * apply the namespace change, 
  * apply a custom patch for seemingly Gandiva (15137e2)
    * this patch will break in future versions of double-conversion, but it is at least documented in the update script
* the update script is run for the current version 

The script is based on: https://github.com/apache/arrow/blob/main/cpp/src/arrow/vendored/fast_float/update.sh

### Are these changes tested?

Yes. Built and tests run

Only if users used to depend on using the internal vendored double-conversion libraries for their own code. In such a case, when they upgrade, they would have to specify the `arrow_vendored` namespace. 
* Closes: #35134

### DIFF
--- a/cpp/src/arrow/vendored/double-conversion/README.md
+++ b/cpp/src/arrow/vendored/double-conversion/README.md
@@ -17,7 +17,7 @@
   under the License.
 -->
 
-The files in this directory are vendored from double-conversion git tag v3.1.5.
+The files in this directory are vendored from double-conversion git tag v3.1.5
 
 To update, run:
 
--- a/cpp/src/arrow/vendored/double-conversion/README.md
+++ b/cpp/src/arrow/vendored/double-conversion/README.md
@@ -18,3 +18,18 @@
 -->
 
 The files in this directory are vendored from double-conversion git tag v3.1.5.
+
+To update, run:
+
+```
+./update.sh VERSION
+```
+
+For example:
+
+```
+./update.sh 3.1.5
+```
+
+If there are errors patching changes, you may have to manually intervene and resolve patch errors.
+If so, please update the patches so the modifications are explicit.
--- a/cpp/src/arrow/vendored/double-conversion/update.sh
+++ b/cpp/src/arrow/vendored/double-conversion/update.sh
@@ -1,0 +1,56 @@
+#!/usr/bin/env bash
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+set -eu
+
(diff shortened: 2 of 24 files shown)


cpp/src/arrow/vendored/double-conversion/README.md
```diff
@@ -17,4 +17,4 @@
   under the License.
 -->
 
-The files in this directory are vendored from double-conversion git tag v3.1.5.
+The files in this directory are vendored from double-conversion git tag v3.1.5
```
kou, line 20: Could you add how to use `update.sh`?

kou (commented):
Thanks!

Could you rebase on main for #35145?

I want us to upstream our changes instead of having our patches for easy to maintain.
Could you wait for a response from Gandiva developers? 
https://github.com/apache/arrow/pull/34919#issuecomment-1501420706

cpp/src/arrow/vendored/double-conversion/update.sh
```diff
@@ -0,0 +1,55 @@
...
+if [ "$#" -ne 1 ]; then
+  echo "Usage: $0 VERSION"
+  echo " e.g.: $0 3.2.1"
+  exit 1
+fi
+
+version="$1"
+
+pushd "${source_dir}"
+rm -rf fast_float
```
kou, line 33:
```suggestion
rm -rf double-conversion
```

cpp/src/arrow/vendored/double-conversion/update.sh
```diff
@@ -0,0 +1,55 @@
...
+git clone \
+    --branch "v${version}" \
+    --depth 1 \
+    https://github.com//google/double-conversion.git
+for f in double-conversion/double-conversion/*{.h,.cc}; do
+  mv "${f}" ./
+done
```
kou, line 40: It seems that we need to care about file remove case:

```suggestion
rm *.cc *.h
for f in double-conversion/double-conversion/*{.h,.cc}; do
  mv "${f}" ./
done
```

mikelui:
> Could you wait for a response from Gandiva developers? [#34919 (comment)](https://github.com/apache/arrow/pull/34919#issuecomment-1501420706)

This doesn't change the current status-quo of the patched changes, and instead just makes it explicit. Would it make sense to merge this now, and allow for the upstream changes for the version update?

The patch in the script would actually fail since those files don't exist in the newest version of double-conversion (the version being updated in that PR)

kou (approved):
+1

> Would it make sense to merge this now, and allow for the upstream changes for the version update?

OK. I see.

liujiacheng777:
@kou @mikelui I would like to update the double-conversion from v3.1.5 to v3.2.1. Can I upgrade based on the original v3.1.5 version and then you add arrow_vendored namespace

kou:
Could you do "add arrow_vendored namespace" by the script added by this pull request?

PR #35139, opened 2023-04-14
GH-35138: [Java] Fix ArrowFlightJdbcTimeStampVectorAccessor to deal with calendar

### LINKED ISSUE #35138: [Java] ArrowFlightJdbcTimeStampVectorAccessor does not deal with calendar correctly

The piece of code below in the `class ArrowFlightJdbcTimeStampVectorAccessor` does not deal with timezone correctly.

```java
  private LocalDateTime getLocalDateTime(Calendar calendar) {
    getter.get(getCurrentRow(), holder);
    this.wasNull = holder.isSet == 0;
    this.wasNullConsumer.setWasNull(this.wasNull);
    if (this.wasNull) {
      return null;
    }

    long value = holder.value;

    LocalDateTime localDateTime = this.longToLocalDateTime.fromLong(value);

    if (calendar != null) {
      TimeZone timeZone = calendar.getTimeZone();
      long millis = this.timeUnit.toMillis(value);
      localDateTime = localDateTime
          .minus(timeZone.getOffset(millis) - this.timeZone.getOffset(millis), ChronoUnit.MILLIS);
    }
    return localDateTime;
  }
```

It should call `localDateTime.plus` instead of `localDateTime.minus`.

I hit this issue when integrating this into our own JDBC implementation and found that the timestamp string in the `ResultSet` is wrong when converting between timezones.

### Component(s)

Java

wgtmac:
### Rationale for this change

The piece of code below in the `class ArrowFlightJdbcTimeStampVectorAccessor` does not deal with timezone correctly.

```java
  private LocalDateTime getLocalDateTime(Calendar calendar) {
    getter.get(getCurrentRow(), holder);
    this.wasNull = holder.isSet == 0;
    this.wasNullConsumer.setWasNull(this.wasNull);
    if (this.wasNull) {
      return null;
    }

    long value = holder.value;

    LocalDateTime localDateTime = this.longToLocalDateTime.fromLong(value);

    if (calendar != null) {
      TimeZone timeZone = calendar.getTimeZone();
      long millis = this.timeUnit.toMillis(value);
      localDateTime = localDateTime
          .minus(timeZone.getOffset(millis) - this.timeZone.getOffset(millis), ChronoUnit.MILLIS);
    }
    return localDateTime;
  }
```

I hit this issue when integrating this into our own JDBC implementation and found that the timestamp string in the `ResultSet` is wrong when converting between timezones.

It should call `localDateTime.plus` instead of `localDateTime.minus`.

### Are these changes tested?

Actually these conversions are covered by plenty of tests but unfortunately they share the same issues. This commit fixes them all.

No.
* Closes: #35138

### DIFF
--- a/java/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcTimeStampVectorAccessor.java
+++ b/java/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcTimeStampVectorAccessor.java
@@ -96,7 +96,7 @@ public class ArrowFlightJdbcTimeStampVectorAccessor extends ArrowFlightJdbcAcces
       TimeZone timeZone = calendar.getTimeZone();
       long millis = this.timeUnit.toMillis(value);
       localDateTime = localDateTime
-          .minus(timeZone.getOffset(millis) - this.timeZone.getOffset(millis), ChronoUnit.MILLIS);
+          .plus(timeZone.getOffset(millis) - this.timeZone.getOffset(millis), ChronoUnit.MILLIS);
     }
     return localDateTime;
   }
--- a/java/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/utils/DateTimeUtils.java
+++ b/java/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/utils/DateTimeUtils.java
@@ -36,7 +36,7 @@ public class DateTimeUtils {
   }
 
   /**
-   * Subtracts given Calendar's TimeZone offset from epoch milliseconds.
+   * Adds given Calendar's TimeZone offset from epoch milliseconds.
    */
   public static long applyCalendarOffset(long milliseconds, Calendar calendar) {
     if (calendar == null) {
@@ -47,7 +47,7 @@ public class DateTimeUtils {
     final TimeZone defaultTz = TimeZone.getDefault();
 
     if (tz != defaultTz) {
-      milliseconds -= tz.getOffset(milliseconds) - defaultTz.getOffset(milliseconds);
+      milliseconds += tz.getOffset(milliseconds) - defaultTz.getOffset(milliseconds);
     }
 
     return milliseconds;
--- a/java/flight/flight-sql-jdbc-core/src/test/java/org/apache/arrow/driver/jdbc/accessor/impl/text/ArrowFlightJdbcVarCharVectorAccessorTest.java
+++ b/java/flight/flight-sql-jdbc-core/src/test/java/org/apache/arrow/driver/jdbc/accessor/impl/text/ArrowFlightJdbcVarCharVectorAccessorTest.java
@@ -510,14 +510,23 @@ public class ArrowFlightJdbcVarCharVectorAccessorTest {
     Text value = new Text("2021-07-02");
     when(getter.get(0)).thenReturn(value.copyBytes());
 
-    Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("America/Sao_Paulo"));
-    Date result = accessor.getDate(calendar);
+    {
+      Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("America/Sao_Paulo"));
+      Date result = accessor.getDate(calendar);
+      calendar.setTime(result);
 
-    calendar = Calendar.getInstance(TimeZone.getTimeZone("Etc/UTC"));
-    calendar.setTime(result);
+      collector.checkThat(dateTimeFormat.format(calendar.getTime()),
+          equalTo("2021-07-01T21:00:00.000Z"));
+    }
 
-    collector.checkThat(dateTimeFormat.format(calendar.getTime()),
-        equalTo("2021-07-02T03:00:00.000Z"));
+    {
+      Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("Etc/UTC"));
+      Date result = accessor.getDate(calendar);
+      calendar.setTime(result);
+
+      collector.checkThat(dateTimeFormat.format(calendar.getTime()),
+          equalTo("2021-07-02T00:00:00.000Z"));
+    }
   }
 
   @Test
@@ -547,13 +556,21 @@ public class ArrowFlightJdbcVarCharVectorAccessorTest {
     Text value = new Text("02:30:00");
     when(getter.get(0)).thenReturn(value.copyBytes());
 
-    Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("America/Sao_Paulo"));
-    Time result = accessor.getTime(calendar);
+    {
+      Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("America/Sao_Paulo"));
+      Time result = accessor.getTime(calendar);
+      calendar.setTime(result);
 
-    calendar = Calendar.getInstance(TimeZone.getTimeZone("Etc/UTC"));
-    calendar.setTime(result);
+      collector.checkThat(timeFormat.format(calendar.getTime()), equalTo("23:30:00.000Z"));
+    }
 
-    collector.checkThat(timeFormat.format(calendar.getTime()), equalTo("05:30:00.000Z"));
+    {
+      Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("Etc/UTC"));
+      Time result = accessor.getTime(calendar);
+      calendar.setTime(result);
+
+      collector.checkThat(timeFormat.format(calendar.getTime()), equalTo("02:30:00.000Z"));
+    }
   }
 
   @Test
@@ -584,14 +601,23 @@ public class ArrowFlightJdbcVarCharVectorAccessorTest {
     Text value = new Text("2021-07-02 02:30:00.000");
     when(getter.get(0)).thenReturn(value.copyBytes());
 
-    Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("America/Sao_Paulo"));
-    Timestamp result = accessor.getTimestamp(calendar);
+    {
+      Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("America/Sao_Paulo"));
+      Timestamp result = accessor.getTimestamp(calendar);
+      calendar.setTime(result);
 
-    calendar = Calendar.getInstance(TimeZone.getTimeZone("Etc/UTC"));
-    calendar.setTime(result);
+      collector.checkThat(dateTimeFormat.format(calendar.getTime()),
+          equalTo("2021-07-01T23:30:00.000Z"));
+    }
 
-    collector.checkThat(dateTimeFormat.format(calendar.getTime()),
-        equalTo("2021-07-02T05:30:00.000Z"));
+    {
+      Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("Etc/UTC"));
+      Timestamp result = accessor.getTimestamp(calendar);
+      calendar.setTime(result);
+
+      collector.checkThat(dateTimeFormat.format(calendar.getTime()),
+          equalTo("2021-07-02T02:30:00.000Z"));
+    }
   }
 
   private void assertGetBoolean(Text value, boolean expectedResult) throws SQLException {
--- a/java/flight/flight-sql-jdbc-core/src/test/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcDateVectorAccessorTest.java
+++ b/java/flight/flight-sql-jdbc-core/src/test/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcDateVectorAccessorTest.java
@@ -130,7 +130,7 @@ public class ArrowFlightJdbcDateVectorAccessorTest {
 
       long offset = timeZone.getOffset(resultWithoutCalendar.getTime());
 
-      collector.checkThat(resultWithoutCalendar.getTime() - result.getTime(), is(offset));
+      collector.checkThat(result.getTime() - resultWithoutCalendar.getTime(), is(offset));
       collector.checkThat(accessor.wasNull(), is(false));
     });
   }
@@ -160,7 +160,7 @@ public class ArrowFlightJdbcDateVectorAccessorTest {
 
       long offset = timeZone.getOffset(resultWithoutCalendar.getTime());
 
-      collector.checkThat(resultWithoutCalendar.getTime() - result.getTime(), is(offset));
+      collector.checkThat(result.getTime() - resultWithoutCalendar.getTime(), is(offset));
       collector.checkThat(accessor.wasNull(), is(false));
     });
   }
--- a/java/flight/flight-sql-jdbc-core/src/test/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcTimeStampVectorAccessorTest.java
+++ b/java/flight/flight-sql-jdbc-core/src/test/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcTimeStampVectorAccessorTest.java
@@ -182,7 +182,7 @@ public class ArrowFlightJdbcTimeStampVectorAccessorTest {
       long offset = timeZone.getOffset(resultWithoutCalendar.getTime()) -
           timeZoneForVector.getOffset(resultWithoutCalendar.getTime());
 
-      collector.checkThat(resultWithoutCalendar.getTime() - result.getTime(), is(offset));
+      collector.checkThat(result.getTime() - resultWithoutCalendar.getTime(), is(offset));
       collector.checkThat(accessor.wasNull(), is(false));
     });
   }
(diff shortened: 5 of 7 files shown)


wgtmac:
@lidavidm @davisusanibar @jduo Could you please take a look?

java/flight/flight-sql-jdbc-core/src/test/java/org/apache/arrow/driver/jdbc/accessor/impl/text/ArrowFlightJdbcVarCharVectorAccessorTest.java
```diff
@@ -510,14 +510,23 @@ public void testShouldGetDateReturnValidDateWithCalendar() throws Exception {
...
-    Date result = accessor.getDate(calendar);
+    {
+      Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("America/Sao_Paulo"));
+      Date result = accessor.getDate(calendar);
+      calendar.setTime(result);
 
-    calendar = Calendar.getInstance(TimeZone.getTimeZone("Etc/UTC"));
-    calendar.setTime(result);
+      collector.checkThat(dateTimeFormat.format(calendar.getTime()),
+          equalTo("2021-07-01T21:00:00.000Z"));
```
lidavidm, line 519: OK. So the assumption is that the underlying value is in UTC, and are converting to the destination time zone?

Derby does something different, though:

```java
    SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSXXX");
    try (final Connection connection = DriverManager.getConnection("jdbc:derby:memory:foo;create=true")) {
      try (final Statement statement = connection.createStatement()) {
        statement.executeUpdate("CREATE TABLE foo (bar DATE)");
        statement.executeUpdate("INSERT INTO foo (bar) VALUES '2021-07-02'");
        try (final ResultSet rs = statement.executeQuery("SELECT * FROM foo")) {
          assertThat(rs.next()).isTrue();
          Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("America/Sao_Paulo"));
          final Date date = rs.getDate(1, calendar);
          System.out.println(date);
          calendar.setTime(date);
          System.out.println(sdf.format(calendar.getTime()));
        }
      }
    }
```

Derby appears to just attach the timezone to the value, instead of adjusting it from an assumed source time zone. The output is:

```
2021-07-02
2021-07-02T12:00:00.000+09:00
```

wgtmac: It depends on the whether the timezone field is set in the TimeStampVector.
- If timezone is not set, it assumes UTC and converts from that.
- If timezone is set, it converts from the provided timezone.

If user does not request a different timezone, then the result is good. Otherwise the problem emerges.

wgtmac:
CI failure is unrelated.

```
[ RUN      ] AggregateSchema.SingleKeyAndSegmentKey
/arrow/cpp/src/arrow/testing/gtest_util.cc:241: Failure
Value of: left.Equals(right, check_metadata)
  Actual: false
Expected: true
schemas 'y: int32
z: int32
hash_count: int64' and 'z: int32
y: int32
hash_count: int64' should have compared equal
[  FAILED  ] AggregateSchema.SingleKeyAndSegmentKey (0 ms)
[----------] 6 tests from AggregateSchema (13 ms total)
```

java/flight/flight-sql-jdbc-core/src/test/java/org/apache/arrow/driver/jdbc/accessor/impl/text/ArrowFlightJdbcVarCharVectorAccessorTest.java
lidavidm: > If timezone is not set, it assumes UTC and converts from that to target timezone.

This seems a bit off though, since we don't _know_ the timezone in that case specifically. (Not that the current behavior seems any better!) It seems like the right behavior here would be to either treat this as an error, or behave like Derby and treat the user-supplied calendar to mean, "assume the timestamp is a timestamp as viewed in this time zone".

wgtmac: Here is the code where it assumes UTC if timezone is missing from the vector: https://github.com/apache/arrow/blob/eaa1a1ea74e2b37020e0133f1460cb251e5f4974/java/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcTimeStampVectorAccessor.java#L180

Yes, I agree with you that we shouldn't blindly assume timezone to be UTC if unset. In Apache Hive, this usually means timestamp without timezone where the values are assumed to be in the default timezone in the JVM (which is the local timezone in the usual case).

wgtmac: IMO, any behavior change may need a discussion as it will affect downstream users. This patch only focuses on the bug fix.

lidavidm: Fair, though arguably this is replacing buggy behavior with buggy behavior 🙂 

Let's put a caveat in the documentation page for the driver? 

We should consider aligning with what Derby appears to do, too. Either that, or check a few other drivers to see the expected behavior here.

I'm still not sure the conversion here is correct, though. From the docs, java.sql.Date is the millisecond value (in GMT) of midnight of a date in some time zone. 

So if we have the date "2021-07-02" in America/Sao_Paulo, then the GMT timestamp [should be 2021-07-02 03:00](https://www.wolframalpha.com/input?i=2021-07-02+00%3A00+Sao+Paulo+in+GMT). So the new behavior doesn't seem correct and the old behavior is what's expected.

Or if we are assuming that the date is "2021-07-02" in UTC, and converting that to Sao Paulo, that makes no sense - Date isn't supposed to store a time, so we shouldn't treat it as one.

wgtmac: The `accessor.getDate(calendar)` above is to implement `ResultSet.getDate(int, Calendar)` and returns a `java.sq.Date`. Given two facts:
- `java.sql.Date`: A milliseconds value represents the number of milliseconds that have passed since January 1, 1970 00:00:00.000 GMT. To conform with the definition of SQL DATE, the millisecond values must be 'normalized' by setting the hours, minutes, seconds, and milliseconds to zero in the particular time zone with which the instance is associated.
- `ResultSet.getDate(int, Calendar)`: Retrieves the value of the designated column in the current row of this ResultSet object as a java.sql.Date object in the Java programming language. This method uses the given calendar to construct an appropriate millisecond value for the date if the underlying database does not store timezone information.

Therefore I agree with you that we should assume the underlying values we are reading do not have timezone information and the behavior would be much easier to understand if we simply attach the user-supplied timezone to the values to return.

The original buggy code that this PR tries to solve actually is not here. The modification here was a result of that fix. Please see my other comments above. Let me double check if there is something wrong.

java/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcTimeStampVectorAccessor.java
```diff
@@ -96,7 +96,7 @@ private LocalDateTime getLocalDateTime(Calendar calendar) {
       TimeZone timeZone = calendar.getTimeZone();
       long millis = this.timeUnit.toMillis(value);
       localDateTime = localDateTime
-          .minus(timeZone.getOffset(millis) - this.timeZone.getOffset(millis), ChronoUnit.MILLIS);
+          .plus(timeZone.getOffset(millis) - this.timeZone.getOffset(millis), ChronoUnit.MILLIS);
```
wgtmac, line 99: The original buggy code I want to solve is here. It tries to convert local date time from vector timezone to user-supplied timezone. We should add the timezone offset but not subtract it. A number of test cases need to be fixed as their expected values are computed with the same issue.

java/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/utils/DateTimeUtils.java
```diff
@@ -47,7 +47,7 @@ public static long applyCalendarOffset(long milliseconds, Calendar calendar) {
     final TimeZone defaultTz = TimeZone.getDefault();
 
     if (tz != defaultTz) {
-      milliseconds -= tz.getOffset(milliseconds) - defaultTz.getOffset(milliseconds);
+      milliseconds += tz.getOffset(milliseconds) - defaultTz.getOffset(milliseconds);
```
wgtmac, line 50: The problem is that I made this line of change in order to fix the test failures of this case: https://github.com/apache/arrow/blob/eaa1a1ea74e2b37020e0133f1460cb251e5f4974/java/flight/flight-sql-jdbc-core/src/test/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcTimeStampVectorAccessorTest.java#L292 and let me paste the code snippet of that case below:

```java
  @Test
  public void testShouldGetStringBeConsistentWithVarCharAccessorWithCalendar() throws Exception {
    // Ignore for TimeStamp vectors with TZ, as VarChar accessor won't consider their TZ
    Assume.assumeTrue(
        vector instanceof TimeStampNanoVector || vector instanceof TimeStampMicroVector ||
            vector instanceof TimeStampMilliVector || vector instanceof TimeStampSecVector);
    Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone(AMERICA_VANCOUVER));
    assertGetStringIsConsistentWithVarCharAccessor(calendar);
  }

  private void assertGetStringIsConsistentWithVarCharAccessor(Calendar calendar) throws Exception {
    try (VarCharVector varCharVector = new VarCharVector("",
        rootAllocatorTestRule.getRootAllocator())) {
      varCharVector.allocateNew(1);
      ArrowFlightJdbcVarCharVectorAccessor varCharVectorAccessor =
          new ArrowFlightJdbcVarCharVectorAccessor(varCharVector, () -> 0, (boolean wasNull) -> {
          });

      accessorIterator.iterate(vector, (accessor, currentRow) -> {
        final String string = accessor.getString();
        varCharVector.set(0, new Text(string));
        varCharVector.setValueCount(1);

        Timestamp timestampFromVarChar = varCharVectorAccessor.getTimestamp(calendar);
        Timestamp timestamp = accessor.getTimestamp(calendar);

        collector.checkThat(timestamp, is(timestampFromVarChar));
        collector.checkThat(accessor.wasNull(), is(false));
      });
    }
  }
```

wgtmac: In the test above, the user-supplied timezone is `AMERICA_VANCOUVER`. When the input text to varCharVector is `2021-07-09 00:00:00.0`, varCharVectorAccessor.getTimestamp(calendar) returns `2021-07-09 07:00:00.0`. This does not make sense to me at all. @lidavidm

wgtmac: To summarize, we have two bugs for now.
- conversion between time zones are incorrect.
- It assumes utc if time zone is not set in the timestamp vector. Same for varchar vector if it tries to interpret as timestamp values.

For the fix of HSBC implementation of getDate/getTimestamp with a calendar as input, I’d propose following:
- convert between timezones if timestamp vector has set timezone and user supplied calendar is not null.
- Throw unsupported operation exception if vector does not have timezone set and user has provided a timezone. The reason is that the returned timestamp object cannot carry timezone info.

IIUC, the timezone info in your derby example comes from the SimpleDateFormat but not the resultset itself.


WDYT? @lidavidm

java/flight/flight-sql-jdbc-core/src/test/java/org/apache/arrow/driver/jdbc/accessor/impl/text/ArrowFlightJdbcVarCharVectorAccessorTest.java
lidavidm: > The original buggy code that this PR tries to solve actually is not here. 

Right, I'm just trying to use this to understand what's happening here.

java/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/utils/DateTimeUtils.java
lidavidm: > varCharVectorAccessor.getTimestamp(calendar) returns 2021-07-09 07:00:00.0. This does not make sense to me at all.

That's [midnight in Vancouver expressed as GMT](https://www.wolframalpha.com/input?i=2021-07-09+07%3A00%3A00+GMT+in+Vancouver) so it's consistent with the other case described above. 

> conversion between time zones are incorrect.

It seems unintuitive, but at least consistent: it will take that date at midnight in the given time zone, and express it in GMT, which seems to match the docstring of java.sql.Date.

> Throw unsupported operation exception if vector does not have timezone set and user has provided a timezone. The reason is that the returned timestamp object cannot carry timezone info. Or return the timestamp as is to pretend it is in the user supplied timezone?

That may also work, but I suppose if the user is explicitly passing a Calendar, the current behavior is what they actually want? This is more a flaw of the original Java date/time APIs in that Date itself doesn't carry the timezone so the display value is weird.

wgtmac: Right, at least these behaviors are consistent to each other. It is unintuitive since timestamp values are always returned in the GMT regardless of what calendar is supplied. I cannot say this is wrong because it seems to be aligned with the SQL `TIMESTAMP_WITH_TIMEZONE` type.

There was a [discussion](https://docs.google.com/document/d/1gNRww9mZJcHvUDCXklzjFEQGpefsuR_akCDfWsdE35Q) in the Hadoop community to work on the consistency of timestamp semantics across different SQL-on-Hadoop engines. It proposed a `TIMESTAMP_WITH_LOCAL_TIMEZONE` type in addition to the SQL `TIMESTAMP_WITHOUT_TIMEZONE` and `TIMESTAMP_WITH_TIMEZONE` types:
- SQL `TIMESTAMP_WITHOUT_TIMEZONE`: Always display the same wall clock time regardless of timezone.
- SQL `TIMESTAMP_WITH_TIMEZONE`: The timestamp value is stored in a specific timezone (can be non-UTC) and can be converted to any timezone for display.
- New `TIMESTAMP_WITH_LOCAL_TIMEZONE`: The timestamp value is stored in the UTC and always display it in the local timezone.

Most of the SQL engines in the aforementioned doc now support `TIMESTAMP_WITH_LOCAL_TIMEZONE` type and so do the open file formats like the Apache Parquet and Apache Orc. I made this PR to align the behavior with `TIMESTAMP_WITH_LOCAL_TIMEZONE` type. Nevertheless I am no an expert on JDBC so I am afraid that I may not know what is the expected behavior from the end users. 

Please feel free to close this PR if you think the current behavior is not an issue, or involve more people to the discussion if a fix is required. Thanks for the discussion in detail!

lidavidm: Thanks for bearing with me. The mess of date/time APIs here is confusing...

For Arrow: `timestamp[s, TZ]` appears equivalent to `TIMESTAMP_WITH_LOCAL_TIMEZONE`. Arrow always stores the underlying value in UTC. `timestamp[s]` appears to be `TIMESTAMP_WITHOUT_TIMEZONE`. Arrow does not have an equivalent to `TIMESTAMP_WITH_TIMEZONE`. 

So the question here is still the semantics of `getDate(int, Calendar)`. The method looks ill-defined to me if `java.sql.Date` is not supposed to carry time or timezone information in the first place...The API docs say "This method uses the given calendar to construct an appropriate millisecond value for the date if the underlying database does not store timezone information."

I do see things like this: https://stackoverflow.com/questions/9202857/timezones-in-sql-date-vs-java-sql-date

> This means that when you work in UTC+1 and ask the database for a DATE a compliant implementation does exactly what you've observed: return a java.sql.Date with a milliseconds value that corresponds to the date in question at 00:00:00 UTC+1 independently of how the data got to the database in the first place.

So that also seems to match the existing behavior, by my reading.

I looked at Postgres's JDBC driver: 

https://github.com/pgjdbc/pgjdbc/blob/3c3e5b62349c42d9d3dfe1fe0102776d4de8c353/pgjdbc/src/main/java/org/postgresql/jdbc/PgResultSet.java#L532

and

https://github.com/pgjdbc/pgjdbc/blob/3c3e5b62349c42d9d3dfe1fe0102776d4de8c353/pgjdbc/src/main/java/org/postgresql/jdbc/TimestampUtils.java#L639-L653

> Truncate date part so in given time zone the date would be formatted as 00:00

I think that also matches with the current behavior, as unintuitive as it is.

wgtmac: Thanks for the additional links! It is really a notorious topic :)

PR #35145, opened 2023-04-14
GH-35144: [C++] Fix a unit test broken when the output order of the aggregate node changed

### LINKED ISSUE #35144: [C++] Unit test regression introduced by #34931

#34931 changed the order of of the aggregate node.  There was one unit test that was not updated with the new order and is now failing.  The new unit test was introduced recently in #34904 so I think the PR for #34931 probably wasn't rebased in the last day or two and so missed the check.

### Component(s)

C++

westonpace:
### Rationale for this change

The unit test was broken by a recent change and needed to be updated.

The unit test has been updated to expect the correct order.

### Are these changes tested?

Yes

No
* Closes: #35144


icexelloss (approved):

icexelloss:
Thanks @westonpace - I found the same issue. I thought all tests passed before I merged it but I must have missed somehow. Thanks for fixing!

westonpace:
> I thought all tests passed before I merged it but I must have missed somehow. 

They probably did all pass but they ran before #34904 was merged in.  This kind of thing happens every so often when two PRs conflict with each other.

PR #14198, opened 2022-09-21
ARROW-17349: [C++] Allow casting map types

wjones127:
* Implements casting functions from `MapType` -> `[MapType, list<struct>, large_list<struct>]`, with support for mapping the field names to new ones.
* Adds a new `Datum::View()` method to map an array to a new type.
* Use `Datum::View()` in early return of cast dispatching (if types are "equal"). Right now this is useful for mapping field names for map arrays, but it could also be used for list types once we enable some way to check if list types are equal except for its field name. See also: https://issues.apache.org/jira/browse/ARROW-14999

PR #35146, opened 2023-04-14
GH-29781: [C++][Parquet] Switch to use compliant nested types by default

### LINKED ISSUE #29781: [C++][Parquet] Default to compliant nested types in Parquet writer
In C++ there is already an option to get the "compliant_nested_types" (to have the list columns follow the Parquet specification), and ARROW-11497 exposed this option in Python.

This is still set to False by default, but in the source it says "TODO: At some point we should flip this.", and in ARROW-11497 there was also some discussion about what it would take to change the default.

cc @emkornfield @pitrou

**Reporter**: [Will Jones](https://issues.apache.org/jira/browse/ARROW-14196) / @wjones127
**Assignee**: [Will Jones](https://issues.apache.org/jira/browse/ARROW-14196) / @wjones127
#### Related issues:
- [[Python][Parquet] Allow to select columns of a list field without requiring the list component names](https://github.com/apache/arrow/issues/29862) (relates to)
- [[Python] pyarrow parquet writer for list does not conform with Apache Parquet specification](https://github.com/apache/arrow/issues/27376) (is related to)
#### PRs and other links:
- [GitHub Pull Request #14651](https://github.com/apache/arrow/pull/14651)

<sub>**Note**: *This issue was originally created as [ARROW-14196](https://issues.apache.org/jira/browse/ARROW-14196). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

wjones127:
### Rationale for this change

This has been a long-standing TODO.

* Closes: #29781

### DIFF
--- a/cpp/src/parquet/properties.h
+++ b/cpp/src/parquet/properties.h
@@ -934,7 +934,7 @@ class PARQUET_EXPORT ArrowWriterProperties {
     /// \brief When enabled, will not preserve Arrow field names for list types.
     ///
     /// Instead of using the field names Arrow uses for the values array of
-    /// list types (default "item"), will use "entries", as is specified in
+    /// list types (default "item"), will use "element", as is specified in
     /// the Parquet spec.
     ///
     /// This is enabled by default.
--- a/cpp/src/parquet/properties.h
+++ b/cpp/src/parquet/properties.h
@@ -879,8 +879,7 @@ class PARQUET_EXPORT ArrowWriterProperties {
           coerce_timestamps_unit_(::arrow::TimeUnit::SECOND),
           truncated_timestamps_allowed_(false),
           store_schema_(false),
-          // TODO: At some point we should flip this.
-          compliant_nested_types_(false),
+          compliant_nested_types_(true),
           engine_version_(V2),
           use_threads_(kArrowDefaultUseThreads),
           executor_(NULLPTR) {}
--- a/cpp/src/parquet/properties.h
+++ b/cpp/src/parquet/properties.h
@@ -937,13 +937,13 @@ class PARQUET_EXPORT ArrowWriterProperties {
     /// list types (default "item"), will use "entries", as is specified in
     /// the Parquet spec.
     ///
-    /// This is disabled by default, but will be enabled by default in future.
+    /// This is enabled by default.
     Builder* enable_compliant_nested_types() {
       compliant_nested_types_ = true;
       return this;
     }
 
-    /// Preserve Arrow list field name (default behavior).
+    /// Preserve Arrow list field name.
     Builder* disable_compliant_nested_types() {
       compliant_nested_types_ = false;
       return this;
--- a/cpp/src/parquet/arrow/arrow_schema_test.cc
+++ b/cpp/src/parquet/arrow/arrow_schema_test.cc
@@ -988,7 +988,7 @@ TEST_F(TestConvertArrowSchema, ParquetLists) {
   //   }
   // }
   {
-    auto element = PrimitiveNode::Make("string", Repetition::OPTIONAL,
+    auto element = PrimitiveNode::Make("element", Repetition::OPTIONAL,
                                        ParquetType::BYTE_ARRAY, ConvertedType::UTF8);
     auto list = GroupNode::Make("list", Repetition::REPEATED, {element});
     parquet_fields.push_back(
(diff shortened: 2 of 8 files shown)


cpp/src/parquet/properties.h
```diff
@@ -819,8 +819,7 @@ class PARQUET_EXPORT ArrowWriterProperties {
           coerce_timestamps_unit_(::arrow::TimeUnit::SECOND),
           truncated_timestamps_allowed_(false),
           store_schema_(false),
-          // TODO: At some point we should flip this.
```
mapleFU, line 822:
```
/// This is disabled by default, but will be enabled by default in future.
```

Should we change this in Loc 880?

wgtmac: The line 320 in the parquet.rst doc should also be changed.
```
   std::shared_ptr<ArrowWriterProperties> arrow_props = ArrowWriterProperties::Builder()
      .enable_deprecated_int96_timestamps() // default False
      ->store_schema() // default False
      ->enable_compliant_nested_types() // default False
      ->build();

```

mapleFU (approved):
```java
public abstract class ConversionPatterns {

  static final String MAP_REPEATED_NAME = "key_value";
  private static final String ELEMENT_NAME = "element";
```

Go through the Java code, seems that it also uses "element". LGTM

mapleFU:
In C++, code shows `force`. But when I go through code in arrow-rs, it doesn't say is force. Maybe we should ask the parquet maillist later?

wjones127:
> In C++, code shows `force`. But when I go through code in arrow-rs, it doesn't say is force. And though parquet-format use "element" in description, it doesn't says it's forced. Maybe we should ask the parquet maillist later?

The rules were laid out earlier here:

https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#backward-compatibility-rules

Before this PR I did work to make sure that the Arrow C++ implementation (1) didn't care about the field names in equality comparison ([PR](https://github.com/apache/arrow/pull/14847)) and (2) could cheaply cast between types that differed only in field names ([PR](https://github.com/apache/arrow/pull/14198)).

mapleFU:
Thanks! So, this flag could make converting and reading a bit cheaper?

wgtmac:
Is it time to revive this PR and check it in? @wjones127

wgtmac (approved):
LGTM! cc @pitrou

cpp/src/parquet/properties.h
```diff
@@ -938,13 +937,13 @@ class PARQUET_EXPORT ArrowWriterProperties {
     /// list types (default "item"), will use "entries", as is specified in
```
pitrou, line 937: It seems this docstring is incorrect as per the Parquet spec, or am I reading it wrong?
```suggestion
    /// list types (default "item"), will use "element", as is specified in
```

See https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists

mapleFU: Oops.

```c++
MapType::MapType(std::shared_ptr<DataType> key_type, std::shared_ptr<Field> item_field,
                 bool keys_sorted)
    : MapType(::arrow::field("key", std::move(key_type), false), std::move(item_field),
              keys_sorted) {}

MapType::MapType(std::shared_ptr<Field> key_field, std::shared_ptr<Field> item_field,
                 bool keys_sorted)
    : MapType(
          ::arrow::field("entries",
                         struct_({std::move(key_field), std::move(item_field)}), false),
          keys_sorted) {}
```

Found `"entries"` in `arrow/type.cc`, and spec is using "element". 

And here maybe we're using "entries":

```c++
/// In a field with Map type, the field has a child Struct field, which then
/// has two children: key type and the second the value type. The names of the
/// child fields may be respectively "entries", "key", and "value", but this is
/// not enforced.
///
/// Map
/// ```text
///   - child[0] entries: Struct
///     - child[0] key: K
///     - child[1] value: V
/// ```
/// Neither the "entries" field nor the "key" field may be nullable.
```

According to spec. Map should use "key_value". Seems it's incorrect? Am I right?

wjones127: You are right, that should be "element".

Parquet will always have them as "element". Arrow has it's own spec, but isn't prescriptive for the name of the list type. Either way, I don't think we should care too much about this detail; these field names should be pretty much invisible to users.

mapleFU (approved):

pitrou (approved):

Issue #35049, opened 2023-04-11
[R] Improve configure script for contributor experience


paleolimbot:
### Describe the enhancement requested

There are a few things that I've noticed over the last few months of working with/installing Arrow and the R package on various platforms. I think some of these may have come up in previous tickets and as I go through actually trying some of these I will attempt to locate them.

The main idea is that when a contributor does a git clone of Arrow and opens up the package in RStudio or VScode, the static nightly builds of Arrow built for the R package should be sufficient for a user to run `devtools:test()`. When I was debugging the "files are kept open too long on Windows" problem I used that strategy to avoid setting up a Windows Arrow build and it worked great! (Maybe this is even documented already...if it is, moving that documentation to a more prominent place may be helpful).

Another common thing I've run up against is that the configure script enables `pkg-config` by default. The build of Arrow resolved by pkg-config is almost never the right one...it's usually too old or has the wrong features enabled. At the very least we need to check the version of whatever gets resolved by pkg-config and maybe we should consider disabling it by default. This is a particular issue with homebrew.

### Component(s)

R

nealrichardson:
I think the homebrew issue is separate from pkg-config. The brew check is after the pkg-config path: https://github.com/apache/arrow/blob/main/r/configure#L128 and unlike the pkg-config block above it, there is no check/warning about version mismatch. We should probably only use the homebrew library if there is a version match. Probably also ok if we wanted to make homebrew apache-arrow opt-in via environment variable.

Possibly unrelated and just FYI, we do use `pkg-config` elsewhere (later) in configure to read the `.pc` files to know which other `-l` lib flags need to be added. This is used even in the bundled static build.

paleolimbot:
When I'm going through this for real I will double check, but on my machine I have `pkg-config` on PATH, which a little too happily resolves to the homebrew install of Arrow. This was the issue Lionel had when attempting to do the dplyr PR.

There is an open ticket somewhere about using "arrow" vs "arrow-dataset" as the pkg-config name, too (arrow-dataset is more likely to resolve to the feature set we need).

nealrichardson:
Sure, I believe you (and I'm not going to `brew install apache-arrow` and mess my env up to find out 😹). I don't think anything has ever been removed from the `configure` script, only added, so there's lots of if-elses in there and it's possible some aren't being reached often or ever. It's ripe for a (careful) refactor, for sure.

nealrichardson:
Related: #35140

nealrichardson:
[...]

Issue #34523, opened 2023-03-09
[R] Unable to compile R package with GCS on macOS M1


amoeba:
### Describe the bug, including details regarding any error messages, version, and platform.

I'm encountering an issue when trying to build the R package from a local build of Arrow C++ but only when I have GCS enabled. The error happens when I run `R CMD INSTALL --preclean --no-multiarch .`:

> symbol not found in flat namespace '__ZN4absl12lts_2023012510FormatTimeENSt3__117basic_string_viewIcNS1_11char_traitsIcEEEENS0_4TimeENS0_8TimeZoneE

This looked very similar to what was reported in https://github.com/apache/arrow/issues/32487, though I'm not building tests. I first tried setting `CMAKE_CXX_STANDARD=17` which didn't help. @nealrichardson recommended `absl_SOURCE=BUNDLED` which also didn't help. I do have abseil via Homebrew.

I'm including some details of how I'm building everything below. I'm building off of main (2f3f41f05e4bfae24344ee7a4b39ad6719f98751):

<details>
<summary>How I'm building libarrow and the R package</summary>

```
$ echo $ARROW_HOME
/Users/bryce/builds/arrow-arm64

$ pwd
/Users/bryce/src/apache/arrow/cpp/build

$ cmake .. \
      -GNinja \
      -DARROW_COMPUTE=ON \
      -DARROW_CSV=ON \
      -DARROW_DATASET=ON \
      -DARROW_FILESYSTEM=ON \
      -DARROW_JSON=ON \
      -DARROW_PARQUET=ON \
      -DARROW_INSTALL_NAME_RPATH=OFF \
      -DARROW_S3=ON \
      -DARROW_GCS=ON \
      -DARROW_MIMALLOC=OFF \
      -DARROW_WITH_BROTLI=ON \
      -DARROW_WITH_BZ2=ON \
      -DARROW_WITH_LZ4=ON \
      -DARROW_WITH_SNAPPY=ON \
      -DARROW_WITH_ZLIB=ON \
      -DARROW_WITH_ZSTD=ON \
      -DARROW_EXTRA_ERROR_CONTEXT=ON \
      -DCMAKE_INSTALL_PREFIX=$ARROW_HOME \
      -DCMAKE_BUILD_TYPE=Release \
      -DGTest_SOURCE=BUNDLED \
      -DCMAKE_CXX_STANDARD=17 \
      -Dabsl_SOURCE=BUNDLED

$ ninja install
[...]

PR #35147, opened 2023-04-14
GH-35140: [R] Rewrite configure script and ensure we don't use mismatched libarrow

### LINKED ISSUE #31989: [R] Improve package installation (on Fedora 36) with arrow system packages
Fedora 36 includes both a 'libarrow' and a 'libarrow-dataset' package under pkg-config names 'arrow' and 'arrow-dataset', respectively. When we do `install.packages("arrow")` on Fedora 36, the pkg-config for 'arrow' gets picked up first, and we get a minimal arrow install. Installing with `NOT_CRAN=true ARROW_USE_PKG_CONFIG=false` results in a full compiled installation because binaries are not available for fedora-36 (according to the install output).

Reported by Roger Bivand (while testing GDAL + Arrow support since Fedora 36 is the easiest place to do that due to system Arrow): https://github.com/paleolimbot/narrow/issues/7

Reproducble using the `fedora:36` docker container:

```bash

# docker run --rm -it fedora:36
dnf install -y R libarrow-dataset-devel cmake
R -e 'install.packages("arrow", repos = "https://cloud.r-project.org/")'
R -e 'arrow::arrow_info()$capabilities'
```

**Reporter**: [Dewey Dunnington](https://issues.apache.org/jira/browse/ARROW-16639) / @paleolimbot

<sub>**Note**: *This issue was originally created as [ARROW-16639](https://issues.apache.org/jira/browse/ARROW-16639). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

### LINKED ISSUE #35140: [R] configure script for R fails to set PKG_CFLAGS

arrow version: 11.0.0.3
R version: 4.2.3
platform: arm64

I found that current `r/configure`  fails to set PKG_CFLAGS parsing ARROW_OPTS_CMAKE file. 

This is a log I tried to install arrow to my m1 mac.
```
> install.packages("arrow")
Installing package into ‘/opt/homebrew/lib/R/4.2/site-library’
(as ‘lib’ is unspecified)
trying URL 'https://cran.ism.ac.jp/src/contrib/arrow_11.0.0.3.tar.gz'
Content type 'application/x-gzip' length 3921484 bytes (3.7 MB)
==================================================
downloaded 3.7 MB

* installing *source* package ‘arrow’ ...
files ‘tools/cpp/src/arrow/Testing/Temporary/CTestCostData.txt’, ‘tools/cpp/src/arrow/Testing/Temporary/LastTest.log’ are missing
** using staged installation
*** Arrow C++ libraries found via pkg-config at /opt/homebrew/Cellar/apache-arrow/11.0.0_3/lib
**** Warning: library version mismatch
**** C++ is 11.0.0 but R is 11.0.0.3
**** If installation fails, upgrade the C++ library to match
**** or retry with ARROW_USE_PKG_CONFIG=false
PKG_CFLAGS= 
PKG_LIBS=-L/opt/homebrew/Cellar/apache-arrow/11.0.0_3/lib -larrow
```
The log shows `PKG_CFLAGS` is empty.
This causes that capabilities, such as dataset, parquet,  etc..., are FALSE.
```
> arrow_info()
Arrow package version: 11.0.0.3

Capabilities:
               
dataset   FALSE
substrait FALSE
parquet   FALSE
json      FALSE
s3        FALSE
gcs       FALSE
utf8proc   TRUE
re2        TRUE
snappy     TRUE
gzip       TRUE
brotli     TRUE
zstd       TRUE
lz4        TRUE
lz4_frame  TRUE
lzo       FALSE
bz2        TRUE
jemalloc   TRUE
mimalloc  FALSE

Memory:
                  
Allocator jemalloc
Current    0 bytes
Max        0 bytes

Runtime:
                        
SIMD Level          none
Detected SIMD Level none

Build:
                                    
C++ Library Version           11.0.0
C++ Compiler              AppleClang
C++ Compiler Version 14.0.3.14030022
```

This is caused by `$LIB_DIR` is empty at this line
https://github.com/apache/arrow/blob/5c73973a23deb68bef75be8acd743763307ac12f/r/configure#L255

I found that the deletion of line at this pull request
https://github.com/apache/arrow/pull/14235/files#diff-089697faebdb7820ca629a2bb316b878cc0ba18a5bfb0b60996f8dbcd1fa11e7L234
causes this problem.
```
LIB_DIR=`echo $PKG_DIRS | sed -e 's/^-L//'`
```

I have already checked that to recover this line resolves this problem.

This is the log.

```
install.packages("./arrow", repos=NULL, type="source")
Installing package into ‘/opt/homebrew/lib/R/4.2/site-library’
(as ‘lib’ is unspecified)
* installing *source* package ‘arrow’ ...
files ‘tools/cpp/src/arrow/Testing/Temporary/CTestCostData.txt’, ‘tools/cpp/src/arrow/Testing/Temporary/LastTest.log’ are missing
file ‘configure’ has the wrong MD5 checksum
** using staged installation
*** Arrow C++ libraries found via pkg-config at /opt/homebrew/Cellar/apache-arrow/11.0.0_3/lib
**** Warning: library version mismatch
**** C++ is 11.0.0 but R is 11.0.0.3
**** If installation fails, upgrade the C++ library to match
**** or retry with ARROW_USE_PKG_CONFIG=false
PKG_CFLAGS=  -DARROW_R_WITH_PARQUET -DARROW_R_WITH_DATASET -DARROW_R_WITH_JSON -DARROW_R_WITH_S3
PKG_LIBS=-L/opt/homebrew/Cellar/apache-arrow/11.0.0_3/lib -larrow_dataset -lparquet -larrow 
** libs
make: Nothing to be done for `all'.
installing to /opt/homebrew/lib/R/4.2/site-library/00LOCK-arrow/00new/arrow/libs
** R
** inst
** byte-compile and prepare package for lazy loading
** help
*** installing help indices
** building package indices
** testing if installed package can be loaded from temporary location
** checking absolute paths in shared objects and dynamic libraries
** testing if installed package can be loaded from final location
** testing if installed package keeps a record of temporary installation path
* DONE (arrow)
> library(arrow)
Some features are not enabled in this build of Arrow. Run `arrow_info()` for more information.

Attaching package: ‘arrow’

The following object is masked from ‘package:utils’:

    timestamp

> arrow_info()
Arrow package version: 11.0.0.3

Capabilities:
               
dataset    TRUE
substrait FALSE
parquet    TRUE
json       TRUE
s3         TRUE
gcs       FALSE
utf8proc   TRUE
re2        TRUE
snappy     TRUE
gzip       TRUE
brotli     TRUE
zstd       TRUE
lz4        TRUE
lz4_frame  TRUE
lzo       FALSE
bz2        TRUE
jemalloc   TRUE
mimalloc  FALSE

Memory:
                  
Allocator jemalloc
Current    0 bytes
Max        0 bytes

Runtime:
                        
SIMD Level          none
Detected SIMD Level none

Build:
                                    
C++ Library Version           11.0.0
C++ Compiler              AppleClang
C++ Compiler Version 14.0.3.14030022
```

### Component(s)

R

nealrichardson:
I've significantly rewritten `r/configure` to make it easier to reason about and harder for issues like https://github.com/apache/arrow/pull/34229 and #35140 to happen. I've also added a version check to make sure that we don't obviously try to use a system C++ library that doesn't match the R package version. Making sure this was applied in all of the right places and handling what to do if the versions didn't match was the impetus for the whole refactor. 

`configure` has been broken up into some functions, and the flow of the script is, as is documented at the top of the file:

```
# * Find libarrow on the system. If it is present, make sure
#   that its version is compatible with the R package.
# * If no suitable libarrow is found, download it (where allowed)
#   or build it from source.
# * Determine what features this libarrow has and what other
#   flags it requires, and set them in src/Makevars for use when
#   compiling the bindings.
# * Run a test program to confirm that arrow headers are found
```

All of the detection of CFLAGS and `-L` dirs etc. happen in one place now, and they all prefer using `pkg-config` to read from the libarrow build what libraries and flags it requires, rather than hard-coding. (autobrew is the only remaining exception, but I didn't feel like messing with that today.) This should make the builds more future proof, should make it so more build configurations work (e.g. I suspect that a static build in ARROW_HOME wouldn't have gotten picked up correctly because it didn't add `-larrow_bundled_dependencies` to the libs, but now it will), and it may eliminate the redundant `-l` and `-D` setting I've observed in some builds (not harmful but definitely sloppy).

Version checking has been added in an R script for ease of testing (and for easier handling of arithmetic), and there is an accompanying `test-check-versions.R` added. These are run on all the builds that use `ci/scripts/r_test.sh`. 

### Behavior changes

* If libarrow is found on the system (via ARROW_HOME, pkg-config, or brew), but the version does not match, it will not be used, and we will try a bundled build. This should mean that users installing a released version will never have libarrow version problems. 
* If both the found C++ library and R package are on matching dev versions (i.e. not identical given the x.y.z.9000 vs x+1.y.z-SNAPSHOT difference), it will proceed with a warning that you may need to rebuild if there are issues. This means that regular developers will see an extra message in the build output.
* autobrew is only used on a release version unless you set FORCE_AUTOBREW=true. This eliminates another source of version mismatches (C++ release version, R dev version).
* The path where you could set `LIB_DIR` and `INCLUDE_DIR` env vars has been removed. Use `ARROW_HOME` instead.

* Closes: #35140
* Closes: #31989


nealrichardson:
If this is in fact the fix for the conda package issues (https://github.com/conda-forge/r-arrow-feedstock/issues/56#issuecomment-1518745545, https://github.com/conda-forge/r-arrow-feedstock/issues/62), then we would have a way to test this in the nightly conda packaging jobs. We currently build the packages but don't install them and assert anything about their capabilities. 

Perhaps we can add something to https://github.com/apache/arrow/blob/main/dev/tasks/conda-recipes/r-arrow/build.sh to load the installed package and test that it has parquet, dataset, etc.

nealrichardson:
@github-actions crossbow submit conda-linux-*-r4* conda-osx-*-r4*

r/configure
```diff
@@ -252,6 +252,10 @@ echo "#include $PKG_TEST_HEADER" | ${TEST_CMD} >/dev/null 2>&1
 
 if [ $? -eq 0 ]; then
   # Check for features
+  if [ "$LIB_DIR" = "" ]; then
+    # In case LIB_DIR wasn't previously set, infer from PKG_DIRS
+    LIB_DIR=`echo $PKG_DIRS | sed -e 's/^-L//'`
+  fi
```
kou, line 258: I think that the following approach (we must set `LIB_DIR` when `PKG_*` are set) is better than this approach:

```diff
diff --git a/r/configure b/r/configure
index 65528bdc3..f452fffb1 100755
--- a/r/configure
+++ b/r/configure
@@ -126,10 +126,11 @@ else
     if [ "$UNAME" = "Darwin" ] && [ "$FORCE_BUNDLED_BUILD" != "true" ]; then
       if [ "$FORCE_AUTOBREW" != "true" ] && [ "`command -v brew`" ] && [ "`brew ls --versions ${PKG_BREW_NAME}`" != "" ]; then
         echo "*** Using Homebrew ${PKG_BREW_NAME}"
-        BREWDIR=`brew --prefix`
+        BREWDIR=`brew --prefix ${PKG_BREW_NAME}`
+        LIB_DIR="$BREWDIR/lib"
         PKG_LIBS="-larrow -larrow_bundled_dependencies"
-        PKG_DIRS="-L$BREWDIR/opt/$PKG_BREW_NAME/lib $PKG_DIRS"
-        PKG_CFLAGS="-I$BREWDIR/opt/$PKG_BREW_NAME/include $PKG_CFLAGS"
+        PKG_DIRS="-L$LIB_DIR $PKG_DIRS"
+        PKG_CFLAGS="-I$BREWDIR/include $PKG_CFLAGS"
       else
         echo "*** Downloading ${PKG_BREW_NAME}"
         if [ -f "autobrew" ]; then
@@ -144,7 +145,9 @@ else
         if [ $? -ne 0 ]; then
           echo "Failed to retrieve binary for ${PKG_BREW_NAME}"
         fi
-        # autobrew sets `PKG_LIBS`, `PKG_DIRS`, and `PKG_CFLAGS`
+        # autobrew sets `PKG_LIBS`, `PKG_DIRS`, `PKG_CFLAGS`, and `BREWDIR`
+        # but `LIB_DIR` isn't set.
+        LIB_DIR="$BREWDIR/lib"
       fi
     else
       if [ "${NOT_CRAN}" = "true" ]; then
```

dev/tasks/conda-recipes/r-arrow/build.sh
```diff
@@ -14,3 +14,5 @@ fi
 
 # ${R_ARGS} necessary to support cross-compilation
 ${R} CMD INSTALL --build r/. ${R_ARGS}
+# Ensure that features are enabled in the R build (feel free to add others)
+${R} -s -e 'library(arrow); stopifnot(arrow_with_dataset(), arrow_with_parquet(), arrow_with_s3())'
```
kou, line 18: @h-vetinari Could you review conda part?

h-vetinari: That part is fine and already working (as in: shows the build fails) in https://github.com/conda-forge/r-arrow-feedstock/pull/63.

However, I don't see how the BREW-specific patch above would apply to conda-forge (we'd need `$PREFIX`), nor do I know what changed between 10 & 11. While taking a first look, my comment in that conda-forge PR was:
> Do we need to do something like in the upstream R [Makefile](https://github.com/apache/arrow/blob/apache-arrow-11.0.0/r/Makefile#L37-L38) to discover the deps correctly?

r/configure
nealrichardson: @kou is this possibly relevant? `PKG_LIBS` and `PKG_DIRS` seem to be set correctly but for some reason `LIB_DIR` isn't. 

```diff
diff --git a/r/configure b/r/configure
index 65528bdc38..8e130c53b4 100755
--- a/r/configure
+++ b/r/configure
@@ -110,7 +110,7 @@ else
     PKG_CFLAGS="$PKGCONFIG_CFLAGS $PKG_CFLAGS"
     PKG_LIBS="${PKGCONFIG_LIBS}"
     PKG_DIRS="${PKGCONFIG_DIRS}"
-    LIB_DIR=${FOUND_LIB_DIR}
+    LIB_DIR="${FOUND_LIB_DIR}"
 
     # Check for version mismatch
     PC_LIB_VERSION=`pkg-config --modversion arrow`
```

kou: Sorry. Could you explain more the context for it? Can I see a build log for the case? This case https://github.com/apache/arrow/issues/35140 ? Or other case?

If this is for #35140, I think that the isn't related because `FOUND_LIB_DIR` doesn't have any space:

```text
*** Arrow C++ libraries found via pkg-config at /opt/homebrew/Cellar/apache-arrow/11.0.0_3/lib
```

kou: Ah, #35140 uses Apache Arrow 11.0.0.
I was seeing the main branch.

Yes. The change https://github.com/apache/arrow/pull/35147#discussion_r1175945751 will fix #35140. And it's already in the main branch: https://github.com/apache/arrow/pull/34229

h-vetinari: From the conda-side, we were able to fix it by specifying `LIB_DIR=$PREFIX/lib` in our build script, see https://github.com/conda-forge/r-arrow-feedstock/pull/63

nealrichardson: 🤦 my mistake, sorry I missed #34229, that explains some of the confusion. I'll tidy this PR up with any remaining useful changes and get it merged.

nealrichardson:
I've applied the patches and added a bunch of comments to `r/configure` to help with any upcoming refactoring. I want to do a bit more testing of the various branches locally before we move forward, so I've marked this as draft for now.

nealrichardson:

nealrichardson:
I will inspect the successful builds to make sure features are enabled as they should be. Of the failures:

* `test-r-offline-maximal` is a download timeout for orc (which we don't even use). https://github.com/ursacomputing/crossbow/actions/runs/4833915427/jobs/8614507086#step:4:50
* `test-r-arrow-backwards-compatibility` I don't quite understand on first glance, will have to inspect, says it can't find libcurl https://github.com/ursacomputing/crossbow/actions/runs/4833911218/jobs/8614498685#step:7:27
* `r-binary-packages` has at least one failure in parsing LIB_DIR, in which it was assumed to have only one -L in it but there were two https://github.com/ursacomputing/crossbow/actions/runs/4833911390/jobs/8614512347#step:14:25930

nealrichardson:

nealrichardson:
The `r-binary-packages` failures the first time on mac could be real issues with the bundled build, which was being triggered because we are on a dev version so now skipping autobrew unless FORCE_AUTOBREW is set. On 10.13, snappy had a compilation error. I'm not inclined to care about that. But on macOS 11, the package built but couldn't find absl symbols, which is odd because `-larrow_bundled_dependencies` was included. Hopefully the other LIB_DIR issue was somehow responsible? In any case, if it's a real issue, it will probably reappear somewhere else.

nealrichardson:
Conda failures all appear to be in python (perhaps the pandas related tests). There's one job failing in the `r-binary-packages` job, something with the LIB_DIR again, will investigate. Other than that, and inspecting the other jobs to make sure that features are correctly enabled, this is about done.

nealrichardson:

nealrichardson:
I checked a handful of the builds and they all showed the correct features enabled. We could update https://github.com/apache/arrow/blob/main/dev/tasks/macros.jinja#L278 and perhaps other places to ensure that features other than just parquet are enabled, like what I added to the conda packaging. But I'll wait for review before poking this further.

The segfault in the pyarrow interaction (this time on flight tests) is happening a lot on the `R / AMD64 Ubuntu 20.04 R 4.2 Force-Tests true`, but it is also happening on `main`, so I don't think it is related.

r/configure
```diff
@@ -58,191 +99,284 @@ if [ ! "`${R_HOME}/bin/R CMD config CXX17`" ]; then
...
+  else
+    set_pkg_vars_without_pc $1
+  fi
+
+  # If on Raspberry Pi, need to manually link against latomic
+  # See: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81358 for similar example
+  if grep raspbian /etc/os-release >/dev/null 2>&1; then
+    PKG_CFLAGS="$PKG_CFLAGS -DARROW_CXXFLAGS=-latomic"
+    PKG_LIBS="-latomic $PKG_LIBS"
+  fi
```
kou, line 264: We need this only when `pkg-config` doesn't exist. Because it's cared in C++: https://github.com/apache/arrow/commit/74af39d486b5a750a8099abc163935e5c8d911bf

```suggestion
    # If on Raspberry Pi, need to manually link against latomic
    # See: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81358 for similar example
    if grep raspbian /etc/os-release >/dev/null 2>&1; then
      PKG_CFLAGS="$PKG_CFLAGS -DARROW_CXXFLAGS=-latomic"
      PKG_LIBS="$PKG_LIBS -latomic"
    fi
  fi
```

Or we may want to move this to `set_pkg_vars_without_pc`.

nealrichardson: 👍 I'll move it in `set_pkg_vars_without_pc`

r/configure
```diff
@@ -58,191 +99,284 @@ if [ ! "`${R_HOME}/bin/R CMD config CXX17`" ]; then
...
-        # Set some default values
-        if [ "${LIBARROW_BINARY}" = "" ]; then
-          export LIBARROW_BINARY=true
-        fi
-        if [ "${LIBARROW_MINIMAL}" = "" ]; then
-          export LIBARROW_MINIMAL=false
-        fi
-      fi
+      # This case must be ARROW_DEPENDENCY_SOURCE=BUNDLED.
+      # TODO: is there no way we can infer from ArrowOptions.cmake or arrow.pc if these are needed?
```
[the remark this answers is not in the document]
nealrichardson: Thanks, I'll update the comment and remove the TODO

r/configure
```diff
@@ -58,191 +100,282 @@ if [ ! "`${R_HOME}/bin/R CMD config CXX17`" ]; then
...
+# with brew, but it is generally not linked. We can over-ride this and find
+# openssl but setting OPENSSL_ROOT_DIR (which cmake will pick up later in
+# the installation process). FWIW, arrow's cmake process uses this
+# same process to find openssl, but doing it now allows us to catch it in
+# nixlibs.R and throw a nicer error.
+if [ "$UNAME" = "Darwin" ] && [ "${OPENSSL_ROOT_DIR}" = "" ] && [ "`command -v brew`" ]; then
+  if brew --prefix openssl >/dev/null 2>&1; then
+    export OPENSSL_ROOT_DIR="`brew --prefix openssl`"
+    export PKG_CONFIG_PATH="${OPENSSL_ROOT_DIR}/lib/pkgconfig${PKG_CONFIG_PATH:+:${PKG_CONFIG_PATH}}"
+  fi
```
nealrichardson, line 121: @kou can we simplify this one further?

```suggestion
if [ "${OPENSSL_ROOT_DIR}" = "" ] && brew --prefix openssl >/dev/null 2>&1; then
  export OPENSSL_ROOT_DIR="`brew --prefix openssl`"
  export PKG_CONFIG_PATH="${OPENSSL_ROOT_DIR}/lib/pkgconfig${PKG_CONFIG_PATH:+:${PKG_CONFIG_PATH}}"
```

nealrichardson:
> Just catching up on this after some PTO! I will give this a test ride on all my local environments today 🙂

🙏  Some combinations of things I tested locally, with a dev build of arrow in `ARROW_HOME` and having done `brew install apache-arrow`:

* Usual way, with `ARROW_HOME` set: you see the new message warning you that you're on dev versions and if the build fails, update your C++ build
* Edit the version number in `"$ARROW_HOME/lib/pkgconfig/arrow.pc"` and see that it is rejected because the versions don't match.
* Unset `ARROW_HOME`: Homebrew apache-arrow is found, but it fails the version check, so it isn't used. Since we're on a dev version, autobrew isn't attempted and instead the bundled build starts (I ctrl-C'd though, not wanting to wait for it).
* Put `"$(brew --prefix apache-arrow)/lib/pkgconfig` on `PKG_CONFIG_PATH` and see that `pkg-config` finds it (but then it is rejected because the versions don't match)

r/configure
```diff
@@ -58,191 +100,282 @@ if [ ! "`${R_HOME}/bin/R CMD config CXX17`" ]; then
...
+    export PKG_CONFIG_PATH="${_LIBARROW_FOUND}/lib/pkgconfig${PKG_CONFIG_PATH:+:${PKG_CONFIG_PATH}}"
   else
-    PKG_CONFIG_AVAILABLE=false
-  fi
-  if [ "$PKG_CONFIG_AVAILABLE" = "true" ] && [ "$ARROW_USE_PKG_CONFIG" != "false" ]; then
-    # Set the search paths and compile flags
-    PKGCONFIG_CFLAGS=`pkg-config --cflags --silence-errors ${PKG_CONFIG_NAME}`
-    PKGCONFIG_LIBS=`pkg-config --libs-only-l --libs-only-other --silence-errors ${PKG_CONFIG_NAME}`
-    PKGCONFIG_DIRS=`pkg-config --libs-only-L --silence-errors ${PKG_CONFIG_NAME}`
+    _LIBARROW_FOUND="false"
```
jonkeane, line 152: Would it be helpful to have a message here too? Or is one thrown later? I really really like that all the other branches here have clear "this is the branch it's on now" — that's been frustrating to diagnose before.

r/configure
```diff
@@ -58,191 +100,282 @@ if [ ! "`${R_HOME}/bin/R CMD config CXX17`" ]; then
...
+}
+
+arrow_built_with() {
+  # Function to check cmake options for features
+  grep -i 'set('"$1"' "ON")' $ARROW_OPTS_CMAKE >/dev/null 2>&1
+}
+
+##############
+# Main logic #
+##############
```
jonkeane, line 349: This is a little pedantic, but I wonder if it would be more straightforward to have this up top and the functions down below (kinda like you define `arrow_built_with` after the function that uses it)? 

I'm not absolutely tied to it, but having the operative thing up top and supporting things below it is a nice way to read things

r/configure
```diff
@@ -58,191 +100,282 @@ if [ ! "`${R_HOME}/bin/R CMD config CXX17`" ]; then
...
+  if [ "$_LIBARROW_FOUND" = "false" ]; then
+    # If we haven't found a suitable version of libarrow, build it
+    if [ "$UNAME" = "Darwin" ] && ! echo $VERSION | grep -q "000"; then
+      # Only autobrew on release version (for testing, use FORCE_AUTOBREW above)
+      # (dev versions end in .9000, and nightly gets something like .10000xxx)
+      do_autobrew
+    else
+      do_bundled_build
+    fi
+  fi
```
jonkeane, line 367: Is it overkill to call out this order for the envvars in the top-matter of this file? That `FORCE_BUNDLED_BUILD` takes precedence over `FORCE_AUTOBREW` which takes precedence over the natural order of things. 

It's "obvious" when you know what those things do in detail to a certain extent, but not everyone has that in their heads.

r/tools/check-versions.R
```diff
@@ -0,0 +1,56 @@
...
+    msg <- c(
+      sprintf("*** > Packages are both on development versions (%s, %s)", cpp_version, r_version),
+      "*** > If installation fails, rebuild the C++ library to match the R version",
+      "*** > or retry with FORCE_BUNDLED_BUILD=true"
+    )
+    cat(paste0(msg, "\n", collapse = ""))
+  } else if (r_version != cpp_version) {
+    cat(
+      sprintf(
+        "**** Not using found C++ library: version (%s) does not match R package (%s)\n",
```
jonkeane, line 44:
```suggestion
        "**** C++ library conflict: found version (%s) does not match R package (%s)\n",
```

Something about "Not using found C++" strikes me as a bit off — almost like malicious compliance style or something?

Separately (and this might be too hard) but would it be possible to say _where_ that thing was found from? either brew|system install or a path might help folks identify what's up? It should be in an earlier log line so it's probably not ultra necessary here, but we might already have that envvar here to use?

r/tools/check-versions.R
```diff
@@ -0,0 +1,56 @@
...
+        cpp_version,
+        r_version
+      )
+    )
+    stop("version mismatch")
+    # Add ALLOW_VERSION_MISMATCH env var to override stop()? (Could be useful for debugging)
```
jonkeane, line 50: 🙏 Not necessary to ship this, of course, but I will add a 🗳️ to this I have needed to | wanted to do this while debugging in the past.

jonkeane (approved):
Thanks for this! So, so much cleaner + clearer and I love that there's much less needing to remember which of 3 paths you're on in nested if/elses.

r/configure
nealrichardson: If you get here, the next message you'll see is something about building libarrow, either with autobrew or something that says which directory the C++ source has been found in. I think it will be clear enough where we are in the flow, but I don't feel strongly about it.

r/configure
nealrichardson: Yeah I agree that would be nice.

r/configure
nealrichardson: IMO it's overkill, and moving this function to the top will make it more visible too. I'm pretty sure the only use of FORCE_AUTOBREW is in our CI.

r/configure
kou: Yes!
Sorry, I didn't notice this in the previous review. :<

r/configure
```diff
@@ -58,191 +100,282 @@ if [ ! "`${R_HOME}/bin/R CMD config CXX17`" ]; then
...
+  if [ -n "$(find "$LIB_DIR" -name 'libarrow_bundled_dependencies.*')" ]; then
+    PKG_LIBS="$PKG_LIBS -larrow_bundled_dependencies"
+  fi
+
+  # If on Raspberry Pi, need to manually link against latomic
+  # See: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81358 for similar example
+  # pkg-config will handle this for us automatically, see ARROW-6312
+  if grep raspbian /etc/os-release >/dev/null 2>&1; then
+    PKG_CFLAGS="$PKG_CFLAGS -DARROW_CXXFLAGS=-latomic"
+    PKG_LIBS="-latomic $PKG_LIBS"
```
kou, line 291: I think that we need to APPEND `-latomic` not PREPEND because `-larrow` needs `-latomic`:

```suggestion
    PKG_LIBS="$PKG_LIBS -latomic"
```

kou (approved):
+1

nealrichardson:
> I did get a build error after unsetting ARROW_HOME (see below); however, it worked beautifully on MacOS x86! I don't think my M1 problem should block this...this rewrite makes it much easier to fix if it does turn out to be a script issue.
> 
> I'll try on Windows and on a few suitable Ubuntu docker images tomorrow.
> 
> ```
> unable to load shared object '/Users/deweydunnington/Library/R/arm64/4.2/library/00LOCK-r/00new/arrow/libs/arrow.so':
>   dlopen(/Users/deweydunnington/Library/R/arm64/4.2/library/00LOCK-r/00new/arrow/libs/arrow.so, 0x0006): symbol not found in flat namespace (__ZN4absl12lts_2023012510FormatTimeENSt3__117basic_string_viewIcNS1_11char_traitsIcEEEENS0_4TimeENS0_8TimeZoneE)
> Error: loading failed
> Execution halted
> ERROR: loading failed
> * removing ‘/Users/deweydunnington/Library/R/arm64/4.2/library/arrow’
> * restoring previous ‘/Users/deweydunnington/Library/R/arm64/4.2/library/arrow’
> 
> Exited with status 1.
> ```

That looks similar to the build failure I saw [here](https://github.com/ursacomputing/crossbow/actions/runs/4833911390/jobs/8614891010), where it appears that we are bundling absl because there is a (newer) version installed by homebrew but we require an exact version match (because absl is very unstable, it seems): https://github.com/ursacomputing/crossbow/actions/runs/4833911390/jobs/8614891010#step:13:416

I don't understand well enough how the linker resolves things to understand how we would insist that we prefer the static symbols in `-larrow_bundled_dependencies` over shared libraries found in `-L/opt/homebrew/lib`. @kou would reordering the `-L` and `-l`s help? In that build, [PKG_LIBS](https://github.com/ursacomputing/crossbow/actions/runs/4833911390/jobs/8614891010#step:13:1638) is `-L/Users/voltrondata/tmp/RtmpQAJbFi/R.INSTALLca181e5538f5/arrow/libarrow/arrow-11.0.0.100000506/lib -L/opt/homebrew/lib -larrow /opt/homebrew/lib/libsnappy.1.1.10.dylib /opt/homebrew/lib/libre2.10.0.0.dylib /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk/usr/lib/libbz2.tbd /opt/homebrew/lib/libaws-cpp-sdk-config.dylib /opt/homebrew/lib/libaws-cpp-sdk-transfer.dylib /opt/homebrew/lib/libaws-cpp-sdk-identity-management.dylib /opt/homebrew/lib/libaws-cpp-sdk-cognito-identity.dylib /opt/homebrew/lib/libaws-cpp-sdk-sts.dylib /opt/homebrew/lib/libaws-cpp-sdk-s3.dylib /opt/homebrew/lib/libaws-cpp-sdk-core.dylib -larrow_bundled_dependencies -R/opt/homebrew/lib -lbrotlidec -R/opt/homebrew/lib -lbrotlienc -lthrift -lz -llz4 -lzstd -lutf8proc`

In any case, I doubt this is a regression introduced in this PR but I could be wrong.

nealrichardson:
Also @paleolimbot you don't need to test Windows because this PR doesn't modify `configure.win`.

r/tools/check-versions.R
nealrichardson: Yeah it should be in the line immediately above this in the logs.

kou:
I think that this problem is related to `-I` not `-L`/`-l`: https://github.com/apache/arrow/issues/34523#issuecomment-1463069526

kou:
Ah, wait. The job was passed at https://github.com/apache/arrow/pull/35147#issuecomment-1529017959 .
It may not be related to `-I`. I'l take a look at this.

nealrichardson:
> Ah, wait. The job was passed at [#35147 (comment)](https://github.com/apache/arrow/pull/35147#issuecomment-1529017959) . It may not be related to `-I`. I'll take a look at this.

It passed there because I changed it not to use the bundled build. That said, it didn't pick up the wrong absl, so maybe there's still something to learn from it.

kou:
Hmm. It seems that autobrew not Homebrew was used in the passed job.

kou:
Ah, OK. Thanks.

nealrichardson:
In the failed build, `PKG_CFLAGS=-DARROW_STATIC -DUTF8PROC_EXPORTS -I/Users/voltrondata/tmp/RtmpQAJbFi/R.INSTALLca181e5538f5/arrow/libarrow/arrow-11.0.0.100000506/include -I/opt/homebrew/include`

nealrichardson:
autobrew does not use pkg-config to set PKG_CFLAGS, so `-I/opt/homebrew/include` is not present in that build

kou:
I think that this is the same problem as https://github.com/apache/arrow/issues/34523 .
And this is a C++ part problem not an R part problem.
So I think that we can work on this as a separated pull request.
  
`__ZN4absl12lts_2023012510FormatTimeENSt3__117basic_string_viewIcNS1_11char_traitsIcEEEENS0_4TimeENS0_8TimeZoneE` is  `absl::lts_20230125::FormatTime(std::__1::basic_string_view<char, std::__1::char_traits<char> >, absl::lts_20230125::Time, absl::lts_20230125::TimeZone)`. And it's used at https://github.com/apache/arrow/blob/a3a01726b1ca195c3d26fec084c0ff19453e6866/cpp/src/arrow/filesystem/gcsfs_internal.cc#L241 .

Abseil was mixed when we build C++'s `libarrow.a` not R's `arrow.so`.

r/configure
```diff
@@ -17,39 +17,81 @@
...
+#
+# * Installing a released version from source, as from CRAN, with
+#   no other prior setup
+#   * On macOS, autobrew is used to retrieve libarrow and dependencies
+#   * On Linux, the nixlibs.R build script will download or build
+# * Installing a released version but first installing libarrow.
+#   It will use pkg-config and brew to search for libraries.
+# * Installing a development version from source as a user.
+#   nixlibs.R will be used to build from source. System packages that
+#   are not on the same dev version will be skipped.
```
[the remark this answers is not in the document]
nealrichardson:
```suggestion
#   Any libarrow found on the system corresponding to a release will not match
#   the dev version and will thus be skipped. nixlibs.R will be used to build libarrow.
```

nealrichardson:

nealrichardson:
I don't fully understand the devdocs failure, but it has to do with the fact that arrow 12.0.0 has been released (https://github.com/ursacomputing/crossbow/actions/runs/4862485406/jobs/8668915213#step:9:3878 vs. https://github.com/ursacomputing/crossbow/actions/runs/4833910722/jobs/8614497534#step:9:3880 where it passed before). The cflags/libs are identical between the two builds. Rebasing to pull in the new version number would probably solve it, though I'm not sure it's worth the electricity to confirm it (that's an expensive build for very little value, it seems).

nealrichardson:
> It doesn't seem to be pulling the hosted static libraries for development, which I expected on Ubuntu/x86 (but maybe it's not supposed to?).

We don't host binaries for x.y.z.9000, the nightly versions are like x.y.z.1000123, so if you wanted it to pull a nightly for a .9000 dev build, we'd have to teach it how to find the most recent one. Something for you (or whoever) to consider on #35049 I suppose.

nealrichardson:
Made #35391 for the segfault on the ubuntu force tests job, which I see on `main` as well.

PR #35148, opened 2023-04-14
GH-35133: [Go] Fails to build on 32-bit platform

### LINKED ISSUE #35133: [Go] Fails to build on 32-bit platforms

The current code fails to build on 32-bit platforms (at least on armel, armhf, mips, mipsel, linux-386 and windows-386) with the following error

```text
github.com/apache/arrow/go/v12/arrow/array
/go/pkg/mod/github.com/apache/arrow/go/v12@v12.0.0-20230414125938-59fe4ce90e2f/arrow/array/dictionary.go:1564:22: math.MaxUint32 (untyped int constant 4294967295) overflows int
```

### Component(s)

Go

zeroshade:
### Rationale for this change
Because in Go untyped `int` values use whatever the system's default size is, an untyped constant value of `math.MaxUint32` is larger than can fit in a 32-bit integer, which would be what `int` is on a 32-bit system.

As a result, we need to cast this failed situation so that we explicitly use a `int64` to ensure this builds on 32-bit systems.

* Closes: #35133

### DIFF
--- a/go/arrow/array/dictionary.go
+++ b/go/arrow/array/dictionary.go
@@ -45,12 +45,12 @@ import (
 //
 // For example, the array:
 //
-//      ["foo", "bar", "foo", "bar", "foo", "bar"]
+//	["foo", "bar", "foo", "bar", "foo", "bar"]
 //
 // with dictionary ["bar", "foo"], would have the representation of:
 //
-//      indices: [1, 0, 1, 0, 1, 0]
-//      dictionary: ["bar", "foo"]
+//	indices: [1, 0, 1, 0, 1, 0]
+//	dictionary: ["bar", "foo"]
 //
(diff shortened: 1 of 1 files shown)


zeroshade:
CC @srebhan I don't have easy access to a 32-bit system to test this on, can you try testing this PR on whatever system you found the original issue on? Thanks!

srebhan:
@zeroshade we are "cross-compiling" (Telegraf) by setting GOOS and GOARCH accordingly. So there is no need to have physical access to such a machine. An example of what needs to be set can be found [here](https://github.com/influxdata/telegraf/blob/2fed77e02ad963c7aeb150cfa282bddfd715b927/.circleci/config.yml#L186).

zeroshade:
Gonna close this in favor of #35159

PR #35159, opened 2023-04-16
GH-35133: [Go] fix for `math.MaxUint32 overflows int` error in 32-bit arch

### LINKED ISSUE #35133: [Go] Fails to build on 32-bit platforms

sunpe:
### Rationale for this change

When compiling on i386 arch, an error like `math.MaxUint32 (untyped int constant 4294967295) overflows int` will appear. The Int type is 32-bit and in 32-bit arch it overflows. In my PR, I cast `int` to `unint` to avoid this error.

* Closes: #35133

### DIFF
--- a/go/arrow/array/dictionary.go
+++ b/go/arrow/array/dictionary.go
@@ -1561,7 +1561,7 @@ func (u *unifier) GetResultWithIndexType(indexType arrow.DataType) (arrow.Array,
 	case arrow.INT16:
 		toobig = dictLen > math.MaxInt16
 	case arrow.UINT32:
-		toobig = dictLen > math.MaxUint32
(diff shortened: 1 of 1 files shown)


zeroshade (approved):
Thanks for the contribution! I'll merge this after the CI passes

Issue #35160, opened 2023-04-16
[Go]: ApproxEqual for strings


yevgenypats:
### Describe the enhancement requested

In lot's of databases null characters (`\x00`) are not accepted in strings (For example [postgres](https://stackoverflow.com/questions/28813409/are-null-bytes-allowed-in-unicode-strings-in-postgresql-via-python) but not only). I think given it's such a common gotch in dbs this would be a great use-case for an `arrayApproxEqualString` method just like we have `arrayApproxEqualFloat16`.

### Component(s)

Go

PR #35161, opened 2023-04-16
GH-35160: [Go] Add arrayApproxEqualString to ignore null chars

### LINKED ISSUE #56: ApproxEqual for strings

In lot's of databases null characters (`\x00`) are not accepted in strings (For example [postgres](https://stackoverflow.com/questions/28813409/are-null-bytes-allowed-in-unicode-strings-in-postgresql-via-python) but not only). I think given it's such a common gotch in dbs this would be a great use-case for an `arrayApproxEqualString` method just like we have `arrayApproxEqualFloat16`.

### Component(s)

Go

yevgenypats:

* Closes: apache/arrow-go#56
* GitHub Issue: #56

### DIFF
--- a/go/arrow/array/binarybuilder.go
+++ b/go/arrow/array/binarybuilder.go
@@ -19,6 +19,7 @@ package array
 import (
 	"bytes"
 	"encoding/base64"
+	"encoding/hex"
 	"fmt"
 	"math"
 	"reflect"
@@ -319,7 +320,10 @@ func (b *BinaryBuilder) UnmarshalOne(dec *json.Decoder) error {
 	case string:
 		data, err := base64.StdEncoding.DecodeString(v)
 		if err != nil {
-			return err
+			data, err = hex.DecodeString(v)
+			if err != nil {
+				return err
+			}
 		}
 		b.Append(data)
 	case []byte:
--- a/go/arrow/array/compare.go
+++ b/go/arrow/array/compare.go
@@ -19,6 +19,7 @@ package array
 import (
 	"fmt"
 	"math"
+	"strings"
 
 	"github.com/apache/arrow/go/v12/arrow"
 	"github.com/apache/arrow/go/v12/arrow/float16"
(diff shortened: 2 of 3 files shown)


go/arrow/array/binarybuilder.go
```diff
@@ -319,7 +320,10 @@ func (b *BinaryBuilder) UnmarshalOne(dec *json.Decoder) error {
 	case string:
 		data, err := base64.StdEncoding.DecodeString(v)
 		if err != nil {
-			return err
+			data, err = hex.DecodeString(v)
+			if err != nil {
+				return err
+			}
```
zeroshade, line 326: this seems odd to add here. Should we also add this to `AppendValueFromString` and update the docs that you can also use hex to specify binary strings?

go/arrow/array/compare.go
```diff
@@ -630,6 +631,34 @@ func validityBitmapEqual(left, right arrow.Array) bool {
 	return true
 }
 
+func stripNulls(s string) string {
+	return strings.ReplaceAll(s, "\x00", "")
+}
+
```
zeroshade, line 637: In the Arrow spec a String array should be all valid utf8. So shouldn't this only be for trimming trailing Nulls? (ie: use `strings.TrimRight` instead of `strings.ReplaceAll` )?

To this end, can you have a test with a string that contains a null inside it that *isn't* at the end and ensure that this approxequal only trims nulls, and doesn't strip them from within the string?

zeroshade (changes requested):

go/arrow/array/binarybuilder.go
yevgenypats: Hmm....Prob it was in another branch and ended up here. will remove

go/arrow/array/compare.go
yevgenypats: Interesting, wasn't aware of the valid utf8 specification. In that case there is no check in `string.Append` [method](https://github.com/cloudquery/arrow/blob/main/go/arrow/array/string.go#L321) to make sure there are no nulls inside the string. Should we add this check or string nulls inside the string? wdyt?

zeroshade: So for performance reasons we don't validate the utf8 strings on Append currently and leave it up to the producer to ensure that they are passing valid utf-8 strings when constructing the array (if it's not valid utf-8 they should be using a Binary array). 

On my list of things to do eventually is a "Validate" method for each array type like the C++ library has. That "Validate" method would do the UTF-8 validity check on the buffer so that a consumer can choose *when* they take the performance hit for validating the utf-8.

zeroshade:
@yevgenypats are you still planning on working on this? It still needs the that hex change removed etc.

yevgenypats:
Let's close this for now and I might open this once again a bit later, Im not sure if this is needed after your comment regarding nulls.

Issue #34826, opened 2023-03-31
[R] Modernize error handling


paleolimbot:
### Describe the bug, including details regarding any error messages, version, and platform.

In a number of places we have a pattern where we do:

```r
tryCatch(
  something_hard_with_confusing_errors(),
  error = function(e) {
    if (grepl("confusing error", conditionMessage(e)) {
      stop("A better error message")
    } else {
      stop(e)
    }
  }
)
```

When we do this we loose the backtrace of the confusing error (which, although confusing to the user, is a better indicator of where the error occurred when we try to debug it). In more than one case, inspecting the original error has resulted in one of us realizing that the original error was not the one we expected. In more than one case it has resulted in one of us (me at least) manually editing some code to generate a custom build that prints out the original error (as opposed to just running the code).

Since we started doing this, rlang has made it relatively easy to throw a "classed error" and to include an original error in the backtrace so that you can do:

```r
tryCatch(
  something_hard_with_confusing_errors(),
  some_custom_error_class = function(e) {
    abort("A better error message", parent = e)
  }
}
```

If for some reason the original error should really really really not get printed, we could inspect the value of REPREX_QUIET or ARROW_R_DEV (such that we could suggest that an issue opener set one of them and run their code again in the event their data is too big to reprex).

### Component(s)

R

nealrichardson:
I agree we can do better by using the new tooling. One thing that makes this a little trickier is that usually the error message we're inspecting comes from the Arrow C++ library, where we can't control what class the R error is. 

I wonder if `cpp11`, or our custom wrapping, could at least add a class to the errors from cpp-land? That way, even if we still need to inspect the original error message to know what to do with it, you could limit that logic to the `cpp_error` class and not `error` in general, so that other kinds of errors would bubble up.

paleolimbot:
Agreed! We could probably get the class specific to something like `arrow_status_invalid` or `arrow_status_not_implemented`. We emit just about everything from here: https://github.com/apache/arrow/blob/main/r/src/arrow_types.h#L99-L117

Probably we'd need an R function like `stop_for_status()` that calls `rlang::abort()` (and then call it from C++).

thisisnic:
This issue has been marked as stale because it has had no activity in the past 2 years. Please remove the stale label or comment below, or this issue will be closed in 14 days.

PR #35172, opened 2023-04-17
GH-35035: [R] Implement names<- for Schemas 

### LINKED ISSUE #35035: [R] Implement `names<-` for Schemas

We occasionally have requests relating to updating schemas and we could make them easier to work with.  We could implement the S3 method `names<-` for Schema object, to enable users to be able to do something like:

```
my_schema <- schema(x = int8(), y = string())
names(my_schema) <- c("y", "x")
```
The specific use-case I'm thinking about here is when a user wants to read a schema from a source, update it, and then re-use it to read/write.

### Component(s)

R

thisisnic:
* Closes: #35035


r/tests/testthat/test-schema.R
```diff
@@ -260,3 +260,22 @@ test_that("as_schema() works for StructType objects", {
...
+test_that("schema name assignment", {
+  schm <- schema(x = int8(), y = string(), z = double())
+  expect_identical(names(schm), c("x", "y", "z"))
+  names(schm) <- c("a", "b", "c")
+  expect_identical(names(schm), c("a", "b", "c"))
+  expect_error(names(schm) <- "f")
+  expect_error(names(schm) <- letters)
+  expect_error(names(schm) <- character(0))
+  expect_error(names(schm) <- NULL)
+  expect_error(names(schm) <- c(TRUE, FALSE))
```
paleolimbot, line 273: Can you either add a regex for these or use `expect_snapshot()` to ensure that we aren't swallowing an unrelated error here?

thisisnic: I agree with your point, but we also discourage testing of error messages which come from libarrow. I guess I can add some custom error handling that we can test for.

nealrichardson: L269-271 are all the same, right? strings but not the right length, so libarrow will raise.

L272-273 will raise from the cpp11 wrapping since you haven't passed strings. 

So maybe you can reduce this to two tests: one that looks for the "wrong number of args" message, and one that looks for the type mismatch message. Those seem safe enough to make assertions about, and minimizes the risk from asserting about other libraries' error messages.

This is another case that #34826 would help with.

paleolimbot (approved):

PR #35173, opened 2023-04-17
GH-34775: [R] arrow_table: as.data.frame() sometimes returns a tbl and sometimes a data.frame

### LINKED ISSUE #34775: [R] arrow_table: as.data.frame() sometimes returns a tbl and sometimes a data.frame

thisisnic:
Features of this PR:
* Ensures that calling `as.data.frame()` on Arrow objects returns base R `data.frame` objects.
* Drops the `class` attribute metadata of input objects of `data.frame` class (i.e. that don't have inherit from any additional classes other than `data.frame`).  This results in us sacrificing roundtrip class fidelity for `data.frame` objects (i.e. if we input a base R data.frame, convert it to an Arrow Table, and then convert it back to R, we get a tibble).  However, we now have consistency in the type of returned objects, retain roundtrip fidelity for other (non-class) metadata, and guarantee that `as.data.frame()` returns a base R data.frame.  Users who wish to input and return a `data.frame` object can call `as.data.frame()` on the returned object.
* Implements `dplyr::collect()` for StructArrays so that these objects can still be returned as tibbles if needed.
* Renames `expect_data_frame()` to `expect_equal_data_frame()` for clarity, and updates it to convert both the object and expected object to data.frames.

* Closes: #34775


paleolimbot (approved):
This looks good! It seems like a cleaner solution than what we currently have. I like the idea of dropping metadata on the way in where possible because I seem to remember that we can skip some calls from C++ into R if there is no metadata to restore which speeds things up a bit.

PR #35178, opened 2023-04-17
GH-34852: [C++][Go][Java][FlightRPC] Add support for ordered data

### LINKED ISSUE #34852: [C++][Java][FlightRPC] Add support for ordered data

Based on the proposal in https://docs.google.com/document/d/1jhPyPZSOo2iy0LqIJVUs9KWPyFULVFJXTILDfkadx2g/edit# .
See also the discussion thread: https://lists.apache.org/thread/247z3t06mf132nocngc1jkp3oqglz7jp

Currently, the endpoints within a FlightInfo [explicitly have no ordering](https://github.com/apache/arrow/blob/4a5fadbb7200ae71a151841bac3cbf35dac10258/format/Flight.proto#L269-L270):

> There is no ordering defined on endpoints. Hence, if the returned data has an ordering, it should be returned in a single endpoint.

This is unnecessarily limiting. Systems can and do implement distributed sorts, but they can’t reflect this in Flight RPC. And users are asking; see this [Stack Overflow question](https://stackoverflow.com/questions/75180268/apache-arrow-flight-getting-sorted-data-from-multiple-endpoints).

Proposal:

Add a flag to `FlightInfo`:

```protobuf
message FlightInfo {
  // FlightEndpoints are in the same order as the data.
  bool endpoints_ordered = 6;
}
```

If set, the client may assume that the data is sorted in the same order as the endpoints. Otherwise, the client cannot make any assumptions (as before).

### Component(s)

C++, Java,FlightRPC

### LINKED ISSUE #35085: [Go][FlightRPC] Add support for ordered data

#34852 will include `.proto` change, document update and C++ implementation.
But we need one more implementation to making this "ordered data" proposal official.
See also: #35084

(Go is one of candidates. If anyone wants to work on this with other language, they are welcome.)

### Component(s)

FlightRPC, Go

kou:
### Rationale for this change

No ordering is unnecessarily limiting.  Systems can and do implement distributed sorts, but they can’t reflect this in Flight RPC.

These changes add `FlightInfo.ordered`.

### Are these changes tested?

Yes.

Yes.

* Closes: #34852
* Closes: #35085


kou:
An integration test is added.
We can review this but we can NOT merge this yet. We need one more implementation and a formal vote before we merge this.

docs/source/format/Flight.rst
```diff
@@ -90,9 +90,13 @@ A client that wishes to download the data would:
...
-   requested. There is no ordering defined on endpoints or the data
-   within, so if the dataset is sorted, applications should return
-   data in a single endpoint.
+   requested. If ``FlightInfo.ordered`` is set, returned endpoints are
+   in the same order as the data. The client can read ordered data by
+   reading data from returned endpoints in order from front to
+   back. Note that a client may ignore ``FlightInfo.ordered``. If an
+   ordering is important and the client may ignore
+   ``FlightInfo.ordered``, applications should return data in a single
+   endpoint.
```
lidavidm, line 99: Maybe to be clear, leave in an "Otherwise, there is no ordering defined on endpoints or the data within"?

format/Flight.proto
```diff
@@ -266,14 +266,28 @@ message FlightInfo {
...
+   * "FlightInfo.ordered = true" or should return the all data in a
+   * single endpoint.
+   *
+   * Note that a client may ignore "FlightInfo.ordered = true". If an
+   * ordering is important for an application, an application must
+   * choose one of them:
+   *
+   * * An application requires that all clients must read data in
+   *   returned endpoints order.
+   * * An application must return the all data in a single endpoint.
```
lidavidm, line 279: Compared to the wording in Flight.rst, it is not as clear what this is trying to say

dev/archery/archery/integration/runner.py
```diff
@@ -430,6 +430,11 @@ def run_all_tests(with_cpp=True, with_java=True, with_js=True,
             "middleware",
             description="Ensure headers are propagated via middleware.",
         ),
+        Scenario(
+            "ordered",
+            description="Ensure FlightInfo.ordered is supported.",
+            skip={"Java", "JS", "C#", "Go", "Rust"},
```
lidavidm, line 436: I can put together Java support too.

cpp/src/arrow/flight/integration_tests/test_integration.cc
```diff
@@ -271,6 +274,148 @@ class MiddlewareScenario : public Scenario {
...
+  Status MakeClient(FlightClientOptions* options) override { return Status::OK(); }
+
+  Status RunClient(std::unique_ptr<FlightClient> client) override {
+    ARROW_ASSIGN_OR_RAISE(auto info,
+                          client->GetFlightInfo(FlightDescriptor::Command("ordered")));
+    const auto& endpoints = info->endpoints();
+    std::vector<FlightEndpoint> ordered_endpoints(endpoints.size());
+    // If the data is ordered, read data from front to back.
+    // If the data is NOT ordered, read data from back to front.
+    if (info->ordered()) {
```
lidavidm, line 363: Shouldn't we be asserting that this is true here?

cpp/src/arrow/flight/sql/example/sqlite_server.cc
```diff
@@ -305,7 +305,7 @@ class SQLiteFlightSqlServer::Impl {
                           EncodeTransactionQuery(query, command.transaction_id));
     std::vector<FlightEndpoint> endpoints{FlightEndpoint{std::move(ticket), {}}};
     ARROW_ASSIGN_OR_RAISE(auto result,
-                          FlightInfo::Make(*schema, descriptor, endpoints, -1, -1))
+                          FlightInfo::Make(*schema, descriptor, endpoints, -1, -1, false))
```
lidavidm, line 308: SQLite doesn't tell us whether the result set is sorted or not, so I suppose technically this should always be 'true'? (That said, this server also only ever returns one endpoint so it's moot either way.)

lidavidm (approved):
LGTM! Just a few minor comments.

cpp/src/arrow/flight/types.h
```diff
@@ -554,6 +556,9 @@ class ARROW_FLIGHT_EXPORT FlightInfo {
   /// The total number of bytes in the dataset. If unknown, set to -1
   int64_t total_bytes() const { return data_.total_bytes; }
 
+  /// Whether endpoints are in the same order as the data.
+  bool ordered() const { return data_.ordered; }
```
wgtmac, line 560: I am not familiar with flight so maybe a dumb question: does the user know which order it is? Like ascending/descending and what are the sorted columns?

lidavidm: Good question, but no, this is just a semantic flag to indicate that (unlike the current status) there is _some_ ordering on the dataset. It's still up to the application to define what that means.

cpp/src/arrow/flight/integration_tests/test_integration.cc
kou: Ah, it may be a more straightforward approach.
I'll use the approach.

cpp/src/arrow/flight/sql/example/sqlite_server.cc
kou: We can use `true` here because this flight server always returns one endpoint.
But I think that `false` is better here because `false` means "unknown order".

A real Flight SQL server will set `true` here only when it analyzes the given query and detects `ORDER BY` in a main `SELECT`.
How about adding a comment something like "TODO: Set ordered to true only when the query has ORDER BY" here?

dev/archery/archery/integration/runner.py
kou: Wow! Thanks!
Could you push it to this branch when you work on it? Or you can create another branch for it.

docs/source/format/Flight.rst
kou: OK!

format/Flight.proto
kou: I've copied some missing descriptions from `Flight.rst`.

lidavidm (approved):
LGTM! Thanks!

kou:

kou:
I took a look at the Go implementation. It seems that we don't need to change the Go implementation for this. (We just need to regenerated `Flight.pb.go`.)

I'll add a Go integration test for this before we start a discussion for this proposal.

kou:
Preview URL: http://crossbow.voltrondata.com/pr_docs/35178/format/Flight.html

go/arrow/internal/flight_integration/scenario.go
```diff
@@ -526,6 +528,168 @@ func (m *middlewareScenarioTester) GetFlightInfo(ctx context.Context, desc *flig
...
+			"Actual:\n" +
+			"%s\n" +
+			"num-rows: %d\n" +
+			"num-cols: %d",
+			expected_table.Schema(),
+			expected_table.NumRows(),
+			expected_table.NumCols(),
+			table.Schema(),
+			table.NumRows(),
+			table.NumCols())
```
kou, line 625: If we have `arrow.Table.ToString()`, we can simplify this.

zeroshade: I agree, we should definitely add a `String()` method to `arrow.Table`. Do you think it's worth adding it to this PR? or should we just make an issue and have it as a separate change?

go/arrow/flight/server.go
```diff
@@ -52,14 +52,9 @@ type (
 	Empty                           = flight.Empty
 )
 
-// FlightService_ServiceDesc is the grpc.ServiceDesc for the FlightService
-// server. It should only be used for direct call of grpc.RegisterService,
-// and not introspected or modified (even as a copy).
-var FlightService_ServiceDesc = flight.FlightService_ServiceDesc
-
```
zeroshade, line 59: Why is this being removed? We added it to make it easy for others to incorporate a Flight Server into an existing grpc service.

go/arrow/internal/flight_integration/scenario.go
zeroshade: for now, maybe just marshal the table to JSON for the error output?

kou: I think that separated issue is better to focus on "ordered data support" in this pull request.
We don't have `array.TableToJSON()` for now... (We have `array.RecordToJSON()`.

go/arrow/flight/server.go
kou: I got the following error with this:

```text
 arrow/flight/server.go:58:40: undefined: flight.FlightService_ServiceDesc
```

`FlightService_ServiceDesc` was renamed to `_FlightService_ServiceDesc` (`_` was prepended) in generated `Flight_grpc.pb.go`: https://github.com/apache/arrow/pull/35178/files#diff-3b9deca07311cc509c9563b866266d044d8d368cc3225f8c4524172576c50f43R632

And we can't use it because it's not exported:

```text
arrow/flight/server.go:58:40: _FlightService_serviceDesc not exported by package flight
```

If this is important for us, could you regenerate `go/arrow/flight/internal/flight/{Flight.pb.go,Flight_grpc.pb.go}` with old `protoc` and push them to this branch? I don't have old `protoc`...

See also https://github.com/apache/arrow/pull/35178/files#diff-8739a8ac1525784934c2c972465c29a068f72cb631ab2e6b36f74cfb2849abaeR21 for `protoc` versions.

go/arrow/internal/flight_integration/scenario.go
kou: Created an issue for `arrow.Table.ToString()`: https://github.com/apache/arrow/issues/35296

dev/archery/archery/integration/runner.py
lidavidm: https://github.com/kou/arrow/pull/11

kou: Thanks!
Merged.

go/arrow/flight/server.go
zeroshade: Interesting, i see that the updated versions for the `Register` method actually takes a `*grpc.Server` which should likely be sufficient to add a flight service to an existing server. So I think this is fine, but we should definitely mark this as a breaking change as a result.

kou: OK.
I also add "**This PR includes breaking changes to public APIs.**" to the pull request description.

docs/source/format/Flight.rst
```diff
@@ -90,9 +90,14 @@ A client that wishes to download the data would:
...
-   within, so if the dataset is sorted, applications should return
-   data in a single endpoint.
+   requested. If ``FlightInfo.ordered`` is set, returned endpoints are
+   in the same order as the data. Otherwise, there is no ordering
+   defined on endpoints or the data within. The client can read
+   ordered data by reading data from returned endpoints in order from
+   front to back. Note that a client may ignore
+   ``FlightInfo.ordered``. If an ordering is important and the client
+   may ignore ``FlightInfo.ordered``, applications should return data
+   in a single endpoint.
```
alamb, line 100: Here is some suggested updates to this wording based on the mailing list discussion https://lists.apache.org/thread/0q1s84p17rtdz3q81wvylckob1hsx83z

I am not sure this is correct but I tried to encode my understanding

```suggestion
   requested. 
   
   If ``FlightInfo.ordered`` is true, this signals there is some
   order between data from different endpoints. 
   Clients should produce the same results as if the data returned 
   from each of the endpoints was concatenated, in order, from front to back. 
   If ``FlightInfo.ordered`` is not set, the client may return data  from from any 
   of the endpoints in arbitrary order. Data from any specific endpoint
   must be returned in order, but the data from different endpoints may be 
   interleaved to allow parallel fetches.  
   Note that since some clients may ignore ``FlightInfo.ordered``,
   if ordering is important and client support can not be ensured, 
   servers should return a single endpoint.
```

kou: Thanks!!!

> data  from from

Is this a typo of `data from`?

@lidavidm Could you also review this suggestion?

lidavidm: Thanks for the suggestion - the new wording sounds good to me (Kou already pointed out the typo)

alamb: Updated wording:

```suggestion
   requested. 
   
   If ``FlightInfo.ordered`` is true, this signals there is some
   order between data from different endpoints. 
   Clients should produce the same results as if the data returned 
   from each of the endpoints was concatenated, in order, from front to back. 
   If ``FlightInfo.ordered`` is not set, the client may return data  from any 
   of the endpoints in arbitrary order. Data from any specific endpoint
   must be returned in order, but the data from different endpoints may be 
   interleaved to allow parallel fetches.  
   Note that since some clients may ignore ``FlightInfo.ordered``,
   if ordering is important and client support can not be ensured, 
   servers should return a single endpoint.
```

kou:

docs/source/format/Flight.rst
kou: Thanks! Applied!

docs/source/format/Flight.rst
```diff
@@ -117,7 +129,9 @@ A client that wishes to download the data would:
    The client must consume all endpoints to retrieve the complete data
    set. The client can consume endpoints in any order, or even in
    parallel, or distribute the endpoints among multiple machines for
-   consumption; this is up to the application to implement.
+   consumption; this is up to the application to implement. The client
+   can also use ``FlightInfo.ordered``. See the previous item for
+   details of ``FlightInfo.ordered``.
```
kou, line 134: I forgot to update this sentence...
If anyone has a suggestion for wording, please share it.

kou:

kou:
The vote thread: https://lists.apache.org/thread/gnxv7cm5l01zgfos7mw52gr2337qofln

The vote result (passed):  https://lists.apache.org/thread/v05hxvffc118qtd4z4yqclryb7qslgfv

I'll merge this.

alamb:
Thank you @kou

PR #35183, opened 2023-04-17
GH-35175: [C#] Decimal128 / 256 Types should alloc dynamic byteWidth based on precision and scale

### LINKED ISSUE #35175: [C#] Decimal128 / 256 Builders should alloc dynamic byteWidth based on precision and scale

Builder(Decimal128Type type) : base(type, 16)
Builder(Decimal256Type type) : base(type, 32)

### Component(s)

C#

Platob:
Compute from precision and scale to get the optimal byteWidth to store values instead of forcing to 16 and 32

It is tested in Arrow.Tests.Types

It should not impact end users since its stored in variable ByteWidth

* Closes: #35175

### DIFF
--- a/csharp/test/Apache.Arrow.Tests/Decimal128ArrayTests.cs
+++ b/csharp/test/Apache.Arrow.Tests/Decimal128ArrayTests.cs
@@ -14,7 +14,6 @@
 // limitations under the License.
 
 using System;
-using System.Collections.Generic;
 using Apache.Arrow.Types;
 using Xunit;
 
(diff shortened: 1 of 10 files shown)


csharp/test/Apache.Arrow.Tests/Decimal128ArrayTests.cs
```diff
@@ -41,7 +40,7 @@ public void AppendThenGetGivesNull()
                     var array = builder.Build();
 
                     Assert.Equal(3, array.Length);
-                    Assert.Equal(array.Data.Buffers[1].Length, array.ByteWidth * 3);
+                    Assert.Equal(array.Data.Buffers[1].Length, array.ByteWidth * 3 + array.ByteWidth - 1);
```
wjones127, line 43: What is the purpose of this PR? It seems like it makes this computation unnecessarily complex. Decimals are either 16 or 32 bytes per the Arrow specification.

Platob: yeah indeed seeing cpp implementation its fixed

I saw in ParquetSharp implementation returning a Decimal128Type with 9 byteWidth for precision = 15 and scale = 5 which was different from 16

Platob:
Not conform with Arrow specs

PR #35191, opened 2023-04-17
GH-35190: [Go] Correctly handle null values in CSV reader

### LINKED ISSUE #35190: [Go] CSV Reader fails to read empty values as null

When the CSV reader is passed empty string as one of the possible null values, it still returns this as an empty string rather than null, even when used with the `WithNullReader(stringsCanBeNull=true)` option. 

### Component(s)

Go

hermanschaaf:
The way the CSV reader handles null values currently makes it impossible for empty strings to be interpreted as null, even when this is explicitly enabled through the `csv.WithNullReader(stringsCanBeNull: true, "", "NULL", "null")` option. This change fixes this.
* Closes: #35190


go/arrow/csv/reader.go
```diff
@@ -767,7 +767,7 @@ func (r *Reader) parseList(field array.Builder, str string) {
 
 func (r *Reader) parseBinaryType(field array.Builder, str string) {
 	// specialize the implementation when we know we cannot have nulls
-	if str != "" && r.isNull(str) {
+	if r.isNull(str) {
 		field.AppendNull()
 		return
 	}
```
zeroshade, line 773: Do we currently have a test that verifies you can still get an empty string if it is not in the list of Null values? / `stringsCanBeNull` == `false` ?

hermanschaaf: I think not, but happy to add one here

hermanschaaf: Added two new test cases for this

zeroshade (approved):

Issue #34830, opened 2023-03-31
[C++] Expand test coverage for FieldPath and FieldRef Get/Find methods


benibus:
### Describe the enhancement requested

The `FieldRef::Find*` and `FieldPath::Get` functions provide many different overloads, however, many of these paths don't appear to be covered in `arrow-type-test`.

This is more of a concern for `FieldPath::Get`, which extracts child fields from arrays, chunked arrays, record batches, tables, etc. The underlying implementations aren't exactly trivial - nor are they equivalent. In addition, there's an (unwritten?) expectation that these functions are zero-copy so perhaps this should be tested explicitly.

In both cases, we could utilize test fixtures to do this in a uniform way.

### Component(s)

C++

thisisnic:
This issue has been marked as stale because it has had no activity in the past 2 years. Please remove the stale label or comment below, or this issue will be closed in 14 days.

PR #35197, opened 2023-04-17
GH-14946: [C++] Add flattening FieldPath/FieldRef::Get methods

### LINKED ISSUE #14946: [C++] Should FieldRef::Get* for a StructArray return the "raw" child array or the "flattened" field array?
Related to the discussion in https://github.com/apache/arrow/pull/14697#issuecomment-1346633393 (but also to the python equivalent APIs like in https://github.com/apache/arrow/pull/14781#issuecomment-1339114243). 

Currently, the `FieldRef`/`FieldPath` methods to get this field from a StructArray returns the child array (i.e. `FieldRef::GetOne/GetAll` or `FieldPath::Get()` when passing those methods a StructArray (or RecordBatch with struct typed field). 
Basically, under the hood, it does `struct_array.data().child_data[idx]`.

On the other hand, if you use the same `FieldRef`/`FieldPath` object in a compute context, and for example pass this to the `struct_field` kernel, the result you get is the "flattened" field array (I find the "flatten" term somewhat confusing), which combines the top-level validity bitmap of the struct array with the validity bitmap of the child array, to ensure you have a correct indication of missing values in the result.  
Basically, under the hood this kernel is doing `struct_array.GetFlattenedField(idx)` instead.

So I am wondering if those two usage patterns of FieldRefs should be consistent? Although that probably depends on what the intended use case is for the direct FieldRef::GetOne / FieldPath::Get ? (currently, I don't see those used in actual code, except as helper in some testing code)

cc @pitrou @westonpace @lidavidm 

### Component(s)

C++

benibus:
### Rationale for this change

The current `FieldPath::Get` methods - when extracting nested child values - don't combine the child's null bitmap with higher-level parent bitmaps. While this is often preferable (it allows for zero-copy), there are cases where higher level "flattening" version is useful - e.g. when pre-processing sort keys for structs.

- Adds `FieldPath::GetFlattened` methods alongside the existing  `FieldPath::Get` overloads
- Adds `GetAllFlattened`, `GetOneFlattened` and `GetOneOrNoneFlattened` methods to `FieldRef`
- Adds a couple internal helpers for dealing with both `Get` variants in templates
- Overhauls the `FieldPath` tests in an effort to improve coverage and generalize across the supported input types

More significantly, this alters the `FieldPathGetImpl` internals to use a new `NestedSelector` class. The reason for this is that the prior method required presenting a vector of instantiated child values for each depth level prior to selection. With support for flattening (and recently, `ChunkedArrays`), this becomes a problem since we need to explicitly create all of those child values for each depth level despite the fact that we're only going to select one. So these changes allow any expensive instantiations to be deferred to selection time.

This also indirectly solves an issue that surfaced in the new tests, which is that `FieldPath::Get` would return incorrect nested values when sliced `Array`s are involved. This is because the underlying child data's offset/length weren't being adjusted based on the parent.

### Are these changes tested?

Yes (tests are included)

Yes, this adds methods to a public API
* Closes: #14946


benibus:
Probably worth mentioning that this (indirectly) addresses most of https://github.com/apache/arrow/issues/34830 as well - at least on the `FieldPath` side of things.

benibus:
@zeroshade @felipecrv

cpp/src/arrow/type.cc
```diff
@@ -1328,20 +1363,54 @@ Result<std::shared_ptr<Array>> FieldPath::Get(const Array& array) const {
...
-  auto columns = ChunkedArrayRef(chunked_array).Flatten();
+  auto columns = ChunkedArrayRef(chunked_array).FlattenZeroCopy();
   return FieldPathGetImpl::Get(this, columns);
 }
 
+Result<std::shared_ptr<Array>> FieldPath::GetFlattened(const Array& array) const {
+  if (array.type_id() != Type::STRUCT) {
+    return FieldPathGetImpl::NonStructError();
+  }
+  auto&& struct_array = checked_cast<const StructArray&>(array);
```
felipecrv, line 1384: `&&`?

cpp/src/arrow/type.h
```diff
@@ -1698,10 +1698,41 @@ class ARROW_EXPORT FieldPath {
...
+
  private:
   std::vector<int> indices_;
 };
 
+namespace internal {
+
+template <typename T>
+using FieldPathGetType =
+    decltype(std::declval<FieldPath>().Get(std::declval<T>()).ValueOrDie());
```
felipecrv, line 1721: I have a hard time inferring what comes out of this. An explicit type trait struct with specializations would be more informative and produce cleaner error messages for users.

felipecrv: Only now I noticed you've just moved this code here.

cpp/src/arrow/type.cc
```diff
@@ -1129,18 +1129,20 @@ class ChunkedArrayData : public ChunkedColumn {
...
     ArrayDataVector chunks(num_chunks());
     for (int chunk_idx = 0; chunk_idx < num_chunks(); ++chunk_idx) {
-      const auto& child_data = chunk(chunk_idx)->child_data;
-      DCHECK_EQ(columns.size(), child_data.size());
-      DCHECK(child_type->Equals(child_data[column_idx]->type));
-      chunks[chunk_idx] = child_data[column_idx];
+      const auto& parent = chunk(chunk_idx);
+      const auto& children = parent->child_data;
+      DCHECK_EQ(columns.size(), children.size());
+      auto child = children[column_idx]->Slice(parent->offset, parent->length);
```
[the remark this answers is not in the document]
benibus: Yes (a bug that I introduced in https://github.com/apache/arrow/pull/34537). It gives incorrect results when retrieving a nested child from an array sliced from another array.

Although, now that I'm thinking about it, I'd imagine the same problem would occur for the non-chunked `FieldPath::Get` variants as well. I'll need to test it and create a separate issue.

cpp/src/arrow/type.cc
benibus: Equivalent to `const auto&` in this case, but yeah... probably less clear and not exactly necessary (same for some of the test code).

felipecrv:
@benibus can you ping the people who understood the issue in this PR?

benibus:
@westonpace I feel you're probably most qualified to look at this given your comments on the original issue.

benibus:
cc @pitrou

cpp/src/arrow/type_test.cc
```diff
@@ -364,152 +365,344 @@ TEST(TestField, TestMerge) {
...
+  }
+  template <typename I>
+  void AssertOutputsEqual(const std::shared_ptr<Array>& expected,
+                          const std::shared_ptr<Array>& actual) const {
+    AssertArraysEqual(*expected, *actual);
+  }
+  template <typename I>
+  void AssertOutputsEqual(const std::shared_ptr<ChunkedArray>& expected,
+                          const std::shared_ptr<ChunkedArray>& actual) const {
+    if constexpr (std::is_same_v<I, ChunkedArray>) {
```
pitrou, line 582: Can you comment to explain this to the reader?

cpp/src/arrow/type_test.cc
```diff
@@ -693,25 +851,6 @@ TEST(TestFieldRef, DotPathRoundTrip) {
...
-
-  // retrieving fields with nested indices
-  EXPECT_EQ(FieldPath({0}).Get(s), f0);
-  EXPECT_EQ(FieldPath({1, 0}).Get(s), f1_0);
-  EXPECT_EQ(FieldPath({2, 0}).Get(s), f2_0);
-  EXPECT_EQ(FieldPath({2, 1, 0}).Get(s), f2_1_0);
-  EXPECT_EQ(FieldPath({2, 1, 1}).Get(s), f2_1_1);
-}
-
 TEST(TestFieldRef, Nested) {
```
pitrou, line 854: It seems quite a lot of effort was expended to test `FieldPath` comprehensively (thanks a lot for that!), but the additional `FieldRef` methods do not seem to be exercised here? (am I mistaken?)

cpp/src/arrow/type_test.cc
```diff
@@ -364,152 +365,344 @@ TEST(TestField, TestMerge) {
...
-  ASSERT_OK_AND_ASSIGN(auto v2_0, FieldPath({2, 0}).Get(*chunked_array));
-  ASSERT_OK_AND_ASSIGN(auto v2_1, FieldPath({2, 1}).Get(*chunked_array));
-  ASSERT_OK_AND_ASSIGN(auto v2_1_0, FieldPath({2, 1, 0}).Get(*chunked_array));
-
-  for (const auto& v : {v0, v1, v2_0, v2_1, v2_1_0}) {
-    EXPECT_EQ(v->num_chunks(), chunked_array->num_chunks());
+
+  static std::shared_ptr<ChunkedArray> SliceToChunkedArray(const Array& array,
+                                                           int num_chunks) {
+    ARROW_CHECK(num_chunks > 0 && array.length() >= num_chunks);
```
pitrou, line 502: Generally we should favor `ASSERT` or `EXPECT` macros in tests, since they should arrow the tests to proceed (unless the test would otherwise crash just after the failed check).

cpp/src/arrow/type_test.cc
```diff
@@ -364,152 +365,344 @@ TEST(TestField, TestMerge) {
...
+class TestFieldPath : public FieldPathTestFixture {
+ protected:
+  template <typename I, bool Flattened = false>
+  void TestGetWithInvalidIndex() const {
+    const auto& input = case_->GetInput<I>();
+    for (const auto& path :
+         {FieldPath({2, 1, 0}), FieldPath({1, 2, 0}), FieldPath{1, 1, 2}}) {
+      EXPECT_RAISES_WITH_MESSAGE_THAT(IndexError,
+                                      ::testing::HasSubstr("index out of range"),
+                                      internal::GetChild<Flattened>(*input, path));
```
pitrou, line 602: Since the `IndexError` has a more detailed error message, is it possible to test it precisely at least for a hard-coded case?

pitrou (changes requested):
Thanks a lot for this work! This looks generally good, just a bunch of relatively minor suggestions.

cpp/src/arrow/type.h
```diff
@@ -1698,10 +1698,45 @@ class ARROW_EXPORT FieldPath {
...
+template <bool Flattened, typename T>
+std::enable_if_t<!Flattened, Result<FieldPathGetType<T>>> GetChild(
+    const T& root, const FieldPath& path, MemoryPool* = NULLPTR) {
+  return path.Get(root);
+}
+template <bool Flattened, typename T>
+std::enable_if_t<Flattened, Result<FieldPathGetType<T>>> GetChild(
+    const T& root, const FieldPath& path, MemoryPool* pool = NULLPTR) {
+  return path.GetFlattened(root, pool);
+}
```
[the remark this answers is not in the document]
benibus: Yeah, unfortunately `if constexpr` wouldn't work here since `GetFlattened` doesn't have overloads for `Schema`, `Field`, etc.

But I'm fine with just duplicating the definitions as well (in that case, the helpers probably wouldn't be too useful outside of the tests so no need to expose them).

cpp/src/arrow/type_test.cc
benibus: Agreed. I wrote the tests before adding the `FieldRef` methods and then neglected to come back to them... Fortunately, it should be possible to use the same test case (for the most part).

cpp/src/arrow/type_test.cc
benibus: This function is only used to generate the shared test case so I figured that aborting here would be appropriate (otherwise, anything could happen).

cpp/src/arrow/type.h
```diff
@@ -1886,6 +1902,15 @@ class ARROW_EXPORT FieldRef : public util::EqualityComparable<FieldRef> {
     }
     return out;
   }
+  template <typename T>
+  std::vector<GetType<T>> GetAllFlattened(const T& root,
+                                          MemoryPool* pool = NULLPTR) const {
+    std::vector<GetType<T>> out;
+    for (const auto& match : FindAll(root)) {
+      out.push_back(match.GetFlattened(root, pool).ValueOrDie());
```
pitrou, line 1910: Uh. Unlike `GetAll`, this can fail for various reasons such as failure to allocate enough memory. In that case, we'd probably want to return an error instead of dying out.

cpp/src/arrow/type.h
```diff
@@ -1905,6 +1935,15 @@ class ARROW_EXPORT FieldRef : public util::EqualityComparable<FieldRef> {
...
     return match.Get(root).ValueOrDie();
   }
+  template <typename T>
+  Result<GetType<T>> GetOneOrNoneFlattened(const T& root,
+                                           MemoryPool* pool = NULLPTR) const {
+    ARROW_ASSIGN_OR_RAISE(auto match, FindOneOrNone(root));
+    if (match.empty()) {
+      return static_cast<GetType<T>>(NULLPTR);
+    }
+    return match.GetFlattened(root, pool).ValueOrDie();
```
pitrou, line 1945: Same here.

cpp/src/arrow/type_test.cc
```diff
@@ -364,152 +365,341 @@ TEST(TestField, TestMerge) {
   }
 }
 
-TEST(TestFieldPath, Basics) {
-  auto f0 = field("alpha", int32());
-  auto f1 = field("beta", int32());
-  auto f2 = field("alpha", int32());
-  auto f3 = field("beta", int32());
-  Schema s({f0, f1, f2, f3});
+struct FieldPathTestCase {
```
pitrou, line 368: Given that this is adding quite a lot of test code, and this file is already long, I think it may be time to move the `Field{Ref,Path}` tests to a separate test module.

cpp/src/arrow/type.h
benibus: In that case, I'm a little confused as to why the non-flattened variants don't forward those errors either - since the standard `FieldPath::Get` methods can also fail (which was true prior to this PR).

I'm mostly referring to `GetOne` and `GetOneAndNone` here, as they already return a `Result`. Regardless, I'll propagate those errors for the new methods.

pitrou (approved):
+1, thanks for the update! Will merge if CI is green.

pitrou:
CI failures are unrelated.

PR #35210, opened 2023-04-18
GH-35209: [Python] Add a type alias for pa.dictionary(pa.int32(), pa.string())

randolf-scholz:
Fixes #35209
* Closes: #35209

### DIFF
--- a/python/pyarrow/tests/test_csv.py
+++ b/python/pyarrow/tests/test_csv.py
@@ -952,6 +952,23 @@ class BaseCSVTableRead(BaseTestCSV):
             'b': [2.5, -3.0, None],
         }
 
+    def test_dictionary(self):
+        # Infer floats with a custom decimal point
+        parse_options = ParseOptions(delimiter=';')
+        rows = b"a;b\nx;x\ny;y\nx;x"
+
+        schema = pa.schema([('a', pa.dictionary(pa.int32(), pa.string())),
+                            ('b', "dictionary[int32,string]")])
+        convert_options = ConvertOptions(column_types=schema)
+        table = self.read_bytes(rows, parse_options=parse_options,
+                                convert_options=convert_options)
+
+        assert table.schema == schema
+        assert table.to_pydict() == {
+            'a': ["x", "y", "x"],
+            'b': ["x", "y", "x"],
+        }
+
     def test_simple_timestamps(self):
         # Infer a timestamp column
         rows = (b"a,b,c\n"
--- a/python/pyarrow/types.pxi
+++ b/python/pyarrow/types.pxi
@@ -4790,7 +4790,7 @@ cdef dict _type_aliases = {
     'duration[us]': duration('us'),
     'duration[ns]': duration('ns'),
     'month_day_nano_interval': month_day_nano_interval(),
-    'dictionary[int32,string]' : dictionary(int32, string),
+    'dictionary[int32,string]' : dictionary(int32(), utf8()),
 }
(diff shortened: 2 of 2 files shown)


python/pyarrow/tests/test_csv.py
```diff
@@ -952,6 +952,23 @@ def test_decimal_point(self):
             'b': [2.5, -3.0, None],
         }
 
+    def test_dictionary(self):
```
danepitkin, line 955: Can you add to the test case `test_schema.py::test_type_for_alias()`? IMO you can delete this test afterwards.

python/pyarrow/types.pxi
```diff
@@ -4790,6 +4790,7 @@ cdef dict _type_aliases = {
     'duration[us]': duration('us'),
     'duration[ns]': duration('ns'),
     'month_day_nano_interval': month_day_nano_interval(),
+    'dictionary[int32,string]': dictionary(int32(), utf8()),
```
danepitkin, line 4793:
```suggestion

    # Composite types
    'dictionary[int32,string]': dictionary(int32(), utf8()),
```

danepitkin: I wonder if it makes sense to separate composite types from Arrow's primitive types into a different list? Theoretically we could auto generate them for completeness, although I wouldn't want to introduce unnecessary computation on startup. We could lazy-load composite types on usage.

danepitkin (commented):
This is the first composite type alias added to this list. I don't see any issue in adding it, but also it leaves the user wondering why only a subset is implemented.

python/pyarrow/tests/test_csv.py
randolf-scholz: > This is the first composite type alias added to this list. I don't see any issue in adding it, but also it leaves the user wondering why only a subset is implemented.

I wrote about this in the issue, please correct me if I am wrong, but it seems that `dictionary[int32, string]` is the "standard" dictionary type that has some useful kernels. At least I noticed horrible performance degradation when trying to read CSV data with `int16`, `uint32` or `int64` indexed dictionary types. So adding only `dictionary[int32, string]` would be to protect the user from using unoptimized composite types. (If this is a bug, I can open another issue).

danepitkin:
Sorry to go back on this, but after some more thought I'm a bit hesitant to add composite type string aliases. There's hundreds of possibilities (or more) and there can also be several variations of string representations each (e.g. spaces vs no spaces). We also aren't matching the string representation of `pa.dictionary.__str__` here, which looks something like this: `DictionaryType(dictionary<values=string, indices=int32, ordered=0>)`. 

@jorisvandenbossche @AlenkaF any thoughts here?

randolf-scholz:
@danepitkin I'd add `dictionary[string,int32]` since it is likely the most common one. Maybe one should add `dictionary[int8,string]`, `dictionary[int16,string]` and `dictionary[int64,string]` immediately as well? How common are non-string dictionaries anyway?

If all possible combinations with non-string types are needed one could write a custom lookup instead. 

Regarding the representation one could do either:

- differentiate `DictionaryType.__repr__` and `DictionaryType.__str__`, so that the latter returns the type-code
- add an additional `name` / `type_code` property to each dtype that returns the abbreviated type code
    - for example, numpy has `name` attribute
    - `np.dtype("int32").name` ⇝ "int32"
    - `np.dtype("datetime64[ns]").name` ⇝ "datetime64[ns]"

PR #35215, opened 2023-04-18
GH-35212: [Go] Add ability to show full call stack with ARROW_CHECKED_MAX_RETAINED_FRAMES

### LINKED ISSUE #35212: [Go] Add more frames to Checked Allocator memory leak output

The `CheckedAllocator` output right now provides a single line of context, roughly indicating which type was involved. This is fine for testing within Arrow itself, but in large applications this is often not enough to narrow down where a memory leak is coming from. It would be very helpful if the full call stack could be provided in the output.

It is technically possible to repeatedly run with different values of `ARROW_CHECKED_ALLOC_FRAMES` and `ARROW_CHECKED_REALLOC_FRAMES` but this is quite tedious.

### Component(s)

Go

hermanschaaf:
This adds support for printing the full call stack when a leak is reported by the checked memory allocator. An `ARROW_CHECKED_MAX_RETAINED_FRAMES` environment variable controls how many frames are retained. To keep this completely backwards-compatible, the default right now is zero. In this case, the reported error is exactly the same as before. When a higher value is given though, a longer call stack is printed. For example:

Before (same as `ARROW_CHECKED_MAX_RETAINED_FRAMES=0`):

```
checked_allocator.go:160: LEAK of 64 bytes FROM github.com/apache/arrow/go/v12/arrow/array.(*TimestampBuilder).newData line 2396
```

After (with `ARROW_CHECKED_MAX_RETAINED_FRAMES=100`):

```
checked_allocator.go:160: LEAK of 64 bytes FROM github.com/apache/arrow/go/v12/arrow/array.(*TimestampBuilder).newData line 2396
     github.com/apache/arrow/go/v12/arrow/array.(*TimestampBuilder).NewTimestampArray line 2386
     github.com/apache/arrow/go/v12/arrow/array.(*TimestampBuilder).NewArray line 2380
     github.com/apache/arrow/go/v12/arrow/array.(*RecordBuilder).NewRecord line 346
     github.com/cloudquery/plugin-sdk/v2/testdata.GenTestData line 243
     github.com/cloudquery/plugin-sdk/v2/plugins/destination.testMigration line 53
     github.com/cloudquery/plugin-sdk/v2/plugins/destination.(*PluginTestSuite).destinationPluginTestMigrate.func5 line 239
     testing.tRunner line 1576
     runtime.goexit line 1172
```
* Closes: #35212

### DIFF
--- a/go/arrow/memory/checked_allocator.go
+++ b/go/arrow/memory/checked_allocator.go
@@ -66,13 +66,13 @@ func (a *CheckedAllocator) Reallocate(size int, b []byte) []byte {
 		return out
 	}
 
+	newptr := uintptr(unsafe.Pointer(&out[0]))
 	a.allocs.Delete(oldptr)
-	ptr := uintptr(unsafe.Pointer(&out[0]))
 	pcs := make([]uintptr, maxRetainedFrames)
 	runtime.Callers(reallocFrames, pcs)
 	callersFrames := runtime.CallersFrames(pcs)
 	if pc, _, l, ok := runtime.Caller(reallocFrames); ok {
-		a.allocs.Store(ptr, &dalloc{pc: pc, line: l, sz: size, callersFrames: callersFrames})
+		a.allocs.Store(newptr, &dalloc{pc: pc, line: l, sz: size, callersFrames: callersFrames})
 	}
 
 	return out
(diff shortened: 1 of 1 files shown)


zeroshade (approved):
this looks awesome, thanks for this!

PR #35222, opened 2023-04-18
GH-35267: [C#] Serialize TotalBytes and TotalRecords in FlightInfo

### LINKED ISSUE #35267: [C#] TotalBytes and TotalRecords not being serialized by C# flight library

When converting from C# object to proto object the implementation forgot to set these properties.

### Component(s)

C#

jeremyosterhoudt:
Currently the `TotalBytes` and `TotalRecords` properties are not serialized into the `FlightInfo` ProtoBuf for C#/.NET.  This breaks compatibility with Arrow Flight implementation done in other languages (c++ for example).

This PR adds the `TotalBytes` and `TotalRecords` properties to the protoBuf `FlightInfo` for the C# implementation.
* Closes: #35267

### DIFF
--- a/csharp/src/Apache.Arrow.Flight/FlightInfo.cs
+++ b/csharp/src/Apache.Arrow.Flight/FlightInfo.cs
@@ -39,7 +39,7 @@ namespace Apache.Arrow.Flight
             TotalRecords = flightInfo.TotalRecords;
         }
 
-        public FlightInfo(Schema schema, FlightDescriptor descriptor, IReadOnlyList<FlightEndpoint> endpoints, long totalRecords = 0, long totalBytes = 0)
+        public FlightInfo(Schema schema, FlightDescriptor descriptor, IReadOnlyList<FlightEndpoint> endpoints, long totalRecords = -1, long totalBytes = -1)
         {
             Schema = schema;
             Descriptor = descriptor;
--- a/csharp/src/Apache.Arrow.Flight/FlightInfo.cs
+++ b/csharp/src/Apache.Arrow.Flight/FlightInfo.cs
@@ -64,7 +64,9 @@ namespace Apache.Arrow.Flight
             var response = new Protocol.FlightInfo()
             {
                 Schema = serializedSchema,
-                FlightDescriptor = Descriptor.ToProtocol()
+                FlightDescriptor = Descriptor.ToProtocol(),
+                TotalBytes = TotalBytes,
+                TotalRecords = TotalRecords
             };
 
             foreach(var endpoint in Endpoints)
--- a/csharp/test/Apache.Arrow.Flight.TestWeb/FlightHolder.cs
+++ b/csharp/test/Apache.Arrow.Flight.TestWeb/FlightHolder.cs
@@ -56,7 +56,7 @@ namespace Apache.Arrow.Flight.TestWeb
                 new FlightEndpoint(new FlightTicket(_flightDescriptor.Paths.FirstOrDefault()), new List<FlightLocation>(){
                     new FlightLocation(_location)
                 })
-            });
+            }, 2, 4);
         }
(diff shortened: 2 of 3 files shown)


csharp/test/Apache.Arrow.Flight.TestWeb/FlightHolder.cs
```diff
@@ -56,7 +56,7 @@ public FlightInfo GetFlightInfo()
                 new FlightEndpoint(new FlightTicket(_flightDescriptor.Paths.FirstOrDefault()), new List<FlightLocation>(){
                     new FlightLocation(_location)
                 })
-            });
+            }, 2, 4);
```
westonpace, line 59: I don't really know how this test works but is this verified anywhere?  Otherwise it seems you might as well use the defaults. Also, should you do `totalRecords: 2, totalBytes: 4` for future readability (I don't really know what the convention is in the C# implementation so feel free to ignore)?

westonpace (changes requested):
It would be nice if this addition was tested in some way.  Would that be possible?

csharp/test/Apache.Arrow.Flight.TestWeb/FlightHolder.cs
jeremyosterhoudt: There is an existing test that verifies all of the properties, but it's not obvious unless you dig into it.  The test has a few assumptions baked into test flight server that is spun up on startup that make it hard to inject an expected amount.  

However, I was able enhanced the test server to calculate the batch length and total bytes.  I've also added an additional test to ensure those values match what is expected for those fields.

csharp/test/Apache.Arrow.Flight.Tests/FlightTests.cs
```diff
@@ -312,5 +312,26 @@ await foreach(var recordBatch in getStream.ResponseStream)
...
+            List<FlightInfo> expectedFlightInfo = new List<FlightInfo>();
+
+            expectedFlightInfo.Add(GivenStoreBatches(flightDescriptor1, new RecordBatchWithMetadata(expectedBatch)));
+
+            var listFlightStream = _flightClient.ListFlights();
+
+            var actualFlights = await listFlightStream.ResponseStream.ToListAsync();
+            var result = actualFlights.First();
+
+            //Expected values can be found in the FlightHolder class
```
westonpace, line 332:
```suggestion
```
I'm not sure this comment is still needed since it appears you moved away from hard-coded values.

csharp/src/Apache.Arrow.Flight/FlightInfo.cs
```diff
@@ -64,7 +64,9 @@ internal Protocol.FlightInfo ToProtocol()
             var response = new Protocol.FlightInfo()
             {
                 Schema = serializedSchema,
-                FlightDescriptor = Descriptor.ToProtocol()
+                FlightDescriptor = Descriptor.ToProtocol(),
+                TotalBytes = TotalBytes,
+                TotalRecords = TotalRecords
```
westonpace, line 69: This isn't related to your PR but should `totalRecords` and `totalBytes` default to -1 in the constructor?  From https://arrow.apache.org/docs/format/Flight.html it appears that `-1` is the value to use for "unknown".

westonpace (changes requested):
Thanks for adding the test.  Two small changes and then I think this is probably good to go.  However, this does not meet our criteria for `MINOR` PRs (these are limited to grammar, usage, spelling, and documentation).  Can you go ahead and also create an issue for this?

westonpace (approved):
This looks good now.  I went ahead and added an issue since it isn't technically minor.  I'll leave it open for a few days in case @eerhardt wants to take a look.  Otherwise I'll merge this in next week.

PR #35223, opened 2023-04-18
GH-35214: [CI][C++] Incremental clang-tidy

### LINKED ISSUE #35214: [CI][C++] Add clang-tidy and IWYU C++ CI jobs

AFAIK these two checks aren't run, but would be very valuable. However, both are slow, so it's most useful if we can run them only on files that have been changed. There is a script to detect changed files, but it is outdated:

https://github.com/apache/arrow/blob/ccc9d2c8da9220f9ed414ab8c71ed4af92fd5f34/cpp/build-support/get-upstream-commit.sh#L25

See also: https://github.com/apache/arrow/issues/16763

### Component(s)

C++, Continuous Integration

wjones127:
### Rationale for this change

Testing to see if we can make clang-tidy part of our workflow.

* Closes: #35214

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -340,6 +340,16 @@ if(${CLANG_TIDY_FOUND})
                     --compile_commands
                     ${CMAKE_BINARY_DIR}/compile_commands.json
                     ${COMMON_LINT_OPTIONS}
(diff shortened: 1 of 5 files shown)


wjones127:
Closing for now since I won't have time to pursue this.

PR #35230, opened 2023-04-19
GH-34949: [C++][Parquet] Enable page index by columns

### LINKED ISSUE #34949: [C++][Parquet] Parquet writer should control set of columns to enable page index

Once https://github.com/apache/arrow/pull/34054 is merged, it is nice to provide a writer option to select set of columns to enable page index.

### Component(s)

C++, Parquet

wgtmac:
### Rationale for this change

Currently parquet writer only supports enabling page index for all columns. It would be good to enable/disable at the column level as sometimes it may not be useful for some columns but it pays to create them.

Similar to `WriterProperties::Builder::enable_dictionary/disable_dictionary`, this patch adds `WriterProperties::Builder::enable_write_page_index/disable_write_page_index` and keep it backward compatible to enable/disable for all columns.

### Are these changes tested?

Added `ParquetPageIndexRoundTripTest::EnablePerColumn` to cover the new settings.

Yes, users are now more flexible to enable/disable page index.
* Closes: #34949

### DIFF
--- a/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
+++ b/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
@@ -26,6 +26,7 @@
 
 #include <cstdint>
 #include <functional>
+#include <set>
 #include <sstream>
 #include <vector>
 
@@ -5235,7 +5236,8 @@ class ParquetPageIndexRoundTripTest : public ::testing::Test {
     ASSERT_OK_AND_ASSIGN(buffer_, sink->Finish());
   }
 
-  void ReadPageIndexes(int expect_num_row_groups, int expect_num_pages) {
+  void ReadPageIndexes(int expect_num_row_groups, int expect_num_pages,
+                       const std::set<int>& expect_columns_without_index = {}) {
     auto read_properties = default_arrow_reader_properties();
     auto reader = ParquetFileReader::Open(std::make_shared<BufferReader>(buffer_));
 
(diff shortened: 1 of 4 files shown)


wjones127 (approved):
This looks good to me 👍

mapleFU:
By the way, can we just disable the column index?

I've a case that I don't want to collect the Column Index, because statistics is not important for me. ( I'm sure I'll read the whole file), however, offset index can be used.

wgtmac:
> By the way, can we just disable the column index?
> 
> I've a case that I don't want to collect the Column Index, because statistics is not important for me. ( I'm sure I'll read the whole file), however, offset index can be used.

I have thought about this. However, it would be complex if we want to control ColumnIndex and OffsetIndex separately for individual columns. What about splitting `WriterProperties::Builder::enable_write_page_index/disable_write_page_index` into `WriterProperties::Builder::enable_write_column_index/disable_write_column_index` and WriterProperties::Builder::enable_write_offset_index/disable_write_offset_index`?

mapleFU:
Well, I think just having Offset Index can optimizing IO, but I don't know how can we do when we only have Column Index

wgtmac:
Actually we can archive it by disabling statistics on all columns. Without column statistics, ColumnIndexes are dropped automatically.

mapleFU:
Yes, but seems it's a bit trickey here :)
Maybe we can document it?

PR #35231, opened 2023-04-19
GH-35234: [Go] Fix skip argument to Callers

### LINKED ISSUE #35234: [Go] Two additional frames need to be skipped in Checked Allocator output

When ARROW_CHECKED_MAX_RETAINED_FRAMES is set to a value greater than zero, the frames currently printed include the frame below the caller and the caller itself, which means the caller is listed twice. We should skip 2 additional frames in the `Callers` code as done in https://github.com/apache/arrow/pull/35231

### Component(s)

Go

hermanschaaf:
Follow-up to https://github.com/apache/arrow/pull/35215. It's mostly cosmetic, but without the additional 2 skips passed in to `Callers`, the Caller frame is repeated in the stack trace, and one frame below it is added as well. With this change, the checked allocator stack trace contains no duplicates.

I had this change locally but didn't realize it wasn't pushed to the PR branch, sorry about that :bow:
* Closes: #35234


zeroshade:
@hermanschaaf can you try rebaseing from main and seeing if that alleviates the test failure?

hermanschaaf:
@zeroshade Just updated to the latest main; looks like all the tests passed except for `Go / AMD64 Debian 11 Go 1.18`. The test failure still seems unrelated to this PR:

```
--- FAIL: TestPreparedStatementNoSchema (0.01s)
    driver_test.go:744: 
        	Error Trace:	/arrow/go/arrow/flight/flightsql/driver/driver_test.go:744
        	Error:      	Received unexpected error:
        	            	EOF
        	Test:       	TestPreparedStatementNoSchema
FAIL
FAIL	github.com/apache/arrow/go/v12/arrow/flight/flightsql/driver	1.150s
```

zeroshade (approved):
I agree, the failure is unrelated to this PR. if it fails in main after merging this I'll file an issue to look into it.

PR #35242, opened 2023-04-20
GH-35171: [C++][Parquet] Implement CRC for data page v2

### LINKED ISSUE #35171: [C++][Parquet] Parquet Support CRC on DataPage V2

Currently, we already support crc on DataPage V1 and Dictionary Page. Finally we need to support DataPage v2.
DataPage v2 is not used online, and parquet-testing not have data page v2 with crc, because parquet-mr didn't implement it. So we cannot use the file from parquet-testing.

### Component(s)

C++, Parquet

mapleFU:
### Rationale for this change

Implement crc for data page v2.

1. When writing, able to write crc for data page v2
2. When reading, able to read crc for data page v2
3. Testing

### Are these changes tested?

Yes, but parquet-testing don't have crc page v2 test. So no integration test

* Closes: #35171


mapleFU:
@wjones127 @wgtmac PTAL :)

mapleFU:
Test failed because parquet-mr in test file write "incorrect" Data Page crc in v2. I will disable them first.

mapleFU:
Oh, it's not that problem, damn...

mapleFU:
After re-create page header, the test below failed:

```
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] WriteDictionary/TestArrowWriteDictionary.Statistics/0, where GetParam() = 4-byte object <00-00 00-00>
[  FAILED  ] WriteDictionary/TestArrowWriteDictionary.Statistics/1, where GetParam() = 4-byte object <01-00 00-00>
```

mapleFU:
@westonpace Hi Pace, I find that `current_page_header_` message Deserialize will not mark `__isset` field to unset when meet a new message. When I change to use a new header, `TestArrowWriteDictionary.Statistics` failed. And when I go through this unittest, I found that I give a wrong stats.

For test index:

```C++
ArrayFromJSON(::arrow::int32(), R"([0, 1, null, 0, 1, null])"),
```

It generate:

```
RowGroup1 [ Page1 [0, 1] Page2 [ null ] ]
RowGroup2 [ Page1 [0, 1] Page2 [ null ] ]
```

So, I think page2 should not have min-max stats, and I fix it here. Would you mind take a look at this patch? Thanks

mapleFU:
I will find out why fuzzing failed..

mapleFU:
The fuzzing find out that:

```
/arrow/cpp/src/generated/parquet_types.cpp:4425:18: runtime error: load of value 16, which is not a valid value for type 'PageType::type'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /arrow/cpp/src/generated/parquet_types.cpp:4425:18 in 
1
```

I guess previous it already has that problem...

cpp/src/parquet/column_reader.cc
```diff
@@ -487,16 +489,15 @@ std::shared_ptr<Page> SerializedPageReader::NextPage() {
 
     const PageType::type page_type = LoadEnumSafe(&current_page_header_.type);
 
-    // TODO(PARQUET-594) crc checksum for DATA_PAGE_V2
-    if (properties_.page_checksum_verification() &&
-        (page_type == PageType::DATA_PAGE || page_type == PageType::DICTIONARY_PAGE) &&
-        current_page_header_.__isset.crc) {
+    if (properties_.page_checksum_verification() && current_page_header_.__isset.crc) {
```
wgtmac, line 492: This change works for standard parquet files. However, if user has customized parquet to support index page, this may break their workload.

cpp/src/parquet/file_deserialize_test.cc
```diff
@@ -814,6 +834,30 @@ TEST_F(TestPageSerde, DictCrcCheckNonExistent) {
                          /* verification_checksum */ true, /* has_dictionary */ true);
 }
 
+TEST_F(TestPageSerde, DataPageV2CrcCheckSuccessful) {
```
wgtmac, line 837: Should we also add test with `has_dictionary=true` for completeness?

wgtmac: nit: switching to parameterized test would be more favorable here.

mapleFU: > Should we also add test with `has_dictionary=true` for completeness?

Sure, let me try it

cpp/src/parquet/column_reader.cc
mapleFU: Yes, so should I extract a pageTypeHasCrc ?

mapleFU:
Gently ping @wjones127

mapleFU:
Comments solved.

mapleFU:
( https://github.com/apache/arrow/pull/35242#issuecomment-1518558830 this happens because deserialized a bad enum value doesn't trigger ubsan, but assign it to another object may trigger...)

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -4092,7 +4092,7 @@ TEST_P(TestArrowWriteDictionary, Statistics) {
       {{"b", "a"}, {"b", "a"}}, {{"c", "c"}, {"c", "c"}}, {{"d", "a"}, {"d", "a"}}};
   const std::vector<std::vector<std::vector<bool>>> expected_has_min_max_by_page = {
       {{true, true}, {true, true}},
-      {{true, true}, {true, true}},
+      {{true, false}, {true, false}},
```
wjones127, line 4095: Why did this change?

wjones127:
Most of the changes look good. But not sure why that test changed.

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
mapleFU: https://github.com/apache/arrow/pull/35242#issuecomment-1517491968

This is mentioned here. Let me explain it:
1. Previously, `PageHeader` reuse same object, so `__isset` would not been cleared
2. So, `has_min_max` will keep "true" if first page has min-max, and second page not has.
3. In test case, the second page should not has min-max, but it has been set. I think previous test is a bug here. You can check it again

wjones127: That makes sense. Thanks for explaining.

wjones127 (approved):

mapleFU:
I guess https://github.com/apache/arrow/actions/runs/4832925654/jobs/8612322729?pr=35242 and https://github.com/apache/arrow/actions/runs/4832925654/jobs/8612322729?pr=35242 are timeout, and https://github.com/apache/arrow/actions/runs/4832925659/jobs/8612318748?pr=35242 is pandas issue. But I'm not sure =_=

PR #35249, opened 2023-04-20
GH-35247: [C++] Add Arrow Substrait support for stddev/variance

### LINKED ISSUE #35247: [C++] Add Arrow Substrait support for stddev/variance

Recently, ibis-substrait added [support for stddev/variance](https://github.com/ibis-project/ibis-substrait/pull/583/). This issue will add support for the same in Arrow Substrait.

### Component(s)

C++

rtpsw:
See #35247
* Closes: #35247

### DIFF
--- a/cpp/src/arrow/engine/substrait/extension_set.cc
+++ b/cpp/src/arrow/engine/substrait/extension_set.cc
@@ -944,6 +944,26 @@ ExtensionIdRegistry::SubstraitAggregateToArrow DecodeBasicAggregate(
                                call.id().name, " to have at least one argument");
       }
       case 1: {
+        std::shared_ptr<compute::FunctionOptions> options = nullptr;
+        if (arrow_function_name == "stddev" || arrow_function_name == "variance") {
+          auto maybe_dist = call.GetOption("distribution");
+          if (maybe_dist) {
+            auto& prefs = **maybe_dist;
+            if (prefs.size() != 1) {
+              return Status::Invalid("expected a single preference for ",
+                                     arrow_function_name, " but got ", prefs.size());
+            }
+            int ddof;
+            if (prefs[0] == "POPULATION") {
+              ddof = 1;
+            } else if (prefs[0] == "SAMPLE") {
+              ddof = 0;
+            } else {
+              return Status::Invalid("unknown distribution preference ", prefs[0]);
+            }
+            options = std::make_shared<compute::VarianceOptions>(ddof);
+          }
+        }
         fixed_arrow_func += arrow_function_name;
 
         ARROW_ASSIGN_OR_RAISE(compute::Expression arg, call.GetValueArg(0));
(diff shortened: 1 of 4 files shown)


icexelloss (approved):
LGTM

ianmcook (changes requested):
This should pass `VarianceOptions` with `ddof` set to `1` when the `distribution` option is set to `POPULATION`

rtpsw:
> This should pass `VarianceOptions` with `ddof` set to `1` when the `distribution` option is set to `POPULATION`

Good catch. I'll work on adding this.

cpp/src/arrow/engine/substrait/extension_set.cc
```diff
@@ -944,6 +944,26 @@ ExtensionIdRegistry::SubstraitAggregateToArrow DecodeBasicAggregate(
                                call.id().name, " to have at least one argument");
       }
       case 1: {
+        std::shared_ptr<compute::FunctionOptions> options = nullptr;
+        if (arrow_function_name == "stddev" || arrow_function_name == "variance") {
+          auto maybe_dist = call.GetOption("distribution");
```
icexelloss, line 949: Can you add a link to the substrait spec here:
https://github.com/substrait-io/substrait/blob/73228b4112d79eb1011af0ebb41753ce23ca180c/extensions/functions_arithmetic.yaml#L1240

icexelloss:
Minor comments otherwise looks good

icexelloss:
@ianmcook @westonpace Do you want to take a look? Ideally I want to merge this soon.

ianmcook:
Thanks @rtpsw & @icexelloss! Is there a test exercising the `"preference": ["POPULATION"]` case? It would be nice to have that.

icexelloss:
> Thanks @rtpsw & @icexelloss! Is there a test exercising the `"preference": ["SAMPLE"]` case? It would be nice to have that.

Sure - are you ok with just copying the json plan (but with preference = SAMPLE) into another test case ?

ianmcook:
> > Thanks @rtpsw & @icexelloss! Is there a test exercising the `"preference": ["SAMPLE"]` case? It would be nice to have that.
> 
> Sure - are you ok with just copying the json plan (but with preference = SAMPLE) into another test case ?

That sounds fine to me!

cpp/src/arrow/engine/substrait/extension_set.cc
```diff
@@ -944,6 +944,30 @@ ExtensionIdRegistry::SubstraitAggregateToArrow DecodeBasicAggregate(
...
+          // https://github.com/substrait-io/substrait/blob/
+          // 73228b4112d79eb1011af0ebb41753ce23ca180c/
+          // extensions/functions_arithmetic.yaml#L1240
+          auto maybe_dist = call.GetOption("distribution");
+          if (maybe_dist) {
+            auto& prefs = **maybe_dist;
+            if (prefs.size() != 1) {
+              return Status::Invalid("expected a single preference for ",
+                                     arrow_function_name, " but got ", prefs.size());
+            }
```
westonpace, line 959: Technically multiple preferences means the user would prefer the first value but will be ok with the second.  In other words, it's similar to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept

So you should be able to handle this case.  However, no producers do this yet.

westonpace (approved):
This looks good.  You could maybe handle the multiple-preferences situation, but I don't see that being required for quite a while.  Let me know if you'd rather defer it.

PR #35253, opened 2023-04-20
GH-35252: [C++] Use FindGTestAlt.cmake by ArrowTesting

### LINKED ISSUE #35252: [C++] find_package(Arrow) may be failed without CMAKE_CXX_STANDARD project

`FindGTestAlt.cmake` uses `-DCMAKE_CXX_STANDARD=${CMAKE_CXX_STANDARD}` to check whether found GoogleTest can be used with Apache Arrow C++.

`find_package(Arrow)` also uses `FindGTestAlt.cmake`. Apache Arrow C++ always sets `CMAKE_CXX_STANDARD`. So `-DCMAKE_CXX_STANDARD=${CMAKE_CXX_STANDARD}` has no problem. But a project that uses `find_package(Arrow)` may not set `CMAKE_CXX_STANDARD`. In this case, `try_compile()` is failed with the following message:

```text
The CXX_STANDARD property on target "cmTC_ce38e" contained an invalid
  value: "".
```

* GoogleTest is used by `ArrowTesting` not `Arrow`. So we should put it to `ArrowTesting/`.
* `CMAKE_CXX_STANDARD` check is needed only when Apache Arrow C++ finds GoogleTest. It's not needed when `find_package(Arrow)` because suitable GoogleTest should be found by `find_package(GTestAlot)`. 

### Component(s)

C++

kou:
### Rationale for this change

`find_package(Arrow)` is failed when a project doesn't set `CMAKE_CXX_STANDARD`.

* `find_package(Arrow)` doesn't need to use `FindGTestAlt.cmake`. `find_package(ArrowTesting)` needs it.
* `FindGTestAlt.cmake` doesn't need to run availability check when it's called from `find_package(ArrowTesting)`. Because it's already done when Apache Arrow C++ is built.

### Are these changes tested?

Yes.

Yes.
* Closes: #35252


kou:
@github-actions crossbow submit -g nightly-packaging -g nightly-tests -g nightly-release

kou:
+1

PR #35257, opened 2023-04-20
GH-35256: [Go] Add ToMap to Metadata

### LINKED ISSUE #35256: [Go]: Add ToMap to Metadata

`Metadata` object right now has `MetadataFrom(kv map[string]string)` that receives map. It will be useful also to have the opposite.

### Component(s)

Go

yevgenypats:

* Closes: #35256
* Closes: #35256

### DIFF
--- a/go/arrow/schema_test.go
+++ b/go/arrow/schema_test.go
@@ -22,6 +22,7 @@ import (
 	"testing"
 
 	"github.com/apache/arrow/go/v12/arrow/endian"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestMetadata(t *testing.T) {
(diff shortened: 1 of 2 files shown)


go/arrow/schema_test.go
```diff
@@ -99,6 +100,9 @@ func TestMetadata(t *testing.T) {
 			if got, want := tc.md.String(), tc.serialize; got != want {
 				t.Fatalf("invalid stringer: got=%q, want=%q", got, want)
 			}
+			if len(tc.kvs) != 0 {
+				assert.Equal(t, tc.kvs, md.ToMap())
+			}
```
zeroshade, line 105: how about adding a round trip test:

create a map, construct the metadata from the map, then call `ToMap` on the result and assert the two maps are equal.

yevgenypats: I think we already have a round trip test here. We have `MetadataFrom` call on line 81

zeroshade: ah perfect

PR #35272, opened 2023-04-21
GH-35271: [R] `[.ArrowTabular` should return ChunkedArray when num_columns is 1

### LINKED ISSUE #35271: `[.ArrowTabular` should return ChunkedArray when num_columns is 1

For consistency of behavior with DataFrame, I think it is necessary to return a Chunked Array if there is only one column in the result.

```r
> all.equal(mtcars[, 1], mtcars$mpg)
[1] TRUE

> all.equal(arrow_table(mtcars)[, 1], arrow_table(mtcars)$mpg)
[1] FALSE
```

### Component(s)

R

eitsupi:
### Rationale for this change

For consistency of behavior with DataFrame, `Table[, "foo"]` is necessary to return a Chunked Array.

### Are these changes tested?

Yes.

Yes.

* Closes: #35271

### DIFF
--- a/r/R/arrow-tabular.R
+++ b/r/R/arrow-tabular.R
@@ -135,6 +135,9 @@ as.data.frame.ArrowTabular <- function(x, row.names = NULL, optional = FALSE, ..
   if (!missing(i)) {
     x <- filter_rows(x, i, ...)
   }
+  if (x$num_columns == 1L) {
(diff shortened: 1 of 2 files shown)


eitsupi:
Sorry, I didn't realize this was already controlled by the `drop` argument. Close.

PR #35284, opened 2023-04-22
GH-35244: [C#] DataType should compare values

### LINKED ISSUE #35244: [C#] DataType should compare values

Actually it uses the classic equal,

Maybe it would be easier and more handy to compare types based on their TypeId, name, and other characteristics

`interface IArrowType : IEquatable<ArrowType>`

### Component(s)

C#

Platob:
### Rationale for this change

Compare DataTypes easier

Hashcode + Equals on all Types

### Are these changes tested?

yes

Unless they compare types by reference, no impacts
* Closes: #35244
* Closes: #35244


westonpace (changes requested):
This would be a pretty major (and subtle) breaking change.  In addition, `ArrayDataTypeComparer.cs` already exists so I think this change is mostly for convenience.

Also, `ArrowType` is a reference type.  So I think it will be natural for C# developers to expect it to behave as a reference type.

Maybe a better approach would be to enhance `ArrayDataTypeComparer.cs` to be an instance of [`IEqualityComparer<IArrowType>`](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.iequalitycomparer-1?redirectedfrom=MSDN&view=net-7.0)?  This would allow you to use the custom equality comparer for things like unit tests and hash tables but avoid imposing a breaking or unintuitive change?

Platob:
Indeed yes, better close this pr for a new dev from scratch

PR #35294, opened 2023-04-23
[DO NOT MERGE] test whether features are enabled in r-arrow conda packages

nealrichardson:
Companion to https://github.com/apache/arrow/pull/35147. The new test should fail here but pass on that PR.


nealrichardson:
@github-actions crossbow submit conda-linux-*-r4* conda-osx-*-r4*

nealrichardson:
This is not right and no longer useful anyway. See the PR linked above.

PR #35298, opened 2023-04-24
GH-35297: [C++][IPC] Fix schema deserialization of map field

### LINKED ISSUE #35297: [C++][IPC] Schema deserialization does not preserve metadata of map field

The code below does not deal with field metadata like `nullable` and `metadata` of key and item sub-fields of map type.

```cpp
Status ConcreteTypeFromFlatbuffer(flatbuf::Type type, const void* type_data,
                                  FieldVector children, std::shared_ptr<DataType>* out) {
  switch (type) {
    case flatbuf::Type::NONE:
      return Status::Invalid("Type metadata cannot be none");
    ...
    case flatbuf::Type::Map:
      if (children.size() != 1) {
        return Status::Invalid("Map must have exactly 1 child field");
      }
      if (children[0]->nullable() || children[0]->type()->id() != Type::STRUCT ||
          children[0]->type()->num_fields() != 2) {
        return Status::Invalid("Map's key-item pairs must be non-nullable structs");
      }
      if (children[0]->type()->field(0)->nullable()) {
        return Status::Invalid("Map's keys must be non-nullable");
      } else {
        auto map = static_cast<const flatbuf::Map*>(type_data);
        *out = std::make_shared<MapType>(children[0]->type()->field(0)->type(),
                                         children[0]->type()->field(1)->type(),
                                         map->keysSorted());
      }
      return Status::OK();
    ...
    default:
      return Status::Invalid("Unrecognized type: " + ToChars(static_cast<int>(type)));
  }
}
```

The issue can be easily reproduced by a roundtrip test of schema serialization with a non-null map item field.

### Component(s)

C++

wgtmac:
### Rationale for this change

The arrow schema deserialization from flatbuffer does not preserve nullable and metadata of sub-fields in the map type.

Fix map type deserialization by creating map type using field objects instead of type objects for sub-types.

### Are these changes tested?

Add several new test cases in the arrow/ipc/read_write_test.cc.

No.
* Closes: #35297

### DIFF
--- a/cpp/src/arrow/ipc/metadata_internal.cc
+++ b/cpp/src/arrow/ipc/metadata_internal.cc
@@ -367,9 +367,9 @@ Status ConcreteTypeFromFlatbuffer(flatbuf::Type type, const void* type_data,
         return Status::Invalid("Map's keys must be non-nullable");
       } else {
         auto map = static_cast<const flatbuf::Map*>(type_data);
-        *out = std::make_shared<MapType>(children[0]->type()->field(0)->type(),
-                                         children[0]->type()->field(1)->type(),
-                                         map->keysSorted());
+        *out =
+            std::make_shared<MapType>(children[0]->type()->field(0),
+                                      children[0]->type()->field(1), map->keysSorted());
       }
       return Status::OK();
     case flatbuf::Type::FixedSizeList:
--- a/cpp/src/arrow/ipc/metadata_internal.cc
+++ b/cpp/src/arrow/ipc/metadata_internal.cc
@@ -367,9 +367,9 @@ Status ConcreteTypeFromFlatbuffer(flatbuf::Type type, const void* type_data,
         return Status::Invalid("Map's keys must be non-nullable");
       } else {
         auto map = static_cast<const flatbuf::Map*>(type_data);
-        *out =
-            std::make_shared<MapType>(children[0]->type()->field(0),
-                                      children[0]->type()->field(1), map->keysSorted());
+        *out = std::make_shared<MapType>(children[0]->type()->field(0)->WithName("key"),
+                                         children[0]->type()->field(1)->WithName("value"),
+                                         map->keysSorted());
       }
       return Status::OK();
     case flatbuf::Type::FixedSizeList:
--- a/cpp/src/arrow/ipc/read_write_test.cc
+++ b/cpp/src/arrow/ipc/read_write_test.cc
@@ -283,7 +283,7 @@ TEST_F(TestSchemaMetadata, NestedFields) {
 // Verify that nullable=false is well-preserved for child fields of map type.
 TEST_F(TestSchemaMetadata, MapField) {
   auto key = field("key", int32(), false);
-  auto item = field("item", int32(), false);
+  auto item = field("value", int32(), false);
   auto f0 = field("f0", std::make_shared<MapType>(key, item));
   Schema schema({f0});
   CheckSchemaRoundtrip(schema);
@@ -291,14 +291,14 @@ TEST_F(TestSchemaMetadata, MapField) {
 
 // Verify that key value metadata is well-preserved for child fields of nested type.
 TEST_F(TestSchemaMetadata, NestedFieldsWithKeyValueMetadata) {
-  auto inner = field("inner", std::make_shared<Int64Type>(), false,
-                     key_value_metadata({"foo"}, {"bar"}));
-
-  auto f0 = field("f0", list(inner), false, key_value_metadata({"k1"}, {"v1"}));
-  auto f1 = field("f1", struct_({inner}), false, key_value_metadata({"k2"}, {"v2"}));
-  auto f2 = field("f3", std::make_shared<MapType>(inner, inner), false,
-                  key_value_metadata({"k3"}, {"v3"}));
-
+  auto metadata = key_value_metadata({"foo"}, {"bar"});
+  auto inner_field = field("inner", int32(), false, metadata);
+  auto f0 = field("f0", list(inner_field), false, metadata);
+  auto f1 = field("f1", struct_({inner_field}), false, metadata);
+  auto f2 = field("f2",
+                  std::make_shared<MapType>(field("key", int32(), false, metadata),
+                                            field("value", int32(), false, metadata)),
+                  false, metadata);
   Schema schema({f0, f1, f2});
   CheckSchemaRoundtrip(schema);
 }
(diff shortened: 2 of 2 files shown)


wgtmac:
Could you please take a look? @lidavidm @wjones127

lidavidm:
Hmm, what was the behavior before?

entries/keys can never be nullable: https://github.com/apache/arrow/blob/cd14e2019ce93e1d6ffadc2556cc4275eeadd9f2/format/Schema.fbs#L124

so if it was allowing it before, we may have a compatibility issue on our hands...

wgtmac:
Yes, "entries" field nor the "key" field cannot be nullable. But the "value" field can be nullable and the bug always sets nullable to true to the "value" field after deserialization.

cpp/src/arrow/ipc/read_write_test.cc
```diff
@@ -271,6 +280,29 @@ TEST_F(TestSchemaMetadata, NestedFields) {
   CheckSchemaRoundtrip(schema);
 }
 
+// Verify that nullable=false is well-preserved for child fields of map type.
+TEST_F(TestSchemaMetadata, MapField) {
+  auto key = field("key", int32(), false);
+  auto item = field("item", int32(), false);
```
wgtmac, line 286: You may reproduce the bug with this test case without the fix. @lidavidm

wgtmac:
> 2. Preserve nullability of map keys field. I'm +0.1 on this. It's not supposed to be nullable. However I suppose it's possible for the child to have a null bit map, so maybe it's better to take care of that in the `ValidateFull` method?

I suppose this is always non-nullable for map keys field, otherwise it indicates a bug on the map type before serialization.

> 3. Preserve metadata of key and value fields. I'm +0.1. Seems fine but I'm not sure about the use case.

This is useful in my use case where each field is equipped with some extra information in the IPC file. 

> 4. Preserve field names of Map subfields. I'm -0.7 on this. The spec dictates they should be certain names, and I don't see good reason to let them deviate from this much. Right now we **don't** preserve them in the C Data Interface.

Good catch! I missed that. My question: is there any chance to prevent creating invalid sub-field names of map type?

lidavidm:
> > 3\. Preserve nullability of map keys field. I'm +0.1 on this. It's not supposed to be nullable. However I suppose it's possible for the child to have a null bit map, so maybe it's better to take care of that in the `ValidateFull` method?
> 
> I suppose this is always non-nullable for map keys field, otherwise it indicates a bug on the map type before serialization.

It's validated: 

https://github.com/apache/arrow/blob/5de56928e0fe43f02005552eee058de57ffb2682/cpp/src/arrow/array/array_nested.cc#L404-L406

That said, I don't actually see a check on the field's nullability - we should fix that.

> Preserve metadata of key and value fields. I'm +0.1. Seems fine but I'm not sure about the use case.

We should just do this (as Gang has done) for consistency. No reason to special case it just because there isn't an obvious use case.

> Good catch! I missed that. My question: is there any chance to prevent creating invalid sub-field names of map type?

It's not validated:

https://github.com/apache/arrow/blob/5de56928e0fe43f02005552eee058de57ffb2682/cpp/src/arrow/type.cc#L532-L537

This is also a potential compatibility issue. Off the top of my head, I might prefer: (1) ensure Validate checks for these cases, and (2) starting a bucket of things to enforce for a future MetadataVersion bump. (A hard error may invalidate files that used to be readable, and silently fixing the names may cause issues with things expecting round trips.) Or we could treat this as a C++ implementation problem, and (1) start issuing a warning + make sure Validate checks this, (2) turn this into a hard error with a C++/Python-specific flag to allow reading invalid files.

wgtmac:
Nullability of key field is only validated in the `MapType::Make` but not the constructors and it should be fixed.

https://github.com/apache/arrow/blob/main/cpp/src/arrow/type.cc#L544-L559

wgtmac:
I only see the description of map field names in the C Data Interface documentation: https://arrow.apache.org/docs/format/CDataInterface.html
```
As specified in the Arrow columnar format, the map type has a single child type named entries, itself a 2-child struct type of (key, value).
```

Does it only apply to the C Data Interface or it is included in the specs somewhere that I have missed? If field names are part of the specs, I think we can silently fix the field names upon map type creation and write a warn log.

lidavidm:
The canonical reference is the Flatbuffers:

https://github.com/apache/arrow/blob/9009dd76e4e9a2f4f13340ebf4173e71813b359b/format/Schema.fbs#L104-L128

So the names are not actually required, only the (non)nullability.

lidavidm:
> Nullability of key field is only validated in the `MapType::Make` but not the constructors and it should be fixed.

Unfortunately the constructors don't have a way to report this since we don't use exceptions :/

lidavidm:
However:

I think we should definitely preserve metadata for all fields and nullability for the value field. Let's fix that and kick off a discussion for the rest?

wgtmac:
> However:
> 
> I think we should definitely preserve metadata for all fields and nullability for the value field. Let's fix that and kick off a discussion for the rest?

Agreed. I will update the PR and leave the field name as is.

cpp/src/arrow/ipc/metadata_internal.cc
```diff
@@ -367,8 +367,8 @@ Status ConcreteTypeFromFlatbuffer(flatbuf::Type type, const void* type_data,
         return Status::Invalid("Map's keys must be non-nullable");
       } else {
         auto map = static_cast<const flatbuf::Map*>(type_data);
-        *out = std::make_shared<MapType>(children[0]->type()->field(0)->type(),
-                                         children[0]->type()->field(1)->type(),
+        *out = std::make_shared<MapType>(children[0]->type()->field(0)->WithName("key"),
```
wgtmac, line 370: How about this? Nullability of key field has been checked above. So here I just enforce key/value field name with other attributes unchanged. @lidavidm @wjones127

lidavidm: I think that's OK, since that's effectively what it was doing before when we just took the type. And now we preserve metadata.

lidavidm: However we should also probably force the nullability of key to `false` here to further match existing behavior.

wgtmac: Line 366 has checked this already. I don't think we need to explicitly set it to false here.

lidavidm: Ah, d'oh. thanks

lidavidm (approved):
LGTM!

wjones127 (approved):

PR #35299, opened 2023-04-24
GH-35285: [C#] Need refacto IArrowArrayBuilder for nested types

### LINKED ISSUE #35285: [C#] Need refacto IArrowArrayBuilder for nested types

Hello,

The interfaces IArrowArrayBuilder<out TArray, out TBuilder> forces the dev to return TBuilder

It makes harder the devs on nested structures

I don't see how it is necessary since its as the same behavior as void, its just cleaner for one line code if we want do it

Is it possible to move value append methods in IArrowArrayBuilder<out TArray> ?

```c#
public interface IArrowArrayBuilder
    {
        int Length { get; }
    }

    public interface IArrowArrayBuilder<out TArray> : IArrowArrayBuilder
        where TArray : IArrowArray
    {
        TArray Build(MemoryAllocator allocator);
    }

    public interface IArrowArrayBuilder<out TArray, out TBuilder> : IArrowArrayBuilder<TArray>
        where TArray : IArrowArray
        where TBuilder : IArrowArrayBuilder<TArray>
    {
        TBuilder Reserve(int capacity);
        TBuilder Resize(int length);
        TBuilder Clear();
    }


    public interface IArrowArrayBuilder<T, out TArray, out TBuilder> : IArrowArrayBuilder<TArray, TBuilder>
        where TArray : IArrowArray
        where TBuilder : IArrowArrayBuilder<TArray>
    {
        TBuilder Append(T value);
        TBuilder Append(ReadOnlySpan<T> span);
        TBuilder AppendRange(IEnumerable<T> values);
        TBuilder AppendNull();
        TBuilder Swap(int i, int j);
        TBuilder Set(int index, T value);
    }
```

### Component(s)

C#

Platob:
Make more native Builders with raw bit / byte access memory

Buffer
Data
Array
* Closes: #35285


csharp/src/Apache.Arrow/Builder/BufferBuilder.cs
```diff
@@ -0,0 +1,356 @@
...
+                }
+                Length = 0;
+            }
+        }
+
+        private const int DefaultCapacity = 64;
+        public int ByteLength { get; private set; }
+
+        public Memory<byte> Memory { get; private set; }
+        public BitBuffer BitOverhead { get; }
```
westonpace, line 52: This approach of using a BitBuffer for BitOverhead is inefficient because it is constantly allocating small byte spans.  I created a quick benchmark comparing this with a simpler approach for demonstration though its possible I'm missing something or did something incorrect:

https://gist.github.com/westonpace/35b619c4c9e90824d93ecb8306a9e169

```
|                Method | NumBits |          Mean |        Error |     StdDev |
|---------------------- |-------- |--------------:|-------------:|-----------:|
|           AppendFalse |       1 |      17.82 ns |     0.101 ns |   0.095 ns |
|            AppendTrue |       1 |      19.65 ns |     0.317 ns |   0.296 ns |
| AppendFalseNoOverhead |       1 |      10.42 ns |     0.072 ns |   0.067 ns |
|  AppendTrueNoOverhead |       1 |      10.30 ns |     0.100 ns |   0.089 ns |
|           AppendFalse |     100 |     257.41 ns |     1.035 ns |   0.968 ns |
|            AppendTrue |     100 |     293.54 ns |     1.179 ns |   1.102 ns |
| AppendFalseNoOverhead |     100 |      36.18 ns |     0.284 ns |   0.266 ns |
|  AppendTrueNoOverhead |     100 |      34.80 ns |     0.094 ns |   0.088 ns |
|           AppendFalse |   10000 |  22,638.86 ns |    77.641 ns |  68.827 ns |
|            AppendTrue |   10000 |  25,443.68 ns |    65.197 ns |  60.985 ns |
| AppendFalseNoOverhead |   10000 |      49.20 ns |     0.159 ns |   0.141 ns |
|  AppendTrueNoOverhead |   10000 |      49.24 ns |     0.180 ns |   0.168 ns |
|           AppendFalse |  100000 | 235,708.02 ns |   633.895 ns | 592.946 ns |
|            AppendTrue |  100000 | 271,133.50 ns | 1,123.712 ns | 996.141 ns |
| AppendFalseNoOverhead |  100000 |     346.36 ns |     0.456 ns |   0.380 ns |
|  AppendTrueNoOverhead |  100000 |     351.31 ns |     1.170 ns |   1.094 ns |
```

csharp/src/Apache.Arrow/Builder/BufferBuilder.cs
```diff
@@ -0,0 +1,356 @@
...
+
+            for (int i = 0; i < count; i++)
+                span[i] = value;
+
+            return AppendStructs<T>(span);
+        }
+
+        internal IBufferBuilder ReserveBytes(int numBytes)
+        {
+            EnsureAdditionalBytes(numBytes);
```
westonpace, line 243: Coming from the C++ implementation I would be surprised to learn that `ReserveBytes` reserved _additional_ bytes and didn't simply ensure bytes.  Maybe name this method `ReserveAdditionalBytes`?

csharp/src/Apache.Arrow/Field.Builder.cs
```diff
@@ -25,7 +28,7 @@ public class Builder
         {
             private Dictionary<string, string> _metadata;
             private string _name;
-            private IArrowType _type;
+            internal IArrowType _type;
```
westonpace, line 31: What's the motivation for this change?

csharp/src/Apache.Arrow/Types/Decimal256Type.cs
```diff
@@ -17,6 +17,9 @@ namespace Apache.Arrow.Types
 {
     public sealed class Decimal256Type: FixedSizeBinaryType
     {
+        // max int decimal 79228162514264337593543950335M : precision = 29
+        // max scaled decimal 1.2345678901234567890123456789M : scale = 28
```
westonpace, line 21: Can we clean these up?

westonpace (changes requested):
There's a lot of good stuff here, thanks for these contributions.  I went ahead and took a first look, focusing mostly on the buffer builder.  I think we can change to an approach that doesn't use bit overhead which should be more efficient.

westonpace:
Also, it seems there is already ArrowBuffer.BitmapBuilder which is a buffer builder that is optimized for bits (and https://github.com/apache/arrow/pull/35342 adds an efficient `AppendRange`).  Perhaps you could use that and leave your buffer builder focused on non-bit types? Either way, it might be good to align on one approach to maintain.

CC @asmirnov82

csharp/src/Apache.Arrow/Builder/BufferBuilder.cs
Platob: During its lifetime, the builder will allocate only 1 time 8 bits in bit overhead

On Ever bit(s) append, it will check if the 8 bits are written (like the old bitmap builder checking if it is % 8)

If it is full (8 bits are written) i will write a new byte in the Memory Buffer and reset all values in the BitOverhead to false setting his length to 0 an dread to recieve other bits

```csharp
public class BenchmarkBits
{
    public static long ElapsedTicks(int repetition, Action action)
    {
        Stopwatch stopwatch = Stopwatch.StartNew();

        for (int i = 0; i < repetition; i++)
            action();

        stopwatch.Stop();

        return stopwatch.ElapsedTicks / repetition;
    }

    private readonly ITestOutputHelper output;

    private static readonly bool[] trueBits = Enumerable.Range(0, 100).Select(_ => true).ToArray();
    private static readonly bool[] falseBits = Enumerable.Range(0, 100).Select(_ => false).ToArray();

    public BenchmarkBits(ITestOutputHelper output)
    {
        this.output = output;
    }

    [Fact]
    public void Bench()
    {
        output.WriteLine($"Elapsed {ElapsedTicks(1000000, MakeOld)}");
        output.WriteLine($"Elapsed {ElapsedTicks(1000000, MakeNew)}");
    }

    private static void MakeNew()
    {
        var builder = new BufferBuilder(64);

        builder.AppendBit(true).AppendBit(true)
            .AppendBits(trueBits)
            .AppendBits(falseBits)
            .Build();
    }

    private static void MakeOld()
    {
        var builder = new ArrowBuffer.BitmapBuilder(64);

        builder.Append(true).Append(true)
            .AppendRange(trueBits)
            .AppendRange(falseBits)
            .Build();
    }
}
````

On 1 million iterations the new implementation is in mean 168ticks
On 1 million iterations the old implementation is in mean 312 ticks

Comment (Platob) [reply csharp/src/Apache.Arrow/Builder/BufferBuilder.cs:52]: Same for sequential append it looks better, thus it scales good with larger value range

Old was in mean 1183 Ticks vs 938 Ticks for the new

```csharp
[Fact]
public void Bench()
{
    output.WriteLine($"Elapsed {ElapsedTicks(100000, MakeOld)}");
    output.WriteLine($"Elapsed {ElapsedTicks(100000, MakeNew)}");
}

private static void MakeNew()
{
    var builder = new BufferBuilder(64);

    for (int i = 0; i < 1000; i++)
        builder.AppendBit(true);

    builder.Build();
}

private static void MakeOld()
{
    var builder = new ArrowBuffer.BitmapBuilder(64);

    for (int i = 0; i < 1000; i++)
        builder.Append(true);

    builder.Build();
}
```

Platob:
> Also, it seems there is already ArrowBuffer.BitmapBuilder which is a buffer builder that is optimized for bits (and #35342 adds an efficient `AppendRange`). Perhaps you could use that and leave your buffer builder focused on non-bit types? Either way, it might be good to align on one approach to maintain.
> 
> CC @asmirnov82

Indeed, but for nested structures it needs specify the builders to check how to merge correctly 2 arrow buffers knowing if they are bit based or byte based

the main reason would be merge 2 array data by just checking their datatypes since their structures should be the same

```csharp
public virtual IArrayBuilder AppendValues(ArrayData data)
{
    // TODO: Make better / recursive fields data type check
    if (data.DataType.TypeId != DataType.TypeId)
        throw new ArgumentException($"Cannot append data type {data.DataType} in builder with data type {DataType}");

    NullCount += data.NullCount;
    Length += data.Length;

    Reserve(data.Length);

    for (int i = 0; i < Buffers.Length; i++)
    {
        IValueBufferBuilder current = Buffers[i];
        ArrowBuffer other = data.Buffers[i];

        if (current.ValueBitSize % 8 == 0)
        {
            // Full byte encoded
            current.AppendBytes(other.Span);
        }
        else
        {
            // Safe copy Bytes and remaining bits
            int end = (data.Length * current.ValueBitSize) / 8;

            current.AppendBytes(other.Span.Slice(0, end));

            Span<bool> bits = BitUtility.BytesToBits(other.Span.Slice(end)).Slice(0, data.Length - end * 8);
            current.AppendBits(bits);
        }
    }

    if (Children != null && data.Children != null)
    {
        for (int i = 0; i < Children.Length; i++)
        {
            Children[i].AppendValues(data.Children[i]);
        }
    }

    if (Dictionary != null && data.Dictionary != null)
    {
        Dictionary.AppendValues(data.Dictionary);
    }

    return this;
}
```

Platob:
Need to close this pr, too many structural changes to handle large memory allocation (the buffer length are in int32, should use buffer which can handle uint64 length at least for 2GB string elements)

PR #35303, opened 2023-04-24
GH-35302: [Go] Improve unsupported type error message in pqarrow

### LINKED ISSUE #35302: [Go] Parquet library should include details about unsupported types in the error message

Currently when using the pqarrow library with an unsupported type, the error is:

```
not implemented yet
```

This makes it hard to know which type exactly is not supported. It would be nice if this error were more verbose and included the type, e.g.

```
support for INTERVAL_DAY_TIME not implemented yet
```

### Component(s)

Go

hermanschaaf:
* Closes: #35302

### DIFF
--- a/go/parquet/pqarrow/schema.go
+++ b/go/parquet/pqarrow/schema.go
@@ -388,7 +388,7 @@ func fieldToNode(name string, field arrow.Field, props *parquet.WriterProperties
 		}
 		return schema.MapOf(field.Name, keyNode, valueNode, repFromNullable(field.Nullable), -1)
 	default:
-		return nil, xerrors.New("not implemented yet")
+		return nil, xerrors.Errorf("support for %v not implemented yet", field.Type.ID())
 	}
 
 	return schema.NewPrimitiveNodeLogical(name, repType, logicalType, typ, length, fieldIDFromMeta(field.Metadata))
--- a/go/parquet/pqarrow/schema.go
+++ b/go/parquet/pqarrow/schema.go
@@ -388,7 +388,7 @@ func fieldToNode(name string, field arrow.Field, props *parquet.WriterProperties
 		}
 		return schema.MapOf(field.Name, keyNode, valueNode, repFromNullable(field.Nullable), -1)
 	default:
-		return nil, xerrors.Errorf("support for %v not implemented yet", field.Type.ID())
+		return nil, fmt.Errorf("%w: support for %s", arrow.ErrNotImplemented, field.Type.ID())
 	}
 
 	return schema.NewPrimitiveNodeLogical(name, repType, logicalType, typ, length, fieldIDFromMeta(field.Metadata))
--- a/go/parquet/pqarrow/schema_test.go
+++ b/go/parquet/pqarrow/schema_test.go
@@ -416,27 +416,26 @@ func TestListStructBackwardCompatible(t *testing.T) {
 // when support for these types is added.
 func TestUnsupportedTypes(t *testing.T) {
 	unsupportedTypes := []struct {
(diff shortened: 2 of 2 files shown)


go/parquet/pqarrow/schema.go
```diff
@@ -388,7 +388,7 @@ func fieldToNode(name string, field arrow.Field, props *parquet.WriterProperties
 		}
 		return schema.MapOf(field.Name, keyNode, valueNode, repFromNullable(field.Nullable), -1)
 	default:
-		return nil, xerrors.New("not implemented yet")
+		return nil, xerrors.Errorf("support for %v not implemented yet", field.Type.ID())
```
zeroshade, line 391: probably makes more sense to just do `field.Type` instead of `field.Type.ID()` which would be more generic (it would give "MAP" instead of "Map<string, int>" etc....

zeroshade (changes requested):
Can you add a unit test that checks for the error?

go/parquet/pqarrow/schema.go
hermanschaaf: The `switch` that this is the default for is on `field.Type.ID()`, which is why I thought it makes more sense to use `field.Type.ID()` in the error as well. WDYT?

hermanschaaf:
@zeroshade Sure, though do you have any ideas on how to implement such a test? Would it be okay to have a test that checks for types that are currently unsupported, knowing we'd need to update the test later when support for those types are added?

zeroshade:
@hermanschaaf yup exactly. It's perfectly fine to check for types that are currently unsupported even though we'd have to update it later when support is added. In the meantime the test will ensure we don't break having that information in the error message.

hermanschaaf:
@zeroshade Added a test, please have another look :bow: I still think it's fine/better to use the ID in the message there because that's what the switch statement is based on, but let me know what you think.

go/parquet/pqarrow/schema.go
zeroshade: okay, that makes sense. I tend to like being more specific but I agree that since that's what we're switching on, it's good to be clear rather than expecting a consumer to understand the difference.

zeroshade, line 391: let's use `fmt.Errorf("%w: support for %s", arrow.ErrNotImplemented, field.Type.ID())`

zeroshade (changes requested):

go/parquet/pqarrow/schema.go
hermanschaaf: Updated 👍

go/parquet/pqarrow/schema_test.go
```diff
@@ -410,3 +411,33 @@ func TestListStructBackwardCompatible(t *testing.T) {
...
+		t.Run(tc.typ.ID().String(), func(t *testing.T) {
+			arrowFields := make([]arrow.Field, 0)
+			arrowFields = append(arrowFields, arrow.Field{Name: "unsupported", Type: tc.typ, Nullable: true})
+			arrowSchema := arrow.NewSchema(arrowFields, nil)
+			_, err := pqarrow.ToParquet(arrowSchema, nil, pqarrow.NewArrowWriterProperties())
+			assert.Error(t, err)
+			if err != nil {
+				wantMsg := fmt.Sprintf("support for %s not implemented yet", tc.typ.ID())
+				assert.Equal(t, wantMsg, err.Error())
+			}
```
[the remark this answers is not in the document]
hermanschaaf: Updated, thanks 👍

zeroshade (approved):

PR #35307, opened 2023-04-24
GH-35306: [Go] Fix Schema.Fields() to return copy of fields

### LINKED ISSUE #35306: [Go]: Schema Fields returns reference instead of copy

All methods on `Schema` returns objects that shouldn't mutate the schema itself apart from `Fields()`.

### Component(s)

Go

yevgenypats:
* Closes: #35306

### DIFF
--- a/go/arrow/schema.go
+++ b/go/arrow/schema.go
@@ -194,7 +194,11 @@ func (sc *Schema) WithEndianness(e endian.Endianness) *Schema {
 func (sc *Schema) Endianness() endian.Endianness { return sc.endianness }
 func (sc *Schema) IsNativeEndian() bool          { return sc.endianness == endian.NativeEndian }
 func (sc *Schema) Metadata() Metadata            { return sc.meta }
-func (sc *Schema) Fields() []Field               { return sc.fields }
+func (sc *Schema) Fields() []Field { 
+	fields := make([]Field, len(sc.fields))
+	copy(fields, sc.fields)
+	return fields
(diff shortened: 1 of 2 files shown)


go/arrow/schema.go
```diff
@@ -194,7 +194,11 @@ func (sc *Schema) WithEndianness(e endian.Endianness) *Schema {
 func (sc *Schema) Endianness() endian.Endianness { return sc.endianness }
 func (sc *Schema) IsNativeEndian() bool          { return sc.endianness == endian.NativeEndian }
 func (sc *Schema) Metadata() Metadata            { return sc.meta }
-func (sc *Schema) Fields() []Field               { return sc.fields }
+func (sc *Schema) Fields() []Field { 
+	fields := make([]Field, len(sc.fields))
+	copy(fields, sc.fields)
+	return fields
+}
```
zeroshade, line 201: I'm not sure I like this. From a performance standpoint, I'd prefer to return the slice as-is rather than a copy of the slice. Perhaps we should add a new function like `FieldsCopy()` or something to that end?

yevgenypats: We can do that but currently all methods return copy of things - like `Metadata` for example and `Field(i)` apart from this one (which also bit me in our code as I wasn't expecting the returned value to change the internal state)

zeroshade: @yevgenypats can you just add a test for this?

PR #35311, opened 2023-04-24
GH-35310: [Go] Incorrect value decimal128 from string

### LINKED ISSUE #35310: [Go] Incorrect value when creating decimal128 from string

Hi All,

I came across this behaviour when creating a decimal128 from a string:

```
func TestDecFromStringStrangeness(t *testing.T) {
	num, err := decimal128.FromString("2112.33", 38, 2)
	require.NoError(t, err)
	require.Equal(t, decimal128.New(0, 211233), num)
}
```
The value created seems to be out by one on the smallest digit:
```
=== RUN   TestDecFromStringBugMaybe
    decimal_test.go:54: 
        	Error Trace:	/Users/tfox/Development/tektite/types/decimal_test.go:54
        	Error:      	Not equal: 
        	            	expected: decimal128.Num{lo:0x33921, hi:0}
        	            	actual  : decimal128.Num{lo:0x33920, hi:0}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,3 +1,3 @@
        	            	 (decimal128.Num) {
        	            	- lo: (uint64) 211233,
        	            	+ lo: (uint64) 211232,
        	            	  hi: (int64) 0
        	Test:       	TestDecFromStringBugMaybe
--- FAIL: TestDecFromStringBugMaybe (0.00s)

```

### Component(s)

Go

zeroshade:
### Rationale for this change
Fix edge case of rounding when parsing a decimal128 value from string with a high precision.

Adding a simple 0.5 rounding based on the signbit before we truncate the value to an integer.

### Are these changes tested?
Yes, an edge case test case was added.

Some values that were previously incorrect will now be correct.

* Closes: #35310


lidavidm (commented):
Can we also add a test with a negative value?

zeroshade:
@lidavidm added same test but with negative version of the number

lidavidm (approved):

PR #35313, opened 2023-04-24
added None check in feather.write_feather

simone1999:
without this None check the call to _feather.write_feather does crash the python interpreter as it tries to access memory locations it is not allowed to access. This means it does NOT simply raise an exception, but completely crashes the entire Python process immediately. Can be reproduced by passing None as df to write_feather before this patch is activated.


kszucs:
Thanks @simone1999 for your PR! 

Could you please cover it with tests and create a corresponding issue as described in https://github.com/apache/arrow/pull/35313#issuecomment-1520955260 ?

PR #35322, opened 2023-04-25
GH-35316: [C++][FlightSQL] Use RowsToBatches() instead of ArrayFromJSON() in SQLite example server

### LINKED ISSUE #35316: [C++][FlightSQL] SQLite example server uses arrow::ArrayFromJSON

It's a function included in `libarrow_testing.so` not `libarrow.so`.
We should not depend on `libarrow_testing.so` because we don't require `ARROW_BUILD_TESTS=ON` for `cpp/examples/arrow/flight_sql_example.cc`.

### Component(s)

C++, FlightRPC

kou:
### Rationale for this change

`ArrayFromJSON()` requires `libarrow_testing.so`. But `cpp/examples/arrow/flight_sql_example.cc` uses the SQLite example server without `libarrow_testing.so` dependency.

Use `RowsToBatches()` in `libarrow.so` instead of `ArrayFromJSON()`.

### Are these changes tested?

Yes by hand.

Yes.
* Closes: #35316

### DIFF
--- a/cpp/src/arrow/flight/sql/example/sqlite_server.cc
+++ b/cpp/src/arrow/flight/sql/example/sqlite_server.cc
@@ -573,9 +573,10 @@ class SQLiteFlightSqlServer::Impl {
 
   arrow::Result<std::unique_ptr<FlightDataStream>> DoGetTypeInfo(
       const ServerCallContext& context, const GetXdbcTypeInfo& command) {
-    const std::shared_ptr<RecordBatch>& type_info_result =
-        command.data_type.has_value() ? DoGetTypeInfoResult(command.data_type.value())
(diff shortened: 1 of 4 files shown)


lidavidm (approved):
Thanks!

This is a bit complex though for what I would think is a common case. Perhaps we should consider some better utilities for this. (ADBC uses a column-wise, templated helper, though I suppose column-wise would also be messy here. https://github.com/apache/arrow-adbc/blob/a1dd00f88d76beaccfce000098d4cdb3944c015d/c/validation/adbc_validation_util.h#L304-L311)

kou:
> Perhaps we should consider some better utilities for this.

I agree with you.

PR #35340, opened 2023-04-26
MINOR: [C++] Add missing RunEndEncodedType dtor

felipecrv:
### Rationale for this change

Avoid problems with LTO compilation (see commit 24da3be9560c5)

Definition of a virtual dtor for RunEndEncodedType.

### Are these changes tested?

N/A.

### DIFF
--- a/cpp/src/arrow/type.cc
+++ b/cpp/src/arrow/type.cc
@@ -799,6 +799,8 @@ RunEndEncodedType::RunEndEncodedType(std::shared_ptr<DataType> run_end_type,
                std::make_shared<Field>("values", std::move(value_type), true)};
 }
 
+RunEndEncodedType::~RunEndEncodedType() = default;
+
 std::string RunEndEncodedType::ToString() const {
   std::stringstream s;
   s << name() << "<run_ends: " << run_end_type()->ToString()
--- a/cpp/src/arrow/type.h
+++ b/cpp/src/arrow/type.h
@@ -1232,6 +1232,7 @@ class ARROW_EXPORT RunEndEncodedType : public NestedType {
 
   explicit RunEndEncodedType(std::shared_ptr<DataType> run_end_type,
                              std::shared_ptr<DataType> value_type);
+  ~RunEndEncodedType() override;
 
   DataTypeLayout layout() const override {
     // A lot of existing code expects at least one buffer


mapleFU:
Just curious why need a explicit dtor here, seems that it can benifits LTO?

felipecrv:
> Just curious why need a explicit dtor here, seems that it can benifits LTO?

Since we don't mark these classes `final`, they can be extended by users and without a virtual dtor (that's what the `override` means -- virtual), users can get into issues if they destroy the instance via the base interface [1].

Making the dtor non-inline is good for binary size reduction and should have little perf impact as type destruction is not part of hot loops in Arrow. [2]

[1] https://stackoverflow.com/questions/461203/when-to-use-virtual-destructors
[2] https://chromium.googlesource.com/chromium/src/+/HEAD/docs/speed/binary_size/optimization_advice.md#optimizing-native-code

mapleFU:
1. without virtual dtor, the program still works. Because base class already declares a virtual dtor.
2. Maybe it will, I'm not familiar with that

From https://github.com/apache/arrow/commit/24da3be9560c5ee4eddbd62a815af553902504b8 . Seems previously it's for lto and dyn lib

mapleFU (approved):
After talk with C++ experts, I think `~RunEnEncodedType` may benifits dynamic linking libraries and avoid some problem. LGTM

felipecrv:
@zeroshade can I get this minor change merged? :)

pitrou:
I'm a bit surprised this would be needed at all. The destructor is already virtual in the base class and so it's inherited as such. Is there a use case that fails without this?

@bkietz @benibus What do you think?

benibus:
It suppose it could be the same reasoning used with the `NestedType` dtor (which this class inherits)? https://github.com/apache/arrow/blob/d7483ae4fa9acddf44740b102adaee6675a0423e/cpp/src/arrow/type.h#L344-L346

mapleFU:
Previous problem maybe occured when LTO enabled ( https://github.com/apache/arrow/commit/24da3be9560c5ee4eddbd62a815af553902504b8 )

I guess when two dynamic libraries have different compiling arguments (like one has `-frtti`, another one is `-fno-rtti`, or using different compilers), there would be some error occured here. And it might (lightly) benifits compiling time.

References:

* http://eel.is/c++draft/basic.scope.scope#4
* http://eel.is/c++draft/class.virtual#2

felipecrv:
@pitrou it's hypothetical but possible as I explained in https://github.com/apache/arrow/pull/35340#issuecomment-1523520894

And @mapleFU has brought up the point about the LTO build needing this somehow. Note that all classes in the file have done this and the class I added wasn't following the pattern.

pitrou:
> @pitrou it's hypothetical but possible as I explained in [#35340 (comment)](https://github.com/apache/arrow/pull/35340#issuecomment-1523520894)

I don't agree with this comment. The destructor is defined as `virtual` in the base class. You don't need to redefine it in the derived class.

> And @mapleFU has brought up the point about the LTO build needing this somehow. Note that all classes in the file have done this and the class I added wasn't following the pattern.

Fair enough. But can you change the PR description to match this?

felipecrv:
@pitrou alright. I simplified the description of the PR.

I messed up in my rationale and mixed two good practices that I have internalized and always followed blindly: (1) base class should have `virtual` dtor and derived class should have `~Derived() override` to guard against that base class changing and inadvertently becoming non-virtual [1] (not the case here as the base type class will never loose its virtual members)

[1] https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-override.html

Sorry to get all language-lawyerly. The real motivation behind the PR is that I saw a class that I had added not following the established practices in the source file. I try to match the existing Arrow style even I don't understand the rationale. There was the LTO issue that I didn't know about after all. Thank you @mapleFU for bringing it up.

pitrou (approved):
Thanks for the update !

PR #13810, opened 2022-08-06
GH-32605: [C#] Extend ArrowBuffer.BitmapBuilder to improve performance of array concatenation

asmirnov82:
Extend ArrowBuffer.BitmapBuilder with Append method overloaded with ReadOnlySpan<byte> parameter. 

This allows to add validity bits to the builder more efficiently (especially for cases when initial validity bits are added to newly created empty builder).  More over it makes BitmapBuilder API more consistent (for example ArrowBuffer.Builder<T> does have such method).

Currently adding new bits to existing bitmap is implemented in ArrayDataConcatenator, Code adds bit by bit in a cycle converting each to a boolean value:

for (int i = 0; i < length; i++)
{
    builder.Append(span.IsEmpty || BitUtility.GetBit(span, i));
}
* Closes: #32605

PR #35342, opened 2023-04-26
GH-32605: [C#] Extend validity buffer api

### LINKED ISSUE #32605: [C#] Extend ArrowBuffer.BitmapBuilder to improve performance of array concatenation
Extend ArrowBuffer.BitmapBuilder with Append method overloaded with ReadOnlySpan<byte> parameter. 

This allows to add validity bits to the builder more efficiently (especially for cases when initial validity bits are added to newly created empty builder).  More over it makes BitmapBuilder API more consistent (for example ArrowBuffer.Builder<T> does have such method).

Currently adding new bits to existing bitmap is implemented in ArrayDataConcatenator, Code adds bit by bit in a cycle converting each to a boolean value:

for (int i = 0; i < length; i++)
{
    builder.Append(span.IsEmpty || BitUtility.GetBit(span, i));
}

Initial problem was described in this email: https://lists.apache.org/thread/kls6tjq2hclsvd16tw901ooo5soojrmb

PR: https://github.com/apache/arrow/pull/13810

**Reporter**: [Alexey Smirnov](https://issues.apache.org/jira/browse/ARROW-17330)
#### PRs and other links:
- [GitHub Pull Request #13810](https://github.com/apache/arrow/pull/13810)

<sub>**Note**: *This issue was originally created as [ARROW-17330](https://issues.apache.org/jira/browse/ARROW-17330). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

asmirnov82:
Add a method to the ValidityBuffer that adds the same bool value length times without allocating an Enumerable.Repeat object

### Rationale for this change

See more details in the code review comments in https://github.com/apache/arrow/pull/13810

* Closes: #32605


csharp/src/Apache.Arrow/ArrowBuffer.BitmapBuilder.cs
```diff
@@ -138,6 +138,28 @@ public BitmapBuilder AppendRange(IEnumerable<bool> values)
...
+                for (int i = 0; i < length; i++)
+                {
+                    BitUtility.SetBit(Span, Length, value);
+                    Length++;
+                }
+                                
+                SetBitCount += value ? length : 0;
+
+                return this;
+            }
```
westonpace, line 161: If you really want to eke out performance you could use [Span.Fill](https://learn.microsoft.com/en-us/dotnet/api/system.span-1.fill?view=net-7.0) for the whole bytes in the middle (assuming `length > 8`).  E.g. something like this:

```
    public BitmapBuilder AppendRange2(bool value, int length)
    {
        // Use simpler method if there aren't many values
        if (length < 20)
        {
            EnsureAdditionalCapacity(length);

            for (int i = 0; i < length; i++)
            {
                BitUtility.SetBit(Span, Length, value);
                Length++;
            }

            SetBitCount += value ? length : 0;

            return this;
        }

        // Otherwise do the work to figure out how to copy whole bytes

        int spaceLeftInCurrentByte = (8 - (Length % 8)) % 8;
        // Bits we will be using to finish up the current byte
        int remainderBits = Math.Min(length, spaceLeftInCurrentByte);
        // Whole bytes that will be set
        int wholeBytes = (length - remainderBits) / 8;
        // Bits to set after the last whole byte
        int trailingBits = length - remainderBits - (wholeBytes * 8);
        int newBytes = (trailingBits > 0) ? wholeBytes + 1 : wholeBytes;

        EnsureAdditionalCapacity(length);

        var span = Memory.Span;
        for (int i = 0; i < remainderBits; i++)
        {
            BitUtility.SetBit(span, Length, value);
            Length++;
        }

        if (wholeBytes > 0)
        {
            var fill = (byte)(value ? 0xFF : 0x00);
            // Length is guaranteed to be a multiple of 8 here
            span.Slice(Length / 8, wholeBytes).Fill(fill);
        }
        Length += wholeBytes * 8;

        for (int i = 0; i < trailingBits; i++)
        {
            BitUtility.SetBit(span, Length, value);
            Length++;
        }

        SetBitCount += value ? length : 0;
        return this;
    }
```

The results are pretty significant once `length` gets large enough:

```
|------------------ |-------- |--------------:|-----------:|-----------:|
|  AppendRangeFalse |       1 |      5.546 ns |  0.0364 ns |  0.0323 ns |
|   AppendRangeTrue |       1 |      5.613 ns |  0.0484 ns |  0.0429 ns |
| AppendRange2False |       1 |      5.835 ns |  0.0751 ns |  0.0666 ns |
|  AppendRange2True |       1 |      5.782 ns |  0.0280 ns |  0.0248 ns |
|  AppendRangeFalse |     100 |    281.145 ns |  0.6840 ns |  0.6063 ns |
|   AppendRangeTrue |     100 |    278.547 ns |  0.9460 ns |  0.8849 ns |
| AppendRange2False |     100 |     17.507 ns |  0.0414 ns |  0.0346 ns |
|  AppendRange2True |     100 |     17.108 ns |  0.0698 ns |  0.0653 ns |
|  AppendRangeFalse |   10000 | 27,397.477 ns | 51.9248 ns | 46.0300 ns |
|   AppendRangeTrue |   10000 | 26,872.102 ns | 59.5216 ns | 49.7032 ns |
| AppendRange2False |   10000 |     14.870 ns |  0.1078 ns |  0.1009 ns |
|  AppendRange2True |   10000 |     19.864 ns |  0.0747 ns |  0.0698 ns |
```

westonpace (approved):
This looks good (and I appreciate the new test).  I've suggested a slight improvement if you're interested.  Otherwise, we can probably proceed with this.

csharp/src/Apache.Arrow/ArrowBuffer.BitmapBuilder.cs
asmirnov82: Thanks for you suggestion. I like the idea! Similar approach is used in BitUtility.CountBits method, so I added a complementary implementation of SetBits method to BitUtility and used it in BitmapBuilder. I also changed CountBits to follow project official coding style.

csharp/src/Apache.Arrow/BitUtility.cs
```diff
@@ -62,6 +62,64 @@ public static void SetBit(Span<byte> data, int index, bool value)
...
         }
 
+        /// <summary>
+        /// Set the number of bits in a span of bytes starting
+        /// at a specific index, and limiting to length.
+        /// </summary>
+        /// <param name="data">Span to set bits value.</param>
+        /// <param name="offset">Bit index to start counting from.</param>
+        /// <param name="length">Maximum of bits in the span to consider.</param>
+        public static void SetBits(Span<byte> data, int index, int length, bool value)
```
eerhardt, line 72: Since this is a public API, can we do some argument checking? For example, `index + length < data.Length` and both `index` and `length` are non-negative.

csharp/src/Apache.Arrow/ArrowBuffer.BitmapBuilder.cs
```diff
@@ -138,6 +138,24 @@ public BitmapBuilder AppendRange(IEnumerable<bool> values)
                 return this;
             }
 
+            /// <summary>
+            /// Append multiple bits.
+            /// </summary>
+            /// <param name="value">Value of bits to append.</param>
+            /// <param name="length">Number of times the value should be added.</param>
+            /// <returns>Returns the builder (for fluent-style composition).</returns>
+            public BitmapBuilder AppendRange(bool value, int length)
```
eerhardt, line 147: Should check that `length` is non-negative.

eerhardt (commented):
Just a couple comments on argument checking. Should be good to merge once those are resolved.

csharp/src/Apache.Arrow/BitUtility.cs
asmirnov82: You are right. However, after taking a closer look at BitUtility class I have some concern that BitUtility should be a public API. And here are reasons:
1) There aren't any argument checking in all other methods (however, there are debug.asserts - I think it's done due to perfomance reason)
2) This class is implemented with an assumption of being used with Bitmap only, when we know that the number of meaningful bits should be less than Max.Int (because each bit corresponds to the validity of value in Value.Span, wich length is limit by Int). In general case the amount of bit inside a Span<byte> can be Max,Int * 8 and correct methods for working with bits inside a span should have long value for bit index and for length.

I added required checks into SetBits and CountBits methods (as this two are very similar) and didn't change other methods. You may see the resulting code in PR, but actually it looks intermediate compromise between two solutions:
 
change all BitUtility methods to use longs for bit indices (with possible performance drawdown)

or

make BitUtility internal (breaking change, in case somebody uses BitUtility in their code)

Would you please make some inputs on that?

csharp/src/Apache.Arrow/ArrowBuffer.BitmapBuilder.cs
asmirnov82: Fixed

csharp/src/Apache.Arrow/BitUtility.cs
eerhardt: I think a decent compromise for this PR is to not add a new public API to BitUtility. Instead, make the new method `internal`. Then you don't need to add argument checking to this method.

Then log an issue proposing making the whole `BitUtility` class `internal`. We should make that decision separate of this PR.

csharp/src/Apache.Arrow/BitUtility.cs
```diff
@@ -62,73 +62,154 @@ public static void SetBit(Span<byte> data, int index, bool value)
...
         /// </summary>
-        /// <param name="data">Span to count bits</param>
-        /// <param name="offset">Bit offset to start counting from</param>
-        /// <returns>Count of set (one) bits</returns>
-        public static int CountBits(ReadOnlySpan<byte> data, int offset) =>
-            CountBits(data, offset, data.Length * 8 - offset);
+        /// <param name="data">Span to count bits.</param>
+        /// <param name="index">Bit index to start counting from.</param>
+        /// <returns>Count of set (one) bits.</returns>
+        public static int CountBits(ReadOnlySpan<byte> data, int index) =>
```
eerhardt, line 146: These changes are technically breaking changes as well. Let's not make them here. If we make the whole class internal, we can refactor this code in any way we want.

asmirnov82: Fixed. I reverted the change related to checking attributes and throwing exception. Other changes are related only to naming, so I left them.

csharp/src/Apache.Arrow/BitUtility.cs
asmirnov82: Fixed

csharp/src/Apache.Arrow/BitUtility.cs
eerhardt: > Other changes are related only to naming, so I left them.

They are still a source breaking change though. Someone could call this method as:

```C#
BitUtility.CountBits(data, offset: 1)
```

And now their code won't compile.

eerhardt: This probably isn't a major issue, especially if we are considering making the class internal. Just more of an "FYI, this is a breaking change."

csharp/src/Apache.Arrow/BitUtility.cs
```diff
@@ -62,73 +62,146 @@ public static void SetBit(Span<byte> data, int index, bool value)
...
+
+            long spanLengthInBits = (long)data.Length * 8;
+
+            if (index < 0 || index >= spanLengthInBits)
+                throw new ArgumentOutOfRangeException(nameof(index));
+
+            int endBitIndex = checked(index + length - 1);
+
+            if (length < 0 || endBitIndex >= spanLengthInBits)
+                throw new ArgumentOutOfRangeException(nameof(length));
```
eerhardt, line 85: Do we still need this argument checking now that the method is `internal`?

asmirnov82: Removed now

csharp/src/Apache.Arrow/BitUtility.cs
asmirnov82: Got it. Thanks

eerhardt (approved):
LGTM. Thanks for the contribution, @asmirnov82!

PR #35347, opened 2023-04-26
GH-35270: [C++] Use Buffer instead of raw buffer in hash join internals

### LINKED ISSUE #35270: Memory leak in SwissTable::grow_double()

There are two allocation in grow_double. If the first suceeded but the second failed, the function will return without Free the first buffer 
```
RETURN_NOT_OK(pool_->Allocate(block_size_total_after, &blocks_new));
RETURN_NOT_OK(pool_->Allocate(hashes_size_total_after, &hashes_new_8B));
```

https://github.com/apache/arrow/blob/main/cpp/src/arrow/compute/key_map.cc#L670

### Component(s)

C++

westonpace:
### Rationale for this change

The current code has two storage buffers in the key map which are allocated with MemoryPool::Allocate which does not use smart pointers.  This could have led to a potential memory leak in an OOM scenario where the first allocate fails and it also led to some convoluted code keeping track of the previously allocated size in order to properly call Free.

Furthermore, it seems that this key map could have been getting potentially copied in the swiss join code.  While that was probably not happening (since the copy happened before the key map was initialized) it is still an easy recipe for an accidental double-free later on as we maintain the class.

Those raw buffers are changed to std::shared_ptr<Buffer> to avoid these issues.

### Are these changes tested?

Somewhat, the existing unit tests should ensure we didn't cause a regression.  I didn't introduce a regression test to introduce this potential bug because it would be very difficult to do so.

No

* Closes: #35270

### DIFF
--- a/cpp/src/arrow/compute/key_map.h
+++ b/cpp/src/arrow/compute/key_map.h
@@ -80,9 +80,11 @@ class ARROW_EXPORT SwissTable {
 
   void num_inserted(uint32_t i) { num_inserted_ = i; }
 
-  uint8_t* blocks() const { return blocks_; }
+  uint8_t* blocks() const { return blocks_->mutable_data(); }
 
-  uint32_t* hashes() const { return hashes_; }
+  uint32_t* hashes() const {
+    return reinterpret_cast<uint32_t*>(hashes_->mutable_data());
+  }
 
   /// \brief Extract group id for a given slot in a given block.
   ///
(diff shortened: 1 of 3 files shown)


cpp/src/arrow/compute/key_map.h
```diff
@@ -226,12 +228,12 @@ class ARROW_EXPORT SwissTable {
   // ---------------------------------------------------
   // * Empty bucket has value 0x80. Non-empty bucket has highest bit set to 0.
   //
-  uint8_t* blocks_;
+  std::shared_ptr<Buffer> blocks_;
 
   // Array of hashes of values inserted into slots.
   // Undefined if the corresponding slot is empty.
   // There is 64B padding at the end.
```
mapleFU, line 235: Should here still mention padding here? Since Buffer already ensure padding?

mapleFU (approved):

cpp/src/arrow/compute/key_map.h
westonpace: Buffer ensures that allocations are aligned (and I think, at least 64 bytes) but I don't know if it actually guarantees padding.  I think this buffer needs to go 64 bytes past the end regardless of the size which is a slightly different requirement.

mapleFU: Got it, padding and aligned are different, I missed it previously. the patch LGTM

pitrou (approved):
+1, nice robustness improvement

Issue #34890, opened 2023-04-04
ArrowNotImplementedError: Function 'dictionary_encode' has no kernel matching input types (dictionary<values=string, indices=uint8, ordered=0>)


randolf-scholz:
### Describe the bug, including details regarding any error messages, version, and platform.

`dictionary_encode` should probably just return the data as-is if it is already encoded as a dictionary.

### Component(s)

Python

randolf-scholz:
MWE:

```python
import pyarrow as pa

dict_array = pa.array(["banana", "orange"]).dictionary_encode()
dict_array.dictionary_encode()  # ✔
pa.compute.dictionary_encode(dict_array)  # ✘
```

randolf-scholz:
Using chunked arrays, `array.dictionary_encode()` breaks as well:

```python
import pyarrow as pa

dict_array = pa.chunked_array([["banana", "orange"], ["orange", "apple"]]).dictionary_encode()
dict_array.dictionary_encode()  # ✘
pa.compute.dictionary_encode(dict_array)  # ✘
```

danepitkin:
Honestly, it might be better to keep the error and remove support for the python `DictionaryArray.dictionary_encode()` method. The errors being thrown all come from the same compute API in the native C++ implementation. The python method is literally just 
```
    def dictionary_encode(self):
        return self
```

randolf-scholz:
One case when this bug appears in the wild is when trying to pivot a `pandas.DataFrame` that has been loaded with the `pyarrow` backend:

```python
import pandas as pd

df = (
    pd.DataFrame([("A", 1), ("B", 2), ("C", 3)], columns=["var", "val"])
    .astype({"var": "string", "val": "float32"})
    .astype({"var": "category", "val": "float32"})
)

# write and reload as parquet with pyarrow backend
df.to_parquet("demo.parquet")
df = pd.read_parquet("demo.parquet", dtype_backend="pyarrow")
print(df.dtypes)  # var is now dictionary[int32,string]

df.pivot(columns=["var"], values=["val"])  # ✘ ArrowNotImplementedError
```

jorisvandenbossche:
But it is certainly true that for convenience (to avoid having to check if the input array is already dict encoded before calling `dictionary_encode`), it might be nice that `pyarrow.compute.dictionary_encode` would just be a no-op for dictionary encoded arrays.

ddutt:
I see this problem even for pandas frames with groupby followed by any operation such as nunique, count etc.

[...]

PR #35356, opened 2023-04-27
GH-34588:[C++][Python] Add a MetaFunction for "dictionary_decode"

### LINKED ISSUE #34588: [C++][Python] Add compute kernel for "dictionary_decode"
The pyarrow DictionaryArray class has a `dictionary_decode()` method, as a counterpart of the `dictionary_encode()`. However, this is not an actual kernel, but just implemented manually (it's also just a simple "take" of course):

https://github.com/apache/arrow/blob/fe88d9ad5c346786842913c8d2a369db099b5406/python/pyarrow/array.pxi#L2506-L2510

It would be nice to make an actual kernel for this, just like we have a "dictionary_encode" kernel (and eg also "run_end_encode" and "run_end_decode" kernels). Among other things, this makes it available as a function in pyarrow.compute (and for example ensures you can call this on a chunked array more easily).

R-JunmingChen:
**Rationale for this change**
This PR is for [Issue-34588](https://github.com/apache/arrow/issues/34588). Discussing with @westonpace, a MetaFunction for "dictionary_decode" is implemented instead of adding a compute kernel.

**What changes are included in this PR?**
C++: Meta Function of dictionary_decode.
Python: Test

**Are these changes tested?**
One test in tests/test_compute.py

* Closes: #34588


cpp/src/arrow/compute/dictionary_decode.cc
```diff
@@ -0,0 +1,95 @@
...
+}
+
+}  // namespace internal
+
+Result<Datum> DictionaryDecode(const Datum& value, ExecContext* ctx) {
+  return CallFunction("dictionary_decode", {value}, ctx);
+}
+
+Result<std::shared_ptr<Array>> DictionaryDecode(const Array& value, ExecContext* ctx) {
+  ARROW_ASSIGN_OR_RAISE(Datum result, DictionaryDecode(Datum(value), ctx));
```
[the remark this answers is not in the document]
R-JunmingChen: > Should it just return an error result or raise exception?
I think we should just return an error as [conventions](https://arrow.apache.org/docs/cpp/conventions.html) says. Exceptions are for the third-party codes.

R-JunmingChen:
Hi, @westonpace @AlenkaF, if you have spared time, please review this PR

AlenkaF (commented):
Thank you for the contribution @R-JunmingChen !

I had a look at the Python part. Just have some minor nits.
- Can we add a combo of encode & decode to the test?
- There are some corrections needed from doctest and linter checks:

```
--- original//arrow/python/pyarrow/compute.py
+++ fixed//arrow/python/pyarrow/compute.py
@@ -387,7 +387,7 @@
     array : decoded Array
         The dictionary_decode result as a new Array
     """
-    if(not isinstance(arr.type, pa.DictionaryType)):
+    if (not isinstance(arr.type, pa.DictionaryType)):
         raise TypeError("Must pass a dictionary array")
 
     return call_function("dictionary_decode", [arr], memory_pool)
--- original//arrow/python/pyarrow/tests/test_compute.py
+++ fixed//arrow/python/pyarrow/tests/test_compute.py
@@ -1750,7 +1750,7 @@
 
 def test_dictionary_decode():
     array = pa.array(["a", "a", "b", "c", "b"])
-    dictionary_array = pa.array(["a", "a", "b", "c", "b"], 
+    dictionary_array = pa.array(["a", "a", "b", "c", "b"],
                                 pa.dictionary(pa.int8(), pa.string()))
 
     assert array != dictionary_array
```

```
_________________ [doctest] pyarrow.compute.dictionary_decode __________________
340     arr : Array-like
341     memory_pool : MemoryPool, optional
342         memory pool to use for allocations during function execution.
343 
344     Examples
345     --------
346     >>> import pyarrow as pa
347     >>> import pyarrow.compute as pc
348     >>> x = pa.array(["a", "a", "b"], pa.dictionary(pa.int8(), pa.string()))
349     >>> x
Expected:
    <pyarrow.lib.DictionaryArray object at ...>
Got:
    <pyarrow.lib.DictionaryArray object at 0x7f6adc3da050>
    <BLANKLINE>
    -- dictionary:
      [
        "a",
        "b"
      ]
    -- indices:
      [
        0,
        0,
        1
      ]

/opt/conda/envs/arrow/lib/python3.9/site-packages/pyarrow/compute.py:349: DocTestFailure
__________________ [doctest] pyarrow.lib.default_memory_pool ___________________
125 default_memory_pool()
126 
127     Return the process-global memory pool.
128 
129     Examples
130     --------
131     >>> default_memory_pool()
Expected:
    <pyarrow.MemoryPool backend_name=... bytes_allocated=0 max_memory=...>
Got:
    <pyarrow.MemoryPool backend_name=jemalloc bytes_allocated=192 max_memory=1205120>
```

In the latter you can use `...` instead of specific bytes allocated etc.

R-JunmingChen:
Hi, @AlenkaF, I have fixed issues you mentioned. It's stange that there is still a lint error in R since I have no change on R code.

AlenkaF:
> Hi, @AlenkaF, I have fixed issues you mentioned. It's stange that there is still a lint error in R but I make no change on R code.

The linter error is actually coming from the C++ part where `dictionary_decode.cc` is missing an ending newline.
But I agree it is not very clear from the log. Here is the snippet:

```
FAILED: CMakeFiles/lint 
cd /tmp/arrow-lint-r6og459d/cpp-build && /usr/local/bin/python /arrow/cpp/build-support/run_cpplint.py --cpplint_binary /arrow/cpp/build-support/cpplint.py --exclude_globs /arrow/cpp/build-support/lint_exclusions.txt --source_dir /arrow/cpp/src --source_dir /arrow/cpp/examples --source_dir /arrow/cpp/tools --quiet
/arrow/cpp/src/arrow/compute/dictionary_decode.cc:95:  Could not find a newline character at the end of the file.  [whitespace/ending_newline] [5]
Done processing /arrow/cpp/src/arrow/compute/dictionary_decode.cc
Total errors found: 1
```

python/pyarrow/compute.py
```diff
@@ -333,6 +333,68 @@ def _make_global_functions():
...
+    <pyarrow.lib.StringArray object at ...>
+    [
+      "a",
+      "a",
+      "b"
+    ]
+
+    Returns
+    -------
+    array : decoded Array
```
AlenkaF, line 389:
```suggestion
    decoded : Array
```

python/pyarrow/compute.py
```diff
@@ -333,6 +333,68 @@ def _make_global_functions():
...
+    >>> x_decode
+    <pyarrow.lib.StringArray object at ...>
+    [
+      "a",
+      "a",
+      "b"
+    ]
+
+    The dictionary_decode actually call Cast in the deep implementation.
+    You can use Cast to decode DictionaryArray.
```
AlenkaF, line 376:
```suggestion
    ``dictionary_decode`` actually calls ``pyarrow.compute.cast`` in the
    deep implementation. You can also use `cast` function to decode a DictionaryArray.
```

AlenkaF (commented):
Just two more nitpicks in the docstring text. But the Python side LGTM +1.
I would wait for somebody reviewing the C++ side before approving though.

cpp/src/arrow/compute/dictionary_decode.cc
```diff
@@ -0,0 +1,95 @@
...
+namespace arrow {
+
+namespace compute {
+namespace internal {
+
+namespace {
+
+const FunctionDoc dictionary_decode_doc{"Decodes a DictionaryArray to an Array",
+                                        "The Function will call cast to really decode.",
+                                        {"dictionary_array"}};
```
westonpace, line 47:
```suggestion
const FunctionDoc dictionary_decode_doc{"Decodes a DictionaryArray to an Array",
                                        ("Return a plain-encoded version of the array input\n"
                                         "This function does nothing if the input is not a dictionary."),
                                        {"dictionary_array"}};
```
The fact that it calls `cast` is an implementation detail that the user doesn't really need to know.

cpp/src/arrow/compute/dictionary_decode.cc
```diff
@@ -0,0 +1,95 @@
...
+ public:
+  DictionaryDecodeMetaFunction()
+      : MetaFunction("dictionary_decode", Arity::Unary(), dictionary_decode_doc) {}
+
+  Result<Datum> ExecuteImpl(const std::vector<Datum>& args,
+                            const FunctionOptions* options,
+                            ExecContext* ctx) const override {
+    if (args[0].type() == nullptr || args[0].type()->id() != Type::DICTIONARY) {
+      return Status::TypeError("Expected a DictonaryArray");
+    }
```
westonpace, line 58: Instead of returning an error here it might be nice to just return the input itself.  The fact that `dictionary_encode` throws an error if the input is already a dictionary [has been seen as a bad thing](https://github.com/apache/arrow/issues/34890).

cpp/src/arrow/compute/dictionary_decode.cc
```diff
@@ -0,0 +1,95 @@
...
+
+    if (args[0].is_array()) {
+      ARROW_CHECK_NE(args[0].array()->dictionary, nullptr);
+      TypeHolder to_type(args[0].array()->dictionary->type);
+      CastOptions castOption = CastOptions::Safe(to_type);
+      return CallFunction("cast", args, &castOption, ctx);
+    } else if (args[0].is_chunked_array()) {
+      ARROW_CHECK_NE(args[0].chunked_array()->chunk(0), nullptr);
```
westonpace, line 66: This will fail for an empty chunked array (e.g. zero chunks)

cpp/src/arrow/compute/dictionary_decode.cc
```diff
@@ -0,0 +1,95 @@
...
+
+  Result<Datum> ExecuteImpl(const std::vector<Datum>& args,
+                            const FunctionOptions* options,
+                            ExecContext* ctx) const override {
+    if (args[0].type() == nullptr || args[0].type()->id() != Type::DICTIONARY) {
+      return Status::TypeError("Expected a DictonaryArray");
+    }
+
+    if (args[0].is_array()) {
+      ARROW_CHECK_NE(args[0].array()->dictionary, nullptr);
```
westonpace, line 61: Instead of grabbing the dictionary from the array/chunk it would probably be easier to grab it from the datum's type property.  Then you don't even need the if/else loop I think:

```
  DictionaryType* dict_type = checked_cast<DictionaryType>(args[0].type().get());
  CastOptions cast_options = CastOptions::Safe(dict_type->value_type());
```

cpp/src/arrow/compute/dictionary_decode.cc
```diff
@@ -0,0 +1,95 @@
...
+      TypeHolder to_type(args[0].array()->dictionary->type);
+      CastOptions castOption = CastOptions::Safe(to_type);
```
westonpace, line 63:
```suggestion
      CastOptions cast_option = CastOptions::Safe(to_type);
```

cpp/src/arrow/compute/dictionary_decode.cc
```diff
@@ -0,0 +1,95 @@
...
+      return CallFunction("cast", args, &castOption, ctx);
+    } else if (args[0].is_chunked_array()) {
+      ARROW_CHECK_NE(args[0].chunked_array()->chunk(0), nullptr);
+      ARROW_CHECK_NE(args[0].chunked_array()->chunk(0)->data(), nullptr);
+      ARROW_CHECK_NE(args[0].chunked_array()->chunk(0)->data()->dictionary, nullptr);
+      TypeHolder to_type(args[0].chunked_array()->chunk(0)->data()->dictionary->type);
+      CastOptions castOption = CastOptions::Safe(to_type);
```
westonpace, line 70:
```suggestion
      CastOptions cast_option = CastOptions::Safe(to_type);
```

cpp/src/arrow/compute/dictionary_decode.h
```diff
@@ -0,0 +1,61 @@
...
+class Array;
+
+namespace compute {
+
+class ExecContext;
+
+// ----------------------------------------------------------------------
+// Convenience invocation APIs for a number of kernels
+
+/// \brief decode a dictionary encoded array to normal array
```
westonpace, line 43:
```suggestion
/// \brief decode a dictionary encoded array to a normal array
```

cpp/src/arrow/compute/dictionary_decode.h
```diff
@@ -0,0 +1,61 @@
...
+/// \param[in] value dictionary array to decode
+/// \param[in] ctx the function execution context, optional
+/// \return the resulting array
+///
+ARROW_EXPORT
+Result<std::shared_ptr<Array>> DictionaryDecode(const Array& value,
+                                                ExecContext* ctx = NULLPTR);
+
+/// \brief decode a dictionary encoded array to normal array
```
westonpace, line 52:
```suggestion
/// \brief decode a dictionary encoded array to a normal array
```

westonpace (changes requested):
Thanks for the PR!  Sorry it took me a while to get to look at it.  Can you put the function in `vector_hash` next to `dictionary_encode`?  I agree it is a bit weird since hashing is not used but I think it would better to keep the functions together than to introduce a new file.

cpp/src/arrow/compute/dictionary_decode.cc
R-JunmingChen: User's opinion matters. Let's change.

R-JunmingChen:
> Can you put the function in `vector_hash` next to `dictionary_encode`? I agree it is a bit weird since hashing is not used but I think it would better to keep the functions together than to introduce a new file.

Sure, have moved. Do we need additional comment for that?

westonpace:
> Sure, have moved. Do we need additional comment for that?

I think it's ok, but if you want we can add a simple comment like

```
// This function does not use any hashing utilities but is kept in this file to be near dictionary_encode
```

python/pyarrow/compute.py
```diff
@@ -333,6 +333,68 @@ def _make_global_functions():
 _make_global_functions()
 
 
+def dictionary_decode(arr, memory_pool=None):
```
westonpace, line 336: Many of the compute functions are automatically described from the function registry (through some kind of python metaclass magic).  I don't really know how this works.  Maybe we still need to define this method since it is a meta function?  But I think it might work even if you don't define this method.  Can you try removing this method and see if the code still works?

CC @jorisvandenbossche or @bkietz who might know more details.

python/pyarrow/tests/test_compute.py
```diff
@@ -1755,6 +1755,20 @@ def test_logical():
...
+    dictionary_array = array.dictionary_encode()
+    dictionary_array_decode = dictionary_array.dictionary_decode()
+
+    assert array != dictionary_array
+
+    assert array == dictionary_array_decode
+    assert array == pc.dictionary_decode(dictionary_array)
+
+    with pytest.raises(TypeError):
+        pc.dictionary_decode(array)
```
westonpace, line 1769: Shouldn't this be ok now?

westonpace (changes requested):
Thanks.  I have a few more questions but I think this is getting close.

python/pyarrow/tests/test_compute.py
R-JunmingChen: Good Catch. I also check the input type in python code, let me remove the check

python/pyarrow/array.pxi
```diff
@@ -2542,8 +2542,14 @@ cdef class DictionaryArray(Array):
...
         Decodes the DictionaryArray to an Array.
+
+        See :func:`pyarrow.compute.dictionary_decode` for usage.
+
+        Returns
+        -------
+        dictionary_decode : Array
         """
-        return self.dictionary.take(self.indices)
+        return _pc().dictionary_decode(self)
```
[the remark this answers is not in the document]
R-JunmingChen: The old approach works well. I am not sure the risk of subtle difference. I think it's good to unify the dictionay_decode in Python.  But your consideration make senses, the change has a risk affecting user who use the old dictionary_decode.

python/pyarrow/compute.py
R-JunmingChen: Do you mean although we don't define this function, but we can call dictionary_decode in Python directly? I will try

R-JunmingChen: It indeed work without the definition

westonpace (approved):
Thanks for this addition!

python/pyarrow/array.pxi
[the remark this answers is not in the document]
westonpace: > I think it's good to unify the dictionay_decode logic in Python

This is a good point too.  I'll let @jorisvandenbossche or @AlenkaF weigh in as well.  We can move forward as-is for now and add this back if they want.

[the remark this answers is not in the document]
jorisvandenbossche: One other reason to keep the original is that it is _possible_ to build pyarrow without compute, and then the old version would still work while using the compute version would raise (although I am not sure how important this is, as many things won't work in that case)

PR #35362, opened 2023-04-28
GH-35361: [C++] Remove Perl dependency from cpp/build-support/run-test.sh

### LINKED ISSUE #35361: [C++] Remove Perl dependency from cpp/build-support/run-test.sh

Perl may not be so portable as before

See also:

https://lists.apache.org/thread/ys0w82slp3d7w3ngprqnqjlwmzf5brqc

> I have uploaded the log [1] for the run using conda with gandiva active. It
> looks like there is an issue with libcrypt.so causing these tests to
> segfault.
>
> 1: https://gist.github.com/assignUser/cba0a13875de9d6a4f31000f585244f0

https://lists.apache.org/thread/w47852gv4h9ngkjdnfhnmnqnx6km5nlb

> libcrypto.so isn't related on the segmentation fault. It's
> just for relating to showing backtrace.
>
> > perl: error while loading shared libraries: libcrypt.so.1:
> > cannot open shared object file: No such file or directory
>
> This is happen at
> https://github.com/apache/arrow/blob/main/cpp/build-support/run-test.sh#L42
> :
>
>     TEST_NAME=$(echo $TEST_FILENAME | perl -pe 's/\..+?$//') # Remove path and extension (if any).

### Component(s)

C++

kou:
### Rationale for this change

Because Perl in conda doesn't work on Manjaro Linux.

* Use `sed` instead of `perl`.
* Remove `stacktrace_addr2line.pl`. If we still need it, we can re-implement the addr2line filter in other language such as Python or shell later.

### Are these changes tested?

Yes by hand on Manjaro Linux.

No.
* Closes: #35361


kou:
@assignUser You can try this for `TEST_DEFAULT=0 TEST_SOURCE=0 TEST_CPP=1 USE_CONDA=1 dev/release/verify-release-candidate.sh 12.0.0 0` with the following patch:

```diff
diff --git a/dev/release/verify-release-candidate.sh b/dev/release/verify-release-candidate.sh
index 779e71a38..334348a8f 100755
--- a/dev/release/verify-release-candidate.sh
+++ b/dev/release/verify-release-candidate.sh
@@ -599,6 +599,8 @@ test_and_install_cpp() {
     ARROW_CMAKE_OPTIONS="${ARROW_CMAKE_OPTIONS:-} -G ${CMAKE_GENERATOR}"
   fi
 
+  cp ${SOURCE_DIR}/../../cpp/build-support/run-test.sh ${ARROW_SOURCE_DIR}/cpp/build-support/
+
   cmake \
     -DARROW_BOOST_USE_SHARED=ON \
     -DARROW_BUILD_EXAMPLES=OFF \
```

I confirmed that the verification command passes with this change on the `manjarolinux/base` Docker image.

assignUser:
Yes checking out this PR and applying the patch does work :tada: Thanks for taking this on!

kou:
That's good to know. :-)
Could you share it on the vote thread too?

kou (commented):
+1

PR #35364, opened 2023-04-28
GH-35363: [C++] Fix Substrait schema names and for segmented aggregation

### LINKED ISSUE #35363: [C++] Fix Substrait schema names and for segmented aggregation
### Describe the bug

Arrow Substrait does not pass `segment_keys`, as it should, when creating the schema. In addition, Arrow Substrait accepts a plan where the number of names does not match the size of the output schema for the plan. Both of these need to be fixed.

### Component(s)

C++

rtpsw:
See https://github.com/apache/arrow/issues/35363
* Closes: #35363

### DIFF
--- a/cpp/src/arrow/engine/substrait/options.cc
+++ b/cpp/src/arrow/engine/substrait/options.cc
@@ -205,12 +205,17 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
       ARROW_ASSIGN_OR_RAISE(auto aggregate, internal::ParseAggregateMeasure(
                                                 agg_measure, ext_set, conv_opts,
                                                 /*is_hash=*/!keys.empty(), input_schema));
+      aggregate.name = aggregate.function;
+      for (auto& field_ref : aggregate.target) {
+        ARROW_ASSIGN_OR_RAISE(auto field, field_ref.GetOne(*input_schema));
+        aggregate.name += "_" + field->name();
+      }
       aggregates.push_back(std::move(aggregate));
     }
 
-    ARROW_ASSIGN_OR_RAISE(auto aggregate_schema,
-                          acero::aggregate::MakeOutputSchema(
-                              input_schema, keys, /*segment_keys=*/{}, aggregates));
+    ARROW_ASSIGN_OR_RAISE(
+        auto aggregate_schema,
+        acero::aggregate::MakeOutputSchema(input_schema, keys, segment_keys, aggregates));
 
     return internal::MakeAggregateDeclaration(
         std::move(inputs[0].declaration), std::move(aggregate_schema),
--- a/cpp/src/arrow/engine/substrait/options.cc
+++ b/cpp/src/arrow/engine/substrait/options.cc
@@ -205,11 +205,6 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
       ARROW_ASSIGN_OR_RAISE(auto aggregate, internal::ParseAggregateMeasure(
                                                 agg_measure, ext_set, conv_opts,
                                                 /*is_hash=*/!keys.empty(), input_schema));
-      aggregate.name = aggregate.function;
-      for (auto& field_ref : aggregate.target) {
-        ARROW_ASSIGN_OR_RAISE(auto field, field_ref.GetOne(*input_schema));
-        aggregate.name += "_" + field->name();
-      }
       aggregates.push_back(std::move(aggregate));
     }
 
--- a/cpp/src/arrow/engine/substrait/options.cc
+++ b/cpp/src/arrow/engine/substrait/options.cc
@@ -208,9 +208,9 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
       aggregates.push_back(std::move(aggregate));
     }
 
-    ARROW_ASSIGN_OR_RAISE(
-        auto aggregate_schema,
-        acero::aggregate::MakeOutputSchema(input_schema, keys, segment_keys, aggregates));
+    ARROW_ASSIGN_OR_RAISE(auto aggregate_schema,
+                          acero::aggregate::MakeOutputSchema(
+                              input_schema, keys, segment_keys, aggregates));
 
     return internal::MakeAggregateDeclaration(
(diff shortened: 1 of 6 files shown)


rtpsw:
cc @westonpace, @icexelloss

cpp/src/arrow/engine/substrait/serde_test.cc
```diff
@@ -3937,7 +3937,7 @@ TEST(Substrait, ProjectWithMultiFieldExpressions) {
             }]
           }
         },
-        "names": ["A", "B", "C", "D"]
```
icexelloss, line 3940: Why this change?

icexelloss: This seems wrong. The plan has 4 output columns:
direct reference 0, 1, 2 and a new column with scalar Function

cpp/src/arrow/engine/substrait/options.cc
```diff
@@ -205,12 +205,17 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
       ARROW_ASSIGN_OR_RAISE(auto aggregate, internal::ParseAggregateMeasure(
                                                 agg_measure, ext_set, conv_opts,
                                                 /*is_hash=*/!keys.empty(), input_schema));
+      aggregate.name = aggregate.function;
```
icexelloss, line 208: Why do we need this logic?

cpp/src/arrow/engine/substrait/serde_test.cc
rtpsw: Without this change, [this check](https://github.com/apache/arrow/pull/35364/files#diff-11e93118405fd9709ff4ffb62aa7fbbc6ff8ca5bee7aa34f8a0ae951360d7a6eR158-R161) raises. IIUC, the plan has 3 outputs because its root is a projection with 3 emit indices.

cpp/src/arrow/engine/substrait/options.cc
rtpsw: Without this logic, the aggregate schema ends up with an empty name for each aggregate. This is because the [aggregate decoding](https://github.com/apache/arrow/blob/16dbd98e8f525884649e5a33096c74a455f64610/cpp/src/arrow/engine/substrait/extension_set.cc#L929) is [returning an aggregate with an empty name](https://github.com/apache/arrow/blob/16dbd98e8f525884649e5a33096c74a455f64610/cpp/src/arrow/engine/substrait/extension_set.cc#L980-L981), which in turn is likely because it has no access to the input schema against which to resolve the field reference (`*arg_ref`). The code here does have access to this schema and resolves the field reference.

cpp/src/arrow/engine/substrait/options.cc
```diff
@@ -205,12 +205,17 @@ class DefaultExtensionProvider : public BaseExtensionProvider {
...
+      for (auto& field_ref : aggregate.target) {
+        ARROW_ASSIGN_OR_RAISE(auto field, field_ref.GetOne(*input_schema));
+        aggregate.name += "_" + field->name();
```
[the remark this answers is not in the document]
rtpsw: Good point. I'll make this consistent shortly.

cpp/src/arrow/engine/substrait/options.cc
icexelloss: > Without this logic, the aggregate schema ends up with an empty name for each aggregate

While empty name is not great. I wonder if it actually breaks anything - substrait doesn't use names to refer to intermediate columns so IIUC the names here doesn't matter for correctness. (Although I agree meanful names are better for debug).

Does any code breaks without this change?

cpp/src/arrow/engine/substrait/serde_test.cc
icexelloss: I see

icexelloss: These names here (A, B, C) doesn't match the “output_schema" below (A, A1, A+1). Is this expected?

icexelloss: (Not that I expect your code change has anything to do with this, but looks quite confusing to me so I asked)

cpp/src/arrow/engine/substrait/options.cc
rtpsw: I think you're right and this is mostly useful in debugging. However, note that the schema produced by the pre-PR code is inconsistent - aggregate columns have empty names while other column have "normal" names (derived from the input). So, this change make sense at least for consistency.

> Does any code breaks without this change?

No. However, after adding this logic I needed to change one test case that checks the aggregate column name. There must be a reason for the existence of this test case, but I don't know it. @westonpace ?

cpp/src/arrow/engine/substrait/serde_test.cc
rtpsw: Right, not related to my changes herer. The invocations look like `CheckRoundTripResult` -> `AssertTablesEqualIgnoringOrder` -> `AssertTablesEqual`, and the latter (surprisingly, I guess) ignores column names for some reason.

cpp/src/arrow/engine/substrait/options.cc
icexelloss: @rtpsw I am happy to merge the code that fixes the the segmented aggregation bug (didn't pass the segmented key). However for the internal aggregation names change we would need validate that this naming is consistent with other aggregation codepath and/or refactor the naming logic to be shared (e.g., move that logic into ParseAggregateMeasure). I would suggest revert that change for now so we can move forward with the bug fix faster.

rtpsw: Done.

icexelloss:
@rtpsw LGTM. Can you fix the lint error?

rtpsw:
@westonpace, @icexelloss: is this good to go?

icexelloss:
Thanks @rtpsw - LGTM.

icexelloss (commented):
LGTM

PR #35368, opened 2023-04-29
GH-34722: [C++][Parquet] Minor: Update wording of Parquet NextPage

### LINKED ISSUE #34722: [C++][Parquet] Minor: Declare lifetime of `shared_ptr<Page>` on PageReader

In C++ Parquet, `SerializedPageReader` holds a `decryption_buffer_` and `decompression_buffer_`, and when `NextPage` called, it will reuse that buffer. So, `Page`'s  buffer data is bounded on `SerializedPageReader::NextPage`, when NextPage is called, it's data might be reset.

I have two problems here:
1. If it's by design, why don't we cache raw data in SerializePageReader?
2. Should we note these in comments?

### Component(s)

C++, Parquet

mapleFU:
### Rationale for this change

This change update wording of Parquet `NextPage`. It will reuse same decompression/decrypt buffer internal. So, use should attension it syntax.

No code change, just add some comments.

### Are these changes tested?

No need

No.

* Closes: #34722


mapleFU:
@wjones127 Mind take a look? Just some comments

cpp/src/parquet/column_reader.cc
```diff
@@ -275,6 +275,9 @@ class SerializedPageReader : public PageReader {
   }
 
   // Implement the PageReader interface
+  //
+  // SerializedPageReader reuse same decryption buffer and decompression buffer,
```
wgtmac, line 279:
```suggestion
  // SerializedPageReader reuses same decryption buffer and decompression buffer,
```

wgtmac: Should we make it clear when and which buffer will be reused? For example,
- If compression is enabled, decompression_buffer_ will be reused.
- If compression is not enabled but encryption is enabled, decryption_buffer_ will be reused.
- Neither encryption nor compression is enabled, the buffer is obtained directly from reading the stream.

cpp/src/parquet/column_reader.h
```diff
@@ -152,6 +152,10 @@ class PARQUET_EXPORT PageReader {
 
   // @returns: shared_ptr<Page>(nullptr) on EOS, std::shared_ptr<Page>
   // containing new Page otherwise
+  //
+  // In SerializedPageReader, it reuses same decryption buffer and
```
wgtmac, line 156: This is implementation detail, should we expose it to the doc string?

cpp/src/parquet/column_reader.cc
mapleFU: Seems that `decryption_buffer_` and `decompression_buffer_` is not conflict, and finally returned value might be one of them

cpp/src/parquet/column_reader.cc
```diff
@@ -275,6 +275,10 @@ class SerializedPageReader : public PageReader {
   }
 
   // Implement the PageReader interface
+  //
+  // SerializedPageReader reuses same decryption buffer and decompression buffer
+  // internal, so, if NextPage() is called, content of previous page might be
+  // invalidated.
```
wjones127, line 281:
```suggestion
  // The returned Page contains references that aren't guaranteed to live
  // beyond the next call to NextPage(). SerializedPageReader reuses the 
  // decryption and decompression buffers internally, so if NextPage() is 
  // called then the content of previous page might be invalidated.
```

cpp/src/parquet/column_reader.h
```diff
@@ -152,6 +152,10 @@ class PARQUET_EXPORT PageReader {
 
   // @returns: shared_ptr<Page>(nullptr) on EOS, std::shared_ptr<Page>
   // containing new Page otherwise
+  //
+  // In PageReader, it may reuse same underlying buffer, so, if
+  // NextPage() is called, content of previous page might be
+  // invalidated.
```
wjones127, line 158:
```suggestion
  // The returned Page may contain references that aren't guaranteed to live
  // beyond the next call to NextPage().
```

wjones127 (commented):
Perhaps we can just frame the API docs around the lifetimes of the returned objects (how long is it guaranteed to be valid?) and not how it is implemented?

wjones127 (commented):
Perhaps we can just frame the API docs around the lifetimes of the returned objects (how long is it guaranteed to be valid?) and not how it is implemented?

mapleFU:
Sure, changed it

wjones127 (approved):

PR #35376, opened 2023-05-01
GH-35375: [C++][FlightRPC] Add `arrow::flight::ServerCallContext::incoming_headers()`

### LINKED ISSUE #35375: [C++][FlightRPC] Add `arrow::flight::ServerCallContext::incoming_headers()`

It returns headers sent by a client.

We can get them only in `arrow::flight::ServerMiddlewareCactory::StartCall()` for now. But they're useful for in each RPC call.

### Component(s)

C++, FlightRPC

kou:
### Rationale for this change

It returns headers sent by a client.

We can get them only in `arrow::flight::ServerMiddlewareCactory::StartCall()` for now. But they're useful for in each RPC call.

Add the method.

### Are these changes tested?

Yes.

Yes.
* Closes: #35375

### DIFF
--- a/cpp/src/arrow/flight/test_util.cc
+++ b/cpp/src/arrow/flight/test_util.cc
@@ -477,7 +477,7 @@ class FlightTestServer : public FlightServerBase {
   Status ListIncomingHeaders(const ServerCallContext& context, const Action& action,
                              std::unique_ptr<ResultStream>* out) {
     std::vector<Result> results;
-    auto prefix = static_cast<std::string_view>(*(action.body));
+    std::string_view prefix(*action.body);
     for (const auto& header : context.incoming_headers()) {
       if (header.first.substr(0, prefix.size()) != prefix) {
         continue;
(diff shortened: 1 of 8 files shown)


cpp/src/arrow/flight/test_util.cc
```diff
@@ -474,12 +474,31 @@ class FlightTestServer : public FlightServerBase {
     return Status::OK();
   }
 
+  Status ListIncomingHeaders(const ServerCallContext& context, const Action& action,
+                             std::unique_ptr<ResultStream>* out) {
+    std::vector<Result> results;
+    auto prefix = static_cast<std::string_view>(*(action.body));
```
lidavidm, line 480: nit, but IMO static_cast is a bit of a confusing way to write an implicit conversion

lidavidm (approved):
Thanks!

cpp/src/arrow/flight/test_util.cc
kou: OK. How about `std::string_view prefix(*action.body)`?

lidavidm (approved):
Thanks!

PR #35378, opened 2023-05-01
GH-35377: [C++][FlightRPC] Add a `ServerCallContext` parameter to `arrow::flight::ServerAuthHandler` methods

### LINKED ISSUE #35377: [C++][FlightRPC] Add a `ServerCallContext` parameter to `arrow::flight::ServerAuthHandler` methods

Because they are also RPC calls to like others such as `ListFlights()` and `DoGet()`.

We should keep backward compatibility with this enhancement.

### Component(s)

C++, FlightRPC

kou:
### Rationale for this change

Because they are also RPC calls to like others such as `ListFlights()` and `DoGet()`.

Add with `ServerCallContext` versions.

Old signature versions still exist for keeping backward compatibility.

### Are these changes tested?

Yes.

Yes.

* C++ API: Don't include any breaking changes.
* C GLib API: Includes a breaking change. `context` argument is added.

* Closes: #35377

### DIFF
--- a/cpp/src/arrow/flight/server_auth.h
+++ b/cpp/src/arrow/flight/server_auth.h
@@ -28,6 +28,8 @@ namespace arrow {
 
 namespace flight {
 
+class ServerCallContext;
+
 /// \brief A reader for messages from the client during an
 /// authentication handshake.
 class ARROW_FLIGHT_EXPORT ServerAuthReader {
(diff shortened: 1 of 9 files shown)


cpp/src/arrow/flight/server_auth.h
```diff
@@ -55,8 +57,54 @@ class ARROW_FLIGHT_EXPORT ServerAuthHandler {
...
   /// can send and read responses from the client at any time.
+  ///
+  /// If this version is implemented, the deprecated Authentication()
+  /// without ServerCallContext version isn't used. So we can
+  /// implement the deprecated version like the following:
+  ///
+  ///   Status Authenticate(ServerAuthSender* outgoing,
+  ///                       ServerAuthReader* incoming) override {
+  ///     return Status::NotImplemented("This version is never used");
+  ///   }
```
lidavidm, line 68: We could change the deprecated overload from pure virtual to just virtual, and add a default no-op/always-failing implementation. That way people who only implement the new overload won't have to update code when we remove the deprecated overload. What do you think?

cpp/src/arrow/flight/server_auth.h
```diff
@@ -55,8 +57,54 @@ class ARROW_FLIGHT_EXPORT ServerAuthHandler {
...
+    return Authenticate(outgoing, incoming);
+  }
+  /// \brief Authenticate the client on initial connection. The server
+  /// can send and read responses from the client at any time.
+  /// \param[in] outgoing The writer for messages to the client.
+  /// \param[in] incoming The reader for messages from the client.
+  /// \return Status OK if this authentication is succeeded.
+  /// \deprecated Deprecated since 13.0.0. Implement the Authentication()
+  /// with ServerCallContext version instead.
   virtual Status Authenticate(ServerAuthSender* outgoing, ServerAuthReader* incoming) = 0;
```
lidavidm, line 85: I suppose we can't mark this as deprecated to the compiler without triggering a bunch of warnings on ourselves, can we...

lidavidm (commented):
LGTM.

cpp/src/arrow/flight/server_auth.h
kou: Oh, it's a good idea! I'll do it.

kou: Done.

cpp/src/arrow/flight/server_auth.h
kou: We can mark this as deprecated but client codes that overrides this don't receive any deprecated warning from their compiler. Because client codes don't call this method. Our server implementations call this method instead.

Anyway, I've added `ARROW_DEPRECATED()` because a `-Wdocumentation-deprecated-sync` warning is reported without it: https://github.com/kou/arrow/actions/runs/4849400904/jobs/8641350382#step:9:1437

```text
/Users/runner/work/arrow/arrow/cpp/src/arrow/flight/server_auth.h:83:8: error: declaration is marked with '\deprecated' command but does not have a deprecation attribute [-Werror,-Wdocumentation-deprecated-sync]
  /// \deprecated Deprecated since 13.0.0. Implement the Authentication()
      ~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/runner/work/arrow/arrow/cpp/src/arrow/flight/server_auth.h:85:3: note: add a deprecation attribute to the declaration to silence this warning
  virtual Status Authenticate(ServerAuthSender* outgoing, ServerAuthReader* incoming) = 0;
  ^
```

cpp/src/arrow/flight/server_auth.h
```diff
@@ -55,23 +57,68 @@ class ARROW_FLIGHT_EXPORT ServerAuthHandler {
...
+                                  "::Authenticate() isn't implemented");
+  }
+  /// \brief Validate a per-call client token.
+  /// \param[in] token The client token. May be the empty string if
+  /// the client does not provide a token.
+  /// \param[out] peer_identity The identity of the peer, if this
+  /// authentication method supports it.
+  /// \return Status OK if the token is valid, any other status if
+  /// validation failed
+  virtual Status IsValid(const ServerCallContext& context, const std::string& token,
```
lidavidm, line 91: Looks like doxygen complains unless there's a `\param` for this

kou: Good catch!
I forgot to add it...

lidavidm (approved):

PR #35384, opened 2023-05-01
GH-35383: [C++] Prefer max_concurrency over executor capacity to avoid segmentation fault

### LINKED ISSUE #35383: [CI] Some TPCH benchmarks started failing

When  https://github.com/apache/arrow/pull/34912 was merged, some TPCH benchmarks started failing. Here is a build with failed benchmarks: https://buildkite.com/apache-arrow/arrow-bci-benchmark-on-test-mac-arm/builds/2835

https://github.com/apache/arrow/pull/35364 fixed some of these failing TPCH benchmarks but these benchmarks are still failing:

Screen Shot 2023-05-01 at 11 29 41 AM

### Component(s)

Benchmarking

westonpace:
### Rationale for this change

The recent change (#34912) calculates the max concurrency using `plan->query_context()->executor()->GetCapacity()`.  This is later used to initialize the kernel states.  However, this is different than what we used to use.  The previous method used was `plan->query_context()->max_concurrency()` which is slightly different(if the aggregate node IS run in parallel then we initialize one state for each CPU thread, one for each I/O thread, and one for the calling user thread).

This is unfortunately a bit complicated as `max_concurrency` would not be a good indicator to use when determining if the plan is running in parallel or not.  So we need to query both properties and use them in their respective spots.

Now, `max_concurrency` is used to figure out how many thread local states need to be initialized and `GetCapacity` is used to figure out if there are multiple CPU threads or not.

### Are these changes tested?

The bug was caught by the benchmarks which is a bit concerning.  Most of the CI have a very small number of CPU threads and don't experience much concurrency and so I think we just didn't see this pattern. Or possibly, this pattern is only experienced in the legacy way that pyarrow launches exec plans.

No.
* Closes: #35383

### DIFF
--- a/cpp/src/arrow/acero/aggregate_node.cc
+++ b/cpp/src/arrow/acero/aggregate_node.cc
@@ -442,6 +442,7 @@ class ScalarAggregateNode : public ExecNode, public TracedNode {
                            aggs_[i].options ? aggs_[i].options->ToString() : "<NULLPTR>"},
                           {"function.kind", std::string(kind_name()) + "::Consume"}});
       KernelContext batch_ctx{plan()->query_context()->exec_context()};
+      DCHECK_LT(thread_index, states_[i].size());
       batch_ctx.SetState(states_[i][thread_index].get());
 
       std::vector<ExecValue> column_values;
--- a/cpp/src/arrow/acero/aggregate_node.cc
+++ b/cpp/src/arrow/acero/aggregate_node.cc
@@ -398,8 +398,7 @@ class ScalarAggregateNode : public ExecNode, public TracedNode {
     auto aggregates = aggregate_options.aggregates;
     const auto& keys = aggregate_options.keys;
     const auto& segment_keys = aggregate_options.segment_keys;
-    const auto concurreny =
-        plan->query_context()->exec_context()->executor()->GetCapacity();
+    const auto concurreny = plan->query_context()->max_concurrency();
 
     if (keys.size() > 0) {
       return Status::Invalid("Scalar aggregation with some key");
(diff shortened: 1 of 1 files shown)


westonpace:
@ursabot please benchmark lang=R

cpp/src/arrow/acero/aggregate_node.cc
```diff
@@ -398,25 +398,25 @@ class ScalarAggregateNode : public ExecNode, public TracedNode {
     auto aggregates = aggregate_options.aggregates;
     const auto& keys = aggregate_options.keys;
     const auto& segment_keys = aggregate_options.segment_keys;
-    const auto concurreny =
-        plan->query_context()->exec_context()->executor()->GetCapacity();
+    const auto concurrency = plan->query_context()->max_concurrency();
+    // We can't use concurrency == 1 because that include I/O concurrency
```
icexelloss, line 402: Do you have an issue with this code?
https://github.com/apache/arrow/blob/be12888997c81b1fb7947f6284be1256edd4d3e4/cpp/src/arrow/acero/hash_join_node.cc#L929

westonpace: I don't think so.  You'll notice, in the very next line, it is doing the same calculation that max_concurrency does.

westonpace:
The mac arm benchmarks are now passing.  It looks like the 9960x is offline so I can't actually tell.

lidavidm (approved):

icexelloss (approved):
LGTM

cpp/src/arrow/acero/aggregate_node.cc
icexelloss: I see. Thanks!

westonpace:
> Does this mean that:
> 
>     1. `plan->query_context()->executor()->GetCapacity()`: config cpu threads
> 
>     2. `plan->query_context()->max_concurrency()` query maximum DOP
> 
> 
> Am I right?

@mapleFU 

Yes, that is correct.

PR #35396, opened 2023-05-02
GH-35390: [Python] Consolidate some APIs in Table and RecordBatch

### LINKED ISSUE #35390: [Python] Consolidate some identical pyarrow Table and RecordBatch APIs

Table and RecordBatch share identical implementations of many APIs. Let's put this in their shared parent class.

### Component(s)

Python

danepitkin:
### Rationale for this change

Table and RecordBatch share many similar/identical APIs. Let's start consolidating them into the parent class. This diff refactors:

`__dataframe__`
`__eq__`
`__getitem__`
`_ensure_integer_index`
`column`
`columns`
`field`
`sort_by`

### Are these changes tested?

Relying on existing test cases for this refactor.

No
* Closes: #35390

### DIFF
--- a/python/pyarrow/table.pxi
+++ b/python/pyarrow/table.pxi
@@ -1600,7 +1600,7 @@ cdef class _Tabular(_PandasConvertible):
 
         Returns
         -------
-        columsn : list of Array (for RecordBatch) or list of ChunkedArray (for Table)
+        columns : list of Array (for RecordBatch) or list of ChunkedArray (for Table)
 
         Examples
         --------
--- a/python/pyarrow/table.pxi
+++ b/python/pyarrow/table.pxi
@@ -1593,6 +1593,45 @@ cdef class _Tabular(_PandasConvertible):
         """
         return self._column(self._ensure_integer_index(i))
 
+    @property
+    def columns(self):
+        """
+        List of all columns in numerical order.
+
+        Returns
+        -------
+        columsn : list of Array (for RecordBatch) or list of ChunkedArray (for Table)
+
+        Examples
+        --------
+        Table (works similarly for RecordBatch)
+
+        >>> import pyarrow as pa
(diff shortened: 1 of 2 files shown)


danepitkin:
Build error is unrelated. The error exists on main (and in other PRs).

python/pyarrow/table.pxi
```diff
@@ -1963,106 +2113,16 @@ cdef class RecordBatch(_Tabular):
...
-                return field_indices[0]
-        elif isinstance(i, int):
-            return i
-        else:
-            raise TypeError("Index must either be string or integer")
+        return self._schema
 
-    def column(self, i):
+    @property
+    def columns(self):
```
AlenkaF, line 2119: Could we consolidate `columns` also? It seems both could be using `_column` or `column`.

python/pyarrow/table.pxi
```diff
@@ -1500,6 +1627,68 @@ cdef class _Tabular(_PandasConvertible):
...
+            each entry is a tuple with column name
+            and sorting order ("ascending" or "descending")
+        **kwargs : dict, optional
+            Additional sorting options.
+            As allowed by :class:`SortOptions`
+
+        Returns
+        -------
+        Table or RecordBatch
+            A new tabular object sorted according to the sort keys.
```
AlenkaF, line 1681: Could we move the example from the `Table` `sort_by` to here?

AlenkaF (commented):
Looking great! Adding just two comments.

python/pyarrow/table.pxi
danepitkin: Absolutely!

danepitkin:
Updated the diff to address comments!

AlenkaF:
Thanks! Will merge, the failure is not related and the connected tests on that build are passing 👍

PR #35398, opened 2023-05-03
GH-35260: [R] Add s3_init helper for adjusting log level

### LINKED ISSUE #35260: [R] Allow users to adjust AWS S3 log level from R

The AWS SDK has built-in logging and the C++ Arrow S3 implementation provides a way to initialize S3 with a log level other than the default (Fatal). Setting the log level above the default of Fatal is useful for both users and developers who might want to understand what the AWS SDK is doing behind the scenes. My immediate use case is to measure the timing and number of HTTP requests `arrow::write_dataset` triggers for a workload I'm troubleshooting.

The Python bindings expose such an option as:

```python
import pyarrow
from pyarrow import fs

fs.initialize_s3(fs.S3LogLevel.Debug)
```

As far as I can tell, the R bindings don't expose anything similar. Would this be worth a PR? It could look something like:

```r
library(arrow)

arrow:::s3_init("Debug") # Output starts printing to the console
s3_bucket(....) # Prints lots of debug info to console
```

I'm not sure whether my example `s3_init` should be exported or not so I'm also curious for thoughts on that.

### Component(s)

R

amoeba:

Adds a non-exported helper function, `s3_init`, that lets users optionally control the log level of the AWS C++ SDK as well as the number of threads it uses for its I/O event loop. This brings R's functionality into alignment with PyArrow.

### Are these changes tested?

Tests are included just to cover input validation and I've manually tested the code locally to make sure it's working as expected.

This function is not currently exported but we can discuss this. I'm thinking it probably should be and that it should be linked from within the documentation so users can find it.
* Closes: #35260


r/R/filesystem.R
```diff
@@ -442,6 +442,57 @@ default_s3_options <- list(
   request_timeout = -1
 )
 
+# Enumerate possible values for AWS SDK log levels, used by s3_init
+s3_log_levels <- c("off", "fatal", "error", "warn", "info", "debug", "trace")
```
thisisnic, line 446: Instead of defining these here, you can set the `log_level` parameter in the function definition to this vector instead.

r/R/filesystem.R
```diff
@@ -442,6 +442,57 @@ default_s3_options <- list(
...
+    stop(
+      "Argument 'log_level' must be one of: ",
+      paste(
+        paste(s3_log_levels[seq_len(length(s3_log_levels) - 1)], collapse = ", "),
+        s3_log_levels[length(s3_log_levels)],
+        sep = ", or "),
+      paste0(" but was '", log_level_normalized, "'."),
+      call. = FALSE)
+    return(invisible())
+  }
```
thisisnic, line 480: This level of error messaging is awesome, and I love the detail here.  However, you can simplify the code in this function by setting the default value of `log_level` to the vector of possible values and then using `rlang::arg_match()` (see docs at https://rlang.r-lib.org/reference/arg_match.html) to do the hard work for you.

I see you've accounted for converting values to lowercase, but this could be handled elsewhere.

r/R/filesystem.R
```diff
@@ -442,6 +442,57 @@ default_s3_options <- list(
...
+
+  # Validate num_event_loop_threads
+  if (num_event_loop_threads < 1) {
+    stop(
+      "Argument 'num_event_loop_threads' should be a positive integer ",
+      "greater than or equal to 1 but was '", num_event_loop_threads, "'.",
+      call. = FALSE
+    )
+
+    return(invisible())
```
thisisnic, line 490: Why return `invisible()` here?

thisisnic (changes requested):
Thanks for this PR, it'll be good to have this extra level of logging exposed for debugging purposes!

I'm a little bit hesitant about using a function that sets state in a session in this way; there's something which feels a little unidiomatic.

In the sparklyr package, they use a function to collect entries from the spark log, though that presupposes the log entries already having been collected elsewhere.

We already use `options()` to control things like use of multithreading (i.e. via `arrow.use_threads`), or whether the `pull()` function returns a vector not (`arrow.pull_as_vector`).  Perhaps we keep `s3_init()` as an internal function, but conditionally call it if some option (e.g. `arrow.s3_log_level`) is set?  

Actually, was this always intended to be an internal function (as I see it's not been exported), or was the `@export` tag missed?

r/R/filesystem.R
amoeba: Good point, seems fine to just do that.

r/R/filesystem.R
amoeba: Ah, that looks a nice improvement. Thanks for the pointer.

r/R/filesystem.R
amoeba: Now that I look at it with fresh eyes, that's pretty silly of me. No need for it. Good catch.

amoeba:
Thanks for taking a look @thisisnic.

> I'm a little bit hesitant about using a function that sets state in a session in this way; there's something which feels a little unidiomatic.

Yeah I agree, esp about it not feeling idiomatic, though I think it's hard to avoid due to how the AWS C++ SDK is designed. Reading https://docs.aws.amazon.com/sdk-for-cpp/v1/developer-guide/basic-use.html, it looks like any session should init the API once, which is the point where the application specifies the log level it desires for the remainder of the session. I think we could totally do something with options as you suggest but it would make it easier for the user to run into this scenario:

```r
options(arrow.s3_log_level = "fatal")
s3_bucket() # Indirectly causes initialization of the API with log level set to Fatal
options(arrow.s3_log_level = "debug") # <---- Has no effect, log level stays set to Fatal
```

With the implementation in this PR, the user would instead get an error:

```r
> arrow:::s3_init(log_level = "debug")
Error: Invalid: S3 was already initialized.  It is safe to use but the options passed in this call have been ignored.
```

> Actually, was this always intended to be an internal function (as I see it's not been exported), or was the @export tag missed?

I left it unexported just until it was discussed in review and since I wasn't sure. I'm leaning towards export but am curious what you think.

thisisnic:
It would make it easier for the user to run into this scenario:
> 
> ```r
> options(arrow.s3_log_level = "fatal")
> s3_bucket() # Indirectly causes initialization of the API with log level set to Fatal
> options(arrow.s3_log_level = "debug") # <---- Has no effect, log level stays set to Fatal
> ```
> With the implementation in this PR, the user would instead get an error:
> 
> ```r
> > arrow:::s3_init(log_level = "debug")
> Error: Invalid: S3 was already initialized.  It is safe to use but the options passed in this call have been ignored.
> ```

Ah, that's a shame; I agree that the silent swallowing of errors isn't great.  Would it be any better if we called `s3_init()` inside `s3_bucket()` instead, and rather than using the `options` above, we update the signature of `s3_bucket()` to take a new parameter, `log_level` to define the value, or would we still run into the same problem? 

> I left it unexported just until it was discussed in review and since I wasn't sure. I'm leaning towards export but am curious what you think.

I feel like it'd be helpful to expose this functionality to (advanced) end users, but it's still unclear how best to do this in a way which would be consistent with their expectations for how this would work. It's tricky!

amoeba:
> ...would we still run into the same problem?

I think we would still have this issue:

```r
s3_bucket("my-buk", log_level = "fatal")
s3_bucket("my-buk", log_level = "debug") # Doesn't do what user expects because S3 is already init'd
```

We could issue a helpful warning here, though we could also do that with the proposed s3_init function too.

@paleolimbot thank you for spitballing! I think see where you're going but I think we could avoid keeping state in the R side and just wrap [`arrow::fs::IsS3Initialized()`](https://github.com/apache/arrow/blob/main/cpp/src/arrow/filesystem/s3fs.cc#L2709).

I think the crux of the problem is that the AWS C++ SDK is forcing us to have immutable state and IMO is that isn't a common pattern in R. It might be possible make this state mutable by re-initializing the SDK on the user's behalf when we detect we need to (e.g. the user changes the log level after `s3_bucket()` but before `bucket$ls()` ) but that feels clunky and I'm not sure how well it'd work. I think shutting down the API (`arrow::fs::FinalizeS3()`) involves blocking until all SDK resources (e.g., threads) are free'd up.

I might do a bit of searching around to see if there are other packages that deal with this kind of situation. If we ultimately don't find something we like, I'd be fine closing this PR and saving it for another time.

thisisnic:
> We could issue a helpful warning here

IMO, that would be plenty here.

amoeba:
Okay, I'll refactor and ping you when it's ready for another look. Thanks @thisisnic !

amoeba:
Hey @thisisnic, this is ready for another look. With this refactor, you can turn logging on either by:

```r
s3_bucket("my_bucket", log_level = "debug")
# or
S3FileSystem$create(log_level = "debug")
```

And doing either a second time with log_level set issues a warning:

> Warning message:
> S3 is already initialized. Argument 'log_level' ignored. 

But calling s3_bucket/S3FileSystem$create w/o `log_level` set repeatedly does not and the initially set log_level is retained.

r/R/filesystem.R
```diff
@@ -384,7 +387,24 @@ S3FileSystem <- R6Class("S3FileSystem",
     region = function() fs___S3FileSystem__region(self)
   )
 )
-S3FileSystem$create <- function(anonymous = FALSE, ...) {
+S3FileSystem$create <- function(
+  anonymous = FALSE,
+  log_level = c("fatal", "error", "warn", "info", "debug", "trace", "off"),
+  ...) {
+  # Init S3 now only if log_level is set
+  # TODO: This is clunky
```
amoeba, line 395: I thought this looked clunky but wasn't sure how best to guard this code. It seems like the way you're supposed to use `rlang::arg_match` is to make the first option in your set of options the default and just write code like,

```r
S3FileSystem$create <- function(log_level = c("fatal", "error", "warn", "info", "debug", "trace", "off")) {
  log_level <- rlang::arg_match(log_level)
  InitS3(log_level)

  # continue
}
```

However we don't always want to call InitS3(), hence the reason for guarding it like I did. Any suggestions welcome.

paleolimbot: With apologies if we're sending you around in circles, I don't know if `S3Filesystem$create()` is the right place for this. Why not an `s3_initialize()` function? If somebody wants to set the logging level, they might do something like:

```r
library(arrow)
s3_initialize(log_level = "verbose")
```

Alternatively, `options()` could be another somewhat idiomatic way to do this:

```r
library(arrow)
options(arrow.s3.log_level = "verbose")

# ...first call to something S3-y would check getOption("arrow.s3.log_level") if
# S3 hadn't been initialized yet.
```

amoeba: No worries at all. A stand-alone `s3_initialize` was how I did my first version of this PR but the discussion lead us down putting the functionality inside `S3FileSystem$create`. `options()` was also suggested before and is pretty good but has one downside, described in https://github.com/apache/arrow/pull/35398#issuecomment-1533867236. Maybe the downside I describe there is acceptable and we just document it?

paleolimbot: Ah, I should have done a better job revisiting the discussion/reviewing the first time!. I think putting it as an argument inside `S3Filesystem$create()` is goofy because the lifecycle of the parameter has nothing to do with the lifecycle of the filesystem that is returned. Somebody is much more likely to attempt `s3_bucket(log_level = ...)` with varying log levels than put calls to `s3_initialize()` all over the place.

Of course, any of these alternatives is better than none, which is what we currently have! I don't really have a strong feeling here and am happy to defer to earlier discussion.

To more literally help with the original comment here, you could do:

```r
$create <- function(log_level = NULL, ...) {
  log_level <- log_level %||% "whater_the_default_value_is"
  log_level <- rlang::arg_match(log_level, c("vebose", "other", "choices"))
  ...
}
```

...and list the choices in the documentation.

amoeba:
Thanks for the reviews @thisisnic and @paleolimbot. Now that we've reviewed two different implementations of this, is one a clear winner? I'm in favor of having a standalone `s3_init` that the user calls out of band of any S3 function calls which produces an error if it's called multiple times in a session. I think I'm also in favor of not exporting it at this time, and we can always choose to export it later on if we find we like it enough.

amoeba:
I had another thought about how to do something idiomatic while not confusing users: What about just having an environment variable like `ARROW_S3_LOG_LEVEL` that all C++-based implementations could make use of? So whenever S3 initialization happens on the C++ side, the env var is checked and the log level is set accordingly.

amoeba:
It does, though how it works is a bit odd: #35575 due to how the Python imports are structured. It Though it is currently more to similar to an s3_init helper from R.

thisisnic:
@amoeba - should I close this PR if you're going with #36115 instead?

amoeba:
Yep! Closing in favor of approach outlined in https://github.com/apache/arrow/issues/36115.

Thanks @thisisnic and @paleolimbot for the review on this one.

PR #35407, opened 2023-05-03
MINOR: [R] ARROW_ACERO should be ON by default in bundled build

nealrichardson:
To match ARROW_DATASET. Without this, the default CRAN version on Linux won't have Acero enabled.

This should be cherry-picked for the 12.0.0 CRAN submission. cc @thisisnic


thisisnic (approved):
Good catch!

PR #35428, opened 2023-05-04
GH-35423: [C++][Parquet] Parquet PageReader Force decompression buffer resize smaller

### LINKED ISSUE #35423: [C++][Parquet] "Data size too large" error with byte-stream-split encoded data since Arrow 12.0.0

Arrow 12.0.0 has a regression where it can crash when reading byte-stream split encoded data written by itself or older versions of Arrow:

```python
import numpy as np
import pyarrow as pa
import pyarrow.parquet as pq

x = pa.array(np.linspace(0.0, 1.0, 1_000_000), type=pa.float32())
table = pa.Table.from_arrays([x], names=['x'])
pq.write_table(table, 'data.parquet', use_dictionary=False, use_byte_stream_split=True)

table = pq.read_table('data.parquet')
print(table)
```
This crashes with:
```
  File "/home/.../write_read_data.py", line 9, in <module>
	table = pq.read_table('data.parquet')
			^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/adam/.local/share/virtualenvs/arrow12-nieuEBn0/lib64/python3.11/site-packages/pyarrow/parquet/core.py", line 2986, in read_table
	return dataset.read(columns=columns, use_threads=use_threads,
		   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/adam/.local/share/virtualenvs/arrow12-nieuEBn0/lib64/python3.11/site-packages/pyarrow/parquet/core.py", line 2614, in read
	table = self._dataset.to_table(
			^^^^^^^^^^^^^^^^^^^^^^^
  File "pyarrow/_dataset.pyx", line 546, in pyarrow._dataset.Dataset.to_table
  File "pyarrow/_dataset.pyx", line 3449, in pyarrow._dataset.Scanner.to_table
  File "pyarrow/error.pxi", line 115, in pyarrow.lib.check_status
OSError: Data size too large for number of values (padding in byte stream split data page?)
```
But the above code works fine with pyarrow 11.0.0 and and 10.0.1.

It appears that #34140 caused this regression. I tested building pyarrow on the current main branch (commit 42d42b1194d8a672e13dac10a8102573f787f70d) and could reproduce the error, but it was fixed after I reverted the merge of that PR (commit c31fb46544b9c8372e799138bad9223162169473).

### Component(s)

Parquet

mapleFU:
### Rationale for this change

See in the issue. 

1. SerializedPageReader will reuse same decompression buffer, **and didn't resize if next page is smaller than previous one**, so, `buffer->size()` might not equal to `page.uncompressed_size()`.
2. So, `data_size` in decoder would be equal to previous page size.
3. When it goes to BYTE_STREAM_SPLIT, BYTE_STREAM_SPLIT will check the size not too large. This will cause throw an exception. When it goes to other readers, here is no bug, because uninitialized memory will not be accessed

Change reader and add test

### Are these changes tested?

Yes

No

* Closes: #35423


mapleFU:
The real reason is that:

```c++
std::shared_ptr<Buffer> SerializedPageReader::DecompressIfNeeded(
    std::shared_ptr<Buffer> page_buffer, int compressed_len, int uncompressed_len,
    int levels_byte_len) {
  if (decompressor_ == nullptr) {
    return page_buffer;
  }
  if (compressed_len < levels_byte_len || uncompressed_len < levels_byte_len) {
    throw ParquetException("Invalid page header");
  }

  // Grow the uncompressed buffer if we need to.
  if (uncompressed_len > static_cast<int>(decompression_buffer_->size())) {
    PARQUET_THROW_NOT_OK(
        decompression_buffer_->Resize(uncompressed_len, /*shrink_to_fit=*/false));
  }

  if (levels_byte_len > 0) {
    // First copy the levels as-is
    uint8_t* decompressed = decompression_buffer_->mutable_data();
    memcpy(decompressed, page_buffer->data(), levels_byte_len);
  }

  // Decompress the values
  PARQUET_THROW_NOT_OK(decompressor_->Decompress(
      compressed_len - levels_byte_len, page_buffer->data() + levels_byte_len,
      uncompressed_len - levels_byte_len,
      decompression_buffer_->mutable_data() + levels_byte_len));

  return decompression_buffer_;
}
```

(1) Here the buffer size will only grow up, and never thrink.
(2) Page.buffer uses this buffer size
(3) BYTE_STREAM_SPLIT will check this size

mapleFU:
@adamreeve Would you mind try this?

cpp/src/parquet/column_reader.cc
```diff
@@ -857,7 +857,10 @@ class ColumnReaderImplBase {
   // first page with this encoding.
   void InitializeDataDecoder(const DataPage& page, int64_t levels_byte_size) {
     const uint8_t* buffer = page.data() + levels_byte_size;
-    const int64_t data_size = page.size() - levels_byte_size;
+    // PageReader may reuse the underlying buffer, so data_size
+    // should use `page.uncompressed_size() - levels_byte_size` rather
+    // than `page.size() - level_byte_size`.
+    const int64_t data_size = page.uncompressed_size() - levels_byte_size;
```
wgtmac, line 863: nit: it would be helpful to comment this is now a precise uncompressed data size, compared to an upper bound in the past.

cpp/src/arrow/dataset/file_parquet_test.cc
```diff
@@ -222,6 +222,42 @@ TEST_F(TestParquetFileFormat, WriteRecordBatchReaderCustomOptions) {
...
+TEST_F(TestParquetFileFormat, WriteRecordBatchReaderEncoding) {
+  auto float_schema = schema({field("td", float32())});
+  auto options =
+      checked_pointer_cast<ParquetFileWriteOptions>(format_->DefaultWriteOptions());
+  options->writer_properties = parquet::WriterProperties::Builder()
+                                   .created_by("TestParquetFileFormat")
+                                   ->disable_statistics()
+                                   ->encoding(parquet::Encoding::BYTE_STREAM_SPLIT)
+                                   ->disable_dictionary()
+                                   ->compression(Compression::SNAPPY)
```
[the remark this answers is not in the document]
mapleFU: Updated

cpp/src/parquet/column_reader.cc
wgtmac: Did you mean here: https://github.com/apache/arrow/blob/main/cpp/src/parquet/column_page.h#L53-L54 ?

It currently says `the total size in bytes of the page's data buffer` and it is kind of misleading. I agree we need to improve it. Something like `its size may exceed the actual size in use`? Please suggest a better statement @wjones127 

Or should we fix the `size()` function to always return the actual size in use to get rid of the confusion?

mapleFU: So, here we should make `uncompressed_size` and `size` be equal (at least in DATA_PAGE V1?) I can fix it later this week

wjones127: Yes, that's where I meant. If there's some way to fix the `size()` function I would prefer that.

wjones127: Yeah I'm thinking could we instead fix change these lines:

https://github.com/apache/arrow/blob/99ac74e6890767e1a3cf96d7aff0b96b9503153c/cpp/src/parquet/column_reader.cc#L584-L588

To something like (change the inequality to `!=`):

```cpp
 // Grow the uncompressed buffer if we need to. Regardless, make sure the buffers
 // size is reported correctly.
 if (uncompressed_len != static_cast<int>(decompression_buffer_->size())) { 
   PARQUET_THROW_NOT_OK( 
       decompression_buffer_->Resize(uncompressed_len, /*shrink_to_fit=*/false)); 
 } 
```

Would that fix the same bug?

mapleFU:
@wgtmac @wjones127 I've change the implemention in `SerializedPageReader` internal, mind take a look?

wjones127 (approved):
This looks better. Thanks @mapleFU!

mapleFU:
```
In file included from /arrow/cpp/src/arrow/builder_benchmark.cc:32:
/arrow/cpp/src/arrow/util/benchmark_util.h:150:8: error: 'void arrow::MemoryPoolMemoryManager::Stop(benchmark::MemoryManager::Result*)' marked 'override', but does not override
  150 |   void Stop(benchmark::MemoryManager::Result* result) override {
      |        ^~~~
/arrow/cpp/src/arrow/util/benchmark_util.h:201:36: error: cannot declare field 'arrow::BenchmarkMemoryTracker::manager_' to be of abstract type 'arrow::MemoryPoolMemoryManager'
  201 |   ::arrow::MemoryPoolMemoryManager manager_;
      |                                    ^~~~~~~~
/arrow/cpp/src/arrow/util/benchmark_util.h:142:7: note:   because the following virtual functions are pure within 'arrow::MemoryPoolMemoryManager':
  142 | class MemoryPoolMemoryManager : public benchmark::MemoryManager {
      |       ^~~~~~~~~~~~~~~~~~~~~~~
In file included from /arrow/cpp/src/arrow/builder_benchmark.cc:27:
/opt/conda/envs/arrow/include/benchmark/benchmark.h:408:16: note:     'virtual void benchmark::MemoryManager::Stop(benchmark::MemoryManager::Result&)'
  408 |   virtual void Stop(Result& result) = 0;
      |                ^~~~
```

Emm, seems this isn't caused by me? I'm not sure how it happens...

wgtmac (approved):
Thanks @mapleFU and @wjones127! LGTM

cpp/src/parquet/file_deserialize_test.cc
```diff
@@ -196,6 +197,9 @@ class TestPageSerde : public ::testing::Test {
                         bool verification_checksum, bool has_dictionary = false,
                         bool write_data_page_v2 = false);
 
+  void TestPageCompressionRoundTrip(int num_pages,
+                                    const std::function<int(int)>& getPageSize);
```
pitrou, line 201: Why not allow simply passing a vector of page sizes?
```suggestion
  void TestPageCompressionRoundTrip(const std::vector<int32_t>& page_sizes);
```

cpp/src/parquet/file_deserialize_test.cc
```diff
@@ -753,7 +755,19 @@ TEST_F(TestPageSerde, Compression) {
 
     ResetStream();
   }
-}  // namespace parquet
+}
+
+TEST_F(TestPageSerde, Compression) {
+  // The pages keep getting larger
```
pitrou, line 761: Can you add a reference to the GH bug id so as to track what motivates these tests?

cpp/src/arrow/dataset/file_parquet_test.cc
```diff
@@ -222,6 +222,46 @@ TEST_F(TestParquetFileFormat, WriteRecordBatchReaderCustomOptions) {
                     *actual_schema);
 }
 
+#ifdef ARROW_WITH_SNAPPY
+
+TEST_F(TestParquetFileFormat, WriteRecordBatchReaderByteStreamSplit) {
```
pitrou, line 227: Is it necessary to use datasets for this test, or can the test be implemented in the regular Parquet C++ test suite?

pitrou (changes requested):
Thanks for the analysis and fix @mapleFU , I posted a couple suggestions.

cpp/src/parquet/file_deserialize_test.cc
mapleFU: Sure. Previously here is for testing crc and compression. I'll add motivation for my new tests here

mapleFU:
Resolve comment except https://github.com/apache/arrow/pull/35428#discussion_r1188637986 . If not neccessary I can just remove it ( personally I think dataset didn't testing ByteStreamSplit and Compression before, so it's ok to add one)

cpp/src/arrow/dataset/file_parquet_test.cc
pitrou: I'll let @westonpace comment on whether this test is relevant for the datasets layer.

westonpace: If the feature is well tested in the `file_deserialize_test.cc` then I think we should remove this.  Even though it was found using datasets I don't think the root cause was in datasets and it doesn't appear that there are any changes to `file_parquet.cc`.

mapleFU: Okay, changes are removed.

mapleFU:
Changes in dataset testing is removed. Would you mind take a look again?

mapleFU:
Change the description of this patch

pitrou (approved):

Issue #35201, opened 2023-04-17
[C++][Gandiva] Add a Secondary Cache to cache gandiva object code


schavan6:
### Describe the enhancement requested

Arrow Gandiva has a primary cache but this cache doesn't persist across
restarts. Integrate a new API in project and filter make calls that
allow the user to specify the implementation of the secondary cache by
providing a c++ and java interface to this persistent cache.

### Component(s)

C++, C++ - Gandiva, Java

PR #35447, opened 2023-05-05
GH-35201: [C++][Gandiva] Add a Secondary Cache to cache gandiva object code

### LINKED ISSUE #203: [C++][Gandiva] Add a Secondary Cache to cache gandiva object code
Arrow Gandiva has a primary cache but this cache doesn't persist across
restarts. Integrate a new API in Project and Filter and make calls that
allow the user to specify the implementation of the secondary cache by
providing a c++ and java interface to this persistent cache.

### Component(s)

C++, C++ - Gandiva, Java

schavan6:
### Rationale for this change

Arrow gandiva has a primary cache but this cache doesn't persist across restarts.

Integrate a new API in project and filter and make calls that allow the user to specify the implementation of the secondary cache by providing a c++ and java interface to this persistent cache.

### Are these changes tested?
Yes.

No.

* Closes: apache/arrow-java#203

### DIFF
--- a/cpp/src/gandiva/cache.h
+++ b/cpp/src/gandiva/cache.h
@@ -50,6 +50,11 @@ class Cache {
     cache_.insert(cache_key, module);
   }
 
+  void Clear() {
+    std::lock_guard<std::mutex> lock(mtx_);
+    cache_.clear();
+  }
+
  private:
   LruCache<KeyType, ValueType> cache_;
   std::mutex mtx_;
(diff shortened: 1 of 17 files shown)


kou:
Did you run a benchmark with this?
How much overhead (slow down) with this?

How about adding support for customizing the current cache (`LLVMGenerator::GetCache()`) (using an user defined cache object) instead of adding a new cache mechanism?

PR #35449, opened 2023-05-05
GH-35448: [C++] Fix detection of %z in strptime format

### LINKED ISSUE #35448: [C++] Detection of %z in strptime format fails in certain cases, not returning timestamp with tz=UTC

https://github.com/apache/arrow/pull/11358 refactored the timezone offset handling in string parsing (both in read_csv, casting str->timestamp and strptime). 

The general rule was that if there is a `%z` in the format string you provide to `strptime`, we apply the offset and return a timestamp type with tz=UTC. 
But this seems to work only in certain cases:

```python
# space before %z -> works
>>> pc.strptime(["5/1/2020 +0100", None, "12/11/1900 -0130"], format="%m/%d/%Y %z", unit='us').type
TimestampType(timestamp[us, tz=UTC])
# no space before %z -> doesn't work
>>> pc.strptime(["5/1/2020+0100", None, "12/11/1900-0130"], format="%m/%d/%Y%z", unit='us').type
TimestampType(timestamp[us])
```

### Component(s)

C++

jorisvandenbossche:
### Rationale for this change

See gh-35448 for the failing example. The current code in `GetZone` was assuming there was always some character between the `%z` and the preceding `%` code (like a whitespace, or `-` or `/`). That is often not the case with `%z` (in time formats like `00:00+01`, the `+` is part of `%z`, and so the format is `%H:%M%z` without character between `%M` and `%z`)
 
### Are these changes tested?

Test is added

The result type will no now correctly have a `tz=UTC` parametrization
* Closes: #35448

### DIFF
--- a/cpp/src/arrow/compute/kernels/scalar_string_test.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_string_test.cc
@@ -1900,6 +1900,12 @@ TYPED_TEST(TestStringKernels, StrptimeZoneOffset) {
   StrptimeOptions options("%m/%d/%Y %z", TimeUnit::MICRO, /*error_is_null=*/true);
   this->CheckUnary("strptime", input1, timestamp(TimeUnit::MICRO, "UTC"), output1,
                    &options);
+
+  // format without whitespace before %z (GH-35448)
+  std::string input2 = R"(["2020-05-01T00:00+0100", null, "1900-12-11T00:00-0130"])";
+  StrptimeOptions options("%Y-%m-%dT%H:%M%z", TimeUnit::MICRO, /*error_is_null=*/true);
+  this->CheckUnary("strptime", input2, timestamp(TimeUnit::MICRO, "UTC"), output1,
+                   &options);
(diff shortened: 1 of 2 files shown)


cpp/src/arrow/compute/kernels/scalar_string_test.cc
```diff
@@ -1900,6 +1900,12 @@ TYPED_TEST(TestStringKernels, StrptimeZoneOffset) {
   StrptimeOptions options("%m/%d/%Y %z", TimeUnit::MICRO, /*error_is_null=*/true);
   this->CheckUnary("strptime", input1, timestamp(TimeUnit::MICRO, "UTC"), output1,
                    &options);
```
kou, line 1902: Could you append `1` to `options` (`options1`) for consistency?

kou (approved):
+1

westonpace (approved):

pitrou (approved):
+1, thanks @jorisvandenbossche

PR #35451, opened 2023-05-05
[FORMAT]: Minor Fix `VarBinary` examples to use Values buffer

clickingbuttons:
I'm implementing Arrow and noticed an inconsistency in the docs. After reading the cpp and Rust implementations and talking to @tustvold I believe VarBinary is actually laid out according to this change.


alamb:
Thank you @clickingbuttons

tustvold (approved):

PR #35453, opened 2023-05-05
GH-35331: [Python] Expose Parquet sorting metadata

### LINKED ISSUE #35331: [C++][Python] Expose `sorting_columns` in RowGroupMetaData for Parquet files

wjones127:
### Rationale for this change

Parquet supports metadata to indicate sort order of the data. This exposes that in a low-level API through row group metadata.

* Adds a new class `pyarrow.parquet.SortingColumn` to represent sorting of a particular column. Since this represents column references in terms of Parquet's standard for column indices, provided method to convert to and from column names.
* Exposes sort order of individual row groups in the metadata.

### Are these changes tested?

Yes, I've added unit tests for all new functionality.

This adds new APIs and provides documentation for them.

* Closes: #35331

### DIFF
--- a/python/pyarrow/_parquet.pxd
+++ b/python/pyarrow/_parquet.pxd
@@ -24,7 +24,7 @@ from pyarrow.includes.libarrow cimport (CChunkedArray, CScalar, CSchema, CStatus
                                         CKeyValueMetadata,
                                         CRandomAccessFile, COutputStream,
                                         TimeUnit, CRecordBatchReader)
-from pyarrow.lib cimport _Weakrefable
+from pyarrow.lib cimport (_Weakrefable, pyarrow_unwrap_schema)
 
 
 cdef extern from "parquet/api/schema.h" namespace "parquet::schema" nogil:
@@ -322,11 +322,17 @@ cdef extern from "parquet/api/reader.h" namespace "parquet" nogil:
         int64_t total_uncompressed_size() const
         unique_ptr[CColumnCryptoMetaData] crypto_metadata() const
 
+    struct CSortingColumn" parquet::SortingColumn":
+        int column_idx
+        c_bool descending
+        c_bool nulls_first
+
     cdef cppclass CRowGroupMetaData" parquet::RowGroupMetaData":
         c_bool Equals(const CRowGroupMetaData&) const
-        int num_columns()
-        int64_t num_rows()
-        int64_t total_byte_size()
+        int num_columns() const
+        int64_t num_rows() const
+        int64_t total_byte_size() const
+        vector[CSortingColumn] sorting_columns() const
         unique_ptr[CColumnChunkMetaData] ColumnChunk(int i) const
 
     cdef cppclass CFileMetaData" parquet::FileMetaData":
@@ -410,6 +416,7 @@ cdef extern from "parquet/api/writer.h" namespace "parquet" nogil:
             Builder* disable_dictionary()
             Builder* enable_dictionary()
             Builder* enable_dictionary(const c_string& path)
+            Builder* set_sorting_columns(vector[CSortingColumn] sorting_columns)
             Builder* disable_statistics()
             Builder* enable_statistics()
             Builder* enable_statistics(const c_string& path)
@@ -502,8 +509,8 @@ cdef extern from "parquet/arrow/schema.h" namespace "parquet::arrow" nogil:
 
     CStatus ToParquetSchema(
         const CSchema* arrow_schema,
-        const ArrowReaderProperties& properties,
-        const shared_ptr[const CKeyValueMetadata]& key_value_metadata,
+        const WriterProperties& properties,
+        const ArrowWriterProperties& arrow_properties,
         shared_ptr[SchemaDescriptor]* out)
 
 
@@ -567,7 +574,8 @@ cdef shared_ptr[WriterProperties] _create_writer_properties(
     data_page_version=*,
     FileEncryptionProperties encryption_properties=*,
     write_batch_size=*,
-    dictionary_pagesize_limit=*) except *
+    dictionary_pagesize_limit=*,
+    sorting_columns=*) except *
 
 
 cdef shared_ptr[ArrowWriterProperties] _create_arrow_writer_properties(
--- a/python/pyarrow/_parquet.pyx
+++ b/python/pyarrow/_parquet.pyx
@@ -531,8 +531,7 @@ cdef class SortingColumn:
 
     >>> import pyarrow.parquet as pq
     >>> [pq.SortingColumn(0), pq.SortingColumn(1, descending=True)]
-    [SortingColumn(column_index=0, descending=False, nulls_first=False),
-     SortingColumn(column_index=1, descending=True, nulls_first=False)]
+    [SortingColumn(column_index=0, descending=False, nulls_first=False), SortingColumn(column_index=1, descending=True, nulls_first=False)]
 
     Convert the sort_order into the list of sorting columns with 
     ``from_sort_order`` (note that the schema must be provided as well):
@@ -541,8 +540,7 @@ cdef class SortingColumn:
     >>> schema = pa.schema([('id', pa.int64()), ('timestamp', pa.timestamp('ms'))])
     >>> sorting_columns = pq.SortingColumn.from_sort_order(schema, sort_order)
     >>> sorting_columns
-    (SortingColumn(column_index=0, descending=False, nulls_first=False),
-     SortingColumn(column_index=1, descending=True, nulls_first=False))
+    (SortingColumn(column_index=0, descending=False, nulls_first=False), SortingColumn(column_index=1, descending=True, nulls_first=False))
 
     Convert back to the sort order with ``to_sort_order``:
 
--- a/python/pyarrow/_parquet.pyx
+++ b/python/pyarrow/_parquet.pyx
@@ -505,7 +505,7 @@ cdef class SortingColumn:
     Parameters
     ----------
     column_index : int
-        Index of column data is sorted by.
+        Index of column that data is sorted by.
     descending : bool, default False
         Whether column is sorted in descending order.
     nulls_first : bool, default False
@@ -691,14 +691,6 @@ cdef class SortingColumn:
     def nulls_first(self):
         """Whether null values appear before valid values (bool)."""
         return self.nulls_first
-
-    def to_dict(self):
-        """Convert to dictionary representation."""
-        return {
-            'column_index': self.column_index,
-            'descending': self.descending,
-            'nulls_first': self.nulls_first
-        }
 
 
 cdef class RowGroupMetaData(_Weakrefable):
--- a/python/pyarrow/_parquet.pyx
+++ b/python/pyarrow/_parquet.pyx
@@ -18,6 +18,7 @@
 # cython: profile=False
 # distutils: language = c++
 
+from collections.abc import Sequence
 from textwrap import indent
 import warnings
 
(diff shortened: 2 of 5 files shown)


wjones127:
Intended usage:

```python
import pyarrow as pa
import pyarrow.parquet as pq
import pyarrow.compute as pc

tab = pa.table({
    'a': pa.array([1, 3, 2, 2]),
    'b': pa.array(['a', 'b', 'c', 'd'])
})

sort_keys = [('a', 'ascending'), ('b', 'descending')]
sorted_tab = tab.sort_by(sort_keys)

pq.write_table(tab, 'data.parquet', sorting_columns=sort_keys)

metadata = pq.read_metadata('data.parquet')
metadata.row_group(0).sorting_columns

pq_file = pq.ParquetFile('data.parquet')
sort_order, null_placement = pq_file.sort_order
assert sort_order == sort_keys
assert null_placement == "at_end"
```

wjones127:
@westonpace This PR might interest you. It occurred to me while working with this we can persist the sort order of data in Parquet and retrieve it. Makes me wonder where we could integrate it into the rest of the code base.

python/pyarrow/parquet/core.py
```diff
@@ -867,6 +915,10 @@ def _sanitize_table(table, new_schema, flavor):
     it will restore the timezone (Parquet only stores the UTC values without
     timezone), or columns with duration type will be restored from the int64
     Parquet column.
+sorting_columns : Sequence of SortingColumn, default None
+    Specify the sort order of the data being written. The writer does not sort
+    the data nor does it verify that the data is sorted. The sort order is
```
mapleFU, line 920: So, user him/her self should ensure the correctness of sorting_columns?

westonpace:
> @westonpace This PR might interest you. It occurred to me while working with this we can persist the sort order of data in Parquet and retrieve it. Makes me wonder where we could integrate it into the rest of the code base.

It is interesting.  Though one challenge would be that a single file doesn't neccesarily tell you about an entire dataset.  For example, if the file foo.parquet is sorted by "date" then are all files in that dataset sorted by date?  Does foo.parquet come before bar.parquet?

That being said, we could probably hook it into exec plans pretty easily for the cases where the source dataset is a single file.

There may also be some benefit in knowing that individual files are sorted (even if the dataset is not) but I'm not quite sure how exactly we'd exploit that yet.

wjones127:
> Though one challenge would be that a single file doesn't necessarily tell you about an entire dataset. For example, if the file foo.parquet is sorted by "date" then are all files in that dataset sorted by date? Does foo.parquet come before bar.parquet?

Yeah, if you wanted to do that, you'd have to combine it with other metadata to get the sorting relationship between partitions and within partitions. Between partitions, you could sort the partition values and prepend that to the sorting order. For within partitions, you could aggregate the row group stats to file-level stats, and use those to see if they combine to an overall ordering.

So, for example, a dataset like:

```
part=A/file1.parquet (min x: 1, max x: 10)
part=A/file1.parquet (min x: 10, max x: 20)
part=B/file1.parquet (min x: 5, max x: 20)
```

If each of the files returned a sort order of `x, ascending`, then you could construct the dataset to have an overall ordering of `[(part, ascending), (x, ascending)]`.

Honestly, it would be nice just to be able to get a dataset to be considered "ordered" by its partition values. For example, if I had two tables partitioned by date, it would be nice to be able to join them via a sort merge join on that date column while using the inherent sorting of the partition values. (Having flashbacks to some frustrations in my days working with PySpark.)

westonpace:
> Honestly, it would be nice just to be able to get a dataset to be considered "ordered" by its partition values. For example, if I had two tables partitioned by date, it would be nice to be able to join them via a sort merge join (or maybe asof join?) on that date column while using the inherent sorting of the partition values. (Having flashbacks to some frustrations in my days working with PySpark.)

We are quite close to being able to do this.  In particular, I think we are very close to being able to take a query like "SELECT MEAN(x) FROM dataset GROUP BY part" and, instead of having aggregation be a pipeline breaker, have it scan values in partition order and emit values as soon as a partition is exhausted.  Here is what we have, and what is missing:

 * DONE: ExecBatch has an `index` property to determine its sorting.
 * DONE: ExecNode has an `ordering` property which says which columns its output are ordered by
 * DONE: The scan node has the ability to scan a dataset in order, this means the AsyncGenerator<ExecBatch> will emit exec batches in a deterministic order (filename, batch_in_file)
 * NEEDED: The scan node needs to detect if the dataset is partitioned, ask the dataset for fragments ordered by partition, and then assign its `ordering` property based on the partition columns (this sounds harder than it is)
 * DONE: The aggregate node is able to define "segment keys" which, if given, it will use to break up the output
 * NEEDED: The aggregate node's segment keys currently only work if run in single threaded mode.  We need to support parallelism.  This should be easily obtainable by using a sequencing queue
 * NOTE: The above assumes that the user intelligently crafts the plan (e.g. they know to assign "part" as a segment key when creating the aggregate node).  In reality, this is something that would happen in a planner (it would notice that "part" is a partition key, figure out that it can configure the scan to emit in sequence, and then use part when creating the aggregate node).  This means this feature will need to be exploited by reasonable intelligent users.  However, this is in line with Acero being a dumb execution plan without a planner.

python/pyarrow/_parquet.pxd
```diff
@@ -502,8 +509,8 @@ cdef extern from "parquet/arrow/schema.h" namespace "parquet::arrow" nogil:
 
     CStatus ToParquetSchema(
         const CSchema* arrow_schema,
-        const ArrowReaderProperties& properties,
-        const shared_ptr[const CKeyValueMetadata]& key_value_metadata,
+        const WriterProperties& properties,
+        const ArrowWriterProperties& arrow_properties,
```
wjones127, line 513: This declaration was never used and was out of date.

python/pyarrow/parquet/core.py
wjones127: Yes.

wjones127:
@westonpace would you have time to review next week?

westonpace:
Yes.

python/pyarrow/_parquet.pyx
```diff
@@ -1665,6 +1919,82 @@ cdef shared_ptr[ArrowWriterProperties] _create_arrow_writer_properties(
...
+        shared_ptr[SchemaDescriptor] sp_parquet_schema
+        shared_ptr[WriterProperties] props = _create_writer_properties()
+        shared_ptr[ArrowWriterProperties] arrow_props = _create_arrow_writer_properties(
+            use_deprecated_int96_timestamps=False,
+            coerce_timestamps=None,
+            allow_truncated_timestamps=False,
+            writer_engine_version="V2"
+        )
+
+    sp_arrow_schema = pyarrow_unwrap_schema(arrow_schema)
```
mapleFU, line 1934: so a nested path could be like `a.b.c`?

python/pyarrow/_parquet.pyx
```diff
@@ -1665,6 +1919,82 @@ cdef shared_ptr[ArrowWriterProperties] _create_arrow_writer_properties(
 
     return arrow_properties
 
+cdef _name_to_index_map(Schema arrow_schema):
+    cdef:
+        shared_ptr[CSchema] sp_arrow_schema
+        shared_ptr[SchemaDescriptor] sp_parquet_schema
+        shared_ptr[WriterProperties] props = _create_writer_properties()
+        shared_ptr[ArrowWriterProperties] arrow_props = _create_arrow_writer_properties(
+            use_deprecated_int96_timestamps=False,
```
mapleFU, line 1928: Is these default schema? or these doesn't matter?

python/pyarrow/_parquet.pyx
```diff
@@ -494,6 +495,212 @@ cdef class ColumnChunkMetaData(_Weakrefable):
...
+    cdef c_bool descending
+    cdef c_bool nulls_first
+
+    def __init__(self, int column_index, c_bool descending=False, c_bool nulls_first=False):
+        self.column_index = column_index
+        self.descending = descending
+        self.nulls_first = nulls_first
+
+    @classmethod
+    def from_sort_order(cls, Schema schema, sort_keys, null_placement='at_end'):
```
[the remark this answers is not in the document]
westonpace: ah, I see, perhaps to align `pc.sort_indices`?

python/pyarrow/_parquet.pyx
```diff
@@ -494,6 +495,212 @@ cdef class ColumnChunkMetaData(_Weakrefable):
...
+        ----------
+        schema : Schema
+            Schema of the input data.
+        sorting_columns : tuple of SortingColumn
+            Columns to sort the input on.
+
+        Returns
+        -------
+        sort_keys : tuple of (name, order) tuples
+        null_placement : {'at_start', 'at_end'}
```
westonpace, line 641: I assume this means "one of `at_start` or `at_end`"?  Is using curly braces here a standard way to do this?  The only way I'm familiar with is python typings which would be `Literal['at_start', 'at_end']` but I'm not really an expert in docstrings so I wouldn't be surprised if this is some pattern I'm just unfamiliar with.

python/pyarrow/_parquet.pyx
```diff
@@ -494,6 +495,212 @@ cdef class ColumnChunkMetaData(_Weakrefable):
...
+        """Whether null values appear before valid values (bool)."""
+        return self.nulls_first
+
+    def to_dict(self):
+        """Convert to dictionary representation."""
+        return {
+            'column_index': self.column_index,
+            'descending': self.descending,
+            'nulls_first': self.nulls_first
+        }
```
westonpace, line 701: Is this needed?

python/pyarrow/tests/parquet/test_metadata.py
```diff
@@ -301,6 +301,91 @@ def test_parquet_write_disable_statistics(tempdir):
...
+    )
+    writer = pa.BufferOutputStream()
+    _write_table(table, writer, sorting_columns=sorting_columns)
+    reader = pa.BufferReader(writer.getvalue())
+
+    # Can retrieve sorting columns from metadata
+    metadata = pq.read_metadata(reader)
+    set_sorting_columns = {metadata.row_group(i).sorting_columns
+                           for i in range(metadata.num_row_groups)}
+    assert set_sorting_columns == set([sorting_columns])
```
westonpace, line 386: Why a set here?  Can't we assert that there will only be one row group?

python/pyarrow/parquet/core.py
westonpace: There is a danger here but I think you explain it here well enough.  I can't think of any way to be more cautionary without just being overly verbose.

westonpace (approved):
A few minor questions / thoughts but no major concerns.

python/pyarrow/_parquet.pyx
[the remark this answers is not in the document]
wjones127: Yeah, was aiming to make familiar to users of that API.

python/pyarrow/_parquet.pyx
wjones127: Yeah it's part of numpydoc style: https://numpydoc.readthedocs.io/en/latest/format.html#parameters

python/pyarrow/_parquet.pyx
wjones127: I don't think so anymore. I'll remove.

python/pyarrow/_parquet.pyx
wjones127: Yeah these don't matter much. We just create them because the `ToParquetSchema` method asks for it.

python/pyarrow/_parquet.pyx
wjones127: Yes.

mapleFU (approved):

wgtmac:
@wjones127 Do you want to revive this?

mapleFU:
Seems I and weston all approve this pull request, what is the remaining blocker for this? Can we just rebase and move forward?

judahrand:
> Seems I and weston all approve this pull request, what is the remaining blocker for this? Can we just rebase and move forward?

Is there more work which needs doing here? I'm happy to pick this up and rebase?

PR #35455, opened 2023-05-06
GH-34375: [C++][Parquet] Ignore page header stats when page index enabled

### LINKED ISSUE #34375: [C++][Parquet] Page header does not save statistics once page index is enabled

Once writing page index is supported, we should not save page statistics in the data page header as it is duplicated. 

Although page stats is disabled by parquet-mr globally ([link](https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java#L1763)), we cannot do that because a lot of test cases in the parquet-cpp relies on the stats from page header.

### Component(s)

C++, Parquet

wgtmac:
### Rationale for this change

Page-level statistics are probably not used in production, and after adding column indexes they are useless. 
parquet-mr already stopped writing them in https://issues.apache.org/jira/browse/PARQUET-1365.

Once page index is enabled for one column, it does not write page stats to the header any more.

### Are these changes tested?

Added a test to check page stats have been skipped.

Yes (behavior change when page index is enabled).

* Closes: #34375

### DIFF
--- a/cpp/src/parquet/column_writer.cc
+++ b/cpp/src/parquet/column_writer.cc
@@ -460,7 +460,11 @@ class SerializedPageWriter : public PageWriter {
         ToThrift(page.definition_level_encoding()));
     data_page_header.__set_repetition_level_encoding(
         ToThrift(page.repetition_level_encoding()));
-    data_page_header.__set_statistics(ToThrift(page.statistics()));
+
+    // Write page statistics only when page index is not enabled.
+    if (column_index_builder_ == nullptr) {
+      data_page_header.__set_statistics(ToThrift(page.statistics()));
+    }
 
     page_header.__set_type(format::PageType::DATA_PAGE);
     page_header.__set_data_page_header(data_page_header);
@@ -479,7 +483,11 @@ class SerializedPageWriter : public PageWriter {
         page.repetition_levels_byte_length());
 
     data_page_header.__set_is_compressed(page.is_compressed());
-    data_page_header.__set_statistics(ToThrift(page.statistics()));
+
+    // Write page statistics only when page index is not enabled.
+    if (column_index_builder_ == nullptr) {
+      data_page_header.__set_statistics(ToThrift(page.statistics()));
+    }
 
     page_header.__set_type(format::PageType::DATA_PAGE_V2);
     page_header.__set_data_page_header_v2(data_page_header);
--- a/cpp/src/parquet/properties.h
+++ b/cpp/src/parquet/properties.h
@@ -525,7 +525,8 @@ class PARQUET_EXPORT WriterProperties {
     /// Enable writing page index in general for all columns. Default disabled.
     ///
     /// Page index contains statistics for data pages and can be used to skip pages
-    /// when scanning data in ordered and unordered columns.
+    /// when scanning data in ordered and unordered columns. Note that it does not
+    /// write statistics to the page header once page index is enabled.
     ///
     /// Please check the link below for more details:
     /// https://github.com/apache/parquet-format/blob/master/PageIndex.md
@@ -541,6 +542,8 @@ class PARQUET_EXPORT WriterProperties {
     }
 
     /// Enable writing page index for column specified by `path`. Default disabled.
+    /// Note that it does not write statistics to the page header once page index is
+    /// enabled.
     Builder* enable_write_page_index(const std::string& path) {
       page_index_enabled_[path] = true;
       return this;
--- a/cpp/src/parquet/properties.h
+++ b/cpp/src/parquet/properties.h
@@ -544,8 +544,6 @@ class PARQUET_EXPORT WriterProperties {
     }
 
     /// Enable writing page index for column specified by `path`. Default disabled.
-    /// Note that it does not write statistics to the page header once page index is
-    /// enabled.
     Builder* enable_write_page_index(const std::string& path) {
       page_index_enabled_[path] = true;
       return this;
--- a/cpp/src/parquet/properties.h
+++ b/cpp/src/parquet/properties.h
@@ -524,9 +524,11 @@ class PARQUET_EXPORT WriterProperties {
 
     /// Enable writing page index in general for all columns. Default disabled.
     ///
-    /// Page index contains statistics for data pages and can be used to skip pages
-    /// when scanning data in ordered and unordered columns. Note that it does not
-    /// write statistics to the page header once page index is enabled.
+    /// Writing statistics to the page index disables the old method of writing
+    /// statistics to each data page header.
+    /// The page index makes filtering more efficient than the page header, as
+    /// it gathers all the statistics for a Parquet file in a single place,
+    /// avoiding scattered I/O.
     ///
     /// Please check the link below for more details:
     /// https://github.com/apache/parquet-format/blob/master/PageIndex.md
--- a/docs/source/cpp/parquet.rst
+++ b/docs/source/cpp/parquet.rst
@@ -304,6 +304,8 @@ Statistics are enabled by default for all columns. You can disable statistics fo
 all columns or specific columns using ``disable_statistics`` on the builder.
 There is a ``max_statistics_size`` which limits the maximum number of bytes that
 may be used for min and max values, useful for types like strings or binary blobs.
+If a column has enabled page index using ``enable_write_page_index``, then it does
+not write statistics to the page header because it is duplicated in the ColumnIndex.
 
 There are also Arrow-specific settings that can be configured with
 :class:`parquet::ArrowWriterProperties`:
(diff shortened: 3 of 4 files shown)


cpp/src/parquet/column_writer.cc
```diff
@@ -460,7 +460,11 @@ class SerializedPageWriter : public PageWriter {
         ToThrift(page.definition_level_encoding()));
     data_page_header.__set_repetition_level_encoding(
         ToThrift(page.repetition_level_encoding()));
-    data_page_header.__set_statistics(ToThrift(page.statistics()));
+
+    // Write page statistics only when page index is not enabled.
```
mapleFU, line 464: Seems that this will invalidate page filter?

cpp/src/parquet/column_writer.cc
```diff
@@ -479,7 +483,11 @@ class SerializedPageWriter : public PageWriter {
         page.repetition_levels_byte_length());
 
     data_page_header.__set_is_compressed(page.is_compressed());
-    data_page_header.__set_statistics(ToThrift(page.statistics()));
+
+    // Write page statistics only when page index is not enabled.
```
mapleFU, line 487: what if that Page Index is canceled, like all values is nan?

cpp/src/parquet/column_writer.cc
wgtmac: Yes, in this case it is more efficient to use the column index instead.

cpp/src/parquet/column_writer.cc
wgtmac: In this case, the page stats in the header is also not very useful. Actually in the parquet-mr, it does not write page stats into the header any more.

wjones127 (approved):
I am good with this. Since this is change in behavior, I'll wait for others to chime in before merging.

emkornfield:
Seems OK to me as writing column indices is opt-in?

wgtmac:
> Seems OK to me as writing column indices is opt-in?

Yes, by default page index is disabled. I am not sure if we should turn on it by default someday to align with what parquet-mr does.

wjones127:
> Yes, by default page index is disabled. I am not sure if we should turn on it by default someday to align with what parquet-mr does.

I think we eventually want to turn it on by default. We just release support for it in 12.0.0, but it's not yet used for any data skipping in the Parquet reader or dataset scanner (right?). I think we should add support for leveraging it in the readers in 13.0.0, and then if it seems to be working for users we can turn in on by default in 14.0.0. Does that seem like a reasonable plan?

wgtmac:
> I think we should add support for leveraging it in the readers in 13.0.0

Yes, but I don't think the time frame is sufficient for a complete page skipping implementation. We can support it progressively.

wjones127:
Sure. So I suppose modify my proposal to: if we fully support predicate pushdown for page index in version N, let's set this to default in version N + 1 (assuming there are no unresolved issues we find).

mapleFU (approved):
Personally I'm good with this, but maybe some document is required. I guess maybe a user previously uses page filter. Now when he/she enables page index, the filter would not work

About statistics, though nan will break the page index, and page statistics and make full use of nan-counts. I think this on-going patch helps https://github.com/apache/parquet-format/pull/196

wgtmac:
> I guess maybe a user previously uses page filter. Now when he/she enables page index, the filter would not work

I think the documentation is also missing for stats filtering on the page header. Since page index is an ongoing effort and it is not enabled by default, users will not get confused by this for now.

> About statistics, though nan will break the page index, and page statistics and make full use of nan-counts. I think this on-going patch helps https://github.com/apache/parquet-format/pull/196

Due to the inactivity in that PR, I assume it will take a long time to move forward as we need somehow to reach a consensus from the community.

cpp/src/parquet/column_writer.cc
```diff
@@ -460,7 +460,11 @@ class SerializedPageWriter : public PageWriter {
         ToThrift(page.definition_level_encoding()));
     data_page_header.__set_repetition_level_encoding(
         ToThrift(page.repetition_level_encoding()));
-    data_page_header.__set_statistics(ToThrift(page.statistics()));
+
+    // Write page statistics only when page index is not enabled.
+    if (column_index_builder_ == nullptr) {
+      data_page_header.__set_statistics(ToThrift(page.statistics()));
+    }
```
pitrou, line 467: This only disables writing the statistics, but still computes them, which is suboptimal.

pitrou (changes requested):
See comment below. Also the doc should be updated, and perhaps also the `WriterProperties` docstrings: https://arrow.apache.org/docs/cpp/parquet.html#writer-properties

cpp/src/parquet/column_writer.cc
wgtmac: The stats computation is still required to build page index (a.k.a. the page stats of ColumnIndex).

pitrou: Ok, I see. Also, why do you use the wording "page index"? It's called column index in the spec.

pitrou: Hmm, that's true, my bad.

wgtmac:
Could you please take a look again? @pitrou

cpp/src/parquet/properties.h
```diff
@@ -525,7 +525,8 @@ class PARQUET_EXPORT WriterProperties {
     /// Enable writing page index in general for all columns. Default disabled.
     ///
     /// Page index contains statistics for data pages and can be used to skip pages
-    /// when scanning data in ordered and unordered columns.
+    /// when scanning data in ordered and unordered columns. Note that it does not
```
pitrou, line 528: What is "it" in this context? The sentence isn't very clear.

wgtmac: Here "it" denotes to the parquet writer or column writer.

pitrou: Perhaps:
> Writing statistics to the page index disables the old method of writing statistics to each data page header.
> The page index makes filtering more efficient than the page header, as it gathers all the statistics for a Parquet file in a single place, avoiding scattered I/O.

docs/source/cpp/parquet.rst
```diff
@@ -573,20 +575,14 @@ Miscellaneous
...
 +--------------------------+----------+----------+---------+
-| Bloom Filter             | ✓        | ✓        | \(2)    |
+| Bloom Filter             | ✓        | ✓        | \(1)    |
 +--------------------------+----------+----------+---------+
-| CRC checksums            | ✓        | ✓        | \(3)    |
+| CRC checksums            | ✓        | ✓        |         |
 +--------------------------+----------+----------+---------+
 
-* \(1) Access to the Column and Offset Index structures is provided, but
-  data read APIs do not currently make any use of them.
```
pitrou, line 586: Why remove this note? AFAIU it is still valid (we do not expose any high-level filtering feature).

pitrou:
Did you try to benchmark writing the page index vs. writing statistics in data page headers?
Perhaps in the future we can enable the page index by default?

wgtmac:
> Did you try to benchmark writing the page index vs. writing statistics in data page headers? Perhaps in the future we can enable the page index by default?

I didn't benchmark the writer time because it should spend more time when page index is enabled because:
- It only skips writing stats to the thrift-encoded header but the stats comparison (which is heavy) in the column writer still does the job.
- Page index builder also does more work than just serializing stats. It also collects sorting order from page stats, collects page offsets and so on.

The main goal of skipping writing stats to page header mainly is to reduce the file size as they are duplicated and easier to get from the column index. We have internally enabled page index by default. The benefit brought by page index is promising.

cpp/src/parquet/properties.h
```diff
@@ -541,6 +544,8 @@ class PARQUET_EXPORT WriterProperties {
     }
 
     /// Enable writing page index for column specified by `path`. Default disabled.
+    /// Note that it does not write statistics to the page header once page index is
+    /// enabled.
```
pitrou, line 548: Update this comment as well? Or just removed it as things are explained in more detail above.

wgtmac: Sorry for missing that! Fixed.

pitrou:
> We have internally enabled page index by default. The benefit brought by page index is promising.

Did you measure the benefits (other than file size)?

pitrou (approved):

wgtmac:
Yes, it could achieve 2X~4X acceleration of reading column chunk from cloud object store using page offset index (with I/O coalescing) compared to issue single I/O of a full column chunk (which is currently by default).

pitrou:
Nice, I see. What about the column index, are you using it?

wgtmac:
> Nice, I see. What about the column index, are you using it?

Yes, we did a naive implementation of predicate push down similar to what the parquet-mr does.

fatemehp:
@wgtmac could you elaborate on how you are achieving 2X~4X acceleration of reading the column chunk from the cloud object store you mentioned above? Are you reading the column chunk page by page?

wgtmac:
> @wgtmac could you elaborate on how you are achieving 2X~4X acceleration of reading the column chunk from the cloud object store you mentioned above? Are you reading the column chunk page by page?

In short, collect all offset/length ranges of required pages, then coalesce them into reasonable I/O chunks and issue async reads before reading any page.

PR #35462, opened 2023-05-06
GH-35461: [C++] Add function Seek to BufferOutputStream

### LINKED ISSUE #35461: [C++] Add function Seek to BufferOutputStream

Add function Seek to BufferOutputStream

### Component(s)

C++

jinchengchenghh:
* Closes: #35461

### DIFF
--- a/cpp/src/arrow/io/memory.cc
+++ b/cpp/src/arrow/io/memory.cc
@@ -110,6 +110,21 @@ Status BufferOutputStream::Write(const void* data, int64_t nbytes) {
   return Status::OK();
 }
 
+Status BufferOutputStream::Seek(int64_t position) {
+  if (ARROW_PREDICT_FALSE(!is_open_)) {
+    return Status::IOError("OutputStream is closed");
+  }
+  DCHECK(buffer_);
+  if (position < 0) {
+    return Status::IOError("Seek out of bounds");
+  }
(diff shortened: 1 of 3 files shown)


cpp/src/arrow/io/memory.cc
```diff
@@ -110,6 +110,21 @@ Status BufferOutputStream::Write(const void* data, int64_t nbytes) {
   return Status::OK();
 }
 
+Status BufferOutputStream::Seek(int64_t position) {
```
wgtmac, line 113: I'd prefer creating a separate class for the seekable output stream as its semantics is append-only and downstream applications are using it with that assumption.

jinchengchenghh: Do you have recommended class name for seekable output stream?

wgtmac: Before discussing the name, could you please justify the use case? 

Also cc @wjones127

westonpace: On the read side we have:

InputStream - non-seekable input
ReadableFile - seekable input

So I think we should use `WriteableFile` and not `OutputStream`.  Also note that some filesystems do not support seekable output (e.g. S3).

jinchengchenghh: I optimize my code to use `FixedSizeBufferWriter`, so I don't need this feature now, thanks for your kindly review!

PR #35464, opened 2023-05-06
GH-35125: [C++][Acero] Add a self-defined io-executor in QueryOptions

### LINKED ISSUE #35125: [C++] Specify IO executor in Acero QueryContext

We want to use a dedicated IO executor other than arrow default in scheduling IO task. The QueryContext constructor only set the IOContext's memory pool. Could it provide a way to also specify the executor?

### Component(s)

C++

mapleFU:
### Rationale for this change

Support a custom io-executor in Acero QueryOptions.

Allow user to define a customed io-executor.

### Are these changes tested?

Currently not

User can specify an io-executor

* Closes: #35125


mapleFU:
```c++
Result<std::shared_ptr<Table>> DeclarationToTable(Declaration declaration,
                                                  QueryOptions query_options) {
  if (query_options.custom_cpu_executor != nullptr) {
    return Status::Invalid("Cannot use synchronous methods with a custom CPU executor");
  }
```

By the way, should I notice and add io-executor here? @westonpace

westonpace:
> By the way, should I notice and add io-executor here? @westonpace

No, that's ok.  In a synchronous context we need to specify the CPU executor (because we are going to use a serial executor).  However, it is ok if there is a custom IO executor.

westonpace (commented):
This is ok on its own but the I/O executor isn't being used yet.  We can do that in follow-up PRs.  However, we should eventually switch things to start using this new IO context.  For example:

https://github.com/apache/arrow/blob/3948c426927514ab6d3165255b4717af9446e949/cpp/src/arrow/dataset/scan_node.cc#L64

https://github.com/apache/arrow/blob/3948c426927514ab6d3165255b4717af9446e949/cpp/src/arrow/acero/source_node.cc#LL335C15-L335C15

mapleFU:
Got it, thanks! I'm a bit busy these days, I'll finish it and testing this weekend.

cpp/src/arrow/acero/query_context.cc
```diff
@@ -23,10 +23,17 @@ namespace arrow {
 using arrow::internal::CpuInfo;
 namespace acero {
 
+io::IOContext getIoContext(QueryOptions& opts, ExecContext& exec_context) {
```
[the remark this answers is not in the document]
mapleFU: I think `exec_context` can be make as constant, but `custom_io_executor` should be non-const, so `QueryOptions` cannot have `const`

mapleFU:
@westonpace comment fixed, mind take a look?

cpp/src/arrow/acero/query_context.cc
```diff
@@ -23,10 +23,19 @@ namespace arrow {
 using arrow::internal::CpuInfo;
 namespace acero {
 
+namespace {
+io::IOContext GetIoContext(QueryOptions& opts, const ExecContext& exec_context) {
```
westonpace, line 27:
```suggestion
io::IOContext GetIoContext(const QueryOptions& opts, const ExecContext& exec_context) {
```

westonpace (changes requested):
Sorry, last small nit and I think we are good to go.

cpp/src/arrow/acero/query_context.cc
mapleFU: I think exec_context can be make as constant, but custom_io_executor should be non-const, so QueryOptions cannot have const. Should I use `mutable` for `io::Executor custom_io_executor` or some other methods?

mapleFU:
@westonpace I've fixed the comments :-)

westonpace (approved):
Thank you for adding this.

Issue #31142, opened 2022-02-15
[Dev] Rename default branch to 'main'


asfimport:
Parent issue, see subtasks

**Reporter**: [Neal Richardson](https://issues.apache.org/jira/browse/ARROW-15689) / @nealrichardson
#### Subtasks:
- [X] https://github.com/apache/arrow/issues/31143
- [X] https://github.com/apache/arrow/issues/31144
- [X] https://github.com/apache/arrow/issues/31145
- [X] https://github.com/apache/arrow/issues/31146
- [X] https://github.com/apache/arrow/issues/31147
- [x] https://github.com/apache/arrow/issues/20117
- [x] https://github.com/apache/arrow/issues/31148
- [x] https://github.com/apache/arrow/issues/31412
- [x] https://github.com/apache/arrow/issues/20161
- [x] https://github.com/apache/arrow/issues/32804
- [x] https://github.com/apache/arrow/issues/33005
- [x]  #33215
- [x] #34256
- [x] #34257

<sub>**Note**: *This issue was originally created as [ARROW-15689](https://issues.apache.org/jira/browse/ARROW-15689). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

jorisvandenbossche:
I opened https://issues.apache.org/jira/browse/INFRA-24214 to ask for the rename, and https://github.com/apache/arrow/pull/34218 to rename all occurences in our own repo (but short term those urls also get redirected by github)

assignUser:
I have added a few follow ups. Tagging @kevingurney and @lafiona so they are aware where this is tracked and how it is moving forward!

kevingurney:
I've taken ownership of #31412 !

jorisvandenbossche:
That reminds me that we also at some point need to ask INFRA for switching the branch name on the arrow-site repo as well. Are we ready to do that? Or are there still tasks to be done in preparation?

assignUser:
It looks like the site workflows were already prepared to work with main instead of master.  I don't see anything else depending on the default branch there but I am not super familiar with that repo so not 100% certain.

kevingurney:
@jorisvandenbossche and @assignUser - I'll verify that there aren't any remaining dependencies in the arrow-site workflows as part of my investigation of #31412.

kevingurney:
Following up on this discussion:

[...]

PR #35467, opened 2023-05-06
MINOR: [Go][CI][Benchmarking] Fix 'Run Benchmarks' steps of 'Go' CI check (broken on `main`, fails with `TypeError`)

jgehrcke:
### Rationale for this change

As part of https://github.com/apache/arrow/issues/31142 we did not rename `master` to `main` in [`ci/scripts/go_bench_adapt.py`](https://github.com/apache/arrow/blob/b43f4cd4ce28e2ddb7bad89081e141f381b09d5e/ci/scripts/go_bench_adapt.py#L34).

This was not immediately noticed in CI because we accidentally switched to a code branch intended for local dev (cf. the `else:` block [here](https://github.com/apache/arrow/blob/b43f4cd4ce28e2ddb7bad89081e141f381b09d5e/ci/scripts/go_bench_adapt.py#L41)), also see [this commit message](https://github.com/apache/arrow/commit/885e050b6b2c367d266793cba8d601c805a4e17d).

Since then we were unintentionally using a feature in conbench/benchadapt to read commit information from the local file system, instead of reading it from CI-focused environment variables.

That even worked for a while.

However, last week, that feature was [removed from benchadapt](https://github.com/conbench/conbench/pull/1135#issuecomment-1537110851) (certainly assuming that this is _not_ used in CI anywhere).

That is also when `BenchmarkResult.github` started to potentially become `None`. Which `mypy` knows about, see:
image

Well, but further down in code, `parsed.github['commit']` was used w/o protection against this `None` case, and that is how we started to see CI run failures with
```
...
  File "/home/runner/work/arrow/arrow/ci/scripts/go_bench_adapt.py", line 94, in _transform_results
    parsed.run_name = f"{parsed.run_reason}: {parsed.github['commit']}"
TypeError: 'NoneType' object is not subscriptable
```
See e.g.  https://github.com/apache/arrow/actions/runs/4859969743/jobs/8742255231 -- thank you @zeroshade for reporting this.

- Clarification in code via comments
- More expressive variable naming, type hints
- Additional assert statement guarding against running 'local dev' code in CI
- Minor formatting changes via black in VSCode (that's OK I hope?)

A tiny feature/convenience for local dev: the env var `CONBENCH_CUSTOM_RUN_REASON` can be used to inject custom info.
```python
    # Reflect 'local dev' scenario in run_reason. Allow user to (optionally)
    # inject a custom piece of information into the run reason here, from
    # environment.
    run_reason = "localdev"
    custom_reason_suffix = os.getenv("CONBENCH_CUSTOM_RUN_REASON")
    if custom_reason_suffix is not None:
        run_reason += f" {custom_reason_suffix.strip()}"
```

### Are these changes tested?

No. Relied on mypy/pylint here.

### DIFF
--- a/ci/scripts/go_bench_adapt.py
+++ b/ci/scripts/go_bench_adapt.py
@@ -38,11 +38,8 @@ SCRIPTS_PATH = ARROW_ROOT / "ci" / "scripts"
 github_commit_info: Optional[Dict] = None
 
 if os.environ.get("CONBENCH_REF") == "main":
-    # Assume GitHub Actions CI. The environment
-    # variable lookups below are expected to fail when
-    # not running in GitHub Actions. See
-    # https://github.com/conbench/conbench/blob/7c4968e631ecdc064559c86a1174a1353713b700/benchadapt/python/benchadapt/result.py#L66
-    # for a specification of this `github` argument.
+    # Assume GitHub Actions CI. The environment variable lookups below are
+    # expected to fail when not running in GitHub Actions.
     github_commit_info = {
         "repository": os.environ["GITHUB_REPOSITORY"],
         "commit": os.environ["GITHUB_SHA"],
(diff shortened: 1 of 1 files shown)


zeroshade (approved):
LGTM thanks for the quick analysis and fix!

PR #35471, opened 2023-05-07
GH-35176: [C++] Add support for disabling threading for emscripten

### LINKED ISSUE #35176: [C++] Add support for disablging threading for emscripten

I've built most of arrow (pyarrow and dependencies) for emscripten. It would be good to have a way to disable threading, as a lot of emscripten use is in browsers where threading may not be available.

At the moment I just put in dummy pthreads, which means some functionality in e.g.datasets fails because it assumes threading is available.

### Component(s)

C++

joemarshall:
As previously discussed in #35176 this is a patch that adds an option `ARROW_DISABLE_THREADING`. When it is turned on, arrow threadpool and serial executors don't spawn threads, and instead run tasks in the main thread when futures are waited for.

It doesn't mess with threading in projects included as dependencies, e.g. multithreaded malloc implementations because if you're building for a non threaded environment, you can't use those anyway.

Basically where this is at is that it runs the test suite okay, and I think should work well enough to be a backend for pandas on emscripten/pyodide, so subject to reviews etc. it is worth merging (and would be jolly convenient for me if it was).

What this means is:
1) It is possible to use arrow in non-threaded emscripten/webassembly environments (with some build patches specific to emscripten which I'll put in once this is in)
2) Most of arrow just works, albeit slower in parts.

Things that don't work and probably won't:
1) Server stuff that relies on threads. Not a massive problem I think because environments with threading restrictions are currently typically also restricted from making servers anyway (i.e. they are web browsers)
2) Anything that relies on actually doing two things at once (for obvious reasons)

Things that don't work yet and could be fixed in future:
1) use of asynchronous file/network APIs in emscripten which would mean I/O could work efficiently in one thread.
2) asofjoin - right now the implementation relies on std::thread - it needs refactoring to work with threadpool like everything else in arrow, but I'm not sure I am expert enough in the codebase to do it well.
* Closes: #35176


kou:
In general, I think that `ARROW_ENABLE_*` is better than `ARROW_DISABLE_*` for readability.

joemarshall:
Yeah I was torn on that. I guess I thought disable implies that threading is a default and that this is something for special cases? I'm easy though.

kou:
There are many default `ON` options. So `ARROW_ENABLE_*=ON` doesn't imply that the feature is disabled by default. 
See also: https://github.com/apache/arrow/blob/main/cpp/cmake_modules/DefineOptions.cmake

joemarshall:
Okay, I'll flip that

joemarshall:
That's flipped, it's now ARROW_ENABLE_THREADING, and an option is added to defineoptions.cmake

cmdline-working.txt
```diff
@@ -0,0 +1,10 @@
+export CMAKE_TOOLCHAIN_FILE=/home/pszjm2/arrow/Emscripten.cmake
```
pitrou, line 1: This is probably not meant to be part of this PR, is it?

cpp/cmake_modules/DefineOptions.cmake
```diff
@@ -210,6 +210,10 @@ takes precedence over ccache if a storage backend is configured" ON)
 
   define_option(ARROW_WITH_MUSL "Whether the system libc is musl or not" OFF)
 
+  define_option(ARROW_ENABLE_THREADING "Enable threading in arrow core. "
+                ON)
```
pitrou, line 214:
```suggestion
  define_option(ARROW_ENABLE_THREADING "Enable threading in Arrow core. "
                ON)
```

cpp/src/arrow/acero/CMakeLists.txt
```diff
@@ -236,7 +246,9 @@ if(ARROW_BUILD_BENCHMARKS)
     target_link_libraries(arrow-acero-expression-benchmark PUBLIC arrow_acero_static)
     target_link_libraries(arrow-acero-filter-benchmark PUBLIC arrow_acero_static)
     target_link_libraries(arrow-acero-project-benchmark PUBLIC arrow_acero_static)
-    target_link_libraries(arrow-acero-asof-join-benchmark PUBLIC arrow_acero_static)
+    if(!gig)
```
pitrou, line 249: Hmm, what is this?

cpp/src/arrow/acero/asof_join_node.cc
```diff
@@ -1520,6 +1520,9 @@ class AsofJoinNode : public ExecNode {
       // Plan has already aborted.  Do not start process thread
       return Status::OK();
     }
+    #ifndef ARROW_ENABLE_THREADING
+      return Status::ExecutionError("ASOF join requires threading enabled");
```
pitrou, line 1524: Probably move this at the beginning of this method? Also use a `#else` clause around the regular implementation.

And I would probably make this `Status::NotImplemented`.

cpp/src/arrow/acero/asof_join_node_test.cc
```diff
@@ -1380,7 +1380,7 @@ T GetEnvValue(const std::string& var, T default_value) {
 }  // namespace
 
 TEST(AsofJoinTest, BackpressureWithBatchesGen) {
-  int num_batches = GetEnvValue("ARROW_BACKPRESSURE_DEMO_NUM_BATCHES", 10);
+  int num_batches = GetEnvValue("ARROW_BACKPRESSURE_DEMO_NUM_BATCHES", 2);
```
pitrou, line 1383: Not sure we want to modify this parameter? cc @westonpace

cpp/src/arrow/acero/bloom_filter.cc
```diff
@@ -426,6 +426,9 @@ void BloomFilterBuilder_Parallel::CleanUp() {
 
 std::unique_ptr<BloomFilterBuilder> BloomFilterBuilder::Make(
     BloomFilterBuildStrategy strategy) {
+#ifndef ARROW_ENABLE_THREADING
+  strategy=BloomFilterBuildStrategy::SINGLE_THREADED;
```
pitrou, line 430:
```suggestion
  strategy = BloomFilterBuildStrategy::SINGLE_THREADED;
```

cpp/src/arrow/acero/bloom_filter_test.cc
```diff
@@ -472,8 +472,10 @@ TEST(BloomFilter, Basic) {
 
   std::vector<BloomFilterBuildStrategy> strategy;
   strategy.push_back(BloomFilterBuildStrategy::SINGLE_THREADED);
-#ifndef ARROW_VALGRIND
+#ifdef ARROW_ENABLE_THREADING
+#ifndef ARROW_VALGRIND 
   strategy.push_back(BloomFilterBuildStrategy::PARALLEL);
+#endif
 #endif
```
pitrou, line 479: Or, more tersely
```suggestion
#if defined(ARROW_ENABLE_THREADING) && !defined(ARROW_VALGRIND)
  strategy.push_back(BloomFilterBuildStrategy::PARALLEL);
#endif
```

cpp/src/arrow/acero/task_util.cc
```diff
@@ -328,10 +328,13 @@ Status TaskSchedulerImpl::ScheduleMore(size_t thread_id, int num_tasks_finished)
   }
 
   ARROW_DCHECK(register_finished_);
-
+#ifndef ARROW_ENABLE_THREADING
+    return ExecuteMore(thread_id, 1, true);
+#else
   if (use_sync_execution_) {
```
pitrou, line 334: Shouldn't instead `use_sync_execution_` to be always true when threading is disabled? @westonpace

cpp/src/arrow/util/config.h.cmake
```diff
@@ -57,5 +57,6 @@
 #cmakedefine ARROW_WITH_MUSL
 #cmakedefine ARROW_WITH_OPENTELEMETRY
 #cmakedefine ARROW_WITH_UCX
+#cmakedefine ARROW_ENABLE_THREADING
```
pitrou, line 60: Let's try to keep this alphabetically-ordered?

.github/workflows/cpp.yml
```diff
@@ -57,11 +57,19 @@ jobs:
...
       matrix:
         include:
+          - arch: amd64
+            clang-tools: "14"
+            image: ubuntu-cpp
+            llvm: "14"
+            runs-on: ubuntu-latest
+            title: AMD64 ubuntu-cpp no threads
+            ubuntu: "22.04"
+            enable_threading: "OFF"
```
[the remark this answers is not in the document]
kou: The following change will work:

```diff
diff --git a/dev/tasks/tasks.yml b/dev/tasks/tasks.yml
index 6d07c65b4..103c11acc 100644
--- a/dev/tasks/tasks.yml
+++ b/dev/tasks/tasks.yml
@@ -1162,6 +1162,15 @@ tasks:
       image: ubuntu-cpp
 {% endfor %}
 
+  test-ubuntu-22.04-cpp-no-threading:
+    ci: github
+    template: docker-tests/github.linux.yml
+    params:
+      env:
+        UBUNTU: 22.04
+      flags: "-e ARROW_ENABLE_THREADING=OFF"
+      image: ubuntu-cpp
+
   test-ubuntu-20.04-cpp-thread-sanitizer:
     ci: github
     template: docker-tests/github.linux.yml
```

You can run the job by commenting `@github-actions crossbow submit test-ubuntu-22.04-cpp-no-threading` in a PR.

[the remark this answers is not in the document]
assignUser: Agreed! A nightly would be much more appropriate.

cpp/src/arrow/acero/asof_join_node_test.cc
joemarshall: oops, sorry left in from testing.  Reverting.

.github/workflows/cpp.yml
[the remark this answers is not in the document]
joemarshall: okay, reverted and added nightly

cpp/src/arrow/acero/task_util.cc
joemarshall: have done

joemarshall:
@github-actions crossbow submit test-ubuntu-22.04-cpp-no-threading

pitrou:
@github-actions crossbow submit test-ubuntu-22.04-cpp-no-threading

kou:
@joemarshall Could you create a feature branch for this instead of using the `main` branch?
`@github-actions crossbow submit test-ubuntu-22.04-cpp-no-threading` doesn't work with the `main` branch.

kou:
> A feature branch in my repo?

Yes.

joemarshall:
This is now in #35672 which is the same but from a differently named branch because that means that the CI stuff (crossbow submit commands) actually work, whereas they don't work if the name of the source branch is main.

PR #35476, opened 2023-05-08
GH-35275: [Java] Ensure VectorSchemaRoot slice returns a new root

### LINKED ISSUE #35275: [Java] Why does VectorSchemaRoot::slice return reference to same root when slice index is [0, rowCount] ?

Hi Arrow community, 

In the VectorSchemaRoot::slice implementation it checks if the slice covers the entirety of the root and returns the reference to the same root if so.

```
    if (index == 0 && length == rowCount) {
      return this;
    }
```

Looks like this is done as an optimization to prevent creating another root + transferring to it.

Doesn't this optimization make `slice` inconsistent? Sometimes you get back a _new_  autocloseable resource that has to be separately closed and sometimes you get back the same root. 

So if you want to calculate something based on a set of different slices of your root then you have to manually check whether the current slice you are requesting isn't the whole root to avoid closing the original root. 

Since `splitAndTransfer` do no copying of the actual data, is this optimization worth this tradeoff? 

### Component(s)

Java

ParthChonkar:
### Rationale for this change
The optimization in the `slice` function for VectorSchemaRoot that returns the reference to the original root when the slice covers the entire root should be removed because this is an inconsistent interface (sometimes you get a totally new autocloseable and sometimes you get the original one). 

Also - in the [this documentation page](https://arrow.apache.org/docs/java/vector_schema_root.html#vectorschemaroot) it explicitly says that "A **NEW** VectorSchemaRoot can be sliced from an existing root without copying data:" so any time you call `slice`, the slice root should be a new one (even if it's the entire root). 

The test for slice has also been changed to have more coverage of the slice function (not just slicing from the start index).  

Remove self reference return in `VectorSchemaRoot#split` and expand test coverage for the function.

### Are these changes tested?
Yes

Yes 

* Closes: apache/arrow#35275
* GitHub Issue: apache/arrow#35275
* GitHub Issue: #35275

### DIFF
--- a/java/vector/src/main/java/org/apache/arrow/vector/VectorSchemaRoot.java
+++ b/java/vector/src/main/java/org/apache/arrow/vector/VectorSchemaRoot.java
@@ -327,10 +327,6 @@ public class VectorSchemaRoot implements AutoCloseable {
     Preconditions.checkArgument(index + length <= rowCount,
         "index + length should <= rowCount");
 
-    if (index == 0 && length == rowCount) {
-      return this;
-    }
-
     List<FieldVector> sliceVectors = fieldVectors.stream().map(v -> {
       TransferPair transferPair = v.getTransferPair(v.getAllocator());
       transferPair.splitAndTransfer(index, length);
--- a/java/vector/src/test/java/org/apache/arrow/vector/TestVectorSchemaRoot.java
+++ b/java/vector/src/test/java/org/apache/arrow/vector/TestVectorSchemaRoot.java
@@ -195,29 +195,33 @@ public class TestVectorSchemaRoot {
   public void testSlice() {
     try (final IntVector intVector = new IntVector("intVector", allocator);
          final Float4Vector float4Vector = new Float4Vector("float4Vector", allocator)) {
-      intVector.setValueCount(10);
-      float4Vector.setValueCount(10);
-      for (int i = 0; i < 10; i++) {
+      final int numRows = 10;
+      intVector.setValueCount(numRows);
+      float4Vector.setValueCount(numRows);
+      for (int i = 0; i < numRows; i++) {
         intVector.setSafe(i, i);
         float4Vector.setSafe(i, i + 0.1f);
       }
+
       final VectorSchemaRoot original = new VectorSchemaRoot(Arrays.asList(intVector, float4Vector));
 
-      VectorSchemaRoot slice1 = original.slice(0, original.getRowCount());
-      assertEquals(original, slice1);
-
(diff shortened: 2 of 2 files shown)


java/vector/src/test/java/org/apache/arrow/vector/TestVectorSchemaRoot.java
```diff
@@ -195,29 +195,33 @@ public void testRemoveVector() {
...
+            // validate data
+            final IntVector childIntVector = (IntVector) slice.getFieldVectors().get(0);
+            final Float4Vector childFloatVector = (Float4Vector) slice.getFieldVectors().get(1);
+            for (int i = 0; i < sliceLength; i++) {
+              final int originalIndex = i + sliceIndex;
+              assertEquals(originalIndex, childIntVector.get(i));
+              assertEquals(originalIndex + 0.1f, childFloatVector.get(i), 0);
+            }
+          }
+        }
```
lidavidm, line 221: Is there anything testing that the slice is actually a copy of the original and not a view?

lidavidm (commented):
Thanks. I think this makes sense.

This is technically a breaking change, since it's possible (but unlikely) that code may have relied on the sharing before, so I updated the PR description.

java/vector/src/main/java/org/apache/arrow/vector/VectorSchemaRoot.java
```diff
@@ -327,10 +327,6 @@ public VectorSchemaRoot slice(int index, int length) {
     Preconditions.checkArgument(index + length <= rowCount,
         "index + length should <= rowCount");
 
-    if (index == 0 && length == rowCount) {
-      return this;
```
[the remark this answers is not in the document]
ParthChonkar: For each vector `splitAndTransfer` will derive a new ArrowBuffer (unlike `transfer`) for the vector's underlying buffer(s) so it won't invalidate the original root. But both will point to the same memory location.

java/vector/src/test/java/org/apache/arrow/vector/TestVectorSchemaRoot.java
ParthChonkar: I'm not sure I understand 😅 - slice should be no-copy right?

lidavidm:
That test has been flaky for a long time, but nobody has been able to reproduce it on Windows or really spend the time to investigate it

davisusanibar:
> @davisusanibar can we just file a PR to disable it only for Windows?

Issues related: https://github.com/apache/arrow/issues/33237 / https://github.com/apache/arrow/issues/35488

PR #35478, opened 2023-05-08
MINOR: [CI][R] Nightly R upload fails due to wrong branch

assignUser:
The [R nightly upload](https://github.com/apache/arrow/actions/runs/4915965925/jobs/8779129556) has failed since we renamed the default branch to main in crossbow due to the hardcoded branch name.

### DIFF
--- a/.github/workflows/r_nightly.yml
+++ b/.github/workflows/r_nightly.yml
@@ -58,7 +58,7 @@ jobs:
           fetch-depth: 0
           path: crossbow
           repository: ursacomputing/crossbow
-          ref: master
+          ref: main
       - name: Set up Python
         uses: actions/setup-python@v4
         with:


nealrichardson:
Thanks. Why did we not see this was failing? Is there anything we can do to improve our monitoring?

raulcd:
We probably could add a job that validates that the last published nightlies (that could be good for Java, R, Python) are not older than 48 hours. It might be redundant on some cases but might potentially raise these type of issues sooner.

assignUser:
I agree visibility is pretty bad... Usually someone will get an email for failing builds (not sure who is for arrow?) maybe we can get those forwarded to the builds@ mailing list, though that might get spammy fast with all the fails on main we tend to have..

thisisnic (approved):
Thanks for fixing this!

raulcd:
I've created the issue linked above to track this improvement

PR #35479, opened 2023-05-08
GH-35411: [MATLAB] Create a templated C++ Proxy Class for Numeric Arrays

### LINKED ISSUE #35411: [MATLAB] Create a templated C++ Proxy Class for Numeric Arrays

Create a templated C++ Proxy Class for Numeric Arrays. This is a followup to the pull request #34563 in which we added basic support for the Float64 type. We want to generalize this code to work for  the rest of the numeric types, i.e., float32, [u]int[x][y]. 

### Component(s)

MATLAB

sgilmore10:
### Rationale for this change

This pull request is a followup to #34563. To facilitate the implementation of future array types, we would like to first create a C++ template class for numeric arrays. We also want to start adding basic tests for the array functionality in the MATLAB interface. 

1. Added a C++ template Class called `NumericArray` templated on `CType`.
2. Re-implemented the `Float64Array` C++ proxy class in terms of the new template class, i.e. `NumericArray<double>`.
3. Added a method called `double()` on the MATLAB Float64Array class to convert the arrow.[Type]Array to a MATLAB `double` array.
4. Added basic tests for round-tripping float64 arrays.
5. Created a base C++ proxy `Array` class that all proxy array classes will inherit from.
6. Renamed `Print()` to `ToString()` and made it return a string instead of printing to the screen.

### Are these changes tested?

Yes, we added automated test cases to the test class `tFloat64Array.m`. In addition, we manually qualified these changes on macOS.

Yes, the `Print()` method is no longer public and there is now a method called `double()` on `arrow.array.Float64Array`. 

Included below is a simple example of using the `double()` method:

```matlab
>> arrowArray = arrow.array.Float64Array([1, 2, 3])            

arrowArray = 

[
  1,
  2,
  3
]

>> matlabArray = double(arrowArray)

matlabArray =

     1
     2
     3

>> class(arrowArray)

ans =

    'arrow.array.Float64Array'

>> class(matlabArray)

ans =

    'double'
```

### Future Directions

1. Support the rest of the numeric types.
2. Add an abstract MATLAB base class called `arrow.array.Array`.
3. Continue building out the methods (e.g. `length()`)
4.  Support `null` values (validity bitmap).
5. Handle converting non-ascii characters from `UTF-8` to `UTF-16`.
6. Handle errors in the C++ layer. 

* Closes: #35411


matlab/src/cpp/arrow/matlab/array/proxy/numeric_array.h
```diff
@@ -0,0 +1,93 @@
...
+
+            // Get the mxArray from constructor arguments
+            const ::matlab::data::TypedArray<CType> numeric_mda = constructor_arguments[0];
+            const ::matlab::data::TypedArray<bool> make_copy = constructor_arguments[1];
+
+            // Get raw pointer of mxArray
+            auto it(numeric_mda.cbegin());
+            auto dt = it.operator->();
+
+            const bool make_deep_copy = make_copy[0];
```
kou, line 46:
```suggestion
            const auto make_deep_copy = make_copy[0];
```

matlab/src/cpp/arrow/matlab/array/proxy/numeric_array.h
```diff
@@ -0,0 +1,93 @@
...
+                    if (maybe_array.ok()) {
+                        array = *maybe_array;
+                    }
+                }
+            } else {
+                // Pass pointer to Arrow array constructor that takes a buffer
+                // Do not make a copy when creating arrow::Buffer
+                std::shared_ptr<arrow::Buffer> buffer(
+                                                      new arrow::Buffer(reinterpret_cast<const uint8_t*>(dt),
+                                                                        sizeof(CType) * numeric_mda.getNumberOfElements()));
```
kou, line 64:
```suggestion
                auto buffer = std::make_shared<arrow::Buffer>(reinterpret_cast<const uint8_t*>(dt),
                                                              sizeof(CType) * numeric_mda.getNumberOfElements());
```

matlab/src/cpp/arrow/matlab/array/proxy/numeric_array.h
```diff
@@ -0,0 +1,93 @@
...
+                
+                // Construct arrow::NumericArray specialization using arrow::Buffer.
+                // pass in nulls information...we could compute and provide the number of nulls here too
+                std::shared_ptr<arrow::Array> array_wrapper(
+                                                            new arrow::NumericArray<ArrowType>(numeric_mda.getNumberOfElements(), buffer,
+                                                                                               nullptr, // TODO: fill validity bitmap with data
+                                                                                               -1));
```
kou, line 71: How about creating an `ArrayData` https://github.com/apache/arrow/blob/main/cpp/src/arrow/array/data.h and then creating an `Array` by `MakeArray()` https://github.com/apache/arrow/blob/main/cpp/src/arrow/array/util.h#L34-L38 ?

We may be able to share the same implementation with other array types with this approach.

matlab/src/cpp/arrow/matlab/array/proxy/numeric_array.h
```diff
@@ -0,0 +1,93 @@
...
+                array = array_wrapper;
+            }
+        }
+
+    protected:
+        void ToMatlab(libmexclass::proxy::method::Context& context) override {
+            using ArrowArrayType = typename arrow::CTypeTraits<CType>::ArrayType;
+
+            const size_t num_elements = static_cast<size_t>(array->length());
```
kou, line 80:
```suggestion
            const auto num_elements = static_cast<size_t>(array->length());
```

matlab/src/matlab/+arrow/+array/Float64Array.m
```diff
@@ -20,25 +20,38 @@
...
     end
 
     methods (Access=protected)
         function displayScalarObject(obj)
-            obj.Print();
+            disp(obj.ToString());
         end
     end
 
+    methods (Access = private)
```
kou, line 52:
```suggestion
    methods (Access=private)
```

matlab/test/arrow/array/tFloat64Array.m
```diff
@@ -24,18 +28,77 @@ function verifyOnMatlabPath(testCase)
...
+        function Basic(testCase, MakeDeepCopy)
+            A = arrow.array.Float64Array([1, 2, 3], DeepCopy=MakeDeepCopy);
             className = string(class(A));
             testCase.verifyEqual(className, "arrow.array.Float64Array");
         end
+
+        function ShallowCopy(testCase)
+        % By default, Float64Array does not create a deep copy on
+        % construction when constructed from a MATLAB array. Instead,
+        % it stores a shallow copy of the array keep the memory alive.
```
kou, line 41:
```suggestion
            % By default, Float64Array does not create a deep copy on
            % construction when constructed from a MATLAB array. Instead,
            % it stores a shallow copy of the array keep the memory alive.
```

matlab/test/arrow/array/tFloat64Array.m
```diff
@@ -24,18 +28,77 @@ function verifyOnMatlabPath(testCase)
...
+            testCase.verifyEqual(double(A), [1 2 3]');
+
+            A = arrow.array.Float64Array([1, 2, 3], DeepCopy=false);
+            testCase.verifyEqual(A.MatlabArray, [1 2 3]);
+            testCase.verifyEqual(double(A), [1 2 3]');
+        end
+
+        function DeepCopy(testCase)
+        % Verify Float64Array does not store shallow copy of the MATLAB
+        % array if DeepCopy=true was supplied.
```
kou, line 53:
```suggestion
            % Verify Float64Array does not store shallow copy of the MATLAB
            % array if DeepCopy=true was supplied.
```

sgilmore10:
Thanks for the feedback @kou! Let us know if you have any other comments.

kou (approved):
+1

PR #35481, opened 2023-05-08
GH-35482: [Go] Append nulls to values in `array.FixedSizeListBuilder.AppendNull`

### LINKED ISSUE #35482: Don't require manually appending nulls to values in `array.FixedSizeListBuilder.AppendNull`

Currently there's a gotcha to use `array.FixedSizeListBuilder.AppendNull`: you have to manually append null values to the underlying list.

This differs from `array.FixedSizeListBuilder.AppendEmptyValue` that handles it by itself.

### Component(s)

Go

candiduslynx:
### Rationale for this change

Currently there's a gotcha to use `array.FixedSizeListBuilder.AppendNull`: you have to manually append null values to the underlying list.

This changes this behavior to match the one of `array.FixedSizeListBuilder.AppendEmptyValue` that handles it by itself.

Add `AppendNull` call to the values builder in `array.FixedSizeListBuilder.AppendNull`.

### Are these changes tested?

Yes, the `Example_fixedSizeListArray` was modified to express the changed behavior.

Added `AppendNull` call to the values builder in `array.FixedSizeListBuilder.AppendNull`.
Previously, manually adding these calls was required.
After this change a simple `array.FixedSizeListBuilder.AppendNull` is enough to maintain the correct indices.

* Closes: #35482

### DIFF
--- a/go/arrow/array/fixed_size_list.go
+++ b/go/arrow/array/fixed_size_list.go
@@ -207,6 +207,9 @@ func (b *FixedSizeListBuilder) Append(v bool) {
 func (b *FixedSizeListBuilder) AppendNull() {
 	b.Reserve(1)
 	b.unsafeAppendBoolToBitmap(false)
+	for i := int32(0); i < b.n; i++ {
+		b.values.AppendNull()
+	}
 }
 
 func (b *FixedSizeListBuilder) AppendEmptyValue() {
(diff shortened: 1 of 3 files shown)


go/arrow/array/fixed_size_list.go
```diff
@@ -207,6 +207,9 @@ func (b *FixedSizeListBuilder) Append(v bool) {
 func (b *FixedSizeListBuilder) AppendNull() {
```
zeroshade, line 207: Can you just add a docstring comment here that mentions this will add the null values to the underlying value builder? Since this is a breaking change i want to make sure the documentation updates accordingly to ensure people can understand why their code broke. (if they don't read the changelog when v13 gets released haha).

zeroshade (approved):
Just one request for a clarified docstring. otherwise this is good.

Issue #15032, opened 2022-12-19
pyarrow won't cast date32 to date64


Thomas-Hirsch:
### Describe the bug, including details regarding any error messages, version, and platform.

I'm trying to build a tool to confirm that a given parquet file conforms to an expected schema with [our metadata format](https://github.com/moj-analytical-services/mojap-metadata). There appears to be a bug with how pyarrow converts dates, however.

<details>
<summary>I have a dummy dataset, called test.csv (expand to view)</summary>

|my_int|animal |my_email                 |my_datetime        |my_date   |
|------|-------|-------------------------|-------------------|----------|
|16    |NA     |omar65@waller.info       |2013-01-14 15:54:20|1993-09-27|
|13    |cat    |nataliechavez@hotmail.com|2006-12-16 01:44:21|1989-04-08|
|13    |dog    |stephanie12@leblanc.com  |1972-06-19 11:22:59|2006-10-01|
|18    |NA     |cameronscott@bradshaw.com|2009-07-11 00:45:28|1992-07-24|
|10    |fish   |rossrebecca@gould.net    |1970-09-11 22:26:13|2011-11-30|
|13    |cat    |escobarryan@gmail.com    |2018-09-18 16:18:37|1979-01-22|
|16    |dog    |shelly51@mitchell.com    |2002-08-26 10:25:03|1981-03-27|
|11    |dog    |kwilliams@hotmail.com    |1989-02-26 16:33:37|2012-06-19|
|19    |chicken|lisacharles@kennedy.com  |1974-07-25 04:41:22|2003-11-06|
|11    |chicken|olsonchelsea@gray.org    |1992-04-03 09:47:30|1972-07-06|
</details>

I have a user-generated schema that resolves to this:
```
my_int: int64
animal: string
my_email: string
my_datetime: timestamp[s]
my_date: date64[ms]
```

However, the last column seems to be read by pyarrow as `date32[day]`, and won't cast otherwise:
```python
import pyarrow as pa
import pyarrow.parquet as pq

table = pa.csv.read_csv("test.csv")
table.cast(my_custom_schema)
pq.write_table(table, "test.parquet")
table_arrow_schema = pq.read_schema("tests/data/parquet_validator/table1.parquet")
table_arrow_schema
```
gives
```
my_int: int64
animal: string
my_email: string
my_datetime: timestamp[ms]
my_date: date32[day]
```

### Component(s)

Parquet, Python

wjones127:
[...]

PR #35494, opened 2023-05-08
Fix underlying type for DATE64

sam-lunt:
When converting a column of type DATE64 to parquet, the underlying type is INT32 instead of INT64. This updates the type to be the correct width.

### DIFF
--- a/cpp/src/parquet/arrow/schema.cc
+++ b/cpp/src/parquet/arrow/schema.cc
@@ -372,7 +372,7 @@ Status FieldToNode(const std::string& name, const std::shared_ptr<Field>& field,
       logical_type = LogicalType::Date();
       break;
     case ArrowTypeId::DATE64:
-      type = ParquetType::INT32;
+      type = ParquetType::INT64;
       logical_type = LogicalType::Date();
       break;
     case ArrowTypeId::TIMESTAMP:


wjones127:
I don't think this is quite right. There isn't a 1-1 correspondence between Arrow's physical types and Parquet's physical types. Parquet always stores dates as int32: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#date. We map to Arrow's Date64 if that's what the original Arrow type was.

sam-lunt:
I see that the unit tests failed, so I'd be inclined to agree that this change isn't quite right, haha.

That said, I do think there is an issue with the current implementation, since `date64` doesn't survive a round trip from Arrow to Parquet and back:
``` py
import datetime
import sys
import pandas as pd
import pyarrow as pa
import pyarrow.parquet as paq

print(f"{sys.version=}")
print(f"{pa.__version__=}")
print()

schema = pa.schema(
    [
        ("date32", pa.date32()),
        ("date64", pa.date64()),
    ]
)

print("schema:", schema, "", sep="\n")

df = pd.DataFrame(
    {
        "date32": datetime.date(2020, 3, 15),
        "date64": datetime.date(2020, 3, 15),
    },
    index=pd.RangeIndex(1),
)

rb = pa.RecordBatch.from_pandas(df, schema=schema)
print("rb.schema:", rb.schema, "", sep="\n")

w = paq.ParquetWriter("test.parquet", version="2.6", schema=schema)
w.write(rb)
w.close()

f = paq.ParquetFile("test.parquet")
t = f.read()
print("t.schema:", t.schema, sep="\n")
```

On my machine (running x86_64 Linux), I see the following output:
```
sys.version='3.11.3 (main, Apr  5 2023, 15:52:25) [GCC 12.2.1 20230201]'
pa.__version__='10.0.1'

schema:
date32: date32[day]
date64: date64[ms]

rb.schema:
date32: date32[day]
date64: date64[ms]
-- schema metadata --
pandas: '{"index_columns": [], "column_indexes": [{"name": null, "field_n' + 412

t.schema:
date32: date32[day]
date64: date32[day]
```

wjones127:
> That said, I do think there is an issue with the current implementation, since date64 doesn't survive a round trip from Arrow to Parquet and back:

Ah that makes sense. I think this is related to https://github.com/apache/arrow/issues/15032

We should probably improve it so that all types roundtrip.

I'm closing this PR, but if you are interested in making a PR to make more types round trip I would review it.

PR #35496, opened 2023-05-08
GH-33856: [C#] Implement C Data Interface for C#

### LINKED ISSUE #33856: [C#] Implement C Data Interface for C#

### LINKED ISSUE #33857: [C#] Implement C Stream Interface in C#

Follow up to #33856.

### Component(s)

C#

CurtHagenlocher:
### Rationale for this change

This continues implementing the C Data Interface for C# with integration for `ArrowArray`, `RecordBatch` and streams.

- Adds classes `CArrowArray` and `CArrowStream` to represent the C API structures.
- Adds interface `IArrowArrayStream` to represent an array stream or record batch reader.
- Adds classes `CArrowArrayImporter`, `CArrowArrayExporter`, `CArrowArrayStreamImporter` and `CArrowArrayExporter` to marshal between C# and C representations.
- Augments the native memory representation to support (reasonably safe) ownership of memory by external code.

### Are these changes tested?

Yes. Testing is largely done via the Python C API interface.

Yes, this adds new user-facing APIs to import and export C# structures using the C API.

The default time unit for Time64Type was previously milliseconds. This does not appear to be valid, so it has been changed to nanoseconds.

- Closes: #33856
- Closes: #33857
* Closes: #33856


csharp/src/Apache.Arrow/Ipc/IArrowArrayStream.cs
```diff
@@ -0,0 +1,24 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+
+namespace Apache.Arrow.Ipc
+{
+    public interface IArrowArrayStream : IArrowReader, IDisposable
```
CurtHagenlocher, line 20: I don't know why IArrowReader wasn't made disposable, but changing that now is breaking and so not doing when we can just define a new interface.

csharp/src/Apache.Arrow/C/CArrowArrayStreamExporter.cs
```diff
@@ -0,0 +1,120 @@
...
+            {
+                IArrowArrayStream arrayStream = FromPointer(cArrayStream->private_data);
+                CArrowSchemaExporter.ExportSchema(arrayStream.Schema, cSchema);
+                return 0;
+            }
+            catch (Exception)
+            {
+                return 1;
+            }
+        }
```
CurtHagenlocher, line 77: Oops, I seem to have forgotten about these.

csharp/src/Apache.Arrow/C/CArrowArrayExporter.cs
```diff
@@ -0,0 +1,210 @@
...
+            if (cArray->release != null)
+            {
+                throw new ArgumentException("Cannot export array to a struct that is already initialized.", nameof(cArray));
+            }
+
+            ExportedAllocationOwner allocationOwner = new ExportedAllocationOwner();
+            try
+            {
+                ConvertArray(allocationOwner, array.Data, cArray);
+                cArray->release = (delegate* unmanaged[Stdcall]<CArrowArray*, void>)Marshal.GetFunctionPointerForDelegate<ReleaseArrowArray>(ReleaseArray);
```
eerhardt, line 60: I think we can do this another way. If you add `[UnmanagedCallersOnly]` to the `private unsafe static void ReleaseArray(CArrowArray* cArray)` method, then this line should just be:

```suggestion
                cArray->release = (delegate* unmanaged[Stdcall]<CArrowArray*, void>)&ReleaseArray;
```

eerhardt: This applies for all the function pointers we need to set on these structs.

csharp/src/Apache.Arrow/C/CArrowArrayExporter.cs
```diff
@@ -0,0 +1,210 @@
...
+        private unsafe static void ConvertRecordBatch(ExportedAllocationOwner sharedOwner, RecordBatch batch, CArrowArray* cArray)
+        {
+            cArray->length = batch.Length;
+            cArray->offset = 0;
+            cArray->null_count = 0;
+            cArray->release = (delegate* unmanaged[Stdcall]<CArrowArray*, void>)Marshal.GetFunctionPointerForDelegate<ReleaseArrowArray>(ReleaseArray);
+            cArray->private_data = null;
+
+            cArray->n_buffers = 1;
+            cArray->buffers = (byte**)Marshal.AllocCoTaskMem(IntPtr.Size);
```
eerhardt, line 167: Why are we using `AllocCoTaskMem` here and in other places?

csharp/src/Apache.Arrow/C/CArrowArrayExporter.cs
```diff
@@ -0,0 +1,210 @@
...
+            if (cArray->n_buffers > 0)
+            {
+                cArray->buffers = (byte**)Marshal.AllocCoTaskMem(array.Buffers.Length * IntPtr.Size);
+                for (int i = 0; i < array.Buffers.Length; i++)
+                {
+                    ArrowBuffer buffer = array.Buffers[i];
+                    IntPtr ptr;
+                    if (!buffer.TryExport(sharedOwner, out ptr))
+                    {
+                        throw new NotSupportedException(); // TODO
```
eerhardt, line 132: Can we address this TODO in this PR?

csharp/src/Apache.Arrow/Memory/ExportedAllocationOwner.cs
```diff
@@ -0,0 +1,55 @@
...
+namespace Apache.Arrow.Memory
+{
+    internal sealed class ExportedAllocationOwner : INativeAllocationOwner, IDisposable
+    {
+        private readonly List<IntPtr> _pointers = new List<IntPtr>();
+        private int _allocationSize;
+
+        ~ExportedAllocationOwner()
+        {
+            Dispose();
```
eerhardt, line 29: This should implement the Disposable pattern.

https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose

csharp/src/Apache.Arrow/Memory/NativeMemoryManager.cs
```diff
@@ -15,24 +15,32 @@
...
         private readonly int _length;
+        private readonly INativeAllocationOwner _owner;
 
         public NativeMemoryManager(IntPtr ptr, int offset, int length)
         {
             _ptr = ptr;
             _offset = offset;
             _length = length;
+            _owner = NativeMemoryAllocator.ExclusiveOwner;
+        }
```
eerhardt, line 36:
```suggestion
        public NativeMemoryManager(IntPtr ptr, int offset, int length)
            : this(NativeMemoryAllocator.ExclusiveOwner, ptr, offset, length)
        {
        }
```

csharp/src/Apache.Arrow/Memory/NativeMemoryManager.cs
```diff
@@ -64,20 +72,36 @@ public override void Unpin()
...
+            if (ptr != IntPtr.Zero)
+            {
+                _owner.Release(ptr, _offset, _length);
+            }
+        }
 
-            lock (this)
+        bool IOwnableAllocation.TryAcquire(out IntPtr ptr, out int offset, out int length)
+        {
+            // TODO: implement refcounted buffers?
```
eerhardt, line 84: Will this get addressed in this PR? If not, we should open an issue for it.

csharp/test/Apache.Arrow.Tests/CDataInterfacePythonTests.cs
```diff
@@ -166,6 +169,79 @@ private static dynamic GetPythonSchema()
             }
         }
 
+        private IArrowArray GetTestArray()
```
eerhardt, line 172: Can we add some tests using the TestData that has all types?

csharp/src/Apache.Arrow/Arrays/ArrowArrayFactory.cs
```diff
@@ -25,6 +25,8 @@ public static IArrowArray BuildArray(ArrayData data)
         {
             switch (data.DataType.TypeId)
             {
+                case ArrowTypeId.Null:
```
eerhardt, line 28: Can we add some tests for the Null array?

eerhardt, line 28: We should also update 

https://github.com/apache/arrow/blob/122fc86576b03ebd130e027200c5eda6a185d2ee/docs/source/status.rst?plain=1#L32-L35

for this new functionality.

We should also turn on this integration test as well:

https://github.com/apache/arrow/blob/122fc86576b03ebd130e027200c5eda6a185d2ee/dev/archery/archery/integration/datagen.py#L1672-L1673

eerhardt (commented):
Thanks for the contribution, @CurtHagenlocher!

I got some time tonight to take a swipe through and posted some comments here. I haven't looked deep into the Memory changes yet. I will when I get time.

csharp/src/Apache.Arrow/Memory/ExportedAllocationOwner.cs
CurtHagenlocher: I thought that advice applies only to non-sealed classes.

csharp/src/Apache.Arrow/C/CArrowArrayExporter.cs
CurtHagenlocher: It would still work correctly if an array were exported via the C API to another bit of managed code which consumed it (via the C API)?

csharp/src/Apache.Arrow/C/CArrowArrayExporter.cs
CurtHagenlocher: derp

csharp/src/Apache.Arrow/Arrays/ArrowArrayFactory.cs
CurtHagenlocher: I should be able to add tests within the next day.

csharp/src/Apache.Arrow/Arrays/NullArray.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        public int Offset => Data.Offset;
+
+        public int NullCount => Data.NullCount;
+
+        public void Dispose() { }
+        public bool IsNull(int index) => true;
+        public bool IsValid(int index) => false;
+
+        public void Accept(IArrowArrayVisitor visitor)
```
CurtHagenlocher, line 107: Implementation can be simplified.

csharp/src/Apache.Arrow/Memory/ExportedAllocationOwner.cs
eerhardt: Only the `virtual Dispose(bool)` part applies to non-sealed classes.

The part you still want is ` GC.SuppressFinalize(this);` when `Dispose()` is called on the object.

>> The Dispose method performs all object cleanup, so the garbage collector no longer needs to call the objects' [Object.Finalize](https://learn.microsoft.com/en-us/dotnet/api/system.object.finalize) override. Therefore, the call to the [SuppressFinalize](https://learn.microsoft.com/en-us/dotnet/api/system.gc.suppressfinalize) method prevents the garbage collector from running the finalizer.

This has performance advantages because the objects don't need to survive from GC gen 0 to gen 1 just to run the finalizer.

csharp/src/Apache.Arrow/C/CArrowArrayExporter.cs
eerhardt: > It would still work correctly if an array were exported via the C API to another bit of managed code which consumed it (via the C API)?

Yes because it is the `release` pointer is an unmanaged function pointer, and not a managed delegate. It is a bit hard to explain. But even though the caller is "managed" (i.e. .NET code), the way the method is invoked is through an unmanged function pointer. So it is still an "UnmanagedCaller".

eerhardt: > Is it worth using conditional compilation to optimize for .NET 5+?

Maybe trying writing a benchmark test in https://github.com/apache/arrow/tree/main/csharp/test/Apache.Arrow.Benchmarks to compare the difference? If we see major differences, it may make sense to split the compilation.

eerhardt:
Current test failure is:

```
Unhandled exception: System.NotSupportedException: JsonArrowType not supported: null
   at Apache.Arrow.IntegrationTest.IntegrationCommand.ToArrowType(JsonArrowType type) in /arrow/csharp/test/Apache.Arrow.IntegrationTest/IntegrationCommand.cs:line 176
   at Apache.Arrow.IntegrationTest.IntegrationCommand.CreateField(Builder builder, JsonField jsonField) in /arrow/csharp/test/Apache.Arrow.IntegrationTest/IntegrationCommand.cs:line 152
   at Apache.Arrow.IntegrationTest.IntegrationCommand.<>c__DisplayClass17_1.<CreateSchema>b__0(Builder f) in /arrow/csharp/test/Apache.Arrow.IntegrationTest/IntegrationCommand.cs:line 145
   at Apache.Arrow.Schema.Builder.Field(Action`1 fieldBuilderAction) in /arrow/csharp/src/Apache.Arrow/Schema.Builder.cs:line 53
   at Apache.Arrow.IntegrationTest.IntegrationCommand.CreateSchema(JsonSchema jsonSchema) in /arrow/csharp/test/Apache.Arrow.IntegrationTest/IntegrationCommand.cs:line 145
   at Apache.Arrow.IntegrationTest.IntegrationCommand.Validate() in /arrow/csharp/test/Apache.Arrow.IntegrationTest/IntegrationCommand.cs:line 75
   at Apache.Arrow.IntegrationTest.IntegrationCommand.Execute() in /arrow/csharp/test/Apache.Arrow.IntegrationTest/IntegrationCommand.cs:line 58
   at Apache.Arrow.IntegrationTest.Program.<>c.<<Main>b__0_0>d.MoveNext() in /arrow/csharp/test/Apache.Arrow.IntegrationTest/Program.cs:line 49
```

Need to update:

https://github.com/apache/arrow/blob/3948c426927514ab6d3165255b4717af9446e949/csharp/test/Apache.Arrow.IntegrationTest/IntegrationCommand.cs#L162-L176

CurtHagenlocher:
Yes, already fixed locally. I'm faulting in knowledge of the codebase as I encounter failing tests... .

csharp/src/Apache.Arrow/C/CArrowArrayExporter.cs
eerhardt: Actually, looking at this deeper, the current code may have problems since the managed Delegate might be GC'd.

See https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.getfunctionpointerfordelegate?view=net-8.0

> You must manually keep the delegate from being collected by the garbage collector from managed code. The garbage collector does not track references to unmanaged code.

CurtHagenlocher: Yes, you'd think I'd have remembered this given that I pointed it out in a private email last month :/.

CurtHagenlocher: I think I've taken care of the lifetime issues and would probably file a work item to test for optimization.

csharp/test/Apache.Arrow.Tests/TestData.cs
```diff
@@ -48,7 +50,7 @@ public static RecordBatch CreateSampleRecordBatch(int length, int columnSetCount
                 builder.Field(CreateField(Date32Type.Default, i));
                 builder.Field(CreateField(Date64Type.Default, i));
                 builder.Field(CreateField(Time32Type.Default, i));
-                builder.Field(CreateField(Time64Type.Default, i));
```
CurtHagenlocher, line 51: The time unit on Time64Type.Default seems like an odd choice in that it produces a type which can't be expressed via the C API.

eerhardt: I'd like to understand this better. Why can't `TimeUnit.Millisecond` work with the C API?

CurtHagenlocher: Time64 with milliseconds is not explicitly enumerated at https://arrow.apache.org/docs/format/CDataInterface.html and it would need a new format specifier to work. This is because "tt" is used for both Time32 and Time64 and the time unit is used to determine which one it will be. (This is perhaps an error in the specification, but the spec is what it is.)

davidhcoe: it looks like Tim32 can be seconds or milliseconds where as Time64 should be microseconds or nanoseconds.

davidhcoe (commented):
Reviewed for high level functionality and confirmed unit tests pass.

csharp/src/Apache.Arrow/Arrays/NullArray.cs
```diff
@@ -0,0 +1,109 @@
...
+                return this;
+            }
+
+            private void CheckIndex(int index)
+            {
+                if (index < 0 || index >= Length)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(index));
+                }
+            }
```
westonpace, line 80: Is this function used?

csharp/src/Apache.Arrow/Arrays/NullArray.cs
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        public ArrayData Data { get; }
+
+        public NullArray(ArrayData data)
+        {
+            data.EnsureDataType(ArrowTypeId.Null);
+            data.EnsureBufferCount(0);
+            Data = data;
```
westonpace, line 89: Minor nit: should we ensure the null count and length are equal?

csharp/src/Apache.Arrow/Memory/ExportedAllocationOwner.cs
```diff
@@ -0,0 +1,62 @@
...
+using System;
+using System.Collections.Generic;
+using System.Runtime.InteropServices;
+
+namespace Apache.Arrow.Memory
+{
+    internal sealed class ExportedAllocationOwner : INativeAllocationOwner, IDisposable
+    {
+        private readonly List<IntPtr> _pointers = new List<IntPtr>();
+        private int _allocationSize;
```
westonpace, line 25: Is this used for anything?

csharp/src/Apache.Arrow/C/CArrowArrayImporter.cs
```diff
@@ -0,0 +1,277 @@
...
+                {
+                    throw new ArgumentException("Tried to import an array that has already been released.", nameof(cArray));
+                }
+            }
+
+            protected override void FinalRelease()
+            {
+                if (_cArray->release != null)
+                {
+                    _cArray->release(_cArray);
```
westonpace, line 112: Do you need to free `_cArray`?

Nevermind.  I see that `_cArray` comes from the caller of `ImportArray` and they are responsible for deleting it (which it seems they can always do as soon as the call to `ImportArray` is finished).

csharp/src/Apache.Arrow/Memory/ImportedAllocationOwner.cs
```diff
@@ -0,0 +1,68 @@
...
+        public IMemoryOwner<byte> AddMemory(IntPtr ptr, int offset, int length)
+        {
+            if (_referenceCount <= 0)
+            {
+                throw new ObjectDisposedException(typeof(ImportedAllocationOwner).Name);
+            }
+
+            NativeMemoryManager memory = new NativeMemoryManager(this, ptr, offset, length);
+            Interlocked.Increment(ref _referenceCount);
+            Interlocked.Add(ref _managedMemory, length);
```
westonpace, line 41: Minor nit: you could probably move this statement inside the if right below it.

csharp/src/Apache.Arrow/C/CArrowArrayStreamImporter.cs
```diff
@@ -0,0 +1,135 @@
...
+                {
+                    throw new ArgumentException("Tried to import an array stream that has already been released.", nameof(cArrayStream));
+                }
+
+                CArrowSchema* cSchema = CArrowSchema.Create();
+                try
+                {
+                    if (_cArrayStream->get_schema(_cArrayStream, cSchema) != 0)
+                    {
+                        throw new Exception("This needs to be better");
```
westonpace, line 73: Perhaps, for now, just "Unexpected error received from external stream. Errno: {errno}"?  At the very least I think we should include the error number in the error message.

csharp/src/Apache.Arrow/C/CArrowArrayStreamImporter.cs
```diff
@@ -0,0 +1,135 @@
...
+                    throw new ObjectDisposedException(typeof(ImportedArrowArrayStream).Name);
+                }
+
+                RecordBatch result = null;
+                CArrowArray* cArray = CArrowArray.Create();
+                try
+                {
+                    if (_cArrayStream->get_next(_cArrayStream, cArray) != 0)
+                    {
+                        throw new Exception("This too needs to be better");
```
westonpace, line 106: Same as above.  Let's include the error number.

csharp/src/Apache.Arrow/C/CArrowArrayStreamImporter.cs
```diff
@@ -0,0 +1,135 @@
...
+                }
+                finally
+                {
+                    if (result == null)
+                    {
+                        CArrowArray.Free(cArray);
+                    }
+                }
+
+                return new ValueTask<RecordBatch>(result);
```
westonpace, line 121: You could, potentially, accept a task scheduler and schedule a new task to call `get_next`, allowing this to be more accurately async.  Though that should definitely be a follow-up and probably depends on what you're interfacing with (e.g. is the underlying stream performing I/O and slow?)

westonpace (approved):
A few minor nits but this looks very well thought out to me.

csharp/src/Apache.Arrow/Memory/ExportedAllocationOwner.cs
CurtHagenlocher: Ah, nice catch. This was for memory pressure bookkeeping that I forgot to add.

eerhardt (approved):
Thanks, @CurtHagenlocher! This is great!

csharp/src/Apache.Arrow/C/CArrowArrayStreamImporter.cs
```diff
@@ -0,0 +1,137 @@
...
+
+            public ValueTask<RecordBatch> ReadNextRecordBatchAsync(CancellationToken cancellationToken = default)
+            {
+                if (_disposed)
+                {
+                    throw new ObjectDisposedException(typeof(ImportedArrowArrayStream).Name);
+                }
+
+                RecordBatch result = null;
+                CArrowArray* cArray = CArrowArray.Create();
```
adamreeve, line 102: Hi @CurtHagenlocher, thanks for implementing this! I've started testing using this to support reading Parquet data as Arrow record batches in [ParquetSharp](https://github.com/G-Research/ParquetSharp).

One concern I've found is that it appears that an imported stream will leak memory as these `CArrowArray` instances are allocated for each batch but they're never freed, as `ImportedArrowArray.FinalRelease` will call the `release` callback but never deallocates the `CArrowArray` struct itself.

Is that correct or am I missing something here?

Would it make sense for the imported array to take ownership of the `CArrowArray` struct and deallocate it after calling release?

This seems to be an issue for external use of the C Data Interface API too, eg. if I want to return an `IArrowArrayStream` from a method I can't free the `CArrowArrayStream` that was used to import it until after the user is finished with the stream, which is awkward.

CurtHagenlocher: Yes, this has already been discovered and there's a PR out for fixing it: https://github.com/apache/arrow/pull/35810

CurtHagenlocher: Okay, my pattern matching was a bit too eager for the first problem. Yes, that looks like a leak and I think your suggestion is right; that on this code path the ImportedArrowArray needs to remember that it owns the allocation.

The second problem feels different, because in the first case we always know that we were the ones who allocated the CArrowArray but I'm not entirely sure we know that about the CArrowArrayStream. It's true for the stream we got from Python in the test case, but that was using a pyarrow-specific API. The flavor of the C API as a whole does suggest that it will usually be the case for the caller to have to allocate the CArrowArrayStream without (I think) quite making it explicit.

On the whole, I suspect that both importers should take a flag which says whether or not to deallocate the structure afterwards. I'm not convinced about the right default for the flag given the relative risks of leaking memory vs deallocating it inappopriately.

csharp/src/Apache.Arrow/C/CArrowArrayExporter.cs
```diff
@@ -0,0 +1,211 @@
...
+                throw new ArgumentNullException(nameof(array));
+            }
+            if (cArray == null)
+            {
+                throw new ArgumentNullException(nameof(cArray));
+            }
+            if (cArray->release != null)
+            {
+                throw new ArgumentException("Cannot export array to a struct that is already initialized.", nameof(cArray));
+            }
```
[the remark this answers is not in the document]
CurtHagenlocher: I was following the existing pattern in CArrowSchemaExporter. But I also think that the documentation, in saying that "A released structure is indicated by setting its release callback to NULL. Before reading and interpreting a structure’s data, consumers SHOULD check for a NULL release callback and treat it accordingly (probably by erroring out)." suggests that this check is appropriate.

csharp/src/Apache.Arrow/C/CArrowArrayStreamImporter.cs
CurtHagenlocher: Filed #35988

adamreeve: Right, yeah having a flag to say whether the structure should be deallocated would solve both problems if it was public and implemented for both the array and stream. I think keeping the current behaviour of not deallocating makes sense as a default, given a small memory leak is less of an issue than incorrectly deallocating, and then that keeps allocation and deallocation symmetric by default.

It wouldn't be a big deal if only the first problem was solved though, I can solve the second problem with some extra indirection by introducing a wrapper for the imported stream that deallocates the struct on dispose.

csharp/src/Apache.Arrow/C/CArrowArrayExporter.cs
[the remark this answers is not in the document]
CurtHagenlocher: This is part of why C is awful ;).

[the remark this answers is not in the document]
CurtHagenlocher: Addressed as part of https://github.com/apache/arrow/pull/35996.

csharp/src/Apache.Arrow/C/CArrowArrayStreamImporter.cs
CurtHagenlocher: https://github.com/apache/arrow/pull/35996

adamreeve: Thanks! That solution of moving the struct into the imported array or stream is much nicer

csharp/src/Apache.Arrow/C/CArrowArrayExporter.cs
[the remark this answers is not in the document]
paleolimbot: I'm sorry I missed this and I see that it's been solved. In nanoarrow we definitely assume that pointer output arguments point to uninitialized memory (and strive to not touch that memory until failure is impossible). There are a few places where we do something like 

```c
struct ArrowArray tmp;
tmp.release = NULL;
// stuff with tmp that might fail
if (had_error) {
  if (tmp.release != NULL) {
    tmp.release(&tmp);
  }

  return;
}

ArrowArrayMove(&tmp, out);
return;
```

...to simplify (to the extent that anything in C is simple) the error handling.

PR #35499, opened 2023-05-09
GH-35304: [C++][ORC] Support attributes conversion

### LINKED ISSUE #35304: [Python] Read ORC metadata

When reading an ORC schema, the metadata of the fields isn't exposed. For more information check: https://github.com/apache/iceberg/issues/6973#issuecomment-1465090060

### Component(s)

Python

wgtmac:
### Rationale for this change

Apache Orc has a per column attribute map and Apache Iceberg depends on this to encode its field metadata. However, the C++ Orc adapter does not know it which makes it difficult to support pyarrow and pyiceberg.

Both reader and writer support Orc attributes conversion from/to arrow field metadata.

### Are these changes tested?

Added two test cases to make sure the Orc adapter can preserve the attributes well.

No.
* Closes: #35304

### DIFF
--- a/cpp/src/arrow/adapters/orc/adapter_test.cc
+++ b/cpp/src/arrow/adapters/orc/adapter_test.cc
@@ -506,6 +506,116 @@ TEST(TestAdapterRead, ReadCharAndVarcharType) {
   ASSERT_EQ(nullptr, record_batch);
 }
 
+TEST(TestAdapterRead, ReadFieldAttributes) {
+  const std::string id_key = "iceberg.id";
+  const std::string required_key = "iceberg.required";
+
+  auto set_attributes = [&](liborc::Type* type, const std::string& id,
+                            const std::string& required) {
+    type->setAttribute(id_key, id);
+    type->setAttribute(required_key, required);
+  };
+
+  auto check_attributes = [&](const std::shared_ptr<arrow::Field>& field,
+                              const std::string& expect_id,
+                              const std::string& expect_required) {
+    auto field_metadata = field->metadata();
+    ASSERT_NE(field_metadata, nullptr);
+    ASSERT_EQ(expect_id, field_metadata->Get(id_key));
+    ASSERT_EQ(expect_required, field_metadata->Get(required_key));
+  };
+
+  auto c1_type = liborc::createPrimitiveType(liborc::TypeKind::INT);
+  set_attributes(c1_type.get(), "1", "true");
+
+  auto c2_elem_type = liborc::createPrimitiveType(liborc::TypeKind::INT);
+  set_attributes(c2_elem_type.get(), "3", "false");
+  auto c2_type = liborc::createListType(std::move(c2_elem_type));
+  set_attributes(c2_type.get(), "2", "false");
+
+  auto c3_key_type = liborc::createPrimitiveType(liborc::TypeKind::INT);
+  set_attributes(c3_key_type.get(), "5", "true");
+  auto c3_value_type = liborc::createPrimitiveType(liborc::TypeKind::INT);
+  set_attributes(c3_value_type.get(), "6", "false");
+  auto c3_type = liborc::createMapType(std::move(c3_key_type), std::move(c3_value_type));
+  set_attributes(c3_type.get(), "4", "false");
+
+  auto c4_sub_type = liborc::createPrimitiveType(liborc::TypeKind::INT);
+  set_attributes(c4_sub_type.get(), "8", "false");
+  auto c4_type = liborc::createStructType();
+  c4_type->addStructField("c4_1", std::move(c4_sub_type));
+  set_attributes(c4_type.get(), "7", "false");
+
+  auto orc_type = liborc::createStructType();
+  orc_type->addStructField("c1", std::move(c1_type));
+  orc_type->addStructField("c2", std::move(c2_type));
+  orc_type->addStructField("c3", std::move(c3_type));
+  orc_type->addStructField("c4", std::move(c4_type));
+
+  MemoryOutputStream mem_stream(kDefaultMemStreamSize);
+  auto writer = CreateWriter(/*stripe_size=*/1024, *orc_type, &mem_stream);
+  writer->close();
+
+  std::shared_ptr<io::RandomAccessFile> in_stream(std::make_shared<io::BufferReader>(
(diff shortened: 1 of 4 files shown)


wgtmac:
@Fokko Could you please take a look? Thanks!

Fokko (approved):
As far as my C++ skills go, this looks good. I think this PR is a good start, the next step would be to expose this in Python as well.

wgtmac:
Thanks @Fokko!

Actually I didn't add any new API. Therefore `pyiceberg` should be able to get the missing info from arrow schema field metadata via `pyarrow`. I am not familiar with pyarrow, so please correct me if I am wrong. @jorisvandenbossche

cpp/src/arrow/adapters/orc/util.cc
```diff
@@ -1176,11 +1190,32 @@ Result<std::unique_ptr<liborc::Type>> GetOrcType(const Schema& schema) {
...
   }
   return std::move(out_type);
 }
 
+Result<std::shared_ptr<const KeyValueMetadata>> GetFieldMetadata(
+    const liborc::Type* type) {
+  if (type == nullptr) {
+    return nullptr;
+  }
+  const auto keys = type->getAttributeKeys();
```
pitrou, line 1204: You might actually want to avoid the metadata allocation here.
```suggestion
  const auto keys = type->getAttributeKeys();
  if (keys.empty()) {
    return nullptr;
  }
```

cpp/src/arrow/adapters/orc/util.h
```diff
@@ -23,6 +23,7 @@
 #include "arrow/array/builder_base.h"
 #include "arrow/result.h"
 #include "arrow/status.h"
+#include "arrow/util/key_value_metadata.h"
```
pitrou, line 26: Only the forward declarations should be necessary here.
```suggestion
#include "arrow/type_fwd.h"
```

cpp/src/arrow/adapters/orc/adapter_test.cc
```diff
@@ -506,6 +506,116 @@ TEST(TestAdapterRead, ReadCharAndVarcharType) {
...
+  ARROW_EXPECT_OK(writer->Write(*expected_output_table));
+  ARROW_EXPECT_OK(writer->Close());
+
+  EXPECT_OK_AND_ASSIGN(auto buffer, buffer_output_stream->Finish());
+  std::shared_ptr<io::RandomAccessFile> in_stream(new io::BufferReader(buffer));
+  EXPECT_OK_AND_ASSIGN(
+      auto reader, adapters::orc::ORCFileReader::Open(in_stream, default_memory_pool()));
+  EXPECT_OK_AND_ASSIGN(auto actual_output_table, reader->Read());
+  ASSERT_OK(actual_output_table->ValidateFull());
+  AssertTablesEqual(*expected_output_table, *actual_output_table, false, false);
```
pitrou, line 612: Can you qualify the two boolean arguments?
```suggestion
  AssertTablesEqual(*expected_output_table, *actual_output_table, /*xxx=*/ false, /*yyy=*/ false);
```

cpp/src/arrow/adapters/orc/adapter_test.cc
```diff
@@ -506,6 +506,116 @@ TEST(TestAdapterRead, ReadCharAndVarcharType) {
...
+  write_options.compression = Compression::UNCOMPRESSED;
+  EXPECT_OK_AND_ASSIGN(auto writer, adapters::orc::ORCFileWriter::Open(
+                                        buffer_output_stream.get(), write_options));
+
+  auto schema = ::arrow::schema(
+      {::arrow::field("c0", ::arrow::int64(), true, key_value_metadata({"k0"}, {"v0"})),
+       ::arrow::field("c1", ::arrow::utf8(), true, key_value_metadata({"k1"}, {"v1"})),
+       ::arrow::field(
+           "c2", ::arrow::list(::arrow::field("item", ::arrow::int64(), true,
+                                              key_value_metadata({"k0"}, {"ddv0"}))))});
```
pitrou, line 601: Could you also add the other kinds of nested fields? (struct, map, union)

pitrou (changes requested):
Thanks @wgtmac ! Here are some comments and suggestions.

cpp/src/arrow/adapters/orc/util.cc
```diff
@@ -1176,11 +1190,32 @@ Result<std::unique_ptr<liborc::Type>> GetOrcType(const Schema& schema) {
...
     ARROW_ASSIGN_OR_RAISE(auto orc_subtype, GetOrcType(*field->type()));
+    SetAttributes(field, orc_subtype.get());
     out_type->addStructField(field->name(), std::move(orc_subtype));
   }
   return std::move(out_type);
 }
 
+Result<std::shared_ptr<const KeyValueMetadata>> GetFieldMetadata(
+    const liborc::Type* type) {
+  if (type == nullptr) {
```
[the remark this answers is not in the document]
wgtmac: I follow the same behavior here: https://github.com/apache/arrow/blob/main/cpp/src/arrow/adapters/orc/util.cc#L1078

wgtmac:
Thanks @pitrou! Welcome back!

cpp/src/arrow/adapters/orc/util.cc
```diff
@@ -951,6 +952,15 @@ Status WriteBatch(const Array& array, int64_t orc_offset,
   }
 }
 
+void SetAttributes(const std::shared_ptr<arrow::Field>& field, liborc::Type* type) {
+  if (field->HasMetadata()) {
+    const auto& metadata = field->metadata();
+    for (int j = 0; j < metadata->size(); j++) {
```
pitrou, line 958:
```suggestion
    for (int64_t j = 0; j < metadata->size(); j++) {
```

wgtmac: Oops, that was refactored out of an inner loop but I forgot to rename it.

pitrou (approved):
Thanks for the update!

jorisvandenbossche:
> Actually I didn't add any new API. Therefore `pyiceberg` should be able to get the missing info from arrow schema field metadata via `pyarrow`. I am not familiar with pyarrow, so please correct me if I am wrong

Yes, that should be the case

PR #35502, opened 2023-05-09
GH-32282: [R] Update case_when() binding to match changes in dplyr

thisisnic:
This PR implements the `.default` argument for `case_when` and raises errors if non-NULL values are supplied for `.ptype` and `.size`.  
* Closes: #32282

### DIFF
--- a/r/R/dplyr-funcs-conditional.R
+++ b/r/R/dplyr-funcs-conditional.R
@@ -140,5 +140,6 @@ register_bindings_conditional <- function() {
         value
       )
     )
-  })
+  }, notes = "`.ptype` and `.size` arguments not supported"
+  )
 }
--- a/r/R/dplyr-funcs-conditional.R
+++ b/r/R/dplyr-funcs-conditional.R
@@ -90,7 +90,15 @@ register_bindings_conditional <- function() {
     out
   })
 
-  register_binding("dplyr::case_when", function(..., .default = NULL) {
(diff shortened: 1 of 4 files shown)


thisisnic:
We could implement `.ptype` in a follow-up ticket if there is user interest in it. A possible implementation might involve calling cast on the `case_when` expression to cast it to the correct type.  We might want to use `infer_type()` so users can pass in e.g. `integer()` like in the `.ptype` example in the dplyr tests.  To do this, we can implement `infer_type.DataType` which just passes the input object back out - then users can input either R objects or Arrow datatypes to this function.

r/R/dplyr-funcs-conditional.R
```diff
@@ -90,7 +90,15 @@ register_bindings_conditional <- function() {
     out
   })
 
-  register_binding("dplyr::case_when", function(...) {
+  register_binding("dplyr::case_when", function(..., .default = NULL, .ptype = NULL, .size = NULL) {
```
eitsupi, line 93: > raises errors if non-NULL values are supplied for .ptype and .size.

Would it be better to mention the limitation with the `notes`?

https://github.com/apache/arrow/blob/7f3e87d2d5e6f0568dbb92b468d911717d9aed91/r/R/dplyr-funcs.R#L63-L64

paleolimbot (approved):
Thanks!

PR #35505, opened 2023-05-09
GH-35389: [Python] Fix coalesce_keys=False option in join operation

### LINKED ISSUE #35389: [Python] Table.join no longer respecting `coalesce_keys` parameter (PyArrow 12)

Table.join has an attribute `coalesce_keys`, which the documentation says:

> coalesce_keys[bool](https://docs.python.org/3/library/stdtypes.html#bltin-boolean-values), default [True](https://docs.python.org/3/library/constants.html#True)
> If the duplicated keys should be omitted from one of the sides in the join result.

In PyArrow v11, the columns used to perform the join were retained in the resultant table when this parameter was set to `False`. However, in v12, the column from the 'right' table (the one in the parameters) is omitted from the result.

A review of the change log for v12 doesn't suggest this change in behaviour is intentional.

This change in behaviour was observed in the matrix regression testing for [Opteryx](https://github.com/mabel-dev/opteryx), which has matrix regression testing across Mac, Linux and Windows, for Python versions 3.8, 3.9, 3.10 and 3.11, and for Left and Inner JOINs - all variations in the test matrix appear to have this same behaviour. 

No error is observed, only a change in behaviour.

### Component(s)

Python

jorisvandenbossche:
### Rationale for this change

During the refactor of the join cython code to use the new `pyarrow.acero` code, I accidentally ignored the `coalesce_keys=False` option. 
This PR restores the previous behaviour (by not passing a custom subset of column names to the HashJoinNode, but relying on its default behaviour to include all fields from left and right data (depending on the join type)).

### Are these changes tested?

Expanded the existing tests to now properly cover the coalesce_keys=False option for all join types.

Fixes a regression in 12.0, restoring behaviour of 11.0
* Closes: #35389

### DIFF
--- a/python/pyarrow/acero.py
+++ b/python/pyarrow/acero.py
@@ -23,7 +23,7 @@
 # cython: language_level = 3
 
 from pyarrow.lib import Table
-from pyarrow.compute import Expression
+from pyarrow.compute import Expression, field
 
 from pyarrow._acero import (  # noqa
     _group_by,
@@ -52,6 +52,13 @@ except ImportError:
 
 def _dataset_to_decl(dataset, use_threads=True):
     decl = Declaration("scan", ScanNodeOptions(dataset, use_threads=use_threads))
+
+    # Get rid of special dataset columns
+    # "__fragment_index", "__batch_index", "__last_in_fragment", "__filename"
+    projections = [field(f) for f in dataset.schema.names]
+    decl = Declaration.from_sequence(
+        [decl, Declaration("project", ProjectNodeOptions(projections))]
+    )
 
     filter_expr = dataset._scan_options.get("filter")
     if filter_expr is not None:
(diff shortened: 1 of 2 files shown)


python/pyarrow/acero.py
```diff
@@ -53,6 +53,13 @@ class InMemoryDataset:
 def _dataset_to_decl(dataset, use_threads=True):
     decl = Declaration("scan", ScanNodeOptions(dataset, use_threads=use_threads))
 
+    # Get rid of special dataset columns
+    # "__fragment_index", "__batch_index", "__last_in_fragment", "__filename"
+    projections = [field(f) for f in dataset.schema.names]
+    decl = Declaration.from_sequence(
+        [decl, Declaration("project", ProjectNodeOptions(projections))]
+    )
```
jorisvandenbossche, line 61: @westonpace do you know if there is an easier way to do this? (there is not an option to just turn off adding those fields in the scanner to start with?)

westonpace: This was on my "new scan node" todo list but sadly that list has been stalled for a while now.

python/pyarrow/tests/test_exec_plan.py
```diff
@@ -115,12 +120,19 @@ def test_joins(jointype, expected, use_threads, use_datasets):
...
         r = r.sort_by("colB")
     else:
         r = r.sort_by("colA")
+    if coalesce_keys:
+        if jointype in ("inner", "left outer"):
+            expected = expected.drop(["colB"])
+        elif jointype == "right outer":
+            expected = expected.drop(["colA"])
+        elif jointype == "full outer":
+            expected = expected.drop(["colB"]).set_column(0, "colA", [[1, 2, 6, 99]])
```
westonpace, line 135: Does this work?  I wouldn't expect this behavior.

westonpace (commented):
For some reason, I thought coalescing a full outer join required an actual call to the `coalesce` function (e.g. with a project node).  However, I don't see that here.  Maybe you just aren't coalescing on a full outer join (which is probably fine)?

westonpace (approved):
Ah, thanks for the clarification.  This looks good then.

PR #35512, opened 2023-05-09
WIP: Only testing ARM self-hosted runners

raulcd:


pitrou:
@raulcd Should this be kept open?

raulcd:
I want to keep testing the possibility of adding cross-compilation capabilities but we can close the PR as my tests can be performed outside the PR. I can reopen once there is something that could potentially be merged.

PR #35513, opened 2023-05-09
GH-35506: [C++] Support First and Last aggregators in Substrait

### LINKED ISSUE #35506: [C++] Support First and Last aggregators in Substrait

Currently, Ibis-Substrait defines First and Last aggregators while Arrow-Substrait does not support them. This issue will add the support.

### Component(s)

C++

rtpsw:
### Rationale for this change

See #35506

"first" and "last" are registered as aggregates in Substrait.

### Are these changes tested?

Yes, The test-cases cover grouped and non-grouped aggregation with "first" and "last", as well as "null" and non-"null" outputs.

Yes, "first" and "last" will be valid in a consumed Substrait plan.
* Closes: #35506


rtpsw:
cc @icexelloss

rtpsw:
cc @icexelloss

rtpsw:
This [CI job](https://github.com/apache/arrow/actions/runs/4926654580/jobs/8802535799?pr=35513) has an unrelated error that seems worth looking into separately.

cpp/src/arrow/engine/substrait/function_test.cc
```diff
@@ -574,6 +575,8 @@ struct AggregateTestCase {
   std::shared_ptr<DataType> output_type;
   // The aggregation takes zero columns as input
   bool nullary = false;
+  // The aggregation is ordered
+  bool ordered = false;
```
icexelloss, line 579: If aggregation is ordered, shouldn't this be true?

cpp/src/arrow/engine/substrait/serde.h
```diff
@@ -96,6 +96,26 @@ ARROW_ENGINE_EXPORT Result<std::shared_ptr<acero::ExecPlan>> DeserializePlan(
...
+/// \param[in,out] plan an ExecNode corresponding to the single toplevel relation in the
+/// Substrait Plan
+/// \param[in] buf a buffer containing the protobuf serialization of a Substrait Plan
+/// message
+/// \param[in] consumer node that consumes the batches produced by each toplevel Substrait
+/// relation
+/// \param[in] registry an extension-id-registry to use, or null for the default one.
+/// \param[out] ext_set_out if non-null, the extension mapping used by the Substrait
+/// \param[in] conversion_options options to control how the conversion is to be done.
+ARROW_ENGINE_EXPORT Status DeserializePlan(
```
icexelloss, line 113: Why this change

cpp/src/arrow/engine/substrait/serde.h
```diff
@@ -133,13 +153,34 @@ ARROW_ENGINE_EXPORT Result<std::vector<acero::Declaration>> DeserializePlans(
...
+/// Substrait Plan
+/// \param[in] buf a buffer containing the protobuf serialization of a Substrait Plan
+/// message
+/// \param[in] write_options write options of a node consuming the batches produced by
+/// each toplevel Substrait relation
+/// \param[in] registry an extension-id-registry to use, or null for the default one.
+/// \param[out] ext_set_out if non-null, the extension mapping used by the Substrait
+/// Plan is returned here.
+/// \param[in] conversion_options options to control how the conversion is to be done.
+ARROW_ENGINE_EXPORT Status DeserializePlan(
```
icexelloss, line 178: Why this change?

cpp/src/arrow/engine/substrait/serde.h
icexelloss: Also, shouldn't this return Result object similar to other Deserialize* function?

icexelloss:
@rtpsw there seems to be some refactoring changes. It is not clear why those changes are in this PR, can you comment?

icexelloss:
> What changes are included in this PR?
> "first" and "last" are registered as aggregates in Substrait.

There are more changes then just "register" first and last aggregates it seems. It would better if you can explain them upfront.

cpp/src/arrow/engine/substrait/function_test.cc
rtpsw: The pre-PR behavior is equivalent to post-PR behavior with `ordered = false`. In pre-PR, all tested aggregators had no trouble with the default executor. In post-PR, the `first` and `last` aggregators are ordered and with `ordered = false` there is an error `NotImplemented: Using ordered aggregator in multiple threaded execution is not supported`.

cpp/src/arrow/engine/substrait/serde.cc
```diff
@@ -236,14 +236,13 @@ ARROW_ENGINE_EXPORT Result<PlanInfo> DeserializePlan(
 
 namespace {
 
-Result<std::shared_ptr<acero::ExecPlan>> MakeSingleDeclarationPlan(
-    std::vector<acero::Declaration> declarations) {
+Status MakeSingleDeclarationPlan(acero::ExecPlan* plan,
```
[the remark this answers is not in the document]
rtpsw: This allows one to pass a preconfigured plan instead of getting a default-configured plan. The tester uses this to pass a plan with a non-default executor `ctx_for_ordered`.

cpp/src/arrow/engine/substrait/serde.h
rtpsw: Same rationale as in [this post](https://github.com/apache/arrow/pull/35513#discussion_r1188823725).

cpp/src/arrow/engine/substrait/serde.h
rtpsw: Same rationale as in [this post](https://github.com/apache/arrow/pull/35513#discussion_r1188823725).

westonpace (commented):
I don't see `first` / `last` in https://github.com/substrait-io/substrait/blob/main/extensions/functions_aggregate_generic.yaml

So wouldn't this be an Arrow extension function and we should use `kArrowSimpleExtensionFunctionsUri`.

rtpsw:
> I don't see first / last in https://github.com/substrait-io/substrait/blob/main/extensions/functions_aggregate_generic.yaml
>
> So wouldn't this be an Arrow extension function and we should use kArrowSimpleExtensionFunctionsUri.

You're right, this PR is ahead of Ibis-Substrait. @icexelloss, should we create a corresponding PR for Ibis-Substrait?

icexelloss:
> You're right, this PR is ahead of Ibis-Substrait. @icexelloss, should we create a corresponding PR for Ibis-Substrait?

Yeah I think so, since first and last is defined in ibis now

cpp/src/arrow/engine/substrait/serde.cc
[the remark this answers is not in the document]
icexelloss: I think you can use https://github.com/apache/arrow/blob/c3acc91c3e21f225ee7cc3cb3f489f0153365f5a/cpp/src/arrow/acero/exec_plan.h#L566

Instead and can avoid changing the serde files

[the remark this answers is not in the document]
icexelloss: i.e., instead of de serializing into a plan, you can de serializing into a declaration and use single-threaded or multi-threaded execution depending on ordered/non-ordered.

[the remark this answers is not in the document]
rtpsw: Done. It turned out not easy to use `DeclarationToTable` here because of the extra setup needed for linking the input table and the consumer, but I did separate out the deserialization and obtained simpler code.

cpp/src/arrow/engine/substrait/serde.h
```diff
@@ -89,8 +89,7 @@ ARROW_ENGINE_EXPORT Result<std::vector<acero::Declaration>> DeserializePlans(
 /// \param[out] ext_set_out if non-null, the extension mapping used by the Substrait
 /// \param[in] conversion_options options to control how the conversion is to be done.
 /// Plan is returned here.
-/// \return an ExecNode corresponding to the single toplevel relation in the Substrait
-/// Plan
+/// \return an ExecPlan for the Substrait Plan
```
icexelloss, line 92: Revert this?

cpp/src/arrow/engine/substrait/serde.h
```diff
@@ -133,8 +132,7 @@ ARROW_ENGINE_EXPORT Result<std::vector<acero::Declaration>> DeserializePlans(
 /// \param[out] ext_set_out if non-null, the extension mapping used by the Substrait
 /// Plan is returned here.
 /// \param[in] conversion_options options to control how the conversion is to be done.
-/// \return a vector of ExecNode declarations, one for each toplevel relation in the
-/// Substrait Plan
+/// \return an ExecPlan for the Substrait Plan
```
icexelloss, line 135: Revert this?

cpp/src/arrow/engine/substrait/function_test.cc
icexelloss: Can you add some comment to make this more clear? Just looking at
```
  // The aggregation is ordered
  bool ordered = false;
```
It's kind of confusing

icexelloss:
@rtpsw Left some comments

icexelloss (changes requested):
Overall looks good. Left some comments.

cpp/src/arrow/engine/substrait/function_test.cc
```diff
@@ -601,6 +603,7 @@ std::shared_ptr<Table> GetOutputTableForAggregateCase(
 std::shared_ptr<acero::ExecPlan> PlanFromAggregateCase(
     const AggregateTestCase& test_case, std::shared_ptr<Table>* output_table,
     bool with_keys) {
+  static ExecContext ctx_for_ordered;
```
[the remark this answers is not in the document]
rtpsw: The `static` makes the `ExecContext` instance global (and restricted to the function's scope), preventing its premature destruction. This is useful because the method it is passed to takes a pointer, rather than a `shared_ptr`, to an `ExecContext`. This instance is only used when the aggregation-case is ordered.

cpp/src/arrow/engine/substrait/serde.h
rtpsw: The pre-PR comment is incorrect, so I fixed it.

cpp/src/arrow/engine/substrait/serde.h
rtpsw: The pre-PR comment is incorrect, so I fixed it.

cpp/src/arrow/engine/substrait/function_test.cc
```diff
@@ -626,10 +629,22 @@ std::shared_ptr<acero::ExecPlan> PlanFromAggregateCase(
...
+  struct TestConsumerFactory {
+    std::shared_ptr<acero::SinkNodeConsumer> operator()() { return consumer; }
+    std::shared_ptr<acero::SinkNodeConsumer> consumer;
+  };
   EXPECT_OK_AND_ASSIGN(
-      std::shared_ptr<acero::ExecPlan> plan,
-      DeserializePlan(*substrait, std::move(consumer), default_extension_id_registry(),
-                      /*ext_set_out=*/nullptr, conversion_options));
+      auto declarations,
+      DeserializePlans(*substrait, TestConsumerFactory{std::move(consumer)},
```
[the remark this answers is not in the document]
rtpsw: No, because the function takes a `ConsumerFactory` instance, and a lambda won't compile here.

westonpace:
> You're right, this PR is ahead of Ibis-Substrait. @icexelloss, should we create a corresponding PR for Ibis-Substrait?

Wouldn't the PR need to be in Substrait itself?  If you want to use https://github.com/substrait-io/substrait/blob/main/extensions/functions_aggregate_generic.yaml then it would seem that file is the one that needs to change.

westonpace:
ibis-substrait and arrow have to agree on function names (and by name I mean both the URI and the name).

substrait-io/substrait is the main repository for substrait and defines a standard set of functions that are likely to be interesting to all producers and consumers.  I do think first/last belong in that list, however, there may be some discussion to figure out exactly how to classify them (no existing aggregate functions today depend on order so are these aggregate functions or some kind of special new function).

So far, to my understanding, functions have fallen into two categories:

 * Well defined functions
 
 These functions are defined in substrait-io/substrait.  They have "official" names that are part of the Substrait spec and should be long standing.  In this case, agreement is easy, official names should always be preferred.  To use an official name it is important that Arrow has a mapping that maps the official name to the Arrow function name.

 * Arrow-specific functions

These functions are available in Arrow, but not defined yet in substrait-io/substrait.  Since there is no official name we instead use a special URI `urn:arrow:substrait_simple_extension_function` with the Arrow function name.  One current limitation of Arrow-specific functions is that we cannot specify options.  However, one benefit is that no mapping is required.  The function name in the Substrait plan should always match the Arrow function name exactly.

I'm confused by this PR.  It is adding an Arrow mapping, but there is no official function (yet).  It will "work" if both ibis-substrait & arrow act like it's already in Substrait but I feel this is going to be confusing to people.  Anyone looking at the Arrow repo, for example, might think that there really is an official first/last function.  I think it's best to use the arrow-specific URI until the official function is adopted.

icexelloss:
Thanks @westonpace for catching this and explain in details. I agree with you.

To be clear, is the ask to use "urn:arrow:substrait_simple_extension_function" instead of "https://github.com/substrait-io/substrait/blob/main/extensions/functions_aggregate_generic.yaml" for the function URI?

cpp/src/arrow/engine/substrait/function_test.cc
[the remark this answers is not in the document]
icexelloss: This still feel strange to me that this is needed. Does something like this not work?
https://github.com/apache/arrow/blob/401ae19051e9120fedb8372a13cd68e08798c00a/cpp/src/arrow/acero/benchmark_util.cc#L49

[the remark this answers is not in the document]
icexelloss: Looks like you can use this one?
https://github.com/apache/arrow/blob/main/cpp/src/arrow/acero/exec_plan.h#L72

westonpace:
> To double check, is the ask to use "urn:arrow:substrait_simple_extension_function" instead of "https://github.com/substrait-io/substrait/blob/main/extensions/functions_aggregate_generic.yaml" for the function URI?

Yes.  Furthermore, if you use `urn:arrow:substrait_simple_extension_function` then you don't need a mapping at all.

However, since this is an aggregate function, I suspect we will need to fix `arrow::engine::ExtensionIdRegistry::SubstraitAggregateToArrow` so that it behaves similarly to `arrow::engine::ExtensionIdRegistry::SubstraitCallToArrow`:

```
  Result<SubstraitCallToArrow> GetSubstraitCallToArrow(
      Id substrait_function_id) const override {
    // NEED TO ADD THIS CASE TO GetSubstraitAggregateToArrow
    if (substrait_function_id.uri == kArrowSimpleExtensionFunctionsUri) {
      return kSimpleSubstraitToArrow;
    }
    auto maybe_converter = substrait_to_arrow_.find(substrait_function_id);
    if (maybe_converter == substrait_to_arrow_.end()) {
      if (parent_) {
        return parent_->GetSubstraitCallToArrow(substrait_function_id);
      }
      return Status::NotImplemented(
          "No conversion function exists to convert the Substrait function ",
          substrait_function_id.uri, "#", substrait_function_id.name,
          " to an Arrow call expression");
    }
    return maybe_converter->second;
  }
```

rtpsw:
Thanks, @westonpace. I'll work on a revision accordingly.

cpp/src/arrow/engine/substrait/extension_set.cc
```diff
@@ -1113,9 +1113,11 @@ struct DefaultExtensionIdRegistry : ExtensionIdRegistryImpl {
                                            DecodeBasicAggregate("mean")));
     DCHECK_OK(AddSubstraitAggregateToArrow({kSubstraitArithmeticFunctionsUri, "std_dev"},
                                            DecodeBasicAggregate("stddev")));
-    DCHECK_OK(
-        AddSubstraitAggregateToArrow({kSubstraitAggregateGenericFunctionsUri, "count"},
-                                     DecodeBasicAggregate("count")));
+    for (const auto& fn_name : {"count", "first", "last"}) {
+      DCHECK_OK(
+          AddSubstraitAggregateToArrow({kSubstraitAggregateGenericFunctionsUri, fn_name},
```
icexelloss, line 1118: Should we uri the arrow specifc uri for first/last instead of the substrait one?

rtpsw: Right. Fixed.

cpp/src/arrow/engine/substrait/function_test.cc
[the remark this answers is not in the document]
rtpsw: OK, I did something similar.

cpp/src/arrow/engine/substrait/serde.h
icexelloss: Oh you are right I am looking at the wrong thing

cpp/src/arrow/engine/substrait/function_test.cc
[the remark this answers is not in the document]
rtpsw: Here are [a couple of instances](https://github.com/search?q=repo%3Aapache%2Farrow%20%22static%20ExecContext%22&type=code) where this is used.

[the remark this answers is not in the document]
icexelloss: Ok since there is existing usage of this pattern I am ok with it.

cpp/src/arrow/engine/substrait/function_test.cc
rtpsw: Clarified.

icexelloss (approved):
LGTM

icexelloss:
@westonpace I am gonna go head and merge this one. If you have any comments please let @rtpsw know and he can address as follow up.

PR #35514, opened 2023-05-09
GH-35515: [C++][Python] Add non decomposable aggregation UDF

### LINKED ISSUE #35515: [C++][Python]  Non decomposable aggregation UDF

Non decomposable aggregation is aggregation that cannot be split into consume/merge/finalize. This is often when the logic rewritten with external python libraries (numpy, pandas, statmodels, etc) and those either cannot be decomposed or not worthy the effect (these are often one-off function instead of reusable one). This PR implements the support for non decomposable aggregation UDFs.

The major issue with non decomposable UDF is that the UDF needs to see all data at once, unlike scalar UDF where UDF only needs to see a batch at a time. This makes non decomposable not so useful as it is same as collect all the data to a pd.DataFrame and apply the UDF on it. However, one very application of non decomposable UDF is with segmented aggregation. To refresh, segmented aggregation works on ordered data and passed one logic chunk at a time (e.g., all data with the same date). With segmented aggregation and non decomposable aggregation UDF, the user can apply any custom aggregation logic over large stream of ordered data, with the memory overhead of a single segment.

### Component(s)

C++, Python

icexelloss:
### Rationale for this change
Non decomposable aggregation is aggregation that cannot be split into consume/merge/finalize. This is often when the logic rewritten with external python libraries (numpy, pandas, statmodels, etc) and those either cannot be decomposed or not worthy the effect (these are often one-off function instead of reusable one). This PR implements the support for non decomposable aggregation UDFs.

The major issue with non decomposable UDF is that the UDF needs to see all data at once, unlike scalar UDF where UDF only needs to see a batch at a time. This makes non decomposable not so useful as it is same as collect all the data to a pd.DataFrame and apply the UDF on it. However, one very application of non decomposable UDF is with segmented aggregation. To refresh, segmented aggregation works on ordered data and passed one logic chunk at a time (e.g., all data with the same date). With segmented aggregation and non decomposable aggregation UDF, the user can apply any custom aggregation logic over large stream of ordered data, with the memory overhead of a single segment.

This PR is currently WIP and not ready for review.

So far I have implemented the minimal amount of code to make a basic test working but needs clean up, error handling etc.

* [x] First round of self review
* [x] Second round of self review
* [x] Implement and test unary
* [x] Implement and test varargs
* [x] Implement and test Acero support with segmented aggregation

### Are these changes tested?
Added new test calling with compute and acero.

The compute tests calls the aggregation on the full array. The acero test callings the aggregation with segmented aggregation.

* Closes: #35515

### DIFF
--- a/cpp/src/arrow/engine/substrait/extension_set.cc
+++ b/cpp/src/arrow/engine/substrait/extension_set.cc
@@ -417,8 +417,21 @@ ExtensionIdRegistry::SubstraitAggregateToArrow DecodeBasicAggregate(
         return compute::Aggregate{std::move(fixed_arrow_func),
                                   options ? std::move(options) : nullptr, *arg_ref, ""};
       }
-      default:
-        break;
+      default: {
+
+        fixed_arrow_func += arrow_function_name;
+        std::vector<FieldRef> target;
+        for (int i = 0; i < call.size(); i++) {
+          ARROW_ASSIGN_OR_RAISE(compute::Expression arg, call.GetValueArg(i));
+          const FieldRef* arg_ref = arg.field_ref();
+          if (!arg_ref) {
+          return Status::Invalid("Expected an aggregate call ", call.id().uri, "#",
+                                 call.id().name, " to have a direct reference");
+          }
+          target.emplace_back(*arg_ref);
+        }
+        return compute::Aggregate{std::move(fixed_arrow_func), nullptr, target, ""};
+      }
     }
     return Status::NotImplemented(
         "Only nullary and unary aggregate functions are currently supported");
--- a/cpp/src/arrow/engine/substrait/extension_set.cc
+++ b/cpp/src/arrow/engine/substrait/extension_set.cc
@@ -981,12 +981,12 @@ ExtensionIdRegistry::SubstraitAggregateToArrow DecodeBasicAggregate(
             options = std::make_shared<compute::VarianceOptions>(ddof);
           }
         }
-
         fixed_arrow_func += arrow_function_name;
+
         std::vector<FieldRef> target;
         for (int i = 0; i < call.size(); i++) {
           ARROW_ASSIGN_OR_RAISE(compute::Expression arg, call.GetValueArg(i));
-          const FieldRef* arg_ref = arg.field_ref();
+          FieldRef* arg_ref = arg.field_ref();
           if (!arg_ref) {
             return Status::Invalid("Expected an aggregate call ", call.id().uri, "#",
                                    call.id().name, " to have a direct reference");
@@ -994,7 +994,7 @@ ExtensionIdRegistry::SubstraitAggregateToArrow DecodeBasicAggregate(
           target.emplace_back(std::move(*arg_ref));
         }
         return compute::Aggregate{std::move(fixed_arrow_func),
-                                  options ? std::move(options) : nullptr, target, ""};
+                                  options ? std::move(options) : nullptr, std::move(target), ""};
       }
     }
   };
--- a/cpp/src/arrow/engine/substrait/extension_set.cc
+++ b/cpp/src/arrow/engine/substrait/extension_set.cc
@@ -363,6 +363,68 @@ ExtensionIdRegistry::SubstraitAggregateToArrow kSimpleSubstraitAggregateToArrow 
   return DecodeBasicAggregate(std::string(call.id().name))(call);
 };
 
+ExtensionIdRegistry::SubstraitAggregateToArrow DecodeBasicAggregate(
+    const std::string& arrow_function_name) {
+  return [arrow_function_name](const SubstraitCall& call) -> Result<compute::Aggregate> {
+    std::string fixed_arrow_func;
+    if (call.is_hash()) {
+      fixed_arrow_func = "hash_";
+    }
+
+    switch (call.size()) {
+      case 0: {
+        if (call.id().name == "count") {
+          fixed_arrow_func += "count_all";
+          return compute::Aggregate{std::move(fixed_arrow_func), ""};
+        }
+        return Status::Invalid("Expected aggregate call ", call.id().uri, "#",
+                               call.id().name, " to have at least one argument");
+      }
+      case 1: {
+        std::shared_ptr<compute::FunctionOptions> options = nullptr;
+        if (arrow_function_name == "stddev" || arrow_function_name == "variance") {
+          // See the following URL for the spec of stddev and variance:
+          // https://github.com/substrait-io/substrait/blob/
+          // 73228b4112d79eb1011af0ebb41753ce23ca180c/
+          // extensions/functions_arithmetic.yaml#L1240
+          auto maybe_dist = call.GetOption("distribution");
+          if (maybe_dist) {
+            auto& prefs = **maybe_dist;
+            if (prefs.size() != 1) {
+              return Status::Invalid("expected a single preference for ",
+                                     arrow_function_name, " but got ", prefs.size());
+            }
+            int ddof;
+            if (prefs[0] == "POPULATION") {
+              ddof = 1;
+            } else if (prefs[0] == "SAMPLE") {
+              ddof = 0;
+            } else {
+              return Status::Invalid("unknown distribution preference ", prefs[0]);
+            }
+            options = std::make_shared<compute::VarianceOptions>(ddof);
+          }
+        }
+        fixed_arrow_func += arrow_function_name;
+
+        ARROW_ASSIGN_OR_RAISE(compute::Expression arg, call.GetValueArg(0));
+        const FieldRef* arg_ref = arg.field_ref();
+        if (!arg_ref) {
+          return Status::Invalid("Expected an aggregate call ", call.id().uri, "#",
+                                 call.id().name, " to have a direct reference");
+        }
+
+        return compute::Aggregate{std::move(fixed_arrow_func),
+                                  options ? std::move(options) : nullptr, *arg_ref, ""};
+      }
+      default:
+        break;
+    }
+    return Status::NotImplemented(
+        "Only nullary and unary aggregate functions are currently supported");
+  };
+}
+
 struct ExtensionIdRegistryImpl : ExtensionIdRegistry {
   ExtensionIdRegistryImpl() : parent_(nullptr) {}
   explicit ExtensionIdRegistryImpl(const ExtensionIdRegistry* parent) : parent_(parent) {}
@@ -934,68 +996,6 @@ ExtensionIdRegistry::SubstraitCallToArrow DecodeConcatMapping() {
                           GetValueArgs(call, 0));
     value_args.push_back(compute::literal(""));
     return compute::call("binary_join_element_wise", std::move(value_args));
-  };
-}
-
-ExtensionIdRegistry::SubstraitAggregateToArrow DecodeBasicAggregate(
-    const std::string& arrow_function_name) {
-  return [arrow_function_name](const SubstraitCall& call) -> Result<compute::Aggregate> {
-    std::string fixed_arrow_func;
-    if (call.is_hash()) {
-      fixed_arrow_func = "hash_";
-    }
-
-    switch (call.size()) {
-      case 0: {
-        if (call.id().name == "count") {
-          fixed_arrow_func += "count_all";
-          return compute::Aggregate{std::move(fixed_arrow_func), ""};
-        }
-        return Status::Invalid("Expected aggregate call ", call.id().uri, "#",
-                               call.id().name, " to have at least one argument");
-      }
-      case 1: {
-        std::shared_ptr<compute::FunctionOptions> options = nullptr;
-        if (arrow_function_name == "stddev" || arrow_function_name == "variance") {
-          // See the following URL for the spec of stddev and variance:
-          // https://github.com/substrait-io/substrait/blob/
-          // 73228b4112d79eb1011af0ebb41753ce23ca180c/
-          // extensions/functions_arithmetic.yaml#L1240
-          auto maybe_dist = call.GetOption("distribution");
-          if (maybe_dist) {
-            auto& prefs = **maybe_dist;
-            if (prefs.size() != 1) {
-              return Status::Invalid("expected a single preference for ",
-                                     arrow_function_name, " but got ", prefs.size());
-            }
-            int ddof;
-            if (prefs[0] == "POPULATION") {
-              ddof = 1;
-            } else if (prefs[0] == "SAMPLE") {
-              ddof = 0;
-            } else {
-              return Status::Invalid("unknown distribution preference ", prefs[0]);
-            }
-            options = std::make_shared<compute::VarianceOptions>(ddof);
-          }
-        }
-        fixed_arrow_func += arrow_function_name;
-
-        ARROW_ASSIGN_OR_RAISE(compute::Expression arg, call.GetValueArg(0));
-        const FieldRef* arg_ref = arg.field_ref();
-        if (!arg_ref) {
-          return Status::Invalid("Expected an aggregate call ", call.id().uri, "#",
-                                 call.id().name, " to have a direct reference");
-        }
-
-        return compute::Aggregate{std::move(fixed_arrow_func),
-                                  options ? std::move(options) : nullptr, *arg_ref, ""};
-      }
-      default:
-        break;
-    }
-    return Status::NotImplemented(
-        "Only nullary and unary aggregate functions are currently supported");
   };
 }
 
--- a/cpp/src/arrow/engine/substrait/extension_set.cc
+++ b/cpp/src/arrow/engine/substrait/extension_set.cc
@@ -986,15 +986,17 @@ ExtensionIdRegistry::SubstraitAggregateToArrow DecodeBasicAggregate(
         std::vector<FieldRef> target;
         for (int i = 0; i < call.size(); i++) {
           ARROW_ASSIGN_OR_RAISE(compute::Expression arg, call.GetValueArg(i));
-          FieldRef* arg_ref = arg.field_ref();
+          const FieldRef* arg_ref = arg.field_ref();
           if (!arg_ref) {
             return Status::Invalid("Expected an aggregate call ", call.id().uri, "#",
                                    call.id().name, " to have a direct reference");
           }
-          target.emplace_back(std::move(*arg_ref));
+          // Copy arg_ref here because field_ref() return const FieldRef*
+          target.emplace_back(*arg_ref);
         }
         return compute::Aggregate{std::move(fixed_arrow_func),
-                                  options ? std::move(options) : nullptr, std::move(target), ""};
+                                  options ? std::move(options) : nullptr,
+                                  std::move(target), ""};
       }
     }
   };
--- a/cpp/src/arrow/engine/substrait/extension_set.cc
+++ b/cpp/src/arrow/engine/substrait/extension_set.cc
@@ -363,81 +363,6 @@ ExtensionIdRegistry::SubstraitAggregateToArrow kSimpleSubstraitAggregateToArrow 
   return DecodeBasicAggregate(std::string(call.id().name))(call);
 };
 
-ExtensionIdRegistry::SubstraitAggregateToArrow DecodeBasicAggregate(
-    const std::string& arrow_function_name) {
-  return [arrow_function_name](const SubstraitCall& call) -> Result<compute::Aggregate> {
-    std::string fixed_arrow_func;
-    if (call.is_hash()) {
-      fixed_arrow_func = "hash_";
-    }
-
-    switch (call.size()) {
-      case 0: {
-        if (call.id().name == "count") {
-          fixed_arrow_func += "count_all";
-          return compute::Aggregate{std::move(fixed_arrow_func), ""};
-        }
-        return Status::Invalid("Expected aggregate call ", call.id().uri, "#",
-                               call.id().name, " to have at least one argument");
-      }
-      case 1: {
-        std::shared_ptr<compute::FunctionOptions> options = nullptr;
-        if (arrow_function_name == "stddev" || arrow_function_name == "variance") {
-          // See the following URL for the spec of stddev and variance:
-          // https://github.com/substrait-io/substrait/blob/
-          // 73228b4112d79eb1011af0ebb41753ce23ca180c/
-          // extensions/functions_arithmetic.yaml#L1240
-          auto maybe_dist = call.GetOption("distribution");
-          if (maybe_dist) {
-            auto& prefs = **maybe_dist;
-            if (prefs.size() != 1) {
-              return Status::Invalid("expected a single preference for ",
-                                     arrow_function_name, " but got ", prefs.size());
-            }
-            int ddof;
-            if (prefs[0] == "POPULATION") {
-              ddof = 1;
-            } else if (prefs[0] == "SAMPLE") {
-              ddof = 0;
-            } else {
-              return Status::Invalid("unknown distribution preference ", prefs[0]);
-            }
-            options = std::make_shared<compute::VarianceOptions>(ddof);
-          }
-        }
-        fixed_arrow_func += arrow_function_name;
-
-        ARROW_ASSIGN_OR_RAISE(compute::Expression arg, call.GetValueArg(0));
-        const FieldRef* arg_ref = arg.field_ref();
-        if (!arg_ref) {
-          return Status::Invalid("Expected an aggregate call ", call.id().uri, "#",
-                                 call.id().name, " to have a direct reference");
-        }
-
-        return compute::Aggregate{std::move(fixed_arrow_func),
-                                  options ? std::move(options) : nullptr, *arg_ref, ""};
-      }
-      default: {
-
-        fixed_arrow_func += arrow_function_name;
-        std::vector<FieldRef> target;
-        for (int i = 0; i < call.size(); i++) {
-          ARROW_ASSIGN_OR_RAISE(compute::Expression arg, call.GetValueArg(i));
-          const FieldRef* arg_ref = arg.field_ref();
-          if (!arg_ref) {
-          return Status::Invalid("Expected an aggregate call ", call.id().uri, "#",
-                                 call.id().name, " to have a direct reference");
-          }
-          target.emplace_back(*arg_ref);
-        }
-        return compute::Aggregate{std::move(fixed_arrow_func), nullptr, target, ""};
-      }
-    }
-    return Status::NotImplemented(
-        "Only nullary and unary aggregate functions are currently supported");
-  };
-}
-
 struct ExtensionIdRegistryImpl : ExtensionIdRegistry {
   ExtensionIdRegistryImpl() : parent_(nullptr) {}
   explicit ExtensionIdRegistryImpl(const ExtensionIdRegistry* parent) : parent_(parent) {}
@@ -1009,6 +934,68 @@ ExtensionIdRegistry::SubstraitCallToArrow DecodeConcatMapping() {
                           GetValueArgs(call, 0));
     value_args.push_back(compute::literal(""));
     return compute::call("binary_join_element_wise", std::move(value_args));
+  };
+}
+
+ExtensionIdRegistry::SubstraitAggregateToArrow DecodeBasicAggregate(
+    const std::string& arrow_function_name) {
+  return [arrow_function_name](const SubstraitCall& call) -> Result<compute::Aggregate> {
+    std::string fixed_arrow_func;
+    if (call.is_hash()) {
+      fixed_arrow_func = "hash_";
+    }
+
+    switch (call.size()) {
+      case 0: {
+        if (call.id().name == "count") {
+          fixed_arrow_func += "count_all";
+          return compute::Aggregate{std::move(fixed_arrow_func), ""};
+        }
+        return Status::Invalid("Expected aggregate call ", call.id().uri, "#",
+                               call.id().name, " to have at least one argument");
+      }
+      default: {
+        // Handles all arity > 0
+
+        std::shared_ptr<compute::FunctionOptions> options = nullptr;
+        if (arrow_function_name == "stddev" || arrow_function_name == "variance") {
+          // See the following URL for the spec of stddev and variance:
+          // https://github.com/substrait-io/substrait/blob/
+          // 73228b4112d79eb1011af0ebb41753ce23ca180c/
+          // extensions/functions_arithmetic.yaml#L1240
+          auto maybe_dist = call.GetOption("distribution");
+          if (maybe_dist) {
+            auto& prefs = **maybe_dist;
+            if (prefs.size() != 1) {
+              return Status::Invalid("expected a single preference for ",
+                                     arrow_function_name, " but got ", prefs.size());
+            }
+            int ddof;
+            if (prefs[0] == "POPULATION") {
+              ddof = 1;
+            } else if (prefs[0] == "SAMPLE") {
+              ddof = 0;
+            } else {
+              return Status::Invalid("unknown distribution preference ", prefs[0]);
+            }
+            options = std::make_shared<compute::VarianceOptions>(ddof);
+          }
+        }
+
+        fixed_arrow_func += arrow_function_name;
+        std::vector<FieldRef> target;
+        for (int i = 0; i < call.size(); i++) {
+          ARROW_ASSIGN_OR_RAISE(compute::Expression arg, call.GetValueArg(i));
+          const FieldRef* arg_ref = arg.field_ref();
+          if (!arg_ref) {
+            return Status::Invalid("Expected an aggregate call ", call.id().uri, "#",
+                                   call.id().name, " to have a direct reference");
+          }
+          target.emplace_back(*arg_ref);
+        }
+        return compute::Aggregate{std::move(fixed_arrow_func), nullptr, target, ""};
+      }
+    }
   };
 }
 
--- a/cpp/src/arrow/engine/substrait/extension_set.cc
+++ b/cpp/src/arrow/engine/substrait/extension_set.cc
@@ -991,9 +991,10 @@ ExtensionIdRegistry::SubstraitAggregateToArrow DecodeBasicAggregate(
             return Status::Invalid("Expected an aggregate call ", call.id().uri, "#",
                                    call.id().name, " to have a direct reference");
           }
-          target.emplace_back(*arg_ref);
+          target.emplace_back(std::move(*arg_ref));
         }
-        return compute::Aggregate{std::move(fixed_arrow_func), nullptr, target, ""};
+        return compute::Aggregate{std::move(fixed_arrow_func),
+                                  options ? std::move(options) : nullptr, target, ""};
       }
     }
   };
--- a/python/pyarrow/_compute.pyx
+++ b/python/pyarrow/_compute.pyx
@@ -2676,6 +2676,8 @@ def register_scalar_function(func, function_name, function_doc, in_types, out_ty
     """
     Register a user-defined scalar function.
 
+    This API is EXPERIMENTAL.
+
     A scalar function is a function that executes elementwise
     operations on arrays or scalars, i.e. a scalar function must
     be computed row-by-row with no state where each output row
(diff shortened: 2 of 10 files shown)


python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -65,6 +70,26 @@ struct PythonUdfKernelInit {
   std::shared_ptr<OwnedRefNoGIL> function;
 };
 
+struct ScalarUdfAggregator : public compute::KernelState {
```
icexelloss, line 73: "Scalar" as supposed to the "grouped" aggregator which has difference interface:
https://github.com/apache/arrow/blob/main/cpp/src/arrow/compute/kernels/hash_aggregate.cc#L66

python/pyarrow/tests/test_udf.py
```diff
@@ -39,6 +39,30 @@ class MyError(RuntimeError):
...
 
 
+@pytest.fixture(scope="session")
+def unary_agg_func_fixture():
+    """
+    Register a unary aggregate function
+    """
+
+    def func(ctx, x):
+        return pa.array([len(x)])
```
icexelloss, line 49: This is not the ideal interface - user shouldn't need to wrap the result in an pa.array of one row.

This is currently implemented this way to reuse the code to pass the result back to C++ code in an object that is well understood (array).

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -101,6 +126,98 @@ struct PythonTableUdfKernelInit {
...
+      values.insert(values.end(), other_state.values.begin(), other_state.values.end());
+      return Status::OK();
+    }
+
+    Status Finalize(compute::KernelContext* ctx, Datum* out) {
+      auto state = arrow::internal::checked_cast<PythonUdfScalarAggregatorImpl*>(ctx->state());
+      std::shared_ptr<OwnedRefNoGIL>& function = state->agg_function;
+      const int num_args = input_schema->num_fields();
+      // Ignore batch length here
+      ScalarUdfContext udf_context{ctx->memory_pool(), 0};
```
icexelloss, line 170: TODO: Use the proper UDF Context

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -101,6 +126,97 @@ struct PythonTableUdfKernelInit {
...
+
+    ~PythonUdfScalarAggregatorImpl() {
+      if (_Py_IsFinalizing()) {
+        agg_function->detach();
+      }
+    }
+
+    Status Consume(compute::KernelContext* ctx, const compute::ExecSpan& batch) {
+      num_rows = batch.length;
+      ARROW_ASSIGN_OR_RAISE(auto rb, batch.ToExecBatch().ToRecordBatch(input_schema, ctx->memory_pool()));
```
icexelloss, line 153: Limitation: Needing twice of the memory to concat

cpp/src/arrow/engine/substrait/extension_set.cc
```diff
@@ -355,6 +355,68 @@ ExtensionIdRegistry::SubstraitCallToArrow kSimpleSubstraitToArrow =
   return ::arrow::compute::call(std::string(call.id().name), std::move(args));
 };
 
+ExtensionIdRegistry::SubstraitAggregateToArrow DecodeBasicAggregate(
```
icexelloss, line 358: TODO: Revert this change after https://github.com/apache/arrow/pull/35608 is merged

cpp/src/arrow/engine/substrait/extension_set.cc
```diff
@@ -592,6 +654,10 @@ struct ExtensionIdRegistryImpl : ExtensionIdRegistry {
 
   Result<SubstraitAggregateToArrow> GetSubstraitAggregateToArrow(
       Id substrait_function_id) const override {
+
+    if (substrait_function_id.uri == kArrowSimpleExtensionFunctionsUri) {
```
icexelloss, line 658: TODO: Revert this change after https://github.com/apache/arrow/pull/35608 is merged

cpp/src/arrow/engine/substrait/extension_set.cc
```diff
@@ -926,68 +992,6 @@ ExtensionIdRegistry::SubstraitCallToArrow DecodeConcatMapping() {
   };
 }
 
-ExtensionIdRegistry::SubstraitAggregateToArrow DecodeBasicAggregate(
```
icexelloss, line 929: TODO: Revert this change after https://github.com/apache/arrow/pull/35608 is merged

python/pyarrow/_compute.pyx
```diff
@@ -2743,6 +2749,84 @@ def register_scalar_function(func, function_name, function_doc, in_types, out_ty
                                           out_type, func_registry)
 
 
+def register_aggregate_function(func, function_name, function_doc, in_types, out_type,
+                                func_registry=None):
+    """
+    Register a user-defined scalar function.
```
icexelloss, line 2755: Fix doc

python/pyarrow/_compute.pyx
```diff
@@ -2743,6 +2749,84 @@ def register_scalar_function(func, function_name, function_doc, in_types, out_ty
...
+    >>> func.name
+    'py_add_func'
+    >>> answer = pc.call_function(func_name, [pa.array([20])])
+    >>> answer
+    <pyarrow.lib.Int64Array object at ...>
+    [
+      21
+    ]
+    """
+    return _register_scalar_like_function(get_register_aggregate_function(),
```
icexelloss, line 2825: Rename this - "scalar like" is no longer accurate

python/pyarrow/conftest.py
```diff
@@ -278,3 +278,29 @@ def unary_function(ctx, x):
                                 {"array": pa.int64()},
                                 pa.int64())
     return unary_function, func_name
+
+
+@pytest.fixture(scope="session")
+def unary_agg_func_fixture():
```
icexelloss, line 284: This is used in both test_udf.py and test_substrait.py

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -101,6 +126,99 @@ struct PythonTableUdfKernelInit {
...
+        PyObject* data = wrap_array(c_data);
+        PyTuple_SetItem(arg_tuple.obj(), arg_id, data);
+      }
+
+      OwnedRef result(agg_cb(function->obj(), udf_context, arg_tuple.obj()));
+      RETURN_NOT_OK(CheckPyError());
+
+      // unwrapping the output for expected output type
+      if (is_array(result.obj())) {
+        ARROW_ASSIGN_OR_RAISE(std::shared_ptr<Array> val, unwrap_array(result.obj()));
```
icexelloss, line 195: TODO: Change to use Scalar instead of array

python/pyarrow/tests/test_substrait.py
```diff
@@ -436,6 +436,7 @@ def table_provider(names, _):
     """
 
     buf = pa._substrait._parse_json_plan(substrait_query)
+    breakpoint()
```
icexelloss, line 439: remove this

python/pyarrow/tests/test_udf.py
```diff
@@ -228,11 +253,11 @@ def check_scalar_function(func_fixture,
         if all_scalar:
             batch_length = 1
 
-    expected_output = function(mock_scalar_udf_context(batch_length), *inputs)
     func = pc.get_function(name)
     assert func.name == name
 
     result = pc.call_function(name, inputs, length=batch_length)
+    expected_output = function(mock_scalar_udf_context(batch_length), *inputs)
```
icexelloss, line 260: Move this down to compute "result" first then "expected", which is how normally this is done in test

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -101,6 +126,99 @@ struct PythonTableUdfKernelInit {
...
+        return Status::OK();
+      } else {
+        return Status::TypeError("Unexpected output type: ", Py_TYPE(result.obj())->tp_name,
+                                 " (expected Array)");
+      }
+      // *out = Datum((int32_t)table->num_rows());
+      return Status::OK();
+    }
+
+    int32_t num_rows = 0;
```
icexelloss, line 214: Remove this

cpp/src/arrow/engine/substrait/extension_set.cc
```diff
@@ -954,7 +954,9 @@ ExtensionIdRegistry::SubstraitAggregateToArrow DecodeBasicAggregate(
         return Status::Invalid("Expected aggregate call ", call.id().uri, "#",
                                call.id().name, " to have at least one argument");
       }
-      case 1: {
+      default: {
```
icexelloss, line 957: This is so that we can support UDFs which can have arity > 1

cpp/src/arrow/engine/substrait/extension_set.cc
```diff
@@ -979,23 +981,21 @@ ExtensionIdRegistry::SubstraitAggregateToArrow DecodeBasicAggregate(
...
+        fixed_arrow_func += arrow_function_name;
+        std::vector<FieldRef> target;
+        for (int i = 0; i < call.size(); i++) {
+          ARROW_ASSIGN_OR_RAISE(compute::Expression arg, call.GetValueArg(i));
+          const FieldRef* arg_ref = arg.field_ref();
+          if (!arg_ref) {
+            return Status::Invalid("Expected an aggregate call ", call.id().uri, "#",
+                                   call.id().name, " to have a direct reference");
+          }
+          target.emplace_back(*arg_ref);
```
icexelloss, line 994: Use std::move here to avoid copy

python/pyarrow/_compute.pyx
```diff
@@ -2738,9 +2744,83 @@ def register_scalar_function(func, function_name, function_doc, in_types, out_ty
...
+    ----------
+    func : callable
+        A callable implementing the user-defined function.
+        The first argument is the context argument of type
+        UdfContext.
+        Then, it must take arguments equal to the number of
+        in_types defined. It must return Scalar matching the
+        out_type.
+
+        To define a varargs function, pass a callable that takes
```
icexelloss, line 2775: TODO: Double check the doc for varargs functions

python/pyarrow/tests/test_substrait.py
```diff
@@ -605,3 +605,151 @@ def table_provider(names, schema):
     expected = pa.Table.from_pydict({"out": [1, 2, 3]})
 
     assert res_tb == expected
+
+
+def test_aggregate_udf_basic(varargs_agg_func_fixture):
```
icexelloss, line 610: This test case matches how we would this it internally. The end user would define sth like
```
def foo(v: pd.Series):
    return np.nanmean(v)

summarize(table, agg=foo, columns=['v'], by='time')
```

We would then wrap the foo into a function that Acero is expecting (a varargs UDF)
```
def get_acero_func(func):
      def acero_func(ctx, *args):
            return pa.scalar(func(*[arg.to_pandas() for arg in args]))
```
And also register it in Acero on the fly.

python/pyarrow/tests/test_udf.py
icexelloss: Fixed

cpp/src/arrow/engine/substrait/extension_set.cc
icexelloss: Done

cpp/src/arrow/engine/substrait/extension_set.cc
icexelloss: Done

cpp/src/arrow/engine/substrait/extension_set.cc
icexelloss: Done

python/pyarrow/_compute.pyx
icexelloss: Done

cpp/src/arrow/engine/substrait/extension_set.cc
```diff
@@ -979,23 +981,21 @@ ExtensionIdRegistry::SubstraitAggregateToArrow DecodeBasicAggregate(
...
         }
-
-        return compute::Aggregate{std::move(fixed_arrow_func),
-                                  options ? std::move(options) : nullptr, *arg_ref, ""};
+        return compute::Aggregate{std::move(fixed_arrow_func), nullptr, target, ""};
```
icexelloss, line 996: TODO: Fix options

icexelloss:
@westonpace I think the PR is in a good shape for review. I will be out until 5/28 so please take your time to review this and I will address them when I get back. If there are other people that you think should be on the reviewer list I can ping them as well.

icexelloss:
@westonpace any chance you can take a look at this? Or maybe some else can help review?

westonpace:
> @westonpace any chance you can take a look at this? Or maybe some else can help review?

I'll commit to looking at this tonight.

westonpace:
Also, it may be interesting to @felipecrv who has played around with the kernels some.

cpp/src/arrow/engine/substrait/extension_set.cc
```diff
@@ -979,23 +981,22 @@ ExtensionIdRegistry::SubstraitAggregateToArrow DecodeBasicAggregate(
...
+          if (!arg_ref) {
+            return Status::Invalid("Expected an aggregate call ", call.id().uri, "#",
+                                   call.id().name, " to have a direct reference");
+          }
+          target.emplace_back(std::move(*arg_ref));
         }
-
         return compute::Aggregate{std::move(fixed_arrow_func),
-                                  options ? std::move(options) : nullptr, *arg_ref, ""};
+                                  options ? std::move(options) : nullptr, target, ""};
```
felipecrv, line 997: Is `std::move(target)` possible here?

cpp/src/arrow/engine/substrait/extension_set.cc
felipecrv: Yes

cpp/src/arrow/engine/substrait/extension_set.cc
```diff
@@ -979,23 +981,22 @@ ExtensionIdRegistry::SubstraitAggregateToArrow DecodeBasicAggregate(
...
           }
         }
-        fixed_arrow_func += arrow_function_name;
 
-        ARROW_ASSIGN_OR_RAISE(compute::Expression arg, call.GetValueArg(0));
-        const FieldRef* arg_ref = arg.field_ref();
-        if (!arg_ref) {
-          return Status::Invalid("Expected an aggregate call ", call.id().uri, "#",
-                                 call.id().name, " to have a direct reference");
+        fixed_arrow_func += arrow_function_name;
```
felipecrv, line 985: This line was glued to the previous block before it's part of the function name building process.

cpp/src/arrow/engine/substrait/extension_set.cc
```diff
@@ -979,23 +981,22 @@ ExtensionIdRegistry::SubstraitAggregateToArrow DecodeBasicAggregate(
...
+        std::vector<FieldRef> target;
+        for (int i = 0; i < call.size(); i++) {
+          ARROW_ASSIGN_OR_RAISE(compute::Expression arg, call.GetValueArg(i));
+          const FieldRef* arg_ref = arg.field_ref();
```
felipecrv, line 989: Remove `const` so it can be moved below.

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -234,6 +351,56 @@ Status RegisterTabularFunction(PyObject* user_function, UdfWrapperCallback wrapp
...
+
+  if (registry == NULLPTR) {
+    registry = compute::GetFunctionRegistry();
+  }
+
+  static auto default_scalar_aggregate_options = compute::ScalarAggregateOptions::Defaults();
+  auto aggregate_func = std::make_shared<compute::ScalarAggregateFunction>(
+      options.func_name, options.arity, options.func_doc, &default_scalar_aggregate_options);
+
+  Py_INCREF(agg_function);
```
westonpace, line 377: Why is this needed?  Can we wrap `agg_function` in an `OwnedRef` instead?

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -101,6 +125,99 @@ struct PythonTableUdfKernelInit {
...
+        ARROW_ASSIGN_OR_RAISE(std::shared_ptr<Scalar> val, unwrap_scalar(result.obj()));
+        if (*output_type != *val->type) {
+          return Status::TypeError("Expected output datatype ", output_type->ToString(),
+                                   ", but function returned datatype ",
+                                   val->type->ToString());
+        }
+        out->value = std::move(val);
+        return Status::OK();
+      } else {
+        return Status::TypeError("Unexpected output type: ", Py_TYPE(result.obj())->tp_name,
```
westonpace, line 208: Can we add a test case for this?

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -101,6 +125,99 @@ struct PythonTableUdfKernelInit {
...
+      }
+
+      OwnedRef result(agg_cb(function->obj(), udf_context, arg_tuple.obj()));
+      RETURN_NOT_OK(CheckPyError());
+
+      // unwrapping the output for expected output type
+      if (is_scalar(result.obj())) {
+        ARROW_ASSIGN_OR_RAISE(std::shared_ptr<Scalar> val, unwrap_scalar(result.obj()));
+        if (*output_type != *val->type) {
+          return Status::TypeError("Expected output datatype ", output_type->ToString(),
```
westonpace, line 201: Can we add a test case for this?

westonpace (changes requested):
A few suggestions but overall this seems pretty clear.  I will refer you to https://github.com/apache/arrow/pull/14320#issuecomment-1372473478 which I found to be a pretty compelling suggestion for a better API for registering UDFs.

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -65,6 +69,26 @@ struct PythonUdfKernelInit {
...
+  virtual Status Finalize(compute::KernelContext* ctx, Datum* out) = 0;
+};
+
+arrow::Status AggregateUdfConsume(compute::KernelContext* ctx, const compute::ExecSpan& batch) {
+  return checked_cast<ScalarUdfAggregator*>(ctx->state())->Consume(ctx, batch);
+}
+
+arrow::Status AggregateUdfMerge(compute::KernelContext* ctx, compute::KernelState&& src,
+                                compute::KernelState* dst) {
+  return checked_cast<ScalarUdfAggregator*>(dst)->MergeFrom(ctx, std::move(src));
```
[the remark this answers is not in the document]
icexelloss: I think you are right it shouldn't matter. Will remove.

cpp/src/arrow/engine/substrait/extension_set.cc
icexelloss: Sorry I am not sure what do you mean here - do you want me to change anything?

felipecrv: To keep the line glued to the previous block as it was before. It's a nitpick, but makes a potential function extraction in the future more obvious.

python/pyarrow/src/arrow/python/udf.cc
icexelloss: I am actually no sure about how Py reference counting should work so I think I could use some help here. Currently I wrote this by trial and error and found if I don't have the Py_INCREF here it would segfault. 

@westonpace What is the best way for me to understand how Py_INCREF and OwnedRef are supposed to work here?

icexelloss: I am actually no sure about how Py reference counting should work so I think I could use some help here. Currently I wrote this by trial and error and found if I don't have the Py_INCREF here it would segfault. 

@westonpace What is the best way for me to understand how Py_INCREF and OwnedRef are supposed to work here?

icexelloss: My current probably incorrect understanding is that if we are passing a PyOjbect to a OwnedRef or OwnedRefNoGIL then we need to call Py_INCREF because the de constructor of OwnedRef and OwnedRefNoGIL will Py_XDECREF?

cpp/src/arrow/engine/substrait/extension_set.cc
icexelloss: @felipecrv My understanding is with `const FieldRef*` then the move below has no effect / need to copy the data (because it is pointer to const FieldRef) and with `FieldRef*` the move operator can then actually move the object and not copying it?

python/pyarrow/conftest.py
```diff
@@ -278,3 +278,59 @@ def unary_function(ctx, x):
...
+    func_name = "y=sum_mean(x...)"
+    func_doc = {"summary": "Varargs aggregate",
+                "description": "Varargs aggregate"}
+
+    pc.register_aggregate_function(func,
+                                   func_name,
+                                   func_doc,
+                                   {
+                                       "x": pa.int64(),
+                                       "y": pa.float64()
```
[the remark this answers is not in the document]
icexelloss: (Realized I doubt explained this in https://github.com/apache/arrow/pull/35514/files/f1d3bcbd803a95e378d09121c5061787a68755d2#r1196685513

python/pyarrow/src/arrow/python/udf.cc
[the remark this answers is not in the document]
icexelloss: Actually IIUC. The move operator is needed here per example
```
// Simple move constructor
A(A&& arg) : member(std::move(arg.member)) // the expression "arg.member" is lvalue
{}
 
// Simple move assignment operator
A& operator=(A&& other)
{
    member = std::move(other.member);
    return *this;
}
```

from https://en.cppreference.com/w/cpp/utility/move

(Code doesn't compile if I remove this)

cpp/src/arrow/engine/substrait/extension_set.cc
felipecrv: You're right. This is one of the many ways copies can creep in.

But looking at the definition of `field_ref` I noticed it can only return `const`, so removing `const` won't really be possible. So I recommend removing the `std::move(*arg_ref)` below so it doesn't give the wrong impression that a move is happening.

felipecrv: More details on the move const issue: https://stackoverflow.com/questions/27810535/why-does-calling-stdmove-on-a-const-object-call-the-copy-constructor-when-pass

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -101,6 +125,99 @@ struct PythonTableUdfKernelInit {
...
+        arrow::Table::FromRecordBatches(input_schema, values)
+      );
+      ARROW_ASSIGN_OR_RAISE(
+        table, table->CombineChunks(ctx->memory_pool())
+      );
+
+      UdfContext udf_context{ctx->memory_pool(), table->num_rows()};
+      for (int arg_id = 0; arg_id < num_args; arg_id++) {
+        // Since we combined chunks thComere is only one chunk
+        std::shared_ptr<Array> c_data = table->column(arg_id)->chunk(0);
```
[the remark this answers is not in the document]
icexelloss: Added test to make sure we raise a reasonable error

cpp/src/arrow/engine/substrait/extension_set.cc
icexelloss: I see. Updated

cpp/src/arrow/engine/substrait/extension_set.cc
icexelloss: Thanks! Removed `std::move(*arg_ref)`

python/pyarrow/_compute.pyx
```diff
@@ -2665,11 +2665,19 @@ cdef get_register_tabular_function():
...
+    reg.register_func = RegisterAggregateFunction
+    return reg
+
+
 def register_scalar_function(func, function_name, function_doc, in_types, out_type,
                              func_registry=None):
     """
     Register a user-defined scalar function.
 
+    This API is EXPERIMENTAL.
```
icexelloss, line 2679: @westonpace Added this because I think this is fairly new/evolving API. Does marking at Experimental look ok to you?

python/pyarrow/src/arrow/python/udf.cc
icexelloss: Added!

python/pyarrow/_compute.pyx
```diff
@@ -2738,9 +2744,83 @@ def register_scalar_function(func, function_name, function_doc, in_types, out_ty
...
+        UdfContext.
+        Then, it must take arguments equal to the number of
+        in_types defined. It must return Scalar matching the
+        out_type.
+        To define a varargs function, pass a callable that takes
+        varargs. The in_type needs to match in type of inputs when
+        the function gets called.
+
+    function_name : str
+        Name of the function. This name must be globally unique.
```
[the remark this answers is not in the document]
icexelloss: Clarfied

python/pyarrow/_compute.pyx
```diff
@@ -2738,9 +2744,83 @@ def register_scalar_function(func, function_name, function_doc, in_types, out_ty
...
+    >>> func_doc["summary"] = "simple mean udf"
+    >>> func_doc["description"] = "compute mean"
+    >>>
+    >>> def compute_mean(ctx, array):
+    ...     return pa.scalar(np.nanmean(array))
+    >>>
+    >>> func_name = "py_compute_mean"
+    >>> in_types = {"array": pa.int64()}
+    >>> out_type = pa.float64()
+    >>> pc.register_aggregate_function(compute_mean, func_name, func_doc,
```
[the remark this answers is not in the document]
icexelloss: Updated

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -65,6 +69,26 @@ struct PythonUdfKernelInit {
...
+}
+
+arrow::Status AggregateUdfMerge(compute::KernelContext* ctx, compute::KernelState&& src,
+                                compute::KernelState* dst) {
+  return checked_cast<ScalarUdfAggregator*>(dst)->MergeFrom(ctx, std::move(src));
+}
+
+arrow::Status AggregateUdfFinalize(compute::KernelContext* ctx, arrow::Datum* out) {
+  auto udf = checked_cast<ScalarUdfAggregator*>(ctx->state());
+  return SafeCallIntoPython([&]() -> Status {return udf->Finalize(ctx, out);});
```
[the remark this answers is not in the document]
icexelloss: Refactored and moved to `PythonUdfScalarAggregatorImpl`

python/pyarrow/src/arrow/python/udf.cc
icexelloss: Added

icexelloss:
@felipecrv @westonpace I think all the comments are addressed except for this one: https://github.com/apache/arrow/pull/35514#discussion_r1205081272 (which needs a bit more discussion since I am not sure the best way to handle python ref counting here.

Please take another look thanks!

cpp/src/arrow/engine/substrait/extension_set.cc
icexelloss: Good catch added

icexelloss:
@westonpace I believe this PR is good to go. The failed CI seems unrelated. I have checked the Py refcount and it seems OK (I will add details in the comment thread above)

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -101,6 +125,99 @@ struct PythonTableUdfKernelInit {
...
+
+    Status Consume(compute::KernelContext* ctx, const compute::ExecSpan& batch) {
+      ARROW_ASSIGN_OR_RAISE(auto rb, batch.ToExecBatch().ToRecordBatch(input_schema, ctx->memory_pool()));
+      values.push_back(rb);
+      return Status::OK();
+    }
+
+    Status MergeFrom(compute::KernelContext* ctx, compute::KernelState&& src) {
+      const auto& other_state = checked_cast<const PythonUdfScalarAggregatorImpl&>(src);
+      values.insert(values.end(), other_state.values.begin(), other_state.values.end());
```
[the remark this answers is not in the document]
icexelloss: Updated!

python/pyarrow/src/arrow/python/udf.cc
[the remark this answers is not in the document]
icexelloss: Added test

python/pyarrow/src/arrow/python/udf.cc
icexelloss: I looked into this more and observed the following.

The refcount of Python function is 2 before registering with Arrow, and is 3 after registering. After invoking the function with arrow compute, the refcount doesn't increase. (still 3) So I believe the code is correct.

On the other hand, I did observe scalar UDF might have a refcount leaking (I observed the refcount increases by 1 after calling the function with arrow compute). Which I can look into separately.

In summary, my understanding is that every time we wrap the PyObject* into a OwnedRef, we need to increase the refcount by 1 so the refcount doesn't change after the destructor of OwnedRef gets invoked. I am 90% certain this is correct understanding of how it should work but happy to hear more thoughts @westonpace

westonpace:
@icexelloss I'll take another look through today.

icexelloss:
> @icexelloss I'll take another look through today.

Thank you!

python/pyarrow/_compute.pyx
westonpace: That's fine.  UDFs are also marked experimental in the [pyarrow docs](https://github.com/apache/arrow/blob/407423394cb988669a0b5716899810a924c0aa4c/docs/source/python/compute.rst?plain=1#L390).

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -15,15 +15,19 @@
 // specific language governing permissions and limitations
 // under the License.
 
-#include "arrow/python/udf.h"
+#include <iostream>
```
westonpace, line 18:
```suggestion
```

python/pyarrow/src/arrow/python/udf.cc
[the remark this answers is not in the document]
westonpace: Yep.  The `move` is indeed needed.  C++ is confusing :cold_sweat:

python/pyarrow/conftest.py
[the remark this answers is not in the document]
westonpace: Ok, so the test case is verifying that the python function can take in `*args` if needed (even though it still lists the args when registering)?

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -101,6 +125,103 @@ struct PythonTableUdfKernelInit {
   UdfWrapperCallback cb;
 };
 
+struct PythonUdfScalarAggregatorImpl : public ScalarUdfAggregator {
+  PythonUdfScalarAggregatorImpl(UdfWrapperCallback agg_cb,
+                                std::shared_ptr<OwnedRefNoGIL> agg_function,
+                                std::vector<std::shared_ptr<DataType>> input_types,
+                                std::shared_ptr<DataType> output_type)
+      : agg_cb(agg_cb), agg_function(agg_function), output_type(output_type) {
+    Py_INCREF(agg_function->obj());
```
westonpace, line 134: This increment seems redundant given you already have one [here](https://github.com/apache/arrow/pull/35514/files#diff-d269c07da082f1b2d9b8f8628effd64aff74c163e325d253c2246410e0d1c3d0R381).

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -101,6 +125,103 @@ struct PythonTableUdfKernelInit {
   UdfWrapperCallback cb;
 };
 
+struct PythonUdfScalarAggregatorImpl : public ScalarUdfAggregator {
+  PythonUdfScalarAggregatorImpl(UdfWrapperCallback agg_cb,
+                                std::shared_ptr<OwnedRefNoGIL> agg_function,
+                                std::vector<std::shared_ptr<DataType>> input_types,
+                                std::shared_ptr<DataType> output_type)
+      : agg_cb(agg_cb), agg_function(agg_function), output_type(output_type) {
```
westonpace, line 133:
```suggestion
      : agg_cb(std::move(agg_cb)), agg_function(agg_function), output_type(std::move(output_type)) {
```
Minor nit

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -101,6 +125,103 @@ struct PythonTableUdfKernelInit {
...
+    Py_INCREF(agg_function->obj());
+    std::vector<std::shared_ptr<Field>> fields;
+    for (size_t i = 0; i < input_types.size(); i++) {
+      fields.push_back(std::move(field("", input_types[i])));
```
westonpace, line 137:
```suggestion
      fields.push_back(field("", input_types[i]));
```
No need to call `std::move` on something that doesn't have a name.

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -101,6 +125,103 @@ struct PythonTableUdfKernelInit {
...
+    }
+    input_schema = schema(std::move(fields));
+  };
+
+  ~PythonUdfScalarAggregatorImpl() {
```
westonpace, line 142:
```suggestion
  ~PythonUdfScalarAggregatorImpl() override {
```

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -101,6 +125,103 @@ struct PythonTableUdfKernelInit {
...
+    if (_Py_IsFinalizing()) {
+      agg_function->detach();
+    }
+  }
+
+  Status Consume(compute::KernelContext* ctx, const compute::ExecSpan& batch) {
```
westonpace, line 148:
```suggestion
  Status Consume(compute::KernelContext* ctx, const compute::ExecSpan& batch) override {
```

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -101,6 +125,103 @@ struct PythonTableUdfKernelInit {
...
+    ARROW_ASSIGN_OR_RAISE(
+        auto rb, batch.ToExecBatch().ToRecordBatch(input_schema, ctx->memory_pool()));
+    values.push_back(std::move(rb));
+    return Status::OK();
+  }
+
+  Status MergeFrom(compute::KernelContext* ctx, compute::KernelState&& src) {
```
westonpace, line 155:
```suggestion
  Status MergeFrom(compute::KernelContext* ctx, compute::KernelState&& src) override {
```

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -101,6 +125,103 @@ struct PythonTableUdfKernelInit {
...
+    auto& other_values = checked_cast<PythonUdfScalarAggregatorImpl&>(src).values;
+    values.insert(values.end(), std::make_move_iterator(other_values.begin()),
+                  std::make_move_iterator(other_values.end()));
+
+    other_values.erase(other_values.begin(), other_values.end());
+    return Status::OK();
+  }
+
+  Status Finalize(compute::KernelContext* ctx, Datum* out) {
```
westonpace, line 164:
```suggestion
  Status Finalize(compute::KernelContext* ctx, Datum* out) override {
```

These `override` suggestions are pretty minor but can be nice.  It just makes it a little easier to track down what went wrong if the parent class' signature changes.

westonpace (approved):
A few more very minor suggestions but, overall, I think this is fine.

python/pyarrow/conftest.py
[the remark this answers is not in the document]
icexelloss: That is correct

icexelloss:
I checked failed CI jobs and those seem unrelated.

python/pyarrow/src/arrow/python/udf.cc
icexelloss: Admitted there could be some redundancy here. I created an follow up to take a closer look:

https://github.com/apache/arrow/issues/36000

PR #35518, opened 2023-05-09
MINOR: [R] Add archery command for revdep checks to packaging checklist

paleolimbot:
This PR adds the actual archery command that runs the reverse dependency checks instead of linking to a half-baked gist!


thisisnic (approved):
Muchas gracias!

PR #35522, opened 2023-05-09
GH-35521: [C++] Hash null bitmap only if null count is 0

### LINKED ISSUE #35521: [C++] Hashing array scalar with null bitmap and non-null 0s bitmap produces different hashes.

Null bitmaps in arrays are generally set to `nullptr` when the null count for the array is 0. However, there are cases where the null bitmap is set to a 0s buffer instead. Semantically, a 0s bitmap and a `nullptr` bitmap are the same thing. However, the hashing algorithm for array scalars will hash the null bitmap if it exists, and the resulting hash is not 0. This behavior can lead to two scalars with the same semantic value, but different internal values of the null bitmap, to hash to different values.

### Component(s)

C++

micah-white:
### Are these changes tested?

I am new to contributing and am having a hard time creating a good test for this. The steps to reproduce this bug originally are too complicated for a simple test. I included my attempt at making a good test in the PR, but some help would be nice.
-->

No.

* Closes: #35521

### DIFF
--- a/cpp/src/arrow/scalar.cc
+++ b/cpp/src/arrow/scalar.cc
@@ -153,10 +153,10 @@ struct ScalarHashImpl {
 
   Status ArrayHash(const ArrayData& a) {
     RETURN_NOT_OK(StdHash(a.length) & StdHash(a.GetNullCount()));
-    if (a.GetNullCount() != 0) {
+    if (a.GetNullCount() != 0 && a.buffers[0] != nullptr) {
       // We can't visit values without unboxing the whole array, so only hash
       // the null bitmap for now. Only hash the null bitmap if the null count
-      // is 0 to ensure hash consistency.
+      // is not 0 to ensure hash consistency.
       RETURN_NOT_OK(BufferHash(*a.buffers[0]));
     }
     for (const auto& child : a.child_data) {
--- a/cpp/src/arrow/scalar.cc
+++ b/cpp/src/arrow/scalar.cc
@@ -153,9 +153,10 @@ struct ScalarHashImpl {
 
   Status ArrayHash(const ArrayData& a) {
     RETURN_NOT_OK(StdHash(a.length) & StdHash(a.GetNullCount()));
-    if (a.buffers[0] != nullptr) {
+    if (a.GetNullCount() != 0) {
       // We can't visit values without unboxing the whole array, so only hash
-      // the null bitmap for now.
+      // the null bitmap for now. Only hash the null bitmap if the null count
+      // is 0 to ensure hash consistency.
       RETURN_NOT_OK(BufferHash(*a.buffers[0]));
     }
     for (const auto& child : a.child_data) {
--- a/cpp/src/arrow/scalar_test.cc
+++ b/cpp/src/arrow/scalar_test.cc
@@ -1064,7 +1064,6 @@ class TestListScalar : public ::testing::Test {
   using ScalarType = typename TypeTraits<T>::ScalarType;
 
   void SetUp() {
-    //     type_ = std::make_shared<T>(int16());
     type_ = MakeListType<T>(int16(), 3);
     value_ = ArrayFromJSON(int16(), "[1, 2, null]");
   }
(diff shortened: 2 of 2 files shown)


cpp/src/arrow/scalar.cc
```diff
@@ -153,9 +153,10 @@ struct ScalarHashImpl {
 
   Status ArrayHash(const ArrayData& a) {
     RETURN_NOT_OK(StdHash(a.length) & StdHash(a.GetNullCount()));
-    if (a.buffers[0] != nullptr) {
+    if (a.GetNullCount() != 0) {
```
micah-white, line 156: IIUC, the null count should get cached from calling `GetNullCount()` above, so this line does not have performance degradations.

cpp/src/arrow/scalar_test.cc
```diff
@@ -1106,6 +1106,25 @@ class TestListScalar : public ::testing::Test {
...
+    auto data = empty_bitmap_scalar.value->data()->buffers[1];
+    std::vector<uint8_t> bitmap_data = {0,0,0};
+    auto null_bitmap = std::make_shared<Buffer>(bitmap_data.data(), 3);
+
+    std::shared_ptr<Int16Array> arr(new Int16Array(3, data, null_bitmap, 0));
+    ASSERT_TRUE(arr->null_count() == 0);
+    // this line fails - I don't know how to create an array with a null bitmap
+    // that is all 0s.
+    ASSERT_TRUE(arr->data()->buffers[0] != nullptr);
+    ScalarType set_bitmap_scalar(arr);
```
micah-white, line 1123: As I said in the description, I don't know how to make an array with a 0s null bitmap. I ran into this bug when using the `filter()` function on a `LargeListArray` of strings. The scalars in the array would have the null bitmap set even if there were no nulls, changing the hash values. I assume using that exact method is too circumspect to be used here.

mapleFU (approved):
This LGTM @wjones127 @pitrou @westonpace Mind take a look?

mapleFU:
Would you mind take a look at the failed CI tests:

```
[ RUN      ] TestListScalar/0.TestHashing
C:/projects/arrow/cpp/src/arrow/scalar_test.cc(1122): error: Value of: arr->data()->buffers[0] != nullptr
  Actual: false
Expected: true
[  FAILED  ] TestListScalar/0.TestHashing, where TypeParam = class arrow::ListType (0 ms)
[----------] 3 tests from TestListScalar/0 (0 ms total)
[----------] 3 tests from TestListScalar/1, where TypeParam = class arrow::LargeListType
[ RUN      ] TestListScalar/1.Basics
[       OK ] TestListScalar/1.Basics (0 ms)
[ RUN      ] TestListScalar/1.ValidateErrors
[       OK ] TestListScalar/1.ValidateErrors (0 ms)
[ RUN      ] TestListScalar/1.TestHashing
C:/projects/arrow/cpp/src/arrow/scalar_test.cc(1122): error: Value of: arr->data()->buffers[0] != nullptr
  Actual: false
Expected: true
[  FAILED  ] TestListScalar/1.TestHashing, where TypeParam = class arrow::LargeListType (0 ms)
[----------] 3 tests from TestListScalar/1 (0 ms total)
[----------] 3 tests from TestListScalar/2, where TypeParam = class arrow::FixedSizeListType
[ RUN      ] TestListScalar/2.Basics
[       OK ] TestListScalar/2.Basics (0 ms)
[ RUN      ] TestListScalar/2.ValidateErrors
[       OK ] TestListScalar/2.ValidateErrors (0 ms)
[ RUN      ] TestListScalar/2.TestHashing
C:/projects/arrow/cpp/src/arrow/scalar_test.cc(1122): error: Value of: arr->data()->buffers[0] != nullptr
  Actual: false
Expected: true
[  FAILED  ] TestListScalar/2.TestHashing, where TypeParam = class arrow::FixedSizeListType (0 ms)
```

?

jorisvandenbossche:
@pitrou small example using python (you don't actually need to filter or slice yourself, since a scalar is already a slice into its parent array):

```
In [31]: arr1 = pa.array([[0, 1], [2, 3]])

In [32]: scalar1 = arr1[0]

In [33]: arr2 = pa.array([[0, 1], [2, None]])

In [34]: scalar2 = arr2[0]

In [35]: scalar1
Out[35]: <pyarrow.ListScalar: [0, 1]>

In [36]: scalar2
Out[36]: <pyarrow.ListScalar: [0, 1]>

In [37]: hash(scalar1)
Out[37]: 6972737373264176731

In [38]: hash(scalar2)
Out[38]: 5286180417804377197

In [39]: scalar1.values.buffers()
Out[39]: 
[None,
 <pyarrow.Buffer address=0x7fbc56a08200 size=32 is_cpu=True is_mutable=True>]

In [40]: scalar2.values.buffers()
Out[40]: 
[<pyarrow.Buffer address=0x7fbc56a08180 size=1 is_cpu=True is_mutable=True>,
 <pyarrow.Buffer address=0x7fbc56a08280 size=32 is_cpu=True is_mutable=True>]
```

Those two scalars are equal, so should have the same hash. But the one has a validity bitmap, and the other not.

cpp/src/arrow/scalar_test.cc
jorisvandenbossche: Yeah, it seems we explicitly don't keep the bitmap around if there are no nulls in the Array construction:

https://github.com/apache/arrow/blob/ec29c6ffc3cb1af4db4903d9877b2f0b548a3ad9/cpp/src/arrow/array/data.cc#L53-L55

One way around this would be to create a ListArray, and then get one element of it as a scalar (like my code snippet in python)

jorisvandenbossche (commented):
Thanks for looking into this!

cpp/src/arrow/scalar_test.cc
pitrou: Yes, just rewrite the Python snippet (or some equivalent) in C++.  It should actually be easy, using `ArrayFromJSON` and `Array::GetScalar`.

cpp/src/arrow/scalar.cc
pitrou: Indeed, however, a NullArray can have a non-zero null count but without a null bitmap. So you must make sure to check the buffers as well.

cpp/src/arrow/scalar.cc
```diff
@@ -153,9 +153,10 @@ struct ScalarHashImpl {
 
   Status ArrayHash(const ArrayData& a) {
     RETURN_NOT_OK(StdHash(a.length) & StdHash(a.GetNullCount()));
-    if (a.buffers[0] != nullptr) {
+    if (a.GetNullCount() != 0 && a.buffers[0] != nullptr) {
```
westonpace, line 156: IIRC, it's also possible that `a.buffers[0] == nullptr` if all the elements are valid.  Is it possible that we still get differing hashes in this case?

* All elements valid and equal and validity bitmap present (would hash the validity bitmap)
* All elements valid and equal and validity bitmap missing (would not hash the validity bitmap)

micah-white: Yes, that's the bug that I was trying to fix. Since both cases have the same semantic value, their hashes should be the same.

jorisvandenbossche (approved):
Looks good to me

It might be nice to test the case of a NullScalar as well (for which you had to change some code), but the test coverage for hashing is quite low to start with ..

pitrou (approved):
I did a few cosmestic changes. Thank you @micah-white !

PR #35523, opened 2023-05-10
GH-35392: [R][CI] Bump the R versions we test to include 4.3

### LINKED ISSUE #35392: [R][CI] Bump the R versions we test to include 4.3

See #12980 for where we did this a year ago.

### Component(s)

Continuous Integration, R

thisisnic:
* Closes: #35392

### DIFF
--- a/.github/workflows/r.yml
+++ b/.github/workflows/r.yml
@@ -242,8 +242,8 @@ jobs:
       fail-fast: false
       matrix:
         config:
-        - { rtools: 42, rversion: "4.2" }
-        - { rtools: 42, rversion: "devel" }
+        - { rtools: 43, rversion: "4.3" }
+        - { rtools: 43, rversion: "devel" }
     env:
       ARROW_R_CXXFLAGS: "-Werror"
       _R_CHECK_TESTS_NLINES_: 0
@@ -255,7 +255,7 @@ jobs:
           fetch-depth: 0
       - run: mkdir r/windows
       - name: Download artifacts
-        if: ${{ matrix.config.rtools == 42 }}
+        if: ${{ matrix.config.rtools == 43 }}
         uses: actions/download-artifact@v3
         with:
           name: libarrow-rtools40-ucrt64.zip
--- a/.github/workflows/r.yml
+++ b/.github/workflows/r.yml
@@ -60,7 +60,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        r: ["4.2"]
+        r: ["4.3"]
         ubuntu: [20.04]
         force-tests: ["true"]
     env:
@@ -318,7 +318,7 @@ jobs:
             timeout = 3600
           )
       - name: Run lintr
-        if: ${{ matrix.config.rversion == '4.2' }}
+        if: ${{ matrix.config.rversion == '4.3' }}
         env:
           NOT_CRAN: "true"
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.env
+++ b/.env
@@ -71,7 +71,7 @@ NUMBA=latest
 NUMPY=latest
 PANDAS=latest
 PYTHON=3.8
-R=4.2
+R=4.3
 SPARK=master
 TURBODBC=latest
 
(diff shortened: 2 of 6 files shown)


thisisnic:
Not sure that everything is expected to work here yet as https://cloud.r-project.org/bin/linux/ubuntu/ hasn't been updated to mention 4.3 (could be a docs issue, but perhaps not).

thisisnic:

.github/workflows/r.yml
```diff
@@ -234,8 +234,8 @@ jobs:
       fail-fast: false
       matrix:
         config:
-        - { rtools: 42, rversion: "4.2" }
```
nealrichardson, line 237: Do we still need to test on rtools 42? I don't know the difference between 42 and 43 to know if not having any jobs that use it is a risk.

assignUser: >From the user perspective, Rtools43 is the same as Rtools42. It uses newer versions of the compiler toolchain and libraries, and hence some package authors will have to extend their make files to link additional libraries. Maintainers of CRAN and Bioconductor packages may use [these patches](https://www.r-project.org/nosvn/winutf8/ucrt3/patches/) for reference or re-use them in their code.

Does not look like there are any changes that affect us, so we can probably just update to 43?

nealrichardson: Different compiler version is potentially concerning, but the question is if there is enough risk to justify another hour-long build. Probably not?

What about nightly packages?

thisisnic:

.github/workflows/r.yml
thisisnic: > What about nightly packages?

How do you mean?

assignUser: I just  tried to update the version used in nightly packages and that's a no go for now (at least for the 10.13 runner...) as rig doesn't find 4.3 due to url change and even compiling from source I was unable to make any newer rig versions run on 10.13. We could do symlinking by hand or something but I'd rather not... hopefully we will be able to drop 10.13 soon :crossed_fingers: 

Also another thing regarding rtools: do we want to continure building urct + 32/64 bit versions of libarrow? imo it would be sensible to drop at the minimum the 32 bit build... but we don't really have a support policy so :shrug: (something I plan to bring up on the ml again)

paleolimbot: I don't think we should drop the 32-bit build unless we run into a situation where the build fails and we don't have the developer time to fix it. I don't think either of those things are true right now? We definitely still want to support the same as the tidyverse support policy within the limits of our developer capacity...old R is very common in government and on supercomputer/cluster environments.

paleolimbot: RSPM provides binary builds for older Windows (I just checked, and `install.packages("arrow")` on Windows/R 4.0.6 gives the latest version without an attempted source build.

assignUser: Well then people who want to use arrow on old versions on windows can use RSPM? I am strongly +1 on dropping 32bit support. Win 7 has been eol for 3 years and I don't think it is our responsibility to support such outdated configurations. I would also assume that the percentage of users that use this is rather small? So not worth it from that direction either (compare to java 8 discussion on the ml)

paleolimbot: I don't think it will cause a problem, but worth asking on the mailing list. Alternatively, you could just leave it in as long as it's not costing you development time?

nealrichardson: RSPM (technically, PPPM today) will need the libarrow binaries to build for those older R versions, so it's not a solution to the problem, it's actually a reason to keep building the 32-bit packages.

assignUser: So (as we are following the past-4 rule I guess?) we will drop the 32bit build when 4.6 is released (aka in 3! years)?  My only reason to drop it is shortening of build times, as we build 3 libarrow versions in each windows build so :shrug:

nealrichardson: Valid concern, though in regular CI we only build the ucrt64 version, the 32-bit is only nightly, so the build time is less of an issue there. (I think the libarrow builds are also parallelized.)

That said, we already "violate" the tidyverse standard for version compatibility (Windows only) because we require R 4.0 on Windows due to C++17. 

I think I'm landing in the same place as @paleolimbot here: leave it in unless/until the 32-bit builds are causing undue developer burden, which it seems like today they are not.

thisisnic:
@github-actions autotune

jonkeane:
This is done already, yeah?

PR #35525, opened 2023-05-10
GH-35528:[Java] Fix RangeEqualsVisitor comparing BitVector with different begin index

### LINKED ISSUE #35528: [Java]bugfix: RangeEqualsVisitor compare BitVector false when compared vectors have different begin index.

In origin code, when compared vectors have different begin index, it will compare vectors value with same index, which is error logic. this pr is to fix it.

### Component(s)

Java

wenxlan:
### Rationale for this change

bugfix: RangeEqualsVisitor compare BitVector false when compared vectors have different begin index. 

In origin code, when compared vectors have different begin index, it will compare vectors value with same index, which is error logic. this pr is to fix it.

only changes RangeEqualsVisitor.compareBaseFixedWidthVectors method, and add test

### Are these changes tested?
yes.

no.

* Closes: #35528


java/vector/src/test/java/org/apache/arrow/vector/compare/TestRangeEqualsVisitor.java
```diff
@@ -26,13 +26,7 @@
...
 import org.apache.arrow.memory.BufferAllocator;
 import org.apache.arrow.memory.RootAllocator;
-import org.apache.arrow.vector.BigIntVector;
-import org.apache.arrow.vector.Float4Vector;
-import org.apache.arrow.vector.Float8Vector;
-import org.apache.arrow.vector.IntVector;
-import org.apache.arrow.vector.LargeVarCharVector;
-import org.apache.arrow.vector.VarCharVector;
-import org.apache.arrow.vector.ZeroVector;
+import org.apache.arrow.vector.*;
```
lidavidm, line 29: nit: I realize this was probably the IDE but can this be reverted? For review it's easier to list out imports

lidavidm (commented):
LGTM overall. Thanks! 

(Just one comment.)

java/vector/src/test/java/org/apache/arrow/vector/compare/TestRangeEqualsVisitor.java
wenxlan: it is reverted now～

lidavidm (approved):
LGTM, will merge after CI

lidavidm (approved):
Thanks!

PR #35527, opened 2023-05-10
GH-35526: [CI][C++] Fixing arrow::internal::IsNullRunEndEncoded redeclared

mapleFU:
### Rationale for this change

Fixing arrow::internal::IsNullRunEndEncoded redeclared. The CI will report:

```
'bool arrow::internal::IsNullRunEndEncoded(const arrow::ArrayData&, int64_t)' redeclared without dllimport attribute after being referenced with dll linkage
```

And:

```
warning: 'selection_vector_type' may be used uninitialized
```

### What changes are included in this PR?

1. Add `ARROW_FRIEND_EXPORT` for friend function
2. Initialize pointer in gandiva with `nullptr`

### Are these changes tested?

No

### Are there any user-facing changes?

No

* Closes: #35526

PR #35533, opened 2023-05-10
MINOR: [C++] Add missing ARROW_EXPORT to IsNull impls

felipecrv:
### Rationale for this change

Fix MinGW CI builds. (Example https://github.com/apache/arrow/actions/runs/4918687806/jobs/8806728909)

ARROW_EXPORT internal functions that are called from inlined functions like `IsNull` and `IsValid`.

### Are these changes tested?

N/A.

More symbols are exported in the .dll.

### DIFF
--- a/cpp/src/arrow/array/data.cc
+++ b/cpp/src/arrow/array/data.cc
@@ -64,14 +64,14 @@ static inline void AdjustNonNullable(Type::type type_id, int64_t length,
 
 namespace internal {
 
-bool IsNullSparseUnion(const ArrayData& data, int64_t i) {
+ARROW_EXPORT bool IsNullSparseUnion(const ArrayData& data, int64_t i) {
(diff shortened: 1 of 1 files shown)


felipecrv:
@pitrou

pitrou:
Looks like Github Actions is currently broken...

mapleFU:
I've create an issue and submit a request for this: https://github.com/apache/arrow/pull/35527

pitrou (commented):
`ARROW_EXPORT` should only be useful in declarations, so I'm not sure why you're adding it in `data.cc`.

felipecrv:
> `ARROW_EXPORT` should only be useful in declarations, so I'm not sure why you're adding it in `data.cc`.

I already had them in the header and got the wrong impression from the windows compiler errors. @mapleFU's fix really is the way to go -- with the `ARROW_FRIEND_EXPORT` -- that's the real missing export declaration.

PR #35541, opened 2023-05-11
GH-35498: [C++] Fix source node batch realignment

### LINKED ISSUE #35498: [C++][Parquet] Parquet write_to_dataset performance regression

On Linux, `pyarrow.parquet.write_to_dataset` shows a large performance regression in Arrow 12.0 versus 11.0.

The following results were collected using Ubuntu 22.04.2 LTS (5.15.0-71-generic), Intel Haswell 4-core @ 3.6GHz, 16 GB RAM, Samsung 840 Pro SSD. They are elapsed times in seconds to write a single int64 column of integers [0,..., _length_-1] with no compression and no multi-threading:

| Array length | Arrow 11 (s) | Arrow 12 (s) |
|-----------------:|--------:|--------:|
| 1,000,000 | 0.1 | 0.1|
| 2,000,000 | 0.2 | 0.4 |
| 4,000,000 | 0.3 | 1.6 |
| 8,000,000 | 0.8 | 6.2 |
| 16,000,000 | 2.3 | 24.4 |
| 32,000,000 | 6.5 | 94.1 |
| 64,000,000 | 13.5 | 371.7 |

The output directory was deleted before each run.
```
"""check.py"""
import sys
import time
import numpy as np
import pyarrow as pa
import pyarrow.parquet as pq

def main():
    path = '/tmp/test.parquet'
    length = 10_000_000 if len(sys.argv) < 2 else int(sys.argv[1])
    table = pa.Table.from_arrays([pa.array(np.arange(length))], names=['A'])
    t0 = time.perf_counter()
    pq.write_to_dataset(
        table, path, schema=table.schema, use_legacy_dataset=False, use_threads=False, compression=None
    )
    duration = time.perf_counter() - t0
    print(f'{duration:.2f}s')

if __name__ == '__main__':
    main()
```

Running `git bisect` on local builds leads me to this commit: 660d259f525d301f7ff5b90416622698fa8a5e9c: [C++] Add ordered/segmented aggregation Substrait extension (#34627).

Following that change, Flamegraphs show a lot of additional time spent in `arrow::util::EnsureAlignment` calling glibc `memcpy`:

Before ~1.3s (ddd0a337174e57cdc80b1ee30dc7e787acfc09f6)
good-ddd0a33 perf

After ~9.6s (660d259f525d301f7ff5b90416622698fa8a5e9c)
bad-660d259 perf

Reading and `pyarrow.parquet.write_table` appear unaffected.

### Component(s)

C++, Parquet

rtpsw:
### Rationale for this change

Currently, source node uses too high a value for realignment, which [caused a performance degradation](https://github.com/apache/arrow/issues/35498).

The source node batch realigns an array to its byte width (if it is more than 1).

### Are these changes tested?

The changes are tested for correctness by the existing tests. The performance is checked by regression jobs.

Only performance.

* Closes: #35498

### DIFF
--- a/cpp/src/arrow/acero/source_node.cc
+++ b/cpp/src/arrow/acero/source_node.cc
@@ -105,10 +105,11 @@ struct SourceNode : ExecNode, public TracedNode {
             }
             ExecBatch batch = morsel.Slice(offset, batch_size);
             for (auto& value : batch.values) {
-              if (value.is_array()) {
-                ARROW_ASSIGN_OR_RAISE(value, arrow::util::EnsureAlignment(
-                                                 value.make_array(), ipc::kArrowAlignment,
-                                                 default_memory_pool()));
+              if (value.is_array() && value.type()->byte_width() > 1) {
+                ARROW_ASSIGN_OR_RAISE(
+                    value, arrow::util::EnsureAlignment(value.make_array(),
+                                                        value.type()->byte_width(),
+                                                        default_memory_pool()));
               }
             }
             if (has_ordering) {
--- a/cpp/src/arrow/acero/source_node.cc
+++ b/cpp/src/arrow/acero/source_node.cc
@@ -34,6 +34,7 @@
 #include "arrow/util/align_util.h"
 #include "arrow/util/async_generator.h"
 #include "arrow/util/async_util.h"
+#include "arrow/util/bit_util.h"
 #include "arrow/util/checked_cast.h"
 #include "arrow/util/future.h"
 #include "arrow/util/logging.h"
@@ -105,10 +106,11 @@ struct SourceNode : ExecNode, public TracedNode {
             }
             ExecBatch batch = morsel.Slice(offset, batch_size);
             for (auto& value : batch.values) {
-              if (value.is_array() && value.type()->byte_width() > 1) {  // GH-35498
+              int64_t width = value.type()->byte_width();
+              if (value.is_array() && is_fixed_width(value.type()->id()) && width > 1 &&
+                  bit_util::IsPowerOf2(width)) {
                 ARROW_ASSIGN_OR_RAISE(
-                    value, arrow::util::EnsureAlignment(value.make_array(),
-                                                        value.type()->byte_width(),
+                    value, arrow::util::EnsureAlignment(value.make_array(), width,
                                                         default_memory_pool()));
               }
             }
--- a/cpp/src/arrow/acero/source_node.cc
+++ b/cpp/src/arrow/acero/source_node.cc
@@ -105,7 +105,7 @@ struct SourceNode : ExecNode, public TracedNode {
             }
             ExecBatch batch = morsel.Slice(offset, batch_size);
             for (auto& value : batch.values) {
-              if (value.is_array() && value.type()->byte_width() > 1) {
+              if (value.is_array() && value.type()->byte_width() > 1) {  // GH-35498
                 ARROW_ASSIGN_OR_RAISE(
                     value, arrow::util::EnsureAlignment(value.make_array(),
                                                         value.type()->byte_width(),


jorisvandenbossche:
@ursabot please benchmark

rtpsw:
I see a bunch of CI job ([here](https://github.com/apache/arrow/actions/runs/4945533592/jobs/8842394608?pr=35541), [here](https://github.com/apache/arrow/actions/runs/4945533592/jobs/8842394874?pr=35541), [here](https://github.com/apache/arrow/actions/runs/4945533592/jobs/8842395019?pr=35541) and [here](https://ci.appveyor.com/project/ApacheSoftwareFoundation/arrow/builds/47016479)) are failing due to OOM, presumably on realignment, but the allocation sizes are not large. More concerning is [this failure](https://github.com/apache/arrow/actions/runs/4945533592/jobs/8842394197?pr=35541), which shows an invalid alignment of 3. Presumably, the data type of the value being realigned is some kind of struct. @westonpace, what are your thoughts? Perhaps there is an API (or idiom) to get the byte width of the largest field of the data type, which I think would be the correct alignment in such a case?

westonpace:
Structs do not have any buffers themselves beyond the validity buffer.  The validity buffer will have a bit width of 1.  I don't think we need to worry about aligning any buffers with a width of 8 or less (e.g. only worry about it once the width is 16)

One problem with this PR is that it doesn't address offsets buffers.  For example, in a string / list / binary array there is a 32-bit offsets buffer which should be aligned to 32.  These arrays won't show up as fixed width arrays.

I was working on this a bit yesterday as well.  I ended up with something like this...

```
bool CheckMallocAlignment(const ArrayData& array) {
  auto type_id = array.type->storage_id();
  if (type_id == Type::DICTIONARY) {
    // The values array will be checked separately
    type_id = ::arrow::internal::checked_pointer_cast<DictionaryType>(array.type)
                  ->index_type()
                  ->id();
  }
  switch (array.type->id()) {
    case Type::NA:
    case Type::FIXED_SIZE_LIST:
    case Type::FIXED_SIZE_BINARY:
    case Type::BOOL:
    case Type::INT8:
    case Type::UINT8:
    case Type::DECIMAL128:
    case Type::DECIMAL256:
    case Type::SPARSE_UNION:
    case Type::RUN_END_ENCODED:
    case Type::STRUCT:
      // These have no buffers or all buffers need only byte alignment
      return true;
    case Type::INT16:
    case Type::UINT16:
    case Type::HALF_FLOAT:
      return CheckValuesAlignment(array, 16);
    case Type::INT32:
    case Type::UINT32:
    case Type::FLOAT:
    case Type::STRING:
    case Type::BINARY:
    case Type::DATE32:
    case Type::TIME32:
    case Type::LIST:
    case Type::MAP:
    case Type::DENSE_UNION:
    case Type::INTERVAL_MONTHS:
      return CheckValuesAlignment(array, 32);
    case Type::INT64:
    case Type::UINT64:
    case Type::DOUBLE:
    case Type::LARGE_BINARY:
    case Type::LARGE_LIST:
    case Type::LARGE_STRING:
    case Type::DATE64:
    case Type::TIME64:
    case Type::TIMESTAMP:
    case Type::DURATION:
    case Type::INTERVAL_DAY_TIME:
      return CheckValuesAlignment(array, 64);
    case Type::INTERVAL_MONTH_DAY_NANO:
      return CheckValuesAlignment(array, 128);
    default:
      return true;
  }
}
```

It's not pretty.  I also tried using a visitor pattern but it was ending up to be more complicated (though arguably a bit more future proof were we to add more fixed-width types).  `CheckValuesAlignment` checks the buffer at index 1.  Fortunately, it seems that in all cases, the buffer that needs alignment seems to be at that position.

rtpsw:
I was thinking along the lines of your work; you're clearly ahead of me on this.

>  I don't think we need to worry about aligning any buffers with a width of 8 or less

We do. In failures I observed internally, the cause was misalignment of buffer addresses (which were even numerically odd) that were expected to be 4- or 8-byte aligned.

I think the switch-statement should be refactored into `type_traits.h`, if it ends up as part of the solution.

I think the correct idea here is that alignment applies to buffers, not arrays. That is, the format of the buffer is what determines the required alignment, if any, and I guess this format is determined by the buffer index within the array and the array's storage type.

@westonpace, should you or I take this forward?

westonpace:
> We do. In failures I observed internally, the cause was misalignment of buffer addresses (which were even numerically odd) that were expected to be 4- or 8-byte aligned.

Sorry, I meant 8bits or less (e.g. we can safely assume that everything has at least 1 byte of alignment)

> I think the correct idea here is that alignment applies to buffers, not arrays. That is, the format of the buffer is what determines the required alignment, if any, and I guess this format is determined by the buffer index within the array and the array's storage type.

Yes, unfortunately.

> @westonpace, should you or I take this forward?

I think I should have time to make a PR today.  Sorry to steal a task, I should've marked yesterday that I was working on this.

rtpsw:
> Sorry, I meant 8bits or less (e.g. we can safely assume that everything has at least 1 byte of alignment)

Agreed. I included this condition in my recent commit here.

> > I think the correct idea here is that alignment applies to buffers, not arrays.
> 
> Yes, unfortunately.

I'd suggest considering adding something like `DataType::GetBufferAlignment(int index)` to return the required alignment of  buffer at `index` that must be a power-of-2 or otherwise be 0 for when no alignment is required. This would force any future data type to provide this info.

> Sorry to steal a task

OK, you owe me one :)

rtpsw:
Also, consider using `{Check/Ensure}Alignment(Buffer, int64_T)` from `util/align_util.cc`.

pitrou:
Is this obsoleted by #35565?

rtpsw:
> Is this obsoleted by #35565?

Yes. Closing this.

PR #35543, opened 2023-05-11
GH-35542 : [R] Implement schema extraction function

### LINKED ISSUE #35542: [R] Implement schema extraction function

See #35071 for rationale

### Component(s)

R

thisisnic:
``` r
library(arrow)
mtcarrow <- arrow_table(mtcars)
schema(mtcarrow)
#> Schema
#> mpg: double
#> cyl: double
#> disp: double
#> hp: double
#> drat: double
#> wt: double
#> qsec: double
#> vs: double
#> am: double
#> gear: double
#> carb: double
#> 
#> See $metadata for additional Schema metadata
```

<sup>Created on 2023-05-11 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

* Closes: #35542


eitsupi:
I am wondering if the function name `schema` is too generic and how about to be changed to something like `arrow_schmea`?

r/R/schema.R
```diff
@@ -240,10 +226,45 @@ print_schema_fields <- function(s) {
...
+schema.default <- Schema$create
+
+#' @export
+schema.ArrowTabular <- function(x) x$schema
+
+#' @export
+schema.RecordBatchReader <- function(x) x$schema
+
+#' @export
+schema.Dataset <- function(x) x$schema
```
nealrichardson, line 267: Can you also add this for arrow_dplyr_query? Might involve calling implicit_schema, or `x[["schema"]] %||% implicit_schema(x)` or something.

nealrichardson:
> > I am wondering if the function name schema is too generic and how about to be changed to something like arrow_schmea?
> 
> I think that's a good point...particularly since it's an S3 method. `infer_schema()` or `extract_schema()` might also be good choices...we already have `as_schema()` for converting a schema-like object to a Schema. (FWIW I use `infer_nanoarrow_schema()` to do this kind of thing in nanoarrow).

`schema()` exists as a function now (and has since the beginning of the package), this PR seems to be adding additional cases for using it to extract the `$schema` attribute from Arrow objects, and it's doing it via S3 methods rather than `if`s.  If we're worried about the name "schema" colliding with other packages, we would have already seen it already--has anyone?

For what it's worth, I'm personally not a fan of renaming to `verb_schema()`: it's pulling an attribute of the object, so it feels natural that the accessor have the same name as the attribute, as we have e.g. `length.ArrowDatum <- function(x) x$length()`. As a (👴) developer, I don't want to have to remember what `verb_` goes in front, I'd rather just `$schema`.

nealrichardson:
> Yes, but then `schema()` both constructs a schema AND pulls the schema attribute...one function doing more than one thing is also confusing although there's certainly precedent for it in pre-tidyverse R.
> 

Yeah I totally get your concern and this is why I didn't do this before, despite many times typing `schema(ds)` and then remembering that instead I had to type `ds$schema` to get the schema. I'm just not convinced (anymore) that it's less confusing than the status quo, where `getter()` exists for some attributes but not others.

> The `as_schema()` method is perhaps more appropriate (and also perhaps easier to remember than "infer" or "extract")?

I'm not sure that feels natural in these cases: `as_schema(ds)` feels like I'm converting the Dataset to something else, where what I really want is to access the Schema that the Dataset contains.

r/R/schema.R
thisisnic: Will do.  Just to check, when do we end up with `arrow_dplyr_query` objects where `x[["schema"]]` isn't `NULL`?

thisisnic:
While I agree that it's not the best that `schema()` would do both construction and extraction here, the aim here was to come up with something intuitive for users that avoids having to use the `$` for extraction, and I don't think that any of the alternatives quite fit that; I agree that the `as_*` functions sound more conversion than extraction.  I just spent ~30 minutes browsing through docs for other packages (mainly tidyverse/tidymodels/r-lib etc) to see how they handle this kind of thing, but there aren't any good analogous examples.

I also considered whether there are other attributes we might want to implement extraction functions like this for, but I think schemas are the only case and so here's no general problem to solve.  Given that multiple people (Neal, myself, and also Francois on the original ticket) report having tried to use `schema()` to extract a schema despite knowing that `schema()` is (also) the function for creating schemas, it might be the best intuitive option we have right now?

thisisnic:
Though it looks like it's not possible to get this working sensibly as an S3 generic anyway, may need to redesign.

r/R/schema.R
nealrichardson: I seem to recall that when we collapse a query and call implicit_schema, it assigns $schema in the nested adq, but maybe I remember wrong.

paleolimbot (commented):
First, I left a note about how you might be able to get the S3 to work although there's no perfect way to go about it (notably, `schema(!!!stuff)` is going to be problematic because the first argument will always be evaluated by not rlang).

Second, how about:

- The S3 generic is `infer_schema()`. We already have `infer_type()` (changed from `type()` for similar reasons)...`as_schema()` definitely feels wrong for all the reasons that have been mentioned.
- We use `infer_schema()` internally, but
- If somebody calls `schema(one_unnamed_arg)`, dispatch to `infer_schema()` for interactive use. I too have typed `schema(something)` and done mild cursing because I couldn't remember what combination of `$` and `as_whatever()` I needed.

thisisnic:
Thanks @paleolimbot, sounds like a good middle ground there, will give that a go.

r/R/schema.R
```diff
@@ -240,10 +226,48 @@ print_schema_fields <- function(s) {
...
+#' schema(tab2)
+#' @export
+schema <- function(...){
+  UseMethod("schema")
+}
+
 #' @export
-#' @rdname Schema
-schema <- Schema$create
+schema.default <- Schema$create
```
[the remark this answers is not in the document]
thisisnic: I don't understand this; how can `..object` be missing, but `...` be present?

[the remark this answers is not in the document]
paleolimbot: (i.e., if you only pass named arguments, `...` will be present but `..object` will not)

r/R/schema.R
```diff
@@ -244,10 +230,59 @@ print_schema_fields <- function(s) {
...
+  dots <- list2(...)
+
+  if (length(dots) == 1 && !is.list(dots[[1]]) && is.null(names(dots)) && !inherits(dots[[1]], "Field")) {
+    return(infer_schema(dots[[1]]))
+  }
+
+  Schema$create(...)
+}
+
+#' Create a schema or extract one from an object.
```
paleolimbot, line 267:
```suggestion
#' Extract a schema from an object
```

(maybe?)

r/R/schema.R
```diff
@@ -244,10 +230,59 @@ print_schema_fields <- function(s) {
...
+
+#' @export
+infer_schema.ArrowTabular <- function(x) x$schema
+
+#' @export
+infer_schema.RecordBatchReader <- function(x) x$schema
+
+#' @export
+infer_schema.Dataset <- function(x) x$schema
+
```
paleolimbot, line 283: I think you added a way to do this for `data.frame` already that might make sense to wire up here?

thisisnic: That feels beyond the scope of this PR, happy to do it in a follow-up if you open a ticket.

r/R/schema.R
```diff
@@ -244,10 +230,59 @@ print_schema_fields <- function(s) {
...
+#' df <- data.frame(col1 = 2:4, col2 = c(0.1, 0.3, 0.5))
+#' tab1 <- arrow_table(df)
+#' schema(tab1)
+#' tab2 <- arrow_table(df, schema = schema(col1 = int8(), col2 = float32()))
+#' schema(tab2)
+#' @export
+schema <- function(...) {
+  dots <- list2(...)
+
+  if (length(dots) == 1 && !is.list(dots[[1]]) && is.null(names(dots)) && !inherits(dots[[1]], "Field")) {
```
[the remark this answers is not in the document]
thisisnic: I thought about this, but `infer_schema.Field()` doesn't quite fit, as we're creating rather than inferring a schema with it. 

Thanks for the list suggestion, added.

r/R/schema.R
```diff
@@ -244,10 +230,59 @@ print_schema_fields <- function(s) {
...
+    return(infer_schema(dots[[1]]))
+  }
+
+  Schema$create(...)
```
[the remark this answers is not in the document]
thisisnic: Nah, it doesn't make any difference.  I've intentionally left it as-is as the intention is more obvious like that.

r/R/dplyr-collect.R
```diff
@@ -192,5 +192,6 @@ implicit_schema <- function(.data) {
       aggregate_types(.data, hash, old_schm)
     )
   }
-  schema(!!!new_fields)
+
+  schema(new_fields)
```
paleolimbot, line 196: If you feel strongly about the ability to pass a bare list instead of `!!!bare_list` I think this is a better fit for a separate PR: it is not related to the ability to use `schema()` on a Dataset/ArrowTabular/arrow_dplyr_query and is a departure from behaviour elsewhere in the package (e.g., none of `record_batch()`, `arrow_table()`, or `struct()` support this).

r/R/schema.R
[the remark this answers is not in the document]
paleolimbot: As above, if you feel strongly about this I think it is a better fit for another PR: our usage everywhere else in the package is to capture the dots once using `list2()` and use `!!!` if they need to be passed on to another function.

r/tests/testthat/test-schema.R
```diff
@@ -292,3 +292,18 @@ test_that("schema name assignment", {
   expect_identical(names(schm2), c("col1", "col2"))
   expect_identical(names(schm2$r_metadata$columns), c("col1", "col2"))
 })
+
+test_that("schema extraction", {
+  tbl <- arrow_table(example_data)
```
paleolimbot, line 297: This might need `skip_if_not_available("dataset")` to pass the minimal nightly tests

r/R/schema.R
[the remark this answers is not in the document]
thisisnic: If the usage elsewhere is like this then I'm happy to change it here.  You didn't say that in your previous feedback. I don't really care what we do here given it doesn't actually matter in any meaningful way.  I will update it to your suggestion.

r/tests/testthat/test-schema.R
```diff
@@ -292,3 +292,9 @@ test_that("schema name assignment", {
...
+  schema(tbl)
+  expect_equal(schema(tbl), tbl$schema)
+})
```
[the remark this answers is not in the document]
thisisnic: I'd thought I'd commented here but apparently not; we already test `schema(some_list)` further up:

https://github.com/apache/arrow/blob/5854d0405366c025427919470327318f0089e03a/r/tests/testthat/test-schema.R#L254-L268

r/tests/testthat/test-schema.R
thisisnic: Thanks, updated!

r/R/dplyr-collect.R
paleolimbot: I'm sorry, I had no idea it was an existing feature. Could it be in the documentation for `...` in `schema()`?

paleolimbot (approved):
Looks great! Thank you!

PR #35248, opened 2023-04-20
GH-35235: [CI][Python] Pandas upstream_devel and nightlies are failing

AlenkaF:
* Closes: #35235

PR #35546, opened 2023-05-11
GH-35250: [Python] Add test for datetime column conversion to pandas

### LINKED ISSUE #35250: [Python] Add explicit test for passing of name to pandas_api.series in ChunkedArray

This is a follow up to https://github.com/apache/arrow/pull/35248 which fixed failing tests for new versions of pandas. We should add explicit tests for passing of `name` to `pandas_api.series` in ChunkedArray.

See: https://github.com/apache/arrow/pull/35248/files#r1172412311

### Component(s)

Python

AlenkaF:
### Rationale for this change
In pandas version 2.1.0 `__from_arrow__` method has been implemented for `DatetimeTZDtype` extension type. See https://github.com/pandas-dev/pandas/pull/52201.

Due to this change, the conversion from pyarrow table column with datetime uses the newly implemented `__from_arrow__` method for the pandas dtype and `pandas_api.series` to construct the pandas series. We had to pass the name in the `pandas_api.series` also, to fix failing tests in https://github.com/apache/arrow/pull/35248.

This PR adds explicit test for this change.
* Closes: #35250

### DIFF
--- a/python/pyarrow/tests/test_pandas.py
+++ b/python/pyarrow/tests/test_pandas.py
@@ -4822,16 +4822,10 @@ def test_column_conversion_for_datetime():
     # GH-35235
     # pandas implemented __from_arrow__ for DatetimeTZDtype
     # https://github.com/pandas-dev/pandas/pull/52201
-    if Version(pd.__version__) <= Version("2.0.1"):
-        pytest.skip("__from_arrow__ added to DatetimeTZDtype in pandas version 2.1.0")
-
     arr = pd.Series(pd.date_range("2012", periods=2, tz="Europe/Brussels"),
                     name="datetime_column")
     table = pa.table({"datetime_column": pa.array(arr)})
     table_col = table.column("datetime_column")
-
-    pandas_dtype = table_col.type.to_pandas_dtype()
-    assert hasattr(pandas_dtype, '__from_arrow__')
 
     result = table_col.to_pandas()
     assert result.name == "datetime_column"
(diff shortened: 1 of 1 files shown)


python/pyarrow/tests/test_pandas.py
```diff
@@ -4816,3 +4816,23 @@ def test_unhashable_map_keys_with_pydicts():
...
             assert np.array_equal(tup1[0], tup2[0])
             assert tup1[1] == tup2[1]
+
+
+def test_column_conversion_for_datetime():
+    # GH-35235
+    # pandas implemented __from_arrow__ for DatetimeTZDtype
+    # https://github.com/pandas-dev/pandas/pull/52201
+    if Version(pd.__version__) <= Version("2.0.1"):
+        pytest.skip("__from_arrow__ added to DatetimeTZDtype in pandas version 2.1.0")
```
jorisvandenbossche, line 4826: Do we need to skip the test for older pandas? I would assume that everything (except for the `assert hasattr(pandas_dtype, '__from_arrow__')`) should work the same with older pandas? (essentially this is testing that we _preserve_ this behaviour with newer pandas)

AlenkaF: Yes, you are correct!
I will put the main part of the test outside this skip and leave `assert hasattr(pandas_dtype, '__from_arrow__')` in it.

jorisvandenbossche: Or just leave out that check, in the end that's testing an implementation detail of pandas, not something we control

jorisvandenbossche (approved):

PR #35548, opened 2023-05-11
GH-34789: [CI][Python] Update the Python versions in our nightly CI matrix

### LINKED ISSUE #34789: [CI][Python] Update the Python versions in our nightly CI matrix
Generally, we tend to test with quite old python versions. For the core tests like the `test-conda-python-xx` builds, we test on the full range of supported python versions (3.7-3.11). But for some other custom integrations tests, we test mostly with old versions:

https://github.com/apache/arrow/blob/ddd0a337174e57cdc80b1ee30dc7e787acfc09f6/dev/tasks/tasks.yml#L1506-L1513

Pandas doesn't actually support python 3.7 anymore, so because we use that version for the "latest" build, we don't actually get latest pandas.

Also for example the HDFS tests are using python 3.7

jorisvandenbossche:
* Closes: #34789

### DIFF
--- a/dev/tasks/tasks.yml
+++ b/dev/tasks/tasks.yml
@@ -1193,24 +1193,24 @@ tasks:
       image: conda-python
 {% endfor %}
 
-  test-conda-python-3.8-hypothesis:
+  test-conda-python-3.11-hypothesis:
     ci: github
     template: docker-tests/github.linux.yml
     params:
       env:
-        PYTHON: 3.8
+        PYTHON: 3.11
         HYPOTHESIS_PROFILE: ci
         PYARROW_TEST_HYPOTHESIS: ON
         # limit to execute hypothesis tests only
         PYTEST_ARGS: "-m hypothesis"
       image: conda-python-pandas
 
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
@github-actions crossbow submit test-conda-python-3.11-hypothesis test-conda-python-3.10-substrait

raulcd (commented):
Thanks @jorisvandenbossche! Any reason why for some of the jobs we have bumped to Python 3.10 and for others to Python 3.11?

jorisvandenbossche:
Just to have a bit variation in the versions, I didn't think too much about it. If something seems odd, feel free to comment on it!

raulcd:
Looks good to me, just wondering whether there was any reason behind it! Thanks!

jorisvandenbossche:
@github-actions crossbow submit  test-conda-python-3.10-hdfs-2.9.2 test-conda-python-3.10-hdfs-3.2.1

jorisvandenbossche:
The HDFS ones are failing on main with the same error (which seems an actual test failure to fix), so merging

PR #35554, opened 2023-05-11
GH-35553: [JAVA] Fix unwrap() in NettyArrowBuf

### LINKED ISSUE #35553: [Java] unwrap() in NettyArrowBuf causes breaks when debug / error messages are logged

The implementation of NettyArrowBuf unwrap() is throwing UnsupportedOperationException, but the ByteBuf interface says unwrap() should return null when a buffer cannot be unwrapped.  Netty's AbstractByteBuf base class calls unwrap() in toString() and prints extra info if the result is non-null. Since toString() is called every time a debug or error message is printed, real errors are getting masked and enabling debug logging is likely to cause unexpected errors.

Seems like a simple fix just to return null from unwrap() in NettyArrowBuf? I will submit a PR with the change if someone wants to take a look.

### Component(s)

Java

martin-traverse:
Suggested fix for #35553

No API changes, this just brings the implementation in line with the API as specified by Netty.

CI for Java run successfully here (not sure if this is publically visible):

https://github.com/martin-traverse/arrow/actions/runs/4949700024

* Closes: #35553

### DIFF
--- a/java/memory/memory-netty/src/main/java/io/netty/buffer/NettyArrowBuf.java
+++ b/java/memory/memory-netty/src/main/java/io/netty/buffer/NettyArrowBuf.java
@@ -110,7 +110,14 @@ public class NettyArrowBuf extends AbstractByteBuf implements AutoCloseable {
 
   @Override
   public ByteBuf unwrap() {
-    throw new UnsupportedOperationException("Unwrap not supported.");
+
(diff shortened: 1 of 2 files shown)


lidavidm (approved):
Thanks!

Is it possible to add a small unit test?

lidavidm (approved):

Issue #32276, opened 2022-07-01
[C++][FlightRPC] Flight generates misaligned buffers


asfimport:
Protobuf's wire format design + our zero-copy serializer/deserializer mean that buffers can end up misaligned. On some Arrow versions, this can cause segfaults in kernels assuming alignment (and generally violates expectations). 

We should:
- Possibly include buffer alignment in array validation
- See if we can adjust the serializer to somehow pad things properly
- See if we can do anything about this in the deserializer
  
  Example:
  ```python
  
  import pyarrow as pa
  import pyarrow.flight as flight
  
  class TestServer(flight.FlightServerBase):
      def do_get(self, context, ticket):
          schema = pa.schema(
              [
                  ("index", pa.int64()),
                  ("int8", pa.float64()),
                  ("int16", pa.float64()),
                  ("int32", pa.float64()),
              ]
          )
          return flight.RecordBatchStream(pa.table([
              [0, 1, 2, 3],
              [0, 1, None, 3],
              [0, 1, 2, None],
              [0, None, 2, 3],
          ], schema=schema))
  
  
  with TestServer() as server:
      client = flight.connect(f"grpc://localhost:{server.port}")
      table = client.do_get(flight.Ticket(b"")).read_all()
      for col in table:
          print(col.type)
          for chunk in col.chunks:
              for buf in chunk.buffers():
                  if not buf: continue
                  print("buffer is 8-byte aligned?", buf.address % 8)
              chunk.cast(pa.float32())
  ```
  
  On Arrow 8
  ```
  
  int64
  buffer is 8-byte aligned? 1
  double
  buffer is 8-byte aligned? 1
  buffer is 8-byte aligned? 1
  double
  buffer is 8-byte aligned? 1
  buffer is 8-byte aligned? 1
  double
  buffer is 8-byte aligned? 1
  buffer is 8-byte aligned? 1
  ```
  On Arrow 7
  ```
  
  int64
  buffer is 8-byte aligned? 4
  double
  buffer is 8-byte aligned? 4
  buffer is 8-byte aligned? 4
  fish: Job 1, 'python ../test.py' terminated by signal SIGSEGV (Address boundary error)
  ```

**Reporter**: [David Li](https://issues.apache.org/jira/browse/ARROW-16958) / @lidavidm

[...]

PR #10290, opened 2021-05-10
ARROW-12725: [C++][Compute] Column at a time hash and comparison in group by

michalursa:
Current implementation of hash group by (GrouperFastImpl class) converts input ExecBatches to row-oriented format,
then hashes and compares rows as if they were a single column. 

It is more efficient to avoid relatively costly encoding and instead compute hashes of individual columns in column-oriented format mixing them together, and similarly comparing column-oriented data to row-oriented data in the hash table without converting. 

Encoding only happens for a subset of input rows that are inserted into the hash table - they introduce new groups. 
Keys in hash table remain stored as row-oriented.

There are 3 components that changed:
- key_compare.cc - contains implementation of one column at a time comparison between column-oriented and row-oriented data
- key_hash.cc - implements hashing individual columns of column-oriented key data and mixing the individual results together for a final multi-column hash
- key_encode.cc - implements encoding in row-format only a selection of input rows; previously all input rows were encoded without support for selective encoding

Performance before and after:
Here are selected results (in millions of rows per second) of arrow-compute-aggregate-benchmark for 0.01% nulls in input columns (Ubuntu VM, Clang 10.0, Intel I9-10980XE): 
| Test | Before (AVX2) | After (AVX2) | Before (Scalar) | After (Scalar) |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| TinyString  | 53  | 96 | 25 | 38 |
| SmallString  | 48 | 80 | 23 | 36 |
| MediumString  | 47 | 65 | 22 | 32 |
| TinyInt  | 340  | 407 | 130| 136 |
| SmallInt  | 246  | 287 | 131 | 137 |
| MediumInt  | 194 | 208 | 106 | 111 |

PR #35565, opened 2023-05-11
GH-35498: [C++] Relax EnsureAlignment check in Acero from requiring 64-byte aligned buffers to requiring value-aligned buffers

### LINKED ISSUE #35498: [C++][Parquet] Parquet write_to_dataset performance regression

westonpace:
### Rationale for this change

Various compute kernels and Acero internals rely on type punning.  This is only safe when the buffer has appropriate alignment (e.g. casting uint8_t* to uint32_t* is only safe if the buffer has 4-byte alignment).  To avoid errors we enforced 64-byte alignment in Acero.  However, this is too strict.  While Arrow's allocators will always generate 64-byte aligned buffers this is not the case for numpy's allocators (and presumably many others).  This PR relaxes the constraint so that we only require value-aligned buffers.

The main complexity here is determining which buffers need aligned and how much.  A special flag kMallocAlignment is added which can be specified when calling CheckAlignment or EnforceAlignment to only require value-alignment and not a particular number.

### Are these changes tested?

Yes

No
* Closes: #35498


westonpace:
@ursabot please benchmark

westonpace:
@ursabot please benchmark

westonpace:
> ['Python', 'R'] benchmarks have high level of regressions.

I'm running the benchmarks again but, as best I can tell, these regressions are noise, though it is quite difficult to say for sure.  E.g. on the 9960x there are 5 non-js regressions and 21 non-js improvements.  On arm I see the same improvements and a bunch of regressions unrelated to the PR.

westonpace:
I have tested the reproduction case in the original issue.  Without this change I get about 10x worse than arrow 11.  With this change I get the same performance as arrow 11.

westonpace:
CC  @rtpsw / @sanjibansg

cpp/src/arrow/util/align_util.cc
```diff
@@ -30,12 +32,120 @@ bool CheckAlignment(const Buffer& buffer, int64_t alignment) {
...
 }
 
-bool CheckAlignment(const ArrayData& array, int64_t alignment) {
-  for (const auto& buffer : array.buffers) {
-    if (buffer) {
-      if (!CheckAlignment(*buffer, alignment)) return false;
+namespace {
+
+// Some buffers are frequently type-punned.  For example, in an int32 array the
+// values buffer is frequently cast to int32_t*
```
felipecrv, line 38: Can't we expect at least 8-byte alignment? :/

cpp/src/arrow/util/align_util.cc
```diff
@@ -30,12 +32,120 @@ bool CheckAlignment(const Buffer& buffer, int64_t alignment) {
...
+                                   // check before casting to an array of structs
+    case Type::BOOL:               // Always treated as uint8_t*
+    case Type::INT8:               // Always treated as uint8_t*
+    case Type::UINT8:              // Always treated as uint8_t*
+    case Type::DECIMAL128:         // Always treated as uint8_t*
+    case Type::DECIMAL256:         // Always treated as uint8_t*
+    case Type::SPARSE_UNION:       // Types array is uint8_t, no offsets array
+    case Type::RUN_END_ENCODED:    // No buffers
+    case Type::STRUCT:             // No buffers beyond validity
+      // These have no buffers or all buffers need only byte alignment
```
felipecrv, line 81: Shouldn't this be recursive for `RUN_END_ENCODED`?

```cpp
max(GetMallocValuesAlignment(*array.run_ends()),
    GetMallocValuesAlignment(*array.values()));
```

cpp/src/arrow/util/align_util.cc
```diff
@@ -30,12 +32,120 @@ bool CheckAlignment(const Buffer& buffer, int64_t alignment) {
...
+// return a buffer that is 4 byte aligned) or common libraries (e.g. numpy) but it is
+// not currently guaranteed by flight (GH-32276).
+//
+// By happy coincedence, for every data type, the only buffer that might need wider
+// alignment is the second buffer (at index 1).  This function returns the expected
+// alignment (in bits) of the second buffer for the given array to safely allow this cast.
+//
+// If the array's type doesn't have a second buffer or the second buffer is not expected
+// to be type punned, then we return 8.
+int GetMallocValuesAlignment(const ArrayData& array) {
```
felipecrv, line 55: Name suggestion: `RequiredBuffersAlignment`. I was confused by the "Get" when it's not getting the alignment of the existing buffers, but calculating the desired/required alignment for this array.

jorisvandenbossche:
> I am not sure why it doesn't show up on the landing page linked from the bot comment

Whoops, they of course _do_ show up, just have to sort by the z-score or change percentage in decreasing order, so that all the positive values show first.

cpp/src/arrow/util/align_util.cc
pitrou, line 55: Since an array can have several buffers, I think this is too coarse-grained. Let's have something like:
```c++
int GetRequiredBufferAlignment(const DataType& type, int buffer_index) {
  if (buffer_index == 0) {
    // Either null bitmap or 8-bit union type ids
    return 1;
  }
  switch (type.id()) {
    case Type::INT16:
    case Type::UINT16:
    case Type::HALF_FLOAT:
      return 2;
    case Type::INT32:
    case Type::UINT32:
    case Type::FLOAT:
    case Type::DATE32:
    case Type::TIME32:
    case Type::LIST:         // Offsets may be cast to int32_t*, data is in child array
    case Type::MAP:          // This is a list array
    case Type::DENSE_UNION:  // Has an offsets buffer of int32_t*
    case Type::INTERVAL_MONTHS:  // Stored as int32_t*
      return 4;
    case Type::INT64:
    case Type::UINT64:
    case Type::DOUBLE:
    case Type::LARGE_LIST:    // Offsets may be cast to int64_t*
    case Type::DATE64:
    case Type::TIME64:
    case Type::TIMESTAMP:
    case Type::DURATION:
    case Type::INTERVAL_DAY_TIME:  // Stored as two contiguous 32-bit integers but may be
                                   // cast to struct* containing both integers
      return 8;
    case Type::INTERVAL_MONTH_DAY_NANO:  // Stored as two 32-bit integers and a 64-bit
                                         // integer
      return 16;
    case Type::STRING:
    case Type::BINARY:  // Offsets may be cast to int32_t*, data is only uint8_t*
      return (buffer_index == 1) ? 4 : 1;
    case Type::LARGE_STRING:
    case Type::LARGE_BINARY:  // Offsets may be cast to int64_t*
      return (buffer_index == 1) ? 8 : 1;
    default:
      // Everything else doesn't have buffers with non-trivial alignement requirements
      return 1;
  }
}
```

pitrou:
> e.g. casting uint8_t* to uint32_t* is only safe if the buffer has 4-byte alignment

I'm not sure where you got that from the C++ spec?

If I try to understand https://en.cppreference.com/w/cpp/language/reinterpret_cast (especially the "Type Aliasing" paragraph), it seems only `uint32_t*` to `uint8_t*` is spec-compliant, not the other way round.

> Whenever an attempt is made to read or modify the stored value of an object of type DynamicType through a glvalue of type AliasedType, the behavior is undefined unless one of the following is true:

>  * AliasedType and DynamicType are similar.
>  * AliasedType is the (possibly [cv](https://en.cppreference.com/w/cpp/language/cv)-qualified) signed or unsigned variant of DynamicType.
>  * AliasedType is [std::byte](https://en.cppreference.com/w/cpp/types/byte), (since C++17) char, or unsigned char: this permits examination of the [object representation](https://en.cppreference.com/w/cpp/language/object#Object_representation_and_value_representation) of any object as an array of bytes.

bkietz (changes requested):
> So ideally I think we should remove the entire code that reallocates buffers to fix their alignment, but short from that we should strive to be as conservative and granular as possible (see the suggestion I posted above).

I tend to agree with narrowing the scope of alignment enforcement: it's part of the contract of the arrow format that buffers be aligned appropriately for the integers etc stored in them. Therefore to my mind we should rather avoid ever constructing an ArrayData which would fail CheckAlignment. If there is a producer which doesn't provide that guarantee (flight, GH-32276) then alignment should be enforced on those buffers before they escape the scope of the faulty producer. In the context of this PR, that'd mean: we *only* enforce alignment on buffers coming out of flight- and that may be the only producer for which we *check* alignment until some other producer is proven to occasionally emit misaligned data.

> e.g. casting uint8_t* to uint32_t* is only safe if the buffer has 4-byte alignment

To be more pedantic about this: It is only legal to reinterpret_cast from a pointer to bytes to some other type T if those bytes are storage for an existing (lifetime has started) object of type T (or a type similar to T). The bytes which we're casting must therefore be aligned to `alignof(T)` because an object of T can never start its lifetime in unaligned storage. When we mmap some bytes which correspond to a buffer of signed 32 bit integers, we're implicitly treating those bytes as storage where integers have started their lifetime (since we're not in c++23 we can't make this explicit with [`start_lifetime_as_array`](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2679r2.pdf)). Since integers are fairly trivial (and this sort of technically-UB-but-what-could-go-wrong is so widely depended upon in C++ development), this is a liberty I don't expect any compiler to punish us for.

pitrou:
> I tend to agree with narrowing the scope of alignment enforcement: it's part of the contract of the arrow format that buffers be aligned appropriately for the integers etc stored in them.

To be nitpicky, the Arrow format doesn't _require_ anything about alignment. Implementations can choose to enforce some requirements if they wish to (which implies the kind of opportunistic realignment shenanigans discussed in this PR).

Also, my view is that the problems with unaligned buffers are largely theoretical until a concrete problem is reliably diagnosed as such.

westonpace:
I tried for a while to reproduce some kind of alignment related issue on godbolt and was unsuccessful.  I then did some research.  Most of the examples I could find either:

A. Demonstrated potential performance penalties (which were often cpu-specific) associated with aligned loads
B. Were specific to 32-bit ARM processors (which crash on unaligned loads)

I think A is mitigated by the fact that any performance benefit is unlikely to outweigh the cost of the allocation / copy required by EnsureAlignment.  B is not important as we don't support 32-bit builds (outside of maybe windows 32-bit) that I'm aware of.

The root of all this discussion is from some assertions in (what is now) `compare_internal.cc` which would DCHECK that a buffer was aligned.  These appear to have been added defensively, and not in response to any actual issue: https://github.com/apache/arrow/pull/10290#discussion_r688021792

So how do we feel about:

 * Proceed with adding this capability to `CheckAlignment` / `EnsureAlignment` (the utility could still be a useful feature for someone that needs aligned buffers for some other reason)
 * Remove all defensive alignment assertions (this is a DCHECK to ensure that a buffer is aligned that is sometimes used before type punning)
 * Change the call in `SourceNode` from "ensure buffers are malloc aligned" to "check if buffers are malloc aligned" and, if not, log a warning along the lines of `One or more buffers are unaligned.  This may lead to suboptimal performance.  Check to see if it is possible for the data source to generate aligned buffers.`

CC @bkietz @pitrou @felipecrv

rtpsw:
> I tried for a while to reproduce some kind of alignment related issue on godbolt and was unsuccessful

But a misalignment can be reproduced [using Flight](https://github.com/apache/arrow/issues/32276), right?

> Remove all defensive alignment assertions

I can confirm this resolved the internal problem on my platform. Do we know that no other supported platform would have a problem with misalignment? Perhaps it's worth adding test-cases for tolerance to misalignment, even if they could only generate warnings?

felipecrv:
> A. Demonstrated potential performance penalties (which were often CPU-specific) associated with the unaligned loads

"A" makes compute kernel performance more noisy and unpredictable. IMO, that alone could be a reason to do our best to feed aligned data to compute kernels even with the potential alloc and copy before the kernel runs.

> B. Were specific to 32-bit ARM processors (which crash on unaligned loads)

"B" is a sign that relying on unaligned memory access makes us much more exposed to CPU bugs. In trusting modern CPUs unaligned-memory access machinery, we expose ourselves to risk. That machinery is even less trustworthy on   SIMD instruction sets.

CPU vendors started supporting unaligned memory access as a response to bad software, not because it makes for elegant and reliable hardware design.

westonpace:
CC @lidavidm too who might be interested in the discussion.

**I'd like to pick something and move forward on it soon.** There is a significant performance regression that affects a very common use case (data from numpy) which we all agree should be fixed (we all agree we should not be requiring 64-byte alignment) and I think it may justify a 12.0.1 patch.

I do not want to get too delayed with discussion on how we should handle a considerably more rare case (buffers from flight) that it isn't clear has extensive use in production.

If there is no clear consensus by tomorrow then I will probably just pick something and advocate that we go with that for now and try to get it merged in early next week.

lidavidm:
I'll take up the Flight issue for 13.0.0.

But yes, it wasn't clear to me whether the original issue was related to alignment at all.

cpp/src/arrow/util/align_util.cc
```diff
@@ -30,12 +32,120 @@ bool CheckAlignment(const Buffer& buffer, int64_t alignment) {
...
+    case Type::FIXED_SIZE_BINARY:  // Fixed size binary could be dangerous but the
+                                   // compute kernels don't type pun this.  E.g. if
+                                   // an extension type is storing some kind of struct
+                                   // here then the user should do their own alignment
+                                   // check before casting to an array of structs
+    case Type::BOOL:               // Always treated as uint8_t*
+    case Type::INT8:               // Always treated as uint8_t*
+    case Type::UINT8:              // Always treated as uint8_t*
+    case Type::DECIMAL128:         // Always treated as uint8_t*
+    case Type::DECIMAL256:         // Always treated as uint8_t*
```
mapleFU, line 77:
```c++
struct DecimalToIntegerMixin {
  template <typename OutValue, typename Arg0Value>
  OutValue ToInteger(KernelContext* ctx, const Arg0Value& val, Status* st) const {
    constexpr auto min_value = std::numeric_limits<OutValue>::min();
    constexpr auto max_value = std::numeric_limits<OutValue>::max();

    if (!allow_int_overflow_ && ARROW_PREDICT_FALSE(val < min_value || val > max_value)) {
      *st = Status::Invalid("Integer value out of bounds");
      return OutValue{};  // Zero
    } else {
      return static_cast<OutValue>(val.low_bits());
    }
  }
```

I'm not so familiar with acero, so I wonder why `Decimal128` and `Decimal256` can use 1B, isn't them `Decimal128` and `Decimal256`?

felipecrv:
I feel like the discussion went a little bit too far on the merits of alignment and where it should be enforced, when the change @westonpace is proposing here (relaxing the very aggressive 64-BYTE alignment) should not be controversial at all: it doesn't even violate the theoretical UB concerns from compilers.

I'm going to review the code and we should move the general alignment discussions to the issue instead of having it on this PR.

cpp/src/arrow/util/align_util.h
```diff
@@ -70,8 +70,26 @@ inline BitmapWordAlignParams BitmapWordAlign(const uint8_t* data, int64_t bit_of
...
 
 // Functions to check if the provided Arrow object is aligned by the specified alignment
+
+/// \brief if this is specified in one of the CheckAlignment or EnsureAlignment functions
+/// then the funciton will ensure each buffer is suitably aligned for the data type of the
+/// array.  For example, given an int32 buffer the validity buffer must be a multiple of 8
+/// and the values buffer must be a multiple of 32.  Given a large_string buffer the
+/// validity buffer and values buffers must be multiples of 8 and the offsets buffer must
+/// be a multiple of 64.
+constexpr int64_t kMallocAlignment = -3;
```
felipecrv, line 80: This has potential for becoming quite confusing as `malloc` does indeed have an alignment that is dependent on system/toolchain. Could this be named something like `kTypeSpecificAlignment` instead?

felipecrv (commented):
LGTM. Left some comments about naming.

cpp/src/arrow/util/align_util.cc
westonpace: This has been renamed to `RequiredValueAlignmentForBuffer` and it takes in a type now instead of an array.

cpp/src/arrow/util/align_util.cc
westonpace: Renamed to `RequiredValueAlignmentForBuffer`

westonpace: Done, except I return `1` if `buffer_index > 1` also (e.g. binary arrays)

cpp/src/arrow/type_traits.cc
```diff
@@ -0,0 +1,94 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+#include "arrow/type_traits.h"
+
+#include "arrow/util/logging.h"
+
+namespace arrow {
+
+namespace {}  // namespace
```
felipecrv, line 24: Early close

cpp/src/arrow/util/align_util.cc
westonpace: I've updated the expected alignment for decimal128/decimal256 arrays to 8.

cpp/src/arrow/type_traits.cc
```diff
@@ -0,0 +1,94 @@
...
+    case Type::LARGE_LIST:    // Offsets may be cast to int64_t*
+    case Type::LARGE_STRING:  // Offsets may be cast to int64_t*
+    case Type::DATE64:
+    case Type::TIME64:
+    case Type::TIMESTAMP:
+    case Type::DURATION:
+    case Type::INTERVAL_MONTH_DAY_NANO:  // Stored as two 32-bit integers and a 64-bit
+                                         // integer
+      return 8;
+    default:
```
felipecrv, line 86: Can we keep this without a `default` so when new types are added by me and @bkietz we get a compiler error reminding us to think about this function?

cpp/src/arrow/type_traits.cc
westonpace: Oops.  There used to be a function there :).  I've fixed this now.

cpp/src/arrow/util/align_util.h
```diff
@@ -72,12 +72,13 @@ namespace util {
...
 
 /// \brief if this is specified in one of the CheckAlignment or EnsureAlignment functions
+///
 /// then the funciton will ensure each buffer is suitably aligned for the data type of the
 /// array.  For example, given an int32 buffer the validity buffer must be a multiple of 8
 /// and the values buffer must be a multiple of 32.  Given a large_string buffer the
 /// validity buffer and values buffers must be multiples of 8 and the offsets buffer must
 /// be a multiple of 64.
-constexpr int64_t kMallocAlignment = -3;
+constexpr int64_t kValueAlignment = -3;
```
felipecrv, line 81: Nice!

cpp/src/arrow/type_traits.cc
westonpace: Thanks for pointing this out.  I thought a switch case without a default just silently ignored options but I was wrong.

westonpace: I've removed the default (and manually listed the remaining invalid options)

westonpace:
> This seems fine to me, with the caveat that always emitting a warning is inflexible for users who definitely want to avoid unaligned access. Perhaps they're building for a platform we're not aware of where unaligned access has more disastrous consequences than we've seen, or maybe they're just running with a sanitizer which detects unaligned access and emits an error even if the platform would allow the access. Therefore, I think that for the time being there should be a build option or at least an environment variable which such a user could set to ensure alignment.

How about a runtime option (proposal updated to include this)?

> Also I think both the build option and the warning should be tagged with the number of an issue tracking (or, later: a link to documentation on) producers of unaligned data and how to mitigate that

In theory the user could be feeding their own unaligned data completely unrelated to any Arrow component and so I'm not sure how to accomplish this.

cpp/src/arrow/util/align_util.cc
mapleFU: Thanks, though I think maintaining a `std::array<uint64_t, NWORDS>;` for Decimal, and use `8` as magic number for alignas is a bit hacking. I'm +1 on this change. Maybe we can use some alignas constexpr later?

westonpace:
I went ahead and added the environment variable too.  Now it should be fully configurable.  I tested the environment variable manually:

```
./debug/arrow-acero-plan-test --gtest_filter=ExecPlanExecution.UnalignedInput
...
/home/pace/dev/arrow/cpp/src/arrow/acero/source_node.cc:76: An input buffer was poorly aligned.  This could lead to crashes or poor performance on some hardware.  Please ensure that all Acero sources generate aligned buffers.
...
[  PASSED  ] 1 test.

ACERO_ALIGNMENT_HANDLING=abort ./debug/arrow-acero-plan-test --gtest_filter=ExecPlanExecution.UnalignedInput
...
/home/pace/dev/arrow/cpp/src/arrow/acero/plan_test.cc:1724: Failure
Failed
'DeclarationToStatus(plan)' failed with Invalid: An input buffer was poorly aligned and UnalignedBufferHandling is set to kAbort
/home/pace/dev/arrow/cpp/src/arrow/acero/source_node.cc:155  HandleUnalignedBuffers(&batch, unaligned_buffer_handling)
[  FAILED  ] ExecPlanExecution.UnalignedInput (13 ms)
...

ACERO_ALIGNMENT_HANDLING=ignore ./debug/arrow-acero-plan-test --gtest_filter=ExecPlanExecution.UnalignedInput
...
[  PASSED  ] 1 test.

ACERO_ALIGNMENT_HANDLING=warn ./debug/arrow-acero-plan-test --gtest_filter=ExecPlanExecution.UnalignedInput
...
/home/pace/dev/arrow/cpp/src/arrow/acero/source_node.cc:76: An input buffer was poorly aligned.  This could lead to crashes or poor performance on some hardware.  Please ensure that all Acero sources generate aligned buffers.
...
[  PASSED  ] 1 test.

ACERO_ALIGNMENT_HANDLING=reallocate ./debug/arrow-acero-plan-test --gtest_filter=ExecPlanExecution.UnalignedInput
...
/home/pace/dev/arrow/cpp/src/arrow/acero/plan_test.cc:1725: Failure
  initial_bytes_allocated
    Which is: 256
  default_memory_pool()->total_bytes_allocated()
    Which is: 320
[  FAILED  ] ExecPlanExecution.UnalignedInput (15 ms)
...
```

cpp/src/arrow/acero/exec_plan.cc
```diff
@@ -358,10 +359,38 @@ std::optional<int> GetNodeIndex(const std::vector<ExecNode*>& nodes,
   return std::nullopt;
 }
 
+const std::string kAceroAlignmentHandlingEnvVar = "ACERO_ALIGNMENT_HANDLING";
```
felipecrv, line 362: C-style string as suggested by the linter is better for constants. And that's what `GetEnvVar` takes under-the hood, so a heap-alloc for std::string never happens.

cpp/src/arrow/acero/source_node.cc
```diff
@@ -143,8 +143,14 @@ struct SourceNode : ExecNode, public TracedNode {
...
             ExecBatch batch = morsel.Slice(offset, batch_size);
-            UnalignedBufferHandling unaligned_buffer_handling =
+            std::optional<UnalignedBufferHandling> opt_unaligned_buffer_handling =
                 plan_->query_context()->options().unaligned_buffer_handling;
+            UnalignedBufferHandling unaligned_buffer_handling;
+            if (opt_unaligned_buffer_handling.has_value()) {
+              unaligned_buffer_handling = *opt_unaligned_buffer_handling;
+            } else {
+              unaligned_buffer_handling = GetDefaultUnalignedBufferHandling();
+            }
```
felipecrv: ternary use maybe?

```cpp
auto unaligned_buffer_handling = opt_unaligned_buffer_handling.has_value() ?
      *opt_unaligned_buffer_handling : 
      GetDefaultUnalignedBufferHandling();
```

cpp/src/arrow/type_traits.cc
```diff
@@ -86,9 +86,13 @@ int RequiredValueAlignmentForBuffer(Type::type type_id, int buffer_index) {
     case Type::DICTIONARY:
     case Type::EXTENSION:
     case Type::MAX_ID:
-      Status::Invalid("Could not check alignment for type id ", type_id).Warn();
+      Status::Invalid("RequiredValueAlignmentForBuffer called with invalid type id ",
+                      type_id)
+          .Warn();
```
felipecrv, line 91: You could put a `break` here and move this message to where you added the other status before the `return 1` at the end. Both situations are only possible with invalid arguments being passed anyways.

cpp/src/arrow/acero/exec_plan.cc
westonpace: Fair enough.  I only saw the string accepting variant and missed the char* version.  I've switched.

cpp/src/arrow/acero/source_node.cc
westonpace: Even better, I switched to `std::optional::value_or` :)

cpp/src/arrow/type_traits.cc
westonpace: Done.  I was scratching my head trying to figure out how to unify these.  Thanks.

cpp/src/arrow/type_traits.cc
```diff
@@ -0,0 +1,96 @@
...
+namespace arrow {
+
+int RequiredValueAlignmentForBuffer(Type::type type_id, int buffer_index) {
+  if (buffer_index != 1) {
+    // If the buffer index is 0 then either:
+    //  * The array type has no buffers, thus this shouldn't be called anyways
+    //  * The array has a validity buffer at 0, no alignment needed
+    //  * The array is a union array and has a types buffer at 0, no alignment needed
+    // If the buffer index is > 1 then, in all current cases, it represents binary
+    //  data and no alignment is needed
```
pitrou, line 31: Hmm, I took a look again and it's more complicated. Unions don't have a null bitmap, but in Arrow C++ we still reserve an empty slot in buffer 0 for the non-existent union null bitmap. Therefore, the dense union offsets buffer is number 2, not 1, and needs to be word-aligned...

cpp/src/arrow/testing/gtest_util.h
```diff
@@ -532,4 +532,9 @@ class ARROW_TESTING_EXPORT GatingTask {
   std::shared_ptr<Impl> impl_;
 };
 
+/// \brief modify an array so that the buffer at index 1 (if it has one) is unaligned
+ARROW_TESTING_EXPORT std::shared_ptr<ArrayData> UnalignValues(const ArrayData& array);
+/// \brief modify an array so that the buffer at index 1 (if it has one) is unaligned
+ARROW_TESTING_EXPORT std::shared_ptr<Array> UnalignValues(const Array& array);
```
pitrou, line 538: Why not instead "unalign" all buffers, even if no particular alignment is required? This would probably make things more interesting?

cpp/src/arrow/util/align_util_test.cc
```diff
@@ -278,4 +279,225 @@ TEST(EnsureAlignment, Table) {
...
+};
+
+template <typename T>
+class MallocAlignmentRequired : public MallocAlignment {};
+template <typename T>
+class MallocAlignmentNotRequired : public MallocAlignment {};
+
+TYPED_TEST_SUITE(MallocAlignmentRequired, TypesRequiringSomeKindOfAlignment);
+TYPED_TEST_SUITE(MallocAlignmentNotRequired, TypesNotRequiringAlignment);
+
```
pitrou, line 410: Why not also a test that aligned values do not need any realignment, or did I miss an existing test for this?

pitrou: Something like:
```c++
TYPED_TEST(MallocAlignmentRequired, AlreadyAligned) {
  std::shared_ptr<ArrayData> data = SampleArray<TypeParam>();
  ASSERT_OK_AND_ASSIGN(
      std::shared_ptr<ArrayData> aligned,
      util::EnsureAlignment(data, util::kValueAlignment, default_memory_pool()));

  ASSERT_OK(ValidateArrayFull(*aligned));
  AssertArraysEqual(*MakeArray(data), *MakeArray(aligned));
  this->CheckUnmodified(*data, *aligned);
}
```

cpp/src/arrow/testing/gtest_util.h
```diff
@@ -532,4 +532,9 @@ class ARROW_TESTING_EXPORT GatingTask {
   std::shared_ptr<Impl> impl_;
 };
 
+/// \brief modify an array so that the buffer at index 1 (if it has one) is unaligned
```
pitrou, line 535: Also mention that this doesn't recurse into children and dictionary?

cpp/src/arrow/acero/exec_plan.h
```diff
@@ -496,6 +496,16 @@ struct ARROW_ACERO_EXPORT Declaration {
   std::string label;
 };
 
+/// \brief describes how to handle unaligned buffers
```
pitrou, line 499:
```suggestion
/// \brief How to handle unaligned buffers
///
/// Please see the QueryOptions docstring for details.
```

cpp/src/arrow/acero/exec_plan.h
```diff
@@ -555,6 +565,34 @@ struct ARROW_ACERO_EXPORT QueryOptions {
   ///
   /// If set then the number of names must equal the number of output columns
   std::vector<std::string> field_names;
+
+  /// Various compute functions and acero internals will type pun array
```
pitrou, line 569:
```suggestion
  /// \brief Policy for unaligned buffers in source data
  ///
  /// Various compute functions and acero internals will type pun array
```

cpp/src/arrow/acero/exec_plan.h
```diff
@@ -555,6 +565,34 @@ struct ARROW_ACERO_EXPORT QueryOptions {
...
+  /// buffers from uint8_t* to some kind of value type (e.g. we might
+  /// cast to int32_t* to add two int32 arrays)
+  ///
+  /// If the buffer is poorly algined (e.g. an int32 array is not aligned
+  /// on a 4-byte boundary) then this is technically undefined behavior.
```
pitrou, line 574:
```suggestion
  /// If the buffer is poorly aligned (e.g. an int32 array is not aligned
  /// on a 4-byte boundary) then this is technically undefined behavior in C++.
```

cpp/src/arrow/acero/exec_plan.h
```diff
@@ -555,6 +565,34 @@ struct ARROW_ACERO_EXPORT QueryOptions {
...
+  /// However, most modern compilers and CPUs are fairly tolerant of this
+  /// behavior and nothing bad (beyond a small hit to performance) is likely
+  /// to happen.
+  ///
+  /// Note that this only applies to source buffers.  All buffers allocated interally
```
pitrou, line 579:
```suggestion
  /// Note that this only applies to source buffers.  All buffers allocated internally
```

cpp/src/arrow/acero/source_node.cc
```diff
@@ -52,6 +52,45 @@ using arrow::internal::MapVector;
...
+                "to kAbort");
+          }
+          break;
+        case UnalignedBufferHandling::kWarn:
+          if (!arrow::util::CheckAlignment(*value.array(),
+                                           arrow::util::kValueAlignment)) {
+            ARROW_LOG(WARNING)
+                << "An input buffer was poorly aligned.  This could lead to crashes or "
+                   "poor performance on some hardware.  Please ensure that all Acero "
+                   "sources generate aligned buffers.";
```
pitrou, line 79:
```suggestion
                   "sources generate aligned buffers, or change the unaligned buffer "
                   "handling configuration.";
```

pitrou (commented):
Thanks a lot for this! I posted a number of minor comments, though this is a very nice and sound improvement on the principle.

cpp/src/arrow/acero/exec_plan.h
```diff
@@ -555,6 +565,34 @@ struct ARROW_ACERO_EXPORT QueryOptions {
...
+  /// by Acero will be suitably aligned.
+  ///
+  /// If this field is set to kWarn then Acero will check if any buffers are unaligned
```
bkietz, line 582: Perhaps this section of the comment could be broken into comments for the members of `UnalignedBufferHandling`?

cpp/src/arrow/acero/plan_test.cc
```diff
@@ -1704,5 +1704,43 @@ TEST(ExecPlanExecution, SegmentedAggregationWithBatchCrossingSegment) {
...
+  // By default we should warn and so the plan should finish ok
+  ASSERT_OK(DeclarationToStatus(plan));
+  ASSERT_EQ(initial_bytes_allocated, default_memory_pool()->total_bytes_allocated());
+
+  QueryOptions query_options;
+
+  // Nothing should happen if we ignore alignment
+  query_options.unaligned_buffer_handling = UnalignedBufferHandling::kIgnore;
+  ASSERT_OK(DeclarationToStatus(plan, query_options));
+  ASSERT_EQ(initial_bytes_allocated, default_memory_pool()->total_bytes_allocated());
```
bkietz, line 1732:
```suggestion
#ifndef ARROW_UBSAN
  query_options.unaligned_buffer_handling = UnalignedBufferHandling::kIgnore;
  ASSERT_OK(DeclarationToStatus(plan, query_options));
  ASSERT_EQ(initial_bytes_allocated, default_memory_pool()->total_bytes_allocated());
#endif
```

cpp/src/arrow/util/align_util_test.cc
```diff
@@ -278,4 +279,225 @@ TEST(EnsureAlignment, Table) {
...
+  auto ty = sparse_union({field("ints", int64()), field("strs", utf8())}, {2, 7});
+  auto ints = ArrayFromJSON(int64(), "[0, 1, 2, 3]");
+  auto strs = ArrayFromJSON(utf8(), R"(["a", null, "c", "d"])");
+  auto ids = ArrayFromJSON(int8(), "[2, 7, 2, 7]")->data()->buffers[1];
+  const int length = 4;
+  SparseUnionArray arr(ty, length, {ints, strs}, ids);
+  return arr.data();
+}
+
+class MallocAlignment : public ::testing::Test {
```
bkietz, line 377:
```suggestion
class ValueAlignment : public ::testing::Test {
```

bkietz (approved):
LGTM, just a few nits

cpp/src/arrow/util/align_util.cc
```diff
@@ -94,12 +135,17 @@ bool CheckAlignment(const Table& table, int64_t alignment,
   return all_aligned;
 }
 
+// Most allocators require a minimum of 8-byte alignment.
+constexpr int64_t kMinimumAlignment = 8;
```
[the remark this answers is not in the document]
westonpace: The problem arises when we are checking for value alignment.  For example, if we encounter a buffer that requires 4 byte alignment then we will call:

```
EnsureAlignment(buffer, 4, memory_pool);
```

If that buffer was not 4-byte aligned then we would try and allocate a new buffer:

```
    ARROW_ASSIGN_OR_RAISE(
        auto new_buffer,
        AllocateBuffer(buffer->size(), 4, memory_pool));
```

This would then raise.

Another option would be to change the signature of `EnsureAlignment`:

```
Result<std::shared_ptr<Buffer>> EnsureAlignment(std::shared_ptr<Buffer> buffer,
                                                int64_t expected_alignment,
                                                int64_t new_alignment_if_needed,
                                                MemoryPool* memory_pool)
```

Then we could pass `4` for `expected_alignment` and `kDefaultBufferAlignment` for `new_alignment_if_needed`.  I wasn't sure if it was worth cluttering the signature of `EnsureAlignment` or not.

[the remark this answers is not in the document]
pitrou: > This would then raise.

Hmm, why would it? I think I'm missing something here.

cpp/src/arrow/acero/exec_plan.h
westonpace: I don't believe comments on enum members show up in our API docs.  Though, one could argue that no one reads those :)

cpp/src/arrow/util/align_util_test.cc
```diff
@@ -278,4 +279,225 @@ TEST(EnsureAlignment, Table) {
...
+TYPED_TEST_SUITE(MallocAlignmentNotRequired, TypesNotRequiringAlignment);
+
+TYPED_TEST(MallocAlignmentRequired, RoundTrip) {
+  std::shared_ptr<ArrayData> data = SampleArray<TypeParam>();
+  std::shared_ptr<ArrayData> unaligned = UnalignValues(*data);
+  ASSERT_OK_AND_ASSIGN(
+      std::shared_ptr<ArrayData> aligned,
+      util::EnsureAlignment(unaligned, util::kValueAlignment, default_memory_pool()));
+
+  AssertArraysEqual(*MakeArray(data), *MakeArray(aligned));
```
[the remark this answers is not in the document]
westonpace: Added.

cpp/src/arrow/util/align_util_test.cc
```diff
@@ -278,4 +279,225 @@ TEST(EnsureAlignment, Table) {
...
+}
+
+TYPED_TEST(MallocAlignmentNotRequired, RoundTrip) {
+  std::shared_ptr<ArrayData> data = SampleArray<TypeParam>();
+  std::shared_ptr<ArrayData> unaligned = UnalignValues(*data);
+  ASSERT_OK_AND_ASSIGN(
+      std::shared_ptr<ArrayData> aligned,
+      util::EnsureAlignment(unaligned, util::kValueAlignment, default_memory_pool()));
+
+  AssertArraysEqual(*MakeArray(data), *MakeArray(aligned));
```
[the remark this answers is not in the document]
westonpace: Added a check

cpp/src/arrow/util/align_util_test.cc
```diff
@@ -278,4 +279,225 @@ TEST(EnsureAlignment, Table) {
...
+  ASSERT_OK_AND_ASSIGN(
+      aligned_ree,
+      util::EnsureAlignment(aligned_ree, util::kValueAlignment, default_memory_pool()));
+
+  this->CheckModified(*unaligned_ree->child_data[0], *aligned_ree->child_data[0]);
+  this->CheckUnmodified(*unaligned_ree->child_data[1], *aligned_ree->child_data[1]);
+}
+
+TEST_F(MallocAlignment, Dictionary) {
+  // Dictionary values require alignment, dictionary keys do not
```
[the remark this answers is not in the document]
westonpace: Fixed

cpp/src/arrow/util/align_util_test.cc
```diff
@@ -278,4 +279,225 @@ TEST(EnsureAlignment, Table) {
...
+
+  std::shared_ptr<ArrayData> unaligned_dict = std::make_shared<ArrayData>(*array->data());
+  unaligned_dict->dictionary = UnalignValues(*unaligned_dict->dictionary);
+  unaligned_dict = UnalignValues(*unaligned_dict);
+
+  std::shared_ptr<ArrayData> aligned_dict = std::make_shared<ArrayData>(*unaligned_dict);
+
+  ASSERT_OK_AND_ASSIGN(
+      aligned_dict,
+      util::EnsureAlignment(aligned_dict, util::kValueAlignment, default_memory_pool()));
```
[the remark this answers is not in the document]
westonpace: Added a check

cpp/src/arrow/util/align_util_test.cc
```diff
@@ -278,4 +279,225 @@ TEST(EnsureAlignment, Table) {
...
+}
+
+TEST_F(MallocAlignment, Extension) {
+  std::shared_ptr<Array> array = ExampleSmallint();
+
+  std::shared_ptr<ArrayData> unaligned = UnalignValues(*array->data());
+
+  ASSERT_OK_AND_ASSIGN(
+      std::shared_ptr<ArrayData> aligned,
+      util::EnsureAlignment(unaligned, util::kValueAlignment, default_memory_pool()));
```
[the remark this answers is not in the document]
westonpace: Added a check

cpp/src/arrow/util/align_util.h
```diff
@@ -82,29 +101,109 @@ ARROW_EXPORT bool CheckAlignment(const Array& array, int64_t alignment);
...
 // completely aligned.
+
+/// \brief calculate which (if any) chunks in a chunked array are unaligned
+/// \param array the array to check
+/// \param alignment the alignment to check for
+/// \param needs_alignment an output vector that will store the results of the check
+///        it must be set to a valid vector.  Extra elements will be added to the end
+///        of the vector for each chunk that is checked.  `true` will be stored if
+///        the chunk is unaligned.
+/// \param offset an optional offset to specify which chunk to start checking at
```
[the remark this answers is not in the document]
westonpace: I changed the wording to:

```
/// \param offset the index of the chunk to start checking
```

cpp/src/arrow/util/align_util.h
```diff
@@ -70,8 +70,27 @@ inline BitmapWordAlignParams BitmapWordAlign(const uint8_t* data, int64_t bit_of
 namespace util {
 
 // Functions to check if the provided Arrow object is aligned by the specified alignment
+
+/// \brief if this is specified in one of the CheckAlignment or EnsureAlignment functions
+///
+/// then the function will ensure each buffer is suitably aligned for the data type of the
```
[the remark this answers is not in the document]
westonpace: Done

cpp/src/arrow/util/align_util.h
```diff
@@ -70,8 +70,27 @@ inline BitmapWordAlignParams BitmapWordAlign(const uint8_t* data, int64_t bit_of
...
+/// array.  For example, given an int32 buffer the validity buffer must be a multiple of 8
+/// and the values buffer must be a multiple of 32.  Given a large_string buffer the
+/// validity buffer and values buffers must be multiples of 8 and the offsets buffer must
+/// be a multiple of 64.
+constexpr int64_t kValueAlignment = -3;
+
+/// \brief calculate if the buffer's address is a multiple of `alignment`
+/// \param buffer the buffer to check
+/// \param alignment the alignment to check for
```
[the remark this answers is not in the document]
westonpace: I changed all of these to:

```
/// \param alignment the alignment (in bytes) to check for
```

cpp/src/arrow/util/align_util.h
```diff
@@ -70,8 +70,27 @@ inline BitmapWordAlignParams BitmapWordAlign(const uint8_t* data, int64_t bit_of
...
 
 // Functions to check if the provided Arrow object is aligned by the specified alignment
+
+/// \brief if this is specified in one of the CheckAlignment or EnsureAlignment functions
+///
+/// then the function will ensure each buffer is suitably aligned for the data type of the
+/// array.  For example, given an int32 buffer the validity buffer must be a multiple of 8
+/// and the values buffer must be a multiple of 32.  Given a large_string buffer the
+/// validity buffer and values buffers must be multiples of 8 and the offsets buffer must
+/// be a multiple of 64.
```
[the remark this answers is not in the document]
westonpace: Yes, my mistake.  I've cleaned up this comment:

```
/// array.  For example, given an int32 buffer the values buffer's address must be a
/// multiple of 4.  Given a large_string buffer the offsets buffer's address must be a
/// multiple of 8.
```

cpp/src/arrow/util/align_util.h
```diff
@@ -70,8 +70,27 @@ inline BitmapWordAlignParams BitmapWordAlign(const uint8_t* data, int64_t bit_of
...
+constexpr int64_t kValueAlignment = -3;
+
+/// \brief calculate if the buffer's address is a multiple of `alignment`
+/// \param buffer the buffer to check
+/// \param alignment the alignment to check for
 ARROW_EXPORT bool CheckAlignment(const Buffer& buffer, int64_t alignment);
+/// \brief calculate if all buffer's in the array data are aligned
```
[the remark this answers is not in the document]
westonpace: I've clarified that it does include these.

cpp/src/arrow/util/align_util.h
```diff
@@ -70,8 +70,27 @@ inline BitmapWordAlignParams BitmapWordAlign(const uint8_t* data, int64_t bit_of
...
+/// \param array the array data to check
+/// \param alignment the alignment to check for
 ARROW_EXPORT bool CheckAlignment(const ArrayData& array, int64_t alignment);
+/// \brief calculate if all buffer's in the array are aligned
```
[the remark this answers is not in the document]
westonpace: Done

cpp/src/arrow/util/align_util.h
```diff
@@ -82,29 +101,109 @@ ARROW_EXPORT bool CheckAlignment(const Array& array, int64_t alignment);
 // of the constituent objects during the EnsureAlignment function where certain
 // objects can be ignored for further checking if we already know that they are
 // completely aligned.
+
+/// \brief calculate which (if any) chunks in a chunked array are unaligned
```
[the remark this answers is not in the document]
westonpace: Done

cpp/src/arrow/util/align_util.cc
```diff
@@ -94,12 +135,17 @@ bool CheckAlignment(const Table& table, int64_t alignment,
...
+
 Result<std::shared_ptr<Buffer>> EnsureAlignment(std::shared_ptr<Buffer> buffer,
                                                 int64_t alignment,
                                                 MemoryPool* memory_pool) {
```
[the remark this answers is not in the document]
westonpace: I added a check and still want to add a test.

cpp/src/arrow/util/align_util.cc
```diff
@@ -30,12 +33,50 @@ bool CheckAlignment(const Buffer& buffer, int64_t alignment) {
...
+}
+
+}  // namespace
+
+bool CheckAlignment(const ArrayData& array, int64_t alignment) {
+  if (!CheckSelfAlignment(array, alignment)) {
+    return false;
+  }
 
   if (array.type->id() == Type::DICTIONARY) {
```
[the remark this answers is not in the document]
westonpace: I changed this to `if (array.dictionary)`

cpp/src/arrow/util/align_util.cc
```diff
@@ -112,10 +158,17 @@ Result<std::shared_ptr<ArrayData>> EnsureAlignment(std::shared_ptr<ArrayData> ar
                                                    MemoryPool* memory_pool) {
   if (!CheckAlignment(*array_data, alignment)) {
     std::vector<std::shared_ptr<Buffer>> buffers_ = array_data->buffers;
```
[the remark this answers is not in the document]
westonpace: Good catch.  I've fixed this.

cpp/src/arrow/type_traits.h
```diff
@@ -1309,6 +1309,28 @@ static inline int offset_bit_width(Type::type type_id) {
...
+/// by flight (GH-32276).
+///
+/// We call this "value aligned" and this method will calculate that required alignment.
+///
+/// \param type_id the type of the array containing the buffer
+///                Note: this should be the indices type for a dictionary array since
+///                A dictionary array's buffers are indices.  It should be the storage
+///                type for an extension array.
+/// \param buffer_index the index of the buffer to check, for example 0 will typically
+///                     give you the alignment expected of the validity buffer
```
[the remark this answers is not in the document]
westonpace: Fixed

cpp/src/arrow/type_traits.cc
westonpace: I was not aware of that.  I'll add a test case for this.

cpp/src/arrow/testing/gtest_util.h
westonpace: I clarified the comment

cpp/src/arrow/acero/exec_plan.h
westonpace: Fixed.

cpp/src/arrow/acero/exec_plan.h
westonpace: Fixed.

cpp/src/arrow/acero/exec_plan.h
westonpace: Fixed.

cpp/src/arrow/acero/exec_plan.h
westonpace: Fixed.

cpp/src/arrow/acero/exec_plan.h
```diff
@@ -496,6 +496,16 @@ struct ARROW_ACERO_EXPORT Declaration {
   std::string label;
 };
 
+/// \brief describes how to handle unaligned buffers
+enum class UnalignedBufferHandling { kWarn, kIgnore, kReallocate, kAbort };
```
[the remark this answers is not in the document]
westonpace: I've changed this to `kError`.

cpp/src/arrow/acero/source_node.cc
westonpace: Fixed.

cpp/src/arrow/util/align_util.cc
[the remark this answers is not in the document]
pitrou: Hmm, we should fix our wrapper around `posix_memalign` then, IMHO.
I think we should also use `max(kDefaultBufferAlignment, alignment)` in this PR.

cpp/src/arrow/acero/plan_test.cc
westonpace: I use `ARROW_UBSAN` regularly and so I was curious why this wasn't a problem.  It turns out we never pass `-fsanitize=alignment` to ubsan.  Perhaps we should add this flag?

cpp/src/arrow/util/align_util_test.cc
westonpace: Good catch.  Fixed.

cpp/src/arrow/acero/plan_test.cc
pitrou: Perhaps open a separate issue for that, but it may open a can of worms :-)

cpp/src/arrow/util/align_util.cc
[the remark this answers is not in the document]
westonpace: >  I think we should also use max(kDefaultBufferAlignment, alignment) in this PR.

I think that is reasonable.  I'll update the comments to explain this.

cpp/src/arrow/type_traits.cc
westonpace: I've fixed dense unions and test these independently now.

cpp/src/arrow/testing/gtest_util.h
westonpace: I now unalign all buffers.  It was simpler anyways.

cpp/src/arrow/util/align_util_test.cc
westonpace: I added this test for all types.  It verifies that the array generated by `SampleArray` (i.e. an array generated with kDefaultBufferAlignment`) is suitably aligned for `kValueAlignment` and that `EnsureAlignment` does not modify it.

cpp/src/arrow/util/align_util.cc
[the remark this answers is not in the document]
westonpace: I've done this, and gotten rid of `kMinimumAlignment`.

> Hmm, we should fix our wrapper around posix_memalign then, IMHO.

We can do this in a follow-up PR.  I've created https://github.com/apache/arrow/issues/35793

cpp/src/arrow/util/align_util.cc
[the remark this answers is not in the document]
westonpace: I've added a general check (and test) for zero/negative alignment and a specific check for `kValueAlignment` mentioning that it is not an valid option for buffers.

cpp/src/arrow/util/align_util.h
```diff
@@ -70,8 +70,34 @@ inline BitmapWordAlignParams BitmapWordAlign(const uint8_t* data, int64_t bit_of
 namespace util {
 
 // Functions to check if the provided Arrow object is aligned by the specified alignment
+
+/// \brief Special alignment value to use data type-specific alignment
+///
+/// If this is passed as the `alignment` in one of the CheckAlignment or EnsureAlignment
+/// functions, then the function will ensure ensure each buffer is suitably aligned
+/// for the data type of the
```
mapleFU, line 78: The line break is weird here

mapleFU (approved):
Rest LGTM!

cpp/src/arrow/util/align_util.h
westonpace: Good catch.  I've cleaned up the comment.

raulcd:
@github-actions crossbow submit test-build-vcpkg-win

PR #34227, opened 2023-02-16
GH-34223: [Java] Java Substrait Consumer JNI call to ACERO C++

davisusanibar:
* Closes: #34223

The purpose of this PR is to implement:

1. JNI Wrappers to consume Acero capabilities that execute Substrait Plans. 
2. Java base code to offer API that consume Substrait Plans. 
3. Initial Substrait documentation

PR #35570, opened 2023-05-12
GH-34252: [Java] Support ScannerBuilder::Project or ScannerBuilder::Filter as a Substrait proto extended expression

### LINKED ISSUE #34252: [Java] Filter and project Datasets with Substrait expressions

The features requested in https://github.com/apache/arrow/issues/33985 will allow Substrait expressions to be passed to the Dataset Filter and Project methods. After that is implemented, we should expose it to Java through the JNI Dataset bindings.

### Component(s)

Java

davisusanibar:
### Rationale for this change

To close https://github.com/apache/arrow/issues/34252

This is a proposal to try to solve:
1. Receive a list of Substrait scalar expressions and use them to Project a Dataset
- [x] Draft a Substrait Extended Expression to test (this will be generated by 3rd party project such as Isthmus)
- [x] Use C++ draft PR to Serialize/Deserialize Extended Expression proto messages
- [x] Create JNI Wrapper for ScannerBuilder::Project 
- [x] Create JNI API
- [x] Testing coverage
- [x] Documentation

Current problem is: `java.lang.RuntimeException: Inferring column projection from FieldRef FieldRef.FieldPath(0)`. Not able to infer by column position by able to infer by colum name. This problem is solved by https://github.com/apache/arrow/pull/35798

This PR needs/use this PRs/Issues:
- https://github.com/apache/arrow/pull/34834
- https://github.com/apache/arrow/pull/34227
- https://github.com/apache/arrow/issues/35579

2. Receive a Boolean-valued Substrait scalar expression and use it to filter a Dataset
- [x] Working to identify activities

### Are these changes tested?

Initial unit test added.

No
* Closes: #34252


davisusanibar:
HI @lidavidm, this PR enable Dataset Projections and Filters using Substrait Extended Expressions. This depends of another 02 PRs yet:
- https://github.com/apache/arrow/pull/35798
- https://github.com/apache/arrow/pull/34834

Could be possible when you have some time to give some feedback for Java side implementation?

Thank you in advance.

lidavidm:
@davisusanibar is it possible to rebase so it's clear which commits are for this PR and which are from the other PRs?

davisusanibar:
> @davisusanibar is it possible to rebase so it's clear which commits are for this PR and which are from the other PRs?

Just merge with main branch and another 02 PRs dependents

davisusanibar:
> What I meant is if you could stage this as a rebase on top of those PRs, so that there's a separation. Otherwise I can review as is but I will assume no C++ or Python changes are from this PR specifically

Please consider that any change as part of cpp/*, python/* won't be part of this PR. Only consider: java/* and docs/*

docs/source/java/dataset.rst
```diff
@@ -158,6 +156,21 @@ Or use shortcut construtor:
...
 Then all columns will be emitted during scanning.
 
+Projection (Produce New Columns) and Filters
+============================================
+
+User can specify projections (new columns) or filters in ScanOptions. For example:
+
+.. code-block:: Java
+
+   ByteBuffer substraitExtendedExpressions = ...; // createExtendedExpresionMessageUsingSubstraitPOJOClasses
```
lidavidm, line 166: remove the comment? Or write it out as a sentence on a new line (`// Use Substrait APIs to create an Expression and serialize to a ByteBuffer`)

docs/source/java/substrait.rst
```diff
@@ -102,6 +104,335 @@ Here is an example of a Java program that queries a Parquet file using Java Subs
     0	ALGERIA	0	 haggle. carefully final deposits detect slyly agai
     1	ARGENTINA	1	al foxes promise slyly according to the regular accounts. bold requests alon
 
+Executing Projections and Filters Using Extended Expressions
+============================================================
+
+Using `Extended Expression`_ we could leverage our current Dataset operations to
+also support Projections and Filters by. To gain access to Projections and Filters
+is needed to define that operations using current Extended Expression Java POJO
+classes defined into `Substrait Java`_ project.
```
lidavidm, line 113:
```suggestion
Dataset also supports projections and filters with Substrait's extended expressions.
This requires the substrait-java library.
```

docs/source/java/substrait.rst
```diff
@@ -102,6 +104,335 @@ Here is an example of a Java program that queries a Parquet file using Java Subs
...
+
+Here is an example of a Java program that queries a Parquet file to project new
+columns and also filter then based on Extended Expression definitions. This example
+show us:
+
+- Load TPCH parquet file Nation.parquet.
+- Produce new Projections and apply Filter into dataset using extended expression definition.
+    - Expression 01 - CONCAT: N_NAME || ' - ' || N_COMMENT = col 1 || ' - ' || col 3.
+    - Expression 02 - ADD: N_REGIONKEY + 10 = col 1 + 10.
+    - Expression 03 - FILTER: N_NATIONKEY > 18 = col 3 > 18.
```
lidavidm, line 123:
```suggestion
This Java program:

- Loads a Parquet file containing the "nation" table from the TPC-H benchmark.
- Projects two new columns:
    - ``N_NAME || ' - ' || N_COMMENT``
    - ``N_REGIONKEY + 10``
- Applies a filter: ``N_NATIONKEY > 18``
```

docs/source/java/substrait.rst
```diff
@@ -102,6 +104,335 @@ Here is an example of a Java program that queries a Parquet file using Java Subs
...
+    import io.substrait.proto.SimpleExtensionURI;
+    import io.substrait.type.NamedStruct;
+    import io.substrait.type.Type;
+    import io.substrait.type.TypeCreator;
+    import io.substrait.type.proto.TypeProtoConverter;
+
+    public class ClientSubstraitExtendedExpressions {
+      public static void main(String[] args) throws Exception {
+        // create extended expression for: project two new columns + one filter
+        String binaryExtendedExpressions = createExtendedExpresionMessageUsingPOJOClasses();
```
lidavidm, line 163: Um, why use a String? Just pass it around as a ByteBuffer in the first place.

docs/source/java/substrait.rst
```diff
@@ -102,6 +104,335 @@ Here is an example of a Java program that queries a Parquet file using Java Subs
...
+        // project and filter dataset using extended expression definition - 03 Expressions:
+        // Expression 01 - CONCAT: N_NAME || ' - ' || N_COMMENT = col 1 || ' - ' || col 3
+        // Expression 02 - ADD: N_REGIONKEY + 10 = col 1 + 10
+        // Expression 03 - FILTER: N_NATIONKEY > 18 = col 3 > 18
```
lidavidm, line 167:
```suggestion
```

docs/source/java/substrait.rst
```diff
@@ -102,6 +104,335 @@ Here is an example of a Java program that queries a Parquet file using Java Subs
...
+    import io.substrait.proto.SimpleExtensionDeclaration;
+    import io.substrait.proto.SimpleExtensionURI;
+    import io.substrait.type.NamedStruct;
+    import io.substrait.type.Type;
+    import io.substrait.type.TypeCreator;
+    import io.substrait.type.proto.TypeProtoConverter;
+
+    public class ClientSubstraitExtendedExpressions {
+      public static void main(String[] args) throws Exception {
+        // create extended expression for: project two new columns + one filter
```
lidavidm, line 162:
```suggestion
```

docs/source/java/substrait.rst
```diff
@@ -102,6 +104,335 @@ Here is an example of a Java program that queries a Parquet file using Java Subs
...
+
+        ExtendedExpression extendedExpression = extendedExpressionBuilder.build();
+
+        // Print JSON
+        System.out.println(
+            JsonFormat.printer().includingDefaultValueFields().print(
+                extendedExpression));
+        // Print binary representation
+        System.out.println(Base64.getEncoder().encodeToString(
+            extendedExpression.toByteArray()));
```
lidavidm, line 418: ...there's not really much point printing this

docs/source/java/substrait.rst
```diff
@@ -102,6 +104,335 @@ Here is an example of a Java program that queries a Parquet file using Java Subs
...
+                    expressionReferenceBuilderProjectOne).
+                addReferredExpr(1,
+                    expressionReferenceBuilderProjectTwo).
+                addReferredExpr(2,
+                    expressionReferenceBuilderFilterOne).
+                setBaseSchema(of.toProto());
+        extendedExpressionBuilder.addAllExtensionUris(extensionUris.values());
+        extendedExpressionBuilder.addAllExtensions(extensions);
+
+        ExtendedExpression extendedExpression = extendedExpressionBuilder.build();
```
lidavidm, line 410: How stable is this? If the format is likely to change then I wonder if it's worth having so much code that might get stale quickly

java/dataset/src/test/java/org/apache/arrow/dataset/TestDataset.java
```diff
@@ -79,6 +79,7 @@ protected List<ArrowRecordBatch> collectTaskData(Scanner scan) {
       List<ArrowRecordBatch> batches = new ArrayList<>();
       while (reader.loadNextBatch()) {
         VectorSchemaRoot root = reader.getVectorSchemaRoot();
+        System.out.println(root.getSchema());
```
lidavidm, line 82:
```suggestion
```

java/dataset/src/main/java/org/apache/arrow/dataset/substrait/JniWrapper.java
```diff
@@ -69,5 +69,8 @@ public native void executeSerializedPlan(String planInput, String[] mapTableToMe
    * @param memoryAddressOutput the memory address where RecordBatchReader is exported.
    */
   public native void executeSerializedPlan(ByteBuffer planInput, String[] mapTableToMemoryAddressInput,
-                                                      long memoryAddressOutput);
+                                           long memoryAddressOutput);
+
+  // add description
```
lidavidm, line 74: This needs to be done?

java/dataset/src/main/java/org/apache/arrow/dataset/substrait/AceroSubstraitConsumer.java
```diff
@@ -90,6 +91,15 @@ public ArrowReader runQuery(ByteBuffer plan, Map<String, ArrowReader> namedTable
     return execute(plan, namedTables);
   }
 
+  public List<String> runDeserializeExpressions(ByteBuffer plan) {
```
lidavidm, line 94: Docstrings? I don't think we want these at all though?

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
```diff
@@ -56,10 +58,26 @@ public ScanOptions(long batchSize, Optional<String[]> columns) {
...
+  }
+
+  /**
+   * Constructor.
+   * @param batchSize Maximum row number of each returned {@link org.apache.arrow.vector.ipc.message.ArrowRecordBatch}
+   * @param columns (Optional) Projected columns. {@link Optional#empty()} for scanning all columns. Otherwise,
+   *                Only columns present in the Array will be scanned.
+   * @param projectExpression (Optional) Expressions to evaluate to produce columns
+   */
+  public ScanOptions(long batchSize, Optional<String[]> columns, Optional<ByteBuffer> projectExpression) {
```
lidavidm, line 71: We should just use a builder at this point...

java/dataset/src/main/java/org/apache/arrow/dataset/jni/NativeDataset.java
```diff
@@ -36,12 +36,14 @@ public NativeDataset(NativeContext context, long datasetId) {
   }
 
   @Override
+  @SuppressWarnings("ArrayToString")
```
lidavidm, line 39: What are we suppressing here?

java/dataset/src/main/java/org/apache/arrow/dataset/jni/JniWrapper.java
```diff
@@ -67,14 +69,16 @@ private JniWrapper() {
   /**
    * Create Scanner from a Dataset and get the native pointer of the Dataset.
    * @param datasetId the native pointer of the arrow::dataset::Dataset instance.
-   * @param columns desired column names.
-   *                Columns not in this list will not be emitted when performing scan operation. Null equals
-   *                to "all columns".
+   * @param columnsSubset desired column names. Columns not in this list will not be emitted when performing scan
+   *                      operation. Null equals to "all columns".
+   * @param columnsToProduce Set expressions which will be evaluated to produce the materialized columns. Null equals
+   *                         to "no produce".
```
lidavidm, line 75:
```suggestion
   * @param columnsToProduce Expressions to materialize new columns (if desired).
```

java/dataset/src/main/cpp/jni_wrapper.cc
```diff
@@ -696,3 +722,44 @@ JNIEXPORT void JNICALL
...
+JNIEXPORT jobjectArray JNICALL
+    Java_org_apache_arrow_dataset_substrait_JniWrapper_executeDeserializeExpressions (
+    JNIEnv* env, jobject, jobject expression) {
+  JNI_METHOD_START
+  auto *buff = reinterpret_cast<jbyte*>(env->GetDirectBufferAddress(expression));
+  int length = env->GetDirectBufferCapacity(expression);
+  std::shared_ptr<arrow::Buffer> buffer = JniGetOrThrow(arrow::AllocateBuffer(length));
+  std::memcpy(buffer->mutable_data(), buff, length);
+  // execute expression
+      arrow::engine::BoundExpressions round_tripped =
```
lidavidm, line 740: C++ code needs to be formatted

java/dataset/src/main/cpp/jni_wrapper.cc
```diff
@@ -696,3 +722,44 @@ JNIEXPORT void JNICALL
...
+      )
+    );
+    env->SetObjectArrayElement(
+      extendedExpressionOutput,
+      j++,
+      env->NewStringUTF(
+        round_tripped.named_expressions[i].expression.ToString().c_str()
+      )
+    );
+  }
```
lidavidm, line 762:
```suggestion
  jobjectArray extendedExpressionOutput = (jobjectArray)env->NewObjectArray(totalExpression*2,env->FindClass("java/lang/String"),0);
  int j = 0;
  for (const auto& expression : round_tripped.named_expressions) {
    env->SetObjectArrayElement(
      extendedExpressionOutput,
      j++,
      env->NewStringUTF(expression.name.c_str())
    );
    env->SetObjectArrayElement(
      extendedExpressionOutput,
      j++,
      env->NewStringUTF(expression.expression.ToString().c_str())
    );
  }
```

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
```diff
@@ -204,4 +206,132 @@ public void testRunBinaryQueryNamedTableNation() throws Exception {
...
+    List<String> extededExpressionList =
+        new AceroSubstraitConsumer(rootAllocator()).runDeserializeExpressions(substraitExpression);
+    assertEquals(3, extededExpressionList.size() / 2);
+    assertEquals("add_two_to_column_a", extededExpressionList.get(0));
+    assertEquals("add(FieldPath(0), 2)", extededExpressionList.get(1));
+    assertEquals("concat_column_a_and_b", extededExpressionList.get(2));
+    assertEquals("binary_join_element_wise(FieldPath(1), FieldPath(1), \"\")", extededExpressionList.get(3));
+    assertEquals("filter_id_lower_than_20", extededExpressionList.get(4));
+    assertEquals("(FieldPath(0) < 20)", extededExpressionList.get(5));
+  }
```
lidavidm, line 237: I don't think this test is useful. We're just testing the C++ code without any purpose.

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
```diff
@@ -204,4 +206,132 @@ public void testRunBinaryQueryNamedTableNation() throws Exception {
...
+    String binaryExtendedExpressions = "Ch4IARIaL2Z1bmN0aW9uc19hcml0aG1ldGljLnlhbWwKHggCEhovZnVuY3Rpb25zX2NvbXBhcmlz" +
+        "b24ueWFtbBIRGg8IARoLYWRkOmkzMl9pMzISFBoSCAIQARoMY29uY2F0OnZjaGFyEhIaEAgCEAIaCmx0OmFueV9hbnkaMQoaGhgaBCoCEAE" +
+        "iCBoGEgQKAhIAIgYaBAoCKAIaE2FkZF90d29fdG9fY29sdW1uX2EaOwoiGiAIARoEYgIQASIKGggSBgoEEgIIASIKGggSBgoEEgIIARoVY2" +
+        "9uY2F0X2NvbHVtbl9hX2FuZF9iGjcKHBoaCAIaBAoCEAEiCBoGEgQKAhIAIgYaBAoCKBQaF2ZpbHRlcl9pZF9sb3dlcl90aGFuXzIwIhoKA" +
+        "klECgROQU1FEg4KBCoCEAEKBGICEAEYAg==";
+    Map<String, String> metadataSchema = new HashMap<>();
+    metadataSchema.put("parquet.avro.schema", "{\"type\":\"record\",\"name\":\"Users\"," +
+        "\"namespace\":\"org.apache.arrow.dataset\",\"fields\":[{\"name\":\"id\"," +
+        "\"type\":[\"int\",\"null\"]},{\"name\":\"name\",\"type\":[\"string\",\"null\"]}]}");
+    metadataSchema.put("writer.model.name", "avro");
```
lidavidm, line 257: Why do we need this?

lidavidm (changes requested):

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
davisusanibar: I just discovered this, all the Dataset response message attach this schema metadata in their response messages. This was not detected because only Fields or Data was compared but if all schema is needed to compare we nee to add this metadata to the expected messages.

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
lidavidm: To be clear: I don't think we should have this API, period.

docs/source/java/dataset.rst
davisusanibar: Added comment on a new line

docs/source/java/substrait.rst
davisusanibar: Changed

docs/source/java/substrait.rst
davisusanibar: Deleted

java/dataset/src/main/java/org/apache/arrow/dataset/substrait/JniWrapper.java
davisusanibar: Deleted

java/dataset/src/main/java/org/apache/arrow/dataset/substrait/AceroSubstraitConsumer.java
davisusanibar: Deleted

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
davisusanibar: Added builder class

java/dataset/src/main/java/org/apache/arrow/dataset/jni/NativeDataset.java
davisusanibar: Deleted

java/dataset/src/main/cpp/jni_wrapper.cc
davisusanibar: Deleted

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
davisusanibar: Deleted

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
davisusanibar: Compare by Fields

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
```diff
@@ -25,8 +26,9 @@
  * Options used during scanning.
  */
 public class ScanOptions {
-  private final Optional<String[]> columns;
+  private final Optional<String[]> columnsSubset;
   private final long batchSize;
+  private Optional<ByteBuffer> columnsProduceOrFilter;
```
zinking, line 31: I'd suggest we still expose a simple unbound string in the interface, but leaving the construction (ser/deser) process in the dataset jar itself.

I mean it's good to use subtrait as underlying implementation, but maybe we should just keep the interfaces simple.

zinking:
I guess people are okay with current projection usage and not seeking for subtrait integration for that.  
for me, I am looking for a method to pass my java filter down to the native scanner.  and at this stage only the simplest filter expressions, not the ones with function calls etc ( I guess these can be followed up separately). 

in a sense https://github.com/apache/arrow/pull/14287/files satisfies what I wanted but it is active and closed. 
I'm generally good with using subtrait in the implementation, but I'd suggest let's keep the java interface simple.

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
davisusanibar: Hey @lidavidm, what's your take on this?

lidavidm: The string parsing can and should be kept separate. The actual interface should reflect Substrait. (We could add a convenience method to set the filter via a string, which gets immediately parsed, for instance.)

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
```diff
@@ -49,24 +51,72 @@ public ScanOptions(String[] columns, long batchSize) {
...
+    private final Optional<String[]> columnsSubset;
+    private Optional<ByteBuffer> columnsProduceOrFilter = Optional.empty();
+
+    /**
+     * Constructor.
+     * @param batchSize Maximum row number of each returned {@link org.apache.arrow.vector.ipc.message.ArrowRecordBatch}
+     * @param columnsSubset (Optional) Projected columns. {@link Optional#empty()} for scanning all columns. Otherwise,
+     *                Only columns present in the Array will be scanned.
+     */
+    public Builder(long batchSize, Optional<String[]> columnsSubset) {
```
lidavidm, line 94: I'd say for a builder's constructor, there's no need for arguments; just use the builder methods.

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
```diff
@@ -49,24 +51,72 @@ public ScanOptions(String[] columns, long batchSize) {
...
+      this.columnsSubset = columnsSubset;
+    }
+
+    /**
+     * Define binary extended expression message for projects new columns or applies filter.
+     *
+     * @param columnsProduceOrFilter (Optional) Expressions to evaluate to projects new columns or applies filter.
+     * @return the ScanOptions configured.
+     */
+    public Builder columnsProduceOrFilter(Optional<ByteBuffer> columnsProduceOrFilter) {
```
lidavidm, line 106: This needs to be named something clearer. `substraitExtendedExpression`?

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
```diff
@@ -49,24 +51,72 @@ public ScanOptions(String[] columns, long batchSize) {
...
+     *                Only columns present in the Array will be scanned.
+     */
+    public Builder(long batchSize, Optional<String[]> columnsSubset) {
+      Preconditions.checkNotNull(columnsSubset);
+      this.batchSize = batchSize;
+      this.columnsSubset = columnsSubset;
+    }
+
+    /**
+     * Define binary extended expression message for projects new columns or applies filter.
```
lidavidm, line 101:
```suggestion
     * Set the Substrait extended expression.
     *
     * <p>Can be used to filter data and/or project new columns.
```

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
lidavidm: Please revise the rest of the code based on this.

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
lidavidm, line 31: Why is this not `final` like the others?

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
```diff
@@ -49,24 +51,72 @@ public ScanOptions(String[] columns, long batchSize) {
...
+    this.columnsProduceOrFilter = Optional.empty();
   }
 
   public ScanOptions(long batchSize) {
     this(batchSize, Optional.empty());
   }
 
-  public Optional<String[]> getColumns() {
-    return columns;
+  public Optional<String[]> getColumnsSubset() {
```
lidavidm, line 68: I don't think there's a compelling reason to rename everything. I do think it is time to add proper docstrings.

java/dataset/src/main/cpp/jni_wrapper.cc
```diff
@@ -470,12 +471,37 @@ JNIEXPORT jlong JNICALL Java_org_apache_arrow_dataset_jni_JniWrapper_createScann
...
       JniGetOrThrow(dataset->NewScan());
   JniAssertOkOrThrow(scanner_builder->Pool(pool));
-  if (columns != nullptr) {
-    std::vector<std::string> column_vector = ToStringVector(env, columns);
+  if (columns_subset != nullptr) {
+    std::vector<std::string> column_vector = ToStringVector(env, columns_subset);
     JniAssertOkOrThrow(scanner_builder->Project(column_vector));
   }
+  if (columns_to_produce_or_filter != nullptr) {
+    auto *buff = reinterpret_cast<jbyte*>(env->GetDirectBufferAddress(columns_to_produce_or_filter));
```
lidavidm, line 479: const?

java/dataset/src/main/cpp/jni_wrapper.cc
```diff
@@ -470,12 +471,37 @@ JNIEXPORT jlong JNICALL Java_org_apache_arrow_dataset_jni_JniWrapper_createScann
...
+    int length = env->GetDirectBufferCapacity(columns_to_produce_or_filter);
+    std::shared_ptr<arrow::Buffer> buffer = JniGetOrThrow(arrow::AllocateBuffer(length));
+    std::memcpy(buffer->mutable_data(), buff, length);
+    arrow::engine::BoundExpressions bounded_expression =
+      JniGetOrThrow(arrow::engine::DeserializeExpressions(*buffer));
+    std::vector<arrow::compute::Expression> project_exprs;
+    std::vector<std::string> project_names;
+    arrow::compute::Expression filter_expr;
+    int filter_count = 0;
+    for(arrow::engine::NamedExpression named_expression : bounded_expression.named_expressions) {
```
lidavidm, line 489: const&, or at least use & and move below

java/dataset/src/main/cpp/jni_wrapper.cc
```diff
@@ -470,12 +471,37 @@ JNIEXPORT jlong JNICALL Java_org_apache_arrow_dataset_jni_JniWrapper_createScann
...
+      if (named_expression.expression.type()->id() == arrow::Type::BOOL) {
+        if (filter_count > 0) {
+          JniThrow("The process only support one filter expression declared");
```
lidavidm, line 492:
```suggestion
          JniThrow("Only one filter expression may be provided");
```

java/dataset/src/main/cpp/jni_wrapper.cc
```diff
@@ -470,12 +471,37 @@ JNIEXPORT jlong JNICALL Java_org_apache_arrow_dataset_jni_JniWrapper_createScann
...
+        }
+        filter_expr = named_expression.expression;
+        filter_count++;
```
lidavidm, line 495: You can track this instead with `optional<Expression> filter_expr`

java/dataset/src/main/cpp/jni_wrapper.cc
```diff
@@ -470,12 +471,37 @@ JNIEXPORT jlong JNICALL Java_org_apache_arrow_dataset_jni_JniWrapper_createScann
...
+      } else {
+        project_exprs.push_back(named_expression.expression);
+        project_names.push_back(named_expression.name);
+      }
+    }
+    JniAssertOkOrThrow(scanner_builder->Project(project_exprs, project_names));
```
lidavidm, line 501: move

java/dataset/src/main/cpp/jni_wrapper.cc
```diff
@@ -470,12 +471,37 @@ JNIEXPORT jlong JNICALL Java_org_apache_arrow_dataset_jni_JniWrapper_createScann
...
-    std::vector<std::string> column_vector = ToStringVector(env, columns);
+  if (columns_subset != nullptr) {
+    std::vector<std::string> column_vector = ToStringVector(env, columns_subset);
     JniAssertOkOrThrow(scanner_builder->Project(column_vector));
   }
+  if (columns_to_produce_or_filter != nullptr) {
+    auto *buff = reinterpret_cast<jbyte*>(env->GetDirectBufferAddress(columns_to_produce_or_filter));
+    int length = env->GetDirectBufferCapacity(columns_to_produce_or_filter);
+    std::shared_ptr<arrow::Buffer> buffer = JniGetOrThrow(arrow::AllocateBuffer(length));
+    std::memcpy(buffer->mutable_data(), buff, length);
```
lidavidm, line 482: I think we do this a few times by now, factor out a helper

java/dataset/src/main/cpp/jni_wrapper.cc
```diff
@@ -470,12 +471,37 @@ JNIEXPORT jlong JNICALL Java_org_apache_arrow_dataset_jni_JniWrapper_createScann
...
+    arrow::engine::BoundExpressions bounded_expression =
+      JniGetOrThrow(arrow::engine::DeserializeExpressions(*buffer));
```
lidavidm, line 484: That said, you don't necessarily need to copy to a new buffer here if you want to avoid that; you can directly wrap a pointer + length in a buffer (so long as the buffer does not escape)

java/dataset/src/main/cpp/jni_wrapper.cc
lidavidm: That said, we should design the API to separate project and filter expressions in the first place.

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
lidavidm, line 106: Don't take Optional here.

docs/source/java/substrait.rst
```diff
@@ -102,6 +104,323 @@ Here is an example of a Java program that queries a Parquet file using Java Subs
     0	ALGERIA	0	 haggle. carefully final deposits detect slyly agai
     1	ARGENTINA	1	al foxes promise slyly according to the regular accounts. bold requests alon
 
+Executing Projections and Filters Using Extended Expressions
+============================================================
+
+Dataset also supports projections and filters with Substrait's extended expressions.
```
lidavidm, line 110: Link to what this means

docs/source/java/substrait.rst
```diff
@@ -102,6 +104,323 @@ Here is an example of a Java program that queries a Parquet file using Java Subs
...
+    import java.util.Arrays;
+    import java.util.Base64;
+    import java.util.HashMap;
+    import java.util.List;
+    import java.util.Optional;
+
+    public class ClientSubstraitExtendedExpressions {
+        public static void main(String[] args) throws Exception {
+            // create extended expression for: project two new columns + one filter
+            ByteBuffer binaryExtendedExpressions = createExtendedExpresionMessageUsingPOJOClasses();
```
lidavidm, line 157: There is no reason to be so verbose

lidavidm: Long method names hurt clarity in the docs where the width is limited

docs/source/java/substrait.rst
```diff
@@ -102,6 +104,323 @@ Here is an example of a Java program that queries a Parquet file using Java Subs
...
+            // project and filter dataset using extended expression definition - 03 Expressions:
+            // Expression 01 - CONCAT: N_NAME || ' - ' || N_COMMENT = col 1 || ' - ' || col 3
+            // Expression 02 - ADD: N_REGIONKEY + 10 = col 1 + 10
+            // Expression 03 - FILTER: N_NATIONKEY > 18 = col 3 > 18
+            projectAndFilterDataset(binaryExtendedExpressions);
+        }
+
+        public static void projectAndFilterDataset(ByteBuffer binaryExtendedExpressions) {
+            String uri = "file:////Users/dsusanibar/voltron/fork/consumer-testing/tests/data/tpch_parquet/nation.parquet";
```
lidavidm, line 166: Try not to put our company name in strings?

lidavidm (changes requested):

docs/source/java/substrait.rst
davisusanibar: changed

docs/source/java/substrait.rst
davisusanibar: changed

java/dataset/src/main/cpp/jni_wrapper.cc
davisusanibar: changed

java/dataset/src/main/cpp/jni_wrapper.cc
davisusanibar: added

java/dataset/src/main/cpp/jni_wrapper.cc
davisusanibar: Can you give me an example, I didn't catch the idea.

java/dataset/src/main/cpp/jni_wrapper.cc
davisusanibar: added

java/dataset/src/main/cpp/jni_wrapper.cc
davisusanibar: added

java/dataset/src/main/cpp/jni_wrapper.cc
davisusanibar: added

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
davisusanibar: deleted

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
davisusanibar: rollback

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
davisusanibar: deleted

java/dataset/src/main/java/org/apache/arrow/dataset/source/Dataset.java
```diff
@@ -32,4 +32,14 @@ public interface Dataset extends AutoCloseable {
...
+
+  /**
+   * Create a new Scanner, using the provided options,
+   * that contains the binary representation of the Substrait
+   * Extended Expression.
+   *
+   * @param options options used during creating Scanner
+   * @return the Scanner instance
+   */
+  Scanner newSubstraitScan(ScanOptions options);
```
lidavidm, line 44: The point of having a ScanOptions object is that we shouldn't need a second function like this, right? newScan should just do the right thing based on the options.

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
```diff
@@ -69,4 +83,8 @@ public Optional<String[]> getColumns() {
   public long getBatchSize() {
     return batchSize;
   }
+
+  public ByteBuffer getSubstraitExtendedExpression() {
```
lidavidm, line 87: Either both getters should use Optional or neither should.

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
```diff
@@ -58,6 +60,18 @@ public ScanOptions(long batchSize, Optional<String[]> columns) {
     this.columns = columns;
   }
 
+  /**
+   * Constructor.
+   * @param batchSize Maximum row number of each returned {@link org.apache.arrow.vector.ipc.message.ArrowRecordBatch}
+   * @param substraitExtendedExpression Extended expression to evaluate for project new columns or apply filter.
+   */
+  public ScanOptions(long batchSize, ByteBuffer substraitExtendedExpression) {
```
lidavidm, line 68: We can and still should have a builder, I was just saying that it doesn't make sense to pass Optional as a constructor parameter here...

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
lidavidm, line 87: Why don't we separate out projection and filter? C++ does this.

java/dataset/src/main/cpp/jni_wrapper.cc
```diff
@@ -484,6 +493,56 @@ JNIEXPORT jlong JNICALL Java_org_apache_arrow_dataset_jni_JniWrapper_createScann
...
+  arrow::MemoryPool* pool = reinterpret_cast<arrow::MemoryPool*>(memory_pool_id);
+  if (pool == nullptr) {
+    JniThrow("Memory pool does not exist or has been closed");
+  }
+  std::shared_ptr<arrow::dataset::Dataset> dataset =
+      RetrieveNativeInstance<arrow::dataset::Dataset>(dataset_id);
+  std::shared_ptr<arrow::dataset::ScannerBuilder> scanner_builder =
+      JniGetOrThrow(dataset->NewScan());
+  JniAssertOkOrThrow(scanner_builder->Pool(pool));
+  if (substrait_expr_produce_or_filter != nullptr) {
```
lidavidm, line 514: this wasn't renamed?

java/dataset/src/main/cpp/jni_wrapper.cc
```diff
@@ -484,6 +493,56 @@ JNIEXPORT jlong JNICALL Java_org_apache_arrow_dataset_jni_JniWrapper_createScann
...
+    std::shared_ptr<arrow::Buffer> buffer = LoadArrowBufferFromByteBuffer(env,
+                                                            substrait_expr_produce_or_filter);
+    std::vector<arrow::compute::Expression> project_exprs;
+    std::vector<std::string> project_names;
+    std::optional<arrow::compute::Expression> filter_expr;
+    const arrow::engine::BoundExpressions bounded_expression =
+          JniGetOrThrow(arrow::engine::DeserializeExpressions(*buffer));
+    for(arrow::engine::NamedExpression named_expression :
```
lidavidm, line 522:
```suggestion
    arrow::engine::BoundExpressions bounded_expression =
          JniGetOrThrow(arrow::engine::DeserializeExpressions(*buffer));
    for(arrow::engine::NamedExpression& named_expression :
```

java/dataset/src/main/cpp/jni_wrapper.cc
```diff
@@ -484,6 +493,56 @@ JNIEXPORT jlong JNICALL Java_org_apache_arrow_dataset_jni_JniWrapper_createScann
...
+                                        bounded_expression.named_expressions) {
+      if (named_expression.expression.type()->id() == arrow::Type::BOOL) {
+        if (filter_expr.has_value()) {
+          JniThrow("Only one filter expression may be provided");
+        }
+        filter_expr = named_expression.expression;
+      } else {
+        project_exprs.push_back(named_expression.expression);
+        project_names.push_back(named_expression.name);
```
lidavidm, line 531:
```suggestion
        project_exprs.push_back(std::move(named_expression.expression));
        project_names.push_back(std::move(named_expression.name));
```

java/dataset/src/main/cpp/jni_wrapper.cc
```diff
@@ -484,6 +493,56 @@ JNIEXPORT jlong JNICALL Java_org_apache_arrow_dataset_jni_JniWrapper_createScann
...
+      }
+    }
+    JniAssertOkOrThrow(scanner_builder->Project(std::move(project_exprs), std::move(project_names)));
+    JniAssertOkOrThrow(scanner_builder->Filter(*std::move(filter_expr)));
```
lidavidm, line 535:
```suggestion
    JniAssertOkOrThrow(scanner_builder->Filter(*filter_expr));
```

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
```diff
@@ -27,6 +28,7 @@
 public class ScanOptions {
   private final Optional<String[]> columns;
   private final long batchSize;
+  private ByteBuffer substraitExtendedExpression;
```
lidavidm, line 31: This must be final.

lidavidm (changes requested):

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
davisusanibar: Yes, that will be an option. But currently we could define using the same extended expression:
- Or a projection
- Or a filter
- Or a projection + filter at the same time

I could divide that, then, 03 new options/methods will be created, or 01 methods with 03 new types.... Does it make sense?

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
davisusanibar: Oh, ok, let me consider the filter again

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
lidavidm: ? You'd have `setProjection`, `setFilter`, and either or both could be null (and both are exclusive with `columns`)

davisusanibar: Added

davisusanibar: Thank you, added

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
davisusanibar: This is optional, is up to the user to configure this or not, it is part of the builder option

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
davisusanibar: added

java/dataset/src/main/java/org/apache/arrow/dataset/source/Dataset.java
davisusanibar: changed

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
lidavidm: Please take another look at this. I'm not going to review this.

davisusanibar: There are: projection, filter and projectionAndFilter, that could by used at ScanOptions builder time, then this method is a helper needed for evaluate what of these option was used.

Could you help me to clarify what is your comment when you mention take another look on what is your recommendation please?

davisusanibar:
Hi @danepitkin could you help me also with a code review?

java/dataset/src/main/cpp/jni_wrapper.cc
```diff
@@ -458,8 +467,8 @@ JNIEXPORT void JNICALL Java_org_apache_arrow_dataset_jni_JniWrapper_closeDataset
  * Signature: (J[Ljava/lang/String;JJ)J
  */
 JNIEXPORT jlong JNICALL Java_org_apache_arrow_dataset_jni_JniWrapper_createScanner(
-    JNIEnv* env, jobject, jlong dataset_id, jobjectArray columns, jlong batch_size,
-    jlong memory_pool_id) {
+    JNIEnv* env, jobject, jlong dataset_id, jobjectArray columns,
+    jobject substrait_extended_expression, jlong batch_size, jlong memory_pool_id) {
```
danepitkin, line 471: Instead of passing `substrait_extended_expression`, can we pass in `filter` and `projection` parameters?

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
```diff
@@ -69,4 +73,106 @@ public Optional<String[]> getColumns() {
   public long getBatchSize() {
     return batchSize;
   }
+
+  private ByteBuffer getProjection() {
```
danepitkin, line 77: `getProjection` and `getFilter` should return optional values similar to `getColumns` if we remove `getSubstraitExtendedExpression` (see below).

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
```diff
@@ -27,6 +28,9 @@
 public class ScanOptions {
   private final Optional<String[]> columns;
   private final long batchSize;
+  private ByteBuffer projection;
+  private ByteBuffer filter;
```
danepitkin, line 32: Should projection/filter be `final` if we have a builder for this object? We want the object to be immutable after creation I think.

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
danepitkin: Hey @davisusanibar , I think we should remove `getProjectionAndFilter` and `getSubstraitExtendedExpression`. If the user wants to set both, they can set filter and projection separately.

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
```diff
@@ -69,4 +73,106 @@ public Optional<String[]> getColumns() {
...
+    private ByteBuffer filter;
+    private ByteBuffer projectionAndFilter;
+
+    /**
+     * Constructor.
+     * @param batchSize Maximum row number of each returned {@link org.apache.arrow.vector.ipc.message.ArrowRecordBatch}
+     * @param columns (Optional) Projected columns. {@link Optional#empty()} for scanning all columns. Otherwise,
+     *                Only columns present in the Array will be scanned.
+     */
+    public Builder(long batchSize, Optional<String[]> columns) {
```
danepitkin, line 122: Should a Builder API only enforce mandatory args in its constructor (e.g. `batchSize`)? `columns` is optional and can have its own builder method.

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
```diff
@@ -204,4 +205,93 @@ public void testRunBinaryQueryNamedTableNation() throws Exception {
       }
     }
   }
+
```
danepitkin, line 208: IMO it would be nice to see separate tests for `filter` and `projection` functionality.

java/dataset/src/main/cpp/jni_wrapper.cc
davisusanibar: It is possible to do that. It may be necessary to invoke Cpp twice to get bounded expressions, once for each filter and once for each projection. What do you think?

danepitkin: I think that should be fine. At this point we have already crossed the JNI boundary so I don't anticipate much performance impact in making the call twice. Ideally, we should keep the implementation flexible enough such that somebody can easily add support for Acero compute expressions as well. e.g. from the user perspective a `filter` can be either a compute expression or a substrait binary blob that gets parsed into a compute expression (wrapped in named/bounded expression objects). We don't need to add that additional functionality in this PR, though.

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
davisusanibar: Added

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
davisusanibar: Changed

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
davisusanibar: Changed

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
davisusanibar: Changed

java/dataset/src/main/cpp/jni_wrapper.cc
davisusanibar: Changed

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
```diff
@@ -69,4 +74,77 @@ public Optional<String[]> getColumns() {
...
     return batchSize;
   }
+
+  public Optional<ByteBuffer> getSubstraitExpressionProjection() {
+    return substraitExpressionProjection;
+  }
+
+  public Optional<ByteBuffer> getSubstraitExpressionFilter() {
+    return substraitExpressionFilter;
+  }
```
danepitkin, line 84: Can we call these `substrait_projection` and `substrait_filter`? I think we can either leave out the word "expression" or else change it "susbstrait_extended_expression_X" if we want to be verbose. I'm curious if other folks have thoughts on readability. Substrait will probably be a new concept to many Arrow Java users so I think it would be good have consistent and clear naming here.

If we change the naming, it would be best to change it everywhere e.g. in JNI/C++, too.

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
```diff
@@ -204,4 +205,167 @@ public void testRunBinaryQueryNamedTableNation() throws Exception {
...
+  @Test
+  public void testBaseParquetReadWithExtendedExpressionsFilter() throws Exception {
+    final Schema schema = new Schema(Arrays.asList(
+        Field.nullable("id", new ArrowType.Int(32, true)),
+        Field.nullable("name", new ArrowType.Utf8())
+    ), null);
+    // Substrait Extended Expression: Filter:
+    // Expression 01: WHERE ID < 20
+    String binarySubstraitExpressionFilter = "Ch4IARIaL2Z1bmN0aW9uc19jb21wYXJpc29uLnlhbWwSEhoQCAIQAhoKbHQ6YW55X2F" +
```
danepitkin, line 217: Nit: Would it be more descriptive to call this `base64EncodedSubstraitFilter`? Then it would get decoded into a var called `substraitFilter`. I usually don't like to get too picky with naming, but in this case it would be nice to be highly descriptive of what these random-looking strings are.

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
```diff
@@ -204,4 +205,167 @@ public void testRunBinaryQueryNamedTableNation() throws Exception {
...
+        Dataset dataset = datasetFactory.finish();
+        Scanner scanner = dataset.newScan(options);
+        ArrowReader reader = scanner.scanBatches()
+    ) {
+      assertEquals(schema.getFields(), reader.getVectorSchemaRoot().getSchema().getFields());
+      int rowcount = 0;
+      while (reader.loadNextBatch()) {
+        rowcount += reader.getVectorSchemaRoot().getRowCount();
+      }
+      assertEquals(3, rowcount);
```
danepitkin, line 242: Optional: Would it be easy to replace the final assertion in these test cases with an exact match of values instead of just row count? If not, then ignore this comment.

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
```diff
@@ -204,4 +205,167 @@ public void testRunBinaryQueryNamedTableNation() throws Exception {
...
+    }
+  }
```
danepitkin, line 283: Can we delete the code that should not run after the expected Exception is thrown?

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
```diff
@@ -204,4 +205,167 @@ public void testRunBinaryQueryNamedTableNation() throws Exception {
...
+      int rowcount = 0;
+      while (reader.loadNextBatch()) {
+        rowcount += reader.getVectorSchemaRoot().getRowCount();
+      }
+      assertEquals(5, rowcount);
+    }
+  }
+
+  @Test
+  public void testBaseParquetReadWithExtendedExpressionsProjectAndFilter() throws Exception {
```
danepitkin, line 325: Optional: There's a decent amount of duplicated code in the test cases. Do you think it would be possible to create a generic function that can be parameterized for each test case?

maybe something like this?
```
private ArrowReader scanParquetFileUsingSubstrait(String base64EncodedSubstraitFilter, String base64EncodedSubstraitProjection) throws Exception {...}
```

danepitkin (commented):
Great work! I left some minor comments (optional = can merge without updating, nit = nitpicking small/insignificant code changes)

docs/source/java/dataset.rst
```diff
@@ -159,6 +157,26 @@ Or use shortcut construtor:
...
 Then all columns will be emitted during scanning.
 
+Projection (Produce New Columns) and Filters
+============================================
+
+User can specify projections (new columns) or filters in ScanOptions. For example:
+
+.. code-block:: Java
+
+   ByteBuffer substraitExtendedExpressions = ...;
```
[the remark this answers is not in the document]
davisusanibar: Added

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
davisusanibar: Changed

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
davisusanibar: Changed

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
davisusanibar: Changed

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
davisusanibar: Deleted

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
davisusanibar: Added

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
```diff
@@ -69,4 +74,77 @@ public Optional<String[]> getColumns() {
...
+
+  /**
+   * Builder for Options used during scanning.
+   */
+  public static class Builder {
+    private final long batchSize;
+    private Optional<String[]> columns;
+    private ByteBuffer substraitProjection;
+    private ByteBuffer substraitFilter;
```
danepitkin, line 93: nit: could `Optional<ByteBuffer>` work here?

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
danepitkin: Nice!

nit: You can remove the rowcount assertion since you already check for exact values.

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
danepitkin: I see you added a helper function for base64 encoded filters. I was actually thinking of putting most of the test code in this helper function though like this (pseudo code):

```
private ArrowReader scanParquetFileUsingSubstrait(Optional<String> base64EncodedSubstraitFilter, Optional<String> base64EncodedSubstraitProjection) throws Exception {
    final Schema schema = new Schema(Arrays.asList(
        Field.nullable("id", new ArrowType.Int(32, true)),
        Field.nullable("name", new ArrowType.Utf8())
    ), null);
    ByteBuffer substraitExpressionFilter = getByteBuffer(base64EncodedSubstraitFilter);
    ByteBuffer substraitExpressionProjection = getByteBuffer(base64EncodedSubstraitProjection);
    ParquetWriteSupport writeSupport = ParquetWriteSupport
        .writeTempFile(AVRO_SCHEMA_USER, TMP.newFolder(), 19, "value_19", 1, "value_1",
            11, "value_11", 21, "value_21", 45, "value_45");
    ScanOptions options = new ScanOptions.Builder(/*batchSize*/ 32768)
        .columns(Optional.empty())
        .substraitFilter(substraitExpressionFilter)
        .substraitProjection(substraitExpressionProjection)
        .build();
    try (
        DatasetFactory datasetFactory = new FileSystemDatasetFactory(rootAllocator(), NativeMemoryPool.getDefault(),
            FileFormat.PARQUET, writeSupport.getOutputURI());
        Dataset dataset = datasetFactory.finish();
        Scanner scanner = dataset.newScan(options);
        ArrowReader reader = scanner.scanBatches()
    ) {
      assertEquals(schema.getFields(), reader.getVectorSchemaRoot().getSchema().getFields());
      return reader;
}


 @Test
  public void testBaseParquetReadWithExtendedExpressionsFilter() throws Exception {
    // Substrait Extended Expression: Filter:
    // Expression 01: WHERE ID < 20
    String base64EncodedSubstraitFilter = "Ch4IARIaL2Z1bmN0aW9uc19jb21wYXJpc29uLnlhbWwSEhoQCAIQAhoKbHQ6YW55X2F" +
        "ueRo3ChwaGggCGgQKAhABIggaBhIECgISACIGGgQKAigUGhdmaWx0ZXJfaWRfbG93ZXJfdGhhbl8yMCIaCgJJRAoETkFNRRIOCgQqAhA" +
        "BCgRiAhABGAI=";
    try (
        ArrowReader reader = scanParquetFileUsingSubstrait(base64EncodedSubstraitFilter, Optional.empty())
    ) {
      while (reader.loadNextBatch()) {
        assertTrue(reader.getVectorSchemaRoot().getVector("id").toString().equals("[19, 1, 11]"));
        assertTrue(reader.getVectorSchemaRoot().getVector("name").toString()
            .equals("[value_19, value_1, value_11]"));
      }
  }
}

 @Test
...
```

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
```diff
@@ -204,4 +205,170 @@ public void testRunBinaryQueryNamedTableNation() throws Exception {
...
+        Scanner scanner = dataset.newScan(options);
+        ArrowReader reader = scanner.scanBatches()
+    ) {
+      assertEquals(schema.getFields(), reader.getVectorSchemaRoot().getSchema().getFields());
+      int rowcount = 0;
+      while (reader.loadNextBatch()) {
+        rowcount += reader.getVectorSchemaRoot().getRowCount();
+        assertTrue(reader.getVectorSchemaRoot().getVector("id").toString().equals("[19, 1, 11]"));
+        assertTrue(reader.getVectorSchemaRoot().getVector("name").toString()
+            .equals("[value_19, value_1, value_11]"));
```
danepitkin, line 241: Is it possible to compare the actual vector instead of the string?

danepitkin (commented):
Great work! Overall LGTM

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
davisusanibar: Added

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
davisusanibar: For that purpose, a ValueVector needs to be created on the fly, mutating with fixed data added. It would be possible for me to do that, but it would add a step that will not be relevant to the purpose of this uni test. Let me confirm that this should be added

```
      IntVector valueVector = new IntVector("id", rootAllocator());
      valueVector.allocateNew(3);
      valueVector.set(0, 19);
      valueVector.set(1, 1);
      valueVector.set(2, 11);
      valueVector.setValueCount(3);
      ...
      assertEquals(reader.getVectorSchemaRoot().getVector("id").toString(), valueVector.toString());
```

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
davisusanibar: The validation is also useful for ensuring rowcount works correctly. How do you feel about this?

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
davisusanibar: This is due to the fact that ArrowReader references like DatasetFactory/Dataset/Scanner must be alive, so if we put all of them in try-catch, ArrowReader will be invalidated. We can create instance variables, but this adds more confusion in declare, assign values, and release these resources variables properly.

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
danepitkin: Ah I see. I was hoping there would be a `.to_pojo()` type of option where you could compare to a Java ListArray or something like that. Maybe better not to add too much unnecessary code to the test cases..

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
danepitkin: Makes sense, let's keep the tests the way you have it. I agree that sounds like the better option.

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
danepitkin: Looks good!

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
```diff
@@ -204,4 +205,173 @@ public void testRunBinaryQueryNamedTableNation() throws Exception {
...
+            .equals("[21, 3, 13]"));
+        assertTrue(reader.getVectorSchemaRoot().getVector("concat_column_a_and_b").toString()
+            .equals("[value_19 - value_19, value_1 - value_1, value_11 - value_11]"));
+        rowcount += reader.getVectorSchemaRoot().getRowCount();
+      }
+      assertEquals(3, rowcount);
+    }
+  }
+
+  private static ByteBuffer getByteBuffer(String base64EncodedSubstraitFilter) {
```
danepitkin, line 371: `base64EncodedSubstraitFilter` -> `base64EncodedSubstrait` since this can be both a filter or projection.

danepitkin:
I added one last small comment about the tests, but otherwise LGTM! I would approve it if I had the capability.

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
davisusanibar: Thank you, changed

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
```diff
@@ -204,4 +205,173 @@ public void testRunBinaryQueryNamedTableNation() throws Exception {
...
+            .equals("[value_19 - value_19, value_1 - value_1, value_11 - value_11]"));
+        rowcount += reader.getVectorSchemaRoot().getRowCount();
+      }
+      assertEquals(3, rowcount);
+    }
+  }
+
+  private static ByteBuffer getByteBuffer(String base64EncodedSubstrait) {
+    byte[] substraitFilter = Base64.getDecoder().decode(base64EncodedSubstrait);
+    ByteBuffer substraitExpressionFilter = ByteBuffer.allocateDirect(substraitFilter.length);
```
danepitkin, line 373: Don't forget about the rest of this function! e.g. `substraitExpressionFilter` and `substraitFilter`

davisusanibar: Changed

java/dataset/src/main/java/org/apache/arrow/dataset/scanner/ScanOptions.java
```diff
@@ -69,4 +74,77 @@ public Optional<String[]> getColumns() {
...
+      this.batchSize = batchSize;
+    }
+
+    /**
+     * Set the Projected columns. Empty for scanning all columns.
+     *
+     * @param columns Projected columns. Empty for scanning all columns.
+     * @return the ScanOptions configured.
+     */
+    public Builder columns(Optional<String[]> columns) {
```
danepitkin, line 109: One more thing: We don't need `Optional<>` parameters for the builder APIs. We should expect the user to passing us a valid object. Same with `substraitProjection` and `substraitFilter`

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
lidavidm: I think we should consider what to do here; possibly an Iterator/Iterable/Stream that gives Java objects would be sufficient (and you could collect into a Java collection and then use standard assertions). Can you file a follow-up task?

java/dataset/src/main/cpp/jni_wrapper.cc
```diff
@@ -474,6 +484,39 @@ JNIEXPORT jlong JNICALL Java_org_apache_arrow_dataset_jni_JniWrapper_createScann
...
+  if (substrait_projection != nullptr) {
+    std::shared_ptr<arrow::Buffer> buffer = LoadArrowBufferFromByteBuffer(env,
+                                                            substrait_projection);
+    std::vector<arrow::compute::Expression> project_exprs;
+    std::vector<std::string> project_names;
+    arrow::engine::BoundExpressions bounded_expression =
+          JniGetOrThrow(arrow::engine::DeserializeExpressions(*buffer));
+    for(arrow::engine::NamedExpression& named_expression :
+                                        bounded_expression.named_expressions) {
+      if (!(named_expression.expression.type()->id() == arrow::Type::BOOL)) {
```
lidavidm, line 496:
```suggestion
      if (named_expression.expression.type()->id() != arrow::Type::BOOL) {
```

lidavidm: Why do we have this in the first place? I think this was left over from a refactor as seen below. It should be perfectly fine to project a BOOL column.

java/dataset/src/main/cpp/jni_wrapper.cc
```diff
@@ -474,6 +484,39 @@ JNIEXPORT jlong JNICALL Java_org_apache_arrow_dataset_jni_JniWrapper_createScann
...
+    for(arrow::engine::NamedExpression& named_expression :
+                                        bounded_expression.named_expressions) {
+      if (named_expression.expression.type()->id() == arrow::Type::BOOL) {
+        if (filter_expr.has_value()) {
+          JniThrow("Only one filter expression may be provided");
+        }
+        filter_expr = named_expression.expression;
+      }
+    }
+    JniAssertOkOrThrow(scanner_builder->Filter(*filter_expr));
```
lidavidm, line 518: This will crash if you provide an empty list of expressions.

java/dataset/src/main/cpp/jni_wrapper.cc
```diff
@@ -474,6 +484,39 @@ JNIEXPORT jlong JNICALL Java_org_apache_arrow_dataset_jni_JniWrapper_createScann
...
+    arrow::engine::BoundExpressions bounded_expression =
+          JniGetOrThrow(arrow::engine::DeserializeExpressions(*buffer));
+    for(arrow::engine::NamedExpression& named_expression :
+                                        bounded_expression.named_expressions) {
+      if (named_expression.expression.type()->id() == arrow::Type::BOOL) {
+        if (filter_expr.has_value()) {
+          JniThrow("Only one filter expression may be provided");
+        }
+        filter_expr = named_expression.expression;
+      }
```
lidavidm, line 516: Throw if the expression is not of type BOOL.

docs/source/java/substrait.rst
```diff
@@ -102,6 +104,349 @@ Here is an example of a Java program that queries a Parquet file using Java Subs
...
+                    Scanner scanner = dataset.newScan(options);
+                    ArrowReader reader = scanner.scanBatches()
+            ) {
+                while (reader.loadNextBatch()) {
+                    System.out.println(
+                            reader.getVectorSchemaRoot().contentToTSVString());
+                }
+            } catch (Exception e) {
+                e.printStackTrace();
+            }
```
lidavidm, line 184: just declare everything as `throws Exception`

docs/source/java/substrait.rst
```diff
@@ -102,6 +104,349 @@ Here is an example of a Java program that queries a Parquet file using Java Subs
...
+                    setSelection(
+                            Expression.FieldReference.newBuilder().
+                                    setDirectReference(
+                                            Expression.ReferenceSegment.newBuilder().
+                                                    setStructField(
+                                                            Expression.ReferenceSegment.StructField.newBuilder().setField(
+                                                                    2)
+                                                    )
+                                    )
+                    );
```
lidavidm, line 199: nit: how are you formatting examples? I think it would make sense to just use `google-java-format` which has a more compact style. In our docs, going too far to the right is unreadable.

lidavidm (changes requested):

java/dataset/src/test/java/org/apache/arrow/dataset/substrait/TestAceroSubstraitConsumer.java
davisusanibar: Just filled https://github.com/apache/arrow/issues/37728

docs/source/java/substrait.rst
davisusanibar: changed

docs/source/java/substrait.rst
davisusanibar: added

java/dataset/src/main/cpp/jni_wrapper.cc
davisusanibar: added

java/dataset/src/main/cpp/jni_wrapper.cc
davisusanibar: changed

java/dataset/src/main/cpp/jni_wrapper.cc
davisusanibar: updated

docs/source/java/substrait.rst
```diff
@@ -102,6 +104,350 @@ Here is an example of a Java program that queries a Parquet file using Java Subs
...
+        // project and filter dataset using extended expression definition - 03 Expressions:
+        // Expression 01 - CONCAT: N_NAME || ' - ' || N_COMMENT = col 1 || ' - ' || col 3
+        // Expression 02 - ADD: N_REGIONKEY + 10 = col 1 + 10
+        // Expression 03 - FILTER: N_NATIONKEY > 18 = col 3 > 18
+        projectAndFilterDataset();
+      }
+
+      public static void projectAndFilterDataset() {
+        //String uri = "file:///Users/dsusanibar/data/tpch_parquet/nation.parquet";
+        String uri = "file:////Users/dsusanibar/voltron/fork/consumer-testing/tests/data/tpch_parquet/nation.parquet";
```
lidavidm, line 164: Can we not put company names in strings? And remove the redundant comment.

davisusanibar: Sorry, second time same error

davisusanibar: Just updated

lidavidm (approved):

PR #35574, opened 2023-05-12
GH-35559: [Java] Implementing JDBC Flight Stream Result Set asynchronous VectorSchemaRoot Producer

### LINKED ISSUE #35559: [Java] JDBC Flight Stream Result Set asynchronous VectorSchemaRoot Producer
The current [ArrowFlightJdbcFlightStreamResultSet](https://github.com/apache/arrow/blob/10bc4585b46048ec3717c198dc7e648a55cd08c9/java/flight/flight-sql-jdbc-driver/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightJdbcFlightStreamResultSet.java) gets a VectorSchemaRoot and then executes it every time ArrowFlightStreamResultSet.next() gets called. There is a performance gain to be had if a double buffer was implemented so that the consumption of the VectorSchemaRoot can be separated from the production.

The goal of this ticket is to implement double buffering such that:
- VectorSchemaRoots are produced asynchronously and stored in a blocking queue.
- when ArrowFlightStreamResultSet.next() gets called, the Result Set just gets the next available VectorSchemaRoot from the Blocking queue and executes it.

### Component(s)
Java

### Component(s)

Java

ArgusLi:
### Rationale for this change

The change leads to a performance increase when there is more than 2 million rows of data:
image

- Implement a double buffer blocking queue for VectorSchemaRoots.
- Produce VectorSchemaRoots Asynchronously.
- Add reading "buffersize" driver property to initalise blocking queue with a custom value.
- Use VectorLoader and VectorUnloader.

### Are these changes tested?

Yes. There is no change in functionality, so the existing tests are adequate.

No.

* Closes: #35559


lidavidm:
Want to rebase this? Looks like it won't build until conflicts are fixed

lidavidm:
Is 0.5% performance improvement noticeable or even statistically significant?

ArgusLi:
> Is 0.5% performance improvement noticeable or even statistically significant?

@lidavidm the posted results were taken from an automated test that we've discovered may not be the best indicator of performance. A more recent manual test fetching 500k rows has shown that this change led to a 4% faster performance - 114,256 ms vs 118,706 ms.

The performance enhancement is iterative, we wanted to create this PR early to engage the community and get feedback as early as possible. More results are pending as we are working with a customer to see results in their lab. Our customer has a use case that leads to a fast consumer, they move millions of rows over the network back and forth, so there was a hypothesis that implementing double buffering would lead to a performance increase.

ArgusLi:
Going to recreate the pull request in a cleaner fashion. Apologies all.

PR #35578, opened 2023-05-12
MINOR: [Doc] Update Parquet documentation website links

amoeba:
### Rationale for this change

While reviewing the Arrow docs on Parquet, I noticed two links to https://parquet.apache.org/documentation/latest/ which 404 for me. As far as I can tell, the docs site was redone and moved to https://parquet.apache.org/docs some time in 2022.

### Are these changes tested?

I didn't go so far as to find the associated Jira issue for the docs website refactor so I'd appreciate if someone more familiar with the project could double-check if the old URL is indeed permanently dead.

### DIFF
--- a/docs/source/cpp/parquet.rst
+++ b/docs/source/cpp/parquet.rst
@@ -27,7 +27,7 @@ Reading and writing Parquet files
 .. seealso::
    :ref:`Parquet reader and writer API reference <cpp-api-parquet>`.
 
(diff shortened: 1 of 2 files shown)


mapleFU (approved):
Nice catch!

pitrou (approved):
Thanks for this @amoeba ! This LGTM.

PR #35580, opened 2023-05-13
GH-35296: [Go] Add arrow.Table.String()

### LINKED ISSUE #35296: [Go] Add arrow.Table.String()

If we have it, we can print a `arrow.Table` easily.

FYI: C++ has `arrow::Table::ToString()`:  https://arrow.apache.org/docs/cpp/api/table.html#_CPPv4NK5arrow11RecordBatch8ToStringEv

See also https://github.com/apache/arrow/pull/35178#discussion_r1174584004 for a use case.

### Component(s)

Go

vibhatha:
This PR fixes https://github.com/apache/arrow/issues/35296
* Closes: #35296

### DIFF
--- a/go/arrow/array/table.go
+++ b/go/arrow/array/table.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"strings"
 	"sync/atomic"
 
 	"github.com/apache/arrow/go/v14/arrow"
@@ -255,6 +256,32 @@ func (tbl *simpleTable) Release() {
 		}
 		tbl.cols = nil
 	}
+}
+
+func (tbl *simpleTable) String() string {
+	o := new(strings.Builder)
+	o.WriteString(tbl.Schema().String())
+	o.WriteString("\n")
+
+	for i := 0; i < int(tbl.NumCols()); i++ {
+		col := tbl.Column(i)
+		chunked := col.Data()
+		chunks := chunked.Chunks()
+		o_array := new(strings.Builder)
+		o_array.WriteString(col.Field().Name + ": [")
+		numChunks := len(chunks)
+		for j := 0; j < numChunks; j++ {
+			chunk := chunked.Chunk(j)
+			if j != numChunks-1 {
+				o_array.WriteString(chunk.String() + ", ")
+			} else {
+				o_array.WriteString(chunk.String())
+			}
+
+		}
+		o.WriteString(o_array.String() + "]\n")
+	}
+	return o.String()
 }
 
 // TableReader is a Record iterator over a (possibly chunked) Table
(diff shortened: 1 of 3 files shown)


zeroshade:
@vibhatha please add a unit test for this

vibhatha:
@zeroshade addressed the reviews.

go/arrow/array/table.go
```diff
@@ -257,6 +258,32 @@ func (tbl *simpleTable) Release() {
...
+	o := new(strings.Builder)
+	o.WriteString(tbl.Schema().String())
+	o.WriteString("\n")
+
+	for i := 0; i < int(tbl.NumCols()); i++ {
+		col := tbl.Column(i)
+		chunked := col.Data()
+		chunks := chunked.Chunks()
+		o_array := new(strings.Builder)
+		o_array.WriteString(col.Field().Name + ": [")
```
zeroshade, line 271: why the need for the separate builder? why not just keep writing to `o` the whole time?

go/arrow/array/table.go
```diff
@@ -257,6 +258,32 @@ func (tbl *simpleTable) Release() {
...
+		numChunks := len(chunks)
+		for j := 0; j < numChunks; j++ {
+			chunk := chunked.Chunk(j)
+			if j != numChunks-1 {
+				o_array.WriteString(chunk.String() + ", ")
+			} else {
+				o_array.WriteString(chunk.String())
+			}
+
```
zeroshade, line 280: This should be rewritten as 

```go
for j, chunk := range col.Data().Chunks() {
        if j != 0 {
                o.WriteString(", ")
        }
        o.WriteString(chunk.String())
}
```

go/arrow/array/table.go
vibhatha: yeah we could use the `o`. Added just for the readability.

zeroshade: I'd prefer to avoid the extra builder because of the cost of creating a new builder for every column. It would be more readable to avoid the intermediate string builder (not to mention potentially large allocations as it ends up having to deep copy the string)

amol-:
@vibhatha Could you please address any pending review? So that @zeroshade can re-review

vibhatha:
@amol- sure I will address them soon. Sorry about the delay.

go/arrow/array/table.go
vibhatha: You're correct, I updated the code.

go/arrow/array/table.go
vibhatha: I updated the code. :+1:

vibhatha:
Updated the code and waiting for CIs. 
cc @amol- I will ping for reviews once this stage is in clear.

vibhatha:
@zeroshade one CI is failing not exactly sure if it is related.

zeroshade (approved):

PR #35582, opened 2023-05-13
GH-34787: [Python] Accept zero_copy_only=False for ChunkedArray.to_numpy

### LINKED ISSUE #34787: [Python] Accept keywords in ChunkedArray.to_numpy() for compatibility with Array.to_numpy()

Currently, if you have "a" pyarrow array object that can be either array or chunked array, and you call `to_numpy` with something that cannot be zero copy, you need to do:

```python
if isinstance(array, pyarrow.Array):
    np_arr = array.to_numpy(zero_copy_only=False)
else:
    np_arr = array.to_numpy()
```

instead of simply:

```python
np_arr = array.to_numpy(zero_copy_only=False)
```
because the chunked array version doesn't accept the keyword. To make this easier, I think we should just accept that keyword in `ChunkedArray.to_numpy()` as well (and only accept False as a possible value)

### Component(s)

Python

jjerphan:
### Rationale for this change

Addresses GH-34787.

I think this proposes the minimal viable changes to have the same signature as proposed in GH-34787.

### Are these changes tested?

Yes, for now, a single simple test has been introduced to test the correct behavior of `ChunkedArray.to_numpy` with respect to the addition of `zero_copy_only`.

### Comments

Looking at the signature of `pa.Array.to_numpy`:

https://github.com/apache/arrow/blob/b211c1896916e3d0bbad1ffc88a30ddb9c90a5f0/python/pyarrow/array.pxi#L1486-L1512

I think `writable` can be added as a default-valued parameter to `pa.ChunkedArray.to_numpy`.

What do you think?

* Closes: #34787


AlenkaF:
Thank you for the contribution @jjerphan!

> I think writable can be added as a default-valued parameter to pa.ChunkedArray.to_numpy.

Yes, I think that would be useful to add 👍

jjerphan:
> Thank you for the contribution @jjerphan!
> 
> > I think writable can be added as a default-valued parameter to pa.ChunkedArray.to_numpy.
> 
> Yes, I think that would be useful to add +1

Thanks for the feedback, @AlenkaF!

Should it be done in another PR?

AlenkaF:
> Should it be done in another PR?

I would personally add it in this PR. WIll change the title to include both changes.

python/pyarrow/table.pxi
```diff
@@ -491,6 +497,11 @@ cdef class ChunkedArray(_PandasConvertible):
             PandasOptions c_options
             object values
 
+        if zero_copy_only:
```
alexswo, line 500: nit, how about putting the `if/raise` statement before the `cdef`?

python/pyarrow/tests/test_array.py
```diff
@@ -168,6 +168,28 @@ def test_to_numpy_zero_copy():
...
+        chunked_arr.to_numpy(zero_copy_only=True)
+
+    np_arr = chunked_arr.to_numpy()
+
+    chunked_arr = None
+    import gc
+    gc.collect()
+
+    # Ensure base is still valid
+    assert np_arr.base is not None
```
jorisvandenbossche, line 188: I am not sure this part is needed here, as this is not actually zero copy (the base is still set, because we concatenate on the arrow side, and then still keep this concatenated buffer alive through a capsule as the numpy array's base).

jorisvandenbossche (approved):
Looks good to me! It's a goode idea to add the `writeable` keyword here as well, but I think also fine to leave that for another PR.

jjerphan:
Thank you for the review, @jorisvandenbossche! I will come back to this PR when I get some bandwidth.

python/pyarrow/tests/test_array.py
jjerphan: OK, so we do not want to test the presence of `np_arr.base` here?

jorisvandenbossche: I don't think that is needed, it's in any case not specific to this test

jorisvandenbossche:
Thanks @jjerphan! Already merging this, so it can be included in pyarrow 13.0.0. But a follow-up PR to add `writeable` as well is certainly welcome.

PR #35592, opened 2023-05-15
GH-35539: [C++] Remove use of internal header files from public header file

### LINKED ISSUE #35539: [C++] Arrow 12.0.0-1 cpp exposes internal headers

it exposes compute/key_map.h.
However, compute/key_map.h relies on compute/util_internal.h which is not exposed.

Ubuntu

### Component(s)

C++

ildipo:

Remove use of compute/util_internal.h in compute/key_map.h

* Closes: #35539


pitrou:
Thanks @ildipo . I also removed use of `arrow/util/logging.h` which should be avoided in public headers.

PR #35597, opened 2023-05-15
GH-35596: [C++][CI] Improve compilation caching with PCG

### LINKED ISSUE #35596: [C++] PCG breaks compilation caching

Noticing that I sometimes got unexpected slow rebuilds, I tried to dive a bit into ccache debug output and found out that by default the PCG headers use the `__DATE__` and `__TIME__` macros, which makes builds non-deterministic:
https://github.com/imneme/pcg-cpp/issues/59

We should perhaps simply remove PCG's `struct static_arbitrary_seed`, which would make compilation caching more efficient. @benibus @bkietz @assignUser 

### Component(s)

C++

pitrou:
### Rationale for this change

The PCG headers use the `__DATE__` and `__TIME__` macros, which makes builds non-deterministic:
https://github.com/imneme/pcg-cpp/issues/59

Deterministic builds are useful for a number of reasons. One is security audits of binary artifacts, another (that directly affects us) is making compilation caching as efficient as possible. Preprocessor-based compilation caching breaks when time-dependent macros are used.

Remove the `struct static_arbitrary_seed` construct from PCG.
Also, enable a gcc/clang warning that detects the use of non-deterministic preprocessor macros.

### Are these changes tested?

Yes, by the additional warning (which is turned into an error in "checkin" warnings mode).

No.

* Closes: #35596


benibus (approved):

assignUser (approved):
Sounds good, it seems this is not a feature used in arrow (or it's development)?

PR #35603, opened 2023-05-16
GH-35559: [Java] Implementing JDBC Flight Stream Result Set asynchronous VectorSchemaRoot Producer

### LINKED ISSUE #35559: [Java] JDBC Flight Stream Result Set asynchronous VectorSchemaRoot Producer

ArgusLi:
### Rationale for this change

The change leads to a performance increase when there is more than 2 million rows of data:
image
These results were taken from an automated test that we've discovered may not be the best indicator of performance.

A more recent manual test fetching 500k rows has shown that this change led to a 4% faster performance - 114,256 ms vs 118,706 ms.

The performance enhancement is iterative, we wanted to create this PR early to engage the community and get feedback as early as possible. More results are pending as we are working with a customer to see results in their lab. Our customer has a use case that leads to a fast consumer, they move millions of rows over the network back and forth, so there was a hypothesis that implementing double buffering would lead to a performance increase.

- Implement a double buffer blocking queue for VectorSchemaRoots.
- Produce VectorSchemaRoots Asynchronously.
- Add reading "buffersize" driver property to initalise blocking queue with a custom value.
- Use VectorLoader and VectorUnloader.

### Are these changes tested?

Yes. There is no change in functionality, so the existing tests are adequate.

No.

* Closes: #35559


lidavidm:
So gRPC should already be buffering data, meaning that this change is mostly about pipelining the IPC work - except IPC deserialization should be very cheap. That could explain why the benefits are small.

4% is something, but the effects still seem very small.

If performance with millions of rows is a concern, has there been any profiling or tracing to identify the bottlenecks?

lidavidm:
> So gRPC should already be buffering data

The flip side is that due to a design decision in gRPC-Java, it's actually hard for a Java _server_ to properly saturate the connection, because gRPC uses a fixed-size buffer for its backpressure indicator. That may also be something to explore, if not already explored. (See https://github.com/grpc/proposal/pull/135)

ArgusLi:
@lidavidm That is quite interesting... thanks for the link!

We did do some profiling - multiple thread dumps while the JDBC client was fetching data and in all the snapshots it was waiting for next().

```
DBeaver: Read data [SELECT * FROM "mongo.yelp"."review"]awaiting notification on [ 0x00000007cb2c85a0 ]
at jdk.internal.misc.Unsafe.park(java.base@17/Native Method)
at java.util.concurrent.locks.LockSupport.park(java.base@17/Unknown Source)
at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionNode.block(java.base@17/Unknown Source)
at java.util.concurrent.ForkJoinPool.unmanagedBlock(java.base@17/Unknown Source)
at java.util.concurrent.ForkJoinPool.managedBlock(java.base@17/Unknown Source)
at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(java.base@17/Unknown Source)
at java.util.concurrent.LinkedBlockingQueue.take(java.base@17/Unknown Source)
at cfjd.org.apache.arrow.flight.FlightStream.next(FlightStream.java:231)
at org.apache.arrow.driver.jdbc.ArrowFlightJdbcFlightStreamResultSet.next(ArrowFlightJdbcFlightStreamResultSet.java:184)
at org.jkiss.dbeaver.model.impl.jdbc.exec.JDBCResultSetImpl.next(JDBCResultSetImpl.java:272)
```

lidavidm:
That's not the right thread to focus on then. If most of the time is spent waiting for `take`, that means gRPC isn't populating the queue fast enough, i.e. it seems most of the time is actually waiting for data transfer. (Or possibly we're being slow at processing messages from gRPC, which is FlightStream.Observer#onNext.) And so then I'd be curious about things like, what is the bandwidth the client is getting? What is the max theoretical bandwidth, what is the observed bandwidth in a synthetic benchmark (like iperf3)?

4%, without further qualification, could still be in the realm of natural variation.

ArgusLi:
Thanks David, you've given us quite a lot to think about. We'll investigate the configurable back pressure threshold and look at the customer's use case a bit more. Thanks once again.

ArgusLi:
@lidavidm the customer has come back with data that shows this change leads to a decrease of 5-40% in time taken. Unfortunately, they showed the data in a screen share during a meeting, and due to client confidentiality, we are not at liberty to share this data. Given that we have data that shows there is a performance enhancement due to this change, we’d like to push this PR forward.

lidavidm:
Hmm, I'm still a little surprised, since it's unclear why this is faster, for the reasons I already gave. But I'll give it a review when I find time.

ArgusLi:
Thanks @lidavidm!!!

java/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightConnection.java
```diff
@@ -110,6 +110,7 @@ private static ArrowFlightSqlClientHandler createNewClientHandler(
     } catch (final SQLException e) {
       try {
         allocator.close();
+        allocator.getChildAllocators().forEach(BufferAllocator::close);
```
lidavidm, line 113: 1) Can we use AutoCloseables?
2) This needs to come before the root allocator is closed anyways

java/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightJdbcFlightStreamResultSet.java
```diff
@@ -204,6 +259,21 @@ public boolean next() throws SQLException {
...
+      flightStreamQueue.close();
+    } else if (currentFlightStream != null) {
+      // close is only called for currentFlightStream if there's no queue
+      currentFlightStream.close();
+    }
+
+    List<VectorSchemaRoot> roots = new LinkedList<>();
+    vectorSchemaRoots.drainTo(roots);
+    roots.forEach(AutoCloseables::closeNoChecked);
+    ofNullable(currentRoot).ifPresent(AutoCloseables::closeNoChecked);
```
lidavidm, line 274: It would be preferable to do this all as one big AutoCloseables call if possible (also, is LinkedList strictly necessary here vs ArrayList?)

java/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightJdbcFlightStreamResultSet.java
```diff
@@ -73,6 +86,7 @@ public final class ArrowFlightJdbcFlightStreamResultSet
                                        final Meta.Frame firstFrame) throws SQLException {
     super(null, state, signature, resultSetMetaData, timeZone, firstFrame);
     this.connection = connection;
+    initializeVectorSchemaRootsQueue();
   }
```
lidavidm, line 90: Would it be possible to have both constructors delegate to a single constructor, then inline initializeVectorSchemaRootsQueue into that constructor?

java/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightJdbcFlightStreamResultSet.java
```diff
@@ -45,15 +54,18 @@
  */
 public final class ArrowFlightJdbcFlightStreamResultSet
     extends ArrowFlightJdbcVectorSchemaRootResultSet {
-
+  private static final String BLOCKING_QUEUE_PARAM = "buffersize";
```
lidavidm, line 57: Can we add this to the documentation?

Can we add a test of setting this parameter?

java/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightJdbcFlightStreamResultSet.java
```diff
@@ -130,39 +146,89 @@ protected AvaticaResultSet execute() throws SQLException {
...
+      executeNextRoot();
+    }
+  }
+
+  private BufferAllocator getAllocator() {
+    if (allocator == null) {
+      allocator = connection.getBufferAllocator().newChildAllocator("vsr-copier", 0, Long.MAX_VALUE);
     }
+
+    return allocator;
```
lidavidm, line 170: Can this also be just initialized in the constructor?

java/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightJdbcFlightStreamResultSet.java
```diff
@@ -130,39 +146,89 @@ protected AvaticaResultSet execute() throws SQLException {
...
+      ofNullable(currentRoot).ifPresent(AutoCloseables::closeNoChecked);
+      currentRoot = vectorSchemaRoots.poll(10, TimeUnit.SECONDS);
+      execute(currentRoot, schema);
+    } catch (InterruptedException e) {
+      throw new SQLException("Could not take root from the queue", e);
+    }
+  }
+
+  private void storeRootsFromStreamAsync() {
+    CompletableFuture.runAsync(() -> {
```
lidavidm, line 213: This uses a global thread pool, can we instead use a pool we control?

java/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightJdbcFlightStreamResultSet.java
```diff
@@ -130,39 +146,89 @@ protected AvaticaResultSet execute() throws SQLException {
...
+      while (vectorSchemaRoots.remainingCapacity() > 0) {
```
lidavidm, line 214: multiple calls to `next` might spawn this task multiple times, and then they may push too many items onto the queue, right?

java/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightJdbcFlightStreamResultSet.java
```diff
@@ -130,39 +146,89 @@ protected AvaticaResultSet execute() throws SQLException {
...
+      vectorSchemaRoots.put(ofNullable(transformedRoot).orElse(theRoot));
+    } catch (InterruptedException e) {
+      throw new SQLException("Could not put root to the queue", e);
     }
   }
 
+  private void executeNextRoot() throws SQLException {
+    try {
+      ofNullable(currentRoot).ifPresent(AutoCloseables::closeNoChecked);
+      currentRoot = vectorSchemaRoots.poll(10, TimeUnit.SECONDS);
```
lidavidm, line 205: Do we want to have a hard timeout like this? We don't know how long it might take for the server to respond

java/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightJdbcFlightStreamResultSet.java
```diff
@@ -130,39 +146,89 @@ protected AvaticaResultSet execute() throws SQLException {
...
+          storeRoot(currentFlightStream.getRoot());
+        } catch (SQLException e) {
+          throw new RuntimeException(e);
+        }
+      }
+    });
+  }
+
   @Override
   public boolean next() throws SQLException {
```
lidavidm, line 230: The control flow between here and the background task are hard to follow; it would help to document what's going on

java/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightJdbcFlightStreamResultSet.java
```diff
@@ -130,39 +146,89 @@ protected AvaticaResultSet execute() throws SQLException {
...
+  }
 
+  private void storeRoot(VectorSchemaRoot originalRoot) throws SQLException {
+    VectorSchemaRoot theRoot = cloneRoot(originalRoot);
+    VectorSchemaRoot transformedRoot = null;
     if (transformer != null) {
       try {
-        currentVectorSchemaRoot = transformer.transform(originalRoot, currentVectorSchemaRoot);
+        transformedRoot = transformer.transform(theRoot, null);
+        theRoot.close();
```
lidavidm, line 189: Can we use try-with-resources here over explicit close?

lidavidm:
Additionally, it seems CI is failing

java/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightJdbcFlightStreamResultSet.java
```diff
@@ -99,23 +136,33 @@ static ArrowFlightJdbcFlightStreamResultSet fromFlightInfo(
...
 
-  private void loadNewFlightStream() throws SQLException {
-    if (currentFlightStream != null) {
-      AutoCloseables.closeNoChecked(currentFlightStream);
+  private void initializeVectorSchemaRootsQueue(int blockingQueueCapacity) {
+    if (vectorSchemaRoots == null) {
+      vectorSchemaRoots = new LinkedBlockingQueue<>(blockingQueueCapacity);
     }
-    this.currentFlightStream = getNextFlightStream(true);
   }
```
lidavidm, line 166: Does this really need to be a separate method? It can just be inlined into the constructor. (Also, won't vectorSchemaRoots always be null to start with? If we initialize it in the constructor, we can make it final too.)

lidavidm (approved):

lidavidm:
That said, it seems things are still failing. (Possibly close() needs to wait for any background tasks to complete/shut down the thread pool?)

PR #35613, opened 2023-05-16
GH-35611: [C++] Remove unnecessary safe operations for ListBuilder and BinaryBuilder

### LINKED ISSUE #35611: [C++] Remove unnecessary safe operations for ListBuilder and BinaryBuilder

There are several safety checks/operations that can be optimized to enhance performance of ListBuilder and BinaryBuild, including: 
1.  Redundant safety checks
2. Unnecessary safe `Append` after `Reserve` is called
3. In AppendArraySlice, space is reserved for each array element but pre-allocation for the whole batch is possible

### Component(s)

C++

js8544:
### Rationale for this change

There are several safety checks/operations that can be optimized to enhance performance of ListBuilder and BinaryBuild

1. Remove redundant safety checks
2. replace Append with UnsafeAppend when possible
3. In AppendArraySlice, pre-allocate space for the whole batch.

### Are these changes tested?

Yes, by existing tests in array_test.cc.

No. 

* Closes: #35611

### DIFF
--- a/cpp/src/arrow/array/builder_binary.h
+++ b/cpp/src/arrow/array/builder_binary.h
@@ -67,7 +67,7 @@ class BaseBinaryBuilder
 
   Status Append(const uint8_t* value, offset_type length) {
     ARROW_RETURN_NOT_OK(Reserve(1));
-    ARROW_RETURN_NOT_OK(AppendNextOffset());
+    UnsafeAppendNextOffset();
     // Safety check for UBSAN.
     if (ARROW_PREDICT_TRUE(length > 0)) {
       ARROW_RETURN_NOT_OK(ValidateOverflow(length));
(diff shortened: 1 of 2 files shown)


cpp/src/arrow/array/builder_binary.h
```diff
@@ -67,7 +67,7 @@ class BaseBinaryBuilder
 
   Status Append(const uint8_t* value, offset_type length) {
     ARROW_RETURN_NOT_OK(Reserve(1));
-    ARROW_RETURN_NOT_OK(AppendNextOffset());
+    UnsafeAppendNextOffset();
```
mapleFU, line 70: Hi, since `Append` is a public interface, how can user make sure that the offset is reserved before?

cpp/src/arrow/array/builder_nested.h
```diff
@@ -202,6 +201,11 @@ class BaseListBuilder : public ArrayBuilder {
     const int64_t num_values = value_builder_->length();
     return offsets_builder_.Append(static_cast<offset_type>(num_values));
   }
+
+  void UnsafeAppendNextOffset() {
```
benibus, line 205: Minor nit, but perhaps we could use this in `AppendNulls` and `AppendEmptyValues`.

benibus (commented):
Thanks. This looks sound IMO.

cpp/src/arrow/array/builder_nested.h
js8544: I'm not sure. In `AppendNulls` and `AppendEmptyValues` the same offset is appended multiple times. Currently this number is only computed once. If we use `UnsafeAppendNextOffset` there, the line `const int64_t num_values = value_builder_->length();` will be computed multiple times.

pitrou:
@js8544 Could you rebase on git main to try and get passing CI?

pitrou:
Ok, thank you @js8544 !

PR #35626, opened 2023-05-16
GH-35617: [Docs] Current n_buffers use in C API example

### LINKED ISSUE #35617: [Docs] C API example, allocation of child buffers references wrong `n_buffers`

This is something spotted by `gcc-12` as I was working around the C API examples.  When (near the very bottom of the [C API documentation](https://arrow.apache.org/docs/format/CDataInterface.html)) the example allocates buffers, it references the parent structure in

```c
malloc(sizeof(void*) * array->n_buffers);
```

This should be 

```c
malloc(sizeof(void*) * child->n_buffers);
```

in both cases, and `child` is updated just above with `n_buffers` being 2 and 3, respectively, as opposed to the 1 from the parent.   

Making the change makes the nags

      warning: array subscript 1 is outside array bounds of ‘void[8]'

      note: at offset 8 into object of size 8 allocated by ‘malloc’ 
      
go away.  I can send a PR if it helps as I have this updated in a fork.

### Component(s)

C, Documentation, Website

eddelbuettel:
### Rationale for this change

As described in #35617 the example in docs suffers from a likely copy-and-paste error as the `n_buffers` value references the _parent_ rather that the two children where this is used. `g++-12` spots this and bemoans that assigning to, respectively, `buffers[1]` and then `buffers[1]` and `buffers[2]` is out of bounds.

The example is corrected.

### Are these changes tested?

Yes, locally.  The warnings under `g++-12` go away.

Not in behavior but the documentation is improved.

* Closes: #35617

### DIFF
--- a/docs/source/format/CDataInterface.rst
+++ b/docs/source/format/CDataInterface.rst
@@ -907,7 +907,7 @@ transferring ownership to the consumer:
          // Bookkeeping
          .release = &release_malloced_array
       };
-      child->buffers = malloc(sizeof(void*) * array->n_buffers);
+      child->buffers = malloc(sizeof(void*) * child->n_buffers);
       child->buffers[0] = float32_nulls;
       child->buffers[1] = float32_data;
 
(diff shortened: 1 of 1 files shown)


kou (approved):
+1

Good catch!

PR #35612, opened 2023-05-16
GH-35594: [R][C++] Bump vendored date library

thisisnic:
There was an issue with tzdb 0.4.0 and the shipped arrow `tz.cpp`.

This PR bumps the vendored version of the date library to commit `cc4685a21e4a4fdae707ad1233c61bbaff241f93`.
* Closes: #35594

Issue #35600, opened 2023-05-15
Allow setting path to timezone db in pyarrow through python API


rohanjain101:
### Describe the enhancement requested

The timezone DB is required to be at %USERPROFILE%\Downloads\tzdata in pyarrow on windows. It would be nice to have a python API to change the path where this db should be found, I believe this is currently only possible through the C++ implementation using arrow:initialize.

It would also be nice if the timezone db was not required if tzdata/pytz are installed.

### Component(s)

Python

jorisvandenbossche:
> It would also be nice if the timezone db was not required if tzdata/pytz are installed.

If I remember correctly, the problem here is that this uses a different format than what the tz implementation we use (https://howardhinnant.github.io/date/tz.html) expects. 
See eg https://github.com/apache/arrow/issues/28868#issuecomment-1378032344 and https://github.com/apache/arrow/issues/31472 for previous discussion about this.

jorisvandenbossche:
And the R bindings already expose this, which essentially is a small wrapper around the setting a C++ option (`arrow::GlobalOptions.timezone_db_path`):

https://github.com/apache/arrow/blob/f6e447944f2a2ab108d5971daf351b7443bc96fb/r/src/config.cpp#L38-L47

So it should certainly be possible to add a similar function to pyarrow.

mattijn:
Thanks for all the work on arrow!

I'm on windows and when running the following:

```python
from datetime import datetime
import pyarrow.compute as pc

pc.strftime(datetime(2004, 8, 1))
```
I get the following error:
> ```cmd
> ArrowInvalid: Cannot locate timezone 'UTC': Timezone database not found at "D:\Users\Hoek\Downloads\tzdata"
> ```

Observe, that the error path refers to `D:`.

But in OP the following is mentioned:

> The timezone DB is required to be at %USERPROFILE%\Downloads\tzdata in pyarrow on windows.

But if I check:
```python
! echo %USERPROFILE%
```

> ```cmd
> C:\Users\Hoek
> ```
and
```python
from pathlib import Path
Path.home()
```

> ```cmd
> WindowsPath('C:/Users/Hoek')
> ```

It is both linking to `C:`. How you know it is on `D:`? `D:` is right though.. 
There is no `Downloads` folder for me on `C:\Users\Hoek` (therefor https://stackoverflow.com/a/74292266/2459096 errored)

I assume it is just early days and this user experience will improve soon. 
Again, thanks for all the hard work!

[...]

Issue #31472, opened 2022-03-28
[Python] Use tzdata timezone database on Windows


asfimport:
In ARROW-13168, we enabled setting the path of the text-based database engine at runtime. This allowed R to use the tzdb package for the timezone database, since it uses the text format.

However, it doesn't seem like tzdata Python package ships that text format. They do have [a "compact" text format](https://github.com/python/tzdata/blob/master/src/tzdata/zoneinfo/tzdata.zi), which _might_ be compatible with our vendored date library. Otherwise, we'd likely have to wait for binary format support in https://github.com/HowardHinnant/date/issues/564

**Reporter**: [Will Jones](https://issues.apache.org/jira/browse/ARROW-16054) / @wjones127
**Watchers**: [Rok Mihevc](https://issues.apache.org/jira/browse/ARROW-16054) / @rok
#### Related issues:
- https://github.com/apache/arrow/issues/28868 (is related to)

<sub>**Note**: *This issue was originally created as [ARROW-16054](https://issues.apache.org/jira/browse/ARROW-16054). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

AlenkaF:
PR to skip the failing tests on Windows that require a tz database: https://github.com/apache/arrow/pull/36996 (interim solution)

h-vetinari:
Are things still being backported to the branch for 13.0? If so, I think https://github.com/apache/arrow/pull/36996 would make sense, as it will cause test failures on windows otherwise.

CC @raulcd

raulcd:
> Are things still being backported to the branch for 13.0? If so, I think #36996 would make sense, as it will cause test failures on windows otherwise.
> 
> CC @raulcd

From my understanding the original error was introduced on https://github.com/apache/arrow/issues/36323 which is not part of 13.0.0

h-vetinari:
> From my understanding the original error was introduced on #36323 which is not part of 13.0.0

Ah OK great! Thanks for the info.

Background: I remember seeing a tzdata failure when I had built from source a while ago, but I had already [tested](https://github.com/conda-forge/arrow-cpp-feedstock/pull/1124) one of the rc's, and it passed as well on windows.

PR #35637, opened 2023-05-17
GH-35658: [Packaging] Sync conda recipes with feedstocks

### LINKED ISSUE #35658: [Packaging] Sync with conda-forge feedstocks

Regular updates to catch updates after 12.0 and new pinnings that arrived in the meantime

### Component(s)

Packaging

h-vetinari:
Corresponds to https://github.com/conda-forge/arrow-cpp-feedstock/pull/1053 for pyarrow (one failing test on our infra that needs debugging), as well as the state of the feedstock for r-arrow after https://github.com/conda-forge/r-arrow-feedstock/pull/65
* Closes: #35658


h-vetinari:
@github-actions crossbow submit -g conda

h-vetinari:
Interestingly, the failure of `test_total_bytes_allocated` on linux from https://github.com/conda-forge/arrow-cpp-feedstock/pull/1053 is not present here. But a weird error on windows:
```
E   pyarrow.lib.ArrowInvalid: Cannot locate timezone 'UTC': Timezone database not found at "C:\Users\VssAdministrator\Downloads\tzdata"
```

kou:
I don't know why we didn't need timezone database for Windows jobs before but we have a shell script to download timezone database:  https://github.com/apache/arrow/blob/main/ci/scripts/download_tz_database.sh

We're using the script on Windows too.

h-vetinari:
> I don't know why we didn't need timezone database for Windows jobs before but we have a shell script to download timezone database:

Sounds like we should just add a dependency on the [`tzdata`](https://github.com/conda-forge/tzdata-feedstock) conda-forge package?

assignUser:
> Sounds like we should just add a dependency on the [`tzdata`](https://github.com/conda-forge/tzdata-feedstock) conda-forge package?

That sound's good. Thanks for taking care of the sync work, much appreciated!

kou:
It may not work because it seems that `tzdata` doesn't include `WindowsZones.xml`.

h-vetinari:
> It may not work because it seems that `tzdata` doesn't include `WindowsZones.xml`.

Why does arrow depend on this in the first place? It's the first package I see that doesn't work with `tzdata`. Also, based on the error message, it seems to be looking in the wrong place? I.e. `C:\Users\VssAdministrator\Downloads\tzdata` instead of `%LIBRARY_PREFIX%` (aside from the fact that inndeed, our tzdata does not [contain](https://github.com/regro/libcfgraph/blob/master/artifacts/tzdata/conda-forge/noarch/tzdata-2023c-h71feb2d_0.json#L231-L835) `WindowsZones.xml`).

The tzdata stuff is an arrow 13 problem though. More urgently, in https://github.com/conda-forge/arrow-cpp-feedstock/pull/1053, I currently have a failing test `test_total_bytes_allocated` on arrow 12 (though not arrow 11 or before), but some commit between arrow 12 and current `main` seems to have fixed this - any ideas?

raulcd:
As a note, our nightly builds for wheels on Windows are also failing at the moment with the same error:
```
 E   pyarrow.lib.ArrowInvalid: Cannot locate timezone 'UTC': Timezone database not found at "C:\Users\ContainerAdministrator\Downloads\tzdata"
```
* [wheel-windows-cp310-amd64](https://github.com/ursacomputing/crossbow/actions/runs/4989689010/jobs/8933937653)
* [wheel-windows-cp311-amd64](https://github.com/ursacomputing/crossbow/actions/runs/4989688097/jobs/8933935592)
* [wheel-windows-cp37-amd64](https://github.com/ursacomputing/crossbow/actions/runs/4989683949/jobs/8933925508)
* [wheel-windows-cp38-amd64](https://github.com/ursacomputing/crossbow/actions/runs/4989686872/jobs/8933932531)
* [wheel-windows-cp39-amd64](https://github.com/ursacomputing/crossbow/actions/runs/4989690690/jobs/8933941809)
The first failure is from 10 days ago since this commit was introduced:
https://github.com/apache/arrow/commit/053b5eecb582a93bc92e6793768fc66d2c5d5046

kou:
Ah, sorry. I misread this problem.

It seems that `WindowsZones.xml` isn't related to this problem.

But tzdata path was hard-coded on Windows: https://github.com/apache/arrow/blob/main/cpp/src/arrow/vendored/datetime/tz.cpp#L265

So our code will not find tzdata installed by conda.

Could you open a new issue for tzdata related problem? We can work on this as a separated issue.

raulcd:
The datetime vendored library seems to have had a change lately: https://github.com/apache/arrow/pull/35612/files
And is quite important for the R release to be packed for 12.0.1. I am wondering whether this issue will have to be fixed for the 12.0.1 release too.

h-vetinari:
> 


> The tzdata stuff is an arrow 13 problem though. More urgently, in [conda-forge/arrow-cpp-feedstock#1053](https://github.com/conda-forge/arrow-cpp-feedstock/pull/1053), I currently have a failing test `test_total_bytes_allocated` on arrow 12 (though not arrow 11 or before), but some commit between arrow 12 and current `main` seems to have fixed this - any ideas?

As I mentioned, the tzdata thing is much less urgent for us in conda-forge than the issue of that failing tests on the feedstock. I now opened a separate issue so this gets (hopefully...) more traction: #35728

jorisvandenbossche:
> Why does arrow depend on this in the first place? It's the first package I see that doesn't work with `tzdata`.

Arrow unfortunately can't work with the tz database from `tzdata`, because the underlying date library we use doesn't support the binary format, see https://github.com/apache/arrow/issues/31472 and linked issues.

In other places we skip tests that requite tz database access on Windows, so that should be done for the new scalar test as well, I assume.

jorisvandenbossche:
I opened https://github.com/apache/arrow/pull/35735 to skip the test requiring a tz database on windows.

h-vetinari:
> Arrow unfortunately can't work with the tz database from `tzdata`, because the underlying date library we use doesn't support the binary format, see #31472 and linked issues.

I don't mind shipping arrow's vendored `WindowsZones.xml` in a pyarrow-specific location (i.e. site-packages). I think it would be better to do that than just fail? timezone-operations are pretty common (unless I misunderstand something), so not supporting that for the conda-forge builds would suck IMO

jorisvandenbossche:
>  so not supporting that for the conda-forge builds would suck IMO

Yes, but to be clear, we currently don't support it for any form of distribution (also not wheels), the user is required to download it manually. So while I also certainly would like to see this solved, it's not something where conda-forge is worse.

Although it might be easier to solve for conda-forge:

> I don't mind shipping arrow's vendored `WindowsZones.xml` in a pyarrow-specific location (i.e. site-packages/pyarrow; we can patch pyarrow to point to the right location).

That should be possible yes (I think it doesn't even need to be in a pyarrow-specific location, it could also be in a "proper" location for data). Currently this would require a patch in the C++ code, with https://github.com/apache/arrow/issues/35600 it should be an easier patch in the pyarrow's `__init__.py`

One issue with vendoring is that this doesn't make it easy to update the version / ensure you always have a recent tz database, also when installing an somewhat older pyarrow version.

h-vetinari:
> One issue with vendoring is that this doesn't make it easy to update the version / ensure you always have a recent tz database, also when installing an somewhat older pyarrow version.

Absolutely, that's why it would be nice if this could be handled through a "proper" dependency like `tzdata`. I haven't dived deep into https://github.com/apache/arrow/issues/31472, but I don't understand why the format that `tzdata` has couldn't be transformed or otherwise made to work for arrow?

jorisvandenbossche:
That's probably technically possible to implement such transformation, but someone still needs to do the work for that. A potential easier path forward is to update date.h to work with the compiled database on Windows (https://github.com/HowardHinnant/date/issues/564)

h-vetinari:
@github-actions crossbow submit -g conda

h-vetinari:
@github-actions crossbow submit -g conda

h-vetinari:
@kou @pitrou @raulcd @assignUser @jorisvandenbossche 
This is green now and matches the state of the feedstocks. There are some more minor improvements on the way (a few but not all) of the test skips can be removed, but I can also put those in the next sync PR after arrow 13 is officially out.

assignUser (approved):
Great work, thanks @h-vetinari ! I will merge in a day or so if there are not objections

dev/tasks/conda-recipes/arrow-cpp/build-arrow.sh
```diff
@@ -30,27 +30,21 @@ fi
...
+    # point to a usable protoc/grpc_cpp_plugin if we're cross-compiling
+    EXTRA_CMAKE_ARGS="${EXTRA_CMAKE_ARGS} -DProtobuf_PROTOC_EXECUTABLE=$BUILD_PREFIX/bin/protoc"
+    if [[ ! -f ${BUILD_PREFIX}/bin/${CONDA_TOOLCHAIN_HOST}-clang ]]; then
+        ln -sf ${BUILD_PREFIX}/bin/clang ${BUILD_PREFIX}/bin/${CONDA_TOOLCHAIN_HOST}-clang
+    fi
+    EXTRA_CMAKE_ARGS="${EXTRA_CMAKE_ARGS} -DCLANG_EXECUTABLE=${BUILD_PREFIX}/bin/${CONDA_TOOLCHAIN_HOST}-clang"
+    EXTRA_CMAKE_ARGS="${EXTRA_CMAKE_ARGS} -DLLVM_LINK_EXECUTABLE=${BUILD_PREFIX}/bin/llvm-link"
     sed -ie "s;protoc-gen-grpc.*$;protoc-gen-grpc=${BUILD_PREFIX}/bin/grpc_cpp_plugin\";g" ../src/arrow/flight/CMakeLists.txt
-    sed -ie 's;"--with-jemalloc-prefix\=je_arrow_";"--with-jemalloc-prefix\=je_arrow_" "--with-lg-page\=14";g' ../cmake_modules/ThirdpartyToolchain.cmake
+    sed -ie 's;"--with-jemalloc-prefix\=je_arrow_";"--with-jemalloc-prefix\=je_arrow_" "--with-lg-page\=16";g' ../cmake_modules/ThirdpartyToolchain.cmake
```
kou, line 47: We can use `-DARROW_JEMALLOC_LG_PAGE=16` instead of rewriting `ThirdpartyToolchain.cmake`.

kou (approved):
+1

raulcd (approved):

dev/tasks/conda-recipes/arrow-cpp/build-arrow.sh
h-vetinari: That sounds great, I'll change that on the feedstock and in the next sync

PR #35654, opened 2023-05-17
GH-35652: [Go][Compute] Allow executing Substrait Expressions using Go Compute

### LINKED ISSUE #35652: [Go][Compute] Allow Executing Substrait Expressions through Go Arrow Compute

Rather than using its own Expression objects and interfaces, let's have Arrow Go just leverage Substrait directly and provide the ability to execute scalar substrait expressions. There isn't a facility yet to perform full evaluation of projections and relations or full plans yet, but we can execute individual expression objects to provide a way to describe and specify compute expressions which are more complex than single operations, leveraging substrait-go to define them.

### Component(s)

Go

zeroshade:
### Rationale for this change
Providing the ability to execute more complex expressions than single operations by leveraging Substrait's expression objects and deprecating the existing separate Expression interfaces in Go Arrow compute. This provides a quick integration with Substrait Expressions and ExtendedExpressions to start building more integrations.

This PR provides:

* an extension registry for Go arrow to provide mappings between Arrow and substrait for functions and for types along with other custom mappings if necessary. 
* Facilities to convert between Arrow data types and Substrait types
* Functions to evaluate Substrait expression objects with Arrow data as the input
* Functions to evaluate Substrait field references against Arrow data and arrow schemas

### Are these changes tested?
Yes, unit tests are included.

Existing `compute.Expression` and its friends are being marked as deprecated.

* Closes: #35652


go/arrow/internal/flight_integration/cmd/arrow-flight-integration-server/__debug_bin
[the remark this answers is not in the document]
zeroshade: -_- i don't even know how that got checked in, i'll remove and add it to the git ignore

go/arrow/array/util_test.go
```diff
@@ -522,3 +522,29 @@ func TestRecordBuilderUnmarshalJSONExtraFields(t *testing.T) {
 
 	assert.Truef(t, array.RecordEqual(rec1, rec2), "expected: %s\nactual: %s", rec1, rec2)
 }
+
+func TestJSON(t *testing.T) {
```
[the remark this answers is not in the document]
zeroshade: honestly? I don't know. It must have gotten caught in a git stash pop that i did at some point. I'll take this out.

felipecrv:
> CC @benibus @felipecrv if you two feel like taking a look / reviewing on this despite your lack of familiarity with Go

Yes, but this is a big one, so next week.

go/arrow/compute/exprs/exec.go
```diff
@@ -0,0 +1,631 @@
...
+		return compute.NewDatum(scalar.NewFloat32Scalar(v.Value)), nil
+	case *expr.PrimitiveLiteral[float64]:
+		return compute.NewDatum(scalar.NewFloat64Scalar(v.Value)), nil
+	case *expr.PrimitiveLiteral[string]:
+		return compute.NewDatum(scalar.NewStringScalar(v.Value)), nil
+	case *expr.PrimitiveLiteral[types.Timestamp]:
+		return compute.NewDatum(scalar.NewTimestampScalar(arrow.Timestamp(v.Value), &arrow.TimestampType{Unit: arrow.Microsecond})), nil
+	case *expr.PrimitiveLiteral[types.TimestampTz]:
+		return compute.NewDatum(scalar.NewTimestampScalar(arrow.Timestamp(v.Value),
+			&arrow.TimestampType{Unit: arrow.Microsecond, TimeZone: TimestampTzTimezone})), nil
```
westonpace, line 204: What is TimestampTzTimezone?

go/arrow/compute/exprs/exec.go
```diff
@@ -0,0 +1,631 @@
...
+		kvArr, err := array.NewStructArray([]arrow.Array{keyArr, valArr}, []string{"key", "value"})
+		if err != nil {
+			return nil, err
+		}
+		defer kvArr.Release()
+
+		return compute.NewDatumWithoutOwning(scalar.NewMapScalar(kvArr)), nil
+	case *expr.StructLiteral:
+		fields := make([]scalar.Scalar, len(v.Value))
+		names := make([]string, len(v.Value))
```
westonpace, line 320: I didn't think the struct literal had names?  Or maybe you're making empty strings for names here?

go/arrow/compute/exprs/exec.go
```diff
@@ -0,0 +1,631 @@
...
+				// reverse the bytes
+				for i := len(data)/2 - 1; i >= 0; i-- {
+					opp := len(data) - 1 - i
+					data[i], data[opp] = data[opp], data[i]
+				}
+			}
+
+			return compute.NewDatum(scalar.NewDecimal128Scalar(val,
+				&arrow.Decimal128Type{Precision: v.Precision, Scale: v.Scale})), nil
+		case *types.UserDefinedLiteral: // not yet implemented
```
westonpace, line 345: Return an error here?

go/arrow/compute/exprs/exec.go
```diff
@@ -0,0 +1,631 @@
...
+		return nil, err
+	}
+
+	var param compute.Datum
+	if sref, ok := ref.(*expr.StructFieldRef); ok {
+		if sref.Field < 0 || sref.Field >= int32(len(input.Values)) {
+			return nil, arrow.ErrInvalid
+		}
+		param = input.Values[sref.Field]
+		ref = ref.GetChild()
```
westonpace, line 459: Is this a while loop?  A `StructField` reference is a recursively nested structure (e.g. 0.3.1) to mean "grab the 0th field, then grab the 3rd field of that, then grab the 1th field of that).

go/arrow/compute/exprs/exec.go
```diff
@@ -0,0 +1,631 @@
...
+	if !arrow.TypeEqual(out.(compute.ArrayLikeDatum).Type(), expectedType) {
+		return nil, fmt.Errorf("%w: referenced field %s was %s, but should have been %s",
+			arrow.ErrInvalid, ref, out.(compute.ArrayLikeDatum).Type(), expectedType)
+	}
+
+	return out, nil
+}
+
+func executeScalarBatch(ctx context.Context, input compute.ExecBatch, exp expr.Expression, ext ExtensionIDSet) (compute.Datum, error) {
+	if !exp.IsScalar() {
```
westonpace, line 477: What's a non-scalar expression?

go/arrow/compute/exprs/exec.go
```diff
@@ -0,0 +1,631 @@
...
+				args[i] = compute.NewDatum(scalar.NewStringScalar(string(v)))
+			case expr.Expression:
+				args[i], err = executeScalarBatch(ctx, input, v, ext)
+				if err != nil {
+					return nil, err
+				}
+				defer args[i].Release()
+
+				if args[i].Kind() != compute.KindScalar {
+					allScalar = false
```
westonpace, line 532: Why do you care about `allScalar`?

go/arrow/compute/exprs/exec.go
```diff
@@ -0,0 +1,631 @@
...
+				newArgs[i], err = compute.CastDatum(ctx, arg, compute.SafeCastOptions(argTypes[i]))
+				if err != nil {
+					return nil, err
+				}
+				defer newArgs[i].Release()
+			}
+		}
+		if newArgs != nil {
+			args = newArgs
+		}
```
westonpace, line 583: Technically this kind of implicit casting shouldn't be needed.  Substrait plans should always represent casts with explicit casts.

westonpace (commented):
I took a look at the expression evaluation.  I'll try and take a look at the function decoding later.  There's a lot of good stuff here!

go/arrow/compute/arithmetic.go
```diff
@@ -627,6 +627,8 @@ func RegisterScalarArithmetic(reg FunctionRegistry) {
 	}{
 		{"sub_unchecked", kernels.OpSub, decPromoteAdd, subUncheckedDoc},
 		{"sub", kernels.OpSubChecked, decPromoteAdd, subDoc},
+		{"subtract_unchecked", kernels.OpSub, decPromoteAdd, subUncheckedDoc},
+		{"subtract", kernels.OpSubChecked, decPromoteAdd, subDoc},
```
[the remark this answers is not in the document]
zeroshade: just made things easier since for all of the *other* functions I was able to have a simple loop since the names matched and I didn't want to have a separate case for only subtract. It's not strictly needed, it just made some code cleaner.

go/arrow/compute/exprs/exec.go
```diff
@@ -0,0 +1,631 @@
...
+		defer arg.Release()
+
+		dt, _, err := FromSubstraitType(e.Type, ext)
+		if err != nil {
+			return nil, fmt.Errorf("%w: could not determine type for cast", err)
+		}
+
+		var opts *compute.CastOptions
+		switch e.FailureBehavior {
+		case types.BehaviorThrowException:
```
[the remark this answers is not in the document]
zeroshade: ah, so as things currently stand there's no option for checking for overflows in substrait casting?

go/arrow/compute/exprs/exec.go
zeroshade: It's a constant defined as "UTC", defined at the top of `types.go`, matching the inline function `TimestampTzTimezoneString` in the c++ (`cpp/src/arrow/engine/substrait/type_internal.h`)

go/arrow/compute/exprs/exec.go
zeroshade: correct, it just uses empty strings for the names.

go/arrow/compute/exprs/exec.go
zeroshade: it'll hit the case of returning `arrow.ErrNotImplemented` which is at the bottom of the function for anything which doesn't return during the switch.

go/arrow/compute/exprs/exec.go
zeroshade: Currently, that would consist of aggregate / window functions I believe (or anything containing one as an argument). Equivalent to Acero's definition of `scalar` vs `vector` functions, i.e. evaluated element-wise vs array-wise.

go/arrow/compute/exprs/exec.go
zeroshade: In the case where an expression is a scalar function whose arguments consist *solely* of enum values, we don't want the exec batch length for the input to that function to reflect the top level input. It's a special case where the input length for that evaluation is 1 since there's only the one value (the enum).

go/arrow/compute/exprs/exec.go
zeroshade: Good point! this was carried over from my original work where I was evaluating general expressions before I converted this to evaluating substrait. I'll remove this to simplify it

lidavidm (approved):
I'm good with this, though admittedly I'm not up to date on Substrait specifics.

go/arrow/compute/exprs/exec.go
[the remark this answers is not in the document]
westonpace: Correct.

westonpace (approved):
I added a bit more rambling but overall this seems good to me.

go/arrow/compute/exprs/exec.go
zeroshade: Yea, that's how I understood things. So the `SUM(y)` should return false for the `IsScalar` check based on the current stuff in substrait-go, WindowFunctions return false for `IsScalar` currently.

go/arrow/compute/exprs/exec.go
zeroshade: The only case I can think of it being a valid function would be a generative function, i.e. a function that generates a result set itself like `SELECT RAND()` or whatnot, but that takes some enum arguments.

go/arrow/compute/exprs/builders.go
```diff
@@ -0,0 +1,445 @@
...
+	for len(dotpath) > 0 {
+		subscript := dotpath[0]
+		dotpath = dotpath[1:]
+		switch subscript {
+		case '.':
+			// next element is a name
+			n := parseName()
+			switch ct := curType.(type) {
+			case *arrow.StructType:
+				idx, found := ct.FieldIdx(n)
```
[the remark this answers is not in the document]
zeroshade: It's a map. The `arrow.StructType` in Go currently doesn't support multiple fields with the same name as it maintains a `map[string]int` to map field names to indices. `FieldIdx` just does a lookup in the map.

go/arrow/compute/exprs/builders.go
```diff
@@ -0,0 +1,445 @@
...
+				case arrow.INT16:
+					keyLiteral = expr.NewPrimitiveLiteral(int16(idx), false)
+				case arrow.INT32:
+					keyLiteral = expr.NewPrimitiveLiteral(int32(idx), false)
+				case arrow.INT64:
+					keyLiteral = expr.NewPrimitiveLiteral(int64(idx), false)
+				case arrow.FLOAT32:
+					keyLiteral = expr.NewPrimitiveLiteral(float32(idx), false)
+				case arrow.FLOAT64:
+					keyLiteral = expr.NewPrimitiveLiteral(float64(idx), false)
```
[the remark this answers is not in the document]
zeroshade: so, there's nothing in the spec that says a float *can't* be a map key which is why I didn't mark it as unsupported. But you make a good point about the call to `atoi`. I'll change the float cases to unsupported for now and if anyone has an issue I can address it then.

go/arrow/compute/exprs/builders.go
```diff
@@ -0,0 +1,445 @@
...
+		for {
+			idx := strings.IndexAny(dotpath, `\[.`)
+			if idx == -1 {
+				name += dotpath
+				dotpath = ""
+				break
+			}
+
+			if dotpath[idx] != '\\' {
+				// subscript for a new field ref
```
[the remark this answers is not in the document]
zeroshade: On line 120, the `name += dotpath[:idx] + string(dotpath[idx+1])` will end up adding the `]` to the `name`, and then we consume the `]` on the next line by doing `dotpath = dotpath[idx+2:]`. For this function, the `parseName` there isn't necessarily a requirement that there be a matching `]`, since you only end up with a `[` here if it's escaped with `\\`.

For the index case of something like `[2]`, the `]` is consumed by line 199 when we do `dotpath = dotpath[subend+1:]` to skip past the `]` (as `dotpath[subend]` is the closing `]`).

Issue #35598, opened 2023-05-15
[MATLAB] Add a public `Valid` property to to the MATLAB `arrow.array.<Array>` classes to query Null values (i.e. validity bitmap support)


kevingurney:
### Describe the enhancement requested

Currently, the `arrow.array.<Array>` classes do not support querying the Null values (i.e. validity bitmap) on an Arrow array. Support for encoding Null values is an important part of the Arrow memory format, so the MATLAB Interface to Arrow should support it.

There are likely multiple different APIs that the MATLAB interface should have to support Null values robustly. However, to focus on incremental delivery, we can start by adding a public `Valid` property to the `arrow.array.<Array>` classes, which would return a `logical` array of null values in the given array.

### Component(s)

MATLAB

kevingurney:
take

kou:
Issue resolved by pull request 35655
https://github.com/apache/arrow/pull/35655

PR #35655, opened 2023-05-17
GH-35598: [MATLAB] Add a public `Valid` property to to the `MATLAB arrow.array.<Array>` classes to query Null values (i.e. validity bitmap support)

kevingurney:
### Rationale for this change

Currently, the `arrow.array.<Array>` classes do not support querying the Null values (i.e. validity bitmap) on an Arrow array. Support for encoding Null values is an important part of the Arrow memory format, so the MATLAB Interface to Arrow should support it.

There are likely multiple different APIs that the MATLAB interface should have to support Null values robustly. However, to focus on incremental delivery, we can start by adding a public `Valid` property to the `arrow.array.<Array>` classes, which would return a `logical` array of null values in the given array.

1. Added a new public property `Valid` to the `arrow.array.Array` superclass.
2. Implemented basic null value handling for `arrow.array.Float64Array` (i.e. treat `NaN` values in the input MATLAB array as null values in the corresponding `arrow.array.Float64Array`).
3. Implement null value substitution (i.e. substitute null values with `NaN`) for `Float64Array` in `toMATLAB` and `double` conversion methods.

Example of creating an `arrow.array.Float64Array` from a MATLAB `double` array containing `NaN` values:

```matlab
>> matlabArray = [1, 2, NaN, 4, NaN]'

matlabArray =

     1
     2
   NaN
     4
   NaN

>> arrowArray = arrow.array.Float64Array(matlabArray)

arrowArray = 

[
  1,
  2,
  null,
  4,
  null
]

>> arrowArray.Valid

ans =

  5×1 logical array

   1
   1
   0
   1
   0

>> all(~isnan(matlabArray) == arrowArray.Valid)

ans =

  logical

   1
```

### Are these changes tested?

Yes, we have added the following test points for the `Valid` property of `arrow.array.Float64Array`:

1. `ValidBasic`
2. `ValidNoNulls`
4. `ValidAllNulls`
5. `ValidEmpty`

Yes.

There is now a public property `Valid` on the arrow.array.Float64Array` class which is a MATLAB `logical` array encoding the null values in the underlying Arrow array, where `true` indicates an element is valid (i.e. not null) and `false` indicates that an element is invalid (i.e. null).

### Future Directions

1. Implement more null value related methods like `isvalid`, `isnull`, `packagedValidityBitmap`, etc.
2. Add null value (i.e. `Valid` property) support to the rest of the `arrow.array.Array` subclasses.

### Notes

1. Thank you to @sgilmore10 for your help with this pull request!


assignUser:
I don't feel comfortable "officially" reviewing these changes but the reasoning is sound and the quality of the PR description is great. It has been nice to see the constant stream of recent contributions to the matlab component! :tada:

matlab/src/cpp/arrow/matlab/array/proxy/array.cc
```diff
@@ -40,4 +43,25 @@ namespace arrow::matlab::array::proxy {
...
+        size_t array_length = static_cast<size_t>(array->length());
+        
+        // If the Arrow array has no null values, then return a MATLAB
+        // logical array that is all "true" for the validity bitmap.
+        if (array->null_count() == 0) {
+            ::matlab::data::ArrayFactory factory;
+            auto validity_buffer = factory.createBuffer<bool>(array_length);
+            auto validity_buffer_ptr = validity_buffer.get();
+            std::fill(validity_buffer_ptr, validity_buffer_ptr + array_length, true);
+            ::matlab::data::TypedArray<bool> valid_elements_mda = factory.createArrayFromBuffer<bool>({array_length, 1}, std::move(validity_buffer));
```
kou, line 57: Can we use `auto` here?

```suggestion
            auto valid_elements_mda = factory.createArrayFromBuffer<bool>({array_length, 1}, std::move(validity_buffer));
```

matlab/src/cpp/arrow/matlab/array/proxy/array.cc
```diff
@@ -40,4 +43,25 @@ namespace arrow::matlab::array::proxy {
...
+            context.outputs[0] = valid_elements_mda;
+            return;
+        }
+
+        auto validity_bitmap = array->null_bitmap();
+        ::matlab::data::TypedArray<bool> valid_elements_mda = arrow::matlab::bit::bitUnpackArrowBuffer(validity_bitmap, array_length);
```
kou, line 63: Ditto.

matlab/src/cpp/arrow/matlab/bit/bit_pack_matlab_logical_array.cc
```diff
@@ -0,0 +1,64 @@
...
+        // Call into a loop-unrolled Arrow utility for better performance when bit-packing.
+        auto generator = [&]() -> bool { return *(unpacked_bool_data_iterator++); };
+        const int64_t start_offset = 0;
+
+        auto mutable_data = packed_validity_bitmap_buffer->mutable_data();
+
+        arrow::internal::GenerateBitsUnrolled(mutable_data, start_offset, unpacked_buffer_length, generator);
+
+        auto buffer_result = std::static_pointer_cast<arrow::Buffer>(packed_validity_bitmap_buffer);
+        return arrow::Result<std::shared_ptr<arrow::Buffer>>{buffer_result};
```
kou, line 61: Can we simplify this?

* `std::shared_ptr<arrow::ResizableBuffer> packed_validity_bitmap_buffer` -> `auto packed_validity_bitmap_buffer`
*  `auto buffer_result = std::static_pointer_cast<arrow::Buffer>(packed_validity_bitmap_buffer); return arrow::Result<std::shared_ptr<arrow::Buffer>>{buffer_result};` -> `return packed_validity_bitmap_buffer`

```suggestion
        ARROW_ASSIGN_OR_RAISE(auto packed_validity_bitmap_buffer,  arrow::AllocateResizableBuffer(packed_buffer_length));

        // Get pointers to the internal uint8_t arrays behind arrow::Buffer and mxArray
        // Get raw bool array pointer from MATLAB logical array.
        // Get an iterator to the raw bool data behind the MATLAB logical array.
        auto unpacked_bool_data_iterator = matlab_logical_array.cbegin();

        // Iterate over the mxLogical array and write bit-packed bools to the arrow::Buffer.
        // Call into a loop-unrolled Arrow utility for better performance when bit-packing.
        auto generator = [&]() -> bool { return *(unpacked_bool_data_iterator++); };
        const int64_t start_offset = 0;

        auto mutable_data = packed_validity_bitmap_buffer->mutable_data();

        arrow::internal::GenerateBitsUnrolled(mutable_data, start_offset, unpacked_buffer_length, generator);

        return packed_validity_bitmap_buffer;
```

matlab/src/cpp/arrow/matlab/array/proxy/numeric_array.h
```diff
@@ -68,12 +81,11 @@ class NumericArray : public arrow::matlab::array::proxy::Array {
                 auto data_buffer = std::make_shared<arrow::Buffer>(reinterpret_cast<const uint8_t*>(dt),
                                                               sizeof(CType) * numeric_mda.getNumberOfElements());
 
-                // TODO: Implement null support
-                std::shared_ptr<arrow::Buffer> null_buffer = nullptr;
+                // Pack the validity bitmap values.
+                auto packed_validity_bitmap = has_validity_bitmap ? arrow::matlab::bit::bitPackMatlabLogicalArray(constructor_arguments[2]).ValueOrDie() : nullptr;
```
kou, line 85: In general, we should not call `ValueOrDie()` without checking `ok()` in library. If `!ok()`, an user program is aborted.

Can we raise an exception instead of abort on error?
(This may be a follow-up task.)

matlab/src/cpp/arrow/matlab/array/proxy/numeric_array.h
```diff
@@ -26,11 +26,19 @@
...
+#include "arrow/matlab/bit/bit_pack_matlab_logical_array.h"
 
 #include "libmexclass/proxy/Proxy.h"
 
 namespace arrow::matlab::array::proxy {
 
+namespace {
+const uint8_t* getUnpackedValidityBitmap(const ::matlab::data::TypedArray<bool>& valid_elements) {
+    const auto valid_elements_iterator(valid_elements.cbegin());
+    return reinterpret_cast<const uint8_t*>(valid_elements_iterator.operator->());
```
kou, line 38: Wow!

`::matlab::data::TypedArray` doesn't provide a method that returns it's raw data, right?

kevingurney: Unfortunately, the current design of the MATLAB Data APIs doesn't provide an easier way of extracting the raw data without making a copy. The `->operator()` syntax is the only way to achieve this at the moment.

We know this isn't a great solution, and we have reached out to the relevant product teams at MathWorks to see if we can come up with a better solution for a future MATLAB release.

matlab/src/cpp/arrow/matlab/array/proxy/numeric_array.h
kevingurney: That's a very good point!

We need to add a bit more infrastructure to mathworks/libmexclass to make exception handling more straightforward. So, it might be easier if we follow up with a separate pull request to handle this after updating libmexclass.

We have been exploring a few different approaches to error handling in libmexclass, but we haven't decided which one makes the most sense yet. We know this is something that needs to be prioritized in the near term.

I've captured this in https://github.com/apache/arrow/issues/35768.

matlab/src/cpp/arrow/matlab/array/proxy/array.cc
kevingurney: Fixed.

matlab/src/cpp/arrow/matlab/array/proxy/array.cc
kevingurney: Fixed.

matlab/src/cpp/arrow/matlab/array/proxy/array.cc
```diff
@@ -40,4 +43,25 @@ namespace arrow::matlab::array::proxy {
         auto length_mda = factory.createScalar(array->length());
         context.outputs[0] = length_mda;
     }
+
+    void Array::valid(libmexclass::proxy::method::Context& context) {
+        size_t array_length = static_cast<size_t>(array->length());
```
kou, line 48:
```suggestion
        auto array_length = static_cast<size_t>(array->length());
```

matlab/src/cpp/arrow/matlab/array/proxy/numeric_array.h
kou: I see.

matlab/src/cpp/arrow/matlab/bit/bit_unpack_arrow_buffer.cc
```diff
@@ -0,0 +1,41 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+#include "arrow/matlab/bit/bit_unpack_arrow_buffer.h"
+
+#include "arrow/util/bitmap_visit.h"
+
+namespace arrow::matlab::bit {
+    ::matlab::data::TypedArray<bool> bitUnpackArrowBuffer(const std::shared_ptr<arrow::Buffer>& packed_buffer, int64_t length) {
+        const uint8_t* packed_buffer_ptr = packed_buffer->data();
```
kou, line 24:
```suggestion
        const auto packed_buffer_ptr = packed_buffer->data();
```

matlab/src/cpp/arrow/matlab/bit/bit_unpack_arrow_buffer.cc
```diff
@@ -0,0 +1,41 @@
...
+
+        ::matlab::data::ArrayFactory factory;
+        
+        const size_t array_length = static_cast<size_t>(length);
```
kou, line 28:
```suggestion
        const auto array_length = static_cast<size_t>(length);
```

kou (approved):
+1

PR #35656, opened 2023-05-17
GH-33321: [Python] Support converting to non-nano datetime64 for pandas >= 2.0

### LINKED ISSUE #33321: [Python] Support converting to non-nano datetime64 for pandas >= 2.0
Pandas is adding capabilities to store non-nanosecond datetime64 data. At the moment, we however always do convert to nanosecond, regardless of the timestamp resolution of the arrow table (and regardless of the pandas metadata).

Using the development version of pandas:

```Java

In [1]: df = pd.DataFrame({"col": np.arange("2012-01-01", 10, dtype="datetime64[s]")})

In [2]: df.dtypes
Out[2]: 
col    datetime64[s]
dtype: object

In [3]: table = pa.table(df)

In [4]: table.schema
Out[4]: 
col: timestamp[s]
-- schema metadata --
pandas: '{"index_columns": [{"kind": "range", "name": null, "start": 0, "' + 423

In [6]: table.to_pandas().dtypes
Out[6]: 
col    datetime64[ns]
dtype: object
```

This is because we have a `coerce_temporal_nanoseconds` conversion option which we hardcode to True (for top-level columns, we hardcode it to False for nested data). 

When users have pandas >= 2, we should support converting with preserving the resolution. We should certainly do so if the pandas metadata indicates which resolution was originally used (to ensure correct roundtrip). 
We _could_ (and at some point also _should_) also do that by default if there is no pandas metadata (but maybe only later depending on how stable this new feature is in pandas, as it is potentially a breaking change for our users if you use eg pyarrow to read a parquet file).

**Reporter**: [Joris Van den Bossche](https://issues.apache.org/jira/browse/ARROW-18124) / @jorisvandenbossche
#### Related issues:
- [[Python][CI] Build with pandas master/nightly failure related to timedelta64 resolution](https://github.com/apache/arrow/issues/33286) (is related to)

<sub>**Note**: *This issue was originally created as [ARROW-18124](https://issues.apache.org/jira/browse/ARROW-18124). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

danepitkin:
Do not coerce temporal types to nanosecond when pandas >= 2.0 is imported, since pandas now supports s/ms/us time units.

This PR adds support for the following Arrow -> Pandas conversions, which previously all defaulted to `datetime64[ns]` or `datetime64[ns, <TZ>]`:
```
date32 -> datetime64[ms]
date64 -> datetime64[ms]
datetime64[s] -> datetime64[s]
datetime64[ms] -> datetime64[ms]
datetime64[us] -> datetime64[us]
datetime64[s, <TZ>] -> datetime64[s, <TZ>]
datetime64[ms, <TZ>] -> datetime64[ms, <TZ>]
datetime64[us, <TZ>] -> datetime64[us, <TZ>]
```
### Rationale for this change

Pandas 2.0 introduces proper support for temporal types.

### Are these changes tested?

Yes. Pytests added and updated.

Yes, arrow-to-pandas default conversion behavior will change when users have pandas >= 2.0, but a legacy option is exposed to provide backwards compatibility.

* Closes: #33321


danepitkin:
I'm looking for early feedback to see if this is the right approach. There are many test cases that will need updating, but I didn't want to tackle them yet in case we take a different approach.

jorisvandenbossche (commented):
Yes, that approach looks good, and is actually simpler than I thought it would be since we already control this with the single option switch (for the code, the tests will indeed get a bit messier).

I think one question is whether we want to make that option public through the `to_pandas` API, so people could still override it to get nanoseconds if they want (to get back the pre-pandas-2.0 behaviour).

danepitkin:
> Yes, that approach looks good, and is actually simpler than I thought it would be since we already control this with the single option switch (for the code, the tests will indeed get a bit messier).
> 
> I think one question is whether we want to make that option public through the `to_pandas` API, so people could still override it to get nanoseconds if they want (to get back the pre-pandas-2.0 behaviour).

I'll expose this! I agree we'd its best to allow continued use of the legacy behavior for awhile.

python/pyarrow/types.pxi
```diff
@@ -40,10 +42,20 @@ cdef dict _pandas_type_map = {
     _Type_HALF_FLOAT: np.float16,
     _Type_FLOAT: np.float32,
     _Type_DOUBLE: np.float64,
-    _Type_DATE32: np.dtype('datetime64[ns]'),
-    _Type_DATE64: np.dtype('datetime64[ns]'),
-    _Type_TIMESTAMP: np.dtype('datetime64[ns]'),
-    _Type_DURATION: np.dtype('timedelta64[ns]'),
+    _Type_DATE32: np.dtype('datetime64[D]'),
```
jorisvandenbossche, line 45: pandas only supports the range of second to nanoseconds, so for dates we should maybe still default to `datetime64[s]`? (otherwise I assume this conversion would happen anyway on the pandas side)

python/pyarrow/types.pxi
```diff
@@ -1122,6 +1150,19 @@ cdef class DurationType(DataType):
...
+        Return the equivalent NumPy / Pandas dtype.
+
+        Examples
+        --------
+        >>> import pyarrow as pa
+        >>> d = pa.duration('ms')
+        >>> d.to_pandas_dtype()
+        timedelta64[ms]
+        """
+        return _get_pandas_type(_Type_TIMESTAMP, self.unit)
```
jorisvandenbossche, line 1164: Should this use _Type_DURATION instead? 
(which also raises the question: is this tested? Although I suppose you haven't yet bothered with updating all the tests)

python/pyarrow/array.pxi
```diff
@@ -1674,8 +1682,11 @@ cdef _array_like_to_pandas(obj, options, types_mapper):
     original_type = obj.type
     name = obj._name
 
-    # ARROW-3789(wesm): Convert date/timestamp types to datetime64[ns]
-    c_options.coerce_temporal_nanoseconds = True
+    # ARROW-33321 reenables support for date/timestamp conversion in pandas >= 2.0
+    from pyarrow.vendored.version import Version
+    if pandas_api.loose_version < Version('2.0.0'):
+        # ARROW-3789(wesm): Convert date/timestamp types to datetime64[ns]
+        c_options.coerce_temporal_nanoseconds = True
```
jorisvandenbossche, line 1689: Should this only be set of the argument was _not_ specified by the user? Or are we fine with forcing this in case of pandas<2.0, since that's what we did in the past anyway, and is also the only useful behaviour (letting the user specify `coerce_temporal_nanoseconds=False` in a conversion to pandas with pandas<2.0 doesn't make much sense, I suppose, since then pandas will convert it to nanoseconds anyhow, and the user never sees the non-nanosecond data). So after writing this: I assume the above is fine ;)

python/pyarrow/array.pxi
```diff
@@ -775,6 +776,11 @@ cdef class _PandasConvertible(_Weakrefable):
             expected to return a pandas ExtensionDtype or ``None`` if the
             default conversion should be used for that type. If you have
             a dictionary mapping, you can pass ``dict.get`` as function.
+        coerce_temporal_nanoseconds : bool, default False
+            A legacy option to coerce date32, date64, datetime, and timestamp
+            types to use nanoseconds when converting to pandas. This was the
+            default behavior in pandas version 1.x. In pandas version 2.0,
+            non-nanosecond time units are now supported.
```
jorisvandenbossche, line 783: We should probably clarify that this keyword is ignored for conversions to pandas with pandas<2.0, and force set to True (regardless of what the user would specify here)

python/pyarrow/types.pxi
danepitkin: Great catch! You're correct that this isn't tested yet.

python/pyarrow/types.pxi
danepitkin: Thank you! Numpy supports [D]ay, but pandas does not.

danepitkin: One thing I found is that Parquet only support [ms], [us], and [ns]. So now several pyarrow dataset tests are failing because datasets with [D]ay units are being converted to [ms] units. I'm somewhat inclined to convert date32 to [ms] by default so we don't have to add a conversion from [ms] -> [s] when doing a parquet roundtrip. Or.. we just let it happen and modify the tests.

danepitkin: This wasn't a problem before when everything was coerced to [ns], which parquet supports.

danepitkin:
Current update: the tests failing locally for me are 1) parquet dataset roundtrips where date32 days are converted to milliseconds instead of seconds because seconds are not supported in parquet and 2) all TZ-aware timestamps are defaulted to nanoseconds (aka I need to add support for other time units in c++).

For (1), I mentioned in another comment that we can convert date32 to millisecond instead of second. For (2), I just need to add support, but it's going to grow this PR even larger unfortunately..

python/pyarrow/types.pxi
jorisvandenbossche: > I'm somewhat inclined to convert date32 to [ms] by default so we don't have to add a conversion from [ms] -> [s] when doing a parquet roundtrip

Yes, that sounds as a good idea (then it also gives the same for date32 vs date64)

jorisvandenbossche: > several pyarrow dataset tests are failing because datasets with [D]ay units are being converted to [ms] units

Can you point to which test is failing? Because this is about conversion from pyarrow to pandas, right? (not arrow<->parquet roundtrip, which should be able to preserve our date32 type because we store the arrow schema)

jorisvandenbossche:
>  For (2), I just need to add support, but it's going to grow this PR even larger unfortunately..

If PR size is a concern, this is also something that could be done as a precursor. It's actually already an issue that shows in conversion to numpy as well:

```
# no timezone -> this preserves the unit
>>> pa.array([1, 2, 3], pa.timestamp('us')).to_numpy()
array(['1970-01-01T00:00:00.000001', '1970-01-01T00:00:00.000002',
       '1970-01-01T00:00:00.000003'], dtype='datetime64[us]')

# with timezone -> always converts to nanoseconds
>>> pa.array([1, 2, 3], pa.timestamp('us', tz="Europe/Brussels")).to_numpy()
...
ArrowInvalid: Needed to copy 1 chunks with 0 nulls, but zero_copy_only was True

>>> pa.array([1, 2, 3], pa.timestamp('us', tz="Europe/Brussels")).to_numpy(zero_copy_only=False)
array(['1970-01-01T00:00:00.000001000', '1970-01-01T00:00:00.000002000',
       '1970-01-01T00:00:00.000003000'], dtype='datetime64[ns]')
```

While this could also be perfectly zero-copy to microseconds in the case with a timezone (we just return the underlying UTC values anyway)

python/pyarrow/types.pxi
danepitkin: The current tests manipulate the types so test cases pass, but these were the tests that originally were failling (the tests in there current state are a bit of a mess right now, I need to go back and clean them up once the implementation actually appears to work properly) : 
```
FAILED pyarrow/tests/parquet/test_dataset.py::test_write_to_dataset_with_partitions[True] - AssertionError: Attributes of DataFrame.iloc[:, 4] (column name="date") are different
FAILED pyarrow/tests/parquet/test_dataset.py::test_write_to_dataset_with_partitions[False] - AssertionError: Attributes of DataFrame.iloc[:, 4] (column name="date") are different
FAILED pyarrow/tests/parquet/test_dataset.py::test_write_to_dataset_with_partitions_and_schema[True] - AssertionError: Attributes of DataFrame.iloc[:, 4] (column name="date") are different
FAILED pyarrow/tests/parquet/test_dataset.py::test_write_to_dataset_with_partitions_and_schema[False] - AssertionError: Attributes of DataFrame.iloc[:, 4] (column name="date") are different
FAILED pyarrow/tests/parquet/test_dataset.py::test_write_to_dataset_with_partitions_and_index_name[True] - AssertionError: Attributes of DataFrame.iloc[:, 4] (column name="date") are different
FAILED pyarrow/tests/parquet/test_dataset.py::test_write_to_dataset_with_partitions_and_index_name[False] - AssertionError: Attributes of DataFrame.iloc[:, 4] (column name="date") are different
FAILED pyarrow/tests/parquet/test_dataset.py::test_write_to_dataset_no_partitions[True] - AssertionError: Attributes of DataFrame.iloc[:, 3] (column name="date") are different
FAILED pyarrow/tests/parquet/test_dataset.py::test_write_to_dataset_no_partitions[False] - AssertionError: Attributes of DataFrame.iloc[:, 3] (column name="date") are different
```

danepitkin:
```
    @pytest.mark.filterwarnings("ignore:'ParquetDataset.schema:FutureWarning")
    def _test_write_to_dataset_with_partitions(base_path,
                                               use_legacy_dataset=True,
                                               filesystem=None,
                                               schema=None,
                                               index_name=None):
        import pandas as pd
        import pandas.testing as tm

        import pyarrow.parquet as pq

        # ARROW-1400
        output_df = pd.DataFrame({'group1': list('aaabbbbccc'),
                                  'group2': list('eefeffgeee'),
                                  'num': list(range(10)),
                                  'nan': [np.nan] * 10,
                                  'date': np.arange('2017-01-01', '2017-01-11',
                                                    dtype='datetime64[D]')})
        output_df["date"] = output_df["date"]
        cols = output_df.columns.tolist()
        partition_by = ['group1', 'group2']
        output_table = pa.Table.from_pandas(output_df, schema=schema, safe=False,
                                            preserve_index=False)
        pq.write_to_dataset(output_table, base_path, partition_by,
                            filesystem=filesystem,
                            use_legacy_dataset=use_legacy_dataset)

        metadata_path = os.path.join(str(base_path), '_common_metadata')

        if filesystem is not None:
            with filesystem.open(metadata_path, 'wb') as f:
                pq.write_metadata(output_table.schema, f)
        else:
            pq.write_metadata(output_table.schema, metadata_path)

        # ARROW-2891: Ensure the output_schema is preserved when writing a
        # partitioned dataset
        dataset = pq.ParquetDataset(base_path,
                                    filesystem=filesystem,
                                    validate_schema=True,
                                    use_legacy_dataset=use_legacy_dataset)
        # ARROW-2209: Ensure the dataset schema also includes the partition columns
        if use_legacy_dataset:
            with pytest.warns(FutureWarning, match="'ParquetDataset.schema'"):
                dataset_cols = set(dataset.schema.to_arrow_schema().names)
        else:
            # NB schema property is an arrow and not parquet schema
            dataset_cols = set(dataset.schema.names)

        assert dataset_cols == set(output_table.schema.names)

        input_table = dataset.read(use_pandas_metadata=True)

        input_df = input_table.to_pandas()

        # Read data back in and compare with original DataFrame
        # Partitioned columns added to the end of the DataFrame when read
        input_df_cols = input_df.columns.tolist()
        assert partition_by == input_df_cols[-1 * len(partition_by):]

        input_df = input_df[cols]
        # Partitioned columns become 'categorical' dtypes
        for col in partition_by:
            output_df[col] = output_df[col].astype('category')
        # if schema is None and Version(pd.__version__) >= Version("2.0.0"):
        #     output_df['date'] = output_df['date'].astype('datetime64[ms]')
>       tm.assert_frame_equal(output_df, input_df)
E       AssertionError: Attributes of DataFrame.iloc[:, 4] (column name="date") are different
E
E       Attribute "dtype" are different
E       [left]:  datetime64[s]
E       [right]: datetime64[ms]
```

jorisvandenbossche:
For the tz aware update, that also influences the (currently untested) to_numpy behaviour, which you can test with the following change:

```diff
--- a/python/pyarrow/tests/test_array.py
+++ b/python/pyarrow/tests/test_array.py
@@ -211,9 +211,10 @@ def test_to_numpy_writable():
         arr.to_numpy(zero_copy_only=True, writable=True)
 
 
+@pytest.mark.parametrize('tz', [None, "UTC"])
 @pytest.mark.parametrize('unit', ['s', 'ms', 'us', 'ns'])
-def test_to_numpy_datetime64(unit):
-    arr = pa.array([1, 2, 3], pa.timestamp(unit))
+def test_to_numpy_datetime64(unit, tz):
+    arr = pa.array([1, 2, 3], pa.timestamp(unit, tz=tz))
     expected = np.array([1, 2, 3], dtype="datetime64[{}]".format(unit))
     np_arr = arr.to_numpy()
     np.testing.assert_array_equal(np_arr, expected)
```

python/pyarrow/src/arrow/python/arrow_to_pandas.cc
```diff
@@ -1569,7 +1572,31 @@ class DatetimeWriter : public TypedPandasWriter<NPY_DATETIME> {
...
+      ConvertDatetimeLikeNanos<int32_t, 86400000L>(*data, out_values);
+    } else if (type == Type::DATE64) {
+      ConvertDatetimeLikeNanos<int64_t, 1LL>(*data, out_values);
+    } else {
+      const auto& ts_type = checked_cast<const TimestampType&>(*data->type());
+      DCHECK_EQ(TimeUnit::MILLI, ts_type.unit())
+          << "Should only call instances of this writer "
+          << "with arrays of the correct unit";
+      ConvertNumericNullable<int64_t>(*data, kPandasTimestampNull,
+                                      this->GetBlockColumnStart(rel_placement));
```
jorisvandenbossche, line 1594:
```suggestion
      ConvertNumericNullable<int64_t>(*data, kPandasTimestampNull, out_values);
```

python/pyarrow/src/arrow/python/arrow_to_pandas.cc
```diff
@@ -1569,7 +1572,31 @@ class DatetimeWriter : public TypedPandasWriter<NPY_DATETIME> {
...
+  using DatetimeWriter<TimeUnit::MILLI>::DatetimeWriter;
+
+  Status CopyInto(std::shared_ptr<ChunkedArray> data, int64_t rel_placement) override {
+    Type::type type = data->type()->id();
+    int64_t* out_values = this->GetBlockColumnStart(rel_placement);
+    if (type == Type::DATE32) {
+      // Convert from days since epoch to datetime64[ms]
+      ConvertDatetimeLikeNanos<int32_t, 86400000L>(*data, out_values);
+    } else if (type == Type::DATE64) {
+      ConvertDatetimeLikeNanos<int64_t, 1LL>(*data, out_values);
```
jorisvandenbossche, line 1587: Can probably use `ConvertNumericNullable` here as well (like in the branch below, that avoids doing the multiplication with 1)

python/pyarrow/src/arrow/python/arrow_to_pandas.cc
```diff
@@ -1569,7 +1572,31 @@ class DatetimeWriter : public TypedPandasWriter<NPY_DATETIME> {
...
+class DatetimeMilliWriter : public DatetimeWriter<TimeUnit::MILLI> {
+ public:
+  using DatetimeWriter<TimeUnit::MILLI>::DatetimeWriter;
+
+  Status CopyInto(std::shared_ptr<ChunkedArray> data, int64_t rel_placement) override {
+    Type::type type = data->type()->id();
+    int64_t* out_values = this->GetBlockColumnStart(rel_placement);
+    if (type == Type::DATE32) {
+      // Convert from days since epoch to datetime64[ms]
+      ConvertDatetimeLikeNanos<int32_t, 86400000L>(*data, out_values);
```
jorisvandenbossche, line 1585: I am wondering that if we do such naive multiplication here, we can get overflow errors for out-of-bounds timestamps (but this is already the case for the current code converting to nanoseconds, as well, to be clear)

python/pyarrow/tests/parquet/common.py
```diff
@@ -176,8 +176,8 @@ def alltypes_sample(size=10000, seed=0, categorical=False):
         # TODO(wesm): Test other timestamp resolutions now that arrow supports
         # them
         'datetime': np.arange("2016-01-01T00:00:00.001", size,
-                              dtype='datetime64[ms]').astype('datetime64[ns]'),
-        'timedelta': np.arange(0, size, dtype="timedelta64[ns]"),
+                              dtype='datetime64[ms]'),
```
jorisvandenbossche, line 179: We can maybe keep both original ns and new ms resolution? (to test both)

python/pyarrow/tests/test_pandas.py
```diff
@@ -70,7 +70,7 @@ def _alltypes_example(size=100):
         # TODO(wesm): Pandas only support ns resolution, Arrow supports s, ms,
         # us, ns
         'datetime': np.arange("2016-01-01T00:00:00.001", size,
-                              dtype='datetime64[ms]').astype("datetime64[ns]"),
+                              dtype='datetime64[ms]'),
```
jorisvandenbossche: Or also here keep two resolutions?

danepitkin:
> For the tz aware update, that also influences the (currently untested) to_numpy behaviour, which you can test with the following change:
> 
> ```diff
> --- a/python/pyarrow/tests/test_array.py
> +++ b/python/pyarrow/tests/test_array.py
> @@ -211,9 +211,10 @@ def test_to_numpy_writable():
>          arr.to_numpy(zero_copy_only=True, writable=True)
>  
>  
> +@pytest.mark.parametrize('tz', [None, "UTC"])
>  @pytest.mark.parametrize('unit', ['s', 'ms', 'us', 'ns'])
> -def test_to_numpy_datetime64(unit):
> -    arr = pa.array([1, 2, 3], pa.timestamp(unit))
> +def test_to_numpy_datetime64(unit, tz):
> +    arr = pa.array([1, 2, 3], pa.timestamp(unit, tz=tz))
>      expected = np.array([1, 2, 3], dtype="datetime64[{}]".format(unit))
>      np_arr = arr.to_numpy()
>      np.testing.assert_array_equal(np_arr, expected)
> ```

Thank you! Updated and it passes out of the gate 🎉

python/pyarrow/tests/test_pandas.py
```diff
@@ -1202,7 +1211,7 @@ def test_table_convert_date_as_object(self):
         df_datetime = table.to_pandas(date_as_object=False)
         df_object = table.to_pandas()
 
-        tm.assert_frame_equal(df.astype('datetime64[ns]'), df_datetime,
+        tm.assert_frame_equal(df.astype('datetime64[ms]'), df_datetime,
```
jorisvandenbossche, line 1218: Do we have coverage for testing that it stays nanoseconds if you specify `coerce_temporal_nanoseconds=True`?

python/pyarrow/tests/test_pandas.py
danepitkin: Added s, ms, us!

python/pyarrow/src/arrow/python/arrow_to_pandas.cc
danepitkin: Technically, I think this specific multiplication is fine. INT32_MAX * 86400000 = 1.8554259e+17, while INT64_MAX = 9.223372e+18.

python/pyarrow/tests/parquet/test_dataset.py
```diff
@@ -1313,6 +1312,15 @@ def _test_write_to_dataset_with_partitions(base_path,
...
     for col in partition_by:
         output_df[col] = output_df[col].astype('category')
+
+    if schema:
+        expected_date_type = schema.field_by_name('date').type.to_pandas_dtype()
+    else:
+        # Arrow to Pandas v2 will convert date32 to [ms]. Pandas v1 will always
+        # silently coerce to [ns] due to non-[ns] support.
+        expected_date_type = 'datetime64[ms]'
+    output_df["date"] = output_df["date"].astype(expected_date_type)
```
danepitkin, line 1322: There is no issue with reading datsets, I just had to update this test to check if the resulting pandas df matched the overridden schema types if a schema is applied. There is probably a better way to check then what I've done here..

danepitkin:
Should we run crossbow on this change to? IIRC only committers can trigger it. Does it run the same tests as the nightly builds?

jorisvandenbossche:

python/pyarrow/array.pxi
```diff
@@ -721,12 +722,12 @@ cdef class _PandasConvertible(_Weakrefable):
         integer_object_nulls : bool, default False
             Cast integers with nulls to objects
         date_as_object : bool, default True
-            Cast dates to objects. If False, convert to datetime64[ns] dtype.
+            Cast dates to objects. If False, convert to datetime64 dtype.
```
jorisvandenbossche, line 725: Do we want to specify this is [ns] for older pandas and [ms] for pandas >= 2.0?

python/pyarrow/tests/test_pandas.py
```diff
@@ -1199,13 +1223,25 @@ def test_table_convert_date_as_object(self):
...
         df_object = table.to_pandas()
 
-        tm.assert_frame_equal(df.astype('datetime64[ns]'), df_datetime,
+        tm.assert_frame_equal(df.astype(expected_type), df_datetime,
                               check_dtype=True)
         tm.assert_frame_equal(df, df_object, check_dtype=True)
 
+    def test_table_coerce_temporal_nanoseconds(self):
+        df = pd.DataFrame({'date': [date(2000, 1, 1)]}, dtype='datetime64[ms]')
+        table = pa.Table.from_pandas(df)
```
jorisvandenbossche, line 1236: We should maybe parametrize this test with using date32/date64/timestamp? Or is the date32/64 case tested elsewhere?

python/pyarrow/array.pxi
```diff
@@ -775,6 +776,13 @@ cdef class _PandasConvertible(_Weakrefable):
             expected to return a pandas ExtensionDtype or ``None`` if the
             default conversion should be used for that type. If you have
             a dictionary mapping, you can pass ``dict.get`` as function.
+        coerce_temporal_nanoseconds : bool, default False
+            Only applicable to pandas version >= 2.0.
+            A legacy option to coerce date32, date64, datetime, and timestamp
```
jorisvandenbossche, line 781:
```suggestion
            A legacy option to coerce date32, date64, and timestamp
```

python/pyarrow/src/arrow/python/arrow_to_pandas.cc
jorisvandenbossche: Ah, yes, when milliseconds is the target this is probably fine. For the nanoseconds case, though, this gives wrong results. Opened https://github.com/apache/arrow/issues/36084 about that.

jorisvandenbossche:
The test build with pandas nightly has one related failure:

```
_____ TestConvertDateTimeLikeTypes.test_timestamp_to_pandas_out_of_bounds ______

self = <pyarrow.tests.test_pandas.TestConvertDateTimeLikeTypes object at 0x7f022e57aee0>

    def test_timestamp_to_pandas_out_of_bounds(self):
        # ARROW-7758 check for out of bounds timestamps for non-ns timestamps
        # that end up getting coerced into ns timestamps.
    
        for unit in ['s', 'ms', 'us']:
            for tz in [None, 'America/New_York']:
                arr = pa.array([datetime(1, 1, 1)], pa.timestamp(unit, tz=tz))
                table = pa.table({'a': arr})
    
                msg = "would result in out of bounds timestamp"
                with pytest.raises(ValueError, match=msg):
                    print(arr.to_pandas(coerce_temporal_nanoseconds=True))
    
                with pytest.raises(ValueError, match=msg):
                    table.to_pandas(coerce_temporal_nanoseconds=True)
    
>               with pytest.raises(ValueError, match=msg):
E               Failed: DID NOT RAISE <class 'ValueError'>

opt/conda/envs/arrow/lib/python3.10/site-packages/pyarrow/tests/test_pandas.py:1495: Failed
```

jorisvandenbossche:
The dask test build has a whole bunch of failures(https://github.com/ursacomputing/crossbow/actions/runs/5275817890/jobs/9541737015), but that is because they have tests for reading/writing parquet, and they will have to update their tests, similarly as we are updating our own parquet tests here. 

All failures show something like 

```
E       Attribute "dtype" are different
E       [left]:  datetime64[ns]
E       [right]: datetime64[us]
```

so that is the expected failure.

(although I am just thinking, the proposed change to update parquet to write nanoseconds by default (https://github.com/apache/arrow/issues/35746) would help for keeping those tests working as before ..)

jorisvandenbossche (commented):
Great work!

In test_pandas.py, there is a `test_timedeltas_(no_)nulls` test. Maybe you could parametrize that one as well to have some coverage for timedelta/duration

Note: I reviewed most of the PR in detail, I only still have to go through the parquet tests

python/pyarrow/array.pxi
danepitkin: I will add the case for pandas 1.x! For pandas 2.0, the conversion will match the pyarrow time unit so it could range from `s` to `ns`. The only surprise from the user's perspective is date32, which translates `Day` to `ms`.

python/pyarrow/tests/test_pandas.py
danepitkin: Its not, good catch! Will update.

python/pyarrow/array.pxi
danepitkin: oof, good catch!

python/pyarrow/tests/test_pandas.py
```diff
@@ -1452,26 +1487,27 @@ def test_timestamp_to_pandas_out_of_bounds(self):
 
                 msg = "would result in out of bounds timestamp"
                 with pytest.raises(ValueError, match=msg):
-                    arr.to_pandas()
+                    print(arr.to_pandas(coerce_temporal_nanoseconds=True))
```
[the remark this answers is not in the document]
danepitkin: 🤦

python/pyarrow/tests/test_pandas.py
```diff
@@ -1199,13 +1223,25 @@ def test_table_convert_date_as_object(self):
...
-        tm.assert_frame_equal(df.astype('datetime64[ns]'), df_datetime,
+        tm.assert_frame_equal(df.astype(expected_type), df_datetime,
                               check_dtype=True)
         tm.assert_frame_equal(df, df_object, check_dtype=True)
 
+    def test_table_coerce_temporal_nanoseconds(self):
+        df = pd.DataFrame({'date': [date(2000, 1, 1)]}, dtype='datetime64[ms]')
+        table = pa.Table.from_pandas(df)
+        result_df = table.to_pandas(
+            coerce_temporal_nanoseconds=True, date_as_object=False)
```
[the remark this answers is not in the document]
danepitkin: I went ahead and parametrized for dates + timestamps!

python/pyarrow/tests/test_pandas.py
```diff
@@ -1199,13 +1223,25 @@ def test_table_convert_date_as_object(self):
...
+        expected_df = df.astype('datetime64[ns]')
+        tm.assert_frame_equal(result_df, expected_df)
+        result_df = table.to_pandas(
+            coerce_temporal_nanoseconds=False, date_as_object=False)
```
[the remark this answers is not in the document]
danepitkin: I went ahead and parametrized for dates + timestamps!

python/pyarrow/tests/test_pandas.py
```diff
@@ -1199,13 +1223,25 @@ def test_table_convert_date_as_object(self):
...
+        df_datetime = table.to_pandas(date_as_object=False,
+                                      coerce_temporal_nanoseconds=coerce_to_ns)
         df_object = table.to_pandas()
 
-        tm.assert_frame_equal(df.astype('datetime64[ns]'), df_datetime,
+        tm.assert_frame_equal(df.astype(expected_type), df_datetime,
                               check_dtype=True)
         tm.assert_frame_equal(df, df_object, check_dtype=True)
 
+    def test_table_coerce_temporal_nanoseconds(self):
```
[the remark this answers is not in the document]
danepitkin: Good call!

python/pyarrow/tests/test_schema.py
```diff
@@ -45,7 +46,10 @@ def test_type_integers():
 
 
 def test_type_to_pandas_dtype():
-    M8_ns = np.dtype('datetime64[ns]')
+    M8 = np.dtype('datetime64[ms]')
+    if _pandas_api.is_v1():
```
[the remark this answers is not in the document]
danepitkin: I'll go ahead and update. I think this is the better approach!

python/pyarrow/tests/test_pandas.py
```diff
@@ -67,10 +67,14 @@ def _alltypes_example(size=100):
...
-        'datetime': np.arange("2016-01-01T00:00:00.001", size,
-                              dtype='datetime64[ms]').astype("datetime64[ns]"),
+        'datetime[s]': np.arange("2016-01-01T00:00:00.001", size,
+                                 dtype='datetime64[s]'),
+        'datetime[ms]': np.arange("2016-01-01T00:00:00.001", size,
+                                  dtype='datetime64[ms]'),
+        'datetime[us]': np.arange("2016-01-01T00:00:00.001", size,
+                                  dtype='datetime64[us]'),
+        'datetime[ns]': np.arange("2016-01-01T00:00:00.001", size,
+                                  dtype='datetime64[ns]'),
```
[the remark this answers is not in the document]
danepitkin: Will do!

python/pyarrow/tests/test_array.py
```diff
@@ -212,8 +212,9 @@ def test_to_numpy_writable():
...
 
 @pytest.mark.parametrize('unit', ['s', 'ms', 'us', 'ns'])
-def test_to_numpy_datetime64(unit):
-    arr = pa.array([1, 2, 3], pa.timestamp(unit))
+@pytest.mark.parametrize('tz', [None, "UTC"])
+def test_to_numpy_datetime64(unit, tz):
+    arr = pa.array([1, 2, 3], pa.timestamp(unit, tz=tz))
     expected = np.array([1, 2, 3], dtype="datetime64[{}]".format(unit))
     np_arr = arr.to_numpy()
     np.testing.assert_array_equal(np_arr, expected)
```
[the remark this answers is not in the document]
danepitkin: Looks like this is tested in test_pandas.py in `test_timestamps_with_timezone`. It will call the following function on a pandas series and I confirmed it is converted into a pyarrow TimestampArray (vs a ChunkedArray)

```
def _check_series_roundtrip(s, type_=None, expected_pa_type=None):
    arr = pa.array(s, from_pandas=True, type=type_)

    if type_ is not None and expected_pa_type is None:
        expected_pa_type = type_

    if expected_pa_type is not None:
        assert arr.type == expected_pa_type

    result = pd.Series(arr.to_pandas(), name=s.name)
    tm.assert_series_equal(s, result)
```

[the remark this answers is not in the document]
danepitkin: I did add s/ms/us time unit support to the above test case as well!

danepitkin:
TODO: 
* Fix Numpy date32 regression
* Fix Pandas nightly test failure

(all other comments are addressed to the best of my knowledge)

python/pyarrow/tests/parquet/common.py
danepitkin: Good suggestion!

python/pyarrow/src/arrow/python/arrow_to_pandas.cc
```diff
@@ -2060,9 +2097,10 @@ static Status GetPandasWriterType(const ChunkedArray& data, const PandasOptions&
     case Type::DATE64:
       if (options.date_as_object) {
         *output_type = PandasWriter::OBJECT;
+      } else if (options.coerce_temporal_nanoseconds) {
+        *output_type = PandasWriter::DATETIME_NANO;
       } else {
-        *output_type = options.coerce_temporal_nanoseconds ? PandasWriter::DATETIME_NANO
-                                                           : PandasWriter::DATETIME_DAY;
+        *output_type = PandasWriter::DATETIME_MILLI;
```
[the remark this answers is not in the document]
danepitkin: Fixed! I ended up adding a `to_numpy` bool in `PandasOptions`.

danepitkin:
All feedback has been applied! This PR is ready for re-review.

I've added support for passing `coerce_temporal_nanoseconds` to `to_pandas_dtype` internally. I thought it best not to expose this option in the user-facing API since it is legacy default behavior. This does mean we have to check if we are using an extension type in `to_pandas` and call the extension types `to_pandas_dtype()` method.

danepitkin:

python/pyarrow/array.pxi
jorisvandenbossche: > For pandas 2.0, the conversion will match the pyarrow time unit so it could range from `s` to `ns`

This is for dates (so date32/date64), so isn't that always "ms" unit?

danepitkin:
The pandas integration test failure is:

```
>               table.column('a').to_pandas(
                    safe=False, coerce_temporal_nanoseconds=True)

opt/conda/envs/arrow/lib/python3.10/site-packages/pyarrow/tests/test_pandas.py:1535: 
pyarrow/array.pxi:868: in pyarrow.lib._PandasConvertible.to_pandas
    ???
pyarrow/table.pxi:472: in pyarrow.lib.ChunkedArray._to_pandas
    ???
opt/conda/envs/arrow/lib/python3.10/site-packages/pandas/core/dtypes/dtypes.py:866: in __from_arrow__
    array = array.cast(pyarrow.timestamp(unit=self._unit), safe=True)
pyarrow/table.pxi:556: in pyarrow.lib.ChunkedArray.cast
    ???
opt/conda/envs/arrow/lib/python3.10/site-packages/pyarrow/compute.py:402: in cast
    return call_function("cast", [arr], options, memory_pool)
pyarrow/_compute.pyx:572: in pyarrow._compute.call_function
    ???
pyarrow/_compute.pyx:367: in pyarrow._compute.Function.call
    ???
pyarrow/error.pxi:144: in pyarrow.lib.pyarrow_internal_check_status
    ???

>   ???
E   pyarrow.lib.ArrowInvalid: Casting from timestamp[s, tz=America/New_York] to timestamp[ns] would result in out of bounds timestamp: -62135596800
```

The issue here is that the pyarrow user can specify the `safe=False` option, but pyarrow internally can't pass that option to the pandas `DatetimeTZDtype.__from_arrow_` method: https://github.com/pandas-dev/pandas/blob/main/pandas/core/dtypes/dtypes.py#L897.  

The only option is to either disallow usage of options in pyarrow's `to_pandas()` methods for pandas ExtensionDType objects or ignore the __from_arrow__ method and do the conversion in pyarrow instead. 

Another approach is to explicitly add option parameters to `__from_arrow__`, but his adds a higher level of complexity to package integration.

danepitkin:
I'll first try to merge https://github.com/apache/arrow/pull/36314 and rebase afterwards. That way, the issue can be fixed properly here before merging.

danepitkin:

danepitkin:
Will wait to rebase on top of https://github.com/apache/arrow/pull/36137 before merging.

jorisvandenbossche:
@danepitkin I merged the Parquet version 2.6 PR https://github.com/apache/arrow/pull/36137, so this can be updated now

danepitkin:

python/pyarrow/tests/parquet/test_dataset.py
```diff
@@ -1313,6 +1312,15 @@ def _test_write_to_dataset_with_partitions(base_path,
     # Partitioned columns become 'categorical' dtypes
     for col in partition_by:
         output_df[col] = output_df[col].astype('category')
+
+    if schema:
+        expected_date_type = schema.field_by_name('date').type.to_pandas_dtype()
+    else:
+        # Arrow to Pandas v2 will convert date32 to [ms]. Pandas v1 will always
+        # silently coerce to [ns] due to non-[ns] support.
+        expected_date_type = 'datetime64[ms]'
```
jorisvandenbossche, line 1321: This comment is not fully correct, I think (when converting the pandas dataframe to pyarrow, we actually don't have date32, but timestamp type). But then I also don't understand how this test is passing ..

jorisvandenbossche: So what actually happens with pandas 2.x: when we create a DataFrame with datetime64[D], that gets converted to datetime64[s] (closest supported resolution to "D"). Then roundtripping to parquet turns that into "ms" (because "s" is not supported by Parquet)

With older pandas this gets converted to datetime64[ns], will come back from Parquet as "us", and converted back to "ns" when converting to pandas. But this `astype("datetime64[ms]")` essentially doesn't do anything, i.e. pandas does preserve the "ns" because it doesn't support "ms", and hence the test also passes for older pandas. 

Maybe it's simpler to just test with a DataFrame of nanoseconds, which now works the same with old and new pandas, and then we don't have to add any comment or astype.

jorisvandenbossche: > Maybe it's simpler to just test with a DataFrame of nanoseconds, which now works the same with old and new pandas, and then we don't have to add any comment or astype.

Hmm, trying that out locally fails (but only with the non-legacy code path), and digging in, it seems that we are still writing Parquet v 1 files with the dataset API ... 
Will open a separate issue and PR to quickly fix that separately.

jorisvandenbossche: -> https://github.com/apache/arrow/pull/36538

jorisvandenbossche (approved):
I pushed a small clean-up on top of my PR to fix the Parquet version for the dataset writer. Further checked all changes to the parquet tests as well, and all looks good!

jorisvandenbossche:

jorisvandenbossche:
And with this PR and the Parquet v2.6 update combined, the failures in the dask builds are now much smaller (just one failure that was testing that a timestamp would overflow by being casted to nanoseconds)

jorisvandenbossche:
Thanks @danepitkin!

danepitkin:
Thanks @jorisvandenbossche for the collaboration and support!

python/pyarrow/tests/test_pandas.py
```diff
@@ -4179,20 +4258,20 @@ def test_to_pandas_extension_dtypes_mapping():
     assert isinstance(result['a'].dtype, pd.PeriodDtype)
 
 
-def test_array_to_pandas():
+@pytest.mark.parametrize("arr",
+                         [pd.period_range("2012-01-01", periods=3, freq="D").array,
+                          pd.interval_range(1, 4).array])
```
kou, line 4263: @jorisvandenbossche @danepitkin We can't use `pd` here because pandas may not be available.
This causes an error on "no pandas" environment: https://github.com/apache/arrow/actions/runs/5496447565/jobs/10016477233
This PR's CI succeeded because our "Without Pandas" job installed pandas implicitly. It has been fixed by #36542.

Could you open an issue for this and fix this?

jorisvandenbossche: Fixed -> https://github.com/apache/arrow/pull/36586

PR #21, opened 2016-03-08
ARROW-43: Python: format array values to in __repr__ for interactive computing

wesm:

PR #35672, opened 2023-05-18
GH-35176: [C++] Add support for disabling threading for emscripten

### LINKED ISSUE #35176: [C++] Add support for disablging threading for emscripten

joemarshall:
As previously discussed in #35176 this is a patch that adds an option `ARROW_ENABLE_THREADING`. When it is turned off, arrow threadpool and serial executors don't spawn threads, and instead run tasks in the main thread when futures are waited for.

It doesn't mess with threading in projects included as dependencies, e.g. multithreaded malloc implementations because if you're building for a non threaded environment, you can't use those anyway.

Basically where this is at is that it runs the test suite okay, and I think should work well enough to be a backend for pandas on emscripten/pyodide.

What this means is:
1) It is possible to use arrow in non-threaded emscripten/webassembly environments (with some build patches specific to emscripten which I'll put in once this is in)
2) Most of arrow just works, albeit slower in parts.

Things that don't work and probably won't:
1) Server stuff that relies on threads. Not a massive problem I think because environments with threading restrictions are currently typically also restricted from making servers anyway (i.e. they are web browsers)
2) Anything that relies on actually doing two things at once (for obvious reasons)

Things that don't work yet and could be fixed in future:
1) use of asynchronous file/network APIs in emscripten which would mean I/O could work efficiently in one thread.
2) asofjoin - right now the implementation relies on std::thread - it needs refactoring to work with threadpool like everything else in arrow, but I'm not sure I am expert enough in the codebase to do it well.
* Closes: #35176


joemarshall:
@kou Could you try and run the github actions bot on here. If it works then I'll copy all the description across and remove the previous PR

thanks,
Joe

kou:
@github-actions crossbow submit test-ubuntu-22.04-cpp-no-threading

ci/scripts/cpp_build.sh
```diff
@@ -129,6 +129,7 @@ cmake \
   -DARROW_WITH_UTF8PROC=${ARROW_WITH_UTF8PROC:-ON} \
   -DARROW_WITH_ZLIB=${ARROW_WITH_ZLIB:-OFF} \
   -DARROW_WITH_ZSTD=${ARROW_WITH_ZSTD:-OFF} \
+  -DARROW_ENABLE_THREADING=${ARROW_ENABLE_THREADING:-ON} \
```
kou, line 132: Could you keep this list in alphabetical order?

ci/scripts/cpp_test.sh
```diff
@@ -80,6 +80,12 @@ pushd ${build_dir}
 if [ -z "${PYTHON}" ] && ! which python > /dev/null 2>&1; then
   export PYTHON="${PYTHON:-python3}"
 fi
+if [[ "${ARROW_ENABLE_THREADING}" == "OFF" ]]; then
+# if threading is disabled, some tests take longer to run
+# but we can get away with running more tests in parallel because we know they wont make loads of threads
```
kou, line 85:
```suggestion
  # if threading is disabled, some tests take longer to run
  # but we can get away with running more tests in parallel because we know they won't make loads of threads
```

ci/scripts/cpp_test.sh
```diff
@@ -80,6 +80,12 @@ pushd ${build_dir}
...
+  ARROW_CTEST_TIMEOUT=${ARROW_CTEST_TIMEOUT:-900}
```
kou, line 86: It seems that max test time is "239.99 sec" in our CI:

https://github.com/ursacomputing/crossbow/actions/runs/5018372445/jobs/8997694994#step:6:3234

> ```text
> 89/103 Test  #21: arrow-compute-scalar-if-else-test .........   Passed  239.99 sec
> ```

Why is this needed?

ci/scripts/cpp_test.sh
```diff
@@ -80,6 +80,12 @@ pushd ${build_dir}
...
+  n_jobs=$((n_jobs*4))
```
kou, line 87: I don't think that this is a good idea...
If we use more test processes than the number of cores, each test process competes CPU resource with others...

cpp/cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -5190,4 +5190,4 @@ message(STATUS "All bundled static libraries: ${ARROW_BUNDLED_STATIC_LIBS}")
 
 configure_file("src/arrow/util/config.h.cmake" "src/arrow/util/config.h" ESCAPE_QUOTES)
 install(FILES "${ARROW_BINARY_DIR}/src/arrow/util/config.h"
-        DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/arrow/util")
+        DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/arrow/util")
```
kou, line 5193: Could you revert this needless change?

cpp/src/arrow/acero/asof_join_node.cc
```diff
@@ -1514,14 +1514,19 @@ class AsofJoinNode : public ExecNode {
   }
 
   Status StartProducing() override {
-    ARROW_ASSIGN_OR_RAISE(process_task_, plan_->query_context()->BeginExternalTask(
-                                             "AsofJoinNode::ProcessThread"));
-    if (!process_task_.is_valid()) {
-      // Plan has already aborted.  Do not start process thread
+    #ifndef ARROW_ENABLE_THREADING
+      return Status::NotImplemented("ASOF join requires threading enabled");
+    #else
```
kou, line 1519: Can we use the following approach?

```cpp
Status StartProducing() override {
#ifndef ARROW_ENABLE_THREADING
  return Status::NotImplemented("ASOF join requires threading enabled");
#endif
  ARROW_ASSIGN_OR_RAISE(process_task_, plan_->query_context()->BeginExternalTask(
  // use existing code as-is.
```

Putting existing code to `#else ... #endif` reduces readability.

cpp/src/arrow/acero/bloom_filter_test.cc
```diff
@@ -507,7 +507,11 @@ TEST(BloomFilter, Scaling) {
   num_build.push_back(4000000);
 
   std::vector<BloomFilterBuildStrategy> strategy;
+#ifndef ARROW_ENABLE_THREADING
+  strategy.push_back(BloomFilterBuildStrategy::SINGLE_THREADED);
+#else
   strategy.push_back(BloomFilterBuildStrategy::PARALLEL);
+#endif
```
kou, line 514: Could you use `#ifdef ... #else ... #endif` not `#ifndef ... #else ... #endif` as much as possible? A `#ifndef` (invert) and `#else` (invert) pair reduces readability.

```cpp
#ifdef ARROW_ENABLE_THREADING
  strategy.push_back(BloomFilterBuildStrategy::PARALLEL);
#else
  strategy.push_back(BloomFilterBuildStrategy::SINGLE_THREADED);
#endif
```

kou (commented):
Could you revert needless changes and fix lint failures?

ci/scripts/cpp_test.sh
joemarshall: Reverted this stuff - I think a bug was causing the slow tests in my build.

cpp/cmake_modules/ThirdpartyToolchain.cmake
joemarshall: oops done now.

joemarshall:
Right - that's fixed, I think ready for review

cpp/CMakeLists.txt
```diff
@@ -370,6 +370,7 @@ if(ARROW_ENABLE_TIMING_TESTS)
   add_definitions(-DARROW_WITH_TIMING_TESTS)
 endif()
 
+
```
kou, line 373: Could you revert needless changes like this?
It seems that there are more needless changes. Could you check again?

joemarshall: Darn, I ran auto-lint on everything, sorry,  will remove them.

kou (commented):
(My review isn't completed yet but I submit this for now. Sorry.)

cpp/src/arrow/testing/gtest_util.cc
```diff
@@ -725,8 +726,24 @@ void TestInitialized(const ArrayData& array) {
 }
 
 void SleepFor(double seconds) {
+#ifdef ARROW_ENABLE_THREADING
   std::this_thread::sleep_for(
       std::chrono::nanoseconds(static_cast<int64_t>(seconds * 1e9)));
+#else
+  std::chrono::duration<double> secs_left = std::chrono::duration<double>(seconds);
```
kou, line 733:
```suggestion
  auto secs_left = std::chrono::duration<double>(seconds);
```

cpp/src/arrow/testing/gtest_util.cc
```diff
@@ -725,8 +726,24 @@ void TestInitialized(const ArrayData& array) {
...
+  auto start_time = std::chrono::steady_clock::now();
+  auto end_time = start_time + secs_left;
+  auto now = start_time;
+  while (now < end_time) {
```
kou, line 737:
```suggestion
  while (std::chrono::steady_clock::now() < end_time) {
```

cpp/src/arrow/testing/gtest_util.cc
```diff
@@ -725,8 +726,24 @@ void TestInitialized(const ArrayData& array) {
...
+    bool run_task = arrow::internal::SerialExecutor::RunTasksOnAllExecutors(true);
+    now = std::chrono::steady_clock::now();
+    if (!run_task) {
+      // all executors are empty, just sleep for the rest of the time
+      std::this_thread::sleep_for(end_time - now);
```
kou, line 742:
```suggestion
      std::this_thread::sleep_for(end_time - std::chrono::steady_clock::now());
```

cpp/src/arrow/testing/gtest_util.cc
```diff
@@ -1048,16 +1118,32 @@ class GatingTask::Impl : public std::enable_shared_from_this<GatingTask::Impl> {
                                  " seconds) waiting for the gating task to be unlocked");
     }
     num_finished_++;
+#else
+    // can't wait here for anything, so make a future to do the waiting
+    num_running_++;
+    auto future = RunTaskFuture();
+    future.Wait();
+    return;
```
kou, line 1126:
```suggestion
```

cpp/src/arrow/testing/gtest_util.cc
```diff
@@ -1048,16 +1118,32 @@ class GatingTask::Impl : public std::enable_shared_from_this<GatingTask::Impl> {
                                  " seconds) waiting for the gating task to be unlocked");
     }
     num_finished_++;
+#else
+    // can't wait here for anything, so make a future to do the waiting
+    num_running_++;
+    auto future = RunTaskFuture();
+    future.Wait();
```
kou, line 1125: Can we just use `BusyWait()` here?

cpp/src/arrow/util/async_generator_test.cc
```diff
@@ -1486,13 +1490,24 @@ TEST(TestAsyncUtil, ReadaheadFailed) {
...
     return DeferNotOk(thread_pool->Submit([&, count]() -> Result<TestInt> {
       gating_task->Task()();
       if (count == 0) {
         return Status::Invalid("X");
       }
       return TestInt(count);
     }));
+#else
+    // if threading is disabled, we can't call Task() as we do below because it will
+    // never return and will block everything
```
kou, line 1503: Should we skip this test instead of adding codes here?

cpp/src/arrow/util/async_generator_test.cc
```diff
@@ -1530,6 +1547,16 @@ TEST(TestAsyncUtil, ReadaheadFailedWaitForInFlight) {
       // These are our in-flight tasks
       return TestInt(0);
     }));
+#else
+    // if threading is disabled, we can't call Task() as we do below because it will
+    // never return and will block everything
```
kou, line 1552: Should we skip this test instead of adding codes here?

cpp/src/arrow/util/future.cc
```diff
@@ -89,7 +90,7 @@ class ConcreteFutureImpl : public FutureImpl {
       case ShouldSchedule::IfUnfinished:
         return !in_add_callback;
       case ShouldSchedule::IfDifferentExecutor:
-        return !callback_record.options.executor->OwnsThisThread();
+        return !(callback_record.options.executor->IsCurrentExecutor());
```
kou, line 93: Why do we need this change?
Can we just disable (or error) this feature without threading?

cpp/src/arrow/util/future.cc
```diff
@@ -149,17 +150,38 @@ class ConcreteFutureImpl : public FutureImpl {
   }
 
   void DoWait() {
+#ifdef ARROW_ENABLE_THREADING
     std::unique_lock<std::mutex> lock(mutex_);
-
```
kou, line 153: Could you revert needless change?

cpp/src/arrow/util/thread_pool.cc
```diff
@@ -183,6 +247,105 @@ void SerialExecutor::RunLoop() {
   }
   state_->current_thread = {};
 }
+#else   // ARROW_ENABLE_THREADING
+bool SerialExecutor::RunTasksOnAllExecutors(bool once_only) {
```
kou, line 251: Why do we need this?
Can we disable (or error) features that need this method?

cpp/src/arrow/util/thread_pool.cc
```diff
@@ -532,6 +695,65 @@ int ThreadPool::DefaultCapacity() {
   return capacity;
 }
 
+#else  // ARROW_ENABLE_THREADING
+ThreadPool::ThreadPool() {
```
kou, line 699: Can we return `Status::NotImplemented` from all `ThreadPool` methods instead of implementing `ThreadPool` without threading?

cpp/src/arrow/util/thread_pool_test.cc
```diff
@@ -341,6 +342,7 @@ TEST(SerialExecutor, AsyncGeneratorWithCleanup) {
   // must run before the terminal item is delivered from the iterator.
   bool follow_up_ran = false;
   Iterator<TestInt> iter =
+
```
kou, line 345:
```suggestion
```

cpp/src/arrow/util/thread_pool_test.cc
```diff
@@ -598,8 +603,11 @@ TEST_F(TestThreadPool, OwnsCurrentThread) {
   ASSERT_FALSE(pool->OwnsThisThread());
   ASSERT_FALSE(one_failed);
 }
-
```
kou, line 601: Could you revert all needless changes?

cpp/src/arrow/util/thread_pool_test.cc
```diff
@@ -845,8 +877,11 @@ TEST_F(TestThreadPoolForkSafety, Basics) {
 }
 
 TEST_F(TestThreadPoolForkSafety, MultipleChildThreads) {
-  // ARROW-15593: race condition in after-fork ThreadPool reinitialization
-  // when SpawnReal() was called from multiple threads in a forked child.
+// ARROW-15593: race condition in after-fork ThreadPool reinitialization
+// when SpawnReal() was called from multiple threads in a forked child.
```
kou, line 881: Could you move this after `GTEST_SKIP()`?

cpp/src/arrow/util/thread_pool.cc
joemarshall: @kou I was going to do that initially, but the problem is that there is so much code that assumes that there is an IO pool and a cpu threadpool and that they are different things that it was easier just to keep track of them. And even code that uses serialexecutors assumes that other things will run concurrently because of the i/o pools. I found that rather than refactoring all code that uses thread pools, it was easier to make the threadpool & serialexecutors do some basic cooperative multitasking instead. This way also has an advantage in that it should be possible to incrementally patch in asynchronous features where the non-threading OS supports them, e.g. in emscripten it might use async calls to javascript for some types of i/o.

cpp/src/arrow/util/thread_pool.cc
joemarshall: I think pretty much anything that does I/O assumes that the i/o calls and serialexecutor calls run in parallel. Without this stuff, serialexecutors can't run at the same time as I/O things which breaks quite a lot of things including loading parquet files. Executors need to be able to assume that other things will happen if they don't have any tasks ready to run or else bad things happen.

cpp/src/arrow/util/future.cc
joemarshall: I'm not sure - I didn't really know what this feature was used for and who uses it, so I made sure I didn't break it.

cpp/src/arrow/testing/gtest_util.cc
joemarshall: I don't think so, but I can't remember why

joemarshall:
@kou - I did a couple of things on this - firstly I fixed most of those comments you had (and added comments where I think things might not need fixing). Secondly, I ran a build in arm64 on oracle cloud, and the test that failed in CI works just fine on there, so maybe that is just a fluke.

kou:
Sorry. I can't review this immediately because I don't have enough time for now. It seems that I need to run tests with `-DARROW_ENABLE_THREADING=OFF`  on GDB , write some codes and so on to consider what approach is reasonable. I hope that I can do it in the next month.

(@westonpace or @pitrou may be able to review this.)

westonpace (changes requested):
Thank you for working on this.  This will be a great feature.  I haven't fully understood everything but I think I am getting the picture.  Could you just give a general description of how you might have multiple serial executors?  Is this from multiple user threads?  Or is this becuase the CPU executor and the I/O executor need to be two distinct executors for some reason?

cpp/src/arrow/util/thread_pool.cc
```diff
@@ -60,11 +61,33 @@ struct SerialExecutor::State {
...
+std::unordered_set<SerialExecutor*> SerialExecutor::all_executors;
+SerialExecutor* SerialExecutor::current_executor = NULL;
+SerialExecutor* SerialExecutor::GetCurrentExecutor() { return current_executor; }
+SerialExecutor* SerialExecutor::last_called_executor = NULL;
+
+#endif
+
+SerialExecutor::SerialExecutor() : state_(std::make_shared<State>()) {
+#ifndef ARROW_ENABLE_THREADING
+  all_executors.insert(this);
```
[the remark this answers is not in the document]
joemarshall: Yep, I'm assuming that the main use case for disabling threading is use of arrow in emscripten or similarly constrained environments where threading doesn't work (and synchronisation primitives may also not work for that matter - in emscripten they are no ops)

joemarshall:
> Thank you for working on this.  This will be a great feature.  I haven't fully understood everything but I think I am getting the picture.  Could you just give a general description of how you might have multiple serial executors?  Is this from multiple user threads?  Or is this becuase the CPU executor and the I/O executor need to be two distinct executors for some reason?


So there's two ways to end up with multiple executors - firstly there's lots of code that assumes that while an executor is running a cpu based task, it can fire and wait for i/o calls. In order to support this kind of waiting in CPU processing code, we still have to keep track of things in multiple executors.

There's also I think some stuff in other things which use multiple executors to run things in parallel whilst keeping each set of executions neatly serialised. And I think things that use nested executors. I found it easier to keep the multiple executors in existence logically, even if in reality there is only ever one thing happening at any time. I did try putting everything in a single executor and a lot of tests failed, so I figured it made more sense to keep the logical structure.

The only other difference is that instead of waiting on a condition when a serial executor is held up by I/o or whatever, it instead loops through all other executors to see if they have any tasks that need running.

Because of that, the concept of a current executor, i.e. the innermost executor being called is used instead of the thread id to detect if a particular executor is running the currently executing task. 

The other potential advantage of keeping the logical structure of things like this is that in future, there is potential for support for asynchronous i/o even in systems without threading - e.g. in webassembly there is an extension (currently experimental only, but standards track ) which enables creating and waiting on JavaScript futures. With that, whilst you still often don't have threads, you can do i/o in parallel with computation. That could integrate with this code with minimal modifications I think.

cpp/src/arrow/util/thread_pool.h
```diff
@@ -393,8 +420,23 @@ class ARROW_EXPORT SerialExecutor : public Executor {
     RunLoop();
     return final_fut;
   }
+
+#ifndef ARROW_ENABLE_THREADING
+  // we have to run tasks from all live executors
+  // during RunLoop if we don't have threading
+  static std::unordered_set<SerialExecutor*> all_executors;
```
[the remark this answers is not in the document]
joemarshall: Because CPU processing can wait for i/o. And in theory (not sure but I think in practice in acero) people can create multiple executors with dependencies across each other, which should still work in this case, just with things in other executors being run when one executor waits for the other.

cpp/src/arrow/util/thread_pool.h
```diff
@@ -194,6 +196,11 @@ class ARROW_EXPORT Executor {
   // Executor. Returns false if this Executor does not support this property.
   virtual bool OwnsThisThread() { return false; }
 
+  // Return true if this is the current executor being called
+  // n.b. this defaults to just calling OwnsThisThread
+  // unless the threadpool is disabled
+  virtual bool IsCurrentExecutor() { return OwnsThisThread(); }
```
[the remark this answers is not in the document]
joemarshall: It's because in threading disabled mode, there can be multiple nested executors running, with one executor waiting on another. OwnsThisThread assumes that each executor has a thread, and uses the thread id to check if the code is running under one particular executor, which doesn't work without threads.

cpp/src/arrow/util/thread_pool.cc
```diff
@@ -60,11 +61,33 @@ struct SerialExecutor::State {
...
 };
 
-SerialExecutor::SerialExecutor() : state_(std::make_shared<State>()) {}
+#ifndef ARROW_ENABLE_THREADING
+// list of all SerialExecutor objects - as we need to run tasks from all pools at once in
+// Run()
+std::unordered_set<SerialExecutor*> SerialExecutor::all_executors;
+SerialExecutor* SerialExecutor::current_executor = NULL;
+SerialExecutor* SerialExecutor::GetCurrentExecutor() { return current_executor; }
+SerialExecutor* SerialExecutor::last_called_executor = NULL;
```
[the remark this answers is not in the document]
joemarshall: Done (and commented on what the variables mean)

cpp/src/arrow/testing/gtest_util.cc
```diff
@@ -1036,7 +1052,60 @@ class GatingTask::Impl : public std::enable_shared_from_this<GatingTask::Impl> {
     return unlocked_future_;
   }
 
+  void _WaitForEndOrUnlocked(std::chrono::time_point<std::chrono::steady_clock> end_time,
```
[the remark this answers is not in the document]
joemarshall: because my brain was in python mode, sorry. fixed.

cpp/src/arrow/util/future.cc
```diff
@@ -149,17 +150,39 @@ class ConcreteFutureImpl : public FutureImpl {
...
+    auto start = std::chrono::steady_clock::now();
+    std::chrono::duration<double> fsec = std::chrono::duration<double>(seconds);
+    while (std::chrono::steady_clock::now() - start < fsec) {
+      // run one task then check time
+      if (IsFutureFinished(state_)) {
+        return true;
+      }
+      arrow::internal::SerialExecutor::RunTasksOnAllExecutors(true);
+    }
+    return IsFutureFinished(state_);
```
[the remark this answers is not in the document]
joemarshall: I think this is the only time it turns up outside test specific code though. I could pull it the equivalent (busywait) from gtest_util into threadpool or somewhere instead, but I don't know if it is worth it?

jorisvandenbossche:
@github-actions crossbow submit test-ubuntu-22.04-cpp-no-threading

westonpace (commented):
Apologies for the delay.  Just a few minor suggestions and I think we can get this in.

cpp/src/arrow/util/thread_pool.cc
```diff
@@ -183,6 +267,110 @@ void SerialExecutor::RunLoop() {
...
+  bool run_task = true;
+  bool keep_going = true;
+  while (keep_going) {
+    run_task = false;
+    keep_going = false;
+    for (auto it = globalState->all_executors.begin();
+         it != globalState->all_executors.end(); ++it) {
+      if (globalState->last_called_executor != NULL) {
+        // always rerun loop if we have a last_called_executor, otherwise
+        // we drop out before everything is
```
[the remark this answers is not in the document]
joemarshall: oops

cpp/src/arrow/util/thread_pool.cc
```diff
@@ -183,6 +267,110 @@ void SerialExecutor::RunLoop() {
...
+          // now we can start running things
+          globalState->last_called_executor = NULL;
+        }
+        // skip until after we have seen the last executor we called
+        // so that we do things nicely in turn
+        continue;
+      }
+      SerialExecutor* exe = *it;
+      // don't make reentrant calls inside a serialexecutor
+      // or more than the number of concurrent tasks set on a threadpool
```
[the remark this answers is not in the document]
joemarshall: fixed

cpp/src/arrow/util/future.cc
```diff
@@ -149,17 +150,39 @@ class ConcreteFutureImpl : public FutureImpl {
...
+#ifdef ARROW_ENABLE_THREADING
     std::unique_lock<std::mutex> lock(mutex_);
 
     cv_.wait(lock, [this] { return IsFutureFinished(state_); });
+#else
+    while (true) {
+      if (IsFutureFinished(state_)) {
+        return;
+      }
+      arrow::internal::SerialExecutor::RunTasksOnAllExecutors(true);
```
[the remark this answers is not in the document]
joemarshall: I've put a warning in if it waits in here too long

joemarshall:
@westonpace sorry, I went off sick for a few days, and only just caught up with this. I've think I've fixed all the broken things / missing comments now.

I left those tests in, just because I felt they were useful as examples of what does and doesn't work in single threaded mode, Feel free to remove if you think that's a bad idea.

joemarshall:
@westonpace Fixed merge conflicts with recent changes (just to bloom filter testing) and merged up to date. hopefully we're ready to merge now.

westonpace (approved):
Threading is tricky.  I wouldn't be surprised if a deadlock of some kind (or several) is still existing in the single-threaded implementation somewhere.  However, I think we've reached the point where we should go ahead and merge this as an experimental feature and start trying it out.

I think the changes to the thread-enabled (e.g. the default) path in this PR are minimal and safe.  So I think we should proceed with merging.

Thank you very much for your efforts putting this together @joemarshall

westonpace:
@kou any concerns?

cpp/src/arrow/testing/gtest_util.cc
```diff
@@ -1036,7 +1051,60 @@ class GatingTask::Impl : public std::enable_shared_from_this<GatingTask::Impl> {
     return unlocked_future_;
   }
 
+  void WaitForEndOrUnlocked(std::chrono::time_point<std::chrono::steady_clock> end_time,
+                            arrow::internal::Executor* executor, Future<> future) {
+    if (unlocked_) {
+      num_finished_++;
+      future.MarkFinished(Status::OK());
+      return;
```
kou, line 1059:
```suggestion
```

cpp/src/arrow/testing/gtest_util.cc
```diff
@@ -1036,7 +1051,60 @@ class GatingTask::Impl : public std::enable_shared_from_this<GatingTask::Impl> {
...
+    } else {
+      if (std::chrono::steady_clock::now() > end_time) {
+        num_finished_++;
+        future.MarkFinished(
+            Status::Invalid("Task unlock never happened - if threads are disabled you "
+                            "can't wait on gatedtask"));
+        return;
```
kou, line 1066:
```suggestion
```

cpp/src/arrow/testing/gtest_util.cc
kou: Or:

```cpp
if (unlocked_) {
  ...
  return;
}

if (std::chrono::steady_clock::now() > end_time) {
  ...
  return;
}

SleepABit();
...
```

cpp/src/arrow/testing/gtest_util.cc
```diff
@@ -1036,7 +1051,60 @@ class GatingTask::Impl : public std::enable_shared_from_this<GatingTask::Impl> {
...
+    num_running_++;
+    // post the unlock check as a separate task
+    // otherwise we'll never let anything else run
+    // so nothing can unlock us
+    using Clock = std::chrono::steady_clock;
+    using DurationDouble = std::chrono::duration<double>;
+    using DurationClock = std::chrono::steady_clock::duration;
+
+    auto start_time = Clock::now();
+    std::chrono::duration<double> secs_left =
```
kou, line 1089:
```suggestion
    auto secs_left =
```

cpp/src/arrow/testing/gtest_util.cc
```diff
@@ -1036,7 +1051,60 @@ class GatingTask::Impl : public std::enable_shared_from_this<GatingTask::Impl> {
...
+        std::chrono::duration<double>(timeout_seconds_);
+    std::chrono::time_point<Clock> end_time =
```
kou, line 1091: Can we use `auto` here?

cpp/src/arrow/testing/gtest_util.cc
```diff
@@ -1036,7 +1051,60 @@ class GatingTask::Impl : public std::enable_shared_from_this<GatingTask::Impl> {
...
+        std::chrono::time_point_cast<DurationClock, Clock, DurationDouble>(start_time +
+                                                                           secs_left);
+    arrow::internal::Executor* executor = arrow::internal::GetCpuThreadPool();
```
kou, line 1094:
```suggestion
    auto executor = arrow::internal::GetCpuThreadPool();
```

cpp/src/arrow/testing/gtest_util.cc
```diff
@@ -1036,7 +1051,60 @@ class GatingTask::Impl : public std::enable_shared_from_this<GatingTask::Impl> {
...
+    Future<> future = Future<>::Make();
```
kou, line 1095:
```suggestion
    auto future = Future<>::Make();
```

cpp/src/arrow/util/async_generator_test.cc
kou: @joemarshall Could you check this?

cpp/src/arrow/util/async_generator_test.cc
kou: @joemarshall Could you check this?

cpp/src/arrow/util/future.cc
```diff
@@ -149,17 +150,47 @@ class ConcreteFutureImpl : public FutureImpl {
...
   bool DoWait(double seconds) {
+#ifdef ARROW_ENABLE_THREADING
     std::unique_lock<std::mutex> lock(mutex_);
 
     cv_.wait_for(lock, std::chrono::duration<double>(seconds),
                  [this] { return IsFutureFinished(state_); });
     return IsFutureFinished(state_);
+#else
+    auto start = std::chrono::steady_clock::now();
+    std::chrono::duration<double> fsec = std::chrono::duration<double>(seconds);
```
kou, line 184:
```suggestion
    auto fsec = std::chrono::duration<double>(seconds);
```

cpp/src/arrow/util/thread_pool.cc
```diff
@@ -183,6 +267,110 @@ void SerialExecutor::RunLoop() {
...
+            globalState->last_called_executor == *it) {
+          // found the last one (or it doesn't exist ih the set any more)
+          // now we can start running things
+          globalState->last_called_executor = NULL;
+        }
+        // skip until after we have seen the last executor we called
+        // so that we do things nicely in turn
+        continue;
+      }
+      SerialExecutor* exe = *it;
```
kou, line 299:
```suggestion
      auto exe = *it;
```

kou (commented):
Sorry. I don't still have enough time to review this carefully...
I trust @westonpace's judgement.

Could you update the PR description to reflect the latest changes? I can read and comment it.

kou:
> The description at the top of the PR is up to date now (and has a brief changelog of what changed from initial PR)

Could you replace the initial description directly instead of adding a new "Updated since initial PR:" section?
The latest decision is the most important thing for us.
If we want to see the history, we can use GitHub's history view UI. (See the "edited" pull down in the top of the description area.)

joemarshall:
> > The description at the top of the PR is up to date now (and has a brief changelog of what changed from initial PR)
> 
> Could you replace the initial description directly instead of adding a new "Updated since initial PR:" section? The latest decision is the most important thing for us. If we want to see the history, we can use GitHub's history view UI. (See the "edited" pull down in the top of the description area.)

Okay, it's up to date.

joemarshall:
@kou think we're good to merge now?

cpp/src/arrow/util/async_generator_test.cc
```diff
@@ -1521,6 +1532,7 @@ TEST(TestAsyncUtil, ReadaheadFailedWaitForInFlight) {
   auto in_flight_gating_task = GatingTask::Make();
   auto source = [&]() -> Future<TestInt> {
     auto count = counter++;
+
```
kou, line 1535:
```suggestion
```

kou (approved):
+1

The failure is https://github.com/apache/arrow/issues/37067 .

Issue #35538, opened 2023-05-10
[C++] Arrow 12 cpp doesn't compile with protobuf 23.0 


smahapatra1:
### Describe the bug, including details regarding any error messages, version, and platform.

cpp/src/arrow/engine/substrait/serde.cc:27:10: fatal error: google/protobuf/stubs/status.h: No such file or directory
   27 | #include <google/protobuf/stubs/status.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

### Component(s)

C++, FlightRPC

westonpace:
Issue resolved by pull request 35673
https://github.com/apache/arrow/pull/35673

PR #35673, opened 2023-05-18
GH-35538: [C++] Remove unnecessary status.h include from protobuf

felipecrv:
### Rationale for this change

Newer versions of protobuf use `absl::Status` instead of `google::protobuf::util::Status`. The status variables in this file are type-annotated with `auto` and we can count on the header that declares the status-returning function we are calling to have included the header that defines `Status`.

Removing an include of a protobuf header that don't exist in later versions of the library.

### Are these changes tested?

I haven't tried building with protobuf 23, but I can be sure this specifically fixes the problem reported in #35538.


felipecrv:
@westonpace

pitrou:

westonpace (approved):
The commit that introduced this include did so via IWYU.  My guess is IWYU maybe got confused between arrow::Status and protobuf Status.

PR #35674, opened 2023-05-18
GH-35359: [C++] FixedSizeListArray.flatten() errors if all elements are null

### LINKED ISSUE #35359: [C++] FixedSizeListArray.flatten() raises if all elements are null
Windows 10
python=3.11.2
pyarrow=11.0.0

```python
import pyarrow as pa

array = pa.array([None, None], type=pa.list_(pa.int32()))
array.flatten()

# Output:
<pyarrow.lib.Int32Array object at 0x000001EBA976FDC0>
[]
```

Same array but with fixed size raises `ArrowInvalid`
```python
array = pa.array([None, None], type=pa.list_(pa.int32(), list_size=1))
array.flatten()

# Output
ArrowInvalid                              Traceback (most recent call last)
Cell In[69], line 1
----> 1 array.flatten()

File ~\AppData\Local\miniconda3\envs\dev311_1\Lib\site-packages\pyarrow\array.pxi:1871, in pyarrow.lib.BaseListArray.flatten()

File ~\AppData\Local\miniconda3\envs\dev311_1\Lib\site-packages\pyarrow\compute.py:239, in _make_generic_wrapper.<locals>.wrapper(memory_pool, *args)
    237 if args and isinstance(args[0], Expression):
    238     return Expression._call(func_name, list(args))
--> 239 return func.call(args, None, memory_pool)

File ~\AppData\Local\miniconda3\envs\dev311_1\Lib\site-packages\pyarrow\_compute.pyx:355, in pyarrow._compute.Function.call()

File ~\AppData\Local\miniconda3\envs\dev311_1\Lib\site-packages\pyarrow\error.pxi:144, in pyarrow.lib.pyarrow_internal_check_status()

File ~\AppData\Local\miniconda3\envs\dev311_1\Lib\site-packages\pyarrow\error.pxi:100, in pyarrow.lib.check_status()

ArrowInvalid: Must pass at least one array
```

Expected behavior: should return an empty array or, since fixed size lists must allocate memory for null slots, should return an array of two nulls (equal to `array.values`)

This issue may be related to the fix implemented for this issue: https://issues.apache.org/jira/browse/ARROW-16174

### Component(s)

Python

zeroshade:
Added a check inside of `FlattenListArray` before the call to `Concatenate` so that it will return an empty array of the appropriate type if there are no non null fragments to return. This avoids receiving the error from concatenate that you must pass at least one array to it.

### Are these changes tested?
Unit tests are added.

* Closes: #35359

### DIFF
--- a/cpp/src/arrow/array/array_nested.cc
+++ b/cpp/src/arrow/array/array_nested.cc
@@ -177,6 +177,8 @@ Result<std::shared_ptr<Array>> FlattenListArray(const ListArrayT& list_array,
   // Final attempt to avoid invoking Concatenate().
   if (non_null_fragments.size() == 1) {
     return non_null_fragments[0];
+  } else if (non_null_fragments.size() == 0) {
+    return MakeEmptyArray(value_array->type(), memory_pool);
   }
 
   return Concatenate(non_null_fragments, memory_pool);
--- a/cpp/src/arrow/array/array_nested.cc
+++ b/cpp/src/arrow/array/array_nested.cc
@@ -156,6 +156,12 @@ Result<std::shared_ptr<Array>> FlattenListArray(const ListArrayT& list_array,
                                  list_array.value_offset(list_array_length));
   }
 
+  // Second shortcut: if the list array is *all* nulls, then just return
+  // an empty array.
+  if (list_array.null_count() == list_array.length()) {
+    return MakeEmptyArray(value_array->type(), memory_pool);
+  }
+
   // The ListArray contains nulls: there may be a non-empty sub-list behind
   // a null and it must not be contained in the result.
   std::vector<std::shared_ptr<Array>> non_null_fragments;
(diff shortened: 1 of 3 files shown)


cpp/src/arrow/array/array_nested.cc
```diff
@@ -177,6 +177,8 @@ Result<std::shared_ptr<Array>> FlattenListArray(const ListArrayT& list_array,
   // Final attempt to avoid invoking Concatenate().
   if (non_null_fragments.size() == 1) {
     return non_null_fragments[0];
+  } else if (non_null_fragments.size() == 0) {
+    return MakeEmptyArray(value_array->type(), memory_pool);
```
felipecrv, line 181: Might also be worth it to add and `} else if (list_array.null_count() == list_array->length()) {` to the `if (list_array.null_count() == 0)` above and avoid the calculation of fragments altogether when we are lucky and `null_count` is already calculated for us.

felipecrv: You can force the null calculation in the tests by calling `GetNullCount()`.

zeroshade: Done

felipecrv: But I think you need to keep both checks. `null_count()` can return `kUnknownNullCount` here. :sweat:

felipecrv: Thanks!

felipecrv (approved):

pitrou (approved):
+1, thanks @zeroshade and thanks @felipecrv for the review

Issue #35581, opened 2023-05-13
[C++] ArraySpan::FillFromScalar is unsafe


bkietz:
### Describe the bug, including details regarding any error messages, version, and platform.

`ArraySpan::FillFromScalar` can store pointers into the structure itself (specifically, into `ArraySpan::scratch_space`) which produces an ArraySpan which easily be unsafely moved and copied:

```c++
ArraySpan span;
span.FillFromScalar(scalar);

UseSpan(span); // Fine; the original span is still alive.

return span; // Undefined Behavior; the returned copy views
             // a stack variable whose lifetime has ended.
```

The capability to view a Scalar as an ArraySpan can be preserved and made safer by restricting access to the span to an explicitly delineated scope:

```c++
ArraySpan::FillFromScalar(scalar, [&](ArraySpan span) {
  UseSpan(span); // Fine; the viewed scratch space is alive inside this scope.
});
```

This has the pleasant side effect of reducing the size of ArraySpan by 16 bytes.

### Component(s)

C++

felipecrv:
```cpp
return span; // Undefined Behavior; the returned copy views
             // a stack variable whose lifetime has ended.
```

We can check if the buffers are pointing to scratch space and re-wire them if they are pointing to the previous value.

felipecrv:
```cpp
ArraySpan span;
span.FillFromScalar(scalar);

UseSpan(span); // Fine; the original span is still alive.

return span; // Undefined Behavior; the returned copy views
             // a stack variable whose lifetime has ended.
 ```
 
 Due to RVO, the returned `ArraySpan` is constructed at the caller's stack and this is not biting us, I guess. I think this is a matter of always using destination-passing style when dealing with `ArraySpan` and never copying or moving it.
 
 Copy and Move constructors haven't been `= delete;`d in the class, I wonder if they could be still. I think we are allowed to do so as it's laveled `EXPERIMENTAL`.

**Comment by bkietz** [comment]
This is not biting us currently because AFAICT the only place we use FillFromScalar is https://github.com/apache/arrow/blob/cd6e2a4d2b9373b942da18b4cc82cb41431764d9/cpp/src/arrow/compute/exec.cc#L319-L330 where they are in a `std::vector` from which they are never moved. This is subtle too: as long as the original ArraySpan remains valid copies will be valid too, so

```c++
ArraySpan Thing::get(int i) { return vector_of_spans_[i]; }
```

Might not segfault if the `Thing` happens to be kept alive.

[...]

PR #35677, opened 2023-05-18
GH-35675 [C++] Don't copy the ArraySpan into the REE ArraySpan

### LINKED ISSUE #35675: [C++] Copying a REE ArraySpan that has buffers using scratch buffers could lead to memory errors

More details in #35581. To avoid this risk, I want to also avoid copying `ArraySpan` into `RunEndEncodedArraySpan` when it's instantiated.

### Component(s)

C++

felipecrv:
### Rationale for this change

Copying ArraySpan is not without issues (see #35581), and this change seems good by itself as it makes it easier for the compiler to SROA [1] the whole `RunEndEncodedArraySpan` class.

[1] https://www.llvm.org/docs/Passes.html#sroa-scalar-replacement-of-aggregates

 - Make `array_span` private in `ree_util::RunEndEncodedArraySpan`
 - Allow construction of  `ree_util::RunEndEncodedArraySpan` with separate `offset` and `length`
 - Change `RunEndEncodedBuilder` to avoid a copy of the `ArraySpan` being iterated over
 - Prevent instantiation based on implicit conversion from `ArrayData` to `ArraySpan`
 
### Are these changes tested?

Yes, by the existing unit tests of the various uses of `RunEndEncodedArraySpan`.

No meaningful changes other than some small tweaks in the interface of a recently added class.

* Closes: #35675

### DIFF
--- a/cpp/src/arrow/util/ree_util.h
+++ b/cpp/src/arrow/util/ree_util.h
@@ -227,8 +227,9 @@ class RunEndEncodedArraySpan {
     int64_t physical_pos_;
   };
 
-  explicit RunEndEncodedArraySpan(const ArrayData& data)
-      : RunEndEncodedArraySpan(ArraySpan{data}) {}
+  // Prevent implicit ArrayData -> ArraySpan conversion in
+  // RunEndEncodedArraySpan instantiation.
+  RunEndEncodedArraySpan(const ArrayData& data) = delete;
 
   /// \brief Construct a RunEndEncodedArraySpan from an ArraySpan and new
   /// absolute offset and length.
--- a/cpp/src/arrow/util/ree_util.h
+++ b/cpp/src/arrow/util/ree_util.h
@@ -230,17 +230,35 @@ class RunEndEncodedArraySpan {
   explicit RunEndEncodedArraySpan(const ArrayData& data)
       : RunEndEncodedArraySpan(ArraySpan{data}) {}
 
-  explicit RunEndEncodedArraySpan(const ArraySpan& array_span)
-      : array_span{array_span}, run_ends_(RunEnds<RunEndCType>(array_span)) {
-    assert(array_span.type->id() == Type::RUN_END_ENCODED);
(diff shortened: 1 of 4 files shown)


cpp/src/arrow/util/ree_util.h
```diff
@@ -227,20 +227,39 @@ class RunEndEncodedArraySpan {
...
-    assert(array_span.type->id() == Type::RUN_END_ENCODED);
+  /// \brief Construct a RunEndEncodedArraySpan from an ArraySpan and new
+  /// absolute offset and length.
+  ///
+  /// RunEndEncodedArraySpan{span, off, len} is equivalent to:
+  ///
+  ///   span.SetSlice(off, len);
+  ///   RunEndEncodedArraySpan{span}
+  ///
+  /// ArraySpan::SetSlice() updates the null_count to kUnknownNullCount, but
```
bkietz, line 242: Could this be fixed by making SetSlice() update null_count differently? If the original span has zero nulls, then shouldn't the slice also have zero nulls? See for example https://github.com/apache/arrow/pull/35628/files#diff-c5d8f67626f30c66a79d1f677ee81264a9b044280e942cfa846e07306b25365bR478

felipecrv: That improvement seems unrelated to this change and affects code that is much more widely used. Could you create an issue for that? My comment here is just an informal proof that the equivalence here holds even though I don't perform all the operations that `SetSlice` does based on my class only taking ArraySpans of REEs.

bkietz (approved):
LGTM

felipecrv:
@bkietz I rebased and force-pushed to see if CI passes now.

felipecrv:
One test is timing out. Can someone re-trigger it instead of me pushing again?  cc @bkietz @pitrou

felipecrv:
@pitrou this is ready for merging.

pitrou (approved):

PR #35679, opened 2023-05-18
GH-32276: [C++][FlightRPC] Align buffers from Flight

### LINKED ISSUE #32276: [C++][FlightRPC] Flight generates misaligned buffers

lidavidm:
### Rationale for this change

Ensure buffers from Flight are at least 8-byte aligned to meet general expectations of downstream users.

Manually align buffers after deserialization.

### Are these changes tested?

Yes

Flight data will be aligned, at the cost of an additional copy.
* Closes: #32276

### DIFF
--- a/cpp/src/arrow/flight/transport/grpc/serialization_internal.cc
+++ b/cpp/src/arrow/flight/transport/grpc/serialization_internal.cc
@@ -25,6 +25,7 @@
 #include <vector>
 
 #include "arrow/flight/platform.h"
+#include "arrow/type_fwd.h"
 
 #if defined(_MSC_VER)
 #pragma warning(push)
@@ -55,6 +56,7 @@
 #include "arrow/flight/transport/grpc/util_internal.h"
 #include "arrow/ipc/message.h"
 #include "arrow/ipc/writer.h"
+#include "arrow/util/align_util.h"
 #include "arrow/util/bit_util.h"
 #include "arrow/util/logging.h"
 
@@ -379,6 +381,14 @@ static const uint8_t kPaddingBytes[8] = {0, 0, 0, 0, 0, 0, 0, 0};
         if (!ReadBytesZeroCopy(wrapped_buffer, &pb_stream, &out->body)) {
           return ::grpc::Status(::grpc::StatusCode::INTERNAL,
                                 "Unable to read FlightData body");
+        }
+        // XXX: due to where we sit, we can't use a custom allocator
+        // XXX: any error here will likely crash or hang gRPC!
+        auto status =
+            util::EnsureAlignment(std::move(out->body), 64, default_memory_pool())
+                .Value(&out->body);
+        if (!status.ok()) {
+          return {::grpc::StatusCode::INTERNAL, status.ToString()};
         }
       } break;
       default:
--- a/cpp/src/arrow/flight/test_definitions.cc
+++ b/cpp/src/arrow/flight/test_definitions.cc
@@ -19,6 +19,8 @@
 
 #include <chrono>
 
+#include <gmock/gmock.h>
+
 #include "arrow/array/array_base.h"
 #include "arrow/array/array_dict.h"
 #include "arrow/array/util.h"
@@ -26,6 +28,8 @@
 #include "arrow/flight/test_util.h"
 #include "arrow/table.h"
 #include "arrow/testing/generator.h"
+#include "arrow/testing/gtest_util.h"
+#include "arrow/util/align_util.h"
 #include "arrow/util/checked_cast.h"
 #include "arrow/util/config.h"
 #include "arrow/util/logging.h"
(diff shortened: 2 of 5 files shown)


lidavidm:
I inlined some stuff from @westonpace's PR but I'd rather wait and be able to use those utilities directly. That said it demonstrates the problem and the solution. 

Given more time we might be able to avoid this, but I'd need to dig deeper into Protobuf & do some benchmarking. (In particular: instead of going gRPC slices -> CodedInputStream -> Arrow buffers, I'd rather directly go from gRPC slices -> Arrow.)

cpp/src/arrow/flight/transport/grpc/serialization_internal.cc
```diff
@@ -380,6 +382,14 @@ ::grpc::Status FlightDataDeserialize(ByteBuffer* buffer,
           return ::grpc::Status(::grpc::StatusCode::INTERNAL,
                                 "Unable to read FlightData body");
         }
+        // XXX: due to where we sit, we can't use a custom allocator
+        // XXX: any error here will likely crash or hang gRPC!
+        auto status =
+            util::EnsureAlignment(std::move(out->body), 64, default_memory_pool())
```
pitrou, line 388: Isn't 64 a bit heavy-handed? Basically, and assuming the distribution of gRPC slice alignments is uniform, we will reallocate almost all incoming buffers?

cpp/src/arrow/flight/transport/grpc/serialization_internal.cc
```diff
@@ -380,6 +382,14 @@ ::grpc::Status FlightDataDeserialize(ByteBuffer* buffer,
           return ::grpc::Status(::grpc::StatusCode::INTERNAL,
                                 "Unable to read FlightData body");
         }
+        // XXX: due to where we sit, we can't use a custom allocator
+        // XXX: any error here will likely crash or hang gRPC!
```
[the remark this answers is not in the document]
lidavidm: This is true of any error returned here. Last I checked, gRPC doesn't actually handle errors despite this being a fallible API, instead choosing to crash.

lidavidm:
Given the nature of the gRPC API there's not a great way to customize it at runtime. If someone complains about this in the future we can point them here then. Possibly we can make it a compile-time toggle, though that's effectively useless to most users.

@westonpace @rtpsw were the most recent to mention this. If there's not a real desire for this then I'll close the PR.

pitrou:
This could be settable using an env var. But I don't think we should realign buffers by default. This can decrease performance and increase memory fragmentation for unsuspecting users.

westonpace:
Now that we have a workaround enabled in Acero I don't have any urgent need for this.  I'm fine with waiting for it to become a problem (and it may never do so).

orlp:
@pitrou 

> But I don't think we should realign buffers by default. 

The [arrow standard requires it](https://arrow.apache.org/docs/format/Columnar.html#buffer-alignment-and-padding).

> Implementations are recommended to allocate memory on aligned addresses (multiple of 8- or 64-bytes) and pad (overallocate) to a length that is a multiple of 8 or 64 bytes. When serializing Arrow data for interprocess communication, these alignment and padding requirements are **enforced**.

cpp/src/arrow/flight/transport/grpc/serialization_internal.cc
EnricoMi: This should be configurable as not all code using the C++ flight client is sensitive to alignment.

PR #35690, opened 2023-05-19
GH-35684: [Go][Parquet] Fix nil dereference with nil list array

### LINKED ISSUE #35684: [Go][Parquet] panic for listof(types) if null

parquet writer will panic on any of the following types if the value null:

```
		{Name: "f1", Type: arrow.ListOf(arrow.FixedWidthTypes.Date32)},
		{Name: "f2", Type: arrow.ListOf(arrow.FixedWidthTypes.Date64)},
		{Name: "f3", Type: arrow.ListOf(arrow.FixedWidthTypes.Timestamp_us)},
		{Name: "f4", Type: arrow.ListOf(arrow.FixedWidthTypes.Timestamp_ms)},
		{Name: "f5", Type: arrow.ListOf(arrow.FixedWidthTypes.Time32ms)},
		{Name: "f6", Type: arrow.ListOf(arrow.FixedWidthTypes.Time64ns)},
		{Name: "f7", Type: arrow.ListOf(arrow.FixedWidthTypes.Time64us)},
```

### Component(s)

Go

Reproduction is here - https://github.com/apache/arrow/pull/35685

Im not sure though how to fix it as the parquet code is quite complex. cc @zeroshade

zeroshade:
### Rationale for this change
If a column to be written to parquet is a list array with only nulls in it, then the child will be empty and potentially have nil buffers. This case isn't checked in the `writeDenseArrow` function, so we blindly attempt to grab the bytes from the buffers which might be nil. 

Adding a check of `leafArr.Len()` at the top of `writeDenseArrow`. If the leaf array's length is > 0 we can be sure that the data buffer is non-nil. So we can just bail if the leaf array is empty, nothing to write.

### Are these changes tested?
Unit test is added.

* Closes: #35684

### DIFF
--- a/go/parquet/pqarrow/encode_arrow.go
+++ b/go/parquet/pqarrow/encode_arrow.go
@@ -246,10 +246,6 @@ type binary64arr interface {
 }
 
 func writeDenseArrow(ctx *arrowWriteContext, cw file.ColumnChunkWriter, leafArr arrow.Array, defLevels, repLevels []int16, maybeParentNulls bool) (err error) {
-	if leafArr.Len() == 0 {
-		return
-	}
-
 	if leafArr.DataType().ID() == arrow.EXTENSION {
 		extensionArray := leafArr.(array.ExtensionArray)
 		// Replace leafArr with its underlying storage array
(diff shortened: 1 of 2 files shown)


zeroshade:
CC @yevgenypats please give this a try and see that it fixes your issue.

felipecrv (approved):

go/parquet/pqarrow/encode_arrow.go
```diff
@@ -246,6 +246,10 @@ type binary64arr interface {
 }
 
 func writeDenseArrow(ctx *arrowWriteContext, cw file.ColumnChunkWriter, leafArr arrow.Array, defLevels, repLevels []int16, maybeParentNulls bool) (err error) {
+	if leafArr.Len() == 0 {
```
benibus, line 249: Can't comment on it directly since it's not in the diff, but it's not immediately obvious to me why we'd handle this case differently: https://github.com/apache/arrow/blob/c8363fa209fed4dfa30a0eee70283d62fc6f1627/go/parquet/pqarrow/encode_arrow.go#L270-L275

Also, it seems like that branch may never be taken now? Unless it's only there to cover an edge-case where `leafArr` was reassigned to an empty storage array after the initial `Len() == 0` check: https://github.com/apache/arrow/blob/c8363fa209fed4dfa30a0eee70283d62fc6f1627/go/parquet/pqarrow/encode_arrow.go#L253-L257

benibus (commented):
Seems fairly straightforward! (although I'm not sure what the call is on that test failure). Just one comment/question for now.

go/parquet/pqarrow/encode_arrow.go
zeroshade: Turns out you're right and it ended up causing a different test to fail. So it looks like i just needed to add the checks for whether or not the buffer is nil *sigh*. I've updated this

yevgenypats (approved):

yevgenypats:
Thanks looks good to me!

PR #35691, opened 2023-05-19
GH-34785: [C++][Parquet] Skeleton for Parquet bloom filter writer

### LINKED ISSUE #34785: [C++][Parquet] Allow writing BloomFilter for specific column

In parquet spec, we can have a BloomFilter for a column in one RowGroup. Writer can set BloomFilter when writing columns. The setting patterns would be like `{write_bloom, ndv}` in column properties.

### Component(s)

C++, Parquet

mapleFU:
### Rationale for this change

Currently we allow reading bloom filter for specific column and rowgroup, now this patch allow it writing BF.

This patch is just a skeleton. If reviewer thinks interface would be OK, I'll go on and add testing.

Allow writing bf:

* [x] Add WriterProperties config for writing bloom filter, including bf and (per-rowgroup) ndv estimation.
* [x] Add BloomFilterWriter for parquet
* [x] From FileSerializer to ColumnWriter, adding bloomfilter
* [x] Ensure Bloom Filter info is written to the file
* [ ] Testing logic above

### Are these changes tested?

Currently not, after reviewers think it's ok, I'll continue testing it.

User can create Bloom Filter in parquet with C++ api

* Closes: #34785

### DIFF
--- a/cpp/src/parquet/bloom_filter.h
+++ b/cpp/src/parquet/bloom_filter.h
@@ -101,6 +101,12 @@ class PARQUET_EXPORT BloomFilter {
   /// @return hash result.
   virtual uint64_t Hash(const FLBA* value, uint32_t len) const = 0;
 
+  // Variant of const pointer argument to facilitate template
+  uint64_t Hash(const int32_t* value) const { return Hash(*value); }
+  uint64_t Hash(const int64_t* value) const { return Hash(*value); }
+  uint64_t Hash(const float* value) const { return Hash(*value); }
+  uint64_t Hash(const double* value) const { return Hash(*value); }
+
   virtual ~BloomFilter() {}
 
  protected:
--- a/cpp/src/parquet/bloom_filter_writer.cc
+++ b/cpp/src/parquet/bloom_filter_writer.cc
@@ -27,13 +27,6 @@ bool RowGroupBloomFilterReference::GetBloomFilterOffsets(
   return false;
 }
 
-void BloomFilterWriter::DropRowGroupBloomFilter(
-    const std::shared_ptr<schema::ColumnPath>& col_path) {
-  if (properties_.bloom_filter_enabled(col_path)) {
-    row_group_bloom_filters_.back().erase(col_path->ToDotString());
-  }
-}
-
 void BloomFilterWriter::AppendRowGroup() { row_group_bloom_filters_.emplace_back(); }
 
 void BloomFilterWriter::WriteTo(::arrow::io::OutputStream* sink,
@@ -44,7 +37,6 @@ void BloomFilterWriter::WriteTo(::arrow::io::OutputStream* sink,
     return;
   }
 
-  size_t total_serialized_count = 0;
   for (const auto& row_group_bloom_filters : row_group_bloom_filters_) {
     reference_.AppendRowGroup();
 
@@ -70,7 +62,6 @@ void BloomFilterWriter::WriteTo(::arrow::io::OutputStream* sink,
         iter->second->WriteTo(sink);
         PARQUET_ASSIGN_OR_THROW(int64_t pos, sink->Tell());
         reference_.AddBloomFilter(column_id, offset, static_cast<int32_t>(pos - offset));
-        total_serialized_count++;
       }
     }
   }
--- a/cpp/src/parquet/bloom_filter_writer.cc
+++ b/cpp/src/parquet/bloom_filter_writer.cc
@@ -1,0 +1,98 @@
+#include "bloom_filter_writer.h"
+
+#include "arrow/io/interfaces.h"
+
+#include "parquet/encryption/encryption.h"
+#include "parquet/encryption/internal_file_encryptor.h"
+#include "parquet/exception.h"
+#include "parquet/properties.h"
+
+namespace parquet {
+
+void RowGroupBloomFilterReference::AppendRowGroup() { references_.emplace_back(); }
+
+void RowGroupBloomFilterReference::AddBloomFilter(int32_t column_id, int64_t offset,
+                                                  int32_t length) {
+  DCHECK(!references_.empty());
+  references_.back().emplace(column_id, Reference{offset, length});
+}
+
+bool RowGroupBloomFilterReference::GetBloomFilterOffsets(
+    size_t row_group_ordinal, const std::map<int32_t, Reference>** out) const {
+  if (row_group_ordinal < references_.size()) {
+    *out = &references_.at(row_group_ordinal);
+    return true;
+  }
+  *out = nullptr;
+  return false;
+}
+
+void BloomFilterWriter::DropRowGroupBloomFilter(
+    const std::shared_ptr<schema::ColumnPath>& col_path) {
+  if (properties_.bloom_filter_enabled(col_path)) {
+    row_group_bloom_filters_.back().erase(col_path->ToDotString());
+  }
+}
+
+void BloomFilterWriter::AppendRowGroup() { row_group_bloom_filters_.emplace_back(); }
+
+void BloomFilterWriter::WriteTo(::arrow::io::OutputStream* sink,
+                                const SchemaDescriptor* schema,
+                                InternalFileEncryptor* encryptor) {
+  if (row_group_bloom_filters_.empty()) {
+    // Return quickly if there is no bloom filter
+    return;
+  }
+
+  size_t total_serialized_count = 0;
+  for (const auto& row_group_bloom_filters : row_group_bloom_filters_) {
+    reference_.AppendRowGroup();
+
+    // the whole row group has no bloom filter
+    if (row_group_bloom_filters.empty()) {
+      continue;
+    }
+
+    // serialize bloom filter by ascending order of column id
+    for (int32_t column_id = 0; column_id < schema->num_columns(); ++column_id) {
+      const auto column_path = schema->Column(column_id)->path()->ToDotString();
+      auto meta_encryptor =
+          encryptor ? encryptor->GetColumnMetaEncryptor(column_path) : nullptr;
+      auto data_encryptor =
+          encryptor ? encryptor->GetColumnDataEncryptor(column_path) : nullptr;
+      if (meta_encryptor || data_encryptor) {
+        ParquetException::NYI("Bloom filter encryption is not implemented");
+      }
+
+      auto iter = row_group_bloom_filters.find(column_path);
+      if (iter != row_group_bloom_filters.cend()) {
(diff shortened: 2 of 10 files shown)


wgtmac:
Thanks for working on this! I will take a look next week after returning from vacation.

cpp/src/parquet/column_writer.cc
```diff
@@ -2299,12 +2319,112 @@ Status TypedColumnWriterImpl<FLBAType>::WriteArrowDense(
   return Status::OK();
 }
 
+template <typename DType>
+void TypedColumnWriterImpl<DType>::UpdateBloomFilter(const T* values,
+                                                     int64_t num_values) {
+  if (bloom_filter_) {
+    for (int64_t i = 0; i < num_values; ++i) {
+      bloom_filter_->InsertHash(bloom_filter_->Hash(values + i));
```
pitrou, line 2327: This will be very slow in any case, because there is a double virtual indirection (first `bloom_filter_->Hash`, then `hasher_->Hash`).

I think you want to add batch methods, such as `BloomFilter::InsertHashes(const double* values, int64_t num_values)`.

pitrou:
@mapleFU I don't have time for a comprehensive review right now, I just took a very quick look.

cpp/src/parquet/column_writer.cc
mapleFU: Emmmm, I'll try to seperate a patch for batch methods

cpp/src/parquet/column_writer.cc
```diff
@@ -2299,12 +2319,112 @@ Status TypedColumnWriterImpl<FLBAType>::WriteArrowDense(
...
+template <>
+void TypedColumnWriterImpl<BooleanType>::UpdateBloomFilterSpaced(const bool*, int64_t,
+                                                                 const uint8_t*,
+                                                                 int64_t) {
+  DCHECK(bloom_filter_ == nullptr);
+}
+
+template <typename DType>
+void TypedColumnWriterImpl<DType>::UpdateBloomFilter(const ::arrow::Array& values) {
+  ParquetException::NYI("UpdateBloomFilter via arrow::Array only supports BYTE_ARRAY");
```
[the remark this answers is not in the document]
mapleFU: Yeah I'll finish it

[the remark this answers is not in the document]
wgtmac: I guess only BYTE_ARRAY type will issue this call.

cpp/src/parquet/file_writer.cc
```diff
@@ -290,14 +294,29 @@ class RowGroupSerializer : public RowGroupWriter::Contents {
...
+          bloom_filter_writer_ && properties_->bloom_filter_enabled(path)
+              ? bloom_filter_writer_->GetOrCreateBloomFilter(
+                    path,
+                    [&]() {
+                      int64_t ndv = properties_->bloom_filter_ndv(path);
+
+                      auto bloom_filter = std::make_unique<BlockSplitBloomFilter>();
+                      // TODO(mwish): add configurable ndv
+                      bloom_filter->Init(BlockSplitBloomFilter::OptimalNumOfBytes(
+                          static_cast<uint32_t>(ndv),
```
wgtmac, line 307: So here we imply the ndv cannot exceed UINT32_MAX? Then we probably need to change its type in the writer properties.

cpp/src/parquet/file_writer.cc
```diff
@@ -290,14 +294,29 @@ class RowGroupSerializer : public RowGroupWriter::Contents {
...
+                          /*fpp=*/0.05));
```
wgtmac, line 308: `fpp` worth to be a config parameter.

cpp/src/parquet/file_writer.cc
```diff
@@ -335,6 +354,7 @@ class FileSerializer : public ParquetFileWriter::Contents {
       row_group_writer_.reset();
 
       WritePageIndex();
+      WriteBloomFilter();
```
wgtmac, line 357: According to the [specs](https://github.com/apache/parquet-format/blob/master/BloomFilter.md),
- The Bloom filter data can be stored before the page indexes after all row groups.
- Or it can be stored between row groups

So `WriteBloomFilter()` need to be placed above `WritePageIndex()`

cpp/src/parquet/bloom_filter_writer.cc
```diff
@@ -0,0 +1,89 @@
...
+
+  // release memory actively
+  row_group_bloom_filters_.clear();
+}
+
+const RowGroupBloomFilterReference& BloomFilterWriter::reference() const {
+  return reference_;
+}
+
+BloomFilter* BloomFilterWriter::GetOrCreateBloomFilter(
```
wgtmac, line 77: It seems that this function always serves `Create` but no `Get`.

cpp/src/parquet/file_writer.cc
wgtmac: BTW, if they are stored between row groups, it would help relax memory pressure.

mapleFU:
@pitrou I move batch exec in a separated patch: https://github.com/apache/arrow/pull/35731

cpp/src/parquet/bloom_filter.h
```diff
@@ -101,6 +101,12 @@ class PARQUET_EXPORT BloomFilter {
   /// @return hash result.
   virtual uint64_t Hash(const FLBA* value, uint32_t len) const = 0;
 
+  // Variant of const pointer argument to facilitate template
```
emkornfield, line 104: instead of doing these, could we make const-ref parameters for the pointer types?

cpp/src/parquet/bloom_filter_writer.h
```diff
@@ -0,0 +1,74 @@
...
+#include "bloom_filter.h"
+
+#include <map>
+#include <vector>
+
+namespace parquet {
+
+class PARQUET_EXPORT RowGroupBloomFilterReference {
+ public:
+  struct Reference {
```
emkornfield, line 12: I think there is already a structure we are using for indexing and in other places of offset and length, lets please use that?

cpp/src/parquet/bloom_filter_writer.h
```diff
@@ -0,0 +1,74 @@
...
+    int64_t offset;
+    int32_t length;
+  };
+
+  /// Append a new row group to host all incoming bloom filters.
+  void AppendRowGroup();
+
+  /// Add reference to the serialized bloom filter.
```
emkornfield, line 20: its not clear what offset and length mean here?

cpp/src/parquet/bloom_filter_writer.h
```diff
@@ -0,0 +1,74 @@
...
+  void AddBloomFilter(int32_t column_id, int64_t offset, int32_t length);
+
+  /// Get bloom filter offsets of a specific row group.
+  bool GetBloomFilterOffsets(size_t row_group_ordinal,
+                             const std::map<int32_t, Reference>** out) const;
```
emkornfield, line 25: why the double pointer?

cpp/src/parquet/bloom_filter_writer.h
```diff
@@ -0,0 +1,74 @@
...
+      : properties_(properties) {}
+  /// Append a new row group to host all incoming bloom filters.
+  void AppendRowGroup();
+
+  /// Return a BloomFilter defined by `col_path`.
+  ///
+  /// * If the col_path has a bloom filter, create a BloomFilter in
+  ///  `row_group_bloom_filters_` and return.
+  /// * Otherwise, return nullptr.
+  BloomFilter* GetOrCreateBloomFilter(
```
emkornfield, line 51: should return be const?  please document the ownership contract (does caller take ownership or is retained by this class)?

cpp/src/parquet/bloom_filter_writer.h
```diff
@@ -0,0 +1,74 @@
...
+      const std::shared_ptr<schema::ColumnPath>& col_path,
```
emkornfield, line 52: prefer `const schema::ColumnPath&` unless this is optional.

cpp/src/parquet/bloom_filter_writer.h
```diff
@@ -0,0 +1,74 @@
...
+      const std::function<std::unique_ptr<BloomFilter>()>& bf_creator);
+
+  /// Serialize all bloom filters with header and bitset in the order of row group and
+  /// column id. Column encryption is not implemented yet. The side effect is that it
+  /// deletes all bloom filters after they have been flushed.
+  void WriteTo(::arrow::io::OutputStream* sink, const SchemaDescriptor* schema,
```
emkornfield, line 58: const SchemaDescriptr& unless schema is optional (int which case behavior should be documentated).

Typically output parameters should be the last arguments (e.g. sink).

cpp/src/parquet/bloom_filter_writer.h
```diff
@@ -0,0 +1,74 @@
...
+               InternalFileEncryptor* encryptor = nullptr);
+
+  /// Return reference to serialized bloom filters. It is undefined behavior until
+  /// WriteTo() has been called.
+  const RowGroupBloomFilterReference& reference() const;
+
+ private:
+  const WriterProperties& properties_;
+
+  std::vector<std::map<std::string, std::unique_ptr<BloomFilter>>>
```
emkornfield, line 68: are BloomFilter value types (i.e. should this be a unique ptr  or plain old object.

please document what std::string means in the map.  and what each element in the vector represents.

cpp/src/parquet/file_writer.cc
mapleFU: May I ask that why bloom filter data should before PageIndex?

cpp/src/parquet/bloom_filter.h
mapleFU: It means that we wrap it in a template or "PhysicalType"?

cpp/src/parquet/bloom_filter_writer.h
mapleFU: Sure, we can reuse same structure

cpp/src/parquet/bloom_filter_writer.h
mapleFU: I've changed to `column_ordinal` for leaf columns

cpp/src/parquet/bloom_filter_writer.h
mapleFU: > const SchemaDescriptr& unless schema is optional (int which case behavior should be documentated).

Yes I think it's a better C++ style, however, arrow parquet already uses `const SchemaDescriptor* schema` everywhere. I don't want to break that.

PR #35694, opened 2023-05-19
GH-35692: [C++][Parquet] Parquet: fixing FixedSizeListReader::AssembleArray when has null FixedSizeList

### LINKED ISSUE #35692: Null fixed-length lists cannot be read from parquet file

At least from pyarrow (and possibly for c++ more generally), null fixed-length lists cannot be read into a table from a parquet file, but they _can_ be written.

```python
import pyarrow as pa
import pyarrow.parquet as pq
array = pa.nulls(1, pa.list_(pa.int32(), 2))
table = pa.table([array], names=["values"])
pq.write_table(table, "list_table.parquet")
pq.read_table("list_table.parquet")
```
gives
```
  File "/Users/swnelson/code/personal/quivr/.quivr-venv/lib/python3.11/site-packages/pyarrow/parquet/core.py", line 2986, in read_table
    return dataset.read(columns=columns, use_threads=use_threads,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/swnelson/code/personal/quivr/.quivr-venv/lib/python3.11/site-packages/pyarrow/parquet/core.py", line 2614, in read
    table = self._dataset.to_table(
            ^^^^^^^^^^^^^^^^^^^^^^^
  File "pyarrow/_dataset.pyx", line 546, in pyarrow._dataset.Dataset.to_table
  File "pyarrow/_dataset.pyx", line 3449, in pyarrow._dataset.Scanner.to_table
pyarrow.lib.ArrowInvalid: Expected all lists to be of size=2 but index 1 had size=0
```

This error does _not_ occur for variable-length lists.

This is with pyarrow version 12.0.0.

### Component(s)

Parquet, Python

mapleFU:
### Rationale for this change

1. Parquet will regard `FixedSizeList` as regular list, and write rep-levels and def-levels if possible
2. When read from Parquet, if no null (or nested null, like `FixedSizeList` doesn't has null but it parent has) is contained, the file can be read correctly
3. However, when read from Parquet with nullable, the `AssembleArray` will not work well. It takes continuous "child array" and child offsets, and need to reserve extra padding space for "null". Previous reader doesn't implement it.

To be more clear:

```
for FixedSizeList(2): [[0, 0], null]

data to be Assemble Array:
null: [0 1]
data: [0 0]
offsets: [0, 2, 2]

required output:
null: [0 1]
data: [0, 0, undefined, undefined]
```

previous reader doesn't implement that required output

Rebuild the `FixedSizeList` when having null member

### Are these changes tested?

Yes

No

* Closes: #35692

### DIFF
--- a/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
+++ b/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
@@ -3010,6 +3010,27 @@ TEST(ArrowReadWrite, ListOfStructOfList1) {
   CheckSimpleRoundtrip(table, 2);
 }
 
+TEST(ArrowReadWrite, ListOfStructOfList2) {
+  using ::arrow::field;
+  using ::arrow::list;
+  using ::arrow::struct_;
+
+  auto type =
+      list(field("item",
+                 struct_({field("a", ::arrow::int16(), /*nullable=*/false),
+                          field("b", list(::arrow::int64()), /*nullable=*/false)}),
+                 /*nullable=*/false));
+
+  const char* json = R"([
+      [{"a": 123, "b": [1, 2, 3]}],
+      null,
+      [],
+      [{"a": 456, "b": []}, {"a": 789, "b": [null]}, {"a": 876, "b": [4, 5, 6]}]])";
+  auto array = ::arrow::ArrayFromJSON(type, json);
+  auto table = ::arrow::Table::Make(::arrow::schema({field("root", type)}), {array});
+  CheckSimpleRoundtrip(table, 2);
+}
+
 TEST(ArrowReadWrite, ListWithNoValues) {
   using ::arrow::Buffer;
   using ::arrow::field;
@@ -3074,7 +3095,7 @@ TEST(ArrowReadWrite, FixedSizeList) {
   CheckSimpleRoundtrip(table, 2, props_store_schema);
 }
 
-TEST(ArrowReadWrite, NullableFixedSizeList) {
+TEST(ArrowReadWrite, FixedSizeList2) {
   using ::arrow::field;
   using ::arrow::fixed_size_list;
   using ::arrow::struct_;
--- a/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
+++ b/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
@@ -3074,6 +3074,26 @@ TEST(ArrowReadWrite, FixedSizeList) {
   CheckSimpleRoundtrip(table, 2, props_store_schema);
 }
 
+TEST(ArrowReadWrite, NullableFixedSizeList) {
+  using ::arrow::field;
+  using ::arrow::fixed_size_list;
+  using ::arrow::struct_;
+
+  auto type = fixed_size_list(::arrow::int16(), /*size=*/3);
+
+  const char* json = R"([
+      [1, 2, 3],
+      null,
+      [4, 2, 6],
+      null,
+      [7, 8, 3],
+      null])";
+  auto array = ::arrow::ArrayFromJSON(type, json);
+  auto table = ::arrow::Table::Make(::arrow::schema({field("root", type)}), {array});
+  auto props_store_schema = ArrowWriterProperties::Builder().store_schema()->build();
+  CheckSimpleRoundtrip(table, 2, props_store_schema);
+}
+
 TEST(ArrowReadWrite, ListOfStructOfList2) {
   using ::arrow::field;
   using ::arrow::list;
--- a/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
+++ b/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
@@ -3082,37 +3082,40 @@ TEST(ArrowReadWrite, NullableFixedSizeList) {
   auto type = fixed_size_list(::arrow::int16(), /*size=*/3);
 
   const char* json = R"([
-      [null, 2, 3],
       null,
-      [4, null, 6],
+      [null, 1, 2],
+      [null, 3, 4],
       null,
-      [7, 8, null],
-      null])";
+      [5, null, 6],
+      [7, null, 8],
+      null,
+      [9, 10, null],
+      null,
+      null,
+      [11, 12, 13],
+      [14, 15, 16]])";
   auto array = ::arrow::ArrayFromJSON(type, json);
   auto table = ::arrow::Table::Make(::arrow::schema({field("root", type)}), {array});
   auto props_store_schema = ArrowWriterProperties::Builder().store_schema()->build();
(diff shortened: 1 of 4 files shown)


spenczar:
Reading https://arrow.apache.org/docs/format/Columnar.html#fixed-size-list-layout more closely, I'm not so sure that the zero-sized elements are correct:
 
> A fixed size list type is specified like FixedSizeList<T>[N], where T is any type (primitive or nested) and N is a 32-bit signed integer representing the length of the lists.
>
> A fixed size list array is represented by a values array, which is a child array of type T. T may also be a nested type. The value in slot j of a fixed size list array is stored in an N-long slice of the values array, starting at an offset of j * N.

`j * N` makes it sound like offsets are always fixed size for fixed size lists. In fact, that document makes it sound like there shouldn't be an offsets buffer at all, since fixed-size lists use "slots". 

Could this actually be a writer bug?

mapleFU:
@spenczar Hi, I think it's just a reader bug. The written data is not corrupt.

1. Parquet will regard `FixedSizeList` as regular list, and write rep-levels and def-levels if possible
2. When read from Parquet, if no null (or nested null, like `FixedSizeList` doesn't has null but it parent has) is contained, the file can be read correctly
3. However, when read from Parquet with nullable, the `AssembleArray` will not work well. It takes continuous "child array" and child offsets, and need to reserve extra padding space for "null". Previous reader doesn't implement it.

mapleFU:
I've submit a fixing, but I'm not familiar with `FixedSizeListArrayBuilder` and `Array`, so I may made some mistakes. Can @pitrou helps me review this patch?

cpp/src/parquet/arrow/reader.cc
```diff
@@ -673,13 +675,39 @@ class PARQUET_NO_EXPORT FixedSizeListReader : public ListReader<int32_t> {
...
     const auto& type = checked_cast<::arrow::FixedSizeListType&>(*field()->type());
     const int32_t* offsets = reinterpret_cast<const int32_t*>(data->buffers[1]->data());
+    // Need to reassemble data with correct offsets and paddings.
+    if (field()->nullable() && data->null_count != 0) {
+      std::unique_ptr<::arrow::ArrayBuilder> temp_builder;
+      ARROW_RETURN_NOT_OK(::arrow::MakeBuilder(::arrow::default_memory_pool(),
+                                               field()->type(), &temp_builder));
+      auto fixed_sized_list_builder =
+          checked_cast<::arrow::FixedSizeListBuilder*>(temp_builder.get());
+      for (int x = 1; x <= data->length; x++) {
```
mapleFU, line 685: (I guess using validity_bitmap is better here, I may change it after I wake up)

mapleFU:
@spenczar Would you mind have a try on this patch? The code might not be the final commit code, but I want to make sure that problem got fixed :-)

mapleFU:
I've found another unrelated issue when writing unittests: https://github.com/apache/arrow/issues/35697

pitrou:
@emkornfield Can you take a look here? I'm not sure I understand the issue.

emkornfield:
@mapleFU its not clear to me that this approach works for all level of nestings.  When I was thinking about the problem, I thought it would be better to populate and make use of [null_slot_usage](https://github.com/apache/arrow/blob/main/cpp/src/parquet/level_conversion.h#L53) on reading each array back.

mapleFU:
I've tested the nested case. But I can take a look at `null_slot_usage`, thanks a lot!

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -3030,25 +3051,48 @@ TEST(ArrowReadWrite, FixedSizeList) {
...
+      [9, 10, null],
+      null,
+      null,
+      [11, 12, 13],
+      [14, 15, 16]])";
   auto array = ::arrow::ArrayFromJSON(type, json);
   auto table = ::arrow::Table::Make(::arrow::schema({field("root", type)}), {array});
-  CheckSimpleRoundtrip(table, 2);
+  auto props_store_schema = ArrowWriterProperties::Builder().store_schema()->build();
+  CheckSimpleRoundtrip(table, 2, props_store_schema);
```
emkornfield, line 3077: I think we need more tests here that can cover nulls in the middle and multiple levels of nested fixed sizes lists with other types of lists.

mapleFU:
I got what you mean, I didn't know `null_slot_usage` previously, thanks a lot!

mapleFU:
```
  if (valid_bits_writer.has_value()) {
    valid_bits_writer->Finish();
  }
  if (offsets != nullptr) {
    output->values_read = offsets - orig_pos;
  } else if (valid_bits_writer.has_value()) {
    output->values_read = valid_bits_writer->position();
  }
  if (output->null_count > 0 && level_info.null_slot_usage > 1) {
    throw ParquetException(
        "Null values with null_slot_usage > 1 not supported."
        "(i.e. FixedSizeLists with null values are not supported)");
  }
```

Seems that I should support the handling logic here?

mapleFU:
I'll close this and submit another patch using `null_slot_usage`

emkornfield:
If this code works (and we don't think null_slot_usage can be more efficient) it is ok to pursue this approach, so feel free to reopen

mapleFU:
Emmm I'm sure it works, `null_slot_usage` will be more efficient because it doesn't copy more data, but my solution may "copy" the underlying leaf data. But I think my solution is right. Should you keep this as a quick fixing? I'm a bit busy this week, and may only have time for `null_slot_usage` fixing this weekend. @pitrou @emkornfield

mapleFU:
https://github.com/apache/arrow/pull/35694#issuecomment-1596728191
@emkornfield @pitrou would you mind help for this? When we have:
1. `FixedSizeList<FixedSizeList<int, 2>, 2>`
2.`FixedSizeList<List<FixedSizeList<int, 2>, 2>>`

What should  `null_slot_usage` be? Should it be? Should them be:
1. `(2)` for `FixedSizeList<int, 2>`, and `(4)` for `FixedSizeList<FixedSizeList<int, 2>, 2>`?
2. `(2)` for `FixedSizeList<int, 2>`, and `(4)` for `FixedSizeList<FixedSizeList<int, 2>, 2>`, what would be for `FixedSizeList<List<FixedSizeList<int, 2>, 2>>`?

spenczar:
Is there any way I can help get some motion here? This remains a pretty nasty blocker for some applications of Arrow and Parquet for us.

mapleFU:
I believe this patch will solve part of the problem. However, it might has some performance lost, since it need to copy the underlying data. Feel free to use this as a patch

Using `null_slot` might be better here. But I need some help to tell me how to implement it.

PR #35719, opened 2023-05-23
GH-35712: [CI][C++] DNM: Debug arrow-array-test slow on macOS

### LINKED ISSUE #35712: [CI][C++] arrow-array-test slow on macOS

This is rather minor, but `arrow-array-test` takes more than one minute on our C++ macOS CI. I even saw it sporadically time out.

It would be nice to find out the cause, as it may be an underlying performance problem in Arrow C++.
```
 25/101 Test   #1: arrow-array-test ..........................   Passed  103.65 sec
```

### Component(s)

C++

mapleFU:

* Closes: #35712


mapleFU:
Close and not reuse it here, because I think I can reuse this to debug in the future

PR #35720, opened 2023-05-23
MINOR: [Docs][C++] Improve Input Shapes section of Compute doc

js8544:
1.  The function that only takes arrays should be `array_sort_indices`, not `sort_indices`.
2. IMO The word `either` implies that `array_sort_indices` also accepts chunked arrays but in fact it doesn't. So I changed `either` to `particular input types`.
3. There are no mention of two dimensional data types, i.e. RecordBatch and Table, throughout the Input Shapes section, which I feel like is a miss. So I added `sort_indices` as an example for them.


kou (approved):
+1

pitrou (approved):
Thanks a lot @js8544 !

PR #35723, opened 2023-05-23
GH-32832: [Go] support building with tinygo

### LINKED ISSUE #32832: [Go] Unable to build with tinygo
I was hoping to use TinyGo to build WASM binaries with Arrow.  TinyGo can generate builds that are [1% the size](https://tinygo.org/getting-started/overview/#:~:text=The%20only%20difference%20here%2C%20is,used%2C%20and%20the%20associated%20runtime.&text=In%20this%20case%20the%20Go,size%20(251k%20before%20stripping)!) of those generated with Go (significant for applications hosted on the web).

Arrow's use of `reflect.SliceHeader` fields limits the portability of the code.  For example, the `Len` and `Cap` fields are assumed to be `int` here: https://github.com/apache/arrow/blob/go/v9.0.0/go/arrow/bitutil/bitutil.go#L158-L159

Go's [reflect package warns](https://github.com/golang/go/blob/go1.19/src/reflect/value.go#L2675-L2685) that the SliceHeader "cannot be used safely or portably and its representation may change in a later release."

Attempts to build a WASM binary using the github.com/apache/arrow/go/v10 module result in failures like this:
```Java

tinygo build -tags noasm -o test.wasm ./main.go
```
```Java
             
# github.com/apache/arrow/go/v10/arrow/bitutil
../../go/pkg/mod/github.com/apache/arrow/go/v10@v10.0.0-20220831082949-cf27001da088/arrow/bitutil/bitutil.go:158:10: invalid operation: h.Len / uint64SizeBytes (mismatched types uintptr and int)
../../go/pkg/mod/github.com/apache/arrow/go/v10@v10.0.0-20220831082949-cf27001da088/arrow/bitutil/bitutil.go:159:10: invalid operation: h.Cap / uint64SizeBytes (mismatched types uintptr and int)
```

This happens because TinyGo uses `uintptr` for the corresponding types: https://github.com/tinygo-org/tinygo/blob/v0.25.0/src/reflect/value.go#L773-L777

This feels like an issue with TinyGo, and it has been ticketed there multiple times (see https://github.com/tinygo-org/tinygo/issues/1284).  They lean on the warnings in the Go sources that use of the SliceHeader fields makes code unportable and suggest changes to the libraries that do not heed this warning.

I don't have a suggested fix or alternative for Arrow's use of SliceHeader fields, but I'm wondering if there would be willingness on the part of this package to make WASM builds work with TinyGo.  Perhaps the TinyGo authors could also offer suggested changes.

**Reporter**: [Tim Schaub](https://issues.apache.org/jira/browse/ARROW-17584)
#### PRs and other links:
- [GitHub Pull Request #14026](https://github.com/apache/arrow/pull/14026)

<sub>**Note**: *This issue was originally created as [ARROW-17584](https://issues.apache.org/jira/browse/ARROW-17584). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

chriscasola:
### Rationale for this change

To support compiling with tinygo which enables use of arrow in environments where binary size is important, like web assembly.

Using an internal JSON package that uses `goccy/go-json` for regular builds as it does currently, but uses the native `encoding/json` for tinygo builds. This is necessary because go-json has a lot of code that is incompatible with tinygo.

Remove dependency on `parquet` package from non-parquet code since it is also incompatible with tinygo.

Other minor tweaks for compatibility with tinygo.

### Are these changes tested?

Should we add a build step that compiles the example with tinygo?

None.

* Closes: #32832


chriscasola:
@zeroshade we're still testing this on our end, but it does work. I'm curious what your thoughts are.

go/arrow/array/string.go
```diff
@@ -79,16 +79,12 @@ func (a *String) ValueOffsets() []int32 {
...
 	beg := a.array.data.offset
 	end := beg + a.array.data.length
 	data := a.values[a.offsets[beg]:a.offsets[end]]
 
-	s := (*reflect.SliceHeader)(unsafe.Pointer(&ret))
-	s.Data = (*reflect.StringHeader)(unsafe.Pointer(&data)).Data
-	s.Len = len(data)
-	s.Cap = len(data)
-	return
+	return []byte(data)
```
zeroshade, line 87: this causes a copy which isn't what we want. This could end up creating a copy of the entire array essentially which would be pretty bad. I assume this change is because of the issues with `SliceHeader`/`StringHeader` in tinygo?

An alternative solution here would be to just pull the bytes from the buffer directly:

```go
if a.array.data.buffers[2] != nil {
    return a.array.data.buffers[2].Bytes()[a.offsets[beg]:a.offsets[end]]
}
return nil
```

This way we preserve the fact that we're not copying the data.

go/arrow/array/string.go
```diff
@@ -221,16 +217,12 @@ func (a *LargeString) ValueOffsets() []int64 {
...
```
zeroshade, line 225: same comment as above

go/arrow/bitutil/bitutil.go
```diff
@@ -150,15 +150,12 @@ const (
...
-	s.Data = h.Data
-	s.Len = h.Len / uint64SizeBytes
-	s.Cap = h.Cap / uint64SizeBytes
+	if cap(b) < uint64SizeBytes {
+		return nil
+	}
 
-	return res
+	h := (*reflect.SliceHeader)(unsafe.Pointer(&b))
+	return unsafe.Slice((*uint64)(unsafe.Pointer(h.Data)), cap(b)/uint64SizeBytes)[:len(b)/uint64SizeBytes]
```
zeroshade, line 158: if we're sure this is only ever called with a slice with a non-zero length, this could be simplified to 

```go
return unsafe.Slice((*uint64)(unsafe.Pointer(&b[0])), cap(b)/uint64SizeBytes)[:len(b)/uint64SizeBytes]
```

go/arrow/compute/arithmetic_test.go
```diff
@@ -2621,7 +2621,7 @@ func (us *UnaryArithmeticSigned[T]) TestAbsoluteValue() {
 				fn(`[]`, `[]`)
 				// scalar/arrays with nulls
 				fn(`[null]`, `[null]`)
-				fn(`[1, null -10]`, `[1, null, 10]`)
+				fn(`[1, null, -10]`, `[1, null, 10]`)
```
zeroshade, line 2624: so this is interesting, how was this not failing already?

go/arrow/compute/cast_test.go
```diff
@@ -2589,7 +2589,7 @@ func (c *CastSuite) TestStructToDifferentNullabilityStruct() {
 		}
 		srcNonNull, _, err := array.FromJSON(c.mem, arrow.StructOf(fieldsSrcNonNullable...),
 			strings.NewReader(`[
-				{"a": 11, "b": 32, "c", 95},
+				{"a": 11, "b": 32, "c": 95},
```
zeroshade, line 2592: same comment as before, how is this not already failing?

go/internal/hashing/xxh3_memo_table.go
```diff
@@ -166,16 +168,13 @@ func (s *BinaryMemoTable) Size() int {
 }
 
 // helper function to easily return a byte slice for any given value
-// regardless of the type if it's a []byte, parquet.ByteArray,
-// parquet.FixedLenByteArray or string.
+// regardless of the type.
```
zeroshade, line 171: should probably clarify that this only works for `[]byte`, `string` or something which fulfills the `ByteSlice` interface

zeroshade (changes requested):
In general this LGTM, just a few nit picks

go/arrow/array/string.go
chriscasola: Yes we changed this because of the `SliceHeader` issue.

When would this cause a copy? Aren't we slicing the string which is referencing the same underlying data?

go/arrow/bitutil/bitutil.go
chriscasola: I'm not sure I can guarantee that. Are you suggesting we add a check just above and `return nil` if the length is `< 1`?

go/arrow/compute/arithmetic_test.go
chriscasola: The goccy JSON parser seems to be more forgiving than the built-in one.

go/arrow/array/string.go
zeroshade: Strings in Go are immutable, so unless the compiler can confirm that `[]byte(str)` is only used as a temporary it will do a copy of the bytes to construct the byte slice from the string.

go/arrow/bitutil/bitutil.go
zeroshade: nah, if we can't guarantee that it's a non-zero length then this is likely fine as is.

chriscasola:
@zeroshade thanks for the review, I think I addressed everything.

What are you thoughts on building with tinygo in CI? The CI setup on this repo seems complex so I'm not sure what that would entail, but it would be nice to avoid new code being introduced that does not build.

go/arrow/array/string.go
zeroshade: This still needs to get updated to no longer be copying the string (like the equivalent function for `(*String)`

zeroshade:
@chriscasola Do you think it's sufficient to just have a single x86-64/linux CI run with tinygo? Or do you think we'd need for arm and other cases too?

chriscasola:
> @chriscasola Do you think it's sufficient to just have a single x86-64/linux CI run with tinygo? Or do you think we'd need for arm and other cases too?

I think a single architecture is fine.

.github/workflows/go.yml
```diff
@@ -394,3 +394,20 @@ jobs:
       - name: Test
         shell: bash
         run: ci/scripts/go_test.sh $(pwd)
+
+  tinygo:
+    name: TinyGo
+    runs-on: ubuntu-latest
+    if: ${{ !contains(github.event.pull_request.title, 'WIP') }}
+    env:
+      TINYGO_VERSION: 0.27.0
```
zeroshade, line 403: i would suggest adding an image to the docker-compose.yml and a dockerfile to `ci/docker/` which is an image with tinygo installed and then use `archery docker run` like other examples.

ci/scripts/go_tinygo_example.sh
```diff
@@ -0,0 +1,27 @@
...
+# under the License.
+
+set -ex
+
+cd ~
+pushd /src
+tinygo build -tags noasm -o ~/example_tinygo arrow/_examples/helloworld/main.go
+popd
+
+./example_tinygo
```
zeroshade, line 27: any reason to not just have it run the unit tests?

ci/scripts/go_tinygo_example.sh
```diff
@@ -0,0 +1,27 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+set -ex
+
+cd ~
+pushd /src
+tinygo build -tags noasm -o ~/example_tinygo arrow/_examples/helloworld/main.go
```
zeroshade, line 24: should we update all of the go build constraints for `noasm` to be `noasm || tinygo` or whatever the equivalent is?

ci/scripts/go_tinygo_example.sh
chriscasola: Unfortunately tinygo can't compile/run unit tests right now. At least it can't handle the ones in this project. See [this issue](https://github.com/tinygo-org/tinygo/issues/1359#issuecomment-687668399).

ci/scripts/go_tinygo_example.sh
chriscasola: @zeroshade I attempted to do this, but it seems like package outside arrow might also have a `noasm` flag. Even after adding the `tinygo` condition everywhere there was a `noasm` condition, I still couldn't build with out the `noasm` tag. Are you okay omitting this for now?

zeroshade: yea that's fine for now. We should probably figure out which package it is that has the `noasm` condition etc.

zeroshade (changes requested):
Just one request: can you add to the `go/arrow/doc.go` file's package comment documentation the info about building with tinygo like needing the `noasm` tag, etc.?

Otherwise this looks good to me.

zeroshade (approved):

tschaub:
Cool to see this got in.  I wonder if anybody looked into what it would take to get the Parquet packages working as well.  I may give it another try at some point.

zeroshade:
I believe the current blocker on the parquet packages is that something in the Thrift dependency is not compatible with tinygo

Issue #35040, opened 2023-04-11
[C++] Change Scalar::CastTo to do safe cast by default and allow to specify cast options ?


jorisvandenbossche:
### Describe the enhancement requested

See https://github.com/apache/arrow/issues/34901 for a longer discussion, but summarizing: the `pyarrow.Scalar` object has a `cast()` method, but in contrast with other cast methods in pyarrow it does an unsafe cast by default. We should probably change this to do a safe cast by default, and at the same time also allow to specify CastOptions (so a user can still choose to do an unsafe cast). 

Example:

```python
# scalar behaviour
>>> pa.scalar(1.5)
<pyarrow.DoubleScalar: 1.5>
>>> pa.scalar(1.5).cast(pa.int64())
<pyarrow.Int64Scalar: 1>

# vs array behaviour
>>> pa.array([1.5]).cast(pa.int64())
...
ArrowInvalid: Float value 1.5 was truncated converting to int64
```

The python cast() method calls the C++ `Scalar::ToCast`:

https://github.com/apache/arrow/blob/e488942cd552ac36a46d40477c1b0326a626ed98/cpp/src/arrow/scalar.h#L99-L100

which currently indeed doesn't have the option to pass CastOptions.

In addition, it seems that for casting Scalars, we do have a somewhat custom implementation, and this doesn't use the generic Cast implementation (from the compute kernels), but has a custom `CastImpl` in scalar.cc. Not fully sure about the reason for this, but maybe historically we wanted to have scalar casting without relying on the optional compute module? (cfr https://github.com/apache/arrow/issues/25025)

### Component(s)

C++

westonpace:
I think this is a good idea.  The result will probably be a slight bit slower but working with scalars is already slightly slower anyways and it would be nice to have only a single implementation.

westonpace:
Also, we've since extracted the cast kernels from the rest of the kernels and declared them as "essential kernels" that will always be present (even if ARROW_COMPUTE is OFF) so I think it is safe to rely on them.

danepitkin:
FYI I tried swapping the Scalar.cast() implementation to the compute kernel and current (but small) test suite passed while also fixing https://github.com/apache/arrow/issues/35370.

```
...
    def cast(self, object target_type=None, safe=None, options=None, memory_pool=None):
        return _pc().cast(self, target_type, safe=safe,
                          options=options, memory_pool=memory_pool)
...
```

AlenkaF:
[...]

PR #12536, opened 2022-03-01
ARROW-13168: [C++][R] Enable runtime timezone database for Windows

wjones127:
This allows for runtime configuration of the timezone database on Windows for C++ and R. Python will be handled later because it's available timezone libraries use the binary rather than text format, which is not yet supported the vendored date library.

For R, Windows will only support the "C" locale, since (as far as I can tell) that's the only locale supported by the MingW std::locale implementation. I think R itself gets around this by implementing a completely custom version of `strftime()` and friends.

PR #35735, opened 2023-05-24
GH-35040: [Python] Skip test_cast_timestamp_to_string on Windows because it requires tz database

jorisvandenbossche:
### Rationale for this change

Fix up of https://github.com/apache/arrow/pull/35395, skipping one of the tests added in that PR on Windows, because the test requires access to a tz database.


jorisvandenbossche:

wgtmac (approved):

pitrou:
@rok does this look right? We are able to somehow use a timezone database on Windows, IIRC?

jorisvandenbossche:
Yeah, the better fix would actually be to detect if the database is available, and only skip if not found on windows. Because this patch skips the tests altogether, even when it can actually work (eg on appveyor the tests pass because we download the tzdata during appveyor setup). But just to note: we currently do this plain skip for other tests requiring the database as well.

I assume checking if the path `%USERPROFILE%\Downloads\tzdata` exists might be a sufficient check at the moment (since that's the place where tz.cpp will look, and we currently don't yet allow customizing that in pyarrow).

danepitkin:
Thanks for fixing! There are other tests in this file that also use `pytest.importorskip("pytz")`, should they be updated as well? It's not clear to me why this test case behaves differently.

python/pyarrow/tests/test_scalars.py
```diff
@@ -295,6 +296,9 @@ def test_cast():
         pa.scalar('foo').cast('int32')
 
 
+# TODO: We should test on windows once ARROW-13168 is resolved.
+@pytest.mark.skipif(sys.platform == 'win32',
+                    reason="Timezone database is not available on Windows yet")
```
rok, line 301: Perhaps a more correct wording (so as to not confuse the reader) would be:
```suggestion
@pytest.mark.skipif(sys.platform == 'win32',
                    reason="Timezone database is sometimes not available on Windows yet")
```

rok (commented):
For completeness - I believe db download for windows was added by @wjones127 in https://github.com/apache/arrow/pull/12536.

wjones127:
> For completeness - I believe db download for windows was added by @wjones127 in #12536.

That was just for C++ and R. I don't think we figured out how to do it for Python, since we need the text form of the IANA database while Python's timezone packages provide the binary form.

raulcd (approved):
We already have 3 tests being skipped on Windows for the same reason.
https://github.com/apache/arrow/blob/main/python/pyarrow/tests/test_compute.py#L1851
https://github.com/apache/arrow/blob/main/python/pyarrow/tests/test_compute.py#L2042
https://github.com/apache/arrow/blob/main/python/pyarrow/tests/test_compute.py#L2042
I am +1 to merge

rok (approved):

python/pyarrow/tests/test_compute.py
```diff
@@ -1846,17 +1850,18 @@ def test_strptime():
...
     timezones = ["CET", "UTC", "Europe/Ljubljana"]
 
-    formats = ["%a", "%A", "%w", "%d", "%b", "%B", "%m", "%y", "%Y", "%H",
-               "%I", "%p", "%M", "%z", "%Z", "%j", "%U", "%W", "%c", "%x",
-               "%X", "%%", "%G", "%V", "%u"]
+    formats = ["%a", "%A", "%w", "%d", "%b", "%B", "%m", "%y", "%Y", "%H", "%I",
+               "%p", "%M", "%z", "%Z", "%j", "%U", "%W", "%%", "%G", "%V", "%u"]
+    if sys.platform != "win32":
+        # Locale-dependent formats don't match on Windows
+        formats.extend(["%c", "%x", "%X"])
```
jorisvandenbossche, line 1864: The error on Appveyor we got was:

```
    @pytest.mark.skipif(sys.platform == "win32" and not util.windows_has_tzdata(),
                        reason="Timezone database is not installed on Windows")
    def test_strftime():
        times = ["2018-03-10 09:00", "2038-01-31 12:23", None]
        timezones = ["CET", "UTC", "Europe/Ljubljana"]
    
        formats = ["%a", "%A", "%w", "%d", "%b", "%B", "%m", "%y", "%Y", "%H",
                   "%I", "%p", "%M", "%z", "%Z", "%j", "%U", "%W", "%c", "%x",
                   "%X", "%%", "%G", "%V", "%u"]
    
        for timezone in timezones:
            ts = pd.to_datetime(times).tz_localize(timezone)
            for unit in ["s", "ms", "us", "ns"]:
                tsa = pa.array(ts, type=pa.timestamp(unit, timezone))
                for fmt in formats:
                    options = pc.StrftimeOptions(fmt)
                    result = pc.strftime(tsa, options=options)
                    expected = pa.array(ts.strftime(fmt))
>                   assert result.equals(expected)
E                   assert False
E                    +  where False = <built-in method equals of pyarrow.lib.StringArray object at 0x0000023767338600>(<pyarrow.lib.StringArray object at 0x0000023767338830>\n[\n  "Sat Mar 10 09:00:00 2018",\n  "Sun Jan 31 12:23:00 2038",\n  null\n])
E                    +    where <built-in method equals of pyarrow.lib.StringArray object at 0x0000023767338600> = <pyarrow.lib.StringArray object at 0x0000023767338600>\n[\n  "03/10/18 09:00:00",\n  "01/31/38 12:23:00",\n  null\n].equals
pyarrow\tests\test_compute.py:1872: AssertionError
```

So it seems that we create a string like "Sat Mar 10 09:00:00 2018", but the python version we compare with gives "03/10/18 09:00:00". According to docs for `%c`, the former (our result) is actually correct. 
But since we are checking matching results in Python in this test, just skipping the ones where those don't match.

jorisvandenbossche:
I updated the PR to actually check if the tzdata database was present, and only skip the tests if not. Because we do actually already download the tzdb on Appveyor (not on the nightly wheel builds, so therefore the tests were failing there). 
While Python doesn't yet allow you to configure the path (and use the data from the `tzdata` package), as mentioned above, if you manually download it to the correct location, the C++ code will already discover it (as was done for R).

jorisvandenbossche:

jorisvandenbossche:

Issue #35641, opened 2023-05-17
[CI][C++] Bad compile/link times despite ccache


pitrou:
### Describe the bug, including details regarding any error messages, version, and platform.

Example here on the AMD64 Conda C++ build:
https://github.com/apache/arrow/actions/runs/5005430904/jobs/8969381538?pr=33634

```
=== ccache statistics before build
===
Cacheable calls:   284 / 894 (31.77%)
  Hits:             14 / 284 ( 4.93%)
    Direct:         14 /  14 (100.0%)
    Preprocessed:    0 /  14 ( 0.00%)
  Misses:          270 / 284 (95.07%)
Uncacheable calls: 610 / 894 (68.23%)
Local storage:
  Cache size (GB): 0.1 / 1.0 (11.35%)
  Hits:             14 / 284 ( 4.93%)
  Misses:          270 / 284 (95.07%)

[ ... CMake runs ... ]

real	27m46.563s
user	49m43.936s
sys	4m28.705s

===
=== ccache statistics after build
===
Cacheable calls:    568 / 1788 (31.77%)
  Hits:             298 /  568 (52.46%)
    Direct:         233 /  298 (78.19%)
    Preprocessed:    65 /  298 (21.81%)
  Misses:           270 /  568 (47.54%)
Uncacheable calls: 1220 / 1788 (68.23%)
Local storage:
  Cache size (GB):  0.1 /  1.0 (11.35%)
  Hits:             298 /  568 (52.46%)
  Misses:           270 /  568 (47.54%)
```

So ccache was completely successful in this build (0 misses during the entire CMake configure / build), yet CMake took 27 minutes.

We should investigate:
* disabling building static libraries (`ARROW_BUILD_STATIC`) to save on linking times
* why there are more than 600 uncacheable calls and only 280 cacheable calls, which is the majority of the ~860 compilation calls (the "Building CXX" lines in the logs)

Note that the "AMD64 Conda C++" build can be reproduced using [Archery](https://arrow.apache.org/docs/dev/developers/continuous_integration/docker.html) with the `conda-cpp` image:
```console
$ archery docker run conda-cpp
```

### Component(s)

C++

pitrou:
cc @kou @felipecrv

pitrou:
It seems "Could not use precompiled header" is the reason, at least on `conda-cpp`, for uncacheable compiler calls.

kou:
If precompiled header is a problem, we may be able to fix this by the following change:

```diff
[...]

PR #35738, opened 2023-05-24
MINOR: [CI][C++] Make ASan job faster

pitrou:
ASAN + UBSAN massively increase the generated code size.

The default compile cache size (1 GB) wasn't sufficient to hold all generated outputs, therefore the cache hit rate would be close to 0% (a cached object would be evicted before the first time it could be requested again).

Also, save 5GB in the build directory by disabling static libraries.

### DIFF
--- a/.github/workflows/cpp.yml
+++ b/.github/workflows/cpp.yml
@@ -49,7 +49,7 @@ permissions:
   contents: read
 
 env:
-  ARROW_ENABLE_TIMING_TESTS: "OFF"
+  ARROW_ENABLE_TIMING_TESTS: OFF
   DOCKER_VOLUME_PREFIX: ".docker/"
 
 jobs:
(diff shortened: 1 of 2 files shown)


pitrou:
cc @kou

pitrou:
I verified that it massively improves caching on a local Docker build.

pitrou:
Example job with full caching: https://github.com/pitrou/arrow/actions/runs/5068416511/jobs/9101198018

kou (approved):
+1

Wow! Good catch!

It seems that we may be able to reduce more build time because there are still 1780 uncachable calls:

https://github.com/pitrou/arrow/actions/runs/5068416511/jobs/9101198018#step:6:2951

```text
  Uncacheable:              1780
```

#35641 may be related.

assignUser:
Very nice!

fyi you can use `--allow-empty` to trigger ci without having to do any changes (as long some changes have been made already)

PR #35751, opened 2023-05-24
GH-35765: [C++] Split vector_selection.cc into more compilation units

### LINKED ISSUE #35765: [C++] Split vector_selection.cc into more compilation units

`vector_selection.cc` contains multiple kernel implementations — `take`, `filter`, `drop_null`... — and many specializations. It's hard to see what is related to what and if we mix in kernel specializations for run-end encoded filters and data it will get really hard to review for completeness

### Component(s)

C++

felipecrv:
### Rationale for this change

When working on #35001 I had a hard time figuring where to place the code for all possible combinations of filters and REE data. `vector_selection.cc` is hard to follow with so many kernels implemented in a single file. This PR splits the two biggest ones: filter and take. Stuff that can be shared by both stays is in `vector_selection_internal.cc` and `vector_selection.cc` is concerned with the registering of the functions and a few smaller kernels.

- [x]  `vector_selection_(internal|take|filter).(cc|h)` source files were extracted from `vector_selection.cc`

### Are these changes tested?

Yes, by existing tests.

* Closes: #35765


felipecrv:
@pitrou @westonpace I tried to separate the code-moving commits from the ones that make changes to explain what's being done step-by-step.

cpp/src/arrow/compute/kernels/vector_selection_filter.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include "arrow/array/data.h"
+#include "arrow/compute/api_vector.h"
+#include "arrow/compute/kernels/vector_selection_internal.h"
+
+namespace arrow {
+namespace compute {
+namespace internal {
+
+int64_t GetBitmapFilterOutputSize(const ArraySpan& filter,
```
pitrou, line 31: Add a docstring here?

cpp/src/arrow/compute/kernels/vector_selection_internal.h
```diff
@@ -0,0 +1,66 @@
...
+  ArrayKernelExec exec;
+};
+
+void RegisterSelectionFunction(const std::string& name, FunctionDoc doc,
+                               VectorKernel base_kernel, InputType selection_type,
+                               const std::vector<SelectionKernelData>& kernels,
+                               const FunctionOptions* default_options,
+                               FunctionRegistry* registry);
+
+Status PreallocateData(KernelContext* ctx, int64_t length, int bit_width,
```
pitrou, line 44: The name `PreallocateData` is probably too general, since it's selection-specific?

cpp/src/arrow/compute/kernels/vector_selection.cc
```diff
@@ -58,2119 +60,22 @@ using internal::OptionalBitIndexer;
 namespace compute {
 namespace internal {
 
+using FilterState = OptionsWrapper<FilterOptions>;
+using TakeState = OptionsWrapper<TakeOptions>;
+
 int64_t GetFilterOutputSize(const ArraySpan& filter,
                             FilterOptions::NullSelectionBehavior null_selection) {
```
pitrou, line 67: Is this function still useful, since it's a trivial wrapper around `GetBitmapFilterOutputSize`?

cpp/src/arrow/compute/kernels/vector_selection_internal.cc
```diff
@@ -0,0 +1,815 @@
...
+  }
+};
+
+#undef LIFT_BASE_MEMBERS
+
+// ----------------------------------------------------------------------
+
+template <typename Impl>
+Status FilterExec(KernelContext* ctx, const ExecSpan& batch, ExecResult* out) {
+  // TODO: where are the values and filter length equality checked?
```
pitrou, line 730: Here apparently:
https://github.com/apache/arrow/blob/6bd31f37ae66bd35594b077cb2f830be57e08acd/cpp/src/arrow/compute/function.cc#L270-L273

pitrou (changes requested):
This PR is not minor in code changes, even though most of it is moving code around. Can you open a GH issue?

cpp/src/arrow/compute/kernels/vector_selection.cc
felipecrv: It becomes this in #35750.
```cpp
int64_t GetFilterOutputSize(const ArraySpan& filter,
                            FilterOptions::NullSelectionBehavior null_selection) {
  if (filter.type->id() == Type::BOOL) {
    return GetBitmapFilterOutputSize(filter, null_selection);
  }
  return GetREEFilterOutputSize(filter, null_selection);
```

 I will move it completely to `vector_selection_filter.cc` now and remove it here, but I haven't decided yet if the REEx* filter kernels are going to be implemented in `vector_selection_filter.cc` or in a separate `.cc`, so I might undo this later.

cpp/src/arrow/compute/kernels/vector_selection_internal.cc
felipecrv: Nice. I can remove the old `TODO` comment then.

cpp/src/arrow/compute/kernels/vector_selection_filter.h
felipecrv: Removed the declaration from header. :heavy_check_mark:

assignUser:
For reference: A minor PR touches <=2 files with no changes to behaviour. (e.g. fixing typos and similar things)

cpp/src/arrow/compute/kernels/vector_selection_internal.h
felipecrv: Renamed to `PreallocatePrimitiveArrayData`. It's here because it's used by both `take` and `filter` on primitive arrays.

cpp/src/arrow/compute/kernels/vector_selection_filter.h
```diff
@@ -0,0 +1,39 @@
...
+// under the License.
+
+#pragma once
+
+#include <cstdint>
+#include <memory>
+
+#include "arrow/array/data.h"
+#include "arrow/compute/api_vector.h"
+#include "arrow/compute/kernels/vector_selection_internal.h"
```
kou, line 25: We should not include `*internal.h` from a non `*internal.h` headers because `*internal.h` aren't installed.

felipecrv: `vector_selection_filter.h` is also internal — included only by `vector_selection.cc`. Should I rename it with an `internal` suffix in the name?

felipecrv: But I just realized it's not needed. :smile:  This is why I made `PopulateFilterKernels` take an output param at some point. That allows me to forward-declare `SelectionKernelData` and not include the full header.

felipecrv: As it's been done with `vector_selection_take.h`

assignUser: IIRC the `internal.h` suffix is important as the cmake uses it to decide which headers get installed. So if it is actually a private/internal header it should be renamed imo. But @kou would know best!

kou: > IIRC the `internal.h` suffix is important as the cmake uses it to decide which headers get installed. So if it is actually a private/internal header it should be renamed imo.

Correct.

felipecrv: Makes sense. I'm also renaming `vector_selection_(filter|take).(h|cc)` and pushing now.

pitrou:
@felipecrv Is this ready for review again?

felipecrv:
> @felipecrv Is this ready for review again?

Yes.

pitrou (approved):
+1, thanks for doing this @felipecrv

PR #35758, opened 2023-05-25
GH-35757: [C++][Parquet] Build encodings metadata from actual encodings

### LINKED ISSUE #35757: [C++][Parquet] ColumnChunkMetadata: using `encoding_stats` to build `encodings`

Currently, the logic of building `ColumnChunkMetadata ` is complex, introduce so much unneccessary logics.

Here I'd like to using `encoding_stats` to build the content of `encodings`.

### Component(s)

C++, Parquet

mapleFU:
### Rationale for this change

This patch uses `encoding_stats` to build the encodings. Currently, `ColumnChunk.encodings` is repeated ( May has multiple `PLAIN` ) and having some wired logic by getting `WriterProperties`. And if later we want to support Fallback to non "PLAIN" encoding, `encoding_stats` would be right, but `encoding` would be "PLAIN". So this patch using `encoding_stats` to build the `encoding`, and force them to be consistent.

### Are these changes tested?

Already has test

Some behavior would be changed, but I think my change is right :)

* Closes: #35757


cpp/src/parquet/metadata.cc
```diff
@@ -1462,40 +1462,45 @@ class ColumnChunkMetaDataBuilder::ColumnChunkMetaDataBuilderImpl {
     column_chunk_->meta_data.__set_total_compressed_size(compressed_size);
 
     std::vector<format::Encoding::type> thrift_encodings;
+    std::vector<format::PageEncodingStats> thrift_encoding_stats;
+    // Force Emplace encoding for RL/DL.
+    thrift_encodings.push_back(ToThrift(Encoding::RLE));
```
wgtmac: Is this true? I think `BIT_PACKED` is still supported.

wgtmac: https://github.com/apache/arrow/blob/6d2df074e624a6a4462a1539be8dadf19cf39df4/cpp/src/parquet/column_writer.cc#L182-L186

cpp/src/parquet/metadata.cc
```diff
@@ -1462,40 +1462,45 @@ class ColumnChunkMetaDataBuilder::ColumnChunkMetaDataBuilderImpl {
...
-    for (const auto& entry : dict_encoding_stats) {
-      format::PageEncodingStats dict_enc_stat;
-      dict_enc_stat.__set_page_type(format::PageType::DICTIONARY_PAGE);
-      dict_enc_stat.__set_encoding(ToThrift(entry.first));
-      dict_enc_stat.__set_count(entry.second);
-      thrift_encoding_stats.push_back(dict_enc_stat);
+      for (const auto& entry : dict_encoding_stats) {
+        format::PageEncodingStats dict_enc_stat;
+        dict_enc_stat.__set_page_type(format::PageType::DICTIONARY_PAGE);
+        format::Encoding::type dict_encoding = ToThrift(entry.first);
```
wgtmac, line 1471: CMIW, `dict_encoding` here will be `PLAIN_DICTIONARY` in v1 and `PLAIN` in v2?

cpp/src/parquet/metadata.cc
```diff
@@ -1462,40 +1462,45 @@ class ColumnChunkMetaDataBuilder::ColumnChunkMetaDataBuilderImpl {
...
 
     std::vector<format::Encoding::type> thrift_encodings;
+    std::vector<format::PageEncodingStats> thrift_encoding_stats;
+    // Force Emplace encoding for RL/DL.
+    thrift_encodings.push_back(ToThrift(Encoding::RLE));
+    // Add dictionary page encoding stats
     if (has_dictionary) {
-      thrift_encodings.push_back(ToThrift(properties_->dictionary_index_encoding()));
+      // For Parquet V1, Dictionary Data Page and Dictionary Index Page encodings
+      // is PLAIN_DICTIONARY, but the actual DATA is PLAIN, so force adding PLAIN.
```
wgtmac:
```suggestion
      // For dictionary encoding in the v1 format, both dictionary page and data
      // page uses PLAIN_DICTIONARY. Since the dictionary page applies PLAIN
      // encoding, we should always add it.
```

wgtmac: What will happen if `PLAIN` is not added to the thrift_encodings list?

wgtmac: I was thinking if this is a wrong behavior which has existed for a long time. Actually in the parquet file here, it only shows `PLAIN_DICTIONARY`, no `PLAIN` in the encoding list.

https://github.com/apache/arrow/blob/6bd31f37ae66bd35594b077cb2f830be57e08acd/cpp/src/parquet/reader_test.cc#L788-L797

cpp/src/parquet/metadata.cc
mapleFU: Oops, I just keep the logic here as before. It's ok to add or remove it :-)

cpp/src/parquet/metadata.cc
mapleFU: I think this just prints the EncodingStats?

wgtmac: Yes, if `EncodingStats` exists it only prints it. Otherwise it prints the encoding list.

mapleFU: So confusion...

cpp/src/parquet/metadata.cc
mapleFU: Yes ...

mapleFU:
```
=== sccache stats after the build ===
sccache: error: Timed out waiting for server startup
```

emm..

pitrou:
Can you expand the PR description to explain a bit better what this is about? I don't understand just by looking at the code.

mapleFU:
@pitrou This patch uses `encoding_stats` to build the encodings. Currently, `ColumnChunk.encodings` is repeated ( May has multiple `PLAIN` ) and having some wired logic by getting `WriterProperties`. And if later we want to support Fallback to non "PLAIN" encoding, `encoding_stats` would be right, but `encoding` would be "PLAIN". So this patch using `encoding_stats` to build the `encoding`, and force them to be consistent. I also remove some duplicated encoding here.

cpp/src/parquet/metadata.cc
```diff
@@ -1462,40 +1462,48 @@ class ColumnChunkMetaDataBuilder::ColumnChunkMetaDataBuilderImpl {
...
+        if (iter == thrift_encodings.end()) {
+          thrift_encodings.push_back(dict_encoding);
+        }
       }
-    } else {  // Dictionary not enabled
-      thrift_encodings.push_back(ToThrift(properties_->encoding(column_->path())));
+      // Dictionary Data Page Encoding.
+      thrift_encodings.push_back(ToThrift(properties_->dictionary_page_encoding()));
     }
+    // Force add encoding for RL/DL.
```
pitrou, line 1484: Even if there are no RL/DL?

cpp/src/parquet/metadata.cc
pitrou: Can you add a comment for that?

cpp/src/parquet/metadata.cc
```diff
@@ -1462,40 +1462,48 @@ class ColumnChunkMetaDataBuilder::ColumnChunkMetaDataBuilderImpl {
...
-      data_enc_stat.__set_encoding(ToThrift(entry.first));
+      format::Encoding::type data_encoding = ToThrift(entry.first);
+      data_enc_stat.__set_encoding(data_encoding);
       data_enc_stat.__set_count(entry.second);
       thrift_encoding_stats.push_back(data_enc_stat);
+      if (data_encoding == format::Encoding::PLAIN_DICTIONARY &&
+          properties_->version() == ParquetVersion::PARQUET_1_0) {
+        // For Parquet V1, Dictionary Data Page and Dictionary Index Page encodings
+        // is PLAIN_DICTIONARY, but the actual DATA is PLAIN, so force adding PLAIN.
+        data_encoding = format::Encoding::PLAIN;
```
pitrou, line 1498: So we are not putting the same values in `encodings` and `encoding_stats`? What does parquet-mr do?

cpp/src/parquet/metadata_test.cc
```diff
@@ -66,7 +66,6 @@ std::unique_ptr<parquet::FileMetaData> GenerateTableMetaData(
   // column metadata
   col1_builder->SetStatistics(stats_int);
   col2_builder->SetStatistics(stats_float);
-  dict_encoding_stats.clear();
```
pitrou, line 69: Why remove this line? Since `/*has_dictionary=*/false` below, perhaps we need to pass empty dict encoding stats?

cpp/src/parquet/metadata_test.cc
```diff
@@ -209,7 +208,7 @@ TEST(Metadata, TestBuildAccess) {
     ASSERT_EQ(10, rg2_column1->data_page_offset());
     ASSERT_EQ(26, rg2_column2->data_page_offset());
     ASSERT_EQ(2, rg2_column1->encoding_stats().size());
-    ASSERT_EQ(2, rg2_column2->encoding_stats().size());
+    ASSERT_EQ(3, rg2_column2->encoding_stats().size());
```
pitrou, line 211: Can we also check the actual encodings values in this test?

mapleFU: Sure, will check it tomorrow

mapleFU:
I go through the Parquet-mr's code, and I think they're doing the better:

```Java
  void writeColumnChunk(ColumnDescriptor descriptor,
      long valueCount,
      CompressionCodecName compressionCodecName,
      DictionaryPage dictionaryPage,
      BytesInput bytes,
      long uncompressedTotalPageSize,
      long compressedTotalPageSize,
      Statistics<?> totalStats,
      ColumnIndexBuilder columnIndexBuilder,
      OffsetIndexBuilder offsetIndexBuilder,
      BloomFilter bloomFilter,
      Set<Encoding> rlEncodings,
      Set<Encoding> dlEncodings,
      List<Encoding> dataEncodings,
      BlockCipher.Encryptor headerBlockEncryptor,
      int rowGroupOrdinal,
      int columnOrdinal,
      byte[] fileAAD) throws IOException {
    startColumn(descriptor, valueCount, compressionCodecName);

    state = state.write();
    if (dictionaryPage != null) {
      byte[] dictonaryPageHeaderAAD = null;
      if (null != headerBlockEncryptor) {
        dictonaryPageHeaderAAD = AesCipher.createModuleAAD(fileAAD, ModuleType.DictionaryPageHeader,
            rowGroupOrdinal, columnOrdinal, -1);
      }
      writeDictionaryPage(dictionaryPage, headerBlockEncryptor, dictonaryPageHeaderAAD);
    }

    if (bloomFilter != null) {
      // write bloom filter if one of data pages is not dictionary encoded
      boolean isWriteBloomFilter = false;
      for (Encoding encoding : dataEncodings) {
        if (encoding != Encoding.RLE_DICTIONARY) {
          isWriteBloomFilter = true;
          break;
        }
      }
      if (isWriteBloomFilter) {
        currentBloomFilters.put(String.join(".", descriptor.getPath()), bloomFilter);
      }
    }
    LOG.debug("{}: write data pages", out.getPos());
    long headersSize = bytes.size() - compressedTotalPageSize;
    this.uncompressedLength += uncompressedTotalPageSize + headersSize;
    this.compressedLength += compressedTotalPageSize + headersSize;
    LOG.debug("{}: write data pages content", out.getPos());
    currentChunkFirstDataPage = out.getPos();
    bytes.writeAllTo(out);
    encodingStatsBuilder.addDataEncodings(dataEncodings);
    if (rlEncodings.isEmpty()) {
      encodingStatsBuilder.withV2Pages();
    }
    currentEncodings.addAll(rlEncodings);
    currentEncodings.addAll(dlEncodings);
    currentEncodings.addAll(dataEncodings);
    currentStatistics = totalStats;

    this.columnIndexBuilder = columnIndexBuilder;
    this.offsetIndexBuilder = offsetIndexBuilder;

    endColumn();
  }
```

Parquet-mr's `ColumnChunkPageWriteStore` will record that:
1. When `WritePage` is called, what rl/dl/data encoding is used
2. They will be collected in a `Set`, and be written to column chunk

For RL/DL, it has the logic below in `ColumnWriterV1`:

```
  private ValuesWriter newColumnDescriptorValuesWriter(int maxLevel) {
    if (maxLevel == 0) {
      return new DevNullValuesWriter();
    } else {
      return new RunLengthBitPackingHybridValuesWriter(
          getWidthFromMaxInt(maxLevel), MIN_SLAB_SIZE, pageSizeThreshold, allocator);
    }
  }
```

and in v2:

```
  @Override
  ValuesWriter createRLWriter(ParquetProperties props, ColumnDescriptor path) {
    return path.getMaxRepetitionLevel() == 0 ? NULL_WRITER : new RLEWriterForV2(props.newRepetitionLevelEncoder(path));
  }

  @Override
  ValuesWriter createDLWriter(ParquetProperties props, ColumnDescriptor path) {
    return path.getMaxDefinitionLevel() == 0 ? NULL_WRITER : new RLEWriterForV2(props.newDefinitionLevelEncoder(path));
  }
```

So, I guess, if not has rl/dl, the encoding would not be collected. It's wired that, some when it's `DevNullValuesWritter`, the level is `BIT_PACKED`:

```
  @Override
  public Encoding getEncoding() {
    return BIT_PACKED;
  }
```

For Dictionary, I guess they don't extra handle the dictionary logic.

cpp/src/parquet/metadata.cc
mapleFU: The behavior is pasted here: https://github.com/apache/arrow/pull/35758#issuecomment-1565248796

cpp/src/parquet/metadata.cc
mapleFU: Done

mapleFU:
@pitrou @wgtmac I've pasted the parquet-mr logic here

cpp/src/parquet/metadata_test.cc
```diff
@@ -160,8 +159,22 @@ TEST(Metadata, TestBuildAccess) {
     ASSERT_EQ(DEFAULT_COMPRESSION_TYPE, rg1_column2->compression());
     ASSERT_EQ(nrows / 2, rg1_column1->num_values());
     ASSERT_EQ(nrows / 2, rg1_column2->num_values());
-    ASSERT_EQ(3, rg1_column1->encodings().size());
-    ASSERT_EQ(3, rg1_column2->encodings().size());
+    {
+      std::set<parquet::Encoding::type> encodings{parquet::Encoding::RLE,
+                                                  parquet::Encoding::RLE_DICTIONARY,
+                                                  parquet::Encoding::PLAIN};
+      auto& encoding_vec = rg1_column1->encodings();
```
pitrou, line 166: Nit: style
```suggestion
      const auto& encoding_vec = rg1_column1->encodings();
```

cpp/src/parquet/metadata_test.cc
```diff
@@ -160,8 +159,22 @@ TEST(Metadata, TestBuildAccess) {
...
+      ASSERT_EQ(encodings, std::set<parquet::Encoding::type>(encoding_vec.begin(),
+                                                             encoding_vec.end()));
+    }
+    {
+      std::set<parquet::Encoding::type> encodings{parquet::Encoding::RLE,
+                                                  parquet::Encoding::RLE_DICTIONARY,
+                                                  parquet::Encoding::PLAIN};
+      auto& encoding_vec = rg1_column2->encodings();
```
pitrou, line 174:
```suggestion
      const auto& encoding_vec = rg1_column2->encodings();
```

cpp/src/parquet/metadata_test.cc
```diff
@@ -160,8 +159,22 @@ TEST(Metadata, TestBuildAccess) {
...
+      ASSERT_EQ(encodings, std::set<parquet::Encoding::type>(encoding_vec.begin(),
+                                                             encoding_vec.end()));
+    }
```
pitrou, line 177: Perhaps define and use a helper function:
```c++
void AssertEncodings(const ColumnChunkMetaData&, const std::set<parquet::Encoding::type>& expected);
```

cpp/src/parquet/metadata.cc
pitrou: I still don't understand why you're doing this. Here is what parquet-mr does for v1 and v2 data pages:
https://github.com/apache/parquet-mr/blob/f15b9c42997e3e0af991462529942e044ace6651/parquet-column/src/main/java/org/apache/parquet/column/values/factory/DefaultV1ValuesWriterFactory.java#L41-L47

https://github.com/apache/parquet-mr/blob/f15b9c42997e3e0af991462529942e044ace6651/parquet-column/src/main/java/org/apache/parquet/column/values/factory/DefaultV2ValuesWriterFactory.java#L45-L51

cpp/src/parquet/metadata.cc
```diff
@@ -1462,40 +1462,50 @@ class ColumnChunkMetaDataBuilder::ColumnChunkMetaDataBuilderImpl {
...
+      for (const auto& entry : dict_encoding_stats) {
+        format::PageEncodingStats dict_enc_stat;
+        dict_enc_stat.__set_page_type(format::PageType::DICTIONARY_PAGE);
+        // Dictionary Encoding would be PLAIN_DICTIONARY in v1 and
+        // PLAIN in v2.
+        format::Encoding::type dict_encoding = ToThrift(entry.first);
+        dict_enc_stat.__set_encoding(dict_encoding);
+        dict_enc_stat.__set_count(entry.second);
+        thrift_encoding_stats.push_back(dict_enc_stat);
+        auto iter =
```
pitrou, line 1477: Can define a small helper function:
```c++
auto add_encoding = [&thrift_encodings](format::Encoding::type value) {
  auto it = std::find(thrift_encodings.begin(), thrift_encodings.end(), value);
  if (it == thrift_encodings.end()) {
    thrift_encodings.push_back(value);
  }
};
```

pitrou (changes requested):

cpp/src/parquet/metadata.cc
mapleFU: Okay, this just compatible with previous logic: https://github.com/apache/arrow/blob/main/cpp/src/parquet/metadata.cc#L1467-L1470

I think I can remove it

pitrou: I think so, yes.

cpp/src/parquet/metadata.cc
```diff
@@ -1462,40 +1462,42 @@ class ColumnChunkMetaDataBuilder::ColumnChunkMetaDataBuilderImpl {
...
+        dict_enc_stat.__set_count(entry.second);
+        thrift_encoding_stats.push_back(dict_enc_stat);
+        add_encoding(dict_encoding);
       }
-    } else {  // Dictionary not enabled
-      thrift_encodings.push_back(ToThrift(properties_->encoding(column_->path())));
     }
+    // Force add encoding for RL/DL.
+    // TODO(mwish): BIT_PACKED encoding is supported, but it will not be
+    //  added here. And when RL/DL is empty, it will not be added.
```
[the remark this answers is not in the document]
mapleFU: Sorry for didn't make it clear, I've update it to

```
    // FIXME(mwish): BIT_PACKED encoding is supported, but it would not be
    //  added here even in format v1.
    //  And when RL/DL is empty, RLE would be added.
```

cpp/src/parquet/metadata.cc
```diff
@@ -1462,40 +1462,44 @@ class ColumnChunkMetaDataBuilder::ColumnChunkMetaDataBuilderImpl {
...
+        dict_enc_stat.__set_count(entry.second);
+        thrift_encoding_stats.push_back(dict_enc_stat);
+        add_encoding(dict_encoding);
+      }
     }
+    // Force add encoding for RL/DL.
+    // BIT_PACKED is supported in `LevelEncoder`, but would only be used
+    // in benchmark and testing.
+    // And even when RL/DL is empty(`column_->max_definition_level() == 0`),
+    // RLE would be added here.
```
pitrou, line 1490:
```suggestion
    // And for now, we always add RLE even if there are no levels at all,
    // while parquet-mr is more fine-grained.
```

pitrou (approved):

cpp/src/parquet/metadata.cc
```diff
@@ -1462,40 +1462,44 @@ class ColumnChunkMetaDataBuilder::ColumnChunkMetaDataBuilderImpl {
...
+        // Dictionary Encoding would be PLAIN_DICTIONARY in v1 and
+        // PLAIN in v2.
+        format::Encoding::type dict_encoding = ToThrift(entry.first);
+        dict_enc_stat.__set_encoding(dict_encoding);
+        dict_enc_stat.__set_count(entry.second);
+        thrift_encoding_stats.push_back(dict_enc_stat);
+        add_encoding(dict_encoding);
+      }
     }
+    // Force add encoding for RL/DL.
```
wgtmac, line 1486:
```suggestion
    // Always add encoding for RL/DL.
```

wgtmac (approved):

PR #35762, opened 2023-05-25
GH-35761: [Go] Fix map comparison in TypeEqual

hermanschaaf:
Fixes some edge cases when doing type comparison for the map type.
* Closes: #35761


zeroshade (approved):
LGTM

PR #35764, opened 2023-05-25
GH-35763: [Go] Fix TypeEqual for lists

### LINKED ISSUE #35763: [Go] TypeEqual doesn't compare nullability in some cases

Due to a bug in the code, when checking type equivalence for two lists and also checking metadata, TypeEqual returns true when nullability differs between the lists.

### Component(s)

Go

hermanschaaf:
This fixes an edge case for list comparisons where TypeEqual would return true even when the nullability of the lists differed.
* Closes: #35763


zeroshade (approved):
LGTM thanks!

Issue #34898, opened 2023-04-05
[Python] Provide a way to restore a schema from its string representation


barbuz:
### Motivation
I need to store a schema in a permanent way, as I'm building a process where several CSV files with the same structure will be converted to Parquet over time, but inferring the schema of the data from each file does not guarantee consistency. I could use something like pickle to dump a binary version of the schema but this is not very portable, and a human-readable representation would allow to do simple modifications by hand if the need arises (e.g. adding or removing columns). For this reason I'm trying to save the schema in a json file and then to read it and re-build a pyarrow Schema. 

The first step is easy thanks to each type having a nice string representation, but going backwards is harder as I could not find any way of building a type or a Field from a string that did not break on more complex types such as `timestep[ms, tz=utc]`. I ended up having to implement my own function to parse string representations of types and build the appropriate pyarrow objects.

### Idea
The basic idea would be implementing a function that takes a string representation of a type and returns the corresponding pyarrow type object. 
Other things that could make this process easier would be a `to_dict()` function for Schema that basically builds the dictionary obtained combining the `names` and `types` lists of a Schema, or a `from_string(s)` function that can reverse the `to_string()` function.

I can contribute my code to convert strings to type objects, but as I've never contributed to this project before I would like some advice on whether this is something desired and what would be the best way to integrate this functionality with the rest of the code base.

### Component(s)

Python

danepitkin:
Hey @barbuz,

It is a neat idea, but I would vote not to support this in arrow itself for a few reasons:
1) string representations are mostly seen as debug output
2) maintaining compatibility over time adds complexity
3) it isn't too hard to implement your own serialization/deserialization

Either way, feel free to check out good first issues in python if you want to become a contributor! https://github.com/apache/arrow/issues?q=is%3Aopen+is%3Aissue+label%3Agood-first-issue+label%3A%22Component%3A+Python%22

westonpace:
The Arrow IPC format defines how to serialize a schema.  You can use an empty IPC file to represent a schema in a portable way:

```
>>> import pyarrow as pa
>>> import pyarrow.ipc
>>> my_schema = pa.schema([pa.field("x", pa.timestamp("ms")), pa.field("y", pa.int32())])
>>> table = pa.Table.from_batches([], schema=my_schema)
>>> with pyarrow.ipc.RecordBatchFileWriter("/tmp/schema.arrow", schema=my_schema) as writer:
...   writer.write_table(table)
... 
[...]

PR #7110, opened 2020-05-05
ARROW-8952: [C++] WIP Support for textual, JSON schema representation

chrish42:
Draft pull request for discussion of possible text schema for Arrow, using Flatbuffers' existing serialization to JSON feature (via `flatbuffers::GenerateText()`, etc.) on the Arrow schema. Not intented to necessarily hint at what the final code would look like, but to allow people to discuss the general approach and see examples of the format.

This approach has the advantages of not introducing any new dependencies, and reusing existing Flatbuffers for much of the feature.

Once there is agreement on this approach, I can then polish this into a mergeable state. In the meantime, the (temporary) `arrow-schema` program takes a hard-coded schema, and outputs the (proposed) corresponding JSON, so people can see what the text schemas would look like with this proposal.

As an example, the following Arrow schema definition:
```
  std::vector<std::shared_ptr<arrow::Field>> schema_vector = {
    arrow::field("id", arrow::int64()),
    arrow::field("cost", arrow::float64()),
    arrow::field("cost_components", arrow::list(arrow::float64()))};
  auto schema = arrow::Schema(schema_vector);
```
translates to:
```
{
  fields: [
    {
      name: "id",
      nullable: true,
      type_type: "Int",
      type: {
        bitWidth: 64,
        is_signed: true
      },
      children: [

      ]
    },
    {
      name: "cost",
      nullable: true,
      type_type: "FloatingPoint",
      type: {
        precision: "DOUBLE"
      },
      children: [

      ]
    },
    {
      name: "cost_components",
      nullable: true,
      type_type: "List",
      type: {
      },
      children: [
        {
          name: "item",
          nullable: true,
          type_type: "FloatingPoint",
          type: {
            precision: "DOUBLE"
          },
          children: [

          ]
        }
      ]
    }
  ]
}
```

Seems pretty straightforward to read for anyone who knows about Arrow's data representation to me. A bit verbose for the simpler cases, however. Maybe we would need to add a couple extra keys at the top level... Versioning for the text schema? Something that says this is an Arrow schema? Or maybe that's not necessary.

PR #35769, opened 2023-05-25
GH-38718: [Go][Format][Integration] Add StringView/BinaryView to Go implementation

### LINKED ISSUE #38718: [Go][Integration] Add support for Utf8View to the go implementation

Utf8View was added to the format in
https://github.com/apache/arrow/issues/35627 and c++ support added in https://github.com/apache/arrow/pull/37792

### Component(s)

Go, Integration

zeroshade:
### Rationale for this change
See #35628 for the rationale and description of the StringView/BinaryView array types.

This change is adding Go as a second implementation of it.

Add Array Types for `StringView` and `BinaryView` along with `StringViewType` and `BinaryViewType` and necessary enums and builders. These arrays can be round tripped through JSON and IPC.

### Are these changes tested?
Yes, unit tests have been added and integration tests run

* Closes: [#38718](https://github.com/apache/arrow/issues/38718)
* Closes: #38718


go/arrow/datatype_binary.go
```diff
@@ -83,16 +93,168 @@ func (t *LargeStringType) Layout() DataTypeLayout {
...
+//
+//	IO Long String |----|----|----|----|
+//	                ^    ^     ^     ^
+//	                |    |     |     |
+//	              size prefix buffer index and offset to out-of-line portion
+//
+// Adapted from TU Munich's UmbraDB [1], Velox, DuckDB.
+//
+// [1]: https://db.in.tum.de/~freitag/papers/p29-neumann-cidr20.pdf
+type StringHeader struct {
```
bkietz, line 128: I'm not familiar enough with Go to know whether you can attach alignment requirements to a struct. Reading this it seems that it'd be only aligned to a 4 byte boundary, which would cause problems for the c++ impl where we expect them to be aligned to 8 bytes.

zeroshade: The only way to modify the alignment requirements of a struct in go is to modify the types of its fields. It is going to pick the alignment for the struct based on the largest alignment needed for a given member. In this case using `unsafe.Alignof(arrow.StringHeader{})` reports an alignment of 4 bytes. That said, as long as the size of 16 bytes for the struct is maintained, I don't think there'll be an issue? In addition I believe we explicitly handle the alignment for IPC input/output.

I guess there could be an issue with the C-Data interface if the buffer ends up not 8-byte aligned? but then again, I don't think you added a proposal for the new format string to indicate this across the C-Data interface yet :smile:

go/arrow/datatype_binary.go
```diff
@@ -83,16 +93,168 @@ func (t *LargeStringType) Layout() DataTypeLayout {
...
+}
+
+type BinaryViewType struct{}
+
+func (*BinaryViewType) ID() Type              { return BINARY_VIEW }
+func (*BinaryViewType) Name() string          { return "binary_view" }
+func (*BinaryViewType) String() string        { return "binary_view" }
+func (*BinaryViewType) IsUtf8() bool          { return false }
+func (*BinaryViewType) binary()               {}
+func (*BinaryViewType) view()                 {}
```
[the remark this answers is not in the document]
zeroshade: convenience interfaces. in `datatype.go` there's `BinaryDataType` which defines an interface adding `IsUtf8() bool` which requires a `binary()` method so that you can easily write code that works for *all* of these binary interfaces via type switches or interfaces.

Then I added this `view()` method for a `BinaryViewDataType` so you can write code targeting both BinaryView and StringView without needing to duplicate code by just using the interface in the type switch.

go/arrow/datatype_binary.go
```diff
@@ -83,16 +93,168 @@ func (t *LargeStringType) Layout() DataTypeLayout {
...
+func (sh *StringHeader) BufferIndex() uint32 {
+	return endian.Native.Uint32(sh.data[StringHeaderPrefixLen:])
+}
+
+func (sh *StringHeader) BufferOffset() uint32 {
+	return endian.Native.Uint32(sh.data[StringHeaderPrefixLen+4:])
+}
+
+func (sh *StringHeader) InlineBytes() (data []byte) {
+	debug.Assert(sh.IsInline(), "calling InlineBytes on non-inline StringHeader")
```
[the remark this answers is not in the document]
zeroshade: fair point, i'll add those

go/arrow/datatype_binary.go
zeroshade: That's what I was thinking and is why i didn't do that in the first place. I mean at that rate why not just make the entire thing `[16]byte` right? lol

bkietz: You're correct that the alignment only really matters if Go is communicating with C++ through shared memory or the C-ABI, so getting it up to 8 is currently speculative generality at this point. That said, adding string view to the C-ABI is a pretty natural follow up to this work so I think it'd be worthwhile to be prepared. I won't insist for now

zeroshade:
CC @felipecrv @benibus

lquerel (approved):
Interesting optimization. I'm not entirely sure why these new types of builders/arrays are named XXXView, considering that, from what I understand, they store small strings inline in a fixed width data array and encode the index/offset for longer strings, with the remainder of the string being stored externally. However, I don't have a better name to suggest at the moment :)

In general, as a user of this library, I would find it useful to add more comments to methods whose usage isn't immediately clear from the name.

go/arrow/array/binary.go
```diff
@@ -318,6 +319,116 @@ func arrayEqualLargeBinary(left, right *LargeBinary) bool {
...
+}
+
+func (a *BinaryView) ValueHeader(i int) *arrow.StringHeader {
+	if i < 0 || i >= a.array.data.length {
+		panic("arrow/array: index out of range")
+	}
+	return &a.values[a.array.data.offset+i]
+}
+
+func (a *BinaryView) Value(i int) []byte {
```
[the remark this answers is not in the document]
felipecrv: This is a concern present everywhere in Arrow though, so a comment here could be understood as implying that places without a comment like this allow buffers to be mutated.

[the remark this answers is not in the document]
lquerel: This is probably a reflex from my Rust experience where this slice will be immutable. In the context of Go, I agree with you that sporadically adding a comment might be counterproductive if we do not apply this globally.

go/arrow/array/binary.go
```diff
@@ -318,6 +319,116 @@ func arrayEqualLargeBinary(left, right *LargeBinary) bool {
...
+	}
+	// golang marshal standard says that []byte will be marshalled
+	// as a base64-encoded string
+	return json.Marshal(vals)
+}
+
+func arrayEqualBinaryView(left, right *BinaryView) bool {
+	leftBufs, rightBufs := left.dataBuffers, right.dataBuffers
+	for i := 0; i < left.Len(); i++ {
+		if left.IsNull(i) {
```
felipecrv, line 420: Shouldn't `right` also be Null here for the slot to be skipped?

go/arrow/datatype_stringheader.go
```diff
@@ -0,0 +1,136 @@
...
+
+package arrow
+
+import (
+	"bytes"
+	"unsafe"
+
+	"github.com/apache/arrow/go/v13/arrow/endian"
+	"github.com/apache/arrow/go/v13/arrow/internal/debug"
+	"github.com/apache/arrow/go/v13/arrow/memory"
```
felipecrv, line 25: v13 here needs to be updated on every release?

go/arrow/array/binary.go
candiduslynx: this is already checked in the `baseArrayEqual` func (~ len, data type, nulls are same)

go/arrow/array/binary.go
```diff
@@ -318,6 +319,116 @@ func arrayEqualLargeBinary(left, right *LargeBinary) bool {
...
+
+type BinaryView struct {
+	array
+	values      []arrow.StringHeader
+	dataBuffers []*memory.Buffer
+}
+
+func NewBinaryViewData(data arrow.ArrayData) *BinaryView {
+	a := &BinaryView{}
+	a.refCount = 1
```
candiduslynx, line 335:
```suggestion
	a := &BinaryView{refCount: 1}
```

go/arrow/array/binary.go
```diff
@@ -318,6 +319,116 @@ func arrayEqualLargeBinary(left, right *LargeBinary) bool {
...
+			o.WriteString(NullValueStr)
+		default:
+			fmt.Fprintf(&o, "%q", a.ValueString(i))
+		}
+	}
+	o.WriteString("]")
+	return o.String()
+}
+
+func (a *BinaryView) ValueStr(i int) string {
```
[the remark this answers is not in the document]
candiduslynx: `ValueStr` is paired with `AppendValueFromString` and has a contract of being able to reconstruct the same data with `b.AppendValueFromString(a.ValueStr(i))`

candiduslynx (commented):
a first round

go/arrow/array/binary.go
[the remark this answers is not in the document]
zeroshade: At the top level, there do already exist comments that state that it is intended that all Arrow Arrays be immutable. I agree with the concern that adding a comment here specifically could be counterproductive. If you can think of a good place to put such a comment that would be more universal, I'd be more than happy to do so.

go/arrow/datatype_stringheader.go
zeroshade: Yea, the version bumping scripts already do find and replace for all the files. Ideally we'd be able to update our release process to do more minor releases and fewer major releases (leaving major releases *only* for breaking changes) but that's an entirely different conversation.

go/arrow/internal/flatbuf/BinaryView.go
```diff
@@ -0,0 +1,52 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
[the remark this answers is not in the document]
zeroshade: this is auto generated via calling `flatc`, so the pascal case is what the Flatbuffers compiler is generating.

go/arrow/array/binary.go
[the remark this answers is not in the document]
zeroshade: I'll add the clarifying comments

dev/archery/archery/integration/datagen.py
```diff
@@ -743,6 +763,72 @@ class LargeStringColumn(_BaseStringColumn, _LargeOffsetsMixin):
...
+                    char_buffers.append(v)
+                else:
+                    if len(char_buffers) == 0:
+                        char_buffers.append(v)
+                    elif len(char_buffers[-1]) + len(v) > DEFAULT_BUFFER_SIZE:
+                        char_buffers.append(v)
+                    else:
+                        offset = len(char_buffers[-1])
+                        char_buffers[-1] += v
+                        assert len(char_buffers[-1]) <= DEFAULT_BUFFER_SIZE
```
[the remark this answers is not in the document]
zeroshade: @bkietz Is there a reason for this assert to exist given the if/elif clauses above it? I pretty much just took this from your C++ PR for the integration tests.

go/arrow/datatype_stringheader_inline_tinygo.go
```diff
@@ -0,0 +1,35 @@
...
+	"reflect"
+	"unsafe"
+
+	"github.com/apache/arrow/go/v13/arrow/internal/debug"
+)
+
+func (sh *StringHeader) InlineData() (data string) {
+	debug.Assert(sh.IsInline(), "calling InlineData on non-inline StringHeader")
+
+	h := (*reflect.StringHeader)(unsafe.Pointer(&data))
```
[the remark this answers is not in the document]
zeroshade: yea. The `Len` in the `StringHeader` for tinygo is a `uintptr` instead of an `int` which is what go1.19 has. Compare line 33 of each file.

go/arrow/array/binarybuilder.go
```diff
@@ -358,6 +359,301 @@ func (b *BinaryBuilder) UnmarshalJSON(data []byte) error {
...
+			if !hdr.IsInline() {
+				b.blockBuilder.UnsafeAppendString(hdr, vv)
+			}
+		}
+	}
+
+	b.builder.unsafeAppendBoolsToBitmap(valid, len(v))
+}
+
+func (b *BinaryViewBuilder) AppendValueFromString(s string) error {
```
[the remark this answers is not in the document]
zeroshade: done, let me know if you think it is sufficient or needs some more information

go/arrow/array/binary.go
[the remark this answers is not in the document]
zeroshade: done, let me know if you think it is sufficient or needs some more information

go/arrow/datatype_binary.go
```diff
@@ -83,16 +83,50 @@ func (t *LargeStringType) Layout() DataTypeLayout {
...
+func (*BinaryViewType) binary()               {}
+func (*BinaryViewType) view()                 {}
+func (t *BinaryViewType) Fingerprint() string { return typeFingerprint(t) }
+func (*BinaryViewType) Layout() DataTypeLayout {
+	variadic := SpecVariableWidth()
+	return DataTypeLayout{Buffers: []BufferSpec{SpecBitmap(),
+		SpecFixedWidth(StringHeaderSizeBytes)}, VariadicSpec: &variadic}
+}
+
+type StringViewType struct{}
```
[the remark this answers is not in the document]
zeroshade: done

go/arrow/datatype_binary.go
```diff
@@ -83,16 +83,50 @@ func (t *LargeStringType) Layout() DataTypeLayout {
 func (t *LargeStringType) OffsetTypeTraits() OffsetTraits { return Int64Traits }
 func (LargeStringType) IsUtf8() bool                      { return true }
 
+type BinaryViewType struct{}
```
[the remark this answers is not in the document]
zeroshade: done

go/arrow/datatype_stringheader_inline_tinygo.go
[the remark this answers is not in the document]
candiduslynx: how sad, but now I see why it's warranted.
However, a comment to note the diff for those implementations would be nice to have (not required, but nice to have nonetheless).

go/arrow/ipc/file_reader.go
```diff
@@ -430,13 +430,18 @@ func (src *ipcSource) fieldMetadata(i int) *flatbuf.FieldNode {
...
-	src     ipcSource
-	ifield  int
-	ibuffer int
-	max     int
-	memo    *dictutils.Memo
-	version MetadataVersion
+	src       ipcSource
+	ifield    int
+	ibuffer   int
+	ivariadic int
```
candiduslynx, line 441: maybe these should be grouped together into an embedded struct? (`iStruct` :chuckle:)

go/arrow/ipc/writer.go
```diff
@@ -605,6 +606,33 @@ func (w *recordEncoder) visit(p *Payload, arr arrow.Array) error {
...
+		typeWidth := int64(arrow.StringHeaderSizeBytes)
+		minLength := paddedLength(arrLen*typeWidth, kArrowAlignment)
+
+		switch {
+		case needTruncate(int64(data.Offset()), values, minLength):
+			// non-zero offset: slice the buffer
+			offset := int64(data.Offset()) * typeWidth
+			// send padding if available
+			len := minI64(bitutil.CeilByte64(arrLen*typeWidth), int64(values.Len())-offset)
+			values = memory.NewBufferBytes(values.Bytes()[offset : offset+len])
```
candiduslynx, line 622: shouldn't `memory.SliceBuffer` be used instead (for `value.Retain()` underneath?

go/arrow/internal/flatbuf/RecordBatch.go
```diff
@@ -128,8 +128,42 @@ func (rcv *RecordBatch) Compression(obj *BodyCompression) *BodyCompression {
 }
 
 /// Optional compression of the message body
+/// Some types such as Utf8View are represented using a variable number of buffers.
```
candiduslynx, line 131: I wonder why `///` is used instead of `//`

go/arrow/array/binarybuilder.go
```diff
@@ -370,6 +371,327 @@ func (b *BinaryBuilder) UnmarshalJSON(data []byte) error {
...
+	}
+
+	dataBuffers := b.blockBuilder.Finish()
+	data = NewData(b.dtype, b.length, append([]*memory.Buffer{
+		b.nullBitmap, b.data}, dataBuffers...), nil, b.nulls, 0)
+	b.reset()
+
+	if b.data != nil {
+		b.data.Release()
+		b.data = nil
```
candiduslynx, line 674: should `b.nullBitmap` be also released somewhere around here?

go/arrow/array/binarybuilder.go
```diff
@@ -370,6 +371,327 @@ func (b *BinaryBuilder) UnmarshalJSON(data []byte) error {
...
+		}
+	}
+	return
+}
+
+func (b *BinaryViewBuilder) NewBinaryViewArray() (a *BinaryView) {
+	data := b.newData()
+	a = NewBinaryViewData(data)
+	data.Release()
+	return
```
candiduslynx, line 687:
```suggestion
	data := b.newData()
	defer data.Release()
	return NewBinaryViewData(data)
```
?

go/arrow/array/bufferbuilder.go
```diff
@@ -151,3 +153,115 @@ func (b *bufferBuilder) unsafeAppend(data []byte) {
...
+// Release may be called simultaneously from multiple goroutines.
+func (b *multiBufferBuilder) Release() {
+	debug.Assert(atomic.LoadInt64(&b.refCount) > 0, "too many releases")
+
+	if atomic.AddInt64(&b.refCount, -1) == 0 {
+		for i, buf := range b.blocks {
+			buf.Release()
+			b.blocks[i] = nil
+		}
+	}
```
candiduslynx, line 183:
```suggestion
	if atomic.AddInt64(&b.refCount, -1) == 0 {
		b.Reset()
	}
```

go/arrow/datatype_stringheader.go
```diff
@@ -0,0 +1,136 @@
...
+	stringHeaderInlineSize = 12
+)
+
+func IsStringHeaderInline(length int) bool {
+	return length < stringHeaderInlineSize
+}
+
+// StringHeader is a variable length string (utf8) or byte slice with
+// a 4 byte prefix and inline optimization for small values (12 bytes
+// or fewer). This is similar to Go's standard string  but limited by
```
candiduslynx, line 39:
```suggestion
// or fewer). This is similar to Go's standard string but limited by
```

go/arrow/internal/flatbuf/RecordBatch.go
zeroshade: You'd have to ask the maintainers of `flatc` lol

go/arrow/internal/flatbuf/RecordBatch.go
```diff
@@ -128,8 +128,42 @@ func (rcv *RecordBatch) Compression(obj *BodyCompression) *BodyCompression {
...
+/// For each such Field in the pre-ordered flattened logical schema, there will be
+/// an entry in variadicCounts to indicate the number of extra buffers which belong
+/// to that Field.
+func (rcv *RecordBatch) MutateVariadicCounts(j int, n int64) bool {
+	o := flatbuffers.UOffsetT(rcv._tab.Offset(12))
+	if o != 0 {
+		a := rcv._tab.Vector(o)
+		return rcv._tab.MutateInt64(a+flatbuffers.UOffsetT(j*8), n)
+	}
+	return false
```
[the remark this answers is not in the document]
zeroshade: I'm not going to modify the generated code. Everything in this folder is generated by flatc from the .fbs files.

go/arrow/internal/testing/gen/random_array_gen.go
```diff
@@ -350,6 +350,40 @@ func (r *RandomArrayGenerator) LargeString(size int64, minLength, maxLength int6
...
+		}
+		return string(out)
+	}
+
+	for i := 0; i < lengths.Len(); i++ {
+		if lengths.IsValid(i) {
+			bldr.Append(gen(lengths.Value(i)))
+		} else {
+			bldr.AppendNull()
+		}
```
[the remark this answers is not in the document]
zeroshade: I'm not going to modify the generated code. Everything in this folder is generated by flatc from the .fbs files.

go/arrow/ipc/writer.go
zeroshade: good point! nice catch

go/arrow/internal/arrjson/arrjson.go
```diff
@@ -2179,3 +2230,113 @@ func durationToJSON(arr *array.Duration) []interface{} {
...
+		Inlined   *string `json:"INLINED,omitempty"`
+	}
+
+	o := make([]interface{}, arr.Len())
+	for i := range o {
+		hdr := arr.ValueHeader(i)
+		if hdr.IsInline() {
+			data := hdr.InlineData()
+			if isBinary {
+				data = strings.ToUpper(hex.EncodeToString(hdr.InlineBytes()))
```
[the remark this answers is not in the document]
zeroshade: The spec for the integration test is to specify that the binary data should be upper-cased hex characters. @bkietz can comment on whether or not the ToUpper is necessary and if we want to relax the spec.

go/arrow/internal/arrjson/arrjson.go
```diff
@@ -2179,3 +2230,113 @@ func durationToJSON(arr *array.Duration) []interface{} {
...
+			}
+			o[i] = StringHeader{
+				Size:    hdr.Len(),
+				Inlined: &data,
+			}
+		} else {
+			idx, off := int(hdr.BufferIndex()), int(hdr.BufferOffset())
+			prefix := hdr.Prefix()
+			encodedPrefix := strings.ToUpper(hex.EncodeToString(prefix[:]))
```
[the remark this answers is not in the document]
zeroshade: Let's keep this discussion to https://github.com/apache/arrow/pull/35769#discussion_r1238716504

go/arrow/array/binarybuilder.go
zeroshade: That is handled by the call to `b.reset()` above this

go/arrow/array/binarybuilder.go
zeroshade: where possible I actually prefer the manual release here as it is *technically* ever so slightly more performant to avoid the `defer` when it is not needed. If there were more lines between the call to `newData` and the return I'd do the defer, but since it's only one line I prefer to follow the same pattern that the rest of the `New*Array` methods are using. What do you think?

go/arrow/ipc/file_reader.go
zeroshade: since they are only used in the `arrayLoaderContext` i'm not sure what the benefit would be to doing so?

go/arrow/array/binarybuilder.go
candiduslynx: a nit, so I'm OK either way

go/arrow/ipc/file_reader.go
candiduslynx: a nit, so probably none

candiduslynx (approved):

zeroshade:
Thanks for the reviews @lquerel and @candiduslynx!!

go/arrow/internal/arrjson/arrjson.go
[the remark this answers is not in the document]
bkietz: I think relaxing the spec is not a trivial change; it'd require all implementers to update how they consume binary data. That said, although it'd be a lot of work updating the integration JSON (perhaps even including a real JSON schema :scream:) is definitely worth a follow up.

See also:
- https://github.com/apache/arrow/pull/7110
- https://github.com/apache/arrow/issues/34898

dev/archery/archery/integration/datagen.py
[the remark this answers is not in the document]
bkietz: This was intended to be didactic/a sanity check. It could probably be replaced with comments

go/arrow/datatype_viewheader.go
```diff
@@ -0,0 +1,138 @@
...
+//	                |----|----|----|----|
+//		                ^    ^     ^     ^
+//		                |    |     |     |
+//		              size prefix buffer index and offset to out-of-line portion
+//
+// Adapted from TU Munich's UmbraDB [1], Velox, DuckDB.
+//
+// [1]: https://db.in.tum.de/~freitag/papers/p29-neumann-cidr20.pdf
+type ViewHeader struct {
+	size uint32
```
bkietz, line 63: `size` is signed too now
```suggestion
	size int
```

go/arrow/datatype_viewheader_inline_tinygo.go
```diff
@@ -0,0 +1,35 @@
...
+package arrow
+
+import (
+	"reflect"
+	"unsafe"
+
+	"github.com/apache/arrow/go/v14/arrow/internal/debug"
+)
+
+func (sh *ViewHeader) InlineData() (data string) {
```
bkietz, line 28: The utility of these over `InlineBytes` isn't clear to me

go/arrow/datatype_viewheader.go
```diff
@@ -0,0 +1,138 @@
...
+func (sh *ViewHeader) BufferIndex() uint32 {
+	return endian.Native.Uint32(sh.data[StringViewPrefixLen:])
+}
+
+func (sh *ViewHeader) BufferOffset() uint32 {
+	return endian.Native.Uint32(sh.data[StringViewPrefixLen+4:])
+}
+
+func (sh *ViewHeader) InlineBytes() (data []byte) {
+	debug.Assert(sh.IsInline(), "calling InlineBytes on non-inline StringHeader")
```
bkietz, line 91: At the risk of offering an opinion well out of my expertise, would it be preferable to return nil here instead?
```suggestion
  if !sh.IsInline() {
    return nil
  }
```

go/arrow/internal/arrdata/arrdata.go
```diff
@@ -1155,6 +1156,65 @@ func makeRunEndEncodedRecords() []arrow.Record {
 	return recs
 }
 
+func makeStringViewRecords() []arrow.Record {
```
bkietz, line 1159: It might be worthwhile to ensure at least one chunk has multiple data buffers

go/arrow/array/bufferbuilder.go
```diff
@@ -151,3 +153,112 @@ func (b *bufferBuilder) unsafeAppend(data []byte) {
 	copy(b.bytes[b.length:], data)
 	b.length += len(data)
 }
+
+type multiBufferBuilder struct {
+	refCount  int64
+	blockSize int
```
bkietz, line 159: This should be settable so that users (and tests) can configure their own block size
```suggestion
	BlockSize int
```

go/arrow/internal/testing/gen/random_array_gen.go
```diff
@@ -351,6 +351,40 @@ func (r *RandomArrayGenerator) LargeString(size int64, minLength, maxLength int6
 	return bldr.NewArray()
 }
 
+func (r *RandomArrayGenerator) StringView(size int64, minLength, maxLength int64, nullProb float64) arrow.Array {
```
bkietz, line 354: specifying max buffer size might also be useful here

bkietz (changes requested):

go/arrow/datatype_viewheader_inline_tinygo.go
zeroshade: Provides the value as a `string` instead of `[]byte` for consumers without performing a copy. so rather than consumers having to work out / implement the zero-copy conversion themselves we provide it for them.

go/arrow/array/bufferbuilder.go
zeroshade: the `multiBufferBuilder` isn't exported itself, so a user wouldn't even have access to the buffer builder in order to set the blocksize. I can add a method to the `BinaryViewBuilder` that would allow setting the blocksize

go/arrow/datatype_viewheader_inline_tinygo.go
bkietz: Ah, then maybe 
```suggestion
func (sh *ViewHeader) InlineString() (data string) {
```

zeroshade: that's fair. renamed

go/arrow/type_traits_string_view.go
```diff
@@ -0,0 +1,53 @@
...
+package arrow
+
+import (
+	"reflect"
+	"unsafe"
+
+	"github.com/apache/arrow/go/v13/arrow/endian"
+)
+
+var StringHeaderTraits stringHeaderTraits
```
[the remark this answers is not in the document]
candiduslynx: Here we see usage of `ViewHeader` for `StringHeaderSizeBytes`, so I suggest renaming this to `ViewHeaderSizeBytes` & rename file as well.

go/arrow/datatype_viewheader.go
```diff
@@ -0,0 +1,138 @@
...
+func (sh *ViewHeader) Equals(buffers []*memory.Buffer, other *ViewHeader, otherBuffers []*memory.Buffer) bool {
+	if sh.sizeAndPrefixAsInt() != other.sizeAndPrefixAsInt() {
+		return false
+	}
+
+	if sh.IsInline() {
+		return sh.inlinedAsInt64() == other.inlinedAsInt64()
+	}
+
+	data := buffers[sh.BufferIndex()].Bytes()[sh.BufferOffset() : sh.BufferOffset()+sh.size]
```
candiduslynx, line 125: You could add a helper func (maybe with asserts?):
```go
func (sh *ViewHeader) variadicData(buffers []*memory.Buffer) []byte {
	offset := sh.BufferOffset()
	return buffers[sh.BufferIndex()].Bytes()[offset : offset+sh.size]
}
```
& then use it as `return bytes.Equal(sh.variadicData(buffers), other.variadicData(otherBuffers))`

go/arrow/datatype.go
```diff
@@ -272,6 +277,10 @@ func (b BufferSpec) Equals(other BufferSpec) bool {
 type DataTypeLayout struct {
 	Buffers []BufferSpec
 	HasDict bool
+	// if this is non-nil, the number of buffers expected is only
+	// lower-bounded by len(buffers). Buffers beyond this lower bound
+	// are expected to conform to this variadic spec.
```
candiduslynx, line 282:
```suggestion
	// VariadicSpec is what the buffers beyond len(Buffers) are expected to conform to.
```

go/arrow/array/string_test.go
```diff
@@ -619,3 +619,176 @@ func TestStringValueLen(t *testing.T) {
 		assert.Equal(t, len(v), slice.ValueLen(i))
 	}
 }
+func TestStringViewArray(t *testing.T) {
```
candiduslynx, line 622: Also requires `AppendValueFromString`/`ValueStr` contract test.

go/arrow/array/binarybuilder.go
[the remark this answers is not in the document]
candiduslynx: Also requires `AppendValueFromString`/`ValueStr` contract test for binary view.

go/arrow/array/binarybuilder.go
```diff
@@ -370,6 +371,334 @@ func (b *BinaryBuilder) UnmarshalJSON(data []byte) error {
...
+// of a byte slice, avoiding the extra copy that would occur if you simply
+// did []byte(v).
+//
+// This is different than AppendValueFromString which exists for the
+// Builder interface, in that this expects raw binary data which is
+// appended unmodified. AppendValueFromString expects base64 encoded binary
+// data instead.
+func (b *BinaryViewBuilder) AppendString(v string) {
+	// create a []byte without copying the bytes
+	// in go1.20 this would be unsafe.StringData
```
candiduslynx, line 487: You could make the `go1.20` & `!go1.20` files for different implementations given you already did this for view headers.

go/arrow/array/bufferbuilder.go
```diff
@@ -151,3 +153,112 @@ func (b *bufferBuilder) unsafeAppend(data []byte) {
...
+	return buf.Cap() - buf.Len()
+}
+
+func (b *multiBufferBuilder) Reset() {
+	b.currentOutBuffer = 0
+	for i, block := range b.blocks {
+		block.Release()
+		b.blocks[i] = nil
+	}
+	b.blocks = nil
```
candiduslynx, line 237: Is setting `b.blocks[i] = nil` necessary?
Could also be simplifies with `b.Finish()`:
```suggestion
	out := b.Finish()
	for i, block := range out {
		block.Release()
	}
```

go/arrow/array/bufferbuilder.go
```diff
@@ -151,3 +153,112 @@ func (b *bufferBuilder) unsafeAppend(data []byte) {
...
+		string
+		int
+	}{val, len(val)}))
+	b.UnsafeAppend(hdr, v)
+}
+
+func (b *multiBufferBuilder) Finish() (out []*memory.Buffer) {
+	b.currentOutBuffer = 0
+	out = b.blocks
+	b.blocks = nil
```
candiduslynx, line 262:
```suggestion
	out, b.blocks = b.blocks, nil
```

candiduslynx (changes requested):

go/arrow/datatype_viewheader.go
bkietz: It just seemed that since `/Inline(Bytes|String)/` is always used inside `if IsInline {}`, it might be preferable to write `if inline := sh.InlineBytes(); inline != nil {}`

go/arrow/ipc/metadata.go
```diff
@@ -1186,18 +1202,25 @@ func writeDictionaryMessage(mem memory.Allocator, id int64, isDelta bool, size,
...
 	if codec != -1 {
 		bodyCompressFB = writeBodyCompression(b, codec)
 	}
 
+	flatbuf.RecordBatchStartVariadicBufferCountsVector(b, len(variadicCounts))
+	for i := len(variadicCounts) - 1; i >= 0; i-- {
+		b.PrependInt64(variadicCounts[i])
+	}
+	vcFB := b.EndVector(len(variadicCounts))
+
```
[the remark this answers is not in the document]
bkietz: I think it probably would be preferable to leave the variadic buffer counts absent rather than empty if none are necessary

go/arrow/type_traits_string_view.go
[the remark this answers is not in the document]
zeroshade: bah i missed a spot when doing the renaming! Thanks for catching this!

go/arrow/ipc/metadata.go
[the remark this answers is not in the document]
zeroshade: good point, i'll make that update

go/arrow/ipc/file_reader.go
```diff
@@ -476,6 +487,9 @@ func (ctx *arrayLoaderContext) loadArray(dt arrow.DataType) arrow.ArrayData {
 	case *arrow.BinaryType, *arrow.StringType, *arrow.LargeStringType, *arrow.LargeBinaryType:
 		return ctx.loadBinary(dt)
 
+	case arrow.BinaryViewDataType:
```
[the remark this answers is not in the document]
zeroshade: `BinaryViewDataType` is an interface, not a concrete type. `StringViewDataType` implements the `BinaryViewDataType` interface. The concrete types are `BinaryViewType` and `StringViewType`. The types that end with `DataType` in the name are interfaces (like `FixedWidthDataType`). So we don't need a separate case for `StringViewType` because they are processed identically

go/arrow/internal/arrjson/arrjson.go
```diff
@@ -818,6 +826,7 @@ type Array struct {
 	Offset   interface{}           `json:"OFFSET,omitempty"`
 	Size     interface{}           `json:"SIZE,omitempty"`
 	Children []Array               `json:"children,omitempty"`
+	Variadic []string              `json:"VARIADIC_BUFFERS,omitempty"`
```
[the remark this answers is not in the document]
zeroshade: The integration testing JSON format has a defined format which uses caps for the array data values themselves and lower case for structural data like children, metadata, and data types.

go/arrow/datatype_viewheader.go
```diff
@@ -0,0 +1,141 @@
...
+func (sh *ViewHeader) BufferIndex() int32 {
+	return int32(endian.Native.Uint32(sh.data[StringViewPrefixLen:]))
+}
+
+func (sh *ViewHeader) BufferOffset() int32 {
+	return int32(endian.Native.Uint32(sh.data[StringViewPrefixLen+4:]))
+}
+
+func (sh *ViewHeader) InlineBytes() (data []byte) {
+	debug.Assert(sh.IsInline(), "calling InlineBytes on non-inline StringHeader")
```
candiduslynx, line 91:
```suggestion
	debug.Assert(sh.IsInline(), "calling InlineBytes on non-inline ViewHeader")
```

candiduslynx (approved):

go/arrow/datatype_viewheader.go
```diff
@@ -0,0 +1,141 @@
...
+	"unsafe"
+
+	"github.com/apache/arrow/go/v15/arrow/endian"
+	"github.com/apache/arrow/go/v15/arrow/internal/debug"
+	"github.com/apache/arrow/go/v15/arrow/memory"
+)
+
+const (
+	StringViewPrefixLen  = 4
+	stringViewInlineSize = 12
```
candiduslynx, line 30:
```suggestion
	ViewPrefixLen  = 4
	viewInlineSize = 12
```

candiduslynx (changes requested):

go/arrow/array/binary.go
```diff
@@ -318,6 +319,126 @@ func arrayEqualLargeBinary(left, right *LargeBinary) bool {
...
+}
+
+func (a *BinaryView) GetOneForMarshal(i int) interface{} {
+	if a.IsNull(i) {
+		return nil
+	}
+	return a.Value(i)
+}
+
+func (a *BinaryView) MarshalJSON() ([]byte, error) {
```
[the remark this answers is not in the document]
zeroshade: The tests in `util_test.go` for `TestArrRecordsJSONRoundTrip` cover this as we have String/BinaryView records in the test data which get run as round trip using `RecordFromJSON` and `json.Encode`

go/arrow/array/binary.go
```diff
@@ -318,6 +319,126 @@ func arrayEqualLargeBinary(left, right *LargeBinary) bool {
...
+
+// ValueStr is paired with AppendValueFromString in that it returns
+// the value at index i as a string: Semantically this means that for
+// a null value it will return the string "(null)", otherwise it will
+// return the value as a base64 encoded string suitable for CSV/JSON.
+//
+// This is always going to be less performant than just using ValueString
+// and exists to fulfill the Array interface to provide a method which
+// can produce a human readable string for a given index.
+func (a *BinaryView) ValueStr(i int) string {
```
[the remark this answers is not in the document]
zeroshade: added

candiduslynx (approved):

bkietz (approved):

PR #35780, opened 2023-05-26
GH-15060: [JS] Add LargeUtf8 type

### LINKED ISSUE #15060: [JS] Missing support for LargeUtf8 datatype

Currently there is no support for the large_string or largeutf8 datatype in the javascript libraries.

### Component(s)

JavaScript

domoritz:
This pull request adds support for the LargeUTF8 type in Arrow. Now we can create, decode, and encode these vectors. However, while the offset vectors support 64 bit integers, note that the value buffers are limited to a length of 32 bits meaning that LargeUTF8 vectors cannot yet be larger than UTF8 vectors. We will see how we can address this limitation in a follow up pull request. The issue is that JS typed arrays can be at most 2**31-1 elements long (implementation defined). This pull request also fixes a bug in a rounding method which prevented us from supporting large vectors so it's already a big step forward.

Fixes #15060.
* Closes: #15060


pitrou:
Why not add `LargeBinary` at the same time?

domoritz:
> Why not add LargeBinary at the same time?

I will look at it when I get this working. I just need some help wrapping it up.

js/src/builder.ts
```diff
@@ -289,7 +290,7 @@ export abstract class Builder<T extends DataType = any, TNull = any> {
             valueOffsets = _offsets?.flush(length);
         } else if (valueOffsets = _offsets?.flush(length)) { // Variable-width primitives (Binary, Utf8), and Lists
             // Binary, Utf8
-            data = _values?.flush(_offsets.last());
+            data = _values?.flush(Number(_offsets.last()));
```
domoritz, line 293: I think this will fail for very large offsets that are larger than `number` allows but I think that's inevitable.

js/src/enum.ts
```diff
@@ -201,6 +201,7 @@ export enum Type {
     SparseUnion = -24,
     IntervalDayTime = -25,
     IntervalYearMonth = -26,
+    LargeUtf8 = -27,
```
domoritz, line 204: Is this right?

js/src/builder.ts
domoritz: Inevitable because JS supports indexing with numbers only.

js/src/type.ts
```diff
@@ -39,9 +39,11 @@ export type IsSigned = { 'true': true; 'false': false };
 export interface DataType<TType extends Type = Type, TChildren extends TypeMap = any> {
     readonly TType: TType;
     readonly TArray: any;
+    readonly TOffset: any;
     readonly TValue: any;
     readonly TChildren: TChildren;
     readonly ArrayType: any;
+    readonly OffsetType: TypedArrayConstructor<Int32Array> | BigIntArrayConstructor<BigInt64Array>;
```
domoritz, line 46: Maybe this should be `OffsetArrayType`?

js/src/enum.ts
kylebarron: >  * **Note**: Only enum values 0-17 (NONE through Map) are written to an Arrow IPC payload.

Seems like `LargeUtf8` would be given a positive number then

FrNecas: I think it should be 20, according to the flatbuffers file that you get when you build the package (also the comment about 0-17) is probably outdated.

FrNecas (changes requested):
While working on adding `Duration` type, I've noticed that you are missing a few pieces to make a new type work properly:
 - add handling for the type to `src/visitor/bytelength.ts`
 - add handling for the type to `src/visitor/vectorloader.ts`
 - add a case for the type to `src/ipc/metadata/message.ts` so that `tableFromIPC` works. This is the relevant code: https://github.com/apache/arrow/blob/main/js/src/ipc/metadata/message.ts#L426

FrNecas:
One more place where this needs to be added: https://github.com/apache/arrow/blob/main/js/src/ipc/metadata/json.ts#L159

pitrou:
Also, here as well, need to remove the corresponding skips in the integration tests:
https://github.com/apache/arrow/blob/b9453a2a33e935e26dc8358846ad671aefdcb4d0/dev/archery/archery/integration/datagen.py#L1668-L1670

ArcRiiad:
Some work left on this PR? :p

domoritz:
Yes. We need to have a buffer that supports lengths of more than what js supports natively.

geclos:
Hi,

Very interested in this PR moving forward. What is the status? Any way I could be of help?

domoritz:
I left it unfinished before I went on vacation and just got back. What's left is addressing the comments here and also implementing a buffer that supports chunks so it can be larger than a normal js array. pRs against this branch would be awesome.

domoritz:
Looks like it's pretty easy to exceed the buffer sizes, though.

```ts
import { constants } from 'node:buffer';
import { LargeUtf8, makeBuilder } from './src/Arrow.dom.ts';

// make the longest posstible string (longer will crash with `RangeError: Invalid string length`)
const MAX_STRING_LENGTH = constants.MAX_STRING_LENGTH;
console.log(`MAX_STRING_LENGTH: ${MAX_STRING_LENGTH}`);
const longString = "a".repeat(MAX_STRING_LENGTH);

const builder = makeBuilder({ type: new LargeUtf8 });

// length of vector
const N = 5;

console.log(`Building vector with total string length (potential need for offsets): ${MAX_STRING_LENGTH * N} or 2^${Math.ceil(Math.log2(MAX_STRING_LENGTH * N))}`);

for (let i = 0; i < N; i++) {
    builder.append(longString);
}
// add string to force another offset
builder.append("");
builder.finish();
const vector = builder.toVector();

console.log(vector);
```

js/src/enum.ts
domoritz: Fixed it but we should probably do https://github.com/apache/arrow/issues/39048.

js/src/type.ts
domoritz: Changed

kylebarron:
I do feel that supporting the full spec (i.e. _having_ a `LargeUtf8` type) and supporting massive strings larger than `Number.MAX_SAFE_INTEGER` are two separate objectives.

I personally feel that it's fine to use `number` for indexing (with some safety checks to avoid overflows) and to instruct people to use Vectors if they need an offset value larger than `Number.MAX_SAFE_INTEGER`. I think "a new data structure that chunks" is a lot of added complexity when `Vector` is _already_ a data structure that chunks.

js/src/enum.ts
```diff
@@ -137,8 +137,7 @@ export enum MessageHeader {
  * nested type consisting of other data types, or another data type (e.g. a
  * timestamp encoded as an int64).
  *
- * **Note**: Only enum values 0-18 (NONE through Duration) are written to an Arrow
- * IPC payload.
+ * **Note**: Only positive enum values are written to an Arrow IPC payload.
```
kylebarron, line 140:
```suggestion
 * **Note**: Only non-negative enum values are written to an Arrow IPC payload.
```
Maybe this is pedantic; does `0` get written to IPC as the original comment suggested?

domoritz:
Sounds good. Thanks for the input. 

Vectors are chunked but buffers in builders are not. So I think I want to add a chunked buffer or figure out how to flush to vectors when the buffer is full.

domoritz:
Is it actually an issue that chunks have different lengths? Seems to work fine here

```js
const b1 = makeBuilder({ type: new Int32 });
const b2 = makeBuilder({ type: new Int32 });

const d1 = [] as Data[];
const d2 = [] as Data[];

b1.append(1);
b1.append(2);
d1.push(b1.flush());
b1.append(3);
d1.push(b1.flush());

b2.append(4);
d2.push(b2.flush());
b2.append(5);
b2.append(6);
d2.push(b2.flush());

const v1 = new Vector(d1);
const v2 = new Vector(d2);

const table = new Table({ v1, v2 });
console.log(table.toArray());
```

kylebarron:
But then you call

```ts
table.batches[0].numRows
// 1
```

Implicitly, a batch has a constant number of rows across all `Data` instances.

When I test it in Python, surprisingly the `Table` constructor _allows_ it, but then when you call `to_batches` it reaggregates chunks into consistent record batches:

```py
import pyarrow as pa

d1 = [
    pa.array([1, 2], pa.int32()),
    pa.array([3], pa.int32()),
]
d2 = [
    pa.array([4], pa.int32()),
    pa.array([5, 6], pa.int32()),
]

c1 = pa.chunked_array(d1)
c2 = pa.chunked_array(d2)
table = pa.table({"v1": c1, "v2": c2})
table
# pyarrow.Table
# v1: int32
# v2: int32
# ----
# v1: [[1,2],[3]]
# v2: [[4],[5,6]]

# to_batches regroups into batches of length 1
table.to_batches()
# [pyarrow.RecordBatch
#  v1: int32
#  v2: int32
#  ----
#  v1: [1]
#  v2: [4],
#  pyarrow.RecordBatch
#  v1: int32
#  v2: int32
#  ----
#  v1: [2]
#  v2: [5],
#  pyarrow.RecordBatch
#  v1: int32
#  v2: int32
#  ----
#  v1: [3]
#  v2: [6]]
```

domoritz:
Hmm, then we can't really automatically flush in the vector builder. That's a bummer. Maybe then I need to implement a chunked buffer that can satisfy the `DataBuffer` type. I need to look at what implications that may have. 

Maybe @trxcllnt has some suggestion here.

domoritz:
Is that a problem? Looking at the C++ docs at https://arrow.apache.org/docs/cpp/api/array.html#classarrow_1_1_chunked_array it seems like it should be okay if chunk lengths are not consistent. 

Screenshot 2023-12-05 at 15 08 28

Using a vector with chunked data would be a lot cleaner than e.g. implementing a chunked buffer (then we would have a chunked data buffer inside a vector that could also be chunked).

kylebarron:
> Is that a problem? Looking at the C++ docs at https://arrow.apache.org/docs/cpp/api/array.html#classarrow_1_1_chunked_array it seems like it should be okay if chunk lengths are not consistent.

It's not the ChunkedArray you want to look at. That indeed allows each `Data` instance within a _single_ `Vector` to have different lengths. But if you look at the [doc for `RecordBatch`](https://arrow.apache.org/docs/cpp/api/table.html#_CPPv4N5arrow11RecordBatchE) it indeed says

> Collection of **equal-length arrays** matching a particular [Schema](https://arrow.apache.org/docs/cpp/api/datatype.html#classarrow_1_1_schema)

(emphasis mine). So two columns can't have different lengths within a single `Batch`. 

Maybe you're right that this is effectively the same as what pyarrow is doing. It's hard to tell if those batches are views on the existing buffers, like the slicing Arrow JS's `batches` have on the existing buffers.

pitrou:
> Maybe you're right that this is effectively the same as what pyarrow is doing. It's hard to tell if those batches are views on the existing buffers

What do "those batches" denote here? I might be able to answer if you phrase the question more precisely :-)

domoritz:
It looks like what happens is that when you put differently chunked vectors into a table, the record batch boundaries are created at the chunk gaps from all vectors. So if you have two vectors with chunk lengths of 2,3 and 3,2, you will get record batches of length 2,1,2. I would hope that the data buffers are views on the existing buffers. The only issue I can see with this approach is that in the case of variable width vectors (like utf8)---where we have a data buffer where we might reference the same position multiple times from the offset buffer---we would need to copy the data buffer in serialized record batches.

pitrou:
> The only issue I can see with this approach is that in the case of variable width vectors (like utf8)---where we have a data buffer where we might reference the same position multiple times from the offset buffer

This is impossible given the [layout](https://arrow.apache.org/docs/format/Columnar.html#variable-size-binary-layout) of the binary types.
It is possible for the [binary view](https://arrow.apache.org/docs/format/Columnar.html#variable-size-binary-view-layout) types, though.

js/src/enum.ts
```diff
@@ -175,6 +174,7 @@ export enum Type {
     FixedSizeList = 16, /** Fixed-size list. Each value occupies the same number of bytes */
     Map = 17, /** Map of named logical types */
     Duration = 18, /** Measure of elapsed time in either seconds, milliseconds, microseconds or nanoseconds. */
+    LargeUtf8 = 20, /** Large variable-length string as List<Char> */
```
kylebarron, line 177: Did you skip 19 on purpose?

js/src/util/buffer.ts
```diff
@@ -208,10 +208,12 @@ export async function* toArrayBufferViewAsyncIterator<T extends TypedArray>(Arra
...
 
 /** @ignore */
-export function rebaseValueOffsets(offset: number, length: number, valueOffsets: Int32Array) {
+export function rebaseValueOffsets(offset: number, length: number, valueOffsets: Int32Array): Int32Array;
+export function rebaseValueOffsets(offset: bigint, length: number, valueOffsets: BigInt64Array): BigInt64Array;
+export function rebaseValueOffsets(offset: number | bigint, length: number, valueOffsets: any) {
     // If we have a non-zero offset, create a new offsets array with the values
     // shifted by the start offset, such that the new start offset is 0
-    if (offset !== 0) {
+    if (offset != 0) {
```
kylebarron, line 216: is this necessary to compare the bigint to a number?

kylebarron (approved):
lgtm!

js/src/enum.ts
domoritz: We haven't implemented it so I'm skipping it for now.

geclos:
Thanks for your work guys!

PR #35782, opened 2023-05-26
GH-X: [JS] Remove WideValue

domoritz:
I think we can use bigint now.


domoritz:
We actually need this because decimals are stored as typed arrays.

PR #12460, opened 2022-02-18
ARROW-13530: [C++] Implement cumulative sum compute function

JabariBooker:
Creating new compute function to perform a cumulative sum on a given array/vector.

PR #35787, opened 2023-05-26
GH-35786: [C++] Add pairwise_diff function

### LINKED ISSUE #35786: [C++] Add pairwise_diff function

Similar to pandas' [Series.Diff](https://pandas.pydata.org/docs/reference/api/pandas.Series.diff.html), the `pairwise_diff` function computes the first order difference of an array, i.e. output[i] = input[i] - input[i-period] where period is configurable.

### Component(s)

C++

js8544:
### Rationale for this change

Add a `pairwise_diff` function similar to pandas' [Series.Diff](https://pandas.pydata.org/docs/reference/api/pandas.Series.diff.html), the function computes the first order difference of an array.

I followed [these instructions](https://github.com/apache/arrow/pull/12460#issuecomment-1057520554). The function is implemented for numerical, temporal and decimal types. Chuck arrays are not yet supported.

### Are these changes tested?

Yes. They are tested in vector_pairwise_test.cc and in python/pyarrow/tests/compute.py.

Yes, and docs are also updated in this PR.

* Closes: #35786

### DIFF
--- a/cpp/src/arrow/compute/api_vector.h
+++ b/cpp/src/arrow/compute/api_vector.h
@@ -670,6 +670,7 @@ Result<Datum> CumulativeMin(
 ///
 /// \param[in] array array input
 /// \param[in] options options, specifying overflow behavior and period
+/// \param[in] check_overflow whether to return error on overflow
 /// \param[in] ctx the function execution context, optional
 /// \return result as array
 ARROW_EXPORT
--- a/cpp/src/arrow/compute/api_vector.h
+++ b/cpp/src/arrow/compute/api_vector.h
@@ -670,7 +670,8 @@ Result<Datum> CumulativeMin(
 ///   Diff([1, 4, 9, 10, 15]) = [null, 3, 5, 1, 5].
 /// With p = 2,
 ///   Diff([1, 4, 9, 10, 15]) = [null, null, 8, 6, 6]
-///
+/// p can also be negative, in which case the diff is computed in
+/// the opposite direction.
 /// \param[in] array array input
 /// \param[in] options options, specifying overflow behavior and period
 /// \param[in] check_overflow whether to return error on overflow
--- a/cpp/src/arrow/compute/api_vector.h
+++ b/cpp/src/arrow/compute/api_vector.h
@@ -234,6 +234,21 @@ class ARROW_EXPORT CumulativeOptions : public FunctionOptions {
 };
 using CumulativeSumOptions = CumulativeOptions;  // For backward compatibility
 
+/// \brief Options for pairwise functions
+class ARROW_EXPORT PairwiseDiffOptions : public FunctionOptions {
+ public:
+  explicit PairwiseDiffOptions(double periods = 1, bool check_overflow = false);
+  static constexpr char const kTypeName[] = "PairwiseDiffOptions";
+  static PairwiseDiffOptions Defaults() { return PairwiseDiffOptions(); }
+
+  /// Periods to shift for applying the binary operation, accepts negative values.
+  int64_t periods = 1;
+
+  /// When true, returns an Invalid Status when overflow is detected, otherwise result is
+  /// wrapped around.
+  bool check_overflow = false;
+};
+
 /// @}
 
 /// \brief Filter with a boolean selection filter
@@ -649,6 +664,22 @@ ARROW_EXPORT
 Result<Datum> CumulativeMin(
     const Datum& values, const CumulativeOptions& options = CumulativeOptions::Defaults(),
     ExecContext* ctx = NULLPTR);
+
+/// \brief Return the first order difference of an array.
+///
+/// Computes the first order difference of an array, i.e. output[i] = input[i] - input[i -
+/// p] if i >= p, otherwise output[i] = null, where p is the period. For example, with p =
+/// 1, Diff([1, 4, 9, 10, 15]) = [null, 3, 5, 1, 5]. With p = 2, Diff([1, 4, 9, 10, 15]) =
+/// [null, null, 8, 6, 6]
+///
+/// \param[in] datum array input
+/// \param[in] options options, specifying overflow behavior and period
+/// \param[in] ctx the function execution context, optional
+/// \return result as array
+ARROW_EXPORT
+Result<std::shared_ptr<Array>> PairwiseDiff(const Array& array,
+                                            const PairwiseDiffOptions& options,
+                                            ExecContext* ctx = NULLPTR);
 
 // ----------------------------------------------------------------------
 // Deprecated functions
--- a/cpp/src/arrow/compute/api_vector.h
+++ b/cpp/src/arrow/compute/api_vector.h
@@ -237,7 +237,7 @@ using CumulativeSumOptions = CumulativeOptions;  // For backward compatibility
 /// \brief Options for pairwise functions
 class ARROW_EXPORT PairwiseDiffOptions : public FunctionOptions {
  public:
-  explicit PairwiseDiffOptions(double periods = 1, bool check_overflow = false);
+  explicit PairwiseDiffOptions(int64_t periods = 1, bool check_overflow = false);
   static constexpr char const kTypeName[] = "PairwiseDiffOptions";
   static PairwiseDiffOptions Defaults() { return PairwiseDiffOptions(); }
 
--- a/cpp/src/arrow/compute/api_vector.h
+++ b/cpp/src/arrow/compute/api_vector.h
@@ -672,7 +672,7 @@ Result<Datum> CumulativeMin(
 /// 1, Diff([1, 4, 9, 10, 15]) = [null, 3, 5, 1, 5]. With p = 2, Diff([1, 4, 9, 10, 15]) =
 /// [null, null, 8, 6, 6]
 ///
-/// \param[in] datum array input
+/// \param[in] array array input
 /// \param[in] options options, specifying overflow behavior and period
 /// \param[in] ctx the function execution context, optional
 /// \return result as array
--- a/cpp/src/arrow/compute/api_vector.h
+++ b/cpp/src/arrow/compute/api_vector.h
@@ -235,18 +235,14 @@ class ARROW_EXPORT CumulativeOptions : public FunctionOptions {
 using CumulativeSumOptions = CumulativeOptions;  // For backward compatibility
 
 /// \brief Options for pairwise functions
-class ARROW_EXPORT PairwiseDiffOptions : public FunctionOptions {
+class ARROW_EXPORT PairwiseOptions : public FunctionOptions {
  public:
-  explicit PairwiseDiffOptions(int64_t periods = 1, bool check_overflow = false);
-  static constexpr char const kTypeName[] = "PairwiseDiffOptions";
-  static PairwiseDiffOptions Defaults() { return PairwiseDiffOptions(); }
+  explicit PairwiseOptions(int64_t periods = 1);
+  static constexpr char const kTypeName[] = "PairwiseOptions";
+  static PairwiseOptions Defaults() { return PairwiseOptions(); }
 
   /// Periods to shift for applying the binary operation, accepts negative values.
   int64_t periods = 1;
-
-  /// When true, returns an Invalid Status when overflow is detected, otherwise result is
-  /// wrapped around.
-  bool check_overflow = false;
 };
 
 /// @}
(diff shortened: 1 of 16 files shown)


js8544:
@lidavidm @pitrou @westonpace Hi guys, would any of you mind taking a look at this PR? It's been sitting around for almost a month. Thanks in advance!

lidavidm (approved):
Seems reasonable to me

js8544:
@pitrou @bkietz Would you mind having some extra look at this PR? Thanks!

cpp/src/arrow/compute/api_vector.h
```diff
@@ -603,6 +614,24 @@ Result<Datum> CumulativeSum(
     const CumulativeSumOptions& options = CumulativeSumOptions::Defaults(),
     ExecContext* ctx = NULLPTR);
 
+/// \brief Return the first order difference of an array.
+///
+/// Computes the first order difference of an array, i.e. output[i] = input[i] - input[i -
+/// p] if i >= p, otherwise output[i] = null, where p is the period. For example, with p =
+/// 1, Diff([1, 4, 9, 10, 15]) = [null, 3, 5, 1, 5]. With p = 2, Diff([1, 4, 9, 10, 15]) =
+/// [null, null, 8, 6, 6]
```
bkietz, line 622: Looks like you had formatting here which got munged, guessing:
```suggestion
/// Computes the first order difference of an array, i.e.
///   output[i] = input[i] - input[i - p]  if i >= p
///   output[i] = null                     otherwise
/// where p is the period. For example, with p = 1,
///   Diff([1, 4, 9, 10, 15]) = [null, 3, 5, 1, 5].
/// With p = 2,
///   Diff([1, 4, 9, 10, 15]) = [null, null, 8, 6, 6]
```

cpp/src/arrow/compute/kernels/vector_pairwise.cc
```diff
@@ -0,0 +1,272 @@
...
+    VisitTwoArrayValuesInline<InputType, InputType>(left, right, valid_func, null_func);
+    RETURN_NOT_OK(status);
+    RETURN_NOT_OK(builder.AppendNulls(-periods));
+  }
+  RETURN_NOT_OK(builder.FinishInternal(result));
+  return Status::OK();
+}
+
+template <typename InputType, typename OutputType, typename Op>
+Status PairwiseDiffKernel(KernelContext* ctx, const ExecSpan& batch, ExecResult* out) {
```
bkietz, line 112: Nit:
```suggestion
Status PairwiseDiffExec(KernelContext* ctx, const ExecSpan& batch, ExecResult* out) {
```

docs/source/cpp/compute.rst
```diff
@@ -1835,3 +1835,25 @@ replaced, based on the remaining inputs.
   results in a corresponding null in the output.
 
   Also see: :ref:`if_else <cpp-compute-scalar-selections>`.
+
+Pairwise functions
+~~~~~~~~~~~~~~~~~~~~
+Pairwise functions are unary vector functions that performs a binary operation on
```
bkietz, line 1841:
```suggestion
Pairwise functions are unary vector functions that perform a binary operation on 
```

bkietz (changes requested):
This mostly looks good to me. However, in light of the strong relationship between this Function and subtract I'd prefer to see more reuse of that function's existing logic. For example when retrieving the output type for pairwise_diff($in_type), couldn't we just as easily call subtract's DispatchExact with ($in_type, $in_type)? And instead of referencing Subtract's Ops directly, couldn't we use the kernel retrieved from subtract's DispatchExact- just passing the relevant slices of the input as the arguments to the subtract kernel? I think this would have negligible impact on performance and would greatly reduce the future maintenance burden for this function, since any new types added to subtract will then automatically be supported by pairwise_diff.

js8544:
> This mostly looks good to me. However, in light of the strong relationship between this Function and subtract I'd prefer to see more reuse of that function's existing logic. For example when retrieving the output type for pairwise_diff($in_type), couldn't we just as easily call subtract's DispatchExact with ($in_type, $in_type)? And instead of referencing Subtract's Ops directly, couldn't we use the kernel retrieved from subtract's DispatchExact- just passing the relevant slices of the input as the arguments to the subtract kernel? I think this would have negligible impact on performance and would greatly reduce the future maintenance burden for this function, since any new types added to subtract will then automatically be supported by pairwise_diff.

Hi @bkietz, do you mean something like this:
```cpp
  auto subtract_func = registry->GetFunction("subtract").ValueOrDie();
  for (const auto& type : types) {
    auto kernel = subtract_func->DispatchExact({type, type}).ValueOrDie();
    // reuse kernel's exec and signature
  }
```

bkietz:
@js8544 yes, exactly. I think we could go a step further and loop over subtracts kernels directly, without the need to go through dispatchExact

js8544:
> I think we might be able to go a step further and loop over subtract's kernels directly, without the need to list input types explicitly and go through DispatchExact.

That sounds better indeed! I'll try this now.

js8544:
@bkietz I've changed it to looping over subtract's kernels and wrapping their signature and kernel exec.

cpp/src/arrow/compute/kernel.h
```diff
@@ -293,6 +298,10 @@ class ARROW_EXPORT OutputType {
   OutputType(Resolver resolver)  // NOLINT implicit construction
       : kind_(COMPUTED), resolver_(std::move(resolver)) {}
 
+  /// \brief For backward compatibility
+  OutputType(ResolverFuncPtr resolver)  // NOLINT implicit construction
+      : kind_(COMPUTED), resolver_(std::move(resolver)) {}
```
js8544, line 303: Since the constructors are not marked explicit, many existing codes implicitly pass a function pointer as `OutputType`. To make them compile, I'll have to either add this constructor or change all occurences. I chose the easier path.

js8544: Nevermind, some compilation fails due to ambiguous call. I guess I'll have to change all occurences.

js8544: Actually making this a templated method is more elegant. Problem solved.

cpp/src/arrow/compute/kernels/vector_pairwise.cc
```diff
@@ -0,0 +1,182 @@
...
+struct PairwiseKernelData {
+  InputType input;
+  OutputType output;
+  ArrayKernelExec exec;
+};
+
+void RegisterPairwiseDiffKernels(std::string_view func_name,
+                                 std::string_view base_func_name, const FunctionDoc& doc,
+                                 FunctionRegistry* registry) {
+  VectorKernel base_kernel;
```
bkietz, line 138: Nit: since the base function here is "subtract", this naming could cause readers to conclude that the kernel is one of "subtract"'s kernels.
```suggestion
  VectorKernel kernel;
```

cpp/src/arrow/compute/kernels/vector_pairwise.cc
```diff
@@ -0,0 +1,182 @@
...
+                                               doc, GetDefaultPairwiseOptions());
+
+  auto base_func_result = registry->GetFunction(std::string(base_func_name));
+  DCHECK_OK(base_func_result.status());
+  const auto& base_func = checked_cast<const ScalarFunction&>(**base_func_result);
+  DCHECK_EQ(base_func.arity().num_args, 2);
+
+  for (const auto& base_func_kernel : base_func.kernels()) {
+    const auto& base_func_kernel_sig = base_func_kernel->signature;
+    if (base_func_kernel_sig->in_types()[0].Equals(base_func_kernel_sig->in_types()[1])) {
```
bkietz, line 153: Nit: de-nest
```suggestion
    if (!base_func_kernel_sig->in_types()[0].Equals(base_func_kernel_sig->in_types()[1])) {
      continue;
    }
```

cpp/src/arrow/compute/kernels/vector_pairwise.cc
```diff
@@ -0,0 +1,182 @@
...
+      OutputType out_type(base_func_kernel_sig->out_type());
+      // Need to wrap base output resolver
+      if (out_type.kind() == OutputType::COMPUTED) {
+        const auto& base_resolver = base_func_kernel_sig->out_type().resolver();
+        auto resolver = [&base_resolver](KernelContext* ctx,
+                                         const std::vector<TypeHolder>& input_types) {
+          return base_resolver(ctx, {input_types[0], input_types[0]});
+        };
+        out_type = OutputType(resolver);
```
bkietz, line 162: Again, I think it's better to err on the side of safety when we can and just copy:
```suggestion
        out_type = OutputType([base_resolver=base_func_kernel_sig->out_type().resolver()
        ](KernelContext* ctx,
                                         const std::vector<TypeHolder>& input_types) {
          return base_resolver(ctx, {input_types[0], input_types[0]});
        });
```

docs/source/cpp/compute.rst
```diff
@@ -1847,3 +1847,25 @@ replaced, based on the remaining inputs.
   results in a corresponding null in the output.
 
   Also see: :ref:`if_else <cpp-compute-scalar-selections>`.
+
+Pairwise functions
+~~~~~~~~~~~~~~~~~~~~
+Pairwise functions are unary vector functions that perform a binary operation on 
+a pair of elements in the input array, typically on adjacent elements. The n-th
+output is computed by applying the binary operation on the n-th and (n-p)-th, 
+where p is the period. The default period is 1. The period can also be negative.
```
bkietz, line 1856:
```suggestion
output is computed by applying the binary operation to the n-th and (n-p)-th inputs, 
where p is the period. The default period is 1, in which case the binary
operation is applied to adjacent pairs of inputs. The period can also be
negative, in which case the n-th output is computed by applying the binary
operation to the n-th and (n+abs(p))-th inputs.
```

bkietz (changes requested):
This is looking great, thanks for refactoring!
Just a few nits

cpp/src/arrow/compute/kernels/vector_pairwise.cc
```diff
@@ -0,0 +1,182 @@
...
+struct PairwiseState : KernelState {
+  PairwiseState(const PairwiseOptions& options, const ArrayKernelExec& scalar_exec)
+      : periods(options.periods), scalar_exec(scalar_exec) {}
+
+  int64_t periods;
+  const ArrayKernelExec& scalar_exec;
+};
+
+KernelInit GeneratePairwiseInit(const ArrayKernelExec& scalar_exec) {
+  return [&scalar_exec](KernelContext* ctx, const KernelInitArgs& args) {
```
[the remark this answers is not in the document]
js8544: Updated to inline lambda and capturing by value.

cpp/src/arrow/compute/kernels/vector_pairwise.cc
```diff
@@ -0,0 +1,182 @@
...
+/// A generic pairwise implementation that can be reused by different ops.
+Status PairwiseExecImpl(KernelContext* ctx, const ArraySpan& input,
+                        const ArrayKernelExec& scalar_exec, int64_t periods,
+                        ArrayData* result) {
+  auto offset = abs(periods);
+  offset = std::min(offset, input.length);
+  auto exec_length = input.length - offset;
+  // prepare bitmap
+  auto null_start = periods > 0 ? 0 : exec_length;
+  auto non_null_start = periods > 0 ? offset : 0;
```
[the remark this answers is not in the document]
js8544: Updated. It's much clearer now. thanks for the suggestion!

cpp/src/arrow/compute/kernels/vector_pairwise.cc
js8544: Updated. Sorry for the confusion!

cpp/src/arrow/compute/kernels/vector_pairwise.cc
js8544: Updated

cpp/src/arrow/compute/kernels/vector_pairwise.cc
js8544: Done

bkietz (approved):
LGTM, thanks!

bkietz:
macOs glib&python failures are network failures while communicating with brew https://github.com/apache/arrow/actions/runs/5408688637/jobs/9835499198?pr=35787

macOS c++ failure is a known issue https://github.com/apache/arrow/issues/36329

ubuntu c++ failure is an s3 flake https://github.com/apache/arrow/actions/runs/5408688635/jobs/9828017525?pr=35787#step:7:4892

R test failure is a known issue https://github.com/apache/arrow/issues/36346

PR #35790, opened 2023-05-26
GH-35789: [C++] Remove check_overflow from CumulativeSumOptions

### LINKED ISSUE #35789: [C++] cumulative_sum ignores CumulativeSumOptions::check_overflow

There are two variants of cumsum, `cumulative_sum` always wraps around on overflow and `cumulative_sum` always return `Status::Invalid` on overflow, regardless of the `check_overflow` parameter in CumulativeSumOptions.
For example:
```cpp
CumulativeSumOptions options(/*start=*/0, /*skip_nulls=*/true, /*check_overflow=*/true);
auto res = CallFunction("cumulative_sum", {ArrayFromJSON(int8(), "[127, 1]")}, &options);
std::cout << res->make_array()->ToString() << std::endl;
// prints [127, -128], but user might expect an error returned.
```

The document of `cumulative_sum` also doesn't memtion `check_overflow` at all, and `check_overflow` is not exposed to  pyarrow. So I believe the best approach to avoid confusion is to remove `check_overflow` from CumulativeSumOptions.

The only place where `check_overflow` is used is in the convenience function `CumulativeSum` defined in api_vector.h, which calls `cumulative_sum_checked` or `cumulative_sum` depending on the parameter. It can be replaced by a bool parameter in the convenience function.

### Component(s)

C++

js8544:
### Rationale for this change

There are two variants of cumsum, `cumulative_sum` always wraps around on overflow and `cumulative_sum_checked` always return `Status::Invalid` on overflow, regardless of the `check_overflow` parameter in CumulativeSumOptions.
For example:
```cpp
CumulativeSumOptions options(/*start=*/0, /*skip_nulls=*/true, /*check_overflow=*/true);
auto res = CallFunction("cumulative_sum", {ArrayFromJSON(int8(), "[127, 1]")}, &options);
std::cout << res->make_array()->ToString() << std::endl;
// prints [127, -128], but user might expect an error returned.
```

I believe it's more of a ambiguity rather than a bug. The document of `cumulative_sum` also doesn't mention `check_overflow` at all, and `check_overflow` is not exposed to pyarrow. So IMO the best approach to avoid confusion is to remove `check_overflow` from CumulativeSumOptions.

The only place where `check_overflow` is used is in the C++ convenience function `CumulativeSum` defined in api_vector.h, which calls `cumulative_sum_checked` or `cumulative_sum` depending on the parameter. It can be replaced by a bool parameter in the C++ convenience function.

1. `check_overflow` is removed from CumulativeSumOptions.
2. Add a bool check_overflow parameter to the C++ convenience function CumulativeSum.

### Are these changes tested?

It doesn't affect the current tests.

Yes, but only the C++ convenience function because check_overflow is not used in the kernel implementation and not exposed to pyarrow anyways.

* Closes: #35789

### DIFF
--- a/cpp/src/arrow/compute/api_vector.h
+++ b/cpp/src/arrow/compute/api_vector.h
@@ -213,10 +213,8 @@ class ARROW_EXPORT PartitionNthOptions : public FunctionOptions {
 /// \brief Options for cumulative sum function
 class ARROW_EXPORT CumulativeSumOptions : public FunctionOptions {
  public:
-  explicit CumulativeSumOptions(double start = 0, bool skip_nulls = false,
-                                bool check_overflow = false);
-  explicit CumulativeSumOptions(std::shared_ptr<Scalar> start, bool skip_nulls = false,
-                                bool check_overflow = false);
+  explicit CumulativeSumOptions(double start = 0, bool skip_nulls = false);
+  explicit CumulativeSumOptions(std::shared_ptr<Scalar> start, bool skip_nulls = false);
   static constexpr char const kTypeName[] = "CumulativeSumOptions";
   static CumulativeSumOptions Defaults() { return CumulativeSumOptions(); }
 
@@ -226,9 +224,6 @@ class ARROW_EXPORT CumulativeSumOptions : public FunctionOptions {
   /// If true, nulls in the input are ignored and produce a corresponding null output.
   /// When false, the first null encountered is propagated through the remaining output.
   bool skip_nulls = false;
-
-  /// When true, returns an Invalid Status when overflow is detected
-  bool check_overflow = false;
 };
 
 /// @}
@@ -601,7 +596,7 @@ ARROW_EXPORT
 Result<Datum> CumulativeSum(
     const Datum& values,
     const CumulativeSumOptions& options = CumulativeSumOptions::Defaults(),
-    ExecContext* ctx = NULLPTR);
+    bool check_overflow = false, ExecContext* ctx = NULLPTR);
 
 // ----------------------------------------------------------------------
 // Deprecated functions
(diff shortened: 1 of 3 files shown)


pitrou:
> There are two variants of cumsum, `cumulative_sum` always wraps around on overflow and `cumulative_sum` always return `Status::Invalid` on overflow,

This does not seem to make sense to me, is there a typo?

pitrou:
Regardless, the unused option is misleading, so I agree with removing it.

js8544:
> > There are two variants of cumsum, `cumulative_sum` always wraps around on overflow and `cumulative_sum` always return `Status::Invalid` on overflow,
> 
> This does not seem to make sense to me, is there a typo?

Sorry! It's a typo indeed. I meant `cumulative_sum_checked`.

cpp/src/arrow/compute/api_vector.h
```diff
@@ -601,7 +596,7 @@ ARROW_EXPORT
 Result<Datum> CumulativeSum(
     const Datum& values,
     const CumulativeSumOptions& options = CumulativeSumOptions::Defaults(),
-    ExecContext* ctx = NULLPTR);
+    bool check_overflow = false, ExecContext* ctx = NULLPTR);
```
pitrou, line 599: Can you document the  additional param in the docstring above?
Also, can you add a small test that the param is taken into account?

pitrou (changes requested):

cpp/src/arrow/compute/api_vector.h
js8544: There were no doc for CumulativeSum above... But I added one anyways.
Also added a small test.

pitrou (approved):
Thanks for this @js8544 . Let's just wait for CI now.

js8544:
The CI failure looks weird. I pushed an empty commit to trigger a rerun.

pitrou:
The CI failure was probably unrelated, but let's wait for it.

PR #35792, opened 2023-05-26
GH-20047: [MATLAB] Enable GitHub Actions CI for MATLAB Interface on Windows

### LINKED ISSUE #20047: [MATLAB] Enable GitHub Actions CI for MATLAB Interface on Windows
Enable CI and Testing for MATLAB submissions to the [Apache Arrow project](https://github.com/apache/arrow) on GitHub. This task can be worked on after [MATLAB Actions](https://github.com/matlab-actions/setup-matlab) is enabled on Windows.

**Reporter**: [Fiona La](https://issues.apache.org/jira/browse/ARROW-15196) / @lafiona

<sub>**Note**: *This issue was originally created as [ARROW-15196](https://issues.apache.org/jira/browse/ARROW-15196). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

leihou6116:
### Rationale for this change
Currently GitHub Actions builds arrow/matlab on Linux and Mac but not on Windows. This change is to enable the build on Windows such that we have qualification coverage on all popular platforms.

The build on windows uses Ninja and Visual Studio to build. This submission is aimed to enable the CI on Windows. In the future, we'll try to build using Ninja and MinGW.

### Are these changes tested?
Yes. When I submitted the change, I verified that the Windows CI was enabled that tests were run successfully.

No.

### Future Direction
On Windows, use Ninja+MinGW to build to remove the dependency on Visual Studio.

* Closes: #20047

### DIFF
--- a/.github/workflows/matlab.yml
+++ b/.github/workflows/matlab.yml
@@ -103,6 +103,12 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
+      - name: Install ninja-build
+        run: brew install ninja
+      - uses: msys2/setup-msys2@v2
+      - name: Setup MSYS2
+        shell: msys2 {0}
+        run: bash ci/scripts/msys2_setup.sh "matlab"
       - name: Install MATLAB
         uses: matlab-actions/setup-matlab@v1
       - name: Build MATLAB Interface
--- a/.github/workflows/matlab.yml
+++ b/.github/workflows/matlab.yml
@@ -111,7 +111,7 @@ jobs:
         uses: matlab-actions/setup-matlab@v1
       - name: Build MATLAB Interface
         run: bash ci/scripts/matlab_build.sh $(pwd)
-      - name: Run MATLAB Testst
+      - name: Run MATLAB Tests
         env:
           # Add the installation directory to the MATLAB Search Path by
           # setting the MATLABPATH environment variable.
--- a/.github/workflows/matlab.yml
+++ b/.github/workflows/matlab.yml
@@ -105,6 +105,8 @@ jobs:
           fetch-depth: 0
       - name: Install MATLAB
         uses: matlab-actions/setup-matlab@v1
+      - name: Enable Developer Command Prompt for Microsoft Visual C++
+        uses: ilammy/msvc-dev-cmd@v1
       - name: Build MATLAB Interface
         run: bash ci/scripts/matlab_build.sh $(pwd)
       - name: Run MATLAB Tests
--- a/.github/workflows/matlab.yml
+++ b/.github/workflows/matlab.yml
@@ -104,7 +104,7 @@ jobs:
         with:
           fetch-depth: 0
       - name: Install ninja-build
-        run: brew install ninja
+        run: choco install ninja
       - uses: msys2/setup-msys2@v2
       - name: Setup MSYS2
         shell: msys2 {0}
--- a/.github/workflows/matlab.yml
+++ b/.github/workflows/matlab.yml
@@ -103,10 +103,6 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
-      - uses: msys2/setup-msys2@v2
-      - name: Setup MSYS2
-        shell: msys2 {0}
-        run: bash ci/scripts/msys2_setup.sh "matlab"
       - name: Install MATLAB
         uses: matlab-actions/setup-matlab@v1
       - name: Build MATLAB Interface
--- a/.github/workflows/matlab.yml
+++ b/.github/workflows/matlab.yml
@@ -103,8 +103,6 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
-      - name: Install ninja-build
-        run: choco install ninja
       - uses: msys2/setup-msys2@v2
       - name: Setup MSYS2
         shell: msys2 {0}
--- a/.github/workflows/matlab.yml
+++ b/.github/workflows/matlab.yml
@@ -94,3 +94,30 @@ jobs:
         uses: matlab-actions/run-tests@v1
         with:
           select-by-folder: matlab/test 
+  windows:
+    name: AMD64 Windows 2022 MATLAB
+    runs-on: windows-latest
+    if: ${{ !contains(github.event.pull_request.title, 'WIP') }}
+    steps:
+      - name: Check out repository        
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Install ninja-build
+        run: choco install ninja
+      - uses: msys2/setup-msys2@v2
+      - name: Setup MSYS2
+        shell: msys2 {0}
+        run: bash ci/scripts/msys2_setup.sh "matlab"
+      - name: Install MATLAB
+        uses: matlab-actions/setup-matlab@v1
+      - name: Build MATLAB Interface
+        run: bash ci/scripts/matlab_build.sh $(pwd)
+      - name: Run MATLAB Testst
+        env:
+          # Add the installation directory to the MATLAB Search Path by
+          # setting the MATLABPATH environment variable.
+          MATLABPATH: matlab/install/arrow_matlab
+        uses: matlab-actions/run-tests@v1
+        with:
+          select-by-folder: matlab/test 
--- a/.github/workflows/matlab.yml
+++ b/.github/workflows/matlab.yml
@@ -103,12 +103,6 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
-      - name: Install ninja-build
-        run: choco install ninja
-      - uses: msys2/setup-msys2@v2
-      - name: Setup MSYS2
-        shell: msys2 {0}
-        run: bash ci/scripts/msys2_setup.sh "matlab"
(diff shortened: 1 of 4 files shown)


matlab/CMakeLists.txt
```diff
@@ -542,6 +542,23 @@ if(UNIX
   endif()
 endif()
 
+if(WIN32)
```
kou, line 545: It seems that we have similar code for `Apple`.
Can we unify them by using `CMAKE_SHARED_LIBRARY_PATH` variable?
https://cmake.org/cmake/help/latest/variable/CMAKE_SHARED_LIBRARY_SUFFIX.html

kou (commented):
It seems that this job takes about 50 minutes: https://github.com/mathworks/arrow/actions/runs/5091930919/jobs/9152601067

Can we reduce build time?

assignUser:
For build time reduction using ccache as in the cpp workflows would probably be best as most of the build time comes from arrow. 

There is also an issue with the ` ilammy/msvc-dev-cmd@v1` action as it is not on the asf infra white list. Ideally we would just get rid of the action and do it manually, we might even have a script that does this already in dev/ or ci/ :D

matlab/CMakeLists.txt
leihou6116: Thank you for your suggestion. It would be great to unify the code. I'll try and get update soon.

leihou6116:
Closed accidentally.

assignUser:
Thanks this inspired me to open #35896  :D

.github/workflows/matlab.yml
```diff
@@ -94,3 +130,48 @@ jobs:
...
+        id: get-date
+        shell: bash
+        run: |
+          echo "date=$(/bin/date -u "+%Y%m%d")" >> $GITHUB_OUTPUT
+      - name: Cache ccache
+        uses: actions/cache@v3
+        with:
+          path: |
+            cpp/**
+          key: cpp-ccache-for-matlab-windows-${{ steps.get-date.outputs.date }}-${{ hashFiles('cpp/**') }}
```
assignUser, line 163: If you hash the cpp files why also add the date? Why is a cache from yesterday for state x not valid today?

leihou6116: Hi Jacob, that is a good question. I added the date such that we could track the history about when the latest cache was generated. The restore-key doesn't match the date. So it should find the latest cache even though the creation date in the key name is not the current date. My current version doesn't reduce build time. I need to figure out why, either the changing-date causing the problem or the value of "path" is not right. If you know the right value for files under cpp/**, please teach me. I really appreciate it. In the meantime, I'll do exercise by trying different values. Sorry for so many trial submissions.

assignUser: Ah I see, makes sense :)

It looks like it is working now: https://github.com/mathworks/arrow/actions/runs/5204445685/jobs/9388659172#step:9:73 :tada:

leihou6116:
I addressed all review feedbacks:

- Unified the installation code in CMakeLists when ARROW is not found.
- Used ccache to reduce build time. It saves 20-30 min on each platform.
- Removed the dependency on ilammy/msvc-dev-cmd@v1 when using Microsoft Visual Studio as compiler.

Please help review the latest change. Thanks for reviewing it.

matlab/CMakeLists.txt
```diff
@@ -520,28 +503,48 @@ if(UNIX
...
+  set_target_properties(arrow_matlab PROPERTIES INSTALL_RPATH $ORIGIN)
+endif()
+
+if(NOT Arrow_FOUND)
+  # If Arrow_FOUND is false, Arrow is built by the arrow_shared target and needs
+  # to be copied to CMAKE_PACKAGED_INSTALL_DIR.
+
+  if(APPLE)
+    # Install libarrow.dylib (symlink) and the real files it points to.
+    # on MacOS, we need to match these files: libarrow.dylib
```
kou, line 517:
```suggestion
    # On macOS, we need to match these files: libarrow.dylib
```

matlab/CMakeLists.txt
```diff
@@ -520,28 +503,48 @@ if(UNIX
...
+    #                                         libarrow.1300.dylib
+    #                                         libarrow.1300.0.0.dylib
+    # where the version number might change.
+    set(SHARED_LIBRARY_VERSION_REGEX "${ARROW_SHARED_LIB_FILENAME}(([.][0-9]+)?([.][0-9]+)?([.][0-9]+)?)${CMAKE_SHARED_LIBRARY_SUFFIX}")
```
kou, line 521: Can we simplify this?

```suggestion
    set(SHARED_LIBRARY_VERSION_REGEX "${ARROW_SHARED_LIB_FILENAME}([.][0-9]+)*${CMAKE_SHARED_LIBRARY_SUFFIX}")
```

matlab/CMakeLists.txt
```diff
@@ -520,28 +503,48 @@ if(UNIX
...
+  elseif(UNIX 
+         AND NOT APPLE
```
kou, line 523: This is redundant.

```suggestion
```

matlab/CMakeLists.txt
```diff
@@ -520,28 +503,48 @@ if(UNIX
...
+         AND NOT CYGWIN)
```
kou, line 524: Do we want to support Cygwin? If we don't, how about removing this condition?

matlab/CMakeLists.txt
```diff
@@ -520,28 +503,48 @@ if(UNIX
...
+    # Install libarrow.so (symlink) and the real files it points to.
+    # On Linux, we need to match these files: libarrow.so
+    #                                         libarrow.so.1200
+    #                                         libarrow.so.1200.0.0
+    # where the version number might change.
+    set(SHARED_LIBRARY_VERSION_REGEX "${ARROW_SHARED_LIB_FILENAME}${CMAKE_SHARED_LIBRARY_SUFFIX}(([.][0-9]+)?([.][0-9]+)?([.][0-9]+)?)")
```
kou, line 530: Can we simplify this?

```suggestion
    set(SHARED_LIBRARY_VERSION_REGEX "${ARROW_SHARED_LIB_FILENAME}${CMAKE_SHARED_LIBRARY_SUFFIX}([.][0-9]+)*")
```

matlab/CMakeLists.txt
leihou6116: Hi Kou, thank you for your review feedback. In CmakeLists, the regular expression doesn't support {}. So I cannot use regular expression like this ([.][0-9]+)){0,3} to match numbers up to 3 times. I have to literally list all of them as (([.][0-9]+)?([.][0-9]+)?([.][0-9]+)?). 

=== Cmake regular expression document. ===
https://cmake.org/cmake/help/latest/command/string.html#regex-specification

matlab/CMakeLists.txt
leihou6116: Good catch. I'll remove this. Thank you.

matlab/CMakeLists.txt
leihou6116: Hi Kou, we only want to match libarrow.so and libarrow.so.<number>.<number>.<number>. We don't want to match other files. That is why we chose to not use "*".

matlab/CMakeLists.txt
leihou6116: In the future, we might want to allocate resource to support Cygwin. It would be great that matlab/arrow supports as many platforms as possible in the long run.

kou (approved):
+1

Could you fix CMake lint CI failure by `archery lint --cmake-format --fix`?

leihou6116:
The failed and cancelled checks are unrelated to my change. The action job for only my change passed, see https://github.com/mathworks/arrow/actions/runs/5335009120

kou:
Could you fix the "Dev / Lint C++, Python, R, Docker, RAT" job by `archery lint --cmake-format --fix`?

https://github.com/apache/arrow/actions/runs/5380716556/jobs/9772450410?pr=35792#step:5:873

```text
ERROR __main__.py:618: Check failed: /arrow/matlab/CMakeLists.txt
```

See also:
* https://arrow.apache.org/docs/dev/developers/continuous_integration/archery.html
* https://arrow.apache.org/docs/dev/developers/cpp/development.html#code-style-linting-and-ci

pitrou:
@leihou6116 Can you answer @kou 's request above? Also, can you merge or rebase from git main?

leihou6116:
I'm working on the rebasing issue. This is my first time working on GitHub. I'm sorry that I messed up the rebase history. I'm working on it and try to come up with a good solution.

kou:
@kevingurney @sgilmore10 Could you help @leihou6116?

sgilmore10:
It looks like the MATLAB macOS 11 build failed due to the sporadic linker error we've seen on and off. I think if we re-run that job, it should pass.

leihou6116:
> It looks like the MATLAB macOS 11 build failed due to the sporadic linker error we've seen on and off. I think if we re-run that job, it should pass.
> 
> We're still looking into that issue. It's unfortunately difficult to reliably reproduce.

The lastest submission - fix linting error - passed on all three platforms.

kou (approved):
+1

PR #35798, opened 2023-05-26
GH-35579: [C++] Support non-named FieldRefs in Parquet scanner

### LINKED ISSUE #35579: [C++] Enable support on field_ref compute expression for also Column Indice

Current `field_ref` Expression are able to support ref by Column Name but does not offer support for Column Index.

Will be useful to also support Column index in case some integration need to pass Column index instead of Column name.

Reproduce message error:
```c++
void reproduceInferringColumnProjection(){
    const std::string& directory_base = "/diretory_of_your_parquet_file/nation_tpch/";;
    std::shared_ptr<arrow::fs::LocalFileSystem> fs =
            std::make_shared<arrow::fs::LocalFileSystem>();
    arrow::fs::FileSelector selector;
    selector.base_dir = directory_base;
    selector.recursive = true;
    std::vector<arrow::fs::FileInfo> file_infos = fs->GetFileInfo(selector).ValueOrDie();
    int num_printed = 0;
    auto format =
            std::make_shared<arrow::dataset::ParquetFileFormat>();
    arrow::dataset::FileSystemFactoryOptions options;
    std::shared_ptr<arrow::dataset::DatasetFactory> dataset_factory =
            (arrow::dataset::FileSystemDatasetFactory::Make(fs, selector, format, options)).ValueOrDie();
    std::shared_ptr<arrow::dataset::Dataset> dataset = (dataset_factory->Finish()).ValueOrDie();
    arrow::dataset::ScannerBuilder scanner_builder(dataset);
    // Error: NotImplemented: Inferring column projection from FieldRef FieldRef.FieldPath(0)
    scanner_builder.Project({compute::call("add", {compute::field_ref(0), compute::literal(10)})}, {"column_0"});
    // OK there are support in case we use Column Name instead of Column Index
    // scanner_builder.Project({compute::call("add", {compute::field_ref("n_nationkey"), compute::literal(10)})}, {"column_0"});
    std::shared_ptr<arrow::dataset::Scanner> scanner = scanner_builder.Finish().ValueOrDie();
    std::shared_ptr<arrow::Table> table = scanner->ToTable().ValueOrDie();
    std::cout << "Table with " << table->num_rows() << " rows and " << table->num_columns() << " columns" << std::endl;
    std::cout << table->ToString() << std::endl;
}
```

Message error:

```
NotImplemented: Inferring column projection from FieldRef FieldRef.FieldPath(0)
```

### Component(s)

C++

benibus:
### Rationale for this change

When setting projections/filters for the file system scanner, the Parquet implementation requires that all materialized `FieldRef`s be position-independent (containing only names). However, it may be useful to support index-based field lookups as well - assuming the dataset schema is known.

Adds a translation step for field refs prior to looking them up in the fragment schema. A known dataset schema is required to do this reliably, however (since the fragment schema may be a sub/superset of the dataset schema) - so in the absence of one, we fall back to the existing behavior.

### Are these changes tested?

Yes (tests are included)

Yes

* Closes: #35579

### DIFF
--- a/cpp/src/arrow/dataset/file_parquet.cc
+++ b/cpp/src/arrow/dataset/file_parquet.cc
@@ -224,23 +224,11 @@ Status ResolveOneFieldRef(
   return Status::OK();
 }
 
-bool IsNamedFieldRef(const FieldRef& ref) {
-  if (ref.IsName()) return true;
-  if (const auto* nested_refs = ref.nested_refs()) {
-    for (const auto& nested_ref : *nested_refs) {
-      if (!nested_ref.IsName()) return false;
-    }
-    return true;
-  }
-  return false;
-}
-
 // Converts a field ref into a position-independent ref (containing only a sequence of
 // names) based on the dataset schema. Returns `false` if no conversion was needed.
-Status MaybeConvertFieldRef(const FieldRef& ref, const Schema& dataset_schema,
-                            FieldRef* out) {
-  if (ARROW_PREDICT_TRUE(IsNamedFieldRef(ref))) {
-    return Status::OK();
+Result<FieldRef> MaybeConvertFieldRef(FieldRef ref, const Schema& dataset_schema) {
+  if (ARROW_PREDICT_TRUE(ref.IsNameSequence())) {
+    return std::move(ref);
   }
 
   ARROW_ASSIGN_OR_RAISE(auto path, ref.FindOne(dataset_schema));
@@ -254,9 +242,8 @@ Status MaybeConvertFieldRef(const FieldRef& ref, const Schema& dataset_schema,
     child_fields = &child_field.type()->fields();
   }
 
-  *out =
-      named_refs.size() == 1 ? std::move(named_refs[0]) : FieldRef(std::move(named_refs));
-  return Status::OK();
+  return named_refs.size() == 1 ? std::move(named_refs[0])
+                                : FieldRef(std::move(named_refs));
 }
 
 // Compute the column projection based on the scan options
@@ -287,7 +274,8 @@ Result<std::vector<int>> InferColumnProjection(const parquet::arrow::FileReader&
     // In the (unlikely) absence of a known dataset schema, we require that all
     // materialized refs are named.
     if (options.dataset_schema) {
-      ARROW_RETURN_NOT_OK(MaybeConvertFieldRef(ref, *options.dataset_schema, &ref));
+      ARROW_ASSIGN_OR_RAISE(
+          ref, MaybeConvertFieldRef(std::move(ref), *options.dataset_schema));
     }
     RETURN_NOT_OK(ResolveOneFieldRef(manifest, ref, field_lookup, duplicate_fields,
                                      &columns_selection));
--- a/cpp/src/arrow/dataset/file_parquet.cc
+++ b/cpp/src/arrow/dataset/file_parquet.cc
@@ -224,6 +224,41 @@ Status ResolveOneFieldRef(
   return Status::OK();
 }
 
+bool IsValidFieldRef(const FieldRef& ref) {
+  if (ref.IsName()) return true;
+  if (const auto* nested_refs = ref.nested_refs()) {
+    for (const auto& nested_ref : *nested_refs) {
+      if (!nested_ref.IsName()) return false;
+    }
+    return true;
+  }
+  return false;
+}
+
+// Converts a field ref into a position-independent ref (containing only a sequence of
+// names) based on the dataset schema. Returns `false` if no conversion was needed.
+Result<bool> ValidateFieldRef(const FieldRef& ref, const Schema& dataset_schema,
+                              FieldRef* out) {
+  if (ARROW_PREDICT_TRUE(IsValidFieldRef(ref))) {
+    return false;
+  }
+
+  ARROW_ASSIGN_OR_RAISE(auto path, ref.FindOne(dataset_schema));
+  std::vector<FieldRef> named_refs;
+  named_refs.reserve(path.indices().size());
+
+  const FieldVector* child_fields = &dataset_schema.fields();
+  for (auto index : path) {
+    const auto& child_field = *(*child_fields)[index];
+    named_refs.emplace_back(child_field.name());
+    child_fields = &child_field.type()->fields();
+  }
+
+  *out =
+      named_refs.size() == 1 ? std::move(named_refs[0]) : FieldRef(std::move(named_refs));
+  return true;
+}
+
 // Compute the column projection based on the scan options
 Result<std::vector<int>> InferColumnProjection(const parquet::arrow::FileReader& reader,
                                                const ScanOptions& options) {
@@ -248,7 +283,12 @@ Result<std::vector<int>> InferColumnProjection(const parquet::arrow::FileReader&
   }
 
   std::vector<int> columns_selection;
-  for (const auto& ref : field_refs) {
+  for (auto& ref : field_refs) {
+    // In the (unlikely) absence of a known dataset schema, we require that all
+    // materialized refs are named.
+    if (options.dataset_schema) {
+      ARROW_RETURN_NOT_OK(ValidateFieldRef(ref, *options.dataset_schema, &ref));
+    }
     RETURN_NOT_OK(ResolveOneFieldRef(manifest, ref, field_lookup, duplicate_fields,
                                      &columns_selection));
   }
--- a/cpp/src/arrow/dataset/file_parquet.cc
+++ b/cpp/src/arrow/dataset/file_parquet.cc
@@ -224,7 +224,7 @@ Status ResolveOneFieldRef(
   return Status::OK();
 }
 
-bool IsValidFieldRef(const FieldRef& ref) {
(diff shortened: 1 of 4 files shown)


cpp/src/arrow/dataset/file_parquet.cc
```diff
@@ -224,6 +224,41 @@ Status ResolveOneFieldRef(
...
+    return true;
+  }
+  return false;
+}
+
+// Converts a field ref into a position-independent ref (containing only a sequence of
+// names) based on the dataset schema. Returns `false` if no conversion was needed.
+Result<bool> ValidateFieldRef(const FieldRef& ref, const Schema& dataset_schema,
+                              FieldRef* out) {
+  if (ARROW_PREDICT_TRUE(IsValidFieldRef(ref))) {
```
mapleFU, line 242: Since you support LookUp by index, why `PREDICT_TRUE` is used here?

mapleFU (commented):
What would happen if `manifest` not matches input in nested schema? Would extra checking is required here?
Assume we have schema evolution in some of the dataset file, how to handle these file with mismatched Field?

cpp/src/arrow/dataset/file_parquet.cc
benibus: Mostly because all existing user code uses named lookups. Plus, I'd imagine indexed lookups would be fairly rare in practice since it requires knowing the exact structure (i.e. field order) of the dataset schema upfront, which may not be predictable if it was inferred from multiple files.

(It's probably not very consequential though)

benibus:
> What would happen if manifest not matches input in nested schema? Would extra checking is required here?
Assume we have schema evolution in some of the dataset file, how to handle these file with mismatched Field?

We shouldn't need any additional checks there, I don't think. The routine for resolving discrepancies between the dataset schema and file manifest (`ResolveOneFieldRef`) is unchanged - and the logic is still in terms of field names exclusively. All this PR really does is convert indexed refs into named refs (using the dataset schema) before those checks occur.

That case should probably be reflected in the tests though... (anecdotally, it _did_ work in my ad hoc testing)

mapleFU:
Yeah, I know what you mean, however, Parquet files in a dataset might has different schema, the most typical case is at: https://iceberg.apache.org/docs/latest/evolution/

Assume user insert a column, name might be better than FieldIndex, because it can maintain some consistency.

If we're sure we don't need to support that case, or user can make sure that file has same schema, then I'm +1 on this patch

cpp/src/arrow/dataset/file_parquet.cc
```diff
@@ -224,6 +224,41 @@ Status ResolveOneFieldRef(
...
+      if (!nested_ref.IsName()) return false;
+    }
+    return true;
+  }
+  return false;
+}
+
+// Converts a field ref into a position-independent ref (containing only a sequence of
+// names) based on the dataset schema. Returns `false` if no conversion was needed.
+Result<bool> ValidateFieldRef(const FieldRef& ref, const Schema& dataset_schema,
```
lidavidm, line 240: nit: the naming is a little misleading; for instance ValidateFieldRef might be more something like MaybeConvertFieldRef?

lidavidm: Additionally the return value isn't ever used so maybe just return Status?

cpp/src/arrow/dataset/file_parquet.cc
```diff
@@ -224,6 +224,41 @@ Status ResolveOneFieldRef(
   return Status::OK();
 }
 
+bool IsValidFieldRef(const FieldRef& ref) {
```
lidavidm, line 227: maybe IsNamedFieldRef?

lidavidm (approved):

lidavidm:
I'm not sure what the problem is @mapleFU? We still support named refs even with this PR. This just allows the user to also provide indices, and we resolve them into names against the overall dataset schema, so that should actually allow for schema evolution if the positions of those fields changes.

lidavidm:
I think right now, of course, we can't yet unify files of different schemas into a consistent schema. But this doesn't affect that either way.

mapleFU (approved):

mapleFU:
OK, thanks for your explanition @lidavidm .
I think when schema evolution happens, seek by index might provide unconsistent result, and seek by name doesn't has that problem. But if we we can't yet unify files of different schemas into a consistent schema, I'm +1 on this patch

lidavidm:
What (would) happen is we resolve the file schemas into an overall dataset schema, _then_ resolve any indices against that unified schema _back into names_, so that issue shouldn't come up

mapleFU:
Okay, I think currently `parquet::SchemaManifest` can build the bridge from arrow to parquet and parquet to arrow, but for file with different schema, it need to follow some rules, maybe by `"PARQUET:field_id"` or others. I'm ok on this patch now!

lidavidm:
There's build failures, but I think they're unrelated?

benibus:
> There's build failures, but I think they're unrelated?

I think so... I'm seeing similar macos failures elsewhere after a fresh rebase

lidavidm:
Probably what happened in https://github.com/conda-forge/cpp-opentelemetry-sdk-feedstock/issues/29

We need to set WITH_STL=ON for the bundled OpenTelemetry build

cpp/src/arrow/dataset/file_parquet.cc
```diff
@@ -248,7 +283,12 @@ Result<std::vector<int>> InferColumnProjection(const parquet::arrow::FileReader&
   }
 
   std::vector<int> columns_selection;
-  for (const auto& ref : field_refs) {
+  for (auto& ref : field_refs) {
+    // In the (unlikely) absence of a known dataset schema, we require that all
+    // materialized refs are named.
+    if (options.dataset_schema) {
```
westonpace, line 289: Heh, at this point, if we don't have a dataset schema, I think something has gone very wrong.

cpp/src/arrow/dataset/file_parquet.cc
```diff
@@ -224,6 +224,41 @@ Status ResolveOneFieldRef(
...
+      if (!nested_ref.IsName()) return false;
+    }
+    return true;
+  }
+  return false;
+}
+
+// Converts a field ref into a position-independent ref (containing only a sequence of
+// names) based on the dataset schema. Returns `false` if no conversion was needed.
+Status MaybeConvertFieldRef(const FieldRef& ref, const Schema& dataset_schema,
```
westonpace, line 240: Minor nit: Can we return `Result<FieldRef>` instead?

westonpace (commented):
This looks good, just a few thoughts.

cpp/src/arrow/dataset/file_parquet.cc
benibus: Haha I figured as much... That being said, some of the existing tests do leave it null.

cpp/src/arrow/dataset/file_parquet_test.cc
```diff
@@ -694,5 +694,55 @@ TEST(TestParquetStatistics, NullMax) {
   EXPECT_EQ(stat_expression->ToString(), "(x >= 1)");
 }
 
+// Tests round-trip projection with nested/indexed FieldRefs
+// https://github.com/apache/arrow/issues/35579
+TEST(TestRoundTrip, ProjectedFieldRefs) {
```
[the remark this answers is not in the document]
benibus: Yeah, I only did it this way to more or less match the example in the original issue. Using the existing fixtures would certainly be more convenient. AFAIK the only difference is that it would test the intermediate batch rather than doing the full round trip on the projection - but that's consistent with the other tests.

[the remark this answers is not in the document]
westonpace: Ok.  I don't feel too strongly here.  So feel free to let me know which you prefer.

cpp/src/arrow/dataset/file_parquet.cc
```diff
@@ -224,6 +224,41 @@ Status ResolveOneFieldRef(
...
+bool IsNamedFieldRef(const FieldRef& ref) {
+  if (ref.IsName()) return true;
+  if (const auto* nested_refs = ref.nested_refs()) {
+    for (const auto& nested_ref : *nested_refs) {
+      if (!nested_ref.IsName()) return false;
+    }
+    return true;
+  }
+  return false;
+}
```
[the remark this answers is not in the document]
benibus: Moved it into `FieldRef` in the update. Looking at it now though, I suspect it may be too niche to justify its place there - at least on its own.

Methods that transform a ref into a flat vector of names or indices might be more useful in general (but less trivial, of course).

westonpace (approved):

cpp/src/arrow/dataset/file_parquet.cc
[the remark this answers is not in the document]
westonpace: I don't think it's too niche.  I feel like I have run into situations a few times now in the scanner where I've needed to know if a ref is all-names, all-indices, or mixed (a lot of the new scanner stuff normalizes to all-indices).  We do have FieldPath already which is a flat vector of indices.

westonpace:
CI issues seem unrelated

PR #35810, opened 2023-05-28
GH-35809: [C#] Improvements to the C Data Interface

### LINKED ISSUE #35809: [C#] Issues with the C data interface implementation.

I took a look at the code in the `Apache.Arrow.C` namespace and noticed the following things:

* The native callbacks can use `[UnmanagedCallersOnly]` on supported frameworks (.NET 5+). This way, taking the pointer to them will be more efficient and does not involve delegates.
* We are manually zeroing the structs we just allocated (https://github.com/apache/arrow/blob/05fe0d25834fd1629d71ceb51f0281b44a511f94/csharp/src/Apache.Arrow/C/CArrowArray.cs#L54-L63 for example). We can zero them by just writing `*ptr = default` which is more efficient.
* When releasing an exported array stream [we don't free the `GCHandle` to the managed object](https://github.com/apache/arrow/blob/05fe0d25834fd1629d71ceb51f0281b44a511f94/csharp/src/Apache.Arrow/C/CArrowArrayStreamExporter.cs#L116-L121), causing a memory leak.
* When a function on an imported array fails, we don't include the result of `get_last_error` in the exception message.
* We free the last exported array stream error message with `Marshal.AllocCoTaskMem`, while we had allocated it with `Marshal.AllocHGlobal`. https://github.com/apache/arrow/blob/05fe0d25834fd1629d71ceb51f0281b44a511f94/csharp/src/Apache.Arrow/C/CArrowArrayStreamExporter.cs#L173
* And more smaller stuff.

I will open a PR to address them.

### Component(s)

C#

teo-tsirpanis:
### Rationale for this change

This PR fixes issues identified while reading the code of the `Apache.Arrow.C` namespace.

See each commit message for more details.

### Are these changes tested?

Using the existing test suite.

No.
* Closes: #35809


kou:
> See each commit message for more details.

Could you write it in the pull request description because we use squash merge and the merge commit only uses the pull request description.

It seems that this pull request has several logical changes. It may be better that you split this pull request and associated issue to multiple small pull requests/issues for easy to review.

csharp/src/Apache.Arrow/C/CArrowArrayExporter.cs
```diff
@@ -15,16 +15,21 @@
...
 using System;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using Apache.Arrow.Memory;
 
 namespace Apache.Arrow.C
 {
     public static class CArrowArrayExporter
     {
+#if NET5_0_OR_GREATER
```
pitrou, line 26: Perhaps add a comment explaining why this needs .Net 5.0+? Or will it be obvious to a C# developer?

csharp/src/Apache.Arrow/C/CArrowArrayExporter.cs
```diff
@@ -203,6 +207,10 @@ private unsafe static void ReleaseArray(CArrowArray* cArray)
         private unsafe static void Dispose(void** ptr)
         {
             GCHandle gch = GCHandle.FromIntPtr((IntPtr)(*ptr));
+            if (!gch.IsAllocated)
```
pitrou, line 210: Add a comment explaining when this might occur? (perhaps if an exception occurs while exporting the array?)

csharp/src/Apache.Arrow/C/CArrowArrayExporter.cs
teo-tsirpanis: `UnmanagedCallersOnlyAttribute` was introduced in .NET 5.

csharp/src/Apache.Arrow/C/CArrowArrayExporter.cs
```diff
@@ -184,13 +189,12 @@ private unsafe static void ConvertRecordBatch(ExportedAllocationOwner sharedOwne
             cArray->dictionary = null;
         }
 
+#if NET5_0_OR_GREATER
+        [UnmanagedCallersOnly(CallConvs = new[] { typeof(CallConvStdcall) })]
```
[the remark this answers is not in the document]
teo-tsirpanis: See https://github.com/apache/arrow/pull/34133#discussion_r1122532439. Ideally we would have used the default calling convention, but that would not be suppported on anything earlier than .NET 5. And in 64-bit platforms the calling convention doesn't matter either way.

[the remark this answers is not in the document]
pitrou: Well, the entire point (mostly) of implementing the C Data Interface is to be compatible with non-.Net producers/consumers. Those are extremely likely to use the platform default. So we should get it right at least when possible, i.e. on .Net >= 5.0.

As for https://stackoverflow.com/questions/34832679/is-the-callingconvention-ignored-in-64-bit-net-applications , does it apply here? It's talking about `DllImport`, which might be different from this?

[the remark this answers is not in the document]
pitrou: In any case, keeping the default convention seems more theoretically sound (and forward-looking, perhaps).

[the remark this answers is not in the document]
pitrou: cc @westonpace @lidavidm for opinions.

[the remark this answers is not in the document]
teo-tsirpanis: If I understand correctly, are you saying to change this: https://github.com/apache/arrow/blob/e628ca50078c749ea7a10b53defe12cbec7d581f/csharp/src/Apache.Arrow/C/CArrowArray.cs#L41

into this?

```csharp
#if NET5_0_OR_GREATER
public delegate* unmanaged<CArrowArray*, void> release;
#else
public delegate* unmanaged[Stdcall]<CArrowArray*, void> release;
#endif
```

That would cause an incompatible API surface between the assembly compiled for .NET 6 and that compiled for the earlier frameworks. We have two options:

* Lie and keep the `stdcall` calling convention on the function pointers.
* Use the default unmanaged calling convention but support the C interface only on .NET 6+ (we don't target 5 as it is unsupported).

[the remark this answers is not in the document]
pitrou: If you take a look for example at the  Arrow C++ implementation, its release functions are entirely private. For example `ReleaseExportedSchema` below is in the anonymous namespace, which doesn't expose the function publicly:
https://github.com/apache/arrow/blob/e628ca50078c749ea7a10b53defe12cbec7d581f/cpp/src/arrow/c/bridge.cc#L107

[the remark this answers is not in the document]
teo-tsirpanis: So I should make the members of these structs private? That also works.

westonpace (approved):
It sounds like we might've worked through most of the issues.  If I understand it correctly, if someone wants to use 32-bit windows, then they need to use .NET 5 or greater, which seems like a reasonable tradeoff to me.

Given we don't have any 32-bit tests for any of the C# APIs, there may be other issues as well.  We can maybe just move forward and worry about 32-bit APIs if a user emerges?

CurtHagenlocher:
> It sounds like we might've worked through most of the issues. If I understand it correctly, if someone wants to use 32-bit windows, then they need to use .NET 5 or greater, which seems like a reasonable tradeoff to me.

I'd understood it differently (though perhaps incorrectly). I thought that using the explicit [Stdcall] would work everywhere except 32-bit non-Windows platforms, while removing the [Stdcall] would prevent things from working on .NET < 5.

westonpace:
> I'd understood it differently (though perhaps incorrectly). I thought that using the explicit [Stdcall] would work everywhere except 32-bit non-Windows platforms, while removing the [Stdcall] would prevent things from working on .NET < 5.

Yes, looking at the code now I think you're right.  I interpreted the above conversation as:

 * [This](https://github.com/apache/arrow/pull/35810#discussion_r1210668239) would be safe but lead to an unstable API surface
 * If we make these methods/fields private we don't have to worry about the API surface
 
However, looking at the code, I don't see that change.  Are we still waiting on that change?  Or is there some reason this wouldn't work?

teo-tsirpanis:
I will make the fields `internal` soon; I'm pretty busy these days.

teo-tsirpanis:
I made them internal.

teo-tsirpanis:
Done @westonpace.

pitrou:
@teo-tsirpanis @westonpace Is there anything left to do here? It would be nice to have this in 13.0.0.

pitrou:
I will say again that I think the callbacks should be `Cdecl`, not `Stdcall`. `Stdcall` is the calling convention for WinAPI functions and nothing else, AFAICT. `Cdecl` is the default calling convention for any other C/C++ code, such as the callbacks that producers will pass in the `release` member.

See: https://www.codeproject.com/Articles/1388/Calling-Conventions-Demystified

pitrou:
(and, yes, it would only make a difference on 32-bit x86 machines, and even then, perhaps in some cases the two are equivalent)

westonpace (approved):
I'm still not entirely an expert in calling conventions, and I don't fully understand.  I do think the default should be cdecl.  However, given this only affects extremely rare systems, and we don't even know if there are any users using those systems, I don't want to hold up legitimate fixes while worrying over calling conventions.

pitrou:
Feel free to merge :-)

westonpace:
Sorry, didn't see your comment on #36506 but it looks like you got the change in just in time.  Fortunately, it seems CI is passing :)  I'll keep monitoring it.

westonpace:
All tests passed, thanks for your attention and the cleanup @teo-tsirpanis

csharp/src/Apache.Arrow/C/CArrowArrayStream.cs
```diff
@@ -35,15 +35,23 @@ public unsafe struct CArrowArrayStream
         ///
         /// Return value: 0 if successful, an `errno`-compatible error code otherwise.
         ///</summary>
-        public delegate* unmanaged[Stdcall]<CArrowArrayStream*, CArrowSchema*, int> get_schema;
```
eerhardt, line 38: If we make these function pointers `internal`, how are users of these APIs supposed to call them?

teo-tsirpanis: They should not; the importer will take care of calling them.

eerhardt: I don't fully follow. If I wanted to take some C arrow data (let's say coming from C++) and natively work with it - not using the "normal" C# ArrowTypes, Schema, RecordBatches, etc. but instead manually calling into the native functions on the CArrowArrayStream, that is no longer possible? Why not?

The reason for interoping at the native layer would be for performance - I wouldn't need to allocate a bunch of managed objects just to interact with the Arrow information coming from some other library.

teo-tsirpanis: See https://github.com/apache/arrow/pull/35810#issuecomment-1584531095 for an explanation on why the function pointer fields are internal.

eerhardt: Another option would be to only publicly expose the function pointer fields on `net5.0+`. Keep them `internal` for netstandard and netfx where they need to be declared `Cdecl` (and honestly function pointers aren't really meant to be used on netfx and netstandard since they came in C# 9 - https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version).

That way users on net5+ can still call the function pointers, if they need to. For netstandard and netfx, they are no worse off - they can't call them with this change anyway.

PR #35848, opened 2023-05-31
GH-35846: [Python] Bump min required numpy version to 1.22.2

### LINKED ISSUE #35846: Minimum required numpy version (1.16.6) has security vulnerability

We currently use version 11.0.0 of pyarrow. We run Snyk checks to find vulnerabilities in our third party dependencies

In the version we use, but also on main, the setup.py file states numpy >= 1.16.6 is required (https://github.com/apache/arrow/blob/main/python/setup.py#L451)

This Python version is affected by the following vulnerability: https://security.snyk.io/vuln/SNYK-PYTHON-NUMPY-2321964

To solve the issue, version 1.22.2 or higher should be required

### Component(s)

Python

diegohavenstein:
The minimum required Numpy version (1.16.6) is affected by the following vulnerability:
https://security.snyk.io/vuln/SNYK-PYTHON-NUMPY-2321964

With this commit, the minimum required numpy version is increased to 1.22.2, such that Snyk checks do not fail anymore.

* Closes: #35846

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -448,7 +448,7 @@ class BinaryDistribution(Distribution):
 
 
 install_requires = (
-    'numpy >= 1.16.6',
+    'numpy >= 1.22.2',
 )
 
 


pitrou:
Why would we do this? You can upgrade Numpy to a newer version if you want, but we shouldn't dictate users' security policy.

Also, this "security" issue looks more theoretical than practical (you need a memory allocation to fail).

PR #35851, opened 2023-05-31
MINOR [Go] Add nested struct test for pqarrow

candiduslynx:
@zeroshade I've found a disturbing behavior with structs (namely, nested ones) in pqarrow.

This blocks https://github.com/cloudquery/filetypes/pull/172

Could you please take a look?

### DIFF
--- a/go/parquet/pqarrow/encode_arrow_test.go
+++ b/go/parquet/pqarrow/encode_arrow_test.go
@@ -931,6 +931,64 @@ func (ps *ParquetIOTestSuite) TestReadDecimals() {
 	ps.True(array.Equal(expected, chunked.Chunk(0)))
 }
 
+func (ps *ParquetIOTestSuite) TestReadNestedStruct() {
+	mem := memory.NewCheckedAllocator(memory.DefaultAllocator)
+	defer mem.AssertSize(ps.T(), 0)
+
+	dt := arrow.StructOf(arrow.Field{
+		Name: "nested",
+		Type: arrow.StructOf(
+			arrow.Field{Name: "bool", Type: arrow.FixedWidthTypes.Boolean},
+			arrow.Field{Name: "int32", Type: arrow.PrimitiveTypes.Int32},
+			arrow.Field{Name: "int64", Type: arrow.PrimitiveTypes.Int64},
+		),
+	})
+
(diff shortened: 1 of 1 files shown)


go/parquet/pqarrow/encode_arrow_test.go
```diff
@@ -931,6 +931,64 @@ func (ps *ParquetIOTestSuite) TestReadDecimals() {
...
+	rgw := writer.AppendRowGroup()
+	cw, err := rgw.NextColumn()
+	ps.NoError(err)
+
+	props := pqarrow.NewArrowWriterProperties(pqarrow.WithAllocator(mem))
+	ctx := pqarrow.NewArrowWriteContext(context.TODO(), &props)
+	ps.NoError(pqarrow.WriteArrowToColumn(ctx, cw, expected, nil, nil, false))
+	ps.NoError(cw.Close())
+	ps.NoError(rgw.Close())
+	ps.NoError(writer.Close())
```
zeroshade, line 978: I'm extremely confused here, you're writing a nested struct arrow array to a column expecting a single leaf bytearray of decimal data.....? It feels like this should error at a minimum. There's no way this should work, unless I'm missing something.

candiduslynx:
@zeroshade I found an issue with the test & it should now pass, but I'm really confused about `array.NewStructData` behavior in conjunction with pqarrow.
Could you please take a look at the code in https://github.com/cloudquery/filetypes/pull/172?
Specifically, the issue I see is that the offsets/null bitmaps for the nested field are propagated strangely.
It [panics](https://github.com/cloudquery/filetypes/actions/runs/5134969218/jobs/9239730687?pr=172#step:6:8) while trying to create Struct array from data, so I don't know where to look.

candiduslynx:
I guess the issue can be connected to https://github.com/apache/arrow/blob/main/go/parquet/pqarrow/encode_arrow_test.go#L1040-L1042
```
// current impl of ArrayEquals for structs doesn't correctly handle nulls in the parent
// with a non-nullable child when comparing. Since after the round trip, the data in the
// child will have the nulls, not the original data.
```

candiduslynx:
I'm closing this for now, however, it would've been nice to be able to use `array.NewStructData` on the read values

PR #35860, opened 2023-05-31
GH-35730: [C++] Add the ability to specify custom schema on a dataset write

### LINKED ISSUE #35730: [Python] write_dataset does not preserve non-nullable columns in schema

When writing a table whose schema has nullable columns using write_dataset the nullable info is not saved

To reproduce

```
import pyarrow as pa
import pyarrow.parquet as pq
import datetime as dt
import pyarrow.dataset as ds

table = pa.Table.from_arrays([[1,2,3],[None,5,None], [dt.date(2023,1,1), dt.date(2023,1,2), dt.date(2023,1,3)]],
    schema=pa.schema([pa.field("x", pa.int64(), nullable=False), pa.field("y", pa.int64(), nullable=True), pa.field("date", pa.date32(), nullable=True)]))
print(table.schema)
# schema shows  column 'x' as not nullable

pq.write_to_dataset(table, parquet_test1", partitioning=['date'], partitioning_flavor='hive')
dataset = ds.dataset("parquet_test1", format="parquet", partitioning="hive")
dataset.to_table().schema
# column 'x' is nullable

pa.dataset.write_dataset(table, "parquet_test2", partitioning=['date'], partitioning_flavor='hive', format='parquet')
dataset = ds.dataset("parquet_test2", format="parquet", partitioning="hive")
dataset.to_table().schema
# column 'x' is nullable

```

### Component(s)

Python

westonpace:
### Rationale for this change

The dataset write node previously allowed you to specify custom key/value metadata on a write node.  This was added to support saving schema metadata.  However, it doesn't capture field metadata or field nullability.  This PR replaces that capability with the ability to specify a custom schema instead.  The custom schema must have the same number of fields as the input to the write node and each field must have the same type.

Added `custom_schema` to `WriteNodeOptions` and removed `custom_metadata`.

### Are these changes tested?

Yes, I added a new C++ unit test to verify that the custom info is applied to written files.

No.  Only new functionality (which is user facing)

* Closes: #35730

### DIFF
--- a/cpp/src/arrow/dataset/file_base.cc
+++ b/cpp/src/arrow/dataset/file_base.cc
@@ -387,16 +387,16 @@ Status WriteBatch(
 
 class DatasetWritingSinkNodeConsumer : public acero::SinkNodeConsumer {
  public:
-  DatasetWritingSinkNodeConsumer(std::shared_ptr<const KeyValueMetadata> custom_metadata,
+  DatasetWritingSinkNodeConsumer(std::shared_ptr<Schema> custom_schema,
                                  FileSystemDatasetWriteOptions write_options)
-      : custom_metadata_(std::move(custom_metadata)),
+      : custom_schema_(std::move(custom_schema)),
         write_options_(std::move(write_options)) {}
 
   Status Init(const std::shared_ptr<Schema>& schema,
               acero::BackpressureControl* backpressure_control,
               acero::ExecPlan* plan) override {
-    if (custom_metadata_) {
-      schema_ = schema->WithMetadata(custom_metadata_);
+    if (custom_schema_) {
+      schema_ = custom_schema_;
     } else {
       schema_ = schema;
     }
@@ -434,7 +434,7 @@ class DatasetWritingSinkNodeConsumer : public acero::SinkNodeConsumer {
                       });
   }
 
-  std::shared_ptr<const KeyValueMetadata> custom_metadata_;
+  std::shared_ptr<Schema> custom_schema_;
   std::unique_ptr<internal::DatasetWriter> dataset_writer_;
   FileSystemDatasetWriteOptions write_options_;
   Future<> finished_ = Future<>::Make();
@@ -453,13 +453,13 @@ Status FileSystemDataset::Write(const FileSystemDatasetWriteOptions& write_optio
 
   // The projected_schema is currently used by pyarrow to preserve the custom metadata
   // when reading from a single input file.
-  const auto& custom_metadata = scanner->options()->projected_schema->metadata();
+  const auto& custom_schema = scanner->options()->projected_schema;
 
   acero::Declaration plan = acero::Declaration::Sequence({
       {"scan", ScanNodeOptions{dataset, scanner->options()}},
       {"filter", acero::FilterNodeOptions{scanner->options()->filter}},
       {"project", acero::ProjectNodeOptions{std::move(exprs), std::move(names)}},
-      {"write", WriteNodeOptions{write_options, custom_metadata}},
+      {"write", WriteNodeOptions{write_options, custom_schema}},
   });
 
   return acero::DeclarationToStatus(std::move(plan), scanner->options()->use_threads);
@@ -475,16 +475,38 @@ Result<acero::ExecNode*> MakeWriteNode(acero::ExecPlan* plan,
 
   const WriteNodeOptions write_node_options =
       checked_cast<const WriteNodeOptions&>(options);
-  const std::shared_ptr<const KeyValueMetadata>& custom_metadata =
-      write_node_options.custom_metadata;
+  const std::shared_ptr<Schema>& custom_schema = write_node_options.custom_schema;
   const FileSystemDatasetWriteOptions& write_options = write_node_options.write_options;
+
+  const std::shared_ptr<Schema>& input_schema = inputs[0]->output_schema();
+
+  if (custom_schema != nullptr) {
+    if (custom_schema->num_fields() != input_schema->num_fields()) {
+      return Status::Invalid(
+          "The provided custom_schema did not have the same number of fields as the "
+          "data.  The custom schema can only be used to add metadata / nullability to "
+          "fields and cannot change the type or number of fields.");
+    }
+    for (int field_idx = 0; field_idx < input_schema->num_fields(); field_idx++) {
+      if (!input_schema->field(field_idx)->type()->Equals(
+              custom_schema->field(field_idx)->type())) {
+        return Status::Invalid("The provided custom_schema specified type ",
+                               custom_schema->field(field_idx)->type()->ToString(),
+                               " for field ", field_idx, "and the input data has type ",
+                               input_schema->field(field_idx),
+                               "The custom schema can only be used to add metadata / "
+                               "nullability to fields and "
+                               "cannot change the type or number of fields.");
+      }
+    }
+  }
 
   if (!write_options.partitioning) {
     return Status::Invalid("Must provide partitioning");
   }
 
   std::shared_ptr<DatasetWritingSinkNodeConsumer> consumer =
-      std::make_shared<DatasetWritingSinkNodeConsumer>(custom_metadata, write_options);
+      std::make_shared<DatasetWritingSinkNodeConsumer>(custom_schema, write_options);
 
   ARROW_ASSIGN_OR_RAISE(
       auto node,
--- a/cpp/src/arrow/dataset/file_base.cc
+++ b/cpp/src/arrow/dataset/file_base.cc
@@ -455,11 +455,14 @@ Status FileSystemDataset::Write(const FileSystemDatasetWriteOptions& write_optio
   // when reading from a single input file.
   const auto& custom_schema = scanner->options()->projected_schema;
 
+  WriteNodeOptions write_node_options(write_options);
+  write_node_options.custom_schema = custom_schema;
+
   acero::Declaration plan = acero::Declaration::Sequence({
       {"scan", ScanNodeOptions{dataset, scanner->options()}},
       {"filter", acero::FilterNodeOptions{scanner->options()->filter}},
       {"project", acero::ProjectNodeOptions{std::move(exprs), std::move(names)}},
-      {"write", WriteNodeOptions{write_options, custom_schema}},
+      {"write", std::move(write_node_options)},
   });
 
   return acero::DeclarationToStatus(std::move(plan), scanner->options()->use_threads);
--- a/cpp/src/arrow/dataset/file_base.cc
+++ b/cpp/src/arrow/dataset/file_base.cc
@@ -475,14 +475,22 @@ Result<acero::ExecNode*> MakeWriteNode(acero::ExecPlan* plan,
 
   const WriteNodeOptions write_node_options =
       checked_cast<const WriteNodeOptions&>(options);
-  const std::shared_ptr<Schema>& custom_schema = write_node_options.custom_schema;
+  std::shared_ptr<Schema> custom_schema = write_node_options.custom_schema;
+  const std::shared_ptr<const KeyValueMetadata>& custom_metadata =
+      write_node_options.custom_metadata;
   const FileSystemDatasetWriteOptions& write_options = write_node_options.write_options;
 
   const std::shared_ptr<Schema>& input_schema = inputs[0]->output_schema();
 
   if (custom_schema != nullptr) {
+    if (custom_metadata) {
+      return Status::TypeError(
+          "Do not provide both custom_metadata and custom_schema.  If custom_schema is "
+          "used then custom_schema->metadata should be used instead of custom_metadata");
+    }
+
     if (custom_schema->num_fields() != input_schema->num_fields()) {
-      return Status::Invalid(
+      return Status::TypeError(
           "The provided custom_schema did not have the same number of fields as the "
           "data.  The custom schema can only be used to add metadata / nullability to "
           "fields and cannot change the type or number of fields.");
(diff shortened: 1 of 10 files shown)


westonpace:
Starting in draft stage as I believe we should get some python unit tests in here and @anjakefala volunteered to do some.

anjakefala:
So far, I added 2 basic tests, based on my understanding of the feature!

The basic case where you write a single table, which contains a field with nullability specified, passes.

Note that this one:

```
  # we can specify the nullability of a field through the schema                                                                                   
    pa.dataset.write_dataset(table_no_null, tempdir/"nulltest2", schema=schema_nullable)                                                             
    dataset = ds.dataset(tempdir/"nulltest2", format="parquet")                                                                                      
    assert dataset.to_table().schema.equals(schema_nullable) 
```

is failing for now. I did not specify nullability in the table's schema, but then specified it in `write_dataset(schema=)`.

Is it expected that the returned dataset would have a field with nullability?

In this example `table` has a field specified with nullability, while `table_no_null` does not:

```
pa.dataset.write_dataset([table_no_null, table], tempdir/"nulltest2", schema=schema_nullable) 
```

The resulting schema also does not have nullability.

anjakefala:
I have confirmed that

```
 pa.dataset.write_dataset(table, tempdir/"nulltest2", schema=schema_nullable, format="parquet") 
```
has a field with nullability.

The following do not:

```
pa.dataset.write_dataset([table_no_null, table], tempdir/"nulltest2", schema=schema_nullable, format="parquet")  
```

or 

```
pa.dataset.write_dataset([table, table_no_null], tempdir/"nulltest2", schema=schema_nullable, format="parquet") 
```

westonpace:
> The following do not:
> 
> pa.dataset.write_dataset([table_no_null, table], tempdir/"nulltest2", schema=schema_nullable, format="parquet")  
> 
> or
> 
> pa.dataset.write_dataset([table, table_no_null], tempdir/"nulltest2", schema=schema_nullable, format="parquet") 
> 

These lines failed for me with the following error:

```
pyarrow/dataset.py:936: in write_dataset
    data = InMemoryDataset(data, schema=schema)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   raise ArrowTypeError(
E   pyarrow.lib.ArrowTypeError: Item has schema
E   x: int64
E   y: int64
E   which does not match expected schema
E   x: int64 not null
E   y: int64
```

I thought this was supported and it took me a moment to track down what was going on.  The error is actually being raised before the C++ call to write the dataset.  Pyarrow is taking the two inputs (`table`, `table_no_null`) and trying to put them in an `InMemoryDataset` and specifying the schema.  The constructor for `InMemoryDataset` is verifying that all the tables it has been given have the same schema and throwing an error because it was given a table whose schema does not match the dataset's schema.

If this is the same error you were getting then I think we can call this an invalid scenario and we don't have to support it (at least for this PR.  Arguably, you could evolve a table into the correct schema if adding it to an InMemoryDataset but that's a different feature).

This is kind of confusing because @anjakefala and I were testing earlier and you are allowed to create an `InMemoryDataset` with tables / batches who have the same types / nullability but different field metadata.  So I created an additional python test case for field metadata which does verify the "two tables but mixed metadata can be overridden by an explicit schema" call.

westonpace:
Oh, and thank you for adding the test case :)

anjakefala:
> If this is the same error you were getting then I think we can call this an invalid scenario and we don't have to support it (at least for this PR. Arguably, you could evolve a table into the correct schema if adding it to an InMemoryDataset but that's a different feature).

Good to know!

jorisvandenbossche (approved):
Looks good to me! 

I just have one question about whether we should also validate the names of the schema; the rest are minor code style nitpicks in the test code, for which I can also push some changes.

cpp/src/arrow/dataset/file_base.cc
```diff
@@ -475,16 +475,38 @@ Result<acero::ExecNode*> MakeWriteNode(acero::ExecPlan* plan,
...
-  const std::shared_ptr<const KeyValueMetadata>& custom_metadata =
-      write_node_options.custom_metadata;
+  const std::shared_ptr<Schema>& custom_schema = write_node_options.custom_schema;
   const FileSystemDatasetWriteOptions& write_options = write_node_options.write_options;
 
+  const std::shared_ptr<Schema>& input_schema = inputs[0]->output_schema();
+
+  if (custom_schema != nullptr) {
+    if (custom_schema->num_fields() != input_schema->num_fields()) {
+      return Status::Invalid(
```
pitrou, line 485: `Status::TypeError` here?

cpp/src/arrow/dataset/file_base.cc
```diff
@@ -475,16 +475,38 @@ Result<acero::ExecNode*> MakeWriteNode(acero::ExecPlan* plan,
...
+          "The provided custom_schema did not have the same number of fields as the "
+          "data.  The custom schema can only be used to add metadata / nullability to "
+          "fields and cannot change the type or number of fields.");
+    }
+    for (int field_idx = 0; field_idx < input_schema->num_fields(); field_idx++) {
+      if (!input_schema->field(field_idx)->type()->Equals(
+              custom_schema->field(field_idx)->type())) {
+        return Status::Invalid("The provided custom_schema specified type ",
```
pitrou, line 493: Same here.

cpp/src/arrow/dataset/file_base.h
```diff
@@ -463,15 +463,21 @@ struct ARROW_DS_EXPORT FileSystemDatasetWriteOptions {
...
-      std::shared_ptr<const KeyValueMetadata> custom_metadata = NULLPTR)
-      : write_options(std::move(options)), custom_metadata(std::move(custom_metadata)) {}
+  explicit WriteNodeOptions(FileSystemDatasetWriteOptions options,
+                            std::shared_ptr<Schema> custom_schema = NULLPTR)
+      : write_options(std::move(options)), custom_schema(std::move(custom_schema)) {}
 
   /// \brief Options to control how to write the dataset
   FileSystemDatasetWriteOptions write_options;
-  /// \brief Optional metadata to attach to written batches
-  std::shared_ptr<const KeyValueMetadata> custom_metadata;
```
jorisvandenbossche, line 474: Instead of removing this option (as a breaking change), we could in theory still allow the user to specify one of both?

(I am not using the C++ API for this, so I don't know how useful this would be / how cumbersome it is to specify the schema if you only want to specify metadata. From the DatasetWriter point of view, this is a fine change of course since there we already have the full schema)

paleolimbot (approved):
I looked at the R bits and the change seems reasonable; however, I would second that we make this a non-source-breaking change.

cpp/src/arrow/dataset/file_base.cc
westonpace: Switched.

cpp/src/arrow/dataset/file_base.h
westonpace: If it were a new feature I would argue it's not worth it (A user could technically use `DeclarationToSchema` to get the output schema of the plan leading up to the write and then attach custom metadata to that).  However, given we have already released `custom_metadata`, and I would like Acero's API to start being stable, I suppose I should set an example.  Thanks for the nudge.  I have restored `custom_metadata`

westonpace:
> I looked at the R bits and the change seems reasonable; however, I would second that we make this a non-source-breaking change.

Thank you.  I have restored the old field so this is no longer a breaking change.

cpp/src/arrow/dataset/file_base.h
```diff
@@ -463,13 +463,21 @@ struct ARROW_DS_EXPORT FileSystemDatasetWriteOptions {
 /// \brief Wraps FileSystemDatasetWriteOptions for consumption as compute::ExecNodeOptions
 class ARROW_DS_EXPORT WriteNodeOptions : public acero::ExecNodeOptions {
  public:
-  explicit WriteNodeOptions(
-      FileSystemDatasetWriteOptions options,
-      std::shared_ptr<const KeyValueMetadata> custom_metadata = NULLPTR)
-      : write_options(std::move(options)), custom_metadata(std::move(custom_metadata)) {}
+  explicit WriteNodeOptions(FileSystemDatasetWriteOptions options,
+                            std::shared_ptr<Schema> custom_schema = NULLPTR)
```
jorisvandenbossche, line 467: Should still constructor also have custom_metadata be added back?  
Or do you only use this in attribute-setting style (`write_options.custom_metadata = ..`) ?

cpp/src/arrow/dataset/file_base.cc
jorisvandenbossche: Yes, to be honest for having a wrong number of fields, I would also raise a ValueError/Invalid (it's a wrong value, not a wrong type)

cpp/src/arrow/dataset/file_base.cc
```diff
@@ -475,16 +475,38 @@ Result<acero::ExecNode*> MakeWriteNode(acero::ExecPlan* plan,
...
+  if (custom_schema != nullptr) {
+    if (custom_schema->num_fields() != input_schema->num_fields()) {
+      return Status::Invalid(
+          "The provided custom_schema did not have the same number of fields as the "
+          "data.  The custom schema can only be used to add metadata / nullability to "
+          "fields and cannot change the type or number of fields.");
+    }
+    for (int field_idx = 0; field_idx < input_schema->num_fields(); field_idx++) {
+      if (!input_schema->field(field_idx)->type()->Equals(
+              custom_schema->field(field_idx)->type())) {
```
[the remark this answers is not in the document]
westonpace: Changing the names should be safe.  Admittedly, a user could also do this name change by inserting a project node before the write node.

I could be convinced otherwise but I don't think this does any harm and I think, as a user, I would expect this behavior, so it wouldn't be surprising that the names changed.

cpp/src/arrow/dataset/file_base.h
westonpace: Fixed.

cpp/src/arrow/dataset/file_base.cc
pitrou: A schema error is a type error IMHO.

westonpace:
The failing test on AMD64 is unrelated (flight ucx test).  There is still one R test running but I think this is good to go now.

cpp/src/arrow/dataset/file_base.h
```diff
@@ -470,6 +471,15 @@ class ARROW_DS_EXPORT WriteNodeOptions : public acero::ExecNodeOptions {
 
   /// \brief Options to control how to write the dataset
   FileSystemDatasetWriteOptions write_options;
+  /// \brief Optional schema to attach to all written batches
+  ///
+  /// By default, we will use the output schema of the input.
+  ///
+  /// This can be used to alter schema metadata, field nullability, or field metadata
```
pitrou, line 478: Nit
```suggestion
  /// This can be used to alter schema metadata, field nullability, or field metadata.
```

cpp/src/arrow/dataset/file_base.h
```diff
@@ -34,6 +34,7 @@
 #include "arrow/filesystem/filesystem.h"
 #include "arrow/io/file.h"
 #include "arrow/util/compression.h"
+#include "arrow/util/key_value_metadata.h"
```
pitrou, line 37: This shouldn't be necessary if `arrow/type_fwd.h` is included.

pitrou:
@jorisvandenbossche Do you want to take a look at the Python test changes?

cpp/src/arrow/dataset/file_base.h
westonpace: I removed this include and explicitly included `arrow/type_fwd.h`.

python/pyarrow/tests/test_dataset.py
```diff
@@ -5192,13 +5188,45 @@ def test_preserve_nullability_parquet(tempdir):
...
+    schema_no_meta = pa.schema([
+        pa.field("x", pa.int64()),
+        pa.field("y", pa.int64())])
+
+    array = [[1, 2, 3], [None, 5, None]]
+    table = pa.Table.from_arrays(array, schema=schema_metadata)
+    table_no_meta = pa.Table.from_arrays(array, schema=schema_no_meta)
+
+    # If no schema is provided the schema of the first table will be used
+    pa.dataset.write_dataset([table, table_no_meta],
```
anjakefala:
```
ds.write_dataset([table, table_no_meta],
```

python/pyarrow/tests/test_dataset.py
```diff
@@ -5192,13 +5188,45 @@ def test_preserve_nullability_parquet(tempdir):
...
+                             tempdir / "field_metatest", format="parquet")
+    dataset = ds.dataset(tempdir / "field_metatest", format="parquet")
+    assert dataset.to_table().schema.equals(schema_metadata, check_metadata=True)
+
+    pa.dataset.write_dataset([table_no_meta, table],
+                             tempdir / "field_metatest2", format="parquet")
+    dataset = ds.dataset(tempdir / "field_metatest2", format="parquet")
+    assert not dataset.to_table().schema.equals(schema_metadata, check_metadata=True)
+
+    pa.dataset.write_dataset([table_no_meta, table],
```
anjakefala:
```
ds.write_dataset([table_no_meta, table],
```

anjakefala (approved):
I like the python test changes!

Weston took out the invalid test case, and added one that covers how the field metadata gets applied.

jorisvandenbossche:
> Do you want to take a look at the Python test changes?

There were no changes to this since my previous review (and push), so all good!

westonpace:
I'm going to go ahead and merge if green.  I think it's quite late for Joris and I believe Raul will be starting the patch build soon.  If Joris finds an issue later then we can still scrap the patch release candidate for a fix (though hopefully we don't need to).

jorisvandenbossche (approved):

raulcd:
@github-actions crossbow submit test-r-ubuntu-22.04

PR #35864, opened 2023-06-01
GH-35747: Add FixedSizeListArray.to_numpy_ndarray and FixedSizeListArray.from_numpy_ndarray

### LINKED ISSUE #35747: [Python] Add from_numpy_ndarray and to_numpy_ndarray to ListArray types

Interoperation between numpy `ndarray`s and Arrow's ListArray types (ListArray, LargeListArray, FixedSizeListArray) is a bit tricky. 

It's hard to construct values: one must convert to a Python list-of-lists first, which is unnecessarily expensive:

```py
>>> import numpy as np
>>> import pyarrow as pa
>>> np_values = np.ones((3, 2), np.float64())
>>> pa_dtype = pa.list_(pa.float64())
>>> pa_values= pa.array(np_values, type=pa_dtype)
  File "pyarrow/array.pxi", line 323, in pyarrow.lib.array
  File "pyarrow/array.pxi", line 83, in pyarrow.lib._ndarray_to_array
pyarrow.lib.ArrowInvalid: only handle 1-dimensional arrays
>>> pa_values = pa.array(np_values.tolist(), type=pa_dtype)
<pyarrow.lib.ListArray object at 0x11ba433a0>
[
  [
    1,
    1
  ],
  [
    1,
    1
  ],
  [
    1,
    1
  ]
]
```

Likewise, converting to a numpy ndarray from a Pyarrow ListArray type is tricky, as described in https://github.com/apache/arrow/issues/35622. That issue describes trickiness with FixedSizeListArrays, but the same is true of ListArrays, which often _might_ have equal-length lists in every entry, making them amenable to presentation as an ndarray.

I'd like to propose the following 6 new methods:
 - `FixedSizeListArray.from_numpy_ndarray(values, type)`: 
   Constructs a new FixedSizeListArray from `values`, which must be a numpy ndarray with `ndim == 2`. 
   `type` is optional; it will be looked up from the ndarray's `dtype` if unset. 
   If `type` is set, values of the ndarray's dtype must be convertible to the provided type.
   
 - `FixedSizeListArray.to_numpy_ndarray(self)`: 
   Returns the FixedSizeListArray's values as a numpy ndarray with a shape of `(len(self), self.type.list_size)`. 
 
   If any of the FixedSizeListArray's values are `null`, raises an error.
   
   If any of the FixedSizeListArray's values _contain_ a `null`, then returns a `ndarray` with `nan` in the null spots, and with `dtype` set to `float64`, or `None` in the null spots and `dtype` of `object` if a conversion to `float64` is not possible. This matches the behavior of `Array.to_numpy` for primitive types.
   
- `ListArray.from_numpy_ndarray(values, type)`: 
  Works just like `FixedSizeListArray.from_numpy_ndarray`.
- `ListArray.to_numpy_ndarray(self)`: 
  Works like `FixedSizeListArray.to_numpy_ndarray`, with an additional check that all list elements are of equal length. If any are different, then raises an error.

and same for LargeListArray as for ListArray, bringing the total to 6.

The FixedSizeListArray methods already have an implementation in the `FixedShapeTensor` extension type. Those implementation are actually a bit more complicated because of tensors' support for permutations:
https://github.com/apache/arrow/blob/95c33d82e67c9500b08ea7086a46b447d65bc14b/python/pyarrow/array.pxi#L3149
https://github.com/apache/arrow/blob/95c33d82e67c9500b08ea7086a46b447d65bc14b/python/pyarrow/array.pxi#L3164

### Component(s)

Python

spenczar:

#35747 asks for additions to all three ListArray types (ListArray, FixedSizeListArray, and LargeListArray). This PR only adds for FixedSizeListArray. That's just because it's both the easiest to add, and the most obviously useful scenario.

This implementation is just in Python and follows the general pattern of FixedShapeTensorArrays.

An alternative implementation would be to do this in [python/pyarrow/src/arrow/python/numpy_convert.cc](https://github.com/apache/arrow/blob/cbc17a98dfd9fb660523f61ecd361b157d278372/python/pyarrow/src/arrow/python/numpy_convert.cc). We could detect 2D Numpy arrays and convert them to FixedShapeTensorArrays. That would be a much more significant change; I think it would change the behavior of other calls like `pyarrow.array()`. It would also allow for much more careful management of copies and memory; it should be possible to implement both to_ and from_ with zero copy for primitive numeric types. However, it would be a lot more complex, and this is my first Arrow contribution, so I figured I'd start a bit smaller.

### Are these changes tested?

Yes.

Yes, and I think I added sufficient documentation.
* Closes: #35747

### DIFF
--- a/python/pyarrow/array.pxi
+++ b/python/pyarrow/array.pxi
@@ -2294,6 +2294,46 @@ cdef class FixedSizeListArray(BaseListArray):
         result.validate()
         return result
 
+    @staticmethod
+    def from_numpy_ndarray(obj):
+        """
+        Convert a 2D numpy ndarray to a fixed size list array.
+
+        The first dimension of the ndarray is the array length, and the second
+        dimension is the list size.
+
+        The numpy array needs to be C-contiguous.
+
+        Parameters
+        ----------
+        obj : numpy.ndarray
+            The numpy array to convert.
+
+        Returns
+        -------
+        FixedSizeListArray
+        """
+        if not obj.flags["C_CONTIGUOUS"]:
+            raise ValueError("The data in the numpy array needs to be in a single, "
+                             "C-style contiguous segment.")
+
+        if len(obj.shape) != 2:
+            raise NotImplementedError("Only 2D numpy arrays are supported")
+
+        list_size = obj.shape[1]
+        if list_size <= 0:
+            raise ValueError("The list size needs to be positive")
+        array = np.ravel(obj, order="C")
+
+        return FixedSizeListArray.from_arrays(array, list_size=list_size)
+
+    def to_numpy_ndarray(self):
+        """
+        Output self as a 2D numpy ndarray.
+        """
+        flat = np.asarray(self.values)
+        return flat.reshape((len(self), self.type.list_size))
+
     @property
     def values(self):
         cdef CFixedSizeListArray* arr = <CFixedSizeListArray*> self.ap
--- a/python/pyarrow/tests/test_array.py
+++ b/python/pyarrow/tests/test_array.py
@@ -2825,6 +2825,61 @@ def test_fixed_size_list_array_flatten_with_slice():
     assert array[2:].flatten() == pa.array([3], type=pa.float64())
 
 
+def test_fixed_size_list_array_from_numpy_ndarray():
+    values = np.array([[1, 2], [3, 4], [5, 6]], dtype=np.int64)
+    result = pa.FixedSizeListArray.from_numpy_ndarray(values)
+    assert result.to_pylist() == [[1, 2], [3, 4], [5, 6]]
+    assert result.type == pa.list_(pa.int64(), 2)
+
+    # String-types should work
+    values = np.array([["a", "b"], ["c", "d"], ["e", "f"]], dtype=np.object_)
+    result = pa.FixedSizeListArray.from_numpy_ndarray(values)
+    assert result.to_pylist() == [["a", "b"], ["c", "d"], ["e", "f"]]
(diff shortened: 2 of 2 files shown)


AlenkaF:
Thank you @spenczar for making a contribution!
And sorry it took a while to get a review.

The implementation in Python instead of C++ makes sense. It might also make sense to think about the value of having this case (numpy arrays of 1 vs 2 dimensions, how one could still use 2-dim arrays in pyarrow avoiding an error) added to the documentation without adding two new functions to the code. I think that could be enough for a user needing to have an option for 2-dim numpy arrays.

rok:
Hey @spenczar, would you consider:
`FixedSizeListArray`:
- "casting" to `pa.FixedShapeTensorArray` and then using `to_numpy`
- `pa.FixedShapeTensorArray.from_numpy_ndarray` and then getting the fixed sized list type storage?
`ListArray/LargeListArray`:
- "casting" to `pa.VariableShapeTensor` and then using `to_numpy` (once https://github.com/apache/arrow/pull/40354 is merged)
- `pa.VariableShapeTensorArray.from_numpy_ndarray` anf then geting the list type storage?

This could work either currently as a workaround or as machinery for this PR.
In the latter case there's a minor caveat - `FixedShapeTensorArray` and `VariableShapeTensorArray` are available only if json is enabled ad compile time (`-DARROW_JSON=ON`).

spenczar:
Cleaning up old un-merged PRs in my queue. I still think this was a good idea but it appears to have no movement; I understand it is blocked by #40354.

rok:
I hope we cycle back to this once #40354 is merged. And sorry for the wait.

PR #35865, opened 2023-06-01
GH-35740: Add documentation for list arrays' values property

### LINKED ISSUE #35740: [Python][Docs] 'values' attribute is missing documentation for ListArray and its friends

The Python ListArray, LargeListArray, and FixedSizeListArray types all have a '.values' attribute which is undocumented, but useful. In #35622, @jorisvandenbossche pointed out that the `values` attribute can be used for assembling Numpy arrays without a copy, without falling all the way down to the buffer level.

### Component(s)

Documentation, Python

spenczar:
Just docs.

I'm a little shaky on my understanding of exactly what's going on with FixedSizeListArray.values, and its behavior with nulls, so that wording might deserve a careful read.
* Closes: #35740

### DIFF
--- a/python/pyarrow/array.pxi
+++ b/python/pyarrow/array.pxi
@@ -2057,8 +2057,6 @@ cdef class ListArray(BaseListArray):
         Return the underlying array of values which backs the ListArray
         ignoring the array's offset.
 
-        This is logically a concatenation of all the sub-lists in this array.
-
         If any of the list elements are null, but are backed by a
         non-empty sub-list, those elements will be included in the
         output.
@@ -2211,8 +2209,6 @@ cdef class LargeListArray(BaseListArray):
         """
         Return the underlying array of values which backs the LargeListArray
         ignoring the array's offset.
-
-        This is logically a concatenation of all the sub-lists in this array.
 
         If any of the list elements are null, but are backed by a
         non-empty sub-list, those elements will be included in the
@@ -2437,8 +2433,6 @@ cdef class FixedSizeListArray(BaseListArray):
         """
         Return the underlying array of values which backs the
         FixedSizeListArray.
-
-        This is logically a concatenation of all the sub-lists in this array.
 
         Note even null elements are included.
 
--- a/python/pyarrow/array.pxi
+++ b/python/pyarrow/array.pxi
@@ -1914,7 +1914,7 @@ cdef class BaseListArray(Array):
         The returned Array is logically a concatenation of all the sub-lists
         in this Array.
 
-        Note that this method is different from ``self.values()`` in that
+        Note that this method is different from ``self.values`` in that
         it takes care of the slicing offset as well as null elements backed
         by non-empty sub-lists.
 
@@ -2053,6 +2053,38 @@ cdef class ListArray(BaseListArray):
 
     @property
     def values(self):
+        """
+        Return the underlying array of values which backs the ListArray.
+
+        This is logically a concatenation of all the sub-lists in this array.
+
+        Note even null values are included. If any of the list slots
+        are null, but are backed by a non-empty sub-list, those values
+        will be included in the output.
+
+        Compare with ``flatten``, which returns only the non-null
+        values.
+
+        Returns
+        -------
+        values : Array
+
+        Examples
+        --------
+        >>> import pyarrow as pa
+        >>> array = pa.array([[1, 2], None, [3, 4, None, 6]])
+        >>> array.values
+        <pyarrow.lib.Int64Array object at ...>
+        [
+          1,
+          2,
+          3,
+          4,
+          null,
+          6
+        ]
+
+        """
         cdef CListArray* arr = <CListArray*> self.ap
         return pyarrow_wrap_array(arr.values())
 
@@ -2140,6 +2172,40 @@ cdef class LargeListArray(BaseListArray):
 
     @property
     def values(self):
+        """
+        Return the underlying array of values which backs the LargeListArray.
+
+        This is logically a concatenation of all the sub-lists in this array.
+
+        Note even null values are included. If any of the list slots
+        are null, but are backed by a non-empty sub-list, those values
+        will be included in the output.
+
+        Compare with ``flatten``, which returns only the non-null
+        values.
+
+        Returns
+        -------
+        values : Array
+
+        Examples
+        --------
+        >>> import pyarrow as pa
+        >>> array = pa.array(
+        ...     [[1, 2], None, [3, 4, None, 6]],
+        ...     type=pa.large_list(pa.int32()),
+        ... )
+        >>> array.values
+        <pyarrow.lib.Int32Array object at 0x12d174100>
+        [
+          1,
+          2,
+          3,
+          4,
+          null,
+          6
+        ]
+        """
         cdef CLargeListArray* arr = <CLargeListArray*> self.ap
         return pyarrow_wrap_array(arr.values())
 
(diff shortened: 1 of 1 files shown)


AlenkaF:
Thank you for the contribution @spenczar !

I think the main difference between `values` and `flatten` is when the array is a slice into another array (see description in https://arrow.apache.org/docs/dev/cpp/api/array.html#_CPPv4N5arrow18FixedSizeListArrayE). You could play around with looking at the code for `array = pa.array([[1, 2], None, [3, None]],type=pa.large_list(pa.int32()))`and `array.slice(0, 1)`.

You are correct in the understanding that the behaviour of `FixedSizeListArray.values` is a bit different. I am now aware of the reason for this, but the difference can be seen in the link to the docs from before (https://arrow.apache.org/docs/dev/cpp/api/array.html#_CPPv4N5arrow9ListArrayE).

Some nit for finish: there are doctest errors that need to be fixed with adding `...`:
```
___________ [doctest] pyarrow.lib.FixedSizeListArray.values.__get__ ____________
2378         values : Array
2379 
2380         Examples
2381         --------
2382         >>> import pyarrow as pa
2383         >>> array = pa.array(
2384         ...     [[1, 2], None, [3, None]],
2385         ...     type=pa.list_(pa.int32(), 2)
2386         ... )
2387         >>> array.values
Differences (unified diff with -expected +actual):
    @@ -1,3 +1,3 @@
    -<pyarrow.lib.Int32Array object at 0x12d1743a0>
    +<pyarrow.lib.Int32Array object at 0x7f3828fa5440>
     [
       1,

/opt/conda/envs/arrow/lib/python3.9/site-packages/pyarrow/lib.cpython-39-x86_64-linux-gnu.so:2387: DocTestFailure
_____________ [doctest] pyarrow.lib.LargeListArray.values.__get__ ______________
2189         values : Array
2190 
2191         Examples
2192         --------
2193         >>> import pyarrow as pa
2194         >>> array = pa.array(
2195         ...     [[1, 2], None, [3, 4, None, 6]],
2196         ...     type=pa.large_list(pa.int32()),
2197         ... )
2198         >>> array.values
Differences (unified diff with -expected +actual):
    @@ -1,3 +1,3 @@
    -<pyarrow.lib.Int32Array object at 0x12d174100>
    +<pyarrow.lib.Int32Array object at 0x7f3828faeb40>
     [
       1,
```

spenczar:
Thanks for the explanation @AlenkaF, that makes a lot of sense, and I learned a bunch about Arrow internals from playing with `slice`. Do you think the docs I wrote need any edits?

I think I've fixed the doctest issue, too, we'll see if it passes CI now.

python/pyarrow/array.pxi
```diff
@@ -2053,6 +2053,38 @@ cdef class ListArray(BaseListArray):
 
     @property
     def values(self):
+        """
+        Return the underlying array of values which backs the ListArray.
```
AlenkaF, line 2057:
```suggestion
        Return the underlying array of values which backs the ListArray
        ignoring array’s offsets.
```
Same for LargeListArray ...

python/pyarrow/array.pxi
```diff
@@ -2053,6 +2053,38 @@ cdef class ListArray(BaseListArray):
...
+
+        This is logically a concatenation of all the sub-lists in this array.
+
+        Note even null values are included. If any of the list slots
+        are null, but are backed by a non-empty sub-list, those values
+        will be included in the output.
+
+        Compare with ``flatten``, which returns only the non-null
+        values.
```
AlenkaF, line 2066:
```suggestion
        values taking into consideration array’s offsets.
```
Same for LargeListArray ...

python/pyarrow/array.pxi
```diff
@@ -2296,6 +2362,40 @@ cdef class FixedSizeListArray(BaseListArray):
 
     @property
     def values(self):
+        """
+        Return the underlying array of values which backs the
+        FixedSizeListArray.
+
+        This is logically a concatenation of all the sub-lists in this array.
+
+        Note even null values are included.
```
AlenkaF, line 2371:
```suggestion
        Note even null elements are included.
```

python/pyarrow/array.pxi
```diff
@@ -2140,6 +2172,40 @@ cdef class LargeListArray(BaseListArray):
...
     @property
     def values(self):
+        """
+        Return the underlying array of values which backs the LargeListArray.
+
+        This is logically a concatenation of all the sub-lists in this array.
+
+        Note even null values are included. If any of the list slots
+        are null, but are backed by a non-empty sub-list, those values
+        will be included in the output.
```
AlenkaF, line 2182:
```suggestion
        If any of the list elements are null, but backed by a non-empty sub-list,
        those elements will be included in the output.
```
Same for ListArray ..

AlenkaF (commented):
Thanks for the corrections!
Some minor nits ...

python/pyarrow/array.pxi
```diff
@@ -2053,6 +2053,38 @@ cdef class ListArray(BaseListArray):
...
+        """
+        Return the underlying array of values which backs the ListArray.
+
+        This is logically a concatenation of all the sub-lists in this array.
+
+        Note even null values are included. If any of the list slots
+        are null, but are backed by a non-empty sub-list, those values
+        will be included in the output.
+
+        Compare with ``flatten``, which returns only the non-null
```
jorisvandenbossche, line 2065:
```suggestion
        Compare with :meth:`flatten`, which returns only the non-null
```

(the the online docs will link to it)

python/pyarrow/array.pxi
```diff
@@ -2053,6 +2053,38 @@ cdef class ListArray(BaseListArray):
...
+        <pyarrow.lib.Int64Array object at ...>
+        [
+          1,
+          2,
+          3,
+          4,
+          null,
+          6
+        ]
+
```
jorisvandenbossche, line 2086: Adding one more example of a sliced array might illustrate the difference between `values` and `.flatten()`

jorisvandenbossche (commented):
Thanks for the PR @spenczar! 
I added a few more comments. I think if we expand it even more (eg additional example), we can further clarify the difference between values and flatten(). But note that what you have here is already a clear improvement, so it's certainly fine as well to just focus on finalizing the edits you have now in this PR.

python/pyarrow/array.pxi
```diff
@@ -2053,6 +2053,38 @@ cdef class ListArray(BaseListArray):
 
     @property
     def values(self):
+        """
+        Return the underlying array of values which backs the ListArray.
+
+        This is logically a concatenation of all the sub-lists in this array.
```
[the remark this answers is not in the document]
spenczar: Personally, I find it useful to know the "dimensionality" of this object - that it's a flat structure, and it feels like a concatenation of all the sub-lists. What do you think of taking out the "logically" qualifier?

Or we could just drop it. The example is pretty clear.

python/pyarrow/array.pxi
spenczar: This is a major improvement - thanks for the suggestion!

spenczar:
Hello! Anything else to be done here? This is affecting some intersphinx stuff I'm doing in a mild way.

spenczar:
Is there anything I can do to get this merged?

python/pyarrow/array.pxi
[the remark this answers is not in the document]
AlenkaF: Does the previous sentence, "Return the underlying array of values ..." provide enough information about the "dimensionality" of the object? If yes, we could remove this second sentence as Joris suggests?

AlenkaF (approved):
Sorry for the delay @spenczar!

I would like us to address the unresolved comment before merging. Otherwise the PR looks good 👍

python/pyarrow/array.pxi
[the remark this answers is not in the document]
spenczar: Sure, I've removed the second sentence.

PR #35867, opened 2023-06-01
GH-35866: [Go] Provide a copy in `arrow.NestedType.Fields()` implementations

### LINKED ISSUE #35866: [Go] `arrow.StructType` fields are mutable

After `arrow.StructType` was constructed the fields are considered immutable (lice columns in `arrow.Schema`, but they aren't

### Component(s)

Go

candiduslynx:
### Rationale for this change

Data types should be left immutable after they've been constructed.

Provide a copy of fields in the following implementations:
* `*arrow.StructType.Fields()`
* `*arrow.unionType.Fields()`

### Are these changes tested?

`arrow.TestFieldsImmutability` was added to ensure the behavior.

Now the fields for structs & unions will be immutable.

* Closes: #35866

### DIFF
--- a/go/arrow/datatype_nested.go
+++ b/go/arrow/datatype_nested.go
@@ -27,6 +27,9 @@ import (
 
 type NestedType interface {
 	DataType
+
+	// Fields method provides a copy of NestedType fields
+	// (so it can be safely mutated and will not result in updating the NestedType).
 	Fields() []Field
 }
 
@@ -288,6 +291,8 @@ func (t *StructType) String() string {
 	return o.String()
 }
 
+// Fields method provides a copy of StructType fields
+// (so it can be safely mutated and will not result in updating the StructType).
 func (t *StructType) Fields() []Field {
 	fields := make([]Field, len(t.fields))
 	copy(fields, t.fields)
@@ -495,6 +500,8 @@ func (t *unionType) init(fields []Field, typeCodes []UnionTypeCode) {
 	}
 }
 
+// Fields method provides a copy of union type fields
+// (so it can be safely mutated and will not result in updating the union type).
 func (t unionType) Fields() []Field {
 	fields := make([]Field, len(t.children))
 	copy(fields, t.children)
--- a/go/arrow/datatype_nested.go
+++ b/go/arrow/datatype_nested.go
@@ -288,7 +288,12 @@ func (t *StructType) String() string {
 	return o.String()
 }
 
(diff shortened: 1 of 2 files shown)


go/arrow/datatype_nested.go
```diff
@@ -288,7 +288,12 @@ func (t *StructType) String() string {
 	return o.String()
 }
 
-func (t *StructType) Fields() []Field   { return t.fields }
+func (t *StructType) Fields() []Field {
```
zeroshade, line 291: Can you add a godoc comment on the methods to indicate to callers that this will make a copy of the Fields?

go/arrow/datatype_nested.go
```diff
@@ -490,7 +495,12 @@ func (t *unionType) init(fields []Field, typeCodes []UnionTypeCode) {
 	}
 }
 
-func (t unionType) Fields() []Field            { return t.children }
+func (t unionType) Fields() []Field {
+	fields := make([]Field, len(t.children))
+	copy(fields, t.children)
+	return fields
+}
```
zeroshade, line 502: same here. add a godoc comment to this to indicate it makes a copy of the fields

zeroshade (changes requested):
LGTM just want to have the documentation indicate that it makes a copy so that callers can choose to instead use the `Field(i)` methods if they want to avoid the copies

candiduslynx:
> LGTM just want to have the documentation indicate that it makes a copy so that callers can choose to instead use the `Field(i)` methods if they want to avoid the copies

Just FYI, `Field(i)` provides a copy of the `arrow.Field` (note: the `Type` field will point to the same instance, but the pointer itself is copied)

go/arrow/datatype_nested.go
candiduslynx: 1e81c723eb665da8f59519710c749e0fda435676

go/arrow/datatype_nested.go
candiduslynx: 1e81c723eb665da8f59519710c749e0fda435676

zeroshade:
@candiduslynx true, (a mistake in my wording) i meant if they didn't want to make a copy of the entire slice of fields. Copying an individual field is extremely cheap, but under certain circumstances it could be desireable to avoid copying the entire slice of fields. 

Either way most times that a slice is returned from a method in go, the average usage assumes that it isn't making a copy of all the elements so it's good to ensure the godoc is clear.

zeroshade (approved):

PR #34392, opened 2023-03-01
GH-34391: [C++] Future as-of-join-node hangs on distant times

rtpsw:
See #34391
* Closes: #34391

Note that the `TestBasic7Forward` test-case included in the PR reproduces the hang in the pre-PR code.

PR #35874, opened 2023-06-01
GH-35838: [C++] Add backpressure test for asof join node

### LINKED ISSUE #35838: [C++] Backpressure broken in asof join node

Originally the `InputState` was created with...

```
  static Result<std::unique_ptr<InputState>> Make(
      size_t index, TolType tolerance, bool must_hash, bool may_rehash,
      KeyHasher* key_hasher, ExecNode* node, ExecNode* output,
      std::atomic<int32_t>& backpressure_counter,
      const std::shared_ptr<arrow::Schema>& schema, const col_index_t time_col_index,
      const std::vector<col_index_t>& key_col_index) {
    constexpr size_t low_threshold = 4, high_threshold = 8;
    std::unique_ptr<BackpressureControl> backpressure_control =
        std::make_unique<BackpressureController>(node, output, backpressure_counter);
...

          InputState::Make(i, tolerance_, must_hash_, may_rehash_, key_hashers_[i].get(),
                           inputs[i], this, backpressure_counter_,
                           inputs[i]->output_schema(), indices_of_on_key_[i],
                           indices_of_by_key_[i]));
```

The `BackpressureController` is given `node` (which is `inputs[i]`) and `output` (which is `this`).

After https://github.com/apache/arrow/commit/dcdeab76b4ba4bac0b4115925c820961ef70ae0d the code changed to:

```
  static Result<std::unique_ptr<InputState>> Make(
      size_t index, TolType tolerance, bool must_hash, bool may_rehash,
      KeyHasher* key_hasher, AsofJoinNode* node, ExecNode* output,
      std::atomic<int32_t>& backpressure_counter,
      const std::shared_ptr<arrow::Schema>& schema, const col_index_t time_col_index,
      const std::vector<col_index_t>& key_col_index) {
    constexpr size_t low_threshold = 4, high_threshold = 8;
    std::unique_ptr<BackpressureControl> backpressure_control =
        std::make_unique<BackpressureController>(node, output, backpressure_counter);

...

          InputState::Make(i, tolerance_, must_hash_, may_rehash_, key_hashers_[i].get(),
                           this, inputs[i], backpressure_counter_,
                           inputs[i]->output_schema(), indices_of_on_key_[i],
                           indices_of_by_key_[i]));
```

Note that `BackpressureControl` is now created with `node` (which is `this`) and then `output` (which is `inputs[i]`).  This means, when the asof join node decides to pause it will call `PauseProducing` on `AsofJoinNode` which currently does nothing:

```
  void PauseProducing(ExecNode* output, int32_t counter) override {}
  void ResumeProducing(ExecNode* output, int32_t counter) override {}
```

### Component(s)

C++

rtpsw:

Passing the correct nodes to the backpressure controller, along with better parameter naming/doc. Also added reusable gate-classes (`Gate`, `GatedNodeOptions` and `GatedNode`) that enable holding all input batches until a gate is released, in order to support more robust backpressure testing in this PR.

### Are these changes tested?

Yes.

No.

* Closes: #35838

### DIFF
--- a/cpp/src/arrow/acero/asof_join_node.cc
+++ b/cpp/src/arrow/acero/asof_join_node.cc
@@ -529,7 +529,7 @@ class KeyHasher {
   size_t index_;
   std::vector<col_index_t> indices_;
   std::vector<KeyColumnMetadata> metadata_;
-  const RecordBatch* batch_;
+  std::atomic<const RecordBatch*> batch_;
   std::vector<HashType> hashes_;
   LightContext ctx_;
   std::vector<KeyColumnArray> column_arrays_;
--- a/cpp/src/arrow/acero/asof_join_node.cc
+++ b/cpp/src/arrow/acero/asof_join_node.cc
@@ -370,15 +370,10 @@ struct MemoStore {
     times_.swap(memo.times_);
   }
 
-  // Updates the current time to `ts` if it is less. A different thread may win the race
-  // to update the current time to more than `ts` but not to less. Returns whether the
-  // current time was changed from its value at the beginning of this invocation.
+  // Updates the current time to `ts` if it is less. Returns true if updated.
   bool UpdateTime(OnType ts) {
-    OnType prev_time = current_time_;
-    bool update = prev_time < ts;
-    while (prev_time < ts && !current_time_.compare_exchange_weak(prev_time, ts)) {
-      // intentionally empty - standard CAS loop
-    }
+    bool update = current_time_ < ts;
+    if (update) current_time_ = ts;
     return update;
   }
 
@@ -822,8 +817,11 @@ class InputState {
         ++batches_processed_;
         latest_ref_row_ = 0;
         have_active_batch &= !queue_.TryPop();
-        if (have_active_batch)
+        if (have_active_batch) {
           DCHECK_GT(queue_.UnsyncFront()->num_rows(), 0);  // empty batches disallowed
+          key_hasher_->Invalidate();  // batch changed - invalidate key hasher's cache
+          memo_.UpdateTime(GetTime(queue_.UnsyncFront().get(), 0));  // time changed
+        }
       }
     }
     return have_active_batch;
@@ -899,8 +897,6 @@ class InputState {
 
   Status Push(const std::shared_ptr<arrow::RecordBatch>& rb) {
     if (rb->num_rows() > 0) {
-      key_hasher_->Invalidate();  // batch changed - invalidate key hasher's cache
-      memo_.UpdateTime(GetTime(rb.get(), 0));  // time changed - update in MemoStore
       queue_.Push(rb);  // only after above updates - push batch for processing
     } else {
       ++batches_processed_;  // don't enqueue empty batches, just record as processed
--- a/cpp/src/arrow/acero/asof_join_node.cc
+++ b/cpp/src/arrow/acero/asof_join_node.cc
@@ -495,9 +495,9 @@ class KeyHasher {
                        4 * kMiniBatchLength * sizeof(uint32_t));
   }
 
-  void Invalidate() {
-    batch_ = NULLPTR;  // invalidate cached hashes for batch - required when it changes
-  }
+  // invalidate cached hashes for batch - required when it changes
+  // only this method can be called concurrently with HashesFor
+  void Invalidate() { batch_ = NULLPTR; }
 
   // compute and cache a hash for each row of the given batch
   const std::vector<HashType>& HashesFor(const RecordBatch* batch) {
--- a/cpp/src/arrow/acero/asof_join_node.cc
+++ b/cpp/src/arrow/acero/asof_join_node.cc
@@ -370,10 +370,15 @@ struct MemoStore {
     times_.swap(memo.times_);
   }
 
-  // Updates the current time to `ts` if it is less. Returns true if updated.
+  // Updates the current time to `ts` if it is less. A different thread may win the race
+  // to update the current time to more than `ts` but not to less. Returns whether the
+  // current time was changed from its value at the beginning of this invocation.
   bool UpdateTime(OnType ts) {
-    bool update = current_time_ < ts;
-    if (update) current_time_ = ts;
+    OnType prev_time = current_time_;
+    bool update = prev_time < ts;
+    while (prev_time < ts && !current_time_.compare_exchange_weak(prev_time, ts)) {
+      // intentionally empty - standard CAS loop
+    }
     return update;
   }
 
@@ -817,11 +822,8 @@ class InputState {
         ++batches_processed_;
         latest_ref_row_ = 0;
         have_active_batch &= !queue_.TryPop();
-        if (have_active_batch) {
+        if (have_active_batch)
           DCHECK_GT(queue_.UnsyncFront()->num_rows(), 0);  // empty batches disallowed
-          key_hasher_->Invalidate();  // batch changed - invalidate key hasher's cache
-          memo_.UpdateTime(GetTime(queue_.UnsyncFront().get(), 0));  // time changed
-        }
       }
     }
     return have_active_batch;
@@ -897,6 +899,8 @@ class InputState {
 
   Status Push(const std::shared_ptr<arrow::RecordBatch>& rb) {
     if (rb->num_rows() > 0) {
+      key_hasher_->Invalidate();  // batch changed - invalidate key hasher's cache
+      memo_.UpdateTime(GetTime(rb.get(), 0));  // time changed - update in MemoStore
       queue_.Push(rb);  // only after above updates - push batch for processing
     } else {
       ++batches_processed_;  // don't enqueue empty batches, just record as processed
--- a/cpp/src/arrow/acero/asof_join_node.cc
+++ b/cpp/src/arrow/acero/asof_join_node.cc
@@ -668,18 +668,19 @@ class InputState {
 
   static Result<std::unique_ptr<InputState>> Make(
       size_t index, TolType tolerance, bool must_hash, bool may_rehash,
-      KeyHasher* key_hasher, ExecNode* node, AsofJoinNode* output,
+      KeyHasher* key_hasher, ExecNode* input, AsofJoinNode* node,
       std::atomic<int32_t>& backpressure_counter,
       const std::shared_ptr<arrow::Schema>& schema, const col_index_t time_col_index,
       const std::vector<col_index_t>& key_col_index) {
     constexpr size_t low_threshold = 4, high_threshold = 8;
     std::unique_ptr<BackpressureControl> backpressure_control =
-        std::make_unique<BackpressureController>(node, output, backpressure_counter);
+        std::make_unique<BackpressureController>(/*node=*/input, /*output=*/node,
+                                                 backpressure_counter);
     ARROW_ASSIGN_OR_RAISE(auto handler,
                           BackpressureHandler::Make(low_threshold, high_threshold,
                                                     std::move(backpressure_control)));
     return std::make_unique<InputState>(index, tolerance, must_hash, may_rehash,
-                                        key_hasher, output, std::move(handler), schema,
+                                        key_hasher, node, std::move(handler), schema,
                                         time_col_index, key_col_index);
   }
 
--- a/cpp/src/arrow/acero/asof_join_node.cc
+++ b/cpp/src/arrow/acero/asof_join_node.cc
@@ -668,19 +668,19 @@ class InputState {
 
   static Result<std::unique_ptr<InputState>> Make(
       size_t index, TolType tolerance, bool must_hash, bool may_rehash,
-      KeyHasher* key_hasher, ExecNode* input, AsofJoinNode* node,
+      KeyHasher* key_hasher, ExecNode* asof_input, AsofJoinNode* asof_node,
       std::atomic<int32_t>& backpressure_counter,
       const std::shared_ptr<arrow::Schema>& schema, const col_index_t time_col_index,
       const std::vector<col_index_t>& key_col_index) {
     constexpr size_t low_threshold = 4, high_threshold = 8;
     std::unique_ptr<BackpressureControl> backpressure_control =
-        std::make_unique<BackpressureController>(/*node=*/input, /*output=*/node,
-                                                 backpressure_counter);
+        std::make_unique<BackpressureController>(
+            /*node=*/asof_input, /*output=*/asof_node, backpressure_counter);
     ARROW_ASSIGN_OR_RAISE(auto handler,
                           BackpressureHandler::Make(low_threshold, high_threshold,
                                                     std::move(backpressure_control)));
     return std::make_unique<InputState>(index, tolerance, must_hash, may_rehash,
-                                        key_hasher, node, std::move(handler), schema,
+                                        key_hasher, asof_node, std::move(handler), schema,
                                         time_col_index, key_col_index);
   }
 
--- a/cpp/src/arrow/acero/asof_join_node_test.cc
+++ b/cpp/src/arrow/acero/asof_join_node_test.cc
@@ -1362,8 +1362,8 @@ TRACED_TEST(AsofJoinTest, TestUnorderedOnKey, {
 })
 
 struct BackpressureCounters {
-  int32_t pause_count = 0;
-  int32_t resume_count = 0;
+  std::atomic<int32_t> pause_count = 0;
+  std::atomic<int32_t> resume_count = 0;
 };
 
 struct BackpressureCountingNodeOptions : public ExecNodeOptions {
--- a/cpp/src/arrow/acero/asof_join_node_test.cc
+++ b/cpp/src/arrow/acero/asof_join_node_test.cc
@@ -31,6 +31,7 @@
 #ifndef NDEBUG
 #include "arrow/acero/options_internal.h"
 #endif
+#include "arrow/acero/map_node.h"
 #include "arrow/acero/test_nodes.h"
 #include "arrow/acero/test_util_internal.h"
 #include "arrow/acero/util.h"
@@ -1381,18 +1382,88 @@ void TestBackpressure(BatchesMaker maker, int num_batches, int batch_size,
   ASSERT_OK_AND_ASSIGN(auto r0_batches, make_shift(r0_schema, 1));
   ASSERT_OK_AND_ASSIGN(auto r1_batches, make_shift(r1_schema, 2));
 
-  Declaration l_src = {
-      "source", SourceNodeOptions(
-                    l_schema, MakeDelayedGen(l_batches, "0:fast", fast_delay, noisy))};
-  Declaration r0_src = {
-      "source", SourceNodeOptions(
-                    r0_schema, MakeDelayedGen(r0_batches, "1:slow", slow_delay, noisy))};
-  Declaration r1_src = {
-      "source", SourceNodeOptions(
-                    r1_schema, MakeDelayedGen(r1_batches, "2:fast", fast_delay, noisy))};
+  struct BackpressureCounters {
+    int32_t pause_count = 0;
+    int32_t resume_count = 0;
+  };
+
+  struct BackpressureTestNodeOptions : public ExecNodeOptions {
+    BackpressureCounters* counters;
+  };
+
+  struct BackpressureTestNode : public MapNode {
+    BackpressureTestNode(ExecPlan* plan, std::vector<ExecNode*> inputs,
+                         std::shared_ptr<Schema> output_schema,
+                         const BackpressureTestNodeOptions& options)
+        : MapNode(plan, inputs, output_schema), counters(options.counters) {}
+
+    static Result<ExecNode*> Make(ExecPlan* plan, std::vector<ExecNode*> inputs,
+                                  const ExecNodeOptions& options) {
+      RETURN_NOT_OK(ValidateExecNodeInputs(plan, inputs, 1, "BackpressureTestNode"));
+      auto bp_options = static_cast<const BackpressureTestNodeOptions&>(options);
+      return plan->EmplaceNode<BackpressureTestNode>(
+          plan, inputs, inputs[0]->output_schema(), bp_options);
+    }
+
+    const char* kind_name() const override { return "BackpressureTestNode"; }
+    Result<ExecBatch> ProcessBatch(ExecBatch batch) override { return batch; }
+
+    void PauseProducing(ExecNode* output, int32_t counter) override {
+      ++counters->pause_count;
+      inputs()[0]->PauseProducing(this, counter);
+    }
+    void ResumeProducing(ExecNode* output, int32_t counter) override {
+      ++counters->resume_count;
+      inputs()[0]->ResumeProducing(this, counter);
+    }
+
+    BackpressureCounters* counters;
+  };
+
+  auto exec_reg = default_exec_factory_registry();
+  std::string bp_test = "backpressure_test";
+  if (!exec_reg->GetFactory(bp_test).ok()) {
+    ASSERT_OK(exec_reg->AddFactory(bp_test, BackpressureTestNode::Make));
+  }
+
+  struct SourceConfig {
+    std::string name_prefix;
+    bool is_fast;
+    std::shared_ptr<Schema> schema;
+    decltype(l_batches) batches;
+
+    std::string name() const { return name_prefix + ";" + (is_fast ? "fast" : "slow"); }
+  };
+
+  // must have at least one fast and one slow
+  std::vector<SourceConfig> source_configs = {
+      {"0", true, l_schema, l_batches},
+      {"1", false, r0_schema, r0_batches},
+      {"2", true, r1_schema, r1_batches},
+  };
+
+  std::vector<BackpressureCounters> bp_counters(source_configs.size());
+  std::vector<Declaration> src_decls;
+  std::vector<std::shared_ptr<BackpressureTestNodeOptions>> bp_options;
+  std::vector<Declaration::Input> bp_decls;
+  for (size_t i = 0; i < source_configs.size(); i++) {
+    const auto& config = source_configs[i];
+    src_decls.emplace_back(
+        "source", SourceNodeOptions(
+                      config.schema,
+                      MakeDelayedGen(config.batches, config.name(),
+                                     config.is_fast ? fast_delay : slow_delay, noisy)));
+    bp_options.push_back(std::make_shared<BackpressureTestNodeOptions>());
+    bp_options.back()->counters = &bp_counters[i];
+    std::shared_ptr<ExecNodeOptions> options = bp_options.back();
+    std::vector<Declaration::Input> bp_in = {src_decls.back()};
+    Declaration bp_decl = {bp_test, bp_in, std::move(options)};
+    bp_decls.push_back(bp_decl);
+  }
 
   Declaration asofjoin = {
-      "asofjoin", {l_src, r0_src, r1_src}, GetRepeatedOptions(3, "time", {"key"}, 1000)};
+      "asofjoin", bp_decls,
+      GetRepeatedOptions(source_configs.size(), "time", {"key"}, 1000)};
 
   ASSERT_OK_AND_ASSIGN(std::unique_ptr<RecordBatchReader> batch_reader,
                        DeclarationToReader(asofjoin, /*use_threads=*/false));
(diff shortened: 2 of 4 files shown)


icexelloss:
@westonpace Are you OK with me merging this to unblock internal issue? @rtpsw Is working on adding tests now but it might take a while.

cpp/src/arrow/acero/asof_join_node.cc
```diff
@@ -668,13 +668,13 @@ class InputState {
 
   static Result<std::unique_ptr<InputState>> Make(
       size_t index, TolType tolerance, bool must_hash, bool may_rehash,
-      KeyHasher* key_hasher, AsofJoinNode* node, ExecNode* output,
+      KeyHasher* key_hasher, AsofJoinNode* node, ExecNode* input,
```
icexelloss, line 671: This still looking confusing:

(1) BackpressureController takes `ExecNode* node, ExecNode* output` and this one nows takes `AsofJoinNode* node, ExecNode* input` which is inconsistent

(2) InputState::Make uses to take inputs[i] but now it takes "node" which is the AsofJoinNode

icexelloss: It is also confusing that 

```
    return std::make_unique<InputState>(index, tolerance, must_hash, may_rehash,
                                        key_hasher, node, std::move(handler), schema,
                                        time_col_index, key_col_index);
```
On line 681 passes the asof join node to the input state instead of the input node, why is it?

icexelloss: @rtpsw I think it is easier to revert this line as to unblock the internal first:
https://github.com/apache/arrow/pull/34392/files#diff-5493b6ae7ea2a4d5cfb581034c076e9c4be7608382168de6d1301ef67b6c01eeR1410

Then work on cleaning up changes introduced in GH-36391. The code is quite confusing now.

cpp/src/arrow/acero/asof_join_node.cc
```diff
@@ -668,13 +668,13 @@ class InputState {
...
       std::atomic<int32_t>& backpressure_counter,
       const std::shared_ptr<arrow::Schema>& schema, const col_index_t time_col_index,
       const std::vector<col_index_t>& key_col_index) {
     constexpr size_t low_threshold = 4, high_threshold = 8;
     std::unique_ptr<BackpressureControl> backpressure_control =
-        std::make_unique<BackpressureController>(node, output, backpressure_counter);
+        std::make_unique<BackpressureController>(input, node, backpressure_counter);
```
pitrou, line 677: Since there's been some confusion, can you make parameter names explicit?
```suggestion
        std::make_unique<BackpressureController>(/*xxx=*/ input, /*yyy=*/ node, backpressure_counter);
```

cpp/src/arrow/acero/asof_join_node.cc
rtpsw: > I think it is easier to revert this line

This reversion alone doesn't compile because `inputs[i]` is not of type `AsofJoinNode*`.

icexelloss: It's probably easier for me to do it instead of back and forth, opened:

https://github.com/apache/arrow/pull/35878

icexelloss: I am fine with figure out the cleaner way to do this in follow up to GH-36391. But for now I think it's easier to just revert the change.

icexelloss: > The two places have a different perspective of what is an input and what is an output (which likely caused confusion in the first place).

We should probably fix the variable naming in the follow PR to GH-36391 how to call these things then. But for now let's just revert to what was before.

cpp/src/arrow/acero/asof_join_node.cc
rtpsw: Done.

cpp/src/arrow/acero/asof_join_node_test.cc
```diff
@@ -1406,6 +1458,9 @@ void TestBackpressure(BatchesMaker maker, int num_batches, int batch_size,
     total_length += batch->num_rows();
   }
   ASSERT_EQ(static_cast<int64_t>(num_batches * batch_size), total_length);
+
+  ASSERT_GT(pause_count, 0);
```
icexelloss, line 1462: Can we validate pause resume counter for all sources? i.e the slow table should not have been paused, but the fast tables should?

cpp/src/arrow/acero/asof_join_node.cc
westonpace: Maybe use more specific names like `asof_input` and `asof_node`?

westonpace (commented):
I do have a little bit of concern that this test may become flaky on CI machines which have notoriously weird timing.  However, we can give it a few runs and see how it goes.  If it does become a problem I'd suggest we invent some kind of "gated source" so we can do something like (very rough pseudocode)...

```
GatedSourceNode gated_left = GatedSourceNode(l_batches);
GatedSourceNode gated_right = GatedSourceNode(r_batches);
gated_right.Ungate();

// Start plan

WaitFor([&] { right_bp_options->pause_count > 0; });
ASSERT_EQ(right_bp_options->resume_count, 0);
ASSERT_EQ(left_bp_options->pause_count, 0);
ASSERT_EQ(left_bp_options->resume_count, 0);

gated_left->Ungate();
AssertPlanFinishes();
```

I'm off today and Monday but I can sketch up a possible `GatedSourceNode` implementation when I get back if needed.

rtpsw:
> I do have a little bit of concern that this test may become flaky on CI machines

Indeed it is flaky - [this](https://github.com/apache/arrow/actions/runs/5155477352/jobs/9285205333?pr=35874) and [this](https://github.com/apache/arrow/actions/runs/5155477352/jobs/9285205732?pr=35874) and [this](https://ci.appveyor.com/project/ApacheSoftwareFoundation/arrow/builds/47206703) CI jobs have failed on the new check. I'll make a quick attempt to fix this.

I agree relying on timing is not good, though that can be said about the pre-PR backpressure test code too, and this PR won't be worse in this respect.

cpp/src/arrow/acero/asof_join_node.cc
rtpsw: Done.

cpp/src/arrow/acero/asof_join_node_test.cc
rtpsw: We may have crossed - the [previous commit](https://github.com/apache/arrow/pull/35874/commits/7d40e92c7d96e34935b2af3f84f74d4e1d8cec49) already has one pause and one resume counter for each source. The issue is that sometimes (due to non-deterministic timing of operations) no pause/resume is requested on one or both of the fast sources. Because of this, in the [recent commit](https://github.com/apache/arrow/pull/35874/commits/3e5fdf9a4a7e48500882f1592d1949c772dca358) I am attempting this test logic: there is a pause request on a fast source if and only if there is a resume one too.

icexelloss:
Thinking about loud here:

What we want to test is that if the through put of asof join node is slower than the source, then we would pause the source. Two potential ways that I think we can reliably do this:
(1) Add some sort of "debug options" to manipulate the behavior of asof join to make it run slower. (i.e. Sleep a few seconds before actually starting the work in the processing thread)
(2) Add a downstream node to asof join that processes data slowly (similar to a slow data sink), i.e., process one batch per second. This way, the backpressure would be pushed from the slow sink to asof join then to the data sources.

I think I prefer (2) a bit more because this affects represents a real life case of slow sink. 

@westonpace I am not sure if the idea of GatedSourceNode is similar or different, but happy to hear

cpp/src/arrow/acero/asof_join_node_test.cc
```diff
@@ -1406,6 +1477,20 @@ void TestBackpressure(BatchesMaker maker, int num_batches, int batch_size,
...
+    BackpressureCounters& counters = counters_by_is_fast[source_configs[i].is_fast];
+    counters.pause_count += bp_counters[i].pause_count;
+    counters.resume_count += bp_counters[i].resume_count;
+  }
+  ASSERT_EQ(counters_by_is_fast.size(), 2);
+  ASSERT_GT(counters_by_is_fast[true].pause_count, 0);
+  ASSERT_GT(counters_by_is_fast[true].resume_count, 0);
+  // runs on some slow machines may not see any pause/resume, but if at least one pause is
+  // seen then at least one resume must also be seen
+  ASSERT_EQ(counters_by_is_fast[false].pause_count > 0,
```
icexelloss, line 1492: I don't think we really need to check the slow sources, because it is entirely possible that it is not paused at all in the expected behavior (since it is slower than asof join?)

rtpsw: I think we do. My understanding is that the intention of the pre-PR test code was to drive the as-of-join slower using a slower input to it, which should lead to backpressure from the as-of-join node. The additional intention of the post-PR test code is to check that indeed this backpressure happens, and this is observed at the test nodes inserted after each source.

rtpsw:
> Thinking about loud here:
> 
> What we want to test is that if the through put of asof join node is slower than the source, then we would pause the source. Two potential ways that I think we can reliably do this: (1) Add some sort of "debug options" to manipulate the behavior of asof join to make it run slower. (i.e. Sleep a few seconds before actually starting the work in the processing thread) (2) Add a downstream node to asof join that processes data slowly (similar to a slow data sink), i.e., process one batch per second. This way, the backpressure would be pushed from the slow sink to asof join then to the data sources.
> 
> I think I prefer (2) a bit more because this affects represents a real life case of slow sink.
> 
> @westonpace I am not sure if the idea of GatedSourceNode is similar or different, but happy to hear

While I'm not sure exactly what Weston has in mind, my understanding is that the GatedSourceNode's goal is to avoid flakiness due to non-deterministic timing. IMO, both (1) and (2) above could be flaky due to non-deterministic timing.

Between (1) and (2) I also wouldn't prefer (1) because the debug-options would change the behavior of the as-of-join node being tested, and I prefer to change the code driving it instead.

icexelloss:
>  IMO, both (1) and (2) above could be flaky due to non-deterministic timing
I think this can be avoided if we make sure the slow sink is blocking. i.e. 

```
slow_sink.pause_processing() # make sure slow_sink stop processing so what we can trigger backpresure
# validate backpresure is triggered
slow_sink.resume_processing()
# validate plan exists normally and resume counter is expected
```

rtpsw:
I added a delaying node after the as-of-join-node and used different backpressure detection checks. I observed that the as-of-join-node calls to `Pause` and `Resume` are independent, because the former/latter occurs when crossing the high/low threshold up/down.

rtpsw:
Prior to the [recent commit](https://github.com/apache/arrow/pull/35874/commits/6db6574432d1c5f997b2496d5f7072a03dbcb946), I noticed that some (macOS, I think) CI jobs were timing out. I figured this happened because `gate()` never returned true, which means the queue never crossed a threshold, likely due to an extreme timing of threaded operations. Since the low and high queue thresholds are 4 and 8, respectively, my guess was the number of batches in the backpressure tests (previously 10) was too small. Doubling the number of batches should therefore help, and indeed this seems to work.

cpp/src/arrow/acero/asof_join_node_test.cc
```diff
@@ -1360,6 +1361,102 @@ TRACED_TEST(AsofJoinTest, TestUnorderedOnKey, {
       schema({field("time", int64()), field("key", int32()), field("r0_v0", float64())}));
 })
 
+struct BackpressureCounters {
+  int32_t pause_count = 0;
+  int32_t resume_count = 0;
```
pitrou, line 1366: Should these be atomic? Presumably they can be updated from multiple threads.

pitrou:
@github-actions crossbow submit ubuntu-cpp-thread-sanitizer

cpp/src/arrow/acero/asof_join_node_test.cc
pitrou: Confirmed by TSAN job.

rtpsw:
In the [recent commit](https://github.com/apache/arrow/pull/35874/commits/380aa94bd04df435a0be7c6ee05a5e2581008f77), I included the changes in [this PR](https://github.com/apache/arrow/pull/35904), which are necessary to avoid TSAN errors.

cpp/src/arrow/acero/asof_join_node.cc
```diff
@@ -529,7 +529,7 @@ class KeyHasher {
   size_t index_;
   std::vector<col_index_t> indices_;
   std::vector<KeyColumnMetadata> metadata_;
-  const RecordBatch* batch_;
+  std::atomic<const RecordBatch*> batch_;
```
icexelloss, line 532: Above it is saying "the key hasher is not thread-safe", if so, why do we care if this is atomic here?

icexelloss: Are we using the key hasher in multiple thread?

rtpsw: It is queried from one thread but invalidated from another. We discussed offline that this can be simplified so that the key hasher would only be used from one thread, but this is (currently?) out of scope for this PR.

westonpace (approved):
This works but I find it a little confusing. It seems strange to me that the delaying node would be placed after the asof join node. I was thinking it would be placed on the slow input to ensure that it did not deliver any batches.

However, if you were to do that, the entire plan would hang. This is because the delaying node synchronously blocks. This means the CPU thread is taken out of use. When use_threads=false (as it is when we use the asof join node) then that means the only working thread is blocked.

The reason it works today is because the delay happens after the asof join node and so it is actually blocking the processing thread and not a CPU thread. It was surprising to me that the processing thread was the thread that called InputReceived. This means, if you have something like...

Source -> AsofJoinNode -> Project -> Aggregation

Then the processing thread will be the one calling Project and Aggregation which is odd. However, this issue isn't very relevant to the backpressure problem at hand.

So it seems this works because there is enough data initially delivered to release at least one batch. This batch gets caught in the delayed node which hangs the processing thread. Since the processing thread is hung then enough data accumulates in the inputs that it will eventually pause the inputs.

If we want to proceed with this design then I think that is ok. For the sake of completeness I am including an example of what I had in my mind. A "GatedNode" would not block but it would not emit any batches (they would just queue up in the scheduler) until the gate is unlocked.

So, with this idea, the gated node would be placed over the slow source. You could get rid of the entire idea of "delayed" sources (and all the accompanying sleeps). The plan would be started and we wait to ensure backpressure is hit. Once backpressure is hit we release the gate and then confirm that the plan resumes and finishes.

Feel free to take my idea or continue with this one.

cpp/src/arrow/acero/asof_join_node.cc
icexelloss: I see in that case you can update the doc to explain how this class should be used in multi-threaded execution? Currently looks like the doc about thread safety is not correct, i.e., it is used by multiple thread while the docstring says it's not thread safe.

cpp/src/arrow/acero/asof_join_node_test.cc
```diff
@@ -1360,6 +1361,102 @@ TRACED_TEST(AsofJoinTest, TestUnorderedOnKey, {
...
+    RETURN_NOT_OK(ValidateExecNodeInputs(plan, inputs, 1, kKindName));
+    auto bp_options = static_cast<const BackpressureDelayingNodeOptions&>(options);
+    return plan->EmplaceNode<BackpressureDelayingNode>(
+        plan, inputs, inputs[0]->output_schema(), bp_options);
+  }
+
+  const char* kind_name() const override { return kKindName; }
+  Result<ExecBatch> ProcessBatch(ExecBatch batch) override {
+    while (!gate()) {
+      SleepFor(delay_seconds);
```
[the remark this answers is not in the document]
rtpsw: I went with `SleepABit`.

cpp/src/arrow/acero/asof_join_node.cc
rtpsw: I ended up just fixing it to be single-threaded.

icexelloss:
Let me take a look at this later

icexelloss:
Per discussion @rtpsw let's use the gated node because it doesn't rely on timing, update this PR and then we should be able to merge.

rtpsw:
> Per discussion @rtpsw let's use the gated node because it doesn't rely on timing, update this PR and then we should be able to merge.

This approach ran into a problem. See [this post](https://github.com/rtpsw/arrow/pull/4#issuecomment-1589441338).

cpp/src/arrow/acero/asof_join_node.cc
icexelloss: > I ended up just fixing it to be single-threaded.
Nice - can you point me where the fix is?

rtpsw: Sorry, I only pushed it in [the recent commit](https://github.com/apache/arrow/pull/35874/commits/e29520b4974e9bbc4ef08687913ce0199ab08b8a).

icexelloss (changes requested):
To clarify status of this PR - we should update the test to use GatedNode instead

rtpsw:
> To clarify status of this PR - we should update the test to use GatedNode instead

Done.

cpp/src/arrow/acero/asof_join_node.cc
```diff
@@ -370,15 +370,10 @@ struct MemoStore {
     times_.swap(memo.times_);
   }
 
-  // Updates the current time to `ts` if it is less. A different thread may win the race
-  // to update the current time to more than `ts` but not to less. Returns whether the
-  // current time was changed from its value at the beginning of this invocation.
+  // Updates the current time to `ts` if it is less. Returns true if updated.
   bool UpdateTime(OnType ts) {
-    OnType prev_time = current_time_;
```
icexelloss, line 377: Are these change out of the scope of this PR? If so, can you create separate ones for them?

cpp/src/arrow/acero/asof_join_node.cc
```diff
@@ -668,18 +663,19 @@ class InputState {
 
   static Result<std::unique_ptr<InputState>> Make(
       size_t index, TolType tolerance, bool must_hash, bool may_rehash,
-      KeyHasher* key_hasher, ExecNode* node, AsofJoinNode* output,
+      KeyHasher* key_hasher, ExecNode* asof_input, AsofJoinNode* asof_node,
```
icexelloss, line 666: Can you other functions in this file consistent with this naming here? e.g.
https://github.com/apache/arrow/blob/main/cpp/src/arrow/acero/asof_join_node.cc#L541

cpp/src/arrow/acero/asof_join_node.cc
```diff
@@ -821,8 +817,11 @@ class InputState {
         ++batches_processed_;
         latest_ref_row_ = 0;
         have_active_batch &= !queue_.TryPop();
-        if (have_active_batch)
+        if (have_active_batch) {
           DCHECK_GT(queue_.UnsyncFront()->num_rows(), 0);  // empty batches disallowed
+          key_hasher_->Invalidate();  // batch changed - invalidate key hasher's cache
```
icexelloss, line 822: Seems unrelated change - can you make a separate PR for this?

cpp/src/arrow/acero/asof_join_node_test.cc
```diff
@@ -1360,9 +1366,209 @@ TRACED_TEST(AsofJoinTest, TestUnorderedOnKey, {
...
+  }
+  void ResumeProducing(ExecNode* output, int32_t counter) override {
+    ++counters->resume_count;
+    inputs()[0]->ResumeProducing(this, counter);
+  }
+
+  BackpressureCounters* counters;
+};
+
+class Gate {
```
icexelloss, line 1419: Can you move the Gated node to a separate file? It can be reused in other back pressure related test too.

cpp/src/arrow/acero/asof_join_node_test.cc
```diff
@@ -1381,36 +1587,85 @@ void TestBackpressure(BatchesMaker maker, int num_batches, int batch_size,
...
+      DeclarationToExecBatchesAsync(asofjoin, exec_ctx);
+
+  auto has_bp_been_applied = [&] {
+    int total_paused = 0;
+    for (const auto& counters : bp_counters) {
+      total_paused += counters.pause_count;
+    }
+    // One of the inputs is gated.  The other two will eventually be paused by the asof
+    // join node
+    return total_paused >= 2;
```
icexelloss, line 1651: Shouldn't total_paused be >= 3? (The gated one + the other 2)?

icexelloss: Also, is this check robust? i.e. what happens if the gated node counter is > 1? Shouldn't we be checking that all counters are >1 here? (not the total)

cpp/src/arrow/acero/asof_join_node_test.cc
```diff
@@ -1381,36 +1587,85 @@ void TestBackpressure(BatchesMaker maker, int num_batches, int batch_size,
...
+
+  gate.ReleaseAllBatches();
+  ASSERT_FINISHES_OK_AND_ASSIGN(BatchesWithCommonSchema batches, batches_fut);
+
+  size_t total_resumed = 0;
+  for (const auto& counters : bp_counters) {
+    total_resumed += counters.resume_count;
   }
-  ASSERT_EQ(static_cast<int64_t>(num_batches * batch_size), total_length);
+  ASSERT_GE(total_resumed, 2);
```
icexelloss, line 1664: Shouldn't this be at least 3? (three sources)

icexelloss (changes requested):
@rtpsw A few comments:
(1) Please separate out threading related changes to asof join - those should be in separate PR (outside of original scope)
(2) Please update the original iGH ssue to better reflect the change in this PR
(3) other inline comments

cpp/src/arrow/acero/asof_join_node_test.cc
```diff
@@ -1360,9 +1366,209 @@ TRACED_TEST(AsofJoinTest, TestUnorderedOnKey, {
...
+  int num_allowed_batches_ = 0;
+  std::mutex mutex_;
+};
+
+struct GatedNodeOptions : public ExecNodeOptions {
+  explicit GatedNodeOptions(Gate* gate) : gate(gate) {}
+  Gate* gate;
+};
+
+struct GatedNode : public ExecNode, public TracedNode {
```
icexelloss, line 1470: Can you add documentation for this class to describe the purpose of this?

cpp/src/arrow/acero/asof_join_node.cc
rtpsw: The location you are pointing to has a different perspective of the input and output, and should not be changed. Also note that the class there could easily be pulled out for reuse, and should not be tied to specific internals of as-of-join node.

cpp/src/arrow/acero/asof_join_node.cc
rtpsw: Created #36094 and will revert here.

cpp/src/arrow/acero/asof_join_node.cc
rtpsw: Created #36094 and will revert here.

cpp/src/arrow/acero/asof_join_node.cc
icexelloss: Can you elaborate why input and output perspective are different?

rtpsw: We discussed this [here](https://github.com/apache/arrow/pull/35874#discussion_r1213557505).

cpp/src/arrow/acero/asof_join_node_test.cc
rtpsw: Done.

cpp/src/arrow/acero/asof_join_node_test.cc
rtpsw: Done.

cpp/src/arrow/acero/asof_join_node_test.cc
icexelloss: It looks the resume counter for the ungated nodes are >= 1 in all cases. So I think we can check them (both resume and paused) individually here. Instead of total.

cpp/src/arrow/acero/asof_join_node.cc
icexelloss: Not sure I fully understand, but from my understanding both the InputState and BackpressureController takes two node, one node presents the input table of asof join, and another is the asof join node itself. So my point here is it's better to make them consistent naming to avoid confusion. And you seem to say that InputState and BackpressureController have different perspective of input and output, which I do not fully get but perhaps I misunderstood what you mean.

icexelloss: (This is not a major point but I want to make sure I understand what you are saying)

rtpsw:
> (2) Please update the original iGH ssue to better reflect the change in this PR

It's authored by Weston - I don't have permission to edit. I edited [the description of this PR](https://github.com/apache/arrow/pull/35874#issue-1736711712) instead,

rtpsw:
This [CI job failure](https://github.com/apache/arrow/actions/runs/5282401504/jobs/9557256027?pr=35874) suggests that the [tightening condition](https://github.com/apache/arrow/pull/35874#discussion_r1229992514) (in [this commit](https://github.com/apache/arrow/pull/35874/commits/c458c2ee7442725d0a82a644a53dae52d51338e5)) isn't always true.

cpp/src/arrow/acero/asof_join_node.cc
rtpsw: We are looking at two nodes: a node pushing to an as-of-join node. The code here has the perspective of the latter node, so the former node is seen as an input, which is why the code has `asof_input`. The `BackpressureController` has the perspective of the former node, which it controls, so the latter node is seen as an output. `BackpressureController` is designed to be reusable outside the context of `AsofJoinNode` (even though currently it is not exposed).

cpp/src/arrow/acero/asof_join_node_test.cc
rtpsw: Empirically, yes. I'll try this.

icexelloss: I don't think this is purely empirically - Asof join should apply back-pressure to all other nodes when one if the input is gated, otherwise it is a bug I think.

icexelloss:
> > (2) Please update the original iGH ssue to better reflect the change in this PR
> 
> It's authored by Weston - I don't have permission to edit. I edited [the description of this PR](https://github.com/apache/arrow/pull/35874#issue-1736711712) instead,

Ok we can leave it as is then.

cpp/src/arrow/acero/asof_join_node.cc
icexelloss: @rtpsw Do we still need to change this. If so, why?

cpp/src/arrow/acero/asof_join_node_test.cc
```diff
@@ -1381,36 +1443,93 @@ void TestBackpressure(BatchesMaker maker, int num_batches, int batch_size,
...
+
+  auto has_bp_been_applied = [&] {
+    // One of the inputs is gated.  The other two will eventually be paused by the asof
+    // join node
+    for (size_t i = 0; i < source_configs.size(); i++) {
+      const auto& counters = bp_counters[i];
+      if (source_configs[i].is_gated) {
+        if (counters.pause_count > 0) return false;
+      } else {
+        if (counters.pause_count != 1) return false;
```
icexelloss, line 1509: Why `!=1` here instead of `> 0`? i.e., if the non-gated node has pause_count > 1, is it considered "bp has not been applied"?

cpp/src/arrow/acero/asof_join_node_test.cc
```diff
@@ -1381,36 +1443,93 @@ void TestBackpressure(BatchesMaker maker, int num_batches, int batch_size,
...
+  Future<BatchesWithCommonSchema> batches_fut =
+      DeclarationToExecBatchesAsync(asofjoin, exec_ctx);
+
+  auto has_bp_been_applied = [&] {
+    // One of the inputs is gated.  The other two will eventually be paused by the asof
+    // join node
+    for (size_t i = 0; i < source_configs.size(); i++) {
+      const auto& counters = bp_counters[i];
+      if (source_configs[i].is_gated) {
+        if (counters.pause_count > 0) return false;
```
icexelloss, line 1507: The logic here is a bit confusing. If pause_count > 0 then bp has not been applied? (Shouldn't it been the other way around?)

rtpsw: The logic around here checks the following expectations of correct application of backpressure: a gated node should not have been paused (checked by the current line) whereas an non-gated node should have been paused once (checked by the next if-statement).

icexelloss (approved):
LGTM

Issue #34391, opened 2023-03-01
[C++] Future as-of-join-node hangs on distant times


rtpsw:
### Describe the bug, including details regarding any error messages, version, and platform.

Future as-of-join-node goes into an infinite loop when right-table times are distant. A specific test case reproducing this is:
```
/*l*/ {R"([[0, 1, 1]])", R"([[1000, 2, 2]])", R"([[2000, 1, 3]])"},
/*r0*/ {R"([[0, 1, 10], [1500, 1, 11], [2500, 1, 12]])"},
/*r1*/ {R"([[0, 1, 100], [1500, 1, 101], [2500, 1, 102]])"},
/*exp_nokey*/ {R"([[0, 0, 1, 10, 100], [1000, 0, 2, 11, 101], [2000, 0, 3, 12, 102]])"},
/*exp_emptykey*/ {R"([[0, 1, 1, 10, 100], [1000, 2, 2, 11, 101], [2000, 1, 3, 12, 102]])"},
/*exp*/ {R"([[0, 1, 1, 10, 100], [1000, 2, 2, null, null], [2000, 1, 3, 12, 102]])"}, 1000);
```

### Component(s)

C++

PR #35878, opened 2023-06-01
GH-35838: [C++] Fix asof join backpresure

### LINKED ISSUE #35838: [C++] Backpressure broken in asof join node

icexelloss:
### Rationale for this change

To fix a bug in asof join backpresure handling.

This is reverting a bug introduced in GH-34391 on this line that breaks asof join backpresure
https://github.com/apache/arrow/pull/34392/files#diff-5493b6ae7ea2a4d5cfb581034c076e9c4be7608382168de6d1301ef67b6c01eeR1410

### Are these changes tested?
No. However code change simply reverts to the state before the bug was introduced in GH-34391 and therefore should be pretty safe (we have tested that the code before GH-34391 is working). In the meantime @rtpsw is working on adding tests around this but I would like to merge this to unblock internal  issues around this.

* Closes: #35838

### DIFF
--- a/cpp/src/arrow/acero/asof_join_node.cc
+++ b/cpp/src/arrow/acero/asof_join_node.cc
@@ -679,7 +679,7 @@ class InputState {
                           BackpressureHandler::Make(low_threshold, high_threshold,
                                                     std::move(backpressure_control)));
     return std::make_unique<InputState>(index, tolerance, must_hash, may_rehash,
-                                        key_hasher, node, std::move(handler), schema,
+                                        key_hasher, output, std::move(handler), schema,
                                         time_col_index, key_col_index);
   }
 
--- a/cpp/src/arrow/acero/asof_join_node.cc
+++ b/cpp/src/arrow/acero/asof_join_node.cc
@@ -538,7 +538,7 @@ class KeyHasher {
 
 class BackpressureController : public BackpressureControl {
  public:
-  BackpressureController(ExecNode* node, AsofJoinNode* output,
+  BackpressureController(ExecNode* node, ExecNode* output,
                          std::atomic<int32_t>& backpressure_counter)
       : node_(node), output_(output), backpressure_counter_(backpressure_counter) {}
 
@@ -547,7 +547,7 @@ class BackpressureController : public BackpressureControl {
 
  private:
   ExecNode* node_;
-  AsofJoinNode* output_;
+  ExecNode* output_;
   std::atomic<int32_t>& backpressure_counter_;
 };
 
--- a/cpp/src/arrow/acero/asof_join_node.cc
+++ b/cpp/src/arrow/acero/asof_join_node.cc
@@ -538,7 +538,7 @@ class KeyHasher {
 
 class BackpressureController : public BackpressureControl {
  public:
-  BackpressureController(ExecNode* node, ExecNode* output,
+  BackpressureController(ExecNode* node, AsofJoinNode* output,
                          std::atomic<int32_t>& backpressure_counter)
       : node_(node), output_(output), backpressure_counter_(backpressure_counter) {}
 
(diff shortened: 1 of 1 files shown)


cpp/src/arrow/acero/asof_join_node.cc
```diff
@@ -1407,7 +1407,7 @@ class AsofJoinNode : public ExecNode {
       ARROW_ASSIGN_OR_RAISE(
           auto input_state,
           InputState::Make(i, tolerance_, must_hash_, may_rehash_, key_hashers_[i].get(),
-                           this, inputs[i], backpressure_counter_,
+                           inputs[i], this, backpressure_counter_,
```
icexelloss, line 1410: Simply reverting this line where the bug is introduced:
https://github.com/apache/arrow/pull/34392/files#diff-5493b6ae7ea2a4d5cfb581034c076e9c4be7608382168de6d1301ef67b6c01eeR1410

@rtpsw is working on adding tests but it could take some time.

icexelloss:
cc @westonpace I prefer to merge this to unblock internal issues while @rtpsw  works on adding the test if that is OK

cpp/src/arrow/acero/asof_join_node.cc
rtpsw: Note that I have just added tests in #35874, though this PR can proceed, if so desired.

icexelloss:
@rtpsw In order to get the head into a non buggy state first. I will going to merge this so that @rtpsw can pull this and continue to work on testing + reducing confusion of the variable naming per discussions https://github.com/apache/arrow/pull/35874#discussion_r1213516786

westonpace:
@icexelloss makes sense

PR #35880, opened 2023-06-01
MINOR: [Documentation][Python] Minor tweaks to docs around IPC messages

amoeba:
### Rationale for this change

@jorisvandenbossche pointed out that the docstring for RecordBatch.serialize might not be clear enough and that users may be surprised to find out a Schema isn't included. This PR makes that clearer and improves up related documentation.

### Are these changes tested?

Yes. I built the docs and verified the rendering was correct and I ran the changed example locally. I didn't --doctest-cython it because it looks like that hasn't been set up for this module.

No, these are just docs improvements.

### DIFF
--- a/python/pyarrow/table.pxi
+++ b/python/pyarrow/table.pxi
@@ -2302,7 +2302,7 @@ cdef class RecordBatch(_Tabular):
         <pyarrow.Buffer address=0x... size=... is_cpu=True is_mutable=True>
 
         Reconstruct RecordBatch from IPC message Buffer and original Schema
-        
+
         >>> pa.ipc.read_record_batch(buf, batch.schema)
         pyarrow.RecordBatch
         n_legs: int64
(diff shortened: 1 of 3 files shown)


python/pyarrow/table.pxi
```diff
@@ -2292,8 +2297,17 @@ cdef class RecordBatch(_Tabular):
         >>> animals = pa.array(["Flamingo", "Parrot", "Dog", "Horse", "Brittle stars", "Centipede"])
         >>> batch = pa.RecordBatch.from_arrays([n_legs, animals],
         ...                                     names=["n_legs", "animals"])
-        >>> batch.serialize()
+        >>> buf = batch.serialize()
+        >>> buf
         <pyarrow.Buffer address=0x... size=... is_cpu=True is_mutable=True>
+        >>> # Reconstruct RecordBatch from IPC message Buffer and original Schema
```
AlenkaF, line 2303:
```suggestion

        Reconstruct RecordBatch from IPC message Buffer and original Schema
        
```

AlenkaF (approved):
Thank you for the PR!
The change looks good to me, just a minor nit in the docstrings.

> I didn't --doctest-cython it because it looks like that hasn't been set up for this module.

The `.pxi` files are checked with doctest under the file `lib.pyx`. So if you run
```
python -m pytest --doctest-cython python/pyarrow/lib.pyx 
``` 
that should test the changes in the `table.pxi`.

PR #35702, opened 2023-05-20
GH-35779: [R][Documentation] Document workaround for window-like functionality

dgreiss:
Issue #29537 describes how to perform an implicit window function. 

It was discussed on the mailing list: https://lists.apache.org/thread/b16ghtb8q9hyl64ks3dp9ftm7pvlnsdk to document this operation in the vignette. It's still not clear if this is a preferred way to apply these operations, but there is potential for significant performance on large data sets

* Closes: #35779

PR #35882, opened 2023-06-02
GH-35709: [R][Documentation] Document passing data to duckdb for windowed aggregates

### LINKED ISSUE #35709: [R] Add mention of moving data to/from DuckDB to the vignettes

#35702 documents how to get around the fact we can't do window functions by discussing how to do this with joins, but we should also mention we can do this by passing data to/from DuckDB

### Component(s)

R

dgreiss:
### Rationale for this change

#35702 documents how to use joins for computing windowed aggregates. This documents an alternative solution by passing data to duckdb. This use case was also mentioned on the [duckdb blog](https://duckdb.org/2021/12/03/duck-arrow.html).

Changes to vignette. 

* Closes: #35709

### DIFF
--- a/r/vignettes/data_wrangling.Rmd
+++ b/r/vignettes/data_wrangling.Rmd
@@ -1,7 +1,7 @@
 ---
 title: "Data analysis with dplyr syntax"
 description: >
-  Learn how to use the dplyr backend supplied by arrow 
+  Learn how to use the dplyr backend supplied by arrow
 output: rmarkdown::html_vignette
 ---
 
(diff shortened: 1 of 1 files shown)


thisisnic (commented):
Thanks for making this PR!  This looks good - one addition to suggest; would you mind adding in a call to `to_arrow()` in the code example? It's not technically needed before the `collect()`, but it'd be helpful to make users aware of it in case they have pipelines where they want to run more code in Arrow before pulling the data into memory.

dgreiss:
I rebased and I think I cleaned up this PR. Let me know if there are any other changes needed

thisisnic (approved):
Fantastic, thanks!

PR #35886, opened 2023-06-02
GH-35287: [C++][Parquet] Add CodecOptions to customize the compression parameter

### LINKED ISSUE #35287: [C++][Parquet] Expose the API to fine tune the window_bits parameter of ZLIB compression

ZLIB library supports history buffers of different sizes by setting windowBits parameter, while currently in Arrow it is set as a fixed value. It makes sense that setting the window size to the maximum number will give a better performance, while it does not provide too much flexibility. As we know there would be some scenarios where the memory efficiency is very limited, or the software stack is relatively old and do not have a large memory capacity. For those cases, the user may want to set the window bits to a small number to save some memory. At least it would be much more flexible to give users such an option to choose the window size when de/compressing. If that makes sense, we would like to add such a property for the user to set the ZLIB(GZIP) window_bits, and will ensure not affecting the existing function calls. Thanks!

### Component(s)

C++, Parquet

yyang52:
### Rationale for this change

Based on https://github.com/apache/arrow/issues/35287, we'd like to add a CodecOptions to make more compression parameters (such as window_bits) customizable when creating the Codec for parquet writer.

Authored-by: Yang Yang [yang10.yang@intel.com](mailto:yang10.yang@intel.com)
Co-authored-by: Rambacher, Mark [mark.rambacher@intel.com](mailto:mark.rambacher@intel.com)

Add CodecOptions and replace `compression_level` when creating the Codec. The design is basically based on previous discussions. 
### Are these changes tested?

Yes

Yes, when user creates the `WriterProperties`

* Closes: #35287


mapleFU (commented):
Seems parquet part change all compression level to options. Do we need to keep api compatible?

cpp/src/parquet/properties.h
```diff
@@ -172,8 +172,8 @@ class PARQUET_EXPORT ColumnProperties {
     max_stats_size_ = max_stats_size;
   }
 
-  void set_compression_level(int compression_level) {
```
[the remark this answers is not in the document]
yyang52: For now I replace the compression_level with CodecOptions, could also keep that for compatibility :)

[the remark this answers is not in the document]
yyang52: seems like python codec api also takes compression_level as input, which also rely on cpp API (where we changed to Codec::Create(Compression::type, CodecOptions)). should we also modify python API or keep Codec::Create(Compression::type, compression_level)?

cpp/src/parquet/column_writer.h
```diff
@@ -87,8 +88,9 @@ class PARQUET_EXPORT PageWriter {
 
   static std::unique_ptr<PageWriter> Open(
       std::shared_ptr<ArrowOutputStream> sink, Compression::type codec,
-      int compression_level, ColumnChunkMetaDataBuilder* metadata,
-      int16_t row_group_ordinal = -1, int16_t column_chunk_ordinal = -1,
+      const std::shared_ptr<CodecOptions>& codec_options,
```
mapleFU, line 91: This should be put to the end for capability

cpp/src/parquet/properties.h
```diff
@@ -192,6 +197,8 @@ class PARQUET_EXPORT ColumnProperties {
 
   int compression_level() const { return compression_level_; }
```
mapleFU, line 198: Since `Codec` contains `compression_level_`, would this be duplicated or useless?

mapleFU:
cc @pitrou @wgtmac

cpp/src/parquet/column_writer.cc
```diff
@@ -683,20 +683,21 @@ class BufferedPageWriter : public PageWriter {
...
+    const std::shared_ptr<CodecOptions>& codec_options,
+    ColumnChunkMetaDataBuilder* metadata, int16_t row_group_ordinal,
+    int16_t column_chunk_ordinal, MemoryPool* pool, bool buffered_row_group,
+    std::shared_ptr<Encryptor> meta_encryptor, std::shared_ptr<Encryptor> data_encryptor,
+    bool page_write_checksum_enabled, ColumnIndexBuilder* column_index_builder,
+    OffsetIndexBuilder* offset_index_builder) {
   if (buffered_row_group) {
     return std::unique_ptr<PageWriter>(new BufferedPageWriter(
-        std::move(sink), codec, compression_level, metadata, row_group_ordinal,
+        std::move(sink), codec, codec_options, metadata, row_group_ordinal,
```
wgtmac, line 694: I am not sure if we are able to move `codec` into `codec_options` as well. I don't have a preference here.

cpp/src/arrow/util/compression.h
```diff
@@ -124,7 +168,9 @@ class ARROW_EXPORT Codec {
 
   /// \brief Create a codec for the given compression algorithm
   static Result<std::unique_ptr<Codec>> Create(
-      Compression::type codec, int compression_level = kUseDefaultCompressionLevel);
+      Compression::type codec,
```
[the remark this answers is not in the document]
yyang52: Yes, that could solve the compatibility issue of different APIs. Just to make it clear, you suggest marking the newly added API with CodecOptions as deprecated right? Or the previous one?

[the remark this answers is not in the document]
wgtmac: I mean deprecating the existing one.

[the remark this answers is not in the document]
yyang52: Great, that makes sense :)

cpp/src/parquet/column_writer.cc
```diff
@@ -249,9 +249,9 @@ int LevelEncoder::Encode(int batch_size, const int16_t* levels) {
 class SerializedPageWriter : public PageWriter {
  public:
   SerializedPageWriter(std::shared_ptr<ArrowOutputStream> sink, Compression::type codec,
-                       int compression_level, ColumnChunkMetaDataBuilder* metadata,
-                       int16_t row_group_ordinal, int16_t column_chunk_ordinal,
-                       bool use_page_checksum_verification,
+                       const std::shared_ptr<CodecOptions>& codec_options,
```
pitrou, line 252: No need for `shared_ptr` here.
```suggestion
                       const CodecOptions& codec_options,
```

cpp/src/parquet/column_writer.h
```diff
@@ -21,6 +21,7 @@
 #include <cstring>
 #include <memory>
 
+#include "arrow/util/compression.h"
```
pitrou, line 24:
```suggestion
#include "arrow/type_fwd.h"
```

cpp/src/parquet/column_writer.h
```diff
@@ -87,8 +88,9 @@ class PARQUET_EXPORT PageWriter {
 
   static std::unique_ptr<PageWriter> Open(
       std::shared_ptr<ArrowOutputStream> sink, Compression::type codec,
-      int compression_level, ColumnChunkMetaDataBuilder* metadata,
```
pitrou, line 90: Let's not remove this API but create a new one taking `const CodecOptions&`, to avoid breaking API compatibility.

cpp/src/parquet/types.cc
```diff
@@ -51,9 +51,30 @@ bool IsCodecSupported(Compression::type codec) {
...
+       << " not supported in Parquet format";
+    throw ParquetException(ss.str());
+  }
+
+  PARQUET_ASSIGN_OR_THROW(result, Codec::Create(codec, codec_options));
+  return result;
+}
+
+// Deprecated and use CodecOptions to create Codec instead
 std::unique_ptr<Codec> GetCodec(Compression::type codec, int compression_level) {
```
pitrou, line 78: Here as well, can you implement this overload in terms of the one above?

r/src/compression.cpp
```diff
@@ -23,7 +23,8 @@
 // [[arrow::export]]
 std::shared_ptr<arrow::util::Codec> util___Codec__Create(arrow::Compression::type codec,
                                                          R_xlen_t compression_level) {
-  return ValueOrStop(arrow::util::Codec::Create(codec, compression_level));
+  return ValueOrStop(arrow::util::Codec::Create(
+      codec, std::make_shared<CodecOptions>(CodecOptions(compression_level))));
```
pitrou, line 27: Can probably revert this.

pitrou (changes requested):

pitrou:
@yyang52 There are Python test failures on CI.

Also, is it possible to pass a base `CodecOptions` when a derived class exists? (e.g. for gzip). Ideally it should be supported.

cpp/src/arrow/util/compression.h
```diff
@@ -124,7 +168,15 @@ class ARROW_EXPORT Codec {
...
+      const std::shared_ptr<CodecOptions>& codec_options =
+          std::make_shared<CodecOptions>(kUseDefaultCompressionLevel));
+
+  /// \brief Create a codec for the given compression algorithm
+  /// \deprecated and left for backwards compatibility.
+  ARROW_DEPRECATED("Use CodecOptions to create Codec")
```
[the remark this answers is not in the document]
wgtmac: Sounds reasonable. I was thinking it is always preferable to use the new `CodecOptions` to set the level.

cpp/src/arrow/util/compression_zlib.cc
```diff
@@ -17,15 +17,14 @@
 
 #include "arrow/util/compression_internal.h"
 
+#include <zconf.h>
+#include <zlib.h>
```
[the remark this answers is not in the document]
yyang52: Sure, that might be due to some local format auto-change, will revert those.

cpp/src/arrow/util/compression.h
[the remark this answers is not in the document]
yyang52: That makes more sense, will keep the existing one as it was

yyang52:
> @yyang52 There are Python test failures on CI.
> 
> Also, is it possible to pass a base `CodecOptions` when a derived class exists? (e.g. for gzip). Ideally it should be supported.

Yes, I think it supports that (as checked in TEST(TestCodecMisc, SpecifyCompressionLevel) to pass base CodecOptions to different Codecs)

cpp/src/arrow/util/compression.h
```diff
@@ -107,6 +107,46 @@ class ARROW_EXPORT Decompressor {
...
+class ARROW_EXPORT GZipCodecOptions : public CodecOptions {
+ public:
+  GZipFormat::type gzip_format = GZipFormat::GZIP;
+  int window_bits = kGZipDefaultWindowBits;
+};
+
+// ----------------------------------------------------------------------
+// brotli codec options implementation
+
+constexpr int kBrotliDefaultWindowBits = 22;
```
mapleFU, line 143: Can it use `BROTLI_DEFAULT_WINDOW`?

cpp/src/parquet/column_io_benchmark.cc
```diff
@@ -40,8 +40,8 @@ std::shared_ptr<Int64Writer> BuildWriter(int64_t output_size,
                                          ColumnDescriptor* schema,
                                          const WriterProperties* properties,
                                          Compression::type codec) {
-  std::unique_ptr<PageWriter> pager =
-      PageWriter::Open(dst, codec, Codec::UseDefaultCompressionLevel(), metadata);
+    std::unique_ptr<PageWriter> pager =
```
mapleFU, line 43: fmt?

cpp/src/parquet/column_writer.cc
```diff
@@ -690,26 +690,40 @@ class BufferedPageWriter : public PageWriter {
...
+    const std::shared_ptr<CodecOptions>& codec_options,
+    ColumnChunkMetaDataBuilder* metadata, int16_t row_group_ordinal,
+    int16_t column_chunk_ordinal, MemoryPool* pool, bool buffered_row_group,
+    std::shared_ptr<Encryptor> meta_encryptor, std::shared_ptr<Encryptor> data_encryptor,
+    bool page_write_checksum_enabled, ColumnIndexBuilder* column_index_builder,
+    OffsetIndexBuilder* offset_index_builder) {
   if (buffered_row_group) {
     return std::unique_ptr<PageWriter>(new BufferedPageWriter(
-        std::move(sink), codec, compression_level, metadata, row_group_ordinal,
+        std::move(sink), codec, *codec_options.get(), metadata, row_group_ordinal,
```
mapleFU, line 701: I think just `*codec_options` is ok?

cpp/src/arrow/util/compression.h
```diff
@@ -107,6 +107,50 @@ class ARROW_EXPORT Decompressor {
...
+/// \brief Compression codec options
+class ARROW_EXPORT CodecOptions {
+ public:
+  CodecOptions(int compression_level = kUseDefaultCompressionLevel) {
+    compression_level_ = compression_level;
+  }
+  virtual ~CodecOptions() = default;
+
+  int compression_level_;
+};
```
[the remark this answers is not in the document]
mapleFU: I guess @yyang52 's idea is that `CodecOptions` can explicit built from `compression_level`, am I right?

cpp/src/parquet/properties.h
mapleFU: Now we have a `compression_level` and a `CodecOptions`. Previously, user can set `compression_level`, and the compression level would be changed.

After this patch, if user change the compression level, the codec will not be changed.

Can we make `CodecOptions` mutate it's `compression_level`, and make `set_compression_level` set the level in `CodecOptions` directly?

mapleFU (commented):
I'm general ok with the parquet side changes now, there are some comments.

cpp/src/arrow/util/compression.h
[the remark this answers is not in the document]
yyang52: yes, then we could do one-line built for CodecOptions with `compression_level`, which has been used in several places

cpp/src/parquet/properties.h
yyang52: that makes sense :)

cpp/src/arrow/util/compression.h
yyang52: yes, but then we need to include the `brotli` header here as well?

yyang52: yes, I think that's just how it is implemented now in `GZip/BrotliCodecOptions`?

mapleFU: Sure, I notice that this patch hard code the encoder's internal into `compression.h`, so I'm afraid if they're changed in the 3rd libraries, we'll get confused here.

yyang52: yeah it seems better to use their `BROTLI_DEFAULT_WINDOW`

cpp/src/arrow/util/compression.h
```diff
@@ -107,6 +107,46 @@ class ARROW_EXPORT Decompressor {
...
+
+// ----------------------------------------------------------------------
+// gzip codec options implementation
+
+struct GZipFormat {
```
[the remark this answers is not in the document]
yyang52: I just move `GZipFormat` from `compression_internal.h` to `compression.h` to be able to use it in `CodecOptions` without changing the type.

cpp/src/parquet/properties.h
```diff
@@ -174,6 +175,16 @@ class PARQUET_EXPORT ColumnProperties {
 
   void set_compression_level(int compression_level) {
     compression_level_ = compression_level;
```
mapleFU, line 177: Can we just remove `compression_level_`?

cpp/src/parquet/properties.h
mapleFU: Can we just get it from `codec_options_`?

mapleFU (commented):
The parquet part looks ok. However:
1. Should we remove "compression_level_` in Builder and properties, and make the interface as a wrapper for visiting `CodecOptions`?
2. Would `CodecOptions` with explicit cast be API broken @pitrou ?

cpp/src/parquet/column_writer.h
pitrou: Also, unless the `PageWriter` needs to keep ownership of the codec options, this should simply be `const CodecOptions& = {}`.

cpp/src/parquet/column_writer_test.cc
```diff
@@ -162,6 +163,25 @@ class TestPrimitiveWriter : public PrimitiveTypedTest<TestType> {
     ASSERT_NO_FATAL_FAILURE(this->ReadAndCompare(compression, num_rows, enable_checksum));
   }
 
+  void TestRequiredWithCodecOptions(
+      Encoding::type encoding, Compression::type compression, bool enable_dictionary,
+      bool enable_statistics, int64_t num_rows = SMALL_SIZE,
+      int compression_level = Codec::UseDefaultCompressionLevel(),
+      ::arrow::util::GZipFormat::type format = ::arrow::util::GZipFormat::GZIP,
+      int window_bits = ::arrow::util::kGZipDefaultWindowBits,
+      bool enable_checksum = false) {
```
pitrou, line 172: Why not simply take a `const CodecOptions&` here?

cpp/src/parquet/properties.h
```diff
@@ -154,6 +154,7 @@ class PARQUET_EXPORT ColumnProperties {
         statistics_enabled_(statistics_enabled),
         max_stats_size_(max_stats_size),
         compression_level_(Codec::UseDefaultCompressionLevel()),
+        codec_options_(std::make_shared<CodecOptions>()),
```
pitrou, line 157: Can we just leave this null-initialized by default?

cpp/src/parquet/properties.h
pitrou: +1, I think we can just keep the codec options.

cpp/src/parquet/properties.h
```diff
@@ -618,6 +677,7 @@ class PARQUET_EXPORT WriterProperties {
     std::unordered_map<std::string, Encoding::type> encodings_;
     std::unordered_map<std::string, Compression::type> codecs_;
     std::unordered_map<std::string, int32_t> codecs_compression_level_;
```
pitrou, line 679: Let's remove this and only keep `codec_options_`?

pitrou (commented):
Thanks for the update @yyang52 ! I think this still needs some improvement, see comments below.

cpp/src/arrow/util/compression.h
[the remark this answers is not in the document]
wgtmac: It was an internal class in the past. As it is moved to a public header now, changing it to enum class would be preferable.

cpp/src/arrow/util/compression.h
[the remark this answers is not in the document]
wgtmac: That sounds reasonable for now. But it would be harder to use once we add more parameters.

cpp/src/arrow/util/compression.h
```diff
@@ -107,6 +107,46 @@ class ARROW_EXPORT Decompressor {
...
+  CodecOptions(int compression_level = kUseDefaultCompressionLevel) {
+    compression_level_ = compression_level;
+  }
+  virtual ~CodecOptions() = default;
+
+  int compression_level_;
+};
+
+// ----------------------------------------------------------------------
+// gzip codec options implementation
```
wgtmac, line 122:
```suggestion
// GZip codec options implementation
```

cpp/src/arrow/util/compression.h
```diff
@@ -107,6 +107,46 @@ class ARROW_EXPORT Decompressor {
...
+    GZIP,
+  };
+};
+
+constexpr int kGZipDefaultWindowBits = 15;
+
+class ARROW_EXPORT GZipCodecOptions : public CodecOptions {
+ public:
+  GZipFormat::type gzip_format = GZipFormat::GZIP;
+  int window_bits = kGZipDefaultWindowBits;
```
wgtmac, line 137: This could be easily out of sync with the upstream. What about using `std::optional<int>` and always use the library default value if it is unset?

cpp/src/arrow/util/compression.h
wgtmac: Same with my previous comment. We can use `BROTLI_DEFAULT_WINDOW` in the `.cc` file

cpp/src/parquet/properties.h
wgtmac: +1 on this. `compression_level_` is duplicated.

cpp/src/parquet/column_writer.h
```diff
@@ -21,6 +21,7 @@
 #include <cstring>
 #include <memory>
 
+#include "arrow/type_fwd.h"
```
wgtmac, line 24: Is this required? It should be included by `parquet/platform.h`.

cpp/src/parquet/column_writer.h
wgtmac: In addition to the comments above, should be consolidate `Compression::type codec` and `const std::shared_ptr<CodecOptions>& codec_options`? codec_options should know codec.

cpp/src/parquet/column_writer.h
wgtmac: Should we mark the old one as deprecated? The new one should always be preferable and we don't want to maintain different overloads in the long term.

cpp/src/parquet/properties.h
```diff
@@ -190,7 +193,14 @@ class PARQUET_EXPORT ColumnProperties {
 
   size_t max_statistics_size() const { return max_stats_size_; }
 
-  int compression_level() const { return compression_level_; }
+  int compression_level() const { return codec_options_->compression_level_; }
+
+  const std::shared_ptr<CodecOptions>& codec_options() const {
+    if (!codec_options_) {
+      return std::make_shared<CodecOptions>();
```
wgtmac, line 200: This is a problem:
```
C:\projects\arrow\cpp\src\parquet\properties.h(200) : error C2220: the following warning is treated as an error
C:\projects\arrow\cpp\src\parquet\properties.h(200) : warning C4172: returning address of local variable or temporary
ninja: build stopped: subcommand failed.
```

yyang52: this seems to be fixed, but there are Python errors that I have no ideas on the hints

yyang52:
Addressed most comments with most CI checks passed (2 fails seem also happening on other PRs). One left thing is hardcoding `window_bits` in `CodecOptions` may easily be out-of-sync with third-party dependency, as mentioned by @wgtmac. Do you have any suggestions/thoughts on that? Please also have a look at other changes I made and see if any other improvements needed. Thanks! @pitrou @mapleFU @wgtmac

wgtmac:
https://github.com/apache/arrow/actions/runs/5367297586/jobs/9737382909?pr=35886

There seems to be lint error.

yyang52:
> https://github.com/apache/arrow/actions/runs/5367297586/jobs/9737382909?pr=35886
> 
> There seems to be lint error.

Doing some fixes, could you help to re-trigger other checks? Thanks!

cpp/src/arrow/util/compression.h
pitrou: That sounds like a good idea.

cpp/src/arrow/util/compression.h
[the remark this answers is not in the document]
pitrou: @yyang52 I'm not sure I understand what you're saying. Both definitions allow writing `CodecOptions{compression_level}`, right?

cpp/src/arrow/util/compression.h
mapleFU: I also vote for this

cpp/src/parquet/platform.h
```diff
@@ -24,7 +24,8 @@
 #include "arrow/io/interfaces.h"  // IWYU pragma: export
 #include "arrow/status.h"         // IWYU pragma: export
 #include "arrow/type_fwd.h"       // IWYU pragma: export
-#include "arrow/util/macros.h"    // IWYU pragma: export
+#include "arrow/util/compression.h"
```
pitrou, line 27: I don't think adding this inclusion is necessary, right?

cpp/src/parquet/column_writer.h
pitrou: We can indeed.

cpp/src/arrow/util/compression.h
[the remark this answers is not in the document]
yyang52: Per tests, we may need to do this? Am I missing something?
```  
CodecOptions codec_options;
codec_options.compression_level_ = compression_level;
return Codec::Create(codec_type, codec_options);
```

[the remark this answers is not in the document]
pitrou: Doesn't `CodecOptions{compression_level}` work?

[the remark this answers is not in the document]
yyang52: It gives me the compilation error:
```
error: no matching constructor for initialization of 'arrow::util::CodecOptions'
  return Codec::Create(codec_type, CodecOptions{compression_level});
                                   ^           ~~~~~~~~~~~~~~~~~~~
/home/yangyang/arrow/cpp/src/arrow/util/compression.h:111:20: note: candidate constructor (the implicit copy constructor) not viable: no known conversion from 'int' to 'const arrow::util::CodecOptions' for 1st argument
class ARROW_EXPORT CodecOptions {
                   ^
/home/yangyang/arrow/cpp/src/arrow/util/compression.h:111:20: note: candidate constructor (the implicit default constructor) not viable: requires 0 arguments, but 1 was provided
1 error generated.
```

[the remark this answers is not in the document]
pitrou: Ah, sorry! Let's forget it then.

cpp/src/parquet/platform.h
yyang52: If not, then it will prevent us to initialize as `CodecOptions{}`? Am I doing this right?
```
column_writer.h:101:43: error: invalid use of incomplete type 'parquet::CodecOptions' (aka 'arrow::util::CodecOptions')
      const CodecOptions& codec_options = CodecOptions{});
                                          ^~~~~~~~~~~~~~
```

cpp/src/parquet/properties.h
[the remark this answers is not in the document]
mapleFU: I guess both can be used

cpp/src/parquet/platform.h
mapleFU: Maybe forward declr like `default_memory_pool()` can be used?

pitrou: @yyang52 You don't need it in `platform.h`.

cpp/src/arrow/util/compression.h
```diff
@@ -107,6 +107,44 @@ class ARROW_EXPORT Decompressor {
...
 
+/// \brief Compression codec options
+class ARROW_EXPORT CodecOptions {
+ public:
+  explicit CodecOptions(int compression_level = kUseDefaultCompressionLevel) {
+    compression_level_ = compression_level;
+  }
+  virtual ~CodecOptions() = default;
+
+  int compression_level_;
```
mapleFU, line 118: So would this be changed to a `std::nullopt`?

yyang52: yeah I made some changes to these part, will push those soon.

cpp/src/parquet/platform.h
yyang52: yes, could I include that directly in `column_writer.h`, as the forward declaration is not sufficient to initialize an object as far as I learned?

yyang52:
May you help to review the code change as well as trigger the remaining CI checks? Thanks! @mapleFU @pitrou @wgtmac

yyang52:
There're 6 failing checks, but seems like not introduced by this PR?

yyang52:
Hi, may I know if there is any change needed or concerns regarding this PR? Thanks! 
@mapleFU @pitrou @wgtmac

mapleFU:
I'm a little busy these days, I'll take a look tonight

yyang52:
> I'm a little busy these days, I'll take a look tonight

Sure, take your time. Just let me know if more actions needed, thanks!

pitrou:
@yyang52 We're busy preparing a release right now. You might start by merging/rebasing from main to get a more passing CI.

yyang52:
> @yyang52 We're busy preparing a release right now. You might start by merging/rebasing from main to get a more passing CI.

Sure, no rush! I will rebase from main and re-run CI checks.

yyang52:
> May you help to review the code change as well as trigger the remaining CI checks?

May you help to trigger the remaining CI checks again? Thanks! @pitrou

cpp/src/parquet/column_writer.h
```diff
@@ -85,6 +87,22 @@ class PARQUET_EXPORT PageWriter {
...
+      std::shared_ptr<Encryptor> header_encryptor = NULLPTR,
+      std::shared_ptr<Encryptor> data_encryptor = NULLPTR,
+      bool page_write_checksum_enabled = false,
+      // column_index_builder MUST outlive the PageWriter
+      ColumnIndexBuilder* column_index_builder = NULLPTR,
+      // offset_index_builder MUST outlive the PageWriter
+      OffsetIndexBuilder* offset_index_builder = NULLPTR,
+      const CodecOptions& codec_options = CodecOptions{});
+
+  ARROW_DEPRECATED("Use the one with CodecOptions instead")
```
pitrou, line 105:
```suggestion
  ARROW_DEPRECATED("Deprecated in 13.0.0. Use CodecOptions-taking overload instead.")
```

cpp/src/arrow/util/compression.h
```diff
@@ -107,6 +108,40 @@ class ARROW_EXPORT Decompressor {
...
 
+/// \brief Compression codec options
+class ARROW_EXPORT CodecOptions {
+ public:
+  explicit CodecOptions(int compression_level = kUseDefaultCompressionLevel) {
+    compression_level_ = compression_level;
+  }
+  virtual ~CodecOptions() = default;
+
+  int compression_level_ = kUseDefaultCompressionLevel;
```
pitrou, line 119: Forgot this, but style nit (this is a public member, it's weird to add a trailing underscore):
```suggestion
  int compression_level = kUseDefaultCompressionLevel;
```

pitrou (changes requested):
@yyang52 There are a couple of comments left to address, can you take a look?

mapleFU (approved):
Parquet part looks ok to me

yyang52:
> @yyang52 There are a couple of comments left to address, can you take a look?

Doing some fix, may you help to take a look and trigger remaining CI checks? Thanks! @pitrou

wgtmac (approved):
LGTM. Nice work!

pitrou:

pitrou (approved):

pitrou:
Thanks a lot for your contribution @yyang52 ! I hope you didn't mind the delays.

yyang52:
> 

Thanks for your kind support! It's a pleasure to make contribution to Arrow community! @pitrou @mapleFU @wgtmac

cpp/src/parquet/file_writer.cc
```diff
@@ -291,12 +304,24 @@ class RowGroupSerializer : public RowGroupWriter::Contents {
...
-          properties_->page_checksum_enabled(), ci_builder, oi_builder);
+      auto codec_options = properties_->codec_options(path)
+                               ? (properties_->codec_options(path)).get()
+                               : nullptr;
+
+      std::unique_ptr<PageWriter> pager;
+      if (!codec_options) {
+        pager = PageWriter::Open(sink_, properties_->compression(path), col_meta,
+                                 row_group_ordinal_, static_cast<int16_t>(column_ordinal),
+                                 properties_->memory_pool(), false, meta_encryptor,
```
mapleFU, line 315: This part lost `buffered_row_group_`, and set it to false. Sorry that I didn't found it at first time

yyang52: Sorry for this missing and thanks for the fix!

PR #35887, opened 2023-06-02
MINOR: [Docs] Fix typo in docs for dense union data type

nenb:
## Rationale for this change / What changes are included in this PR?

Fixed typo in `Columnar.rst`. 

I think this typo may have crept in from an [older](https://wesm.github.io/arrow-site-test/format/Layout.html#example-layout-dense-union) version of the docs when validity bitmaps were allowed for Union types.

## Are these changes tested?

N/A

## Are there any user-facing changes?

Only in the docs.

### DIFF
--- a/docs/source/format/Columnar.rst
+++ b/docs/source/format/Columnar.rst
@@ -622,7 +622,7 @@ will have the following layout: ::
 
     * Children arrays:
       * Field-0 array (f: Float32):
-        * Length: 2, Null count: 1
+        * Length: 3, Null count: 1
         * Validity bitmap buffer: 00000101
 
         * Value Buffer:


nenb:
@AlenkaF Apologies for the notification, I wasn't sure how else to add a reviewer.

Please do let me know if this not good behaviour, and what I should do in future.

Thanks for all your work!

AlenkaF:
Thank you for the contribution @nenb, pinging me like this is fine for sure 👍 

I think the change is correct. The null count is now added to the first child array (field `f`), so the length of this array has to be increased by 1.

Will merge.

AlenkaF (approved):

PR #35889, opened 2023-06-02
GH-32566: [C++] Connect parquet to the new scan node

### LINKED ISSUE #32566: [C++] Create fragment scanners for csv/parquet/orc/ipc
Once we have the basic scan node ready (with an initial implementation based on in-memory fragments) then we can add over the file-format versions.  We may also want to consider adding JSON support for datasets at this time.

**Reporter**: [Weston Pace](https://issues.apache.org/jira/browse/ARROW-17288) / @westonpace
**Assignee**: [Weston Pace](https://issues.apache.org/jira/browse/ARROW-17288) / @westonpace
#### PRs and other links:
- [GitHub Pull Request #14663](https://github.com/apache/arrow/pull/14663)

<sub>**Note**: *This issue was originally created as [ARROW-17288](https://issues.apache.org/jira/browse/ARROW-17288). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

westonpace:
### Rationale for this change

This ended up being considerably more change than just connecting parquet to the new scan node.  In order to do this I had to refactor the scan node itself somewhat.  It introduces the concept of scan tasks (or maybe scan streams would be a better name) to help clarify the concept of a row group (which I didn't have to worry about with CSV).  I also introduced the staging area which is a slightly different approach to sequencing that I think will be much simpler.

The new scan node now supports the parquet format.

### Are these changes tested?

Yes

There are breaking changes to the scan2 node but this feature hasn't really been released yet.
* Closes: #32566

### DIFF
--- a/cpp/src/parquet/arrow/reader.cc
+++ b/cpp/src/parquet/arrow/reader.cc
@@ -16,7 +16,6 @@
 // under the License.
 
 #include "parquet/arrow/reader.h"
-#include <sys/types.h>
 
 #include <algorithm>
 #include <cstring>
--- a/cpp/src/parquet/arrow/reader.cc
+++ b/cpp/src/parquet/arrow/reader.cc
@@ -16,10 +16,12 @@
 // under the License.
 
 #include "parquet/arrow/reader.h"
+#include <sys/types.h>
 
 #include <algorithm>
 #include <cstring>
 #include <memory>
+#include <queue>
 #include <unordered_set>
 #include <utility>
 #include <vector>
@@ -58,6 +60,7 @@ using arrow::Future;
 using arrow::Int32Array;
 using arrow::ListArray;
 using arrow::MemoryPool;
+using arrow::RecordBatch;
 using arrow::RecordBatchReader;
 using arrow::ResizableBuffer;
 using arrow::Result;
@@ -326,6 +329,23 @@ class FileReaderImpl : public FileReader {
     return ReadRowGroups(row_groups, Iota(reader_->metadata()->num_columns()), table);
   }
 
+  Result<AsyncBatchGenerator> DoReadRowGroupsAsync(
+      const std::vector<int>& row_groups, const std::vector<int>& indices,
+      ::arrow::internal::Executor* cpu_executor, bool allow_sliced_batches);
+
+  AsyncBatchGenerator ReadRowGroupsAsync(const std::vector<int>& row_groups,
+                                         const std::vector<int>& indices,
+                                         ::arrow::internal::Executor* cpu_executor,
+                                         bool allow_sliced_batches) override {
+    Result<AsyncBatchGenerator> batch_gen =
+        DoReadRowGroupsAsync(row_groups, indices, cpu_executor, allow_sliced_batches);
+    if (batch_gen.ok()) {
+      return batch_gen.MoveValueUnsafe();
+    }
+    return ::arrow::MakeFailingGenerator<std::shared_ptr<RecordBatch>>(
+        batch_gen.status());
+  }
+
   Status ReadRowGroup(int row_group_index, const std::vector<int>& column_indices,
                       std::shared_ptr<Table>* out) override {
     return ReadRowGroups({row_group_index}, column_indices, out);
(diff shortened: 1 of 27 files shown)


westonpace:
This is very much still a draft.  There are still a lot of tests to add and some TODOs (column projection and row filtering) but I don't expect the overall structure to change too much if anyone wanted to take an early look.

westonpace:
On the bright side, I can now reach max parallelism with about 3GB of RAM, regardless of the size of row groups (and performance looks to be about 10% better but still very early to say that)

cpp/src/parquet/arrow/reader.cc
```diff
@@ -1243,6 +1285,124 @@ Status FileReaderImpl::ReadRowGroups(const std::vector<int>& row_groups,
...
+                    "encountered that was too large to fit in a single batch.");
+              }
+              state->overflow.push(std::move(next_batch));
+            } else {
+              first = std::move(next_batch);
+            }
+          }
+          if (!first) {
+            // TODO(weston): Test this case
+            return Status::Invalid("Unexpected empty row group");
```
mapleFU, line 1364: Hmmm I guess a RowGroup can be empty currently. You can easily generate a case like this using python `write_table`

```
table.len() == 10000
write_table(table, 2000)
```

mapleFU (commented):
So, this would prevent from `use_thread` deadlock?

cpp/src/parquet/arrow/reader.cc
```diff
@@ -16,10 +16,12 @@
 // under the License.
 
 #include "parquet/arrow/reader.h"
+#include <sys/types.h>
```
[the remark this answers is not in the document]
westonpace: No, thank you for noticing.  I will clean this up soon.

westonpace:
> So, this would prevent from use_thread deadlock?

Yes.  Since this method is async then the caller can choose not to block.  Previously we used `GetRecordBatchGenerator` which is very similar to the method I ended up creating (`ReadRowGroupsAsync`).  The difference is that `GetRecordBatchGenerator` ignores `batch_size` and `ReadRowGroupsAsync` does not.  My hope is that, eventually, `GetRecordBatchGenerator` can be deprecated and removed.

westonpace:
I think the changes here are probably too extensive to expect review.  I will be breaking this PR up into multiple PRs.

 * Adding a new `ReadRowGroupsAsync` to the file reader
 * Reworking the scan node
 * Adding parquet support

cpp/src/parquet/arrow/reader.cc
westonpace: Doesn't this create a table with 5 row groups?  Why would this be an empty row group?

westonpace:
The first of these PRs is now available:

 [ ] #36779

cpp/src/parquet/arrow/reader.cc
mapleFU:
```c++
  Status WriteTable(const Table& table, int64_t chunk_size) override {
    RETURN_NOT_OK(table.Validate());

    if (chunk_size <= 0 && table.num_rows() > 0) {
      return Status::Invalid("chunk size per row_group must be greater than 0");
    } else if (!table.schema()->Equals(*schema_, false)) {
      return Status::Invalid("table schema does not match this writer's. table:'",
                             table.schema()->ToString(), "' this:'", schema_->ToString(),
                             "'");
    } else if (chunk_size > this->properties().max_row_group_length()) {
      chunk_size = this->properties().max_row_group_length();
    }

    auto WriteRowGroup = [&](int64_t offset, int64_t size) {
      RETURN_NOT_OK(NewRowGroup(size));
      for (int i = 0; i < table.num_columns(); i++) {
        RETURN_NOT_OK(WriteColumnChunk(table.column(i), offset, size));
      }
      return Status::OK();
    };

    if (table.num_rows() == 0) {
      // Append a row group with 0 rows
      RETURN_NOT_OK_ELSE(WriteRowGroup(0, 0), PARQUET_IGNORE_NOT_OK(Close()));
      return Status::OK();
    }

    for (int chunk = 0; chunk * chunk_size < table.num_rows(); chunk++) {
      int64_t offset = chunk * chunk_size;
      RETURN_NOT_OK_ELSE(
          WriteRowGroup(offset, std::min(chunk_size, table.num_rows() - offset)),
          PARQUET_IGNORE_NOT_OK(Close()));
    }
    return Status::OK();
  }
```

It's from this code. It's easy to flush the rowgroup that row_num == 0

westonpace: I manually created some empty row groups.  It turns this branch should be unreachable because, further up, we will have noticed that "remaining rows" is 0 and returned an end marker.  I've updated this code and added a test case for this scenario in #36779

mapleFU: Okay, great!

westonpace:
The second PR is now available.  Once the two pre-reqs merge I will undraft this.

PR #35899, opened 2023-06-04
GH-35909: [Go] Deprecate `arrow.MapType.ValueField` & `arrow.MapType.ValueType` methods

### LINKED ISSUE #35909: Deprecate `arrow.MapType.ValueField` & `arrow.MapType.ValueType` methods

Instead of those methods `arrow.MapType.ElemField` & `arrow.MapType.Elem` should be used (as they correspond to treating MapType as list-like

### Component(s)

Go

candiduslynx:
### Rationale for this change

Follow-up for #35885 

* Added `ElemField() Field` to `arrow.ListLikeType` interface
* Added `ElemField() Field` to `arrow.MapType` implementation
* Added deprecation notice to `arrow.MapType.ValueField` & `arrow.MapType.ValueType`
* Fixed a bug in `go/arrow/array/map.go` (`NewMapBuilderWithType` used `ValueType` instead of `ItemType`)

### Are these changes tested?

Compile-time assertion for corresponding types.

* Added `ElemField() Field` to `arrow.ListLikeType` interface
* Added `ElemField() Field` to `arrow.MapType` implementation
* Added deprecation notice to `arrow.MapType.ValueField` & `arrow.MapType.ValueType`

* Closes: #35909

### DIFF
--- a/go/parquet/pqarrow/encode_arrow.go
+++ b/go/parquet/pqarrow/encode_arrow.go
@@ -36,23 +36,17 @@ import (
 
 // get the count of the number of leaf arrays for the type
 func calcLeafCount(dt arrow.DataType) int {
-	if dt, ok := dt.(arrow.ListLikeType); ok {
-		return calcLeafCount(dt.Elem())
-	}
-
-	switch dt.ID() {
-	case arrow.EXTENSION:
-		return calcLeafCount(dt.(arrow.ExtensionType).StorageType())
-	case arrow.SPARSE_UNION, arrow.DENSE_UNION:
-		panic("arrow type not implemented")
-	case arrow.DICTIONARY:
-		return calcLeafCount(dt.(*arrow.DictionaryType).ValueType)
-	case arrow.STRUCT:
+	switch dt := dt.(type) {
+	case arrow.ExtensionType:
+		return calcLeafCount(dt.StorageType())
+	case arrow.NestedType:
 		nleaves := 0
-		for _, f := range dt.(*arrow.StructType).Fields() {
+		for _, f := range dt.Fields() {
 			nleaves += calcLeafCount(f.Type)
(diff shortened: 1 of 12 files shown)


candiduslynx:
@zeroshade I wonder if you'll be open to removing the `ValueType` & `ValueField` for `arrow.MapType` (so that it doesn't have the duplicate methods with the same results)?

candiduslynx:
@zeroshade any chance to get an initial review today?

go/parquet/pqarrow/encode_arrow.go
```diff
@@ -36,19 +36,17 @@ import (
 
 // get the count of the number of leaf arrays for the type
 func calcLeafCount(dt arrow.DataType) int {
+	if dt, ok := dt.(arrow.ListLikeType); ok {
+		return calcLeafCount(dt.Elem())
+	}
```
zeroshade, line 41: could we further simply this by doing 

```go
if dt, ok := dt.(arrow.NestedType); ok {
    nleaves := 0
    for _, f := range dt.Fields() {
        nleaves += calcLeafCount(f.Type)
    }
    return nleaves
}
```
Which will cover struct types in addition to the list/map types?

zeroshade (changes requested):
LGTM just one nit

go/parquet/pqarrow/encode_arrow.go
candiduslynx: Done in 9ce73c4a0e8ccbc0c1f0fdc632666df4dc947750.
I removed the panic on the union types then (as they are nested, BTW).

zeroshade (approved):

candiduslynx:
@zeroshade I think the tests have passed.
Any additional issues in the PR?

PR #35902, opened 2023-06-04
GH-35837: [C++] Acero will hang if StopProducing is called while backpressure is applied on the source node

### LINKED ISSUE #35837: [C++] Acero will hang if StopProducing is called while backpressure is applied on the source node

The source node creates an external task called `SourceNode::DatasetScan` which is managed outside the exec plan.  This task is actually an `arrow::Loop` call which is iterating the source async generator.  That loop can be paused and resumed to apply backpressure.

When StopProducing arrives the source node sets `stop_requested_=true`.  However, `stop_requested_` is not processed until the next iteration through the loop.  If the plan is paused then the loop is not iterating and nothing notices that `stop_requested_` changed and the external task never completes and the plan then never completes.

### Component(s)

C++

rtpsw:

The stopping logic for `SourceNode` now ensures the node is first resumed.

### Are these changes tested?

Yes. A new `source_node_test` tester is added for this.

No.

* Closes: #35837

### DIFF
--- a/cpp/src/arrow/acero/source_node.cc
+++ b/cpp/src/arrow/acero/source_node.cc
@@ -259,6 +259,7 @@ struct SourceNode : ExecNode, public TracedNode {
         return;
       }
       to_finish = backpressure_future_;
+      backpressure_future_ = Future<>::MakeFinished();
     }
     to_finish.MarkFinished();
   }
@@ -270,6 +271,7 @@ struct SourceNode : ExecNode, public TracedNode {
       std::lock_guard<std::mutex> lg(mutex_);
       if (!backpressure_future_.is_finished()) {
         to_finish = backpressure_future_;
+        backpressure_future_ = Future<>::MakeFinished();
       }
     }
     if (to_finish.is_valid()) {
--- a/cpp/src/arrow/acero/source_node.cc
+++ b/cpp/src/arrow/acero/source_node.cc
@@ -263,6 +263,22 @@ struct SourceNode : ExecNode, public TracedNode {
     to_finish.MarkFinished();
   }
 
+  Status StopProducing() override {
+    // GH-35837: ensure node is not paused
+    Future<> to_finish;
+    {
+      std::lock_guard<std::mutex> lg(mutex_);
+      if (!backpressure_future_.is_finished()) {
+        to_finish = backpressure_future_;
+      }
+    }
+    if (to_finish.is_valid()) {
+      to_finish.MarkFinished();
+    }
+    // only then stop
+    return ExecNode::StopProducing();
+  }
+
   Status StopProducingImpl() override {
     std::unique_lock<std::mutex> lock(mutex_);
     stop_requested_ = true;
--- a/cpp/src/arrow/acero/source_node_test.cc
+++ b/cpp/src/arrow/acero/source_node_test.cc
@@ -1,0 +1,109 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#include <gtest/gtest.h>
+
+#include "arrow/acero/exec_plan.h"
+#include "arrow/acero/map_node.h"
+#include "arrow/acero/options.h"
+#include "arrow/acero/test_nodes.h"
+
+namespace arrow {
+namespace acero {
+
+struct PauseThenStopNodeOptions : public ExecNodeOptions {
+  explicit PauseThenStopNodeOptions(int num_pass) : num_pass(num_pass) {}
+
+  int num_pass;
+};
+
+struct PauseThenStopNode : public MapNode {
+  static constexpr const char* kKindName = "PauseThenStopNode";
+  static constexpr const char* kFactoryName = "pause_then_stop";
+
+  static void Register() {
+    auto exec_reg = default_exec_factory_registry();
+    if (!exec_reg->GetFactory(kFactoryName).ok()) {
+      ASSERT_OK(exec_reg->AddFactory(kFactoryName, PauseThenStopNode::Make));
+    }
+  }
+
+  PauseThenStopNode(ExecPlan* plan, std::vector<ExecNode*> inputs,
+                    std::shared_ptr<Schema> output_schema,
+                    const PauseThenStopNodeOptions& options)
+      : MapNode(plan, inputs, output_schema), num_pass(options.num_pass) {}
+
+  static Result<ExecNode*> Make(ExecPlan* plan, std::vector<ExecNode*> inputs,
+                                const ExecNodeOptions& options) {
+    RETURN_NOT_OK(ValidateExecNodeInputs(plan, inputs, 1, kKindName));
+    auto bp_options = static_cast<const PauseThenStopNodeOptions&>(options);
+    if (bp_options.num_pass < 2) {
+      return Status::Invalid("num_pass must be at least 2");
+    }
+    return plan->EmplaceNode<PauseThenStopNode>(plan, inputs, inputs[0]->output_schema(),
+                                                bp_options);
+  }
+
+  const char* kind_name() const override { return kKindName; }
+  Result<ExecBatch> ProcessBatch(ExecBatch batch) override {
+    if (num_pass == 2) {
+      inputs()[0]->PauseProducing(this, 1);
+    } else if (num_pass == 1) {
+      ARROW_RETURN_NOT_OK(inputs()[0]->StopProducing());
+    }
+    if (num_pass > 0) --num_pass;
+    return batch;
+  }
+
+  int num_pass;
+};
+
+TEST(SourceNode, PauseThenStop) {
+  PauseThenStopNode::Register();
+
+  constexpr int num_pass = 2, num_batches = 10, batch_size = 1;
+  ASSERT_GE(num_pass, 2);
+  ASSERT_GT(num_batches, num_pass);
+  auto t_schema = schema({field("time", int32()), field("value", int32())});
+  ASSERT_OK_AND_ASSIGN(auto t_batches,
+                       MakeIntegerBatches({[](int row) -> int64_t { return row; },
+                                           [](int row) -> int64_t { return row + 1; }},
+                                          t_schema, num_batches, batch_size));
+
+  Declaration t_src = {
+      "source", SourceNodeOptions(t_batches.schema,
+                                  MakeDelayedGen(t_batches, "t_src", /*delay_sec=*/0.5,
+                                                 /*noisy=*/false))};
+  Declaration ctrl = {
+      PauseThenStopNode::kFactoryName, {t_src}, PauseThenStopNodeOptions(num_pass)};
+
+  ASSERT_OK_AND_ASSIGN(std::unique_ptr<RecordBatchReader> batch_reader,
+                       DeclarationToReader(ctrl, /*use_threads=*/false));
+
+  int64_t total_length = 0;
+  for (;;) {
+    ASSERT_OK_AND_ASSIGN(auto batch, batch_reader->Next());
+    if (!batch) {
+      break;
+    }
+    total_length += batch->num_rows();
+  }
+  ASSERT_EQ(static_cast<int64_t>(num_pass * batch_size), total_length);
+}
+
+}  // namespace acero
+}  // namespace arrow
--- a/cpp/src/arrow/acero/source_node_test.cc
+++ b/cpp/src/arrow/acero/source_node_test.cc
@@ -31,17 +31,8 @@ struct PauseThenStopNodeOptions : public ExecNodeOptions {
   int num_pass;
 };
 
+template <typename ThisNode>
 struct PauseThenStopNode : public MapNode {
-  static constexpr const char* kKindName = "PauseThenStopNode";
-  static constexpr const char* kFactoryName = "pause_then_stop";
-
-  static void Register() {
-    auto exec_reg = default_exec_factory_registry();
-    if (!exec_reg->GetFactory(kFactoryName).ok()) {
-      ASSERT_OK(exec_reg->AddFactory(kFactoryName, PauseThenStopNode::Make));
-    }
-  }
-
   PauseThenStopNode(ExecPlan* plan, std::vector<ExecNode*> inputs,
                     std::shared_ptr<Schema> output_schema,
                     const PauseThenStopNodeOptions& options)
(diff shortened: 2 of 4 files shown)


rtpsw:
cc @icexelloss

cpp/src/arrow/acero/source_node.cc
```diff
@@ -263,6 +263,22 @@ struct SourceNode : ExecNode, public TracedNode {
     to_finish.MarkFinished();
   }
 
+  Status StopProducing() override {
+    // GH-35837: ensure node is not paused
+    Future<> to_finish;
+    {
+      std::lock_guard<std::mutex> lg(mutex_);
+      if (!backpressure_future_.is_finished()) {
```
icexelloss, line 271: What does this condition mean here

cpp/src/arrow/acero/source_node_test.cc
```diff
@@ -0,0 +1,109 @@
...
+      ARROW_RETURN_NOT_OK(inputs()[0]->StopProducing());
+    }
+    if (num_pass > 0) --num_pass;
+    return batch;
+  }
+
+  int num_pass;
+};
+
+TEST(SourceNode, PauseThenStop) {
```
icexelloss, line 75: Can you add comments to explain what this is testing? i.e., "Testing that Plan can be properly stopped when back presure is applied to the source node.

cpp/src/arrow/acero/source_node_test.cc
```diff
@@ -0,0 +1,109 @@
...
+  PauseThenStopNode::Register();
+
+  constexpr int num_pass = 2, num_batches = 10, batch_size = 1;
+  ASSERT_GE(num_pass, 2);
```
icexelloss, line 79: What are these assertion doing?

cpp/src/arrow/acero/source_node_test.cc
```diff
@@ -0,0 +1,109 @@
...
+    }
+    return plan->EmplaceNode<PauseThenStopNode>(plan, inputs, inputs[0]->output_schema(),
+                                                bp_options);
+  }
+
+  const char* kind_name() const override { return kKindName; }
+  Result<ExecBatch> ProcessBatch(ExecBatch batch) override {
+    if (num_pass == 2) {
+      inputs()[0]->PauseProducing(this, 1);
+    } else if (num_pass == 1) {
```
icexelloss, line 65: When will this be 1?

icexelloss: Oh nvm, I see you are decrementing it

cpp/src/arrow/acero/source_node_test.cc
```diff
@@ -0,0 +1,109 @@
...
+    auto bp_options = static_cast<const PauseThenStopNodeOptions&>(options);
+    if (bp_options.num_pass < 2) {
+      return Status::Invalid("num_pass must be at least 2");
+    }
+    return plan->EmplaceNode<PauseThenStopNode>(plan, inputs, inputs[0]->output_schema(),
+                                                bp_options);
+  }
+
+  const char* kind_name() const override { return kKindName; }
+  Result<ExecBatch> ProcessBatch(ExecBatch batch) override {
```
icexelloss, line 62: Can you explain why this is supposed to work? Looks like the first batch will pause the input, then the second batch will stop? Could it happen that the second batch doesn't arrive because of the pause?

cpp/src/arrow/acero/source_node.cc
rtpsw: This code block implements essentially the same logic as `ResumeProducing`, which @westonpace could likely explain better than myself.

cpp/src/arrow/acero/source_node_test.cc
rtpsw: The tester won't work when the assertions on `num_pass` and `num_batches` fail. They prevent incorrect changes of the constants.

cpp/src/arrow/acero/source_node_test.cc
rtpsw: `num_pass` is decremented with each batch and stops at 0, so when `num_pass` is `1` (resp. `2`) the code is handling the last (resp. second-from-last) batch out of `n` - the initial value of `num_pass`.

icexelloss:
@rtpsw Curious why did you say "This PR contains a "Critical Fix"? Is this generated by some template?

rtpsw:
> @rtpsw Curious why did you say "This PR contains a "Critical Fix"? Is this generated by some template?

Yes, it's coming from the PR-template for the description. It asks for this labeling in certain cases, like when the issue can cause a crash (or hang, I guess).

cpp/src/arrow/acero/source_node_test.cc
rtpsw: This works because `PauseProducing` is not an immediate operation, which is useful for the current tester. In a debugging session, I observed that [the backpressure condition is detected immediately](https://github.com/apache/arrow/blob/a0d28deefd660adc6b14c4ac5fe4ed4175d4fedb/cpp/src/arrow/acero/source_node.cc#L213-L214) but [the loop](https://github.com/apache/arrow/blob/a0d28deefd660adc6b14c4ac5fe4ed4175d4fedb/cpp/src/arrow/acero/source_node.cc#L191) leads to [checking for termination](https://github.com/apache/arrow/blob/a0d28deefd660adc6b14c4ac5fe4ed4175d4fedb/cpp/src/arrow/util/future.h#L810) only after `StopProducing` is called. I suspect the non-immediacy is a consequence of managing using futures. @westonpace should be able to explain in more detail.

cpp/src/arrow/acero/source_node_test.cc
rtpsw: Done.

cpp/src/arrow/acero/source_node_test.cc
```diff
@@ -0,0 +1,109 @@
...
+
+  int64_t total_length = 0;
+  for (;;) {
+    ASSERT_OK_AND_ASSIGN(auto batch, batch_reader->Next());
+    if (!batch) {
+      break;
+    }
+    total_length += batch->num_rows();
+  }
+  ASSERT_EQ(static_cast<int64_t>(num_pass * batch_size), total_length);
```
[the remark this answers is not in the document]
rtpsw: > Can we test this in a way that https://github.com/apache/arrow/blob/main/cpp/src/arrow/acero/exec_plan.h#L107 is actually called? (Since this what happens in real case).

Done.

cpp/src/arrow/acero/source_node.cc
```diff
@@ -263,6 +263,22 @@ struct SourceNode : ExecNode, public TracedNode {
...
+        to_finish = backpressure_future_;
```
westonpace, line 272:
```suggestion
        to_finish = backpressure_future_;
        backpressure_future_ = Future<>::MakeFinished();
```

We might want to do this here and in `ResumeProducing`.  Otherwise, if `ResumeProducing` and `StopProducing` are called at the same time then both calls might try and mark the same future finished (which isn't valid).

cpp/src/arrow/acero/source_node_test.cc
icexelloss: Can you add a code comment please

cpp/src/arrow/acero/source_node_test.cc
westonpace: That sounds about right.  In practice, at any given moment, we assume the source node is in the middle of obtaining the next batch (e.g. reading a batch from flight or reading a batch from the disk or some other operation).  This is the primary loop.  If we strip away the async stuff (and ignore synchronization) it would look something like...

```
void StartProducing() {
  while (true) {
    while (!paused_) {
      Batch next = source.Poll();
      if (!next) {
        // Finished reading the source
        return;
      }
      output_->InputReceived(next);
    }
    paused_cv_.wait();
  }
}

void PauseProducing() {
  paused_ = true;
}

void ResumeProducing() {
  paused_ = false;
  paused_cv_.notify_all();
}
```

Still, I would naively expect what Li said might technically be possible (e.g. 99% of the time when PauseProducing is called the above loop will be in `source.Poll` because that's the slow part. So you will get at least one batch after the pause.  However, if the timing is unlucky, maybe it is between `source.Poll` and the `!paused` check and you don't end up getting one more batch.

However, this is all running with `use_threads=false` which means the polling loop and the loop here are maybe all the same thread and so maybe the behavior is deterministic and correct.  I'd have to look further.

westonpace (approved):
I think the fix is valid.

I'm not sure about the test but it's probably ok.  In order to trigger the issue it's not enough to call pause and then stop.  You need to call pause, wait for the source thread to break out of the loop, and then call stop.  I wonder if there might be some false positive where it calls pause and then calls stop before the source thread has had a chance to break out of the loop.

However, there is no easy way I can think of to know for certain that the source node has broken out of its loop.

Still, a test that might have a chance at a false positive is better than no test at all.

cpp/src/arrow/acero/source_node.cc
icexelloss: @rtpsw Is this addressed?

cpp/src/arrow/acero/source_node_test.cc
icexelloss: @rtpsw Does this mean that the current test relies on one more batch to be delivered after PauseProducing is called, other wise this test will deadlock and never finishes? (i.e., we never call StopProducing)

icexelloss: Although I suspect with the current implementation this is guaranteed, but I feel the fact that the test relies on that isn't great. It is possible to write the test in a more robust fashion? 

i.e., reusing the gated node idea, we can:
(1) Start a query with gated node closed so that back pressure is applied to source
(2) Call StopProducing 
(3) Validate that the execution is properly cancelled.

cpp/src/arrow/acero/source_node_test.cc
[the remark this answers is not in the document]
icexelloss: Maybe instead of using a thread pool and custom exec context, what if we 
(1) have a PauseNode/GateNode that calls PauseProducing as soon as it receives input
(2) expose a debug method that check if a SourceNode is paused. (SourceNode::IsPaused())
(3) use DeclarationToReader to start the plan
(4) Wait for SourceNode to be paused (using SourceNode::IsPaused())
(5) Call StopProducing on the created ExecPlan

I think the main difference from what Weston's above is the IsPaused() method, which might be useful in other back pressure related test as well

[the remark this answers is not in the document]
icexelloss: Some more thoughts. I think we need
(1) A reliable/controlled way to trigger back pressure (similar to Weston's gate node idea)
(2) A way to validate that a Source node is paused because of backpresure (either via SourceNode::IsPaused()) or sth else
(3) A way to stop the execution while the query is running (either by providing an executor/thread to run the query while the main thread stop the execution, or have the main thread run the query and have a separate timed thread to stop the execution. Or just have one thread (if we manage to do this with just one thread)

I agree with Weston that this test sort of works, but since we are touching back pressure/cancellation quite a bit I think it is worthwhile to improve the testing around this.

cpp/src/arrow/acero/source_node.cc
rtpsw: Yes, [here](https://github.com/apache/arrow/pull/35902/files#diff-0f58fd9023e0533b4e9149839729afa86c16341c1bd5ddd8c7578ef6bc441968R262) and [here](https://github.com/apache/arrow/pull/35902/files#diff-0f58fd9023e0533b4e9149839729afa86c16341c1bd5ddd8c7578ef6bc441968R274).

cpp/src/arrow/acero/source_node_test.cc
rtpsw: I changed to an immediate stop after pause, which is a simpler way to avoid relying on this guarantee.

cpp/src/arrow/acero/source_node_test.cc
rtpsw: Done.

cpp/src/arrow/acero/source_node_test.cc
```diff
@@ -0,0 +1,161 @@
...
+      ARROW_RETURN_NOT_OK(static_cast<ThisNode*>(this)->DoStopProducing());
+    }
+    if (num_pass > 0) --num_pass;
+    return batch;
+  }
+
+  int num_pass;
+};
+
+struct PauseThenStopInputNode : public PauseThenStopNode<PauseThenStopInputNode> {
```
icexelloss, line 65: What is the difference purpose of `PauseThenStopInputNode` and `PauseThenStopPlanNode`

cpp/src/arrow/acero/source_node_test.cc
```diff
@@ -0,0 +1,161 @@
...
+      ARROW_RETURN_NOT_OK(
+          exec_reg->AddFactory(kFactoryName, PauseThenStopPlanNode::Make));
+    }
+    return Status::OK();
+  }
+
+  using PauseThenStopNode::PauseThenStopNode;
+
+  Status DoStopProducing() {
+    plan()->StopProducing();
```
icexelloss, line 103: A node that shutdown the entire plan looks weird - I don't think any other node does this. Can we trigger the StopProducing the normal way? (i.e., from the thread that invokes the execution).

icexelloss: In general, I would prefer the test case to as close to real scenario as possible. I.e., the real scenario is:
(1) Back pressure is applied 
(2) ExecPlan::StopProducing() by user
(3) Plan doesn't stop

So the test case here ideally should be close to that case

cpp/src/arrow/acero/source_node_test.cc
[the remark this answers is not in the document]
icexelloss: > I don't think it would help us to test whether a (source) node is paused. The current issue is about the pausing behavior, which manifests in backpressure, not about the pausing state, which may well just be a flag (that may even lack behavior, in an extreme case).

It is not perfect and relies on the assumption (if PauseProducing() is called on input node, then backpressure is properly applied). If we really want to check the source node is indeed paused, then I think we need to do what Weston suggest above which is perhaps a bit overkill. Checking the backpresure future or Checking PauseProducing() is called is a reasonable mid ground IMO.

> I don't think pausing is the object of testing here; backpressure behavior is, and pausing is just a tool to ensure backpressure.
Similar to above, yes it is not the object of testing, but is a proxy for back pressure since I am not aware of a way to check backpressure directly (the backpresure_future_ is a private variable)

> I can relate to this, though. If we're aiming for heavy testing of backpressure/cancellation interactions, we should probably create an issue for each kind of testing we have in mind. In my mind, the current PR deals with the current issue and has enough value on its own.
The test I have in mind is described above:
```
(1) have a PauseNode/GateNode that calls PauseProducing as soon as it receives input
(2) expose a debug method that check if a SourceNode is paused. (SourceNode::IsPaused())
(3) use DeclarationToReader to start the plan
(4) Wait for SourceNode to be paused (using SourceNode::IsPaused())
(5) Call StopProducing on the created ExecPlan
```
I agree this PR seems to fix the bug however I'd prefer to have a test that can reflect the actual usage, i.e. ExecPlan::StopProducing() gets called by user when back pressure is applied

cpp/src/arrow/acero/source_node_test.cc
rtpsw: > A node that shutdown the entire plan looks weird

Wouldn't a fatal error detected by the node be a good reason to do this?

> I don't think any other node does this

Perhaps. But it should still useful to test this -  a call to `StopProducing` is supposed to work from anywhere.

> Can we trigger the StopProducing the normal way? (i.e., from the thread that invokes the execution).

Sure. Why not as an additional test?

cpp/src/arrow/acero/source_node_test.cc
rtpsw: The former calls `StopProducing` on the input node and the latter on the plan. Pre-PR, both cause a hang, so we need to cover these cases.

icexelloss:
FYI I am removing "This PR contains a Critical Fix" because I don't think this applies to one of the below:

"
(a) a security vulnerability, (b) a bug that caused incorrect or invalid data to be produced, or (c) a bug that causes a crash (even when the API contract is upheld)
"

icexelloss (changes requested):
Request to add a test that call StopProducing() from the test thread after planning is running and backpressure is applied

westonpace:
> FYI I am removing "This PR contains a Critical Fix" because I don't think this applies to one of the below:
> 
> "
> (a) a security vulnerability, (b) a bug that caused incorrect or invalid data to be produced, or (c) a bug that causes a crash (even > when the API contract is upheld)
> "

I think a deadlock counts as a crash or, at least, should be marked a critical fix (we can update the wording).

icexelloss:
> > FYI I am removing "This PR contains a Critical Fix" because I don't think this applies to one of the below:
> > "
> > (a) a security vulnerability, (b) a bug that caused incorrect or invalid data to be produced, or (c) a bug that causes a crash (even > when the API contract is upheld)
> > "
> 
> I think a deadlock counts as a crash or, at least, should be marked a critical fix (we can update the wording).



> > FYI I am removing "This PR contains a Critical Fix" because I don't think this applies to one of the below:
> > "
> > (a) a security vulnerability, (b) a bug that caused incorrect or invalid data to be produced, or (c) a bug that causes a crash (even > when the API contract is upheld)
> > "
> 
> I think a deadlock counts as a crash or, at least, should be marked a critical fix (we can update the wording).

Sounds good. Let me add it back then.

icexelloss:
In the spirit of not wasting too much time here arguing what is a good test, I am going to approve and merge what you have. It's that the test I had in mind but it is better than nothing.

icexelloss (approved):

PR #35907, opened 2023-06-05
GH-35906: [Docs] Enable building the documentation without having pyarrow installed

### LINKED ISSUE #35906: [Docs] Enable the build of the documentation without pyarrow built from source

We should enable the build of the documentation (all of it, only format section or only developers section) without having pyarrow built from source (that is without pyarrow or with pyarrow installed as a binary).

Connected to the work in https://github.com/apache/arrow/issues/30627 and https://github.com/apache/arrow/issues/35609.

### Component(s)

Documentation

AlenkaF:
### Rationale for this change

Ease the process of building the documentation for dev purposes.

`conf.py` is updated in a way to permit having pyarrow not installed (from source or as a binary).
In case pyarrow is not available:
- `docs/source/python` folder will be excluded from the build of the documentation
-  version of the documentation will be set to `'0.0.0-local-docs-build'`

I have tested the changes for cases when:
- pyarrow was built from source
- without pyarrow
- pyarrow was installed from PyPI, version 12.0.0

with building all of the docs and only format/developers sections.
* Closes: #35906

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -49,6 +49,7 @@ sys.path.extend([
 try:
     import pyarrow
     exclude_patterns = []
+    pyarrow_version =  pyarrow.__version
 
     # Conditional API doc generation
 
@@ -92,6 +93,7 @@ try:
         pyarrow.parquet.encryption = sys.modules['pyarrow.parquet.encryption'] = mock.Mock()
 except (ImportError, LookupError):
     exclude_patterns = ['python']
+    pyarrow_version =  ""
     cuda_enabled = False
     flight_enabled = False
 
@@ -206,10 +208,10 @@ author = u'Apache Software Foundation'
 # built documents.
 #
 # The short X.Y version.
-version = os.environ.get('ARROW_DOCS_VERSION',"")
+version = os.environ.get('ARROW_DOCS_VERSION', pyarrow_version)
 
 # The full version, including alpha/beta/rc tags.
-release = os.environ.get('ARROW_DOCS_VERSION',"")
+release = os.environ.get('ARROW_DOCS_VERSION', pyarrow_version)
 
 if "+" in release:
     release = release.split(".dev")[0] + " (dev)"
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -90,7 +90,7 @@ try:
     except ImportError:
         parquet_encryption_enabled = False
         pyarrow.parquet.encryption = sys.modules['pyarrow.parquet.encryption'] = mock.Mock()
-except:
+except (ImportError, LookupError):
     exclude_patterns = ['python']
     cuda_enabled = False
     flight_enabled = False
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -49,7 +49,7 @@ sys.path.extend([
 try:
     import pyarrow
     exclude_patterns = []
-    pyarrow_version =  pyarrow.__version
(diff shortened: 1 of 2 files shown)


docs/source/conf.py
```diff
@@ -38,15 +38,63 @@
...
+        orc_enabled = False
+        pyarrow.orc = sys.modules['pyarrow.orc'] = mock.Mock()
+
+    try:
+        import pyarrow.parquet.encryption
+        parquet_encryption_enabled = True
+    except ImportError:
+        parquet_encryption_enabled = False
+        pyarrow.parquet.encryption = sys.modules['pyarrow.parquet.encryption'] = mock.Mock()
+except:
```
jorisvandenbossche, line 93:
```suggestion
except ImportError:
```

?

docs/source/conf.py
```diff
@@ -158,11 +206,18 @@
 # built documents.
 #
 # The short X.Y version.
-version = os.environ.get('ARROW_DOCS_VERSION',
-                         pyarrow.__version__)
+try:
+    version = os.environ.get('ARROW_DOCS_VERSION',
+                             pyarrow.__version__)
```
jorisvandenbossche, line 211: I think you could try to avoid passing `pyarrow.__version__` to this function, as that will fail if pyarrow wasn't imported, even when ARROW_DOCS_VERSION is specified. 

So that we allow you to set the version with that env variable, also for the case pyarrow isn't installed? (I don't know how important this use case is, though)

jorisvandenbossche (commented):
Thanks for the PR!

One question: this is excluding the full of the python docs. Longer term, should we try to _only_ exclude the API docs? (there are currently a few files in the `python/` directory that require pyarrow to be installed to run the examples in the `ipython` directives, but many files also don't have this, and could in theory be built without pyarrow being installed)

docs/source/conf.py
AlenkaF: So yeah, we could default to an empty string if `ARROW_DOCS_VERSION` is not specified. I guess that can't happen (env var not being set) when publishing the docs with the release, right?

docs/source/conf.py
jorisvandenbossche: That might be specific to you local install (it's still finding the editable dev version?), so best to also include ImportError (if you are in an env without any pyarrow installed, it should give an ImportError)

jorisvandenbossche:
> I can try a bit further to see where I get, if you think this would be beneficial? But I do think if one is building the docs without even binary version of pyarrow then the need for python docs is not very high.

Yes, it certainly is less important (given that you need some python environment anyway, given that we are using sphinx to build the docs). So let's not worry about that use case here.

AlenkaF:
This PR should be ready now @jorisvandenbossche

docs/source/conf.py
jorisvandenbossche: > So yeah, we could default to an empty string if `ARROW_DOCS_VERSION` is not specified.

I think it would be nice to still use `pyarrow.__version__` _if_ it is installed, and the env variable is not set. It's nice to see a correct version number when building the docs locally for development (when you typically don't specify that env variable).

You could for example use `pyarrow_version =  pyarrow.__version` and `pyarrow_version = ""` in the if/else branches above when trying to import pyarrow, and then use that variable here (`version = os.environ.get('ARROW_DOCS_VERSION', pyarrow_version)`)

jorisvandenbossche (approved):

PR #35920, opened 2023-06-05
GH-18547: [Java] Support re-emitting dictionaries in ArrowStreamWriter

adamreeve:
### Rationale for this change

This allows writing IPC streams where dictionary values change between record batches.

* Add new abstract `void ensureDictionariesWritten(DictionaryProvider provider, Set<Long> dictionaryIdsUsed)` to the base `ArrowWriter` class
* Move existing logic that only writes dictionaries once into the `ArrowFileWriter` class
* Implement replacement dictionary writing in `ArrowStreamWriter` by keeping copies of previously written dictionaries

### Are these changes tested?

Yes, I've added a new unit test for this

Yes, `ArrowStreamWriter` will now write replacement dictionaries when dictionary values change between batches.

There may also be some performance impact on `ArrowStreamWriter` even when replacement dictionaries are not needed, due to taking an initial copy of dictionaries when writing the first batch and then comparing dictionaries for equality each time a new batch is written.

`ArrowWriter` has a new abstract `ensureDictionariesWritten` method. This will only affect users directly inheriting from  `ArrowWriter` rather than `ArrowFileWriter` or `ArrowStreamWriter`.

There's also a behaviour change to `ArrowWriter`, where previously dictionaries were read from a `DictionaryProvider` on construction, but this is now delayed until the first batch is written.
* Closes: #18547

### DIFF
--- a/java/tools/src/main/java/org/apache/arrow/tools/EchoServer.java
+++ b/java/tools/src/main/java/org/apache/arrow/tools/EchoServer.java
@@ -135,6 +135,7 @@ public class EchoServer {
         Preconditions.checkState(reader.bytesRead() == writer.bytesWritten());
         LOGGER.debug(String.format("Echoed %d records", echoed));
         reader.close(false);
+        writer.close();
       }
     }
 
--- a/java/vector/src/main/java/org/apache/arrow/vector/ipc/ArrowStreamWriter.java
+++ b/java/vector/src/main/java/org/apache/arrow/vector/ipc/ArrowStreamWriter.java
@@ -21,11 +21,18 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.channels.Channels;
 import java.nio.channels.WritableByteChannel;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 
+import org.apache.arrow.util.AutoCloseables;
+import org.apache.arrow.vector.FieldVector;
 import org.apache.arrow.vector.VectorSchemaRoot;
+import org.apache.arrow.vector.compare.VectorEqualsVisitor;
 import org.apache.arrow.vector.compression.CompressionCodec;
 import org.apache.arrow.vector.compression.CompressionUtil;
+import org.apache.arrow.vector.dictionary.Dictionary;
 import org.apache.arrow.vector.dictionary.DictionaryProvider;
 import org.apache.arrow.vector.ipc.message.IpcOption;
 import org.apache.arrow.vector.ipc.message.MessageSerializer;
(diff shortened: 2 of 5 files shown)


java/tools/src/main/java/org/apache/arrow/tools/EchoServer.java
```diff
@@ -135,6 +135,7 @@ public void run() throws IOException {
         Preconditions.checkState(reader.bytesRead() == writer.bytesWritten());
         LOGGER.debug(String.format("Echoed %d records", echoed));
         reader.close(false);
+        writer.close();
```
adamreeve, line 138: The `ArrowStreamWriter` not being closed meant that the copies of dictionary vectors it holds weren't being closed, so the allocator was throwing an `IllegalStateException` due to outstanding buffers. This is a behaviour change that users might need to watch out for.

adamreeve:
From what I can tell, the Flight code doesn't use `ArrowWriter` or `ArrowStreamWriter`, so I don't think the test failure for `org.apache.arrow.flight.TestLargeMessage.getLargeMessage` is related to my changes.

lidavidm:
@davisusanibar can you give this a look?

java/tools/src/main/java/org/apache/arrow/tools/EchoServer.java
lidavidm: Thanks. I think it's reasonable to expect that you must close the writer anyways (and that you should generally try-with-resources these writers).

java/vector/src/main/java/org/apache/arrow/vector/ipc/ArrowStreamWriter.java
```diff
@@ -121,4 +129,45 @@ public static void writeEndOfStream(WriteChannel out, IpcOption option) throws I
   protected void endInternal(WriteChannel out) throws IOException {
     writeEndOfStream(out, option);
   }
+
+  @Override
+  protected void ensureDictionariesWritten(DictionaryProvider provider, Set<Long> dictionaryIdsUsed)
```
lidavidm, line 134: Oof, so for each batch we write, we need to re-compare all dictionaries? That may also be worth noting in the PR description.

lidavidm: That said, this is broadly in line with the C++ implementation: https://github.com/apache/arrow/blob/766e25440250dde9117e19245389badb5ed7678c/cpp/src/arrow/ipc/writer.cc#L1121

The C++ implementation takes a number of shortcuts to avoid having to compare the full data buffers on each batch. However, because vectors are mutable in the Java implementation, we can't easily make the same optimization here.

lidavidm (approved):
LGTM. Thank you!

lidavidm:
The failing check is a long-standing flaky test that I disabled recently

java/vector/src/main/java/org/apache/arrow/vector/ipc/ArrowStreamWriter.java
adamreeve: Yeah I don't think there's a way to avoid this due to needing to account for mutable vectors, I've added a comment about this in the PR description.

PR #35930, opened 2023-06-06
GH-35926: [C++][Parquet] PageIndex allow only building Offset Index

### LINKED ISSUE #35926: [C++][Parquet] Parquet PageIndex allow only building Offset Index

Currently, if `page_index` is enabled, the Offset-index and Column-index would be built.

However, sometimes, user only need offset-index to find the pages, and statistics is not useful, we're sure these column has bad distribution and better not collect there stats. So, here we can allow only build offset index.

### Component(s)

C++, Parquet

mapleFU:
### Rationale for this change

Allow only build Offset Index for some column

Add a property `enable_column_index`.

### Are these changes tested?

Yes

User can disable column index for some column

* Closes: #35926

### DIFF
--- a/cpp/src/parquet/properties.h
+++ b/cpp/src/parquet/properties.h
@@ -139,7 +139,7 @@ static constexpr Encoding::type DEFAULT_ENCODING = Encoding::PLAIN;
 static const char DEFAULT_CREATED_BY[] = CREATED_BY_VERSION;
 static constexpr Compression::type DEFAULT_COMPRESSION_TYPE = Compression::UNCOMPRESSED;
 static constexpr bool DEFAULT_IS_PAGE_INDEX_ENABLED = false;
-static constexpr bool DEFAULT_IS_COLUMN_INDEX_ENABLED = false;
+static constexpr bool DEFAULT_IS_COLUMN_INDEX_ENABLED = true;
 
 class PARQUET_EXPORT ColumnProperties {
  public:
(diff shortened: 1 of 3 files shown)


mapleFU:
cc @wgtmac @pitrou @emkornfield

cpp/src/parquet/properties.h
```diff
@@ -147,14 +148,16 @@ class PARQUET_EXPORT ColumnProperties {
...
+                   bool page_index_enabled = DEFAULT_IS_PAGE_INDEX_ENABLED,
+                   bool column_index_enabled = DEFAULT_IS_COLUMN_INDEX_ENABLED)
       : encoding_(encoding),
         codec_(codec),
         dictionary_enabled_(dictionary_enabled),
         statistics_enabled_(statistics_enabled),
         max_stats_size_(max_stats_size),
         compression_level_(Codec::UseDefaultCompressionLevel()),
-        page_index_enabled_(DEFAULT_IS_PAGE_INDEX_ENABLED) {}
+        page_index_enabled_(page_index_enabled),
```
wgtmac, line 159: What about splitting `enable_page_index` into `enable_column_index` and `enable_offset_index`? This looks much cleaner to me. Though it may introduce a weird case where enable_column_index=true && enable_offset_index=false.

wgtmac: This feature is pretty new so I think it should be fine.

mapleFU:
> Statistics can already be disabled per-column, does it make sense to add this setting as well?

Oh yes, it can help. Should we support that or just disable statistics here?

pitrou:
> Oh yes, it can help. Should we support that or just disable statistics here?

Support what? I don't understand.

mapleFU:
Sorry for confusing, I means should I add a flag for only disable offset index? Or just disable statistic is enough. I guess disable statistics is suitable for my case

pitrou:
@wgtmac What do you think? Is there any case where one might want to enable statistics but disable the offset index?

mapleFU:
Emmm I'll close this first, I think `disable_statistics` already fixed my problem. If any one want's same, welcome to open it again

wgtmac:
> @wgtmac What do you think? Is there any case where one might want to enable statistics but disable the offset index?

I don't think so. If that's true, they can disable page index and use that from the page headers.

PR #35933, opened 2023-06-06
GH-35599: [Python] Canonical fixed-shape tensor extension array/type is not picklable.

### LINKED ISSUE #35599: [Python] Canonical fixed-shape tensor extension array/type is not picklable.

The [fixed-shape tensor extension type](https://arrow.apache.org/docs/python/extending_types.html#fixed-size-tensor) does not appear to be picklable. Given that pickling Arrow data is supported in general and is used in Python-centric systems such as Ray, supporting pickling for canonical extension types/arrays seems reasonable.

### Extension Type

```python
pickle.loads(pickle.dumps(pa.fixed_shape_tensor(pa.int64(), (2, 2))))
```
raises the error:
```
KeyError                                  Traceback (most recent call last)
File .../venv/lib/python3.9/site-packages/pyarrow/types.pxi:4798, in pyarrow.lib.type_for_alias()

KeyError: 'extension<arrow.fixed_shape_tensor>'
```

### Extension Array

```python
tensor_type = pa.fixed_shape_tensor(pa.int32(), (2, 2))
arr = [[1, 2, 3, 4], [10, 20, 30, 40], [100, 200, 300, 400]]
storage = pa.array(arr, pa.list_(pa.int32(), 4))
tensor_array = pa.ExtensionArray.from_storage(tensor_type, storage)
pickle.loads(pickle.dumps(tensor_array))
```
raises the ~same error:
```
KeyError                                  Traceback (most recent call last)
File .../venv/lib/python3.9/site-packages/pyarrow/types.pxi:4798, in pyarrow.lib.type_for_alias()

KeyError: 'extension<arrow.fixed_shape_tensor>'

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
Cell In[13], line 1
----> 1 pickle.loads(pickle.dumps(tensor_array))

File .../venv/lib/python3.9/site-packages/pyarrow/types.pxi:4800, in pyarrow.lib.type_for_alias()

ValueError: No type alias for extension<arrow.fixed_shape_tensor>
```

## Environment

- pyarrow 12.0.0
- Python 3.9
- MacOS

## Possible Solution

It seems like we might be able to implement `__reduce__` on [`FixedShapeTensorType`](https://github.com/apache/arrow/blob/2d76d9a526f9827283bb7dfac60715b6ad4aec34/python/pyarrow/types.pxi#L1511C12-L1587) such that it uses the `__arrow_ext_serialize__` serialization protocol? E.g.
```python
def __reduce__(self):
    return type(self).__arrow_ext_deserialize__, (self.storage, self.__arrow_ext_serialize__())
```

### Component(s)

Python

AlenkaF:
This PR adds `__reduce__` method to the `FixedShapeTensorType`.
* Closes: #35599

### DIFF
--- a/python/pyarrow/types.pxi
+++ b/python/pyarrow/types.pxi
@@ -1586,6 +1586,10 @@ cdef class FixedShapeTensorType(BaseExtensionType):
     def __arrow_ext_class__(self):
         return FixedShapeTensorArray
 
+    def __reduce__(self):
+        return fixed_shape_tensor, (self.value_type, self.shape,
+                                    self.dim_names, self.permutation)
+
 
 cdef class PyExtensionType(ExtensionType):
     """
--- a/python/pyarrow/tests/test_extension_type.py
+++ b/python/pyarrow/tests/test_extension_type.py
@@ -1306,3 +1306,19 @@ def test_extension_to_pandas_storage_type(registered_period_type):
         # Check the usage of types_mapper
         result = table.to_pandas(types_mapper=pd.ArrowDtype)
         assert isinstance(result["ext"].dtype, pd.ArrowDtype)
+
+
+def test_tensor_type_is_picklable():
+    # GH-35599
+
+    expected_type = pa.fixed_shape_tensor(pa.int32(), (2, 2))
+    result = pickle.loads(pickle.dumps(expected_type))
+
+    assert result == expected_type
+
+    arr = [[1, 2, 3, 4], [10, 20, 30, 40], [100, 200, 300, 400]]
+    storage = pa.array(arr, pa.list_(pa.int32(), 4))
+    expected_arr = pa.ExtensionArray.from_storage(expected_type, storage)
+    result = pickle.loads(pickle.dumps(expected_arr))
+
+    assert result == expected_arr


python/pyarrow/types.pxi
```diff
@@ -1586,6 +1586,10 @@ cdef class FixedShapeTensorType(BaseExtensionType):
     def __arrow_ext_class__(self):
         return FixedShapeTensorArray
 
+    def __reduce__(self):
```
AlenkaF, line 1589: I have started with implementing the `__reduce__` method for `FixedShapeTensorType`. I have to think about how to/if it is possible to use the extension type parameters when defining the method on the `BaseExtensionType`.

AlenkaF: As the logic for the `__arrow_ext_deserialize__` of the `FixedShapeTensorType` is happening in C++ we can not use `__arrow_ext_deserialize__` and `__arrow_ext_serialize__` to define `__reduce__` here.

But we can do that for `ExtensionType class`. I have created a separate issue for it: https://github.com/apache/arrow/issues/36038

jorisvandenbossche (approved):
Thanks!

pitrou:
Instead of defining `__reduce__` for each concrete extension type, why not define it on `BaseExtensionType` by making of use of C++-level (de)serialization?

jorisvandenbossche:
We actually did that somewhat as a follow-up in https://github.com/apache/arrow/pull/36170, defining `__reduce__` on `ExtensionType` (which can be moved to `BaseExtensionType`), but with using the Python APIs instead of the C++ APIs (so using `self.__arrow_ext_deserialize__` instead of `self.ext_type.Deserialize`), which then requires those dunders to be implemented (even when it is wrapping one defined in C++). This can indeed be cleaned up.

pitrou:
A cleanup would nice indeed, to avoid such [spurious additions](https://github.com/apache/arrow/pull/38008/files#diff-9f6b1034bb57bace094700a1b4748c88e40c9f98aec94a959ca466b5cbce0401).

PR #35944, opened 2023-06-06
GH-36168: [C++][Python] Support halffloat for Arrow list to pandas

### LINKED ISSUE #36168: [Python] Support halffloat for Arrow list to pandas

When we translate `float16` from `pyarrow` to `pandas`, no error occurs:
```
>>> df = pd.DataFrame({"a": [np.float16(1), np.float16(2), np.float16(3)]})
>>> table = pa.Table.from_pandas(df)
>>> df_new = table.to_pandas()
```
But, when we use list:
```
>>> df = pd.DataFrame({"a": [[np.float16(1)], [np.float16(2)], [np.float16(3)]]})
>>> table = pa.Table.from_pandas(df)
>>> df_new = table.to_pandas()
pyarrow.lib.ArrowNotImplementedError: Not implemented type for Arrow list to pandas: halffloat
```

### Component(s)

Python

izveigor:
### Rationale for this change
When we translate `float16` from `pyarrow` to `pandas`, no error occurs:
```
>>> df = pd.DataFrame({"a": [np.float16(1), np.float16(2), np.float16(3)]})
>>> table = pa.Table.from_pandas(df)
>>> df_new = table.to_pandas()
```
But, when we use list:
```
>>> df = pd.DataFrame({"a": [[np.float16(1)], [np.float16(2)], [np.float16(3)]]})
>>> table = pa.Table.from_pandas(df)
>>> df_new = table.to_pandas()
pyarrow.lib.ArrowNotImplementedError: Not implemented type for Arrow list to pandas: halffloat
```

### Are these changes tested?
Yes

Yes

* Closes: #36168


AlenkaF:
Thanks for the ping @izveigor!
Can we add the tests to the python side? Also, I think this would then need an issue opened and linked.

AlenkaF (approved):
Thank you! LGTM

python/pyarrow/tests/test_array.py
```diff
@@ -3347,6 +3347,17 @@ def test_to_pandas_timezone():
...
 
 
+@pytest.mark.pandas
+def test_to_pandas_float16_list():
+    # https://github.com/apache/arrow/issues/36168
+    arr = pa.array([[np.float16(1)], [np.float16(2)], [np.float16(3)]])
+    s = arr.to_pandas()
+    assert s is not None
+    base_refcount = sys.getrefcount(s.values.base)
+    assert base_refcount == 2
```
jorisvandenbossche, line 3357: What's the reason for checking the ref count?

I think you can create the expected result using numpy: `expected = np.array([1, 2, 3], dtype="float16")`, and then check that the result is the same.

jorisvandenbossche (commented):
Thanks for the PR!

python/pyarrow/tests/test_array.py
```diff
@@ -3347,6 +3347,15 @@ def test_to_pandas_timezone():
     assert s.dt.tz is not None
 
 
+@pytest.mark.pandas
+def test_to_pandas_float16_list():
+    # https://github.com/apache/arrow/issues/36168
+    expected = [[np.float16(1)], [np.float16(2)], [np.float16(3)]]
+    arr = pa.array(expected)
+    result = arr.to_pandas()
+    assert result.tolist() == expected
```
jorisvandenbossche, line 3356: Small comment: this conversion to list and comparing list equality doesn't actually check that the correct dtype was created (the individual elements are still equal, even when they would for example be float16 vs float32), so maybe we can still check the actual dtype, something like `result[0].dtype == "float16"`

jorisvandenbossche (approved):
Thanks!

PR #35950, opened 2023-06-06
GH-35948: [Go] Only cast `int8` and `unit8` to `float64` when JSON marshaling arrays

### LINKED ISSUE #35948: [Go] Marshaling `int64` or `unit64` arrays to JSON can cause loss of data

When marshaling `array.Int64` or `array.Uint64` to JSON the original `int64` or `unit64` values can be lost.
The cause is that the values are casted to `float64` before marshaling, and that type is not big enough to hold all possible `int64` or `unit64` values.

Relevant code:
https://github.com/apache/arrow/blob/4ec231b23432e9af340ed1a10d914f9bee84aca7/go/arrow/array/numeric.gen.go#L104
https://github.com/apache/arrow/blob/4ec231b23432e9af340ed1a10d914f9bee84aca7/go/arrow/array/numeric.gen.go#L199

Adding the following tests to [`numeric_test.go`](https://github.com/apache/arrow/blob/4ec231b23432e9af340ed1a10d914f9bee84aca7/go/arrow/array/numeric_test.go) result in failures:

```go
func TestInt64MarshalJSON(t *testing.T) {
	pool := memory.NewCheckedAllocator(memory.NewGoAllocator())
	defer pool.AssertSize(t, 0)

	var (
		vs = []int64{-5474557666971701248}
	)

	b := array.NewInt64Builder(pool)
	defer b.Release()

	for _, v := range vs {
		b.Append(v)
	}

	arr := b.NewArray().(*array.Int64)
	defer arr.Release()

	jsonBytes, err := json.Marshal(arr)
	if err != nil {
		t.Fatal(err)
	}
	got := string(jsonBytes)
	want := `[-5474557666971701248]`
	if got != want {
		t.Fatalf("got=%s, want=%s", got, want)
	}
}

func TestUInt64MarshalJSON(t *testing.T) {
	pool := memory.NewCheckedAllocator(memory.NewGoAllocator())
	defer pool.AssertSize(t, 0)

	var (
		vs = []uint64{14697929703826477056}
	)

	b := array.NewUint64Builder(pool)
	defer b.Release()

	for _, v := range vs {
		b.Append(v)
	}

	arr := b.NewArray().(*array.Uint64)
	defer arr.Release()

	jsonBytes, err := json.Marshal(arr)
	if err != nil {
		t.Fatal(err)
	}
	got := string(jsonBytes)
	want := `[14697929703826477056]`
	if got != want {
		t.Fatalf("got=%s, want=%s", got, want)
	}
}
```

A possible solution is to only do the casting to `float64` when the array is of `unit8` like the comment says, or skip the casting for `int64` and `unit64` arrays

### Component(s)

Go

erezrokah:
### Rationale for this change

See [issue](https://github.com/apache/arrow/issues/35948).

Based on https://github.com/apache/arrow/blob/9fb8697dcb442f63317c7d6046393fb74842e0ae/go/arrow/array/numeric.gen.go.tmpl#L120 it seems we should only do the casting for `unit8` and `int8` types.

This PR fixes the numeric template so it only casts `int8` and `unit8` to `float64` when JSON marshaling arrays to avoid losing data for other types.

### Are these changes tested?

Yes

The user facing change is that if a user was JSON marshaling arrays that has `uint64` or `int64` values they can now get different (more accurate) values than before 

* Closes: #35948

### DIFF
--- a/go/arrow/array/numeric.gen.go
+++ b/go/arrow/array/numeric.gen.go
@@ -101,9 +101,7 @@ func (a *Int64) MarshalJSON() ([]byte, error) {
 	vals := make([]interface{}, a.Len())
 	for i := 0; i < a.Len(); i++ {
 		if a.IsValid(i) {
-
 			vals[i] = a.values[i]
-
 		} else {
 			vals[i] = nil
 		}
@@ -198,9 +196,7 @@ func (a *Uint64) MarshalJSON() ([]byte, error) {
 	vals := make([]interface{}, a.Len())
 	for i := 0; i < a.Len(); i++ {
 		if a.IsValid(i) {
-
 			vals[i] = a.values[i]
-
 		} else {
 			vals[i] = nil
 		}
@@ -295,9 +291,7 @@ func (a *Float64) MarshalJSON() ([]byte, error) {
 	vals := make([]interface{}, a.Len())
 	for i := 0; i < a.Len(); i++ {
 		if a.IsValid(i) {
-
 			vals[i] = a.values[i]
-
 		} else {
 			vals[i] = nil
 		}
@@ -392,9 +386,7 @@ func (a *Int32) MarshalJSON() ([]byte, error) {
 	vals := make([]interface{}, a.Len())
 	for i := 0; i < a.Len(); i++ {
 		if a.IsValid(i) {
-
 			vals[i] = a.values[i]
-
 		} else {
 			vals[i] = nil
 		}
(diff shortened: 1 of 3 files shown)


zeroshade:
@erezrokah The failing test is indeed a flaky test, I haven't yet been able to reproduce it locally in order to fix it which annoys me. But you're fine.

zeroshade (approved):
LGTM! Thanks much for this!

erezrokah:
Thanks for the quick review and merge!

PR #35953, opened 2023-06-06
GH-35935: [C++] Clean interruption of a Acero plan with `use_threads=false`

### LINKED ISSUE #35935: [C++] Clean interruption of a Acero plan with `use_threads=false`

In internal experiments, we observed that an Acero plan with `use_threads=false` may not fully respond to an interruption. In this case, even though the signal is handled, the source nodes continue to push batches until done. Part of this issue is to devise testing for this condition.

### Component(s)

C++

rtpsw:

The execution plan set-up code is refactored to call `StopProducing` on the plan in the reader's `Close`. Originally, `StopProducing` was called from the destructor of `BatchConverter`, but with `use_threads=false` a hang occurs prior to this point in the code.

### Are these changes tested?

Yes - a test that early-closes a plan reader.

Yes; this will fix a hang visible to users.

* Closes: #35935


rtpsw:
cc @icexelloss @westonpace

rtpsw:
In an internal experiment, I observed that the changes here and along with those in #35902 fixed the issue. I'd like to get some feedback on what tests to add in order to verify this. @westonpace, WDYT?

icexelloss:
@rtpsw Can you add what is the rationale for this PR?

cpp/src/arrow/acero/exec_plan.cc
```diff
@@ -997,9 +1001,11 @@ struct BatchConverter {
   AsyncGenerator<std::optional<ExecBatch>> exec_batch_gen;
   std::shared_ptr<Schema> schema;
   std::shared_ptr<ExecPlan> exec_plan;
+  StopSource stop_source;
+  StopToken stop_token = stop_source.token();
 };
 
-Result<AsyncGenerator<std::shared_ptr<RecordBatch>>> DeclarationToRecordBatchGenerator(
+Result<std::shared_ptr<BatchConverter>> DeclarationToBatchConverter(
```
icexelloss, line 1008: What is batch converter? Why do we change the API here?

cpp/src/arrow/acero/exec_plan.cc
```diff
@@ -1025,21 +1031,31 @@ Result<std::unique_ptr<RecordBatchReader>> DeclarationToReader(Declaration decla
...
+          : arrow::internal::SerialExecutor::IterateGenerator<
+                std::shared_ptr<RecordBatch>>(std::move(make_gen), &ser_exec));
 
   struct PlanReader : RecordBatchReader {
-    PlanReader(std::shared_ptr<Schema> schema,
-               std::unique_ptr<Iterator<std::shared_ptr<RecordBatch>>> iterator)
-        : schema_(std::move(schema)), iterator_(std::move(iterator)) {}
+    PlanReader(std::shared_ptr<Schema> schema, std::shared_ptr<BatchConverter> converter,
+               std::unique_ptr<Iterator<std::shared_ptr<RecordBatch>>> iterator,
+               arrow::internal::SerialExecutor* ser_exec)
```
icexelloss, line 1054: Why does plan reader takes a SerialExecutor? What if this is not serial execution?

cpp/src/arrow/acero/exec_plan.h
```diff
@@ -602,6 +603,9 @@ struct ARROW_ACERO_EXPORT QueryOptions {
   /// If this field is not set then it will be treated as kWarn unless overridden
   /// by the ACERO_ALIGNMENT_HANDLING environment variable
   std::optional<UnalignedBufferHandling> unaligned_buffer_handling;
+
+  /// \brief An optional stop-token for the query. Defaults to unstoppable.
+  StopToken stop_token = StopToken::Unstoppable();
```
icexelloss, line 608: This API feels strange. Why would we want the user to pass a stop token in QueryOptions?

rtpsw: What do you mean by user? If you mean end-user, then the end-user does not need to change this, because it is set by higher-level APIs. If you mean caller, then this options allows the caller to signal stopping via the stop-token. Specifically, [a stop is requested on the stop-source](https://github.com/apache/arrow/pull/35953/files#diff-70e843181eb977ed11ad131f79a0f54a0bf9e30a0cb752d440658ee0cc16cc34R962) from which [this stop-token was created](https://github.com/apache/arrow/pull/35953/files#diff-70e843181eb977ed11ad131f79a0f54a0bf9e30a0cb752d440658ee0cc16cc34R1005); then, [the stop-token is passed to the scheduler](https://github.com/apache/arrow/pull/35953/files#diff-70e843181eb977ed11ad131f79a0f54a0bf9e30a0cb752d440658ee0cc16cc34R194) which will detect the signal via the stop-token.

This approach was suggested by @westonpace, so he may be able to say more about why.

cpp/src/arrow/util/thread_pool.h
```diff
@@ -301,8 +301,10 @@ class ARROW_EXPORT SerialExecutor : public Executor {
   /// approach is to use a stop token to cause the generator to exhaust early.
   template <typename T>
   static Iterator<T> IterateGenerator(
-      internal::FnOnce<Result<std::function<Future<T>()>>(Executor*)> initial_task) {
+      internal::FnOnce<Result<std::function<Future<T>()>>(Executor*)> initial_task,
+      internal::SerialExecutor** serial_executor_out = NULLPTR) {
```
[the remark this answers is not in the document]
rtpsw: It is a bit strange. At least so far I didn't find a better way to expose, given that the [`SerialExecutor` constructor is private](https://github.com/apache/arrow/blob/daacbcc4c5f0e435b2158896584a4385dbf38986/cpp/src/arrow/util/thread_pool.h#L370-L371). I needed to expose it to [send the stop-token to it](https://github.com/apache/arrow/pull/35953/files#diff-70e843181eb977ed11ad131f79a0f54a0bf9e30a0cb752d440658ee0cc16cc34R1076).

cpp/src/arrow/acero/exec_plan.cc
rtpsw: The SerialExecutor is needed for a clean interruption in case of `use_threads=false` (see other posts here for more details). If the plan execution is not serial then [it is null](https://github.com/apache/arrow/pull/35953/files#diff-70e843181eb977ed11ad131f79a0f54a0bf9e30a0cb752d440658ee0cc16cc34R1043) and not used.

icexelloss:
> @rtpsw Not sure I understand what is going on here. 

> In this case, even though the signal is handled, the source nodes continue to push batches until done.
Can you explain what is the cause of this?

cpp/src/arrow/acero/exec_plan.h
icexelloss: > If you mean caller, then this options allows the caller to signal stopping via the stop-token. 
Why does the caller need to signal stopping via the stop-token? Doesn't https://github.com/apache/arrow/blob/main/cpp/src/arrow/acero/exec_plan.h#L107 already allow user to stop the execution?

cpp/src/arrow/acero/exec_plan.cc
rtpsw: [The batch converter](https://github.com/apache/arrow/pull/35953/files#diff-70e843181eb977ed11ad131f79a0f54a0bf9e30a0cb752d440658ee0cc16cc34R947) is mainly responsible for [converting exec-batches to record-batches](https://github.com/apache/arrow/pull/35953/files#diff-70e843181eb977ed11ad131f79a0f54a0bf9e30a0cb752d440658ee0cc16cc34R965). We change the (note: internal) API here because [this new `IterateGenerator`-invoking code-branch](https://github.com/apache/arrow/pull/35953/files#diff-70e843181eb977ed11ad131f79a0f54a0bf9e30a0cb752d440658ee0cc16cc34R1048-R1049), which is specific for the `use_threads=false` case, needs access to `make_gen`, which in turn needs access to the batch converter. With this refactoring, it turned out that `DeclarationToBatchConverter` was the only API needed, so it replaced `DeclarationToRecordBatchGenerator`, and the generator was constructed on top of it.

rtpsw:
> Can you explain what is the cause of this?

For a detailed answer, I'd have to repro the experiment and investigate. I suspect that, pre-PR, the scheduler got stuck before `StopProducing` could get called.

cpp/src/arrow/acero/exec_plan.h
icexelloss: Hmm..OK. It feels strange that StopProducing does not work with Serial execution and we need a different way to stop an execution with serial execution. I would like to understand (1) why serial execution doesn't work with StopProducing now and (2) why we cannot make `StopProducing` to work (with serial execution.)

rtpsw: I imagine it's possible to make `StopProducing` be sufficient to shut down the plan even if it is serially executing, based on the facilities in this PR. However, there may be unforeseen interactions so I'd suggest deferring this to a separate discussion/issue.

icexelloss:
Typo: “There is a couple of PR/discussion”

cpp/src/arrow/acero/exec_plan.h
rtpsw: @westonpace, the approach here is from an offline discussion with you that was active just a few days ago, and in my experiments `StopProducing` wasn't enough. What progress with `StopProducing` are you referring to? What is the approach you are currently in favor of? If it ends up being straightforward, then I'd be in favor too, but I'd like to first validate this is the case.

westonpace: If I recall the problem with `StopProducing` was that sometimes we were still seeing a hang even after `StopProducing` was called.  However, this was later revealed to be the "paused source node doesn't respect stop producing" problem.  Beyond that everything seemed to work correctly with `StopProducing`.

rtpsw:
> I think we should consolidate the discussion first, figure out what is needed to be done

The situation is that we have #35902 as a fix related to `StopProducing`, as [mentioned by Weston](https://github.com/apache/arrow/pull/35953/#discussion_r1221591136), and the question is [whether it is sufficient for a clean interruption](https://github.com/apache/arrow/pull/35953/#discussion_r1222178918) or otherwise this PR is needed. I'll get back with an answer for this later. In the meantime, this PR is on hold.

cpp/src/arrow/acero/exec_plan.h
icexelloss: Thanks @westonpace for chiming in. I would agree with you here.

icexelloss:
@rtpsw Do you mean that `StopProducing` of the source node is not hit?

rtpsw:
> @rtpsw Do you mean that `StopProducing` of the source node is not hit?

Correct. I have now confirmed that it is not hit even in the version based on this PR. So, the next step should be to make sure it is called (in the internal experiment) and then see what the conclusion is.

rtpsw:
I got the same results (#35902 isn't sufficient for clean interruption but this PR is) when using an earlier version of Arrow as the base. This is good evidence that the cause the results are correct. If this PR is not favored, there should be some middle-point between #35902 and this PR that would both work and be favored. @westonpace, any insights about this possible middle-point?

icexelloss:
@rtpsw Can you provide more details on what doesn't work after applying #35902?

rtpsw:
We opted to switch to an approach in which the plan is stopped directly, when the plan reader closes. @westonpace, I believe this is along the lines you suggested.

cpp/src/arrow/util/thread_pool.h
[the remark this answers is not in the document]
rtpsw: The `StopProducing` approach ([noted here](https://github.com/apache/arrow/pull/35953/#issuecomment-1589892510)) worked in an internal experiment. Side note: the pre-PR version also had a call to `StopProducing`, in the destructor of `BatchConverter`, but it was too late - a hang happens earlier.

cpp/src/arrow/acero/exec_plan.h
icexelloss: @rtpsw Should we revert this change?

rtpsw: You're right.

cpp/src/arrow/util/thread_pool.h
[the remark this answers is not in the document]
icexelloss: If I read the code correctly - it seems like BatchConverter destructor is called when the PlanReader gets destructed (since it is a member of the batch_iterator which is a member of the PlanReader) (related code: https://github.com/apache/arrow/blob/e53db939bfad2f20e332172ab4f453add1dc680d/cpp/src/arrow/acero/exec_plan.cc#L1028)

This means that the destructor doesn't gets called during `Close()` / `iterator_.reset()`, which would explain why it didn't work before. @westonpace Does that sound correct?

cpp/src/arrow/acero/exec_plan.h
rtpsw: Done.

cpp/src/arrow/util/thread_pool.h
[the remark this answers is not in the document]
rtpsw: That sounds right though an additional point is that `Close` hanged in the loop for draining the remaining batches, so the destructor was never reached.

icexelloss:
I haven’t reviewed yet - Can you update the description?

icexelloss:
Can you also please add a test?

cpp/src/arrow/acero/exec_plan.cc
```diff
@@ -951,7 +951,6 @@ struct BatchConverter {
     if (exec_plan->finished().is_finished()) {
       return;
     }
-    exec_plan->StopProducing();
```
icexelloss, line 954: Why removing this?

cpp/src/arrow/acero/exec_plan.h
```diff
@@ -32,6 +32,7 @@
 #include "arrow/compute/exec.h"
 #include "arrow/compute/ordering.h"
 #include "arrow/type_fwd.h"
+#include "arrow/util/cancel.h"
```
icexelloss, line 35: Revert?

cpp/src/arrow/acero/exec_plan.cc
```diff
@@ -1001,18 +1000,24 @@ struct BatchConverter {
 
 Result<AsyncGenerator<std::shared_ptr<RecordBatch>>> DeclarationToRecordBatchGenerator(
     Declaration declaration, QueryOptions options,
-    ::arrow::internal::Executor* cpu_executor, std::shared_ptr<Schema>* out_schema) {
+    ::arrow::internal::Executor* cpu_executor, std::shared_ptr<Schema>* out_schema,
+    std::shared_ptr<ExecPlan>* out_plan) {
```
icexelloss, line 1004: Please document this argument

icexelloss (changes requested):
Reviewed with comments.

Please update the PR description and add a test (calls Reader.Close() to stop an execution)

icexelloss:
FYI I am removing "This PR contains a Critical Fix" because I don't think this applies to one of the below:

"
(a) a security vulnerability, (b) a bug that caused incorrect or invalid data to be produced, or (c) a bug that causes a crash (even when the API contract is upheld)
"

rtpsw:
In the [recent commit](https://github.com/apache/arrow/pull/35953/commits/15cc16873aea9d5270f8b8ae12650d478afa8b51), we ended up moving the `StopProducing` call from the destructor of `BatchConverter`, which is too late due to a hang occurring earlier, into `PlanReader::Close` before the draining loop, which required exposing the plan to the reader.

cpp/src/arrow/acero/exec_plan.h
rtpsw: Done.

cpp/src/arrow/acero/exec_plan.cc
rtpsw: It's moved, not removed. See [this post](https://github.com/apache/arrow/pull/35953/#issuecomment-1594187748).

rtpsw:
> I haven’t reviewed yet - Can you update the description?

Done.

rtpsw:
> FYI I am removing "This PR contains a Critical Fix" because I don't think this applies to one of the below:
> 
> " (a) a security vulnerability, (b) a bug that caused incorrect or invalid data to be produced, or (c) a bug that causes a crash (even when the API contract is upheld) "

The bug causes a hang, which is even worse than a crash, when the API contract is upheld. Whatever you decide.

cpp/src/arrow/acero/exec_plan.cc
rtpsw: What is the policy for documenting arguments? Most arguments are not documented, and in many cases their meaning is clear. I think this is the case here too.

rtpsw:
> Please update the PR description and add a test (calls Reader.Close() to stop an execution)

Done.

rtpsw:
> Can you also please add a test?

Done.

cpp/src/arrow/acero/exec_plan.cc
icexelloss: I am not sure about this. It's not clear to me what's the intent of calling StopProducing here in the first place. And `reader->Close()` isn't called in all cases so I think it is safer to leave this unchanged.

cpp/src/arrow/acero/plan_test.cc
```diff
@@ -767,6 +767,24 @@ TEST(ExecPlanExecution, DeclarationToReader) {
...
+  ASSERT_OK_AND_ASSIGN(std::unique_ptr<RecordBatchReader> reader,
+                       DeclarationToReader(plan, /*use_threads=*/false));
+
+  // read only a few batches
+  for (size_t i = 0; i < 10; i++) {
+    ASSERT_OK(reader->Next());
+  }
+  // then close
+  EXPECT_RAISES_WITH_MESSAGE_THAT(Cancelled, HasSubstr("Plan was cancelled early"),
+                                  reader->Close());
```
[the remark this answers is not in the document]
rtpsw: Reinstating this invocation requires a guarantee that `ExecPlan::StopProducing()` is idempotent, otherwise if `PlanReader::Close()` occurs then the invocation here which must occur as well may misbehave. I don't see this in guarantee in the documentation. @westonpace, WDYT?

cpp/src/arrow/acero/exec_plan.cc
rtpsw: Reinstating this invocation requires a guarantee that `ExecPlan::StopProducing()` is idempotent, otherwise if `PlanReader::Close()` occurs then the invocation here which must occur as well may misbehave. I don't see this in guarantee in the documentation. @westonpace, WDYT?

icexelloss: @rtpsw Can you try to figure out if ExecPlan::StopProducing() is idempotent

cpp/src/arrow/acero/plan_test.cc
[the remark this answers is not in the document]
icexelloss: If the expectation of "close" here is to "cancel the query gracefully", then I think the reasonable behaviors should be returning status OK if the cancellation is successful. I don't think raises an exception "Plan is cancelled" is the correct behavior then the user actually wanted to "cancel the plan".

cpp/src/arrow/acero/exec_plan.cc
icexelloss: I think it is worthwhile to at least explain that the `out_plan` is used for cancellation, otherwise this API looks confusion - it is not intuitive to return exec plan in a method called DeclarationToRecordBatchGenerator)?

cpp/src/arrow/acero/exec_plan.cc
rtpsw: Looking at `StopProducing`, it has an atomic-bool guarding against double-execution, I'll reinstate.

cpp/src/arrow/acero/plan_test.cc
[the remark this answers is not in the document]
rtpsw: The currently tested behavior exists pre-PR, so be aware that you're asking to change it. I don't know what the expectation in this case is because there is no documentation for `RecordBatchReader::Close()` and `PlanReader::Close()` that addresses this. There are other C++ reading APIs out there where a close operation is expected to report errors, possibly via a separate call ([like `std::ifstream::close()`](https://en.cppreference.com/w/cpp/io/basic_ifstream/close)), so I suspect people's expectations about this may vary. I'm fine with whatever behavior is in consensus, and in any case documentation of it should be added.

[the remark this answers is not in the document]
rtpsw: I actually included the fix you requested, so please take a look. In any case, I'm fine with whatever is decided.

icexelloss (approved):
LGTM

PR #35954, opened 2023-06-06
GH-35952: [R] Ensure that schema metadata can actually be set as a named character vector

### LINKED ISSUE #35952: [R] Ensure that schema metadata can actually be set as a named character vectory

Before #35860, setting schema metadata as a named character vector was possible. The changes in the R bindings made this error, as exposed in a reverse dependency check on the 12.0.1 release branch.

``` r

schema <- schema(x = int32())
schema$metadata <- c("name" = "value")
#> Error in metadata[["r"]]: subscript out of bounds
```

<sup>Created on 2023-06-06 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### Component(s)

R

paleolimbot:
This wasn't necessarily a regression (reprex fails in 12.0.0 as well), although the comments suggest that assigning a named character vector will work when assigning schema metadata and this feature appears to be used by at least one of our dependencies (sfarrow). Given that the sfarrow check passes on 12.0.0, there is possibly also a place in our code that returns a named character vector rather than a list. I've confirmed that this fix solves the reverse dependency failure building the sfarrow example vignette.

``` r

schema <- schema(x = int32())
schema$metadata <- c("name" = "value")
schema$metadata
#> $name
#> [1] "value"
```

<sup>Created on 2023-06-06 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>
* Closes: #35952

### DIFF
--- a/r/R/schema.R
+++ b/r/R/schema.R
@@ -229,9 +229,13 @@ prepare_key_value_metadata <- function(metadata) {
       call. = FALSE
     )
   }
+
+  metadata <- as.list(metadata)
+
   if (!is_empty(metadata) && is.list(metadata[["r"]])) {
     metadata[["r"]] <- .serialize_arrow_r_metadata(metadata[["r"]])
   }
+
   map_chr(metadata, as.character)
 }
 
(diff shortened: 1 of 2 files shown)


thisisnic (approved):
Fantastic, thanks!

PR #35955, opened 2023-06-06
GH-35949: [R] CSV File reader options class objects should print the selected values

### LINKED ISSUE #35949: [R] CSV File reader options class objects should print the selected values

This would make it much easier for debugging

### Component(s)

R

thisisnic:
Fixes #35949
* Closes: #35949

### DIFF
--- a/r/R/csv.R
+++ b/r/R/csv.R
@@ -466,12 +466,13 @@ CsvReadOptions <- R6Class("CsvReadOptions",
   public = list(
     encoding = NULL,
     print = function(...) {
-      cat("CsvReadOptions")
-      cat("\nencoding:", self$encoding)
-      cat("\ncolumn_names:", self$column_names)
-      cat("\nblock_size:", self$block_size)
-      cat("\nskip_rows:", self$skip_rows)
-      cat("\nautogenerate_column_names:", self$autogenerate_column_names)
+      cat("CsvReadOptions\n")
+      for (attr in c(
+        "column_names", "block_size", "skip_rows", "autogenerate_column_names",
+        "use_threads", "skip_rows_after_names", "encoding"
+      )) {
+        cat(sprintf("%s: %s\n", attr, self[[attr]]))
+      }
       invisible(self)
     }
   ),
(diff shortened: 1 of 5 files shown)


thisisnic:
I started on this as it'd be useful for some debugging I'm doing, though I wonder if there's a more efficient way of doing this than copying the code for each attribute.  Perhaps not?

r/R/csv.R
```diff
@@ -464,10 +464,22 @@ CsvTableReader$create <- function(file,
...
   public = list(
-    encoding = NULL
+    encoding = NULL,
+    print = function(...) {
+      cat("CsvReadOptions")
+      cat("\nencoding:", self$encoding)
+      cat("\ncolumn_names:", self$column_names)
+      cat("\nblock_size:", self$block_size)
+      cat("\nskip_rows:", self$skip_rows)
+      cat("\nautogenerate_column_names:", self$autogenerate_column_names)
```
paleolimbot, line 474: Probably a `for()` loop here?

```r
for (attr in c("encoding", ...)) {
  cat(sprintf("%s: %s\n", attr, self[[attr]]))
}
```

paleolimbot (commented):
Seems like a good idea to me! This will almost certainly save you or somebody else some time in the future.

r/R/csv.R
thisisnic: Thanks! I was also checking in that there's no way of doing the other bits more programmatically, but doesn't appear to be, so I'll just plough on!

r/tests/testthat/test-dataset-csv.R
```diff
@@ -574,3 +574,11 @@ test_that("open_delim_dataset params passed through to open_dataset", {
 
   expect_equal(ds$time, "16-01-2023")
 })
+
```
paleolimbot, line 577: One more test here for the data access part (as opposed to the printing part)? Like:

```r
test_that("CSV options field access", {
  options <- CsvReadOptions$create(skip_rows = 102)
  expect_identical(options$skip_rows, 102)
  # ... for all the options we just added the ability to access
})
```

...for each option we added a

paleolimbot (commented):
Awesome! Just one more test and it's good to go!

r/tests/testthat/test-dataset-csv.R
thisisnic: Added

paleolimbot (approved):
Thanks! I made a ticket for the CI failure, which seems to affect every debian/testing-based job: https://github.com/apache/arrow/issues/36030

PR #35956, opened 2023-06-06
MINOR: [Go] Add IsNull method to dictionary builder

metalmatze:
Given that this function is super similar to the previously added `GetValueIndex` I've decided to not open an Issue first.

### Rationale for this change

This is helpful to check if previously appended rows have been null or not.

### Are these changes tested?

The changes have been tested in our unit tests building on top of this. I'm happy to follow up with a unit test if deemed necessary. 

The addition of `IsNull(i int) bool` to the `BinaryDictionaryBuilder`.

### DIFF
--- a/go/arrow/array/dictionary.go
+++ b/go/arrow/array/dictionary.go
@@ -45,12 +45,12 @@ import (
 //
 // For example, the array:
 //
-//      ["foo", "bar", "foo", "bar", "foo", "bar"]
+//	["foo", "bar", "foo", "bar", "foo", "bar"]
 //
 // with dictionary ["bar", "foo"], would have the representation of:
 //
-//      indices: [1, 0, 1, 0, 1, 0]
-//      dictionary: ["bar", "foo"]
+//	indices: [1, 0, 1, 0, 1, 0]
+//	dictionary: ["bar", "foo"]
 //
 // The indices in principle may be any integer type.
 type Dictionary struct {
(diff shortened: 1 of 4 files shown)


go/arrow/array/dictionary.go
```diff
@@ -1293,6 +1293,29 @@ func (b *BinaryDictionaryBuilder) InsertStringDictValues(arr *String) (err error
...
+	case *Int32Builder:
+		return b.builder.nullBitmap.Len() != 0 && bitutil.BitIsNotSet(b.builder.nullBitmap.Bytes(), i)
+	case *Uint64Builder:
+		return b.builder.nullBitmap.Len() != 0 && bitutil.BitIsNotSet(b.builder.nullBitmap.Bytes(), i)
+	case *Int64Builder:
+		return b.builder.nullBitmap.Len() != 0 && bitutil.BitIsNotSet(b.builder.nullBitmap.Bytes(), i)
+	default:
+		return false
+	}
+}
```
zeroshade, line 1317: If you just add `IsNull` to the underlying `builder` struct in `builder.go` then *all* the builders will get it for free as they all embed the base builder. It also means you wouldn't need this switch here.

metalmatze: That's cool! Let me tackle that!

zeroshade (changes requested):
LGTM. can you please add some unit tests? After adding some tests, I'd be happy to merge this.

zeroshade (approved):

PR #35958, opened 2023-06-07
GH-35926: [C++][Parquet] Allow disabling ColumnIndex by disabling statistics

### LINKED ISSUE #35926: [C++][Parquet] Parquet PageIndex allow only building Offset Index

mapleFU:
When statistics are disabled for a column, then only OffsetIndex should be emitted, not ColumnIndex.
Also, fix a bug in the `ColumnProperties` constructor where `page_index_enabled` would not be forwarded correctly.

* Closes: #35926


mapleFU:
cc @wgtmac @pitrou

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -5350,6 +5352,30 @@ TEST_F(ParquetPageIndexRoundTripTest, SimpleRoundTrip) {
                             /*null_counts=*/{1}}));
 }
 
+TEST_F(ParquetPageIndexRoundTripTest, SimpleRoundTripWithStatsDisabled) {
+  auto writer_properties = WriterProperties::Builder()
+                               .enable_write_page_index()
+                               ->disable_statistics()
```
wgtmac, line 5358: Should we test this per column?

mapleFU: Added now

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -5370,12 +5370,53 @@ TEST_F(ParquetPageIndexRoundTripTest, SimpleRoundTripWithStatsDisabled) {
...
+    ])"}));
+
+  ReadPageIndexes(/*expect_num_row_groups=*/2, /*expect_num_pages=*/1);
+
+  ColumnIndexObject empty_column_index{};
+  empty_column_index.empty = true;
+  EXPECT_THAT(
+      column_indexes_,
+      ::testing::ElementsAre(
+          empty_column_index,
```
wgtmac, line 5404: Why add a new `empty` flag? Isn't `ColumnIndexObject{}` enough to represent a missing ColumnIndex?

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -5356,7 +5357,6 @@ TEST_F(ParquetPageIndexRoundTripTest, SimpleRoundTripWithStatsDisabled) {
   auto writer_properties = WriterProperties::Builder()
                                .enable_write_page_index()
                                ->disable_statistics()
-                               ->max_row_group_length(4)
```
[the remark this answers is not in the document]
mapleFU: No need to split rowgroup here, just ignore it

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
mapleFU: Just want to make clear that it's explicit built from a `nullptr`

wgtmac: That seems duplicate. A valid ColumnIndex must have at least one page. So `null_pages`, `min_values` and `max_values` will not be empty and share the same size if ColumnIndexObject is created from a non null ColumnIndex.

mapleFU: Updated

cpp/src/parquet/properties.h
```diff
@@ -154,7 +154,7 @@ class PARQUET_EXPORT ColumnProperties {
         statistics_enabled_(statistics_enabled),
         max_stats_size_(max_stats_size),
         compression_level_(Codec::UseDefaultCompressionLevel()),
-        page_index_enabled_(DEFAULT_IS_PAGE_INDEX_ENABLED) {}
+        page_index_enabled_(page_index_enabled) {}
```
wgtmac, line 157: Good catch!

wgtmac (approved):
LGTM

mapleFU:
@pitrou Mind take a look?

mapleFU:
@pitrou @wgtmac Do you still have any comments?

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -5350,6 +5350,70 @@ TEST_F(ParquetPageIndexRoundTripTest, SimpleRoundTrip) {
...
+      [2,     "b",  [1, 2]   ],
+      [3,     "c",  [null]   ],
+      [null,  "d",  []       ],
+      [5,     null, [3, 3, 3]],
+      [6,     "f",  null     ]
+    ])"}));
+
+  ReadPageIndexes(/*expect_num_row_groups=*/1, /*expect_num_pages=*/1);
+  for (auto& column_index : column_indexes_) {
+    // Means page is empty.
```
pitrou, line 5372: Which page is empty? I don't understand this comment.

pitrou: Why not test against `ColumnIndexObject{}` instead?

mapleFU: Sure, changed

pitrou (approved):
+1, thanks @mapleFU !

pitrou:
The remaining CI failures are unrelated, will merge.

PR #35966, opened 2023-06-07
GH-35965: [Go] Fix `Decimal256DictionaryBuilder`

### LINKED ISSUE #35965: `Decimal256DictionaryBuilder` does not work correctly

`Decimal256DictionaryBuilder` cannot append `decimal256` value and prefers `decimal128` instead.
```
cannot use v (variable of type decimal256.Num) as decimal128.Num value in argument to builder.Append
```
The problem is in this line: https://github.com/apache/arrow/blob/main/go/arrow/array/dictionary.go#L1376

### Component(s)

Go

izveigor:
### Rationale for this change
`Decimal256DictionaryBuilder` cannot append `decimal256` value and prefers `decimal128` instead.
```
cannot use v (variable of type decimal256.Num) as decimal128.Num value in argument to builder.Append
```

### Are these changes tested?
Yes

Yes

* Closes: #35965

### DIFF
--- a/go/arrow/array/dictionary.go
+++ b/go/arrow/array/dictionary.go
@@ -28,6 +28,7 @@ import (
 	"github.com/apache/arrow/go/v13/arrow"
 	"github.com/apache/arrow/go/v13/arrow/bitutil"
 	"github.com/apache/arrow/go/v13/arrow/decimal128"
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Thanks for catching this oversight!

PR #35976, opened 2023-06-07
MINOR: [Go] Use proper methods in compute

candiduslynx:
Blocks #35973
@zeroshade I believe this is a proper code (use Item instead of Value for map type).
I'll leave the filling in & the issue creation up to you.

### DIFF
--- a/go/arrow/compute/exprs/exec.go
+++ b/go/arrow/compute/exprs/exec.go
@@ -285,15 +285,15 @@ func literalToDatum(mem memory.Allocator, lit expr.Literal, ext ExtensionIDSet) 
 				return nil, fmt.Errorf("%w: key type mismatch for %s, got key with type %s",
 					arrow.ErrInvalid, mapType, scalarKey.DataType())
 			}
-			if !arrow.TypeEqual(mapType.ValueType(), scalarValue.DataType()) {
-				return nil, fmt.Errorf("%w: value type mismatch for %s, got key with type %s",
+			if !arrow.TypeEqual(mapType.ItemType(), scalarValue.DataType()) {
+				return nil, fmt.Errorf("%w: value type mismatch for %s, got value with type %s",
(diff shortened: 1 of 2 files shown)


candiduslynx:
@zeroshade after seeing this appear in th codebase once again, maybe it'll be better to remove the value x methods from map type completely, wdyt?
They clearly cause a lot of misunderstanding between usual key-value semantics & key-item in map type.

zeroshade:
@candiduslynx I just noticed this myself and merged #35983 to fix it before I saw this PR. 

I think it's fine for now to leave them as deprecated rather than breaking people. We can remove them after v13 is released.

candiduslynx:
> @candiduslynx I just noticed this myself and merged #35983 to fix it before I saw this PR.
> 
> I think it's fine for now to leave them as deprecated rather than breaking people. We can remove them after v13 is released.

@zeroshade kindly take a look at https://github.com/apache/arrow/pull/35983/files#r1222024655

zeroshade:
That said, it looks like I screwed up and used `Elem` when I should have used `Item` haha. Can you rebase this and then put back the fix to use Item on line 626 of `go/arrow/compute/exprs/types.go` ? I'll merge this after that.

candiduslynx:
> That said, it looks like I screwed up and used `Elem` when I should have used `Item` haha. Can you rebase this and then put back the fix to use Item on line 626 of `go/arrow/compute/exprs/types.go` ? I'll merge this after that.

That's why I've deprecated the `ValueX` method in the first place =)

candiduslynx:
@zeroshade I've solved the conflicts using GitHub UI, hope it's fine

zeroshade (approved):
Thanks for the catch! I'll merge after the CI passes.

candiduslynx:
> Thanks for the catch! I'll merge after the CI passes.

I'll merge the main into the #35973 (mem leak fix) afterwards, so you can take a look alreay

Issue #35975, opened 2023-06-07
[Go] C Data Interface can't import decimal256


lidavidm:
### Describe the enhancement requested

It's just not implemented: https://github.com/apache/arrow/blob/9be7074f85d6057a92bc008766e8f30d58b30cf7/go/arrow/cdata/cdata.go#L232-L238

### Component(s)

Go

PR #35978, opened 2023-06-07
GH-35974: [Go] Don't panic if importing C Array Stream fails

### LINKED ISSUE #35974: [Go] C Data Interface should import schema up front instead of panicking

Same issue that bit the C++ implementation: Schema() is infallible so it can't be implemented on top of C Data Interface (where get_schema is fallible) without panicking. We should import the schema up front and report the error properly.

### Component(s)

Go

lidavidm:
### Rationale for this change
Panicking is rude.

Import C Array Stream schemas up front and report the error.

### Are these changes tested?

Yes.

`ImportCArrayStream` (which cannot fail, only panic) is deprecated in favor of `ImportCRecordReader` (which can return an error).
* Closes: #35974

### DIFF
--- a/go/arrow/cdata/cdata.go
+++ b/go/arrow/cdata/cdata.go
@@ -717,7 +717,7 @@ func importCArrayAsType(arr *CArrowArray, dt arrow.DataType) (imp *cimporter, er
 	return
 }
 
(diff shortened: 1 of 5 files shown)


lidavidm:
Note a practical reason why we might panic on import is #35975

zeroshade (approved):
LGTM. thanks!

PR #35980, opened 2023-06-07
GH-35979: [C++] Refactor Acero scalar and hash aggregation into separate files

### LINKED ISSUE #35979: [C++] Refactor Acero scalar and hash aggregation into separate files

Refactor the existing aggregate_node.cc into separate files (scalar and group aggregate respectively)
 
### Component(s)

C++

ildipo:
### Rationale for this change

Some refactoring to simplify relations development and pave the way for implementing window aggregation.

Existing Acero aggregation (scalar and group-by) sources have been refactored into separate files with no changes.

* Closes: #35979


cpp/src/arrow/acero/CMakeLists.txt
```diff
@@ -29,7 +29,11 @@ endmacro()
 
 set(ARROW_ACERO_SRCS
     accumulation_queue.cc
-    aggregate_node.cc
```
ildipo, line 32: `aggregate_node.cc` was split into `scalar_aggregate_node.cc`, `groupby_aggregate_node.cc` and `aggregate_internal.cc` - the existing code was moved as-is

icexelloss:
Thanks @ildipo! 

@westonpace Since this is a fairly large PR I think it would be beneficial if reviewers get an understanding of high level design in some way (maybe a meeting? Or some shared doc?). WDYT?

Also @ianmcook IIRC you also mentioned interested in reviewing windowing design/functionality

cpp/src/arrow/acero/groupby_aggregate_node.cc
```diff
@@ -0,0 +1,447 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
ildipo, line 1: All of this was inside `aggregate_node.cc`

cpp/src/arrow/acero/scalar_aggregate_node.cc
```diff
@@ -0,0 +1,322 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
ildipo, line 1: All of this was inside `aggregate_node.cc`

westonpace:
@icexelloss happy to meet in person and go over the high level design.  Just from a glance it looks like the aggregate node was split into the scalar and group-by cases with common parts put into aggregate_internal (I'm guessing to help share code between the aggregate cases and the window aggregates).  Then two new window nodes were added.  It seems that both of these nodes rely on the InputReceived being called in a sorted order in the same way that the asof join node does?

ildipo:
@westonpace that is correct

ildipo:
I decided to split this into 2 to simplify review.
This now contains only refactoring and the additional options in the relations but not the window implementation.

cpp/src/arrow/acero/options.h
```diff
@@ -297,6 +297,26 @@ class ARROW_ACERO_EXPORT ProjectNodeOptions : public ExecNodeOptions {
   std::vector<std::string> names;
 };
 
+/// windows proceed left to right
+class ARROW_ACERO_EXPORT WindowAggregateArgs {
```
icexelloss, line 301: Revert this?

cpp/proto/substrait/extension_rels.proto
```diff
@@ -69,3 +69,4 @@ message SegmentedAggregateRel {
   // A list of one or more aggregate expressions along with an optional filter.
   repeated substrait.AggregateRel.Measure measures = 3;
 }
+
```
icexelloss, line 72: Nit: revert this change? (Since we don't touch this file at all probably cleaner if we don't leave a new line change)

icexelloss:
@ildipo LGTM. Minor comments on the substrait file (since we don't change it probably better not to leave untouched)

Lint issue probably need:
```
archery lint --cmake-format
```

Do you mind also updating the title and description to better reflect the reduced change set?

icexelloss (approved):
LGTM

icexelloss:
@westonpace I plan to merge this later Today if you don't have objections.

icexelloss:
@westonpace Since I didn't hear from you I went ahead and merged this. (No changes to external headers so we should be fine).

But let me know in case you have some more comments we can do it as a follow up.

PR #35983, opened 2023-06-07
GH-35982: [Go] Fix go1.18 broken builds

### LINKED ISSUE #35982: [Go] go1.18 builds broken with deprecated message

#35654 was merged without manually rebasing first so it didn't catch the change that deprecated usage of a few functions in it until after the merge and builds. Usage of `(*MapType).ValueType()` was deprecated which is causing the linting step to break all go1.18 builds (since we only build the compute on go1.18+)

### Component(s)

Go

zeroshade:
### Rationale for this change
Fix usage of deprecated functions to fix the Go 1.18 builds.

* Closes: #35982

### DIFF
--- a/go/arrow/compute/exprs/types.go
+++ b/go/arrow/compute/exprs/types.go
@@ -623,7 +623,7 @@ func ToSubstraitType(dt arrow.DataType, nullable bool, ext ExtensionIDSet) (type
 		if err != nil {
 			return nil, err
 		}
-		valueType, err := ToSubstraitType(dt.ValueType(), dt.ValueField().Nullable, ext)
+		valueType, err := ToSubstraitType(dt.Elem(), dt.ElemField().Nullable, ext)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


lidavidm (approved):

go/arrow/compute/exprs/types.go
```diff
@@ -623,7 +623,7 @@ func ToSubstraitType(dt arrow.DataType, nullable bool, ext ExtensionIDSet) (type
 		if err != nil {
 			return nil, err
 		}
-		valueType, err := ToSubstraitType(dt.ValueType(), dt.ValueField().Nullable, ext)
+		valueType, err := ToSubstraitType(dt.Elem(), dt.ElemField().Nullable, ext)
```
candiduslynx, line 626: This is wrong, as the value here is a value in `key-value pair` sense, so it has to be `Item` in `arrow.MapType`

PR #35989, opened 2023-06-08
GH-34351: [C++][Parquet] Statistics: add detail documentation and tiny optimization

### LINKED ISSUE #34351: [C++][Parquet] Statistics Merge ignore setting `has` flag

mapleFU:

1. This patch does some tiny optimizations on Parquet C++ Statistics. It does:

```
For min-max, using std::string. Because assume the case like that:
EncodedStatistics c1;
// do some operations
EncodedStatistics c2 = c1;
c2.set_max("dasdasdassd");
After c2 set, c1 would be set too. So I use std::string here.
```

2. Force clear ndv count during merging, and set `has_distinct_count_ = false`, and add some comments
3. Add some specification in Statistics API

### Are these changes tested?

Yes

No

* Closes: #34351


mapleFU:
Some notes:
1. Builder of `Statistics` will always have `has_null_count_`, and ignore `has_distinct_count_` (not build when building `EncodedStatistics`). `has_min_max_` will be `false` if page is all nulls or nulls and NaN
2. Reader will able to set `distinct_count` and `has_distinct_count`. Which would be valid, but when it call `Encoded`, the status will not be included in `EncodedStatistics`

mapleFU:
Another problem is that:

```c++
template <typename DType>
static std::shared_ptr<Statistics> MakeTypedColumnStats(
    const format::ColumnMetaData& metadata, const ColumnDescriptor* descr) {
  // If ColumnOrder is defined, return max_value and min_value
  if (descr->column_order().get_order() == ColumnOrder::TYPE_DEFINED_ORDER) {
    return MakeStatistics<DType>(
        descr, metadata.statistics.min_value, metadata.statistics.max_value,
        metadata.num_values - metadata.statistics.null_count,
        metadata.statistics.null_count, metadata.statistics.distinct_count,
        metadata.statistics.__isset.max_value || metadata.statistics.__isset.min_value,
        metadata.statistics.__isset.null_count,
        metadata.statistics.__isset.distinct_count);
  }
  // Default behavior
  return MakeStatistics<DType>(
      descr, metadata.statistics.min, metadata.statistics.max,
      metadata.num_values - metadata.statistics.null_count,
      metadata.statistics.null_count, metadata.statistics.distinct_count,
      metadata.statistics.__isset.max || metadata.statistics.__isset.min,
      metadata.statistics.__isset.null_count, metadata.statistics.__isset.distinct_count);
}
```

For the Arrow write file, this is ok, however, when `!metadata.statistics.__isset.null_count`, the `num_values` would be wrong. Currently I don't know how to fix it.

mapleFU:
cc @wgtmac @pitrou would you mind take a look? Since `Statistics` is released, I'm afraid my change will break some invariant...

cpp/src/parquet/statistics.h
```diff
@@ -149,10 +148,10 @@ class PARQUET_EXPORT EncodedStatistics {
   // the true minimum for aggregations and there is no way to mark that a
   // value has been truncated and is a lower bound and not in the page.
   void ApplyStatSizeLimits(size_t length) {
-    if (max_->length() > length) {
+    if (max_.length() > length) {
```
wgtmac, line 151: Should we clear `max_` as well?

cpp/src/parquet/statistics.cc
```diff
@@ -472,34 +474,40 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
     comparator_ = std::static_pointer_cast<TypedComparator<DType>>(comp);
     TypedStatisticsImpl::Reset();
     has_null_count_ = true;
-    has_distinct_count_ = true;
+    has_distinct_count_ = false;
```
wgtmac, line 477: nit: has_distinct_count_ is set to false in `TypedStatisticsImpl::Reset();`

cpp/src/parquet/statistics.cc
```diff
@@ -472,34 +474,40 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
...
   }
 
+  // Currently, this function will call by testing.
```
wgtmac, line 480:
```suggestion
  // Currently, this function will only be called by testing.
```

cpp/src/parquet/statistics.cc
```diff
@@ -472,34 +474,40 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
...
     TypedStatisticsImpl::IncrementNullCount(null_count);
-    IncrementDistinctCount(distinct_count);
+    SetDistinctCount(distinct_count);
 
     Copy(min, &min_, min_buffer_.get());
     Copy(max, &max_, max_buffer_.get());
     has_min_max_ = true;
   }
 
+  // Currently, this function will call by ColumnChunkMetaData during read.
```
wgtmac, line 495:
```suggestion
  // Currently, this function will only be called by ColumnChunkMetaData during read.
```

cpp/src/parquet/statistics.cc
```diff
@@ -552,12 +560,19 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
 
   void Merge(const TypedStatistics<DType>& other) override {
     this->num_values_ += other.num_values();
-    if (other.HasNullCount()) {
+    // Merge always runs when Merge builder's page statistics
+    // into column chunk statistics, so it tent to have null count.
+    if (ARROW_PREDICT_TRUE(other.HasNullCount())) {
```
wgtmac, line 565: Should we check this->has_null_count_ as well?

cpp/src/parquet/statistics.cc
```diff
@@ -552,12 +560,19 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
...
+    } else {
+      this->has_null_count_ = false;
     }
-    if (other.HasDistinctCount()) {
-      this->statistics_.distinct_count += other.distinct_count();
+    // Distinct count cannot be merged.
+    if (has_distinct_count_) {
+      has_distinct_count_ = false;
     }
+    // If !other.HasMinMax, might be all-nulls or nulls and nan,
```
wgtmac, line 574: We cannot differentiate at least two states:
- (1) the stats has not seen any value except null or nan, so it does not have min/max
- (2) the stats has seen some valid values but min/max are truncated due to size limit

So if currently the stats is in the state (2), we should not merge min/max from the incoming stats

mapleFU: The interesting thing is, currently, "merge" is only called by writer, and writer is always the case (1)...
@pitrou Do you have any idea here? Since `Statistics` is already public.

cpp/src/parquet/statistics.cc
```diff
@@ -552,12 +560,19 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
...
+    // Merge always runs when Merge builder's page statistics
+    // into column chunk statistics, so it tent to have null count.
+    if (ARROW_PREDICT_TRUE(other.HasNullCount())) {
       this->statistics_.null_count += other.null_count();
+    } else {
+      this->has_null_count_ = false;
     }
-    if (other.HasDistinctCount()) {
-      this->statistics_.distinct_count += other.distinct_count();
+    // Distinct count cannot be merged.
```
pitrou, line 570: I understand the reasoning, but this might break existing usage. Perhaps we should instead document that the distinct count can be an upper bound?

Can you use `git log` or `git blame` to find out who touched this code previously, and ping them to get their opinion?

cpp/src/parquet/statistics_test.cc
```diff
@@ -377,6 +377,35 @@ class TestStatistics : public PrimitiveTypedTest<TestType> {
     ASSERT_EQ(total->max(), std::max(statistics1->max(), statistics2->max()));
   }
 
+  void TestMergeEmpty() {
+    EncodedStatistics encoded_statistics1;
+    encoded_statistics1.has_distinct_count = false;
```
pitrou, line 382: This doesn't test the semantic change you're making in this PR, is it?

cpp/src/parquet/statistics.cc
mapleFU: Emmm currently writer **never** set `has_distinct_count_`, so I change it as disabled. In my opinion, user **should** explicit set `has_distinct_count_` when building, like:

```c++
FinishColumnChunk() {
  if(!fallback_) {
     set_distinct_count()
  }
}
```

The current `has_distinct_count_` is absolutely confusing...

cpp/src/parquet/statistics.cc
```diff
@@ -463,6 +463,8 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
  public:
   using T = typename DType::c_type;
 
+  // Currently, this function will be called by ColumnWriter to create the
+  // statistics collector during write.
```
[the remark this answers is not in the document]
mapleFU: I guess it's ok but it makes cleaning the statistics much more complex...

cpp/src/parquet/statistics.cc
mapleFU: https://github.com/apache/arrow/commit/43c7154894f819e617dbf3d669dbe1185f05893b#diff-f8a090a40e214aba5ebff534411a9febf213b725597afb750395d2a7483d9bc1

I guess these code is written in 2016 and never changed the syntax for add distinct count. Maybe it's designed like dictionary in Arrow IPC Streaming?

cpp/src/parquet/statistics.cc
mapleFU: @pitrou As Gang mentioned here

1. In theory, `has_null_count_` is not added able if another statistics has `has_null_count_ = true`
2. However, in our implementions, builder will always set `has_null_count_ == true`

Maybe I can use the same way to treat them?

cpp/src/parquet/statistics.cc
pitrou: Hmm... ok. I took a look at parquet-mr and the Rust Parquet implementation, neither of them seems to handle `distinct_count`. So I guess this is ok.

cpp/src/parquet/statistics.cc
pitrou: I think we can keep the code as-is indeed.

cpp/src/parquet/statistics.cc
```diff
@@ -552,12 +560,19 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
...
       this->statistics_.null_count += other.null_count();
+    } else {
+      this->has_null_count_ = false;
     }
-    if (other.HasDistinctCount()) {
-      this->statistics_.distinct_count += other.distinct_count();
+    // Distinct count cannot be merged.
+    if (has_distinct_count_) {
+      has_distinct_count_ = false;
     }
```
pitrou, line 573: Simply:
```suggestion
    has_distinct_count_ = false;
```

cpp/src/parquet/statistics.cc
pitrou: Hmm... sorry, but I don't understand what you mean with "not added able".

cpp/src/parquet/statistics.cc
mapleFU: I can discard it and add support for it using `set` later.

cpp/src/parquet/statistics.cc
```diff
@@ -602,6 +617,7 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
   }
 
   EncodedStatistics Encode() override {
+    // Currently, distinct_count will never be set in EncodedStatistics.
```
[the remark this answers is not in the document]
mapleFU: Yes we should, let me fix it.

cpp/src/parquet/statistics.cc
wgtmac: CMIW, `null_count_` can be safely accumulated and merged, so `has_null_count_` is set to true and `null_count_` is set to 0 at its initial stage. During the building process, `has_null_count_` should always be true unless something wrong happens.

On the contrary, `distinct_count_` cannot be merged and actually it is not set in the building process. It makes sense to set it to false at its initial stage.

I think @mapleFU means `distinct_count_` cannot be merged.

pitrou: Ok, fair enough.

cpp/src/parquet/statistics.cc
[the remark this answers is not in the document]
mapleFU: I change *will be* to *will tend to be*, I think this is more proper here

cpp/src/parquet/column_writer.cc
```diff
@@ -1346,7 +1346,12 @@ class TypedColumnWriterImpl : public ColumnWriterImpl, public TypedColumnWriter<
 
   EncodedStatistics GetChunkStatistics() override {
     EncodedStatistics result;
-    if (chunk_statistics_) result = chunk_statistics_->Encode();
+    if (chunk_statistics_) {
+      if (has_dictionary_ && !fallback_) {
+        chunk_statistics_->SetDistinctCount(current_dict_encoder_->num_entries());
```
mapleFU, line 1351: @pitrou I think this is the proper way to set distinct count. Another method is

```c++
if (chunk_statistics_) { 
  result = chunk_statistics_->Encode();
  if (has_dictionary_ && !fallback_) {
    result.set_distinct_count(...);
  }
}
```

Do you think this is ok?

mapleFU:
> data = [1, 2, 2, None, 4], type = DataType(uint16), physical_type = 'INT32'
min_value = 1, max_value = 4, null_count = 1, num_values = 4, distinct_count = 0

After my reflect, the `distinct_count` need to be fixed. Can you confirm that @pitrou ?

pitrou:
@mapleFU How was this data produced?

pitrou:
(and, yes, `distinct_count = 0` is obviously wrong here :-))

mapleFU:
From here: https://ci.appveyor.com/project/ApacheSoftwareFoundation/arrow/builds/47327471
Let me draft the fix the testing here...

cpp/src/parquet/column_writer.cc
pitrou: I _think_ it's ok. But I don't think it's very important to write the distinct count, if we're only able to write it for dictionary-encoded chunks.

mapleFU: Ok so let me just remove it. I can use `distinct_count` to inspect dictionary, but I guess it's really useless

cpp/src/parquet/statistics.cc
wgtmac, line 477: nit: add a comment here to explain why has_null_count_ and has_distinct_count_ have different defaults as we have discussed here

cpp/src/parquet/statistics.cc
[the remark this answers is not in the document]
wgtmac: I'd rather not to say who will call this. Instead, we can simply say it is used to create an empty stats and update it from follow-up writes.

cpp/src/parquet/statistics.cc
```diff
@@ -472,34 +474,40 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
     comparator_ = std::static_pointer_cast<TypedComparator<DType>>(comp);
     TypedStatisticsImpl::Reset();
     has_null_count_ = true;
-    has_distinct_count_ = true;
+    has_distinct_count_ = false;
   }
 
+  // Currently, this function will tend to be called by testing.
```
wgtmac, line 480: Ditto, we can say it is used to create stats from provided values.

cpp/src/parquet/statistics.cc
```diff
@@ -472,34 +474,40 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
...
     TypedStatisticsImpl::IncrementNullCount(null_count);
-    IncrementDistinctCount(distinct_count);
+    SetDistinctCount(distinct_count);
 
     Copy(min, &min_, min_buffer_.get());
     Copy(max, &max_, max_buffer_.get());
     has_min_max_ = true;
   }
 
+  // Currently, this function will tend to be called by ColumnChunkMetaData during read.
```
wgtmac, line 495: Ditto, we can say it is used to create stats from a thrift Statistics object.

cpp/src/parquet/statistics.cc
```diff
@@ -463,6 +463,8 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
  public:
   using T = typename DType::c_type;
 
+  // Currently, this function will tend to be called by ColumnWriter to create the
```
pitrou, line 466: Nit (and same below)
```suggestion
  // Currently, this constructor will tend to be called by ColumnWriter to create the
```

cpp/src/parquet/statistics.h
```diff
@@ -166,13 +167,13 @@ class PARQUET_EXPORT EncodedStatistics {
   void set_is_signed(bool is_signed) { is_signed_ = is_signed; }
 
   EncodedStatistics& set_max(const std::string& value) {
-    *max_ = value;
+    max_ = value;
```
pitrou, line 170: If you want to further micro-optimize:
```c++
  EncodedStatistics& set_max(std::string value) {
    max_ = std::move(value);
    has_max = true;
    return *this;
  }
```

cpp/src/parquet/statistics_test.cc
```diff
@@ -377,6 +377,35 @@ class TestStatistics : public PrimitiveTypedTest<TestType> {
...
+    EXPECT_TRUE(statistics2->HasDistinctCount());
+    auto s2 = std::dynamic_pointer_cast<TypedStatistics<TestType>>(statistics2);
+
+    auto total = MakeStatistics<TestType>(this->schema_.Column(0));
+    total->Merge(*s1);
+    total->Merge(*s2);
+
+    EXPECT_FALSE(total->HasDistinctCount());
+    EXPECT_FALSE(total->HasMinMax());
+    EXPECT_EQ(2000, total->num_values());
```
pitrou, line 406: Can you tell the null count as well?

mapleFU: Updated

cpp/src/parquet/metadata.cc
```diff
@@ -87,20 +87,28 @@ std::string ParquetVersionToString(ParquetVersion::type ver) {
 template <typename DType>
 static std::shared_ptr<Statistics> MakeTypedColumnStats(
     const format::ColumnMetaData& metadata, const ColumnDescriptor* descr) {
+  DCHECK(metadata.__isset.statistics);
+  int64_t num_values = 0;
```
mapleFU, line 91: I've handle it here. Maybe it's not right. But I think it's compatible with previous code

mapleFU:
Emm why gandiva test failed...

```
[----------] 2 tests from TestExpressionRegistry
[ RUN      ] TestExpressionRegistry.VerifySupportedFunctions
/Users/runner/work/arrow/arrow/cpp/src/gandiva/expression_registry_test.cc:48: Failure
Expected: (element) != (functions.end()), actual: 8-byte object <40-9D 8C-24 E1-7F 00-00> vs 8-byte object <40-9D 8C-24 E1-7F 00-00>
function signature date64[ms] date_diff(date64[ms], int64) missing in supported functions.

/Users/runner/work/arrow/arrow/cpp/src/gandiva/expression_registry_test.cc:48: Failure
Expected: (element) != (functions.end()), actual: 8-byte object <40-9D 8C-24 E1-7F 00-00> vs 8-byte object <40-9D 8C-24 E1-7F 00-00>
function signature timestamp[ms] date_diff(timestamp[ms], int64) missing in supported functions.

[  FAILED  ] TestExpressionRegistry.VerifySupportedFunctions (166 ms)
```

pitrou:
> Emm why gandiva test failed...

That happens from time to time on the macOS CI, it's unrelated.

cpp/src/parquet/statistics.cc
```diff
@@ -543,7 +557,7 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
     ResetCounts();
     has_min_max_ = false;
     has_distinct_count_ = false;
-    has_null_count_ = false;
+    has_null_count_ = true;
```
mapleFU, line 560: page statistics will call `Reset`, this will clear `has_null_count_`, however, this time it should have `has_null_count_ = 0`

cpp/src/parquet/statistics.cc
```diff
@@ -610,8 +629,12 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
     if (HasNullCount()) {
       s.set_null_count(this->null_count());
     }
-    // num_values_ is reliable and it means number of non-null values.
-    s.all_null_value = num_values_ == 0;
+    if (HasNullCount()) {
```
mapleFU, line 632: This refactor is a bit tricky. When `ColumnChunkMetadata` read a statistics from thrift, and if the thrift not have `null_count`, the `num_values` would be wrong. And in C++ Parquet, we'll never meet this case, because writer always has `has_null_count_`

mapleFU:
@pitrou @wgtmac Mind take a look? I've update the logic for `num_values` within `Statistics`

cpp/src/parquet/metadata.cc
```diff
@@ -87,20 +87,29 @@ std::string ParquetVersionToString(ParquetVersion::type ver) {
 template <typename DType>
 static std::shared_ptr<Statistics> MakeTypedColumnStats(
     const format::ColumnMetaData& metadata, const ColumnDescriptor* descr) {
+  DCHECK(metadata.__isset.statistics);
+  int64_t num_non_null_values = 0;
+  if (metadata.statistics.__isset.null_count) {
+    num_non_null_values = metadata.num_values - metadata.statistics.null_count;
+  } else {
+    // NOTE: statistics should `metadata.statistics.__isset.null_count`
+    // has a invalid non-null value number.
```
pitrou, line 96: What is a "invalid non-null value number"? I don't understand.

cpp/src/parquet/metadata.cc
```diff
@@ -87,20 +87,29 @@ std::string ParquetVersionToString(ParquetVersion::type ver) {
...
+    num_non_null_values = metadata.num_values;
```
pitrou, line 97: Why change this? `has_null_count` will be false anyway.

cpp/src/parquet/statistics.cc
```diff
@@ -471,35 +473,47 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
     auto comp = Comparator::Make(descr);
     comparator_ = std::static_pointer_cast<TypedComparator<DType>>(comp);
     TypedStatisticsImpl::Reset();
+    // Currently, writer will always write `distinct_count` to EncodedStatistics.
```
pitrou, line 476: Do you mean `null_count`?

cpp/src/parquet/statistics.cc
pitrou: Why would `num_values` be wrong?

cpp/src/parquet/statistics.cc
mapleFU: Yes, my fault

cpp/src/parquet/statistics.cc
mapleFU: Can a `!metadata.statistics.__isset.null_count` might means:

1. The underlying data has null
2. But it doesn't write it into statistics

If it can, when `!metadata.statistics.__isset.null_count`, `metadata.num_values - metadata.statistics.null_count` cannot represent the non-null count

mapleFU:
Emmm I've tried it, I think `metadata.__isset.statistics && !metadata.statistics.__isset.null_count` might means it has null but not set it's count. We use `metadata.num_values - metadata.statistics.null_count` to get `num_values` in `Statistics` (`num_values` is just used to build page index )

 If all writer can make sure that `metadata.__isset.statistics && !metadata.statistics.__isset.null_count` means no nulls, the currently implemention is right. If we regard `num_values` in `Statistics` as "wrongable", it's also ok.

cpp/src/parquet/statistics.cc
wgtmac: What about adding something like `ResetHasFlags()` below?
```cpp
void ResetHasFlags() {
  has_min_max_ = false;
  has_distinct_count_ = false;
  has_null_count_ = true;
}
```
We can call it in the constructors and Reset() to make sure the initial states are consistent.

cpp/src/parquet/statistics.cc
```diff
@@ -552,12 +568,17 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
 
   void Merge(const TypedStatistics<DType>& other) override {
     this->num_values_ += other.num_values();
+    // Merge always runs when Merge builder's page statistics
+    // into column chunk statistics, so it tent to have null count.
```
wgtmac, line 572:
```suggestion
    // into column chunk statistics, so it tends to have null count.
```

cpp/src/parquet/statistics.cc
```diff
@@ -463,6 +463,8 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
  public:
   using T = typename DType::c_type;
 
+  // Currently, this constructor will tend to be called by ColumnWriter to create the
```
wgtmac, line 466: I still don't think comment with `currently` and `tend` is a good idea as it may easily be out of sync and confuses users. It would be good enough to simply describe its supposed use case.

cpp/src/parquet/statistics.cc
mapleFU: Updated

mapleFU:
@pitrou @wgtmac I've fix the comments, mind take a look?

cpp/src/parquet/statistics.cc
```diff
@@ -610,8 +619,12 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
     if (HasNullCount()) {
       s.set_null_count(this->null_count());
     }
-    // num_values_ is reliable and it means number of non-null values.
-    s.all_null_value = num_values_ == 0;
+    if (HasNullCount()) {
```
pitrou, line 622: Can you merge this with the `if` before?

cpp/src/parquet/statistics.cc
```diff
@@ -610,8 +619,12 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
...
+      // num_values_ is reliable and it means number of non-null values.
+      s.all_null_value = num_values_ == 0;
```
pitrou, line 624: Can you add a statistics test for `all_null_value`? Including cases where `num_values_` is not reliable.

cpp/src/parquet/statistics.cc
```diff
@@ -627,7 +640,12 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
   T min_;
   T max_;
   ::arrow::MemoryPool* pool_;
-  int64_t num_values_ = 0;  // # of non-null values.
+  // # of non-null values.
```
pitrou, line 643:
```suggestion
  // Number of non-null values.
```

cpp/src/parquet/statistics.cc
```diff
@@ -627,7 +640,12 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
...
+  // num_values_ would be reliable when has_null_count_,
+  // if it's created from a page thrift statistics, and statistics
+  // doesn't have null_count, but page has null data, `num_values_`
+  // would be equal to value including nulls.
```
pitrou, line 647: This isn't very readable. Do you want to split this into two sentences?

cpp/src/parquet/statistics.cc
```diff
@@ -637,8 +655,9 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
 
   void Copy(const T& src, T* dst, ResizableBuffer*) { *dst = src; }
 
-  void IncrementDistinctCount(int64_t n) {
-    statistics_.distinct_count += n;
+  void SetDistinctCount(int64_t n) {
+    // distinct count can only be "set", and cannot be increment.
```
pitrou, line 659:
```suggestion
    // distinct count can only be "set", and cannot be incremented.
```

pitrou (changes requested):
@mapleFU Thanks for this. I think we want to have more tests for statistics behavior, and to make sure that we don't introduce any regressions.

cpp/src/parquet/statistics.cc
mapleFU: Added

cpp/src/parquet/statistics.cc
mapleFU: Tried it but maybe I'm not good for docs

mapleFU:
I've fixed the comments and added some tests @pitrou

mapleFU:
Is there any other things I need to fix?

cpp/src/parquet/statistics_test.cc
```diff
@@ -377,6 +380,79 @@ class TestStatistics : public PrimitiveTypedTest<TestType> {
...
+
+  void TestMergeMinMax() {
+    this->GenerateData(1000);
+
+    auto chunk_statistics = MakeStatistics<TestType>(this->schema_.Column(0));
+
+    {
+      auto page_statistics1 = MakeStatistics<TestType>(this->schema_.Column(0));
+      std::vector<uint8_t> valid_bits(
+          bit_util::BytesForBits(static_cast<uint32_t>(this->values_.size())) + 1, 0);
```
pitrou, line 433: `valid_bits` is not even used, is it?

cpp/src/parquet/statistics_test.cc
```diff
@@ -377,6 +380,79 @@ class TestStatistics : public PrimitiveTypedTest<TestType> {
...
+      EXPECT_FALSE(encoded_stats1.has_max);
+    }
+    {
+      auto page_statistics2 = MakeStatistics<TestType>(this->schema_.Column(0));
+      page_statistics2->Update(this->values_ptr_, this->values_.size(), 0);
+
+      chunk_statistics->Merge(*page_statistics2);
+      auto encoded_stats2 = chunk_statistics->Encode();
+      EXPECT_TRUE(encoded_stats2.has_min);
+      EXPECT_TRUE(encoded_stats2.has_max);
```
pitrou, line 452: Hmm... why are those true? Also, why not also check the actual min/max values?

pitrou (changes requested):
Really, I expect the tests to exercise a more varied number of situations _and_ to check all statistics fields (include min/max values, not just whether there is a min/max).

Also, please write test helpers so that each test doesn't have 10+ lines of boilerplate.

cpp/src/parquet/statistics_test.cc
mapleFU: I'll remove it

mapleFU:
Emmm what kind of test should I added here?
1. I test the behavior when `ResetFlag` is called
2. `num_distinct_value` and other being merged
3. `min-max` beging merged
4. `null-value`

Should I then add:
1. `null-count` and merge
2. behavior of different constructor

Is that enough?

cpp/src/parquet/statistics_test.cc
mapleFU: Or we can distinct from `all is null or nan` and `not has min-max, cannot be merged`. I prefer to compatible with previous currently

mapleFU:
```c++
  void SetMinMaxPair(std::pair<T, T> min_max) {
    // CleanStatistic can return a nullopt in case of erroneous values, e.g. NaN
    auto maybe_min_max = CleanStatistic(min_max);
    if (!maybe_min_max) return;
    auto min = maybe_min_max.value().first;
    auto max = maybe_min_max.value().second;
    if (!has_min_max_) {
      has_min_max_ = true;
      Copy(min, &min_, min_buffer_.get());
      Copy(max, &max_, max_buffer_.get());
    } else {
      Copy(comparator_->Compare(min_, min) ? min_ : min, &min_, min_buffer_.get());
      Copy(comparator_->Compare(max_, max) ? max : max_, &max_, max_buffer_.get());
    }
  }
```

If `other` has min-max, our code will always set `has_min_max_ = true` here

mapleFU:
@pitrou Sorry for writing ugly test yesterday, I didn't make clear the required testing. Now I tried to organize them and added some comments. Would you mind take a look?

cpp/src/parquet/statistics.cc
```diff
@@ -552,12 +556,17 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
 
   void Merge(const TypedStatistics<DType>& other) override {
     this->num_values_ += other.num_values();
+    // Merge always runs when Merge builder's page statistics
+    // into column chunk statistics, so it usually has null count.
```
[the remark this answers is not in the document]
wgtmac: Additionally, I don't think we need to add comment above. We don't know how external users call `Statistics::Merge()`. Rather we can comment that it tries to maintain every field at its best effort.

cpp/src/parquet/statistics.cc
```diff
@@ -552,12 +556,17 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
...
+    } else {
+      this->has_null_count_ = false;
     }
-    if (other.HasDistinctCount()) {
-      this->statistics_.distinct_count += other.distinct_count();
-    }
+    // Distinct count cannot be merged.
+    has_distinct_count_ = false;
+    // If !other.HasMinMax, might be all-nulls or nulls and nan,
+    // so, not clear `this->has_min_max_` here.
```
wgtmac, line 569:
```suggestion
    // Do not clear min/max here if the other side does not provide
    // min/max which may happen when other is an empty stats or all
    // its values are null and/or NaN.
```

cpp/src/parquet/statistics.cc
```diff
@@ -627,7 +638,11 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
   T min_;
   T max_;
   ::arrow::MemoryPool* pool_;
-  int64_t num_values_ = 0;  // # of non-null values.
+  // Number of non-null values.
+  // num_values_ would be reliable when has_null_count_.
+  // If statistics is created from a page thrift statistics, and statistics
+  // doesn't have null_count, `num_values_` may include null values.
```
wgtmac, line 644:
```suggestion
  // Number of non-null values.
  // Please note that num_values_ is reliable when has_null_count_ is set.
  // When has_null_count_ is not set, e.g. a page statistics created from
  // a statistics thrift message which doesn't have the optional null_count, 
  // `num_values_` may include null values.
```

mapleFU:
```
[ RUN      ] TwoLevelCacheWithExpirationTest.CleanupPeriodOk
/arrow/cpp/src/parquet/encryption/two_level_cache_with_expiration_test.cc:118: Failure
  lifetime2->size()
    Which is: 0
  2
[  FAILED  ] TwoLevelCacheWithExpirationTest.CleanupPeriodOk (525 ms)
```

No idea why this is failed...

cpp/src/parquet/statistics.cc
```diff
@@ -648,6 +664,16 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
...
   }
 
+  void ResetHasFlags() {
+    this->has_min_max_ = false;
+    // writer will write `distinct_count` to EncodedStatistics.
+    // So, disable it when reset.
+    this->has_distinct_count_ = false;
+    // writer will always write `null_count` to EncodedStatistics.
+    // So, enable it when reset.
+    this->has_null_count_ = true;
```
wgtmac, line 674:
```suggestion
    // has_min_max_ will only be set when it meets any valid value.
    this->has_min_max_ = false;
    // has_distinct_count_ will only be set once SetDistinctCount()
    // is called because distinct count calculation is not cheap and
    // disabled by default.
    this->has_distinct_count_ = false;
    // Null count calculation is cheap and enabled by default.
    this->has_null_count_ = true;
```

cpp/src/parquet/statistics_test.cc
```diff
@@ -590,6 +593,173 @@ TYPED_TEST(TestNumericStatistics, Equals) {
...
+    VerifyMergedStatistics(*statistics1, *statistics2,
+                           [](TypedStatistics<TestType>* merged_statistics) {
+                             EXPECT_FALSE(merged_statistics->HasDistinctCount());
+                             EXPECT_FALSE(merged_statistics->Encode().has_distinct_count);
+                           });
+  }
+
+  // If a page is all nulls or nan, min and max should be set to null.
+  // And the all-null page being merged with a page having min-max,
+  // the result should have min-max.
```
wgtmac, line 652:
```suggestion
  // If all values in a page are null or nan, its stats should not set min-max.
  // Merging its stats with another page having good min-max stats should not
  // drop the valid min-max from the latter page.
```

cpp/src/parquet/statistics_test.cc
```diff
@@ -590,6 +593,173 @@ TYPED_TEST(TestNumericStatistics, Equals) {
...
+  // and page merge with it should also not have null_count.
+  void TestMergeNullCount() {
+    this->GenerateData(1000);
+
+    // Page should have null-count even if no nulls
+    std::shared_ptr<TypedStatistics<TestType>> statistics1;
+    {
+      statistics1 = MakeStatistics<TestType>(this->schema_.Column(0));
+      statistics1->Update(this->values_ptr_, /*num_values=*/this->values_.size(),
+                          /*null_count*/ 0);
```
wgtmac, line 695:
```suggestion
                          /*null_count=*/ 0);
```

cpp/src/parquet/statistics.cc
```diff
@@ -609,9 +618,11 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
     }
     if (HasNullCount()) {
       s.set_null_count(this->null_count());
+      // num_values_ is reliable and it means number of non-null values.
+      s.all_null_value = num_values_ == 0;
     }
-    // num_values_ is reliable and it means number of non-null values.
-    s.all_null_value = num_values_ == 0;
+    // Currently, distinct count would not be written.
+    // So, not call set_distinct_count.
```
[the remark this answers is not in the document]
mapleFU: Hmmm that's merge. I think Merge is different from Reset

cpp/src/parquet/statistics.cc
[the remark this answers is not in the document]
mapleFU: > If Merge is called twice with other.HasNullCount()=false and then other.HasNullCount()=true, here may have problem. 

Why? If `other.HasNullCount()` called, the flag would never be true.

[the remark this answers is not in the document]
mapleFU: I know what you mean. This is a bit confusing:

1. `IncrementNullCount` will set `has_null_count_` to true ( but I think it's not neccessary ). The `page_statistics_` would call `IncrementNullCount` to set it( although at the first time `has_null_count_` would be true for it
2. `chunk_statistics_` will only call `merge`, once it meets a `Statistics` without `has_null_count_`, it's `has_null_count_` would never be true.

cpp/src/parquet/statistics_test.cc
```diff
@@ -590,6 +593,173 @@ TYPED_TEST(TestNumericStatistics, Equals) {
...
+
+TYPED_TEST(TestStatisticsHasFlag, MergeDistinctCount) {
+  ASSERT_NO_FATAL_FAILURE(this->TestMergeDistinctCount());
+}
+
+TYPED_TEST(TestStatisticsHasFlag, MergeNullCount) {
+  ASSERT_NO_FATAL_FAILURE(this->TestMergeNullCount());
+}
+
+TYPED_TEST(TestStatisticsHasFlag, TestMergeMinMax) {
```
wgtmac, line 755:
```suggestion
TYPED_TEST(TestStatisticsHasFlag, MergeMinMax) {
```

cpp/src/parquet/statistics_test.cc
```diff
@@ -590,6 +593,173 @@ TYPED_TEST(TestNumericStatistics, Equals) {
...
+  ASSERT_NO_FATAL_FAILURE(this->TestMergeMinMax());
+}
+
+TYPED_TEST(TestStatisticsHasFlag, NotHasNullValues) {
+  ASSERT_NO_FATAL_FAILURE(this->TestNotHasNullValue());
```
wgtmac, line 760:
```suggestion
TYPED_TEST(TestStatisticsHasFlag, MissingNullCount) {
  ASSERT_NO_FATAL_FAILURE(this->TestMissingNullCount());
```

cpp/src/parquet/statistics_test.cc
```diff
@@ -590,6 +593,173 @@ TYPED_TEST(TestNumericStatistics, Equals) {
...
+  // If statistics doesn't have null count, all_null_value should be false.
+  void TestNotHasNullValue() {
+    EncodedStatistics encoded_statistics1;
+    encoded_statistics1.has_null_count = false;
+    auto statistics1 = Statistics::Make(this->schema_.Column(0), &encoded_statistics1,
+                                        /*num_values=*/1000);
+    auto s1 = std::dynamic_pointer_cast<TypedStatistics<TestType>>(statistics1);
+    EXPECT_FALSE(s1->HasNullCount());
+    auto encoded = s1->Encode();
+    EXPECT_FALSE(encoded.all_null_value);
```
wgtmac, line 741: Check other attributes of `encoded` as well?

cpp/src/parquet/statistics_test.cc
```diff
@@ -590,6 +593,173 @@ TYPED_TEST(TestNumericStatistics, Equals) {
...
 
+template <typename TestType>
+class TestStatisticsHasFlag : public TestStatistics<TestType> {
+ public:
+  void SetUp() override {
+    TestStatistics<TestType>::SetUp();
+    this->SetUpSchema(Repetition::OPTIONAL);
+  }
+
+  std::shared_ptr<TypedStatistics<TestType>> BuildMergedStatistics(
```
wgtmac, line 604:
```suggestion
  std::shared_ptr<TypedStatistics<TestType>> MergeStatistics(
```

cpp/src/parquet/statistics_test.cc
```diff
@@ -590,6 +593,173 @@ TYPED_TEST(TestNumericStatistics, Equals) {
...
+    chunk_statistics->Merge(stats1);
+    chunk_statistics->Merge(stats2);
+    return chunk_statistics;
+  }
+
+  void VerifyMergedStatistics(const TypedStatistics<TestType>& stats1,
+                              const TypedStatistics<TestType>& stats2,
+                              const std::function<void(TypedStatistics<TestType>*)>& fn) {
+    fn(BuildMergedStatistics(stats1, stats2).get());
+    fn(BuildMergedStatistics(stats2, stats1).get());
```
wgtmac, line 616:
```suggestion
                              const std::function<void(TypedStatistics<TestType>*)>& test_fn) {
    ASSERT_NO_FATAL_FAILURE(test_fn(MergeStatistics(stats1, stats2).get()));
    ASSERT_NO_FATAL_FAILURE(test_fn(MergeStatistics(stats2, stats1).get()));
```

wgtmac: CMIW, directly calling `.get()` on the shared_ptr may introduce lifetime issue if the returned shared_ptr object is released before fn uses the raw pointer.

cpp/src/parquet/statistics_test.cc
```diff
@@ -590,6 +593,173 @@ TYPED_TEST(TestNumericStatistics, Equals) {
...
+  }
+
+  // Distinct count should set to false when Merge is called.
+  void TestMergeDistinctCount() {
```
wgtmac, line 620: It would be good if we can test all flags in these cases even when the main purpose of a specific case is not for those flags.

cpp/src/parquet/statistics_test.cc
mapleFU: > CMIW, directly calling .get() on the shared_ptr may introduce lifetime issue if the returned shared_ptr object is released before fn uses the raw pointer.

See https://en.cppreference.com/w/cpp/language/lifetime#Temporary_object_lifetime

> All temporary objects are destroyed as the last step in evaluating the [full-expression](https://en.cppreference.com/w/cpp/language/expressions#Full-expressions) that (lexically) contains the point where they were created, and if multiple temporary objects were created, they are destroyed in the order opposite to the order of creation. This is true even if that evaluation ends in throwing an exception.

cpp/src/parquet/statistics.cc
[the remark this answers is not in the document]
mapleFU: During test I've test that both `!has_null_count_ merge with has_null_count_` and `has_null_count_ merge with !has_null_count_`

cpp/src/parquet/statistics_test.cc
mapleFU: Isn't here is already enough some test for other properties? I guess I can add it but it might make tests confusion.

cpp/src/parquet/statistics.cc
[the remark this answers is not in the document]
wgtmac: OK, thanks for confirming!

cpp/src/parquet/statistics_test.cc
wgtmac: Thanks for the info!

cpp/src/parquet/statistics_test.cc
```diff
@@ -590,6 +593,173 @@ TYPED_TEST(TestNumericStatistics, Equals) {
...
+                             EXPECT_TRUE(merged_statistics->Encode().has_min);
+                             EXPECT_TRUE(merged_statistics->Encode().has_max);
+                           });
+  }
+
+  // Default statistics should have null_count even if no nulls is written.
+  // However, if statistics read from thrift, it might not have null_count,
+  // and page merge with it should also not have null_count.
+  void TestMergeNullCount() {
+    this->GenerateData(1000);
```
wgtmac, line 688:
```suggestion
    this->GenerateData(/*num_values=*/1000);
```

cpp/src/parquet/statistics_test.cc
```diff
@@ -590,6 +593,173 @@ TYPED_TEST(TestNumericStatistics, Equals) {
...
+  void TestNotHasNullValue() {
+    EncodedStatistics encoded_statistics1;
+    encoded_statistics1.has_null_count = false;
+    auto statistics1 = Statistics::Make(this->schema_.Column(0), &encoded_statistics1,
+                                        /*num_values=*/1000);
+    auto s1 = std::dynamic_pointer_cast<TypedStatistics<TestType>>(statistics1);
+    EXPECT_FALSE(s1->HasNullCount());
+    auto encoded = s1->Encode();
+    EXPECT_FALSE(encoded.all_null_value);
+  }
```
wgtmac, line 742:
```suggestion
  void TestNotHasNullValue() {
    EncodedStatistics encoded_statistics;
    encoded_statistics.has_null_count = false;
    auto statistics = Statistics::Make(this->schema_.Column(0), &encoded_statistics,
                                       /*num_values=*/1000);
    auto typed_stats = std::dynamic_pointer_cast<TypedStatistics<TestType>>(statistics);
    EXPECT_FALSE(typed_stats->HasNullCount());
    auto encoded = typed_stats->Encode();
    EXPECT_FALSE(encoded.all_null_value);
  }
```

cpp/src/parquet/statistics.cc
```diff
@@ -552,12 +555,18 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
 
   void Merge(const TypedStatistics<DType>& other) override {
     this->num_values_ += other.num_values();
+    // Merge always runs when Merge builder's page statistics
+    // into column chunk statistics, so it usually has null count.
```
wgtmac, line 559:
```suggestion
    // null_count is always valid when merging page statistics into
    // column chunk statistics.
```

cpp/src/parquet/statistics_test.cc
```diff
@@ -590,6 +593,173 @@ TYPED_TEST(TestNumericStatistics, Equals) {
...
+                           [](TypedStatistics<TestType>* merged_statistics) {
+                             EXPECT_TRUE(merged_statistics->HasMinMax());
+                             EXPECT_TRUE(merged_statistics->Encode().has_min);
+                             EXPECT_TRUE(merged_statistics->Encode().has_max);
+                           });
+  }
+
+  // Default statistics should have null_count even if no nulls is written.
+  // However, if statistics read from thrift, it might not have null_count,
+  // and page merge with it should also not have null_count.
```
wgtmac, line 686:
```suggestion
  // Default statistics should have null_count even if no nulls is written.
  // However, if statistics is created from thrift message, it might not
  // have null_count. Merging statistics from such page will result in an
  // invalid null_count as well.
```

cpp/src/parquet/statistics_test.cc
```diff
@@ -590,6 +593,173 @@ TYPED_TEST(TestNumericStatistics, Equals) {
...
+  // Merging its stats with another page having good min-max stats should not
+  // drop the valid min-max from the latter page.
+  void TestMergeMinMax() {
+    this->GenerateData(1000);
+    // Create a statistics object without min-max.
+    std::shared_ptr<TypedStatistics<TestType>> statistics1;
+    {
+      statistics1 = MakeStatistics<TestType>(this->schema_.Column(0));
+      statistics1->Update(this->values_ptr_, /*num_values=*/0,
+                          /*null_count*/ this->values_.size());
```
wgtmac, line 660:
```suggestion
                          /*null_count=*/ this->values_.size());
```

wgtmac (approved):
LGTM. Thanks!

mapleFU:
Gently ping @pitrou

pitrou (changes requested):
LGTM in general, just two minor comments

cpp/src/parquet/statistics.cc
```diff
@@ -609,9 +618,10 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
     }
     if (HasNullCount()) {
       s.set_null_count(this->null_count());
+      // num_values_ is reliable and it means number of non-null values.
+      s.all_null_value = num_values_ == 0;
     }
-    // num_values_ is reliable and it means number of non-null values.
-    s.all_null_value = num_values_ == 0;
+    // FIXME(mwish): distinct count is not encoded for now.
```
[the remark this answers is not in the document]
mapleFU: Created. But actually I found a lot of issues be forgotten https://github.com/apache/arrow/issues/36505

pitrou, line 624:
```suggestion
    // TODO (GH-36505): distinct count is not encoded for now.
```

pitrou (approved):

PR #35991, opened 2023-06-08
GH-35990: [CI][C++][Windows] Don't use -l for "choco list"

### LINKED ISSUE #35990: [CI][C++][Windows] "choco list -l" isn't available with choco v2

https://github.com/apache/arrow/actions/runs/5208662945/jobs/9397507544#step:3:38

```text
Chocolatey v2.0.0
Invalid argument -l. This argument has been removed from the list command and cannot be used.
```

https://docs.chocolatey.org/en-us/guides/upgrading-to-chocolatey-v2-v6#the-list-command-now-lists-local-packages-only-and-the-local-only-and-lo-options-have-been-removed

> The List Command Now Lists Local Packages Only and the --local-only and -lo Options Have Been Removed
>
> In version 1.0.0 of Chocolatey CLI, we added notices that the choco list command will list only local packages, and deprecated the -l and it's alias options. See this [GitHub issue for more information](https://github.com/chocolatey/choco/issues/158). We have also removed the -a and it's alias options from the list command as it no longer made sense to have that option once side-by-side installs were removed.

### Component(s)

C++, Continuous Integration

kou:
### Rationale for this change

Because it's removed and needless now.

https://docs.chocolatey.org/en-us/guides/upgrading-to-chocolatey-v2-v6#the-list-command-now-lists-local-packages-only-and-the-local-only-and-lo-options-have-been-removed

> The List Command Now Lists Local Packages Only and the --local-only and -lo Options Have Been Removed
>
> In version 1.0.0 of Chocolatey CLI, we added notices that the choco list command will list only local packages, and deprecated the -l and it's alias options. See this [GitHub issue for more information](https://github.com/chocolatey/choco/issues/158). We have also removed the -a and it's alias options from the list command as it no longer made sense to have that option once side-by-side installs were removed.

Just removed "-l".

### Are these changes tested?

Yes.

No.
* Closes: #35990

### DIFF
--- a/.github/workflows/cpp.yml
+++ b/.github/workflows/cpp.yml
@@ -267,7 +267,7 @@ jobs:
             /d 1 `
             /f
       - name: Installed Packages
-        run: choco list -l
+        run: choco list
       - name: Install Dependencies
         run: choco install -y --no-progress openssl
       - name: Checkout Arrow


pitrou (approved):

pitrou:
Thanks @kou !

PR #35996, opened 2023-06-08
GH-35988: [C#] The C data interface implementation can leak on import

### LINKED ISSUE #35988: [C#] The C data interface implementation can leak on import

An imported stream will leak memory as CArrowArray instances are allocated for each batch but they're never freed.

Code which imports an IArrowArrayStream may have trouble freeing the CArrowArrayStream that was used to import it because it can't necessarily track the stream's lifetime.

### Component(s)

C#

CurtHagenlocher:

To ensure proper cleanup, immediately copies the contents of the C structure into the imported class for arrays and streams.
Relaxes the requirement when exporting that the target structure appear uninitialized.

### Are these changes tested?

Existing tests pass. We don't as yet seem to have a good way to test for memory leaks so no new tests have been added.

No.

* Closes: #35988

### DIFF
--- a/csharp/src/Apache.Arrow/C/CArrowArrayImporter.cs
+++ b/csharp/src/Apache.Arrow/C/CArrowArrayImporter.cs
@@ -44,14 +44,13 @@ namespace Apache.Arrow.C
         /// </examples>
         /// <param name="ptr">The pointer to the array being imported</param>
         /// <param name="type">The type of the array being imported</param>
-        /// <param name="freeOnRelease">If true, frees the memory for the CArrowArray on final release.</param>
         /// <returns>The imported C# array</returns>
-        public static unsafe IArrowArray ImportArray(CArrowArray* ptr, IArrowType type, bool freeOnRelease = false)
+        public static unsafe IArrowArray ImportArray(CArrowArray* ptr, IArrowType type)
         {
             ImportedArrowArray importedArray = null;
             try
             {
-                importedArray = new ImportedArrowArray(ptr, freeOnRelease);
+                importedArray = new ImportedArrowArray(ptr);
                 return importedArray.GetAsArray(type);
             }
             finally
@@ -80,14 +79,13 @@ namespace Apache.Arrow.C
         /// </examples>
         /// <param name="ptr">The pointer to the record batch being imported</param>
         /// <param name="schema">The schema type of the record batch being imported</param>
-        /// <param name="freeOnRelease">If true, frees the memory for the CArrowArray on final release.</param>
         /// <returns>The imported C# record batch</returns>
-        public static unsafe RecordBatch ImportRecordBatch(CArrowArray* ptr, Schema schema, bool freeOnRelease = false)
+        public static unsafe RecordBatch ImportRecordBatch(CArrowArray* ptr, Schema schema)
         {
             ImportedArrowArray importedArray = null;
             try
             {
-                importedArray = new ImportedArrowArray(ptr, freeOnRelease);
+                importedArray = new ImportedArrowArray(ptr);
                 return importedArray.GetAsRecordBatch(schema);
             }
             finally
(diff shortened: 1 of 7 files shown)


csharp/src/Apache.Arrow/C/CArrowArrayImporter.cs
```diff
@@ -29,7 +29,8 @@ public static class CArrowArrayImporter
         /// </summary>
         /// <remarks>
         /// This will call the release callback once all of the buffers in the returned
-        /// IArrowArray are disposed.
+        /// IArrowArray are disposed. If freeOnRelease is set, it will also free the memory
```
eerhardt, line 32: Usually comments about the parameters go in a `/// <param name="name">` comment.

csharp/src/Apache.Arrow/C/CArrowArrayImporter.cs
```diff
@@ -42,12 +43,12 @@ public static class CArrowArrayImporter
         /// IArrowArray importedArray = CArrowArrayImporter.ImportArray(importedPtr);
         /// </code>
         /// </examples>
-        public static unsafe IArrowArray ImportArray(CArrowArray* ptr, IArrowType type)
+        public static unsafe IArrowArray ImportArray(CArrowArray* ptr, IArrowType type, bool freeOnRelease)
```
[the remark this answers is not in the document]
CurtHagenlocher: I was originally going to default to false, but maybe forcing the user to think about it isn't a bad idea?

CurtHagenlocher:
> The second change is IMHO the symptom of an implementation issue. `ImportedArrowArray` currently stores the given raw pointer `CArrowArray* _cArray`. Instead, it should probably store its own `CArrowArray` struct like the C++ importer does:

Ah, that's much nicer.

csharp/src/Apache.Arrow/C/CArrowArrayStreamImporter.cs
```diff
@@ -99,24 +93,17 @@ public ValueTask<RecordBatch> ReadNextRecordBatchAsync(CancellationToken cancell
...
+                    if (cArray.release != null)
                     {
-                        result = CArrowArrayImporter.ImportRecordBatch(cArray, _schema);
-                    }
-                }
-                finally
-                {
-                    if (result == null)
-                    {
-                        CArrowArray.Free(cArray);
```
eerhardt, line 119: Who is freeing the memory that cArray references now? Or does the `RecordBatch` that gets returned take ownership of this memory now?

csharp/src/Apache.Arrow/C/CArrowArrayImporter.cs
```diff
@@ -90,42 +96,49 @@ public static unsafe RecordBatch ImportRecordBatch(CArrowArray* ptr, Schema sche
...
+                cArray->release = null;
             }
 
             protected override void FinalRelease()
             {
-                if (_cArray->release != null)
+                if (_cArray.release != null)
                 {
-                    _cArray->release(_cArray);
+                    fixed (CArrowArray * cArray =  &_cArray)
```
eerhardt, line 119: (nit)
```suggestion
                    fixed (CArrowArray* cArray =  &_cArray)
```

csharp/src/Apache.Arrow/C/CArrowArrayStreamImporter.cs
CurtHagenlocher: ImportRecordBatch always took ownership of the array data before. What's changed is that the memory for cArray is now allocated on the stack instead of the heap.

csharp/src/Apache.Arrow/C/CArrowArrayImporter.cs
```diff
@@ -90,42 +96,49 @@ public static unsafe RecordBatch ImportRecordBatch(CArrowArray* ptr, Schema sche
...
             protected override void FinalRelease()
             {
-                if (_cArray->release != null)
+                if (_cArray.release != null)
                 {
-                    _cArray->release(_cArray);
+                    fixed (CArrowArray* cArray = &_cArray)
+                    {
+                        cArray->release(cArray);
+                    }
```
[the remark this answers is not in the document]
CurtHagenlocher: Without the "fixed", in theory the GC could move the current object while we're inside the call. This would change the location of the embedded struct.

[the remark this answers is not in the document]
pitrou: Yes, I was wondering about the reason. Thank you!

csharp/src/Apache.Arrow/C/CArrowSchemaExporter.cs
```diff
@@ -50,10 +50,6 @@ public static unsafe void ExportType(IArrowType datatype, CArrowSchema* schema)
             {
                 throw new ArgumentNullException(nameof(schema));
             }
-            if (schema->release != null)
-            {
-                throw new ArgumentException("Cannot export schema to a struct that is already initialized.");
```
adamreeve, line 55: For what it's worth, it looks like this change fixes an issue I ran into using a C# exported stream from an older version of C++ Arrow, where the schema struct used when importing the stream was uninitialized: https://github.com/apache/arrow/blob/apache-arrow-10.0.1/cpp/src/arrow/c/bridge.cc#L1782 (this has since been fixed)

westonpace (commented):
If I understand correctly the basic premise is that, on import, we let `CArrowArray` be in the garbage collector pool since it should be safe to destroy if we happen to lose all references to it (unlike an exported array which we assume will have external references)?  That seems valid to me.

pitrou:
As far as I can understand, the code looks good now (but I'm not a C# developer).

eerhardt (approved):
LGTM. I'll merge this soon, if I don't hear any objections.

westonpace (approved):

PR #35997, opened 2023-06-08
GH-35576: [C++] Make Decimal{128,256}::FromReal more accurate

### LINKED ISSUE #35576: [C++] Decimal{128,256}::FromReal accuracy loss on non-small scale values

```
a = pa.array([545803904.0], type=pa.float32())
>>> a.cast(pa.decimal128(38, 18))
<pyarrow.lib.Decimal128Array object at 0x000001DE8FF59840>
[
  545803886.966396699654750208
]
>>>
```

While encoding of decimal128 and float32 is very different, the difference after the cast  seems much larger than expected.

### Component(s)

Python

pitrou:
The original algorithm for real-to-decimal conversion did its computations in the floating-point domain, accumulating rounding errors especially for large scale or precision values, such as:
```
>>> pa.array([1234567890.]).cast(pa.decimal128(38, 11))
<pyarrow.lib.Decimal128Array object at 0x7f05f4a3f1c0>
[
  1234567889.99999995904
]
>>> pa.array([1234567890.]).cast(pa.decimal128(38, 12))
<pyarrow.lib.Decimal128Array object at 0x7f05f494f9a0>
[
  1234567890.000000057344
]
```

The new algorithm strives to avoid precision loss by doing all its computations in the decimal domain. However, negative scales, which are presumably infrequent, fall back on the old algorithm.

* Closes: #35576

### DIFF
--- a/cpp/src/arrow/util/basic_decimal.cc
+++ b/cpp/src/arrow/util/basic_decimal.cc
@@ -29,6 +29,7 @@
 
 #include "arrow/util/bit_util.h"
 #include "arrow/util/config.h"  // for ARROW_USE_NATIVE_INT128
+#include "arrow/util/decimal_internal.h"
 #include "arrow/util/endian.h"
 #include "arrow/util/int128_internal.h"
 #include "arrow/util/int_util_overflow.h"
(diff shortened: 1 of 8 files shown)


pitrou:
@felipecrv @benibus Would either you have time to review this?

cpp/src/arrow/util/decimal_internal.h
```diff
@@ -0,0 +1,494 @@
...
+    BasicDecimal256FromLE(
+        {0ULL, 13416929297035424256ULL, 7104660214910016933ULL, 7965459555662261ULL}),
+    BasicDecimal256FromLE(
+        {0ULL, 5042084454387381248ULL, 15706369927971514489ULL, 79654595556622613ULL}),
+    BasicDecimal256FromLE(
+        {0ULL, 13527356396454709248ULL, 9489746690038731964ULL, 796545955566226138ULL})};
+
+#undef BasicDecimal256FromLE
+
+constexpr int kCeilLog2PowersOfTen[76 + 1] = {
```
westonpace, line 430: Perhaps a small comment explaining what this is

cpp/src/arrow/util/basic_decimal.cc
```diff
@@ -951,14 +631,6 @@ DecimalStatus BasicDecimal128::Divide(const BasicDecimal128& divisor,
   return DecimalDivide(*this, divisor, result, remainder);
 }
 
-bool operator==(const BasicDecimal128& left, const BasicDecimal128& right) {
-  return left.high_bits() == right.high_bits() && left.low_bits() == right.low_bits();
-}
-
-bool operator!=(const BasicDecimal128& left, const BasicDecimal128& right) {
-  return !operator==(left, right);
-}
```
westonpace, line 960: Is this going away because it is superceded by the version in GenericBasicDecimal?

cpp/src/arrow/util/decimal.cc
```diff
@@ -44,195 +45,291 @@ using internal::SafeLeftShift;
...
-    1000000000000000ULL,
-    10000000000000000ULL,
-    100000000000000000ULL,
-    1000000000000000000ULL
-    // clang-format on
+struct BaseDecimalRealConversion {
+  // Return 10**exp, with a fast lookup, assuming `exp` is withing bounds
+  template <typename Real>
+  static Real PowerOfTen(int32_t exp) {
+    DCHECK(exp >= -76 && exp <= 76);
```
westonpace, line 54: Do we want to make `76` a constant instead of a magic number?

cpp/src/arrow/util/decimal_test.cc
```diff
@@ -802,34 +814,21 @@ class TestDecimalFromReal : public ::testing::Test {
     const std::vector<ParamType> params{
         // clang-format off
         {0.0f, 1, 0, "0"},
-        {-0.0f, 1, 0, "0"},
```
westonpace, line 805: Why remove these cases?  We can still convert negative numbers right?  Isn't it just less precise?

python/pyarrow/tests/test_compute.py
```diff
@@ -1822,6 +1825,128 @@ def test_fsl_to_fsl_cast(value_type):
...
+        # the iterative algorithm (GH-35576).
+        diff_digits = abs(actual - expected) * 10**decimal_ty.scale
+        limit = 2 if decimal_ty.precision < max_precision - 1 else 4
+        assert diff_digits <= limit, (
+            f"float_val = {float_val!r}, precision={decimal_ty.precision}, "
+            f"expected = {expected!r}, actual = {actual!r}, "
+            f"diff_digits = {diff_digits!r}")
+
+
+# XXX Cannot test float32 as case generators above assume float64
```
westonpace, line 1932: Is this a TODO (should we create an issue?) or is it a fact, in which case we can get rid of the `XXX`?

westonpace (approved):
I won't promise to follow the conversion algorithm entirely :) but this looks well tested and the bit shifting routines make sense.

felipecrv (approved):
I can't claim I went through all the details of the implementation, but it looks like the kind of code that doesn't have rarely-taken branches that wouldn't be exercised from the unit tests. LGTM.

cpp/src/arrow/util/decimal_test.cc
pitrou: Because `CheckDecimalFromReal` and `CheckDecimalFromRealIntegerString` now automatically deduce the negative test cases from the positive ones.

cpp/src/arrow/util/basic_decimal.cc
pitrou: Yes!

cpp/src/arrow/util/decimal.cc
pitrou: Good idea, will do.

python/pyarrow/tests/test_compute.py
pitrou: Not sure. I wouldn't create an issue for it as the effort-benefit ratio is probably low.

pitrou:
I think I have addressed all review comments. Would you like to take another look? @westonpace @felipecrv

felipecrv (approved):

cpp/src/arrow/util/decimal_test.cc
```diff
@@ -868,24 +867,29 @@ template <typename T>
...
-            FromFloatTestParam{-1.8446746e+15f, 20, 4, "-1844674627273280.7168"},
-            // Almost 10**38 (minus 2**103)
-            FromFloatTestParam{9.999999e+37f, 38, 0,
-                               "99999986661652122824821048795547566080"},
-            FromFloatTestParam{-9.999999e+37f, 38, 0,
-                               "-99999986661652122824821048795547566080"}};
+    return {
+        // -- Stress the 24 bits of precision of a float
+        // 2**63 + 2**40
+        FromFloatTestParam{9.223373e+18f, 19, 0, "9223373136366403584"},
```
huberylee, line 873: Hi, the expected return value of ``FromFloatTestParam{5.76460752e13f, 18, 4, "57646075230342.3488"}``  is ``57646073774080.0000``, which seems different from the original value, does that meet expectations？

pitrou: What do you call "the original value" in that context?
Both values are actually equal in  `float32`:
```python
>>> np.float32(5.76460752e13) == np.float32(57646073774080)
True
```

huberylee: > What do you call "the original value" in that context? Both values are actually equal in `float32`:
> 
> ```python
> >>> np.float32(5.76460752e13) == np.float32(57646073774080)
> True
> ```

I see what you mean. Thanks!

PR #36009, opened 2023-06-09
GH-35500: [C++][Go][Java][FlightRPC] Add support for result set expiration

### LINKED ISSUE #35500: [C++][Go][FlightRPC] Add support for result set expiration

Based on the proposal in https://docs.google.com/document/d/1jhPyPZSOo2iy0LqIJVUs9KWPyFULVFJXTILDfkadx2g/edit# .
See also the discussion thread: https://lists.apache.org/thread/247z3t06mf132nocngc1jkp3oqglz7jp

Currently, it is undefined whether a client can call DoGet more than once. Clients may want to retry requests, and servers may not want to persist a query result forever. 

### Proposal

Add an expiration time to FlightEndpoint. If present, clients may assume they can retry DoGet requests. Otherwise, clients should avoid retrying DoGet requests.

```protobuf
message FlightEndpoint {
  // In UTC.
  google.protobuf.Timestamp expiration_time = 3;
}
```

This proposal is not a full retry protocol.

Also, add “pre-defined” actions to Flight RPC for working with result sets. These are pre-defined Protobuf messages with standardized encodings for use with DoAction:
* ~~CancelQuery~~ CancelFlightInfo: Asynchronously cancel the execution of a distributed query. (Replaces the equivalent Flight SQL action.)
* ~~RefreshQuery~~ RefreshFlightEndpoint: Request an extension of the expiration of a FlightEndpoint.
* ~~CloseQuery~~ ~CloseFlightInfo: Close a FlightInfo so that the server can clean up resources early.~ (Only `CancelFlightInfo` will be enough.)

This lets the ADBC/JDBC/ODBC drivers for Flight SQL explicitly manage result set lifetimes. These can be used with Flight SQL as regular actions.

### Prior Art

* [Dremio](https://docs.dremio.com/software/advanced-administration/job-results-cleanup/): job results are cleaned up after a set time.
* [Google BigQuery Storage](https://github.com/googleapis/googleapis/blob/master/google/cloud/bigquery/storage/v1/storage.proto): read sessions include an explicit expiration time.
* [Snowflake](https://docs.snowflake.com/en/user-guide/querying-persisted-results.html): result sets persist for 24 hours.

### Component(s)

C++, FlightRPC

kou:
### Rationale for this change

Currently, it is undefined whether a client can call DoGet more than once. Clients may want to retry requests, and servers may not want to persist a query result forever.

Add an expiration time to FlightEndpoint. If present, clients may assume they can retry DoGet requests. Otherwise, clients should avoid retrying DoGet requests.

This is not a full retry protocol.

Also, add "pre-defined" actions to Flight RPC for working with result sets. These are pre-defined Protobuf messages with standardized encodings for use with DoAction:

  * CancelFlightInfo: Asynchronously cancel the execution of a distributed query. (Replaces the equivalent Flight SQL action.)
  * RefreshFlightEndpoint: Request an extension of the expiration of a FlightEndpoint.

This lets the ADBC/JDBC/ODBC drivers for Flight SQL explicitly manage result set lifetimes. These can be used with Flight SQL as regular actions.

#### Backward compatibility

Flight SQL's CancelQuery is deprecated by Flight RPC's CancelFlightInfo. But some clients may not be able to migrate to CancelFlightInfo entirely.

If a client can assume that a server requires 13.0.0 or later, the client can always use CancelFlightInfo. Otherwise, the client need to use CancelQuery (for old server) and/or CancelFlightInfo (for newer server).

A server needs to implement only CancelFlightInfo. Because Flight SQL server libraries provide the default CancelQuery implementation that delegates to CancelFlightInfo. Clients can use both of Flight SQL's CancelQuery and Flight RPC's CancelFLightInfo by this feature.

### Are these changes tested?

Yes.

Yes.

* Closes: #35500


dev/archery/archery/integration/runner.py
```diff
@@ -435,6 +435,37 @@ def run_all_tests(with_cpp=True, with_java=True, with_js=True,
...
+            "expiration_time:refresh_flight_info",
+            description=("Ensure FlightEndpoint.expiration_time and "
+                         "RefreshFlightEndpoint are working as expected."),
+            skip={"Java", "Go", "JS", "C#", "Rust"},
+        ),
+        Scenario(
+            "expiration_time:close_flight_info",
+            description=("Ensure FlightEndpoint.expiration_time and "
+                         "CloseFlightInfo are working as expected."),
+            skip={"Java", "Go", "JS", "C#", "Rust"},
```
lidavidm, line 467: I can work on Java next week

format/Flight.proto
```diff
@@ -321,6 +348,13 @@ message FlightEndpoint {
    * represent redundant and/or load balanced services.
    */
   repeated Location location = 2;
+
+  /*
+   * Expiration time of this stream. If present, clients may assume
+   * they can retry DoGet requests. Otherwise, clients should avoid
+   * retrying DoGet requests.
```
lidavidm, line 355: I wonder if wording it as "it is application-defined whether DoGet requests may be retried" might be better since technically that is how it was before

cpp/src/arrow/flight/serialization_internal.cc
```diff
@@ -147,6 +169,15 @@ Status ToProto(const FlightEndpoint& endpoint, pb::FlightEndpoint* pb_endpoint)
   for (const Location& location : endpoint.locations) {
     RETURN_NOT_OK(ToProto(location, pb_endpoint->add_location()));
   }
+  if (endpoint.expiration_time.has_value()) {
+    const auto expiration_time = endpoint.expiration_time.value();
+    const auto since_epoch = expiration_time.time_since_epoch();
+    const auto since_epoch_ns =
+        std::chrono::duration_cast<std::chrono::nanoseconds>(since_epoch).count();
+    auto pb_expiration_time = pb_endpoint->mutable_expiration_time();
+    pb_expiration_time->set_seconds(since_epoch_ns / 1000000000);
```
lidavidm, line 178: nit: constant for the nanoseconds to seconds conversion? (Do we already have one?)

cpp/src/arrow/flight/client.cc
```diff
@@ -569,6 +569,36 @@ Status FlightClient::DoAction(const FlightCallOptions& options, const Action& ac
   return DoAction(options, action).Value(results);
 }
 
+arrow::Result<std::unique_ptr<ActionCancelFlightInfoResult>>
+FlightClient::CancelFlightInfo(const FlightCallOptions& options, const FlightInfo& info) {
+  ARROW_ASSIGN_OR_RAISE(auto body, info.SerializeToString());
+  Action action{ActionType::kCancelFlightInfo.type, Buffer::FromString(body)};
+  ARROW_ASSIGN_OR_RAISE(auto stream, DoAction(options, action));
+  ARROW_ASSIGN_OR_RAISE(auto result, stream->Next());
```
lidavidm, line 577: We should drain the rest of the DoAction stream since that's the only way we can get the server error (if any). You can see how this is done in the Flight SQL codebase: https://github.com/apache/arrow/blob/766e25440250dde9117e19245389badb5ed7678c/cpp/src/arrow/flight/sql/client.cc#L131-L137

cpp/src/arrow/flight/sql/client.h
```diff
@@ -324,6 +324,7 @@ class ARROW_FLIGHT_SQL_EXPORT FlightSqlClient {
   Status Rollback(const FlightCallOptions& options, const Savepoint& savepoint);
 
   /// \brief Explicitly cancel a query.
+  /// Deprecated since 13.0.0. Use FlightClient::CancelFlightInfo() instead.
```
lidavidm, line 327: Doxygen has a [`\deprecated`](https://www.doxygen.nl/manual/commands.html#cmddeprecated) command

format/Flight.proto
```diff
@@ -189,6 +191,31 @@ message Result {
   bytes body = 1;
 }
 
+/*
+ * The result of the CancelFlightInfo action.
+ *
+ * The result should be stored in Result.body.
+ */
+message ActionCancelFlightInfoResult {
```
kou, line 199: We may want to remove `Action`:

```suggestion
message CancelFlightInfoResult {
```

format/Flight.proto
```diff
@@ -189,6 +191,31 @@ message Result {
...
+  enum CancelResult {
```
kou, line 200: We may want to remove `Cancel`:

```suggestion
  enum Result {
```

Or we may want to move this to top-level and return this enum directly instead of wrapping by `ActionCancelFlightInfoResult` message.

format/Flight.proto
```diff
@@ -189,6 +191,31 @@ message Result {
...
+    // The cancellation status is unknown. Servers should avoid using
+    // this value (send a NOT_FOUND error if the requested query is
+    // not known). Clients can retry the request.
+    CANCEL_RESULT_UNSPECIFIED = 0;
```
kou, line 204: `CANCEL_RESULT_` prefix may be redundant.

```suggestion
    UNSPECIFIED = 0;
```

kou: We should use prefix.

https://protobuf.dev/programming-guides/style/#enums

> Prefer prefixing enum values instead of surrounding them in an enclosing message.

cpp/src/arrow/flight/client.cc
lidavidm: I think `ReadResult` can only be in an `_internal.h` header because it uses Protobuf. But `DrainResultStream` could become `ResultStream::Drain`.

format/Flight.proto
lidavidm: +1 from me

format/Flight.proto
lidavidm: At least in Protobuf/gRPC itself, you can only return messages and not bare enums

format/Flight.proto
lidavidm: Yes, the reason (in case you didn't find it) is that enums in Protobuf use C-style namespacing, which is carried over into generated code. So it will end up as `ActionCancelFlightInfoResult::[CANCEL_RESULT_]UNSPECIFIED`, and so if you ever need a second enum, you need to prefix the name.

format/Flight.proto
zeroshade: +1 from me too

zeroshade (commented):
@kou would you like me contribute the Go implementation? Or do you have that in the works? I have the time to do so if you need / want the help.

kou:
@zeroshade Thanks for the offer! I have the Go implementation in local and I just need to implement one more integration test to cover all cases implemented in the C++ implementation. I'll be able to complete it and push to this branch today. But the Go implementation is needed to be cleaned up. Please review and/or push improvements to this branch after I push the Go implementation.

cpp/src/arrow/flight/sql/server.h
```diff
@@ -594,7 +594,16 @@ class ARROW_FLIGHT_SQL_EXPORT FlightSqlServerBase : public FlightServerBase {
   virtual Status EndTransaction(const ServerCallContext& context,
                                 const ActionEndTransactionRequest& request);
 
+  /// \brief Attempt to explicitly cancel a FlightInfo.
+  /// \param[in] context  The call context.
+  /// \param[in] info     The FlightInfo to cancel.
+  /// \return             The cancellation result.
+  virtual arrow::Result<CancelFlightInfoResult> CancelFlightInfo(
+      const ServerCallContext& context, const FlightInfo& info);
```
kou, line 602: This is not tested.

We may want to add `flight_sql:expiration_time:*` integration tests that use `CancelFlightInfo`/`CloseFlightInfo`/`RefreshFlightEndpoint`. But most of codes will be duplicated with `expiration_time:*` integration tests...

@lidavidm What do you think about this?

kou:
@zeroshade I've pushed the Go implementation but it's still dirty and remains some TODOs in the Flight SQL part...
(We need to consider how to store `FlightInfo`/`FlightEndpoint` for `CancelFlightInfo`/`CloseFlightInfo`/`RefreshFlightEndpoint` for the Flight SQL part. See also: https://github.com/apache/arrow/pull/36009#discussion_r1228899238 )

TODO (helps are welcome!):
* Update document
* Clean up the C++ implementation
* Clean up the Go implementation
* Consider how to keep backward compatibility the Flight SQL API

cpp/src/arrow/flight/sql/server.h
lidavidm: I don't think we need separate integration tests for Flight SQL if we're testing them in Flight RPC. Just unit tests would be sufficient.

lidavidm:
For Flight SQL, we could mark the old Cancel action deprecated in Protobuf and in the APIs?

go/arrow/flight/flightsql/client_test.go
```diff
@@ -60,6 +60,18 @@ func (m *FlightServiceClientMock) AuthenticateBasicToken(_ context.Context, user
 	return args.Get(0).(context.Context), args.Error(1)
 }
 
+func (m *FlightServiceClientMock) CancelFlightInfo(ctx context.Context, info *FlightInfo, opts ...grpc.CallOption) (CancelFlightInfoResult, error) {
+	panic("not implemented") // TODO: Implement
```
zeroshade, line 64: can this just follow the pattern above of `args := m.Called(info, opts)` and `return args.Get(0).(CancelFlightInfoResult), args.Error(1)` so we can use the mock to create the tests?

go/arrow/flight/server.go
```diff
@@ -49,9 +49,36 @@ type (
 	Action                          = flight.Action
 	ActionType                      = flight.ActionType
 	Result                          = flight.Result
+	CancelFlightInfoResult    = flight.CancelFlightInfoResult
+	CancelResult = flight.CancelResult
```
zeroshade, line 53: indentation

zeroshade (changes requested):

format/Flight.proto
kou: @lidavidm ping

cpp/src/arrow/flight/client.cc
kou: @lidavidm ping

lidavidm: Ah, sorry.

The `Any` was a choice by the original Dremio developers, it gives you more type safety in Protobuf but is largely redundant, yes. (That said, even when wrapped in Any, we can still provide a function/method to parse it right?)

lidavidm: But I would prefer not to wrap in Any, so we can keep it like this.

format/Flight.proto
lidavidm: > Can we use CancelStatus or something instead of CancelResult to use Result in CancelResult and ActionCancelFlightInfoResult?

Sorry, what are you proposing here?

> I moved CancelResult to top-level to reduce ActionCancelFlightInfoResult::... prefix.

I think that's fine.

cpp/src/arrow/flight/client.cc
kou: > even when wrapped in Any, we can still provide a function/method to parse it right?

Right.

> But I would prefer not to wrap in Any, so we can keep it like this.

OK! Thanks!

format/Flight.proto
kou: > Sorry, what are you proposing here?

I wanted to say that how about the following (or other better name)?

```diff
index cda7dae31..5ccdd5b9a 100644
--- a/format/Flight.proto
+++ b/format/Flight.proto
@@ -196,20 +196,20 @@ message Result {
  *
  * This is used by CancelFLightInfoResult.result.
  */
-enum CancelResult {
+enum CancelStatus {
   // The cancellation status is unknown. Servers should avoid using
   // this value (send a NOT_FOUND error if the requested query is
   // not known). Clients can retry the request.
-  CANCEL_RESULT_UNSPECIFIED = 0;
+  CANCEL_STATUS_UNSPECIFIED = 0;
   // The cancellation request is complete. Subsequent requests with
   // the same payload may return CANCELLED or a NOT_FOUND error.
-  CANCEL_RESULT_CANCELLED = 1;
+  CANCEL_STATUS_CANCELLED = 1;
   // The cancellation request is in progress. The client may retry
   // the cancellation request.
-  CANCEL_RESULT_CANCELLING = 2;
+  CANCEL_STATUS_CANCELLING = 2;
   // The query is not cancellable. The client should not retry the
   // cancellation request.
-  CANCEL_RESULT_NOT_CANCELLABLE = 3;
+  CANCEL_STATUS_NOT_CANCELLABLE = 3;
 }
 
 /*
@@ -218,7 +218,7 @@ enum CancelResult {
  * The result should be stored in Result.body.
  */
 message CancelFlightInfoResult {
-  CancelResult result = 1;
+  CancelStatus status = 1;
 }
 
 /*
```

lidavidm: Ah, to avoid confusion with Arrow `Result`? 

I guess Arrow also uses `Status`, so both are the same to me. If using `CancelStatus` is clearer to you, then it sounds good to me.

kou: > Ah, to avoid confusion with Arrow `Result`?

No. I just want to avoid `Result` duplication in `CancelFlightInfoResult` and `CancelResult`.

kou: BTW, we have `message Result` in `Flight.proto`. It may confuse us with Arrow `Result`...

lidavidm: > BTW, we have `message Result` in `Flight.proto`. It may confuse us with Arrow `Result`...

Yes, the Flight one actually happened to come first, but it's unfortunate...

> No. I just want to avoid Result duplication in CancelFlightInfoResult and CancelResult.

Ok. I think `status` is also good then.

kou: OK!

cpp/src/arrow/flight/sql/server.h
kou: OK. Done.

kou:
> For Flight SQL, we could mark the old Cancel action deprecated in Protobuf and in the APIs?

It makes sense. Done.

cpp/src/arrow/flight/types.h
```diff
@@ -630,6 +697,11 @@ class ARROW_FLIGHT_EXPORT ResultStream {
 
   ARROW_DEPRECATED("Deprecated in 8.0.0. Use Result-returning overload instead.")
   Status Next(std::unique_ptr<Result>* info);
+
+  /// \brief Read and drop the all rest messages to get error from a server.
```
lidavidm, line 701:
```suggestion
  /// \brief Read and drop the remaining messages to get the error (if any) from a server.
```

lidavidm (commented):
LGTM (I only looked at C++)

I'll start working on Java side.

cpp/src/arrow/flight/types.h
kou: Thanks!

go/arrow/flight/client.go
```diff
@@ -348,6 +352,92 @@ func (c *client) Authenticate(ctx context.Context, opts ...grpc.CallOption) erro
...
+	res, err := stream.Recv()
+	if err != nil {
+		return
+	}
+	if err = proto.Unmarshal(res.Body, &result); err != nil {
+		return
+	}
+	for {
+		_, err = stream.Recv()
+		if errors.Is(err, io.EOF) {
```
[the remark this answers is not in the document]
kou: Good catch! I missed `err = nil`.

go/arrow/flight/client.go
```diff
@@ -348,6 +352,92 @@ func (c *client) Authenticate(ctx context.Context, opts ...grpc.CallOption) erro
...
+			break
+		}
+		if err != nil {
+			return nil, err
+		}
+	}
+	return &refreshedEndpoint, nil
+}
```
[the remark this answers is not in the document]
kou: You're right.

go/arrow/flight/flightsql/client_test.go
kou: OK. I've changed to use mock and added tests of them.

go/arrow/flight/server.go
kou: Oh, sorry.
I've ran `go fmt arrow/flight/server.go` in `go/`.

BTW, I got many change when I run `go fmt ./...`. What command do you use for formatting?

go/arrow/flight/client.go
```diff
@@ -348,10 +352,87 @@ func (c *client) Authenticate(ctx context.Context, opts ...grpc.CallOption) erro
 	return c.authHandler.Authenticate(ctx, &clientAuthConn{stream})
 }
 
+// Ensure the result of a DoAction is fully consumed
+func ReadUntilEOF(stream FlightService_DoActionClient) error {
```
zeroshade, line 356: do we need this to be exposed and exported? If not, I'd prefer to keep it unexported unless you think it'll be specifically useful for users.

go/arrow/flight/client.go
```diff
@@ -348,10 +352,87 @@ func (c *client) Authenticate(ctx context.Context, opts ...grpc.CallOption) erro
...
+	action.Body, err = proto.Marshal(info)
+	if err != nil {
+		return
+	}
+	stream, err := c.DoAction(ctx, &action, opts...)
+	if err != nil {
+		return
+	}
+	err = ReadUntilEOF(stream)
+	return
```
zeroshade, line 409: `return ReadUntilEOF(stream)`?

go/arrow/flight/flightsql/server.go
```diff
@@ -640,7 +668,17 @@ type Server interface {
 	// EndTransaction commits or rollsback a transaction
 	EndTransaction(context.Context, ActionEndTransactionRequest) error
 	// CancelQuery attempts to explicitly cancel a query
+	// Deprecated: Since 13.0.0. If you can require all clients
+	// use 13.0.0 or later, you can use only CancelFlightInfo and
+	// you don't need to use CancelQuery. Otherwise, you may need
+	// to use CancelQuery and/or CancelFlightInfo.
```
zeroshade, line 674: Having `flightSqlServer` have its own implementation of `CancelQuery` as I mentioned above also has the nice effect here of allowing users to switch entirely to `CancelFlightInfo` safely regardless of the client as the server will do the right thing then.

cpp/src/arrow/flight/integration_tests/test_integration.cc
```diff
@@ -410,6 +413,470 @@ class OrderedScenario : public Scenario {
...
+            statuses_[index].cancelled = true;
+            cancel_status = CancelStatus::kCancelled;
+          }
+        } else {
+          cancel_status = CancelStatus::kNotCancellable;
+        }
+        auto cancel_result = CancelFlightInfoResult{cancel_status};
+        ARROW_ASSIGN_OR_RAISE(auto serialized, cancel_result.SerializeToString());
+        results.push_back(Result{Buffer::FromString(std::move(serialized))});
+      }
```
lidavidm, line 532: Hmm, this exposes something a little underspecified - the Flight SQL equivalent always returned a single result. Should we stick to that?

lidavidm: Per-endpoint results may be more flexible, but I'm not sure it's a useful level of flexibility

cpp/src/arrow/flight/integration_tests/test_integration.cc
```diff
@@ -410,6 +413,470 @@ class OrderedScenario : public Scenario {
...
+    }
+    // Re-reads after expired
+    for (const auto& endpoint : info->endpoints()) {
+      if (!endpoint.expiration_time.has_value()) {
+        continue;
+      }
+      const auto& expiration_time = endpoint.expiration_time.value();
+      if (expiration_time > Timestamp::clock::now()) {
+        std::this_thread::sleep_for(expiration_time - Timestamp::clock::now());
+      }
```
lidavidm, line 647: I wonder if this might be flaky in CI (at least in the prior loop), since if it takes more than 2 seconds somehow it'll fail above. Maybe there's no need to test the actual expiration of the ticket? Or, we can treat the expiration as a counter instead of a real expiration time to make things not dependent on the clock.

cpp/src/arrow/flight/integration_tests/test_integration.cc
lidavidm: Actually, the client only ever reads the first result anyways right? So was this meant to only send a single result?

lidavidm:
Java support: https://github.com/kou/arrow/pull/12

Though I haven't tested the integration bits yet

lidavidm:
Ok, looks like the Java-C++ integration passes locally for me

go/arrow/flight/flightsql/server.go
kou: You're right.
The document is wrong. Users need only implement `CancelFlightInfo` and it works with `CancelQuery` and `CancelFlightInfo` from clients.

cpp/src/arrow/flight/integration_tests/test_integration.cc
kou: Done.

kou:
Thanks!
I've merged the Java implementation!

(You can push your changes to this branch directly. :-)

java/flight/flight-sql/src/main/java/org/apache/arrow/flight/sql/FlightSqlProducer.java
```diff
@@ -383,13 +424,38 @@ default void beginTransaction(ActionBeginTransactionRequest request, CallContext
...
    * Explicitly cancel a query.
    *
    * @param info     The FlightInfo of the query to cancel.
    * @param context  Per-call context.
    * @param listener Whether cancellation succeeded.
+   * @deprecated Prefer {@link #cancelFlightInfo(FlightInfo, CallContext, StreamListener)}.
    */
+  @Deprecated
   default void cancelQuery(FlightInfo info, CallContext context, StreamListener<CancelResult> listener) {
     listener.onError(CallStatus.UNIMPLEMENTED.toRuntimeException());
```
kou, line 460: @lidavidm How about providing the default implementation that uses `cancelFlightInfo` so that users don't need to implement both of `cancelQuery()` and `cancelFlightInfo()` to support both of  the `CancelQuery`/`CancelFlightInfo` actions?
(The C++/Go implementations do it.)

lidavidm: Ah, whoops. I'll do that!

kou:
Hmm. Some integration tests failed:

https://github.com/apache/arrow/actions/runs/5328790706/jobs/9653800237?pr=36009#step:7:14620

```text
 -- Arrow Fatal Error --
NotImplemented: StatementExecution failed: Flight returned unimplemented error, with message: CancelFlightInfo not implemented
```


https://github.com/apache/arrow/actions/runs/5328790706/jobs/9653800237?pr=36009#step:7:14592

```text
Invalid flight: expired
```

lidavidm:
Hmm, I wonder if the expirations are because Java runs too slowly in CI.

java/flight/flight-sql/src/main/java/org/apache/arrow/flight/sql/FlightSqlProducer.java
lidavidm: Done!

lidavidm:
The Java integration passes locally for me, so I do think it's a timing issue. It might be enough to just test that the expiration time and refresh gets passed through and parsed properly, and not necessarily need to implement the expiration semantics? Since what we care about in the integration tests is just that the data makes it through properly. (And then we might use fixed timestamps for the expirations so all implementations know what values to expect.)

go/arrow/flight/flightsql/server.go
```diff
@@ -511,8 +511,36 @@ func (BaseServer) BeginSavepoint(context.Context, ActionBeginSavepointRequest) (
 	return nil, status.Error(codes.Unimplemented, "BeginSavepoint not implemented")
 }
 
-func (BaseServer) CancelQuery(context.Context, ActionCancelQueryRequest) (CancelResult, error) {
-	return CancelResultUnspecified, status.Error(codes.Unimplemented, "CancelQuery not implemented")
+func (b *BaseServer) CancelQuery(context context.Context, request ActionCancelQueryRequest) (CancelResult, error) {
+	result, err := b.CancelFlightInfo(context, request.GetInfo())
+	if err != nil {
+		return CancelResultUnspecified, err
+	}
```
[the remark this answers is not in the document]
zeroshade: So what we could do is check whether or not the server they passed has a `CancelQuery` method, and if it does then we dispatch to that. Otherwise, default to calling `CancelFlightInfo`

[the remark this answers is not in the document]
kou: You mean that we keep backward compatibility for users who are using `CancelQuery`, right?
If so, could you push a code for it to this branch? I don't know much about `reflect` package in Go...

kou:
OK. Let's remove expired checks.

kou:
Ah, should we remove the following checks too?

We can read multiple times from endpoints that have expiration time:

```diff
diff --git a/cpp/src/arrow/flight/integration_tests/test_integration.cc b/cpp/src/arrow/flight/integration_tests/test_integration.cc
index f253f8db7..b8a75ad6d 100644
--- a/cpp/src/arrow/flight/integration_tests/test_integration.cc
+++ b/cpp/src/arrow/flight/integration_tests/test_integration.cc
@@ -617,26 +617,20 @@ class ExpirationTimeDoGetScenario : public Scenario {
     ARROW_ASSIGN_OR_RAISE(
         auto info, client->GetFlightInfo(FlightDescriptor::Command("expiration_time")));
     std::vector<std::shared_ptr<arrow::Table>> tables;
-    // First read from all endpoints
     for (const auto& endpoint : info->endpoints()) {
       ARROW_ASSIGN_OR_RAISE(auto reader, client->DoGet(endpoint.ticket));
       ARROW_ASSIGN_OR_RAISE(auto table, reader->ToTable());
-      tables.push_back(table);
-    }
-    // Re-reads only from endpoints that have expiration time
-    for (const auto& endpoint : info->endpoints()) {
-      if (endpoint.expiration_time.has_value()) {
-        ARROW_ASSIGN_OR_RAISE(auto reader, client->DoGet(endpoint.ticket));
-        ARROW_ASSIGN_OR_RAISE(auto table, reader->ToTable());
-        tables.push_back(table);
+      if (tables.size() == 0) {
+        if (endpoint.expiration_time.has_value()) {
+          return Status::Invalid("The first endpoint must not have expiration time");
+        }
       } else {
-        auto reader = client->DoGet(endpoint.ticket);
-        if (reader.ok()) {
-          return Status::Invalid(
-              "Data that doesn't have expiration time "
-              "shouldn't be readable multiple times");
+        if (!endpoint.expiration_time.has_value()) {
+          return Status::Invalid("The ", tables.size(),
+                                 "-th endpoint must have expiration time");
         }
       }
+      tables.push_back(table);
     }
     ARROW_ASSIGN_OR_RAISE(auto table, ConcatenateTables(tables));
 
@@ -645,13 +639,9 @@ class ExpirationTimeDoGetScenario : public Scenario {
     ARROW_ASSIGN_OR_RAISE(auto builder,
                           RecordBatchBuilder::Make(schema, arrow::default_memory_pool()));
     auto number_builder = builder->GetFieldAs<UInt32Builder>(0);
-    // First reads
     ARROW_RETURN_NOT_OK(number_builder->Append(0));
     ARROW_RETURN_NOT_OK(number_builder->Append(1));
     ARROW_RETURN_NOT_OK(number_builder->Append(2));
-    // Re-reads only from endpoints that have expiration time
-    ARROW_RETURN_NOT_OK(number_builder->Append(1));
-    ARROW_RETURN_NOT_OK(number_builder->Append(2));
     ARROW_ASSIGN_OR_RAISE(auto expected_record_batch, builder->Flush());
     std::vector<std::shared_ptr<RecordBatch>> expected_record_batches{
         expected_record_batch};
```

We can read from refreshed endpoints:

```diff
diff --git a/cpp/src/arrow/flight/integration_tests/test_integration.cc b/cpp/src/arrow/flight/integration_tests/test_integration.cc
index f253f8db7..5f212ba2d 100644
--- a/cpp/src/arrow/flight/integration_tests/test_integration.cc
+++ b/cpp/src/arrow/flight/integration_tests/test_integration.cc
@@ -779,13 +779,6 @@ class ExpirationTimeRefreshFlightEndpointScenario : public Scenario {
   Status RunClient(std::unique_ptr<FlightClient> client) override {
     ARROW_ASSIGN_OR_RAISE(auto info,
                           client->GetFlightInfo(FlightDescriptor::Command("expiration")));
-    std::vector<std::shared_ptr<arrow::Table>> tables;
-    // First read from all endpoints
-    for (const auto& endpoint : info->endpoints()) {
-      ARROW_ASSIGN_OR_RAISE(auto reader, client->DoGet(endpoint.ticket));
-      ARROW_ASSIGN_OR_RAISE(auto table, reader->ToTable());
-      tables.push_back(table);
-    }
     // Refresh all endpoints that have expiration time
     std::vector<FlightEndpoint> refreshed_endpoints;
     Timestamp max_expiration_time;
@@ -811,56 +804,7 @@ class ExpirationTimeRefreshFlightEndpointScenario : public Scenario {
       }
       refreshed_endpoints.push_back(std::move(refreshed_endpoint));
     }
-    // Expire all not refreshed endpoints
-    {
-      std::vector<Timestamp> refreshed_expiration_times;
-      for (const auto& endpoint : refreshed_endpoints) {
-        refreshed_expiration_times.push_back(endpoint.expiration_time.value());
-      }
-      std::sort(refreshed_expiration_times.begin(), refreshed_expiration_times.end());
-      if (refreshed_expiration_times[0] < max_expiration_time) {
-        return Status::Invalid(
-            "One or more refreshed expiration time "
-            "are shorter than original expiration time\n",
-            "Original:  ", max_expiration_time.time_since_epoch().count(), "\n",
-            "Refreshed: ", refreshed_expiration_times[0].time_since_epoch().count(),
-            "\n");
-      }
-      if (max_expiration_time > Timestamp::clock::now()) {
-        std::this_thread::sleep_for(max_expiration_time - Timestamp::clock::now());
-      }
-    }
-    // Re-reads only from refreshed endpoints
-    for (const auto& endpoint : refreshed_endpoints) {
-      ARROW_ASSIGN_OR_RAISE(auto reader, client->DoGet(endpoint.ticket));
-      ARROW_ASSIGN_OR_RAISE(auto table, reader->ToTable());
-      tables.push_back(table);
-    }
-    ARROW_ASSIGN_OR_RAISE(auto table, ConcatenateTables(tables));
 
-    // Build expected table
-    auto schema = arrow::schema({arrow::field("number", arrow::uint32(), false)});
-    ARROW_ASSIGN_OR_RAISE(auto builder,
-                          RecordBatchBuilder::Make(schema, arrow::default_memory_pool()));
-    auto number_builder = builder->GetFieldAs<UInt32Builder>(0);
-    // First reads
-    ARROW_RETURN_NOT_OK(number_builder->Append(0));
-    ARROW_RETURN_NOT_OK(number_builder->Append(1));
-    ARROW_RETURN_NOT_OK(number_builder->Append(2));
-    // Re-reads only from refreshed endpoints
-    ARROW_RETURN_NOT_OK(number_builder->Append(1));
-    ARROW_RETURN_NOT_OK(number_builder->Append(2));
-    ARROW_ASSIGN_OR_RAISE(auto expected_record_batch, builder->Flush());
-    std::vector<std::shared_ptr<RecordBatch>> expected_record_batches{
-        expected_record_batch};
-    ARROW_ASSIGN_OR_RAISE(auto expected_table,
-                          Table::FromRecordBatches(expected_record_batches));
-
-    // Check read data
-    if (!table->Equals(*expected_table)) {
-      return Status::Invalid("Read data isn't expected\n", "Expected:\n",
-                             expected_table->ToString(), "Actual:\n", table->ToString());
-    }
     return Status::OK();
   }
 };
```

lidavidm:
That sounds good to me.

go/arrow/flight/flightsql/server.go
```diff
@@ -927,10 +937,78 @@ func (f *flightSqlServer) ListActions(_ *flight.Empty, stream flight.FlightServi
...
+			err    error
+		)
+
+		if err = proto.Unmarshal(cmd.Body, &info); err != nil {
+			return status.Errorf(codes.InvalidArgument, "unable to unmarshal FlightInfo for CancelFlightInfo: %s", err.Error())
+		}
+
+		if result, err = f.srv.CancelFlightInfo(stream.Context(), &info); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
kou: Thanks! It works!

lidavidm:
I'll fix up the Java in a moment

format/Flight.proto
```diff
@@ -183,6 +183,15 @@ message Action {
   bytes body = 2;
 }
 
+/*
+ * The request of the RenewFlightEndpoint action.
+ *
+ * The request should be stored in Action.body.
+ */
+message RenewFlightEndpointRequest {
```
lidavidm, line 191: Hmm, at this point should we just have request/result messages for all 3 actions?

lidavidm: Ah, I saw the ML message just now, ignore me.

kou: Ah, sorry. I'm still changing now. I'll add `CancelFlightInfoRequest` later.

But I didn't have a plan to add `RenewFlightEndpointResult`. Should we add it?

NOTE: The ML message: https://lists.apache.org/thread/ofw6bydhyhcqqn3t8013gx0k6fbw8lvv

> * Remove CloseFlightInfo (if nobody objects it)
> * RefreshFlightEndpoint ->
>    RenewFlightEndpoint
> * RenewFlightEndpoint(FlightEndpoint) ->
>   RenewFlightEndpoint(RenewFlightEndpointRequest)
> * CancelFlightInfo(FlightInfo) ->
>   CancelFlightInfo(CancelFlightInfoRequest)

lidavidm: I think your plan on the ML is fine!

kou: OK!

java/flight/flight-core/src/main/java/org/apache/arrow/flight/RenewFlightEndpointRequest.java
```diff
@@ -0,0 +1,70 @@
...
+
+  public RenewFlightEndpointRequest(FlightEndpoint endpoint) {
+    this.endpoint = Objects.requireNonNull(endpoint);
+  }
+
+  RenewFlightEndpointRequest(Flight.RenewFlightEndpointRequest proto) throws URISyntaxException {
+    this(new FlightEndpoint(proto.getEndpoint()));
+  }
+
+  public FlightEndpoint getFlightEndpoint() {
```
kou, line 39: @lidavidm How about renaming this to `GetEndpoitn()` because `FlightInfo` uses `getEndpoints()` not `getFlightEndpoints()`?

kou (commented):
@lidavidm I've added `CloseFlightInfoRequest` and removed `CloseFlightInfo`.

kou:
https://github.com/apache/arrow/actions/runs/5368022930/jobs/9738596690?pr=36009#step:7:14222

```text
Unknown error
Invalid flight: expired: 1: 2 seconds
```

```text
FAILED TEST: expiration_time:do_get Java producing,  Java consuming
```

Hmm. It seems that the Java implementation takes 2+ seconds for `GetFlightInfo` and `DoGet` on CI. I think that it's too slow even we run it on CI. Anyway, I'll increase expiration time.

kou:
The vote carried: https://lists.apache.org/thread/no26s310qn3v0n5x830d50k598fh0pvr

I'll merge this.

PR #36012, opened 2023-06-09
GH-35859: [Python] Actually change the default row group size to 1Mi

### LINKED ISSUE #35859: [Python] New row_group_size default of 1 Mi not working

In the release notes of Arrow 12 say that the default row group size has been lowered from 64 Mi to 1 Mi. But here's what's happening in practice (PyArrow from conda-forge):

```py
In [1]: import pyarrow.parquet as pq

In [2]: import pyarrow as pa

In [3]: pa.__version__
Out[3]: '12.0.0'

In [12]: pq.write_table(pa.Table.from_pydict({"x": [42] * (64 * 1024 * 1024 + 1)}), "/tmp/x")

In [13]: pq.read_metadata("/tmp/x")
Out[13]:
<pyarrow._parquet.FileMetaData object at 0x7fa6b81eaa70>
  created_by: parquet-cpp-arrow version 12.0.0
  num_columns: 1
  num_rows: 67108865
  num_row_groups: 2  # <=============
  format_version: 2.6
  serialized_size: 485

In [14]: pq.write_table(pa.Table.from_pydict({"x": [42] * (64 * 1024 * 1024 + 0)}), "/tmp/x")

In [15]: pq.read_metadata("/tmp/x")
Out[15]:
<pyarrow._parquet.FileMetaData object at 0x7fa6b823d490>
  created_by: parquet-cpp-arrow version 12.0.0
  num_columns: 1
  num_rows: 67108864
  num_row_groups: 1  # <=============
  format_version: 2.6
  serialized_size: 376
```

### Component(s)

Python

westonpace:
### Rationale for this change

In #34280 the default row group size was changed to 1Mi.  However, this was accidentally reverted (for python, but not C++) in #34435 

The problem is that there is both an "absolute max row group size for the writer" and a "row group size to use for this table"  The pyarrow user is unable to set the former property.

The behavior in pyarrow was previously "If no value is given in the call to write_table then don't specify anything and let the absolute max apply"

The first fix changed the absolute max to 1Mi.  However, this made it impossible for the user to use a larger row group size.  The second fix changed the absolute max back to 64Mi.  However, this meant the default didn't change.

This change leaves the absolute max at 64Mi.  However, if the user does not specify a row group size, we no longer "just use the table size" and instead use 1Mi.

### Are these changes tested?

Yes, a unit test was added.

Yes, the default row group size now truly changes to 1Mi.  This change was already announced as part of #34280
* Closes: #35859

### DIFF
--- a/python/pyarrow/_parquet.pyx
+++ b/python/pyarrow/_parquet.pyx
@@ -1767,7 +1767,7 @@ cdef class ParquetWriter(_Weakrefable):
             int64_t c_row_group_size
 
         if row_group_size is None or row_group_size == -1:
-            c_row_group_size = ctable.num_rows()
+            c_row_group_size = min(ctable.num_rows(), 1024*1024)
         elif row_group_size == 0:
             raise ValueError('Row group size cannot be 0')
         else:
--- a/python/pyarrow/_parquet.pyx
+++ b/python/pyarrow/_parquet.pyx
@@ -42,6 +42,8 @@ from pyarrow.lib import (ArrowException, NativeFile, BufferOutputStream,
 
 cimport cpython as cp
 
+_DEFAULT_ROW_GROUP_SIZE = 1024*1024
+_MAX_ROW_GROUP_SIZE = 64*1024*1024
 
 cdef class Statistics(_Weakrefable):
     """Statistics for a single column in a single row group."""
@@ -1595,7 +1597,7 @@ cdef shared_ptr[WriterProperties] _create_writer_properties(
     # The user can always specify a smaller row group size (and the default
     # is smaller) when calling write_table.  If the call to write_table uses
     # a size larger than this then it will be latched to this value.
-    props.max_row_group_length(64*1024*1024)
+    props.max_row_group_length(_MAX_ROW_GROUP_SIZE)
 
     properties = props.build()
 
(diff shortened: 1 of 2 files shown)


python/pyarrow/_parquet.pyx
```diff
@@ -1767,7 +1767,7 @@ cdef class ParquetWriter(_Weakrefable):
             int64_t c_row_group_size
 
         if row_group_size is None or row_group_size == -1:
-            c_row_group_size = ctable.num_rows()
+            c_row_group_size = min(ctable.num_rows(), 1024*1024)
```
mapleFU, line 1770: should we declare a constant rather than 1024*1024 directly?

jorisvandenbossche (approved):

python/pyarrow/_parquet.pyx
westonpace: I'll defer to @jorisvandenbossche only because I don't actually know what a constant should like like in this file (style wise).  I can't find any good examples.

jorisvandenbossche: Either way is fine for me, but so it would like the following (defined at the top of the file, after the imports):

https://github.com/apache/arrow/blob/2ce4a38fcf1fbd0d759b6c2cdf657f27ee6f02a9/python/pyarrow/_dataset.pyx#L42-L46

(but for something that is not reused multiple times, it is less worth it, I think)

westonpace: Ok,I have to fix a lint issue anyways so I'll add a constant real quick for readability.

PR #36015, opened 2023-06-09
GH-36014: [Go] Allow duplicate field names in structs

### LINKED ISSUE #36014: [Go] Handle duplicate field names in structs

This affects the C Data Interface since all schemas are encoded as structs

### Component(s)

Go

lidavidm:
### Rationale for this change

The package segfaulted before.

Allow duplicate field names (in general methods will return the first field if there are duplicates)

### Are these changes tested?

Yes

Yes

* Closes: #36014

### DIFF
--- a/go/arrow/datatype_nested.go
+++ b/go/arrow/datatype_nested.go
@@ -246,13 +246,12 @@ func (*FixedSizeListType) Layout() DataTypeLayout {
 // of relative types, called its fields.
 type StructType struct {
 	fields []Field
-	index  map[string]int
+	index  map[string][]int
 	meta   Metadata
 }
 
 // StructOf returns the struct type with fields fs.
 //
-// StructOf panics if there are duplicated fields.
 // StructOf panics if there is a field with an invalid DataType.
 func StructOf(fs ...Field) *StructType {
 	n := len(fs)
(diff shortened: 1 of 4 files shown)


go/arrow/datatype_nested.go
```diff
@@ -309,17 +309,28 @@ func (t *StructType) Fields() []Field {
...
+// If there are multiple fields with the given name, FieldIdx returns
+// the index of the first first such field.
 func (t *StructType) FieldIdx(name string) (int, bool) {
 	i, ok := t.index[name]
-	return i, ok
+	if ok {
+		return i[0], true
+	}
+	return -1, false
 }
```
zeroshade, line 334: We should add a new method that will return the []int by name lookup

lidavidm: Added the same methods Schema has, though the inconsistency admittedly bugs me (why is FieldsByName ([]Field, bool) but FieldIndices is just []int, and then there's FieldIndices vs FieldIdx -.-)

zeroshade: Blame InfluxDB who created the original impl :stuck_out_tongue: lol

zeroshade (approved):
LGTM

zeroshade:
@lidavidm Actually, with this, can you remove the `skip_category('Go')` from archery/datagen.py for the `generate_duplicate_fieldnames_case()`? Or do you want to enable the integration test as a separate PR?

lidavidm:
Oh! I'll fix that, one moment

lidavidm:
Looks like it passes.

zeroshade:
LGTM!

Issue #35638, opened 2023-05-17
Process parquet rowgroups without Arrow conversion


alippai:
### Describe the usage question you have. Please include as many useful details as  possible.

I'd like to read a Parquet file and append an Arrow table to the new Parquet file created based on the old file and the new table added as a new row group.
Can I read the Parquet rowgroup by rowgroup, decide to drop any or use them and assemble a new Parquet file without doing the (de)serialization to Arrow?

### Component(s)

C++, Parquet, Python

mapleFU:
Seems that you want a "append" syntax, and want to avoid read->covert to arrow->writeback?

I guess current Parquet code cannot support this :-(

alippai:
Yes, something like that. My usecase is writing data to a small parquet file daily, changing the last 3 days. I don’t have exact numbers to support this extra api yet, but wanted to ask first.

I can imagine this is not a common case to keep/drop row groups based on the stats or append new row groups - feel free to close the issue, please

alippai:
Speaking of this... is it a good practice to use row groups instead of hive partitions or is that considered an anti-pattern when speaking of parquet? Would that be a good addition to pyarrow dataset to optionally ensure the parquet rowgroups contains only one partition?

westonpace:
> My usecase is writing data to a small parquet file daily, changing the last 3 days. I don’t have exact numbers to support this extra api yet, but wanted to ask first.
> 
> I can imagine this is not a common case to keep/drop row groups based on the stats or append new row groups - feel free to close the issue, please

I would say that it is a very common thing for users to want to do.  However, parquet is often not the correct layer of abstraction to introduce this capability.  For example, the table formats like Iceberg, Delta Lake, and Hudi have all come up with ways to handle this.

Appending data to parquet groups has been asked for several times.  I've seen arguments that it is simply not possible without rewriting the file (because thrift uses a lot of absolute file offsets and those offsets, in the portions of the file you are not changing, would become invalid) but I have not investigated it thoroughly enough myself.

> Speaking of this... is it a good practice to use row groups instead of hive partitions or is that considered an anti-pattern when speaking of parquet?

There are pros and cons to both.  Row groups can be more flexible than hive partitions (e.g. each row group contains statistics for ALL columns and not just some and row group filters can include things like bloom filters).  However, hive partitions support append operations (you can always add more files to the month=July folder but you can't add more data to an existing row group).

westonpace:
[...]

PR #36027, opened 2023-06-11
GH-36028: [Docs][Parquet] Detailed parquet format support and parquet integration status

### LINKED ISSUE #36028: [Docs][Parquet] Document Parquet implementation status

There is no comprehensive documentation on the publicly available parquet features. The docs of different parquet featuers are scattered across different repositories and their issue queues. Related note: https://github.com/apache/arrow/issues/35638#issuecomment-1584966456 

### Component(s)

Documentation

alippai:
This is a draft skeleton for: https://github.com/apache/arrow/issues/35638#issuecomment-1584966456
* Closes: #36028


alippai:
I'm sure this is too detailed in some places also there is a good chance that it misses many useful features.

My approach was going through the [parquet-format changelog](https://github.com/apache/parquet-format/blob/master/CHANGES.md), the thrift file, the parquet-mr, arrow and arrow-rs issue queue. 

I've intentionally tried to avoid 2.4-2.10 parquet format version info as it'd imply that the 2.9 features include 2.6 features which might not reflect the reality. Instead of that I've tried to focus on the end-user public API and providing a flat list of features instead. I'm open for different approaches as well.

alippai:
@tustvold @mapleFU @westonpace @wgtmac What do you think? Would this be useful?

docs/source/status.rst
```diff
@@ -348,3 +348,107 @@ Notes:
...
+| Format                                    | C++   | Python | Java   | Go    | Rust  |
+|                                           |       |        |        |       |       |
++===========================================+=======+========+========+=======+=======+
+| Basic compression                         |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Brotli, LZ4, ZSTD                         |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| LZ4_RAW                                   |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Hive-style partitioning                   |       |        |        |       |       |
```
tustvold, line 366: I'm not sure I'd consider this a feature of the parquet implementation, it is more a detail of the query engine imo?

docs/source/status.rst
```diff
@@ -348,3 +348,107 @@ Notes:
...
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Column metadata                           |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Chunk metadta                             |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Sorting column                            |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| ColumnIndex statistics                    |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Page statistics                           |       |        |        |       |       |
```
tustvold, line 380: What is this referring to?

docs/source/status.rst
```diff
@@ -348,3 +348,107 @@ Notes:
...
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Partition append / delete                 |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| RowGroup append / delete                  |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Page append / delete                      |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Page CRC32 checksum                       |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Parallel partition processing             |       |        |        |       |       |
```
tustvold, line 422: IMO this is a query engine detail, not a detail of the file format?

docs/source/status.rst
```diff
@@ -348,3 +348,107 @@ Notes:
...
++-------------------------------------------+-------+--------+--------+-------+-------+
+| ColumnIndex statistics                    |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Page statistics                           |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Statistics min_value                      |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| xxHash based bloom filter                 |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| bloom filter length                       |       |        |        |       |       |
```
tustvold, line 386: What is this?

docs/source/status.rst
```diff
@@ -348,3 +348,107 @@ Notes:
...
++-------------------------------------------+-------+--------+--------+-------+-------+
+| FIXED_LEN_BYTE_ARRAY                      |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Complete Delta encoding support           |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Complete RLE support                      |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| BYTE_STREAM_SPLIT                         |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Partition pruning on the partition column |       |        |        |       |       |
```
tustvold, line 402: Again this is a detail of the query engine not the parquet implementation imo

docs/source/status.rst
```diff
@@ -348,3 +348,107 @@ Notes:
...
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Modular encryption                        |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| External column data                      |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Nanosecond support                        |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| FIXED_LEN_BYTE_ARRAY                      |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Complete Delta encoding support           |       |        |        |       |       |
```
tustvold, line 396: I think it would be clearer if you listed the actual encodings

docs/source/status.rst
```diff
@@ -348,3 +348,107 @@ Notes:
...
+| LZ4_RAW                                   |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Hive-style partitioning                   |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| File metadata                             |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| RowGroup metadata                         |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Column metadata                           |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
```
tustvold, line 373: You can't not support this metadata, as otherwise the parquet file can't be read?

docs/source/status.rst
```diff
@@ -348,3 +348,107 @@ Notes:
...
+| Chunk metadta                             |       |        |        |       |       |
```
tustvold, line 374: I'm not sure what this is and how it differs from ColumnChunk

docs/source/status.rst
```diff
@@ -348,3 +348,107 @@ Notes:
...
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Page pruning using statistics             |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Page pruning using bloom filter           |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Partition append / delete                 |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| RowGroup append / delete                  |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Page append / delete                      |       |        |        |       |       |
```
tustvold, line 418: I don't think any support page appending, the rust implementation supports appending column chunks though

docs/source/status.rst
```diff
@@ -348,3 +348,107 @@ Notes:
...
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Parallel RowGroup processing              |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Parallel Page processing                  |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Storage-aware defaults (1)                |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Adaptive concurrency (2)                  |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Adaptive IO when pruning used (3)         |       |        |        |       |       |
```
tustvold, line 432: I'm not sure which parquet reader these features are based off, but my 2 cents is that they indicate a problematic IO abstraction that relies on prefetching heuristics instead of pushing vectored IO down into the IO subsystem (which the Rust implementation does).

docs/source/status.rst
```diff
@@ -348,3 +348,107 @@ Notes:
...
++-------------------------------------------+-------+--------+--------+-------+-------+
+| BYTE_STREAM_SPLIT                         |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Partition pruning on the partition column |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| RowGroup pruning using statistics         |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| RowGroup pruning using bloom filter       |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Page pruning using projection pushdown    |       |        |        |       |       |
```
tustvold, line 408:
```suggestion
| Column Pruning using projection pushdown    |       |        |        |       |       |
```

docs/source/status.rst
```diff
@@ -348,3 +348,107 @@ Notes:
...
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Page pruning using statistics             |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Page pruning using bloom filter           |       |        |        |       |       |
```
tustvold, line 412: I don't think this is supported by the format, bloom filters are per column chunk

docs/source/status.rst
```diff
@@ -348,3 +348,107 @@ Notes:
...
+
+
+Parquet format public API details
+=================================
+
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Format                                    | C++   | Python | Java   | Go    | Rust  |
+|                                           |       |        |        |       |       |
++===========================================+=======+========+========+=======+=======+
+| Basic compression                         |       |        |        |       |       |
```
tustvold, line 360: I wonder if we could have separate tables for supported physical types, encodings and compression

tustvold (commented):
Left some comments, I would personally restrict this table to feature of the actual file readers and not query engine functionality like partitioning and concurrency - imo these are not features of a parquet implementation, but rather a query system. IMO a parquet implementation should not be unilaterally making concurrency decisions, but rather exposing APIs to allow query engines to distribute the work how they deem fit. Similarly partitions are a catalog detail

I would also suggest having separate tables for supported types, encodings, compression and feature support.

docs/source/status.rst
alippai: While arrow-rs needs datafusion for this functionality, arrow handles it without Acero. I don't have strong opinion though

docs/source/status.rst
alippai: Like I said there is a good chance I made a mistake here. I saw this in the thrift spec: ColumnChunk->ColumnMetadata->Statistics

docs/source/status.rst
alippai: It's part of the arrow API in python

docs/source/status.rst
alippai: https://github.com/apache/parquet-format/blob/c766945d90935ebcd4e03fee13aad2b6efcadce3/src/main/thrift/parquet.thrift#L766

docs/source/status.rst
alippai: Same, it's part of the current API, but I agree it's not consistent across implementations.

docs/source/status.rst
alippai: Yes, likely some / most of the Page references should be ColumnChunk. I'll read about this more.

docs/source/status.rst
alippai: I wanted to capture the IO pushdown section https://arrow.apache.org/blog/2022/12/26/querying-parquet-with-millisecond-latency/#io-pushdown but also added more. Likely out of scope as none of the implementations goes into details or provides an API

alippai:
Thanks @tustvold. I'll address the Page vs ColumnChunk issues and other improvement ideas.

docs/source/status.rst
wgtmac: +1 for this.

docs/source/status.rst
wgtmac: I agree with @tustvold, `partitioning` is more like a high-level use case on top of file format.

docs/source/status.rst
wgtmac: Are these intended for the completeness of fields defined in the metadata? If yes, probably they worth a separate table and indicate the states of each field. But that sounds too complicated.

docs/source/status.rst
```diff
@@ -348,3 +348,107 @@ Notes:
 * \(1) Through JNI bindings. (Provided by ``org.apache.arrow.orc:arrow-orc``)
 
 * \(2) Through JNI bindings to Arrow C++ Datasets. (Provided by ``org.apache.arrow:arrow-dataset``)
+
+
+Parquet format public API details
+=================================
+
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Format                                    | C++   | Python | Java   | Go    | Rust  |
```
wgtmac, line 357: The `Java` column could be misleading here. In the arrow repo, there is a java dataset reader to support reading from parquet dataset. If this is for parquet-mr, then it can be easily out of sync.

docs/source/status.rst
wgtmac: Could we organize these items in a layered fashion? Maybe this is a good start point: https://arrow.apache.org/docs/cpp/parquet.html#supported-parquet-features

docs/source/status.rst
tustvold: OMG, they finally added it - amazing, will get that incorporated into the rust writer/reader

wgtmac: > OMG, they finally added it - amazing, will get that incorporated into the rust writer/reader

I just added it recently :) Please note that the latest format is not released yet so the parquet-mr does not know `bloom_filter_length` now.

docs/source/status.rst
zeroshade: Is the intention to indicate that the metadata is available through a public API rather than saying whether or not it is supported in general, since as @tustvold says, you have to support the metadata otherwise the file can't be read.

westonpace:
I'll repeat what the rest said about engine/format differences and maybe offer some clarification.

In C++ the picture is pretty clear, as the APIs tend to be focused on implementation:

There is a C++ parquet module which is purely a parquet reader.
There is a C++ datasets library which, using Acero, offers a lot of features on top of this

In pyarrow the picture is pretty muddled, as the APIs are more focused on user experience:

There is a pyarrow.parquet module, however, many of its features are powered by C++ datasets.  For example, the pyarrow.parquet module can read from S3 even the the C++ parquet module has no concept of S3 (it just has an abstraction for input streams).

So I agree with the others that we should probably not base the features on the python API.

westonpace:
Although...to play devil's advocate...it might be odd when a feature is available in the parquet reader, but not yet exposed in the query component.  For example, there is some row skipping and bloom filters in the C++ parquet reader, but we haven't integrated those into the datasets layer yet.

pitrou:
Agreed with @westonpace.
I created https://issues.apache.org/jira/browse/PARQUET-2310 to propose adding this in the Parquet docs.

alippai:
Moved it to the parquet-site repo: https://github.com/apache/parquet-site/pull/34

PR #36029, opened 2023-06-11
GH-35987: [C++] Pin brew protobuf version to 21

### LINKED ISSUE #35987: [C++] Add support for Protobuf 23.2 on non-Windows

We can't resolve `google::protobuf::internal::ThreadSafeArena::thread_cache_` with Protobuf 23.2 on non-Windows:

```text
 /Applications/Xcode_14.2.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: ../../release/libarrow.a(util_avx2.cc.o) has no symbols
ld: illegal thread local variable reference to regular symbol __ZN6google8protobuf8internal15ThreadSafeArena13thread_cache_E for architecture x86_64
```

I think that https://github.com/protocolbuffers/protobuf/pull/12983 solves this but it's not released yet.

Workarounds:
1. Specify `PROTOBUF_USE_DLLS` explicitly on non-Windows too.
2. Use bundled (old) Protobuf

### Component(s)

C++

js8544:
### Rationale for this change

A temporary workaround for link errors caused by protobuf 23.

Pin brew protobuf version to 21 for now.

### Are these changes tested?

To be tested by CI pipelines.

No

* Closes: #35987


js8544:
@kou CI looks good now :)

kou (approved):
+1

Thanks!

I'll keep #35987 open because this is a workaround.

kou:
It seems that `protobuf@21` and `grpc@1.53` aren't linked from `brew --prefix`.
I confirmed that we can use them by specifying the paths of them explicitly: `cmake ... -DCMAKE_PREFIX_PATH="$(brew --prefix protobuf@21);$(brew --prefix grpc@1.54)"`

Hmm. We may want to set these path implicitly like we did for LLVM: https://github.com/apache/arrow/blob/b4ac585ecb4da610cc64e346e564ca86594aec53/cpp/cmake_modules/FindLLVMAlt.cmake#L46-L56

kou:
#36086

js8544:
Sorry! My bad for not checking this. No wonder why this check has been timing out a lot...

PR #36042, opened 2023-06-13
GH-34338: [Java] Removing the automatic enabling of BaseAllocator.DEBUG on -ea

### LINKED ISSUE #34338: [Java] BaseAllocator.DEBUG should be opt-in as HistoricalLog is immensely expensive and breaks profiling

BaseAllocator.DEBUG is currently enabled automatically any time assertions are enabled via -ea which is typically all test runs and JUnit usages in IDEs, including debugging and profiling based on JUnit tests.

The problem with this is that BaseAllocator.DEBUG causes ArrowBuf to maintain a HistoricalLog with 5 history entries. To achieve this, a highly expensive event containing a full stack trace must be created for each allocation event with the vast majority of them thrown away.

For example: In an application that generates many small (100kB) Arrow data frames, each consisting of around 50 vectors, profiling parts of the test suite related to Arrow generation shows that the creation the StackTrace instances needed for the rolling 5 entry HistoricalLog consume 85% of all program CPU and are responsible for 80% of all Java memory allocation. 

During the applications full Maven test run, which is not a debugging scenario but sees BaseAllocator.DEBUG turned on anyway due to -ea, the cost of HistoricalLog increases the total test run-time from 4 seconds (with arrow.memory.debug.allocator=false ) to 14 seconds, a substantial slow-down for no real value and a large cost when spread over every test and CI run.

Profiling also becomes challenging, because the cost of StackTrace spread out across the program pollutes and distorts profiling results and flame graphs become largely unusable.

These negative effects occur in every Arrow Java application by default, unless explicitly turned off.

I propose that the default of arrow.memory.debug.allocator be reversed to false by removing the automatic enabling of BaseAllocator.DEBUG on -ea. If this is not acceptable, then at minimum the maintenance of HistoricalLog should be separately and specifically disabled.

If discoverability is a dominating concern, then in debugging or diagnostic messages, in place of the information presented from HistoricalLog we could instead place a message saying that HistoricalLog is disabled and describing the system property needed to enable it.

### Component(s)

Java

davisusanibar:
### Rationale for this change

To close apache/arrow#34338.

Removing the automatic enabling of BaseAllocator.DEBUG on -ea

### Are these changes tested?

Unit test added.

Yes.
* Closes: apache/arrow#34338
* GitHub Issue: #34338


java/memory/memory-core/src/main/java/org/apache/arrow/memory/BaseAllocator.java
```diff
@@ -53,9 +53,10 @@ abstract class BaseAllocator extends Accountant implements BufferAllocator {
     if (propValue != null) {
       DEBUG = Boolean.parseBoolean(propValue);
     } else {
-      DEBUG = AssertionUtil.isAssertionsEnabled();
+      DEBUG = false;
     }
-    logger.info("Debug mode " + (DEBUG ? "enabled." : "disabled."));
+    logger.info("Debug mode " + (DEBUG ? "enabled."
+        : "disabled. Enable that by the following VM option -Darrow.memory.debug.allocator=true."));
```
lidavidm, line 59:
```suggestion
        : "disabled. Enable with the VM option -Darrow.memory.debug.allocator=true."));
```

java/memory/memory-core/src/test/java/org/apache/arrow/memory/TestArrowBuf.java
```diff
@@ -138,12 +129,44 @@ public void testSetBytesBigEndian() {
...
       buf.setBytes(0, data);
       byte[] actual = new byte[expected.length];
       buf.getBytes(0, actual);
       assertArrayEquals(expected, actual);
     }
   }
 
+  @Test
+  public void testEnabledAssertion() {
+    ((Logger) LoggerFactory.getLogger("org.apache.arrow")).setLevel(Level.TRACE);
```
lidavidm, line 143: We need to reset the level after the test runs.

java/memory/memory-core/src/test/java/org/apache/arrow/memory/TestArrowBuf.java
```diff
@@ -138,12 +129,44 @@ public void testSetBytesBigEndian() {
...
+    ClassLoader.getSystemClassLoader().setDefaultAssertionStatus(true);
```
lidavidm, line 144: We probably want to reset this, too

java/memory/memory-core/src/test/java/org/apache/arrow/memory/TestArrowBuf.java
```diff
@@ -138,12 +129,44 @@ public void testSetBytesBigEndian() {
...
+      fieldDebug.setAccessible(true);
+      Field modifiersDebug = Field.class.getDeclaredField("modifiers");
+      modifiersDebug.setAccessible(true);
+      modifiersDebug.setInt(fieldDebug, fieldDebug.getModifiers() & ~Modifier.FINAL);
+      fieldDebug.set(null, true);
+      try (BufferAllocator allocator = new RootAllocator(128)) {
+        ArrowBuf buf = allocator.buffer(2);
+      } catch (Exception e) {
+        assertTrue(e.getMessage().contains("event log for:")); // JDK8, JDK11
+        fieldDebug.set(null, false);
```
lidavidm, line 166: put this in a `finally` block?

java/memory/memory-core/src/test/java/org/apache/arrow/memory/TestArrowBuf.java
lidavidm: Don't we need to reset it to its original value, instead of assuming what it was?

davisusanibar: Added

lidavidm: How does this test even work, actually? The property is already set by the time you get here.

davisusanibar: Ohhh thank you, you are right, assertions are enabled by default in all the unit test scope, then I remove the property and only validate the message.

java/memory/memory-core/src/test/java/org/apache/arrow/memory/TestArrowBuf.java
```diff
@@ -138,12 +129,48 @@ public void testSetBytesBigEndian() {
...
+        ArrowBuf buf = allocator.buffer(expected.length)) {
       buf.setBytes(0, data);
       byte[] actual = new byte[expected.length];
       buf.getBytes(0, actual);
       assertArrayEquals(expected, actual);
     }
   }
 
+  @Test
+  public void testEnabledAssertion() {
```
lidavidm, line 142:
```suggestion
  /**
   * Test that allocation history is not recorded even though
   * assertions are enabled in tests (GH-34338).
   */
  public void testEnabledAssertion() {
```

lidavidm:
@davisusanibar can you look into the flaky build? This has happened a lot.

java/memory/memory-core/src/test/java/org/apache/arrow/memory/TestArrowBuf.java
```diff
@@ -138,12 +129,52 @@ public void testSetBytesBigEndian() {
...
+  /**
+   * Test that allocation history is not recorded even though
+   * assertions are enabled in tests (GH-34338).
+   */
+  public void testEnabledAssertion() {
+    ((Logger) LoggerFactory.getLogger("org.apache.arrow")).setLevel(Level.TRACE);
+    try (BufferAllocator allocator = new RootAllocator(128)) {
+      ArrowBuf buf = allocator.buffer(2);
+    } catch (Exception e) {
+      assertFalse(e.getMessage().contains("event log for:"));
```
lidavidm, line 151: Can we use assertThrows

davisusanibar: Changed

java/memory/memory-core/src/test/java/org/apache/arrow/memory/TestArrowBuf.java
```diff
@@ -138,12 +129,52 @@ public void testSetBytesBigEndian() {
...
+    } finally {
+      ((Logger) LoggerFactory.getLogger("org.apache.arrow")).setLevel(null);
+    }
+  }
+
+  @Test
+  public void testEnabledHistoricalLog() {
```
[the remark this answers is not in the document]
davisusanibar: We could decouple our test that need to override system property but at the end the unique way to inject a parameter will be by reflection.

By definition there is not way to share system properties between main code and test code, for that reason what other framework do are using reflection to change that at runtime.

The initial scope of this test is JDK11, for another JDK17+ versions this test validate only the message 'java.lang.NoSuchFieldException: modifiers'

java/memory/memory-core/src/test/java/org/apache/arrow/memory/TestArrowBuf.java
```diff
@@ -165,14 +166,14 @@ public void testEnabledHistoricalLog() {
...
-      } catch (Exception e) {
+        allocator.buffer(2);
+        Exception e = assertThrows(IllegalStateException.class, () -> allocator.close());
         assertTrue(e.getMessage().contains("event log for:")); // JDK8, JDK11
       } finally {
         fieldDebug.set(null, false);
       }
     } catch (Exception e) {
-      assertTrue(e.toString().contains("java.lang.NoSuchFieldException: modifiers")); // JDK17+
+      assertTrue(e.toString().equals("java.lang.NoSuchFieldException: modifiers")); // JDK17+
```
lidavidm, line 176: Why was this changed to equals and not contains?

davisusanibar: To be more accurately

lidavidm: It's rather brittle

java/memory/memory-core/src/test/java/org/apache/arrow/memory/TestArrowBuf.java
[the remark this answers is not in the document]
lidavidm: So the test becomes worthless on JDK17+. I would rather we find a proper way to test this instead of a brittle test that will silently become invalid over time.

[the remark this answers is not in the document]
lidavidm: That doesn't have to block this PR, but I don't consider this test as a proper validation of the debug mode.

java/memory/memory-core/src/test/java/org/apache/arrow/memory/TestArrowBuf.java
davisusanibar: Ok, let me maintain contains instead.

davisusanibar:
Just filled a ticket for Cyclonedx errors: https://github.com/apache/arrow/issues/36371

java/memory/memory-core/src/test/java/org/apache/arrow/memory/TestArrowBuf.java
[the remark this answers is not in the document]
lidavidm: Can you file a ticket?

lidavidm (approved):

PR #36051, opened 2023-06-13
GH-36050: [Docs][C] Fix memory leak in C export documentation

### LINKED ISSUE #36050: [Docs] C producer examples leak memory?

Perhaps my limited C knowledge is making me miss something obvious, but I think the producer examples on [this page](https://arrow.apache.org/docs/format/CDataInterface.html#exporting-a-struct-float32-utf8-array) of the documentation leak memory? Child schemas are malloced:

```c
 //
 // Initialize child type #0
 //
 child = schema->children[0] = malloc(sizeof(struct ArrowSchema));
```

Then in the release implementation, we free the array of child pointers and call release on the children array, but don't actually free the child ArrowSchema allocations:

```c
static void release_malloced_type(struct ArrowSchema* schema) {
   int i;
   for (i = 0; i < schema->n_children; ++i) {
      struct ArrowSchema* child = schema->children[i];
      if (child->release != NULL) {
         child->release(child);
      }
   }
   free(schema->children);
   // Mark released
   schema->release = NULL;
}
```

Since (as far as I can tell?) the docs don't otherwise explicitly state what the lifetime of these children is, it could be nice to fix this? Reading the C++ ExportArray implementation, it seems like these are in fact freed in the release call via the private_data pointer in at least the C++ implementation.

### Component(s)

C, Documentation

lord:
### Rationale for this change
Fix memory leak in C export examples. Fixes #36050.

Change to example in docs

### Are these changes tested?
Docs only change.

No
* Closes: #36050

### DIFF
--- a/docs/source/format/CDataInterface.rst
+++ b/docs/source/format/CDataInterface.rst
@@ -783,6 +783,7 @@ Export the array type as a ``ArrowSchema`` with C-malloc()ed children:
          if (child->release != NULL) {
             child->release(child);
          }
+         free(child);
       }
       free(schema->children);
       // Mark released
@@ -857,6 +858,7 @@ transferring ownership to the consumer:
          if (child->release != NULL) {
             child->release(child);
          }
+         free(child);
       }
       free(array->children);
       // Free buffers


lord:
@zeroshade

kou (approved):
+1

Good catch!

PR #36054, opened 2023-06-13
WIP: [R] Verify CRAN release-12.0.1-rc1

thisisnic:


thisisnic:
@github-actions crossbow submit --group r

thisisnic:
valgrind timed out, devdocs often fails, conda isn't necessary for the release, and I have no idea why the homebrew build failed (it seems to be building from source instead of relying on the binary, so the binary not being there shouldn't be it), but have resubmitted the job to check.

thisisnic:
> The homebrew failure is the known failure when installing the latest protobuf version with brew: #35987

Thanks Raul.

assignUser:
12.0.1 is on CRAN :tada:

PR #36058, opened 2023-06-14
GH-36047: [C++][Compute] Add support for duration types to IndexIn and IsIn

### LINKED ISSUE #36047: ArrowNotImplementedError: Function 'is_in' has no kernel matching input types (duration[s])

```python
import numpy as np
import pyarrow as pa

dt = np.linspace(0, 100) * np.timedelta64(1, "s")
arr = pa.array(dt)  # duration[s]
pa.compute.is_in(arr, arr)  # ✘ ArrowNotImplementedError
```

Was kind of surprised by this, since presumably an `is_in` function should be pretty generic.

### Component(s)

Python (version 12.0.0)

js8544:
Add support for duration types to IndexIn and IsIn
* Closes: #36047

### DIFF
--- a/cpp/src/arrow/compute/kernels/scalar_set_lookup_test.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_set_lookup_test.cc
@@ -39,6 +39,7 @@
 #include "arrow/testing/gtest_compat.h"
 #include "arrow/testing/gtest_util.h"
 #include "arrow/type.h"
+#include "arrow/type_fwd.h"
 #include "arrow/type_traits.h"
 
 namespace arrow {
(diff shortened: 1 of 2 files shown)


westonpace (approved):
This looks like a good to me.  I tested it out a bit in python and it seems to work.  I do find the casting to be a bit arbitrary here.  For example, it seems odd that I can do `is_in(milliseconds, seconds)` but not `is_in(seconds, milliseconds)`.

However, users can always explicitly cast, and I think this is better for a follow-up if someone feels strongly about it.

js8544:
> This looks like a good to me. I tested it out a bit in python and it seems to work. I do find the casting to be a bit arbitrary here. For example, it seems odd that I can do `is_in(milliseconds, seconds)` but not `is_in(seconds, milliseconds)`.
> 
> However, users can always explicitly cast, and I think this is better for a follow-up if someone feels strongly about it.

Actually it's possible to do `is_in(seconds, milliseconds)` if the casting is safe. For example, `is_in([1s, 2s], [1000ms, 2000ms])` would return `[true, true]`, while `is_in([1s, 2s], [1ms, 2ms])` would throw an error.

Currently the kernels only try to cast `value_set` into query keys' type. It does make more sense if they cast both ways. I'll try to implement this in another PR.

cpp/src/arrow/compute/kernels/scalar_set_lookup_test.cc
```diff
@@ -238,6 +239,29 @@ TEST_F(TestIsInKernel, TimeTimestamp) {
...
+              "[true, false, false, true, true]", /*skip_nulls=*/true);
+  }
+
+  // Different units, invalid cast
+  ASSERT_RAISES(Invalid, IsIn(ArrayFromJSON(duration(TimeUnit::SECOND), "[0, 1, 2]"),
+                              ArrayFromJSON(duration(TimeUnit::MILLI), "[0, 2]")));
+
+  // Different units, valid cast
+  CheckIsIn(ArrayFromJSON(duration(TimeUnit::MILLI), "[0, 1, 2000]"),
+            ArrayFromJSON(duration(TimeUnit::SECOND), "[0, 2]"), "[true, false, true]");
```
pitrou, line 262: What happens if conversion from SECOND to MILLI overflows?

cpp/src/arrow/compute/kernels/scalar_set_lookup_test.cc
```diff
@@ -238,6 +239,29 @@ TEST_F(TestIsInKernel, TimeTimestamp) {
...
+              "[true, false, false, true, true]", /*skip_nulls=*/true);
+
+    // Duplicates in right array
+    CheckIsIn(type, "[1, null, 5, 1, 2]", "[2, 1, 1, null, 2]",
+              "[true, true, false, true, true]", /*skip_nulls=*/false);
+    CheckIsIn(type, "[1, null, 5, 1, 2]", "[2, 1, 1, null, 2]",
+              "[true, false, false, true, true]", /*skip_nulls=*/true);
+  }
+
+  // Different units, invalid cast
```
pitrou, line 256: Interesting... ideally, the values could be cast to MILLI instead of casting the value_set to SECOND. Don't have to solve this here, though. :-)

js8544: #36204 will solve this.

cpp/src/arrow/compute/kernels/scalar_set_lookup_test.cc
js8544: It calls Cast with CastOptions::Safe(), so it should return an error.

PR #36061, opened 2023-06-14
GH-34788: [Python][Packaging][CI] Drop Python 3.7 support

### LINKED ISSUE #34788: [Python][Packaging][CI] Drop Python 3.7 support
Previous issue for Python 3.6: https://github.com/apache/arrow/issues/28450

We dropped Python 3.6 support in pyarrow 7.0 (early 2022). If we follow the same pattern, we can drop 3.7 support for pyarrow 13 (so we can start doing it after the 12.0 release), since by the time of that release it should have reached end of life (https://endoflife.date/python)

raulcd:
### Rationale for this change

Python 3.7 is End of Support on 27th June 2023. We can drop support for pyarrow 13.0.0.
See: https://endoflife.date/python

Remove support for 3.7. Update minimum required Python version and update some CI jobs to use the minimum Python supported version

### Are these changes tested?

They will be tested via Archery

Yes in terms of not supporting older versions of Python. This will be a breaking change for some users but not for the API itself
* Closes: #34788


raulcd:

raulcd:
@github-actions crossbow submit -g python

.github/workflows/python.yml
```diff
@@ -47,41 +47,41 @@ jobs:
     name: ${{ matrix.title }}
     runs-on: ubuntu-latest
     if: ${{ !contains(github.event.pull_request.title, 'WIP') }}
-    timeout-minutes: 60
+    timeout-minutes: 90
```
raulcd, line 50: I've done this to fix the MacOS 12 job that was timing out after 60 minutes. Happy to revert if we think we should tackle the issue differently or on another PR.

raulcd:
The hypothesis, spark and hdfs failures are known nightly failures.

jorisvandenbossche (approved):
Looks good to me!

.github/workflows/python.yml
kou: It seems that we use bundled Protobuf instead of using Homebrew's Protobuf:
https://github.com/apache/arrow/actions/runs/5267649495/jobs/9523177390#step:5:236

```text
-- Could NOT find Protobuf (missing: Protobuf_LIBRARIES Protobuf_INCLUDE_DIR) (Required is at least version "3.0.0")
```

I should have confirmed this from CI log explicitly in #36029...

python/requirements-wheel-test.txt
```diff
@@ -7,8 +7,7 @@ pytest-lazy-fixture
 pytz
 tzdata; sys_platform == 'win32'
 
-numpy==1.19.5; platform_system == "Linux"   and platform_machine == "aarch64" and python_version <  "3.7"
-numpy==1.21.3; platform_system == "Linux"   and platform_machine == "aarch64" and python_version >= "3.7" and python_version < "3.11"
+numpy==1.21.3; platform_system == "Linux"   and platform_machine == "aarch64" and python_version >= "3.8" and python_version < "3.11"
```
kou, line 10:
```suggestion
numpy==1.21.3; platform_system == "Linux"   and platform_machine == "aarch64" and python_version < "3.11"
```

dev/tasks/tasks.yml
```diff
@@ -1517,9 +1516,9 @@ tasks:
       image: conda-python-hdfs
 {% endfor %}
 
-{% for python_version, spark_version, test_pyarrow_only, numpy_version in [("3.7", "v3.1.2", "false", "latest"),
-                                                                           ("3.8", "v3.2.0", "false", "1.23"),
-                                                                           ("3.9", "master", "false", "latest")] %}
+{% for python_version, spark_version, test_pyarrow_only, numpy_version in [("3.8", "v3.1.2", "false", "latest"),
```
kou, line 1519: Do we still need Spark 3.1.2?

I think that Spark 3.1 reached EOL:

https://spark.apache.org/versioning-policy.html

> Feature release branches will, generally, be maintained with bug fix releases for a period of 18 months.

And 3.1.1 was released on 2021-03-02:

https://spark.apache.org/news/index.html

> [Spark 3.1.1 released](https://spark.apache.org/news/spark-3-1-1-released.html)
>
> March 2, 2021

3.2 might reach EOL because 3.2.0 was released on 2021-10-13:

> [Spark 3.2.0 released](https://spark.apache.org/news/spark-3-2-0-released.html)
>
> October 13, 2021

@kiszk Do you know what Spark versions are maintained?

.github/workflows/python.yml
raulcd: Would you revert it back to 60 minutes and fix that on a different PR? I think that's probably better then.

dev/tasks/tasks.yml
raulcd: @kou should we move the spark versions conversation to update them to its own ticket?

kou: Ah, yes. It makes sense.

raulcd:
@github-actions crossbow submit -g python

raulcd:

dev/tasks/tasks.yml
raulcd: I have created this issue to track it: https://github.com/apache/arrow/issues/36199

raulcd:
I've rebased main and I am re-running tests to verify all is good before merging.

raulcd:
The failures on spark-maser, hypothesis and the arm64 wheels are already happening on our nightly builds.
@kou do you have any other concern or are you ok merging this?

kou (approved):
+1

PR #36067, opened 2023-06-14
GH-36059: [C++][Compute] Reserve space for hashtable for scalar lookup functions

### LINKED ISSUE #36059: [C++] Performance of building up HashTable (MemoTable) in is_in kernel
From some benchmarking in pandas, @phofl noticed that the "is_in" implementation of Arrow is considerably slower compared to pandas' (khash-based) implementation, _if_ the value_set is big (and thus when the execution time is mostly dominated by creating the hashtable, and not by the actual lookups). 

Small illustration comparing `pc.is_in` with `pandas.core.algorithms.isin` (the equivalent pandas function) using a larger value_set with a small array (so that we are mostly timing the hashtable creation):
```
arr = pa.array(np.random.choice(np.arange(1000), 100))
for n in [10_000, 100_000, 1_000_000, 10_000_000]:
    print(n)
    value_set = pa.array(np.arange(n))
    %timeit pc.is_in(arr, value_set)
    %timeit pd.core.algorithms.isin(np.asarray(arr), np.asarray(value_set))
```

gives

```
10000
158 µs ± 581 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
66.8 µs ± 473 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
100000
8.51 ms ± 110 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
627 µs ± 41.8 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
1000000
103 ms ± 8.74 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
28.5 ms ± 4.09 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
10000000
1.26 s ± 33 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
171 ms ± 22 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

So pandas is often 4-10x faster. I am not familiar with our HashTable implementation or whether it's a fully fair comparison, but it seems this is at least an indication there is room for performance improvement (and I suppose this might be for HashTable-based functions in general, not just "is_in").

js8544:
### Rationale for this change

Performance improvement for scalar lookup functions.

1. Reserve space for hashtable for scalar lookup functions before the insertion process.
2. A benchmark that demonstrates the improvement. Data is pasted in the comments.

### Are these changes tested?

By existing unit tests.

No.
* Closes: #36059

### DIFF
--- a/cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
@@ -36,7 +36,7 @@ namespace {
 
 template <typename Type>
 struct SetLookupState : public KernelState {
-  explicit SetLookupState(MemoryPool* pool) : lookup_table(), memory_pool(pool) {}
+  explicit SetLookupState(MemoryPool* pool) : memory_pool(pool) {}
 
   Status Init(const SetLookupOptions& options) {
     if (options.value_set.is_array()) {
@@ -105,7 +105,7 @@ struct SetLookupState : public KernelState {
   }
 
   using MemoTable = typename HashTraits<Type>::MemoTableType;
-  std::optional<MemoTable> lookup_table;
+  std::optional<MemoTable> lookup_table;  // use optional for delayed initialization
   MemoryPool* memory_pool;
   // When there are duplicates in value_set, the MemoTable indices must
   // be mapped back to indices in the value_set.
--- a/cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
@@ -36,16 +36,23 @@ namespace {
 
 template <typename Type>
 struct SetLookupState : public KernelState {
-  explicit SetLookupState(MemoryPool* pool) : lookup_table(pool, 0) {}
+  explicit SetLookupState(MemoryPool* pool) : lookup_table(pool, 0), memory_pool(pool) {}
 
   Status Init(const SetLookupOptions& options) {
     if (options.value_set.is_array()) {
       const ArrayData& value_set = *options.value_set.array();
       memo_index_to_value_index.reserve(value_set.length);
+      lookup_table =
+          MemoTable(memory_pool,
+                    ::arrow::internal::HashTable<char>::kLoadFactor * value_set.length);
       RETURN_NOT_OK(AddArrayValueSet(options, *options.value_set.array()));
     } else if (options.value_set.kind() == Datum::CHUNKED_ARRAY) {
       const ChunkedArray& value_set = *options.value_set.chunked_array();
       memo_index_to_value_index.reserve(value_set.length());
+      lookup_table =
+          MemoTable(memory_pool,
+                    ::arrow::internal::HashTable<char>::kLoadFactor * value_set.length());
+
       int64_t offset = 0;
       for (const std::shared_ptr<Array>& chunk : value_set.chunks()) {
         RETURN_NOT_OK(AddArrayValueSet(options, *chunk->data(), offset));
@@ -99,6 +106,7 @@ struct SetLookupState : public KernelState {
 
   using MemoTable = typename HashTraits<Type>::MemoTableType;
   MemoTable lookup_table;
+  MemoryPool* memory_pool;
   // When there are duplicates in value_set, the MemoTable indices must
   // be mapped back to indices in the value_set.
   std::vector<int32_t> memo_index_to_value_index;
--- a/cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
@@ -61,8 +61,8 @@ struct SetLookupState : public KernelState {
     } else {
       return Status::Invalid("value_set should be an array or chunked array");
     }
-    if (!options.skip_nulls && lookup_table.value().GetNull() >= 0) {
-      null_index = memo_index_to_value_index[lookup_table.value().GetNull()];
+    if (!options.skip_nulls && lookup_table->GetNull() >= 0) {
+      null_index = memo_index_to_value_index[lookup_table->GetNull()];
     }
     return Status::OK();
   }
@@ -82,7 +82,7 @@ struct SetLookupState : public KernelState {
         DCHECK_EQ(memo_index, memo_size);
         memo_index_to_value_index.push_back(index);
       };
-      RETURN_NOT_OK(lookup_table.value().GetOrInsert(
+      RETURN_NOT_OK(lookup_table->GetOrInsert(
           v, std::move(on_found), std::move(on_not_found), &unused_memo_index));
       ++index;
       return Status::OK();
@@ -96,7 +96,7 @@ struct SetLookupState : public KernelState {
         DCHECK_EQ(memo_index, memo_size);
         memo_index_to_value_index.push_back(index);
       };
-      lookup_table.value().GetOrInsertNull(std::move(on_found), std::move(on_not_found));
+      lookup_table->GetOrInsertNull(std::move(on_found), std::move(on_not_found));
       ++index;
       return Status::OK();
     };
@@ -272,7 +272,7 @@ struct IndexInVisitor {
     VisitArraySpanInline<Type>(
         data,
         [&](T v) {
-          int32_t index = state.lookup_table.value().Get(v);
+          int32_t index = state.lookup_table->Get(v);
           if (index != -1) {
             bitmap_writer.Set();
 
@@ -366,7 +366,7 @@ struct IsInVisitor {
     VisitArraySpanInline<Type>(
         this->data,
         [&](T v) {
-          if (state.lookup_table.value().Get(v) != -1) {
+          if (state.lookup_table->Get(v) != -1) {
             writer.Set();
           } else {
             writer.Clear();
--- a/cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
@@ -36,7 +36,7 @@ namespace {
 
 template <typename Type>
 struct SetLookupState : public KernelState {
-  explicit SetLookupState(MemoryPool* pool) : lookup_table(pool, 0), memory_pool(pool) {}
+  explicit SetLookupState(MemoryPool* pool) : lookup_table(), memory_pool(pool) {}
 
   Status Init(const SetLookupOptions& options) {
     if (options.value_set.is_array()) {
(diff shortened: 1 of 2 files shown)


js8544:
Benchmark result:
Before:
```
----------------------------------------------------------------------------------------
Benchmark                              Time             CPU   Iterations UserCounters...
----------------------------------------------------------------------------------------
IndexInInt32LargeSet/100            2883 ns         2883 ns       246728 bytes_per_second=13.2302M/s items_per_second=3.46823M/s
IndexInInt32LargeSet/1000          10322 ns        10322 ns        63804 bytes_per_second=3.6958M/s items_per_second=968.833k/s
IndexInInt32LargeSet/10000        139318 ns       139316 ns         4987 bytes_per_second=280.389k/s items_per_second=71.7795k/s
IndexInInt32LargeSet/100000      4588807 ns      4588452 ns          165 bytes_per_second=8.51322k/s items_per_second=2.17938k/s
IndexInInt32LargeSet/1000000    90538493 ns     90536500 ns           10 bytes_per_second=441.811/s items_per_second=110.453/s
IndexInInt32LargeSet/10000000 1428412088 ns   1428369286 ns            1 bytes_per_second=28.004/s items_per_second=7.00099/s
IsInInt32LargeSet/100               2833 ns         2833 ns       234016 bytes_per_second=13.4639M/s items_per_second=3.52948M/s
IsInInt32LargeSet/1000             10376 ns        10376 ns        67001 bytes_per_second=3.6765M/s items_per_second=963.773k/s
IsInInt32LargeSet/10000           148294 ns       148284 ns         4769 bytes_per_second=263.43k/s items_per_second=67.4381k/s
IsInInt32LargeSet/100000         4463505 ns      4463293 ns          147 bytes_per_second=8.75195k/s items_per_second=2.2405k/s
IsInInt32LargeSet/1000000       75740587 ns     75738687 ns           10 bytes_per_second=528.132/s items_per_second=132.033/s
IsInInt32LargeSet/10000000    1317214434 ns   1317158744 ns            1 bytes_per_second=30.3684/s items_per_second=7.5921/s
```

After
```
----------------------------------------------------------------------------------------
Benchmark                              Time             CPU   Iterations UserCounters...
----------------------------------------------------------------------------------------
IndexInInt32LargeSet/100            2659 ns         2658 ns       266800 bytes_per_second=14.3493M/s items_per_second=3.76159M/s
IndexInInt32LargeSet/1000           9759 ns         9759 ns        70169 bytes_per_second=3.9091M/s items_per_second=1024.75k/s
IndexInInt32LargeSet/10000         93311 ns        93309 ns         7477 bytes_per_second=418.634k/s items_per_second=107.17k/s
IndexInInt32LargeSet/100000      1729160 ns      1729162 ns          453 bytes_per_second=22.5904k/s items_per_second=5.78315k/s
IndexInInt32LargeSet/1000000    79806851 ns     79805041 ns            9 bytes_per_second=501.221/s items_per_second=125.305/s
IndexInInt32LargeSet/10000000 1112193163 ns   1112056579 ns            1 bytes_per_second=35.9694/s items_per_second=8.99235/s
IsInInt32LargeSet/100               2518 ns         2518 ns       280148 bytes_per_second=15.1526M/s items_per_second=3.97215M/s
IsInInt32LargeSet/1000              9625 ns         9624 ns        72600 bytes_per_second=3.96356M/s items_per_second=1039.02k/s
IsInInt32LargeSet/10000            93759 ns        93754 ns         7256 bytes_per_second=416.647k/s items_per_second=106.662k/s
IsInInt32LargeSet/100000         1608221 ns      1608152 ns          495 bytes_per_second=24.2903k/s items_per_second=6.21832k/s
IsInInt32LargeSet/1000000       73113133 ns     73112833 ns            9 bytes_per_second=547.1/s items_per_second=136.775/s
IsInInt32LargeSet/10000000    1207785785 ns   1207731078 ns            1 bytes_per_second=33.12/s items_per_second=8.27999/s
```

cpp/src/arrow/compute/kernels/scalar_set_lookup_benchmark.cc
```diff
@@ -91,35 +91,51 @@ static void IsInStringLargeSet(benchmark::State& state) {
...
 }
 
 static void IsInInt64SmallSet(benchmark::State& state) {
-  SetLookupBenchmarkNumeric<Int64Type>(state, "is_in_meta_binary", state.range(0));
+  SetLookupBenchmarkNumeric<Int64Type>(state, "is_in_meta_binary", state.range(0),
+                                       1 << 18);
+}
+
+static void IsInInt32LargeSet(benchmark::State& state) {
+  SetLookupBenchmarkNumeric<Int32Type>(state, "is_in_meta_binary", state.range(0), 10);
```
westonpace, line 138: Why `10` here instead of `1 << 18`?

westonpace (commented):
Just one question, otherwise this makes sense to me.

cpp/src/arrow/compute/kernels/scalar_set_lookup_benchmark.cc
js8544: If query length is `1 << 18`, the execution time would be dominated by the "lookup" phase and the optimization I implemented would have very little effect.  The idea is that the new `LargeSet` benchmarks are to measure the performance of the "preparation" phase, while the existing `SmallSet` benchmarks are to measure the "lookup" phase.

pitrou: Well, we should keep the benchmarks meaningful and relevant. A 10-length input array does not really represent real-world use cases.

We can also add this optimization without adding micro-benchmarks for it.

cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
```diff
@@ -36,16 +36,23 @@ namespace {
 
 template <typename Type>
 struct SetLookupState : public KernelState {
-  explicit SetLookupState(MemoryPool* pool) : lookup_table(pool, 0) {}
+  explicit SetLookupState(MemoryPool* pool) : lookup_table(pool, 0), memory_pool(pool) {}
```
pitrou, line 39: Ok, but this makes an initial memory allocation for the lookup table which is immediately undone in `Init`.

A possible way around that is to make `lookup_table` a `std::optional<MemoTable>`, to allow deferred initialization.

js8544: Good point. Changed it to `std::optional`

cpp/src/arrow/compute/kernels/scalar_set_lookup_benchmark.cc
js8544: Ok, how about 1000?

pitrou: That sounds fine to me.

js8544: Done

cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
```diff
@@ -54,8 +61,8 @@ struct SetLookupState : public KernelState {
     } else {
       return Status::Invalid("value_set should be an array or chunked array");
     }
-    if (!options.skip_nulls && lookup_table.GetNull() >= 0) {
-      null_index = memo_index_to_value_index[lookup_table.GetNull()];
+    if (!options.skip_nulls && lookup_table.value().GetNull() >= 0) {
+      null_index = memo_index_to_value_index[lookup_table.value().GetNull()];
```
pitrou, line 65: Can probably use the dereference operator?
```suggestion
    if (!options.skip_nulls && lookup_table->GetNull() >= 0) {
      null_index = memo_index_to_value_index[lookup_table->GetNull()];
```

cpp/src/arrow/compute/kernels/scalar_set_lookup_benchmark.cc
```diff
@@ -56,8 +56,8 @@ static void SetLookupBenchmarkString(benchmark::State& state,
 template <typename Type>
 static void SetLookupBenchmarkNumeric(benchmark::State& state,
                                       const std::string& func_name,
-                                      const int64_t value_set_length) {
-  const int64_t array_length = 1 << 18;
+                                      const int64_t value_set_length,
+                                      const int64_t array_length) {
```
pitrou, line 60: perhaps add `state.counters["value_set_length"] = value_set_length` somewhere below to make results more readable? (same in the other bench function)

cpp/src/arrow/compute/kernels/scalar_set_lookup_benchmark.cc
```diff
@@ -91,35 +91,52 @@ static void IsInStringLargeSet(benchmark::State& state) {
 }
 
 static void IndexInInt8SmallSet(benchmark::State& state) {
-  SetLookupBenchmarkNumeric<Int8Type>(state, "index_in_meta_binary", state.range(0));
+  SetLookupBenchmarkNumeric<Int8Type>(state, "index_in_meta_binary", state.range(0),
+                                      1 << 18);
```
pitrou, line 95: Hmm, can we avoid the repeated hardcoded value?
You could define a constant such as `constexpr int64 kArrayLengthWithSmallSet = 1 << 18`.

cpp/src/arrow/compute/kernels/scalar_set_lookup_benchmark.cc
```diff
@@ -134,10 +151,12 @@ BENCHMARK(IndexInInt8SmallSet)->RangeMultiplier(4)->Range(2, 8);
 BENCHMARK(IndexInInt16SmallSet)->RangeMultiplier(4)->Range(2, 64);
 BENCHMARK(IndexInInt32SmallSet)->RangeMultiplier(4)->Range(2, 64);
 BENCHMARK(IndexInInt64SmallSet)->RangeMultiplier(4)->Range(2, 64);
+BENCHMARK(IndexInInt32LargeSet)->RangeMultiplier(10)->Range(100, 10000000);
```
pitrou, line 154: Let's make the benchmarks shorter a bit? We don't need so many and so large value sets IMHO.
```suggestion
BENCHMARK(IndexInInt32LargeSet)->RangeMultiplier(100)->Range(1000, 1000000);
```

cpp/src/arrow/compute/kernels/scalar_set_lookup_benchmark.cc
js8544: done

cpp/src/arrow/compute/kernels/scalar_set_lookup_benchmark.cc
js8544: done

pitrou (approved):

PR #36073, opened 2023-06-14
GH-36036: [C++][Python][Parquet] Implement Float16 logical type

### LINKED ISSUE #36036: [C++][Parquet] Implement Float16 logical type

There's currently an active proposal to add a half-float logical type to the Parquet spec here: https://github.com/apache/parquet-format/pull/184. Following the discussion in the PR/issue, the general consensus was that we should go ahead and implement support for the type before moving forward with a vote.

To summarize, this would currently entail adding a float16 logical type (based on a 2-byte fixed-size binary physical type) and implementing read/write support. We also want to ensure that its ordering requirements are consistent with the native floating point types, min/max values are properly handled in `Statistics`, etc.

### Component(s)

C++, Parquet

benibus:
### Rationale for this change

There is currently an active proposal to support half-float types in Parquet. For more details/discussion, see the links in this PR's accompanying issue.

This PR implements basic support for a `Float16LogicalType` in accordance with the proposed spec. More specifically, this includes:

- Changes to `parquet.thrift` and regenerated `parqet_types` files
- Basic `LogicalType` class definition, method impls, and enums
- Support for specialized comparisons and column statistics

In the interest of scope, this PR does not currently deal with arrow integration and byte split encoding - although we will want both of these features resolved before the proposal is approved.

### Are these changes tested?

Yes (tests are included)

Yes

* Closes: #36036


benibus:
cc @anjakefala @zeroshade @pitrou

cpp/src/parquet/float_internal.h
```diff
@@ -0,0 +1,61 @@
...
+
+#include "arrow/util/bit_util.h"
+#include "arrow/util/ubsan.h"
+#include "parquet/types.h"
+
+namespace parquet {
+
+struct float16 {
+  constexpr static uint16_t min() { return 0b1111101111111111; }
+  constexpr static uint16_t max() { return 0b0111101111111111; }
```
mapleFU, line 31: should it follow the `min` `max` lowest` ?

cpp/src/parquet/float_internal.h
```diff
@@ -0,0 +1,61 @@
...
+#include <cstdint>
+#include <cstring>
+
+#include "arrow/util/bit_util.h"
+#include "arrow/util/ubsan.h"
+#include "parquet/types.h"
+
+namespace parquet {
+
+struct float16 {
```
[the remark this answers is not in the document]
pitrou: We might even want to define a `arrow::Float16` struct for easier handling of half-precision floats (together with a bunch of operators, and perhaps overrides for `std::isnan`, `std::numeric_limits` and friends)?

For example:
```c++
struct Float16 {
  uint16_t value;

  static Float16 FromBytes(const uint8_t*) { ... }
  static Float16 FromBytes(std::array<uint8_t, 2>) { ... }
  std::array<uint8_t, 2> ToBytes() const { ... }

  friend bool operator== ...
};

static_assert(std::is_trivial_v<Float16>);
```

cpp/src/parquet/statistics_test.cc
```diff
@@ -1073,50 +1125,217 @@ void CheckExtrema() {
...
+template <>
+void TestFloatStatistics<Float16LogicalType>::TestNaNs() {
+  constexpr int kNumValues = 8;
+  constexpr int kValueLen = sizeof(uint16_t);
+
+  NodePtr node = this->MakeNode("f", Repetition::OPTIONAL);
+  ColumnDescriptor descr(node, 1, 1);
+
+  const uint16_t nan_int = 0b1111110010101010;
+  const uint16_t min_int = 0b1010010111000110;
```
pitrou, line 1291: Can you comment on FP values for non-NaN constants?

pitrou (commented):
Thanks @benibus ! Please note that the Thrift definitions will have to be synchronized from https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift once the Parquet format spec addition is accepted.

cpp/src/parquet/float_internal.h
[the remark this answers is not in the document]
zeroshade: I was going to make this same suggestion, so +1 from me on putting these helpers into `arrow/util`

cpp/src/parquet/statistics_test.cc
```diff
@@ -972,12 +1016,20 @@ TEST_F(TestStatisticsSortOrderFLBA, UnknownSortOrder) {
...
+}
+
 template <typename Stats, typename Array, typename T = typename Array::value_type>
 void AssertMinMaxAre(Stats stats, const Array& values, T expected_min, T expected_max) {
   stats->Update(values.data(), values.size(), 0);
   ASSERT_TRUE(stats->HasMinMax());
-  EXPECT_EQ(stats->min(), expected_min);
-  EXPECT_EQ(stats->max(), expected_max);
+  EXPECT_EQ(stats->EncodeMin(), EncodeValue(expected_min));
+  EXPECT_EQ(stats->EncodeMax(), EncodeValue(expected_max));
```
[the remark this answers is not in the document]
benibus: Yeah basically. The testing header does define one but it expects a predetermined length (which is not 2).

benibus:
Thanks! Before moving forward on some of the suggestions, I just want to ensure that I interpreted the endianness requirement from the proposal correctly. As per [the spec](https://github.com/apache/parquet-format/pull/184): 

> The `FLOAT16` annotation represents half-precision floating-point numbers in the 2-byte IEEE little-endian format.

Is this in line with what I've done here (i.e. the `FLBA` itself representing a little-endian binary float) or does it imply something else?

pitrou:
> Is this in line with what I've done here (i.e. the `FLBA` itself representing a little-endian binary float)

Yes!

cpp/src/parquet/statistics.cc
```diff
@@ -277,11 +278,54 @@ template <bool is_signed>
 struct CompareHelper<FLBAType, is_signed>
     : public BinaryLikeCompareHelperBase<FLBAType, is_signed> {};
 
+struct Float16CompareHelper {
+  using T = FLBA;
+
+  static T DefaultMin() { return T{float16::max_ptr()}; }
+  static T DefaultMax() { return T{float16::min_ptr()}; }
```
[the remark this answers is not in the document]
benibus: They're actually meant to be swapped like that (the impl basically treats them as null-like values). For instance, this is the equivalent for native floats: https://github.com/apache/arrow/blob/5e993f9ca402199a613d7abaea163c5cd9b3dccb/cpp/src/parquet/statistics.cc#L84-L85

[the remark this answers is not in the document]
wgtmac: Thanks for the explanation!

cpp/src/arrow/util/float16.h
```diff
@@ -0,0 +1,179 @@
...
+  }
+  friend constexpr bool operator!=(Float16Base lhs, Float16Base rhs) {
+    return !(lhs == rhs);
+  }
+
+  friend constexpr bool operator<(Float16Base lhs, Float16Base rhs) {
+    if (lhs.is_nan() || rhs.is_nan()) return false;
+    return Float16Base::CompareLt(lhs, rhs);
+  }
+  friend constexpr bool operator>(Float16Base lhs, Float16Base rhs) { return rhs < lhs; }
```
pitrou, line 94: Nit, but it's a bit weird to have this one delegate to its counterpart, while `operator>=` doesn't.

cpp/src/arrow/util/float16.h
```diff
@@ -0,0 +1,179 @@
...
+  friend constexpr bool operator<=(Float16Base lhs, Float16Base rhs) {
+    if (lhs.is_nan() || rhs.is_nan()) return false;
+    return !Float16Base::CompareLt(rhs, lhs);
+  }
+  friend constexpr bool operator>=(Float16Base lhs, Float16Base rhs) {
+    if (lhs.is_nan() || rhs.is_nan()) return false;
+    return !Float16Base::CompareLt(lhs, rhs);
+  }
+
+  friend std::ostream& operator<<(std::ostream& os, Float16Base arg) {
```
pitrou, line 105: This means we should include `<iosfwd>` (and by not defining the method in this header, we avoid including the entire iostream library).

cpp/src/arrow/util/float16.h
```diff
@@ -0,0 +1,179 @@
...
+  /// \brief Copy the value's bytes in native-endian byte order
+  void ToBytes(uint8_t* dest) const { std::memcpy(dest, &value_, sizeof(value_)); }
+  /// \brief Return the value's bytes in native-endian byte order
+  std::array<uint8_t, 2> ToBytes() const {
+    std::array<uint8_t, 2> bytes;
+    ToBytes(bytes.data());
+    return bytes;
+  }
+
+  void ToLittleEndian(uint8_t* dest) const {
```
pitrou, line 68: It would be nice to add docstrings for all public methods.

cpp/src/arrow/util/float16.h
```diff
@@ -0,0 +1,179 @@
...
+        return (lhs.bits() & 0x7fff) > (rhs.bits() & 0x7fff);
+      } else {
+        // Handle +/-0
+        return !lhs.is_zero() || rhs.bits() != 0;
+      }
+    } else if (rhs.signbit()) {
+      return false;
+    } else {
+      // Both are positive
+      return (lhs.bits() & 0x7fff) < (rhs.bits() & 0x7fff);
```
pitrou, line 130: If they're both positive, then why do we need to AND the bits?

cpp/src/arrow/util/float16.h
```diff
@@ -0,0 +1,179 @@
...
+ private:
+  // Comparison helpers that assume neither operand is NaN
+  static constexpr bool CompareEq(Float16Base lhs, Float16Base rhs) {
+    return (lhs.bits() == rhs.bits()) || (lhs.is_zero() && rhs.is_zero());
+  }
+  static constexpr bool CompareLt(Float16Base lhs, Float16Base rhs) {
+    if (lhs.signbit()) {
+      if (rhs.signbit()) {
+        // Both are negative
+        return (lhs.bits() & 0x7fff) > (rhs.bits() & 0x7fff);
```
pitrou, line 121: I don't think ANDing changes anything here?

cpp/src/arrow/util/float16_test.cc
```diff
@@ -0,0 +1,160 @@
...
+};
+
+template <typename Operator>
+class Float16OperatorTest : public ::testing::Test {
+ public:
+  void TestCompare(const std::vector<TestValue>& test_values) {
+    const auto num_values = static_cast<int>(test_values.size());
+
+    // Check all combinations of operands in both directions
+    for (int offset = 0; offset < num_values; ++offset) {
```
pitrou, line 101: This is a strange way to write the double loop. Why not simply:
```c++
for (int i = 0; i < num_values; ++i) {
  for (int j = 0; i < num_values; ++j) {
```

cpp/src/parquet/statistics.cc
```diff
@@ -53,6 +55,25 @@ namespace {
 constexpr int value_length(int value_length, const ByteArray& value) { return value.len; }
 constexpr int value_length(int type_length, const FLBA& value) { return type_length; }
 
+// Static "constants" for normalizing float16 min/max values. These need to be expressed
+// as pointers because `Float16LogicalType` represents an FLBA.
+const uint8_t* float16_lowest() {
+  static const auto bytes = std::numeric_limits<Float16>::lowest().ToLittleEndian();
```
pitrou, line 61: This could probably made `constexpr` by spelling out the `#if ARROW_LITTLE_ENDIAN` dance explicitly in `ToLittleEndian` and friends.
Not sure that's useful here, but worth remembering.

cpp/src/parquet/statistics.cc
```diff
@@ -277,11 +298,42 @@ template <bool is_signed>
...
 
+struct Float16CompareHelper {
+  using T = FLBA;
+
+  static T DefaultMin() { return T{float16_max()}; }
+  static T DefaultMax() { return T{float16_lowest()}; }
+
+  static T Coalesce(T val, T fallback) {
+    return val.ptr != nullptr && Float16::FromLittleEndian(val.ptr).is_nan() ? fallback
+                                                                             : val;
```
pitrou, line 309: Shouldn't this be, rather (also adding parentheses for clarity):
```suggestion
    return (val.ptr == nullptr || Float16::FromLittleEndian(val.ptr).is_nan()) ? fallback
                                                                               : val;
```

cpp/src/parquet/statistics.cc
```diff
@@ -458,6 +540,16 @@ std::pair<ByteArray, ByteArray> TypedComparatorImpl<false, ByteArrayType>::GetMi
   return GetMinMaxBinaryHelper<false>(*this, values);
 }
 
+static LogicalType::Type::type LogicalTypeId(const ColumnDescriptor* descr) {
```
pitrou, line 543: Not sure the `static` is required (aren't we already in the anonymouns namespace?).

cpp/src/parquet/statistics.cc
```diff
@@ -525,6 +616,19 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
   bool Equals(const Statistics& raw_other) const override {
     if (physical_type() != raw_other.physical_type()) return false;
 
+    const auto logical_id = LogicalTypeId(*this);
+    switch (logical_id) {
+      // Only compare against logical types that influence the interpretation of the
+      // physical type
+      case LogicalType::Type::FLOAT16:
+        if (LogicalTypeId(raw_other) != logical_id) {
+          return false;
```
pitrou, line 625: What if `this` is not a float16 but `raw_other` is?

pitrou:
By the way:

> In the interest of scope, this PR does not currently deal with arrow integration and byte split encoding - although we will want both of these features resolved before the proposal is approved.

Do you plan to tackle Arrow integration in a subsequent PR? This PR cannot be integrated before the proposal is approved, AFAICT.

cpp/src/arrow/util/float16.h
```diff
@@ -0,0 +1,179 @@
...
+  std::array<uint8_t, 2> ToBigEndian() const {
+    return Float16Base{bit_util::ToBigEndian(value_)}.ToBytes();
+  }
+
+  friend constexpr bool operator==(Float16Base lhs, Float16Base rhs) {
+    if (lhs.is_nan() || rhs.is_nan()) return false;
+    return Float16Base::CompareEq(lhs, rhs);
+  }
+  friend constexpr bool operator!=(Float16Base lhs, Float16Base rhs) {
+    return !(lhs == rhs);
```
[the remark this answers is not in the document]
benibus: Are you sure? Just double-checked and native float equivalents return true for me (i.e. `1.0f != std::numeric_limits<float>::quiet_NaN()`, `std::numeric_limits<float>::quiet_NaN() != std::numeric_limits<float>::quiet_NaN()`)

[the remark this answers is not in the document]
pitrou: Uh. It appears you're right. I must have been misremembering the (wacky) rules...

benibus:
> Do you plan to tackle Arrow integration in a subsequent PR? This PR cannot be integrated before the proposal is approved, AFAICT.

Yes, my plan _was_ to handle Arrow integration in a subsequent PR (in tandem with the Java implementation), and then move forward with the proposal.

However, I think you're right that we'll need to roll those features into this PR. Just did a closer re-reading of @julienledem's comments on https://github.com/apache/parquet-format/pull/184, and I believe we're going to want the full C++/Java PRs ready (but **not** merged) before approving the proposal - so we'll probably need to sit on this one in the meantime.

That being said, I can maintain this PR until then. Plus we may need to make adjustments if new requirements come in from the Java side.

cpp/src/arrow/util/float16.h
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  void ToLittleEndian(uint8_t* dest) const {
+    Float16Base{bit_util::ToLittleEndian(value_)}.ToBytes(dest);
+  }
+  std::array<uint8_t, 2> ToLittleEndian() const {
+    return Float16Base{bit_util::ToLittleEndian(value_)}.ToBytes();
+  }
+
+  void ToBigEndian(uint8_t* dest) const {
```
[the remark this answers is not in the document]
benibus: Not sure how generally useful it would be either. I only really added it for symmetry given that the little-endian requirement is Parquet-specific.

pitrou:
Well, if Arrow integration does not come in this PR, then at least some testing of reading/writing using non-Arrow Parquet APIs should still be added.

pitrou:
We probably want a Java implementation as well according to https://github.com/apache/parquet-format/pull/184#issuecomment-1428363226

wgtmac:
> We probably want a Java implementation as well according to [apache/parquet-format#184 (comment)](https://github.com/apache/parquet-format/pull/184#issuecomment-1428363226)

OK, if that is a must. I can spare some time to do this. Would be good after this PR gets completed.

benibus:
@wgtmac If you're willing to do that then it'd be greatly appreciated. I'll try to help where I can (or at least get some more eyes on it).

> Well, if Arrow integration does not come in this PR, then at least some testing of reading/writing using non-Arrow Parquet APIs should still be added.

I'm planning to add Arrow support to this PR in the coming days, but I can add non-Arrow read/write tests if we need more coverage.

benibus:
@wgtmac Are you still interested in working on the Java side of this?

benibus:
No worries! I'll ping some others and keep you updated if necessary.

zeroshade:
@benibus Should the float16 type's usage in the Parquet code be put behind a Parquet version check? i.e. should it error out if reading/writing a Parquet v1.0 file as opposed to a v2.6 or whatever?

anjakefala:
I just pushed a commit that adds PyArrow test coverage for float16 Parquet. We can add [Python] to the PR name.

cpp/src/arrow/util/float16.h
```diff
@@ -0,0 +1,198 @@
...
+namespace util {
+
+/// \brief Base class for an IEEE half-precision float, encoded as a `uint16_t`
+///
+/// The exact format is as follows (from MSB to LSB):
+/// - bit 0:     sign
+/// - bits 1-5:  exponent
+/// - bits 6-15: mantissa
+///
+/// NOTE: Methods in the class should not mutate the unerlying value or produce copies.
```
bkietz, line 41:
```suggestion
/// NOTE: Methods in the class should not mutate the underlying value or produce copies.
```

bkietz: I'm not sure why this division is beneficial. Is this implementation inspired by another which makes this distinction? If possible, I think it'd be preferable to be simpler and have a single class.

If you're looking at BasicDecimal/Decimal, those classes are kept separate so that Gandiva can include them in generated LLVM IR (so methods which reference std:: types and functions are kept out). Similar support for float16 is out of scope here, I would think.

cpp/src/arrow/util/float16.cc
```diff
@@ -0,0 +1,28 @@
...
+// under the License.
+
+#include <ostream>
+
+#include "arrow/util/float16.h"
+
+namespace arrow {
+namespace util {
+
+std::ostream& operator<<(std::ostream& os, Float16Base arg) { return (os << arg.bits()); }
```
bkietz, line 25: I think a missing piece from this PR is conversion to/from `float`. With that defined, we could print the equivalent `float` value instead and test that each `float16_test_cc::TestValue` is internally consistent

cpp/src/arrow/util/float16_test.cc
```diff
@@ -0,0 +1,168 @@
...
+  }
+};
+
+using OperatorTypes =
+    ::testing::Types<CompareEq, CompareNe, CompareLt, CompareGt, CompareLe, CompareGe>;
+
+TYPED_TEST_SUITE(Float16OperatorTest, OperatorTypes);
+
+TYPED_TEST(Float16OperatorTest, Compare) { this->TestCompare(g_test_values); }
+
```
bkietz, line 122: Instead of going through TYPED_TEST etc, I think it'd be more readable to use:

```c++
TEST(Float16Test, Compare) {
  auto ExpectOp = [](std::string op_name, auto op) {
    ARROW_SCOPED_TRACE(op_name);
    const auto num_values = static_cast<int>(g_test_values.size());

    // Check all combinations of operands in both directions
    for (int i = 0; i < num_values; ++i) {
      for (int j = 0; j < num_values; ++j) {
        ARROW_SCOPED_TRACE(i, ",", j);

        auto [a16, a32] = test_values[i];
        auto [b16, b32] = test_values[j];

        // Results for float16 and float32 should be the same
        ASSERT_EQ(op(a16, b16), op(a32, b32));
      }
    }
  };

  ExpectOp("equal", [](auto l, auto r) { return l == r; });
  // ...
}
```

cpp/src/parquet/arrow/reader_internal.cc
```diff
@@ -713,6 +715,77 @@ Status TransferDecimal(RecordReader* reader, MemoryPool* pool,
...
+/// We do this by:
+/// 1. Creating an arrow::BinaryArray from the RecordReader's builder
+/// 2. Allocating a buffer for the arrow::HalfFloatArray
+/// 3. Converting the little-endian bytes in each BinaryArray entry to native-endian
+/// halffloat (uint16_t) values
+Status TransferHalfFloat(RecordReader* reader, MemoryPool* pool,
+                         const std::shared_ptr<Field>& field, Datum* out) {
+  auto binary_reader = dynamic_cast<BinaryRecordReader*>(reader);
+  DCHECK(binary_reader);
+  ::arrow::ArrayVector chunks = binary_reader->GetBuilderChunks();
```
bkietz, line 775: Maybe I'm missing something, but I think this should be specifically a fixed size binary array:
```suggestion
/// 1. Creating an arrow::FixedSizeBinaryArray from the RecordReader's builder
/// 2. Allocating a buffer for the arrow::HalfFloatArray
/// 3. Converting the little-endian bytes in each FixedSizeBinaryArray entry to native-endian
/// halffloat (uint16_t) values
Status TransferHalfFloat(RecordReader* reader, MemoryPool* pool,
                         const std::shared_ptr<Field>& field, Datum* out) {
  auto* binary_reader = dynamic_cast<FLBARecordReader*>(reader);
  DCHECK(binary_reader);
  ::arrow::ArrayVector chunks = binary_reader->GetBuilderChunks();
```

cpp/src/parquet/arrow/reader_internal.cc
```diff
@@ -713,6 +715,77 @@ Status TransferDecimal(RecordReader* reader, MemoryPool* pool,
   return Status::OK();
 }
 
+static inline Status ConvertToHalfFloat(const Array& array,
```
bkietz, line 718: Could this return Result<std::shared_ptr<Array>> instead?

bkietz (changes requested):

cpp/src/parquet/statistics.cc
```diff
@@ -527,9 +616,28 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
 
   void IncrementNumValues(int64_t n) override { num_values_ += n; }
 
+  static bool IsMeaningfulLogicalType(LogicalType::Type::type type) {
```
mapleFU, line 619: Emm... May I ask why only F16 is the meaningFul logical type?

cpp/src/arrow/util/float16.h
benibus: Honestly, the only reason for the base class was to make testing on the Parquet side more convenient, since the value data needs to be stored externally (see `BufferedFloat16` [here](https://github.com/apache/arrow/blob/2f06b0b0577f53547da9bc471d69c49749ae8e89/cpp/src/parquet/statistics_test.cc#L66)). I'm not sure if this is a good enough justification for a public API though - and I agree that a single class would be better, so I'll try a different approach for the tests.

cpp/src/arrow/util/float16.cc
benibus: Yep... the lack of conversion functions is the big elephant in the room. I initially avoided going there since I didn't expect they'd be needed for this PR, but their utility has become kinda obvious at this point.

cpp/src/parquet/statistics.cc
benibus: This just happens to be first logical type that dramatically affects how the physical value (a `FLBA` in this case) should be interpreted for comparisons. But I tried to generalize things in case any similar logical types are added in the future.

wgtmac:
@benibus There's an update on my side. @zhangjiashen is interested in working on this in parquet-mr. I will help him make some progress.

benibus:
@wgtmac Thanks for the update! I should mention that I actually made some progress on the Java side myself - although I didn't get very deep into the more substantial stuff (statistics, comparators). That being said, I'd be happy to let @zhangjiashen take the wheel since it'd allow me to focus on the C++/Go side (I dont _really_ know Java) - plus, I'll be away next week. If you want, I can provide my work in some form as a starting point.

I'm happy to help where I can, of course - particularly with the floating point logic. Right now I'm working on conversions between short/half-precision floats for this PR, so if the need arises for similar functionality in parquet-mr then stay tuned.

cpp/src/parquet/statistics_test.cc
```diff
@@ -1082,6 +1115,45 @@ void TestStatisticsSortOrder<FLBAType>::SetValues() {
...
+  values_buf_.resize(kNumBytes);
+  uint8_t* ptr = values_buf_.data();
+  for (int i = 0; i < NUM_VALUES; ++i) {
+    Float16(u16_vals[i]).ToLittleEndian(ptr);
+    values_[i].ptr = ptr;
+    ptr += kValueLen;
+  }
+
+  stats_[0]
+      .set_min(std::string(reinterpret_cast<const char*>(values_[0].ptr), kValueLen))
```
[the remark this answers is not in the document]
benibus: That's correct.

wgtmac:
> If you want, I can provide my work in some form as a starting point.

Yes, it would be good to continue the work based on your local change on the Java side.

> Right now I'm working on conversions between short/half-precision floats for this PR, so if the need arises for similar functionality in parquet-mr then stay tuned.

IMO, the float16 support in the parquet-mr should be simpler than here. At least it does not require to provide a full-featured float16 data type implementation. CMIW, the goal of parquet-mr is to support read/write float16 values (and stats/page index) in the form of `byte[2]`. It is the engine's responsibility to handle conversion from/to any customized float16 type.

cpp/src/parquet/statistics_test.cc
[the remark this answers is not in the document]
wgtmac: Thanks for confirmation!

benibus:
@wgtmac Hopefully this branch can be of some use: https://github.com/benibus/parquet-mr/tree/PARQUET-1647-float16. It should cover most of the initial boilerplate, at least.

For generating the thrift definitions, I installed the [pending parquet-format branch](https://github.com/apache/parquet-format/pull/184) to my local Maven repo and edited the pom.xml to target the SNAPSHOT version. There may be a better way of doing that kind of thing, of course...

wgtmac:
> For generating the thrift definitions, I installed the [pending parquet-format branch](https://github.com/apache/parquet-format/pull/184) to my local Maven repo and edited the pom.xml to target the SNAPSHOT version. There may be a better way of doing that kind of thing, of course...

Thank you very much! I did the same thing to install the unmerged thrift PR.

cpp/src/parquet/statistics.cc
mapleFU: Thanks for you explanation!

cpp/src/parquet/arrow/reader_internal.cc
```diff
@@ -713,6 +715,77 @@ Status TransferDecimal(RecordReader* reader, MemoryPool* pool,
...
+  const int64_t length = binary_array.length();
+  // Allocate data for the output halffloat array
+  ARROW_ASSIGN_OR_RAISE(auto data, ::arrow::AllocateBuffer(length * byte_width, pool));
+  uint8_t* out_ptr = data->mutable_data();
+
+  const int64_t null_count = binary_array.null_count();
+  // Copy the values to the output array in native-endian format
+  if (null_count > 0) {
+    for (int64_t i = 0; i < length; ++i, out_ptr += byte_width) {
+      Float16 f16{0};
```
mapleFU, line 741: The code looks good to me, but why can't we blink memcpy when it's `LITTLE_ENDIAN`? Since if it's null, the value is undefined?

cpp/src/parquet/arrow/reader_internal.cc
mapleFU: `BinaryRecordReader` can be regard as a base class, so it would work here. however I think it would be more clear with @bkietz 's advice

cpp/src/parquet/column_writer.cc
```diff
@@ -2305,6 +2307,74 @@ struct SerializeFunctor<
...
+      }
+    }
+    return Status::OK();
+  }
+
+ private:
+  FLBA ToFLBA(const uint16_t* value_ptr) const {
+    return FLBA{reinterpret_cast<const uint8_t*>(value_ptr)};
+  }
+  FLBA ToFLBA(uint16_t value) {
```
mapleFU, line 2361: The code look good to me, but the two mode here is a bit confusing, would different name for these `ToFLBA` better? (avoid mis-use the `ToFLBA` and touch `scratch_` )

mapleFU (commented):
Rest LGTM, thanks ben!

cpp/src/parquet/statistics_test.cc
```diff
@@ -1082,6 +1109,45 @@ void TestStatisticsSortOrder<FLBAType>::SetValues() {
...
+      0b1100010100000000,  // -5.0
+      0b1100010000000000,  // -4.0
+      0b1100001000000000,  // -3.0
+      0b1100000000000000,  // -2.0
+      0b1011110000000000,  // -1.0
+      0b0000000000000000,  // +0.0
+      0b0011110000000000,  // +1.0
+      0b0100000000000000,  // +2.0
+      0b0100001000000000,  // +3.0
+      0b0100010000000000,  // +4.0
```
pitrou, line 1135: Can you reorder the values so that they are not pre-sorted? This would exercise the minmax routines a bit better.

cpp/src/parquet/statistics.cc
```diff
@@ -527,9 +616,28 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
...
+        return true;
+      default:
+        return false;
+    }
+  }
+
   bool Equals(const Statistics& raw_other) const override {
     if (physical_type() != raw_other.physical_type()) return false;
 
+    const auto logical_id = LogicalTypeId(*this);
```
pitrou, line 631: Perhaps store the logical type id in the `TypedStatisticsImpl` constructor, instead of calling `LogicalTypeId` from multiple places?

cpp/src/parquet/column_writer.cc
```diff
@@ -2305,6 +2307,74 @@ struct SerializeFunctor<
...
+// Write Arrow to Float16
+
+// Requires a custom serializer because Float16s in Parquet are stored as a 2-byte
+// (little-endian) FLBA, whereas in Arrow they're a native `uint16_t`. Also, a temporary
+// buffer is needed if there's an endian mismatch.
+template <>
+struct SerializeFunctor<::parquet::FLBAType, ::arrow::HalfFloatType> {
+  Status Serialize(const ::arrow::HalfFloatArray& array, ArrowWriteContext* ctx,
+                   FLBA* out) {
+#if ARROW_LITTLE_ENDIAN
```
pitrou, line 2320: None of the Parquet C++ implementation is big-endian compatible, so I'm not sure you should bother specifically here. We also don't have any CI for this.

cpp/src/parquet/arrow/arrow_schema_test.cc
```diff
@@ -906,22 +908,6 @@ TEST_F(TestConvertArrowSchema, ArrowFields) {
...
 }
 
-TEST_F(TestConvertArrowSchema, ArrowNonconvertibleFields) {
-  struct FieldConstructionArguments {
-    std::string name;
-    std::shared_ptr<::arrow::DataType> datatype;
-  };
-
-  std::vector<FieldConstructionArguments> cases = {
-      {"float16", ::arrow::float16()},
```
pitrou, line 916: Instead of removing this test, can we use another Arrow datatype whose conversion to Parquet isn't supported?

cpp/src/parquet/arrow/reader_internal.cc
pitrou: Also, Parquet C++ has no big-endian compatibility currently.

pitrou: memcpy isn't even needed. You can (or should be able to) just reuse the buffers. For example by calling `array.View(type)`.

cpp/src/parquet/arrow/arrow_schema_test.cc
```diff
@@ -851,6 +851,8 @@ TEST_F(TestConvertArrowSchema, ArrowFields) {
        ParquetType::FIXED_LEN_BYTE_ARRAY, 7},
       {"decimal(32, 8)", ::arrow::decimal(32, 8), LogicalType::Decimal(32, 8),
        ParquetType::FIXED_LEN_BYTE_ARRAY, 14},
+      {"float16", ::arrow::float16(), LogicalType::Float16(),
+       ParquetType::FIXED_LEN_BYTE_ARRAY, 2},
```
pitrou, line 855: This will test Arrow to Parquet, can also you add a test for the other way round in `TestConvertParquetSchema`?

cpp/src/parquet/arrow/test_util.h
```diff
@@ -65,12 +65,44 @@ struct Decimal256WithPrecisionAndScale {
   static constexpr int32_t scale = PRECISION - 1;
 };
 
+inline std::vector<uint16_t> RandomHalfFloatValues(size_t size, uint16_t min,
```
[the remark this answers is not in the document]
pitrou: And perhaps `random_real` can be enhanced to accept a `Float16` type parameter?

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -916,9 +925,15 @@ TYPED_TEST(TestParquetIO, SingleColumnOptionalReadWrite) {
...
 TYPED_TEST(TestParquetIO, SingleColumnOptionalDictionaryWrite) {
-  // Skip tests for BOOL as we don't create dictionaries for it.
-  if (TypeParam::type_id == ::arrow::Type::BOOL) {
-    return;
+  switch (TypeParam::type_id) {
+    // Skip tests for BOOL as we don't create dictionaries for it.
+    case ::arrow::Type::BOOL:
+    // Skip tests for HALF_FLOAT as it's not currently supported by `dictionary_encode`
+    case ::arrow::Type::HALF_FLOAT:
+      GTEST_SKIP();
```
pitrou, line 933: Let's add a skip message for clarity (though I'm not sure GTest always displays it).
```suggestion
      GTEST_SKIP() << "dictionary_encode not supported for float16" ;
```

pitrou: Also do the same for bool above.

pitrou (commented):
Thanks a lot for this.
In addition to what others reviewers posted, here are a bunch of comments.

Also, you'll need to update the doc for supported types here:
https://arrow.apache.org/docs/cpp/parquet.html#logical-types

cpp/src/arrow/util/float16.h
```diff
@@ -0,0 +1,192 @@
...
+    }
+  }
+};
+
+static_assert(std::is_trivial_v<Float16>);
+
+}  // namespace util
+}  // namespace arrow
+
+// TODO: Not complete
```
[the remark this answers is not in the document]
bkietz: Dropping overloads and specializations into std is UB except where specifically allowed, and I don't see that noted for [isnan](https://eel.is/c++draft/cmath.syn). We could get around that a little by having `isnan.h`:

```c++
#include <cmath>
#include "arrow/util/float16.h"
namespace arrow::util {
using std::isnan;
constexpr bool isnan(Float16 f) { return f.isnan(); }
}
```

... I'm not sure if that's worthwhile though; we'd only get any benefit out of that when we need to call isnan and we happen to be in a sufficiently generic context that we don't wish to have explicitly different code for `float` and `Float16`. I'd guess it's not worth the trouble

anjakefala:
@wgtmac @zhangjiashen for the parquet-mr implementation, could you let me know when you have a branch/PR available on GitHub? =) I published the original request to add float16 to Parquet, and would love to keep track of the implementation work.

benibus:
@zhangjiashen @wgtmac Just an update, but the Java implementation has recently risen in priority at my company so it's looking like I'll need to continue the work on my branch and (hopefully) have a PR up in the next week or two. However, let me know if there's already been significant work done. If so, I'd be happy to credit/co-author the changes. Regardless, you're welcome to contribute once the PR is up!

zhangjiashen:
> @zhangjiashen @wgtmac Just an update, but the Java implementation has recently risen in priority at my company so it's looking like I'll need to continue the work on my branch and (hopefully) have a PR up in the next week or two. However, let me know if there's already been significant work done. If so, I'd be happy to credit/co-author the changes. Regardless, you're welcome to contribute once the PR is up!

@benibus I am working on Java implementation by continuing your changes and will create a PR once it is ready by end of this week

anjakefala:
@zhangjiashen Please link to the PR here when it is available! =)

benibus:
@zhangjiashen Awesome, thanks! Just let us know if anything changes.

BTW, when you do open the PR, there's already any existing ticket that you can reference [here](https://issues.apache.org/jira/browse/PARQUET-1647) - so no need to create a new one.

zhangjiashen:
> @zhangjiashen Awesome, thanks! Just let us know if anything changes.
> 
> BTW, when you do open the PR, there's already any existing ticket that you can reference [here](https://issues.apache.org/jira/browse/PARQUET-1647) - so no need to create a new one.

@wgtmac  @benibus @anjakefala , created a [PR](https://github.com/apache/parquet-mr/pull/1142) for some changes, please help take a look when you get a chance

cpp/src/arrow/util/float16.h
```diff
@@ -0,0 +1,207 @@
...
+  friend constexpr bool operator<=(Float16 lhs, Float16 rhs) {
+    if (lhs.is_nan() || rhs.is_nan()) return false;
+    return !Float16::CompareLt(rhs, lhs);
+  }
+  friend constexpr bool operator>=(Float16 lhs, Float16 rhs) { return rhs <= lhs; }
+
+  ARROW_FRIEND_EXPORT friend std::ostream& operator<<(std::ostream& os, Float16 arg);
+
+ protected:
+  uint16_t value_;
```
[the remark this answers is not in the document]
benibus: Hmm... I could go either way, but I'm leaning towards keeping it as is - mostly because it mirrors the properties of the native FP types, which seems to be something we're (at least implicitly) aspiring to do here. Plus, if we decide that the compiler/stl optimizations aren't worth the risk, then we could add the default initializer later without as much consequence as the reverse.

cpp/src/arrow/util/float16.h
```diff
@@ -0,0 +1,207 @@
...
+///
+/// The exact format is as follows (from LSB to MSB):
+/// - bits 0-10:  mantissa
+/// - bits 10-15: exponent
+/// - bit 15:     sign
+///
+class ARROW_EXPORT Float16 {
+ public:
+  Float16() = default;
+  constexpr explicit Float16(uint16_t value) : value_(value) {}
```
benibus, line 45: Since the class now supports construction from `float`/`double`, it _seems_ like a bad idea for this to be a public constructor as well (an integer won't implicitly convert to floating point like one might expect) - so I might change some things around. However, LMK if there are any suggestions for the API in general.

benibus:
@ianmcook If you're on this branch, then this best way to do this (for now) is to use [arrow::util::Float16](https://github.com/apache/arrow/blob/b1f5448b1439cfa8f6845ebee6e9d429ef8a1436/cpp/src/arrow/util/float16.h#L42) to downcast to `uint16_t` before calling `Append`.

ianmcook:
@benibus thanks! Using `arrow::util::Float16(float_value).bits()` solves it. I updated the example Gist accordingly.

bkietz (approved):

benibus:
Now that the proposal is merged, would anyone be willing to take a final look and potentially merge this? I've gone ahead and rebased for good measure.

I should also mention that the [Go implementation](https://github.com/apache/arrow/pull/37599) was merged today, but it'll eventually need this PR's `parquet.thrift` to re-generate its files in the future.

cpp/src/arrow/util/float16.h
```diff
@@ -0,0 +1,207 @@
...
+#if ARROW_LITTLE_ENDIAN
+    return ToLittleEndian();
+#else
+    return ToBigEndian();
+#endif
+  }
+
+  /// \brief Copy the value's bytes in little-endian byte order
+  void ToLittleEndian(uint8_t* dest) const {
+    FromBits(::arrow::bit_util::ToLittleEndian(bits_)).ToBytes(dest);
```
pitrou, line 109: This implementation is weird, why call `FromBits` here?
```suggestion
    const auto bytes = ToLittleEndian();
    std::memcpy(dest, bytes.data(), bytes.size());
```

cpp/src/parquet/arrow/test_util.h
```diff
@@ -65,12 +66,30 @@ struct Decimal256WithPrecisionAndScale {
   static constexpr int32_t scale = PRECISION - 1;
 };
 
+inline void RandomHalfFloatValues(int64_t n, uint32_t seed,
```
pitrou, line 69: We don't need to line the definition here, can we put in a `.cc` file instead?

pitrou (approved):
LGTM except for the nits above.

Would it be useful to add tests checking that one write or read a Parquet FLOAT16 with a byte length != 2?

mapleFU (approved):
LGTM

( It's a bit weird for me that fp16 can only use PLAIN, RLE_DICT and ... DELTA_BYTE_ARRAY, aha...)

cpp/src/parquet/arrow/test_util.h
benibus: We could either create a corresponding `parquet/arrow/test_util.cc` or put it in `parquet/test_util.cc`. I'm not sure if the former is worth doing in this case, though.

pitrou: Given this is not Arrow-specific, putting it in `parquet/test_util.cc` would be fine (should probably also move the declaration then).

benibus:
> Would it be useful to add tests checking that one write or read a Parquet FLOAT16 with a byte length != 2?

Probably... I'll try to add something, assuming it wouldn't be redundant.

> ( It's a bit weird for me that fp16 can only use PLAIN, RLE_DICT and ... DELTA_BYTE_ARRAY, aha...)

Yeah, I think we're going to specifically address the encodings as a follow-up since there's been some recent discussion in that area.

benibus:
Update: I don't think it'd be very useful to add tests for different type lengths at the reader/writer level since it isn't even possible to construct a `PrimitiveNode` with an inapplicable length without throwing an exception. I suppose that's why there aren't similar test for other (e.g. decimal, uuid) types AFAICT.

However, it turns out that I forgot to add the relevant tests to `schema_test.cc`, so I went ahead and included those - along with the other requested changes.

pitrou:
LGTM, I'm waiting for a last CI check and will merge afterwards.

mapleFU:
Also we're going to release this in parquet-2.10

https://github.com/apache/parquet-format/pull/219

PR #36077, opened 2023-06-14
GH-36076: [C++] Remove deprecated cli flag

### LINKED ISSUE #36076: [C++] deperacted gbenchmark cli flag causes error

Google benchmark removed the long deprecated `color_print` cli flag in [1.6.1](https://github.com/google/benchmark/commit/ce92bbfb903308bf31530cfb69a36bb06e16d368). This will cause the benchmarks to error when run if benchmark version >=1.6.1 is used. We bundle 1.7.1 since 12.0.0.

We set this flag unconditionally to false (for historic reasons I assume as this change is from 2017) [here](https://github.com/apache/arrow/blob/main/cpp/cmake_modules/BuildUtils.cmake#L570). I think we should not set the flag at all which will default to detecting if the env supports tty colors. If this causes cli issue we should set the correct flag and make it configurable.

### Component(s)

C++

assignUser:
### Rationale for this change
The deprecated flag causes errors on benchmark >= 1.6.1.

No, benchmark defaults to an auto option to determine if colors are supported or not.
* Closes: #36076

### DIFF
--- a/cpp/cmake_modules/BuildUtils.cmake
+++ b/cpp/cmake_modules/BuildUtils.cmake
@@ -613,7 +613,8 @@ function(ADD_BENCHMARK REL_BENCHMARK_NAME)
            ${BUILD_SUPPORT_DIR}/run-test.sh
            ${CMAKE_BINARY_DIR}
            benchmark
-           ${BENCHMARK_PATH}
+           ${BENCHMARK_PATH})
+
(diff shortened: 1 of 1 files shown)


assignUser:

assignUser:
@ursabot please benchmark

anjakefala (approved):

kou (approved):
+1

PR #36094, opened 2023-06-15
GH-36092: [C++] Simplify concurrency in as-of-join node

rtpsw:

The key hasher invalidation and memo-store time updating are moved to the processing thread.

### Are these changes tested?

Yes, by existing tests.

No.
* Closes: #36092

### DIFF
--- a/cpp/src/arrow/acero/asof_join_node.cc
+++ b/cpp/src/arrow/acero/asof_join_node.cc
@@ -344,7 +344,7 @@ struct MemoStore {
   // the time of the current entry, defaulting to 0.
   // when entries with a time less than T are removed, the current time is updated to the
   // time of the next (by-time) and now-current entry or to T if no such entry exists.
-  std::atomic<OnType> current_time_;
+  OnType current_time_;
   // current entry per key
   std::unordered_map<ByType, Entry> entries_;
   // future entries per key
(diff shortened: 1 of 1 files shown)


rtpsw:
cc @westonpace, @icexelloss

rtpsw:
cc @westonpace, @icexelloss

cpp/src/arrow/acero/asof_join_node.cc
```diff
@@ -364,21 +364,16 @@ struct MemoStore {
...
-  // current time was changed from its value at the beginning of this invocation.
+  // Updates the current time to `ts` if it is less. Returns true if updated.
   bool UpdateTime(OnType ts) {
-    OnType prev_time = current_time_;
-    bool update = prev_time < ts;
-    while (prev_time < ts && !current_time_.compare_exchange_weak(prev_time, ts)) {
-      // intentionally empty - standard CAS loop
-    }
+    bool update = current_time_ < ts;
+    if (update) current_time_ = ts;
```
westonpace, line 376:
```suggestion
    current_time_ = std::min(current_time_, ts);
```
Minor nit: Maybe simpler?

westonpace (approved):
This makes sense to me. Let me verify I understand:

Before, on InputReceived, we would update the memo for an input based on the new batch.

Now, the only thing that really happens on InputReceived, is the batch is placed into the appropriate input queue.  The update to the memo table happens when teh processing thread runs, as it is processing batches for an input.

icexelloss:
> This makes sense to me. Let me verify I understand:
> 
> Before, on InputReceived, we would update the memo for an input based on the new batch.
> 
> Now, the only thing that really happens on InputReceived, is the batch is placed into the appropriate input queue. The update to the memo table happens when teh processing thread runs, as it is processing batches for an input.

Yes exactly

icexelloss:
@rtpsw Looks like you added another atomic variable here:
https://github.com/apache/arrow/pull/35874/files#r1218191706

Can you fix that one too?

icexelloss:
> @rtpsw Looks like you added another atomic variable here: https://github.com/apache/arrow/pull/35874/files#r1218191706
> 
> Can you fix that one too?

Nvm looks like you fixed this

icexelloss (approved):
LGTM

cpp/src/arrow/acero/asof_join_node.cc
rtpsw: The method still needs to return a Boolean, so we can't remove the `bool update` line, but we can replace the if-statement.

rtpsw:
> This makes sense to me. Let me verify I understand:
> 
> Before, on InputReceived, we would update the memo for an input based on the new batch.
> 
> Now, the only thing that really happens on InputReceived, is the batch is placed into the appropriate input queue. The update to the memo table happens when teh processing thread runs, as it is processing batches for an input.

As @icexelloss said, this is correct. However, there is another difference to note. The pre-PR code invalidates the key hasher and updates the memo-store time given any batch, whereas the post-PR code does so given all but the first batch (after `batches_processed_` is incremented). This is fine because the key hasher is initialized as invalid and the memo-store time is initialized to the lowest value.

PR #36102, opened 2023-06-15
MINOR: [C++] Make clang-tidy stop complaining about C arrays

felipecrv:
### Rationale for this change

`std::array` is rarely a good fit for the kind of low-level array and buffer manipulation that Arrow does, so having `clang-tidy` complaining about C-style arrays confuses more than it helps.

Change in `.clang-tidy`.

### Are these changes tested?

I checked on my editor. Warnings not shown anymore.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -21,6 +21,7 @@ Checks: |
   -clang-analyzer-alpha*,
   google-*,
   modernize-*,
+  -modernize-avoid-c-arrays,
   -modernize-use-trailing-return-type,
   -modernize-use-nodiscard,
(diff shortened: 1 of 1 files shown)


felipecrv:
@pitrou @bkietz

pitrou (approved):
This sounds good to me. It seems our standard lint routine does not run clang-tidy, does it?

felipecrv:
> This sounds good to me. It seems our standard lint routine does not run clang-tidy, does it?

No, but there is an issue about enabling it https://github.com/apache/arrow/issues/35214

Part of the fix is fixing the warnings or disabling some of them like I'm doing here. :-)

PR #36107, opened 2023-06-15
GH-35705: [R] Rename docs page from acero

### LINKED ISSUE #35705: [R] Rename docs page from `acero`

At the moment, users need to type `?acero` to access the docs which show which tidyverse functions are implemented to run on Acero, but novice users might not know what Acero is, and don't need to, which probably makes "Acero" hard to remember.  Can we rename the docs alias to something like `arrow-tidyverse` to simplify them finding this information?

### Component(s)

R

thisisnic:
Fixes #35705 
* Closes: #35705

### DIFF
--- a/r/R/dplyr-funcs-doc.R
+++ b/r/R/dplyr-funcs-doc.R
@@ -352,4 +352,6 @@
 #' * [`starts_with()`][tidyselect::starts_with()]
 #'
 #' @name acero
+#'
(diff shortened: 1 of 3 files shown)


thisisnic:
I'm not entirely convinced we need all of these aliases - good to hear others' thoughts on this.

paleolimbot (approved):
Thank you!

PR #36108, opened 2023-06-15
GH-36098: [MATLAB] Change C++ proxy constructors to accept an options struct instead of a cell array containing the arguments

### LINKED ISSUE #36098: [MATLAB]: Change C++ proxy constructors to accept an options struct instead of a cell array containing the arguments

It will be cleaner if we pack the input arguments to C++ proxy classes in structs instead of a cell array. This will let us access the arguments by name in the C++ code. 

### Component(s)

MATLAB

sgilmore10:
### Rationale for this change

It would be better if we passed an option struct to the C++ proxy constructors instead of a cell array containing the input arguments. If we pass in a struct, we can access the inputs by name. For example, if the NumericArray proxy class accepted an options struct with fields named `MatlabArray`, `Valid`, and `DeepCopy`, we could access the values like so in its `make` function:

```cpp
        static libmexclass::proxy::MakeResult make(const libmexclass::proxy::FunctionArguments& constructor_arguments) {
            ::matlab::data::StructArray opts = constructor_arguments[0];
            const ::matlab::data::TypedArray<CType> data_mda = opts[0]["MatlabArray"];
            const ::matlab::data::TypedArray<bool> valid_mda = opts[0]["Valid"];
            const ::matlab::data::TypedArray<bool> make_copy_mda = opts[0]["DeepCopy"];
    }
```

It's easier to reason about the code above than the code snippet below:

```cpp
        static libmexclass::proxy::MakeResult make(const libmexclass::proxy::FunctionArguments& constructor_arguments) {
            const ::matlab::data::TypedArray<CType> data_mda = constructor_arguments[0];
            const ::matlab::data::TypedArray<bool> valid_mda = constructor_arguments[1];
            const ::matlab::data::TypedArray<bool> make_copy_mda = constructor_arguments[2];
    }
```
Using options structs also enables us to support syntaxes at construction-time. We can query a field on the struct to determine which fields we should expect to be there. 

1. The `NumericArray` C++ proxy classes accepts a struct with the fields `MatlabArray`, `Valid`, and `DeepCopy` at construction time.
2. The `BooleanArray` C++ proxy class accepts a struct with the fields `MatlabArray` and `Valid` at construction-time.

### Are these changes tested?

Existing test cases cover these changes.

No, these changes are not user-facing.

### Future Directions

In a followup PR, we plan on adding an `ArgumentParser` class. This would abstract out how you access fields on scalar `matlab::data::StructArray` objects. The `NumericArray` code would then look something like this:

```cpp
        static libmexclass::proxy::MakeResult make(const libmexclass::proxy::FunctionArguments& constructor_arguments) {
             ArgumentParser args{constructor_arguments[0]};
             const ::matlab::data::TypedArray<CType> data_mda = args["MatlabArray"];
             const ::matlab::data::TypedArray<bool> valid_mda = args["Valid"];
             const ::matlab::data::TypedArray<bool> make_copy_mda = args["DeepCopy"]; 
    }
```

### Notes

Thank you to @kevingurney for all the help and the idea for the `ArgumentParser` class.

* Closes: #36098

### DIFF
--- a/matlab/src/cpp/arrow/matlab/array/proxy/boolean_array.cc
+++ b/matlab/src/cpp/arrow/matlab/array/proxy/boolean_array.cc
@@ -24,9 +24,11 @@
 namespace arrow::matlab::array::proxy {
 
         libmexclass::proxy::MakeResult BooleanArray::make(const libmexclass::proxy::FunctionArguments& constructor_arguments) {
+            ::matlab::data::StructArray opts = constructor_arguments[0];
+
             // Get the mxArray from constructor arguments
-            const ::matlab::data::TypedArray<bool> logical_mda = constructor_arguments[0];
-            const ::matlab::data::TypedArray<bool> validity_bitmap_mda = constructor_arguments[1];
+            const ::matlab::data::TypedArray<bool> logical_mda = opts[0]["MatlabArray"];
+            const ::matlab::data::TypedArray<bool> validity_bitmap_mda = opts[0]["Valid"];
 
             // Pack the logical data values.
             auto maybe_packed_logical_buffer = arrow::matlab::bit::bitPackMatlabLogicalArray(logical_mda);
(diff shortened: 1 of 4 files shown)


kou (approved):
+1

kou:
@austin3dickeyd Conbench reported that this pull request may have a performance regression but it should not be happen. Because we don't have MATLAB related benchmarks and this pull request changes only MATLAB related codes.

Can we reduce false-positive Conbench reports?

PR #36122, opened 2023-06-16
GH-36120: [C#] Support schema metadata through the C API

### LINKED ISSUE #36120: [C#] Support schema and field metadata in the C Data Interface

.NET support for import and export of schemas over the C data interface was added in #33856, but metadata in fields and schemas is currently ignored.

I'm happy to take a look at this unless this is something @wjones127 or @CurtHagenlocher was already planning on doing?

### Component(s)

C#

CurtHagenlocher:

Import and export of field- and schema-level metadata via the C API.

### Are these changes tested?

Yes

* Closes: #36120

### DIFF
--- a/csharp/src/Apache.Arrow/C/CArrowSchemaExporter.cs
+++ b/csharp/src/Apache.Arrow/C/CArrowSchemaExporter.cs
@@ -16,8 +16,10 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Runtime.InteropServices;
+using System.Text;
 using Apache.Arrow.Types;
 
 namespace Apache.Arrow.C
@@ -83,8 +85,7 @@ namespace Apache.Arrow.C
         {
             ExportType(field.DataType, schema);
             schema->name = StringUtil.ToCStringUtf8(field.Name);
-            // TODO: field metadata
-            schema->metadata = null;
+            schema->metadata = ConstructMetadata(field.Metadata);
             schema->flags = GetFlags(field.DataType, field.IsNullable);
         }
 
@@ -104,8 +105,8 @@ namespace Apache.Arrow.C
         public static unsafe void ExportSchema(Schema schema, CArrowSchema* out_schema)
         {
             var structType = new StructType(schema.FieldsList);
-            // TODO: top-level metadata
             ExportType(structType, out_schema);
+            out_schema->metadata = ConstructMetadata(schema.Metadata);
         }
 
         private static char FormatTimeUnit(TimeUnit unit) => unit switch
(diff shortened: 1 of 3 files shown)


CurtHagenlocher:
@wjones127 has offered to review ;)

csharp/src/Apache.Arrow/C/CArrowSchemaImporter.cs
```diff
@@ -281,21 +283,53 @@ public Field GetAsField()
...
+            private unsafe static IReadOnlyDictionary<string, string> GetMetadata(byte* metadata)
+            {
+                if (metadata == null)
+                {
+                    return null;
+                }
+
+                IntPtr ptr = (IntPtr)metadata;
+                int count = Marshal.ReadInt32(ptr);
+                if (count == 0)
```
wjones127, line 311: Perhaps, to be somewhat defensive:

```suggestion
                if (count <= 0)
```

csharp/src/Apache.Arrow/C/CArrowSchemaImporter.cs
```diff
@@ -281,21 +283,53 @@ public Field GetAsField()
...
+
+                Dictionary<string, string> result = new Dictionary<string, string>(count);
+                for (int i = 0; i < count; i++)
+                {
+                    result[ReadString(ref ptr)] = ReadString(ref ptr);
+                }
+                return result;
+            }
+
+            private unsafe static string ReadString(ref IntPtr ptr)
```
wjones127, line 325: Might want to name more specifically here:

```suggestion
            private unsafe static string ReadMetadataString(ref IntPtr ptr)
```

csharp/test/Apache.Arrow.Tests/CDataInterfacePythonTests.cs
wjones127, line 1: Could you add a test that involves some multi-byte unicode characters?

csharp/src/Apache.Arrow/C/CArrowSchemaExporter.cs
```diff
@@ -239,6 +240,51 @@ private static long GetFlags(IArrowType datatype, bool nullable = true)
...
+                WriteString(ref ptr, lengths[i++], pair.Key);
+                WriteString(ref ptr, lengths[i++], pair.Value);
+            }
+
+            Debug.Assert((long)(IntPtr)ptr - (long)result == size);
+
+            return (byte*)result;
+        }
+
+        private unsafe static void WriteString(ref byte* ptr, int length, string str)
```
wjones127, line 277:
```suggestion
        private unsafe static void WriteMetadataString(ref byte* ptr, int length, string str)
```

csharp/src/Apache.Arrow/C/CArrowSchemaImporter.cs
```diff
@@ -281,21 +283,53 @@ public Field GetAsField()
...
+            {
+                int length = Marshal.ReadInt32(ptr);
```
wjones127, line 327: Perhaps we should also throw an exception if length is negative? Or will `GetString` already do that?

wjones127 (changes requested):
Looks pretty good. Would mainly like to see a test for multi-byte characters, and then have a few smaller nits you can consider.

wjones127 (approved):
Looks good now. Thanks!

PR #36125, opened 2023-06-16
GH-34620: [C#] Support DateOnly and TimeOnly on .NET 6.0+

### LINKED ISSUE #34620: [CSharp] Support encoding `DateOnly` and `TimeOnly` values on .NET 6.

.NET 6 added `DateOnly` and `TimeOnly` types that encode either only a date or a wall clock time. After #34618 gets merged which multi-targets to .NET 6, we can add support for them, when encoding `(Date|Time)(32|64)Array`s.

I will do it.

### Component(s)

C#

CurtHagenlocher:

Date32Array and Date64Array now support DateOnly values for construction and reading on .NET 6.0 and later.
Time32Array and Time64Array now support TimeOnly values for construction and reading on .NET 6.0 and later.
A new TimeArrayBuilder type is used to share logic between Time32Array.Builder and Time64Array.Builder just as the DateArrayBuilder does for the date array types.

### Are these changes tested?

Yes
* Closes: #34620


csharp/src/Apache.Arrow/Arrays/Time64Array.cs
```diff
@@ -113,5 +134,33 @@ public Time64Array(ArrayData data)
...
+            if (value == null)
+            {
+                return null;
+            }
+
+            var unit = ((Time64Type)Data.DataType).Unit;
+            return unit switch
+            {
+                TimeUnit.Microsecond => new TimeOnly(value.Value * TicksPerMicrosecond),
+                TimeUnit.Nanosecond => new TimeOnly(value.Value / NanosecondsPerTick),
```
westonpace, line 160: Should this be rounded?  Although I guess you mention it might be truncated above.

csharp/test/Apache.Arrow.Tests/TimeOnlyTests.cs
```diff
@@ -0,0 +1,111 @@
...
+                // Assert
+                var array = builder.Build();
+                Assert.Equal(1, array.Length);
+                Assert.Null(array.GetTime(0));
+                Assert.Null(array.GetValue(0));
+            }
+
+            [Theory]
+            [MemberData(nameof(GetTestData))]
+            public void AppendTimeGivesSameTime(TimeOnly time, TimeUnit timeUnit)
```
westonpace, line 93: Can you clean up these lint checks (possibly just remove `timeUnit`)?

westonpace (changes requested):
This looks pretty solid.  Just one question about the microsecond/nanosecond tick constants.

csharp/src/Apache.Arrow/Arrays/Time64Array.cs
```diff
@@ -24,17 +25,25 @@ namespace Apache.Arrow
     /// </summary>
     public class Time64Array : PrimitiveArray<long>
     {
+        private const long TicksPerMicrosecond = 1000;
+        private const long NanosecondsPerTick = 100;
```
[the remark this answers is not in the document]
CurtHagenlocher: No, you follow just fine; it's the code that's wrong :/.

csharp/src/Apache.Arrow/Arrays/Date32Array.cs
```diff
@@ -108,5 +118,21 @@ public Date32Array(ArrayData data)
...
+        /// Get the date at the specified index
+        /// </summary>
+        /// <param name="index">Index at which to get the date.</param>
+        /// <returns>Returns a <see cref="DateOnly" />, or <c>null</c> if there is no object at that index.
+        /// </returns>
+        public DateOnly? GetDateOnly(int index)
+        {
+            int? value = GetValue(index);
+            return value.HasValue
+                ? DateOnly.FromDayNumber(_epochDayNumber + value.Value)
```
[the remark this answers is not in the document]
CurtHagenlocher: ArgumentOutOfRangeException. Like System.DateTime, the maximum date is 9999-12-31,

csharp/src/Apache.Arrow/Arrays/Time64Array.cs
CurtHagenlocher: Yes, unfortunately. You'd need to keep it in integer form to deal with higher precision.

csharp/test/Apache.Arrow.Tests/TimeOnlyTests.cs
CurtHagenlocher: There's actually a bug in the test which is why the other bug went uncaught.

westonpace (approved):
Thanks for addressing the feedback.

PR #36130, opened 2023-06-16
GH-36129: [Python] Consolidate common APIs in Table and RecordBatch

### LINKED ISSUE #36129: [Python] Consolidate common python functionality between Table and RecordBatch classes

This will reduce duplication of code in table.pxi. There is a common parent class `_Tabular` that can house these methods.

### Component(s)

Python

danepitkin:
### Rationale for this change

Deduplicate python code in subclasses. The following functions are moved to the shared parent class:
```
__init__
__len__
column_names
from_pylist
from_pydict
itercolumns
to_pylist
to_pydict
```

Refactoring and minor test updates.

### Are these changes tested?

Yes

No
* Closes: #36129

### DIFF
--- a/python/pyarrow/table.pxi
+++ b/python/pyarrow/table.pxi
@@ -1600,6 +1600,28 @@ cdef class _Tabular(_PandasConvertible):
         return self._column(self._ensure_integer_index(i))
 
     @property
+    def column_names(self):
+        """
+        Names of the Table or RecordBatch columns.
+
+        Returns
+        -------
+        list of str
+
+        Examples
+        --------
+        Table (works similarly for RecordBatch)
+
+        >>> import pyarrow as pa
+        >>> table = pa.Table.from_arrays([[2, 4, 5, 100],
+        ...                               ["Flamingo", "Horse", "Brittle stars", "Centipede"]],
+        ...                               names=['n_legs', 'animals'])
+        >>> table.column_names
+        ['n_legs', 'animals']
+        """
+        return [self.field(i).name for i in range(self.num_columns)]
+
+    @property
     def columns(self):
         """
         List of all columns in numerical order.
@@ -1749,6 +1771,18 @@ cdef class _Tabular(_PandasConvertible):
         animals: string
         -- schema metadata --
         n_legs: 'Number of legs per animal'
+
+        Construct a Table from a dictionary of arrays with pyarrow schema:
+
+        >>> my_schema = pa.schema([
+        ...     pa.field('n_legs', pa.int64()),
+        ...     pa.field('animals', pa.string())],
+        ...     metadata={"n_legs": "Number of legs per animal"})
+        >>> pa.Table.from_pydict(pydict, schema=my_schema).schema
+        n_legs: int64
+        animals: string
+        -- schema metadata --
+        n_legs: 'Number of legs per animal'
         """
 
         return _from_pydict(cls=cls,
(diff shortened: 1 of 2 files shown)


AlenkaF (commented):
Thanks for the PR, happy to see this work continue!

I added some minor nit comments.
Also been thinking, if it would be worth including `.itercolumns()` for `RecordBatch` in the tests somehow?

python/pyarrow/table.pxi
```diff
@@ -1513,12 +1513,18 @@ cdef class _Tabular(_PandasConvertible):
...
+
     def __repr__(self):
         if not self._is_initialized():
             raise ValueError("This object's internal pointer is NULL, do not "
                              "use any methods or attributes on this object")
         return self.to_string(preview_cols=10)
 
+    def _column(self, int i):
+        raise NotImplementedError
+
```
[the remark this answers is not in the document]
danepitkin: I plan to leave it as-is for now since it utilizes C objects (`self.table` or `self.batch`). I haven't found a way to consolidate C/Cython code since that code requires the explicit object types in order to compile. If we found a way, we could consolidate a large chunk of Cython! It might not be possible, or if it is possible it might give worse performance.

At this point, I think the best next steps are to leave the C/Cython as-is and focus on implementing missing APIs that might exist in one class and not the other.

python/pyarrow/table.pxi
```diff
@@ -1787,6 +1941,35 @@ cdef class _Tabular(_PandasConvertible):
         """
         return _pc().take(self, indices)
 
+    def to_pydict(self):
+        raise NotImplementedError
```
jorisvandenbossche, line 1945: FWIW, looking at the current `to_pydict` implementations, while they are not exactly the same right now, I think it should be quite easy to rewrite it slightly so that it works for both table and record batch (in the end you iterate over the number of columns, and get each field/column and to_pylist on that)

danepitkin: I'll go ahead and include this change! I'll check some basic timing just to see if anything major happens with performance.

python/pyarrow/table.pxi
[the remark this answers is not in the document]
AlenkaF: Thanks so much for the explanation! Helped me understand this part much better 👍

python/pyarrow/table.pxi
danepitkin: I ran the following code in IPython and received the results below. No performance difference found in my basic testing.

```
import random
import string
import pyarrow as pa

arr0 = [random.randint(0, 2_000_000) for _ in range(1_000_000)]
arr1 = [''.join(random.choices(string.ascii_uppercase + string.digits, k=10)) for _ in range(1_000_000)]
table = pa.table([arr0, arr1], ["c0", "c1"])
```
```%timeit table.to_pydict()```

Before
`2.03 s ± 7.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`
After
`1.99 s ± 7.99 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`

danepitkin:
I feel pretty good about this PR now. I think it could be ready to merge! All comments have been addressed to the best of my knowledge.

AlenkaF (approved):
LGTM +1

jorisvandenbossche (approved):
Thanks!

PR #36132, opened 2023-06-16
MINOR: [Java] Fix dependencies scope in flight-sql module

laurentgo:
### Rationale for this change
flight-sql module advertises dependencies which are not actually used by the published code, causing those dependencies to show up in downstream projects.

Fix dependencies scope in flight-sql module:
* move hamcrest to test scope as it is only used during unit tests
* add optional flag for commons-cli as it is only used in the example class

### Are these changes tested?
No new test. Confirming that there's no build issue

No


lidavidm (approved):
Thank you!

PR #36134, opened 2023-06-16
GH-36133: [C#] Support Set on the `BinaryArray` Builder

### LINKED ISSUE #36133: [C#][Arrow] Implement Set on the BinaryArray Builder

Implement the Set method on the Binary and String Array builders to maintain consistency with other Array Builders which support the Set method.

### Component(s)

C#

jeremyosterhoudt:

Implemented the Set method on both the `BinaryArray` and `StringArray` builders to match functionality of other Builders (`PrimitiveArray.Builder` for example).

Closes #36133
* Closes: #36133

### DIFF
--- a/csharp/src/Apache.Arrow/Arrays/BinaryArray.cs
+++ b/csharp/src/Apache.Arrow/Arrays/BinaryArray.cs
@@ -319,8 +319,8 @@ namespace Apache.Arrow
 
                 return Instance;
 #else
-                //There is not a clean way to shift the bytes with Span.CopyTo which was added in .NetStandard 2.1/.NET5
-                //Given .NetStandard past EOL, keep existing functionality the same for those users.
+                //There is not a clean way to shift the bytes without Span.CopyTo which was added in .NetStandard 2.1/.NET5
+                //For that reason this feature is only avaiable when running versions of .NetStandard 2.1/.NET5 and higher
                 throw new NotImplementedException();
 #endif
             }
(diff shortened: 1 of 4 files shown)


csharp/src/Apache.Arrow/Arrays/BinaryArray.cs
```diff
@@ -258,8 +259,77 @@ public TBuilder Swap(int i, int j)
...
+                {
+                    Set(i + index, newValues[i]);
+                }
+
+                ValidityBuffer.Set(offset, true);
+
+                return Instance;
+#else
+                //There is not a clean way to shift the bytes with Span.CopyTo which was added in .NetStandard 2.1/.NET5
+                //Given .NetStandard past EOL, keep existing functionality the same for those users.
```
danmoseley, line 323: nit, .NET Standard 2.0/2.1 do not have an EOL, they define a standard, not a runtime or product. All future .NET's including .NET 6, 7, and 8 are expected to support .NET Standard 2.1. It's true that .NET Standard is getting increasingly less useful as .NET evolves while .NET Framework is static.

you probably know this -- just suggesting this comment ought to say ".NET 5 and lower are EOL"

jeremyosterhoudt: Great point!  I've updated the comment to reflect that the feature is only available to users running NetStandard 2.1/NET 5 and higher.

danmoseley: Looks good to me, but I am not an Arrow person I just happened to come here so I'll let someone else review...

CurtHagenlocher (changes requested):
I made some suggestions; there also appears to be a failing build.

CurtHagenlocher:
Closed as there is no activity for two years and the source has moved to a new repository.

PR #36137, opened 2023-06-16
GH-35746: [Parquet][C++][Python] Switch default Parquet version to 2.6

### LINKED ISSUE #35746: [Parquet][C++][Python] Bump the default format version from 2.4 -> 2.6

Parquet format version 2.6 introduces the NanoSecond time unit for Time and Timestamp logical types.

### Component(s)

Parquet

anjakefala:
Change the default parquet version to 2.6.

Discussed in: 
* [ML](https://lists.apache.org/thread/027g366yr3m03hwtpst6sr58b3trwhsm)
* [Issue](https://github.com/apache/arrow/issues/35746)

* Closes: #35746


jorisvandenbossche:
Taking a quick look at some of the Python failures:

- `test_write_error_deletes_incomplete_file`: I assume the test relied on parquet.write_table to raise an error because of nanosecods being present, and they would get truncated. So either we need to add something else that errors, or we can add `version="2.4"` to keep it using the older parquet version and thus it still errors
- `test_timestamp_restore_timezone_nanosecond`: it is creating the expected result with microseconds, but now it comes back as nanoseconds. I think here we want to test explicitly that the timezone is preserved when the resolution changes, so I would also add a `version="2.4"` to keep the test as it was. But you could parametrize the test just above to use different resolutions (ms, us, ns) to cover all of them.
- `test_parquet_version_timestamp_differences`: also needs to be updated because the `expected` now contains nanoseconds instead of microseconds (but there are some comments there that are a bit outdated it seems ..)

jorisvandenbossche:
For C++, the test failures are in

```
 [  FAILED  ] 2 tests, listed below:
[  FAILED  ] TestArrowReadWrite.DateTimeTypes
[  FAILED  ] TestArrowReadWrite.ParquetVersionTimestampDifferences
```

and I suppose those will also be updated to reflect the new defaults (nanoseconds are preserved), or updated to test both old default of 2.4 and new of 2.6

anjakefala:
@jorisvandenbossche This was so helpful!

Yeah, with the `_check_roundtrip` calls, it seems that when the version is not explicitly set, it uses the default for `write_table`. And sometimes, the assumption was that this default version was '1.0'. Tested behaviour didn't change for 2.4 and so it didn't matter that the new default version was being used for `_check_roundtrip`.

python/pyarrow/tests/parquet/test_datetime.py
```diff
@@ -378,28 +379,31 @@ def test_parquet_version_timestamp_differences():
...
 
-    # Using Parquet version 2.0, coercing to milliseconds or microseconds
-    # is allowed
     expected = pa.Table.from_arrays([a_us, a_us, a_us, a_us], names)
-    _check_roundtrip(table, expected, version='2.6', coerce_timestamps='us')
+    for ver in all_versions:
+        _check_roundtrip(table, expected, version=ver, coerce_timestamps='us')
 
     # TODO: after pyarrow allows coerce_timestamps='ns', tests like the
     # following should pass ...
```
jorisvandenbossche, line 409: Could also be left for a follow-up, but we could now allow this (in `_create_arrow_writer_properties` in _parquet.pyx we would need to update the cython bindings to pass it through, I think on C++ side it's already implemented)

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -3334,7 +3334,7 @@ TEST(ArrowReadWrite, NestedRequiredOuterOptional) {
     auto arrow_writer_props = ArrowWriterProperties::Builder();
     arrow_writer_props.store_schema();
     if (inner_type->id() == ::arrow::Type::UINT32) {
-      writer_props.version(ParquetVersion::PARQUET_2_4);
+      writer_props.version(ParquetVersion::PARQUET_2_6);
```
jorisvandenbossche, line 3337: The default writer_props created above with `WriterProperties::Builder();` will now use 2_6 by default, so potentially this could be removed alltogether

jorisvandenbossche (commented):
The python changes all look good!

@mapleFU could you have a look at the C++ test changes? 
At the moment it is mostly updating the tests for the new default, not fully sure how well the explicit versions are covered in other tests.

cpp/src/parquet/column_writer_test.cc
```diff
@@ -649,7 +649,7 @@ TYPED_TEST(TestPrimitiveWriter, DictionaryFallbackVersion1_0) {
 }
 
 TYPED_TEST(TestPrimitiveWriter, DictionaryFallbackVersion2_0) {
-  this->TestDictionaryFallbackEncoding(ParquetVersion::PARQUET_2_4);
+  this->TestDictionaryFallbackEncoding(ParquetVersion::PARQUET_2_6);
```
mapleFU, line 652: Emmm should we traverse all 2_0, 2_4, 2_6 ? Or just 2_0 is ok?

mapleFU:
The test in C++ looks ok. But I'm afraid that we might leak some test under 2.4, so maybe we can separate and test 2.4 and 2.6?

cpp/src/parquet/column_writer_test.cc
anjakefala: I am working on having the tests cover at least 2.4 and 2.6! I started something locally, and will push once I have it built.

python/pyarrow/tests/parquet/test_datetime.py
anjakefala: Yes, on the C++ side it appears to be implemented!

```
Expected 'WriteTable(*input_table, ::arrow::default_memory_pool(), CreateOutputStream(), input_table->num_rows(), properties, arrow_coerce_to_nanos_properties)' to fail with NotImplemented, but got OK
```

anjakefala:
`AsofJoinTest` needs to be updated as well.

jorisvandenbossche:

cpp/src/parquet/column_writer_test.cc
mapleFU:
```C++
this->TestDictionaryFallbackEncoding(ParquetVersion::PARQUET_2_4);
this->TestDictionaryFallbackEncoding(ParquetVersion::PARQUET_2_6);
```

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -1973,7 +1983,9 @@ TEST(TestArrowReadWrite, ParquetVersionTimestampDifferences) {
                               field("ts:us", t_us), field("ts:ns", t_ns)});
   auto input_table = Table::Make(input_schema, {a_s, a_ms, a_us, a_ns});
 
-  auto parquet_version_1_properties = ::parquet::default_writer_properties();
```
mapleFU, line 1976: nice catch!

mapleFU (approved):

jorisvandenbossche (approved):

jorisvandenbossche:
There might still be a few cases where the test coverage for all options can be improved (see some of the comments), but if needed, let's do that in a follow-up PR. 
I am going to merge this now, because we have another PR waiting on this one that we also want to get in before the feature freeze.

jorisvandenbossche:
Thanks @anjakefala!

PR #36152, opened 2023-06-17
GH-36151: [Java] Add `volatile` declaration to `keyPosition` in `ParallelSearcher`

### LINKED ISSUE #36151: `keyPosition` variable in ParallelSearcher should be marked `volatile`

`ParallelSearcher` dispatches multiple threads to search a large vector. The main synchronization variable here is `keyPosition`, which can be read and written to by multiple threads. All threads should have a consistent view of the state of this variable, hence it should be marked `volatile`.

### Component(s)

Java

rtadepalli:
### Rationale for this change

Multiple threads can read/write to the `keyPosition` variable. Thus, it should be declared `volatile` so that all threads have a consistent view of the value of this variable.

This PR just marks an instance member `volatile`.

### Are these changes tested?

Existing tests should suffice as no functionality is being changed.

No, there are no user facing changes. There are no changes to the public API either.
* Closes: #36151

### DIFF
--- a/java/algorithm/src/main/java/org/apache/arrow/algorithm/search/ParallelSearcher.java
+++ b/java/algorithm/src/main/java/org/apache/arrow/algorithm/search/ParallelSearcher.java
@@ -52,7 +52,7 @@ public class ParallelSearcher<V extends ValueVector> {
   /**
    * The position of the key in the target vector, if any.
    */
-  private int keyPosition = -1;
+  private volatile int keyPosition = -1;
 
   /**
    * Constructs a parallel searcher.


lidavidm (commented):
Shouldn't the right way to fix this be to use an AtomicInteger and compareAndSet?

rtadepalli:
`keyPosition` is only being read and written to, both of which should be one instruction each. Per my understanding, I don't think we need `AtomicInteger` here; `volatile` should suffice. Please let me know if I am missing something.

lidavidm (approved):

Issue #35728, opened 2023-05-24
[Python] `test_total_bytes_allocated` failing on CI


h-vetinari:
### Describe the bug, including details regarding any error messages, version, and platform.

In https://github.com/conda-forge/arrow-cpp-feedstock/pull/1053, I currently have a failing test `test_total_bytes_allocated` on arrow 12 (though not arrow 11 or before), but some commit between arrow 12 and current `main` seems to have fixed this[^1] - any ideas?

This is currently holding up all our migrations on the arrow-feedstock and is very high priority for us in conda-forge. Possible solutions:
* someone tells me that it's fine to skip that test
* we identify the relevant changes that regressed this from 11->12 and revert them (on the feedstock)
* we identify what fixed the issue between 12 & main and I backport those changes (on the feedstock)

[^1]: as confirmed in #35637

_Originally posted by @h-vetinari in https://github.com/apache/arrow/issues/35637#issuecomment-1552650935_
            

### Component(s)

Packaging

assignUser:
@westonpace is it fine to skip that test or do you know if that fix was already added for 12.0.1?

h-vetinari:
Thanks @assignUser for chiming in! Some more context. It did not appear for the last successful [build](https://github.com/conda-forge/arrow-cpp-feedstock/pull/1042) of 12.0.0, then appeared for a [migration](https://github.com/conda-forge/arrow-cpp-feedstock/pull/1053) (of simultaneously re2, aws-crt & google-cloud-cpp; because reasons...), but only on linux, and not for any of the older arrow versions (which received the same migrator PRs).

> Single test failure here, and only on linux. Normally I'd go with a skip, but this looks like it might be leaking memory? Any ideas? The test in question [has no](https://github.com/apache/arrow/blob/apache-arrow-12.0.0/python/pyarrow/tests/test_array.py#L40-L41) comments or documentation AFAICS.
> 
> ```
> =================================== FAILURES ===================================
> __________________________ test_total_bytes_allocated __________________________
> 
>     def test_total_bytes_allocated():
> >       assert pa.total_allocated_bytes() == 0
> E       assert 1216 == 0
> E        +  where 1216 = <built-in function total_allocated_bytes>()
> E        +    where <built-in function total_allocated_bytes> = pa.total_allocated_bytes
> ```
> 
> It's also interesting that this doesn't appear in any of the equivalent PRs to the maintenance branches, e.g. #1057.

[...]

PR #36162, opened 2023-06-19
GH-21761: [Python] accept pyarrow scalars in array constructor

### LINKED ISSUE #21761: [Python] accept pyarrow values / scalars in constructor functions ?
Currently, functions like \`pyarrow.array` don't accept pyarrow Arrays, or also not scalars of it:

```Java

In [42]: arr = pa.array([1, 2, 3])

In [43]: pa.array(arr)
...
ArrowInvalid: Could not convert 1 with type pyarrow.lib.Int64Value: did not recognize Python value type when inferring an Arrow data type

In [44]: pa.array(list(arr))
...
ArrowInvalid: Could not convert 1 with type pyarrow.lib.Int64Value: did not recognize Python value type when inferring an Arrow data type
```

Do we want to allow those / recognize those here? (the first case could even have a fastpath, as we don't need to do it element by element).

Also scalars are not supported:

```Java

In [46]: type(arr.sum())
Out[46]: pyarrow.lib.Int64Scalar

In [47]: pa.array([arr.sum()])
...
ArrowInvalid: Could not convert 6 with type pyarrow.lib.Int64Scalar: did not recognize Python value type when inferring an Arrow data type
```

And also in other functions we don't accept arrow scalars / values:

```Java

In [48]: string = pa.array(['a'])[0]

In [49]: type(string)
Out[49]: pyarrow.lib.StringValue

In [50]: pa.field(string, pa.int64())
...
TypeError: expected bytes, pyarrow.lib.StringValue found
```
 

 

**Reporter**: [Joris Van den Bossche](https://issues.apache.org/jira/browse/ARROW-5295) / @jorisvandenbossche
#### Related issues:
- [[Python] pa.array() doesn't support pa.lib.TimestampScalar objects](https://github.com/apache/arrow/issues/26862) (duplicates)
- [[Python] Array constructor doesn't support arrow scalars.](https://github.com/apache/arrow/issues/33594) (is duplicated by)
- [[Python] Scalar constructors should be callable](https://github.com/apache/arrow/issues/31648) (relates to)
- [[Python] Allow pa.array/pa.chunked_array to infer pa.NA when in a non pyarrow container](https://github.com/apache/arrow/issues/31995) (relates to)

<sub>**Note**: *This issue was originally created as [ARROW-5295](https://issues.apache.org/jira/browse/ARROW-5295). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

AlenkaF:
### Rationale for this change

Currently, `pyarrow.array `doesn't accept list of pyarrow Scalars and this PR adds a check to allow that.

* Closes: #21761

### DIFF
--- a/python/pyarrow/src/arrow/python/inference.cc
+++ b/python/pyarrow/src/arrow/python/inference.cc
@@ -29,7 +29,6 @@
 
 #include "arrow/scalar.h"
 #include "arrow/status.h"
-#include "arrow/type.h"
 #include "arrow/util/decimal.h"
 #include "arrow/util/logging.h"
 
--- a/python/pyarrow/src/arrow/python/inference.cc
+++ b/python/pyarrow/src/arrow/python/inference.cc
@@ -462,6 +462,12 @@ class TypeInferrer {
 
     RETURN_NOT_OK(Validate());
 
+    if (arrow_scalar_count_ > 0 and arrow_scalar_count_ + none_count_ != total_count_){
+      return Status::Invalid(
+              "pyarrow scalars cannot be mixed "
+              "with other Python scalar values currently");
+    }
+
     if (numpy_dtype_count_ > 0) {
       // All NumPy scalars and Nones/nulls
       if (numpy_dtype_count_ + none_count_ == total_count_) {
@@ -501,12 +507,6 @@ class TypeInferrer {
               "numpy.datetime64 scalars cannot be mixed "
               "with other Python scalar values currently");
       }
-    }
-
-    if (arrow_scalar_count_ > 0 and arrow_scalar_count_ + none_count_ != total_count_){
-      return Status::Invalid(
-              "pyarrow scalars cannot be mixed "
-              "with other Python scalar values currently");
     }
 
     if (list_count_) {
@@ -576,19 +576,13 @@ class TypeInferrer {
   }
 
   Status VisitArrowScalar(PyObject* obj, bool* keep_going) {
-    Result<std::shared_ptr<Scalar>> result = arrow::py::unwrap_scalar(obj);
-    if (!result.ok()){
-      return internal::InvalidValue(obj, "Oh, what?");
-    }
-    std::shared_ptr<Scalar> scalar = result.ValueOrDie();
-
+    ARROW_ASSIGN_OR_RAISE(auto scalar, arrow::py::unwrap_scalar(obj));
     // Check that all the scalar types for the sequence are the same
-    std::shared_ptr<DataType> type = (*scalar->type).GetSharedPtr();
-    if (arrow_scalar_count_ > 0 and type != scalar_type_) {
+    if (arrow_scalar_count_ > 0 and *scalar->type != *scalar_type_) {
       return internal::InvalidValue(obj,
                                     "cannot mix scalars with different types");
     }
-    scalar_type_ = type;
+    scalar_type_ = scalar->type;
     ++arrow_scalar_count_;
     return Status::OK();
   }
--- a/python/pyarrow/src/arrow/python/inference.cc
+++ b/python/pyarrow/src/arrow/python/inference.cc
@@ -79,7 +79,7 @@ Status ImportPresentIntervalTypes(OwnedRefNoGIL* interval_types_tuple) {
 
 }  // namespace
 
-int import_pyarrow();
+import_pyarrow();
 
 #define _NUMPY_UNIFY_NOOP(DTYPE) \
   case NPY_##DTYPE:              \
--- a/python/pyarrow/src/arrow/python/inference.cc
+++ b/python/pyarrow/src/arrow/python/inference.cc
@@ -27,7 +27,9 @@
 #include <utility>
 #include <vector>
 
+#include "arrow/scalar.h"
 #include "arrow/status.h"
+#include "arrow/type.h"
 #include "arrow/util/decimal.h"
 #include "arrow/util/logging.h"
 
@@ -76,6 +78,8 @@ Status ImportPresentIntervalTypes(OwnedRefNoGIL* interval_types_tuple) {
 }
 
 }  // namespace
+
+int import_pyarrow();
 
 #define _NUMPY_UNIFY_NOOP(DTYPE) \
   case NPY_##DTYPE:              \
@@ -340,6 +344,7 @@ class TypeInferrer {
         decimal_count_(0),
         list_count_(0),
         struct_count_(0),
+        arrow_scalar_count_(0),
         numpy_dtype_count_(0),
         interval_count_(0),
         max_decimal_metadata_(std::numeric_limits<int32_t>::min(),
@@ -391,6 +396,8 @@ class TypeInferrer {
     } else if (PyUnicode_Check(obj)) {
       ++unicode_count_;
       *keep_going = make_unions_;
+    } else if(arrow::py::is_scalar(obj)) {
+      RETURN_NOT_OK(VisitArrowScalar(obj, keep_going));
     } else if (PyArray_CheckAnyScalarExact(obj)) {
       RETURN_NOT_OK(VisitDType(PyArray_DescrFromScalar(obj), keep_going));
     } else if (PySet_Check(obj) || (Py_TYPE(obj) == &PyDictValues_Type)) {
@@ -496,6 +503,12 @@ class TypeInferrer {
       }
     }
 
+    if (arrow_scalar_count_ > 0 and arrow_scalar_count_ + none_count_ != total_count_){
+      return Status::Invalid(
+              "pyarrow scalars cannot be mixed "
+              "with other Python scalar values currently");
+    }
+
     if (list_count_) {
       std::shared_ptr<DataType> value_type;
       RETURN_NOT_OK(list_inferrer_->GetType(&value_type));
@@ -534,6 +547,8 @@ class TypeInferrer {
       *out = utf8();
     } else if (interval_count_) {
       *out = month_day_nano_interval();
+    } else if (arrow_scalar_count_) {
+      *out = scalar_type_;
     } else {
       *out = null();
     }
@@ -557,6 +572,24 @@ class TypeInferrer {
         RETURN_NOT_OK(it.second.Validate());
       }
     }
+    return Status::OK();
+  }
+
+  Status VisitArrowScalar(PyObject* obj, bool* keep_going) {
+    Result<std::shared_ptr<Scalar>> result = arrow::py::unwrap_scalar(obj);
+    if (!result.ok()){
+      return internal::InvalidValue(obj, "Oh, what?");
+    }
+    std::shared_ptr<Scalar> scalar = result.ValueOrDie();
+
+    // Check that all the scalar types for the sequence are the same
+    std::shared_ptr<DataType> type = (*scalar->type).GetSharedPtr();
+    if (arrow_scalar_count_ > 0 and type != scalar_type_) {
+      return internal::InvalidValue(obj,
+                                    "cannot mix scalars with different types");
+    }
+    scalar_type_ = type;
+    ++arrow_scalar_count_;
     return Status::OK();
   }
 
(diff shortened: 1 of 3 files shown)


python/pyarrow/tests/test_convert_builtin.py
```diff
@@ -2363,3 +2363,113 @@ def test_array_from_pylist_offset_overflow():
...
+    sequence = seq([pa.scalar("a"), np.float16("0.1"), None])
+    with pytest.raises(pa.ArrowInvalid,
+                       match="pyarrow scalars cannot be mixed with other "
+                             "Python scalar values currently"):
+        pa.array(sequence)
+
+    with pytest.raises(pa.ArrowInvalid,
+                       match="Cannot append scalar of type string "
+                             "to builder for type int32"):
+        pa.array([pa.scalar("a")], type=pa.int32())
```
AlenkaF, line 2475: This should fail with a different message.
Also casting `int64` to `int8` should work but it currently doesn't.

python/pyarrow/src/arrow/python/inference.cc
```diff
@@ -77,6 +79,8 @@ Status ImportPresentIntervalTypes(OwnedRefNoGIL* interval_types_tuple) {
 
 }  // namespace
 
+int import_pyarrow();
```
jorisvandenbossche, line 82:
```suggestion
import_pyarrow();
```

python/pyarrow/tests/test_convert_builtin.py
```diff
@@ -2363,3 +2363,113 @@ def test_array_from_pylist_offset_overflow():
...
+    ([None, datetime.date.today()], [None, pa.scalar(datetime.date.today())]),
+    ([datetime.time(1, 1, 1), None], [pa.scalar(datetime.time(1, 1, 1)), None]),
+    ([datetime.timedelta(seconds=10)], [pa.scalar(datetime.timedelta(seconds=10))]),
+    ([None, datetime.datetime(2014, 1, 1)], [
+     None, pa.scalar(datetime.datetime(2014, 1, 1))]),
+    ([pa.MonthDayNano([1, -1, -10100])], [pa.scalar(pa.MonthDayNano([1, -1, -10100]))]),
+    (["a", "b"], [pa.scalar("a"), pa.scalar("b")]),
+    ([b"a", b"b"], [pa.scalar(b"a"), pa.scalar(b"b")]),
+    ([1, 2, 3], pa.scalar([1, 2, 3])),
+    (["a", "b"], pa.scalar(["a", "b"])),
```
jorisvandenbossche, line 2384:
```suggestion
    ([[1, 2, 3]], [pa.scalar([1, 2, 3])]),
    ([["a", "b"]], [pa.scalar(["a", "b"])]),
```

?

python/pyarrow/tests/test_convert_builtin.py
jorisvandenbossche: > Also casting `int64` to `int8` should work but it currently doesn't.

I suppose it is the builder that currently doesn't support that? 
(it might also be fine to not support this for this initial PR)

python/pyarrow/src/arrow/python/inference.cc
AlenkaF: This change produced an error:
```
/arrow/python/pyarrow/src/arrow/python/inference.cc
  /arrow/python/pyarrow/src/arrow/python/inference.cc:82:17: error: expected constructor, destructor, or type conversion before ';' token
     82 | import_pyarrow();
        |   
```

I do not fully understand the usage of this function. But if we need to call it before using any other function in the pyarrow C++ we could do an if statement or something?

AlenkaF:
I am not sure why linter is failing on this PR. I will try to rebase and see if it helps.

python/pyarrow/src/arrow/python/inference.cc
AlenkaF: Removing `import_pyarrow();` altogether also works so I went with that.

AlenkaF:
The tests failing already have an open issue: https://github.com/apache/arrow/issues/35728, so I think this PR is ready for review.

One missing thing are the test for sets where the elements in the set get reordered when getting passed to the `pa.array` constructor. Not sure there is a solution for it.
https://github.com/apache/arrow/blob/3da9ba67144643d5331aeda7accb5c5795aeb1df/python/pyarrow/tests/test_convert_builtin.py#L2448-L2449

python/pyarrow/src/arrow/python/inference.cc
```diff
@@ -560,6 +573,17 @@ class TypeInferrer {
     return Status::OK();
   }
 
+  Status VisitArrowScalar(PyObject* obj, bool* keep_going) {
```
danepitkin, line 576: I think we can add an `/* unused */` comment for `keep_going` since we do want to check all values.

python/pyarrow/tests/test_convert_builtin.py
```diff
@@ -2363,3 +2363,136 @@ def test_array_from_pylist_offset_overflow():
     assert isinstance(arr, pa.ChunkedArray)
     assert len(arr) == 2**31
     assert len(arr.chunks) > 1
+
+
```
danepitkin, line 2367: Great test cases! I see we aren't testing all scalar types defined in https://arrow.apache.org/docs/python/api/datatypes.html, but I don't think we need to since many of them take identical code paths, right? e.g. date32 vs date64

danepitkin (commented):
Nice work!

python/pyarrow/tests/test_convert_builtin.py
AlenkaF: Yeah, I took approx one example for the group of datatypes represented in the specific `PyPrimitiveConverter` class that holds an `Append` method in `python/pyarrow/src/arrow/python/python_to_arrow.cc`. There should be no difference between for example `date32` vs `date64`. But will add an extra case to the test just to be sure 👍

python/pyarrow/tests/test_convert_builtin.py
```diff
@@ -2363,3 +2363,144 @@ def test_array_from_pylist_offset_overflow():
...
+@parametrize_with_collections_types
+@pytest.mark.parametrize(('data', 'scalar_data', 'value_type'), [
+    ([True, False, None], [pa.scalar(True), pa.scalar(False), None], pa.bool_()),
+    ([1, 2, None], [pa.scalar(1), pa.scalar(2), None], pa.int64()),
```
jorisvandenbossche, line 2371:
```suggestion
    ([1, 2, None], [pa.scalar(1), pa.scalar(2), pa.scalar(None, pa.int64())], pa.int64()),
```

To have a case with a null scalar instead of None as well

python/pyarrow/tests/test_convert_builtin.py
```diff
@@ -2363,3 +2363,144 @@ def test_array_from_pylist_offset_overflow():
...
+
+    result = pa.array(seq(scalar_data), type=value_type)
+    assert expect.equals(result)
+
+
+def test_array_accepts_pyarrow_scalar_from_compute():
+    arr = pa.array([1, 2, 3])
+    result = pa.array([arr.sum()])
+    expect = pa.array([6])
+    assert expect.equals(result)
```
jorisvandenbossche, line 2470: Given that this is essentially tested by the test above (we know that the sum() returns a scalar), you can maybe drop this one

python/pyarrow/src/arrow/python/inference.cc
```diff
@@ -27,7 +27,9 @@
 #include <utility>
 #include <vector>
 
+#include "arrow/scalar.h"
 #include "arrow/status.h"
+#include "arrow/type.h"
```
jorisvandenbossche, line 32: Is this additional include needed? (we already use shared pointers to DataType in this file, so I would expect your diff to not need this)

jorisvandenbossche (commented):
Looks good! Some small comments

python/pyarrow/src/arrow/python/inference.cc
AlenkaF: Correct, I guess it must be a leftover. Will remove it 👍

AlenkaF:
The failing tests are all due to a known issue with `test_total_bytes_allocated`.

python/pyarrow/tests/test_convert_builtin.py
```diff
@@ -2363,3 +2363,141 @@ def test_array_from_pylist_offset_overflow():
...
+    (
+        [(1, 'bar')],
+        [pa.scalar([('a', 1), ('b', 'bar')], type=pa.struct(
+            [('a', pa.int8()), ('b', pa.string())]))],
+        pa.struct([('a', pa.int8()), ('b', pa.string())])
+    )
+])
+def test_array_accepts_pyarrow_scalar(seq, data, scalar_data, value_type):
+    if type(seq(scalar_data)) == set:
+        pytest.skip("TODO: The elements in the set get reordered.")
```
jorisvandenbossche, line 2461: One more thing I forgot, we should maybe resolve this comment. Either just keep skipping it (and remove the TODO), or actually address it. 

The actual behaviour is expected: the iteration order of a set is not defined (and not the same as the order it was created), so it is OK that this doesn't pass out of the box. The question is whether we want to change the assert specifically for this case, or just leave it as is (the skip)

AlenkaF: I would change the assert. Will think about how to do it and hopefully make a commit soon =)

AlenkaF: Created (probably not the nicest way) to test sets also ([bfb17c6](https://github.com/apache/arrow/pull/36162/commits/bfb17c693cce362998f3bc80675e22f01d935123)).

python/pyarrow/tests/test_convert_builtin.py
```diff
@@ -2457,14 +2457,29 @@ def test_array_from_pylist_offset_overflow():
...
-    assert expect.equals(result)
+
+    if type(seq(scalar_data)) == set:
+        try:
+            import pyarrow.compute as pc
+            expect = [expect[a] for a in pc.array_sort_indices(expect).to_pylist()]
+            result = [result[a] for a in pc.array_sort_indices(result).to_pylist()]
+            expect == result
+        except pa.lib.ArrowNotImplementedError:
+            set(expect) == set(result)
```
jorisvandenbossche, line 2470: I think it's probably good enough to just always use the fallback (or could also be `sorted(expect.to_pylist()) == sorted(result.to_pylist())` ?)

(that would avoid some of the complexity)

AlenkaF: That was my first try, but in the case of dictionary type the set mixed the elements differently 🤷‍♀️ 
But `sorted(expect.to_pylist()) == sorted(result.to_pylist())` is a great idea! Tried without the `to_pylist` but it ofcourse didn't work. Will update today 👍

AlenkaF: There is always a case that doesn't fit in this PR 🤣  

```python
TypeError: unhashable type: 'list'
TypeError: unhashable type: 'dict'
```

I can do `try-except` and take the items of the lists in the case of a `TypeError` from above (in case of ListType and StructType as `to_pylist()` returns a dictionary in a list).

jorisvandenbossche: > There is always a case that doesn't fit in this PR

Which case exactly gives that error message with ``sorted(expect.to_pylist()) == sorted(result.to_pylist())``?

AlenkaF: `ListType` and `StructType`, that is:
- `data = [(1, 'bar')], scalar_data = [<pyarrow.StructScalar: [('a', 1), ('b', 'bar')]>], value_type = StructType(struct<a: int8, b: string>)`
- `data = [(1, 2)], scalar_data = [<pyarrow.StructScalar: [('a', 1), ('b', 2)]>], value_type = StructType(struct<a: int8, b: int8>)`
- `data = [['a', 'b']], scalar_data = [<pyarrow.ListScalar: ['a', 'b']>], value_type = ListType(list<item: string>)`
- `data = [[1, 2, 3]], scalar_data = [<pyarrow.ListScalar: [1, 2, 3]>], value_type = ListType(list<item: int64>)`

jorisvandenbossche: Hmm. Sorry for the back and forth here, but then I would maybe go back to the initial skip (with an updated comment). I don't think this is worth the complexity. We know that sets can be iterated multiple times (just with an non-deterministic order), and won't give a problem here, accept being difficult to test with general data. 
(eg, while testing some other options, I also noticed that for the case of null type, for a set it actually created a different array because of de-duplicating the sequence)

AlenkaF: Agree, will do 👍

jorisvandenbossche (approved):

AlenkaF:
Will merge this today as the CI is green 👍

jorisvandenbossche:
Thanks Alenka!

PR #36163, opened 2023-06-19
GH-36095: [Go] Add doc for `pqarrow.FileWriter.WriteBuffered`

### LINKED ISSUE #36095: Make `WriteBuffered` a default

A follow-up to https://github.com/cloudquery/filetypes/pull/203.

Basically, we saw a drastic improvement after switching to a buffered mode, so maybe making `Write` be buffered and introducing a `WriteUnbuffered` func to perform as current `Write` is a better way?

### Component(s)

Go

candiduslynx:
### Rationale for this change

Docs to help people decide on the best-performing option.

Doc change only.

### Are these changes tested?

N/A

Doc

* Closes: #36095

### DIFF
--- a/go/parquet/pqarrow/file_writer.go
+++ b/go/parquet/pqarrow/file_writer.go
@@ -135,7 +135,7 @@ func (fw *FileWriter) RowGroupTotalBytesWritten() int64 {
 }
 
 // WriteBuffered will either append to an existing row group or create a new one
-// based on the record length and RowGroupTotalBytesWritten.
+// based on the record length and max row group length.
 //
 // Additionally, it allows to manually break your row group by
 // checking RowGroupTotalBytesWritten and calling NewBufferedRowGroup,
--- a/go/parquet/pqarrow/file_writer.go
+++ b/go/parquet/pqarrow/file_writer.go
@@ -105,7 +105,7 @@ func (fw *FileWriter) NewRowGroup() {
 
 // NewBufferedRowGroup starts a new memory Buffered Row Group to allow writing columns / records
 // without immediately flushing them to disk. This allows using WriteBuffered to write records
-// and decide where to break your rowgroup based on the TotalBytesWritten rather than on the max
+// and decide where to break your row group based on the TotalBytesWritten rather than on the max
 // row group len. If using Records, this should be paired with WriteBuffered, while
 // Write will always write a new record as a row group in and of itself.
 func (fw *FileWriter) NewBufferedRowGroup() {
@@ -134,6 +134,13 @@ func (fw *FileWriter) RowGroupTotalBytesWritten() int64 {
 	return 0
 }
 
+// WriteBuffered allows to write records and decide where to break your row group
+// based on the TotalBytesWritten rather than on the max row group len.
+// If using Records, this should be paired with NewBufferedRowGroup,
+// while Write will always write a new record as a row group in and of itself.
+//
+// Performance-wise WriteBuffered might be more favorable than Write
+// especially if dealing with lots of records that have only a small amount of rows.
 func (fw *FileWriter) WriteBuffered(rec arrow.Record) error {
 	if !rec.Schema().Equal(fw.schema) {
 		return fmt.Errorf("record schema does not match writer's. \nrecord: %s\nwriter: %s", rec.Schema(), fw.schema)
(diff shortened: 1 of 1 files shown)


go/parquet/pqarrow/file_writer.go
```diff
@@ -134,6 +134,13 @@ func (fw *FileWriter) RowGroupTotalBytesWritten() int64 {
 	return 0
 }
 
+// WriteBuffered allows to write records and decide where to break your row group
+// based on the TotalBytesWritten rather than on the max row group len.
```
zeroshade, line 138: TotalBytesWritten will still break row groups on the max row group len, not TotalBytesWritten. It does, however, allow users to use TotalBytesWritten to decide when to break row groups themselves by checking `TotalBytesWritten` and calling `NewBuffereredRowGroup` themselves.

go/parquet/pqarrow/file_writer.go
```diff
@@ -134,6 +134,13 @@ func (fw *FileWriter) RowGroupTotalBytesWritten() int64 {
...
+// If using Records, this should be paired with NewBufferedRowGroup,
+// while Write will always write a new record as a row group in and of itself.
+//
+// Performance-wise WriteBuffered might be more favorable than Write
+// especially if dealing with lots of records that have only a small amount of rows.
```
zeroshade, line 143: Mention that the tradeoff is that more memory will be utilized to keep the whole row group buffered in memory before it starts writing (since Parquet files must write an entire column before writing the next column).

go/parquet/pqarrow/file_writer.go
```diff
@@ -181,7 +188,10 @@ func (fw *FileWriter) WriteBuffered(rec arrow.Record) error {
 }
 
 // Write an arrow Record Batch to the file, respecting the MaxRowGroupLength in the writer
-// properties to determine whether or not a new row group is created while writing.
+// properties to determine whether a new row group is created or not while writing.
```
zeroshade, line 191: Let's update this to say `to determine whether the record is broken up into more than one row group` and add a statement that it will always create at least one row group for the record, or something like that.

go/parquet/pqarrow/file_writer.go
```diff
@@ -181,7 +188,10 @@ func (fw *FileWriter) WriteBuffered(rec arrow.Record) error {
...
+//
+// Performance-wise Write might be more favorable than WriteBuffered
+// especially if dealing with records that have a lot of rows.
```
zeroshade, line 194: Let's mention that this would be more performant in the case of a highly restricted memory environment, or if you have very large records with lots of rows, potentially close to the `max row group length`.

go/parquet/pqarrow/file_writer.go
candiduslynx: Done in 9c4ca94137a143986e69feb07a0ab94baee7e334

go/parquet/pqarrow/file_writer.go
candiduslynx: Done in 9c4ca94137a143986e69feb07a0ab94baee7e334

go/parquet/pqarrow/file_writer.go
candiduslynx: Done in 9c4ca94137a143986e69feb07a0ab94baee7e334

go/parquet/pqarrow/file_writer.go
candiduslynx: Done in 9c4ca94137a143986e69feb07a0ab94baee7e334

go/parquet/pqarrow/file_writer.go
```diff
@@ -134,13 +134,19 @@ func (fw *FileWriter) RowGroupTotalBytesWritten() int64 {
 	return 0
 }
 
-// WriteBuffered allows to write records and decide where to break your row group
-// based on the TotalBytesWritten rather than on the max row group len.
-// If using Records, this should be paired with NewBufferedRowGroup,
-// while Write will always write a new record as a row group in and of itself.
+// WriteBuffered will either append to an existing row group or create a new one
+// based on the record length and RowGroupTotalBytesWritten.
```
zeroshade, line 138: `WriteBuffered` itself never checks `RowGroupTotalBytesWritten`, unless i'm missing something.

candiduslynx: fixed a typo in 25d51d5a58e86ca527d31547ec21e6878e7a852f

zeroshade (approved):
LGTM thanks!

PR #36165, opened 2023-06-19
MINOR: [Go] DictionaryBuilder reads null bitmap from idxBuilder

metalmatze:
This updates the `dictionaryBuilder` to read the idxBuilder's nullBitmap such that it not only panics but reads the correct NULLs. 

### Rationale for this change

Fix the method for the dictionaryBuilder. 

This updates the `dictionaryBuilder` to read the idxBuilder's nullBitmap such that it not only panics but reads the correct NULLs. 

### Are these changes tested?

Yes, updated the unit test.

No

### DIFF
--- a/go/arrow/array/dictionary.go
+++ b/go/arrow/array/dictionary.go
@@ -717,6 +717,8 @@ func (b *dictionaryBuilder) ResetFull() {
 
 func (b *dictionaryBuilder) Cap() int { return b.idxBuilder.Cap() }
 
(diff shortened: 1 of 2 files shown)


metalmatze:
Turns out this is actually not a bug where the underlying bitmap is null, but we need to overwrite the dictionaryBuilder to use the idxBuilder's `IsNull`.
```go
func (b *dictionaryBuilder) IsNull(i int) bool { return b.idxBuilder.IsNull(i) }
```

I'll update this PR tomorrow.

zeroshade (approved):

PR #36169, opened 2023-06-20
GH-36198: [Go] Remove deprecated equality checks

### LINKED ISSUE #36198: Remove deprecated equality checks

There are several [deprecated comparison functions](https://github.com/apache/arrow/blob/e6221c45de7b4c675bcedfac9afb87f6a131dad9/go/arrow/array/compare.go) that have long been deprecated and were supposed to be removed in v9, we're working on v13 now. 

I propose to remove those functions or at least remove several usages of those functions in the Go arrow package itself. 

### Component(s)

Go

metalmatze:
### Rationale for this change

There are several deprecated comparison functions that have long been deprecated.
This PR unblocks the removal of those functions as nothing depends on them anymore.

Given the functions have been deprecated for a long time and might have been removed in v9, I removed them in this PR already, given this is for v13. 

### Are these changes tested?

Unit tests are passing locally. 

Removal of old deprecated functions for comparing arrays. 

* Closes: #36198

### DIFF
--- a/go/arrow/array/compare.go
+++ b/go/arrow/array/compare.go
@@ -37,7 +37,7 @@ func RecordEqual(left, right arrow.Record) bool {
 	for i := range left.Columns() {
 		lc := left.Column(i)
 		rc := right.Column(i)
-		if !ArrayEqual(lc, rc) {
+		if !Equal(lc, rc) {
 			return false
 		}
 	}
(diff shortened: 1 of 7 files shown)


kou:
Could you open an issue for this pull request because this is not a MINOR change?
See also our MINOR definition: https://github.com/apache/arrow/blob/main/CONTRIBUTING.md#Minor-Fixes

zeroshade:
I agree with @kou's statement that this isn't a MINOR change and should have an issue opened for it. Also it looks like you missed one `ArrayEqual` in the `cdata_test.go` file.

zeroshade (approved):

PR #36180, opened 2023-06-20
GH-36178: [C++]support prefetching for ReadRangeCache lazy mode

### LINKED ISSUE #36178: [C++] Support prefetch in ReadRangeCache Lazy mode

The current [ReadRangeCache](https://github.com/apache/arrow/blob/main/cpp/src/arrow/io/caching.h#L100) support a laze mode where each range is read upon an [explicit request](https://github.com/apache/arrow/blob/main/cpp/src/arrow/io/caching.cc#L255), and a non-lazy mode where all ranges are read concurrently [at once](https://github.com/apache/arrow/blob/main/cpp/src/arrow/io/caching.cc#L168). In some scenarios, it would help to support a prefetch mode (which naturally is bound to the lazy mode) such that  explicitly reading one range will prefetching one or more following ranges (the number of prefetched ranges should be configurable via cache options). 

### Component(s)

C++

jp0317:
### Rationale for this change

The current [ReadRangeCache](https://github.com/apache/arrow/blob/main/cpp/src/arrow/io/caching.h#L100) support a laze mode where each range is read upon an [explicit request](https://github.com/apache/arrow/blob/main/cpp/src/arrow/io/caching.cc#L255), and a non-lazy mode where all ranges are read concurrently [at once](https://github.com/apache/arrow/blob/main/cpp/src/arrow/io/caching.cc#L168). In some scenarios, it would help to support a prefetch mode (which naturally is bound to the lazy mode) such that  explicitly reading one range will prefetching one or more following ranges . 

Add a new cache option `prefetch_limit` to define the maximum number of ranges to be prefetched when reading one range in lazy mode.

### Are these changes tested?

Yes. A new unit test is included

No.
* Closes: #36178

### DIFF
--- a/cpp/src/arrow/io/caching.cc
+++ b/cpp/src/arrow/io/caching.cc
@@ -35,13 +35,15 @@ namespace io {
 CacheOptions CacheOptions::Defaults() {
   return CacheOptions{internal::ReadRangeCache::kDefaultHoleSizeLimit,
                       internal::ReadRangeCache::kDefaultRangeSizeLimit,
-                      /*lazy=*/false};
+                      /*lazy=*/false,
+                      /*prefetch_limit=*/0};
 }
 
 CacheOptions CacheOptions::LazyDefaults() {
   return CacheOptions{internal::ReadRangeCache::kDefaultHoleSizeLimit,
                       internal::ReadRangeCache::kDefaultRangeSizeLimit,
-                      /*lazy=*/true};
+                      /*lazy=*/true,
+                      /*prefetch_limit=*/0};
 }
 
 CacheOptions CacheOptions::MakeFromNetworkMetrics(int64_t time_to_first_byte_millis,
@@ -125,7 +127,7 @@ CacheOptions CacheOptions::MakeFromNetworkMetrics(int64_t time_to_first_byte_mil
                                       (1 - ideal_bandwidth_utilization_frac))));
   DCHECK_GT(range_size_limit, 0) << "Computed range_size_limit must be > 0";
 
-  return {hole_size_limit, range_size_limit, false};
+  return {hole_size_limit, range_size_limit, /*lazy=*/false, /*prefetch_limit=*/0};
 }
 
 namespace internal {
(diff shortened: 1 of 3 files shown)


emkornfield:
@jp0317 could you sync to head to see if CI issues go away?

@lidavidm would you mind taking a look?

cpp/src/arrow/io/caching.cc
```diff
@@ -204,6 +206,17 @@ struct ReadRangeCache::Impl {
     if (it != entries.end() && it->range.Contains(range)) {
       auto fut = MaybeRead(&*it);
       ARROW_ASSIGN_OR_RAISE(auto buf, fut.result());
+      if (options.lazy && options.prefetch_limit > 0) {
+        int64_t num_prefetched = 0;
+        for (auto next_it = it + 1;
+             next_it != entries.end() && num_prefetched++ < options.prefetch_limit;
```
lidavidm, line 212: nit: instead of confusing postfix, just increment the counter inside the loop

cpp/src/arrow/io/caching.h
```diff
@@ -43,10 +43,14 @@ struct ARROW_EXPORT CacheOptions {
   int64_t range_size_limit;
   /// \brief A lazy cache does not perform any I/O until requested.
   bool lazy;
+  /// \brief The maximum number of ranges to be prefetched. This is only used
```
lidavidm, line 46: Explain what prefetching means here?

cpp/src/arrow/io/memory_test.cc
```diff
@@ -845,13 +845,61 @@ TEST(RangeReadCache, Lazy) {
...
+}
+
 TEST(CacheOptions, Basics) {
   auto check = [](const CacheOptions actual, const double expected_hole_size_limit_MiB,
                   const double expected_range_size_limit_MiB) -> void {
     const CacheOptions expected = {
         static_cast<int64_t>(std::round(expected_hole_size_limit_MiB * 1024 * 1024)),
         static_cast<int64_t>(std::round(expected_range_size_limit_MiB * 1024 * 1024)),
-        /*lazy=*/false};
+        /*lazy=*/false, /*prefetch_limit=*/0};
```
lidavidm, line 902: For compatibility, we should probably add explicit constructors here now

lidavidm: Ah, I thought you had added the value here because it was necessary, or was it just to be explicit?

lidavidm: I suppose from looking it up again that it would be OK to omit this too, which is all I was wondering about.

lidavidm (approved):
I'm curious about the cases where you've found this useful?

mapleFU:
So, in this patch, when we have io `[0, 1, ...99]`

1. When 0 is readed, the `1-9` will be prefetched
2. When 1 is readed, it would wait for the result being readed, and fetch `10-17`?

Am I right?

PR #36184, opened 2023-06-20
GH-36182: [Gandiva][C++] Fix substring_index function when index is negative.

### LINKED ISSUE #36182: [C++] Bug in substring_index with negative index.

substring_index("Abc.DE.fGh", '.', -2) returns "fGh" but it should return "DE.fGh" (ie starting from the second occurrence of the delimiter from the right). The proposed behavior matches the behavior of other databases.

substring_index was added here: https://github.com/apache/arrow/issues/32374

### Component(s)

C++ - Gandiva

lriggs:
### Rationale for this change

substring_index("Abc.DE.fGh", '.', -2) returns "fGh" but it should return "DE.fGh" (ie starting from the second occurrence of the delimiter from the right). The proposed behavior matches the behavior of other databases.

Fixed reverse index calculation and updated unit tests.

### Are these changes tested?

Yes, unit tests and integration testing.

Function behavior change with substring_index function.

* Closes: #36182

### DIFF
--- a/cpp/src/gandiva/gdv_function_stubs_test.cc
+++ b/cpp/src/gandiva/gdv_function_stubs_test.cc
@@ -464,7 +464,7 @@ TEST(TestGdvFnStubs, TestSubstringIndex) {
   EXPECT_FALSE(ctx.has_error());
 
   out_str = gdv_fn_substring_index(ctx_ptr, "Abc.DE.fGh", 10, ".", 1, -2, &out_len);
-  EXPECT_EQ(std::string(out_str, out_len), "fGh");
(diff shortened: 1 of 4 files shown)


pitrou:
@js8544 Would you like to review this?

js8544:
+1. I checked with other dbs with `substring_index` function, namely MySQL, Spark, SingleStore and Doris. This PR will indeed make Gandiva's behavior consistent with them.

pitrou (approved):
I trust @js8544 's judgement here. Let's merge.

PR #36192, opened 2023-06-20
PARQUET-2316: [C++] Allow partial PreBuffer in the parquet FileReader

jp0317:
### Rationale for this change

The current FileReader can only work in  one of the two modes, coalescing (when Prebuffer is called) and non-coalescing (when Prefufer is not called), due to the if statement [here](https://github.com/apache/arrow/blob/main/cpp/src/parquet/file_reader.cc#L203)

Since Prebuffer is basically caching all specified column chunks, it would raise concerns on memory usage for systems with tight memory budget. In such scenarios, one may want to Prebuffer some small chunks while being able to read the rest chunks using  BufferedInputStream.

Changes to support partial prebuffer on a subset of column chunks and a unit test

### Are these changes tested?

Yes.

No.

### DIFF
--- a/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
+++ b/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
@@ -2382,6 +2382,34 @@ TEST(TestArrowReadWrite, WaitCoalescedReads) {
   ASSERT_EQ(actual_batch->num_rows(), num_rows);
 }
 
+// Use coalesced reads and non-coaleasced reads for different column chunks.
+TEST(TestArrowReadWrite, CoalescedReadsAndNonCoalescedReads) {
+  const int num_columns = 5;
+  const int num_rows = 128;
+
+  std::shared_ptr<Table> expected;
+  ASSERT_NO_FATAL_FAILURE(MakeDoubleTable(num_columns, num_rows, 1, &expected));
+
+  std::shared_ptr<Buffer> buffer;
+  ASSERT_NO_FATAL_FAILURE(WriteTableToBuffer(expected, num_rows / 2,
+                                             default_arrow_writer_properties(), &buffer));
+
+  std::unique_ptr<FileReader> reader;
+  ASSERT_OK_NO_THROW(OpenFile(std::make_shared<BufferReader>(buffer),
+                              ::arrow::default_memory_pool(), &reader));
+
+  ASSERT_EQ(2, reader->num_row_groups());
+
+  // Pre-buffer 3 columns in the 2nd row group.
+  reader->parquet_reader()->PreBuffer({1}, {0, 1, 4}, ::arrow::io::IOContext(),
+                                      ::arrow::io::CacheOptions::Defaults());
+  ASSERT_OK(reader->parquet_reader()->WhenBuffered({1}, {0, 1, 4}).status());
+
+  ASSERT_OK_AND_ASSIGN(auto actual, ReadTableManually(reader.get()));
+
+  AssertTablesEqual(*actual, *expected, /*same_chunk_layout=*/false);
+}
+
 TEST(TestArrowReadWrite, GetRecordBatchReaderNoColumns) {
   ArrowReaderProperties properties = default_arrow_reader_properties();
   const int num_rows = 10;
--- a/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
+++ b/cpp/src/parquet/arrow/arrow_reader_writer_test.cc
@@ -2384,11 +2384,12 @@ TEST(TestArrowReadWrite, WaitCoalescedReads) {
 
 // Use coalesced reads and non-coaleasced reads for different column chunks.
 TEST(TestArrowReadWrite, CoalescedReadsAndNonCoalescedReads) {
-  const int num_columns = 5;
-  const int num_rows = 128;
+  constexpr int num_columns = 5;
+  constexpr int num_rows = 128;
 
   std::shared_ptr<Table> expected;
-  ASSERT_NO_FATAL_FAILURE(MakeDoubleTable(num_columns, num_rows, 1, &expected));
+  ASSERT_NO_FATAL_FAILURE(
+      MakeDoubleTable(num_columns, num_rows, /*nchunks=*/1, &expected));
 
   std::shared_ptr<Buffer> buffer;
   ASSERT_NO_FATAL_FAILURE(WriteTableToBuffer(expected, num_rows / 2,
@@ -2401,9 +2402,12 @@ TEST(TestArrowReadWrite, CoalescedReadsAndNonCoalescedReads) {
   ASSERT_EQ(2, reader->num_row_groups());
 
   // Pre-buffer 3 columns in the 2nd row group.
-  reader->parquet_reader()->PreBuffer({1}, {0, 1, 4}, ::arrow::io::IOContext(),
+  const std::vector<int> row_groups = {1};
+  const std::vector<int> column_indices = {0, 1, 4};
+  reader->parquet_reader()->PreBuffer(row_groups, column_indices,
+                                      ::arrow::io::IOContext(),
                                       ::arrow::io::CacheOptions::Defaults());
-  ASSERT_OK(reader->parquet_reader()->WhenBuffered({1}, {0, 1, 4}).status());
+  ASSERT_OK(reader->parquet_reader()->WhenBuffered(row_groups, column_indices).status());
 
   ASSERT_OK_AND_ASSIGN(auto actual, ReadTableManually(reader.get()));
 
--- a/cpp/src/parquet/file_reader.cc
+++ b/cpp/src/parquet/file_reader.cc
@@ -179,7 +179,7 @@ class SerializedRowGroup : public RowGroupReader::Contents {
                      std::shared_ptr<::arrow::io::internal::ReadRangeCache> cached_source,
                      int64_t source_size, FileMetaData* file_metadata,
                      int row_group_number, const ReaderProperties& props,
-                     const std::unordered_set<int>& prebuffered_column_chunks,
+                     std::unordered_set<int> prebuffered_column_chunks,
                      std::shared_ptr<InternalFileDecryptor> file_decryptor = nullptr)
       : source_(std::move(source)),
         cached_source_(std::move(cached_source)),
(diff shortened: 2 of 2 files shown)


cpp/src/parquet/file_reader.cc
```diff
@@ -268,6 +272,7 @@ class SerializedRowGroup : public RowGroupReader::Contents {
   std::unique_ptr<RowGroupMetaData> row_group_metadata_;
   ReaderProperties properties_;
   int row_group_ordinal_;
+  std::unordered_set<int>& prebuffered_column_chunks_;
```
wgtmac, line 275: IMO, this would be better to use `const std::unordered_set<int>`, without reference.

cpp/src/parquet/file_reader.cc
```diff
@@ -178,13 +179,15 @@ class SerializedRowGroup : public RowGroupReader::Contents {
                      std::shared_ptr<::arrow::io::internal::ReadRangeCache> cached_source,
                      int64_t source_size, FileMetaData* file_metadata,
                      int row_group_number, const ReaderProperties& props,
+                     std::unordered_set<int>& prebuffered_column_chunks,
```
wgtmac, line 182: `const std::unordered_set<int>&` or `std::unordered_set<int>`

wgtmac (changes requested):

wgtmac:
Could you please fix the lint issue here: https://github.com/apache/arrow/actions/runs/5329892746/jobs/9656017234?pr=36192

There is also a test failure, not sure if it is related: https://github.com/apache/arrow/actions/runs/5329892751/jobs/9656017548?pr=36192

```
The following tests FAILED:
Errors while running CTest
	 63 - arrow-dataset-file-parquet-test (Timeout)
Error: Process completed with exit code 8.
```

Also cc @pitrou

wgtmac:
There are some tests failures: 

https://ci.appveyor.com/project/ApacheSoftwareFoundation/arrow/builds/47375345

```
[----------] 32 tests from TestScan/TestParquetFileFormatScan
[ RUN      ] TestScan/TestParquetFileFormatScan.ScanRecordBatchReader/0Threaded16b1024r
unknown file: error: SEH exception with code 0xc0000005 thrown in the test body.
[  FAILED  ] TestScan/TestParquetFileFormatScan.ScanRecordBatchReader/0Threaded16b1024r, where GetParam() = Threaded16b1024r (30 ms)
```

It seems that there are ASAN issues: https://github.com/apache/arrow/actions/runs/5352677942/jobs/9707827704?pr=36192

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -2382,6 +2382,34 @@ TEST(TestArrowReadWrite, WaitCoalescedReads) {
   ASSERT_EQ(actual_batch->num_rows(), num_rows);
 }
 
+// Use coalesced reads and non-coaleasced reads for different column chunks.
+TEST(TestArrowReadWrite, CoalescedReadsAndNonCoalescedReads) {
+  const int num_columns = 5;
+  const int num_rows = 128;
```
wgtmac, line 2388:
```suggestion
  constexpr int num_columns = 5;
  constexpr int num_rows = 128;
```

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -2382,6 +2382,34 @@ TEST(TestArrowReadWrite, WaitCoalescedReads) {
...
+
+  std::shared_ptr<Table> expected;
+  ASSERT_NO_FATAL_FAILURE(MakeDoubleTable(num_columns, num_rows, 1, &expected));
```
wgtmac, line 2391:
```suggestion
  ASSERT_NO_FATAL_FAILURE(MakeDoubleTable(num_columns, num_rows, /*nchunks=*/1, &expected));
```

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -2382,6 +2382,34 @@ TEST(TestArrowReadWrite, WaitCoalescedReads) {
...
+  std::unique_ptr<FileReader> reader;
+  ASSERT_OK_NO_THROW(OpenFile(std::make_shared<BufferReader>(buffer),
+                              ::arrow::default_memory_pool(), &reader));
+
+  ASSERT_EQ(2, reader->num_row_groups());
+
+  // Pre-buffer 3 columns in the 2nd row group.
+  reader->parquet_reader()->PreBuffer({1}, {0, 1, 4}, ::arrow::io::IOContext(),
+                                      ::arrow::io::CacheOptions::Defaults());
+  ASSERT_OK(reader->parquet_reader()->WhenBuffered({1}, {0, 1, 4}).status());
```
wgtmac, line 2406:
```suggestion
  const std::vector<int> row_groups = {1};
  const std::vector<int> column_indices = {0, 1, 4};
  reader->parquet_reader()->PreBuffer(row_groups, column_indices, ::arrow::io::IOContext(),
                                      ::arrow::io::CacheOptions::Defaults());
  ASSERT_OK(reader->parquet_reader()->WhenBuffered(row_groups, column_indices).status());
```

wgtmac (commented):
Thanks for fixing the CI! I can confirm that the only failure is untreated to this patch. I have left some comments in addressing the readability in the test. Rest LGTM

cpp/src/parquet/file_reader.cc
```diff
@@ -351,8 +364,11 @@ class SerializedFile : public ParquetFileReader::Contents {
     cached_source_ =
         std::make_shared<::arrow::io::internal::ReadRangeCache>(source_, ctx, options);
     std::vector<::arrow::io::ReadRange> ranges;
+    prebuffered_column_chunks_.clear();
```
wgtmac, line 367: After double check, I think there could be concurrent issues. If `PreBuffer` and `GetRowGroup` can be called concurrently, `cached_source_` and `prebuffered_column_chunks_` might be in an inconsistent state. For example, `GetRowGroup` reads an old state in `prebuffered_column_chunks_`, then `PreBuffer` creates brand new `cached_source_` and `prebuffered_column_chunks_`. After that, `GetRowGroup` may pass the new `cached_source_` to the RowGroupReader. I believe the followup read will get an error complaining `Status::Invalid("ReadRangeCache did not find matching cache entry")`.

cc @westonpace

wgtmac: That sounds reasonable. Thanks!

wgtmac (approved):

pitrou:
@westonpace @lidavidm Could one of you take a look?

cpp/src/parquet/file_reader.cc
```diff
@@ -178,13 +179,15 @@ class SerializedRowGroup : public RowGroupReader::Contents {
                      std::shared_ptr<::arrow::io::internal::ReadRangeCache> cached_source,
                      int64_t source_size, FileMetaData* file_metadata,
                      int row_group_number, const ReaderProperties& props,
+                     const std::unordered_set<int>& prebuffered_column_chunks,
```
lidavidm, line 182: nit: take this by move and move in the constructor? (so that the caller chooses whether to copy or not)

lidavidm (approved):

cpp/src/parquet/file_reader.cc
```diff
@@ -297,9 +302,17 @@ class SerializedFile : public ParquetFileReader::Contents {
...
+    // only be updated within Prebuffer().
+    auto prebuffered_column_chunks_iter = prebuffered_column_chunks_.find(i);
+    if (prebuffered_column_chunks_iter != prebuffered_column_chunks_.end()) {
+      prebuffered_column_chunks = prebuffered_column_chunks_iter->second;
+    }
+
     std::unique_ptr<SerializedRowGroup> contents = std::make_unique<SerializedRowGroup>(
         source_, cached_source_, source_size_, file_metadata_.get(), i, properties_,
-        file_decryptor_);
+        prebuffered_column_chunks, file_decryptor_);
```
mapleFU, line 315: Here it will copy the `prebuffered_column_chunks` in mapping, would you mind change it to move, and change the argument in constructor to `std::unordered_map`?

mapleFU (approved):

wgtmac (approved):

cpp/src/parquet/file_reader.cc
lidavidm: The move here means that if you read this row group again, it won't make use of the pre-buffered data. Is that intentional? If so, it may be clearer to explicitly erase the map entry after moving as well.

lidavidm: argh, you're right, we copied here

cpp/src/parquet/arrow/arrow_reader_writer_test.cc
```diff
@@ -2382,6 +2382,38 @@ TEST(TestArrowReadWrite, WaitCoalescedReads) {
...
+
+  // Pre-buffer 3 columns in the 2nd row group.
+  const std::vector<int> row_groups = {1};
+  const std::vector<int> column_indices = {0, 1, 4};
+  reader->parquet_reader()->PreBuffer(row_groups, column_indices,
+                                      ::arrow::io::IOContext(),
+                                      ::arrow::io::CacheOptions::Defaults());
+  ASSERT_OK(reader->parquet_reader()->WhenBuffered(row_groups, column_indices).status());
+
+  ASSERT_OK_AND_ASSIGN(auto actual, ReadTableManually(reader.get()));
```
westonpace, line 2412: Does ReadTableManually read the entire table?

westonpace (commented):
I don't see anything wrong with this.  I do think it might be hard to use.  How is a user going to know which row groups or columns they are supposed to prebuffer?  Maybe this is for a situation where a user needs to read the same parquet file over and over again and they can highly tune this operation?

Also, you may want to look into clearing the pre-buffer cache after a row group is read.  That could help start releasing memory as the file is being read.

Would this support a use case allowing us to use prebuffering and still read one row group at a time like:

PreBuffer row group 0
Read row group 0
PreBuffer row group 1
Read row group 1
...

wgtmac:
@westonpace Do you still have any concern?

westonpace:
@wgtmac no concerns.  We can go ahead and merge this one.  Having more options and different ways to read the files is good I think as long as we document how it works and I think this PR is pretty clear.

westonpace (approved):

wgtmac:
Thanks for the reply! @westonpace

PR #36204, opened 2023-06-21
GH-36203: [C++] Support casting in both ways for is_in and index_in

### LINKED ISSUE #36203: [C++] Support casting in both ways for is_in and index_in

This is a follow up of https://github.com/apache/arrow/pull/36058#pullrequestreview-1488682384. Currently it only try to cast the value set to input type, not the other way around. This causes some valid input types to be rejected.

### Component(s)

C++

js8544:
### Rationale for this change

This is a follow up of https://github.com/apache/arrow/pull/36058#pullrequestreview-1488682384. Currently it only try to cast the value set to input type, not the other way around. This causes some valid input types to be rejected.

The kernels will first try to case value_set to input type during preparation. If it doesn't work, it would try to cast input to value_set type before the lookup happens.

### Are these changes tested?

Yes. 

Some previously rejected input types will now be valid.

* Closes: #36203

### DIFF
--- a/cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
@@ -231,9 +231,16 @@ struct InitStateVisitor {
       if (cast_result.ok()) {
         options.value_set = *cast_result;
       } else if (CanCast(*arg_type.type, *options.value_set.type())) {
-        // Will try to cast input array to value set type during kernel exec
+        // Avoid casting from non binary types to string like above
+        // Otherwise, will try to cast input array to value set type during kernel exec
+        if ((options.value_set.type()->id() == Type::STRING ||
+             options.value_set.type()->id() == Type::LARGE_STRING) &&
+            !is_base_binary_like(arg_type.id())) {
+          return Status::Invalid("Array type didn't match type of values set: ",
+                                 *arg_type, " vs ", *options.value_set.type());
+        }
       } else {
-        return Status::Invalid("Input type doesn't match type of values set: ", *arg_type,
+        return Status::Invalid("Array type doesn't match type of values set: ", *arg_type,
                                " vs ", *options.value_set.type());
       }
     }
@@ -277,12 +284,12 @@ struct IndexInVisitor {
   }
 
   template <typename Type>
-  Status ProcessIndexIn(const SetLookupState<Type>& state, const ArraySpan& data) {
+  Status ProcessIndexIn(const SetLookupState<Type>& state, const ArraySpan& input) {
     using T = typename GetViewType<Type>::T;
     FirstTimeBitmapWriter bitmap_writer(out_bitmap, out->offset, out->length);
     int32_t* out_data = out->GetValues<int32_t>(1);
     VisitArraySpanInline<Type>(
-        data,
+        input,
         [&](T v) {
           int32_t index = state.lookup_table->Get(v);
           if (index != -1) {
@@ -386,11 +393,11 @@ struct IsInVisitor {
   }
 
   template <typename Type>
-  Status ProcessIsIn(const SetLookupState<Type>& state, const ArraySpan& data) {
+  Status ProcessIsIn(const SetLookupState<Type>& state, const ArraySpan& input) {
     using T = typename GetViewType<Type>::T;
     FirstTimeBitmapWriter writer(out->buffers[1].data, out->offset, out->length);
     VisitArraySpanInline<Type>(
-        data,
+        input,
         [&](T v) {
           if (state.lookup_table->Get(v) != -1) {
             writer.Set();
--- a/cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
@@ -34,6 +34,7 @@ using internal::HashTraits;
 namespace compute::internal {
 namespace {
 
+// This base class enables non-templated access to the value set type
 struct SetLookupStateBase : public KernelState {
   std::shared_ptr<DataType> value_set_type;
 };
@@ -221,7 +222,6 @@ struct InitStateVisitor {
                              " vs ", *options.value_set.type());
     }
 
-    auto hash_table_type = arg_type;
     if (!options.value_set.is_arraylike()) {
       return Status::Invalid("Set lookup value set must be Array or ChunkedArray");
     } else if (!options.value_set.type()->Equals(*arg_type)) {
@@ -232,14 +232,13 @@ struct InitStateVisitor {
         options.value_set = *cast_result;
       } else if (CanCast(*arg_type.type, *options.value_set.type())) {
         // Will try to cast input array to value set type during kernel exec
-        hash_table_type = options.value_set.type();
       } else {
         return Status::Invalid("Input type doesn't match type of values set: ", *arg_type,
                                " vs ", *options.value_set.type());
       }
     }
 
-    RETURN_NOT_OK(VisitTypeInline(*hash_table_type, this));
+    RETURN_NOT_OK(VisitTypeInline(*options.value_set.type(), this));
     return std::move(result);
   }
 };
@@ -319,9 +318,9 @@ struct IndexInVisitor {
   Status ProcessIndexIn() {
     const auto& state = checked_cast<const SetLookupState<Type>&>(*ctx->state());
     if (!data.type->Equals(state.value_set_type)) {
-      auto materized_input = data.ToArrayData();
+      auto materialized_input = data.ToArrayData();
       ARROW_ASSIGN_OR_RAISE(auto casted_input,
-                            Cast(*materized_input, state.value_set_type,
+                            Cast(*materialized_input, state.value_set_type,
                                  CastOptions::Safe(), ctx->exec_context()));
       return ProcessIndexIn(state, *casted_input.array());
     }
--- a/cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
@@ -31,12 +31,15 @@ namespace arrow {
 using internal::checked_cast;
 using internal::HashTraits;
 
-namespace compute {
-namespace internal {
+namespace compute::internal {
 namespace {
 
+struct SetLookupStateBase : public KernelState {
+  std::shared_ptr<DataType> value_set_type;
+};
+
 template <typename Type>
-struct SetLookupState : public KernelState {
+struct SetLookupState : public SetLookupStateBase {
   explicit SetLookupState(MemoryPool* pool) : memory_pool(pool) {}
 
   Status Init(const SetLookupOptions& options) {
@@ -65,6 +68,7 @@ struct SetLookupState : public KernelState {
     if (!options.skip_nulls && lookup_table->GetNull() >= 0) {
       null_index = memo_index_to_value_index[lookup_table->GetNull()];
     }
+    value_set_type = options.value_set.type();
     return Status::OK();
   }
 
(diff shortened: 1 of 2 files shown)


pitrou:
@js8544 Is this ready for review?

js8544:
> @js8544 Is this ready for review?

Yes, please!

cpp/src/arrow/compute/kernels/scalar_set_lookup_test.cc
```diff
@@ -253,11 +253,11 @@ TEST_F(TestIsInKernel, TimeDuration) {
               "[true, false, false, true, true]", /*skip_nulls=*/true);
   }
 
-  // Different units, invalid cast
-  ASSERT_RAISES(Invalid, IsIn(ArrayFromJSON(duration(TimeUnit::SECOND), "[0, 1, 2]"),
-                              ArrayFromJSON(duration(TimeUnit::MILLI), "[0, 2]")));
+  // Different units, cast value_set to values
+  CheckIsIn(ArrayFromJSON(duration(TimeUnit::SECOND), "[0, 2]"),
+            ArrayFromJSON(duration(TimeUnit::MILLI), "[0, 1, 2000]"), "[true, true]");
```
pitrou, line 258: Ok, but you're choosing an easy case here.
What if values = `[0s, 2s]` and value_set = `[1ms, 2ms, 2000ms]`? Will it fail because of truncation during casting? Will it return an incorrect value? Will it try casting the other way round?

cpp/src/arrow/compute/kernels/scalar_set_lookup_test.cc
```diff
@@ -822,6 +822,37 @@ TEST_F(TestIndexInKernel, Boolean) {
...
+  // Although value_set cannot be cast to int8, but int8 is castable to float
+  CheckIndexIn(input, ArrayFromJSON(float32(), "[1.0, 2.5, 3.1, 5.0]"),
+               "[null, 0, null, null, null, 3, null, null, null]");
+
+  // Allow implicit casts between binary types...
+  CheckIndexIn(ArrayFromJSON(binary(), R"(["aaa", "bbb", "ccc", null, "bbb"])"),
+               ArrayFromJSON(fixed_size_binary(3), R"(["aaa", "bbb"])"),
+               "[0, 1, null, null, 1]");
+  CheckIndexIn(ArrayFromJSON(utf8(), R"(["aaa", "bbb", "ccc", null, "bbb"])"),
+               ArrayFromJSON(large_utf8(), R"(["aaa", "bbb"])"), "[0, 1, null, null, 1]");
```
pitrou, line 844: Can you check the other way round as well for both checks above?

cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
```diff
@@ -215,16 +221,24 @@ struct InitStateVisitor {
...
+      auto cast_result =
           Cast(options.value_set, CastOptions::Safe(arg_type.GetSharedPtr()),
-               ctx->exec_context()));
+               ctx->exec_context());
+      if (cast_result.ok()) {
+        options.value_set = *cast_result;
+      } else if (CanCast(*arg_type.type, *options.value_set.type())) {
+        // Will try to cast input array to value set type during kernel exec
+      } else {
+        return Status::Invalid("Input type doesn't match type of values set: ", *arg_type,
```
pitrou, line 236: Ideally, this should have been `TypeError` instead of `Invalid` here, but let's keep consistent for now.
Perhaps open an issue to change this later?

cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
```diff
@@ -263,11 +277,8 @@ struct IndexInVisitor {
   }
 
   template <typename Type>
-  Status ProcessIndexIn() {
+  Status ProcessIndexIn(const SetLookupState<Type>& state, const ArraySpan& data) {
```
pitrou, line 280: The `data` argument here shadows the `data` member variable, which is confusing and error-prone. Do you need it?

cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
```diff
@@ -359,13 +386,11 @@ struct IsInVisitor {
   }
 
   template <typename Type>
-  Status ProcessIsIn() {
+  Status ProcessIsIn(const SetLookupState<Type>& state, const ArraySpan& data) {
```
pitrou, line 389: Same remark.

pitrou (changes requested):
Thanks a lot! This is a desirable improvement and the approach looks good.

cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
js8544: Created #36345

cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
js8544: Changed `data` to `input`

cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
js8544: Changed `data` to `input`

cpp/src/arrow/compute/kernels/scalar_set_lookup_test.cc
js8544: Updated the test case. Since only safe casts are allowed, it will first try casting [1ms, 2ms, 2000ms] to [0s, 2s] and fail, then cast the other way around and succeed. The result is [null, true] as expected.

cpp/src/arrow/compute/kernels/scalar_set_lookup_test.cc
js8544: Done. I also explicitly forbid casting non-binary to binary, same as the tests down below.

cpp/src/arrow/compute/kernels/scalar_set_lookup_test.cc
```diff
@@ -126,25 +126,40 @@ TEST_F(TestIsInKernel, ImplicitlyCastValueSet) {
...
+  CheckIsIn(input, ArrayFromJSON(float32(), "[1.0, 2.5, 3.1, 5.0]"),
+            "[false, true, false, false, false, true, false, false, false]");
 
   // Allow implicit casts between binary types...
   CheckIsIn(ArrayFromJSON(binary(), R"(["aaa", "bbb", "ccc", null, "bbb"])"),
             ArrayFromJSON(fixed_size_binary(3), R"(["aaa", "bbb"])"),
             "[true, true, false, false, true]");
+  CheckIsIn(ArrayFromJSON(fixed_size_binary(3), R"(["aaa", "bbb", "ccc", null, "bbb"])"),
+            ArrayFromJSON(binary(), R"(["aaa", "bbb"])"),
+            "[true, true, false, false, true]");
```
pitrou, line 139: Can you perhaps make things more interesting by making entries that are not 3-bytes long?
```suggestion
  CheckIsIn(ArrayFromJSON(binary(), R"(["aaa", "bb", "ccc", null, "bbb"])"),
            ArrayFromJSON(fixed_size_binary(3), R"(["aaa", "bbb"])"),
            "[true, false, false, false, true]");
  CheckIsIn(ArrayFromJSON(fixed_size_binary(3), R"(["aaa", "bbb", "ccc", null, "bbb"])"),
            ArrayFromJSON(binary(), R"(["aa", "bbb"])"),
            "[false, true, false, false, true]");
```

cpp/src/arrow/compute/kernels/scalar_set_lookup_test.cc
```diff
@@ -253,11 +268,11 @@ TEST_F(TestIsInKernel, TimeDuration) {
               "[true, false, false, true, true]", /*skip_nulls=*/true);
   }
 
-  // Different units, invalid cast
-  ASSERT_RAISES(Invalid, IsIn(ArrayFromJSON(duration(TimeUnit::SECOND), "[0, 1, 2]"),
-                              ArrayFromJSON(duration(TimeUnit::MILLI), "[0, 2]")));
+  // Different units, cast value_set to values
+  CheckIsIn(ArrayFromJSON(duration(TimeUnit::SECOND), "[0, 2]"),
+            ArrayFromJSON(duration(TimeUnit::MILLI), "[1, 2, 2000]"), "[false, true]");
```
pitrou, line 273: More explicitly ensure that the cast actually happens (either 2 or 2000 could be the successful match...)
```suggestion
  CheckIsIn(ArrayFromJSON(duration(TimeUnit::SECOND), "[0, 1, 2]"),
            ArrayFromJSON(duration(TimeUnit::MILLI), "[1, 2, 2000]"), "[false, false, true]");
```

cpp/src/arrow/compute/kernels/scalar_set_lookup_test.cc
```diff
@@ -253,11 +268,11 @@ TEST_F(TestIsInKernel, TimeDuration) {
               "[true, false, false, true, true]", /*skip_nulls=*/true);
   }
 
-  // Different units, invalid cast
-  ASSERT_RAISES(Invalid, IsIn(ArrayFromJSON(duration(TimeUnit::SECOND), "[0, 1, 2]"),
-                              ArrayFromJSON(duration(TimeUnit::MILLI), "[0, 2]")));
+  // Different units, cast value_set to values
```
pitrou, line 271: Is the comment correct? It will attempt to cast value_set to values, fail because of truncation, and then cast values to value_set.

cpp/src/arrow/compute/kernels/scalar_set_lookup_test.cc
```diff
@@ -779,11 +794,11 @@ TEST_F(TestIndexInKernel, TimeDuration) {
   CheckIndexIn(duration(TimeUnit::SECOND), "[null, null, null, null]", "[null]",
                "[0, 0, 0, 0]");
 
-  // Different units, invalid cast
-  ASSERT_RAISES(Invalid, IndexIn(ArrayFromJSON(duration(TimeUnit::SECOND), "[0, 1, 2]"),
-                                 ArrayFromJSON(duration(TimeUnit::MILLI), "[0, 2]")));
+  // Different units, cast value_set to values
```
pitrou, line 797: Similar comments here :-)

cpp/src/arrow/compute/kernels/scalar_set_lookup_test.cc
```diff
@@ -822,6 +837,50 @@ TEST_F(TestIndexInKernel, Boolean) {
...
+  auto expected = ArrayFromJSON(int32(), ("[null, null, 0, 1, null,"
+                                          "2, null, 3, null]"));
+  AssertArraysEqual(*expected, *out.make_array());
+
+  // Although value_set cannot be cast to int8, but int8 is castable to float
+  CheckIndexIn(input, ArrayFromJSON(float32(), "[1.0, 2.5, 3.1, 5.0]"),
+               "[null, 0, null, null, null, 3, null, null, null]");
+
+  // Allow implicit casts between binary types...
+  CheckIndexIn(ArrayFromJSON(binary(), R"(["aaa", "bbb", "ccc", null, "bbb"])"),
```
pitrou, line 855: Similar suggestions here as for `is_in`.

pitrou (changes requested):
Thanks a lot! Just a few more suggestions to improve the tests.

cpp/src/arrow/compute/kernels/scalar_set_lookup_test.cc
js8544: Done

cpp/src/arrow/compute/kernels/scalar_set_lookup_test.cc
js8544: Done

cpp/src/arrow/compute/kernels/scalar_set_lookup_test.cc
js8544: Comment updated.

cpp/src/arrow/compute/kernels/scalar_set_lookup_test.cc
js8544: Done

cpp/src/arrow/compute/kernels/scalar_set_lookup_test.cc
js8544: Done

pitrou (approved):

PR #36205, opened 2023-06-21
GH-16604: [C++][FlightRPC] Add async Flight client

lidavidm:
### Rationale for this change

Draft an async client API for Flight in C++.

An implementation of DoAction, GetFlightInfo, and DoGet.

The next challenge will be to figure out how to handle writes.

### Are these changes tested?

Sort of.

Adds new APIs.
* Closes: #16604


lidavidm:
CC @pitrou @zeroshade @wjones127 if you were interested in async APIs.

I had to add a small hook to ipc::StreamDecoder. I think it would be preferable to extract a formal "IPC state machine" from ipc/reader.cc and make sure that it's consistent with/used by both StreamDecoder and RecordBatchStreamReader.

TODOs:

- [ ] Figure out how to handle writes (DoPut, DoExchange)
- [ ] Switch to using TransportStatus to report errors (and maybe provide an arrow::Status on the side, or convert it to an error detail; see below)
- [ ] Add RPC error details to TransportStatus (this is part of Flight RPC already, but adding it to TransportStatus will help formalize it)
- [ ] Add tests

Things not handled here:

- [ ] Server APIs
- [ ] Python bindings (though I do want to sketch out how this works with asyncio or AnyIO; it seems community may prefer AnyIO: https://github.com/apache/arrow-adbc/issues/224)
- [ ] Enabling this to work with StopToken (we need StopToken to be able to call a callback, I think)
- [ ] Per-read timeouts (i.e. allow a long/unlimited timeout for a streaming RPC, but stop the RPC if the server does not send a message for a long time)
- [ ] Making the sync API wrap the async API

TODOs.txt
```diff
@@ -0,0 +1,17 @@
+- Use a uniform API for all async Flight client APIs:
+
+  void Foo(const FlightCallOptions&, shared_ptr<ReadListener<T>> listener);
+
+  - shared_ptr, or raw pointer? (gRPC uses the latter...)
```
pitrou, line 5: Or weak_ptr? This way the caller can easily decide the lifetime (like a raw pointer), while not risking crashes (unlike a raw pointer).

cpp/src/arrow/flight/types_async.h
```diff
@@ -0,0 +1,116 @@
...
+///
+/// A single listener may not be used for multiple concurrent RPC
+/// calls.  The application MUST hold the listener alive until
+/// OnFinish() is called and has finished.
+template <typename T>
+class ARROW_FLIGHT_EXPORT AsyncListener : public AsyncListenerBase {
+ public:
+  /// \brief Get the next server result.
+  /// This will never be called concurrently with itself or OnFinish.
+  virtual void OnNext(T message) = 0;
```
pitrou, line 66: If the templating is annoying, perhaps this could be `void OnNext(std::any)`?

pitrou: Also, shouldn't this be able to return a `Status`?

cpp/src/arrow/flight/types_async.h
```diff
@@ -0,0 +1,116 @@
...
+/// \brief Callbacks for DoPut.
+class ARROW_FLIGHT_EXPORT IpcPutter : public AsyncListener<std::unique_ptr<Buffer>> {
+ public:
+  // TODO: which of these can be made const T&?
+
+  /// \brief Begin writing an IPC stream.  May only be called once.
+  ///   Must be called before writing any record batches.
+  void Begin(std::shared_ptr<Schema> schema,
+             ipc::IpcWriteOptions options = ipc::IpcWriteOptions::Defaults());
+  /// \brief Write a record batch.
```
pitrou, line 89: This is an async-write, right?

cpp/src/arrow/flight/types_async.h
```diff
@@ -0,0 +1,116 @@
...
+  /// \param[in] last If true, this is the last write on this stream.
+  ///   This may let the underlying transport optimize the write.
+  ///   Semantically equivalent to calling Write then DoneWriting.
+  void Write(FlightStreamChunk chunk, bool last = false);
+  /// \brief Indicate that the client is done writing.  Must be
+  ///   called, or OnFinish will never be called.
+  void DoneWriting();
+
+  /// \brief Begin or Write finished.  The application may Write() again.
+  virtual void OnWritten() = 0;
```
pitrou, line 108: So the application cannot issue several `Write` calls if the corresponding `OnWritten` notifications haven't been received yet?

cpp/src/arrow/flight/types_async.h
lidavidm: I think it's best to not have multiple error paths. If we return a Status here, what is the implementation supposed to do with it? It can at best just give it straight back to OnFinish.

cpp/src/arrow/flight/types_async.h
lidavidm: Yes, this part is still very much a WIP.

cpp/src/arrow/flight/types.h
```diff
@@ -716,5 +720,130 @@ class ARROW_FLIGHT_EXPORT SimpleResultStream : public ResultStream {
...
+///   like Protocol Buffers for sophisticated API clients.
+class BinaryStatusDetail : public StatusDetail {
+ public:
+  constexpr static const char* kTypeId = "flight::BinaryStatusDetail";
+  explicit BinaryStatusDetail(std::vector<std::byte> details)
+      : details_(std::move(details)) {}
+  const char* type_id() const override { return kTypeId; }
+  std::string ToString() const override;
+
+  const std::vector<std::byte> details() const { return details_; }
```
pitrou, line 822: It seems a bit gratuitous to introduce `std::byte` in our codebase. Why not use `std::string` simply?

cpp/src/arrow/flight/types_async.h
lidavidm: Right, only one Write can be 'active' at any time.

cpp/src/arrow/flight/types.cc
```diff
@@ -771,5 +774,258 @@ arrow::Result<std::string> BasicAuth::SerializeToString() const {
...
+
+TransportStatus TransportStatus::FromCodeStringAndMessage(const std::string& code_str,
+                                                          std::string message) {
+  int code_int = 0;
+  try {
+    code_int = std::stoi(code_str);
+  } catch (...) {
+    return TransportStatus{
+        TransportStatusCode::kUnknown,
+        message + ". Also, server sent unknown or invalid Arrow status code " + code_str};
```
pitrou, line 912: You mean "Flight status code"?

cpp/src/arrow/flight/types.h
```diff
@@ -716,5 +720,130 @@ class ARROW_FLIGHT_EXPORT SimpleResultStream : public ResultStream {
...
 };
 
+/// \defgroup flight-error Error Handling
+/// Types for handling errors from RPCs.  Flight uses a set of status
+/// codes standardized across Flight implementations, so these types
+/// let applications work directly with those codes instead of having
+/// to translate to and from Arrow Status.
+
+/// \brief Abstract status code for an RPC as per the Flight
+///   specification.
```
pitrou, line 730: How come this is different from FlightStatusCode? Do we want to deprecate the latter?

TODOs.txt
```diff
@@ -0,0 +1,17 @@
...
+  - How does this look for writes?
+  - Maybe we always call AddHold and RemoveHold? (It's unclear to me
+    why you'd want to read/write after the stream ends...)
+
+- StopToken needs to be able to run a callback, so that we can cancel gRPC calls.
```
pitrou, line 10: Yes, I think there's an issue already open for this? It should be doable now that we use a helper thread for signal-based cancellation.

The main question is the API: do we want to ensure that callbacks are automatically removed at the end of a scope? Something else?

cpp/src/arrow/flight/types.h
lidavidm: FlightStatusCode was a kludge in the original implementation to add to the arrow::StatusCode list without modifying StatusCode. This was refactored out of transport.h and intends to be definitive. (Both gRPC and UCX backends already translate to it, then the synchronous API translates that into arrow::Status + FlightStatusCode). I would rather make a clean break than try to repurpose or extend FlightStatusCode (which also contains statuses that do not actually exist in Flight on the wire.)

TODOs.txt
lidavidm: I don't think there's a scope that would work, so we would have to explicitly deregister it before calling OnFinish.

lidavidm: Or I suppose if we make it the caller's responsibility, it can be deregistered when the listener itself is cleaned up.

cpp/src/arrow/flight/types_async.h
pitrou: It can decide to automatically cancel the RPC and log the error gracefully.

cpp/src/arrow/flight/types_async.h
pitrou: But why not let the caller buffer writes?

lidavidm: We could do that (the underlying API does not, but that's OK), but then we probably need to consider backpressure.

cpp/src/arrow/flight/types_async.h
lidavidm: I would rather we be explicit; for instance, Arrow C++'s logging doesn't integrate into any application logging framework, which is annoying when you get unexpected, uncontrollable console spam.

cpp/src/arrow/flight/types_async.h
pitrou: Indeed... but backpressure is precisely the kind of concern that the transport is better equipped to deal with, no?

TODOs.txt
lidavidm: Ah. 

So frankly, I don't intend to tackle it here, so I'd rather wait and see how things look once it's wired up. (My thought, though, was that if you provide a StopToken in the CallOptions - which is already possible - we should attach and detach the callback at appropriate times.)

TODOs.txt
lidavidm: We could use non-const reference to express this more strongly? weak_ptr seems rather awkward to use here unless you already have a shared_ptr (in which case, you'd probably rather just pass shared_ptr), but I dislike forcing heap allocation here.

lidavidm:
Async DoPut is implemented, though there is a case not yet handled (and I haven't exposed backpressure yet/and we probably need some mutexes)

TODOs.txt
pitrou: The main challenge IMHO is to ensure proper lifetime management. If you don't pass a managed pointer (either shared or weak), the async machinery might invoke a listener that doesn't exist anymore.

By the way, some cancellation mechanism might also be desired?

cpp/src/arrow/flight/types_async.h
pitrou: The `PauseProducing`/`ResumeProducing` mechanism might be an inspiration.

TODOs.txt
lidavidm: TryCancel is included as a method.

I can switch to forcing a shared pointer again...

cpp/src/arrow/flight/types_async.h
lidavidm: Not here I think (it's not possible to implement with gRPC, and is that really a proven mechanism? Did Acero ever actually implement it or was it only a theoretical API)

pitrou: That API is modeled after [Twisted's](https://docs.twisted.org/en/stable/core/howto/producers.html#push-producers) flow-control mechanism.

Another possibility is to expose a [asyncio-like](https://docs.python.org/3.13/library/asyncio-stream.html#asyncio.StreamWriter.drain) API, for example a `Future<> Drain()` method on the transport.

cpp/src/arrow/flight/types.h
pitrou: We're already using `std::string` for both string and binary data accross the codebase.

cpp/src/arrow/flight/flight_test.cc
```diff
@@ -1019,6 +1027,37 @@ TEST_F(TestFlightClient, DoAction) {
...
+      }
+
+      arrow::Future<std::vector<Result>> future_ =
+          arrow::Future<std::vector<Result>>::Make();
+      std::vector<Result> results_;
+    };
+    // XXX: if you don't wait for rpc to finish, UB occurs because the
+    // subclass part of the listener will be occurred, and so gRPC
+    // will call back into OnFinish but this will be gone it will have
+    // been replaced by "pure virtual" stub
```
pitrou, line 1051: Seems like a good argument to have a shared_ptr somewhere? :-)

cpp/src/arrow/flight/transport.h
```diff
@@ -223,44 +236,30 @@ ARROW_FLIGHT_EXPORT
...
+  virtual void TryCancel() {}
 
-  /// \brief Convert an abstract transport status to a C++ status.
-  Status ToStatus() const;
+  /// Only needed for DoPut/DoExchange
+  virtual void Begin(std::shared_ptr<Schema> schema, ipc::IpcWriteOptions options) {}
+  /// Only needed for DoPut/DoExchange
+  virtual void Write(arrow::flight::FlightStreamChunk chunk, bool last) {}
+  /// Only needed for DoPut/DoExchange
+  virtual void DoneWriting() {}
```
pitrou, line 258: Is there a difference between calling `DoneWriting` and calling `Write(..., last=true)`?

cpp/src/arrow/flight/types_async.h
```diff
@@ -0,0 +1,116 @@
...
+/// \brief Callbacks for results from async RPCs that read Arrow data.
+class ARROW_FLIGHT_EXPORT IpcListener : public AsyncListener<FlightStreamChunk> {
+ public:
+  /// \brief Get the IPC schema.
+  /// This will never be called concurrently with itself, OnNext, or OnFinish.
+  virtual void OnSchema(std::shared_ptr<Schema> schema) = 0;
+};
+
+/// \brief Callbacks for DoPut.
+class ARROW_FLIGHT_EXPORT IpcPutter : public AsyncListener<std::unique_ptr<Buffer>> {
```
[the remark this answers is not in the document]
lidavidm: DoPut returns metadata from the server (the intent was to use this to implement resumable uploads)

cpp/src/arrow/flight/transport.h
lidavidm: No, I'm actually removing this (this is a slight optimization that gRPC implements but I don't think it's worth exposing here)

lidavidm: this being `last`, that is

cpp/src/arrow/flight/types_async.h
lidavidm: I am going to implement queueing, at least.

I think Start/StopProducing could be implemented on top of just knowing the queue depth.

gRPC already has backpressure, anyways: it's "one write" at a time but the write is considered complete once it enters an internal buffer, so effectively gRPC is giving you backpressure by delaying the OnWrite callback.

lidavidm: But Flight-level queueing is also useful to smooth over the fact that one "write" may actually be multiple gRPC-level writes

pitrou: If all transports give you the queue depth, then why not?
But the open question is how the application would know what to do with the queue depth. It's not easy to interpret.

lidavidm: Well, but only having Start/StopProducing means baking in the logic into a slower-moving library instead.  

For queue depth: on each OnWritten callback, the application would check the queue depth and enqueue messages until there's enough work in the buffer. If it wants to be more sophisticated, it can track (or we could expose) an estimate of the bytes enqueued instead. If you really want to be sophisticated, you could estimate bandwidth by also tracking the time between OnWritten callbacks.

For both gRPC and UCX, we'd have to bake some variant of this logic into the library.

pitrou: > For queue depth: on each OnWritten callback, the application would check the queue depth and enqueue messages until there's enough work in the buffer.

1) doesn't the transport have better knowledge to decide whether there is "enough work in the buffer"?

2) if the application decides to postpone the next write, to which deadline does it postpone it? does it fire a timer to check the queue depth, say, 1ms later?

The point of having something like `Drain` or `ResumeProducing` is that the _transport_ informs the application that there is now enough room in the buffer.

lidavidm: 1: well, yes, that's why gRPC doesn't call the OnWritten callback until it's ready for more work!

2: the application would just do the check the next time OnWritten is called (it's called for *each* message sent). So you can think of OnWritten as ResumeProducing, if you want.

pitrou:
For the record, if you added support for a single method (for example DoGet?), it might make reviewing a bit easier?

cpp/src/arrow/flight/types_async.h
pitrou: > the application would just do the check the next time OnWritten is called (it's called for _each_ message sent). So you can think of OnWritten as ResumeProducing, if you want.

Ah! Well, fair enough, then.

lidavidm: I'll try to clarify that (clearly this is confusing/not obvious)

TODOs.txt
lidavidm: Hmm, the problem is then we have a circular reference (and weak_ptr just punts the problem). The caller has state (the callbacks), and the transport has state (the actual RPC), and they both need to reference each other (the callbacks need to update the transport, and the transport needs to invoke the callbacks). 

I suppose we could very carefully drop the weak/shared_ptr holding the callbacks from the RPC state when the RPC finishes, breaking the loop...? But I would rather not create the situation in the first place.

For what it's worth, the gRPC API expects the caller to hold the RPC state alive (that's the grpc::ClientContext).

I suppose a different solution would be to force the transport to hold all the state (e.g. by stashing it in the client object). This makes the client worse for concurrent RPCs (now they're all blocking on that central state, even if only for a short time).

Another might be to change the return type from `void` to some object containing the RPC state. The caller would then hold this alive (and internally it would own the callbacks). You would still have to heap-allocate the callback structure.

lidavidm: I'll draft out the shared_ptr locally (I suppose you can always wrap a reference to a stack allocated object if the allocation matters)

lidavidm:
It turns out queueing writes is quite complicated since gRPC apparently does not wait for your callback to finish before invoking it again, so we are going to have to take some locks...

lidavidm:
Rebased/squashed and converted GetFlightInfo to take shared_ptr. If that looks roughly reasonable I'll split out all other changes and add tests for async GetFlightInfo.

pitrou:
@lidavidm Sounds good to me!

lidavidm:
Split out here: https://github.com/apache/arrow/pull/36517

PR #35921, opened 2023-06-06
GH-34921: [C++][Python][Java] Require CMake 3.16 or later

kou:
### Rationale for this change

We need to support Ubuntu 20.04 (CMake 3.16) or later, CentOS 7 (CMake 3.17) or later, and Debian GNU/Linux bullseye (CMake 3.18) or later.

So we can require CMake 3.16 or later.

### What changes are included in this PR?

Require CMake 3.16 or later.

### Are these changes tested?

Yes.

### Are there any user-facing changes?

Yes.
* Closes: #34921

PR #36206, opened 2023-06-21
GH-36090: [C++] Add testing libraries for Acero & Datasets

### LINKED ISSUE #36090: [C++] Excessive increase in debug build times

Here is a full debug build with Acero disabled (572 build steps), as of changeset 95972cca1f65179b63a9fe5262d1092547e3d137:
```
real	2m22,463s
user	38m39,322s
sys	2m26,218s
```

Here is the same build with Acero enabled (642 build steps):
```
real	2m39,304s
user	45m11,410s
sys	2m51,336s
```

So compiling Acero + tests is worth 7 minutes of CPU time, which seems a bit large.

More generally, Arrow C++ compile times (in debug mode) seem to have increased by 25% - or 7 additional minutes of CPU time - between 11.0.0 and 12.0.0, which is not sustainable:

* debug build on 11.0.0 (552 build steps):
```
real	2m17,473s
user	37m37,749s
sys	2m24,311s
```
* same debug build on 12.0.0 (634 build steps):
```
real	2m35,782s
user	44m12,064s
sys	2m48,723s
```

### Measurements notes

* `ccache` was disabled
* CUDA, Dataset, Flight, Gandiva were disabled
* Compute, GCS, Parquet, Orc, S3 were enabled
* shared libraries only
* using `ld.gold`
* target build directory resides on a [zram](https://en.wikipedia.org/wiki/Zram) filesystem

### Component(s)

C++

westonpace:
### Rationale for this change

The acero & datasets tests rely on some common sources.  The current build includes those in every test executable as sources.  This was a bit of a change introduced as we extracted acero into its own library.  This causes those files to be built again and again and this caused an increase in build times.

This change creates test libraries for each of those modules.  These test libraries include the common sources and are built once.  These libraries can then be included in the tests.  This avoids the duplicate builds.

Changes to how common test sources are built and linked.

### Are these changes tested?

Yes.  These tests still run as part of the unit tests.

I'm not entirely sure if the cmake.in and pc.in files are tested or not as I don't know entirely how they factor into things.  However, I believe these are only needed if users want to rely on these test libraries and I'm not sure how critical that is.

No.

* Closes GH-36090
* Closes: #36090


lidavidm:
If you don't want a 'real' library, you could use a CMake object library, so that the object files for common sources are built once and then linked into each binary.

westonpace:
Kou had suggested this as well. I started by just copy/pasting flight (but must've missed something as CI tests are trying to build it even when testing is off).  I'll look into making it an object library to see if that's simpler.

westonpace:
I assume I can just ignore `ARROW_ACERO_TEST_LINKAGE` for object libraries

westonpace:
Actually, it looks like we are already using object libraries:

```
  if(WIN32
     OR (CMAKE_GENERATOR STREQUAL Xcode)
     OR CMAKE_VERSION VERSION_LESS 3.12
     OR NOT ARROW_POSITION_INDEPENDENT_CODE)
    # We need to compile C++ separately for each library kind (shared and static)
    # because of dllexport declarations on Windows.
    # The Xcode generator doesn't reliably work with Xcode as target names are not
    # guessed correctly.
    # We can't use target for object library with CMake 3.11 or earlier.
    # See also: Object Libraries:
    # https://cmake.org/cmake/help/latest/command/add_library.html#object-libraries
    set(USE_OBJLIB OFF)
  else()
    set(USE_OBJLIB ON)
  endif()

  if(USE_OBJLIB)
    # Generate a single "objlib" from all C++ modules and link
    # that "objlib" into each library kind, to avoid compiling twice
    add_library(${LIB_NAME}_objlib OBJECT ${ARG_SOURCES})
```

I don't know if these requirements (e.g. XCode, WIN32) apply for these testing libraries or not but it seems we can't always use object libraries in all situations.

lidavidm:
I think that sets up an object library that gets linked into a shared and a static library.

But ugh, I forgot about the dllimport/dllexport issues. That said...it shouldn't matter for test code only? There would be no import/export issue since it's all the same binary.

westonpace:
> I think that sets up an object library that gets linked into a shared and a static library.

Ah, ok.

> But ugh, I forgot about the dllimport/dllexport issues. That said...it shouldn't matter for test code only? There would be no import/export issue since it's all the same binary.

That makes sense to me.

lidavidm (approved):

kou (approved):
+1

We need CMake 3.11 or later to use an OBJECT library for `target_link_libraries()` but it'll be done by GH-35921 soon.

kou:
Can we close GH-36090 by only this?

Or do we need more pull requests to close GH-36090? If so, we may want to create an issue for each sub-task.

kou:
We don't need to wait for #35921.

PR #36210, opened 2023-06-21
GH-36177: [MATLAB] Add the Type object hierarchy to the MATLAB interface

### LINKED ISSUE #36177: [MATLAB] Add the `Type` object hierarchy to the MATLAB interface

We should add the `Type` object hierarchy to the MATLAB interface. Eventually, we should add a `Type` property to `arrow.array.Array` class.

### Component(s)

MATLAB

sgilmore10:
### Rationale for this change

As we continue to build up the MATLAB Interface, we need to add the `Type` class hierarchy to the interface to support objects with schemas.

1. Added an enum class called `arrow.type.ID` with the following values:
```matlab
>> enumeration(arrow.type.ID.Boolean)

Enumeration members for class 'arrow.type.ID':

    Boolean
    UInt8
    Int8
    UInt16
    Int16
    UInt32
    Int32
    UInt64
    Int64
    Float32
    Float64
```

2. Added an abstract class representing a type called `arrow.type.Type`. It defines one property named `ID` that all concrete subclasses must define. `ID` must be an `arrow.type.ID`.

3. Defined an abstract class called `arrow.type.PrimitiveType` that inherits from `arrow.type.Type`. All primitive type classes inherit from this class.

4. Defined the following concrete type classes: 
    - `arrow.type.BooleanType`
    - `arrow.type.Float32Type`
    - `arrow.type.Float64Type`
    - `arrow.type.Int8Type`
    - `arrow.type.UInt8Type`
    - `arrow.type.Int16Type`
    - `arrow.type.UInt16Type`
    - `arrow.type.Int32Type`
    - `arrow.type.UInt32Type`
    - `arrow.type.Int64Type`
    - `arrow.type.UInt64Type`

### Are these changes tested?

Yes, added the following test classes: 
    - `tID.m`
    - `tBooleanType.m`
    - `tFloat32Type.m`
    - `tFloat64Type.m`
    - `tInt8Type.m`
    - `tUInt8Type.m`
    - `tInt16Type.m`
    - `tUInt16Type.m`
    - `tInt32Type.m`
    - `tUInt32Type.m`
    - `tInt64Type.m`
    - `tUInt64Type.m`

### Future Directions

1. Add a new property to `arrow.array.Array` called `Type`.
2. Add a `arrow.type.Field` class.
3. Add a `arrow.tabular.Schema` class.
4. As we add more array types, we will extend `arrow.type.ID` and add the appropriate type class. 

* Closes: #36177

### DIFF
--- a/matlab/src/matlab/+arrow/+type/ID.m
+++ b/matlab/src/matlab/+arrow/+type/ID.m
@@ -1,0 +1,52 @@
+% Licensed to the Apache Software Foundation (ASF) under one or more
+% contributor license agreements.  See the NOTICE file distributed with
+% this work for additional information regarding copyright ownership.
+% The ASF licenses this file to you under the Apache License, Version
+% 2.0 (the "License"); you may not use this file except in compliance
+% with the License.  You may obtain a copy of the License at
+%
+%   http://www.apache.org/licenses/LICENSE-2.0
+%
+% Unless required by applicable law or agreed to in writing, software
+% distributed under the License is distributed on an "AS IS" BASIS,
+% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+% implied.  See the License for the specific language governing
+% permissions and limitations under the License.
+
+classdef ID < uint64
+%ID Data type enumeration
+    enumeration
+        Boolean (1)
+        UInt8   (2)
+        Int8    (3)
+        UInt16  (4)
+        Int16   (5)
+        UInt32  (6)
+        Int32   (7)
+        UInt64  (8)
+        Int64   (9)
+        % Float16 (10) not yet supported
+        Float32 (11)
+        Float64 (12)
+    end
+
+    methods
+        function bitWidth = bitWidth(obj)
+            import arrow.type.ID
+            switch obj
+                case ID.Boolean
+                    bitWidth = 1;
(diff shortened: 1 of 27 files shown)


matlab/src/matlab/+arrow/+type/ID.m
```diff
@@ -0,0 +1,52 @@
...
+        Int32   (7)
+        UInt64  (8)
+        Int64   (9)
+        % Float16 (10) not yet supported
+        Float32 (11)
+        Float64 (12)
+    end
+
+    methods
+        function bitWidth = bitWidth(obj)
```
kou, line 34: FYI: Type classes in Apache Arrow C++ also bit width by `bit_width()` method.

kou (approved):
+1

FYI: Apache Arrow C++ also provides type objects: https://github.com/apache/arrow/blob/main/cpp/src/arrow/type.h

matlab/src/matlab/+arrow/+type/ID.m
sgilmore10: Hi @kou,

We debated creating proxy `type` objects versus mirroring them in MATLAB. We decided against creating proxy objects for now, but can always revisit this decision later. From the user's perspective, there would be no difference.

PR #36211, opened 2023-06-21
GH-36209: [Java] Upgrade Netty due to security vulnerability

### LINKED ISSUE #36209: [Java] Upgrade Netty due to CVE

CVE https://github.com/advisories/GHSA-6mjq-h674-j845 fixed in Netty 4.1.94.Final

I don't think this relates to Arrow usage with Netty, but there is an API change that Arrow is not compatible with and needs a patch.

### Component(s)

Java

BryanCutler:
### Rationale for this change

Upgrading Netty dependency due to CVE https://github.com/advisories/GHSA-6mjq-h674-j845
This also requires a patch to arrow-memory

Upgrading Netty, gRPC and Protobuf dependencies

### Are these changes tested?
Existing tests

No

netty-handler SniHandler 16MB allocation

The SniHandler can allocate up to 16MB of heap for each channel during the TLS handshake. When the handler or the channel does not have an idle timeout, it can be used to make a TCP server using the SniHandler to allocate 16MB of heap.
 
https://github.com/advisories/GHSA-6mjq-h674-j845

* Closes: apache/arrow#36209
* GitHub Issue: #36209

### DIFF
--- a/java/memory/memory-netty/src/main/java/io/netty/buffer/PooledByteBufAllocatorL.java
+++ b/java/memory/memory-netty/src/main/java/io/netty/buffer/PooledByteBufAllocatorL.java
@@ -161,7 +161,7 @@ public class PooledByteBufAllocatorL {
     }
 
     private UnsafeDirectLittleEndian newDirectBufferL(int initialCapacity, int maxCapacity) {
-      PoolThreadCache cache = threadCache();
+      PoolArenasCache cache = threadCache();
       PoolArena<ByteBuffer> directArena = cache.directArena;
 
       if (directArena != null) {
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -33,7 +33,7 @@
     <dep.junit.jupiter.version>5.9.0</dep.junit.jupiter.version>
     <dep.slf4j.version>1.7.25</dep.slf4j.version>
     <dep.guava-bom.version>31.1-jre</dep.guava-bom.version>
-    <dep.netty-bom.version>4.1.82.Final</dep.netty-bom.version>
+    <dep.netty-bom.version>4.1.94.Final</dep.netty-bom.version>
     <dep.grpc-bom.version>1.49.1</dep.grpc-bom.version>
     <dep.protobuf-bom.version>3.21.6</dep.protobuf-bom.version>
     <dep.jackson-bom.version>2.15.1</dep.jackson-bom.version>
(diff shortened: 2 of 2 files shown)


BryanCutler:
@lidavidm this is a bit of a problem because Arrow memory needed to be patched to use the fixed version. Is there any discussion of doing a 12.0.2 release that could include this?

lidavidm:
I think 12.0.2 is unlikely, given the 13.0.0 code freeze is a few weeks away, but if you raise it on the ML we can see.

I think this can affect arrow flight, given gRPC uses Netty.

lidavidm (approved):

BryanCutler:
Thanks @lidavidm !

lidavidm:
The code freeze is proposed to be July 10th (https://lists.apache.org/thread/f9r0dsd65ohdtcvc7fnnlfs23n3z0n7f). It would then be one to several weeks to chase down release blockers, prepare binaries, and vote on the release, depending on if any last-minute issues are found.

Generally releases are conducted quarterly.

idelpivnitskiy:
Thanks for response @lidavidm,
I understand your general release cadence, but would like to highlight that this incompatibility does not allow upgrading Netty version for anything else in the same classpath. It will be great for the project if Arrow can find a way to either backport it to older versions (maybe `MethodHandler`s with runtime Netty version check can help), or find a way to stop using **internal** Netty API completely (preferred approach). 

In the meanwhile, are there any known downsides or issues for users switching from `arrow-memory-netty` to `arrow-memory-unsafe`? I couldn't find much information about side-effects. Assuming that they are interchangeable.

lidavidm:
We would love to be more agile with releases. However, it takes quite a bit of maintainer effort. Help is welcome.

@danepitkin was exploring if reflection could help. If you have ideas on how exactly that could work, that may be useful. I'm not familiar with why the code is structured like this, other than performance (but I'm not aware of benchmarks for this).

`arrow-memory-unsafe` has not received as much usage, in my estimation, because `netty` was for a long time the only implementation and is still the 'default' in effect. Also, if your application also uses Netty, the `unsafe` implementation is not aware of memory allocated by Netty and vice versa (may have to tweak the JVM native memory limit).

lidavidm:
I filed https://github.com/apache/arrow/issues/36562 to investigate this in the future. 

(Though you gave me an idea; maybe we could make InnerAllocator here an interface and dispatch between a naive version and an internals-using version at runtime; presumably the JIT could devirtualize the call over time.)

PR #36215, opened 2023-06-21
GH-35760: [C++] C Data Interface helpers should also run checks in non-debug mode

### LINKED ISSUE #35760: [C++] `ArrowArrayStreamRelease` et al should clear the `ArrowArrayStream`

This is supposed to be redundant; the `release` method is supposed to do this itself (as evidenced by the asserts), but this would guard against producers that forget to do so (like https://github.com/apache/arrow-nanoarrow/issues/203) and prevent runtime crashes.

### Component(s)

C++

zeroshade:
Replaced `assert`s in the `Arrow*Release` methods with a macro which outputs to `stderr` and calls `std::abort` directly with a message. This way the check still happens in release mode. The `ARROW_CHECK*` macros require C++ and this is intended to be a C file which others can vendor.

Instead of only failing in debug mode, now these release methods will fail if the release callback fails to set the `release` member to `null` even in release mode. Leading to potential calls to `std::abort` that weren't previously occurring.

* Closes: #35760

### DIFF
--- a/cpp/src/arrow/c/helpers.h
+++ b/cpp/src/arrow/c/helpers.h
@@ -27,7 +27,7 @@
 #define CHECK_OR_DIE(condition, msg)                          \
   do {                                                        \
     if (!(condition)) {                                       \
-      fprintf(stderr, "%s:%d:: %s", __FILE__, __LINE__, msg); \
+      fprintf(stderr, "%s:%d:: %s", __FILE__, __LINE__, (msg)); \
       std::abort();                                           \
     }                                                         \
   } while (0)
--- a/cpp/src/arrow/c/helpers.h
+++ b/cpp/src/arrow/c/helpers.h
@@ -25,7 +25,7 @@
 
 #define ARROW_C_ASSERT(condition, msg)                          \
   do {                                                          \
-    if (!(condition)) {                                         \      
+    if (!(condition)) {                                         \
       fprintf(stderr, "%s:%d:: %s", __FILE__, __LINE__, (msg)); \
       abort();                                                  \
     }                                                           \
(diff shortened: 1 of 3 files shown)


paleolimbot (approved):
This seems reasonable!

cpp/src/arrow/c/helpers.h
```diff
@@ -18,10 +18,20 @@
...
+#include <stdlib.h>
 #include <string.h>
 
 #include "arrow/c/abi.h"
 
+#define CHECK_OR_DIE(condition, msg)                          \
+  do {                                                        \
+    if (!(condition)) {                                       \
+      fprintf(stderr, "%s:%d:: %s", __FILE__, __LINE__, msg); \
+      std::abort();                                           \
```
pitrou, line 31: This needs to be C code, not C++.

cpp/src/arrow/c/helpers.h
```diff
@@ -18,10 +18,20 @@
...
+#include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
 
 #include "arrow/c/abi.h"
 
+#define CHECK_OR_DIE(condition, msg)                          \
+  do {                                                        \
+    if (!(condition)) {                                       \
+      fprintf(stderr, "%s:%d:: %s", __FILE__, __LINE__, msg); \
```
pitrou, line 30: Let's allow `msg` to be complex.
```suggestion
      fprintf(stderr, "%s:%d:: %s", __FILE__, __LINE__, (msg)); \
```

cpp/src/arrow/c/helpers.h
```diff
@@ -18,10 +18,20 @@
 #pragma once
 
 #include <assert.h>
```
pitrou, line 20: Can perhaps remove this inclusion now?

cpp/src/arrow/c/helpers.h
zeroshade: good point, my mistake there at using `std::` it should be just `abort`.

pitrou (approved):

pitrou:
@zeroshade `go/arrow/cdata/arrow/helpers.h` also needs to be updated but it's not a trivial copy of the C++ source anymore.

pitrou:
@zeroshade Can you open an issue for the Go followup?

Issue #36128, opened 2023-06-16
[Compute] ArrowNotImplementedError: Function 'multiply' has no kernel matching input types (duration[s], int32)


randolf-scholz:
### Describe the enhancement requested

```python
import pyarrow as pa

arr = pa.array([1,2,3], type=pa.duration("s"))
mul = pa.array([1,2,3], type=pa.int32())

pa.compute.multiply(arr, [1,2,3])  # ✔ works
pa.compute.multiply(arr, mul)  # ✘ ArrowNotImplementedError: Function 'multiply' has no kernel matching input types (duration[s], int32)
```

Multiplication of `duration`-type with python `int` works, but not with pyarrow's `int32`. It should in priunciple work with any integer type.

### Component(s)

Python

westonpace:
@js8544 recently fixed something very similar.  I don't recall if it would have addressed this issue or not.

js8544:
> 

Not really, that was for the `is_in` kernels, but I'll submit a PR for this issue.

PR #36219, opened 2023-06-21
GH-36218:  [CI][Go] Run benchmark steps only on the main branch

### LINKED ISSUE #36218: [CI][Go] Run benchmark steps only on the main branch

> Is there any reason we are raising an AssertionError here?

Yes, this code is not supposed to execute in GHA for a non-default branch:
https://github.com/apache/arrow/blob/e6221c45de7b4c675bcedfac9afb87f6a131dad9/ci/scripts/go_bench_adapt.py#L46

```
else:
    # Assume that the environment is not GitHub Actions CI. Error out if that
    # assumption seems to be wrong.
    assert os.getenv("GITHUB_ACTIONS") is None

    # This is probably a local dev environment, for testing. In this case, it
    # does usually not make sense to provide commit information (not a
    # controlled CI environment). Explicitly keep `github_commit_info=None` to
    # reflect that (to not send commit information).
```

In that sense I think the failure is expected and does / should not block merge. If the redness is a problem then we can also sys.exit(0) instead. Although that might result in mean false-positive green signal in the future.

_Originally posted by @jgehrcke in https://github.com/apache/arrow/issues/36195#issuecomment-1600599390_

---

Ok, I think the problem is that the PR from dependabot is coming from a branch on the apache/arrow repository instead of a fork hence those are executed which are not if they are from a fork branch due to:
`if: success() && github.event_name == 'push' && github.repository == 'apache/arrow'`
Probably we should not error and just do a `sys.exit(0)` but now I understand why this is failing here.

_Originally posted by @raulcd in https://github.com/apache/arrow/issues/36195#issuecomment-1600624989_

### Component(s)

Continuous Integration, Go

kou:
### Rationale for this change

Because our benchmark tool supports only the main branch.

Add branch name check to the existing `if:`s.

### Are these changes tested?

No.

No.
* Closes: #36218

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -94,7 +94,8 @@ jobs:
         if: >-
           success() &&
           github.event_name == 'push' &&
-          github.repository == 'apache/arrow'
+          github.repository == 'apache/arrow' &&
+          github.ref_name == 'main'
         env:
           ARCHERY_DOCKER_USER: ${{ secrets.DOCKERHUB_USER }}
           ARCHERY_DOCKER_PASSWORD: ${{ secrets.DOCKERHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


westonpace:
Did you mean  #36218 ?  I don't think #36128 is related.

kou:
Oh, sorry. You're right. Fixed.

assignUser (approved):
+1

Issue #35896, opened 2023-06-03
[CI] Refactor ccache steps into reusable action


assignUser:
### Describe the enhancement requested

We have a lot of different workflows that all use ccache in almost identical ways. In keeping with DRY we should refactor this into an action we can call from the different workflows.

### Component(s)

Continuous Integration

thisisnic:
This issue has been marked as stale because it has had no activity in the past 2 years. Please remove the stale label or comment below, or this issue will be closed in 14 days.

PR #36221, opened 2023-06-21
GH-36220: [CI] Run the "Docker Push" step only on the main branch

### LINKED ISSUE #36220: [CI] Run the "Docker Push" step only on the main branch

We want to cache only images that are built on the main branch.

We don't need to cache images that are built on other branches such as branches created by Dependabot.

### Component(s)

Continuous Integration

kou:
### Rationale for this change

We want to cache only images that are built on the main branch.

We don't need to cache images that are built on other branches such as branches created by Dependabot.

Add a branch name check.

### Are these changes tested?

No.

No.
* Closes: #36220

### DIFF
--- a/.github/workflows/cpp.yml
+++ b/.github/workflows/cpp.yml
@@ -114,7 +114,11 @@ jobs:
           ulimit -c unlimited
           archery docker run ${{ matrix.image }}
       - name: Docker Push
-        if: success() && github.event_name == 'push' && github.repository == 'apache/arrow'
+        if: >-
+          success() &&
+          github.event_name == 'push' &&
+          github.repository == 'apache/arrow' &&
+          github.ref_name == 'main'
         env:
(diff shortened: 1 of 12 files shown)


assignUser (approved):
+1

.github/workflows/cpp.yml
```diff
@@ -114,7 +114,11 @@ jobs:
           ulimit -c unlimited
           archery docker run ${{ matrix.image }}
       - name: Docker Push
-        if: success() && github.event_name == 'push' && github.repository == 'apache/arrow'
+        if: >-
+          success() &&
```
assignUser, line 118: We might be able to reduce code duplication a bit by refactoring this into an action (similar to #35896 )

kou: It's a good idea!

raulcd: I've created an issue to track it: https://github.com/apache/arrow/issues/36235

PR #36119, opened 2023-06-16
GH-35879: [C++] Bump bundled google-cloud-cpp to 2.12.0

kou:
### Rationale for this change

The version will fix #35318.

### What changes are included in this PR?

Use the latest released version.

### Are these changes tested?

Yes.

### Are there any user-facing changes?

Yes.
* Closes: #35879

PR #36228, opened 2023-06-22
GH-36227: [C++] New GcsOption to set the project id

### LINKED ISSUE #36227: [C++] Add option to set the default project id in GcsFilesystem

To create a bucket the client library must be configured with a default project. This can be done via the `GOOGLE_CLOUD_CPP` environment variable, or via `google::cloud::storage::ProjectIdOption`.  There is no way to set the option in the `GcsFilesystem` class, which forces applications and test to rely on spooky environment variables.

### Component(s)

C++

coryan:
### Rationale for this change

This fixes #36227, originally motivated by the problems in #36119, but seems like a valuable feature in any case.

- Refactor some code to make it testable.
- Add a new `std::optional<std::string>` field to the `GcsOptions` class.

### Are these changes tested?

Yes, I expanded the unit tests.

Yes. I updated the field documentation.  If I missed some documentation please let me know.

I am also not familiar with the steps required to update the Python wrappers, if there is some documentation to follow I would appreciate it.  I can expand this PR or send a separate one, your call.

* Closes: #36227

### DIFF
--- a/cpp/src/arrow/filesystem/gcsfs.h
+++ b/cpp/src/arrow/filesystem/gcsfs.h
@@ -77,6 +77,13 @@ struct ARROW_EXPORT GcsOptions {
   /// This will be ignored if non-empty metadata is passed to OpenOutputStream.
   std::shared_ptr<const KeyValueMetadata> default_metadata;
 
+  /// \brief The project to use for creating buckets.
+  ///
+  /// If not set, the library uses the GOOGLE_CLOUD_PROJECT environment
+  /// variable. Most I/O operations do not need a project id, only applications
+  /// that create new buckets need a project id.
+  std::optional<std::string> project_id;
+
   bool Equals(const GcsOptions& other) const;
 
   /// \brief Initialize with Google Default Credentials
--- a/cpp/src/arrow/filesystem/gcsfs.h
+++ b/cpp/src/arrow/filesystem/gcsfs.h
@@ -27,9 +27,13 @@
 
 namespace arrow {
 namespace fs {
+namespace internal {
 
 // Opaque wrapper for GCS's library credentials to avoid exposing in Arrow headers.
 struct GcsCredentialsHolder;
+
+}  // namespace internal
+
 class GcsFileSystem;
 
 /// \brief Container for GCS Credentials and information necessary to recreate them.
@@ -41,7 +45,9 @@ class ARROW_EXPORT GcsCredentials {
   TimePoint expiration() const { return expiration_; }
   const std::string& target_service_account() const { return target_service_account_; }
   const std::string& json_credentials() const { return json_credentials_; }
-  const std::shared_ptr<GcsCredentialsHolder>& holder() const { return holder_; }
+  const std::shared_ptr<internal::GcsCredentialsHolder>& holder() const {
+    return holder_;
+  }
 
  private:
   GcsCredentials() = default;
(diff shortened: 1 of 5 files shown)


cpp/src/arrow/filesystem/gcsfs_internal.h
```diff
@@ -53,6 +64,8 @@ ARROW_EXPORT Result<std::shared_ptr<const KeyValueMetadata>> FromObjectMetadata(
 
 ARROW_EXPORT std::int64_t Depth(std::string_view path);
 
+google::cloud::Options AsGoogleCloudOptions(const GcsOptions& options);
```
kou, line 67:
```suggestion
ARROW_EXPORT google::cloud::Options AsGoogleCloudOptions(const GcsOptions& options);
```

cpp/src/arrow/filesystem/gcsfs_internal.h
```diff
@@ -30,6 +32,15 @@
 
 namespace arrow {
 namespace fs {
+struct GcsCredentialsHolder {
```
kou, line 35:
```suggestion
struct ARROW_EXPORT GcsCredentialsHolder {
```

kou (approved):
+1

Thanks!

cpp/src/arrow/filesystem/gcsfs_internal.h
coryan: Done.

cpp/src/arrow/filesystem/gcsfs_internal.h
coryan: I am not sure we want to do this.  That symbol was not exported before, and was not part of the public API:

https://github.com/apache/arrow/blob/cd6b2b5dca468cee6704ade4ce539c0a09874607/cpp/src/arrow/filesystem/gcsfs.cc#L38-L43

arguably it should live in `arrow::fs::internal`.  Thoughts?

kou: > it should live in `arrow::fs::internal`. 

Ah, yes. I agree with it.

coryan: Done

kou: Thanks!

cpp/src/arrow/filesystem/gcsfs.h
```diff
@@ -77,6 +83,13 @@ struct ARROW_EXPORT GcsOptions {
   /// This will be ignored if non-empty metadata is passed to OpenOutputStream.
   std::shared_ptr<const KeyValueMetadata> default_metadata;
 
+  /// \brief The project to use for creating buckets.
+  ///
+  /// If not set, the library uses the GOOGLE_CLOUD_PROJECT environment
+  /// variable. Most I/O operations do not need a project id, only applications
+  /// that create new buckets need a project id.
+  std::optional<std::string> project_id;
```
pitrou, line 91: FTR, an optional was probably not necessary as I suppose an empty string cannot possibly be a valid project id?

coryan: You are probably right.  There is a difference between "not set" (which implies "use the `GOOGLE_CLOUD_PROJECT` environment variable") and "set to empty" (which implies "ignore the environment variable and use the empty string"). Is this distinction useful?  Arguably not.

I am working on the Python fixes today, I can change this to `std::string` if you think it is best to do so.

pitrou: It's probably ok to keep the optional.
There's already a PR for Python here: https://github.com/apache/arrow/pull/36376

raulcd: hey @coryan I gave the Python PR a try to try and fix some of the CI failures. I am trying to fix some of our nightly builds to try and get to a better state before the code freeze. Sorry for stepping in :)

coryan: No worries at all, I am sure your fixes will be better.  And I am getting very efficient at "did nothing" 😄

PR #36231, opened 2023-06-22
GH-36128: [C++][Compute] Allow multiplication between duration and all integer types

### LINKED ISSUE #36128: [Compute] ArrowNotImplementedError: Function 'multiply' has no kernel matching input types (duration[s], int32)

js8544:
### Rationale for this change

Currently durations can only be multiplied with int64, but no other integer types.

Allow duration types to be multiplied with any integer type.
1. For `multiply`, new kernels are added to support the new types.
2. For `multiply_checked`, integers will be casted to int64. We can't add new kernels because the MultiplyChecked op class requires both operands to have the same type.

### Are these changes tested?

Yes.

No.
* Closes: #36128

### DIFF
--- a/cpp/src/arrow/compute/kernels/scalar_arithmetic.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_arithmetic.cc
@@ -618,6 +618,21 @@ ArrayKernelExec GenerateArithmeticWithFixedIntOutType(detail::GetTypeId get_id) 
   }
 }
 
+void PromoteIntegerForDurationArithmetic(std::vector<TypeHolder>* types) {
+  bool has_duration = std::any_of(types->begin(), types->end(), [](const TypeHolder& t) {
+    return t.id() == Type::DURATION;
+  });
+
+  if (!has_duration) return;
+
+  // Require implicit casts to int64 to match duration's bit width
+  for (auto& type : *types) {
+    if (is_integer(type.id())) {
+      type = int64();
+    }
+  }
+}
+
 struct ArithmeticFunction : ScalarFunction {
   using ScalarFunction::ScalarFunction;
 
@@ -643,21 +658,9 @@ struct ArithmeticFunction : ScalarFunction {
         }
       }
 
-      if (name_ == "multiply_checked") {
-        // MultiplyChecked requires the same type for both operands, therefore we can't
-        // easily make kernels for integers other than int64, so we just cast integers to
-        // int64
-        if ((*types)[0].id() == Type::DURATION && is_integer((*types)[1].id())) {
-          ReplaceTypes(int64(), &(*types)[1], 1);
-        } else if ((*types)[1].id() == Type::DURATION && is_integer((*types)[0].id())) {
-          ReplaceTypes(int64(), &(*types)[0], 1);
-        }
-      } else if (name_ == "divide" || name_ == "divide_checked") {
-        // Similar to MultipyChecked, Divide and DivideChecked require both operands to
-        // have the same type
-        if ((*types)[0].id() == Type::DURATION && is_integer((*types)[1].id())) {
-          ReplaceTypes(int64(), &(*types)[1], 1);
-        }
+      if (name_ == "multiply" || name_ == "multiply_checked" || name_ == "divide" ||
+          name_ == "divide_checked") {
+        PromoteIntegerForDurationArithmetic(types);
       }
     }
 
@@ -1488,19 +1491,9 @@ void RegisterScalarArithmetic(FunctionRegistry* registry) {
 
   // Add multiply(duration, int64) -> duration
   for (auto unit : TimeUnit::values()) {
-    for (auto numeric : NumericTypes()) {
-      if (!is_integer(numeric->id())) continue;
-
-      auto exec =
-          GenerateInteger<ScalarBinary, Int64Type, Int64Type, Multiply>(numeric->id());
-      DCHECK_OK(multiply->AddKernel({numeric, duration(unit)}, duration(unit), exec));
-
-      auto reverse_exec =
-          GenerateInteger<ScalarBinaryReverse, Int64Type, Int64Type, Multiply>(
-              numeric->id());
-      DCHECK_OK(
-          multiply->AddKernel({duration(unit), numeric}, duration(unit), reverse_exec));
-    }
+    auto exec = ArithmeticExecFromOp<ScalarBinaryEqualTypes, Multiply>(Type::DURATION);
+    DCHECK_OK(multiply->AddKernel({duration(unit), int64()}, duration(unit), exec));
+    DCHECK_OK(multiply->AddKernel({int64(), duration(unit)}, duration(unit), exec));
   }
 
   DCHECK_OK(registry->AddFunction(std::move(multiply)));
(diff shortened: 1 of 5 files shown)


pitrou:
The CI failure on AMD64 Conda C++ is unrelated.

cpp/src/arrow/compute/kernels/scalar_arithmetic.cc
```diff
@@ -640,6 +642,17 @@ struct ArithmeticFunction : ScalarFunction {
...
+      if (name_ == "multiply_checked") {
+        // MultiplyChecked requires the same type for both operands, therefore we can't
+        // easily make kernels for integers other than int64, so we just cast integers to
+        // int64
+        if ((*types)[0].id() == Type::DURATION && is_integer((*types)[1].id())) {
+          ReplaceTypes(int64(), &(*types)[1], 1);
+        } else if ((*types)[1].id() == Type::DURATION && is_integer((*types)[0].id())) {
+          ReplaceTypes(int64(), &(*types)[0], 1);
+        }
+      }
```
pitrou, line 655: Are there other places where a similar pattern may be desired? Should this be factored out somehow?
cc @lidavidm @bkietz  for opinions.

cpp/src/arrow/compute/kernels/scalar_arithmetic_test.cc
```diff
@@ -1152,6 +1152,53 @@ TEST(TestBinaryArithmetic, AddWithImplicitCastsUint64EdgeCase) {
                                      ArrayFromJSON(uint64(), "[18446744073709551615]")}));
 }
 
+TEST(TestBinaryArithmetic, MultiplyDuration) {
```
pitrou, line 1155: We already have `TestTemporalMultiplyDuration` in `scalar_temporal_test.cc`, can you augment that test instead?

Also, notice `TestTemporalDivideDuration` which hints that the same improvement could be done for division...

cpp/src/arrow/compute/kernels/scalar_arithmetic.cc
```diff
@@ -640,6 +642,17 @@ struct ArithmeticFunction : ScalarFunction {
           ReplaceTypes(type, types);
         }
       }
+
+      if (name_ == "multiply_checked") {
```
pitrou, line 646: As pointed out below, duration also supports dividing by an integer.

cpp/src/arrow/compute/kernels/scalar_arithmetic_test.cc
js8544: Oh, I didn't notice this. I've changed the tests.

pitrou (approved):
LGTM. I'll wait for @lidavidm and @bkietz 's opinion on the `DispatchBest` implementation.

cpp/src/arrow/compute/kernels/scalar_arithmetic.cc
lidavidm: I don't remember any, but it's been quite a while since I worked with Acero.

lidavidm: I think we can leave it as-is.

bkietz: Personally, I'd prefer to see a uniform approach to adding this support to `multiply` and `multiply_checked`: either both should have kernels matching each type or both should just use the numeric promotion. (Adding kernels to multiply_checked would require adding Call overloads to the MultiplyChecked op, but is doable.) Of those two options, I'd prefer to see promotion for consistency; multiply(int8, int64) requires an implicit cast to int64 so it seems odd to specifically add direct support for multiply(int8, duration) but not for any of the other arithmetic kernels.

More directly in answer to your question, I would also like to see this logic extracted to a helper function. It may only be used in this DispatchBest, but just in case another Function requires this logic it'd be best to try to avoid copypasta. Furthermore it'll keep this DispatchBest from getting too unreadable. Something like:

```c++
if (name_ == "multiply" || name_ == "multiply_checked" ||
    name_ == "divide" || name_ == "divide_checked") {
  PromoteIntegerForDurationArithmetic(types);
}

void PromoteIntegerForDurationArithmetic(std::vector<TypeHolder>* types) {
  bool has_duration = false;
  for (const auto& type : *types) {
    if (type.id() == Type::DURATION) {
      has_duration = true;
    }
  }
  if (!has_duration) return;

  // Require implicit casts to int64 to match duration's bit width
  for (auto& type : *types) {
    if (is_integer(type.id())) {
      type = int64();
    }
  }
}
```

It should be fine to let this function do the replacement `(int8, duration) -> (int64, duration)` even for divide; this just defers the failed kernel lookup to the following DispatchExact call.

bkietz (changes requested):

cpp/src/arrow/compute/kernels/scalar_arithmetic.cc
js8544: It does make more sense considering that muliply uses type promotion already. I've updated the code.

cpp/src/arrow/compute/kernels/scalar_arithmetic.cc
```diff
@@ -616,6 +617,21 @@ ArrayKernelExec GenerateArithmeticWithFixedIntOutType(detail::GetTypeId get_id)
   }
 }
 
+void PromoteIntegerForDurationArithmetic(std::vector<TypeHolder>* types) {
+  bool has_duration = std::any_of(types->begin(), types->end(), [](const TypeHolder& t) {
```
bkietz, line 621: :rocket:

bkietz (changes requested):

cpp/src/arrow/compute/kernels/scalar_arithmetic.cc
```diff
@@ -616,6 +617,21 @@ ArrayKernelExec GenerateArithmeticWithFixedIntOutType(detail::GetTypeId get_id)
   }
 }
 
+void PromoteIntegerForDurationArithmetic(std::vector<TypeHolder>* types) {
```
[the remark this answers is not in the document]
js8544: My bad. I've moved it over.

bkietz (approved):

js8544:
CI failures are unrelated. @pitrou @bkietz Is this PR good to merge?

cpp/src/arrow/compute/kernels/scalar_arithmetic.cc
pitrou: Looks much cleaner now, thanks!

pitrou (approved):

Issue #36000, opened 2023-06-08
[Python][C++] Simplify/Validate PyObject refcount in udf.cc


icexelloss:
### Describe the bug, including details regarding any error messages, version, and platform.

In udf.cc, there is interaction between C++ and Python where the C++ code holds a user defined Python object inside a Kernel and invoke it over input data. 

Due to this interaction, there are a number of `Py_INCREF` call as well as using `OwnRef` to manage the refcount for the PyObject (which represents the user defined function. The current code for managing it is either over complicated and/or bug where the C++ code might be over increasing the refcount. While this is not a huge issue (it probably cause the Python function to never be garbage collected), it could use some clean up / documentation as least to improve maintainability. 

Appendix:
* Related thread:  https://github.com/apache/arrow/pull/35514#discussion_r1205081272

### Component(s)

C++

PR #36253, opened 2023-06-22
GH-36252: [Python] Add non decomposable hash aggregate UDF 

### LINKED ISSUE #36252: [C++][Python] Non decomposable aggregation UDF (Hash version)

In https://github.com/apache/arrow/issues/35515,

I have implemented a Scalar version of the non decomposable UDF (Scalar as in SCALAR_AGGREGATE). I would like to support the Hash version of it (Hash as in HASH_AGGREGATE)

### Component(s)

C++

icexelloss:
### Rationale for this change

In https://github.com/apache/arrow/issues/35515,

I have implemented a Scalar version of the non decomposable UDF (Scalar as in SCALAR_AGGREGATE). I would like to support the Hash version of it (Hash as in HASH_AGGREGATE)

With this PR, user can register an aggregate UDF once with `pc.register_aggregate_function` and it can be used as both scalar aggregate function and hash aggregate function.

Example:

```
def median(x):
    return pa.scalar(np.nanmedian(x))

pc.register_aggregate_function(func=median, func_name='median_udf', ...)

table = ...
table.groupby("id").aggregate([("v", 'median_udf')])
```

The main changes are:
* In ResigterAggregateFunction (udf.cc), we now register the function both as a scalar aggregate function and a hash aggregate function (with signature adjustment for hash aggregate kernel because we need to append the grouping key)
* Implemented PythonUdfHashAggregateImpl, similar to the PythonUdfScalarAggregateImpl. In Consume, it will accumulate both the input batches and the group id array. In Merge, it will merge the input batches and group id array (with the group_id_mapping). In Finalize, it will apply groupings to the accumulated batches to create one record batch per group, then apply the UDF over each group.
* Some code clean up - `UdfWrapperCallback` objects are named `cb` (previously, `agg_cb` or `wrapper`) now and the user defined python function is now just called `function` (previously `agg_function`) 

For table.groupby().aggregate(...), the space complexity is O(n) where n is the size of the table (and therefore, is not very useful). However, this is more useful in the segmented aggregation case, where the space complexity of O(s), where s the size of the segments.

### Are these changes tested?
Added new test in test_udf.py (with table.group_by().aggregate() and test_substrait.py (with segmented aggregation)

Yes with this change, user can call use registered aggregate UDF with `table.group_by().aggregate() ` or Acero's segmented aggregation.

### Checklist

- [x] Self Review
- [x] API Documentation

* Closes: #36252

### DIFF
--- a/cpp/src/arrow/compute/row/grouper.h
+++ b/cpp/src/arrow/compute/row/grouper.h
@@ -149,7 +149,7 @@ class ARROW_EXPORT Grouper {
   ///       []
   ///   ]
   static Result<std::shared_ptr<ListArray>> MakeGroupings(
-      const UInt32Array& ids, uint32_t num_groups,
+      const UInt32Array& ids, uint32_t  ,
       ExecContext* ctx = default_exec_context());
 
   /// \brief Produce a ListArray whose slots are selections of `array` which correspond to
--- a/cpp/src/arrow/compute/row/grouper.h
+++ b/cpp/src/arrow/compute/row/grouper.h
@@ -149,8 +149,7 @@ class ARROW_EXPORT Grouper {
   ///       []
   ///   ]
   static Result<std::shared_ptr<ListArray>> MakeGroupings(
-      const UInt32Array& ids, uint32_t  ,
-      ExecContext* ctx = default_exec_context());
+      const UInt32Array& ids, uint32_t, ExecContext* ctx = default_exec_context());
 
   /// \brief Produce a ListArray whose slots are selections of `array` which correspond to
   /// the provided groupings.
--- a/cpp/src/arrow/compute/row/grouper.h
+++ b/cpp/src/arrow/compute/row/grouper.h
@@ -149,7 +149,8 @@ class ARROW_EXPORT Grouper {
   ///       []
   ///   ]
   static Result<std::shared_ptr<ListArray>> MakeGroupings(
-      const UInt32Array& ids, uint32_t, ExecContext* ctx = default_exec_context());
+      const UInt32Array& ids, uint32_t num_groups,
+      ExecContext* ctx = default_exec_context());
 
   /// \brief Produce a ListArray whose slots are selections of `array` which correspond to
   /// the provided groupings.
--- a/python/pyarrow/_compute.pyx
+++ b/python/pyarrow/_compute.pyx
@@ -2824,7 +2824,7 @@ def register_aggregate_function(func, function_name, function_doc, in_types, out
     >>> answer
     <pyarrow.DoubleScalar: 30.0>
     >>> table = pa.table([pa.array([1, 1, 2, 2]), pa.array([10, 20, 30, 40])], names=['k', 'v'])
-    >>> result = table.group_by('k').aggregate(['v', 'py_compute_median'])
+    >>> result = table.group_by('k').aggregate([('v', 'py_compute_median')])
     >>> result
     pyarrow.Table
     k: int64
--- a/python/pyarrow/_compute.pyx
+++ b/python/pyarrow/_compute.pyx
@@ -2767,6 +2767,9 @@ def register_aggregate_function(func, function_name, function_doc, in_types, out
     This is often used with ordered or segmented aggregation where groups
     can be emit before accumulating all of the input data.
 
+    Note that currently size of any input column can not exceed 2 GB limit
+    (all groups combined).
+
     Parameters
     ----------
     func : callable
--- a/python/pyarrow/_compute.pyx
+++ b/python/pyarrow/_compute.pyx
@@ -2767,8 +2767,8 @@ def register_aggregate_function(func, function_name, function_doc, in_types, out
     This is often used with ordered or segmented aggregation where groups
     can be emit before accumulating all of the input data.
 
-    Note that currently size of any input column can not exceed 2 GB limit
-    (all groups combined).
+    Note that currently the size of any input column can not exceed 2 GB
+    for a single segment (all groups combined).
 
     Parameters
     ----------
--- a/python/pyarrow/_compute.pyx
+++ b/python/pyarrow/_compute.pyx
@@ -2823,6 +2823,15 @@ def register_aggregate_function(func, function_name, function_doc, in_types, out
     >>> answer = pc.call_function(func_name, [pa.array([20, 40])])
     >>> answer
     <pyarrow.DoubleScalar: 30.0>
+    >>> table = pa.table([pa.array([1, 1, 2, 2]), pa.array([10, 20, 30, 40])], names=['k', 'v'])
+    >>> result = table.group_by('k').aggregate(['v', 'py_compute_median'])
+    >>> result
+    pyarrow.Table
+    k: int64
+    v_py_compute_median: double
+    ----
+    k: [[1,2]]
+    v_py_compute_median: [[15,35]]
     """
     return _register_user_defined_function(get_register_aggregate_function(),
                                            func, function_name, function_doc, in_types,
--- a/python/pyarrow/conftest.py
+++ b/python/pyarrow/conftest.py
@@ -289,7 +289,7 @@ def unary_agg_func_fixture():
     """
     from pyarrow import compute as pc
 
-    def func(ctx, x, *args):
+    def func(ctx, x):
         return pa.scalar(np.nanmean(x))
 
     func_name = "mean_udf"
--- a/python/pyarrow/conftest.py
+++ b/python/pyarrow/conftest.py
@@ -20,6 +20,8 @@ import pyarrow as pa
 from pyarrow import Codec
 from pyarrow import fs
 
+import numpy as np
+
 groups = [
     'acero',
     'brotli',
@@ -283,10 +285,9 @@ def unary_func_fixture():
 @pytest.fixture(scope="session")
 def unary_agg_func_fixture():
     """
-    Register a unary aggregate function
+    Register a unary aggregate function (mean)
     """
     from pyarrow import compute as pc
-    import numpy as np
 
     def func(ctx, x, *args):
         return pa.scalar(np.nanmean(x))
@@ -312,7 +313,6 @@ def varargs_agg_func_fixture():
     Register a unary aggregate function
     """
     from pyarrow import compute as pc
-    import numpy as np
 
     def func(ctx, *args):
         sum = 0.0
@@ -320,7 +320,7 @@ def varargs_agg_func_fixture():
             sum += np.nanmean(arg)
         return pa.scalar(sum)
 
-    func_name = "y=sum_mean(x...)"
+    func_name = "sum_mean"
     func_doc = {"summary": "Varargs aggregate",
                 "description": "Varargs aggregate"}
 
--- a/python/pyarrow/conftest.py
+++ b/python/pyarrow/conftest.py
@@ -288,13 +288,14 @@ def unary_agg_func_fixture():
     from pyarrow import compute as pc
     import numpy as np
 
-    def func(ctx, x):
+    def func(ctx, x, *args):
         return pa.scalar(np.nanmean(x))
 
-    func_name = "y=avg(x)"
+    func_name = "mean_udf"
     func_doc = {"summary": "y=avg(x)",
                 "description": "find mean of x"}
 
+    breakpoint()
     pc.register_aggregate_function(func,
                                    func_name,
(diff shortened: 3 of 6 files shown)


cpp/src/arrow/compute/row/grouper.h
```diff
@@ -149,8 +149,7 @@ class ARROW_EXPORT Grouper {
   ///       []
   ///   ]
   static Result<std::shared_ptr<ListArray>> MakeGroupings(
-      const UInt32Array& ids, uint32_t num_groups,
-      ExecContext* ctx = default_exec_context());
+      const UInt32Array& ids, uint32_t, ExecContext* ctx = default_exec_context());
```
icexelloss, line 152: Revert this change - accidental

python/pyarrow/conftest.py
```diff
@@ -283,15 +285,14 @@ def unary_function(ctx, x):
...
 def unary_agg_func_fixture():
     """
-    Register a unary aggregate function
+    Register a unary aggregate function (mean)
     """
     from pyarrow import compute as pc
-    import numpy as np
 
-    def func(ctx, x):
+    def func(ctx, x, *args):
```
icexelloss, line 292: Revert this

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -355,18 +536,10 @@ Status RegisterTabularFunction(PyObject* user_function, UdfWrapperCallback wrapp
       wrapper, options, registry);
 }
 
-Status AddAggKernel(std::shared_ptr<compute::KernelSignature> sig,
```
icexelloss, line 358: This is inlined now.

icexelloss:
@westonpace I would like a request a review on this PR. The code should be relatively straight forward and similar to https://github.com/apache/arrow/pull/35514 so hopefully no confusion/surprises here. 

For the implementation of the grouping, I used an approach similar to `GroupedListImpl` aggregator and `partition.cc`

For the registration, I decided to register both scalar/hash kernel with one API `register_aggregate_function` because I think scalar/hash difference is not really something user should be worry about (from user's point of view, it is just "aggregation", whether it is hash/scalar is an compute implementation detail).

Let me know if those sounds OK to you.

westonpace:
@icexelloss I should have some time to take a look tomorrow

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -15,16 +15,28 @@
 // specific language governing permissions and limitations
 // under the License.
 
-#include "arrow/python/udf.h"
+#include <iostream>
```
westonpace, line 18:
```suggestion
```

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -215,9 +249,162 @@ struct PythonUdfScalarAggregatorImpl : public ScalarUdfAggregator {
...
+
+struct PythonUdfHashAggregatorImpl : public HashUdfAggregator {
+  PythonUdfHashAggregatorImpl(std::shared_ptr<OwnedRefNoGIL> function,
+                              UdfWrapperCallback cb,
+                              std::vector<std::shared_ptr<DataType>> input_types,
+                              std::shared_ptr<DataType> output_type)
+      : function(function), cb(std::move(cb)), output_type(std::move(output_type)) {
+    Py_INCREF(function->obj());
+    std::vector<std::shared_ptr<Field>> fields;
+    for (size_t i = 0; i < input_types.size(); i++) {
```
westonpace, line 267:
```suggestion
    std::vector<std::shared_ptr<Field>> fields;
    fields.reserve(input_types.size());
    for (size_t i = 0; i < input_types.size(); i++) {
```

Minor nit

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -215,9 +249,162 @@ struct PythonUdfScalarAggregatorImpl : public ScalarUdfAggregator {
...
+  };
+
+  ~PythonUdfHashAggregatorImpl() override {
+    if (_Py_IsFinalizing()) {
+      function->detach();
+    }
+  }
+
+  /// @brief Same as ApplyGrouping in parition.cc
+  /// Replicated the code here to avoid complicating the dependencies
```
westonpace, line 280:
```suggestion
  // same as ApplyGrouping in parition.cc
  // replicated the code here to avoid complicating the dependencies
```
I'm not sure how valuable this comment is in general though.

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -215,9 +249,162 @@ struct PythonUdfScalarAggregatorImpl : public ScalarUdfAggregator {
...
+    ARROW_ASSIGN_OR_RAISE(
+        std::shared_ptr<RecordBatch> rb,
+        batch.ToExecBatch().ToRecordBatch(input_schema, ctx->memory_pool()));
+
+    // This is similar to GroupedListImpl
+    // last array is the group id
+    const ArraySpan& groups_array_data = batch[batch.num_values() - 1].array;
+    DCHECK_EQ(groups_array_data.offset, 0);
+    int64_t batch_num_values = groups_array_data.length;
+    const auto* batch_groups = groups_array_data.GetValues<uint32_t>(1, 0);
```
westonpace, line 313: Why not just `groups_array_data.GetValues<uint32_t>(1);`?

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -215,9 +249,162 @@ struct PythonUdfScalarAggregatorImpl : public ScalarUdfAggregator {
...
+    auto g = group_id_mapping.GetValues<uint32_t>(1);
+    for (uint32_t other_g = 0; static_cast<int64_t>(other_g) < other.num_values;
+         ++other_g) {
+      // Different state can have different group_id mappings, so we
+      // need to translate the ids
+      RETURN_NOT_OK(groups.Append(g[other_raw_groups[other_g]]));
+    }
+
+    num_values += other.num_values;
+    return Status::OK();
```
westonpace, line 337: Does `num_groups` need to be updated here?

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -215,9 +249,162 @@ struct PythonUdfScalarAggregatorImpl : public ScalarUdfAggregator {
...
+                          Grouper::MakeGroupings(UInt32Array(num_values, groups_buffer),
+                                                 static_cast<uint32_t>(num_groups)));
+
+    ARROW_ASSIGN_OR_RAISE(auto table,
+                          arrow::Table::FromRecordBatches(input_schema, values));
+    ARROW_ASSIGN_OR_RAISE(auto rb, table->CombineChunksToBatch(ctx->memory_pool()));
+    UdfContext udf_context{ctx->memory_pool(), table->num_rows()};
+
+    if (rb->num_rows() == 0) {
+      return Status::Invalid("Finalized is called with empty inputs");
```
westonpace, line 355: Why is this a problem?

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -215,9 +249,162 @@ struct PythonUdfScalarAggregatorImpl : public ScalarUdfAggregator {
...
+    const int num_args = input_schema->num_fields() - 1;
+
+    ARROW_ASSIGN_OR_RAISE(auto groups_buffer, groups.Finish());
+    ARROW_ASSIGN_OR_RAISE(auto groupings,
+                          Grouper::MakeGroupings(UInt32Array(num_values, groups_buffer),
+                                                 static_cast<uint32_t>(num_groups)));
+
+    ARROW_ASSIGN_OR_RAISE(auto table,
+                          arrow::Table::FromRecordBatches(input_schema, values));
+    ARROW_ASSIGN_OR_RAISE(auto rb, table->CombineChunksToBatch(ctx->memory_pool()));
```
westonpace, line 351: There are some cases where this won't be possible.  For example, if you have an array of strings then the array may only have 2GB of string data (regardless of how many elements it has).  So any single group can't have more than 2GB of string data.  I don't know this is fatal but you may want to mention in the user docs somewhere or wrap this failure with extra context.

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -392,24 +569,109 @@ Status RegisterAggregateFunction(PyObject* agg_function, UdfWrapperCallback agg_
...
+  compute::ScalarAggregateKernel kernel(std::move(sig), std::move(init),
+                                        AggregateUdfConsume, AggregateUdfMerge,
+                                        AggregateUdfFinalize, /*ordered=*/false);
+  RETURN_NOT_OK(aggregate_func->AddKernel(std::move(kernel)));
   RETURN_NOT_OK(registry->AddFunction(std::move(aggregate_func)));
   return Status::OK();
 }
 
+/// @brief Create a new UdfOptions with adjustment for hash kernel
+/// @param options User provided udf options
```
westonpace, line 591:
```suggestion
/// \brief Create a new UdfOptions with adjustment for hash kernel
/// \param options User provided udf options
```

westonpace (changes requested):
This is a good set of tests. It's nice and convenient that the same python implementation can work for both.  I have a few minor suggestions / questions.  I think the only concerning thing is that we need to restrict group sizes to things that will fit in a single batch.  Do you think this will be a problem for your use cases?

icexelloss:
> I think the only concerning thing is that we need to restrict group sizes to things that will fit in a single batch. Do you think this will be a problem for your use cases?

Thanks @westonpace. Currently we only plan to use this with segmented aggregation so each group is not going to be very large. (grouping inside a segment), so I don't think it would be a problem.

python/pyarrow/src/arrow/python/udf.cc
icexelloss: Good catch - updated

python/pyarrow/src/arrow/python/udf.cc
icexelloss: Updated

python/pyarrow/src/arrow/python/udf.cc
icexelloss: Added note in the user doc

icexelloss:
@westonpace This should be clean now - another look?

icexelloss:
Gentle ping @westonpace anything else  you want me to change here?

python/pyarrow/_compute.pyx
```diff
@@ -2767,6 +2767,9 @@ def register_aggregate_function(func, function_name, function_doc, in_types, out
     This is often used with ordered or segmented aggregation where groups
     can be emit before accumulating all of the input data.
 
+    Note that currently size of any input column can not exceed 2 GB limit
+    (all groups combined).
```
westonpace, line 2771:
```suggestion
    Note that currently the size of any input column can not exceed 2 GB
    for a single segment (all groups combined).
```

python/pyarrow/src/arrow/python/udf.cc
```diff
@@ -215,9 +247,164 @@ struct PythonUdfScalarAggregatorImpl : public ScalarUdfAggregator {
...
+  std::shared_ptr<DataType> output_type;
+};
+
+struct PythonUdfHashAggregatorImpl : public HashUdfAggregator {
+  PythonUdfHashAggregatorImpl(std::shared_ptr<OwnedRefNoGIL> function,
+                              UdfWrapperCallback cb,
+                              std::vector<std::shared_ptr<DataType>> input_types,
+                              std::shared_ptr<DataType> output_type)
+      : function(function), cb(std::move(cb)), output_type(std::move(output_type)) {
+    Py_INCREF(function->obj());
```
westonpace, line 263: These `INCREF`'s still seem superfluous to me but I don't think it's critical.  We could test in a follow-up using temporary function registries to see if we are preventing UDF functions from being garbage collected.

westonpace (approved):
Minor wording suggestion.  Otherwise this looks good.

python/pyarrow/src/arrow/python/udf.cc
icexelloss: I agree with you. I plan to address this in https://github.com/apache/arrow/issues/36000 but haven't got to it.

icexelloss:
Thanks @westonpace. I applied your suggestion and will merge once CI passes.

icexelloss:
CI failure is unrelated. Merging.

PR #36256, opened 2023-06-22
GH-36255: [C++] Add benchmarks for "if_else" kernel on lists

### LINKED ISSUE #36255: [C++] Add benchmarks for "if_else" kernel on lists

"if_else" kernels kernels with list input can have their perf improved at the cost of code specialization. I'm working on this problem and would like to introduce benchmarks that measure how it behaves currently.

### Component(s)

C++

felipecrv:
### Rationale for this change

"if_else" kernels kernels with list input can have their perf improved at the cost of code specialization. I'm working on this problem and would like to introduce benchmarks that measure how it behaves currently.

Only the new benchmarks and addition of `items_per_second` metric on the others.

### Are these changes tested?

By running the benchmarks.

No.
* Closes: #36255

### DIFF
--- a/cpp/src/arrow/compute/kernels/scalar_if_else_benchmark.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_if_else_benchmark.cc
@@ -57,9 +57,49 @@ struct GetBytesProcessed<Type, enable_if_base_binary<Type>> {
   }
 };
 
+template <typename DataType>
+struct GetBytesProcessed<DataType, enable_if_list_view_like<DataType>> {
+  static int64_t Get(const std::shared_ptr<Array>& arr) {
+    using OffsetType = typename TypeTraits<DataType>::OffsetType::c_type;
+
+    int64_t total_bytes = 0;
+    if constexpr (DataType::type_id == Type::LIST ||
+                  DataType::type_id == Type::LARGE_LIST) {
+      total_bytes += (arr->length() + 1) * sizeof(OffsetType);
+    } else if constexpr (DataType::type_id == Type::LIST_VIEW) {
+      // ListView has an extra buffer for the sizes and it uses the same type as the
+      // offsets buffer.
+      total_bytes += 2 * arr->length() * sizeof(OffsetType);
+    } else {
+      ADD_FAILURE() << "GetBytesProcessed dispatching not implemented for "
+                    << arr->type()->ToString();
+    }
+
+    auto child_array = std::static_pointer_cast<ListArray>(arr)->values();
+    switch (child_array->type_id()) {
+      case Type::UINT32:
+        total_bytes += GetBytesProcessed<UInt32Type>::Get(child_array);
+        break;
+      case Type::UINT64:
+        total_bytes += GetBytesProcessed<UInt64Type>::Get(child_array);
+        break;
+      case Type::STRING:
+        total_bytes += GetBytesProcessed<StringType>::Get(child_array);
+        break;
+      case Type::LARGE_STRING:
+        total_bytes += GetBytesProcessed<LargeStringType>::Get(child_array);
+        break;
+      default:
+        ADD_FAILURE() << "GetBytesProcessed dispatching not implemented for "
+                      << child_array->type()->ToString();
+        break;
+    }
+    return total_bytes;
+  }
+};
+
 template <typename Type>
-static void IfElseBench(benchmark::State& state) {
-  auto type = TypeTraits<Type>::type_singleton();
+static void IfElseBench(benchmark::State& state, const std::shared_ptr<DataType>& type) {
   using ArrayType = typename TypeTraits<Type>::ArrayType;
 
   int64_t len = state.range(0);
@@ -85,11 +125,25 @@ static void IfElseBench(benchmark::State& state) {
                           (GetBytesProcessed<BooleanType>::Get(cond) +
                            GetBytesProcessed<Type>::Get(left) +
                            GetBytesProcessed<Type>::Get(right)));
+  state.SetItemsProcessed(state.iterations() * (len - offset));
 }
 
 template <typename Type>
-static void IfElseBenchContiguous(benchmark::State& state) {
+static void IfElseBench(benchmark::State& state) {
   auto type = TypeTraits<Type>::type_singleton();
+  return IfElseBench<Type>(state, type);
+}
+
+template <typename ListType, typename ValueType>
+static void IfElseBenchList(benchmark::State& state) {
+  auto value_type = TypeTraits<ValueType>::type_singleton();
+  auto list_type = std::make_shared<ListType>(value_type);
+  return IfElseBench<ListType>(state, list_type);
+}
+
+template <typename Type>
+static void IfElseBenchContiguous(benchmark::State& state,
+                                  const std::shared_ptr<DataType>& type) {
   using ArrayType = typename TypeTraits<Type>::ArrayType;
 
   int64_t len = state.range(0);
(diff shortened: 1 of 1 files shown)


felipecrv:
@pitrou

cpp/src/arrow/compute/kernels/scalar_if_else_benchmark.cc
```diff
@@ -57,9 +57,45 @@ struct GetBytesProcessed<Type, enable_if_base_binary<Type>> {
...
+struct GetBytesProcessed<DataType, enable_if_list_like<DataType>> {
+  static int64_t Get(const std::shared_ptr<Array>& arr) {
+    using OffsetType = typename TypeTraits<DataType>::OffsetType::c_type;
+
+    int64_t total_bytes = 0;
+    if constexpr (DataType::type_id == Type::LIST ||
+                  DataType::type_id == Type::LARGE_LIST) {
+      total_bytes += (arr->length() + 1) * sizeof(OffsetType);
+    } else {
+      ADD_FAILURE() << "GetBytesProcessed dispatching not implemented for "
```
pitrou, line 70: One alternative is (you need to include `arrow/util/logging.h` if not already done):
```suggestion
      ARROW_LOG(FATAL) << "GetBytesProcessed dispatching not implemented for "
```

cpp/src/arrow/compute/kernels/scalar_if_else_benchmark.cc
```diff
@@ -57,9 +57,45 @@ struct GetBytesProcessed<Type, enable_if_base_binary<Type>> {
...
+        total_bytes += GetBytesProcessed<LargeStringType>::Get(child_array);
+        break;
+      default:
+        ADD_FAILURE() << "GetBytesProcessed dispatching not implemented for "
+                      << child_array->type()->ToString();
+        break;
+    }
+    return total_bytes;
+  }
+};
```
pitrou, line 95: Note: we may also not bother with the complexity; if things are getting out of hand, just return 0 from `GetBytesProcessed` for list-like types.

cpp/src/arrow/compute/kernels/scalar_if_else_benchmark.cc
```diff
@@ -57,9 +57,45 @@ struct GetBytesProcessed<Type, enable_if_base_binary<Type>> {
...
+    if constexpr (DataType::type_id == Type::LIST ||
+                  DataType::type_id == Type::LARGE_LIST) {
+      total_bytes += (arr->length() + 1) * sizeof(OffsetType);
+    } else {
+      ADD_FAILURE() << "GetBytesProcessed dispatching not implemented for "
+                    << arr->type()->ToString();
+    }
+
+    auto child_array = std::static_pointer_cast<ListArray>(arr)->values();
+    switch (child_array->type_id()) {
```
pitrou, line 75: May want to use `VisitTypeInline` to avoid rewriting the switch by hand.

cpp/src/arrow/compute/kernels/scalar_if_else_benchmark.cc
```diff
@@ -85,11 +121,25 @@ static void IfElseBench(benchmark::State& state) {
                           (GetBytesProcessed<BooleanType>::Get(cond) +
                            GetBytesProcessed<Type>::Get(left) +
                            GetBytesProcessed<Type>::Get(right)));
+  state.SetItemsProcessed(state.iterations() * (len - offset));
```
pitrou, line 124: How about we also make the parameters more readable?
```suggestion
  state.SetItemsProcessed(state.iterations() * (len - offset));
  state.counters["offset"] = static_cast<double>(offset);
```

cpp/src/arrow/compute/kernels/scalar_if_else_benchmark.cc
```diff
@@ -403,36 +499,45 @@ static void ChooseBench64(benchmark::State& state) {
...
 
+// IfElse: Lists
+BENCHMARK(IfElseBenchListUInt32)->Args({kNumItems, 0});
+BENCHMARK(IfElseBenchListUInt64)->Args({kNumItems, 0});
+BENCHMARK(IfElseBenchListString32)->Args({kNumItems, 0});
+BENCHMARK(IfElseBenchListString64)->Args({kNumItems, 0});
+BENCHMARK(IfElseBenchListUInt32Contiguous)->Args({kNumItems, 0});
+BENCHMARK(IfElseBenchListUInt64Contiguous)->Args({kNumItems, 0});
+BENCHMARK(IfElseBenchListString32Contiguous)->Args({kNumItems, 0});
+BENCHMARK(IfElseBenchListString64Contiguous)->Args({kNumItems, 0});
```
pitrou, line 520: Perhaps we want to cut down on the number of benchmarks? I don't think we need to benchmark both 32- and 64-bit ints, or normal and large strings.

pitrou (commented):
LGTM on the principle. Just a couple minor suggestions.

cpp/src/arrow/compute/kernels/scalar_if_else_benchmark.cc
felipecrv: Done. Confirmed this works even in Release builds.

cpp/src/arrow/compute/kernels/scalar_if_else_benchmark.cc
felipecrv: I think the size of the data is important here to differentiate between list array containing lists of small and big sizes.

cpp/src/arrow/compute/kernels/scalar_if_else_benchmark.cc
felipecrv: I rewrote all the GetBytesProcessed overloads using a visitor now.

cpp/src/arrow/compute/kernels/scalar_if_else_benchmark.cc
```diff
@@ -24,42 +24,79 @@
...
-struct GetBytesProcessed<BooleanType> {
-  static int64_t Get(const std::shared_ptr<Array>& arr) { return arr->length() / 8; }
-};
+  explicit GetBytesProcessedVisitor(const Array* arr) : arr(arr) {}
 
-template <typename Type>
-struct GetBytesProcessed<Type, enable_if_number<Type>> {
-  static int64_t Get(const std::shared_ptr<Array>& arr) {
+  Status RecurseInto(const Array* child_arr) {
+    // return VisitTypeInline(*child_arr->type(), this);
```
pitrou, line 42: Did you mean to re-enable this or just to remove this function? `RecurseInto` is currently a no-op...

cpp/src/arrow/compute/kernels/scalar_if_else_benchmark.cc
```diff
@@ -24,42 +24,79 @@
...
-    return arr->length() * sizeof(OffsetType) +
-           std::static_pointer_cast<ArrayType>(arr)->total_values_length();
+    total_bytes += arr->length() * sizeof(OffsetType) +
+                   internal::checked_cast<const ArrayType*>(arr)->total_values_length();
+    return Status::OK();
+  }
+
+  template <typename ArrowType>
+  std::enable_if_t<is_list_like_type<ArrowType>::value &&
+                       !is_fixed_size_list_type<ArrowType>::value,
```
pitrou, line 69: I think you can simply use `enable_if_var_size_list<ArrowType, Status>`

cpp/src/arrow/compute/kernels/scalar_if_else_benchmark.cc
```diff
@@ -24,42 +24,79 @@
...
+    using ArrayType = typename TypeTraits<ArrowType>::ArrayType;
+    using OffsetType = typename TypeTraits<ArrowType>::OffsetType::c_type;
+
+    if constexpr (ArrowType::type_id == Type::LIST ||
+                  ArrowType::type_id == Type::LARGE_LIST) {
+      total_bytes += (arr->length() + 1) * sizeof(OffsetType);
+      auto child_array = internal::checked_cast<const ArrayType*>(arr)->values();
+      return RecurseInto(child_array.get());
+    } else {
+      return Unimplemented(type);
```
pitrou, line 81: `MAP` should actually work fine as well.

cpp/src/arrow/compute/kernels/scalar_if_else_benchmark.cc
felipecrv: My bad. I commented this to reduce the template error noise when compiling and forgot to re-enable.

cpp/src/arrow/compute/kernels/scalar_if_else_benchmark.cc
felipecrv: Indeed. :sweat_smile: Fixing.

pitrou (approved):

PR #36265, opened 2023-06-23
GH-36264: [R] Add scalar() function 

### LINKED ISSUE #36264: [R] Add scalar() function

We have functions for many other data container creation but not for `Scalar$create()`.

See also #20396 for a follow-up

### Component(s)

R

thisisnic:
Fixes #36264 36264
* Closes: #36264


thisisnic:
I feel like there may be a reason we haven't done this previously, though am unsure?

paleolimbot (approved):
Definitely worth having! Perhaps it didn't happen because we also have `Expression$scalar()` which is what all the bindings use.

It's true that `scalar()` is a tiny bit generic...the closest thing I found was `rlang::is_scalar_xx()` and `assertthat::is.scalar()` (which helpfully returns `TRUE` for an `arrow::Scalar`. I think it's fine! There's always `arrow_scalar()` if you're worried.

thisisnic:
> It's true that `scalar()` is a tiny bit generic...the closest thing I found was `rlang::is_scalar_xx()` and `assertthat::is.scalar()` (which helpfully returns `TRUE` for an `arrow::Scalar`. I think it's fine! There's always `arrow_scalar()` if you're worried.

Thanks for looking for those!  If you're happy with this as-is, I'm happy too.  Will merge once CI is green.

PR #36266, opened 2023-06-23
GH-35901: [C++][Python] pyarrow.csv.write_csv crashes when writing tables containing FixedSizeBinaryArray

### LINKED ISSUE #35901: [Python] pyarrow.csv.write_csv crashes when writing tables containing FixedSizeBinaryArray

Suppose we have a large array of binary strings
```
import numpy as np
import pyarrow as pa
import pyarrow.csv as pcsv

nparr = np.frombuffer(np.random.randint(65, 91, int(4E8), 'u1'), 'S4')
```
We want to construct a FixedSizedBinaryArray and Table
```
fixedarr = pa.array(nparr, pa.binary(4))
fixedtable = pa.Table.from_arrays([fixedarr], names=['fixedsize'])
```

Alternatively, construct a BinaryArray and Table
```
binarr = pa.array(nparr, pa.binary())
bintable = pa.Table.from_arrays([binarr], names=['binary'])
```
Setting up the csv write option
```
csvoption = pcsv.WriteOptions(include_header=False, batch_size=2048, delimiter='|', quoting_style='none')
```
Writing the table containing BinaryArray works
```
pcsv.write_csv(bintable, 'binary.csv', write_options=csvoption)
```

But writing the table containing FixedSizedBinaryArray crashes after first `batch_size`
```
pcsv.write_csv(fixedtable, 'fixedsize.csv', write_options=csvoption)
```
Note both tables are compatible with `pyarrow.parquet.write_table` though.

**Environment**: 
Windows 10
pyarrow 12.0.0
Python 3.10.10

### Component(s)

Python

vibhatha:
### Rationale for this change

Crash occurs when writing a CSV file using `pyarrow.csv.write_csv` with a `FixedSizedBinaryArray`. And the crash is sensitive to the datasize. And this PR fixes this [issue](https://github.com/apache/arrow/issues/35901).

This PR includes a minor change to a the logic in `scalar_cast_string.cc`.

### Are these changes tested?

I have added a Python test to evaluate the user's issue and replicated the bug.
And the test case I have added test it parametricly. Also I tested existing C++
tets.

Issue filed by a user: https://github.com/apache/arrow/issues/35901

* Closes: #35901


felipecrv:
Is this still a draft or ready for review?

felipecrv (approved):

westonpace (changes requested):
Just one small concern about the test.  Thanks for the fix.

python/pyarrow/tests/test_csv.py
```diff
@@ -1968,3 +1968,27 @@ def test_write_csv_decimal(tmpdir, type_factory):
...
+
+    csvoption = WriteOptions(include_header=True, batch_size=2048,
+                             delimiter='|', quoting_style='none')
+
+    write_csv(fixed_table, file_name, write_options=csvoption)
+
+    assert os.path.exists(file_name)
+
+    res_table = read_csv(file_name)
+    assert fixed_table.shape, res_table.shape
```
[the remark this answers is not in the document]
vibhatha: Updated the code. Thanks for catching this one.

python/pyarrow/tests/test_csv.py
```diff
@@ -1968,3 +1968,28 @@ def test_write_csv_decimal(tmpdir, type_factory):
...
+    int(1E2),
+    int(1E4),
+    int(1E6)
+))
+def test_large_binary_write_to_csv(tmpdir, data_size):
+    file_name = tmpdir / "fixedsize_"+str(data_size)+".csv"
+
+    nparr = np.frombuffer(np.random.randint(65, 91, data_size, 'u1'), 'S4')
+
+    fixed_arr = pa.array(nparr, pa.string())
```
westonpace, line 1983:
```suggestion
    fixed_arr = pa.array(nparr, pa.binary(4))
```
This test, as written, doesn't actually reproduce the error.  Changing this to be fixed size binary does trigger the error (assuming you compile without your fix).  You will need to update the comparison though because fixed size binary will be read back as string without any other options.  So you will also need to specify the column type (using convert_options) when reading from CSV.

westonpace (changes requested):
I believe the fix is correct.  We need to change the test slightly to make sure it is regressing the underlying problem.

python/pyarrow/tests/test_csv.py
vibhatha: You're correct, I have missed that point. Test cases have been updated.

westonpace (changes requested):
Looks like the test is failing (I think you might need to combine chunks on the result before comparing).  Also, looks like there is a minor lint issue.

python/pyarrow/tests/test_csv.py
```diff
@@ -1972,6 +1972,38 @@ def test_write_csv_decimal(tmpdir, type_factory):
...
+    parse_options = ParseOptions(delimiter="|")
+    convert_options = ConvertOptions(column_types={"fixedsize": pa.binary(4)})
+    read_options = ReadOptions(block_size=2048)
+
+    res_table = read_csv(file_name, parse_options=parse_options,
+                         convert_options=convert_options,
+                         read_options=read_options)
+    res_table = res_table.combine_chunks()
+
+    assert res_table.column(0).chunks[0] == fixed_table.column(0).chunks[0]
```
AlenkaF, line 2004:
```suggestion
    if not res_table.equals(fixed_table):
        # Better error output
        assert res_table.to_pydict() == fixed_table.to_pydict()
```

python/pyarrow/tests/test_csv.py
```diff
@@ -1972,6 +1972,41 @@ def test_write_csv_decimal(tmpdir, type_factory):
...
+    read_options = ReadOptions(block_size=2048)
+
+    res_table = read_csv(file_name, parse_options=parse_options,
+                         convert_options=convert_options,
+                         read_options=read_options)
+    res_table = res_table.combine_chunks()
+
+    if not res_table.equals(fixed_table):
+        assert assert res_table.schema == fixed_table.schema
+        assert assert res_table.num_rows == fixed_table.num_rows
```
AlenkaF, line 2006:
```suggestion
        assert res_table.schema == fixed_table.schema
        assert res_table.num_rows == fixed_table.num_rows
```

AlenkaF:
Hm, now the Python 3.9 build is passing and 3.8 is failing. I am creating 3.9 env locally to see what I can find, hope to stop the suggest change-commit spam =)

AlenkaF:
I have found out the cause for the random flakey failures in the CI.

The code in the tests is generating a numpy array with random strings of length 4. They can also include the string `"NULL"` which currently gets converted to `None`.

<details>
<summary>Reproducible example</summary>

```python
import pyarrow as pa
import numpy as np
import os
from pyarrow.csv import (read_csv, write_csv,
                         ConvertOptions, ParseOptions,
                         ReadOptions, WriteOptions)

file_name = "fixedsize_NULL"
nparr = np.array([b'NULL'])
fixed_arr = pa.array(nparr, pa.binary(4))
fixed_table = pa.Table.from_arrays([fixed_arr], names=['fixedsize'])

write_options = WriteOptions(include_header=True, batch_size=2048,
                                delimiter='|')
write_csv(fixed_table, file_name, write_options=write_options)

assert os.path.exists(file_name)

parse_options = ParseOptions(delimiter="|")
convert_options = ConvertOptions(column_types={"fixedsize": pa.binary(4)})
read_options = ReadOptions(block_size=2048)

res_table = read_csv(file_name, parse_options=parse_options,
                        convert_options=convert_options,
                        read_options=read_options)
```

```python
>>> nparr
array([b'NULL'], dtype='|S4')
>>> fixed_table
pyarrow.Table
fixedsize: fixed_size_binary[4]
----
fixedsize: [[4E554C4C]]
>>> res_table
pyarrow.Table
fixedsize: fixed_size_binary[4]
----
fixedsize: [[null]]
```
</details>

We should use `convert_options = ConvertOptions(column_types={"fixedsize": pa.binary(4)}, quoted_strings_can_be_null=False)` so that `"NULL"` is not converted to `None`.

Will add the change via suggestions.

vibhatha:
> I have found out the cause for the random flakey failures in the CI.
> 
> The code in the tests is generating a numpy array with random strings of length 4. They can also include the string `"NULL"` which currently gets converted to `None`.
> Reproducible example
> 
> We should use `convert_options = ConvertOptions(column_types={"fixedsize": pa.binary(4)}, quoted_strings_can_be_null=False)` so that `"NULL"` is not converted to `None`.
> 
> Will add the change via suggestions.

Oh Wow! Awesome catch @AlenkaF :pray: 
Thanks for working on this one.

AlenkaF:
I wasn't able to get the right indent though so I expect, at least, the linter error 😬 😁

pitrou:
Please explicitly describe the rationale in the PR description. "This PR fixes an already filed issue which has been evaluated as a bug" is not an appropriate description.

cpp/src/arrow/compute/kernels/scalar_cast_string.cc
```diff
@@ -338,10 +338,12 @@ BinaryToBinaryCastExec(KernelContext* ctx, const ExecSpan& batch, ExecResult* ou
...
-    ARROW_ASSIGN_OR_RAISE(
-        output->buffers[0],
-        arrow::internal::CopyBitmap(ctx->memory_pool(), input.buffers[0].data,
-                                    input.offset, input.length));
+    if (input.buffers[0].data != nullptr) {
+      ARROW_ASSIGN_OR_RAISE(
+          output->buffers[0],
+          arrow::internal::CopyBitmap(ctx->memory_pool(), input.buffers[0].data,
+                                      input.offset, input.length));
+    }
```
pitrou, line 346: This seems to fix a bug in the C++ kernels, so, could you add a test in this area as well?
There are existing tests in https://github.com/apache/arrow/blob/f98a13250d10dba248a2bb85989d6b80265e82d8/cpp/src/arrow/compute/kernels/scalar_cast_test.cc#L2059 which you could expand.

python/pyarrow/tests/test_csv.py
```diff
@@ -1972,6 +1972,39 @@ def test_write_csv_decimal(tmpdir, type_factory):
     assert out.column('col').cast(type) == table.column('col')
 
 
+@pytest.mark.parametrize("data_size", (
+    int(1E2),
+    int(1E4),
+    int(1E6)
```
pitrou, line 1978: Why the parameterization? What is the intent here?

python/pyarrow/tests/test_csv.py
```diff
@@ -1972,6 +1972,39 @@ def test_write_csv_decimal(tmpdir, type_factory):
...
+))
+def test_large_binary_write_to_csv(tmpdir, data_size):
+    file_name = tmpdir / "fixedsize_"+str(data_size)+".csv"
+
+    nparr = np.frombuffer(np.random.randint(65, 91, data_size, 'u1'), 'S4')
+
+    fixed_arr = pa.array(nparr, pa.binary(4))
+    fixed_table = pa.Table.from_arrays([fixed_arr], names=['fixedsize'])
+    fixed_table = fixed_table.combine_chunks()
```
pitrou, line 1987: Is this line actually necessary for this test?

python/pyarrow/tests/test_csv.py
anjakefala: From playing with it, `int(1E6)` is the one that actually reproduces the bug (results in a segfault on main). 

Is there value in also having `int(1E2)` or `int(1E4)` to make sure there wasn't a regression with relatively smaller binary arrays? If there isn't, we can remove those parameters.

python/pyarrow/tests/test_csv.py
anjakefala: Not necessary anymore! It will be updated in the next push.

anjakefala: (It's not. =)

vibhatha:
> Please explicitly describe the rationale in the PR description. "This PR fixes an already filed issue which has been evaluated as a bug" is not an appropriate description.

@pitrou I updated the description. Is the updated description helpful?

python/pyarrow/tests/test_csv.py
pitrou: I don't think there is value in that, no.

cpp/src/arrow/compute/kernels/scalar_cast_string.cc
anjakefala: I attempted to expand the test! I'm still new to Arrow-C++, could you take a look at it?

cpp/src/arrow/compute/kernels/scalar_cast_test.cc
```diff
@@ -2104,6 +2104,11 @@ TEST(Cast, BinaryToString) {
     // ARROW-16757: we no longer zero copy, but the contents are equal
     ASSERT_NE(invalid_utf8->data()->buffers[1].get(), strings->data()->buffers[2].get());
     ASSERT_TRUE(invalid_utf8->data()->buffers[1]->Equals(*strings->data()->buffers[2]));
+
+    // ARROW-35901: check that casting with first buffer being a `nullptr` works
```
pitrou, line 2108:
```suggestion
    // GH-35901: check that casting with first buffer being a `nullptr` works
```

cpp/src/arrow/compute/kernels/scalar_cast_test.cc
```diff
@@ -2104,6 +2104,11 @@ TEST(Cast, BinaryToString) {
...
+    auto fixed_array_null = ArrayFromJSON(from_type, "[]");
+    fixed_array_null->data()->buffers[0] = NULLPTR;
+    CheckCast(fixed_array_null, ArrayFromJSON(string_type, "[]"));
```
pitrou, line 2111: Did you make sure that this fails without the patch?

cpp/src/arrow/compute/kernels/scalar_cast_test.cc
```diff
@@ -2104,6 +2104,11 @@ TEST(Cast, BinaryToString) {
     // ARROW-16757: we no longer zero copy, but the contents are equal
     ASSERT_NE(invalid_utf8->data()->buffers[1].get(), strings->data()->buffers[2].get());
     ASSERT_TRUE(invalid_utf8->data()->buffers[1]->Equals(*strings->data()->buffers[2]));
+
+    // ARROW-35901: check that casting with first buffer being a `nullptr` works
+    auto fixed_array_null = ArrayFromJSON(from_type, "[]");
+    fixed_array_null->data()->buffers[0] = NULLPTR;
```
pitrou, line 2110: Let's make sure that this is a valid array?
```suggestion
    fixed_array_null->data()->buffers[0] = NULLPTR;
    ASSERT_OK(fixed_array_null->ValidateFull());
```

python/pyarrow/tests/test_csv.py
```diff
@@ -1972,6 +1972,33 @@ def test_write_csv_decimal(tmpdir, type_factory):
     assert out.column('col').cast(type) == table.column('col')
 
 
+def test_large_binary_write_to_csv(tmpdir):
+    data_size = int(1E6)
```
pitrou, line 1976: Can you add a comment mentioning the GH issue?

python/pyarrow/tests/test_csv.py
```diff
@@ -1972,6 +1972,33 @@ def test_write_csv_decimal(tmpdir, type_factory):
...
+    file_name = tmpdir / "fixedsize_"+str(data_size)+".csv"
```
pitrou, line 1977: I'm not sure why this is parameterizing the file name, but in any case, should try to follow PEP 8 and put spaces around operators.

(also, you can use a f-string)

python/pyarrow/tests/test_csv.py
```diff
@@ -1972,6 +1972,33 @@ def test_write_csv_decimal(tmpdir, type_factory):
...
+
+    nparr = np.frombuffer(np.random.randint(65, 91, data_size, 'u1'), 'S4')
+
+    fixed_arr = pa.array(nparr, pa.binary(4))
+    fixed_table = pa.Table.from_arrays([fixed_arr], names=['fixedsize'])
+
+    write_options = WriteOptions(include_header=True, batch_size=2048,
+                                 delimiter='|')
```
pitrou, line 1985: There is nothing wrong with this, but the custom delimiter seems gratuitous. Is the batch size related to the data size?

pitrou: (if it is, this means you could have a much smaller `data_size` and something like `batch_size = data_size // 4`? also with a comment mentioning why this is important for this test)

cpp/src/arrow/compute/kernels/scalar_cast_test.cc
anjakefala: It now passes with the patch and segfaults on main! The patch was doing the check after an `offset` check. This only affected sliced `Buffers`.

python/pyarrow/tests/test_csv.py
anjakefala: I'm pretty sure both `batch_size` and `delimiter` were picked up from the code that originally caught the bug: https://github.com/apache/arrow/issues/35901.

The `data_size` seems important for triggering the case (smaller `data_sizes` weren't triggering it). I'll play with `batch_size`, and remove the delimiter.

anjakefala: Turns out `batch_size` is not important, either!

vibhatha: The `batch_size` doesn't matter much, but for the sake of error reproduction, I used it. But the data size is important. I agree with @anjakefala

PR #36270, opened 2023-06-23
GH-36251: [MATLAB] Add `Type` property to `arrow.array.Array`

### LINKED ISSUE #36251: [MATLAB] Add `Type` property to `arrow.array.Array` class

Now that we've the initial version of the `arrow.type.Type` class hierarchy, we should add a `Type` property to the `arrow.array.Array` base class. 

### Component(s)

MATLAB

sgilmore10:
### Rationale for this change

Now that we've added the `arrow.type.Type` class hierarchy to the interface, we should add a property named `Type` to the base `arrow.array.Array` class.

All concrete subclasses of the `arrow.array.Array` define a property named `Type`:

```matlab
>> a = arrow.array.Float64Array([1 2 3 4])

a = 

[
  1,
  2,
  3,
  4
]
>> a.Type

ans = 

  Float64Type with properties:

            ID: Float64
      BitWidth: 64
     NumFields: 0
    NumBuffers: 2
```

### Are these changes tested?
Yes, we added test cases in `hNumeric.m` and `tBooleanArray.m`.

Yes.

### Note

1. I noticed that the `tInt16Array.m` test class did not define the `NullSubstitutionValue` property. This meant the test class was abstract, so its tests were not running. That was my mistake. We'll think about ways to verify all the appropriate tests are running in CI.
2. As @kou mentioned, we think it may be worth creating "Type" C++ proxy classes, especially when we start implementing nested array types. For now, we'll leave them as MATLAB-only classes, but will probably change their implementation soon. 
3. Thank you to @kevingurney for the help!
* Closes: #36251

### DIFF
--- a/matlab/src/matlab/+arrow/+array/Array.m
+++ b/matlab/src/matlab/+arrow/+array/Array.m
@@ -1,21 +1,21 @@
+% Licensed to the Apache Software Foundation (ASF) under one or more
+% contributor license agreements.  See the NOTICE file distributed with
+% this work for additional information regarding copyright ownership.
+% The ASF licenses this file to you under the Apache License, Version
+% 2.0 (the "License"); you may not use this file except in compliance
+% with the License.  You may obtain a copy of the License at
+%
+%   http://www.apache.org/licenses/LICENSE-2.0
+%
+% Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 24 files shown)


kou (approved):
+1

PR #36286, opened 2023-06-25
GH-36280: [Python][Parquet] Export C++ WriteRecordBatch in Python API

### LINKED ISSUE #36280: [Python][Parquet] Allow `write_batch` to directly write batch

Currently, Python parquet's `ParquetWriter.write_batch` will fallback to call the `write_table`, `write_table` will call `NewRowGroup`, after write, it will finish the RowGroup, which will causing lots of RowGroups if batch is small.

Nowadays, `WriteRecordBatch` is supported in C++ ParquetWriter, so, it's possible that Python use `WriteRecordBatch` rather than `WriteTable`.

### Component(s)

Parquet, Python

mapleFU:
### Rationale for this change

Export C++ WriteRecordBatch in Python API to allow memory saving

* [x] Add `write_record_batch` in pyx
* [ ] Export to `ParquetWriter` api

### Are these changes tested?

* [ ] Will do

user can using it

* Closes: #36280

### DIFF
--- a/python/pyarrow/_parquet.pxd
+++ b/python/pyarrow/_parquet.pxd
@@ -23,7 +23,7 @@ from pyarrow.includes.libarrow cimport (CChunkedArray, CScalar, CSchema, CStatus
                                         CTable, CMemoryPool, CBuffer,
                                         CKeyValueMetadata,
                                         CRandomAccessFile, COutputStream,
-                                        TimeUnit, CRecordBatchReader)
+                                        TimeUnit, CRecordBatchReader, CRecordBatch)
 from pyarrow.lib cimport _Weakrefable
 
 
--- a/python/pyarrow/_parquet.pxd
+++ b/python/pyarrow/_parquet.pxd
@@ -532,6 +532,7 @@ cdef extern from "parquet/arrow/writer.h" namespace "parquet::arrow" nogil:
         CStatus WriteTable(const CTable& table, int64_t chunk_size)
         CStatus NewRowGroup(int64_t chunk_size)
         CStatus Close()
+        CStatus WriteRecordBatch(const CRecordBatch& batch)
 
         const shared_ptr[CFileMetaData] metadata() const
 
--- a/python/pyarrow/_parquet.pyx
+++ b/python/pyarrow/_parquet.pyx
@@ -1779,6 +1779,19 @@ cdef class ParquetWriter(_Weakrefable):
             check_status(self.writer.get()
                          .WriteTable(deref(ctable), c_row_group_size))
 
+    def write_record_batch(self, RecordBatch record_batch):
+        """
+        Write RecordBatch.
+
+        Parameters
+        ----------
+        record_batch : RecordBatch
+        """
+
+        with nogil:
+            check_status(self.writer.get()
+                         .WriteRecordBatch(deref(record_batch.batch)))
+
     @property
     def metadata(self):
         cdef:
--- a/python/pyarrow/_parquet.pyx
+++ b/python/pyarrow/_parquet.pyx
@@ -1692,7 +1692,6 @@ cdef class ParquetWriter(_Weakrefable):
         int64_t write_batch_size
         int64_t dictionary_pagesize_limit
         object store_schema
-        object write_page_index
 
     def __cinit__(self, where, Schema schema not None, use_dictionary=None,
                   compression=None, version=None,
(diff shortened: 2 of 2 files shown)


pitrou:
> Export C++ WriteRecordBatch in Python API to allow memory saving

Can you explain which memory saving you're talking about? Can you show an example?

mapleFU:
>  Can you explain which memory saving you're talking about? Can you show an example?

When user writes to parquet table row-by-row, we usally batch the rows, then converting the batch to arrow, and write arrow batch to parquet, and when file is large enough, we will close writer and write to oss. Usally when we write to parquet, `parquet::arrow::WriteTable` is finally called, it will finish RowGroup once. So our files end with hundred or even thousands of RowGroups, which causing Metadata grows huge.

This patch exports `parquet::arrow::FileWriter::WriteBatch`, which will buffer the input in one RowGroup. Comparing to direct buffer the records in memory or in arrow file, buffering to parquet might be more straightforward @pitrou

pitrou:
> So our files end with hundred or even thousands of RowGroups, which causing Metadata grows huge.

Hundres of row groups for which file size?

> This patch exports `parquet::arrow::FileWriter::WriteBatch`, which will buffer the input in one RowGroup. Comparing to direct buffer the records in memory or in arrow file, buffering to parquet might be more straightforward @pitrou

But does it increase performance? Decrease file size? By how much?

mapleFU:
@pitrou When RowGroup number grows, there're two problems:
1. size of Metadata in a File grows large, specially when there are lots of columns
2.during reading, `RowGroup(idx)` will be called many times, which making reading from file slow

At lease I just want a `write_record_batch_without_flush_rowgroup` to make writing to parquet more lightweight

mapleFU:
I've a case, for example, we batch every 2000 rows. And write a file with 200,000 rows, the meta would be like:

```
  "TotalRows": "205434",
  "NumberOfRowGroups": "103",
  "NumberOfRealColumns": "97",
  "NumberOfColumns": "97",
```

The file is 3.5Mib, after changing to one row-group, the file can be just 450kb. This also making reading from it slower

jorisvandenbossche:
> Should I just change it to `write_record_batch`, or adding a argument to control it to avoid breaking previous behavior? 

I think we could also consider just "breaking" the current behaviour. It will still honor the max row_group_size, just no longer create one row group per batch that you are writing. 

Or only do this if `row_group_size` keyword is None.

What happens if you do a `write_table` after a `write_batch` if the `write_batch` didn't yet finalize the current row group?

mapleFU:
@jorisvandenbossche 

```c++
  Status NewRowGroup(int64_t chunk_size) override {
    if (row_group_writer_ != nullptr) {
      PARQUET_CATCH_NOT_OK(row_group_writer_->Close());
    }
    PARQUET_CATCH_NOT_OK(row_group_writer_ = writer_->AppendRowGroup());
    return Status::OK();
  }
```

`parquet::arrow::FileWriterImpl::WriteTable` will call `NewRowGroup` for every chunk, which means it will first close previous RowGroup, then split input table to `chunk` by users row group size, and call `AppendRowGroup` to create non-buffered row-group writer for every chunk.

`parquet::arrow::FileWriterImpl::WriteRecordBatch` will close previous rowgroup if previous row-group is not a "buffered" row-group, and append `RecordBatch` to the buffered row-group.

So:

1. If only `WriteTable` is called, every call will create **at lease one** new row group.
2. If only `WriteRecord` is called, every call might not create row group.
3. If `WriteRecordBatch` is called after `WriteTable`, it will find that there is not in-flight row-group, so it will create a buffered row-group and write to it.
4. If `WriteTable` is called after `WriteRecordBatch`, it will find that there is a in-flight row-group, so it will close the buffered row-group, create a non-buffered one, and write to it

pitrou:
Thanks for the explanation @mapleFU . Perhaps we can add an optional argument to `parquet::arrow::FileWriterImpl::WriteTable` to tell it whether to act more like `WriteRecordBatch`?

Something like:
```c++
  /// \brief Write a Table to Parquet.
  ///
  /// If `use_buffering` is false, then any pending row group is closed
  /// at the beginning and at the end of this call.
  /// If `use_buffering` is true, this function reuses an existing
  /// buffered row group until the chunk size is met, and leaves
  /// the last row group open for further writes.
  /// It is recommended to set `use_buffering` to true to minimize
  /// the number of row groups, especially when calling `WriteTable`
  /// with small tables.
  ///
  /// \param table Arrow table to write.
  /// \param chunk_size maximum number of rows to write per row group.
  /// \param use_buffering Whether to potentially buffer data.
  virtual ::arrow::Status WriteTable(
      const ::arrow::Table& table, int64_t chunk_size = DEFAULT_MAX_ROW_GROUP_LENGTH,
      bool use_buffering = false) = 0;
```

jorisvandenbossche:
Yes, that is what I suggested above as well, to just change the behaviour, at least for the current `write_batch`

mapleFU:
I'll add a `WriteRecordBatch` with hint row-group size, and then bump it to Python script.

PR #36289, opened 2023-06-25
GH-35273: [C++] Add integer round kernels

### LINKED ISSUE #35273: BUG: pyarrow.compute.round fails for large integer within int64

```
In [11]: import pyarrow as pa

In [12]: pa.compute.round(pa.array([1372636858620000589]))
ArrowInvalid: Integer value 1372636858620000589 not in range: -9007199254740992 to 9007199254740992

In [13]: pa.__version__
Out[13]: '11.0.0'
```

Not sure why the rounding range is narrower  (2^53) that the int64 range

### Component(s)

Python

js8544:
### Rationale for this change

Currently `round` casts integers to floats which causes undesired behavior.

Add round kernels for integer types.

### Are these changes tested?

Yes.

No.

* Closes: #35273

### DIFF
--- a/cpp/src/arrow/compute/kernels/scalar_round.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_round.cc
@@ -397,7 +397,7 @@ struct RoundImpl<Type, RoundMode::HALF_TO_ODD> {
   template <typename T = Type>
   static constexpr enable_if_integer_value<T> Round(const T val, const T floor,
                                                     const T multiple, Status* st) {
-    if ((floor / multiple) % 2 == 1) {
+    if ((floor / multiple) % 2 != 0) {
       return floor;
     }
     return RoundImpl<T, RoundMode::TOWARDS_INFINITY>::Round(val, floor, multiple, st);
--- a/cpp/src/arrow/compute/kernels/scalar_round.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_round.cc
@@ -132,7 +132,7 @@ struct RoundUtil {
         Pow10Struct<18>::value, Pow10Struct<19>::value};
 
     auto digits10 = std::numeric_limits<T>::digits10;
-    return lut[std::min(power, static_cast<int64_t>(digits10))];
+    return static_cast<T>(lut[std::min(power, static_cast<int64_t>(digits10))]);
   }
 };
 
--- a/cpp/src/arrow/compute/kernels/scalar_round.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_round.cc
@@ -125,7 +125,7 @@ struct RoundUtil {
         10000000000000000ULL,
         100000000000000000ULL,
         1000000000000000000ULL
-        // clang-format on 
+        // clang-format on
     };
 
     return static_cast<T>(lut[power]);
--- a/cpp/src/arrow/compute/kernels/scalar_round.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_round.cc
@@ -84,25 +84,6 @@ bool IsPositive(const Scalar& scalar) {
   return visitor.result;
 }
 
-// N.B. take care not to conflict with type_traits.h as that can cause surprises in a
-// unity build
-
-// A constexpr helper struct to compute powers of 10 at compile time
-// Can use a consteval function once we force C++20
-template <int Exp>
-struct Pow10Struct {
- private:
-  static constexpr uint64_t half_pow = Pow10Struct<Exp / 2>::value;
-
- public:
-  static constexpr uint64_t value = half_pow * half_pow * (Exp % 2 ? 10 : 1);
-};
-
-template <>
-struct Pow10Struct<0> {
-  static constexpr uint64_t value = 1;
-};
-
 struct RoundUtil {
   // Calculate powers of ten with arbitrary integer exponent
   template <typename T>
@@ -124,13 +105,28 @@ struct RoundUtil {
     DCHECK_GE(power, 0);
     DCHECK_LE(power, std::numeric_limits<T>::digits10);
     static constexpr uint64_t lut[] = {
-        Pow10Struct<0>::value,  Pow10Struct<1>::value,  Pow10Struct<2>::value,
-        Pow10Struct<3>::value,  Pow10Struct<4>::value,  Pow10Struct<5>::value,
-        Pow10Struct<6>::value,  Pow10Struct<7>::value,  Pow10Struct<8>::value,
-        Pow10Struct<9>::value,  Pow10Struct<10>::value, Pow10Struct<11>::value,
-        Pow10Struct<12>::value, Pow10Struct<13>::value, Pow10Struct<14>::value,
-        Pow10Struct<15>::value, Pow10Struct<16>::value, Pow10Struct<17>::value,
-        Pow10Struct<18>::value, Pow10Struct<19>::value};
+        // clang-format off
+        1ULL,
+        10ULL,
+        100ULL,
+        1000ULL,
+        10000ULL,
+        100000ULL,
+        1000000ULL,
+        10000000ULL,
+        100000000ULL,
+        1000000000ULL,
+        10000000000ULL,
+        100000000000ULL,
+        1000000000000ULL,
+        10000000000000ULL,
+        100000000000000ULL,
+        1000000000000000ULL,
+        10000000000000000ULL,
+        100000000000000000ULL,
+        1000000000000000000ULL
+        // clang-format on 
+    };
 
     return static_cast<T>(lut[power]);
   }
--- a/cpp/src/arrow/compute/kernels/scalar_round.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_round.cc
@@ -27,6 +27,7 @@
 #include "arrow/compute/cast.h"
 #include "arrow/compute/kernel.h"
 #include "arrow/compute/kernels/base_arithmetic_internal.h"
+#include "arrow/compute/kernels/codegen_internal.h"
 #include "arrow/compute/kernels/common_internal.h"
 #include "arrow/compute/kernels/util_internal.h"
 #include "arrow/type.h"
@@ -121,7 +122,7 @@ struct RoundUtil {
   template <typename T>
   static enable_if_integer_value<T> Pow10(int64_t power) {
     DCHECK_GE(power, 0);
-
+    DCHECK_LE(power, std::numeric_limits<T>::digits10);
     static constexpr uint64_t lut[] = {
         Pow10Struct<0>::value,  Pow10Struct<1>::value,  Pow10Struct<2>::value,
         Pow10Struct<3>::value,  Pow10Struct<4>::value,  Pow10Struct<5>::value,
@@ -131,8 +132,7 @@ struct RoundUtil {
         Pow10Struct<15>::value, Pow10Struct<16>::value, Pow10Struct<17>::value,
         Pow10Struct<18>::value, Pow10Struct<19>::value};
 
-    auto digits10 = std::numeric_limits<T>::digits10;
-    return static_cast<T>(lut[std::min(power, static_cast<int64_t>(digits10))]);
+    return static_cast<T>(lut[power]);
   }
 };
 
@@ -433,6 +433,13 @@ struct RoundOptionsWrapper<RoundOptions, CType> : public OptionsWrapper<RoundOpt
   static Result<std::unique_ptr<KernelState>> Init(KernelContext* ctx,
                                                    const KernelInitArgs& args) {
     if (auto options = static_cast<const OptionsType*>(args.options)) {
+      if constexpr (is_integer_value<CType>::value) {
+        if (-options->ndigits > std::numeric_limits<CType>::digits10) {
+          return Status::Invalid("Rounding to ", options->ndigits,
+                                 " digits is out of range for type ",
+                                 args.inputs[0].ToString());
+        }
+      }
       return std::make_unique<RoundOptionsWrapper>(*options);
     }
     return Status::Invalid(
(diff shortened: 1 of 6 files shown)


westonpace:
CC @EpsilonPrime do you think you could take a look?

EpsilonPrime:
Hi, I've been looking through the PR and it looks pretty good so far.  I'm going to take another pass to see if any of this makes it harder to implement round for Decimal types (I suspect not but it's worth checking).  In the meantime could you add benchmarks for the new kernels?  Thanks!

EpsilonPrime:
Actually it looks like the benchmarks are already there.  I'll run them locally to see what kind of improvement there is.  I suspect there will be a noticeable improvement from this PR because the expensive conversion to float won't need to occur.

js8544:
> Actually it looks like the benchmarks are already there. I'll run them locally to see what kind of improvement there is. I suspect there will be a noticeable improvement from this PR because the expensive conversion to float won't need to occur.

https://gist.github.com/js8544/8471c3106bbaff473fb7bddf4c56b4de I just ran it locally and here's the result.

docs/source/cpp/compute.rst
```diff
@@ -563,30 +563,34 @@ representation based on the rounding criterion.
...
   :struct:`RoundOptions` specifies the rounding precision in terms of number
   of digits.  A negative value corresponds to digits in the non-fractional
   part.  For example, -2 corresponds to rounding to the nearest multiple of
   100 (zeroing the ones and tens digits).  Default value of ``ndigits`` is 0
-  which rounds to the nearest integer.
+  which rounds to the nearest integer. For integer inputs a non-negative 
+  ``ndigits`` value is ignored and the input is returned unchanged. For integer
+  inputs, if ``-ndigits`` is larger than the maximum number of digits the 
+  input type can hold, it is truncated to the maximum digit. For example, 
+  ``round([123], ndigits=-4, round_mode=DOWN)`` returns [100] for ``int8`` type.
```
EpsilonPrime, line 586: (as the ndigits value is silently reduced to -2)

EpsilonPrime:
I ran the benchmarks as well (locally on an M1 Macbook Pro):
`archery benchmark diff review36289 main --benchmark-filter=RoundArrayBenchmark`

The summary version is:

- int8, uint8         130-1200x faster
- int16, uint16      100-870x faster
- int32, uint32     70-830x faster
- int64, uint64     100-700x faster
- float/double       no noticible effect

https://gist.github.com/EpsilonPrime/658c90020a5964064e803cfb7e4761b2

docs/source/cpp/compute.rst
```diff
@@ -563,30 +563,34 @@ representation based on the rounding criterion.
...
+  For integer inputs, an error is returned on overflow.
 * \(3) Round to a multiple where the ``multiple`` option of
   :struct:`RoundToMultipleOptions` specifies the rounding scale.  The rounding
-  multiple has to be a positive value.  For example, 100 corresponds to
-  rounding to the nearest multiple of 100 (zeroing the ones and tens digits).
-  Default value of ``multiple`` is 1 which rounds to the nearest integer.
+  multiple has to be a positive value and can be casted to input type.  
+  For example, 100 corresponds to ounding to the nearest multiple of 100
```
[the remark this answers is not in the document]
js8544: Done

docs/source/cpp/compute.rst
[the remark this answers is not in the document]
js8544: I agree that we should align with RoundToMultiple. I changed the behavior to rejecting the operation when -ndigits is too large.

js8544:
@EpsilonPrime Sorry for the delay, I've updated the PR according to your suggestions. Please re-review it when it's convenient for you. Thanks!

EpsilonPrime:
The changes look great to me.  Thanks for implementing this!

docs/source/cpp/compute.rst
```diff
@@ -563,30 +563,32 @@ representation based on the rounding criterion.
 +-------------------+------------+-------------+-------------------------+----------------------------------+--------+
 | floor             | Unary      | Numeric     | Float32/Float64/Decimal |                                  |        |
 +-------------------+------------+-------------+-------------------------+----------------------------------+--------+
-| round             | Unary      | Numeric     | Float32/Float64/Decimal | :struct:`RoundOptions`           | (1)(2) |
+| round             | Unary      | Numeric     | Input Type              | :struct:`RoundOptions`           | (1)(2) |
```
westonpace, line 566: So I think this is technically a breaking change right?

Before, if we had something like:

```
x = pa.array([1, 2], pa.int32())
y = pc.round(x)
```

Then `y` would be a double array. Now, `y` will be an `int32` array.  I think this is correct and the old behavior was unintentional so I think it is an ok breaking change.  Still, we should make sure to mark the PR as a breaking change if my understanding is correct so that we document it as such in the release notes.

CC @jorisvandenbossche for second opinion.

docs/source/cpp/compute.rst
```diff
@@ -563,30 +563,32 @@ representation based on the rounding criterion.
...
 | trunc             | Unary      | Numeric     | Float32/Float64/Decimal |                                  |        |
 +-------------------+------------+-------------+-------------------------+----------------------------------+--------+
 
-* \(1) Output value is a 64-bit floating-point for integral inputs and the
-  retains the same type for floating-point and decimal inputs.  By default
-  rounding functions displace a value to the nearest integer using
-  HALF_TO_EVEN to resolve ties.  Options are available to control the rounding
-  criterion.  Both ``round`` and ``round_to_multiple`` have the ``round_mode``
-  option to set the rounding mode.
+* \(1)  By default rounding functions displace a value to the nearest
```
westonpace, line 573:
```suggestion
* \(1)  By default rounding functions change a value to the nearest 
```
I realize this wasn't your PR but I think we can use simpler English here.

westonpace (changes requested):
Can you quickly comment on whether you agree this is a breaking change or not?  Then I think we can approve this PR.

docs/source/cpp/compute.rst
js8544: Right, it should be a breaking change. @jorisvandenbossche could you please confirm if it's acceptable?

cpp/src/arrow/compute/kernels/scalar_round.cc
```diff
@@ -96,8 +117,29 @@ struct RoundUtil {
...
+        Pow10Struct<3>::value,  Pow10Struct<4>::value,  Pow10Struct<5>::value,
+        Pow10Struct<6>::value,  Pow10Struct<7>::value,  Pow10Struct<8>::value,
+        Pow10Struct<9>::value,  Pow10Struct<10>::value, Pow10Struct<11>::value,
+        Pow10Struct<12>::value, Pow10Struct<13>::value, Pow10Struct<14>::value,
+        Pow10Struct<15>::value, Pow10Struct<16>::value, Pow10Struct<17>::value,
+        Pow10Struct<18>::value, Pow10Struct<19>::value};
+
+    return static_cast<T>(lut[power]);
+  }
 };
```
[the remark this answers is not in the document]
js8544: Now it does seem like an overkill to me. I've changed it to the simpler way.

docs/source/cpp/compute.rst
jorisvandenbossche: Yes, fully agreed with the summary of Weston above: this was unintentional behaviour (because of automatic casting for numeric types), and it's fine to correct this with a breaking change.

docs/source/cpp/compute.rst
```diff
@@ -563,30 +563,32 @@ representation based on the rounding criterion.
 +-------------------+------------+-------------+-------------------------+----------------------------------+--------+
 | floor             | Unary      | Numeric     | Float32/Float64/Decimal |                                  |        |
```
jorisvandenbossche, line 564: I would have expected that "floor" kernel is a small wrapper around "round" with a specific RoundOptions value. If the output type of "round" changes, that doesn't also change "floor"?

js8544: Not really. Because I didn't add floor kernels for integer types, floor(int) would still be dispatched to floor(float) kernels, and thus calling the round functions for floats.

jorisvandenbossche: But now that Round itself supports rounding integers, how hard would it be to expand the floor/trunc registration to integer types as well? 
(like `MakeUnaryRoundFunction` was updated to loop through all NumericTypes instead of just float32/float64)

pitrou: Is it at all useful to expose those functions for integer inputs?

jorisvandenbossche: Just to have consistent behaviour with the generic round (i.e. always preserve the input type). But that alone is maybe not worth it.

jorisvandenbossche: Although it might also simplify things, given that the `RoundIntegerToFloatingPointFunction` to explicitly cast int to float which is still used for floor/trunc/ceil could then be removed (but didn't check the code in detail)

js8544: https://github.com/apache/arrow/issues/36786 I've created another issue for this.

westonpace (approved):

js8544:
I think this PR is ready to be merged. CI failures are unrelated. @pitrou Would you mind merging this? Thanks!

docs/source/cpp/compute.rst
```diff
@@ -563,30 +563,32 @@ representation based on the rounding criterion.
...
 +-------------------+------------+-------------+-------------------------+----------------------------------+--------+
-| round             | Unary      | Numeric     | Float32/Float64/Decimal | :struct:`RoundOptions`           | (1)(2) |
+| round             | Unary      | Numeric     | Input Type              | :struct:`RoundOptions`           | (1)(2) |
 +-------------------+------------+-------------+-------------------------+----------------------------------+--------+
-| round_to_multiple | Unary      | Numeric     | Float32/Float64/Decimal | :struct:`RoundToMultipleOptions` | (1)(3) |
+| round_to_multiple | Unary      | Numeric     | Input Type              | :struct:`RoundToMultipleOptions` | (1)(3) |
```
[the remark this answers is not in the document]
js8544: Done

cpp/src/arrow/compute/kernels/scalar_round_arithmetic_test.cc
```diff
@@ -1052,7 +1118,56 @@ TYPED_TEST(TestUnaryRoundUnsigned, Round) {
...
     this->SetRoundNdigits(pair.first);
-    this->AssertUnaryOp(Round, values, ArrayFromJSON(float64(), pair.second));
+    this->AssertUnaryOp(Round, values,
+                        ArrayFromJSON(this->type_singleton(), pair.second));
+  }
+
+  // Test different rounding mode
+  // skip uint8 because of its small range
+  if constexpr (!std::is_same_v<TypeParam, UInt8Type>) {
+    std::string values("[0, 1, 13, 50, 115, 176, 200, 250]");
```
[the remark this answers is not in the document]
js8544: Done

cpp/src/arrow/compute/kernels/scalar_round_arithmetic_test.cc
```diff
@@ -1014,21 +1031,70 @@ TYPED_TEST(TestUnaryRoundSigned, Round) {
...
+    }};
+    for (const auto& pair : round_modes_and_expected) {
+      this->SetRoundMode(pair.first);
+      this->AssertUnaryOp(Round, values,
+                          ArrayFromJSON(this->type_singleton(), pair.second));
+    }
+  }
+
+  // An overly large ndigits would cause an error
+  if constexpr (std::is_same_v<TypeParam, Int8Type>) {
```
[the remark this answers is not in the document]
js8544: Right. The if constexpr is removed.

cpp/src/arrow/compute/kernels/scalar_round_arithmetic_test.cc
```diff
@@ -1014,21 +1031,70 @@ TYPED_TEST(TestUnaryRoundSigned, Round) {
...
     this->SetRoundNdigits(pair.first);
-    this->AssertUnaryOp(Round, values, ArrayFromJSON(float64(), pair.second));
+    this->AssertUnaryOp(Round, values,
+                        ArrayFromJSON(this->type_singleton(), pair.second));
+  }
+
+  // Test different rounding mode
+  // skip int8 because of its small range
+  if constexpr (!std::is_same_v<TypeParam, Int8Type>) {
+    std::string values("[0, 1, -13, -50, 115, -176, 200, 250]");
```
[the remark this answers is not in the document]
js8544: I kept -50 and added -150, so that no two options result in identical results. (Changing -50 to -150 would make HALF_TO_ODD and HALF_UP the same).

js8544:
friendly ping :)

docs/source/cpp/compute.rst
```diff
@@ -627,8 +635,8 @@ The example values are given for default values of ``ndigits`` and ``multiple``.
 +-----------------------+--------------------------------------------------------------+---------------------------+
 
 The following table gives examples of how ``ndigits`` (for the ``round``
-function) and ``multiple`` (for ``round_to_multiple``) influence the operance
-performed, respectively.
+function) and ``multiple`` (for ``round_to_multiple`` and ``round_binary``) 
+influence the operance performed, respectively.
```
pitrou, line 639:
```suggestion
influence the operation performed, respectively.
```

docs/source/cpp/compute.rst
```diff
@@ -627,8 +635,8 @@ The example values are given for default values of ``ndigits`` and ``multiple``.
 +-----------------------+--------------------------------------------------------------+---------------------------+
 
 The following table gives examples of how ``ndigits`` (for the ``round``
-function) and ``multiple`` (for ``round_to_multiple``) influence the operance
-performed, respectively.
+function) and ``multiple`` (for ``round_to_multiple`` and ``round_binary``)
```
pitrou, line 638: Hmm, the second input for `round_binary` is equivalent to `ndigits`, not `multiple`, right?

pitrou (changes requested):
Thanks for the update, just two nits

docs/source/cpp/compute.rst
js8544: My bad, fixed.

pitrou (approved):

PR #36294, opened 2023-06-26
GH-36293: [C++] Use ipc_write_options.memory_pool for compressed buffer and shrink after compression

### LINKED ISSUE #34025: [C++] Arrow IPC buffer compression forces the use of default_memory_pool

In the CompressBuffer, it calls AllocateBuffer without passing the memory_pool argument https://github.com/apache/arrow/blob/177a0d4400e61f44dbd045ce05150bba28dd1d74/cpp/src/arrow/ipc/writer.cc#L183

Is there any special reason why it doesn't use the memory_pool in ipc write options? In our use case, we want to manage the memory allocation and avoid using the default memory pool.

### Component(s)

C++

### LINKED ISSUE #36293: [C++] Shrink compressed Arrow IPC buffer

We use the Arrow IPC data format for shuffle operation in our project https://github.com/oap-project/gluten. To evaluate the performance, we run TPCH benchmarks of SF2T on a bare metal machine. By immediately shrinking the buffer after buffer compression, we observed that the average memory usage decreases 31%, and the total runtime decreases 7%.

### Component(s)

C++

marin-ma:
### Rationale for this change

Described in issue #36293 #34025.

* Allocate buffer for compressed data using the memory pool given by the user
* Shrink compressed data buffer after compression to conserve memory, as the compressed data might be much smaller than the theoretical max compressed data size

### Are these changes tested?

Covered by existing tests.

No.

* Closes: #36293
* Closes: #34025

### DIFF
--- a/cpp/src/arrow/ipc/writer.cc
+++ b/cpp/src/arrow/ipc/writer.cc
@@ -192,8 +192,9 @@ class RecordBatchSerializer {
     int64_t maximum_length = codec->MaxCompressedLen(buffer.size(), buffer.data());
     int64_t prefixed_length = buffer.size();
 
-    ARROW_ASSIGN_OR_RAISE(auto result,
-                          AllocateResizableBuffer(maximum_length + sizeof(int64_t)));
+    ARROW_ASSIGN_OR_RAISE(
+        auto result,
+        AllocateResizableBuffer(maximum_length + sizeof(int64_t), options_.memory_pool));
     ARROW_ASSIGN_OR_RAISE(auto actual_length,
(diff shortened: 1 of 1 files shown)


pitrou (approved):
Nice catch indeed, thank you!

felipecrv:
@pitrou should we avoid setting `default_memory_pool()` as a default argument value? At least in internal APIs?

It bothers me that the compiler could ensure we always inject a pool, but it's easy to mistakenly use `default_memory_pool()` because it's the default argument value in so many functions and constructors.

felipecrv:
I created an issue #36360 but I won't be working on it right now. I marked it as good-first-issue as well.

PR #36310, opened 2023-06-26
GH-36309: [C++] Add ability to cast between scalars of list-like types

### LINKED ISSUE #36309: [C++] Ability to cast between scalars of list-like types

List scalars are very simple and don't change much between the list types (the scalar wraps an array) and currently it's not possible to cast between (list, large-list, and fixed-size-list scalars).

### Component(s)

C++

felipecrv:
### Rationale for this change

Makes it easier to work generically with different list types.

More cast implementations between list-like types of scalars.

### Are these changes tested?

Yes. With new unit tests.
* Closes: #36309

### DIFF
--- a/cpp/src/arrow/scalar.cc
+++ b/cpp/src/arrow/scalar.cc
@@ -1127,19 +1127,14 @@ Status CastImpl(const StructScalar& from, StringScalar* to) {
   return Status::OK();
 }
 
-template <typename T>
-using is_list_or_list_view_type =
-    std::integral_constant<bool, std::is_same<T, ListType>::value ||
-                                     std::is_same<T, ListViewType>::value ||
-                                     std::is_same<T, LargeListType>::value ||
-                                     std::is_same<T, FixedSizeListType>::value>;
-
-// casts between list and list-view types
+// casts between variable-length and fixed-length list types
 template <typename ToScalar>
-enable_if_t<is_list_or_list_view_type<typename ToScalar::TypeClass>::value, Status>
+enable_if_t<is_list_like_type<typename ToScalar::TypeClass>::value &&
+                ToScalar::TypeClass::type_id != Type::MAP,
+            Status>
 CastImpl(const BaseListScalar& from, ToScalar* to) {
   DCHECK(from.type->id() == Type::LIST || from.type->id() == Type::LARGE_LIST ||
-         from.type->id() == Type::LIST_VIEW || from.type->id() == Type::FIXED_SIZE_LIST);
+         from.type->id() == Type::FIXED_SIZE_LIST);
 
   if constexpr (sizeof(typename ToScalar::TypeClass::offset_type) < sizeof(int64_t)) {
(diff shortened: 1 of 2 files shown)


cpp/src/arrow/scalar.cc
```diff
@@ -1127,6 +1127,28 @@ Status CastImpl(const StructScalar& from, StringScalar* to) {
...
 }
 
+// casts between variable-length and fixed-length list types
+template <typename ToScalar>
+enable_if_t<is_list_like_type<typename ToScalar::TypeClass>::value &&
+                ToScalar::TypeClass::type_id != Type::MAP,
+            Status>
+CastImpl(const BaseListScalar& from, ToScalar* to) {
+  DCHECK(from.type->id() == Type::LIST || from.type->id() == Type::LARGE_LIST ||
+         from.type->id() == Type::FIXED_SIZE_LIST);
```
pitrou, line 1137: Careful: `MapScalar` inherits from `BaseListScalar`. Are you sure it can't use this overload?

cpp/src/arrow/scalar_test.cc
```diff
@@ -1135,6 +1136,25 @@ class TestListScalar : public ::testing::Test {
     ASSERT_NE(a0->hash(), b0->hash());
   }
 
+  void TestCast(ScalarType& scalar, const std::shared_ptr<DataType>& to_type) {
```
pitrou, line 1139: Can we avoid non-const refs?
```suggestion
  void TestCast(const ScalarType& scalar, const std::shared_ptr<DataType>& to_type) {
```

cpp/src/arrow/scalar_test.cc
```diff
@@ -391,6 +390,8 @@ class TestRealScalar : public ::testing::Test {
     ASSERT_TRUE(list_nan.ApproxEquals(list_other_nan, options));
   }
 
+  void TestListOf() { TestListOf<ListScalar>(list(type_)); }
```
pitrou, line 393: Is it me, or is `TestListOf` never called with types other than plain list?

pitrou: And can we add a cast test between map and non-map-list-like somewhere?

pitrou (changes requested):

cpp/src/arrow/scalar.cc
felipecrv: I will use `is_var_length_list` in the enable-if and expand the test cases.

felipecrv: Used a better predicate for the enable_if here and extended it to gracefully handle Map as well.

cpp/src/arrow/scalar_test.cc
felipecrv: I added `large_list` here, but note that the cast tests are the ones relevant here.

pitrou (approved):
+1, thanks for the update @felipecrv

PR #36314, opened 2023-06-26
GH-36096: [Python] Call __from_arrow__ in Array.to_pandas

### LINKED ISSUE #36096: [Python] Array vs ChunkedArray to_pandas discrepancy for pandas extension dtypes implementing __from_arrow__
When you have a arrow type that maps to a pandas extension dtype implementing `__from_arrow__`, the `ChunkedArray.to_pandas` conversion will call the `pd_dtype.__from_arrow__`, but the `Array.to_pandas` version does not.

Typically, if only encounter such dtypes this if you have an pyarrow ExtensionArray as well, and in `ExtensionArray.to_pandas`, we also check for the dtype having `__from_arrow__`:

https://github.com/apache/arrow/blob/475b5b9463b64bec4e03a47e3277076db246bd35/python/pyarrow/array.pxi#L3094-L3104

But the base class `Array.to_pandas` doesn't do this. And recently, one of the pandas dtypes that map to a non-extension array on the pyarrow side (i.e. DatetimeTZDtype, mapping to timestamp with tz) added `__from_arrow__`. Which means that for this dtype, the conversion takes a different code path. 

```
from datetime import datetime
import pyarrow as pa

arr = pa.array([datetime(1, 1, 1)], pa.timestamp("s", tz="America/New_York"))
table = pa.table({'a': arr})
# doesn't call DatetimeTZDtype.__from_arrow__
arr.to_pandas()
# ChunkedArray does call that
table["a"].to_pandas()
```

danepitkin:
### Rationale for this change

Array.to_pandas should mimic ChunkedArray.to_pandas implementation. Notably, there is a missing call to `__from_arrow__` if the attribute exists.

### Are these changes tested?

Requires dev pandas. Can manually kick off integration tests to test with pandas nightly.

No
* Closes: #36096

### DIFF
--- a/python/pyarrow/array.pxi
+++ b/python/pyarrow/array.pxi
@@ -1475,6 +1475,17 @@ cdef class Array(_PandasConvertible):
         return self.take(indices)
 
     def _to_pandas(self, options, types_mapper=None, **kwargs):
+        pandas_dtype = None
+        try:
+            pandas_dtype = self.type.to_pandas_dtype()
+        except NotImplementedError:
+            pass
+
+        # pandas ExtensionDtype that implements conversion from pyarrow
+        if hasattr(pandas_dtype, '__from_arrow__'):
+            arr = pandas_dtype.__from_arrow__(self)
+            return pandas_api.series(arr, copy=False)
+
         return _array_like_to_pandas(self, options, types_mapper=types_mapper)
 
     def __array__(self, dtype=None):
--- a/python/pyarrow/array.pxi
+++ b/python/pyarrow/array.pxi
@@ -1475,22 +1475,6 @@ cdef class Array(_PandasConvertible):
         return self.take(indices)
 
     def _to_pandas(self, options, types_mapper=None, **kwargs):
-        pandas_dtype = None
-
-        if types_mapper:
-            pandas_dtype = types_mapper(self.type)
-        elif self.type.id == _Type_EXTENSION:
-            try:
-                pandas_dtype = self.type.to_pandas_dtype()
-            except NotImplementedError:
-                pass
-
-        # Only call __from_arrow__ for Arrow extension types or when explicitly
-        # overridden via types_mapper
-        if hasattr(pandas_dtype, '__from_arrow__'):
-            arr = pandas_dtype.__from_arrow__(self)
-            return pandas_api.series(arr, copy=False)
-
         return _array_like_to_pandas(self, options, types_mapper=types_mapper)
 
     def __array__(self, dtype=None):
@@ -1689,6 +1673,21 @@ cdef _array_like_to_pandas(obj, options, types_mapper):
 
     original_type = obj.type
     name = obj._name
+    dtype = None
+
+    if types_mapper:
+        dtype = types_mapper(original_type)
+    elif original_type.id == _Type_EXTENSION:
+        try:
+            dtype = original_type.to_pandas_dtype()
+        except NotImplementedError:
+            pass
+
+    # Only call __from_arrow__ for Arrow extension types or when explicitly
+    # overridden via types_mapper
+    if hasattr(dtype, '__from_arrow__'):
+        arr = dtype.__from_arrow__(obj)
+        return pandas_api.series(arr, name=name, copy=False)
 
     # ARROW-3789(wesm): Convert date/timestamp types to datetime64[ns]
     c_options.coerce_temporal_nanoseconds = True
@@ -3106,26 +3105,6 @@ cdef class ExtensionArray(Array):
         cdef Array result = pyarrow_wrap_array(<shared_ptr[CArray]> ext_array)
         result.validate()
         return result
-
-    def _to_pandas(self, options, types_mapper=None, **kwargs):
-        pandas_dtype = None
-
-        if types_mapper:
-            pandas_dtype = types_mapper(self.type)
-        else:
-            try:
-                pandas_dtype = self.type.to_pandas_dtype()
-            except NotImplementedError:
-                pass
-
-        # Only call __from_arrow__ for Arrow extension types or when explicitly
-        # overridden via types_mapper
-        if hasattr(pandas_dtype, '__from_arrow__'):
-            arr = pandas_dtype.__from_arrow__(self)
-            return pandas_api.series(arr, copy=False)
-
-        # otherwise convert the storage array with the base implementation
-        return Array._to_pandas(self.storage, options, types_mapper, **kwargs)
 
 
 class FixedShapeTensorArray(ExtensionArray):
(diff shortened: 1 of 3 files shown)


danepitkin:

jorisvandenbossche:
Hmm, so the failing tests point out an issue with this approach: we have some keywords to control the conversion, notably `timestamp_as_object` in this case. 
And if the user passes this, and we just call `pandas_dtype.__from_arrow__` nonetheless, this keyword gets ignored.

But, we also already have this problem, as this already happens for the ChunkedArray conversion:

```
In [8]: from datetime import datetime
   ...: import pyarrow as pa
   ...: 
   ...: arr = pa.array([datetime(2001, 1, 1)], pa.timestamp("s", tz="America/New_York"))
   ...: table = pa.table({'a': arr})

In [9]: arr.to_pandas(timestamp_as_object=True)
Out[9]: 
0    2000-12-31 19:00:00-05:00
dtype: object

In [10]: table["a"].to_pandas(timestamp_as_object=True)
Out[10]: 
0   2000-12-31 19:00:00-05:00
Name: a, dtype: datetime64[ns, America/New_York]
```

jorisvandenbossche:
One option could be to decide that we _only_ want to use this `pandas_dtype.__from_arrow__` if you actually have an extension type on the arrow side (and so don't do this for timestamp data, also not for the ChunkedArray case), or when you specified a pandas extension dtype with `types_mapper`. Then for all default cases, we (pyarrow) just own the conversion and can ensure the keywords are honored.

python/pyarrow/array.pxi
```diff
@@ -1475,6 +1475,17 @@ cdef class Array(_PandasConvertible):
         return self.take(indices)
 
     def _to_pandas(self, options, types_mapper=None, **kwargs):
+        pandas_dtype = None
+        try:
+            pandas_dtype = self.type.to_pandas_dtype()
```
jorisvandenbossche, line 1480: Something that we should also do the the other places where this logic lives (eg ChunkedArray._to_pandas), but I think we should give precedence to `types_mapper`, if specified. Because as a user, if you specify that keyword, you want to be able to override the default conversion.

danepitkin:

danepitkin:

python/pyarrow/array.pxi
```diff
@@ -1476,12 +1476,17 @@ cdef class Array(_PandasConvertible):
...
         pandas_dtype = None
-        try:
-            pandas_dtype = self.type.to_pandas_dtype()
-        except NotImplementedError:
-            pass
 
-        # pandas ExtensionDtype that implements conversion from pyarrow
+        if types_mapper:
+            pandas_dtype = types_mapper(self.type)
+        elif issubclass(type(self.type), ExtensionType):
```
jorisvandenbossche, line 1482:
```suggestion
        elif self.type.id == _Type_EXTENSION:
```

python/pyarrow/array.pxi
```diff
@@ -3104,12 +3109,18 @@ cdef class ExtensionArray(Array):
...
         pandas_dtype = None
-        try:
-            pandas_dtype = self.type.to_pandas_dtype()
-        except NotImplementedError:
-            pass
+        types_mapper = kwargs.get('types_mapper', None)
+
+        if types_mapper:
+            pandas_dtype = types_mapper(self.type)
+        elif issubclass(type(self.type), ExtensionType):
```
jorisvandenbossche: This check isn't needed here, since this is in the ExtensionArray subclass. 

But more generally, I am wondering if this method is actually still needed now we are adding similar logic to the base class' `to_pandas`

danepitkin: The only difference here now is calling `self.type.to_pandas_dtype()` vs `self.storage.type.to_pandas_dtype()` (storage gets passed to the base Array function).

danepitkin: I did remove the issubclass check in a recent refactor!

python/pyarrow/array.pxi
danepitkin: Much cleaner! I'll add to ChunkedArray as well.

danepitkin:

python/pyarrow/array.pxi
```diff
@@ -1475,6 +1475,22 @@ cdef class Array(_PandasConvertible):
...
+            try:
+                pandas_dtype = self.type.to_pandas_dtype()
+            except NotImplementedError:
+                pass
+
+        # Only call __from_arrow__ for Arrow extension types or when explicitly
+        # overridden via types_mapper
+        if hasattr(pandas_dtype, '__from_arrow__'):
+            arr = pandas_dtype.__from_arrow__(self)
+            return pandas_api.series(arr, copy=False)
```
jorisvandenbossche, line 1492: Given that this code is both here and in the ChunkedArray version, and both methods also end up calling `_array_like_to_pandas`, we can maybe move this logic there to avoid duplicating it.

python/pyarrow/array.pxi
jorisvandenbossche: Currently we are indeed passing down `self.storage`, but I seem to remember that we added similar logic to the underlying C++ ConvertArrayToPandas to also (by default) use the storage array to do the actual conversion. So it _might_ actually work by just removing the overridden method completely.

python/pyarrow/array.pxi
danepitkin: +1 on code deduplication, I'll give it a shot. The only differences currently are the parameters passed into `pandas_api.series(arr)`. Arrays want copy=False, and ChunkedArray has enough info to add name=self._name.

danepitkin: From my testing, `Array` does always have a `None` "column" name. Good to know `ChunkedArray` could benefit from `copy=False`! I thought maybe it was left out on purpose since the memory is not guaranteed to be contiguous.

python/pyarrow/array.pxi
danepitkin: Looks like you are right! https://github.com/apache/arrow/blob/0cea12ffc87915e3808d153d4d8e5f115e48d487/python/pyarrow/src/arrow/python/arrow_to_pandas.cc#L2457

Running pytests locally seems to agree as well. I'll run the integration tests with this change to verify.

danepitkin:

python/pyarrow/array.pxi
jorisvandenbossche: > I thought maybe it was left out on purpose since the memory is not guaranteed to be contiguous.

ChunkedArray itself can indeed consist of multiple chunks, but at the moment we reach the `pandas_api.series(arr, copy=False)` code, we already converted the ChunkedArray to a single contiguous numpy array (or pandas ExtensionArray). And anyway, this `copy=False` is mostly to indicate to pandas they don't need to do an extra copy (https://github.com/apache/arrow/pull/34593).

jorisvandenbossche (approved):
Nice!

PR #36315, opened 2023-06-26
GH-36236: [R][Python] Add performance comparison between Release Candidate and previous release

### LINKED ISSUE #36236: [Release] Add performance comparison between Release Candidate and previous release

On  the 12.0.0 release we introduced a performance regression and we had to create the 12.0.1 release to fix it.
The performance regression could have been found during release candidate verification if we would have had some performance comparison between the new Release Candidate and the old Release.
There has been some discussion on the Arrow biweekly community call to introduce a step during the release to generate a conbench comparison between the new RC and the previous release.

This issue is to track this improvement and add this step so we have more data around performance when voting for the release.

### Component(s)

Release

boshek:
### Rationale for this change

This is WIP automated reporting product that compares two commits. The intent here is that this will compare a RC and a previous release via the conbench API and then summarize that information graphically. Ideally this is run automatically as part of a release candidate process. 

It is described well in #36236 

This PR adds a quarto document which can be rendered into a static html file. To specify which commits should be compared one can set the `BASELINE_GIT_COMMIT` and `CONTENDER_GIT_COMMIT` variables. 

@raulcd has offered to determine a process to auto-detect the correct shas. 

I am happy to work on the CI to render this. I think an R + renv + quarto action workflow would work pretty well here and ensure some stability for R package versions. However, I have not checked in a renv lockfile as of yet so as not to pre-suppose any solutions.  

Right now we are only looking at comparisons with the `ursa-i9-9960x` machine. Let me know if we want to expand beyond that. I am attaching a zipped html file which contains an example static html doc. In the future hopefully we will have a better process to share rendering artifacts:

[arrow-release-report.html.zip](https://github.com/apache/arrow/files/11875579/arrow-release-report.html.zip)

* Closes: #36236

### DIFF
--- a/dev/release/perf-report/_plots.qmd
+++ b/dev/release/perf-report/_plots.qmd
@@ -1,0 +1,5 @@
+```{r}
+#| results: asis
+cat("###", bm_name, "\n")
+plot_comparison(conbench_list[[lang]][[bm_name]])
+```
--- a/dev/release/perf-report/arrow-release-report.qmd
+++ b/dev/release/perf-report/arrow-release-report.qmd
@@ -29,8 +29,8 @@ Sys.unsetenv("CONBENCH_EMAIL")
 Sys.unsetenv("CONBENCH_URL")
 Sys.unsetenv("CONBENCH_PASSWORD")
 
-baseline_git_commit <- Sys.getenv("BASELINE_GIT_COMMIT", "9736dde84bb2e6996d1d12f6a044c33398e3c3a3") 
-contender_git_commit <-  Sys.getenv("CONTENDER_GIT_COMMIT", "6af660f48472b8b45a5e01b7136b9b040b185eb1")
+baseline_git_commit <- Sys.getenv("BASELINE_GIT_COMMIT", "6af660f48472b8b45a5e01b7136b9b040b185eb1") 
+contender_git_commit <-  Sys.getenv("CONTENDER_GIT_COMMIT", "ac2d207611ce25c91fb9fc90d5eaff2933609660")
 hardware_name <- "ursa-i9-9960x"
 
 library(dplyr)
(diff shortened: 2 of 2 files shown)


assignUser:
I think it would make sense to make this into a crossbow job that is triggered during the rc preparation process as we already have everything setup over there to be able to publish static pages via s3:

- trigger conbench on rc commit (via comment on rc PR?)
- start job in a later step (e.g. after binaries are prepared/signed) with benchmarked commit and commit from last-release-tag (insert bash :magic_wand: here)
- insert link to rendered perf report into rc vote message

What do you think @raulcd ?

boshek:
> I think it would make sense to make this into a crossbow job that is triggered during the rc preparation process as we already have everything setup over there to be able to publish static pages via s3:
> 
> * trigger conbench on rc commit (via comment on rc PR?)
> * start job in a later step (e.g. after binaries are prepared/signed) with benchmarked commit and commit from last-release-tag (insert bash 🪄 here)
> * insert link to rendered perf report into rc vote message
> 
> What do you think @raulcd ?

@assignUser I was thinking I could create a repository dispatch workflow over in crossbow which renders this report. I don't know if this was also implied but I also think we _could_ move this raw file into the crossbow repo as well (though that obviously isn't required). What do you think? I'd have some cycles to start on that so this step:

> * start job in a later step (e.g. after binaries are prepared/signed) with benchmarked commit and commit from last-release-tag (insert bash 🪄 here)

just has to construct the payload and trigger the rendering.

assignUser:
>  create a repository dispatch workflow 

That would be the "normal" way to do this but I think in this case it would be better to create a crossbow job to do it. That way we can also make it an option to only render the report and upload it as an artifact without publishing it. e.g. to review easier on changes to the report.

The windows vcpkg job would be a simple example to follow: https://github.com/apache/arrow/blob/main/dev/tasks/tasks.yml#L1538

We can also pair on this if you'd like :)

boshek:
Closing this PR is favour of implementing this over in crossbow: https://github.com/ursacomputing/crossbow/pull/61

PR #36316, opened 2023-06-27
GH-36317: [C++] Return a BufferVector from CleanListOffsets

### LINKED ISSUE #36317: [C++] Simplify CleanListOffsets

It takes output parameters, but it could return a `Result<BufferVector>` instead.

### Component(s)

C++

felipecrv:
### Rationale for this change

`CleanListOffsets` is taking output parameters when it could return multiple buffers in a vector exactly how they are supposed to be used to construct the list-like array.

Besides that, having the logic for setting the validity buffer split between it and the caller is unnecessarily confusing.

 - Addition of a new constructor for `MapArray` that can accept a pre-allocated `BufferVector`
 - Removal of output parameter in `CleanListOffsets` and change of return type

### Are these changes tested?

Yes, by existing tests. A pre-existing `MapArray` ctor delegates to this new ctor, so all tests that exercise the old ctor also exercise the new.
* Closes: #36317

### DIFF
--- a/cpp/src/arrow/array/array_nested.cc
+++ b/cpp/src/arrow/array/array_nested.cc
@@ -337,17 +337,23 @@ MapArray::MapArray(const std::shared_ptr<DataType>& type, int64_t length,
 }
 
 MapArray::MapArray(const std::shared_ptr<DataType>& type, int64_t length,
+                   BufferVector buffers, const std::shared_ptr<Array>& keys,
+                   const std::shared_ptr<Array>& items, int64_t null_count,
+                   int64_t offset) {
+  auto pair_data = ArrayData::Make(type->fields()[0]->type(), keys->data()->length,
+                                   {nullptr}, {keys->data(), items->data()}, 0, offset);
+  auto map_data =
+      ArrayData::Make(type, length, std::move(buffers), {pair_data}, null_count, offset);
+  SetData(map_data);
+}
+
+MapArray::MapArray(const std::shared_ptr<DataType>& type, int64_t length,
                    const std::shared_ptr<Buffer>& offsets,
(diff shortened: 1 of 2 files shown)


kou:
Could you open an issue for non MINOR change?

Our MINOR definition: https://github.com/apache/arrow/blob/main/CONTRIBUTING.md#Minor-Fixes
(This is also linked from the auto generated message: https://github.com/apache/arrow/pull/36316#issuecomment-1608586156 )

felipecrv:
> Could you open an issue for non MINOR change?
> 
> Our MINOR definition: https://github.com/apache/arrow/blob/main/CONTRIBUTING.md#Minor-Fixes (This is also linked from the auto generated message: [#36316 (comment)](https://github.com/apache/arrow/pull/36316#issuecomment-1608586156) )

Done.

cpp/src/arrow/array/array_nested.cc
```diff
@@ -74,9 +76,8 @@ Status CleanListOffsets(const Array& offsets, MemoryPool* pool,
     // Copy valid bits, ignoring the final offset (since for a length N list array,
     // we have N + 1 offsets)
     ARROW_ASSIGN_OR_RAISE(
-        auto clean_valid_bits,
+        auto clean_validitty_buffer,
```
pitrou, line 79: Obvious typo in variable name here.

pitrou (changes requested):
LGTM except for one nit. Thanks @felipecrv !

cpp/src/arrow/array/array_nested.cc
felipecrv: Fixed.

pitrou (approved):

PR #36321, opened 2023-06-27
MINOR: [R] Require cmake 3.16 in bundled build script

nealrichardson:
Followup to #35921. The nixlibs.R build script checks for suitable cmake on the system, so since we require newer cmake for C++ now, we need to update the check here. 

I refactored slightly so that there now is a variable in the script called `cmake_minimum_required`, same as in the CMakeLists.txt files throughout the project, so hopefully the next time we update the cmake version, we'll find this file better.

I also bumped the version of cmake that we download (if cmake 3.16 is not found) to the latest release.

cc @kou

### DIFF
--- a/r/tools/nixlibs.R
+++ b/r/tools/nixlibs.R
@@ -481,13 +481,13 @@ build_libarrow <- function(src_dir, dst_dir) {
   invisible(status)
 }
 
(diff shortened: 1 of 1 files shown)


kou (approved):
+1

Oh, thanks! I missed this...

PR #36322, opened 2023-06-27
GH-36318: [Go] only decode lengths for the number of existing values, not for all nvalues.

### LINKED ISSUE #36318: [Go][Parquet] DeltaLengthByteArray encoding breaks with null values in go library

The DeltaLengthByteArray encoding cannot handle null values in the go library.
It fails on 
a) all nulls in one batch - here it reaches an index out of bounds error when initializing the delta bit packing decoder
b) some nulls in one batch - here it continues to decode the lengths for all rows and interprets actual data for lengths information. Data that later is missing when it wants to decode it.

I have a fix and will update the Issue as soon as the PR is created.

### Component(s)

Go

convoi:
### Rationale for this change

Fixes issue 36318.
DeltaLengthBinaryArray Encoding fails to handle null values.

Instead of decoding lengths for "all" values (even the undefined ones), we only decode the lengths for the actually set values.
The Go Version of arrow was unable to read parquet files it produced itself if the mentioned encoding was used but the values contain nulls.

### Are these changes tested?

Tests are included.

No.

* Closes: #36318

### DIFF
--- a/go/parquet/internal/encoding/delta_byte_array_test.go
+++ b/go/parquet/internal/encoding/delta_byte_array_test.go
@@ -1,3 +1,19 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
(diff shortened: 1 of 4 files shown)


zeroshade (approved):

go/parquet/internal/encoding/delta_byte_array_test.go
```diff
@@ -0,0 +1,31 @@
+package encoding
```
zeroshade, line 1: You need to add the Apache license to the top of this file please. That will fix the failing CI

Issue #29678, opened 2021-09-22
[C++] Vendor a span implementation


asfimport:
std::span is a basic vocabulary type which won't be available to us until c++20. We should vendor an implementation similar to our util::optional

**Reporter**: [Ben Kietzman](https://issues.apache.org/jira/browse/ARROW-14083) / @bkietz

<sub>**Note**: *This issue was originally created as [ARROW-14083](https://issues.apache.org/jira/browse/ARROW-14083). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

bkietz:
A span polyfill was added in https://github.com/apache/arrow/pull/36334

PR #36334, opened 2023-06-27
GH-15281: [C++] Replace bytes_view alias with span

### LINKED ISSUE #15281: [C++] basic_string_view<long> will be invalid in future libc++

Libc++ has made a change to stop supporting std::char_traits (and thus, also, basic_string and basic_string_view) for non-character types: https://discourse.llvm.org/t/deprecating-std-string-t-for-non-character-t/66779

Apparently, such support was never required by the standard, and is considered a mistake.

The initial libc++ change made this a hard error, but after reports of breakage, it was downgraded to a deprecation warning. See review thread https://reviews.llvm.org/D138307. It is now expected to become a hard error again in LLVM 18.

One of the projects broken is Apache Arrow, due to the usage here:
https://github.com/apache/arrow/blob/53752adc6b81166cd4ee7db5a819494042f29197/cpp/src/arrow/util/bitmap.h#L52

E.g., 
```
/usr/include/c++/v1/string_view:279:45: error: implicit instantiation of undefined template 'std::char_traits<unsigned long>'
    static_assert((is_same<_CharT, typename traits_type::char_type>::value),
                                            ^
arrow/util/bitmap.h:153:16: note: in instantiation of template class 'std::basic_string_view<unsigned long>' requested here
    View<Word> words[N];
               ^
arrow/compute/kernels/scalar_if_else.cc:1595:15: note: in instantiation of function template specialization 'arrow::internal::Bitmap::VisitWords<3UL, (lambda at arrow/compute/kernels/scalar_if_else.cc:1595:35), unsigned long>' requested here
      Bitmap::VisitWords(bitmaps, [&](std::array<uint64_t, 3> words) {
              ^
arrow/compute/kernels/scalar_if_else.cc:1652:14: note: in instantiation of function template specialization 'arrow::compute::internal::(anonymous namespace)::ExecArrayCaseWhen<arrow::BooleanType>' requested here
      return ExecArrayCaseWhen<Type>(ctx, batch, out);
             ^
arrow/compute/kernels/codegen_internal.h:1173:47: note: in instantiation of member function 'arrow::compute::internal::(anonymous namespace)::CaseWhenFunctor<arrow::BooleanType>::Exec' requested here
      return Generator<BooleanType, Args...>::Exec;
                                              ^
arrow/compute/kernels/scalar_if_else.cc:2738:17: note: in instantiation of function template specialization 'arrow::compute::internal::GenerateTypeAgnosticPrimitive<arrow::compute::internal::(anonymous namespace)::CaseWhenFunctor>' requested here
    auto exec = GenerateTypeAgnosticPrimitive<CaseWhenFunctor>(*type);
```

I suspect "View" needs to be changed to use some other type than std::basic_string_view (could be std::span if you required c++20, but you probably don't want to require c++20 yet.)

### Component(s)

C++

bkietz:
### Rationale for this change

basic_string_view was never really intended to support types other than chars; the template which the STL intends for use in this situation is std::span 

Since std::span is added in c++20, a minimal polyfill is added named util::span.
- doesn't include support for static extents
- some accessors like `front()` are not present in util::span

### User facing changes:

- The broken alias bytes_view and its header are removed. I don't expect this to trouble anybody much since it isn't used in much of our API, and if it's really necessary it will be trivial to add it back with `using bytes_view = span<uint8_t const>;` or so

* Closes: #15281


pitrou:
@benibus Do you want to review this?

pitrou:
Should also close #29678.

bkietz:
To workaround https://github.com/apache/arrow/issues/36379 for this PR I could add an overload of FormatMatcherDescription which forwards to the v1.11 one when necessary but won't be called otherwise. However that seems pretty hacky. Please advise

pitrou:
Let's not overengineer it. You just want to check the pointer/length of a span, so you can write a trivial helper function `AssertSpansEqual` (or `AssertSpansIdentical` if you want to stress that it's not a logical comparison of the underlying values).

cpp/src/arrow/util/span.h
```diff
@@ -0,0 +1,124 @@
...
+
+  constexpr bool operator==(span const& other) const {
+    if (size_ != other.size_) return false;
+
+    T* ptr = data_;
+    for (T const& e : other) {
+      if (*ptr++ != e) return false;
+    }
+    return true;
+  }
```
felipecrv, line 100: Does the compiler turns this into `memcmp` when `T` is an integer type?

felipecrv (commented):
@bkietz you remove an existing public header. Shouldn't that be mentioned in the PR description as an user-facing change?

cpp/src/arrow/util/span.h
bkietz: I've played with this a bit and couldn't get the compiler to do it automatically. I'll use `if constexpr (std::is_integral_v<T>)` to dispatch to memcmp

cpp/src/arrow/util/bitmap.h
```diff
@@ -387,10 +384,10 @@ class ARROW_EXPORT Bitmap : public util::ToStringOstreamable<Bitmap>,
   int64_t length() const { return length_; }
 
   /// string_view of all bytes which contain any bit in this Bitmap
```
benibus, line 386: Maybe we should alter these method comments now? e.g:
```suggestion
  /// span of all bytes which contain any bit in this Bitmap
```

benibus (commented):
Thanks, this will be useful! In general, LGTM - just one nit.

felipecrv (approved):

cpp/src/arrow/util/span.h
```diff
@@ -0,0 +1,129 @@
...
+
+  constexpr size_t size() const { return size_; }
+  constexpr size_t size_bytes() const { return size_ * sizeof(T); }
+  constexpr bool empty() const { return size_ == 0; }
+
+  constexpr T& operator[](size_t i) { return data_[i]; }
+  constexpr const T& operator[](size_t i) const { return data_[i]; }
+
+  constexpr span subspan(size_t offset) const {
+    if (offset > size_) return {};
```
pitrou, line 81: Should we still return a non-null pointer?
```suggestion
    if (offset > size_) return {data_, 0};
```

cpp/src/arrow/util/span.h
```diff
@@ -0,0 +1,129 @@
...
+
+  template <
+      typename R,
+      typename DisableUnlessConstructibleFromDataAndSize =
+          decltype(span<T>(std::data(std::declval<R&>()), std::size(std::declval<R&>()))),
+      typename DisableUnlessSimilarTypes = std::enable_if_t<std::is_same_v<
+          std::decay_t<std::remove_pointer_t<decltype(std::data(std::declval<R&>()))>>,
+          std::decay_t<T>>>>
+  // NOLINTNEXTLINE runtime/explicit, non-const reference
+  constexpr span(R& range) : span{std::data(range), std::size(range)} {}
```
pitrou, line 67: Should this be `R&&`?

cpp/src/arrow/util/span.h
bkietz: alright

cpp/src/arrow/util/span.h
bkietz: Alright, I'll add a test suite to check more construction cases against std::span

cpp/src/arrow/util/span.h
```diff
@@ -0,0 +1,129 @@
...
+      out.size_ = count;
+    }
+    return out;
+  }
+
+  constexpr bool operator==(span const& other) const {
+    if (size_ != other.size_) return false;
+
+    if constexpr (std::is_integral_v<T>) {
+      return std::memcmp(data_, other.data_, size_bytes()) == 0;
```
pitrou, line 97: Note that `memcmp` doesn't accept non-null pointers, even if size is 0. Perhaps we should guard against this (the fallback impl below doesn't have this problem AFAICT).

bkietz: Good catch, I'll add an explicit check for `data_`.

bkietz: ... I'm surprised the sanitizer didn't tell me about this one

pitrou: Perhaps because you don't have tests for the equality operator?

bkietz: Added a test for equality

pitrou:

pitrou:

pitrou:

pitrou (approved):
Made a small improvement to the tests, I'll merge now. Thank you @bkietz !

Issue #36331, opened 2023-06-27
[C++][CI] Sporadic errors in `arrow-acero-asof-join-node-test`


pitrou:
### Describe the bug, including details regarding any error messages, version, and platform.

Various errors seem to occur intermittently on CI.

One example:
```
[ RUN      ] AsofJoinTest.BackpressureWithBatchesGen
/arrow/cpp/src/arrow/acero/asof_join_node_test.cc:1516: Failure
Value of: has_bp_been_applied()
  Actual: false
Expected: true
[  FAILED  ] AsofJoinTest.BackpressureWithBatchesGen (10693 ms)
```
from https://github.com/apache/arrow/actions/runs/5391899018/jobs/9789453616?pr=36215#step:6:3333

Another example:
[logs](https://gist.github.com/pitrou/d1814bde928d3b08fd29a923053a9ed0)
from https://github.com/apache/arrow/actions/runs/5376323460/jobs/9753323965#step:10:338 

### Component(s)

C++, Continuous Integration

pitrou:
@rtpsw @icexelloss

icexelloss:
Thanks @pitrou - I noticed this the other day as well: https://github.com/apache/arrow/issues/36248

We can probably disable this test until we figure out how to make this more robust

kou:
@rtpsw Could you add "take" only comment here to assign you to this issue?

rtpsw:
take

PR #36336, opened 2023-06-28
GH-36329: [C++][CI] Use OpenSSL 3 on macOS

### LINKED ISSUE #36329: [C++][CI] OpenSSL link error in Gandiva on macOS

As seen on CI:
```
Undefined symbols for architecture x86_64:
  "_EVP_MD_get_size", referenced from:
      gandiva::gdv_hash_using_openssl(long long, void const*, unsigned long, evp_md_st const*, unsigned int, int*) in libgandiva.a(hash_utils.cc.o)
ld: symbol(s) not found for architecture x86_64
```
https://github.com/apache/arrow/actions/runs/5390688418/jobs/9786417735#step:9:1716

Also, it would be nice if the CMake configure logs shows the architecture being targeted (perhaps `CMAKE_SYSTEM_PROCESSOR`?).

### Component(s)

C++, Continuous Integration

kou:
### Rationale for this change

GitHub Actions self-hosted runner for macOS has
/usr/local/include/openssl/ provided by OpenSSL 3 (`openssl@3`). Our include paths have `... -isystem /usr/local/include -isystem /usr/local/opt/openssl@1.1/include ...`. It means that `/usr/local/include/openssl/...` is used for `#include <openssl/...>`.

If we mix OpenSSL 3 headers and OpenSSL 1.1 libraries, we may get some problems such as a link error.

This uses OpenSSL 3 instead of OpenSSL 1.1 because GitHub Actions self-hosted runner for macOS provides OpenSSL 3 by /usr/local/include/openssl/. Note that `$(brew --prefix openssl@3)/include` isn't linked as /usr/local/include/openssl` by default. So I think that Homebrew GitHub Actions self-hosted runner for macOS does it explicitly.

Other solution: Unlinking `/usr/local/include/openssl` by `brew unlink openssl@3`. But there is no reason to use OpenSSL 1.1 for us. So this PR doesn't use this solution.

### Are these changes tested?

Yes.

Yes.
* Closes: #36329


kou:
+1

The "C++ / AMD64 macOS 12 C++" is still failing but it's caused by #36331/#36248 .

cpp/cmake_modules/FindOpenSSLAlt.cmake
```diff
@@ -22,17 +22,24 @@ endif()
...
+      set(OPENSSL_ROOT_DIR ${OPENSSL_BREW_PREFIX})
     else()
-      execute_process(COMMAND ${BREW} --prefix "openssl"
-                      OUTPUT_VARIABLE OPENSSL_BREW_PREFIX
+      execute_process(COMMAND ${BREW} --prefix "openssl@3.0"
+                      OUTPUT_VARIABLE OPENSSL3_BREW_PREFIX
                       OUTPUT_STRIP_TRAILING_WHITESPACE)
-      if(OPENSSL_BREW_PREFIX)
-        set(OPENSSL_ROOT_DIR ${OPENSSL_BREW_PREFIX})
+      if(OPENSSL11_BREW_PREFIX)
```
pitrou, line 34: Should this be `OPENSSL3_BREW_PREFIX`?

cpp/cmake_modules/FindOpenSSLAlt.cmake
```diff
@@ -22,17 +22,24 @@ endif()
 if(APPLE AND NOT OPENSSL_ROOT_DIR)
   find_program(BREW brew)
   if(BREW)
-    execute_process(COMMAND ${BREW} --prefix "openssl@1.1"
-                    OUTPUT_VARIABLE OPENSSL11_BREW_PREFIX
+    execute_process(COMMAND ${BREW} --prefix "openssl"
```
pitrou, line 25: I'm curious: why first try "openssl", then "openssl@3.0", then "openssl@1.1"?

Wouldn't "openssl" cover all other cases? I don't know how brew works...

raulcd:

raulcd:
Triggering java-jars because they seem to be failing on the nightlies due to this reason:
```
  "_EVP_MD_get_size", referenced from:
      gandiva::gdv_hash_using_openssl(long long, void const*, unsigned long, evp_md_st const*, unsigned int, int*) in libgandiva.a(unity_3_cxx.cxx.o)
```

raulcd:
There seems to be a lot of failures around `gandiva` on the MacOs java-jars job. I am not sure if they are related:
```
 The following tests FAILED:
	 20 - arrow-compute-scalar-type-test (Failed)
	 38 - arrow-substrait-substrait-test (Failed)
	 44 - arrow-acero-asof-join-node-test (Failed)
	 78 - gandiva-internals-test (Failed)
	 80 - gandiva-filter-test (Failed)
	 81 - gandiva-projector-test (Failed)
	 82 - gandiva-projector-build-validation-test (Failed)
	 83 - gandiva-if-expr-test (Failed)
	 84 - gandiva-literal-test (Failed)
	 85 - gandiva-boolean-expr-test (Failed)
	 86 - gandiva-binary-test (Failed)
	 87 - gandiva-date-time-test (Failed)
	 88 - gandiva-to-string-test (Failed)
	 89 - gandiva-utf8-test (Failed)
	 90 - gandiva-hash-test (Failed)
	 91 - gandiva-in-expr-test (Failed)
	 92 - gandiva-null-validity-test (Failed)
	 93 - gandiva-decimal-test (Failed)
	 94 - gandiva-decimal-single-test (Failed)
	 95 - gandiva-filter-project-test (Failed)
	 96 - gandiva-projector-test-static (Failed)
```

cpp/cmake_modules/FindOpenSSLAlt.cmake
kou: `openssl` is the default OpenSSL formula. In general, it refers the latest OpenSSL formula (`openssl@3.1` now).
If `openssl@3.1` isn't installed, `brew --prefix openssl` is failed.

Then, this process falls back to `openssl@3.0`, then `openssl@1.1`. Because we want to use newer OpenSSL as much as possible.

cpp/cmake_modules/FindOpenSSLAlt.cmake
kou: Ah, yes. Good catch!

Ah, wait. I should have used `OPENSSL30_...` for it.

kou:
`gandiva-*` tests were crashed without backtrace... So we can't debug this. We may be able debug this by `ssh` to the host.

Anyway, how about tracking the failures as a separated issue? Because Gandiva tests in "AMD64 macOS 12 C++" aren't failed.

raulcd:
> Anyway, how about tracking the failures as a separated issue? Because Gandiva tests in "AMD64 macOS 12 C++" aren't failed.

Sounds good to me.

kou:
Created: #36404

I'll merge this.

PR #36338, opened 2023-06-28
GH-36337: [Ruby] Relax required Apache Arrow C++ version

### LINKED ISSUE #36337: [Ruby] Relax required Apache Arrow C++ version

Apache Arrow follows Semantic Versioning. So we can use all Apache Arrow C++ `12.*.*` for Red Arrow 12.0.1. Apache Arrow C GLib `12.*.*` will also work but Red Arrow 12.0.1 may depend on changes for Apache Arrow C GLib 12.0.1. So we should keep depending on the same version of Apache Arrow C GLib.

### Component(s)

Ruby

kou:
### Rationale for this change

Apache Arrow follows Semantic Versioning. So we can use all Apache Arrow C++ `12.*.*` for Red Arrow 12.0.1. Apache Arrow C GLib `12.*.*` will also work but Red Arrow 12.0.1 may depend on changes for Apache Arrow C GLib 12.0.1. So we should keep depending on the same version of Apache Arrow C GLib.

Relax required Apache Arrow C++ version.

### Are these changes tested?

Yes.

Yes.
* Closes: #36337

### DIFF
--- a/ruby/red-arrow/ext/arrow/extconf.rb
+++ b/ruby/red-arrow/ext/arrow/extconf.rb
@@ -41,8 +41,6 @@ end
 unless required_pkg_config_package([
                                      "arrow",
                                      Arrow::Version::MAJOR,
-                                     Arrow::Version::MINOR,
-                                     Arrow::Version::MICRO,
                                    ],
                                    conda: "libarrow",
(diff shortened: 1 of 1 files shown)


kou:
+1

PR #36341, opened 2023-06-28
GH-36340: [Java] Address race condition in allocator logger thread

### LINKED ISSUE #36340: [Java] Flakey test TestBaseAllocator#testMemoryUsage

Test introduced by #35314 is flakey, and result in log messages not being detected. 

In an attempt to debug it by adding the list of log messages, the following output is observed:
```
Error:  Failures: 
Error:    TestBaseAllocator.testMemoryUsage:1127 Log messages are:
[TRACE] Memory Usage: 
null
```

The only possible explanation is that there's a side effect of the logging thread being started before the allocator field of the parent class being set and observable yet

### Component(s)

Java

laurentgo:
### Rationale for this change
Address a race condition where the allocator logger thread starts logging before the class is fully initialized

Change initialization order within the allocator to address the race condition.

Add list of log messages captured during the test run if the assertion failed.

### Are these changes tested?
Yes.

No.

Closes apache/arrow#36340
* Closes: #36340


lidavidm (approved):
Thanks!

PR #36344, opened 2023-06-28
GH-26153: [C++] Share common codes for RecordBatchStreamReader and StreamDecoder

### LINKED ISSUE #26153: [C++] RecordBatchStreamReader should use StreamDecoder
There's no reason to duplicate some of the stream reading logic, and re-using StreamDecoder would ensure the behaviour of both classes matches.

**Reporter**: [Antoine Pitrou](https://issues.apache.org/jira/browse/ARROW-10142) / @pitrou
**Assignee**: [Kouhei Sutou](https://issues.apache.org/jira/browse/ARROW-10142) / @kou

<sub>**Note**: *This issue was originally created as [ARROW-10142](https://issues.apache.org/jira/browse/ARROW-10142). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

kou:
### Rationale for this change

Because they (pull-based and push-based) must have the same behavior.

This PR extracts reusable codes to StreamDecoderInternal from StreamDecoderImpl. External API isn't changed for
RecordBatchStreamReader and StreamDecoder.

This PR adds some external API to implement this:

* arrow::Status::ToStringWithoutContextLines(): This is only for testing. We can get stable result of ASSERT_RAISES_WITH_MESSAGE() with/without -DARROW_EXTRA_ERROR_CONTEXT=ON by this.

  We can extract this and related changes to separated PR if we want.

* arrow::ipc::Listener::OnRecordBatchWithMetadataDecoded(): Because RecordBatchStreamReader wants not only RecordBatch but also custom metadata. OnRecordBatchWithMetadataDecoded() receives RecordBatchWithMetadata. OnRecordBatchDecoded() still exists and it's used by default for backward compatibility.

* arrow::ipc::CollectListener::metadatas(), arrow::ipc::CollectListener::num_record_batches(), arrow::ipc::CollectListener::PopRecordBatch(), arrow::ipc::CollectListener::PopRecordBatchWithMetadat(): If we add these APIs, we can use CollectListner in RecordBatchStreamReader. We can create an internal listener only for RecordBatchStreamReader if don't want to extend CollectListener.

### Are these changes tested?

Yes.

Yes.

`arrow::ipc::CollectListener::record_batches()` returns `const std::vector<std::shared_ptr<RecordBatch>>&` instead of `std::vector<std::shared_ptr<RecordBatch>>`.

* Closes: #26153

### DIFF
--- a/cpp/src/arrow/ipc/reader.cc
+++ b/cpp/src/arrow/ipc/reader.cc
@@ -916,7 +916,7 @@ class StreamDecoderInternal : public MessageDecoderListener {
                                       &out_schema_, &field_inclusion_mask_,
                                       &swap_endian_));
 
-    num_required_initial_dictionaries_ = dictionary_memo_.fields().num_fields();
+    num_required_initial_dictionaries_ = dictionary_memo_.fields().num_dicts();
     num_read_initial_dictionaries_ = 0;
     if (num_required_initial_dictionaries_ == 0) {
       state_ = State::RECORD_BATCHES;
@@ -930,7 +930,7 @@ class StreamDecoderInternal : public MessageDecoderListener {
   Status OnInitialDictionaryMessageDecoded(std::unique_ptr<Message> message) {
     if (message->type() != MessageType::DICTIONARY_BATCH) {
       return Status::Invalid("IPC stream did not have the expected number (",
-                             dictionary_memo_.fields().num_fields(),
+                             num_required_initial_dictionaries_,
                              ") of dictionaries at the start of the stream");
     }
     RETURN_NOT_OK(ReadDictionary(*message));
--- a/cpp/src/arrow/ipc/reader.cc
+++ b/cpp/src/arrow/ipc/reader.cc
@@ -849,88 +849,113 @@ Status ReadDictionary(const Message& message, const IpcReadContext& context,
   return ReadDictionary(*message.metadata(), context, kind, reader.get());
 }
 
-// ----------------------------------------------------------------------
-// RecordBatchStreamReader implementation
-
-class RecordBatchStreamReaderImpl : public RecordBatchStreamReader {
+// Streaming format decoder
+class StreamDecoderInternal : public MessageDecoderListener {
  public:
-  Status Open(std::unique_ptr<MessageReader> message_reader,
-              const IpcReadOptions& options) {
-    message_reader_ = std::move(message_reader);
-    options_ = options;
+  enum State {
+    SCHEMA,
+    INITIAL_DICTIONARIES,
+    RECORD_BATCHES,
+    EOS,
+  };
 
-    // Read schema
-    ARROW_ASSIGN_OR_RAISE(std::unique_ptr<Message> message, ReadNextMessage());
-    if (!message) {
-      return Status::Invalid("Tried reading schema message, was null or length 0");
+  explicit StreamDecoderInternal(std::shared_ptr<Listener> listener,
+                                 IpcReadOptions options)
+      : listener_(std::move(listener)),
+        options_(std::move(options)),
+        state_(State::SCHEMA),
+        field_inclusion_mask_(),
+        num_required_initial_dictionaries_(0),
+        num_read_initial_dictionaries_(0),
+        dictionary_memo_(),
+        schema_(nullptr),
+        out_schema_(nullptr),
+        stats_(),
+        swap_endian_(false) {}
+
+  Status OnMessageDecoded(std::unique_ptr<Message> message) override {
+    ++stats_.num_messages;
+    switch (state_) {
+      case State::SCHEMA:
+        ARROW_RETURN_NOT_OK(OnSchemaMessageDecoded(std::move(message)));
+        break;
+      case State::INITIAL_DICTIONARIES:
+        ARROW_RETURN_NOT_OK(OnInitialDictionaryMessageDecoded(std::move(message)));
+        break;
+      case State::RECORD_BATCHES:
+        ARROW_RETURN_NOT_OK(OnRecordBatchMessageDecoded(std::move(message)));
+        break;
+      case State::EOS:
+        break;
     }
-
-    RETURN_NOT_OK(UnpackSchemaMessage(*message, options, &dictionary_memo_, &schema_,
-                                      &out_schema_, &field_inclusion_mask_,
-                                      &swap_endian_));
     return Status::OK();
   }
 
-  Status ReadNext(std::shared_ptr<RecordBatch>* batch) override {
-    ARROW_ASSIGN_OR_RAISE(auto batch_with_metadata, ReadNext());
-    *batch = std::move(batch_with_metadata.batch);
-    return Status::OK();
+  Status OnEOS() override {
+    state_ = State::EOS;
+    return listener_->OnEOS();
   }
 
-  Result<RecordBatchWithMetadata> ReadNext() override {
-    if (!have_read_initial_dictionaries_) {
-      RETURN_NOT_OK(ReadInitialDictionaries());
-    }
+  Listener* raw_listener() const { return listener_.get(); }
 
-    RecordBatchWithMetadata batch_with_metadata;
-    if (empty_stream_) {
-      // ARROW-6006: Degenerate case where stream contains no data, we do not
-      // bother trying to read a RecordBatch message from the stream
-      return batch_with_metadata;
-    }
+  std::shared_ptr<Schema> schema() const { return out_schema_; }
 
-    // Continue to read other dictionaries, if any
-    std::unique_ptr<Message> message;
-    ARROW_ASSIGN_OR_RAISE(message, ReadNextMessage());
+  ReadStats stats() const { return stats_; }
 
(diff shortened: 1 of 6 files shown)


cpp/src/arrow/ipc/reader.h
```diff
@@ -301,9 +323,39 @@ class ARROW_EXPORT CollectListener : public Listener {
     return record_batches_;
   }
 
+  /// \return the all decoded metadatas
+  std::vector<std::shared_ptr<KeyValueMetadata>> metadatas() const { return metadatas_; }
```
kou, line 327: We may want to use better name for a vector of metadata...

cpp/src/arrow/ipc/reader.cc
```diff
@@ -951,60 +977,85 @@ class RecordBatchStreamReaderImpl : public RecordBatchStreamReader {
...
-  IpcReadOptions options_;
-  std::vector<bool> field_inclusion_mask_;
-
-  bool have_read_initial_dictionaries_ = false;
-
-  // Flag to set in case where we fail to observe all dictionaries in a stream,
-  // and so the reader should not attempt to parse any messages
-  bool empty_stream_ = false;
+  Result<RecordBatchWithMetadata> ReadNext() override {
+    auto collect_listener = static_cast<CollectListener*>(raw_listener());
```
pitrou, line 1021: `checked_cast` here?

cpp/src/arrow/ipc/reader.cc
```diff
@@ -951,60 +977,85 @@ class RecordBatchStreamReaderImpl : public RecordBatchStreamReader {
...
+    while (collect_listener->num_record_batches() == 0) {
```
pitrou, line 1022: Should we also check that `state() != EOS`?

cpp/src/arrow/ipc/reader.h
```diff
@@ -301,9 +317,39 @@ class ARROW_EXPORT CollectListener : public Listener {
...
+
+  /// \return the number of collected record batches
+  size_t num_record_batches() const { return record_batches_.size(); }
```
pitrou, line 324: Return `int64_t` here?

cpp/src/arrow/ipc/reader.h
```diff
@@ -301,9 +317,39 @@ class ARROW_EXPORT CollectListener : public Listener {
...
+  /// remove it from record_batches
+  RecordBatchWithMetadata PopRecordBatchWithMetadata() {
+    RecordBatchWithMetadata record_batch_with_metadata;
+    if (record_batches_.empty()) {
+      return record_batch_with_metadata;
+    }
+    record_batch_with_metadata.batch =
+        std::move(record_batches_[record_batches_.size() - 1]);
+    record_batch_with_metadata.custom_metadata =
+        std::move(metadatas_[metadatas_.size() - 1]);
```
pitrou, line 343:
```suggestion
    record_batch_with_metadata.batch = std::move(record_batches_.back());
    record_batch_with_metadata.custom_metadata = std::move(metadatas_.back());
```

cpp/src/arrow/ipc/reader.h
```diff
@@ -251,7 +251,8 @@ class ARROW_EXPORT Listener {
   /// \see StreamDecoder
   virtual Status OnEOS();
 
-  /// \brief Called when a record batch is decoded.
+  /// \brief Called when a record batch is decoded and
+  /// OnReocrdBatchDecoded() isn't overrided.
```
pitrou, line 255:
```suggestion
  /// \brief Called when a record batch is decoded and
  /// OnRecordBatchWithMetadataDecoded() isn't overrided.
```

cpp/src/arrow/ipc/reader.h
pitrou: I think this is ok.

cpp/src/arrow/ipc/reader.h
```diff
@@ -301,9 +317,39 @@ class ARROW_EXPORT CollectListener : public Listener {
     return record_batches_;
   }
 
+  /// \return the all decoded metadatas
+  std::vector<std::shared_ptr<KeyValueMetadata>> metadatas() const { return metadatas_; }
```
pitrou, line 321: But why is this returning a copy? It could return a const-ref. Same for `record_batches()`...

cpp/src/arrow/ipc/reader.cc
```diff
@@ -951,60 +977,85 @@ class RecordBatchStreamReaderImpl : public RecordBatchStreamReader {
...
-        }
-      }
+  std::shared_ptr<Listener> listener_;
+  const IpcReadOptions options_;
+  State state_;
+  std::vector<bool> field_inclusion_mask_;
+  int num_required_initial_dictionaries_;
+  int num_read_initial_dictionaries_;
+  DictionaryMemo dictionary_memo_;
+  std::shared_ptr<Schema> schema_;
```
pitrou, line 987: Is `schema_` used anywhere?

pitrou (commented):
Nice!

cpp/src/arrow/ipc/reader.h
kou: Thanks! It's what I want!

cpp/src/arrow/ipc/reader.cc
kou: https://github.com/apache/arrow/pull/36344/files#diff-e992169684aea9845ac776ada4cbb2b5dc711b49e5a3fbc6046c92299e1aefceR940 and https://github.com/apache/arrow/pull/36344/files#diff-e992169684aea9845ac776ada4cbb2b5dc711b49e5a3fbc6046c92299e1aefceR954 .
(I understand that this PR's diff is difficult to review...)

cpp/src/arrow/ipc/reader.cc
kou: Good catch!

cpp/src/arrow/ipc/reader.cc
kou: Ah, it's better.

cpp/src/arrow/ipc/reader.h
kou: Oh...

cpp/src/arrow/ipc/reader.h
kou: Ah, I just used the same signature as `record_batches()` but we should use const-ref.
I can't remember why I didn't use const-ref for `record_batches()` but I just doesn't notice it...

I also use const-ref for `record_batches()` too. It breaks a backward compatibility but I hope that nobody doesn't use this.

cpp/src/arrow/ipc/reader.h
kou: Ah, yes. Other places use `int64_t` or `int` for the number of record batches.

kou:
The integration test failures seems related:

https://github.com/apache/arrow/actions/runs/5407939797/jobs/9826488521?pr=36344#step:7:14040

```text
################# FAILURES #################
FAILED TEST: nested_dictionary Rust producing,  C++ consuming

FAILED TEST: shared_dict C++ producing,  C++ consuming
```

kou:
The integration test failures are fixed.

pitrou:
It seems like Homebrew is broken currently? @kou

cpp/src/arrow/ipc/reader.cc
pitrou: Ahh... Isn't it error-prone that `schema()` returns the filtered schema (`out_schema_`) but `OnSchemaDecoded` is called with the unfiltered schema?

kou:
> It seems like Homebrew is broken currently?

I've fixed it by #36403 and rebased on main.

cpp/src/arrow/ipc/reader.cc
kou: I can understand your concern.

How about adding `Listener::OnSchemaDecoded(std::shared_ptr<Schema> schema, std::shared_ptr<Schema> filtered_schema)` and it calls `Listener::OnSchemaDecoded(std::shared_ptr<Schema> schema)` by default for keeping backward compatibility? (Is `filtered_schema` name good? I think that `out_schema` is difficult to understand.)

If we want to do it, I want to do it in a separated pull request instead of including it to this pull request. Because it's not related to this pull request.

kou:
AppVeyor failure: #36405

cpp/src/arrow/ipc/reader.cc
pitrou: That's a good idea, thanks for suggesting it.

kou: OK. I've opened #36407 for it.

pitrou:
The R failures look unexpected. @paleolimbot @thisisnic Any clue about them?

thisisnic:
R failures are unrelated; the fix is in #36397

kou:
I think that #36381 introduced it and it's tracked by #36396.

kou:
#36397 was merged and I've rebased on main.

kou:
The R failures have gone.
I'll merge this.

PR #36358, opened 2023-06-28
GH-36345: [C++] Prefer TypeError over Invalid in IsIn and IndexIn kernels

### LINKED ISSUE #36345: [C++][Compute] Prefer TypeError over Invalid in IndexIn and IsIn kernels

When input type and value_set type miss match, TypeError should be used instead of Invalid.

### Component(s)

C++

js8544:
### Rationale for this change

`TypeError` should be returned when `values` and `value_set` have incompatible types in IsIn and IndeIx.
`Invalid` is still returned if the types are compatible but casting the values fails (for example because of overflow or truncation).

When casting between types is not supported, return TypeError instead of Invalid.

### Are these changes tested?

Yes.

No.

* Closes: #36345

### DIFF
--- a/cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
+++ b/cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
@@ -209,7 +209,7 @@ struct InitStateVisitor {
       const auto& ty1 = checked_cast<const TimestampType&>(*arg_type);
       const auto& ty2 = checked_cast<const TimestampType&>(*options.value_set.type());
       if (ty1.timezone().empty() ^ ty2.timezone().empty()) {
-        return Status::Invalid(
+        return Status::TypeError(
             "Cannot compare timestamp with timezone to timestamp without timezone, got: ",
             ty1, " and ", ty2);
       }
@@ -218,8 +218,8 @@ struct InitStateVisitor {
       // This is a bit of a hack, but don't implicitly cast from a non-binary
       // type to string, since most types support casting to string and that
       // may lead to surprises. However, we do want most other implicit casts.
-      return Status::Invalid("Array type didn't match type of values set: ", *arg_type,
-                             " vs ", *options.value_set.type());
+      return Status::TypeError("Array type didn't match type of values set: ", *arg_type,
+                               " vs ", *options.value_set.type());
     }
 
     if (!options.value_set.is_arraylike()) {
@@ -236,12 +236,12 @@ struct InitStateVisitor {
         if ((options.value_set.type()->id() == Type::STRING ||
              options.value_set.type()->id() == Type::LARGE_STRING) &&
             !is_base_binary_like(arg_type.id())) {
-          return Status::Invalid("Array type didn't match type of values set: ",
-                                 *arg_type, " vs ", *options.value_set.type());
+          return Status::TypeError("Array type didn't match type of values set: ",
+                                   *arg_type, " vs ", *options.value_set.type());
         }
       } else {
-        return Status::Invalid("Array type doesn't match type of values set: ", *arg_type,
-                               " vs ", *options.value_set.type());
+        return Status::TypeError("Array type doesn't match type of values set: ",
+                                 *arg_type, " vs ", *options.value_set.type());
       }
     }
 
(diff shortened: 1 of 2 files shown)


js8544:
cc @pitrou

cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
```diff
@@ -326,9 +326,13 @@ struct IndexInVisitor {
     const auto& state = checked_cast<const SetLookupState<Type>&>(*ctx->state());
     if (!data.type->Equals(state.value_set_type)) {
       auto materialized_input = data.ToArrayData();
-      ARROW_ASSIGN_OR_RAISE(auto casted_input,
-                            Cast(*materialized_input, state.value_set_type,
-                                 CastOptions::Safe(), ctx->exec_context()));
+      auto cast_result = Cast(*materialized_input, state.value_set_type,
+                              CastOptions::Safe(), ctx->exec_context());
+      if (ARROW_PREDICT_FALSE(!cast_result.ok())) {
```
pitrou, line 331: Well, we want to keep the original error. This is precisely to distinguish the different kinds of errors:
* `TypeError`: casting from type X to type Y is not supported at all
* `Invalid`: casting from type X to type Y _is_ supported, but the actual values fail casting (e.g. because of truncation or overflow)

cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
```diff
@@ -423,11 +427,15 @@ struct IsInVisitor {
...
     if (!data.type->Equals(state.value_set_type)) {
-      auto materialized_data = data.ToArrayData();
-      ARROW_ASSIGN_OR_RAISE(auto casted_data,
-                            Cast(*materialized_data, state.value_set_type,
-                                 CastOptions::Safe(), ctx->exec_context()));
-      return ProcessIsIn(state, *casted_data.array());
+      auto materialized_input = data.ToArrayData();
+      auto cast_result = Cast(*materialized_input, state.value_set_type,
+                              CastOptions::Safe(), ctx->exec_context());
+      if (ARROW_PREDICT_FALSE(!cast_result.ok())) {
```
pitrou, line 433: Same here of course.

cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
pitrou: Note that `Cast` also does the distinction, with the caveat that it returns `NotImplemented` instead of `TypeError` (which IMHO is not great):
```python
>>> a = pa.array([[1000,2000], [3000]])
>>> a.type
ListType(list<item: int64>)
>>> pc.cast(a, pa.list_(pa.int32()))
<pyarrow.lib.ListArray object at 0x7f1ccf65ab60>
[
  [
    1000,
    2000
  ],
  [
    3000
  ]
]
>>> pc.cast(a, pa.list_(pa.int8()))
[...]
ArrowInvalid: Integer value 1000 not in range: -128 to 127
>>> pc.cast(a, pa.utf8())
[...]
ArrowNotImplementedError: Unsupported cast from list<item: int64> to utf8 using function cast_string
```

pitrou: So you could perhaps refine this as:
```c++
      if (!cast_result.ok()) {
        if (cast_result.IsNotImplemented()) {
          return Status::TypeError("Array type didn't match type of values set: ",
                                   *data.type, " vs ", *state.value_set_type);
        }
        return cast_result;
      }
```

pitrou (changes requested):

cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
js8544: Done.

cpp/src/arrow/compute/kernels/scalar_set_lookup.cc
js8544: Done.

pitrou (approved):

PR #36362, opened 2023-06-28
MINOR: [Docs] add note about how comment bot assigns users

anjakefala:
### Rationale for this change

Each time I wanted to use it, I had to look up the syntax in the github workflows. Most new users wouldn't know about it existing. 

https://github.com/apache/arrow/blob/e5de6a59f410a3255cc84138b44fc5802b627afc/.github/workflows/comment_bot.yml#L182

Doc addition of text teaching about "take".

Nope

### DIFF
--- a/docs/source/developers/guide/step_by_step/finding_issues.rst
+++ b/docs/source/developers/guide/step_by_step/finding_issues.rst
@@ -69,8 +69,7 @@ in the comments.
    When you find a GitHub issue you would like to work on, please mention
    your interest in the comment section of that issue; that way we will know
    you are working on it.
-   To assign yourself to the issue type 'take' (and nothing else) in a comment
-   on that issue.
+   Consider assigning yourself to the issue (:ref:`issue-assignment`) when the work starts.
 
 Also, do not hesitate to ask questions in the comment. You can get some
 pointers about where to start and similar issues already solved.
--- a/docs/source/developers/guide/step_by_step/finding_issues.rst
+++ b/docs/source/developers/guide/step_by_step/finding_issues.rst
@@ -69,6 +69,8 @@ in the comments.
    When you find a GitHub issue you would like to work on, please mention
    your interest in the comment section of that issue; that way we will know
    you are working on it.
+   To assign yourself to the issue type 'take' (and nothing else) in a comment
+   on that issue.
 
 Also, do not hesitate to ask questions in the comment. You can get some
 pointers about where to start and similar issues already solved.


docs/source/developers/guide/step_by_step/finding_issues.rst
```diff
@@ -69,6 +69,8 @@ in the comments.
    When you find a GitHub issue you would like to work on, please mention
    your interest in the comment section of that issue; that way we will know
    you are working on it.
+   To assign yourself to the issue type 'take' (and nothing else) in a comment
+   on that issue.
```
kou, line 73: We have a note for `take` at https://arrow.apache.org/docs/dev/developers/bug_reports.html#issue-assignment .
How about linking to the section instead of writing the same content here?

anjakefala: Oh, that's where it is! Sure.

anjakefala: (I pushed a change, but I managed to build docs locally, and it does not work as I expected. I need to understand how linking to a subdomain on a seperate page works. `:ref:` might work. I will move to "ready for review" when it is building correctly.

anjakefala:
@kou The link was added! I tested it locally, and it works.

docs/source/developers/guide/step_by_step/finding_issues.rst
```diff
@@ -69,6 +69,7 @@ in the comments.
    When you find a GitHub issue you would like to work on, please mention
    your interest in the comment section of that issue; that way we will know
    you are working on it.
+   Consider `assigning yourself to the issue <../../bug_reports.html#issue-assignment>`_ when the work starts.
```
kou, line 72: Can we use `:ref:...` markup here?

```suggestion
   Consider :ref:`assigning yourself to the issue <issue-assignment>`_ when the work starts.
```

anjakefala: @kou The way you wrote it didn't work out, it doesn't seem to be valid rst markup? I followed the syntax [here](https://devguide.python.org/documentation/markup/) and made the change.

anjakefala:
Ready for review!

mapleFU (approved):

docs/source/developers/guide/step_by_step/finding_issues.rst
kou: Oh, sorry.
I thought that it works. But I'm not an expert of reST.

FYI:

https://devguide.python.org/documentation/markup/#cross-linking-markup

> Alternatively, you can reference any label (not just section titles) if you provide the link text `` :ref:`link text <reference-label>` ``.

kou (approved):
+1

docs/source/developers/guide/step_by_step/finding_issues.rst
anjakefala: Ahhhh, so the `_` was the problem.

kou: Ah, I should have removed it!

PR #36364, opened 2023-06-28
GH-36353: [R] Fix package version references to be text only and never numeric

### LINKED ISSUE #36353: [R] Fix package version references to be text only and never numeric

I received the following email from CRAN today:

```
Dear maintainers,

This concerns the CRAN packages

  Ternary V8 arrow circumplex dabestr dm downlit ggplot2 haven igraph
  isopam jsonlite jtools maketools nomnoml opencpu origin pROC
  plainview readxl rmarkdown runjags sf snakecase stringstatic
  supernova testdat testthat useful usethis vetiver viridis

A recent PR (<https://bugs.r-project.org/show_bug.cgi?id=18548>) showed
that whereas version inputs to numeric_version() and package_version()
were always documented to be "suitable character vectors" (or numeric
vesion objects), the implementation also accepted numeric inputs via
as.character() coercion, the results of which however are (at least
currently) dependent on the setting of the OutDec option.  Given that
numeric input is certainly error-prone, imagine

  as.numeric_version(1.20)
  as.numeric_version(1.5-1)

R Core has decided it would be best to gradually phase out such inputs.

Thus, With c84610 R-devel now *warns* about such inputs, which will
typically not show in the check results (but immediately giving errors
would cause too much disruption).  Some of these warnings can be found
by code analyis.  I list these uses below, can you please fix as
necessary?

When doing so, please check the check outputs for such warnings, e.g. by

  grep -r 'invalid non-character version specification' PKGNAME.Rcheck

(One can for now set env var _R_CALLS_INVALID_NUMERIC_VERSION_ to
something non-empty to get the call stacks for the warnings; this may
need increasing option warning.length from its default.)

Please fix before 2023-07-12.

Please note that CRAN will be closed for submissions between 2023-07-21
2023-08-08.

Best
-k

***********************************************************************
$arrow
$arrow$`arrow/R/duckdb.R`
$arrow$`arrow/R/duckdb.R`[[1]]
getRversion() >= 4
```

Instead of doing a patch release, we can fold this in with the 13.0.0 release (preferably ahead of the CRAN summer holidays), but we should fix this and perhaps add a check to a CI job so we don't fall foul if this by accident in future.

### Component(s)

R

thisisnic:
Fixes #36353
* Closes: #36353

### DIFF
--- a/r/R/duckdb.R
+++ b/r/R/duckdb.R
@@ -95,7 +95,7 @@ run_duckdb_examples <- function() {
     packageVersion("duckdb") > "0.2.7" &&
     requireNamespace("dplyr", quietly = TRUE) &&
     requireNamespace("dbplyr", quietly = TRUE) &&
-    getRversion() >= 4
+    getRversion() >= "4"
 }
 
 # Adapted from dbplyr


paleolimbot (approved):
Thanks!

PR #36365, opened 2023-06-28
GH-36384: [Go] Schema: NumFields

### LINKED ISSUE #36384: [Go] Expose number of fields in schema

Previously if one wanted to iterate over the fields in the schema you would call the Fields() function and just iterate over the slice. However, due to [this commit](https://github.com/rtpsw/arrow/commit/802674b73c94c84388a6775b424ebe4f6e04274e) there is now an allocation and copy that happens when that's called. So to iterate over the fields without allocations one now must use the Field(i int) method; however that means a user must already know exactly how many fields are in the schema which isn't possible today.

It would be idea if there were a `NumFields() int` method on schemas allowing users to iterate over the fields without copying.

### Component(s)

Go

thorfour:
### Rationale for this change

Previously if one wanted to iterate over the fields in the schema you would call the `Fields()` function and just iterate over the slice. However, due to [this commit](https://github.com/rtpsw/arrow/commit/802674b73c94c84388a6775b424ebe4f6e04274e) there is now an allocation and copy that happens when that's called. So to iterate over the fields without allocations one now must use the `Field(i int)` method; however that means a user must already know exactly how many fields are in the schema which isn't possible today.

This adds a simple `NumFields() int` method that returns the number of fields in a schema to allow a user to iterate over all the fields without having to copy them. 

Expose the number of fields in a schema for iteration over fields.

Single function added `NumFields() int` to schema

### Are these changes tested?

N/A

Yes this is a new API

* Closes: #36384

### DIFF
--- a/go/arrow/schema.go
+++ b/go/arrow/schema.go
@@ -194,12 +194,13 @@ func (sc *Schema) WithEndianness(e endian.Endianness) *Schema {
 func (sc *Schema) Endianness() endian.Endianness { return sc.endianness }
 func (sc *Schema) IsNativeEndian() bool          { return sc.endianness == endian.NativeEndian }
 func (sc *Schema) Metadata() Metadata            { return sc.meta }
-func (sc *Schema) Fields() []Field { 
+func (sc *Schema) Fields() []Field {
 	fields := make([]Field, len(sc.fields))
 	copy(fields, sc.fields)
 	return fields
 }
-func (sc *Schema) Field(i int) Field             { return sc.fields[i] }
+func (sc *Schema) Field(i int) Field { return sc.fields[i] }
+func (sc *Schema) NumFields() int    { return len(sc.fields) }
 
 func (sc *Schema) FieldsByName(n string) ([]Field, bool) {
 	indices, ok := sc.index[n]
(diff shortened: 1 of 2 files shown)


kou:
Could you open a new issue for non-MINOR change instead of using `MINOR: ` title?
See also our MINOR definition: https://github.com/apache/arrow/blob/main/CONTRIBUTING.md#Minor-Fixes

thorfour:
> Could you open a new issue for non-MINOR change instead of using `MINOR: ` title? See also our MINOR definition: https://github.com/apache/arrow/blob/main/CONTRIBUTING.md#Minor-Fixes

Ah my bad, missed that. Will open

thorfour:
This addresses #36384

go/arrow/schema.go
```diff
@@ -194,12 +194,13 @@ func (sc *Schema) WithEndianness(e endian.Endianness) *Schema {
...
 func (sc *Schema) Metadata() Metadata            { return sc.meta }
-func (sc *Schema) Fields() []Field { 
+func (sc *Schema) Fields() []Field {
 	fields := make([]Field, len(sc.fields))
 	copy(fields, sc.fields)
 	return fields
 }
-func (sc *Schema) Field(i int) Field             { return sc.fields[i] }
+func (sc *Schema) Field(i int) Field { return sc.fields[i] }
+func (sc *Schema) NumFields() int    { return len(sc.fields) }
```
zeroshade, line 203: Can you add a small unit test for this?

thorfour: added

zeroshade (approved):

PR #36368, opened 2023-06-28
GH-36053:  [C++] summarizing a variable results in NA at random, while there is no NA in the subset of data

### LINKED ISSUE #36053: [C++][R] summarizing a variable results in NA at random, while there is no NA in the subset of data.

```
td <- tempdir()
tzip <- file.path(td, "reprex.zip")
download.file("https://drive.google.com/uc?export=download&id=1-KefpiALDtUg0PrCUgpMAaE0903jVWWm", destfile = tzip)
unzip(tzip, exdir = td)
tlink <- file.path(td, "co_ano_mes=1997-01-01")
library(dplyr)
arrow_dset <- arrow::open_dataset(
     tlink,
     format = "parquet"
)
arrow_dset%>%
count(fluxo, vl_frete_miss=is.na(vl_frete))%>%collect()
```

No missing values for vl_frete when fluxo=="imp"

>   fluxo vl_frete_miss     n   
> <chr> <lgl>         <int>  
> exp   TRUE   35546  
> imp   FALSE         42332

```
replicate(10,  arrow_dset%>% 
  group_by(fluxo) %>% 
  summarise(vl_frete=sum(vl_frete)) %>% 
  collect %>% 
  filter(fluxo=="imp") %>%
  pull(vl_frete))
```

>      [1]        NA        NA 154149785        NA 154149785        NA 154149785 154149785
>      [9] 154149785 154149785

### Component(s)

R

westonpace:
### Rationale for this change

When merging two aggregate states we were failing to use the correct `no_nulls` field.  This field tells us whether we should return `null` if `skip_nulls=False` (if `no_nulls` is false then we return null).

Since we were reading the wrong field we would sometimes emit null even when a column didn't actually have any nulls.

Fixed the bug.

### Are these changes tested?

Yes, I added a new unit test that reproduced this failure quite reliably.

No.
* Closes: #36053

### DIFF
--- a/cpp/src/arrow/acero/aggregate_node_test.cc
+++ b/cpp/src/arrow/acero/aggregate_node_test.cc
@@ -22,9 +22,13 @@
 
 #include <memory>
 
+#include "arrow/acero/test_util_internal.h"
+#include "arrow/compute/api_aggregate.h"
 #include "arrow/result.h"
 #include "arrow/table.h"
 #include "arrow/testing/gtest_util.h"
+#include "arrow/util/bit_util.h"
+#include "arrow/util/string.h"
 
 namespace arrow {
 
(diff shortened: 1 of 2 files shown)


bkietz (approved):
LGTM

PR #36372, opened 2023-06-29
GH-36369: [C++][FlightRPC] Fix a hang bug in FlightClient::Authenticate*()

### LINKED ISSUE #36369: [C++][FlightRPC] arrow::flight::FlightClient::AuthenticateBasicToken is blocked against Ballista

Ballista uses `admin`/`password` for (Basic) authentication: https://arrow.apache.org/ballista/user-guide/flightsql.html#a-name-tool-use-the-driver-in-your-favorite-data-tool

The following Apache Arrow C++ Flight RPC client is blocked: 

```cpp
#include <iostream>

#include <arrow/flight/client.h>

static arrow::Status
run(void)
{
  ARROW_ASSIGN_OR_RAISE(auto location,
                        arrow::flight::Location::Parse("grpc://127.0.0.1:50050"));
  arrow::flight::FlightClientOptions options;
  ARROW_ASSIGN_OR_RAISE(auto client,
                        arrow::flight::FlightClient::Connect(location, options));
  arrow::flight::FlightCallOptions call_options;
  ARROW_ASSIGN_OR_RAISE(auto bearer_token,
                        client->AuthenticateBasicToken(call_options, "admin", "password"));
  return arrow::Status::OK();
}

extern "C" int
main(void)
{
  auto status = run();
  if (!status.ok()) {
    std::cout << status.ToString() << std::endl;
    return 1;
  }
  return 0;
}
```

`stream->Finish()` in `GrpcClientImpl::AuthenticateBasicToken()` is blocked:

https://github.com/apache/arrow/blob/main/cpp/src/arrow/flight/transport/grpc/grpc_client.cc#L762

[`grpc::ClientReaderWriter< W, R >::Finish` document](https://grpc.github.io/grpc/cpp/classgrpc_1_1_client_reader_writer.html#a0fe28c9fa3d01f716c9e453245b69b17) refers [`grpc::internal::ClientStreamingInterface::Finish()` document](https://grpc.github.io/grpc/cpp/classgrpc_1_1internal_1_1_client_streaming_interface.html#a567f213cc0a5df2a03becda3e5711e25) and it says:

> It is appropriate to call this method exactly once when both:
>
>  * the calling code (client-side) has no more message to send (this can be declared implicitly by calling this method, or explicitly through an earlier call to WritesDone method of the class in use, e.g. [ClientWriterInterface::WritesDone](https://grpc.github.io/grpc/cpp/classgrpc_1_1_client_writer_interface.html#aff19574252338e9ac1b5446e82ed8ac5) or [ClientReaderWriterInterface::WritesDone](https://grpc.github.io/grpc/cpp/classgrpc_1_1_client_reader_writer_interface.html#a52f4e5d5ac7fe0e4995cb337aa0ecfc8)).
>  * there are no more messages to be received from the server (which can be known implicitly, or explicitly from an earlier call to [ReaderInterface::Read](https://grpc.github.io/grpc/cpp/classgrpc_1_1internal_1_1_reader_interface.html#adfb50ace59b4b9666a46548d3f732ee8) that returned "false").

We call `stream->WritesDone()` explicitly for the former. But we don't call `stream->Read()` until it returns `false`. Should we call `stream->Read()` before we call `stream->Finish()` like the following?

```diff
diff --git a/cpp/src/arrow/flight/transport/grpc/grpc_client.cc b/cpp/src/arrow/flight/transport/grpc/grpc_client.cc
index 5abecd91a..b382d1e11 100644
--- a/cpp/src/arrow/flight/transport/grpc/grpc_client.cc
+++ b/cpp/src/arrow/flight/transport/grpc/grpc_client.cc
@@ -740,6 +740,15 @@ class GrpcClientImpl : public internal::ClientTransport {
     RETURN_NOT_OK(auth_handler_->Authenticate(&outgoing, &incoming));
     // Explicitly close our side of the connection
     bool finished_writes = stream->WritesDone();
+    pb::HandshakeResponse response;
+    if (!stream->Read(&response)) {
+      return MakeFlightError(FlightStatusCode::Internal,
+                             "No handshake response from server");
+    }
+    if (stream->Read(&response)) {
+      return MakeFlightError(FlightStatusCode::Internal,
+                             "Too much handshake response from server");
+    }
     RETURN_NOT_OK(FromGrpcStatus(stream->Finish(), &rpc.context));
     if (!finished_writes) {
       return MakeFlightError(FlightStatusCode::Internal,
@@ -759,6 +768,15 @@ class GrpcClientImpl : public internal::ClientTransport {
         stream = stub_->Handshake(&rpc.context);
     // Explicitly close our side of the connection.
     bool finished_writes = stream->WritesDone();
+    pb::HandshakeResponse response;
+    if (!stream->Read(&response)) {
+      return MakeFlightError(FlightStatusCode::Internal,
+                             "No handshake response from server");
+    }
+    if (stream->Read(&response)) {
+      return MakeFlightError(FlightStatusCode::Internal,
+                             "Too much handshake response from server");
+    }
     RETURN_NOT_OK(FromGrpcStatus(stream->Finish(), &rpc.context));
     if (!finished_writes) {
       return MakeFlightError(FlightStatusCode::Internal,
```

If we call `stream->Read()` before `stream->Finish()`, the test program isn't blocked.

@lidavidm What do you think about this?

### Component(s)

C++, FlightRPC

kou:
### Rationale for this change

We need to drain all responses from server before we call gRPC's Finish() to avoid hanging.

Read all responses before calling Finish().

### Are these changes tested?

Yes.

Yes.
* Closes: #36369

### DIFF
--- a/cpp/src/arrow/flight/transport/grpc/grpc_client.cc
+++ b/cpp/src/arrow/flight/transport/grpc/grpc_client.cc
@@ -732,52 +732,16 @@ class GrpcClientImpl : public internal::ClientTransport {
                       std::unique_ptr<ClientAuthHandler> auth_handler) override {
     auth_handler_ = std::move(auth_handler);
     ClientRpc rpc(options);
-    std::shared_ptr<
-        ::grpc::ClientReaderWriter<pb::HandshakeRequest, pb::HandshakeResponse>>
-        stream = stub_->Handshake(&rpc.context);
-    GrpcClientAuthSender outgoing{stream};
-    GrpcClientAuthReader incoming{stream};
-    RETURN_NOT_OK(auth_handler_->Authenticate(&outgoing, &incoming));
-    // Explicitly close our side of the connection
-    bool finished_writes = stream->WritesDone();
-    if (!finished_writes) {
-      return MakeFlightError(FlightStatusCode::Internal,
-                             "Could not finish writing before closing");
-    }
-    // Drain the read side, as otherwise gRPC Finish() will hang. We
-    // only call Finish() when the client closes the writer or the
-    // reader finishes, so it's OK to assume the client no longer
-    // wants to read and drain the read side.
-    pb::HandshakeResponse response;
-    while (stream->Read(&response)) {
-    }
-    RETURN_NOT_OK(FromGrpcStatus(stream->Finish(), &rpc.context));
-    return Status::OK();
+    return AuthenticateInternal(rpc);
   }
 
   arrow::Result<std::pair<std::string, std::string>> AuthenticateBasicToken(
       const FlightCallOptions& options, const std::string& username,
       const std::string& password) override {
-    // Add basic auth headers to outgoing headers.
     ClientRpc rpc(options);
+    // Add basic auth headers to outgoing headers.
     AddBasicAuthHeaders(&rpc.context, username, password);
-    std::shared_ptr<
-        ::grpc::ClientReaderWriter<pb::HandshakeRequest, pb::HandshakeResponse>>
-        stream = stub_->Handshake(&rpc.context);
-    // Explicitly close our side of the connection.
-    bool finished_writes = stream->WritesDone();
-    if (!finished_writes) {
-      return MakeFlightError(FlightStatusCode::Internal,
-                             "Could not finish writing before closing");
-    }
-    // Drain the read side, as otherwise gRPC Finish() will hang. We
-    // only call Finish() when the client closes the writer or the
-    // reader finishes, so it's OK to assume the client no longer
-    // wants to read and drain the read side.
-    pb::HandshakeResponse response;
-    while (stream->Read(&response)) {
-    }
-    RETURN_NOT_OK(FromGrpcStatus(stream->Finish(), &rpc.context));
+    RETURN_NOT_OK(AuthenticateInternal(rpc));
     // Grab bearer token from incoming headers.
     return GetBearerTokenHeader(rpc.context);
   }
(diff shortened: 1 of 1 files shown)


cpp/src/arrow/flight/transport/grpc/grpc_client.cc
```diff
@@ -740,11 +740,24 @@ class GrpcClientImpl : public internal::ClientTransport {
...
                              "Could not finish writing before closing");
     }
+    // Drain the read side, as otherwise gRPC Finish() will hang. We
+    // only call Finish() when the client closes the writer or the
+    // reader finishes, so it's OK to assume the client no longer
+    // wants to read and drain the read side.
+    pb::HandshakeResponse response;
+    if (!stream->Read(&response)) {
+      return MakeFlightError(FlightStatusCode::Internal,
+                             "No handshake response from server");
```
lidavidm, line 754: Do we expect another response? I think the auth_handler_ should've read everything.

This could probably just be

```cpp
while (stream->Read(&response)) {}
```

here and below.

I think it's OK to either ignore all extra responses, or return an error.

kou: OK. I've simplified this.

lidavidm (approved):

cpp/src/arrow/flight/transport/grpc/grpc_client.cc
```diff
@@ -740,11 +740,18 @@ class GrpcClientImpl : public internal::ClientTransport {
...
                              "Could not finish writing before closing");
     }
+    // Drain the read side, as otherwise gRPC Finish() will hang. We
+    // only call Finish() when the client closes the writer or the
+    // reader finishes, so it's OK to assume the client no longer
+    // wants to read and drain the read side.
+    pb::HandshakeResponse response;
+    while (stream->Read(&response)) {
+    }
+    RETURN_NOT_OK(FromGrpcStatus(stream->Finish(), &rpc.context));
```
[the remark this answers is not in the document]
kou: > Can we perhaps factor out all this in a helper method?

Done.

> (also, why do we ignore any read response?)

For `Authenticate()`, user defined `auth_handler_->Authenticate()` may or may not read a response from a server. If `auth_handler_->Authenticate()` isn't read a response, we can read it here. But we can't do nothing here for the read response. So we can just ignore the read response.

For `AuthenticateBasicToken()`, we can read one response from a server here. But we don't need to use the response for basic token authentication. So we can just ignore the read response.

A server may return multiple responses but we don't need them here. So we can ignore them too.

thisisnic:
The remaining R failures will be fixed when #36397 is merged

lidavidm (approved):

kou:
Green.
I'll merge this.

PR #36373, opened 2023-06-29
GH-36154: [JS][CI] Use `jest` cache in CI

### LINKED ISSUE #36154: [JS][CI] Timeout on `macos` in GitHub Actions

https://github.com/apache/arrow/actions/runs/5097749959/jobs/9164462790

`macos` in `name: NodeJS` times out.

```
AMD64 macOS 11 NodeJS 18
The job running on runner GitHub Actions 25 has exceeded the maximum execution time of 60 minutes.
```

### Component(s)

JavaScript

abetomo:
### Rationale for this change

Use the cache of `test` to reduce test time.

Use `jest` cache with `actions/cache@v3`.

### Are these changes tested?

In GitHub Actions, test time was reduced.

* Closes: #36154

### DIFF
--- a/.github/workflows/js.yml
+++ b/.github/workflows/js.yml
@@ -94,10 +94,8 @@ jobs:
         uses: actions/cache@v3
         with:
           path: js/.jest-cache
-          key: js-jest-cache-${{ runner.os }}-${{ matrix.node }}-${{ hashFiles('js/yarn.lock') }}
-          restore-keys: |
-            js-jest-cache-${{ runner.os }}-${{ matrix.node }}-
-            js-jest-cache-${{ runner.os }}-
+          key: js-jest-cache-${{ runner.os }}-${{ hashFiles('js/yarn.lock') }}
+          restore-keys: js-jest-cache-${{ runner.os }}-
       - name: Install NodeJS
         uses: actions/setup-node@v3
         with:
@@ -126,10 +124,8 @@ jobs:
         uses: actions/cache@v3
         with:
           path: js/.jest-cache
-          key: js-jest-cache-${{ runner.os }}-${{ matrix.node }}-${{ hashFiles('js/yarn.lock') }}
-          restore-keys: |
-            js-jest-cache-${{ runner.os }}-${{ matrix.node }}-
-            js-jest-cache-${{ runner.os }}-
+          key: js-jest-cache-${{ runner.os }}-${{ hashFiles('js/yarn.lock') }}
+          restore-keys: js-jest-cache-${{ runner.os }}-
       - name: Install NodeJS
         uses: actions/setup-node@v3
         with:
--- a/.github/workflows/js.yml
+++ b/.github/workflows/js.yml
@@ -80,7 +80,7 @@ jobs:
     name: AMD64 macOS 11 NodeJS ${{ matrix.node }}
     runs-on: macos-latest
     if: github.event_name == 'push'
-    timeout-minutes: 60
+    timeout-minutes: 90
     strategy:
       fail-fast: false
       matrix:
@@ -90,6 +90,14 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
+      - name: Jest Cache
+        uses: actions/cache@v3
+        with:
+          path: js/.jest-cache
+          key: js-jest-cache-${{ runner.os }}-${{ matrix.node-version }}-${{ hashFiles('js/yarn.lock') }}
+          restore-keys: |
+            js-jest-cache-${{ runner.os }}-${{ matrix.node-version }}-
+            js-jest-cache-${{ runner.os }}-
       - name: Install NodeJS
         uses: actions/setup-node@v3
         with:
@@ -114,6 +122,14 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
+      - name: Jest Cache
+        uses: actions/cache@v3
+        with:
+          path: js/.jest-cache
+          key: js-jest-cache-${{ runner.os }}-${{ matrix.node-version }}-${{ hashFiles('js/yarn.lock') }}
+          restore-keys: |
+            js-jest-cache-${{ runner.os }}-${{ matrix.node-version }}-
+            js-jest-cache-${{ runner.os }}-
       - name: Install NodeJS
         uses: actions/setup-node@v3
         with:
--- a/.github/workflows/js.yml
+++ b/.github/workflows/js.yml
@@ -94,9 +94,9 @@ jobs:
         uses: actions/cache@v3
         with:
           path: js/.jest-cache
-          key: js-jest-cache-${{ runner.os }}-${{ matrix.node-version }}-${{ hashFiles('js/yarn.lock') }}
+          key: js-jest-cache-${{ runner.os }}-${{ matrix.node }}-${{ hashFiles('js/yarn.lock') }}
           restore-keys: |
-            js-jest-cache-${{ runner.os }}-${{ matrix.node-version }}-
+            js-jest-cache-${{ runner.os }}-${{ matrix.node }}-
             js-jest-cache-${{ runner.os }}-
       - name: Install NodeJS
         uses: actions/setup-node@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/js.yml
```diff
@@ -90,6 +90,14 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
+      - name: Jest Cache
+        uses: actions/cache@v3
+        with:
+          path: js/.jest-cache
+          key: js-jest-cache-${{ runner.os }}-${{ matrix.node-version }}-${{ hashFiles('js/yarn.lock') }}
```
kou, line 97:
```suggestion
          key: js-jest-cache-${{ runner.os }}-${{ matrix.node }}-${{ hashFiles('js/yarn.lock') }}
```

.github/workflows/js.yml
```diff
@@ -90,6 +90,14 @@ jobs:
...
+          restore-keys: |
+            js-jest-cache-${{ runner.os }}-${{ matrix.node-version }}-
```
kou, line 99:
```suggestion
            js-jest-cache-${{ runner.os }}-${{ matrix.node }}-
```

.github/workflows/js.yml
```diff
@@ -90,6 +90,14 @@ jobs:
...
+            js-jest-cache-${{ runner.os }}-
```
kou, line 100: Do we need this?
(Can we reuse cache for different Node.js version?)

.github/workflows/js.yml
```diff
@@ -114,6 +122,14 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
+      - name: Jest Cache
+        uses: actions/cache@v3
+        with:
+          path: js/.jest-cache
+          key: js-jest-cache-${{ runner.os }}-${{ matrix.node-version }}-${{ hashFiles('js/yarn.lock') }}
+          restore-keys: |
+            js-jest-cache-${{ runner.os }}-${{ matrix.node-version }}-
```
kou, line 131:
```suggestion
          key: js-jest-cache-${{ runner.os }}-${{ matrix.node }}-${{ hashFiles('js/yarn.lock') }}
          restore-keys: |
            js-jest-cache-${{ runner.os }}-${{ matrix.node }}-
```

domoritz:
Great. Do you have some before and after measurements?

.github/workflows/js.yml
abetomo: Oh, thanks. Fixed.

.github/workflows/js.yml
abetomo: The cache is reused even if different Node.js versions are used.
I have confirmed this by switching Node.js versions locally.

abetomo:
@domoritz Thanks for the review.
Here is the link to the measurement results on GitHub Actions.
https://github.com/apache/arrow/issues/36154#issuecomment-1611331523

.github/workflows/js.yml
kou: Then, can we remove `-${{ matrix.node }}` from `key` and `restore-keys`?

abetomo: Will you be testing on more than one `node` in the future?
(like `node: [18, 20]`.)

I just tested a few times locally and it seemed a little faster to use the cache of the same Node.js version.
It is certainly unnecessary if you don't test with multiple `node`.

kou: We'll not use multiple Node.js versions unless we reduce CI time. (40+ minutes is too long to run multiple jobs. If we want to test against multiple Node.js versions, we'll use nightly CI that are defined in `dev/tasks/tasks.yml`.)

If the difference between with/without `-${{ matrix.node }}`, we want to unify it to reduce cache size. apache/arrow has multiple language implementations. So cache limit provided by GitHub Actions is small for us. So we want to reduce cache size as much as possible.

abetomo: I got it.
I removed `matrix.node`.

domoritz:
Great. The times seem faster although not by as much as I expected. Are we sure that caching doesn’t introduce issues where tests are stale? What is in the cache?

abetomo:
https://jestjs.io/docs/code-transformation
> Jest will cache the result of a transformation and attempt to invalidate that result based on a number of factors, such as the source of the file being transformed and changing configuration.

Jest caches transformed codes.
In most cases, the cache should not cause problems.

The documentation notes the following.
https://jestjs.io/docs/cli#--cache
> The cache should only be disabled if you are experiencing caching related problems.

So, it is not always possible to say that problems will not occur.

domoritz:
So should we use the cache across lock files as well then?

abetomo:
Do you mean to cache `node_modules` as well?
That seems better.

domoritz:
No, caching node modules is a separate issue and I don't think it'll help much. What I mean is that in https://github.com/apache/arrow/pull/36373/files#diff-9992b6f0e329a6dc49913c62f6efe2102f7a64f55d4bdab3d47df86292db5e87R97 we key by the lockfile but if the jest cache is automatically only used when its applicable, we can remove the restriction to specific node and lockfile versions.

abetomo:
At first, I too thought that the key was `js-jest-cache-${{ runner.os }}` would be enough.
However, if the key is the same, the cache stored on CI is not updated and the same cache is restored all the time.
So I thought it would be better to update the jest cache on CI when the lock file is updated, so I included the lock file in the key.

domoritz:
Why would it be an issue if the cache is not updated? It feels odd to update based on the lock file rather than some other test related property.

abetomo:
I agree that using a lock file is not the best way to do this.
I would like to be able to key in changes to files in `src` and `test`, but which is best?
Is it better to use `github.sha`?

If the cache is not updated, the cache will not be usable and the speed improvement will be less effective.

kou:
What are the inputs for jest's cache? We should use a hash value for them in `key`.

domoritz:
What we really want is a way to use a cache and still update it. Anything else will not catch many cases and the CI will still time out.

abetomo:
Since the cache for jest is a transformed code, it is created from the code in `src` and `test`.
I could not find a way to use those hash values for the `key` in a good way.
I thought `github.sha` seemed to be the most like it.

kou:
Then, we can use `js-jest-cache-${{ runner.os }}-${{ hashFiles('js/**/*.ts', ''js/test/**/*.ts') }}` for `key` like other our workflow files. If we can't reuse most cache when we update dependencies, we can also add `js/yarn.lock` like `js-jest-cache-${{ runner.os }}-${{ hashFiles('js/**/*.ts', ''js/test/**/*.ts', 'js/yarn.lock') }}`.

We should not use `github.sha`. If we use it, our cache is always updated even when we don't change `js/{src,test}/**/*.ts`. It will waste our available cache limit.

abetomo:
@kou Thanks for the information.
I did not know that you could use `*` in `hashFiles` or that you could specify more than one.
`js-jest-cache-${{ runner.os }}-${{ hashFiles('js/src/**/*.ts', ''js/test/**/*.ts') }}` looks good.

@domoritz What do you think?

assignUser:
> What we really want is a way to use a cache and still update it.

This is not possible due to the way actions/cache works once a cache with a certain key is created it is immutable. But using the ts  file hashes in the key should work as you intend imo. (unless dependencies also come into play than you might want to ad the lock file as kou said)

domoritz:
Makes sense that caches are immutable. Do we have a way to check whether the cache is used? If it’s used even if dependencies change, I think we should not include the lock file in the hash.

assignUser:
I am not sure what you mean with if the cache is used? Used by jest? I don't know @abetomo? Though actions/cache sets an output value if there was a cache hit or not, if that is what you mean.

kou:
The existing cache is used even when dependencies are only changed because we have `restore-keys: js-jest-cache-${{ runner.os }}-`.
But the cache isn't updated if we don't have `'js/yarn.lock'` in `hashFiles()`. Because the cache key isn't changed.
If jest's cache depends on dependencies too, we should add `js/yarn.lock` to `hashFiles()` too.

domoritz:
What I mean is that I’d like to make sure we know when the jest cache is actually used by jest or whether jest ignored it because the cache wasn’t applicable to the code.

abetomo:
The cache is generated in `cacheDirectory`.
If there is no cache (the cache is not used), a new cache file will be generated.
If the cache is used, no new file is generated.
This will help us determine if you used the cache or not.

There does not seem to be any option in jest to more easily know if the cache is used or not.

trxcllnt (approved):

abetomo:
I tried it and when the package related to the code transform changes (when the lock file is updated), jest cache is also updated.

What about the key `js-jest-cache-${{ runner.os }}-${{ hashFiles('js/src/**/*.ts', ''js/test/**/*.ts', 'js/yarn.lock') }}`?

kou (approved):
+1

I'll merge in a few days unless nobody objects this.

abetomo:
Thanks for the merge.

PR #36387, opened 2023-06-29
GH-36045:[Python] Improve usability of pc.map_lookup / MapLookupOptions

### LINKED ISSUE #36045: [Python] Improve usability of pc.map_lookup / MapLookupOptions
Trying to use `pc.map_lookup` today, I noticed some minor usability issues that could be improved:

```
arr = pa.array([{'a': 'a1', 'b': 'b1'}, {'a': 'a2', 'b': 'b2'}, {'a': 'a1', 'b': 'b3'}], 
               type=pa.map_(pa.string(), pa.string()))
```

1\. Passing the key value as a python string instead of scalar errors:

```
# works fine
>>> pc.map_lookup(arr, pa.scalar("a"), "all")
<pyarrow.lib.ListArray object at 0x7f099d2fabc0>
[
  [
    "a1"
  ],
  [
    "a2"
  ],
  [
    "a1"
  ]
]

# gives confusing error
>>> pc.map_lookup(arr, "a", "all")
...
ArrowInvalid: map_lookup: query_key can't be empty.
```

If not passing a Scalar object, we should probably just pass the python value to `pa.scalar(..)` for convenience.

2\. Allow to specify the `occurence` keyword, given that this is not necessarily obvious, it's nice to be able to specify the named argument:

```
>>> pc.map_lookup(arr, pa.scalar("a"), occurence="first")
...
TypeError: __init__() takes exactly 3 positional arguments (2 given)
```

I suppose we need to give it a default value for this, though (to make our auto-generation work, or we can look into improving the autogeneration of the signature)

R-JunmingChen:
### Rationale for this change
This PR is for #36045, which aims to improve usability of pc.map_lookup / MapLookupOptions  

For `query_key` which is not a subclass of `pyarrow.lib.Scalar`, we convert it to `scalar`

### Are these changes tested?

Yes, add one test in tests/test_compute.py::test_map_lookup
* Closes: #36045


python/pyarrow/_compute.pyx
```diff
@@ -1565,14 +1566,17 @@ class MapLookupOptions(_MapLookupOptions):
...
+    query_key : Scalar or Object can be converted to Scalar
         The key to search for.
     occurrence : str
         The occurrence(s) to return from the Map
         Accepted values are "first", "last", or "all".
     """
 
     def __init__(self, query_key, occurrence):
+        if not issubclass(type(query_key), lib.Scalar):
+            query_key = pa.scalar(query_key)
```
jorisvandenbossche, line 1578:
```suggestion
            query_key = lib.scalar(query_key)
```

and then you don't need to add the `import pyarrow as pa` at the top

python/pyarrow/_compute.pyx
```diff
@@ -1565,14 +1566,17 @@ class MapLookupOptions(_MapLookupOptions):
...
-    query_key : Scalar
+    query_key : Scalar or Object can be converted to Scalar
         The key to search for.
     occurrence : str
         The occurrence(s) to return from the Map
         Accepted values are "first", "last", or "all".
     """
 
     def __init__(self, query_key, occurrence):
+        if not issubclass(type(query_key), lib.Scalar):
```
jorisvandenbossche, line 1577:
```suggestion
        if not isinstance(query_key, lib.Scalar):
```

python/pyarrow/tests/test_compute.py
```diff
@@ -3092,6 +3092,12 @@ def test_map_lookup():
     result_all = pa.array([[1], None, None, [5, 7], None],
                           type=pa.list_(pa.int32()))
 
+    assert pc.map_lookup(arr,
+                         'one', 'first') == result_first
+    assert pc.map_lookup(arr,
+                         'one', 'last') == result_last
+    assert pc.map_lookup(arr,
+                         'one', 'all') == result_all
```
jorisvandenbossche, line 3100: Adding just the first case should be sufficient to cover this (the value for occurrence shouldn't impact how the query_key is converted)

jorisvandenbossche (commented):
Thanks for the contribution!

Added a few small comments

R-JunmingChen:
Thanks for suggestions. Have merged.

python/pyarrow/tests/test_compute.py
```diff
@@ -3092,6 +3092,8 @@ def test_map_lookup():
     result_all = pa.array([[1], None, None, [5, 7], None],
                           type=pa.list_(pa.int32()))
 
+    assert pc.map_lookup(arr,
+                         'one', 'first') == result_first
```
jorisvandenbossche, line 3096:
```suggestion
    assert pc.map_lookup(arr, 'one', 'first') == result_first
```

jorisvandenbossche (approved):
Looking good, thanks!

PR #36418, opened 2023-06-30
GH-36417: [C++] Add Buffer::data_as, Buffer::mutable_data_as

### LINKED ISSUE #36417: [C++] Add Buffer::data_as

Buffer's data is usually cast to a non-byte type before use, which requires a lot of boilerplate with `reinterpret_cast`. If the buffer happens to live in an `ArrayData` we can use `GetValues` but bare buffers don't have access to this. Let's add an accessor which does the same thing for bare buffers: `const T* Buffer::data_as<T>() const` and `T* Buffer::mutable_data_as<T>()` will greatly simplify usage of buffers.

### Component(s)

C++

bkietz:
### Rationale for this change

There's a lot of boilerplate in casting buffer bytes. A templated accessor will decrease that.

- `{Buffer, BufferBuilder, BufferSpan}::{data_as<T>, mutable_data_as<T>}`
- rewriting a few files' worth of reinterpret_casts to use these accessors to showcase usage
- `Buffer::FromVector`, a std::vector analog to Buffer::FromString
- some tangential cleanup of unnecessary copies in the Array classes

### Are these changes tested?

Not really. The rewritten files exercise the new accessors

No, but I've noticed `Buffer::operator[]` again and that should probably be deprecated.

* Closes: #36417

### DIFF
--- a/cpp/src/arrow/array/array_base.h
+++ b/cpp/src/arrow/array/array_base.h
@@ -106,7 +106,7 @@ class ARROW_EXPORT Array {
   /// \see GetNullCount
   int64_t ComputeLogicalNullCount() const;
 
-  std::shared_ptr<DataType> type() const { return data_->type; }
+  const std::shared_ptr<DataType>& type() const { return data_->type; }
   Type::type type_id() const { return data_->type->id(); }
 
   /// Buffer for the validity (null) bitmap, if any. Note that Union types
@@ -251,7 +251,7 @@ class ARROW_EXPORT PrimitiveArray : public FlatArray {
                  int64_t null_count = kUnknownNullCount, int64_t offset = 0);
 
   /// Does not account for any slice offset
-  std::shared_ptr<Buffer> values() const { return data_->buffers[1]; }
+  const std::shared_ptr<Buffer>& values() const { return data_->buffers[1]; }
 
  protected:
   PrimitiveArray() : raw_values_(NULLPTR) {}
--- a/cpp/src/arrow/array/array_dict.cc
+++ b/cpp/src/arrow/array/array_dict.cc
@@ -108,7 +108,7 @@ DictionaryArray::DictionaryArray(const std::shared_ptr<DataType>& type,
 
 const std::shared_ptr<Array>& DictionaryArray::dictionary() const {
   if (!dictionary_) {
-    // FIXME this isn't thread safe
+    // TODO(GH-36503) this isn't thread safe
     dictionary_ = MakeArray(data_->dictionary);
   }
   return dictionary_;
--- a/cpp/src/arrow/array/array_dict.cc
+++ b/cpp/src/arrow/array/array_dict.cc
@@ -50,7 +50,7 @@ using internal::CopyBitmap;
 // ----------------------------------------------------------------------
 // DictionaryArray
 
-std::shared_ptr<Array> DictionaryArray::indices() const { return indices_; }
+const std::shared_ptr<Array>& DictionaryArray::indices() const { return indices_; }
 
 int64_t DictionaryArray::GetValueIndex(int64_t i) const {
   const uint8_t* indices_data = data_->buffers[1]->data();
@@ -106,8 +106,9 @@ DictionaryArray::DictionaryArray(const std::shared_ptr<DataType>& type,
   SetData(data);
 }
 
-std::shared_ptr<Array> DictionaryArray::dictionary() const {
+const std::shared_ptr<Array>& DictionaryArray::dictionary() const {
   if (!dictionary_) {
+    // FIXME this isn't thread safe
     dictionary_ = MakeArray(data_->dictionary);
   }
   return dictionary_;
--- a/cpp/src/arrow/array/array_nested.cc
+++ b/cpp/src/arrow/array/array_nested.cc
@@ -611,10 +611,9 @@ const std::shared_ptr<Array>& StructArray::field(int i) const {
   return boxed_fields_[i];
 }
 
-const std::shared_ptr<Array>& StructArray::GetFieldByName(const std::string& name) const {
-  static std::shared_ptr<Array> null;
+std::shared_ptr<Array> StructArray::GetFieldByName(const std::string& name) const {
   int i = struct_type()->GetFieldIndex(name);
-  return i == -1 ? null : field(i);
+  return i == -1 ? nullptr : field(i);
 }
 
 Result<ArrayVector> StructArray::Flatten(MemoryPool* pool) const {
--- a/cpp/src/arrow/array/array_nested.cc
+++ b/cpp/src/arrow/array/array_nested.cc
@@ -470,11 +470,11 @@ const FixedSizeListType* FixedSizeListArray::list_type() const {
   return checked_cast<const FixedSizeListType*>(data_->type.get());
 }
 
-std::shared_ptr<DataType> FixedSizeListArray::value_type() const {
+const std::shared_ptr<DataType>& FixedSizeListArray::value_type() const {
   return list_type()->value_type();
 }
 
-std::shared_ptr<Array> FixedSizeListArray::values() const { return values_; }
+const std::shared_ptr<Array>& FixedSizeListArray::values() const { return values_; }
 
 Result<std::shared_ptr<Array>> FixedSizeListArray::FromArrays(
     const std::shared_ptr<Array>& values, int32_t list_size) {
@@ -611,9 +611,10 @@ const std::shared_ptr<Array>& StructArray::field(int i) const {
   return boxed_fields_[i];
 }
 
-std::shared_ptr<Array> StructArray::GetFieldByName(const std::string& name) const {
+const std::shared_ptr<Array>& StructArray::GetFieldByName(const std::string& name) const {
+  static std::shared_ptr<Array> null;
   int i = struct_type()->GetFieldIndex(name);
-  return i == -1 ? nullptr : field(i);
+  return i == -1 ? null : field(i);
 }
 
 Result<ArrayVector> StructArray::Flatten(MemoryPool* pool) const {
--- a/cpp/src/arrow/buffer.h
+++ b/cpp/src/arrow/buffer.h
@@ -49,7 +49,6 @@ namespace arrow {
 /// The following invariant is always true: Size <= Capacity
 class ARROW_EXPORT Buffer {
  public:
-  Buffer() = delete;
   ARROW_DISALLOW_COPY_AND_ASSIGN(Buffer);
 
   /// \brief Construct from buffer and size without copying memory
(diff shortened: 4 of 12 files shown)


cpp/src/arrow/array/array_base.h
```diff
@@ -106,7 +106,7 @@ class ARROW_EXPORT Array {
   /// \see GetNullCount
   int64_t ComputeLogicalNullCount() const;
 
-  std::shared_ptr<DataType> type() const { return data_->type; }
+  const std::shared_ptr<DataType>& type() const { return data_->type; }
```
felipecrv, line 109: Nice. This always annoyed me.

cpp/src/arrow/array/array_dict.cc
```diff
@@ -106,8 +106,9 @@ DictionaryArray::DictionaryArray(const std::shared_ptr<DataType>& type,
   SetData(data);
 }
 
-std::shared_ptr<Array> DictionaryArray::dictionary() const {
+const std::shared_ptr<Array>& DictionaryArray::dictionary() const {
   if (!dictionary_) {
+    // FIXME this isn't thread safe
```
felipecrv, line 111: File issue and mention it here?

felipecrv: Or move `dictionary_` init to constructor?

cpp/src/arrow/array/array_nested.cc
```diff
@@ -611,9 +611,10 @@ const std::shared_ptr<Array>& StructArray::field(int i) const {
   return boxed_fields_[i];
 }
 
-std::shared_ptr<Array> StructArray::GetFieldByName(const std::string& name) const {
+const std::shared_ptr<Array>& StructArray::GetFieldByName(const std::string& name) const {
+  static std::shared_ptr<Array> null;
```
felipecrv, line 615: Having to add a static like this might be reason enough to continue returning a value instead of a `&`.

cpp/src/arrow/buffer.h
```diff
@@ -50,6 +50,8 @@ namespace arrow {
 /// The following invariant is always true: Size <= Capacity
 class ARROW_EXPORT Buffer {
  public:
+  Buffer() = delete;
```
felipecrv, line 53: move the DISALLOW_COPY_AND_ASSIGN here as well?

cpp/src/arrow/buffer.h
```diff
@@ -137,6 +139,25 @@ class ARROW_EXPORT Buffer {
...
 
+  /// \brief Construct an immutable buffer that takes ownership of the contents
+  /// of an std::vector (without copying it).
+  ///
+  /// \param[in] data a string to own
+  /// \return a new Buffer instance
+  template <typename T>
+  static std::shared_ptr<Buffer> FromVector(std::vector<T> vec) {
+    auto* data = reinterpret_cast<uint8_t*>(vec.data());
+    auto size = static_cast<int64_t>(vec.size());
```
felipecrv, line 150: `* sizeof(T>` missing?

cpp/src/arrow/buffer.h
```diff
@@ -137,6 +139,25 @@ class ARROW_EXPORT Buffer {
...
+    return std::shared_ptr<Buffer>(
+        new Buffer{data, size},
+        [vec = std::move(
+             vec)  // Keep the vector's buffer alive inside the destructor until after we
+                   // have deleted the Buffer. Note we can't use this trick in FromString
+                   // since std::string's data is inline for short strings so moving
+                   // invalidates pointers into the string's buffer.
```
felipecrv, line 157: `clang-format` will have a nicer time if you place this comment after the `,` above.

cpp/src/arrow/buffer.h
```diff
@@ -137,6 +139,25 @@ class ARROW_EXPORT Buffer {
   /// \return a new Buffer instance
   static std::shared_ptr<Buffer> FromString(std::string data);
 
+  /// \brief Construct an immutable buffer that takes ownership of the contents
+  /// of an std::vector (without copying it).
+  ///
+  /// \param[in] data a string to own
+  /// \return a new Buffer instance
+  template <typename T>
+  static std::shared_ptr<Buffer> FromVector(std::vector<T> vec) {
```
felipecrv, line 148: I would call this `FromPrimitiveVector` because people frequently mistakenly assume converting an `std::vector<std::string>` to an Arrow array is an straightforward operations.

And would it be OK to declare `vec` as `&&` so accidental copies aren't to easy to create?

cpp/src/arrow/buffer_builder.h
```diff
@@ -118,6 +118,11 @@ class ARROW_EXPORT BufferBuilder {
     return Status::OK();
   }
 
+  /// \brief Append the given data to the buffer
+  ///
+  /// The buffer is automatically expanded if necessary.
+  Status Append(std::string_view v) { return Append(v.data(), v.size()); }
```
felipecrv, line 124: Nice!

cpp/src/arrow/buffer_builder.h
```diff
@@ -196,6 +202,14 @@ class ARROW_EXPORT BufferBuilder {
   int64_t length() const { return size_; }
   const uint8_t* data() const { return data_; }
   uint8_t* mutable_data() { return data_; }
+  template <typename T>
+  const T* data_as() const {
+    return reinterpret_cast<const T*>(data());
```
felipecrv, line 207: Nit: just return `data_` so Debug builds don't have to put code to call `data()` :D

cpp/src/arrow/compute/kernels/hash_aggregate.cc
```diff
@@ -293,10 +293,10 @@ struct GroupedCountImpl : public GroupedAggregator {
...
     auto other = checked_cast<GroupedCountImpl*>(&raw_other);
 
-    auto counts = reinterpret_cast<int64_t*>(counts_.mutable_data());
-    auto other_counts = reinterpret_cast<const int64_t*>(other->counts_.mutable_data());
+    auto* counts = counts_.mutable_data_as<int64_t>();
+    const auto* other_counts = other->counts_.data_as<int64_t>();
 
-    auto g = group_id_mapping.GetValues<uint32_t>(1);
+    const auto* g = group_id_mapping.buffers[1]->data_as<uint32_t>();
     for (int64_t other_g = 0; other_g < group_id_mapping.length; ++other_g, ++g) {
```
felipecrv, line 300: This would work before with a `NULL` `buffers[1]` if `group_id_mapping.length == 0`, so I'm a bit afraid of this change specifically.

felipecrv (commented):
Looks good!

cpp/src/arrow/array/array_dict.cc
pitrou: Or use https://en.cppreference.com/w/cpp/memory/shared_ptr/atomic

cpp/src/arrow/array/array_nested.cc
pitrou: Agreed. The refcount cost is also not terribly annoying since we're also doing a hash lookup, presumably.

cpp/src/arrow/buffer.h
pitrou: Or just use `// clang-format off` / `// clang-format on` around those lines.

cpp/src/arrow/array/array_dict.cc
mapleFU: Atomic shared pointer tent to use `mutex` (folly has implement one using atomic swap). I guess it may hurts the performance

pitrou: Interesting, though an uncontended mutex should be reasonably fast.

cpp/src/arrow/buffer.h
bkietz: I'll rename and add a static assert for triviality of T.

I would prefer not to forbid copies just to prevent them from happening accidentally. This method would not be used in a performance critical context anyway. We build buffers with BufferBuilder in perf-critical contexts; I anticipate this mostly being useful in tests where requiring a call to something like `kj::cp` would get annoying.

cpp/src/arrow/array/array_nested.cc
westonpace: A static like this is also an implicit mutex so it has the same synchronization cost as a refcount

cpp/src/arrow/compute/kernels/hash_aggregate.cc
```diff
@@ -2464,9 +2464,9 @@ struct GroupedCountDistinctImpl : public GroupedAggregator {
     ARROW_ASSIGN_OR_RAISE(std::shared_ptr<Buffer> remapped_g,
                           AllocateBuffer(uniques.length * sizeof(uint32_t), pool_));
 
-    const auto* g_mapping = group_id_mapping.GetValues<uint32_t>(1);
-    const auto* other_g = uniques[1].array()->GetValues<uint32_t>(1);
-    auto* g = reinterpret_cast<uint32_t*>(remapped_g->mutable_data());
+    const auto* g_mapping = group_id_mapping.buffers[1]->data_as<uint32_t>();
+    const auto* other_g = uniques[1].array()->buffers[1]->data_as<uint32_t>();
```
westonpace, line 2468: Why is `data_as` preferred over `GetValues`?

cpp/src/arrow/buffer.h
felipecrv: Fair enough.

felipecrv: I have to accept that C++ can't prevent accidental copies. One has to remain vigilant at all times :)

cpp/src/arrow/array/array_dict.cc
bkietz: Follow up filed here https://github.com/apache/arrow/issues/36503

will update the comment to
```suggestion
    // TODO(GH-36503) this isn't thread safe
```

cpp/src/arrow/buffer.h
bkietz: Yep, thanks

cpp/src/arrow/array/array_nested.cc
bkietz: reverted

cpp/src/arrow/compute/kernels/hash_aggregate.cc
bkietz: reverted

cpp/src/arrow/compute/kernels/hash_aggregate.cc
bkietz: ... IMHO it looks more natural/prettier to specify which buffer before describing what you're going to do with the buffer.

It's not one-to-one either since GetValues accounts for the offset. I've reverted replacements of GetValues.

cpp/src/arrow/buffer.h
```diff
@@ -137,6 +140,25 @@ class ARROW_EXPORT Buffer {
...
   static std::shared_ptr<Buffer> FromString(std::string data);
 
+  /// \brief Construct an immutable buffer that takes ownership of the contents
+  /// of an std::vector (without copying it).
+  ///
+  /// \param[in] vec a vector to own
+  /// \return a new Buffer instance
+  template <typename T>
+  static std::shared_ptr<Buffer> FromVectorOfTrivial(std::vector<T> vec) {
+    static_assert(std::is_trivial_v<T>);
```
pitrou, line 150: Instead of the wacky name, why not make the compile error more explicit?
```suggestion
  static std::shared_ptr<Buffer> FromVector(std::vector<T> vec) {
    static_assert(std::is_trivial_v<T>, "Buffer::FromVector needs a trivial type");
```

westonpace (approved):

cpp/src/arrow/buffer.h
bkietz: I was following @felipecrv's comment. This seems fine, though; the compilation error should remove any doubts about what the function can convert

cpp/src/arrow/buffer.h
```diff
@@ -137,6 +140,25 @@ class ARROW_EXPORT Buffer {
...
+    auto* data = reinterpret_cast<uint8_t*>(vec.data());
+    auto size_in_bytes = static_cast<int64_t>(vec.size() * sizeof(T));
```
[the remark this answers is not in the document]
bkietz: Alright, I'll add that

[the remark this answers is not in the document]
bkietz: Should I do the same for FromString's buffer?

cpp/src/arrow/buffer.h
felipecrv: https://github.com/apache/arrow/pull/36418#discussion_r1248323736

cpp/src/arrow/buffer.h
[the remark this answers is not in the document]
pitrou: Good question. I wonder if current implementations of `std::string` might return a null pointer, given the small string optimization. But it probably doesn't hurt to special case it.

pitrou:

pitrou (approved):

PR #36419, opened 2023-06-30
GH-36363: [MATLAB] Create proxy classes for the DataType class hierarchy

sgilmore10:
### Rationale for this change

In the original pull request in which we added the MATLAB `arrow.type.<Type>Type` classes (e.g. `arrow.type.Float32Type`), we did implement these classes as proxies. At the time, we weren't sure if it would be advantageous to implement the type classes as proxies, but now realize it will be for composite data structures, i.e. `Schema`, `StructArray`, `ListArray`. 

1. All classes within the `arrow.type.Type` class hierarchy are implemented as proxies. 

### Are these changes tested?

Yes, we had existing tests for these classes. 

No.

### Future Directions

1. In a followup PR request, we plan on integrating the proxy type classes and the array classes so that they share the same underlying C++` arrow::DataType` object. We thought doing so in this change would be too much code churn.

### Notes

Thank you @kevingurney for the help!

* Closes: #36363

### DIFF
--- a/matlab/src/cpp/arrow/matlab/type/proxy/primitive_ctype.h
+++ b/matlab/src/cpp/arrow/matlab/type/proxy/primitive_ctype.h
@@ -40,14 +40,14 @@ class PrimitiveCType : public arrow::matlab::type::proxy::FixedWidthType {
     using ArrowDataType = arrow_type_t<CType>;
     
     public:
-        PrimitiveCType(std::shared_ptr<ArrowDataType> primitive_type) : arrow::matlab::type::proxy::FixedWidthType(primitive_type) {
+        PrimitiveCType(std::shared_ptr<ArrowDataType> primitive_type) : arrow::matlab::type::proxy::FixedWidthType(std::move(primitive_type)) {
         }
 
         ~PrimitiveCType() {}
 
         static libmexclass::proxy::MakeResult make(const libmexclass::proxy::FunctionArguments& constructor_arguments) {
             auto data_type = arrow::CTypeTraits<CType>::type_singleton();
-            return std::make_shared<PrimitiveCType>(std::static_pointer_cast<ArrowDataType>(data_type));
+            return std::make_shared<PrimitiveCType>(std::static_pointer_cast<ArrowDataType>(std::move(data_type)));
         }
 };
 
--- a/matlab/src/cpp/arrow/matlab/type/proxy/primitive_ctype.h
+++ b/matlab/src/cpp/arrow/matlab/type/proxy/primitive_ctype.h
@@ -40,7 +40,7 @@ class PrimitiveCType : public arrow::matlab::type::proxy::Type {
     using ArrowDataType = arrow_type_t<CType>;
     
     public:
-        PrimitiveCType(const std::shared_ptr<ArrowDataType> primitive_type) : Type(primitive_type) {
+        PrimitiveCType(std::shared_ptr<ArrowDataType> primitive_type) : Type(primitive_type) {
             REGISTER_METHOD(PrimitiveCType, bitWidth);
         }
 
--- a/matlab/src/cpp/arrow/matlab/type/proxy/primitive_ctype.h
+++ b/matlab/src/cpp/arrow/matlab/type/proxy/primitive_ctype.h
@@ -1,0 +1,55 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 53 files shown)


kevingurney:
Added as myself as reviewer per https://github.com/apache/arrow/pull/36549.

matlab/src/cpp/arrow/matlab/type/proxy/type.h
```diff
@@ -0,0 +1,43 @@
...
+
+namespace arrow::matlab::type::proxy {
+
+class Type : public libmexclass::proxy::Proxy {
+    public:
+        Type(std::shared_ptr<arrow::DataType> type);
+    
+        virtual ~Type() {}
+
+        std::shared_ptr<arrow::DataType> getType();
```
kevingurney, line 32: Since we are going to likely have similar "getter" methods for other concrete `Proxy` subclasses in the future, maybe we should pick a more generic name like `"unwrap"` here to keep things more consistent across the code base. I believe this is the [pattern followed by `pyarrow`](https://github.com/apache/arrow/issues/36429).

matlab/src/cpp/arrow/matlab/type/proxy/type.cc
```diff
@@ -0,0 +1,50 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#include "arrow/matlab/type/proxy/type.h"
+
+namespace arrow::matlab::type::proxy {
+
+    Type::Type(std::shared_ptr<arrow::DataType> type) : data_type{type} {
+        // Register Proxy methods.
```
kevingurney, line 23: We can probably remove this comment now since we have been following this pattern across the code base for a while now.

kevingurney (approved):
+1

matlab/src/cpp/arrow/matlab/type/proxy/type.cc
```diff
@@ -0,0 +1,49 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#include "arrow/matlab/type/proxy/type.h"
+
+namespace arrow::matlab::type::proxy {
+
+    Type::Type(std::shared_ptr<arrow::DataType> type) : data_type{type} {
```
kou, line 22: Do we want to use `variable{...}` style instead of `variable_name(...)` here?

BTW, we may want to use `std::move(type)` here.

matlab/src/cpp/arrow/matlab/type/proxy/primitive_ctype.h
```diff
@@ -0,0 +1,55 @@
...
+template<typename CType>
+using enable_if_primitive = std::enable_if_t<is_primitive<CType>::value, bool>; 
+
+template<typename CType, enable_if_primitive<CType> = true>
+class PrimitiveCType : public arrow::matlab::type::proxy::FixedWidthType {
+    
+    using ArrowDataType = arrow_type_t<CType>;
+    
+    public:
+        PrimitiveCType(std::shared_ptr<ArrowDataType> primitive_type) : arrow::matlab::type::proxy::FixedWidthType(primitive_type) {
```
kou, line 43: Do we need `std::move()` here?

matlab/src/cpp/arrow/matlab/type/proxy/primitive_ctype.h
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        ~PrimitiveCType() {}
+
+        static libmexclass::proxy::MakeResult make(const libmexclass::proxy::FunctionArguments& constructor_arguments) {
+            auto data_type = arrow::CTypeTraits<CType>::type_singleton();
+            return std::make_shared<PrimitiveCType>(std::static_pointer_cast<ArrowDataType>(data_type));
```
kou, line 50: Do we need `std::move()` here?

matlab/src/cpp/arrow/matlab/type/proxy/type.cc
sgilmore10: That's a good point. I'll `std::move` `type`.

matlab/src/cpp/arrow/matlab/type/proxy/primitive_ctype.h
sgilmore10: Will do.

matlab/src/cpp/arrow/matlab/type/proxy/primitive_ctype.h
sgilmore10: Will do

matlab/src/cpp/arrow/matlab/type/proxy/type.cc
kevingurney: @kou - I will go ahead and merge @sgilmore10's changes at this point based on her response.

However, let us know if you think this still needs changes. We can always follow up with another PR to address remaining feedback.

Marking this as resolved.

PR #36424, opened 2023-07-01
GH-36423: [C++][Compute] Support "or" in `Expression::IsSatisfiable`

### LINKED ISSUE #36423: [C++][Compute] Can we support `or` in `Expression::IsSatisfiable`

When I go through the code, I found `Expression::IsSatisfiable` has handled case for "and" and "and_kleene" function, but "or" and "or_kleene" is not support.
If it's possible, I'd like to submit the code and testing for it

### Component(s)

C++

mapleFU:
### Rationale for this change

This patch supports "or" in `Expression::IsSatisfiable`, if all it child is not satisfiable, the expression itself is not satisfiable.

Support "or" and "or_kleene" in `IsSatisfiable`

### Are these changes tested?

Yes

No

* Closes: #36423


mapleFU:
I've no idea why https://github.com/apache/arrow/actions/runs/5430438414/jobs/9876174167?pr=36424 failed...

kou:
It's #36424.

mapleFU:
@westonpace Hi, would you mind take a look?

pitrou:
@bkietz Do you want to take a look at this?

felipecrv (approved):

cpp/src/arrow/compute/expression.cc
```diff
@@ -362,8 +362,20 @@ bool Expression::IsSatisfiable() const {
...
+      return false;
+    }
+  }
+
+  if (call->function_name == "or_kleene" || call->function_name == "or") {
+    bool all_unsatisfiable =
+        std::all_of(call->arguments.begin(), call->arguments.end(),
+                    [](const Expression& arg) { return !arg.IsSatisfiable(); });
+    if (all_unsatisfiable) {
+      return false;
```
bkietz, line 378: IMO, using the std algorithms in this way is less readable than the for loop, please use that instead
```suggestion
    for (const Expression& arg : call->arguments) {
      if (!arg.IsSatisfiable()) return false;
    }
    return true;
  }

  if (call->function_name == "or_kleene" || call->function_name == "or") {
    for (const Expression& arg : call->arguments) {
      if (arg.IsSatisfiable()) return true;
    }
    return false;
```

Or alternatively rewrite to return the result of the std algorithms directly
```suggestion
    return std::all_of(call->arguments.begin(), call->arguments.end(),
                       [](const Expression& arg) { return arg.IsSatisfiable(); });
  }

  if (call->function_name == "or_kleene" || call->function_name == "or") {
    return std::any_of(call->arguments.begin(), call->arguments.end(),
                       [](const Expression& arg) { return arg.IsSatisfiable(); });
```

bkietz (approved):
LGTM, one minor nit:

cpp/src/arrow/compute/expression_test.cc
```diff
@@ -461,6 +461,29 @@ TEST(Expression, IsSatisfiable) {
...
+           call("or", {literal(null), field_ref("bool")}),
+       }) {
+    ARROW_SCOPED_TRACE(might_true.ToString());
+    EXPECT_TRUE(Bind(might_true).IsSatisfiable());
+  }
+
+  for (const auto& never_true : {
+           // N.B. this is "or_kleene"
+           or_(literal(false), literal(null)),
+           call("or", {literal(false), literal(null)}),
```
pitrou, line 479: Doesn't this test simply exercise constant folding? @bkietz What do you think?

cpp/src/arrow/compute/expression.cc
mapleFU: You're right. I've changed to the second one

cpp/src/arrow/compute/expression_test.cc
bkietz: IsSatisfiable does not (currently) do any canonicalization or simplification of the expression, so even things which could be trivially constant folded are not caught. We usually only call IsSatisfiable after simplifying by a guarantee or otherwise modifying an expression based on information available only to the caller, so we shouldn't try simplifying in IsSatisfiable since that's probably redundant and more expensive than we need.

@mapleFU could you add this to the doccomment for IsSatisfiable?

```diff
   /// Return true if this expression could evaluate to true. Will return true for any
-  /// unbound, non-boolean, or unsimplified Expressions
+  /// unbound or non-boolean Expressions. IsSatisfiable does not (currently) do any
+  /// canonicalization or simplification of the expression, so even Expressions
+  /// which are be unsatisfiable may spuriously return `true` here. This function is
+  /// intended for use in predicate pushdown where a filter expression is simplified
+  /// by a guarantee, so it assumes that trying to simplify again would be redundant.
```

mapleFU: done

c_glib/arrow-glib/error.hpp
```diff
@@ -46,7 +46,7 @@ namespace garrow {
     if (status.ok()) {
       return TRUE;
     } else {
-      std::string context = std::move(context_func());
+      std::string context = context_func();
```
pitrou, line 49: Why this change @mapleFU ?

cpp/src/arrow/compute/expression.h
```diff
@@ -94,7 +94,11 @@ class ARROW_EXPORT Expression {
   bool IsNullLiteral() const;
 
   /// Return true if this expression could evaluate to true. Will return true for any
-  /// unbound, non-boolean, or unsimplified Expressions
+  /// unbound or non-boolean Expressions. IsSatisfiable does not (currently) do any
+  /// canonicalization or simplification of the expression, so even Expressions
+  /// which are be unsatisfiable may spuriously return `true` here. This function is
```
pitrou, line 99: Typo
```suggestion
  /// which are unsatisfiable may spuriously return `true` here. This function is
```

c_glib/arrow-glib/error.hpp
mapleFU: done

pitrou (approved):

pitrou:

cpp/src/arrow/compute/expression.cc
felipecrv: I should have seen this. :D

c_glib/arrow-glib/error.hpp
kou: Oh, sorry. I'll fix it by #36531.

PR #36430, opened 2023-07-01
GH-36429: [C++][Python] Wrap and Unwrap Functions for Other PyArrow Objects

### LINKED ISSUE #36429: Wrap / Unwrap Functions for Other PyArrow Objects

The PyArrow C++ module provides wrapper and unwrapper functions for a variety of Arrow entities, including schemas, arrays, record batches, and tables. Examples are:
- `arrow::python::wrap_table` or `pyarrow_wrap_table`: `std::shared_ptr<arrow::Table>` to `PyObject*`
- `arrow::python::unwrap_table` or `pyarrow_unwrap_table`: `PyObject*` to `std::shared_ptr<arrow::Table>`

However, it is missing conversion functions for newer entities, like:
- Dataset
- Filesystem
- Compute Expression
- Scanner
- RecordBatchReader

It would be helpful to expose such functions for 3rd-party APIs and tools.

### Component(s)

C++, Python

srilman:
### Rationale for this change

It is helpful for 3rd-party code that uses Arrow to sometimes create or get Arrow entities in Python code and then pass it in to C++ extension code. Or on the other hand, use the C++ code for missing options and then pass to Python.

This PR includes wrapper and unwrapper functions additional PyArrow objects like datasets, compute expressions, filesystems, and so on.

### Are these changes tested?

Not right now since none of the other wrapper functions were tested.

Yes, additional user-facing functions were added, and the documentation was updated

* Closes: #36429

### DIFF
--- a/docs/source/python/integration/extending.rst
+++ b/docs/source/python/integration/extending.rst
@@ -146,6 +146,41 @@ C++ objects.
    Return whether *obj* wraps an Arrow C++ :type:`SparseCSRMatrix` pointer;
    in other words, whether *obj* is a :py:class:`pyarrow.SparseCSRMatrix` instance.
 
+.. function:: bool arrow::py::is_expression(PyObject* obj)
+
+   Return whether *obj* wraps an Arrow C++ :type:`compute::Expression` pointer;
+   in other words, whether *obj* is a :py:class:`pyarrow.compute.Expression` instance.
+
+.. function:: bool arrow::py::is_filesystem(PyObject* obj)
+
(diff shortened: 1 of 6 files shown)


pitrou:
@srilman There are CI failures, can you take a look?

Also, it looks like this PR will not work if some components such as dataset have been disabled. You need to be a bit more fine-grained. You can take a look at `arrow/util/config.h` for available preprocessor defines.

srilman:
Thanks @JerAguilon! I couldn't figure out the circular dependency issue myself. I will rebase my PR on top of yours

Issue #36360, opened 2023-06-28
[C++] Reduce number of internal APIs that define default_memory_pool() as default argument value


felipecrv:
### Describe the enhancement requested

Default argument values are convenient, but create implicit uses that are hard to see by simply scanning code at function callsites.

Function that accept a `MemoryPool *` should (ideally) only call functions that perform allocations with that memory pool. Due to many functions being declared with `MemoryPool *pool = default_memory_pool()`, it's very easy to accidentally use the default memory pool when you should be passing down the memory pool you received.

Fixing this issue has two steps:

1) Changing internal functions to not define `default_memory_pool()` as default argument and changing the callsites to pass a memory pool they receive from callers of explicitly pass the `default_memory_pool()`.

2) Change declarations of functions that call `default_memory_pool()` to take a `MemoryPool *` parameter instead. Then go to step 1 again.

Related issues:
 - #36293
 - #34025

### Component(s)

C++

felipecrv:
@wjones127 @pitrou @marin-ma

ShaiviAgarwal2:
@felipecrv 
The enhancement requested is to improve the code by making it easier to see when a function is using the default memory pool instead of the intended memory pool. This is done by changing the internal functions to not define `default_memory_pool()` as the default argument and changing the callsites to explicitly pass the intended memory pool. Additionally, functions that call `default_memory_pool()` should be changed to take a `MemoryPool * `parameter instead.

```
void myFunction(int arg1, int arg2, MemoryPool *pool = default_memory_pool()) {
    // function implementation using pool instead of default_memory_pool()
}

// Updated version:
void myFunction(int arg1, int arg2, MemoryPool *pool) {
    // function implementation using pool instead of default_memory_pool()
}

// Call site before update:
myFunction(arg1, arg2);

// Updated call site:
MemoryPool *pool = getMemoryPool(); // get the intended memory pool
myFunction(arg1, arg2, pool);

```

felipecrv:
> This is done by changing the internal functions to not define default_memory_pool() as the default argument and changing the callsites to explicitly pass the intended memory pool.

Yes, of course. The bad consequence of this is that it breaks APIs that are already public. The multi-step approach I suggested in the description (temporary removal of the default + callsite fixes) would at least ensure the Arrow codebase itself is not using `default_memory_pool()` by accident.

Eventually, we could break some of the APIs and drop the `pool = default_memory_pool()` from parameter declarations.

[...]

PR #36431, opened 2023-07-02
GH-35511: [C++] Util: add memory_pool in `SwapEndianArrayData`

### LINKED ISSUE #35511: [C++] SwapEndianArrayData should allocate from a configurable pool

Currently, `SwapEndianArrayData()` will always allocate new buffers from the default memory pool. It should instead conform to other allocating functions and accept a `MemoryPool*` argument:

```
Result<std::shared_ptr<ArrayData>> SwapEndianArrayData(
    const std::shared_ptr<ArrayData>& data, MemoryPool* = default_memory_pool());
```

### Component(s)

C++

mapleFU:
### Rationale for this change

Add memory_pool argument in `SwapEndianArrayData` to avoid allocating all from default memory pool.

Add memory_pool argument in `SwapEndianArrayData` 

### Are these changes tested?

Maybe tested by ipc

Maybe user would use the new interface

* Closes: #35511


mapleFU:
@bkietz Hi, would you mind take a look?

pitrou:
@felipecrv Would you have a bit of time to review this? Thanks!

cpp/src/arrow/array/util.cc
```diff
@@ -296,11 +301,14 @@ class ArrayDataEndianSwapper {
...
 Result<std::shared_ptr<ArrayData>> SwapEndianArrayData(
-    const std::shared_ptr<ArrayData>& data) {
+    const std::shared_ptr<ArrayData>& data, MemoryPool* pool) {
   if (data->offset != 0) {
     return Status::Invalid("Unsupported data format: data.offset != 0");
   }
-  ArrayDataEndianSwapper swapper(data);
+  if (pool == nullptr) {
+    return Status::Invalid("Unsupported pool == nullptr");
+  }
```
felipecrv, line 310: This being an internal API, should we even have this validation?

pitrou: No indeed!

pitrou: Also, in general it's assumed that people don't pass nullptr where a pointer is expected. There may be specific exceptions where the argument is optional.

mapleFU: Removed now

felipecrv: I think a `DCHECK(pool)` is enough here.

The reason we pass `*` instead of `&` in functions like this is that we follow the Google Style convention that says non-const "reference" parameters should be pointers instead of `&`. A consequence of that convention is that we expect the caller to do the right thing and pass a non-null mutable parameter unless stated in the function documentation that `nullptr` is allowed.

cpp/src/arrow/array/util.cc
```diff
@@ -77,7 +77,9 @@ class ArrayDataWrapper {
 
 class ArrayDataEndianSwapper {
  public:
-  explicit ArrayDataEndianSwapper(const std::shared_ptr<ArrayData>& data) : data_(data) {
+  explicit ArrayDataEndianSwapper(const std::shared_ptr<ArrayData>& data,
+                                  MemoryPool* pool)
+      : data_(data), pool_(pool) {
```
felipecrv, line 82: Did you grep for all calls to `ArrayDataEndianSwapper` and explicitly pass a pool or the `default_memory_pool()` explicitly?

Context: https://github.com/apache/arrow/issues/36360

felipecrv (commented):
A few comments.

cpp/src/arrow/array/util.cc
felipecrv: ( I only saw your comments after I posted mine) :)

felipecrv (approved):

cpp/src/arrow/array/util.cc
mapleFU: Quick answer: yes, `ArrayDataEndianSwapper` is merly used. I've capture all callings.

And `pool_` haven't been default initialized. So compiler will prevent from that calling

felipecrv: > So compiler will prevent from that calling

But you have `= default_memory_pool()` as a default value argument, no?

mapleFU: 1. class `ArrayDataEndianSwapper` has prevent from it
2. function `SwapEndianArrayData` doesn't. It's only used in `ipc`, and I've changed them. You can check it again

felipecrv: No need to check. I just wanted to confirm. :-) PR is approved.

mapleFU: Fall sleep now, thanks a lot!

PR #36435, opened 2023-07-02
GH-35710: [R] Followup improvements to new configure script

### LINKED ISSUE #35710: [R] Issues with configure script

I found two issues with the latest rewrite of the configure script for the R package:

1. There's a new script to check whether the library version and the R package version are compatible. This is fine, but for arrow 11, there are versions 11.0.0.2 and 11.0.0.3 on CRAN. The script does not remove the fourth component, and thus it stops when it finds that 11.0.0 != 11.0.0.3. My understanding is that versions a.b.c.d of the R package should be considered compatible to a.b.c of the library.
2. `LIB_DIR` is hardcoded [here](https://github.com/apache/arrow/blob/f3500f65c0b15e285cba2b1822385248b3424f84/r/configure#L266) to `something/lib`. In many systems, this is `lib64`, not `lib`, and `pkgconfig` provides this info.

### Component(s)

R

nealrichardson:
### Rationale for this change

See #35710

* [Get `LIB_DIR` from `pkg-config` where possible](https://github.com/apache/arrow/pull/36435/commits/c8d09eb7c1077f580100b808a307f92afd17dd38), to handle the possibility that it is `lib64` and not just `lib` on some platforms.
* [Allow x.y.z.1 to use x.y.z C++ library](https://github.com/apache/arrow/pull/36435/commits/a77f909830c89846fd5119de56c536a333e4f44b), so that apt/yum official release packages can be used with patched versions submitted to CRAN.

### Are these changes tested?

The version check change has a unit test. The LIB_DIR change hopefully is well enough covered by our existing CI.

* Closes: #35710

### DIFF
--- a/r/configure
+++ b/r/configure
@@ -263,7 +263,6 @@ do_autobrew () {
 # Once libarrow is obtained, this function sets `PKG_LIBS`, `PKG_DIRS`, and `PKG_CFLAGS`
 # either from pkg-config or by inferring things about the directory in $1
 set_pkg_vars () {
-  LIB_DIR="$1/lib"
   if [ "$PKG_CONFIG_AVAILABLE" = "true" ]; then
     set_pkg_vars_with_pc
   else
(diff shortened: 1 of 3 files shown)


kou (approved):
+1

Can we use more descriptive PR title?

How about writing commit message contents to the PR description too because our merge script only uses the PR description for merged commit. (Commit messages are discarded.)

kou (approved):
Yes! Thanks!

PR #36436, opened 2023-07-02
GH-36247: [R] Add write_csv_dataset

### LINKED ISSUE #36247: [R] Add function `write_csv_dataset()`

Add function `write_csv_dataset()` (we don't currently support the functionality needed for `write_delim_dataset()`, and `write_tsv_dataset()`), to mirror our `open_csv_dataset()` function

### Component(s)

R

dgreiss:
### Rationale for this change

Create a convenience wrapper around `write_dataset` for csv files. 

Adds a `write_csv_dataset()` 

### Are these changes tested?

Yes a few tests were added. 

Yes a new function has been added. If this looks good I can add more to the docs as well. 
* Closes: #36247


thisisnic (commented):
Thanks for making this PR!  

This is looking good so far, and have added some comments about potential changes.

One thought I had is that when I was designing `open_csv_dataset()` etc, the parameters in that function are different to those in `open_dataset()` as they are intended to reflect those in `read_csv_arrow()`.  `read_csv_arrow()` in turn mirrors the parameters in `readr::read_csv()`.  The idea behind this is that you can use the same code as you would if you were using `readr` for the workflows.  

The implementation of `write_csv_arrow()` doesn't do this (maybe it should, but that hasn't been done yet), but we could do those things here.

In the implementation of `open_csv_dataset()`, I also intentionally don't allow extra things to be passed in via the `...`, to try to keep things simpler to reason about (advanced users can use `open_dataset()` and pass in whatever they like there, if they need to).

What I'm wondering is, if we want to take the opportunity to expose the options differently here, to create an API for users that is easier to reason about.

For example, `readr::write_csv` has the following options exposed in its parameters:

* `na`
* `append`
* `col_names`
* `quote`
* `escape`
* `eol`

We could take these, work out if they map to the Arrow options nicely, and do the hard work inside of the function to convert them for the user, a bit like we do for the CSV reader.

What do you think?  I appreciate that this sounds like a whole lot of work, so I'd be happy to break it down into smaller tasks and do some bits myself if you didn't want to do the whole lot.

dgreiss:
> The implementation of `write_csv_arrow()` doesn't do this (maybe it should, but that hasn't been done yet), but we could do those things here.

Got it, I created a separate issue #36700  to add the other functions `write_delim_arrow()` and `write_tsv_arrow()`

> In the implementation of `open_csv_dataset()`, I also intentionally don't allow extra things to be passed in via the `...`, to try to keep things simpler to reason about (advanced users can use `open_dataset()` and pass in whatever they like there, if they need to).

Makes sense, I'll update it to follow that convention. 

> What I'm wondering is, if we want to take the opportunity to expose the options differently here, to create an API for users that is easier to reason about.
> 
> For example, `readr::write_csv` has the following options exposed in its parameters:
> 
> * `na`
> * `append`
> * `col_names`
> * `quote`
> * `escape`
> * `eol`
> 
> We could take these, work out if they map to the Arrow options nicely, and do the hard work inside of the function to convert them for the user, a bit like we do for the CSV reader.

Right now `na`, `eol` and `delim` map to Arrow's `null_string`, `eol` and `delimiter` and I've exposed all of them in the PR. I can make sure those options get added as well in #36700. 

The mapping of `readr` options does complicate things if a user uses both Arrow and `readr` options (eg. `write_dataset(ds, file, delim = ',', delimiter = ';')`). So we'll have to handle that possibility by either throwing an error or defaulting to the Arrow option.

r/R/dataset-write.R
```diff
@@ -178,12 +178,32 @@ write_dataset <- function(dataset,
   }
 
   path_and_fs <- get_path_and_filesystem(path)
+
+  dots <- list(...)
+  if (format %in% c("csv", "tsv") && any(c("delimiter", "delim") %in% names(dots))) {
+    stop("Do not set a delimiter for csv or tsv formats.")
+  }
```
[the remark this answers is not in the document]
dgreiss: I think you're right, let's let users specify a csv format with any delimiter.

r/R/dataset-format.R
```diff
@@ -76,9 +76,10 @@ FileFormat <- R6Class("FileFormat",
 )
 FileFormat$create <- function(format, schema = NULL, ...) {
   opt_names <- names(list(...))
-  if (format %in% c("csv", "text") || any(opt_names %in% c("delim", "delimiter"))) {
+  if (format %in% c("csv", "text", "txt") || any(opt_names %in% c("delim", "delimiter"))) {
     CsvFileFormat$create(schema = schema, ...)
   } else if (format == "tsv") {
+    # This delimiter argument is ignored.
```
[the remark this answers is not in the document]
dgreiss: I don't think that fixes it. I believe the reason the delimiter is ignored, is because it calls `dataset___CsvFileFormat__make()`, which creates a `CsvFileFormat` and `CsvFileFormat` does not have a public member to set the [write options](https://arrow.apache.org/docs/cpp/api/dataset.html#_CPPv4N5arrow7dataset13CsvFileFormatE).

The Arrow guides have an example where they use `ds::FileSystemDataset::Write()`, but the current implementation is to pass an execution plan to Acero through `ExecPlan_Write()`. 

Issue #31601 is a related to expose the `FileSystemDatasetWriteOptions` in R which may be an alternate solution.

dgreiss:
@thisisnic I think it's ready for another review.

thisisnic:
Thanks for all your work on this @dgreiss.  I think that before we push ahead much further, it'd be worth addressing the discrepancies with `write_csv_arrow()`; I've opened #36864 for discussion first.

r/R/csv.R
```diff
@@ -510,32 +510,55 @@ CsvReadOptions$create <- function(use_threads = option_use_threads(),
...
 }
 
-readr_to_csv_write_options <- function(include_header = TRUE,
+readr_to_csv_write_options <- function(col_names = TRUE,
                                        batch_size = 1024L,
-                                       na = "") {
+                                       delim = ",",
+                                       na = "",
+                                       eol = "\n",
+                                       quote = "Needed") {
```
thisisnic, line 518: The valid values for `quote` in `readr::write_delim()` are `("needed", "all", "none")`, so do we need to allow those values here and then convert them to the Arrow-valid version in the body of this function?

r/R/csv.R
```diff
@@ -510,32 +510,55 @@ CsvReadOptions$create <- function(use_threads = option_use_threads(),
...
+CsvWriteOptions$create <- function(include_header = TRUE,
+                                   batch_size = 1024L,
+                                   null_string = "",
+                                   delimiter = ",",
+                                   eol = "\n",
+                                   quoting_style = c("Needed", "AllValid", "None")) {
+  if (missing(quoting_style)) {
+    quoting_style <- "Needed"
+  }
+  quoting_style_opts <- c("Needed", "AllValid", "None")
```
thisisnic, line 541: I think here, you can use `match.arg()` instead

r/R/csv.R
```diff
@@ -510,32 +510,55 @@ CsvReadOptions$create <- function(use_threads = option_use_threads(),
...
+
   assert_that(is.logical(include_header))
+  assert_that(is_integerish(batch_size, n = 1, finite = TRUE), batch_size > 0)
+  assert_that(is.character(delimiter))
   assert_that(is.character(null_string))
   assert_that(!is.na(null_string))
   assert_that(length(null_string) == 1)
   assert_that(!grepl('"', null_string), msg = "na argument must not contain quote characters.")
+  assert_that(is.character(eol))
+  assert_that(quoting_style %in% quoting_style_opts, msg = "quoting_style must be 1 of 'Needed', 'AllValid' or 'None'")
```
thisisnic, line 551: Once you use `match.arg()` above, you won't need this check

r/R/dataset-write.R
```diff
@@ -209,6 +228,202 @@ write_dataset <- function(dataset,
...
+                                max_open_files = 900L,
+                                max_rows_per_file = 0L,
+                                min_rows_per_group = 0L,
+                                max_rows_per_group = bitwShiftL(1, 20),
+                                col_names = TRUE,
+                                batch_size = 1024L,
+                                delim = ",",
+                                na = "",
+                                eol = "\n",
+                                quote = "Needed") {
```
thisisnic, line 315: This makes sense, as we are using the existing `write_dataset()` parameters followed by the `readr::write_delim()` ones, the same as we did for `open_delim_dataset()`.  

We might need to update the value of `quote` here too, to match the `readr::write_delim()` ones. Possible the same with `na`.

I was wondering if we should be enabling `col_types` by passing in a schema, but this isn't a valid parameter in `write_dataset()`, so we shouldn't here.

r/R/dataset-write.R
```diff
@@ -209,6 +228,202 @@ write_dataset <- function(dataset,
...
+                              min_rows_per_group = 0L,
+                              max_rows_per_group = bitwShiftL(1, 20),
+                              col_names = TRUE,
+                              batch_size = 1024L,
+                              na = "",
+                              eol = "\n",
+                              quote = "Needed") {
+  if (!missing(max_rows_per_file) && missing(max_rows_per_group) && max_rows_per_group > max_rows_per_file) {
+    max_rows_per_group <- max_rows_per_file
+  }
```
thisisnic, line 405: Just to check, why do we need to replicate this logic here?

r/R/dataset-write.R
```diff
@@ -209,6 +228,202 @@ write_dataset <- function(dataset,
...
+#' The default is an empty string `""`.
+#' @param eol the end of line character to use for ending rows. The default is `"\n"`.
+#' @param quote How to handle fields which contain characters that need to be quoted.
+#' - `Needed` - Only enclose values in quotes which need them, because their CSV rendering can
+#'  contain quotes itself (e.g. strings or binary values) (the default)
+#' - `AllValid` -   Enclose all valid values in quotes. Nulls are not quoted. May cause readers to
+#' interpret all values as strings if schema is inferred.
+#' - `None` -   Do not enclose any values in quotes. Prevents values from containing quotes ("),
+#' cell delimiters (,) or line endings (\\r, \\n), (following RFC4180). If values
+#' contain these characters, an error is caused when attempting to write.
```
thisisnic, line 294: Can you use `#' @inheritParams write_dataset` here to save copying some of this?

thisisnic (commented):
Thanks - had a look at this with the conversation from the last package dev meeting in mind, and added a few comments below.  Lots of this is looking done and I think we'll just need to iron out details.

r/R/csv.R
dgreiss: I think it's cleaner to only support the Arrow version, otherwise you need to validate and support both options:

```r
write_dataset(ds, path, quote = "Needed")

write_delim_dateset(ds, path, quote = "needed")
```

Also, I'm not sure the behaviour of the `needed` options are identical. Specifically Arrow only encloses for quotes, whereas readr encloses for a delimiter, quote, or newline. The arrow options are [here](https://github.com/apache/arrow/blob/9b3bf08f78f690605394ad3815177ece931f06d3/cpp/src/arrow/csv/options.h#L173-L186): 

```c++
/// \brief Quoting style for CSV writing
enum class ARROW_EXPORT QuotingStyle {
  /// Only enclose values in quotes which need them, because their CSV rendering can
  /// contain quotes itself (e.g. strings or binary values)
  Needed,
  /// Enclose all valid values in quotes. Nulls are not quoted. May cause readers to
  /// interpret all values as strings if schema is inferred.
  AllValid,
  /// Do not enclose any values in quotes. Prevents values from containing quotes ("),
  /// cell delimiters (,) or line endings (\\r, \\n), (following RFC4180). If values
  /// contain these characters, an error is caused when attempting to write.
  None
};

```

The `readr::write_*` options for `quote` [are](https://readr.tidyverse.org/reference/write_delim.html): 

* needed - Values are only quoted if needed: if they contain a delimiter, quote, or newline.
* all - Quote all fields.
* none - Never quote fields.

r/R/dataset-write.R
dgreiss: Otherwise this throws an error: 

```
write_delim_dataset(ds, dst_dir, max_rows_per_file = 5L)

> Error: Invalid: max_rows_per_group must be less than or equal to max_rows_per_file
```

This check gets duplicated on all the `write_*_dataset()` functions, so there may be a way to refactor, but didn't think it was worth the indirection.

r/R/dataset-write.R
dgreiss: Didn't know about this tag. Thanks!

r/R/dataset-write.R
dgreiss: Agree, and if you want to change the schema you could change the schema of the `dataset` before writing.

r/R/csv.R
thisisnic: This is a tricky one; the aim of supporting the `readr` parameters is so that folks can use the `arrow` functions as drop-in replacements for their `readr` etc code, and so it follows that we'd support both the `readr` parameter names and possible values.  For example, we support things like compact column specification in `read_csv_arrow()` even though Arrow itself doesn't support this.

It also feels a bit divergent to have parameter values which are using Pascal Case here as well when everything else is in snake case.

@nealrichardson I'd be keen to get your thoughts here given you were involved in a lot of the design of the earlier functions.

paleolimbot: My take is that if Arrow can't produce identical behaviour (or if we don't have the bandwidth to implement tests for identical behaviour), the readr value shouldn't be accepted (any of us could open a follow-up ticket to request that feature from Arrow C++ developers if it turns out to be important). It sounds like @dgreiss' solution of using the UpperCamelCase version is reasonable...if a user does attempt to `write_csv_arrow(..., quote = "needed")`, they will get an error and can refer to the documentation about accepted values and how they may differ from readr behaviour.

dgreiss: I've went ahead and added the readr options and updated the docs and tests to make sure we are testing the output. Notably[ Arrow quotes all strings and binaries](https://github.com/apache/arrow/blob/a7d5217937ca8b8be77995910f672c5e61e9d699/cpp/src/arrow/csv/writer.h#L31-L39), not just in cases where there is a quote or delimiter in the string:

```c++
// Functionality for converting Arrow data to Comma separated value text.
// This library supports all primitive types that can be cast to a StringArrays.
// It applies to following formatting rules:
//  - For non-binary types no quotes surround values.  Nulls are represented as the empty
//  string.
//  - For binary types all non-null data is quoted (and quotes within data are escaped
//  with an additional quote).
//    Null values are empty and unquoted.
```

I've learned more than I expected about the Arrow CSV writer through this PR 😅. @thisisnic @paleolimbot let me know if there's anything else on this one.

r/_pkgdown.yml
```diff
@@ -138,6 +138,7 @@ articles:
       - developers/data_object_layout
 
 reference:
+
```
thisisnic, line 141:
```suggestion
```

thisisnic (commented):
This is great work, thanks @dgreiss!  I've made a few final suggestions, but keen to get this merged soon once they are resolved unless there's anything big I've missed.

dgreiss:
Thanks I've accepted the changes and it should be good to go.

r/tests/testthat/test-dataset-write.R
```diff
@@ -812,3 +812,208 @@ test_that("write_dataset() errors on data.frame with NULL names", {
...
+  expect_equal(lines[2], "\"1\";\"1\";\"true\";\"a\"")
+
+  dst_dir <- make_temp_dir()
+  write_tsv_dataset(df, dst_dir, quote = "all", eol = "\r\n")
+  ds <- open_dataset(dst_dir, format = "tsv")
+  expect_equal(df, ds |> collect())
+
+  lines <- paste(readLines(paste0(dst_dir, "/part-0.tsv")), sep = "\n")
+  expect_equal(lines[2], "\"1\"\t\"1\"\t\"true\"\t\"a\"")
+
```
thisisnic, line 1010:
```suggestion

```

thisisnic (approved):
Thanks, looks good to merge now!

PR #36463, opened 2023-07-04
GH-36446: [C++] Minor style improvements in ConcatenateImpl

### LINKED ISSUE #36446: [C++] ConcatenateImpl Constructor re-uses moved argument "const ArrayDataVector & in"

In `cpp/src/arrow/array/concatenate.cc` , the `const ArrayDataVector & in` argument is referenced after being moved into the `in_` within the member initializer list:

https://github.com/apache/arrow/blob/7ebc88c8fae62ed97bc30865c845c8061132af7e/cpp/src/arrow/array/concatenate.cc#L172-L191

where `using ArrayDataVector = std::vector<std::shared_ptr<ArrayData>>;` in `type_fwd.h`.

I was surprised the code works and that this seems to result in a copy of `in` to `in_` . But I suspect that either

1. The signature should be `ConcatenateImpl(ArrayDataVector in, MemoryPool* pool)` and `in` should not be re-used after the move, or:
2. The member initializer list should be : `in_(in), pool_(pool), out_(std::make_shared<ArrayData>())`

### Component(s)

C++

mapleFU:
### Rationale for this change

Fixing style in `ConcatenateImpl::ConcatenateImpl`

1. Remove `std::move` on a const-ref
2. Use a foreach loop

### Are these changes tested?

Yes, by existing tests.

No.

* Closes: #36446


mapleFU:
We can also uses:

```
ConcatenateImpl(ArrayDataVector in, MemoryPool* pool)
```

and call move when calling to `ConcatenateImpl::ConcatenateImpl`, but I guess copying `shared_ptr` is not bottleneck here. If reviewer want, I'll do it

kou:
> We can also uses:
> 
> ```
> ConcatenateImpl(ArrayDataVector in, MemoryPool* pool)
> ```
> 
> and call move when calling to `ConcatenateImpl::ConcatenateImpl`, but I guess copying `shared_ptr` is not bottleneck here. If reviewer want, I'll do it

Can we avoid copying `in` by this? Then could you do this?

mapleFU:
@kou sure, done it

mapleFU:
I've pushed a fixing in my branch, but it doesn't occur in this patch. Would this a feature?

pitrou:
@mapleFU There are CI failures, can you take a look?

mapleFU:
@pitrou I've no idea why the test failed. I guess here is another bugs about concat or the calling on concat, I'm a bit busy so maybe take a look at it this weekend

pitrou:
You probably moved something that shouldn't be moved?

mapleFU:
Yes you are right but I didn't found it out. I guess it's not hard to find it if I can reproduce it. Maybe I'll try on windows machine tonigh.

pitrou:
@mapleFU Can you try rebasing/merging first?

mapleFU:
@pitrou Yes, problem still exists, I think this is only on windows, I'll run and check it later.

mapleFU:
Oh I've found the problem, I made a mistake here! The member `in_` of `ConcatenateImpl` is a reference...

cpp/src/arrow/array/concatenate.cc
```diff
@@ -169,22 +169,26 @@ Status PutOffsets(const std::shared_ptr<Buffer>& src, Offset first_offset, Offse
   return Status::OK();
 }
 
+// Implement of `arrow::Concatenate`.
+// N.B. The ConcatenateImpl should accept a reference to a vector of ArrayData,
+// instead of a temporary value.
```
pitrou, line 174: What does it mean? Is it a TODO? Otherwise, I don't see a reason for this comment.

mapleFU: Well not a todo, just tell the user not to use like:

```c++
ConcatenateImpl i{/* get prvalue */GenerateArray()};
i.concate()
```

That would be unsafe, because `ConcatenateImpl` holds a const reference. If you think it's not neccessary, I'll remove it.

pitrou: This is a private API, so I don't think it's necessary.

mapleFU: Fixed now

pitrou (approved):

Issue #36412, opened 2023-06-30
[Python][CI] Fix deprecation warnings in the pandas nightly build


jorisvandenbossche:
The pandas nightly/upstream_devel builds show some warnings from pandas that we should address:

```
=============================== warnings summary ===============================
opt/conda/envs/arrow/lib/python3.10/site-packages/pyarrow/tests/test_dataset.py::test_make_fragment
  /opt/conda/envs/arrow/lib/python3.10/site-packages/numpy/core/fromnumeric.py:59: FutureWarning: 'DataFrame.swapaxes' is deprecated and will be removed in a future version. Please use 'DataFrame.transpose' instead.
    return bound(*args, **kwds)

opt/conda/envs/arrow/lib/python3.10/site-packages/pyarrow/tests/test_dataset.py::test_legacy_write_to_dataset_drops_null
  /opt/conda/envs/arrow/lib/python3.10/site-packages/pyarrow/parquet/core.py:3471: FutureWarning: The default of observed=False is deprecated and will be changed to True in a future version of pandas. Pass observed=False to retain current behavior or observed=True to adopt the future default and silence this warning.
    for keys, subgroup in data_df.groupby(partition_keys):

opt/conda/envs/arrow/lib/python3.10/site-packages/pyarrow/tests/test_feather.py::test_strings[1]
opt/conda/envs/arrow/lib/python3.10/site-packages/pyarrow/tests/test_feather.py::test_strings[1]
opt/conda/envs/arrow/lib/python3.10/site-packages/pyarrow/tests/test_feather.py::test_strings[2]
opt/conda/envs/arrow/lib/python3.10/site-packages/pyarrow/tests/test_feather.py::test_strings[2]
  /opt/conda/envs/arrow/lib/python3.10/site-packages/pyarrow/tests/test_feather.py:107: FutureWarning: Mismatched null-like values None and nan found. In a future version, pandas equality-testing functions (e.g. assert_frame_equal) will consider these not-matching and raise.
    assert_frame_equal(result, expected)

opt/conda/envs/arrow/lib/python3.10/site-packages/pyarrow/tests/test_pandas.py::TestConvertStringLikeTypes::test_pandas_unicode
opt/conda/envs/arrow/lib/python3.10/site-packages/pyarrow/tests/test_pandas.py::TestConvertStringLikeTypes::test_bytes_to_binary
[...]

PR #36465, opened 2023-07-04
GH-23870: [Python] Ensure parquet.write_to_dataset doesn't create empty files for non-observed dictionary (category) values

jorisvandenbossche:

If we partition on a categorical variable with "unobserved" categories (values present in the dictionary, but not in the actual data), the legacy path in `pq.write_to_dataset` currently creates empty files. The new dataset-based path already has the preferred behavior, and this PR fixes it for the legacy path and adds a test for both as well.

This also fixes one of the pandas deprecation warnings listed in https://github.com/apache/arrow/issues/36412

### Are these changes tested?

Yes

Yes, this no longer creates a hive-style directory with one empty file (parquet file with 0 rows) when users have unobserved categories. However, this aligns the legacy path with the new and default dataset-based path.
* Closes: #23870

### DIFF
--- a/python/pyarrow/parquet/core.py
+++ b/python/pyarrow/parquet/core.py
@@ -3468,7 +3468,7 @@ def write_to_dataset(table, root_path, partition_cols=None,
         if len(partition_keys) == 1:
             partition_keys = partition_keys[0]
 
-        for keys, subgroup in data_df.groupby(partition_keys):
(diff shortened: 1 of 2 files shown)


AlenkaF (approved):

jorisvandenbossche:

jorisvandenbossche:
@github-actions crossbow submit *pandas*

PR #36480, opened 2023-07-05
GH-36492: [CI][Python] Add Ubuntu 22.04 nightly build

### LINKED ISSUE #36492: [CI][Python] Add Ubuntu 22.04 nightly build

Ubuntu 22.04 is the most recent LTS version.
It also has recent enough gRPC packages, so Flight will be enabled and tested on this platform.

### Component(s)

Continuous Integration, Python

pitrou:
Ubuntu 22.04 is the most recent LTS version.
It also has recent enough gRPC packages, so Flight will be enabled and tested on this platform.

* Closes: #36492

### DIFF
--- a/dev/tasks/tasks.yml
+++ b/dev/tasks/tasks.yml
@@ -1229,8 +1229,8 @@ tasks:
       image: ubuntu-python
 
   test-ubuntu-22.04-python-3:
-    ci: azure
-    template: docker-tests/azure.linux.yml
+    ci: github
+    template: docker-tests/github.linux.yml
     params:
       env:
         UBUNTU: 22.04
--- a/dev/tasks/tasks.yml
+++ b/dev/tasks/tasks.yml
@@ -1228,6 +1228,14 @@ tasks:
         UBUNTU: 20.04
       image: ubuntu-python
 
+  test-ubuntu-22.04-python-3:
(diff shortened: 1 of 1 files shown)


pitrou:
@kou Should we use Github or Azure for this new job?

pitrou:
@github-actions crossbow submit test-ubuntu-22.04-python-3

assignUser:
I would say GitHub due to.mote capacity. 

Though I am pretty sure we must have a 22.04 job already but I need to check...

pitrou:
@github-actions crossbow submit test-ubuntu-22.04-python-3

pitrou:
> Though I am pretty sure we must have a 22.04 job already but I need to check...

We have some for C++, not for Python AFAICT (not using system packages anyway).

assignUser:
We have `verify-rc-source-python-linux-ubuntu-22.04-amd64` and everything  based on the conda dockerfile (e.g. `conda-python`) is also running on 22.04 though not with system packages as you say...

pitrou:
@kou What do you think?

kou (approved):
+1

I want to use GitHub Actions as much as possible because I don't like Azure Pipelines UI when I check a job result...

I'll create an issue for this and merge this.

PR #36366, opened 2023-06-28
GH-36250: [MATLAB] Add `arrow.array.StringArray` class

kevingurney:
### Rationale for this change

Thanks to @sgilmore10's [recent changes to enable UTF-8 <-> UTF-16 string conversions](#36167), we can now add support for creating Arrow `String` arrays (UTF-8 encoded) from MATLAB `string` arrays (UTF-16 encoded).

### What changes are included in this PR?

1. Added new `arrow.array.StringArray` class that can be constructed from MATLAB [`string`](https://www.mathworks.com/help/matlab/ref/string.html?s_tid=doc_ta) and [`cellstr`](https://www.mathworks.com/help/matlab/ref/cellstr.html) types. **Note**: We explicitly decided to *not* support [`char`](https://www.mathworks.com/help/matlab/ref/char.html?s_tid=doc_ta) arrays for the time being.
2. Factored out code for extracting "raw" `const uint8_t*` from a MATLAB `logical` Data Array into a new function `bit::unpacked_as_ptr` so that it can be reused across multiple Array `Proxy` classes. See https://github.com/apache/arrow/issues/36335.
3. Added new `arrow.type.StringType` type class and associated `arrow.type.ID.String` enum value.
4. Enabled support for creating `RecordBatch` objects from MATLAB `table`s containing `string` data.
5. Updated `arrow::matlab::array::proxy::Array::toString` code to convert from UTF-8 to UTF-16 for display in MATLAB.

**Examples**

*Most MATLAB `string` arrays round-trip*

```matlab
>> matlabArray = ["A"; "B"; "C"]

matlabArray = 

  3x1 string array

    "A"
    "B"
    "C"

>> arrowArray = arrow.array.StringArray(matlabArray)

arrowArray = 

[
  "A",
  "B",
  "C"
]

>> matlabArrayRoundTrip = toMATLAB(arrowArray)          

matlabArrayRoundTrip = 

  3x1 string array

    "A"
    "B"
    "C"

>> isequal(matlabArray, matlabArrayRoundTrip)

ans =

  logical

   1
```

*MATLAB `string(missing)` Values get mapped to `null` by default*

```matlab
>> matlabArray = ["A"; string(missing); "C"]

matlabArray = 

  3x1 string array

    "A"
    <missing>
    "C"

>> arrowArray = arrow.array.StringArray(matlabArray) 

arrowArray = 

[
  "A",
  null,
  "C"
]

>> matlabArrayRoundTrip = toMATLAB(arrowArray) 

[...]

PR #36485, opened 2023-07-05
GH-36483: [C++] Make `UTF8StringToUTF16` and `UTF16StringToUTF8` accept `string_views`

### LINKED ISSUE #36483: [C++] Make `UTF8StringToUTF16` and `UTF16StringToUTF8` accept `string_views`

`UTF8StringToUTF16` and `UTF16StringToUTF8` accept `const` references as inputs. As @kou pointed out, these functions should accept `string_views`. 

### Component(s)

C++

sgilmore10:
### Rationale for this change

As @kou pointed out in #36366, it's better to pass read-only string data via  as string_views than as const references. I recently submitted two new functions called `UTF8StringToUTF16` and `UTF16StringToUTF8`, both of which accept const references. This PR changes the signatures of those functions to accept string_views.

Changed the function signatures of `UTF8StringToUTF16` and `UTF16StringToUTF8`:

1. `arrow::Result<std::u16string> UTF8StringToUTF16(const std::string& source)` ->
    `arrow::Result<std::u16string> UTF8StringToUTF16(std::string_view source)` 

2. `arrow::Result<std::string> UTF16StringToUTF8(const std::u16string& source)` ->
    `arrow::Result<std::string> UTF16StringToUTF8(std::u16string_view source)` 

### Are these changes tested?

Uses existing tests in `utf8_util_test.cc`

No.

Only the MATLAB Interface uses these functions, so  I don't expect these changes to break anyone's code:

1. Search [results](https://github.com/search?q=repo%3Aapache%2Farrow%20UTF8StringToUTF16&type=code) for `UTF8StringToUTF16`
2. Search [results](https://github.com/search?q=repo%3Aapache%2Farrow+UTF16StringToUTF8&type=code) for `UTF16StringToUTF8`

* Closes: #36483

### DIFF
--- a/matlab/src/cpp/arrow/matlab/array/proxy/timestamp_array.cc
+++ b/matlab/src/cpp/arrow/matlab/array/proxy/timestamp_array.cc
@@ -49,7 +49,8 @@ namespace arrow::matlab::array::proxy {
         const mda::TypedArray<mda::MATLABString> units_mda = opts[0]["TimeUnit"];
 
         // extract the time zone string
-        MATLAB_ASSIGN_OR_ERROR(const auto timezone, arrow::util::UTF16StringToUTF8(timezone_mda[0]),
+        const std::u16string u16_timezone = timezone_mda[0];
+        MATLAB_ASSIGN_OR_ERROR(const auto timezone, arrow::util::UTF16StringToUTF8(u16_timezone),
                                error::UNICODE_CONVERSION_ERROR_ID);
 
         // extract the time unit
--- a/matlab/src/cpp/arrow/matlab/array/proxy/timestamp_array.cc
+++ b/matlab/src/cpp/arrow/matlab/array/proxy/timestamp_array.cc
@@ -49,7 +49,7 @@ namespace arrow::matlab::array::proxy {
         const mda::TypedArray<mda::MATLABString> units_mda = opts[0]["TimeUnit"];
 
         // extract the time zone string
-        const std::u16string u16_timezone = timezone_mda[0];
+        const std::u16string& u16_timezone = timezone_mda[0];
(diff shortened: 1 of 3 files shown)


sgilmore10:
I believe the failures are unrelated to the changes in this PR.

matlab/src/cpp/arrow/matlab/array/proxy/timestamp_array.cc
```diff
@@ -49,7 +49,8 @@ namespace arrow::matlab::array::proxy {
         const mda::TypedArray<mda::MATLABString> units_mda = opts[0]["TimeUnit"];
 
         // extract the time zone string
-        MATLAB_ASSIGN_OR_ERROR(const auto timezone, arrow::util::UTF16StringToUTF8(timezone_mda[0]),
+        const std::u16string u16_timezone = timezone_mda[0];
```
kou, line 52: I don't know the type of `timezone_mda[0]` but does this cause a copy?

Can we use `std::u16string_view` here?

```cpp
std::u16string_view u16_timezone(timezone_mda[0]);
```

sgilmore10: Hi @kou, this does cause a copy. Unfortunately, it's not  possible to create a `std::816string_view` from `timezone_mda` directly. However, I can extract a `const` reference from `timezone_mda:

```cpp
const std::u16string& u16_timezone = timezone_mda[0];
```

That should avoid the copy.

sgilmore10: Just pushed the change.

kou (approved):
+1

PR #36486, opened 2023-07-05
GH-34509: [C++][Parquet] Improve docstrings for ArrowReaderProperties::batch_size

### LINKED ISSUE #34509: [C++][Parquet] batch size in ArrowReaderProperties does not affect ReadTable API

set_batch_size in ArrowReaderProperties can be configured when building Parquet's FileReader. However, this option only affects RecordBatchReader. The ReadTable API will still generate a continuous table with only one batch, and the batch_size parameter does not take any effect during ReadTable. (I originally thought it was batch size during the parquet scan) Probably add more explanations to the doc to avoid confusion.

### Component(s)

Parquet

mapleFU:
Make it clear that the `batch_size` setting is ignored by some Parquet FileReader APIs.
* Closes: #34509

### DIFF
--- a/cpp/src/parquet/properties.h
+++ b/cpp/src/parquet/properties.h
@@ -820,8 +820,11 @@ class PARQUET_EXPORT ArrowReaderProperties {
   /// \brief Set the maximum number of rows to read into a record batch.
   ///
   /// Will only be fewer rows when there are no more rows in the file.
+  /// Note that some APIs such as ReadTable may ignore this setting.
   void set_batch_size(int64_t batch_size) { batch_size_ = batch_size; }
-  /// Return the batch size, and it only applies to reading into RecordBatch.
+  /// Return the batch size in rows.
+  ///
+  /// Note that some APIs such as ReadTable may ignore this setting.
   int64_t batch_size() const { return batch_size_; }
 
   /// Enable read coalescing (default false).
--- a/cpp/src/parquet/properties.h
+++ b/cpp/src/parquet/properties.h
@@ -817,11 +817,12 @@ class PARQUET_EXPORT ArrowReaderProperties {
     }
   }
 
-  /// \brief Set the maximum number of rows to read into a chunk or record batch.
+  /// \brief Set the maximum number of rows to read into a arrow chunk or record batch.
   ///
+  /// It's only for `arrow::RecordBatchReader` or RecordBatchGenerator.
   /// Will only be fewer rows when there are no more rows in the file.
   void set_batch_size(int64_t batch_size) { batch_size_ = batch_size; }
-  /// Return the batch size.
+  /// Return the batch size for `arrow::RecordBatchReader` or RecordBatchGenerator.
   int64_t batch_size() const { return batch_size_; }
 
   /// Enable read coalescing (default false).
--- a/cpp/src/parquet/properties.h
+++ b/cpp/src/parquet/properties.h
@@ -817,12 +817,11 @@ class PARQUET_EXPORT ArrowReaderProperties {
     }
   }
 
-  /// \brief Set the maximum number of rows to read into a arrow chunk or record batch.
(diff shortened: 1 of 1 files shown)


mapleFU:
@XinyuZeng @wgtmac Mind take a look?

cpp/src/parquet/properties.h
```diff
@@ -817,11 +817,12 @@ class PARQUET_EXPORT ArrowReaderProperties {
...
   }
 
-  /// \brief Set the maximum number of rows to read into a chunk or record batch.
+  /// \brief Set the maximum number of rows to read into a arrow chunk or record batch.
   ///
+  /// It's only for `arrow::RecordBatchReader` or RecordBatchGenerator.
   /// Will only be fewer rows when there are no more rows in the file.
   void set_batch_size(int64_t batch_size) { batch_size_ = batch_size; }
-  /// Return the batch size.
+  /// Return the batch size for `arrow::RecordBatchReader` or RecordBatchGenerator.
```
wgtmac, line 825: Would it be sufficient to simply say `it only applies to reading into RecordBatch`?

cpp/src/parquet/properties.h
```diff
@@ -817,11 +817,12 @@ class PARQUET_EXPORT ArrowReaderProperties {
     }
   }
 
-  /// \brief Set the maximum number of rows to read into a chunk or record batch.
+  /// \brief Set the maximum number of rows to read into a arrow chunk or record batch.
```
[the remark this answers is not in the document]
mapleFU: `ChunkedArray` I guess ╮(￣▽￣"")╭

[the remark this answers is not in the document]
wgtmac: Then probably we should remove `chunk` here to reduce confusion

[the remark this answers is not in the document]
mapleFU: done

cpp/src/parquet/properties.h
```diff
@@ -817,11 +817,12 @@ class PARQUET_EXPORT ArrowReaderProperties {
     }
   }
 
-  /// \brief Set the maximum number of rows to read into a chunk or record batch.
+  /// \brief Set the maximum number of rows to read into a record batch.
   ///
+  /// It only applies to reading into RecordBatch.
```
wgtmac, line 822: this can be removed as well since it duplicates the above comment.

wgtmac (approved):

mapleFU:
@pitrou would you mind take a look at this? Just a document update, because `ArrowReaderProperties::batch_size` just works on interfaces about `RecordBatch`, like async generate record batch, or `RecordBatchReader`

cpp/src/parquet/properties.h
```diff
@@ -817,11 +817,11 @@ class PARQUET_EXPORT ArrowReaderProperties {
     }
   }
 
-  /// \brief Set the maximum number of rows to read into a chunk or record batch.
+  /// \brief Set the maximum number of rows to read into a record batch.
   ///
   /// Will only be fewer rows when there are no more rows in the file.
```
pitrou, line 822:
```suggestion
  /// Will only be fewer rows when there are no more rows in the file.
  /// Note that some APIs such as ReadTable may ignore this setting.
```

cpp/src/parquet/properties.h
```diff
@@ -817,11 +817,11 @@ class PARQUET_EXPORT ArrowReaderProperties {
...
   void set_batch_size(int64_t batch_size) { batch_size_ = batch_size; }
-  /// Return the batch size.
+  /// Return the batch size, and it only applies to reading into RecordBatch.
```
pitrou, line 824:
```suggestion
  /// Return the batch size in rows.
  ///
  /// Note that some APIs such as ReadTable may ignore this setting.
```

pitrou (approved):
Thanks for noticing this @mapleFU . This limitation is probably unfortunate...

PR #36489, opened 2023-07-05
GH-36488: [C++] Import/Export ArrowDeviceArray

### LINKED ISSUE #36488: [C++] Import/Export ArrowDeviceArray

With the addition of the `ArrowDeviceArray` struct, we should add to the `bridge.h`/`.cc` files methods to allow importing and exporting `ArrowDeviceArray` objects properly leveraging the memory manager and buffers.

### Component(s)

C++

zeroshade:
### Rationale for this change
With the addition of the `ArrowDeviceArray` we should provide import/export functions just like we have for `ArrowArray`.

Adding Import/Export functions to the `bridge.h/.cc` files for C Data Non-CPU data.

This requires adding a device type to buffers and memory managers to propagate through. 

### Are these changes tested?
Yes.

(Still need to add tests to `bridge_test.cc` but I wanted to get eyes on this first)

No.

* Closes: #36488

### DIFF
--- a/cpp/src/arrow/buffer.h
+++ b/cpp/src/arrow/buffer.h
@@ -63,13 +63,13 @@ class ARROW_EXPORT Buffer {
         data_(data),
         size_(size),
         capacity_(size),
-        device_type_(DeviceType::kCPU) {
+        device_type_(DeviceAllocationType::kCPU) {
     SetMemoryManager(default_cpu_memory_manager());
   }
 
   Buffer(const uint8_t* data, int64_t size, std::shared_ptr<MemoryManager> mm,
          std::shared_ptr<Buffer> parent = NULLPTR,
-         std::optional<DeviceType> device_type = std::nullopt)
+         std::optional<DeviceAllocationType> device_type = std::nullopt)
       : is_mutable_(false),
         data_(data),
         size_(size),
@@ -254,7 +254,7 @@ class ARROW_EXPORT Buffer {
 
   const std::shared_ptr<MemoryManager>& memory_manager() const { return memory_manager_; }
 
-  std::optional<DeviceType> device_type() const { return device_type_; }
+  std::optional<DeviceAllocationType> device_type() const { return device_type_; }
 
   std::shared_ptr<Buffer> parent() const { return parent_; }
 
@@ -310,7 +310,7 @@ class ARROW_EXPORT Buffer {
   const uint8_t* data_;
   int64_t size_;
   int64_t capacity_;
-  std::optional<DeviceType> device_type_;
+  std::optional<DeviceAllocationType> device_type_;
 
   // null by default, but may be set
   std::shared_ptr<Buffer> parent_;
--- a/cpp/src/arrow/buffer.h
+++ b/cpp/src/arrow/buffer.h
@@ -75,7 +75,7 @@ class ARROW_EXPORT Buffer {
         size_(size),
         capacity_(size),
         parent_(std::move(parent)) {
-    // will set device_type from the memory manager
+    // SetMemoryManager will also set device_type_
     SetMemoryManager(std::move(mm));
     // if a device type is specified, use that instead. for example:
     // CUDA_HOST. The CudaMemoryManager will set device_type_ to CUDA,
--- a/cpp/src/arrow/buffer.h
+++ b/cpp/src/arrow/buffer.h
@@ -57,12 +57,18 @@ class ARROW_EXPORT Buffer {
   ///
   /// \note The passed memory must be kept alive through some other means
   Buffer(const uint8_t* data, int64_t size)
-      : is_mutable_(false), is_cpu_(true), data_(data), size_(size), capacity_(size), device_type_(DeviceType::CPU) {
+      : is_mutable_(false),
+        is_cpu_(true),
+        data_(data),
+        size_(size),
+        capacity_(size),
+        device_type_(DeviceType::CPU) {
     SetMemoryManager(default_cpu_memory_manager());
   }
 
   Buffer(const uint8_t* data, int64_t size, std::shared_ptr<MemoryManager> mm,
-         std::shared_ptr<Buffer> parent = NULLPTR, DeviceType device_type = DeviceType::UNKNOWN)
+         std::shared_ptr<Buffer> parent = NULLPTR,
+         DeviceType device_type = DeviceType::UNKNOWN)
       : is_mutable_(false),
         data_(data),
         size_(size),
--- a/cpp/src/arrow/buffer.h
+++ b/cpp/src/arrow/buffer.h
@@ -20,6 +20,7 @@
 #include <cstdint>
 #include <cstring>
 #include <memory>
+#include <optional>
 #include <string>
 #include <string_view>
 #include <utility>
@@ -62,13 +63,13 @@ class ARROW_EXPORT Buffer {
         data_(data),
         size_(size),
         capacity_(size),
-        device_type_(DeviceType::CPU) {
+        device_type_(DeviceType::kCPU) {
     SetMemoryManager(default_cpu_memory_manager());
   }
 
   Buffer(const uint8_t* data, int64_t size, std::shared_ptr<MemoryManager> mm,
          std::shared_ptr<Buffer> parent = NULLPTR,
-         DeviceType device_type = DeviceType::UNKNOWN)
+         std::optional<DeviceType> device_type = std::nullopt)
       : is_mutable_(false),
         data_(data),
         size_(size),
@@ -79,7 +80,7 @@ class ARROW_EXPORT Buffer {
     // if a device type is specified, use that instead. for example:
     // CUDA_HOST. The CudaMemoryManager will set device_type_ to CUDA,
     // but you can specify CUDA_HOST as the device type to override it.
-    if (device_type != DeviceType::UNKNOWN) {
+    if (device_type != std::nullopt) {
       device_type_ = device_type;
     }
   }
@@ -253,7 +254,7 @@ class ARROW_EXPORT Buffer {
 
   const std::shared_ptr<MemoryManager>& memory_manager() const { return memory_manager_; }
 
-  DeviceType device_type() const { return device_type_; }
+  std::optional<DeviceType> device_type() const { return device_type_; }
 
   std::shared_ptr<Buffer> parent() const { return parent_; }
 
@@ -309,7 +310,7 @@ class ARROW_EXPORT Buffer {
   const uint8_t* data_;
   int64_t size_;
   int64_t capacity_;
-  DeviceType device_type_;
+  std::optional<DeviceType> device_type_;
 
   // null by default, but may be set
   std::shared_ptr<Buffer> parent_;
--- a/cpp/src/arrow/buffer.h
+++ b/cpp/src/arrow/buffer.h
@@ -57,18 +57,25 @@ class ARROW_EXPORT Buffer {
   ///
   /// \note The passed memory must be kept alive through some other means
   Buffer(const uint8_t* data, int64_t size)
-      : is_mutable_(false), is_cpu_(true), data_(data), size_(size), capacity_(size) {
+      : is_mutable_(false), is_cpu_(true), data_(data), size_(size), capacity_(size), device_type_(DeviceType::CPU) {
     SetMemoryManager(default_cpu_memory_manager());
   }
 
   Buffer(const uint8_t* data, int64_t size, std::shared_ptr<MemoryManager> mm,
-         std::shared_ptr<Buffer> parent = NULLPTR)
+         std::shared_ptr<Buffer> parent = NULLPTR, DeviceType device_type = DeviceType::UNKNOWN)
       : is_mutable_(false),
         data_(data),
         size_(size),
         capacity_(size),
         parent_(std::move(parent)) {
+    // will set device_type from the memory manager
     SetMemoryManager(std::move(mm));
+    // if a device type is specified, use that instead. for example:
+    // CUDA_HOST. The CudaMemoryManager will set device_type_ to CUDA,
+    // but you can specify CUDA_HOST as the device type to override it.
+    if (device_type != DeviceType::UNKNOWN) {
+      device_type_ = device_type;
+    }
   }
 
   Buffer(uintptr_t address, int64_t size, std::shared_ptr<MemoryManager> mm,
@@ -240,6 +247,8 @@ class ARROW_EXPORT Buffer {
 
   const std::shared_ptr<MemoryManager>& memory_manager() const { return memory_manager_; }
 
+  DeviceType device_type() const { return device_type_; }
+
   std::shared_ptr<Buffer> parent() const { return parent_; }
 
   /// \brief Get a RandomAccessFile for reading a buffer
@@ -294,6 +303,7 @@ class ARROW_EXPORT Buffer {
   const uint8_t* data_;
   int64_t size_;
   int64_t capacity_;
+  DeviceType device_type_;
 
   // null by default, but may be set
   std::shared_ptr<Buffer> parent_;
@@ -309,6 +319,7 @@ class ARROW_EXPORT Buffer {
   void SetMemoryManager(std::shared_ptr<MemoryManager> mm) {
     memory_manager_ = std::move(mm);
     is_cpu_ = memory_manager_->is_cpu();
+    device_type_ = memory_manager_->device()->device_type();
   }
 
  private:
--- a/cpp/src/arrow/c/bridge.cc
+++ b/cpp/src/arrow/c/bridge.cc
@@ -674,7 +674,8 @@ Status ExportRecordBatch(const RecordBatch& batch, struct ArrowArray* out,
 //////////////////////////////////////////////////////////////////////////
 // C device arrays
 
-Status ValidateDeviceInfo(const ArrayData& data, std::optional<DeviceAllocationType>* device_type,
+Status ValidateDeviceInfo(const ArrayData& data,
+                          std::optional<DeviceAllocationType>* device_type,
                           int64_t* device_id) {
   for (const auto& buf : data.buffers) {
     if (!buf) {
@@ -1388,7 +1389,8 @@ class ImportedBuffer : public Buffer {
       : Buffer(data, size), import_(std::move(import)) {}
 
   ImportedBuffer(const uint8_t* data, int64_t size, std::shared_ptr<MemoryManager> mm,
-                 DeviceAllocationType device_type, std::shared_ptr<ImportedArrayData> import)
+                 DeviceAllocationType device_type,
+                 std::shared_ptr<ImportedArrayData> import)
       : Buffer(data, size, mm, nullptr, device_type), import_(std::move(import)) {}
 
   ~ImportedBuffer() override {}
--- a/cpp/src/arrow/c/bridge.cc
+++ b/cpp/src/arrow/c/bridge.cc
@@ -547,7 +547,8 @@ void ReleaseExportedArray(struct ArrowArray* array) {
   }
   DCHECK_NE(array->private_data, nullptr);
   auto* pdata = reinterpret_cast<ExportedArrayPrivateData*>(array->private_data);
-  if (pdata->sync_event_.sync_event != nullptr && pdata->sync_event_.release_func != nullptr) {  
+  if (pdata->sync_event_.sync_event != nullptr &&
+      pdata->sync_event_.release_func != nullptr) {
     pdata->sync_event_.release_func(pdata->sync_event_.sync_event);
   }
   delete pdata;
@@ -673,7 +674,7 @@ Status ExportRecordBatch(const RecordBatch& batch, struct ArrowArray* out,
 //////////////////////////////////////////////////////////////////////////
 // C device arrays
 
-Status ValidateDeviceInfo(const ArrayData& data, std::optional<DeviceType>* device_type,
+Status ValidateDeviceInfo(const ArrayData& data, std::optional<DeviceAllocationType>* device_type,
                           int64_t* device_id) {
   for (const auto& buf : data.buffers) {
     if (!buf) {
@@ -704,16 +705,16 @@ Status ValidateDeviceInfo(const ArrayData& data, std::optional<DeviceType>* devi
   return Status::OK();
 }
 
-Result<std::pair<std::optional<DeviceType>, int64_t>> ValidateDeviceInfo(const ArrayData& data) {
-  std::optional<DeviceType> device_type;
+Result<std::pair<std::optional<DeviceAllocationType>, int64_t>> ValidateDeviceInfo(
+    const ArrayData& data) {
+  std::optional<DeviceAllocationType> device_type;
   int64_t device_id = -1;
   RETURN_NOT_OK(ValidateDeviceInfo(data, &device_type, &device_id));
   return std::make_pair(device_type, device_id);
 }
 
-Status ExportDeviceArray(const Array& array, RawSyncEvent sync_event, 
-                         struct ArrowDeviceArray* out,
-                         struct ArrowSchema* out_schema) {
+Status ExportDeviceArray(const Array& array, RawSyncEvent sync_event,
+                         struct ArrowDeviceArray* out, struct ArrowSchema* out_schema) {
   if (sync_event.sync_event != nullptr && sync_event.release_func) {
     return Status::Invalid(
         "Must provide a release event function if providing a non-null event");
@@ -744,8 +745,7 @@ Status ExportDeviceArray(const Array& array, RawSyncEvent sync_event,
   return Status::OK();
 }
 
-Status ExportDeviceRecordBatch(const RecordBatch& batch, 
-                               RawSyncEvent sync_event,
+Status ExportDeviceRecordBatch(const RecordBatch& batch, RawSyncEvent sync_event,
                                struct ArrowDeviceArray* out,
                                struct ArrowSchema* out_schema) {
   if (sync_event.sync_event != nullptr && sync_event.release_func == nullptr) {
@@ -1388,7 +1388,7 @@ class ImportedBuffer : public Buffer {
       : Buffer(data, size), import_(std::move(import)) {}
 
   ImportedBuffer(const uint8_t* data, int64_t size, std::shared_ptr<MemoryManager> mm,
-                 DeviceType device_type, std::shared_ptr<ImportedArrayData> import)
+                 DeviceAllocationType device_type, std::shared_ptr<ImportedArrayData> import)
       : Buffer(data, size, mm, nullptr, device_type), import_(std::move(import)) {}
 
   ~ImportedBuffer() override {}
@@ -1401,17 +1401,16 @@ struct ArrayImporter {
   explicit ArrayImporter(const std::shared_ptr<DataType>& type)
       : type_(type),
         zero_size_buffer_(std::make_shared<Buffer>(kZeroSizeArea, 0)),
-        device_type_(DeviceType::kCPU) {}
+        device_type_(DeviceAllocationType::kCPU) {}
 
   Status Import(struct ArrowDeviceArray* src, const DeviceMemoryMapper& mapper) {
-    ARROW_ASSIGN_OR_RAISE(memory_mgr_,
-                          mapper(src->device_type, src->device_id));
-    device_type_ = static_cast<DeviceType>(src->device_type);
+    ARROW_ASSIGN_OR_RAISE(memory_mgr_, mapper(src->device_type, src->device_id));
+    device_type_ = static_cast<DeviceAllocationType>(src->device_type);
     RETURN_NOT_OK(Import(&src->array));
     import_->sync_event_ = src->sync_event;
     // reset internal state before next import
     memory_mgr_.reset();
-    device_type_ = DeviceType::kCPU;
+    device_type_ = DeviceAllocationType::kCPU;
     return Status::OK();
   }
 
@@ -1758,7 +1757,7 @@ struct ArrayImporter {
   std::shared_ptr<Buffer> zero_size_buffer_;
 
   std::shared_ptr<MemoryManager> memory_mgr_;
-  DeviceType device_type_;
+  DeviceAllocationType device_type_;
 };
 
 }  // namespace
--- a/cpp/src/arrow/c/bridge.cc
+++ b/cpp/src/arrow/c/bridge.cc
@@ -522,6 +522,9 @@ struct ExportedArrayPrivateData : PoolAllocationMixin<ExportedArrayPrivateData> 
 
   std::shared_ptr<ArrayData> data_;
 
+  ReleaseEventFunc sync_release_;
+  void* sync_event_;
+
   ExportedArrayPrivateData() = default;
   ARROW_DEFAULT_MOVE_AND_ASSIGN(ExportedArrayPrivateData);
   ARROW_DISALLOW_COPY_AND_ASSIGN(ExportedArrayPrivateData);
@@ -544,7 +547,11 @@ void ReleaseExportedArray(struct ArrowArray* array) {
         << "Dictionary release callback should have marked it released";
   }
   DCHECK_NE(array->private_data, nullptr);
-  delete reinterpret_cast<ExportedArrayPrivateData*>(array->private_data);
+  auto* pdata = reinterpret_cast<ExportedArrayPrivateData*>(array->private_data);
+  if (pdata->sync_event_ != nullptr && pdata->sync_release_ != nullptr) {
+    pdata->sync_release_(pdata->sync_event_);
+  }
+  delete pdata;
 
   ArrowArrayMarkReleased(array);
 }
@@ -584,6 +591,8 @@ struct ArrayExporter {
     // Store owning pointer to ArrayData
     export_.data_ = data;
 
+    export_.sync_event_ = nullptr;
+    export_.sync_release_ = nullptr;
     return Status::OK();
   }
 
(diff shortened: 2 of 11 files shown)


zeroshade:
CC @kkraus14

cpp/src/arrow/c/bridge.cc
```diff
@@ -1275,7 +1398,19 @@ class ImportedBuffer : public Buffer {
...
+        zero_size_buffer_(std::make_shared<Buffer>(kZeroSizeArea, 0)),
+        device_type_(DeviceType::CPU) {}
+
+  Status Import(struct ArrowDeviceArray* src, const DeviceMemoryMgr& mapper) {
+    ARROW_ASSIGN_OR_RAISE(memory_mgr_,
+                          mapper.get_manager(src->device_type, src->device_id));
+    device_type_ = static_cast<DeviceType>(src->device_type);
+    RETURN_NOT_OK(Import(&src->array));
+    import_->sync_event_ = src->sync_event;
+    memory_mgr_.reset();
```
felipecrv, line 1411: Should `device_type_` be restored as well?

felipecrv: If not, I would initialize it to `UNKNOWN` in the ctor and set it to `UNKNOWN` here as `device_type_` doesn't seem to be used when importing from CPU arrays.

cpp/src/arrow/c/bridge.h
```diff
@@ -166,6 +166,139 @@ Result<std::shared_ptr<RecordBatch>> ImportRecordBatch(struct ArrowArray* array,
...
+
+/// \brief EXPERIMENTAL: Export C++ Array as an ArrowDeviceArray.
+///
+/// The resulting ArrowDeviceArray struct keeps the array data and buffers alive
+/// until its release callback is called by the consumer. All buffers in
+/// the provided array MUST have the same device_type, otherwise an error
+/// will be returned.
+///
+/// If a non-null sync_event is provided, then the sync_release func must also be
+/// non-null. If the sync_event is null, then the sync_release parameter is ignored.
```
felipecrv, line 190: can this be changed from "is ignored" to "is not called"?

cpp/src/arrow/c/bridge.h
```diff
@@ -166,6 +166,139 @@ Result<std::shared_ptr<RecordBatch>> ImportRecordBatch(struct ArrowArray* array,
...
+///
+/// \param[in] array Array object to export
+/// \param[in] sync_event A pointer to an event-like object if necessary for
+/// synchronization, otherwise null. \param[in] sync_release Function pointer to release
+/// the sync event \param[out] out C struct to export the array to \param[out] out_schema
+/// optional C struct to export the array type to
```
felipecrv, line 196: line break needed before \param

felipecrv (changes requested):

cpp/src/arrow/buffer.h
```diff
@@ -57,18 +57,25 @@ class ARROW_EXPORT Buffer {
...
         capacity_(size),
         parent_(std::move(parent)) {
+    // will set device_type from the memory manager
     SetMemoryManager(std::move(mm));
+    // if a device type is specified, use that instead. for example:
+    // CUDA_HOST. The CudaMemoryManager will set device_type_ to CUDA,
+    // but you can specify CUDA_HOST as the device type to override it.
+    if (device_type != DeviceType::UNKNOWN) {
+      device_type_ = device_type;
+    }
```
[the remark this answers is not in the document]
zeroshade: Possibly, we'd have to add a `Validate` or `Valid` method to the memory manager to do so, to allow the memory manager to specify if the device_type is valid for it. Not sure how worthwhile that is though.

cpp/src/arrow/buffer.h
```diff
@@ -57,18 +57,25 @@ class ARROW_EXPORT Buffer {
   ///
   /// \note The passed memory must be kept alive through some other means
   Buffer(const uint8_t* data, int64_t size)
-      : is_mutable_(false), is_cpu_(true), data_(data), size_(size), capacity_(size) {
+      : is_mutable_(false), is_cpu_(true), data_(data), size_(size), capacity_(size), device_type_(DeviceType::CPU) {
```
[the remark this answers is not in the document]
zeroshade: The `is_cpu_` boolean is still useful because of pinned host memory. Both `CUDA_HOST` and `ROCM_HOST` are pinned CPU memory which would have the corresponding device type (since the memory is managed by the device drivers) but is accessible to the CPU.

cpp/src/arrow/c/bridge.cc
```diff
@@ -522,6 +522,9 @@ struct ExportedArrayPrivateData : PoolAllocationMixin<ExportedArrayPrivateData>
 
   std::shared_ptr<ArrayData> data_;
 
+  ReleaseEventFunc sync_release_;
+  void* sync_event_;
```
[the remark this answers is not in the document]
zeroshade: good point, I'll add the initialization.

cpp/src/arrow/c/bridge.cc
```diff
@@ -663,6 +672,115 @@ Status ExportRecordBatch(const RecordBatch& batch, struct ArrowArray* out,
   return Status::OK();
 }
 
+//////////////////////////////////////////////////////////////////////////
+// C device arrays
+
+Result<std::pair<DeviceType, int64_t>> validate_device_info(const ArrayData& data) {
```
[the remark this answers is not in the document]
zeroshade: hmm, that's a good point. It seems to sometimes get used for private methods but rarely, I'll switch this to PascalCase like the rest of the file

cpp/src/arrow/c/bridge.cc
```diff
@@ -663,6 +672,115 @@ Status ExportRecordBatch(const RecordBatch& batch, struct ArrowArray* out,
...
+
+    if (device_type == DeviceType::UNKNOWN) {
+      device_type = buf->device_type();
+      device_id = buf->device()->device_id();
+      continue;
+    }
+
+    if (buf->device_type() != device_type) {
+      return Status::Invalid(
+          "exporting device array with buffers on more than one device.");
```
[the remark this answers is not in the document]
zeroshade: gah, too much of my Go conditioning :stuck_out_tongue: lol Since you don't typically capitalize error messages for Go. I'll fix it.

cpp/src/arrow/c/bridge.cc
```diff
@@ -663,6 +672,115 @@ Status ExportRecordBatch(const RecordBatch& batch, struct ArrowArray* out,
...
+    }
+
+    if (buf->device()->device_id() != device_id) {
+      return Status::Invalid(
+          "exporting device array with buffers on multiple device ids.");
+    }
+  }
+
+  // recursively check the children
+  auto info = std::make_pair(device_type, device_id);
```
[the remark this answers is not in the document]
zeroshade: hmm, I was trying to simplify the logic by not passing the device type and id through the recursion like that. Technically though, we could probably avoid the scenario you're suggesting by simply doing the recursive children check *first* before I check `data.buffers`. 

The only cases where `data.buffers` could be empty always have children or the entire array has a length of 0. In both cases checking the children first to either get `Unknown` (no children or length is 0) solves the problem without having to pass the device type and id through the recursion, allowing me to keep the logic simple. Thoughts?

cpp/src/arrow/c/bridge.cc
zeroshade: That's fair, prevents potential issues. I'll set it to `UNKNOWN`

zeroshade: Or rather, I'll set it to CPU here afterwards just for consistency in case it ends up eventually getting used for CPU arrays.

cpp/src/arrow/c/bridge.h
zeroshade: done

cpp/src/arrow/c/bridge.h
zeroshade: I think it got screwed up by clang-format or something, weird.

zeroshade: I think it got screwed up by clang-format or something, weird.

cpp/src/arrow/c/bridge.cc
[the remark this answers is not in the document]
felipecrv: I think the pattern I suggest means you don't have to rely on any assumption.

You can wrap the auxiliary recursive function for ease of use. This pattern is common in functional programming when turning a recursive function into a tail-call optimized one: create aux function that takes accumulator, return accumulator when done, wrap aux in a simple interface.

```cpp
Status validate_device_info(const ArrayData& data, DeviceType *device_type, int64_t *device_id);

Result<pair> validade_device_info(concst ArrayData& data) {
  DeviceType device_type;
  int64_t device_id;
  RETURN_NOT_OK(validate_device_info(data, &device_type, &device_id));
  return {device_type, device_id};
}
```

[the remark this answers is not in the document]
felipecrv: You won't have to do any comparisons in the second for loop. Just recursively call the auxiliary function passing the accumulators.

cpp/src/arrow/buffer.h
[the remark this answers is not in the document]
kkraus14: While the pointers are accessible to the CPU, there's no guarantee that there isn't stream ordered work happening on them where it could cause a race condition to CPU code accessing the buffer. Given `is_cpu_` is existing ABI, I'm wondering if it wouldn't be a safer option to have it return `False` in these cases where if someone wants to do CPU things against `CUDA_HOST` or `ROCM_HOST` buffers they should explicitly check the device type instead?

cpp/src/arrow/gpu/cuda_context.cc
```diff
@@ -384,7 +384,7 @@ Result<std::shared_ptr<Buffer>> CudaMemoryManager::ViewBufferTo(
   if (to->is_cpu()) {
     // Device-on-CPU view
     ARROW_ASSIGN_OR_RAISE(auto address, GetHostAddress(buf->address()));
-    return std::make_shared<Buffer>(address, buf->size(), to, buf);
+    return std::make_shared<Buffer>(address, buf->size(), to, buf, DeviceType::CUDA_HOST);
   }
```
kkraus14, line 388: I know this is existing code, but I don't think this is safe? It looks like we're taking the address to device memory and then just creating a new buffer wrapping that same address with a `CUDA_HOST` device?

This would only work if the device of `buf` was `CUDA_HOST` or `CUDA_MANAGED` which doesn't look to be the case here.

cpp/src/arrow/c/bridge.cc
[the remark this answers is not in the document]
zeroshade: @felipecrv have a look at the new version when you get a chance?

cpp/src/arrow/gpu/cuda_context.cc
kkraus14: Missed the `GetHostAddress` function which calls into `cuPointerGetAttribute` under the hood with `CU_POINTER_ATTRIBUTE_HOST_POINTER` to get a host accessible pointer. This will normally error, but that's fine since it works as intended. Resolving this.

felipecrv (commented):
LGTM, but I will let @kkraus14 do the final review.

cpp/src/arrow/c/bridge.cc
```diff
@@ -663,6 +672,112 @@ Status ExportRecordBatch(const RecordBatch& batch, struct ArrowArray* out,
   return Status::OK();
 }
 
+//////////////////////////////////////////////////////////////////////////
+// C device arrays
+
+Status ValidateDeviceInfo(const ArrayData& data, DeviceType* device_type,
+                          int64_t* device_id) {
```
benibus, line 679: Not a big deal ATM, but a `DeviceInfo` struct might be good have (not just here, but in general), e.g:
```c++
struct DeviceInfo {
  explicit DeviceInfo(DeviceType type = DeviceType::UNKNOWN, int64_t id = -1)
      : type(type), id(id) {}
  DeviceType type;
  int64_t id;
};
```

cpp/src/arrow/c/bridge.cc
```diff
@@ -663,6 +672,112 @@ Status ExportRecordBatch(const RecordBatch& batch, struct ArrowArray* out,
...
+  for (const auto& buf : data.buffers) {
+    if (!buf) {
+      continue;
+    }
+
+    if (*device_type == DeviceType::UNKNOWN) {
```
benibus, line 685: Pretty sure this loop would fail to catch a discrepancy if all buffers had `DeviceType::UNKNOWN` except for the last one.

felipecrv: But that is fine I think. The constraint imposed here is that non-UNKNOWN buffers have to be distinct. So only one different from UNKNOWN is trivially distinct.

benibus: If that's the case, then the function would return `Invalid` if any non-UNKNOWN buffers were "seen" before an UNKNOWN buffer (unless I'm totally missing something).

kkraus14 (approved):

cpp/src/arrow/device.h
```diff
@@ -29,6 +29,25 @@
 
 namespace arrow {
 
+/// \brief EXPERIMENTAL: Device type enum which matches up with C Data Device types
+enum class DeviceType : char {
+  UNKNOWN = 0,
```
pitrou, line 34: Hmm... we might want to use the `kSomething` convention for these enum values, under the assumption that some software might use those uppercase names for C macros?

cpp/src/arrow/c/bridge.h
```diff
@@ -166,6 +166,140 @@ Result<std::shared_ptr<RecordBatch>> ImportRecordBatch(struct ArrowArray* array,
...
+///
+/// @{
+
+/// \brief EXPERIMENTAL: Type for freeing a sync event
+///
+/// If synchronization is necessary for accessing the data on a device,
+/// a pointer to an event needs to be passed when exporting the device
+/// array. It's the responsibility of the release function for the array
+/// to release the event.
+using ReleaseEventFunc = void (*)(void*);
```
pitrou, line 180: Erm, can this be `std::function<void(void*)>` instead? We are not limited to C types here.

cpp/src/arrow/c/bridge.h
```diff
@@ -166,6 +166,140 @@ Result<std::shared_ptr<RecordBatch>> ImportRecordBatch(struct ArrowArray* array,
...
+///
+/// \param[in] array Array object to export
+/// \param[in] sync_event A pointer to an event-like object if necessary for
+/// synchronization, otherwise null.
+/// \param[in] sync_release Function pointer to release the sync event
+/// \param[out] out C struct to export the array to
+/// \param[out] out_schema optional C struct to export the array type to
+ARROW_EXPORT
+Status ExportDeviceArray(const Array& array, void* sync_event,
+                         ReleaseEventFunc sync_release, struct ArrowDeviceArray* out,
```
pitrou, line 200: I'm curious: does the user have to find out how to produce the right `sync_event` and `sync_release` for the device type? It's not obvious to me who would be able to call this function with the right parameters.

This API should ideally hide the details of the C Data Interface, which it does not here.

cpp/src/arrow/c/bridge.h
```diff
@@ -166,6 +166,140 @@ Result<std::shared_ptr<RecordBatch>> ImportRecordBatch(struct ArrowArray* array,
...
+                               struct ArrowDeviceArray* out,
+                               struct ArrowSchema* out_schema = NULLPTR);
+
+class ARROW_EXPORT DeviceMemoryMgr {
+ public:
+  virtual ~DeviceMemoryMgr() = default;
+
+  virtual Result<std::shared_ptr<MemoryManager>> get_manager(ArrowDeviceType device_type,
+                                                             int64_t device_id) const = 0;
+};
```
pitrou, line 233: If there's only a single virtual method, just use a `std::function`? Though, again, I'm not sure having the user implement this makes a very usable API.

cpp/src/arrow/c/bridge.cc
pitrou: Or at least return `Result<std::pair<DeviceType, int64_t>>` here...

cpp/src/arrow/device.h
pitrou: Also, what does `UNKNOWN` mean here? It doesn't appear in `abi.h`.

cpp/src/arrow/c/bridge.cc
zeroshade: That brings me back to my original implementation vs this tail recursion version that @felipecrv suggested haha. 

This is the internal recursive call, below it is the intended function to call which does return `Result<std::pair<DeviceType, int64_t>>`

cpp/src/arrow/c/bridge.cc
zeroshade: The `UNKNOWN` is coming from line 710, it's what we initialize the value with before we make the recursive call. @pitrou is your suggestion that I should remove the `UNKNOWN` value from the enum and just use `std::optional` everywhere?

cpp/src/arrow/device.h
zeroshade: sure, I'll add the `k` prefix. And `UNKNOWN` was my way of handling when we don't have a device type set, i think i can update this to use your suggestion of `std::optional` to handle this instead.

cpp/src/arrow/c/bridge.h
zeroshade: So the event is related to the stream and whatever would be pushing the data through the device, rather than being related to the buffers objects themselves. So the user would need to know how to produce the right sync event (and therefore the correct release) for whatever the underlying device they are using is. In theory we could instead just template this on the event type rather than accepting the void* directly (but then we would just be `reinterpret_cast`ing it to a `void*` anyway. 

The alternative (or potential evolution of this) would be to encapsulate this like your suggestion https://github.com/apache/arrow/issues/36103 but it would still be up to the user to have created the event and pass it through since the user would need to properly call the method to add triggering the event as a task on the stream for after the data is populated. But we'd still need the user to pass it in. 

I was putting off the creation of that event/stream encapsulation as an evolution of this in a subsequent PR but I can do that with this PR if you think that's better.

cpp/src/arrow/c/bridge.h
zeroshade: So the alternative option would be that the only way for users to get other device handling would be for it to get contributed and merged into the arrow lib directly. By allowing the user to implement this it means that as long as they can implement the `MemoryManager` API they can use whatever devices they want without having to wait until that support is added to the Arrow library itself. 

I could add a default implementation of this which handles CPU / CUDA since those already exist in the arrow lib and then a user can just rely on that default implementation or supply their own?

cpp/src/arrow/c/bridge.cc
pitrou: If you don't like passing a `Result` recursively, you can always build a struct with a method in it...

cpp/src/arrow/c/bridge.cc
pitrou: Yes, please remove `UNKNOWN` from the enum as it shouldn't be a legal value for it.

cpp/src/arrow/c/bridge.h
pitrou: I think having the user devise the `void*` and release func themselves is just making it easier for them to shoot themselves in the foot. That said, you're right that we can _start_ with that in this PR.

In any case, for the sake of type-proofing, perhaps we want to introduce an intermediate struct? (there are too many things that may cast silently to `void*` :-)).
```c++
struct RawSyncEvent {
  void* sync_event = nullptr;
  std::function<void(void*)> release_func;
};
```

cpp/src/arrow/c/bridge.h
pitrou: > I could add a default implementation of this which handles CPU / CUDA since those already exist in the arrow lib and then a user can just rely on that default implementation or supply their own?

Hmm, that sounds fair indeed.

cpp/src/arrow/device.h
```diff
@@ -29,6 +29,25 @@
...
+  CPU = 1,
+  CUDA = 2,
```
felipeblazing, line 36: Should we consider adding a seperate allocation type for allocations that are made using Stream Ordered Allocators [More Details Here](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY__POOLS.html) These allocators are asynchronous with respect to the host. When a user allocates data using a stream ordered allocator the allocator returns immediately and the allocation is guaranteed to be valid before the next operation is added to that particular stream. Because arrow is not handling cuda streams if a user uses the asynchronous allocator then there is no guarantee this allocation will be performed before we attempt to write or read to that memory. 

This means that if we are using async allocators the user should be informed this is the case so that they can do things like capture the stream from whatever library is making the allocation to handle the stream or at least call cudaDeviceSynchronize() to ensure that this allocation is indeed already available.

cpp/src/arrow/c/bridge.cc
zeroshade: the `Result` isn't recursive, the recursion just passes the pointers to the device type and id

cpp/src/arrow/device.h
```diff
@@ -29,6 +29,25 @@
 
 namespace arrow {
 
+/// \brief EXPERIMENTAL: Device type enum which matches up with C Data Device types
+enum class DeviceType : char {
```
felipeblazing, line 33: Is DeviceType the appropriate label for this? It seems to be more a combination of DeviceType with different allocation flavors that the drivers implement for the various device types.

pitrou: Good point, perhaps call it `DeviceAllocationType`?

cpp/src/arrow/c/bridge.h
zeroshade: done

cpp/src/arrow/device.h
zeroshade: It's intended that knowing which devices are stream ordered is up to the users of the library. This is why the DeviceArray includes an sync event that can be populated and if it isn't null it needs to be synchronized on.

Additionally after this PR, I'll be working on an interface to abstract more of the event and synchronization for buffers. I don't think we need a separate allocation type for async allocators, we just need to ensure that we're consistent with the device type enums.

cpp/src/arrow/device.h
zeroshade: I was keeping it named "DeviceType" so that it aligns with the C ABI `ArrowDeviceType` typedef. Do you think it matters for this to be named differently?

felipeblazing: If you think it will cause confusion then that is fine. I found the name DeviceType misleading. I take it updating the C ABI as well is a much more involved task?

cpp/src/arrow/device.h
felipeblazing: Its not a device which is stream ordered but rather an allocation. A user could have allocations that are both made via async allocators and via synchronous ones. If the memory layout doesn't have a way of specifying this then users will have to basically have an extra piece of information which is carried around with every allocation they make in order to ensure  that asynchronously allocated memory  calls synchronize before being accessed. 

The more I think about it though there is a further complication which is knowing which stream it was allocated on and the fact that if you enable per thread default stream this would further complicate situations in which every thread would have its own stream of value 0.  I can understand why it might be necessary to hold off on this for a while but I do think this is something that should be kept in  mind.

cpp/src/arrow/device.h
pitrou: Since we have a class `arrow::Device` which doesn't map 1-to-1 to these enum members, then I think it would be good to use another name indeed.

As for changing the C ABI: well, it wouldn't change the _ABI_, precisely :-) Also the C Device Data Interface is very new and still experimental... However, the names would probably get quite long as well.

pitrou:
@zeroshade It's weird that there are "2 workflows awaiting approval" here. @assignUser do you know what's going on?

assignUser:
This usually only happens for first time contributors but there was a Periode a while back where GitHub seemed to forget this status even for org members... I guess this is happening again :shrug:

cpp/src/arrow/device.h
Rhonda85: I agree

Rhonda85: I agree

zeroshade: @felipeblazing That makes sense, I'll add that to my list of things for this project. For now I'd prefer to hold off on adding those complexities to a future PR rather than this one. Unless you think that this PR in and of itself is going to be problematic without it?

cpp/src/arrow/device.h
zeroshade: In that case i'll update this name to `DeviceAllocationType`

cpp/src/arrow/buffer.h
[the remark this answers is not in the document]
zeroshade: Hmm, I think the intent for the `is_cpu_` flag was solely to know whether or not the data is accessible via CPU rather than caring about the stream ordered work synchronizations. I intend on expanding the existing classes/objects to make it easier to synchronize for stream ordered work as a separate abstraction, so I lean towards leaving it marked as `true` in this case, but I'm open to changing it.

@pitrou @felipeblazing @Rhonda85 what do you all think?

cpp/src/arrow/gpu/cuda_memory.cc
```diff
@@ -480,5 +485,21 @@ Result<uint8_t*> GetHostAddress(uintptr_t device_ptr) {
...
 
+Result<std::shared_ptr<MemoryManager>> DefaultMemoryMapper(ArrowDeviceType device_type,
+                                                           int64_t device_id) {
+  switch (device_type) {
+    case ARROW_DEVICE_CPU:
+      return default_cpu_memory_manager();
+    case ARROW_DEVICE_CUDA:
+    case ARROW_DEVICE_CUDA_HOST:
+    case ARROW_DEVICE_CUDA_MANAGED: {
+      ARROW_ASSIGN_OR_RAISE(auto device, arrow::cuda::CudaDevice::Make(device_id));
```
westonpace, line 496: This is a little weird but caused by this PR.  My first thought, at looking at this, is that we are returning a dangling object since `device` will be destroyed almost immediately.

I think `arrow::cuda::CudaDevice::Make` should probably be `arrow::cuda::CudaDevice::Get` and return `const CudaDevice&` or `const CudaDevice*`.

westonpace (changes requested):
I think I follow this pretty well.  Thanks!  Can you add some tests to `bridge_test.cc` testing out these new methods?

cpp/src/arrow/buffer.h
```diff
@@ -294,6 +310,7 @@ class ARROW_EXPORT Buffer {
   const uint8_t* data_;
   int64_t size_;
   int64_t capacity_;
+  std::optional<DeviceAllocationType> device_type_;
```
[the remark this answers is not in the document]
westonpace: Ah, I see this discussion was already had.

cpp/src/arrow/device.h
felipeblazing: This is a bit of an edge case. If only one process was performing allocations and it was consistent with its allocation strategy, which most are, then you would not really have this issue. So yes its fine to hold off!

cpp/src/arrow/device.h
```diff
@@ -29,6 +29,24 @@
...
+  kVULKAN = 7,
+  kMETAL = 8,
+  kVPI = 9,
+  kROCM = 10,
+  kROCM_HOST = 11,
+  kEXT_DEV = 12,
+  kCUDA_MANAGED = 13,
+  kONEAPI = 14,
+  kWEBGPU = 15,
+  kHEXAGON = 16,
```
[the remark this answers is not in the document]
zeroshade: It's mirroring the list of macros from `abi.h` for the device types

cpp/src/arrow/device.h
```diff
@@ -29,6 +29,24 @@
 
 namespace arrow {
 
+/// \brief EXPERIMENTAL: Device type enum which matches up with C Data Device types
+enum class DeviceAllocationType : char {
```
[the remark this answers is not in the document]
zeroshade: see https://github.com/apache/arrow/pull/36489#discussion_r1261448552

cpp/src/arrow/gpu/cuda_memory.cc
zeroshade: I'll look at this in a subsequent PR rather than make that change here since it'll probably have other effects elsewhere

cpp/src/arrow/buffer.h
```diff
@@ -57,18 +58,31 @@ class ARROW_EXPORT Buffer {
...
       : is_mutable_(false),
         data_(data),
         size_(size),
         capacity_(size),
         parent_(std::move(parent)) {
+    // will set device_type from the memory manager
     SetMemoryManager(std::move(mm));
+    // if a device type is specified, use that instead. for example:
+    // CUDA_HOST. The CudaMemoryManager will set device_type_ to CUDA,
+    // but you can specify CUDA_HOST as the device type to override it.
```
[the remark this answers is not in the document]
zeroshade: They are different, `CUDA` is memory allocated on the device itself while `CUDA_HOST` is memory allocated on the CPU and the pages are pinned so that they are still accessible by the devices. For example: https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1gab84100ae1fa1b12eaca660207ef585b vs https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1g37d37965bfb4803b6d4e59ff26856356

cpp/src/arrow/c/bridge.h
```diff
@@ -166,6 +167,139 @@ Result<std::shared_ptr<RecordBatch>> ImportRecordBatch(struct ArrowArray* array,
...
+///
+/// @{
+
+/// \brief EXPERIMENTAL: Type for freeing a sync event
+///
+/// If synchronization is necessary for accessing the data on a device,
+/// a pointer to an event needs to be passed when exporting the device
+/// array. It's the responsibility of the release function for the array
+/// to release the event. Both can be null if no sync'ing is necessary.
+struct RawSyncEvent {
```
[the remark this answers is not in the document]
zeroshade: I plan on changing this in the future to have a more abstracted class structure for handling events similar to the discussion [here](https://github.com/apache/arrow/issues/36103). So i want to draw a distinction between this and the eventual interface (ie. the next thing i'll be working on)

cpp/src/arrow/c/bridge.h
```diff
@@ -166,6 +167,139 @@ Result<std::shared_ptr<RecordBatch>> ImportRecordBatch(struct ArrowArray* array,
...
+Status ExportDeviceRecordBatch(const RecordBatch& batch, RawSyncEvent sync_event,
+                               struct ArrowDeviceArray* out,
+                               struct ArrowSchema* out_schema = NULLPTR);
+
+using DeviceMemoryMapper =
+    std::function<Result<std::shared_ptr<MemoryManager>>(ArrowDeviceType, int64_t)>;
+
+// ARROW_EXPORT
+// Result<std::shared_ptr<MemoryManager>> DefaultMemoryMapper(ArrowDeviceType device_type,
+//                                                            int64_t device_id);
```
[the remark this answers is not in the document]
zeroshade: forgot to delete this... i'll fix that haha.

cpp/src/arrow/c/bridge_test.cc
```diff
@@ -1112,6 +1130,391 @@ TEST_F(TestArrayExport, ExportRecordBatch) {
...
+  }
+};
+
+class MyDevice : public Device {
+ public:
+  explicit MyDevice(int value) : Device(true), value_(value) {}
+  const char* type_name() const override { return kMyDeviceTypeName; }
+  std::string ToString() const override { return kMyDeviceTypeName; }
+  bool Equals(const Device& other) const override {
+    if (other.type_name() != kMyDeviceTypeName) {
```
westonpace, line 1183: Minor nit / question.  Why compare `type_name` and not `device_type`?

westonpace (approved):
That's quite a comprehensive set of tests.  I think it's probably more exhaustive than strictly needed but that's probably ok.  I'll let others weigh in too but this looks good as far as I can tell.

It might also be good to exercise the sync event but I think, if you have plans to exercise it in practice in a follow-up PR, that is probably ok too.

cpp/src/arrow/c/bridge_test.cc
zeroshade: Mostly because I was just copying the way that `CudaDevice` implemented the function. But in theory you're right, it should compare both. I'll update that.

zeroshade:
Does anyone have any further comments here? If not I'll potentially merge this at the end of the day

zeroshade:
Looks like there's a memory leak identified by ASAN, will merge after I fix that problem.

cpp/src/arrow/gpu/cuda_memory.h
```diff
@@ -258,5 +260,9 @@ Result<uintptr_t> GetDeviceAddress(const uint8_t* cpu_data,
 ARROW_EXPORT
 Result<uint8_t*> GetHostAddress(uintptr_t device_ptr);
 
+ARROW_EXPORT
+Result<std::shared_ptr<MemoryManager>> DefaultMemoryMapper(ArrowDeviceType device_type,
```
pitrou, line 264: It seems weird to have this in `arrow/gpu`, why not under `arrow/c` instead?

cpp/src/arrow/c/bridge_test.cc
```diff
@@ -1112,6 +1130,392 @@ TEST_F(TestArrayExport, ExportRecordBatch) {
...
+    if (buf.size() > 0) {
+      memcpy(dest->mutable_data(), buf.data(), static_cast<size_t>(buf.size()));
+    }
+    return std::move(dest);
+  }
+};
+
+class MyDevice : public Device {
+ public:
+  explicit MyDevice(int value) : Device(true), value_(value) {}
```
pitrou, line 1179: Would be nice to be more explicit: `Device(/*xxx=*/ true)`

cpp/src/arrow/c/bridge_test.cc
```diff
@@ -1112,6 +1130,392 @@ TEST_F(TestArrayExport, ExportRecordBatch) {
...
+      } else {
+        buffers.push_back(nullptr);
+      }
+    }
+
+    arrow::ArrayDataVector children;
+    for (const auto& child : data.child_data) {
+      ARROW_ASSIGN_OR_RAISE(auto dest, ToDeviceData(mm, *child));
+      children.push_back(dest);
+    }
```
pitrou, line 1220: Should take care of the dictionary array as well.

cpp/src/arrow/c/bridge_test.cc
```diff
@@ -1112,6 +1130,392 @@ TEST_F(TestArrayExport, ExportRecordBatch) {
...
+              std::string(c_schema.metadata, kEncodedMetadata2.size()));
+    RecordBatchExportChecker checker{};
+    checker(&c_array, *batch, kMyDeviceType, 1, nullptr);
+
+    // Create batch anew, with the same buffer pointers
+    batch = batch_factory();
+    checker(&c_array, *batch, kMyDeviceType, 1, nullptr);
+  }
+}
+
```
pitrou, line 1518: It would be nice to have tests for importing device arrays.
Also tests that the sync event and its release func are taken into account.

cpp/src/arrow/c/bridge_test.cc
```diff
@@ -1112,6 +1130,392 @@ TEST_F(TestArrayExport, ExportRecordBatch) {
...
+  TestPrimitive(mm, fixed_size_binary(3), R"(["foo", "bar", null])");
+  TestPrimitive(mm, binary(), R"(["foo", "bar", null])");
+  TestPrimitive(mm, large_binary(), R"(["foo", "bar", null])");
+  TestPrimitive(mm, utf8(), R"(["foo", "bar", null])");
+  TestPrimitive(mm, large_utf8(), R"(["foo", "bar", null])");
+
+  TestPrimitive(mm, decimal(16, 4), R"(["1234.5670", null])");
+  TestPrimitive(mm, decimal256(16, 4), R"(["1234.5670", null])");
+
+  TestPrimitive(mm, month_day_nano_interval(), R"([[-1, 5, 20], null])");
```
pitrou, line 1322: TBH, I'm not sure it's useful to copy/paste all those tests, besides the readability burden of duplicate code.

We just want to check that primitive, nested, extension and dictionary are properly handled in general (detailed export behaviour is already tested previously).

PR #35753, opened 2023-05-25
GH-35752: [CI][GLib][Ruby] Pass GITHUB_ACTIONS environment variable to Docker containers

kou:
### Rationale for this change

It's for choosing GitHub Actions suitable output automatically.

test-unit gem checks `GITHUB_ACTIONS=true` and the environment variable is set by GitHub Actions automatically.

### What changes are included in this PR?

Add `GITHUB_ACTIONS:` to `environment:`.

`c_glib/README.md` is updated to run CI jobs.

### Are these changes tested?

Yes.

### Are there any user-facing changes?

No.
* Closes: #35752

PR #36495, opened 2023-07-06
MINOR: [CI] Use GITHUB_PATH instead of GITHUB_ACTIONS to detect environment

assignUser:
### Rationale for this change

In #35753 we started passing  GITHUB_ACTIONS into the docker containers for better log printing. The sccache script used this var to detect if it was run in gha to add sccache to the path. This now causes false positives.

Use GITHUB_PATH directly instead.

### Are these changes tested?
Will trigger crossbow jobs

### DIFF
--- a/ci/scripts/install_sccache.sh
+++ b/ci/scripts/install_sccache.sh
@@ -61,7 +61,7 @@ fi
 tar -xzvf $SCCACHE_ARCHIVE --strip-component=1 --directory $PREFIX --exclude="sccache*/*E*E*"
 chmod u+x $PREFIX/sccache
 
-if [ -z "${GITHUB_PATH}" ]; then
+if [ ! -z "${GITHUB_PATH}" ]; then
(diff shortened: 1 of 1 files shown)


assignUser:
@github-actions crossbow submit conan*

ci/scripts/install_sccache.sh
```diff
@@ -61,7 +61,7 @@ fi
 tar -xzvf $SCCACHE_ARCHIVE --strip-component=1 --directory $PREFIX --exclude="sccache*/*E*E*"
 chmod u+x $PREFIX/sccache
 
-if [ "${GITHUB_ACTIONS}" = "true" ]; then
+if [ -z "${GITHUB_PATH}" ]; then
```
kou, line 64:
```suggestion
if [ -n "${GITHUB_PATH}" ]; then
```

assignUser:
@github-actions crossbow submit conan*

kou (approved):
+1

Thanks!

PR #36499, opened 2023-07-06
GH-36482: [C++][CI] Fix sporadic test failures in AsofJoinBasicTest

### LINKED ISSUE #36482: [C++][CI] Sporadic test failures in AsofJoinBasicTest

We routinely see this kind of failures on CI:
https://github.com/apache/arrow/actions/runs/5464016379/jobs/9945469106#step:7:3365

```

[ RUN      ] AsofJoinNodeTest/AsofJoinBasicTest.TestBasic7Forward/2
/arrow/cpp/src/arrow/testing/gtest_util.cc:505: Failure
Failed
Unequal at absolute position 3

@@ -1, +1 @@
-11
+null
Expected:
  [
    10,
    null,
    12
  ]
Actual:
  [
    10,
    11,
    12
  ]
Google Test trace:
/arrow/cpp/src/arrow/acero/asof_join_node_test.cc:630: Right-1 type: timestamp[ms, tz=UTC]
/arrow/cpp/src/arrow/acero/asof_join_node_test.cc:628: Right-0 type: double
/arrow/cpp/src/arrow/acero/asof_join_node_test.cc:626: Left type: uint16
/arrow/cpp/src/arrow/acero/asof_join_node_test.cc:624: Key type: date32[day]
/arrow/cpp/src/arrow/acero/asof_join_node_test.cc:622: Time type: timestamp[us, tz=UTC]
/arrow/cpp/src/arrow/acero/asof_join_node_test.cc:620: Iteration: 3
/arrow/cpp/src/arrow/acero/asof_join_node_test.cc:611: Types seed: 1688560093425811692
/arrow/cpp/src/arrow/acero/asof_join_node_test.cc:998: AsofJoinBasicTest_TestBasic7_MutateByKey
/arrow/cpp/src/arrow/acero/asof_join_node_test.cc:997: AsofJoinBasicTest_TestBasic7Forward_MutateByKey
AsofJoinTest debug:
AsofjoinNode(0xffffc2670910): received batch from input 2:
time:   [
    1970-01-01 00:00:00.000000,
    1970-01-01 00:00:00.001500,
    1970-01-01 00:00:00.002500
  ]
key2:   [
    1970-01-01,
    1970-01-01,
    1970-01-01
  ]
key:   [
    1970-01-02,
    1970-01-02,
    1970-01-02
  ]
r1_v0:   [
    1970-01-01 00:00:00.100,
    1970-01-01 00:00:00.101,
    1970-01-01 00:00:00.102
  ]

AsofjoinNode(0xffffc2670910): received batch from input 1:
time:   [
    1970-01-01 00:00:00.000000,
    1970-01-01 00:00:00.001500,
    1970-01-01 00:00:00.002500
  ]
key2:   [
    1970-01-01,
    1970-01-01,
    1970-01-01
  ]
key:   [
    1970-01-02,
    1970-01-02,
    1970-01-02
  ]
r0_v0:   [
    10,
    11,
    12
  ]

AsofjoinNode(0xffffc2670910): received batch from input 0:
time:   [
    1970-01-01 00:00:00.000000
  ]
key2:   [
    1970-01-01
  ]
key:   [
    1970-01-02
  ]
l_v0:   [
    1
  ]

AsofjoinNode(0xffffb33fb440): Advancing input 1
AsofjoinNode(0xffffb33fb240): key hasher 1 got hashes [10621589963589458110, 10621589963589458110, 10621589963589458110]
AsofjoinNode(0xffffb33fb340): memo 1 store: for_time=0 row=0 time=0 key=10621589963589458110
AsofjoinNode(0xffffb33fb440): Advancing input 1 hit distant time=1500 at=0
AsofjoinNode(0xffffb33fb360): memo 1 remove: ts=0
AsofjoinNode(0xffffb33fb440): Advancing input 1 updated=0
AsofjoinNode(0xffffb33fb440): Advancing input 2
AsofjoinNode(0xffffb33fb240): key hasher 2 got hashes [10621589963589458110, 10621589963589458110, 10621589963589458110]
AsofjoinNode(0xffffb33fb340): memo 2 store: for_time=0 row=0 time=0 key=10621589963589458110
AsofjoinNode(0xffffb33fb440): Advancing input 2 hit distant time=1500 at=0
AsofjoinNode(0xffffb33fb360): memo 2 remove: ts=0
AsofjoinNode(0xffffb33fb440): Advancing input 2 updated=0
AsofjoinNode(0xffffb33fb240): key hasher 0 got hashes [10621589963589458110]
AsofjoinNode(0xffffb33fb440): Emplace: key=10621589963589458110 lhs_latest_row=0 lhs_latest_time=0
AsofjoinNode(0xffffb33fb440):   i=1 has_entry=1 time=0 row=0 accepted=1
AsofjoinNode(0xffffb33fb440):   i=2 has_entry=1 time=0 row=0 accepted=1
AsofjoinNode(0xffffb33fb6a0): produce batch 0:
time:   [
    1970-01-01 00:00:00.000000
  ]
key2:   [
    1970-01-01
  ]
key:   [
    1970-01-02
  ]
l_v0:   [
    1
  ]
r0_v0:   [
    10
  ]
r1_v0:   [
    1970-01-01 00:00:00.100
  ]

AsofjoinNode(0xffffc2670910): received batch from input 0:
time:   [
    1970-01-01 00:00:00.001000
  ]
key2:   [
    1970-01-01
  ]
key:   [
    1970-01-03
  ]
l_v0:   [
    2
  ]

AsofjoinNode(0xffffb33fb440): Advancing input 1
AsofjoinNode(0xffffb33fb340): memo 1 store: for_time=1000 row=1 time=1500 key=10621589963589458110
AsofjoinNode(0xffffb33fb440): Advancing input 1 hit distant time=2500 at=1000
AsofjoinNode(0xffffb33fb340): memo 1 store: for_time=1000 row=2 time=2500 key=10621589963589458110
AsofjoinNode(0xffffb33fb360): memo 1 remove: ts=1000
AsofjoinNode(0xffffb33fb440): Advancing input 1 updated=1
AsofjoinNode(0xffffb33fb440): Advancing input 2
AsofjoinNode(0xffffb33fb340): memo 2 store: for_time=1000 row=1 time=1500 key=10621589963589458110
AsofjoinNode(0xffffb33fb440): Advancing input 2 hit distant time=2500 at=1000
AsofjoinNode(0xffffb33fb340): memo 2 store: for_time=1000 row=2 time=2500 key=10621589963589458110
AsofjoinNode(0xffffb33fb360): memo 2 remove: ts=1000
AsofjoinNode(0xffffb33fb440): Advancing input 2 updated=1
AsofjoinNode(0xffffb33fb440): Emplace: key=10621589963589458110 lhs_latest_row=0 lhs_latest_time=1000
AsofjoinNode(0xffffb33fb440):   i=1 has_entry=1 time=1500 row=1 accepted=1
AsofjoinNode(0xffffb33fb440):   i=2 has_entry=1 time=1500 row=1 accepted=1
AsofjoinNode(0xffffb33fb6a0): produce batch 1:
time:   [
    1970-01-01 00:00:00.001000
  ]
key2:   [
    1970-01-01
  ]
key:   [
    1970-01-03
  ]
l_v0:   [
    2
  ]
r0_v0:   [
    11
  ]
r1_v0:   [
    1970-01-01 00:00:00.101
  ]

AsofjoinNode(0xffffc2670910): received batch from input 0:
time:   [
    1970-01-01 00:00:00.002000
  ]
key2:   [
    1970-01-01
  ]
key:   [
    1970-01-02
  ]
l_v0:   [
    3
  ]

AsofjoinNode(0xffffb33fb440): Advancing input 1
AsofjoinNode(0xffffb33fb360): memo 1 remove: ts=2000
AsofjoinNode(0xffffb33fb440): Advancing input 2
AsofjoinNode(0xffffb33fb360): memo 2 remove: ts=2000
AsofjoinNode(0xffffb33fb440): Emplace: key=10621589963589458110 lhs_latest_row=0 lhs_latest_time=2000
AsofjoinNode(0xffffb33fb440):   i=1 has_entry=1 time=2500 row=2 accepted=1
AsofjoinNode(0xffffb33fb440):   i=2 has_entry=1 time=2500 row=2 accepted=1
AsofjoinNode(0xffffb33fb6a0): produce batch 2:
time:   [
    1970-01-01 00:00:00.002000
  ]
key2:   [
    1970-01-01
  ]
key:   [
    1970-01-02
  ]
l_v0:   [
    3
  ]
r0_v0:   [
    12
  ]
r1_v0:   [
    1970-01-01 00:00:00.102
  ]

Comparing flattened expected table:
time: timestamp[us, tz=UTC]
key2: date32[day]
key: date32[day]
l_v0: uint16
r0_v0: double
r1_v0: timestamp[ms, tz=UTC]
----
time:
  [
    [
      1970-01-01 00:00:00.000000,
      1970-01-01 00:00:00.001000,
      1970-01-01 00:00:00.002000
    ]
  ]
key2:
  [
    [
      1970-01-01,
      1970-01-01,
      1970-01-01
    ]
  ]
key:
  [
    [
      1970-01-02,
      1970-01-03,
      1970-01-02
    ]
  ]
l_v0:
  [
    [
      1,
      2,
      3
    ]
  ]
r0_v0:
  [
    [
      10,
      null,
      12
    ]
  ]
r1_v0:
  [
    [
      1970-01-01 00:00:00.100,
      null,
      1970-01-01 00:00:00.102
    ]
  ]

with flattened result table:
time: timestamp[us, tz=UTC]
key2: date32[day]
key: date32[day]
l_v0: uint16
r0_v0: double
r1_v0: timestamp[ms, tz=UTC]
----
time:
  [
    [
      1970-01-01 00:00:00.000000
    ],
    [
      1970-01-01 00:00:00.001000
    ],
    [
      1970-01-01 00:00:00.002000
    ]
  ]
key2:
  [
    [
      1970-01-01
    ],
    [
      1970-01-01
    ],
    [
      1970-01-01
    ]
  ]
key:
  [
    [
      1970-01-02
    ],
    [
      1970-01-03
    ],
    [
      1970-01-02
    ]
  ]
l_v0:
  [
    [
      1
    ],
    [
      2
    ],
    [
      3
    ]
  ]
r0_v0:
  [
    [
      10
    ],
    [
      11
    ],
    [
      12
    ]
  ]
r1_v0:
  [
    [
      1970-01-01 00:00:00.100
    ],
    [
      1970-01-01 00:00:00.101
    ],
    [
      1970-01-01 00:00:00.102
    ]
  ]


[  FAILED  ] AsofJoinNodeTest/AsofJoinBasicTest.TestBasic7Forward/2, where GetParam() = AsofJoinBasicParams: MutateByKey (52 ms)
```

### Component(s)

C++, Continuous Integration

rtpsw:

The key hasher is invalidated before the first invocation of `GetKey` (via `GetLatestKey`) after a new batch arrives. In the pre-PR code, this invalidation happens within `Advance`, which is called from `AdvanceAndMemoize` only after `GetLatestKey` is called. The change adds synchronization between the input-receiving- and processing- threads, because avoiding that would require a more complicated and brittle change, e.g., one that involves detecting in the processing thread when a new batch was added to the queue in order to invalidate the key hasher at that time.

### Are these changes tested?

Yes, by existing tests.

No.

* Closes: #36482

### DIFF
--- a/cpp/src/arrow/acero/asof_join_node.cc
+++ b/cpp/src/arrow/acero/asof_join_node.cc
@@ -524,7 +524,7 @@ class KeyHasher {
   size_t index_;
   std::vector<col_index_t> indices_;
   std::vector<KeyColumnMetadata> metadata_;
-  const RecordBatch* batch_;
+  std::atomic<const RecordBatch*> batch_;
   std::vector<HashType> hashes_;
   LightContext ctx_;
   std::vector<KeyColumnArray> column_arrays_;
@@ -819,7 +819,6 @@ class InputState {
         have_active_batch &= !queue_.TryPop();
         if (have_active_batch) {
           DCHECK_GT(queue_.UnsyncFront()->num_rows(), 0);  // empty batches disallowed
-          key_hasher_->Invalidate();  // batch changed - invalidate key hasher's cache
           memo_.UpdateTime(GetTime(queue_.UnsyncFront().get(), 0));  // time changed
         }
       }
@@ -897,7 +896,8 @@ class InputState {
 
   Status Push(const std::shared_ptr<arrow::RecordBatch>& rb) {
     if (rb->num_rows() > 0) {
-      queue_.Push(rb);  // only after above updates - push batch for processing
+      key_hasher_->Invalidate();  // batch changed - invalidate key hasher's cache
+      queue_.Push(rb);            // only now push batch for processing
     } else {
       ++batches_processed_;  // don't enqueue empty batches, just record as processed
     }


pitrou:

pitrou:
@icexelloss Could you take a look at this PR?

pitrou:
@rtpsw It does, though of course the failure wasn't deterministic in the first place.

It also seems to fix things on CI, which is encouraging :-)

rtpsw:
OK, crossing fingers.

pitrou:
Is KeyHasher thread-safe? AFAIU, `key_hasher_` can be used from two threads at a time... `GetKey` calling `key_hasher_->HashesFor` and `Push` calling `key_hasher_->Invalidate`.

pitrou:
I also don't understand why you call `Invalidate` explicitly, while the `KeyHasher` is supposed to invalidate automatically.

However, I see that `KeyHasher` is storing a raw `RecordBatch` pointer, which is problematic. If the previous `std::shared_ptr<RecordBatch>` was destroyed, another `std::shared_ptr<RecordBatch>` could re-allocate the same underlying `RecordBatch` pointer, which is unlikely but not impossible.

Am I missing something here?

rtpsw:
>Is KeyHasher thread-safe? AFAIU, key_hasher_ can be used from two threads at a time... GetKey calling key_hasher_->HashesFor and Push calling key_hasher_->Invalidate.

[This doc](https://github.com/apache/arrow/blob/fd949f8fb416025dc9d77c035c1cac5f45ed1b94/cpp/src/arrow/acero/asof_join_node.cc#L493-L494) addresses thread-safety of `Invalidate` and `HashesFor`.

> I also don't understand why you call Invalidate explicitly, while the KeyHasher is supposed to invalidate automatically.

Presumably, you mean [this invocation](https://github.com/apache/arrow/blob/fd949f8fb416025dc9d77c035c1cac5f45ed1b94/cpp/src/arrow/acero/asof_join_node.cc#L502) of `Invalidate`. Its for safety - it keeps the KeyHasher invalid if any code (now or in the future) fails between that point and [the clean return](https://github.com/apache/arrow/blob/fd949f8fb416025dc9d77c035c1cac5f45ed1b94/cpp/src/arrow/acero/asof_join_node.cc#L518-L519).

I suspect that with the explicit invalidation in place, we could just have `if (batch)` [here](https://github.com/apache/arrow/blob/fd949f8fb416025dc9d77c035c1cac5f45ed1b94/cpp/src/arrow/acero/asof_join_node.cc#L499). I could look into this separately. Given past experience, I'd prefer to make small steps when dealing with as-of-join-node's concurrency.

> However, I see that KeyHasher is storing a raw RecordBatch pointer, which is problematic. If the previous std::shared_ptr<RecordBatch> was destroyed, another std::shared_ptr<RecordBatch> could re-allocate the same underlying RecordBatch pointer, which is unlikely but not impossible.

The explicit invalidation was introduced to resolve exactly this case, which was observed on macOS.

pitrou:
> The explicit invalidation was introduced to resolve exactly this case, which was observed on macOS.

Hmm, this could probably have been fixed more idiomatically by storing a `std::weak_ptr<RecordBatch>`.

westonpace:
I don't have any problems with the fix but I don't understand the bug in the first place.  How was it a race?  The only thread that is looking at these variables is the processing thread correct?

westonpace:
Or maybe the non-determinsitic part was the order in which the batches arrived?  E.g. if all the right input batches arrived before any of the left input batches?

rtpsw:
> Hmm, this could probably have been fixed more idiomatically by storing a std::weak_ptr<RecordBatch> in KeyHasher. And you wouldn't have had to worry about concurrency.

Unfortunately, it's not that simple. The concurrency problem is not caused by an invalid/dangling pointer to a record batch that a weak pointer would have avoided. I'll try to answer both your and Westons's questions.

Some background first. The key hasher is used by the processing thread and caches hashes corresponding to a particular record batch. Therefore, the key hasher (i.e., its cache of hashes) should be invalidated when the record batch being handled by the processing thread changes. The complexity here is that this change is affected by the input-receiving thread. The post-PR code invalidates the key hasher just before this point using a thread-synchronizing operation. This ensures that the cached hashes at this point would not be used by the processing thread calling `HashesFor`; instead, `HashesFor` will compute the hashes for the new record batch the next time `HashesFor` would be invoked.

Given this background, we can see that the problem is not about an invalid/pointer to a record batch but about invalid cached hashes. In the pre-PR code, the failure could happen with the following order of operations:

1. The input-receiving thread pushes a record batch to an empty queue, placing it at the front, thus making it the current record batch being processed.
2. The processing thread invokes `AdvanceAndMemoize`, which invokes `GetLatestKey`, which invokes `GetKey`, which invokes `HashesFor`, which finds hashes that are incorrect for the above batch. This causes the problem, which is a concurrency one because it is driven by a non-deterministic order of operations.

Only thereafter, the processing thread, still in `AdvanceAndMemoize`, invokes `Advance`, which deals with the queue and could detect a new record batch in order to then invalidate the key hasher, but it is too late due to item 2 above.

icexelloss:
@rtpsw The rationale behind the hash caching / invalidation is unclear to me - if the purpose of the KeyHasher here is to compute the hash for each row in the batch, why don't we always keep a immutable tuple of <RecordBatch, std::vector<HashType>> to avoid in the queues to avoid dealing cache invalidation / mutation here?

rtpsw:
> @rtpsw The rationale behind the hash caching / invalidation is unclear to me - if the purpose of the KeyHasher here is to compute the hash for each row in the batch, why don't we always keep a immutable tuple of <RecordBatch, std::vector<HashType>> in the input queues / input state to avoid dealing cache invalidation / mutation here?

The key hasher already encapsulates the batch and its hashes, so I don't see how wrapping the two in a tuple gains anything. As for immutability, you'd need to explain what you have in mind in more detail (or try to implement it) because it could easily degrade performance, which is a major consideration here that greatly impacts the design. The current design has all processing, including for hashes, is done on the processing thread; it sounds like the proposed immutability would shift the computation of hashes to the input-receiving thread, as well as add allocation/deallocation cost for the hashes, and these could easily degrade performance.

> And I don't see clearly why we cannot use the tuple in both places to avoid the cache invalidation issue.

We can and the code effectively does use the "tuple", encapsulated by the key hasher, in both these places. The main question is when and in which thread to compute the hashes. The current code's answer is: upon initial querying for the hashes (of a given batch) within the processing thread.

> One thing that is not clear to me is why do we need rehash (row hash for each batch is deterministic so not clear to me why we need rehash at all) , but it seems to be probably can be avoided by using the recordbatch/hash tuple as well?

This is an orthogonal topic. Rehashing occurs only when the key is a single fixed-width (up to 64 bits, for now) column in which a null value is observed for the first time. So, no rehashing ever occurs if the key spans multiple columns or if the key is a single fixed-width column with no null values. This is a performance optimization - the fast path is when the key is a single fixed-width column without any nulls, which is a common case for which no hashes are computed, since its fixed-width values can be compared directly. The code is being optimistic - it hopes for no null values until it sees the first one.

icexelloss:
> The key hasher already encapsulates the batch and its hashes, so I don't see how wrapping the two in a tuple gains anything. As for immutability, you'd need to explain what you have in mind in more detail (or try to implement it) because it could easily degrade performance, which is a major consideration here that greatly impacts the design. The current design has all processing, including for hashes, done on the processing thread; it sounds like the proposed immutability would shift the computation of hashes to the input-receiving thread, as well as add allocation/deallocation cost for the hashes, and these could easily degrade performance.

Yes the key hasher contains effectively a tuple of <RecordBatch, Hash>. However the tuple is mutable, i.e., the class needs to make sure the Hash is indeed the hash of the RecordBatch (via Invalidate) and IIUC that's where bugs happen. (Either we have a batch and doesn't have the hash, or that we associate the wrong hash with a batch. What I am proposing as an alternative is to have key hasher class produces an *immutable* tuple of <RecordBatch, Hash> so we avoid this wrong hash problem all together. So the code basically becomes:

Currently:
We have queues of RecordBatches, when hash for a RecordBatch is needed, we call key hasher to get the hash (and that's where  the bug comes when we cache the wrong hash)

What I am proposing:
We have queues of immutable <RecordBatch, Hash>, the tuple of created when a batch is received (either in the input received thread or processing thread).  Since we only popping batches from queue in the processing thread, we should always have the hash for the batch that we want to process.

In terms of performance, in both cases, we compute hash once for each input batches, so I don't see why performance would degrade. The extra cost of what I proposing is basically allocating one tuple for each input batches, maybe + some shared pointer copies per batch, both seems fairly cheap.

Do you not agree that this would simplify things?

cpp/src/arrow/acero/asof_join_node.cc
```diff
@@ -524,7 +524,7 @@ class KeyHasher {
   size_t index_;
   std::vector<col_index_t> indices_;
   std::vector<KeyColumnMetadata> metadata_;
-  const RecordBatch* batch_;
+  std::atomic<const RecordBatch*> batch_;
```
icexelloss, line 527: Is the KeyHasher class thread safe now? Since with this change there are two thread touching this class, we should clearly document what the thread safety model of this class is. (What happens if the hash is invalidated in the middle of processing?)

rtpsw:
> What I am proposing as an alternative is to have key hasher class produces an immutable tuple  ...

Yes, this was my understanding of your proposal when I wrote [this post](https://github.com/apache/arrow/pull/36499#issuecomment-1624849882), where I noted performance concerns.

> The extra cost of what I proposing is basically allocating one tuple for each input batches, maybe + some shared pointer copies per batch, both seems fairly cheap.

There's also additional deallocations, additional memory needed, additional costs due to computing in one thread and using the results in another thread, and probably more CPU cache misses. It's conceptually simpler but I don't think we can be sure it'd be more performant.

It's hard to predict performance impact of changes to complex concurrent code.  Ultimately, if we want to know, there's no escaping from implementing the alternative and trying it out. It's going to take some effort to change `RecordBatch` to a tuple in a bunch of places, so whether to go down this path is a decision, I guess for you, to make - let me know your call. Note that in this PR, I was aiming for a minimal change; in fact, it's not even new code but a reversion of part of a [previous PR](https://github.com/apache/arrow/pull/36094).

cpp/src/arrow/acero/asof_join_node.cc
rtpsw: > Is the KeyHasher class thread safe now?

I think so - see below for why.

> Since with this change there are two thread using this class, we should clearly document what the thread safety model of this class is.

[This doc](https://github.com/apache/arrow/blob/fd949f8fb416025dc9d77c035c1cac5f45ed1b94/cpp/src/arrow/acero/asof_join_node.cc#L493-L494) describes the supported concurrency. 

> What happens if the hash is invalidated in the middle of processing?

Here's what I think happens. The hashes are always computed for the record batch found at the front of the queue. The critical (and probably rare) case is when the input-receiving thread gets a record batch during the computation of hashes, whence it invalidates the key hasher and pushes the record batch to the queue. In this case, the important point is that this pushed record batch cannot be at the front of the queue because the queue is not empty and the push is to the back of the queue. This invalidation may only lead to a recomputation of hashes for the same record batch at the front of the queue.

icexelloss:
> > What I am proposing as an alternative is to have key hasher class produces an immutable tuple  ...
> 
> Yes, this was my understanding of your proposal when I wrote [this post](https://github.com/apache/arrow/pull/36499#issuecomment-1624849882), where I noted performance concerns.
> 
> > The extra cost of what I proposing is basically allocating one tuple for each input batches, maybe + some shared pointer copies per batch, both seems fairly cheap.
> 
> There's also additional deallocations, additional memory needed, additional costs due to computing in one thread and using the results in another thread, and probably more CPU cache misses. It's conceptually simpler but I don't think we can be sure it'd be more performant.
> 
> It's hard to predict performance impact of changes to complex concurrent code. Ultimately, if we want to know, there's no escaping from implementing the alternative and trying it out. It's going to take some effort to change `RecordBatch` to a tuple in a bunch of places, so whether to go down this path is a decision, I guess for you, to make - let me know your call. Note that in this PR, I was aiming for a minimal change; in fact, it's not even new code but a reversion of part of a [previous PR](https://github.com/apache/arrow/pull/36094).

Ok I think we agree that an immutable tuple design is simpler but maybe have performance implications here. For the sake of pushing this PR forward, I will try to understand your current approach.

> Given this background, we can see that the problem is not about an invalid/pointer to a record batch but about invalid cached hashes. In the pre-PR code, the failure could happen with the following order of operations:

The input-receiving thread pushes a record batch to an empty queue, placing it at the front, thus making it the current record batch being processed.
The processing thread invokes AdvanceAndMemoize, which invokes GetLatestKey, which invokes GetKey, which invokes HashesFor, which finds hashes that are incorrect for the above batch. This causes the problem, which is a concurrency one because it is driven by a non-deterministic order of operations.

In the case here, since the input thread doesn't not `batch_` variable in side `KeyHasher`. I would expect when the processing thread invokes `AdvanceAndMemoize`, it would find the `batch_` is NULLPTR and then compute the hash for the first batch. What do you mean by " finds hashes that are incorrect for the above batch" here?

rtpsw:
>  I would expect when the processing thread invokes AdvanceAndMemoize, it would find the batch_ is NULLPTR

This case doesn't always occur though it is indeed the common one. It occurs when the input-receiving thread invalidates the key hasher while the processing thread is _not_ executing `HashesFor` concurrently. In this case, the invalidation sets `KeyHasher::batch_` to `nullptr`, and the processing threads finds this `nullptr` in a later invocation of  `HashesFor`. 

The other case (also [explained here](https://github.com/apache/arrow/pull/36499#discussion_r1255982569)) is when the processing thread is executing `HashesFor` concurrently with the input-receiving thread executing `Invalidate`. In this case, `KeyHasher::batch_` may be set to `nullptr` first by the invalidation and then set to the batch passed to `HashesFor`, near the end of this function.

So, `AdvanceAndMemoize` may find either a `nullptr` or some batch in `KeyHasher::batch_`.

> What do you mean by " finds hashes that are incorrect for the above batch" here?

I meant a key hasher cache miss, i.e., that the condition in [this if-statement](https://github.com/apache/arrow/blob/fd949f8fb416025dc9d77c035c1cac5f45ed1b94/cpp/src/arrow/acero/asof_join_node.cc#L499) evaluates to `false`.

icexelloss:
> This case doesn't always occur though it is indeed the common one. It occurs when the input-receiving thread invalidates the key hasher while the processing thread is not executing HashesFor concurrently. In this case, the invalidation sets KeyHasher::batch_ to nullptr, and the processing threads finds this nullptr in a later invocation of HashesFor.
> The other case (also https://github.com/apache/arrow/pull/36499#discussion_r1255982569) is when the processing thread is executing HashesFor concurrently with the input-receiving thread executing Invalidate.

I am trying understand the issue before this PR, but you seems to explaining the behavior after the code change here?

rtpsw:
> I am trying understand the issue pre-PR, but you seems to explaining the behavior after the code change here.

Correct, sorry.

> Can you explain your comment above w.r.t to my question below?

Sure. See below.

> I would expect when the processing thread invokes AdvanceAndMemoize, it would find the batch_ is NULLPTR and then compute the hash for the first batch.

In the pre-PR code, this description is correct for the first batch. For the second batch, `AdvanceAndMemoize` [invokes `GetLatestKey`](https://github.com/apache/arrow/blob/fd949f8fb416025dc9d77c035c1cac5f45ed1b94/cpp/src/arrow/acero/asof_join_node.cc#L865) before it [invokes `Advance`](https://github.com/apache/arrow/blob/fd949f8fb416025dc9d77c035c1cac5f45ed1b94/cpp/src/arrow/acero/asof_join_node.cc#L869), which means `HashesFor` (invoked under `GetLatestKey`) finds `KeyHasher::batch_` referring to the first batch instead of to the second. This is what I meant by "finds hashes that are incorrect for the above batch".

Now, taking this a step further, if the two batches have a different address, then things seem OK because the hashes would then be recomputed. I suspect the bug is triggered when the first and second batches happen to have the same address, which we know is a race condition, leading to the use of incorrect hashes. Granted, I haven't yet nailed down the exact triggering condition in the pre-PR code. It would take some effort - let me know if you'd like this investigated. For the time being, in the current version of the PR, I preferred to find the minimal change that works and I can explain.

icexelloss:
> Now, taking this a step further, if the two batches have a different address, then things seem OK because the hashes would then be recomputed. I suspect the bug is triggered when the first and second batches happen to have the same address, which we know is a race condition, leading to the use of incorrect hashes. Granted, I haven't yet nailed down the exact triggering condition in the pre-PR code. It would take some effort - let me know if you'd like this investigated. For the time being, in the current version of the PR, I preferred to find the minimal change that works and I can explain.

If that is case, then I would argue this is not a concurrency issue, but rather, that the KeyHasher class is broken and failed to cache hash correctly. I remembered we had this issue before and I thought one of your PR fixed this issue, but I cannot seem to find the change anymore. (There was one about Allocator reusing the buffer address for the second batch on MacOS or sth)

rtpsw:
> If that is case, then I would argue this is not a concurrency issue, but rather, that the KeyHasher class is broken and failed to cache hash correctly - you can have the same wrong hash issue too if everything is on one thread

I understand that if you take just the key hasher alone then the caching by address can (be driven to) fail using one thread. However, this argument is ignoring invalidation. I could make a similar argument that the KeyHasher with proper invalidation is correct and can be so even when everything is on one thread. I agree that it would be cleaner to completely remove the address checking from the key hasher, since invalidation makes this not needed, and instead just check for `nullptr` (see `if (batch_)` [here](https://github.com/apache/arrow/pull/36499#issuecomment-1623678477)) as set by the invalidation. I did not do it in the current version of this PR because my goal was the simplest change that fixes. I would prefer to do it in a separate PR, but we could decide to do this in one. In any case, when done it would eliminate the need for the above arguments.

> therefore, I don't think moving things to another thread is the right fix.

Presumably, you're referring to the invalidation from the input-receiving thread. Indeed, there is an alternative where the key hasher operations are all done from one thread, which we discussed starting in [this post](https://github.com/apache/arrow/pull/36499#issuecomment-1624267164). It may be considered better by some standard, certainly doable, yet also more complex. This is because, as discussed, `AdvanceAndMemoize` accesses the key hasher before calling `Advance`, which is where a new batch is handled and the key hasher can be invalidated, but for correctness we need to invalidate before accessing the key hasher. To deal with this in one thread, we'd need to detect a new batch at the front of the queue (in `AdvanceAndMemoize` before accessing the key hasher) as a trigger for invalidation. We can't do this by comparing batch addresses, so we'd probably need to change the queue to contain serially numbered batches (rather than just batches) and do the detection by comparing serial numbers.

Note that up until recently, I did not have access to a platform where I could reproduce the kind of failure discussed here. Now that I do, I'm more confident I could locally test changes relevant to this kind of failure.

> I remembered we had this issue before and I thought one of your PR fixed this issue, but I cannot seem to find the change anymore. (There was one about Allocator reusing the buffer address for the second batch on MacOS or sth)

My understanding of the timeline of relevant changes is:

1. #34392 introduced invalidation, which was originally done from the input-receiving thread. IIRC, this is also the PR where we first observed an infrequent failure on macOS and fixed it.
2. #36094 wished to simplify things by moving operations to one thread, but ended up introducing the bug we're dealing with here.
3. #36499 (the current PR) reverts the moving of the key hasher invalidation.

> Also, are you able to confirm that the failure happened pre-PR is because the second batch has the same address as the first batch?

Right now I don't have debug-print confirmation of this or of the opposite. It would take some effort to get - let me know if you want this done.

rtpsw:
I created #36580 as an alternative solution described [here](https://github.com/apache/arrow/pull/36499#issuecomment-1627524361). It passed the internal testing and I'm waiting to see if it passes CI. After that, we could measure performance and decide whether to go with this or that PR.

westonpace:
I've been able to reproduce errors very reliably with this patch (https://github.com/apache/arrow/commit/25163548b6f851990f188db52a69f1fa0a7a4322)

The patch accurately simulates what I think our real world constraints are on ordering and multi-threading:

 * Each input may deliver its next batch at any time (e.g. maybe input 2 batch 0 comes first and maybe input 1 batch 0 comes first)
 * Within in an input we know that batches will be ordered. (e.g. input 2 batch 5 will always come after input 2 batch 4)
 * A single input will not call InputReceived reentrantly (e.g. a call to input 2 batch 4 InputReceived will finish before the call to input 2 batch 5 InputReceived begins)
 * There are no guarantees across inputs (e.g. there is no guarantee input 3 will deliver batch 4 before input 2 delivers batch 100)

This commit actually generates a lot of errors in many different asof join node tests.  Also, unfortunately, the patch proposed here does not fix all of these errors.

rtpsw:
> @rtpsw https://github.com/apache/arrow/pull/36580 is not exactly what I was asking for.

Right, this PR is [the alternative I described](https://github.com/apache/arrow/pull/36499#issuecomment-1627706130). We could also try the alternative you described, where the hashes are paired with the batch, and check its performance, [as discussed](https://github.com/apache/arrow/pull/36499#issuecomment-1625532994).

> There are still two thread reading/writing the batch_index variable which can lead to race conditions / complex threading model.

There is no batch index variable that is shared between threads. There is a member variable `InputState::batch_index_` that is accessed exclusively by the input-receiving thread and used to number the incoming record batches, and there is a separate value `NumberedRecordBatch::index` that (along with the record batch it numbers) is transferred via the queue to the processing thread, where it is used and set to a local variable `batch_index`.

rtpsw:
> 've been able to reproduce errors very reliably with this patch ([2516354](https://github.com/apache/arrow/commit/25163548b6f851990f188db52a69f1fa0a7a4322))

I'm also seeing errors with this patch. This should be very helpful, thanks!

icexelloss (approved):
For the purpose of moving things forward I am going to approve this PR.

At minimum, this PR fixes the sporadic test failures. Although I am not a big fan of the approach here I think it is acceptable.

We can improve the code structure / simply concurrent models (e.g. using immutable <RecordBatch, hash> tuples) as follow up.

As for the potential issues  @westonpace's test util reveals, we can look into it separately as well.

icexelloss:
@westonpace Are you OK with merging this one as is?

rtpsw:
> As for the potential issues @westonpace's test util reveals, we can look into it separately as well.

#36651

westonpace (approved):
Yes, I'm fine with this as an incremental addition.

PR #36510, opened 2023-07-06
PARQUET-2321: [C++] allow customized buffer size when creating ArrowInputStream for a column PageReader

jp0317:
### Rationale for this change

When buffered stream is enabled, all column chunks, regardless of their actual sizes, are currently sharing the same buffer size which is stored in the shared [read properties](https://github.com/apache/arrow/blob/main/cpp/src/parquet/file_reader.cc#L213).  

Given a limited memory budget, one may want to customize buffer size for different column chunks based on their actual size, i.e., smaller chunks will use consume less memory budget for its buffer.

Codes for customizing buffer size and unit tests. Add a new api based on the `ComputeColumnChunkRange` to expose the actual column chunk sizes. 

### Are these changes tested?

Yes.

Extend some APIs on file_reader.h with a new parameter defining the customized buffer size for a column chunk. The default behavior (if the new  parameter is not specified) will not change.


cpp/src/parquet/properties.h
```diff
@@ -64,7 +64,8 @@ class PARQUET_EXPORT ReaderProperties {
   MemoryPool* memory_pool() const { return pool_; }
 
   std::shared_ptr<ArrowInputStream> GetStream(std::shared_ptr<ArrowInputFile> source,
-                                              int64_t start, int64_t num_bytes);
+                                              int64_t start, int64_t num_bytes,
+                                              int64_t buffer_size = -1);
```
wgtmac, line 68: What about `std::optional<int64_t> buffer_size`?

cpp/src/parquet/file_reader.cc
```diff
@@ -66,7 +66,7 @@ static constexpr int64_t kMaxDictHeaderSize = 100;
 RowGroupReader::RowGroupReader(std::unique_ptr<Contents> contents)
     : contents_(std::move(contents)) {}
 
-std::shared_ptr<ColumnReader> RowGroupReader::Column(int i) {
+std::shared_ptr<ColumnReader> RowGroupReader::Column(int i, int64_t buffer_size) {
```
wgtmac, line 69: TBH, this additional parameter looks a little bit weird here.

cpp/src/parquet/file_reader.h
```diff
@@ -189,6 +190,9 @@ class PARQUET_EXPORT ParquetFileReader {
   ::arrow::Future<> WhenBuffered(const std::vector<int>& row_groups,
                                  const std::vector<int>& column_indices) const;
 
+  /// Return the range of the specified column chunk.
+  ::arrow::io::ReadRange GetColumnChunkRange(int row_group_index, int column_index);
```
wgtmac, line 194: This is not used any where?

cpp/src/parquet/file_reader.h
```diff
@@ -44,7 +44,8 @@ class PARQUET_EXPORT RowGroupReader {
   // An implementation of the Contents class is defined in the .cc file
   struct Contents {
     virtual ~Contents() {}
-    virtual std::unique_ptr<PageReader> GetColumnPageReader(int i) = 0;
+    virtual std::unique_ptr<PageReader> GetColumnPageReader(int i,
```
wgtmac, line 47: I didn't see how the file reader deal with this new parameter. Is it intended for the caller to pass a good buffer_size?

wgtmac (commented):
Thanks for making a PR! However, I am really in doubt about its effectiveness. As the buffered stream does not know the page boundary, it can only issue next read in its best effort. As the caller, we don't know the page size either to tune the buffer_size here. IMO, the best solution is to equip PageReader with page index if available, and plan the read range based on the offset index of each page. WDYT?

wgtmac:
Yep, I understand your overall goal is to control the total memory used in the buffered stream.

My main concern is that parameters like `buffer_size` which should be solely controlled by ReaderProperties now it is at large and runs away into the public API of reader. It is yet not easy for end users to tune the new parameter: e.g. they need to call `ComputeColumnChunkRange` that you have just exposed to get a basic idea of column chunk size and make a decision based on that. Moreover, people in the future may follow the same way to implement their hacky optimizations which tend to make the code base difficult to maintain. It seems to me the benefit it brings does not worth breaking the design.

It would be more favorable if we can add a new setting to ReaderProperties to fine tune the behavior and let the file reader to optimize this internally without changing any public API on the reader side.

mapleFU:
Personally I think create column chunk with a buffer size is tricky. And maybe something like:

```
class ColumnReaderProperties {
   void set_buffer_size(int64_t);
   int64_t buffer_size();
};

class ReaderProperties {
   // default?
   int64_t buffer_size();
   // 
   int64_t buffer_size(int column_id);

private:
  int64_t default_buffer_size_;
  // col -> 
  std::map<int, ColumnReaderProperties> column_properties_;
};
```

would this make code better for this?

cpp/src/parquet/file_reader.cc
```diff
@@ -66,7 +66,8 @@ static constexpr int64_t kMaxDictHeaderSize = 100;
 RowGroupReader::RowGroupReader(std::unique_ptr<Contents> contents)
     : contents_(std::move(contents)) {}
 
-std::shared_ptr<ColumnReader> RowGroupReader::Column(int i) {
+std::shared_ptr<ColumnReader> RowGroupReader::Column(
+    int i, std::optional<ReaderProperties> prop) {
```
wgtmac, line 70: This does not look better than the proposed `ColumnProperties` from @mapleFU

mapleFU: Seems that `RowGroupReader` would already has `properties_`. Could you use that?

wgtmac (changes requested):

cpp/src/parquet/file_reader.cc
mapleFU: @jp0317 In `WriterProperties`, it has a Column-specific writer property, and when column-level property doesn't exist, it will refer to original property. Why not just use that to handling?

And under that, `Column` will get `buffer_size()` without breaking the previous interface.

cpp/src/parquet/file_reader.h
```diff
@@ -44,7 +44,8 @@ class PARQUET_EXPORT RowGroupReader {
   // An implementation of the Contents class is defined in the .cc file
   struct Contents {
     virtual ~Contents() {}
-    virtual std::unique_ptr<PageReader> GetColumnPageReader(int i) = 0;
+    virtual std::unique_ptr<PageReader> GetColumnPageReader(
```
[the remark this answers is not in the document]
wgtmac: `GetColumnChunkRange` is just an internal implementation to roughly get the read range of column chunk. In fact, the exact range of a column chunk can be more complex if IndexPage and ColumnMetaData are considered. It would be a burden to make it public.

cpp/src/parquet/file_reader.cc
wgtmac: Is it an overkill to setup a column-level buffer size of each row group? If you really need control in this granularity, you can create one reader for each row group. Probably enabling buffered stream in this case is not a good idea as its behavior is not that predicable.

ReaderProperties is immutable once reader has been created, mutating it seems not a good idea to me.

cpp/src/parquet/properties.h
```diff
@@ -79,10 +93,45 @@ class PARQUET_EXPORT ReaderProperties {
...
-  void set_buffer_size(int64_t size) { buffer_size_ = size; }
+  /// Return the default size of the buffered stream buffer.
+  int64_t buffer_size() const { return default_column_reader_properties_.buffer_size(); }
+  /// Set the default size of the buffered stream buffer in bytes.
+  void set_buffer_size(int64_t size) {
+    default_column_reader_properties_.set_buffer_size(size);
+  }
+
+  /// Return the size of the buffered stream buffer for a column chunk.
+  int64_t buffer_size(int row_group_index, int column_index) const {
```
wgtmac, line 104: The current change looks better than leaking the property thing into the reader api. However, I'm not sure if it is appropriate to introduce concept like `row_group_index` into ReaderProperties, especially when `ColumnReaderPropertiesMap` is organized into a two-layer design. Sorry but I would -1 on making such change on optimization like this.

WDYT? @pitrou @emkornfield @mapleFU

cpp/src/parquet/properties.h
```diff
@@ -56,15 +57,28 @@ constexpr int32_t kDefaultThriftStringSizeLimit = 100 * 1000 * 1000;
...
   explicit ReaderProperties(MemoryPool* pool = ::arrow::default_memory_pool())
       : pool_(pool) {}
 
   MemoryPool* memory_pool() const { return pool_; }
 
-  std::shared_ptr<ArrowInputStream> GetStream(std::shared_ptr<ArrowInputFile> source,
-                                              int64_t start, int64_t num_bytes);
+  std::shared_ptr<ArrowInputStream> GetStream(
+      std::shared_ptr<ArrowInputFile> source, int64_t start, int64_t num_bytes,
+      std::optional<int64_t> buffer_size = std::nullopt);
```
mapleFU, line 81: why it need a input `buffer_size` when having a PropertiesMap?

pitrou:
Let's step back a bit here. You are talking about limited memory use cases. The default buffer size is `kDefaultBufferSize = 4096 * 4`. Is there a situation where this is too large?

mapleFU (commented):
I'm just back from a bussiness trip so sorry for late reply. I'm generall ok with the change. But it plan the `buffer_size` for per-rowgroup and per-column. Seems that maybe you would like to plan the buffer-size defined by per-row-group?

I guess we'd better plan the `batch_size` per column first. Configuring it per-row-group and per-column is a-bit hard for user.

cpp/src/parquet/properties.cc
```diff
@@ -28,16 +28,20 @@
 namespace parquet {
 
 std::shared_ptr<ArrowInputStream> ReaderProperties::GetStream(
-    std::shared_ptr<ArrowInputFile> source, int64_t start, int64_t num_bytes) {
-  if (buffered_stream_enabled_) {
+    std::shared_ptr<ArrowInputFile> source, int64_t start, int64_t num_bytes,
+    std::optional<int64_t> custom_buffer_size) {
```
mapleFU, line 32: Emmm why do we need a `custom_buffer_size` even when we already have a `buffer_size` internal? This makes building a stream has "default buffer size", "buffer size in ReaderProperties", "buffer size in GetStream"...

cpp/src/parquet/properties.h
mapleFU: Personally I think deciding a buffer-size for per-row-group and per-column is to hack... Perhaps user here has define a `CalculateBufferSize(const ColumnChunkMetadata&)`, and caculating this.

it's ok but I guess it's too complex... Why a consistent buffer_size for each column is not ok? cc @pitrou

PR #36514, opened 2023-07-06
GH-36513: [Dev][C#] Add Dependabot configuration for NuGet

### LINKED ISSUE #36513: [Dev][C#] Add Dependabot configuration for NuGet

Target PR examples:
* https://github.com/apache/arrow/pull/36490
* https://github.com/apache/arrow/pull/36491

Document: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#nuget-cli

### Component(s)

C#, Developer Tools

kou:
### Rationale for this change

We don't want to adjust PR titles from Dependabot manually.

Set the commit message prefix for `/csharp/` from Dependabot.
 
### Are these changes tested?

No. I hope that this works well...

No.
* Closes: #36513

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -23,3 +23,9 @@ updates:
       interval: "weekly"
     commit-message:
       prefix: "MINOR: [CI] "
+  - package-ecosystem: "nuget"
+    directory: "/csharp/"
+    schedule:
+      interval: "weekly"
+    commit-message:
+      prefix: "MINOR: [C#] "


assignUser (approved):
+1

kou:
@eerhardt This causes the following PRs:

* #36525
* #36526
* #36527
* #36528
* #36529
* #36530

Are they useful for us? Or should we revert this configuration?

eerhardt:
> Are they useful for us? Or should we revert this configuration?

Some are useful. But some I would suggest not to take. For example https://github.com/apache/arrow/pull/36530.

I would suggest adding this section:

https://github.com/microsoft/semantic-kernel/blob/8d705b30adf0ef94792dacb86424b172b613ffcd/.github/dependabot.yml#L25-L31

For things like System.*, Microsoft.Bcl.*, etc. we want to consciously upgrade major versions. There are reasons why customers want to stay on older versions, for example 6.0 which is "LTS - long term support", and not upgrade to 7.0, which goes out of support sooner than 6.0.

See https://github.com/microsoft/semantic-kernel/pull/1799 for reasoning why staying on lower major versions is preferred in some cases.

kou:
OK. I've opened #36577. Could you review PRs from Dependabot and merge/close them?

